PR #4622, opened 2015-06-25
[meteor run --no-db] will not start the local mongo server

jboulhous:
Hello,
This is not the use case for an app that uses `meteor-platform`.
This useful if we don't need the local mongo server.
Regards.
Jamal

### DIFF
--- a/tools/commands.js
+++ b/tools/commands.js
@@ -194,6 +194,7 @@ var runCommandOptions = {
   requiresApp: true,
   maxArgs: Infinity,
   options: {
+    'no-db': { type: Boolean },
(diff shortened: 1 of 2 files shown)


jboulhous:
I should add this to help.txt

stubailo:
@jboulhous I'm going to soon be working on a solution that simply doesn't start mongo if you don't have the `mongo` package, which seems preferable.

stubailo:
In Meteor 1.2, we have split up meteor-platform so that removing mongo will actually be possible.

jboulhous:
@stubailo thank you.
I think your solution is actually the best

Issue #2177, opened 2014-05-24
Seen this exception in the client log.


inikkor:
Been there for a few release. It doesn't seem to effect anything as far as I can tell

Uncaught SyntaxError: Unexpected token Y sockjs-0.3.4.js:2018
InfoReceiver.doXhr.xo.onfinish sockjs-0.3.4.js:2018
EventEmitter.emit sockjs-0.3.4.js:151
AbstractXHRObject._start.that.xhr.onreadystatechange sockjs-0.3.4.js:848

Using Chrome Version 35.0.1916.114 m on Windows 8

TravisDart:
An explanation of this error is given in the [Unofficial Meteor Faq](https://github.com/oortcloud/unofficial-meteor-faq#uncaught-syntaxerror-unexpected-token-y):
> Client-side error caused by the server crashing and sending a message starting with "Your app is crashing. Here's the latest log."

Other references to this error:
http://www.telescopeapp.org/docs/customizing/troubleshooting/
https://gist.github.com/Multiply/998e8c5d70881c99d121

RobertLowe:
@glasser this shouldn't be closed, I feel like there is a more elegant way to handle this... 

Could we perhaps catch it and retry would be more appropriate no? 

I think this confuses developers and spams error monitoring services...

Slava:
This error message appears when your server process crashes and Meteor serves a page with console messages starting with "Your app is crashing ...". A "solution" would be to have a mini-ddp server that sends a message like "the app is crashing, refresh the page to display the error messages".

RobertLowe:
@Slava exactly, I feel it should be "just" intelligent enough to understand that these errors happen from time to time (a missed colon for example) and recover or refresh once resolved. I'm not sure a "mini-ddp" is necessary, perhaps just a `try`/`catch` and handling appropriately...

Slava:
@RobertLowe Feel free to propose things in a new issue in a form of concrete code changes.

PR #4639, opened 2015-06-27
Silences #2177

RobertLowe:
Silences the annoying "Uncaught SyntaxError: Unexpected token Y"

Cheers

@Slava you want code, you got code, any thoughts on how the user should be informed?: https://github.com/RobertLowe/meteor/blob/b90927bd2c70d0b07c5caa006f8e3289d1935449/packages/ddp-client/sockjs-0.3.4.js#L2027-L2028

:smile:

### DIFF
--- a/packages/ddp-client/sockjs-0.3.4.js
+++ b/packages/ddp-client/sockjs-0.3.4.js
@@ -2015,7 +2015,24 @@ InfoReceiver.prototype.doXhr = function(base_url, AjaxObject) {
         tref = null;
         if (status === 200) {
             var rtt = (new Date()).getTime() - t0;
-            var info = JSON.parse(text);
+
+// <METEOR>
+  // https://github.com/meteor/meteor/issues/2177
+            // catch parse errors for when meteor server is crashing and
+            // producing error log messages for any request
+            try {
+              var info = JSON.parse(text);
+            } catch (error) {
+              if (text.match(/Your app is crashing. Here's the latest log./)) {
+                // should we hard refresh, pass the error along, or log the
+                // message by some other means like console.log/Meteor.debug
+                return that.emit('finish');
+              } else {
+                throw error;
+              }
+            }
+// </METEOR>
+
             if (typeof info !== 'object') info = {};
             that.emit('finish', info, rtt);
         } else {


Slava:
I think it is a reasonable hack to stop sockjs from throwing exceptions. But the real issue is the UX:

- when a server-side error occurs, there is no way to notice it from the browser
- when sockjs fails, it backs off for too much time, causing the retry time to be postponed for too long, which leads to the developer refreshing the page manually.

I think, even in this solution, it is better to not hard-code the display string in sockjs. That being said, this is a start :)

RobertLowe:
> when a server-side error occurs, there is no way to notice it from the browser

Right, unless we have a 'mini-ddp' like you suggested in #2177. 

Even though the string is presently hardcoded, it still provides from method of detecting a crashed state. Also that string could from from a single source/var/setting, currently "Your app is crashing" is hardcoded in a bunch of [places](https://github.com/meteor/meteor/search?utf8=%E2%9C%93&q=Your+app+is+crashing)

Without the heavier overhead of a 'mini-ddp', we could pass along the state detected [here](https://github.com/meteor/meteor/pull/4639/files#diff-f23e48b45e7069f638988a0b49c9c450R2027)

> when sockjs fails, it backs off for too much time, causing the retry time to be postponed for too long, which leads to the developer refreshing the page manually.

I noticed this as well, at least if we handle with this solution we could potentially lower the retry while the application is in a known crashing state.

Food for though, thanks.

Cheers :smile:

Slava:
Closing this as we are not going to merge it. @anubhav94 is working on an alternative fix for this.

RobertLowe:
@Slava any reference to @anubhav94 fix/discussion?

anubhav94:
Hey @RobertLowe, there is currently no fix / discussion online as of now. I'm still in the early stages of coming up with a solution that will allow users to know of compiler and server side issues within the browser, be able to fix them in their editor and continue developing without having to refresh the page manually.

RobertLowe:
Not directly related to this issue, but closely related @stubailo created [meteor-dev-error-overlay](https://github.com/stubailo/meteor-dev-error-overlay) which tries to provide some feedback when the server is failing by [polling HTTP state](https://github.com/stubailo/meteor-dev-error-overlay/blob/master/dev-error-overlay.js#L21-L22) :+1:

kachkaev:
@RobertLowe [meteor-dev-error-overlay](https://github.com/stubailo/meteor-dev-error-overlay) seems like a nice attempt to make error reporting more informative, but I decided to remove it from my Meteor app shortly after adding. The package triggers calls to the server every seconds (which pollutes the network log) but still does not remove an unclear error message form the console.

Can anyone suggest a way of removing the following in the browser's console log when the server crashes?
```VM1966:2 Uncaught SyntaxError: Unexpected token <```

The message does not confuse me now, but it looks quite unclear at first. Seems like a time-waster that is googled by every new developer. It can be probably easily replaced with something meaningful just with a single `try / catch`. Could anyone from MDG reconsider this issue please?

RobertLowe:
@Slava sorry to beat a dead horse but any thoughts in regards to @kachkaev

Thanks 👍

kachkaev:
@RobertLowe seems like @Slava [left Meteor last year](http://devblog.me/bye-meteor.html). Could anyone else from MDG share their thoughts on this please?

RobertLowe:
Dang, hope he's having a good time at MIT, @stubailo thoughts?

PR #4644, opened 2015-06-28
Fetch emails from GitHub

zimme:
This allows one to determine if an email is verified or not and also gives you all emails the user have registered on GitHub

### DIFF
--- a/packages/github/github_server.js
+++ b/packages/github/github_server.js
@@ -5,12 +5,13 @@ OAuth.registerService('github', 2, null, function(query) {
   var accessToken = getAccessToken(query);
   var identity = getIdentity(accessToken);
   var emails = getEmails(accessToken);
+  var primaryEmail = _.findWhere(emails, {primary: true});
 
   return {
     serviceData: {
       id: identity.id,
       accessToken: OAuth.sealSecret(accessToken),
-      email: identity.email,
+      email: primaryEmail.email,
       username: identity.login,
       emails: emails
     },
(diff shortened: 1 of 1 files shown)


zimme:
This adds one more HTTP request for login with Github which might be a deal breaker?

zimme:
Looking over the github api I found out that the email returned in /user response isn't the github account's primary email but the one which is chosen as the publicly available one.

I could update this PR to use the users primary email as the email for the `email` field

zimme:
I've updated this PR to use the GitHub accounts primary email address instead of the publicly visible one which could be set to none.

zimme:
So I realized this will only work if you provide `user:email` scope or "higher", I'll add a check and only do this if the scope is provided

zimme:
This latest version will use a GitHub account's primary email for the `email` field and provide a list or all of the account's emails in the `emails` field, if one uses the scope `user` or `user:email`. If none of these scopes is used then the `email` field will use the publicly visible email as before and the `emails` field will be an empty array, this is to make sure the emails list is removed if one were to remove the scope.

zimme:
Thinking about it it might not be a good idea to provide potentially defferent data based on the scope which might change rum time with another login request.

What I'd prefer to do is to is to add the `user:email` scope as the default scope which should always be present, to always have the primary email available for github logins

zimme:
I'll start with changing the `email` field to always use the publically visible email as it was before and only add `emails` with all github account's emails if the scope allows it, empty array otherwise.

martijnwalraven:
I agree it makes sense to add the `user:email` scope as the default scope. But we probably want the email field to be the publicly visible email if one is provided, as you suggest in your last comment. If you make these changes, I'll be happy to merge the pull request.

martijnwalraven:
I went ahead and made the changes. Thanks!

Issue #4626, opened 2015-06-25
Remove all verification tokens for a specific email once that email is verified


zimme:
This will break links in the "other" verify email emails, but it will keep the user record cleaner

zimme:
I'll try and put together a PR for this if you don't think this is a weird idea

stubailo:
Yeah, I think this would be a good idea to do.

Thanks for the feature request! We welcome discussions about how to make Meteor better. If you haven't yet, check out our wiki page about [feature requests on GitHub](https://github.com/meteor/meteor/wiki/Contributing-to-Meteor#feature-requests).

If you add a PR for this I'll take a look.

zimme:
I'll try and find sum time

PR #4663, opened 2015-07-01
Remove all verificationTokens for a specific email once verified

zimme:
This will break the other verification email links but it will keep the
user document cleaner. The email will be verified anyways so who cares
that the other verification links isn’t working any more.

#4626, @stubailo, Here you go.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -739,7 +739,7 @@ Meteor.methods({verifyEmail: function (token) {
         {_id: user._id,
          'emails.address': tokenRecord.address},
         {$set: {'emails.$.verified': true},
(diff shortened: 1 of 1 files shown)


stubailo:
LGTM; need to get a security review from one other person first. It looks like if you try to verify again it will just say "Verify email link expired", which sounds fine to me.

@avital?

If we wanted to go above and beyond we could add a second message there like "This email has already been verified.", but this is such an edge case that it probably doesn't matter.

stubailo:
Checked with @Slava to make sure this is OK, going to merge.

stubailo:
Merged: f91da45458f37320569defadeb7b159156b0e2e4

zimme:
:+1:

Issue #4545, opened 2015-06-12
Not able to authenticate user with github if user's email is private on github


frenchbread:
We are using accounts-github package for authenticating a user on a site. Everything works if user's email is public in github profile settings. But if the email is **not publicly** visible and even if permission for email requested, this error pops up:
```
 Exception while invoking method 'login' undefined
```
Also, there is an existing question about this issue on [stackoverflow](http://stackoverflow.com/questions/24689889/the-accounts-github-package-is-causing-my-meteor-user-to-have-a-null-email/24692465#24692465) which might be helpful to review.

anubhav94:
Good point, we should probably be doing something similar to our Google/Facebook packages where we have a default scope of the 'user:emai' and then do a second round trip as seen here: https://developer.github.com/v3/users/emails/. Pull requests welcome!

PR #4666, opened 2015-07-01
Always use identity.email for the email field for github accounts

zimme:
Code introduced in f32f07a can error out because identity.email might
be empty and primaryEmail might be undefined at the same time.

Let’s always use the publicly visible email for the `email` field and if
scope is enough to provide the GitHub account’s list of emails then we
only add that to the `emails` field and developers which use the `user` or
`user:email` scope should use the `emails` field instead of `email` field.

If the scope is revoked, the list of emails will become empty in the
`emails` field.

@martijnwalraven
I think this is the best approach, because with this solution we hand over control to the developers and don't force them to always use the `user:email` scope even if they were to set their own scope.

### DIFF
--- a/packages/github/github_server.js
+++ b/packages/github/github_server.js
@@ -5,13 +5,12 @@ OAuth.registerService('github', 2, null, function(query) {
   var accessToken = getAccessToken(query);
   var identity = getIdentity(accessToken);
   var emails = getEmails(accessToken);
-  var primaryEmail = _.findWhere(emails, {primary: true});
 
   return {
     serviceData: {
       id: identity.id,
       accessToken: OAuth.sealSecret(accessToken),
-      email: identity.email || primaryEmail.email,
+      email: identity.email,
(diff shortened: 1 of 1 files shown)


zimme:
The reason I choose this approach over always adding `user:email` to the scope list is because it's less invasive, and if it turns out people get really confused about the different fields then we consider using the always require `user:email` scope approach.

martijnwalraven:
Thanks for bringing this to my attention. Your pull request wouldn't actually fix #4545, so I've gone ahead and just added a check for the case where primaryEmail is undefined.

I think this approach will give us the best default behavior and still allows people to override the requested permission if they need to do so.

zimme:
:+1:

zimme:
Just make a note that if the GitHub user haven't set a publicly visible email the primary email will be used, because this is new behaviour and might be unexpected for people who are using the `email` field today.

PR #4675, opened 2015-07-02
Adding a DDP Rate Limiter to Meteor Core

anubhav94:
Added a rate limiter at the DDP level to prevent DoS attacks, with default rules added for login, create new user and password reset attempts.


laosb:
I believe it's necessary for a production-ready app. +1 for this.

lorensr:
Does this support rate limiting across multiple servers? Eg my website has a cluster of application servers, and IP X has many connections open, split among my servers. I'd like all of X's activity to count toward a single limit. This type of rate limiting is done by https://github.com/zeroasterisk/Meteor-Throttle with a Mongo collection. While that would be a good default, since it doesn't require any addition setup, it would be great to have an option to configure it to use Redis, which IMO is better suited to the task.

anubhav94:
This doesn't currently support rate limiting across multiple servers, but would create N times the original rate limit if you have N application servers. You brought up a great point and while we don't currently support it, we can keep it in mind for future iterations to the rate limiter!

PR #4718, opened 2015-07-10
Implement "ecmascript" core package using compiler plugins

benjamn:
This is a work in progress in the sense that I know there will be important feedback, but it's feature-complete in the sense that the design seems sound, and everything is working as far as I can tell.

I hope my comments and commit messages explain my approach in more detail, but the short explanation is that we can allow compiler plugins for the `.js` extension, as long as we simply fall back to hard-coded JS support when no plugin is registered for the `.js` extension. The `ecmascript` package obviously cannot depend on itself, and therefore it must be written in plain ES5, but all future apps and packages can use the `ecmascript` package by default, and all past packages/apps can opt into using it by running `meteor add ecmascript`.

To do:
- [x] allow packages that `api.use('ecmascript')` to specify that certain files should not be transpiled
- [x] update [babel-compiler/README.md](https://github.com/meteor/meteor/blob/ecmascript-compiler-plugin/packages/babel-compiler/README.md)
- [x] update metadata in [ecmascript/package.js](https://github.com/meteor/meteor/blob/ecmascript-compiler-plugin/packages/ecmascript/package.js)
- [x] verify that source maps work in all contexts
- [x] provide a polyfill for `Promise`
- [ ] ~~provide a polyfill for `Map`~~
- [ ] ~~provide a polyfill for `Set`~~
- [x] Add some tests of the `ecmascript` package.

I'm going to hold off on including `Map` and `Set` polyfills until we've had a chance to vet the https://github.com/zloirock/core-js implementations in [tool code](https://github.com/meteor/meteor/tree/devel/tools).

cc @dgreensp @glasser @Slava @stubailo

### DIFF
--- a/packages/babel-compiler/README.md
+++ b/packages/babel-compiler/README.md
@@ -1,0 +1,50 @@
+[Babel](http://babeljs.io/) is a parser and transpiler for ECMAScript 2015
+syntax and beyond, which enables some upcoming JavaScript syntax features
+to be used in today's browsers and runtimes.
+
+Meteor's Babel support consists of the following core packages:
+
+* `babel-compiler` - Exposes the [Babel API](https://babeljs.io/docs/usage/api/)
+  on the symbol `Babel`.  For example, `Babel.compile(source, options)`.
+
+* `babel-runtime` - Meteor versions of the external helpers used by
+  Babel-generated code.  Meteor's core packages must run on IE 8 without
+  polyfills, so these helpers cannot assume the existence of
+  `Object.defineProperty`, `Object.freeze`, and so on.
+
+### Babel API
+
+The `babel-compiler` package exports the `Babel` symbol, which exposes
+functionality provided by the
+[`meteor-babel`](https://www.npmjs.com/package/meteor-babel) NPM package,
+which is in turn implmented using the
+[`babel-core`](https://www.npmjs.com/package/babel-core) NPM package.
+Note that you can only use the `babel-compiler` package on the server.
+
+Example:
+
+```js
+var babelOptions = Babel.getDefaultOptions();
+
+// Modify the default options, if necessary:
+babelOptions.whitelist = [
+  "es6.blockScoping", // For `let`
+  "es6.arrowFunctions" // For `=>`
+];
+
+var result = Babel.compile(
+  "let square = (x) => x*x;",
+  babelOptions
+);
+
+// result.code will be something like
+// "var square = function (x) {\n  return x * x;\n};"
+```
+
+Use `Babel.compile(source)` to transpile code using a set of default
+options that work well for Meteor code.
+
+Resources:
+
+* [API docs](https://babeljs.io/docs/usage/api/)
+* [List of transformers](https://babeljs.io/docs/usage/transformers/)
--- a/packages/babel-compiler/babel.js
+++ b/packages/babel-compiler/babel.js
@@ -1,0 +1,61 @@
+var meteorBabel = Npm.require('meteor-babel');
+
+/**
+ * Returns a new object containing default options appropriate for
+ */
+function getDefaultOptions(extraFeatures) {
+  if (extraFeatures) {
+    check(extraFeatures, {
+      // Modify options to enable ES2015 module syntax.
+      modules: Match.Optional(Boolean),
+      // Modify options to enable async/await syntax powered by Fibers.
+      meteorAsyncAwait: Match.Optional(Boolean)
+    });
+  }
+
+  // See https://github.com/meteor/babel/blob/master/options.js for more
+  // information about what the default options are.
+  var options = meteorBabel.getDefaultOptions(extraFeatures);
+
+  // The sourceMap option should probably be removed from the default
+  // options returned by meteorBabel.getDefaultOptions.
+  delete options.sourceMap;
+
+  return options;
+}
+
+Babel = {
+  getDefaultOptions: getDefaultOptions,
+
+  compile: function (source, options) {
+    options = options || getDefaultOptions();
+    return meteorBabel.compile(source, options);
+  },
+
+  // Provided for backwards compatibility; prefer Babel.compile.
+  transformMeteor: function (source, extraOptions) {
+    var options = getDefaultOptions();
+
+    if (extraOptions) {
+      if (extraOptions.extraWhitelist) {
+        options.whitelist.push.apply(
+          options.whitelist,
+          extraOptions.extraWhitelist
+        );
+      }
+
+      for (var key in extraOptions) {
+        if (key !== "extraWhitelist" &&
+            hasOwnProperty.call(extraOptions, key)) {
+          options[key] = extraOptions[key];
+        }
+      }
+    }
+
+    return meteorBabel.compile(code, options);
+  },
+
+  setCacheDir: function (cacheDir) {
+    meteorBabel.setCacheDir(cacheDir);
+  }
+};
--- a/packages/babel-compiler/package.js
+++ b/packages/babel-compiler/package.js
@@ -1,0 +1,17 @@
+Package.describe({
+  name: "babel-compiler",
+  summary: "Parser/transpiler for ECMAScript 6+ syntax",
+  // Tracks the npm version below.  Use wrap numbers to increment
+  // without incrementing the npm version.
+  version: '5.7.1'
+});
+
+Npm.depends({
+  'meteor-babel': '0.4.3'
+});
+
+Package.onUse(function (api) {
+  api.addFiles('babel.js', 'server');
+  api.use('check');
+  api.export('Babel', 'server');
+});
--- a/packages/babel-runtime/babel-runtime.js
+++ b/packages/babel-runtime/babel-runtime.js
@@ -1,0 +1,210 @@
+// The name `babelHelpers` is hard-coded in Babel.  Otherwise we would make it
+// something capitalized and more descriptive, like `BabelRuntime`.
+babelHelpers = {
+  // es6.templateLiterals
+  // Constructs the object passed to the tag function in a tagged
+  // template literal.
+  taggedTemplateLiteralLoose: function (strings, raw) {
+    // Babel's own version of this calls Object.freeze on `strings` and
+    // `strings.raw`, but it doesn't seem worth the compatibility and
+    // performance concerns.  If you're writing code against this helper,
+    // don't add properties to these objects.
+    strings.raw = raw;
+    return strings;
+  },
+
+  // es6.classes
+  // Checks that a class constructor is being called with `new`, and throws
+  // an error if it is not.
+  classCallCheck: function (instance, Constructor) {
+    if (!(instance instanceof Constructor)) {
+      throw new TypeError("Cannot call a class as a function");
+    }
+  },
+
+  // es6.classes
+  inherits: function (subClass, superClass) {
+    if (typeof superClass !== "function" && superClass !== null) {
+      throw new TypeError("Super expression must either be null or a function, not " + typeof superClass);
+    }
+
+    if (superClass) {
+      if (Object.create) {
+        // All but IE 8
+        subClass.prototype = Object.create(superClass.prototype, {
+          constructor: {
+            value: subClass,
+            enumerable: false,
+            writable: true,
+            configurable: true
+          }
+        });
+      } else {
+        // IE 8 path.  Slightly worse for modern browsers, because `constructor`
+        // is enumerable and shows up in the inspector unnecessarily.
+        // It's not an "own" property of any instance though.
+        //
+        // For correctness when writing code,
+        // don't enumerate all the own-and-inherited properties of an instance
+        // of a class and expect not to find `constructor` (but who does that?).
+        var F = function () {
+          this.constructor = subClass;
+        };
+        F.prototype = superClass.prototype;
+        subClass.prototype = new F();
+      }
+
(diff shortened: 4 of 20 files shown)


packages/babel-compiler/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+The `babel-compiler` package exports the `Babel` symbol, which is the same
+object you get in Node from `require("babel-core")`.  You can only use it on the
+server.
+
+Example:
+
+```js
+Babel.transform('var square = (x) => x*x;',
+                { whitelist: ['es6.arrowFunctions'] })
```
benjamn: This README is slightly out of date. I'll add that to the to-do list.

packages/babel-compiler/package.js
```diff
@@ -0,0 +1,17 @@
+Package.describe({
+  name: "babel-compiler",
+  summary: "Parser/transpiler for ECMAScript 6+ syntax",
+  // Tracks the npm version below.  Use wrap numbers to increment
+  // without incrementing the npm version.
+  version: '5.6.17'
```
benjamn: I think we should version this package in step with `meteor-babel`, instead of `babel-core`, since a particular version of `meteor-babel` corresponds to a pegged version of `babel-core` anyway.

packages/babel-runtime/babel-runtime.js
```diff
@@ -0,0 +1,102 @@
+// The name `babelHelpers` is hard-coded in Babel.  Otherwise we would make it
+// something capitalized and more descriptive, like `BabelRuntime`.
+babelHelpers = {
```
benjamn: @dgreensp I copied this file from https://github.com/meteor/react-packages/blob/master/packages/babel-runtime/babel-runtime.js recently, but if you've made additional changes they may need to be ported here

packages/babel-runtime/package.js
```diff
@@ -0,0 +1,12 @@
+Package.describe({
+  name: "babel-runtime",
+  summary: "Runtime support for output of Babel transpiler",
+  version: '0.1.0',
```
benjamn: Not really sure how this package should be versioned… other than, of course, semantically.

packages/ecmascript/ecmascript-tests.js
```diff
@@ -0,0 +1,5 @@
+// Write your tests here!
+// Here is an example.
+Tinytest.add('example', function (test) {
+  test.equal(true, true);
+});
```
benjamn: Need to add real tests (perhaps in a different package, a la https://github.com/meteor/react-packages/tree/master/packages/babel-tests), or remove this file.

packages/ecmascript/ecmascript.js
```diff
@@ -0,0 +1 @@
+// Write your package code here!
```
benjamn: Need to remove this file.

packages/ecmascript/package.js
```diff
@@ -0,0 +1,21 @@
+Package.describe({
+  name: 'ecmascript',
+  version: '0.0.1',
+  // Brief, one-line summary of the package.
+  summary: '',
```
benjamn: Need to update these properties.

tools/build-plugin.js
```diff
@@ -202,16 +201,6 @@ export class SourceProcessorSet {
     // don't use iteration functions, so we can return (and start at #1)
     for (let i = 1; i < parts.length; i++) {
       const extension = parts.slice(i).join('.');
-      // We specially handle 'js' in the build tool, because you can't provide a
-      // plugin to handle 'js' files, because the plugin would need to be built
-      // with JavaScript itself!  Places that hardcode JS are tagged with
-      // #HardcodeJs.
```
benjamn: In fact, you *can* provide a plugin to handle `.js` files, that handler just can't be the only way of handling `.js` files, and in particular it can't handle `.js` files within the plugin package itself.

packages/babel-compiler/package.js
dgreensp: Sounds good.  I guess we start at 6.0 or something :)

packages/ecmascript/plugin.js
```diff
@@ -0,0 +1,38 @@
...
+      hash: result.hash,
+      sourceMap: result.map,
+      bare: !!fileOptions.bare
+    });
+  });
+};
+
+BCp.setDiskCacheDirectory = function(cacheDir) {
+  Babel.setCacheDirectory(cacheDir);
+};
```
benjamn: This method dovetails really well with the caching strategy used by https://www.npmjs.com/package/meteor-babel!

packages/babel-runtime/babel-runtime.js
dgreensp: I've pushed changes in the last hour.  Hopefully the last for a while.

dgreensp:
I want to have a conversation at some point about how you opt into the polyfills, and whether they go in the global environment.

If the polyfills live in `ecmascript` or are implied by it, but they go into the global environment, then if any package you use uses `ecmascript`, you get them in the global environment, which doesn't seem like the way that should be determined.  There's also a direction we could go where Map/Set/Promise aren't global and are more like imported modules that happen to be written to the ES6 spec, but I suspect that will have downsides.

We might as well just have Meteor 1.2 put Map, Set, and Promise in the global environment on both client and server as part of the platform.  (And then I would argue to throw in es5-shim while we're at it!)

tools/build-plugin.js
glasser: Hmm, I think removing `this._hardcodeJs` isn't correct, although I think maybe it was a bad name for the option to this class. The point here is that there is a default compiler for `*.js` but no default linter (unless you changed that in this PR).  By dropping `this._hardcodeJs` it will now make the linter code think that there is a linter for `*.js` even if there isn't, which isn't correct (and probably will cause it to crash when it looks for a sourceProcessors field on the returned classification).

(Not doing a full review this minute, was just poking to make sure you were working on a suitably new version of code (w/ SourceProcessorSet) and you are.)

benjamn: Brought back `this._hardcodeJs` in the current version.

packages/babel-compiler/.versions
```diff
@@ -0,0 +1,3 @@
+babel-compiler@5.4.7
```
glasser: Don't think we need to check this file in. I think it's only created when you run `publish` or `publish-for-arch` and not when you publish things as part of a release? Do you remember what created it? It can probably be deleted. (Does this package have binary components?)

packages/babel-compiler/babel.js
```diff
@@ -0,0 +1,34 @@
...
+          options[key] = extraOptions[key];
+        }
+      }
+    }
+
+    return meteorBabel.compile(code, options);
+  },
+
+  setCacheDirectory: function (cacheDir) {
+    meteorBabel.setCacheDir(cacheDir);
```
glasser: Seems odd to call this method setCacheDir in one place and setCacheDirectory in another.

packages/babel-compiler/package.js
glasser: Don't forget you can also do eg `5.6.17_1` although there are some weird things about underscored versions like that it's hard to properly make prereleases for them.

packages/babel-runtime/.versions
```diff
@@ -0,0 +1,3 @@
+babel-runtime@0.1.0
```
glasser: ditto

packages/ecmascript/package.js
```diff
@@ -0,0 +1,21 @@
...
+  // URL to the Git repository containing the source code for this package.
+  git: '',
+  // By default, Meteor will default to using README.md for documentation.
+  // To avoid submitting documentation, set this field to null.
+  documentation: 'README.md'
+});
+
+Package.registerBuildPlugin({
+  name: 'compile-ecmascript',
+  use: ['compiler-plugin', 'babel-compiler@5.6.17'],
```
glasser: Once @dgreensp reviews https://rbcommons.com/s/meteor/r/56/ this will change to `isobuild:compiler-plugin@1.0.0`.

packages/ecmascript/plugin.js
```diff
@@ -0,0 +1,45 @@
...
+    if (fileOptions.transpile !== false) {
+      var result = Babel.transformMeteor(source, {
+        sourceMap: true,
+        filename: inputFilePath,
+        sourceFileName: "/" + inputFilePath,
+        sourceMapName: "/" + outputFilePath + ".map"
+      });
+
+      toBeAdded.data = result.code;
+      toBeAdded.hsah = result.hash;
```
glasser: `hash`

tools/compiler-plugin.js
```diff
@@ -420,7 +413,7 @@ var PackageSourceBatch = function (unibuild, processor) {
         }
       } else {
         sourceProcessor = sourceProcessorSet.getByExtension(extension);
-        if (! sourceProcessor) {
+        if (! sourceProcessor && extension !== 'js') {
```
glasser: Probably could use some variant of the comment removed a few lines up

tools/isopack.js
```diff
@@ -1117,7 +1117,8 @@ _.extend(Isopack.prototype, {
           // If we're going to write a legacy prelink file later, track the
           // original form of the resource object (with the source in a Buffer,
           // etc) instead of the later version.  #HardcodeJs
-          if (writeLegacyBuilds && resource.type === "source" &&
+          if (writeLegacyBuilds &&
```
glasser: So I think you need to tweak this logic somehow:

```
        if (writeLegacyBuilds) {
          if (_.any(unibuild.resources, function (resource) {
            return resource.type === "source" && resource.extension !== "js"
              && resource.extension !== "css";
          })) {
            // This package cannot be represented as an isopack-1 Isopack
            // because it uses a file implemented by registerCompiler other than
            // the very basic JS and CSS types.
            writeLegacyBuilds = false;
          }
        }
```

The idea here is that if a package only contains JS and CSS files and static assets/head/body (or the output of "legacy source handlers", which still run at package build time and only create JS/CSS/assets/head/body), then we can publish a "fat" package that has files in the pre-1.2 format.  (This code runs only for publishing packages and for processing packages downloaded into the tropohouse (since we aren't making a separate tropohouse for new vs old packages).)

But that should *NOT* be the case if the package contains a "*.js" file but that file is being processed by ecmascript (or something else).  So we should figure out how to turn off writeLegacyBuilds in that case.

(Note that any case where there is a non-trivial source file, there must be a dependency on `isobuild:compiler-plugin`, which should make old versions of the tool not choose the package in version solver anyway.  So in a sense this is just an optimization: not creating the legacy part of the package for package versions that are never going to be looked at by legacy tools anyway. But it's an important optimization, I think.)

packages/babel-compiler/babel.js
benjamn: I'll change `setCacheDirectory` to `setCacheDir`, so that I don't have to change the API of the NPM package.

packages/ecmascript/plugin.js
benjamn: good catch, thanks

tools/isopack.js
benjamn: Hmm, it seems like we need to call `compiler.getActivePluginPackages` to determine if the current isopack is processed by any plugins that register compiler handlers for `.js` files, but it's too early to call `compiler.getActivePluginPackages` here, so that optimization may not be available?

glasser: We could just put something on the resource json thing when we generate it
that says that it is (or isn't) from the default handler...

glasser:
Just did some refactorings on classifyFilename that probably broke your merge.

benjamn:
Just rebased against them!

tools/isopack.js
benjamn: Done: https://github.com/meteor/meteor/commit/95de49841784843c3f9d83837cbc1f00fc5a4956

benjamn:
@glasser @dgreensp I believe this PR is ready for review (to be merged into `origin/batch-plugins`, that is)

glasser:
Just to be super explicit — this change does not provide a built in way for compiler plugins to produce ES2015 output (not that coffeescript etc does that anyway) but it does at least provide some packages that those plugins could use?  (I'm not criticizing, just trying to make sure I understand the scope.)

benjamn:
That's correct. Compiler plugins that want to target ES2015 can depend directly on `babel-compiler` and use `Babel.compile` themselves. If those plugins want to handle `.js` files, then the `ecmascript` package may have to be removed before the other plugin can be enabled.

tools/isopack.js
```diff
@@ -919,6 +919,45 @@ _.extend(Isopack.prototype, {
...
+          resource.extension === "css"
+        );
+      }
+
+      return false;
+    }
+
+    return writeLegacyBuilds && ! _.any(
+      this.unibuilds,
+      unibuild => _.any(
```
glasser: think our style guide only lets you drop parens/curlies/return for single-line arrows?

also you could use Array.some if you want for both of these.

benjamn: Fixed to use `Array.prototype.some` and single-line arrow functions: https://github.com/meteor/meteor/pull/4718/files#diff-b34c69e78e7e754e2ecf517e481c3c36R979

benjamn:
@glasser new logic for determining if a resource can be written in a legacy build: https://github.com/meteor/meteor/commit/89bc8c12871ef3cc8251d21f20dccc221588c193

tools/isopack.js
dgreensp: Fancy!! :)

Issue #2587, opened 2014-09-15
Spacebars : no easy way to handle @first/@last/@index/@odd/@even ?


temsa:
**_2 Upvotes**_ Spacebars is inspired by Handlebars, yet it does not support [@data](http://handlebarsjs.com/reference.html#data) properties as far as I know (at least it is not in the reference).

How should we handle some "{{#each}}" which should not have some element for the last or the first one ?

How should we handle "stripes" as well ?

wursttheke:
Awesome: @index seems to be coming! https://github.com/meteor/meteor/pull/3716

Maybe the rest can be build on top of that?

rossmartin:
Please this would be awesome.

Slava:
@wursttheke it is not hard to implement all of these on top of @index. That's not a priority though.

wursttheke:
Thanks for your answer, @Slava. You're right. Having index is great and the rest shouldn't be too much of a problem. I guess it could be done easily inside a helper function.

mitar:
But will parser allow us to use `@first` as a helper name?

Slava:
@mitar no, it will not. You can implement others in helpers.

mitar:
Crying. :-(

mitar:
@dgreensp [said that if we do a pull request](https://github.com/meteor/meteor/commit/afd240fa6fb1bff4263e7d398e018b694a7033f4#commitcomment-11166036) for `@first` and others, he will take it. :-)

calvin-evans:
:+1:

hwillson:
Migrated over to meteor/blaze#185 - closing here. Thanks!

PR #4729, opened 2015-07-12
Remove docs for @index

chmac:
The docs link to this file as the spacebars readme. But, using `@index` throws:

       client/templates/foo/foo.template.html:15: Expected IDENTIFIER

Should be removed from the docs until it is implemented. Seems to be tracked in #2587 if I understood correctly.

### DIFF
--- a/packages/spacebars/README.md
+++ b/packages/spacebars/README.md
@@ -365,9 +365,6 @@ for example), but it may also be a plain JavaScript array, `null`, or
 An "else" section may be provided, which is used (with no new data
 context) if there are zero items in the sequence at any time.
 
-You can use a special variable `@index` in the body of `#each` to get the
-0-based index of the currently rendered value in the sequence.
-
 ### Reactivity Model for Each
 
 When the argument to `#each` changes, the DOM is always updated to reflect the


Slava:
@index is implemented on devel and this doc string is on devel.

Slava:
I agree, we should release it soon :) The fact that it's been in docs for over 4 month but not in a release is disturbing.

chmac:
Maybe change the link from the docs to point to the released branch instead?

Slava:
The link is already only on devel

Slava:
I see, I misunderstood what you meant in the last comment. I fixed the link in docs to point to the master branch on GitHub.

chmac:
Awesome, thanks a lot. :+1:

Slava:
Thanks for noticing this.

RobertLowe:
@Slava maybe have some continuous deployments for docs, something like: `devel.docs.meteor.com` :neckbeard:

Slava:
@RobertLowe I am not working on the docs at the moment. I think @dominicnguyen and @zhenyasav would be interested to hear your ideas, later. I am all for not deploying the docs manually.

RobertLowe:
Thanks @Slava sorry, I bug you so much, keep being awesome! :smile:

Slava:
I think it is a useful comment and a good idea. Right now the older docs are not available online either.

PR #4735, opened 2015-07-13
Improve Meteor.wrapAsync doc.

trusktr:


trusktr:
I added "The signature of the callback of the wrapped function should be `function(error, result){}`.". Is that right?

stubailo:
Yeah I think so. Thanks!

I think this function should be improved overall, but this is a good start.

stubailo:
Merged: 1f47600bb77b7c97e9b197236ca6a5a80b53a595

rclai:
@trusktr also, if your callback differs from `function (err, result) {}`, you would need to use `Meteor.bindEnvironment()` to handle it. See this [forum post](https://forums.meteor.com/t/how-do-fibers-and-meteor-asyncwrap-work/6087/25?u=lai).

trusktr:
@rclai I think you can just wrap the code with a function too?

```
Meteor.wrapAsync(function(callback) {
  doSomethingThatAcceptsCallbackWithDifferentSignature(function(some, thing, else, error) {
    callback(error, {some: some, thing: thing, else: else})
  })
})()
```

stubailo:
Yeah, I would do the above and continue using `wrapAsync`.

rclai:
Oh!  That's the Golden example!

PR #4736, opened 2015-07-13
Add `delete` to ReactiveDict

### LINKED ISSUE #4438: Add `delete` to ReactiveDict
Right now, you have to `set` to `undefined`, which is not really semantic. Also, it's not documented anywhere.

RobertLowe:
Pull request roulette. :+1: 

Cheers :smile: 

Fixes #4438

### DIFF
--- a/packages/reactive-dict/reactive-dict.js
+++ b/packages/reactive-dict/reactive-dict.js
@@ -165,8 +165,9 @@ _.extend(ReactiveDict.prototype, {
   },
 
   delete: function(key) {
-    var self    = this;
-    var oldKeys = self.keys;
+    var self      = this;
+    var oldKeys   = self.keys;
+    var didRemove = false;
 
     _.each(oldKeys, function(value, oldKey) {
       if (key == oldKey){
@@ -174,12 +175,12 @@ _.extend(ReactiveDict.prototype, {
         changed(self.keyValueDeps[oldKey][value]);
         changed(self.keyValueDeps[oldKey]['undefined']);
         delete self.keys[oldKey]; // clean up
+        self.allDeps.changed();
+        didRemove = true;
       }
     });
 
-    // trigger a change regardless if any matching key found...
-    self.allDeps.changed();
-
+    return didRemove;
   },
 
   _setObject: function (object) {
--- a/packages/reactive-dict/reactive-dict.js
+++ b/packages/reactive-dict/reactive-dict.js
@@ -169,16 +169,14 @@ _.extend(ReactiveDict.prototype, {
     var oldKeys   = self.keys;
     var didRemove = false;
 
-    _.each(oldKeys, function(value, oldKey) {
-      if (key == oldKey){
-        changed(self.keyDeps[oldKey]);
-        changed(self.keyValueDeps[oldKey][value]);
-        changed(self.keyValueDeps[oldKey]['undefined']);
-        delete self.keys[oldKey]; // clean up
-        self.allDeps.changed();
-        didRemove = true;
-      }
-    });
+    if (_.has(oldKeys, key)) {
+      changed(self.keyDeps[key]);
+      changed(self.keyValueDeps[key][oldKeys[key]]);
+      changed(self.keyValueDeps[key]['undefined']);
+      delete self.keys[key]; // clean up
+      self.allDeps.changed();
+      didRemove = true;
+    }
 
     return didRemove;
   },
--- a/packages/reactive-dict/reactive-dict.js
+++ b/packages/reactive-dict/reactive-dict.js
@@ -164,6 +164,24 @@ _.extend(ReactiveDict.prototype, {
     
   },
 
+  delete: function(key) {
+    var self    = this;
+    var oldKeys = self.keys;
+
+    _.each(oldKeys, function(value, oldKey) {
+      if (key == oldKey){
+        changed(self.keyDeps[oldKey]);
+        changed(self.keyValueDeps[oldKey][value]);
+        changed(self.keyValueDeps[oldKey]['undefined']);
+        delete self.keys[oldKey]; // clean up
+      }
+    });
+
+    // trigger a change regardless if any matching key found...
+    self.allDeps.changed();
+
+  },
+
   _setObject: function (object) {
     var self = this;
 
--- a/packages/reactive-dict/reactive-dict-tests.js
+++ b/packages/reactive-dict/reactive-dict-tests.js
@@ -70,7 +70,9 @@ Tinytest.add('ReactiveDict - delete(key) works', function (test) {
   test.equal(equalsUndefined, false);
   test.equal(all, {foo: 'bar', bar: 'foo'});
 
-  dict.delete('foo');
+  var didRemove = dict.delete('foo');
+  test.equal(didRemove, true);
+
   Tracker.flush();
 
   test.isUndefined(val);
@@ -78,4 +80,6 @@ Tinytest.add('ReactiveDict - delete(key) works', function (test) {
   test.equal(equalsUndefined, true);
   test.equal(all, {bar: 'foo'});
 
+  didRemove = dict.delete('barfoobar');
+  test.equal(didRemove, false);
 });
--- a/packages/reactive-dict/reactive-dict-tests.js
+++ b/packages/reactive-dict/reactive-dict-tests.js
@@ -42,3 +42,40 @@ Tinytest.add('ReactiveDict - clear() works', function (test) {
   test.equal(equalsUndefined, true);
   test.equal(all, {});
 });
+
+
+
+Tinytest.add('ReactiveDict - delete(key) works', function (test) {
+  var dict = new ReactiveDict;
+  dict.set('foo', 'bar');
+  dict.set('bar', 'foo');
+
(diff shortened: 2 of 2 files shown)


packages/reactive-dict/reactive-dict.js
```diff
@@ -164,6 +164,24 @@ _.extend(ReactiveDict.prototype, {
...
+        changed(self.keyDeps[oldKey]);
+        changed(self.keyValueDeps[oldKey][value]);
+        changed(self.keyValueDeps[oldKey]['undefined']);
+        delete self.keys[oldKey]; // clean up
+      }
+    });
+
+    // trigger a change regardless if any matching key found...
+    self.allDeps.changed();
+
```
RobertLowe: The reasoning behind putting the `allDeps.changed` and *always* changing, was for more predictability for end-users, we may want to move it up. :smile: 

Let me know

rclai:
Was there a consensus on how to name this method? I remember some issue lying around where there was lots of going back and forth about the naming of this; either `unset` or `remove`, etc.

stubailo:
IMO we should try to match the ES6 Map API as much as possible (although that ship has partially sailed, I still have hope)

So `delete` seems to be ideal: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete

packages/reactive-dict/reactive-dict.js
stubailo: I would be against triggering reactive updates when nothing has changed. Avoiding unnecessary reruns is kind of the whole point of Tracker.

rclai:
I have no problem with that.

packages/reactive-dict/reactive-dict.js
RobertLowe: @stubailo I totally agree, I don't have strong feelings either way.

I mean, that feels like the technically right way to handle it, but for users/devs, I can picture situations where a user would delete a key, and expect a rerun. 

Let's start with the more performant way (only when changed), and if anyone makes a feature request for always triggering reactive updates it could be handle via an option...

@glasser any thoughts?

RobertLowe:
@stubailo sure, it seems like it's only missing the `return true/false`.

I mean, we could have aliases for `unset/remove/delete`, I don't have strong opinions.

packages/reactive-dict/reactive-dict.js
stubailo: I guess the theory of Tracker is that people should never "expect a rerun" - if you write your autoruns in an idempotent way then there shouldn't be side-effects that you are expecting. It's more about eventual consistency than events.

stubailo:
@RobertLowe I feel like having aliases is a Ruby thing - AFAIK no other community does that.

RobertLowe:
@stubailo I've mined too many gems... I enjoy not needing to think about a method name, but I'll go with whatever best practices are here. :smile: 

Updated with minor changes to reactivity & return

RobertLowe:
@stubailo all tidy :smile:

stubailo:
@dgreensp can you take a look?

packages/reactive-dict/reactive-dict.js
```diff
@@ -164,6 +164,23 @@ _.extend(ReactiveDict.prototype, {
     
   },
 
+  delete: function(key) {
+    var self      = this;
```
stubailo: Can we un-align this? We don't align assignments anywhere else in the code AFAIK

packages/reactive-dict/reactive-dict.js
```diff
@@ -164,6 +164,23 @@ _.extend(ReactiveDict.prototype, {
...
+    var oldKeys   = self.keys;
+    var didRemove = false;
+
+    if (_.has(oldKeys, key)) {
+      changed(self.keyDeps[key]);
+      changed(self.keyValueDeps[key][oldKeys[key]]);
+      changed(self.keyValueDeps[key]['undefined']);
+      delete self.keys[key]; // clean up
```
stubailo: I wonder if it's a big problem that we never remove the keyDeps or keyValueDeps? Seems like if you keep changing keys, you will end up with a memory leak...

RobertLowe: right, I hadn't thought of that, we should probably clean up as much as possible.

`clear` should clean up as well, `clear`'s clean up should be a different gh issue, no?

packages/reactive-dict/reactive-dict.js
RobertLowe: aww but they're so pretty and readable, I can un-align... :-1: :smile:

packages/reactive-dict/reactive-dict.js
dgreensp: A Computation can listen on any key, whether or not it exists in the dict, and we still track it.  So removing a key doesn't mean we should go and remove the Dependency objects.  However, it seems like we should _eventually_ clean up the keys.  We currently clean up the values in `keyValueDeps` but not the keys (the second level of the map but not the first), and we never clean up `keyDeps`.  I'm not sure where to put the clean-up code.

packages/reactive-dict/reactive-dict.js
```diff
@@ -164,6 +164,23 @@ _.extend(ReactiveDict.prototype, {
     
   },
 
+  delete: function(key) {
+    var self      = this;
+    var oldKeys   = self.keys;
+    var didRemove = false;
+
+    if (_.has(oldKeys, key)) {
+      changed(self.keyDeps[key]);
```
dgreensp: It feels like most of this method's code could be turned into a call to `this.set(key, undefined)`.

dgreensp:
There's a comment in `setDefault` that basically says to come back to that code when it's possible to clear entries.  Admittedly we already have `clear()`, but that comment should probably be addressed now.

packages/reactive-dict/reactive-dict.js
RobertLowe: @dgreensp At first glance totally, but I think some things pull away from that:

1. A more ES6-ish API (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/delete)
2. I think it's simpler to keep the two apart and distinct

RobertLowe:
@dgreensp right, I think that comment is just outdated now as we have `clear` & `delete`

packages/reactive-dict/reactive-dict.js
dgreensp: Sorry, I didn't mean API-wise, I meant to simplify the implementation and reduce duplication of code.

This change would also have a side effect that code like `Session.get("asdf")` would no longer be invalidated when `asdf` goes from `undefined` to absent (in which case the code still returns `undefined`), but that seems like an improvement if anything.

dgreensp:
The comment raises a question that is still relevant, namely, in a world where keys are removable, should `setDefault` only set if the key is absent, and not if it is present but undefined?  Is it too late for that?

RobertLowe:
@dgreensp I see what you mean. I think a user would just expect `setDefault` to always keep it's default, even if `delete`'d. There should be something analogous to `clearDefault` in my opinion. Thoughts?

dgreensp:
Oh, I was confused, for a minute I thought `setDefault` currently would replace an `undefined` value.  But it currently only replaces an absent value.  (An `undefined` value is stored as `"undefined"`.)

What confused me is that `Session.set("foo", undefined)` doesn't actually do anything if `asdf` doesn't exist.  It doesn't create a key.  So you can still `setDefault` after that.

`setDefault` shouldn't call `get`, that would introduce a new call to `.depend()` that we probably don't want on a set.

When you say "I think a user would just expect setDefault to always keep it's default, even if delete'd," are you proposing remembering the default and restoring it when the key is deleted?  I'm not sure if I like that or not.  It would certainly require some new code in the implementation to store the defaults.

packages/reactive-dict/reactive-dict.js
stubailo: Is it possible to somehow identify when all of the computations that depend on a dependency have been stopped?

stubailo:
What's the point of `setDefault`? Can we just get rid of it? Is there a concrete use case where it should be used and not `set`?

dgreensp:
It's a very common method.  Our example apps use it; I use it.

The use case is setting an initial value for a Session variable, without overwriting the migrated data.

stubailo:
Ah, I guess it's because of migration, right. I always forget.

dalgard:
What is the status on this?

dgreensp:
Polished (72c6622b4223d90d61d775118530021e0dff207b) and merged.

benjamn:
I was getting the following error when I did `git fetch origin` because there was a branch called `pr/4736`:
```sh
fatal: Cannot fetch both refs/pull/4736/head and refs/heads/pr/4736 to refs/remotes/origin/pr/4736
```
so I renamed that branch to [`pr-4736`](https://github.com/meteor/meteor/tree/pr-4736), just in case you still need it.

dgreensp:
Thanks.  I was trying to push to the PR and I'm not sure what happened.  Maybe PR refs aren't tags or branches and can't be pushed normally?  Eh.  I don't need the branch, will delete.

PR #4801, opened 2015-07-23
Upgrades generate_dev_bundle scripts on both Unix and Windows to use npm 3

martijnwalraven:
Only uses npm 3 for generating the bundles, we'll keep using npm 1.x for everything else for now.

### DIFF
--- a/tools/files.js
+++ b/tools/files.js
@@ -23,7 +23,6 @@ var buildmessage = require('./buildmessage.js');
 var watch = require('./watch.js');
 var fiberHelpers = require('./fiber-helpers.js');
 var colonConverter = require("./colon-converter.js");
-var Console = require("./console.js").Console;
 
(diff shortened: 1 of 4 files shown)


tools/files.js
```diff
@@ -643,7 +642,7 @@ files.freeTempDir = function (tempDir) {
       // Don't crash and print a stack trace because we failed to delete a temp
       // directory. This happens sometimes on Windows and seems to be
       // unavoidable.
-      Console.debug(err);
```
stubailo: Why this change?

martijnwalraven: I was pairing with Slava yesterday and we got a Console is undefined error that we didn't understand. The calls happen in a weird context (responding to a signal and clearing up), so that may have something to do with it (although I have no idea what that would do).

PR #4815, opened 2015-07-25
[perf] Remove oplog notification cloning

arunoda:
We do `EJSON.clone` for every oplog entry we receive. Depending the oplog usage it'll use upto 10% - 40% CPU usage for that.
See below:

screen shot 2015-07-26 at 12 18 24 am
See it live: https://ui.kadira.io/cpf/TJNrApwTkPKMc49uq?metric=costlyPaths&pathId=0

I think this `EJSON.clone` add ~1.5 years back when start writing oplog implementation. It also marked for deleting later on.
I checked the codebase for OplogObserverDriver and it's safe and we don't modify the notification. I think we don't plan to add anymore features/refactors to the OplogObserverDriver soon.
So it's safe to remove this EJSON.clone.

That'll simply improve the oplog performance a lot.

### Test Case

* Here's the repo you can use for the test: https://github.com/arunoda/oplog-perf-test

### DIFF
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -96,8 +96,7 @@ _.extend(OplogHandle.prototype, {
 
     var originalCallback = callback;
     callback = Meteor.bindEnvironment(function (notification) {
-      // XXX can we avoid this clone by making oplog.js careful?
-      originalCallback(EJSON.clone(notification));
+      originalCallback(notification);
     }, function (err) {
       Meteor._debug("Error in oplog callback", err.stack);
     });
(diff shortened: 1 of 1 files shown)


Slava:
Looks good to me, we put these clones here in case the user of the package accidentally modifies the objects from cache.

We have even more cloning within the fields projection.

Maybe we can use Object.freeze instead?

arunoda:
`Object.freeze` sounds great. I just add it.

Slava:
Same goes to other places where the document gets to the publish set

arunoda:
May be we need to do a recursive freeze to the notification as well. Then, I'm quite not sure it's worthy doing.

I'm quite not familiar with publish set?

tmeasday:

PR #4829, opened 2015-07-28
Static HTML package and templating refactor

stubailo:
### Goals

1. Make a package called `static-html` that just compiles `<head>` and `<body>` tags inside `.html` files. Much like templating but without the templates.
2. Refactor `templating` to avoid duplicating code.

### Plan

Split out code that would be shared between `templating` and `static-html` into a new `templating-tools` package. These tools could also be used to simplify implementation of other build plugins, like `simple:markdown-templating`.

This also has the added benefit of moving as much code as humanly possible out of the `templating` package, so that it can be reused in other packages.

### Code changes

1. `templating-tools` package and its README
2. `static-html` package and its README
3. `caching-html-compiler` is not new code; it is just code factored out of the batch plugin version of `templating`, but the README and some comments are new.

### Tests

1. `tools/tests/static-html.js` tests static html and error handling
2. `templating-tools/html-scanner-tests.js` tests `scanHtmlForTags` and `compileTagsWithSpacebars` together.

All unit tests pass on this branch.


packages/caching-compiler/caching-compiler.js
```diff
@@ -1,5 +1,5 @@
-const fs = Plugin.fs;
-const path = Plugin.path;
```
Slava: why?

packages/caching-compiler/multi-file-caching-compiler.js
```diff
@@ -1,4 +1,4 @@
-const path = Plugin.path;
```
Slava: why this change?

packages/caching-html-compiler/caching-html-compiler.js
```diff
@@ -0,0 +1,136 @@
...
+      // XXX this path manipulation may be unnecessarily complex
+      let pathPart = path.dirname(filePath);
+      if (pathPart === '.')
+        pathPart = '';
+      if (pathPart.length && pathPart !== path.sep)
+        pathPart = pathPart + path.sep;
+      const ext = path.extname(filePath);
+      const basename = path.basename(filePath, ext);
+
+      // XXX generate a source map
```
dgreensp: Are these new or existing XXXs?

dgreensp:
I love the docs, and how the scanner and handler are now functions!

I'd still like to see the weird `Template.__checkName  etc` code factored out in such a way that it can be used by Jade -- i.e. not tied to Spacebars.  There are only a couple lines in `compileTagsWithSpacebars` that are Spacebars-specific, which suggests somehow making that function more generic.  That sounds hard though, so what might be easier is leaving that overall function and class as is, and just pulling out the lines that generate JS code into public helpers like `TemplateTools.generateTemplateDefinition`.

Why are bodyAttrs checked for conflicts in two places?

packages/caching-html-compiler/caching-html-compiler.js
stubailo: This was existing.

stubailo:
Yes, there are some more places I could de-duplicate code, namely:

1. Checking body attrs for duplicates - I considered factoring this out, but what would the function do - just check if a value is already in a dictionary? Then the only thing left is throwing the right error, and it didn't seem to me like the error message should always be the same. But if you like, I can factor that part out.
2. The template generation code: I totally agree this should be factored out - will do that next.

dgreensp:
For (1), I didn't mean code duplication, just why are there two different checks.  But now I realize one is for within-a-file collisions, and the other is for global collisions.

stubailo:
@dgreensp that's right. I'll put a comment about that.

packages/caching-html-compiler/README.md
```diff
@@ -0,0 +1,37 @@
...
+2. `static-html`
+3. `simple:markdown-templating`
+
+It provides automatic caching and handles communicating with the build plugin APIs. The actual functions that convert HTML into compiled form are passed in as arguments into the constructor, allowing those functions to be unit tested separately from the caching and file system functionality.
+
+-------
+
+### new CachingHtmlCompiler(name, tagScannerFunc, tagHandlerFunc)
+
+Constructs a new CachingHtmlCompiler that can be passed into templating.
```
stubailo: Typo - this should be registerCompiler

stubailo:
@dgreensp addressed your comments.

packages/templating-tools/compile-tags-with-spacebars.js
```diff
@@ -0,0 +1,107 @@
...
+        const renderFuncCode = SpacebarsCompiler.compile(this.tag.contents, {
+          isBody: true,
+          sourceName: "<body>"
+        });
+
+        // We may be one of many `<body>` tags.
+        this.results.js += `
+Template.body.addContent(${renderFuncCode});
+Meteor.startup(Template.body.renderToDocument);
+`;
```
dgreensp: Can you factor this part out, too?  Then I think it's good to go!

PR #4832, opened 2015-07-29
Fix #4804 and throw error when Mongo selector is array

### LINKED ISSUE #4804: Collection.remove() should throw an error when used with an empty array, instead of removing all items
So, with javascript it is really easy to inadvertently pass an array object `Collection.remove(array)` as the selector, and if that array is the empty array, it deletes all documents in that collection! 

`Collection.remove()` should throw an error if passed an array (like it does in mongo!) or at the very least not delete the whole collection if an empty array gets passed as the selector. This seems like a very dangerous query loophole that could easily burn beginners of Meteor.

stubailo:
An array is not a valid selector in Mongo, so we should fail fast in that case. Also, before this fix, if you passed an empty array as a selector, it would select everything just like `{}`.

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -328,6 +328,12 @@ Mongo.Collection._rewriteSelector = function (selector) {
   if (LocalCollection._selectorIsId(selector))
     selector = {_id: selector};
 
+  if (_.isArray(selector)) {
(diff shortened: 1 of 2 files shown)


stubailo:
@Slava, could you take a look when you have time?

@AChusuei, you wanted me to give you a heads up.

PR #4848, opened 2015-07-31
Remove `meteor-platform` from `test-packages` app, fix dependencies

stubailo:
It turns out that `test-packages` automatically included all of `meteor-platform`, meaning all packages were tested with the assumption that everything in Meteor platform was globally available on the client.

Now that we are going to split up meteor platform into smaller modules, this is not an acceptable way to test/build packages. I went through and made sure that every single package that uses something from `meteor-platform` actually directly depends on it.

This miiiiight break some package authors' tests, but really that is a good thing in this case because they will need to actually depend on things that they use.

### Spacebars <-> templating

Spacebars used to depend on templating, but since templating actually generates code that requires Spacebars, it made more sense to do it the other way around. This means that Spacebars can't depend on templating anymore, so I moved Template.dynamic into the templating package.

@dgreensp can you review this? Really it's just the Spacebars part that I'm not 100% sure what the right solution is.


dgreensp:
It's a shame to lose `dynamic.html` in source form.

Can we solve this with unordered dependencies (`api.use('mypackage', {unordered: true})`)?  It doesn't seem like it's that weird for spacebars and templating to have a mutual dependency.

OR, hey, can we move `dynamic.html` into the `templating` package?

stubailo:
_WHOA_ I don't know why I didn't think of that.

Let me see what happens.

stubailo:
@dgreensp wow, that worked great! Awesome idea.

packages/test-helpers/package.js
```diff
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
-  api.use(['underscore', 'tracker', 'ejson', 'tinytest', 'random']);
+  api.use(['underscore', 'tracker', 'ejson', 'tinytest', 'random', 'blaze']);
```
dgreensp: Only the `renderToDiv` helper depends on Blaze, so in theory we could let you use the other helpers without pulling in Blaze.  I guess pulling in Blaze doesn't do any harm, though.

dgreensp:
LGTM!

Issue #1608, opened 2013-11-14
Packages Enhancement: Predefined Groups


aldeed:
This is an enhancement suggestion concerning packages. Looking to see if MDG and others like this idea, or if someone has a better solution.

The idea is to have predefined package groups. Probably have the list of groups in a released JSON file, easy for contributors to do PRs. Example _package-groups.json_:

``` js
{
  "group1": ["package1", "package2", "package3"]
}
```

Then add `meteor` commands `add-group` and `remove-group`:

``` bash
$ meteor add-group group1
$ meteor remove-group group1
```

This would add or remove all the packages individually, as if you had typed `meteor add` for each one. The idea is that `standard-app-packages` could then be a package group instead of a package, which makes it easier to use subsets of packages.

Example 1: Use standard stack without mongo database

``` bash
$ meteor add-group standard
$ meteor remove-group mongo
```

Example 2: Remove standard stack from a new app but use all "accounts-" packages

``` bash
$ meteor remove-group standard
$ meteor add-group accounts 
```

This could be achieved by creating many more "grouper" pseudo-packages like the current `standard-app-packages`, but in my opinion it's more intuitive to think in terms of groups, and that keeps the package list smaller.

It would also be nifty to have `meteor` look for a second _package-groups.json_ file in the user's home directory, so that users could create their own groups for combinations they commonly use. Could be linked with a `meteor create-group pkg1 pkg2 pkg3 pkg4` command for convenience.

PR #4851, opened 2015-07-31
Split meteor-platform into smaller umbrella packages code and discussion

stubailo:
> This PR is currently a work in progress; the plan is listed here.

Right now, we have a large meta-package called `meteor-platform` which contains a lot of neat packages that people might want when building their app. Unfortunately, since everything is in one package, it’s hard to remove parts that you want to replace - you need to remove the platform meta-package and add in each package separately. In particular, if you want to avoid using Blaze for HTML templates, it’s very difficult.

I think there will be a lot of benefits from splitting up the `meteor-platform` package into smaller packages that contain units of functionality.

### Goal

* Make it possible to remove some packages from the initial Meteor app without having to add back in every package individually
* Make it easy to switch to a different view layer
* However, don't confuse people by listing every single package and make it clear what each one does
* Keep in mind that this doesn’t change the Meteor release process at all - `meteor-platform` was just a “convenience” package and didn’t do anything about pinning version numbers. All of that stuff will work the exact same way, but the output of `meteor list` will look a little different.

### Planned packages

After this project, here is the set of packages that will be included by default in a newly created Meteor app:

1. `meteor-base` is the set of packages that basically every single Meteor app will have. If you don’t have these packages, you are probably doing something that isn’t really supported, like building a command line tool or switching out the whole web server stack. It comes with the following packages:
    1. `meteor` - this includes stuff like `Meteor.isClient`, a default handler for `css` files, etc.
    2. `webapp` - this is responsible for handling actual HTTP connections, Websockets, and serving files
    3. `underscore` - almost all of Meteor is built on top of underscore, so it makes sense to let people assume that most or all Meteor apps right now will have this included
    4. `autoupdate` - refreshing the client is a core part of the Meteor development experience, and it’s integrated into several layers of the stack
    5. `ddp` - lots of core parts of Meteor assume that DDP can be used to communicate between client and server
2. `standard-minifiers` minifies your JS and CSS code in production
3. `ecmascript` allows you to write your app using new ES2015 JavaScript features
4. `es5-shim` polyfills some newer APIs in old and non-compliant browsers, in particular IE8
5. `mobile-experience` is a set of cordova-specific packages that set some good defaults when building for mobile. These packages only activate when you are building a native Android or iOS app.
    1. `fastclick` - avoid the 300ms touch delay
    2. `mobile-status-bar` - avoid the status bar information covering up your app content
    3. `launch-screen` - cover the app with a launch image so that people don’t have to see things loading
6. `mongo` is the package that enables Meteor to connect to MongoDB on the server and watch queries in real-time. It also includes Minimongo for the client so that you can publish Mongo documents over DDP. This package will be removable in case you want to use one of the community-supported drivers for alternate databases, and for the desirable future where Meteor supports other databases officially.
7. `blaze-html-templates` compiles your `html` files with Spacebars and includes the Blaze runtime on the client so that the templates can run. If you remove this, you might want to include a different view layer like `react`, or `angular`, and use a package for rendering the starter HTML like `static-html` (also coming out in Meteor 1.2)
8. `tracker` the package that powers a lot of Meteor’s reactive APIs on the client. Including it in the app allows you to use `Tracker.autorun` directly.
9. `session` a simple global reactive dictionary for the client.
10. `jquery` a convenient utility library for the client.
11. `insecure` a prototyping package that lets you make any database modifications from the client.
12. `autopublish` a prototyping package that lets you access the whole database (except sensitive user data) from the client.

### Other changes

1. I’ll try to modify the `meteor` command line tool to avoid running MongoDB if you don’t have the `mongo` package added.
2. `meteor create` will print a helpful message pointing you to the tutorial, documentation, and other materials
3. The `autoupdate` package will be renamed to `hot-code-push` to better reflect what it actually does.
4. `meteor-platform` will no longer be a part of future Meteor releases. Apps upgraded to Meteor 1.2 will be automatically updated to use the packages listed above instead of `meteor-platform`. (Along with a set of packages like `EJSON` and `Random` that used to be in meteor platform but probably shouldn’t have been)

----

Working document (feel free to read and leave comments about the reasoning): https://quip.com/aVGFAGa22zPu

----

TODO:

- [x] Write up plan and post to the forums to get feedback
- [x] Write READMEs for the new umbrella packages
- [x] Remove `meteor-platform` from release
- [x] Add an upgrader to replace `meteor-platform` with components
- [x] Remove `logging` dependency from Blaze
- [x] Fix constraint bug in meteor-platform upgrader

In separate PRs - these are separate chunks of work:
- [x] ~~Figure out if it is hard to not run MongoDB if we don't have the `mongo` package~~ (it is hard)
- [x] Make `meteor create` print a helpful message that takes you to the docs/tutorial: https://github.com/meteor/meteor/pull/4871
- [ ] Rename autoupdate to hot-code-push
- [ ] Do backcompat via dependencies rather than umbrella packages


Primigenus:
This is a great initiative that I've missed for a while. Two others that I've been considering: a group of packages under "backcompat" that define which behaviour is merely included to support old APIs (eg. 0.9.0 or older browsers) and "prototyping" (or something similarly named) that includes autopublish and insecure. It looks like you already called these out in comments in the packages file, but does it make sense to make them into actual packages?

stubailo:
@Primigenus thanks for posting some feedback on this PR! Really appreciate it - I'll put out a call for wider feedback after making some changes prompted by discussions with the team. About your suggestions:

### Backcompat package

Problems related to backcompat:

1. Packages that look for a certain other package being included. For example, `if (Package.deps) {` requires there to be an actual package called `deps`, even if the `tracker` package exports the `Deps` symbol.
2. Code to maintain API compatibility with old versions of a package. So if we change the API of `Tracker`, tracker will contain some old code that enables the old API so that packages don't needlessly break.

Am I missing some others?

In this case, I think (1) could actually be solved by having `tracker` have a hard dependency on `deps` until we bump the major version of `tracker`? (2) should just be code inside the `tracker` package, again, until the major version is bumped.

We need a better strategy for deprecating and renaming packages, I think.

But IMO I don't really see the need for a standalone backcompat package because in most cases it's an application level concern - which browsers do you want to support? which APIs are you using? which packages are you using, and what versions of Meteor were they written for? If there were a backcompat package, we would need a new one for every version of Meteor, and possibly different ones for different parts of the platform. What do you think?

### Prototyping

I think we need to figure out how to do a better job of either (1) not adding insecure and autopublish by default or (2) making it more obvious that you really really need to remove these packages if you are running in production. I don't know that a package name will fix that, especially since there are already many helpful tutorials telling you that you need to remove `insecure` specifically so if we change the name to `prototyping` then people might get even more confused.

Honestly, I would be in favor of an approach where we literally warn the user every single time they start the app - "hey, we added this package by default so that you can prototype fast, but make sure you don't run it in production!" or something similar. Maybe a warning in meteor build or meteor deploy?

### Comments in the packages file

Unfortunately, we've been billing the `packages` file as something that is meteor-internal for most users. I suspect most people use `meteor list` instead of looking at the file, so the comments won't help them at all. Maybe we can make `meteor list` show comments from that file somehow?

Primigenus:
@stubailo I agree with all of your conclusions. It would be nice if we could somehow annotate some of these packages to imply the grouping (backcompat, prototyping) since it looks like changing the names would cause the problems you described. I guess you tried that with comments, so it sounds like looking into a way to surface those comments would be a great next step. Thanks!

msavin:
(Y) I've been looking at Meteor as a build tool more than anything, and that has been one of the buy-ins for me. Right now, people are swapping the client side stuff, and soon, they'll be swapping the databases, so this seems like the right move. I'd also love to easily be able to drop Session :)

AndreasGalster:
So how would you go about not listing all the new smaller packages? This would certainly be a pain in the ass. I would recommend having the possibility to either add the complete package as it is right now or let people choose to just add certain smaller packages.

timbrandin:
Are the static-html package going to be included in the default packages too, and what will it actually do behind the scenes? Anyways this seems to be the right way to go! Great work! :+1:

stubailo:
@timbrandin unfortunately, it's currently not possible to have two handlers for `.html` in your app, so `static-html` can't be used together with `blaze-html-templating`. So if you wanted to use React you would probably remove `blaze-html-templating`, and add back either `static-html`, or some fancy html to React templating package instead, which would take over handling the `html` file extension.

@AndreasGalster I think it is only a pain in the ass in a super new app. Any "real" app will probably include dozens of packages, so these won't be a big deal. IMO it's much more important to show people what functionality they can replace rather than having a smaller number of packages listed. I think Rails does the right thing here; the Gemfile in a new app has a few packages but each one is nicely commented and explained.

IMO the worst thing possible is giving people a totally opaque `meteor-platform` which makes it look like you can never change any packages, and gives you a bunch of stuff you don't need.

timbrandin:
Ah I see, so in theory we could build static sites by removing blaze-html-templating and adding in static-html. I like that possibility!

timbrandin:
How about creating a new flag "optional" in the Package api so that package authors could write packages that support all of the rendering frameworks but parts of the app is only included if one has added this and that.. that would also make it possible to disassemble the meteor-platform package.

Also it would be nice if one could see in `meteor list` which package is included by imply, that would make it easier finding the culprit who added iron:router or whatever package that doesn't fit with your current needs.

mariorodriguespt:
@stubailo I think this is a fantastic idea. 

I've wondered several months ago how could I build a web widget with some meteor packages without having to load the whole thing. This could be very helpful if you want to serve this widget to a non meteor website. An iframe isn't a good idea today for this problem, Hackers News uses 100 mb of RAM and an empty meteor app uses 200mb on my machine.

On the widget scenario I would only need the minimongo package to accomplish it, a guide on this would help a lot. I would love to use some packages in non client side meteor projects, although some modifications are necessary at client side in the core packages code. The use case I had was a client that wanted to display realtime statistics in a widget, I can all do it by self or I could use minimongo and save a lot of time. This would be, in my opinion, a step towards a better integrations with other platforms.

I don't know if this is a side intended intension of this pull request.

rolfnl:
Will you be renaming a package like <code>ecmascript</code> again when there's a new version coming and want to support that too?

The <code>auto publish</code> & <code>insecure</code> packages are actually a great feature for new users and quick prototyping, one of the strengths of the platform imho. It's all over the place already to remove them for production (like you already mentioned). I'd opt for a message anytime you run <code>meteor</code> or when you deploy.
Perhaps there are some extra flags possibilities when creating a project, e.g. to remove default packages <code>meteor create foo --remove=insecure,autopublish</code> .. I don't know.. I don't mind creating a new project and then deleting them right away.

Any chance to replace <code>underscore</code> with <code>lodash</code> throughout meteor? They will probably be merged somehow in the future anyway (I hope and get that sense reading through https://github.com/jashkenas/underscore/issues/2182).

In any case, this is a good move.

stubailo:
@timbrandin about "optional" dependencies - you can already have a "weak" dependency, which means that if the package is added via some other source then it will definitely load before yours. You can check by looking at `Package["package:name"]`. So you could activate different features of your package based on the presence of other packages. Unfortunately, it is not possible to do this at build time right now (I think, I could be wrong though).

@mariorodriguespt unfortunately, I don't think this will help you that much in building your widget - it doesn't enable anything new that you couldn't do before. A good option for you might be to use something like Asteroid, which is a more lightweight client for a Meteor server: https://github.com/mondora/asteroid However, we are looking at making the kind of thing you are doing easier in the future.

@rolfnl I don't know how the release management for `ecmascript` will work, but I imagine that when we add new ES2015/2016/etc features we will bump the version of the package rather than making a new package with a new name. Either way, that's out of the scope of this PR. Replacing underscore is also not in the scope of this project and there's a long lived discussion at #1009.

stubailo:
@timbrandin about a new set of features for `meteor list` - that should probably be a separate feature request, I encourage you to post it separately.

AndreasGalster:
@stubailo Good points. I've just realized a big issue with listing packages comes from well, the long lists. Why not visually group packages based on categories, such as utilities, upload/storage, etc. to make it easier scanning through a large list of packages?

stubailo:
@AndreasGalster @timbrandin I think you both have some good ideas about ways to improve `meteor list` - maybe we should start a forum thread about it or something?

stubailo:
@dgreensp I think this is ready for review! I've split some of the work into separate PRs, going to do those in parallel.

dgreensp:
Even if there isn't one package called "backcompat", maybe the desired thing is to be able to remove a package and have top-level symbols like `UI` and `Handlebars` go away (and maybe other symbols or methods) that exist purely for back-compat?

packages/meteor-base/package.js
```diff
@@ -0,0 +1,36 @@
...
+    // Most Meteor core packages depend on Underscore right now
+    'underscore',
+
+    // The protocol and client/server libraries that Meteor uses to send data
+    'ddp',
+    'livedata', // XXX COMPAT WITH PACKAGES BUILT FOR 0.9.0.
+
+    // A small client-side reactive programming library that underpins most of
+    // Meteor frontend code
+    'tracker',
```
dgreensp: this shouldn't be here

Primigenus:
@dgreensp The reason I had wanted something like backcompat was because I'm pretty obsessive about download size of my Meteor apps, and if I know up front that I don't need the backward compatibility then I would also like to not have to ship that code. If there's an easy way to just excise it, that would be a lot more convenient than having to manually rearrange the meteor-platform package.

ccorcos:
Yay! +:100: ! I like this PR!

1) I like the idea of meta-packages and I see nothing wrong with them. `prototyping`, `backcompat` are great ideas. I also wouldn't mind putting the rest in `meteor-platform` just so `meteor list` is nice and clean.

2) If you're making it possible to take Mongo out of the build tool, then I think this would be a good time to build a simple API for starting and stopping databases in the build tool (allowing to use environment variables and Meteor.settings to set the database url, etc.).

3) One last suggestion -- create an API for `accounts-base` for client / server storage. This way we can store client data however we want (cookies on a browser, a file in an electron app, or a file in a React Native app), and we can save/fetch records from whatever database we want on the server.

timbrandin:
Either a Github issue or a forum thread works, and I think starting a forum
thread is destined to get more attention if you started it. :)

2015-08-04 3:48 GMT+02:00 Sashko Stubailo <notifications@github.com>:

> @dgreensp <https://github.com/dgreensp> I think this is ready for review!
> I've split some of the work into separate PRs, going to do those in
> parallel.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/meteor/meteor/pull/4851#issuecomment-127452478>.
>



-- 
Cheers,
// Tim Brandin @timbrandin <https://twitter.com/#!/timbrandin>
http://timbrandin.com
http://studiointeract.com

ssteinerx:
Something like the jQueryUI selector/builder would be very cool and could be a locally run meteor app to select a matrix of packages with rules for selecting required minor support packages for major packages etc.:

https://jqueryui.com/download/

packages/blaze-html-templates/package.js
```diff
@@ -0,0 +1,21 @@
...
+  // To avoid submitting documentation, set this field to null.
+  documentation: 'README.md'
+});
+
+Package.onUse(function(api) {
+  api.imply([
+    // A library for reactive user interfaces
+    'blaze',
+    'ui', // XXX COMPAT WITH PACKAGES BUILT FOR 0.9.0.
+    'spacebars', // XXX COMPAT WITH PACKAGES BUILT FOR 0.9.0
```
dgreensp: We should clarify all these `XXX COMPAT` messages, because these aren't packages full of back-compat code, and actually some of the reasons for including them are subtle.

For `ui`, it sounds like it's so that `Package.ui` exists.  There's literally no other effect of implying it here, which is quite subtle indeed.  For `spacebars`, it's to export the `Spacebars` symbol.  We should also comment that `ui`, `livedata`, and `deps` are empty packages.

dgreensp:
@Primigenus I see.  I also love small bundles, and I look forward to reducing the size of the Meteor bundle, because I think there's some low-hanging fruit.  Unfortunately, it hasn't been a company priority so far, but the work we're doing now is a good start.

There's really very little code that exists purely for pre-0.9.0 back-compat -- mainly just some name aliases, so a line here or there.  Are you thinking of some code that I'm not thinking of, or are you going off of the comments in the PR?  The `ui`, `livedata`, and `deps` packages are all empty, and `spacebars` is included anyway.  The back-compat going on here is at the level of the package system, not code.

aldeed:
Just wanted to point out that I suggested this a loooonnnnggg time ago. :) So thanks.

https://github.com/meteor/meteor/issues/1608

stubailo:
@aldeed unfortunately it's not as good as what you suggested :P but it accomplishes some similar goals.

@ccorcos if you have some feature requests or something like that, you should have those discussions on the relevant conversation thread. I don't think there is any possible way that creating a storage abstraction for accounts could be part of this PR.

Primigenus:
@dgreensp No, I was thinking of the same stuff - like I said, it's an obsession. An empty file with 7 or 8 lines of Javascript declaring an alias split across a number of packages adds up. It definitely falls in the "low hanging fruit" area of bundle size optimisations we could do for me.

dgreensp:
@Primigenus Oh man, there is low-hanging fruit that is orders of magnitude larger than 7 or 8 lines.  We've known it for a while, we just haven't moved on it.  Just now I removed the `json` and `logging` packages from the required packages (since all browsers have JSON anyway).  The next release will make packages like `session` removable, and longer term, why not `jquery`; why not `underscore`.  To say nothing of not forcing you to ship all your templates down in one bundle, but loading them incrementally.

Primigenus:
@dgreensp Haha, if you consider that to be low hanging fruit too, great!

sebakerckhof:
Related request: https://github.com/meteor/meteor/issues/4897

samcorcos:
@stubailo Yes! The more modular and generalized the core packages, the better. This is definitely a step in the right direction.

msavin:
On this point, it would be great if we can get more data on packages with-in the app, as well as trigger package add/remove - it can lead to some interesting tools :smile:

msavin:
I think it would be great if we can get better package management API too - to get more data on packages in the app (such as versions), as well as easily add/remove/update packages. It could lead to some interesting tools :smile:

juliomac:
I am running my app on Meteor 1.3.2.4 now and I am trying to reduce size and speed up load on client.

I have this up being updated since Meteor version 0.9 or so and I wonder if it is safe to remove ```meteor-platform```?  That is, is ```base``` duplicating packages as it gets in?

Besides, I've seen lots of packages for ```web.cordova``, for instance, that I don't use. Would removing it save me some more load time?

stubailo:
No, it wouldn't! Meteor doesn't load duplicate packages. But please open a new issue rather than commenting on a closed PR if you have further questions or post on the forum!

Slava:
- [ ] asdf

PR #4865, opened 2015-08-03
Collection.exists: New method in Mongo.Collection API

yanisIk:
I decided to add this new method because a lot of people use ``` !Collection.findOne() ``` instead of ``` !Collection.find().count() ``` to check if a document exists.

The latter one is way more performant but less elegant to write. That's why I think it is appropriate to have a method like this one.
Adding it to the Meteor Mongo.Collection API and in the official documentation will increase by a lot the performance of a lot of Meteor applications.

More information about that here (includes benchmark):
https://blog.serverdensity.com/checking-if-a-document-exists-mongodb-slow-findone-vs-find/

### DIFF
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -231,6 +231,24 @@ LocalCollection.Cursor.prototype.count = function () {
   return self._getRawObjects({ordered: true}).length;
 };
 
+/**
+ * @summary Checks if document(s) corresponding to the cursor exist.
+ * @memberOf Mongo.Cursor
+ * @method  exists
+ * @instance
+ * @locus Anywhere
+ * @returns {Boolean}
+ */
+LocalCollection.Cursor.prototype.exists = function () {
+  var self = this;
+
+  if (self.reactive)
+    self._depend({added: true, removed: true},
+                 true /* allow the observe to be unordered */);
+
+  return self._getRawObjects({ordered: true}).length > 0;
+};
+
 LocalCollection.Cursor.prototype._publishCursor = function (sub) {
   var self = this;
   if (! self.collection.name)
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -295,6 +295,32 @@ _.extend(Mongo.Collection.prototype, {
     var argArray = _.toArray(arguments);
     return self._collection.findOne(self._getFindSelector(argArray),
                                     self._getFindOptions(argArray));
+  },
+  
+  /**
+   * @summary Check if the documents in a collection that match the selector exists.
+   * @locus Anywhere
+   * @method exists
+   * @memberOf Mongo.Collection
+   * @instance
+   * @param {MongoSelector} [selector] A query describing the documents to find
+   * @param {Object} [options]
+   * @param {MongoSortSpecifier} options.sort Sort order (default: natural order)
+   * @param {Number} options.skip Number of results to skip at the beginning
+   * @param {Number} options.limit Maximum number of results to return
+   * @param {MongoFieldSpecifier} options.fields Dictionary of fields to return or exclude.
+   * @param {Boolean} options.reactive (Client only) Default `true`; pass `false` to disable reactivity
+   * @param {Function} options.transform Overrides `transform` on the  [`Collection`](#collections) for this cursor.  Pass `null` to disable transformation.
+   * @returns {Boolean}
+   */
+  exists: function (/* selector, options */) {
+    // Collection.exists() (always return true if Collection isn't empty) behaves differently
+    // from Collection.exists(undefined) (always return false).  so be
+    // careful about the length of arguments.
+    var self = this;
+    var argArray = _.toArray(arguments);
+    return self._collection.find(self._getFindSelector(argArray),
+                                 self._getFindOptions(argArray)).count() ? true : false;
   }
 
 });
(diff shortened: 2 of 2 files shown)


yanisIk:
Agreement signed.

sebakerckhof:
Just wondering: would adding limit:1 to the options increase performance of this function?

kookoskar:
@sebakerckhof I believe it would, but then you would have to also pass `true` to `count()` (`count(true)`) as per [MongoDB docs](http://docs.mongodb.org/master/reference/method/cursor.count/). Otherwise you still get the complete count.

I think this would be a welcomed change.

stubailo:
Can we get a discussion going somewhere, perhaps the Meteor forums, discussing how people feel about this feature?

Out of the several different methods we could add to the collection API, why this one? I'm not saying it's the wrong one, but it would be good to see some more discussion first.

kookoskar:
@stubailo For me, Meteor is something that is supposed to take all the clutter away from developing, which it does so pretty well. A big part of this is not having to know the intricate details of everything - you can use the packages for most things, where someone who knows the details has taken care of them.

Why should there be some kind of "aha! got you! you shouldn't be using `findOne()` to check whether something exists!" trap? Instead, provide a clear-cut method for that. I also generally see `coll.exists(selector);` being much more "Meteor-like" and developer-friendly than `coll.find(selector, {limit: 1}).count(true) > 0;`, which is just.. ugh.

stubailo:
@kookostar you brought up another thing, which is that most things can have packages - I think this is actually a case where this function could easily be part of an add-on package (although I'll admit it would be hard to discover!)

Again, I'm not saying this is a bad feature to have, just looking for some more ideas about it.

yanisIk:
@stubailo Personnaly, I think that this doesn't have to be in a package since it is not an additional feature that people need, but just a better way to use MongoDB that can be implemented directly inside the Collection API. 
It is a good practice that everybody should use when trying to check if a document exists so it should be preferably implemented inside the Collection API or in the MongoDB api itself.

As a basic user, I will never go to Atmosphere and ask myself if there is a package that can find document for me, I will just use Collection.findOne() and check if the object is null. Almost all the projects are using this and they don't know that there's a better way to do it (and they probably don't need it), that's why it should be implemented in the Collection API.

This is my opinion of course. I just think that packages should be used for additional features that are not easy to add to the core and more complex than that.

mitar:
We use this feature extensively in PeerDB:
 * [local collection implementation](https://github.com/peerlibrary/meteor-peerdb/blob/master/lib.coffee#L78)
 * [server side implementation](https://github.com/peerlibrary/meteor-peerdb/blob/master/server.coffee#L24)

It really makes some things cleaner to do, and code to read.

I am in support of adding this in. But I would suggest that the author of this pull request checks our implementation. Because the whole point of `exists` for me is that it should work faster than `count()` and current implementation in this pull request makes it only a sugar, and not really improve on performance.

rgoomar:
I think this would be a good feature to add. It does allow for cleaner code.

rclai:
Yes, please add options `{ limit: 1, fields: { _id: 1 }, transform: null }`.

yanisIk:
@rclai I added the options you suggested.
@mitar I read your code and didn't really understand what's the difference, I saw that you use almost the same options that rclai suggested, is it enough or can we improve that ?

mitar:
So there are two things here. `exist` on the collection and `exists` on the cursor. In PeerDB we have both. Let's discuss both. :-)

So for collection `exists`, you use `count()` with limit. I think it is much simpler to [simply use `findOne()`](https://github.com/peerlibrary/meteor-peerdb/blob/master/lib.coffee#L454). It has exactly those properties you want, trying to get any document, if there exist. No need to ask for precise count, just to know if there is any. Also, count with limit might or might not do correct thing because of this bug: #1201

Then there is cursor based `exists`. For those we have two implementations, [for Minimongo](https://github.com/peerlibrary/meteor-peerdb/blob/master/lib.coffee#L79), and [for MongoDB](https://github.com/peerlibrary/meteor-peerdb/blob/master/server.coffee#L24). Those are also important because it is common to pass a cursor to the template and you can then do `if cursor.exists then render content else render something else`.

rclai:
#1201 is such an old bug!

mitar:
Exactly. So you have many installations of Meteor with old and new behavior. So it is better to not try to use `limit` with `count`. :-)

yanisIk:
Oh I see now, thanks. 
So findOne with {transform: null} does the same thing as find with {limit:1} (find returns a cursor so it doesn't transform anything by default right ?)

But I don't see why we should do so many changes for the same thing. The modification on the Mongo.Collection object should be enough, solves the ''problem'' and is accessible everywhere (client and server).

mitar:
Because also `count()` exists in Meteor on the cursor. So if you want to use `exists()` as an optimized/nicer semantic version of `count()` we need a version on the cursor as well. It is really rare to make a query just to check if something exists, and then go and make another query to get data.

kookoskar:
True, `cursor.exists()` should be defined as well. I'm even inclined to think that there should be no `collection.exists()`; only `cursor.exists()`, so the API stays consistent.

yanisIk:
@mitar I just added the ''exists'' function on the cursor too. Let me know if this is correct.

By the way, do I have to add tests for those two methods (sorry, it's the first time that I contribute to a big project like this one).

stubailo:
Yes, these will need to have complete tests to be accepted.

tmeasday:

Issue #4862, opened 2015-08-02
 Meteor.userId() throws an error in validateLoginAttempt/onLogin/onLoginFailure callbacks


brettle:
`Meteor.userId()` is throwing an error when called within `Accounts.validateLoginAttempt`/`Accounts.onLogin`/`Accounts.onLoginFailure` callbacks. This is preventing me from determining which (if any) already logged in user made the login attempt.

Here's a reproduction recipe using [this github repo](https://github.com/brettle/Meteor.userId-throws-error-in-validateLoginAttempt-callback):

```sh
[brettle@localhost git]$ git clone git@github.com:brettle/Meteor.userId-throws-error-in-validateLoginAttempt-callback.git login-callback-bug
Cloning into 'login-callback-bug'...
remote: Counting objects: 13, done.
remote: Compressing objects: 100% (9/9), done.
remote: Total 13 (delta 0), reused 13 (delta 0), pack-reused 0
Receiving objects: 100% (13/13), done.
Checking connectivity... done.
[brettle@localhost git]$ cd login-callback-bug/
[brettle@localhost login-callback-bug]$ meteor
[[[[[ ~/git/login-callback-bug ]]]]]          

=> Started proxy.                             
=> Started MongoDB.                           
=> Started your app.                          

=> App running at: http://localhost:3000/
```

Visit http://localhost:3000/ and click the "Login with correct password" button. You'll get this on the server console:
```sh
I20150801-21:56:11.520(-7)? In validateLoginAttempt callback:
I20150801-21:56:11.521(-7)?   Meteor.userId() throws: Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.
I20150801-21:56:11.521(-7)?   this.userId: undefined
I20150801-21:56:11.522(-7)? In onLogin callback:
I20150801-21:56:11.522(-7)?   Meteor.userId() throws: Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.
I20150801-21:56:11.522(-7)?   this.userId: undefined
```
Click the "Login with incorrect password" button and you'll get this on the server console:

```sh
I20150801-21:56:15.091(-7)? In validateLoginAttempt callback:
I20150801-21:56:15.091(-7)?   Meteor.userId() throws: Error: Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.
I20150801-21:56:15.091(-7)?   this.userId: undefined
I20150801-21:56:15.091(-7)? In onLoginFailure callback:
[...]

PR #4867, opened 2015-08-03
Fix issue #4862. Don't wrap login callbacks with bindEnvironment.

brettle:
Changes to packages/callback-hook:
- Add bindEnvironment option to Hook constructor (defaults to true).
- Add internal helper function dontBindEnvironment() which does all the
  error handling stuff like Meteor.bindEnvironment() but doesn't change
  the environment.
- Use dontBindEnvironment() instead of Meteor.bindEnvironment() when
  bindEnvironment option is false.
- Add tests.

Changes to packages/accounts-base:
- Create hooks with { bindEnvironment: false }
- Add test for whether Meteor.userId() is available in login callbacks.

### DIFF
--- a/packages/accounts-base/accounts_common.js
+++ b/packages/accounts-base/accounts_common.js
@@ -22,10 +22,12 @@ AccountsCommon = function _AccountsCommon(options) {
 
   // Callback exceptions are printed with Meteor._debug and ignored.
   this._onLoginHook = new Hook({
+    bindEnvironment: false,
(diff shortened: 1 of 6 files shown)


stubailo:
@brettle makes sense to me. The context of the login callback should be the login attempt, not the code where the callback was attached.

stubailo:
Merged here: 587e02a692da86eaed03a8ff7a2c5010cda17ba3

stubailo:
@brettle thanks for all of the tests!

PR #4871, opened 2015-08-04
Print link to meteor.com/learn when a new app is created

stubailo:
I'm glad we have that awesome page to link to.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -718,6 +718,13 @@ main.registerCommand({
   Console.info(
     Console.command("meteor"), Console.options({ indent: 2 }));
 
+  Console.info("");
+  Console.info("If this is your first time using Meteor, try the tutorial:");
+  Console.info(
+    Console.url("http://www.meteor.com/tutorials"),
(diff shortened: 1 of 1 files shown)


Primigenus:
It would be nice if Meteor would likely only print this if it thinks you may be new to Meteor. Is there a way to guess that?

stubailo:
@Primigenus I feel like this wouldn't affect experienced users too much - two lines of output could be pretty easy to ignore. Do you disagree?

Primigenus:
@stubailo I feel like from a UX point of view it's good to anticipate what situation the user is in, so if for example we're dealing with a more advanced user maybe we can tell them something in those two lines that they can use rather than skip. That's why I was wondering out loud about it. :)

stubailo:
@Primigenus what would be a more useful thing to print for advanced users?

stubailo:
I didn't hear anything new, so I am merging: 927745a5d5e6c121b9b9582d212502947d836fa8

Primigenus:
Oops. Sorry about that, I was going to think about it and get back to you and then I forgot. Apologies!

PR #4873, opened 2015-08-04
Creating error app to show compiler errors in browser

anubhav94:
Created an error app inside the tool code that loads once the main app crashes, showing the logs that are printed to the Terminal prompt. This app persists a DDP connection to the user and allows developers to never have to go back to the terminal as the file watching api will reload their main app once they save and show the main app if the error is fixed or show the error on the error app.

Need to do:
- [x] Remove port 8600 hard coding
- [x] Remove 'localhost' hard coding for error app host
- [ ] Add in reload server button
- [ ] Hide behind production flag so loaded only in debug only
- [x] Add in app reloading update to Errors app when the error is updating

@Slava can you give this a look?


tools/commands.js
```diff
@@ -392,6 +392,18 @@ function doRunCommand (options) {
     mobileServer = mobileServer + ":" + parsedMobileServer.port;
   }
 
+  let errorAppPath = files.pathJoin(files.convertToStandardPath(__dirname),
```
stubailo: Is there any reason this is `let` and not `const`? https://github.com/meteor/javascript#13.1

anubhav94: Great point - there is no reason and I will change it to const.

stubailo:
I would also change the name of the app to something simpler - having the words "reactive" and "debugger" might imply some features that it doesn't actually have. I feel like it could just be called "development-error-app" or something

anubhav94:
Good point and fixed.

Slava:
The new rebased PR is here: https://github.com/meteor/meteor/pull/4881

PR #4875, opened 2015-08-04
Support prodOnly in addition to debugOnly

dgreensp:
It's very useful to be able to include code only in production.  It's
useful for React integration, but presumably also for many apps.

* Add prodOnly boolean flag alongside debugOnly in Package.describe

* Packages the set prodOnly to true auto-depend on isobuild:prod-only
  (making them error in the old tool where prodOnly isn't supported)

* The `includeDebug` boolean build option is replaced by a string
  named buildMode, which can be 'development' or 'production', just
  like minifyMode.

Tested by self-test.


tools/isobuild/package-source.js
```diff
@@ -547,6 +551,8 @@ _.extend(PackageSource.prototype, {
        * @param {String} options.debugOnly A package with this flag set to true
        * will not be bundled into production builds. This is useful for packages
        * meant to be used in development only.
+       * @param {String} options.prodOnly A package with this flag set to true
```
Slava: why is its type a string?

stubailo: It's also string in debugOnly... both should be updated probably

dgreensp: weird, I'll fix

Slava:
From the superficial look at it, it looks good to me.

I would request adding a couple of tests just for completeness to verify my understanding:

- self test for backwards compatibility if it is possible to test (or at least a test that the magical dependency on `isobuild:prod-only` is added and the latest tool supports it in its features)
- a test on interactions between both `debugOnly` and `prodOnly` flags set in the same package. What should happen?

Also at least internal docs that specify how debugOnly/prodOnly affect the following:
- picked versions (solution from CS)
- load order
- linking (to my knowledge the debugOnly part of the package is not fully linked? or it doesn't have explicit imports?)

dgreensp:
@Slava, holy cow, you are right, it affects load order -- `debugOnly` can lead to different load orders in development and production.  This seems like something we should fix eventually, but maybe for this release we can just document it.

Picked versions are not affected.  `debugOnly` / `prodOnly` packages are simply omitted from linking.

Specifying `debugOnly` and `prodOnly` on the same package should be an error.  Good catch.

I'll put this info in a comment in `package-source.js`.  At least that will be discovered by someone grepping the code.  I'll add a couple more self-tests.

dgreensp:
Ok, I did the things I said, including writing another self-test and writing some notes on `debugOnly` and `prodOnly` in the code.

Can you think of an easy way to test that old Meteor chokes on a package that uses `prodOnly`?  Can I just create an app in `tools/tests/apps` with `.meteor/release` set to a published release like `1.1.0.3`?

Slava:
@dgreensp yeah, I can't think of any good way to test that otherwise.

dgreensp:
I guess what I meant was, is that possible?  It looks like I may have to create a "mock warehouse" or something...

Slava:
@dgreensp I don't know if it is possible.

dgreensp:
Ok.  Looks like it is not possible, or at least not easy.  I will not write that test for now.

dgreensp:
I fixed a bug in this, where the code the imports symbols from other packages is generated.  It's important that this code not generate code to import symbols from debugOnly or prodOnly packages.

I also had to modify the package server to allow prodOnly in package metadata (not in this repo).

The final diff: https://github.com/meteor/meteor/compare/1dadcd6581dae49d2ebba1ce1dc8bc384cf015a6...31a10f3b9553be0336c845148c66c6ec18c59a39

dgreensp:
@glasser ^ You might be interested in reviewing the final diff if you're back today.

PR #4879, opened 2015-08-04
If stdout is non-empty, assume that 'tasklist' and 'netstat' succeeded

ndarilek:
My Meteor app failed after a Windows 10 upgrade. A bit of digging revealed

See my indepth analysis on that issue for more specifics. The TLDR is that
even when I run the child_process invocation by hand in a node REPL it
looks like the command both succeeded and failed (I.e. stdout contains the
successful invocation but `error` is non-null.) This happens for both
mongod and netstat invocations.

I'm sorry, I can't test this because I suspect it requires exposure to
multiple Windows environments and I just have my tablet. My hope is that
MDG has better test infrastructure than I do and can put this through its
paces so it can land in a release quickly.

I also don't know if this is the correct solution. I do know that it
turned my failing Meteor app into one that at least mostly launches,
barring a non-OS-specific package error, and that this fix is letting me
get back to work. :) Perhaps a Node/Windows integration expert can look at
why this command appears to both succeed and fail, but I'm not that
person.

### DIFF
--- a/tools/runners/run-mongo.js
+++ b/tools/runners/run-mongo.js
@@ -76,7 +76,7 @@ if (process.platform === 'win32') {
 
     child_process.exec('tasklist /fi "IMAGENAME eq mongod.exe"',
       function (error, stdout, stderr) {
-        if (error) {
+        if (error && stdout.length == 0) {
           var additionalInfo = JSON.stringify(error);
           if (error.code === 'ENOENT') {
             additionalInfo = "tasklist wasn't found on your system, it usually can be found at C:\\Windows\\System32\\.";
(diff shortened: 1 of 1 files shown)


avital:
@ndarilek When I run:
```
child_process.exec('tasklist /fi "IMAGENAME eq mongod.exe"', function(err, stdout, stderr) {
console.log(stdout, stderr)
})
```

on my Windows 10 machine I don't see "The system cannot find the path specified.". I  wonder how our environments differ...

ndarilek:
I have a bit more information about this issue.

It seems that if you have a value in your cmd's autorun that is no longer present, as I did here, cmd will fail to run the autorun value, produce something on stderr and throw an error with child_process.exec. See this for more context:

http://superuser.com/questions/727316/error-in-command-line-the-system-cannot-find-the-path-specified

Removing the missing files (a leftover from a removed Chocolatey install) made this problem go away. I hadn't noticed them because I don't use cmd as my standard shell.

I still think this PR should land in some form, though. Meteor shouldn't completely fail if something in cmd's autorun is missing. Yes, it's ugly that Windows does this, but Windows itself is unfortunately ugly. :)

tmeasday:

PR #4881, opened 2015-08-05
Fancy page reloading on server error

anubhav94:
@Slava Here's the new PR.

Things left to do:

- [ ] Add in reload server button
- [x] Hide behind production flag so loaded only in debug only


tools/runners/run-all.js
```diff
@@ -163,6 +178,8 @@ class Runner {
 
     if (! self.stopped) {
       buildmessage.enterJob({ title: "starting your app" }, function () {
+        self.errorAppRunner.start();
+        runLog.log('Started error app.', {arrow: true});
```
Slava: I wonder if the users even need to see this line. I.e. they probably don't know what the error app is. As opposed to other things like "starting mongo" and "starting your app". Although, I would agree that the current "starting proxy" is not super helpful either.

tools/runners/run-all.js
```diff
@@ -163,6 +178,8 @@ class Runner {
 
     if (! self.stopped) {
       buildmessage.enterJob({ title: "starting your app" }, function () {
+        self.errorAppRunner.start();
```
Slava: any reason why we can't put it into the `otherRunners` array? What is the difference of not doing it?

tools/runners/run-all.js
anubhav94: I thought it'd be useful to know if the error app ever fails for whatever reason - instead of the regular main app they're trying to load.

tools/runners/run-all.js
anubhav94: I didn't know what the otherRunners array does - if it can fit there, then that works for me!

tools/development-error-app/.meteor/.id
```diff
@@ -0,0 +1,7 @@
+# This file contains a token that is unique to your project.
+# Check it into your repository along with the rest of this directory.
+# It can be used for purposes such as:
+#   - ensuring you don't accidentally deploy one app on top of another
+#   - providing package authors with aggregated statistics
+
+1exd2yi15tnfti17281wy
```
stubailo: Is it bad that this ID will be the same for every project?

tools/development-error-app/README.md
```diff
@@ -0,0 +1 @@
+# reactive-browser-debugger
```
stubailo: Would be good to have like a 1 paragraph description for this app.

tools/development-error-app/development-error-app.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+    refreshing: function () {
+      return Refreshing.find({_id: 'is-app-refreshing'}).fetch()[0].value;
+    }
+  });
+}
+
+Meteor.methods({
+  addErrorMessage: function (errMsg) {
+    if (this.isSimulation)
```
stubailo: You could just define these methods inside the server code to avoid them generating simulations.

tools/runners/run-all.js
```diff
@@ -110,6 +114,22 @@ class Runner {
       noRestartBanner: self.quiet,
     });
 
+    self.errorAppConfig = errorAppConfig;
+
+    if (errorAppConfig.runErrorApp) {
+      self.errorAppRunner = new AppRunner({
+        ...optionsForAppRunner,
```
stubailo: Do we know that we actually want to pass in the rest of the options?

tools/runners/run-proxy.js
```diff
@@ -190,13 +202,26 @@ _.extend(Proxy.prototype, {
...
+        if (self.mode === "hold")
+          break;
+
+        var c = self.websocketQueue.shift();
+        if (self.mode === "errorpage") {
+          self.proxy.ws(c.req, c.socket, c.head, {
+            target: 'http://' + self.proxyToErrorApp + ':' + self.proxyToErrorPort
+          });
+        }
+      }
```
stubailo: `} else {`

Slava:
It looks to me that in the current implementation, the exponential back-off of DDP connections still matters. Is it right?

Slava:
Also, if we are removing the whole idea of "try 3 times in 30 seconds", it is worth removing the loop, the setTimeout, breaks and continues.

tools/development-error-app/.meteor/.id
anubhav94: Not that I can think of - is there a place where this ID is used for a special reason where having it bundled with the same id will bebad?

Slava: I think you can safely delete it. I don't know if it even matters in our case.

anubhav94: Yup, I think it can be deleted and it gets regenerated if it doesn't exist so we should be all set.

anubhav94:
@Slava I think you're right the exponential backoff should matter but I don't think there is a time where it is growing very large. In both cases, we now disconnect the clients and thus they have to reconnect themselves.

Slava:
@anubhav94 I didn't get it, what is 7 seconds for and how are you going to increase it?

anubhav94:
It currently takes ~7 seconds from the time I manually break the app and the error app shows up [start time], fix the error immediately once it loads (looking at console.log for the current time), and then the main app reappears [end time]. I'm using console.log in Meteor.methods which are called upon Meteor.startup(). Changing the exponential backoff doesn't seem to do much because once we remove the reloads, the exponential backoff doesn't really get increased much since it gets another DDP connection to the error-app as soon as the proxy changes mode.

anubhav94:
@Slava Can you give this another look? We have testing now as well for the app.

Slava:
I have looked at it and the high-level parts look good to me. I didn't go into too many details on the last couple of commits.

Assigning this to Sashko since I am not going to work on this project any time soon.

stubailo:
I think this package handles many of the same concerns in a simpler way: https://github.com/stubailo/meteor-dev-error-overlay

PR #4907, opened 2015-08-07
Package-scoped variables, new instruction api.pckgscope(), consequent update of the coffeescript package

TC827:
At this moment it is not possible to write a meteor package in coffeescript and strict mode. See [this stackoverflow question](http://stackoverflow.com/questions/31759640).

This pull request contains the following changes:
* new instruction **api.pckgscope()**  to be used in *package.js* to declare variables with package
 scope. This instruction is similar to **api.export()**  with the difference that in this context there is no
 export.
* the consequent updates to the **isobuild**
* the consequent updates to the **coffeescript package**. The entire problem with share is now
 obsolete
* update to the documentation of the *coffeescript package*. 
* small update of the packaging documentation

### DIFF
--- a/docs/client/full-api/api/packagejs.md
+++ b/docs/client/full-api/api/packagejs.md
@@ -35,6 +35,8 @@ rest of this section will explain the specific API commands in greater detail.
       api.use('iron:router@1.0.0');
       // Give users of this package access to the Templating package.
       api.imply('templating')
+      // Make the scope of EmailIntern internal to this package.
+      api.pckgscope('EmailIntern', 'server');
       // Export the object 'Email' to packages or apps that use this package.
       api.export('Email', 'server');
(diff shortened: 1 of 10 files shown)


TC827:
Test cases: [github:TC827/meteor_pr4907](https://github.com/TC827/meteor_pr4907)

avital:
I think there's a much simpler fix to your issue by modifying `packages/coffeescript/plugin/compile-coffeescript.js` to place the declaration of `__coffeescriptShare` before "use strict" (which is already searched for).

In the longer term, we're planning to move towards ES6 modules for variable scoping instead of our in-house "linker" solution.

stubailo:
Yeah, I don't think we are going to add this API to `package.js`, especially since Avi's fix above should work. I'd look at a PR with that strategy.

PR #4233, opened 2015-04-17
Reusable Accounts constructors

benjamn:
This pull requests reduces the assumption of a singleton `Accounts` namespace in the `accounts-base` core package, making it much easier to manage multiple accounts connections without duplicating code.

At a high level, this refactoring was all about replacing the ad-hoc `Accounts` object with instances of the new `AccountsClient` and `AccountsServer` classes, which inherit from `AccountsCommon` so that methods can be shared (yay isomorphism!).

As an example of a luxury that had to be given up in order to support multiple accounts instances, initialization code that used to run at the top level (like the auto-login code) is now called from the constructor functions.

I think it's now much clearer which methods and properties are available on the client, on the server, or in both places, since each environment has a distinct `.prototype` where available methods are defined.

PR #4930, opened 2015-08-11
Fix Accounts apiAutoBoxes in docs

benjamn:
A number of `autoApiBox`es in the "Accounts" [section](http://docs.meteor.com/#/full/accounts_api) of the docs have been broken since https://github.com/meteor/meteor/pull/4233 was merged. This pull request fixes that documentation.

Although it is tempting to move all `Accounts{Client,Server,Common}` methods into their respective `class` bodies, I have limited this pull request to just those methods that appear in the documentation, in order to make the changes easier to review, while also verifying that JSDoc can handle ECMAScript 2015 syntax. Remember that the style of adding methods manually to `Accounts{Client,Server}.prototype` is still used in other packages like `accounts-password`, and is thus the mixture of method syntax and `.prototype` augmentation is likely to persist for the foreseeable future.


docs/client/full-api/nameToId.js
```diff
@@ -41,18 +40,29 @@ nameToId = {
...
+  "AccountsServer#validateNewUser": "accounts_validatenewuser",
   "Accounts.setPassword": "accounts_setpassword",
   "Accounts.sendEnrollmentEmail": "accounts_sendenrollmentemail",
-  "Accounts.onCreateUser": "accounts_oncreateuser",
+  "AccountsServer#onCreateUser": "accounts_oncreateuser",
   "Accounts.sendResetPasswordEmail": "accounts_sendresetpasswordemail",
   "Accounts.resetPassword": "accounts_resetpassword",
   "Accounts.sendVerificationEmail": "accounts_sendverificationemail",
   "Accounts.emailTemplates": "accounts_emailtemplates",
+  "AccountsClient#onResetPasswordLink": "Accounts-onResetPasswordLink",
```
stubailo: You shouldn't need these entries. they are mostly for backcompat.

stubailo: I'm sorry this isn't more documented... (is that ironic?)

stubailo:
So there’s a real problem right now that there is no way to separate fancy power user APIs from the general-use APIs that cover the most common use case. I think we should at least separate them organizationally in the docs.

My preference would be to have the docs mainly cover the old API - Meteor.user(), Meteor.userId(), and then have a separate section that talks about how to instantiate a new instance of accounts, instead of interspersing the two.

We should also make sure the basic docs still work. Since the doc for `Meteor.userId()` is reused in both places, it definitely shouldn't link to something else.

### Proposed changes

1. Move the canonical documentation back to `Meteor.x` and `Accounts.x`
2. Move all of the docs for `AccountsCommon`, `AccountsClient`, and `AccountsServer` to the bottom of the section, and phrase it as a thing for power users that need to connect to two different accounts servers. The docs for these methods can link to `Meteor.user`, etc by saying "works like like Meteor.userId does for the default connection" or something.
3. Lowercase the instance names in the docs to `accountsClient` to make it clear that you first need to create an instance of `AccountsClient` to call this method.

packages/accounts-base/url_client.js
```diff
@@ -86,21 +86,7 @@ AccountsTest = {
...
- * @param  {Function} callback The function to call. It is given two arguments:
- *
- * 1. `token`: A password reset token that can be passed to
- * [`Accounts.resetPassword`](#accounts_resetpassword).
- * 2. `done`: A function to call when the password reset UI flow is complete. The normal
- * login process is suspended until this function is called, so that the
- * password for user A can be reset even if user B was logged in.
- * @locus Client
- */
+// Documentation for this method can be found in accounts_client.js.
```
stubailo: Why did we decide to move the docs to a different file?

benjamn: Because ES2015 `class` bodies can't be split across multiple files, and I wanted JSDoc to be able to infer the enclosing class from where the `/**...*/` comment is found.

stubailo: I would prefer putting it back here with the `@memberof` annotation.

The main reason we switched to using JSDoc in the first place was to keep the documentation in the same place as the code, and it would be a real shame to break that property to avoid adding a tag to JSDoc, in my opinion.

benjamn: Done.

benjamn:
I believe I've made all the requested changes. Updated demo [here](http://pull-request-4930-accounts-apiautobox-fixes.meteor.com/).

packages/accounts-base/accounts_client.js
```diff
@@ -1,51 +1,132 @@
...
-AccountsClient = function _AccountsClient(options) {
-  AccountsCommon.call(this, options);
-
-  this._loggingIn = false;
-  this._loggingInDeps = new Tracker.Dependency;
-
-  this._loginServicesHandle =
-    this.connection.subscribe("meteor.loginServiceConfiguration");
+/**
+ * @summary Constructor for the `Accounts` namespace on the client.
```
stubailo: Maybe let's call this "client object" instead of namespace? I feel like a namespace isn't something you can construct. Maybe this is too nitpicky.

docs/client/full-api/api/accounts.md
```diff
@@ -255,7 +252,64 @@ Example:
       passwordSignupFields: 'USERNAME_AND_OPTIONAL_EMAIL'
     });
 
-{{> autoApiBox "Accounts.validateNewUser"}}
+<h2 id="advanced_accounts_api"><span>Advanced Accounts</span></h2>
```
stubailo: I'm not a huge fan of the heading "Advanced Accounts" - it could maybe be more descriptive like "accounts for multiple servers" or something, but I can't really come up with anything much better on the spot.

packages/accounts-base/accounts_client.js
stubailo: Also, quick question: is it possible to instantiate one of these on the server? Like if one Meteor server wanted to log into a different Meteor server (we have a similar situation with DDP client/server)

stubailo: Oh, nvm, saw the note: "Note that AccountsClient is currently available only on the client, due to its use of browser APIs such as window.localStorage. In principle, though, it might make sense to establish a client connection from one server to another remote accounts server. Please let us know if you find yourself needing this server-to-server functionality."

Great explanation.

packages/accounts-base/accounts_common.js
```diff
@@ -1,55 +1,223 @@
...
-
-var Ap = AccountsCommon.prototype;
-
 /**
- * @summary Get the current user id, or `null` if no user is logged in. A reactive data source.
- * @locus Anywhere but publish functions
+ * @summary Super-constructor for AccountsClient and AccountsServer.
+ * @locus Anywhere
+ * @class AccountsCommon
+ * @instancename accountsClientOrServer
```
stubailo: Maybe just "accounts" here? It looks a little silly with such a long name. But it's not a huge deal.

stubailo:
Looks good to me mostly! It looks like the backcompat on the URLs is good too.

I'm a tiny bit worried that people might not realize that they can call `AccountsServer#onCreateUser` via `Accounts.onCreateUser`, but maybe we'll cross that bridge when we get there.

packages/accounts-base/accounts_common.js
benjamn: Will do.

docs/client/full-api/api/accounts.md
benjamn: Going with **Accounts (multi-server)** unless anyone objects.

PR #4939, opened 2015-08-11
Implement Promise-returning Meteor.server.{call,apply}Async methods.

benjamn:
This implementation has the (very much intended) side effect of allowing server-side method implementations to return `Promise` objects to delay completion of `Meteor.call` and `Meteor.apply`.

Implementing client-side `Meteor.{call,apply}Async` methods should also be feasible, but that's out of scope for this commit.

Related forum post: https://forums.meteor.com/t/fibers-and-meteor-promise-npm-pacakge/6531/7

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -1538,79 +1538,31 @@ _.extend(Server.prototype, {
     });
   },
 
-  call: function (name /*, arguments */) {
-    // if it's a function, the last argument is the result callback,
-    // not a parameter to the remote method.
-    var args = Array.prototype.slice.call(arguments, 1);
-    if (args.length && typeof args[args.length - 1] === "function")
+  call: function (name, ...args) {
+    if (args.length && typeof args[args.length - 1] === "function") {
+      // If it's a function, the last argument is the result callback, not
+      // a parameter to the remote method.
       var callback = args.pop();
+    }
+
     return this.apply(name, args, callback);
   },
 
-  // @param options {Optional Object}
-  // @param callback {Optional Function}
-  apply: function (name, args, options, callback) {
-    var self = this;
+  // A version of the call method that always returns a Promise.
+  callAsync: function (name, ...args) {
+    return this.applyAsync(name, args);
+  },
 
+  apply: function (name, args, options, callback) {
     // We were passed 3 arguments. They may be either (name, args, options)
     // or (name, args, callback)
-    if (!callback && typeof options === 'function') {
+    if (! callback && typeof options === 'function') {
       callback = options;
       options = {};
     }
     options = options || {};
 
-    if (callback)
-      // It's not really necessary to do this, since we immediately
-      // run the callback in this fiber before returning, but we do it
-      // anyway for regularity.
-      // XXX improve error message (and how we report it)
-      callback = Meteor.bindEnvironment(
-        callback,
-        "delivering result of invoking '" + name + "'"
-      );
-
-    // Run the handler
-    var handler = self.method_handlers[name];
-    var exception;
-    if (!handler) {
-      exception = new Meteor.Error(404, "Method not found");
-    } else {
-      // If this is a method call from within another method, get the
-      // user state from the outer method, otherwise don't allow
-      // setUserId to be called
-      var userId = null;
-      var setUserId = function() {
-        throw new Error("Can't call setUserId on a server initiated method call");
-      };
-      var connection = null;
-      var currentInvocation = DDP._CurrentInvocation.get();
-      if (currentInvocation) {
-        userId = currentInvocation.userId;
-        setUserId = function(userId) {
-          currentInvocation.setUserId(userId);
-        };
-        connection = currentInvocation.connection;
-      }
-
-      var invocation = new DDPCommon.MethodInvocation({
-        isSimulation: false,
-        userId: userId,
-        setUserId: setUserId,
-        connection: connection,
-        randomSeed: DDPCommon.makeRpcSeed(currentInvocation, name)
-      });
-      try {
-        var result = DDP._CurrentInvocation.withValue(invocation, function () {
-          return maybeAuditArgumentChecks(
-            handler, invocation, EJSON.clone(args), "internal call to '" +
-              name + "'");
-        });
-        result = EJSON.clone(result);
-      } catch (e) {
-        exception = e;
-      }
-    }
+    const promise = this.applyAsync(name, args, options);
 
     // Return the result in whichever way the caller asked for it. Note that we
     // do NOT block on the write fence in an analogous way to how the client
@@ -1618,12 +1570,57 @@ _.extend(Server.prototype, {
     // cursor observe callbacks have fired when your callback is invoked. (We
     // can change this if there's a real use case.)
     if (callback) {
-      callback(exception, result);
-      return undefined;
+      promise.then(
+        result => callback(undefined, result),
+        exception => callback(exception)
+      );
+    } else {
+      return promise.await();
     }
-    if (exception)
-      throw exception;
-    return result;
+  },
+
+  // @param options {Optional Object}
+  applyAsync: function (name, args, options) {
+    // Run the handler
+    var handler = this.method_handlers[name];
+    if (! handler) {
+      return Promise.reject(new Meteor.Error(404, "Method not found"));
+    }
+
+    // If this is a method call from within another method, get the
+    // user state from the outer method, otherwise don't allow
+    // setUserId to be called
+    var userId = null;
+    var setUserId = function() {
+      throw new Error("Can't call setUserId on a server initiated method call");
+    };
+    var connection = null;
+    var currentInvocation = DDP._CurrentInvocation.get();
+    if (currentInvocation) {
+      userId = currentInvocation.userId;
+      setUserId = function(userId) {
+        currentInvocation.setUserId(userId);
+      };
+      connection = currentInvocation.connection;
+    }
+
+    var invocation = new DDPCommon.MethodInvocation({
+      isSimulation: false,
+      userId,
+      setUserId,
+      connection,
+      randomSeed: DDPCommon.makeRpcSeed(currentInvocation, name)
+    });
+
+    return new Promise(resolve => resolve(
+      DDP._CurrentInvocation.withValue(
+        invocation,
+        () => maybeAuditArgumentChecks(
+          handler, invocation, EJSON.clone(args),
+          "internal call to '" + name + "'"
+        )
+      )
+    )).then(EJSON.clone);
   },
 
   _urlForSession: function (sessionId) {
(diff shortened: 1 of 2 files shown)


packages/ddp-server/livedata_server.js
```diff
@@ -1601,12 +1567,57 @@ _.extend(Server.prototype, {
...
+    } else {
+      return promise.await();
     }
-    if (exception)
-      throw exception;
-    return result;
+  },
+
+  // @param options {Optional Object}
+  applyAsync: function (name, args, options) {
```
stubailo: I didn't really bring this up before, but call and apply are already asynchronous, in the sense that they take callbacks. Do promises somehow make them more asynchronous?

I guess I'm suggesting finding a new name, like callPromise

packages/ddp-server/livedata_server.js
```diff
@@ -1521,79 +1536,30 @@ _.extend(Server.prototype, {
...
+  call: function (name, ...args) {
+    // If it's a function, the last argument is the result callback,
     // not a parameter to the remote method.
-    var args = Array.prototype.slice.call(arguments, 1);
-    if (args.length && typeof args[args.length - 1] === "function")
-      var callback = args.pop();
-    return this.apply(name, args, callback);
+    return (args.length &&
+            typeof args[args.length - 1] === "function")
+      ? this.apply(name, args, args.pop())
```
stubailo: I'm not really a fan of this code style - I think breaking it up into multiple logical lines makes more sense to me. What really tripped me up is that we are calling pop at the same time as passing the array into apply.

But that's just my opinion, so feel free to ignore.

packages/ddp-server/livedata_server.js
```diff
@@ -692,34 +692,49 @@ _.extend(Session.prototype, {
...
+      promise.then((result) => {
+        finish();
+        if (result !== undefined) {
+          payload.result = result;
+        }
+        self.send(payload);
+      }, (exception) => {
+        finish();
+        payload.error = wrapInternalException(
+          exception, "while invoking method '" + msg.method + "'");
```
stubailo: Convert to template string while we're here?

packages/ddp-server/livedata_server_tests.js
```diff
@@ -156,3 +156,64 @@ Tinytest.addAsync(
...
+
+    const clientCallPromise = new Promise((resolve, reject) => {
+      clientConn.call(
+        "testResolvedPromise",
+        "clientConn.call with callback",
+        (error, result) => error ? reject(error) : resolve(result)
+      );
+    });
+
+    const serverCallAsyncPromise = Meteor.server.callAsync(
```
stubailo: can we add a test for applyAsync too, just in case?

packages/ddp-server/livedata_server.js
benjamn: I think I like that as a general rule: suffix all `Promise`-returning functions with -Promise, instead of -Async.

packages/ddp-server/livedata_server.js
benjamn: I see what you mean about it being confusing. I'll make it clearer.

packages/ddp-server/livedata_server.js
benjamn: Why not!

packages/ddp-server/livedata_server_tests.js
benjamn: Will do.

packages/ddp-server/livedata_server.js
benjamn: Switched back to -Async, per discussion on Slack:
* -Async works better as an adverb than -Promise
* -Async is evocative of / aligned with ES7 `async` functions, which always return `Promise`s
* in a future where we have `async` functions, ideally one won't have to deal directly with things called `Promise`s

lourd:
Nice! Cool to see this. I feel like I've seen this [somewhere](https://github.com/grovelabs/meteor-call-promise) before... :stuck_out_tongue_winking_eye:

benjamn:
I went ahead and pushed [support for returning a `Promise` from a method](https://github.com/meteor/meteor/commit/f5821c88eee587706eb8107f74de2f60c267807f) to https://github.com/meteor/meteor/tree/devel, since that didn't involve any changes to the `Meteor.{call,apply}` API.

I'm not sure we should support `Meteor.server.{call,apply}Async` until we have an analogous story on the client side, and that's trickier because of method stubs.

lourd:
Is there an issue or something tracking the development/design for the `Meteor.{call,apply}Async` methods? I have some thoughts and use cases, not sure where the best place to write them is though.

benjamn:
There's this old hackpad that was written before `Promise`s were a thing: https://meteor.hackpad.com/Separating-call-into-call-and-callAsync-kNjhay2PLlM

Other than that, this PR is probably as good as place as any to discuss the design of these APIs.

My main question for you would be: how much do you care about stub return values? In case that terminology is not familiar, "stubs" are what we call method implementations defined on the client, which can write to Minimongo to update the UI before the actual method call completes. If a stub returns a value, that value is usually only visible to other stubs that called the current stub. However, the hackpad I linked above contains a proposal in which `Meteor.callAsync` would return the result of the stub, but pass the result of the actual method call to the callback. I'm not sure if that use case is something anyone cares about, but I would be happy to know if I'm wrong.

lourd:
Alright, let's get into it :sake: 

> My main question for you would be: how much do you care about stub return values?

Exactly! I've been trying to reason about this the past week. Here's how I'm thinking about it:

There's 4 general outcomes in the flow of a Meteor method from client to server:

1. stub return value
2. stub exception
3. server return value
4. server exception

## Database-only situation
you have a `Posts` collection and a DDP method for inserting posts.

```js
const Posts = new Mongo.Collection('posts');
Meteor.method({
  newPost(doc) {
    check(doc, {
      title: String,
      text: String,
    });
    return Posts.insert(doc)
  }
});
```

Do we care about all 4 cases?

1. Stub return value - you may care about the return value from the stub because you want to route to the new post component, at somewhere like `/posts/_id`
2. Stub exception - you care if there's an exception because that means your arguments were malformed. You also don't need to go run the method on the server then
3. Server return value - you don't really care what this is, as long as there was no exception and the value matches the stub.
4. Server exception - you definitely care because now you've got to handle reverting the UI gracefully

## External server-side API
You have an API that you only call server-side, but you want to simulate the changes immediately

```js
const LightBulbs = new Mongo.Collection('lightBulbs');
Meteor.method({
  setBulbColor(bulbId, color) {
    check(bulbId, String);
    check(color, String);
    const bulb = LightBulbs.findOne(bulbId);

    if ( !this.userId ) {
      throw new Meteor.Error('not-logged-in', 'Seriously bro?');
    }
    if ( !bulb.belongsTo(this.userId) ) {
      throw new Meteor.Error('not-yours', `That ain't your light!`);
    }

    // option 1 -- go async on both sides
    const bulbSwitched ? Meteor.isClient :
      Promise.resolve() :
      bulb.setColor(color); // goes and does some async stuff

    return bulbSwitch.then(() => {
      LightBulbs.update(bulbId, {$set: { color }});
    });

    // option 2 -- keep client sync and let server go async
    if ( Meteor.isClient ) {
      LightBulbs.update(bulbId, {$set: { color }});
    } else {
      return bulb.setColor(color)
        .then((res) => {
          LightBulbs.update(bulbId, {$set: { color }});
        });
    }
  }
}
```

Do we care about all 4 cases?

1. Stub return value - we don't really care about the return value of the update, we care about the side effects, the database updates. There's nothing new being generated.
2. Stub exception - still care, could be giving it wrong type of arguments.
3. Server return value - don't really care about the value itself, just that the method succeeds.
4. Server exception - you definitely care here, because the async API call could have failed. You want to show errors appropriately.

--------

So with those 2 use cases in mind, I actually think that a `Meteor.call` that returns a Promise isn't sufficient if you want to know about all 4 of those cases — you can only find out about 3 of the 4 possible paths. You can't get the return value of the stub *and* and the server. There are certainly cases where that's all that is cared about sometimes, though.

But for the most part, I think it's more isomorphic to use `Meteor.call` with the callback. On the client, you make your stub sync and depend on the side effects. On the server, your method is synchronous because of fibers. On the client, `Meteor.call` returns the method result and throws exceptions if they occur, and the callback gets the possible error and server result. On the server, `Meteor.call` returns the method result and throws exceptions if they occur. (if the callback is given it can be invoked with the same error-result combo, but it's not really needed). So in both cases I can do a `Meteor.call()` and get the same behavior.

benjamn:
@lourd this breakdown of concerns is excellent; thanks for taking the time to write it up. I totally agree with you that something would be lost if we modified `Meteor.call` to return a `Promise` instead of taking a callback, since `Promise`s don't work that well for APIs that need to return multiple values at different times.

This PR is replaced by #5005, where I decided to leave `Meteor.{call,apply}` unchanged, but introduced new `Meteor.{call,apply}Async` methods on both client and server (whereas this PR only implemented async methods on the server). These new methods always return a `Promise` and do not take a callback.

The `Promise` objects returned by `Meteor.{call,apply}Async` default to producing the result/exception from the server, unless you pass `options.returnStubValue` and/or `options.throwStubExceptions` to `Meteor.applyAsync`. The side effects of executing stubs still take place immediately (before the remote call comes back), which seems to be the primary use case for stubs (i.e. writing to Minimongo, rather than returning a value). Also, if you call one stub method from another stub method, the caller receives the stub `Promise` instead of the server `Promise`, so that the simulation can move along without communicating with the server.

While I'm introducing a new API, I'm tempted to make `options.throwStubExceptions` the default for `Meteor.{call,apply}Async`, after reading your analysis. In other words, if a stub is defined, and it throws an exception, the server call never happens. Any feelings about that?

lourd:
No problem, thanks for your work!

I agree on the idea of making `options.throwStubExceptions` the default for the Async methods. I think that's what most people would expect — if my code has an exception, throw it. The alternative has been more difficult to explain and reason about — if my code has an error, it will just log it, and then the call will go to the server, where it will probably also error out, but maybe not if the method's don't match. I would propose making that the default for `Meteor.call` too (in addition to `returnStubValue`, but I guess there's backcompat stuff you have to worry about.

> if you call one stub method from another stub method, the caller receives the stub Promise instead of the server Promise, so that the simulation can move along without communicating with the server.

Hmmm... I'm trying to think of a use case for this but I'm having trouble. When would one be doing async operations in a stub? Or generally calling stubs from other stubs? That feels against the grain of what stubs are intended to be. My inclination is that if I'm using `{call, apply}Async` I only care about the server side results and that's what I expect to be returned in the promise.

deanius:
Can't you use reactivity (in an autorun) to: return stub results/errors, then when the result is ready, return the actual result ? Like this:

deanius:
Like this: https://github.com/okgrow/meteor-promise/blob/master/tests/reactive-promise.es6.js

Issue #4847, opened 2015-07-30
Meteor crashes when HTTP.call receives an error and content is not a string


lintonye:
## Steps to reproduce
- Create a new project `meteor create http-test`
- Add http package `meteor add http`
- Replace the content of `http-test.js` with the following:

```javascript
if (Meteor.isClient) {

  Template.hello.events({
    'click button': function () {
      Meteor.call('get');
    }
  });
}

if (Meteor.isServer) {
  Meteor.methods({
    'get': function() {
      var url = 'http://www.google.com/1.jpg';
      HTTP.call('GET', url, {npmRequestOptions: {encoding: null}});
    }
  })
}
```
- Start Meteor and click the "Click Me" button in the browser

## Result
- After a few seconds, Meteor crashes with the error below:

```
W20150730-15:02:20.278(-7)? (STDERR) TypeError: Object <!DOCTYPE html>
W20150730-15:02:20.278(-7)? (STDERR) <html lang=en>
W20150730-15:02:20.278(-7)? (STDERR)   <meta charset=utf-8>
W20150730-15:02:20.278(-7)? (STDERR)   <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
W20150730-15:02:20.278(-7)? (STDERR)   <title>Error 404 (Not Found)!!1</title>
W20150730-15:02:20.278(-7)? (STDERR)   <style>
[...]

PR #4940, opened 2015-08-11
fix issue #4847 about HTTP package

Rebolon:
The fix just check if content is a string or not. If not, it's because you are using npmRequestOptions.encoding=null so request return a Buffer.

### DIFF
--- a/packages/http/httpcall_common.js
+++ b/packages/http/httpcall_common.js
@@ -5,8 +5,6 @@ makeErrorByStatus = function(statusCode, content) {
     return str.length > length ? str.slice(0, length) + '...' : str;
   };
 
-  // this solution is not browser compatible
-  // var contentToCheck = typeof Buffer != "undefined" && content instanceof Buffer ? content.toString() : content;
   var contentToCheck = typeof content == "string" ? content : content.toString();
   
   var message = "failed [" + statusCode + "]";
--- a/packages/http/httpcall_common.js
+++ b/packages/http/httpcall_common.js
@@ -5,9 +5,14 @@ makeErrorByStatus = function(statusCode, content) {
     return str.length > length ? str.slice(0, length) + '...' : str;
   };
 
+  // this solution is not browser compatible
+  // var contentToCheck = typeof Buffer != "undefined" && content instanceof Buffer ? content.toString() : content;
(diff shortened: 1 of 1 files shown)


packages/http/httpcall_common.js
```diff
@@ -5,9 +5,14 @@ makeErrorByStatus = function(statusCode, content) {
     return str.length > length ? str.slice(0, length) + '...' : str;
   };
 
+  // this solution is not browser compatible
+  // var contentToCheck = typeof Buffer != "undefined" && content instanceof Buffer ? content.toString() : content;
```
stubailo: We shouldn't leave this code in here.

packages/http/httpcall_common.js
```diff
@@ -5,9 +5,14 @@ makeErrorByStatus = function(statusCode, content) {
...
+  var contentToCheck = typeof content == "string" ? content : content.toString();
```
stubailo: I think it would be better to just not print the content to the console in this case? If you're using encoding: null, it's probably binary content and we probably don't want to dump an image to the console?

stubailo:
I think this is a good idea, but we might not want to dump a stringified version of the buffer to the console.

packages/http/httpcall_common.js
Rebolon: Sometime we use encoding: null because of special encoding of the webpage (latin, ...). And when it happens you will have html content.
Maybe we can check if we found <html> node after we have stringified the Buffer. If it's ok, we can continue, if not, we should just return "failed [statusCode]"

Rebolon: But this time we forgot other text format like Json or Xml. Maybe we should just not stringified Buffer and just let the failing message.

Rebolon:
The question is : if we have a statusCode >= 400, does the content can be a binary ? i'm not sure

GauthierD-:
oookkk I understand now why my http.get fail with a 404 error ... 
nice fix :+1:

stubailo:
Sure, I guess we truncate it anyway, so let's go with this for now @Rebolon

stubailo:
Merged in commit beff16e7e1e545bce3bd1b76b3ad2ede104b35b1

PR #4963, opened 2015-08-14
Clarify what Meteor.startup does, eps. on the server

dandv:
A bunch of fellow devs were confused by [what exactly Meteor.startup does, esp. on the server](https://forums.meteor.com/t/solved-when-should-we-use-meteor-startup/2710).


stubailo:
I would say something more like: "Code wrapped in Meteor.startup always runs after all of the app files have loaded, so you should put code here if you want to access shared variables from other files."

stubailo:
Oops, never merged this. Gonna do it.

stubailo:
Merged: 72da78897e5537c1bfd77636ea659be5ed612d3f

PR #4974, opened 2015-08-16
Mention the _id is included in the `user` arg to onCreateUser

dandv:
Reason is people have been assuming onCreateUser is a PRE-create hook, and no user id was available. See

* http://stackoverflow.com/questions/12984637/is-there-a-post-createuser-hook-in-meteor-when-using-accounts-ui-package
* https://forums.meteor.com/t/initializing-some-collections-when-a-user-creates-an-account/3949/6


stubailo:
@dandv it's a little confusing because the document is not yet inserted into the database, but I think your change is strictly an improvement.

stubailo:
Merged to devel: 8a06985fd289139d8edf848e0546340887717aeb

Issue #4933, opened 2015-08-11
Meteor crashes constantly with EPERM on Windows


stubailo:
Originally posted by @fjafjan on the windows-preview repo

I am building a fairly rather simple Meteor project on Windows 8, and despite both re-installing meteor, and manually re-setting the library, after the app has run for a short time meteor crashes and will not restart. Meteor is also unable to run reset, and I haven't been able to identify exactly what it is that causes this arise in the first place. There is a consistent error I get from the ccorcos:swipe, but runtime errors should not be crashing meteor right?

I have tried changing/removing the files in .meteor/client.

My question is both to identify what could be causing this particular error (what does "stat" mean in this context? other error messages I have found have while searching have been far more useful) as well as generally if there is a faster way of hard resetting a meteor project than running meteor create proj_name in a new folder and manually moving over the source files. 

The stacktrace is below.

The packages I use are 

 - meteor-platform 
 - insecure 
 - bevanhunt:leaflet 
 - mdg:geolocation
 - accounts-facebook
 - service-configuration 
 - less stevezhu:lodash
 - ccorcos:swipe
 - iron:router 
 - underscore 
 - coffeescript 
 - accounts-ui
 - accounts-password

```
    C:\Users\Fjafjan\Documents\GitHub\progress>meteor run
    [[[[[ C:\Users\Fjafjan\Documents\GitHub\progress ]]]]]

    => Started proxy.
    => Started MongoDB.

    C:\Users\Fjafjan\AppData\Local\.meteor\packages\meteor-tool\1.1.3\mt-os.windows.x86_32\dev_bundle\lib\node_modules\fibers\future.js:245
                                                    throw(ex);
                                                          ^
    Error: EPERM, stat 'C:\Users\Fjafjan\Documents\GitHub\progress\.meteor\local\build\programs\web.browser\packages'
      at Object.Future.wait (C:\Users\Fjafjan\AppData\Local\.meteor\packages\meteor-tool\1.1.3\mt-os.windows.x86_32\dev_bundle\lib\node_modules\
    fibers\future.js:398:15)
      at Object.wrapper [as stat] (C:\Users\Fjafjan\AppData\Local\.meteor\packages\meteor-tool\1.1.3\mt-os.windows.x86_32\tools\files.js:1350:24
    )
      at C:\Users\Fjafjan\AppData\Local\.meteor\packages\meteor-tool\1.1.3\mt-os.windows.x86_32\tools\files.js:454:43
      at Array.forEach (native)
[...]

PR #4980, opened 2015-08-18
issue #4933: Improved files.rename with handling EPERMs and better wait/retry loop

quetzalcoatl:
Improved files.rename with handling EPERMs and added wait&retry with expanding delays similar to RIMRAF's idea for fixing windows issues


quetzalcoatl:
..but to get it working nice, still an update to rimraf (https://github.com/isaacs/rimraf/issues/86) is needed

PR #4981, opened 2015-08-18
Fix line ending in the middle of "meteor add"

dandv:
Seen at http://docs.meteor.com/#/full/writingpackages


anubhav94:
Thanks @dandv ! Merged.

PR #4987, opened 2015-08-18
corrected redirect URI

kbremner:
The URI has been changed to match the one actually used


jellevandenhooff:
Thanks for the PR! Could you update your patch to compute the URL inside of the `siteUrl` helper in `meet_configure.js` (and perhaps rename the helper to `redirectUrl` at the same time?)

stubailo:
Any updates?

benjamn:

PR #4988, opened 2015-08-18
Correct double negative

Primigenus:
Was this intentional? If so, I don't get it


stubailo:
Nice catch! Merged: 0a29e7614e08f201ff7c98475dbc683b9fc73cd1

PR #5003, opened 2015-08-20
OAuth1Binding - don't mutate the rawHeaders when generating a signature

timhaines:


timhaines:
This fixes a long-standing bug of (post body) params being added to the headers when an oauth signature is generated.  They shouldn't be.

avital:
This looks right, and I'd like to merge it. What's a simple way I can test this behavior? Is it that any call to `new OAuth1Binding(...).call` with params failed?

timhaines:
@avital No - calls with params did not fail.  

The issue that highlighted this problem for me was sending long DMs to Twitter.  Long as in > 2000 characters - Twitter recently updated the max length to 10k, so I was testing that.    

You can quite happily include short params, and Twitter is tolerant of them in the header that they end up in.  But Twitter does reject headers over a certain length.  When the 'text' param grows to > 2000 characters, and is included in the header, Twitter rejects the request.

If you wanted a simple test, you could check that the headers remain the same after they've been passed into this method as a parameter.

avital:
Thanks! Merged into devel and release-1.2

PR #5005, opened 2015-08-20
Implement Promise-returning Meteor.{call,apply}Async methods.

benjamn:
Like #4939, but this PR implements the same functionality on both client and server.

In short, `Meteor.{call,apply}Async` are just like their synchronous counterparts, except that they do not take a callback, and they always return a `Promise`. The eventual value of this `Promise` is the value returned from the server, unless you call `Meteor.applyAsync(method, args, { returnStubValue: true })`, in which case the `Promise` will evaluate to the stub result (though of course the full RPC call will still go through to the server).

The trickiest part was allowing method stubs to return `Promise` objects, while still calling `connection._saveOriginals()` and `connection._retrieveAndStoreOriginals(methodId())` at the right times, in a way that worked for both synchronous `Meteor.{call,apply}` and asychronous `Meteor.{call,apply}Async`. If everything was asynchronous, this would all be a lot easier!

The preliminary "Decompose …" commits demonstrate that there was some opportunity to reuse code between the synchronous and asynchronous versions of these methods.

cc @stubailo @dgreensp @glasser

### DIFF
--- a/packages/ddp-client/client_convenience.js
+++ b/packages/ddp-client/client_convenience.js
@@ -42,18 +42,24 @@ if (Meteor.isClient) {
     }
   };
 
-  Meteor.connection =
+  const conn = Meteor.connection =
     DDP.connect(ddpUrl, {
       onDDPVersionNegotiationFailure: onDDPVersionNegotiationFailure
     });
 
   // Proxy the public methods of Meteor.connection so they can
   // be called directly on Meteor.
-  _.each(['subscribe', 'methods', 'call', 'apply', 'status', 'reconnect',
-          'disconnect'],
-         function (name) {
-           Meteor[name] = _.bind(Meteor.connection[name], Meteor.connection);
-         });
+  _.each([
+    'subscribe',
+    'methods',
+    'call', 'callAsync',
+    'apply', 'applyAsync',
+    'status',
+    'reconnect',
+    'disconnect'
+  ], name => {
+    Meteor[name] = conn[name].bind(conn);
+  });
 } else {
   // Never set up a default connection on the server. Don't even map
   // subscribe/call/etc onto Meteor.
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -1039,7 +1039,17 @@ _.extend(Connection.prototype, {
     return this.applyAsync(name, args);
   },
 
-  applyAsync(name, args, options = {}) {
+  applyAsync(name, args, {
+    onResultReceived,
+    returnStubValue = false,
+    // Note that the default value of options.throwStubExceptions for
+    // Connection#applyAsync is different from Connection#apply. Unless
+    // the caller explicitly disables options.throwStubExceptions,
+    // exceptions thrown by stubs will be exposed to the calling code, and
+    // will prevent the RPC from being sent to the server.
+    throwStubExceptions = true,
+    wait = false,
+  } = {}) {
     // Keep our args safe from mutation (eg if we don't send the message for a
     // while because of a wait method).
     args = EJSON.clone(args);
@@ -1089,19 +1099,19 @@ _.extend(Connection.prototype, {
 
       return rpcPromise = new Promise((resolve, reject) => {
         this._enqueueMethodInvoker(new MethodInvoker({
+          message,
+          onResultReceived,
           methodId: methodId(),
-          callback: (error, result) => {
+          callback(error, result) {
             error ? reject(error) : resolve(result);
           },
           connection: this,
-          onResultReceived: options.onResultReceived,
-          wait: !!options.wait,
-          message: message
-        }), options.wait);
+          wait: !! wait
+        }), wait);
       })
     }
 
-    if (options.throwStubExceptions === false) {
+    if (! throwStubExceptions) {
       // If we explicitly don't care about stub exceptions, then we can go
       // ahead and send the RPC without waiting on the stub.
       ensureRPCSent();
@@ -1109,12 +1119,12 @@ _.extend(Connection.prototype, {
 
     return stubPromise.then(result => {
       const rpcPromise = ensureRPCSent();
-      return options.returnStubValue ? result : rpcPromise;
+      return returnStubValue ? result : rpcPromise;
     }, exception => {
         // If an exception occurred in the stub, and we did not explicitly
         // pass options.throwStubExceptions === false, then abort the RPC
         // and throw the stub exception instead.
-        if (options.throwStubExceptions !== false) {
+        if (throwStubExceptions) {
           throw exception;
         }
 
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -1066,42 +1066,55 @@ _.extend(Connection.prototype, {
     // At this point we're definitely doing an RPC, and we're going to
     // return the value of the RPC to the caller.
 
-    // Send the RPC. Note that on the client, it is important that the
-    // stub have finished before we send the RPC, so that we know we have
-    // a complete list of which local documents the stub wrote.
-    var message = {
-      msg: 'method',
-      method: name,
-      params: args,
-      id: methodId()
-    };
+    let rpcPromise;
+    const ensureRPCSent = () => {
+      if (rpcPromise) {
+        return rpcPromise;
+      }
 
-    // Send the randomSeed only if we used it
-    if (randomSeedGenerator.seed) {
-      message.randomSeed = randomSeedGenerator.seed;
+      // Send the RPC. Note that on the client, it is important that the
+      // stub have finished before we send the RPC, so that we know we have
+      // a complete list of which local documents the stub wrote.
+      const message = {
+        msg: 'method',
+        method: name,
+        params: args,
+        id: methodId()
+      };
+
+      // Send the randomSeed only if we used it
+      if (randomSeedGenerator.seed) {
+        message.randomSeed = randomSeedGenerator.seed;
+      }
+
+      return rpcPromise = new Promise((resolve, reject) => {
+        this._enqueueMethodInvoker(new MethodInvoker({
+          methodId: methodId(),
+          callback: (error, result) => {
+            error ? reject(error) : resolve(result);
+          },
+          connection: this,
+          onResultReceived: options.onResultReceived,
+          wait: !!options.wait,
+          message: message
+        }), options.wait);
+      })
     }
 
-    const rpcPromise = new Promise((resolve, reject) => {
-      this._enqueueMethodInvoker(new MethodInvoker({
-        methodId: methodId(),
-        callback: (error, result) => {
-          error ? reject(error) : resolve(result);
-        },
-        connection: this,
-        onResultReceived: options.onResultReceived,
-        wait: !!options.wait,
-        message: message
-      }), options.wait);
-    });
+    if (options.throwStubExceptions === false) {
+      // If we explicitly don't care about stub exceptions, then we can go
+      // ahead and send the RPC without waiting on the stub.
+      ensureRPCSent();
+    }
 
-    return stubPromise.then(
-      result => options.returnStubValue ? result : rpcPromise,
-      exception => {
-        // If an exception occurred in a stub, and we're ignoring it
-        // because we're doing an RPC and want to use what the server
-        // returns instead, log it so the developer knows
-        // (unless they explicitly ask to see the error).
-        if (options.throwStubExceptions) {
+    return stubPromise.then(result => {
+      const rpcPromise = ensureRPCSent();
+      return options.returnStubValue ? result : rpcPromise;
+    }, exception => {
+        // If an exception occurred in the stub, and we did not explicitly
+        // pass options.throwStubExceptions === false, then abort the RPC
+        // and throw the stub exception instead.
+        if (options.throwStubExceptions !== false) {
           throw exception;
         }
 
@@ -1115,7 +1128,7 @@ _.extend(Connection.prototype, {
           );
         }
 
-        return rpcPromise;
+        return ensureRPCSent();
       }
     );
   },
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -792,23 +792,8 @@ _.extend(Connection.prototype, {
     var enclosing = DDP._CurrentInvocation.get();
     var alreadyInSimulation = enclosing && enclosing.isSimulation;
 
-    // Lazily generate a randomSeed, only if it is requested by the stub.
-    // The random streams only have utility if they're used on both the client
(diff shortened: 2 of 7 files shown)


stubailo:
I love the idea of client stubs throwing errors!

zimme:
Wouldn't it be possible to have the `call`/`apply` calls return promises if there is no callback specified?

Is it really necessary to have separate `callAsync`/`applyAsync` functions for this?

packages/ddp-client/client_convenience.js
```diff
@@ -42,18 +42,24 @@ if (Meteor.isClient) {
...
+    'subscribe',
+    'methods',
+    'call', 'callAsync',
+    'apply', 'applyAsync',
+    'status',
+    'reconnect',
+    'disconnect'
+  ], name => {
+    Meteor[name] = conn[name].bind(conn);
+  });
```
lourd: Could use the builtin `Array.forEach` here instead of Underscore

packages/ddp-client/livedata_connection.js
```diff
@@ -873,42 +821,255 @@ _.extend(Connection.prototype, {
...
+    self._enqueueMethodInvoker(new MethodInvoker({
       methodId: methodId(),
       callback: callback,
       connection: self,
       onResultReceived: options.onResultReceived,
       wait: !!options.wait,
       message: message
-    });
+    }), options.wait);
+
```
lourd: Could use shorthand here:

```js
message,
callback,
methodId: methodId(),
connection: self,
onResultReceived: options.onResultReceived,
wait: !!options.wait
```

packages/ddp-client/livedata_connection.js
```diff
@@ -873,42 +821,255 @@ _.extend(Connection.prototype, {
...
+      if (randomSeedGenerator.seed) {
+        message.randomSeed = randomSeedGenerator.seed;
+      }
+
+      return rpcPromise = new Promise((resolve, reject) => {
+        this._enqueueMethodInvoker(new MethodInvoker({
+          methodId: methodId(),
+          callback: (error, result) => {
+            error ? reject(error) : resolve(result);
+          },
```
lourd: Can do the full shorthand for an object literal function instead of arrow syntax

```js
callback(error, result) {
  error ? reject(error) : resolve(result);
}
```

packages/ddp-client/livedata_connection.js
benjamn: Yep, good call.

packages/ddp-client/livedata_connection.js
benjamn: Oh, right! Much better.

deanius:
Can you use reactivity (in an autorun) to: return stub results/errors, then when the result is ready, return the actual result ? Like this: https://github.com/okgrow/meteor-promise/blob/master/tests/reactive-promise.es6.js

lourd:
I went ahead and polyfilled this until this is merged [here](https://github.com/grovelabs/meteor-call-async). Probably missed a couple edge cases, but the tests cover the primary ones. The package that existed there is now better named as `grove:call-sync` and moved [here](https://github.com/grovelabs/meteor-call-sync).

mitar:
I got a crazy idea and I think we should do it. :-)

So why Meteor promises would not be both standard-compliant and reactive?

So what I have in mind is that a method could return a promise which would have the following methods:
* `then` – standard one
* `catch` – standard one
* `done` – a reactive data source which returns the value once the promise is resolved, otherwise `undefined`
* `error` – a reactive data source which returns the error value once the promise fails, otherwise `undefined`
* `return` – a reactive data source which returns the value once the promise is resolved, or throws an exception if promise fails

This would allow the whole Meteor API to have the same interface both for callback-based code, but also for reactive code. The same we should do for current chaos of callback and reactive states. For example, `Meteor.subscribe` currently accepts both two callback, but also return a handle which can be used as a reactive source. Wouldn't it be much cleaner if it would simple return a promise and then we can map:

* `onReady` – to `then`, with a value which is another promise which returns once the subscription stops
* `onStop` – if it is error, it is mapped to `catch`
* `ready` – mapped to `done`

Of course, promise could also be extended with extra methods, like `stop()` and `subscriptionId`. But the base class would be the same.

Those are just example, but it would solve a big problem for me in Meteor which is that there are often cases where some state change is observable only through callback, some are through reactive variable, but I believe that all should be provided through all. And promises could refactor all of them together.

dgreensp:
@mitar, I've had the same thought that intuitively, waiting for a subscription to be ready should be like waiting for a Promise.  It would be nice if you could write `Meteor.subscribe(...).then(...)`.  However, I don't really see how the idea of a subscription handle being a Promise goes farther than that.  There's no "value" that becomes available only when a subscription is ready, and the fact that the handle may be stopped doesn't map to anything.  The Promise aspect of a subscription handle would be a hack that makes it easy to await `ready()`, and captures literally nothing else.

If we really wanted to use Promises instead of callbacks to respond to state changes, we could cook up a pattern like:

```
var handle = Meteor.subscribe(...).beReady().then(...)
await handle.beStopped()
```

The properties `ready` and `stopped` can be viewed either as reactive booleans you may want to observe, or as state-change events you might want to respond to with a callback, and here we'd be adding a third interpretation, that the property going from `false` to permanent `true` represents the resolution of a value-less Promise.

The point is, it seems to make some amount of sense to map a subscription being ready to a Promise resolving, but it's not because of some fundamental alignment between subscriptions and Promises, but because there's a nice API for waiting for a Promise to resolve, compared to waiting for an object to reactively change state.

deanius:
Good idea @mitar , but i see the relationship going the other way - The handle returned by `subscribe` can contain a promise field. 

https://github.com/deanius/meteor-promise/blob/master/tests/subscription-ready.js

I like the idea of unifying reactivity with promises - this is a 'promising' discussion !

mitar:
Hm, maybe there is some more background needed here. So all started when I was thinking of how to good communication in Blaze Components ([more on this here](https://github.com/peerlibrary/meteor-blaze-components/issues/82)). In short, I am seeing reactive variables as strict super set of events. It is like events + state.

So you can use reactivity both to do things like:

```javascript
Tracker.autorun(function (computation) {
  dragging.set(mouseDown.get() && mouseMoved.get() && !mouseUp.get());
});
```

With events you would have to remember state manually. On the other hand, it is easy (with a helper which abstract the stuff below) to do events:

```javascript
Tracker.autorun(function (computation) {
  if (!dragging.get() && isOnline.get()) return;
  computation.stop();

  // Do something on start of dragging and user is online.
});
```

So I really like how you can compose stuff and make things happen when multiple conditions occur. This is really hard to do with events.

Now enter promises. The are not completely the same as events, but they are still the old-school callback style programming. You have to define callbacks to handle them. Combining them requires new helper functions (while people are used to autorun). And they can be resolved only once, you cannot change the values afterwards. So in some way for me it looks again just a special (limited) case of a reactive variable.

So this is why I am suggesting that promises could be simply a reactive variables with additional helper functions. So if we look at current reactive variable. Isn't promise then only:

```javascript.
ReactiveVar.prototype.then = function (fun) {
  var self = this;
  self._thenHandle = Tracker.autorun(function (computation) {
    if (!self.get()) return;
    computation.stop();
    if (self._errorHandle) self._errorHandle.stop();
    Tracker.nonreactive(function () {
      fun(self.get());
    });
  });
});
```

So those are just ideas. I do not have a concrete proposal. But for me it really feels that we are missing an opportunity here and we could design a great system where we have a Meteor reactive variable for use in Meteor, but then also it is compatible with other code which expects promises. (This is especially useful now that we also have [server-side autorun](https://github.com/peerlibrary/meteor-server-autorun).)

So yea, maybe this is just what you are saying, an extension to reactive variable to expose also promise-like interface. But what I would also like to see is the opposite, that promises behave like reactive variables as well. So that you can combine them inside autorun if needed. So maybe we do not have to change much in `subscribe`, only add promise interface to all its reactive variables.

Maybe all Meteor promises should have `promise.return()` which would be a reactive source which returns the value (once promise resolves, and from then on the same value all the time), or throws an exception of promise fails.

mitar:
Here is [some code which uses this pattern](https://github.com/mitar/council-app/blob/master/packages/core/documents/storagefile.coffee). See `UploadStatus`. This is for me such empowered promise. It can has both `done` and `error`, but also reactive progress. I can then map current [method call callback to it](https://github.com/mitar/council-app/blob/master/packages/core/documents/storagefile.coffee#L60).

And then I can autorun-wait for the [state change](https://github.com/mitar/council-app/blob/master/packages/council/editor.coffee#L35).

So the API here is not yet the cleanest. But this is where I got an idea that there is something here. So there is something in this combining of callbacks with promises with reactivity.

benjamn:
I very much want to use `Promise`s everywhere I/we can, but they're not always a good fit for ongoing sequences of events.

There are a couple of patterns floating around for asynchronous event streams in JS.

One is [`AsyncIterator`](https://github.com/zenparsing/async-iteration/), which is a kind of iterator object whose `.next()` method returns `Promise<{value,done}>` instead of `{value,done}` objects. Given an async iterator `ait`, you're supposed to consume it like this:
```js
ait.next().then(function handler({value, done}) {
  if (! done) {
    // do something with value
    ait.next().then(handler);
  }
}).catch(error => ...);
```
You have to really love `Promise`s to find this code pretty, but it's arguably the most natural way of generalizing the `Iterator` interface to handle asynchronous sequences.

Another pattern is [`Observable`](https://github.com/zenparsing/es-observable), which represents a stream of values that can have multiple subscribers. Creating an `Observable` object is kind of complicated (similar to creating a `Promise` object), but consuming it is very easy:
```js
// Subscribe to a stream of keydown events on inputElement.
let subscription = listen(inputElement, "keydown").subscribe({
    next(val) { console.log("Received key command: " + val) },
    error(err) { console.log("Received an error: " + err) },
    complete() { console.log("Stream complete") },
});

// Unsubscribe from the sequence at any time:
subscription.unsubscribe();
```
Here's [a full example](https://github.com/zenparsing/es-observable/blob/master/README.md), including an implementation of the `listen` function.

Of all these abstractions (`Promise`, `AsyncIterator`, and `Observable`), I think the one that makes the most sense as a reactive data source is `Observable`. Using a `Promise` as a reactive variable makes sense if there's only one state transition (from undefined to the final value), but `Observable`s allow for change over time, cancelation of individual subscriptions, and methods like `.map` and `.filter`.

The `Observable` proposal is at stage 1 right now, so I expect it to be finalized in ECMAScript 2016. We'll add a polyfill to the `ecmascript-runtime` package as soon as it moves to stage 2.

mitar:
> Using a Promise as a reactive variable makes sense if there's only one state transition (from undefined to the final value),

Yes, and many status variables in Meteor are like that. So I think for now I would go just for those. `Observable` can be later on added on top of existing cursor API.

BTW, will there be `cursor.iterator()` available?

tmeasday:
@benjamn the only downside of this (otherwise awesome) approach is that there's (AFAICT) no way to get both the stub and true return value. If I'm right about this, how can I implement the following optimistic UX?:

```js
const options = {throwStubExceptions: true, returnStubValue: true};
const post = Meteor.apply('Post.create', args, options, (error, realPost) => {
  if (error) {
    // freak out, undo what I did optimistically below
  }

  // possibly do something new in the UI with the realPost data
});

// optimistically do something with the stub's return value
Router.go('postShow', post);
```

If the answer is "you can't with `callAsync`", I guess that's OK, but it's kind of frustrating to not have a single API that allows all the different things you might want to do, UX wise.

I'm not sure what is an isn't kosher for "subclassing" a promise, but what would seem to be a better API to me would be something like:

```js
Meteor.callAsync('Post.create', args)
  .thenStub(post => { // or maybe `.stub.then()`?
    // optimistically do something with the stub's return value
  })
  .catch(error => {
      // freak out, undo what I did optimistically above
  })
  .then(realPost => {
    // possibly do something new in the UI with the realPost data
  });
```

benjamn:
It might not be totally crazy for `Meteor.callAsync` to add additional methods to the `Promise` it returns, e.g.:
```js
callAsync(name, ...args) {
  let promise = this.applyAsync(name, args);
  promise.now = function now(onStubSuccess, onStubError) {...};
  return promise;
}
```

tmeasday:
I am advocating for an approach where stub results/exceptions aren't the default that people would use without thinking (clearly that should be the server result/exception), but that it *is* useful in many cases, and it'd be nice to be able to use it without passing esoteric options.

If we made this happen by default, the other question would be should a stub exception stop the method from executing on the server? I think this is something that's useful, as then you can use stubs for client-side validation of methods, but perhaps it's too surprising of a default. @stubailo what do you think?

mitar:
Hoho, but maybe then having reactive variables instead of promises solves this, because you can first return a stub value and once the real value comes to the client, reactive variable updates to that value, or stays the same if everything is OK. In this respect than it works the same as other database operations on the client side. So if you use promise as a promise then you loose stub operations (callback is called only after server confirms), but if you use them as reactive variables then you can have both.

tmeasday:
@mitar - I'm not sure though because you need to differentiate between the two. So now do you have the reactive var returning `{result, isSimulation}`? That's kind of ugly.

On the subject of the "equivalency" of promises and reactive vars, I certainly agree that not all reactive vars work as promises, but @benjamn have you ever thought about on the client somehow making all promises reactive? Something like `promise.reactiveValue()`?

mitar:
> I'm not sure though because you need to differentiate between the two.

Do you? With Minimongo you do not. You just get a value which is a result of a simulated method call, and once the method call returns it is hopefully the same value and there is nothing to be changed, or the output is otherwise updated to the last/true value.

tmeasday:
Definitely. There are a variety of UX patterns that you might want to follow, depending on how optimistic you are about the method succeeding. In some, you do stuff immediately with the simulated value; in others, you wait for the real value to come back and show a "waiting" state in the meantime.

mitar:
Some ideas then:

```javascript
promise.value()
```

```javascript
promise.finalValue()
```

```javascript
promise.isSimulated()
```

```javascript
promise.error()
```

```javascript
promise.return()
```

So `value` changes first to simulation, then the final value. `finalValue` changes to final value only. `isSimulated` is true until the `value` is final. `error` returns an exception if something happen. `return` returns the `value`, but if there is an exception, it triggers dependency invalidation, but then throws the exception.

benjamn:
Closing this now because it's been a long time, and these changes didn't really preserve the concept of optimistic updates (stubs), since a function returning a `Promise` can only have one result.

deanius:
Just saying- did this thread continue somewhere or is it unresolved- @benjamn ? I'm super-interested in making Observables and these kinds of APIs ergonomic in Meteor and have been prototyping some of my own patterns.

mitar:
I opened https://github.com/meteor/meteor-feature-requests/issues/327. I think one way to resolve this would be that promises are returned only if `returnStubValue` is set to `false`. Then the promise would only serve as a way to return the value from the server (once callback comes back), but not the value during simulation.

Issue #2294, opened 2014-07-09
Spiderable package removes all script tags, preventing the use of JSON-LD


sbking:
When trying to include rich data for search engines using JSON-LD, you need to include a `script` tag with `type="application/ld+json"`. [Spiderable doesn't seem to discriminate between different types of `script` tags, removing them all](https://github.com/meteor/meteor/blob/devel/packages/spiderable/spiderable.js#L85). 

I believe an exception should be made for scripts with type `application/ld+json`, because the JSON-LD method of adding rich data to your markup works **really** well with Meteor. You can define schema.org data in one place (Collections or objects in your JS), use it in your templates with clean markup free of microdata/RDFa attributes, and then just `JSON.stringify()` the schema data into a `<script type="type="application/ld+json">` tag.

I have it working for my project by changing the RegExp on spiderable.js:85 to:

```
/<script(?![^>]+ld\\+json)[^>]+>(.|\\n|\\r)*?<\\/script\\s*>/ig
```

hwillson:
Hi all - deprecating the `spiderable` package is currently being considered. The reasons for this include:

- Google has deprecated the [AJAX crawling specification](https://webmasters.googleblog.com/2015/10/deprecating-our-ajax-crawling-scheme.html) that the `spiderable` package is based on;
- Search engine crawlers (like Google's) have gotten much better at dealing with javascript/AJAX based applications, without the need for any special handling;
- There are now several better options for SEO goodness, like pre-rendering services (e.g. [Prerender.io](https://prerender.io/) - recommended in the Guide's [Enabling SEO](https://guide.meteor.com/deployment.html#seo) section and [supported out of the box with Galaxy](http://galaxy-guide.meteor.com/seo.html)), server-side rendering techniques, etc.

As such, I'll close this issue off for now. If anyone feels like this is a bad decision and can justify why `spiderable` should not be deprecated, please post back with your reasons. Thanks for opening this issue originally!

P.S. --> Anyone who wants to keep using the `spiderable` package after it has been deprecated will still be able to, by version pinning to the last non-deprecated version of `spiderable`.

PR #5015, opened 2015-08-23
Spiderable: Replace regex with DOM manipulation + keep JSON-LD

tomwasd:
I think/hope @glasser suggested something along these lines in this issue: https://github.com/meteor/meteor/issues/2294

It gets rid of the regex for removing script tags and the fragment meta tag and uses DOM manipulation instead. It also keeps JSON-LD tags which are useful for semantic SEO - https://developers.google.com/schemas/formats/json-ld?hl=en


tmeasday:

PR #5016, opened 2015-08-23
Remove unused RunCommand and add promise-based execFileSync/Async

martijnwalraven:
The execFileSync function is meant to resemble the similarly-named Node
0.12 synchronous process creation API, but instead of being fully
blocking it uses a promise-based implementation. You can also use
execFileAsync directly, which returns a promise.

Some functionality is currently missing but could be added when the
need arises (e.g. support for timeout, maxBuffer, and encoding options).

Eventually, these versions should replace the ones in
tools/utils/utils.js and tools/tool-testing/selftest.js.

### DIFF
--- a/tools/utils/processes.js
+++ b/tools/utils/processes.js
@@ -1,123 +1,144 @@
-var Future = require('fibers/future');
-var _ = require('underscore');
-var child_process = require('child_process');
-var Console = require('../console/console.js').Console;
+import _ from 'underscore';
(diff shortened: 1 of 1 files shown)


stubailo:
Looks great to me!

martijnwalraven:
Thanks for having a look at it! I'll merge it as part of the cordova-improvements branch, because `commands-cordova.js` on devel still depends on `RunCommand`.

PR #5024, opened 2015-08-25
Add case-insensitive accounts-password utility functions

stubailo:
You can see the added methods by looking at the changes to `names.json`, or the docs ToC.

@avital can you review this please?


docs/client/full-api/api/passwords.md
```diff
@@ -47,6 +47,31 @@ override this behavior, use [`Accounts.onCreateUser`](#accounts_oncreateuser).
 This function is only used for creating users with passwords. The external
 service login flows do not use this function.
 
+### Managing usernames and emails
+
+Instead of modifying documents in the [`Meteor.users`](#meteor_users) collection directly, use these convenience functions which correctly check for case insensitive duplicates before updates.
```
avital: 80c

docs/client/full-api/api/passwords.md
```diff
@@ -47,6 +47,31 @@ override this behavior, use [`Accounts.onCreateUser`](#accounts_oncreateuser).
...
+
+{{> autoApiBox "Accounts.setUsername"}}
+
+{{> autoApiBox "Accounts.addEmail"}}
+
+By default, an email address is added with `{ verified: false }`. Use
+[`Accounts.sendVerificationEmail`](#Accounts-sendVerificationEmail) to send an email with a link the user can use verify their email address.
```
avital: 80c

packages/accounts-password/password_server.js
```diff
@@ -758,7 +833,97 @@ Meteor.methods({verifyEmail: function (token) {
...
+        'emails.$.verified': verified
+      }});
+      return true;
+    }
+
+    return false;
+  });
+
+  if (didUpdateOwnEmail) {
+    return;
```
avital: I was a little confused at first as to why we don't need something like the "call `checkForCaseInsensitiveDuplicates` both before and after the Mongo operation" pattern here. But then I understood that other than modifying Mongo directly (in which case you're on your own), there's no way for another user to add `newEmail` while this user has it in another case.

Maybe a comment explaining this? Up to you, it may be clearer than I make it seem.

packages/accounts-password/password_tests.js
```diff
@@ -1312,4 +1330,154 @@ if (Meteor.isServer) (function () {
...
+    var thirdEmail = origEmail.toUpperCase();
+    Accounts.addEmail(userId, thirdEmail, true);
+
+    test.equal(Accounts._findUserByQuery({id: userId}).emails, [
+      { address: thirdEmail, verified: true },
+      { address: newEmail, verified: false }
+    ]);
+  });
+
+  Tinytest.add("passwords - add email should fail when there is an existing user with an email only differing in case", function (test) {
```
avital: 80c

packages/accounts-password/password_tests.js
```diff
@@ -1312,4 +1330,154 @@ if (Meteor.isServer) (function () {
...
+      { address: thirdEmail, verified: true }
+    ]);
+
+    Accounts.removeEmail(userId, origEmail);
+
+    test.equal(Accounts._findUserByQuery({id: userId}).emails, [
+      { address: thirdEmail, verified: true }
+    ]);
+  });
+
```
avital: No tests for `Accounts.findUserByUsername` and `Accounts.findUserByEmail`?

avital:
A few comments, one about missing tests, the rest super minor.

Looks great!

packages/accounts-password/password_server.js
stubailo: Added a comment

packages/accounts-password/password_tests.js
stubailo: Added one for each!

stubailo:
@avital I think I've addressed everything, give it another pass?

packages/accounts-password/password_server.js
avital: I'm confused about the location of the comment -- it says the words "We don't have to do a case insensitive check here" right before a line that defines a case insensitive regexp. Maybe you meant to put the comment somewhere else?

packages/accounts-password/password_tests.js
avital: Looks great

stubailo:
Merged as one commit: dac019fda8a5e70b2dbec335b6605b04c0c034bc

PR #5034, opened 2015-08-26
Add remove function to Reactive-Dict

msavin:
I'm looking forward to Dict.all() for debugging tools and building utilities. We also need a way to permanently remove keys from the dictionaries, because setting them to null will still cause them to be returned by Dict.all().

### DIFF
--- a/packages/reactive-dict/reactive-dict.js
+++ b/packages/reactive-dict/reactive-dict.js
@@ -165,6 +165,14 @@ _.extend(ReactiveDict.prototype, {
     });
 
   },
+  
+  remove: function (key) {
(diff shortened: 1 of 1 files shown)


kookoskar:
There's also #4736 open, which seems to have more discussion.

skirunman:
This was done in https://github.com/meteor/meteor/commit/72c6622b4223d90d61d775118530021e0dff207b

benjamn:
Feel free to reopen this if https://github.com/meteor/meteor/commit/72c6622b4223d90d61d775118530021e0dff207b does not address your needs.

PR #5038, opened 2015-08-26
Regression test for #2691

mitar:
cc @avital 

It does not pass. I think because your fix is not correct.


avital:
Yes, you're right. My fix was incorrect and I reverted it. I'll link to this from #2691

avital:
.

avital:
.

avital:
Leaving this open for some future time when we address #2691

benjamn:

keyscores:
@benjamn I think this test should be included since the original issue that motivated this test is still open. It's not a trivial bug.

PR #5044, opened 2015-08-27
Document the sha package

dandv:
The sha package is listed on Meteor's homepage under Utilities:

https://www.meteor.com/utilities

### DIFF
--- a/packages/sha/package.js
+++ b/packages/sha/package.js
@@ -1,12 +1,10 @@
 Package.describe({
-  name: 'sha256',
   version: '1.0.4-plugins.0'
   summary: 'SHA256 implementation',
-  git: 'https://github.com/meteor/meteor/tree/devel/packages/sha',
-  documentation: 'README.md'
+  git: 'https://github.com/meteor/meteor',
 });
 
 Package.onUse(function (api) {
   api.export('SHA256');
-  api.addFiles('sha256.js', ['client', 'server']);
+  api.addFiles('sha256.js');
 });
--- a/packages/sha/package.js
+++ b/packages/sha/package.js
@@ -1,9 +1,12 @@
 Package.describe({
-  summary: "SHA256 implementation",
-  version: "1.0.4-plugins.0"
+  name: 'sha256',
+  version: '1.0.4-plugins.0'
+  summary: 'SHA256 implementation',
+  git: 'https://github.com/meteor/meteor/tree/devel/packages/sha',
+  documentation: 'README.md'
 });
 
 Package.onUse(function (api) {
   api.export('SHA256');
-  api.addFiles(['sha256.js'], ['client', 'server']);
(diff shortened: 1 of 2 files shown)


dgreensp:
I think it's normal style for core packages to lack `name` and `documentation`.

dandv:
I've used the style generated by `meteor create --package`, including apostrophes instead of double quotes.

packages/sha/package.js
```diff
@@ -1,9 +1,12 @@
 Package.describe({
-  summary: "SHA256 implementation",
-  version: "1.0.4-plugins.0"
+  name: 'sha256',
+  version: '1.0.4-plugins.0'
+  summary: 'SHA256 implementation',
+  git: 'https://github.com/meteor/meteor/tree/devel/packages/sha',
+  documentation: 'README.md'
 });
```
dgreensp: put only `summary` and `version` here

packages/sha/package.js
```diff
@@ -1,9 +1,12 @@
...
 
 Package.onUse(function (api) {
   api.export('SHA256');
-  api.addFiles(['sha256.js'], ['client', 'server']);
+  api.addFiles('sha256.js', ['client', 'server']);
```
dgreensp: remove `['client', 'server']`

dgreensp:
Core packages are different.  Make these changes and I will accept.  Thanks for the README!

packages/sha/package.js
dandv: `package.js` had `['client', 'server']` to begin with; I had just left it alone. Gone now.

packages/sha/package.js
dandv: Can we please have the `git` field here so https://atmospherejs.com/meteor/sha has a GitHub/issues link?

packages/sha/package.js
```diff
@@ -1,9 +1,10 @@
...
-  version: "1.0.4-plugins.0"
+  version: '1.0.4-plugins.0'
+  summary: 'SHA256 implementation',
+  git: 'https://github.com/meteor/meteor',
 });
 
 Package.onUse(function (api) {
   api.export('SHA256');
-  api.addFiles(['sha256.js'], ['client', 'server']);
+  api.addFiles('sha256.js');
```
dgreensp: you're right, `['client', 'server']` was there before, I didn't notice that

dgreensp:
According to my sources, Atmosphere will truncate the git URL to just `https://github.com/meteor/meteor`, which is something we want to fix.  Then we will add git URLs to everything.

dandv:
Right. My [last commit](https://github.com/dandv/meteor/commit/d54df8321bf1a8c6d13083a9bcf82c9be1131fb3#diff-869c8e16c7aa72a982167b7a5e4ed500R4) does that truncation.

dgreensp:
ok, merged, thanks

PR #5056, opened 2015-08-28
Fix `echo | meteor shell`

### LINKED ISSUE #5055: `echo 1+1 | meteor shell` should print "2"
This makes it very hard to use `meteor shell` for scripting.

```
$ meteor &

$ echo | meteor shell

/Users/avital/.meteor/packages/meteor-tool/.1.1.4.1k57gpn++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/tools/shell-client.js:106
    process.stdin.setRawMode(true);
                  ^
TypeError: Object #<Socket> has no method 'setRawMode'
    at Socket.onConnect (/Users/avital/.meteor/packages/meteor-tool/.1.1.4.1k57gpn++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/tools/shell-client.js:106:19)
    at Socket.emit (events.js:117:20)
    at Object.afterConnect [as oncomplete] (net.js:887:10)
```

avital:
Fixes #5055

### DIFF
--- a/tools/shell-client.js
+++ b/tools/shell-client.js
@@ -103,7 +103,9 @@ Cp.setUpSocket = function setUpSocket(sock, key) {
 
     process.stderr.write(shellBanner());
     process.stdin.pipe(sock);
-    process.stdin.setRawMode(true);
+    if (process.stdin.setRawMode) { // https://github.com/joyent/node/issues/8204
+      process.stdin.setRawMode(true);
+    }
   }
 
   function onClose() {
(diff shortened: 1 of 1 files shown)


benjamn:
Calling `process.stdin.setRawMode` is only important when you're actually typing keystrokes (like control-D) into the shell. When you're echoing text, all that matters is the text.

My only concern here is that I'm not sure `echo "1+1" | meteor shell` actually echoes `2` to stdout, even with this change. It still seems to start up the shell. So this change is definitely necessary, but perhaps not sufficient to fix #5055.

avital:
Right. It does take time for the shell to load, and the shell doesn't stop until you write `.exit`. There's no way to tell `meteor shell`: "wait until this operation is complete, then exit".

In the benchmarking script that I'm writing, I did:
```
(sleep 5 && echo "$METEOR_SHELL_CODE" && sleep 8 && echo .exit) | ~/meteor/meteor shell 2>&1
```

I'll merge this pull request in and leave #5055 open with a comment explaining what would need to get done to make `meteor shell` totally awesome for scripting.

benjamn:
Sounds good to me!

PR #5072, opened 2015-08-30
Add missing brackets around hexstring opt param

vjau:
In the JSDoc of the Mongo.objectId() method, the param name hexstring was missing brackets, causing the data.js generating script to forget the optional:true attribute for the param.


stubailo:
Thanks! Merged: e5d0badc386bd5bb630a1481ba91844cc8a61f84

PR #5073, opened 2015-08-30
[1.2 RC] Add JSX extension on ecmascript

eXon:
Since we are already using Babel for compiling JS, the ecmascript extension should also support JSX files. Even Facebook is using Babel to compile JSX now. I don't see any reason why it should be on a different package for Meteor 1.2. Any reason why this is not supported?

### DIFF
--- a/packages/ecmascript/package.js
+++ b/packages/ecmascript/package.js
@@ -1,7 +1,7 @@
 Package.describe({
   name: 'ecmascript',
   version: '0.1.3-plugins.5',
-  summary: 'Compiler plugin that supports ES2015+ in all .js files',
+  summary: 'Compiler plugin that supports ES2015+ in all .js and .jsx files',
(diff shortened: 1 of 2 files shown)


eXon:
I am using the JSX package with Meteor 1.2 right now and it is slow as hell to compile. I believe ecmascript is faster because it cache files that doesn't change. Is it right?

stubailo:
The new JSX package in Meteor 1.2 will internally use the same mechanism as the `ecmascript` package. I don't see any reason to couple these packages together, because it would be super reasonable for someone to want to use JSX without ecmascript or vice versa. Also, we can introduce react-specific optimizations into JSX that we wouldn't want to add to the ecmascript package.

@eXon we haven't implemented the caching for JSX yet, you're right, but we will before the 1.2 release.

PR #5078, opened 2015-08-31
Explain how to change logged-in user on client

lorensr:
Those looking at the server-side `this.setUserId` docs will often want `Meteor.userId()` to update on the client as well, so I think it would be good to explain how to do that. 

On a separate note, the docs use the phrase `logged in user`. I believe it should have a hyphen: `logged-in user`. Should I submit a PR for that as well?

https://en.wikipedia.org/wiki/Compound_modifier

### DIFF
--- a/docs/client/full-api/api/methods.md
+++ b/docs/client/full-api/api/methods.md
@@ -84,7 +84,7 @@ the [Meteor accounts system](#accounts_api) then this is handled for you.
 
 {{> autoApiBox "DDPCommon.MethodInvocation#setUserId"}}
 
-Call this function to change the currently logged in user on the
+Call this function to change the currently logged-in user on the
 connection that made this method call. This simply sets the value of
 `userId` for future method calls received on this connection. Pass
 `null` to log out the connection.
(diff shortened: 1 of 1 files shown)


stubailo:
I think this encourages a pretty bad anti-pattern: `Meteor.subscribe('dataFor', Meteor.userId());`

This is something one should never do; instead, you should use `this.userId` inside the publication, which will probably refresh itself once you call `setUserId` on the server. Have you actually encountered this issue with subscriptions not refreshing automatically?

lorensr:
Haha, sorry. That's not actually something I do – I was erroneously translating an example from Weldon's post, which I've removed from the PR.

https://dweldon.silvrback.com/impersonating-a-user

helfer:
@lorensr Actually, @tmeasday just told me that he's working on moving all the docs over to hexo (in the new-hexo-based-docs branch) and should be done with it in the next few days. Could you make a pull-request against that branch or wait until it's merged and then make the PR again?

benjamn:

Issue #5137, opened 2015-09-11
Error compiling {{#each a in (b)}}


wh0:
In Meteor release 1.2-rc.12, when compiling `{{#each a in (b)}}`, I get the following error:

```
=> Errors prevented startup:                  
   
   While processing files with templating (for target web.browser):

   /home/wh0/.meteor/packages/meteor-tool/.1.1.7-rc.12.dvasa3++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/lib/node_modules/fibers/future.js:395:16:
   Cannot call method 'replace' of undefined
   at Object.Future.wait
   (/home/wh0/.meteor/packages/meteor-tool/.1.1.7-rc.12.dvasa3++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/lib/node_modules/fibers/future.js:395:16)
   at CachingHtmlCompiler.processFilesForTarget (caching-compiler.js:312:12)
   at CachingHtmlCompiler.processFilesForTarget (caching-html-compiler.js:40:32)
```

[caching-compiler.js:312](https://github.com/meteor/meteor/blob/8754e3626f866173bf9a89693f3ecc1e3c72cd07/packages/caching-compiler/caching-compiler.js#L312)
[caching-html-compiler.js:40](https://github.com/meteor/meteor/blob/8754e3626f866173bf9a89693f3ecc1e3c72cd07/packages/caching-html-compiler/caching-html-compiler.js#L40)
(can't find `replace` on either of these lines)

`{{#each a in b}}` don't `{{#each (b)}}` cause the crash.

---

steps to reproduce:

```sh
meteor create --release 1.2-rc.12 asdf
cd asdf
rm asdf.css
rm asdf.js
cat <<EOF >asdf.html
<body>
  {{#each a in (b)}}
    Hello
  {{/each}}
</body>
EOF
meteor
```

wh0:
maybe it's this replace:
https://github.com/meteor/meteor/blob/8754e3626f866173bf9a89693f3ecc1e3c72cd07/packages/spacebars-compiler/codegen.js#L122

wh0:
it's not that replace. Here's the real stack trace

```
#0 toJSLiteral <runJavaScript-53>:269:12
#1 _.extend.codeGenPath <runJavaScript-55>:1066:36
#2 _.extend.codeGenTemplateTag <runJavaScript-55>:942:20
[...]

PR #5138, opened 2015-09-12
spacebars-compiler: handle EXPR in each-in

wh0:
#5137 

You can do `{{#each b c d=e}}{{/each}}`. Now you can also do `{{#each a in b c d=e}}{{/each}}`.


stubailo:
Looks pretty good! Will look in depth soon. Next time I would suggest linking to the PR from the issue, I didn't notice it until now.

yyx990803:
Cherry-picked into release-1.2 - there's a small issue when `codeGenInclusionData` returns null, which I fixed. Also added tests.

PR #5141, opened 2015-09-12
Add methods of the form BrowserPolicy.content.allow<ContentType>BlobUrl() to BrowserPolicy

brianlukoff:
This PR is for adding methods of the form BrowserPolicy.content.allow<ContentType>BlobUrl() to BrowserPolicy, to support the "blob:" CSP directive.


zol:
Merged. Thanks @brianlukoff !

PR #5146, opened 2015-09-13
Use Meteor.setTimeout to be able to use tests on the server

mitar:
I am using the same tests file to run tests also for our implementation of server-side autorun. This is necessary for things to be correctly bound.


stubailo:
Merged: 7e97b224441228cb1f07f0b2f4c48ab5ff2eb08a

Thanks!

PR #5161, opened 2015-09-15
Use fast PRNG to generate `_id` for documents created outside methods

avital:
This change harmonizes server document ID generation regardless of whether
it happens inside of a method or not, by using Alea in both cases.

This cuts time of inserting small documents outside of methods
on the server by over 30%, and more importantly makes it easier to be
confident in benchmarking numbers.

---

BACKGROUND

When calling `coll.insert()` on the server within methods, we use the Alea PRNG
(which is fast, can be seeded, and not cryptographically secure) to generate
the `_id` field for the newly created document (unless an `_id` field was
explicitly passed).

The reason we use Alea is so that we can seed the PRNG from the client, as to
ensure consistently chosen IDs for methods that create multiple documents and
run on both client and server.

Prior to this change, when calling `coll.insert()` on the server *not* inside
methods, we'd use Node's cryptographically secure `crypto.getRandomBytes()`
which is slower (due to allocating buffers that need to cross from JS
into native code).

With this change, we always use Alea when generating a document ID.

---

CRYPTOGRAPHICALLY SECURE IDS STILL AVAILABLE

If an app wants to guarantee using a cryptographically secure PRNG
when generating IDs, just generate IDs yourself:
`coll.insert({_id: Random.id(), ...})`.

`Random.id()` still uses a CSPRNG (unless you're on IE, or
on the server and not enough entropy has been collected, which is
basically never the case).

f you *want* the faster Alea algorithm, use `Random.fast.id()`
(The `Random.fast` object has all the same methods as on `Random`)

---

BENCHMARK RESULTS

Here are the measured times for inserting 5000 documents, before
and after this change (on my machine):

Benchmark                         | Before | After
--------------------------------- | ------ | ------
direct insert from `meteor shell` | 2179ms | 1520ms
a method called from a browser    | 1617ms | 1570ms
a method called from the server   | 1491ms | 1487ms
direct insert from the server     | 2272ms | 1445ms

(The benchmark can be found here:
https://github.com/avital/mongo-timing/blob/f32ea073b7bd242ef9294bd916327238c5f03ba2/benchmark2.sh)

### DIFF
--- a/packages/ddp-common/random_stream.js
+++ b/packages/ddp-common/random_stream.js
@@ -49,9 +49,11 @@ DDPCommon.RandomStream.get = function (scope, name) {
     name = "default";
   }
   if (!scope) {
-    // There was no scope passed in;
-    // the sequence won't actually be reproducible.
-    return Random;
+    // There was no scope passed in; the sequence won't actually be
+    // reproducible. but make it fast (and not cryptographically
+    // secure) anyways, so that the behavior is similar to what you'd
+    // get by passing in a scope.
+    return Random.fast;
   }
   var randomStream = scope.randomStream;
   if (!randomStream) {
--- a/packages/ddp-common/random_stream.js
+++ b/packages/ddp-common/random_stream.js
@@ -37,13 +37,14 @@ function randomToken() {
   return Random.hexString(20);
 };
 
-// Returns the random stream with the specified name, in the specified scope.
-// If scope is null (or otherwise falsey) then we will use Random, which will
-// give us as random numbers as possible, but won't produce the same
-// values across client and server.
-// However, scope will normally be the current DDP method invocation, so
-// we'll use the stream with the specified name, and we should get consistent
-// values on the client and server sides of a method call.
+// Returns the random stream with the specified name, in the specified
+// scope. If a scope is passed, then we use that to seed a (not
+// cryptographically secure) PRNG using the fast Alea algorithm.  If
+// scope is null (or otherwise falsey) then we use a generated seed.
+//
+// However, scope will normally be the current DDP method invocation,
+// so we'll use the stream with the specified name, and we should get
+// consistent values on the client and server sides of a method call.
 DDPCommon.RandomStream.get = function (scope, name) {
   if (!name) {
     name = "default";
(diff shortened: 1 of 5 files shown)


avital:
@glasser, would you have time to review this change?

glasser:
Sure.

Initial impression: call it `Random.insecure` instead of `Random.fast`.  There's a reason it's not the default, so we should give it a name that emphasizes why you should think carefully about using it instead of why you should be excited about using it.

packages/ddp-common/random_stream.js
```diff
@@ -49,9 +49,11 @@ DDPCommon.RandomStream.get = function (scope, name) {
     name = "default";
   }
   if (!scope) {
-    // There was no scope passed in;
-    // the sequence won't actually be reproducible.
-    return Random;
+    // There was no scope passed in; the sequence won't actually be
```
glasser: The comment above this method is now wrong.

glasser:
This requires a doc or history.md note to make clear that anyone relying on the behavior for security needs the explicit Random.id() now.

avital:
Thanks @glasser, I'll make those changes. Did you go through the entire change or should I expect potentially more comments?

glasser:
I did a full first pass

packages/ddp-common/random_stream.js
avital: updated

avital:
and added History.md note. doesn't look like the docs need changing.

@glasser i think this is ready to be merged

glasser:
:shipit:

avital:
Merged in d1ae8f25be68c2f2d79a68295ebd5576ed27b5fb

mitar:
How does one make so that secure IDs are generated also in the simulation mode?

Can secure IDs be enabled for the whole collection with some options to the collection constructor? (Like when we can choose between string and ObjectID IDs? Maybe there could also be "secure string" ID options there.)

avital:
The problem is that Meteor doesn't ship with an algorithm that's both
cryptographically secure and can be seeded (which is necessary for latency
compensated methods that generate multiple documents)

mitar:
Hm, I thought that this was how it was before?

Is there an extension point where community could then develop packages for this?

glasser:
(it's less "Meteor doesn't ship with an algorithm that's both cryptographically secure and can be seeded" and more "being cryptographically secure literally means it can't be seeded (predictable)")

mitar:
Oh, I get it. The text in history doc was a bit hard to understand. :-) OK. So this deals only with IDs outside of a method call, which are now made in the same way as inside a method call. So it does not change the behavior for method calls and situation when you want the same ID between client and server.

avital:
Yes Mitar that's right.

Issue #5181, opened 2015-09-15
{{#each a in b}} stops reactive updates from outside


wh0:
In Meteor 1.2-rc.15, the `each..in` template tag doesn't allow reactive references to values outside the tag.

```handlebars
{{#with getOrder}}
  <p>Give me {{qty}} of everything:</p>
  <ul>
    {{#each item in items}}
      <li>{{qty}}x {{item}}</li>
    {{/each}}
  </ul>
{{/with}}
```
(suppose `getOrder` is a reactive helper that returns something like `{qty: 5, items: ['pancakes']}`)

When `qty` changes, it'll update partially, to show:

> Give me 6 of overything:
> - 5x pancakes

test app: https://github.com/wh0/meteor-bug5181

wh0:
cc @mitar

wh0:
https://github.com/meteor/meteor/blob/81b2d3e6eecb5cedd9a14696824300acfdf3536c/packages/blaze/builtins.js#L178 basically snapshots the immediately surrounding context when each item view is created

wh0:
added pr #5182

stubailo:
Fixed, should come out in the next RC (17)

stubailo:
Released now as part of `1.2-rc.17`!

PR #5182, opened 2015-09-15
blaze: don't snapshot data context in each..in

wh0:
#5181

When using the new-style `each..in` tag, the data context in items is now reactive. Instead of creating a `Blaze.With` with a snapshot of the surrounding context, it now creates a plain old `Blaze.View`. Accessing the data context inside the item view now traverses out to reactively access whatever data context is outside.


yyx990803:
Thanks! This has been merged into the 1.2 release branch and will be merged back into devel after the release.

PR #5190, opened 2015-09-16
Convert to HTML error page

stubailo:
One big reason we need this is that the text format doesn't work on Windows
because of line endings. We could just replace the line endings based on
platform, but this seemed better overall.

### DIFF
--- a/tools/runners/run-proxy.js
+++ b/tools/runners/run-proxy.js
@@ -172,40 +172,7 @@ _.extend(Proxy.prototype, {
 
       var c = self.httpQueue.shift();
       if (self.mode === "errorpage") {
-        // XXX serve an app that shows the logs nicely and that also
-        // knows how to reload when the server comes back up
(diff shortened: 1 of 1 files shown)


tools/runners/run-proxy.js
```diff
@@ -174,13 +174,37 @@ _.extend(Proxy.prototype, {
...
+    </style>
+  </head>
+
+  <body>
+    <h3>Your app is crashing. Here's the latest log:</h3>
+
+    <pre>`);
 
         _.each(runLog.getLog(), function (item) {
           c.res.write(item.message + "\n");
```
glasser: html-escape (&lt; and friends), I think

stubailo: Darn! Missed that. Do we have any utilities like this in Meteor somewhere that I could use?

glasser: There's one in blaze preamble.js.  Dunno about tools.

stubailo:
Merged.

Issue #3799, opened 2015-02-23
meteor/cordova/ios not allowed to load local resource on IOS


dotnetwise:
We are using the file transfer plugin to download a file to 
`file:///var/mobile/Containers/Data/Application/6BB3B78C-0BAA-42FF-93F0-C70B80B86944/tmp/cdv_photo_003.jpg`

However, then trying to access that path with an `<img` tag will give us security errors:
`Not allowed to load local resource: file:///var/mobile/Containers/Data/Application/6BB3B78C-0BAA-42FF-93F0-C70B80B86944/tmp/cdv_photo_003.jpg`

This won't happen in a normal `cordova` app since it starts via `file:///` protocol

However `meteor` breaks this beacuse the app will start on `http://meteor.local`

So, how to give access to the app's `file:///` sandbox files?

Slava:
You are right, there is no good way to do it right now.

dotnetwise:
So, why enforce `http://meteor.local` then?

I think there should be an option to be on the file protocol so that would be the same as normal cordova app.

What about `file:///meteor.local` to be "served" the same way as `http://meteor.local`
This way would also fix the `android 2.3` buggy `http://meteor.local` and would also keep the same functionality/behavior as for a normal cordova (meteorless) app.

What do you think?

Slava:
The environment on `file://` urls doesn't match the environment you get in the browser on a regular website. A lot of this has to do with cross-origin requests, single-page routing, hash fragments and problems coming out of this approach.

dotnetwise:
Yes, but none of them are a problem in a normal cordova app.

And if we are migrating from an existing app towards meteor then we will have a broken app, just because the app won't run in a file:/// environment anymore.

So, at least give us the option in the `mobile-config.js` to specify the `http://meteor.local` or `file:///meteor.local` or anything else.

ddresch:
What exactly does this mean? It's not possible to access downloaded files in any way? Are there any solutions or practices how to download and store larger assets like sounds etc. AND access these later on? 

I mean this is not an rare use case, it's pretty common I guess. `appcache` is probably not the best solution to build on if you handle bigger files, I got no experience with it so this is more a question.

glasser:
Just closed #3229 as duplicate of this. It contains a good reproduction:  https://github.com/jperl/repro-meteor-3229

glasser:
https://github.com/phonegap/phonegap-app-developer/issues/134
[...]

PR #5203, opened 2015-09-18
Ability to expose additional path over http://meteor.local/<prefix> on Cordova

wojtkowiak:
This is a solution for problems mentioned here https://github.com/meteor/meteor/issues/3799
@martijnwalraven: I know you are planning on rewriting the whole thing but take a look and let me know if this could maybe be accepted at least as a solution for now. If you have a different approach to this problem also let me know. This is just something I made for my team and we are already using this in some apps for like a half a year now. If this does not fit your plans that is ok, maybe at least you will be able to use some of this code :smile: 
The crucial part is here: https://github.com/meteor/com.meteor.cordova-update/pull/6

Code lacks tests but I am having a problem with `meteor test-packages --ios-device` - whenever I change anything in the test meteor goes into endless loop saying that the client was modified x1560 
times and goes up every second. No idea currently, maybe I should report it as bug.

Api is described in methods comments.

Demo:
1. checkout https://github.com/wojtkowiak/test
2. checkout https://github.com/wojtkowiak/com.meteor.cordova-update/tree/additional-data-path
3. checkout https://github.com/wojtkowiak/meteor/tree/additional-path
4. on line https://github.com/wojtkowiak/meteor/blob/additional-path/packages/webapp/package.js#L22 change the tarball url to file:///path/to/checkouted/com.meteor.cordova-update
5. run the app from checkout

### DIFF
--- a/packages/meteor/cordova_environment.js
+++ b/packages/meteor/cordova_environment.js
@@ -6,3 +6,94 @@
  */
 Meteor.isCordova = true;
 
+Meteor.Cordova = Meteor.Cordova || {};
+
+Meteor.Cordova._additionalDataPath = null;
+Meteor.Cordova._additionalDataUrlPrefix = 'data';
+
+Meteor.Cordova._trimFileProtocol = function(path) {
+    var fileProtocol = 'file://';
+
+    if (path.substr(0, fileProtocol.length) === fileProtocol)
+        path = path.substr(fileProtocol.length);
+
+    return path;
+};
+
+/**
+ * @memberOf Meteor
+ * @summary Sets an additional path on the mobile device that will be available in the Cordova environment using the url http://meteor.local/data
+ * @locus Client
+ * @param {String} [path] An absolute, valid path to a directory on the mobile device
+ */
+Meteor.Cordova.setAdditionalDataPath = function(path) {
+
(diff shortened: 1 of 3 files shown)


martijnwalraven:
Thanks! I'm indeed in the process of rewriting the plugins, so I don't think this is a good time to merge a pull request. I will take a good look at your approach though, because I agree exposing additional paths is probably the way to go. I'm planning on releasing a preview of the new plugins soon, hopefully sometime next month.

The `meteor test-packages --ios-device` bug does seem like something worth reporting, especially if you have a good reproduction. Does this also happen on 1.2 RC?

wojtkowiak:
Well if we would decide already how the API should like and ensure that it will not change I guess this could be merged earlier. But that just my opinion. 
Anyway let me know if I can help in any way. If you already have your work somewhere I can even port my changes to your new solution. I'd really like to see my team using standard release of meteor instead of my fork :smile:

About the error I will prepare a repro and submit it.

wojtkowiak:
Just a reminder, if I can help in any way let me know.

webwurst:
Any news on this? :)

martijnwalraven:
It's not in the first Meteor 1.3 Cordova beta yet, but it is on the list of features to work on.

stubailo:
This is going to be a first-class feature in the new Meteor cordova thing.

PR #5212, opened 2015-09-19
Extract handler result publish code

mitar:
This allows easier extending by packages.

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -1035,6 +1035,10 @@ _.extend(Subscription.prototype, {
     if (self._isDeactivated())
       return;
 
+    self._publishHandlerResult(res);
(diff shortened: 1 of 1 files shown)


mitar:
cc @stubailo 

This would make it really easy for my server-side reactive publish package to not have to duplicate core code. :-)

stubailo:
So the idea is that your package runs the handler reactively and uses the factored out publish result function? Makes sense to me.

mitar:
Yes. It should not change anything for you. But if we get this out in 1.2, then it will be great to release out reactive server side publish shortly afterwards. :-)

stubailo:
Merged: 60eaf91624831bee84f32b09aed66b881cd9b87f

Thanks!

Sorry I couldn't get it in before 1.2.

PR #5218, opened 2015-09-19
Remove template tags

shadedprofit:
At the beginning and end of the **quick-start.md**, there were **template tags** that didn't seem to belong with the content.  I removed them to help make the document less confusing.


Slava:
the document is compiled to a Spacebars template and is used in the docs. This is why there are tags. The document is not intended to be a stand-alone markdown file.

stubailo:
Yup!

Issue #2232, opened 2014-06-12
Need a consistent and cross-browser approach to logging caught errors


alanshaw:
**_1 Upvote**_ I can't debug my app in Firefox. The error messages are really hard to decipher:
![screen shot 2014-06-12 at 09 22 22](https://cloud.githubusercontent.com/assets/152863/3255430/b96fdcea-f20a-11e3-82c5-bf5ae51ee6db.png)

Same exception in Chrome:
![screen shot 2014-06-12 at 09 23 36](https://cloud.githubusercontent.com/assets/152863/3255448/ee7a44fc-f20a-11e3-84b8-248a38082d63.png)

It seems as though the error message isn't being logged.

MattiSG:
Duplicate of #1211?

ffxsam:
I don't understand why the error shows up so differently in Firefox. How are errors being dumped? Are these not using `throw Error` or `console.error`? Should Meteor be using a more standard way of error reporting?

lalomartins:
All right, after about two days trying my hand at this… I got a much better situation on Firefox (without compromising Chrome). You can see my work on https://github.com/lalomartins/meteor/tree/error-improvements, and a test project at https://github.com/lalomartins/meteor-test-errors.

It's by no means a complete fix for this, especially not as described by @glasser — I touched tracker and blaze error reporting, there are still heaps more, so not “consistent”. A debugging package might make sense here, or at least a proper debugging API in the meteor package, but I'm not going to venture designing that on my own; maybe I'll start a hackpad.

This also fixes Firefox complaining about assigning to __proto__ as described in #4472.

lalomartins:
Ultimately (and somehow it seems nobody in the previous discussion noticed this, including myself), the main problem is that in Firefox we don't have the actual exception text (type + message), so it's hard to debug what actually happened. We were focusing on how ugly the stack looks, and didn't notice this.

That is definitely a Meteor bug; Meteor is counting on the stack property containing the type + message data, which it does in Chrome, but is not a standard.

(Incidentally, my branch does fix that, at least for Meteor errors.)

Diggsey:
Nice work @lalomartins - with these changes I will finally get zero warnings in my meteor app - a first as long as I've used meteor...

alanshaw:
Thank you @lalomartins

ffxsam:
Thanks! I've never understood why the errors are clean in Chrome but not Firefox. I thought we were supposed to be striving towards completely cross-platform solutions?

[...]

PR #5221, opened 2015-09-21
Error improvements

lalomartins:
Partial attempt at https://github.com/meteor/meteor/issues/2232, also removing some browser warnings by eliminating `__proto__` manipulation

### DIFF
--- a/packages/blaze/exceptions.js
+++ b/packages/blaze/exceptions.js
@@ -31,9 +31,10 @@ Blaze._reportException = function (e, msg) {
   if (! debugFunc)
     // adapted from Tracker
     debugFunc = function () {
-      return (typeof Meteor !== "undefined" ? Meteor._debug :
-              ((typeof console !== "undefined") && console.log ? console.log :
-               function () {}));
+      return (typeof Meteor !== "undefined" ? Meteor._error :
+              ((typeof console !== "undefined") &&
+               (console.error ? console.error : console.log ? console.log :
+               function () {})));
     };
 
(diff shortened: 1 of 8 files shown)


stubailo:
This looks like the beginning of a good change, since error reporting is something that could be much improved in Meteor.

However, it probably won't be possible to accept this without a good set of tests ensuring that errors are printed correctly - then we can run them in all browsers through our testing system and make sure that we don't break it in the future.

Do you have any ideas about how to test this?

lalomartins:
For a more reliable test, we'd want to hook up webdriver and check the actual console output. I think it might be overkill in this case, though.

What's the level of isolation in tinytest? A quick look at the source indicates it's zero… hmm. The straightforward way to test these is to stub/replace console.log, but I can see that having all sorts of side effects.

On the other hand, maybe adding a test-friendly console object in tinytest itself could be a good idea? But that would be a separate project.

stubailo:
> The straightforward way to test these is to stub/replace console.log, but I can see that having all sorts of side effects.

I think that could be sufficient, but would that method work in IE8/9?

lalomartins:
I just had a chance to check it, and yes, it works on 8 and 9.

I'll try to set aside some time this weekend to work on this — both the tests and refreshing the branch.

lalomartins:
For the record — I'm not using Meteor enough ATM that I can justify the time investment, or in fact, that I'd remember how to do this. Next time I have a project, I'll refresh this branch and give the testing a go. Might be later this month, but it's not certain yet.

stubailo:
OK, going to close it for now then.

Issue #5250, opened 2015-09-23
crucial error in simple-todos-angular


sclausen:
adding ng-include to the body leads to a source simple-todos-angular.compiled.html.js which obviously is wrong:
```javascript
Meteor.startup(function() {
  document.body.setAttribute('ng-include', ''todos-list.ng.html'');
  document.body.setAttribute('ng-controller', 'TodosListCtrl');
});
```
I don't know where this code is generated, therefor I can only provide a hack around this problem:
change the attribute directive ng-include to an element directive.
``` html
<body ng-controller="TodosListCtrl">
  <ng-include src="'todos-list.ng.html'"></ng-include>
</body>
```

sclausen:
Okay, even after this, my whole todos-app doesn't really work without an error.
![simple-todos-angular](https://cloud.githubusercontent.com/assets/916076/10038987/0cd45a62-61ce-11e5-96ac-8dda8de60cf8.png)

gjr-qm:
I think this is the same bug as #5232 (single quote in body tag attribute value).

You could try escaping the quotes by putting a backslash in front of them:
`<body ng-include="\'todos-list.ng.html\'">`

sclausen:
Yeah, this would be another good hack to fix the error. And afterall, the example app is still not useable. There must be somthing wrong on a more basic level for the quotation thing and the blank white page without errors, but as I said, I can't find the source code responsible for this… :(

yyx990803:
Thanks for reporting the issue - @stubailo has pushed a fix, it will be out in the next patch release (soon!)

jkalbfeld:
\m/  I thought I was a complete moron until I read this thread.  Just wrote some kickass stuff in the earlier version of Meteor, stole the code for a new app I scaffolded today and it was completely borked.

stubailo:
#5232 is a duplicate, going to track there.

PR #5253, opened 2015-09-23
not working meteor-include replaced

sclausen:
As described in #5250 the simple-todos-angular example has two issues. First, the ng-include in the body tag doesn't work properly, as described in #5232. Workarounds have been discussed [here](https://github.com/Urigo/angular-meteor/issues/702) and second, the `<meteor-include src="loginButtons"></meteor-include>` doesn't work. As @gjr-qm described, the solution for the last issue would be as follows: add the package `urigo:angular-blaze-template` and replace 
``` html
<meteor-include src="loginButtons"></meteor-include>
```
with
``` html
<blaze-template name="loginButtons"></blaze-template>
```
After thins changes and if #5232 is fixed, the simple-todos-angular example will work.


rgoomar:
@Urigo Is this fix fine? The Angular tutorial is currently unusable and this fixes the problem.

Urigo:
@sclausen @rgoomar  thanks for the valuable feedback!
I have fixed and updated the tutorial.
Hopefully it will be deployed soon.

Urigo:
The updated tutorial is now deployed.
Thank you so much

avital:
Happy to see that this was resolved!

Issue #5246, opened 2015-09-22
1.2: How to enable @decorators in ecmascript (Babel) package


pbastowski:
I have lots of ES2015 code in my projects that relies on Decorators, babel options `{ stage: 0 }`. However, Meteor 1.2 `ecmascript` package does not enable `{ stage: 0 }` by default and generates compile errors

    While processing files with ecmascript (for target web.browser):
    client/app/app.es6.js:30: client/app/app.es6.js: Unexpected token (30:0)

when it encounters code such as in the screen shot below from **app.es6.js**. (See the `@` at line 30 character 0)? That's a decorator.

![image](https://cloud.githubusercontent.com/assets/5649024/10034587/11be0208-6192-11e5-9114-80187da30d96.png)

The code above works with Meteor 1.1 and my existing package `pbastowski:angular-babel`, but not with Meteor 1.2 and package `ecmascript`.

Is there a setting or a configuration file, like babel.json, where I can configure my own Babel settings?

BrianRosamilia:
The whitelist is hardcoded to the features MDG included in the release notes, it would seem.

You can whitelist it yourself here https://github.com/meteor/meteor/blob/devel/packages/ecmascript/ecmascript.js and see what happens

Push 'es7.decorators' into babelOptions.whitelist

I'm pretty shocked they didn't find a way to be permissive with es7 features, at least in the client.

Related :  even if you hack in async/await by changing the ecmascript package to whitelist it, it uses a custom compiler plugin that doesn't work properly right now.

Sanjo:
You need to copy the ecmascript and babel-compiler packages to your app packages folder and add the option here:

* https://github.com/meteor/meteor/blob/release/METEOR%401.2.0.1/packages/ecmascript/ecmascript.js#L6
* https://github.com/meteor/meteor/blob/release/METEOR%401.2.0.1/packages/babel-compiler/babel-compiler.js#L51

I'm not sure if the `stage: 0` option works. `babelOptions.whitelist.push('<FEATURE_NAME>'` definitely works.

pbastowski:
@BrianRosamilia, @Sanjo  Thanks guys. I saw that code yesterday also, but thought that perhaps "they" provided some way to set custom options. Alas, it would appear they didn't.

Ok, I found that it is babel-compiler.js, which needs to be modified. So, I pushed 'es7.decorators' and later 'es7.classProperties' to the whitelist, which fixes the Babel errors.

Then, I got the browser complaining about
````
[...]

PR #5260, opened 2015-09-23
Add missing babel helpers required for es7.decorators and es6.modules

pbastowski:
Added missing babel helpers required for es7.decorators and es6.modules (see also issue https://github.com/meteor/meteor/issues/5246).

The `ecmascript` Meteor 1.2 package supports custom whitelisting of various Babel transformers through the use of the `.babelrc` file, placed in the project root directory. This is standard Babel functionality, please read https://babeljs.io/docs/usage/babelrc/.

However, some of the transformers require in-browser runtime support through extra functions called Babel helpers, which are defined on the `babelHelpers` object in the `babel-runtime.js` file of the `babel-runtime` package. 

**ES7 decorators** and **ES6 modules** require the following helper functions, which are not included as standard: `createDecoratedClass`, `defineDecoratedPropertyDescriptor` and `interopRequireDefault`. This PR adds these helper functions to the `babel-runtime` package.

`.babelrc` must be formatted as in the example below.
```json
{
    "whitelist": [
        "es7.decorators",
        "es7.classProperties",
        "es7.exportExtensions",
        "es7.comprehensions",
        "es6.modules"
    ]
}
````

### DIFF
--- a/packages/babel-compiler/babel.js
+++ b/packages/babel-compiler/babel.js
@@ -1,7 +1,7 @@
 var meteorBabel = Npm.require('meteor-babel');
 
-// Read in user config from optional babel.json file, which must be located 
-// in the project root directory. babel.json must be formatted as in the 
+// Read in user config from optional .babelrc file, which must be located 
+// in the project root directory. .babelrc must be formatted as in the 
 // example below.
 //{
 //    "whitelist": [
@@ -18,7 +18,7 @@ var babelUserConfig = function () {
   var path = Npm.require('path');
 
   var appdir = process.env.PWD || process.cwd();
-  var babelOptionsFilePath = path.join(appdir, 'babel.json');
+  var babelOptionsFilePath = path.join(appdir, '.babelrc');
 
   if (fs.existsSync(babelOptionsFilePath)) {
     return JSON.parse(fs.readFileSync(babelOptionsFilePath, {encoding: 'utf8'}));
@@ -50,7 +50,7 @@ function getDefaultOptions(extraFeatures) {
   // information about what the default options are.
   var options = meteorBabel.getDefaultOptions(extraFeatures);
 
-  // Bring in user Babel config options from babel.json
+  // Bring in user Babel config options from .babelrc
   for (propName in babelUserConfig) {
     if (babelUserConfig.hasOwnProperty(propName)) {
       var prop = babelUserConfig[propName];
--- a/packages/babel-compiler/babel.js
+++ b/packages/babel-compiler/babel.js
@@ -1,5 +1,30 @@
 var meteorBabel = Npm.require('meteor-babel');
 
+// Read in user config from optional babel.json file, which must be located 
+// in the project root directory. babel.json must be formatted as in the 
+// example below.
+//{
+//    "whitelist": [
+//        "es7.decorators",
+//        "es7.classProperties",
+//        "es7.exportExtensions",
+//        "es7.comprehensions",
+//        "es6.modules"
+//    ],
+//    "stage": 0
+//}
+var babelUserConfig = function () {
+  var fs = Npm.require('fs');
+  var path = Npm.require('path');
+
+  var appdir = process.env.PWD || process.cwd();
+  var babelOptionsFilePath = path.join(appdir, 'babel.json');
+
+  if (fs.existsSync(babelOptionsFilePath)) {
+    return JSON.parse(fs.readFileSync(babelOptionsFilePath, {encoding: 'utf8'}));
+  }
+}();
+
 function validateExtraFeatures(extraFeatures) {
   if (extraFeatures) {
     check(extraFeatures, {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rclai** [comment]
Huh? Modules? Is this all it would take to get imports to work in Meteor?

**Comment by pbastowski** [comment]
No it isn't :) But this at least transpiles the "import" keyword to an equivalent require call. You still need to supply the require library of some kind. I have my own `pbastowski:require` that does the job for my apps.

**Comment by rclai** [comment]
Oh I see.

**Comment by pbastowski** [comment]
That last commit's title was meant to say "Fixed bad createDecoratedClass function definition".

**Comment by Mistereo** [comment]
Maybe it's better to use `.babelrc` (https://babeljs.io/docs/usage/babelrc/) instead of `babel.json`?

**Comment by pbastowski** [comment]
Interesting that you mention this, as I also read this same page earlier today and was thinking about updating this PR. 

So, I think yes is the correct answer.

**Comment by pbastowski** [comment]
While I was preparing the updated PR, which renames `babel.json` to `.babelrc`, I found something quite exciting. `.babelrc` is already implemented and provides the exact same functionality as my `babel.json` does!

So, no need to implement `babel.json`. I removed that functionality from this PR completely.

So, the only thing that was missing to enable the support for ES7 decorators and `import .... from` were the three babelHelpers functions that I added to `babel-runtime.js`. These were apparently removed on purpose (see README.md in the babel-runtime package), because they were described in Babel documentation as "experimental". Experimental as they may be, they are what we use in Angular2, the angular2-now library and TypeScript implements them too. 

The point is, we need ES7 decorators as part of Babel and this PR adds the babelHelpers functions that are needed to support them.



So, this PR now consists only of the additions to the `babel-runtime.js` file.

**Comment by pbastowski** [comment]
OK, this is the final nail in the coffin for this PR, which I will close after some of you good people help me verify my findings. below:

- As I mentioned in my previous posts above, `.babelrc` allows us to whitelist extra Babel transformers
- Some transformers require specific helper functions to be added to the `babalHelpers` object
- I just discovered that the `babelHelpers` object is visible on `window`, so, we can add to it whatever functions we wish and desire.

So, create a file `client/lib/babel-helpers.js` in your project and add the following to it:
````javascript
// Provides support for es7.decorators
window.babelHelpers.defineDecoratedPropertyDescriptor = function (target, key, descriptors) {
    var _descriptor = descriptors[key];
    if (!_descriptor) return;
    var descriptor = {};

    for (var _key in _descriptor) descriptor[_key] = _descriptor[_key];

    descriptor.value = descriptor.initializer ? descriptor.initializer.call(target) : undefined;
    Object.defineProperty(target, key, descriptor);
};

// Provides support for es7.decorators
window.babelHelpers.createDecoratedClass = (function () {
    function defineProperties(target, descriptors, initializers) {
        for (var i = 0; i < descriptors.length; i++) {
            var descriptor = descriptors[i];
            var decorators = descriptor.decorators;
            var key = descriptor.key;
            delete descriptor.key;
            delete descriptor.decorators;
            descriptor.enumerable = descriptor.enumerable || false;
            descriptor.configurable = true;
            if ('value' in descriptor || descriptor.initializer) descriptor.writable = true;
            if (decorators) {
                for (var f = 0; f < decorators.length; f++) {
                    var decorator = decorators[f];
                    if (typeof decorator === 'function') { descriptor = decorator(target, key, descriptor) || descriptor; } else { throw new TypeError('The decorator for method ' + descriptor.key + ' is of the invalid type ' + typeof decorator); }
                }
                if (descriptor.initializer !== undefined) {
                    initializers[key] = descriptor;
                    continue;
                }
            }
            Object.defineProperty(target, key, descriptor);
        }
    }

    return function (Constructor, protoProps, staticProps, protoInitializers, staticInitializers) {
        if (protoProps) defineProperties(Constructor.prototype, protoProps, protoInitializers);
        if (staticProps) defineProperties(Constructor, staticProps, staticInitializers);
        return Constructor;
    };
})();

window.babelHelpers.interopRequireDefault = function (obj) {
    return obj && obj.__esModule ? obj : {
        "default": obj
    };
};
````

Now, create the file `.babelrc` in the root of your project and add the following to it:
```javascript
{
    "whitelist": [
        "es7.decorators",
        "es7.classProperties",
        "es7.exportExtensions",
        "es7.comprehensions",
        "es6.modules"
    ]
}
```

Now, your application that uses the `ecmascript` package and uses ES7 decorators, possibly because it uses the `pbastowski:angular2-now` package, should process both decorators and `import .... from` statements as expected and previously accomplished with `grigio:babel` and/or my very own `pbastowski:angular-babel` packages.

Could someone please let me know if the above worked for them?

Thanks.

pbastowski:
Closed, because it is not required as per my explanation above.

stubailo:
OK, sounds good! I think we'll be improving our ES2015+ support over time anyway, hopefully it meets your needs!

pbastowski:
@stubailo  I create a package `pbastowski:ecmascript-extras`, which adds the features I mentioned above on top of the existing `ecmascript` package.

Issue #689, opened 2013-02-15
[bug] Can't update meteor behind a proxy on Linux!


aadrian:
Was trying to update meteor from 0.5.4 with:

> meteor update

behind a proxy, but it seems to give all the time:

> "Failed to download manifest."

This is quite unexpected since the system is setup correctly, wget and curl work too, and "npm", "bower" and all other dependency managers work without problems too behind the proxy.

Thanks in advance.

daxpablo:
That's what worked for me:

export HTTP_PROXY="http://user:password@1.2.3.4:5678" && curl https://install.meteor.com/ | sh

kenorb:
Same here.

```
$ meteor update
Failed to download manifest.
$ curl https://install.meteor.com
curl: (35) SSL peer handshake failed, the server most likely requires a client certificate to connect
```

PR #5289, opened 2015-09-25
(minimongo) Fix pull modifier not excepting null values

DAB0mB:
Originally created because of [angular-meteor issue #689](https://github.com/Urigo/angular-meteor/issues/689). From now on the $pull modifier would except null values.


yyx990803:
Thank you!

PR #5300, opened 2015-09-27
Refactor duplicated full adder function expression

kencheung:


stubailo:
Seems legit to me; @dgreensp can you verify?

dgreensp:
These are two different class constructors that happen to take three arguments and assign them `this.a`, `this.b`, and `this.c`.  Although they're verbose, I don't think we should abstract across them.

PR #5345, opened 2015-10-01
Refactor duplicated function expression for greeting

kencheung:


benjamn:
This looks good to me, but I have to say this would be a lovely application of ECMAScript 2015 [template strings](http://www.2ality.com/2015/01/template-strings-html.html).

Would you be willing to add `api.use("ecmascript")` to the `Package.onUse` callback in `meteor/packages/accounts-password/package.js`, and then use a multi-line template string?

Something like this:
```js
      return `${greeting}

${welcomeMsg}, simply...
Thanks.
`;
```

kencheung:
I have made the changes as suggested. Please advice if I misunderstand anything. Thank you.

packages/accounts-password/email_templates.js
```diff
@@ -2,6 +2,23 @@
...
+
+function greet(welcomeMsg) {
+  return function(user, url) {
+      var greeting = (user.profile && user.profile.name) ?
+            ("Hello " + user.profile.name + ",") : "Hello,";
+      return `${greeting}
+
+${welcomeMsg}, simply click the link below.
+
+url
```
benjamn: `${url}`?

kencheung: Fixed. Thank you.

benjamn:
Perfect, thanks!

Issue #5363, opened 2015-10-04
Maybe standard-minifiers shouldn't re-minify .min.js files?


jedwards1211:
I'd be happy to make a PR, just want someone to acknowledge that this is a desired feature before I go to the effort

Those of us who are using Webpack are discovering that it takes a long time for Meteor to minify Webpack bundles.  Since we can minify more quickly with Webpack's UglifyJS plugin (since it operates on modules individually), it would be nice if we could just output bundles as `server.bundle.min.js` and Meteor wouldn't run them through UglifyJS again.

stubailo:
Yeah, I think that could be very reasonable - I can't think of any reason someone would name their file with `.min.js` if it wasn't minified already. I'll accept the PR if it's a minimal change!

jedwards1211:
Cool!  I'll make the PR soon

jedwards1211:
Do you know of any good way I could add the option to disable bypassing `.min.js` files?  (Besides an environment variable), just in case someone needs it?  And should I print a message about "assuming yourfile.min.js is already minified" on the off chance it would be a surprise to some dev?

laosb:
Fixed in #6986 !

PR #5378, opened 2015-10-07
don't re-minify .min.js files in standard-minifiers

jedwards1211:
fix https://github.com/meteor/meteor/issues/5363

### DIFF
--- a/packages/standard-minifiers/plugin/minify-js.js
+++ b/packages/standard-minifiers/plugin/minify-js.js
@@ -40,10 +40,12 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
         path: file.getPathInBundle()
       });
     }
-    allJs += UglifyJSMinify(file.getContentsAsString(), minifyOptions).code;
-    allJs += '\n\n';
+    else {
+      allJs += UglifyJSMinify(file.getContentsAsString(), minifyOptions).code;
+      allJs += '\n\n';
 
-    Plugin.nudge();
+      Plugin.nudge();
+    }
   });
 
   if (files.length) {
--- a/packages/standard-minifiers/plugin/minify-js.js
+++ b/packages/standard-minifiers/plugin/minify-js.js
@@ -34,18 +34,12 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
   var allJs = '';
   files.forEach(function (file) {
     if (/\.min\.js$/.test(file.getPathInBundle())) {
-      file.addJavaScript({
-        data: file.getContentsAsBuffer(),
-        sourceMap: file.getSourceMap(),
-        path: file.getPathInBundle()
-      });
-    }
-    else {
+      allJs += file.getContentsAsString();
+    } else {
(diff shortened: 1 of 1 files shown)


sebakerckhof:
Won't this still call the minifier?

jedwards1211:
@sebakerckhof Whoops, yup! I forgot to copy over an else block from my package.

jedwards1211:
Okay, now it matches how I was doing it in my own package.

jedwards1211:
I investigated a little bit how I could write an automated test for this, but it didn't seem very obvious; if anything tests that `standard-minifiers` works, I didn't find it.

sebakerckhof:
I could be mistaken, but I think it is still incorrect:
1) You'll end up with multiple JS files instead of one, but you probably still only want one http request
2) Js files will get out of order, which might cause errors
2) If the first JS file is a .min.js file, it will be overwritten with the allJS.

I think you need to add everything to the allJS variable, but only skip the minifying step for already minified files.

jedwards1211:
@sebakerckhof ah, yes, good points, you are clearly more familiar with the build process than I am.  Let me update it

jedwards1211:
Does everything look good now?

stubailo:
This definitely needs a test. The tests are currently written in a super custom test framework called "self-test", here: https://github.com/meteor/meteor/blob/devel/tools/tests/standard-minification.js

If you can write the test, I'll accept this immediately - otherwise it will have to wait until someone writes the test.

jedwards1211:
Okay.  Is there any way to avoid this kind of thing?
```
    run.waitSecs(60);
```
I would have thought that `run.match('Client modified -- refreshing');` waits until that gets printed to the console and the wait times are unnecessary.  No?

stubailo:
@jedwards1211 that doesn't actually wait 60 seconds, it actually just extends the timeout of the next command since `run.match` has a built-in timeout.

I'll admit it's pretty confusingly named.

jedwards1211:
Ahhhh, very good.  If there are any guides to self-test anywhere or if you have tips on where I can find the self-test API in the code, please let me know.

jedwards1211:
I just realized: you can use the `.es5.js` extension to work around the hanging on large bundles issue, but then you can't use `.min.js`.  What we are creating here is **convention without configuration**.

If we really want to serve devs best, we should also make a way to turn these things on and off for any arbitrary files individually.  ESLint knows this; they let you use comments like
```js
/* eslint-disable no-console */
console.log('freedom!');
/* eslint-enable no-console */
```

We could adopt the same approach here, scanning each file for a comment:
```js
/* meteor-disable standard-minifiers */
```

stubailo:
I see what you are saying! It would be reasonable to have a file that shouldn't be compiled or minified. In this particular case, I can't think of any reason you would want a minified file to be compiled?

jedwards1211:
That's a good point.  So maybe a second PR to disable compilation of .min.js files?

stubailo:
Yeah for now that could be a good workaround.

benjamn:

PR #5386, opened 2015-10-07
Update simple-todos.js

anand-vishal:
It works without *input*, also in the event maps documentation (http://docs.meteor.com/#/full/eventmaps) the various formats don't have the third part like in the example above. I have just started learning this, so sorry if I am missing something.


stubailo:
Unfortunately updating this app is not that simple, since it's a part of a multi-step tutorial. I'd suggest filing an issue here: https://github.com/meteor/tutorials

PR #5398, opened 2015-10-09
adding check for NODE_ENV before overriding in 'meteor run'

em0ney:
Very simple improvement here.

When running `meteor` or `meteor run`, the NODE_ENV is set to 'development'.  In some cases when we are wanting to test production-like behaviour we try to override this variable locally, but are never able to.  Found this line and adding a check to wrap around it.


benjamn:

PR #5401, opened 2015-10-09
Fix capitalization

theavish:
Fixed capitalization on lines 8-10


zimme:
I think the lower case is intended as it reads as a full sentence, but I might be wrong?

stubailo:
Yeah. I don't think this was an accident.

PR #5414, opened 2015-10-12
Allow PORT as a string if it matches named pipe pattern. Fixes #4413

### LINKED ISSUE #4413: Ports are not always numbers
In Webapp.js, parseInt is used on the env.PORT. The env.PORT need not always be a number, but can be a named pipe.

https://github.com/meteor/meteor/blob/40ec4760f049b3a312b6b228aa4bf08f91f241aa/packages/webapp/webapp_server.js#L742

This issue can be hit when running Meteor with IIS on Windows, which uses a named pipe to feed the connections to Node. Removing the parseInt function call works fine for named pipes and ports on local tests. express-generator ships with a port parsing function which is more robust - maybe a better option than dropping the parseInt entirely.

ericterpstra:
Windows Server uses 'named pipes' instead of a port number. This allows named pipes to be used according to this pattern: [Pipe Names](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365783(v=vs.85).aspx).  
Fixes https://github.com/meteor/meteor/issues/4413

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -439,7 +439,7 @@ var getUrlPrefixForArch = function (arch) {
     '' : '/' + '__' + arch.replace(/^web\./, '');
 };
 
-// parse port to see if its a named pipe. If so, return as-is (String), otherwise return as Int
+// parse port to see if its a Windows Server style named pipe. If so, return as-is (String), otherwise return as Int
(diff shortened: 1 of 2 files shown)


stubailo:
Looks great to me!

stubailo:
Merged: 412e502fa9ec9e794b05e7faf2d071cf818a48fc

Issue #5372, opened 2015-10-06
meteor/stylus crashes (surprising path) when trying to import nib


swestgate:
**_3 Upvotes**_ I've run into an issue where including `@import 'nib'` in my main .styl file causes meteor to crash (detailed error below). I suspect this may be related to the fact that I'm running Windows with two drives, C: holds the OS and my profile, but D: holds most program files and is where I develop my meteor projects. It looks like the path to nib's support files isn't properly accounting for the drive letter.

Note: the error below chokes on reset.styl because I have `global-reset()` right after the import statement. Without it I get the same error, but with the vendor.styl file instead.

```
=> Errors prevented startup:

   While processing files with stylus (for target web.browser):

   C:\Users\myuser\AppData\Local\.meteor\packages\meteor-tool\1.1.9\mt-os.windows.x86_32\tools\static-assets\server\mini-files.js:34:13:
   Surprising path:
   /Users/myuser/AppData/Local/.meteor/packages/stylus/2.511.0_2/plugin.compileStylusBatch.os/npm/compileStylusBatch/node_modules/nib/lib/nib/reset.styl
   at toDosPath
   (C:\Users\myuser\AppData\Local\.meteor\packages\meteor-tool\1.1.9\mt-os.windows.x86_32\tools\static-assets\server\mini-files.js:34:13)
   at Object.convertToOSPath
   (C:\Users\myuser\AppData\Local\.meteor\packages\meteor-tool\1.1.9\mt-os.windows.x86_32\tools\static-assets\server\mini-files.js:47:12)
   at Object.wrapper (C:\tools\fs\files.js:1310:25)
   at importer.readFile (plugin/compile-stylus.js:119:21)
   at Array.map (native)
   at processSourcemap (plugin/compile-stylus.js:139:52)
   at StylusCompiler.compileOneFile (plugin/compile-stylus.js:169:23)
   at multi-file-caching-compiler.js:115:45
   at multi-file-caching-compiler.js:144:68
   at
   C:\Users\myuser\AppData\Local\.meteor\packages\stylus\2.511.0_2\plugin.compileStylusBatch.os\npm\caching-compiler\node_modules\async\lib\async.js:182:20
   at replenish
   (C:\Users\myuser\AppData\Local\.meteor\packages\stylus\2.511.0_2\plugin.compileStylusBatch.os\npm\caching-compiler\node_modules\async\lib\async.js:317:21)
   at
[...]

PR #5418, opened 2015-10-13
Idea for Windows Stylus fix

stubailo:
Might fix issue #5372

I don't have the time to test this right now, but if someone could run it on their Windows computer and let me know that would be awesome!

Also, we should have automated tests for this and we should run them on Windows.


benjamn:

PR #5423, opened 2015-10-14
Use _.has for checking minimongo NO_CREATE_MODIFIERS

tim-smart:
In somes cases Object.prototype.$set could be definied by another framework etc.


stubailo:
> In somes cases Object.prototype.$set could be definied by another framework

Is there a particular case when this happens? I wonder what the actual issue is that you ran into.

tmeasday:

PR #5430, opened 2015-10-14
Fix issue #1201

okland:
Allow count apply skip limit in mongo driver.

Fix problem on server when calling count on server with limit - 
C.find(q, {limit: N}).count() -> 
        count can be > N
with this fix  C.find(q, {limit: N}).count(true) ->
        count can be <= N
Allow like the default mongo driver to send flag whether to applySkipLimit with the count


stubailo:
We would also need this to work in Minimongo, would you be good with adding it? Also, this should probably include a test.

okland:
The problem is that in MiniMongo package the count function always count only the num of objects that pass the query (with the limit ) , Like using the flag  - C.find(q, {limit: N}).count(true), i can make it like that by default also on the mongo driver, So they will behave in the same manner.

benjamn:

PR #5435, opened 2015-10-15
Google's OAuth credentials page does not permit trailing slashes. This patch resolves this.

JonTheNiceGuy:


packages/google/google_configure.js
```diff
@@ -1,7 +1,14 @@
 Template.configureLoginServiceDialogForGoogle.helpers({
   siteUrl: function () {
     return Meteor.absoluteUrl();
-  }
+  },
+  siteUrlNoTrail: function () {
+    if (Meteor.absoluteUrl().slice(-1) === '/') {
+      return Meteor.absoluteUrl().slice(0, 1);
```
stubailo: Should this be `slice(0, -1);` instead?

stubailo:
Great idea, but I'm not sure the code is correct.

JonTheNiceGuy:
Damn yes, you're right. I have to apologise, I don't typically do JavaScript, I spotted this issue in a project that is using meteor, and I thought I could cludge my way to fixing it. I don't even have a dev environment set up, as I don't have the first clue about tying meteor together to do auth. I probably should have just submitted a bug.

How do you want me to proceed? Close this down and file a bug, or try to suss it out in this pr?

stubailo:
If you can actually test the code, then a PR would be great, if you don't have time then an issue is fine too.

JonTheNiceGuy:
Closed, in favour of issue #5486.

PR #5448, opened 2015-10-16
Add buildMode option in compiler plugins

eXon:
For some compiler plugins (let's say Webpack :)), it is VERY important to know if we are building in production or development mode. It adds stuff on development for hot module replacement and start a dev server as an example.

By passing the buildMode option to the plugins, it will solve this problem. I badly need this. This is how people are doing it right now: `NODE_ENV=production meteor --production` to run in production. I would rather have only `meteor --production` and know we are running in production.

Let me know if I need to fix/add anything!

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -692,6 +692,7 @@ class Target {
   _runCompilerPlugins() {
     buildmessage.assertInJob();
     const processor = new compilerPluginModule.CompilerPluginProcessor({
+      buildMode: this.buildMode,
       unibuilds: this.unibuilds,
       arch: this.arch,
       isopackCache: this.isopackCache,
--- a/tools/isobuild/compiler-plugin.js
+++ b/tools/isobuild/compiler-plugin.js
@@ -67,6 +67,7 @@ const LINKER_CACHE = new LRU({
 
 exports.CompilerPluginProcessor = function (options) {
   var self = this;
+  self.buildMode = options.buildMode;
   self.unibuilds = options.unibuilds;
   self.arch = options.arch;
   self.isopackCache = options.isopackCache;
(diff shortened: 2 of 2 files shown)


eXon:
I don't know why the CI gave this error:

```
  2| Error: Can't kill running mongo (pid 8702).
  2|     at /tools/runners/run-mongo.js:310:11
  2|     at Array.forEach (native)
  2|     at Function._.each._.forEach (/home/ubuntu/meteor/dev_bundle/lib/node_modules/underscore/underscore.js:79:11)
  2|     at findMongoAndKillItDead (/tools/runners/run-mongo.js:283:5)
  2|     at launchOneMongoAndWaitForReadyForInitiate (/tools/runners/run-mongo.js:400:7)
  2|     at launchMongo (/tools/runners/run-mongo.js:656:7)
  2|     at [object Object]._.extend._startOrRestart (/tools/runners/run-mongo.js:756:19)
  2|     at [object Object]._.extend.start (/tools/runners/run-mongo.js:714:10)
  2|     at /tools/runners/run-all.js:168:26
package-tests: add packages to app ... 
```

Tests are running fine locally.

stubailo:
@dgreensp what do you think of this? Is there any reason we shouldn't do it this way?

Also I'm not a fan of the word `buildMode` - perhaps the public API should be `buildEnvironment` or just `environment` or something?

Sanjo:
:+1: for this addition.
The name buildMode is currently the name that it used for this option in the Meteor tool.

stubailo:
@Sanjo what things could we break if we renamed it?

Sanjo:
> what things could we break if we renamed it?

@stubailo Nothing. Just wanted to promote to keep this name consistent between the public API and internally in the Meteor tool. (I personally find the name buildMode accurate for this)

dgreensp:
I'm a fan of exposing it, but I can't think of a better name than buildMode.  The word environment means other things to me, like the set of environment variables.

Slava:
Note that the mode of development or production is passed into minifiers already. I also don't think the code of compiler plugins correctly caches the change of environment (because it is assumed to not matter).

eXon:
@Slava This might not be the case for most compilers but knowing if we are in production or not for others is a big deal for the Webpack compiler (probably not a lot). 

As an example, the compiler will add a few things to activate HMR (this allows to reload the change you made in realtime, without a page refresh). There is no way to wait at the minify stage to do this.

Right now I am using process.argv to guess but I think it would be more robust to have the real value from Meteor.

Slava:
@eXon I am just giving more info on the situation.

benjamn:

Issue #5427, opened 2015-10-14
Tests compilation


Sanjo:
TLDR: Compiling the tests into a separate build target would enable me (Velocity) to speed up the test execution a lot.

With our current mirror approach in Velocity, we run the Meteor tool separately for each mirror (= testing mode). On average you use three mirrors. One for Cucumber, Jasmine server tests and Jasmine client tests. That means that four separate Meteor tool instances run for the app.
With the current approach therefore a lot of time is spend on building the same app (four times).

The only difference between the mirrors is the set of tests that are included. Therefore we could build the app only once and then spawn the built app directly and load some tests based on an argument that we pass to the programs/server/main.js.

My sticking point right now is that the tests should get compiled, so that our users can us ES2015 and CoffeeScript. I tried to implement it, but have some hard time to get it fully working. So I would appreciate some help.

tmeasday:
Hi @Sanjo, thanks for this one, sorry it's taken a while to look at it.

Is this still something that you want for your non-velocity projects? I've been considering all the various ways Velocity integrates with the tool and ideas on how to make that better and I'm working on (what I hope is a) really simple proposal to get a better app testing story going forward. I'll post something to the forums soon I hope.

But let me know your latest thoughts on this PR.

Sanjo:
Yes. I still think the PR is a good idea. I enables me to write a fast implementation for server unit testing for sanjo:jasmine. The only bigger missing part right now for me.

tmeasday:
So sanjo:jasmine somehow runs a mirror using the pre-built test code without the use of velocity?

Sanjo:
Yeah, I would just spawn a modified version of `.meteor/local/build/programs/server/boot.js`. This doesn't have the overhead of the second Meteor build tool running.

abernix:
Closing as I think this is no longer relevant or needs to be re-visited with the newer testing setups.

PR #5457, opened 2015-10-17
Tests compilation

Sanjo:
As already stated in https://github.com/meteor/meteor/issues/5427, test compilation with the Meteor tool is an important missing piece for Velocity, that enables us to do performance optimizations and support ES2015 and CoffeeScript for all test frameworks.

With this pull-request the Meteor tool compiles tests also. Test files get the `isTest: true` flag in the program.json manifests and are not loaded by the Meteor app. Velocity can load the tests files when needed on demand.

Currently I'm looking for feedback on my approach.

__Still to do:__
* [ ] Test with `meteor build` / production build
* [ ] Add a selftest for this


Sanjo:
> did the CI tests catch this?

Yes. But I caught it myself 5 seconds AFTER I pushed the commit :D (like always when I do something to quick).

stubailo:
So the way I understand this design, it compiles all of the test files just the same as always, except then doesn't add them to the boilerplate for the app's HTML? Does it still compile the tests when building for production? Probably ideally it wouldn't include test files in the production bundle at all, right?

Sanjo:
> So the way I understand this design, it compiles all of the test files just the same as always, except then doesn't add them to the boilerplate for the app's HTML? Does it still compile the tests when building for production? Probably ideally it wouldn't include test files in the production bundle at all, right?

Exactly as you wrote. The tests should only be compiled in the development buildMode. I have a new commit nearly ready that will exclude the tests from the production build. I will finish it on this weekend.

Sanjo:
I just pushed the last thing that I have done some weeks ago. The tests indicate that it doesn't fully work.

PR #5465, opened 2015-10-18
Hook to alter service prov, query update or create

calvinfroedge:
This pull request removes the need for hacky workarounds (several packages do this) to merge social accounts. This PR allows the user to modify the query logic used to determine if a newly connected service should be merged with an existing service. An example follows below. I have currently monkey patched the Meteor source to add this hook:

```
Accounts.onUpdateUserFindExistingUserHooks = [
 	function(selector, serviceName, serviceData, options){
		_.each(selector, function(value, key){
			var obj = {};
			obj[key] = value;
			or.push(obj);
		});

		var email = null;

		var pushEmail = function(val){
			or.push({emails: {$elemMatch: {address: val}}});
		}

		if(serviceData.emails){
			serviceData.emails.map(function(email){
				var val = email.email || email.address;
				if(val && email.verified) pushEmail(val);
			});
		}

		var newSelector = {$or: or};
		return newSelector;
	} 
]
```

This checks if an email verified by a third party provider already appears in the user database. If so, it modifies the selector used to find an existing user:

```
var user = this.users.findOne(selector);
```

Meteor then updates the existing user record and finishes the login process. This is a significant improvement over the existing practice in many packages of using the `Accounts.onCreateUser` hook to query for similar information, then delete the old account and merge its properties on the new account. The intention (I think) of such solutions is to merge accounts while preserving the new login session. However, this results in potential data loss / data integrity issues.

**My solution is not opinionated on how this should happen, and leaves it up to the developer to define.**

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1348,6 +1348,15 @@ Ap.updateOrCreateUserFromExternalService = function (
     selector[serviceIdKey] = serviceData.id;
   }
 
+  // HOOK: Check if an update is needed to the user account from user supplied hook,
+  // caller should provide array of filter functions for modifying selector
+  var _checkUpdateHooks = Accounts.onUpdateUserFindExistingUserHooks || [];
+  if(_checkUpdateHooks && Array.isArray(_checkUpdateHooks) && _checkUpdateHooks.length > 0){
+	  _checkUpdateHooks.map(function (func) { //Each update hook can check the selector and apply a callback
+			selector = func(selector, serviceName, serviceData, options); //Each hook should return a modified selector
+		});
+	}
+
   var user = this.users.findOne(selector);
 
   if (user) {
--- a/packages/accounts-base/accounts_tests.js
+++ b/packages/accounts-base/accounts_tests.js
@@ -100,6 +100,27 @@ Tinytest.add('accounts - updateOrCreateUserFromExternalService - Twitter', funct
   Meteor.users.remove(uid1);
 });
 
+Tinytest.add('accounts - updateOrCreateUserFromExternalService - WITH HOOK', function (test) {
+	// Create user in the first pass
+	var user1 = Accounts.updateOrCreateUserFromExternalService('twitter', {id: 'test', monkey: 'blue'}, {profile: {externalServiceHook: 1}});
+
+	// Set a new selector to query the user with (could be $or selector, or whatever Mongo selector the user wants to find a user to update)
+	Accounts.onUpdateUserFindExistingUserHooks = [
+		function(selector, serviceName, serviceData, options){
+			return {"profile.externalServiceHook": options.profile.externalServiceHook};
+		}
+	];
+
+	var user2 = Accounts.updateOrCreateUserFromExternalService('github', {id: 'test', monkey: 'red'}, {profile: {externalServiceHook: 1}});
+
+	test.equal(user1.userId, user2.userId);
+	var users = Meteor.users.find({"profile.externalServiceHook": 1}).fetch();
+	test.length(users, 1);
+
+	//Cleanup
+	Accounts.onUpdateUserFindExistingUserHooks = [];
+	Meteor.users.remove(user1.userId);
+});
 
 Tinytest.add('accounts - insertUserDoc username', function (test) {
   var userIn = {


brettle:
I'm afraid I have to recommend against accepting this PR because I feel it encourages a poor security practice. The provided example code illustrates my concern. Consider the following attack scenario: 

An app supports login with either the Giggle service or the LinkedOut service. User has a Giggle account and logs into the app using his Giggle account which provides his verified email addreess of "user@example.com". User either doesn't have a LinkedOut account, or does but doesn't trust LinkedOut to vouch for his authenticity (e.g. because he uses a weaker password there). Attacker attacks LinkedOut (or the user's weaker LinkedOut password) and creates or logs into a LinkedOut account with "verified" email address of "user@example.com". Attacker signs in to app using that LinkedOut account and, because it has the same verified email address as the user's Giggle account, the attacker is logged in to the user's app account. When the user signed up with the app using his Giggle account, he did not realize (and had no reason to expect) that he was trusting LinkedOut (or his LinkedOut password) to not get hacked.

Regarding other solutions that delete the old account and merge it's properties into the new account, I agree that this can cause data loss / data integrity issues. If this is your primary concern, you might look at my `brettle:accounts-add-service` package. It preserves the existing account and adds the new service to it. It also ensures that the same user has actually logged in to both services so that the above security problem is avoided. Lastly, it also supports `accounts-password`, which your approach does not.

Sorry to be so negative but I'd hate to see Meteor encourage what I view as a poor security practice, especially when there are, IMO, better existing solutions to the problem you are trying to solve.

I hope all that makes sense and I hope you continue to contribute!

calvinfroedge:
@brettle Did you even read the code for the PR? This does support accounts-password because it does not modify it at all. Meteor as a framework already handles merging accounts based on the query that's returned. And the code I gave was just a sample of how a user CAN hook into the framework, it's not the commit code itself. The code in the commit just provides a hook. It's up to the user to determine what happens.

There are `not better existing solutions` and you didn't even read the commit code.

brettle:
For the record, I *did* read your code (both the commit and your example) before posting my comment.

When I said that your code does not support `accounts-password` I was referring to the fact that it does not support merging of a new password-based login into an existing account. This is because your PR only alters the method that deals with external services, as you acknowledge. I was trying to distinguish your approach from an approach like `brettle:accounts-add-service` which does support such merging. Sorry for not being more clear.

As for whether developers are capable of making their own decisions on which authentication services they trust, my concern is not which services the developer trusts, but which services the **user** trusts. The point of OAuth is to allow users and developers to delegate authentication to one or more services that they **mutually** trust. If I sign up with an app using a service which I trust and the app allows someone else to access my account by authenticating using a service which I don't trust, then the app has betrayed my trust. My concern with your PR is that it makes it particularly easy for developers to write apps that do that (as your example illustrates).

Regarding whether there are "better existing solutions" , my claim was that there were better existing solutions "to the problem you are trying to solve." My understanding of that problem is that you want to be able to merge a new login service into an existing account without data loss / data integrity issues. The `brettle:accounts-add-service` package is an example of an existing package that solves that problem with the only restriction being that the user must already logged in to the existing account when he logs in using the new service. If your goal is to solve the problem without that restriction, then I agree that `brettle:accounts-add-service` does not do that, but I also feel that removing that restriction is a betrayal of the user's trust.

stubailo:
I'd like to hear more from developers of other accounts merging packages, such as:

- @splendido of [accounts-meld](https://atmospherejs.com/splendido/accounts-meld)
- @lirbank of [accounts-merge](https://atmospherejs.com/mikael/accounts-merge)

splendido:
@stubailo I think such a hook should be heavily documented, at least, and there should be a better way to add new hook callbacks than directly accessing an `Accounts.onUpdateUserFindExistingUserHooks` field...

But, in any case, I see it very risky to let _external code_ modifying the selector object used to interact with `Meteor.users` collection.

For sure I think it'll be source of many headache for developers messing up with it without knowing well the underlying details about the `Accounts` system

calvinfroedge:
```there should be a better way to add new hook callbacks than directly accessing an Accounts.onUpdateUserFindExistingUserHooks field...```

I agree with this.

```For sure I think it'll be source of many headache for developers messing up with it without knowing well the underlying details about the Accounts system```

External code is used *after* a user is created already. That's how @splendido's plugin works. I would say that deleting old user IDs is also pretty risky. If you allow hooks at all, you should consider that *order matters*, and being able to hook into different points of a process is very important, especially for something like authentication.

I was able to accomplish this same requirement using [Auth0 Rules](https://auth0.com/docs/rules) within a matter of minutes. They're hooks work exactly this way:

1) User authenticates with 3rd party provider
2) User object is returned to the app
3) The app can modify the 3rd party data or update an existing user *before* upsert happens.

Here is an example implementation:
https://github.com/auth0/rules/blob/master/rules/link-users-by-email.md

One important difference here is that the rule actually queries the API, wheras in this meteor hook I modified the `selector.` Is it possible modifying the query is not the ideal solution? Yes. *I'm just saying there is a definite need for hooks / rules before user account actions take place.*

splendido:
I know I'm the author of `splendido:accounts-meld` but despite this, generally speaking, could having such a hook into the core accounts system be seen as a weak point for the platform as a whole?

The security of an application, from the accounting system point of view, would then be not only up to MDG but also up to the developer and the end user should convince herself that the developer was good too before registering on the app.
I know it's like this in any case, since a malicious developer could do whatever she likes to cheat accounts...
But since one of the selling point for Meteor is a very strong accounting system integrated into core, it won't be a good publicity for it the case of a developer which messed up with the selector modification and assigned a new user to the wrong (already existing) account...
Does this make any sense?

A completely different story would be discovering `splendido:accounts-meld` having a big security bug: in this case Meteor would still be safe and everyone should go blame splendido! ;-)

calvinfroedge:
@splendido A malicious developer could already use the existing hooks to do bad things if they wanted to. Look, I'm not even using Meteor any more because of issues like this. As a developer, I don't want someone else to make presumptions on what I should and shouldn't be able to do. I don't want a framework that assumes I'm stupid. I want a framework that allows me to spend *radically* less time creating products, gives me *fine grained* control over the details for when I need to worry about them, and assumes *that I am smart enough to reason about the implications of the code I write.* If MDG wants to make the next Wordpress, great, good for them. I'll stick with tools that save me time AND allow me fine-grained control over how they work.

I want *middleware*, I want *hooks*, I want *events*, I want *callbacks*. These are the things that sophisticated application developers need in a framework. If MDG is appealing to the developer who *doesn't* want to know about the details, and is *ok* with accepting the limitations of the framework because they couldn't *imagine* using their own brains to make *decisions*, that is MDG's business.

splendido:
I'm really sorry if my last post was offending, but that was not my intention for sure

splendido:
I really think adding such a hook into core should come with a lot of documentation and examples for the newbes (not the fools...) to clearly understand how to use it and get up and running flowlessly.

This said, I might even think this could be a reasonable proposal, but I wouldn't merge this PR in its current status. Obviously this is only my personal thinking...

tmeasday:

Issue #5320, opened 2015-09-29
static-html / caching-html-compiler have a jQuery dependency


mxab:
with 1.2.0.2 the jquery issue is back
see https://github.com/meteor/meteor/issues/5223

mxab:
@stubailo was there any specific reason why you went back to use jQuery to set the body attributes?

stubailo:
I introduced a bug when I removed it... it's more work than I thought!

Going to keep this open as a feature request.

mxab:
Hmm ok, but if I'm not mistaken everyone who wants to use angular instead of blaze needs to depend on jQuery explicitly otherwise the application will break right?
Currently my application depends on jQuery anyway so I can live with it, but I'm also trying to setup a [angular-meteor-platform](https://github.com/planet-training/angular-meteor-platform) package which would now need the jQuery dependency.
I haven't done any React, but I guess also there jQuery is optional

stubailo:
Yep, I know - hopefully we can fix it soon.

stubailo:
I'd love to look at a pull request for this. Here's the relevant code:

https://github.com/meteor/meteor/blob/9c46763e3f62cbe813769b1f88f5554b00c6605e/packages/caching-html-compiler/caching-html-compiler.js#L112

Basically we need to change this code to not use jQuery to set attributes. But we need to make sure to escape the attribute keys/values properly.

mxab:
Ok I see, I try to look into it

mxab:
@stubailo could you provide some samples where the escaping is needed? or are there some tests for this already?

stubailo:
No there are no tests yet. One example is if you do `<body x='"y"' ...` then a naive implementation without quote escaping will do some wrong things. This is what happened to me.

mxab:
so the generated markup should contain ```<body x='\"y\"'``` right? But did you write ```<body x='"y"'``` yourself in the firstplace and wouldn't this be invalid already?

ozsay:
@stubailo is this implementation good enough ?

```js
Meteor.startup(function() {
    var attrs = ${JSON.stringify(compileResult.bodyAttrs)};
    for (var prop in attrs) {
      document.body.setAttribute(prop, attrs[prop]);
    }
  });
```

What does this implementation lack?

stubailo:
This could work I think - I ran into some problems with escaping quotes but it looks like this wouldn't have that problem. Submit a pr to remove the deep and I'll accept it.

[...]

Issue #5223, opened 2015-09-21
static-html / caching-html-compiler have a jQuery dependency


mxab:
Hi,
if I use the static-html package, don't have jQuery in my dependency and use attributes on the body tag, the application fails since the caching-html-compiler uses jQuery to attach attributes to the body element.

https://github.com/meteor/meteor/blob/devel/packages/caching-html-compiler/caching-html-compiler.js#L112

This is my custom platform package
https://github.com/planet-training/angular-meteor-platform/blob/1.2-with-rc-deps/package.js

and this its sample project
https://github.com/mxab/meteor-ng-app/tree/attr_on_body

One thing to fix this is to depend on jQuery in the static html package but having jQuery as dependency on the "static-html" package is a bit of a paradox :)

stubailo:
Ooooooooh, that's a good point. I assume it's easy to change that code to not use JQuery, will do that now.

stubailo:
Fixed, releasing in RC 19: https://github.com/meteor/meteor/commit/e6bc9413f3f7313225dc4ae87e2e25a85ca0c59a

PR #5466, opened 2015-10-18
Remove unused jQuery dependency

timwachsener:


stubailo:
It's used by the compiled output to set body attributes. See:

https://github.com/meteor/meteor/issues/5320
#5223

To remove this dependency, you need to change the output to use plan JS: https://github.com/meteor/meteor/issues/5320#issuecomment-145156813

PR #5473, opened 2015-10-19
Fix documentation for Cordova.depends

benweissmann:
The link to the github documentation on Meteor's cordova integration pointed to an outdated page that redirects to the new page, which means the anchor link was totally broken.

In addition, the example using a github tarball URL used the short form of the sha, which Meteor doesn't accept -- trying to use that example would throw `must declare exact version of dependency: org.apache.cordova.camera@https://github.com/apache/cordova-plugin-camera/tarball/d84b875c`


stubailo:
Thanks!

Issue #5067, opened 2015-08-30
[BUG] [batch-plugins] Long wait time after my compiler handles sources and when Meteor continues.


trusktr:
The new `registerCompiler` API calls my compiler's (`rocket:module`'s) `processFilesForTarget` method. After `processFilesForTarget` returns, sometimes there's a period of time in the console with no output from Meteor, lasting about 1 minute.

I've got caching working in `rocket:module`, and `rocket:module` completes its tasks within seconds. The lag after `rocket:module`'s `processFilesForTarget` method returns is the only thing preventing quick refreshes for me right now.

Any ideas what might be happening? Is there any info I can provide?

cc: @glasser

stubailo:
Do you have a version of the package I can clone to try it out? Then I can add some logging or something to the command line tool to see what it's doing during that time.

trusktr:
@stubailo Yeah, I published `rocket:module` 0.8.2 on Atmosphere. It uses `api.versionsFrom('1.2-rc.7')`.

So, 

```
meteor create test
cd test
meteor update --release 1.2-rc.7
meteor add rocket:module
```

Then create an `npm.json` file and list some NPM dependencies in there (same format as package.json's `dependencies` field, for now):

```js
{
  "react": "^0.13.0",
  "foo": "^x.x.x",
  "bar": "^x.x.x",
  ... etc ...
}
```

Then create at least one entry point in the app. We'll create two: `client/entry.js` and `server/entry.js`.

In those files, begin importing the NPM dependencies like you would with CommonJS or ES6 module syntax. You can also import local files.

For example, `client/entry.js`:

```js
import React from 'react'
import foo from 'foo'
import bar from 'bar'

import something from './path/to/local/file'

Meteor.startup(x=>
  React.render(<div>Hello.</div>, document.body)
)
```

aliases for modules can be specified in `rocket-module.json`:

```js
{
  "aliases": {
    "elephant": "react"
  }
}
```

then you can do

```js
import React from 'elephant'
```

After you've started the app, the first build will always take longer, of course, and then everything should be cached. Make changes to the files, and rocket:module will rebuild in a matter of seconds.

You'll see output like this in the console:

```
[rocket:module] Compiling for platform "web.browser"...
npm WARN install Couldn't install optional dependency: Unsupported
npm WARN EPEERINVALID jss-props-sort@0.1.0 requires a peer of jss@^1.0.0 but none was installed.
[rocket:module] Done. Elapsed time: 6181ms
```

[...]

Issue #4310, opened 2015-05-01
Module system in Meteor.


trusktr:
Is there any plan to have a module system for Meteor (ES6 Modules perhaps)?

I'm using Webpack to make bundles (one can be made for the server and one for the client). Each bundle can `require`/ or `import` common things used on both sides. It works nicely...

But problems can arise when (for example) a webpack bundle needs to depend on an Atmosphere package's globally exposed API, and that API depends on another Atmosphere package that contains an external non-Meteor library that the initial webpack bundle depends on too, causing duplicate code to exist in the app: one in the webpack bundle, and one in an Atmosphere package. For example, a library I'm using in my webpack bundle depends on [Famous](https://www.npmjs.com/package/famous) from npmjs.com, while a library I need from Atmosphere depends on [raix:famono](https://atmospherejs.com/raix/famono) which supplies a global version of Famous, so I have two Famous libraries in my app, which can lead to unexpected things happening because two sets of code that depend on Famous both want to use a Famous `Engine`, but there's no out of the box way for them to share the same `Engine`. A quick solution would be to port the Atmosphere package depending on raix:famono to npm, depend on Famous from npm instead of raix:famono, then I can use that library in my webpack bundle and there will be a single Famous `Engine` in the app.

It can be harder for large teams in large organizations to adopt Meteor for projects that are projected to be big because it doesn't have a module system (where modules with file-level scopes can export things that *aren't* put in global scope and other file-level scopes can import those things directly) in a world where the best tools outside of Meteor (webpack, browserify, etc) already have this (CommonJS for importing and exporting things at first, but AMD/ES6 with extensions to the tools).

Imagine what happens when a Meteor app gets huge, with hundreds of thousands of lines of code. The largest Meteor app I've seen in person was 12,000 lines and I could already see the complications.

Galaxy will scale application bandwidth, but we also need something that scales developer productivity in a standard way that the whole rest of the JavaScript community uses (CJS and now ES6 Modules, goodbye AMD). Meteor's packaging system is evolving aside from the JavaScript community's overall standard instead of working with it.

**Is Meteor's packaging system made so that beginners can get started quickly?** It's obvious that adding packages to a Meteor app and having things exposed in global scope makes it easy for novices to just start writing code (perhaps spaghetti code). This problem can be solved by simply telling people to make two entry points, one for server, one for client, and for common code to be put in a third place to be imported by either of the two sides. The tutorial can briefly explain how to import and export things from modules.

[...]

PR #5475, opened 2015-10-20
[WORK IN PROGRESS] Implement ECMAScript 2015 modules

benjamn:
## Summary

My plan is to use Babel to compile ECMAScript 2015 `import` and `export` syntax to CommonJS, and then use a new primitive called `Meteor.install` to implement a CommonJS module system on both the client and the server. Running the same bundled code everywhere will allow the Meteor module system to be truly isomorphic.

## Comparison to existing tools

Why invent `Meteor.install` instead of borrowing the bundle formats of [Browserify](http://browserify.org/) or [Webpack](https://webpack.github.io/)? The goal of those tools is to make server-side Node code runnable on the client. Either tool would probably work well for Meteor, if the server-side portion of a Meteor app looked like a Node project, but that is unfortunately not the case. Given the complexities of the Meteor package system (and especially build plugins), the difficulty of getting Browserify or Webpack to traverse the dependency graph of a Meteor app outweighed the complexity of implementing that logic within the existing Meteor build process.

That is not to say there is nothing from Browserify or Webpack that we can reuse! For example, it should be possible to use the same [resolve](https://www.npmjs.com/package/resolve) NPM package that Browserify uses; the trick is invoking it at the right point in the build process: after build plugins run, but before the bundled files are linked together, with package variables imported and exported in a backwards-compatible way, combining source maps from all the input files into a single source map for the final bundle, with a set of rules about which modules should be treated as entry points and which modules should be evaluated on demand. There are a lot of details to get right, so it makes sense to pick and choose what code we borrow from other projects and what we write from scratch.

`Meteor.install` is comparable to the output formats of Browserify or Webpack, but those output formats are tightly coupled with the tools that generate them, and I wanted `Meteor.install` to make sense as a standalone API. For example, the Webpack tool generates a sequential numbering of all the modules in the bundle, then rewrites every `require(id)` call as a `__webpack_require__(n)` call, where `n` is the number of the required module. This numbering strategy works because Webpack can assume the code that generates the numbers also generates the output bundle. If you divorce those two aspects of the module delivery system, then the output format becomes significantly less appealing. Numbering modules also limits the effectiveness of caching in the build system, since the values passed to `__webpack_require__` often have to be rewritten when the dependency graph changes in some unrelated part of the application. Furthermore, since the generated numbers do not preserve the original structure of the application, and must be determined at build time, passing dynamic string values to `require` is effectively forbidden. And of course the coupling between tool and output format becomes even more complicated when you want to generate separate bundles to be loaded at different times on different pages, instead of generating a single monolithic bundle.

In short, the argument for reusing the output formats of Browserify or Webpack is only compelling if you're already using those tools during your build process, in their entirety. Without that harmonious integration of tool and output format, designing a new output format from scratch becomes a lot more appealing.

## `Meteor.install` in detail

With that background out of the way, let's take a look at how `Meteor.install` works.

### Basic usage

`Meteor.install` takes either one or two arguments: an optional package name, and a tree of nested objects and function expressions that describe the structure of your application's directory tree. Objects correspond to directories, object property names correspond to directory or file names within directories, and object property values correspond to file contents (objects represent nested directories, and functions represent module files).

For example, the following code adds a module called `"foo.js"` and an NPM package called `bar` to a Meteor package called `"test-package"`:
```js
var require = meteorInstall("test-package", {
  "foo.js": function(require, exports, module) {
    exports.foo = true;
    exports.bar = require("bar");
  },
  node_modules: {
    bar: {
      "index.js": function(require, exports, module) {
        exports.bar = true;
      }
    }
  }
});
```
In situations where you know you have installed all the modules that might be `require`d, you can now call `require("./foo").bar` to evaluate the `foo.js` module, which calls `require("bar")`, which is resolved to a package in the `node_modules` "directory," which contains an entry point called `"index.js"` whose `module.exports` object becomes the value of the expression `require("./foo").bar`. If you're familiar with [the Node module loading system](https://nodejs.org/api/modules.html#modules_modules), this logic should be pretty familiar. It's designed to work exactly the same way.

### Asynchronous loading

A great virtue of `Meteor.install` is that you can call it multiple times, filling in the structure of installed modules incrementally. In such situations, you may not know exactly when all the dependencies of a module will become available, so it is important to declare those dependencies using an array expression instead of just a function expression:
```js
// This require function will be used in the next snippet of example code.
var require = meteorInstall("test-package", {
  "foo.js": [
    // This array expression indicates that the foo.js module depends on some
    // possibly-not-yet-defined module identified as "bar".
    "bar",
    function(require, exports, module) {
      exports.foo = true;
      exports.bar = require("bar");
    }
  ]
});
```
Now you can require `foo.js` asynchronously, using [`require.ensure`](https://webpack.github.io/docs/code-splitting.html#require-ensure), and trust that its callback will be run when all dependencies become available:
```js
// In the context of the "test-package" package, evaluate the following function
// as soon as it becomes safe to call `require("./foo")`:
require.ensure(["./foo"], function (require) {
  require("./foo");
});
```
If all the function needs to do is `require` the depended-upon modules, then you can omit the function and use the following shorthand:
```js
require.ensure(["./foo"]);
```
Note that the `require` method for a package can be obtained by calling `Meteor.install(packageName)`, even if you have nothing new to install, and for application code by calling `Meteor.install()` with no arguments.

For more examples, please refer to [the tests](https://github.com/benjamn/meteor/blob/wip-Meteor.install/packages/meteor/common_environment_test.js).

## Next steps

Looking ahead to the next couple of weeks, I'll be filling in this pull request with server-side logic for generating `Meteor.install`-style bundles, but I wanted to get this explanation of `Meteor.install` out in the open sooner rather than later.


rozzzly:
Thank you +1

stubailo:
Does this include any ideas for how to conditionally load modules on different architectures, for example Cordova/Browser? Or different environments, like production/development?

benjamn:
@stubailo different combinations of architecture and environment will evaluate different `Meteor.install` bundles. If you're building for `web.browser`, `Meteor.install` will never install anything from the `server/` directory, and vice-versa for `os` architectures and `client/` directories.

The interesting case is when a file is evaluated on multiple architectures and/or multiple environments (the isomorphic dream!). My current thinking is that we should simply ignore `require`d dependencies in those files, so instead of calling `require` in an `if (Meteor.isClient) {...}` block in an isomorphic file, you have to move that code into a file inside a `client/` directory.

benjamn:
After writing this up, I no longer like the idea of `Meteor.install` sometimes evaluating modules (asynchronously) and sometimes just installing them.

Instead, I'd like to move the full burden of evaluating modules into the `require` function, both the one returned by `Meteor.install` and the one available to installed modules. Note that this function can always be obtained, even if you have nothing new to install, by calling `Meteor.install(packageName)`, or just `Meteor.install()` for app code.

To avoid inventing a new syntax, I'm inclined to borrow [Webpack's `require.ensure` API](https://webpack.github.io/docs/code-splitting.html#require-ensure). I like `require.ensure` because it does not force dependencies to be evaluated before the callback gets called. Instead, `require.ensure` calls the callback as soon as all the dependencies are *ready* to be required, so you can do something like
```js
require.ensure(["a", "b"], function (require) {
  require("a");
  if (shouldEvaluateB()) {
    require("b");
  }
});
```
Because I love `Promise`s, I would also like to provide an API called `require.promise`, which could be used like so:
```js
require.promise("a").then(function (a) {
  // a === require("a")
});
```
To join on multiple promises, you could do
```js
Promise.all([
  require.promise("a"),
  require.promise("b")
]).then(function ([a, b]) {
  // a === require("a");
  // b === require("b");
});
```
but that's a lot of boilerplate compared to something like
```js
require.promise(["a", "b"]).then(function([a, b]) {
  // a === require("a");
  // b === require("b");
});
```
Let me know if you have strong feelings about how `require.promise` ought to work!

arunoda:
This is great. This question is just to make sure I got it correct.

* `require.ensure` and all these promise API are something happening internally. We(as normal meteor users) interact with `import` and `export` (no promise stuff)?
* Can we export modules inside the Meteor app? What if there is entry files for different architectures and one for all. Then we can build the app as we need by requiring.
* Is it possible to dynamically build modules? If so, We can alter Router to load all the modules need to load the first page. Then load others asynchronously.

vjau:
I don't get where all this boilerplate is gonna be used, in your code or in special meta files like package.js ?

trusktr:
> Either tool would probably work well for Meteor, if the server-side portion of a Meteor app looked like a Node project, but that is unfortunately not the case. Given the complexities of the Meteor package system (and especially build plugins), the difficulty of getting Browserify or Webpack to traverse the dependency graph of a Meteor app outweighed the complexity of implementing that logic within the existing Meteor build process.

Can you expand on that?

I was thinking of adding the ability to import anything from any other package (which isn't that hard, just haven't worked much on it lately as I haven't had much need for it yet) similar to the `less` package's `@import` mechanism.

The concept in `rocket:module` is this: take each package of the app and treat it as an NPM package, make a package.json for each package that lists NPM dependencies, then make a root package that depends on all the packages, then `npm install` in the root package, then compile all entry points using `webpack`, and voila (with a few tweaks) we have compiled entry point bundles for each package (`entry.js` files) that can be added back using `file.addJavascript`.

(By the way, Webpack handles module de-duping, which `rocket:module` takes advantage of to ensure that modules are only compiled into the app a single time (thanks in part to NPM v3's flat dependency file structure). There's room for performance improvements in `rocket:module`, but it works.

So, in theory, even when using purely the `registerCompiler` API, I was able to make all this `import`/`export` stuff happen (minus a few small caveats which I've been lazy to fix due to time constraints, but am now wanting to leave one job in order to find more time for personal projects like `rocket:module`) and there's a clear path to making it easy to being able to import anything from any package like `less`'s `@import`s.

On another note, if making all this from scratch instead of using an existing tool like Webpack, why not just use SystemJS, which implements the ES6 `System` API -- or make an API based on the upcoming System API -- so that Meteor will be legitimately ready for when ES6 modules actually land in browsers?

I was thinking of doing this in `rocket:module` (taking the advantage of the System API) by swapping out Webpack with JSPM (JSPM is the most advanced in this regard).

JSPM already handles dependency de-duping and forking (forking where a package A depends on C@1.x.x and B depends on C@2.x.x) much better than NPM v3 (even before NPM v3 existed), and much faster.

I really think aligning with the System API is the way to go, with Webpack as a second choice (although at the time I made `rocket:module`, webpack was easier to configure so I used it over JSPM, but JSPM has made nice strides lately and has a few more nice ones coming up).

trusktr:
The only downside of `rocket:module` based on what you've described is that it itself is a build plugin, and you'd like for ES6 modules to run regardless of (right?). Could a solution to that be build plugin priority (the order they are defined with `api.use` in a package)? Then files would simply pass on to the next build plugin.

trusktr:
> Numbering modules also limits the effectiveness of caching in the build system, since the values passed to __webpack_require__ often have to be rewritten when the dependency graph changes in some unrelated part of the application

`rocket:module` has Webpack caching working, kept alive between calls to `processFilesForTarget`. It reduced build time a lot already. The only things left to make performance better is how we treat the creation of the NPM pseudo-packages, which are currently being re-written every single time `processFilesForTarget` is called, and some other things unrelated to Webpack caching.

There's also this bug in Meteor (I'm fairly certain it's meteor since it happens after the call to my `processFilesForTarget` call has returned) that is preventing full performance: https://github.com/meteor/meteor/issues/5067

I haven't had time to make a simple reproduction, and @stubailo hasn't had the time to look into the more complex case that I emailed him yet. I live nearby, in West Oakland, so I'd be willing to set up a meeting to go over it. I'm no expert like you guys are with Meteor and I haven't found the time to find the problem myself yet.

trusktr:
> And of course the coupling between tool and output format becomes even more complicated when you want to generate separate bundles to be loaded at different times on different pages, instead of generating a single monolithic bundle.

I don't think the tools are in the way. The entry points are simply compiled files that are put back into their place with `file.addJavascript`. I believe it can be a separate API that is capable of loading the file (regardless of how it got compiled) at some point in time separate from when the bundle loads, and API not coupled to the build plugin at all.

trusktr:
> To avoid inventing a new syntax, I'm inclined to borrow Webpack's require.ensure API.

The one that has already been invented is official, and it's `System.register` and friends.

trusktr:
@stubailo 

> Does this include any ideas for how to conditionally load modules on different architectures, for example Cordova/Browser?

This already exists with `api.addFiles` when you specify an architecture, doesn't it, and build plugins (like `rocket:module`) work with this.

But, like I mentioned, I'm no expert like you guys on the internals of Meteor, I'm more familiar with it's API, so I might not see as clear a picture as you guys see, but I do believe that the concept of trying to satisfy the `System` API would carry a long way, since that's the future.

trusktr:
Could you guys perhaps provide a high-level list of the things that a module system will allow end users to do with Meteor? Then I can compare to what `rocket:module` already does (and what is/was planned).

trusktr:
@arunoda 

> Can we export modules inside the Meteor app? What if there is entry files for different architectures and one for all. Then we can build the app as we need by requiring.

We can already do this with `rocket:module`.

> Is it possible to dynamically build modules? If so, We can alter Router to load all the modules need to load the first page. Then load others asynchronously.

I think an approach might be to modify `api.addFiles` to allow for loading those files by name (including entrypoint files compiled by `rocket:module` or any other build plugin). For example:

```js
api.addFiles('foo.js', 'server', {
  bundle: 'lorem'
})
```

If and only if there's the `bundle` option in the 3rd argument to `api.addFiles`, then the files added in that call will be separated from the main bundle and into a separate bundle called "lorem". The client can then do something like

```js
Meteor.loadBundle('import')
```

which queries the server for the bundle.

If Meteor wants to play well with JavaScript standards, then that idea can be replaced with a `System`-based API, f.e.

```js
System.import('lorem') // perhaps that loads a bundle, or perhaps that loads a "lorem" meta package that imports all the files from the server with HTTP/2 in one single connection.
```

trusktr:
@rclai said in #4310 

> I think a core package (called modules or something like that) should be made to make a Meteor app function via modules, otherwise it should function the way it does now in order to keep simplicity.

That would be essentially what `rocket:module` is (although it's not entirely finished yet and needs polishing, but the ideas are already complete except for cross-package imports which I feel is the last major idea that needs to be added).

Also, @MacRusher's `universe:modules` is already using SystemJS, so that can also serve as some inspiration towards using the `System` API.

casperanderseneu:
@trusktr

I read the thread where rocket:module was compared with universe:module

How is the meteor-rocket/webpack implementation different from
thereactivestack/meteor-webpack ?

What does rocket:module add on top of webpack, that justify another package
on top of a webpack package?

I like a future scenario where

1. meteor embraces the eco-system and makes it easy to use existing up to
the edge industry tools and packages, rather than going out of its way to
implement its own abstractions, with pain for users not benefiting from the
general advances taking place in the bigger world.

2. we can conditionally based on whatever we can learn from the initial
request provide a base client with login and a ssr backend for seo

3. action/event/routes trigger loading of additional modules
(code-splitting)

4. hot module replacement (remove and add) on condition works in both dev
and prod, according to webpack docs it is not impossible for prod too.

5. standalone build of modules, so they can be build and served separately
from the main client, like lazy-loading from external sources, or whatever
is better.

trusktr:
@casperanderseneu 

> How is the meteor-rocket/webpack implementation different from
thereactivestack/meteor-webpack ?

rocket:wepack merely provides the `Wepack` symbol that contains a refernence to Webpack from NPM, that's all. it doesn't add anything extra. `rocket:module` uses that. Thereactivestack/meteor-webpack on the other hand starts to do some things like `rocket:module`, but all in the same package as webpack.

> 1. meteor embraces the eco-system and makes it easy to use existing up to
the edge industry tools and packages, rather than going out of its way to
implement its own abstractions, with pain for users not benefiting from the
general advances taking place in the bigger world.

I think that's already the goal of the efforts here and in the packages.

> 2. we can conditionally based on whatever we can learn from the initial
request provide a base client with login and a ssr backend for seo
> 3. action/event/routes trigger loading of additional modules
(code-splitting)
> 4. hot module replacement (remove and add) on condition works in both dev
and prod, according to webpack docs it is not impossible for prod too.
> 5. standalone build of modules, so they can be build and served separately
from the main client, like lazy-loading from external sources, or whatever
is better.

I think all those cases can be covered via the `System` API. :+1:

rozzzly:
So today I spent a few hours looking at how **[`[vazco/universe-modules]`](https://github.com/vazco/universe-modules)** works; did a lot of digging arround in isobuild. Here's my take on things:

Although I think webpack is great, Using it so we can get ES6 modules would force the developer into using webpack. That opens up a whole new can of worms. I've been watching [**`[jedwards1211/meteor-webpack-react]`**`/issues`](https://github.com/jedwards1211/meteor-webpack-react/issues) for some time now. As you can see, it doesn't work 100% out of the box. Webpack and meteor are both large, distinct pieces of software. Hence, you simply can't get them to work together without a non-insignificant amount of effort. While the webpack approach has some awesome benefits, I see the potential of innumerable edge-cases as a deal breaker.

------------

 It took me a while to piece together the logic of @benjamn's proposal, and it seems it would work logically. However, as @arunoda asked:
> require.ensure and all these promise API are something happening internally. We(as normal meteor users) interact with import and export (no promise stuff)?

I was going to ask the same thing. Because while that proposal is well conceived, where that boilerplate would be used/to what degree is not apparent. If this was something taken care of by a babel plugin where all I needed to do was write my `import` statement and then all of that was abstracted away, I would be fine with that! But creating an entirely new `meteor.install(...)` syntax kind of defeats the purpose of adopting the ES6 module syntax, because at that point its kind of just another module system. I'd love to hear your thinking @benjamn, that sounds like it could work quite well as the internal mechanism for modules, but I couldn't condone that as the outward-facing api. 

-------------

**[`[vazco/universe-modules]`](https://github.com/vazco/universe-modules)** does not have either dependency. No additional syntax (beyond the initial `System.import(...)`) and no need to configure webpack. It does everything I wan't from modules. Namely, the concise `import` and `export` syntax. **However there is one big issue** that is keeping me from using it today. All of the files to be compiled with support for the es6 modules, must have a `*.import.js` extension. `*.js` files do not get picked up by the package's custom compiler, and are processed as normal (ie: get thrown out into the global scope). Starting with Meteor 1.2.0, It's now possible to register a compiler that handles all the `.js` extension. In fact, that's exactly what the `ecmascript` package does to feed the code into Babel. 

So here is my vision.

What if we took what `universe-modules` does, but flipped it around. That is, all the files in our app or packages (if specified of course) would be processed in the way that `*.import.js` are currently handled, ignored by the usual meteor load order. Then any of the files with a new **`*.entrypoint.js`** extension would be kept in the global scope, processed by the default compiler.  


(to be continued)

trusktr:
@rozzzly 

> What if we took what universe-modules does, but flipped it around. That is, all the files in our app or packages (if specified of course) would be processed in the way that *.import.js are currently handled, ignored by the usual meteor load order. Then any of the files with a new *.entrypoint.js extension would be kept in the global scope, processed by the default compiler.

That's exactly what `rocket:module` does.

rozzzly:
Not quite, it uses webpack. I haven't had a chance to  investigate `rocket:module`, go through and read all of the source code. But that is closer to what I'm getting at.

Originally, I was hoping I could use a regex here:
```
Plugin.registerCompiler({
    extensions: ['import.js', 'import.jsx'],
    filenames: []
}, function () {
    return new UniverseBabelCompiler();
});
```
https://github.com/vazco/universe-modules/blob/master/build-plugin.js

But after digging into isobuild, I find that it keeps a [map of `SourceProcessor`s indexed by extension/filename](https://github.com/meteor/meteor/blob/devel/tools/isobuild/build-plugin.js#L181). So to support regex, isobuild would require a small patch. **However** after reading a little bit of `rocket:module`. I see it filter's out the entrypoint files in the compiler's `processFile(...)` itself. In fact, it explicitly states it wants to handle those files too.

````
   Plugin.registerCompiler({
        // TODO: Add css, typescript, coffeescript, etc.
        extensions: [
            'js', 'jsx',

            'entry.js', 'entry.jsx',
            'client.entry.js', 'client.entry.jsx',
            'server.entry.js', 'server.entry.jsx'
        ],
        filenames: [
            'npm.json',
            'rocket-module.json',

            'entry.js', 'entry.jsx',
            'client.entry.js', 'client.entry.jsx',
            'server.entry.js', 'server.entry.jsx'
        ]
    }, x=> new RocketModuleCompiler)
```
https://github.com/meteor-rocket/module/blob/master/plugin/plugin.js#L729

> I'm not sure why it asks for the `entry.js` there because it's going to get picked up thanks to `js` anyway so it'd be redundant. _If there's some special reason, I'd love for someone to point it out to me, thanks_ :wink:  

------

I have an idea that would get my implementation working in just a few lines. So, I'm going to go try that out. I'll report back in a few hours.

deanius:
Great start @benjamn - I agree with the points above about preferring to use `import/export` instead of the new functions defined above. If things can be compiled to `require.promise` etc from I think I'd prefer it. But thanks for putting this reco out early so people busy learning Webpack like myself can see and maybe influence what an MDG, true-Meteor solution might look like :+1:

rozzzly:
+1 @benjamn  I totally agree with @deanius, there have been quite a few issues opened in the past few days, but you were the first person to actually put something on the table. Thanks my friend. It makes me quite happy to see quickly meteor is adapting and what a healthy opensauce community we've got here. _Everybody give themselves a pat on the back._

(disclaimer: first as an offical Meteor solution. [**`[rocket:module]`**](https://github.com/meteor-rocket/module) and [**`[universe:modules]`**](https://github.com/universe/modules) have been around for quite a while. _Don't feel let out, you're awesome too!_)

Sanjo:
@benjamn Have you considered using [SystemJS](https://github.com/systemjs/systemjs) with the [SystemJS module format](https://github.com/ModuleLoader/es6-module-loader/blob/master/docs/system-register.md) and module loader? I think [universe:modules](https://atmospherejs.com/universe/modules) shows that it works. It is also not coupled to any build tool. Babel already supports transpiling to this module format however ;-)

For Meteor packages we could prefix the module path with `{acme:package-name}/` ([see documentation](https://github.com/vazco/universe-modules/#loading-modules-from-packages)) instead of having the package name separate from the path.

The system should support code-splitting (bundles) and a way to lazy load the bundles from the server when they are required. This is also already supported by SystemJS. ;-)

benjamn:
@arunoda @vjau your hopes/suspicions are correct: none of this `Meteor.install` boilerplate will appear in application or package code, it's just going to be the output format of the Meteor build system. In other words, if you open up your browser console and peel back the source maps, you'll see all of your code wrapped by `Meteor.install`, but otherwise you won't have to worry about it. I mentioned `require.ensure` for two reasons: I wanted to draw some attention to the fact that `Meteor.install` supports asynchronous module loading, and I think Webpack has demonstrated that `require.ensure` is a good mechanism for identifying potential bundle boundaries. If Meteor supports bundle splitting in the future, there's a good chance it will be based on `require.ensure`.

I'm planning a longer, higher-level forum post about the goals (and non-goals) of this project, but the very short version is that Meteor 1.3 will enable the Babel transform that converts ES2015 `import` and `export` to CommonJS `require` and `exports`. Although the emphasis will be on ES2015 modules, the use of CommonJS behind the scenes is more than just an implementation detail, because it will make NPM packages work without modification in application code, on both the client and the server, without any wrapper Meteor packages. Just create a `node_modules` directory anywhere in your app, `npm install` some packages into it, and `import` or `require` them to your heart's content.

I'll say more about things like controlling file load order using `import` statements (a big improvement over the current set of rules), and loading modules lazily rather than eagerly, in that forum post. Stay tuned!

@trusktr I very much like the `rocket:module` strategy of creating an NPM package for each Meteor package, with a single umbrella package, then using `npm` (version 3.x, presumably?) to install them so that common dependencies get hoisted to the top level, where they can be shared among packages.

@sanjo `System.register` will be more appealing when/if it becomes part of the standard library in at least some JavaScript engines, but until then the polyfill has a pretty big runtime footprint (not a deal-breaker, just something that makes me hesitate), and the API for defining modules is much more cumbersome than the ES2015-to-CommonJS transform. With `System.register`, the bundle code must provide setter functions for every imported variable, and those variables all have to be hoisted into a separate scope. The CommonJS transform achieves the same live-binding semantics by turning imported vars into object member expressions, which is easier to understand and explain. Yes, you might not be able to refer to those variables in an `eval` string, but `eval` becomes unreliable pretty much any time you're using a transpiler, so I think that's not much of a loss.

vjau:
That's a good news @benjamn , thanks for clarification.
For me one of the strong selling point of your implementation would be full "compatibility" with Typescript, i don't know if you are familiar with the subject.

benjamn:
@rozzzly 

> Although I think webpack is great, Using it so we can get ES6 modules would force the developer into using webpack. That opens up a whole new can of worms. I've been watching [jedwards1211/meteor-webpack-react]/issues for some time now. As you can see, it doesn't work 100% out of the box. Webpack and meteor are both large, distinct pieces of software. Hence, you simply can't get them to work together without a non-insignificant amount of effort. While the webpack approach has some awesome benefits, I see the potential of innumerable edge-cases as a deal breaker.

That's been exactly my experience. As much as I would like to use Webpack, and as much as I hope that a future version of Meteor will encourage a project structure that is easier for Webpack to consume, Meteor 1.3 has to be backwards compatible in a way that seems to forbid using Webpack or Browserify in their entirety (though reusing certain parts of them is a really good idea).

> So here is my vision.

> What if we took what universe-modules does, but flipped it around. That is, all the files in our app or packages (if specified of course) would be processed in the way that *.import.js are currently handled, ignored by the usual meteor load order. Then any of the files with a new *.entrypoint.js extension would be kept in the global scope, processed by the default compiler.

That's essentially my vision, too, except that I'm planning to treat any files inside an `imports/` or `node_modules/` directory as non-entry-points, while files outside those directories will be loaded the same way as before, according to usual Meteor load order. If you're really excited about modules, you can put your entire app into `client/imports/` and `server/imports/`, and then have just one entry-point module per environment, perhaps `client/main.js` and `server/main.js`.

benjamn:
@vjau you will probably still have to use a compiler plugin for `.ts` files, but as long as TypeScript can generate CommonJS code that uses `require` and `exports`, then you won't have to worry about how that's implemented anymore.

vjau:
The problem is not the compile step, but for the compiler to know which scope each file has access to. With current package system, it's very complicated/impossible.

benjamn:
@vjau got it (I think)… so `import` statements will allow the compiler to do better static analysis of your TypeScript code?

vjau:
Yes @benjamn, every "free" variable appairing in your code must be "justified" or typescript will complain.
Please note that there are also some changes/enhancement in the work 
https://github.com/Microsoft/TypeScript/issues/5039

trusktr:
@rozzzly 

> I'm not sure why [rocket:module] asks for the entry.js there because it's going to get picked up thanks to js anyway so it'd be redundant. If there's some special reason, I'd love for someone to point it out to me, thanks

I just put the extra ones there for readability, although yeah, it's not necessary.

@benjamn 

> I very much like the rocket:module strategy of creating an NPM package for each Meteor package, with a single umbrella package, then using npm (version 3.x, presumably?) to install them so that common dependencies get hoisted to the top level, where they can be shared among packages.

Yep, NPM v3 for the flat file structure most of the time (except when there's dependency forks). JSPM still handles forks with a flat file structure, and is still much faster than NPM.

> That's essentially my vision, too, except that I'm planning to treat any files inside an imports/ or node_modules/ directory as non-entry-points, while files outside those directories will be loaded the same way as before

That's what I'm doing in `rocket:module`, except the folder is `modules` instead of `imports`.

rozzzly:
#### how about a `.gitignore` type syntax marking (non)imports? that would be perfect, imho

minimal configuration, no forced structure/naming conventions 

------------------------------------------------

> The CommonJS transform achieves the same live-binding semantics by **turning imported vars into object member expressions**, _which is easier to understand and explain_. Yes, you might not be able to refer to those variables in an eval string, but eval becomes unreliable pretty much any time you're using a transpiler so I think that's not much of a loss. --- @benjamn 


I've been playing around with it and I understand what `System js` is doing, but I'm not following you there. Care to link/point me in the right direction? Couldn't come up with a google query that would enlighten me :/ 

-----------------------------

and I really love the sound of npm modules working naturally, but this is even more impressive

> I very much like the rocket:module strategy of creating an NPM package for each Meteor package, with a single umbrella package, then using npm (version 3.x, presumably?) to install them so that common dependencies get hoisted to the top level, where they can be shared among packages.

This is one of the things I found so appealing about the webpack route. Would this align with the way MDG is approaching modules? seems like it would be quite involved. 

------------------------------


@trusktr  fair enough

trusktr:
> how about a .gitignore type syntax marking (non)imports? that would be perfect, imho

I like that idea. Being able to use regexes would convenient for this purpose, as well as in package.js for adding files so we don't have to tediously list every file.

glasser:
This is super exciting!  I love where you're going with this.

We chatted a little about this directly, but my main comments on this PR are:

- It looks like this API treats "top level package name" as a special concept.  I think this makes a lot of sense in the Meteor context, but maybe it's a little too hardcoded, and we should just imagine Meteor.install calls as defining a big tree and you're allowed to call `Meteor.install` to augment the things accessible from any part of that tree, not just what are effectively top-level subdirectories?
- It would be really great if this format was also what ends up implementing the current pseudo-global imports and exports created by Linker, so that we end up with one mechanism (with interop between linker exports and CommonJS imports and vice versa) rather than two parallel mechanisms.
- Consider calling the API something other than `Meteor.install` — both because `Meteor` itself is already a symbol created by, well, this API, and because it's generic enough that it probably will be useful outside of Meteor.
- Consider making the main entry `Meteor.install` (or whatever you call it) be a specific default instance of a tree, but be based on an even lower-level API that lets you make multiple such trees.  If for no other reason than testability (drop your guids!) and general avoidance of global singletons.

Sanjo:
> Consider making the main entry Meteor.install (or whatever you call it) be a specific default instance of a tree, but be based on an even lower-level API that lets you make multiple such trees. If for no other reason than testability (drop your guids!) and general avoidance of global singletons.

:+1:  Please. __Use case:__ Create a clean app state for each unit test, to have isolated tests.

ccorcos:
One thing to consider with this build system stuff are build-system plugins. 

http://puppybits.github.io/talks/More_Speed_More_Money.html#1

trusktr:
The cool thing about the existing tools like webpack/browserify/jspm is that I can make things like

```js
import fragment from './path/to/fragement.glsl'
import foo from './path/to/foo.coffee'
import './path/to/styles.css'
```

work in all three (although substack [disagrees with import overloading](https://gist.github.com/substack/68f8d502be42d5cd4942#overloading-require), people love and [do it anyways](https://www.npmjs.com/package/browserify-css)). It'd be really nice for that to work in Meteor too, without having to modify it.

The bottom line is that I'd like to write libraries that can be used anywhere, with any of the popular build tools and in Meteor.

I don't think import overloading is part of the upcoming loader spec, so it'll still be a thing to be implemented by tooling.

How do you feel about that @benjamn and @glasser?

benjamn:
@trusktr my plan is that the Meteor module system will use the final, plain-CommonJS/ES5 output from compiler plugins when it scans for `require`d dependencies. That means you'll be able to import any kind of file that can be compiled to JS by a plugin, and compiler plugins can inject their own `require` calls, even if there was no corresponding `import` or `require` in the source code.

The `coffeescript` package is a good example: because the output is just JS, you'll be able to `require("./path/to/foo.coffee")` from a `.js` file, and you'll get back whatever `./path/to/foo.coffee` exported (by modifying `exports` or `module.exports`). I'm actually not sure what kind of module syntax CoffeeScript supports, but as long as the compiled output uses `require` and `exports`, it should be able to be consumed by any other Meteor code.

I would like to stay away from `import`/`require` overloading as much as possible, simply because those tool-specific module identifier conventions are unlikely to become part of the language any time soon.

It would be cool if Webpack/Browserify/jspm plugins could be consumed by Meteor as if they were Meteor-style compiler plugins, but in the meantime I believe your best option is to use/write Meteor plugins that output CommonJS.

trusktr:
Sounds good. So it's just a matter of matching webpack *-loaders or browserify *ify transforms with Meteor build plugins, and code can remain the same. I also agree about not using prefixes as in `require('css!path/to/style.css')`. I suppose the build plugin for CSS in the case of `require('path/to/style.css')` would create a module that automatically writes the CSS to the document, and exports nothing. Simple enough!

benjamn:
> I suppose the build plugin for CSS in the case of require('path/to/style.css') would create a module that automatically writes the CSS to the document, and exports nothing. Simple enough!

Exactly! I also have a dream that one day `require("path/to/style.less")` will return an object containing all the variables defined in `style.less`, so that you don't have to duplicate those variable declarations in your JS.

ccorcos:
@benjamn I'm right there with you -- I think the answer will eventually be styles via JS.

trusktr:
@ccorcos 

> styles via JS

There's already [JSS](https://github.com/jsstyles/jss), which I think is awesome.

ccorcos:
@trusktr whoa. pretty neat. Why haven't I been using this?!

trusktr:
@benjamn Here's a really good reason to consider using SystemJS: https://github.com/systemjs/systemjs/issues/895#issuecomment-154134762

benjamn:
Rebased against https://github.com/meteor/meteor/pull/4619 and added a pretty exciting (though still experimental) commit that compiles stylesheets to JS modules when possible: https://github.com/benjamn/meteor/commit/24dd60f83f4c8acc60ea0056047a1d68b8837905

ccorcos:
awesome! It would be cool (if its not too hard to implement) if these stylesheets can be require'd in JS so you can load only the styles you need for a certain component...

trusktr:
@ccorcos @benjamn

> It would be cool (if its not too hard to implement) if these stylesheets can be require'd in JS so you can load only the styles you need for a certain component...

That's currently possible with JSS (the syntax being JS object notation). If styles are automatically applied to the page when they are `import`ed, as in

```js
import 'path/to/some-style.less'
```

then it won't be possible to selectively load styles because all imports will be loaded when the app starts running.

@benjamn it might be nice to be able to import a helper, so f.e.

```js
import someStyle from 'path/to/some-style.less'

// ...
    // apply it when needed:
    someStyle()
```

benjamn:
@trusktr yeah, having to call an imported function makes a lot of sense, and opens up other possibilities, like returning something useful from the function. Does JSS have a way to remove a set of styles from the page?

trusktr:
@benjamn Yep,

```js
import jss from 'jss'

let sheet = jss.createStyleSheet(...)

sheet.attach()
sheet.detach()
```

mist123:
Is there any milestone for this? 1.3? 2.0?
Can't wait to see this in Meteor.

benjamn:
@mist123 This stuff is definitely going into 1.3, though I anticipate 2.0 will be an opportunity to break backwards compatibility a bit, so (for example) you won't have to put your modules in an `imports/` directory to prevent them from being immediately evaluated.

trusktr:
@benjamn How can I test this? Can I specify some preview version with `meteor --release`?

stubailo:
Today, we had a meeting between the team working on the Meteor Guide, Ben, and some other folks, thinking about some edge cases we need to consider before we can safely write the guide using modules in Meteor 1.3. It would be great to get all of your feedback and see if there are any issues we didn't consider!

https://forums.meteor.com/t/ideas-from-mdg-meteor-1-3-the-meteor-guide-and-es2015-modules/13591

packages/geojson-utils/package.js
```diff
@@ -4,8 +4,10 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
+  api.use('modules');
   api.export('GeoJSON');
-  api.addFiles(['pre.js', 'geojson-utils.js', 'post.js']);
```
avital: Good riddance

packages/modules/modules.js
```diff
@@ -0,0 +1,27 @@
+meteorInstall = makeInstaller();
+
+meteorInstall({
+  node_modules: {
+    modules: {
+      "css.js": function (r, exports) {
```
avital: What is this? This lets users call `require("css.js").addStyles("...css text...")`? I couldn't find it used anywhere else in the code.

packages/modules/modules.js
```diff
@@ -0,0 +1,27 @@
...
+        exports.addStyles = function (css) {
+          var doc = document;
+          var style = doc.createElement("style");
+
+          style.setAttribute("type", "text/css");
+
+          if (style.styleSheet) {
```
avital: Can we add a comment explaining this if clause? It is a browser compatibility thing?

packages/modules/modules.js
```diff
@@ -0,0 +1,27 @@
...
+            style.styleSheet.cssText = css;
+          } else {
+            style.appendChild(doc.createTextNode(css));
+          }
+
+          return (
+            doc.head ||
```
avital: Same here. Browser compatibility?

tools/isobuild/compiler-plugin.js
```diff
@@ -91,7 +93,13 @@ export class CompilerPluginProcessor {
     var sourceProcessorsWithSlots = {};
 
     var sourceBatches = _.map(self.unibuilds, function (unibuild) {
+      const { pkg: { name }, arch } = unibuild;
+      const sourceRoot = name
```
avital: ah is this one of the places where `sourceRoot` matters? it's either read from a isopack or from a package directory on disk?

packages/modules/modules.js
avital: Found where it's used.

tools/isobuild/compiler-plugin.js
```diff
@@ -357,16 +368,36 @@ class ResourceSlot {
...
+    if (self.packageSourceBatch.usesModules()) {
+      const data = new Buffer(
+        'require("modules/css").addStyles(' + JSON.stringify(
+          files.convertToStandardLineEndings(options.data)
+        ) + ');\n'
+      );
+
+      self.jsOutputResources.push({
+        type: "js",
+        refreshable: true, // TODO Really?
```
avital: I couldn't find anywhere in the code that reads this property. Did I miss something?

tools/isobuild/compiler-plugin.js
```diff
@@ -487,6 +530,46 @@ class PackageSourceBatch {
...
+          self.importedSymbolToPackageName[symbol.name] = depUnibuild.pkg.name;
+        }
+      });
+
+      self.usedPackageNames[depUnibuild.pkg.name] = true;
+    });
+  }
+
+  usesModules() {
+    return _.isString(this.sourceRoot) &&
```
avital: What is this line for?

tools/isobuild/compiler-plugin.js
```diff
@@ -533,57 +616,26 @@ class PackageSourceBatch {
...
     const isApp = ! self.unibuild.pkg.name;
     const linkerOptions = {
       useGlobalNamespace: isApp,
+      sourceRoot: self.sourceRoot,
+      nodeModulesPath: self.unibuild.nodeModulesPath,
       // I was confused about this, so I am leaving a comment -- the
       // combinedServePath is either [pkgname].js or [pluginName]:plugin.js.
       // XXX: If we change this, we can get rid of source arch names!
-      combinedServePath: isApp ? null :
+      combinedServePath: isApp ? "/app.js" :
```
avital: What is this change about?

tools/isobuild/isopack-cache.js
```diff
@@ -118,6 +118,27 @@ _.extend(exports.IsopackCache.prototype, {
     });
   },
 
+  getSourceRoot(name, arch) {
+    const packageInfo = this._packageMap.getInfo(name);
+
+    if (packageInfo) {
+      if (packageInfo.kind === "local") {
```
avital: ah, so this is in a way the cause of the discrepancy between local and versioned packages in how they find files that are being imported.

tools/isobuild/package-api.js
```diff
@@ -336,6 +337,29 @@ _.extend(PackageAPI.prototype, {
...
+      const source = {
+        relPath: path,
+        fileOptions: {
+          mainModule: true
+        }
+      };
+
+      const oldMain = filesForArch.main;
+      if (oldMain) {
+        // It's not an error to call api.mainModule multiple times, but
```
avital: Why wouldn't we make this an error?

tools/isobuild/linker.js
```diff
@@ -145,32 +150,121 @@ _.extend(Module.prototype, {
...
+      // returned from this method in a singleton array.
+      servePath: self.combinedServePath,
+    };
+
     // Emit each file
-    _.each(self.files, function (file) {
-      if (!_.isEmpty(chunks)) {
-        chunks.push("\n\n\n\n\n\n");
+    if (self.useMeteorInstall) {
+      const tree = {};
```
avital: Can we put this clause in a separate function? This function is getting *very* long. (I know that this is definitely by far not the worst offender in tool code)

tools/isobuild/linker.js
```diff
@@ -145,32 +150,121 @@ _.extend(Module.prototype, {
...
+
+      _.each(self.files, function (file) {
+        if (file.bare) {
+          // Bare files will be added in between the synchronous require
+          // calls below.
+          return;
+        }
+
+        const parts = file.installPath.split("/");
```
avital: this code is very technical. i'm about to dig in to understand what it does but this should definitely have a comment above it in "human english" explaining what it does.

tools/isobuild/linker.js
avital: What is `tree`? I think there should be a comment.

tools/isobuild/linker.js
```diff
@@ -145,32 +150,121 @@ _.extend(Module.prototype, {
...
+          const isLastPart = i === parts.length - 1;
+          t = _.has(t, part)
+            ? t[part]
+            : t[part] = isLastPart ? file : {};
+        });
+      });
+
+      let moduleCount = 0;
+
+      function walk(t) {
```
avital: ditto, there should be a comment explaining this function

tools/isobuild/linker.js
```diff
@@ -145,32 +150,121 @@ _.extend(Module.prototype, {
...
+
+      if (moduleCount === 0) {
+        // If no files were actually added to the chunks array, roll back
+        // to before the `var require = meteorInstall(` chunk.
+        chunks.length = chunksLengthBeforeWalk;
+      }
+
+      // Now that we have installed everything in this package or
+      // application, immediately require the non-lazy modules and
+      // evaluate the bare files.
```
avital: I'm confused -- the original code didn't distinguish `bare` files specially, why do we do that now?

tools/isobuild/linker.js
```diff
@@ -241,9 +335,26 @@ var File = function (inputFile, module) {
   // produced by plugins)
   self.sourceHash = inputFile.hash || watch.sha1(self.source);
 
+  // The path of the source file, relative to the root directory of the
+  // package or application.
+  self.sourcePath = inputFile.sourcePath;
+
+  // Absolute module identifier to use when installing this file via
+  // meteorInstall.
+  self.installPath = inputFile.installPath || inputFile.sourcePath;
```
avital: When is `inputFile.installPath` falsey? Probably warrants a comment. (I haven't reviewed import-scanner.js yet)

tools/isobuild/js-analyze.js
```diff
@@ -13,6 +19,112 @@ function tryToParse(source) {
   }
 }
 
+var dependencyKeywordPattern = /\b(require|import|export)\b/g;
+
+export function findImportedModuleIdentifiers(source) {
```
avital: Please comment this function

tools/isobuild/js-analyze.js
```diff
@@ -13,6 +19,112 @@ function tryToParse(source) {
...
+    possibleIndexes.push(match.index);
+  }
+
+  if (!possibleIndexes.length) {
+    return {};
+  }
+
+  const ast = tryToParse(source);
+
+  function walk(node, left, right) {
```
avital: Please comment this function

packages/modules/modules.js
benjamn: Yeah that's the old-IE case: http://stackoverflow.com/a/524721/128454

packages/modules/modules.js
benjamn: In this case, I suppose `doc.getElementByTagName("head").item(0)` would work equally well in all browsers.

tools/isobuild/compiler-plugin.js
```diff
@@ -91,7 +93,13 @@ export class CompilerPluginProcessor {
     var sourceProcessorsWithSlots = {};
 
     var sourceBatches = _.map(self.unibuilds, function (unibuild) {
+      const { pkg: { name }, arch } = unibuild;
+      const sourceRoot = name
+        ? self.isopackCache.getSourceRoot(name, arch)
```
benjamn: Right, I added `self.isopackCache.getSourceRoot` to abstract over the differences between local/checkout/installed packages. Also note the `name` variable is `null` when we're bundling an app.

tools/isobuild/compiler-plugin.js
benjamn: I couldn't tell what the `refreshable` property does, either, though it was in the original code (preserved below, for the non-modules case).

tools/isobuild/compiler-plugin.js
benjamn: There are still some cases where we don't have a reliable `sourceRoot`, so this property will be `undefined`, such as when we're bundling a plugin.

tools/isobuild/compiler-plugin.js
benjamn: We use the `combinedServePath` in source maps, so it's wise to make sure it's always a string, but the precise value of the string doesn't affect anything besides source maps.

tools/isobuild/isopack-cache.js
benjamn: Yep!

tools/isobuild/js-analyze.js
benjamn: Will do.

tools/isobuild/js-analyze.js
benjamn: The basic idea is that `findImportedModuleIdentifiers` uses a regular expression to scan quickly for possible locations of certain tokens (`require`, `import`, `export`), then uses that location information to steer the AST traversal, so that it visits on subtrees that contain interesting tokens. The AST traversal determines if the tokens were actually what we thought they were (a `require` function call, or an `import` or `export` statement).

tools/isobuild/linker.js
benjamn: I'll add comments, but this `walk` function is similar to `JSON.stringify(tree)`, except the final result (`chunks`) is an array of strings and `SourceNode`s.

tools/isobuild/linker.js
benjamn: The output of `file.getPrelinkedOutput` is different now. In the original code, we could just output the results of `file.getPrelinkedOutput` for both `bare` and non-`bare` files, but now the output for non-`bare` files is a blob of `meteorInstall` code, so it's not possible to insert `bare` code in between the bundled modules. More importantly, `meteorInstall` doesn't actually evaluate any module code; that happens when we call `require`. If it's important for a `bare` file to be evaluated between two modules, then we need to output code that `require`s the first module, then code for the `bare` module, then code that `require`s the second module.

I would like to see `bare` files (and the `self.useGlobalNamespace` thing above) go away entirely, but that would probably break back compat.

tools/isobuild/linker.js
benjamn: Hopefully never. I'll add an assertion.

tools/isobuild/package-api.js
benjamn: I guess I was imagining you might do
```js
api.mainModule("main.js");
```
and then later
```js
api.mainModule("server_main.js", "server");
```
to "override" the main module for server code, leaving `main.js` as the entry point for `web.browser` and `web.cordova`.

I'll admit that kind of pattern isn't very compelling with only two or three architectures in play, but it might be a reasonable thing to do in the future?

trusktr:
@benjamn Will ES features still be handled by a separate `ecmascript` package in 1.3?

Also, what about ES2015 generators (I don't see them listed in `ecmascript`'s README) and all the ES2016 features like `async/await`? The advantage of using my own package `rocket:module` is that I've enabled those awesome things.

Maybe there can be a way to opt into the full set of features, based on ES stages (using Babel's new presets)?

AdamBrodzinski:
> Why invent Meteor.install instead of borrowing the bundle formats of Browserify or Webpack?

@benjamn  Will this fix the untenable reload times though? Using Webpack in Meteor works even from outside core and it reduces reloads from 17 seconds to 1.4 seconds :frowning: 

Don't get me wrong, adding ES6 modules is awesome but adding it onto a sandy foundation seems like a waste of your time.

trusktr:
17 seconds? What are you compiling? How big is your project? What build
plugins are you using? That seems ridiculously high. Mine is always like
1second or less with pure Meteor, and with webpack a little more (via
rocket:module).

/#!/JoePea

AdamBrodzinski:
> 17 seconds? What are you compiling? How big is your project? What build
plugins are you using? That seems ridiculously high. Mine is always like
1second or less with pure Meteor, and with webpack a little more (via
rocket:module).

I'm not sure if it's a package that was causing it or not... both are using React. I stopped building with the Meteor Sass package because it would crash with CI, so I switched to the NPM version and build outside. I think the only other build plugin was browserify for a couple of small npm modules.

Another project is on 1.1 and is taking around 9 seconds per reload. Hearing of your build times makes me want to try to re-upgrade to 1.2 again for this app to see if it's any different.

It's a ~30k LoC project with around 200 files (can't remember offhand). I tried using 1.2 but it was actually a second slower after the upgrade. So I swapped everything out to a pure webpack setup and made it essentially drop the bundle into the Meteor core folder on change (in dev mode it uses the dev server though).

bensmeets:
fwiw, I also get build much more in the direction of 17 seconds instead of 1 second :) @ 1.2 and even when having just created a brand new meteor project. It takes meteor at least 5 seconds to notice a file has been changed. Then the whole building even needs to start.

That's with approx 60 files. Nothing fancy. The blank meteor project has the same issue.

For the life of me, I can't understand why meteor keeps trying to reinvent the wheel :)

faceyspacey:
@benjamn 
>That's been exactly my experience. As much as I would like to use Webpack, and as much as I hope that a future version of Meteor will encourage a project structure that is easier for Webpack to consume, Meteor 1.3 has to be backwards compatible in a way that seems to forbid using Webpack or Browserify in their entirety (though reusing certain parts of them is a really good idea).

React overnight has--and even more so will become--the default view layer for Meteor. I can't bare to see the next bunch of months (until Meteor 2.0) be perfect minus one thing: hot code reloading. It will really put Meteor developers using React at a disadvantage compared to their NPM/React counterparts. In addition, if we did get this truly right, it would be one more thing to quiet the legions of frustrated Blaze developers. It's a key part of this whole story. It ties the story of the new Module system together with the new View system, React (and whatever official React-specific addons/mixins it gets).

Whatever we can do to make @trusktr and @eXon's packages seamless to integrate I think should be the least we can do. Currently--and I could be wrong--you can't use these packages with other packages that depend on `ecmascript`. i recently built a compiler plugin, and had to doubly do the work of `ecmascript`. So, if it's just a matter of making the plugin api support multiple compilers that can "consume" the file structure (per extension), that could be a low hanging fruit to make sure we dont have to wait until Meteor 2.0 until we are up to speed with the React world. Either way, now would be a very ripe time to facilitate better React integration.

ps.  After reading how thorough and *on top of it* all you guys are, I'm very happy to know this is in good hands.

faceyspacey:
@rozzzly 
>What if we took what universe-modules does, but flipped it around. That is, all the files in our app or packages (if specified of course) would be processed in the way that *.import.js are currently handled, ignored by the usual meteor load order. Then any of the files with a new *.entrypoint.js extension would be kept in the global scope, processed by the default compiler.

@benjamn 
>That's essentially my vision, too, except that I'm planning to treat any files inside an imports/ or node_modules/ directory as non-entry-points, while files outside those directories will be loaded the same way as before, according to usual Meteor load order. If you're really excited about modules, you can put your entire app into client/imports/ and server/imports/, and then have just one entry-point module per environment, perhaps client/main.js and server/main.js.

Awesome! Whatever we can do to move like monkey's from vine to vine will make these transitions so much easier.

AdamBrodzinski:
@trusktr I just upgraded to 1.2 and this time it wasn't slower but with .1 of a second (within the margin of me using a stopwatch). The app before and after was 11.5-12 seconds from hitting save to the change displaying on screen.

@benjamn Will the new module system help this at all?

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+  pathNormalize,
+  pathDirname,
+  pathBasename,
+  pathExtname,
+  statOrNull,
+  readFile,
+  convertToPosixPath,
+} from "../fs/files.js";
+
+export default class ImportScanner {
```
avital: Can there be a comment here explaining what this class does?

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+    this.usedPackageNames = usedPackageNames;
+    this.nodeModulesPath = nodeModulesPath;
+    this.absPathToOutputIndex = {};
+    this.outputFiles = [];
+    this.extensions = extensions;
+  }
+
+  addInputFiles(...args) {
+    const Ap = Array.prototype;
+    const flatFiles = Ap.concat.apply(Ap, args);
```
avital: Maybe just use `_.toArray`?

avital: Wait, what is this code even more? isn't `args` already an array at this point?

stubailo: Also, isn't there `Array.from` which does the same thing?

benjamn: This code was originally written to do the work of [this code](https://github.com/meteor/meteor/blob/ff64b73253694858ecbc0446b9d88e22cfcbe35d/tools/isobuild/compiler-plugin.js#L523-L528), but I never ended up changing `getResources`.

It may be worth noting that `Array.prototype.concat.apply(Array.prototype, array)` will flatten any nested arrays embedded in the given `array` (like the code in `getResources`), so it's definitely different from both `_.toArray` and `Array.from`.

benjamn: Oh wait, no, I remember now: I was trying to support both of the following styles:
* passing a single array of files to `addInputFiles`
* passing multiple file arguments to `addInputFiles`

I guess it's a sign of something that I forgot why this code was worthwhile… :blush:

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+    nodeModulesPath,
+  }) {
+    assert.ok(isString(sourceRoot));
+
+    this.name = name;
+    this.bundleArch = bundleArch;
+    this.sourceRoot = sourceRoot;
+    this.usedPackageNames = usedPackageNames;
+    this.nodeModulesPath = nodeModulesPath;
+    this.absPathToOutputIndex = {};
```
avital: This needs to be commented. What are "output indexes"?

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+      // in the bundle if they are actually imported.
+      file.lazy = this._isFileLazy(file);
+
+      // Files that are eagerly evaluated are effectively "imported" as
+      // entry points.
+      file.imported = ! file.lazy;
+
+      file.installPath = this._getInstallInfo(absPath).path;
+
+      if (has(this.absPathToOutputIndex, absPath)) {
```
avital: When does this happen? Don't we process every file once?

Or is this because someone may add the same file more than once when calling `addFiles`?

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+
+  _splitPath(path) {
+    const partsInReverse = [];
+    for (let dir; (dir = pathDirname(path)) !== path; path = dir) {
+      partsInReverse.push(pathBasename(path));
+    }
+    return partsInReverse.reverse();
+  }
+
+  _tryToResolveImportedPath(id, path) {
```
avital: I think this function, and honestly almost every function in this class, should have a comment.

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+
+    while (resolved && resolved.stat.isDirectory()) {
+      resolved = this._resolvePkgJsonMain(resolved.path) ||
+        this._joinAndStat(dirPath, "index.js");
+    }
+
+    return resolved && resolved.path;
+  }
+
+  _joinAndStat(...joinArgs) {
```
avital: This isn't really "join and stat", right? It's more something like "try various meanings of require paths, and return the result of stat while you're at it"?

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+    if (exactResult && exactStat.isDirectory()) {
+      // After trying all available file extensions, fall back to the
+      // original result if it was a directory.
+      return exactResult;
+    }
+
+    return null;
+  }
+
+  _resolveAbsolute(id) {
```
avital: I never knew you could do `require("/foo.js")` and honestly still am not sure what it does.

I've been looking at https://nodejs.org/api/modules.html#modules_all_together (which should also probably be linked to some place in this file) and can't tell what it's supposed to do. Is it the same as `require('./foo.js')`?

avital: (after reading the code)

so it loads from the top-level of the package or app? ...ok, i guess.

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+    extensions = [".js", ".json"],
+    usedPackageNames = {},
+    nodeModulesPath,
+  }) {
+    assert.ok(isString(sourceRoot));
+
+    this.name = name;
+    this.bundleArch = bundleArch;
+    this.sourceRoot = sourceRoot;
+    this.usedPackageNames = usedPackageNames;
```
avital: This needs to be commented.

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+    this.nodeModulesPath = nodeModulesPath;
+    this.absPathToOutputIndex = {};
+    this.outputFiles = [];
```
avital: What is the format of elements in this array? It should be documented.

tools/isobuild/import-scanner.js
glasser: I think you're misreading the confusing pseudocode at that link.

Specifically, when it says:

```
2. If X begins with './' or '/' or '../'
   a. LOAD_AS_FILE(Y + X)
```

I'm pretty sure that "Y + X" doesn't actually mean "concatenate those strings"; it means `path.resolve(Y, X)`.  ie, "resolve the path X as if you tried to open it with your current directory Y".  So if X is absolute, it literally just means that path X.

If you test, you can see that this is what `require("/foo.js")` does in normal Node: it literally opens `/foo.js`.

(Of course if you want you could make Meteor's require do something different for an absolute path, though that strikes me as a pretty bad idea.  If you desperately need a syntax for "relative to the app root", the syntax  I implemented for less/etc `@import` was `{}/foo.js`, though it's certainly not pretty.)

trusktr: What about an `app` keyword that represents the app, so  `import something from 'app/foo.js'`. That name can be reserved (no meteor package on Atmosphere can be called `app` (no MDG packes currently are)). On the same note, since if an import/require doesn't start with path symbols ('./' or '/' or '../') then the first identifier is considered to be a package name, why not allow Meteor package names here without the curlies, so f.e. `import Foo from 'some:package/path/to/Foo'` where some:package is a Meteor package.

trusktr: That idea is without having checked if colons are allowed in imports, but I don't think it matters since this is a custom implementation. The alternative would be to use isopack names (`some_package`) but it's not coherent.

avital: Do modules need to be able to import files from the app? As the code stands now, the app is always loaded so all globals are available, and app files that aren't in `imports/` don't use `export`/`module.exports`.

Regarding importing from another package, I *think* that works in the current implementation?

On the general note of calling `require` with an absolute path, my recommendation would be to just disallow that. The app will not work on other machines, etc. I don't see why we would support that feature.

trusktr: Good question! From the end user's point of view, packages are relied upon by the app, so it might make sense for the importing to be from app side if we intend to let users feel like they have control of their apps.

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+
+      if (bundlingClientApp && (dir === "server" ||
+                                dir === "private")) {
+        // If we're bundling an app for a client architecture, any files
+        // contained by a server-only directory that is not contained by
+        // a node_modules directory must be ignored.
+        return;
+      }
+
+      if (dir === "node_modules") {
```
avital: This clause is to support something like `require("node_modules/bar")`? Which does the same as `require("./bar")`? Does anyone do that? Is this just to support the same way that Node's `require` works?

tools/isobuild/import-scanner.js
```diff
@@ -0,0 +1,431 @@
...
+      }
+
+      const installInfo = this._getInstallInfo(absImportedPath);
+      if (! installInfo) {
+        // The given path cannot be installed on this architecture.
+        return;
+      }
+
+      if (! installInfo.inNodeModules) {
+        // At this point, if the file is not in a node_modules directory,
```
avital: I'm struggling to understand this. Maybe I just don't have the right mental model of this class.

If in a package you call `mainModule` or `addFiles` on just a single file, and that file imports another file that's not in a `node_modules` directory, then will this code ignore that import? Or maybe what gets passed to `addInputFiles` in this class is all of the JS files in the package?

tools/isobuild/import-scanner.js
benjamn: You rarely see `node_modules` in an imported/required module identifier, because the way to import something from a `node_modules` directory is simply to `require("bar")` *without* a `./` or `../`.

This whole loop is attempting to determine whether a module should be allowed to be installed. This clause specifically relaxes all the restrictions above for files in `node_modules` directories, because you wouldn't want an npm package to break because it put some of its modules in a `server/` or `packages/` directory.

avital:
@benjamn Finally done with a first pass on the review!

tools/isobuild/import-scanner.js
benjamn: An important concept in this implementation is that `meteorInstall` does not have to preserve the exact structure of files and directories from the developer's file system. Instead, `meteorInstall` has the opportunity to define a simplified directory structure that allows the usual Node lookup algorithm to work as expected, without a bunch of special cases. You can think of `_getInstallInfo` as a method that maps real file system paths to normalized virtual `.installPath`s for use with `meteorInstall`. Sometimes `_getInstallInfo` will reject a path, which just means that module won't be installed by `meteorInstall`, and can't be imported at runtime (for example, any file in a `server/` directory when we're building a bundle for the client).

Given the current implementation, the virtual directory structure expressed via `meteorInstall` might look something like this:
```
<root>
├── app # same for all apps, allows apps to have their own local node_modules directory
│   ├── eager1.js # a file that will be required immediately when the app starts
│   ├── eager2.js
│   ├── imports
│   │   ├── lazy1.js # a file that must be imported
│   │   └── lazy2.js
│   └── node_modules
│       └── moment # a local npm package installed by the app developer
└── node_modules
    ├── pkg1 # a Meteor package with two npm dependencies
    │   └── node_modules
    │       ├── npm1 # in reality, the source files for these npm packages live in some .npm or 
    │       └── npm2 # isopacket directory, but _getNodeModulesInstallInfo normalizes them thus
    ├── pkg2 # a Meteor package with one npm dependency
    │   └── node_modules
    │       └── npm3
    └── pkg3.js # generated stub for a Meteor package not using modules: module.exports = Package.pkg3
```

What to I mean by special cases? For example, the actual source files for that `npm3` dependency might reside in different places depending on whether the `pkg2` Meteor package was installed from a checkout, was in a local `packages/` directory, or came from Atmosphere, so it's vitally important that `meteorInstall` always installs `pkg2` as `/node_modules/pkg2/` and the `npm3` dependency as `/node_modules/pkg2/node_modules/npm3/`.

Given this structure, it is technically possible to `import "/app/imports/lazy2"`, though I would be willing to consider disabling support for absolute (`/`) identifiers, which would effectively force all imported identifiers to be either relative (`./` or `../`) or top-level (i.e., referring to something in `node_modules`).

Even if we forbid absolute identifiers, it would still be possible for packages to use `../../app/` relative identifiers to import things from app code. If we wanted to forbid that, we could depart from the Node lookup algorithm by inventing a new special case, or give the `/app` directory a randomized name, but I don't think either of those options is really worth the added complexity. Node developers can explore the entire file system using relative identifiers, but no one is so foolish as to do that in real code.

benjamn: By the way, when/if Meteor ever moves in the direction of saying apps are just Node projects and packages are simply `node_modules` installed in those projects, the virtual directory structure described above will very likely become the actual directory structure.

glasser: I'd suggest disallowing absolute paths, then, as you suggest (rather than making them do something distinct from what Node usually does), to help the potential transition towards "just Node projects" instead of creating some new behavior to preserve.

tools/isobuild/import-scanner.js
trusktr: That would let users run `npm install` anywhere in their app, which sounds like a really good idea if you're aiming for the broadest support for NPM within a Meteor app. This will let people use their favorite workflows (or none at all) while still ultimately having a Meteor app! :+1:

tools/isobuild/import-scanner.js
trusktr: As long as Atmosphere is around, the ideal for me would be to only have to specify entrypoint files in a package's package.js. Other files that are ever imported would need to be detected automatically and given to users via Atmosphere (requiring some static analysis). A temporary solution would be to just ship all files (which is less efficient, but will at least work)

tools/isobuild/import-scanner.js
trusktr: > I'd suggest disallowing absolute paths

What about absolute paths just starting from the app root and a simple documentation about this?

trusktr: Will users use their own copy of npm?

benjamn: For now, yeah, you'd be responsible for getting stuff into `node_modules/` in whatever way made sense for you. Probably using your own `npm`, but maybe also with `git clone` or something like that.

In the future, maybe `meteor npm` could be a shortcut for `/path/to/meteor/dev_bundle/bin/npm` so you don't have to have your own version of `npm` installed in order to use it.

tools/isobuild/import-scanner.js
benjamn: > Or maybe what gets passed to addInputFiles in this class is all of the JS files in the package?

Yes, that's what happens now, thanks to this commit: https://github.com/benjamn/meteor/commit/21a0780194bd9b40ad384b80ae66262a010d9174

As @trusktr points out, this may be less efficient, but from what I can tell most packages don't include (m)any files that were never added with `api.addFiles`, so removing those unnecessary calls to `api.addFiles` and adding all files automatically shouldn't be that much more expensive than what we did before.

tools/isobuild/import-scanner.js
trusktr: Cool! I like the flexibility of this! Can I use a preview version of this via the `meteor` cli?

trusktr: Can I specify a local version of Meteor via the cli?

stubailo: The best way to use Meteor locally is to make an alias in your shell to run the `meteor` binary at the root of the repo.

trusktr: Sweeet. :+1:

benjamn: We should probably do the same for `meteor node` (i.e., make it an alias for `/path/to/meteor/dev_bundle/bin/node`) while we're at it…

trusktr: I wonder what people would use it for?

/#!/JoePea

trusktr:
An idea: maybe `require('/absolute/path')` should always by deafult import relative to the root of the context. For example, if in app code, that would import relative to the app root. If in package code, that would import relative to the package root. Optionally, package names could be used for cross-package imports, but for security, I think there should be a way for a package to specify that files can be exported from outside.

So, `require('some:package/path/to/file')` will only work if that package has allowed that in package.js.

But, I guess, for security, someone should be careful what things their packages export anyways. This brings up a question: should we even worry about cross package importing with Meteor packages (and thus encouraging Atmosphere packages), or just start encouraging NPM packages and not worry about cross-package importing for Meteor packages?

trusktr:
Another idea: we can extend package.json, including a field that Meteor recognizes, that can be used for a package to specify which files shouldn't be importable outside of the package. That doesn't exist with NPM as far as I know, and could even serve as an idea for NPM. It'd be nice to control what people can and can no import from a library (for example, allowing only public APIs to be imported).

tools/isobuild/import-scanner.js
avital: @benjamn and I talked about this, we should just make it be a single argument that's an array.

vjau:
FYI, path modules resolution in Typescript 1.8 (next version)
https://github.com/Microsoft/TypeScript/issues/5039

timothyarmes:
Hi,

This sounds great. I have a question concerning the replication of modules. I currently use cosmos:browserify to include my nam modules, and using browserify we can replace the `require` statements with other code so that we can avoid including chucks of code (such as the react runtime!) multiple times. 

Will your system allow this too?

Tim

JeremyBYU:
Thanks your for the work. I have been trying out Meteor 1.3 and I really like it. I thought I should post some of the issues I have been facing using meteor 1.3 beta with new import syntax. You can see my git repo here for a better explanation [link](https://github.com/JeremyBYU/meteor-import)

#### NPM install must be called before launching meteor first time

Be sure to use npm install before you launch meteor. Npm will install modules into node_modules folder.
There seems to be a bug where the build cache in .meteor/local never receives the packages from node_modules during refresh. The only way to fix it is to delete the .meteor/local folder to see changes in node_modules. Probably a bug that needs to be reported or fixed.
Using Windows 10....
#### Meteor react package vs npm react
Couldnt get npm react to work (react installed from npm).  I think it needs browserify or something.  Kept complaining about process.env stuff. I just used the official meteor react package.

#### NPM React packages needing reference to react
```
global = {};
global.React = React;
```
You need this code to load before anything. A react component was looking for react.  It checks to find it in ```global.React```. Since Meteors official react package uses the global variable ```React``` this is pretty easy.

#### Referencing NPM React Components
I dont get this but I had to do full path to reference the Formsy react component in my custom components. My custom react components are in client/imports.
```
let Formsy = require('../../node_modules/formsy-react');  
```
Really annoying, if someone can figure out a cleaner way that would be great.  

#### Some NPM React packages just dont work
Some of the npm packages I wanted to use just didn't work.  I wanted to use React Pure css components.  Didnt work. Complained about babel not being installed as a dependency....yada yada

trusktr:
> A react component was looking for react. It checks to find it in global.React

You should file an issue in the component's tracker so they can hopefully fix that. :)

> ```
> let Formsy = require('../../node_modules/formsy-react');  
> ```

That might possibly be a bug with Meteor 1.3. It should definitely know what you mean by `require('formsy-react')`. Have you tried ES6 `import` syntax? DOes that change the result?

JeremyBYU:
ES6 Import syntax doesnt work in jsx files (for Meteor 1.3 beta). However I read in some other thread that there is a fix on the way for that.

trusktr:
@JeremyBYU Good question! https://forums.meteor.com/t/does-the-ecmascript-package-support-jsx/9830/5

zimt28:
[Good news](https://github.com/jashkenas/coffeescript/issues/3162#issuecomment-176545595) for CoffeeScript users; looks like we'll be able to use ES2015 module syntax without backticks :+1:

PR #5501, opened 2015-10-22
Update the account username docs

greggman:
Note: I'm a complete meteor noob but as far as I can tell this field no longer exists or certainly not as documented

For one the none of the `accounts-xxx` packages ask for a username. [`accounts-password`](https://atmospherejs.com/meteor/accounts-password) only asks for email and password, no username is set. [`accounts-facebook`](https://atmospherejs.com/meteor/accounts-facebook) sets username but does no check for uniqueness.

Even if this is a bad PR I hope you'll consider the docs seem to be out of date. Maybe someone that knows it better can submit a better PR.

### DIFF
--- a/docs/client/basic/sections/accounts.md
+++ b/docs/client/basic/sections/accounts.md
@@ -46,7 +46,7 @@ user document:
 ```
 {
   _id: "bbca5d6a-2156-41c4-89da-0329e8c99a4f",  // Meteor.userId()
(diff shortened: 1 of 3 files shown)


stubailo:
I don't think this is correct. First, usernames are certainly expected to be unique. Second, you can add a validator to make it required.

I agree that the accounts docs aren't great, but the modification here isn't the right one.

PR #5516, opened 2015-10-22
Adding new Facebook user-agent to spiderable

AlexeyMK:
According to Facebook
(https://developers.facebook.com/docs/sharing/webmasters/crawler), they
have a new (May 2014) crawler named Facebot. Spiderable doesn't
currently work for Facebot. It ought to.  

PS. I shed a tear along with @stuabilo that this is even a thing.

### DIFF
--- a/packages/spiderable/spiderable_server.js
+++ b/packages/spiderable/spiderable_server.js
@@ -14,6 +14,7 @@ var urlParser = Npm.require('url');
 // here. I shed a silent tear.
 Spiderable.userAgentRegExps = [
   /^facebookexternalhit/i,
+  /^Facebot/,
   /^linkedinbot/i,
   /^twitterbot/i,
   /^slackbot-linkexpanding/i


stubailo:
Goodness gracious. Is there a way to test that it works?

AlexeyMK:
It's a third party integration :-/. I think we may just have to take
Facebook at their word.

I thought about putting together a test for the user agent white listing,
but given that there weren't tests for any of the other user agent regexs,
I figured, meh.

We could move the list of user agent regexs into a config file or
something, so you could ostensibly eventually load them from some
repository of "user agents to allow", but this feels like overkill.

Also, if looks like Google is slowly deprecating the approach that makes
spiderable work (
https://developers.google.com/webmasters/ajax-crawling/docs/getting-started?hl=en),
so really the answer is to finish server side rendering by 2016.

stubailo:
Well, I guess this is the best we can do right now.

PR #5535, opened 2015-10-25
Add 'meteor console' as an alias for 'meteor shell'

mbrookes:
Accept "muscle memory" `console` as valid command for those coming from a Rails background.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -388,14 +388,14 @@ main.registerCommand(_.extend(
 ///////////////////////////////////////////////////////////////////////////////
 // shell
 ///////////////////////////////////////////////////////////////////////////////
-
-main.registerCommand({
-  name: 'shell',
+var shellCommandOptions = {
   requiresRelease: false,
   requiresApp: true,
   pretty: false,
   catalogRefresh: new catalog.Refresh.Never()
-}, function (options) {
+};
+
+function doShellCommand(options) {
   if (!options.appDir) {
     Console.error(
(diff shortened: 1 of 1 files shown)


stubailo:
I don't know how we feel about aliases for commands, I'm not sure if there are possible complications down the road when names of commands don't map one-to-one with the actual command, for example if someone is writing a wrapper script or similar. What do you think @benjamn?

sebakerckhof:
One problem of aliases is that meteor guides and tutorials might use different terms which can confuse new users. Maybe aliases could be used just to give hints what the correct command is.
E.g. somebody types; `meteor console` and gets a message: `Did you mean meteor shell?`

mbrookes:
@sebakerckhof - that's a fair point (albeit fairly unlikely for a hidden alias), but your solution make sense. @stubailo - rework or close?

stubailo:
Yeah I think we should do the hint that tells you to call meteor shell instead.

day:
I agree if a command doesn't exist, it shouldn't just be silently aliased. However, 'console' was certainly the more obvious choice *for me* as well (admittedly also coming from a Rails background). Since (strictly speaking) a console is piece of hardware, and a shell is a command-line interpreter, 'shell' is the more correct. So, +1 for the hint compromise on this.

stubailo:
Any news?

mbrookes:
Sorry Sashko, got caught up in other things. I have submitted #5593, so will close this.

PR #5540, opened 2015-10-26
Safety check for Cordova apps hanging on splash screen #4038

markoshust:


markoshust:
Are there any reasons why this build/tests failed?

markoshust:
Ah... it appears this type of logic was already built into the refactored code. I'll close this ticket out.

PR #5545, opened 2015-10-26
Improve error message when value is null.

zacharydenton:
Currently, when `value` is null, the error message is (for example) `Expected string, got object`. This changes the error message to `Expected string, got null`, which aids debugging.

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -152,7 +152,7 @@ var testSubtree = function (value, pattern) {
       if (typeof value === typeofChecks[i][1])
         return false;
       return {
-        message: "Expected " + typeofChecks[i][1] + ", got " + typeof value,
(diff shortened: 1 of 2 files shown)


stubailo:
Makes sense to me! I hope people aren't parsing the error output here, then it would technically be a breaking change.

It would also be great to have a test for this special case, could you add that?

zacharydenton:
@stubailo I added some tests for the `check` error messages (https://github.com/zacharydenton/meteor/commit/a7d90f6b59c84512f01dc7ec25f886a975b730fc). Looks like there's a problem with CI #3 today though.

stubailo:
Wow, awesome! Thanks for all the tests. Super great work. This is the best kind of PR.

PR #5550, opened 2015-10-27
Correct History for 1.2.1

rgoomar:
Both Mac and Linux are based off Unix, but Mac is not based off Linux, therefore, you can run it on a `non-Unix` machine rather than a `non-Linux` machine.


rgoomar:
Actually, looking at it again, if it is only relevant on non-Linux, close this PR. I may have mis interpreted it.

stubailo:
No, this literally means "Linux", since Galaxy runs Linux. Good catch, though!

Issue #5096, opened 2015-09-02
Only add collection methods if allow/deny is being used


zimt28:
**_1 Upvote**_ When creating a collection `new Mongo.Collection('posts')`, Meteor will create some methods automatically: `posts/insert`, `posts/update`, `posts/remove`, maybe some more. Please prefix these methods somehow, like `_meteor/posts/insert`; as creating a `posts/insert` method for a `posts` collection seems to be the right name (that's why you're using it), it shouldn't be blocked.

stubailo:
Couldn't agree more!

As a workaround, I've been using `add` instead of `insert` for my own methods.

stubailo:
@zimt28 in a very late response, would it work for you to simply not have these methods if you're not using allow/deny for that collection?

mitar:
+1

zimt28:
@stubailo This would work great for me!

mquandalle:
I also support collection methods customization as @stubailo started to implement in #5559.

One use case I encountered in Wekan is that I wanted to define a set of *mutations* at the model level (instead of dispersing minimongo queries all over the code base) and at the same time benefit from re-usable security checks à la allow/deny. Currently I ended up defining my mutations like this:

https://github.com/wekan/wekan/blob/dd101543920148a48b74541ebbbaca97f6bc6d11/models/boards.js#L124-L213

Which I am pretty happy with. I have some ideas to improve this API (and the underlined DDP method calls) and the abstraction of `CollectionPrototype._defineMutationMethods` will be super helpful!

mquandalle:
Well I read a little bit too quickly, `CollectionPrototype._defineMutationMethods` isn’t the kind of abstraction I hoped it was, but still the possibility to remove and replace allow/deny mechanism is great.

stubailo:
Yeah I'm like 5% into implementing this stuff - the first step is taking exactly what exists now and refactoring a bit, the second step will be adding new APIs to decouple allow/deny and collections.

stubailo:
> abstraction I hoped it was

What are you hoping for?

mquandalle:
It’s just that from reading the name I thought it was a function to define methods to mutate the collection, something that would that would put a DDP method under a dedicated “namespace” and expose the collection object to this method, and then things like Mongo  `insert`, `update`, and `delete` or even things like `rename` or `changeColor` like my above example in Wekan could rely on this.

stubailo:
You can do this easily in user-space, though, right?

```js
[...]

PR #5559, opened 2015-10-28
WIP: Move allow/deny entirely into a separate package and refactor insert/update/remove

stubailo:
As a first start to making it possible to use collections without allow/deny, I'm working on splitting out the code. This is a first step to solving #5096, and in general will be good for reducing the size of the `mongo` package. Consider that all of the code I'm moving was in _one file_, called `collection.js`.

In order to enable this, I needed to refactor `insert`/`update`/`remove` so that I could actually understand the control flow of those functions, which I think is a benefit in itself.

This should result in zero user-facing changes; the next PR will have some of the changes related to making it possible to not use allow/deny at all, or enable it per-collection.

@avital to review.

### DIFF
--- a/packages/allow-deny/allow-deny.js
+++ b/packages/allow-deny/allow-deny.js
@@ -1,0 +1,493 @@
+///
+/// Remote methods and access control.
+///
+
+// Restrict default mutators on collection. allow() and deny() take the
+// same options:
+//
+// options.insert {Function(userId, doc)}
+//   return true to allow/deny adding this document
+//
+// options.update {Function(userId, docs, fields, modifier)}
+//   return true to allow/deny updating these documents.
+//   `fields` is passed as an array of fields that are to be modified
+//
+// options.remove {Function(userId, docs)}
+//   return true to allow/deny removing these documents
+//
+// options.fetch {Array}
+//   Fields to fetch for these validators. If any call to allow or deny
+//   does not have this option then all fields are loaded.
+//
+// allow and deny can be called multiple times. The validators are
+// evaluated as follows:
+// - If neither deny() nor allow() has been called on the collection,
+//   then the request is allowed if and only if the "insecure" smart
+//   package is in use.
+// - Otherwise, if any deny() function returns true, the request is denied.
+// - Otherwise, if any allow() function returns true, the request is allowed.
+// - Otherwise, the request is denied.
+//
+// Meteor may call your deny() and allow() functions in any order, and may not
+// call all of them if it is able to make a decision without calling them all
+// (so don't include side effects).
+
+AllowDeny = {
+  CollectionPrototype: {}
+};
+
+// In the `mongo` package, we will extend Mongo.Collection.prototype with these
+// methods
+const CollectionPrototype = AllowDeny.CollectionPrototype;
+
+/**
+ * @summary Allow users to write directly to this collection from client code, subject to limitations you define.
+ * @locus Server
+ * @method allow
+ * @memberOf Mongo.Collection
+ * @instance
+ * @param {Object} options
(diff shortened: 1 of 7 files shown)


packages/allow-deny/allow-deny.js
```diff
@@ -0,0 +1,493 @@
...
+AllowDeny = {
+  CollectionPrototype: {}
+};
+
+// In the `mongo` package, we will extend Mongo.Collection.prototype with these
+// methods
+const CollectionPrototype = AllowDeny.CollectionPrototype;
+
+/**
+ * @summary Allow users to write directly to this collection from client code, subject to limitations you define.
```
avital: 80c (I can't even see the whole line in GitHub this way)

packages/allow-deny/allow-deny.js
```diff
@@ -0,0 +1,493 @@
...
+ * @param {Function} options.insert,update,remove Functions that look at a proposed modification to the database and return true if it should be allowed.
+ * @param {String[]} options.fetch Optional performance enhancement. Limits the fields that will be fetched from the database for inspection by your `update` and `remove` functions.
+ * @param {Function} options.transform Overrides `transform` on the  [`Collection`](#collections).  Pass `null` to disable transformation.
+ */
+CollectionPrototype.allow = function(options) {
+  addValidator(this, 'allow', options);
+};
+
+/**
+ * @summary Override `allow` rules.
```
avital: I don't know if you're looking for feedback on the docs, but I don't understand the description of `deny` as "override `allow` rules". Or maybe there is an additional sentence missing or something?

avital: Oh. This is just copied over from the old collection.js.

packages/allow-deny/allow-deny.js
avital: Oh! This was also just copied over... Please ignore

packages/mongo/collection.js
```diff
@@ -437,6 +432,77 @@ var throwIfSelectorIsNotId = function (selector, methodName) {
...
+  }
+
+  // On inserts, always return the id that we generated; on all other
+  // operations, just return the result from the collection.
+  var chooseReturnValueFromCollectionResult = function (result) {
+    if (doc._id) {
+      return doc._id;
+    }
+
+    // XXX what is this for??
```
avital: What came out of this, I thought we talked about this and you figured out some way around it?

packages/mongo/collection.js
```diff
@@ -461,149 +574,54 @@ var throwIfSelectorIsNotId = function (selector, methodName) {
...
+    return;
+  }
 
-    // both sync and async, unless we threw an exception, return ret
-    // (new document ID for insert, num affected for update/remove, object with
-    // numberAffected and maybe insertedId for upsert).
-    return ret;
-  };
-});
+  if (convertResult) {
```
avital: It would be a tad shorter to do `convertResults = convertResults || _.identity` (or even can you do that as a default argument value in ES5?)

Anyhow, not important enough to change now, it just came to mind.

avital: (And then get rid of the `if` clause)

avital:
LGTM! Two small comments, I wouldn't block on those to merge

packages/mongo/collection.js
stubailo: Yeah in the end I don't think my solution actually worked.

packages/mongo/collection.js
stubailo: Interesting - I actually think the way it is now is more readable, rather than using `_.identity`. But I'll change it!

PR #5561, opened 2015-10-29
Adds Meteor.stubMethods and Meteor.unstubMethods

Sanjo:
### Intro

This addition enables developers to unit test code that call Meteor methods in a simple way.

The pull request adds two methods: `Meteor.stubMethods` and `Meteor.unstubMethods`. Both are intended to be used in client unit test code. `Meteor.stubMethods` ca replaced the method implementation a method with a fake implementation.

### Example

The added unit tests in this pull request show how use the API.

In addition here is an example of an scenario where you would use those methods.

App code:
```js
App.Services.MyService = {
  myMethod: function() {
    Meteor.call('myMethod', function(error, result) {
      if (!error) {
        Session.set('theResult', result);
      } else {
        Session.set('theResult', 'Error!');
      }
    });
  }
}
```

Test:
```js
describe('MyService', function() {
  describe('myMethod', function() {
    afterEach(function() {
      Meteor.unstubMethods(['myMethod']);
    });

    it('sets the result in the session when there is no error', function() {
      Meteor.stubMethods({
        'myMethod': function () {
          return 'noErrorResult';
        });
      });
      spyOn(Session, 'set');
      App.Services.MyService.myMethod();
      expect(Session.set).toHaveBeenCalledWith('theResult', 'noErrorResult');
    });

    it('does not set the result in the session when there is an error', function() {
      Meteor.stubMethods({
        'myMethod': function () {
          throw new Error('Oh no');
        });
      });
      spyOn(Session, 'set');
      App.Services.MyService.myMethod();
      expect(Session.set).toHaveBeenCalledWith('theResult', 'Error!');
    });
  });
});
```

### Epilogue

This is one step into the direction of making Meteor app testing simple. Your constructive feedback is welcomed ;-)


stubailo:
How do you call a mocked method?

Sanjo:
> How do you call a mocked method?

Exactly as you call a normal Meteor method (with Meteor.call or Meteor.apply). Mocking a method is something that you would do in the unit test setup. We will create an example to illustrate the usage and will post it here.

samhatoum:
> How do you call a mocked method?

Something like this:

```javascript
App.Services.MyService = {
  aMethod: function() {
    Meteor.call('myMethod', function(error, result) {
      if (!error) {
        Session.set('theResult', result);
      } 
    });
  }
}
```

```javascript
describe('My Service', function() {
  afterEach(function() {
    Meteor.unstubMethods(['myMethod']);
  });
  it('sets the result in the session when is no error', function() {
    Meteor.stubMethods({
      'myMethod': function () {
        return 'noErrorResult';
      });
    });
    spyOn(Session, 'set');
    App.Services.SomeService.aMethod();
    expect(Session.set).toHaveBeenCalledWith('theResult', 'noErrorResult');
  });
  it('sets does not set the result in the session when there is an error', function() {
    Meteor.stubMethods({
      'myMethod': function () {
        throw new Error('Oh no');
      });
    });
    spyOn(Session, 'set');
    App.Services.SomeService.aMethod();
    expect(Session.set).toHaveBeenCalledWith('theResult', 'noErrorResult');
  });
});
```

Sanjo:
@stubailo I have added an example in the initial post and improved the description.

packages/ddp-client/livedata_connection.js
```diff
@@ -426,6 +427,78 @@ _.extend(MethodInvoker.prototype, {
   }
 });
 
+// A MethodMockInvoker calls a mock implementation for a method without
```
stubailo: I know this isn't consistent with the style of the rest of the file (since it needs to be heavily refactored, IMO) but can we put this in a new file in the package?

Also, it would be cool to use ES2015 class syntax here, seems like a perfect spot for it!

stubailo:
I quite like this idea - is there any info/documentation I can read about what people are doing to test methods now and how this makes it easier, or what hacks it can replace? Curious what the alternative possible designs are.

Sanjo:
>  is there any info/documentation I can read about what people are doing to test methods now and how this makes it easier, or what hacks it can replace? Curious what the alternative possible designs are.

Just to make sure, that you understand it correctly: This pull request will not improve testing methods. Testing methods can be done by using Meteor.call. This pull request will improve testing method result handling code (the callback) for client side code.

Other solutions to this problem are:
* Move as much code out of the handler function as possible. I don't think that introducing indirection just to make the code testable is ideal.
* Replace the Meteor.call implementation with code that calls the callback immediately with the wanted result and error in your test. This achieves the same goal. But it is not as simple to use as Meteor.stubMethods.

stubailo:
> This pull request will not improve testing methods. Testing methods can be done by using Meteor.call. This pull request will improve testing method result handling code (the callback) for client side code.

Right, thanks for clarifying. I definitely had the wrong idea.

Let me put it a different way: as someone who isn't well-acquainted with testing, is there some other source that can back up that this PR is a good idea? I think the technical implementation is good, but I don't trust myself to make that decision without more information. For example, does @samhatoum think we should accept this?

Sanjo:
Had another idea. To make it clearer that those APIs are for testing, we could expose them as `Meteor.Testing.stubMethods` and `Meteor.Testing.unstubMethods`. This namespace can also contain other futures testing helpers.

stubailo:
Hmm actually the only thing that really needs to be in the Method code itself is the thing that decides between calling the real vs. mocked method, right? Could the rest be in a different file/package?

mitar:
Hm, why exactly does this have to be in the core? I have done thins like this in the past:

```javascript
Meteor.methods({
  foo: wrapMethod('foo', function () {
     ...
  }
});
```

mitar:
And then `wrapMethod` can make mocking and stuff?

samhatoum:
@stubailo 
> does @samhatoum think we should accept this?

Yes. To answer your question about how it is done otherwise, we currently we do this:

```javascript
var _methodNameToResponseMap = {}
VelocityHelpers.stubMethod = function (methodName, error, result) {
  _methodNameToResponseMap[methodName] = {error: error, result: result}
  spyOn(Meteor, 'call').and.callFake(function (methodName) {
    var callback = arguments[arguments.length - 1]
    var response = _methodNameToResponseMap[methodName]
    callback(response.error, response.result)
  })
}
```
and then use it like this:
```javascript
VelocityHelpers.stubMethod('user/createAndLogin', null, 'someId');
```
This works in Jasmine, but it does not work for any other testing package. We would need to write another version for Sinon (used in Mocha), another for Cucumber etc. 

@mitar 
> why exactly does this have to be in the core?

For the reasons above, and for the reason that @Sanjo said above:

> I don't think that introducing indirection just to make the code testable is ideal.

When the core has support for testing, it means we write less code in our app, which in turn means less bugs!

stubailo:
OK, @Sanjo can you address my code comment about moving the code into a different file?

Sanjo:
@stubailo Yes, doing it now. Can I introduce the Meteor.Testing namespace?

stubailo:
Why not just `Testing`? Is the problem that it _would_ be `Methods.Testing.stub`, but there's no `Methods` namespace?

Sanjo:
Yeah, `Testing` works even better.

mitar:
Should it then be in a separate package, only testing package? Maybe added to some of existing core testing packages?

Sanjo:
I would create a new package named `testing` that introduces the `Testing` namespace. Similar to the `meteor` package role. This package can be `debugOnly: true` I think, so the testing helpers are not shipped in the production build.

Maybe I could also move the method stubbing code into a package `ddp-client-testing` and make the `testing` package imply the `ddp-client-testing` package.

tmeasday:
So good to see this appearing guys. Can we do the same to `Meteor.subscribe` please? ;)

stubailo:
Yeah this all seems like a good idea @Sanjo!

tmeasday:
I think our recommendation is to use ValidatedMethod or a similar wrapper to achieve this. A `Meteor.methods()` global that isn't imported is kind of crappy IMO.

@Sanjo do you still like this pattern?

PR #5562, opened 2015-10-29
Allow the other types of URLs that npm supports.

tmeasday:
See for instance #844 for people running into the
fact that our lack of support for git protocols means
dealing with private npm packages is a pain.

### DIFF
--- a/tools/utils/utils.js
+++ b/tools/utils/utils.js
@@ -486,7 +486,7 @@ exports.isUrlWithSha = function (x) {
 exports.isNpmUrl = function (x) {
   // These are the various protocols that NPM supports, which we use to download NPM dependencies
   // See https://docs.npmjs.com/files/package.json#git-urls-as-dependencies
-  return exports.isUrlWithSha(x) || /^(git|git\+ssh|git\+http|git\+https)?:\/\/.*#/.test(x);
+  return exports.isUrlWithSha(x) || /^(git|git\+ssh|git\+http|git\+https)?:\/\//.test(x);
 };
 
 exports.isPathRelative = function (x) {
--- a/tools/utils/utils.js
+++ b/tools/utils/utils.js
@@ -477,10 +477,16 @@ exports.isUrlWithFileScheme = function (x) {
 };
 
 exports.isUrlWithSha = function (x) {
-  // These are the various protocols that NPM supports, which we use to download
-  // both NPM and Cordova dependencies
+  // Is a URL with a fixed SHA? We use this for Cordova -- although theoretically we could use
+  // a URL like isNpmUrl(), there are a variety of problems with this, 
+  // see https://github.com/meteor/meteor/pull/5562
+  return /^https?:\/\/.*[0-9a-f]{40}/.test(x);
+}
+
+exports.isNpmUrl = function (x) {
+  // These are the various protocols that NPM supports, which we use to download NPM dependencies
   // See https://docs.npmjs.com/files/package.json#git-urls-as-dependencies
-  return /^(https|git|git\+ssh|git\+http|git\+https)?:\/\/.*[0-9a-f]{40}/.test(x);
+  return exports.isUrlWithSha(x) || /^(git|git\+ssh|git\+http|git\+https)?:\/\/.*#/.test(x);
 };
 
 exports.isPathRelative = function (x) {
@@ -493,21 +499,21 @@ exports.isPathRelative = function (x) {
 //
 // This is talking about NPM/Cordova versions specifically, not Meteor versions.
 // It does not support the wrap number syntax.
-exports.ensureOnlyExactVersions = function (dependencies) {
+exports.ensureOnlyExactVersions = function (dependencies, {forNpm}) {
   _.each(dependencies, function (version, name) {
     // We want a given version of a smart package (package.js +
     // .npm/npm-shrinkwrap.json) to pin down its dependencies precisely, so we
     // don't want anything too vague. For now, we support semvers and urls that
     // name a specific commit by SHA.
-    if (! exports.isExactVersion(version)) {
+    if (! exports.isExactVersion(version, {forNpm})) {
       throw new Error(
         "Must declare exact version of dependency: " + name + '@' + version);
     }
   });
 };
-exports.isExactVersion = function (version) {
-  return semver.valid(version) || exports.isUrlWithSha(version)
-    || exports.isUrlWithFileScheme(version);
+exports.isExactVersion = function (version, {forNpm}) {
+  return semver.valid(version) || exports.isUrlWithFileScheme(version)
+    || (forNpm ? exports.isNpmUrl(version) : exports.isUrlWithSha(version));
 };
 
 
--- a/tools/utils/utils.js
+++ b/tools/utils/utils.js
@@ -499,21 +499,21 @@ exports.isPathRelative = function (x) {
 //
 // This is talking about NPM/Cordova versions specifically, not Meteor versions.
 // It does not support the wrap number syntax.
-exports.ensureOnlyValidVersions = function (dependencies, {forNpm}) {
+exports.ensureOnlyValidVersions = function (dependencies, {forCordova}) {
   _.each(dependencies, function (version, name) {
     // We want a given version of a smart package (package.js +
     // .npm/npm-shrinkwrap.json) to pin down its dependencies precisely, so we
     // don't want anything too vague. For now, we support semvers and urls that
     // name a specific commit by SHA.
-    if (! exports.isValidVersion(version, {forNpm})) {
+    if (! exports.isValidVersion(version, {forCordova})) {
       throw new Error(
         "Must declare valid version of dependency: " + name + '@' + version);
     }
   });
 };
-exports.isValidVersion = function (version, {forNpm}) {
+exports.isValidVersion = function (version, {forCordova}) {
   return semver.valid(version) || exports.isUrlWithFileScheme(version)
-    || (forNpm ? exports.isNpmUrl(version) : exports.isUrlWithSha(version));
+    || (forCordova ? exports.isUrlWithSha(version): exports.isNpmUrl(version));
 };
 
 
--- a/tools/utils/utils.js
+++ b/tools/utils/utils.js
@@ -493,25 +493,25 @@ exports.isPathRelative = function (x) {
   return x.charAt(0) !== '/';
 };
 
-// If there is a version that isn't exact, throws an Error with a
+// If there is a version that isn't valid, throws an Error with a
 // human-readable message that is suitable for showing to the user.
 // dependencies may be falsey or empty.
 //
 // This is talking about NPM/Cordova versions specifically, not Meteor versions.
 // It does not support the wrap number syntax.
-exports.ensureOnlyExactVersions = function (dependencies, {forNpm}) {
+exports.ensureOnlyValidVersions = function (dependencies, {forNpm}) {
   _.each(dependencies, function (version, name) {
     // We want a given version of a smart package (package.js +
     // .npm/npm-shrinkwrap.json) to pin down its dependencies precisely, so we
     // don't want anything too vague. For now, we support semvers and urls that
     // name a specific commit by SHA.
-    if (! exports.isExactVersion(version, {forNpm})) {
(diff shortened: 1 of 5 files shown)


tmeasday:
Martijn, you asked to have a play with this on cordova.

martijnwalraven:
I tried using these URL schemes with Cordova, but I'm running into some trouble.

For http/https I get the following error:
```
fatal: Unable to find remote helper for 'git+https'
```
This [Stack Overflow question](http://stackoverflow.com/questions/31322495/git-fatal-unable-to-find-remote-helper-for-githttps) suggests that may actually be a bug in npm, but I'm not sure if that is what is going on here.

Using ssh seems to be accepted, but with `meteor add cordova:phonegap-facebook-plugin@git+ssh://github.com/Wizcorp/phonegap-facebook-plugin.git#5dbb1583168558b4447a13235283803151cb04ec` I get this as a result:
```
   Permission denied (publickey).
   fatal: Could not read from remote repository.
   
   Please make sure you have the correct access rights
   and the repository exists.
```

The issue is probably that our current plugin specification syntax does not allow specifying a username, because we already use @ to separate the plugin ID and version or URL. So `meteor add cordova:phonegap-facebook-plugin@git+ssh://git@github.com:Wizcorp/phonegap-facebook-plugin.git#5dbb1583168558b4447a13235283803151cb04ec` is not allowed. 

We could change the code in [`tools/cli/commands-packages.js`](https://github.com/meteor/meteor/blob/devel/tools/cli/commands-packages.js#L1871) to deal with this, but I'm afraid the syntax will be confusing to humans as well :) SO ideally I'd like to get rid of the plugin ID there. The Cordova CLI doesn't require it to be specified either. But that does mean we should download/install the plugins as part of the `meteor add` command, instead of just adding them to `cordova-plugins`. What do you think?

As a separate issue, do you still plan on allowing branches and tags in addition to SHAs?

tmeasday:
Hey @martijnwalraven - thanks for having a look. Let's see now:

1. Reading the linked npm issue, it sounds like `git+https://` requires a careful URL format. What was the URL you were using?

2. I don't really understand the implications of the Cordova thing well enough to speak to it, but I suspect people will want to be able to do this. Alternatively we could change the separator we use on the command line from `@`?  `cordova:phonegap-facebook-plugin:git+ssh...` would make sense to me also.

3. Re branches and tags -- I am scared of the implications but @glasser seems to want us to do it, so I guess I should just do it. I was held back slightly by the semantics of the method changing (ie. calling it `isExactVersion()` wouldn't really be right anymore and the error messages should change). 

  I'd be happy to spend some time doing what makes sense to me if you're happy to review/sanity check it.

martijnwalraven:
1. I think I was using `meteor add cordova:phonegap-facebook-plugin@git+https://github.com/Wizcorp/phonegap-facebook-plugin.git#5dbb1583168558b4447a13235283803151cb04ec` (can't test it right now, I'm at the airport). Definitely not a relative URL as seems to be suggested by the npm issue.

2. I don't think replacing `@` by `:` really improves the situation, and it would lead to more confusion if we change this now. The easiest solution is probably for me to fix the code to deal with the double `@` and leave the larger changes until later.

3. Yeah, I would also need to change the error messages in the Cordova code, but I think it is worth it. Only allowing SHAs seems unnecessarily restrictive. Are you planning on changing this for Meteor 1.3?

tmeasday:
> Are you planning on changing this for Meteor 1.3?

I don't really have any plans around this but I'm happy to take a look at the npm side of it.

mitar:
@martijnwalraven: I think you are confusing things. This pull request is for `Npm.depends`, not for Meteor package dependencies.

mitar:
I tried with:

```
Npm.depends({
  'xxx': 'git+ssh://github.com/private/xxx.git#fabe272133ef6f4c6bec39c38f320aa6dfb25fc3'
});
```

But it fails with `Permission denied (publickey)`. Running the same command from my shell:

```
npm install 'git+ssh://github.com/private/xxx.git#fabe272133ef6f4c6bec39c38f320aa6dfb25fc3'
```

Succeeds. Is it possible that Meteor command line tool is not running under the same user as I am otherwise developing?

mitar:
Hm, but:

```
"./meteor/dev_bundle/bin/npm" "install" "git+ssh://github.com/sunblaze-ucb/node-secureworker.git#c12aaeeec6ef4df448354eac1d2"
```

fails with no access to my key.

mitar:
OK, this works. I had to add `git@` for username.

```
Npm.depends({
  'xxx': 'git+ssh://git@github.com/private/xxx.git#fabe272133ef6f4c6bec39c38f320aa6dfb25fc3'
});
```

Can this be released soon. :-)

tmeasday:
@martijnwalraven 

1. that URL worked fine (at least up to the point of trying to find a `package.json`) when I added it as an `Npm.depends` URL, FWIW.

2. Yes, you are right, let's try and get the double `@` thing working

3. I will try to do this tomorrow.

martijnwalraven:
Something else I've been thinking about is that if we accept branches and tags in addition to SHAs, we then need a way to update these dependencies. Not sure how that would work for `Npm.depends`, but with Cordova plugins there is currently no explicit mechanism for this, and we only update plugins when the URL changes.

wh0:
why do we restrict what protocols are allowed?

tmeasday:
@martijnwalraven - to your last point, I did some minor investigations regarding npm packages + branches:

1. Once you get a branch to load, an `npm-shinkwrap.json` file is created (with the exact commit that's current for the branch) and subsequent builds of the project will use this exact version (even if you e.g. use `meteor reset` locally). From my perspective this is great as it means we get repeatable builds between members of the team (assuming they check this file in) and for deployment.

2. If you delete that `npm-shinkwrap.json` file, the project will rebuild, and fetch the latest version from the branch.

So, it *is* possible to update the dependency, although of course it's in no ways "nice". Unless there's some other command that I don't know about that would do this for me.

It seems to me that loosening things and getting this behaviour is strictly better than the status quo (even if it's not perfect), so I'm inclined to say we should do it. But it depends on the situation w/ Cordova. What happens there?

(You can loosen the requirement by simply changing `isUrlWithSha` to `  return /^(https|git|git\+ssh|git\+http|git\+https)?:\/\//.test(x);`)

martijnwalraven:
Hmmm, I'm afraid the behavior with Cordova would be confusing. Basically, adding a plugin (or package with a plugin dependency) install the version pointed at by the specified branch/tag as expected. But there is no shrinkwrap mechanism, so someone else installing that same plugin at a later time might get a different version. Furthermore, we reinstall all plugins when we make any changes to the list of plugins, so adding or removing another plugin might also install a different version. And removing/adding a plugin would also be the only way to force updating the installed version (besides removing `.meteor/local/cordova-build` completely).

Then again, vanilla Cordova has no plugin updating mechanism either, and has the same issues with non-repeatable builds. So maybe these inconsistencies are acceptable?

tmeasday:
But if I understand correctly, you don't typically repeatedly build your Cordova project, right? - as in someone builds it once and then checks in the built output (so in a sense, the "shinkwrapped" version). From then on all developers would be using the same built version until someone updates the Cordova build.

This is in contrast to the Meteor Cordova world, where it basically rebuilds from scratch on each machine that runs it. (Please correct me if I have this wrong).

The above seems really bad. I don't personally feel comfortable with a situation where you and I can have the exact same thing checked out from git and be running different code.

mitar:
Can we get this merged in for NPM stuff and for Cordova you can discuss it more later? ;-)

tmeasday:
@martijnwalraven do you think we can split the npm stuff from the cordova stuff? The easier approach would be to merge this as-is (without the commit-ish support).

tmeasday:
@martijnwalraven I committed some changes that split things up. Can you take a look when you have a moment?

tools/utils/utils.js
```diff
@@ -457,9 +457,16 @@ exports.isUrlWithFileScheme = function (x) {
...
+  // Is a URL with a fixed SHA? We use this for Cordova -- although theoretically we could use
+  // a URL like isNpmUrl(), there are a variety of problems with this, 
+  // see https://github.com/meteor/meteor/pull/5562
   return /^https?:\/\/.*[0-9a-f]{40}/.test(x);
+}
+
+exports.isNpmUrl = function (x) {
+  // These are the various protocols that NPM supports, which we use to download NPM dependencies
+  // See https://docs.npmjs.com/files/package.json#git-urls-as-dependencies
+  return exports.isUrlWithSha(x) || /^(git|git\+ssh|git\+http|git\+https)?:\/\/.*#/.test(x);
```
martijnwalraven: Should we also allow URLs without a SHA for `http(s)`? Could we just add `https?` to the protocols instead of using `isUrlWithSha`? 

Now that we we allow more flexible URLs, is there any reason still to require an explicit branch instead of using the default one?

tools/utils/utils.js
```diff
@@ -472,21 +479,21 @@ exports.isPathRelative = function (x) {
...
+    if (! exports.isExactVersion(version, {forNpm})) {
       throw new Error(
         "Must declare exact version of dependency: " + name + '@' + version);
     }
   });
 };
-exports.isExactVersion = function (version) {
-  return semver.valid(version) || exports.isUrlWithSha(version)
-    || exports.isUrlWithFileScheme(version);
+exports.isExactVersion = function (version, {forNpm}) {
```
martijnwalraven: Now that we no longer require a SHA in `isNpmUrl`, does `isExactVersion` still make sense? Or should we call it something like `isValidVersion`? (Coming to think of it, `isUrlWithFileScheme` never guaranteed an exact version.)

martijnwalraven: This may be bikeshedding, but would it make sense to turn the flag around and use `forCordova` instead of `forNpm`? Strictly taken, Cordova also uses NPM, so ultimately it is a URL to be passed to NPM.

tmeasday: I think these are both fair points.

tools/utils/utils.js
tmeasday: > Should we also allow URLs without a SHA for http(s)?

I guess npm says "You may specify a tarball URL in place of a version range".

But one nice property of the various `#commit-ish` URLs is that npm stores the actual commit hash in the shinkwrap file so you get consistent builds. If we allow arbitrary changeable[1] URLs, we lose that.

So I'd vote against it. Plus I don't recall people asking for it.

> Could we just add https? to the protocols instead of using isUrlWithSha?

No, because the `#commit-ish` thing is a property of `git+` URLs. (and the above)

> Now that we we allow more flexible URLs, is there any reason still to require an explicit branch instead of using the default one?

I'm just following npm's format here: https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

[1] Ok in theory a URL that includes a hash is changeable but let's not get pathological.

tmeasday: Hmm I guess it says "the default is master" So maybe we don't need it. Ok, I'll remove the requirement for `#`.

tmeasday:
@martijnwalraven I made the changes discussed above and it appears to work. Let me know

tools/utils/utils.js
martijnwalraven: Ah, I didn't get the difference between `https://` and `git+https://`. With Cordova, you use `https://` to download from Git, so that actually supports `#commit-ish` URLs.

(So it allows `https://github.com/apache/cordova-plugin-console.git#f055daec45575bf08538f885e09c85a0eba363ff` or `https://github.com/apache/cordova-plugin-console.git#r0.2.0`).

martijnwalraven:
@tmeasday Looks great, I think this is ready to be merged!

tmeasday:
Hmm, looks like I made some self-tests fail by changing the error message to say "valid version" rather than "exact version". I'm not quite sure what's best here..

martijnwalraven:
I think ''valid version" makes more sense now, so you could just change the expectations in the self test?

tmeasday:
Oh, I need to merge devel into this. Should I rebase this on top of devel or merge devel in here?

martijnwalraven:
Rebasing is probably cleaner, that's what I usually do.

tmeasday:
Ok, force pushed. I guess let's double check the test pass and then.. who merges? You or me?

martijnwalraven:
I'll leave the honor to you :)

PR #5575, opened 2015-10-30
Allow piping commands to `meteor shell` via STDIN

benjamn:
These changes make it possible to use `meteor shell` for single-command evaluation, with only the command output echoing to `stdout`. Errors are printed to `stderr`, and the `meteor shell` process returns a non-zero error code.

Example:
```js
> meteor create asdf
> cd asdf
> meteor &
> echo Meteor.isServer | meteor shell
true
```

cc @avital @glasser

### DIFF
--- a/tools/shell-client.js
+++ b/tools/shell-client.js
@@ -81,9 +81,44 @@ Cp.connect = function connect() {
   });
 };
 
+Cp.setUpSocketForSingleUse = function (sock, key) {
+  sock.on("connect", function () {
+    const inputBuffers = [];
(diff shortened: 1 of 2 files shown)


benjamn:
Since this hasn't drawn any objections, and it works, and the changes are confined to the `meteor shell` code, I'm going to go ahead and merge it.

deiangi:
Don't you normally pipe in the other way around? 

$meteor shell | echo Meteor.isServer
expected$ true

deiangi:
and btw that is an awesome feature. what version supports that?

PR #5585, opened 2015-11-01
Make Script More Portable With #!/usr/bin/env

TomFreudenberg:
To make the scripts more compatible on different linux environments, shebang should be changed in general to `#!/usr/bin/env bash` instead `#!/bin/bash`

### DIFF
--- a/docs/canonically_diff_docs.sh
+++ b/docs/canonically_diff_docs.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
 echo "Please run this script from the docs app folder."
 echo "Make sure you have phantomjs installed!"
 
--- a/examples/unfinished/benchmark/run-local.sh
+++ b/examples/unfinished/benchmark/run-local.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 PORT=9000
 if [ -z "$NUM_CLIENTS" ]; then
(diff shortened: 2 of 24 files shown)


TomFreudenberg:
Hmm, I checked the errors on `ci/circleci`. I am currently not able to see if this is something which comes through my changes or are invalid in `devel` branch currently.

If you guide me, I will test and provide a patch if something is wrong in this PR.

stubailo:
Looks great. Apparently there is some history here: https://github.com/meteor/meteor/pull/1148

But we're going to accept it now since you're doing some great work around making Meteor run on different platforms.

TomFreudenberg:
Hi Sashko ( @stubailo ) thanks ... I am sorry hadn't seen that PR before but it is great that you accepted and merged.

stubailo:
Would it be possible to elaborate here on which environments this is necessary for? Just for historical value.

TomFreudenberg:
Hi Sashko, yes, I will write it down here. I started a compatible check issue and will go thru a number of distros to check where meteor / meteor universal will run. For those are popular and not yet official supported, we will provide a dev_bundle on bintray - We like the idea of meteor everywhere :-).

TomFreudenberg:
Here are the first one's

FreeBSD, OpenBSD, NetBSD

TomFreudenberg:
Next: busybox / debootstrap

tangorri:
So this mean I can try to compile/run rocketchat on freebsd now ?

PR #5586, opened 2015-11-01
Polling observer driver - Allow to customize polling Interval

okland:
Allow users to change polling default interval per cursor, Important when using polling instead of oplog.
For example: Collection.find({}, {_disableOplog:true, _pollingInterval:4000}) so the pooling will happen at least once in  4 sec.


zol:
Thanks @okland - this seems reasonable.

PR #5592, opened 2015-11-02
Split minifiers in css and js minifiers

sebakerckhof:
Splits the `minifiers` and `standard-minifiers` package into separate css and js minifiers.
`standard-minifiers` is now an umbrella package that implies both minifiers.


stubailo:
Going to review this as soon as I have time, please let me know if I don't in the next 2-3 days.

stubailo:
Looks good to me! Hopefully this makes it easier for people to write custom minifiers!

stubailo:
Note about tests: all of the tests that fail here also fail on `devel`, so I chose to ignore them.

PR #5601, opened 2015-11-03
Fix HTTP.call to not explode on undefined content

### LINKED ISSUE #5565: http_common - content may be undefined
This line caused my application to crash. `content` was undefined and threw error because `.toString()` cannot be called on `undefined`

https://github.com/meteor/meteor/blob/devel/packages/http/httpcall_common.js#L8

To Fix:
```javascript
var contentToCheck = typeof content == "string" || typeof content == 'undefined' ? content : content.toString();
```

stubailo:
Fixes #5565


stubailo:
I should add a test!

PR #5602, opened 2015-11-03
Check for the existence of a keyValueDep in ReactiveDict clear

### LINKED ISSUE #5530: Reactive Dict errors on clearing values that have never been retrieved
Hi There

Found a weird one, I replicated in code, and have replicated simply in the chrome dev console so I don't believe a demo is necessary as its incredibly easy to replicate, but I can set one up if necessary.

With Reactive Dict included in your project doing the following in chrome dev tools

````js
var thing = new ReactiveDict();
thing.set('one', 'some value');
thing.clear();
````

produces this error : 

````js
reactive-dict.js:166 : Uncaught TypeError: Cannot read property '"some value"' of undefined(…)
(anonymous function)	@	reactive-dict.js:166
_.each._.forEach	@	underscore.js:142
_.extend.clear	@	reactive-dict.js:164
(anonymous function)	@	VM10168:2
InjectedScript._evaluateOn	@	VM9632:875
InjectedScript._evaluateAndWrap	@	VM9632:808
InjectedScript.evaluate	@	VM9632:664
````

If however you use the 'get'' prior to calling `.clear()` this does not occur, though calling .all() does not seem to stop the error for occurring when clearing afterwards.

stubailo:
Fixes #5530

From @sprohaska's comment: https://github.com/meteor/meteor/issues/5530#issuecomment-151186970

### DIFF
--- a/packages/reactive-dict/reactive-dict-tests.js
+++ b/packages/reactive-dict/reactive-dict-tests.js
@@ -40,6 +40,12 @@ Tinytest.add('ReactiveDict - clear() works', function (test) {
   var dict = new ReactiveDict;
   dict.set('foo', 'bar');
 
+  // Clear should not throw an error now
+  // See issue #5530
+  dict.clear();
+
+  dict.set('foo', 'bar');
+
   var val, equals, equalsUndefined, all;
   Tracker.autorun(function() {
     val = dict.get('foo');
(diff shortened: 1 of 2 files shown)


stubailo:
Should add a test for this.

PR #5622, opened 2015-11-06
Case insensitive email lookup in forgotPassword method

### LINKED ISSUE #5621: Accounts.forgotPassword() is case sensitive for email-based requests to reset an account password
I am finding that `Accounts.forgotPassword` is case sensitive for email-based requests.

For example, when a user registered as **tester@tester.com** tries an Accounts.forgotPassword request with **Tester@tester.com**, the response from the server is `User not found`.

This seems odd to me because the login process is case *insensitive* when using email as the username, i.e. that user could login with **TeSTeR@tEstEr.Com** with no problem. But the `Accounts.forgotPassword()` seems to be applying different rules.

Is this the expected behavior?

Originally posted [here in the forum](https://forums.meteor.com/t/is-accounts-forgotpassword-case-sensitive-for-email-based-accounts/12926).

edit: Pretty sure this is related to [issue 550](https://github.com/meteor/meteor/issues/550)

zimme:
Should fix #5621. Fix not verified.


zimme:
I don't think the failing tests was my fault, seems like the tests couldn't run properly

martijnwalraven:
Thanks for getting a start on fixing this! I'd be happy to merge a fix, but there is some work remaining so it would be great if you wanted to take this on:
- With your change, `forgotPassword` now passes the whole `options` object to `_findUserByQuery`. It checks whether it contains `email`, but it may contain additional fields. For instance, if you also pass `options.id` or `options.username`, these will take precedence and be used to look up the user instead. So you probably want to make sure to pass an `options` object with only an `email` key.
- Could you add a test to `password_tests.js`, and make sure it fails without this fix and passes with it?

Let me know if you don't feel like working on this or are unable to do so, so we can add it to our own to do list.

PS: You're right about the failing tests, we're having some issues with our CI setup.

zimme:
I passed the whole `options` object because there's a `check(options, {email: String});` before making sure the object only contains `email`, as it's not `Match.ObjectIncluding` so no other values are allowed.

I can add a test that checks against the same email with a different case then the one used to create the user.

martijnwalraven:
Ah, you're absolutely right, I misinterpreted the `check`! That test sounds good, so if you could add that, I'll merge the fix. Thanks again!

zimme:
Something that hit me while working on this... it would have been so much easier to just add do this.
```js
user = {
  username: 'zimme',
  usernameLowerCase: 'zimme',
  emails: [{
    address: 'Simon.R.K.Fridlund@gmail.com',
    addressLowerCase: 'simon.r.k.fridlund@gmail.com',
    verified: true,
  },
}
```

martijnwalraven:
There was actually quite a bit of discussion about this (see https://github.com/meteor/meteor/issues/550). Among other things, we wanted the solution to be backwards compatible and we didn't want to open up the possibility of the two fields getting out of sync.

zimme:
I can see that being a concern. What I did now was adding the email in it's original form to the "resetToken" so I didn't have to use the regex/permutations selector to lookup the email when doing the resetPassword method. if you feel that's the wrong approach, feel free to change whatever you think is needed.

martijnwalraven:
I need to take a closer look at this, and I'll probably want to make some changes, but I won't get to this until the end of the week.

zimme:
GLHF =)

martijnwalraven:
I ran into trouble trying to modify the pull request directly, so I had to create a separate branch: https://github.com/meteor/meteor/tree/case-insensitive-forgot-password

I've cleaned up some of the test code to get rid of email variables shared between tests, and have made sure `forgotPassword` respects the passed in email and doesn't just use the first address in the user's `emails` list. Also, `Accounts.sendResetPasswordEmail` continues to expect a case sensitive email address (as do similar server functions).

Because this code has potential security implications, I'm going to ask a team member to do a quick review, but I expect it to be merged to devel soon.

zimme:
Tracked in #5674

PR #5649, opened 2015-11-11
Don't leak internal paths

mitar:


stubailo:
What do you mean by leak? Presumably, that log logs to the server, and the server knows about all of the paths? I'm not sure what the before and after of this PR is based on this title and code change.

mitar:
Hmmm. I think it went to the `curl` output, so to the request. But now that I am looking at this I must admit I do not see how this would happen. :-) Hm, `Log.error` is not send to the client ever, maybe in debug mode?

stubailo:
I don't think it's ever sent anywhere. The only possibility that comes to mind is maybe Kadira does something with logs? But like, if server-side logs are leaked to the client then that would be a massive, massive security bug, and one we don't have as far as I know.

mitar:
Yea.

Issue #5630, opened 2015-11-07
Improve `check` package to call Match.Where function only once


bySabi:
The function on Match.Where is called two time is not fail.

```file.js
checkFn = Match.Where(function(x) {
  console.log('checkFn');
  return true;
})

check('', checkFn);
```

output this: 

```
=> App running at: http://localhost:3000/
I20151107-17:19:27.946(1)? checkFn
I20151107-17:19:27.946(1)? checkFn
```

bySabi:
I recreate more clear with this example:

```
checkFn = Match.Where(function(x) {
    console.log('checkFn');
    return true;
});

function callmeOne(arg) {
  console.log('start callmeOne');

  check(arg, checkFn);
  console.log('end callmeOne');
}

callmeOne('value');
```

the output is:

```
=> App running at: http://localhost:3000/
I20151108-20:39:28.198(1)? start callmeOne
I20151108-20:39:28.199(1)? checkFn
I20151108-20:39:28.199(1)? checkFn
I20151108-20:39:28.200(1)? end callmeOne
```

`Match.test` function is afected too.

bySabi:
Created a repo https://github.com/bySabi/twice-match-where-call
Using meteor 1.2.1 on OSX.

stubailo:
I don't think enforcing the number of times `check` calls each function is part of its contract, so this isn't really a bug. Do you somehow have a very complex validation function?

bySabi:
Thanks for take time in this @stubailo.
I you read this forum thread https://forums.meteor.com/t/match-where-on-check-function-called-two-times-why/13005  your release than this "mis behaviour" is not there before 1.2.1 release.

I have a validation function that log input inside Match.Where. Now I have duplicate message logs cause this. The doc not same nothing that number os check calls is not guaranteed. Is not guaranteed the numbers of call by contract then in any moment this number can be 2 or even more, in some use cases this could have serious performace penalty all over the place.  What is the point on make more than one check on same "check"? 
If check enforcing is not quaranteed we must move to a guaranteed aproach like?
`if (check_this(args)) throw Meteor.Error('ERROR!')`

stubailo:
> check enforcing is not quaranteed

It's definitely guaranteed that it will work. But I agree that if it was only called once in 1.2 and is now twice in 1.2.1, that's a regression in performance. I think it was from this PR: https://github.com/meteor/meteor/pull/4584

I'd love to take another PR to fix it to only get called once.

[...]

PR #5651, opened 2015-11-11
fix Match.Where function called twice bug

bySabi:
Hi @stubailo I found the problem on `check` package related to issue:  https://github.com/meteor/meteor/issues/5630
Please review this PR it


stubailo:
Looks awesome!

Thanks for catching this.

PR #5663, opened 2015-11-16
Make sure it's clear that DDPRateLimiter is on the server

stubailo:
Technically, this is a "breaking change" since code that is loaded
on client and server will now throw an error when calling DDPRateLimiter,
but the API didn't do anything on the client anyway so it's probably
more like a bug fix.


stubailo:
Tests are in progress, @avital can you give this a quick look?

avital:
:+1:

PR #5667, opened 2015-11-17
Replace all uses of Future in tools/ code with equivalent Promise-based code

benjamn:
This branch has been sitting around since May 2015. After rebasing it against https://github.com/meteor/meteor/pull/4619, I decided I'd rather not rebase it again if I can help it.

Note that many of these commits use `Promise.prototype.await` to yield the current `Fiber` until the `Promise` is resolved, which (like `Future.prototype.wait`) allows the code creating the `Promise` to avoid returning the `Promise` to the caller. This works great in an environment with `Fiber`s, but it might be better, if this code was being written from scratch, to stick to `Promise.prototype.then` or `async` functions and `await` expressions, so that the calling code can decide how to orchestrate the asynchronous operations.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -6,7 +6,6 @@ var buildmessage = require('../utils/buildmessage.js');
 var auth = require('../meteor-services/auth.js');
 var authClient = require('../meteor-services/auth-client.js');
 var config = require('../meteor-services/config.js');
-var Future = require('fibers/future');
 var runLog = require('../runners/run-log.js');
 var utils = require('../utils/utils.js');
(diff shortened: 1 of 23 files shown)


benjamn:
Now that all the integration tests are passing, I'm inclined to merge this by end of day tomorrow if there are no objections. However, since there are a lot of changes, and any bugs introduced might be subtle (e.g., differences in timing between `Future` and `Promise` callbacks), I think it's probably best *not* to squash the commits, in case we need to diagnose problems using `git bisect`.

cc @dgreensp @martijnwalraven @glasser @avital

PR #5668, opened 2015-11-17
Fix docs to reflect EJSON format instead of Any

### LINKED ISSUE #5657: Meteor.publish doesn't recognize methods of an object passed as an argument in subscribe call
Server (coffee)
```coffee
Meteor.publish 'posts', (args)->
 userId = @userId
 Posts.find(cityId:args.cityId(userId))
```
Client
```coffee
obj=
  cityId:(userId)->Users.findOne(userId).city
Meteor.subscribe 'posts', obj
```
Throws `Exception from sub posts id WcuyuY6eWNBxFw8yK TypeError: Object #<Object>
 has no method cityId` in server console.

rgoomar:
- This is to avoid any confusion of passing in any argument when it is checking for an EJSONable value
- Fixes #5657

### DIFF
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -472,7 +472,7 @@ _.extend(Connection.prototype, {
    * @locus Client
    * @param {String} name Name of the subscription.  Matches the name of the
    * server's `publish()` call.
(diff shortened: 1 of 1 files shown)


stubailo:
Thanks Rishi!

PR #5674, opened 2015-11-17
Add case insensitive email lookup to forgotPassword function

martijnwalraven:
This is a continuation of https://github.com/meteor/meteor/pull/5622, but I couldn't find a way to modify the pull request directly, so I had to create a separate branch.

I've cleaned up some of the test code to get rid of email variables shared between tests, and have made sure `forgotPassword` respects the passed in email and doesn't just use the first address in the user's `emails` list. Also, `Accounts.sendResetPasswordEmail` continues to expect a case sensitive email address (as do similar server functions).

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -500,11 +500,16 @@ Accounts.setPassword = function (userId, newPlaintextPassword, options) {
 Meteor.methods({forgotPassword: function (options) {
   check(options, {email: String});
 
-  var user = Meteor.users.findOne({"emails.address": options.email});
+  var user = Accounts._findUserByQuery({email: options.email});
   if (!user)
     throw new Meteor.Error(403, "User not found");
 
-  Accounts.sendResetPasswordEmail(user._id, options.email);
+  const emails = _.pluck(user.emails || [], 'address');
+  const caseSensitiveEmail = _.find(emails, email => {
+    return email.toLowerCase() === options.email.toLowerCase();
+  });
+
+  Accounts.sendResetPasswordEmail(user._id, caseSensitiveEmail);
 }});
 
 // send the user an email with a link that when opened allows the user
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -500,7 +500,7 @@ Accounts.setPassword = function (userId, newPlaintextPassword, options) {
 Meteor.methods({forgotPassword: function (options) {
   check(options, {email: String});
 
-  var user = Accounts._findUserByQuery({email: options.email});
+  var user = Accounts.findUserByEmail(options.email);
   if (!user)
     throw new Meteor.Error(403, "User not found");
 
--- a/packages/accounts-base/package.js
+++ b/packages/accounts-base/package.js
@@ -43,7 +43,7 @@ Package.onUse(function (api) {
   api.addFiles('accounts_common.js', ['client', 'server']);
   api.addFiles('accounts_server.js', 'server');
 
-  api.addFiles('accounts_rate_limit.js');
+  api.addFiles('accounts_rate_limit.js', 'server');
   api.addFiles('url_server.js', 'server');
 
   // accounts_client must be before localstorage_token, because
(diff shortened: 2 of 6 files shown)


packages/accounts-password/password_server.js
```diff
@@ -500,11 +500,16 @@ Accounts.setPassword = function (userId, newPlaintextPassword, options) {
 Meteor.methods({forgotPassword: function (options) {
   check(options, {email: String});
 
-  var user = Meteor.users.findOne({"emails.address": options.email});
+  var user = Accounts._findUserByQuery({email: options.email});
```
zimme: Sounds reasonable.

packages/accounts-password/password_server.js
```diff
@@ -500,11 +500,16 @@ Accounts.setPassword = function (userId, newPlaintextPassword, options) {
...
   if (!user)
     throw new Meteor.Error(403, "User not found");
 
-  Accounts.sendResetPasswordEmail(user._id, options.email);
+  const emails = _.pluck(user.emails || [], 'address');
+  const caseSensitiveEmail = _.find(emails, email => {
```
zimme: The reason I put this logic inside of `sendResetPasswordEmail` was so that when a developer from server-side code manually are using `sendResetPasswordEmail` doesn't have to care about finding and using the caseSensisiveEmail.

Which they will have to do now, otherwise the `resetPassword` method will fail, as it expects the email in the resetToken to be of the same case.

martijnwalraven: My reasoning was that the ability to specify a case insensitive email is a convenience, but email addresses should still be treated as case sensitive. Only allowing this in `forgotPassword` and keeping `sendResetPasswordEmail` case sensitive seemed to make sense for consistency with `sendEnrollmentEmail` and `sendVerificationEmail`.

What use case do you have in mind for invoking `sendResetPasswordEmail` manually on the server?

zimme: The case I had in mind isn't hard at all to just pick the stored email and use that, but just over all. If the account system support case insensitive emails I would expect that to be the case for all email related functions, or is my thinking completely flawed? :stuck_out_tongue_closed_eyes:

martijnwalraven: We've had some long discussions about this... The outcome was that we're not treating stored usernames and email addresses as fully case insensitive, but do want to expose some convenience functions on the client. I agree drawing the line there is kind of arbitrary and can be confusing.

Not going all the way with case insensitivity is mostly for backwards compatibility (i.e. multiple existing users with a username or email only differing by case). But in addition, it turns out the local part of email addresses is actually case sensitive according to the RFC, although that is probably highly theoretical.

In server code, you can use `findUserByUsername` and `findUserByEmail` to do a case insensitive lookup manually, but we hadn't considered the need to know which actual email address matched to pass it on to other functions, and as you found out that now requires you to filter for the right email yourself.

The alternative would be to make all `send<...>Email` functions use the same case insensitive logic, but I'm not sure that is worth it at this point.

zimme: For me personally, I have no big problem with the solution as it stands. I'm just thinking about what would yield the smallest surprise for developers.

As long as I know that there's a reason behind the decisions made I'm good. :+1:

packages/accounts-password/password_server.js
stubailo: Wouldn't `Accounts.findUserByEmail` be more appropriate here?

stubailo:
This looks good to me, @martijnwalraven!

:+1:

zimme:
> Why not use Accounts.findUserByEmail() which is case insensitive?

Don't know why I didn't think of this... @martijnwalraven maybe the code can be simplified with using that?

stubailo:
Oh yeah I meant sounds good after we use findUserByEmail

martijnwalraven:
Yeah, using `findUserByEmail` is indeed cleaner. The pluck code is still needed however, because neither `_findUserByQuery` nor `findUserByEmail` tells you which email address actually matched.

zimme:
So this could be closed as of 5931bcdae362e1026ceb8a08e5a4b053ce5340b7?

martijnwalraven:
I think so :)

Issue #5665, opened 2015-11-17
User logging in or out overwrites onReconnect and code using onReconnect breaks resume-on-reconnect


brettle:
**_1 Upvote**_ The docs say that `connection.onReconnect` can be used to set "a function to be called as the first step of reconnecting." However, if the user logs in or out using the connection, the accounts system will overwrite any `onReconnect` function that has been set. This overwriting occurs [here](https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_client.js#L220) and [here](https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_client.js#L321). There is also the related issue that other code setting `connection.onReconnect` will cause the resume-on-reconnect functionality to stop working.

 If you really need a reproduction repo, I can put one together, but I'd prefer to just include appropriate tests in a PR which fixes the problem.

Here's the fix I propose:
1. Deprecate `connection.onReconnect`.
2. Add a `connection.onReset(callback)` method that registers `callback` as one of potentially several callbacks to be called when the connection is reset/reconnected (in addition to calling `connection.onReconnect()`. Like `onLogin`, etc. `connection.onReset(callback)` would return an object with a `stop()` method that could be used to unregister the callback.
3. Change `AccountsClient.callLoginMethod` to call `onReset(callback)` instead of setting `onReconnect`, so that other code can't break the resume-on-reconnect functionality. 

Would a PR along these lines be considered?

avital:
Hey @brettle, thanks for the clearly explained issue.

You're right -- we document the `onReconnect` API even though for all practical purposes it's not useable (unless it's a connection that never logs in I suppose).

I'm curious to know what your use-case is. Can you elaborate on that?

If I were to make this change I wouldn't even keep `onReconnect` around since as you pointed out it couldn't really have been used anyways. I'd make it `onReconnect(callback)`. And that would need to return a handle that has an `.unregister()` method that so that the accounts system can unregister the original handle when there's a login or logout on the connection.

brettle:
Hi @avital. I have 2 specific use cases at the moment.

[...]

PR #5677, opened 2015-11-18
Fix bug #5665: add DDP.onReconnect(), deprecate conn.onReconnect

brettle:
Deprecate use of `connection.onReconnect = func`. Instead, a new
`DDP.onReconnect(callback)` method should be used to register callbacks to call
when a connection reconnects. The connection that is reconnecting is passed as
the only argument to `callback`. This is used by the accounts system to relogin
on reconnects without interfering with other code which uses
`connection.onReconnect`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,13 @@
 ## v.NEXT
 
+* Deprecate use of `connection.onReconnect = func`. Instead, a new
+  `DDP.onReconnect(callback)` method should be used to register callbacks to
+  call when a connection reconnects. The connection that is reconnecting is
+  passed as the only argument to `callback`. This is used by the accounts system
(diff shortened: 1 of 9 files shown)


timbotnik:
Thanks! We will take a look at this more deeply.

tmeasday:

hwillson:
As discussed in https://github.com/meteor/meteor/issues/5665#issuecomment-324711884, I'm re-opening this. This PR was headed down the right path (and `onReconnect` is still broken), so it would be great to get this functionality in place.

hwillson:
Since this PR's branch is commit locked, I'll continue this work in PR https://github.com/meteor/meteor/pull/9092.

Issue #5633, opened 2015-11-08
Great slowdown because of the interaction between a Minimongo query and subscription


mitar:
***1 Upvote*** I want my Saturday back.

See reproduction [here](https://github.com/mitar/meteor-issue5633).

The issue is that I have one subscription which returns the latest document in a collection. And then I display that latest document. So I have a helper like:

```javascript
Template.hello.helpers({
  latest: function () {
    return Documents.findOne({}, {sort: {createdAt: -1}});
  }
});
```

And then I want to subscribe to the whole collection as well, so I open another subscription to a publish function which publishes all documents. That subscription takes (in the example app) for 500 documents **more than 10 seconds to get to the ready state**. The interesting thing is that if I comment out the content of the `latest` template helper, subscription finishes at normal speed (0.3 seconds, which still feels a bit large for localhost app, but OK).

It seems that the bottleneck is in the Minimongo query which is then slowing down everything else, including the reception and processing of DDP messages. So it seems that for every new document Minimongo is recomputing the query, and this grows super-linearly, it is slower and slower with number of documents in the collection.

I would guess this is potentially huge slowdown for many applications because. Especially cases like infinity scroll where you have an order specified in Minimongo and you are subscribing to a new larger publish of same documents.

steph643:
Are you sure this is not due to the usual Blaze slowdown, as described [here](https://github.com/meteor/meteor/issues/3979)?

mitar:
No. If I do:

```
Tracker.autorun(function () {
  // Comment this line out to make subscription fast.
  return Documents.findOne({}, {sort: {createdAt: -1}});
});
```

So not using Blaze, but just an autorun to open a query, things work as slow as before. But if I remove it, then they are fast.

So it is not Blaze related. It might be that it happens with Blaze more often (especially with template-level subscriptions) but it is happening because of open query with sorting.

mitar:
One very interesting thing is also that if I change publish from:

```javascript
Meteor.publish('last-document', function () {
  return Documents.find({}, {sort: {createdAt: -1}, limit: 1});
});
```

To (so `-1` to `1`):

```javascript
Meteor.publish('last-document', function () {
  return Documents.find({}, {sort: {createdAt: 1}, limit: 1});
});
```

Then client side subscription is fast as well. But interesting, if I change then client side to match the order (so change `-1` to `1` in `sort`):

```javascript
Template.hello.helpers({
  latest: function () {
[...]

Issue #4073, opened 2015-03-31
Unnecessary re-rendering in Blaze because with's ReactiveVar uses default equals function


mitar:
***1 Upvote*** Blaze uses in its [Blaze.With](https://github.com/meteor/meteor/blob/devel/packages/blaze/builtins.js#L16) view `ReactiveVar` with default equality function:

> A function of two arguments, called on the old value and the new value whenever the ReactiveVar is set.  If it returns true, no set is performed.  If omitted, the default `equalsFunc` returns true if its arguments are `===` and are of type number, boolean, string, undefined, or null.

This means that for most data contexts which are objects re-rendering is triggered because equality function returns false. Even if new and old objects are referentially or content-wise completely equal.

The issue is that this triggers again and again `Template.currentData()` which might be in autoruns which do complex code.

So the questions are: why not provide some configurability here? Somehow one-size fits all approach is not good enough here. Some objects are equal when they are referentially equal. Some objects are equal when `EJSON.equals` returns true and it is cheaper for equality then recomputing the autorun reactive var is used in. For some you want fast failure like it is now.

Maybe a set of utilities around reactive dependencies would be beneficial which would help? https://github.com/peerlibrary/meteor-computed-field does not help with `Template.currentData()` because it does not use Blaze autorun.

In any case, the main issue is that internally all with data contexts are rerendered. And if this triggers then dynamic template inclusion to rerun, then you have things rebuilding and re-rendering all over the place.

dgreensp:
It's true that being able to configure how `{{#with}} does comparisons might be a nice analogue of React's `shouldComponentUpdate`.  At the moment, combined with the fact that too many common patterns in Blaze take a dependency on the data context, some inefficiency results.

As a workaround for particular cases where you really don't want to re-render a template unnecessarily, you can use a `ReactiveVar` on the template instance and configure its `equalsFunc`.

mitar:
My current workaround is:

```
@currentData = new ReactiveVar null, EJSON.equals
@autorun (computation) =>
  @currentData.set Template.getData()
```

Probably this is what you meant as well. :-)

hpx7:
+1 (see also #3966)

mitar:
This is especially problematic with `#each`.

mitar:
Closing in favor of: https://github.com/meteor/blaze/issues/68

Issue #1710, opened 2013-12-25
Allow d3.js at server.


livingyang:
Sometimes we need parse csv file, and d3.js has an excellent csv parser. But d3.js only enable at client now. Could you enable it at server?

PR #5680, opened 2015-11-19
Process DDP updates in batches

mitar:
See #5633 for discussion.

### DIFF
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -40,9 +40,9 @@ var Connection = function (url, options) {
     retry: true,
     respondToPings: true,
     // When updates are coming within this ms interval, batch them together.
-    updatesBatchingInterval: 5,
-    // But if batching is happening for more than this ms, process it anyway.
-    updatesBatchingMaxAge: 500
+    bufferedWritesInterval: 5,
+    // Flush buffers immediately if writes are happening continuously for more than this many ms.
+    bufferedWritesMaxAge: 500
   }, options);
 
   // If set, called when we reconnect, queuing method calls _before_ the
@@ -178,16 +178,16 @@ var Connection = function (url, options) {
   // if we're blocking a migration, the retry func
   self._retryMigrate = null;
 
-  self.__processBatchedUpdates = Meteor.bindEnvironment(self._processBatchedUpdates, "livedata batch", self);
+  self.__flushBufferedWrites = Meteor.bindEnvironment(self._flushBufferedWrites, "ddp buffered writes", self);
   // Collection name -> array of messages.
-  self._batchedUpdates = {};
-  // When current batching of a updates started, in ms timestamp.
-  self._batchedUpdatesStart = null;
-  // Timeout handle for the next processing of batched updates.
-  self._batchedUpdatesProcessingHandle = null;
+  self._bufferedWrites = {};
+  // When current buffer of updates started, in ms timestamp.
+  self._bufferedWritesStartedAt = null;
+  // Timeout handle for the next processing of all pending writes
+  self._bufferedWritesFlushHandle = null;
 
-  self._updatesBatchingInterval = options.updatesBatchingInterval;
-  self._updatesBatchingMaxAge = options.updatesBatchingMaxAge;
+  self._bufferedWritesInterval = options.bufferedWritesInterval;
+  self._bufferedWritesMaxAge = options.bufferedWritesMaxAge;
 
   // metadata for subscriptions.  Map from sub ID to object with keys:
   //   - id
@@ -1218,56 +1218,56 @@ _.extend(Connection.prototype, {
       // We'll now process and all of our buffered messages, reset all stores,
       // and apply them all at once.
       _.each(self._messagesBufferedUntilQuiescence, function (bufferedMsg) {
-        self._processOneDataMessage(bufferedMsg, self._batchedUpdates);
+        self._processOneDataMessage(bufferedMsg, self._bufferedWrites);
       });
       self._messagesBufferedUntilQuiescence = [];
     } else {
-      self._processOneDataMessage(msg, self._batchedUpdates);
+      self._processOneDataMessage(msg, self._bufferedWrites);
     }
 
     // Sometimes we'll need to disable deferred batching
     //  to make the livedata tests readable / synchronous.
     // Also, it's more responsive to flush batches when subscriptions start / end
-    if (self._updatesBatchingInterval === 0 || msg.msg === "nosub" || msg.msg === "ready") {
-      self._processBatchedUpdates();
+    if (self._bufferedWritesInterval === 0 || msg.msg === "nosub" || msg.msg === "ready") {
+      self._flushBufferedWrites();
       return;
     }
 
-    if (self._batchedUpdatesStart === null) {
-      self._batchedUpdatesStart = new Date().valueOf();
+    if (self._bufferedWritesStartedAt === null) {
+      self._bufferedWritesStartedAt = new Date().valueOf();
     }
-    else if (self._batchedUpdatesStart + self._updatesBatchingMaxAge < new Date().valueOf()) {
-      self._processBatchedUpdates();
+    else if (self._bufferedWritesStartedAt + self._bufferedWritesMaxAge < new Date().valueOf()) {
+      self._flushBufferedWrites();
       return;
     }
 
-    if (self._batchedUpdatesProcessingHandle) {
-      clearTimeout(self._batchedUpdatesProcessingHandle);
+    if (self._bufferedWritesFlushHandle) {
+      clearTimeout(self._bufferedWritesFlushHandle);
     }
-    self._batchedUpdatesProcessingHandle = setTimeout(self.__processBatchedUpdates,
-                                                      self._updatesBatchingInterval);
+    self._bufferedWritesFlushHandle = setTimeout(self.__flushBufferedWrites,
+                                                      self._bufferedWritesInterval);
   },
 
-  _processBatchedUpdates: function () {
+  _flushBufferedWrites: function () {
     var self = this;
 
-    if (self._batchedUpdatesProcessingHandle) {
-      clearTimeout(self._batchedUpdatesProcessingHandle);
-      self._batchedUpdatesProcessingHandle = null;
+    if (self._bufferedWritesFlushHandle) {
+      clearTimeout(self._bufferedWritesFlushHandle);
+      self._bufferedWritesFlushHandle = null;
     }
 
-    self._batchedUpdatesStart = null;
+    self._bufferedWritesStartedAt = null;
 
-    if (self._resetStores || !_.isEmpty(self._batchedUpdates)) {
+    if (self._resetStores || !_.isEmpty(self._bufferedWrites)) {
       // Begin a transactional update of each store.
       _.each(self._stores, function (s, storeName) {
-        s.beginUpdate(_.has(self._batchedUpdates, storeName) ?
-                      self._batchedUpdates[storeName].length : 0,
+        s.beginUpdate(_.has(self._bufferedWrites, storeName) ?
+                      self._bufferedWrites[storeName].length : 0,
                       self._resetStores);
       });
       self._resetStores = false;
 
-      _.each(self._batchedUpdates, function (updateMessages, storeName) {
+      _.each(self._bufferedWrites, function (updateMessages, storeName) {
         var store = self._stores[storeName];
         if (store) {
           _.each(updateMessages, function (updateMessage) {
@@ -1286,7 +1286,7 @@ _.extend(Connection.prototype, {
         }
       });
 
-      self._batchedUpdates = {};
+      self._bufferedWrites = {};
 
       // End update transaction.
       _.each(self._stores, function (s) { s.endUpdate(); });
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -1227,7 +1227,8 @@ _.extend(Connection.prototype, {
 
     // Sometimes we'll need to disable deferred batching
     //  to make the livedata tests readable / synchronous.
-    if (self._updatesBatchingInterval === 0) {
+    // Also, it's more responsive to flush batches when subscriptions start / end
+    if (self._updatesBatchingInterval === 0 || msg.msg === "nosub" || msg.msg === "ready") {
       self._processBatchedUpdates();
       return;
     }
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -1225,6 +1225,13 @@ _.extend(Connection.prototype, {
       self._processOneDataMessage(msg, self._batchedUpdates);
     }
 
+    // Sometimes we'll need to disable deferred batching
+    //  to make the livedata tests readable / synchronous.
+    if (self._updatesBatchingInterval === 0) {
+      self._processBatchedUpdates();
+      return;
+    }
+
     if (self._batchedUpdatesStart === null) {
       self._batchedUpdatesStart = new Date().valueOf();
     }
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -178,7 +178,8 @@ var Connection = function (url, options) {
   // if we're blocking a migration, the retry func
   self._retryMigrate = null;
 
-  self.__flushBufferedWrites = Meteor.bindEnvironment(self._flushBufferedWrites, "ddp buffered writes", self);
+  self.__flushBufferedWrites = Meteor.bindEnvironment(
+    self._flushBufferedWrites, "flushing DDP buffered writes", self);
   // Collection name -> array of messages.
   self._bufferedWrites = {};
   // When current buffer of updates must be flushed at, in ms timestamp.
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -38,7 +38,11 @@ var Connection = function (url, options) {
     reloadWithOutstanding: false,
     supportedDDPVersions: DDPCommon.SUPPORTED_DDP_VERSIONS,
     retry: true,
-    respondToPings: true
+    respondToPings: true,
+    // When updates are coming within this ms interval, batch them together.
+    updatesBatchingInterval: 5,
+    // But if batching is happening for more than this ms, process it anyway.
+    updatesBatchingMaxAge: 500
   }, options);
 
   // If set, called when we reconnect, queuing method calls _before_ the
@@ -173,6 +177,17 @@ var Connection = function (url, options) {
   self._updatesForUnknownStores = {};
   // if we're blocking a migration, the retry func
   self._retryMigrate = null;
+
+  self.__processBatchedUpdates = _.bind(self._processBatchedUpdates, self);
+  // Collection name -> array of messages.
+  self._batchedUpdates = {};
+  // When current batching of a updates started, in ms timestamp.
+  self._batchedUpdatesStart = null;
+  // Timeout handle for the next processing of batched updates.
+  self._batchedUpdatesProcessingHandle = null;
+
+  self._updatesBatchingInterval = options.updatesBatchingInterval;
+  self._updatesBatchingMaxAge = options.updatesBatchingMaxAge;
 
   // metadata for subscriptions.  Map from sub ID to object with keys:
   //   - id
@@ -1183,9 +1198,6 @@ _.extend(Connection.prototype, {
   _livedata_data: function (msg) {
     var self = this;
 
-    // collection name -> array of messages
-    var updates = {};
-
     if (self._waitingForQuiescence()) {
       self._messagesBufferedUntilQuiescence.push(msg);
 
@@ -1206,22 +1218,48 @@ _.extend(Connection.prototype, {
       // We'll now process and all of our buffered messages, reset all stores,
       // and apply them all at once.
       _.each(self._messagesBufferedUntilQuiescence, function (bufferedMsg) {
-        self._processOneDataMessage(bufferedMsg, updates);
+        self._processOneDataMessage(bufferedMsg, self._batchedUpdates);
       });
       self._messagesBufferedUntilQuiescence = [];
     } else {
-      self._processOneDataMessage(msg, updates);
+      self._processOneDataMessage(msg, self._batchedUpdates);
     }
 
-    if (self._resetStores || !_.isEmpty(updates)) {
+    if (self._batchedUpdatesStart === null) {
+      self._batchedUpdatesStart = new Date().valueOf();
+    }
+    else if (self._batchedUpdatesStart + self._updatesBatchingMaxAge < new Date().valueOf()) {
+      self._processBatchedUpdates();
+      return;
+    }
+
+    if (self._batchedUpdatesProcessingHandle) {
+      clearTimeout(self._batchedUpdatesProcessingHandle);
+    }
+    self._batchedUpdatesProcessingHandle = setTimeout(self.__processBatchedUpdates,
+                                                      self._updatesBatchingInterval);
+  },
+
+  _processBatchedUpdates: function () {
+    var self = this;
+
+    if (self._batchedUpdatesProcessingHandle) {
+      clearTimeout(self._batchedUpdatesProcessingHandle);
+      self._batchedUpdatesProcessingHandle = null;
+    }
+
+    self._batchedUpdatesStart = null;
+
+    if (self._resetStores || !_.isEmpty(self._batchedUpdates)) {
       // Begin a transactional update of each store.
       _.each(self._stores, function (s, storeName) {
-        s.beginUpdate(_.has(updates, storeName) ? updates[storeName].length : 0,
+        s.beginUpdate(_.has(self._batchedUpdates, storeName) ?
+                      self._batchedUpdates[storeName].length : 0,
                       self._resetStores);
       });
       self._resetStores = false;
 
-      _.each(updates, function (updateMessages, storeName) {
+      _.each(self._batchedUpdates, function (updateMessages, storeName) {
         var store = self._stores[storeName];
         if (store) {
           _.each(updateMessages, function (updateMessage) {
@@ -1239,6 +1277,8 @@ _.extend(Connection.prototype, {
                                      updateMessages);
         }
       });
+
+      self._batchedUpdates = {};
 
       // End update transaction.
       _.each(self._stores, function (s) { s.endUpdate(); });
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -1250,24 +1250,28 @@ _.extend(Connection.prototype, {
 
   _flushBufferedWrites: function () {
     var self = this;
-
     if (self._bufferedWritesFlushHandle) {
       clearTimeout(self._bufferedWritesFlushHandle);
       self._bufferedWritesFlushHandle = null;
     }
 
     self._bufferedWritesStartedAt = null;
+    self._performWrites(self._bufferedWrites);
+    self._bufferedWrites = {};
+  },
 
-    if (self._resetStores || !_.isEmpty(self._bufferedWrites)) {
+  _performWrites: function(updates){
+    var self = this;
+
+    if (self._resetStores || !_.isEmpty(updates)) {
       // Begin a transactional update of each store.
       _.each(self._stores, function (s, storeName) {
-        s.beginUpdate(_.has(self._bufferedWrites, storeName) ?
-                      self._bufferedWrites[storeName].length : 0,
+        s.beginUpdate(_.has(updates, storeName) ? updates[storeName].length : 0,
                       self._resetStores);
       });
       self._resetStores = false;
 
-      _.each(self._bufferedWrites, function (updateMessages, storeName) {
+      _.each(updates, function (updateMessages, storeName) {
         var store = self._stores[storeName];
         if (store) {
           _.each(updateMessages, function (updateMessage) {
@@ -1285,8 +1289,6 @@ _.extend(Connection.prototype, {
                                      updateMessages);
         }
       });
-
-      self._bufferedWrites = {};
 
       // End update transaction.
       _.each(self._stores, function (s) { s.endUpdate(); });
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -181,8 +181,8 @@ var Connection = function (url, options) {
   self.__flushBufferedWrites = Meteor.bindEnvironment(self._flushBufferedWrites, "ddp buffered writes", self);
   // Collection name -> array of messages.
   self._bufferedWrites = {};
-  // When current buffer of updates started, in ms timestamp.
-  self._bufferedWritesStartedAt = null;
+  // When current buffer of updates must be flushed at, in ms timestamp.
+  self._bufferedWritesFlushAt = null;
   // Timeout handle for the next processing of all pending writes
   self._bufferedWritesFlushHandle = null;
 
@@ -1233,10 +1233,10 @@ _.extend(Connection.prototype, {
       return;
     }
 
-    if (self._bufferedWritesStartedAt === null) {
-      self._bufferedWritesStartedAt = new Date().valueOf();
+    if (self._bufferedWritesFlushAt === null) {
+      self._bufferedWritesFlushAt = new Date().valueOf() + self._bufferedWritesMaxAge;
     }
-    else if (self._bufferedWritesStartedAt + self._bufferedWritesMaxAge < new Date().valueOf()) {
+    else if (self._bufferedWritesFlushAt < new Date().valueOf()) {
       self._flushBufferedWrites();
       return;
     }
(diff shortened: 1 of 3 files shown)


tmeasday:
@mitar, remind me; do we still want to merge this, or is there little point given what we discovered about minimongo?

mitar:
I think it is still work in progress. We could merge it as it is, it does not hurt and it enables future optimizations. But we might also wait for fixing the Minimongo as well.

mitar:
I think it would be really great if we would fix the Minimongo issue. :-) Any takers? I really think it impacts many Meteor apps we assume it is rendering for slowdowns.

nathan-muir:
Hmmm... after digging through a few things, here's what I've noticed.

There seems to be two tiers of optimisations available:

1. Queries that call insert/ update/ delete _and have a skip/limit_ will always call `_recomputeResults`. 
2. Others will call some combination of `_(insert|update|remove)InResults`

I think @glasser's comments in #5633 are in reference to [2]
> You could do something like, if a query (observe) gets more than a certain number of changes while paused, we stop updating it and mark it as "dirty" and recompute it at resume time.

However, [1] seems to be the low hanging fruit in terms of easy performance gains, and semantics.

I think something as simple as
```js
LocalCollection.prototype._recomputeResults = function (query, oldResults) {
  var self = this;
  if (self.paused){
    self.dirty = true;
    return;
  }
  /* ... rest of function ... */
};
LocalCollection.prototype.resumeObservers = function () {
  var self = this;
  // No-op if not paused.
  if (!this.paused)
    return;

  // Unset the 'paused' flag. Make sure to do this first, otherwise
  // observer methods won't actually fire when we trigger them.
  this.paused = false;

  for (var qid in this.queries) {
    var query = self.queries[qid];
    if (query.dirty){
       // re-compute results will do the diff automatically. 
       self._recomputeResults(query, query.resultsSnapshot);
    } else {
      // Diff the current results against the snapshot and send to observers.
      // pass the query object for its observer callbacks.
      LocalCollection._diffQueryChanges(
        query.ordered, query.resultsSnapshot, query.results, query,
        { projectionFn: query.projectionFn });
      }
      query.resultsSnapshot = null;
  }
  self._observeQueue.drain();
};
```

mitar:
That sounds a great idea. Do you want to add this directly to this pull request?

nathan-muir:
@mitar Yep! I'll make some patches now, and test against your example app.

mitar:
I will give you access to my fork then. Please modify only this branch.

nathan-muir:
Preliminary results look good:

Before ~6s solid processing.
image

After ~ 0.4s light processing (two blocks of 40ms uninterrupted processing).
image

mitar:
That is with my example?

nathan-muir:
Yep, that's with https://github.com/mitar/meteor-issue5633

mitar:
That is amazing. I love this collaboration. Me adding pausing and you adding dirty flag. :-) So if I understand correctly, the dirty flag is a new thing you added? I am curious, why it has to be per query and not per (MiniMongo) collection?

> As for optimising _insertInResults and related functions. I see a few blocking factors:

I do not see those as issue. We currently pause only when we think it is not needed. So I would really just make things thirty when paused and this is it. And then we can leave to pausing logic to determine when to pause and when not to.

> I think we could simply mark any query as query.dirty=true if _recomputeResults is called while paused. There is no advantage to recomputing early - it can all be done during resumeObservers.

Doesn't this address both 1. and 2. above? This is why I like it. Because it is agnostic to operations and it is very minimal.

Or are you trying to optimize queries also while things are not paused? So that if I call insert during normal operation that we should not recompute until values are needed?

So one thing I wrote above is that I do not understand why we are computing queries eagerly at all. Why do not all changes just mark it dirty and then when you call query it gets recomputed? Is it because some changes are simpler than whole recompilation? Then maybe all changes should be put in the queue. So what we could do is make `dirty` a queue of changes we store. And then if query is such which always recompute everything, we just recompute. If the query is such which just can modify some state we then apply it at first read. And we can then decide at that moment how to apply it: if it is short queue we maybe run it, otherwise we maybe recompute it all.

nathan-muir:
> That is amazing. I love this collaboration.

I think this worked out just right, as I've been scouring over livedata/ minimongo lately with some experiments; this PR just happened to be in an area of current knowledge.

>> As for optimising _insertInResults and related functions. I see a few blocking factors: 

> I do not see those as issue

The reason why I brought them up, is that `insertInResults` is efficient compared to recomputing the whole result set.

> Doesn't this address both 1. and 2. above? This is why I like it. Because it is agnostic to operations and it is very minimal.

I'll re-word it:

There seems to be two paths taken for each document inserted/updated/deleted:

1. Queries that have skip/limit_ will overwrite `query.results = getRawObjects()` every invocation of call insert/ update/ delete. _This is really inefficient!_
2. Other queries will mutate `query.results` by just adding/ removing/ updating the doc as necessary. _This is already somewhat efficient_

So, seeing as (2) is already _pretty good_ it's not as straight forward to optimise without precise examples.

mitar:
> (Aside: have been looking at using immutable objects internally in minimongo to make integration with react easier/faster).

Oh yes, that would improve performance even further. See this ticket: https://github.com/meteor/meteor/issues/4073

Currently in Tracker, Blaze and everywhere else where we use default `ReactiveVar`'s equality we recompute stuff every time something is an object. It would be great if could know that if something is equal by reference is also equal by content. Then we could just use `===` in `ReactiveVar` and optimize stuff a lot.

> So, seeing as (2) is already pretty good it's not as straight forward to optimise without precise examples.

OK, so second part of my comment above was then about this. So to make it more concrete, with:

* for 1., we are addressing this with this pull request
* for 2., we could queue all mutations (like `insertInResults`) into a queue and do not apply them immediately, then when somebody wants to read from the query, we check the length of the queue,
  * if the length is < N, we apply it and return the results
  * otherwise we recompute the whole query

We then vary N and see what should be the default value. My guess is around 100. It could even be parameter to `find`.

I do not think this second thing would be too hard to implement?

The issue is that if query is observed, then "somebody wants to read from the query" means that it should run every time anyway. So maybe this is not so useful optimization.

Maybe a better next optimization would be to improve how queries are done in the autorun. Currently, queries inside an autorun use [observeChanges](https://github.com/meteor/meteor/blob/devel/packages/minimongo/minimongo.js#L534) to know when to invalidate. But you do not really need to know exactly what is the new value, only that it would be changed.

Also, now looking at this code I see one other low hanging fruit. We should implement the function like this:

```javascript
LocalCollection.Cursor.prototype._depend = function (changers, _allow_unordered) {
  var self = this;

  if (Tracker.active) {
    var handle = null;
    var v = new Tracker.Dependency;
    v.depend();
    var notifyChange = function () {
      v.changed();
      if (handle) handle.stop();
    };

    var options = {
      _suppress_initial: true,
      _allow_unordered: _allow_unordered
    };
    _.each(['added', 'changed', 'removed', 'addedBefore', 'movedBefore'],
           function (fnName) {
             if (changers[fnName])
               options[fnName] = notifyChange;
           });

    // observeChanges will stop() when this computation is invalidated
    handle = self.observeChanges(options);
  }
};
```

See, now on the first invalidation we stop the observe. There is no more need for it. Dependency has been invalidated. So observe does not have to computed anymore. I think this could potentially have huge impact for queries inside an autorun. What do you think?

BTW, have you run all tests locally? CircleCI has some issues it seems. Timeouts.

packages/minimongo/minimongo.js
```diff
@@ -957,6 +967,11 @@ LocalCollection._updateInResults = function (query, doc, old_doc) {
 // oldResults is guaranteed to be ignored if the query is not paused.
 LocalCollection.prototype._recomputeResults = function (query, oldResults) {
   var self = this;
+  if (self.paused){
```
mitar: Style.

packages/minimongo/minimongo.js
```diff
@@ -1079,11 +1094,17 @@ LocalCollection.prototype.resumeObservers = function () {
 
   for (var qid in this.queries) {
     var query = self.queries[qid];
-    // Diff the current results against the snapshot and send to observers.
-    // pass the query object for its observer callbacks.
-    LocalCollection._diffQueryChanges(
-      query.ordered, query.resultsSnapshot, query.results, query,
-      { projectionFn: query.projectionFn });
+    if (query.dirty){
```
mitar: Style.

mitar:
But yes, this looks really cool. I think this should be merged in at this level already.

nathan-muir:
> The issue is that if query is observed, then "somebody wants to read from the query" means that it should run every time anyway. So maybe this is not so useful optimization.

Yeah, there's not much of a way around this. However - I think the DDP event batching will already improve things for case (2), as it minimises the calls to `LocalCollection._diffQueryChanges`

> BTW, have you run all tests locally? CircleCI has some issues it seems. Timeouts.

Ah, just run some tests on `livedata`, fixed a few bugs due to async.

Still have an issue in `livedata - publisher errors with onError callback` as the `onReady` is not being called any more.

packages/ddp-client/livedata_connection.js
```diff
@@ -1206,22 +1218,55 @@ _.extend(Connection.prototype, {
...
       });
       self._messagesBufferedUntilQuiescence = [];
     } else {
-      self._processOneDataMessage(msg, updates);
+      self._processOneDataMessage(msg, self._batchedUpdates);
+    }
+
+    // Sometimes we'll need to disable deferred batching
+    //  to make the livedata tests readable / synchronous.
+    if (self._updatesBatchingInterval === -1){
```
mitar: Style. Space before `{`.

mitar: Also, this could probably be `0`, no?

nathan-muir:
Okay, tests should be passing now.

But it would be good for someone else the solution - basically we flush any pending updates in Livedata if a "nosub" or "ready" message is received.

There may be a more elegant way to achieve the same result though.

mitar:
Hm, I have issues parsing your middle sentence in the previous comment.

nathan-muir:
Fixed!

mitar:
Hm, an interesting idea. But also potentially problematic in the case of `subscribe` inside an autorun, no? Especially in the infinity scroll case. Because there you unsubscribe and resubscribe for almost the same messages, and in fact you would not want to do recomputation in the middle between those, but only at the end. What do you think? Is this a problem?

I think this pull request also lacks tests. We are introducing some new options and behavior. We should test for that. It would be also good to test things with observe and autorun. To see how things behave there and assure that things stay optimized. Because here for example, you have two parts which have to work together, batching of DDP requests and dirty flag. Imagine that batching stops working for some reason, we should be able to observe that with a test.

mitar:
> review the solution 

Is this the solution to fixing the tests or just one extra optimization?

mitar:
In fact I think that my comment about `nosub`/`ready` is not correct. Those messages do not come between unsubscribe/resubscribe, just at the end of that cycle, which is good. So I think this is good how it is now.

mitar:
Have you seen: https://github.com/meteor/meteor/commit/f0404b6e0d50ce80949f896d346d5b2557da5494

nathan-muir:
> ... you would not want to do recomputation in the middle between those, but only at the end. ...

I think this would be ideal in an _optimal_ scenario, but it would require too much engineering for little result. Even in the example app you provided, it still does 2 batches/recomputes. Which is much better than `N=1000` recomputes; but still not the ideal case of only 1 recompute.

I think the next tier of optimisation would be to offload diff/recompute to a webworker, as anything over 16ms can block the UI thread enough to ruin animations etc.

> I think this pull request also lacks tests.

I had to disable ddp batches in `ddp-client/livedata_connection_tests.js`, as most of the tests are write+assert (otherwise it would just be re-writing the tests to write+flush+assert).

There's probably _some_ tests that could test batching, but I haven't had a chance to read through & re-write them.

> It would be also good to test things with observe and autorun. To see how things behave there and assure that things stay optimized. Because here for example, you have two parts which have to work together, batching of DDP requests and dirty flag. Imagine that batching stops working for some reason, we should be able to observe that with a test.

Livedata - Yes we can test batching, and ensure that all types of messages/actions are processed correctly / in-order.

Minimongo - I'm not really sure what we can test here; previously the code was just doing lots of re-computes which were completely ignored/wasted. We can test that the new code-paths haven't broken the API of observe; but I'm not sure we should test/enforce that only a single re-compute happens per pause/batch of data?


> > review the solution

> Is this the solution to fixing the tests or just one extra optimization?

I just wanted to confirm that flushing updates at `nosub` / `ready` was optimal / correct.

> Have you seen: f0404b6

Yep, looks like a good optimisation, and shouldn't affect this one at all.

mitar:
> I think this would be ideal in an optimal scenario, but it would require too much engineering for little result.

As I commented later, this is in fact not a problem and it already behaves correctly.

> I think the next tier of optimisation would be to offload diff/recompute to a webworker, as anything over 16ms can block the UI thread enough to ruin animations etc.

+1

> Next Steps

+1

mitar:
This is also interesting: https://github.com/meteor/meteor/blob/devel/packages/minimongo/minimongo.js#L126

nathan-muir:
@tmeasday As the only person from MDG that's commented on this; I would just like to follow up to see what can be done to get this merged.

The changes to `minimongo` are pretty basic, and the performance improvement substantial.

tmeasday:
@nathan-muir just back from vacation. Will try to look at this later this week.

stubailo:
Gonna take a detailed look this week and try to merge this!

packages/ddp-client/livedata_connection.js
```diff
@@ -1490,6 +1541,11 @@ _.extend(Connection.prototype, {
 
     var self = this;
 
+    // Lets make sure there are no buffered writes before returning result.
+    if (!_.isEmpty(self._bufferedWrites)) {
```
nathan-muir: Not sure about this one; 

I added the `_.isEmpty` check here, as every call to `_flushBufferedWrites` will call `_runAfterUpdateCallbacks` (even if empty). So flushing without this predicate would mean that the after-update callbacks would run on every method result.

Alternatively the call to `_runAfterUpdateCallbacks` could be predicated / hoisted out to another function.

packages/ddp-client/livedata_connection_tests.js
```diff
@@ -2,7 +2,10 @@ var newConnection = function (stream) {
   // Some of these tests leave outstanding methods with no result yet
   // returned. This should not block us from re-running tests when sources
   // change.
-  return new LivedataTest.Connection(stream, {reloadWithOutstanding: true});
+  return new LivedataTest.Connection(stream, {
+    reloadWithOutstanding: true,
+    bufferedWritesInterval: 0
```
nathan-muir: I've disabled buffering for the tests, as they all written to expect synchronous behaviour.

Not sure if we're better adding new tests or trying to adapt/modify existing ones.

packages/minimongo/minimongo.js
```diff
@@ -957,6 +965,14 @@ LocalCollection._updateInResults = function (query, doc, old_doc) {
 // oldResults is guaranteed to be ignored if the query is not paused.
 LocalCollection.prototype._recomputeResults = function (query, oldResults) {
   var self = this;
+  if (self.paused) {
+    // There's no reason to recompute the results now as we're still paused.
+    // By flagging the query as "dirty", the recompute will be performed
+    // when resumeObservers is called.
+    query.dirty = true;
```
nathan-muir: It might be worth clearing `query.results` here?

```js
      if (query.ordered) {
        query.results = [];
      } else {
        query.results.clear();
      }
```

packages/ddp-client/livedata_connection.js
```diff
@@ -38,7 +38,11 @@ var Connection = function (url, options) {
     reloadWithOutstanding: false,
     supportedDDPVersions: DDPCommon.SUPPORTED_DDP_VERSIONS,
     retry: true,
-    respondToPings: true
+    respondToPings: true,
+    // When updates are coming within this ms interval, batch them together.
+    bufferedWritesInterval: 5,
+    // Flush buffers immediately if writes are happening continuously for more than this many ms.
+    bufferedWritesMaxAge: 500
```
stubailo: How would one override these options for the default Meteor connection? I can imagine that 500ms might be too long for some apps that have basically continuous updates. (Maybe those don't exist?)

packages/ddp-client/livedata_connection.js
```diff
@@ -174,6 +178,17 @@ var Connection = function (url, options) {
   // if we're blocking a migration, the retry func
   self._retryMigrate = null;
 
+  self.__flushBufferedWrites = Meteor.bindEnvironment(self._flushBufferedWrites, "ddp buffered writes", self);
```
stubailo: Line length.

packages/ddp-client/livedata_connection.js
stubailo: Yeah this seems like a fine idea to me.

:+1:

packages/ddp-client/livedata_connection_tests.js
stubailo: We definitely need some new tests for this buffering functionality, that might be very non-trivial to write.

packages/ddp-client/livedata_connection.js
nathan-muir: TBH; I'm not sure about this one (it was added by @mitar). Does `Websocket.onmessage` take precedence over `setTimeout` in every browser/engine?

I was going to we-write this as a maximum buffer size, rather than a temporal boundary.

mitar:
I am glad this is moving forward. I am a bit busy at the moment with other stuff, but I am OK with the direction all this is moving. Thank you for all the work, both of you.

packages/ddp-client/livedata_connection.js
mitar: Do we really need double `_`?

packages/ddp-client/livedata_connection.js
mitar: No no. Do not change this. The point here is that what is really important to users (I believe) is framerate. Not size of data. You want to have predictable framerate.

To understand what is happening here is that in most cases there is only 5 ms delay, not 500 ms delay. So if changes are coming sporadically, then you have a 5 ms delay while buffering is waiting if there is no other change. If there is no other change, then change go through immediately. If there are changes for more than 500 ms coming without any 5 ms pause between any of those change, then after 500 ms of those sustained changes we flush them. I think this is also good. Because if you have 500 ms of sustained changes coming through then the whole point of this pull request was that we are not passing those changes through because they hog the whole rendering of those changes.

I think those defaults are pretty reasonable. 5 ms delay for rare changes, and 500 ms flushing of big changes which will probably change a lot of DOM anyway. But I agree that there should be a way to change those settings for the default connection. But what would be good API for this? Simple some `Meteor.defaultConnectionSettings` which you can set/override in your app and is then used when creating a default connection?

mitar: (At least that was how I implemented. Not sure if current implementation still works like that. It seems so.)

stubailo: Yeah I think having a setting for it would be good. We can either put it in Meteor.settings, or have a runtime function call for it (which makes sense to me since it would be totally reasonable to want to change the delay on certain pages of your app that want to be more realtime).

Perhaps something like `DDPClient.setBufferMaxAge`?

packages/ddp-client/livedata_connection_tests.js
tmeasday: I wrote such a test: https://github.com/meteor/meteor/commit/74230beba8777cb9104c4180525e2a3fe09772c4

Appreciate people's thoughts: 
 - Is this testing the right thing?
 - Is it sufficiently thorough or do we want to test more stuff (other combination of message timings, other data messages, not just in the stub?)
 - Is there any way to deterministically know when the max age is hit? It was hitting it at inconsistent/unexpected times for me, which I put down to the content of the test taking relatively significant time to run, as you can see in my comments.

stubailo:
@tmeasday that test looks good to me. So should we merge this now? I'm trying to think if I'm forgetting anything.

mitar:
We have not yet added a way to pass options to the default connection. I do not have good idea here, so if you have some, feel free to just do it.

packages/ddp-client/livedata_connection.js
maxnowack: A runtime function on the DDP connection would be great, since we could change the behavior for each connection at runtime.

mitar: I think adding changing this after connection is established might add extra complexity. Not sure if it is worth it. I do not see how much more it would add though, besides just changing those values.

tmeasday:
I think we decided that it's not vital to be able to change it. I actually can't remember what the exact reasoning was @stubailo? (always write things down grrr).

mitar:
Maybe because in the GraphQL world you will have much higher latencies anyway?

mitar:
I think in the worst case people can change internal `_bufferedWritesMaxAge` and other fields directly. I think they are exposed through `Meteor.connection`, no?

stubailo:
> Maybe because in the GraphQL world you will have much higher latencies anyway?

Mm, I think it's far too early to think about making these kinds of comparisons.

Anyway, I'll try to get this in by the end of the week.

packages/ddp-client/livedata_connection.js
maxnowack: I think you're right. It's more important to merge the PR than achieving this flexibility. It should be possible to implement this later.

mitar:
It is interesting, that in the [minimongo README](https://github.com/meteor/meteor/tree/devel/packages/minimongo) it is written:

> All massive replacements happen in between pauseObservers and resumeObservers calls, so any user code (such as Blaze) sees the whole change in one tick (helps to avoid the flicker).

Interesting that this seems not to be fact currently, but it will be after this pull request?

mitar:
So, are we merging this?

stubailo:
I'm worried about the one failing test (server-to-server minimongo). But I think it would be reasonable to just merge it anyway. We keep passing this around internally because it's hard to reason about asynchronous logic like this to fix the test, and make sure @tmeasday's new test passes.

glasser:
I'd be a little concerned about the test failure, since it's actually surfacing a "shouldn't happen" error from within the server:

```
I20160314-16:32:36.478(-7)? Exception in handling DDP message: Error: Expected not to find a document already present for an add
I20160314-16:32:36.478(-7)?     at Object.self._connection.registerStore.update (packages/mongo/collection.js:163:19)
I20160314-16:32:36.478(-7)?     at Object.store.(anonymous function) [as update] (packages/ddp-client/livedata_connection.js:464:1)
I20160314-16:32:36.478(-7)?     at packages/ddp-client/livedata_connection.js:1280:1
I20160314-16:32:36.479(-7)?     at Array.forEach (native)
I20160314-16:32:36.479(-7)?     at Function._.each._.forEach (packages/underscore/underscore.js:105:1)
I20160314-16:32:36.479(-7)?     at packages/ddp-client/livedata_connection.js:1279:1
I20160314-16:32:36.479(-7)?     at Function._.each._.forEach (packages/underscore/underscore.js:113:1)
I20160314-16:32:36.479(-7)?     at [object Object]._.extend._performWrites (packages/ddp-client/livedata_connection.js:1276:1)
I20160314-16:32:36.479(-7)?     at [object Object]._.extend._flushBufferedWrites (packages/ddp-client/livedata_connection.js:1261:1)
I20160314-16:32:36.479(-7)?     at [object Object]._.extend._livedata_result (packages/ddp-client/livedata_connection.js:1547:1)
```

It looks like the server is re-sending adds for messages it already sent, which is concerning.

glasser:
OK, the server isn't re-sending it, but minimongo is re-receiving it.

packages/ddp-client/livedata_connection.js
```diff
@@ -1206,13 +1219,52 @@ _.extend(Connection.prototype, {
...
+
+  _flushBufferedWrites: function () {
+    var self = this;
+    if (self._bufferedWritesFlushHandle) {
+      clearTimeout(self._bufferedWritesFlushHandle);
+      self._bufferedWritesFlushHandle = null;
     }
 
+    self._bufferedWritesFlushAt = null;
+    self._performWrites(self._bufferedWrites);
```
glasser: OK, I've briefly looked into the failing test.

These two lines are safe in normal JS but not in fiberized JS.  Specifically, it's possible for more messages to come down the network before `_performWrites` returns, which can lead to sending the same buffered writes to `_performWrites` again!

This is what is happening in the failing test.  Specifically, the structure is such that for some reason synchronously to `_performWrites` we call the `onReady` callback for a subscription (this is via `_runAfterUpdateCallbacks`).  In the case of the failing test, that onReady is a testAsyncMulti `expect` function, which believe it or not actually ends up invoking the next step of the test synchronously.  This means that `_performWrites` never returns in this case.

Changing these two lines to
```
    var writes = self._bufferedWrites;
    self._bufferedWrites = {};
    self._performWrites(writes);
```
fixes the failing test.  It does leave one wondering if there is a further problem.

I think in the old code it was moderately unproblematic that `_runAfterUpdateCallbacks` could block, since it was the final thing called from the "receive message" callback.  But now it can also be called from near the beginning of `_livedata_result`, which is a little hairier.  It's a little bad that user code (sub ready callbacks and method call callbacks) which can block will prevent return values from being received (including the onResultReceived callback which is explicitly supposed to be the one that doesn't have to wait on other stuff).  But I guess that's relatively unavoidable when you're using Fibers.

I would definitely make the change I suggested above.  I'm not sure if there are deeper changes needed to make this stuff more bulletproof — but maybe people don't use minimongo connected to DDP clients too much in Node anyway.

glasser: (I do see a failure in "reconnect auto-login" after applying this, though it passed when I re-tried.)

mitar: Just random question, do you think this previous version was more robust: https://github.com/mitar/meteor/commit/748b0b157141665390a8e13c05569be56c83fe1d

mitar: >  It's a little bad that user code (sub ready callbacks and method call callbacks) which can block will prevent return values from being received (including the onResultReceived callback which is explicitly supposed to be the one that doesn't have to wait on other stuff). But I guess that's relatively unavoidable when you're using Fibers.

But this is only when connecting as a client on the server, no? Because on the proper client you do not have Fibers?

glasser: re more robust, I assume you mean before that commit?  Hmm, I mean, it doesn't call the after-update callbacks in the critical section, but it still invokes the "store API", which you could imagine could yield.

Correct that this whole concern is probably only a problem on the server, but people do use server-to-server DDP to some degree.

nathan-muir:
> These two lines are safe in normal JS but not in fiberized JS

Good point! I didn't see any fiberized code, but I guess there wasn't any `noYieldsAllowed` so it's possible.

> Changing these two lines to...

Okay - That change has been added to the PR.

> But now it can also be called from near the beginning of _livedata_result, which is a little hairier.

Agreed. Is there anything preventing the callbacks from being deferred with `process.nextTick` or `Meteor.setTimeout`? 

> (I do see a failure in "reconnect auto-login" after applying this, though it passed when I re-tried.)

Tests passed first time here;

glasser:
@tmeasday asked me the same question about calling the user callbacks later.  Unfortunately I think it would take a lot of research for me to be able to answer this question; I don't know what sorts of guarantees apps think they have about this sort of thing. One thing I've learned over the years is that the set of Meteor APIs, some of which call callbacks synchronously, some asynchronously, lead to a lot of emergent behavior and implicit user expectations.  See eg #2315.  Hopefully future API design can avoid this by being consistently async.

mitar:
If I understand correctly, those apps which would do server-to-server DDP connections and would find issues with this buffering, could set `bufferedWritesInterval` to 0 to disable it?

Alternatively, we could just disable buffering for server-to-server DDP connections by default (and user would have to explicitly enable it).

mitar:
Anyway, is this then ready to be merged?

tmeasday:
@mitar I think we are too late in the RC process for 1.3 to include this now, given it (potentially) introduces some significant (and possibly breaking) changes to DDP's behavior on the client. 

I'd like this one to spend some time in beta-land before we ship it so a variety of people have a chance to test it on their apps. I propose we merge it to devel right after 1.3 ships and try and get a beta of the feature out soon after so we can all play with it in real apps.

mitar:
@tmesday: Ping on this one?

tmeasday:
Thanks for the reminder @mitar

copleykj:
I just did a little dance when I noticed this...

mitar:
Photo or it didn't happen.

TomFreudenberg:
Thanks for all that work! Great improvement.

abernix:
This is a great improvement, however it turns out to be a breaking change for anyone using the `kadirahq/fast-render` package.

I have opened an issue over there (kadirahq/fast-render#166) about it and perhaps it can be corrected, though some guidance might be appreciated from those involved in this change.  FastRender wraps the `_livedata_data` method and probably needs some compatibility changes.

abernix:
I do find it a bit strange that this change was snuck in on what I consider to be a patch release of Meteor and not even included in the History/Changelog.

lorensr:
I imagine the Changelog entry was just overlooked. I think in general
changing private methods doesn't count as a breaking change for semver, and
that's a risk you take when writing a package that uses them, but it would
be nice for people to try a search for the method and warn packages that
use it

abernix:
@lorensr I'm sure `History.md` (Change Log) was an oversight.  And you're absolutely right that internal methods are considered fair-game for *whatever*.  However, it was mentioned by @tmeasday that this could be a breaking change to the way DDP works and that a beta-cycle *might* be in order.  Instead, this was pushed in what was essentially [a bug-fix release (1.3.2.1)](https://github.com/meteor/meteor/compare/release-1.3.2...release-1.3.2.1).  (not sure semver is even in play here)

I honestly wouldn't even remotely have expected anyone to try search for a third-party package that might be affected (outside of core code) by changing a private method, however I would have expected this sort of commit in a beta-channel-release for at least a few days.

Oh, and I ain't even mad!  Was just surprised.

mitar:
Oh, this was already released? Didn't know that. :-)

mitar:
Although, this is probably a change which warrants a blog post announcement. People would probably find it really great to have.

benjamn:
This is my fault for cutting the `release-1.3.2.2` branch from `devel` instead of the old `release-1.3.2.1` branch. This definitely should have waited until 1.3.3.

Is this grounds for un-recommending 1.3.2.2? Note that 1.3.2.1 was never recommended.

tmeasday:
> Is this grounds for un-recommending 1.3.2.2

It might be? I kind of hoped to have some folks test 1.3.3 w/ their production apps to flush out bugs like this fast-render one (and more subtle issues)

tmeasday:
We've unrecommended 1.3.2.2 and will release 1.3.2.3 soon without this change.

sergiotapia:
@tmeasday Is 1.3.2.3 recommended now? https://github.com/meteor/meteor/releases

abernix:
@sergiotapia Give them a *few* minutes, hah.. :smile:

Also, you can use this meteor-tool command to see what's up:
```
meteor show --show-all METEOR
```

(Answer: No, it's not *currently*).

sergiotapia:
Sure I was just wondering how Meteor releases work. It seems they aren't tied to Github releases, so I'll use that command you pasted to keep tabs on releases.

copleykj:
@abernix that's a friggin handy command right there! Thanks for that.

sergiotapia:
Awesome, it works now :) `meteor update --release 1.3.2.3` No errors in my app.

abernix:
@tmeasday You wrote a `buffering data` test (74230beba8777cb9104c4180525e2a3fe09772c4) at one point in this PR – was there any consensus on whether or not that test was the right way to do it?  Seems like a reasonable method to me, but it doesn't seem to have got included in the PR and (when I added it manually) periodically fails for me (about 3/10 times).

tmeasday:
@abernix, I think that it probably should get merged but we should fix the test to pass properly.

I think probably we can do it by using [sinon](http://sinonjs.org)'s time faking mode (or something equivalent). 

Are you interested in sending a PR? I'd be keen to merge it.

abernix:
@tmeasday Sweet!  Just submitted PR #6880 that should address this! Also `fast-render` and `flow-router-ssr` have been fixed, so at this point we'll see if anything else goes wonky in 1.3.3 beta.  Cheers. 🍰

mitar:
Hm, thinking now, I think this pull request will break this package: https://arunoda.github.io/meteor-streams/ We should test it and see how we could make it so that it continues to work, which will in turn break Rocket.Chat (as one of most popular projects I know of which use meteor-streams).

I would claim that Rocket.Chat should move away for streams anyway for their purpose can could just use [control-mergebox package](https://github.com/peerlibrary/meteor-control-mergebox) ([#1710](https://github.com/RocketChat/Rocket.Chat/issues/1710)), but probably we should look into how to make it continue to work. Maybe we could have some way to not batch changes per-collection?

tmeasday:
I guess they can just turn it off completely but I think it'd definitely worth communicating with @arunoda et al. Let's do so on the forum thread (just testing out the beta now, will post the thread here when I create it).

tmeasday:
Forum thread: https://forums.meteor.com/t/1-3-3-beta-w-ddp-batching-now-available/22644

mitar:
I discovered a subtle but ugly bug with buffering: #8088

Issue #5664, opened 2015-11-16
Accounts.forgotPassword and Accounts.verifyEmail throw error instead of pass error in callback


mishavee:
Where the email address is blank, or the token is blank, the methods throw a new error, but the docs say they return errors in the callback.

The bad thing about this is undocumented behaviour, plus needing two forms of error handling.

Code in: meteor/packages/accounts-password/password_client.js

stubailo:
I guess there is a bit of distinction of a runtime error that can't be predicted, vs. a programming error where the argument is of the wrong type. In an ideal world, these arguments could be statically checked.

However, I'd definitely accept a PR for returning these errors through the callback instead of throwing them.

mishavee:
Thanks @stubailo, that sounds fun. Is there a guide to submitting a pull-request? Which branch do I clone from, and pull-request to? Can I link it to this issue? Is there automated tests to extend?
"Questions, questions, always questions" - I'm looking at other pull requests to see what to do.

stubailo:
Unfortunately there isn't a great guide about it. But basically:

1. You branch off of `devel` and pull request to `devel`
2. Run the tests with `./meteor test-packages` in the root directory of the repo
3. The tests should be in the same package as the code you'll be changing, it is also probably a good idea to add more tests about the error handling in this case
3. Link to the issue by mentioning the issue in the PR description

mishavee:
Thanks for the good directions :)  I found some similar info in [Contributing-to-Meteor](https://github.com/meteor/meteor/wiki/Contributing-to-Meteor) but your steps are more useful.

mishavee:
Hi @stubailo, I've had fun: cloned the repo, created a new local branch from devel, patched the file, created two tests that pass, and filled in the contributors form. I tried creating a pull request from github desktop, no permission to publish ... same if I try to push the new branch.

Do I need to fork into another github repo just to do a pull request?
Thanks for your time!

stubailo:
Yes, you can only do a pull request on github from a fork, which is kind of
like a branch but on github.

mishavee:
Okay I did a fork, added a branch with the changes, created a pull request to devel: #5681
Now it's waiting for the tests to run. Yay, thanks for your help @stubailo!

laosb:
Fixed in #7117!

PR #5681, opened 2015-11-19
Issue #5664 Client Accounts.forgotPassword and .verifyEmail pass error in callback

mishavee:
To resolve issue #5664.
Accounts.forgotPassword and Accounts.verifyEmail to return error in callback instead of exception.
bugfix in packages/accounts-password/password_client.js
Two additional tests in packages/accounts-password/password_tests.js

### DIFF
--- a/packages/accounts-password/password_client.js
+++ b/packages/accounts-password/password_client.js
@@ -116,7 +116,7 @@ Accounts.createUser = function (options, callback) {
   if (typeof options.password !== 'string')
     throw new Error("options.password must be a string");
   if (!options.password) {
-    callback(new Meteor.Error(400, "Password may not be empty"));
+    callback && callback(new Meteor.Error(400, "Password may not be empty"));
     return;
   }
 
@@ -153,7 +153,7 @@ Accounts.changePassword = function (oldPassword, newPassword, callback) {
 
   check(newPassword, String);
   if (!newPassword) {
-    callback(new Meteor.Error(400, "Password may not be empty"));
+    callback && callback(new Meteor.Error(400, "Password may not be empty"));
     return;
   }
 
@@ -209,7 +209,7 @@ Accounts.changePassword = function (oldPassword, newPassword, callback) {
  */
 Accounts.forgotPassword = function(options, callback) {
   if (!options.email) {
-    callback(new Meteor.Error(400, "Must pass options.email"));
+    callback && callback(new Meteor.Error(400, "Must pass options.email"));
     return;
   }
   Accounts.connection.call("forgotPassword", options, callback);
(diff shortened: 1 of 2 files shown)


packages/accounts-password/password_client.js
```diff
@@ -208,8 +208,10 @@ Accounts.changePassword = function (oldPassword, newPassword, callback) {
  * @param {Function} [callback] Optional callback. Called with no arguments on success, or with a single `Error` argument on failure.
  */
 Accounts.forgotPassword = function(options, callback) {
-  if (!options.email)
-    throw new Error("Must pass options.email");
+  if (!options.email) {
+    callback(new Meteor.Error(400, "Must pass options.email"));
```
zimme: `callback` is optional, you should check that it's available before using it.
Should we throw if there is no `callback`?

packages/accounts-password/password_client.js
```diff
@@ -255,8 +257,10 @@ Accounts.resetPassword = function(token, newPassword, callback) {
  * @param {Function} [callback] Optional callback. Called with no arguments on success, or with a single `Error` argument on failure.
  */
 Accounts.verifyEmail = function(token, callback) {
-  if (!token)
-    throw new Error("Need to pass token");
+  if (!token) {
+    callback(new Meteor.Error(400, "Need to pass token"));
```
zimme: Same as previous comment

mishavee:
Good pickup @zimme, I've altered the code to only use the callback if it is passed in.
This applies to forgotPassword, verifyEmail, createUser, changePassword, resetPassword.
I've updated tests to cover this too.

stubailo:
Hey, this looks great. Unfortunately, I don't have time to look at this in depth right now. I'd suggest one change off the bat, which is that another option would be to throw the type error if a callback is not passed. Otherwise, you'd going to have a function call silently fail with no way for the developer to find out? Or maybe that's the point here, I'm not sure.

mishavee:
Thanks for the feedback @stubailo :)  I think the point is the error is either of interest to the developer, or ignored ... not great practice, but seems appropriate. I'm happy to have another way!

packages/accounts-password/password_client.js
```diff
@@ -116,7 +116,7 @@ Accounts.createUser = function (options, callback) {
   if (typeof options.password !== 'string')
     throw new Error("options.password must be a string");
```
zimme: Shouldn't this be handled by the callback too?

mishavee: I think type checking errors need to be thrown as exceptions to get noticed during development, so the developer will realise they're sending in the wrong thing when they test it.
On the other hand, unhandled exceptions in production are ugly for the users.

tmeasday:

Issue #5650, opened 2015-11-11
Is the Google oAuth 'profile' scope really required?


MasterAM:
Meteor's `google` package requests the 'profile' permission/scope with Google oAuth.

When reading the [Google documentation](https://developers.google.com/+/web/api/rest/oauth#profile), it may seem like this scope is indeed mandatory for getting the user's id. AFAIK, this is not really the case.

The `email` scope is sufficient for getting the required info.
To test this, I forked the `google` package, removed the `profile` scope, leaving only the `email` scope.
It worked very well. The only thing that was missing in the created user object was the `locale` field. 

I believe that this scope can (and should) be made optional,  since requesting unneeded permissions, especially in a very popular authentication mechanism, may deter some users from registering to our apps.

I will gladly submit this as a PR, if you find it worthy.

rdickert:
Thanks for pointing this out! If you are confident it is not necessary, I'd be in favor of removing it. Thanks also for offering to do a PR - I'd say go for it.

MasterAM:
Thanks for the response.

I created a PR for the issue.

I think that the `google` and `accounts-google` packages should be configurable via `Meteor.settings`, since there are some options that are not supported at the moment that I find important, in particular: adding default scopes and deciding whether or not to force the Google account selection popup to appear.

I believe that having that construct in place will make it easier to set options that override the defaults and can, in turn, be overridden by the options supplied to `Meteor.loginWithGoogle()`.

This may also make keeping backward-compatibility easier.

What do you think about this approach?

laosb:
Close since the PR was merged and the functionality was released in 1.4.0.1.

PR #5699, opened 2015-11-23
Using the google `email` scope for oAuth.

MasterAM:
This PR is related to https://github.com/meteor/meteor/issues/5650.

The `email` scope is enough for getting the required data, including the
user ID, so no need for the `profile` scope, which may deter users.

### DIFF
--- a/packages/google/google_client.js
+++ b/packages/google/google_client.js
@@ -23,9 +23,9 @@ Google.requestCredential = function (options, credentialRequestCompleteCallback)
 
   var credentialToken = Random.secret();
 
-  // always need this to get user id from google.
-  var requiredScope = ['profile'];
-  var scope = ['email'];
+  // we need the email scope to get user id from google.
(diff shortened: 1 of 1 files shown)


timbotnik:
This seems reasonable, thanks for the PR!  However it may have implications for apps that are relying on certain fields being there.  Could you identify the backwards compatibility issues to watch out for here?  We'll take a look at this when we can.

MasterAM:
AFAICT, only the `locale` field was missing when I did not use the `profile` scope.
Only a handful of fields are whitelisted by the server side by default anyway.

I think that there can be some improvements made to the package configurability that will make this sort if compatibility issues easier to resolve (or nonexistent).
I will add those to the issue thread.

MasterAM:
Any thought about this merging this PR before the packages are ripped out of the repository?

tmeasday:

PR #5702, opened 2015-11-24
Use methods when we have them

mitar:


mitar:
BTW, `getCordovaPlatforms` is really ugly. Assumption that any non-default platform is a Cordova platform? Where does this assumption come from?

stubailo:
This makes sense. And I think the assumption that all platforms are cordova platforms is a historical thing right now. Clearly if in the future we have electron or similar, we'll need to do a bit of refactoring to remove that assumption.

mitar:
Yea, I am just adding one extra platform now and it is painful. :-(

arboleya:
@mitar Can you tell what extra platform is that?

mitar:
Not yet.

@stubailo: In fact at the end it was pretty easy. Amazing.

arboleya:
@mitar If it's about a Desktop layer based on Electron and you're inclined to have a chat about it, even on pvt, I started a [project](https://github.com/arboleya/electrify) a while ago and have evolved some [ideas](https://github.com/arboleya/electrify/issues/24) ever since, besides some other things I have in mind.

mitar:
That is great. And no, it is not that. But from my experience now it is not really so hard to create a new target. BTW, have you seen [this project](https://github.com/peerlibrary/meteor-package-builder).

arboleya:
Nope, never heard of. Is it abandoned? What it's all about? Gotta check it later.

arboleya:
Ok, I see you've wrapped Meteor in order to provide a `package` method access **from inside** Meteor itself. But why? Since we cannot execute a specific Meteor method without starting Meteor first, this would generate an extra step for running stuff.

I mean, Electrify was firstly built as a standalone Meteor package, in which one would have to open a second tab to run  `meteor shell` in order to call Electrify's (server) methods. It was awful, so I've reverted it to a more practical npm package, and If I understood what you were trying to achieve, the same `meteor shell` thing would be needed. It's something to consider.

As for the `node-meteor-package-builder` repo, does it suggests similar possibilities of those achieved using the [`publish-release`](http://practicalmeteor.com/using-meteor-publish-release-to-extend-the-meteor-command-line-tool/) command?

Well, let me know if there's something I didn't get. This topic interest me. Tks

PR #5703, opened 2015-11-24
Be more precise which platforms are Cordova

mitar:


timbotnik:
Thanks!

mitar:
Hm, tests are failing, but I would not say that because of this pull request?

timbotnik:
Yeah looks like the failures are `ecmascript` related.

zol:
Closing this as the PR is lacking an explanation or ticket reference and the code it targets has since been refactored.

PR #5710, opened 2015-11-26
Make sure no user configuration is used

mitar:
Reproducible builds and stuff.

### DIFF
--- a/scripts/generate-dev-bundle.sh
+++ b/scripts/generate-dev-bundle.sh
@@ -53,6 +53,8 @@ cp -R "$DIR/lib/node_modules/npm" "$DIR/bundled-npm"
 # export path so we use the downloaded node and npm
 export PATH="$DIR/bin:$PATH"
 
+# Disable any NPM user configuration so that there are no differences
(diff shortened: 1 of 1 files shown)


stubailo:
Can you add a comment so that the next person who comes along knows why this is here?

mitar:
Done.

benjamn:
Looks good to me, and rebasing my https://github.com/meteor/meteor/tree/npm3 against these changes should be trivial. Thanks for reporting the problem and providing a solution!

benjamn:
After thinking about this some more, I decided it's better to forbid user configuration files dynamically in https://github.com/meteor/meteor/blob/devel/tools/isobuild/meteor-npm.js than to obscure the lookup path when we build npm (which is what https://github.com/meteor/meteor/commit/adaf28b4aabd25cb2e8515bffc78a88e89dbe80b did, if I understand correctly).

mitar:
I have not found a way to forbid user configuration files dynamically. There is no command line switch for this for it that I would found. Or am I missing something?

PR #5713, opened 2015-11-26
Serve manifest.json file correctly for the main architecture

mitar:
Otherwise it is stored into the `staticFiles` as `manifest.json` (instead of `/manifest.json`) which is then never matched.


stubailo:
Looks reasonable!

PR #5714, opened 2015-11-26
Correct URLs for non-main architectures

### LINKED ISSUE #5706: Manifest contains invalid URLs for non-standard architectures
If I have a non-standard architecture `web.foobar`, then its manifest is served under `/__foobar/manifest.json`, which is great, but the manifest itself has URLs which are not rooted under that path:

```
{
  path: "packages/meteor.js",
  where: "client",
  type: "js",
  cacheable: true,
  url: "/packages/meteor.js?2a844cb96311115f63cb890c87c79881233bdabc",
  sourceMap: "packages/meteor.js.map",
  sourceMapUrl: "/packages/2a844cb96311115f63cb890c87c79881233bdabc.map",
  size: 712,
  hash: "2a844cb96311115f63cb890c87c79881233bdabc"
},
```

See `url` and `sourceMapUrl`. They should be `/__foobar/packages/meteor.js?2a844cb96311115f63cb890c87c79881233bdabc` and `/__foobar/packages/2a844cb96311115f63cb890c87c79881233bdabc.map`, respectively.

mitar:
Fixes #5706.

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -466,12 +466,10 @@ var runWebAppServer = function () {
         if (! clientJsonPath || ! clientDir || ! clientJson)
           throw new Error("Client config file not parsed.");
 
-        var urlPrefix = getUrlPrefixForArch(arch);
-
         var manifest = clientJson.manifest;
         _.each(manifest, function (item) {
           if (item.url && item.where === "client") {
-            staticFiles[urlPrefix + getItemPathname(item.url)] = {
+            staticFiles[getItemPathname(item.url)] = {
               absolutePath: path.join(clientDir, item.path),
               cacheable: item.cacheable,
               hash: item.hash,
@@ -483,7 +481,7 @@ var runWebAppServer = function () {
             if (item.sourceMap) {
               // Serve the source map too, under the specified URL. We assume all
               // source maps are cacheable.
-              staticFiles[urlPrefix + getItemPathname(item.sourceMapUrl)] = {
+              staticFiles[getItemPathname(item.sourceMapUrl)] = {
                 absolutePath: path.join(clientDir, item.sourceMap),
                 cacheable: true
               };
@@ -499,6 +497,8 @@ var runWebAppServer = function () {
         };
 
         WebApp.clientPrograms[arch] = program;
+
+        var urlPrefix = getUrlPrefixForArch(arch);
 
         // Serve the program as a string at /foo/<arch>/manifest.json
         // XXX change manifest.json -> program.json
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -422,6 +422,7 @@ WebAppInternals.staticFilesMiddleware = function (staticFiles, req, res, next) {
   }
 };
 
+// XXX Duplicate of a function in bundler.js.
 var getUrlPrefixForArch = function (arch) {
   // XXX we rely on the fact that arch names don't contain slashes
   // in that case we would need to uri escape it
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -357,16 +357,29 @@ class File {
     this.cacheable = true;
   }
 
+  // XXX Duplicate of a function in webapp_server.js.
+  getUrlPrefixForArch(arch) {
+    // XXX we rely on the fact that arch names don't contain slashes
+    // in that case we would need to uri escape it
+
+    // We add '__' to the beginning of non-standard archs to "scope" the url
+    // to Meteor internals.
+    return arch === 'web.browser' ?
+      '' : '/' + '__' + arch.replace(/^web\./, '');
+  }
+
   // Given a relative path like 'a/b/c' (where '/' is this system's
   // path component separator), produce a URL that always starts with
   // a forward slash and that uses a literal forward slash as the
   // component separator.
-  setUrlFromRelPath(relPath) {
+  setUrlFromRelPath(relPath, arch) {
     var url = relPath;
 
     if (url.charAt(0) !== '/') {
       url = '/' + url;
     }
+
+    url = this.getUrlPrefixForArch(arch) + url;
 
     // XXX replacing colons with underscores as colon is hard to escape later
     // on different targets and generally is not a good separator for web.
@@ -767,7 +780,7 @@ class Target {
         f.setTargetPathFromRelPath(relPath);
 
         if (isWeb) {
-          f.setUrlFromRelPath(resource.servePath);
+          f.setUrlFromRelPath(resource.servePath, this.arch);
         } else {
           unibuildAssets[resource.path] = resource.data;
         }
@@ -799,7 +812,7 @@ class Target {
           f.setTargetPathFromRelPath(relPath);
 
           if (isWeb) {
-            f.setUrlFromRelPath(resource.servePath);
+            f.setUrlFromRelPath(resource.servePath, this.arch);
           }
 
           if (resource.type === 'js' && isOs) {
@@ -924,7 +937,7 @@ class Target {
         }
 
         if (file.path) {
-          newFile.setUrlFromRelPath(file.path);
+          newFile.setUrlFromRelPath(file.path, this.arch);
           newFile.targetPath = file.path;
         } else {
           newFile.setUrlToHash('.js', '?meteor_js_resource=true');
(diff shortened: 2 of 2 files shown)


stubailo:
@avital, could you take a look at this? Looks like it doesn't hurt anything, and unblocks something that @mitar is doing.

mitar:
Just to be clear, it is not blocking me, because I have to anyway use a fork to support new target. But I am sending pull requests to Meteor for things I had to fix to support the new target. After pulling all these pull requests in then it is pretty straightforward to add another target. So those are general things which should fix overall Meteor support for custom targets.

So I am not sure who is currently consumer of `manifest.json`. I assume that Cordova is? Does it ever read `/__cordova/manifest.json`? Because with this pull request URLs in that manifest would change (to correct ones), but maybe the consumer is "fixing" them on their side?

martijnwalraven:
The issue here is indeed that Cordova downloads `/__cordova/manifest.json` during a hot code push and currently expects the URLs to be relative to the manifest path (ignoring the `/` at the beginning).

This seems like a reasonable change, but it has to be coordinated with updating the Cordova plugin in. I'm also wondering what this would mean for backwards compatibility (e.g. an app with an older version of the plugin connecting to a server with this change applied).

mitar:
> So this seems like a reasonable change, but it has to be coordinated with updating the Cordova plugin. 

I can look into that, but I cannot test it.

> I'm also wondering what this would mean for backwards compatibility (e.g. an app with an older version of the plugin connecting to a server with this change applied).

In theory we could update `format` to signal that, but code is not checking for `format` anywhere, so this would not really help. On the other hand we could make new code check `format` and if it is old version, it uses the old way of parsing URLs, if new, then the new one. But this would of course make autoupdating code pretty ugly, handling both cases.

martijnwalraven:
The [rewritten Cordova plugin](https://github.com/meteor/cordova-plugin-meteor-webapp) does check for `format`, ignoring the update if it is anything other than `web-program-pre1`. It would have been nice if the old plugin did the same, so we could simply update the format :)

mitar:
So what should we do?

mitar:
(BTW, I also extended manifest with a list of dependencies each file has, which allows then to have dynamic loading of files (you know which other files you have to load, if they are not already). Is this something MDG would be interested in having in core?)

martijnwalraven:
I don't see an easy solution for now, so I'm tempted to leave it as is, especially so soon before a release. Is it an option for you to also interpret these URLs as relative, the same way the Cordova plugin does?

We probably want to make more changes to the build and packaging system after Meteor 1.3, so it may make more sense to wait until then.

martijnwalraven:
I'm not sure what you mean by dynamic loading of files here?

mitar:
So is the code in the meteor repository in use, for cordova plugin? If so, then we can at least fix one direction (people with new plugin connecting to old Meteor instances).

> I'm not sure what you mean by dynamic loading of files here?

This is unrelated to this pull request. But it makes such an entry:

```
    {
      "path": "packages/ejson.js",
      "where": "client",
      "type": "js",
      "cacheable": true,
      "url": "/packages/ejson.js?a021b77b39385122cd8e4eab4df0d0bfe299c9df",
      "dependencies": [
        "packages/meteor.js",
        "packages/underscore.js",
        "packages/base64.js"
      ],
      "sourceMap": "packages/ejson.js.map",
      "sourceMapUrl": "/packages/a021b77b39385122cd8e4eab4df0d0bfe299c9df.map",
      "size": 81713,
      "hash": "a021b77b39385122cd8e4eab4df0d0bfe299c9df"
    }
```

This allows in my code one to do something like:

```
dependenciesReady('packages/ejson.js').then(function () {
  // EJSON and everything else it needs is now loaded.
});
```

And keep the list of `<script>` tags initially send by Meteor very small.

I am just asking if this is something which would be of interest and I can open a ticket, pull request, or something, and we can discuss it further there.

mitar:
>  Is it an option for you to also interpret these URLs as relative, the same way the Cordova plugin does?

Hardly, because in my case I also allow remote URLs, CDNs and stuff in the manifest.

martijnwalraven:
Yes, the plugin is in use on the `cordova-improvements` branch, which is where the [Meteor 1.3 Cordova betas](https://github.com/meteor/meteor/issues/6050#issuecomment-180843752) are built from.

I don't see how allowing the new plugin to interpret both old and new versions of the manifest would help solve backwards compatibility. We still wouldn't be able to update the manifest version we serve because the old plugin wouldn't know how to deal with it.

We're definitely interested in improving loading performance, and that looks like a smart solution. It would be great to hear more about your experiences and get your input on the design of a future build system. If possible, I think we'd prefer a more automated approach integrated with the modules system and normal importing behavior. Webpack has mechanisms for code splitting that do something similar for instance.

Also wondering, how does your dependency system help you in production, where there would only be one minified file?

mitar:
> I don't see how allowing the new plugin to interpret both old and new versions of the manifest would help solve backwards compatibility. We still wouldn't be able to update the manifest version we serve because the old plugin wouldn't know how to deal with it.

OK, so the solution is to release 1.3 with plugin which is checking the format, and then in 1.4 we change the format and fix the paths?

> If possible, I think we'd prefer a more automated approach integrated with the modules system and normal importing behavior. Webpack has mechanisms for code splitting that do something similar for instance.

Yes, this is done currently only around the concept of packages. So this is done around Meteor 1.2.

With webpack probably manifest file should also be updated and instead of having a list of packages, it should be a list of modules, no? But then the idea that you have list of dependencies would work the same. Only units would be smaller.

> Also wondering, how does your dependency system help you in production, where there would only be one minified file?

We minimize each package file separately.

mitar:
Where should we discuss dynamic loading of packages? I think I have a working solution. Pretty simple in fact.

martijnwalraven:
For now, I think opening an issue is probably the best way. In the future this seems like a good candidate for a design discussion, perhaps in a separate repo.

mitar:
Opened #6203.

mitar:
Fixing the URLs in the manifest file can also then solve the assets URL case. Then you could resolve the asset by its path into the full URL by reading the manifest file.

mitar:
@martijnwalraven, will Meteor 1.3 have this new Cordova plugin? If not, then maybe we should add to the 1.3 the code which allows it to parse both formats, and then in 1.4 we move to generate the correct URLs.

mitar:
See https://github.com/meteor/meteor/issues/6222 for more motivation why this would be useful to be solved early.

benjamn:

mitar:
Can we just reopen?  There is little benefit in loosing all this conversation in the new pull request?

benjamn:
Absolutely. Sorry the PR reaping process is kinda blunt :)

stubailo:
I love these conversations with copy-pasted text on multiple issues, lol

benjamn:
@martijnwalraven can we just merge this now? Any lingering objections?

mitar:
> I love these conversations with copy-pasted text on multiple issues, lol

Oh, and I thought I was talking to a bot. :-) So it is a manual bot, real I, not AI. :-)

benjamn:
AAI, if you will.

mitar:
@benjamn: No, I do not think this can be merged as it is now. From my understanding, we should bump the manifest version number, and Cordova plugin should be upgraded to support new and old version file, so to resolve URLs based on the version file.

If @martijnwalraven agrees, I can look into this.

mitar:
Augmented and assisted intelligence? ;-)

martijnwalraven:
Yeah, this would break hot code push on existing Cordova apps, both those running on Meteor 1.3 and earlier. Although the plugin does check the manifest version in Meteor 1.3, it just fails if it isn't the supported version.

Working on changing this only makes sense if we are planning to add more architectures. As I understand it, Mitar's main use case is https://github.com/meteor/meteor/issues/6222. There is some discussion on that issue, but basically I felt adding a separate architecture for web workers would complicate the build process further and it would be better to wait until we can support multiple bundles in the build tool. Curious to hear what you think about this @benjamn.

mitar:
> Yeah, this would break hot code push on existing Cordova apps, both those running on Meteor 1.3 and earlier. Although the plugin does check the manifest version in Meteor 1.3, it just fails if it isn't the supported version.

And I tried to get code in which would make [it more robust](https://github.com/meteor/meteor/pull/5714#issuecomment-182157355). :-(

No, my use case is to provide alternative architectures altogether. Web workers were just one example I made because I was already working on other aspects of adding more architectures.

zol:
Hey @mitar - Took another look at this and chatted about it with the team. As it's not a trivial change we're not going to take this PR until there's clear demand for multiple architectures from others in the community. It's not on our roadmap to add more architectures at this stage.

PR #5715, opened 2015-11-26
Update AccountsUIWrapper.jsx

okland:
Fix warning of using deprecated React.findDOMNode


stubailo:
Actually isn't the correct approach now to use `this.refs.container` directly?

stubailo:
We need to update the tutorial, that will automatically update this app. We're going to do this before 1.3 release.

okland:
+1

PR #5716, opened 2015-11-26
Update simple-todos-react.jsx

okland:
Minor fix user ReactDOM.render instead of React.render - remove warning in console. :)


stubailo:
Thanks!

PR #5729, opened 2015-11-29
Providing access to the raw access token data in OAuth1.

chfritz:
Extended oauth1_bindings with just one line to return the raw access
token response to the login handler for the service. This is required
as some services do provide additional information together with the
access token (http://oauth.net/core/1.0/#rfc.section.6.3.2). Currently
in the meteor oauth1 package only the `accessToken` and
`accessTokenSecret` are stored.

This is similar to `rauth`'s get_raw_acces_token in python
(http://rauth.readthedocs.org/en/latest/api/#rauth.OAuth1Service.get_raw_access_token).

### DIFF
--- a/packages/oauth1/oauth1_binding.js
+++ b/packages/oauth1/oauth1_binding.js
@@ -67,6 +67,7 @@ OAuth1Binding.prototype.prepareAccessToken = function(query, requestTokenSecret)
 
   self.accessToken = tokens.oauth_token;
   self.accessTokenSecret = tokens.oauth_token_secret;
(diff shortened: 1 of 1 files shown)


stubailo:
Which login services is this relevant for?

chfritz:
This was required for Zotero (https://atmospherejs.com/chfritz/accounts-zotero).

https://www.zotero.org/support/dev/web_api/v3/oauth:
"In addition to receiving the token, API consumers using OAuth will need to retrieve the user's user ID from the response parameters returned by Zotero.org."

tmeasday:

Issue #5441, opened 2015-10-15
Blaze should not throw error when calling a function on falsy data context


jagi:
Hi, I've noticed inconsistency in how Blaze treats rendering template with not existing data. Let's say we have the following template and data context is empty.

```hbs
{{user.firstName}}
```

It will just print nothing and no error will be printed. Not let's say we have the following template.

```hbs
{{user.fullName 'Mr.'}}
```

Now with empty data context the following error will be thrown `Uncaught Error: Can't call non-function: undefined`. Here is a reproduction repository https://github.com/jagi/blaze-function-error.

I was able to make a workaround using the `#with` expression (but it's ugly) or I can just omit displaying template using the `Template.subscriptionsReady` helper. However, I want to display template even if there is no data. I want the template to be filled with data when subscriptions are ready.

zimme:
It expects `user.fullName` to be a template helper as you provide arguments to it? and as `user.fullname` is `undefined` it's not a template helper function which can be called?

zimme:
But you're right, there needs to be a check if the function you wanna call isn't a registered template helper that the function from the data context is present before calling it, and just do nothing otherwise. for the behaviour to be consistent with the rest of blaze behaviour.

jagi:
If the `user.full` name is a function in the `user` object then it's possible to pass arguments to this function as it would be a helper. It's a nice way of abstracting helpers from template.

However, right now if the `user` object is empty then it throws error and it shouldn't because it does not throw error when it's property.

stubailo:
Yeah, I think Blaze usually goes out of its way to avoid these kinds of runtime null errors, so we must have missed a spot there.

I'd be happy to look at a PR improving this.

stubailo:
Should be fixed!

jagi:
Great, thanks!

mitar:
Interesting, I think I reported that in #4230, but @glasser didn't like the idea. But now it just got merged.

Comparing the change, I think mine was similar: https://github.com/peerlibrary/meteor-blaze-components/issues/21

I would argue though that there is one extra change needed. Change from:

```
return value.apply(null, newArgs);
```

To:

```
var res = value.apply(null, newArgs);
if (res == null) return null;
return res;
```

Also, if `value == null`, I would always return `null`, even when `value === undefined`. Otherwise things like `{{> user.firstNameTemplate}}` explode.

PR #5736, opened 2015-12-01
Fix https://github.com/meteor/meteor/issues/5441

skishore:
I've verified that the new tests pass. Two of the new test cases were broken prior to this change. I've also verified that this change fixes the example program from the issue page, in a really hacky way - I manually patched the fix into the build directory. What's the better way to verify a package fix?


stubailo:
Looks great! Thanks.

Fixes #5441

PR #5743, opened 2015-12-03
Unnecessary variable

mitar:
Probably makes things a bit faster.


mquandalle:
There have been several unused variables in the tools (eg.  #3235). It should be fairly easy to run a static analysis tool like EsLint to automate their removal.

mitar:
Yes, but this one computes something. :-) So slows down stuff.

stubailo:
Looks great!

PR #5750, opened 2015-12-04
Tinytest README - first draft

robfallows:
@tmeasday : PR as requested. Note I have retained the guide style (it's nice and friendly), but removed references that suggest this is a part of the guide.

Also, I've removed the "how do I test ..." sections. I no longer think they are appropriate in the context of a README, and especially as the official guide position will not be recommending Tinytest. There are resources around which go into this in some detail and links to those should probably be added.

I've added a reference to smithy:describe as you asked.

### DIFF
--- a/packages/tinytest/README.md
+++ b/packages/tinytest/README.md
@@ -1,26 +1,313 @@
-This is an internal Meteor package.
+# TinyTest
 
-API:
-test.ok(doc)
-test.expect_fail()
-test.fail(doc)
-test.exception(exception)
-test.runId()
-test.equal(actual, expected, message, not)
-test.notEqual(actual, expected, message)
-test.instanceOf(obj, klass, message)
-test.notInstanceOf(obj, klass, message)
-test.matches(actual, regexp, message)
-test.notMatches(actual, regexp, message)
-test.throws(func)
-test.isTrue(v, msg)
-test.isFalse(v, msg)
-test.isNull(v. msg)
-test.isNotNull(v, msg)
-test.isUndefined(v, msg)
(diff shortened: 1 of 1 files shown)


stubailo:
This is awesome!

stubailo:
I think this is much better than what we have now, and correct as far as I can see. Thank you!

tmeasday:
Thanks so much @robfallows, this is really great

robfallows:
@stubailo, @tmeasday : you're very welcome. Thanks for the opportunity to contribute :smile:

packages/tinytest/README.md
```diff
@@ -1,26 +1,314 @@
...
+
+> ... Ideally, each test case is independent from the others. Substitutes such as method stubs, mock objects, fakes, and test harnesses can be used to assist testing a module in isolation.
+
+The object of a unit test is to prove functionality you have written. It is not to prove third-party or other package or library code which may form a part of your codebase. This may require that you head up your tests with *stubs, mocks/spies* or *fakes* which satisfy the demands of your code, but are non-functional (or have limited functionality). Alternatively, instead of placing the stubs in your test file, you could put them into a separate `stubs.js` file, which you `api.addFiles`, or even create and add a stubs package which you then `api.use`.
+
+From [martinfowler.com](http://www.martinfowler.com/articles/mocksArentStubs.html):
+
+> Fake objects actually have working implementations, but usually take some shortcut which makes them not suitable for production (an in memory database is a good example).
```
covall: Minimongo is a good example that the last example here is not correct or rather out of date.

PR #5752, opened 2015-12-05
duplicate class names fix #5632

### LINKED ISSUE #5632: Repeated class names not rendered correctly
Using Semantic-UI sometimes you need to add the same class name twice, i.e.

```html
<div class="ui two column computer one column mobile grid container">
```

but this is translated into
```html
<div class="ui two column computer one mobile grid container">
```
breaking the interface.
This is because of [DiffingAttributeHandler](https://github.com/meteor/meteor/blob/devel/packages/blaze/attrs.js#L68) call `this.parseValue`which  reduces the repeated element into one since the element is the key in the array where values are stored
```javascipt
parseValue: function (attrString) {
    var tokens = {};

    _.each(attrString.split(' '), function(token) {
      if (token)
        tokens[token] = token;
    });
    return tokens;
  }
```

Here is the repo to reproduce the error with step by step guide: https://github.com/bitIO/repeated-class-issue

talha-asad:
Fix for semantic ui duplicate classes and other similar frameworks that make use of repeated classes with ordering.

### DIFF
--- a/packages/blaze/attrs.js
+++ b/packages/blaze/attrs.js
@@ -102,9 +102,13 @@ var ClassHandler = DiffingAttributeHandler.extend({
   parseValue: function (attrString) {
     var tokens = {};
 
-    _.each(attrString.split(' '), function(token) {
-      if (token)
-        tokens[token] = token;
+    var similarTokens = 0;
+    _.each(attrString.split(' '), function(token, i) {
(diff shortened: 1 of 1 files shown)


talha-asad:
Sorry i broke a few contribution guidelines. I'll submit another one.

PR #5753, opened 2015-12-05
duplicate class names fixes #5632

### LINKED ISSUE #5632: Repeated class names not rendered correctly

talha-asad:
***2 Upvotes*** Fix for Semantic UI duplicate classes and other similar frameworks that make use of repeated classes where order of classes matter.

It would be appreciated if you could get this merged asap.

### DIFF
--- a/packages/blaze/attrs.js
+++ b/packages/blaze/attrs.js
@@ -102,9 +102,13 @@ var ClassHandler = DiffingAttributeHandler.extend({
   parseValue: function (attrString) {
     var tokens = {};
 
-    _.each(attrString.split(' '), function(token) {
-      if (token)
-        tokens[token] = token;
+    var similarTokens = 0;
+    _.each(attrString.split(' '), function(token, i) {
+      if (token) {
+        // Duplicate class names (e.g: Semantic UI)
+        if (tokens[token] && i > 0) tokens[Object.keys(tokens)[(i-similarTokens++)-1]] += ' ' + token;
+        else tokens[token] = token;
+      }
     });
     return tokens;
   }
--- a/packages/blaze/attrs.js
+++ b/packages/blaze/attrs.js
@@ -106,7 +106,7 @@ var ClassHandler = DiffingAttributeHandler.extend({
     _.each(attrString.split(' '), function(token, i) {
       if (token) {
         // Duplicate class names (e.g: Semantic UI)
-        if (tokens[token] && i > 0) tokens[Object.keys(tokens)[(i-similarTokens++)-1]] += ' ' + token;
+        if (tokens[token] && i > 0) tokens[Object.keys(tokens)[(i - similarTokens++) - 1]] += ' ' + token;
         else tokens[token] = token;
       }
     });


stubailo:
Can you add a comment explaining what the following code does?

```js
tokens[Object.keys(tokens)[(i - similarTokens++) - 1]] += ' ' + token;
```

I think it could be refactored a bit to make it more readable, rather than a complex one-liner.

stubailo:
Also, we'll need tests to make sure this doesn't break in the future.

talha-asad:
I am using it locally and so far i haven't had a situation where it would have caused a problem. 

Basically what i do is if the same classname is repeated, i attach it to the dict's last value after a space. This preserves the order of the classes. This only happens if the same class is repeated again. In other words, instead of having a dict with exactly the same keys and values, we group more than one class if it repeats and lock it with the classname it appeared with.

I could split it up in multiple lines, but i think its best if we illustrate it with an example:
```js
//For non-repetitive (e.g "ui calendar right") classes we produce the following dict:

{
  "ui": "ui",
  "calendar": "calendar",
  "right": "right"
}

// For repetitive (e.g "ui middle aligned center aligned segment") classes we product the following dict:
{
  "ui": "ui",
  "middle": "middle",
  "aligned": "aligned",
  "center": "center aligned",
  "segment": "segment"
}

```

talha-asad:
Let me know what is a suitable comment for this and i'll get it in. Just seemed a bit difficult for me to explain this with a comment, might need some help in that regard ;)

I ran it through the tests that were already part of the package and it passed them all.

talha-asad:
@stubailo I'd want to get this in 1.3 and i know you guys are pretty busy but i think this is a small but vital fix, if possible please help me get this in the core. I know Blaze isn't much of a priority anymore but i think this would be fundamental for something like Blaze 2 as well.

mikkhait:
Please accept this PR

masca88:
:+1: 

> @stubailo I'd want to get this in 1.3 and i know you guys are pretty busy but i think this is a small but vital fix, if possible please help me get this in the core. I know Blaze isn't much of a priority anymore but i think this would be fundamental for something like Blaze 2 as well.

franzip:
:+1: 

Would be awesome to see it in Meteor 1.3!

jlukic:
:+1: :wink:

talha-asad:
I think we still need a good comment to explain what i did lol. I am bad at coming up with good comments.

voidale:
+1 just got hit by this with semantic ui

cluxter:
+1 still not working here despite having the latest version of Meteor and semantic-ui.

tmeasday:
Thanks for this PR, sorry we never got around to merging it. We are currently cleaning up the repository and removing old PRs; if you think this PR still has value and are interested in resurrecting it, we encourage you to re-submit it against https://github.com/meteor/blaze

PR #5771, opened 2015-12-09
Enable ES2015 async/await by default in "ecmascript" package

avital:


packages/ecmascript-runtime/package.js
```diff
@@ -7,17 +7,20 @@ Package.describe({
...
 
 Npm.depends({
-  "meteor-ecmascript-runtime": "0.2.6"
+  "meteor-ecmascript-runtime": "0.2.6",
+  "regenerator": "0.8.42"
 });
 
 Package.onUse(function(api) {
   api.use("modules");
+  api.use("promise");
```
benjamn: Maybe comment that we're only depending on `promise` because regenerator needs it?

avital: sure

Slava:
yay?

avital:
yay!

PR #5778, opened 2015-12-11
Allow new Mongo.Collection to be called multiple times for the same collection name

aldeed:
- No error is thrown when constructing a `Mongo.Collection` twice for the same name
- If mutation methods were already created for a named collection, we skip them and do not throw an error
- `Mongo.Collection` constructor takes a new option, `defineMutationMethods`, which can be set to false to skip creating the default mutation methods
- `_defineMutationMethods` function performance is slightly improved by returning earlier if there is no need to loop through method types

This could obviously have some edge cases effects I'm sure, but I can't think of any and all tests still pass. Maybe something weird would happen if defining it twice with different types of _id?

There is of course the chance that two packages, or a package and app, unknowingly use the same collection name and don't get an error. That possibility is not reason enough to keep throwing the error, though, IMO. (In fact this error often bites me when I _want_ to declare the same mongo collection as a package because the package does not export their reference.) Perhaps just a console.warn?

The reason for `defineMutationMethods: false` is so that packages can subclass `Mongo.Collection` and provide their own validated mutation methods, turning off the default ones.

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -22,6 +22,7 @@ Mongo = {};
 
 The default id generation technique is `'STRING'`.
  * @param {Function} options.transform An optional transformation function. Documents will be passed through this function before being returned from `fetch` or `findOne`, and before being passed to callbacks of `observe`, `map`, `forEach`, `allow`, and `deny`. Transforms are *not* applied for the callbacks of `observeChanges` or to cursors returned from publish functions.
+ * @param {Boolean} options.defineMutationMethods Set to `false` to skip setting up the mutation methods that enable insert/update/remove from client code. Default `true`.
  */
 Mongo.Collection = function (name, options) {
   var self = this;
@@ -209,21 +210,41 @@ Mongo.Collection = function (name, options) {
       getDoc: function(id) {
         return self.findOne(id);
       },
-      
+
       // To be able to get back to the collection from the store.
       _getCollection: function () {
         return self;
       }
     });
 
-    if (!ok)
-      throw new Error("There is already a collection named '" + name + "'");
+    if (!ok) {
+      const message = `There is already a collection named "${name}"`;
+      if (options._suppressSameNameError === true) {
+        // XXX In theory we do not have to throw when `ok` is falsy. The store is already defined
+        // for this collection name, but this will simply be another reference to it and everything
+        // should work. However, we have historically thrown an error here, so for now we will
+        // skip the error only when `_suppressSameNameError` is `true`, allowing people to opt in
+        // and give this some real world testing.
+        console.warn ? console.warn(message) : console.log(message);
+      } else {
+        throw new Error(message);
+      }
+    }
   }
 
   // XXX don't define these until allow or deny is actually used for this
   // collection. Could be hard if the security rules are only defined on the
   // server.
-  self._defineMutationMethods();
+  if (options.defineMutationMethods !== false) {
+    try {
+      self._defineMutationMethods({ useExisting: (options._suppressSameNameError === true) });
+    } catch (error) {
+      // Throw a more understandable error on the server for same collection name
+      if (error.message === `A method named '/${name}/insert' is already defined`)
+        throw new Error(`There is already a collection named "${name}"`);
+      throw error;
+    }
+  }
 
   // autopublish
   if (Package.autopublish && !options._preventAutopublish && self._connection
--- a/packages/allow-deny/allow-deny.js
+++ b/packages/allow-deny/allow-deny.js
@@ -70,8 +70,9 @@ CollectionPrototype.deny = function(options) {
   addValidator(this, 'deny', options);
 };
 
-CollectionPrototype._defineMutationMethods = function() {
+CollectionPrototype._defineMutationMethods = function(options) {
   const self = this;
+  options = options || {};
 
   // set to true once we call any allow or deny methods. If true, use
   // allow/deny semantics. If false, use insecure mode semantics.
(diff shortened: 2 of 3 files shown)


projectn00b:
Can we have collection hooks baked in?

packages/mongo/collection.js
```diff
@@ -210,15 +210,15 @@ Mongo.Collection = function (name, options) {
         return self.findOne(id);
       },
     });
-
-    if (!ok)
-      throw new Error("There is already a collection named '" + name + "'");
+    
+    // If ok is false, that means we already registered the store for this collection name, but
```
serkandurusoy: I think the best option would be a `warnIfCollectionExists` property on the Collection constructor and then a warning on the console if it set.

stubailo:
> Can we have collection hooks baked in?

An easy way to write collection hooks in your app would be to use ES2015 classes to override certain methods on the collection, like: http://guide.meteor.com/collections.html#hooks

going to take a look at this PR today!

serkandurusoy:
> override certain methods on the collection

@stubailo that would not be chained, though, right? The last override wins!

stubailo:
@serkandurusoy yes - and there are big benefits to avoiding action at a distance like that. It could be pretty hard to understand what is going on in your app if any package or part of your app could add hooks to any collection.

serkandurusoy:
I use hooks heavily and often find the need to contextually divide them based on what kind of task the hook is going to do. I also heavilty use a pattern where client/server hooks run different tasks.

stubailo:
What do you use them for? Perhaps hooks are a crucial feature, it would be good to know what you're running into.

serkandurusoy:
A few frequent usecases I have for hooks (matb33:collection-hooks) are

* Modify the document
 * Add default values
 * Apply constraints
 * Clean up some fields
 * Set or unset conditional fields (fields that depend on the value of another field or document or collection)
* Creating notifications, in the form of
 * email,
 * a record on a notifications collection, 
 * a call to an api (internal or external)
 * setting some reactive datasource on the client
* Update some related item on another collection
 * some counts,
 * denormalized fields
 * continue on a series of actions that may or may not need to be within a transaction

These all do have other means of achieving the desired results, but I've grown quite fond of a hooks-based design because it does provide context around events and actions and allow me to centralize them, similar *in some ways* to the dispatcher pattern of flux.

*PS: Please let me know if you want to carry this over to another thread.*

aldeed:
I don't know why CI tests failed but they all pass locally.

stubailo:
I just realized I haven't thought of this - what happens if you declare the same collection twice on the client? Do they both get the data?

serkandurusoy:
Hmm, right. Perhaps a better api would be for collection2 to provide something like `baseCollection` that just returns the non-decorated version of the collection.

zodern:
In zodern:pureadmin-mongo it sometimes declares the same collection twice. MiniMongo throws an error:
```
Error: There is already a collection named 'queue'
    at new p.Collection (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:62:3310)
    at eval (eval at <anonymous> (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:180:54), <anonymous>:78:18)
    at http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:80:2042
    at Function.e._withTemplateInstanceFunc (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:80:4982)
    at n (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:80:1948)
    at null.<anonymous> (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:80:2826)
    at n (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:79:19836)
    at Object.t.nonreactive (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:31:4260)
    at http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:79:19748
    at Object.d._withCurrentView (http://localhost:4000/packages/zodern_pure-admin/output/pure-admin.js:79:22545)

```

projectn00b:
One of the major issues with matb33:collection-hooks is that you cannot do
a mongodb call within a hook. Say for instance you had to get a value from
mongodb to insert into before insert it would not allow it.

serkandurusoy:
@projectn00b how so? I use matb33:collection-hooks in usecases where I lookup extra information from other collections.

aldeed:
@stubailo I think they both get the same data by pointing to the same store. Because `registerStore` will be called every time but exits pretty much immediately after the first time.

It's a good sign that existing tests pass, but maybe we need to think of some additional tests to add, checking that pub/sub works properly for multiple instances of the same collection. allow/deny should work fine since all client instances will call the same mutation method, but probably should have tests to prove that, too.

aldeed:
Regarding hooks, my opinion is that they are useful in enough valid cases to be warranted. Making sure they aren't misused is an argument for good documentation of them, not entirely omitting them.

However, this PR, and in general supporting subclassing and custom mutation methods, opens the door to future packages that can provide hooks without it needing to be part of the core pkg.

projectn00b:
@SerkanDurusoy a simple find or findone usually yields an error when using
within the hook itself.

serkandurusoy:
@projectn00b I assure you they work. If you would like to share your usecase, I'd like to offer you help. But let's do that either on SO or the matb33 issue tracker.

dropfen:
I cannot understand why the test failed. circleci is not giving useful information so far.. 

I came to this topic because of another problem: When you try to create a new AccountsClient with a separate DDP connection, the AccountsCommon constructor tries to create a new Collection with the name "users".. see #6070

zol:
Hey @aldeed . I'm happy to help get this merged if you still require the functionality  but as per the comment history it looks like at least more investigation needs to happen regarding the behavior it introduces on the client (ideally, more tests). I'm keen to get this PR either closed or merged.

We've recently fixed up the CI should if you want to pick up working on it again, please rebase the PR against the latest `devel` branch.

aldeed:
@zol I rebased and rewrote this as a completely backward compatible opt-in change since there are some questions about the effects.

- New `suppressSameNameError` option can be set to `true` to see only a warning about collections with the same name already being defined
- New `defineMutationMethods` option can be set to `false` to skip creating the allow/deny mutation methods.

zol:
Hey, It looks like the tests are failing in our CI system. Do they pass locally?

Just one change (apart from tests passing) I'd like before merging this is to rename `suppressSameNameError` to `_ suppressSameNameError` and remove it from the docs. We can surface this option later down the track if it proves to be reliable.

aldeed:
@zol I changed to `_suppressSameNameError`. All tests pass locally but one test failed on CI, "drop collection", which does not appear to be related to anything I did, but I'm not sure.

abernix:
@aldeed That test is failing on a regular basis – it's probably (read: almost positively certainly) not your change. It worked fine on Circle but failed on Travis:

CircleCI:
```
S: tinytest - mongo-livedata - oplog - drop collection/db : OK
```

I just restarted your Travis test.  Will see what happens.

abernix:
🏁✅🎆

zol:
Merged. Thanks guys!

foobarbecue:
Are there any contradindications / dangers associated with using  `_suppressSameNameError`?

aldeed:
@foobarbecue I have followed the code paths a bit and I don't personally see any problems with it, but the goal is to have people try it out and see if it causes any problems.

foobarbecue:
Cool, I've been using it on a [project](https://github.com/foobarbecue/lifestream/) and haven't encountered any problems so far. Thanks!

Issue #5800, opened 2015-12-14
Localmarket example app: Geolocation not EJSON stringifiable


matthisk:
In the localmarket example app a Geolocation is send to a (server side) method in share-overlay.js line 53. However the Geolocation returned from `Geolocation.currentLocation` can not be serialized as EJSON, this results in a server side error when calling `createActivity` in activities.js.

timbotnik:
Thanks for the bug report - we'll check into this.

matthisk:
I also created a pull request that fixes this issue #5801, it is not perfect but it works. Didn't test it out yet on all platforms though.

hwillson:
Closing as part of the old issue cleanup process. If needed, please consider re-opening this issue in the [Local Market](https://github.com/meteor/localmarket) repo. Thanks!

PR #5801, opened 2015-12-14
Fix issue 5800: send object instead of Geolocation to createActivity

matthisk:
The localmarket example app does not seem to be working correctly in browser when creating a new activity. This pr should fix the issue described in #5800


tmeasday:

PR #5834, opened 2015-12-19
Fixed incorrect property name in documentation

sheeldotme:
As per the method summary, and independent verification, `name` should be used rather than `method` to identify the method or subscription targeted for rate limiting.

```- `name`: The name of the method or subscription being called```


stubailo:
Merged: a28aa7f95c395bddc4d7401a55b90333b2ad35a1

Thanks!

PR #5841, opened 2015-12-20
Fix CSS URL rewriting for applications deployed under a sub path

### LINKED ISSUE #5837: Incorrect URL rewriting in CSS files if the application is deployed under a sub-path
I provided a description of the issue under the commit that introduced it, https://github.com/meteor/meteor/commit/1e4838ccd38c2df142591a67d675ac38eb8a5630#commitcomment-15072716.

mquandalle:
I provided a description of the issue under the commit that introduced it, https://github.com/meteor/meteor/commit/1e4838ccd38c2df142591a67d675ac38eb8a5630#commitcomment-15072716.

### DIFF
--- a/packages/minifier-css/minifier.js
+++ b/packages/minifier-css/minifier.js
@@ -81,7 +81,8 @@ CssTools = {
   // For performance reasons this function acts by side effect by modifying the
   // given AST without doing a deep copy.
   rewriteCssUrls: function (ast) {
-    rewriteRules(ast.stylesheet.rules);
+    var mergedCssPath = "/";
+    rewriteRules(ast.stylesheet.rules, mergedCssPath);
   }
 };
 
@@ -94,7 +95,7 @@ if (typeof Profile !== 'undefined') {
          });
 }
 
-var rewriteRules = function (rules) {
+var rewriteRules = function (rules, mergedCssPath) {
   _.each(rules, function(rule, ruleIndex) {
 
     // Recurse if there are sub-rules. An example:
@@ -102,10 +103,10 @@ var rewriteRules = function (rules) {
     //         .rule { url(...); }
     //     }
     if (_.has(rule, 'rules')) {
-      rewriteRules(rule.rules);
+      rewriteRules(rule.rules, mergedCssPath);
     }
 
-    var basePath = pathDirname(rule.position.source);
+    var basePath = pathJoin("/", pathDirname(rule.position.source));
 
     // Set the correct basePath based on how the linked asset will be served.
     // XXX This is wrong. We are coupling the information about how files will
@@ -119,7 +120,7 @@ var rewriteRules = function (rules) {
         basePath = "/";
 
     _.each(rule.declarations, function(declaration, declarationIndex) {
-      var parts, resource, absolutePath, quotes, oldCssUrl, newCssUrl;
+      var parts, resource, absolutePath, relativeToMergedCss, quote, oldCssUrl, newCssUrl;
       var value = declaration.value;
 
       // Match css values containing some functional calls to `url(URI)` where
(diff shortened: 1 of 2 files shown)


packages/minifiers-css/urlrewriting-tests.js
```diff
@@ -32,12 +32,12 @@ Tinytest.add("minifiers - url rewriting when merging", function (test) {
   t('http://', 'http://', 'malformed URL');
 
   parseOptions.source = 'application/client/dir/other-style.css';
-  t('./image.png', '/image.png', 'base path is root');
-  t('./child/image.png', '/child/image.png', 'child directory from root');
-  t('child/image.png', '/child/image.png', 'child directory from root');
-  t('/image.png', '/image.png', 'absolute url');
-  t('"/image.png"', '"/image.png"', 'double quoted url');
-  t("'/image.png'", "'/image.png'", 'single quoted url');
+  t('./image.png', 'image.png', 'base path is root');
```
Slava: can you please explain again on this concrete example why `/image.png` is incorrect and how `image.png` is correct if the ROOT_URL is set to `http://site.com/subpath/`?

mquandalle: Sure, it’s actually pretty simple. Let’s launch a Meteor app with:

    ROOT_URL="http://localhost:3000/subpath" meteor

and let's assume that the served CSS file contains something like

```css
background: url('/image.png');
```

The browser will resolve this resource by GETing the file `http://localhost:3000/image.png` whereas the file is actually served at `http://localhost:3000/subpath/image.png` resulting in a Not Found error. The problem is that absolute paths are relative to the domain name root, not necessarily the application root (in case the app is served under a sub-path).

I will provide a reproduction to make the issue even clearer.

mquandalle:
Here is a minimal reproduction of the issue: https://github.com/mquandalle/meteor-css-url-rewriting-subpath-issue

packages/minifiers-css/minifiers.js
```diff
@@ -121,17 +122,39 @@ var rewriteRules = function (rules) {
...
+        // in which case the resources linked in the merged CSS file would miss
+        // the `myapp/` prefix. Since this path prefix is only known at launch
+        // time (rather than build time) we can't use absolute paths to link
+        // resources in the generated CSS.
+        //
+        // Instead we transform absolute paths to make them relative to the
+        // merged CSS, leaving to the browser the responsibility to calculate
+        // the final resource links (by adding the application deployment
+        // prefix, here `myapp/`, if applicable).
+        relativeToMergedCss = path.relative(mergedCssPath, absolutePath);
```
Slava: is this using the node version of `path`? Can you check that it will work on Windows? For safety, I recommend to change it to `files.pathRelative(...`

packages/minifiers-css/minifiers.js
```diff
@@ -110,7 +111,7 @@ var rewriteRules = function (rules) {
         basePath = "/";
 
     _.each(rule.declarations, function(declaration, declarationIndex) {
-      var parts, resource, absolutePath, quotes, oldCssUrl, newCssUrl;
+      var parts, resource, absolutePath, relativeToMergedCss, quote, oldCssUrl, newCssUrl;
```
Slava: might be worth breaking down into two lines to 1) logically group the variables and 2) make the lines max 80 chars long

mquandalle: What about using ES6 `let` and avoid this line of `var` explicit hosting?

packages/minifiers-css/minifiers.js
mquandalle: Ok, let’s use `files.pathRelative`.

mquandalle: Is this function actually exposed to packages?

timbotnik:
Thanks for your help @Slava & thanks for the PR @mquandalle.

mquandalle:
Any plan to review the change @timbotnik?

mquandalle:
Anyone?

tmeasday:
@mquandalle this makes sense to me, sorry for the delays.

Can we rebase this off current devel (or merge devel into it) so we can get passing tests?

mquandalle:
Thanks for the positive response @tmeasday. I was a bit afraid that this issue would be closed when I saw the notification on GitHub :)

I’ll rebase on devel.

packages/minifier-css/minifier.js
```diff
@@ -130,17 +131,39 @@ var rewriteRules = function (rules) {
...
+        // in which case the resources linked in the merged CSS file would miss
+        // the `myapp/` prefix. Since this path prefix is only known at launch
+        // time (rather than build time) we can't use absolute paths to link
+        // resources in the generated CSS.
+        //
+        // Instead we transform absolute paths to make them relative to the
+        // merged CSS, leaving to the browser the responsibility to calculate
+        // the final resource links (by adding the application deployment
+        // prefix, here `myapp/`, if applicable).
+        relativeToMergedCss = path.relative(mergedCssPath, absolutePath);
```
tmeasday: We'll still need to resolve this on windows.

One option is to use `path.posix.relative` in node 4 if we just merge this to the 1.4 branch.

tmeasday:
@mquandalle any idea about that failing self test? It seems relevant.

You can run it with `meteor self-test "standard-minifiers - CSS splitting"`

tmeasday:
@mquandalle will you have a chance to look at this?

tmeasday:
Ok, the self test failing seemed to just be bad luck. Passing now.

Still need to resolve the `path.relative` thing. I might merge this into `release-1.4` next week.

mquandalle:
Shouldn't it be enough to require the patched version of the `path` lib, like so:

```js
const path = Plugin.path;
```

?

tmeasday:
I'll admit I'm confused. What's `Plugin`?

mquandalle:
http://docs.meteor.com/api/packagejs.html#build-plugin-file-system

tmeasday:
Hmm, we can't use the `Plugin` API in this file as it's not a build plugin. Probably the easiest thing to do is to follow the lead of what's already in this file: https://github.com/meteor/meteor/blob/master/packages/minifier-css/minifier.js#L159-L179

I'll try patching this PR

tmeasday:
Ok, closed in favor of https://github.com/meteor/meteor/pull/7357. If the tests pass I will merge at last

PR #5845, opened 2015-12-20
Allow to find the corresponding collection from the store

mitar:
I needed this so many times I lost the count.


mitar:
Used in https://github.com/peerlibrary/meteor-subscription-data.

tmeasday:
+1 (to this or some other way to get an existing instance of a collection w/ a given name)

stubailo:
Seems legit!

mitar:
Superb!

Issue #5269, opened 2015-09-24
Rebuild time grew 50% after updating to 1.2


avalanche1:
**_1 Upvote**_ It now loads compileLessBatch, compile-ecmascript, CosmosBrowserify, builds for web.cordova.
![meteor-load](https://cloud.githubusercontent.com/assets/10534563/10066980/214f821a-62a9-11e5-918b-2f513971c75f.gif)
Refresh time grew accordingly.
I'm on win7, btw.

Slava:
Things to do if you want to improve your report:

- post what packages you are using and what versions of the packages?
- post how many files of each extension are you using in your project? how big are the files?
- post what files are you changing that triggers the rebuild? server-side file? client-side file? CSS? template? etc?
- CosmosBrowserify is usually very slow. What npm packages are you trying to load with browserify?
- post output with the `METEOR_PROFILE=50` env variable set.

avalanche1:
Here you go:
```
accounts-password              1.1.3  Password support for accounts
audit-argument-checks          1.0.4  Try to detect inadequate input sanitiza...
bigdsk:inputmask               3.1.63  JQuery InputMask for Meteor
blaze-html-templates           1.0.1  Compile HTML templates into reactive UI...
check                          1.0.6  Check whether a value matches a pattern
coffeescript                   1.0.9  Javascript dialect with fewer braces an...
dburles:lazy                   1.0.0  A handy little development tool
ejson                          1.0.7  Extended and Extensible JSON library
fastclick                      1.0.7  Faster touch events on mobile
jquery                         1.11.4  Manipulate the DOM using CSS selectors
kadira:debug                   2.2.3  Full Stack Debugging Solution for Meteor
logging                        1.0.8  Logging facility.
manuel:viewmodel               1.9.12  MVVM, two-way data binding, and compon...
manuel:viewmodel-explorer      1.0.5  Side panel for manuel:viewmodel
meteor-base                    1.0.1  Packages that every Meteor app needs
meteorhacks:fast-render        2.10.0  Render your app before the DDP connect...
meteorhacks:subs-manager       1.6.2  Subscriptions Manager for Meteor
momentjs:moment                2.10.6  Moment.js (official): parse, validate,...
mongo                          1.1.1  Adaptor for using MongoDB and Minimongo...
mquandalle:bower               1.5.2  Use Bower packages in your Meteor app
mquandalle:jade                0.4.3_1  Jade template language
[...]

PR #5859, opened 2015-12-21
Add profiling information to clearly outline slow legacy-style compilers

Slava:
Helps to profile apps, like wekan.

https://github.com/meteor/meteor/issues/5269

### DIFF
--- a/tools/isobuild/compiler.js
+++ b/tools/isobuild/compiler.js
@@ -323,7 +323,11 @@ var lintUnibuild = function ({isopack, isopackCache, sourceArch}) {
 // options.isopack.
 //
 // Returns a list of source files that were used in the compilation.
-var compileUnibuild = function (options) {
+var compileUnibuild = Profile(
+  function (options) {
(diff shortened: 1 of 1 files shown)


mquandalle:
:+1: I removed most of legacy compilers in Wekan in c69f993 but the app still relies on some (like those exported by `tap-i18n`) and they are not easy to find in the PROFILE logs.

Slava:
@mquandalle they won't be in the profiler logs until somebody merges this :( @stubailo

stubailo:
Re-running the tests - they seemed to have some weird behavior. If they are in some reasonable state will merge!

Slava:
Yeah, I don't know. The change in this PR doesn't affect any functional behavior, only profiling reporting (which doesn't have selftests)

stubailo:
OK merged, hopefully the tests are wrong... as they often are. Why is it so hard to write reliable tests? Such an existential question.

mquandalle:
I don’t know, but the [Circle CI logs](https://circleci.com/gh/meteor/meteor/447) are super artistic :p

Circle CI

PR #5893, opened 2015-12-27
Expose makeAttributeHandler

mitar:
Exposing `makeAttributeHandler`. This is aligned with the comment:

> XXX make it possible for users to register attribute handlers!

This is necessary for Blaze Components for: https://github.com/peerlibrary/meteor-blaze-components/issues/99

@stubailo, It would be great if this could be merged in for Meteor 1.3. :-) Then I can release new version when Meteor 1.3 is released.

### DIFF
--- a/packages/blaze/attrs.js
+++ b/packages/blaze/attrs.js
@@ -65,7 +65,7 @@ AttributeHandler.extend = function (options) {
 //
 // Extended below to support classes, SVG elements and styles.
 
-var DiffingAttributeHandler = AttributeHandler.extend({
+Blaze._DiffingAttributeHandler = AttributeHandler.extend({
   update: function (element, oldValue, value) {
     if (!this.getCurrentValue || !this.setValue || !this.parseValue)
       throw new Error("Missing methods in subclass of 'DiffingAttributeHandler'");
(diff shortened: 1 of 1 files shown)


stubailo:
Hmm I think this is weird since it allows you to replace a base class that is extended by lots of other stuff.

Why not `Blaze.registerAttributeHandler`, and then you would register handlers for `onClick`, `onMouseOver`, etc. etc.? Plus, you could probably override the built-in ones for `class`, etc in this way.

It would only be a few more lines to add this logic rather than just exposing the object.

mitar:
No, the main thing I need is to be able to wrap the `Blaze._makeAttributeHandler`. This is the main feature. I am exposing `Blaze._DiffingAttributeHandler` so that one can create new class by extending it. The same as `Blaze._AttributeHandler` is already exposed.

You cannot really override `Blaze._AttributeHandler` or `Blaze._DiffingAttributeHandler` because at that point other classes are already defined. So there is no way to really interfere with something.

So the `Blaze._makeAttributeHandler` is what I use to hook up new handler which knows how to attach event handlers. You can see [here](https://github.com/peerlibrary/meteor-blaze-components/blob/master/client.coffee#L57) how I am using it.

mitar:
@zol: So now I can just merge this? Or what would be the process?

zol:
@mitar - Yes please move this across to meteor/blaze and go ahead and merge it. I'll close this out on meteor/meteor

Issue #3867, opened 2015-03-05
onLogout callback on the server side


mitar:
***1 Upvote*** It would be great to have an onLogout callback on the server side. This would allow one to cleanup and react to login changes. Useful for SSO and special things associated with logged in state.

Probably a client side part would be useful as well, but you can get that in a way with the reactive `Meteor.userId()`.

peterchoo:
This.  It would be incredibly useful.

I have a package which manages some per session data internally.  If the user refreshes, this data is no longer accessible, as it is restricted by sessionId.  But if the user logs out (in this case triggered from another package) I have no way of clearing my data.

sbrych:
+1

mitar:
The current workaround we are using is creating an empty publish on the server side to which client subscribes. And then that empty publish is keeping the local collection populated with information:

Server side:

```
sessions = new Mongo.Collection null

Meteor.onConnection (connection) ->
  id = share.sessions.insert
    _id: connection.id
    userId: null
  connection.onClose ->
    share.sessions.remove id

Meteor.publish 'sessionDummy', ->
  share.sessions.update @connection.id, $set:
    userId: @userId
  # Don't actually publish anything.
  []
```

Client:

```
Meteor.subscribe 'sessionDummy'
```

stubailo:
Is this use case not covered by the user status package? https://github.com/mizzao/meteor-user-status

mitar:
Yes, it has to use the same ugly trick with dummy empty subscription as I described above: https://github.com/mizzao/meteor-user-status/blob/master/status.coffee#L208

peterchoo:
@stubailo Thanks for pointing that package out.  I'll look into using it for the time being.

I strongly feel that the accounts package should provide an onLogout hook system to mirror the onLogin hooks.  It would be extremely useful if the hooks had available to them the connection which triggered it, and the user object.

Thank you!

stubailo:
Yep, sounds like a great idea. In general we don't have enough callbacks for all kinds of things.

aunz:
@mitar 
connection.onClose  only works if a user closes the browser/
If the user only clicks the logout button, the onclose callback will not be called?

mitar:
The logout you get through dummy subscription.

jagill:
Would be very useful.

lorensr:
[...]

PR #5894, opened 2015-12-27
Issue #3867 Add server side onLogout callback

erencay:
In order to track logged in users it is handy to have a logout callback on the server. Otherwise you have to apply some unhealthy hacks like mentioned in this issue #3867. Since I am new to both Meteor and Node.js I pretty much followed the other server side callback implementations.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -125,6 +125,11 @@ AccountsServer = class AccountsServer extends AccountsCommon {
   /// LOGOUT HOOK
   ///
 
+  /**
+   * @summary Register a callback to be called after a logged in user logs out.
+   * @locus Server
+   * @param {Function} func The callback to be called after a logged in user logs out. Takes the logged out user as parameter.
+   */
   onLogout(func) {
     this._onLogoutHook = func;
   }
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -120,6 +120,14 @@ AccountsServer = class AccountsServer extends AccountsCommon {
 
     this._onCreateUserHook = func;
   }
+
+  ///
+  /// LOGOUT HOOK
+  ///
+
+  onLogout(func) {
+    this._onLogoutHook = func;
+  }
 };
 
 var Ap = AccountsServer.prototype;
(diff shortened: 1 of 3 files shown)


packages/accounts-base/accounts_server.js
```diff
@@ -525,11 +533,15 @@ Ap._initServerMethods = function () {
   };
 
   methods.logout = function () {
+    var user = accounts.users.findOne(this.userId)
```
stubailo: Can we only fetch the user if there is a logout hook? Basically put this inside the `if`

erencay: I see your concerns. But to my understanding, we have to track the `userId` somehow, since we will lost the value after `userId` set to null, right? We either have to duplicate the `if` statement at the beginning and avoid unnecessary database hits  or we have to assign the `userId` to a local variable and use it in the `if` statement. *(Assuming we should not call the hook before `userId` actually sets to null.)*

stubailo: Yes you can just use a local variable.

stubailo:
Any progress on this? Still waiting on that one fix.

erencay:
Thanks for the reminder, I totally forgot about that pull request. Please let me know if something missing.

skirunman:
We could use this onLogout hook so I hope this gets implemented.

Should this PR also not include some updates to documentation ala [Accounts.onLogin](http://docs.meteor.com/#/full/accounts_onlogin)?

packages/accounts-base/accounts_server.js
```diff
@@ -525,11 +538,17 @@ Ap._initServerMethods = function () {
...
 
   methods.logout = function () {
+    var userId = this.userId;
     var token = accounts._getLoginToken(this.connection.id);
     accounts._setLoginToken(this.userId, this.connection, null);
     if (token && this.userId)
       accounts.destroyToken(this.userId, token);
     this.setUserId(null);
+
+    if(accounts._onLogoutHook) {
```
tmeasday: Please add a space between `if` and `(`

packages/accounts-base/accounts_server.js
```diff
@@ -525,11 +538,17 @@ Ap._initServerMethods = function () {
   };
 
   methods.logout = function () {
+    var userId = this.userId;
```
tmeasday: Why not just use `this.userId` inside the `if`?

packages/accounts-password/password_tests.js
```diff
@@ -590,6 +593,24 @@ if (Meteor.isClient) (function () {
     logoutStep
   ]);
 
+  testAsyncMulti("passwords - onLogout hook", [
+    function (test, expect) {
+      this.username = Random.id();
+      this.password = "password";
+      Accounts.createUser(
+        {username: this.username, password: this.password,
+        profile: { testOnLogoutHook: true }},
```
tmeasday: Can we format this as
```
{
  username: ...,
  password: ...,
}
```

tmeasday:
@erencay I made a few more comments, then I'll be happy to merge it

tmeasday:
Actually, #6889 seems like a more mature version of this. Thank you @erencay but I'll focus on that PR.

Issue #5552, opened 2015-10-27
A way to extend Spacebars compiler


mitar:
Currently it is not really possible to extend Spacbars compiler because the package loads before other packages on the server and you cannot really monkey-patch it or something. And once you get to your package, it seems all templates are already compiled? It would be great if there would be a way to inject yourself somehow there, or add stuff.

I am looking into this because of https://github.com/peerlibrary/meteor-blaze-components/issues/99. Because I would like to support such syntax:

```
<template name="ButtonComponent">
  <button class="red" onClick="onClick">Click me</button>
</template>
```

Where `onClick` attribute would not be rendered as attribute, but would be removed from HTMLJS and added to an event map for that DOM element once it is rendered.

(The above code would call `onClick` on the component instance for the `click` event.)

stubailo:
Injecting code into build plugins is not currently possible from outside of the build plugin, and I don't think the architecture of build plugins really allows that at all. The correct thing to do here would be to create your own version of the `templating` plugin and ask people to switch to that one. In fact, the templating package is very nicely factored to allow you to easily do just that: https://github.com/meteor/meteor/blob/devel/packages/templating/plugin/compile-templates.js

However, as a personal preference, I would prefer to have an extra attribute in the DOM over having to switch build plugins.

mitar:
If Blaze Components include my `templating` plugin and some other package includes official `templating` plugin, what happens? Which one will be able to process `.html` files?

mitar:
Could package's `imply` help here?

stubailo:
Build plugins are actually per-package. So different packages can easily use different HTML build plugins :]

You can't use both of them in the _same_ app or package though. (Think of an app like a big package that has automatic file loading, it's basically like that internally)

mitar:
OK, so if I have package-based apps, things should work. Great.

mitar:
This is not completely smooth because one has to copy most of the `templating` package around. I split it into two packages to make this easier: https://github.com/meteor/meteor/pull/5903

mitar:
See also: https://github.com/meteor/meteor/issues/5911

mitar:
See also: https://github.com/meteor/meteor/issues/5911

PR #5903, opened 2015-12-29
Support easier creation of server side Blaze rendering

mitar:
This splits `templating` package into two packages, while keeping backwards compatibility.

This allows one to develop an alternative `templating` package which can for example render templates differently (see https://github.com/meteor/meteor/issues/5552 and https://github.com/peerlibrary/meteor-blaze-components/issues/99), and to provide server side compilation of templates with an alternative HTML compilation plugin. But one can still reuse code for `Template`.

Ideally, dynamic template would also be part of the base package, but it seems there is no way to add the HTML file and say "later on some other package will provide a plugin how to compile the HTML file, or to ignore it".


stubailo:
OK, I think this is a good change overall (in general I think decoupling the runtime and build plugin is good as well).

I think we could could make everything work if we did this:

1. `templating-build-plugin` has just the HTML compiler
2. `templating` has just the dynamic and templating.js stuff, and depends on `templating-build-plugin`

Note that this will actually allow you to use only the compiler, or only the runtime, because the dependency on the build plugin inside templating is restricted to that package only.

mitar:
> Note that this will actually allow you to use only the compiler, or only the runtime, because the dependency on the build plugin inside templating is restricted to that package only.

Hm, but will then this work? Because currently one makes a dependency on the `templating` (or `blaze-html-templates`) package when you want to process `*.html` in your app/package. If what you are saying is true, then dependency on the `templating` will not do this anymore? Does this mean that we also have to change that `blaze-html-templates` implies `templating-build-plugin`? But then `templating` will not have to depend on the `templating-build-plugin`?

Anyway, one other thing I would like to achieve is to be able to use all the content of the `templating` runtime also on the server side. So my build plugin works on the server side (in fact, I just removed the `archMatching` restriction and [things work](https://github.com/peerlibrary/meteor-blaze-components/blob/master/compile-templates.js)). But the issue is that by default `dynamic.html` cannot be processed on the server side. Any ideas how to make this work? One option would be that `dynamic.html` would be added as an asset which you can then compile on the server side?

Or should we remove the `archMatchin` [restriction](https://github.com/meteor/meteor/blob/devel/packages/templating/plugin/compile-templates.js#L3) and all client-only stuff in templating package and this is it? So then it is to the developers to decide to put `*.html` files only to the client side, or if they want, they can put it on the server side as well.

mitar:
Hm, when trying to publish a package with:

```
Package.registerBuildPlugin({
  name: "compileBlazeComponentsTemplatesBatch",
  use: [
    'caching-html-compiler',
    'ecmascript',
    'templating-tools',
    'spacebars-compiler',
    'html-tools'
  ],
  sources: [
    'patch-compiling.js',
    'compile-templates.js'
  ]
});
```

I am getting:

```
error: You must specify a version constraint for package caching-html-compiler
error: You must specify a version constraint for package ecmascript
error: You must specify a version constraint for package templating-tools
error: You must specify a version constraint for package spacebars-compiler
error: You must specify a version constraint for package html-tools
```

Is this normal?

stubailo:
Yes I don't know if versionsfrom exists for build plugins.

stubailo:
Actually have you tried putting a versionsfrom in your package.onUse?

mitar:
Yes, [I have it](https://github.com/peerlibrary/meteor-blaze-components/blob/master/package.js) but it is not respected. Should I open a ticket about this?

mitar:
See also: https://github.com/meteor/meteor/issues/5911

mitar:
See [also this for some issues with current approach](https://github.com/orionjs/orion/pull/381).

It seems that this support for multiple handlers for the same `*html` file is probably problematic. Probably most people want to imply `templating` because they want to expose `Template` automatically. But they do not really care how it is compiled or which package is compiling.

stubailo:
OK I was wrong at first, we'd need three packages: just the runtime, just the build plugin, and then both.

And I'd say that people implying templating is just a bug in the package - there is no reason a package with useful runtime stuff should be implying a build plugin unless it's explicitly trying to be some kind of meta package.

Anyway, what changes would you need to core to export the right template stuff on the server?

mitar:
> Anyway, what changes would you need to core to export the right template stuff on the server?

I would just remove all stuff which is limiting `templating` symbols and files to the client only. I manually copied files to Blaze Components, included them, and they work. The only tricky thing is #5913.

So I would suggest that my pull request is modified to create those three packages. And that all files and symbols are made to work both on the client and server. The only changes I had to do was:
* https://github.com/peerlibrary/meteor-blaze-components/blob/master/server.coffee
* https://github.com/peerlibrary/meteor-blaze-components/blob/master/patch-compiling.js#L87

So the problem for `<head>` is that it rendered by server, but it is added to the client to do so.

mitar:
So, what should we do about this?

stubailo:
Not sure. @zol ?

zol:
This is a reasonable product change however I'd like someone technical to sign off on the PR, with an eye towards not breaking anything rather than whether it's the perfect solution. @stubailo are you in a position where you can give me the green light given you've already looked at this or should I find someone in the Meteor team?

stubailo:
Yeah this looks fine to me. But looks like it should be in the Blaze repository now, right?

zol:
@mitar - Could you please move this over to meteor/blaze and go ahead and merge it. I'd love to test out our process around collaborating through the split repo and iron out any wrinkles. I'm going to close it out on meteor/meteor

mitar:
OK, but I am not completely satisfied with the design here. It would be really good to get some input here what should be the design of the solution.

mitar:
Also, when will [blaze be removed from this repository](https://github.com/meteor/meteor/tree/devel/packages/blaze)? I am afraid this will lead to issues soon.

PR #5906, opened 2015-12-30
Fixed small typo in function scope explanation

brunocassol:
`var x = function x () {}`

should instead read:

`var x = function () {}`


stubailo:
I actually think both are fine, and the one in the code sample is better since it's not anonymous, I think.

stubailo:
Either way, the whole section is wrong since `var x = function ...` is fundamentally different from a function declaration since function declarations are hoisted.

PR #5915, opened 2016-01-01
Use scopes to disambiguate between Meteor Packages and NPM Modules

### LINKED ISSUE #5909: [1.3-modules-beta.2] Importing twilio fails because underscore isn't imported correctly
Installing `twilio` under `node_modules` doesn't work - where-as `Npm.require("twilio")` works within a package.

Example app at `https://github.com/nathan-muir/meteor-1.3-npm-issue` just `npm install && meteor`.

Error thrown:

```
TypeError: Object [object Object] has no method 'extend'
    at meteorInstall.app.node_modules.twilio.lib.AccessToken.js (app/node_modules/twilio/lib/AccessToken.js:21:1)
    at fileEvaluate (packages/modules/.npm/package/node_modules/install/install.js:183:1)
    at require (packages/modules/.npm/package/node_modules/install/install.js:75:1)
    at meteorInstall.app.node_modules.twilio.lib.index.js (app/node_modules/twilio/lib/index.js:31:1)
    at fileEvaluate (packages/modules/.npm/package/node_modules/install/install.js:183:1)
    at require (packages/modules/.npm/package/node_modules/install/install.js:75:1)
    at meteorInstall.app.server.test.js (/Users/nathan/Projects/meteor-npm-issue/.meteor/local/build/programs/server/app/app.js:41:15)
    at fileEvaluate (packages/modules/.npm/package/node_modules/install/install.js:183:1)
    at require (packages/modules/.npm/package/node_modules/install/install.js:75:1)
    at /Users/nathan/Projects/meteor-npm-issue/.meteor/local/build/programs/server/app/app.js:39549:1
```

Peeking at `AccessToken.js` shows the following:

```
var _ = require('underscore'); // 1
/* snip */
_.extend(IpMessagingGrant.prototype, { // 21
  key: 'ip_messaging',
```

So, it appears the issue is `require('underscore')` isn't importing the right version / one with the correct exports.

nathan-muir:
Fixes #5909 

Currently, if any code attempts to `import` a NPM module which collides with a Meteor Package name - the module system would import the Meteor package. This causes NPM modules that `require("underscore")`, or `require("babel-runtime")` to fail as they receive the Meteor versions of these modules.

After taking a look at a few solutions, (eg, prefixing with special characters), I stumbled upon [NPM Scopes](https://docs.npmjs.com/misc/scope)

This PR changes `Isobuild.ImportScanner` to only recognise Meteor Packages when scoped (currently `meteor`).

So, now if you `require("underscore")` you'll get the NPM module.

But if you'd like the Meteor version, you simply `require("@meteor/underscore")`.

This change applies to all imports of meteor packages - even if there is no NPM name conflict - all imports must be prefixed with `@meteor/`

### DIFF
--- a/packages/accounts-base/accounts_url_tests.js
+++ b/packages/accounts-base/accounts_url_tests.js
@@ -1,4 +1,4 @@
-import {AccountsTest} from "accounts-base";
+import {AccountsTest} from "@meteor/accounts-base";
 
 Tinytest.add("accounts - parse urls for accounts-password",
   function (test) {
@@ -9,7 +9,7 @@ Tinytest.add("accounts - parse urls for accounts-password",
 
     _.each(actions, function (hashPart) {
       var fakeToken = "asdf";
-      
+
       var hashTokenOnly = "#/" + hashPart + "/" + fakeToken;
       AccountsTest.attemptToMatchHash(hashTokenOnly, function (token, action) {
         test.equal(token, fakeToken);
--- a/tools/isobuild/import-scanner.js
+++ b/tools/isobuild/import-scanner.js
@@ -35,6 +35,7 @@ export default class ImportScanner {
     sourceRoot,
     usedPackageNames = {},
     nodeModulesPath,
+    packageModuleScope = "meteor",
   }) {
     assert.ok(isString(sourceRoot));
 
@@ -43,6 +44,7 @@ export default class ImportScanner {
     this.sourceRoot = sourceRoot;
     this.usedPackageNames = usedPackageNames;
     this.nodeModulesPath = nodeModulesPath;
+    this.packageModuleScope = packageModuleScope;
     this.absPathToOutputIndex = {};
     this.outputFiles = [];
   }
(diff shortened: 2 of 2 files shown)


mquandalle:
Related: a6a5373

nathan-muir:
@benjamn Would be good to get your feedback on this solution instead of #5933 / a6a5373

eg. With only a6a5373, I can easily break my app that already uses meteor's underscore by just `npm install some-package-that-uses-underscore`

benjamn:
@nathan-muir you raise a valid point, and I absolutely agree that the right/better answer is to namespace Meteor packages somehow, so that they can't collide with npm packages.

I believe we can accomplish that namespacing without imagining that Meteor packages have a special npm scope, by simply reorganizing the directory structure of installed modules:
```
meteorInstall({...
├── example.css
├── example.html
├── example.js
└── node_modules
    ├── meteor
    │   ├── accounts-base.js
    │   └── blaze.js
    ├── moment
    │   ├── index.js
    │   └── package.json
    └── react
        └── package.json
```
In other words, if Meteor packages appear to be installed in `node_modules/meteor/`, then packages like `underscore` can be imported like so: `import {_} from "meteor/underscore"` or `var _ = require("meteor/underscore")._`. Of course, what you probably want to do is to `npm install underscore` locally and then `import _ from "underscore"`, which will work correctly because `underscore` refers to the npm package an not the Meteor package.

benjamn:
0a6324013ada0f32f86b09465ccaa74c6081900c implements the `meteor/...` prefixing I described above, which should fix the same problem without introducing the concept of scopes.

I definitely want to support npm scopes in the future, and there may even be a `@meteor/...` or `@mdg/...` scope in that future world. In the meantime, I would prefer not to give the impression that Meteor supports scopes, when we only kind of support one special scope (`@meteor`). I don't think full support will be that hard to implement, but it will require (at the very least) an upgrade of the `npm` command installed in the Meteor dev bundle.

Thanks for taking the time to put together this pull request, and especially for pointing out how easy it is to break an app by installing an npm package using npm 3.

PR #5916, opened 2016-01-01
Refactor reset password and enroll account

bwobst:
Was working with `Accounts.sendResetPasswordEmail` and `Accounts.sendEnrollmentEmail` and noticed a comment that the two could be refactored.  I went ahead and abstracted the 99% of the identical code from both and wrapped it in a 'sendEmail' method.  The only real difference between the two (which were a good 50 lines of code each) was the key `resetPassword` and `enrollAccount`.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -517,54 +517,7 @@ Meteor.methods({forgotPassword: function (options) {
  * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
  */
 Accounts.sendResetPasswordEmail = function (userId, email) {
-  // Make sure the user exists, and email is one of their addresses.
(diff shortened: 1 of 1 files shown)


mquandalle:
Why are you using a Meteor method and not a simple JavaScript function. By writing a method you allow clients to call `sendEmail` with the arguments they want, which could lead to some user spamming.

bwobst:
Good point.  A regular function would be good, but I thought wrapping it as
a method was best practice.

stubailo:
Nope, methods should only be used for code that is intended to be called from the client. Everything else should be a plain function.

stubailo:
@bwobst please refactor so that I can accept the PR!

bwobst:
No problem, plain function it is.

This PR is being replaced by PR#5958. This one I just edited in the browser.  The other is properly linked up to a local fork so I can push additional changes if needed.

Issue #5911, opened 2015-12-31
Improve error message for conflicting file handlers


mitar:
This error message:

> conflict: two packages included in the app (templating and peerlibrary:blaze-components) are both trying to handle *.html

Does not really help you a lot because many packages `imply` `templating`, so it is unclear which package exactly you have to remove. See [this list here](https://github.com/peerlibrary/meteor-blaze-components/issues/111#issuecomment-168109852). `templating` is not there, but it is still been complained about.

mitar:
The issue is that some packages are implying other packages. See [here](https://github.com/orionjs/orion/pull/381).

It seems that this support for multiple handlers for the same `*html` file is probably problematic.

stubailo:
Yikes. I'd say that for almost all packages, implying templating is a bug. I don't know how we could specify a better error message there, without rewriting a lot of constraint solver logic or something.

Perhaps we could have a way to determine which package is loading templating for you, by displaying some sort of tree visualization of dependencies.

mitar:
> Yikes. I'd say that for almost all packages, implying templating is a bug.

I would not agree. Even Blaze Components were implying `templating` in the past. So that you could just add Blaze Components to your package and then use it and add HTML files to your package. Requiring one to add one extra sibling dependency is a bit strange. The idea should be that the package can define all dependencies you need to use it properly. Currently you have to imply `templating` if you want to make a package which uses HTML templates.

stubailo:
What if the person using your package wants to use a different HTML handler? Like in this exact case? I'd say a package that uses React implying JSX would be an anti-pattern.

mitar:
No, I would say that probably there should be three packages, a package which implies HTML handler, and another which does not, and the handler itself.

mitar:
> What if the person using your package wants to use a different HTML handler?

In the case of Blaze Components that is not supported. Or that HTML handler would have to additionally extend Blaze Components HTML handler.

stubailo:
I see, I guess in that case the only option would be to remove Orion or whatever is causing the conflict.

mitar:
Of course, the main issue is that it is not really possible to extend `templating` package. In fact, the only thing I do, is add [an extra file](https://github.com/peerlibrary/meteor-blaze-components/blob/master/package.js#L20) (and make it run on the server side).

chompomonim:
> I see, I guess in that case the only option would be to remove Orion or whatever is causing the conflict.

[...]

Issue #5618, opened 2015-11-05
Allow for multiple compilers of same file type.


AlexFrazer:
Use case: I am using `ecmascript` plugin, but also want to create a documentation generator which will similarly process javascript files. Something like this would be cool:

```
PluginHandlers.ecmascript.extend({
  processForPackage(file, options) {
  }
});
```

rozzzly:
I mean you could always fork `ecmascript` (and the actual compiler package it depends on, `babel-compiler`). Then modify the compiler's code that handles the per-file compilation to run your docgen before/after its babelified.

rozzzly:
Although the compiler code is not something you typically would care about being documented given that es5 code is pretty ugly.

stubailo:
Yeah, I think this is a legitimate feature we should have. I also am not sure if this is a great application of that feature, though, since I feel like you could achieve the documentation functionality with an external tool. Are you planning on somehow serving these docs as part of your app UI?

egoldblum:
I too would like the ability to register additional build compilers for a single file type.  In my case, I would like to statically rewrite certain paths in `.html` files at build time; particularly in `head.html`.

As to the "achieving same functionality with an external tool": this feels a bit strange to me given the fact that linters & minifiers are first class citizens in the build process.  I can easily achieve the same functionality a linter plugin gives me using an external tool (in fact, I do), so it seems somewhat arbitrary to make that the deciding factor of the validity of this feature.

stubailo:
> so it seems somewhat arbitrary to make that the deciding factor of the validity of this feature.

I'd argue this makes the linter feature of Meteor less useful than some other feature that would be hard to achieve using an external tool.

sungwoncho:
I am writing a helper for flow router that can be used in React components. My solution was to register a compiler for `.jsx` extensions to look for the helper strings and evaluate them. But this cannot be done because no two packages can handle the same file extensions. I think this case is a good application of this proposed feature.

stubailo:
> I am writing a helper for flow router that can be used in React components.

Why would a helper for flow router require parsing source code?

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

[...]

PR #5936, opened 2016-01-05
Allow access to build plugins

mitar:
See #5911 and #5552 for discussion.

cc @stubailo

### DIFF
--- a/tools/isobuild/isopack.js
+++ b/tools/isobuild/isopack.js
@@ -661,6 +661,30 @@ _.extend(Isopack.prototype, {
         sourceProcessorSet.addSourceProcessor(sp);
       },
 
+      _getByExtension: function (sourceProcessorSet, extension) {
+        return sourceProcessorSet.getByExtension(extension);
+      },
+
+      _getByFilename: function (sourceProcessorSet, filename) {
+        return sourceProcessorSet.getByFilename(filename);
+      },
+
+      _getCompilerByExtension: function (extension) {
+        return Plugin._getByExtension(isopack.sourceProcessors.compiler, extension);
+      },
+
+      _getMinifierByExtension: function (extension) {
+        return Plugin._getByExtension(isopack.sourceProcessors.minifier, extension);
+      },
+
+      _getCompilerByFilename: function (filename) {
+        return Plugin._getByFilename(isopack.sourceProcessors.compiler, filename);
+      },
+
+      _getMinifierByFilename: function (filename) {
+        return Plugin._getByFilename(isopack.sourceProcessors.minifier, filename);
+      },
+
       // Compilers are part of the Batch Plugins API.
       //
       // A compiler plugin is provided by packages to participate in


stubailo:
@avital - I think @mitar wants this to be able to show a better error message for conflicting build plugins, what do you think?

@mitar - I think it would be good to have a more descriptive message for the commit, so that when people look at the history for this file they can know why these methods are there. Also, would it maybe be better to have one PR for the whole thing with error messages rather than just these methods?

avital:
I don't understand this pull request, and definitely don't understand the motivation behind it. @mitar, can you explain?

mitar:
We can discuss it here, but I thought it is better to discuss it in relevant tickets (#5911 and #5552). I just made this pull request so that we can discuss easier there.

Anyway, the point of this pull request is that currently you have issues when two plugins are trying to handle the same file extension. This change allows for a plugin to detect that some other plugin is trying to handle its extension and adapt the behavior. So before you blindly register a plugin, you can check if plugin already exists. Maybe then you do not register anything, maybe then you replace that plugin. Or patch it, or chain it, or whatever.

mitar:
> I think it would be good to have a more descriptive message for the commit, so that when people look at the history for this file they can know why these methods are there.

Sure. I can cleanup the pull request after we decide it is OK.

> Also, would it maybe be better to have one PR for the whole thing with error messages rather than just these methods?

The point is that error messages become a non-problem once you have these methods available.

Slava:
> Anyway, the point of this pull request is that currently you have issues when two plugins are trying to handle the same file extension. This change allows for a plugin to detect that some other plugin is trying to handle its extension and adapt the behavior. So before you blindly register a plugin, you can check if plugin already exists. Maybe then you do not register anything, maybe then you replace that plugin. Or patch it, or chain it, or whatever.

How is allowing an auto-magical behavior like a good idea? From the app developer prospective, a random build plugin might be registered by some deep-deep dependency. And because that deep dependency is present, another build plugin decides to flip its behavior to something else?

mitar:
That is already completely like that. Whole Meteor allows monkey-patching. I can monkey-patch Blaze and make it behave completely differently. Why would now this be an exception?

Of course, developer should make sure that things are not too unexpected. That behavior is only extended and not changed.

Do you have a better proposal how to easily allow this? I think it is the best compromise between time spend developing it, and then seeing how much demand is really for this feature. It is exactly how the rest of Meteor is done by exposing various `_internals` around which you can then monkey-patch.

So please, let's be practical here and not block innovation just because somebody can break something. This can be done on many other levels already.

benjamn:

mitar:
Can we just reopen?  There is little benefit in loosing all this conversation in the new pull request?

benjamn:
Absolutely. Sorry the PR reaping process is kinda blunt :)

mitar:
I think you  could achieve daily pull request closing quota by merging some of my other easier to merge and pretty straightforward pull requests. ;-)

zol:
I'm going to close this PR for two reasons:

(1) In light of the fact that a transition to npm is on the medium term roadmap, we don't want to increase the surface area of the current build plugin APIs.

(2) There is considerable trepidation from the core team in exposing the build plugin system's internals in this way, including legitimate concerns around promoting a questionable usage pattern.

mitar:
@stubailo, @tmeasday: I would like to bid again for reopening this pull request. Accessing other build plugins from plugins would allow us to implement `<script language="coffee"></script>` tag in Blaze. So one could during compilation of Blaze template, access build plugin for `coffee` extension and get it to build the file.

But it seems that compiles to not really have an API which allows string to string compilation? Or we could just use `File` object with `data` and pass it to `compileOneFile`.

tmeasday:
Seems reasonable to me. What do you think @benjamn?

mitar:
Ping? @tmeasday, @benjamn 

I am not completely sure if the above would work though, because tool chain plugin is also doing some postprocessing. But maybe with some options to plugins like "bare" it can serve as a simple compiler from strings to strings, when a different plugin needs that. But I think we can add such options later to plugins, if it will be needed.

tmeasday:
Well I will reopen it anyway.

mitar:
Can this get merged please? We will also need it for https://github.com/meteor/blaze/issues/110.

benjamn:
I wanted to work out a better way to provide this functionality before closing this PR, but I haven't had a chance to work on it.

Specifically, it would be great to have a system for chaining together multiple compiler plugins to process source files in a series of steps.

Tricky aspects of that design problem include (but are not limited to): deciding how compiler plugins should specify which source/intermediate files they care about, and making sure plugins run in the right order with respect to one another.

Babel presets and plugins (if you're familiar with those) are a good example of the complexity of configuring a compiler pipeline with multiple processing steps.

It may seem like exposing source processors to package code would help solve your problem, but it creates a huge new problem for Meteor the framework, because it makes a previously internal API public, which keeps us from changing how compiler plugins work in the future, which would ironically limit our ability to provide this exact functionality in a better way.

mitar:
This is why methods are prefixed with `_`. :-) So that we can start building this things now, see what is needed in the future, and that can inform better future design process here.

But in short, for me at the moment the interesting thing is to be able to say: please compile this string and return. No special things here like matching files and stuff you mentioned above. Just so that inside `<script>` tag in Blaze I can embed CoffeeScript, or Stylus inside `<style>` tag. And reuse existing plugins for that.

abernix:
Housekeeping: Just tagging as related to #5618

PR #5937, opened 2016-01-05
Allow middleware to set http status code (fixes #5820)

### LINKED ISSUE #5820: Allow middleware to set the HTTP status code
@stubailo requested someone create an issue on this repo back in April but I can't find one here. Apologies if this is a duplicate.

The problem is discussed at https://github.com/iron-meteor/iron-router/issues/1055 but basically boils down to changing https://github.com/meteor/meteor/blob/devel/packages/webapp/webapp_server.js#L711 from:
```javascript
res.writeHead(200, headers);
res.write(boilerplate);
res.end();
return undefined;
```
to:
```javascript
var statusCode = res.statusCode ? res.statusCode : 200;
res.writeHead(statusCode, headers);
res.write(boilerplate);
res.end();
return undefined;
```

This would allow the router to return a 404 (or any non 200) status code (so Google etc. know not to index the url) but still render the boilerplate for the app so we can display the notFound template configured in IronRouter.

mjmasn:
This change will allow routers and other middleware to set the HTTP status code to something other than 200, for example returning a 404 if the route does not exist on the client or server. It retains the default 200 response if `res.statusCode` has not been set.

My first PR here, couldn't see an obvious way to add tests for this functionality as it's inside an anonymous function but happy to be enlightened.

In the case of Iron Router, there is already some commented-out code that this will allow to work as long as routes are defined on client & server (https://github.com/iron-meteor/iron-router/blob/devel/lib/router_server.js#L80-L89). There may be other SEO considerations (for the JS-rendering search engines) - if I click a client-side route from a 404 (where Iron Router NotFound template is rendered), perhaps the page should be refreshed to give Googlebot etc. the correct status code? Maybe Google already thought of that... I don't know.

Fixes #5820

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -708,7 +708,8 @@ var runWebAppServer = function () {
       return undefined;
     }
 
-    res.writeHead(200, headers);
+    var statusCode = res.statusCode ? res.statusCode : 200;
+    res.writeHead(statusCode, headers);
     res.write(boilerplate);
     res.end();
     return undefined;


stubailo:
Looks great!

solace:
Apologies, but how long does it usually take for a merged pull to be released? Or are there staged milestones?

mjmasn:
@solace I would assume it would be in the next release but I don't know for sure. Luckily Meteor is 'just' a collection of modules so if you need this functionality now, you can always copy the webapp directory from meteor/packages (devel branch) into your app's packages folder. Just remember to remove it when you upgrade meteor as there may be version constraint issues.

PR #5939, opened 2016-01-05
Added option to extend ddp for custom messages

okland:
Added option to extend ddp for custom messages.

By either giving option in connection creation `options.customMessages`.
Or dynamically by extending Meteor.connection or other Connection
```
_.extend(Meteor.connection._customMessages, {
    updateBatch: function() {}, 
    subscriptionRedirect: function(){}
});
```

I needed this for couple of use cases :
1. Handling sending batch updates from server to client.
2. Do subscription redirect - send the client to get subscription from another server.

### DIFF
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -76,6 +76,9 @@ var Connection = function (url, options) {
   self._heartbeatInterval = options.heartbeatInterval;
   self._heartbeatTimeout = options.heartbeatTimeout;
 
+  // Option to extend ddp messages with custom messages 
+  self._customMessages = options.customMessages || {};
+  
   // Tracks methods which the user has tried to call but which have not yet
   // called their user callback (ie, they are waiting on their result or for all
   // of their writes to be written to the local cache). Map from method ID to
@@ -257,6 +260,8 @@ var Connection = function (url, options) {
     else if (msg.msg === 'result')
       self._livedata_result(msg);
     else if (msg.msg === 'error')
+      self._livedata_error(msg);
+    else if (_.include(_.keys(self._customMessages), msg.msg))
       self._livedata_error(msg);
     else
       Meteor._debug("discarding unknown livedata message type", msg);


stubailo:
I doubt we would add new messages to DDP right now, since that would break a lot of existing DDP clients. There would need to be a big reason and a lot of discussion to do so.

Perhaps you could simulate those custom messages via custom `added` messages, in a way that is specific to your application?

okland:
@stubailo Currently i feel that moving message like 'subscriptionRedirect' ( Needed in case that somebody subscribe to subscription that is already handled by another server, to reuse the live query, the multiplexer and the logic on constructing the query )  into the added feels a bit weird, Moreover i think that this kind of extension doesn't break DDP clients just let anybody who wants to expend their functionality do so. I think that this feature will open a window for a lot of innovation in the community, Will allow to solve much of the most burning  issues regarding scalability.

stubailo:
It could be helpful, you're right. Either way, it shouldn't come in the form of a pull request. We should first have a discussion on the forum about this and a more detailed design document, so that there is a chance to decide if these are the correct messages to be added.

DDP is not the part of the system where we can afford to accept a PR without careful consideration and discussion.

okland:
@stubailo  i would like to re-open this PR for discussion, 

I'm attaching a medium post in which i explain in details why we should extend the ddp protocol with custom messages (extension that will not break existing clients)
[Should we allow custom ddp messages ?](https://medium.com/@omriklinger/should-we-allow-custom-ddp-messages-a5fff26673e2#.10rpqof5y)

This solution is a required tool for solving scale problems of production apps (already solving mine). For example @avital solution  https://github.com/meteor/meteor/commit/23b2d1b517ebe0b09989eb44c2f5defe72023164, the problem of the specified production app, could have been optimized in many ways using this solution. For example sending the data in a batch, stringify messages once per subscription update and sent to all listening clients. Choosing to integrate this into the core will let developers share their solutions with the community, get feedback and make those solutions available to every production app. Instead of the current state where we cannot share those solutions because they are based on changes in core packages.

I think this solution make even more sense when looking forward into using reactive graphQL, for sending cached data ( All at once)

BTW in the original PR there is an error, and the code not updating to the latest version because the PR is close, The latest version of the code is https://github.com/okland/meteor/commit/703af71ebb16ff4e04639b08e8a255c1bda15ae7

stubailo:
I can't reopen the PR for some reason, maybe you can do it on your end?

okland:
I re-opend it as new one https://github.com/meteor/meteor/pull/6117

PR #5951, opened 2016-01-07
Webapp caching improvements

martijnwalraven:
This is a more comprehensive version of https://github.com/meteor/meteor/pull/5900.

As discussed on that PR, using `location.replace()` instead of `location.reload()` can significantly improve reloading performance because it avoids unnecessary cache validations.

One problem noted there is that we currently serve non-cacheable assets with a `max-age` of a day, which means that changes to assets (such as an image in `/public`) would not be detected if we switch to `location.replace()`. 

Setting the `max-age` to this value was supposedly done to avoid image flickering during reloads (see #773). I'm not sure about the original reasoning, but thinking about it some more I don't see how this would help. With `location.reload()`, the browser basically ignores `max-age` anyway and it therefore always sends at least a conditional request. That still leads to flickering, as far as I can see (only tested on Safari and Chrome). 

So I believe serving assets with a `max-age` of 0 and switching to `location.replace()` should not actually make a difference in this respect (but it does give us the performance improvements we want for assets that *are* cacheable).

Another point was that the presence of a `Last-Modified` header based on the file date would interfere with caching, because browsers would make requests conditional on the date as well as the ETag. 

In order to disable the generation of a `Last-Modified` header, I needed to upgrade the `send` and `connect` dependencies. (I've upgraded `connect` to the latest 2.x version because I did not want to risk a major version upgrade).

I've also smuggled in a small addition I needed for Cordova hot code push downloads, which is to make sure we don't inadvertently download the default index page when a file is not found. The code for this is in the last commit.

### DIFF
--- a/packages/reload/reload.js
+++ b/packages/reload/reload.js
@@ -218,8 +218,10 @@ Reload._reload = function (options) {
 
   var tryReload = function () { _.defer(function () {
     if (Reload._migrate(tryReload, options)) {
-      // Tell the browser to shut down this VM and make a new one
-      window.location.reload();
+      // Make the browser reload the page
+      // Using location.replace() instead of location.reload() avoids
+      // validating assets with the server if we still have a fresh cached copy.
+      window.location.replace(window.location.href);
     }
   }); };
 
--- a/packages/webapp/.npm/package/npm-shrinkwrap.json
+++ b/packages/webapp/.npm/package/npm-shrinkwrap.json
@@ -1,65 +1,387 @@
 {
   "dependencies": {
     "connect": {
-      "version": "2.9.0",
+      "version": "2.30.2",
       "dependencies": {
-        "qs": {
-          "version": "0.6.5"
+        "basic-auth-connect": {
+          "version": "1.0.0"
         },
-        "cookie-signature": {
-          "version": "1.0.1"
-        },
-        "buffer-crc32": {
-          "version": "0.2.1"
-        },
-        "cookie": {
-          "version": "0.1.0"
-        },
-        "bytes": {
-          "version": "0.2.0"
-        },
-        "fresh": {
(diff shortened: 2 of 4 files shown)


martijnwalraven:
I did some more tests to see if these changes affect image flickering on reload. As far as I can see, there is no difference between the previous and the new caching approach. Safari and Chrome on Mac do show flickering in both cases, Firefox and Internet Explorer do not. Curiously, neither does Safari on iOS.

Browser | Image flickering
--- | ---
Chrome 47 (Mac) | yes
Safari 9.0.2 | yes
Firefox 43 (Mac) | no
Chrome 47 (Android) | yes
Safari (iOS 9.2) | no
IE 11 on Windows 7 | no
IE 10 on Windows 7 | no

avital:
LGTM!

(And to make sure I understand -- eliminating the 1 day caching on static assets means that now updating an image file in `public/` will correctly lead to the browser showing the new image?)

martijnwalraven:
Yep, that's the idea. 

About the image flickering, I'm not actually sure the results above are valid. I don't think the browser itself makes the difference, but whether the image has been validated when the page is rendered. And that might depend on a whole host of other factors (including the network conditions, etc.). But at least I am still convinced these changes don't make a difference, so that should be ok.

avital:
Merge it!

tmeasday:
Thanks for working on this in such details @martijnwalraven

mjgallag:
This fix is awesome. I just noticed this issue in 1.2.1. I watched a hot code reload where CSS wasn't modified (only js) and thus hash file name was identical. But due to reload and Last-Modified/Etag having been changed not only did it check with server, but it downloaded the whole file again because it didn't get a 304 hit. It appears this will solve that issue, thanks a lot @martijnwalraven!

PR #5959, opened 2016-01-08
Refactored sendPasswordResetEmail and sendEnrollmentEmail

bwobst:
Noticed a comment on `sendEnrollmentEmail` that said `sendPasswordResetEmail` and `sendEnrollmentEmail` could/should be refactored, so I took a stab at it and added a `sendEmail` function that removes duplicate code.


bwobst:
By the way, I forked the project, but it's pulled in other commits that I did not make.

PR #5981, opened 2016-01-13
Expose options for performance tuning `observe` on the server

n1mmy:
These can be useful to tune performance for application specific data patterns.

Parameters exposed and documented:
- disableOplog
- pollingInterval
- pollingThrottle

Tested with:
```
m = new Mongo.Collection("foo");
if (Meteor.isServer) {
  Meteor.publish("foo", function () { return m.find({_id: '7'}, {disableOplog: true}); });
  Facts.setUserIdFilter(function (userId) { return true; });
} else if (Meteor.isClient) {
  Meteor.subscribe("foo");
}
```
in a new app with `facts` and not `autopublish`

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -280,6 +280,9 @@ _.extend(Mongo.Collection.prototype, {
    * @param {MongoFieldSpecifier} options.fields Dictionary of fields to return or exclude.
    * @param {Boolean} options.reactive (Client only) Default `true`; pass `false` to disable reactivity
    * @param {Function} options.transform Overrides `transform` on the  [`Collection`](#collections) for this cursor.  Pass `null` to disable transformation.
+   * @param {Boolean} options.disableOplog (Server only) Pass true to disable oplog-tailing on this query. This affects the way server processes calls to `observe` on this query. Disabling the oplog can be useful when working with data that updates in large batches.
+   * @param {Number} options.pollingInterval (Server only) How often to poll this query when observing on the server. In milliseconds. Defaults to 10 seconds.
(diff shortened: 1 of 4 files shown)


packages/mongo/collection.js
```diff
@@ -280,6 +280,9 @@ _.extend(Mongo.Collection.prototype, {
    * @param {MongoFieldSpecifier} options.fields Dictionary of fields to return or exclude.
    * @param {Boolean} options.reactive (Client only) Default `true`; pass `false` to disable reactivity
    * @param {Function} options.transform Overrides `transform` on the  [`Collection`](#collections) for this cursor.  Pass `null` to disable transformation.
+   * @param {Boolean} options.disableOplog (Server only) Pass true to disable oplog-tailing on this query. This affects the way server processes calls to `observe` on this query. Disabling the oplog can be useful when working with data that updates in large batches.
+   * @param {Number} options.pollingInterval (Server only) How often to poll this query when observing on the server. In milliseconds. Defaults to 10 seconds.
```
avital: Can this be renamed to `pollingIntervalMs`? It was confusing, especially when it says, "the default is 10 seconds".

packages/mongo/collection.js
```diff
@@ -280,6 +280,9 @@ _.extend(Mongo.Collection.prototype, {
...
+   * @param {Number} options.pollingThrottle (Server only) Minimum time to allow between re-polling. Increasing this will save CPU and mongo load at the expense of slower updates to users. Decreasing this is not recommended. In milliseconds. Defaults to 50 milliseconds.
```
avital: Ditto `pollingThrottleMs`?

avital:
Sorry for the late review!

n1mmy:
Options updated and re-tested.

avital:
Awesome

wearhere:
I'm curious about `pollingThrottle` —what would cause the driver to poll sooner than `pollingInterval`?

PR #6005, opened 2016-01-14
Add meteor login / logout info to docs

nagelflorian:
I recently came across the issue of how to handle Galaxy deployments with third-party continiuous integration providers. Unfortunately I couldn't find any infos in the documentation on how to handle `meteor login` without sharing my main credentials. Luckily I came across [this post](https://medium.com/@natestrauser/migrating-meteor-apps-from-modulus-to-galaxy-with-continuous-deployment-from-codeship-aed2044cabd9#.j4a68a16v) and the usage of `METEOR_SESSION_FILE` to generate a session token. To promote good security practices we should add this to the docs.

### DIFF
--- a/docs/client/full-api/commandline.html
+++ b/docs/client/full-api/commandline.html
@@ -70,6 +70,16 @@ You can use the `--package` option, to create a new package. If used in an
 existing app, this command will create a package in the packages
 directory.
 
+<h3 id="meteorloginlogout">meteor login / logout</h3>
+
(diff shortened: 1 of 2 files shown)


stubailo:
Ah yes! We were going to put this either in the galaxy docs or the Meteor Guide but didnt get around to it.

This is a good start, but we should put it in the deployment section of the guide, opened an issue for that.

PR #6008, opened 2016-01-15
Allow literal functions for Minimongo sorting

### LINKED ISSUE #2730: Allow specifying a comparator function for sorting a cursor
I couldn't find any existing tickets requesting this - my apologies if I missed it.

When a sort is requested using the `sort` option of `Collection#find` on the client side, it looks to me as if that sorting is done entirely on the client side. The mongo sort spec is converted into a comparator by a `Minimongo.Sorter` object which is simply passed to the built-in sort function. I would like the ability to simply specify the comparator directly. Currently, I can do this:

```
var cursor = MyCollection.find(query);
cursor.sorter = {
  getComparator: function() {
    return function(a, b) {
      // return 0, 1, or -1
    }
  }
};
```

This is essentially mimicking the interface of a `Minimongo.Sorter` object. It seems to work, but it involves mucking with the internals of cursors a little more than I'm comfortable with. `Minimongo.Sorter` contains a lot of code which I'm not duplicating and whose purpose I'm not sure of, so this might be missing something important.

Would it be possible to get a better and more official API for doing this? The following seems like it would be ideal:

```
MyCollection.find(query, {
  sort: function(a, b) {
    // return 0, 1, or -1
  }
});
```

Thanks.

mitar:
Fixes #2730.

### DIFF
--- a/docs/client/full-api/api/collections.md
+++ b/docs/client/full-api/api/collections.md
@@ -803,6 +803,10 @@ The last form will only work if your JavaScript implementation
 preserves the order of keys in objects. Most do, most of the time, but
 it's up to you to be sure.
 
+For local collections you can pass a comparator function which receives two
+document objects, and returns -1 if the first document comes first in order,
+1 if the second document comes first, or 0 if neither document comes before
+the other. This is a Minimongo extension to MongoDB.
 
 
 {{> apiBoxTitle name="Field Specifiers" id="fieldspecifiers"}}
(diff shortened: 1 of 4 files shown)


stubailo:
Running tests locally, then going to rebase and merge.

stubailo:
Merged! abe2c5e01121ef7734373ae215c95a44bf5f944e

PR #6042, opened 2016-01-19
Better polyfill for Error.captureStackTrace in Meteor.makeErrorType.

benjamn:
The trick of returning an object other than `this` from the constructor function makes the classes created by `Meteor.makeErrorType` impossible to subclass, unless the subclass is very careful to return the same object from its own constructor.

Fortunately, the only reason we needed an actual `Error` instance in this code was to get access to its `.stack` property, which we can just as easily borrow from an auxiliary `Error` object.

cc @stubailo @dgreensp

### DIFF
--- a/packages/meteor/errors.js
+++ b/packages/meteor/errors.js
@@ -4,26 +4,19 @@
 //
 Meteor.makeErrorType = function (name, constructor) {
   var errorClass = function (/*arguments*/) {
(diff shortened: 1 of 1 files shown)


benjamn:
This was merged as f0a84785dbc07c8fe3bb5bbc02e541ba206e7d96.

Issue #6052, opened 2016-01-21
1.3b4 PIXI import error ('fs' missing)


jiku:
***1 Upvote*** Trying to `import PIXI from "pixi.js"` I'm getting

```
Error: Cannot find module 'fs'
    at require (http://localhost:3000/packages/modules-runtime.js?hash=bfd7d4eb668102fe6d61ecbda4bab9035cd5af94:105:21)
    at meteorInstall.node_modules.pixi.js.src.core.renderers.webgl.filters.SpriteMaskFilter.js (http://localhost:3000/packages/modules.js?hash=b6b40b4043dfd34996d56e2169a4f735335af624:12260:10)
```

I've made a PR (hope I'm doing it right) #6053.
As noted, it seems to be an issue with PIXI.

chiluap:
+1 same here

newsiberian:
@chiluap, hello, faced with the same error in i18next-node-fs-backend on [1.3-beta.8]. Did you find any workaround?

jiku:
`i18next-node-fs-backend` works here, with something like

```javascript
import * as i18nextBackend from 'i18next-node-fs-backend';
console.log(i18nextBackend);
```

on server only. Should it work on the client? See #5963.

With PIXI, it's a client-only (mostly?) lib AFAIK. Not sure why it depends on fs.

For a workaround to your issue, if it should work on the client I guess you could do something like

  "dependencies": {
    "fs": "*",
    "path": "*",
    "i18next-node-fs-backend": "*"
  }

where "fs" at least seems just be an empty placeholder module someone made. Tested here on client and it now traces

```shell
Object {type: "backend"}
default: Backend(services)
arguments: (...)
caller: (...)
length: 1
name: "Backend"
prototype: Backend
type: "backend"
__proto__: ()
<function scope>
type: "backend"
__proto__: Object
```

benjamn:
This is fixed in beta.16, thanks to #6056.

PR #6053, opened 2016-01-21
jiku/modules/pixi

jiku:
For issue #6052.

Maybe related issues with PIXI through Webpack

https://github.com/pixijs/pixi.js/issues/1854
https://github.com/pixijs/pixi.js/issues/2078

Where they set

node: {
    fs: "empty"
}


benjamn:
Thanks very much for submitting a pull request against the test app!

benjamn:
This is fixed in beta.16, thanks to #6056.

PR #6060, opened 2016-01-21
Allow HTTP OPTIONS method to correctly support CORS preflight

mitar:
Otherwise CORS preflight does not work for `manifest.json` file. CORS preflight can happen if you are having some extra headers in your request for `manifest.json` (so if the request is classified as [non-simple](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests)).


mitar:
:heart:

PR #6080, opened 2016-01-25
mongo url is good for an hour now

rdickert:
This functionality was changed by @multilinear. Change the cli help to match this.


multilinear:
I just checked the code, yup one hour

Issue #6070, opened 2016-01-24
Not possible to create multiple AccountsClients


dropfen:
if `Accounts` exists it's not possible to create a  `new AccountsClient` because under the hood `AccountsCommon` is trying to create `new Mongo.Collection('users'..)` and this is not allowed as mentioned by this PR-Issue #5778 

https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_common.js#L23

dropfen:
Could someone explain me, if this is not a bug?

tmeasday:
Hey @dropfen - The use case for a second `AccountsClient` is to connect to the accounts system on a *second* server -- thus you'd provide a different connection to the construction. Consequently, you wouldn't run into this issue of a duplicate collection -- you can have two collections with the same name on different connections.

Do you have a specific issue you were running into or were you just confused about how this could work?

dropfen:
Hey @tmeasday, thanks for the reply.
(I though, I did passed the connection, but I didn't passed it via options object, but as an argument)

However, a second ago, I tried it again and it worked, I could create another AccountsClient Object called **A2** with **M2** (which is a DDP.connection) as connection, but I would like to connect to the same application on another domain, to perform some of the tasks on another server.

As I understood I can do it, but I cannot login to **M2** with my **Accounts._storedLoginToken**.

I tried 
```
A2.loginWithToken(Accounts._storedLoginToken(), function(err, res){});
// => error: You've been logged out by the server. Please log in again.

```

Maybe you could give me a hint, how to share the same token between the connections?

tmeasday:
I don't think you can use the same login token on both connections as the token is generated by the server, and thus will be different on both servers.

dropfen:
Hmm.. this are not good news. I tried to perform a password login on two servers, it worked only on the one, where I initially created the user. This means it has something to do with the collection, which cannot be shared between connections, right?

tmeasday:
I mean you might just have to call `.createUser()` against both servers with the same username and password, right.

dropfen:
surprisingly **A2** which is an instance of AccountsClient, does not have `.createUser` method

dropfen:
Oh, actually it's working.. I had a typo in my mongodb declaration for the second server.. arggg..

Now it's working with `loginWithToken`method.. just fine ;)
(It has to, because the db is the same..)

A lot of thanks for helping!

PR #6083, opened 2016-01-26
Return existing 'users' collection, if it exists, otherwise a new one #BugFix

dropfen:
Related to #6070 

Summary: If you try to create a new AccountsClient after the regular Accounts Object was created, you will get en Error because the underlying AccountsCommon constructor will go to create a new "users" collection, right? 

So there is a discussion going on, how to return an existing collection from the Mongo.Collection constructor itself in #5778, but this quick fix, will make at least the AccountsClient work as it should.

### DIFF
--- a/packages/accounts-base/accounts_common.js
+++ b/packages/accounts-base/accounts_common.js
@@ -18,12 +18,8 @@ AccountsCommon = class AccountsCommon {
     this.connection = undefined;
     this._initConnection(options || {});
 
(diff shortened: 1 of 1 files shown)


benjamn:
Thanks for taking the time to identify this workaround. As you suggest, I believe #5778 will be a better way to fix this problem, so let's focus discussion on that issue, rather than fixing it in this one spot.

PR #6085, opened 2016-01-26
Clone serverWatchSet to not modify it

mitar:
Otherwise `settingsWatchSet` is added to `bundleResult.serverWatchSet` which is at the same also `cachedServerWatchSet`.

Also a fix for style.

### DIFF
--- a/tools/runners/run-app.js
+++ b/tools/runners/run-app.js
@@ -652,7 +652,7 @@ _.extend(AppRunner.prototype, {
       };
     }
 
-    var serverWatchSet = bundleResult.serverWatchSet;
+    var serverWatchSet = bundleResult.serverWatchSet.clone();
     serverWatchSet.merge(settingsWatchSet);
 
     // We only can refresh the client without restarting the server if the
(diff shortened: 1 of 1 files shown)


mitar:
Ping.

benjamn:
This seems to interfere with the assumptions of this code below
```js
if (! canRefreshClient) {
  // Restart server on client changes if we can't refresh the client.
  serverWatchSet = combinedWatchSetForBundleResult(bundleResult);
}
```
Specifically, if `! canRefreshClient`, we throw away the cloned `serverWatchSet` you created above, which effectively discards the `settingsWatchSet` files you added to it.

Does that make sense? Is that cause for concern?

mitar:
Maybe then `serverWatchSet.merge(settingsWatchSet);` should be called after `! canRefreshClient`.

So what worries me is that you modify something which I see as input (`serverWatchSet`) internally.

But probably it does not matter. Feel free to close this pull request.

stubailo:
> But probably it does not matter. Feel free to close this pull request.

Doing this.

PR #6101, opened 2016-01-27
Fixed manifest.json path on Windows

mitar:
This made URL paths with `\` which then did not correctly resolve. There is no need to be OS specific here, it breaks things.


mitar:
:heart:

PR #6125, opened 2016-02-01
Make #each use _id in object or as getter in object's prototype

### LINKED ISSUE #6124: Make #each use _id in object or as getter in object's prototype
When feeding #each an array of objects, if the objects have _id in them at the instance level than each uses the full range of observable related dom transformations which means that interaction with third party libraries doesn't break, because each keeps track of the actual dom nodes that move and when, say, sortable moves them, it respects that they are already moved.

However if _id is a getter up the objects prototype chain, each does not use that and falls back to index based "change" notification which can easily break in this situation.  Making #each more aggressively use _id would be a better solution in many cases I think.

in packages/observe-sequence/observe_sequence.js:
```
seqChangedToArray = function (lastSeqArray, array, callbacks) {
  var idsUsed = {};
  var seqArray = _.map(array, function (item, index) {
    var id;
    if (typeof item === 'string') {
      // ensure not empty, since other layers (eg DomRange) assume this as well
      id = "-" + item;
    } else if (typeof item === 'number' ||
               typeof item === 'boolean' ||
               item === undefined) {
      id = item;
    } else if (typeof item === 'object') {
      id = (item && _.has(item, '_id')) ? item._id : index;
    } else {
      throw new Error("{{#each}} doesn't support arrays with " +
                      "elements of type " + typeof item);
    }

    var idString = idStringify(id);
    if (idsUsed[idString]) {
      if (typeof item === 'object' && '_id' in item)
        warn("duplicate id " + id + " in", array);
      id = Random.id();
    } else {
      idsUsed[idString] = true;
    }

    return { _id: id, item: item };
  });

  return seqArray;
};
```
The line `id = (item && _.has(item, '_id')) ? item._id : index;`
could be changed to `id = (item && 'id' in item) ? item._id : index;`

In this case objects which hide the _id in a prototype chained getter (such as an ORM type implementation of denormalized documents, just as an example of my use case) would still work.

lassombra:
Fixes #6124 as described in that issue

### DIFF
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -292,7 +292,7 @@ seqChangedToArray = function (lastSeqArray, array, callbacks) {
                item === undefined) {
       id = item;
     } else if (typeof item === 'object') {
-      id = (item && '_id' in item) ? item._id : index;
+      id = (item && ('_id' in item)) ? item._id : index;
     } else {
       throw new Error("{{#each}} doesn't support arrays with " +
                       "elements of type " + typeof item);
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -292,7 +292,7 @@ seqChangedToArray = function (lastSeqArray, array, callbacks) {
                item === undefined) {
       id = item;
     } else if (typeof item === 'object') {
-      id = (item && '_id' in item)) ? item._id : index;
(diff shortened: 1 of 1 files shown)


lassombra:
Note that circleci is erroring on tests completely unrelated to this pull request.

avital:
LGTM, but can we add parenthesis around the `in` clause to make the order of operations explicit?

lassombra:
Sure no problem, I'll get that in right now

stubailo:
Merged, thanks!

Issue #6140, opened 2016-02-02
Weird behavior using check and objects with prototypes


sebakerckhof:
Check seems to only check for an object's own properties, e.g. this will run fine, without affecting the `child` object:
```js
var parent = {foo:"bar"};
var child = Object.assign(Object.create(parent),{bar:"foo"}); 
check(child, {bar: String,});
```

Let's say you have an object `child` as above and you run this code:
```js
check(child, {bar: String});
myCollection.insert(child);
```
This will insert a document with properties `foo` and `bar` in myCollection.

I know that in 99% of the cases `check` will be used on values coming from JSON (through DDP), but I'd like to argue that the package shouldn't make this assumption and a developer could reasonably assume/expect object `child` to only contain property `bar` after running through `check`. Therefore the current behavior, as described above, is weird and error-prone.

tmeasday:
I'd probably argue the real issue here is that minimongo is pickup up prototypical properties and inserting them into the DB. This seems the wrong behavior to me. I think it's a duplicate of https://github.com/meteor/meteor/issues/4260 really.

sebakerckhof:
I don't agree, because even if you leave out the minimongo part, the behavior of `check` is still weird to me:
```js
var parent = {foo:"bar"};
var child = Object.assign(Object.create(parent),{bar:"foo"}); 
check(child, {bar: String});
if(child.foo){ 
  console.log("this is not what I'd expect"); 
}
```

Since `check` is its own package and actually quite useful for a lot of stuff (I believe @stubailo mentioned it could be a good stand-alone NPM package for the larger JS community), I think it makes sense to make it work with objects that have prototypes.

+ It's a really small change, I can provide a PR, but I want to verify if there's a specific reason not to do it that way.

tmeasday:
@sebakerckhof We are really recommending people use simple-schema to do schema checks these days and if anything looking to deprecate check. Perhaps that was the package @stubailo was referring to?

I'd be hesitant to change the semantics of `check()` given it's such a security critical feature and we'd have to think through the implications. It does sound like what you are proposing would be an in improvement though. @stubailo what do you think?

stubailo:
@tmeasday I did suggest that if we plan on keeping around check, we should make it an npm package and encourage all kinds of people to use it. It looks like considering non-own properties would only make check _more_ restrictive, not less, so it doesn't look like it would be a security issue.

[...]

PR #6148, opened 2016-02-03
Make check check inherited properties

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/6140

### DIFF
--- a/History.md
+++ b/History.md
@@ -13,6 +13,8 @@
 * Move `DDPRateLimiter` to the server only, since it won't work if it is called from the client. It
   will now error if referenced from the client at all.
 
+* `check` will now check all keys of an object, including inherited ones #6140
+
 * Allow `git+` URL schemes for npm dependencies #844
 
 Patches contributed by GitHub users vereed, devgrok, ...
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -325,7 +325,16 @@ var testSubtree = function (value, pattern) {
       requiredPatterns[key] = subPattern;
   });
 
-  for (var keys = _.keys(value), i = 0, length = keys.length; i < length; i++) {
+  //XXX: replace with underscore's _.allKeys if Meteor updates underscore to 1.8+ (or lodash)
+  var allKeys = function(obj){
+    var keys = [];
+    if (_.isObject(obj)){
+      for (var key in obj) keys.push(key);
+    }
+    return keys;
+  }
+
+  for (var keys = allKeys(value), i = 0, length = keys.length; i < length; i++) {
     var key = keys[i];
     var subValue = value[key];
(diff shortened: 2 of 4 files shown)


stubailo:
Merged! Thanks!

https://github.com/meteor/meteor/commit/eb897bc91744946390bf38e56f80780a9786b801

mitar:
This goes directly against current documentation. Citing:

> Matches an Object with the given keys, with values matching the given patterns. If any pattern is a Match.Optional, that key does not need to exist in the object. The value may not contain any keys not listed in the pattern. **The value must be a plain Object with no special prototype.**

Also, the `Object` documentation says:

> Matches any **plain** Object with any keys;

Personally, I have been using this property of `check` to make sure nobody gives me any fancy object which has overridden not just property, but getters, setters, and other fancy stuff which might have side effects and so on.

I think this is pretty drastic backwards incompatible change and I would vote to revert it.

If people need this feature, make something like `Match.FullObject({key1: pattern1, key2: pattern2, ...})` and `Match.FullObjectIncluding({key1: pattern1, key2: pattern2, ...})`.

sebakerckhof:
> Personally, I have been using this property of check to make sure nobody gives me any fancy object which has overridden not just property, but getters, setters, and other fancy stuff which might have side effects and so on.

Well, then you might have security issues, because check just ignored the prototype before this PR.

mitar:
Yes. The fix should probably then be that it throws an error if it is not a plain object.

sebakerckhof:
But checking if an object is a plain object is quite tricky (You'd have to check all property descriptors, what if `Object` itself is modified, etc.) While on the other hand if you really want to ensure an object is a plain object, you can easily sanitize it before giving it to check (`JSON.parse(JSON.stringify(foo))`)

mitar:
> But checking if an object is a plain object is quite tricky

I have the following functions in my code often:

```
function isTrueObject(value) {
  return _.isObject(value) && !_.isArray(value) && !_.isFunction(value);
}

function isPlainObject(obj) {
  if (isTrueObject(obj)) {
    return false;
  }
  if (obj.constructor !== Object) {
    return false;
  }
  return true;
}
```

I assume `Object` is not changed because it is on the server.

> you can easily sanitize it before giving it to check

The whole point of calling `check` is to have a simple one command to verify arguments of methods. Not that I have first to sanitize and then call `check`.

> I do agree I should have modified the docs with this PR, because currently the docs are not correct.

Yes, by that you are introducing backwards incompatibility. It is not just changing docs.

stubailo:
> The fix should probably then be that it throws an error if it is not a plain object.

I agree, my mistake on accepting a PR that doesn't do this. Is this an easy check to make? @mitar should/can we just add the above code to check?

PR #6160, opened 2016-02-04
Missed filter to pass _getCollection through

mitar:
When making change into a pull request (#5845) I missed this file change.

This now provides a way to access collections from stores on the client.

### DIFF
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -442,7 +442,8 @@ _.extend(Connection.prototype, {
     // implemented by 'store' into a no-op.
     var store = {};
     _.each(['update', 'beginUpdate', 'endUpdate', 'saveOriginals',
-            'retrieveOriginals', 'getDoc'], function (method) {
+            'retrieveOriginals', 'getDoc',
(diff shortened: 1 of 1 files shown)


mitar:
cc @stubailo, please merge this one as well. ;-)

mitar:
Or, not missed, but by allowing this you can also do:

```
Meteor.connection._stores['CollectioName']._getCollection()
```

Which helps a lot with current issue that you cannot register the same collection multiple times.

dburles:
Hey, wondering why API isn't "public"? It could replace https://github.com/dburles/mongo-collection-instances

mitar:
So that packages like mongo-collection-instances can use it and expose it and maintain a stable API despite internal API changing through time.

dburles:
Oh cool, I had wondered if that was the idea!

msavin:
@mitar this is client side only right? is there a server-side equivalent?

mttstt:
ping

mitar:
This can be on the server, if you are making a DDP client connection from the server. What do you mean by "server side equivalent"? What would you like to get?

PR #6176, opened 2016-02-06
Emphasize the tree structure in the profile output

mquandalle:
I took inspiration from the Unix [`tree`](https://en.wikipedia.org/wiki/Tree_%28Unix%29) command to modify the profile output to emphasize the hierarchal relations between the tree items.

A gist containing both the old and the new proposed formating is available [here](https://gist.github.com/mquandalle/ae441310cdd466eddf8b), though GitHub formating isn’t as good as the one on a terminal because Github insert some vertical space in between the text lines.

cc @dgreensp


stubailo:
Seems like a good idea to me!

Issue #6108, opened 2016-01-28
Allowing calling Meteor methods with an option to not retry them on hot-reload


mitar:
Currently, if there is any pending Meteor method being run and the server reloads (DDP connection is re-established) those Meteor methods are retried. It would be great if one could disable that for particular method calls.

tmeasday:
Agreed, I think this would be a useful feature. I don't think we'll implement this in the current livedata implementation. @stubailo - would we take a PR that implemented it?

stubailo:
Yeah this is something I've actually wanted for a while - the ability to completely disable retries, and throw an error if we don't know that the method succeeded, or have reason to believe it failed.

Would super accept a PR for this, and would love to make it the default in ValidatedMethod. (whether or not you are OK with retrying would depend on whether the method is idempotent)

lassombra:
I'm interested in tackling this but have a question about how people would like to see this implemented.  As I see it, there are a couple of for sure details and a couple of options.

Defining a method as idempotent:

- Could be done in method definition.
- Could be an option passed to call

Handling idempotent with reconnect:
- Should throw an error.  
- Error should describe the situation
- Reconnect would ideally have a way to query the old subscription id and try to get the result of the method to handle dropped connection vs server reload.

Default setting:
- Should probably be methods are not idempotent (current behavior)
- Could be idempotent methods by default?

I am leaning towards defining in method definition, not at method call.  I am also leaning towards defaulting to non-idempotent for reliable backwards compatible behavior.

I've begun work on this and will submit a PR when done.

stubailo:
I think you have the opposite meaning of idempotent. An idempotent method is one that is okay to call multiple times.

Canceling retries has to be an option passed to Meteor.apply because the client doesn't have any way of accessing definitions of server-only methods. Then we can integrate that option into a method calling wrapper like ValidatedMethod.

The default should be what happens in current meteor for backward compatibility reasons.

tmeasday:
Oh I and I was going to add:

> Reconnect would ideally have a way to query the old subscription id and try to get the result of the method to handle dropped connection vs server reload.

Do you mean "old connection id"? This sounds like something that's out of scope; we can't guarantee the client will connect to the same web server and I don't believe there's any way to identify this connection as being the "same" as an old one (like what property would you reliably use?)

lassombra:
Yeah, I'm tired and was thinking backwards and inside out.  
[...]

PR #6180, opened 2016-02-07
Create noRetry option on Connection.apply

lassombra:
Implements #6108.

Test is included, as is updated documentation.
All tests in ddp-client package (including new test for #6108) pass.

Implemented as simply a `noRetry` flag on apply which triggers it to fail with a Meteor.Error(409) instead of retrying on reconnect.

### DIFF
--- a/docs/client/full-api/api/methods.md
+++ b/docs/client/full-api/api/methods.md
@@ -74,7 +74,7 @@ it will re-call the method when it reconnects. This means that a client may
 call a method multiple times when it only means to call it once. If this
 behavior is problematic for your method, consider attaching a unique ID
 to each method call on the client, and checking on the server whether a call
-with this ID has already been made.
+with this ID has already been made.  Alternatively, you can use [`Meteor.apply`](#meteor_apply) with the noRetry option set to true.
 
 {{> autoApiBox "DDPCommon.MethodInvocation#userId"}}
 
@@ -178,7 +178,10 @@ even if the method's writes are not available yet, you can specify an
 
 `Meteor.apply` is just like `Meteor.call`, except that the method arguments are
 passed as an array rather than directly as arguments, and you can specify
-options about how the client executes the method.
+options about how the client executes the method.  Options permitted are:
+- `wait` (Client only): If true, don't send this method until all previous method calls have completed, and don't send any subsequent method calls until this one is completed.
+- `onResultReceived` (Client only): This callback is invoked with the error or result of the method (just like `asyncCallback`) as soon as the error or result is available. The local cache may not yet reflect the writes performed by the method.
+- `noRetry` (Client only): if true, don't send this method again on reload, simply call the callback with an error (will be `Meteor.Error(409)`)
 
 <h2 id="ddpratelimiter"><span>DDPRateLimiter</span></h2>
 
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -274,11 +274,44 @@ var Connection = function (url, options) {
     msg.support = self._supportedDDPVersions;
     self._send(msg);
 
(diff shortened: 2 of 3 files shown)


packages/ddp-client/livedata_connection.js
```diff
@@ -274,11 +274,44 @@ var Connection = function (url, options) {
...
+            if (loopMethod.methodId === methodInvoker.methodId) {
+              break;
+            }
+          }
+          // Remove from current method block. This may leave the block empty, but we
+          // don't move on to the next block until the callback has been delivered, in
+          // _outstandingMethodFinished.
+          currentMethodBlock.splice(i, 1);
+          // make sure that the method is told that it failed.
+          methodInvoker.receiveResult(Meteor.Error(409, 'Method is non-idempotent but attempted to call a second time',
```
stubailo: I would change this to an error string - for example `"invocation-failed"`. Error numbers in DDP are deprecated and I'd like to get rid of them if possible. Also, including the word "idempotent" in the message might be confusing for developers who don't know what that means.

docs/client/full-api/api/methods.md
```diff
@@ -178,7 +178,10 @@ even if the method's writes are not available yet, you can specify an
 
 `Meteor.apply` is just like `Meteor.call`, except that the method arguments are
 passed as an array rather than directly as arguments, and you can specify
-options about how the client executes the method.
+options about how the client executes the method.  Options permitted are:
```
stubailo: We don't need to list these options here, right? They should be in the API box above.

packages/ddp-client/livedata_connection.js
```diff
@@ -703,6 +737,7 @@ _.extend(Connection.prototype, {
    * @param {Object} [options]
    * @param {Boolean} options.wait (Client only) If true, don't send this method until all previous method calls have completed, and don't send any subsequent method calls until this one is completed.
    * @param {Function} options.onResultReceived (Client only) This callback is invoked with the error or result of the method (just like `asyncCallback`) as soon as the error or result is available. The local cache may not yet reflect the writes performed by the method.
+   * @param (Boolean) options.noRetry (Client only) if true, don't send this method again on reload, simply call the callback with the error
```
stubailo: The documentation is generated from here, see how to update it in this hackpad: https://meteor.hackpad.com/Automatically-Generating-API-Docs-using-JSDoc-EpPmd2iuFEH

stubailo:
This is amazing! I especially like that I can tell that the code doesn't do anything different if the `noRetry` option isn't passed.

Going to merge this now, and fix the docs later.

stubailo:
Amazing contribution! Merged as one commit: c80c7414931dc1051dea93a918cc1ae2850d45c7

I made some changes, hopefully that is OK with you!

packages/ddp-client/livedata_connection.js
lassombra: Yeah, I was actually going to change that just as soon as I got home after re-reading the style guide today.  If 'invocation-failed' is used, the test should also be updated to expect that.

docs/client/full-api/api/methods.md
lassombra: Yeah, missed that the api box was there, glad it's not needed.

packages/ddp-client/livedata_connection.js
lassombra: And I see now that I somehow did parenthesis instead of braces, otherwise the documentation will update as needed.

lassombra:
Totally ok with the changes.  I made note of one where the test no longer reflects the new contract.  I could provide another pull request, but I suspect it'd be easier for you to just modify the one line...

stubailo:
Yep, doing that now!

stubailo:
Done: 165ea96e0d44f4b371b6f339b8fa237537f8b60f

Thanks for noticing.

PR #6190, opened 2016-02-08
[Docs] Proper detection of key codes for search

fabienb4:
Avoid opening search on function (FN) keys press (`String.fromCharCode()` returns letters with them).

### DIFF
--- a/docs/client/search.js
+++ b/docs/client/search.js
@@ -19,12 +19,10 @@ $(document).on("keydown", function (event) {
   }
 });
 
-// Open search with any non-special key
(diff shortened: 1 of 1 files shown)


stubailo:
Sure, looks good! Will be deployed when the docs hit 1.3, unless it's absolutely crucial it's deployed now for some reason.

fabienb4:
No, it's not crucial for sure. It's just annoying that pressing F5 (for example) actually brings up the search, and then refresh. I like it when things work smoothly ;-)

PR #6192, opened 2016-02-08
Add package reflection to the header of files

### LINKED ISSUE #6188: A way to get current Meteor package name
In 1.2, I am not sure if there is a difference with 1.3, but it would be very useful to be able to get current Meteor package name. This would allow one to construct asset URLs without hard-coding stuff. I am doing some stuff with automatically resolving dependencies and loading them through the manifest file, so that would be also useful there.

mitar:
Fixes #6188.

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -947,6 +947,17 @@ var getHeader = function (options) {
     );
   }
 
+  if (options.combinedServePath || options.name) {
+    chunks.push("/* Package reflection */\n");
+    if (options.name) {
+      chunks.push("var PackageName = '" + options.name + "';\n");
+    }
+    if (options.combinedServePath) {
+      chunks.push("var PackagePath = '" + options.combinedServePath.replace(/^\//, '') + "';\n");
+    }
+    chunks.push("\n");
+  }
+
   return chunks.join('');
 };
 
@@ -1114,7 +1125,9 @@ export var fullLink = Profile("linker.fullLink", function (inputFiles, {
   // into a single scope.
   var header = getHeader({
     imports,
-    packageVariables: _.union(assignedVariables, declaredExports)
+    packageVariables: _.union(assignedVariables, declaredExports),
+    combinedServePath,
+    name
   });
 
   let exportsName;
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -950,10 +950,11 @@ var getHeader = function (options) {
   if (options.combinedServePath || options.name) {
     chunks.push("/* Package reflection */\n");
     if (options.name) {
-      chunks.push("var PackageName = '" + options.name + "';\n");
+      chunks.push("var PackageName = " + JSON.stringify(options.name) + ";\n");
     }
     if (options.combinedServePath) {
-      chunks.push("var PackagePath = '" + options.combinedServePath.replace(/^\//, '') + "';\n");
+      chunks.push("var PackagePath = " +
+        JSON.stringify(options.combinedServePath.replace(/^\//, '')) + ";\n");
     }
     chunks.push("\n");
   }


tools/isobuild/linker.js
```diff
@@ -612,6 +612,17 @@ var getHeader = function (options) {
     chunks.push("var " + options.packageVariables.join(', ') +
                 ";\n\n");
   }
+  if (options.combinedServePath || options.name) {
+    chunks.push("/* Package reflection */\n");
+    if (options.name) {
+      chunks.push("var PackageName = " + JSON.stringify(options.name) + ";\n");
+    }
+    if (options.combinedServePath) {
+      chunks.push("var PackagePath = " +
```
stubailo: What would one use `PackagePath` for? Is it necessary for the asset use case?

stubailo:
I think this is a great idea. What are the pros and cons of attaching this information to the already-global `Package` variable instead of creating a new one?

mitar:
See #6203 for some related ideas.

tools/isobuild/linker.js
mitar: In my world this is the same as the path in the manifest file. So you can use it to resolve the entry in the manifest file where you can get some extra information.

mitar:
> What are the pros and cons of attaching this information to the already-global Package variable instead of creating a new one?

Only that I do not know how I would do that? Like naming. Maybe `Package.self.name`? `Package._name`?

tools/isobuild/linker.js
mitar: But frankly the main use case for me was for dynamic loading (#6203) of modules. So that I could lookup into the manifest file and see dependencies (one other change I added to the manifest file for each entry there). But then I discovered that it is simply easier to list directly all dependencies in the header of the file. Instead of having a key to resolve in manifest to load dependencies.

stubailo:
Moved to Zol's plate, I don't know what to do about this. But I agree it's a good idea.

zol:
@mitar - this looks reasonable and given that you have a need for it I'd like to merge it. @stubailo, how were you thinking it would get added to the `Package` global?

For instance, let's say currently my `Package` var is:

```
Package = {
  ui: {
     {Blaze: Object, UI: Object, Handlebars: Object}
  }
}
```

What would it become with your proposed API?

stubailo:
Probably just 'Package.name' or something.

mitar:
`Package.name` would mean that there should never be a core package name with name `name`? :-)

mitar:
What about `Package._name` and `Package._path`?

mitar:
> this looks reasonable and given that you have a need for it I'd like to merge it.

Just to be  clear, at the moment I am using this in one my branch with also other changes (like dynamic loading). I am not sure how much does it help on its own, but I know that it would help later on with improving how one can construct asset URLs on the client side.

zol:
@mitar Yeah, I don't like `name` either, not because it prohibits a core package from existing but it doesn't conform with the conventions it's peers are following (`name` acts like a namespace not a package reference). I personally would go with `__name__` and `__path__` to distinguish that these objects have special meaning.

mitar:
The only issue with `__name__` and `__path__` I have is that it looks very similar to some internal JavaScript object properties (like `__proto__`).

mitar:
OK, I am making this pull request more complicated and against my otherwise approach of "perfect is an enemy of good". ;-) Maybe I should just be quiet and do it as you propose. :-)

mitar:
Hm, package names are pretty restricted. Maybe we can use some symbol which is forbidden in a package name?

zol:
That's a good idea and good point re: internal js properties.

mitar:
> That's a good idea and good point re: internal js properties.

Not sure what you are saying. That we should then go with `__name__` or not? :-) So, just make a call and I will do it.

zol:
Hey, Sorry for the delay - I meant your point regarding the naming of internal properties was a good one. Having thought about it more I think we should still use two underscores because using only one would break our own convention of 'private, don't use'.

mitar:
What about `__name` and `__path`?

benjamn:
Note that you can use `module.id` for something like this when using modules: `require(module.id)` always give you back the current `module.exports`.

mitar:
But not name and path, no?

zol:
Hey @mitar - Sorry I dropped the ball on this one earlier. I'd move forward and merge it except it seems somewhat pointless now that we know we're going to be moving towards a full npm transition as soon as possible. As a result I'm going to close the PR out now, sorry about the lost time/work involved. Please create an issue and subsequent PR as we work on the modules transition if it looks like this use-case is not going to be possible with our proposed solution.

mitar:
> Please create an issue and subsequent PR as we work on the modules transition

Maybe just add it to the list. :-) So some package/module reflection capability. :-)

mitar:
Could we reopen this pull request and merge it? It seems NPM is not yet coming.

hwillson:
Re-opening sounds reasonable to me @mitar. I haven't had a chance to review the changes, but I'll take a look early next week. Thanks!

hwillson:
Hi @mitar - would you mind rebasing? Thanks!

mitar:
Rebased.

zimme:
Did we want to put this into `Package` under `_name` and `path` instead of new variables?

benjamn:
`Package` is a global variable which appears the same to every package, so I don't think that's an option. This reminds me of why `System.import` was a terrible API compared to a per-module dynamic `import()` function—`System` was global, so there could be only one version of the `System.import` method, so you always had to pass the current `module.id` as a second argument, which was super annoying.

Speaking of per-module variables, I would be much more interested in finding a way to attach metadata like this to the `module` object, rather than introducing new pseudo-global variables like `PackageName` and `PackagePath`. Specifically, `module.package.{name,path}` would be ideal.

Before anyone points out this idea requires you to be using the `modules` package, let me preemptively state that I do not consider that a valid objection.

zimme:
Would that mean old packages wanting to use this would have to be re-written using modules or would they just be able to import from `modules` to get to the new vars?

benjamn:
@zimme Old packages would have to include `api.use("modules")` in their `package.js`, but they would very probably still work with no additional changes, thanks to our tireless back-compat efforts.

It's also worth noting that the current version of this PR would only work for packages used in next-version-or-later Meteor apps, so there's not much of a backwards compatibility story here, in any case.

I'm going to close this PR now, because I think it's unlikely we will implement this feature by declaring new package variables. Thanks for raising the idea, all the same!

mitar:
Hm, so what is the path forward here?

benjamn:
For package name:
```js
const packageName = module.id.startsWith("/node_modules/meteor/")
  ? module.id.split("/")[3]
  : null; // Indicates an app instead of a package.
```
I'm not convinced your `PackagePath` variable is useful, since it no longer makes sense in production (where all the JS is combined), and the location/contents of manifest files isn't something I would encourage you to rely on.

mitar:
So the above works only for modules? The main reason why package name is useful for me is that one does not have to hardcode paths to package assets on the client side. If I have an asset on the client side, and I want to access it, how do I get to the URL of that asset?

PR #6201, opened 2016-02-09
For path to match the one in the manifest

mitar:
This code tries to match [`setUrlFromRelPath`](https://github.com/meteor/meteor/blob/devel/tools/isobuild/bundler.js#L364) but it seems it was not updated when escaping was added.

Probably this should be extracted somewhere out. But for now, it can be at least synced.

Only cosmetic change.

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -431,7 +431,7 @@ _.extend(File.prototype, {
     }
 
     var chunks = [];
-    var pathNoSlash = self.servePath.replace(/^\//, "");
+    var pathNoSlash = self.servePath.replace(/^\//, "").replace(/:/g, '_');
 
     if (! self.bare) {
       var closureHeader = "(function(){";


mitar:
And this one, @benjamn.

tools/isobuild/linker.js
```diff
@@ -431,7 +431,7 @@ _.extend(File.prototype, {
     }
 
     var chunks = [];
-    var pathNoSlash = self.servePath.replace(/^\//, "");
+    var pathNoSlash = self.servePath.replace(/^\//, "").replace(/:/g, '_');
```
benjamn: Can you use `require("../utils/colon-converter.js").convert` here instead?

benjamn:
Also, if you rebase, the tests will probably pass now, which is a requirement for merging to `devel`.

zol:
Thanks @mitar . I'm happy to merge this if you make @benjamn's requested change and rebase against the latest `devel`.

zol:
Closing this due to inactivity. Please open a new PR with the suggested updates.

mitar:
Groan. I just didn't have time. ;-)

zol:
Sorry :)

PR #6211, opened 2016-02-10
Rename separated minifiers to singular name and use separate versions…

sebakerckhof:
This is a follow up PR for this PR: https://github.com/meteor/meteor/pull/5592

It does 2 things:
1) Rename a single minifier to its singular name, because it's just better naming.
2) Use separate minifiers in new projects and when upgrading to 1.3, because otherwise swapping out a minifier is cumbersome.
In most cases a minifier will only handle css or js and not both. 
Take for example my [autoprefixer](https://atmospherejs.com/seba/minifiers-autoprefixer) plugin. If a user wants to use this and had included the `standard-minifiers` package, he will have to do 3 things:

1. remove standard-minifiers 
2. add seba:minifiers-autoprefixer
3. add back standard-minifier-js

After this PR, this would be:

1. remove standard-minifier-css
2. add seba:minifiers-autoprefixer


stubailo:
@juliancwirko what do you think about this?

juliancwirko:
Yeah, I tought about this too when rewriting postcss package ;) See new readme: https://github.com/juliancwirko/meteor-postcss/tree/meteor-1.3-in-progress 
I think this change is needed because otherwise swapping minifiers is quite odd (although still simple) ;) 
With this change in default we will have separated js and css minifier without `standard-minifiers` 'umbrella package'. Sounds good for me.

Of course I don't know exactly how this change will affect backward compatibility and other relations, but it seems that there shouldn't be any problems.

sebakerckhof:
The `standard-minifiers` package is still present for BC.

stubailo:
OK I'm good to merge this, and it's going to be up to you @sebakerckhof and @juliancwirko to update your docs to make sure people know which packages to remove.

stubailo:
Should go out in the next beta release, @benjamn can we put this on some sort of list?

Issue #3876, opened 2015-03-06
Match.Optional should accept null in addition to undefined


stubailo:
Checking arguments to methods becomes quite silly otherwise.

First of all, arguments that are passed in as `undefined` are converted to `null` by DDP, so you can't use `Match.Optional` in its current state to validate missing arguments in methods on the server.

Second, even if you used `Match.OneOf(null, <type>)`, it still wouldn't work because the method stub on the client actually _does_ get undefined.

Perhaps we should reconsider our code style guidelines of setting things to `null` instead of `undefined`, or modify our default tools to accept `null` as meaning "missing value".

meonkeys:
+1

rijk:
+1

voidale:
+1

panphora:
+1

mjesuele:
This just bit me as well, and is rather unintuitive behavior. +1!!

acemtp:
+1!!!

Jank1310:
+1

LewisW:
+1

stubailo:
I would accept a pull request for a new call, `Match.Maybe`, which has the correct behaviour - accepts `undefined`, `null`, and the passed value.

Then we can deprecate `Optional`. I don't think it would be good to silently change the behavior because it could introduce undiscoverable security holes into people's apps.

If someone acts fast, we can get this into Meteor 1.3!!!

stubailo:
Fixed this by adding `Match.Maybe`. Going to make the docs recommend that instead.

mitar:
We developed those extensions: https://github.com/peerlibrary/meteor-check-extension

We have `Match.OptionalOrNull` to address this issue.

lirbank:
A simple workaround, for pre 1.3, is to substitute Match.Optional(<type>) with Match.OneOf(null, undefined, <type>) - works on both client and server.

maxenceC:
@lirbank thank you for the workaround !

PR #6220, opened 2016-02-11
Added Match.Maybe to allow optional matching with null support

LewisW:
This pull request implements the proposed Match.Maybe suggested in #3876. Match.Maybe adds support for a null value being accepted as an optional value. This is necessary because DDP converts undefined values to null values.

I have not updated the docs yet, I wanted to check I was on the right track before adding something to the docs.

### DIFF
--- a/docs/client/full-api/api/check.md
+++ b/docs/client/full-api/api/check.md
@@ -111,6 +111,10 @@ only if the key is not set as opposed to the value being set to `undefined`.
 
 {{/dtdd}}
 
+{{#dtdd "<code>Match.Maybe(<em>pattern</em>)</code>"}} Behaves like `Match.Optional`
+except it also matches `null`. If used in an object, the behaviour is identical to `Match.optional`.
+{{/dtdd}}
+
 {{#dtdd "<code>Match.OneOf(<em>pattern1</em>, <em>pattern2</em>, ...)</code>"}}
 Matches any value that matches at least one of the provided patterns.
 {{/dtdd}}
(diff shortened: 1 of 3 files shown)


stubailo:
Looks awesome!

Please update the docs and I'll merge.

sebakerckhof:
Very useful.

LewisW:
Documentation updated. I wasn't sure if you wanted me to recommend it over Match.Optional or just document it as an alternative and trust Meteor developers to use it themselves when appropriate.

stubailo:
Merged as one commit: 7a80a8c8517dc721974141d77b7ee70492adcac2

Thank you!

Going to update the docs to make it stand out more, and possibly say that `Optional` is no longer recommended.

abernix:
@stubailo had wrote:
> I don't think it would be good to silently change the behavior (of `Match.Optional`) because it could introduce undiscoverable security holes into people's apps.

Unfortunately, this PR did cause `Match.Optional` to work identical to `Match.Maybe`, even though it wasn't intended to.  There weren't any existing tests for `Match.Optional` on `null` and none were added.  Didn't find a security hole (luckily) but did find this problem. :smiling_imp: 

The current version is just setting `var Maybe = Optional;` but checking the `instanceof` in an unreachable `else if` (because of them being `===`).  Essentially the  `Match.Optional` logic was not being called at all in 1.3 code, instead always using `Match.Maybe`.

See #6735 for more information and #6736 for the PR fix.  Added a bunch of tests that prove the regression and fix too.

stubailo:
Oh man, that's a great catch.

PR #6234, opened 2016-02-14
Non developer readme contribution

Kauress:
I have written a clear, easy-to-comprehend introduction for Meteor. I believe this introduction, free of complex technical terms, will swiftly and thoroughly inform anyone about Meteor.


stubailo:
I think these introductions belong on meteor.com, not the readme - perhaps we should open a discussion on the forums or something, but I would prefer to keep the readme as small as possible.

benjamn:

Issue #6099, opened 2016-01-27
Cross target package dependencies


mitar:
I stumbled across what I think is currently a limitation of the package system (1.2). The issue is that I would like to create a package, which when you add it on `web.browser` target would add files both to `web.browser` and `web.cordova` targets, if the latter is enabled for the whole app. It seems this is not possible. Currently, I have to list both `web.browser` and `web.cordova` when doing `api.use`. This is problematic because the fact about `web.cordova` file is something which is internal to the package implementation. The exposed API and symbols are reasonable only for the `web.browser`. (`web.cordova` only gets some assets and some other tweaks which then make it work with the `web.browser`.)

(The target names here are just example, I in practice have some custom targets.)

avital:
It the example above with `client` and `web.cordova` actually true? I am pretty sure passing in `"client"` into `api.use` makes that package available on both browser and Cordova apps.

mitar:
I will update the description to make it clearer how I would like this to behave.

mitar:
So, to be more concrete, I have a package, which does:

```
api.addFiles('browser.js', 'web.browser');
api.addFiles('cordova.js', 'web.cordova');
```

And I would like to do in another package:

```
api.use('my-package', 'web.browser');
```

But that should add also `cordova.js` to `web.cordova` if it is enabled, but not that user has to list it when doing `api.use`, because it should be internal to the implementation of my package.

One example for this is code for web workers. Imagine that we have `web.worker` target. I would like to create a package `my-super-rendered-in-web-worker` which has `api.js` for `web.browser` and `worker.js` for `web.worker` targets. Users should not know that the package is using web workers internally. You would want only to load the package on the `web.browser` because this is where the public API for the package is. And then my package could check and load `worker.js` from `web.worker` target internally.

mitar:
Hm, I am looking into this and I have issues implementing it. So it is not easy to get metadata in, but the issue is that each target is build separately, it seems. So when a target `web.browser` is being build, there is no way to tell the target `web.worker` that when it will be build, it should include the X extra package (because a package from `web.browser` depends on it).

mitar:
I made #6238 pull request.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

[...]

Issue #6222, opened 2016-02-11
Add new target: web.worker


mitar:
**_1 Upvote**_ This is a ticket to discuss one idea I implemented elsewhere and would like to bring it into the Meteor core.

I would like to add new target `web.worker` to the Meteor. This would allow one to use the whole toolchain (ecmascript, coffeescript, combining multiple packages) for various web workers (web workers, service workers, etc.). This would address the following issues: https://github.com/meteor/meteor/issues/5982 https://github.com/meteor/meteor/issues/2751

I have the code ready in my fork and works pretty nice.

Is this something I should make a pull request? What would be guidelines and design considerations here?

One open issue related to this is, that once you have this, some such feature is needed: https://github.com/meteor/meteor/issues/6099 My current plan to address this to add `api.onUseDependices({target: [listOfDependingTargets]})` which would register that when somebody depends on the package with `target`, also `listOfDependingTargets` should be added.

This allows one to depend a package only for `web.browser`, but would pull in also `web.worker` because it is used internally.

The other thing which is needed is something like https://github.com/meteor/meteor/issues/6203, which allows one to dynamically resolves dependencies. So if I do `importScripts('packages/ejson.js')` this will automatically pull also `packages/base64.js` as a dependency, for example. Works pretty well for both workers and `<script>` tags in my implementation.

This pull request https://github.com/meteor/meteor/pull/5714 should probably also be pulled in to get manifest file have correct URLs for content. In this way then one does not have to hard-code URLs to workers' scripts, but can read manifest and resolve them (I would not provide mechanism for resolving as part of this pull request, but it can be done later on).

mitar:
cc @stubailo: Let's see how discussion of a feature request would work on GitHub.

stubailo:
I'd like to note that this is much more than a feature request, this is a _Very Major_ feature request (in my view). And then pass it to the platform team, @avital and @benjamn.

mitar:
Hm, it is less than 100 lines of code from what I see at the moment. :-)

mitar:
Most of code changes were done because of the Cordova. This would only add one more identifier to all of those lists of possible targets. The rest is more or less already taken care of or already merged in Meteor recently by @martijnwalraven.

mitar:
The dependency thingy is the only one really changing things, but that could be also limited only to this platform.

trusktr-goin:
Would this automatically launch all scripts for `web.worker` in a worker? In the same worker? One worker per file? How would that be configured?

[...]

PR #6238, opened 2016-02-15
Provides architectural dependencies between targets

mitar:
Fixes: #6099 

Now you can use `api.archDepends({client: 'web.worker'})` to get `web.worker` content automatically included when `client` dependency is made.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -422,6 +422,7 @@ class Target {
 
     // the architecture to build
     arch,
+    allArchitectures,
     // projectContextModule.CordovaPluginsFile object
     cordovaPluginsFile,
     // 'development' or 'production'; determines whether debugOnly
@@ -438,6 +439,7 @@ class Target {
 
     // Something like "web.browser" or "os" or "os.osx.x86_64"
     this.arch = arch;
+    this.allArchitectures = allArchitectures;
 
     // All of the Unibuilds that are to go into this target, in the order
     // that they are to be loaded.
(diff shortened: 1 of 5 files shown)


stubailo:
If there was a list of requirements to getting a PR accepted (which we are working on ASAP with @martijnwalraven), having a detailed discussion about the feature and then agreement that it should be in Meteor core would be at the top of the list. The issue referenced was filed by you and doesn't have any discussion in support of this particular addition, so I think it would be a mistake to accept a PR implementing it.

mitar:
> having a detailed discussion about the feature and then agreement that it should be in Meteor core would be at the top of the list. 

Here is the ticket with the discussion: #6222

> doesn't have any discussion in support of this particular addition, so I think it would be a mistake to accept a PR implementing it.

I agree. The pull request is made so that in that those who would like to use it can find it, and give their say on the topic.

mitar:
(Hm, this is an interesting idea, imagine that you could do `meteor update --with-pull-request 6238` and it would create a version of Meteor for you with that pull request added. In this way people could enable Meteor features for them only. And pull request would have the implementation which could also evolve through time. If Meteor would collect stats about `with-pull-request` calls you could then determine which of the pull requests should be merged in.)

stubailo:
@mitar I think packages do a pretty good job of extending Meteor today - for example, you can use a local fork of a package if you really want to (we could make it much easier to use forks without having a local copy, but that's another thing). I think the real problem is that it's basically impossible to extend the Meteor tool in any meaningful way - if there was some concept of plugins or extensions (beyond just compiling a certain file extension) then it would hopefully be easy to add these features.

> The pull request is made so that in that those who would like to use it can find it, and give their say on the topic.

I feel like this is not the role of pull requests, especially if people consider it a problem when pull requests aren't accepted. If even the author of the PR doesn't think it should be merged, is it something that should be open in the first place?

mitar:
> If even the author of the PR doesn't think it should be merged, is it something that should be open in the first place?

Hm, I do not know how you use open source projects, but often check existing pull requests and use them when I need some feature which is missing from core version. GitHub nicely finds such pull requests. And sometimes it is easier just to check the code than having a long discussion how it could be.

Also, I think the PR should be merged, but not yet. So after the discussion is resolved and/or after more people say that they like the idea. This is why pull request is useful. They can test it and +1 it (or -1).

D1no:
> imagine that you could do `meteor update --with-pull-request 6238` and it would create a version of Meteor for you with that pull request added

I am not sure how this would be implemented, but this would be an incredible nice feature to relax the tension between MDG and the Community by allowing people to easily venture ahead for certain features without seeing MDG liable for it. From my personal experience, maintaining a private meteor fork is very disconnected and quite a challenge. This would certainly reduce the ongoing Black vs White binary decision making if something goes into core or not. @stubailo maybe you can provide your 2 cents

martijnwalraven:
One of the problems with pull requests is that they quickly get out of date and you will have to deal with merge conflicts, so automating this would be a challenge. When combining multiple pull requests you will likely run into even more conflicts.

I agree with @stubailo that for most features, packages should be a better solution. And if they are not today we can think about ways to improve their usefulness. The issue with this particular feature request (i.e adding a `web.workers` architecture) is that it requires additions to the build system and tool. And especially with the more fundamental changes we're planning on making to those, this is not the time to add more complexity.

benjamn:

mitar:
Just for record. I think I found some issues in implementation here later on. I have some updated code. If anyone else is interested in this, I can try to iron our the issues and redo a pull request. But for now I will not be spending more time on this because it works for me.

PR #6263, opened 2016-02-18
Update rules `space-after-keywords` and `space-return-throw-case` to `keyword-spacing`

lnmunhoz:
Replace old rules by new one.
- http://eslint.org/docs/rules/space-after-keywords
- http://eslint.org/docs/rules/space-return-throw-case

> Replacement notice: This rule was removed in ESLint v2.0 and replaced by **keyword-spacing** rule. Some style guides will require or disallow spaces following the certain keywords.

### DIFF
--- a/scripts/admin/eslint/.eslintrc
+++ b/scripts/admin/eslint/.eslintrc
@@ -162,11 +162,10 @@
       "before": false,
       "after": true
     }],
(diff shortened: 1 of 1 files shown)


zol:
We're using a fixed version of eslint so this isn't a problem. We'll fix it when we update eslint itself.

PR #6287, opened 2016-02-22
Take a test runner as commandline argument

tmeasday:
This isn't quite the complete solution we are planning, but it is enough to allow others to test alternate runners. 

I've tested this with https://github.com/tmeasday/meteor-mocha/tree/1.3-reporter, and it works against Todos (you need to `meteor remove avital:mocha; meteor add tmeasday:mocha`)


tmeasday:
I'm not convinced the best approach isn't just to allows require the `--driver-package` argument and not get people to add their driver to `.meteor/packages` btw

avital:
I think the difference between an argument that adds a package and a package you add to your app as a default isn't so important ("bikeshedding").

This change looks good.

PR #6290, opened 2016-02-22
Added some help text for meteor test-app

tmeasday:
First pass at this. TBH I'm not totally sure what goes in here. Am I right to call out to the guide? Should we do this in other spots of the help?

### DIFF
--- a/tools/cli/help.txt
+++ b/tools/cli/help.txt
@@ -532,7 +532,65 @@ Options:
   --verbose         Print all output from builds logs.
   --no-lint         Don't run linters used by the tested packages on every
                     test app rebuild.
+>>> test-app
+Test the application
+Usage: meteor test-app [options]
+       meteor test-app --unit [options]
+       meteor test-app --integration [options]
+       
+Runs tests against the application, either in *unit* or *integration* 
+test mode. meteor test-app will start a special app based on a test driver
+which handles the task of running tests and displaying the results in the 
+browser when you visit it.
 
+In unit test mode, no files in your application are eagerly loaded, aside
+from test files (in you tests/ directory or named *.tests.*). You can
+import your app's modules from within your tests and test them in isolation.
+
+In integration test mode, your app is loaded as usual, and your tests can
+inspect and effect the running state. Test files are loaded similarly to in
+unit test mode.
+
+Open the test dashboard in your browser to run the tests and see the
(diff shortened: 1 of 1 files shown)


tools/cli/help.txt
```diff
@@ -532,7 +532,65 @@ Options:
...
+       meteor test-app --unit [options]
+       meteor test-app --integration [options]
+       
+Runs tests against the application, either in *unit* or *integration* 
+test mode. meteor test-app will start a special app based on a test driver
+which handles the task of running tests and displaying the results in the 
+browser when you visit it.
+
+In unit test mode, no files in your application are eagerly loaded, aside
+from test files (in you tests/ directory or named *.tests.*). You can
```
avital: FWIW, Right now it also finds files named `tests.*` or `test.*` or `*.test.*`. (We can change that if we want)

tools/cli/help.txt
```diff
@@ -532,7 +532,65 @@ Options:
...
+import your app's modules from within your tests and test them in isolation.
+
+In integration test mode, your app is loaded as usual, and your tests can
```
avital: Maybe "your app is loaded as usual, and then made hidden, ..."?

tools/cli/help.txt
```diff
@@ -532,7 +532,65 @@ Options:
...
+                    in a browser or in console logs.
+
+  --ios,            Run tests in an emulator or on a mobile device. All of
+  --android,        the tests for client and server will run in addition to
+  --ios-device,     mobile-specific tests.
+  --android-device
+  --test-app-path   Set the directory in which to create a temporary app used
+                    for tests. Defaults to the system's temporary directory,
+                    usually /tmp.
+  --velocity        [Experimental] Execute tests using phantomjs and exit.
```
avital: Is that really the only difference? (If so, wow.)

avital:
This is so cool! I never knew about some of those options, I guess we just got them for free from test-packages?

avital:
And I think calling out the guide is great! Yes, we should do it for more commands!

avital:
The help should explain the default behavior (which is `--integration`). Or, we could require passing an argument, that may be better?

tmeasday:
Please take another go-round

avital:
LGTM

Issue #5784, opened 2015-12-12
Application testing improvements for Meteor 1.3


benjamn:
- [x] Find tests in `tests/` subdirectories
- [x] Optimize `meteor test-app` - copy build dir before starting.
- [x] `Meteor.isTest`, `Meteor.isIntegrationTest`, `Meteor.isUnitTest`.
- [x] Implement `meteor test-app --unit`
- [x] `meteor test-app` to find all test files, even inside `imports/`
- [x] Why do integration tests no longer correctly hide the app UI?
- [x] Why do Blaze unit tests fail on the Todos app
- [x] Can we actually have two test reporter packages that both expose Mocha added to the app?
- [x] Simplify process for building test runner packages (clear CSS, clear DOM, stop router).
- [x] Require passing in a driver package option
- [x] Figure out a directory scheme for acceptance (i.e. outside of Meteor) tests #6321
- [x] Add help text so that `curmeteor help test` works
- [x] Add test to skeleton and example apps. #6303 #6306

philipmjohnson:
Hallelujah!  Relief from app-level testing madness is finally in site!

Sanjo:
Let me know if you have something to show or want some input. I did a lot of the work in Velocity, including the meteor --test-app implementation in the Velocity Meteor fork.

avital:
Here are my current thoughts:

The `meteor test-app` command (that will later be renamed to plain `meteor test`) will be modeled after `meteor test-packages`. Here's how it will work:

* Instead of Velocity mirrors that basically make copies of the app, it will bundle the same app directory into a new temporary build directory. It still uses the same "watch files for changes" logic (which can run in parallel to running the app normally), so you get hot code push for tests.

* This process adds a test reporter package that can be specified as a command line argument as part of this process. This doesn't affect any parallel non-test runs of the app.

* Running with `--unit` only loads test files explicitly passed in as arguments and their dependencies. Running with `--integration` also loads the entire app.

This isn't the entire story but I thought I'd jot down these thoughts in parallel to working on them.

@Sanjo -- what do you think? Have I correctly understood the purpose of Velocity mirrors? Will this plan resolve the need for using them, at least for the case of not running parallel test chunks on different servers?

stubailo:
You mentioned a test reporter package, what about the test runner? For example mocha, tinytest, etc. Right now test-packages takes a special argument, will test-app have something similar?

avital:
[...]

PR #6309, opened 2016-02-23
Require that you specify a driver package for app tests

tmeasday:
For a subtask of #5784

### DIFF
--- a/tools/cli/help.txt
+++ b/tools/cli/help.txt
@@ -540,8 +540,10 @@ Usage: meteor test-app --driver-package <driver> [options]
        
 Runs tests against the application, either in *unit* or *integration*
 (the default) test mode. meteor test-app will start a special app based on
-a test driver (specified with --driver-package) which handles the task of
-running tests and displaying the results in the browser when you visit it.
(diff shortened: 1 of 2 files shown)


tools/cli/help.txt
```diff
@@ -534,14 +534,14 @@ Options:
...
+       meteor test-app --driver-package <driver> --integration [options]
        
-Runs tests against the application, either in *unit* or *integration* 
-test mode. meteor test-app will start a special app based on a test driver
-which handles the task of running tests and displaying the results in the 
-browser when you visit it.
+Runs tests against the application, either in *unit* or *integration*
+(the default) test mode. meteor test-app will start a special app based on
+a test driver (specified with --driver-package) which handles the task of
+running tests and displaying the results in the browser when you visit it.
```
avital: I wonder now, how will people find which driver packages they should use? Should this help article link to the guide where we describe this?

avital:
Left one comment, but LGTM!

tools/cli/help.txt
tmeasday: Yeah, I think it should. I guess I'll put in where I think the URL will live.

Issue #6321, opened 2016-02-24
Changes to testing file matching patterns


tmeasday:
Moving from https://github.com/meteor/meteor/issues/5784#issuecomment-187518773

#### Unit tests:

- matching *.tests.* but not *.integration.tests.*
- or test*

#### Integration tests:

matching *.tests.* but not *.unit.tests.*
- or test*

#### Acceptance tests (ignored by all modes of meteor)

- inside top level `acceptance-tests/` directory.

avital:
I thought we'd do `tests/integration/`. Sounds a little weird to special case acceptance and not integration. And it would be surprising that simply by being in `tests/` as opposed to a file named `test_foo.js` is an integration test instead of a unit test. What do you think?

tmeasday:
So other stuff inside `tests/` are unit tests?

avital:
Maybe the existence of files in `tests/` throws an error? But this might not really matter.

tmeasday:
Ok, changed the requirements above

PR #6338, opened 2016-02-25
Changes to testing file matching patterns #6321

tmeasday:


tools/isobuild/package-source.js
```diff
@@ -1386,10 +1391,14 @@ _.extend(PackageSource.prototype, {
     // Ignore the usual ignorable files.
     sourceReadOptions.exclude.push(...ignoreFiles);
 
-    // Unless we're running tests, ignore source files with name
-    // "*test*.*", "*tests*.*", "test.*", "tests.*"
+    // Unless we're running tests, ignore all test filenames and if we are, ignore the
+    // type of file we *aren't* running
     if (!global.testCommandMetadata) {
-      Array.prototype.push.apply(sourceReadOptions, TEST_FILENAME_REGEXPS);
+      Array.prototype.push.apply(sourceReadOptions.exclude, TEST_FILENAME_REGEXPS);
```
avital: I think you also need to push `UNIT_TEST_FILENAME_REGEXPS` and `INTEGRATION_TEST_FILENAME_REGEXPS` here.

But really, I'd change it a bit and have `TEST_FILENAME_REGEXPS` include both of those already where it's defined (using the ES6 array spread operator), and actually not change the code here.

And then I'd rename `TEST_FILENAME_REGEXPS` to `ALL_TEST_FILENAME_REGEXPS`. That makes the logical relationship between these three arrays match their names.

tools/isobuild/app-test-files.js
```diff
@@ -8,6 +8,18 @@ export const TEST_FILENAME_REGEXPS = [
   /^tests?./
```
avital: At this point we might consider just getting rid of the `tests?.*` form... Not only does it stick out in it's difference from the unit test and integration test filenames, and makes it harder to exactly explain how this feature works -- it also makes colocating tests not as great. It's much better to see `foo.js` and `foo.test.js` right next to each other in your editor than a bunch of files and then a bunch of `test_*` files where you can't even easily see which files have tests.

What do you think?

tmeasday:
@avital please review again

avital:
Looks great! Worth verified at least once that `test-packages` still works (I'm pretty sure it does but the code changed is on that code path as well)

Issue #5787, opened 2015-12-12
Tool changes for Meteor 1.3


benjamn:
Command line options to pick the org to which you’re deploying / display the org it belongs to when you deploy. (https://paper.dropbox.com/doc/Product-Design-Meteor-CLI-for-Galaxy-Deploy-and-Transfer-mCkFOLLcNEzI3vsCuAXel)

avital:
See https://github.com/meteor/galaxy-server/issues/631

timbotnik:
@benjamn let me know when this appears on the radar for you guys, happy to pair on it.  I actually think the first 3 "Deploys fail" tasks can be handled by the existing error message passing pipeline, but we should verify that.  The more pressing issue here is the option to specify a galaxy account name when deploying.  It's a big source of unnecessary complexity and user confusion in galaxy so the sooner we can rid ourselves of it the better.

avital:
https://paper.dropbox.com/doc/Product-Design-Meteor-CLI-for-Galaxy-Deploy-and-Transfer-mCkFOLLcNEzI3vsCuAXel

zol:
@tmeasday - Last week @avital indicated he would pair with you on this ticket hence I'm assigning it over to you.

PR #6348, opened 2016-02-26
Changes to the tool due to #5787

tmeasday:


tools/meteor-services/deploy.js
```diff
@@ -144,7 +144,8 @@ var authedRpc = function (options) {
   var infoResult = deployRpc({
     operation: 'info',
     site: rpcOptions.site,
-    expectPayload: []
+    expectPayload: [],
+    qs: options.qs
```
tmeasday: Note that this means all query strings are passed through to all `/info` requests (not just on deploy). I'm not sure if this is likely to cause problems..

tools/meteor-services/deploy.js
```diff
@@ -368,7 +370,8 @@ var bundleAndDeploy = function (options) {
   var preflight = authedRpc({
     site: site,
     preflight: true,
-    promptIfAuthFails: promptIfAuthFails
+    promptIfAuthFails: promptIfAuthFails,
+    qs: options.rawOptions
```
tmeasday: I think perhaps we don't actually have to pass `qs` here -- this request seems to just do the first half of the deploy anyway (i.e. the info part). I'm not entirely sure I understand the purpose of this, TBH.

tmeasday: Oh, OK I read it more. I guess we do the preflight first to avoid having to wait to bundle to crap out. So this is good.

tools/meteor-services/deploy.js
```diff
@@ -436,40 +439,44 @@ var bundleAndDeploy = function (options) {
...
-    });
+  if (result.payload.message) {
+    Console.info(result.payload.message);
+  } else {
+    var deployedAt = require('url').parse(result.payload.url);
+    var hostname = deployedAt.hostname;
+
+    Console.info('Now serving at http://' + hostname);
+
+    if (! hostname.match(/meteor\.com$/)) {
```
tmeasday: Note that it doesn't appear this block of code actually works any more. I should probably just remove it.

tmeasday:
Note that `meteor authorized X --transfer Y` to mother results in:

```
Couldn't change authorized users: Bad request
```

Not sure if we want to do anything about that.

tools/cli/commands.js
```diff
@@ -1387,6 +1382,8 @@ main.registerCommand({
     return deploy.changeAuthorized(site, "add", options.add);
   } else if (options.remove) {
     return deploy.changeAuthorized(site, "remove", options.remove);
+  } else if (options.transfer) {
+    return deploy.changeAuthorized(site, "transfer", options.transfer);
```
avital: Oh, I thought we said that `transfer` would do the same as `add`? At least make sure @pcarrier knows that we'll have an explicit new `transfer` call that should be already functional when we ship Meteor 1.3.

tools/cli/help.txt
```diff
@@ -431,6 +431,7 @@ View or change authorized users and organizations for a site.
 Usage: meteor authorized <site> [--list]
        meteor authorized <site> --add <username>
        meteor authorized <site> --remove <username>
+       meteor authorized <site> --add <username>
```
avital: You accidentally had `--transfer` here.

tools/meteor-services/deploy.js
avital: I would think if anything that's desirable so that Galaxy could always induce a message in the CLI without requiring sending the entire bundle to the server.

avital: But it's a good thing to know!

tools/meteor-services/deploy.js
```diff
@@ -436,40 +442,44 @@ var bundleAndDeploy = function (options) {
...
-              "You've deployed to a custom domain.",
-              "Please be sure to CNAME your hostname",
-              "to origin.meteor.com, or set an A record to 107.22.210.133.");
-            Console.info('-------------');
-          }
-        });
-      }
-    });
+  if (result.payload.message) {
+    Console.info(result.payload.message);
```
avital: Make sure @pcarrier know that with this change the server should make sure to also send back "Now serving at..."

tools/meteor-services/deploy.js
avital: Remove all the code!! (Not all of it... but still)

avital:
LGTM!

tools/cli/commands.js
avital: Oh! I just now read the comment above. Ignore me :) Good plan.

avital:
Re "Note that meteor authorized X --transfer Y to mother results in: Couldn't change authorized users: Bad request". I don't think we should care.

LGTM (with one minor help correction)

PR #6357, opened 2016-02-27
Better skeleton app

laosb:
A picture shows how it looks:
snip20160227_85
I added some links for beginners to getting started, and instructions to begin with an empty project. Also comes with a better design powered by [Skeleton](http://getskeleton.com/).

But there are some thing we should notice: The basic styles come with `skeleton:skeleton`, which is a 3rd-party package (but official Skeleton wrapper). I don't know if we should imply it with the package system, or add its CSS rules to the CSS file. Also if you want to start building your own, you have to remove the package.

I think with Skeleton beginners can build their first app quickly. But some tutorials may need to add `meteor remove skeleton:skeleton` as a step.

I'm not a guy from Skeleton team to promote Skeleton :-)

### DIFF
--- a/tools/static-assets/skel/.meteor/packages
+++ b/tools/static-assets/skel/.meteor/packages
@@ -12,9 +12,8 @@ reactive-var            # Client-side reactive dictionary for your app
 jquery                  # Helpful client-side library
 tracker                 # Meteor's client-side reactive programming library
 
-standard-minifiers
-#standard-minifier-css   # CSS minifier run for production mode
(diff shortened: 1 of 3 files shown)


liangxiaojuan:
hi

laosb:
Now it works with ReactiveVar instead of Sessions, and was written in pretty ES6.

tmeasday:
Thanks very much for this @laosb.

I took some inspiration from this in changing the skeleton but I think we'll steer clear of something styled for now.

PR #6361, opened 2016-02-27
Update react example to React 0.14

aericson:
The current example code for React does not work. There was a change in 5773b7e3b0a that moved to use ReactDOM but the `.meteor/versions` file was not updated.

Besides fixing this, also made changes to remove warnings and removed the use of `findDOMNode` to access the input. This is no longer needed in 0.14, reference: [Starting with this release, this.refs.giraffe is the actual DOM node](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#dom-node-refs).


tmeasday:
Thanks @aericson,

Will need to update it further to use the new 1.3 React API but this is a good start for React 0.14

PR #6363, opened 2016-02-28
Just returning null does not make a publish ready

mitar:


mitar:
Hm, but how does client-side knows to trigger `ready` on a subscription which returns `null`? Because it still does that?

mitar:
Oh, I see that `ready` is a [noop](https://github.com/meteor/meteor/blob/devel/packages/ddp-server/livedata_server.js#L1286) for universal publish methods. So this is why is not needed.

Issue #6306, opened 2016-02-23
Update skeleton app to follow 1.3 Guide recommendations


tmeasday:
- [ ] App structure
- [ ] Modules
- [ ] Tests

tmeasday:
+ no session / actually use a collection.

stubailo:
Should create client/server dirs probably, should have a landing page that tells you to do the tutorial/guide

laosb:
I've actually done a version of template app before and submitted it as a pr #6357 

Having client/server dirs sounds good, but it is not very useful for an one-cup-of-coffee app. Considering a kind of example or something like Mantra CLI could do this initial for a typical production app instead of a template app.

laosb:
I would try to rewrite it with ReactiveDict and use a collection to build a simple todo (any ideas about some easy usage of collections?)

tmeasday:
@stubailo I am going back and forth about this one.

I think @laosb is right in that there is a difference in use-case between a simple starting point and a full boilerplate. In fact I can kind of see three completing goals in what `meteor create` does:

1. Provide a simple starting point for beginners
  - Here `insecure` and `autopublish` make sense
  - A lot of files and directories is a poor choice
  - Including something to point them to the docs and guide is good.
2. Provide the starting point for a production app (i.e. boilerplate)
  - We'd probably put down every directory here: https://github.com/meteor/guide/blob/structure-content/content/structure.md#example-directory-layout
  - Probably *any* content is just annoying
  - We could also set up a bunch of scripts (e.g. linting?) in `package.json`
  - This seems more opinionated and guide-y. It's good but maybe it shouldn't go in the tool? Maybe we should let the community do it?
3. A mini-tutorial
  - e.g. showing how to use collections
  - including a test

I'm sort of turning against 3. I don't think we should not follow our recommendations (i.e. we should `import` and not use the session), but I wonder if the test is adding much?

I'm wondering if we go with 1. and call out that if they want to see a "real" app, they could run `meteor create --example todos`. And if they want to get a boilerplate, they could run ... ?

stubailo:
Hmm, one thing to note is that Meteor doesn't have scaffolding - in Rails the starter app didn't have much, but you could generate views, tests, etc. with the scaffold command.

[...]

Issue #6347, opened 2016-02-26
meteor run ios-device, bugged when running on localhost but ok on mobile-server (1.3 beta 11)


lukastheiler:
I reproduced this strange behaviour with a very simple app, which just lists the elements of a collection on 1.3 beta 11. 

__OK__: ```meteor run ios ```works fine both in the browser and in the ios simulator

__NOT OK__: ```meteor run ios-device ``` browser ok, but no collection item shows up on the physical ios device. 

__OK__: ```meteor run ios-device --mobile-server lukas13b11.meteor.com``` works both on the browser and the physical device. 

Does anyone know a workaround?

martijnwalraven:
In the `meteor run ios-device` case, are you sure your device is able to connect to the local server? They have to be on the same WiFi network, there might be router settings preventing them from connecting to each other, or the computer could be blocking ports.

lukastheiler:
WIFI wasn't the issue, ```meteor run android-device```worked fine. 
However: i just did a ```rm -rf .meteor/local/build```and ran  ```meteor run ios-device``` - and now it works like charm. Very strange. I'll keep an eye on it and do more rm's. 
So i'm guessing we can close the issue for now. 
Thanks! Lukas

PR #6372, opened 2016-03-01
Refactored app skeleton

tmeasday:
See #6306 

 - Split files into the correct place, given the guide
 - Import html file
 - Removed test [it's not a demo]
 - Use ReactiveVar rather than Session
 - Added an info section to guide people further


tmeasday:
@benjamn sending this to you because I undid some of the stuff we discussed :/ (based on discussions w/ others).

This PR takes some inspiration from #6347 (thank you @laosb), but doesn't go as far as styling the skeleton.

Files are named as they would be per the guide. We remove the test as people should do the tutorial if they want to learn about Meteor features.

tmeasday:
A small question: I changed `"main"` in `package.json` to `server/main.js`. Should we add a `"client"` field?

tools/static-assets/skel/client/main.js
```diff
@@ -0,0 +1,22 @@
+import { Template } from 'meteor/templating';
+import { ReactiveVar } from 'meteor/reactive-var';
+
+import './main.html';
+
+Template.hello.onCreated(function helloOnCreated() {
+  // counter starts at 0
+  this.counter = new ReactiveVar(0);
```
stubailo: Why this instead of `ReactiveDict`?

tools/static-assets/skel/package.json
```diff
@@ -2,10 +2,9 @@
   "name": "~name~",
   "private": true,
   "version": "0.0.1",
-  "main": "~name~.js",
+  "main": "server/main.js",
```
stubailo: Should we just remove this field? IMO it's not meaningful to node.

tools/static-assets/skel/client/main.js
tmeasday: I guess the guide says use a dict huh. I'll change this.

tools/static-assets/skel/package.json
tmeasday: It could be premature. What do you think @benjamn?

PR #6382, opened 2016-03-02
Documentation of how to use 1.3 modules in CoffeeScript

GeoffreyBooth:
Per [discussion](https://github.com/meteor/meteor/issues/5788#issuecomment-181988193) with @avital, some documentation of how to use CommonJS to work with modules in CoffeeScript in Meteor 1.3.

### DIFF
--- a/packages/modules/README.md
+++ b/packages/modules/README.md
@@ -14,7 +14,7 @@ Now, you might be wondering what good the `modules` package is without the `ecma
 
 While the `modules` package is useful by itself, we very much encourage using the `ecmascript` package (and thus `import` and `export`) instead of using `require` and `exports` directly. If you need convincing, here’s a presentation that explains the differences: http://benjamn.github.io/empirenode-2015
 
-## Basic syntax
+## Basic ES2015 syntax
 
 Although there are a number of different variations of `import` and `export` syntax, this section describes the essential forms that everyone should know.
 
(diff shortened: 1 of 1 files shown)


tmeasday:
@benjamn I think perhaps you should look over this?

zol:
Hey, Thanks for your PR. Please sign our [contributor agreement](https://contribute.meteor.com/) before we can accept it.

GeoffreyBooth:
Done!

zol:
Thanks!

Dyscrete:
CoffeeScript hasn't had a lot of progress lately, I think the switch to ES6/7 should be an encouraged standard.

GeoffreyBooth:
@Dyscrete CoffeeScript doesn’t really need a lot of progress, since there’s very little in ES2015 that CoffeeScript didn’t already have. Modules are the only substantial difference, in that CoffeeScript (like ES5) relied on CommonJS. CommonJS `require` statements [are required in Meteor 1.3 for certain situations](http://guide.meteor.com/v1.3/structure.html#using-require), even when using ES2015, so CommonJS support will be a part of Meteor for the foreseeable future. And with CommonJS supported, there’s no reason not to support CoffeeScript as well—there’s nothing MDG needs to do really to keep the CoffeeScript support that Meteor already has. That’s why this PR is simply documentation, explaining to CoffeeScript users how to use modules, without any new code to fix any incompatibilities between Meteor 1.3 and CoffeeScript.

GeoffreyBooth:
You're not using this documentation?

lorensr:
+1 on reopening. Might be better if the info were placed here:

http://docs.meteor.com/#/full/coffeescript

Also with a link to it from this section: 

http://docs.meteor.com/#/full/modules

benjamn:
Sorry, this got closed only because the `release-1.3` branch was deleted post-launch. Can you resubmit the PR?

GeoffreyBooth:
Resubmitted as #6636.

rijk:
Wow, this was really helpful, thanks @GeoffreyBooth !! +1 for including it in the documentation, had to Google for about half an our before I found this.

GeoffreyBooth:
Thanks. You should encourage @benjamn or @zol to merge it in ;)

GeoffreyBooth:
#6636, I mean.

packages/modules/README.md
```diff
@@ -96,6 +96,44 @@ import {default as Value, a, F} from "./exporter";
...
+can be rewritten to use CommonJS `module.exports`:
+
+```coffeescript
+module.exports.insert = new ValidatedMethod # ...
+module.exports = incompleteCountDenormalizer
+```
+
+You can also simply write `exports` instead of `module.exports` if you prefer.
+
+Note that [backticks](http://coffeescript.org/#embedded) do *[not](https://github.com/meteor/meteor/issues/6000)* work; your `.coffee` files will be transpiled into JavaScript, but not then handed off to the `ecmascript` package for further transpiling. Any JavaScript you type in backticks will be passed through unmodified all the way to the browser or Node.js runtime.
```
mitar: This is not true anymore.

PR #6392, opened 2016-03-03
Require plain objects for Match.ObjectIncluding

mitar:
Fixes: #6140

### DIFF
--- a/History.md
+++ b/History.md
@@ -13,9 +13,9 @@
 * Move `DDPRateLimiter` to the server only, since it won't work if it is called from the client. It
   will now error if referenced from the client at all.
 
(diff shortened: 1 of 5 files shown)


mitar:
Very interesting. While working on this I discovered a similar bug in jQuery and reported it: https://github.com/jquery/jquery/issues/2968 And now it was fixed. But a similar bug was discovered in lodash and is also now fixed: https://github.com/lodash/lodash/issues/2085

So now this is future proof. If Meteor will use jQuery version 2 or lodash in the future, this check will still work correctly. :-)

@sebakerckhof thanks for creating an amazing test case which failed on all popular JavaScript projects. :-)

sebakerckhof:
Haha, that's actually pretty cool.

Issue #6379, opened 2016-03-01
Move example apps into their own repo and update the tool to tell you to clone it


tmeasday:
***1 Upvote*** We should just hardcode the tool so that we get

```
$ meteor create --example todos
Please clone the repository from https://github.com/meteor/todos
```

stubailo:
:+1: love it! One more option - create a git org called `meteor-examples` and then just `git clone github.com/meteor-examples/${stuff}`

stubailo:
But I don't know if the proliferation of GitHub orgs would be a good thing. Maybe `github.com/meteor/example-${name}`.

tmeasday:
Yeah I prefer the second. I think I'll rename the todos repository.

Re: the three simple todos examples (and similarly the (now two) full todos examples). Are we best having a single repo for these w/ branches? Or 3 separate repos?

stubailo:
Hmm, I guess the question is do we want to unify the issues for the repos?

tmeasday:
The nice part we get for free out of this is we can now create a meteor/example-boilerplate repository.

tmeasday:
Unifying issues and being able to cherry pick non-UI specific commits I think are great benefits.

Downsides are making harder to clone and perhaps more confusing.

PR #6393, opened 2016-03-03
Removed examples and added a command to printout their repos.

tmeasday:
See #6379 

@martijnwalraven could you review this please? I'm still a bit unsure about tool code. Also I copied the commits over to the relevant repos, you might want to review that.


martijnwalraven:
Looks great! Glad we could preserve commit history. We may want to remove some examples for 1.3 though (at least `clock` and `leaderboard`?).

I'll update `localmarket` so it doesn't generate warnings.

I do get an error for `todos`:
```
   While minifying app stylesheet:
   error: Can't load npm module 'autoprefixer' from packages/minifier-postcss_plugin.js. Check your Npm.depends().
```

tmeasday:
Yeah I'm not sure about clock and leaderboard. They are both pretty simple but a little out-of-date. I'm not sure what use they are really. This probably gets to the larger examples strategy.

Re: the issue w/ Todos -- perhaps you need to `npm install`? If it's not that, let's take the discussion to the todos repo

PR #6419, opened 2016-03-05
Update Contributing.md

laosb:
- Removed the old `meteor-core` mail list from it and use Meteor Forums instead.
- Remove some dead messages of pre-1.0 era.


stubailo:
Thanks! Wow, didn't know it was this old.

Issue #6436, opened 2016-03-08
Allow test drivers to initialize on the server


tmeasday:
Either add `runTests` on the server, or just switch the semantics so there's only one test driver loaded at a time.

This will allow:
  1. Test packages that run themselves phantom for CI
  2. Test packages that run server tests then exit (for quick unit tests)

aldeed:
@tmeasday can close?

tmeasday:
Thanks for the reminder

PR #6455, opened 2016-03-10
For #6436 -- allow test drivers to `.start()` on the server

tmeasday:
***1 Upvote*** @zol do you think this is satisfactory syntax?

A driver package can now define `.runTests()` on the client, which kicks off the test process, and `.start()` on the server which can be used for various purposes as discussed on #6436


aldeed:
Looks good to me

Issue #6454, opened 2016-03-10
`meteorEnv.NODE_ENV` is always `"production"` in test mode


tmeasday:
If `process.env.NODE_ENV` is unset in the tool, it somehow get set between bundling and starting the app process. This stops `meteor-env-dev/server.js` from working as intended.

It seems in normal `meteor run`, this: https://github.com/meteor/meteor/commit/40857bdaff7ea648145ff1f1defb4ad8020287de#commitcomment-16607978 means this never happens. 

We should probably do the same for all runs, as I stated in that comment.

tmeasday:
I'm still not totally comfortable about not knowing *why* it got set to production when it started unset, but I think #6461 does the right thing.

PR #6461, opened 2016-03-11
Ensure that runAll always sets NODE_ENV+buildMode sensibly.

tmeasday:
See the commit note and issue #6454

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -357,19 +357,6 @@ function doRunCommand(options) {
     webArchs.push("web.cordova");
   }
 
-  let nodeEnv = process.env.NODE_ENV;
-  if (options.production) {
-    // If options.production, make sure $NODE_ENV === "production".
-    process.env.NODE_ENV = "production";
-  } else if (nodeEnv) {
-    // If $NODE_ENV is set, override options.production according to that.
-    options.production = nodeEnv === "production";
-  } else {
-    // Otherwise make sure $NODE_ENV is set according to
-    // options.production.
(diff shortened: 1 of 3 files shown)


tmeasday:
@benjamn hopefully you agree with the behaviour I specified for the various run modes in the commit note (repeated here):

`meteor run` -- both should be "development"
`meteor run --production` - both should be "production"
`NODE_ENV=production meteor run` - as above
`NODE_ENV=development meteor run --production` - `--production` should win
`meteor test` - NODE_ENV should be "development", buildMode should be "test"
`NODE_ENV=production meteor test" - NODE_ENV should be "production", buildMode should be "test"

It would be great to be able to write tests for the above :/

benjamn:
With these changes, do we still need both `meteor-env-dev` and `meteor-env-prod`? In fact, can we just roll the `meteorEnv` functionality into the `meteor` package, so we don't have to do [this dance](https://github.com/meteor/meteor/blob/release-1.3/packages/meteor/import_meteorEnv.js)?

tmeasday:
@benjamn I think we can remove them.

These changes will ensure that the `NODE_ENV` is set correctly[1] for `meteor run`, `meteor test` and `meteor test-packages`.

However it won't affect other ways of running Meteor, namely bundling + deploying. Do we need to put something in to set `NODE_ENV` to `"production"` by default on app startup if it's unset to handle these cases? Is there any case I'm not thinking of where it should default to `"development"`?

I'll send a PR and we can discuss there

[1] Correctly according to that list above.

PR #6469, opened 2016-03-11
Feature/census package

Urigo:
Adding the `census` package to track anonymous data to help us improve Meteor (optional package)

Also adding it to `meteor create` and `meteor update` since 1.3


packages/census/setup.js
```diff
@@ -0,0 +1,23 @@
...
+    appSecret: process.env.APP_SECRET,
+    rootUrl: process.env.ROOT_URL,
+    statsServerUrl: process.env.METEOR_PACKAGE_STATS_SERVER_URL,
+    reportRate: process.env.CENSUS_REPORT_RATE || 24 * 60 * 60 * 1000,
+    autoSample: Utils.bool(process.env.CENSUS_AUTO_SAMPLE || true),
+    reportAttempts: process.env.CENSUS_REPORT_ATTEMPTS || 3
+  };
+
+  if (!Config.appId)
+    Utils.warn("App id is not provided, sampling won't start");
```
zol: I don't think we should ever log anything to the console.

tools/upgraders.js
```diff
@@ -227,6 +227,12 @@ the guide about breaking changes here:`,
     packagesFile.writeIfModified();
   },
 
+  "1.3.0-census-package": function (projectContext) {
```
zol: Just wanted to double check that this code will only ever try to add the package once - i.e if the user updates to 1.3, removes the package, then updates to 1.3.1 - it's not going to add the package again. Is that correct?

packages/census/setup.js
```diff
@@ -0,0 +1,23 @@
+// Waiting for environment variables to be initialized
+Meteor.startup(() => {
+  Config = {
+    appId: process.env.APP_ID,
+    appSecret: process.env.APP_SECRET,
+    rootUrl: process.env.ROOT_URL,
+    statsServerUrl: process.env.METEOR_PACKAGE_STATS_SERVER_URL,
```
zol: This doesn't ever seem to get set - shouldn't there be a default value that points back to  the activity server.

tools/static-assets/skel/.meteor/packages
```diff
@@ -19,3 +19,4 @@ ecmascript              # Enable ECMAScript2015+ syntax in app code
 
 autopublish             # Publish all data to the clients (for prototyping)
 insecure                # Allow all DB writes from clients (for prototyping)
+census                  # Track anonymous info to help us improve Meteor
```
mquandalle: First, unless I’m wrong reading the code, I wouldn’t qualify an IP address nor the application URL as some “anonymous info”.

Furthermore I believe that adding a tracking package by default is a very bad signal sent to both the Meteor and the broader JavaScript communities. We don’t expect jQuery to send data to “help the jQuery company”, nor React to send data to Facebook, nor Meteor to send data to MDG. At the very least, if you have a real need for this data, let the developers opt-in – instead of giving them a way to opt-out.

My intuition is that built-in tracking code has a risk to deteriorate Meteor reputation. Maybe we should discuss this issue on the forum?

laosb: Truly agree with @mquandalle . Even MDG wants this package to collect data, I suggest not adding it as a default. I think, it's not bad to add it to the skeleton app, but you should notify the user once the app is created. Also, you should notify users by print some logs like `Census is added, Meteor will use this package to collect data. You can remove it by $ meteor remove census.`
I'm against adding it in a update.

zol:
I wanted to add some clarity around why we plan to include the `census` package. Recently, we were looking at Meteor usage statistics and realized that given our existing telemetry we can only identify about 20% of apps using Meteor. That means a large fraction of our usage comes from apps that we know absolutely nothing about.

This is less than ideal as knowing who our users are and how they're using our platform helps us make informed decisions towards creating a great Meteor roadmap. The insight gained from better telemetry results in a roadmap that increases worldwide Meteor adoption (which helps everyone in the community - more contributors, more add-ons/services, more talent, more jobs). For instance, knowing what breakdown of apps are built by enterprises vs startups vs individuals allows us to tailor our learning materials, tutorials, features and commercial offerings accordingly. Knowing what scale folks use Meteor at (by collecting the `maxSessions` stat - https://github.com/meteor/meteor/pull/6469/files#diff-a7c6c405e95f7fbaeac0c4645b12c77eR27) informs us on how much engineering resources to allocate towards scaling/performance.

Ultimately, we're doing this to benefit everyone in the community by gathering objective data that informs where to spend our resources in the most impactful way. The difficult decision lies in where to strike the balance between the package being too intrusive but useful vs being friendly but useless.

@laosb I like your suggestion of explicitly notifying users that we've added the package.

packages/census/README.md
```diff
@@ -0,0 +1,23 @@
...
+- report.onFail(callback) - Registers a callback which will be invoked with the error on fail.
+- startSampling() - Starts sampling.
+- stopSampling() - Stops sampling.
+
+#### Config
+
+Cenus can be configured by the following enviroment variables:
+
+- APP_ID (required) - The application id.
+- APP_ID (required) - The application secret.
```
StorytellerCZ: Shouldn't this be `APP_SECRET`?

mquandalle:
Thank you very much for you answer @zol. As I believe others will be interested in this development as well I opened [a thread on the forum](https://forums.meteor.com/t/mdg-plan-for-a-tracking-package/19471) so that we can leave this PR for code review.

packages/census/utils.js
```diff
@@ -0,0 +1,32 @@
...
+      if (err && options.attempts--)
+        Utils.request(options, cb);
+      // If request has been succeeded or no more attempts left, invoke callback
+      else
+        cb(err, result);
+    }));
+  },
+
+  // Gets the ip address
+  ip() {
```
juho: Great way to get sued here MDG. Do NOT approve this PR.

mquandalle: Shouldn’t be a “headache for enterprise people” because you can remove all the tracking code with `meteor remove census`.

peterchoo: There is a chance someone who does not look at the forums or misses this issue will not know to do so.  I am against having this as an opt-out package, I do not feel that it is appropriate for MDG to add automatic statistic collection.  I feel this should be an opt-IN package, with only a notification post project creation to say this package exists, how to install it with `meteor add census`, an a link to the documentation/source.

PR #6472, opened 2016-03-13
Fix logic for showing --driver-package error

aldeed:
Previously the "No --driver-package specified for `meteor test`" error would never actually show because the logic was wrong. Now it will.

cc @tmeasday

### DIFF
--- a/packages/meteor/test_environment.js
+++ b/packages/meteor/test_environment.js
@@ -13,12 +13,8 @@ Meteor.isTest = !!TEST_METADATA.isTest;
 Meteor.isAppTest = !!TEST_METADATA.isAppTest;
 
 var testDriverPackageName = TEST_METADATA.driverPackage;
-if (testDriverPackageName) {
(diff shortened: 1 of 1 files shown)


tmeasday:
@aldeed are you sure this wasn't just a consequence of #6473

aldeed:
@tmeasday regardless if you just read the logic in this section, it doesn't make sense.  `if (typeof testDriverPackageName !== "string")` is within `if (testDriverPackageName)`, which would always be a string if it's set, so that error would never get thrown with the original logic.

tmeasday:
OK, sure but we don't want the else block at the end though---this code runs in all modes; the check on `testDriverPackageName` is how we know if we are in a test mode

aldeed:
OK, I see. The error for missing `--driver-package` option when in test mode is already thrown elsewhere. So yeah, this should still be merged since the error here wouldn't ever be thrown, but it's not high priority. I updated my branch to remove the `else`

PR #6488, opened 2016-03-15
Escape script tag in History.md

emgee3:
An unescaped `<script>` tag prevented viewing the full History.md file. Added backticks to restore visibility.

Before the fix:
screen shot 2016-03-14 at 9 49 50 pm


laosb:
Same problem, :+1:

zol:
Thanks!

PR #6494, opened 2016-03-15
A more developer friendly Meteor `underscore` dependency.

bySabi:
PR proposal for make Meteor `underscore` dependency more customizable for developers who
want update to latest `underscore` or `lodash` or use library npm module.

Probably more flames to: https://github.com/meteor/meteor/issues/1009

What change here?

All meteor packages depends on a forked outdated `underscore`. This fork can´t be update without broke core packages but it can be make a internal package and expose to users through a wrapper package.

With this PR:

* `underscore` package become `underscore-base` and namespace `_` goes to `__`.
Core packages `meteor`, `tinytest` and `modules` now depends on `underscore-base`.
* New `underscore` package is a wrapper of `underscore-base` and expose global `_`. All other packages depends on `underscore` remain without change.
* Former `underscore-test` was merged on new `underscore` package.

Advantages?

* Now developers can experiment a little, like update to latest `lodash` module.
Ex:

packages\underscore\packages.js
```
Package.describe({
  summary: "Collection of small helpers: _.map, _.each, ...",
  version: '1.0.6-rc.1'
});

Package.onUse(function(api) {
  api.use('modules');
  api.mainModule('underscore.js', ['client', 'server']);
});

```

packages\underscore\underscore.js
```
_ = require('lodash')._;
_.contains = _.includes;
global._ = _;
```

* For MDG is one step for future migration to `underscore` module.

Probably this PR would not be merged but the 'idea' is still valid. :-)

NOTE: I test changes locally and it working.


zol:
I think your use case is actually addressed by Meteor 1.3. Can you just `npm install` lodash or underscore at whatever version and use `import` or `require` to load it?

bySabi:
What to do if I want test core meteor packages with lodash?

zol:
Not sure what you mean by that.

bySabi:
I mean monkey patch meteor packages without broke all.
Update `underscore` or go to `lodash` for all packages but `meteor`, `tinytest` and `modules`.

zol:
Ah, got it. This is an edge case we're not looking to add support for.

PR #6545, opened 2016-03-20
Include a new "devtools" package in the Meteor skeleton

mquandalle:
There has been a lot of discussions in the JavaScript community about how hard it is to set up a project correctly in our constantly evolving environment. In this ecosystem Meteor has always stood as a developer friendly platform to build JavaScript applications, and I believe there is an opportunity to work more in this direction.

I propose to add a package called `"debugger"` (we could find a better name) that consists of a set of tools activated in development mode only. It currently only contains the [simple:dev-error-overlay](https://atmospherejs.com/simple/dev-error-overlay) package developed by @stubailo, but in the future it could include other tools for things like type checking, hot reloading, state inspectors, etc. as these tools emerge and mature as community packages.

### DIFF
--- a/packages/devtools/README.md
+++ b/packages/devtools/README.md
@@ -2,4 +2,5 @@
 
 A set of tools that improve the developer experience. These tools won’t be included in production builds of your app.
 
-1. [simple:dev-error-overlay](https://atmospherejs.com/simple/dev-error-overlay) - Display a nice overlay in your browser when your Meteor app has a server-side build error. You can also optionally display a desktop notification or play a sound!
+1. [`hot-code-push`](https://atmospherejs.com/meteor/hot-code-push) - Refresh the client automatically when the server has new code.
+2. [simple:dev-error-overlay](https://atmospherejs.com/simple/dev-error-overlay) - Display a nice overlay in your browser when your Meteor app has a server-side build error. You can also optionally display a desktop notification or play a sound!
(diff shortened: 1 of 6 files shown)


stubailo:
I'd say it should either be called "developer-experience" or maybe just be all the packages individually. But I agree having these by default could be a nice advantage and show that meteor cares about developer experience more than almost anyone else.

laosb:
Sounds sweet. But is it quite right to imply a **community** package in a **core** package?

stubailo:
> is it quite right to imply a community package in a core package

I think we should start doing this more - there are a lot of community packages that are pretty essential to the Meteor experience, especially in the realm of testing. Don't have an opinion about my particular package, though.

mquandalle:
I considered the name “developer-experience” but that sounds confusing given the parallel with “mobile-experience” which is a very different kind of package, runtime instead of build-time. Also I think it would be a bit strange to “remove developer-experience” if for some reason someone want to remove this package :-)

I propose the name “devtools”.

stubailo:
Hmm, personally I think basically everyone should have `hot-code-push`, while the dev error overlay is a lot less universal.

Plus, hot code push actually runs in production as well!

I'd rather just add each package individually.

laosb:
Maybe we should just mention these great packages in our Guide.
Or if we have this package, `devtool` is actually a great name.

stubailo:
Yeah I'd be in favor of a guide article about development tools like debugging, version control, etc.

tmeasday:
I think this should be a community package @mquandalle. (The guide article would be awesome!)

PR #6581, opened 2016-03-24
Update History.md

emgee3:
Escape `<SELECT>` tag to fix improperly rendered markdown.


emgee3:
Updated against devel branch at https://github.com/meteor/meteor/pull/6607

PR #6606, opened 2016-03-28
Fixed typo: remove erroneous comma from skeleton

serkandurusoy:
The skeleton app created by `meteor --create` creates a skeleton app, whose `client/main.js` file has erroneous commas after their first and only helper and event map objects.

This does not break anything, but since this is the first thing a new Meteor developer will see, we would not want to leave an unprofessional impression with typos.

CC: @platform

### DIFF
--- a/tools/static-assets/skel/client/main.js
+++ b/tools/static-assets/skel/client/main.js
@@ -11,12 +11,12 @@ Template.hello.onCreated(function helloOnCreated() {
 Template.hello.helpers({
   counter() {
     return Template.instance().counter.get();
-  },
+  }
 });
 
 Template.hello.events({
   'click button'(event, instance) {
     // increment the counter when button is clicked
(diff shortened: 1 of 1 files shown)


serkandurusoy:
@benjamn I guess this one is within your territory. This is a very light fix - although I believe it to be philosophically important - and it is my first PR on meteor so I hope this does not get blocked by some policy nuance.

serkandurusoy:
What does *closed* mean? Is this rejected? Fixed elsewhere? Have I done something wrong? Should I be encouraged/disscouraged for further pr's?

benjamn:
Sorry, that was an accidental consequence of deleting the `release-1.3` branch, now that Meteor 1.3 has been released. Can you re-send this PR against the `devel` branch?

serkandurusoy:
Created https://github.com/meteor/meteor/pull/6615

PR #6607, opened 2016-03-29
Escape HTML tags in History.md

emgee3:
Unescaped HTML tags cause rendering errors in History.md when viewed on Github. This patch escapes them all.


stubailo:
Thanks!

Issue #376, opened 2012-10-08
Allow trailing comma in JSON strings


dandv:
I know the JSON spec forbids a trailing comma in an array, but for better interoperability, it would help if methods like `Meteor.http.get` didn't choke on a final comma in a JSON array.

Use case: consuming third party APIs without having to chop off the comma from the end of `result.content` and parsing it as JSON again (CPU cycles will be consumed on the first attempt to parse the JSON, up to the penultimate character, the comma).

PR #6615, opened 2016-03-29
Fixed typo: remove erroneous comma from skeleton

serkandurusoy:
The skeleton app created by meteor --create creates a skeleton app, whose client/main.js file has erroneous commas after their first and only helper and event map objects.

This does not break anything, but since this is the first thing a new Meteor developer will see, we would not want to leave an unprofessional impression with typos.

CC: @benjamn

### DIFF
--- a/tools/static-assets/skel/client/main.js
+++ b/tools/static-assets/skel/client/main.js
@@ -11,12 +11,12 @@ Template.hello.onCreated(function helloOnCreated() {
 Template.hello.helpers({
   counter() {
     return Template.instance().counter.get();
-  },
+  }
 });
 
 Template.hello.events({
   'click button'(event, instance) {
     // increment the counter when button is clicked
     instance.counter.set(instance.counter.get() + 1);
-  },
+  }
 });


serkandurusoy:
@benjamn I don't know why tests have failed, I believe devel already had them failing.

Anyway, I did some research and it seems the "dangling comma" is actually spec-compliant and is just a stylistic preference.

I still think they should not be there, becasue it is a "perceived problem" and the skeleton app is in my opinion a "marketing tool".

lucfranken:
It can be a disadvantage if you use git for diff's. If you add a second helper it will show 2 changes lines. Because you then add the comma. That's why this might be the spec I suspect. And it does actually make sense in that scenario. 

Only the JSON spec does not allow the trailing comma which make it a bit inconsistent. That has been discussed here: https://github.com/meteor/meteor/issues/376

stubailo:
@serkandurusoy we recommend trailing commas in the Meteor Guide, via suggesting the AirBnB style guide. So expect to see trailing commas everywhere.

stubailo:
Here are directions: http://guide.meteor.com/code-style.html#eslint-installing

I think that gives you everything you need to install and run a linter!

serkandurusoy:
Ha! I guess I need to check back on guide changes more often! Thank you for patiently pointing me in the right direction.

Or perhaps, the guide [could benefit](https://github.com/meteor/guide/issues/305) from a `changelog.md`.

stubailo:
Ah, that's actually how to "Migrating to 1.3" article started out. But I agree it could be useful to have a changelog... What do you think @tmeasday?

Issue #2176, opened 2014-05-23
Meteor.defer() not in docs?


michael-mccrickard:
***1 Upvote*** Checking the page:  docs.meteor.com and I am unable to find the term 'defer' anywhere.  Someone on stackoverflow just helped me solve a fundamental problem that I was having with iron-router and meteor by using Meteor.defer().  I had never heard of this function until then!

KrishnaPG:
Which is the recommended one? `Meteor.defer()` or `Meteor.setTimeout()`.

Looking for a way to handle `DOM ready / document loaded` events (Required by some external JS files)

glasser:
We now do track doc [feature requests](https://github.com/meteor/meteor/wiki/Contributing-to-Meteor#feature-requests) so I'll reopen this.  Pull requests to improve our docs are always welcome!

@KrishnaPG I don't think either of those functions responds to DOM ready.  You can use `Meteor.startup` for that though!

keyscores:
Seems like this is a similar issue as this: https://forums.meteor.com/t/template-layout-rendered-cannot-initialize-jquery-components-for-sub-templates/2706/2

How should we use .startup or .defer in this case

tcastelli:
+1 would be interested to know how to use startup in a similar context. Should we check the loaded template/url in a meteor startup function and then start the things?

fightingtheboss:
Although I'm aware of `Meteor.defer`, I actually came across a situation where this would have easily solved my issue (and ultimately did!) and it would've definitely jogged my memory if it were part of the docs. I can try to make the time to contribute a pull request adding it to the docs if they're being accepted for documentation.

michael-mccrickard:
I use Meteor.defer() quite regularly.  Anytime I have an instruction in my code that doesn’t seem to execute and then I run that same instruction in the console and it works, I wrap it in my code with Meteor.defer() and that usually works.  Typically, it seems to be cases where something in the DOM that I’m depending on has not yet rendered.  You would think that “after” hook in iron-router would allow you to avoid this, but I find that method unreliable.

rclai:
It has been pretty problematic trying to get third party libraries to initialize onRendered mainly on giant templates because even though onRendered runs, the actual target element hasn't completely rendered, and the initializing script runs prematurely. The workaround is to have those target elements inside a separate template with its own onRendered callback and do the initializing there.

woniesong92:
[...]

Issue #4442, opened 2015-05-23
Meteor.defer not covered in docs


ffxsam:
It would be great to see this documented. It's a powerful feature and not mentioned anywhere.

rgoomar:
Oh wow, never noticed that. It also isn't documented in comments either (JSDoc style).
https://github.com/meteor/meteor/blob/master/packages/meteor/timers.js#L67

ekatek:
Agreed. Thanks for catching this!

mitjajez:
This is duplicate of #2176.

stubailo:
Just added it! It will be deployed next time the docs are.

ffxsam:
This issue is so old, I forgot what `Meteor.defer` does! 😆  I guess I'll wait and see when the docs are updated.

PR #6629, opened 2016-03-30
Document Meteor.defer function

skirunman:
This commit resolves issues #2176 and #4442. Many Meteor tutorials and developers use the Meteor.defer function call on both the client and the server, but it is not documented. In addition, this function is even used in the latest [Meteor 1.3 todos example](https://github.com/meteor/todos/blob/9f515946f30331451af07960df845f43448aff3b/imports/ui/components/lists-show.js#L61) and was recently discussed on [Meteor Forums](https://forums.meteor.com/t/todos-app-madness/20239/6). 

I added this to the Meteor Core section right after Meteor.wrapAsync as it flowed best for me in this section. I did not put it in the Timers section as Meteor.defer is not really a timer call per se and in fact can have very slightly different behavior from `Meteor.setTimeout(func, 0)`.

This could be flushed out even further with examples of how/why to use this call on the server and the client, but that is probably best left for the Meteor Guide.

### DIFF
--- a/docs/client/full-api/api/core.md
+++ b/docs/client/full-api/api/core.md
@@ -40,7 +40,7 @@ followed by your application code.
 
 {{> autoApiBox "Meteor.wrapAsync"}}
 
-{{> auotApiBox "Meteor.defer"}}
+{{> autoApiBox "Meteor.defer"}}
 
 {{> autoApiBox "Meteor.absoluteUrl"}}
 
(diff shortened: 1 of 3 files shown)


skirunman:
There is one failure in the CI test, not sure why.

skirunman:
@stubailo Can we merge this. May want mention Meteor.defer in Meteor Guide section on reactive data as well.

stubailo:
Merged as one commit: 4afbabd0dc43ff57b2c262111923609ab35a85a4

Thanks!

PR #6633, opened 2016-03-30
Mention Meteor Guide in README.md.

laosb:
Looks we even didn't metion the great Guide!


laosb:
Seems someone had added it now.

PR #6658, opened 2016-03-31
Remove duplicate call to `DDPCommon.stringifyDDP`

awwx:
This code simplification in ddp-server removes the duplicate call to
`DDPCommon.stringifyDDP` when sending the `connected` message.

The change makes the code easier to understand: it clarifies that
there's nothing magical about the `connnected` message: it is sent in
the same way as the other DDP messages.

It is almost a pure refactoring: if the undocumented `_printSentDDP`
option is false (the default), there's no change in the behavior of
the code.

Previously, if `Meteor._printSentDDP` was set, the server would
display other sent messages but not the `connected` message.  With
this change, when the option is set then all sent DDP messages will be
displayed.

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -294,7 +294,7 @@ var Session = function (server, version, socket, options) {
     httpHeaders: self.socket.headers
   };
 
-  self.send({msg: 'connected', session: self.id});
+  self.send({ msg: 'connected', session: self.id });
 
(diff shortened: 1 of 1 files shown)


zol:
Hey @awwx . I'd happily merge this PR. Could you please rebase against the latest `devel` and ensure the tests pass (we fixed up CI recently).

zol:
Closing this off as I created a rebased version here: https://github.com/meteor/meteor/pull/7114

Issue #6660, opened 2016-03-31
multiline strings are broken after linker builds modules.js


martintreurnicht:
Version : meteor 1.3

When including [openchain](https://www.npmjs.com/package/openchain), i'm getting `Uncaught SyntaxError: Unexpected token ILLEGAL` error

When i look at the javascript i can see that it's including comments next to the line, which is breaking the multiline string (see screenshot). This issue seem to have been resolved by #1350 back in 2013, but it has come back somehow. I had a look at commit https://github.com/meteor/meteor/commit/bfc7f22 and that change hasn't regressed, so i'm not sure why this is happening

Here is some easily reproducible steps:
1. `meteor create test && cd test`
2. `meteor npm install openchain`
3 import it on main.js `import 'openchain'`

<img width="1612" alt="screen shot 2016-03-31 at 11 18 26 am" src="https://cloud.githubusercontent.com/assets/1970728/14186265/5c80a56c-f732-11e5-90e1-4b2a00026a7c.png">

benjamn:
Just tracked this down. That file's line endings are `\r\n`, and since we were [splitting the lines on `\n`](https://github.com/meteor/meteor/blob/8248d86fb2aeb27aac30d0ca0682e0ab8008320c/tools/isobuild/linker.js#L642), lines that appeared to end with `"\\"` actually ended with a `"\\\r"`, which confused [this code](https://github.com/meteor/meteor/blob/8248d86fb2aeb27aac30d0ca0682e0ab8008320c/tools/isobuild/linker.js#L650).

benjamn:
Try `meteor update --release 1.3.1-rc.3` if you have a chance!

martintreurnicht:
@benjamn Awesome! Seems to be working. Thanks for fixing this so quickly

PR #6663, opened 2016-03-31
Fix multiline strings breaking if there is some kind of whitespace after

martintreurnicht:
Fixes issue #6660


zol:
Fixed in 1.3.1-rc.3

PR #6668, opened 2016-04-01
Update Blaze.remove docs.

trusktr:
It's helpful to know that this also destroys the template instance associated with the view being removed. Could possibly be worded better.


stubailo:
that's legit! will get deployed next time we deploy the docs.

Thanks!

Issue #6684, opened 2016-04-02
windows installer text is obsolete and misleading - "our free service"


adaptabi:
![image](https://cloud.githubusercontent.com/assets/485535/14225318/642bd886-f8c7-11e5-9332-93d106692e5f.png)

Unfortunately, the free service was hunted down and now only payed galaxy service is available!

Does anyone still reads these texts? :)

helfer:
Thanks for pointing it out, we have  a pull request that fixes it: https://github.com/meteor/meteor/pull/6685

It will be included once we get around to recompiling the windows installer.

stubailo:
(To clarify, recompiling the installer is an arduous process that involves physical certificate signing keys, very specific visual studio builds, etc, so sorry that it takes a bit longer than it should)

hwillson:
This was completed - thanks for posting this originally!

hwillson:
Ooops - turns out this wasn't fully completed. The submitted changes are still pending a re-compile (https://github.com/meteor/meteor/pull/6685#issuecomment-205147754). Re-opening!

hwillson:
I believe a re-compile has happened since https://github.com/meteor/meteor/issues/6684#issuecomment-276139992, so I'll close this off (again 🙂). Thanks!

abernix:
Actually, this still hasn't happened. 😞

hwillson:
The old Windows installer has been replaced by [`choco install meteor`](https://www.meteor.com/install) in Meteor 1.6. Closing this off (for good this time! 🙂).

PR #6685, opened 2016-04-02
Update text in Windows installer (Fixing #6684)

laosb:
#6684 
I replaced "free service" with "publishing Atmosphere packages", because it might be more useful for developers instead of just advertising Galaxy.


stubailo:
Ah, this referred to the recently shut down `meteor deploy` service. I'm going to keep this PR open because it requires recompiling the installer, which we do very very rarely - if I just merge it nobody will notice the change. But it's definitely worth updating that text soon.

@benjamn and @zol, I can help get this done whenever you think it's a priority, you'll need my help to use the code signing dongle.

laosb:
:) Sorry for my unclear issue title

PR #6698, opened 2016-04-04
Remove _.once and add cache to getPool

### LINKED ISSUE #4323: Email package should allow changing SMTP settings
In my application SMTP is configured by the user in the UI. As a result, they also need to be able to update the configuration from the UI, but today this is not possible. The email package creates a connection pool only once, regardless of whether the `MAIL_URL` environment variable is later changed. In email.js we have:
```
var getPool = _.once(function () {
  // We delay this check until the first call to Email.send, in case someone
  // set process.env.MAIL_URL in startup code.
  var url = process.env.MAIL_URL;
  if (! url)
    return null;
  return makePool(url);
});
```

It would be great if there were some way to update the SMTP configuration while the application is running, either by checking whether the environment variable has changed before sending an email (and creating a new pool if it has), or exposing some hook so the application can manually let the email package know the environment variable has changed.

marceloschmidt:
***1 Upvote*** Fix #4323
Using _.once gives you no option for "clearing the cache" of the ran function.
Using _.memoize, we can pass in the value of process.env.MAIL_URL, so once 
it changes, the function is allowed to be run again, caching the return value
for the current value of process.env.MAIL_URL.

### DIFF
--- a/packages/email/email.js
+++ b/packages/email/email.js
@@ -41,14 +41,17 @@ var makePool = function (mailUrlString) {
   return pool;
 };
 
-var getPool = _.once(function () {
+var getPool = function() {
   // We delay this check until the first call to Email.send, in case someone
-  // set process.env.MAIL_URL in startup code.
+  // set process.env.MAIL_URL in startup code. Then we store in a cache until
+  // process.env.MAIL_URL changes.
   var url = process.env.MAIL_URL;
-  if (! url)
-    return null;
-  return makePool(url);
-});
+  if (this.cacheKey === undefined || this.cacheKey !== url) {
+    this.cacheKey = url;
+    this.cache = url ? makePool(url) : null;
+  }
+  return this.cache;
+}
 
 var next_devmode_mail_id = 0;
 var output_stream = process.stdout;


osv:
+1

martijnwalraven:
@marceloschmidt: Hmmm, but wouldn't this leave you with multiple SMTP client pools if the `MAIL_URL` changes? Looking at the code for [_.memoize()](https://github.com/jashkenas/underscore/blob/master/underscore.js#L781-#L791), it is designed to store multiple values and doesn't actually clear the cache if a new one is computed. The code is pretty simple however, so you could probably implement a similar function that would only store one value at a time.

marceloschmidt:
Yes, I hadn't realized this. Thanks for pointing it out, @martijnwalraven . I'll create a function that deals with a single value but is also 'clearable'.

marceloschmidt:
@martijnwalraven can you take a look at the modified function? Thanks!

martijnwalraven:
This looks reasonable to me, although it could use at least a few comments. 

Since this is the only place `memoizeOne()` is used, it may also make sense to just inline the behavior in `getPool()`. That would also get rid of some duplication (accessing `process.env.MAIL_URL` twice) and may make the code more readable. But that is a judgement call, and going with `memoizeOne()` may also be fine as long as it is documented.

marceloschmidt:
@martijnwalraven that was a great suggestion (I should've thought about that). Can you check out current code and see if it's fit? Thanks!

martijnwalraven:
Looks good to me!

martijnwalraven:
Maybe just squash the commits and change the pull request title?

marceloschmidt:
Done

martijnwalraven:
Sorry to be such a pain, but the commit message still contains the messages from the previous commits. If you can change that, I'll merge it immediately!

marceloschmidt:
Not a problem. I had never used squash commits before and I probably messed that up on the rebase commit. I'll do it in a few minutes and will let you know. Thanks for the patience :)

marceloschmidt:
Done

martijnwalraven:
Thanks, merged!

Issue #6563, opened 2016-03-22
Meteor Core jQuery dependency when using static-html


faburem:
***1 Upvote*** When I create a new project where I remove blaze and jQuery in favor of static-html, jQuery is still loaded. This happens with Meteor 1.2.1 and also with 1.3-rc.5. Here is what I did and my test results:
```
meteor create static-html
meteor remove blaze-html-templates insecure autopublish jquery
meteor add static-html
```
![screen shot 2016-03-22 at 09 54 05](https://cloud.githubusercontent.com/assets/11456790/13946611/eeb537fa-f015-11e5-9457-9c246af489d2.png)
When I run this regularly:
![screen shot 2016-03-22 at 09 48 00](https://cloud.githubusercontent.com/assets/11456790/13946591/d6affff0-f015-11e5-9705-52abfdf6d15c.png)
and the same with --production:
![screen shot 2016-03-22 at 09 53 24](https://cloud.githubusercontent.com/assets/11456790/13946681/57b7f922-f016-11e5-88d1-27a6a35f77a3.png)

faburem:
I traced this down to the following packages:

- **check**: easy to resolve, is there any reason why this currently depends on jQuery (it didn't used to)?
- **boilerplate-generator**: this will be harder to resolve, it currently has a hard dependency on Blaze which in turn depends on jQuery. I definitely think this should be decoupled now that there are many alternatives to Blaze available and actively used.

skirunman:
Looks like part of jQuery lib was added to do object check in this this last commit https://github.com/meteor/meteor/commit/22e5a2611d7fd78abb4aeaddc2615bcb33adc0a9#diff-db148119d1602c9b3d16d4d020e4e7ff

Not sure why there is a strong dependency on jquery package here `api.use('jquery', 'client');`, but I agree this needs to be fixed.

@mitar what are your thoughts as this was your commit?

skirunman:
> it currently has a hard dependency on Blaze 

I did not see the Blaze dependency, but there is a Spacebars one.

```
api.use(['underscore',  'spacebars-compiler',
           'spacebars', 'htmljs', 'ui'], 'server');
```
https://github.com/meteor/meteor/blob/devel/packages/boilerplate-generator/package.js

faburem:
yeah its a bit hidden behind the ui dependency (which btw. is deprecated and resolves to Blaze)

mitar:
[...]

PR #6710, opened 2016-04-05
jquery dependency fix for check package

skirunman:
This is is meant to fix the jquery dependency in the check package for the isPlainObject function by using a version of isPlainObject from lodash, issue #6563 

I was not able to get tests to run in my environment and this should be checked by someone with more experience on Meteor core. I hope this can at least help get this fix implemented soon.


skirunman:
I just want to reiterate this is a work in progress and needs to be tested.

packages/check/match.js
```diff
@@ -3,6 +3,9 @@
 // Things we explicitly do NOT support:
 //    - heterogenous arrays
 
+//XXX: import from main lodash library if Meteor updates to use lodash instead of underscore
+const isPlainObject = Npm.require('lodash.isplainobject');
```
mitar: This works also on client? This is new in 1.3? Interesting.

skirunman: I think you are right that it does not work on client.

packages/check/package.js
```diff
@@ -4,15 +4,17 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
-  api.use(['underscore', 'ejson'], ['client', 'server']);
-  api.use('jquery', 'client');
+  api.use(['underscore', 'modules', 'ejson'], ['client', 'server']);
```
stubailo: I think you want to actually depend on `ecmascript` instead of `modules` here, especially if you are using `const` in the code.

skirunman: I tried that, but got an error with a circular reference that ecmascript package depended on check. I'll just change `const` to `var`.

skirunman:
Closing as solution implemented as shim function.

Issue #6666, opened 2016-04-01
ROOT_URL backstabbed by automatic `meteor update`


skaag:
When running a meteor project for the first time in a fresh environment, if the `~/.meteor` folder does not exist, Meteor will download it from Amazon's cloudfront. That's great, except the `ROOT_URL` environment variable isn't saved/restored, so once the files/libraries are fetched, the project will fail to run.

This scenario is where Meteor is installed by the root user, globally. Then, a non-root user tries to run a meteor project (in which case, `~/.meteor` would obviously be missing).

Example:
Before the update/install" `ROOT_URL = http://myapp.local:3000/`
After the update/install: `ROOT_URL = https://d3sqy0vbqsdhku.cloudfront.net/packages-bootstrap/1.2.1`

To avoid this issue, the Meteor updater needs to either save & restore the previous value of `ROOT_URL`, or it should use a different environment variable entirely.

helfer:
The .meteor folder actually contains files that your app needs, such as which packages are installed. If you're moving your app, you should copy the .meteor directory as well. You can ignore the .meteor/local directory however. If in doubt, refer to the .gitignore file.

stubailo:
I'd be happy to accept a PR to this script to use a different variable than `ROOT_URL`:

https://github.com/meteor/meteor/blob/devel/scripts/admin/launch-meteor

skaag:
Thank you for the opportunity! I just created a PR. Signed the CLA, too. I guess it's waiting on a CI test now.

laosb:
As the pr was merged, closing.

PR #6711, opened 2016-04-05
Fixed issue #6666 + Removed progress bar animation when no tty is pre…

skaag:
This PR contains two very simple fixes:

1. It allows Meteor projects to run smoothly even if an updated version of meteor was fetched automatically (Issue #6666)
2. It hides the `curl` progress bar when not running in a terminal. This is very useful when meteor installs are automated (via ansible, or within vagrant environment, etc).


stubailo:
Looks good to me! Will come out in the next release.

PR #6731, opened 2016-04-07
Fix bad link in modules package documentation

skirunman:
Link should be to Meteor Guide, not non-existent documenation.


laosb:
I think there are more dead links in the document exists, aren't there?

stubailo:
Great catch. I'm sure this isn't the only one.

Issue #6712, opened 2016-04-06
[1.3] `check` on template instances creates infinite recursion


Zodiase:
You could try this in the `todos` demo app.

```JavaScript
Template.App_body.onCreated(function appBodyOnCreated() {
  // This line creates an infinite loop in Meteor 1.3.
  check(this, Match.Optional(Blaze.TemplateInstance));
});
```

In one of my packages I have to check if an optional parameter is an instance of `Blaze.TemplateInstance` and this is what I've been doing. I didn't encounter this error before Meteor 1.3.

Zodiase:
Error in console:
![image](https://cloud.githubusercontent.com/assets/1815174/14302093/e61885c2-fb62-11e5-857c-a2e455dc78a2.png)

abernix:
I don't have a fix, but I thought I'd put what I did find here.

This is happening only with `Match.Optional` (maybe with `Match.Maybe`, but I didn't check) and is occurring because of `EJSON`, not because of `check`.  It's happening when trying to serialize the `Blaze.TemplateInstance` with `EJSON.stringify`.

This is the line throwing the error.  If you remove the `EJSON.stringify` from the error message, it works fine...
https://github.com/meteor/meteor/blob/22e5a2611d7fd78abb4aeaddc2615bcb33adc0a9/packages/check/match.js#L169

Zodiase:
@abernix 
I know the actual infinite recursion is caused by `EJSON.stringify`ing `Blaze.TemplateInstance`, but I believe the cause is still within `check`. Reason being: `check(this, Blaze.TemplateInstance)` works perfectly fine (except it doesn't allow `null`), which means `check` doesn't always have to call `EJSON.stringify`. So it is only logical to me that when using `Match.Optional`, after ruling out the "optional" part of the logic, it will be the exact same scenario as not having the "optional".

abernix:
I would argue that the infinite recursion in `EJSON.stingify` should be fixed too, but it should be a separate issue.   That being said, I have found what created the problem in `check`.  It has to do with the way that `Match.Maybe` was added in 1.3 (via #6220). 

I will try to get a PR open today, but I want to write a few more tests first because there was completely unintended change to `Match.Optional` that went unnoticed (it accepts `null` now! – so identical to the new `.Maybe`!).

abernix:
[...]

PR #6736, opened 2016-04-07
Fix `Match.Optional` to work as it did previously in Meteor 1.2

### LINKED ISSUE #6735: [1.3] `check`'s Match.Optional now allows "null" (contrary to plan)
As found during research for the problem in  #6712, `Match.Optional` now matches `null` when it had not before 1.3.  The (ultimate) decision in Issue #3876 was to keep `Match.Optional` the same (allow `undefined`) and introduce a new `Match.Maybe` which would match `null` and `undefined` in order to **NOT** do this:

> silently change the (`Match.Optional`) behavior because it could introduce undiscoverable security holes into people's apps

Unfortunately, the way that that #6220 was done, it literally did exactly that, and there was no Match test in place to make sure that `null` did not pass `Match.Optional`.

I have an incoming PR to fix this.  This Issue is just for the record.

abernix:
`Match.Optional` is still only supposed to "pass" if the value is `null`
or the specified type.  The new `Match.Maybe` allows `undefined` or
`null` in addition to the specified types. `Match.Optional` is on the
track toward deprecation, however to not break existing code it was
*supposed* to stay working the same as before. (per #3876).

There weren't tests in place to make sure that `Match.Optional` kept
working the same, and the code didn't actually make it keep working the
same.  Hopefully extra tests will make this better.

`.Maybe` has some additional bugs, but should be addressed separately
(see #6712)

Fixes #6735

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -115,7 +115,9 @@ var Optional = function (pattern) {
   this.pattern = pattern;
 };
 
-var Maybe = Optional;
+var Maybe = function (pattern) {
+  this.pattern = pattern;
+};
 
 var OneOf = function (choices) {
   if (_.isEmpty(choices))
--- a/packages/check/match_test.js
+++ b/packages/check/match_test.js
@@ -73,6 +73,7 @@ Tinytest.add("check - check", function (test) {
           }
         }));
       }
+      if ( type !== null ) fails(null, Match.Optional(type)); // Optional doesn't allow null, but does match on null type
       fails(pair[0], [type]);
       fails(pair[0], Object);
     });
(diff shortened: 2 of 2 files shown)


packages/check/match.js
```diff
@@ -115,7 +115,9 @@ var Optional = function (pattern) {
   this.pattern = pattern;
 };
 
-var Maybe = Optional;
+var Maybe = function (pattern) {
+  this.pattern = pattern;
+};
```
abernix: Line 256 below is where the real problem happened since it's an `elseif` that never got reached anymore since both `.Maybe` and `.Optional` were the same object.

abernix:
^^ Updated PR/commit reference to dyslexic "other" issue number I had in my commit msg.  Should be #6712.  Silly, Abernix.

PR #6743, opened 2016-04-07
Fix for `check` max-recursion when called on circular references

### LINKED ISSUE #4587: check fails when with circular object and null

### LINKED ISSUE #6712: [1.3] `check` on template instances creates infinite recursion

abernix:
Fix for `check` max-recursion when called on circular references. This is a stop-gap fix for a bigger issue with `EJSON.stringify` working on circular objects.  #4778 *apparently* fixes most *other* `EJSON` methods but it's not even clear it if it would fix the `.stringify` problem properly (after all, `JSON.stringify` doesn't either).  Also, that PR is over a year old.

This fix takes the approach of trying to fix the calling of `EJSON.stringify` with an object that it can't work with.  It's doing this by trying the native `JSON.stringify` to see if it would work.  If that returns a `TypeError`, then it just dumbs-down the object to the `typeof`.

Also added a lot of tests

I can't say it's the absolute best, but it *will* take care of:
* Fix #6712
* Fix #4587

Depends on #6736 (Match.Optional working properly) [update: it's merged]

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -138,12 +138,40 @@ var ObjectWithValues = function (pattern) {
   this.pattern = pattern;
 };
 
+var stringForErrorMessage = function (value, options) {
+  options = options || {};
+
+  if ( value === null ) return "null";
+
+  if ( options.onlyShowType ) {
+    return typeof value;
+  }
+
+  // Your average non-object things.  Saves from doing the try/catch below for.
+  if ( typeof value !== "object" ) {
+    return EJSON.stringify(value)
+  }
+
+  try {
+    // Find objects with circular references since EJSON doesn't support them yet (Issue #4778 + Unaccepted PR)
+    // If the native stringify is going to choke, EJSON.stringify is going to choke too.
+    JSON.stringify(value);
+  } catch (stringifyError) {
+    if ( stringifyError.name === "TypeError" ) {
+      return typeof value;
+    }
+  }
+
+  return EJSON.stringify(value);
+};
+
 var typeofChecks = [
   [String, "string"],
   [Number, "number"],
   [Boolean, "boolean"],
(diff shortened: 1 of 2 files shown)


benjamn:
Have you considered using Node's `require("util").inspect(value)`?

abernix:
`util` would be ideal.  But because this is also a client-facing library, I was concerned about:

1. Loading it client-side in addition to `EJSON.stringify`, which is already "there", *just* to gain the benefits of a nicer error message (which is all that happens here).  After all, `util` still just does a try/catch with `JSON.stringify`.
2. Potentially changing the existing error messages by using a different library to serialize them.  Though arguably, implementations should be using `Match.test` if they care about doing anything upon failures, I guess some might be try/catching.  Usage of `check` is so extensive within the community, especially within methods and publications, that I just really erred on the side of caution here.

Really, `EJSON` needs the fix here as it will still fail, and `util` is might be the right way to do that, however, I'm super unfamiliar with the reach of `EJSON` and what could be affected.

All in all, if 1 & 2 above aren't a concern, then I can update PR to use `util` and remove `EJSON.stringify` entirely from the `check` package.

Issue #5560, opened 2015-10-28
DDPRateLimiter is not defined


brianmingus:
On the client: Uncaught ReferenceError: DDPRateLimiter is not defined
On the server: ReferenceError: DDPRateLimiter is not defined

```coffeescript
addCommentRule =
  userId: (userId) ->
    type: 'method'
    name: 'getessay'

DDPRateLimiter.addRule addCommentRule, 1, 3000
DDPRateLimiter.setErrorMessage "Please wait three seconds"
```

brianmingus:
I don't believe the official documentation mentions that in order to use the rate limiter you must install a special ddp-rate-limiter package.

stubailo:
Yes, we don't do a good job listing in the docs when you need to add a certain package. For example, this is also the case with `ReactiveVar`.

abernix:
Moving this over to the new `docs` repo at meteor/docs#14 now that the [new docs are up](http://docs.meteor.com/).  I will submit a PR for it over there!  Closing.

PR #6751, opened 2016-04-08
Notify developers in the docs to add packages first. (Fixing #5560)

laosb:
#5560

Also did it for `check`.


abernix:
`check` was already fixed in the new docs and I put an issue/PR in meteor/docs#15 (+14) for `ddp-rate-limiter`.  Closing! :)

PR #6763, opened 2016-04-09
Hint adding packages when "Cannot find module 'meteor/xxx'"

laosb:
When a user encounter `Cannot find module 'meteor/xxx`, most possibly he didn't add the package but used it. Let's give him a hint to add it.


laosb:
@stubailo It's a common mistake. Having this hint could be useful for beginners.

stubailo:
@benjamn could be a simple thing to improve!

tmeasday:
Can you rebase this to be off of current `devel`? I think this branch is off an earlier commit with failing self tests. It should be as easy as (assuming you have a remote called upstream pointing here):

```
# optionally

git fetch upstream
git rebase upstream/devel
git push --force
```

laosb:
`Update Branch`  should did the same thing.

benjamn:
We could also consider removing [this code](https://github.com/meteor/meteor/blob/fb3b3c9cbc5764cccfc63e63a8b26b6437fbab33/tools/isobuild/compiler-plugin.js#L979-L982) so that the user would get a warning at build time, before hitting this runtime error.

laosb:
Yeah that would be nice.

tmeasday:
Thanks @laosb

PR #6773, opened 2016-04-10
Update .eslintrc - Remove last comma

zanot:
I followed this guide about [Meteor code style](http://guide.meteor.com/code-style.html), which has a link to [Set up Sublime Text for ES2015](http://info.meteor.com/blog/set-up-sublime-text-for-meteor-es6-es2015-and-jsx-syntax-and-linting).

In the step 4, we can copy ESLint configuration used by Meteor, but ESLint cannot read this file, it generates an error in the status bar : "cannot read config file" because of the last comma.

I use Sublime Text 3 build 3103.


tmeasday:
This config definitely works in the Meteor code base.

PR #6817, opened 2016-04-14
Expose holdCount within launch-screen package

markoshust:
It'd be a good idea to expose holdCount within this package, so we can tell if LaunchScreen is currently holding. This would be useful in situations where the `reload-on-resume` package is receiving a hot code push/update, and the screen is currently holding. Rather than showing a loading/spinner page, since the splash screen is still being shown, we can just immediately reload the page, preventing an additional loading/spinner page from being shown to expedite the waiting process.

### DIFF
--- a/packages/launch-screen/mobile-launch-screen.js
+++ b/packages/launch-screen/mobile-launch-screen.js
@@ -36,6 +36,7 @@ LaunchScreen = {
 
     // Returns a launch screen handle with a release method
     return {
+      holdCount: holdCount,
       release: release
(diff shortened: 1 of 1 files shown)


stubailo:
I think this would be better implemented as `LaunchScreen.getHoldCount` - it doesn't make sense to me that you can only get that information via `hold()`.

tmeasday:
I agree with @stubailo here @markoshust -- could you please resubmit this PR if you are still interested?

PR #6819, opened 2016-04-14
Remove handle.stop() after update

jamielob:
Removed handle.stop() so that live query continues to run after the first code push, and newer versions continue to download.

Important when delaying reload until after a resume, otherwise it takes multiple resumes to get to the latest bundle.

### DIFF
--- a/packages/autoupdate/autoupdate_cordova.js
+++ b/packages/autoupdate/autoupdate_cordova.js
@@ -48,7 +48,6 @@ Autoupdate._retrySubscription = function() {
         var checkNewVersionDocument = function(doc) {
           var self = this;
           if (doc.version !== autoupdateVersionCordova) {
-            handle && handle.stop();
             newVersionAvailable();
           }
         };


jamielob:
@martijnwalraven RE: our discussion.

jamielob:
Thanks @martijnwalraven! Just took a second look at this - should we also remove this now as it looks like it might be redundant:

https://github.com/jamielob/meteor/blob/7190faedb22abc9e18f51b7d0291c366630caa9b/packages/autoupdate/autoupdate_cordova.js#L55-L58

martijnwalraven:
That's certainly not redundant! Without it, we wouldn't get notified of new versions at all.

jamielob:
Sorry @martijnwalraven, I was in a rush before - I simply meant the handle was redundant now. It could just simply be:

```
//Watch for new versions
ClientVersions.find({_id: 'version-cordova'}).observe({
          added: checkNewVersionDocument,
          changed: checkNewVersionDocument
        });
```

Unless I'm reading it wrong or there is a semantic reason to keep the handle var in there.

PR #6822, opened 2016-04-15
New hexo based docs

tmeasday:
I'll deploy this to S3 tomorrow -- Feel free to try it locally if you are keen ;)

There are still some styling issues that we'll address over the next day or two, but make a note of them if you see them, I guess.


abernix:
These look great, man.  Just submitted the first pull request against this branch (sorry).

Is this going to be a separate repo like `guide` or is it staying nestled within the `meteor` repo for now?

stubailo:
@abernix personally I would prefer to have a separate repo, but this is a good start I think. One thing to consider is that part of the docs are generated from code comments.

abernix:
@stubailo The immediate benefits I saw were
1. Ability and possibly to delegate a community manager for – like the guide.
2. Less intimidating issue/PR-filing process
3. No massive 1 hour CI builds for a doc PR, but maybe there's another way around that.

But ya, the jsdoc stuff would need consideration – semi-fixable with either:
1. git-submodules (but then there's the whole, you know, git-submodules thing 💩),
2. or, (a bit hacky, but) change the `jsdoc` generation script to check if `../../meteor-docs` directory exists and `.git/description` is the meteor docs repo and write it's files there.

Anyway, just a thought.  This is a huge improvement either way.  Just trying to take it to 🌠 .

stubailo:
Yeah I agree there are huge benefits to a separate repo to make contributions much simpler.

tmeasday:
I'm still on the fence about a separate repo. The main pro I see is that you can have a single commit that both adds/changes a feature and updates the docs.

However, if it's a big enough change that it affects the non "api-box" part of the docs, then maybe the guide needs to change too. So coordinating commits across repos is probably a thing no matter what.

The other complexity with separate repos is coordinating versions. If we have a 1.3 and 1.4 docs, then we need to make sure we generate `data.js` from the correct branch of the Meteor repository. This would make a pretty strong case for the submodule I think.

tmeasday:
My vote is for `meteor/docs` with a submodule into `meteor/meteor`.

stubailo:
How would running locally work? How would it get the JSDoc stuff? Perhaps the build script can just clone Meteor and run the doc script?

tmeasday:
I think the instructions would be something like:

```
# to run:
git submodule update --init
./src/scripts/admin/jsdoc/jsdoc.sh
hexo server
```

tmeasday:
Can I take "hooray emoji" ~== this a good idea, you should do it?

stubailo:
Yes!

Issue #6462, opened 2016-03-11
Expire password reset links


lorensr:
When Eve gains access to Bob's email, a drive-by/automated action is to search for and try using all password reset emails. If Bob has used the link, the link no longer works. However, if Bob does not use the link, I don't think that the link expires over time?

Recommendations:

1. Expire it by default. (`services.password.reset`) Most people use the link right away, but expiring after a day should catch 99% of uses. In the worst case, the user can simply get another reset email sent. For reference, a `setInterval` is used for expiring resume tokens: https://github.com/meteor/meteor/blob/0f98ea22fcc3f10cfd36aba66e531bf22c70f96e/packages/accounts-base/accounts_server.js#L1127
2. Provide a configuration function similar to `Accounts.loginExpirationInDays`: http://docs.meteor.com/#/full/accounts_config

cc @michelfloyd

MichelFloyd:
24 hours should be more than adequate. Even 4 hours should cover 99.9% of cases.

stubailo:
I think it would be good to add a similar mechanism to expiring resume tokens, but for the password reset tokens. Would someone want to submit a PR for it?

alphanso:
I would like to give it a try. Is anybody else working on it?

lorensr:
Don't think so!

alphanso:
I have created a draft pull request. Just wanted to ensure that approach to fix the issue is correct. Once I get a sign-off, I will add test and update the pull request.

lorensr:
On a related note, I think I remember `Accounts.onEmailVerificationLink` logs you in? If so, do those links expire?

alphanso:
I checked 
https://github.com/meteor/meteor/blob/devel/packages/accounts-password/password_server.js

It doesn't expire verification link tokens.

alphanso:
@stubailo Please have a look at it

tmeasday:
Fixed by #7534!

serkandurusoy:
When that PR is used with aldeed:collection2 where a user schema is attached to the users collection, it throws error: After filtering out keys not in the schema, your modifier is now empty.

I'll go ahead and post this over at the collection2 repo, but I thought it would not do harm if also noted here, or perhaps there's something we can do about it!

MichelFloyd:
I'm seeing the following possibly related error when deployed to galaxy:

**MongoError: Cannot apply $pull to a non-array value**
```
[...]

PR #6833, opened 2016-04-16
Expire password reset token

alphanso:
Draft fix for issue #6462. This fix is not ready for merge yet.

- [ ] Get a sign off that approach in theory is correct
- [ ] Add test case for configuration
- [ ] Add test case for token expiration error

### DIFF
--- a/packages/accounts-base/accounts_common.js
+++ b/packages/accounts-base/accounts_common.js
@@ -78,6 +78,9 @@ export class AccountsCommon {
   // - loginExpirationInDays {Number}
   //     Number of days since login until a user is logged out (login token
   //     expires).
+  // - passwordResetTokenExpirationInDays {Number}
+  //     Number of days since password reset link creation until it can't
+  //     be used anymore (password reset token expires)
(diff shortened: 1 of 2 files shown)


alphanso:
@data I have created this draft pull request to get a sign-off that approach is right. Please provide guidance :)

tmeasday:
@alphanso this looks generally sensible. I think apart from a test for expired reset tokens (I'm not sure we need one for the configuration part, but if you are keen..) it would be good to also clean up old tokens like we do here: https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_server.js#L1084

tmeasday:
Also, there are some failing tests. I'd rebase off current devel as some of those failures look like flakey ones that have been tightened up, but there are some accounts related ones there worth looking at.

tmeasday:
@alphanso, thanks again for this PR. We're still waiting to hear about the comments above--just to let you know, we'll be closing this PR in a week if not.@alphanso

PR #6838, opened 2016-04-18
Bump the sockjs package

mitar:
This exposes additional headers: https://github.com/sockjs/sockjs-node/blob/master/Changelog

This is useful when working with [Sandstorm](http://sandstorm.io/) which uses some of those headers to convey current host, which can then be used for dynamic `Meteor.absoluteUrl`.

### DIFF
--- a/packages/ddp-server/package.js
+++ b/packages/ddp-server/package.js
@@ -6,7 +6,7 @@ Package.describe({
 
 Npm.depends({
   "permessage-deflate": "0.1.5",
-  sockjs: "0.3.14"
+  sockjs: "0.3.17"
 });
 
 Package.onUse(function (api) {


tmeasday:
@mitar can we rebase this to devel (or just hit the upgrade button on the PR)

mitar:
> 

@mitar can we rebase this to devel (or just hit the upgrade button on the PR)

I do not understand at all what you asked here. It is already on `devel`. What is upgrade button?

tmeasday:
The self tests are failing because it was branched off `devel` before we fixed them. So we need to either merge current `devel` or rebase this off `devel`. 

I believe there's a button the Github UI called "upgrade branch" that you'll see that does the merging automatically.

mitar:
Didn't find the button. Rebased.

stubailo:
Thanks for bumping the version!

PR #6868, opened 2016-04-20
Meteor no longer provides jquery history or layout

skirunman:
Documentation fix to remove references to jquery-history and jquery-layout packages, which Meteor no longer provides.


stubailo:
Thanks! Glad we got rid of these.

Issue #6351, opened 2016-02-26
Allow users to add Babel plugins to their project


clayne11:
***1 Upvote*** Copied from [https://forums.meteor.com/t/meteor-1-3-and-babel-options/15275](url)

The BabelCompiler constructor in Meteor 1.3 takes options via the extraFeatures parameter. For example, here's what the ecmascript package does:

```
Plugin.registerCompiler({
  extensions: ['js'],
}, function () {
  return new BabelCompiler({
    asyncAwait: true,
    modules: true
  });
});

Plugin.registerCompiler({
  extensions: ['jsx'],
}, function () {
  return new BabelCompiler({
    asyncAwait: true,
    modules: true,
    react: true
  });
});
```
It'd be nice if there were more options, and a way for us to configure them. Maybe they can live in package.json (since that will now be a common thing in Meteor 1.3+). It'd be nice to, for example, enable stage-0 features, etc. It might look like this:

```
// package.json
{
  // ..
  "ecmascriptOptions": {
    "classProperties": true
  }
}
```

That would enable class properties4 on top of the default options that the BabelCompiler already applies based on whether a file is a .js or .jsx file.

gadicc:
I think that while it's nice for compiler plugins to be able to add babel plugins, usually users simply want to enable babel plugins for files the `ecmascript` package is already responsible for.  We could - as suggested above by @trusktr - have options somewhere that `ecmascript` can use.  But in light of recent conversations and strategy to be more integrated with the general Javascript ecosystem, I'd suggest simply relying on `.babelrc`.

That way, if someone wants to add a plugin, they simply follow the instructions on that plugin's webpage (i.e. `npm install` and add to `.babelrc`), rather than having a different "Meteor way" of doing it, like we had to do in the past with e.g. npm packages.

I know we disabled `.babelrc` support in https://github.com/meteor/meteor/issues/6016 via https://github.com/meteor/babel/commit/1eb5cbd97c8427f62703930e2d441710943c1720, but I think that was more for a quick fix than a design / policy decision.  Back in the [the original post](https://forums.meteor.com/t/meteor-1-3-and-babel-options/15275/5?u=gadicc) on the forums, @benjamn said:

[...]

PR #6874, opened 2016-04-21
Simple .babelrc support implementation

gadicc:
@benjamn, for when you get started on this (#6351).

I removed all the hotloading-related code, so this is hopefully a lot clearer.  It assumes the changes to `meteor-babel` from my PRs there.

Obviously not looking to have this merged as is.  I think it covers the main points so I hope will help.  You can use just as inspiration or merge and adapt as desired.  Let me know if you'd like anything else from me on this.

I think the main improvement that could be made when having control over all of Meteor vs a package, is to have Meteor watch and restart on a change to the various .babelrc's.  That and a better way to get the project directory :)  See also my notes (in comments) about having removed `babel-preset-meteor` and suggestion to do the same for `react`.

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -48,6 +48,8 @@ BCp.processFilesForTarget = function (inputFiles) {
       }
 
       var babelOptions = Babel.getDefaultOptions(self.extraFeatures);
+      var deps = mergeBabelrcOptions(babelOptions, inputFile);
+      deps.sourceHash = toBeAdded.hash;
 
       babelOptions.sourceMap = true;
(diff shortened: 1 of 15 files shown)


gadicc:
Sample docs :)

https://github.com/gadicc/meteor-react-hotloader/blob/master/docs/babelrc.md

stubailo:
I believe that @benjamn has implemented `.babelrc` support in 1.3.4 independent of this PR, so it doesn't make sense to merge this. Let me know if that's not accurate.

gadicc:
Yes, that's right (in 1.3.3).  Thanks, totally slipped my mind to close this! :D

PR #6880, opened 2016-04-22
DDP Buffering Tests for Meteor 1.3.3

abernix:
This is an adaptation of @tmeasday's 74230beba8777cb9104c4180525e2a3fe09772c4 test that he created for meteor/meteor#5680.  Due to occasional failures, it now uses sinon's `lolex` npm library to allow us to control the setTimeout/setInterval within the test itself, providing for tick-accurate testing. Also nifty because it allows the test to finish in less time than it actually takes.

### DIFF
--- a/packages/ddp-client/livedata_connection_tests.js
+++ b/packages/ddp-client/livedata_connection_tests.js
@@ -98,31 +98,25 @@ Tinytest.add("livedata stub - receive data", function (test) {
 });
 
 Tinytest.add("livedata stub - buffering data", function (test) {
-  var stream = new StubStream();
-
-  // Install special setTimeout that lets us control it in tests, courtesy of sinon's lolex
+  // Install special setTimeout that allows tick-by-tick control in tests using sinonjs 'lolex'
   // This needs to be before the connection is instantiated.
-  var clock = lolex.install();
+  const clock = lolex.install();
+  const tick = (timeout) => clock.tick(timeout);
 
-  var conn = newConnection(stream, {
+  const stream = new StubStream();
+  const conn = newConnection(stream, {
     bufferedWritesInterval: 10,
-    bufferedWritesMaxAge: 40
+    bufferedWritesMaxAge: 40,
   });
 
   startAndConnect(test, stream);
 
-  var coll_name = Random.id();
-  var coll = new Mongo.Collection(coll_name, conn);
+  const coll_name = Random.id();
+  const coll = new Mongo.Collection(coll_name, conn);
 
-  var testCount = function(count) {
-    test.equal(coll.find({}).count(), count);
-  };
+  const testDocCount = (count) => test.equal(coll.find({}).count(), count);
 
-  var wait = function(timeout) {
-    clock.tick(timeout);
-  };
-
-  var addDocument = function() {
+  const addDoc = () => {
     stream.receive({
       msg: 'added',
       collection: coll_name,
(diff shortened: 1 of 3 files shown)


packages/ddp-client/livedata_connection_tests.js
```diff
@@ -95,6 +95,92 @@ Tinytest.add("livedata stub - receive data", function (test) {
   test.isUndefined(conn._updatesForUnknownStores[coll_name]);
 });
 
+Tinytest.add("livedata stub - buffering data", function (test) {
```
abernix: I could wrap this in a `Meteor.isServer` to avoid the error happening on client tests, however I thought this might eventually (soon?) be ported over to using `emcascript`, which might allow this test to work there too.  Let me know if you want me to add it now.

packages/ddp-client/livedata_connection_tests.js
```diff
@@ -95,6 +95,92 @@ Tinytest.add("livedata stub - receive data", function (test) {
...
+  var lolex = Npm.require('lolex');
```
tmeasday: Just use `require()` (or `import lolex from 'lolex'` if we are using `ecmascript` in this package/test?)

packages/ddp-client/livedata_connection_tests.js
tmeasday: I'm not sure I totally understand, but we can definitely add a dependency on `ecmascript`, certainly in the `onTest` section.

tmeasday:
Thanks for this @abernix.

Perhaps you can change the various timings to be more rational? (I think you've just copied the ones over I set originally, which is when I was working around the non-exact `setTimeout` method I was using). Glancing at the timeouts it's not immediately obvious what we are trying to test out.

packages/ddp-client/livedata_connection_tests.js
abernix: @tmeasday Ok, I just didn't see *any* (???_ other internal packages doing anything besides `Npm.depends` and `Npm.require` for actual npm packages so wasn't sure if it was supported or desired.  Do I need to create `package.json` files for `ddp-client` if I go that route?

abernix:
Well, your original timings weren't completely unreasonable.  I did refine them a bit and make them accurate tick-for-tick though.  I'll see if I can do anything more with them though.

tmeasday:
Oh, OK, sure, whatever makes sense.

packages/ddp-client/livedata_connection_tests.js
tmeasday: That's because all the core packages were written prior to 1.3 :)

tmeasday: > Do I need to create package.json files for ddp-client if I go that route?

Nope, you can use `require` for `Npm.require()`-ed things.

packages/ddp-client/livedata_connection_tests.js
```diff
@@ -95,6 +95,92 @@ Tinytest.add("livedata stub - receive data", function (test) {
...
+  wait(6); // 44 ticks
+  addDocument(); // 7 doc
+  wait(6); // 50 ticks
+  addDocument(); // 8 doc
+  wait(6); // 56 ticks! (36 ticks since last flush)
+
+  // We're just shy of the max age (40) right now, so still shouldn't have flushed.
+  testCount(2);
+  // Now we're over the max.
+  wait(4); // Ok, 60 ticks (40 ticks since the last flush and the max we're allowed)
```
tmeasday: But also 10 ticks since the last doc addition, maybe we should make it a little clearer?

abernix: Yea, I wasn't sure how much to go through and explain the math but I'll go through, organize a bit more and update.  Should be by Monday.

tmeasday: Reading it again I don't think much of a change is needed, maybe just make line 170 like 9ms or something so we are at 53 ticks when the final doc comes in.

abernix:
@tmeasday Updated and tried to make it a bit more clear: ES6'd, removed a couple of the early lines which were just confusing (and duplicated), and changed it to be at 53 ticks when the final doc came in.

As a bit of a concern with lolex loading outside of test-mode, I tried doing the `Npm.depends` for `lolex` only for `onTest` however I ran into the same situation as #6888 with it reloading packages indefinitely.

packages/ddp-client/livedata_connection_tests.js
abernix: Done.  The client test was failing when it was run in the browser because of the requirement on `lolex` and the (previous) inability to `Npm.require` in the client.  Now using `ecmascript` and `import` it works perfect.

tmeasday:
Looks great @abernix! Merged into a `ddp-batching` branch, we'll cut a beta from it soon I think.

tmeasday:
Also merged into `release-1.3.3`

PR #6894, opened 2016-04-25
Use dependencies in check package

trusktr:
It might be nice to show by example in the Meteor packages the recommended way of importing things.

It might be nice if there was an option to disable globals so that even if a package uses `api.export` those exports never get attached globally. The only way to use those would-be-meteor-package-exports would be to `require` or `import` them.


tmeasday:
Thanks for the PR @trusktr but we'll be looking to do a pass through the entire code base to do this reasonably soon.

PR #6902, opened 2016-04-25
Eagerly load test files in full app testing mode as well as regular test mode

### LINKED ISSUE #6901: Full app test files inside /import directory not being eagerly loaded in full app testing mode
***1 Upvote*** Files of the form `/import/*.app-test[s].*` used to be eagerly loaded by Meteor (and therefore, when running `meteor test --full-app --driver-package practicalmeteor:mocha`, my full app tests would run). This behavior was changed in https://github.com/meteor/meteor/commit/eb18d59faa1986d912a59bd57a4ecc03ba4b3572, which only eagerly loads test files when `global.testCommandMetadata.isTest` is true, but when running `meteor test --full-app --driver-package practicalmeteor:mocha`, `global.testCommandMetadata` is:

```
{ driverPackage: 'practicalmeteor:mocha',
  isAppTest: true,
  isTest: false }
```

It looks like https://github.com/meteor/meteor/commit/eb18d59faa1986d912a59bd57a4ecc03ba4b3572 made it into Meteor 1.3.2. When testing by hand, I see that my full app tests do run under Meteor 1.3.1, but not under Meteor 1.3.2.1 (I couldn't test Meteor 1.3.2 directly due to some bug in Meteor 1.3.2)

Am I correct in placing my full app tests under the `/imports` directory? The Meteor Guide says that tests should live in the imports directory alongside the relevant code, but it doesn't specifically mention full app tests. I looked to meteor/todos, and it *does* place its `*.app-test[s].*` files in the `imports/` directory. See [/imports/startup/client/routes.app-test.js](https://github.com/meteor/todos/blob/master/imports/startup/client/routes.app-test.js).

Once fix seems fairly straightforward, just change the `isTest` computation [here](https://github.com/meteor/meteor/blob/310d6df4ac2b411425c626094ed3d378206fee18/tools/isobuild/package-source.js#L1330-L1331) to account for `global.testCommandMetadata.isAppTest` as well as `global.testCommandMetadata.isTest`. I'll submit a PR with that change.

jfly:
This fixes #6901.

I'm happy to add/change any test if appropriate.

### DIFF
--- a/tools/isobuild/package-source.js
+++ b/tools/isobuild/package-source.js
@@ -1328,7 +1328,8 @@ _.extend(PackageSource.prototype, {
   _inferFileOptions(relPath, {arch, isApp}) {
     const fileOptions = {};
     const isTest = global.testCommandMetadata &&
-      global.testCommandMetadata.isTest;
+      (global.testCommandMetadata.isTest ||
+       global.testCommandMetadata.isAppTest);
 
     if (isTest) {
       if (isTestFilePath(relPath)) {


jfly:
I can't figure out why circleci is failing (although I haven't dug too hard). Feel free to close this PR.

pmoons:
```
{"groupPath":["tinytest","livedata stub"],"test":"reconnect non-idempotent method","events":[{"sequence":2489,"type":"fail","details":{"type":"true","not":false},"cookie":{"name":"livedata stub - reconnect non-idempotent method","offset":0,"groupPath":["tinytest","livedata stub"],"shortName":"reconnect non-idempotent method"}}]}
```

My PR failed because of the same issue.  Seems to be happening to a lot of people.  I ran a test on a fresh checkout and it was failing as well, so this fail probably has nothing to do with your code.

tmeasday:
Thanks for the PR and nudge @jfly! As I mentioned on the issue, this PR doesn't quite do it because it means that full app tests become equal to normal tests (i.e. *all* non-test files are loaded lazily).

I've taken inspiration from this PR and am working on a complete fix. Thanks again.

jfly:
That makes sense. Thanks for the explanation!

PR #6911, opened 2016-04-26
Error wasn't being instantiated property in livedata connection onReset

### LINKED ISSUE #6906: Tiny Test fails - livedata stub - reconnect non-idempotent method
When checking out a commit as recent as [7012957a](https://github.com/meteor/meteor/commit/7012957a0d1f48c5b141e3e2a34fd450457a7d56), running tests will fail with the `livedata stub - reconnect non-idempotent method` message.

Meteor 1.3.2.4
No extra dependencies.
**Expected behavior**: I expected the tests to pass, but one fails.
**Steps to reproduce**: Check out that commit and run `meteor test-packages ./`

I believe this [line](https://github.com/meteor/meteor/blob/devel/packages/ddp-client/livedata_connection_tests.js#L780) is the culprit.  There is never any error received from the method call when the connection is reset.

abernix:
This was causing the "livedata stub - reconnect non-idempotent method" test to fail because it couldn't properly detect the (very much desired) error.

Fixes #6906

### DIFF
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -302,7 +302,7 @@ var Connection = function (url, options) {
           currentMethodBlock.splice(i, 1);
 
           // Make sure that the method is told that it failed.
-          methodInvoker.receiveResult(Meteor.Error('invocation-failed',
+          methodInvoker.receiveResult(new Meteor.Error('invocation-failed',
             'Method invocation might have failed due to dropped connection. ' +
             'Failing because `noRetry` option was passed to Meteor.apply.'));
         }


pmoons:
Yay 👍  TinyTests are passing again.  Having trouble discerning why CI failed. Maybe something with the cache and `less`?

```
... fail!
                                              
                                             \
                                              
  => match-timeout at ../../../tools/tool-testing/selftest.js:34
                                              
                                             \
                                              
  => Pattern: CACHE(less): Ran (#1) on: ["/subdir/nested-root.less","/top.less"]
                                              
                                             \
                                              
  => Last 100 lines:
                                              
                                             \
                                              
  1| [[[[[ /tmp/mt-1p1k9cg/home/myapp ]]]]]
                                              
```

abernix:
@pmoons - Ehh, they've been failing on a few packages for a while, `compiler-plugins` and `autoupdate` last I checked.

pmoons:
Gotchya. Hopefully this gets merged soon, could fix a lot of failing CI tests for other users.

abernix:
@pmoons Are people actually using `meteor self-test` in their CI?  Aside from how long it takes, how on earth is anyone actually getting them to pass reliably? The hot code push test has been failing randomly for...months?...Years?  I've always just assumed that only the trained eyes of a core team-member could decide if the tests have "passed" or not. 🍻

pmoons:
Not sure.  Just starting trying to contribute to Meteor.  I figured that all those CI tests failing in the PRs from the past few months were because of the `non-idempotent method test failing`.  If the push test is failing randomly for that long though, that's really disturbing. 😨

abernix:
Eh, don't be too deterred – they're fine in general and I think there's just some tests with known/understood issues.  Here's to one less! 🍸

abernix:
Looks like @tmeasday fixed this the same way in 08577711928bd6b0ab09ed80e038f3185ced3717, so closing this PR.

Tests are all passing again! 🎉

Issue #6764, opened 2016-04-09
[1.3] Should warn/fail on using `import` in `meteor shell`


gadicc:
This seems to be coming up almost every day on the forums... maybe one day people will search first, but OTOH, it is a little counter-intuitive that the statement "works" (without error) but "doesn't work" (in result).  We should assume that most users won't understand the different between a full app compile vs REPL.

@benjamn

e.g.

* https://github.com/meteor/guide/issues/312#issuecomment-207054402 (your comment)
* https://forums.meteor.com/t/silly-noob-import-export-question/21040 (2016-04-09)
* https://forums.meteor.com/t/1-3-1-strange-import-behavior/20940/5 (2016-04-07)
* https://forums.meteor.com/t/how-to-import-npm-in-meteor-1-3/17901 (2016-04-07)
* Probably others I missed and from earlier.

I could do a PR towards end of next week, don't think I'll have a chance before then.

benjamn:
Warning is a great idea!

gadicc:
Solved with [reify](https://github.com/benjamn/reify) in Meteor 1.3.3; no warning necessary :D

PR #6914, opened 2016-04-26
meteor shell: warn when using "import" in a way likely to fail

gadicc:
PR for #6764 (cc: @benjamn)

I originally wanted to warn rather than fail, but with the way the result is sent back via a callback, it's one or the other.  Nevertheless, we can avoid failing on cases where we're confident the user actually knows what they're doing (i.e. using the imported var on the same line, see notes in `ecmascript.js` and test cases).

Not sure where the best place is to redirect for more info; happy to amend the (currently linked) original issue to explain what's going on here in more depth, and then I guess it will (helpfully) be an obvious place to report unexpected errors relating to importing from the shell.  Otherwise will amend however you suggest.

### DIFF
--- a/packages/ecmascript/ecmascript.js
+++ b/packages/ecmascript/ecmascript.js
@@ -3,6 +3,21 @@ ECMAScript = {
     const babelOptions = Babel.getDefaultOptions();
     babelOptions.sourceMap = false;
     babelOptions.ast = false;
+
+    /*
+     * Since import *will* work as expected if everything is on one line,
+     * we'll allow a case of the import statement being followed by a
+     * semicolon, some optional whitespace and a word character.
+     */
+    if (command.match(/(?!.*?;\s*\S+)^\s*import/)) {
+      throw new Error(
+        'Using "import" in the Meteor shell is unlikely to work how you ' +
+        'expect.  Try "var something = require(\'moduleId\').default;" ' +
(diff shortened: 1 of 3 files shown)


gadicc:
Just noticed the failing test, but it's unrelated to my commit:

```
{"groupPath":["tinytest","livedata stub"],"test":"reconnect non-idempotent method","events":[{"sequence":2489,"type":"fail","details":{"type":"true","not":false},"cookie":{"name":"livedata stub - reconnect non-idempotent method","offset":0,"groupPath":["tinytest","livedata stub"],"shortName":"reconnect non-idempotent method"}}]}
```

I did add one test which is passing:

```
S: tinytest - ecmascript - server - compileForShell - warn on imports : OK
```

and all tests passed on my system, not sure what's specific about the above test that is failing on CI.

benjamn:
@gadicc While I very much appreciate the pragmatism of this pull request, I have some good news: Meteor 1.3.3 will use [a completely new compiler for ES2015 modules](https://github.com/benjamn/reify). Among numerous other benefits, `import` statements will just work in `meteor shell`, because they now create normal variables rather than `_obj.prop` references.

gadicc:
@benjamn, reify looks superb.  Yes, definitely making imports work how people expect is the better way of solving this :D

PR #4507, opened 2015-06-03
Allow all pending observe callbacks to be run before stop

mitar:
This allows fixing a really nasty race-condition between when observe callbacks are scheduled and when they are run. Currently, you can get observe callbacks scheduled, you can call stop, and then callbacks are run. Which can be quite surprising (that callbacks are run after observe was stopped). Especially if code around stop also cleaned some things observe callbacks expect to still have.

Alternatively, we might want to expose a `flush` method on the handle, which flushes pending callbacks when called?

Also, should handle stop in fact remove all pending callbacks? So I think there should be maybe three options:
* leave callbacks to run after stop
* run all callbacks before stop
* drop all pending callbacks

PR #6917, opened 2016-04-26
Allow all pending observe callbacks to be run before stop (take 2)

mitar:
(This is a followup to https://github.com/meteor/meteor/pull/4507.)

This allows fixing a really nasty race-condition between when observe callbacks are scheduled and when they are run. Currently, you can get observe callbacks scheduled, you can call stop, and then callbacks are run. Which can be quite surprising (that callbacks are run after observe was stopped). Especially if code around stop also cleaned some things observe callbacks expect to still have.

Alternatively, we might want to expose a `flush` method on the handle, which flushes pending callbacks when called?

Also, should handle stop in fact remove all pending callbacks? So I think there should be maybe three options:
* leave callbacks to run after stop
* run all callbacks before stop
* drop all pending callbacks

### DIFF
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -398,21 +398,27 @@ _.extend(LocalCollection.Cursor.prototype, {
     var handle = new LocalCollection.ObserveHandle;
     _.extend(handle, {
       collection: self.collection,
-      stop: function (flush) {
-        if (flush && Meteor.isServer && self.collection._observeQueue._draining) {
+      stop: function () {
+        if (self.reactive)
+          delete self.collection.queries[qid];
+      },
+      flush: function () {
+        // If it is already draining, then draining is hapenning in some other
+        // fiber, so we want to block this fiber as well.
+        if (Meteor.isServer && self.collection._observeQueue._draining) {
           var Future = Npm.require('fibers/future');
           var future = new Future();
           self.collection._observeQueue.queueTask(function () {
             future.return()
           });
-          // Just to make sure that this will be drained. Probably this is a noop because we are
-          // already draining when we enter this code path.
+          // Just to make sure that this will be drained. Probably this is a
+          // noop because we are already draining when we enter this code path.
           self.collection._observeQueue.drain();
           future.wait();
         }
-
-        if (self.reactive)
-          delete self.collection.queries[qid];
+        else {
+          self.collection._observeQueue.drain();
+        }
       }
     });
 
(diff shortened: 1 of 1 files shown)


tmeasday:
Hey @mitar,

Some questions / notes:

1. Does the bug only appear for MiniMongo on the server?
2. Should we be calling `flush` before `stop` automatically?
3. Is there a way to flush only the callbacks relevant to this observer (I guess not)?
4. We definitely need to get a test in to highlight the behavior that this fixes (esp if we do 2.).

It should be possible to write an async test that highlights the situation, I would have thought. Something like:

```js
const cursor = X.find();
const observer = cursor.observe({
  added() {
     Meteor._sleepForMs(100);
     console.log('added called');
  }
});

X.insert({});
X.insert({});
observer.stop();
```

That may or may not work but the bug would come up when your observe callback yields I imagine.

Perhaps we should be passing observe callbacks to another fiber, although that feels like a more drastic change.

mitar:
OK, a disclaimer, this was now quite some time ago. Last year. And the project where I was hitting this issue is done. That particular project had a requirement that exactly the same code should run both on the server and client side. This is why we implemented server-side autorun, fixed Minimongo discrepancies between server and client and so on (some pull requests I made).

> Does the bug only appear for MiniMongo on the server?

We observed this only on the server, I think because it uses fibers so multiple code paths are possible.

> Should we be calling flush before stop automatically?

Probably. I think it is a better semantics. On the other hand, we might want to things: one way to flush, the other to cancel all of those.

> Is there a way to flush only the callbacks relevant to this observer (I guess not)?

It should require some changes (like storing for each callback where it is coming from), but it could be possible. On the other hand, I think this would change semantics in strange ways (some callback being run now in different order than they were queued). Not sure if we want that.

In general I think that when you are calling `.stop()` you are probably OK waiting a bit, because it is not really a time-sensitive part of the code.

> Perhaps we should be passing observe callbacks to another fiber, although that feels like a more drastic change.

There are some issues here anyway, like #2691. I made this test #5038 for it. So yes, calling observes out of the main "thread" should probably be done anyway (but in that case especially on client).

tmeasday:
Actually so I've been repeatedly revisiting this one, and I'll admit I'm confused.

How is it actually possible to reach this race? Each of `insert/added/removed` call `_observeQueue.drain()`, which waits for the queue to drain anyway (i.e. all tasks to complete).

Can you outline (ideally via a test) an example of how this could happen? My example above seems wrong now.

mitar:
I think we should close this for now. I do not have access to that code anymore and was also a random case, not a clear reproduction. I think it is good to have it here if anyone else triggers it and then we can reopen it.

Issue #6907, opened 2016-04-26
Proper import chain for Meteor packages


gadicc:
In [modules/install-packages.js](https://github.com/meteor/meteor/blob/83c2f7a790dbf13f45a153ed5320f4692e706b41/packages/modules/install-packages.js) we have:

```js
  meteorDir[name + ".js"] = function (r, e, module) {
    module.exports = Package[name];
  };
```

Would it be possible to get something like this instead:
```js
  meteorDir[name + ".js"] = [mainModule, function(require, e, module) {
    module.exports = require(mainModule);
  }];
```

This is important for hotloading for updates to flow down correctly, and required for the new version of the original react-hot-loader package (although there is a [workaround](https://github.com/gaearon/react-hot-loader/issues/259) that can be used to catch the update in each individual package too).  And I guess affects possible bundle analysis in the future.

cc: @benjamn as usual :)

I'd say if you tell me that the name of the `mainModule` is already available in `computeJsOutputFilesMap` somewhere I'll submit a PR for this, except I already owe you a much simpler PR from like two weeks ago :(

gadicc:
I think we can just track this via the PR.

abernix:
Let's actually leave it open until resolved, per policy for FRs. :wink:

PR #6931, opened 2016-04-28
Proper import chain for Meteor packages to app

gadicc:
Addresses #6907.

@benjamn, this is what I had in mind and solves the issue for me.  Pending any comments from you I'll play with it for a few more days then add some tests to the PR and rebase.

### DIFF
--- a/tools/isobuild/compiler-plugin.js
+++ b/tools/isobuild/compiler-plugin.js
@@ -776,8 +776,15 @@ export class PackageSourceBatch {
 
       map.forEach((info, name) => {
         if (! name) return;
+
+        let mainModule = _.find(info.files, file => file.mainModule);
+        mainModule = mainModule ?
+          `meteor/${name}/${mainModule.targetPath}` : false;
+
         meteorPackageInstalls.push(
-          "install(" + JSON.stringify(name) + ");\n"
+          "install(" + JSON.stringify(name) +
+            (mainModule ? ", " + JSON.stringify(mainModule) : '') +
+          ");\n"
         );
       });
 
--- a/packages/modules/install-packages.js
+++ b/packages/modules/install-packages.js
@@ -1,4 +1,4 @@
-function install(name) {
+function install(name, mainModule) {
   var meteorDir = {};
 
   // Given a package name <name>, install a stub module in the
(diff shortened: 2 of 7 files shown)


tools/isobuild/compiler-plugin.js
```diff
@@ -776,8 +776,14 @@ export class PackageSourceBatch {
 
       map.forEach((info, name) => {
         if (! name) return;
+
+        let mainModule = _.find(info.files, file => file.mainModule);
+        mainModule = mainModule ?
+          `meteor/${name}/${mainModule.targetPath}` : false;
```
benjamn: This can just be `mainModule.installPath`, I think, since that includes `meteor/${name}/` prefix already.

gadicc: Thanks for the quick look!  `installPath` isn't available yet, since currently the code runs just before the import-scanner.  Would you prefer to move it to after?  The import scanner takes the existing map as input, so not sure if any side effects here, or on mutating the generated scannerMap.

I did a quick test and started getting "Can't find npm module" errors for e.g. 'meteor/underscore', but I might have rushed too much.  Also `installPath` is like `node_modules/meteor/...` so would need to either strip the `node_modules/` or prefix with `/`.

gadicc:
Hope the tests are ok.  Happy to rename the two new packages (in `tests/apps/package-tests`) or test for other things, per your guidance.

gadicc:
@benjamn, this would be appropriate for 1.3.3 I think, yeah?

In light of my earlier comments, I think better to leave the code as is.

Re failing tests, they're not mine:

```
selftest:  - package-tests: change packages during hot code push
```

Unfortunately that last test breaks the CI process, but locally, my test passes (just a simple unit test to make sure the output for modules.js is as expected for both cases).

It's possible to squash directly via github merge now, but happy to rebase if you prefer (I guess that would trigger a new CI run, iirc you and others have made the CI testing more stable recently).

abernix:
@gadicc Please rebase.  While 2-3 reruns of the old (flakier) tests would probably go through, a rebase will almost certainly work the first time and will trigger the build.

gadicc:
@abernix, done; tests passing :D  definitely big improvements in testing... and this community issue management is a game changer!  thanks!

gadicc:
Awesome, thanks @benjamn for getting this in 1.3.3!
cc: @clayne11, think it will help with your current project.

clayne11:
Yes that's amazing! Great work 👍

PR #6938, opened 2016-04-29
Docs <--> Hostory.md inconsistancy about Match.Optional

lucendio:

I am not sure which of them is wrong, but something seems to be a bit odd here. See [History.md](https://github.com/meteor/meteor/blob/devel/History.md#v132) third-last bullet point.
The second sentence regarding the usage in an object seems to make no sense as well, or I am not able to get my head around it.

### DIFF
--- a/docs/client/full-api/api/check.md
+++ b/docs/client/full-api/api/check.md
@@ -122,7 +122,7 @@ check(undefined, Match.Maybe(String)); // OK
 
 {{#dtdd "<code>Match.Optional(<em>pattern</em>)</code>"}}
 
-Behaves like `Match.Maybe` except it doesn't accept `null`. If used in an object, the behavior is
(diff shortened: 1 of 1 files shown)


tmeasday:
@abernix care to comment on this one?

abernix:
The change in this pull-request is incorrect.  The History.md is correct as it is.  However, the functionality of `check`/`Match` on objects is still confusing and is being addressed in a different issue (#6959)

@lucendio `Maybe.Optional` and `Match.Maybe` are treated differently inside of objects because of the extra distinction of "not existing entirely".  See #6959 where I just posted all the example which will hopefully make it more clear for you.  Also, there's a discussion about what's confusing about the current `check` and how it might need to be changed.  You're not the first one to be confused about it, but hopefully you will see from my examples what it all means.  Feel free to discuss more in that issue.

@tmeasday This can be closed.

PR #6940, opened 2016-04-29
[docs-hexo] Clarify that "session" and "check" must be installed

### LINKED ISSUE #6932: [Docs] It's not clear in the docs that "session" and "check" must be added
As forums user **gris** [brought up here](https://forums.meteor.com/t/session-package/22276/2), it's not clear that `session` must be added in order to use it.  Other packages make a very clear distinction.  This probably wasn't updated when the packages were no longer default starting in (I think) 1.2.

The same omission seems to be on `check`.

PR coming.

abernix:
Same thing I did in fd2bc2a8b31bb35ca65cf946e8f1d9171cd8d2f2 except for the (previously unware-of) hexo docs.

Fixes #6932

### DIFF
--- a/docs/source/api/check.md
+++ b/docs/source/api/check.md
@@ -7,6 +7,12 @@ The `check` package includes pattern checking functions useful for checking the 
 of variables and an [extensible library of patterns](#matchpatterns) to specify which types you are
 expecting.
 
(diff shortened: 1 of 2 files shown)


tmeasday:
Thanks @abernix

Issue #6723, opened 2016-04-06
[1.3.1] Update results in Error: Can't find npm module 'babel-runtime/helpers/typeof'.


pors:
Updating from 1.3 to 1.3.1 broke something for me. All was running fine with 1.3, now I get:

```
$ meteor run
[[[[[ ~/dev/reactionic-kitchensink/app ]]]]]

=> Started proxy.
=> Started MongoDB.
W20160406-20:06:32.096(3)? (STDERR)
W20160406-20:06:32.097(3)? (STDERR) /Users/mark/dev/reactionic-kitchensink/app/.meteor/local/build/programs/server/boot.js:324
W20160406-20:06:32.097(3)? (STDERR) }).run();
W20160406-20:06:32.097(3)? (STDERR)    ^
W20160406-20:06:32.213(3)? (STDERR) Error: Can't find npm module 'babel-runtime/helpers/typeof'. Did you forget to call 'Npm.depends' in package.js within the 'modules-runtime' package?
W20160406-20:06:32.213(3)? (STDERR)     at Object.Npm.require (/Users/mark/dev/reactionic-kitchensink/app/.meteor/local/build/programs/server/boot.js:195:17)
```

etc.

eldog:
I had to delete my local meteor installation, i.e.,  run `rm -rf ~/.meteor`, to fix the error.

pors:
@eldog that does it for me as well, thanks!

I now get these warnings, but at least it works again

```
W20160409-11:45:31.527(3)? (STDERR)  - react@0.14.7 installed, react@15.x needed
W20160409-11:45:31.527(3)? (STDERR)  - react-addons-pure-render-mixin@0.14.1 installed, react-addons-pure-render-mixin@15.x needed
```

vittaljk:
Thank you it helped me!!!!!

meyian:
I'm having this error, and deleting the .meteor folder means you can no longer run your application. Need help with this issue, is it possible to have this thread reopened?

meetmagdalene:
+1 for this issue to be reopened. 

I receive this error after updating to meteor 1.3.2.4.

timmygee:
I've re-opened a similar issue here https://github.com/meteor/meteor/issues/6918

namelivia:
I'm also having the same issue, I tried doing `rm -rf ~/.meteor` but then after doing `meteor update && meteor` for reinstalling and launching the application again I am getting the error again.

meyian:
[...]

PR #6944, opened 2016-04-29
Use MaxCDN as our new CDN

n1mmy:
MaxCDN has graciously agreed to host our compiled packages and bootstrap tarballs. Thanks, MaxCDN!

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,3 +103,5 @@ Interested in contributing to Meteor?
 
 We are hiring!  Visit https://www.meteor.com/jobs to
 learn more about working full-time on the Meteor project.
+
+Bandwidth and content distribution graciously provided by [MaxCDN](https://www.maxcdn.com/).
--- a/meteor
+++ b/meteor
@@ -61,7 +61,7 @@ function install_dev_bundle {
     mkdir "$BUNDLE_TMPDIR"
 
     # duplicated in scripts/windows/download-dev-bundle.ps1:
-    DEV_BUNDLE_URL_ROOT="https://d3sqy0vbqsdhku.cloudfront.net/"
+    DEV_BUNDLE_URL_ROOT="https://static-meteor.netdna-ssl.com/"
     # If you set $USE_TEST_DEV_BUNDLE_SERVER then we will download
     # dev bundles copied by copy-dev-bundle-from-jenkins.sh without --prod.
     # It still only does this if the version number has changed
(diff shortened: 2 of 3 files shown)


laosb:
By the way, how can we set up a mirror for compiled packages (which is served on `warehouse.meteor.com` for now I think)?

benjamn:
Awesome! I will test this out and merge it as soon as I can.

abernix:
Something seems to be going on with this.

I've seen a few reports of failed installs and downloads in the last few days.  #6960 [#6723 (comment)](https://github.com/meteor/meteor/issues/6723#issuecomment-216144477) and I think even #6951 (it's disguised in there).  Also, [on StackOverflow](https://stackoverflow.com/questions/36989049/getting-error-while-downloading-meteor-unexpectedly-download-stops).  I'm also not sure if the Windows installer was updated already, however #6955 also speaks of download problems (2x).

Just tried 10 times and I'm not experiencing any problems with it, but as you'll see with #6960, the problem went away when the user switched URLs.

n1mmy:
Hold off on merging this for now... we're having some issues with people in some countries (eg India) not being able to download from the new CDN.

laosb:
China developers always have problems in downloading anything abroad. Could you provide some tips in how to set up a mirror in China? That's what most Chinese developers care about.

n1mmy:
Hi @laosb. Sorry, I don't really have a great answer for you in terms of setting up a mirror in China. The meteor tool doesn't have built-in support for changing the base URL used (`warehouse.meteor.com`). If you run from git you can manually modify https://github.com/meteor/meteor/blob/devel/tools/packaging/warehouse.js#L46 to point elsewhere.

I'm curious, do you think a CDN edge node in Hong Kong would help? That might be an option we explore.

laosb:
@n1mmy will you accept a pr to enable env-override for the hardcoded url?

Hk node wouldn't help much, as China restrict all aboard data transfer, including hk tw and ma. But still better then nothing.

n1mmy:
@laosb Good question. I would think so. @benjamn @tmeasday ?

tmeasday:
Makes sense to me

n1mmy:
@benjamn We cut back over to MaxCDN for installers and it seems to be OK. Clear to merge.

zol:
@n1mmy - according to @benjamn we decided not to go with MaxCDN in the end so I'm going to close this PR. Re-open it if anything changes.

PR #6951, opened 2016-05-01
Create source.sh

shaileshtheroyal:

respected sir,  many people getting problem to download meteor package from curl too including me.proces stops randoms after downloading somewhat package. here i have found the new way which will allow continuous,robust and static (running scipt itself on users computer) download .
i just made source.sh file and copied the installer in it. now we just need to run shell srcipt using  #sh source.sh . and installation starts...
 hope you will include this as optional to user if original was getting problamaatic

yours' meteor user
shaileshvbagade@gmail.com

### DIFF
--- a/source.sh
+++ b/source.sh
@@ -1,0 +1,231 @@
+#!/bin/sh
+
+# This is the Meteor install script!
+#
+# Are you looking at this in your web browser, and would like to install Meteor?
+#
+# MAC AND LINUX:
+#   Just open up your terminal and type:
+#
+#     curl https://install.meteor.com/ | sh
+#
+#   Meteor currently supports:
+#    - Mac: OS X 10.7 and above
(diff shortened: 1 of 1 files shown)


shaileshtheroyal:
plz include this also on site for robust and static download

abernix:
@shaileshtheroyal Who has been having problems?  Do you have examples or discussions?  Open Issues?

There shouldn't be a difference between downloading the exact script and running it in a shell, versus sending the output of curl to a shell.  If you are having a problem, please provide information about what operating system you're using.

stubailo:
> Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

I think this should be discussed in an issue first, so that we can agree on the best solution.

PR #6972, opened 2016-05-04
Slight improvements

mitar:
I think it is somehow important to convey that batching until 500 ms happens only if changes are coming under 5 ms in. If they are not, they things get flushed.


stubailo:
thanks!

PR #6975, opened 2016-05-05
Remove the google `profile` scope.

MasterAM:
This is a re-submission of https://github.com/meteor/meteor/pull/5699.

The `email` scope is enough for getting the required data, including the user ID, so no need for the `profile` scope, which may deter users.

The discussion started at https://github.com/meteor/meteor/issues/5650

Our team anticipates Google to be a major choice for users during signup, so we would like to alleviate some of the concerns that requesting the `profile` scope may bring.

We would also like developers to have more control over more aspects, such as the way the popup window behaves (e.g, forcing it upon login for situations where there is more than one logged-in Google account).

I would really like to hear your opinions in the discussion thread.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,13 @@
 ## v.NEXT
 
+* The `google` package now uses the `email` scope as a mandatory field instead
+  of the `profile` scope. The `profile` scope is still added by default if the
+  `requestPermissions` option is not specified to maintain backward
+  compatibility, but it is now possible to pass an empty array to
+  `requestPermissions` in order to only request the `email` scope, which
+  reduces the amount of permissions requested from the user in the Google
+  popup. [PR #6975](https://github.com/meteor/meteor/pull/6975)
+
 ## v1.4
 
 * Node has been upgraded to 4.4.7.
--- a/packages/google/google_client.js
+++ b/packages/google/google_client.js
@@ -23,9 +23,9 @@ Google.requestCredential = function (options, credentialRequestCompleteCallback)
 
   var credentialToken = Random.secret();
 
-  // always need this to get user id from google.
-  var requiredScope = ['profile'];
-  var scope = ['email'];
+  // we need the email scope to get user id from google.
+  var requiredScope = ['email'];
+  var scope = [];
   if (options.requestPermissions)
     scope = options.requestPermissions;
   scope = _.union(scope, requiredScope);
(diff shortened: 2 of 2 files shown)


stubailo:
I'm a bit worried this might be a breaking change for people who are currently using the profile information in their apps. Perhaps there is something else we could do to allow one to disable the `profile` scope?

MasterAM:
We could make it configurable via Meteor.settings, and have it add the `profile` scope by default.
Having an explicit setting could override it.

Having said that, there are a couple more things I would like to make configurable about the Google auth process. Would you advise to add them to this PR or create a separate one once this one is agreed upon and merged?

tmeasday:
@MasterAM, sorry about the delays here.

A couple of things:

1. I think we should probably just swap the `requiredScope` and `scope` so by default the behavior is the same, but it is now possible to remove the `profile` scope by passing `requestPermissions = []`.

2. I think we should remove `locale` from here: https://github.com/meteor/meteor/blob/devel/packages/google/google_server.js#L4-L5

3. We need a (prominent) changelog entry.

tmeasday:
@MasterAM, thanks again for this PR. We're still waiting to hear about the comments above--just to let you know, we'll be closing this PR in a week if we don't.

MasterAM:
Hi Tom,

Forgot about the PR, thanks for reminding me.

1. No problem. Another option would be deprecating the current API (providing a deprecation message) and introducing a new config options (`requiredPermissions` or `extraPermissions` or any other suitable name) that will use the `email` scope by default, but that might be too much of a hassle. 

  A major version bump may be in place.

2. I don't see why. If a user wants a scope that returns locale (if available), why won't it be whitelisted? Do you expect them to directly mutate the 
When the current `Google.whitelistedFields` list?

3. No problem. Do you want me to write one? If so, where?

tmeasday:
1. I'd advocate the easiest thing :)

2. I mean it just seems weird to have a whitelisted field that we don't know is there for sure. But we could also just leave it I guess.

3. `History.md`

Thanks again!

Also, can you merge `devel` (or rebase) into this branch so the tests pass? Thanks!

MasterAM:
Rebased, made changes and added the changelog entry. I Hope it's not too verbose.

Note to users:
- If you did not specify the `requestPermissions` option, you don't have to change anything.
- If you specified this option and want to request the `profile` permission, you should add it to the array of your requested permissions.
- If you do not need the `profile

tmeasday:
Thanks very much for your hard work on this @MasterAM

PR #6981, opened 2016-05-05
Use longjohn to repair stacktraces on self-test

tmeasday:
@benjamn can you please check that I did the right thing in adding this to the dev bundle

### DIFF
--- a/tools/tests/compiler-plugins.js
+++ b/tools/tests/compiler-plugins.js
@@ -103,8 +103,13 @@ selftest.define("compiler plugin caching - coffee", () => {
     });
 
     // First program built (web.browser) compiles everything.
-    cacheMatch('Ran (#1) on: ' + JSON.stringify(
-      ["/subdir/nested-root." + extension, "/top." + extension]));
+    cacheMatch('Ran (#1) on: ' + JSON.stringify([
+      // Though files in imports directories are compiled, they are marked
+      // as lazy so they will not be loaded unless imported.
+      "/imports/dotdot." + extension,
+      "/subdir/nested-root." + extension,
+      "/top." + extension
+    ]));
     // There is no render execution in the server program, because it has
     // archMatching:'web'.  We'll see this more clearly when the next call later
     // is "#2" --- we didn't miss a call!
--- a/tools/utils/parse-stack.js
+++ b/tools/utils/parse-stack.js
@@ -15,7 +15,7 @@ const _ = require('underscore');
 // return anything past that function. We call this the "user portion"
 // of the stack.
 export function parse(err) {
-  var frames = err.stack.split('\n');
+  let frames = err.stack.split('\n');
 
   frames.shift(); // at least the first line is the exception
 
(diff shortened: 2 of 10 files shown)


tools/utils/parse-stack.js
```diff
@@ -15,10 +15,17 @@ const _ = require('underscore');
 // return anything past that function. We call this the "user portion"
 // of the stack.
 export function parse(err) {
-  const frames = err.stack.split('\n');
+  var frames = err.stack.split('\n');
```
Slava: should be `let`?

tmeasday: Thanks!

benjamn:
Looks good to me!

Now you just need to bump [`$BUNDLE_VERSION`](https://github.com/meteor/meteor/blob/9ea56a8d951c8405d1e66543cc4d43d6e857df23/meteor#L3), rebuild the dev bundle on Jenkins, and publish the new bundles to S3. Or you can manually `npm install longjohn` in `meteor/dev_bundle/lib/node_modules/` for now and we can rebuild the dev bundle later. Let me know if you need any help with that.

It's cool that `longjohn` uses the same version of `source-map-support` that we already use (0.3.2), even though ours is a fork installed from a GitHub tarball URL. I hope `npm` is clever enough to realize that those modules are the same.

tmeasday:
> I hope npm is clever enough to realize that those modules are the same.

npm 3 would be, I don't think npm 2 will unless we dedupe, but I'm not sure what the side effect would be?

tmeasday:
> Now you just need to bump $BUNDLE_VERSION, rebuild the dev bundle on Jenkins, and publish the new bundles to S3. Or you can manually npm install longjohn in meteor/dev_bundle/lib/node_modules/ for now and we can rebuild the dev bundle later. Let me know if you need any help with that.

Oh yeah, that's what I did locally, obviously :)

I'll try and figure out how to publish the dev bundle now, I'll hassle you if I can't figure it out

tools/utils/parse-stack.js
benjamn: Could also do
```js
const frames = err.stack.split("\n").slice(1)
  .filter(f => ! f.match(/^\-{45}$/));
```

tmeasday: :+1:

tmeasday:
I've copied the dev bundle to prod. So this is ready to merge to devel

tools/tests/compiler-plugins.js
```diff
@@ -103,8 +103,13 @@ selftest.define("compiler plugin caching - coffee", () => {
     });
 
     // First program built (web.browser) compiles everything.
-    cacheMatch('Ran (#1) on: ' + JSON.stringify(
-      ["/subdir/nested-root." + extension, "/top." + extension]));
+    cacheMatch('Ran (#1) on: ' + JSON.stringify([
+      // Though files in imports directories are compiled, they are marked
+      // as lazy so they will not be loaded unless imported.
+      "/imports/dotdot." + extension,
```
tmeasday: This now fails on the stylus self test however: https://circleci.com/gh/meteor/meteor/1389

benjamn: Fixed.

tmeasday: Oh it was an actually bug/missing feature that the self-test highlighted. Who knew :)

PR #6986, opened 2016-05-06
Don't reminify *.min.js

laosb:
#5378 Old PR
#5363 Issue

Resubmitted but I'm not ready with writing tests. Personally think that shouldn't have any problems. Pinging the member dealing with the old PR @stubailo

### DIFF
--- a/packages/standard-minifier-js/plugin/minify-js.js
+++ b/packages/standard-minifier-js/plugin/minify-js.js
@@ -33,7 +33,12 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
 
   var allJs = '';
   files.forEach(function (file) {
-    allJs += UglifyJSMinify(file.getContentsAsString(), minifyOptions).code;
+    // Don't reminify *.min.js.
+    if (/\.min\.js$/.test(file.getPathInBundle())) {
+      allJs += file.getContentsAsString();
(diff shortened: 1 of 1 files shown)


laosb:
Personally think .min.css shouldn't be reminified too.

benjamn:
This looks safe and beneficial to me.

benjamn:
Feel free to submit another PR for `.min.css` :)

laosb:
Should this be mentioned in History.md and docs? There might be some cases that people use .min.js for other reasons but still want be minified.

stubailo:
Definitely should be in the docs regardless!

mitar:
Yes, please add tests and docs as well, no?

laosb:
@mitar Learning how to write tests! Is there any document about it? When comes to docs, should I simply add it to the old Meteor-based docs or to the hexo-based docs in the PR?

mitar:
Just check how other tests are done and add to them. But I see that this package does not have them. So maybe then it is on MDG to write first ones. I do not know how to write tests for tooling.

mitar:
For documentation, yes, that is in flux currently.

tmeasday:
> I do not know how to write tests for tooling.

You can use `selftest` for this. I don't think it's well documented. We should fix that!

PR #6989, opened 2016-05-06
Give a hint when using a Chrome blocked port.

laosb:
When you try to visit your app in development with Chrome on a Chrome blocked port, for example, 6000, you would receive an `ERR_UNSAFE_PORT` on Chrome. This PR gives user a hint when they try to do so.
For the blocked port list, see the source: https://src.chromium.org/viewvc/chrome/trunk/src/net/base/net_util.cc?view=markup

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -95,12 +95,85 @@ export function parseServerOptionsForRunCommand(options, runTargets) {
 
 function parsePortOption(portOption) {
   let parsedServerUrl = utils.parseUrl(portOption);
-
+  // Chrome blocked those ports for safety reasons. 
+  // See: https://src.chromium.org/viewvc/chrome/trunk/src/net/base/net_util.cc?view=markup
+  const blockedPorts = [
+    1,    // tcpmux
+	  7,    // echo
+	  9,    // discard
+	  11,   // systat
+	  13,   // daytime
+	  15,   // netstat
+	  17,   // qotd
+	  19,   // chargen
+	  20,   // ftp data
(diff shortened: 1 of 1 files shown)


zol:
@laosb - why is the unsafe port error from Chrome not enough? Also, we now have a process for feature requests that this PR doesn't follow - it should reference an issue that's been classified and can gather community feedback.

laosb:
Actually Chrome doesn't tell you what happened, unless you expand details. In
fact you can only see an `ERR_UNSAFE_PORT`, and you might need to do a search
to figure out what happened.

  

Sorry for not following the process - I didn't read it carefully and I'm now
reading it byte by byte!

  
Sent from [Nylas N1](https://nylas.com/n1?ref=n1), the extensible, open source
mail client.  

On 5月 7 2016, at 3:33 凌晨, Zoltan Olah &lt;notifications@github.com&gt; wrote:  

> [@laosb](https://github.com/laosb) \- why is the unsafe port error from
Chrome not enough? Also, we now have a process for feature requests that this
PR doesn't follow - it should reference an issue that's been classified and
can gather community feedback.

>

> —  
You are receiving this because you were mentioned.  
Reply to this email directly or [view it on GitHub](https://github.com/meteor/
meteor/pull/6989#issuecomment-217538284)

zol:
I'm going to close this as I don't think the trade-off between maintaining a hardcoded list of ports vs 'user friendlyness' is worth it in this case. Happy to be proven wrong via an issue with lot's of upvotes.

PR #6997, opened 2016-05-07
Make setDefault compatible with object type argument

### LINKED ISSUE #6992: ReactiveDict.setDefault does not allow Object argument like ReactiveDict.set
Currently using `reactive-dict` 1.1.7.  After some painful troubleshooting I realized that the `setDefault` method does not handle the case where the argument is an object like the `set` method does. The documentation for `Session` in the API docs says that:

> Set a variable in the session if it hasn't been set before. Otherwise works exactly the same as Session.set.

For this to be true, it should check whether the argument is an object and if it is, iterate over the keys calling `setDefault` on each one.

DrDanRyan:
Make the behavior of ReactiveDict.setDefault match that of ReactiveDict.set when it comes to the Object argument form.  Copied implementation strategy from set method.  Closes #6992.


laosb:
Seems the tests come back to life! Hoooooray!

PR #7015, opened 2016-05-09
Support for Facebook login from Mobile

jasper-lu:
Just reopening this for review.

Basically, the big issue was that when I tried to add Facebook login to a mobile app I was working on (which used Meteor as the server), I couldn't do it without rewriting the accounts-facebook package, because meteor loginWithOAuth only accepted short-term tokens. My change would allow the server to accept long-term tokens too, which would allow mobile log in.

In short, mobile clients such as Facebook only send long-term tokens for mobile login. Meteor as of the first pull request only accepted short term tokens.

@tmeasday  #3515


jasper-lu:
I'll get on working on the test commands and checking if this works with the current version of Meteor asap.

tmeasday:
@jasper-lu can you rebase this to be off of current `devel`? I think this branch is off a very old commit before we had automated tests. It should be as easy as:

```
git fetch
git rebase upstream/devel
git push --force
```

jasper-lu:
Got it. I will work on it tonight.

abernix:
@jasper-lu Any updates? :smile:

jasper-lu:
@abernix I just moved it over. Still needs some testing though.

packages/facebook/facebook_server.js
```diff
@@ -50,16 +50,32 @@ var getTokenResponse = function (query) {
...
-      "https://graph.facebook.com/v2.2/oauth/access_token", {
-        params: {
-          client_id: config.appId,
-          redirect_uri: OAuth._redirectUri('facebook', config),
-          client_secret: OAuth.openSecret(config.secret),
-          code: query.code
-        }
-      }).content;
+    // Request an access token, or request access token info if it was already given
+    // DDP clients should make a post request sending the accessToken before logging in with oauth credentialtoken/secret
```
lorensr: What is the post request? Is that the first `HTTP.get` to `/app` below, or something additional that needs to be done that's not here?

zol:
Hey @jasper-lu . I'd like to merge this however I'm concerned that there may be issues/edge cases in the code that we won't catch without more testing. Could you please open an issue describing the problem this solves and link it to the PR. The intent is to get other people who need this feature to raise their hands and help test your changes.

packages/facebook/facebook_server.js
jasper-lu: Refer to https://github.com/kenyee/android-ddp-client/blob/master/src/com/keysolutions/ddpclient/android/DDPStateSingleton.java

The purpose of this is to support logging in with OAuth for mobile clients. For mobile clients, the post request sending accessToken needs to be done before logging in. (Refer to loginWithOAuth and sendOAuthHTTPRequest)

PR #7016, opened 2016-05-09
Bump Circle CI parallelism to 8x

benjamn:
I've already increased the number of containers from 4 back to 8, but the `scripts/ci.sh` script needs to be updated to take advantage of the new resources.

Submitting this as a pull request not only because we just turned on branch protection for `devel` but also because it seems like a good idea to re-run the test suite after making significant changes to the test infrastructure.

cc @tmeasday @zol @stubailo

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -10,21 +10,39 @@ case $CIRCLE_NODE_INDEX in
 0)
   echo "Running test-packages"
   echo "Running self-test (1): A-Com"
-  ./packages/test-in-console/run.sh && ./meteor self-test --file "^[a-b]|^c[a-n]|^co[a-l]|^compiler-plugins" --exclude "$SELF_TEST_EXCLUDE"
+  ./packages/test-in-console/run.sh
+  ./meteor self-test --headless \
+      --file "^[a-b]|^c[a-n]|^co[a-l]|^compiler-plugins" \
+      --exclude "$SELF_TEST_EXCLUDE"
   ;;
 1)
   echo "Running self-test (2): Con-K"
   echo "Running self-test (3): L-O"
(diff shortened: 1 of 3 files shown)


tmeasday:
LGTM.

zol:
Lgtm, although I don't love that Console.disableSpinner(); is commented out as that breaks the `--headless` flag that you introduced in this PR (if I understand the code correctly). I would prefer `Console.disableSpinner();` to be commented back in and the `--headless` arguments to self test be removed as the temporary solution.

abernix:
Can `circle.yml` just be changed to add a higher timeout?  They support it like this, I believe:

```
test:
  override:
    - ./scripts/ci.sh :
        parallel: true
        timeout: 1800
```

I am highly offended by what that spinner does on CI logs.

benjamn:
Hah, yeah, I tried [disabling the spinner](https://github.com/meteor/meteor/commit/b3bcdbafda77322c5994516dd917955ac7f18706), but then tests failed because sometimes there wasn't any "output" for too long (10 minutes I believe). So we're [back](https://github.com/meteor/meteor/commit/a9d95b2a452fb5b67c5a135dc26a7e4a61d6da4f) to
```
                                             -
                                              
                                             -
                                              
                                             -
                                              
                                             \
                                              
                                             \
                                              
                                             \
                                              
                                             \
                                              

                                             \
                                              
                                             |
                                              
                                             |
                                              
                                             |
                                              
                                             \
                                              
                                             |
                                              
                                             |
                                              
                                             /
                                              
                                             -
                                              
                                             -
                                              
                                             \
                                              
                                             \
                                              
                                             |
                                              
                                             |
                                              
                                             |
                                              
                                             |
                                              
                                             /
                                              
                                             /
                                              
                                             \
                                              
                                             \
                                              
                                             \
                                              
                                             |
                                              
                                             |
                                              
                                             /
                                              
                                             /
```
etc.

abernix:
Ad nauseum!

I'll try a new circle.yml with timeout and no spinner sometime this weekend.

Issue #4439, opened 2015-05-23
Move docs to a separated repo


laosb:
***2 Upvotes*** Meteor is a good platform for both production and education, so I decided to translate Meteor Docs into Chinese, it will be easier for China developers, even students, to play with Meteor. The problem is that docs exist as a folder in the main repo, so if I need to translate, I need to fork the main repo, which is unnecessary for document translating.

ShashankaNataraj:
+1

stubailo:
I think the reason docs are in the same repo right now is mainly so that they can be versioned as part of the same codebase. This means if you check out a release of Meteor in Git, that will also check out the correct version of the docs.

It seems to me like there isn't much of a cost to forking the whole Meteor repo for docs translation, other than just having a lot of unnecessary files in the directory - is there something you would like to do that is currently not possible because of this repository structure?

laosb:
Due to the slow network connection between China Mainland and Github, and the MOST are the "unnecessary" files for docs translating, It's not worth cloning the whole codebase. 
To the point that "can be versioned as part of the same codebase", we can make the docs one of our example apps, so we can `meteor create --example docs` and I believe that it is much easier for those ones only need a docs app instead of a clone of Meteor.

285858315:
+1

laosb:
In this way we can have a separated issues tracker for docs, which can make the docs better.

PR #7017, opened 2016-05-09
Removed docs, pointed people at new repo

tmeasday:
See https://github.com/meteor/docs


laosb:
Ah, that will land #4439 to. Hooray!

PR #7021, opened 2016-05-10
Make setDefault compatible with object type argument

stubailo:
Rebased version of https://github.com/meteor/meteor/pull/6997


packages/reactive-dict/reactive-dict.js
```diff
@@ -79,8 +79,18 @@ _.extend(ReactiveDict.prototype, {
     }
   },
 
-  setDefault: function (key, value) {
+  setDefault: function (keyOrObject, value) {
     var self = this;
+
+    if ((typeof keyOrObject === 'object') && (value === undefined)) {
```
benjamn: These extra parentheses aren't necessary, fwiw.

packages/reactive-dict/reactive-dict.js
```diff
@@ -198,6 +208,14 @@ _.extend(ReactiveDict.prototype, {
     });
   },
 
+  _setDefaultObject: function (object) {
```
benjamn: Why make this `_private`?

tmeasday:
@stubailo are we now good to merge this? It passes!

stubailo:
Doneee

PR #7031, opened 2016-05-11
wrapAsync: Allow use of null value to signal no callback was passed

### LINKED ISSUE #7010: wrapAsync does not handle case where last non-callback argument is a function
I discovered this after trying to wrap some of the functions from the very popular npm package `async`: if the last non-callback argument to the original (unwrapped) function is a function, then the wrapped function will always behave as if a callback was passed and will not behave in the fiber 'sync' style on the server. This happens because the wrapped function looks at the last argument and if it is a function, then it will assume that it is the callback.

For example, `async.map` has a signature:   `(collection, iteratee, callback)`, where `iteratee` is a function and required parameter.  So, you cannot use the result of `Meteor.wrapAsync(async.map)` on the server in a way that makes it wait for the result. If you omit the callback, then Meteor uses `iteratee` as a callback and you end up with garbage results and no sync style wait for the results.

However, this can be hacked by wrapping it in a function that changes the function arguments, like:

``` javascript
const mapFiber = Meteor.wrapAsync(myMap);

function myMap(iteratee, collection, cb) {
    async.map(collection, iteratee, cb);
}


// Later when you want to use it, you have to use the modified argument order
const result = mapFiber(iteratee, collection);
```

This is an easy enough workaround, but I wonder if there could be some option for `Meteor.wrapAsync` that would allow direct wrapping without changing argument order.  Maybe allowing the caller to specify the number of required non-callback function arguments?  Then, `wrapAsync` would know how far back in the argument list to go looking for the callback before deciding that there was none provided on the function call.  In the `map` case above, I could say there are 2 non-callback arguments and then `wrapAsync` would never consider the first two arguments fair game for being chosen as the callback.

DrDanRyan:
Closes #7010

There is a convention in the Javascript community to use the value `null` for a function argument that is intentionally not provided. 

When a function returned by `wrapAsync` (the "wrapped function") is called, the arguments are inspected in reverse order to determine if a callback was provided. If the last argument (that is not undefined) is a function, it is used as the callback.  If the last argument is not a function, then the call is "fiberized" by **appending** a `future.resolver()` to the arguments to be used as a callback in the inner function call.

This behavior is not ideal because there is no way to correctly call a wrapped function when the last non-callback argument is also a function. There are **many** examples of this in the `async` npm library: https://github.com/caolan/async. With the current implementation, replacing the callback argument with `null` in the wrapped function call results in the inner function being called with `null` for the callback and the `future.resolver` is never touched.

I think that if the last (not undefined) parameter in the wrapped function call is `null`, that the inner function should be called with the `future.resolver` **replacing** the `null` value. This provides a way to correctly call any wrapped function "sync" style on the server; just pass `null` for the callback argument. This makes no difference on the client, because if the last argument is not a function, an error is thrown; and, these "sync" style calls are only supposed to be made on the server anyway.

The only break of existing code I can see is when the last non-callback argument is the value `null` and the callback parameter is simply omitted (because that is how you have to do it in the current version).  This shouldn't be very common, and it can be fixed by adding an extra `null` value for the callback parameter.

I think the case where the last non-callback argument is a function is probably much more common and currently there is no call signature to fix that.

Providing the `null` option for the callback makes it so the wrapped function can have any signature and still be called in "sync" style of "async" style on the server.

### DIFF
--- a/packages/meteor/helpers.js
+++ b/packages/meteor/helpers.js
@@ -82,7 +82,7 @@ _.extend(Meteor, {
 
   /**
    * @memberOf Meteor
-   * @summary Wrap a function that takes a callback function as its final parameter. The signature of the callback of the wrapped function should be `function(error, result){}`. On the server, the wrapped function can be used either synchronously (without passing a callback) or asynchronously (when a callback is passed). On the client, a callback is always required; errors will be logged if there is no callback. If a callback is provided, the environment captured when the original function was called will be restored in the callback.
+   * @summary Wrap a function that takes a callback function as its final parameter. The signature of the callback of the wrapped function should be `function(error, result){}`. On the server, the wrapped function can be used either synchronously (by omitting the callback or passing the value `null` for the callback) or asynchronously (when a callback is passed). If the last argument passed to the wrapped function is a function, it will be used as the callback. On the client, a callback is always required; errors will be logged if there is no callback. If a callback is provided, the environment captured when the original function was called will be restored in the callback.
    * @locus Anywhere
    * @param {Function} func A function that takes a callback as its final parameter
    * @param {Object} [context] Optional `this` object against which the original function will be invoked
@@ -99,6 +99,8 @@ _.extend(Meteor, {
         if (type !== "undefined") {
           if (type === "function") {
             callback = arg;
+          } else if (arg === null) {
+            --i; // replace the null value with callback in the function call
           }
           break;
         }


stubailo:
The tests failed, restarted the build just in case it's a temporary thing. But it could be a real failure, in which case we should fix it before it can be merged.

zol:
@benjamn this looks ok to me but I'd like to get your eyes on it before merging it.

benjamn:
This is long-standing legacy code that many developers are using in their apps and packages, so I think we should consider it frozen. But I will admit this change seems useful, and you should definitely feel free to reimplement `Meteor.wrapAsync` in this way, for your own purposes.

PR #7036, opened 2016-05-11
Speed up selftest

tmeasday:
Use symlinking to speed up selftest init times


tmeasday:
This passed in 35 minutes -- I think we could distribute the tests better too: https://circleci.com/gh/meteor/meteor/1577

PR #7039, opened 2016-05-12
[tests] Change destination IP for http error test to something more invalid

### LINKED ISSUE #7038: [selftest] reconnect auto-login test fails on OS X Server
Opening this to track self-test issue on OS X Server as found by @Fabs in #6957:

> get the error `accounts C: reconnect auto-login -  fail — fail - message Issue #4970 has occured.`. I looked #4970 and its closed. It happens with the default tarball mongodb driver, on devel. As it does the httpcall error I mentioned above. 
> I took the time now to read the test carefully, I think maybe I know what the issue could be. I have OSX Server, and the address http://0.0.0.0/ resolves an webpage (apache is running, port 80).
> So the assumption behind the test `HTTP.call("GET", "http://0.0.0.0/", expect(unknownServerCallback));` does not hold on my machine, and thus the test fail. When I turn the server off, then the test stop failing. http://0.0.0.0:3000/ also resolves to my meteor server.

Tom wrote:
> Whoa, `0.0.0.0` wasn't a great choice! Is there a known "bad" IP address? We should change the test to use one that doesn't resolve to localhost (I'm surprised this hasn't come up before). Perhaps `240.0.0.0` is a better choice? https://tools.ietf.org/html/rfc5735

I write:

Yea, `0.0.0.0` is definitely no good.  I would propose `198.51.100.99` as it should factually never be used on any network – and is fine to use for a check-down type test.  `240.*` still could get some use (though unlikely).

I'll drop a PR.

abernix:
Previously 0.0.0.0 was used, but that _can_ refer to the local machine and was causing problems when the test was run on a machine that had something (e.g. Apache, etc.) listening on 0.0.0.0:80.

Fixes #7038

### DIFF
--- a/packages/http/httpcall_tests.js
+++ b/packages/http/httpcall_tests.js
@@ -90,15 +90,16 @@ testAsyncMulti("httpcall - errors", [
       test.isFalse(error.response);
     };
 
-    // 0.0.0.0 is an illegal IP address, and thus should always give an error.
+    const invalidIp = "0.0.0.199";
+    // This is an invalid destination IP address, and thus should always give an error.
     // If your ISP is intercepting DNS misses and serving ads, an obviously
     // invalid URL (http://asdf.asdf) might produce an HTTP response.
-    HTTP.call("GET", "http://0.0.0.0/", expect(unknownServerCallback));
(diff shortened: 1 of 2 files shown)


tmeasday:
Thanks @abernix

Issue #7042, opened 2016-05-12
Support for smtps:// MAIL_URL


MichaelJCole:
Ok, couple of things here:

1) Meteor uses port 465 to indicate secure connection.  [Port 465 was never a standard port and was assigned to something else](http://blog.mailgun.com/25-465-587-what-port-should-i-use/).

2) I'm getting an email transport error similar to this one: [Greeting not received]( https://github.com/nodemailer/nodemailer/issues/441).   Related to #5284

My email provider (Sparkpost is the new Mandrill) wants to use STARTTLS on port 587.  I don't see a way to configure that in Meteor, we just have MAIL_URL.

"smtps" appears to be a common protocol, ([see first example](https://github.com/nodemailer/nodemailer)), but Meteor doesn't allow it ([email code](https://github.com/meteor/meteor/blob/devel/packages/email/email.js#L20-L36)), and only uses secureConnection if the port is 465. 

This seems like a bug, but maybe Meteor makes this choice for a reason?  What's the best way to configure Meteor to use STARTTLS on port 587?

Attached PR sets secure connection if 
```
{ secureConnection: (port === 465) || (mailUrl.protocol === 'smtps:'),
```

This would be a non-breaking change, that is more standards based.

Thanks!

Mike

abernix:
My initial impression is that there is something missing in your PR.   The code snippet you put above:
```
{ secureConnection: (port === 465) || (mailUrl.protocol === 'smtps:'),
```
...is not in the PR.  Will comment more in the PR.

abernix:
Closing as this should have been resolved with PR #7043 which was merged and should be in Meteor as of 1.3.3.1.

PR #7043, opened 2016-05-12
Allow smtps:// in MAIL_URL to enable secureConnection

MichaelJCole:
This backwards compatible patch allows smtps protocol, and secureConnection on port 587.  Port 465 is non-standard and assigned to something else.  See #7042 for details.

### DIFF
--- a/packages/email/email.js
+++ b/packages/email/email.js
@@ -17,9 +17,9 @@ var MailComposer = EmailInternals.NpmModules.mailcomposer.module.MailComposer;
 
 var makePool = function (mailUrlString) {
   var mailUrl = urlModule.parse(mailUrlString);
-  if (mailUrl.protocol !== 'smtp:')
+  if (mailUrl.protocol !== 'smtp:' && mailUrl.protocol !== 'smtps:')
     throw new Error("Email protocol in $MAIL_URL (" +
-                    mailUrlString + ") must be 'smtp'");
+                    mailUrlString + ") must be 'smtp' or 'smtps'");
 
   var port = +(mailUrl.port);
   var auth = false;
--- a/packages/email/email.js
+++ b/packages/email/email.js
@@ -33,7 +33,7 @@ var makePool = function (mailUrlString) {
   var pool = simplesmtp.createClientPool(
     port,  // Defaults to 25
     mailUrl.hostname,  // Defaults to "localhost"
-    { secureConnection: (port === 465),
+    { secureConnection: (port === 465) || (mailUrl.protocol === 'smtps:'),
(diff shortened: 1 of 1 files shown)


MichaelJCole:
Please see #7042

packages/email/email.js
```diff
@@ -17,9 +17,9 @@ var MailComposer = EmailInternals.NpmModules.mailcomposer.module.MailComposer;
 
 var makePool = function (mailUrlString) {
   var mailUrl = urlModule.parse(mailUrlString);
-  if (mailUrl.protocol !== 'smtp:')
+  if (mailUrl.protocol !== 'smtp:' && mailUrl.protocol !== 'smtps:')
     throw new Error("Email protocol in $MAIL_URL (" +
-                    mailUrlString + ") must be 'smtp'");
+                    mailUrlString + ") must be 'smtp' or 'smtps'");
```
abernix: This only changes whether or not an error is thrown when `smtps` is used.  It doesn't change `secureConnection` below.

abernix:
This doesn't appear to change the `secureConnection` value, something must be missing.

abernix:
You had this code in the Issue
```
{ secureConnection: (port === 465) || (mailUrl.protocol === 'smtps:'),
```
...but it's not here in the PR.  This makes the difference between this change working and not working by allowing the additional trigger for `secureConnection`, which would definitely be necessary if the provider didn't support 465.

Pretty strange for SparkPost to just not support it though. I'm aware it is not RFC and it's silly, but it is widely accepted by many (Gmail, Hotmail, Mandrill, SendGrid, MailGun, AWS SES, etc.).  Looks like Postmarkapp doesn't support it either.

Want to update this PR?

MichaelJCole:
@abernix Hey Jesse, there are steps in git, and I missed one.  Added the 2nd commit. Good catch.

Yeah, this PR didn't fix the SparkPost issue.  I'm in a hurry and switched to Mailgun instead.  the 'simpleSMTP' npm package is depricated.  Next steps for Sparkpost issue would be to switch to a new mail transport library, but that's a bigger choice than a drive-by PR :-)

carinakoo:
@MichaelJCole, thanks for this leg work. I'm experiencing this as well with Sparkpost. Do I understand correctly then that there's no known fix for Sparkpost? Enabling `secureConnection` did not solve? Thanks.

MichaelJCole:
Hey @carinakoo I was in a hurry and switched to mailgun with no problems yet.  You might get a better response from Sparkpost because you're not the first.  Give their support a try.

zol:
Looks great, thanks everyone!

PR #7062, opened 2016-05-16
Put cordova.js behind other js file

feit:
When i build cordova app with --production flag, open the app it throw jQuery is not defined, because cordova.js was load before jquery.

### DIFF
--- a/packages/boilerplate-generator/boilerplate_web.cordova.html
+++ b/packages/boilerplate-generator/boilerplate_web.cordova.html
@@ -24,7 +24,6 @@
     }
   </script>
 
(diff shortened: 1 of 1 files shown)


abernix:
Thanks for submitting this.  Pull requests are very important to the Meteor project, however it's very unlikely this will be considered without an Issue attached to it along with a clear reproduction of the problem you are experiencing and proof that it is in actual bug.

Please open a separate issue with a Git repository link to a reproducible example of your problem and make reference to this PR. See more about contributing here: https://github.com/meteor/meteor/blob/devel/Contributing.md

PR #7063, opened 2016-05-16
Increase circleci timeout and disable spinner for CI tests

abernix:
Trying to see if re-enabling the `--headless` spinner that Ben put in place will work if combined with a longer timeout for CircleCI.

...about to find out.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,3 +10,4 @@ test:
   override:
     - ./scripts/ci.sh :
         parallel: true
+        timeout: 1800
--- a/circle.yml
+++ b/circle.yml
@@ -10,4 +10,4 @@ test:
   override:
     - ./scripts/ci.sh :
         parallel: true
-        timeout: 1800
+        timeout: 1200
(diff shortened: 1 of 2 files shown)


abernix:
This worked well – with the exception of the (separate) spinning in the package tests (container 0) the output is significantly more compact on containers 1-7.  The default timeout was 600 on CircleCI – this first test raised it to 1800.  Going to split the difference with 1200 and see how that goes.

abernix:
1200 seconds works!  ✅

I'm not going to bother trying less than that at the moment.  I'd say this is a huge improvement already since it at least provides readable logs for 7/8 containers.  The next step would be making a similar `--headless` flag work with `test-packages`.

benjamn:
I agree this is a huge improvement!

benjamn:
If https://github.com/meteor/meteor/commit/05dfe244382a927fb1284162c3df0d38570a6d2b, I will cherry-pick it onto `devel`.

abernix:
👍 My avoidance to doing this was that CI might *never* end in some bizarre cases? (resulting in queue galore).  Haven't seen runaways yet though, so have another 👍

PR #7066, opened 2016-05-16
Minimongo: Allow `_id` in `$setOnInsert`.

gsuess:
Mongo now allows specifying an `_id` in upserts using the `$setOnInsert` operator, but MiniMongo throws an error. This makes simulations for latency compensation very cumbersome.

However MiniMongo does support setting _id s with `$setOnInsert` as it stands now. Only the exception was thrown incorrectly.

So I have added the tests to the relevant test case and fixed the condition on throwing the exception.

Steps to reproduce: 

```JavaScript
let collection = new LocalCollection;
collection.upsert({foo: 'bar'}, {$setOnInsert: {_id: 'test'}});
```

Expected result:
```JavaScript
isEqual(collection.findOne(), {
  _id: 'test',
  foo: 'bar'
});
```

Actual result:

> Error: MinimongoError: Mod on _id not allowed [409]

### DIFF
--- a/packages/minimongo/minimongo_tests.js
+++ b/packages/minimongo/minimongo_tests.js
@@ -2085,6 +2085,23 @@ Tinytest.add("minimongo - modify", function (test) {
     exceptionWithQuery(doc, {}, mod);
   };
 
+  var upsert = function (query, mod, expected) {
+    var coll = new LocalCollection;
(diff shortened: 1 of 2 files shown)


tmeasday:
Thanks @gsuess!

PR #7072, opened 2016-05-16
[Work in progress] Release 1.5

benjamn:
This pull request is an easy way to collect changes that will be going into Meteor 1.5 and run tests against them in an isolated, consistent environment (CircleCI and TravisCI).

Meteor 1.5 will be the first version of Meteor that is installable from npm, meaning that you can choose which version of Node you use, and you can use the Meteor build tool as a dev dependency of any npm package. That doesn't mean there won't be an easy way to install Meteor with all the prerequisites you might want; it just means you will have the option of providing your own versions of Node, npm, and possibly Mongo, if you need that flexibility.

Milestone: https://github.com/meteor/meteor/milestones/Release%201.5


tools/cli/dev-bundle-bin-helpers.js
```diff
@@ -41,3 +56,31 @@ exports.getEnv = function () {
...
+  require("./flush-buffers-on-exit-in-windows.js");
+
+  return child;
+};
+
+exports.npmInstall = function () {
+  var nodeModulesPath = path.join(__dirname, "..", "..", "node_modules");
+  try {
+    var stat = fs.statSync(nodeModulesPath);
+    if (stat.isDirectory()) {
```
benjamn: If `meteor-tools` is installed with npm v3, it's entirely possible that it won't have a `node_modules` directory because all of its dependencies were hoisted to the top level, so this check will have to be improved.

stubailo:
Super pumped for this!

tomtom87:
cant wait!!

RobertLowe:
It's happening

So excited! 👍

D1no:
Just to clarify
> you **can** use the Meteor build tool as a dev dependency

Does that mean (with some reasonable effort) the meteor client and server can be build / picked up by other build tools such as grunt, webpack etc. or simply dropped into an html file?

Because this sounds like a limitation:
> it just means you will have the option of providing your own versions of Node, npm, and possibly Mongo

Of course, the main reason why many people (including me) are so incredible excited about this announcement is, that it is currently **very hard to drop meteor client files** into an **existing build chain**.

For example: We have a project where we build a range of browser extensions (chrome, Safari, etc) and need to use a custom build chain (webpack) to have a reasonable workflow. It would be a dream to npm install meteor client files and connect them to the running meteor server.

`npm install meteor-client --save`

```js
import {Meteor, Mongo, Tracker} from 'meteor-client';

Meteor.connect('http://localhost:3000').onReady(()=> {
// runtime magic
});

// ....

if(Meteor.isConnected) {
// I am a progressive Web-App
};
```

❤️

benjamn:
Though I appreciate the positive response to this PR, it's pretty far out of date at this point. I am still excited about the possibility of installing Meteor from `npm`; however, the current plan for Meteor 1.5 is to focus on Apollo/GraphQL integration, in order to (among other goals) support databases other than Mongo. I'd be happy to continue talking about the `npm` idea, but it feels incorrect to put the label "Meteor 1.5" on that discussion.

D1no:
I personally see this as a huge disappointment. Meteor has a legacy with residing teams, waiting to fully integrate it into npm driven build chains.

Decoupling meteor is desperately necessary for most projects and also a road to freely add features and live query replacements like apollo via npm. The technical dept in getting meteor to the large JS community is just getting bigger. And I really don't like that, given prior communication.

Forum Thread
https://forums.meteor.com/t/great-news-meteor-apollo-integration-coming-in-1-5/26494/6

AndreasGalster:
What about packages in general? Will we at least see a minor move of existing packages to npm?

I still feel that it is too much of russian roulette getting unwanted packages installed with Meteor. I also would prefer npm before Apollo. The build tools and package/dependency management with meteor is becoming more and more frustrating. We need a better way to analyze which packages end up in our app, currently with meteor packages it is not transparent enough where dependencies are coming from. I don't necessarly thing meteor-core needs to be npm just yet but maybe at least put some of the things from core slowly on npm. 

Some basic examples are: Move the random package and the accounts packages over to npm. Make an incremental move rather than move one big meteor package over to npm.

tomtom87:
How do we move to npm? I'll port any i can to help out would love to use npm over atmosphere seriously would be a dream.

let me know what it entails more and I will start with a package i maintain

laosb:
Is it possible to start with some packages without many dependencies, like check? I think this transition doesn't necessarily to be done in a major version. (The second digit in meteor version number can be treated as major, I think?)

MichaelJCole:
What the heck is Apollo and why do I need it?  The only thing I can find centers on Facebook.

http://www.apollodb.com/apollo_embedded.asp
http://www.i-programmer.info/news/84-database/7441-facebook-apollo-nosql-database.html

Is Meteor going to replace node with something?

jedwards1211:
I have a solution for using Meteor and React without running your app through Meteor dev tools, which means you can use whatever version of node you want: https://github.com/jedwards1211/crater
It also has Webpack hot reloading, React SSR, and allows you to customize the Webpack and Babel config.

Issue #7064, opened 2016-05-16
`meteor mongo` only works with built-in mongo


nicodmf:
Versions:
Meteor: 1.3.2.4
Configuration in development mode : 
- one docker container for meteor
- one docker container for mongodb (3.0)

The `meteor mongo` command doesn't work even with a right $MONGO_URL variable in the environenment. After reading the code : 

https://github.com/meteor/meteor/blob/abd574f38008b45f5e2a6bc322b10bcdde44763a/tools/cli/commands.js#L1134

It's seems that meteor mongo could take an argument, but even with an argument, this doesn't work for dark reasons (Sorry, Meteor does not yet support specific path URLs,).

But with meteor shell, i can use the database and insert document.

As docker is even a development system that a production system, and anyway that to have a mongodb in different server is a usually process, i think the meteor mongo should be upgraded tio work with a distant database.

At less the message should be changed.

Best Regards

nicodmf:
The more angriest with this command is that if i change code and indicate my server this works...

abernix:
You can also just use `mongo mongodb://..` but I suppose it *might* make sense for the `meteor-tool` to be able to connect as well.

I wonder if it wouldn't be best to use the same functionality as `meteor npm` though and simply proxy the request to the built-in `mongo` so you could do `meteor mongo mongodb://...`.  If it can discover the local server's `MONGO_URL` or it's set in the ENV, then it could set it for you.

But then there's the question of what to do with the rest of the mongo commands: `mongodump`, `mongoexport`, `mongoimport`, `mongorestore`, `mongotop`, etc., etc., etc. ?

abernix:
The expectation right now is that if you're running your own Mongo server and setting it with `MONGO_URL`, you know how to use the built-in `mongo` tools directly.

laosb:
The argument is used by Meteor Free Hosting, and I'm not sure if it still works on Galaxy. Back to that time it's pretty useful to do `meteor mongo xxx.meteor.com`.

abernix:
@laosb That hasn't changed though, right?  I don't see why that would have stopped working.  This is basically asking to make it possible to connect `meteor mongo` to non-Galaxy, non-Free-Hosting, non-built-in Mongo.

nicodmf:
As i say on #7068 : I'm not sure that is just a feature, for me at least at beginning this was more a bug or a correction. The process is not clear :

- you can connect distant db on deployment
- but not near db in dev

This is at least a strange effect.

[...]

PR #7077, opened 2016-05-18
`meteor mongo` can detect and use MONGO_URL now.

laosb:
#7064 

This pr simply takes MONGO_URL to run mongo shell if in a app without builtin mongo started.

Should have tests, but I'm still learning how to do it. Is there anyone who are willing to write it?


tmeasday:
Thanks @laosb, I don't think we want to take this. Is there a specific use case here that's not solved by installing Mongo directly on your system?

We'd prefer not to expose the internals of the dev bundle if we can avoid it. For instance, in 1.4 we will ship a version of `mongo` without SSL support because we don't need it. Accepting this would make that more painful.

PR #7080, opened 2016-05-18
Merge 4235

tmeasday:
***1 Upvote*** See #4235


ryneeverett:
:+1: Haven't looked at this code in a while but your change makes sense to me.

PR #7085, opened 2016-05-19
Fix incorrect History entry about Match.Maybe and Match.Optional

### LINKED ISSUE #7083: CHANGELOG is confusing on v1.3.2, regarding change in Match.Optional
In the [changelog for v1.3.2](https://github.com/meteor/meteor/blob/devel/History.md#v132), we can read:

> Match.Optional only passes if the value is null or the specified type, whereas previously it accepted undefined. Use Match.Maybe to allow undefined. #6735

Foremost, this seems like a huge breaking change, which scared us off a lot – we've got `Match.Optional()` sprinkled all over, catch undefined arguments to functions. But after some investigations, it seems like this is not the case?

This is some tests in `meteor shell` with an app on 1.3.2.4:

```
> Match.test(null, Match.Optional(String))
false
> Match.test(undefined, Match.Optional(String))
true # <== This contradicts the changelog
> Match.test(null, Match.Maybe(String))
true
> Match.test(undefined, Match.Maybe(String))
true
```

So if I'm not heavily mistaken, it's still possible for `Match.Optional()` to match and pass on `undefined` in 1.3.2.x. This might have been "fixed" in a later version than 1.3.2.

Here's the docs on `Match.Optional`, being correct:

> Behaves like Match.Maybe except it doesn’t accept null. If used in an object, the behavior is identical to Match.Maybe.

https://docs.meteor.com/api/check.html#matchpatterns

I suggest either clarifying – negating – the changelog paragraph above, or adding a new note to a point release after 1.3.2.

abernix:
The log entry about `Match.*` bugs in 1.3.2 is actually opposite of what was changed.  This clarifies it and adds an entry for the original addition of `Match.Maybe` to the 1.3 log.

This was my mistake for not submitting a `History.md` entry with my fix in #6736 and leaving it (a confusing situation) to be interpreted by the person who wrote the History.  I didn't even notice it was wrong the second time I read it!

Fixes #7083

### DIFF
--- a/History.md
+++ b/History.md
@@ -68,6 +68,8 @@
 
 * Fixed bugs: https://github.com/meteor/meteor/milestones/Release%201.3.2
 
+* Fixed unintended change to `Match.Optional` which caused it to behave the same as the new `Match.Maybe` and incorrectly matching `null` where it previously would not have allowed it. #6735 
+
 ## v1.3.1
 
 * Long isopacket node_modules paths have been shortened, fixing upgrade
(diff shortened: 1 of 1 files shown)


abernix:
Opening a new PR with a single commit and the correct fix. 👎

PR #6964, opened 2016-05-03
Allow --port in "test" to support a bindaddr, same as non-test

abernix:
The `test` mode wasn't passing the proper proxyHost and proxyPort values to runAll.  This changes the behaviour to work the same as the regular `run` mode and also adds tests for both modes to test the bindaddr (localAddress) is taking effect properly.

Fixes #6961

PR #7103, opened 2016-05-23
Adds changelog for PR #6964

abernix:
Adds changelog for PR #6964, (`--port IP:ADDR` support on `meteor test`), which is included in 1.3.3.


tmeasday:
Thanks @abernix

Issue #6957, opened 2016-05-02
Update to Mongo 3.2


tmeasday:
### PR: https://github.com/meteor/meteor/pull/7163

This issue is an attempt to outline what needs to be considered as part of this transition. We [plan](https://github.com/meteor/meteor/blob/devel/Roadmap.md#mongodb-updates) to look at this in the short/medium term, however if someone is keen to contribute to the project and wants to see it land earlier, this would be a great opportunity to make a larger contribution!

### Update Mongo Driver to 3.2 compatible version https://github.com/meteor/meteor/issues/5763

 - [x] We need to update the version of the `node-mongodb-native` package used here: https://github.com/meteor/meteor/blob/devel/packages/npm-mongo/package.js to at least 2.1.0 (I guess we may as well get the latest 2.1.18).

 - [x] Check our uses of the driver in https://github.com/meteor/meteor/blob/devel/packages/mongo/ and other core code. Supposedly the API is backwards compatible, although there are some small issues mentioned here: https://github.com/meteor/meteor/issues/5763#issuecomment-163055637

 - [x] Ensure tests pass

(We should also probably remove the `npm-mongo` wrapper package and just peer-depend on the npm package from `mongo`, but I think we should wait to do this, it'll be a larger issue we have to solve with the 1.4 move to npm).

### Update bundled Mongo in the dev bundle to 3.2 https://github.com/meteor/meteor/issues/5809

 - [x] Update our fork of Mongo [https://github.com/meteor/mongo] to 3.2 (there has been work done already in this direction: https://github.com/meteor/mongo/tree/ssl-r3.2.0). This fork allows us to statically link Mongo. (**Note we didn't actually do this, using vanilla**)
 
 - [x] Update https://github.com/meteor/meteor/blob/devel/scripts/build-mongo-for-dev-bundle.sh to build + include Mongo 3.2. (**Note we didn't actually do this, using vanilla**)

 - [x] Update generate-dev-bundle scripts

 - [x] Ensure that if you run the 3.2 Mongo server against an (older) 2.6 local database, it gives a meaningful error that suggests that you `meteor reset` or look into taking steps to upgrading your db contents

 - [x] Figure out if we need to turn journalling off or if there's a way to get the minimum journal size down.

### Errata

 - [x] Update https://github.com/meteor/meteor/blob/devel/scripts/admin/test-packages-with-mongo-versions.rb to also test 3.2

[...]

PR #7104, opened 2016-05-23
Mongo 3.2 fix oplog

Fabs:
On the path to finish #6957.

- Fix oplog connection
- Fix a test that had too high limits for the new driver


tmeasday:
@Fabs I'm not able to run the self-tests or even start a Meteor app right now on this branch or the `mongo-3.2` branch. I think it's because we need to make similar changes to this PR to `run-mongo.js` around here: https://github.com/meteor/meteor/blob/mongo-3.2/tools/runners/run-mongo.js#L593-L652

I can take a look but I suspect you have more context on this?

Fabs:
Also. I ran this branch against some projects, and the only think breaking so far is `kadira` (They use an internal object Connection on the driver, that is no longer available). I gave them a heads up.

benjamn:
Anything left to do here before we merge this into https://github.com/meteor/meteor/tree/mongo-3.2?

tmeasday:
I'll need to look over @Fabs' latest changes @benjamn, and do some more testing. Assuming it is good I think we are more or less done

tools/runners/run-mongo.js
```diff
@@ -584,35 +584,41 @@ var launchMongo = function (options) {
...
-        var status = yieldingMethod(db.admin(), 'command',
-                                    {replSetGetStatus: 1});
-        if (!(status && status.documents && status.documents[0])) {
+        var status;
+        try{
+          var status = yieldingMethod(db.admin(), 'command',
+                                      {replSetGetStatus: 1});
+        } catch (e) {
+          //Some of the expected results, come as errors.
+          status = e;
```
tmeasday: Does the `startupStatus === 6` error come through as an error or a return value? If the second, I think we should just let the original error through here, rather than catching and then throwing again later.

tools/runners/run-mongo.js
```diff
@@ -584,35 +584,41 @@ var launchMongo = function (options) {
...
+        }
+
+        if (!status) {
           throw status;
```
tmeasday: We can't throw `status` if it's undefined.

tools/runners/run-mongo.js
```diff
@@ -584,35 +584,41 @@ var launchMongo = function (options) {
         });
       }
 
-      var initiateResult = yieldingMethod(
-        db.admin(), 'command', {replSetInitiate: configuration});
+      var initiateResult;
+      try{
```
tmeasday: More small formatting things :)

tools/runners/run-mongo.js
Fabs: I see what you mean. I strongly believe it comes as an error though, as per [the documentation on ReplicaSet States](https://docs.mongodb.com/manual/reference/replica-states/), UNKNOWN(6) state is an Error State.

Fabs: Also, very good we had that discussion because `startupStatus` is now `myState`.

tools/runners/run-mongo.js
Fabs: Ups!

tools/runners/run-mongo.js
```diff
@@ -607,38 +607,43 @@ var launchMongo = function (options) {
...
         }
-        status = status.documents[0];
+
+        // See https://docs.mongodb.com/manual/reference/replica-states/
+        // for information on various states
         if (!status.ok) {
-          if (status.startupStatus === 6) {  // "SOON"
+          // UNKNOWN(6) some of the other replicas are in an UNKNOWN state from
+          // this replica perspective.
+          if (status.myState === 6) {
```
tmeasday: I'm not sure we should include this code unless we can reproduce the code path. 

It looks like the original test was included in this commit: https://github.com/Fabs/meteor/commit/1d5da6b4390512857f222f3e8959dc5dda505093 used by this commit: https://github.com/Fabs/meteor/commit/72c9657b0e8a7b209ded1df67f22dd80e2fab0a2

You can run that test with `./meteor self-test 'mongo failover'`. Perhaps you should try running that and see if we can nail this down.

tools/runners/run-mongo.js
```diff
@@ -607,38 +607,43 @@ var launchMongo = function (options) {
...
-                                    {replSetGetStatus: 1});
-        if (!(status && status.documents && status.documents[0])) {
-          throw status;
+        var status;
+        try{
+          status = yieldingMethod(db.admin(), 'command',
+                                      {replSetGetStatus: 1});
+        } catch (e) {
+          // Some of the expected results, come as errors.
+          status = e;
```
tmeasday: I don't really like this pattern (as a reader of the code I might not expect `status` to be an error).

I'd rather the code looked like:

```js
try {
  status = ...
} catch (e) {
   if (e.myState === 6) {
     // handle non-fatal error
   } else {
     throw e;
   }
}

// now we know status is a status record.

File: tools/runners/run-mongo.js [hunk 6]
```diff
@@ -607,38 +607,43 @@ var launchMongo = function (options) {
         });
       }
 
-      var initiateResult = yieldingMethod(
-        db.admin(), 'command', {replSetInitiate: configuration});
+      try{
```
tmeasday: Can we please fix the formatting in this diff? (whitespace mainly)

tools/runners/run-mongo.js
Fabs: That is good you pointed me to that test. There is also some problems in the test itself regarding the driver changes.

I confess I was a little shy of running self-tests. No longer!! I have seen it happening once or twice while I was developing, but I will make sure that the mongo tests stress this one code branch or that I add a test that does (or remove it, but I think its necessary).

Fabs: Meanwhile I will just commit on the other issues you mentioned.

Fabs: Meanwhile I will just commit on the other issues you mentioned.

tools/runners/run-mongo.js
```diff
@@ -607,38 +607,37 @@ var launchMongo = function (options) {
...
-        if (!status.ok) {
-          if (status.startupStatus === 6) {  // "SOON"
+        var status;
+        try{
+          status = yieldingMethod(db.admin(), 'command',
+                                      {replSetGetStatus: 1});
+        } catch (e) {
+          // See https://docs.mongodb.com/manual/reference/replica-states/
+          // for information on various states
+          if (e.myState === 6) {
```
Fabs: Just so it's known I will still take a better look on this.

Issue #7123, opened 2016-05-26
[1.3.3-beta.2] Email.send not working unless headers object provided.


saimeunt:
Looks like this new feature : **Insert a Date header into emails by default:** https://github.com/meteor/meteor/pull/6916/files leads to a serious regression if `options.headers` is `undefined`.

https://github.com/meteor/meteor/blob/release-1.3.3/packages/email/email.js#L181

saimeunt:
BTW, quick fix would be:

```js
const emailSendCopy = Email.send;
Email.send = (options) => {
  Object.assign(options, { headers: {} });
  emailSendCopy(options);
};
```

abernix:
Thanks for reporting this, @saimeunt. I would expect it to be fixed in the next beta.

If your mail sending is all done from a single entry point, just passing an empty headers object is also an easy fix:
```js
Email.send({
  from: "foo@example.com",
  to: "bar@example.com",
  // more usual properties here
  headers: {},
});
```

abernix:
This has already been fixed with 6e071838af0c9594cdbf9b23ded7ef5806510a86 and should be in the next beta.  Also PR #7128 has been submitted which would have caught this mistake!

abernix:
Thanks for reporting this.  This should be fixed in the new beta that just came out.  You can update to it using `meteor update --release 1.3.3-beta.2`.

Thanks for reporting this and please write back if you're still having any problems!

PR #7128, opened 2016-05-27
Improve Tests for the `email` package

abernix:
* Added `smokeEmailTest` function which sets up test stream
* Splits the previously combined tests into multiple tests
* Created test that catches error in PR #6916, caught in #7123
* Confirmed regression without 6e071838af0c9594cdbf9b23ded7ef5806510a86

Per @tmeasday in https://github.com/meteor/meteor/pull/6916#issuecomment-221953805

### DIFF
--- a/packages/email/email_tests.js
+++ b/packages/email/email_tests.js
@@ -1,12 +1,25 @@
 var streamBuffers = Npm.require('stream-buffers');
 
-Tinytest.add("email - dev mode smoke test", function (test) {
+var devWarningBanner = "(Mail not sent; to enable " +
+  "sending, set the MAIL_URL environment variable.)\n";
+
+function smokeEmailTest(testFunction) {
   // This only tests dev mode, so don't run the test if this is deployed.
   if (process.env.MAIL_URL) return;
 
   try {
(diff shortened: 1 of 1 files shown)


tmeasday:
Looks great, thanks @abernix !

packages/email/email_tests.js
```diff
@@ -37,7 +49,57 @@ Tinytest.add("email - dev mode smoke test", function (test) {
...
+    });
+
+    test.matches(stream.getContentsAsString("utf8"),
+      /^====== BEGIN MAIL #0 ======$[\s\S]+^To: bar@example.com$/m);
+  });
+});
+
+Tinytest.add("email - multiple e-mails same stream", function (test) {
+  smokeEmailTest(function (stream) {
+    // Test if date header is automaticall generated, if not specified
```
AlexanderS: This comment does not look correct. The date header is not checked here and the tests checking the auto generated date header is further down and has the same comment.

abernix: @AlexanderS You're absolutely right. This was a copy/paste mistake on my part when I was splitting the single previous test block into multiple blocks.

PR #7131, opened 2016-05-28
Use babel preset/plugin versions/hash for cache invalidation

gadicc:
Addresses point 1 in https://github.com/meteor/meteor/issues/6351#issuecomment-220944251, cc: @benjamn.

I don't have too much time this weekend but this code has been in my head the whole week and it wants out :D  As usual, completely open to change style / variable names / approach per your guidance, or for you to use just as inspiration if you'd like to go a different route.

Requires my [PRs for meteor/babel](https://github.com/meteor/babel/pulls) (I slipped in the `sourceHash` for the same trouble).  I should probably update my use of `deps` there to `cacheDeps`.  `cacheHashDeps` is even clearer but a little excessive.

Ultimately `Babel.compile(source, babelOptions, cacheDeps)` gets a `cacheDeps` like:

```js
{
  'sourceHash': '2c9f7e231af1a62db195875ca62df7edc253c3c4',
  'babel-preset-stage-0': '6.5.0',
  'babel-plugin-transform-decorators-legacy': '1.3.4',
  'react-hot-loader': '3.0.0-beta.2',
  '/home/../project/localPlugin.js': '6ea5270e40877f07f1f0e41babe7b4ffe4c7c721'
}
```

No tests :(  With a bit more time I could do some unit tests just for some coverage, but realistically it's a fairly simple change where I don't think much can go wrong (famous last words).  Obviously I did test in a simple project though.

Edit: Of course, this doesn't address local plugins with more than 1 file.  Not immediately sure on the best way to handle that case.  Since it's less common, maybe easiest to just not cache for local plugins.  Scanning imports seems excessive.

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -303,7 +303,7 @@ BCp._inferHelper = function (inputFile, babelOptions, cacheDeps, babelrc) {
 
 // 'react-hot-loader/babel' => 'react-hot-loader'
 function packageNameFromTopLevelModuleId(id) {
-  return id.replace(/\/.*$/, '');
+  return id.split("/", 1)[0];
 }
 
 var crypto = Npm.require('crypto');
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -67,6 +67,9 @@ BCp.processOneFileForTarget = function (inputFile, source) {
     sourceMap: null,
     bare: !! fileOptions.bare
   };
+  var cacheDeps = {
+    sourceHash: toBeAdded.hash
+  };
 
   // If you need to exclude a specific file within a package from Babel
   // compilation, pass the { transpile: false } options to api.addFiles
@@ -105,7 +108,7 @@ BCp.processOneFileForTarget = function (inputFile, source) {
       babelOptions.plugins.push(babelModulesPlugin);
     }
 
-    this.inferExtraBabelOptions(inputFile, babelOptions);
+    this.inferExtraBabelOptions(inputFile, babelOptions, cacheDeps);
 
     babelOptions.sourceMap = true;
     babelOptions.filename =
@@ -117,7 +120,7 @@ BCp.processOneFileForTarget = function (inputFile, source) {
 
     try {
       var result = profile('Babel.compile', function () {
-        return Babel.compile(source, babelOptions);
+        return Babel.compile(source, babelOptions, cacheDeps);
       });
     } catch (e) {
       if (e.loc) {
@@ -153,7 +156,7 @@ function profile(name, func) {
   }
 };
 
-BCp.inferExtraBabelOptions = function (inputFile, babelOptions) {
+BCp.inferExtraBabelOptions = function (inputFile, babelOptions, cacheDeps) {
   if (! inputFile.require ||
       ! inputFile.findControlFile ||
       ! inputFile.readAndWatchFile) {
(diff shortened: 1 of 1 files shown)


packages/babel-compiler/babel-compiler.js
```diff
@@ -276,3 +300,25 @@ BCp._inferHelper = function (inputFile, babelOptions, babelrc) {
 
   return false;
 };
+
+// 'react-hot-loader/babel' => 'react-hot-loader'
+function packageNameFromTopLevelModuleId(id) {
+  return id.replace(/\/.*$/, '');
```
benjamn: What about just `id.split("/", 1)[0]`?

packages/babel-compiler/babel-compiler.js
```diff
@@ -216,40 +222,58 @@ BCp._inferHelper = function (inputFile, babelOptions, babelrc) {
...
           presetOrPlugin = inputFile.require(id);
+          presetOrPluginMeta = inputFile.require(
+            packageNameFromTopLevelModuleId(id) + '/package.json');
         }
       } else {
         // If the identifier is not top-level, but relative or absolute,
         // then it will be required as-is, so that you can implement your
         // own Babel plugins locally, rather than always using plugins
         // installed from npm.
         presetOrPlugin = inputFile.require(id);
```
benjamn: There's also `inputFile.resolve(id)`, which could be used to determine the full absolute path of the module (with suitable conversion for Windows), and then we could hash the file contents instead of hashing the stringified function.

benjamn: Of course you'd want to use `inputFile.readAndWatchFile(absPath)` to read the file, so later changes will trigger a rebuild.

benjamn:
> Since it's less common, maybe easiest to just not cache for local plugins. 

This is definitely tempting…

packages/babel-compiler/babel-compiler.js
gadicc: Yes, that would be better :)

packages/babel-compiler/babel-compiler.js
gadicc: Yeah `func.toString()` wouldn't have been my first choice but I wanted to avoid reading the file a 2nd time, maybe I'm a little conservative like that.  Without using `Module._{cache,compile}` directly.  But agree reading the file directly is cleaner and hopefully the OS will have it cached too.  Ultimately though this will depend on our discussion about how to best handle caching for this case.

gadicc:
I think there are only two real options for this case; 1) not cache, 2) scan deps.  All the other "compromises" I thought of would lead to an confusing/frustrating dev experience.  Hashing against all the deps is definitely the right way to go, but maybe it's something to save for another day (I might still have a stab though if I get a chance).

For now we could just explain this clearly in the docs (and mention that publishing to npm is an option to improve performance), and I think also a warning on the console with a link to the docs?  With a local plugin for the *app*, and a big app, this could become detrimental.  This issue is probably also relevant for symlinks in node_modules (e.g. from `npm link` for local dev).

gadicc:
Oh, one compromise was to check if the single file actually even imports/requires any relative modules, and include the hash of the package.json for top level imports.  That's not too bad.  Wouldn't cache dynamic requires, but neither would an import scanner.  My reluctance is just to introduce more cases and tell the dev "ok, if (1) then... if (2) and (a) but not (b) then..." which I think is probably (?) worse than longer load times.

benjamn:
I'm going to go ahead and merge this as-is, and do some follow-up experimentation with the relative module caching. I'll @mention you in a comment when I figure something out (or get stuck again).

gadicc:
Ok, great!  Will still do the BABEL_ENV stuff as soon as I have a chance.

Issue #7090, opened 2016-05-20
meteor list is taking forever


mitar:
Sometimes, `meteor list` inside my [Dockerfile takes forever](https://github.com/tozd/docker-meteor/blob/master/Dockerfile#L18). It seems like it is in a loop.

See attached log for an example. This has been going for more than 12 hours now (for this particular build).

[Terminal Saved Output.txt.gz](https://github.com/meteor/meteor/files/275502/Terminal.Saved.Output.txt.gz)

abernix:
Can you attach `gdb` to it and see where it's stuck?  If it's only in a `Dockerfile` maybe you reproduce in some other TTY-less situation?  screen/ssh or something? 💭 

My computer had a lot of fun with that 1.2Gb text file.

mitar:
It is not stuck. It is working all the time, but it is not progressing.

I was also thinking that it might be TTY thingy. Maybe `meteor list | cat` should be a good way to test it.

abernix:
My only other thought would be that it's stuck on the sqlitedb (locked?) for some reason.

mitar:
OK, if it was not clear. The output is going and going. The log I send is not the end of the output. But it continue for hours in the same way. Working, 100% CPU, but not really progressing.

abernix:
Yeah I got that from the log. My computer was at 200% cpu just opening it.  Haha.

mitar:
Crazy thing is that on Docker Hub it works and image with the same Dockerfile compiles well. Only when trying to compile the image locally.

And this is something new. It has not been a problem in the past.

mitar:
Interesting. It seems to get stuck trying to download core packages. From Docker Hub where it works, I have for example such lines in my log (indentation not preserved):

```
 Task [Downloading 120 packages child] 3719829/?
   Task [Downloading babel-compiler@5.8.24_1...] 3719829/?
 Task [Downloading 120 packages child] 223294/?
   Task [Downloading blaze@2.1.3...] 223294/?
 Task [Downloading 120 packages child] 4483553/?
   Task [Downloading coffeescript@1.0.11...] 4483553/?
 Task [Downloading 120 packages child] 9627494/?
   Task [Downloading cosmos:browserify@0.9.3...] 9627494/?
 Task [Downloading 120 packages child] 333998/?
   Task [Downloading ddp-client@1.2.1...] 333998/?
 Task [Downloading 120 packages child] 111257/?
[...]

PR #7143, opened 2016-05-31
Set timeout for HTTP requests

mitar:
While debugging #7090 I discovered that there is no timeout specified for many HTTP requests Meteor is making. This means that when packaging server (or CDN) is badly behaving, client is waiting indefinitely instead of throwing an error.

### DIFF
--- a/tools/utils/http-helpers.js
+++ b/tools/utils/http-helpers.js
@@ -12,6 +12,7 @@ var auth = require('../meteor-services/auth.js');
 var config = require('../meteor-services/config.js');
 var release = require('../packaging/release.js');
 var Console = require('../console/console.js').Console;
+var timeoutScaleFactor = require('./utils.js').timeoutScaleFactor;
 
 
 // Helper that tracks bytes written to a writable
@@ -242,6 +243,12 @@ _.extend(exports, {
     }
     if (proxy && !options.proxy) {
       options.proxy = proxy;
+    }
+
+    if (! options.timeout) {
+      // 15 seconds for timeout between initial response headers and data,
+      // and between chunks of data while reading the rest of the response.
+      options.timeout = 15000 * timeoutScaleFactor;
     }
 
     // request is the most heavy-weight of the tool's npm dependencies; don't
--- a/tools/utils/http-helpers.js
+++ b/tools/utils/http-helpers.js
@@ -246,9 +246,9 @@ _.extend(exports, {
     }
 
     if (! options.timeout) {
-      // 15 seconds for timeout between initial response headers and data,
+      // 30 seconds for timeout between initial response headers and data,
       // and between chunks of data while reading the rest of the response.
-      options.timeout = 15000 * timeoutScaleFactor;
+      options.timeout = 30000 * timeoutScaleFactor;
     }
 
     // request is the most heavy-weight of the tool's npm dependencies; don't


laosb:
This might sounds too short for people in China, as in most cases it spends more than 30s. We should have a ENV to disable timeout.

abernix:
@laosb The `TIMEOUT_SCALE_FACTOR` env variable exists already and you can set it to a higher multiplier.

laosb:
Didn't realize it. Was that env used in other places?

mitar:
Also, timeout is not to download a file, but more or less timeout between TCP packets. I do not think that you have 15 seconds between packets even in China. :-)

laosb:
Ah, my mistake, sorry!

mitar:
I hoped that my comment in the code is conveying this. :-( Should I improve it?

abernix:
I thought it was fine – in general when talking about HTTP "timeout" this is what is implied – it's never the overall duration of the download, but instead the max time between blocks of data.

My only concern is if this is an _appropriate_ use of `timeoutScaleFactor` – as that's only really used in selftest (usually in `test.waitSecs`) for increasing the delay when awaiting the next command to output to the log.  I don't think it's currently purposed for networking stuff.

abernix:
I also think all future ENV variables should be prefixed with `METEOR_`, but that's a separate issue and not something you've caused with this PR. :)

laosb:
And we should have a cheatsheet for ENVs.

mitar:
I do not know, I can remove it. I do not see really a reason for anyone to increase it to more than 15 seconds anyway.

laosb:
We should not remove it, as nobody knows who will need it.

abernix:
I think this `timeout` is important.  I might actually make it 60 seconds just to cover the worst cases. A 15 second transmission timeout can absolutely happen – especially on a wifi or mobile network when switching modes – but is still recoverable in most all cases.  After a minute, you're very likely down though.  I guess 60 seconds of waiting might be obnoxious in some environments though – maybe 30 is an okay compromise though?

I'd probably wait to see about using `timeoutScaleFactor` based on what core says.

Based on the networking problems I've seen so far (including CDN problems, etc.) I think a procedure for HTTP request retry strategies is the next thing that might need some attention.

mitar:
I think we should merge it with 15 seconds, and provide it with a release candidate. And then see what people give feedback on.

mitar:
But yes, it could also be longer. I just do not want cases like now where it just hangs indefinitely.

abernix:
Definitely don't want infinite hanging.  My vote would be 60 – too short of a timeout and this could actually _increase_ the number of package install problems by terminating them when they still might succeed if left to wait.

benjamn:
Shall we ask our friend `Math.random`?
```js
> Math.round(Math.random() * 45 + 15)
35
```
In all seriousness, I'm subjectively comfortable with 30 seconds.

mitar:
And should I just remove the scaling and make it 30s and this is it?

benjamn:
Yeah let's make it 30 seconds, and keep the scaling factor unless there's a good reason to remove it.

mitar:
Others said that until now scaling factor is used only for testing.

zol:
Lgtm, thanks @mitar

jamiter:
I'm getting the following error when I try to deploy to Galaxy with 1.3.4
```
Error deploying application: Connection error (ETIMEDOUT)
```
Could this be related or is Galaxy simply having issues?

jamiter:
For who is reading the above comment; It has been comfirmed that this caused the deploy issue. A fix will be released soon. See this thread for details:

https://forums.meteor.com/t/1-3-4-breaks-galaxy-deployment-etimedout/25383/7

fdiazsmith:
Can some explain how to change the TIMEOUT_SCALE_FACTOR. Is this a variable stored somewhere, can it be done from the terminal?

benjamn:
On Mac/Linux:
```sh
TIMEOUT_SCALE_FACTOR=10 meteor deploy ...
```

On Windows (PowerShell):
```sh
$env:TIMEOUT_SCALE_FACTOR = "10"
meteor deploy ...
```

laosb:
Unfortunately, even 60s got an exceed in China. Changing TIMEOUT_SCALE_FACTOR...

mitar:
For downloading or uploading?

I think the solution is to restart. Not to extend timeout. With such long delays between TCP packets, TCP connection might not recover anyway.

laosb:
Downloading. Retried several times.

laosb:
An auto-retry-and-resume strategy would make Meteor usable in China, I think.

mitar:
Care of trying to add this feature to [request](https://github.com/request/request) library?

adamgins:
I seems to be getting a similar issue with Meteor 1.3.5.1 I have tried TIMEOUT_SCALE_FACTOR = "10" but I cannot consistently deploy to Galaxy

shadoath:
What is the `TIMEOUT_SCALE_FACTOR` range and what does each value mean?

PR #7147, opened 2016-06-01
Change package creation skeleton to import properly

### LINKED ISSUE #6653: unable to unit test newly created package for 1.3
in Meteor 1.3, the dummy default package code itself does not compile when running test-packages command

running
```
meteor create packbug
cd packbug
meteor create --package test:pack-c
meteor add test:pack-c
meteor test-packages test:pack-c
```

it does not compile and give this
```
Error: Can't find npm module 'meteor/pack-c'. Did you forget to call 'Npm.depends' in package.js within the 'modules-runtime' package?
```

i am trying to understand how the new package modules worked in the new Meteor 1.3, turns out the problem i had lies with the default generated code itself, it is confusing

abernix:
The package skeleton test example was importing the module as if it was outside the package.  This changes it to reference the local package source, relative to itself, so that this sequence of commands will work properly:

```
meteor create my-app1
cd my-app1
meteor create --package my-app1:my-package123
meteor add my-app1:my-package123
meteor test-packages my-app1:my-package123
```

Fixes #6653

### DIFF
--- a/tools/static-assets/skel-pack/~fs-name~-tests.js
+++ b/tools/static-assets/skel-pack/~fs-name~-tests.js
@@ -2,7 +2,7 @@
 import { Tinytest } from "meteor/tinytest";
 
 // Import and rename a variable exported by ~fs-name~.js.
-import { name as packageName } from "meteor/~fs-name~";
+import { name as packageName } from "meteor/~name~";
 
 // Write your tests here!
 // Here is an example.


benjamn:
The test package is different from the non-test package, so I'm worried this will cause a different copy of the module to be imported, rather than importing it from the non-test package.

I suspect the correct fix here is to make sure the `my-app1:` name prefix is included in the `import ... from "meteor/my-app:my-package123"` source identifier.

abernix:
I didn't really think about them being packaged separately and treated it like a main app (also not noticing the lopped-off package prefix led to trying the alternative) – but that makes perfect since since one half is just using the other.  

Fixed and tests already done.

tools/static-assets/skel-pack/~fs-name~-tests.js
```diff
@@ -2,7 +2,7 @@
 import { Tinytest } from "meteor/tinytest";
 
 // Import and rename a variable exported by ~fs-name~.js.
-import { name as packageName } from "meteor/~fs-name~";
+import { name as packageName } from "meteor/~name~";
```
benjamn: Oh wow, glad we support `~name~`!

abernix: Yup, turns out `~fs-name~` was actually derived from it.

Too easy.

PR #7164, opened 2016-06-03
Allow a function to be passed in the 'sort' clause

### LINKED ISSUE #2730: Allow specifying a comparator function for sorting a cursor

MasterAM:
Fixes #2730.

Do you think that it is sufficient or should I create separate checks for `MongoInternals` and `LocalCollection`?

### DIFF
--- a/packages/mongo/collection_tests.js
+++ b/packages/mongo/collection_tests.js
@@ -52,3 +52,48 @@ Tinytest.add('collection - call new Mongo.Collection with defineMutationMethods=
     test.equal(nomethods._connection[handlerPropName]['/' + noMethodCollectionName + '/insert'], undefined);
   }
 );
+
+Tinytest.add('collection - call find with sort function',
+  function (test) {
+    var handlerPropName = Meteor.isClient ? '_methodHandlers' : 'method_handlers';
+    var initialize = function (collection) {
+      collection.insert({a: 2});
+      collection.insert({a: 3});
+      collection.insert({a: 1});
+    };
+
+    var sorter = function (a, b) {
+      return a.a - b.a;
+    }
+
+    var getSorted = function (collection) {
+      return collection.find({}, {sort: sorter}).map(function (doc) { return doc.a; });
+    };
+
+    var collectionName = 'sort' + test.id;
+    var localCollection = new Mongo.Collection(null);
+    var namedCollection = new Mongo.Collection(collectionName, {connection: null});
+
+    initialize(localCollection);
+    test.equal(getSorted(localCollection), [1, 2, 3]);
+
+    initialize(namedCollection);
+    test.equal(getSorted(namedCollection), [1, 2, 3]);
+  }
+);
+
+Tinytest.add('collection - call native find with sort function',
+  function (test) {
+    var collectionName = 'sortNative' + test.id;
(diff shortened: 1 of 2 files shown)


mitar:
I would make two checks, one if `self.name` is set (which should be the same as now), and another if `self.name` is not set (which means it is local collection) allow function.

Maybe there is some better way to know if it is local collection or mongo internals.

MasterAM:
Probably `self._collection instanceof LocalCollection` or something, as named collections without a connection on the server also use `LocalCollection`.

The way it is at the moment, the check is not very rigorous anyhow, (it still allows plenty of invalid selectors), so I am wondering if it is worth the effort and the extra complexity.

mitar:
@stubailo: Can you check and merge this pull request?

packages/mongo/collection_tests.js
```diff
@@ -52,3 +52,48 @@ Tinytest.add('collection - call new Mongo.Collection with defineMutationMethods=
     test.equal(nomethods._connection[handlerPropName]['/' + noMethodCollectionName + '/insert'], undefined);
   }
 );
+
+Tinytest.add('collection - call find with sort function',
+  function (test) {
+    var handlerPropName = Meteor.isClient ? '_methodHandlers' : 'method_handlers';
```
tmeasday: This seems like a spurious line.

tmeasday:
Would it be a problem to just allow the `Function` type through in all cases and let mongo throw an exception? What does the error from Mongo look like? 

(I suspect something like `[Error: Illegal sort clause, must be of the form [['field1', '(ascending|descending)'], ['field2', '(ascending|descending)']]]`)

mitar:
Yes, let's just allow function and this is it.

packages/mongo/collection_tests.js
MasterAM: Indeed, it is. Copy-paste leftover. Sorry for that!

MasterAM:
You are correct about the error. I think that it is more descriptive than a `Match` error anyway.

The first commit simply adds the function type to the spec, so you can merge that one.
The tests were written for the other (separate native vs minimongo spec), but I can adapt them to handle the `Mongo` exception.

tmeasday:
Merged!

PR #7176, opened 2016-06-05
IPv6 Support. closes #7155 

### LINKED ISSUE #7155: IPv6 addresses weren't properly parsed
I'm trying to run meteor on IPv6, but opening app in browser gives 404 "Unknown path".

Using default app that is created with "meteor create xx". It works with IPv4, but it doesn't with IPv6.

Meteor version: 1.3.2.4
Platform: Debian GNU/Linux 8 amd64

By default meteor does not bind to IPv6 address so I tried starting with:
meteor run --port [::]:3000
meteor run --port [xxxx:........:xxxx]:3000

And then opening http://[xxxx:........:xxxx]:3000/ gives 404 "Unknown path"

By searching through source code I suspect that webapp package gives this error but I was unable to dig any further.

chemicstry:
This is my take on fixing IPv6 handling in meteor.

What's changed:
- Format URLs using url.format instead of string concatenation
- Renamed host to hostname. (host is ip+port and hostname is just ip, seems to be mixed up here)

IPv6 is working perfectly with these changes. However, I still may have missed other parts of the code where URLs are constructed using string concatenation.

Please review this PR and I will fix what is needed.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -108,9 +108,9 @@ function parseMobileServerOption(mobileServerOption,
   optionName = 'mobile-server') {
   let parsedMobileServerUrl = utils.parseUrl(
     mobileServerOption,
-    { protocol: 'http://' });
+    { protocol: 'http' });
 
-  if (!parsedMobileServerUrl.host) {
+  if (!parsedMobileServerUrl.hostname) {
     Console.error(`--${optionName} must include a hostname.`);
     throw new main.ExitWithCode(1);
   }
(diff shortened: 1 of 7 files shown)


chemicstry:
I am not sure why coffee script caching test failed though...

tmeasday:
I think the it's a flaky test due to log ordering. Will re-run it now.

tmeasday:
Ok, test passed. Thanks very much @chemicstry, this looks good. Merging.

PR #6889, opened 2016-04-24
Add Accounts.onLogout(func) Hook

pmoons:
`Accounts.onLogout(func)` is practically identical to the implementation of `Accounts.onLogin(func)`.  This PR is an effort to maintain symmetry between the available Login / Logout hooks.

Use cases include destroying active session variables, reverting UI elements, and any other cleanup that needs to occur immediately after the user is logged out.  

This PR is related to the following discussions:
[#5984](https://github.com/meteor/meteor/pull/5894)
#3867 
[#1074](https://github.com/meteor/meteor/issues/1074) 
[#426](https://github.com/meteor/meteor/issues/426)

Corresponding user-facing documentation has also been provided in this PR, as well as appropriate tests.

PR #7187, opened 2016-06-07
Fix #6889

tmeasday:
See https://github.com/meteor/meteor/pull/6889

Todo: 

 - [x] Ensure tests pass
 - [ ] Add history entry
 - [ ] Make changes to `vNEXT` of docs
 - [ ] Follow up on issues linked to the #6889


Ben305:
@tmeasday Thanks a lot for getting this done! Any change to see this in `1.3.3`?

tmeasday:
Nah, we are in rc for 1.3.3 but you can expect to see a beta for 1.4 (with this in it) very soon after the 1.3.3 release

gsonnenf:
Hi, I does the onLogout hook have an argument that returns the user who logged out or his connection? I was playing with a onLogout hook in 1.3.4 and it doesn't seem to have any of that available (perhaps im looking in the wrong place or the real update hasn't been applied yet?)

tmeasday:
Yes, it made it into 1.3.3.1 https://github.com/meteor/meteor/blob/devel/History.md#v1331

It could make sense to add it in another PR if it's not already available.

PR #7190, opened 2016-06-08
Mongo 3.2 update/upsert returns nMatched as numberAffected

Fabs:
- Add tests for the previous behaviour (returning nMatched)
   - I verified they pass on devel too before adding to this branch.
- Fix a bug with upserting when it did not cause any change.


tmeasday:
Looks great, thank you again @Fabs

PR #7203, opened 2016-06-11
Update Google client ID instructions again

ocdtrekkie:
Google changed them. Again. I also decided "Click" is a better verb than
"Press", so I changed that as well.


benjamn:
lgtm 👌

Issue #6883, opened 2016-04-23
Have `meteor update` explain how to update cordova packages


lorensr:
```
> meteor update cordova:cordova-plugin-inappbrowser
=> Errors while upgrading packages:           
                                              
While upgrading packages:
error: cordova:cordova-plugin-inappbrowser: package is not in the project

> meteor list | grep cordova-plugin-inappbrowser
cordova:cordova-plugin-inappbrowser           0.6.0
```

It would be better to say:

```
To update cordova-plugin-inappbrowser, run:

`meteor add cordova-plugin-inappbrowser@VERSION`

where VERSION is, for example: 1.0.0

To look up the latest version of cordova-plugin-inappbrowser, run:

`npm view cordova-plugin-inappbrowser version`
```

abernix:
Fixed via #7233!

PR #7233, opened 2016-06-16
Have `meteor update` explain how to update cordova packages

laosb:
#6883


tools/cli/commands-packages.js
```diff
@@ -1649,6 +1649,22 @@ main.registerCommand({
   } else {
     upgradePackageNames = options.args;
   }
+  var upgradePackagesWithoutCordova = _.reject(upgradePackageNames, function (name) {
+    return name.split(':')[0] === 'cordova'; 
+  });
+  if (upgradePackagesWithoutCordova !== upgradePackageNames) {
```
zol: This code won't work. You can't compare arrays using an equality operator.

nbrady-techempower: @zol isn't he comparing an element in the array?

zol: No, the arguments to `if (upgradePackagesWithoutCordova !== upgradePackageNames) {` are arrays.

nbrady-techempower: Indeed. My eyes went to the first equality operator! :shame:

laosb: Sorry! Too hurry to PR!

laosb:
@zol Edited.

Not sure if it's the best way to do it, but it should works.

tools/cli/commands-packages.js
```diff
@@ -1652,7 +1652,7 @@ main.registerCommand({
   var upgradePackagesWithoutCordova = _.reject(upgradePackageNames, function (name) {
     return name.split(':')[0] === 'cordova'; 
   });
-  if (upgradePackagesWithoutCordova !== upgradePackageNames) {
+  if (upgradePackagesWithoutCordova.sort().toString() !== upgradePackageNames.sort().toString()) {
```
zol: This is hacky. You should use an underscore method, like `_.difference`

laosb:
@zol Found `_.isEqual` and updated the code.

laosb:
@zol: Any updates on dealing with this PR?

tools/cli/commands-packages.js
```diff
@@ -1649,6 +1649,22 @@ main.registerCommand({
   } else {
     upgradePackageNames = options.args;
   }
+  var upgradePackagesWithoutCordova = _.reject(upgradePackageNames, function (name) {
+    return name.split(':')[0] === 'cordova'; 
+  });
+  if (upgradePackagesWithoutCordova !== upgradePackageNames) {
+    // There are some cordova packages in the list to update.
+    // We should tell users how to update cordova packages.
+    Console.warn('Looks like you\'re trying to update cordova packages.');
```
abernix: Really should use double quotes to contain this string.  Otherwise, if someone goes to search for this string in the source code, they might not find it unless they search for it with the extra backslash.

abernix: Additionally, maybe this should be rephrased to:

    It appears that you're trying to upgrade a Cordova package.  To perform this action, you should run:

tools/cli/commands-packages.js
```diff
@@ -1649,6 +1649,22 @@ main.registerCommand({
...
+    Console.warn('To update cordova packages, run:');
+    Console.warn('  meteor add PLUGIN@VERSION');
```
abernix: It still doesn't make it very clear that you should use the `cordova:` prefix.  For example, with the code below, I might assume that I should run:

    meteor add cordova-plugin-inappbrowser@1.0.0

Which would not work.

Perhaps this should say:

    Console.warn('  meteor add cordova:PLUGIN-NAME@VERSION');

laosb:
@abernix As it's a notification when user try to update their cordova packages, I assume that they should have known it. Though adding few bytes cost nothing and may help beginners!

laosb:
@abernix missed your two comments, now edited!

laosb:
@meteor any updates or I should close it?

zol:
I'll take a closer look at it and merge it the next chance I get.

tools/cli/commands-packages.js
```diff
@@ -1649,6 +1649,22 @@ main.registerCommand({
   } else {
     upgradePackageNames = options.args;
   }
+  var upgradePackagesWithoutCordova = _.reject(upgradePackageNames, function (name) {
+    return name.split(':')[0] === 'cordova'; 
+  });
```
abernix: ES6 instead of Underscore?:
```js
const upgradePackagesWithoutCordova = upgradePackageNames.filter(name => name.split(':')[0] !== 'cordova');
```

tools/cli/commands-packages.js
```diff
@@ -1649,6 +1649,22 @@ main.registerCommand({
...
+  if (_.isEqual(upgradePackagesWithoutCordova, upgradePackageNames)) {
```
abernix: If these two variables are equal...

tools/cli/commands-packages.js
```diff
@@ -1649,6 +1649,22 @@ main.registerCommand({
...
+    // We should tell users how to update cordova packages.
+    Console.warn("It appears that you're trying to upgrade a Cordova package.");
+    Console.warn('To perform this action, you should run:');
+    Console.warn('  meteor add cordova:PLUGIN@VERSION');
+    Console.warn('where PLUGIN is your cordova package name and VERSION is, ' +
+                'for example: 1.0.0');
+    Console.warn('To look up the latest version of cordova-plugin-inappbrowser, run:');
+    Console.warn('  npm view PLUGIN version');
+    // Exclude cordova packages
+    upgradePackageNames = upgradePackagesWithoutCordova;
```
abernix: ... then why set them to be the same?

zol: This is a good point, please remove the redundant line.

tools/cli/commands-packages.js
zol: This would be preferable but not mandatory.

tools/cli/commands-packages.js
laosb: @abernix @zol Actually I missed a "'" before _.isEqual. Will update tonight.

laosb: "!" not "'", sorry! (On my phone)

tools/cli/commands-packages.js
laosb: Is es6 available in this file?

tools/cli/commands-packages.js
laosb: Updated. @zol @abernix

tools/cli/commands-packages.js
abernix: Yes, ES6 is available in this file.

tools/cli/commands-packages.js
abernix: That would also make more sense.

@laosb I think changes like this are important and helpful, especially from a tech support standpoint. I understand you've had problems running from checkout in the past due to the firewall, and I think you've got some of those checkout problems fixed now, but in the future if you need help confirming simple changes like this and are unable to test them yourself, please make it very clear in the PR that it's not tested and provide instructions.  You know where to find me on chat too.   I'm happy to help!  :wink:

tools/cli/commands-packages.js
laosb: Good, looks like much code of this file can be updated to use ES6.

abernix: Yes, but I believe only existing code that is _actually being changed_ and new additions are being updated to ES6 at this time.

laosb:
@abernix: I've tested and did some enhancements on this PR.
image

laosb:
@abernix Sounds like you're the assignee now. How do you think about this? I would update this now.

abernix:
Meant to take care of this earlier this week but... **1.4**!  Let's get this closed.  :smile:

(Btw, sorry that going back and forth on these text edits has been a bit painful, but I think the changes and clear messaging/formatting are super important.  In the future, maybe a PR is fine for single-line changes but I think an issue _first_ would be best for larger blocks of text (like this PR).  Easier than doing line-by-line-comments!  What do you think?)

For what I hope is the final update!  How about this for the text? (Let me know if it makes sense or it's not clear before you commit/push):

```
To add/upgrade a Cordova plugin in your Meteor project, run:

    meteor add cordova:PLUGIN-NAME@x.y.z

The 'PLUGIN-NAME' should be an official plugin name (e.g. cordova-plugin-media)
and the 'x.y.z' should be an available version of the plugin.  The latest
version can be found by with the following command:

    meteor npm view PLUGIN-NAME version

The non-Cordova packages will now be updated...
```

I used a shorter plugin name to keep the lines short.

For the actual command part of the text, can you use the `Console.command` function as shown in [other command examples](https://github.com/meteor/meteor/blob/devel/tools/cli/commands.js#L1189-L1199).

Finally, please rebase and if you can provide a screenshot of what it looks like when you're done and it looks good I'll merge it.

tools/cli/commands-packages.js
```diff
@@ -1686,11 +1705,13 @@ main.registerCommand({
 
       // If the user explicitly mentioned some packages to upgrade, they must
       // actually end up in our solution!
-      _.each(options.args, function (packageName) {
-        if (! projectContext.packageMap.getInfo(packageName)) {
-          buildmessage.error(packageName + ': package is not in the project');
-        }
-      });
+      if (options.args.length !== 0) {
```
abernix: Is this `if` statement even necessary?  Doesn't `upgradePackageNames` _have_ to be an array at this point?  If so, `each` will just do nothing anyway.  It just seems redundant.

laosb: I just added this `if` statement according to the comment:

```
// If the user explicitly mentioned some packages to upgrade, they must
// actually end up in our solution!
```

laosb:
@abernix I've merged `devel` into this branch but it sounds like still having conflicts. Not sure how to do it.

I will go with your version of text, since I'm not a English native speaker. Thank you for your turning in words!

laosb:
@abernix Should we do a hard line wrap in paragraph using `Console.warn` for every line, or just wrap in code by `"xxx" + "xxx"` which results in no wrap(or soft wrap by terminal app) in the output?

laosb:
Updated text:

image

abernix:
> Should we do a hard line wrap in paragraph

Yes, hard wrap and newlines before and after the command.

> still having conflicts. Not sure how to do it.

No worry.  I'll take care of it.

laosb:
@abernix Ok, updating to hard wraps.

tools/cli/commands-packages.js
abernix: Ok.  Seems okay since with your conditional since `upgradePackageNames` should only be non-Cordova packages that they specified on the command line.

Also, I would probably just write this as:

    if (options.args.length) {

...since it's an array and the only truth-y value possible is non-zero.

abernix:
We can do without the hard-wraps, but:

* No double spaces between `of  the` or before `(e.g.`
* Newlines before and after commands.  (Easier to find them)

laosb:
image

abernix:
LGTM.  Will merge after CircleCI finishes proof-reading it for us! ;)

Thanks for working through this!

PR #7235, opened 2016-06-16
Removes Velocity testing code

Sanjo:
The support of the Velocity testing functionality has been dropped with Meteor 1.3 by me, who maintained the Velocity Meteor fork that had some additional functionality in it that was required to run tests with Velocity. Therefore the code that is removed in this PR no longer serves any purpose.


benjamn:
Thanks for cleaning up this code, and for working on it in the first place. I applaud you for making significant contributions to the Meteor platform when it wasn't always easy or rewarding. It's my full-time job, so I know how much of an uphill battle it can be. I like to think contributing is getting easier, but we've definitely still got a long road ahead.

Cheers ✊

PR #7245, opened 2016-06-17
Allow `--headless` option for `test` and `test-packages`

jdivy:
In a nutshell, I think the `--headless` option should be available to both the `test` and `test-packages` command.  I couldn't think of a good reason why it's not available, and it would be much nicer to see CI logs that don't have many lines of rotating progress bars.

screen shot 2016-06-17 at 1 11 21 pm

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1413,6 +1413,10 @@ testCommandOptions = {
     deploy: { type: String },
     production: { type: Boolean },
     settings: { type: String },
+    // Indicates whether these self-tests are running headless, e.g. in a
+    // continuous integration testing environment, where visual niceties
+    // like progress bars and spinners are unimportant.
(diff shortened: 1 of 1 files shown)


tools/cli/commands.js
```diff
@@ -1413,6 +1413,10 @@ testCommandOptions = {
     deploy: { type: String },
     production: { type: Boolean },
     settings: { type: String },
+    // Indicates whether these self-tests are running headless, e.g. in a
+    // continuous integration testing environment, where visual niceties
+    // like progress bars and spinners are unimportant.
```
jdivy: I debated on whether or not to even keep this documentation, and of course I left a minor copy/paste error.  Let me know if you would rather just strip this comment altogether.

benjamn:
Looks good to me!

PR #7251, opened 2016-06-18
Generate dev bundle script : fix missing directory

TPXP:
The current `generate_dev_bundle.sh` script forgets to create the `lib/node_modules` directory in the build dir, which ends up in an undefined target when it attempts to add node modules to it. This PR aims at fixing it, so that the script works properly.

Thank you for this great project, and keep up the good work ! :+1:


TPXP:
Oops! In fact, this directory is created when NodeJS is installed. As, I'm trying to use my system-wide binaries, I commented out this part, but in fact there is no problem. Sorry ! :confused:

PR #7261, opened 2016-06-20
Reload: detect href ending with `#`.

### LINKED ISSUE #7239: reload package won't reload app if URL ends with `/#`
1. Run meteor app on `http://localhost:3000/#`.
2. Make a change in client code
3. Wait for it

Keep waiting for it. It never reloads.

This code: 
```
if (window.location.hash) {
  window.location.reload();
} else {
  window.location.replace(window.location.href);
}
```
doesn't work when URL ends with the `#` char.

realyze:
Make sure URLs ending with `#` get properly picked up by `Reload` package and the app refreshes.

Fixes #7239 .

### DIFF
--- a/packages/reload/reload.js
+++ b/packages/reload/reload.js
@@ -223,7 +223,7 @@ Reload._reload = function (options) {
       // with the server if we still have a valid cached copy. This doesn't work
       // when the location contains a hash however, because that wouldn't reload
       // the page and just scroll to the hash location instead.
-      if (window.location.hash) {
+      if (window.location.hash || _.last(window.location.href) === '#') {
         window.location.reload();
       } else {
         window.location.replace(window.location.href);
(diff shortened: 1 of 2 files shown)


packages/reload/reload.js
```diff
@@ -223,7 +223,7 @@ Reload._reload = function (options) {
       // with the server if we still have a valid cached copy. This doesn't work
       // when the location contains a hash however, because that wouldn't reload
       // the page and just scroll to the hash location instead.
-      if (window.location.hash) {
+      if (window.location.hash || _.last(window.location.href) === '#') {
```
benjamn: Can you add `api.use('ecmascript-runtime')` to `reload/package.js` and use `window.location.href.endsWith("#")` here? That will trigger another test run, too, which I think should pass this time…

realyze: Sure thing. Just out of curiosity, what's wrong with using underscore? It's used in this module already. I didn't really understand CircleCI's output - couldn't grok how it's linked to underscore.

benjamn:
Last thing: wanna sign the contributor agreement? https://contribute.meteor.com/

realyze:
Signed.

laosb:
is @meteor-bot down?

Issue #6958, opened 2016-05-03
Allow passing options in the MongoDB driver


tmeasday:
For example, to enable SSL validation (after #6957) - https://github.com/meteor/meteor/issues/5809#issuecomment-190841203

This could look something like 

`Mongo.setConnectionOptions({...});` to be called by the user before they initialize any collections, setting some internal state that is read at the various points we call `new MongoConnection`.

dburles:
@tmeasday just opened a PR, maybe you can take a look

tmeasday:
https://github.com/meteor/meteor/pull/7277

tmeasday:
This issue is fixed but I'm not sure it works in practice if you are using accounts. Have you tested that @dburles?

dburles:
I haven't

abernix:
I'm fairly sure that this doesn't work unless implemented as a package and loaded before (i.e. in `.meteor/packages`) any other package which initiates a collection (such as `accounts-base`).  #7455 and the [related forum post](https://forums.meteor.com/t/meteor-with-the-new-mongodb-3-2-1-compose-io/17763/11) seem to hint at that.  It does seem to work though!

Is it possible that `Meteor.settings` would be a more appropriate way to implement this?  But then, as mentioned in aforementioned links, there's the SSL CA file needing to be loaded from a filesystem path.

tmeasday:
Possibly accounts could do its init in a startup block but that could cause flow on problems..

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

This feature request will be closed here, but anyone interested in migrating this feature request to the new repository (to make sure it stays active), can click [here](https://github.com/meteor/meteor-feature-requests/issues/new?title=Allow%20passing%20options%20in%20the%20MongoDB%20driver&body=Migrated%20from:%20meteor/meteor%236958) to start the feature request migration process. This manual migration process is intended to help identify which of the older feature requests are still considered to be of value to the community. Thanks!

PR #7277, opened 2016-06-23
User defined mongo connection options API for #6958

dburles:
For #6958

I noticed the API has changed for a bunch of the existing options used in `mongo_driver.js` such as `{ safe: true }` now replaced by: https://docs.mongodb.com/manual/core/replica-set-write-concern/ but that is a bit out of scope for this PR. Maybe something to be looked at though.

### DIFF
--- a/packages/mongo/connection_options.js
+++ b/packages/mongo/connection_options.js
@@ -1,0 +1,10 @@
+/**
+ * @summary Allows for user specified connection options
+ * @example http://mongodb.github.io/node-mongodb-native/2.1/reference/connecting/connection-settings/
+ * @locus Server
+ * @param {Object} options User specified Mongo connection options
+ */
+Mongo.setConnectionOptions = function setConnectionOptions (options) {
+  check(options, Object);
+  Mongo._connectionOptions = options;
+}
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -130,7 +130,8 @@ MongoConnection = function (url, options) {
   self._observeMultiplexers = {};
   self._onFailoverHook = new Hook;
 
-  var mongoOptions = {db: {safe: true}, server: {}, replSet: {}};
+  var mongoOptions = _.extend({db: {safe: true}, server: {}, replSet: {}},
+                               Mongo._connectionOptions);
 
   // Set autoReconnect to true, unless passed on the URL. Why someone
   // would want to set autoReconnect to false, I'm not really sure, but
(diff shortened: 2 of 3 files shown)


packages/mongo/mongo_driver.js
```diff
@@ -151,13 +151,10 @@ MongoConnection = function (url, options) {
...
-  // XXX maybe we should have a better way of allowing users to configure the
-  // underlying Mongo driver
+  // Internally the oplog connections specify their own poolSize
+  // which we don't want to overwrite with any user defined value
   if (_.has(options, 'poolSize')) {
-    // If we just set this for "server", replSet will override it. If we just
-    // set it for replSet, it will be ignored if we're not using a replSet.
-    mongoOptions.server.poolSize = options.poolSize;
-    mongoOptions.replSet.poolSize = options.poolSize;
+    mongoOptions.poolSize = options.poolSize;
```
tmeasday: Has this `poolSize` option changed? If so we should probably fix this

packages/mongo/collection.js
```diff
@@ -255,6 +255,12 @@ Mongo.Collection = function (name, options) {
   }
 };
 
+// Allow for user specified connection options
+Mongo.setConnectionOptions = function setConnectionOptions (options) {
+  check(options, Object);
+  Mongo._connectionOptions = options;
+}
+
```
tmeasday: Can we add some jsdoc for this also?

tmeasday:
Looking good @dburles

dburles:
For reference here's the relevant docs for the newer options http://mongodb.github.io/node-mongodb-native/2.1/reference/connecting/connection-settings/

packages/mongo/mongo_driver.js
dburles: Only in how you pass it in, the new mongo driver takes care of where it needs to go

tmeasday: Does the old code work though? It seems surprising that this passed our testing of the new driver..

dburles: Oh right yeah, well as far as I've looked into things it seems like legacy options will work just fine, though I'm not totally sure what happens with `safe`, I think it may just be ignored as it's the default on later Mongo, but I could be wrong.

tmeasday: Maybe remove this from the PR then, I think it muddies the waters a little

dburles: Sure thing

dburles:
Seems pretty good now https://github.com/meteor/meteor/pull/7277/files

packages/mongo/connection_options.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @summary Allows for user specified connection options
+ * @locus Server
+ * @param {Object} options User specified Mongo connection options
```
tmeasday: Can we add a link to the mongo docs so people know what these options are?

dburles: Done

dburles:
Just a heads up, this was definitely merged, but for some reason now it's not included in the `release-1.4` branch

tmeasday:
Any idea what happened here @benjamn?

benjamn:
Not sure why it disappeared, but I've cherry-picked it back onto release-1.4: https://github.com/meteor/meteor/commit/2654fe68be7d212c2ed2ec0c94333fde79648dd0

PR #7279, opened 2016-06-23
Finalize replacement NPM@3 at generate-dev-bundle.sh script

TomFreudenberg:
With release 1.3.4 there was a new variable `$NPM_VERSION` defined in `build-dev-bundle-common.sh`.

Instead checking the right version inside the `generate-dev-bundle.sh` this var is better to use.

In addition, the message is not correct anymore because no special node is delivered anymore. The npm is not build via PORTABLE=1 in the binaries of nodejs.

In case that the NPM is replaced during the bulid this content should be deleted.

Hope that fits to you
Tom


benjamn:
Good catch! Makes total sense to me.

TomFreudenberg:
Thanks Ben

PR #7256, opened 2016-06-19
Adds truthiness check b4 accessing property in facebook package. Fixes #7255

workflow:
Adds truthness check b4 accessing property in facebook package. Fixes #7255:

Trying to login using `facebook@1.2.7` (newest version) in Meteor 1.3.3.1 yields the following type check error:

```
facebook_client.js:40 Uncaught TypeError: Cannot read property 'authType' of null
```

Please do let me know if you need more info to reproduce.
I'm attaching a one-line PR that fixes this.

PR #7281, opened 2016-06-23
Check options exist in twitter package

pradel:
Trying to login with Twitter package give an error:
`Uncaught TypeError: Cannot read property 'force_login' of null`

Make the same thing than #7256 but for twitter package.
This issue was introduced by commit https://github.com/meteor/meteor/commit/001a3f3128d1eaebdea431bf35707748b5dc185e


benjamn:
Thanks!

PR #7285, opened 2016-06-23
Release 1.3.4.1

benjamn:
Adding a timeout to HTTP requests made by the `meteor` command-line tool (#7143) [caused some Galaxy deploys to fail on slower connections](https://forums.meteor.com/t/1-3-4-breaks-galaxy-deployment-etimedout/25383).

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,15 @@
 ## v.NEXT
 
+## v1.3.4.1
+
+* Increased the default HTTP timeout for requests made by the `meteor`
+  command-line tool to 60 seconds (previously 30), and [disabled the
+  timeout completely for Galaxy
+  deploys](https://forums.meteor.com/t/1-3-4-breaks-galaxy-deployment-etimedout/25383/).
+
+* Minor bug fixes: [#7281](https://github.com/meteor/meteor/pull/7281)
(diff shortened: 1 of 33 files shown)


mitar:
Hm, but this is strange. Because this means that somebody is deploying to Galaxy where they have not send any TCP packet for more than 30 seconds?

mitar:
Maybe the issue is that timeout works for downloading data, not for uploading? So while uploading, timeout is not cleared on uploaded chunks?

Maybe this means that there is a need for some improvements to requests package, or some wrapper around it? Which would take into the account both uploading and downloading, and retry each if it fails, instead of just aborting?

Is this an issue also then for package publishing?

benjamn:
@mitar yeah, I think maybe we should have restricted the default timeout to GET requests (the Galaxy publish request is a POST)

mitar:
Yes, but this is probably an upstream bug. That it does not clear timeout while sending.

mitar:
Because if upload really timeouts, we should be throwing an error. Here it seems that data is being send normally, only because nothing is downloaded while upload is happening, timeout triggers.

But we should have also timeout on uploading.

mitar:
I opened: https://github.com/request/request/issues/2259

Issue #7149, opened 2016-06-01
Autoupdate client does not honor WebAppInternals.setBundledJsCssPrefix


egoldblum:
Currently running 1.2.1 on Galaxy.

We set `WebAppInternals.setBundledJsCssPrefix` within a `Meteor.startup`.  In the general case this works fine.

Observed behavior:
After deploying new code, connected browser clients reload automatically, but requests for the new js/css land on our origin not the CDN.

Expected behavior:
Autoupdate'd css `link href` (and javscript `script src`) should continue to respect the set `WebAppInternals.setBundledJsCssPrefix`

https://github.com/meteor/meteor/blob/devel/packages/autoupdate/autoupdate_client.js#L131

laosb:
Yeah, should be and did reproduce on a self-hosted app.

tmeasday:
A simple way to reproduce this locally: 

```bash
# in one terminal
meteor create testapp
cd testapp
echo "WebAppInternals.setBundledJsCssPrefix('http://localhost:4000')" >> server/main.js
meteor

# in another
git clone https://github.com/tmeasday/simple-proxy
cd simple-proxy
npm install
node index.js
```

tmeasday:
@egoldblum I am only seeing the behavior you are referring to for CSS files, not JS files. Are you sure JS files don't hit the CDN?

egoldblum:
@tmeasday you are correct, only rewritten css links cause requests to our origin, not js.

abernix:
This is confirmed fixed in Meteor 1.3.4.2!  Thanks for reporting this problem, @egoldblum.

PR #7316, opened 2016-06-28
Publish CSS documents for autoupdate after rewriting for #7149

tmeasday:
A fix for #7149 

Previously we did not take the JsCssRewriteHook (i.e. CDN host)
into account when doing CSS-only "version-refreshable" HCP.

It is not possible on the client to rewrite CSS urls to take the
hook into account (as it is a function), so instead, we just
publish those CSS URLs post-transform.

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -36,7 +36,11 @@ WebApp.clientPrograms = {};
 // XXX maps archs to program path on filesystem
 var archPath = {};
 
-var bundledJsCssUrlRewriteHook;
+var bundledJsCssUrlRewriteHook = function (url) {
+  var bundledPrefix =
+     __meteor_runtime_config__.ROOT_URL_PATH_PREFIX || '';
+  return bundledPrefix + url;
+};
 
 var sha1 = function (contents) {
   var hash = crypto.createHash('sha1');
@@ -283,13 +287,6 @@ WebAppInternals.generateBoilerplateInstance = function (arch,
     _.clone(__meteor_runtime_config__),
     additionalOptions.runtimeConfigOverrides || {}
   );
-
-  var jsCssUrlRewriteHook = bundledJsCssUrlRewriteHook || function (url) {
-    var bundledPrefix =
-       __meteor_runtime_config__.ROOT_URL_PATH_PREFIX || '';
-    return bundledPrefix + url;
-  };
-
   return new Boilerplate(arch, manifest,
     _.extend({
       pathMapper: function (itemPath) {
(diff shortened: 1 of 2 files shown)


tmeasday:
@Slava if you have a second I suspect you may have some context on this. Hopefully the commit message and the original issue explains the problem I am trying to solve here.

benjamn:
I suspect the test failure was spurious. Rebuilding.

packages/webapp/webapp_server.js
```diff
@@ -574,9 +568,13 @@ var runWebAppServer = function () {
...
-      WebAppInternals.refreshableAssets = {
-        allCss: boilerplateByArch[WebApp.defaultArch].baseData.css
-      };
+      var cssFiles = boilerplateByArch[WebApp.defaultArch].baseData.css;
+      // Rewrite all CSS files (which are written directly to <style> tags)
+      // by autoupdate_client to use the CDN prefix/etc
+      var allCss = _.map(cssFiles, function(cssFile) {
+        return { url: self.bundledJsCssUrlRewriteHook(cssFile.url) };
+      });
+      WebAppInternals.refreshableAssets = { allCss: allCss };
```
benjamn: This can just be `WebAppInternals.refreshableAssets = { allCss };`.

packages/webapp/webapp_server.js
```diff
@@ -818,9 +816,16 @@ WebAppInternals.setInlineScriptsAllowed = function (value) {
   WebAppInternals.generateBoilerplate();
 };
 
+// this is the default implementation, can be overridden below
+WebAppInternals.bundledJsCssUrlRewriteHook = function (url) {
```
benjamn: Why expose this on the `WebAppInternals` namespace instead of keeping it private to the scope of this module?

packages/webapp/webapp_server.js
```diff
@@ -266,30 +264,24 @@ var getBoilerplate = function (request, arch) {
     }
     return memoizedBoilerplate[memHash];
   }
-  
-  var boilerplateOptions = _.extend({ 
-    htmlAttributes: htmlAttributes 
+
+  var boilerplateOptions = _.extend({
+    htmlAttributes: htmlAttributes
   }, _.pick(request, 'dynamicHead', 'dynamicBody'));
```
benjamn: As long as we're touching this code, I'll just note that this could be
```js
var boilerplateOptions = {
  htmlAttributes,
  ..._.pick(request, "dynamicHead", "dynamicBody")
}
```

packages/webapp/webapp_server.js
tmeasday: Yeah, I'll back out that change. I was originally thinking of calling it from autoupdate, but no longer.

packages/webapp/webapp_server.js
tmeasday: That was just my editor. Not sure I want to get into the selective ES2015-izing game.

tmeasday:
I think this diff should be improved

Issue #7326, opened 2016-06-29
Destroying Blaze templates with many sub-templates freezes browser


arggh:
Running newest Meteor 1.3.4.1 (or an older one if you wish), destroying a Blaze-template with many child/sub-templates causes a complete browser lock down for several seconds.

All that's needed is a basic table-element with ~few hundred rows and a couple of columns on each. Rendering the table happens in a user friendly manner, but once it's being teared down the user experience crumbles - the browser freezes completely, making users click repeatedly on the button/link/toggle thinking their clicks are not registered.

Reproduction can be found here: https://github.com/arggh/blaze-tmpl-destroy/tree/nosubs

The issue was already reported (and closed without resolution or fixes) at #4352 and then resurrected in the meteor/blaze -repository: https://github.com/meteor/blaze/issues/44.

I realize that Blaze has been split to it's own repository, but there's not much happening over there and the repository's maintainer stated that the transition is not yet complete, advising to post the issue in both repos.

I think this is such a drag and it's not really that uncommon to have a table with few hundred rows on a website, so maybe it **should be fixed?**

PR #3801, opened 2015-02-24
Tracker profiler package

Slava:
WIP.

- Expose references to computations in Tracker.
- Introduce a package called `tracker-profiler` that counts recomputations of every computation.
- Give Blaze autoruns descriptive names in the [`displayName`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/displayName) field.
- Spacebars generator should give more descriptive names to the views generated for helpers and other symbols (or identifier paths).

@avital

PR #7328, opened 2016-06-29
Set stopped computation to null instead of delete (#7326)

arggh:
Fix for issues https://github.com/meteor/meteor/issues/7326 and https://github.com/meteor/blaze/issues/44

This change in Tracker speeds up the removal of Blaze templates with many sub-templates by approximately 5-10x.

It replaces the `delete` of a stopped computation by setting it to `null`. Read more [here](http://bertanguven.com/preventing-memory-leaks-in-javascript-null-vs-delete)

Reproduction of the issue used as a benchmark (and a test to verify the effect): https://github.com/arggh/blaze-tmpl-destroy/tree/nosubs

Without the fix in Tracker, the first time hiding the table with 550 rows took about **1.5 seconds**. On the second go it took about **2.6 seconds**. Then **3.5 seconds**. Then **~5 seconds** and so on, which is really strange - as if things just got incremented instead of removed.

With the fix applied, hiding the same table takes a constant **300-450ms.**

### DIFF
--- a/packages/promise/.npm/package/npm-shrinkwrap.json
+++ b/packages/promise/.npm/package/npm-shrinkwrap.json
@@ -2,6 +2,7 @@
   "dependencies": {
     "meteor-promise": {
       "version": "0.7.2",
+      "resolved": "https://registry.npmjs.org/meteor-promise/-/meteor-promise-0.7.2.tgz",
       "from": "meteor-promise@0.7.2"
     },
     "promise": {
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -301,7 +301,7 @@ Tracker.Computation.prototype.stop = function () {
     self.stopped = true;
     self.invalidate();
     // Unregister from global Tracker.
-    delete Tracker._computations[self._id];
+    Tracker._computations[self._id] = null;
     for(var i = 0, f; f = self._onStopCallbacks[i]; i++) {
       Tracker.nonreactive(function () {
         withNoYieldsAllowed(f)(self);
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -26,13 +26,6 @@ Tracker.active = false;
  */
 Tracker.currentComputation = null;
 
-// References to all computations created within the Tracker by id.
-// Keeping these references on an underscore property gives more control to
-// tooling and packages extending Tracker without increasing the API surface.
-// These can used to monkey-patch computations, their functions, use
-// computation ids for tracking, etc.
-Tracker._computations = {};
-
 var setCurrentComputation = function (c) {
   Tracker.currentComputation = c;
   Tracker.active = !! c;
@@ -203,9 +196,6 @@ Tracker.Computation = function (f, parent, onError) {
   self._onError = onError;
   self._recomputing = false;
 
-  // Register the computation within the global Tracker.
-  Tracker._computations[self._id] = self;
-
   var errored = true;
   try {
     self._compute();
@@ -300,8 +290,6 @@ Tracker.Computation.prototype.stop = function () {
   if (! self.stopped) {
     self.stopped = true;
     self.invalidate();
-    // Unregister from global Tracker.
-    Tracker._computations[self._id] = null;
     for(var i = 0, f; f = self._onStopCallbacks[i]; i++) {
       Tracker.nonreactive(function () {
         withNoYieldsAllowed(f)(self);


arggh:
On Chrome Canary I observed the steady 300-450ms delay, on an older Chrome it seemed that the Tracker_.computations objects increasing size has a small incremental overhead, but it's barely noticeable compared to the increasing delay with using `delete`.

arggh:
As a quick test, running

```
Tracker._computations = _.pick(Tracker._computations, function(value, key, object) {
    return value !== null;
});
```

in the end of `Tracker._runFlush` seems to take care of the "issue" of ever-increasing amount of keys in the `Tracker._computations`-object without any noticeable performance penalty. It's another story whether it's necessary, since the object will never be enumerated.

As a sidenote, a quick test where I change the `Tracker._computations`-object into an array seems to shave some milliseconds as well, though I think I would need to create a more robust test environment to get solid results. The previous improvement was so huge it was easy to verify.

Any input would be greatly appreciated, since I'm treading on quite uncharted territory for myself (I took my first look at the internals of Meteor late last night), and I would really like to fix this problem (and not cause new ones).

abernix:
Initially I'd say that I think this is probably a great issue to focus on, but unfortunately the CI tests on this PR have not been very accommodating.

I thought it might have been a fluke failure, but at this point I've re-run them 8 times and each time I run it it seems that _completely_ different tests fail:

##### Failed on one pass or another...
```
  - cordova-plugins: remove cordova plugins
  - package-tests: show and search local package
  - package-tests: add package with no builds
  - package-tests: show and search local package
  - linter-plugins: linter plugins - lint package on `meteor publish`
  - static-html: static-html - add static content to head and body
  - colon-converter-tests: can't build local packages with colons
  - compiler-plugins: compiler plugins - compiler throws
  - linter-plugins: linter plugins - linting package with `meteor lint`
  - package-tests: change packages during hot code push
  - package-tests: show and search local package
  - cordova-plugins: remove cordova plugins
  - linter-plugins: linter plugins - linting package with `meteor lint`
  - colon-converter-tests: can't build local packages with colons
```

I haven't had a moment's chance to look into why, but just wanted to document the test retries I'd done.

arggh:
Thanks for being persistent. I was quite puzzled, trying to understand whats going on with Meteor's CI process in CircleCI. Apparently it's running some kind of in-house test suite that's failing due to timeouts: twice it seems the whole process timed out after two hours of `updating package catalog...` or some random test hitting the magically set timeout limit.

I started running the same tests locally, with same parameters. All tests were passing until:

```
linter-plugins: linter plugins - linting package with `meteor lint` ... 
  ... fail!                                                                 
  => match-timeout at ../../../../tools/tests/linter-plugins.js:123         
  => Pattern: /While linting files .* my-package .*Server/                  
  => No output  
```
...a **timeout again.** I changed the `TIMEOUT_SCALE_FACTOR` to 30 and re-run the test succesfully.

It also seems that @benjamn 's CircleCI run just timed out few seconds ago: https://circleci.com/gh/meteor/meteor/1921

arggh:
\o/ CircleCI co-operated today, @abernix.

abernix:
Woohoo!  Not saying it's related, but they did have a number of issues posted to their Status Twitter yesterday.

benjamn:
Can we just remove `Tracker._computations`? Perhaps @Slava knows more, since he added it originally? https://github.com/meteor/meteor/commit/996369e6a5b663e05bd8d8ad8910ddecfbe8ad92

@tmeasday and I are not entirely sure what timing we should be measuring with this reproduction. Can you say a little more about that, @arggh?

maxnowack:
For everyone else who is searching: #3801 is the PR where `Tracker._computations` was introduced.

arggh:
@benjamn sure: 

### Description of the reproduction/benchmark

I built that reproduction to show @mitar how slow it is to tear down a template with lots of sub-templates. The reproduction just generates a view with random data for a table (550 rows) and then uses an if-else structure to show or hide the table.

You can use Chrome's timeline tools to capture the action while _hiding_ the list with and without the fix applied, that is offered in this pull-request. Just be sure to check `JS Profile` in timeline and observe how much time your browser spends _scripting_, or further investigate the flamegraph to find out that `Tracker.Computation.stop()` is the  culprit.

### Why is it slow

The function uses `delete` to remove a property from `Tracker._computations`-object, which essentially is about 8 times slower that just setting the property to null. Using `delete` disables all kind of optimizations and makes the object "slow", and therefore should not be used in a scenario like this, where the function gets called possibly hundreds or thousands of times in a row.

The difference in my tests has been substantial. I have not yet checked in on other parts of Meteor core, maybe there are huge speed gains to be had elsewhere too...?

### How to

1. run the [reproduction](https://github.com/arggh/blaze-tmpl-destroy/tree/nosubs)
2. click "show list" and wait for Blaze to build the table with funny faces and random strings
3. the page should look like this:

tracker_test

4. Start recording the timeline with JS selected
5. Click "hide list" and feel the pain of how long it takes to have the list disappear

**First toggle, without the fix (~1.4 seconds):**
tracker_nofix1

**Second toggle, without the fix (~2.5 seconds):**
tracker_nofix2

**Third toggle, without the fix (~3.6 seconds):**
tracker_nofix3

(sidenote: I do not know, why the duration of scripting keeps growing on each iteration.)

Then if you apply the fix in Tracker...

**First toggle, with the fix (~700ms):**
tracker1

**Second toggle, with the fix (~600ms):**
tracker2

**Third toggle ~530ms**
**Fourth toggle ~600ms**
**Fifth toggle ~550ms**

and it gets rather boring after that.

(sidenote: I realized I had a minor ~100-150ms difference compared to the original benchmarks, which seems to go away if I wait a few seconds after starting the timeline profiling and clicking `hide list`. That way hiding the list takes around 350-450ms each time.)

I have also been running this fix in the actual app I'm building and the fix has made my app usable again. In one particular view, I was benchmarking 6-7 seconds of scripting time when taking down a complex table structured with Blaze. This was reduced down to ~700ms (which is still obvious when everything else in the app feels snappy and fast, but a lot better than 7 seconds).

benjamn:
My one worry with this change is that the `Tracker._computations` object will keep accumulating (tons of) keys, which is potentially a memory leak. For that reason, I would much prefer removing `Tracker._computations` altogether, since it's never been documented and no one (that we know of) uses it. Presumably doing nothing is even faster than setting `Tracker._computations[self._id] = null`.

arggh:
For sure it would be better to remove it altogether, if it's not needed. I also posted some ideas about preventing the possible memory leak 6 days ago in this PR:

> As a quick test, running
> 
> ```Tracker._computations = _.pick(Tracker._computations, function(value, key, object) {
>     return value !== null;
> });```
>
> in the end of Tracker._runFlush seems to take care of the "issue" of ever-increasing amount of keys in the Tracker._computations-object without any noticeable performance penalty. It's another story whether it's necessary, since the object will never be enumerated.

arggh:
Should I update the PR to remove computations tracking altogether?

benjamn:
Yes, please do remove it.

benjamn:
Also, thanks for the detailed explanation of why this matters!

mitar:
What about using maybe ES2016 Map instead? Maybe it has better performance then objects to store mapping. (Because there is no prototype chains and stuff.)

mitar:
BTW, this was introduced here: https://github.com/meteor/meteor/commit/996369e6a5b663e05bd8d8ad8910ddecfbe8ad92

Maybe this could break some external library extending it? This could happen even if we switch it to `Map`.

Personally it seems OK to not track this. But we should inform others of this plan and see if anyone objects.

BTW, maybe WeakMap could be used here as well. In this way if no other code is extending Tracker and reading this map, it might work faster.

benjamn:
Unfortunately, ES6 data structures (especially polyfilled) tend to have worse performance characteristics than plain objects. If you think the culprit is the prototype chain (even though `delete` does not examine the prototype chain), then the right thing to do is to use a prototype-less object: `Object.create(null)`. However, I'm still confident the right thing to do here is to eliminate `Tracker._computations`.

mitar:
Hm, what about the following idea: We wrap deletion and adding to `Tracker._computations` with if statement which checks if `Tracker._computations` itself exists. Then packages which would like to profile Tracker (which seems why this was introduced) can simply create `Tracker._computations` and things will work, but slowly. Otherwise there will be only one simple `if` check.

But I am also not opposing removing it. I am just thinking how to keep the feature around if needed.

tmeasday:
I do see a few code matches for it: https://github.com/search?utf8=✓&q=%22Tracker._computations%22&type=Code&ref=searchresults so perhaps @mitar's idea is a good one.

abernix:
Obviously it's not accounting for what people are doing privately, but [trimming down that code search a bit more](https://github.com/search?utf8=%E2%9C%93&q=%22Tracker._computations%22+language%3Ajavascript+language%3Acoffeescript+-filename%3A.cache+NOT+%22These+can+used+to+monkey-patch+computations%22&type=Code&ref=searchresults) to exclude folks who have simply checked in their entire `.meteor` directory: it seems to be @Slava, `peerlibrary` (:wave: @mitar), @aitoroses and `meteor-space`, who are using it in some form or another.

So, perhaps what @mitar said + a deprecation warning? But also maybe those few use cases aren't truly necessary.

mitar:
(BTW, I am using it in my port of Tracker to server-side. I just copied this part of code from client. But I am not really using it. I will probably do the same thing we will decide to do here to improve performance as well.)

Slava:
I am not using it in any of my active projects.

tmeasday:
So by the sounds of it just removing `_computations` is probably fine. If someone complains we could do as @mitar suggested but by the sounds of it that won't happen.

PR #7340, opened 2016-07-01
Release 1.3.4.2

benjamn:
This is a small bug fix release for issues affecting Meteor 1.3.4.1. You can try it now by running
```sh
meteor update --release 1.3.4.2
```
in any application directory.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,31 @@
 ## v.NEXT
 
-* Adds App.appendToConfig allowing custom tags in config.xml [#7307](https://github.com/meteor/meteor/pull/7307)
+## v1.3.4.2
+
+* The `meteor node` and `meteor npm` commands now respect
+  `.meteor/release` when resolving which versions of `node` and `npm` to
+  invoke. Note that you must `meteor update` to 1.3.4.2 before this logic
+  will take effect, but it will work in all app directories after
+  updating, even those pinned to older versions.
+  [#7338](https://github.com/meteor/meteor/issue/7338)
+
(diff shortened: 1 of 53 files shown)


Ben305:
It would be great if #7349 could be included

rhettlivingston:
@benjamn, I'm guessing this new .meteor/dev_bundle file should not be version controlled and that we should add it to .meteor/.gitignore?

benjamn:
Yeah, that's probably a good idea, though I don't think it would hurt anything for it to be in version control. It's logically similar to `.meteor/release`, which is not `.gitignore`d.

rhettlivingston:
I was keying on its use of an absolute path that only exists on my system.

benjamn:
Ok, you're right, I'll take another look at that tomorrow.

Another thing keeping me from recommending the 1.3.4.2 release is that my `~/.meteor/meteor` symlink is not getting updated to point to `~/.meteor/packages/meteor-tool/1.3.4_2/mt-os.osx.x86_64/meteor` after I update (even if I temporarily recommend the release).

I suppose I was overconfident about this release. I'll try to have a 1.3.4.3 release out soon.

rhettlivingston:
Yes, the link not being updated was unexpected. It was the first thing I
checked. Glad to hear you felt so too.

I'm in the process of bringing meteor/todos up to 1.3.4.x including having
it actually utilize conditional imports and an ESLint configuration that
tolerates them. I will hold off on letting the PR go until we have a
recommended 1.3.4.x beyond 1.3.4.1.

Issue #7292, opened 2016-06-24
METEOR_LOCAL_DIR fails meteor test on 1.3.3+


mikeifomin:
I tried testing with METEOR_LOCAL_DIR  on a fresh meteor(1.3.3) project and it fell with an error:

```
 fs.js:439
   return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                  ^
 Error: ENOENT, no such file or directory '.meteor/local-test/shell/info.json'
     at Object.fs.openSync (fs.js:439:18)
     at Object.fs.writeFileSync (fs.js:978:15)
     at Server.<anonymous> (/Users/mike/tmp/test-app/.meteor/local-test/build/programs/server/shell-server.js:78:10)
     at Server.g (events.js:180:16)
     at Server.emit (events.js:92:17)
     at net.js:1056:10
     at process._tickCallback (node.js:458:13)
```

With Meteor 1.3.2.4 all works fine, but 1.3.3 and beyond fails

## reproduce (on OS X and Ubuntu 14.04 )

1.3.2.4 work fine
``` bash
   meteor create test-app --release=1.3.2.4 && cd test-app && meteor add practicalmeteor:mocha && METEOR_LOCAL_DIR=.meteor/local-test meteor test --driver-package practicalmeteor:mocha
```

1.3.3 **fail** 
```
   meteor create test-app --release=1.3.3 && cd test-app && meteor add practicalmeteor:mocha && METEOR_LOCAL_DIR=.meteor/local-test meteor test --driver-package practicalmeteor:mocha
```

abernix:
`METEOR_LOCAL_DIR` was not supported in Meteor 1.3.2.4 (bfdca2aaf9df635cf7f7397719f2b7cab2596412 didn't exist) so the comment above that it worked couldn't be true.

In 1.3.3.x and 1.3.4.x, `METEOR_LOCAL_DIR` works fine, but based on the way it was written, it's necessary to specify the full path (e.g. `METEOR_LOCAL_DIR=/Users/username/your-meteor-project/.meteor/local-test`) otherwise it's probably (just guessing) trying to find it relative to `process.cwd()`.

@laosb Is this something you want to look at?

laosb:
Yeah I'm try to look at it, but these days I'm facing my Endterm Exam. I'll do further investigation after the exam, exactly this weekend.

mikeifomin:
@abernix thank you. With the absolute path all works as expected.

``` bash 
[...]

PR #7343, opened 2016-07-03
Relative path resolves to project directory when use METEOR_LOCAL_DIR.

laosb:
Issue #7292 

Use `path.resolve` to resolve the relative path to project. May need to take this down to `History.md`.

### DIFF
--- a/tools/project-context.js
+++ b/tools/project-context.js
@@ -1,5 +1,6 @@
 var assert = require("assert");
 var _ = require('underscore');
+var path = require('path');
 
 var archinfo = require('./utils/archinfo.js');
 var buildmessage = require('./utils/buildmessage.js');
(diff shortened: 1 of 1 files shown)


tools/project-context.js
```diff
@@ -89,9 +90,11 @@ _.extend(ProjectContext.prototype, {
     // writes to; used by `meteor test` so that you can test your
     // app in parallel to writing it, with an isolated database.
     // You can override the default .meteor/local by specifying
-    // METEOR_LOCAL_DIR.
+    // METEOR_LOCAL_DIR. You can use relative path if you want it
+    // relative to your project directory.
     self.projectLocalDir = process.env.METEOR_LOCAL_DIR ?
-      process.env.METEOR_LOCAL_DIR : (options.projectLocalDir ||
+      path.resolve(options.projectLocalDir, process.env.METEOR_LOCAL_DIR)
+      : (options.projectLocalDir ||
```
Slava: for consistency with the rest of the codebase this should be `files.pathResolve(options.projectLocalDir, files.convertToStandardPath(process.env.METEOR_LOCAL_DIR))`

Slava:
I can't comment if this is logically the correct fix, but the work with file-system paths should be very careful to work well on Windows. See https://github.com/meteor/meteor/tree/devel/tools/fs

laosb:
Ah, did see the README but didn't notice that `path` also wrapped in `files`. Updated the code and fixed a former mistaken.

laosb:
@Slava Tests passed!

Slava:
I will leave the opportunity for @tmeasday to look at it since I have been detached from this codebase for awhile now.

PR #7344, opened 2016-07-03
Use requestretry in http-helpers

### LINKED ISSUE #7267: Add retrying to package downloader
Currently, the Meteor package installer doesn't retry+resume downloads. Therefore, the tool can hang permanently whilst downloading packages over a flakey connection. 

We should add the ability for the tool to retry and resume package downloads. This will partially address the CDN issues reported in #6960

Using an existing node library with download resumption would be the preferred implementation.

Ben305:
Try to fix #7267 

cc @tmeasday

### DIFF
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -28,7 +28,7 @@ var packageJson = {
     underscore: "1.5.2",
     "source-map-support": "https://github.com/meteor/node-source-map-support/tarball/1912478769d76e5df4c365e147f25896aee6375e",
     semver: "4.1.0",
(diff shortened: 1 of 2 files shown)


laosb:
Do we need to expose the parameters to users so they can change them by ENV?

laosb:
Sounds like your change results in the `error: Can't install npm dependencies. Are you connected to the internet?`

tmeasday:
We'll need to build a new dev bundle for this to pass I think. Thanks for looking at it @Ben305

tmeasday:
Closing in favour of https://github.com/meteor/meteor/pull/7349

Issue #7266, opened 2016-06-21
Add retrying to installer


zol:
Currently, the Meteor install script doesn't retry+resume downloads. Therefore, downloading the Meteor distribution can fail part-way leading to an aborted installation. On a bad network connection where the download continually fails, it's impossible to install Meteor.

We should add the ability for the script to retry and resume downloads. This will partially address the CDN issues reported in #6960

## Suggested implementation
- Refactor curl to save the download to a temporary file created by the `mktemp` command rather than executing `stdout`.
- Using the method outlined [here](https://ilovesymposia.com/2013/04/11/automatically-resume-interrupted-downloads-in-osx-with-curl/), refactor the curl command to resume the download if it fails. You may need to check for multiple exit codes as the post suggests. Alternatively, you could check for a non-0 exit and retry a fixed number of times.

tmeasday:
Done. Have commented on #6960. Closing this, will reopen if it's not successful.

tmeasday:
I should change this file too https://github.com/meteor/meteor/blob/devel/scripts/admin/launch-meteor#L1

PR #7348, opened 2016-07-04
Add retrying to installer

tmeasday:
For #7266. This is code copied from install.meteor.com. The script seems to work, as far as I can test it.

### DIFF
--- a/scripts/admin/launch-meteor
+++ b/scripts/admin/launch-meteor
@@ -82,15 +82,36 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
   fi
   echo "Installing a Meteor distribution in your home directory." 1>&2
 
-  # Only show progress bar animations if we have a tty
-  # (Prevents tons of console junk when installing within a pipe)
-  if [[ -t 1 ]]; then
-    curl --progress-bar --fail "$TARBALL_URL" | tar -xzf - -C "$INSTALL_TMPDIR"
-  else
-    curl -s --fail "$TARBALL_URL" | tar -xzf - -C "$INSTALL_TMPDIR"
-  fi
+  TARBALL_FILE="$(dirname "$METEOR_WAREHOUSE_DIR")/.meteor-tarball-tmp"
+  rm -rf "$TARBALL_FILE"
+
+  echo "Downloading Meteor distribution"
+  # keep trying to curl the file until it works (resuming where possible)
+  MAX_ATTEMPTS=99
+  set +e
+  CODE=1
+  ATTEMPTS=0
+  while [ $CODE -ne 0 ] && [ $ATTEMPTS -lt $MAX_ATTEMPTS ]
+  do
+    ATTEMPTS=$((ATTEMPTS + 1))
+
+    # Only show progress bar animations if we have a tty
+    # (Prevents tons of console junk when installing within a pipe)
+    if [[ -t 1 ]]; then
+      curl --progress-bar --fail "$TARBALL_URL" -C - -o "$TARBALL_FILE"
+    else
+      curl -s --fail "$TARBALL_URL" -C - -o "$TARBALL_FILE"
+    fi
+
+    CODE=$?
+  done
+  set -e
 
   # bomb out if it didn't work, eg no net
+  test -e "${TARBALL_FILE}"
+  tar -xzf "$TARBALL_FILE" -C "$INSTALL_TMPDIR" -o
+  rm -rf "$TARBALL_FILE"
+
   test -x "${INSTALL_TMPDIR}/.meteor/meteor"
   mv "${INSTALL_TMPDIR}/.meteor" "$METEOR_WAREHOUSE_DIR"
   rmdir "${INSTALL_TMPDIR}"
--- a/scripts/admin/launch-meteor
+++ b/scripts/admin/launch-meteor
@@ -97,7 +97,7 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
 
     # Only show progress bar animations if we have a tty
     # (Prevents tons of console junk when installing within a pipe)
-    if [[ -t 1 ]]; then
+    if [ -t 1 ]; then
       curl --progress-bar --fail "$TARBALL_URL" -C - -o "$TARBALL_FILE"
     else
       curl -s --fail "$TARBALL_URL" -C - -o "$TARBALL_FILE"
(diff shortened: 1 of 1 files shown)


scripts/admin/launch-meteor
```diff
@@ -82,15 +82,36 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
...
+  set +e
+  CODE=1
+  ATTEMPTS=0
+  while [ $CODE -ne 0 ] && [ $ATTEMPTS -lt $MAX_ATTEMPTS ]
+  do
+    ATTEMPTS=$((ATTEMPTS + 1))
+
+    # Only show progress bar animations if we have a tty
+    # (Prevents tons of console junk when installing within a pipe)
+    if [[ -t 1 ]]; then
```
abernix: The rest of the additions look POSIX-compatible, but because `/bin/sh` isn't always `bash` (GNU), but sometimes `dash` [such as on Ubuntu](https://wiki.ubuntu.com/DashAsBinSh), the `[[` will break things (and currently is, see #7353).

tmeasday: Whoa, thanks @abernix, what would we do without you :)

Just to fill you in: this is the script installed at `/usr/local/bin/meteor`, which is run with `bash` I guess, thanks to the shebang.

The script at install.meteor.com is similar but part of a private repo (the package server). I foolishly tried to unify them :/

abernix: Yeah, I just sussed out this loader a couple weeks ago, but I incorrectly assumed the shebang was the same as install.meteor.com.  So I guess this is fine – though generally it's best to keep POSIX-compliant – though I find it highly frustrating exactly because of issues like this. :wink:

abernix: Btw, on the `install.meteor.com` script, I just noticed that you completely removed the "IF TTY USE `progress-bar`" conditional.  You should be fine to continue using that so long as you only use single brackets instead of double-brackets:

    if [ -t 1 ]; then

Which I can confirm works in Debian properly when running the command without an interactive terminal (via SSH).  I didn't test with anything else (Docker, for example), but it should be the same as there is no STDIN descriptor.

(Also, I realize this is the wrong issue to discuss the `install.meteor.com` script on, however I don't have a better PR for it since it's not in the repo. :wink:)

tmeasday: Yeah, I just reverted it quickly. I'll give that a try (at a more considered pace this time).

tmeasday:
@benjamn I'll ship this change to the installer first and ping back here in a day or two when we know it has no problems.

scripts/admin/launch-meteor
tmeasday: Deployed now. Let me know if you hear of any problems.

scripts/admin/launch-meteor
```diff
@@ -82,15 +82,36 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
...
   echo "Installing a Meteor distribution in your home directory." 1>&2
 
-  # Only show progress bar animations if we have a tty
-  # (Prevents tons of console junk when installing within a pipe)
-  if [[ -t 1 ]]; then
-    curl --progress-bar --fail "$TARBALL_URL" | tar -xzf - -C "$INSTALL_TMPDIR"
-  else
-    curl -s --fail "$TARBALL_URL" | tar -xzf - -C "$INSTALL_TMPDIR"
-  fi
+  TARBALL_FILE="$(dirname "$METEOR_WAREHOUSE_DIR")/.meteor-tarball-tmp"
```
benjamn: Can we name this file in a way that makes it unique for a given `$TARBALL_URL`, perhaps by hashing the URL and appending the hash? I can imagine someone failing to download one tarball and trying a different version, only to end up appending to the old temporary file, resulting in an invalid download.

scripts/admin/launch-meteor
```diff
@@ -82,15 +82,36 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
...
+    fi
+
+    CODE=$?
+  done
+  set -e
 
   # bomb out if it didn't work, eg no net
+  test -e "${TARBALL_FILE}"
+  tar -xzf "$TARBALL_FILE" -C "$INSTALL_TMPDIR" -o
+  rm -rf "$TARBALL_FILE"
```
benjamn: We should `rm -rf "$TARBALL_FILE"` even if (especially if) the extraction failed.

benjamn: Never mind this; we `rm -rf` it above, before we try to download anything.

scripts/admin/launch-meteor
```diff
@@ -98,9 +98,11 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
     # Only show progress bar animations if we have a tty
     # (Prevents tons of console junk when installing within a pipe)
     if [ -t 1 ]; then
-      curl --progress-bar --fail "$TARBALL_URL" -C - -o "$TARBALL_FILE"
+      curl --progress-bar --fail --continue-at - \
+        "$TARBALL_URL" --output "$TARBALL_FILE"
     else
-      curl -s --fail "$TARBALL_URL" -C - -o "$TARBALL_FILE"
+      curl --silent --fail --continue-at - \
+        "$TARBALL_URL" --output "$TARBALL_FILE"
```
tmeasday: Should we just switch the "verbosity" part out with something like

```sh
VERBOSITY="--silent";
if [ -t 1 ]; then
  VERBOSITY="--progressbar
fi

# later

curl $VERBOSITY --fail...
```

scripts/admin/launch-meteor
```diff
@@ -82,7 +82,10 @@ if [ ! -x "$METEOR_WAREHOUSE_DIR/meteor" ]; then
   fi
   echo "Installing a Meteor distribution in your home directory." 1>&2
 
-  TARBALL_FILE="$(dirname "$METEOR_WAREHOUSE_DIR")/.meteor-tarball-tmp"
+  # Generate the $TARBALL_FILE path based on $TARBALL_URL, but with unsafe
+  # characters replaced by underscores.
+  PART_FILE=".meteor-${TARBALL_URL//[^A-Za-z0-9_.-]/_}.part"
+  TARBALL_FILE="$(dirname "$METEOR_WAREHOUSE_DIR")/${PART_FILE}"
```
tmeasday: What's the purpose of naming the part file based on the tarball, given the file is going to be deleted at the end of the script no matter what?

scripts/admin/launch-meteor
benjamn: That seems nice and DRY :)

scripts/admin/launch-meteor
benjamn: You're right, it probably doesn't matter anymore if you can always just retry the install.

tmeasday:
I think this is ready to go. I've just deployed the equivalent code to install.meteor.com

PR #7349, opened 2016-07-04
7267 use request retry

tmeasday:
Based on #7344. Use the `requestretry` package to make package downloads more resistant to CDN/internet flakiness.

We should consider what values to use for `maxAttempts` and `retryDelay`.

### DIFF
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=0.6.16
+BUNDLE_VERSION=0.6.17
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=0.6.15
+BUNDLE_VERSION=0.6.16
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -28,7 +28,7 @@ var packageJson = {
     underscore: "1.5.2",
     "source-map-support": "https://github.com/meteor/node-source-map-support/tarball/1912478769d76e5df4c365e147f25896aee6375e",
     semver: "4.1.0",
-    request: "2.47.0",
+    requestretry: "1.9.0",
     fstream: "https://github.com/meteor/fstream/tarball/d11b9ec4a13918447c8af7559c243c190744dd1c",
     tar: "1.0.2",
     kexec: "2.0.2",
--- a/tools/utils/http-helpers.js
+++ b/tools/utils/http-helpers.js
@@ -256,10 +256,18 @@ _.extend(exports, {
       delete options.timeout;
     }
 
+    // Configure requestretry
+    if (! _.has(options, "maxAttempts")) {
+      options.maxAttempts = 3;   // try 3 times
+    }
+    if (! _.has(options, "retryDelay")) {
+      options.retryDelay = 30 * 1000;  // wait for 30s before trying again
+    }
+
     // request is the most heavy-weight of the tool's npm dependencies; don't
     // require it until we definitely need it.
     Console.debug("Doing HTTP request: ", options.method || 'GET', options.url);
-    var request = require('request');
+    var request = require('requestretry');
     var req = request(options, callback);
 
 


n1mmy:
@tmeasday It doesn't look like this resumes downloads where the left off on failure.

It still probably helps some people, but I wouldn't be surprised if the most affected people just have their downloads fail over and over.

tmeasday:
@n1mmy I think you are right. Looking around I don't see any node packages that support resuming requests.

Based on my testing, I do suspect that the size of the package tarballs (which are significantly smaller than the full distribution) might mean that resuming is overkill anyway?

So I think this patch might be useful anyway. But it's hard to say.

abernix:
I've heard plenty of users complaining about inability to install packages so I think, as is, this alone is worth a try.

Came across these two packages.  They're not super popular/stable-looking, but perhaps they might work.

* https://www.npmjs.com/package/fast-download
* https://www.npmjs.com/package/jswget

n1mmy:
The issue comes up mostly with the meteor tool package, which is over 100MB =(

Ben305:
I also would like to see this shipped, because the changes won't made the download problems worse, but probably solve some of them.

laosb:
In China, even `kadira:flow-router` can't be downloaded without resume. Resuming is necessary, even not for packages like `meteor-tool`.

tmeasday:
@abernix unfortunately things that aren't drop-in replacements for request would be difficult to integrate with this code: https://github.com/meteor/meteor/blob/master/tools/utils/http-helpers.js#L123

I suspect we may need to alter that code to support resuming to/from a file, basically copying the pattern from here: https://github.com/zenflow/node-fast-download/blob/master/lib/FastDownload.js#L30 and https://github.com/zenflow/node-fast-download/blob/master/lib/FastDownload.js#L63-L68

tmeasday:
Note having looked closer at this code that we should ensure we QA the network related parts of the tool (uploading packages / apps to galaxy / release-for-arch / downloading) before we release this.

tmeasday:
Also, interestingly, downloads are **significantly** faster when I use http not https.

abernix:
@tmeasday Re: https: If it's consistently slower when making successive requests, then that is strange and the rest of what I'm about to say is probably not relevant, but if the bundles are getting purged from the CDN edge caches, it could likely be the https overhead when fetching, decrypting, re-encrypting & forwarding objects from the origin server.  I'm not sure what the cache settings/TTLs on the bundles are, but that could have some effect, I suppose.

That said, I don't seem to be seeing the same issue.  In fact, this shows my https is consistently the same or faster.  Here are some benchmarks of me downloading `kadira:flow-router` (16 MBytes) 50 times on a 100Mbps connection with HTTPS vs HTTP for each of: 1) `wget`, 2) [`download-cli`](https://www.npmjs.com/package/download-cli) (which uses [`got`](https://www.npmjs.com/package/got) behind the scenes) and 3) [`request-cli`](https://www.npmjs.com/package/request-cli) (which uses `request` BTS).

Times listed are average and in parenthesis is stddev %.

#### HTTPS

##### `wget`
```sh
$ perf stat -r 50 -d wget --quiet https://warehouse.meteor.com/builds/mjvBCnpwsLFcXbjzt/1461283328652/ny7jr9wBv9/kadira:flow-router-2.12.1-os+web.browser+web.cordova.tgz
... snip ...
       0.334198264 seconds time elapsed                                          ( +-  8.37% )
```

##### `download-cli`
```sh
$ perf stat -r 50 -d download http://warehouse.meteor.com/builds/mjvBCnpwsLFcXbjzt/1461283328652/ny7jr9wBv9/kadira:flow-router-2.12.1-os+web.browser+web.cordova.tgz
... snip ...
       1.101198478 seconds time elapsed                                          ( +-  3.59% )
```

##### `request-cli`
```sh
$ perf stat -r 50 -d req https://warehouse.meteor.com/builds/mjvBCnpwsLFcXbjzt/1461283328652/ny7jr9wBv9/kadira:flow-router-2.12.1-os+web.browser+web.cordova.tgz
... snip ...
       1.425188913 seconds time elapsed                                          ( +-  0.58% )
```

#### HTTP

##### `wget`
```sh
$ perf stat -r 50 -d wget --quiet http://warehouse.meteor.com/builds/mjvBCnpwsLFcXbjzt/1461283328652/ny7jr9wBv9/kadira:flow-router-2.12.1-os+web.browser+web.cordova.tgz
... snip ...
       0.549323766 seconds time elapsed                                          ( +-  5.49% )
```

##### `download-cli`
```sh
$ perf stat -r 50 -d download http://warehouse.meteor.com/builds/mjvBCnpwsLFcXbjzt/1461283328652/ny7jr9wBv9/kadira:flow-router-2.12.1-os+web.browser+web.cordova.tgz
... snip ...
       1.103604489 seconds time elapsed                                          ( +-  6.37% )
```

##### `request-cli`
```sh
$ perf stat -r 50 -d req http://warehouse.meteor.com/builds/mjvBCnpwsLFcXbjzt/1461283328652/ny7jr9wBv9/kadira:flow-router-2.12.1-os+web.browser+web.cordova.tgz
... snip ...
       1.599599134 seconds time elapsed                                          ( +-  1.62% )
```

benjamn:
A better version of this PR was merged from https://github.com/meteor/meteor/pull/7399. Specifically, that version includes support for resuming partial downloads.

Issue #7338, opened 2016-07-01
NPM and Node versions don't follow project's Meteor release


rhettlivingston:
I'm using multiple versions of Meteor including 1.3.0, 1.3.4.1, and 1.4-beta.6 for this example. My development system is on Ubuntu 16.04, so this isn't a Windows problem.

I've read time and time again that a big advantage of using 'meteor npm' instead of 'npm' in a project is assuring that you are using the right version. So, I do.

While debugging an issue in the meteor/todos configuration, I realized that a component of the problem I was battling was that I was running NPM 3.9.6 when using 'meteor npm' in the meteor/todos project which is a meteor v1.3.0 project (as indicated by .meteor/release).

I then blew away my ~/.meteor directory to clear out all of my meteor versions and get back to ground zero and downloaded a clean copy of meteor/todos. When I ran 'meteor npm install' in the todos, it correctly used Node 0.10.45 and NPM 2.15.1.

Then I downloaded a clean copy of a project that I have that is based on 1.3.4.1 and switched to that directory. When I ran 'meteor npm install' in that directory, it again used Node 0.10.45 and NPM 2.15.1. However, it should be using NPM 3.9.6.

I ran 'meteor' in the 1.3.4.1 app's directory and it went through downloading the 1.3.4.1 versions of packages.

After that, I got the following

```bash
rhett@dreamshot:~/oob/meteors/oob$ meteor --version
Meteor 1.3.4.1
rhett@dreamshot:~/oob/meteors/oob$ meteor npm -v
2.15.1
rhett@dreamshot:~/oob/meteors/oob$ meteor node -v
v0.10.45
rhett@dreamshot:~/oob/meteors/oob$ cd ../todos
rhett@dreamshot:~/oob/meteors/todos$ meteor --version
Meteor 1.3                                    
rhett@dreamshot:~/oob/meteors/todos$ meteor npm -v
2.15.1
rhett@dreamshot:~/oob/meteors/todos$ meteor node -v
v0.10.45
```

I then switched to a clean project that is set to release 1.4-beta.6 and in fact has a NPM package that has a hard dependency on Node 4. Guess what.

```bash
rhett@dreamshot:~/oob1.4/meteors/oob$ meteor --version
Meteor 1.4-beta.6                             
rhett@dreamshot:~/oob1.4/meteors/oob$ meteor npm -v
2.15.1
rhett@dreamshot:~/oob1.4/meteors/oob$ meteor node -v
v0.10.45
```

[...]

PR #7369, opened 2016-07-08
Release 1.3.4.3

benjamn:
Unfortunately, the following claim in the [release notes](https://github.com/meteor/meteor/blob/devel/History.md#v1342) for the [1.3.4.2 release](https://github.com/meteor/meteor/pull/7340) was not entirely true:

* The `meteor node` and `meteor npm` commands now respect `.meteor/release` when resolving which versions of `node` and `npm` to invoke. Note that you must `meteor update` to 1.3.4.2 before this logic will take effect, but it will work in all app directories after updating, even those pinned to older versions. [#7338](https://github.com/meteor/meteor/issue/7338)

Specifically, running `meteor update` to update to 1.3.4.2 was not actually enough to modify the `~/.meteor/meteor` symlink (or `AppData\Local\.meteor\meteor.bat` on Windows), which prevented the new behavior of `meteor node` and `meteor npm` from taking effect.

Thanks to @rhettlivingston for providing [quick feedback](https://github.com/meteor/meteor/pull/7340#issuecomment-231253317) on these concerns!

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,27 @@
 ## v.NEXT
 
+## v1.3.4.3
+
+* Node has been upgraded to 0.10.46.
+
+* `npm` has been upgraded to 3.10.5.
+
+* The `node-gyp` npm package has been upgraded to 3.4.0.
+
+* The `node-pre-gyp` npm package has been upgraded to 0.6.29.
+
+* The `~/.meteor/meteor` symlink (or `AppData\Local\.meteor\meteor.bat` on
+  Windows) will now be updated properly after `meteor update` succeeds. This was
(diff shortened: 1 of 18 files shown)


kachkaev:
@benjamn is there a chance of releasing another `1.4-beta` with this problem solved? Looks like I've got some problems because `meteor node --version` is still `v0.10.45` in `beta.7`.

If a release not require any complex merging, of course.

benjamn:
@kachkaev of course! I haven't even rebased https://github.com/meteor/meteor/tree/release-1.4 against these changes yet :)

PR #7373, opened 2016-07-09
Ability to override the default warehouse url base

laosb:
> Discussed about it at #6944. This PR gives you ability to override the default warehouse urlbase (warehouse.meteor.com) by setting METEOR_WAREHOUSE_URLBASE.

Previous PR #7054 didn't achieve this and hope this will work.

PS: Does this function only used to download packages? Will the url not begins with `https://warehouse.meteor.com`?

cc @n1mmy

### DIFF
--- a/tools/packaging/tropohouse.js
+++ b/tools/packaging/tropohouse.js
@@ -299,6 +299,12 @@ _.extend(exports.Tropohouse.prototype, {
   // XXX: Error handling.
   _downloadBuildToTempDir: function (versionInfo, buildRecord) {
     var url = buildRecord.build.url;
+    
+    // Override the download domain name and protocol if METEOR_WAREHOUSE_URLBASE
+    // provided.
+    if (process.env.METEOR_WAREHOUSE_URLBASE) {
+      url = url.replace(/^[a-zA-Z]+:\/\/[^\/]+/, process.env.METEOR_WAREHOUSE_URLBASE);
+    }
 
     // XXX: We use one progress for download & untar; this isn't ideal:
     // it relies on extractTarGz being fast and not reporting any progress.


laosb:
@meteor Any thoughts?

zol:
@laosb please refrain from direct mentioning folks on issues. I've mixed feelings about this PR and am now wondering whether merging #7054 was the correct thing to do. @laosb as far as I know you're the only one who wants to use this feature. Can you confirm that you've tested this PR and merging it will result in `METEOR_WAREHOUSE_URLBASE` working as intended?

laosb:
Sorry for the hurry but actually I have difficulties in debugging with unreleased version, as the meteor/meteor is a huge repo which taks my whole life to clone. 

Still trying to clone and I'll notify you when I tested it out. Sorry!

laosb:
@zol Sorry for the hurry and it does work on my computer. But to be careful, I'm not sure if other downloads use this function and also downloads from `warehouse.meteor.com`. It works as a hack since I just hijack the address.

abernix:
@laosb So the reason for "hacking" it with this regex is because `packages.meteor.com` returns the _full_ `url` with `https://warehouse.meteor.com/` in the data that it sends, correct?

In general, how is your connectivity to the `packages.meteor.com` server?  Is that going to be a problem as well?  (Since it's the first obstacle that must be overcome before `warehouse.meteor.com`).

laosb:
Yes for the first problem.

In China, both packages.meteor.com and warehouse.meteor.com wasn't blocked, but all foreign sites were limited in speed. The longer the connection was, the lower speed you'll get. So long as we download heavy files from warehouse.meteor.com, the connection may break as we have very low speed. But with packages.meteor.com, in most cases we don't download much so it won't be a problem.

zol:
Ok, I'm going to merge this.

PR #7399, opened 2016-07-13
Package download request resuming

tmeasday:
For #7267 

I wanted to get your thoughts on this approach @benjamn.

Todo:

 - [x] Ensure it actually works with package downloading
 - [x] Figure out if there's a way to make the test meaningful
 - [x] Add a timeout to the retries
 - [x] Better logging
 - [x] Add `stream-buffers` to the dev bundle

### DIFF
--- a/tools/packaging/tropohouse.js
+++ b/tools/packaging/tropohouse.js
@@ -304,7 +304,7 @@ _.extend(exports.Tropohouse.prototype, {
     // it relies on extractTarGz being fast and not reporting any progress.
     // Really, we should create two subtasks
     // (and, we should stream the download to the tar extractor)
-    var packageTarball = httpHelpers.getUrl({
+    var packageTarball = httpHelpers.getUrlWithResuming({
       url: url,
       encoding: null,
       progress: buildmessage.getCurrentProgressTracker(),
--- a/tools/tests/utils-tests.js
+++ b/tools/tests/utils-tests.js
@@ -152,3 +152,30 @@ selftest.define("parse url", function () {
     protocol: "https"
   });
 });
+
+// XXX: WIP -- I'm not sure how to properly test a failed connection.
+// At the moment, I'm thinking I'll just mark this test as "don't run"
+// and use it by hand for testing (which basically means run it, kill my
+// internet, start my internet again, watch what happens)
+selftest.define("resume downloads", function () {
+  const url = 'http://warehouse.meteor.com/builds/Pr7L8f6PqXyqNJJn4/1443478653127/aRiirNrp4v/meteor-tool-1.1.9-os.osx.x86_64+web.browser+web.cordova.tgz';
+
+  const result = require('../utils/http-helpers').getUrlWithResuming({
+    timeout: 1000,
+    url: url,
+    encoding: null,
+    wait: false,
+    progress: {
+      reportProgress({ current, end }) {
+        const percent = current / end * 100;
+        if (Math.random() < 0.1) {
+          console.log(`${percent} %`);
+        }
+      },
+      reportProgressDone() {
+        console.log('done');
+      }
+    }
+  });
+  console.log(result, result.toString().length)
+});
--- a/tools/tests/utils-tests.js
+++ b/tools/tests/utils-tests.js
@@ -1,6 +1,8 @@
 var selftest = require('../tool-testing/selftest.js');
 var utils = require('../utils/utils.js');
 
+import httpHelpers from '../utils/http-helpers';
+
 selftest.define('subset generator', function () {
   var out = [];
   utils.generateSubsetsOfIncreasingSize(['a', 'b', 'c'], function (x) {
@@ -157,9 +159,6 @@ selftest.define("resume downloads", ['net', 'slow'], function () {
   // A reasonably big file that (I think) should take more than 1s to download
   // and that we know the size of
   const url = 'http://warehouse.meteor.com/builds/Pr7L8f6PqXyqNJJn4/1443478653127/aRiirNrp4v/meteor-tool-1.1.9-os.osx.x86_64+web.browser+web.cordova.tgz';
-
-  let firstPass = true;
-  const httpHelpers = require('../utils/http-helpers');
 
   setTimeout(() => {
     httpHelpers._currentRequest.emit('error', 'pretend-http-error');
--- a/tools/tests/utils-tests.js
+++ b/tools/tests/utils-tests.js
@@ -153,14 +153,23 @@ selftest.define("parse url", function () {
   });
 });
 
-// XXX: WIP -- I'm not sure how to properly test a failed connection.
-// At the moment, I'm thinking I'll just mark this test as "don't run"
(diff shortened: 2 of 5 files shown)


tools/utils/http-helpers.js
```diff
@@ -358,6 +369,47 @@ _.extend(exports, {
...
+    const MAX_ATTEMPTS = 5;
+    function attempt(triesRemaining) {
+      const headers = _.extend({}, urlOrOptions.headers, {
+        range: `bytes=${outputStream.size()}-`,
+      });
+      try {
+        return httpHelpers.request({
+          outputStream,
+          headers,
+          ...urlOrOptions,
```
tmeasday: XXX: should be `options`

tools/utils/http-helpers.js
```diff
@@ -358,6 +369,47 @@ _.extend(exports, {
...
+      url: urlOrOptions,
+    };
+
+    const outputStream = new WritableStreamBuffer();
+
+    const MAX_ATTEMPTS = 5;
+    function attempt(triesRemaining) {
+      const headers = _.extend({}, urlOrOptions.headers, {
+        range: `bytes=${outputStream.size()}-`,
+      });
```
benjamn:
```js
const headers = {
  ...urlOrOptions.headers,
  range: `bytes=${outputStream.size()}-`
};
```

tools/utils/http-helpers.js
```diff
@@ -358,6 +369,47 @@ _.extend(exports, {
...
+        return httpHelpers.request({
+          outputStream,
+          headers,
+          ...urlOrOptions,
+        });
+      } catch (e) {
+        console.log(e);
+        if (triesRemaining > 0) {
+          console.log(`Request failed, retrying`);
+          return attempt(triesRemaining - 1);
```
benjamn: What if the failure only counted against `triesRemaining` if the download made no progress (or less than some small amount of progress)?

tools/tests/utils-tests.js
```diff
@@ -152,3 +152,30 @@ selftest.define("parse url", function () {
...
+
+// XXX: WIP -- I'm not sure how to properly test a failed connection.
+// At the moment, I'm thinking I'll just mark this test as "don't run"
+// and use it by hand for testing (which basically means run it, kill my
+// internet, start my internet again, watch what happens)
+selftest.define("resume downloads", function () {
+  const url = 'http://warehouse.meteor.com/builds/Pr7L8f6PqXyqNJJn4/1443478653127/aRiirNrp4v/meteor-tool-1.1.9-os.osx.x86_64+web.browser+web.cordova.tgz';
+
+  const result = require('../utils/http-helpers').getUrlWithResuming({
+    timeout: 1000,
```
benjamn: This should actually test resuming, perhaps by supporting a custom `options.outputStream` object, and passing a test object whose `.write` method throws an HTTP error after the first few chunks are written?

tools/utils/http-helpers.js
```diff
@@ -14,6 +14,7 @@ var release = require('../packaging/release.js');
 var Console = require('../console/console.js').Console;
 var timeoutScaleFactor = require('./utils.js').timeoutScaleFactor;
 
+import { WritableStreamBuffer } from 'stream-buffers';
```
benjamn: The `stream-buffers` package needs to be in `dev_bundle/lib/node_modules`.

tools/utils/http-helpers.js
```diff
@@ -358,6 +369,47 @@ _.extend(exports, {
...
+      }
+    }
+    const response = attempt(MAX_ATTEMPTS).response;
+    const body = outputStream.getContents();
+
+    if (response.statusCode >= 400 && response.statusCode < 600) {
+      const href = response.request.href;
+      throw Error(
+         body ||
+          `Could not get ${href}; server returned [${response.statusCode}]`);
```
benjamn: Maybe this was just for debugging, but I would use the same error message every time, regardless of whether `body` is defined, and then just `return outputStream.getContents()` below, if there was no error.

tools/utils/http-helpers.js
benjamn: Ah, sorry, I see that in your to-do list now.

benjamn:
I like the idea of doing this in memory!

tools/utils/http-helpers.js
benjamn: Perhaps there should be a delay of a few seconds between attempts?

tools/utils/http-helpers.js
benjamn: If you use `options` here, the `options.headers` property will override the `headers` property, so you might as well update `options.headers` above, and remove the `headers,` line here.

tools/utils/http-helpers.js
```diff
@@ -358,6 +369,47 @@ _.extend(exports, {
...
+    const options = _.isObject(urlOrOptions) ? urlOrOptions : {
+      url: urlOrOptions,
+    };
+
+    const outputStream = new WritableStreamBuffer();
+
+    const MAX_ATTEMPTS = 5;
+    function attempt(triesRemaining) {
+      const headers = _.extend({}, urlOrOptions.headers, {
+        range: `bytes=${outputStream.size()}-`,
```
benjamn: Should `range` be capitalized here?

tools/tests/utils-tests.js
tmeasday: Oooh, that might work, let me try it.

tools/utils/http-helpers.js
tmeasday: Yeah I just wanted to check with you that the concept of using a stream buffer to collect the body made sense before bothering to build a new bundle etc.

tools/utils/http-helpers.js
tmeasday: That's part of the plan, yeah.

tools/utils/http-helpers.js
tmeasday: Yeah, I was just re-implementing what `getUrl` does above, but I agree it's sort of weird.

tools/utils/http-helpers.js
tmeasday: :+1:

tools/utils/http-helpers.js
tmeasday: :+1:

tools/utils/http-helpers.js
tmeasday: 👍

tmeasday:
@ben I think this should be good to go, awaiting tests

tools/packaging/tropohouse.js
```diff
@@ -304,7 +304,7 @@ _.extend(exports.Tropohouse.prototype, {
     // it relies on extractTarGz being fast and not reporting any progress.
```
abernix: It's just off the top of the diff here, but I think this `XXX/TODO` would be a huge benefit as it's getting hard to debug when something is stuck `downloading` or something is stuck `extracting`.

I'm getting the feeling that people are getting stuck _after_ the actual download, especially on Windows.

Issue #7297, opened 2016-06-24
1.4-beta.1 - fourseven:scss package broken


mikestokes:
Running OS X El Capitan and upgrading to 1.4-beta.1 there is an issue that seems to be related to the npm and/or node update with the fourseven:scss package for compiling sass files.

Re-creation:

1. Upgrade to Meteor 1.4-beta.1
2. Install the fourseven:scss package using the "meteor add" command
3. Remove NPM packages folder: rm -rf node_modules
4. Install NPM packages: meteor npm install
5. Run meteor

`[[[[[ ~/git/fletchers/fd.shopfloor.meteor ]]]]]

=> Started proxy.
=> Started MongoDB.
=> Errors prevented startup:

   While loading plugin `compileScssBatch` from package `fourseven:scss`:

   /Users/mike/.meteor/packages/fourseven_scss/.3.4.3.fcqret++os.osx.x86_64+web.browser+web.cordova/plugin.compileScssBatch.os.osx.x86_64/npm/node_modules/meteor/compileScssBatch/node_modules/node-sass/lib/extensions.js:158:11:
   The `libsass` binding was not found in
   /Users/mike/.meteor/packages/fourseven_scss/.3.4.3.fcqret++os.osx.x86_64+web.browser+web.cordova/plugin.compileScssBatch.os.osx.x86_64/npm/node_modules/meteor/compileScssBatch/node_modules/node-sass/vendor/darwin-x64-46/binding.node
   This usually happens because your node version has changed.
   Run `npm rebuild node-sass` to build the binding for your current node
   version.
   at Object.sass.getBinaryPath
   (/Users/mike/.meteor/packages/fourseven_scss/.3.4.3.fcqret++os.osx.x86_64+web.browser+web.cordova/plugin.compileScssBatch.os.osx.x86_64/npm/node_modules/meteor/compileScssBatch/node_modules/node-sass/lib/extensions.js:158:11)
   at Object.<anonymous>
   (/Users/mike/.meteor/packages/fourseven_scss/.3.4.3.fcqret++os.osx.x86_64+web.browser+web.cordova/plugin.compileScssBatch.os.osx.x86_64/npm/node_modules/meteor/compileScssBatch/node_modules/node-sass/lib/index.js:16:36)
   at Module._compile (module.js:409:26)
   at Object.Module._extensions..js (module.js:416:10)
   at Module.load (module.js:343:32)
   at Module.Mp.load
[...]

PR #7218, opened 2016-06-14
Release 1.4

benjamn:
Now that [Meteor 1.3.3 is out](https://forums.meteor.com/t/announcing-meteor-1-3-3-ddp-batching-babelrc-support-improved-import-declarations/24874), the [`release-1.4`](https://github.com/meteor/meteor/tree/release-1.4) branch should target [`devel`](https://github.com/meteor/meteor/tree/devel) instead of [`release-1.3.3`](https://github.com/meteor/meteor/tree/release-1.3.3). However, because GitHub does not support changing the target branch of a pull request (despite [significant demand](https://github.com/isaacs/github/issues/18)), I had to open this new pull request.

[Here](https://github.com/meteor/meteor/pull/7035) is the original pull request for Meteor 1.4, and here is the description from that PR:

> This pull request is an easy way to collect changes that will be going into Meteor 1.4 and run tests against them in an isolated, consistent environment (CircleCI and TravisCI).

> At this time, Meteor 1.4 comes with Node 4.4.5. If Node 6 stabilizes soon enough, Meteor 1.4 may ship with Node 6 instead. If not, you will nevertheless be able to use Node 6 with Meteor 1.4 if you're feeling adventurous, as many of the improvements made to enable the Node 4.x upgrade also support Node 6.

> Milestone: https://github.com/meteor/meteor/milestones/Release%201.4

PR #7407, opened 2016-07-14
Release 1.3.5

benjamn:
Various changes cherry-picked from the Meteor 1.4 release branch (#7218):

* Failed Meteor package downloads are now automatically resumed from the point of failure, up to ten times, with a five-second delay between attempts. [#7399](https://github.com/meteor/meteor/pull/7399)

* If an app has no `package.json` file, all packages in `node_modules` will be built into the production bundle. In other words, make sure you have a `package.json` file if you want to benefit from `devDependencies` pruning. 7b2193188fc9e297eefc841ce6035825164f0684

* Binary npm dependencies of compiler plugins are now automatically rebuilt when Node/V8 versions change. [#7297](https://github.com/meteor/meteor/issues/7297)

* Because `.meteor/local` is where purely local information should be stored, the `.meteor/dev_bundle` link has been renamed to `.meteor/local/dev_bundle`.

* The `.meteor/local/dev_bundle` link now corresponds exactly to `.meteor/release` even when an app is using an older version of Meteor. d732c2e649794f350238d515153f7fb71969c526

* When recompiling binary npm packages, the `npm rebuild` command now receives the flags `--update-binary` and `--no-bin-links`, in addition to respecting the `$METEOR_NPM_REBUILD_FLAGS` environment variable. [#7401](https://github.com/meteor/meteor/issues/7401)

* The last solution found by the package version constraint solver is now stored in `.meteor/local/resolver-result-cache.json` so that it need not be recomputed every time Meteor starts up. 69b100d21a33fd7ca58fc2fe2d8997ade4a36af6

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,7 @@
 dependencies:
+  pre:
+    # https://github.com/meteor/docs/blob/version-NEXT/long-form/file-change-watcher-efficiency.md
+    - echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
   cache_directories:
     - "dev_bundle"
     - ".meteor"
--- a/History.md
+++ b/History.md
@@ -2,6 +2,10 @@
 
 ## v1.3.5
 
+* Failed Meteor package downloads are now automatically resumed from the
+  point of failure, up to ten times, with a five-second delay between
+  attempts. [#7399](https://github.com/meteor/meteor/pull/7399)
+
 * If an app has no `package.json` file, all packages in `node_modules`
   will be built into the production bundle. In other words, make sure you
   have a `package.json` file if you want to benefit from `devDependencies`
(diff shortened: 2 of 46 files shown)


circle.yml
```diff
@@ -1,4 +1,7 @@
 dependencies:
+  pre:
+    # https://github.com/meteor/docs/blob/version-NEXT/long-form/file-change-watcher-efficiency.md
+    - echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```
benjamn: This didn't actually work because `fs.inotify.max_user_watches` is not a key that can be modified (even as root) in a Circle CI container. I would revert this commit, but it doesn't seem to do any harm. The tests still run.

benjamn:
Having a lot of trouble with the package server today. Not sure why, but the tool often gets stuck at `Updating package catalog`.

benjamn:
Managed to repair the package server! Back to our regularly scheduled release work :)

PR #7411, opened 2016-07-14
Added postSignUpHook call in the AccountsServer.updateOrCreateUserFromExternalService

oohaysmlm:
Added a call to AccountsTemplates.options.postSignUpHook in the AccountsServer.updateOrCreateUserFromExternalService function so that postSignUpHook is correctly called even when users sign up using an OAuth service.

This has been a big issue for those of us using OAuth services that had code that relied on the postSignUpHook.  This issue was discussed in the meteor-useraccounts/core repository, but actually had to be fixed in the accounts-base package, which is part of the main Meteor repo.  Discussion on this issue can be found here: https://github.com/meteor-useraccounts/core/issues/669#issuecomment-226339972

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1388,9 +1388,21 @@ Ap.updateOrCreateUserFromExternalService = function (
     // insertUserDoc.
     user = {services: {}};
     user.services[serviceName] = serviceData;
+
+    var userId = this.insertUserDoc(options, user);
+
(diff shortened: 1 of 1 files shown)


tmeasday:
Hey @oohaysmlm, thanks for the PR.

It doesn't make sense for Meteor to call out to a third party package. What is the equivalent code for normal signup that triggers this hook in `accounts-base`? 

I think it would make sense to open a feature request issue where you outline the problem and then we can talk about whether it makes sense to implement it.

Issue #7393, opened 2016-07-12
`--production` flag should add a warning


abernix:
There seems to be a trend of users using the `meteor-tool` to run their production app (using `meteor run`) instead of using `meteor build` and running the bundled `node` app.  This is obviously less-than-ideal as the meteor tool has a lot of overhead which is designed to facilitate development but is not necessary (nay, just wrong) to have in production.

This trend seems to be even more popular since the addition of the `--production` flag, which is purely meant to _simulate_ production minification, but does almost nothing else (in fact, maybe nothing at all), as far as I know.

It might even be worth going a step further and renaming `--production` to `--simulate-production` to avoid the confusion of thinking you can use `meteor run --production` on your live deployment.  Really, anything to discourage this behavior.

@laosb mentioned this recently in #7372 and I've been picking up hints of users doing this on a weekly (or more often) basis.

Thoughts?

(EDIT: Just remembered, the flag also excludes `debugOnly` packages)

laosb:
Renaming is a good idea. How about restrict visits by setting the IP `meteor-tool` listening to 127.0.0.1 and not allow 80/443 port for `--production`?

abernix:
> How about restrict visits by setting the IP meteor-tool listening to to 127.0.0.1

This can't work as Cordova needs to be able to connect over the public interface when testing on mobile devices.  This could also prevent testing with different browsers from other OSes from within VirtualBox.

> not allowing 80/443 port for `--production`

I don't know if not allowing is a good idea.  Even if we want to prevent it, doing this could break existing production environments.

Though, perhaps the warning could show when running on port 80 too.  I think 443 might be unnecessary as there is no way to specify certificates – they would be running into other issues, presumably.  Regardless, the user could be running on any port with an Nginx proxy (a configuration we've just seen suggested in another ticket).

Maybe some other informational warnings could be added to the tool, or just something more clear in the docs.

laosb:
Another idea: How about detecting `X-Forwarded-For` on development?

abernix:
Possibly.  We just need to consider that `X-Forwarded-For` is not only for reverse proxies, but also for traditional proxies (SOCKS, etc.).

💭 But... Might be fine? 💭

I still think we could just start with the bare minimum and add more advanced detection if it's a continuing problem.

laosb:
If MDG have some data to analysis, this would be easier. For example, the number of long-running meteor tool instances on linux?

abernix:
[...]

PR #7422, opened 2016-07-17
Add a warning when use --production.

laosb:
#7393 

May need further discussions on this problem.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -325,6 +325,15 @@ function doRunCommand(options) {
     appPort = appPortMatch[2] ? parseInt(appPortMatch[2]) : null;
   }
 
+  if (options.production) {
+    Console.warn(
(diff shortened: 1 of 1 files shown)


tools/cli/commands.js
```diff
@@ -325,6 +325,15 @@ function doRunCommand(options) {
     appPort = appPortMatch[2] ? parseInt(appPortMatch[2]) : null;
   }
 
+  if (options.production) {
+    Console.warn(
```
zol: This is good but the messaging could do with some grammatical tweaks. I'd suggest 

"Warning: The --production flag should only be used to simulate production bundling for testing purposes. Use `meteor build` to create a bundle for production deployment (see: https://guide.meteor.com/deployment.html)".

laosb: Ah, good point!

Issue #7397, opened 2016-07-13
onLogout should provide connection information


gsonnenf:
The onLogout hook was added with a comment that it was analogous to the onLogin hook:

> Adds Accounts.onLogout() a hook directly analogous to Accounts.onLogin()

https://github.com/meteor/meteor/blob/devel/History.md#v1331

I've tested it and it doesn't seem to provide session context analogous to the onLogin hooks.

The login hook code is as follows - 'packages/accounts-base/accounts_server.js' ->
line: 165
```
Ap._successfulLogin = function (connection, attempt) {
  this._onLoginHook.each(function (callback) {
    callback(cloneAttemptWithConnection(connection, attempt));
    return true;
  });
};
```
line 355:
`this._successfulLogin(methodInvocation.connection, attempt);`

The logout hook code is as follows - 'packages/accounts-base/accounts_server.js' ->
line 179: 
```
Ap._successfulLogout = function () {
  this._onLogoutHook.each(function (callback) {
    callback(); <---- no context set
    return true;
  });
};
```

line 535:
 ```
 methods.logout = function () {
    var token = accounts._getLoginToken(this.connection.id);
    accounts._setLoginToken(this.userId, this.connection, null);
    if (token && this.userId)
      accounts.destroyToken(this.userId, token);
    this.setUserId(null);
    accounts._successfulLogout(); // <-------------no arguments passed
  };
```

It would be useful to pass a connection context and perhaps also return the the userId of the person who logged out as an argument (it is currently cleared before the log out is called).

abernix:
Thanks for this feature request.  I think this is reasonable, and this will likely be a candidate for `pull-requests-encouraged`, however I'll cite the [feature request guidelines](https://github.com/meteor/meteor/blob/devel/Contributing.md#feature-requests) for the sake of completeness and because there are a lot of other feature requests:

> You'll have the best chance of getting a change into core if you can build consensus in the community for it. Start by creating a well specified feature request as a Github issue.

> Help drive discussion and advocate for your feature on the Github ticket (and perhaps the forums). The higher the demand for the feature and the greater the clarity of it's specification will determine the likelihood of a core contributor prioritizing your feature by flagging it with the pull-requests-encouraged label.

Specifically, try to work up a consensus (perhaps on the forums) of what fields would be most beneficial, etc.  Shouldn't be too much to figure out here, but +1's accepted!

Related to #6889 #7187

gsonnenf:
[...]

PR #7433, opened 2016-07-18
[7397] onLogout should provide connection information

Ragsboss:
Fix for https://github.com/meteor/meteor/issues/7397. See commit description for more details. Added a new test and ran all existing tests.
@abernix could you please review this?


packages/accounts-base/accounts_server.js
```diff
@@ -176,9 +176,10 @@ Ap._failedLogin = function (connection, attempt) {
   });
 };
 
-Ap._successfulLogout = function () {
+Ap._successfulLogout = function (connection, userId) {
+  var logoutContext = { userId: userId, connection: connection };
   this._onLogoutHook.each(function (callback) {
-    callback();
+    callback(logoutContext);
```
benjamn: I would just call `callback({ userId, connection })` here, to provide a defensive copy of the context information to each callback.

abernix:
@Ragsboss In addition to the comments above, how do you feel about having the `onLogin` receiving the same parameter (`logoutContext`) for consistency between the two going forward?

packages/accounts-base/accounts_server.js
Ragsboss: Will do.

Ragsboss:
@abernix changing onLogin's current argument would be a breaking change. I'm not sure if this is ok? In fact (as I mentioned in my commit message for first commit), I was wondering if I should instead pass logoutContext as { user, connection } instead of { userId, connection } to make it consistent with existing context (aka attempt in the code) that's passed on onLogin callbacks. For reference the onLogin callback receives an object like this based on my tests.
`{"type":"password","allowed":true,"methodName":"login","methodArguments":[{"user":{"username":"admin@abc.com"},"password":{"digest":"d6b5d69d3c396de3ffb35077d0bbee33b4605eb9cabf251af305716e61","algorithm":"sha-256"}}],"user":{"_id":"QoxkdfReN8RpnfBKD","createdAt":"2016-07-16T00:34:47.179Z","services":{"password":{"bcrypt":"$2a$10$tRPe0c40CI7c/1MfK9a9fePXQYvHGmr8sh88q/ZGYMAAl3ACLBWve"},"resume":{"loginTokens":[]}},"username":"admin@abc.com","profile":{"firstname":"Admin"}},"connection":{"id":"mr2SQXqQyWeoHMuyi","clientAddress":"127.0.0.1","httpHeaders":{"referer":"https://192.168.56.103/","x-forwarded-for":"127.0.0.1","x-real-ip":"192.168.56.1","x-forwarded-proto":"http","host":"192.168.56.103","user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/51.0.2704.103 Safari/537.36","accept-language":"en-US,en;q=0.8"}}}`

abernix:
You're right, we wouldn't want that to be a breaking change.   I was mainly rooting for getting the `connection` and `user` parameters to be the same – which you've done.  I think the resolution of `userId` to `user` to achieve this consistency is certainly worth it.  Once this is merged, meteor/docs#66 should cover the documentation for this (if you want to take a look).

LGTM.

Ragsboss:
@benjamn and @abernix thank you for your review and comments. I'm not sure what the process is now. Who can merge this pull request to meteor/devel?

abernix:
Magical pull-request gnomes will come through and make sure we're all thinking clearly.  Give it some time, but they will come. 🍪

(In seriousness, it will be reviewed by the core team in due time – generally within a week)

packages/accounts-base/accounts_server.js
```diff
@@ -176,9 +176,10 @@ Ap._failedLogin = function (connection, attempt) {
   });
 };
 
-Ap._successfulLogout = function () {
+Ap._successfulLogout = function (connection, userId) {
+  const user = userId && this.users.findOne(userId);
```
mitar: Are we sure we want to always make a database query here? Why wouldn't we just pass `userId` and leave to the developer to fetch the user document if they want? I think this is not necessary to do automatically.

mitar: (Moreover, you are fetching the user object with all fields, which can potentially be large.)

Ragsboss: This is to be consistent with other handlers like onLogin.

mitar: I do not see `user` to be passed to `onLogin`? `connection` is cheap, because it is already available. But `user` object is not.

Ragsboss: onLogin does get an object with user property. See following code. This is also mentioned in the documentation and we indeed use this property in our project. 

  var user;
  if (result.userId)
    user = this.users.findOne(result.userId);

  var attempt = {
    type: result.type || "unknown",
    allowed: !! (result.userId && !result.error),
    methodName: methodName,
    methodArguments: _.toArray(methodArgs)
  };
  if (result.error)
    attempt.error = result.error;
  if (user)
    attempt.user = user;

tmeasday:
Great work @Ragsboss!

Issue #7431, opened 2016-07-18
[windows] Wix installer should call WixBroadcastEnvironmentChange


abernix:
This looks like a potential solution for requiring Windows users to restart after installing Meteor and confusion over the `meteor` command not working after installation as in #7322 & #4090.

See [Wix documentation here](http://wixtoolset.org/documentation/manual/v3/customactions/wixsettingchange.html) which says specifically:

> Windows Installer itself sends the WM_SETTINGCHANGE message for settings it changes while processing an MSI package but cannot do so for changes a package makes via custom action. Also, Windows Installer does not send WM_SETTINGCHANGE for environment variable changes when a reboot is pending.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

This feature request will be closed here, but anyone interested in migrating this feature request to the new repository (to make sure it stays active), can click [here](https://github.com/meteor/meteor-feature-requests/issues/new?title=%5Bwindows%5D%20Wix%20installer%20should%20call%20WixBroadcastEnvironmentChange&body=Migrated%20from:%20meteor/meteor%237431) to start the feature request migration process. This manual migration process is intended to help identify which of the older feature requests are still considered to be of value to the community. Thanks!

Issue #6837, opened 2016-04-18
Windows installer cannot properly reinstall


avalanche1:
1) If I delete `AppData/Local/.meteor` folder and try to reinstall meteor with the official installer - it doesn't download anything, but says that meteor has been installed:
![image](https://cloud.githubusercontent.com/assets/10534563/14598845/e67d133a-055c-11e6-8c48-78381cad8e57.png)
Then I have to manually go through the registry and remove every line containing string 'meteor' in order for the installer to actually download meteor files.
Can you make that the installer actually re-downloads all the files every time it's being run - no matter what is in the registry?
2) After all this voodoo, there is no uninstall record for meteor in windows remove\uninstall programs list. That needs to be fixed

benjamn:
Thanks for letting us know about this problem!

@stubailo may have some ideas about how to proceed here

benjamn:
Just ran into this myself. Although this isn't fixed in the current beta of 1.3.3 (`meteor update --release 1.3.3-beta.0`), it will be fixed before the official 1.3.3 release.

PR #7443, opened 2016-07-20
Trigger an event that allows Windows to notice PATH changes

abernix:
This will hopefully be the solution to meteor/meteor#7431


benjamn:
[Here](http://s3.amazonaws.com/meteor-windows/installers/1.3.5.1/InstallMeteor.exe)'s the (unsigned) installer, in case anyone wants to try it.

abernix:
*sigh* This didn't work, as apparently this was only added [in WiX Toolset 3.10.0.1519](https://github.com/wixtoolset/wix3/blob/wix310rtm/History.md#wixbuild-version-31001519).

Also, turns out Windows 10 is the worst offender here and Meteor only recommends rebooting on Windows 8.1.  I guess if WiX Toolset can't be updated, the next best option would be to change the instructions on the "Success" dialog.

I don't know if updating Wix Toolset is an easy task or not. 

(In related news, I can confirm that #6685 is fixed but #6837 doesn't seem to)

Notes:

#### Windows 10
##### Before
* Required reboot or opening/closing ENV variable control panel.
* Existing Command prompt must be reopened
* Opening new command prompt does not work.

##### After
* Same

#### Windows 8.1
##### Before
* Opening new command prompt works.
* Existing Command prompt must be reopened

##### After
* Same

#### Windows 7
##### Before
* Opening new command prompt works.
* Existing Command prompt must be reopened

##### After
* Same

abernix:
Will revisit this when it'll be more reasonable to update the dependencies for the WiX installer which are required to make this work.

Issue #7026, opened 2016-05-11
Downloads still occur with METEOR_OFFLINE_CATALOG set


thiagodelgado111:
***1 Upvote*** ### Meteor and OS info
Meteor 1.3.2.1
Mac OS X 10.11.4 

### Expected behavior
Using the METEOR_OFFLINE_CATALOG=1 should stop Meteor trying to automatically download and upgrade meteor releases and packages.

### Actual behavior
It continue giving me this message:
```
Meteor 1.3.2.4: Improvements to deploying binary npm packages,
importing stylesheets from node_modules, and more:
https://github.com/meteor/meteor/blob/master/History.md#v1324

This release is being downloaded in the background. Update your app to
Meteor 1.3.2.4 by running 'meteor update'.
```

### How to reproduce
Start a meteor v1.3.2.1 app using `METEOR_OFFLINE_CATALOG=1 meteor`

abernix:
This is tough to say what the "Expected Behavior" should be since this is really an undocumented feature that has been mainly mentioned in comments.  Though, it does behave differently than previously defined in those comments:

@glasser once wrote in #2846:
> As a workaround for now, you can disable catalog refreshing entirely by setting the METEOR_OFFLINE_CATALOG environment variable. This will prevent you from getting any package or *release updates* (including updates that will fix this bug!) so I would avoid setting it permanently for every command.

Anyhow, this is occurring because the [`updateMeteorToolSymlink` calls `downloadPackagesMissingFromMap` right before it shows the banner](https://github.com/meteor/meteor/blob/devel/tools/packaging/updater.js#L64) and nothing checks `catalog.official.offline`.

If keeping keeping this undocumented feature around is important, an easy fix would be to do `if (catalog.official.offline) return;`, either:
* [somewhere around here](https://github.com/meteor/meteor/blob/devel/tools/packaging/updater.js#L183)
* Or, **probably better**, in the actual `downloadPackagesMissingFromMap` function [somewhere around here](https://github.com/meteor/meteor/blob/devel/tools/packaging/tropohouse.js#L575)

Both should probably have a mandatory warning that this variable is enabled.

Now, the banner about downloading is actually entirely deceiving because once you see that, the `meteor-tool` is actually already downloaded.  It shows that message on it's next update check when it realizes it's got it, and after the symlink update.  The banner message itself is just hardcoded to say that and nothing checks the offline-ness.

Not sure how critical this all is, but +1's accepted.

[...]

PR #7445, opened 2016-07-21
Add option to skip release check

c9s:
## Summary

In issue #7026, the release updater is not a part of the catalog update, therefore it still runs the updater even if `METEOR_OFFLINE_CATALOG` is specified.

The reason that we really really need this option is, meteor runs way too slow on travis-ci, it spents almost 7 minutes to just *prepare* the application.

## Changes

A `--no-release-check` option is added to the run command in this PR.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -375,7 +375,7 @@ function doRunCommand(options) {
     oplogUrl: process.env.MONGO_OPLOG_URL,
     mobileServerUrl: utils.formatUrl(parsedMobileServerUrl),
     once: options.once,
-    noReleaseCheck: options['no-release-check'],
+    noReleaseCheck: options['no-release-check'] || process.env.METEOR_NO_RELEASE_CHECK,
     cordovaRunner: cordovaRunner
   });
 }
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -256,6 +256,7 @@ var runCommandOptions = {
     'mobile-port': { type: String },
     'app-port': { type: String },
     'debug-port': { type: String },
+    'no-release-check': { type: Boolean },
     production: { type: Boolean },
     'raw-logs': { type: Boolean },
     settings: { type: String },
@@ -374,6 +375,7 @@ function doRunCommand(options) {
     oplogUrl: process.env.MONGO_OPLOG_URL,
     mobileServerUrl: utils.formatUrl(parsedMobileServerUrl),
     once: options.once,
+    noReleaseCheck: options['no-release-check'],
     cordovaRunner: cordovaRunner
   });
 }
--- a/tools/cli/help.txt
+++ b/tools/cli/help.txt
@@ -58,6 +58,7 @@ Options:
   --release        Specify the release of Meteor to use.
   --verbose        Print all output from builds logs.
   --no-lint        Don't run linters used by the app on every rebuild.
+  --no-release-check            Don't run the release updater to check for new releases.
   --allow-incompatible-update   Allow packages in your project to be upgraded or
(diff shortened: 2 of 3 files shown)


c9s:
I'm open to discuss the API and the option name. Please send me the feedbacks then I can update this PR.

abernix:
This provides a flag but doesn't provide any functionality for that flag.  Did you forget to commit part of your fix?

c9s:
Just one moment,  I will update the changes

Jesse Rosenberger <notifications@github.com> 於 2016年7月21日 星期四寫道：

c9s:
There was one missing commit. I've updated the changes.

tools/cli/commands.js
```diff
@@ -375,7 +375,7 @@ function doRunCommand(options) {
     oplogUrl: process.env.MONGO_OPLOG_URL,
     mobileServerUrl: utils.formatUrl(parsedMobileServerUrl),
     once: options.once,
-    noReleaseCheck: options['no-release-check'],
+    noReleaseCheck: options['no-release-check'] || process.env.METEOR_NO_RELEASE_CHECK,
```
abernix: I'm not sure this deserves of an environment variable.   Why wouldn't this just be a `--flag`?

(I realize that Meteor has a lot of other ENV variables and it's a common way to set things, however I think ENVs should be reserved for production environments.  This particular flag wouldn't have any significance in production).

c9s: I am fine to remove the environment variable.  I added this because sometimes we need to run more meteor commands on the development side, if we have the env var, we don't have to append the option everytime.

nathantreid:
I would very much prefer having the environment variable, as I would have it globally enabled. I love updating to the latest version of Meteor ASAP, but production apps are typically a version or 2 behind for stability, and having to wait for Meteor to update when I need to get work done on an existing app is not good.

abernix:
Ok, got around to testing this – **LGTM**, even if I'm not sure I agree with the ENV variable (btw, I suggested my eventual solution in #7559).

I tested this in what I believe to be the most reasonable way:

* Installed Meteor 1.3.5.1 (using older version intentionally so there's something to upgrade)
* Created a Meteor 1.3.5.1 project
* Checked out `release-1.3.5`
* Rebased this simple PR onto that
* Changed the update runner to check for updates every 30 seconds (instead of 3 hours)
* Added some debug just to make sure things are working right.
* Replaced my official 1.3.5.1 installation with the checked-out version (so it would identify as an actual release, not a checkout)
* Ran my 1.3.5.1 project I created above with the `--no-release-check` argument
* Waited an intentionally-long 10 minutes! ☕ 
* Confirmed that there was no network activity background downloading during that time and that the `meteor-tool` symlink was not updated.  Debug message also confirmed that the updater was not running at all (or even scheduled).
* Reinstalled 1.3.5.1 from install.meteor.com
* Changed the update runner to run every 30 seconds manually
* Confirmed that there WAS network activity after 30 seconds, and when the download was done a prompt to install Meteor 1.4.0.1, plus the symlink was changed.

Obviously if you run any of your projects without `--no-release-check` then the global tool will still be updated, but as per usual the app itself will _only_ be updated if you run `meteor update` within that directory.

abernix:
Alternatively, it sounds like disabling auto-updates for _everyone_ would preferred by some folks: https://github.com/meteor/meteor/issues/7026#issuecomment-235622192 (Though I think Internet Explorer adopted that approach at one point and it failed miserably)

c9s:
Thanks!

DominikGuzei:
Thanks to the 7 gods, new and old … this is the best "feature" lately! 👍

PR #7447, opened 2016-07-21
linker optimization

c9s:
## Summary

The linker split the source code by lines repeatedly (at least twice), since splitting the code into lines is a must, we simply split the lines in the constructor and get the lines from the File object.

This PR also rewrites some `_.each` loop, since symbols and lines (N) could be huge, iterating the array elements with the native javascript for loop is faster.


c9s:
wrong branch, sorry

Issue #6058, opened 2016-01-21
Build tool performance improvements


dgreensp:
**See also this doc:** https://github.com/meteor/meteor/blob/release-1.3/tools/PERFORMANCE.md

This issue tracks build tool performance improvements being worked on in the context of Meteor 1.3.

When talking about build tool performance, the following factors should be taken into account:
- Initial build vs. rebuild: As you'd expect, rebuild takes much less time than initial build, because of caching.  A performance improvement that cuts down initial build time may not affect rebuild time.  Also, there's a distinction between a client-only and a client/server rebuild.  In addition, the _first_ rebuild after starting an app is currently slower than subsequent rebuilds because of how the constraint solver does caching.
- Running from release vs. checkout: Most people run Meteor "from a release," meaning they run the installed `meteor` on their machine (with or without `--release`).  When developing and testing Meteor, we often from "from checkout," meaning we run `./meteor` from the checked-out repo.  These two modes have very different performance characteristics.  When you run from checkout, the core packages are loaded from the repo.  This means that, for example, a lot more files are watched, but the constraint solver doesn't have to do nearly as much work because the core package versions are pinned.  While it's easier to debug and inspect Meteor when running from checkout, when working on performance improvements we should focus on the experience of running from release.
- Size and complexity of app:  One app may be slower to build than another because it is larger or more complex on in any number of dimensions, such as the amount of code to compile or CSS to minify; the compiler plugins used; the size of the dependency graph; or the size in bytes of the packages used.  We should make sure "skeleton" (the empty app) is fast, and also that typical "real apps" are fast.
- Hardware and OS:  Numbers will vary based on your CPU speed and disk speed (especially whether you have an SSD or a spinning disk).  Also, the Windows file system has different characteristics from a Mac or Linux file system.

Even though absolute time measurements will vary from computer to computer and app to app, it's useful to talk in rough numbers about where time is going.
#### Overview of Rebuilding

When a "real app" rebuilds running from release, it may take as much as 10-20 seconds on `1.3-modules-beta.4` (the latest release as of the beginning of this work).  Most of this time goes to "bundling."  The bundler is instrumented using Meteor's profiling library, so if you run `METEOR_PROFILE=1 meteor` in a shell, you'll get a decent overview of where the time is going (under the heading "Rebuild App").

Besides bundling, there are a couple other "hidden" sources of rebuild time which you won't see in the profile, but they are small in comparison:
[...]

PR #7449, opened 2016-07-21
#6058: linker/sourcemap optimization

c9s:
## Summary

1. The linker split the source code by lines repeatedly (at least twice), since splitting the code into lines is a must, we simply split the lines in the constructor and get the lines from the File object.

2. This PR also rewrites some `_.each` loop, since symbols and lines (N) could be huge, iterating the array elements with the native javascript for loop is faster.

3. ES6 spread operator is more expansive if it's translated into es5 by babel.  push.apply reduces a lot of computation costs.

## Fixes

Issue #6058

### Benchmark Result

The benchmark was tested with our real application, the test command is to clean all linker cache and re-run the build:

```
rm -rf .meteor/local/bundler-cache && METEOR_PROFILE=500 ~/work/meteor/meteor run
```

Before
```
(#1) Total: 18,569 ms (ProjectContext prepareProjectForBuild)
(#2) Total: 175,310 ms (Build App)
```

After
```
(#1) Total: 17,786 ms (ProjectContext prepareProjectForBuild)
(#2) Total: 164,798 ms (Build App)
```

https://gist.github.com/c9s/cc5891c0d9050c7b9a8f47ee7c0f072e

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -373,8 +373,9 @@ var buildSymbolTree = function (symbolMap) {
         walk[part] = {};
       walk = walk[part];
     }
-    if (value)
+    if (value) {
       walk[lastPart] = value;
+    }
   }
   return ret;
 };
@@ -644,7 +645,8 @@ _.extend(File.prototype, {
         });
       }
 
-      for (var line = 1; line <= lines.length; ++line) {
+      var lineCount = lines.length;
+      for (var line = 1; line <= lineCount; ++line) {
         addIdentityMapping({ line, column: 0 });
       }
 
@@ -1015,8 +1017,8 @@ export var fullLink = Profile("linker.fullLink", function (inputFiles, {
     noLineNumbers
   });
 
-  var i = 0, inputFile;
-  for ( ; i < inputFiles.length ; i++) {
+  var i = 0, inputFile, _len;
+  for (_len = inputFiles.length ; i < _len ; i++) {
     module.addFile(inputFiles[i]);
   }
 
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -942,7 +942,12 @@ var getFooter = function ({
       chunks.push(pkgInit, ";\n");
     } else {
       const scratch = {};
-      _.each(exported, symbol => scratch[symbol] = symbol);
+      // _.each(exported, symbol => scratch[symbol] = symbol);
+      for (var i = 0 ; i < exported.length ; i++) {
+        var symbol = exported[i];
+        scratch[symbol] = symbol;
+      }
+
       const symbolTree = writeSymbolTree(buildSymbolTree(scratch));
       chunks.push("(function (pkg, symbols) {\n",
                   "  for (var s in symbols)\n",
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -908,14 +908,13 @@ var getImportCode = function (imports, header, omitvar) {
 
   // Imports
   var scratch = {};
-  _.each(imports, function (name, symbol) {
-    scratch[symbol] = packageDot(name) + "." + symbol;
-  });
+  for (var symbol in imports) {
+    scratch[symbol] = packageDot(imports[symbol]) + "." + symbol;
+  }
   var tree = buildSymbolTree(scratch);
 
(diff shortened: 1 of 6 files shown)


c9s:
I am not sure why the tests are failing. If I forked from `master`, these tests will pass.

abernix:
Your #7448 also failed which was based on `master`, but PRs should be based on `devel`.

More importantly though, what issue number are you trying to fix?  Pull requests require discussion and planning before they are considered.

Please read the [contribution guidelines](https://github.com/meteor/meteor/blob/devel/Contributing.md#feature-requests) and consider looking at the `./meteor self-test` command so you can see how the CI tests work by [looking at this script](https://github.com/meteor/meteor/blob/devel/scripts/ci.sh#L1).  You can run tests locally and avoid submitting PRs before they are done.

If you're submitting pull requests to optimize the Meteor tool, make sure you have a way to test and show your speed improvements.

c9s:
yes, this PR is for optimizing the Meteor tool. I will give you the profiling result tomorrow.

abernix:
Please open an issue discussing what you plan to fix and showing your profiling then link this PR to it

c9s:
Sure, I will do that tomorrow.

c9s:
@abernix This issue is related to https://github.com/meteor/meteor/issues/6058

Rebuild -> Sourcemap Performance

c9s:
@abernix I've updated the description to include the benchmark result, please take a look.

tools/isobuild/linker.js
```diff
@@ -1059,6 +1072,26 @@ export var fullLink = Profile("linker.fullLink", function (inputFiles, {
...
+      //   file2.source = header1 + SOURCE_MAP_INSTRUCTIONS_COMMENT + header1 
+      //                    + source + footer
+      //
+      //   file3.source = header1 
+      //                    + SOURCE_MAP_INSTRUCTIONS_COMMENT + header1 
+      //                    + SOURCE_MAP_INSTRUCTIONS_COMMENT + header1
+      //                    + source + footer
+      //
+      // is this expected?
+      //
```
c9s: @abernix When I walked through this function, I have some doubt for the header re-assignment, is this expected?

c9s: I think this should be moved to the outside of the map ?

c9s:
@abernix do I have to regenerate the dev bundle file everytime I switch to another branch? or I just run ./meteor --get-ready ?

laosb:
@c9s In your case you don't really need to generate the devbundle since you didn't modify the things in bundle, I think?

c9s:
@laosb Thanks! yeah I think so, I only modify the source files under tools.

c9s:
Is there a reason that meteor tied to node 0.10 ? I found `--prof-process` is not supported in node 0.10 and it's hard to check the profile result.

c9s:
The building speed is also improved on travis.

It used to take 20 minutes to finish a build, now it only takes 10 minutes!

benjamn:
Thanks for working on this!

Can you try pushing more commits at once? Every time you push, it triggers a separate run of the test suite. I'm also happy to cancel the older test runs, but I don't want to step on your toes.

benjamn:
Canceled everything except the most recent test run.

benjamn:
This pull request really needs to be split up to address individual, specific performance optimizations. #6058 is an umbrella performance issue, not really appropriate as a justification for a pull request.

Please take a step back, think about the different problems you're trying to solve, and open some new issues describing those problems. Then we can have a discussion about which ones should be addressed first, and you can solve each one individually, without making lots of loosely-related changes as you go.

I suspect some of these changes are much more important than others, and I would like to understand why. With everything in one big pull request like this, it's hard to learn any lessons that we can apply when writing code in the future.

I don't want to discourage this kind of work, but it has to be more focused than this. If you're not willing to organize your changes by specific goals, I have to warn you we're not likely to merge them.

Thanks for understanding.

c9s:
@benjamn Thanks, the reason I pushed the commit atomically is that I broke something in specific commit, so I break the changes into several commits to CI to tests.

The reason I'm pushing the branch with changes together is, it runs really really slow to build the application on our side, therefore I maintained a branch for the team.

I will separate the optimization target in the issue, and rebase the changes onto different branch.

PR #7457, opened 2016-07-22
Extract .tar.gz files using native tar or 7z.exe when possible

benjamn:
For the sake of portability, the `files.extractTarGz` method is currently implemented in terms of a pure-JavaScript library: https://www.npmjs.com/package/tar

Portability is nice, but speed is also nice. On my machine, it takes 25 seconds to extract the `fourseven:scss` package using the npm `tar` package, whereas it takes only *3.5 seconds* to extract it using the native `tar` command. On Windows, the time drops from 36 seconds to 8 seconds.

The time savings for even larger packages like `meteor-tool` are even more dramatic: 45 seconds vs. 4 seconds on my Mac.

This pull request uses the native commands when possible, but falls back to the npm `tar` library if the native commands don't work for any reason.

For Windows, we now include a standalone `7z.exe` binary in the dev bundle (i.e., `dev_bundle\bin\7z.exe`), though `tar` will also work on Windows if it's in the `$env:PATH`. This is the standard 7-Zip 16.02 program available [here](http://www.7-zip.org/download.html). I believe this usage is compatible with the 7-zip license, according to the [FAQ](http://www.7-zip.org/faq.html).

I think these changes may be too risky to ship with Meteor 1.4, given that it's now in the release candidate stage, but if all goes well we will ship them soon after that release (likely 1.4.1).

cc @tmeasday @zol @abernix

### DIFF
--- a/tools/fs/files.js
+++ b/tools/fs/files.js
@@ -745,6 +745,11 @@ function ensureDirectoryEmpty(dir) {
 function tryExtractWithNativeTar(buffer, tempDir, options) {
   ensureDirectoryEmpty(tempDir);
 
+  if (options.forceConvert) {
+    return Promise.reject(new Error(
+      "Native tar cannot convert colons in package names"));
+  }
+
   return new Promise((resolve, reject) => {
     const flags = options.verbose ? "-xzvf" : "-xzf";
     const tarProc = spawn("tar", [flags, "-"], {
@@ -766,6 +771,11 @@ function tryExtractWithNativeTar(buffer, tempDir, options) {
 
 function tryExtractWithNative7z(buffer, tempDir, options) {
   ensureDirectoryEmpty(tempDir);
+
+  if (options.forceConvert) {
+    return Promise.reject(new Error(
+      "Native 7z.exe cannot convert colons in package names"));
+  }
 
   const exeOSPath = files.convertToOSPath(
     files.pathJoin(files.getCurrentNodeBinDir(), "7z.exe"));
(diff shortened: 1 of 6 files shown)


tools/fs/files.js
```diff
@@ -695,33 +696,31 @@ files.extractTarGz = function (buffer, destPath, options) {
...
 
-  new Promise(function (resolve, reject) {
-    var gunzip = zlib.createGunzip().on('error', reject);
+  const startTime = +new Date;
+  let promise = tryExtractWithNativeTar(buffer, tempDir, options);
 
-    var extractor = new tar.Extract({
-      path: files.convertToOSPath(tempDir)
-    }).on('entry', function (e) {
-      if (process.platform === "win32" || options.forceConvert) {
```
benjamn: The `options.forceConvert` boolean is no longer passed by any callers.

tools/fs/files.js
```diff
@@ -695,19 +696,154 @@ files.extractTarGz = function (buffer, destPath, options) {
   var tempDir = files.pathJoin(parentDir, '.tmp' + utils.randomToken());
   files.mkdir_p(tempDir);
 
+  if (! _.has(options, "verbose")) {
+    options.verbose = require("../console/console.js").Console.verbose;
```
benjamn: Note that `meteor run --verbose` will cause all the output from the `tar` and/or `7z.exe` commands to be displayed in the console, in case they don't seem to be working and we need to get some debugging clues.

s-devaney:
Could the status of the extraction be added to the update process? At the moment it says "downloading" rather than "extracting" etc

abernix:
@s-devaney This is a good idea.  I've looked at this recently, but currently they're in the same sub-task so the message doesn't change.   I think assuming that this PR works okay, it will be possible to improve the messaging.  For now, as Ben mentioned above, `meteor run --verbose` will be super informative.

benjamn:
@s-devaney I'll add a commit that reflects that, thanks!

benjamn:
The build messages now distinguish between downloading, extracting, and loading packages. Thanks to this PR, the extraction is now by far the cheapest part.

benjamn:
@tmeasday should I hold off on merging this until after 1.4.0.1?

tmeasday:
Yeah, although I can create a release branch for 1.4.0.1, so merge it whenever

PR #7461, opened 2016-07-23
Upgrade node to 4.4.7 LTS

### LINKED ISSUE #7460: Upgrading node to 4.4.7 LTS
Since Node 4.4.7 is long term supported version, is there a plan to upgrade the bundled node in meteor from 0.10 to 4.4.7 ?

c9s:
This fixes #7460 to upgrade the WAY TOO OLD node 0.10 to 4.4.7 LTS. (long term support version)

### DIFF
--- a/scripts/build-dev-bundle-common.sh
+++ b/scripts/build-dev-bundle-common.sh
@@ -5,7 +5,7 @@ set -u
 
 UNAME=$(uname)
 ARCH=$(uname -m)
-NODE_VERSION=0.10.46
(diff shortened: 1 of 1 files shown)


benjamn:
Again, see https://github.com/meteor/meteor/pull/7218. Did you really think upgrading Node would be as simple as changing one line? I wish it was that simple. Really, I do. But it's not.

I'm thrilled that you want to help, but *please* make an effort to understand the problems you're solving before submitting untested pull requests like this one.

PR #7462, opened 2016-07-23
#6058: Add --cache-solution option support

c9s:
Add `--cache-solution` to cache the result from SAT resolver, this saves about 10 seconds when starting up the app.

Related to #6058

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -260,6 +260,9 @@ var runCommandOptions = {
     'raw-logs': { type: Boolean },
     settings: { type: String },
     verbose: { type: Boolean, short: "v" },
+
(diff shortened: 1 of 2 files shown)


benjamn:
Sorry, but this is already implemented by these two commits, which are part of Meteor 1.3.5.1: https://github.com/meteor/meteor/commit/79660c16aacfd45773fe3e77a478657b05c4403b https://github.com/meteor/meteor/commit/69b100d21a33fd7ca58fc2fe2d8997ade4a36af6

Feel free to reopen this pull request if you think it has benefits beyond those commits. Also, for future reference, it's not a good idea to add a new command-line option for something like this. If it's a good idea, and there are no risks associated with it, then there's no point in making it optional.

PR #7481, opened 2016-07-26
Improve the meteor build speed on Travis CI

c9s:
This is a simple quick fix for the travis yml config file.

## Summary

run.sh already runs ./meteor --self-test

so the install section defined in travis config can be remove.


zol:
@tmeasday you've worked with travis recently, is this change safe?

tmeasday:
Yeah, makes sense to me. I'm just re-running the circle script.

PR #7483, opened 2016-07-26
build tool optimization: merge ignoreFiles patterns

### LINKED ISSUE #7482: build tool optimization: merge ignoreFiles pattern
This is related to #6058

Right now you can see the pattern of ignoreFiles are like 10. so the file finder has to check the filename for 10 times in the worst case.

If we have N = 10,000 which is the number of files, then we need to match 10,000 \* 10 times, and this causes the matching really slow.

The benefit of merging the patterns is, V8's regexp engine will optimize the pattern, so it will be much faster

c9s:
Fix #7482

### DIFF
--- a/tools/tests/bundle-ignore-files.js
+++ b/tools/tests/bundle-ignore-files.js
@@ -1,0 +1,20 @@
+var selftest = require('../tool-testing/selftest.js');
+var Sandbox = selftest.Sandbox;
+var files = require('../fs/files.js');
+var bundler = require('../isobuild/bundle.js');
+
+selftest.define("bundle-ignore-files", () => {
+  var patterns = bundler.ignoreFiles;
+  var inputs = [
+    '.git/refs',
+    '.git/HEAD',
+    '.meteor/local',
+    '.meteor/packages',
+  ];
+  _.each(inputs, (input) => {
+    let matched = _.any(patterns, (ptn) => {
+      return ptn.test(input);
+    });
+    selftest.expectEqual(matched, true);
+  });
+});
--- a/tools/tests/bundle-ignore-files.js
+++ b/tools/tests/bundle-ignore-files.js
@@ -7,10 +7,11 @@ var bundler = require('../isobuild/bundle.js');
 selftest.define("bundle-ignore-files", () => {
   var patterns = bundler.ignoreFiles;
   var inputs = [
-    '.git/refs',
-    '.git/HEAD',
-    '.meteor/local',
-    '.meteor/packages',
+    '.git/',
+    '.meteor/',
+    '.DS_Store',
+    '.aaabbb.swp',
+    'Thumbs.db',
   ];
   _.each(inputs, (input) => {
     let matched = _.any(patterns, (ptn) => {
--- a/tools/tests/bundle-ignore-files.js
+++ b/tools/tests/bundle-ignore-files.js
@@ -1,3 +1,4 @@
+var _ = require('underscore');
 var selftest = require('../tool-testing/selftest.js');
 var Sandbox = selftest.Sandbox;
 var files = require('../fs/files.js');
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -173,11 +173,13 @@ import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
-    /~$/, /^\.#/, /^#.*#$/,  // emacs swap files
-    /^\..*\.sw.$/,  // vim swap files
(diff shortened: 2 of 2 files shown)


zol:
Can you please write a test for this to ensure the semantics are still correct.

c9s:
Sure

Zoltan Olah <notifications@github.com> 於 2016年7月27日 星期三寫道：

tmeasday:
@c9s, thanks again for this PR. We're still waiting to hear about the comments above--just to let you know, we'll be closing this PR in a week if we don't.

c9s:
I ended up with the test last time because I didn't find out a way to just
run one test case, the whole process is just too time consuming - it takes
10minutes for each iteration

I will take a look at this again this week.

Do you have a simple test case sample for this?

Tom Coleman <notifications@github.com> 於 2016年8月17日 星期三寫道：

c9s:
Let's see if the test works

tools/tests/bundle-ignore-files.js
```diff
@@ -0,0 +1,22 @@
+var _ = require('underscore');
+var selftest = require('../tool-testing/selftest.js');
+var Sandbox = selftest.Sandbox;
+var files = require('../fs/files.js');
+var bundler = require('../isobuild/bundle.js');
```
tmeasday: Should this be `../isobuild/bundle.js`?

You can run this test locally with `./meteor self-test "bundle-ignore-files"`

abernix:
@c9s Any updates on this?  You can isolate and run the specific test (or any specific test) using [the suggestion above](https://github.com/meteor/meteor/pull/7483#discussion_r76891916).

sebakerckhof:
Btw, here's a snippet to test the perf improvements of this change in case there are 10000 files:
```js
const nbFiles = 10000;

const ignoreFiles = [
/~$/, /^\.#/, /^#.*#$/,
     /^\..*\.sw.$/, 
     /^\.DS_Store\/?$/, /^ehthumbs\.db$/, /^Icon.$/, /^Thumbs\.db$/,
     /^\.meteor\/$/,
     /^\.git\/$/ 
     ];

const ignoreFilesCombined = [
    /~$/, /^\.#/,
    /^(\.meteor\/|\.git\/|Thumbs\.db|\.DS_Store\/?|Icon.|ehthumbs\.db|\..*\.sw.|#.*#)$/
];

const re = /[^a-z0-9]+/gi;
const testSet = Array(nbFiles).fill(0).map(i => Math.random().toString(36).replace(re, ''));

function test(name, testSet, ignoreFiles){
	const t0 = performance.now();
	testSet.forEach(file => {
               //https://github.com/meteor/meteor/blob/devel/tools/fs/watch.js#L303
		ignoreFiles.some(re => re.test(file))
	});
	const t1 = performance.now();
	console.log(`${name} took: ${(t1-t0)} msecs`);
}

test('old',testSet,ignoreFiles);
test('new',testSet,ignoreFilesCombined);
```

abernix:
@sebakerckhof Thanks for that benchmark code!  Looks like about a ~50% increase when run from `node` (and even more in Chrome).

On that note, it seems that @c9s doesn't currently have time for this, so if you were willing to re-package this, finish the tests and submit a new PR, I'm confident it would be accepted swiftly.

sebakerckhof:
@tmeasday thanks, I rebased on the current devel and ran:
`git submodule update --init --recursive`
and
`./meteor --get-ready`

But now as soon as a test has something like `selftest.expectEqual(matched, true);`, I get:

```
Errors prevented isopacket load:              
                                              
While loading isopacket `ejson`:
error: Can't load npm module 'meteor-deque' from packages/meteor.js. Check your
Npm.depends().

  ... fail!                                   
  => Test threw exception: Error: isopacket load failed?
    at loadIsopacketFromDisk (/tools/tool-env/isopackets.js:299:11)
    at /tools/tool-env/isopackets.js:96:23
    ...
```

Anyway, for now I just did `./meteor npm install -g meteor-deque` to workaround this and then the test passes (after modifications). You want me to open a PR?

abernix:
Yeah, I haven't seen this either.  I'm curious if something is went wrong in your `--get-ready` step?  Maybe try:

* Committing/Stashing any unsaved changes you might have (i.e. `git stash save`)
* Run `git clean -dfx` to return your `meteor` checkout to pristine shape and remove untracked files (e.g. builds, caches, etc.) .

 _(You can add `-n` to enable "dry-run" mode if you're concerned about what it's going to delete)_
* Just run `./meteor --help` (Not `./meteor --get-ready`).

This should download a bundle and avoid a lot of compilation.

sebakerckhof:
@abernix that seems to work. PR submitted.

abernix:
Great!  Closing this issue since the active PR is that one now!

@c9s Thanks for taking the initial initiative on this - much appreciated.  Hopefully it will surface soon!

Issue #7484, opened 2016-07-26
source map header bug?


c9s:
Please take a look
https://github.com/meteor/meteor/blob/devel/tools/isobuild/linker.js#L1063

the option `includeSourceMapInstructions` is an option which maybe true, the `header` variable is defined and initialized outside of the map closure.

The line 1063 concats `header` and re-assigns the result to `header` variable again. if the closure is executed more than 1 time, we will have repeated `SOURCE_MAP_INSTRUCTIONS_COMMENT` and the repeated header and the header will be getting longer. Is it a bug?

I think the concat suppose to be outside of the closure instead of inside the map closure?

I suspect this is the cause of the source map parsing error.

abernix:
I'm not confirming as a bug, but just marking for the time being. 😁

What you're saying sounds reasonable to me, but someone more familiar with that part of the process would need to comment.

abernix:
Closing as this was fixed in Meteor 1.4.1.1 via aacfbbf5d27631d1eb18eb210e6868f105e467a8.  Great find and thanks for reporting this.  Hopefully this will resolve some of the more weird source map problems that have been occurring!

PR #7485, opened 2016-07-26
Fix source map header for the prelinked files

c9s:
I've described the issue here #7484, please take a look, see if this fix makes sense.

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -1057,27 +1057,28 @@ export var fullLink = Profile("linker.fullLink", function (inputFiles, {
     name
   });
 
+  if (includeSourceMapInstructions) {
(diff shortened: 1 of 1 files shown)


tools/isobuild/linker.js
```diff
@@ -1044,6 +1044,10 @@ export var fullLink = Profile("linker.fullLink", function (inputFiles, {
     packageVariables: _.union(assignedVariables, declaredExports)
   });
 
+  if (includeSourceMapInstructions) {
+    header = SOURCE_MAP_INSTRUCTIONS_COMMENT + "\n\n" + header;
+  }
```
benjamn: Can you move this block down closer to the other lines you hoisted out of the `_.map`? In other words, just after the `var footer = getFooter({...});` call.

benjamn:
This looks good to me, with the one comment I made.

c9s:
@benjamn updated and rebased.

PR #7489, opened 2016-07-26
Reorganize and categorize DDP messages

jazeee:
Categorize the DDP messages to separate subscription from collection messages.
The prior DDP documentation intersperses the two types of messages, and this is a simple reorganization of the bullet points to help with understanding the message types.
No real content change.

### DIFF
--- a/packages/ddp/DDP.md
+++ b/packages/ddp/DDP.md
@@ -99,17 +99,22 @@ a `pong` message. If the received `ping` message includes an `id` field, the
 ## Managing Data:
 
 ### Messages:
-
+#### Subscription
  * `sub` (client -> server):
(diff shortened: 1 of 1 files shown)


zol:
Hey, Thanks for your contribution however I'm going to leave things as they are as this spec has been stable and unchanged for a long time now and we haven't heard from others in the community that it's confusing.

jazeee:
Ok, fair enough.
I also noticed a couple missing DDP commands and fields that I observe from Meteor 1.3.5.1
Specifically, Meteor sends `closed` command when closing the connection.
Also, the error objects have more parameters than are documented.
I created a PR for that #7505

Issue #7412, opened 2016-07-15
Cannot test due to HTML files in node_modules


nicocrm:
`meteor test` is giving an error when certain NPM packages are included, that have HTML files that are not parsed correctly by meteor.  One such package is [jsforce](https://github.com/jsforce/jsforce) - an HTML file in the test directory has a `DOCTYPE`.

To reproduce it on meteor (1.3.4.4):
1. meteor create jsforce-test
2. cd jsforce-test
3. meteor npm install --save jsforce
4. meteor test --driver-package practicalmeteor:mocha --port 3100

The error given is:

   While processing files with templating (for target web.browser):
   node_modules/jsforce/test/browser/client.test.html:1: Can't set DOCTYPE here.  (Meteor sets <!DOCTYPE html> for you)

Removing the offending file allows tests to run.

Is it possible to ignore HTML files inside of node_modules?

abernix:
Thanks for the easy reproduction steps.  I can confirm this happens to me as well.  I've packaged this into this reproduction:

https://github.com/abernix/meteor-issue-7412

joncursi:
Same issue here when using `nyc` and `istanbul` to generate HTML files into a `coverage/` folder. Meteor tries to load this HTML and I don't think I have the tools to be able to block that.

hwillson:
This is no longer an issue with Meteor 1.4.2.3 (the supplied repro doesn't throw the error after updating to 1.4.2.3). Closing - thanks for reporting this originally!

Issue #7494, opened 2016-07-27
`meteor test` picks up matching files from `node_modules`


tmeasday:
Reported: https://forums.meteor.com/t/a-first-implementation-of-meteor-app-testing-for-1-3/17097/75

Reproduction: https://github.com/Gywem/meteor-test-issue

tmeasday:
https://github.com/meteor/meteor/pull/7496

tmeasday:
Ah, this is a dupe of https://github.com/meteor/meteor/issues/7412

PR #7496, opened 2016-07-27
More subtle logic for deciding when test files should be lazily loaded

tmeasday:
For https://github.com/meteor/meteor/issues/7494


tmeasday:
Quick sanity check on this on @benjamn

abernix:
I believe this also will Close #7412

benjamn:
lgtm

PR #7505, opened 2016-07-27
Add missing DDP command and error fields

jazeee:
I noticed a couple missing DDP commands and fields that I observe from Meteor 1.3.5.1
Specifically, Meteor sends `closed` command when closing the connection.
Also, the error objects have more parameters than are documented.

Also, I did a small reorganization of the `Managing Data` section that I really believe improves understanding. However, I can revert that if you really don't want it. (See #7489)
> Categorize the DDP messages to separate subscription from collection messages.
> The prior DDP documentation intersperses the two types of messages, and this is a simple reorganization of the bullet points to help with understanding the message types.

It helped me when refactoring the Java DDP client, which has some issues.


jazeee:
It looks like I need to make some modifications. Please hold on this PR.

jazeee:
I need to process the Java DDP client code further to correct this documentation.

Issue #7491, opened 2016-07-26
Very high CPU usage on 1.3.5.1 when running `meteor`


scharf:
On my mac, when I start `meteor`, node takes 175% CPU and my laptop heats up. The same happens for a coworker too. 

We came to the conclusion, that it may have to do with automatic download of some meteor tools.

_Update:_

The issue appears to be that running Meteor projects that use a version of Meteor less than 1.4 with the 1.4 Meteor tool results in extremely high CPU use. This impacts most users as the 1.4 tool is automatically downloaded and run now that it's been released (regardless of whether users are ready to update to 1.4).

--- 

NOTE: There is a workaround for this issue, as outlined by @abernix : https://github.com/meteor/meteor/issues/7491#issuecomment-235655623

SECOND NOTE: this is fixed by 1.4.0.1. You just need to ensure you have that version on your machine. The download should happen automatically, but if not (check by running `meteor --version` **OUTSIDE** of an app), grab it with `meteor update` outside of an app.

scharf:
After I upgrade to meteor `1.4` the problem is gone. But I should be able to continue working with `1.3.5.1`..

benjamn:
And you can continue working with 1.3.5.1! Any app whose `.meteor/release` file has `METEOR@1.3.5.1` in it will automatically use the older version of the tool. This is a special feature of Meteor that you won't find in many (any?) other framework.

scharf:
...sure I can use `1.3.5.1`, but meteor did something so that the meteor watch process kills my laptop (it's quite hot in my office and having a a high load all the time made my laptop stop working because of overheating).

But `1.4` seems pretty stable already (we usually wait for the `x.x.1` release ;-)

billyvg:
We are also having the same issues using release 1.3.5.1 - meteor dev server runs at 100% CPU with this release, but it seems to be fine at 1.4.

Tried `reset` and reinstall meteor to no avail.

marcuscarr:
I'm running into this problem, too, on 1.2.1.

@scharf @billyvg - Are you using CollectionFS? Kadira showed a lot of Mongo activity, and `db.system.profile` shows a lot of queries to the `cfs.<myFileCollection>.filerecords` collections.

scharf:
I tried another (much simpler) app that connects to the same database -- there I don't see the problem.

marcuscarr:
I commented out all the CFS code and Node is still running like crazy. Guess that was a red herring.

scharf:
[...]

PR #7510, opened 2016-07-28
Pipe stdio rather than inheriting

tmeasday:
For https://github.com/meteor/meteor/issues/7491 it seems using `kexec` or `spawn` with `{stdio: 'inherit'}` caused a high constant CPU load on Node 4. Using `{stdio: 'pipe'}` and manually piping seems fine though.

I haven't figured out what the explanation for it is and why it's not a known node issue (I'm going to investigate further).

One positive of this change is unifying some windows/unix code.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,7 @@
 ## v.NEXT
 
+* Fix an issue where springboarding to older releases caused CPU load [Issue #7491](https://github.com/meteor/meteor/issues/7491)
+
 ## v1.4
 
 * Node has been upgraded to 4.4.7.
--- a/packages/meteor-tool/package.js
+++ b/packages/meteor-tool/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "The Meteor command-line tool",
-  version: '1.4.0-1-rc.0'
+  version: '1.4.0-1-rc.1'
 });
 
 Package.includeTool();
--- a/packages/meteor-tool/package.js
+++ b/packages/meteor-tool/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "The Meteor command-line tool",
-  version: '1.4.0'
+  version: '1.4.0-1-rc.0'
 });
 
 Package.includeTool();
--- a/scripts/admin/meteor-release-experimental.json
+++ b/scripts/admin/meteor-release-experimental.json
@@ -1,6 +1,6 @@
 {
  "track": "METEOR",
- "version": "1.4.0.1-rc.0",
+ "version": "1.4.0.1-rc.1",
  "recommended": false,
  "official": false,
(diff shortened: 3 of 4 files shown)


tmeasday:
TODO: can probably remove `kexec` from the dev bundle.

tmeasday:
Ok, this approach is not going to work, due to our console code not flushing buffers (I think). In any case we don't see buildmessages.

@benjamn suggested an approach where the `springboard` function just exits with a code and the wrapper script picks that up and re-runs a different Meteor's `tools/index.js`. 

A couple of hiccups with that approach:

 1. It would mean the shell script cannot use `exec`
 2. We'd need to sometimes run the second `tools/index.js` with `METEOR_SPRINGBOARD_RELEASE=$new-release`, which might mean we need two exit codes
 3. The JS file would have to communicate both that `$new-release` and the `SCRIPT_DIR` for the new run. I'm not sure of the best way to do that (it might be OK to just literally log it to two lines of stdout, but it feels a little unreliable).

I'll probably give that a try tomorrow, unless the other direction (see the issue) bears fruit.

benjamn:
I think we can unify Unix and Windows, and avoid having to communicate back to the shell script (which would necessarily work differently for Unix and Windows).

Here's the idea: have a tiny grandparent process, implemented in Node, that spawns the process that is currently spawned by the script. It listens for exit codes, and also intercepts each line of stdout, so that the Meteor process can communicate things like the springboarded `execPath` and script directory up to the grandparent process.

When the Meteor process dies with the chosen non-zero code, the grandparent process will parse the final lines of stdout and restart the Meteor process appropriately. Since the grandparent process won't have any dangling resources that need to be freed, it's fine for it to outlive the Meteor process, and because it's all JS, this approach should work the same on Windows.

abernix:
`child_process`.`fork` supports creating an IPC which might be a really clean way to communicate back and forth and avoid trying to rely on parsing output (even if that may be possible via other means).

abernix:
(I say that mainly because I read at least 3 issues today around the webz today about how inefficient stdin/stdout buffer passing can be sometimes, misc bugs, etc... [this was the last one I saw](https://github.com/nodejs/node/issues/3429))

https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options

tmeasday:
[Given the issue will be fixed in the next node release](https://github.com/meteor/meteor/issues/7491#issuecomment-236006548) (I confirmed this here), I'm not sure a complicated workaround is warranted. 

I'm going to follow @abernix's hacky `script` direction to see if I can get an easy short term fix.

tmeasday:
https://github.com/meteor/meteor/compare/7491-no-kexec-2?expand=1

PR #7520, opened 2016-07-29
Fix #7491 by wrapping child process in `script` on OSX

### LINKED ISSUE #7491: Very high CPU usage on 1.3.5.1 when running `meteor`

tmeasday:
See https://github.com/meteor/meteor/issues/7491#issuecomment-235887764

NOTE: this issue should be fixed in the next release of node v4
(4.4.8 or 4.5). When we include that in the dev bundle we should
revert this change

### DIFF
--- a/packages/meteor-tool/package.js
+++ b/packages/meteor-tool/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "The Meteor command-line tool",
-  version: '1.4.0'
+  version: '1.4.0-1-rc.2'
 });
 
 Package.includeTool();
--- a/scripts/admin/meteor-release-experimental.json
+++ b/scripts/admin/meteor-release-experimental.json
@@ -1,6 +1,6 @@
 {
  "track": "METEOR",
- "version": "1.4-rc.2",
+ "version": "1.4.0.1-rc.2",
  "recommended": false,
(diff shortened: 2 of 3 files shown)


tmeasday:
Released as 1.4.0.1-rc.2, can confirm this appears to work (including passing build messages).

tmeasday:
Closed in favor of https://github.com/meteor/meteor/pull/7522

PR #7534, opened 2016-07-31
Expire Password reset tokens

Ben305:
This PR is based on #6833 

I have added some tests for the new functionality and increased the default password expiration time to 3 days.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1115,6 +1115,41 @@ Ap._expireTokens = function (oldestValidDate, userId) {
   // expired tokens.
 };
 
+// Deletes expired password reset tokens from the database.
+//
+// Exported for tests. Also, the arguments are only used by
+// tests. oldestValidDate is simulate expiring tokens without waiting
+// for them to actually expire. userId is used by tests to only expire
+// tokens for the test user.
+Ap._expirePasswordResetTokens = function (oldestValidDate, userId) {
+  var tokenLiftetimeMs = this. _getPasswordResetTokenLifetimeMs();
+
+  // when calling from a test with extra arguments, you must specify both!
+  if ((oldestValidDate && !userId) || (!oldestValidDate && userId)) {
+    throw new Error("Bad test. Must specify both oldestValidDate and userId.");
+  }
+
+  oldestValidDate = oldestValidDate ||
+    (new Date(new Date() - tokenLifetimeMs));
+  var userFilter = userId ? {_id: userId} : {};
+
+  this.users.update(_.extend(userFilter, {
+    $or: [
+      { "services.password.reset.when": { $lt: oldestValidDate } },
+      { "services.password.reset.when": { $lt: +oldestValidDate } }
+    ]
+  }), {
+    $pull: {
+      "services.password.reset": {
+        $or: [
+          { when: { $lt: oldestValidDate } },
+          { when: { $lt: +oldestValidDate } }
+        ]
+      }
+    }
+  }, { multi: true });
+}
+
 // @override from accounts_common.js
 Ap.config = function (options) {
   // Call the overridden implementation of the method.
@@ -1135,6 +1170,7 @@ Ap.config = function (options) {
 function setExpireTokensInterval(accounts) {
   accounts.expireTokenInterval = Meteor.setInterval(function () {
     accounts._expireTokens();
+    accounts._expirePasswordResetTokens();
   }, EXPIRE_TOKENS_INTERVAL_MS);
 }
 
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -664,6 +664,11 @@ Meteor.methods({resetPassword: function (token, newPassword) {
         "services.password.reset.token": token});
       if (!user)
         throw new Meteor.Error(403, "Token expired");
+      var when = user.services.password.reset.when;
+      var tokenLifetimeMs = Accounts._getPasswordResetTokenLifetimeMs();
+      var currentTimeMs = Date.now();
+      if ((currentTimeMs - when) > tokenLifetimeMs)
+        throw new Meteor.Error(403, "Token expired");
       var email = user.services.password.reset.email;
       if (!_.include(_.pluck(user.emails || [], 'address'), email))
         return {
--- a/packages/accounts-password/password_tests.js
+++ b/packages/accounts-password/password_tests.js
@@ -1392,6 +1392,75 @@ if (Meteor.isServer) (function () {
       }, /Incorrect password/);
     });
 
+  Tinytest.add(
+    'passwords - reset password should work when token is not expired',
+    function (test, onComplete) {
+      var username = Random.id();
+      var email = username + '-intercept@example.com';
+
+      var userId = Accounts.createUser({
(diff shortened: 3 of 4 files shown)


Ben305:
@tmeasday What do you think?

packages/accounts-password/password_server.js
```diff
@@ -664,6 +664,11 @@ Meteor.methods({resetPassword: function (token, newPassword) {
         "services.password.reset.token": token});
       if (!user)
         throw new Meteor.Error(403, "Token expired");
+      var when = user.services.password.reset.when;
+      var tokenLifetimeMs = Accounts._getPasswordResetTokenLifetimeMs();
+      var currentTimeMs = Date.now();
+      if((currentTimeMs - when) > tokenLifetimeMs)
```
tmeasday: Formatting (add a space between `if` and `(`)

tmeasday:
This looks great @Ben305!

One minor issue with formatting and it's good to go!

Ben305:
@tmeasday Thanks, I have fixed the formatting now

packages/accounts-password/password_tests.js
```diff
@@ -1392,6 +1392,75 @@ if (Meteor.isServer) (function () {
...
+      var resetPasswordToken = match[1];
+
+      makeTestConnection(
+        test,
+        function () {
+          
+          test.isTrue(Meteor.call("resetPassword", resetPasswordToken, "new-password"));
+          test.isTrue(Meteor.call("login", {user: {username: username}, password: "new-password"}));
+
+          onComplete();
```
benjamn: This test completes prematurely, without calling `onComplete()`, since you used `Tinytest.add` instead of `Tinytest.addAsync`. Unfortunately, that's not the only problem: `makeTestConnection` doesn't modify the default connection used by `Meteor.call`, so these method calls don't work as written.

Ben305: Thanks for fixing and explaining this!

tmeasday: Are we sure the surrounding tests (which call `test.throws`) are working without being async?

packages/accounts-base/accounts_server.js
```diff
@@ -1115,6 +1115,41 @@ Ap._expireTokens = function (oldestValidDate, userId) {
...
 };
 
+// Deletes expired password reset tokens from the database.
+//
+// Exported for tests. Also, the arguments are only used by
+// tests. oldestValidDate is simulate expiring tokens without waiting
+// for them to actually expire. userId is used by tests to only expire
+// tokens for the test user.
+Ap._expirePasswordResetTokens = function (oldestValidDate, userId) {
+  var tokenLiftetimeMs = this. _getPasswordResetTokenLifetimeMs();
```
menelike: This does not look like it should work, `this. _getPasswordResetTokenLifetimeMs();` should be `this._getPasswordResetTokenLifetimeMs();`. Imho the extra space needs to be removed.

benjamn: Also `tokenLiftetimeMs`…

packages/accounts-base/accounts_server.js
```diff
@@ -1115,6 +1115,41 @@ Ap._expireTokens = function (oldestValidDate, userId) {
...
+      "services.password.reset": {
+        $or: [
+          { when: { $lt: oldestValidDate } },
+          { when: { $lt: +oldestValidDate } }
+        ]
+      }
+    }
+  }, { multi: true });
+}
+
```
mjmasn: I think there's an issue here if 'services.password.reset' isn't an array. Not sure why because it looks like it should only update if 'services.password.reset.when' exists. I'm running Meteor 1.4.0.1 but have updated accounts-base to 1.2.10 (maybe this is the issue?) Anyway I'm seeing a few errors in production:

```
MongoError: Cannot apply $pull to a non-array value ...[snip]... at AccountsServer.Ap._expirePasswordResetTokens
```

tmeasday:
@mjmasn the issue is fixed on current devel. @benjamn I suppose we should release a fix to `accounts-base` / `accounts-password`?

benjamn:
Already done!

tmeasday:
@benjamn I mean the fix that I just made 8 minutes ago ;)

mjmasn:
@tmeasday sweet, thanks 👍

olizilla:
I'm hitting
```
MongoError: Cannot apply $pull to a non-array value ...[snip]... at AccountsServer.Ap._expirePasswordResetTokens
```
in production as well, after upgrading. Can we get another release of `accounts-base` with @tmeasday's last fix in pls @benjamn ?

lynchem:
Any idea what release this will go in ?

tmeasday:
It'll out in 1.4.1, to be released very soon (and will include a version you can update to independenly @olizilla)

mitar:
Ah ah ah. This broke my invitation system. :-( I agree that forget password tokens should probably have short invitation time. But invitation to the system tokens should probably have much longer one. At least in my case I expected before upgrading to new Meteor version that they will simply not expire at all.

Is there a way to add configuration to invite/enrollment tokens and that those would have a different expiration time?

tmeasday:
@mitar - `Accounts.config({passwordResetTokenExpirationInDays: ... })`

mitar:
This is for password reset, not for enrollment. I would like that password reset is 3 days expiration, while enrollment is 30 days or more. :-)

tmeasday:
Yeah, the code is the same right now. PR?

mitar:
But currently there is also only one method on the server side to handle both of those cases?

tmeasday:
Well there are two methods but they use the same expiration logic.

tmeasday:
(https://github.com/meteor/meteor/blob/devel/packages/accounts-password/password_server.js#L595)

mitar:
When they get back from the invitation workflow, the [same method is called](https://github.com/meteor/meteor/blob/devel/packages/accounts-password/password_server.js#L652) so at that point it is not possible to know how long the token should be for invitation.

So one way to address is to store in the database when it should expire, and not when it was added. Because the function you linked store as `when` current timestamp.

alanning:
As @mitar mentioned, this broke our enrollment system.  :-(   

Incredibly hard for us to debug this as only symptom was a blank "services.password" field until noticed the "no one complains until 3-days after invite" pattern in support tickets.

Would MDG be open to having this mentioned in the Breaking changes section of the [migration doc](https://guide.meteor.com/1.4-migration.html)?

Also, +1 on splitting password reset and enrollment timeouts.

tmeasday:
@alanning sure, if you'd like to send a PR to the guide, maybe mentioning it was released in 1.4.1.

mitar:
@tmeasday: And what about splitting this configuration?

tmeasday:
Oh, yeah definitely consider that `pull-requests-heavily-encouraged`

mitar:
But how it should be done. There is no way to do it without changing the database to store when to expire instead of time when invitation was sent. Or we would have to add extra flag to say the reason for the token (this could also be useful in general, to have in database why is token there, reset or invite).

tmeasday:
I would suggest the second, but maybe we can open an issue to discuss the details?

mitar:
Done: #7794.

PR #7539, opened 2016-08-01
Switch to concat stream

tmeasday:
Use `concat-stream` instead of `stream-buffers`

`concat-stream` is efficient and uses `Buffer.concat` once, rather than a continuous re-allocation and strcpy into new buffers.

### DIFF
--- a/tools/tests/utils-tests.js
+++ b/tools/tests/utils-tests.js
@@ -177,6 +177,7 @@ selftest.define("resume downloads", ['net'], function () {
             ++interruptCount;
             request.emit('error', 'pretend-http-error');
             request.emit('end');
+            request.abort();
           }
         });
       }
--- a/tools/utils/http-helpers.js
+++ b/tools/utils/http-helpers.js
@@ -14,7 +14,7 @@ var release = require('../packaging/release.js');
 var Console = require('../console/console.js').Console;
 var timeoutScaleFactor = require('./utils.js').timeoutScaleFactor;
 
-import { WritableStreamBuffer } from 'stream-buffers';
+import concatStream from 'concat-stream';
 
 // Helper that tracks bytes written to a writable
 var WritableWithProgress = function (writable, listener) {
@@ -393,8 +393,6 @@ _.extend(exports, {
       url: urlOrOptions,
     };
 
-    const outputStream = new WritableStreamBuffer();
-
     const maxAttempts =
       _.has(options, "maxAttempts")
       ? options.maxAttempts : 10;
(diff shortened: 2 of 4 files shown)


tools/tests/utils-tests.js
```diff
@@ -177,6 +177,7 @@ selftest.define("resume downloads", ['net'], function () {
             ++interruptCount;
             request.emit('error', 'pretend-http-error');
             request.emit('end');
+            request.abort();
```
tmeasday: I'm not sure why we didn't need this before, but I was seeing the request continuing to dump data into the stream.

tools/utils/http-helpers.js
```diff
@@ -14,7 +14,7 @@ var release = require('../packaging/release.js');
 var Console = require('../console/console.js').Console;
 var timeoutScaleFactor = require('./utils.js').timeoutScaleFactor;
 
-import { WritableStreamBuffer } from 'stream-buffers';
+import concatStream from 'concat-stream';
```
benjamn: The default export is the `ConcatStream` constructor, so let's capitalize the first `C` and use `new` below.

tools/utils/http-helpers.js
```diff
@@ -404,13 +402,13 @@ _.extend(exports, {
       ? options.retryDelaySecs : 5;
 
     const masterProgress = options.progress;
+    const outputStream = concatStream();
```
benjamn: Let's use `new ConcatStream({ encoding: "buffer" })` here.

tools/utils/http-helpers.js
```diff
@@ -428,10 +426,9 @@ _.extend(exports, {
         }));
 
       } catch (e) {
-        const size = outputStream.size();
-        const useTry = size === lastSize;
-        const change = size - lastSize;
-        lastSize = outputStream.size();
+        const size = _.reduce(outputStream.body, (s, buf) => s + buf.length, 0);
```
benjamn: We can do a little better than recomputing the total length by keeping a running `sum` variable in the same scope as `outputStream` and only adding the lengths of any new `outputStream.body` elements here. To determine which chunks are new, we can store the previous length of `outputStream.body`.

benjamn:
This is definitely an improvement, but I'm not sure we really need the `concat-stream` dependency. If you look at the `ConcatStream` implementation, the parts we actually need boil down to something as simple as
```js
var Writable = require('stream').Writable;

class ResponseStream extends Writable {
  constructor() {
    super();
    this.body = [];
    this.length = 0;
  }

  _write(chunk, encoding, next) {
    this.body.push(chunk);
    this.length += chunk.length;
    next();
  }

  getBody() {
    return Buffer.concat(this.body);
  }
}
```

tmeasday:
Merged your changes, shall we merge this to `devel`?

PR #7550, opened 2016-08-02
Expose the function that gets fb data from an access token

gsabran:
It's possible to get an access token directly, for instance with mobile application. To make it easier to handle those access tokens on the backend, I split the function that handles the token from the larger function that starts with the code.

This enables things like handling fb login directly from an access token:

```js
let fbAppId = null;
export const loginWithFacebook = (params, req, res) => {
  const fbAccessToken = req.body.token;

  // make sure the access token is for our app
  const fbAppIdForToken = JSON.parse(HTTP.get(
    "https://graph.facebook.com/app", {
      params: { access_token: fbAccessToken }
    }).content).id;
  // get our app id
  if (!fbAppId) {
    const config = ServiceConfiguration.configurations.findOne({service: 'facebook'});
    fbAppId = config.appId;
  }
  if (fbAppId !== fbAppIdForToken) {
    // handle incorrect token
    return;
  }

  const expiresAt = req.body.expiresAt;
  const authResult = Facebook.handleAuthFromAccessToken(fbAccessToken, expiresAt);

  const { userId } = Accounts.updateOrCreateUserFromExternalService(
    'facebook',
    authResult.serviceData,
    authResult.options
  );
  ...
};
```

### DIFF
--- a/packages/facebook/facebook_server.js
+++ b/packages/facebook/facebook_server.js
@@ -2,12 +2,7 @@ Facebook = {};
 
 var querystring = Npm.require('querystring');
 
-
-OAuth.registerService('facebook', 2, null, function(query) {
-
-  var response = getTokenResponse(query);
-  var accessToken = response.accessToken;
-
+Facebook.handleAuthFromAccessToken = function handleAuthFromAccessToken(accessToken, expiresAt) {
   // include all fields from facebook
   // http://developers.facebook.com/docs/reference/login/public-profile-and-friend-list/
   var whitelisted = ['id', 'email', 'name', 'first_name',
@@ -17,9 +12,8 @@ OAuth.registerService('facebook', 2, null, function(query) {
 
   var serviceData = {
     accessToken: accessToken,
-    expiresAt: (+new Date) + (1000 * response.expiresIn)
+    expiresAt: expiresAt
   };
-
 
   var fields = _.pick(identity, whitelisted);
   _.extend(serviceData, fields);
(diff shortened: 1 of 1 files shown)


gsabran:
@meteor-bot done :)

tmeasday:
Thanks @gsabran, seems reasonable.

tmeasday:
@gsabran are you interested in further maintaining the facebook packages? We are looking for maintainers for some of our more peripheral packages.

gsabran:
I'm happy to contribute when it's easy for me! I'm still the only dev at my startup so I don't have much extra bandwidth. And I'd rather not take responsibilities I can't honor. That being said, I'm happy to contribute as much as I can. So I'd rather not tell you that I'd maintain the package, but I'll try to be helpful.

I'm working on integrating 3rd parties services from mobile clients with a Meteor backend. They (so far for Facebook at least) have flows that differ from web based oauth. I'll make PR when I think it makes sense to bring more flexibility in the accounts packages.

VitalyChe:
@gsabran Thank you for your commit, I am just looking into authenticating FB/Google the same way as you proposed for mobile apps. What do you think about Google, could it be done the same way?
Thanks

gsabran:
I haven't look into Google yet, but will probably soon. If that's helpful, here's the full code I'm been using things for Facebook mobile auth in my code: https://gist.github.com/gsabran/8bab053beb05a0a5bf871b4caa00f9b6

PR #7575, opened 2016-08-04
test if a Facebook is valid and corresponds to the register Facebook …

gsabran:
Ask Facebook if a given access token is valid. Additionally, make sure the corresponding Facebook app is the one registered with the Meteor application.

### DIFF
--- a/packages/facebook/facebook_server.js
+++ b/packages/facebook/facebook_server.js
@@ -106,3 +106,27 @@ var getIdentity = function (accessToken, fields) {
 Facebook.retrieveCredential = function(credentialToken, credentialSecret) {
   return OAuth.retrieveCredential(credentialToken, credentialSecret);
 };
+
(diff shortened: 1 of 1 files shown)


gsabran:
I'm not sure I'm catching network error in the best way possible. Any input welcome!

tmeasday:
Hey @gsabran.

Thanks for the PR. 

In general with changes like this, it's best to get consensus around it by opening a feature issue before proposing a PR. See: https://github.com/meteor/meteor/blob/devel/Contributing.md#proposing-your-change

I think this is an example of a feature that could be part of an external package and as such not the kind of change we should merge at the moment.

gsabran:
Ok thanks for the guideline. It'd be a bit weird to have an external package while there's already a Facebook package within Meteor. I know you guys would like this kind of package to be moved out of Meteor..

PR #7576, opened 2016-08-04
add on option to register a callback after a user has been created

gsabran:
The current hook `Accounts.onCreateUser` runs before a user is created and is useful to modify the record that is going to be inserted.
I've added another hook that is called after the user is created and would be useful for functions that require a record in the db, such as sending a verification email.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -113,7 +113,7 @@ export class AccountsServer extends AccountsCommon {
   /**
    * @summary Customize new user creation.
    * @locus Server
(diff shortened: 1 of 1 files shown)


tmeasday:
Which fields are on the user after it's been created that aren't there before?

Let's take this to a feature discussion before proposing a PR.

gsabran:
It's not so much about fields than record existence in database. For instance, `Accounts.sendVerificationEmail` takes the user id in input, not an user object, and then fetches all it needs from the database.

So you can't do `Accounts.onCreateUser((user) => Accounts.sendVerificationEmail(user._id))`

PR #7586, opened 2016-08-05
Rename PACKAGE_DIRS to METEOR_PACKAGE_DIRS

### LINKED ISSUE #7585: PACKAGE_DIRS environment variable is too generic
Since the name `PACKAGE_DIRS` is too generic, some softwares already use this kind of environment variable for loading packages, for example.. GNU Make also use `PACKAGE_DIRS` and perl modules also use this name.

To avoid the conflict, we can't just add it in our bashrc because the meteor package is totally different.

I would suggest adding a prefix `METEOR_` to the existing `PACKAGE_DIRS`, but keep the original one for deprecation notice.

c9s:
Fix #7585 

Prefer `METEOR_PACKAGE_DIRS` over `PACKAGE_DIRS`. This is a backward compatible change.

### DIFF
--- a/tools/packaging/warehouse.js
+++ b/tools/packaging/warehouse.js
@@ -33,7 +33,7 @@
 ///
 /// The warehouse is not used at all when running from a
 /// checkout. Only local packages will be loaded (from
-/// CHECKOUT/packages or within a directory in the PACKAGE_DIRS
+/// CHECKOUT/packages or within a directory in the METEOR_PACKAGE_DIRS or PACKAGE_DIRS
 /// environment variable). The setup of that is handled by release.js.
 
 var os = require("os");
--- a/tools/project-context.js
+++ b/tools/project-context.js
@@ -1,5 +1,6 @@
 var assert = require("assert");
 var _ = require('underscore');
+var path = require('path');
 
 var archinfo = require('./utils/archinfo.js');
 var buildmessage = require('./utils/buildmessage.js');
@@ -596,10 +597,21 @@ _.extend(ProjectContext.prototype, {
     var self = this;
     var searchDirs = [files.pathJoin(self._projectDirForLocalPackages, 'packages')];
 
-    if (! self._ignorePackageDirsEnvVar && process.env.PACKAGE_DIRS) {
-      // User can provide additional package directories to search in
-      // PACKAGE_DIRS (colon-separated).
-      _.each(process.env.PACKAGE_DIRS.split(':'), function (p) {
+    // User can provide additional package directories to search in
+    // METEOR_PACKAGE_DIRS (colon-separated),
+    // This was added after 1.4.0.1
+    var envPackageDirs = process.env.METEOR_PACKAGE_DIRS || process.env.PACKAGE_DIRS;
+
+    // If the PACKAGE_DIRS, warn users to migrate to the new env vars.
+    if (process.env.PACKAGE_DIRS) {
+      console.warn("For compatibility reasons, the 'PACKAGE_DIRS' environment"
+                   + " variable is deprecated and will be removed in a future"
+                   + " release of Meteor.  Developers should now use"
+                   + " 'METEOR_PACKAGE_DIRS'.");
+    }
+    if (! self._ignorePackageDirsEnvVar && envPackageDirs) {
+      // path.delimiter was added in v0.9.3
+      _.each(envPackageDirs.split(path.delimiter), function (p) {
         searchDirs.push(files.pathResolve(p));
       });
     }
--- a/tools/cli/help.txt
+++ b/tools/cli/help.txt
@@ -505,7 +505,7 @@ modified.
 Packages may be specified by name or by path. If a package argument
 contains a '/', it is loaded from a directory of that name; otherwise,
 the package name is resolved according to the usual package search
-algorithm ('packages' subdirectory of the current app, $PACKAGE_DIRS
+algorithm ('packages' subdirectory of the current app, $METEOR_PACKAGE_DIRS
 directories, and core packages in that order). You can test any number
 of packages simultaneously. If you don't specify any package names
 then all available packages will be tested.
@@ -778,7 +778,7 @@ and rebuilds the package from scratch. Packages are specified by name.
 
 If you pass no arguments, this will rebuild all local packages.
 That includes packages found through the
-PACKAGE_DIRS environment variable, local packages in the current
+METEOR_PACKAGE_DIRS environment variable, local packages in the current
 application, and, if running Meteor from a checkout, the packages in
 the checkout. It doesn't include any packages for which we don't have
 the source.


laosb:
@c9s Would you like to adapt the solution @abernix introduced in his issue comment?

*(For your reference)*

> * The existing `PACKAGE_DIRS` should print a deprecation warning that it has been changed to use `METEOR_PACKAGE_DIRS`.  And this text as a `Console.warn` message with no wrapping:
> 
>  ```
> For compatibility reasons, the 'PACKAGE_DIRS' environment variable is deprecated and will be removed in a future release of Meteor.  Developers should now use 'METEOR_PACKAGE_DIRS'.
>  ```
>  * If `process.platform === 'win32'` then an additional message should print:
> 
>    ```
> Windows developers should now use a semi-colon (;) to separate multiple paths.
>    ```
> * **The existing `PACKAGE_DIRS` variable should keep working exactly as it is right now.**
> * A comment `//` should specify what version it was deprecated in with this issue number.
> * The new `METEOR_PACKAGE_DIRS` variable should use the `;` (semi-colon) separator when `process.platform === 'win32'` and use the `:` (colon) separator for everything else.
> 
> That's my thought.

c9s:
sure, all make sense. will see if i have to finish the PR

Wexpo Lyu <notifications@github.com> 於 2016年8月9日 星期二寫道：

c9s:
Updated changes according to the feedbacks.

tools/packaging/warehouse.js
```diff
@@ -33,7 +33,7 @@
 ///
 /// The warehouse is not used at all when running from a
 /// checkout. Only local packages will be loaded (from
-/// CHECKOUT/packages or within a directory in the PACKAGE_DIRS
+/// CHECKOUT/packages or within a directory in the METEOR_PACKAGE_DIRS or PACKAGE_DIRS
```
laosb: Wrap this line, it's longer than usual.

tools/project-context.js
```diff
@@ -596,10 +597,21 @@ _.extend(ProjectContext.prototype, {
...
+    // If the PACKAGE_DIRS, warn users to migrate to the new env vars.
+    if (process.env.PACKAGE_DIRS) {
+      console.warn("For compatibility reasons, the 'PACKAGE_DIRS' environment"
+                   + " variable is deprecated and will be removed in a future"
+                   + " release of Meteor.  Developers should now use"
+                   + " 'METEOR_PACKAGE_DIRS'.");
+    }
+    if (! self._ignorePackageDirsEnvVar && envPackageDirs) {
+      // path.delimiter was added in v0.9.3
+      _.each(envPackageDirs.split(path.delimiter), function (p) {
```
laosb: You should use `files.pathOsDelimiter` instead of `path.delimiter`, since it has been wrapped into `files`, which is meant to fill OS gaps.

laosb:
Last few lines of Travis Build Log:

```
email: updating npm dependencies -- mailcomposer, simplesmtp, stream-buffers...
compileLessBatch: updating npm dependencies -- less...
compileCoffeescript: updating npm dependencies -- coffee-script, source-map...
es5-shim: updating npm dependencies -- es5-shim...
lintJshint: updating npm dependencies -- jshint...
minifier-css: updating npm dependencies -- css-parse, css-stringify...
   Building package minifier-css             \
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).
The build has been terminated.
```

laosb:
Triggered a rebuild for you :)

laosb:
Tests passed!

laosb:
Given the fact that ES6 is available in this file, would you like use ES6 features when possible in the code you changed?

c9s:
sure, which part do you want to use es6?

laosb:
ES6 usage is not necessary, please just fix 2 points I've mentioned in your patch above.

tools/project-context.js
```diff
@@ -596,10 +597,21 @@ _.extend(ProjectContext.prototype, {
...
+    // METEOR_PACKAGE_DIRS (colon-separated),
+    // This was added after 1.4.0.1
+    var envPackageDirs = process.env.METEOR_PACKAGE_DIRS || process.env.PACKAGE_DIRS;
+
+    // If the PACKAGE_DIRS, warn users to migrate to the new env vars.
+    if (process.env.PACKAGE_DIRS) {
+      console.warn("For compatibility reasons, the 'PACKAGE_DIRS' environment"
+                   + " variable is deprecated and will be removed in a future"
+                   + " release of Meteor.  Developers should now use"
+                   + " 'METEOR_PACKAGE_DIRS'.");
```
laosb: `console` -> `Console`

Use `+` on the end of the line, and have the space on the end of the line, like:

```
Console.warn('Foo ' +
  'bar.');
```

or simply use `,` to separate, and have no space on the edge of each line:

```
Console.warn('Foo',
  'bar.');
```

tools/project-context.js
```diff
@@ -596,10 +597,21 @@ _.extend(ProjectContext.prototype, {
...
     var searchDirs = [files.pathJoin(self._projectDirForLocalPackages, 'packages')];
 
-    if (! self._ignorePackageDirsEnvVar && process.env.PACKAGE_DIRS) {
-      // User can provide additional package directories to search in
-      // PACKAGE_DIRS (colon-separated).
-      _.each(process.env.PACKAGE_DIRS.split(':'), function (p) {
+    // User can provide additional package directories to search in
+    // METEOR_PACKAGE_DIRS (colon-separated),
+    // This was added after 1.4.0.1
+    var envPackageDirs = process.env.METEOR_PACKAGE_DIRS || process.env.PACKAGE_DIRS;
```
abernix: Per the original feature request, one of the original requirements was for:

> **The existing PACKAGE_DIRS variable should keep working exactly as it is right now.**

...this was because there over 100 pages of GitHub search results of projects using `PACAKGE_DIRS` already.

By doing it this way (e.g. `METEOR_PACKAGE_DIRS || PACKAGE_DIRS`) and changing the code below to use `path.delimiter`, this requirement is no longer met which means this becomes a breaking-change for Windows projects which are using `PACKAGE_DIRS`.

Therefore, I think it's important that`PACKAGE_DIRS` continue to use `:` regardless of the OS since this is the way it has behaved in the past and it's used quite frequently.

tools/project-context.js
abernix: Yes, the new `METEOR_PACKAGE_DIRS` should use `files.pathOsDelimiter` however `PACKAGE_DIRS` should continue to use `:` explicitly, so this code is not entirely right (see my above comment).

laosb:
@c9s Do you still work on this?

abernix:
@c9s Thanks very much for opening this PR originally.  Assuming you're busy, I've repackaged your changes into #7919.  Closing this PR in favour of that.

c9s:
Thank you

Jesse Rosenberger <notifications@github.com> 於 2016年10月13日 星期四寫道：

PR #7591, opened 2016-08-05
Make commits links in changelog

lieuwex:


laosb:
I guess we can safely use 6-character commit ID in the link text.

lieuwex:
@laosb Should I change it to that?

laosb:
Just my personal thought 💭 @abernix Any thoughts?

tmeasday:
An improvement, thanks

PR #7595, opened 2016-08-05
Use pure-JavaScript implementation for npm-bcrypt package.

benjamn:
According to the `README`, this implementation is approximately 2.7 times slower than native: https://www.npmjs.com/package/bcryptjs

Apps that wish to continue using the native `bcrypt` package should run `meteor npm install --save bcrypt` in the root application directory, and the `npm-bcrypt` package will prefer that implementation.

This change should reduce the pain of Windows developers who have had trouble installing a working compiler toolchain to rebuild the native `bcrypt` package.


packages/npm-bcrypt/wrapper.js
```diff
@@ -0,0 +1,8 @@
+try {
+  // If you really need the native `bcrypt` package, then you should
+  // `meteor npm install --save bcrypt` into the node_modules directory in
+  // the root of your application.
+  exports.NpmModuleBcrypt = require("bcrypt");
+} catch (e) {
+  exports.NpmModuleBcrypt = require("bcryptjs");
```
zol: Should we warn users that they're using the slower JavaScript bcrypt library here? I suspect most users won't read our release notes carefully enough to know this.

benjamn: I guess the question is how often we should warn, since we don't want the warning to be too noisy?

packages/npm-bcrypt/wrapper.js
```diff
@@ -0,0 +1,38 @@
...
+};
+
+try {
+  // If you really need the native `bcrypt` package, then you should
+  // `meteor npm install --save bcrypt` into the node_modules directory in
+  // the root of your application.
+  var bcrypt = require("bcrypt");
+} catch (e) {
+  bcrypt = require("bcryptjs");
+  console.warn([
```
zol: That's great - I think it's acceptable to be noisy in this case as the users who typically care about noisiness are also the ones who will be eager to avoid using `bcryptjs`

mitar:
Is it really a problem if bcrypt is slow? Isn't the whole point of it to be slow?

benjamn:
Yes! For new hashes, you can pick whatever constant factors you want to make the hashing as slow as you like. But it's still worth noting that there are lots of old hashes out there already, and you can't just recompute them, since that would require unknown information like salts and passwords. So a really big performance difference would not be something we could hide. In this case, 2-3x seems tolerable, and you can always install `bcrypt` yourself if you need exactly the same performance characteristics.

mitar:
Yes, but that scary warning ... People might try to install bcrypt and have issues when they would not really have to.

Torsten85:
I got rid of this warning by adding bcrypt as mentioned (```meteor npm install --save bcrypt```). But after building my app for production (```meteor build --architecture os.linux.x86_64```), i still get the message on my production server.

Am I missing something?

abecks:
I also receive this warning in production after using `meteor build --architecture os.linux.x86_64`.

I also tried to manually install `bcrypt` in `bundle/programs/server` via `npm install bcrypt` but I still receive the warning when starting my app.

In my package.json:
``
"bcrypt": "^0.8.7",
``

mitar:
Are you running `meteor npm install` before running `meteor build` on the same machine?

abecks:
No, I was under the impression that `meteor npm install --save bcrypt` was installing it. I will try it now.

mitar:
It installs it, but locally. So do you run then `meteor build` on the same machine. Or do you do this and then run `meteor build` on another machine where you have not run `meteor npm install` as well?

abecks:
I do run `meteor build` on the same machine. Should I try installing `bcrypt` globally then building? I should note that I do have a `bcrypt` folder under `node_modules` in my project root, so it seems installed. I also do not receive the warning when running the app under the `meteor` command. I only receive the warning in production after building.

I am building on OS X 10.11.3 and deploying to CentOS 6. In the deployment, I do an `npm install` in `bundle/programs/server`.

mitar:
Then this seems like a bug.

I think you should open a new issue (or find an existing one) about this issue. Commenting on past pull requests is not very useful from issues management perspective.

mitar:
(Reference a new issue from here.)

Issue #7605, opened 2016-08-08
Surprising path error when using an npm module with long relative path


Kurounin:
**Meteor version:** 1.4.0.1
**OS:** Windows 10 64bit
**Bug repository & steps to reproduce:** https://github.com/Kurounin/meteor-absolute-path-bug
**Description of bug:** If you use an npm module in the project with a long relative path (e.g.: _../../../build/Release/kerberos_) and you place the project in the drive root, when building the project meteor will join the current absolute file path and the required relative path and the resulting path will be _/build/Release/kerberos_ which will cause it to crash with a fatal error.

abernix:
I was able to successfully reproduce this problem on Windows 10 64-bit with Meteor 1.4.0.1 and 1.4.1.3, but after upgrading to Meteor 1.4.2 this problem went away – without the pull-request.

A lot has changed in 1.4.2 with the isobuilder so I would attribute the change to that.  If you find that this is not fixed for you when you update to 1.4.2, please ping me and we can reopen, but I'm fairly confident that this is resolved. :smile:

PR #7606, opened 2016-08-08
Fix for requiring long relative path

Kurounin:
Fix for issue https://github.com/meteor/meteor/issues/7605

**Meteor version:** 1.4.0.1
**OS:** Windows 10 64bit
**Bug repository & steps to reproduce:** https://github.com/Kurounin/meteor-absolute-path-bug
**Description of bug:** If you use an npm module in the project with a long relative path (e.g.: _../../../build/Release/kerberos_) and you place the project in the drive root, when building the project meteor will join the current absolute file path and the required relative path and the resulting path will be _/build/Release/kerberos_ which will cause it to crash with a fatal error.

### DIFF
--- a/tools/static-assets/server/mini-files.js
+++ b/tools/static-assets/server/mini-files.js
@@ -88,6 +88,10 @@ var wrapPathFunction = function (name, partialPaths) {
       var result = f.apply(path, args);
       if (typeof result === "string") {
         result = toPosixPath(result, partialPaths);
+        // if final path is without root, add it from initial argument
+        if (name === 'join' && ! /^\/[A-Za-z](\/|$)/.test(result) && /^\/[A-Za-z](\/|$)/.test(arguments[0])) {
+          result = '/' + arguments[0][1] + result;
+        }
       }
 
       return result;


laosb:
I guess the `@platform` is not working in your description. Removed it.

laosb:
Circle CI gave an unrelated failure:

```
1 failure:
  - static-html: static-html - add static content to head and body
```

Triggered a rerun.

tmeasday:
@Kurounin can you explain exactly which path function is failing and with which arguments, and how this fix solves the issue?

Kurounin:
@tmeasday If you require an npm module (in this case Kerberos) and in Windows you put your meteor project in C:\meteor-project\ then when you start it it will throw an error that it can't find /build/Release/kerberos.
This happens because inside kerberos npm module there is a `require("../../../build/Release/kerberos")` and when meteor does the conversion to unix path (C/meteor-project/) and tries to resolve that path it will result in `/build/Release/kerberos` which will fail the conversion back to Windows path, because no drive letter is present anymore.
My fix prevents removing the drive letter when resolving relative paths in Windows.

It should also be investigated why modules requiring external files (kerberos, edge) fail to find those files when used in meteor projects.

tmeasday:
@Kurounin is it `files.pathRelative`? And what are the arguments and real vs expected output?

I ask because it doesn't seem easy to produce a test for this but I'd like to understand the exact nature of the fix.

Kurounin:
@tmeasday In https://github.com/Kurounin/meteor-absolute-path-bug you can see the full stack.

**/tools/isobuild/resolver.js** runs `const joined = pathJoin('C/meteor-absolute-path-bug', '../../../build/Release/kerberos');`
which will run **/tools/static-assets/server/mini-files.js** `wrapPathFunction()` that will first run `toDosPath()` for each argument, then `join()` them and afterwards convert them back using `toPosixPath()`.

At the end we will have `const joined = '/build/Release/kerberos';` which is not correct.

So when **/tools/isobuild/resolver.js** gets to `const path = pathNormalize(joined);` and runs `wrapPathFunction()` the `toDosPath()` will fail because drive letter is no longer present.

tmeasday:
So you are saying at some point it runs

```js
// I added the '/', I assume it was a typo above?
files.pathJoin('/C/meteor-absolute-path-bug', '../../../build/Release/kerberos')

// which returns (incorrectly)
'/build/Release/kerberos'

// after your patch, it returns
'/C/build/Release/kerberos'
```

But that directory doesn't exist either, right? It seems like the real problem here is why kerberos is requiring a file from a path that make sense?

Looking at the source code, the only places I see it doing `require('../../../build/Release/kerberos')` seem like sensible locations, so is this not an error somewhere else in the stack (it's joining with the wrong `absParentPath` here: https://github.com/meteor/meteor/blob/devel/tools/isobuild/resolver.js#L95)?

Kurounin:
Yes, technically the path after my patch is wrong for the kerberos requirement (meteor tries to resolve relative paths by trying to go up from initial directory until it reaches meteor app root directory), but it returns a valid Windows directory path.

The issue is that both kerberos & edge (try using edge in linux and you will get an error as well, because https://github.com/tjanczuk/edge/blob/master/lib/edge.js will not be able to find `'../build/Release/edge_nativeclr.node'`.
The files themselves exist in the node_modules npm package directory, but not in the built meteor directory. These files are not actually included in the npm package, they are built when running `npm install` so why these aren't added to the meteor build should be investigated as well.

abernix:
Thanks very much for reporting the issue and opening the pull request to try to fix it.

Per my comment on the issue: 

> I was able to successfully reproduce this problem on Windows 10 64-bit with Meteor 1.4.0.1 and 1.4.1.3, but after upgrading to Meteor 1.4.2 this problem went away.

While I'm not sure specifically what resolved the problem, it would seem (and is likely) that the bug existed somewhere else outside this logic.  Anyhow, this PR no longer seems necessary.  Closing but please reopen the issue if the problem continues for you in 1.4.2.

PR #7610, opened 2016-08-09
Update README.md for Windows users.

laosb:


README.md
```diff
@@ -91,6 +93,8 @@ rm -rf ~/.meteor/
 sudo rm /usr/local/bin/meteor
 ```
 
+On Windows, just run the uninstaller.
```
benjamn: Can you say a little more here about going to "Control Panel" > "Uninstall a program"?

laosb: @benjamn That's because different Windows versions have different title of the "Uninstall a program". How about just mention that it was in Control Panel, like "just go to Control Panel and uninstall it"?

benjamn: Sounds good!

PR #7616, opened 2016-08-09
Don't touch ~/.meteor/meteor after background updates.

benjamn:
Just because the background updater manages to download a new release, we should not necessarily update the `~/.meteor/meteor` symlink (or `AppData\Local\.meteor\meteor.bat` on Windows) to refer to the new release. That should happen only when the user explicitly runs `meteor update`, i.e. [here](https://github.com/meteor/meteor/blob/eebf538ee00434edeae60ed2ad053e3f76d1b6a6/tools/cli/commands-packages.js#L1346).

### DIFF
--- a/tools/packaging/updater.js
+++ b/tools/packaging/updater.js
@@ -61,8 +61,6 @@ var checkForUpdate = function (showBanner, printErrors) {
     return;
   }
 
-  updateMeteorToolSymlink(printErrors);
-
   maybeShowBanners();
 };
 


tmeasday:
Also, it might be worth figuring out once and for all if background updates can block development. In my experiments, they do not, but users seem to report that they do fairly frequently. Perhaps they are just seeing a message and assuming the development server isn't running? I'm not sure.

benjamn:
@tmeasday I believe this pull request fixes the primary problem with background updates, namely that background updates expose you to bugs in the latest release, since `~/.meteor/meteor` was getting updated unconditionally and involuntarily.

tmeasday:
@benjamn can we confirm that the symlink will later be put in place?

PR #7619, opened 2016-08-10
Work in progress: extract blaze from core packages

tmeasday:
Removing packages at https://github.com/meteor/blaze from the repository.

### Things that I have been able to do in this branch:

 - Publish a release (`TMEASDAY-RELEASE-TESTING@1.4.0.1`) with all remaining core packages depending on specified versions of relevant blaze packages.

 - Create an app with that release and it seems to work just fine.

### Still to be resolved:

How to deal with isopackets (see https://github.com/meteor/meteor/pull/7619/commits/1346b22d60be4df58f235962f70346481041ab56)

I think the best solution would be to add any non-core packages which are used by isopackets as submodules, in such a way that the code can still be used by the isopacket mechanism, but the packages are *not* picked up by the package / release scripts.

### NOTE

It doesn't make sense to merge this branch as it incorporates the machinery of the test release above. Once we resolve the issue above, I'll make a new branch without the bumped version numbers.


tmeasday:
Also, @benjamn - any thoughts on why I'd get 

> promise.await is not a function

When running self-tests? Could it be the dependencies of the packages removed from the tool in this commit: https://github.com/meteor/meteor/pull/7619/commits/1346b22d60be4df58f235962f70346481041ab56 ?

benjamn:
No, the error about `promise.await` is due to `phantomjs-prebuilt` using a bad `Promise` polyfill that overwrites the `global.Promise` constructor. You should rebase against `devel` to include [this commit](https://github.com/meteor/meteor/pull/7569).

tools/tool-env/isopackets.js
```diff
@@ -227,8 +227,10 @@ var newIsopacketBuildingCatalog = function () {
       // expect them to all build, and we're fine with dying if they don't
       // (there's no worries about needing to springboard).
       isopacketCatalog.initialize({
-        localPackageSearchDirs: [files.pathJoin(
-          files.getCurrentToolsDir(), 'packages')],
+        localPackageSearchDirs: [
+          files.pathJoin(files.getCurrentToolsDir(), 'packages'),
+          files.pathJoin(files.getCurrentToolsDir(), 'packages-for-isopackets', 'blaze', 'packages')
```
benjamn: Nice! If/when we turn the command-line tool into an npm package, we can drop the `packages` directory and move everything isopacket-related into `packages-for-isopackets`.

Last I checked, that list was relatively short:
```
babel-compiler
babel-runtime
base64
blaze
blaze-tools
boilerplate-generator
check
constraint-solver
contsraint-solver
ddp-client
ddp-common
ddp-rate-limiter
ddp-server
deps
diff-sequence
ecmascript
ecmascript-runtime
ejson
html-tools
htmljs
id-map
jquery
logging
logic-solver
meteor
modules
modules-runtime
mongo-id
npm-mongo
observe-sequence
package-version-parser
promise
random
reactive-var
retry
spacebars
spacebars-compiler
tracker
ui
underscore
webapp-hashing
xmlbuilder
```

tmeasday:
Closed in favour of https://github.com/meteor/meteor/pull/7633

Issue #7636, opened 2016-08-12
Can't use private NPM servers that require authentication


sebakerckhof:
Commit https://github.com/meteor/meteor/pull/7162/commits/a8063e81f7ddb25c541af85c9b4b4d48fe890322 made it so that if you use a private NPM server, Meteor tool will check if that one's available instead of the default one. This is good, but the problem is that most private NPM servers require authentication, even for get requests. 

This will result in meteor throwing:
`error: Can't install npm dependencies. Are you connected to the internet?` even if the server is available.

benjamn:
You might need to set the `$NPM_TOKEN` environment variable? https://docs.npmjs.com/private-modules/ci-server-config

sebakerckhof:
That's not the problem. The problem is meteor checks if the registry url is available using a normal http get request. If it isn't or errors, it will halt.

abernix:
This will be fixed in a future stable release of Meteor but is available in beta right now (specifically, `1.4.2-beta.7`)

The beta channel isn't always bug free and you may run into unrelated bugs, but if you're in a position to test that this particular feature has been fixed for you, it would certainly be appreciated.  You can find instructions for using the current beta release in #7668.

Please report back if you still have problems with this!

PR #7637, opened 2016-08-12
Do not check connection with protected npm registries

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/7636

Now it just doesn't check the connection if we need to authenticate for get requests (according to the `always-auth` npm setting). This is because there are multiple ways to authenticate to NPM servers (http auth, certs, bearer tokens?) and it'd be hard to make all of them work.

### DIFF
--- a/tools/isobuild/meteor-npm-userconfig
+++ b/tools/isobuild/meteor-npm-userconfig
@@ -6,3 +6,10 @@
 
 loglevel = error
 progress = false
+
+# By default if the NPM registry is unavailable, it will timeout after more than 60s.
+# These settings make sure we throw an error faster
+fetch-retries = 1
+fetch-retry-factor = 4
+fetch-retry-mintimeout = 1000
+fetch-retry-maxtimeout = 10000
--- a/tools/isobuild/meteor-npm.js
+++ b/tools/isobuild/meteor-npm.js
@@ -996,7 +996,9 @@ var installFromShrinkwrap = function (dir) {
 // dependencies. `npm install` times out after more than a minute.
 var ensureConnected = function () {
   try {
-    httpHelpers.getUrl(process.env.NPM_CONFIG_REGISTRY || "http://registry.npmjs.org");
+    if(!process.env["NPM_CONFIG_ALWAYS-AUTH"]){
+      httpHelpers.getUrl(process.env.NPM_CONFIG_REGISTRY || "http://registry.npmjs.org");
+    }
   } catch (e) {
     buildmessage.error("Can't install npm dependencies. " +
                        "Are you connected to the internet?");
--- a/tools/isobuild/meteor-npm.js
+++ b/tools/isobuild/meteor-npm.js
@@ -886,7 +886,6 @@ var getShrinkwrappedDependencies = function (dir) {
 };
 
 var installNpmModule = function (name, version, dir) {
-  ensureConnected();
 
   var installArg = utils.isNpmUrl(version)
     ? version : (name + "@" + version);
@@ -959,8 +958,6 @@ var installFromShrinkwrap = function (dir) {
       "Can't call `npm install` without a npm-shrinkwrap.json file present");
   }
 
-  ensureConnected();
-
   const tempPkgJsonPath = files.pathJoin(dir, "package.json");
   const pkgJsonExisted = files.exists(tempPkgJsonPath);
   if (! pkgJsonExisted) {
@@ -990,21 +987,6 @@ var installFromShrinkwrap = function (dir) {
       recordLastRebuildVersions(pkgDir);
     }
   });
-};
-
-// ensure we can reach http://npmjs.org before we try to install
-// dependencies. `npm install` times out after more than a minute.
-var ensureConnected = function () {
-  try {
-    if(!process.env["NPM_CONFIG_ALWAYS-AUTH"]){
-      httpHelpers.getUrl(process.env.NPM_CONFIG_REGISTRY || "http://registry.npmjs.org");
-    }
-  } catch (e) {
-    buildmessage.error("Can't install npm dependencies. " +
-                       "Are you connected to the internet?");
-    // Recover by returning false from updateDependencies
-    throw new NpmFailure;
-  }
 };
 
 // `npm shrinkwrap`


laosb:
Sounds good to me.

benjamn:
Maybe we could avoid the problems this PR is meant to address by merely attempting to resolve the hostname via DNS, so authentication never becomes a problem?

```js
function ensureConnected() {
  const hostname = require("url").parse(
    process.env.NPM_CONFIG_REGISTRY || 
    "http://registry.npmjs.org"
  ).hostname;

  return new Promise((resolve, reject) => {
    require("dns").resolve(hostname, (err, ip) => {
      err ? reject(err) : resolve(ip);
    });
  }).await();
}
```

sebakerckhof:
@benjamn wouldn't that change the semantics of `ensureConnected` in a way that makes the check useless?

The way I understand it, `ensureConnected` is used solely to check if a host is online. This is done as a small optimization since NPM would only time out after a minute or so. Therefore this pre-`npm install` check, to save the user that minute.
Am I right that in case `registry.npmjs.org` is offline, it's DNS would still be resolved?

sebakerckhof:
@benjamn any chance of getting this in for the next release?

zol:
@benjamn, @sebakerckhof's fix seems reasonable. [NPM_CONFIG_ALWAYS-AUTH](https://docs.npmjs.com/misc/config#always-auth) is a legit environment variable used for roughly this purpose.

abernix:
@sebakerckhof To be fair, skipping the connection check when `NPM_CONFIG_ALWAYS-AUTH` is set also defeats the intent of `ensureConnected`.

There seem to be a variety of `npm config`-able parameters that are meant to solve this exact problem.  For example: [fetch-retry-maxtimeout](https://docs.npmjs.com/misc/config#fetch-retry-maxtimeout).  The `ensureConnected` logic is fairly old, dating back to f31457a017499b9091fe0b6166a0c2bdb21f23d5 and I'm not sure these config options existed at the time.

How about trying to set `fetch-retry-maxtimeout` to a value lower than `60000` (the default) and completely removing the `ensureConnected` check?  Plus, there are other `fetch-*` config parameters (`fetch-retries`, `fetch-retry-factor` and `fetch-retry-mintimeout`) which could also be helpful.  `ensureConnected` is a function that I have recently questioned myself since during the test environment setup it runs ~25 times.

It might be as easy as putting the `fetch-*` config params into https://github.com/meteor/meteor/blob/devel/tools/isobuild/meteor-npm-userconfig.

Thoughts?

sebakerckhof:
@abernix 

> To be fair, skipping the connection check when NPM_CONFIG_ALWAYS-AUTH is set also defeats the intent of ensureConnected.

True

>  I'm not sure these config options existed at the time.

https://github.com/meteor/meteor/commit/f31457a017499b9091fe0b6166a0c2bdb21f23d5 is from Jan 24, 2013
These properties have been configurable since https://github.com/npm/npm/commit/419ffa05054fc3956ecad9fe8f124a70fcf935b6 which dates back to Jun 28 2012

So it was either overlooked or not used for a good reason, but I assume it's the first.

The formula for the retry is:
```js
Math.min(random * minTimeout * Math.pow(factor, attempt), maxTimeout)
```

Default values of npm are:
minTimeout: 10
factor: 10
retries: 2
maxTimeout: 60

What behavior seems the best to you:
1. Just setting retries to 0, which will give the same behavior as ensureConnected
2. Just setting maxTimeout to 5, then it will retry twice totaling around 10 seconds before it fails?

abernix:
> Just setting retries to 0, which will give the same behavior as ensureConnected

Not retrying at all just seems undesirable to me.  I think we've seen that with other things (installer, package downloads, etc.)

Maybe 3 retries in 20 seconds?  Just putting this out there: `minTimeout: 1000ms`, `factor: 4`, `retries: 3`, `maxTimeout: 21000ms`, which should yield a retry at `1s`, `5s` and `21s`.

sebakerckhof:
sounds reasonable to me. I'll test it out

sebakerckhof:
@abernix . Okay, it works a bit different since those are just the times between the retries, not the times at which it does a retry. + it takes some time to know when a host is unavailable.
So with your parameters you almost end up with again 60s time before you get an error. So I'll tune the parameters a bit and open a new PR.

sebakerckhof:
New PR: https://github.com/meteor/meteor/pull/7811 .

abernix:
Play with the settings a bit and see what you come up with that seems reasonable and report back what happens here.  Then let's wait for input before proceeding with another PR.

sebakerckhof:
Ok, I thought you were part of MDG, my mistake.

These settings seemed ok for me:
fetch-retries = 1
fetch-retry-factor = 4
fetch-retry-mintimeout = 1000
fetch-retry-maxtimeout = 10000 

(maxTimeout not really necessary.)
For an unreachable host, this threw an error in about 20s.

Just copy-pasting from my other PR why I think this is the right approach:

1. It works for registries that require HTTP get authentication, thus fixing #7636
2. It makes builds a bit faster since the ensureConnected check can be called quite a lot (~25 times in @abernix test environment, see: #7637 (comment) ). 
3. We end up with less code to maintain.
4. More flexible behavior. Currently it will retry once, which in my tests throws an error after about 20s for an unreachable host. You could disable retries altogether or allow for more retries.

hwillson:
> Ok, I thought you were part of MDG

He should be - MDG, please make this happen! :-)

abernix:
@sebakerckhof Ok, then this should yield two total attempts (one initial, one retry), which is probably reasonable. I think getting the timeouts to happen any quicker will be difficult since it's dependent on the kernel TCP timeout.  And I could be wrong, but end-users, especially those with unfavorable network conditions, should be able to tweak these settings using environment variables. (e.g. `NPM_CONFIG_FETCH-RETRIES`, etc.) which I believe take precedence.

Does this currently output a pleasing error message, similar to the previous `Can't install npm dependencies.  Are you connected to the internet?` message?  If not, it would be really nice to catch the `ETIMEDOUT` and do so.

Ultimately, I just don't know if `ensureConnected` served another, higher purpose, but I hope this will be a suitable change.  So again, let's wait for more insight.

>> Ok, I thought you were part of MDG
>
> He should be - MDG, please make this happen! :-)

@hwillson Please open a pull-request. :wink:😜

sebakerckhof:
Well, you get a meteor error like:
` error: couldn't install npm package packagename@version: Command failed:`
followed by a typical npm error, which is a bit more obscure (quite a load of text), but IMO is more helpful since you can get quite a lot of different erorrs. Next to a timeout, there can be an authentication error, authorization error, dns error, ... 
The default npm errors do give an indication what the problem could be.

benjamn:
@sebakerckhof Speaking for MDG (on the advice of @abernix), I'll be happy to merge this if you move the contents of #7811 into this PR.

sebakerckhof:
@benjamn done.

benjamn:
Thanks @sebakerckhof!

PR #7656, opened 2016-08-16
Infinitely retry to connect to Mongo on start. Fixes #7655

### LINKED ISSUE #7655: Attempt to reconnect to mongo if initial connection fails.
If I `export MONGO_URL=mongo://localhost/database` and start meteor, it fails to connect to the database, never retries and exits.

```
MongoError: failed to connect to server [xxx:xxx] on first connect
    at Object.Future.wait (xxx/bundle/programs/server/node_modules/fibers/future.js:449:15)
    at new MongoConnection (packages/mongo/mongo_driver.js:216:27)
```

Using Mongo 1.4

rws-github:
Many times when we deploy our Meteor apps, they fail at the start due to a failed initial connection to the database. Starting them again seems to fix the problem. I would like the initial connection attempts to at least retry for a period to prevent the app from failing right from the start.

It looks like this would be a feature or enhancement that requires some discussion. Perhaps the retry should be capped at a number of attempts or that it is configurable along with the retry interval.

There also appears to be a test that checks that a failed connection attempt throws an exception which this change breaks, so a decision on what to do there is necessary or the retry capability could be completely optional and configured with settings or env vars which allows the test to work just fine.

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -123,6 +123,64 @@ var replaceTypes = function (document, atomTransformer) {
   return ret;
 };
 
+var connectToDb = function (future, self, url, mongoOptions) {
+  var connectFuture = new Future;
+  MongoDB.connect(
+    url,
+    mongoOptions,
+    Meteor.bindEnvironment(
+      function (err, db) {
+        if (err) {
+          throw err;
+        }
+
+        // First, figure out what the current primary is, if any.
+        if (db.serverConfig.isMasterDoc) {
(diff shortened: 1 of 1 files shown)


zol:
@rws-github - Please follow our contribution guidelines and update this PR accordingly (e.g sign our contributor agreement, make sure tests pass, provide a proper description).

zol:
@tmeasday - You've been looking at Mongo recently. What do you think of retrying the initial connection?

tmeasday:
@rws-github let's take the discussion to the issue and reopen this PR / a new PR when we've concluded what the best approach is.

PR #7699, opened 2016-08-24
Pass object with error to onLoginFailure hooks

### LINKED ISSUE #7385: [feature-request] Provide arguments to onLogin / onLoginFailure on client
There's some discussion of this at #3572 but that issue is closed and not specific to this issue.

It would be good to have at least the type and error passed to `onLogin`/`onLoginFailure` callbacks on the client. I can sort of understand the argument for not providing the method arguments e.g. email/username & password (although the point is moot, you either trust the client side code you're bundling or it shouldn't be in there) but the other non-sensitive-information arguments that are available on the server should be supplied. These are `type`, `allowed`, `error`, `user`, `connection`, and `methodName` [but not `methodArguments`].

The use case is checking for a specific error on login, which is fine if using a login form that calls `Meteor.loginWithPassword` with a callback but not fine on a session resume / page refresh etc. If a specific error appears I need to redirect the user to a different website along with a login token for that website. I want to provide these details in the error object from the server because I don't see another way of returning anything from a login attempt to the user.

mjmasn:
Fixes #7385 

For consistency with the server methods, this passes an object to the onLoginFailure hooks, but unlike on the server it contains only the error associated with the failed login attempt.

### DIFF
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -191,7 +191,7 @@ Ap.callLoginMethod = function (options) {
       });
     } else {
       self._onLoginFailureHook.each(function (callback) {
-        callback();
+        callback({ error: error });
         return true;
       });
     }
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -191,7 +191,7 @@ Ap.callLoginMethod = function (options) {
       });
     } else {
       self._onLoginFailureHook.each(function (callback) {
-        callback({ error: error });
+        callback({ error });
         return true;
       });
     }


mjmasn:
PR for docs: https://github.com/meteor/docs/pull/82

mjmasn:
@benjamn any chance of getting this merged for the next release? It's a really simple change, but allows us to see errors from all failed logins (i.e. session resumes, token logins), not just password logins. Currently have to edit a custom version of the package every time I update Meteor in all our apps :(

mitar:
This would be really great to have in.

packages/accounts-base/accounts_client.js
```diff
@@ -191,7 +191,7 @@ Ap.callLoginMethod = function (options) {
       });
     } else {
       self._onLoginFailureHook.each(function (callback) {
-        callback();
+        callback({ error: error });
```
benjamn: Can you maybe just change this to use shorthand notation (`callback({ error })`), and see if the CircleCI tests pass after rebasing?

PR #7706, opened 2016-08-25
Fix URL issue in History.md

mbrobbel:
Fix URL to issue fixed in 1.4.1.1 in History.md


laosb:
Just one character change could make a difference. Should merge it!

PR #7117, opened 2016-05-25
Client Accounts.forgotPassword and .verifyEmail pass error in callback

laosb:
Issue #5664 
Previous PR #5681

This rebased @mishavee 's work. Thank you @mishavee !

PR #7722, opened 2016-08-27
Move #7117 entry to v.NEXT in History.md

laosb:
The change was added in the v.NEXT, but later v.NEXT was changed into v1.3.3 for the 1.3.3 release. Now it was in 1.3.3, which is wrong and misleading. This PR move it to v.NEXT since it was not released till now.


laosb:
@tmeasday

tmeasday:
Thanks @laosb

PR #7727, opened 2016-08-29
Added autoprefixer to stylus

mitar:
Nib is very outdated. I was mostly using it to add vendor prefixes. But now people use autoprefixer for that. So I added it to the package.

One can configure which browsers to support using file options:

``` javascript
  api.addFiles([
    'file.styl'
  ], 'client', {
    autoprefixer: {
      browsers: ['last 3 Chrome versions', 'Firefox >= 7', 'Explorer >= 8', 'last 2 versions', '> 1%', 'Firefox ESR', 'Safari >= 4']
    }
  });
```

[See defaults here](https://github.com/ai/browserslist#queries).

### DIFF
--- a/packages/stylus/.npm/plugin/compileStylusBatch/npm-shrinkwrap.json
+++ b/packages/stylus/.npm/plugin/compileStylusBatch/npm-shrinkwrap.json
@@ -1,9 +1,19 @@
 {
   "dependencies": {
     "amdefine": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.0.tgz",
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.1.tgz",
       "from": "amdefine@>=0.0.4"
+    },
+    "autoprefixer": {
+      "version": "6.3.7",
+      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-6.3.7.tgz",
+      "from": "autoprefixer@6.3.7"
+    },
+    "autoprefixer-stylus": {
+      "version": "0.9.4",
+      "resolved": "https://registry.npmjs.org/autoprefixer-stylus/-/autoprefixer-stylus-0.9.4.tgz",
+      "from": "autoprefixer-stylus@0.9.4"
     },
     "balanced-match": {
       "version": "0.4.2",
@@ -14,6 +24,16 @@
       "version": "1.1.6",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.6.tgz",
       "from": "brace-expansion@>=1.0.0 <2.0.0"
+    },
+    "browserslist": {
+      "version": "1.3.6",
+      "resolved": "https://registry.npmjs.org/browserslist/-/browserslist-1.3.6.tgz",
+      "from": "browserslist@>=1.3.4 <1.4.0"
+    },
+    "caniuse-db": {
+      "version": "1.0.30000585",
+      "resolved": "https://registry.npmjs.org/caniuse-db/-/caniuse-db-1.0.30000585.tgz",
+      "from": "caniuse-db@>=1.0.30000488 <2.0.0"
     },
     "concat-map": {
       "version": "0.0.1",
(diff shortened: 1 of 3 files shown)


tmeasday:
Hmm, the downside of this approach is you can't configure imported files from the app. I would suspect a way of configuring autoprefixer globally is what we want, but of course this isn't easy to do (package level configuration, especially for a build plugin).

mitar:
Ahh. I completely forgot about app-level content. I am not using that in my apps.

Yea, then we have to do something similar to what Babel has which is ugly. And I was so happy for options to work so well.

Maybe we should simply create a way to configure top-level options for build tools? `.meteor/stylus.json` would be loaded as options for all `.styl` files? And each package could have `.meteor/<packagename>.json` which would add configuration to `options` for all top-level `addFiles`?

tmeasday:
I was thinking more along the lines of figuring out a good way to provide configuration for a build plugin. Maybe a principled version of what `juliancwirko:postcss` does, which is reading configuration out of `package.json`.

I know this is something people have wanted for a long time (static package-level config). I'm not sure what the argument is why we've never done it.

mitar:
Now that we have package.json this might be good place as well.

tmeasday:
Shall we close this and open a FR that is contingent on figuring out global configuration?

mitar:
Wouldn't it be better just to add global configuration code here? When somebody suggests how to do it? I can also give the person access to my branch to do so.

tmeasday:
Maybe you're right. Figuring out a system of global configuration is probably a bit OTT

mitar:
I updated the pull request. Can we get this merged in? ;-)

mitar:
I opened #7829 for global configuration of build plugins.

sebakerckhof:
The downside of this approach is that it needs to be done for each and every plugin that wants autoprefixing (sass, less, ...) whereas putting it in the minification step works for every preprocessor.
However, if I'm not mistaking, the minification step doesn't run on css imported through javascript, so this would be a way to still get autoprefixing and import your css in your JS files.

mitar:
Also, I need prefixing during development so that I can test things across browsers.

tmeasday:
@mitar what does this do by default? (For files with `fileOptions = {}`?)

tmeasday:
> so this would be a way to still get autoprefixing and import your css in your JS files.

@sebakerckhof why do you think this code would run over CSS imported from JS?

mitar:
> @mitar what does this do by default? (For files with `fileOptions = {}`?)

There is a link to [that](https://github.com/ai/browserslist#queries) in the description of the pull request. ;-)

By default it will automatically prefix anything which has to be prefixed for the following browsers: `> 1%, last 2 versions, Firefox ESR`.

tmeasday:
This is a fairly large change for the default behaviour. Do you think that may cause problems for people?

mitar:
Doesn't prefixing just fix things, not break things? Can it break anything?

But I can also make it so that it is disabled for now by default and you have to enable it. And then once #94 is resolved we can change that.

tmeasday:
Yeah, do you think I'm being too conservative? It's hard to know if it'll be a problem for people beyond larger file sizes.

mitar:
Fine with me. I am using packages-based architecture anyway for my apps. :-)

I think we can be conservative now and work on options. But I would not like that options are blocking this.

tmeasday:
Ok, let me know and I'll merge it.

mitar:
BTW, it seems it also uses [environment variables](https://github.com/ai/browserslist#queries) to configure things. And `browserslist` config file. It would be interesting to see where does it search for.

sebakerckhof:
> @sebakerckhof why do you think this code would run over CSS imported from JS?
Well, not imported css, but at least imported stylus files. If you'd do the same thing with a preprocessor that accepts normal css syntax (like sass or less) you could just rename your `css` files to `scss` or `less` to get them autoprefixed, even when importing through JS.

tmeasday:
@mitar let me know when you've changed the default.

mitar:
Yes, busy with some other stuff at the moment.

mitar:
@tmeasday: I changed it so that by default autoprefixer is disabled.

tmeasday:
Can we not run autoprefixer at all if the option is undefined?

mitar:
That is an option as well. :-)

mitar:
Updated.

tmeasday:
Thanks for bearing with me here @mitar

mjmasn:
@mitar so for app-level stylus files is there a way to enable autoprefixer? Had to move from `mquandalle:stylus` to `stylus` for Meteor 1.6 but just realising prefixing is now broken.

PR #7744, opened 2016-09-01
Expose fields automatically published for the logged-in user

mitar:
Addresses #7743.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -696,6 +696,14 @@ Ap._initAccountDataHooks = function () {
   });
 };
 
+Ap._publicationFields = function () {
+  return {
+    profile: 1,
+    username: 1,
+    emails: 1
+  };
+};
+
 Ap._initServerPublications = function () {
   var accounts = this;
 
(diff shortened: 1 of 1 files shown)


laosb:
Should we make it a public API instead of an underscore-prefixed *private* API?

stubailo:
I think an underscored name is fine for this. We can always add a non-underscored version later.

stubailo:
@mitar is this intended to be the _default_ set of published fields? Or should it represent the current set, if it's modified?

mitar:
Current set. The goal is that you can override the prototype to configure the current set. Which is then published. So you should override the prototype before publish happens. (Hm, now I am not 100% sure that you can really do that in practice.)

stubailo:
Hmm, is overriding methods the right API for this kind of thing? I guess that's how overriding stuff like the Accounts URLs works.

Is the publication inside Meteor.startup at least?

mitar:
It is inside [constructor](https://github.com/peerlibrary/meteor/blob/ff66df982ec6f4275bcd237736843d0288c4e7bb/packages/accounts-base/accounts_server.js#L35). I am not sure where is that called from though.

stubailo:
Looks like it's called synchronously, so you probably wouldn't be able to override this from app code: https://github.com/meteor/meteor/blob/be986fd70926c9dd8eff6d8866205f236c8562c4/packages/accounts-base/server_main.js#L9

stubailo:
Let's close this PR and talk on the issue?

PR #7753, opened 2016-09-06
Move connect.compress() after rawConnectHandlers

### LINKED ISSUE #7754: connect.compress() runs before rawConnectHandlers, causing problems
Meteor 1.4.1.1

Currently `WebApp` runs `connect.compress()` before `rawConnectHandlers` in the middleware stack.  I'm trying to use `webpack-hot-middleware` (and `webpack-dev-middleware` in `rawConnectHandlers` and the compression breaks it; moving `connect.compress()` after `rawConnectHandlers` fixes the problem.  Compression might interfere with other things one tries to do in `rawConnectHandlers`.

In any case `rawConnectHandlers` isn't truly "raw" unless it comes before *everything* else.

jedwards1211:
Fix #7754

Currently `WebApp` runs `connect.compress()` before `rawConnectHandlers` in the middleware stack.  I'm trying to use `webpack-hot-middleware` (and `webpack-dev-middleware` in `rawConnectHandlers` and the compression breaks it; moving `connect.compress()` after `rawConnectHandlers` fixes the problem.  Compression might interfere with other things one tries to do in `rawConnectHandlers`.

In any case `rawConnectHandlers` isn't truly "raw" unless it comes before *everything* else.


zol:
Thanks @jedwards1211 , this looks reasonable.

jedwards1211:
Great, thanks!

barbatus:
Ok now it breaks server side rendering in FlowRouter, since there is no way now to override `res.write` which gets now a compressed buffer instead of string. I think that was the reason why `connect.compress()` had been put before `rawConnectHandlers `.

jedwards1211:
@barbatus I don't use FlowRouter so I'm not following exactly, do you mean FlowRouter also monkeypatches `res.write` like `connect.compress()` does?
To me this is a great example of the pitfalls of convention over configuration, without an option to completely override the conventions.

barbatus:
@jedwards1211 Yes, it does. As now FlowRouter monkeypatches `res.write`, then `connect.compress()` monkeypatches `res.write`, so eventually FlowRouter's `res.write` gets a compressed buffer.

jedwards1211:
@barbatus I'm a bit confused, because the latest version of `compression` calls the `write` that it monkeypatched: https://github.com/expressjs/compression/blob/master/index.js#L205.  Not sure about the old version Meteor is using though...

jedwards1211:
Why does FlowRouter need to monkeypatch `res.write` anyway?

barbatus:
For the server side rendering. For example, If one wanted to pre-render a React component for a particular route and then dump it to the response. So FlowRouter could be helpful with that.

jedwards1211:
@barbatus No, I mean is it really necessary to monkeypatch `res.write` to do SSR?  Is there not a simpler way to do it?  I searched through the code in https://github.com/kadirahq/flow-router/tree/master/server and didn't find anything monkeypatching `res.write`.  Can you show a link to the source that does that?

barbatus:
You'll is a SSR version https://github.com/kadirahq/flow-router/tree/ssr
It monkeypatches res.write particularly here https://github.com/kadirahq/flow-router/blob/ssr/server/route.js#L89

jedwards1211:
I would encourage you guys to either
- make a PR to put a `postCompressHandlers` section into the webapp package where `rawConnectHandlers` used to be or
- work with MDG to find a less fragile solution, some possibilities:
  - an API to put a `bodyGenerator` hook in  `BoilerplateGenerator` instead of doing text search-and-replace on the `<body>` tag in monkeypatched `res.write`.
  - an `generateHead` API in `BoilerplateGenerator` that FlowRouter can call; then FlowRouter could take control of rendering the whole page but still delegate to `BoilerplateGenerator` to render the head.

jedwards1211:
If it's decided to revert this PR though, then I could make a different PR to conditionally disable using `connect.compress()` with an environment variable.

jedwards1211:
So shall I make a PR to revert this and add a `preCompressHandlers` section?

jedwards1211:
@enricofoschi this is the PR

comus:
hi, it breaks server side rendering in `react router ssr` also.
the same, `react router ssr` use `fast-render`, `fast-render` use `Picker` and `InjectData` to preload some subscriptions, directly write to chunk in res.write

the main point is `picker`, it depands on `WebApp.rawConnectHandlers`
https://github.com/meteorhacks/picker/blob/master/lib/instance.js
p.s. may be change it to `WebApp.connectHandlers` ?

jedwards1211:
@comus personally I think monkeypatching `res.write` for things like ssr is a code smell, since I can imagine other ways of doing the ssr that would not have been broken by this.  But obviously we should probably revert this since it violated semver.  In any case, I would strongly encourage you guys to consider alternatives to monkeypatching `res.write`.

enricofoschi:
@jedwards1211 thanks for the quick follow up on this - appreciate it. Afaik monkey patching res.write is the only way to actually inject pre-rendered html on the page. I know this may be out of topic for this discussion, but I'd be interested in knowing alternatives.

jedwards1211:
@enricofoschi I know that's the case now, but I really would rather see MDG add an API to the `BoilerplateGenerator` to allow other packages to inject pre-rendered html.

zol:
We merged a principled way to modify the `head` or `body` tags exactly for this kind of use case here: https://github.com/meteor/meteor/pull/3860

I'd suggest sending a patch to the author of code that monkey patches `res.write` to use `dynamicHead`/`dynamicBody` instead.

jedwards1211:
@zol fantastic!  I don't personally use FlowRouter though so it's not something I want to spend time on -- I was just expressing my opinion to FlowRouter SSR devs that I think that approach will serve their needs better.

jedwards1211:
@barbatus @arunoda as @zol mentioned above, there is now a principled way to modify `head` and `body` tags in SSR that doesn't involve monkeypatching `res.write`.

PR #7774, opened 2016-09-15
Add ws: and wss: sources to default CSP for Cordova

### LINKED ISSUE #7772: iOS 10 blocks websocket connections
- [Update -- Fix](#update----fix)
  - [meteor v1.4](#meteor-v14)
    - [Wait](#wait)
    - [Immediate](#immediate)
  - [meteor < v1.4](#meteor--v14)
- [Problem](#problem)

# Update -- Fix

### Wait

You can wait until #7774 is merged and a new version of 

https://atmospherejs.com/meteor/boilerplate-generator

is published (currently at `1.0.9`), and then do:

```
$ meteor update boilerplate-generator
$ deploy
```

### Immediate

```
git clone git@github.com:gsaynac/meteor.git
cd meteor
git checkout cordova-csp-fix
cp -r packages/boilerplate-generator ~/myapp/packages/
cd ~/myapp
deploy
```

## meteor < v1.4

```
git clone git@github.com:gsaynac/meteor.git
cd meteor
git checkout cordova-csp-fix
cp -r packages/boilerplate-generator ~/myapp/packages/
cd ~/myapp
meteor
```

Now you'll get an error like this one:

```
=> Errors prevented startup:                  
   
   error: No version of boilerplate-generator satisfies all constraints: @=1.0.9, @=1.0.8
   Constraints on package "boilerplate-generator":
   * boilerplate-generator@=1.0.9 <- top level
   * boilerplate-generator@=1.0.8 <- top level
```

This is when running meteor `v1.3.2.3`, but you may see a version on the bottom line other than `1.0.8`, in which case use that in this next step:

- edit `~/myapp/packages/boilerplate-generator/package.js`, changing `1.0.9` to `1.0.8` (or the version you see from above)
- rerun `meteor`, making sure there no longer is a startup error
- deploy

# Problem

It appears that WKWebView in iOS 10 no longer applies the Meteor Cordova CSP (`default-src *`) to websockets, and blocks the connection, with the below message. This breaks production apps that make websocket connections (perhaps only connections other than the default connection to the Meteor server? not sure) when their users update to iOS 10 (which has been released).

> Refused to connect to wss://my.domain.tld/sockjs/015/sdflkj/websocket because it appears in neither the connect-src directive nor the default-src directive of the Content Security Policy.
> SecurityError (DOM Exception 18): The operation is insecure

I think this is the CSP that is currently used in Cordova apps:  

```
<meta http-equiv="Content-Security-Policy" content="default-src * data: blob: 'unsafe-inline' 'unsafe-eval';">
```
- https://github.com/meteor/meteor/blob/master/packages/boilerplate-generator/boilerplate_web.cordova.html#L7
- https://guide.meteor.com/mobile.html#csp

Replacing it in the generated html file in XCode with the below [worked for one person in the forums](https://forums.meteor.com/t/ios-10-compatibility/26065/33) (and it looks like for a couple people earlier in the thread as well):

```
<meta http-equiv="Content-Security-Policy" content="default-src * 'self' data: blob: 'unsafe-inline' 'unsafe-eval' wss://*.ourdomain.tld ">
```
(which has an added `wss://domain` at the end, and a `'self'`, if that matters)

If `wss://*` works, that would be nice, as it would match the default `browser-policy` CSP. 

> Add browser-policy to your app to enable a starter policy. With this starter policy, your app's client code will be able to load content (images, scripts, fonts, etc.) only from its own origin, except that XMLHttpRequests and WebSocket connections can go to any origin.
https://atmospherejs.com/meteor/browser-policy

@martijnwalraven

gsaynac:
This PR should fix #7772 by adding ws: and wss: to the default CSP for Cordova.

### DIFF
--- a/packages/boilerplate-generator/boilerplate_web.cordova.html
+++ b/packages/boilerplate-generator/boilerplate_web.cordova.html
@@ -4,7 +4,7 @@
   <meta name="format-detection" content="telephone=no">
   <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width, height=device-height">
   <meta name="msapplication-tap-highlight" content="no">
-  <meta http-equiv="Content-Security-Policy" content="default-src * data: blob: 'unsafe-inline' 'unsafe-eval';">
+  <meta http-equiv="Content-Security-Policy" content="default-src * data: blob: 'unsafe-inline' 'unsafe-eval' ws: wss:;">
 
 {{! We are explicitly not using bundledJsCssUrlRewriteHook: in cordova we serve assets up directly from disk, so rewriting the URL does not make sense }}
 


lorensr:
This fix is working for me (added updated `packages/boilerplate-generator/` to my project and deployed).

benjamn:
Just published version 1.0.10 of the `boilerplate-generator` package, with this change included. If you're using Meteor 1.4, you can update immediately by running `meteor update boilerplate-generator`.

sandeepmirpuri:
I tried updating using meteor update boilerplate-generator

It doesn't work. Stuck on 1.0.9. Any solution?

PR #7781, opened 2016-09-16
Add crossorigin="anonymous" to script tags

brianlukoff:
If you serve your JS through a CDN, then most errors will report out to Kadira (or a similar tracking tool) as "Script error" because the JS file is hosted on another domain. Adding crossorigin="anonymous" to each script tag enables CORS (as long as the CDN is properly configured and the Meteor app also sets the Access-Control-Allow-Origin headers on JS files).

### DIFF
--- a/packages/boilerplate-generator/boilerplate_web.browser.html
+++ b/packages/boilerplate-generator/boilerplate_web.browser.html
@@ -12,7 +12,7 @@
 {{else}}
 <script type='text/javascript' src='{{rootUrlPathPrefix}}/meteor_runtime_config.js'></script>
 {{/if}}
-{{#each js}}  <script type="text/javascript" src="{{../bundledJsCssUrlRewriteHook url}}"></script>
+{{#each js}}  <script type="text/javascript" crossorigin="anonymous" src="{{../bundledJsCssUrlRewriteHook url}}"></script>
 {{/each}}
 {{#each additionalStaticJs}}
   {{#if ../inlineScriptsAllowed}}


dnish:
Yes, think this is needed, my Kadira logs are full of:

```
Error: Script error.

Error:
    at window.onerror (<anonymous>:0:0)

``` 

Seems that it happens because we server all assets via CDN.

egoldblum:
Definitely needed.  I've wanted this for awhile:   #4783

abernix:
Hmm... I can understand the need for this but I think setting this would implicitly enable CORS for that resource, thus _requiring_ the `Access-Control-Allow-Origin` header be set for the host on the other end which is serving that resource.

I believe the result would be that if the remote host didn't set this header, the script wouldn't be executed at all because of the conflicting policy.  While setting `Access-Control-Allow-Origin` might be common, it isn't set by default in on a variety of hosts and it seems this could potentially break things for a lot of users who didn't have that set on their CDN/private-host/etc.

https://html.spec.whatwg.org/multipage/infrastructure.html#cors-settings-attribute says:
> `crossorigin=anonymous`: Requests for the element will have their mode set to "cors" and their credentials mode set to "same-origin".

and

> The missing value default, used when the attribute is omitted, is the No CORS state.

Thoughts?

dnish:
Meteor has a CDN function:


    WebAppInternals.setBundledJsCssPrefix()

Maybe we can check for it and if it is set we add the anonymous prop.

abernix:
@dnish The `Access-Control-Allow-Origin` which needs to be set has to be set on the CDN, not on Meteor.   That was my concern with it breaking things, as this would be out of Meteor's control.

@brianlukoff Is there a way to "serve your JS through a CDN" without using `setBundledJsCssPrefix`?  Are there other uses of this method for non-CDN reasons? (I don't believe so).  If not, then only setting `crossorigin="anonymous"` when `setBundledJsCssPrefix` is set seems like a good idea for this PR in general, no?

Regardless, if my understanding of this attribute is correct, this PR would still break execution of JS for current users of `setBundledJsCssPrefix` users who don't set `Access-Control-Allow-Origin` on their CDN.  As to what the probability of that situation is, I'm not sure.

dnish:
@abernix We are using CloudFlare via cdn.ourapp.com and setting the Access-Control-Allow-Origin header via Meteor WebApp. CloudFlare caches the header.

abernix:
@dnish Ok, that seems like it would work.  I don't think that "caching-what-you-send" behavior is identical across all CDNs though – I presume it is configurable?

brianlukoff:
What if we add an additional (optional) parameter to `setBundledJsCssPrefix` and then only add `crossorigin="anonymous"` if that option is set -- would that work? I think this is only needed for folks that host their JS on a CDN, and anyone doing that would presumably already be using `setBundledJsCssPrefix`.

abernix:
@brianlukoff That would solve the issue of breaking compatibility since it would essentially become opt-in and the documentation for that option could be clear about the need to properly set the header on the CDN.

Maybe `setBundledJsCssPrefix({ crossOrigin: "anonymous" })` would be appropriate?  This could also allow the user to use [the `use-credentials` configuration](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes), if they so desired.

And the default could remain the same as it is right now?

abernix:
@brianlukoff This sounds more reasonable (and safe) to me after our last notes, but let me know your thoughts.  This will definitely need tests written for it before it has a chance to be merged though.  Is that something you can take care of sometime this week?  Lastly, before merging, we'll need to open an issue or PR with `docs` in `vNEXT` branch so it gets documented.

brianlukoff:
Thanks @abernix - this sounds like a good plan. I may not have time to work on it in the near future, but I'll take a stab as soon as I have the bandwidth.

tmeasday:
I'm not sure I properly understand the original issue here (I've not seen this problem before). Can we be sure to document it a little more (why is the kadira error happening?) when opening the issue?

egoldblum:
@tmeasday Please see the linked issue #4783 for an introduction about why this error occurs

tmeasday:
Thanks, @egoldblum I missed that. So re-reading @brianlukoff's original comment, he is saying the same thing ("the error I see in Kadira is not very informative, because CORS"). That makes sense. Thanks.

egoldblum:
Yep, the browser is preventing error details from leaking cross origin (i.e. your CDN is origin A, your app is server from origin B).  You can tell the browser to allow this through a combination of:

1) Access-Control-Allow-Origin response header from CDN (origin A)
2) setting the `crossorigin` attribute on the `script` tag that loads the meteor bundle (origin B)

2 is where Meteor currently does not provide a configuration hook.

abernix:
@brianlukoff Just a gentle ping to see if you'll have any time to work on this soon?  If not, I'll try to set aside some time to move it forward but would appreciate any help you could provide, especially as you're probably more able to test the results.

brianlukoff:
I'm not going to be able to work on this until after the new year at the earliest, unfortunately. The change to the template is simple (just that one attribute), but I wasn't able quickly figure out the right way to pass that option up the chain.

benjamn:
Closing now due to inactivity, but feel free to reopen if you find more time to work on this!

nadeemja:
Hi! 

I really hope this can be re-opened. We have a lot of users getting blank pages and at the same time our logs are filling up with useless errors.

abernix:
@nadeemja This is waiting for a member of the community to revive it right now.  I'm not sure if @brianlukoff will be able to do this currently, so I've gone ahead and marked [the original issue](https://github.com/meteor/meteor/issues/4783) as `pull-requests-encouraged`.

If you're interested in making a contribution to the Meteor Open-Source project, this would be a great way to do it!  I think most, if not all of the requirements are listed above in this PR, so you might want to review that and see if it's something you could tackle.

nadeemja:
Thanks for chiming in, @abernix. I'd love to contribute, but I haven't the faintest idea where to look for relevant code. Where would I even start to make changes? Honestly, most of the above discussion if flying straight over my head. The developer guide is too general, I feel, in this case. I'm willing to play around if I got some direction. I looked at the commit, and what files where changed, but that was only for generating boilerplate it seems. What if you already have a codebase?

abernix:
@nadeemja The file's changed in the commits within this PR are a good starting point, but the term "boilerplate" might be misleading you.  This boilerplate is the static HTML that is served at the first-load of your application and generated on each initial page load, not "boilerplate" in the sense that's generated once when the application is created.

The Developer guide is general, but it should help you get up and running and then you can likely playing around with the code and see how the various changes affect the output.

Issue #7538, opened 2016-08-01
Documentation about pollingIntervalMs and pollingThrottleMs is unclear


ignl:
I am not sure if documentation is correct for [Collection-find](http://docs.meteor.com/api/collections.html#Mongo-Collection-find). In this [blog post](http://info.meteor.com/blog/tuning-meteor-mongo-livedata-for-scalability) settings for Livedata optimization are called pollingInterval and pollingThrottle. Also I am not sure if default for pollingIntervalMs is really 10s? I tried various combinations but couldn't get it to work on meteor 1.4.0.1. Currently I try like this serverside:

`messages.find({'metadata.thread': threadId}, {sort: {'date' : sort}, limit: limit, pollingThrottle: 12000, pollingInterval: 12000});`

I expect a new message to be pushed to other client every 12s, but it seems meteor just updates them immediately anyway. Is this a correct way to do it?

laosb:
This was not the interval of sending messages to client. It's about the interval of the server's polling database. Also keep in mind that if you modify data on server instead of DB, changes happened immediately, too.

ignl:
Ok you closed it but is the documentation really correct? Is it really called pollingIntervalMs and pollingThrottleMs? If default for pollingInterval is 10s should it be set in ms or s (if I would like to poll db every every 20s should I set 20000 or 20)? And about pollingThrottle from documentation:

> Minimum time to allow between re-polling. Increasing this will save CPU and mongo load at the expense of slower updates to users. Decreasing this is not recommended. In milliseconds. Defaults to 50 milliseconds.

This is exactly what I want. If it doesn't work like that then I would say documentation is really misleading which is exactly what I wrote in title.

laosb:
*slower updates to users* there means slower direct DB update to users, not slower serverside updates to users, I think. The documentation is really unclear about this. Maybe add a link to the blog post you've mentioned? I'll reopen for other's thoughts.

loredanacirstea:
@ignl 

Docs are correct, with the exception of those 10sec (should be 10 ms).

1. Be sure that you are only updating MongoDB and not Minimongo - for example, if you update through Meteor methods,  be sure you don't have client stubs.
2. Try this:

```
messages.find({'metadata.thread': threadId}, 
  {
    sort: {'date' : sort}, 
    limit: limit,
    disableOplog: true, 
    pollingThrottleMs: 12000, 
    pollingIntervalMs: 12000
  }
);
```

[...]

PR #7800, opened 2016-09-21
[docs] Clarify JSDoc for pollingIntervalMs and pollingThrottleMs options

abernix:
This change changes units to be consistent and more clearly indicates when these parameters come into play.

Addresses concerns in meteor/meteor#7538

Fixes meteor/meteor#7538


zol:
Thanks @abernix !

Issue #7786, opened 2016-09-19
tools/static-assets/skel/.gitignore should end with a new line(\n)


oyyq99999:
As the title says.

abernix:
Opened #7801.  From a sysadmin standpoint, I understand the reasoning behind this.  Per my comments in the commit:

> Even though I think the fight is a bit futile, it is considered best practice to have newlines at the end of all files.  At the request of meteor/meteor#7786, this commit adds newlines to the end of the new app skeleton `.gitignore` file.  These skel files are used when generating new meteor apps so this avoids perpetuation of this problem into futrue generations. ;)

I changed a few other `.gitignore` files that were missing it too.  This will be closed when the PR is merged.  Thanks for pointing it out!

PR #7801, opened 2016-09-21
Add Newlines to end of .gitignore files, specifically the new app skel

abernix:
Even though I think the fight is a bit futile, at the request of meteor/meteor#7786 this adds newlines to the end of the new app skeleton .gitignore file.  The skel files are used when generating new meteor apps so this avoids perpetuation of this problem into futrue generations. ;)

I also took the liberty of fixing the few other Meteor files which were lacking newlines (all of which were .gitignore)

Closes meteor/meteor#7786


zol:
Merged, thank you!

PR #7811, opened 2016-09-23
Remove need for registry http get check by tuning npm retry parameters

sebakerckhof:
This PR follows out of the discussion of https://github.com/meteor/meteor/pull/7637

Basically Meteor did a check to see if the npm registry url is available because NPM only times out after more than a minute by default.

However the NPM retry parameters are configurable and by tuning them we can remove this check.
This has a couple of advantages:
1. It works for registries that require HTTP get authentication, thus fixing https://github.com/meteor/meteor/issues/7636
2. It makes builds a bit faster since the ensureConnected check can be called quite a lot (~25 times in @abernix  test environment, see: https://github.com/meteor/meteor/pull/7637#issuecomment-249105535 ). 
3. We end up with less code to maintain.
4. More flexible behavior. Currently it will retry once, which in my tests throws an error after about 20s for an unreachable host. You could disable retries altogether or allow for more retries.

### DIFF
--- a/tools/isobuild/meteor-npm-userconfig
+++ b/tools/isobuild/meteor-npm-userconfig
@@ -6,3 +6,10 @@
 
 loglevel = error
 progress = false
+
+# By default if the NPM registry is unavailable, it will timeout after more than 60s.
+# These settings make sure we throw an error faster
+fetch-retries = 1
+fetch-retry-factor = 4
+fetch-retry-mintimeout = 1000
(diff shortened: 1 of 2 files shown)


abernix:
Sorry, but let's continue the conversation in the other PR so we can keep the conversation in one easy-to-follow location.  The PR can be updated once MDG chimes in on what they want to happen here.  (Really, the discussion should have happened in the original issue anyhow.)   Closing this.

Issue #5666, opened 2015-11-17
meteor build should still output something even when not inside a terminal


mitar:
`meteor build` should still output something occasionally even when not running inside a terminal. The problem is when building a Docker image on Docker Hub that the build timeouts and is terminated prematurely.

As an example, if I do `meteor build . | cat` nothing is output.

zol:
Hey @mitar - I don't understand what this feature request is about. If you can add more color including a step by step description of how the change will behave differently from the current implementation, this would help us evaluate whether this is something we want to add to the tool.

mitar:
So currently, if I run `meteor build` inside a Docker container the stdout does not have a terminal attached to it, so `meteor build` does not output anything to the stdout, it just does it job.

This is a problem if `meteor build` takes a lot of time. Some cloud providers which are building Docker images assume that if there is no activity on the standard output for some time that the build process hanged and they kill it.

So, I would like that `meteor build` outputs at least something (maybe a dot every minute) even when it is not attached to a terminal.

AlbinoGeek:
The issue:

- ` meteor build ` has no output.

Why does this matter?

- Some things (such as Docker remote building, and Amazon Build Farms) will automatically terminate the builder if it has not produced output for the LAST `TIMEOUT` TIME.

Solution:

- Output something (even dots) during build to prevent autotermination.

avital:
Why not just run your own command that runs `meteor build` and also prints dots in parallel?

mitar:
Because they if `meteor build` crashes, the command in parallel have to detect that and also stop, or something. So in fact I have then to add some process monitoring script or something. It feels a bit too complicated.

mitar:
I am trying currently `METEOR_PROGRESS_DEBUG=1 meteor build`.

spasarok:
What was the solution to this issue? My Meteor build is getting hung up on some step, and because there's no logging I don't know what step is causing the problem and I eventually get a timeout.

mitar:
#8037 got merged. So after it is released, you will be able to just do `export METEOR_HEADLESS=true` and it will work. But for build itself you can just do `meteor build --headless .`. Example: https://github.com/tozd/docker-meteor/blob/master/Dockerfile#L21

hwillson:
Since #7814 and #8037 were both merged, I'll close this issue. Thanks!

Issue #5208, opened 2015-09-18
Option to disable build tool spinner


zol:
It would be great if we allowed folks to disable the spinner in the build tool as it generates a heap of spurious logging inside automation tools such as Travis-CI.

justinsb:
`METEOR_PRETTY_OUTPUT=0` is supposed to do this
https://github.com/meteor/meteor/blob/devel/tools/console/console.js#L76

zol:
Thanks @justinsb - that did the trick.

PR #7814, opened 2016-09-23
Add --headless option to the build command

engelgabriel:
Building logs files on Travis-CI are getting over 4mb and 80% are spinners.
There's no point in spinning the spinner when we're running automated builds.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -787,6 +787,10 @@ var buildCommands = {
     server: { type: String },
     // XXX COMPAT WITH 0.9.2.2
     "mobile-port": { type: String },
+    // Indicates whether these build is running headless, e.g. in a
+    // continuous integration building environment, where visual niceties
+    // like progress bars and spinners are unimportant.
+    headless: { type: Boolean },
     verbose: { type: Boolean, short: "v" },
     'allow-incompatible-update': { type: Boolean }
   },
@@ -818,6 +822,11 @@ main.registerCommand(_.extend({ name: 'bundle', hidden: true
 
 var buildCommand = function (options) {
   Console.setVerbose(!!options.verbose);
+  if (options.headless) {
+    // There's no point in spinning the spinner when we're running
+    // automated builds.
+    Console.setHeadless(true);
+  }
   // XXX output, to stderr, the name of the file written to (for human
   // comfort, especially since we might change the name)
 


lucfranken:
@engelgabriel Did you see this: https://github.com/meteor/meteor/issues/5208#issuecomment-141579781 ? It seems to fix your issue already, is that correct?

mitar:
But the downside is that then one gets build timeouts because there is not enough output for too long sometimes. So in fact I am running my build processes with `METEOR_PROGRESS_DEBUG=1` to assure I am getting output all the time.

So something in the middle would be great. No spinners, but at least a dot every minute or so.

benjamn:
The way we solve @mitar's problem (which is a real problem!) for our CircleCI tests is to print [`[ProgressDisplay keepalive]`](https://github.com/meteor/meteor/blob/dce2b20ddbe45da48c37f26813fce8d0c72d8c88/tools/console/console.js#L396) every five minutes if there has been no other output. The exact message doesn't matter of course, but could something like that work here?

mitar:
So is this issues then related: #5666?

So METEOR_PRETTY_OUTPUT=0 then does print something out every 5 minutes?

abernix:
@mitar Hmm, I think that will print out right now in all idle situations when the progress display is actually enabled, but I think progress display is only enabled [under _these_ circumstances](https://github.com/meteor/meteor/blob/devel/tools/console/console.js#L1235-L1250) – seemingly not at all when running without a TTY.  I think that the CIs do emulate a TTY, but not sure about your Docker (Or any Docker?)

So to (kinda) answer your question: Yes, but your milage may vary / I don't know. :grin:

mitar:
Docker does not do TTY. I think we should maybe split this into multiple options. One where you are getting this hearbeat, and another where you can quiet the progress bar. So then I could enable heartbeat, but disable progress bar. And without some strange detection for TTY and things like that (that can be used for defaults).

engelgabriel:
Before this gets merged, we had to use `METEOR_PRETTY_OUTPUT=0` in combination with `travis_retry travis_wait 15 meteor build /tmp/build` so it is far from ideal :(

Is there any downside of accepting this PR?

abernix:
IMHO, LGTM.  Provides consistency with `self-test`, `test-packages` and `test` which already provide this option.

@mitar as to your point, possibly (I don't run under those circumstances, and I don't think I'd want a progress bar without a TTY), but it seems like it would be a separate option and separate PR, no?  Doing anything different with `--headless` here would violate the way those are already setup.  In the future, if that's the way to go, there'd just be a single place to [update that anyway](https://github.com/meteor/meteor/blob/devel/tools/console/console.js#L389-L393).

Issue #7794, opened 2016-09-21
Have different timeouts for expiration of password reset and enrollment


mitar:
#7534 introduced token expiration, but one could want a different expiration for password reset and enrollment (or even disable expiration for enrollment).

We could address this by adding a reason for token to the database, and then use different expiration based on that reason.

tmeasday:
I think this is a reasonable approach. So what would the key look like?

mitar:
`reason: 'reset'` or `reason: 'enroll'`

tmeasday:
So the token would look like:

```
  var tokenRecord = {
    token: token,
    email: email,
    when: when,
    reason: 'reset' | 'enrol'
  };
```

That makes sense to me. And we would add a separate option for setting the two timeouts and add two intervals to do the checks?

mitar:
Yea.

tmeasday:
Ok, I'll mark PR-encouraged then.

alphanso:
@tmeasday I have created a PR for this issue. Please check.

mitar:
This has been resolved.

zarvox:
The patch which purported to fix this issue appears to have a showstopper bug which causes enrollment tokens to be cleaned up under the reset token expiry rules.  The net effect is that under the default settings, enrollment tokens are still destroyed after 3 days, rather than after 30 days, and enrollment tokens can't stay in the database longer than password reset tokens.

In particular, when processing password reset token expiry at https://github.com/meteor/meteor/pull/7817/files#diff-2ae5b6c36ab4132c1a2d0e33b5fd8443R1158 a `tokenFilter` with an `$or` specifying the appropriate token reason is passed to `expirePasswordToken`, but in https://github.com/meteor/meteor/pull/7817/files#diff-2ae5b6c36ab4132c1a2d0e33b5fd8443R1081 that `$or` clause is overwritten by the one which only looks at the `when` field.  Perhaps the two clauses should be added to an `$and` clause instead?

We're working around this by bumping the validity duration of password reset tokens, but figured you'd want a heads-up that `passwordEnrollTokenExpirationInDays` is basically broken.

Let me know if I should file this as a new issue.

abernix:
@zarvox Thanks for reporting this!  Please do file this as a new issue so it doesn't get lost on this closed issue!  Additionally, if you can offer any help to fix it, that would be much appreciated (a pull request would be awesome!).

phuongvietle:
Is this finally resolved?

[...]

Issue #7734, opened 2016-08-30
Schema validation error thrown during reset token expiration


serkandurusoy:
As carried over from #6462 per request from @tmeasday 

The task that runs for password reset token expirations cause the following error to be thrown from collection2:

```
Exception in setInterval callback: Error: services.password.reset.$or is not allowed by the schema
     at getErrorObject (packages/aldeed_collection2-core/lib/collection2.js:451:1)
     at [object Object].doValidate (packages/aldeed_collection2-core/lib/collection2.js:433:1)
     at [object Object].Mongo.Collection.(anonymous function) [as update] (packages/aldeed_collection2-core/lib/collection2.js:185:1)
     at AccountsServer.Ap._expirePasswordResetTokens (packages/accounts-base/accounts_server.js:1137:14)
     at packages/accounts-base/accounts_server.js:1174:14
     at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
     at packages/meteor/timers.js:6:1
     at runWithEnvironment (packages/meteor/dynamics_nodejs.js:110:1)
```

tmeasday:
This seems like a bug in SS dealing with `$push: { "field": { or: ... } }`

serkandurusoy:
Hmm, I'll cross post over there as well

tarunbatra:
i'm getting the
`Exception in setInterval callback: MongoError: Cannot apply $pull to a non-array value`
error as others have mentioned in #6462 . Any updates on this?

tmeasday:
@tarunbatra What Meteor version are you on? Which version of `npm-mongo` are you using?

tarunbatra:
@tmeasday 
Meteor: `1.4.1.1`
npm-mongo : `1.5.48`

tmeasday:
@tarunbatra apologies, I mean to ask your `accounts-base` version.

tmeasday:
(if it's not `1.2.11` I suggest updating it)

tarunbatra:
@tmeasday yep that was it. Thanx! :+1:

abernix:
SimpleSchema shouldn't really be throwing an error for this, but the code that is causing this does not match a Mongo spec.  It just happens to work currently because [the actual value side of the `$unset` field doesn't matter](https://docs.mongodb.com/manual/reference/operator/update/unset/#up._S_unset).

[...]

PR #7817, opened 2016-09-24
Separate timeouts for expiration of password reset and enrollment

alphanso:
@data Fix for issue #7794

### DIFF
--- a/packages/accounts-base/accounts_common.js
+++ b/packages/accounts-base/accounts_common.js
@@ -97,6 +97,7 @@ export class AccountsCommon {
    * @param {Number} options.loginExpirationInDays The number of days from when a user logs in until their token expires and they are logged out. Defaults to 90. Set to `null` to disable login expiration.
    * @param {String} options.oauthSecretKey When using the `oauth-encryption` package, the 16 byte key using to encrypt sensitive account credentials in the database, encoded in base64.  This option may only be specifed on the server.  See packages/oauth-encryption/README.md for details.
    * @param {Number} options.passwordResetTokenExpirationInDays The number of days from when a link to reset password is sent until token expires and user can't reset password with the link anymore. Defaults to 3.
+   * @param {Number} options.passwordEnrollTokenExpirationInDays The number of days from when a link to set inital password is sent until token expires and user can't set password with the link anymore. Defaults to 30.
    */
   config(options) {
     var self = this;
@@ -128,7 +129,7 @@ export class AccountsCommon {
     }
 
     // validate option keys
-    var VALID_KEYS = ["sendVerificationEmail", "forbidClientAccountCreation",
+    var VALID_KEYS = ["sendVerificationEmail", "forbidClientAccountCreation", "passwordEnrollTokenExpirationInDays",
                       "restrictCreationByEmailDomain", "loginExpirationInDays", "passwordResetTokenExpirationInDays"];
     _.each(_.keys(options), function (key) {
       if (!_.contains(VALID_KEYS, key)) {
@@ -217,6 +218,11 @@ export class AccountsCommon {
             DEFAULT_PASSWORD_RESET_TOKEN_EXPIRATION_DAYS) * 24 * 60 * 60 * 1000;
   }
 
+  _getPasswordEnrollTokenLifetimeMs() {
+    return (this._options.passwordEnrollTokenExpirationInDays ||
+        DEFAULT_PASSWORD_ENROLL_TOKEN_EXPIRATION_DAYS) * 24 * 60 * 60 * 1000;
+  }
+
   _tokenExpiration(when) {
     // We pass when through the Date constructor for backwards compatibility;
     // `when` used to be a number.
@@ -259,6 +265,8 @@ Meteor.user = function () {
 var DEFAULT_LOGIN_EXPIRATION_DAYS = 90;
 // how long (in days) until reset password token expires
 var DEFAULT_PASSWORD_RESET_TOKEN_EXPIRATION_DAYS = 3;
+// how long (in days) until enrol password token expires
+var DEFAULT_PASSWORD_ENROLL_TOKEN_EXPIRATION_DAYS = 30;
 // Clients don't try to auto-login with a token that is going to expire within
 // .1 * DEFAULT_LOGIN_EXPIRATION_DAYS, capped at MIN_TOKEN_LIFETIME_CAP_SECS.
 // Tries to avoid abrupt disconnects from expiring tokens.
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1075,6 +1075,21 @@ Ap._generateStampedLoginToken = function () {
 /// TOKEN EXPIRATION
 ///
 
+function expirePasswordToken(accounts, oldestValidDate, tokenFilter, userId) {
+  var userFilter = userId ? {_id: userId} : {};
+
+  accounts.users.update(_.extend(userFilter, tokenFilter, {
+    $or: [
+      { "services.password.reset.when": { $lt: oldestValidDate } },
+      { "services.password.reset.when": { $lt: +oldestValidDate } }
+    ]
+  }), {
+    $unset: {
+      "services.password.reset": ""
+    }
+  }, { multi: true });
+}
+
 // Deletes expired tokens from the database and closes all open connections
 // associated with these tokens.
 //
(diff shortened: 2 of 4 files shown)


packages/accounts-base/accounts_common.js
```diff
@@ -212,9 +213,14 @@ export class AccountsCommon {
             DEFAULT_LOGIN_EXPIRATION_DAYS) * 24 * 60 * 60 * 1000;
   }
 
-  _getPasswordResetTokenLifetimeMs() {
-   return (this._options.passwordResetTokenExpirationInDays ||
-            DEFAULT_PASSWORD_RESET_TOKEN_EXPIRATION_DAYS) * 24 * 60 * 60 * 1000;
+  _getPasswordResetTokenLifetimeMs(reason) {
+    if(reason === 'enrol') {
```
mitar: Style, space before `(`.

packages/accounts-base/accounts_server.js
```diff
@@ -1135,13 +1135,15 @@ Ap._expirePasswordResetTokens = function (oldestValidDate, userId) {
...
 
   this.users.update(_.extend(userFilter, {
+    "services.password.reset.reason": reason,
     $or: [
       { "services.password.reset.when": { $lt: oldestValidDate } },
       { "services.password.reset.when": { $lt: +oldestValidDate } }
     ]
   }), {
     $unset: {
       "services.password.reset": {
```
mitar: This `$unset` looks to me in generally strange. Is this really a valid MongoDB query? Strange.

alphanso: $unset is not changed in this PR. I have just updated the query. Here is the documentation for $unset
https://docs.mongodb.com/manual/reference/operator/update/unset/

packages/accounts-base/accounts_common.js
alphanso: Fixed

packages/accounts-base/accounts_server.js
mitar: So, I think this is completely bogus. `$unset` takes a field name and does not care about the value of the filedname. That $or there is whatever?

This could be simply simplified to:

```
$unset: {
  "services.password.reset": ""
}
```

alphanso: Fixed

abernix:
🇬🇧/🇦🇺/🇳🇿/🌍/🌏: I _swear_ I wouldn't make this argument unless it was for consistency, **_but_**...

The instance of "enrol" (with one L – not wrong!) should be changed to "enroll" (with two Ls) since Meteor currently uses the North American version of "enroll", "enrolled", and "enrollment" in other places.  The rest of the world uses single-L versions of these words: enrol, enroled, enrolment.

😞

<sub>(Warning: It looks so much better with the extra L :wink:)</sub>

mitar:
I agree about two l. Consistency is useful also for making less bugs.

alphanso:
Its a funny fix but I have fixed it :) :D

packages/accounts-base/accounts_server.js
```diff
@@ -1135,18 +1135,14 @@ Ap._expirePasswordResetTokens = function (oldestValidDate, userId) {
   var userFilter = userId ? {_id: userId} : {};
 
   this.users.update(_.extend(userFilter, {
+    "services.password.reset.reason": reason,
```
mitar: This change will now not expire tokens made before we introduced this new field.

Maybe it would be better if we would simply have one query, which would be something like: remove all tokens if (reason is enroll and older than X) OR (reason is not enroll (including no reason) and older than Y).

packages/accounts-base/accounts_server.js
abernix: This is the cause of the bug reported in #7734.

abernix:
Not saying that this won't land, but if this PR doesn't land for any reason, [this fix above should still be applied in a new PR](https://github.com/meteor/meteor/pull/7817#discussion_r80375759).

packages/accounts-base/accounts_server.js
alphanso: Fixed it and also added the test.

benjamn:
Can we introduce a new `AccountsServer#_expirePasswordEnrollTokens` method instead of changing the parameters of `AccountsServer#_expirePasswordResetTokens`? They can still share most of their implementation through a private helper.

alphanso:
Separated the 2 functions. Please check. :)

packages/accounts-base/accounts_common.js
```diff
@@ -212,9 +213,14 @@ export class AccountsCommon {
             DEFAULT_LOGIN_EXPIRATION_DAYS) * 24 * 60 * 60 * 1000;
   }
 
-  _getPasswordResetTokenLifetimeMs() {
-   return (this._options.passwordResetTokenExpirationInDays ||
-            DEFAULT_PASSWORD_RESET_TOKEN_EXPIRATION_DAYS) * 24 * 60 * 60 * 1000;
+  _getPasswordResetTokenLifetimeMs(reason) {
```
abernix: I don't see a reason not to break this down into two functions as well (`_getPasswordResetTokenLifetimeMs` and `_getPasswordEnrollTokenLifetimeMs`).

The name just isn't accurate otherwise.  (i.e. `_getPasswordResetTokenLifetimeMs` is really `_getPasswordResetOrEnrollTokenLifetimeMs` which is just too long).

packages/accounts-base/accounts_server.js
```diff
@@ -1132,21 +1132,52 @@ Ap._expirePasswordResetTokens = function (oldestValidDate, userId) {
...
+    (new Date(new Date() - tokenLifetimeMs));
+
+  var tokenFilter = {
+    "services.password.reset.reason": "enroll"
+  }
+
+  this._expirePasswordToken(oldestValidDate, tokenFilter, userId);
+}
+
+Ap._expirePasswordToken = function (oldestValidDate, tokenFilter, userId) {
```
abernix: I don't think we need to expose NEW `AccountServer` methods.   This could just be private, right?

packages/accounts-base/accounts_server.js
```diff
@@ -1171,7 +1202,8 @@ Ap.config = function (options) {
 function setExpireTokensInterval(accounts) {
   accounts.expireTokenInterval = Meteor.setInterval(function () {
     accounts._expireTokens();
-    accounts._expirePasswordResetTokens();
+    accounts._expirePasswordResetTokens('reset');
```
abernix: Looks like this didn't get updated; this seems to be passing the wrong param.  I believe it should be:

`accounts._expirePasswordResetTokens();`

?

packages/accounts-base/accounts_server.js
```diff
@@ -1171,7 +1202,8 @@ Ap.config = function (options) {
...
+    accounts._expirePasswordResetTokens('enroll');
```
abernix: Looks like this didn't get updated; this would perform the wrong action.  

`accounts._expirePasswordEnrollTokens();`

Right?

packages/accounts-base/accounts_server.js
abernix: Also, I recommend moving this method above where it's first used (i.e. placed at ~1077)

packages/accounts-base/accounts_common.js
```diff
@@ -97,6 +97,7 @@ export class AccountsCommon {
    * @param {Number} options.loginExpirationInDays The number of days from when a user logs in until their token expires and they are logged out. Defaults to 90. Set to `null` to disable login expiration.
    * @param {String} options.oauthSecretKey When using the `oauth-encryption` package, the 16 byte key using to encrypt sensitive account credentials in the database, encoded in base64.  This option may only be specifed on the server.  See packages/oauth-encryption/README.md for details.
    * @param {Number} options.passwordResetTokenExpirationInDays The number of days from when a link to reset password is sent until token expires and user can't reset password with the link anymore. Defaults to 3.
+   * @param {Number} options.passwordEnrollTokenExpirationInDays The number of days from when a link to set inital password is sent until token expires and user can't set password with the link anymore. Defaults to 30.
```
abernix: :+1:

packages/accounts-base/accounts_common.js
alphanso: Separated the functions.

packages/accounts-base/accounts_server.js
alphanso: Fixed it

packages/accounts-base/accounts_server.js
alphanso: Yup. Fixed.

packages/accounts-base/accounts_server.js
alphanso: Yup. Fixed

alphanso:
Made required changes. Please check.

mitar:
+1

abernix:
It should be pointed out that this implementation could be buggy.  Please see https://github.com/meteor/meteor/issues/7794#issuecomment-270253106 (which will hopefully be opened as a new issue soon).  @alphanso, your thoughts would be appreciated, but the bug report sounds very likely due to the order of the `_.extend` arguments.

alphanso:
I had a quick look. We are using multiple $or operator without using $and operator and I believe one of the $or is silently ignored (http://stackoverflow.com/questions/5150061/mongodb-multiple-or-operations) . I believe instead of using _.extend, we should use $and as suggested by @zarvox

mitar:
#8218 was opened to report this.

Issue #7478, opened 2016-07-26
Proposal to use -s as short form of --settings flag in the Meteor CLI tool


yogiben:
`--settings` is probably the most used flag for the cli tool.

We make developers' lives easier by allowing them to use `-p` in addition to `--port` when setting a port.

The same would be true if we allowed `-s` in addition to `--settings`.

e.g. `meteor -s settings.json`

yogiben:
I'd be happy to submit a PR for this.

alphanso:
@abernix  @platform I have created a PR. Please check.

abernix:
Fixed with PR #7819.  Nice work, @alphanso!  This should work its way into an upcoming version.

PR #7819, opened 2016-09-25
Use -s as short form of --settings flag

alphanso:
Fix issue #7478


alphanso:
Updated the PR to change the `deploy` and `test` commands also.

abernix:
Updated the `help.txt` with the shorthand commands and merged.  Thanks!

Issue #7631, opened 2016-08-11
Meteor should check if is running as root and prevent it on Unix


mitar:
Meteor should have a similar check node.js has and prevent running as root on Unix. That if you do `sudo meteor` it says that it will not run and you have to do `sudo meteor --ireallyknowwhatiamdoing` (see what is the flag node.js use) to allow it. Because I think this is so common issue and there is so little reasons why to run meteor with sudo that it would be very useful.

[From discussion here](https://github.com/meteor/meteor/issues/7469#issuecomment-239073439).

mitar:
We should maybe also document a fix if you do run with `sudo`, that you should chown your `.meteor` directory back to yourself: `chown -Rh <username> ~/.meteor`

mhidou:
Should we check only on `meteor run` or in all commands?

abernix:
I don't see a reason not to do it on all commands.  It seems that the only time root would _actually_ be necessary would be if you were using the `meteor` development tool and wanted to bind to a port <= 1024.  In the case that you needed to do that (still arguably necessary in a world of containers and port forwarding), you could pass the flag to allow it.

But, speaking of flag, I can't seem to find a reference to what that `--ireallyknowwhatiamdoing` Node flag is you're referring to, @mitar. 🔎🔍

mitar:
In npm it is called [unsafe-perm](https://docs.npmjs.com/misc/config#unsafe-perm). So I would do `meteor --unsafe-perm run` to be similar.

I run inside Docker building bundles and on Travis CI test-packages as root. :-) So this might be backwards incompatible change if we do it for all commands, but I would still do it. But we probably have to do it for all, because almost all commands could access packaging information and potentially change permissions of files, no?

Alternatively, we could be just output a warning if you run as root and without this flag.

abernix:
Ah, yes.  I just had to use that on the `abernix:meteord` `mup` image recently.  I'm glad I've forgotten already. :wink:

But yea, this would be a breaking change if it was anything more than a warning.  I think it's pretty important to prevent this sort of behavior though.  I'm kinda in favor.

mitar:
In favor of preventing or of warning?

abernix:
In favor of preventing entirely unless the flag is passed.

mitar:
I agree.

mitar:
This has been resolved.

mitar:
This has been resolved.

PR #7821, opened 2016-09-26
Meteor should check if is running as root and prevent it on Unix

mhidou:
Hi,

This is a solution for #7631

```
$ sudo meteor

It seems that you are running meteor as root, if you really know what you are
doing, add the unsafe-perm flag.

meteor --unsafe-perm <command>

```
```
$ sudo meteor --unsafe-perm

You are running meteor as root, don't forget to chown your .meteor directory
back to yourself.

chown -Rh <username> ~/.meteor

...

```

### DIFF
--- a/tools/cli/main.js
+++ b/tools/cli/main.js
@@ -594,7 +594,7 @@ Fiber(function () {
   // tight timetable for 1.0 and there is no advantage to doing it now
   // rather than later. #ImprovingCrossVersionOptionParsing
 
-  var isBoolean = { "--help": true };
+  var isBoolean = { "--help": true, "--unsafe-perm": true };
   var walkCommands = function (node) {
     _.each(node, function (value, key) {
       if (value instanceof Command) {
@@ -744,6 +744,35 @@ Fiber(function () {
 
     // It is a plain old argument!
     rawArgs.push(term);
+  }
+  
+  // Check if meteor is run as root only on UNIX platforms
+  var unixPlatforms = { 
+    'darwin': true,
+    'linux': true,
+    'freebsd': true
+  };
+  var platform = process.platform;
+
+  if (_.has(unixPlatforms, platform)){
+    var isRoot = process.env.USERNAME == 'root';
+    var isRootSafe = _.has(rawOptions, '--unsafe-perm');
+
+    if(isRoot && !isRootSafe){
+      Console.error("");
+      Console.error("It seems that you are running meteor as root, if you really know what you are doing, add the unsafe-perm flag.");
+      Console.error("");
+      Console.error("meteor --unsafe-perm <command>");
+      Console.error("");
+      process.exit(1);
+    }
+    if (isRoot && isRootSafe) {
+      Console.info("");
+      Console.info("You are running meteor as root, don't forget to chown your .meteor directory back to yourself.");
+      Console.info("");
+      Console.info("chown -Rh <username> ~/.meteor");
+      Console.info("");
+    }
   }
 
   // Figure out if we're running in a directory that is part of a Meteor
--- a/tools/cli/main.js
+++ b/tools/cli/main.js
@@ -747,17 +747,9 @@ Fiber(function () {
   }
 
   // Prevent running meteor as root
-
-  var unixPlatforms = {
-    'darwin': true,
-    'linux': true,
-    'freebsd': true
-  };
-  var platform = process.platform;
-
-  if (_.has(unixPlatforms, platform)){
+  if (process.getuid) {
     // On UNIX platforms.
-    if (process.getgid() === 0 && process.getuid() === 0){
+    if (process.getuid() === 0){
       if (_.has(rawOptions, '--unsafe-perm')) {
         // Meteor is running as root and has the --unsafe-perm flag, just notice.
         Console.info("");
--- a/tools/cli/main.js
+++ b/tools/cli/main.js
@@ -745,10 +745,10 @@ Fiber(function () {
     // It is a plain old argument!
     rawArgs.push(term);
   }
-  
-  // Check if meteor is run as root only on UNIX platforms.
-  
-  var unixPlatforms = { 
+
+  // Prevent running meteor as root
+
+  var unixPlatforms = {
     'darwin': true,
     'linux': true,
     'freebsd': true
@@ -757,48 +757,31 @@ Fiber(function () {
 
   if (_.has(unixPlatforms, platform)){
     // On UNIX platforms.
-    var isSudo = !_.isUndefined(process.env.SUDO_UID);
-    
     if (process.getgid() === 0 && process.getuid() === 0){
       if (_.has(rawOptions, '--unsafe-perm')) {
         // Meteor is running as root and has the --unsafe-perm flag, just notice.
         Console.info("");
-        Console.info("You have run Meteor as root.", 
+        Console.info("You have run Meteor as root.",
         "Your permissions in your app directory will be incorrect if you ever attempt to perform any Meteor tasks as your non-root user.",
         "You probably didn't want this, but you can fix it by running the following from the root of your project:");
         Console.info("");
         Console.info("sudo chown -Rh <username> .meteor/local");
         Console.info("");
+        // Fix issue when running `sudo meteor --unsafe-perm`, --unsafe-perm: unknown option.
+        delete rawOptions['--unsafe-perm'];
       }
       else {
-        // Meteor is running as root without the --unsafe-perm flag.
+        // Meteor is running as root without the --unsafe-perm flag, notice and stop.
         Console.error("");
         Console.error("You are attempting to run Meteor as the \"root\" user.",
-        "If you are developing, this is almost certainly *not* what you want to do and will likely result in incorrect file permissions.",);
+        "If you are developing, this is almost certainly *not* what you want to do and will likely result in incorrect file permissions.",
+        "However, if you are running this in a build process (CI, etc.) or you are absolutely sure you know what you are doing,",
+        "add the `--unsafe-perm` flag to this command to proceed.");
         Console.error("");
-        
-        if (isSudo){
-          // Meteor is running with "sudo", changing gid and uid to remove root access.
-          Console.info("Meteor will now run as " + process.env.SUDO_USER);
-          Console.info("");
-          
-          process.setgid(parseInt(process.env.SUDO_GID));
-          process.setuid(parseInt(process.env.SUDO_UID));
-        }
-        
-        Console.info("However, if you are running this in a build process (CI, etc.) or you are absolutely sure you know what you are doing,", 
-        "add the `--unsafe-perm` flag to this command to force meteor running as root.");
-        Console.info("");
-        
-        if (!isSudo){
-          // Meteor is not running with "sudo", exit.
-          process.exit(1);
-        }
+
+        process.exit(1);
       }
     }
-    
-    // Issue when running `sudo meteor --unsafe-perm`, --unsafe-perm: unknown option.
-    // TODO: remove the --unsafe-perm flag.
   }
 
   // Figure out if we're running in a directory that is part of a Meteor
--- a/tools/cli/main.js
+++ b/tools/cli/main.js
@@ -746,7 +746,7 @@ Fiber(function () {
     rawArgs.push(term);
   }
   
-  // Check if meteor is run as root only on UNIX platforms
+  // Check if meteor is run as root only on UNIX platforms.
   var unixPlatforms = { 
     'darwin': true,
     'linux': true,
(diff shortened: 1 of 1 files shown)


mitar:
I would output the second warning. Because it is not true that one wants to chown the directory back to your self. Or, hm, which `.meteor` does Meteor under sudo use? `/home/user/.meteor` or `/root/.meteor`?

Maybe you could display the second warning if the parent of `.meteor` directory is not the same as the `.meteor` directory (or as current running user).

So in Docker I run `meteor` as root, but not with `sudo`. And `.meteor` is inside `/root/.meteor`. There is no point for the second warning then.

tools/cli/main.js
```diff
@@ -745,6 +745,35 @@ Fiber(function () {
...
+    var isRoot = process.env.USERNAME == 'root';
+    var isRootSafe = _.has(rawOptions, '--unsafe-perm');
+
+    if(isRoot && !isRootSafe){
+      Console.info("");
+      Console.info("It seems that you are running meteor as root, if you really know what you are doing, add the unsafe-perm flag.");
+      Console.info("");
+      Console.info("meteor --unsafe-perm <command>");
+      Console.info("");
+      return;
```
mitar: Could you make sure that the process terminates with non-zero exit code?

tools/cli/main.js
```diff
@@ -745,6 +745,35 @@ Fiber(function () {
...
+    'freebsd': true
+  };
+  var platform = process.platform;
+
+  if (_.has(unixPlatforms, platform)){
+    var isRoot = process.env.USERNAME == 'root';
+    var isRootSafe = _.has(rawOptions, '--unsafe-perm');
+
+    if(isRoot && !isRootSafe){
+      Console.info("");
```
mitar: And this should probably be an error, not info, then, if we are failing.

mhidou:
@mitar I agree with you we have to display the second warning only when the command is run with sudo, it makes no sense if you are logged in as root. 

When I run meteor with sudo I have `process.env.SUDO_USER` containing my username. Unfortunately I've found no documentation about this. Is it safe to check `typeof process.env.SUDO_USER == 'string'` instead of checking .meteor folder location? better performances?

tools/cli/main.js
mhidou: Fixed

tools/cli/main.js
mhidou: Fixed

tools/cli/main.js
```diff
@@ -745,6 +745,35 @@ Fiber(function () {
...
+      Console.error("");
+      Console.error("meteor --unsafe-perm <command>");
+      Console.error("");
+      process.exit(1);
+    }
+    if (isRoot && isRootSafe) {
+      Console.info("");
+      Console.info("You are running meteor as root, don't forget to chown your .meteor directory back to yourself.");
+      Console.info("");
+      Console.info("chown -Rh <username> ~/.meteor");
```
abernix: If run as root (with `sudo` or not) it would still use `$HOME/.meteor` (which in most cases would be `/root`) for `meteor-tool`-related stuff which seems like it should be fine and I don't believe is the source of the problem.

I think the issue arises with the `root`-owned files being written in `<app-dir>/.meteor/local/` which would be in the user's home directory, no?  (Though even that should be fine so long as they ALWAYS run every Meteor command with root, but they probably don't understand what they are doing if they're doing this.).

tools/cli/main.js
```diff
@@ -745,6 +745,35 @@ Fiber(function () {
...
+
+    if(isRoot && !isRootSafe){
+      Console.error("");
+      Console.error("It seems that you are running meteor as root, if you really know what you are doing, add the unsafe-perm flag.");
+      Console.error("");
+      Console.error("meteor --unsafe-perm <command>");
+      Console.error("");
+      process.exit(1);
+    }
+    if (isRoot && isRootSafe) {
```
abernix: Maybe add `&& process.env.LOGNAME !== process.env.USER` here to suppress this when it's irrelevant?

tools/cli/main.js
abernix: This message could probably read:

```
You have run Meteor as root.  Your permissions in your app directory will be incorrect
if you ever attempt to perform any Meteor tasks as your non-root user.  You probably
didn't want this, but you can fix it by running the following from the root of your project:

    sudo chown -Rh ${`process.env.LOGNAME`} .meteor/local
```

tools/cli/main.js
```diff
@@ -745,6 +745,35 @@ Fiber(function () {
...
+
+  if (_.has(unixPlatforms, platform)){
+    var isRoot = process.env.USERNAME == 'root';
+    var isRootSafe = _.has(rawOptions, '--unsafe-perm');
+
+    if(isRoot && !isRootSafe){
+      Console.error("");
+      Console.error("It seems that you are running meteor as root, if you really know what you are doing, add the unsafe-perm flag.");
+      Console.error("");
+      Console.error("meteor --unsafe-perm <command>");
```
mitar: I agree.

tools/cli/main.js
mitar: I am not sure. I have seen issues with `~/.meteor` as well. :-)

mitar: But I agree. This warning could be enough, and even if it is not, it will lead users to the right direction.

tools/cli/main.js
```diff
@@ -745,6 +745,35 @@ Fiber(function () {
...
+  // Check if meteor is run as root only on UNIX platforms
+  var unixPlatforms = { 
+    'darwin': true,
+    'linux': true,
+    'freebsd': true
+  };
+  var platform = process.platform;
+
+  if (_.has(unixPlatforms, platform)){
+    var isRoot = process.env.USERNAME == 'root';
```
benjamn: It seems like the canonical way to do this check is [`process.getuid && process.getuid() === 0`](https://github.com/sindresorhus/is-root/blob/86962be252d45e26a78a085e8ad97e7781f2ce34/index.js#L3). The `process.getuid` function is only defined on POSIX platforms, and 0 is the UID of the root user.

mitar:
I cannot upvote yout "Let's use `process.getuid`, please.". :-(

abernix:
This PR is a step in the right direction (and likely still merge-able without this consideration), but this also raises some other concerns about the way Meteor runs its `npm install` commands internally since currently, if the `meteor` tool was run as `root` (again, not recommended), `--unsafe-perm` wouldn't be used

@mitar, I'm curious if you've experienced issues with this in your Docker builds already (and a bit surprised if you haven't!)

Perhaps once this is implemented, it would be ideal to let this argument through to the NPM installing functions accordingly.

mitar:
NPM changed it so that it allows you to run root, but if you do so, then they complain that now you should probably use `sudo`: https://github.com/npm/npm/commit/7288a137f3ea7fafc9d4e7d0001a8cd044d3a22e

So I do not think we have to pass anything to NPM.

mitar:
I have not had any problems with permissions on Docker or Travis CI, but I do have to run Travis CI with `sudo: required`. https://github.com/arunoda/travis-ci-meteor-packages/blob/master/.travis-sample.yml#L3

mhidou:
What about changing gid and uid when meteor is running with sudo?

```console
$ sudo meteor

You are attempting to run Meteor as the "root" user. If you are developing,
this is almost certainly *not* what you want to do and will likely result in
incorrect file permissions.

Meteor will now run as <username>

However, if you are running this in a build process (CI, etc.) or you are
absolutely sure you know what you are doing, add the `--unsafe-perm` flag to
this command to force meteor running as root.

=> Continues as username

```

```console
$ sudo meteor --unsafe-perm

You have run Meteor as root. Your permissions in your app directory will be
incorrect if you ever attempt to perform any Meteor tasks as your non-root
user. You probably didn't want this, but you can fix it by running the
following from the root of your project:

sudo chown -Rh <username> .meteor/local

=> Continues as root

```

```console
# meteor

You are attempting to run Meteor as the "root" user. If you are developing,
this is almost certainly *not* what you want to do and will likely result in
incorrect file permissions.

However, if you are running this in a build process (CI, etc.) or you are
absolutely sure you know what you are doing, add the `--unsafe-perm` flag to
this command to force meteor running as root.

=> Stop

```

```console
# meteor --unsafe-perm

You have run Meteor as root. Your permissions in your app directory will be
incorrect if you ever attempt to perform any Meteor tasks as your non-root
user. You probably didn't want this, but you can fix it by running the
following from the root of your project:

sudo chown -Rh <username> .meteor/local

=> Continues as root

```

tools/cli/main.js
mhidou: When running with sudo `process.env.LOGNAME` contains `'root'`, `process.env.SUDO_USER` contains the user name but available only when run with sudo. This warning is also shown when logged in as root, better to keep `<username>`.

mitar:
I do not think we should be too smart here. Just display an error and instructions how to proceed. I think this is much better than trying to guess things. I think using uid/gid is better than relaying on environment variables.

mhidou:
@mitar @benjamn @abernix PR updated

tools/cli/main.js
```diff
@@ -746,6 +746,44 @@ Fiber(function () {
...
+  // Prevent running meteor as root
+
+  var unixPlatforms = {
+    'darwin': true,
+    'linux': true,
+    'freebsd': true
+  };
+  var platform = process.platform;
+
+  if (_.has(unixPlatforms, platform)){
```
abernix: This entire `unixPlatforms` check (from here up) can go away and just be replaced with:

```js
if (process.getuid) {
```

It will only be defined on `POSIX` systems.

tools/cli/main.js
```diff
@@ -746,6 +746,44 @@ Fiber(function () {
...
+    // On UNIX platforms.
+    if (process.getgid() === 0 && process.getuid() === 0){
```
abernix: Please just check `process.getuid() === 0`.  There is no need to check `process.getgid()` as well since uid/gid is always 0/0 for `root`.

tools/cli/main.js
```diff
@@ -746,6 +746,44 @@ Fiber(function () {
...
+        // Meteor is running as root and has the --unsafe-perm flag, just notice.
+        Console.info("");
+        Console.info("You have run Meteor as root.",
+        "Your permissions in your app directory will be incorrect if you ever attempt to perform any Meteor tasks as your non-root user.",
+        "You probably didn't want this, but you can fix it by running the following from the root of your project:");
+        Console.info("");
+        Console.info("sudo chown -Rh <username> .meteor/local");
+        Console.info("");
+        // Fix issue when running `sudo meteor --unsafe-perm`, --unsafe-perm: unknown option.
+        delete rawOptions['--unsafe-perm'];
```
abernix: Hmm, this is an interesting behavior.  No idea why?

mhidou: If the `--unsafe-perm` flag is not deleted the command emit an error because this option is not in the command's available options.

tools/cli/main.js
mhidou: Fixed

tools/cli/main.js
mhidou: Fixed

benjamn:
Thanks @mhidou and everyone who provided feedback!

abernix:
Just realized that this has a side effect of not allowing you to run `meteor --version` or `meteor --help` as root.

With `meteor --version` you can't even run it with `--unsafe-perm`, but with `meteor --help` you can.  The `--help` isn't hard to quick-fix, but `--version` could potentially initiate a `meteor-tool` download which might cause file permission problems?  Not sure?  Maybe not?

It's possible that this will be purely an annoyance, but thought I'd put it out there/talk out loud.

mitar:
Thanks for reporting.

rclai:
Guys, why am I experiencing this PR when I haven't even upgraded Meteor? https://github.com/meteor/meteor/issues/7959

benjamn:
@rclai The `tools/cli/main.js` code runs before the code that examines `.meteor/release` in order to switch to the app-specific Meteor version, so this logic applies any time you run `meteor`, regardless of `.meteor/release`.

Running Meteor as `root` has been problematic for a while (at least back to Meteor 1.3), so this warning is valid for other versions, if annoying. The fact that `--unsafe-perm` doesn't work is a bug, though.

I believe you can get around the warning by pointing the `~/.meteor/meteor` symlink to a different version, e.g.:
```
cd ~/.meteor
rm meteor
ln -s packages/meteor-tool/1.4.1_3/mt-os.*/meteor ./
```

skirunman:
Yes, this broke our CI script for Bitbucket Pipelines, which runs in a Docker image as root, and what we use to deploy to Galaxy. We currently run `meteor --version` as well as `meteor npm --version`, `meteor node --version` and `meteor add some-package`.

Will the `meteor npm` commands work with `--unsafe-perm` flag? I assume yes, but have not had time to test.

What about ability to run `meteor add some-package`?

benjamn:
Yes, for what it's worth, `meteor npm` and `meteor node` will work as `root`, because that logic [comes before `tools/cli/main.js`](https://github.com/meteor/meteor/blob/f7271e214a689f5f6ae071faaecf581104ec0bed/tools/index.js#L3-L22).

skirunman:
@benjamn Thanks for the quick response.

I guess I'll change our script and wait for a fix to allow running other meteor commands with `--unsafe-flag`. Also need a fix for https://github.com/meteor/meteor/issues/7956

mitar:
I think we should just make sure that `--unsafe-perm` works for all and any command/argument.

abernix:
@mitar Agree and actually, that might actually be the case already (`--version` and `--help` being the only exceptions as I outlined above) but I think the issue here has to do with running old projects since this `unsafe-perm` check happens before the springboard to the older version.

If you're using Meteor 1.4.2 tool on a Meteor 1.4.2 project, everything will work fine.  If you're not running Meteor as `root` at all, this problem doesn't affect you.  But I believe this is what is happening right now if you run the 1.4.2 `meteor` command on a project that isn't 1.4.2, as `root`.

* The 1.4.2 `meteor` command runs the 1.4.2 `meteor-tool`'s `tools/cli/main.js`.  This is normal, as it does this before it springboards to the correct version (and downloads it if necessary).
* `tools/cli/main.js` begins `--unsafe-perm` check...
 * If `--unsafe-perm` is passed, execution continues (unless you're using `--version`, per the bug I mentioned above)
   * BUT, this is a problem when we springboard to the older `meteor-tool` since it won't know how to deal with the `--unsafe-perm` flag and will fail for a another reason.
 * If `--unsafe-perm` is not passed, Meteor `process.exit`'s to prevent the permission debacle.
   * This is a problem because, the older `meteor-tool` won't even be spring-boarded (since we're already exited).

As a work-around for now, if you're using Docker and building a pre-1.4.2 project, you should disable release checking and change your Meteor install to use the previous version of the Meteor tool (and ideally, whatever version of Meteor that you're using):

```sh
export METEOR_NO_RELEASE_CHECK=true # make sure this is set in your environment.
curl https://install.meteor.com/?release=1.4.1.3 | sh # You can also use the exact version you're deploying here.
```

This will have the added benefit of executing much more quickly and avoiding download of a tool that you're not going to use anyway, so you should probably be doing this in a Docker env anyway.

I will try to take a look at this today.

abernix:
Upon thinking about this some more, it would seem the primary issue is the `--unsafe-perm` getting passed to the springboarded version of Meteor.   So I might just propose the commit I just made here https://github.com/abernix/meteor/commit/fd69b0c22ef07b312838f06369e9943c8c0df4ad if anyone has thoughts on it.

I'll drop that as a PR if nobody has any objections and I'm going to try to run it through some of my own CI builds right now.

aldo235:
its stil happend to me , some one have a sollution?

abernix:
@aldo235 Let's discuss in the issue (#7959), not in the pull request.  Thanks.

PR #7823, opened 2016-09-27
Readme: Make clear how to run your app with a local checkout of Meteor

lucfranken:
The README explained how to run Meteor from a checkout: 
https://github.com/meteor/meteor#slow-start-for-developers

The example used is the Meteor docs app, which is available in the Meteor repository itself. However it did not mention how you can run the local Meteor checkout from a local app.

That is the most basic use case, for example to reproduce or even fix a bug. 

That’s why this example is based on creating a new clean app with meteor create so we focus on creating a clean reproduction. That way we can point to it when people want to debug their own issue.

This change makes it easier for beginners to use while it does not negatively impact expert users in my opinion.

**Edit:** Another thing which I am not sure of mentioning: The process described does not work actually. You first have to Fork Meteor and you have to clone your own Fork to your local machine. Not sure if that goes into too much detail. Following the steps described will only work for Meteor maintainers since they do have actual access to the Meteor repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ can run Meteor directly from a Git checkout using these steps:
 
 0. **_(Optional)_ Compile dependencies**
 
-    > This optional step requires `git`, a C and C++ compiler, autotools, and scons.
+    > This optional step requires a C and C++ compiler, autotools, and scons.
     > If this step is skipped, Meteor will simply download pre-built binaries.
 
     To build everything from scratch (`node`, `npm`, `mongodb`, etc.) run the following:
--- a/README.md
+++ b/README.md
@@ -107,23 +107,13 @@ Now you can run meteor directly from the checkout (if you did not
 build the dependency bundle above, this will take a few moments to
 download a pre-build version).
 
+Run at least one Meteor command to install required dependencies, like:
+
 ```bash
 ./meteor --help
 ```
 
-### Local docs
-
-From your checkout, you can read the docs locally. The `/docs` directory is a
-meteor application, so simply change into the `/docs` directory and launch
-the app:
-
-```bash
-cd docs/
-../meteor
-```
-
-You'll then be able to read the docs locally in your browser at
-`http://localhost:3000/`.
+You local Meteor checkout is now ready for use.
 
 Note that if you run Meteor from a git checkout, you cannot pin apps to specific
 Meteor releases or run using different Meteor releases using `--release`.
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Run at least one Meteor command to install required dependencies, like:
 ./meteor --help
 ```
 
-You local Meteor checkout is now ready for use.
+Your local Meteor checkout is now ready for use.
 
 Note that if you run Meteor from a git checkout, you cannot pin apps to specific
 Meteor releases or run using different Meteor releases using `--release`.
--- a/README.md
+++ b/README.md
@@ -85,7 +85,8 @@ can run Meteor directly from a Git checkout using these steps:
     anywhere you would normally call the system `meteor`.  For example,:
 
     ```sh
-    any-app-directory $ /path/to/meteor-checkout/meteor
+    $ cd my-app/
+    $ /path/to/meteor-checkout/meteor run
     ```
 
     > _Note:_ When running from a `git` checkout, you cannot pin apps to specific
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ structure:
 Create the demo app by running:
 
 ```bash
-./meteor create demo-app
+./meteor/meteor create demo-app
 ```
 
 To run the demo-app with the local copy of Meteor:
--- a/README.md
+++ b/README.md
@@ -41,82 +41,55 @@ meteor
 
 ## Slow Start (for developers)
 
-If you want to run on the bleeding edge, or help develop Meteor, you
-can run Meteor directly from a git checkout.
+If you want to run on the bleeding edge, or [help contribute to Meteor](Contributing.md), you
+can run Meteor directly from a Git checkout using these steps:
 
-### Clone
+0. **Clone from GitHub**
 
-```bash
-git clone --recursive git://github.com/meteor/meteor.git
-cd meteor
-```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +63,35 @@ in the root of the `meteor` repository. The typical symptom of not
 updating submodules will be `Error: Depending on unknown package ...`
 when you run most Meteor commands.
 
+### Create testing app
```
lucfranken: Agree that the user knows how to create a Meteor app. Otherwise you don't need a local checkout.

As far as I can see though it's important that the app gets created by the custom installed version (which is why referring to ./meteor/meteor instead of just meteor create) which is consistent with the part on how to run the tests as described here: https://github.com/meteor/meteor/blob/devel/Contributing.md#run-against-your-local-meteor-copy. 

Very open to improvements though to make things even more clear.

README.md
```diff
@@ -80,6 +111,8 @@ download a pre-build version).
 ./meteor --help
 ```
```
lucfranken: I agree adding an alias etc. makes things more comfortable. My approach was to behave like a first time contributor who finds a simple issue he/she wants to fix. Personally I think we should add advanced methods, like adding an alias, into https://github.com/meteor/meteor/blob/devel/Contributing.md for example.

README.md
abernix: > As far as I can see though it's important that the app gets created by the custom installed version

I'm curious why you feel this way as I don't generally find this to be a problem.  (Though I guess it can be pointed out that an app's `.meteor/release` file can have the value `none` instead of a release number, which might be related to whatever problem you're experiencing).

abernix:
What do you mean by the following?

>  The process described does not work actually. You first have to Fork Meteor and you have to clone your own Fork to your local machine. Not sure if that goes into too much detail. Following the steps described will only work for Meteor maintainers since they do have actual access to the Meteor repository.

README.md
```diff
@@ -107,23 +107,13 @@ Now you can run meteor directly from the checkout (if you did not
...
-```bash
-./meteor --help
-```
-
-### Local docs
-
-From your checkout, you can read the docs locally. The `/docs` directory is a
-meteor application, so simply change into the `/docs` directory and launch
-the app:
+Run at least one Meteor command to install required dependencies, like:
```
abernix: :+1:

README.md
```diff
@@ -107,23 +107,13 @@ Now you can run meteor directly from the checkout (if you did not
...
 
 ```bash
-cd docs/
-../meteor
+./meteor --help
 ```
 
-You'll then be able to read the docs locally in your browser at
-`http://localhost:3000/`.
+You local Meteor checkout is now ready for use.
```
abernix: This should say `Your`.

lucfranken:
Fixed

README.md
lucfranken: Fixed typo

README.md
lucfranken: 👍

README.md
```diff
@@ -61,6 +63,35 @@ in the root of the `meteor` repository. The typical symptom of not
...
+./meteor/meteor create demo-app
+```
+
+To run the demo-app with the local copy of Meteor:
+
+```bash
+cd demo-app
+../meteor/meteor
+```
+
```
abernix: I still propose getting rid of this `### Create testing app` section.  The steps above this point still aren't clear and it's not really typical for someone to create a testing app in this fashion when they're developing Meteor features.  They probably got here in the first place because they're working with an existing app and they ran into a bug or wanted a feature.

Let's consider just giving better instructions on how to use this with existing apps (as I posted [here](https://github.com/meteor/meteor/pull/7823#discussion_r80711575))

laosb:
Actually I'm thinking that we should separate this part into docs to keep the readme short.

lorensr:
folder structure might be clearer as:

```
cd ..
meteor create demo-app
cd demo-app
../meteor/meteor
```

abernix:
I've pushed a commit to this PR which cleans up the "Slow Start" section significantly (formatting, etc.) with inspiration from @lucfranken's commits and roughly addresses the original title of this PR (though the PR as submitted actually addressed many issues).

See how these changes look by [visiting this branch's `README.md`](https://github.com/lucfranken/meteor/blob/readme-use-git-checkout-on-app/README.md#slow-start-for-developers).

@laosb Your idea of separating it is the right way to do this.  I think docs would be taking it too far from the repo so maybe another file in this repo?

@lorensr Better yes, but I think specifying any recommended structure here at all is a bit odd.  As I mentioned above, presumably the users came here trying to address a specific bug or feature that they found while working on an existing app.  They likely don't need help setting up a new application, but better instructions for running the checked-out Meteor with any app.  I tried to address that clearly in bullet-point 4.  There's probably for improvement (suggesting a `mymeteor` shell `alias`, etc.).  But that'll be best in another location with more room to breath and expand.  Don't want the steps, or this `README`, to be intimidating.

I think we should get this PR merged and then we can move onto moving it into a different file.  Maybe that will be into `Contributing.md`, but perhaps a new `BUILDING.md` (better names welcome, this is just what Node uses) where details can be elaborated on (like aliases and pro-tips).

A LGTM (or two) and I think this should be merged and we'll drive from there.

README.md
```diff
@@ -41,59 +41,55 @@ meteor
...
+    $ ./meteor --help
+    ```
+
+0. **Ready to Go!**
+
+    Your local Meteor checkout is now ready to use!  You can use this `./meteor`
+    anywhere you would normally call the system `meteor`.  For example,:
+
+    ```sh
+    any-app-directory $ /path/to/meteor-checkout/meteor
```
lorensr: seeing dir listed in bash prompt is uncommon. what about
```sh
cd my-app
/path/to/meteor-checkout/meteor run
```

laosb:
Yes, moving to `BUILDING.md` or something like that would be great. Then we can remove this section from README and just add a link to Quick Start.

README.md
abernix: Sorry, my prompt has unicorns and monkeys and directories. :wink:

Fixed!

PR #7787, opened 2016-09-19
Fix typo in isobuild ignoreFiles

HashDot:
It was impossible to import a file from folder named *Icon*

e.g.: `/imports/components/Icon/index.js`

PR #7835, opened 2016-09-28
Optimize ignoreFiles patterns

sebakerckhof:
This is a continuation of @c9s work (https://github.com/meteor/meteor/pull/7483) to fix (https://github.com/meteor/meteor/issues/7482)

Merging the different regex to one is quite a bit faster than executing them one-by-one (see: https://github.com/meteor/meteor/pull/7483#issuecomment-249938794 for a rough estimate).

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -174,7 +174,7 @@ import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
     /~$/, /^\.#/,
-    /^(\.meteor\/|\.git\/|Thumbs\.db|\.DS_Store\/?|Icon.|ehthumbs\.db|\..*\.sw.|#.*#)$/, 
+    /^(\.meteor\/|\.git\/|Thumbs\.db|\.DS_Store\/?|Icon\.|ehthumbs\.db|\..*\.sw.|#.*#)$/,
       /* .meteor => avoids scanning N^2 files when bundling all packages
         .git => often has too many files to watch 
         ....sw(.) => vim swap files
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -174,7 +174,7 @@ import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
     /~$/, /^\.#/,
-    /^(\.meteor\/|\.git\/|Thumbs\.db|\.DS_Store\/?|Icon\.|ehthumbs\.db|\..*\.sw.|#.*#)$/,
+    /^(\.meteor\/|\.git\/|Thumbs\.db|\.DS_Store\/?|Icon\r|ehthumbs\.db|\..*\.sw.|#.*#)$/,
       /* .meteor => avoids scanning N^2 files when bundling all packages
         .git => often has too many files to watch 
         ....sw(.) => vim swap files
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -173,11 +173,13 @@ import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
-    /~$/, /^\.#/, /^#.*#$/,  // emacs swap files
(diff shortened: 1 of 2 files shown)


tools/tests/bundle-ignore-files.js
```diff
@@ -0,0 +1,15 @@
...
+  const patterns = bundler.ignoreFiles;
+  const matchingInputs = [
+    '.git/',
+    '.meteor/',
+    '.DS_Store',
+    '.aaabbb.swp',
+    'Thumbs.db'
+  ];
+
+  matchingInputs.forEach(input => selftest.expectEqual(patterns.some(p => p.test(input)),true));
```
abernix: Probably a good idea to add some `nonMatchingInputs` and make sure that they fail (`expectEqual`...`false`) so both sides are tested?

sebakerckhof: Yeah, i meant to do that at first, but honestly, the regex is so simple it seemed a bit pointless.

abernix: Well, funny enough, there's a bug in the regex right now. 🙊  #7787 

:smile:

sebakerckhof: Dammit :p.
But still, I mean the concatenated version will match the same things like the separate version and I don't really know how to make a good lists of stuff that doesn't match a regex.

sebakerckhof:
Ok, I added the example from the bug report as one that shouldn't match.

tools/tests/bundle-ignore-files.js
```diff
@@ -12,4 +12,10 @@ selftest.define("bundle-ignore-files", () => {
   ];
 
   matchingInputs.forEach(input => selftest.expectEqual(patterns.some(p => p.test(input)),true));
+
+  const nonMatchingInputs = [
+    '/imports/components/Icon/index.js',
```
abernix: This should actually just be `Icon` (this only matches the segment, not the full path).

Can you add `Icon.` to the `matchingInputs` test too?  Truthfully, I have no idea what an `Icon.` file is in the first place and the original commit (63092aabc4067a505ae1c1269e03f7a2b67b68c0) didn't mention, but it seems to be some sort of cache file that should be ignored and we might as well keep with that.

sebakerckhof: fixed

mjmasn: [This](http://superuser.com/questions/298785/icon-file-on-os-x-desktop) states that the file is actually called `Icon\r`. May need to update that regex again ;)

abernix: @mjmasn Wow...what the file!? That was very likely the original intention.  Thanks for finding that.

@sebakerckhof Could you update?

tools/isobuild/bundler.js
```diff
@@ -173,11 +173,13 @@ import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
-    /~$/, /^\.#/, /^#.*#$/,  // emacs swap files
-    /^\..*\.sw.$/,  // vim swap files
-    /^\.DS_Store\/?$/, /^ehthumbs\.db$/, /^Icon.$/, /^Thumbs\.db$/,
-    /^\.meteor\/$/, /* avoids scanning N^2 files when bundling all packages */
-    /^\.git\/$/ /* often has too many files to watch */
+    /~$/, /^\.#/,
+    /^(\.meteor\/|\.git\/|Thumbs\.db|\.DS_Store\/?|Icon\.|ehthumbs\.db|\..*\.sw.|#.*#)$/,
```
abernix: Per https://github.com/meteor/meteor/pull/7835#discussion_r81148872

Could you change `Icon\.` to `Icon\r`?

tools/tests/bundle-ignore-files.js
```diff
@@ -0,0 +1,22 @@
...
+const bundler = require('../isobuild/bundler.js');
+
+selftest.define("bundle-ignore-files", () => {
+  const patterns = bundler.ignoreFiles;
+  const matchingInputs = [
+    '.git/',
+    '.meteor/',
+    '.DS_Store',
+    '.aaabbb.swp',
+    'Icon.',
```
abernix: Per https://github.com/meteor/meteor/pull/7835#discussion_r81148872

Could you change `Icon.` to `Icon\r`?

tools/isobuild/bundler.js
sebakerckhof: done

PR #7837, opened 2016-09-28
Allow bcc on enrollAccount email for better monitoring

benmgreene:


lucfranken:
Interesting idea. One thing which may worry some is that _this mailbox where the BCC is sent contains all security tokens_ which could be used in a wrong way. 

What is the actual thing you want to monitor? The fact that the e-mails are sent out to your users? I mean: You don't really need those tokens right?

Another point: You are now re-using the from e-mailaddress which may or may not be available. Won't it be better to follow the convention of the Email object (https://docs.meteor.com/api/email.html#Email-send) to have a separate field in the config for that? 

That could be added in the Accounts.emailTemplates:
https://github.com/meteor/meteor/blob/master/packages/accounts-password/email_templates.js#L22

benmgreene:
Hi there. Yeah, I think the security implication that every security token would get emailed to the same address is a valid concern. But on the other hand, I can just go into my email delivery service and see the text of any of these emails anyhow. Adding a `bcc` field just gives me more options of how I consume the same info.

The actual thing I want to monitor is when and to whom these emails are going. Right now, there's really no way I can even trigger a callback that I could use to trigger some other type of real-time notification. I could of course add a notification elsewhere, but in the case of anyone using a community accounts UI package, that winds up being very defeatist.

I guess in the end, I'm not sure why `Accounts.emailTemplates` wouldn't support the same range of functionality that `Email.send` supports. Seems like Meteor devs should have the tools to hang ourselves as we see fit :-P.

Re: re-using the from email address, that's not what I am seeing (or intended). I see:

```
bcc: Accounts.emailTemplates.enrollAccount.bcc
      ? Accounts.emailTemplates.enrollAccount.bcc(user)
      : Accounts.emailTemplates.bcc,
```

Yes, I should add a default `bcc` value (`undefined`) to `Accounts.emailTemplates` -- assuming this passes muster, I'll definitely do that.

mitar:
> But on the other hand, I can just go into my email delivery service and see the text of any of these emails anyhow.

Yes, but that is inside your delivery service, not your e-mail account. How often you log in into your delivery service and how often you login into your e-mail?

I think it would be better to have a simple hook which happens at every enrollment (or reset password) and then you can do whatever you want with that.

abernix:
This should have started off as a feature request issue per the [guidelines for feature requests](https://github.com/meteor/meteor/blob/devel/Contributing.md#feature-requests).

That being said, I think the security concern with this is totally legitimate.  Yes, it exists in other ways (say, by means of your e-mail provider), but that doesn't mean that it should be made easier or duplicated.

> Seems like Meteor devs should have the tools to hang ourselves as we see fit

Disagree.  If the options are there, users will frequently use them without consideration or understanding of the implications.  Adding `bcc` encourages this unsafe behavior and I don't think disallowing `bcc` would be stifling creativity.

@mitar said:
> I think it would be better to have a simple hook which happens at every enrollment

This seems like a better approach to me.  As it stands, I feel like you could probably do this already with [`Accounts.onCreateUser`](https://docs.meteor.com/api/accounts-multi.html#AccountsServer-onCreateUser), in a much more security conscious way by providing exactly what you want in your "signup email", and not including the secure token.

Again though, this should be a feature request issue where it can be reviewed / reflected / voted upon before becoming a PR.  Closing in favor of that and happy to discuss further in that FR.

PR #7668, opened 2016-08-18
Release 1.4.2

benjamn:
As promised in [the 1.4.1 release announcement](https://forums.meteor.com/t/announcing-meteor-1-4-1/28210), Meteor 1.4.2 will be all about rebuild performance, as well as the usual bug fixes post-1.4.1.

The easiest way to get involved is to keep an eye on this pull request, and run `meteor update --release 1.4.2-beta.n` in a test app whenever you see a new `beta.n` version appear. Technically, you should wait at least a half hour after you see the version appear on the [releases](https://github.com/meteor/meteor/releases) page, since it takes a bit of time to publish the `meteor-tool` package for all architectures. If you've helped out with previous beta testing cycles, all of this should be familiar.

A more challenging way to get involved is to take matters into your own hands, and help improve Meteor rebuild performance. If you're up for that, we encourage you to identify specific performance problems, open issues to invite discussion of those problems, and then submit pull requests based on the outcomes of those discussions. Your pull requests should target the https://github.com/meteor/meteor/tree/release-1.4.2 branch instead of `devel`.

Meaningful performance work is not easy, so we really must insist on the process outlined above. Contributions that come out of nowhere are not likely to be considered or accepted, because they tend not to take the whole picture into consideration.

With that caveat out of the way, I don't think I need to convince you that build performance improvements are among the highest-impact work you can possibly do for the Meteor community. More than just improving your own developer experience, you’ll be a hero.

Here are two great documents to read to get started thinking about Meteor performance optimization:

* https://github.com/meteor/meteor/blob/devel/tools/PERFORMANCE.md
* https://medium.com/@lucashansen/faster-meteor-reloads-62cca5b56460#.rj9ysw8ut

To be clear, we are not expecting the community to do all (or even very much) of this work. Without a doubt, the core team will be working on these problems harder than anyone. If you're up for a challenge, though, we're here to support your efforts.

PR #7840, opened 2016-09-28
Reconfigure Mongo replica set when --port differs

### LINKED ISSUE #7563: [1.4] Switching run --port from initial --port stalls forever
Hi,

It seems the `meteor run` command is not working as expected when the `--port` option is specified at least since the latest patch for v1.4.0.1 (or at worst since 1.4.0 release).

Let's start from the beginning.
The command `meteor run` starts a new meteor instance on the port 3000 by default, and a new MongoDB instance on the port 3001 (Meteor Port + 1).
This behavior is working properly, as you can see on this picture I made.

`meteor run`

meteor_run

Now, I can also specify the port thanks to the `--port` options.
If I specify the port 3000, as in the default behavior, everything works fine, as you can see in the output.
Also, there is no difference between using `-p` or `--port`.

`meteor run --port 3000` or `meteor run -p 3000`

meteor_run_3000
.

However, if I try to launch a new meteor instance on another port than the 3000 one, it seems to not work properly, since it display "Launching your app forever" and never display "MongoDB started." as it should.
For example;

`meteor run --port 3002`

meteor_run_3002

I don't know how it has come to this, since it was working in v1.3.x so I guess it is about either the recent release or patch.
The weirdest part is, that if an instance is running on port 3000, I got a Mongo Error if I try to launch an other instance on another port.

dual

This behavior has been found both on Mac and Linux (2 different Ubuntu Server).
This was found in the v1.4.0.1 of Meteor, after the release patch. I don't know if the behavior exists without the patch or even on the 1.3.X release.
Every regular workarounds have been tested : meteor reset, new project with meteor create, export PORT, etc. None seems to fix the issue.

Discussions have been made about it on the Gitter: 
https://gitter.im/meteor/meteor
A forum post has been made, available here: 
https://forums.meteor.com/t/meteor-starts-only-on-port-3000-not-any-other-port/27271/10

My guess is that there has been something changed during the latest release, since it was about the new MongoDB version, and may be (still a guess) the MongoDB instance keeps trying to run on port 3001 when Meteor runs.

In consequence:

- If you try to run your instance for the first time in port XXX (different than 3000), it doesn't work. (Maybe something like if meteor port + 1 different from 3001 it crashes or something).
- If you try to run your second instance in port XXX as the first one runs in port 3000, you got MongoDB Error 100, revealing that the second instance absolutely tries to access the running instance on port 3001, which shouldn't be the expected behavior.

Hope I was pretty clear in this issue, since it is a very strange behavior, and I would like to run two instances of meteor on the same server at the same time, which I cannot at the very moment.

Thanks for reading,
Greetings,
@axelvaindal

benjamn:
[This comment](https://github.com/meteor/meteor/issues/7563#issuecomment-237544943) by @abernix explains the problem, and this PR fixes it.

Fixes #7563 and also the [`bundler-{assets,options}` test failures](https://circleci.com/gh/meteor/meteor/2648#tests/containers/3) that have been plaguing the [Release 1.4.2](https://github.com/meteor/meteor/pull/7668) branch.


tools/runners/run-mongo.js
```diff
@@ -213,8 +213,7 @@ if (process.platform === 'win32') {
             var foundPort = parseInt(m[2], 10);
             var foundPath = m[3];
 
-            if ( (! port || port === foundPort) &&
-                 (! dbDir || dbDir === foundPath)) {
+            if (! dbDir || dbDir === foundPath) {
```
benjamn: Without this change, the only `mongod` processes we'll find are those using the same port we're trying to use, which is not the case if we've changed the `--port`.

Since we're still checking the `--dbPath`, this search won't return any `mongod` processes that don't belong to this app.

tools/runners/run-mongo.js
```diff
@@ -613,10 +632,16 @@ var launchMongo = function (options) {
...
-          db.admin(), 'command', {replSetInitiate: configuration});
+        yieldingMethod(db.admin(), 'command', {
+          replSetInitiate: configuration,
+        });
       } catch (e) {
-        if (e.message !== 'already initialized') {
+        if (e.message === 'already initialized') {
+          yieldingMethod(db.admin(), 'command', {
+            replSetReconfig: configuration,
+            force: true,
```
benjamn: I believe the use of `force: true` is safe here because this code runs in development, as the Mongo server is starting up, before any Meteor code has a chance to modify the database.

Reference: https://docs.mongodb.com/manual/reference/command/replSetReconfig/

tools/runners/run-mongo.js
tmeasday: This seems incorrect. `findMongoPids` is used in two places:

1. [To find the mongo port](https://github.com/meteor/meteor/blob/b484edf66964ad5e90dd8c2b9fc38653e3bad7f5/tools/runners/run-mongo.js#L235) as `findMongoPids(dbDir)`

2. [To kill the running mongo](https://github.com/meteor/meteor/blob/b484edf66964ad5e90dd8c2b9fc38653e3bad7f5/tools/runners/run-mongo.js#L297) as `findMongoPids(null, port)`.

In the first case, this change won't do anything as `port` is undefined. In the second case, this will now match *all* mongo processes, as `dbDir` is `null`.

I don't understand the reason for this change I guess?

tmeasday: Also, shouldn't we make the same change on line 134 (the windows version). Although looking at that code, I don't see how it works in case 1. above either (it has the reverse problem).

abernix: When I was debugging this issue, I didn't recall running into this issue, but yes, the original check raises some question and as @benjamn says:

> Since we're still checking the --dbPath, this search won't return any mongod processes that don't belong to this app.

tools/runners/run-mongo.js
```diff
@@ -537,6 +536,11 @@ var launchMongo = function (options) {
         maybeReadyToTalk();
       }
 
+      if (/[ReplicationExecutor] Locally stored replica set configuration does not have a valid entry for the current node/) {
```
abernix: Hmm... the `[` and `]` aren't escaped so it would never match the actual line but the regex itself doesn't test _against_ anything (à la `.test(data)`) so it would be setting `replSetReadyToBeInitiated` on every log line.  Maybe it's actually fine to just always reconfig?  Heh.

tools/runners/run-mongo.js
abernix: Agree `force` is fine, but perhaps catching this error is all that's necessary?

tools/runners/run-mongo.js
benjamn: Oh, hah, wow, this is terrible, thanks.

tools/runners/run-mongo.js
benjamn: @tmeasday The `findMongoPids(null, port)` version is broken because we pass the new `port`, though `mongod` is still running on the old port, which we don't know. There seems to be an old assumption that `mongod` can run with the same `dbPath` but different ports, so if you kill the one running on the port you're about to use, then all's well. That's just not true anymore, if it ever was.

benjamn:
Reverted the changes to `findMongoPids` in favor of a more general solution: https://github.com/meteor/meteor/pull/7840/commits/295d3d5678228f06ee0ab6c0d60139849a0ea192

tools/runners/run-mongo.js
benjamn: Without `force: true`, the exception indicates that no changes were made against the secondary (because it wasn't primary).

abernix: Sorry, that was poorly worded on my part and separate conversations.

Issue #7857, opened 2016-10-02
Support for new field update operators ($min, $max, $currentDate and $mul)


sebakerckhof:
Since mongo 2.6+ there are some new field operators: $min, $max, $currentDate and $mul.

Minimongo currently does not yet support these modifiers. @stubailo [encouraged](https://forums.meteor.com/t/meteor-1-4-1-1-does-not-support-mongo-3-2-operators-min-max/29900/8?u=seba) to make a PR for this, which I plan to do. So I mainly create this issue to link the PR to an issue.

stubailo:
I'd suggest doing it in several smaller PRs, if possible - that way we can merge each individual one faster.

stubailo:
Oh never mind, I see that you have opened a PR already.

sebakerckhof:
Yeah, I could still split them up if you think it's better, but even with them combined it's a pretty small PR.

stubailo:
Yep, no problem.

tmeasday:
#7858 is merged!

PR #7858, opened 2016-10-02
Implement $min, $max, $currentDate modifiers

sebakerckhof:
Related issue: https://github.com/meteor/meteor/issues/7857

This implements $min, $max and partially $currentDate + tests

I didn't implement $mul, since there can be [numeric type conversions](https://docs.mongodb.com/v2.6/faq/developers/#faq-developers-multiplication-type-conversion) which I don't know how to deal with in JS. There's also no support for `timestamp` type in $currentDate, since from what I understand of [the docs](https://docs.mongodb.com/manual/reference/bson-types/#document-bson-type-timestamp), this is an internal type I can't generate myself.

### DIFF
--- a/packages/minimongo/modify.js
+++ b/packages/minimongo/modify.js
@@ -184,33 +184,40 @@ var NO_CREATE_MODIFIERS = {
 var MODIFIERS = {
   $currentDate: function (target, field, arg) {
     if (typeof arg === "object" && arg.hasOwnProperty("$type")) {
-       if (arg.$type !== "date")
+       if (arg.$type !== "date") {
          throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers");
-    } else if(arg !== true) {
+       }
+    } else if (arg !== true) {
       throw MinimongoError("Invalid $currentDate modifier");
     }
     target[field] = new Date();
   },
   $min: function (target, field, arg) {
-    if (typeof arg !== "number")
+    if (typeof arg !== "number") {
       throw MinimongoError("Modifier $min allowed for numbers only");
+    }
     if (field in target) {
-      if (typeof target[field] !== "number")
+      if (typeof target[field] !== "number") {
         throw MinimongoError("Cannot apply $min modifier to non-number");
-      if (target[field] > arg)
+      }
+      if (target[field] > arg) {
         target[field] = arg;
+      }
     } else {
       target[field] = arg;
     }
   },
   $max: function (target, field, arg) {
-    if (typeof arg !== "number")
+    if (typeof arg !== "number") {
       throw MinimongoError("Modifier $max allowed for numbers only");
+    }
     if (field in target) {
-      if (typeof target[field] !== "number")
+      if (typeof target[field] !== "number") {
         throw MinimongoError("Cannot apply $max modifier to non-number");
-      if (target[field] < arg)
-        target[field] = arg;
+      }
+      if (target[field] < arg) {
+         target[field] = arg;
+      }
     } else {
       target[field] = arg;
     }
--- a/packages/minimongo/modify.js
+++ b/packages/minimongo/modify.js
@@ -182,35 +182,34 @@ var NO_CREATE_MODIFIERS = {
 };
 
 var MODIFIERS = {
-  $currentDate: function(target, field, arg){
-    if(typeof arg === "object" && arg.hasOwnProperty("$type")){
-       if(arg.$type !== "date"){
-        throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers");
-       }
-    }else if(arg !== true){
+  $currentDate: function (target, field, arg) {
+    if (typeof arg === "object" && arg.hasOwnProperty("$type")) {
+       if (arg.$type !== "date")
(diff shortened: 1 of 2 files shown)


packages/minimongo/modify.js
```diff
@@ -182,6 +182,40 @@ var NO_CREATE_MODIFIERS = {
 };
 
 var MODIFIERS = {
+  $currentDate: function(target, field, arg){
+    if(typeof arg === "object" && arg.hasOwnProperty("$type")){
+       if(arg.$type !== "date"){
+        throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers");
+       }
+    }else if(arg !== true){
```
mitar: Style.

sebakerckhof: Is there a style guide somewhere? The only reference I could find in the contributing.md refers to: https://guide.meteor.com/code-style.html#javascript which doesn't say a lot.

sebakerckhof: But most of the code doesn't seem to match that guide?

abernix:
@sebakerckhof Could you rebase this off `devel`?  I believe it should solve the failing tests.

sebakerckhof:
@abernix done.

This would probably need to get an entry in history.md, but I don't know yet under which section it will come.

laosb:
@sebakerckhof In most cases you should just put your entry in "v.NEXT" section :)

packages/minimongo/modify.js
```diff
@@ -182,6 +182,39 @@ var NO_CREATE_MODIFIERS = {
 };
 
 var MODIFIERS = {
+  $currentDate: function (target, field, arg) {
+    if (typeof arg === "object" && arg.hasOwnProperty("$type")) {
+       if (arg.$type !== "date")
+         throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers");
+    } else if(arg !== true) {
```
tmeasday: Spaces around `()` please

packages/minimongo/modify.js
```diff
@@ -182,6 +182,39 @@ var NO_CREATE_MODIFIERS = {
...
+      throw MinimongoError("Invalid $currentDate modifier");
+    }
+    target[field] = new Date();
+  },
+  $min: function (target, field, arg) {
+    if (typeof arg !== "number")
```
tmeasday: We use `{}` for if statements always (or we should anyway).

sebakerckhof: Ok, but that was copied from the already existing $inc function.

packages/minimongo/modify.js
sebakerckhof: Which ones, they have spaces?

tmeasday:
```
} else if(arg !== true) {
```
->
```
} else if (arg !== true) {
```

sebakerckhof: Oh yeah, I missed that one.

sebakerckhof:
Fixed the styling. No problem if you're being strict on styling, but if you want to open up more to contributors it would be good if there was an official styling guide and if MDG also followed that.

tmeasday:
Yeah it's pretty tricky because the code base is pretty large and I guess the styling people used changed over the years. I think following the file the code is in is pretty reasonable.

abernix:
@sebakerckhof Re: Style. As @mitar pointed out earlier in this issue, the [Contributing.md](https://github.com/meteor/meteor/blob/devel/Contributing.md#submitting-pull-requests) links (somewhat annoyingly, via an extra click) to https://guide.meteor.com/code-style.html#style-guide says that [Airbnb style](https://github.com/airbnb/javascript) is recommended.  In my experience, MDG doesn't request changes to PRs which match Airbnb style and new code that is being added does _usually_ match up with that.  But the Meteor code base is far too big to go through and change all the older code that (mostly) was written before there was any such thing as a popular style guide for JavaScript.  It would be risky, bug-introducing, and destroy a lot of Git-blame history to change it all at once, as glorious as that might be.  If there is anything that can be done to make this more clear (like updating Contributing.md), a PR is likely welcome for that.

tmeasday:
Thanks @sebakerckhof!

packages/minimongo/modify.js
```diff
@@ -182,6 +182,39 @@ var NO_CREATE_MODIFIERS = {
...
+      throw MinimongoError("Modifier $min allowed for numbers only");
```
clemenshelm: Before I open an issue on this: Is there a reason why the operator isn't allowed for dates?

mitar: Is this supported in MongoDB? Interesting.

sebakerckhof: There's no specific reason, I wasn't aware it could be used for dates. I'll create a follow-up PR.

clemenshelm: Great, thanks 😊

sebakerckhof: I looked into this a bit further and apparently $min and $max can be used to compare any types, where [BSON type order](https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#faq-dev-compare-order-for-bson-types) is applied. So there's quite a bit of work involved to be fully compliant.

PR #7860, opened 2016-10-02
Add warning when login token is rejected

dhrubins:
When using meteor deploy to galaxy via continuous integration, our deployments suddenly stopped working. We filed a support ticket with galaxy (4218) because our environment hadn't changed (so we thought). After investigating, we found our token simply expired. We only figured this out when we noticed that our CI deployments timed out during a login prompt.

The whole process could have been better if the meteor tool provided a better error message (actually, any error message) when there's a login fail caused by an expired token. This PR attempts to provide at least an indication the token could be the problem, rather than failing silently.

### DIFF
--- a/tools/meteor-services/auth.js
+++ b/tools/meteor-services/auth.js
@@ -65,6 +65,9 @@ var loggedInAccountsConnection = function (token) {
     connection.close();
 
     if (err.error === 403) {
+      // Better than saying nothing at all
+      Console.error("warning: login token was rejected by the server");
+
       // This is not an ideal value for the error code, but it means
       // "server rejected our access token". For example, it expired
       // or we revoked it from the web.


tools/meteor-services/auth.js
```diff
@@ -65,6 +65,9 @@ var loggedInAccountsConnection = function (token) {
     connection.close();
 
     if (err.error === 403) {
+      // Better than saying nothing at all
+      Console.error("warning: login token was rejected by the server");
```
abernix: Right now, adding this message here would actually double up error messages in some places, [like here](https://github.com/meteor/meteor/blob/fe143dd84752bc3dea42ea35d44c2d4e93ba14f5/tools/cli/commands.js#L1904-L1907).

I think there might be a better place to put this, [maybe here?](https://github.com/meteor/meteor/blob/fe143dd84752bc3dea42ea35d44c2d4e93ba14f5/tools/meteor-services/auth.js#L620-L676)

What was the actual solution to your issue? Perhaps that could be explained too.

abernix: Also, I'd question having the word `warning:` in a `Console.error`.  Since it seems like this was an error actually preventing you from continuing, it seems that removing `warning:` would be more appropriate.

dhrubins: Ok, thanks. I'll take a look at that other location. If I make changes, should I just close out this PR and resubmit as a new one?

The solution was to simply regenerate the token - it was a ~30 sec fix, but it took us quite an effort to figure it out on the CI system since there was no indication of the actual problem.

RE: messaging - agreed, we should take the word warning out of the error message. I think it's still good to have an error so it is more likely to be reported, rather than ```console.warn``` which might be overlooked.

abernix: > The solution was to simply regenerate the token

Ok, so the solution was needing to re-`meteor login`, generate a new `METEOR_SESSION_FILE` before doing a `meteor deploy`?  Just trying to figure out where this error was getting thrown.  Plus, I'm mainly curious if we should make this error actually tell them how to fix the problem (i.e. Find something/add something to docs and link to it in the actual error).

You can just push new commits to the same branch (`LegalRobot/meteor-tool-auth-messaging`) and it will update this PR. :)

abernix: @dhrubins Any updates on this?  This PR is appreciated and this sounds bizarre and annoying enough to track down that this would likely be really nice to get a fix in for if you have some time.

dhrubins: Sorry, been busy lately... I think I can close this out today.

abernix:
@dhrubins Gentle nudge. :smile:

benjamn:
Closing now due to inactivity, but feel free to reopen if you have more time to work on this. We'd be happy to take a fix for this.

Louis30:
Reject

Louis30:
Close

PR #7865, opened 2016-10-04
Replaces Meteor configured host with the original one coming in main request

chongwang87:
it takes the protocol (http or https) from Meteor.absoluteUrl() and
replaces the Meteor configured host with the original one coming in main request


abernix:
You are (effectively) resubmitting the same pull request which was previously rejected because it was thought to [have some security implications](https://github.com/meteor/meteor/pull/4480#issuecomment-112579472) (#4480) and you haven't made any changes to account for that.

Let's discuss this further in the issue before opening the same PR (again).

Issue #7673, opened 2016-08-19
[Up for grabs] Optimize ClientTarget#minifyCss


benjamn:
**Normally I would assign this to myself, but it seems like a nice task for an outside contributor, since it shouldn't require a deep understanding of the Meteor build tool. If you're interested in tackling this, or have any insight to share, please comment below!**

A number of [profiling](https://github.com/meteor/meteor/issues/7493#issuecomment-235777225) [reports](https://github.com/meteor/meteor/issues/7493#issuecomment-240765913) have indicated that [`ClientTarget#minifyCss`](https://github.com/meteor/meteor/blob/88db65c8378336d20104792c7f2ce3604afa7d74/tools/isobuild/bundler.js#L1239) is taking too long, especially when an app has many `.css` files.

I suspect a large part of the problem lies in composing CSS source maps, which happens [here](https://github.com/meteor/meteor/blob/88db65c8378336d20104792c7f2ce3604afa7d74/packages/standard-minifier-css/plugin/minify-css.js#L107-L113), and takes more than 8 seconds in the Workpop profile.

I'm not even sure that code is doing what it claims to do, since `originals` contains all the CSS files that are being combined, and `newMap.applySourceMap` is a method that you're supposed to call when there are multiple transforms at work, such as SCSS-to-CSS and CSS-to-minified-CSS.

Suggested approach:

- [ ] To measure your optimizations accurately, you'll want to have a test app with lots of `.css` files, and you'll want to run it with `METEOR_PROFILE=100 meteor`. Make sure you have enough `.css` files so that `ClientTarget#minifyCss` and `composing source maps` shows up in the profile.

- [ ] As a first step, I would try simply disabling the source map combination code, to approximate the ideal performance improvement.

- [ ] As a second step, if you believe the source map combination code is important enough to justify the cost, I would try to rewrite it using [the `sourcemap.SourceNode` API](https://www.npmjs.com/package/source-map#sourcenode).

Your changes should be confined to the `standard-minifier-css` package, which is good news because it means that we can release a new version of that package without having to do a whole new Meteor release.

veered:
I opened a PR that does the most basic possible implementation of this optimization [here](https://github.com/meteor/meteor/pull/5875). Apparently this implementation has some issues, but I've been using it on our large web application for 6+ months without a single issue.

arist0tl3:
To follow up, I've been using the solution @veered posted for ~2 months on a medium-sized project. Client-side changes used to look like:

```
[...]

PR #7877, opened 2016-10-06
Rewrite source map composition for CSS files

klaussner:
The previous implementation had a run time quadratic in the number of CSS files with an associated source map. This new implementation has a linear run time because it iterates only once through all mappings of each CSS file.

Fixes the performance problems discussed in #7673.

### DIFF
--- a/packages/standard-minifier-css/plugin/minify-css.js
+++ b/packages/standard-minifier-css/plugin/minify-css.js
@@ -149,23 +149,28 @@ var mergeCss = Profile("mergeCss", function (css) {
 
     // Find mappings from the concatenated file back to the original files
     concatConsumer.eachMapping(function (mapping) {
-      var consumer = consumers[mapping.source];
-      var source;
+      var source = mapping.source;
+      var consumer = consumers[source];
 
-      // If there is a source map for the original file, e.g., if it has been
-      // compiled from Less to CSS, find the source location in the original's
-      // original file. Otherwise, use the mapping of the concatenated file's
-      // source map.
       var original = {
         line: mapping.originalLine,
         column: mapping.originalColumn
       };
 
+      // If there is a source map for the original file, e.g., if it has been
+      // compiled from Less to CSS, find the source location in the original's
+      // original file. Otherwise, use the mapping of the concatenated file's
+      // source map.
       if (consumer) {
-        original = consumer.originalPositionFor(original);
-        source = original.source;
-      } else {
-        source = mapping.source;
+        var newOriginal = consumer.originalPositionFor(original);
+
+        // Finding the original position should always be possible (otherwise,
+        // one of the source maps would have incorrect mappings). However, in
+        // case there is something wrong, use the intermediate mapping.
+        if (newOriginal.source !== null) {
+          original = newOriginal;
+          source = original.source;
+        }
       }
 
       // Add a new mapping to the final source map
@@ -174,10 +179,7 @@ var mergeCss = Profile("mergeCss", function (css) {
           line: mapping.generatedLine,
           column: mapping.generatedColumn
         },
-        original: {
-          line: original.line,
-          column: original.column
-        },
+        original: original,
         source: source
       });
 
--- a/packages/standard-minifier-css/plugin/minify-css.js
+++ b/packages/standard-minifier-css/plugin/minify-css.js
@@ -119,26 +119,73 @@ var mergeCss = Profile("mergeCss", function (css) {
 
   var newMap;
 
+  // Compose the concatenated file's source map with source maps from the
+  // previous build step if necessary.
   Profile.time("composing source maps", function () {
-    // If any input files had source maps, apply them.
-    // Ex.: less -> css source map should be composed with css -> css source map
-    newMap = sourcemap.SourceMapGenerator.fromSourceMap(
-      new sourcemap.SourceMapConsumer(stringifiedCss.map));
+    var concatConsumer;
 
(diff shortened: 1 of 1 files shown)


packages/standard-minifier-css/plugin/minify-css.js
```diff
@@ -119,26 +119,73 @@ var mergeCss = Profile("mergeCss", function (css) {
...
-    newMap = sourcemap.SourceMapGenerator.fromSourceMap(
-      new sourcemap.SourceMapConsumer(stringifiedCss.map));
+    var concatConsumer;
 
-    Object.keys(originals).forEach(function (name) {
+    newMap = new sourcemap.SourceMapGenerator();
+    concatConsumer = new sourcemap.SourceMapConsumer(stringifiedCss.map);
+
+    // Create a dictionary of source map consumers for fast access
+    var consumers = Object.keys(originals).reduce(function (consumers, name) {
```
benjamn: I may be biased against `Array.prototype.reduce`, but I think this might be clearer if you defined `var consumers = Object.create(null)` above, and simply modified it in a `.forEach` loop.

packages/standard-minifier-css/plugin/minify-css.js
```diff
@@ -119,26 +119,73 @@ var mergeCss = Profile("mergeCss", function (css) {
...
+        } catch (err) {
+          // If we can't apply the source map, silently drop it.
+          //
+          // XXX This is here because there are some less files that
+          // produce source maps that throw when consumed. We should
+          // figure out exactly why and fix it, but this will do for now.
+        }
+      }
+
+      return consumers;
```
benjamn: No need to `return consumers` if you switch to a `.forEach` loop.

packages/standard-minifier-css/plugin/minify-css.js
```diff
@@ -119,26 +119,73 @@ var mergeCss = Profile("mergeCss", function (css) {
...
+      // Add a new mapping to the final source map
+      newMap.addMapping({
+        generated: {
+          line: mapping.generatedLine,
+          column: mapping.generatedColumn
+        },
+        original: {
+          line: original.line,
+          column: original.column
+        },
```
benjamn: It looks like `consumer.originalPositionFor` always [returns a new/unshared object](https://github.com/mozilla/source-map/blob/b37449d3c499ab7225a9bed28d0a218fde987691/lib/source-map-consumer.js#L628), so I think you can just say `original: original` here.

packages/standard-minifier-css/plugin/minify-css.js
```diff
@@ -119,26 +119,73 @@ var mergeCss = Profile("mergeCss", function (css) {
...
+      // compiled from Less to CSS, find the source location in the original's
+      // original file. Otherwise, use the mapping of the concatenated file's
+      // source map.
+      var original = {
+        line: mapping.originalLine,
+        column: mapping.originalColumn
+      };
+
+      if (consumer) {
+        original = consumer.originalPositionFor(original);
```
benjamn: What happens if there's no original position ([this case](https://github.com/mozilla/source-map/blob/b37449d3c499ab7225a9bed28d0a218fde987691/lib/source-map-consumer.js#L637))? Do we still want to call `newMap.addMapping` below?

benjamn:
Circle CI test failures are unrelated to this change.

abernix:
Hmm, this seems to have broken projects that are using `less` with [C++/multi-line/block style comments](http://lesscss.org/features/#features-overview-feature-comments), [like this](https://github.com/abernix/1.4.2-minify-css-error/blob/master/client/main.less) as shown in [this reproduction repo](https://github.com/abernix/1.4.2-minify-css-error/).

```
=> Errors prevented startup:

   While minifying app stylesheet:

   /Users/jesse/Dev/1.4.2-minify-css-error/.meteor/local/isopacks/standard-minifier-css/plugin.minifyStdCSS.os/npm/node_modules/meteor/minifyStdCSS/node_modules/source-map/lib/source-map-consumer.js:704:13:
   "app/client/main.less.css" is not in the SourceMap.
   at SourceMapConsumer_sourceContentFor
   (/Users/jesse/Dev/1.4.2-minify-css-error/.meteor/local/isopacks/standard-minifier-css/plugin.minifyStdCSS.os/npm/node_modules/meteor/minifyStdCSS/node_modules/source-map/lib/source-map-consumer.js:704:13)
   at packages/minifyStdCSS_plugin.js:202:38
   at Array.forEach (native)
   at SourceMapConsumer_eachMapping
   (/Users/jesse/Dev/1.4.2-minify-css-error/.meteor/local/isopacks/standard-minifier-css/plugin.minifyStdCSS.os/npm/node_modules/meteor/minifyStdCSS/node_modules/source-map/lib/source-map-consumer.js:155:14)
   at packages/minifyStdCSS_plugin.js:164:20
   at Function.time (/tools/tool-env/profile.js:305:10)
   at packages/minifyStdCSS_plugin.js:137:11
   at CssToolsMinifier.processFilesForBundle (packages/minifyStdCSS_plugin.js:33:16)
```

klaussner:
Good catch! My assumption that a source map has invalid mappings if an original source position cannot be found (in response to @benjamn's third comment above) was wrong. It looks like `stringifyCss` creates mappings for block comments but the Less compiler doesn't, so there are no original source positions for them. Fix should be ready soon. :)

benjamn:
@abernix @klaussner I'm working on this right now, as a matter of fact! Hopefully our solutions agree…

PR #7886, opened 2016-10-07
Add Cordova Platform Dependency Tests

abernix:
* Adds test which would have caught #7884 (fails without it, passes with it)
* Adds (Unix-only) test which checks for message when platform is already added (previously no test)
* Fix incorrectly redeclared `var` + s/`var`/`let`/

### DIFF
--- a/tools/tests/cordova-platforms.js
+++ b/tools/tests/cordova-platforms.js
@@ -4,7 +4,7 @@ var files = require('../fs/files.js');
 
 selftest.define("add cordova platforms", ["cordova"], function () {
   var s = new Sandbox();
-  var run;
+  let run;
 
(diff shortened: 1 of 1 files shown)


abernix:
Ok, haha, it won't fail on CircleCI, but it will fail on another environment where you run the test where the Android SDK is **not** installed, which is what this PR is intended to fix.

Examples: [My Semaphore CI build for this same PR](https://semaphoreci.com/abernix/meteor/branches/feature-add-cordova-dependency-test/builds/3), or anyone who doesn't have Android SDK installed who runs this locally:

    METEOR_PRETTY_OUTPUT=0 ./meteor self-test "add cordova platforms"

PR #7898, opened 2016-10-08
Fix Travis CI tests by setting mongoOptions.server.reconnectTries = Infinity.

benjamn:
Travis CI builds have been failing intermittently [since around the time #7880 was merged](https://travis-ci.org/meteor/meteor/builds).

I'm going to use this pull request to try to fix those tests. As a start, I've merely bumped the `npm-mongo` and `mongo` package versions. I don't expect that to solve the problem, but it's a start.

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -131,14 +131,17 @@ MongoConnection = function (url, options) {
   self._onFailoverHook = new Hook;
 
   var mongoOptions = _.extend({
-      db: { safe: true },
-      // Set reconnectTries to 0 which means keep trying to reconnect forever,
-      // rather than the default of losing the connection permanently after 30
-      // retries (separated by 1000ms).
-      server: { reconnectTries: 0 },
-      replSet: {}
+    db: { safe: true },
+    // http://mongodb.github.io/node-mongodb-native/2.2/api/Server.html
+    server: {
+      // Reconnect on error.
+      autoReconnect: true,
+      // Try to reconnect forever, instead of stopping after 30 tries (the
+      // default), with each attempt separated by 1000ms.
+      reconnectTries: Infinity
     },
-    Mongo._connectionOptions);
+    replSet: {}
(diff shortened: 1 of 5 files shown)


packages/mongo/mongo_driver.js
```diff
@@ -131,14 +131,17 @@ MongoConnection = function (url, options) {
...
-      // Set reconnectTries to 0 which means keep trying to reconnect forever,
-      // rather than the default of losing the connection permanently after 30
-      // retries (separated by 1000ms).
-      server: { reconnectTries: 0 },
-      replSet: {}
+    db: { safe: true },
+    // http://mongodb.github.io/node-mongodb-native/2.2/api/Server.html
+    server: {
+      // Reconnect on error.
+      autoReconnect: true,
```
tmeasday: I'm not sure this is actually an option any more. Well, actually the documentation is inconsistent.

Here: http://mongodb.github.io/node-mongodb-native/2.2/reference/connecting/connection-settings/ it definitely mentions `autoReconnect` in the same breath as `reconnectTries` although doesn't actually mention that either should be on the `server` key.

Here: `http://mongodb.github.io/node-mongodb-native/2.2/api/Server.html` it doesn't mention `autoReconnect`.

Ok, I see there is something in the code: https://github.com/mongodb/node-mongodb-native/blob/2.2/lib/server.js#L73

benjamn: Ah, you're right, the documentation only mentions `server.socketOptions.autoReconnect`. Does that have anything to do with this?

tmeasday: It ultimately gets passed in here: https://github.com/christkv/mongodb-core/blob/2.0/lib/topologies/server.js#L35

tmeasday: And `server.socketOptions` gets merged with the `server` options anyway here: https://github.com/mongodb/node-mongodb-native/blob/2.2/lib/server.js#L133

Issue #6882, opened 2016-04-22
[cordova] Inappbrowser White Screen during oAuth Login


brianscroggins24:
Currently in one of my Meteor apps i'm using loginWithFacebook and loginWithGoogle auth along with the ServiceConfiguration package. When the inappbrowser comes up to login a user in they are able to enter their credentials. However when they submit those creds the the inappbrowser white screens. The activity only occurs on the device, on the simulator the inappbrowser disappears and logs the user in. I've tried every solution including adding the cordova-inappbrower-plugin to my project and adding loginStyle: 'popup' to the ServiceConfiguration but nothing has work. If anyone know of why this might be happening please shed some light. I'm working on iOS9 for both the simulator and the device.

terenceng2010:
Looks like an upstream issue
https://issues.apache.org/jira/plugins/servlet/mobile#issue/CB-11136

terenceng2010:
my current temporary fix to this issue is to override the cordova-plugin-inappbrowser plugin using by package oauth by adding this line in cordova-plugins:

`cordova-plugin-inappbrowser@https://github.com/ephemer/cordova-plugin-inappbrowser.git#04091fde737519c149e7ad6316971cb6b490c5b3`

tekknikk:
@terenceng2010 your fix works great.

abernix:
Per the debugging I did in #7102 and a repo that @mitiaptest posted, I'm re-posting the repo here as a reproduction for this problem so it will be easy to confirm when it's fixed.  I can confirm this issue and also confirm that it is resolved with the fix that @terenceng2010 suggested (command below)

##### Reproduction (see README.md for steps)
https://github.com/abernix/issue-6882

##### Workaround

    meteor add cordova:cordova-plugin-inappbrowser@https://github.com/ephemer/cordova-plugin-inappbrowser.git#04091fde737519c149e7ad6316971cb6b490c5b3

hexsprite:
@abernix I'm also having this issue. If it's as easy as upgrading the reference to the plugin couldn't this be released? Or is there concern about this causing some other impacts?

hexsprite:
OK, I dug a bit deeper and basically Google is deprecating this form of login. But in the meantime, it would be great to have a working Google oAuth with Meteor accounts on Cordova. So maybe we could cut a release with just this one patch enabled.

https://github.com/apache/cordova-plugin-inappbrowser/pull/162#issuecomment-243054022

In the near future switch to something like:
[...]

PR #7908, opened 2016-10-11
Prevent error w/ no hash in oauth redirect url

ramezrafla:
We have a case where the returned oauth redirect does not have a hash. We propose a graceful exit when that happens. We also propose to increase the timeout to 500ms for closing the popup, to give the server the chance to react to the oauth redirect.

### DIFF
--- a/packages/oauth/oauth_cordova.js
+++ b/packages/oauth/oauth_cordova.js
@@ -48,7 +48,7 @@ OAuth.showPopup = function (url, callback, dimensions) {
       setTimeout(function () {
         popup.close();
         callback();
-      }, 500);
+      }, 100);
     }
   };
 
--- a/packages/oauth/oauth_cordova.js
+++ b/packages/oauth/oauth_cordova.js
@@ -44,7 +44,6 @@ OAuth.showPopup = function (url, callback, dimensions) {
       // https://issues.apache.org/jira/browse/CB-2285.
       //
       // XXX Can we make this timeout smaller?
-
       setTimeout(function () {
         popup.close();
         callback();
--- a/packages/oauth/oauth_cordova.js
+++ b/packages/oauth/oauth_cordova.js
@@ -29,13 +29,10 @@ OAuth.showPopup = function (url, callback, dimensions) {
       var splitUrl = event.url.split("#");
       var hashFragment = splitUrl[1];
 
-      if (! hashFragment) {
-        throw new Error("No hash fragment in OAuth popup?");
+      if (hashFragment) {
+        var credentials = JSON.parse(decodeURIComponent(hashFragment));
+        OAuth._handleCredentialSecret(credentials.credentialToken,credentials.credentialSecret);
       }
-
-      var credentials = JSON.parse(decodeURIComponent(hashFragment));
-      OAuth._handleCredentialSecret(credentials.credentialToken,
-                                    credentials.credentialSecret);
 
       oauthFinished = true;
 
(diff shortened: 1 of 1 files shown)


abernix:
Thanks for opening a PR!

The first part of this PR (and the one described in the title/commit) seems reasonable to me (almost without discussion), but can you talk a little bit about what the case is when the oAuth Redirect does not have a hash?

Please remove the second part (raising the timeout) and make it a separate PR (one issue per PR, and one commit per functional change).  In the new PR, you'll also want to explain much more clearly _why_ you're proposing to raise a timeout that already has a `XXX` comment on it trying to lower it.  I can kind of understand your issue, but it seems like it might be specific to your platform (in other words, some servers might return faster or slower).  Either way, discussion for a separate PR. :smile:

ramezrafla:
Thanks @abernix for the quick review of my PR. Very appreciated.
We reverted the timeout as you instructed.

Here is our case: we use Clever.com which is an educational OAuth service. I am not an expert on OAuth so don't know if the hash is mandatory, but the oauth url is called with query parameters only (namely ```code``` and some custom ones). The change we made to gracefully degrade is the only change required to successfully work with this OAuth service. I recognize that most people will use standard providers (Google, Facebook and Twitter) and we are the exception. This change will not affect them.

abernix:
I've looked at this a few times now and I'm not entirely convinced that this should be necessary, nor do I believe that if this change was implemented the oAuth process would actually be working properly.  More so, the fact that this could flag `oauthFinished` as `true` when there were not credentials is odd because the oAuth process is not actually successful unless `credentialsSecret` (this is supposed to be the `access_token` received from Clever.  This flag being set before it is supposed to seems like it could cause the `pageLoaded` function [not to run again](https://github.com/meteor/meteor/blob/devel/packages/oauth/oauth_cordova.js#L24-L26).

My confusion is further amplified by the fact that `credentialsToken` and `credentialsSecret` are not set directly by the oAuth provider (in your case, Clever), but rather by [this code in `end_of_popup_response.js`](https://github.com/meteor/meteor/blob/devel/packages/oauth/end_of_popup_response.js#L9-L16).

I looked at [Clever.com's oAuth documentation](https://dev.clever.com/instant-login/bearer-tokens) and they're using OAuth 2.0.  Since Meteor uses the oAuth2 implementations of Google, Facebook, Weebo, and GitHub, I'm going to have to presume one of two things:

1. This is a bug in some other part of the process which hasn't been clearly identified here.  #6882, maybe?  It's important to identify and see the bug we're trying to fix here.
2. There is something missing in your implementation of the Clever oAuth 2.0 API.

I think the best way to proceed right now is to request that you open an issue with a reproducible example of the problem you're trying to fix with this PR.  While I didn't go to the extent of making a reproduction when reviewing this PR, I'm happy to create my own Clever dev account and generate my own credentials if you can provide a pre-built application that I can plug those credentials into and see the problem.

If we can clearly get to the bottom of what the PR is trying to fix then perhaps a more clear fix will become obvious, but as of right now it seems that the hash is absolutely mandatory and that its absence would be a concern.

I'm going to close this and request an issue with the problem clearly reproduced be opened instead – we can reopen this if it turns out this is the correct solution.

ramezrafla:
Thanks @abernix 
I'll dig deeper into it, but I do know that our authentication fails with Clever on mobile if we don't have this change. So we'll run off a fork of this package for now. Given we are the only ones using this, I believe your conclusion is safe for the community.

Issue #6952, opened 2016-05-02
Large number of node_modules/files causes "too many open files" on OS X


emlai:
**_1 Upvote**_ I have ~40000 files (~7GB altogether) in my `public` directory. When I run `meteor`, the "building the application" step takes 3 minutes. When I then press <kbd>ctrl</kbd><kbd>c</kbd> to exit it, the following gets printed:

```
fs.js:439
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^
Error: EMFILE, too many open files '/Users/emlai/.meteor/packages/meteor-tool/.1.3.2_4.lbyo5v++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/tools/static-assets/server/shell-server.js'
    at Object.fs.openSync (fs.js:439:18)
    at Object.fs.readFileSync (fs.js:290:15)
    at Object.Module._extensions..js (module.js:473:44)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at /tools/runners/run-app.js:146:7
    at /Users/emlai/.meteor/packages/meteor-tool/.1.3.2_4.lbyo5v++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/tools/tool-env/cleanup.js:24:7
    at Array.forEach (native)
    at Function._.each._.forEach (/Users/emlai/.meteor/packages/meteor-tool/.1.3.2_4.lbyo5v++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/underscore/underscore.js:79:11)
    at /Users/emlai/.meteor/packages/meteor-tool/.1.3.2_4.lbyo5v++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/tools/tool-env/cleanup.js:23:7
    at Object.exports.noYieldsAllowed (/tools/utils/fiber-helpers.js:37:14)
    at runHandlers (/Users/emlai/.meteor/packages/meteor-tool/.1.3.2_4.lbyo5v++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/tools/tool-env/cleanup.js:20:16)
[...]

PR #7915, opened 2016-10-12
Change "max open files" logic in launcher

abernix:
This makes a couple changes to the `ulimit` calls of the Meteor launch script.

This will hopefully be less necessary with future changes to the watcher, but as long as polling mechanisms are in place for the files (e.g. `file.watchFile`), this seems helpful as changes to `ulimit -n` could be necessary.  This POSIX-compatible PR does two things:

* Don't lower `ulimit` max open files if it's already higher than desired  …
 * Previously, if the default ulimit was set to a value higher than 16384, it was lowering it.  While this doesn't do much for users with default OS settings, those that have modified their ulimit settings will have their settings maintained.
* Increase the value of the first `ulimit` max open files change attempt  …
 * Make the first attempt 32768 instead of 16384 since some users have large numbers of files or node_modules trees (meteor/meteor#6952).  This hasn't been raised since Meteor started watching many more files nor since the addition of native NPM.

I know the second bullet point wasn't discussed, but I thought it was close enough to bundle with little concern for side effect.

I think another option for this would be to use the system hard limits (Linux: `ulimit -Hn`, Mac: `sysctl -n kern.maxfiles`) and just set the limit to that, [like FB's `watchman` does](https://github.com/facebook/watchman/blob/e59f2f52e80147ee6639579efacae4b892af932a/listener.cpp#L611-L671).

### DIFF
--- a/meteor
+++ b/meteor
@@ -113,16 +113,24 @@ fi
 DEV_BUNDLE="$SCRIPT_DIR/dev_bundle"
 METEOR="$SCRIPT_DIR/tools/index.js"
 
+# Set the nofile ulimit as high as permitted by the hard-limit/kernel
+if [ "$(ulimit -Sn)" != "unlimited" ]; then
+    if [ "$(uname -s)" = "Darwin" ]; then
+        maxfilesuse="$(sysctl -n kern.maxfilesperproc)"
+    else
+        maxfilesuse="$(ulimit -Hn)"
+    fi
 
-# Try higher ulimit maxfiles settings until permitted by the kernel
-if [ "$(ulimit -n)" != "unlimited" ] && ! [ "$(ulimit -n)" -gt 32768 ] ; then
-    ulimit -n 32768 > /dev/null 2>&1 || \
-    ulimit -n 16384 > /dev/null 2>&1 || \
-    ulimit -n 8192 > /dev/null 2>&1 || \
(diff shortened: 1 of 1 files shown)


benjamn:
I'm happy to merge this, but setting the limit to the max seems ideal to me. Any reason to hold back?

abernix:
I agree.  This was just the "I'm sure this will work" PR.

I would just want to do some additional testing.  My intuition tells me that something along the lines of `value-to-set = sysctl -n kern.maxfiles || ulimit -Hn` would provide sufficient details, but I hadn't looked beyond that and tested on various platforms.  

I can try to visit this in the next couple days.

abernix:
Ok, I went all in and just raise it to what's determined to be the max.

I tested this logic on Ubuntu 14.04, 16.04, busybox, Debian Jessie, Alpine, macOS Sierra, OS X Mountain Lion, OS X Lion and whatever the build-farm Mac is – I think it's Lion too.

I don't know any environment would actually return `unlimited` as a soft limit.  I couldn't find one to test on, but this should still work for that too (by not doing anything) since I don't know any OS where you could actually set a soft nofile limit to unlimited.

/cc @benjamn

benjamn:
Looks reasonable to me! As you know, file watching is more important than ever in 1.4.2, so anything that makes it more responsive and/or reliable is worth a try.

PR #7921, opened 2016-10-13
Move History entry about #7821 higher explain further

abernix:
Just a proposal, but happy to be rejected.

Currently the history entry for meteor/meteor#7821 states that it prints a warning but is actually a breaking change for anyone running the `meteor` tool as `root` right now (for example, CI environments with auto-deploy, etc.).  This moves that history entry higher (a bit more prominent) and explains why.

### DIFF
--- a/History.md
+++ b/History.md
@@ -20,6 +20,12 @@
   been updated to version 2.2.10.
   [#7780](https://github.com/meteor/meteor/pull/7780)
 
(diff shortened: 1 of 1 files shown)


benjamn:
The one test that still hasn't passed will eventually time out, and should be fixed by https://github.com/meteor/meteor/pull/7668/commits/ec34a4ecc6d4a9891ccaa382538f8972d3f2fa7d.

Issue #7584, opened 2016-08-04
Twitter force_login and Facebook authType not working?


MichaelJCole:
This is about OAuth logins, via these [commits](https://github.com/meteor/meteor/pull/6987)

Twitter offers a [force_login parameter](https://dev.twitter.com/oauth/reference/get/oauth/authenticate) to ask the user to re-enter their password.  This is useful for selecting a twitter account in the case someone has more than one account.  It's coded in Meteor [here](https://github.com/meteor/meteor/blob/5d5cfca5b005b3b25f44ecb0284b39ee4336b1a1/packages/twitter/twitter_client.js#L42)

Facebook offers a similar feature through [auth_type](https://developers.facebook.com/docs/facebook-login/reauthentication) which seems to be encoded in Meteor as [authType](https://github.com/meteor/meteor/blob/5d5cfca5b005b3b25f44ecb0284b39ee4336b1a1/packages/facebook/facebook_client.js#L41)

I'm using the [meteor-link-accounts package](https://github.com/yubozhao/meteor-link-accounts/blob/master/core-services/twitter.js#L16) but neither request seems to be adding these parameters to the URL.  (The link shows the code that calls into Meteor Framework).

Is there an example for this?

Here's a [reproduction](https://github.com/MichaelJCole/meteor-7584-force_login).

The expected behavior is that force_login is passed to twitter, but that doesn't appear to be happening.

abernix:
Thanks for opening this issue.

For Facebook's `auth_type`, you're absolutely right – it's implemented incorrectly as `authType` in Meteor.  I don't think it has ever been `authType` [according to the Facebook docs](https://developers.facebook.com/docs/facebook-login/reauthentication) (as you linked to).  Additionally, a cursory look through their API history didn't show anything about it being changed.  It's possible that somewhere in their full JS API (which is not used by Meteor) it's possible that it converts it automatically from `authType` to `auth_type`, but I can't tell.  I've submitted PR #7928 to fix that.

***
As far as Twitter:

Yeah, the implementation of `force_login` is broken there too. *sigh*   I'll try to check it out next week.

abernix:
@MichaelJCole Again, thanks for reporting this.  This should be fixed in the release candidate right now (specifically, `1.4.2-rc.1`).  Hopefully this is finally working properly!

[...]

PR #5693, opened 2015-11-23
Handle force_login parameter in Twitter package

nuvipannu:
Adding support for force_login:  
https://dev.twitter.com/oauth/reference/get/oauth/authenticate :sparkle:

PR #1566, opened 2013-11-05
Add Twitter Options: forceLogin and requestPermissions

timhaines:
This PR preserves backwards compatibility. It adds two options to the Twitter package. When requesting a credential, there are now two options.

The forceLogin option means force_login gets passed through to Twitter, telling it to present the authentication form as a login form, disregarding any user currently signed in to Twitter.com at the time.

The requestPermissions option allows an app to specify a lower permission level if it doesn't want use the one the app is configured with at dev.twitter.com. Valid options are 'read' and 'write'.

Closes #616, #1564

PR #7933, opened 2016-10-18
Support additional params on oAuth1 authorize URL

abernix:
Ok, please bear with this description...

This PR is meant to fix a previous broken implementation of the `force_login` parameter for Twitter (link below) which was reported in #6987.   This appears to be first attempted by very-old PR #1566 as a two-in-one-PR but turned into abandonware and never got merged.

#5693 (another attempt) also never landed but was resurrected into #6987 which did land but unfortunately, there is no way that it could work as the `force_login` parameter would be lost when passed to the Meteor oAuth server before the redirect to the actual service (e.g. Twitter, other oAuth1 implementations) which needs to see this param.

So, with a little help from code from #2404 and using the previously-supported ability to pass a function (versus a string) for an oAuth1 URL, this commit implements (and relocates) a function which safely applies whitelisted params to that URL.

This is slightly different implementation than desired in the original PR, but I'm not sure the facilities I mentioned above existed at that point (or weren't acknowledged).

This introduces a `twitter_common.js` file shared between server and client which indicates which Twitter-supported params are permitted on the authorize step.  The two params which Twitter supports right now are `force_login` and `screen_name`.  (See: https://dev.twitter.com/oauth/reference/get/oauth/authenticate)

This commit removes the non-functional implementation of `force_login` introduced by meteor/meteor#6987 and implements it via the aforementioned method.

As a precaution (and since neither `ecmascript` nor `es5-shim` are used by this package), I stuck with JS ES3.

Closes meteor/meteor#7584
Relates to meteor/docs#94

### DIFF
--- a/packages/twitter/twitter_client.js
+++ b/packages/twitter/twitter_client.js
@@ -37,11 +37,13 @@ Twitter.requestCredential = function (options, credentialRequestCompleteCallback
   }
 
   // Support additional, permitted parameters
-  var addlParams = Twitter.validParamsAuthenticate;
-  for (var i = 0, len = addlParams.length; i < len; i++) {
-    if (options && options[addlParams[i]]) {
-      loginPath += "&" + addlParams[i] + "=" + encodeURIComponent(options[addlParams[i]]);
-    }
+  if (options) {
+    var hasOwn = Object.prototype.hasOwnProperty;
+    Twitter.validParamsAuthenticate.forEach(function (param) {
+      if (hasOwn.call(options, param)) {
+        loginPath += "&" + param + "=" + encodeURIComponent(options[param]);
+      }
+    });
   }
 
   var loginUrl = Meteor.absoluteUrl(loginPath);
--- a/packages/twitter/twitter_client.js
+++ b/packages/twitter/twitter_client.js
@@ -1,5 +1,3 @@
-Twitter = {};
-
 // Request Twitter credentials for the user
 // @param options {optional}  XXX support options.requestPermissions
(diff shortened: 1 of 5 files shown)


packages/twitter/twitter_client.js
```diff
@@ -38,9 +36,12 @@ Twitter.requestCredential = function (options, credentialRequestCompleteCallback
...
 
-  // Handle force login (request the user to enter their credentials)
-  if (options && options.force_login) {
-    loginPath += "&force_login=true";
+  // Support additional, permitted parameters
+  var addlParams = Twitter.validParamsAuthenticate;
+  for (var i = 0, len = addlParams.length; i < len; i++) {
+    if (options && options[addlParams[i]]) {
+      loginPath += "&" + addlParams[i] + "=" + encodeURIComponent(options[addlParams[i]]);
+    }
```
benjamn: `Twitter.validParamsAuthenticate.forEach(function (param) { ... })` would be fine here, I think.

benjamn:
Mostly just wanted to try pushing a commit to a PR source branch :)

abernix:
Haha.  Such a great feature!  It even [triggered my Semaphore CI build](https://semaphoreci.com/abernix/meteor/branches/feature-fix-twitter-auth-params).  Rest assured, they failed miserably – but they're not working great on `devel` right now.

Anyhow, the `Array#forEach` would only be an issue with IE7 and IE8 which don't support it, but I'm not even sure if anything _would_ work in them.  Either way, it should be solved if the dev installed `es5-shim`.

PR #7934, opened 2016-10-18
Adding gap: into CSP

ramezrafla:
In UIWebView on iOS10 we get CSP failure to load gap://ready, this resolves it on both simulator and device. This is an urgent fix.


ramezrafla:
Thanks!

PR #7945, opened 2016-10-21
Update accounts-ui-viewer.html

betheazdavida:


abernix:
Thanks for filing a PR.  What is this change meant to accomplish and why is it necessary?

abernix:
Based on the other activity by that user, I believe this was opened in error.  Closing.

Issue #5626, opened 2015-11-06
Meteor can't get assets with names containing special chars


arboleya:
Yes, I know it doesn't look good to have names like `private/images/maça verde ou limão.png`.

``` javascript
Assets.getBinary('images/maça verde ou limão.png')
 // throw error EACESS file not found..
```

However I think Meteor should find the assets regardless of it's name (with or not special chars).

Makes sense?

zol:
Hey @arboleya . This looks like a valid issue. Could you please save collaborators some time and create a small reproduction repo as described in: https://github.com/meteor/meteor/wiki/Contributing-to-Meteor#reporting-a-bug-in-meteor

arboleya:
Hi @zol, somehow I've lost track of this. I'll put up a repo. :v:

arboleya:
@zol here it is:
https://github.com/arboleya/meteor-assets-bug

zol:
Thanks @arboleya - I've confirmed your repro. Not sure on when we'll have time to work on a fix, PR's are encouraged!

hwillson:
I've dug into this a bit further; this is a [Unicode normalization](http://unicode.org/reports/tr15/) issue. Here's a quick example of where this is failing in the source.

Let's say we're making a `Assets.getText('maça verde.txt')` call. Looking in [boot.js](https://github.com/meteor/meteor/blob/cca97e2784831b4cd966a688842a892d7749d029/tools/static-assets/server/boot.js#L228-L229):

```
...
if (!fileInfo.assets || !_.has(fileInfo.assets, assetPath)) {
  _callback(new Error("Unknown asset: " + assetPath));
...
```

In this conditional, `fileInfo.assets` returns an object that looks like:

```
{ 
  'maça verde.txt': 'assets/app/macaverde.txt' 
}
```

and `assetPath` returns `maça verde.txt`. So the `_.has` call appears to look like:

```
_.has({ 'maça verde.txt': 'assets/app/macaverde.txt' }, 'maça verde.txt')
```

This should return true, but always returns false. This is because the two `maça verde.txt` strings are actually different thanks to Unicode code points; they're essentially lookalikes.

[...]

PR #7963, opened 2016-10-26
Load `program.json` using the Unicode Normalization Form of the loaded JSON string

hwillson:
This is to help address issue #5626. When the `program.json` file is loaded, we're levearging ES2015's [`String.prototype.normalize()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) capabilities to make sure the loaded JSON string uses Unicode Normalization Form.


abernix:
Thanks for working on this, @hwillson.

The need for this PR makes sense for me, but is it possible that we should also be doing a `normalize` on the `assetPath` that is being looked up and not just the dictionary that it's being looked up in?

For example, [this `assetPath`](https://github.com/meteor/meteor/blob/cca97e2784831b4cd966a688842a892d7749d029/tools/static-assets/server/boot.js#L251) and [this `assetPath`](https://github.com/meteor/meteor/blob/cca97e2784831b4cd966a688842a892d7749d029/tools/static-assets/server/boot.js#L226).  Without normalizing both sides of the check, it seems like we could still fail to hit an asset path in some circumstances (maybe different OS or keyboard combinations?).  Am I missing something?

hwillson:
Hi @abernix - thanks for your review/comments. No, you're not missing anything - I thought the same thing. In the end I decided not to add the extra `normalize` calls, as they didn't seem to be necessary. Here's a quick recap of my thinking at the time:

In `server-json.js`, the `fs.readFileSync(serverJsonPath, 'utf8')` call (without being normalized) returns the following string (when showing character escapes):

```
mac\u0327a verde.txt
```

When normalizing the string using `fs.readFileSync(serverJsonPath, 'utf8').normalize('NFC')`, the same string (with character escapes) looks like:

```
ma\xE7a verde.txt
```

So the non-normalized version uses `c\u0327` which is made up of 2 code points: [U+0063 LATIN SMALL LETTER C](https://codepoints.net/U+0063) and [U+0327 COMBINING CEDILLA](https://codepoints.net/U+0327). The normalized version uses `\xE7` which is made up of one code point: [U+00E7 LATIN SMALL LETTER C WITH CEDILLA](https://codepoints.net/U+00E7).

When testing/debugging this issue, I was continuously monitoring the `assetPath` value used in `boot.js`, watching the character escapes used in the string. In all  cases the `assetPath` value always uses a single code point to represent the accented character. So `assetPath` is always already in the following format:

```
ma\xE7a verde.txt
```

I couldn't find a case where `assetPath` ever used 2 code points, so the extra `normalize` call didn't seem to be needed.

The above all being said, as I type this I'm starting to lean towards adding the extra `normalize` calls. There isn't really a performance hit, we're not really adding extra complexity, and it's a nice extra failsafe to have. Even though I couldn't find a case where `assetPath` ever used 2 code points (in my relatively quick testing), programmatically it's possible, so we should probably be a bit more defensive.

I'll get this PR updated accordingly. Thanks again!

hwillson:
Hi @abernix - I've updated the `Assets` code/tests to include string normalization on the `assetPath` when used in `getText`, `getBinary` and `absoluteFilePath`. This should now help address the assets unicode issue from both sides of the equation. Thanks!

abernix:
Thanks, @hwillson.  I _believe_ this PR is good to go right now (subject to further input from core), and I'd assume it'll get review/release by MDG when 1.4.3 work has started (i.e. after 1.4.2.x's) :smile:

abernix:
Ok, follow-up to my previous statement (sorry, just trying to head off any questions):

This will work right now, but do you think we can safely move the `.normalize` into the `files.convertToStandardPath` function?  That seems a bit dangerous at first thought, but I'm not sure how OSes would react to a path that is essentially the same, but simply the other way around.

Alternatively, I'd propose we create a new function in [`mini-files.js`](https://github.com/meteor/meteor/blob/devel/tools/static-assets/server/mini-files.js#L54) which does this `normalize` so we can specify the `NFC` setting in a single place.  Perhaps a `files. unicodeNormalizePath` function?

tools/tests/apps/unicode-asset-app/index.js
```diff
@@ -0,0 +1,22 @@
...
+  // c\u0327 = 2 code points: U+0063 LATIN SMALL LETTER C and U+0327 COMBINING
+  //           CEDILLA
+  // \xE7 = 1 code point: U+00E7 LATIN SMALL LETTER C WITH CEDILLA
+  const filenames = [
+    'maça verde.txt',
+    'mac\u0327a verde.txt',
+    'ma\xE7a verde.txt',
+  ];
+
+  for (let i = 0; i < filenames.length; i++) {
```
abernix: I think this and the following `for` loop should use `Array.prototype.forEach` instead.

tools/tests/apps/unicode-asset-app/index.js
```diff
@@ -0,0 +1,22 @@
...
+    console.log(`${i + 1} - getText: ${Assets.getText(filenames[i])}`);
+  }
+
+  for (let i = 0; i < filenames.length; i++) {
```
abernix: Same with `Array.prototype.forEach`.  But these tests look real good!

hwillson:
@abernix, we think along very similar lines! I originally started to make my changes in `convertToStandardPath`, but stopped as I was a bit worried about affecting other parts of the Meteor codebase that I haven't been testing. I was trying to find a good place to put the shared code changes though, so creating a new function in `mini-files` sounds great to me! I'l adjust - thanks!

tools/tests/apps/unicode-asset-app/index.js
hwillson: I originally wrote them with `forEach`, then wanted to output an index, so I switched to a normal `for` loop. All of this was because I can't seem to EVER remember that `forEach` methods can have callbacks with an index parameter passed in! Grr - need more coffee - I'll get this changed!

hwillson:
I've moved the common `normalize` code into `mini-files.js`, and updated the tests accordingly. I added a test for the new `unicodeNormalizePath` helper function, but wasn't quite sure where it should go. For now I've added it into `tools/tests/assets.js`, but let me know if that should change. Thanks!

tools/static-assets/server/server-json.js
```diff
@@ -1,4 +1,6 @@
 var fs = require("fs");
 var path = require("path");
 var serverJsonPath = path.resolve(process.argv[2]);
-module.exports = JSON.parse(fs.readFileSync(serverJsonPath, 'utf8'));
+module.exports = JSON.parse(
+  fs.readFileSync(serverJsonPath, 'utf8').normalize('NFC')
```
abernix: I guess this isn't using the new `unicodeNormalizePath` because it's not a path?

hwillson: Exactly - we could always consider making the `unicodeNormalizePath` function a little more generic, so it's open to all strings. It wouldn't really fit in `mini-files.js` at that point though; it would belong in a more generic string helper location. I'm not sure where that would be in the context of the build tools though.

abernix:
@hwillson This is merged into `devel` now, but it appears tests are [failing consistently](https://circleci.com/gh/meteor/meteor/tree/devel) now with:

```
  - assets: assets - unicode asset names are allowed
```

Here is one specific test: https://circleci.com/gh/meteor/meteor/3115

I saw no failures in this PR, but it does seem to be because of this.

Any ideas?

hwillson:
Thanks @abernix - very strange! I'll take a look right now and keep you posted.

hwillson:
I just synched with `devel` and ran my own CI build - it passed!

https://circleci.com/gh/hwillson/meteor/12

This is definitely a strange one ...

abernix:
@hwillson You can run just that single test locally by executing:

    ./meteor self-test "assets - unicode asset names are allowed"

That being said – it works fine for me too on OS X.  I tried on my Ubuntu box and it failed.  I guess this is a (possibly interesting) platform-related issue as the files are outputted to different places in the Ubuntu bundle than on OS X.  I'm not sure if the tests should just be changed to check for something different or if this is a legitimate (other?) issue.

benjamn:
I've submitted https://github.com/meteor/meteor/pull/8118 to revert these changes temporarily, but if we can figure out what's going on soon, I'd be happy to add a fix and un-revert the changes on that PR.

hwillson:
@abernix Yes I was running the tests locally as well, and they were/are passing. They only seemed to fail via CI. They weren't failing via my CI build, but I found out why - `./scripts/ci.sh` is hard-coded to run specific tests on specific CircleCI containers; I was only using one container, which means only the warehouse self-tests were running. I've added more, and am now able to replicate the issue via CircleCI (see https://circleci.com/gh/hwillson/meteor/14).

> I'm not sure if the tests should just be changed to check for something different or if this is a legitimate (other?) issue.

Actually, the good news is that the tests are working properly - it's showing a real issue, which only occurs on Linux. On OSX the `program.json` contains the following:

```
{
      "node_modules": "npm/node_modules",
      "sourceMap": "app/app.js.map",
      "path": "app/app.js",
      "assets": {
        "maça verde.jpg": "assets/app/macaverde.jpg",
        "maça verde.txt": "assets/app/macaverde.txt"
      }
    }
```

whereas on Linux it contains:

```
{
      "node_modules": "npm/node_modules",
      "sourceMap": "app/app.js.map",
      "path": "app/app.js",
      "assets": {
        "maça verde.jpg": "assets/app/maaverde.jpg",
        "maça verde.txt": "assets/app/maaverde.txt"
      }
    }
```

The **c** is missing. It looks like this issue is coming from the isobuild builder, which is where I'm focusing right now. 

Thanks for reverting the changes @benjamn - I had to step away for a few meetings, but I'm back on this now. More details soon - thanks guys!

hwillson:
Arg - found it. The problem is happening in the [`_sanitize` function of `tools/isobuild/builder.js`](https://github.com/meteor/meteor/blob/1f883c6f6d2493aaaf0bb2a0da2f720ed944bacc/tools/isobuild/builder.js#L172):

```
part = part.replace(/[^a-zA-Z0-9._\:-]/g, '');
```

On Linux just before this line `part` contains `maça verde.txt`. Right after `part` contains `maaverde.txt`.

On OSX just before this line `part` contains `maça verde.txt`. Right after `part` contains `macaverde.txt`. This is the same Unicode normalization issue, just in another location. Changing the above line to:

```
part = part.normalize('NFC').replace(/[^a-zA-Z0-9._\:-]/g, '');
```

Fixes this issue, so `part` then contains `maaverde.txt` on both Linux and OSX. I'll get this patched - where would you guys prefer to see my changes (since this branch was already merged)?

benjamn:
Despite what I said earlier, I would just make a new pull request and I'll close #8118.

hwillson:
Great - will do!

benjamn:
@hwillson I'd like to get tests passing again sooner rather than later, so I've merged #8118. When you have your fix ready, just `git revert` 5754f8ef3ea5222dd5ab601dd91f130d0092646e and include that in another pull request, please!

hwillson:
@benjamn Certainly - will do; the patch I outlined previously fixed the problem, but I wanted to make sure it was the proper fix. I've found a better place to make the necessary change, which will prevent this issue from happening in others areas of the codebase as well. I'll outline all of this in a new PR I'll be submitting shortly. Thanks again!

abernix:
I thought I'd explain what happened when this was merged into `devel` and suddenly started failing when the tests appeared to have "passed" before.  I've run into this before.

The tests above (for example, the one run on c93756a710c010f976cdb1b49e874b61691a5ad1) passed ✅ because they weren't run on the Meteor CircleCI environment (which has 8 containers), but instead run on @hwillson's CircleCI (which had 1 container).  You'll note this if you look at the URL in the "Details" link of the CI results:

* [https://circleci.com/gh/**hwillson**/meteor/9](https://circleci.com/gh/hwillson/meteor/9)

Essentially, 7 of the containers didn't even run.

This is explained in [this CircleCI docs](https://circleci.com/docs/fork-pr-builds/):

> Note: If otherdev/yourproject is explicitly configured to build on CircleCI in its own right, then we will not run builds in the yourorg/yourproject context for PRs from the fork

I find this to be a rather annoying (and slightly insecure), but it is what it is.  I had to delete my CircleCI configuration for meteor to avoid running it on my own account (or pay $$$ and upgrade to 8 containers).

Anyhow, I have a fix to the ci.sh script stashed somewhere from when I first ran into this same problem.  I'll submit a PR for it as I think this could be considered an improvement to contributor experience.

hwillson:
@abernix I found the same issue, and just finished making a patch for this! 😄 I have it ready to go and was going to open a new issue + PR for the `ci.sh` changes. Basically I'm just adding a quick fix to first see how many available CI nodes there are; if only 1, all self-tests get run on one node. Should I go ahead and submit my issue/PR or hold off for your changes? Thanks - and sorry about this again guys!

hwillson:
Just to add, you can see it working here: https://circleci.com/gh/hwillson/meteor/22

abernix:
@hwillson I think I have even one better.  And no worries at all, this is how things get improved!  Truth be told, your test found what would have been a real bug and did eventually get run and fail at some point in the pipeline.

(And when I say one better, I mean, since `meteor` is free open-source, you can actually raise your CircleCI configuration to 4!  My change will support that.)

hwillson:
Awesome - yes, I quickly raised my CI nodes to 2 yesterday and re-ran, which immediately showed the failed tests (since the unicode asset tests were run on the second node). I'll hold off on my change - something more dynamic that leverages the available number of nodes would be great. Yah for tests working!

PR #7970, opened 2016-10-26
Add support for frame-ancestors CSP option in browser-policy

MichaelOber:
This Pull Request adds support for for the frame-ancestors CSP option in the browser-policy package. Right now, there is no easy way to allow and/or restrict specific domains from framing your app. Using the X-Frame-Options header has the following problems:
- You are only allowed to restrict all domains, only allow same domain, or only allow one other domain. So, if you need to allow multiple domains to frame your app, you can't.
- The X-Frame-Options header does not work in Chrome or Safari. So even if you want to allow another domain to frame your app, you can't guarantee that it will work in those browsers.

As a result of the above, to allow other domains to frame your app, you have to allow _all_ domains to frame your app. 

However, there is a [frame-ancestors CSP](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#frame-ancestors) that allows you to specify which domains can frame your application. This pull request adds an API to do that. 

Calling `BrowserPolicy.content.allowFrameAncestorsOrigin("https://foo.com/"); ` will allow `foo.com` to frame your application. Calling `BrowserPolicy.content.disallowFrameAncestors()` will forbid other domains from framing your application.

A test has been added to verify that the frame-ancestors CSP is added correctly.

There are also some documentation changes to make note of the new policy.

### DIFF
--- a/packages/browser-policy-content/browser-policy-content.js
+++ b/packages/browser-policy-content/browser-policy-content.js
@@ -17,7 +17,7 @@
 // disallowEval()
 //
 // For each type of content (script, object, image, media, font, connect,
-// style), there are the following functions:
+// style, frame, frame-ancestors), there are the following functions:
 // allow<content type>Origin(origin): allows the type of content to be loaded
 // from the given origin
 // allow<content type>DataUrl(): allows the content to be loaded from data: URLs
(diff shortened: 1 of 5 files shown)


zol:
Lgtm, I'm going to merge this, thank you.

Issue #7959, opened 2016-10-25
You are attempting to run Meteor as the "root" user


rclai:
Uh... Why is this error happening? I did not even upgrade any of my Meteor apps at all. Why would this shell flow control be getting hit when I haven't even updated to the new version? This must have come up as a result of you pushing out this latest build.

Anyway, running this:

`meteor run --unsafe-perm --port 3002`

Does not fix the issue, and it's because I'm running an old version of Meteor and all I get is this:

```
You have run Meteor as root. Your permissions in your app directory will be incorrect if you ever attempt to perform any Meteor tasks
as your non-root user. You probably didn't want this, but you can fix it by running the following from the root of your project:

sudo chown -Rh <username> .meteor/local

--unsafe-perm: unknown option.                
Usage: meteor run [target..] [options]

Searches upward from the current directory for the root directory of a
Meteor project, then runs that project in local development
mode. You can use the application by pointing your web browser at
localhost:3000. No internet connection is required.

Whenever you change any of the application's source files, the changes
are automatically detected and applied to the running application.

The application's database persists between runs. It's stored under
the .meteor directory in the root of the project.

If you have added a platform to your app with 'meteor add-platform', you can
pass one of the following targets as an argument to this command.

Targets:
  android          Run on the Android emulator.
  android-device   Run on a connected Android device.
  ios              Run on the iOS simulator.
  ios-device       Open Xcode with the iOS project for this app, where you can
                   run your app on a connected iOS device.

Options:
  --port, -p       Port to listen on (instead of the default 3000). Also
                   uses port N+1 and a port specified by --app-port.
                   Specify as --port=host:port to bind to a specific interface.
  --debug-port     Specify a port to enable server-side debugging. The
                   server will be paused at startup, waiting for incoming
                   connections from debugger clients on the specified port.
  --mobile-server  Location where mobile builds connect to the Meteor server.
                   Defaults to your local IP and the port that the Meteor
                   server binds to. Can include a URL scheme (for
                   example, --mobile-server=https://example.com:443).
  --production     Simulate production mode. Minify and bundle CSS and JS files.
  --raw-logs       Run without parsing logs from stdout and stderr.
  --settings       Set optional data for Meteor.settings on the server.
  --release        Specify the release of Meteor to use.
  --verbose        Print all output from builds logs.
  --no-lint        Don't run linters used by the app on every rebuild.
  --allow-incompatible-update   Allow packages in your project to be upgraded or
[...]

Issue #7559, opened 2016-08-03
Explore an option for `meteor config` similar to `npm config`


abernix:
It would be nice if there was a way to provide configuration options for **non-sensitive** information to the `meteor-tool`, such as `--port`, the proposal for `--skip-release-check` in #7445, and other development items which are frequently only configurable via increasingly-clunky `ENV` variables.  Ideally this would be done using a tool similar to [`npm config`](https://docs.npmjs.com/cli/config) and its [`npmrc`](https://docs.npmjs.com/files/npmrc) derivitive so it could also have "global" preferences where options could be set for all your Meteor projects, including the ability to have a version controlled file in a project that would work across multiple developers, CI builds, etc.

I believe this is only relevant for the development environment specifically for using the `meteor` command line tool, **not for production**, which I believe is fine the way it is using `Meteor.settings` + `settings.json` or the `METEOR_SETTINGS` environment variable.  

The [`config` NPM package](https://www.npmjs.com/package/config) is the first thing that comes to mind, but I'd love to see other proposals.

aaroncalderon:
If meteor where to respect the `~\.*rc` convention other tools have implemented (e.g. git, npm, bower, tsd) and use said configurations we would not have to re-invent the wheel.

My point being, lets honor `~\.npmrc` and the issues we all face with runing `meteor npm install` behind a proxy. And we could also have meteor implement a similar approach, we would be able to do without those environment variables.

abernix:
@aaroncalderon This feature request was not meant to have anything to do with changing NPM settings in relation to Meteor.  It's simply saying that Meteor has certain parameters (of its own), which would be nice to configure with a tool roughly similar to `npm config`.  Meteor currently has no global "tool" config and this is what that was aiming to be.

Briefly though: If you are using `npm config` (without the `--global` flag) in the root of the Meteor project, you will not get the results you are looking for since Meteor's functional root as far as NPM is concerned is not the same – as you have found.  That being said (I haven't tested this), it's plausible that if you use `npm config set proxy <proxy> --global` Meteor would pick it up without you having to go to this effort.  Additionally, I'm fairly sure that the `HTTP_PROXY` (and similar/related) environment variables work properly with Meteor.  Either way, this is a discussion for another issue/the forums as this is about something else entirely.

hwillson:
[...]

PR #7971, opened 2016-10-26
Respect $METEOR_UNSAFE_PERM in addition to --unsafe-perm.

benjamn:
Propagating environment variables to child processes is easier than (re)passing explicit command-line flags, so `--unsafe-perm` is now a shorthand for setting `process.env.METEOR_UNSAFE_PERM = "true"`.

Allowing the use of an environment variable is important for `meteor npm`, since we don't parse or pass through Meteor-specific command-line arguments when running `meteor npm`.

When `$METEOR_UNSAFE_PERM` is set, its value should be propagated to any npm commands via the `$NPM_CONFIG_UNSAFE_PERM` variable.

Helps with #7959.
Follow-up to #7821.

### DIFF
--- a/tools/cli/main.js
+++ b/tools/cli/main.js
@@ -768,33 +768,40 @@ Fiber(function () {
     rawArgs.push(term);
   }
 
-  // Prevent running meteor as root
-  if (process.getuid) {
-    // On UNIX platforms.
-    if (process.getuid() === 0){
-      if (_.has(rawOptions, '--unsafe-perm')) {
-        // Meteor is running as root and has the --unsafe-perm flag, just notice.
-        Console.info("");
-        Console.info("You have run Meteor as root.",
-        "Your permissions in your app directory will be incorrect if you ever attempt to perform any Meteor tasks as your non-root user.",
-        "You probably didn't want this, but you can fix it by running the following from the root of your project:");
-        Console.info("");
-        Console.info("sudo chown -Rh <username> .meteor/local");
-        Console.info("");
-        // Fix issue when running `sudo meteor --unsafe-perm`, --unsafe-perm: unknown option.
-        delete rawOptions['--unsafe-perm'];
-      }
-      else {
-        // Meteor is running as root without the --unsafe-perm flag, notice and stop.
-        Console.error("");
-        Console.error("You are attempting to run Meteor as the \"root\" user.",
-        "If you are developing, this is almost certainly *not* what you want to do and will likely result in incorrect file permissions.",
-        "However, if you are running this in a build process (CI, etc.) or you are absolutely sure you know what you are doing,",
-        "add the `--unsafe-perm` flag to this command to proceed.");
-        Console.error("");
+  if (_.has(rawOptions, "--unsafe-perm")) {
+    process.env.METEOR_UNSAFE_PERM = "true";
+    delete rawOptions["--unsafe-perm"];
+  }
 
-        process.exit(1);
-      }
+  // Prevent running meteor as root on UNIX platforms.
+  if (process.getuid &&
+      process.getuid() === 0) {
+    const unsafePerm = !! (
+      process.env.METEOR_UNSAFE_PERM &&
+      JSON.parse(process.env.METEOR_UNSAFE_PERM));
+
+    if (! unsafePerm) {
+      // Meteor is running as root without $METEOR_UNSAFE_PERM, notice and stop.
+      Console.error("");
+      Console.error(
+        "You are attempting to run Meteor as the root user.",
+        "If you are developing, this is almost certainly *not* what you want to do and will likely result in incorrect file permissions.",
+        "However, if you are running this command in a build process (CI, etc.), or you are absolutely sure you know what you are doing,",
+        "set the $METEOR_UNSAFE_PERM environment variable to proceed."
+      );
+    }
+
+    Console.info("");
+    Console.info(
+      "Even with $METEOR_UNSAFE_PERM or --unsafe-perm, permissions in your app directory will be incorrect if you ever attempt to perform any Meteor tasks as a non-root user.",
+      "If you need to fix your permissions, run the following command from the root of your project:"
+    );
+    Console.info("");
+    Console.info(Console.command("  sudo chown -Rh <username> .meteor/local"));
+    Console.info("");
+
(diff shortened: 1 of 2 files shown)


benjamn:
If the `npm` propagation is desirable, then I think it's a good thing if `$METEOR_UNSAFE_PERM` is propagated to child/springboard processes.

tools/cli/main.js
```diff
@@ -616,7 +616,7 @@ Fiber(function () {
   // tight timetable for 1.0 and there is no advantage to doing it now
   // rather than later. #ImprovingCrossVersionOptionParsing
 
-  var isBoolean = { "--help": true, "--unsafe-perm": true };
```
mitar: I think we should keep command-line flag. I think people know much better how to set that than how to run `METEOR_UNSAFE_PERM=true meteor`. But maybe `unsafe-perm` should just set `METEOR_UNSAFE_PERM` and this is it. :-)

abernix: I agree that it's cleaner; I'm not a fan of environment variables in general, but I guess I don't care either way.  However, this will break anyone (again) who has made the change today to use `--unsafe-perm`.

abernix: Propagation to NPM might be useful/wise though!

benjamn: I'm going to restore the `--unsafe-perm` option, but have it just set `process.env.METEOR_UNSAFE_PERM = "true"`.

benjamn:
Updated to keep supporting `--unsafe-perm`. Please review when you can!

tools/cli/main.js
```diff
@@ -768,33 +768,40 @@ Fiber(function () {
...
+      // Meteor is running as root without $METEOR_UNSAFE_PERM, notice and stop.
+      Console.error("");
+      Console.error(
+        "You are attempting to run Meteor as the root user.",
         "If you are developing, this is almost certainly *not* what you want to do and will likely result in incorrect file permissions.",
-        "However, if you are running this in a build process (CI, etc.) or you are absolutely sure you know what you are doing,",
-        "add the `--unsafe-perm` flag to this command to proceed.");
-        Console.error("");
+        "However, if you are running this command in a build process (CI, etc.), or you are absolutely sure you know what you are doing,",
+        "set the $METEOR_UNSAFE_PERM environment variable to proceed."
```
mitar: `or --unsafe-perm`?

Also, the name of variable should be `METEOR_UNSAFE_PERM` not `$METEOR_UNSAFE_PERM`.

abernix: My recommendation would be the same as @mitar on both counts.

I understand it's semantics and colloquial used to indicate "a variable" but in this case the `$` is not part of the variable in this case – even more so than say, PHP or Perl.  It's shell expansion syntax and isn't used for _setting_ (or even _getting_ in most languages).  However, this form of usage is used in several other places in the code (and `History.md`) – but it is more commonly used without.

abernix: Kind of relevant, and not mentioned in my quick, slumberous reply above last night was clarification as to _why_ I prefer not to use environment variables as a primary means: they are confusing and difficult to convey across different platforms (dollar signs included or not).

For example, I actually don't know how this variable affects Windows users, but the customer-support script to recommend how to set environment variables goes something like this:

> _Hi, so I need you need to set this environment variable: If you're on Windows and using PowerShell, you should run `$Env:METEOR_UNSAFE_PERM="true"`, however if you're using command prompt you should run `set METEOR_UNSAFE_PERM=true` but on Unix you'll want to run `METEOR_UNSAFE_PERM=true`._ 🤕 

It should be as easy as "You should add `--unsafe-perm` to the command"...which is why I've rallied for `--unsafe-perm` and `--no-release-check` on past PRs.  

That being said, I do think it's important to have the environment variables available in many cases (and I understand the importance/use for `meteor npm`, in this case).  Hopefully if #7559 ever comes to fruition it will support both arguments and environment variables alike, [ala `npm config`](https://docs.npmjs.com/misc/config).

abernix:
Are we sure that this flag serves the same exact purpose as NPM's `--unsafe-perm` and should be passed through?  For example, what is [quoted here](https://news.ycombinator.com/item?id=8948165) appears to be true, but slightly different behavior than what we're trying to solve (i.e. Meteor does no UID/GID switching and that's not what we're trying to disable):

> If you run npm with root, npm will run scripts with the user "nobody" by default. This means that npm doesn't run scripts as root even if you run "npm install" with root.
You can set another user instead of nobody with the "user" option and you can disable switching UID/GID using the "unsafe-perm" option but DO NOT DO THIS.

benjamn:
What if we used `METEOR_ALLOW_SUPERUSER` and `--allow-superuser`? `METEOR_ALLOW_ROOT` is tempting but "root" is kind of an overloaded term.

abernix:
I'm inclined to prefer that if my understanding of NPM's `--unsafe-perm` is correct.

I understand there are similarities in the environment that you'd use these flags (NPM `--unsafe-perm` and the Meteor equivalent), but I'm not sure Meteor functionality is the same, especially since additional relevant arguments such as [`user`](https://docs.npmjs.com/misc/config#user) and [`group`](https://docs.npmjs.com/misc/config#group) aren't available.

benjamn:
Great, agreed. I've been having trouble developing any reliable intuition about the meaning of `--unsafe-perm`, whereas `--allow-superuser` seems totally straightforward.

abernix:
Because of these newfound circumstances, I'm hesitant to push for backward compatibility, but I do hate to break the folks who have made this change already.   However, maybe it's fine to just break it right away and get it over with since the widely-used Docker images haven't updated to support this yet.  Or alternatively, perhaps a short-lived apology deprecation message until Meteor 1.4.3, or something: "Hey, we know we told you to add to `--unsafe-perm`, but we should have ...  Sorry.".

abernix:
Unfortunately, per my comments here https://github.com/meteor/meteor/issues/7959#issuecomment-257083014), this didn't work.  In particular, I failed to notice that modified `rawArgs` are completed abandoned.

See my comments here: https://github.com/abernix/meteor/commit/6716b47e7be61d22f813f1c80993f67a81031c3c

Happy to PR it if you want (or feel free to cherry-pick it), or I can rework if you'd prefer.

/cc @benjamn

PR #7985, opened 2016-10-28
add onMessage hook

akvadrako:
Like the `onConnection` hook, this is called for every DDP message received. We have a large meteor app and would like to centralise authorisation for publications and methods in the least error-prone way possible. Without any documented way to hook into the server core, the best we can do is monkey-patch `Session`. This PR provides a better way.

### DIFF
--- a/packages/ddp-server/livedata_server_tests.js
+++ b/packages/ddp-server/livedata_server_tests.js
@@ -82,7 +82,6 @@ testAsyncMulti(
   }]
 );
 
-
 Meteor.methods({
   livedata_server_test_inner: function () {
     return this.connection.id;
@@ -92,6 +91,28 @@ Meteor.methods({
     return Meteor.call('livedata_server_test_inner');
   }
 });
+
+
+Tinytest.addAsync(
+    "livedata server - onMessage hook",
+    function (test, onComplete) {
+
+        var cb = Meteor.onMessage(function (msg, session) {
+            test.equal(msg.method, 'livedata_server_test_inner');
+            cb.stop();
+            onComplete();
+        });
+
+        makeTestConnection(
+            test,
+            function (clientConn, serverConn) {
+                clientConn.call('livedata_server_test_inner');
+                clientConn.disconnect();
+            },
+            onComplete
+        );
+    }
+);
 
 
 Tinytest.addAsync(
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -547,6 +547,11 @@ _.extend(Session.prototype, {
           processNext();
         };
 
+        self.server.onMessageHook.each(function (callback) {
+          callback(msg, self);
+          return true;
+        });
+
         if (_.has(self.protocol_handlers, msg.msg))
           self.protocol_handlers[msg.msg].call(self, msg, unblock);
         else
@@ -1320,6 +1325,11 @@ Server = function (options) {
     debugPrintExceptions: "onConnection callback"
   });
 
+  // Map of callbacks to call when a new message comes in.
+  self.onMessageHook = new Hook({
+    debugPrintExceptions: "onMessage callback"
+  });
+
   self.publish_handlers = {};
   self.universal_publish_handlers = [];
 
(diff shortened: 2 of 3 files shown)


packages/ddp-server/livedata_server_tests.js
```diff
@@ -95,6 +94,28 @@ Meteor.methods({
 
 
 Tinytest.addAsync(
+    "livedata server - onMessage hook",
```
tmeasday: Can we use two spaces please?

tmeasday:
@akvadrako I'm trying to understand the use case here. What is the shape of an typical `onMessage` handler? Do you throw an exception if the message should not be auth-ed? What happens in that case?

packages/ddp-server/livedata_server_tests.js
```diff
@@ -95,6 +94,28 @@ Meteor.methods({
...
+            test.equal(msg.method, 'livedata_server_test_inner');
+            cb.stop();
+            onComplete();
+        });
+
+        makeTestConnection(
+            test,
+            function (clientConn, serverConn) {
+                clientConn.call('livedata_server_test_inner');
+                clientConn.disconnect();
```
benjamn: I believe you need to call `onComplete()` at the end of this callback. See the other `Tinytest.addAsync` tests below for examples.

akvadrako: We already call onComplete() in the handler, and that's important to be sure the handler is called. Isn't that sufficient? It seems wrong to call it twice.

akvadrako:
@tmeasday The use-case is global authorisation checking, and it could throw an error or transform the message - the goal is to return an error to the client. So, something like this (not tested):
```js
function onMessage(msg) {
  if(msg.msg === 'sub' && !checkAuth(msg.name, Meteor.userId())) {
    msg.msg = 'noauth'; // or throw new Meteor.Error('NOAUTH');
  }
}
```

mitar:
Have you seen this package and what they are doing: https://github.com/RocketChat/meteor-streamer/blob/master/packages/rocketchat-streamer/client/client.js

It seems it is possible to already hook into the stream, but will this API make it cleaner for that package to do it as well? @rodrigok, what do you think?

akvadrako:
There is also this one: https://github.com/hitchcott/meteor-method-hooks

But it's a hack based on internal APIs which the ```onMessage()``` hook would make cleaner.

mitar:
Hm, no, method hooks would not benefit from `onMessage` to my understanding. Because it simply wraps existing methods. To me that is a pretty simple an clean API to wrap, accessing public properties like `Meteor.server.method_handlers`. Doing this wrapping on message level would lead to code duplication.

akvadrako:
There are two problems with the method hooks approach.

1. ```Meteor.server``` is not public (at least not documented)
2. it's order-dependant - if a method handler is added after it wraps ```method_handlers```, it won't be wrapped.

mitar:
Why would you define methods after `Meteor.startup`? This looks like a strange design.

But this could be fixed by the package by wrapping `Meteor.methods` function. Make a pull request. :-)

akvadrako:
This patch is cleaner and also supports wrapping subscriptions.

mitar:
I do not see how would method hooks be implemented with this API. Can you show an example?

akvadrako:
Well the current design does not support ```afterMethod``` hooks in a straightforward fashion, but you could transform the message to a no-op or error to implement the before part:

```js
Method.onMessage(msg => {
  if (msg.msg === 'method' && !beforeMethod(msg.msg, msg.params)) {
    msg.msg = 'noauth'; // or throw Error
  }
});
```

It's possible to do anything in this handler, like call the method yourself and do ```session.send``` to reply, but that is a little tricky.

mitar:
You would also have to de-EJSONify `msg.params` no?

Just to be clear, I think such hooks are useful in general. But I am not yet sure if it should really be made public. But at least a hook should exist for sure.

rodrigok:
@mitar Yes, it will be much more clean to handle messages and create new message actions not supported by meteor now.

akvadrako:
It does indeed work at a lower-level that ```methods```, so ideally the session would do a bit more transformation and expose a higher-level message handling API that could be wrapped. So the _default_ message handler would be like:

```js
onMessage({ name, params, send, originalHandler }) {
  originalHandler(name, params, send);
}
```

where ```send``` is ```send(name, params)``` and params are always non-EJSON-ified.

wojtkowiak:
@akvadrako 
>```javascript
function onMessage(msg) {
  if(msg.msg === 'sub' && !checkAuth(msg.name, Meteor.userId())) {
    msg.msg = 'noauth'; // or throw new Meteor.Error('NOAUTH');
  }
}
```

How would you actually be able to send something to client from this handler? Can you prevent method or publish invocation in any way?
Will `Meteor.userId()` work in every case?

benjamn:
This seems like a good idea to me, on the whole. The implementation mimics other hooks, and I think it's fair to say that an `onMessage` hook needs to be careful about its assumptions (e.g. that `Meteor.userId()` is available), given that it's intercepting all messages.

I would also consider a separate pull request that allowed the `onMessage` hook to drop messages, if anyone thinks that would be helpful. Note that we can release new versions of `ddp-server` without doing a full Meteor release, so it's easy to keep shipping incremental improvements.

Issue #7956, opened 2016-10-25
[1.4.2-rc.4] super does not work in some cases because __proto__ is not set


macrozone:
Meteor 1.4.2-rc.4 seems to break some npm packages: https://github.com/vazco/uniforms/issues/117

I could not fully track this down, but what I found out is, that the meteor-version of babel-runtime does not set **proto** / Object.setPrototypeOf

Compare this:
meteor version: https://github.com/meteor/meteor/blob/devel/packages/babel-runtime/babel-runtime.js#L81

with this:
babel version: https://github.com/babel/babel/blob/master/packages/babel-helpers/src/helpers.js#L378

super() is transformed to this:

`var _this = (0, _possibleConstructorReturn3['default'])(this, (_class.__proto__ || Object.getPrototypeOf(_class)).call(this, props));`

When running under meteor 1.4.2, `_class.__proto__` is just an empty anonymous function, whereas under meteor 1.4.1.x `_class.__proto__` is set to a constructor-function

It might be also related to the fact, that this particular package uses static properties. I can't imagine, that inheritance just breaks without noticing. 

Anyone an idea, what's going on here?

benjamn:
I agree this is a bug. Fortunately we can fix it with an update to `babel-runtime` without doing a whole Meteor release.

macrozone:
As 1.4.2 is now released, how can i update babel-runtime separatly?

maartenbusstra:
Is this why my 2x subclassing of `Mongo.Collection` fails? It says `insert` or `remove` are `undefined` functions.

neurosoup:
Same for me, I am subclassing Mongo.Collection and have error 'insert undefined' since 1.4.2 release. Is the issue is linked ?

benjamn:
Here's the [original explanation](https://github.com/meteor/meteor/blob/f7271e214a689f5f6ae071faaecf581104ec0bed/packages/babel-runtime/babel-runtime.js#L112-L124) of why we chose not to set `__proto__`. In short, it wouldn't work in Internet Explorer <10, and any code you wrote that relied on that feature would appear to work in other browsers, but fail in older IEs, where you might not be actively testing.

The good news is that Babel now attempts to work around the limitations of older IEs by using `_class.__proto__ || Object.getPrototypeOf(_class)` instead of `Object.getPrototypeOf(_class)`, thanks to [this commit](https://github.com/babel/babel/commit/10cd6519d8f0374e8beb02a04f20da0123c52414).

[...]

PR #7995, opened 2016-10-31
Rely on npm babel-runtime instead of Meteor helper implementations

benjamn:
The more I've thought about the underlying issues involved in https://github.com/meteor/meteor/issues/7956, the more I believe the time has come for Meteor to stop providing custom implementations of `babel-runtime/helpers/*` and instead simply rely on the npm `babel-runtime` package.

This pull request adds `babel-runtime` to the default `package.json` file for new apps (a la `meteor-node-stubs`), installs those basic packages when running `meteor test-packages` (which is something we should have been doing all along), simplifies the Meteor `babel-runtime` package (it now throws an error if `babel-runtime` has not been installed), and cleans up some other logic that no longer matters.

I'm torn about whether to include this in 1.4.2.1 or wait for 1.4.3, though I believe it is the most reliable fix for #7956, which is (rightfully) in the 1.4.2.1 milestone.


benjamn:
Many tests were broken by the initial version of this pull request, since we weren't installing `babel-runtime` in the various test apps used by `meteor self-test ...`, but that should be fixable.

PR #7997, opened 2016-11-01
Refactor template of `resetPassword` email

Davidyuk:
As I see, current realization of `Accounts.emailTemplates.verifyEmail.text` are equal to result of already defined `greet` function with `"To reset your password"` param. So I suggest to use it instead.


tmeasday:
Good pickup, thanks @Davidyuk

Issue #6337, opened 2016-02-25
Confusing README of Meteor Stylus Repo


ldong:
Hi Meteor Devs,

I don't usually reach out to people like this, especially on other repo [meteor/stylus](https://github.com/meteor/stylus/). It seems to be the best way to file an issue here, since issue list of [meteor/stylus](https://github.com/meteor/stylus/) is closed.

Please, please do NOT write confusing words in the [README](https://atmospherejs.com/meteor/stylus) in [atomspherejs](http://atomspherejs.com), see the following screenshot.

<img width="1219" alt="screen shot 2016-02-24 at 8 04 38 pm" src="https://cloud.githubusercontent.com/assets/5491695/13309867/4edfcd52-db35-11e5-9143-3451b42c9fa9.png">

With the stylus package installed, **.main.styl** files in your application are
automatically compiled to CSS and the results are included in the client CSS
bundle.

Creating `.main.styl` simply does not work. So, instead of saying misleading words like that, one could say, any file ending with `.styl` will be automatically merged and bundled into one css. It was confusing to me, and I don't understand why this [file change](https://github.com/meteor/meteor/commit/d6e816e3fd84b58185e01f3f6307c0dd2d5e578d?diff=split) is not reflected in the current stylus README. Please help, and make sure you update both github [repo](https://github.com/meteor/stylus) and [atomspherejs](http://atomspherejs.com).

BTW, can meteor developers explain why you guys closed issues on [meteor/stylus](https://github.com/meteor/stylus/) repo?

Thank you,
Lin

Menda:
Makes sense to me

Slava:
The meteor/stylus repo you linked to is a fork of an npm module. We did it to submit PRs. The actual package is in this repo.

ldong:
@Slava Thank you for clarifying, I got that. I just wonder why the README was not being updated on the [atmospherejs](https://atmospherejs.com/meteor/stylus) website. i.e.
```
With the `stylus` package installed, `.main.styl` files in your application are
automatically compiled to CSS and the results are included in the client CSS
bundle.
```

And that is confusing. Can you please update that?

lorensr:
`foo.styl` works for me. I've never tried `.main.styl` or `foo.main.styl`.

abernix:
@ldong Thanks for reporting this and sorry it took so long to get around to it.  It will be updated on Atmosphere when the next release is published.  PRs are definitely welcomed for this stuff to make them land quicker.

PR #7998, opened 2016-11-01
Clarify mis-leading wording in the `stylus` README

abernix:
Make it clear that all files with the `.styl` extension are processed and not just `.main.styl` files based on the comment from meteor/meteor#6337.

I believe this `main.styl` syntax might be residue from some ancient version but Meteor definitely processes all `styl` extensions, as seen [here](https://github.com/meteor/meteor/blob/7858fb2431321a11d22cc87bbf1ea591d7594d77/packages/stylus/plugin/compile-stylus.js#L7-L10).

Fixes meteor/meteor#6337

### DIFF
--- a/packages/stylus/README.md
+++ b/packages/stylus/README.md
@@ -6,12 +6,12 @@
 simple syntax and expressive dynamic behavior. It allows for more compact
 stylesheets and helps reduce code duplication in CSS files.
 
-With the `stylus` package installed, `.main.styl` files in your application are
-automatically compiled to CSS and the results are included in the client CSS
-bundle.
+With the `stylus` package installed, files with the `.styl` extension are sent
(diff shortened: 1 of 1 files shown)


tmeasday:
Thanks @abernix. Do we need to republish the `stylus` package for this to come up on Atmosphere?

abernix:
I don't know how it works with these core packages, but `publish` with the `--update` flag would normally fix this on any normal Atmosphere package (without republishing and by only changing metadata), assuming you did it from the `release-1.4.2` branch.

Otherwise, I don't think it's worth bumping the version just for this (relatively old) text issue as it will get bumped automatically when versions are bumped for 1.4.2.1 release (again, if I'm understanding the tool release process correctly).

PR #8038, opened 2016-11-10
Allow dynamic headers and return options

mitar:
This allows dynamic headers (like per e-mail unsubscribe link inside headers), and returns information about the e-mail so that for same caller can log all e-mails or in some other way post-process sending of e-mails.

PR #8039, opened 2016-11-10
Provide a way to post-process emails options

mitar:
Alternative to #8038.

Maybe it is simpler to provide a way to code wrap a function and provide any post-processing of options it wants. It is not the best API design, but it is working well in practice to augment things with very custom things.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -519,14 +519,7 @@ Meteor.methods({forgotPassword: function (options) {
 // send the user an email with a link that when opened allows the user
 // to set a new password, without the old password.
 
-/**
- * @summary Send an email with a link the user can use to reset their password.
- * @locus Server
- * @param {String} userId The id of the user to send email to.
- * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
- * @importFromPackage accounts-base
- */
-Accounts.sendResetPasswordEmail = function (userId, email) {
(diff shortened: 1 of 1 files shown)


abernix:
So the instruction to developers would be:  If you want to do more advanced stuff don't call `Accounts.sendResetPasswordEmail`, but call `Accounts._sendResetPasswordEmail` instead?

#8038 seems like a much preferable implementation.

mitar:
I was more thinking of: monkey-patch `Accounts._sendResetPasswordEmail` to do anything you want before or after, before it is send to `Email.send`.

Yes, but `#8038` does not really allow all things wrapping allows. It allows only modifying headers. This one for example allows one to generate an extra (tracking) token and add it to both headers and e-mail content. The issue is that without this, it is hard to correlate call to header function and text/email template to be able to inject the same token into both.

abernix:
Can we either go with #8038 or send this back to the drawing table (in an issue) to design a better plan to address the concerns?

I mean, I guess we can wait and see what MDG thinks, but I don't think making it easier to monkey-patch (and suggesting that) would be an official solution.  If anything, it seems like it'd be more reasonable to recommend developers to fork `accounts-password` and just change it as they wish.

Opening internals or recommending playing with internals just hasn't been highly-desirable methodology from past (and recent) PRs.

mitar:
OK, #8038 should probably go in anyway.

What about `Accounts.emailTemplates.preprocess` callback function which you can set and it will be called with `user, options, type` and expect `options` to be returned before send is called?

mitar:
I made #8065 instead.

Issue #8021, opened 2016-11-05
Scope of imported variables (nested imports)


klaussner:
I was trying to use nested imports and noticed that the scoping doesn't work as described in [reify's README.md](https://github.com/benjamn/reify#how-it-works). In particular, a variable imported inside an `if` block has the scope of a `var` declaration, not a `let` declaration.

For example:

```js
if (true) {
  import value from './file.js';
}

console.log(value);
```

logs the `value` of the imported file (instead of `undefined`) because the code is transpiled to:

```js
if (true) {
  var value;

  module.import('./file.js', {
    "default": function(v) {
      value = v
    }
  });
}

console.log(value);
```

This leads to problems when using the same name for imports more than once, in different block scopes:

```js
if (true) {
  import value from '/imports/a.js';
  console.log(value);
} else {
  import value from '/imports/b.js';
  console.log(value);
}
```

```
=> Errors prevented startup:

   While processing files with ecmascript (for target web.browser):
   client/client.js:5:9: /client/client.js: Duplicate declaration "value"
```

benjamn:
The real reason for this is that Reify sits at the very end of the compiler toolchain, just before Node runs its version of `eval` if you're using the `require("reify")` strategy, or after all other Babel transforms if you're using it as part of [`meteor-babel`](https://github.com/meteor/babel/blob/8ac177e9c682197aebaf7685c0f3fe7e84f95b8e/index.js#L61)).

Since Reify doesn't have the ability to transpile `let` declarations for older versions of Node, it seems like a bad idea for it to be generating new `let` declarations, if there's any chance that the JS engine does not support `let` natively. Unfortunately, Node 4 still doesn't support `let` outside of strict mode, so it's not enough to know that you're using Node 4. You might think that using `import` implies you're in a module, which implies strict mode, but in Meteor we let people opt into `"use strict"` for their modules, in case they're still using auto-imported "global" variables.

With all of that said, I think we may be able to fix this in https://github.com/meteor/babel by moving the Reify compilation step earlier in the Babel plugin pipeline, even if we can't fix it once and for all in Reify itself. Some Babel transforms insert additional `import` declarations, so we would have to reenable the Babel modules transform, but those `import` declarations won't need (or benefit from) Reify because they'll all be at the top level, and Babel will turn them into `require` calls. If this works, then Babel can transpile the `let` declarations created by Reify, and Reify will only have to handle the `import` declarations from your source code (not the ones Babel itself inserts).

[...]

Issue #8008, opened 2016-11-02
less/scss preprocessors not producing css in 1.4.2


eymiha:
In 1.4.2, I added the less package and created a file `app.less` under client containing

    body { margin: 0 }

ran meteor, and no css was emitted. I removed the package and added fourseven:scss, changed the file to `app.scss`, ran meteor and again, no css was emitted.

I created a new project in 1.3.4, added the less package and copied the old `app.scss` to `app.less` under my new projects client. I ran meteor and voila, css was created from the less file.

benjamn:
How are you determining if CSS was emitted? For me, these reproduction steps result in a `<link>` tag as the first child element of the `<head>`, something like:
```html
<head>
  <link rel="stylesheet" type="text/css" class="__meteor-css__" href="/merged-stylesheets.css?hash=2809b331c0316963f4f876a8e1e7f1fc667ba0ee">
  ...
```
If you visit http://localhost:3000/merged-stylesheets.css?hash=... in your browser, the processed CSS should be visible there.

Perhaps try a CSS rule whose effects are a little more obvious, and/or less vulnerable to being overridden elsewhere?
```css
body {
  background: lightgreen;
}
```

klaussner:
I've noticed a similar problem, but only when importing files. Here's a reproduction: https://github.com/klaussner/meteor-css-issue

Since 1.4.2, adding new `.less` files and importing them while `meteor` is running (sometimes) results in `Uncaught Error: Cannot find module './<filename>.less.css'` in the browser console. The files are built correctly when `meteor` is restarted, though.

Akryum:
[Here is the issue](https://github.com/Akryum/meteor-vue-component/issues/80) in the vue-meteor-component project that relates to this (Meteor 1.4.2). Adding new `.vue` files while meter is running throws a not found error.

benjamn:
Please try this again after running
```
meteor update --release 1.4.2.2-beta.0
```
in an existing app, or
```
meteor create --release 1.4.2.2-beta.0 your-new-app
```
to create a new app using the latest beta version.

klaussner:
`.less` and `.scss` imports are working!

averissimo:
Not working for me in `Meteor 1.4.2.3`

What I've done:

- `meteor add less`
- create client/ui/css/app.less with 'body { background-color: #777; }`
- refreshed browser, restarted meteor and it does not add to the `merged-stylesheets.css` file

It detects a change, but does not update files as this appears in meteor console:

[...]

Issue #7978, opened 2016-10-27
[1.4.2] Changes in npm linked modules no longer cause a rebuild


damonmaria:
My guess is in improving build speed this functionality may have been dropped intentionally, but maybe it's an oversight.

In 1.4.1 if I updated any of the several npm packages I am working on that are `npm link`ed to the `node_modules` of the Meteor project then Meteor would rebuild. This no longer happens in 1.4.2. 

Is this the way it's supposed to be? 

MacOS 10.12.1

benjamn:
Yes, starting watchers for files in `node_modules` was prohibitively expensive, so now (in Meteor 1.4.2) we watch for changes only in the contents of `node_modules` directories. I'd be happy to consider further refinements, but it can't involve watching everything in `node_modules`.

damonmaria:
Totally understand. And I'm willing to lose linked modules causing a rebuild for faster build speed. 

But I wonder if a child of `node_modules` is a symlink (should be a good determinant of `npm lnk`ed) then it could be 'deep watched'? We've moved a lot of our common Meteor code out to npm modules and a lot of our editing goes on in those rather than the project we're working on. 

I just tried to see if I could workaround this manually and trigger a rebuild by `touch`ing the symlink but that didn't work. But then I tried it on an ordinary directory in `node_modules` (`react`) and that didn't trigger a rebuild either. So maybe fsnotify or whatever it is doesn't see that as a change. 

Another possibility is to expose in Meteor shell a way to invalidate a path so we can trigger a rebuild if we want. 

Also, just to complicate things, all our packages are npm scoped packages and so the actual package dir is not directly under `node_modules`. Just mentioning that in case it affects anything.

jchristman:
I second the idea of 'deep watching' symlinks in node_modules. The only way that anything should be linked is if someone is intending on editing that module without having to reinstall or update every time they make a change. I think that is likely the simplest solution.

hexsprite:
the only problem with deep watching symlinks is that they often have their own node_modules inside them... so maybe that could be excluded.

benjamn:
Yep, I think this would only apply to symlinks in the top-level `node_modules` directory in the app, right?

jchristman:
Tldr; watch all symlinks and their subsequent symlinks in their node_modules, but don't watch the normal node_modules of the symlinked things.

[...]

PR #7975, opened 2016-10-26
Release 1.4.2.1

benjamn:
This release will be a bug-fixing follow-up to Meteor 1.4.2.

PR #8044, opened 2016-11-10
Release 1.4.2.2 (and 1.4.2.3)

benjamn:
Minor follow-up release to Meteor 1.4.2.1.

In addition to what's currently on `devel`, here are the issues I'm hoping to address with this release:

Detection of newly added files:
  - [x] https://forums.meteor.com/t/meteor-not-detecting-new-files/31151
  - [x] https://github.com/Akryum/meteor-vue-component/issues/80
  - [x] https://github.com/meteor/meteor/pull/7975#issuecomment-259263650
  - [x] https://github.com/meteor/meteor/issues/8008
  - [x] https://github.com/meteor/meteor/issues/7978#issuecomment-259600435

Reify-related issues (possibly to be postponed to 1.4.3):
  - [ ] https://github.com/meteor/meteor/issues/8021
  - [ ] https://github.com/meteor/meteor/issues/7662
  - [ ] https://github.com/meteor/meteor/pull/8030


pozylon:
Very happy to see you guys adding new features to the accounts system and minimongo.

laosb:
@benjamn Is it OK to add a compatibility breaking change (`accounts-facebook` splitting)in a patch version?

benjamn:
@laosb You might be right… how much do those changes break backwards compatibility? We could always release those packages separately, later.

laosb:
Basically if users update, they need to add one more package to get config UI.

benjamn:
@laosb After thinking about it over the weekend, I've decided to remove the `accounts-facebook` reorganization changes from this release, so that the 1.4.2.2 will be smoother. We should still release new versions of those packages immediately after releasing 1.4.2.2.

benjamn:
Please run `meteor update --release 1.4.2.2-rc.1` and test any functionality that's important to you. I'd like to do the final release today or tomorrow. Also please let me know about any changes that I missed that can't wait for the next release.

abernix:
I likely won't have a chance to test this tonight (your today).

I'm not aware of any other issues which cannot wait for the next release.  The only other 'trending' issue that I've seen in the last couple weeks is that Uglify's lack of harmony features (on the published branch) seems to be rearing its head more and more (#8020, #7994, #5517).  I also haven't had an opportunity to verify that those would be fixed with the [`harmony`](https://github.com/mishoo/UglifyJS2/tree/harmony) branch though so maybe that would be a better fit for 1.4.3 (reify, babel, etc.).

benjamn:
Yeah, last I checked the `harmony` branch of UglifyJS was not ready for production, and was not a smooth upgrade. Crazy that they've survived this long without catching up to the times.

abernix:
@benjamn I'd suggest merging #8045 as well.  It implements code-style improvements to the new "skel"  introduced with #7807 (which is surfacing for the first time in 1.4.2.2 thanks to 0466fdd37f757efea0e70a970b4ad5f2071d9d00 above).  It'd be nice to just use the "improved" version.

benjamn:
After taking a closer look at https://github.com/meteor/meteor/pull/8045, even though Meteor 1.4.2.2 is already final (but not yet "recommended"), I think it's worth immediately publishing a 1.4.2.3 release with those changes. Lots of folks are going to encounter those skeleton files before they know anything else about Meteor, so it seems really important to use good style and code structure.

abernix:
Good call on the move for 1.4.2.3.  Also, if it's any assurance, I saw no issues with Meteor 1.4.2.2-rc.1 in my tests on my own apps.

QHose:
Hi Ben,

I used your suggested beta for the file not change issue. It works better. But I noticed meteor does not recognize small changes consistently. E.g. When I make a A HREF with too much quotes. I save, then meteor shows the error. I remove the extra quote but it does not update.

```<a href="{{URL_Youtube_concept_behind}}""  target="_blank">explained in this video</a>```

to

```<a href="{{URL_Youtube_concept_behind}}"  target="_blank">explained in this video</a>```

I saw the same thing yesterday when I had too much blaze }}}. 

{{template}}}

to

{{template}}

QHose:
mmm, could it be I press save too soon... and meteor is not yet checking my files?

PR #8045, opened 2016-11-11
code style changes for skel-full

dburles:
Updated the code style (and dependencies) for the full scaffold to match (as best as seems reasonable) the Meteor guide.

There is some discrepancy with the recommended eslint configuration. For example there are a number of rules that make sense within the context of a Meteor project, which is exactly why [these have been added](https://github.com/meteor/todos/blob/master/package.json#L41-L61) in the todos example. So to address that I've added a few `eslint-disable` comments where it makes sense, though it's not ideal.

In [some cases like this](https://github.com/meteor/meteor/blob/devel/tools/static-assets/skel-full/imports/ui/components/hello/hello.js#L3) it is pretty much standard practise (in Meteor) to not bother with naming function expressions (`func-names`). So we should probably change that, but as you can see it's not great that it then conflicts with the recommended (stock) linter config.

Perhaps a solution is that it can be handled by the [eslint meteor plugin](https://github.com/dferber90/eslint-plugin-meteor), though I'm not totally sure how that might work since it's designed to work independently to the airbnb rules. @dferber90 any thoughts? :)

So in the meantime until there's a solution like above should we remove the `eslint-disable` comments for the sake of simplicity?

### DIFF
--- a/tools/static-assets/skel-full/imports/api/links/links.js
+++ b/tools/static-assets/skel-full/imports/api/links/links.js
@@ -2,4 +2,6 @@
 
 import { Mongo } from 'meteor/mongo';
 
-export const Links = new Mongo.Collection('links');
+const Links = new Mongo.Collection('links');
+
+export { Links };
--- a/tools/static-assets/skel-full/imports/api/links/links.js
+++ b/tools/static-assets/skel-full/imports/api/links/links.js
@@ -2,6 +2,4 @@
 
 import { Mongo } from 'meteor/mongo';
 
-const Links = new Mongo.Collection('links');
-
-export { Links };
+export const Links = new Mongo.Collection('links');
--- a/tools/static-assets/skel-full/imports/api/links/methods.js
+++ b/tools/static-assets/skel-full/imports/api/links/methods.js
@@ -9,12 +9,6 @@ Meteor.methods({
     check(url, String);
     check(title, String);
 
-    // Check if this is a valid url
-    const re = /((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g;
-    if (!url.match(re)) {
-      throw new Meteor.Error('Invalid URL.');
-    }
-
     return Links.insert({
       url,
       title,
--- a/tools/static-assets/skel-full/imports/api/links/methods.js
+++ b/tools/static-assets/skel-full/imports/api/links/methods.js
@@ -1,8 +1,8 @@
 // Methods related to links
 
 import { Meteor } from 'meteor/meteor';
-import { Links } from './links.js';
 import { check } from 'meteor/check';
+import { Links } from './links.js';
 
 Meteor.methods({
   'links.insert'(title, url) {
@@ -10,10 +10,15 @@ Meteor.methods({
     check(title, String);
 
     // Check if this is a valid url
-    if (!url.match(/((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g)){
-      throw new Meteor.Error('Bad url. I.e https://www.meteor.com');
+    const re = /((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g;
+    if (!url.match(re)) {
+      throw new Meteor.Error('Invalid URL.');
     }
 
-    return Links.insert({url, title, createdAt: new Date()});
-  }
+    return Links.insert({
+      url,
+      title,
+      createdAt: new Date(),
+    });
+  },
 });
--- a/tools/static-assets/skel-full/imports/ui/components/info/info.js
+++ b/tools/static-assets/skel-full/imports/ui/components/info/info.js
@@ -2,8 +2,6 @@ import { Links } from '/imports/api/links/links.js';
 import { Meteor } from 'meteor/meteor';
 import './info.html';
 
-/* eslint-disable func-names, prefer-arrow-callback */
-
 Template.info.onCreated(function () {
   Meteor.subscribe('links.all');
 });
--- a/tools/static-assets/skel-full/imports/ui/components/info/info.js
+++ b/tools/static-assets/skel-full/imports/ui/components/info/info.js
@@ -1,8 +1,10 @@
+import { Links } from '/imports/api/links/links.js';
+import { Meteor } from 'meteor/meteor';
 import './info.html';
-import { Links } from '/imports/api/links/links.js'
-import { Meteor } from 'meteor/meteor'
 
-Template.info.onCreated(function infoOnCreated() {
+/* eslint-disable func-names, prefer-arrow-callback */
+
+Template.info.onCreated(function () {
   Meteor.subscribe('links.all');
 });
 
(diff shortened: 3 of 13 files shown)


tools/static-assets/skel-full/package.json
```diff
@@ -5,6 +5,7 @@
     "start": "meteor run"
   },
   "dependencies": {
-    "meteor-node-stubs": "~0.2.0"
+    "babel-runtime": "^6.18.0",
```
abernix: I can see your concern about the `dependencies` here, but keeping the actual latest version number here isn't critical.

The most recent compatible version of the packages will still be installed anyhow (thanks to semver) when the user runs `meteor npm install`.

And for those who don't run `meteor npm install`, the version listed in their own (newly created) `package.json` will get updated soon after they've created their app since users will be given the warning to `meteor npm install babel-runtime --save`.  Running that will bump this version, and install the latest simultaneously.

They can probably just be updated when someone notices.

tools/static-assets/skel-full/imports/api/links/links.js
```diff
@@ -2,4 +2,6 @@
 
 import { Mongo } from 'meteor/mongo';
 
-export const Links = new Mongo.Collection('links');
+const Links = new Mongo.Collection('links');
+
+export { Links };
```
abernix: Is there a particular reason this is the preferred syntax over the prior?

tools/static-assets/skel-full/imports/api/links/methods.js
```diff
@@ -1,19 +1,24 @@
...
 Meteor.methods({
   'links.insert'(title, url) {
     check(url, String);
     check(title, String);
 
     // Check if this is a valid url
-    if (!url.match(/((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g)){
-      throw new Meteor.Error('Bad url. I.e https://www.meteor.com');
+    const re = /((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g;
+    if (!url.match(re)) {
```
abernix: [`RegExp.prototype.test()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) has long been proven to be more performant than [`String.prototype.match()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/match) in the browser.  These days, I find them to be the same in the Node environment, but for the sake of consistency I'd recommend changing this to:

`if (!re.test(url)) {`

tools/static-assets/skel-full/imports/api/links/methods.js
```diff
@@ -1,19 +1,24 @@
...
 
 Meteor.methods({
   'links.insert'(title, url) {
     check(url, String);
     check(title, String);
 
     // Check if this is a valid url
-    if (!url.match(/((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g)){
-      throw new Meteor.Error('Bad url. I.e https://www.meteor.com');
+    const re = /((http|https)\:\/\/)+[a-zA-Z0-9\.\/\?\:@\-_=#]+\.([a-zA-Z0-9\&\.\/\?\:@\-_=#])*/g;
```
abernix: In general, I dislike having such a subjective RegEx in this sample and I'd almost propose changing the entire method to something else (e.g. `colors.insert` *shrug*) based purely for that reason.

As to the RegEx itself, it doesn't need the `/g` modifier and lacks any anchors (e.g. `^`, `$`).  Honestly, I hope this isn't used to actually verify a URL anywhere.  For example, it would match:

    hey sup regex http://http://http://?. <script></script>

...as a valid URL, amongst many other things.

I propose we get rid of this in any way possible.

tools/static-assets/skel-full/package.json
dburles: Yeah I thought the same, though if it was fairly trivial to implement an automated method I figure it would be worthwhile

tools/static-assets/skel-full/imports/api/links/links.js
dburles: My thinking is:
1. It indicates that it will be likely that you'll export more from this file (methods, etc)
2. I think it's a good convention to place all exports at the base of each file, otherwise it can be difficult to determine what's exported

tools/static-assets/skel-full/imports/api/links/methods.js
stubailo: Honestly validating URLs doesn't seem like a good idea in general. Perhaps you could just check if it starts with `http` or something.

tools/static-assets/skel-full/imports/api/links/links.js
dburles: What's the disadvantage though?

tools/static-assets/skel-full/imports/api/links/methods.js
dburles: I agree, I had considered removing it as part of these updates

tools/static-assets/skel-full/imports/api/links/links.js
stubailo: Writing the exported variable twice? It just seems unnecessary. If there are no advantages then the most concise format wins IMO

dburles: Ok I've gone with the more concise version

tools/static-assets/skel-full/imports/ui/components/info/info.js
```diff
@@ -1,8 +1,10 @@
+import { Links } from '/imports/api/links/links.js';
+import { Meteor } from 'meteor/meteor';
 import './info.html';
-import { Links } from '/imports/api/links/links.js'
-import { Meteor } from 'meteor/meteor'
 
-Template.info.onCreated(function infoOnCreated() {
+/* eslint-disable func-names, prefer-arrow-callback */
```
abernix: I find it slightly weird to have this (or any of these) hovering mid-file.  Perhaps hoist all of these to the top?

But on that note, maybe the `eslint` stuff should just be removed until it's figured out what the best way to do it is?  With having these `eslint-disable` lines here (And thus implying an `eslint` "setup" is in place), it almost seems like setting up the new `package.json` with an `eslintConfig` and incorporate `eslint-plugin-meteor`, etc. might be better.

Anyhow, everything else looks great.

dferber90:
@dburles The way the ESLint ecosystem works is that plugins like `eslint-plugin-meteor` provide additional rules to be used (they can also export default configurations, but it's more common to have separate config packages). Configs like `eslint-config-airbnb` specify which plugins to use (if any), how they should be configured, and they can also configure the core rules. What you wanna do is create `eslint-config-meteor` and specify the rules there. It can probably use `eslint-config-airbnb` behind the scenes and it can also provide some defaults for `eslint-plugin-meteor`. See [Shareable Configs](http://eslint.org/docs/developer-guide/shareable-configs) for how to create one. If you want to chat about this face to face just let me know :) Happy to help

tools/static-assets/skel-full/imports/ui/components/info/info.js
dburles: Yep will remove, the eslint config already setup would be ideal for sure, let's go with that.

dburles:
Nice! thanks @dferber90. I had wondered if it were possible to create a package like that but hadn't looked into it, sounds great. I'll get that setup and let you know once I've got something we can start with :)

iDoMeteor:
One thing to note, is that the linting MDG uses _for_ Meteor is utilized in a considerably different context than that of an app built _with_ Meteor. :)

dburles:
Yeah, just to be clear it's not the Meteor core lint config, it's a lint config built for projects

abernix:
@dburles Do you think you can nuke those `eslint` changes (for now) and then revisit the more proper config in a new PR?  It may be too late to float this, but it would be _nice_ to have these changes in the 1.4.2.2 release (where it will make its first official appearance).

dburles:
@abernix Yeah definitely, in that case I would really like to get the other ones done too, what's the ETA on the release?

dburles:
@abernix actually nevermind, they're all good!

benjamn:
Merged into https://github.com/meteor/meteor/pull/8044, thanks!

dburles:
@abernix @dferber90 @iDoMeteor here's a start, it's not yet published to npm. https://github.com/dburles/eslint-config-meteor ideas/feedback are welcome.

dferber90:
@dburles Did a quick overview and this looks good 👍 

You may want to introduce upper limits to your peer dependencies since new rules introduced into them through major releases can mean that the build of people using `eslint-config-meteor` suddenly fails linting without them doing a major upgrade of anything themselves.

PR #8048, opened 2016-11-14
Keep Hashes in URL when rewriting relative URLs

choyweston:
For certain use cases, like SVG filters, the url needs to have a # symbol. For example, the following css property : 

```filter:url("filters.svg#lightGreen")```

In the current version of the codebase, gets rewritten to : 

```filter:url("http://myapp.com/filters.svg")```

Whereas the proper behaviour should be: 

```filter:url("http://myapp.com/filters.svg#lightGreen")```

This simple change will fix the issue.

### DIFF
--- a/packages/minifier-css/minifier.js
+++ b/packages/minifier-css/minifier.js
@@ -145,7 +145,7 @@ var rewriteRules = function (rules, mergedCssPath) {
         // Rewrite relative paths (that refers to the internal application tree)
         // to absolute paths (addressable from the public build).
         if (isRelative(resource.path)) {
-          absolutePath = pathJoin(basePath, resource.path);
+          absolutePath = pathJoin(basePath, resource.path + resource.hash);
         } else {
           absolutePath = resource.path;
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by choyweston** [comment]
Not sure why this would fail the CircleCI test.

File: packages/minifier-css/minifier.js [hunk 1]
```diff
@@ -149,6 +149,10 @@ var rewriteRules = function (rules, mergedCssPath) {
         } else {
           absolutePath = resource.path;
         }
+        
+        if(resource.hash){
```
abernix: Style: Spaces before and after the parenthesis, please! :)

choyweston:
@abernix I added the tests and fixed the styling. Never added tests before, let me know if there are any mistakes there.

choyweston:
Woohoo! First commit to Meteor :)

abernix:
It appears that this may not have been properly released since `standard-minifier-css` was still pinned to an older version of `minifier-js`.  This should be fixed and officially released now with `standard-minifier-css@1.3.4`.

Thanks, @choyweston!  (If you have a chance to test and confirm, please do!)

PR #8051, opened 2016-11-14
update js minifier

alextaaa:


abernix:
Thank you for submitting this, but please resubmit your pull request targeting the `devel` branch instead of `release-1.4.2.2` per the [submitting pull-requests](https://github.com/meteor/meteor/blob/devel/Contributing.md#submitting-pull-requests) requirements.

Additionally, while it's always reasonable to keep these packages up to date, perhaps you could mention any notable changes that you found beneficial in the new version when opening the new PR (i.e. the reason you wanted to update it).  This makes it easier to determine if any existing Meteor issues might be fixed by this upgrade. :smile:

PR #8065, opened 2016-11-17
Allow a function to further change options object before passing it to Email.send

mitar:
A better version of #8039.

### DIFF
--- a/packages/accounts-password/email_tests.js
+++ b/packages/accounts-password/email_tests.js
@@ -30,8 +30,13 @@ testAsyncMulti("accounts emails - reset password flow", [
       "getInterceptedEmails", this.email, expect((error, result) => {
         test.equal(error, undefined);
         test.notEqual(result, undefined);
-        test.equal(result.length, 2); // the first is the email verification
-        var options = result[1];
+        test.equal(result.length, 2);
+        // the first is the email verification
+        var options = result[0];
+        test.equal(options.headers['My-Custom-Header2'], 'Cool2');
+
+        // the second is the reset password
+        options = result[1];
 
         var re = new RegExp(Meteor.absoluteUrl() + "#/reset-password/(\\S*)");
         var match = options.text.match(re);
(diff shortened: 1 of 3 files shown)


tmeasday:
So my feeling here is that wrapping our email generation functions in something to tweak the output feels like a brittle and less flexible approach than the alternative, which is: simply splitting those functions into two parts:

 A) the token/url generation and writing part
 B) the email generation part

If you want to do custom stuff that isn't easily accomplished, you can just call A) and then build your own email. So for instance, put [this part](https://github.com/meteor/meteor/blob/devel/packages/accounts-password/password_server.js#L531-L555) into a function called `Accounts.generatePasswordResetUrl` or the like.

@mitar what do you think?

mitar:
I would instead maybe split it so that one function returns `options` for mails, and then the other just calls the first one and sends (the existing function)?

abernix:
I think it should be more like "in addition" instead of "instead" and really break it down and provide maximum customizability (some other e-mail template tool/email-sending API, if they wish)

There are already `Accounts.urls.{resetPassword, verifyEmail, enrollAccount}` methods for generating the actual URLs, so maybe the portion @tmeasday linked to should become three separate functions for each of the three e-mail types: `generatePasswordReset`, `generateEnrollment` and `generateVerification` (since each generates a token _slightly_ differently).  Alternatively, a `generateFor` method which accepts a `type`.  Either way, it just returns the `user` (for personalization) and the token.

The portion which creates the options (and seems to be duplicated exactly between the three e-mails, could become a private function which generates the `options`.

```js
function generateEmailOptions(template, { user, email, url }) {
  return {
    to: email,
    from: template.from,
    text: template.text === "function" && template.text(user, url),
    ...
  };
}
```

The bulk of those three `Accounts.send*` e-mail sending methods is then (roughly):

```js
Accounts.sendBlah = function (userId, address) {
  const { user, token } = Accounts.generateBlah({userId, address});
  const url = Accounts.urls.blahEmail(token);
  const options = generateEmailOptions(Accounts.emailTemplates.verifyEmail, { user, email, url });
  Email.send(options);
}
```

Then, if someone wants to send their own complex e-mail or use "Email Service" [Template API](https://mandrillapp.com/api/docs/templates.JSON.html) (for example) to send pre-designed templates (or ones their designers created already), they can just implement their own `sendBlah` method using those simple functions and not use `Accounts.sendBlah` at all.

mitar:
Do we have an agreement on this then?

mitar:
@tmeasday: was this an upvote for yes? So I should restructure the code as suggested above?

benjamn:
I like the idea of making it easier to implement your own `send*` methods, instead of making it possible to influence the behavior of `Accounts.send*` methods. If you have time to look into that, @mitar, please let us know how it goes.

abernix:
My above example could be further simplified by having `Accounts.generateBlah` also return the `url` in addition to the `token` (by calling `Accounts.urls.blahEmail` itself).  Even if someone doesn't need it, the overhead is minimal and would eliminate the extra call in user's custom functions.

mitar:
Closing in favor of #8207 where I implemented this as discussed above.

Issue #6707, opened 2016-04-05
[oAuth] doesn't work with subpath: end_of_popup_response.js get 404


Buom01:
Hi,
When I'm logging in with an external service, all works except a little detail that block all, I'm using a subpath;

> Failed to load resource: the server responded with a status of 404 (Not Found)

Because it's try to get `end_of_popup_response.js` without subpath

So with this bug, i'm not logged in, and the popup is keeped open and white

Packages: 
- Using facebook, google, and twitter login package

Context:
- Tried with my deployed app :/ with Firefox and Chromium on Linux with a logged user (that should works)
- Meteor 1.3 on linux, using demeteorize

Reproduce:
- https://github.com/Buom01/meteor-oauth-bug
  You will need to create github keys

Buom01:
While waiting a patch:
```bash
$ meteor add meteorhacks:picker
```
```javascript
Picker.route('/packages/oauth/end_of_popup_response.js', function(params, req, res, next) {
  res.writeHead(302, {
    'Location': "/y/packages/oauth/end_of_popup_response.js"
  });
  res.end();
});
```

abernix:
@Buom01 I know this is a relatively old issue, but I think another related issue just came in for it.  See my comments here: https://github.com/meteor/meteor/issues/7979#issuecomment-256851071  This sub-path configuration is likely not very common and therefore this hasn't got much attention, but your help in identifying it and seeing if we can work out a fix for it would be appreciated.

Maybe pop into that other issue and see if it sounds like the same problem to you?  I made a recommendation on how it might possibly be fixed but I'm curious what you think.

keyscores:
@abernix I've noticed that OAuth has no tests.  Am I mistaken or is this the case?
See here:
https://github.com/meteor/meteor/blob/master/packages/accounts-oauth/oauth_tests.js

abernix:
@keyscores The tests which exist are in the parent packages: [`oauth`](https://github.com/meteor/meteor/blob/devel/packages/oauth/oauth_tests.js#L16), [`oauth2`](https://github.com/meteor/meteor/blob/devel/packages/oauth2/oauth2_tests.js#L1) and [`oauth1`](https://github.com/meteor/meteor/blob/devel/packages/oauth1/oauth1_tests.js#L1).

keyscores:
@abernix thanks for the quick response. Can you quickly point me to the code which handles the route callback request: http://domain.com/_oauth/github?close 

[...]

PR #8081, opened 2016-11-21
Fix oauth ROOT_URL_PATH_PREFIX

koszta:
Fixes issue #6707

### DIFF
--- a/packages/oauth/oauth_server.js
+++ b/packages/oauth/oauth_server.js
@@ -342,7 +342,10 @@ var renderEndOfLoginResponse = function (options) {
     throw new Error('invalid loginStyle: ' + options.loginStyle);
   }
 
-  var result = template.replace(/##CONFIG##/, JSON.stringify(config));
+  var result = template.replace(/##CONFIG##/, JSON.stringify(config))
+    .replace(
+      /##ROOT_URL_PATH_PREFIX##/, __meteor_runtime_config__.ROOT_URL_PATH_PREFIX
+    );
 
   return "<!DOCTYPE html>\n" + result;
 };
--- a/packages/oauth/oauth_tests.js
+++ b/packages/oauth/oauth_tests.js
@@ -59,3 +59,51 @@ Tinytest.add(
     test.equal(OAuth._retrievePendingCredential(key, secret), cred);
   }
 );
+
+Tinytest.add("oauth - _endOfLoginResponse with popup loginStyle supports ROOT_URL_PATH_PREFIX",
+  function (test) {
+    var rootUrlPathPrefix = __meteor_runtime_config__.ROOT_URL_PATH_PREFIX;
+    __meteor_runtime_config__.ROOT_URL_PATH_PREFIX = '/test-root-url-prefix';
+    var res = {
+      writeHead: function () {},
+      end: function (content) {
+        __meteor_runtime_config__.ROOT_URL_PATH_PREFIX = rootUrlPathPrefix;
+        test.matches(
+          content,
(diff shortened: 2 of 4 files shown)


packages/oauth/oauth_tests.js
```diff
@@ -59,3 +59,51 @@ Tinytest.add(
     test.equal(OAuth._retrievePendingCredential(key, secret), cred);
   }
 );
+
```
abernix: It would be nice to also add a test for the default (current) `ROOT_URL_PATH_PREFIX` behavior (i.e. blank `''` and not modified).

packages/oauth/oauth_tests.js
```diff
@@ -59,3 +59,51 @@ Tinytest.add(
...
+          content,
+          /\/test-root-url-prefix\/packages\/oauth\/end_of_redirect_response\.js/
+        );
+      }
+    };
+    var details = {
+      credentials: {},
+      loginStyle: 'redirect',
+      query: {
+        // {"redirectUrl": "http://localhost:3000/"}
```
koszta: Not specifying it breaks the test:
`TypeError: must start with number, buffer, array or string
    at fromObject (buffer.js:208:11)
    at new Buffer (buffer.js:65:10)
    at Object.OAuth._stateFromQuery (packages/oauth.js:99:14)
    at Object.OAuth._endOfLoginResponse (packages/oauth.js:419:25)`

Specifying it (but not http://localhost:3000/) prints a warning on the console:
`Error in OAuth Server: redirectUrl () is not on the same host as the app (http://localhost:3000/)
`

The comment I added is there simply to make clear what's in the base64 encoded string.

Would this be a better option?
```js
state: new Buffer(JSON.stringify({
  redirectUrl: __meteor_runtime_config__.ROOT_URL
}), 'binary').toString('base64')
```

abernix: > The comment I added is there simply to make clear what's in the base64 encoded string.

Ah!  I didn't realize that it was the encoded version of `state`.  :smile:

Yes, I'd say change it to use the code you provided otherwise that message (`not on the same host as the app not on the same host as the app `) will print out in the (rare) cases when tests are run on a different port (for example, `./meteor --port 3600 test-packages oauth`).

packages/oauth/oauth_server.js
```diff
@@ -342,7 +342,10 @@ var renderEndOfLoginResponse = function (options) {
     throw new Error('invalid loginStyle: ' + options.loginStyle);
   }
 
-  var result = template.replace(/##CONFIG##/, JSON.stringify(config));
+  var result = template.replace(/##CONFIG##/, JSON.stringify(config))
+    .replace(
+      /##ROOT_URL_PATH_PREFIX##/, __meteor_runtime_config__.ROOT_URL_PATH_PREFIX
```
benjamn: Can these `RegExp`s appear more than once? If so, you should use `/##ROOT_URL_PATH_PREFIX##/g` to match multiple times. Otherwise only the first occurrence will be replaced. If you don't think anything needs to be changed, just comment to let me know you've seen this question.

abernix: Fair question. It's fine either way but in the current implementation, no, it's not possible.

However, looking at it from a "this looks like a template" angle with consideration for future changes: I'd still say "never" for `##CONFIG##` and "it's highly unlikely" for `##ROOT_URL_PATH_PREFIX##`.  The chances of [this template](https://github.com/meteor/meteor/blob/devel/packages/oauth/end_of_redirect_response.html) becoming more complicated are slim-to-none as it's just meant to load the one `<script>` which performs a trivial function.  Any cause for repetition of that tag would have to be meditated and I'd probably refute. :wink:

Issue #6785, opened 2016-04-11
Meteor 1.3's windows command-line output


alphashuro:
Meteor 1.3's commandline output on windows 10 always prints at the very first line of the command-line, doesn't clear the previous text first (so it is jumbled up with whatever I had typed when I first opened cmd) and doesn't print complete output probably because of the fact that it keeps overwriting itself at the first line.

benjamn:
This issue will be officially fixed when Meteor 1.3.2 is released this Friday, but you can test it out by doing `meteor update --release 1.3.2-rc.5` now. Thanks for bringing this issue to our attention!

alphashuro:
Awesome stuff, I will test as soon as I boot up windows. What was the issue if I may ask?

alphashuro:
the issue is still there as you can see here in this screen shot:
![image](https://cloud.githubusercontent.com/assets/8580352/14487837/2707a5ba-0165-11e6-9bca-080cf13ebb45.png)
Also it takes REALLY long to start up the app on windows, I sometimes doubt if its even going to start up at all. It spends a lot of time building the application, trying to startup, then building again, and so on.

benjamn:
Hmm, you're right to ask, and I'm afraid I jumped the gun here. A [change](https://github.com/meteor/meteor/pull/6664) was made to `dev_bundle/bin/npm.cmd` that prevents it from echoing every `meteor npm` command, but the problems you're seeing in Powershell are clearly unrelated to that. Sorry for the false hope.

Do you get the same messy output if you run Meteor from the `cmd` prompt?

alphashuro:
Yes I get the same on the cmd prompt


Alpha

0737907955



From: Ben Newman

Sent: Friday, April 15, 12:20 AM

Subject: Re: [meteor/meteor] Meteor 1.3's windows command-line output (#6785)

To: meteor/meteor

Cc: Alpha Shuro



Hmm, you're right to ask, and I'm afraid I jumped the gun here. A change was made to dev_bundle/bin/npm.cmd that prevents it from echoing every meteor npm command, but the problems you're seeing in Powershell are clearly unrelated to that. Sorry for the false hope.


Do you get the same messy output if you run Meteor from the cmd prompt?

alphashuro:
Lol please ignore the extra stuff in that comment, wanted to see if replying to github emails would be a viable way to respond, it seems not!

Rhywden:
[...]

PR #8082, opened 2016-11-21
Fix for console output in PowerShell on Win10

Rhywden:
See https://github.com/meteor/meteor/issues/6785

Basically, 
```
new Array(249).join('\b')
```
is not a good idea for a carriage return. Code has been tested with CMD, PowerShell and PowerShell in Legacy Mode and seems to solve the issue.

### DIFF
--- a/tools/console/console.js
+++ b/tools/console/console.js
@@ -68,7 +68,7 @@ var wordwrap = require('wordwrap');
 var PROGRESS_DEBUG = !!process.env.METEOR_PROGRESS_DEBUG;
 var FORCE_PRETTY=undefined;
 var CARRIAGE_RETURN =
-  (process.platform === 'win32' && process.stdout.isTTY ? new Array(249).join('\b') : '\r');
+  (process.platform === 'win32' && process.stdout.isTTY && process.title !== 'Windows PowerShell' ? new Array(249).join('\b') : '\r');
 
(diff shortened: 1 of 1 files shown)


abernix:
@Rhywden Your solution may very well work just fine so I'm just leaving https://github.com/abernix/meteor/commit/ba49ff53dc51cc78389e6c055e373c701e01125c here as an alternate approach in case this proves to cause problems for Internationalized versions of Windows where `process.title` may not be `Windows PowerShell` (but instead is something in the local language).  I'm not sure if that is ever the case in practice though?

In my commit, I detect the feature using the Windows SDK and change the linefeed in the same way as you did.

Nice job looking at this though!  This is great.

manzonif:
Hi, I still have similar issue. Output start always from top of window (command prompt and power shell) and overwrites the pre-existing text. 
This on Win 10 italian locale.

abernix:
@manzonif What version of Meteor is your app using?

manzonif:
meteor --version
Meteor 1.5

ensemblebd:
Same problem, windows 10, meteor 1.6
Maybe I should just scrap the entire idea of using meteor on windows period, this is getting ridiculous.
Can't run IOS anyway.. Might as well spin up a mac vm or dvi kvm at this point.
No one has a solution for this?  Has anyone tried Cygwin?

PR #8116, opened 2016-11-30
Fix history and add history for #7970

abernix:
As a correction to 8e22818 which inadvertently removed 1.4 from existence.

Good catch by @mitar.  https://github.com/meteor/meteor/commit/8e228180e66a9808f268f1de1b9ead2ec21a8751?diff=unified#commitcomment-20013960

/cc @zol – I think this is what you wanted.


zol:
Thanks @abernix . Mistake with referencing `origin` rather than `upstream` . Kinda surprised there were no conflicts.

PR #8120, opened 2016-12-01
Help address the Unicode asset filename issues discussed in #5626

hwillson:
Hi guys - this PR is intended to address issue #5626 by:

1. Bringing back the changes made in PR #7963 (these changes are still needed - we just need to take things a bit further).
2. Adding new changes to address failing tests on Linux.

As mentioned in https://github.com/meteor/meteor/pull/7963#issuecomment-263975539, the unicode asset tests were passing on OSX but failing on Linux. When a  relative path including a Unicode based file was split up in the `_sanitize` function of `tools/isobuild/builder.js`, the resulting split filenames were different on OSX and Linux. This was caused by the same Unicode normalization issues that PR #7963 was addressing for the Assets API.

While the `_sanitize` function could be patched to address this by first making sure the relative path containing the Unicode based file name is normalized using [`String.prototype.normalize()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/normalize), I think it makes sense to move this fix higher up the code tree. Other areas of the codebase could run into this same issue.

This PR makes a very small change in `tools/isobuild/package-source.js`, within the `_findAssets` function, to normalize incoming asset paths/filenames. This seemed to be the most logical first place assets are found for a unibuild, and have their relative path stored for later use. Normalizing asset  path/filenames here fixes the problems discovered in the builder, but will also help make sure this problem doesn't occur anywhere else the asset path/filenames are referenced internally.

Let me know if you have any comments/questions or need anything changed. Thanks!

P.S. --> I'm very sorry about the broken tests in #7963; those tests weren't broken when I was testing locally on OSX, or broken when running in my own Meteor fork configured on Circle CI with Linux (or so I thought - I found the reason why they appeared to be passing via Circle CI, but weren't really; I'm going to open a separate issue about this as I think this could be easily addressed by tweaking the `./scripts/ci.sh` script a bit). Thanks!


abernix:
The latest [test failure](https://circleci.com/gh/meteor/meteor/3136) on `devel` was a different failure.  It appears to be due to a general CircleCI crash (as it happened on two different containers).  I've re-ran it and expect it to pass.:

```sh
METEOR_PRETTY_OUTPUT=0 ./meteor --get-ready died unexpectedly
```

PR #8122, opened 2016-12-01
Make CI test runner more defensive and a bit more flexible

abernix:
General improvements to the CI script:

* Changes the CI script to make sure it runs all tests, regardless of the environment configuration while still allowing flexibility in the size of the container parallelism.

  While not having any effect on the Meteor configuration this will now run all 8 test groups across a fewer number of containers, if necessary.

  This fixes issues where false-positives are achieved on contributor PRs, like seen on meteor/meteor#7963 and for the reason I explained here: https://github.com/meteor/meteor/pull/7963#issuecomment-264188162

* Permits changing `TIMEOUT_SCALE_FACTOR` using a variable instead of hard-coding it in the script and adds `ADDL_SELF_TEST_EXCLUDE` to define environment specific tests which must be skipped.  For example:

  ```sh
  $ TIMEOUT_SCALE_FACTOR=20 ./scripts/ci.sh # take your time!
  ```

  ```sh
  $ ADDL_SELF_TEST_EXCLUDE="^stubborn-(run-)?test$" ./scripts/ci.sh
  ```

  For example, I run my preliminary (personal) Meteor tests on SemaphoreCI, but there is a particular test that just _will not work_.  This allows to skip that without a hack-y `sed` parse.

* (reverted) 6701357c41078e1e8a90384e4fc71ba0ab3301ff was probably overboard and just cleaned up the way the skipped tests were defined.  I reverted it as one of the features I was used was a `bash`-ism and in the spirit of making `meteor` work in as many places as possible I didn't want to force that.

(We'll see if this PR passes on my own CircleCI with just 4 containers)

/cc @hwillson @benjamn

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -1,6 +1,24 @@
-export TIMEOUT_SCALE_FACTOR=15
-export TEST_PACKAGES_EXCLUDE="less"
-export SELF_TEST_EXCLUDE="^can't publish package with colons|^old cli tests|^logs - logged (in|out)|^mongo - logged (in|out)|^minifiers can't register non-js|^minifiers: apps can't use|^compiler plugins - addAssets"
+#!/bin/sh
+
+#
+# Optional Environment Variables for Configuration
+#
+# - TIMEOUT_SCALE_FACTOR: (default: 15)
+#   A multiplation factor that can be used to raise the wait-time on
+#   various longer-running tests.  Useful for slower (or faster!) hardware.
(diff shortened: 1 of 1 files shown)


abernix:
✅ It did work on my CI and [you'll see in other tests here](https://circleci.com/gh/abernix/meteor/tree/feature%2Fimprove-ci-script) that it failed just before this commit which was the un-rebased version without #8120 (which would have passed previously).

hwillson:
This is awesome @abernix - thanks for putting this together!

Issue #7778, opened 2016-09-15
Meteor.setTimeout() does not return a "Number" type handle


CdavM:
Background:
I'm running Meteor 1.4.1.1 on Ubuntu 16.04. This bug is reproducible on a fresh Meteor app with no dependencies.

Actual behavior:
Running
`x = Meteor.setTimeout(function(){console.log("hello");}, 1000000);`
`typeof(x)` returns an `object`. Here is what mine looks like:

```
{ _called: false,
  _idleTimeout: 1000000,
  _idlePrev: 
   Timer {
     '0': [Function: listOnTimeout],
     _idleNext: [Circular],
     _idlePrev: [Circular],
     msecs: 1000000 },
  _idleNext: 
   Timer {
     '0': [Function: listOnTimeout],
     _idleNext: [Circular],
     _idlePrev: [Circular],
     msecs: 1000000 },
  _idleStart: 179871,
  _onTimeout: [Function],
  _repeat: null }
```

Expected behavior:
`typeof(x)` returns `number` which can be passed into a clearTimeout() function. For instance, running the code in a console returns an integer (in my case, 1146).

More info:
https://docs.meteor.com/api/timers.html#Meteor-clearTimeout 
Implies that a Number is returned as a handle from Meteor.setTimeout(). However, this is not the case. An object is returned instead. 

I think the server-side implementation should mimic both javascript conventions and the client-side.

I am unable to locate the function setTimeout() in the source (and the link on the docs page is not constructive). Please fix it or advise on how to bypass it. I'd like to have something that can be stored into a Mongo collection.

mitar:
Handles from `setTimeout` should be seen as an opaque resource you cannot and should not really be storing into Mongo. See them as a "file descriptor"-like thingy. They are valid just for that one process.

CdavM:
But is there a benefit to the current implementation, where you use "file descriptor"-like things on the client but expose the object on the server? Why not have the same behavior on both ends?

Anyway, the implementation should at least be reflected in the docs, so we either make this bug a feature and document it or change the behavior.

mitar:
> Why not have the same behavior on both ends?

Because it is an opaque value. You should not care what it is. :-)

Anyway, this is a duplicate of: https://github.com/nodejs/node-v0.x-archive/issues/5728

mitar:
So this is just what underlying implementation is returning, in browsers and on node.

See also: https://github.com/nodejs/node-v0.x-archive/pull/7082

[...]

PR #8127, opened 2016-12-02
[docs] Clarify docs for handle returned by Meteor.setTimeout/setInterval

abernix:
Thanks to @CdavM for pointing out this JSdoc discrepancy in #7778.

The value passed into from `Meteor.clearTimeout` or `Meteor.clearInterval` is an opaque handle returned by `Meteor.setTimeout` and `Meteor.setInterval` accordingly.

That is to say, on some engines it will be a `Number` and on others an `Object` but you need not worry about the actual type.

Closes meteor/meteor#7778


zol:
Lgtm, thanks @abernix !

PR #8128, opened 2016-12-02
Remove skipped CI tests which don't exist anymore

abernix:
As a follow-up to cleaning up `ci.sh` yesterday in #8122, I found some tests that don't exist anymore but were still "excluded" from running.  This removes the exceptions for these non-existent tests:

* ^mongo - logged (in|out)
* ^logs - logged (in|out)
* ^can't publish package with colons


zol:
Lgtm, thanks!

PR #8133, opened 2016-12-04
Use CoffeeScript 1.12.0

GeoffreyBooth:
[CoffeeScript 1.12.0 is out](http://coffeescript.org/#changelog). This updates Meteor’s package to use the new version.

I removed the `if` check that passed through the CoffeeScript compiler’s output to `ecmascript` only if certain ES2015+ features were recognized in the original source. Now all coffee output is parsed by `ecmascript`. With the addition of tagged template literals and `for…of` in CoffeeScript 1.12.0, I don’t think it’s possible anymore to reliably detect when ES2015+ features are being used. We should just always assume there could be ES2015+ syntax in the CoffeeScript compiler output, unfortunate though that is from a performance perspective.

### DIFF
--- a/packages/coffeescript/plugin/compile-coffeescript.js
+++ b/packages/coffeescript/plugin/compile-coffeescript.js
@@ -88,29 +88,29 @@ export class CoffeeCompiler extends CachingCompiler {
       inputFile.getDeclaredExports().map(e => e.name)
     );
 
-    if (/`|\b(?:import|export|yield)\b/.test(source)) {
-      // If source contains backticks or features that output as ES2015+,
-      // pass the coffee output through babel-compiler
-      const doubleRoastedCoffee =
-        this.babelCompiler.processOneFileForTarget(inputFile, output.js);
+    // CoffeeScript contains a handful of features that output as ES2015+,
+    // such as modules, generator functions, for…of, and tagged template
+    // literals. Because they’re too varied to detect, pass all CoffeeScript
+    // compiler output through the Babel compiler.
+    const doubleRoastedCoffee =
+      this.babelCompiler.processOneFileForTarget(inputFile, output.js);
 
-      if (doubleRoastedCoffee != null &&
-          doubleRoastedCoffee.data != null) {
-        output.js = doubleRoastedCoffee.data;
+    if (doubleRoastedCoffee != null &&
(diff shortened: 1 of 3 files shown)


packages/coffeescript/plugin/compile-coffeescript.js
```diff
@@ -88,29 +88,29 @@ export class CoffeeCompiler extends CachingCompiler {
...
-          // then BabelCompiler will not transpile it, and there will be
-          // no sourceMap, but that's fine because the original
-          // CoffeeScript sourceMap will still be valid.
-        }
+    // CoffeeScript contains a handful of features that output as ES2015+,
+    // such as modules, generator functions, for…of, and tagged template
+    // literals. Because they’re too varied to detect, pass all CoffeeScript
+    // compiler output through the Babel compiler.
+    const doubleRoastedCoffee =
+      this.babelCompiler.processOneFileForTarget(inputFile, output.js);
```
benjamn: I believe you can enable caching easily if you define a `CoffeeCompiler#setDiskCacheDirectory` method:
```js
class CoffeeCompiler extends CachingCompiler {
  ...
  setDiskCacheDirectory(cacheDir) {
    this.babelCompiler.setDiskCacheDirectory(cacheDir);
  }
  ...
}
```

GeoffreyBooth: How would I be able to test that this works?

benjamn: If you just mean convincing yourself that it makes a difference, you can `console.log({cacheDir})` and then inspect that directory to verify that compiled `.json` files are getting saved by the `BabelCompiler`. You could also put some timing code around `this.babelCompiler.processOneFileForTarget(...)` to make sure it gets faster after the first compile.

If you mean programmatically testing that it works, then I'm not sure what the best strategy is. It's tricky because caching is supposed to be a transparent optimization, so it can be hard to detect/verify from the outside.

GeoffreyBooth: I meant the former. Thanks, I’ll look into this.

benjamn: Looks good to me! I'll wait for the tests to pass and then merge this.

Issue #7849, opened 2016-09-30
Meteor Cordova app still tries to connect to local server after a build with --server option.


dagatsoin:
- METEOR@1.4.1.1
- ROOT_URL (on server) is set to: X.X.X.X:8080
- Build command of the server: `meteor build app/ --directory --server-only --server 0.0.0.0:8080`

(Xpost with an [SO question](http://stackoverflow.com/questions/39785872/how-to-set-a-server-url-to-connect-to-for-a-meteor-cordova-app), but I suspect an issue)

I build my meteor cordova app with `meteor build ../output --server myserverprod.com:8080`

Then I upload the android-armv7-debug.apk to an FTP and redownload it on my devices.

But when I remote debugging I see that it still tries to connect to my local server 192.x.x.x instead of the URL I put in the command line.

I also tested by disable wifi to be sure that it happens with the initial embedded files.

Is there any environment variable to set which are not documented?

abernix:
I realize the docs don't make this 100% clear, but make sure that your `ROOT_URL` includes the protocol (e.g. `https://`, `http://`) and the same with the `--server` argument when building. (`ROOT_URL=http://myserverprod.com:8080` and `--server https://mysererprod.com:8080`)

Also...

> Run command of the server: `meteor build app/ --directory --server-only --server 0.0.0.0:8080`

This command will simply generate a bundle directory of a Node app; this is **not** a "Run command".  Additionally, the `--server` is actually not relevant on this command (only used on the mobile build).  In the output bundle directory, you'll find a `README` with instructions on how to run the Node server.  When starting, you'll define the `ROOT_URL` which should match the `--server` you used when generating the mobile bundle.

dagatsoin:
Note that the doc is also different form the command line help.
The command line says we must write --server**=** and the doc says to use a space. So which one is right?

I will double check the protocol.

abernix:
Either **should** work, but its been suggested that there is a bug with that logic: #7112.  (Though, the reproduction steps did not work for me last I tried).  Please do report over there if you experience the same behavior.

dagatsoin:
EDIT: correct my first post bout run/build.

I did not write the port in ROOT_URL. I will see if it fixes.

abernix:
Please also make sure you're starting the `node` process as suggested in the `README`:

```sh
  $ (cd programs/server && npm install)
  $ export MONGO_URL='mongodb://user:password@host:port/databasename'
[...]

Issue #7291, opened 2016-06-24
Inconsistent Android build with Crosswalk


Exilz:
**_1 Upvote**_ @martijnwalraven Since Meteor 1.3 (I'm currently using Meteor 1.3.3.1), I noticed a strange bug when building Android applications.

The problem is, when I build the project like so `meteor build ../BUILDS/TEST-BUILD-ADNROID/1.0.0 --server localhost:3000` (purely for example), I noticed that the output directory isn't the same everytime.

On my first attempt, the directory `android/project/build` doesn't exist at all.
Then, I deleted it, and typed the previous command again.

Aaand hurray ! here is my `android/project/build/outputs/apk` folder with `android-x86-release-unsigned.apk` & `android-armv7-release-unsigned.apk` as expected.

This gets even stranger when, sometimes, the apk folder is there, but it's not filled with `release` apks, only debug ones. I couldn't reproduce this 100% of the time, but this happens.
I noticed it when I tried to sign them, since jarsigner told me "dafuq dude, it's already been signed".

The only way around this is to `meteor reset` before the first build, fail the first one, delete the destination folder, and succeed on the second try.
Since building is taking a _really long_ time on my projects (25 cordova plugins...), having to build twice really is time consuming for us.

I could reproduce this in an empty project :

```
meteor create testmobile
cd testmobile
meteor add-platform android
meteor add crosswalk
npm install
meteor build ../BUILDS/TEST-BUILD-ADNROID/1.0.0 --server localhost:3000
```

-> No `build` directory.

I don't know if this is related to meteor is itself or crosswalk. I didn't notice this on my Cordova projects not related to meteor, though.

chrisknu:
+1 for a fix. Noticed the same issue, just kindof dealt with it by re-running after doing a meteor reset a few times

pmwisdom:
I think this is the same issue I saw before. The workaround is to run the build script yourself. to generate the APK's. 

See https://github.com/meteor/meteor/issues/6756 for more info.

AnthonyAstige:
I've also noticed inconsistencies in build names / directories, if it's there or not, etc.  Once other piece of perhaps interesting and maybe related information: Sometimes the app says it's build, but it appears to have an old version.

Steps to (sometimes) reproduce

1. Build the app as normal
1. Update your version number in `mobile-config.js`
[...]

Issue #6756, opened 2016-04-08
No APK's Automatically Generated via Meteor Build ( 1.3.1 & 1.3.2 )


pmwisdom:
I've noticed that Meteor Build is behaving quite odd since 1.3.1, for instance, if you do a meteor reset (delete .meteor/local) right before you build, the build will succeed but it won't generate any APK's. If you run meteor, and then build, it will generate APK's.

I also noticed something else that was strange. Lets say I'm on my master branch and I run meteor, and then I switch over to my staging branch and immediately build, it will actually build the last run from my master branch, not the staging branch i'm trying to build. It seems to be building whatever was last run, not the current state of the app at that point and time.

pmwisdom:
I'm not 100% of why this is happening still but a work around if anyone else sees this is to run build from inside the android/project/cordova/ folder of your meteor build. That will generate your APK's.

davidwoody:
@pmwisdom We are seeing the same thing on our project... basically our build script does:

```
rm -rf .meteor/local; rm -rf ../output-build;

meteor build ../output-build --server=$SERVER --mobile-settings $SETTINGS;
```

Meteor 1.3 outputs the`android/project/build` directory, but Meteor 1.3.1 or 1.3.2 does _not_ output the `android/project/build` directory (with the APK's we're after).

conrad-vanl:
Same here. Running the `build` script within `android/project/cordova` does generate APKs, but with filenames that don't match what we're used to seeing (things like 'debug' in the filename)

pmwisdom:
@conrad-vanl you need to run build with the --release flag.

oliverthomasklein:
@pmwisdom I'm also getting this issue. Right now there appears to be no build folder. I'm updated to the latest 1.3.4.4. When you say add the --release flag what release are you using that helped you work around this?

pmwisdom:
@oliverthomasklein I was referring to the --release flag of **cordova** build, not to the **meteor** release. Meaning if you want to generate apks for app store distribution you need to provide the --release flag to the build executable in /cordova-build/platforms/android/cordova/build. If you do not, you will generate debug apks.

ffflorian:
Same problem here, running Meteor 1.4.0.1.

jimmy1993k:
Same problem on 1.3.5.1, you can get the APK's when you run the cordova build command yourself as in when you have build to some build folder: 
`<build-folder>/android/project/cordova/build --release`

[...]

Issue #7112, opened 2016-05-24
hot code should work when you pass server url without "="


crapthings:
this will not work

```
meteor build ../ios --server=http://139.162.2.61:3000
```

this will work, because you can't use "=" here, that will not pass ROOT_URL to cordova meteor_runtime_config

```
meteor build ../ios --server http://139.162.2.61:3000
```

check your building folder to find index.html, you can see that if your **meteor_runtime_config** ROOT_URL is not your server ip:port, that is the issue.

```
<script type="text/javascript">
  __meteor_runtime_config__ = JSON.parse(decodeURIComponent("%7B%22meteorRelease%22%3A%22METEOR%401.3.2.4%22%2C%22ROOT_URL%22%3A%22http%3A%2F%2F139.162.2.61%3A3000%2F%22%2C%22ROOT_URL_PATH_PREFIX%22%3A%22%22%2C%22DDP_DEFAULT_CONNECTION_URL%22%3A%22http%3A%2F%2F139.162.2.61%3A3000%2F%22%2C%22autoupdateVersionCordova%22%3A%2212b3d5a0ecc9b973a11be1b8ac8e8762acd4934f%22%2C%22appId%22%3A%221kzvx321w733oi1ac503b%22%2C%22meteorEnv%22%3A%7B%22NODE_ENV%22%3A%22production%22%2C%22TEST_METADATA%22%3A%22%7B%7D%22%7D%7D"));

  if (/Android/i.test(navigator.userAgent)) {
    // When Android app is emulated, it cannot connect to localhost,
    // instead it should connect to 10.0.2.2
    // (unless we're using an http proxy; then it works!)
    if (!__meteor_runtime_config__.httpProxyPort) {
      __meteor_runtime_config__.ROOT_URL = (__meteor_runtime_config__.ROOT_URL || '').replace(/localhost/i, '10.0.2.2');
      __meteor_runtime_config__.DDP_DEFAULT_CONNECTION_URL = (__meteor_runtime_config__.DDP_DEFAULT_CONNECTION_URL || '').replace(/localhost/i, '10.0.2.2');
    }
  }
</script>
```

i used to asked once

https://github.com/meteor/meteor/issues/3852#issuecomment-76883310

```
 meteor help build
```

this help mislead

```
--server            Location where mobile builds connect to the Meteor server.
                  Defaults to localhost:3000. Can include a URL scheme
                  (for example, --server=https://example.com:443).
```

crapthings:
to make hot code push

[...]

PR #8142, opened 2016-12-12
Cordova project preparation must occur before copying to the build

abernix:
This fixes a regression caused by 88d43a0f16a484a5716050cb7de8066b126c7b28 which is demonstrated in meteor/meteor#7849.

Essentially, with the current implementation some Cordova build elements are "stale" when the build is copied.  For example, if you execute a `meteor run ios` and then `meteor build . --server=http://example.com/` (note: `example.com`) the `config.xml` (`<access origin />`), the boilerplate HTML (`__meteor_runtime_config__`) and other elements of the bundle (`Info.plist` on iOS) will still contain the previously used `http://<local_ip>:3000` address instead of `http://example.com` as they should.

Additionally, it would appear that it's impossible to actually checkout a project and immediately run `meteor build` without running `meteor run (android|ios)` first.

Various work-arounds for this seem to exist, such as running `meteor build` twice, or running `meteor run --server=http://production.com` first.

Ultimately, this is occurring because the bundle is being copied before the Cordova `prepareForPlatform` occurs which I believe was not intended.

There is already a test in place which fails without this fix, but marked as `slow` and therefore not executed on CircleCI.  Specifically, `cordova builds with server options` would have caught this.  Forcibly running this test locally now passes with this change.

Fixes meteor/meteor#7849
Fixes meteor/meteor#7291
Fixes meteor/meteor#6756

Fix-Maybe #7112

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1016,12 +1016,22 @@ ${cordova.displayNameForPlatform(platform)}` }, () => {
               'platforms', platform);
             const platformOutputPath = files.pathJoin(outputPath, platform);
 
+            // Prepare the project once again to ensure that it is up to date
+            // with current build options.  For example, --server=example.com
+            // is utilized in the Cordova builder to write boilerplate HTML and
+            // various config.xml settings (e.g. access policies)
+            if (platform === 'ios') {
+              cordovaProject.prepareForPlatform(platform, buildOptions);
+            } else if (platform === 'android') {
+              cordovaProject.buildForPlatform(platform, buildOptions);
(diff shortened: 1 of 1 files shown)


martijnwalraven:
@abernix: Thanks a lot for looking into these long standing Cordova issues! Your explanation and fix make sense. I don't currently have a Meteor environment installed so I can't verify it, but the code looks good to me.

d62remi:
Do you plan to make a patch for 1.4.2.3 version without update to 1.4.3 with this PR ?

Issue #7954, opened 2016-10-24
Mongo.Collection throws EventEmitter


gsabran:
This code:
```js
try {
  var coll = new Mongo.Collection('__dummy_coll_' + Random.id());  
} catch (e) {
  console.log('error!', e);
  throw e
}
```
is throwing an error that is 
```
error! EventEmitter {
   id: 1,
   s: 
    { options: 
       { disconnectHandler: [Object],
         cursorFactory: [Object],
...
```

This only happens with a real `MONGO_URL`, and not why any url so it might be related to the database configuration. I started Meteor with:
`$  MONGO_URL=mongodb://<user>:<password>@<what-compose.io-told-me-to-put-here> meteor`

This still happens in `1.4.2-rc.3`, but and also in versions since `1.4.1` that I've tested (and possibly earlier)

So I don't know why this is throwing, and why this is throwing an `EventEmitter`

My entire application's code is the default code, replacing `server/main.js` by:

```js
import { Meteor } from 'meteor/meteor';
import { Random } from 'meteor/random';
import { Mongo } from 'meteor/mongo';

Meteor.startup(() => {
  // code to run on server at startup
});

try {
  var coll = new Mongo.Collection('__dummy_coll_' + Random.id());  
} catch (e) {
  console.log('error!', e);
  throw e
}
```

gsabran:
This might be a specific error since reading Meteor's code, I've been able to reproduce it in pure node with:

```js
var MongoDB = require('mongodb');

var url = 'mongodb://<user>:<password>@<what-compose.io-told-me-to-put-here>';
var mongoOptions = {
  db: { safe: true, native_parser: false },
  server: { autoReconnect: true, reconnectTries: Infinity },
  replSet: {}
};

var err;
var cb = function (e, r) { console.log('error:', !!e, 'res', !!r); err = e; };

MongoDB.connect(url, mongoOptions, cb);
```

that logs a similar error. Not really where to go from here!

abernix:
Hmm, when I run the Meteor repo reproduction you provided against my MongoDB (I don't have Compose), it starts up successfully.

And when I run your pure Node script, it returns:

    error: false res true

The `res` (`r`) is indeed an `EventEmitter`, but that's correct as a [`MongoDB.Db`](https://mongodb.github.io/node-mongodb-native/2.2/api/Db.html) is an object that can emit events.  The `error` is not set.   Could this be a problem with connecting to your MongoDB?

[...]

PR #8150, opened 2016-12-13
Upgrade `mongodb` npm package to version 2.2.15.

### LINKED ISSUE #8147: MONGO_URL with multiple hosts throws useless error if any of them are invalid.
I haven't completely tracked down what is causing it, but since I just spent the better part of two days figuring out what the problem actually was I figured I should get it reported..

When starting Meteor with a MONGO_URL that includes more than one mongodb server, if one of them is invalid then the error message that results is entirely unhelpful, consisting of just (yes, this is indeed the entire error):

```
/Users/jasonkohles/.meteor/packages/meteor-tool/.1.4.2_3.bdala++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:313
            throw(ex);
            ^
[object Object]
```

I'm running Meteor 1.4.2.3, and have confirmed that you can very easily reproduce this with pretty much any meteor app:

```
git clone https://github.com/meteor/todos
cd todos
meteor npm install
export MONGO_URL=mongodb://foo,bar/blah
meteor
# (here Meteor crashes with [object Object] error message)
```

abernix:
Trivial PR, but this bumps the `mongodb` version to [2.2.15](https://github.com/mongodb/node-mongodb-native/blob/2.2/HISTORY.md#2215-2016-12-10).  This brings in [`mongodb-core@2.1.2`](https://github.com/christkv/mongodb-core/blob/2.0/HISTORY.md#212-2016-12-10) which has a major version bump for `bson`, going from 0.5.7 to 1.0.1.

Fixes #8147
Relates to #7954 (already closed)

### DIFF
--- a/packages/npm-mongo/.npm/package/npm-shrinkwrap.json
+++ b/packages/npm-mongo/.npm/package/npm-shrinkwrap.json
@@ -1,9 +1,9 @@
 {
   "dependencies": {
     "bson": {
-      "version": "0.5.6",
-      "resolved": "https://registry.npmjs.org/bson/-/bson-0.5.6.tgz",
-      "from": "bson@>=0.5.6 <0.6.0"
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/bson/-/bson-1.0.1.tgz",
+      "from": "bson@>=1.0.1 <1.1.0"
     },
     "buffer-shims": {
       "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


abernix:
Many tests failed for this on CircleCI (like `source maps from an app`) and others timed out, but I'm inclined to believe that it's because of the custom warehouse in use on some tests due to `npm-mongo`'s presence in the tool's isopackets?

abernix:
Recalling this as it's maybe not so trivial.  Despite the fact that they [just released 2.2.16](https://github.com/mongodb/node-mongodb-native/blob/2.2/HISTORY.md#2216-2016-12-13) which fixes a bug that would have likely been a problem in 2.2.15, it also seems something about the bump from 2.2.14 to 2.2.15 or 2.2.16 is problematic.

Will try to revisit soon.

Issue #8158, opened 2016-12-15
Missing users index


elie222:
I think the following index is missing:
```
Meteor.startup(() => {
  Meteor.users.rawCollection().createIndex({ 'services.password.reset.when': 1 });
});
```

I had some of these queries that caused our database to lock up because this index was missing.

stubailo:
Seems like that would go in here: https://github.com/meteor/meteor/blob/8e32da57a128a70fdc09823b5d506758ce6a9b82/packages/accounts-password/password_server.js#L1084

Would be great to see a PR!

abernix:
This has been merged.  Thanks for filing this and for the PR, @elie222!

elie222:
No problem!

elie222:
The index on an array wasn't an issue in the end?

abernix:
Well, it's a potential issue at very high scale much in the same way as #8084 and will need to be fixed in a similar way, but for now it's important to implement as-is.

PR #8163, opened 2016-12-15
Added an index for the users collection

elie222:
An example of a slow query that was slowing down my entire database that this PR should fix:

```
query wc_draft.users query: { $or: [ { services.password.reset.when: { $lt: new Date(1481501706951) } }, { services.password.reset.when: { $lt: 1481501706951.0 } } ] } plan
Summary: COLLSCAN ntoreturn:1000 ntoskip:0 nscanned:64642 nscannedObjects:64642 keyUpdates:0 numYields:208 locks(micros) r:1782624 nreturned:0 reslen:20 1028ms
```

See the discussion here:
https://github.com/meteor/meteor/issues/8158#issuecomment-267456752

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -1061,4 +1061,5 @@ Meteor.users._ensureIndex('services.email.verificationTokens.token',
                           {unique: 1, sparse: 1});
 Meteor.users._ensureIndex('services.password.reset.token',
                           {unique: 1, sparse: 1});
-Meteor.users._ensureIndex('services.password.reset.when');
+Meteor.users._ensureIndex('services.password.reset.when',
+                          {sparse: 1});


glasser:
Did you write that query yourself or is it part of Meteor?

abernix:
That query is from [`expirePasswordToken` in `accounts_server.js`](https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_server.js#L1078-L1091) which I believe would be getting called at an interval in `setExpireTokensInterval`.

glasser:
Ah, yes, looks like the new password reset token expiry feature added in 1.4.1 with https://github.com/meteor/meteor/pull/7534 should have included an index.

I do agree that #8084 suggests that indexes under arrays should be avoided, though.  The login token issue in #8084 is more likely to arise accidentally than with reset tokens but it does create a DOS risk anyway.

elie222:
There are other problematic queries on the users collection too.

This is another one:

query: { services.accessTokens.tokens: { $exists: true, $ne: [] } }
planSummary: COLLSCAN ntoreturn:1000 ntoskip:0 nscanned:64690
nscannedObjects:64690 keyUpdates:0 numYields:1095 locks(micros)
r:3539650 nreturned:0 reslen:20 2785ms

abernix:
@elie222 Any updates on this?  Do you think [`sparse`](https://docs.mongodb.com/manual/core/index-sparse/) is appropriate for this?

elie222:
Yes, I think so.

abernix:
Thanks, @laosb.  LGTM.

glasser:
Hmm, I think this index should be added in the same package that does the query: the query is in accounts-base but the index is in accounts-password.

If you use accounts-base but not accounts-password, the interval is still going to exist but it's going to be doing an unindexed query, which is bad.

I'm not really sure I understand why this code is in accounts-base at all; it seems like it would make much more sense to have it in accounts-password, but in any case the index should be in the same place as the interval.

glasser:
(I suppose you could also use weak dependencies or something to allow accounts-base to only run that code if accounts-password is installed, but I really don't get why it can't just be in accounts-password.)

abernix:
@glasser I think your confusion about why this isn't in [`accounts_password`](https://github.com/meteor/meteor/tree/devel/packages/accounts-password) is warranted.  This particular issue didn't rectify the already mis-placed code, but this stems from in https://github.com/meteor/meteor/pull/7534, which put it there in the first place.

I've opened #8448 to track this.  Thanks!

liitfr:
Hi team !
First of, thanks for your awesome work !

I've just tried Robo3T - a mongo client on my current project.  
Looks like Robo3T throws a   

> Cannot load list of indexes.
> 
> Error:
> Wrong type for field (sparse) int != 8

when I want to list user's indexes. If I drop all indexes and re-create all of them this way :

```
[
...
    {
        "v" : 1,
        "unique" : true,
        "key" : {
            "username" : 1
        },
        "name" : "username_1",
        "ns" : "meteor.users",
        "sparse" : true
    }
...
]
```

Then it works.  
Well, it's maybe not significant, but it looks like `sparse` should equal `true`, not `1` (mongo documentation uses `true` as well)

I'm a noob. Maybe it's not relevant & it's just that Robo3T is somehow too much strict ...
But it could worth this little change ...
  
Cheers & thanks again

abernix:
@liitfr The message `Wrong type for field (sparse) int != 8` seems like an error from Robo3T.  Perhaps you could, using the regular `mongo` shell (available in Meteor with `meteor mongo` from the app root directory) show the output of `db.users.getIndexes()` before and after your changes and also include the Meteor and MongoDB version you're using?

Issue #7829, opened 2016-09-27
Provide global configuration for build plugins


mitar:
So while one can configure build plugins by passing options to `api.addFiles`, there is no such mechanism for app-level files. For example, #7727 would need that.

A proposed idea was to use and parse `package.json` for that.

mitar:
cc @tcastelli, would you be interested in working on a pull request for this?

tcastelli:
I have never tried to do that from package.json, but I could check how postcss and get involved into the pull request, (especially if that could help Blaze getting those nice configs :) )

To be clear, If I understand correctly the idea is to create a way to define custom options for packages that could have special build options. If so, I have some questions
- Would these options available for any kind of package(npm/atmosphere)?
- Is there any reference design/conventions that we must follow for this PR (or we can structure it as we want)?
- Limitations?

mitar:
I do not yet have answers to those questions. :-) I think you will have to investigate it.

Probably anything in Meteor could access those options, but the primary use for me is to access them from build plugins. For example, see https://github.com/meteor/meteor/pull/7727 how I configure build plugin to use `api.addFiles`.

We probably want in `package.json` section for all options, and then for each of packages their own key, which would then have custom values plugin or whoever could access. Something like:

```json
{
  "packageOptions": {
    "stylus": {
      "autoprefixer": {
        "browsers": ["last 3 Chrome versions", "Firefox >= 7", "Explorer >= 8", "last 2 versions", "> 1%", "Firefox ESR", "Safari >= 4"]
      }
    }
  }
}
```

sebakerckhof:
I remember @stubailo opening a related issue when we were discussing autoprefixing: https://github.com/meteor/meteor/issues/5512 .

sebakerckhof:
Ow, I see you already found that one. Anyway, it might have some relevant discussion on how to read the config file.

mitar:
Yes, that was only for minimizers. I am thinking that those options could be used by any package, including build plugins.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

[...]

Issue #6791, opened 2016-04-11
Scan node_modules for packages that can be treated as local Meteor packages


benjamn:
Publishing Meteor packages to npm seems fairly straightforward if we treat `node_modules` directories as similar to `packages` directories, and consider any package with a `package.js` file to be a Meteor package.

This would mean Meteor packages could be hosted by npm, and could also be installed with `npm install` or `meteor npm install`.

The main complexity I can see is that, when importing from such packages, we could use either the npm package name or the `meteor/name` currently used to refer to Meteor packages. It might be possible to support both styles, though!

jiku:
Great idea. If NPM doesn't suddenly begin collecting module 'stack variants' (npm-module, react-npm-module, angular-npm-module, meteor-npm-module to npm-module with variant specified in packages.json metadata or on import) and most modules having Meteor-specific functionality are published as `meteor-npm-module` it makes sense to drop the `meteor/` prefix I think. But might be good to have the option.

Would you want to keep ´package.js´ or transition to something like ´meteor.js´?

benjamn:
I'm thinking the `package.json` file could have a `meteor` section that (among other uses) optionally specifies the name of the control file, e.g.
```js
{
  "name": "user:name",
  "main": "server.js",
  "browser": "client.js",
  "meteor": {
    "control-file": "package.js",
    ...
  }
}
```

jiku:
Yeah that would be even better.

laosb:
Looks good. But what if there comes a normal npm package with a `package.js` in it? And, what if a non-Meteor user use this package by mistake?

jiku:
@laosb hopefully that won't happen. Enforcing use of `meteor.control-file` in `packages.json`, people have to point to a specific file. Even then, or if someone points to the wrong file in `packages.json`, Meteor would still execute it structurally. Not just eager-load the whole file, self-executing functions and all.

talha-asad:
@benjamn I like the idea of putting in a `meteor` section in package.json, instead of using package.js file.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

[...]

PR #8165, opened 2016-12-16
Option to automatically update app node_modules to match package.json, etc.

mattmccutchen:
As of 2020-08-27, I'm making the motivation and design discoverable at the top.  I could file a separate issue in the meteor-feature-requests repo, but filipenevola suggests that's an unnecessary hassle.  Scroll to the bottom for recent discussion.

Meteor should have an option to make all relevant commands automatically install or update app-level npm packages to match `package.json` or similar configuration files.  After Meteor jumped through so many hoops to automate things (including updating of npm dependencies of local Atmosphere packages!) to maintain the single command to run an app from a checkout, it's deeply disappointing that Meteor abandoned this by encouraging app-level npm packages with no support for automatic updating.  In fact, my most important Meteor-based app uses a local Atmosphere package to update the npm packages it needs, and we have held off on migrating it to app-level npm packages because of the lack of automatic updating.

This feature request was originally raised in #6848; benjamn was very skeptical and closed the issue but left the door open for me to submit a PR.  Since then, filipenevola has stated a desire to add the feature.

The original design, based on benjamn's desire not to couple the feature to the npm tool at a time when it had serious limitations compared to other tools for installing npm packages, was to have the user opt in by specifying an arbitrary command to update the packages and a set of files Meteor should watch to trigger re-running the command.  However, my understanding is that since then, the npm tool has largely caught up to its competitors, so it may now make sense to add functionality to Meteor that is specific to the npm tool if there are significant benefits.

The specific proposed syntax to be placed in `package.json` was like this:
```json
  "meteor": {
    "npmAutoInstall": {
       "command": "unset TOOL_NODE_FLAGS && meteor npm prune && meteor npm install",
       "watchSpec": ["package.json", "npm-shrinkwrap.json"]
    }
  }
```
I have not yet checked whether it should be updated due to changes to Meteor since December 2016.  Suggestions welcome.

The implementation was to remember the hashes of the watched files in a hidden file in `node_modules`, not just in memory, so Meteor could avoid running the command even once on startup in case it was slow.  This assumes the user doesn't make manual changes to `node_modules` and expect Meteor to revert them, except that if the user blows away `node_modules` completely (e.g., via the version control system's "clean" command), Meteor will recreate it the next time it is started.

filipenevola notes:
> Another thing that we could try to tackle in the same PR is when you change to a Meteor version that uses a different version of Node.js, this sometimes break your binary dependencies on node_modules and just npm install is not enough to solve, you need to remove your binaries first (maybe there is a npm command for that? not sure) and install.

I have not yet learned how to implement this.

---

As described at:
https://github.com/meteor/meteor/issues/6848#issuecomment-264986992

The basic functionality seems to work, but some fine-tuning may be needed and the code could use cleanup (and probably a self-test case).  @benjamn, may I have your feedback before I put more work into refining this?  Thanks.

Test app:
https://github.com/mattmccutchen/meteor-npm-autoinstall-test

You can switch between the "master" and "with-codemirror" branches.

### DIFF
--- a/tools/isobuild/package-source.js
+++ b/tools/isobuild/package-source.js
@@ -1055,20 +1055,23 @@ _.extend(PackageSource.prototype, {
       return value;
     });
 
-    function makeCacheKey(dir) {
-      return baseCacheKey + "\0" + dir;
-    }
-
     function find(dir, depth, inNodeModules) {
       // Remove trailing slash.
       dir = dir.replace(/\/$/, "");
 
-      // If we're in a node_modules directory, cache the results of the
+      // If this is a node_modules directory, cache the results of the
       // find function for the duration of the process.
-      const cacheKey = inNodeModules && makeCacheKey(dir);
-      if (cacheKey &&
-          cacheKey in self._findSourcesCache) {
-        return self._findSourcesCache[cacheKey];
+      let cacheSaver = null;
+      if (files.pathBasename(dir) == "node_modules") {
+        // Key by absolute path so npmAutoInstall can invalidate by absolute path.
+        const absDir = files.pathJoin(self.sourceRoot, dir);
+        self._findSourcesCache[absDir] = self._findSourcesCache[absDir] || {};
+        if (baseCacheKey in self._findSourcesCache[absDir]) {
+          return self._findSourcesCache[absDir][baseCacheKey];
+        }
+        cacheSaver = (sources) => {
+          self._findSourcesCache[absDir][baseCacheKey] = sources;
+        }
       }
 
       if (loopChecker.check(dir)) {
@@ -1125,8 +1128,8 @@ _.extend(PackageSource.prototype, {
         sources.push(...find(nodeModulesDir, depth + 1, true));
       }
 
-      if (cacheKey) {
-        self._findSourcesCache[cacheKey] = sources;
+      if (cacheSaver) {
+        cacheSaver(sources);
       }
 
       return sources;
@@ -1494,6 +1497,12 @@ _.extend(PackageSource.prototype, {
 
   displayName() {
     return this.name === null ? 'the app' : this.name;
+  }
+});
+
+_.extend(PackageSource, {
+  invalidateSourcesCache(absDir) {
+    delete PackageSource.prototype._findSourcesCache[absDir];
   }
 });
 
--- a/tools/project-context.js
+++ b/tools/project-context.js
@@ -10,12 +10,15 @@ var files = require('./fs/files.js');
 var isopackCacheModule = require('./isobuild/isopack-cache.js');
 var isopackets = require('./tool-env/isopackets.js');
 var packageMapModule = require('./packaging/package-map.js');
+var PackageSource = require('./isobuild/package-source.js');
 var release = require('./packaging/release.js');
 var tropohouse = require('./packaging/tropohouse.js');
 var utils = require('./utils/utils.js');
 var watch = require('./fs/watch.js');
 var Profile = require('./tool-env/profile.js').Profile;
 import { KNOWN_ISOBUILD_FEATURE_PACKAGES } from './isobuild/compiler.js';
+import child_process from 'child_process';
+import { optimisticReadJsonOrNull } from './fs/optimistic.js';
 
 // The ProjectContext represents all the context associated with an app:
 // metadata files in the `.meteor` directory, the choice of package versions
@@ -54,7 +57,8 @@ var STAGE = {
   RESOLVE_CONSTRAINTS: '_downloadMissingPackages',
   DOWNLOAD_MISSING_PACKAGES: '_buildLocalPackages',
   BUILD_LOCAL_PACKAGES: '_saveChangedMetadata',
-  SAVE_CHANGED_METADATA: 'DONE'
+  SAVE_CHANGED_METADATA: '_npmAutoInstall',
+  NPM_AUTO_INSTALL: 'DONE'
 };
 
 _.extend(ProjectContext.prototype, {
@@ -203,6 +207,9 @@ _.extend(ProjectContext.prototype, {
     // Initialized by _buildLocalPackages.
     self.isopackCache = null;
 
+    // Initialized by _npmAutoInstall.
+    self._npmAutoInstallWatchSet = null;
+
     self._completedStage = STAGE.INITIAL;
 
     // The resolverResultCache is used by the constraint solver; to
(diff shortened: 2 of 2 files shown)


abernix:
Sorry for not responding on this a very long time ago, with any comment at all; we realize you put some time into this.  This implementation might have some merit, especially as it seems to offer a very specific way to opt in and generally stay out of the way otherwise.

That said, it does conjure up a potential bottleneck in builds which requires additional consideration (to ensure that the blame is not put on Meteor for this slowness).  We haven't been able to give a lot of additional thought to it yet, but hope to revisit it soon.

One sub-project which should be done first would be a facility/library which handles a `meteor` section of `package.json`, watching it as appropriate (possibly as a dependency/component of [`project-context.js`](https://github.com/meteor/meteor/blob/devel/tools/project-context.js)?).

We've discussed this briefly in other issues: https://github.com/meteor/meteor/issues/6791#issuecomment-208959968, https://github.com/meteor/meteor/issues/8341#issuecomment-278747356 and (kinda, but eventually) https://github.com/meteor/meteor/issues/7829.  Perhaps we should get an issue open for that design first.

mattmccutchen:
Thanks for the update, @abernix.  I put out a prototype in the hope of building momentum around the feature, but I don't see myself spending additional hours trying to design infrastructure to handle use cases with which I have no familiarity with no guarantee that this goes anywhere.  I went ahead and filed #8445.  If you can provide a more detailed design sketch for #8445 and give me some grounds to trust that you'd work with me to get both PRs merged, I may try my hand at it.  Otherwise, I think I'm leaving this behind unless/until the other people who claimed to want the original feature care enough to get involved.  I'm disappointed to see Meteor give up [the single command to run from a checkout](https://github.com/meteor/meteor/issues/6848#issuecomment-212184034) after having jumped through so many hoops to automate other things, but I have to go on with life.

mattmccutchen:
I noticed the milestone settings.  Meteor maintainers, if you're serious about getting this functionality into an upcoming release, let me know and I will be happy to help.  I can rebase the change to the current "devel" branch of Meteor and retest it; then my request stands for feedback before I put more time into polishing it.

filipenevola:
The CLA is still not signed, closing because of this.

mattmccutchen:
I completed the process to sign the CLA, though the CLAassistant comment does not update even if I press the "recheck" link.  I'm not sure why I didn't attend to this the first time.  Please reopen this pull request unless you have also decided to reject the change on the merits; otherwise I might eventually submit a new pull request, or someone else could now that my prototype code is clearly licensed.  (I still doubt that the change will go anywhere, but I don't want it dropped from consideration for a silly reason like the CLA.)

stolinski:
Bump bump. This was marked as closed, but would be a nice "meteor like" addition to the platform.

filipenevola:
Hey @stolinski @mattmccutchen I like this idea my only concern is that we need to have a way to enable or disable this as in some cases you don't want this to happen automatically.

I believe Matt agrees with me as he said this feature would be opt-in:
> I don't see what the problem is when the feature is opt-in and Meteor clearly logs when it is calling the package manager.

Besides that I'm ok and we can create a new PR and start to test these changes.

Another thing that we could try to tackle in the same PR is when you change to a Meteor version that uses a different version of Node.js, this sometimes break your binary dependencies on node_modules and just npm install is not enough to solve, you need to remove your binaries first (maybe there is a npm command for that? not sure) and install.

Also I like the idea of having hooks, so anyone can customize what happens after/before these updates, like use yarn, etc.

@mattmccutchen are you willing to proceed? I can guarantee that you are going to have support from our side. I believe it would be better to open a new PR as CLA check was failing 😞 

If you prefer you could start an issue describing the solution that you would implement, the features that it would provide, the configuration options, etc.

@zodern feel free to add any ideas or concerns here as well.

mattmccutchen:
This is a huge and welcome surprise!  I should be able to resume work on this feature in the next few days.

> Hey @stolinski @mattmccutchen I like this idea my only concern is that we need to have a way to enable or disable this as in some cases you don't want this to happen automatically.
> 
> I believe Matt agrees with me as he said this feature would be opt-in:
> 
> > I don't see what the problem is when the feature is opt-in and Meteor clearly logs when it is calling the package manager.

To clarify, I don't have a personal stance on whether the feature should be on by default but was deferring to the previous Meteor maintainers and will similarly defer to you.  At a minimum, it should be possible to turn it off.

> Another thing that we could try to tackle in the same PR is when you change to a Meteor version that uses a different version of Node.js, this sometimes break your binary dependencies on node_modules and just npm install is not enough to solve, you need to remove your binaries first (maybe there is a npm command for that? not sure) and install.

I agree we should investigate this problem, whether the solution ends up being in Meteor itself or just documenting the way users need to configure Meteor to solve the problem.  I have no prior knowledge of this problem.  Before I research it myself, would anyone care to provide links to more information?

> I believe it would be better to open a new PR as CLA check was failing

The CLA check seems to be fixed, so I think it makes sense to continue in this PR rather than force everyone to re-subscribe to a new one.  Of course, I will satisfy all the requirements that would apply to a new PR.  Let me know if you have another reason to start a new PR.

> If you prefer you could start an issue describing the solution that you would implement, the features that it would provide, the configuration options, etc.

I agree this would be appropriate and will file an issue in the meteor-feature-requests repo soon.

filipenevola:
Hey @mattmccutchen go ahead with this PR, you don't need to open a new one. We can also discuss possible solutions here, you don't need to open a feature request.

Feel free to mention me here if necessary or DM me on [Community Slack](https://join.slack.com/t/meteor-community/shared_invite/enQtODA0NTU2Nzk5MTA3LWY5NGMxMWRjZDgzYWMyMTEyYTQ3MTcwZmU2YjM5MTY3MjJkZjQ0NWRjOGZlYmIxZjFlYTA5Mjg4OTk3ODRiOTc), I'm always around.

mattmccutchen:
I have now summarized the motivation and old design at the top of this PR.  Feedback would be much appreciated before I set about updating the design and implementation.

hexsprite:
> I have not yet learned how to implement this.

I believe it's: `npm rebuild --build-from-source`

that should rebuild all binaries from source for the current Node version.

So you'd want to track which node version was used to install the current `node_modules` directory, and then run that commands if the version changes (Major version change I guess?).

filipenevola:
Hi @mattmccutchen thank you for your updates in the original description.

And I believe @hexsprite is right about his command to recompile dependencies.

mattmccutchen:
A quick update: This is still on my to-do list.  I've had a succession of higher-priority things come up and I'm starting a part-time job, but I hope to get back to it if at all possible in the next week or two.  The textual merge with the latest upstream was relatively trouble-free and I can publish it here if anyone wants, but I hoped to put in some more work to understand the current Meteor build system and ensure that I'm integrating with it properly.  If it seems I won't be able to proceed with this work by myself in a timely manner, I can hand it off to someone else or post the specific questions I'm stuck on here for knowledgeable people to answer.  Thanks filipenevola for the encouragement!

StorytellerCZ:
@matteodem are you still working on this?

mattmccutchen:
I'm assuming that mention was intended for me.

Unfortunately, development on my project that uses Meteor is on indefinite hold.  As long as that remains the case, I think it's unlikely that I'll ever feel personally motivated to work on finishing this PR as opposed to my job and my many other hobby projects.  Also, I no longer feel that I have an adequate sense of how useful the feature would be to users in light of any changes to Meteor that may have occurred since I originally proposed the feature.  I went ahead and published [my rebased commit as of 2020-09-11](https://github.com/mattmccutchen/meteor/commit/d7db95e92355f8d2ac79f58357c3d3c0d941550d) in case it's helpful to anyone.

Regrets for not providing an explicit update sooner (although I suspect people could have guessed what happened on my side).  I'm guessing the inquiry was in preparation for closing the PR to get it off your list, so I'll go ahead and be the one to close it so that GitHub gives me permission to reopen it if need be. :slightly_smiling_face:  (I wish GitHub had a richer set of statuses to help keep currently unactionable items from being distracting rather than just this generic "closed" status, but that's another story...)

If someone else is interested in helping with the work, then that would be a sign that the community actually cares and I would be likely to get involved again.

Issue #8160, opened 2016-12-15
Using arrays defined in vm not recognized by Blaze #each directive


brucejo75:
I am creating arrays in a VM.  Then I am using these arrays in my Blaze template.  Blaze does not recognize that these are valid arrays.

## Environment
meteor v1.4.2.3

## Explanation

Blaze depends on the Meteor observe_sequence package.  And the observe_sequence package uses `instanceof Array` to determine if the sequence is an array.  `instanceof Array` does not work across windows, because there are 2 instances of `Array` one in each window.

There is a writeup [here](http://web.mit.edu/jwalden/www/isArray.html).

## Fix
change these 2 lines to use `Array.isArray(seq)` instead of depending on `instanceof Array`.
[observe_sequence.js#L97](https://github.com/meteor/meteor/blob/fe143dd84752bc3dea42ea35d44c2d4e93ba14f5/packages/observe-sequence/observe_sequence.js#L97)
[observe_sequence.js#L129](https://github.com/meteor/meteor/blob/fe143dd84752bc3dea42ea35d44c2d4e93ba14f5/packages/observe-sequence/observe_sequence.js#L129)

stubailo:
Would be great to see a PR for this - looks like a simple fix!

mitar:
I think we should just use `_.isArray`.

stubailo:
Isn't `Array.isArray` just a standard version of the exact same thing?

mitar:
Yes, and Underscore will use it if it is available, no?

mitar:
BTW, what is VM?

stubailo:
Yeah I guess avoiding underscore dependencies is a lost cause at this point.

mitar:
I mean, we should replace underscore with lodash, but otherwise it is a good idea to have a layer of abstraction in between.

brucejo75:
I am using `vm.browserify` to do my evals to keep the evals safe from global corruption.  They refer to running it in a `vm` (aka virtual machine), but I think it is just really another window.

abernix:
@brucejo75 Would there be anything stopping you from making a PR for this?  In other words, is there anything I/Meteor/we can do to help you submit your first PR?

brucejo75:
Thanks @abernix.  I will take a shot at it. Never submitted a PR to Meteor.  I will carve some time today.

brucejo75:
I may need some help.  Here is what I did:

1. forked meteor
2. made code change in `observe-sequence.js`
[...]

PR #8166, opened 2016-12-17
fix for #8160

brucejo75:
Fix for #8160.

Simple fix to enable arrays created in foreign windows to be valid for Blaze `{{#each}}` loops.

Just changed `instanceof` check to use `_.isArray()`.

### DIFF
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -94,7 +94,7 @@ ObserveSequence = {
 
         if (!seq) {
           seqArray = seqChangedToEmpty(lastSeqArray, callbacks);
-        } else if (seq instanceof Array) {
+        } else if (_.isArray(seq)) {
           seqArray = seqChangedToArray(lastSeqArray, seq, callbacks);
         } else if (isStoreCursor(seq)) {
           var result /* [seqArray, activeObserveHandle] */ =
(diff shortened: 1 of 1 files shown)


benjamn:
Yeah, `instanceof` is an unfortunate pitfall in many cases (duplicate npm packages, too). Thanks for fixing this!

brucejo75:
Hi @ManuelDeLeon, can you give me a repro?  I tried this on the console:

```
> var a = new Array()
undefined
>a
[]
>_.isArray(a)
true
> var b = _.extend(a, {'foo': function foo() { return true;}})
undefined
> b.foo()
true
> _.isArray(b)
true
```
Seems to work.

ManuelDeLeon:
I meant `class MyArray extends Array`. Here's a repro: https://github.com/ManuelDeLeon/test2
```js
Template.body.viewmodel({
  items: [1, 2],
  log_and_return_items: function () {
    console.log(this.items());
    console.log("instanceof Array", this.items() instanceof Array);
    console.log("isArray", _.isArray(this.items()));
    return this.items();
  }
});
```
```html
<body>
    {{#each log_and_return_items}} 
        {{ this }} 
    {{/each}}
</body>
```

The array is converted to `ReactiveArray` and it logs:

```
[1, 2]
instanceof Array true
isArray false
Uncaught Error: {{#each}} currently only accepts arrays, cursors or falsey values.
```

brucejo75:
I see.  `Array.isArray` does not work either.  My first PR too, :-(.

I guess the best we can do is:

**Main Window**: Support Arrays & sub-classed arrays in main window.

**Secondary Window**: Support arrays on Secondary Windows, no sub-classed arrays.  Not sure sub-classed arrays that came from a secondary window make sense anyway, any functions would be in the secondary window Context.

I will work on a new PR with your recommendation.

mitar:
I think `seq instanceof Array || _.isArray(seq)` sounds like a great solution to this.

Issue #3786, opened 2015-02-21
new Meteor.Collection(null) and keys with dots vs a server collection


RobertLowe:
When using a `new Meteor.Collection(null)` on the client side, you can have keys with `.`'s in them.

So for instance it's possible to insert: `http://dbpedia.org/data/George_Carlin.json` into the local collection. However, on the server with a Mongo backed connection you cannot persist keys with `.`'s in them.

`Exception while invoking method '/Thing/insert' Error: key http://dbpedia.org/resource/George_Carlin must not contain '.'`

Which makes sense as Mongo reserves `.` & `$` characters. 

> Field names cannot contain dots (i.e. .) or null characters, and they must not start with a dollar sign (i.e. $). See Dollar Sign Operator Escaping for an alternate approach.

http://docs.mongodb.org/manual/reference/limits/#Restrictions-on-Field-Names

This issue requests that the behaviour between Client & Server be unambiguous.

Thoughts?

Slava:
I think @RobertLowe has found a valid difference in behavior between Mongo and Minimongo:

```
> c = new Package.minimongo.LocalCollection
< d {name: undefined, _docs: d._IdMap, _observeQueue: e._SynchronousQueue, next_qid: 1, queries: Object…}
> c.insert({'a.b': 'c'})
< "rXiJrw4CmDQju5axq"
```

This looks rather harmless but a check in LocalCollection methods wouldn't hurt.

hwillson:
I've submitted PR #8169 to prohibit the use of dot containing field names during Minimongo inserts. Thanks!

abernix:
#8169 has landed.  Thanks for the fix, @hwillson, and thanks for reporting it originally, @RobertLowe!  This will be included in an upcoming 1.4.x release.

laddi:
I updated to 1.4.2.6 the other day and started seeing this error all over since I've been using LocalCollections to store various data, such as kubernetes labels, that contain keys with dots in them.  I understand why this checks needs to be in place when using database-backed collections but I'm hugely disappointed that I will now have to refactor my code, that is not using database-backed collections, significantly because of it.  This should have been implemented with an opt-out when using LocalCollections to minimize regression errors.

hwillson:
[...]

PR #8169, opened 2016-12-19
Adjusted Minimongo to prohibit dot based field name inserts, to fix issue #3786

hwillson:
Hi all - this PR is intended to help address issue #3786. Right now Minimongo allows inserts with dot (".") containing field names, which is against [Mongo's field name restrictions](https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names). The changes in this PR will cause a new exception to be thrown if dot containing field inserts are attempted by Minimongo, with the exception thrown matching the exception thrown by Mongo on the server side. Thanks!

### DIFF
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -544,6 +544,16 @@ LocalCollection.prototype.insert = function (doc, callback) {
   var self = this;
   doc = EJSON.clone(doc);
 
+  // Make sure field names do not contain dots to line up with Mongo's
+  // field name restrictions:
+  // https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
+  if (doc) {
+    const invalidFieldMatches = JSON.stringify(doc).match(/"([^"]*\.[^"]*)":/);
+    if (invalidFieldMatches && invalidFieldMatches.length === 2) {
+      throw MinimongoError(`Field ${invalidFieldMatches[1]} must not contain '.'`);
+    }
+  }
+
   if (!_.has(doc, '_id')) {
     // if you really want to use ObjectIDs, set this global.
     // Mongo.Collection specifies its own ids and does not use this code.
@@ -1110,4 +1120,3 @@ LocalCollection.prototype.resumeObservers = function () {
   }
   self._observeQueue.drain();
 };
-
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -549,7 +549,7 @@ LocalCollection.prototype.insert = function (doc, callback) {
   // https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
   if (doc) {
     JSON.stringify(doc, (key, value) => {
-      if (key.indexOf('.') > -1) {
+      if (_.isString(key) && key.indexOf('.') > -1) {
         throw MinimongoError(`Field ${key} must not contain '.'`);
       }
       return value;
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -544,13 +544,19 @@ LocalCollection.prototype.insert = function (doc, callback) {
   var self = this;
   doc = EJSON.clone(doc);
 
-  // Make sure field names do not contain dots to line up with Mongo's
-  // field name restrictions:
+  // Make sure field names do not contain Mongo restricted
+  // characters ('.', '$', '\0').
   // https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
   if (doc) {
+    const invalidCharMsg = {
+      '.': "contain '.'",
+      '$': "start with '$'",
+      '\0': "contain null bytes",
+    };
     JSON.stringify(doc, (key, value) => {
-      if (_.isString(key) && key.indexOf('.') > -1) {
-        throw MinimongoError(`Field ${key} must not contain '.'`);
+      let match;
+      if (_.isString(key) && (match = key.match(/^\$|\.|\0/))) {
+        throw MinimongoError(`Key ${key} must not ${invalidCharMsg[match[0]]}`);
       }
       return value;
     });
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -548,10 +548,12 @@ LocalCollection.prototype.insert = function (doc, callback) {
   // field name restrictions:
   // https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
   if (doc) {
-    const invalidFieldMatches = JSON.stringify(doc).match(/"([^"]*\.[^"]*)":/);
-    if (invalidFieldMatches && invalidFieldMatches.length === 2) {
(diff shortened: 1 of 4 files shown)


hwillson:
Hmm - Travis failed but from the error log the failures don't look like anything caused by this PR?

hwillson:
Okay, it is my fault - `ecmascript` is missing (thanks @abernix)! Fixing now.

packages/minimongo/minimongo.js
```diff
@@ -544,6 +544,16 @@ LocalCollection.prototype.insert = function (doc, callback) {
   var self = this;
   doc = EJSON.clone(doc);
 
+  // Make sure field names do not contain dots to line up with Mongo's
+  // field name restrictions:
+  // https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
+  if (doc) {
+    const invalidFieldMatches = JSON.stringify(doc).match(/"([^"]*\.[^"]*)":/);
```
abernix: I believe this RegEx might be susceptible to match inadvertently on the `value` side.  

Consider the JSON:

```json
{"value":"silly quoted JSON in here \"︵ (°.°)/ ︵\": inadvertently matches!  eeps"}
```

...which I believe would throw the error incorrectly.

As an alternative, thoughts on using the [`replacer` function](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) of `JSON.stringify`?

```js
JSON.stringify(doc, (key) => {
  if (key.indexOf(".") > -1) {
    throw MinimongoError(`Field ${key} must not contain '.'`);
  }
  return value;
});
```

We could fix the regex, but I'm just thinking that the built-in function might be faster anyway and more able to handle whatever is thrown at it.

hwillson: Awesome idea! I'll make that change - thanks!

RobertLowe:
Does this check for `.`s or does  this check for any non-sane mongo field name which may be the larger issue? I'm not sure `.`s are the only problematic character, we may want a charlist/wordlist or this could be fine to merge as just a one off until someone complains of the ladder

Nice effort 👍

abernix:
I'm assuming dots are the most commonly encountered reason, but it's reasonable to say that this could/should be expanded to prevent the other disallowed configurations: null characters or starting with a dollar sign.  I'm not sure if anything outside of those three things would be a problem though.

hwillson:
Hi guys - yes, this PR was just addressing the original issue of `'.'`'s. That being said it would be pretty trivial to extend it to handle `$`'s and null characters, so I'll make it happen - thanks for the feedback!

hwillson:
Whoa - @abernix, check out the "All checks have passed" section of this PR. My last commit just completely by-passed CircleCI:

screenshot 2016-12-20 10 19 16

I know why - I originally started this PR with my own instance of CircleCI running checks, then disabled my CircleCI instance just before the last commit, to speed up the CI run using Meteor's CI instances. The act of disabling it then committing means I've just been able to by-pass the CircleCI tests completely, and we're all green! Looks like yet another security hole caused by Circle allowing developer CI instances to override project CI instances 😞 ...

mitar:
I have not checked code, but `$rename` operator should probably also check that you are renaming to a valid name? (If that was allowed before.)

hwillson:
Hi @mitar - good point; I just ran some tests with `$rename` and Minimongo. If you try to rename a field to a new field with a `.`, it creates a new nested field/object, so I think we're okay there. If you try to rename a field to a new field with a `$`, Minimongo throws a `Uncaught Error: can't set field named $a` exception so we're okay there as well. If you try to rename with a null byte however, Minimongo does let it go through! So it looks like we should track down and adjust renames to restrict null bytes. I'll dig in an update this PR as needed. Thanks for the additional insight!

mitar:
Hm, maybe then also looking into $set would be beneficial.

BTW, I opened many tickets about difference between client and server side in the past: https://github.com/meteor/meteor/issues?q=is%3Aissue+is%3Aopen+minimongo+author%3Amitar

You could look into them. How I caught them: I made tests run on client and server. So Meteor tests on the client, and Minimongo tests against the server. I think this should be done and would catch many many differences already. See also closed issues which were not all addressed: https://github.com/meteor/meteor/issues?q=is%3Aissue+minimongo+author%3Amitar+is%3Aclosed

hwillson:
Hi all - I've adjusted `$rename` to make sure it doesn't allow null bytes. @mitar - I had previously verified `$set` doesn't allow dot characters, so no changes were necessary. I just checked for `$`'s and null bytes; `$`'s are prohibited but null bytes are currently allowed as well! My recent changes include a patch for this. I've also reviewed all other Minimongo insert/update combinations that would allow invalid field names, and I think we're now covered. So this PR should now fully address the original issue (plus '$'s and null bytes).

With regards to other differences between Minimongo/Mongo, I know - I've been watching several of your issues. I believe that's a can of worms for another day/PR though ...

Thanks again all!

mitar:
You are amazing! Thanks so much!

mitar:
And for other issues, I think the best first step would be to start running tests on both client and server side for Mongo related stuff.

dreich-nucleus:
This is a little late, but there was an unintended consequence of this fix. When doing an upsert where the query/selector has dotted keys, the upsert fails with the "MinimongoError: Key <query field with dot> must not contain '.'" error.

`collection.insert({ val: 100, ref: { b: '123abc', desc: 'ref to other obj' }})`

`collection.upsert({'ref.b': '345xyz'}, {$set: {val: 1000}, $setOnInsert: {ref: {...full info}}})`

The selector is a valid query and the update code will add the selector to the doc to ensure a second upsert actually finds the new document. But it just adds the `ref.b` to the update doc and that blows up in the insert.

We are still stuck on a fairly old meteor (1.4.2) which actually works with the above code because the minimongo version is 1.0.19. However, something in our full unit test suite is pulling in a slightly newer version (1.0.21) and the unit tests fail. Makes it a little tough to deliver code to master when the tests fail. When we upgrade to a newer version, this issue might get worse.

RobertLowe:
@dreich-nucleus in the meantime you could use `collection.rawCollection()` to access the native mongo driver and consider updating

dreich-nucleus:
The unit tests are mocking the real collections as in memory collections - I did not think rawCollection was available on those.
There is a project to upgrade to 1.7 but that won't land before my project is done. I do have a workaround though. It looks like the issue has been fixed in the latest code though.

mitar:
@dreich-nucleus I think you should open a new issue about this.

dreich-nucleus:
I could do that, but what if it is fixed in 1.7? It would be my first issue here and I don't want to make a silly one. Advice is welcome.

mitar:
I think better to comment in already closed PRs is to open an issue. Report in which version you are finding this problem (this is part of good issue) and yes, maybe it will be closed if it is fixed in newer version. But it will definitely get more eyes if it is a new issue and not discussion inside something already closed.

dreich-nucleus:
Sounds good. I figured the context would be helpful, but I will link this discussion to the new issue. Thanks

dreich-nucleus:
Looks like the issue was #8631 (https://github.com/meteor/meteor/issues/8631). Sadly, that issue never came up on my searches. Oh well.

Issue #7843, opened 2016-09-29
Missing link to WebApp project in readme of autoupdate package


lucfranken:
In followup on #7827 here is a link to "WebApp project" which does not exist:

https://github.com/meteor/meteor/tree/8cdab91bc415433aa693e045534996e5a3ba4f05/packages/autoupdate

https://github.com/meteor/meteor/blame/devel/packages/autoupdate/README.md#L12

It seems there is no real list of "projects" anymore. Which is also referred to by Thea in: https://forums.meteor.com/t/community-involvement-setup/29538/2 as "parts of Meteor" or "components".

Is there such a list available which describes the main projects under which the packages are grouped? So we can link to those and there a developer can get an overview of the specific projects of Meteor? 

It seems to me that it's much easier to get a grasp of parts of Meteor, for developers who want to contribute, when you have an idea of how the packages relate to each other. Now I do that just by hand following which packages include each other.

Search on all occurrences of this specific link:
https://github.com/meteor/meteor/search?utf8=✓&q=%22meteor.com%2Fwebapp%22&type=Code

_Found this while researching: https://github.com/meteor/meteor/issues/7760_

abernix:
The meteor.com/webapp link was probably lost during the Meteor.com redesign which happened a long time ago.  You can still find `webapp` in the `packages` directory that we looked at (and linked to) in #7827.

> Search on all occurrences of this specific link:
> https://github.com/meteor/meteor/search?utf8=✓&q=%22meteor.com%2Fwebapp%22&type=Code

As you can see, those are README files from the other packages within the Meteor project.  I think it's reasonable to update all those links (in that search) to be: `[Webapp](https://github.com/meteor/meteor/tree/devel/packages/webapp)`?

It's possible that those (including the one already accepted in #7827) should be relative links (i.e. `[Webapp](../webapp)`), but I don't have a particular feeling either way.  Relative might be easier to update, but less helpful if viewing the `README` in a normal text editor (e.g. `vi`, `emacs`, Atom, Sublime Text, etc.).

Maybe a `README` for this `packages` directory might be a good "table of contents" for the various projects.?

Thoughts?

lucfranken:
@abernix to be clear I am waiting for @theadactyl as it seems she is working on something https://forums.meteor.com/t/community-involvement-setup/29538/59 which might relate to the organization of this. 

I agree with your approach, we should improve this on Github with readme files. But it needs some agreement internally it seems to organise this so let's wait for that?

[...]

PR #8172, opened 2016-12-19
Updated all webapp project URLs (issue #7843)

hwillson:
Hi all - relating to https://github.com/meteor/meteor/issues/7843#issuecomment-260897892, I believe this should take care of the remaining actionable items in #7843. Let me know if you need anything changed. Thanks!


abernix:
This looks good.  My only further consideration would be: should we point them to `master` or is `devel` good enough?  I'd recommend just making them relative, but since these are displayed in Atmosphere, I think they have to be absolute (I could be wrong?).

hwillson:
Thanks for the feedback @abernix - I think you're right; since `devel` could have changes that aren't officially published yet, we shouldn't point to the `devel` docs. Yes, they have to be absolute for Atmosphere. Changes coming shortly - thanks!

hwillson:
Changes made - thanks!

Issue #7758, opened 2016-09-06
[Meteor 1.4.1.1] Mongo Upsert with multi-attribute query results in undefined object in mongo_driver.js:636


BigDSK:
Environment:
- Meteor 1.4.1.1
- OS: MacOSX 10.11.6
- Dependencies: aldeed:simple-schema; aldeed:collections2

Upsert code that worked in versions 1.3.X and below is now failing in 1.4.1.1 with an error in mongo_driver.js:636. The error and stack are below. An example statement is here (see repo below for full reproduction):

```
  Sample.upsert({
      network_id: new Mongo.ObjectID('44915733af80844fa1cef07a'),
      collection_name: 'blah',
      doc_id: new Mongo.ObjectID('44915733af80844fa1cef07b')
    },
    {
      $set: { thing_to_update: 'item1' }
    });
```

Full reproducible repo is here:

[https://github.com/BigDSK/meteor-upsert-error](https://github.com/BigDSK/meteor-upsert-error)

Please view [/server/main.js](https://github.com/BigDSK/meteor-upsert-error/blob/master/server/main.js#L42) for the collection definition and offending code. Workaround is to replace upsert with findOne()/insert()/update() series.

```
Exception in Mongo write: TypeError: Cannot read property 'indexOf' of undefined
     at Function.MongoConnection._isCannotChangeIdError (packages/mongo/mongo_driver.js:636:7)
     at packages/mongo/mongo_driver.js:751:47
     at runWithEnvironment (packages/meteor/dynamics_nodejs.js:110:1)
     at packages/meteor/dynamics_nodejs.js:123:1
     at handleCallback (/Users/dennis/.meteor/packages/npm-mongo/.1.5.49.12xux8a++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/utils.js:96:12)
     at /Users/dennis/.meteor/packages/npm-mongo/.1.5.49.12xux8a++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/collection.js:1043:20
     at executeWrite (/Users/dennis/.meteor/packages/npm-mongo/.1.5.49.12xux8a++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/wireprotocol/3_2_support.js:80:5)
[...]

PR #8175, opened 2016-12-21
Mongo package linting

dr-dimitru:
mongo_driver.js:
  - Lines 66, 1011 and 1352: Linting
  - Line 497: `return` after `throw`
  - Lines 523 - 527: `e` already defined in same function scope (hoisting)
  - Line 937: Missed semicolon
  - Line 1095: Very confused about ES6 “default parameter” set via colon, which probably will be interpreted as Object - Babel failed to build on my end, regarding [this docs](https://mongodb.github.io/node-mongodb-native/api-generated/cursor.html#count) it should be a Boolean, so we set it to `false` by default
  - Line 1120: no need to set variable to `undefined`
  - Lines 1122, 1128, and 1133: Define `doc` above while due to function scope and hoisting `doc` is already allocated inside `while` and `try`

polling_observe_driver.js
  - `newResults` defined inside try, and used in parent scope later

oplog_tailing.js
  - `lastEntry` defined on while loop (line 139), and used later at parent scope

oplog_observe_driver.js
  - Line 425 and 448: `maxBuffered` is defined twice in same scope (hoisting)

mongo_livedata_tests.js
  - Lines 481 - 498: `val` and `x` is defined twice (hoisting)
  - Lines 973 - 975 and 980 - 982: `expectedRemoves` and `expectedAdds` is  already defined on lines 952 and 955

allow_tests.js
  - Line 16: No need to set variable to `undefined`

package.js
  - Beta version bump

Other files
 - Overall linting, missed semicolons


benjamn:
What linter ruleset are you using here? The long lines from compound conditions seem like a step backwards for readability, though other changes seem like good ideas.

dr-dimitru:
Hi @benjamn ,

I'm using SublimeLinter + Babel + eslint with very default setup.

From `.eslintrc`:
```
    "rules": {
        "indent": [
            "error",
            2
        ],
        "linebreak-style": [
            "error",
            "unix"
        ],
        "quotes": [
            "off",
            "single"
        ],
        "semi": [
            "error",
            "always"
        ],
        "no-console": [
            "off", 
            {
                "allow": ["warn", "error", "info", "log", "trace"]
            }
        ]
    }
```

I can change it back on your demand, changed because even on my 15" screen it has space for those lines, and looks different to me. I guess it depends from developer to developer.

packages/mongo/mongo_driver.js
```diff
@@ -1092,7 +1090,7 @@ _.extend(SynchronousCursor.prototype, {
     return self.map(_.identity);
   },
 
-  count: function (applySkipLimit: false) {
+  count: function (applySkipLimit = false) {
```
abernix: 👍

dr-dimitru:
Hi @abernix ,

Thank you for detailed review. Required changes is pushed.

For some reason I can not run tests for this package unless changes is made, please take a look on [this commit](https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c). What do you think?

packages/mongo/mongo_driver.js
```diff
@@ -1120,18 +1120,19 @@ MongoConnection.prototype.tail = function (cursorDescription, docCallback) {
...
         return;
       try {
-        var doc = cursor._nextObject();
+        doc = cursor._nextObject();
       } catch (err) {
         // There's no good way to figure out if this was actually an error
         // from Mongo. Ah well. But either way, we need to retry the cursor
         // (unless the failure was because the observe got stopped).
-        doc = null;
+        // `doc` is already `null` here;
```
abernix: I don't think this change is safe or just a "linting" change – at least not without further consideration or explanation.

This is in a loop.  While `doc` is `null` the first time around, if `cursor._nextObject` throws an error, it would be left as the _previous_ `doc` that was processed on this cursor (loop).  A tailable cursor remains open until additional documents arrive.

Do you disagree?

abernix:
I don't see how https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c is related to this PR.  Is there an actual issue that you're trying to fix with that commit?

packages/mongo/mongo_driver.js
dr-dimitru: @abernix you're right. My bad.

dr-dimitru:
https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c not really related to this PR.

It's popped-out as I tried to run `meteor test-packages ./`, of course including `MONGO_URL` and `MONGO_OPLOG_URL`. Anyways I'm able to run and pass most of tests only after changes [at this commit](https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c), should I start separate ticket/PR?

abernix:
@dr-dimitru Ok, this LGTM now.  Hopefully someone else can take a look over it too and 👍 it.

Regarding https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c: yes, let's make that a separate issue which shows the problem you're encountering, including specifically _which_ tests are failing.  Also, try to keep each commit/PR to one logical change.   That being said, the "fix" reminds me of something that turned up in https://github.com/meteor/meteor/issues/7758#issuecomment-253076694 (re: `errmsg` vs `err`) which we never resolved. The full test suite is a bit tricky at times – you may find solace in looking at `scripts/ci.sh` in the Meteor repo to see what the CI suite runs.

packages/mongo/collection_tests.js
```diff
@@ -89,7 +89,13 @@ Tinytest.add('collection - call native find with sort function',
     if (Meteor.isServer) {
       test.throws(
         function () {
-          nativeCollection.find({}, {sort: function(){}}).map(function (doc) { return doc.a; })
+          nativeCollection
+            .find({}, {
+              sort: function () {},
```
benjamn: Though I'm not going to request changes, in the future you should know you can write methods like this using method syntax:
```js
nativeCollection
  .find({}, {
    sort() {},
  })
```

dr-dimitru: Hi @benjamn ,
Thank you for merge.
I wish to use ES7 syntax, but most of file seems to be ES5. Am I wrong?

PR #8176, opened 2016-12-21
Make sure missed JS/CSS is not cacheble

dr-dimitru:
- Set `Cache-Control` to `no-cache, must-revalidate, proxy-revalidate`
 - Set `Expires` to date in a past


abernix:
Thanks for taking the time to try and improve Meteor however, in the future, please do not open PRs without conversation or issues attached to them.  You have not even explained why you're making this change.  Your contribution is appreciated, but the reasoning and implications need to be discussed first.

Also, your commit messages should more than just repeating the changes that you made in the code without further explanation.  Please see the [guidelines for submitting pull requests](https://github.com/meteor/meteor/blob/devel/Contributing.md#submitting-pull-requests).

At first glance though, I'd say this should be unnecessary as `no-cache` should be more definitive than `must-revalidate`, no?

Please open a new *issue* first demonstrating the problem you're trying to solve or a bug with the current implementation and we can go from there.  I'm happy to re-open this if we determine that is the best course of action.

dr-dimitru:
Hi @abernix ,

From guideline:
```
don't expect your PR to be merged unless it's a trivial and obvious fix (e.g documentation)
```

Please forgive me to assume this as `trivial and obvious fix`.

> At first glance though, I'd say this should be unnecessary as no-cache should be more definitive than must-revalidate, no?

`must-revalidate` will instruct to revalidate requests that was mistakenly cached in a past, due to some rare case.

Here is great statement from [SO thread](http://stackoverflow.com/questions/18148884/difference-between-no-cache-and-must-revalidate): 

> The thing is, with web services, you have to take a negative view and assume the worst for your unknown client apps. Any stale resource has the potential to cause a problem.

Although this comment `Which also means max-age=0, must-revalidate and no-cache are identical` is absolutely true, different browsers has different implementation, keeping in mind not every user is updated to latest browser version and users using all different kind of browsers.

I've made a tests to let you see the difference:

 - Current behaviour: `https://veliov.com/tests/current/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fcurrent%2Fstyle.css)
    - Note: `This response allows all caches to store it.`, but `This response cannot be served from cache without validation.`
    - In my case using CloudFlare this response id 100% cached
 - Suggested behaviour: `https://veliov.com/tests/suggested/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fsuggested%2Fstyle.css)
    - Which actually gives same results, but `Expires` header makes ached result stale right away
 - Thank you @abernix , after test I suggest [this fix](https://github.com/VeliovGroup/meteor/commit/50ee7919561d6b3299c626683e52c2bd75817a49): `https://veliov.com/tests/suggested-fix/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fsuggested-fix%2Fstyle.css)
    - Gives result: `This response can't be stored by a cache.`
    - At the same time gives fallback if any of `Cache-Control` value is not supported or misbehave

Other examples:
 - Without expire: `https://veliov.com/tests/no-expires/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fno-expires%2Fstyle.css)
 - Without expire and must-revalidate: `https://veliov.com/tests/no-expires-no-must-revalidate/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fno-expires-no-must-revalidate%2Fstyle.css)
 - Without must-revalidate: `https://veliov.com/tests/no-must-revalidate/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fno-must-revalidate%2Fstyle.css)
 - With expires only: `https://veliov.com/tests/expires-only/style.css` [RedBot](https://redbot.org/?uri=https%3A%2F%2Fveliov.com%2Ftests%2Fexpires-only%2Fstyle.css)

abernix:
"Trivial and obvious" would be: documentation-typo, a missing semi-colon, a version bump, etc., but even those should come with explanations as to *why* you're doing it, if you're trying to fix a bug, *what* you've tested, etc.  Anyway, thank you for adding the additional conversation! 🙂

I'll start by pointing out that those headers that you are changing are in the default handler and are only used for resources which are no longer available (for example, an old version of your app which is not able to be served).  Meteor already employs other cache-busting techniques which base the filename off the content of the file.

The tests you are showing still would yield the same end result: The resource is not valid to be served from the cache unless it can be revalidated by the server, which it cannot be.  Your suggestion to further add `no-store` only prevents storing of the resource at all – it wouldn't change the cache hit which is still prevented by lack of revalidation.  But again, this is for a resource that doesn't exist and won't be called again and even if it was fetched would return the same result.

I understand that implementations can differ and some older proxies are misconfigured.  For that reason, I do agree with taking a defensive approach (like the "negative view" comment above), but if anything, it seems like `no-store, must-revalidate` might be more appropriate based on [RFC 7234](https://tools.ietf.org/html/rfc7234) (which replaces [RFC 2616](https://tools.ietf.org/html/rfc2616)).  This is further explained in [this more popular SO post](https://stackoverflow.com/questions/49547/how-to-control-web-page-caching-across-all-browsers) about the topic.

Either way, it _still_ seems this change should be unnecessary.  Again: What problem are you trying to fix?  Have you actually seen a problem in certain browsers/clients/proxies?  Is this a problem you've encountered using the back button?

I'll reopen this for now.

dr-dimitru:
Hi @abernix ,

I've meet this problem while playing with caching on client via Service Workers (kind of back button problem). And using CloudFlare as caching proxy at the same time.

When event thrown by AppCache is fired, I'm forcing  Service Worker cache to revalidate used/requested files, with no-cache request. Somewhy with default Cache-Control for missing file old version is returned by CloudFlare.

abernix:
What you're saying doesn't match what CloudFlare says in [their documentation](https://support.cloudflare.com/hc/en-us/articles/202775670-How-Do-I-Tell-CloudFlare-What-to-Cache-)

> If the Cache-Control header is set to "private", "no-store", "no-cache",  or "max-age=0", or if there is a cookie in the response, then CloudFlare will not cache the resource.

When you say:
> In my case using CloudFlare this response is 100% cached

How are you testing this?  All of your RedBot links show the [CloudFlare header](https://support.cloudflare.com/hc/en-us/articles/200168266-What-do-the-various-CloudFlare-cache-responses-HIT-Expired-etc-mean-) indicating that it was not cached by them:

    CF-Cache-Status: MISS

Are you talking specifically about assets? (e.g. CSS, images, etc)?

You might want to [check what they think](https://support.cloudflare.com/hc/en-us/articles/200169556-How-can-I-tell-if-CloudFlare-is-caching-my-site-or-a-specific-file-).  

Have you tried running Meteor from a check out with this change to confirm this fixes it?  A reproduction might help prove a problem here but based on what you've explained so far I don't think this would fix your problem.

dr-dimitru:
@abernix 

The logic behind this:
 1. File exists, and cached on CloudFlare
 2. We update the app, file is not exists anymore
 3. At this point on file request it's not revalidated, and old file is returned form CF

```
CF-Cache-Status: MISS
```

Requests is not cached as all of them initially has `Cache-Control` header which prevents initial caching. Probably it's a good idea to simulate case when file exists before, and now not.

abernix:
@dr-dimitru `CF-Cache-Status: MISS` indicates that it is NOT being cached by CloudFlare so your "number 3" (in the previous comment) should not be true based on that.  Also, if you are saying that you update your app (in "number 2") a new cache-busting filename would be generated and Meteor should _not_ be looking for the old file anymore, right?

Additionally, if "number 3" _was_ happening, wouldn't the proper change for this be somewhere else?  You're trying to change the logic on the code which sets cache headers for files that _don't_ exist and shouldn't be getting requested, and if they DID previous exist, they would have different cache headers than the ones that are set here, right?

This whole behavior you're explaining doesn't match CloudFlare's documentation or the expected result of those headers and I'm just not sure this would fix your problem.  Have you tested this change in practice?  Or is this just a theoretical fix?  Doesn't it seem more likely that there is a problem with your service worker implementation?

dr-dimitru:
> Doesn't it seem more likely that there is a problem with your service worker implementation?

Yes, as I've mentioned before - it's my badly written SW.
As I was playing around and testing it.

> Also, if you are saying that you update your app (in "number 2") a new cache-busting filename would be generated and Meteor should not be looking for the old file anymore, right?

Right, but SW wouldn't use a new files (*it would't even know what currently used files is stale*) unless CF and other caches will revalidate it.

I might be wrong here, so I've made more tests:

First response without `Cache-Control` and `Expires` header (visit few times to make CF cache it):
```
    HTTP/1.1 200 OK
    Date: Fri, 20 Jan 2017 10:17:16 GMT
    Content-Type: text/css; charset=utf-8
    Content-Length: 42
    Connection: keep-alive
    Set-Cookie: __cfduid=d7f835532ccdc02117e4e5d7c3e3650bc1484907436;
        expires=Sat, 20-Jan-18 10:17:16 GMT; path=/;
        domain=.veliov.com; HttpOnly
    CF-Cache-Status: HIT
    Expires: Mon, 13 Feb 2017 10:17:16 GMT
    Cache-Control: public, max-age=2073600
    Server: cloudflare-nginx
    CF-RAY: 3241c6974c10655d-SYD
```

Then add `Cache-Control 'no-cache';` (current behaviour, which returns cached version, even with `pragma:no-cache` request header):
```
    HTTP/1.1 200 OK
    Date: Fri, 20 Jan 2017 10:21:00 GMT
    Content-Type: text/css; charset=utf-8
    Content-Length: 42
    Connection: keep-alive
    Set-Cookie: __cfduid=d65838b72eb7cfa2aa7fdd6d0038d08991484907660;
        expires=Sat, 20-Jan-18 10:21:00 GMT; path=/;
        domain=.veliov.com; HttpOnly
    CF-Cache-Status: HIT
    Expires: Mon, 13 Feb 2017 10:21:00 GMT
    Cache-Control: public, max-age=2073600
    Server: cloudflare-nginx
    CF-RAY: 3241cc0d0c8a65d5-SYD
```

Adding suggested `Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate'` does not forces CF to revalidate the cache. Response is same:
```
    HTTP/1.1 200 OK
    Date: Fri, 20 Jan 2017 10:25:05 GMT
    Content-Type: text/css; charset=utf-8
    Content-Length: 42
    Connection: keep-alive
    Set-Cookie: __cfduid=dc639d7bd858deeb32865bec2bcac17e11484909645;
        expires=Sat, 20-Jan-18 10:25:05 GMT; path=/;
        domain=.veliov.com; HttpOnly
    Cache-Control: public, max-age=2073600
    CF-Cache-Status: HIT
    Expires: Mon, 13 Feb 2017 10:25:05 GMT
    Server: cloudflare-nginx
    CF-RAY: 3241fc8458e26521-SYD
```

So, I decided to try to add `proxy-revalidate` to initial request, and visit few times to make CF cache it.  `proxy-revalidate` is striped from response:
```
    HTTP/1.1 200 OK
    Date: Fri, 20 Jan 2017 10:54:05 GMT
    Content-Type: text/css; charset=utf-8
    Content-Length: 42
    Connection: keep-alive
    Set-Cookie: __cfduid=dc639d7bd858deeb32865bec2bcac17e11484909645;
        expires=Sat, 20-Jan-18 10:54:05 GMT; path=/;
        domain=.veliov.com; HttpOnly
    Cache-Control: public, max-age=2073600
    CF-Cache-Status: HIT
    Expires: Mon, 13 Feb 2017 10:54:05 GMT
    Server: cloudflare-nginx
    CF-RAY: 3241fc8458e26521-SYD
```

I've also played with combinations of:
 - `Cache-Control 'public, proxy-revalidate, must-revalidate'`
 - `Last-Modified $date_gmt`

Before, and after file was cached, but all of them is __failed__, CF stores file an won't revalidate it unless it's become stale in CF cache.

So, I guess if you have no other idea how to force CF to revalidate and update file in its cache after it's changed at origin - this PR may be closed.

harrisonhunter:
Minor note here, running the OWASP ZAP vulnerability scanner suggests changing the cache-control header from 'no-cache' to 'no-cache, no-store, must-revalidate, private' https://www.owasp.org/index.php/Session_Management_Cheat_Sheet#Web_Content_Caching

Not sure if that's important but just noticed that a while ago when checking some security things

abernix:
I'm going to close this because I'm confident that it's not going to fix your problem.

Just to be clear though, I'm not opposed to changing these `Cache-Control` headers to something more resilient, for example, to add `Pragma`, `Expires: 0` (not the date variation currently in the PR, but `0` which is more appropriate) and `Cache-Control` headers such as `no-store`, `must-revalidate`, however these changes would be purely for faulty proxy implementations and shouldn't be necessary.  And specifically, don't seem necessary in your case.  I'm fairly sure that folks are successfully using Meteor with CloudFlare, and I'm also confident that CloudFlare does revalidate resources when instructed to and operates within spec.

@harrisonhunter There is nothing sensitive in this content – it's literally an empty page; there is no vulnerability here.  Again though, I'm not sure anything you'd be receiving from OWASP would be referring to this code unless you're scanning pages that _used_ to exist – which seems unlikely.

abernix:
Oh, additionally, in regards to:

> I decided to try to add proxy-revalidate to initial request

You shouldn't be expecting a `Cache-Control` header sent in a HTTP request to appear in the response.  While there is a `Cache-Control` header on the [request](https://tools.ietf.org/html/rfc7234#section-5.2.1) and the response, there are different supported values for each mode.  It wasn't clear to me if this is what you meant by "initial request".

dr-dimitru:
`initial request` I've meant first response (*typo*)

PR #8179, opened 2016-12-21
Update node-mongodb driver to 2.2.16

OleksandrChekhovskyi:
This fixes an issue where the driver fails to recover from unresponsive primary in a replica set.

Relevant commit in `mongodb-core`:
https://github.com/christkv/mongodb-core/commit/6611ea48e2a9e195fdef1572d3f94db5924b3ee9

### DIFF
--- a/packages/npm-mongo/.npm/package/npm-shrinkwrap.json
+++ b/packages/npm-mongo/.npm/package/npm-shrinkwrap.json
@@ -1,9 +1,9 @@
 {
   "dependencies": {
     "bson": {
-      "version": "0.5.6",
-      "resolved": "https://registry.npmjs.org/bson/-/bson-0.5.6.tgz",
-      "from": "bson@>=0.5.6 <0.6.0"
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/bson/-/bson-1.0.1.tgz",
+      "from": "bson@>=1.0.1 <1.1.0"
     },
     "buffer-shims": {
       "version": "1.0.0",
@@ -31,14 +31,14 @@
       "from": "isarray@>=1.0.0 <1.1.0"
     },
     "mongodb": {
-      "version": "2.2.11",
-      "resolved": "https://registry.npmjs.org/mongodb/-/mongodb-2.2.11.tgz",
-      "from": "mongodb@2.2.11"
+      "version": "2.2.16",
+      "resolved": "https://registry.npmjs.org/mongodb/-/mongodb-2.2.16.tgz",
+      "from": "mongodb@2.2.16"
     },
     "mongodb-core": {
-      "version": "2.0.13",
-      "resolved": "https://registry.npmjs.org/mongodb-core/-/mongodb-core-2.0.13.tgz",
-      "from": "mongodb-core@2.0.13"
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/mongodb-core/-/mongodb-core-2.1.2.tgz",
+      "from": "mongodb-core@2.1.2"
     },
     "process-nextick-args": {
       "version": "1.0.7",
--- a/packages/npm-mongo/package.js
+++ b/packages/npm-mongo/package.js
@@ -3,12 +3,12 @@
 
 Package.describe({
   summary: "Wrapper around the mongo npm package",
-  version: '2.2.11_2',
+  version: '2.2.16_1',
   documentation: null
 });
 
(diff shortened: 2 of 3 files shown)


abernix:
Since you have CircleCI setup on your own `meteor` fork but disabled for this branch, the CircleCI tests didn't run on this PR (at all) but I recently tried upgrading `mongodb` to 2.2.15 in #8150 and they failed.  I also tried 2.2.16 with the same results.  See the test results here: [2.2.15](https://circleci.com/gh/abernix/meteor/tree/feature%2Fupdate-npm-mongodb-2.2.15) and [2.2.16](https://circleci.com/gh/abernix/meteor/tree/feature%2Fupdate-npm-mongodb-2.2.16).

I agree that the bug fix is important, but those tests need to pass before this can be merged.  If you have an opportunity to look into it, please try [running from a checkout](https://github.com/meteor/meteor/#slow-start-for-developers) and [running the tests locally](https://github.com/meteor/meteor/blob/devel/Contributing.md#running-tests-on-meteor-core) and getting the failing tests to pass.  Also, please enable your CircleCI builds for this PR to see the progress.

These tests are all failing consistently from what I've seen:

  - run: run --once with real Mongo
  - run: run and SIGKILL parent process
  - source-maps: source maps from an app
  - shell-tests: meteor shell
  - standard-minification: standard-minifiers - CSS splitting
  - stylus-cross-packages: can import stylus across packages
  - custom-minifier: custom minifier - devel vs prod
  - hot-code-push: css hot code push
  - hot-code-push: versioning hot code push
  - modules: modules - unimported lazy files
  - old: bundler-assets
  - test-modes: 'meteor test --port' accepts/rejects proper values

This might have to do with the major upgrade of `bson`, though I haven't investigated further.

OleksandrChekhovskyi:
I have been investigating the failing tests. The exact reason is still unclear, but this is what I have found so far:
* Disabling warehouse in the sandbox seems to fix the `source-maps` test (possibly this means that some wrong npm dependency is loaded?).
* The test fails because connection to the newly-spawned `mongod` does not finish the initial `open` command. `mongod` sends the reply, but the node.js mongo client does not seem to notice (no messages in the log, when logging is set to "debug" level), and then the connection times out.

OleksandrChekhovskyi:
Updated the patch. The problem was that mongod took significant time to respond to the initial `ismaster` ping (when starting from empty state), and so connection timed out. The fix is to increase the connect timeout used for initial mongo connection. Now all tests seem to pass.

abernix:
Thanks for your continued work on this!

While that does appear to fix the problem, I don't think we should be very comfortable with simply saying "It used to take less than 2 seconds, but now it can take up to 30 seconds".

I think we should get a better explanation as to _why_ this is happening.  Perhaps something in Mongo has changed that needs to be dealt with differently?  Maybe one of the [open issues](https://github.com/christkv/mongodb-core/issues) is related?

Have you tried observing the Mongo logs by printing the logs [from this area](https://github.com/OleksandrChekhovskyi/meteor/blob/14a2bf913ae8539ad731401d68a54774242f9542/tools/runners/run-mongo.js#L533) and seeing what is different between the two versions?

OleksandrChekhovskyi:
The actual timings didn't change, as far as I can see. The problem happens because start of mongo and app build is done in parallel, and build blocks event loop, so even though mongo replied quickly, the driver will get a chance to process the message much later.

The default timeout of 2s for monitoring messages has been added there recently: https://github.com/christkv/mongodb-core/commit/a483635855362f5263506f94b7973e26411ac746

So it didn't get slower, it's just that blocked event loop now caused problems because of small default monitoring timeout, and it didn't before. It's a weird corner case.

I think simply increasing the timeout for that connection (used only to init replica set) is a reasonable fix.

abernix:
Thanks for the additional information/research!  Your diagnosis sounds accurate. 

I suppose a problem could still arise here on a slow computer, but this might be reasonable to move forward for now.

Do you mind formatting that settings `Object` so it's not all on one line?

OleksandrChekhovskyi:
Updated the patch with multi-line formatting.

OleksandrChekhovskyi:
Reminder that the patch is ready to be merged. Is there anything preventing it?

benjamn:
Still need to re-publish `npm-mongo` for these changes to become available, but thankfully we can do that without a whole new Meteor release (since Meteor 1.4).

Thanks for working on this @OleksandrChekhovskyi!

Issue #5964, opened 2016-01-10
Meteor.call with DOM element leads to too much recursion


CaptainN:
I accidentally passed in a DOM reference to a Meteor method call using react like this:

``` JavaScript
// hangs/crashes/too much recursion
Meteor.call("updateOwnEmail", this.refs.email, (error) => { ... });
// works fine
Meteor.call("updateOwnEmail", this.refs.email.value, (error) => { ... });
```

It crashed the page in Chrome after a brief hang, and Firefox kicks out a "too much recursion" message. The rolled out error message shows a lot of EJSON lines (and ES5), so I would guess the serializer is choking on the DOM element, but that's just a guess.

A better outcome would be a console message saying that DOM elements can't be passed to the server.

stubailo:
Yes I agree, it would be great for either EJSON or `Meteor.call` to throw a helpful error when values that can't be serialized are passed in. I'd be happy to look at a pull request to make this improvement.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

This feature request will be closed here, but anyone interested in migrating this feature request to the new repository (to make sure it stays active), can click [here](https://github.com/meteor/meteor-feature-requests/issues/new?title=Meteor.call%20with%20DOM%20element%20leads%20to%20too%20much%20recursion&body=Migrated%20from:%20meteor/meteor%235964) to start the feature request migration process. This manual migration process is intended to help identify which of the older feature requests are still considered to be of value to the community. Thanks!

PR #8201, opened 2016-12-29
Throw error when EJSON.clone circular structure

nlhuykhang:
Check for circular structure of objects passed to `EJSON.clone`, throw error if exists.
In reference to #5964

### DIFF
--- a/packages/ejson/ejson.js
+++ b/packages/ejson/ejson.js
@@ -454,6 +454,27 @@ EJSON.equals = function (a, b, options) {
   }
 };
 
+function isCyclic (obj) {
+  var seenObjects = [];
+
+  function detect (obj) {
+    if (obj && typeof obj === 'object') {
+      if (seenObjects.indexOf(obj) !== -1) {
+        return true;
+      }
+      seenObjects.push(obj);
+      for (var key in obj) {
+        if (obj.hasOwnProperty(key) && detect(obj[key])) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  return detect(obj);
+}
+
 /**
  * @summary Return a deep copy of `val`.
  * @locus Anywhere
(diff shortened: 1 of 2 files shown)


mitar:
Have you done any performance benchmarks with this change? Because it can potentially slow down cloning a lot. And Meteor is doing a lot of cloning everywhere. You are doing one extra iteration over all properties of all objects, and you are also storing all objects you see into an array which could grow a lot, not to mention that scanning the array is O(n) operation, and you scan again and again, which makes this cycle checking operation quite unoptimized.

One option would be that instead of doing one extra pass to detect cycles, you would be remembering what you already cloned and if you get back to what you cloned during cloning process, you throw an error.

Alternatively, you could extend EJSON with additional method which check for cycles. So users can run that themselves before cloning if they suspect an object they want to clone has cycles. Because sometimes you know that there are not cycles, so why checking then?

nlhuykhang:
I did not really think about performance when working on this. I just focused on solving #5964. But you are right, this change could make a negative impact on all Meteor projects.

I will revert `.clone` to the old implementation, and add a new method to EJSON named `.isCyclic` to check for cycles. Use this new method when executing Meteor methods to resolve #5964, sound good?

mitar:
So maybe you could add this method and then also add those checks when method arguments are getting serialized.

abernix:
Please rebase this on `devel` now that #8204 has landed, which should fix the failing tests.

Also, it appears your CircleCI builds aren't enabled for (your) `nlhuykhang/meteor` repo so [none of those](https://circleci.com/gh/nlhuykhang/meteor) ran for this PR.  You can enable up to 4 containers for free on Meteor and it will use them all.

nlhuykhang:
I closed this PR for now as it is not yet finished

nlhuykhang:
After rebasing `devel`, I created a new PR here #8227. I do not know why the build keeps failing even though everything works fine and all tests are passed on my local machine.

Could anyone help me?

Issue #8202, opened 2016-12-29
babel-runtime dependency missing in bare skeleton


mhidou:
When creating an empty app we got this error.

``` 
$ meteor create --bare .

Created a new Meteor app in current directory.

To run your new app:
  meteor npm install
  meteor

If you are new to Meteor, try some of the learning
resources here:
  https://www.meteor.com/learn

$ meteor npm install
$ meteor run

=> Started proxy.
=> Started MongoDB.

# ...

Error: The babel-runtime npm package could not be found in your node_modules directory. Please run the following command to install it:

 meteor npm install --save babel-runtime

# ...

=> Exited with code: 1
=> Your application is crashing. Waiting for file
   change.
```

Adding babel-runtime as a npm dependency in the bare skeleton fixes this issue.

abernix:
Thanks for reporting this!  I opened PR #8203 to fix this.

PR #8203, opened 2016-12-29
Ensure `babel-runtime` is included in "skeleton" variations

abernix:
The `bare` skeleton didn't include the `babel-runtime` in `package.json`, thus making the instructions for starting the project (`meteor npm install` & `meteor`) inaccurate.

This change makes the `bare` skeleton the same as the `full` and "default" versions.

As an unnecessary liberty, since the latest version would be installed anyway, I bumped `meteor-node-stubs` and `babel-runtime` to the latest published version numbers and standardized the format of all three `package.json` files.

Closes meteor/meteor#8202

### DIFF
--- a/tools/static-assets/skel-bare/package.json
+++ b/tools/static-assets/skel-bare/package.json
@@ -5,6 +5,7 @@
     "start": "meteor run"
   },
   "dependencies": {
-    "meteor-node-stubs": "~0.2.0"
+    "babel-runtime": "^6.20.0",
+    "meteor-node-stubs": "~0.2.4"
   }
 }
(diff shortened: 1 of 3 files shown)


klaussner:
Just nitpicking but I would change the order of the two packages because npm sorts them alphabetically when adding other packages with `npm install --save`. Generating them in alphabetical order makes diffs look a little better.

abernix:
@klaussner Perfectly reasonable nit-picking!  Fixed, however Travis CI tests are now failing because the `blaze` repo was rebased (see #8204 & related).  I'll rebase once that is merged.

PR #8205, opened 2016-12-30
Fastclick will break selects on Chrome mobile

elie222:
See here:
https://github.com/ftlabs/fastclick/issues/497
And here:
https://productforums.google.com/forum/#!topic/chrome/Q4Rt6d0C4Qo

I just spent hours trying to figure out why selects haven't been working on my site, and finally found out it was fastclick.

### DIFF
--- a/packages/fastclick/README.md
+++ b/packages/fastclick/README.md
@@ -2,9 +2,7 @@
 [Source code of released version](https://github.com/meteor/meteor/tree/master/packages/fastclick) | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/fastclick)
 ***
 
-```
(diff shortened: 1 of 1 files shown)


packages/fastclick/README.md
```diff
@@ -2,6 +2,10 @@
 [Source code of released version](https://github.com/meteor/meteor/tree/master/packages/fastclick) | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/fastclick)
 ***
 
+```
```
abernix: This shouldn't be a code-fence (` ``` `) (as you can see in the [preview](https://github.com/elie222/meteor/blob/f0a2c8cfe4a252cb928b27bc607767d3952eb00e/packages/fastclick/README.md) it breaks the display.  How about removing both fencing lines and just add `>` in front of the paragraph?

Either way, I think a bold `Warning:` in front of it would be even more helpful:

```
> **Warning:** As of late...
```

abernix:
Don't worry about the failing TravisCI tests.  All new PRs will now fail until #8204 lands.

benjamn:
Updated the metadata of the `fastclick` package on Atmosphere without publishing a new version: https://atmospherejs.com/meteor/fastclick

Issue #5490, opened 2015-10-21
improvement: don't pass options.email to sendVerificationEmail method when creating a new user


SierraGolf:
**problem**
currently the `createUser` implementation for the accounts-password package passes the options.email to the `sendVerification` method, see [here](https://github.com/meteor/meteor/blob/dc3cd6eb92f2bdd1bb44000cdd6abd1e5d0285b1/packages/accounts-password/password_server.js#L1021).

generally this is not a problem. if however the application uses a framework like [collection2](https://github.com/aldeed/meteor-collection2) which performs transformation of the data which is persisted this might cause a problem.

collection2 automatically trims the fields if not specified otherwise. if a user accidentally enters his email address with a space at the end the data will be trimmed by collection2 and persisted correctly but the verification email can not be send because the email from the options is now different from the one persisted, see [here](https://github.com/meteor/meteor/blob/dc3cd6eb92f2bdd1bb44000cdd6abd1e5d0285b1/packages/accounts-password/password_server.js#L739).

**potential solution**
when creating a user just pass the user id to the `sendVerificationEmail` method.

**workaround**
I have created a schema for the options as well which I clean and validate in the onCreateUser hook.

stubailo:
Seems legit, submit a PR?

SierraGolf:
I would be happy to. Is there any documentation on meteors internal testing facilities?
How would I add a deny hook on insert just for one single test in order to not have side-effects on other tests?

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

This feature request will be closed here, but anyone interested in migrating this feature request to the new repository (to make sure it stays active), can click [here](https://github.com/meteor/meteor-feature-requests/issues/new?title=improvement%3A%20don't%20pass%20options.email%20to%20sendVerificationEmail%20method%20when%20creating%20a%20new%20user&body=Migrated%20from:%20meteor/meteor%235490) to start the feature request migration process. This manual migration process is intended to help identify which of the older feature requests are still considered to be of value to the community. Thanks!

PR #8206, opened 2017-01-01
Do not pass options.email to .sendVerificationEmail when creating new user

nlhuykhang:
In reference to #5490

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -1047,7 +1047,7 @@ Meteor.methods({createUser: function (options) {
       // a token to verify the user's primary email, and send it to
       // that address.
       if (options.email && Accounts._options.sendVerificationEmail)
-        Accounts.sendVerificationEmail(userId, options.email);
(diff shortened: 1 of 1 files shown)


abernix:
Sorry this has taken so long to look at, but thank you for submitting this.  I realize the problem [originally reported](https://github.com/meteor/meteor/issues/5490#issue-112631828) might still exist (although there haven't been any other reports in 1.5 years), but this change can't be accepted as currently submitted since in the situation where multiple e-mail addresses are waiting to get "verified", this could allow the wrong e-mail address to get marked as "verified" – something we most certainly do not want to happen.

An alternative approach might include a more aggressive approach which makes sure all e-mail addresses are "trimmed" within the accounts package at all stages, whether during `.find` or `.insert`, or possibly a solution in `Collection2`.  Again though, given the age of the issue and the lack of +1s, it might not actually be an issue.  If you're encountering this problem, I'd encourage taking a different approach to solving this.

Thanks again for trying to make this contribution though! (and thank you for your others!)

PR #8207, opened 2017-01-01
Refactored reset password and enroll account methods

mitar:
To allow easier customization email sending of tokens.

More or less just moving code around.

A new version of #8065 and #8038.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -547,16 +547,16 @@ Meteor.methods({forgotPassword: function (options) {
 }});
 
 /**
- * @summary Generates a reset token record.
+ * @summary Generates a reset token and saves it into the database.
  * @locus Server
- * @param {String} userId The id of the user to generate the reset token record for.
- * @param {String} email Which address of the user to generate the reset token record for. This address must be in the user's `emails` list. If `null`, defaults to the first email in the list.
+ * @param {String} userId The id of the user to generate the reset token for.
+ * @param {String} email Which address of the user to generate the reset token for. This address must be in the user's `emails` list. If `null`, defaults to the first email in the list.
  * @param {String} reason `resetPassword` or `enrollAccount`.
  * @param {Object} [extraTokenData] Optional additional data to be added into the token record.
- * @returns {Object} Object with {email, user, tokenRecord} values.
+ * @returns {Object} Object with {email, user, token} values.
  * @importFromPackage accounts-base
  */
-Accounts.generateResetTokenRecord = function (userId, email, reason, extraTokenData) {
+Accounts.generateResetToken = function (userId, email, reason, extraTokenData) {
   // Make sure the user exists, and email is one of their addresses.
   var user = Meteor.users.findOne(userId);
   if (!user) {
@@ -573,8 +573,9 @@ Accounts.generateResetTokenRecord = function (userId, email, reason, extraTokenD
     handleError("No such email for user.");
   }
 
+  var token = Random.secret();
   var tokenRecord = {
-    token: Random.secret(),
+    token: token,
     email: email,
     when: new Date()
   };
@@ -592,19 +593,26 @@ Accounts.generateResetTokenRecord = function (userId, email, reason, extraTokenD
     _.extend(tokenRecord, extraTokenData);
   }
 
-  return {email, user, tokenRecord};
+  Meteor.users.update({_id: user._id}, {$set: {
+    'services.password.reset': tokenRecord
+  }});
+
+  // before passing to template, update user object with new token
+  Meteor._ensure(user, 'services', 'password').reset = tokenRecord;
+
+  return {email, user, token};
 };
 
 /**
- * @summary Generates an e-mail verification token record.
+ * @summary Generates an e-mail verification token and saves it into the database.
  * @locus Server
- * @param {String} userId The id of the user to generate the  e-mail verification token record for.
- * @param {String} email Which address of the user to generate the e-mail verification token record for. This address must be in the user's `emails` list. If `null`, defaults to the first unverified email in the list.
+ * @param {String} userId The id of the user to generate the  e-mail verification token for.
+ * @param {String} email Which address of the user to generate the e-mail verification token for. This address must be in the user's `emails` list. If `null`, defaults to the first unverified email in the list.
  * @param {Object} [extraTokenData] Optional additional data to be added into the token record.
- * @returns {Object} Object with {email, user, tokenRecord} values.
+ * @returns {Object} Object with {email, user, token} values.
  * @importFromPackage accounts-base
  */
-Accounts.generateVerificationTokenRecord = function (userId, email, extraTokenData) {
+Accounts.generateVerificationToken = function (userId, email, extraTokenData) {
   // Make sure the user exists, and email is one of their addresses.
   var user = Meteor.users.findOne(userId);
   if (!user) {
@@ -626,8 +634,9 @@ Accounts.generateVerificationTokenRecord = function (userId, email, extraTokenDa
     handleError("No such email for user.");
   }
 
+  var token = Random.secret();
   var tokenRecord = {
-    token: Random.secret(),
+    token: token,
     // TODO: This should probably be renamed to "email" to match reset token record.
     address: email,
     when: new Date()
@@ -636,41 +645,6 @@ Accounts.generateVerificationTokenRecord = function (userId, email, extraTokenDa
   if (extraTokenData) {
     _.extend(tokenRecord, extraTokenData);
   }
-
-  return {email, user, tokenRecord};
-};
-
-/**
- * @summary Saves the reset token record into database in user's document.
- * @param {Object} user User for which to save the reset token record for.
- * @param {Object} tokenRecord The reset token record to be saved.
- * @returns {Object} User object with added token record.
- * @importFromPackage accounts-base
- */
-Accounts.saveResetTokenRecord = function (user, tokenRecord) {
-  // clone so that we do not change the input
-  user = EJSON.clone(user);
-
-  Meteor.users.update({_id: user._id}, {$set: {
-    'services.password.reset': tokenRecord
-  }});
-
-  // before passing to template, update user object with new token
-  Meteor._ensure(user, 'services', 'password').reset = tokenRecord;
-
-  return user;
-};
-
-/**
- * @summary Saves the e-mail verification token record into database in user's document.
- * @param {Object} user User for which to save the e-mail verification token record for.
- * @param {Object} tokenRecord The e-mail verification token record to be saved.
- * @returns {Object} User object with added token record.
- * @importFromPackage accounts-base
- */
-Accounts.saveVerificationTokenRecord = function (user, tokenRecord) {
-  // clone so that we do not change the input
-  user = EJSON.clone(user);
 
   Meteor.users.update({_id: user._id}, {$push: {
     'services.email.verificationTokens': tokenRecord
@@ -683,7 +657,7 @@ Accounts.saveVerificationTokenRecord = function (user, tokenRecord) {
   }
   user.services.email.verificationTokens.push(tokenRecord);
 
-  return user;
+  return {email, user, token};
 };
 
 /**
@@ -730,17 +704,16 @@ Accounts.generateOptionsForEmail = function (email, user, url, reason) {
  * @param {String} userId The id of the user to send email to.
  * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
  * @param {Object} [extraTokenData] Optional additional data to be added into the token record.
- * @returns {Object} Object with {email, user, tokenRecord, url, options} values.
+ * @returns {Object} Object with {email, user, token, url, options} values.
  * @importFromPackage accounts-base
  */
 Accounts.sendResetPasswordEmail = function (userId, email, extraTokenData) {
-  const {email: realEmail, user, tokenRecord} =
-    Accounts.generateResetTokenRecord(userId, email, 'resetPassword', extraTokenData);
-  const userWithTokenRecord = Accounts.saveResetTokenRecord(user, tokenRecord);
-  const url = Accounts.urls.resetPassword(tokenRecord.token);
-  const options = Accounts.generateOptionsForEmail(realEmail, userWithTokenRecord, url, 'resetPassword');
+  const {email: realEmail, user, token} =
+    Accounts.generateResetToken(userId, email, 'resetPassword', extraTokenData);
+  const url = Accounts.urls.resetPassword(token);
+  const options = Accounts.generateOptionsForEmail(realEmail, user, url, 'resetPassword');
   Email.send(options);
-  return {email: realEmail, user: userWithTokenRecord, tokenRecord, url, options};
+  return {email: realEmail, user, token, url, options};
 };
 
 // send the user an email informing them that their account was created, with
@@ -757,17 +730,16 @@ Accounts.sendResetPasswordEmail = function (userId, email, extraTokenData) {
  * @param {String} userId The id of the user to send email to.
  * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
  * @param {Object} [extraTokenData] Optional additional data to be added into the token record.
- * @returns {Object} Object with {email, user, tokenRecord, url, options} values.
+ * @returns {Object} Object with {email, user, token, url, options} values.
  * @importFromPackage accounts-base
  */
 Accounts.sendEnrollmentEmail = function (userId, email, extraTokenData) {
-  const {email: realEmail, user, tokenRecord} =
-    Accounts.generateResetTokenRecord(userId, email, 'enrollAccount', extraTokenData);
-  const userWithTokenRecord = Accounts.saveResetTokenRecord(user, tokenRecord);
-  const url = Accounts.urls.enrollAccount(tokenRecord.token);
-  const options = Accounts.generateOptionsForEmail(realEmail, userWithTokenRecord, url, 'enrollAccount');
+  const {email: realEmail, user, token} =
+    Accounts.generateResetToken(userId, email, 'enrollAccount', extraTokenData);
+  const url = Accounts.urls.enrollAccount(token);
+  const options = Accounts.generateOptionsForEmail(realEmail, user, url, 'enrollAccount');
   Email.send(options);
-  return {email: realEmail, user: userWithTokenRecord, tokenRecord, url, options};
+  return {email: realEmail, user, token, url, options};
 };
 
 
@@ -865,7 +837,7 @@ Meteor.methods({resetPassword: function (token, newPassword) {
  * @param {String} userId The id of the user to send email to.
  * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first unverified email in the list.
  * @param {Object} [extraTokenData] Optional additional data to be added into the token record.
- * @returns {Object} Object with {email, user, tokenRecord, url, options} values.
+ * @returns {Object} Object with {email, user, token, url, options} values.
  * @importFromPackage accounts-base
  */
 Accounts.sendVerificationEmail = function (userId, email, extraTokenData) {
@@ -873,13 +845,12 @@ Accounts.sendVerificationEmail = function (userId, email, extraTokenData) {
   // account if they own said address but weren't those who created
   // this account.
 
-  const {email: realEmail, user, tokenRecord} =
-    Accounts.generateVerificationTokenRecord(userId, email, extraTokenData);
-  const userWithTokenRecord = Accounts.saveVerificationTokenRecord(user, tokenRecord);
-  const url = Accounts.urls.verifyEmail(tokenRecord.token);
-  const options = Accounts.generateOptionsForEmail(realEmail, userWithTokenRecord, url, 'verifyEmail');
+  const {email: realEmail, user, token} =
+    Accounts.generateVerificationToken(userId, email, extraTokenData);
+  const url = Accounts.urls.verifyEmail(token);
+  const options = Accounts.generateOptionsForEmail(realEmail, user, url, 'verifyEmail');
   Email.send(options);
-  return {email: realEmail, user: userWithTokenRecord, tokenRecord, url, options};
+  return {email: realEmail, user, token, url, options};
 };
 
 // Take token from sendVerificationEmail, mark the email as verified,
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -830,8 +830,8 @@ Accounts.sendVerificationEmail = function (userId, email) {
 
   const {email: realEmail, user, tokenRecord} = Accounts.generateVerificationTokenRecord(userId, email);
   Accounts.saveVerificationTokenRecord(user, tokenRecord);
-  var url = Accounts.urls.verifyEmail(tokenRecord.token);
-  var options = Accounts.generateOptionsForEmail(realEmail, user, url, 'verifyEmail');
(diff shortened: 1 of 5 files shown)


mitar:
While this allows one to customize how e-mails are being send, the issue s that it does not allow customizing existing methods. For example, `forgotPassword` calls `Accounts.sendResetPasswordEmail` so there is no way to change it to call something else, with custom flow. Now, I could implement my own method, and use that, but then this original method would still stay around. There is no way to unregister it. :-(

mitar:
Thanks. No automatic code linting/style checking? :-(

tmeasday:
Ha! Have you seen the Meteor code base?

abernix:
@mitar, I don't know whether your sad face is sarcastic or not, but I'd assume you're aware that the "automatic code linting/style checking" hasn't changed in a very long time – I really don't think that should come as a surprise to you.  Some stuff is linted automatically, but [a lot is ignored](https://github.com/meteor/meteor/blob/devel/.eslintignore).  There has been the [Meteor Style Guide](https://github.com/meteor/javascript) for quite some time and it is very closely tied to that of Airbnb, which is what most people refer to in "linter" conversations these days.

My opinion is that the reasonable thing to do is make sure code that is added or changed matches that guide and eventually it will work itself out.  I'm personally not in favor of changing all Meteor code (much of which was written before linters were commonplace) as it munges up VCS history and I see little benefit.

Anyhow, as to the actual functionality, this looks pretty good to me, but shouldn't [`Accounts.sendVerificationEmail`](https://github.com/mitar/meteor/blob/443188c25938e8f21789ab20361266c6146fb6e0/packages/accounts-password/password_server.js#L737-L807) get the same treatment?

Yes, you would need to implement your own `forgotPassword` method as you described, but I suppose you could probably block the default using a DDPRateLimiter rule with a maximum of zero requests.  Not sure how far we want to work this as it doesn't seem to have been a highly demanded feature?  Or maybe everyone has just built around it some other way?

packages/accounts-password/password_server.js
```diff
@@ -516,17 +516,17 @@ Meteor.methods({forgotPassword: function (options) {
...
  * @locus Server
  * @param {String} userId The id of the user to send email to.
- * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
+ * @param {String} email Which address of the user's to send the email to. This address must be in the user's `emails` list. If `null`, defaults to the first email in the list.
+ * @param {String} reason String `resetPassword` or `enrollAccount`.
+ * @returns {Object} Object with {email, user, token, url} values.
  * @importFromPackage accounts-base
  */
-Accounts.sendResetPasswordEmail = function (userId, email) {
+Accounts.createPasswordToken = function (userId, email, reason) {
```
abernix: I'm not sure `createPasswordToken` is appropriate here since it's used for both `enroll`, `password` (and hopefully, `verify`?).

packages/accounts-password/password_server.js
```diff
@@ -516,17 +516,17 @@ Meteor.methods({forgotPassword: function (options) {
...
-
 /**
- * @summary Send an email with a link the user can use to reset their password.
+ * @summary Creates options for email sending for reset password and enroll account emails.
+ * You can use this function when customizing a reset password or enroll account email sending.
  * @locus Server
  * @param {String} userId The id of the user to send email to.
- * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
+ * @param {String} email Which address of the user's to send the email to. This address must be in the user's `emails` list. If `null`, defaults to the first email in the list.
+ * @param {String} reason String `resetPassword` or `enrollAccount`.
```
abernix: `String` seems redundant here (and below)?

abernix: I think it's too late to fix this since we didn't oppose to it in #7817, but it would be nice if we didn't have the duality between "resetPassword" and "reset" (and "enrollAccount" and "enroll").

Oh well, I guess.

packages/accounts-password/password_server.js
```diff
@@ -538,44 +538,89 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
   if (!email || !_.contains(_.pluck(user.emails || [], 'address'), email))
     throw new Error("No such email for user.");
 
+  var tokenReason;
```
abernix: Why use `var` here but `let` in other new code? 🙂

packages/accounts-password/password_server.js
```diff
@@ -538,44 +538,89 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+ * You can use this function when customizing a reset password or enroll account email sending.
+ * @locus Server
+ * @param {Object} email Which address of the user's to send the email to.
+ * @param {Object} user The user object to generate options for.
+ * @param {String} url URL to which user is directed to confirm the email.
+ * @param {String} reason String `resetPassword` or `enrollAccount`.
+ * @returns {Object} Options which can be passed to `Email.send`.
+ * @importFromPackage accounts-base
+ */
+Accounts.tokenEmailOptions = function (email, user, url, reason) {
```
abernix: Again, not sure I agree on this function name.  Maybe `Accounts.generateOptionsForEmail`?  It's not really related to a `token` at all and this can be used for any account related e-mail, even one which may not have tokens, right?

packages/accounts-password/password_server.js
mitar: Yea, I do not like this duality between constants, but this is what we have now.

packages/accounts-password/password_server.js
mitar: Verify is different. `createPasswordToken` seems appropriate for me based on what is stored in the database (reset password token).

packages/accounts-password/password_server.js
mitar: Not sure if it can be used for other e-mails.

To me generating (or creating) often means that operation is heavy, this is why I prefer for simple operations to be more like a getter, just saying what it is what you get.

packages/accounts-password/password_server.js
mitar: Because `var` is elsewhere in this function.

abernix: I fully understand your idea of matching surrounding code, but based on that logic you should have used brace-less `if`/`else` conditionals below (for `tokenReason`) (I'm not in favor of that, just saying)

Shouldn't we incrementally move toward more commonly preferred syntax/styles whenever possible?  I don't know how far we can stretch the "surrounding"-style logic – where does it stop?  The function?  The function's function?  The parent class?  The file?

packages/accounts-password/password_server.js
abernix: It could be used for other e-mails.  The bottom line is the function does nothing with a token so it should not be called "token*".

packages/accounts-password/password_server.js
mitar: I am not used to brace-less ifs, so I missed that.

But I was just answering why `var`. I do not really care.

abernix:
This can be rebased on `devel` now that #8204 has landed, which will likely fix the tests.

Also, it appears your CircleCI builds aren't enabled for (your) `mitar/meteor` repo so [none of those](https://circleci.com/gh/mitar/meteor) ran for this PR.  I personally don't think I agree with [the way](https://circleci.com/docs/fork-pr-builds/#quick-start) that CircleCI decides to run these, but it is what it is.  You can enable up to 4 containers for free on Meteor and it will use them all.

benjamn:
👍  to rebasing to fix the tests

packages/accounts-password/password_server.js
abernix: Ok, but even if `verify` is different, does it make sense to use `createPasswordToken` to create an enrollment token?  It could be argued that you'd expect `createEnrollmentToken`.

packages/accounts-password/password_server.js
abernix: Fine either way.  Let's leave it.

abernix:
Still looking to move this PR forward!  I presume you're busy with other things at the moment, @mitar?

mitar:
Yes, I will get back to this.

abernix:
`git rebase`-ing on current `devel` to see what happens!

mitar:
Yea, I still have a tab open.

packages/accounts-password/password_server.js
port80labs: I agree that it should allow for the options to be overridden for other emails as well. For example, we are trying to override the header for emails and add xheaders that can be used by the email processor like sendgrid. these headers could be different for different emails.

hwillson:
Hi @mitar - are you still interested in finishing this PR up? Let me know if not, and I'll look into getting it wrapped up. Thanks!

mitar:
Let me look into this over the weekend. :-)

hwillson:
Hi @mitar - any luck? We're really close to getting this one merged. Let me know if you're busy and I can help wrap this one up. Thanks!

packages/accounts-password/password_server.js
mitar: Maybe we can address this duality when fixing #8084?

mitar:
So what would be a better name than `createPasswordToken`?

mitar:
I made one more iteration on this pull request. I refactored e-mail verification as well. I have not tested this in any special way and I am relaying on test suite to assure I didn't break anything. There should not be much functionality changes except for:
* `sendResetPasswordEmail`, `sendEnrollmentEmail`, `sendVerificationEmail` now return `{email, user, tokenRecord, url, options}`
* they also accept optional `extraTokenData` which is stored into tokens: this is useful to remember some extra metadata for those tokens, like which user invited which user, or what is a tracking ID for this e-mail

The next thing I would need is that one can generate multiple of those tokens. So that user can get multiple invitations and that any of those e-mails work. Or request multiple e-mail resets and that any e-mail work. E-mail verification already works like that.

mitar:
If anyone wants to change any name of anything here, feel free to just to it. Branch is editable by Meteor maintainers.

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,187 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
-    options.text =
-      Accounts.emailTemplates.resetPassword.text(user, resetPasswordUrl);
+  if (reason === 'resetPassword') {
+    tokenRecord.reason = 'reset';
   }
-
-  if (typeof Accounts.emailTemplates.resetPassword.html === 'function') {
-    options.html =
-      Accounts.emailTemplates.resetPassword.html(user, resetPasswordUrl);
+  else if (reason === 'enrollAccount') {
```
hwillson: The `else if`'s here could be moved up a line to sit beside the closing brace, to match the style used in the rest of the file.

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,187 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+/**
+ * @summary Send an email with a link the user can use to reset their password.
+ * @locus Server
+ * @param {String} userId The id of the user to send email to.
+ * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
+ * @returns {Object} Object with {email, user, tokenRecord, url, options} values.
+ * @importFromPackage accounts-base
+ */
+Accounts.sendResetPasswordEmail = function (userId, email, extraTokenData) {
+  const {email: realEmail, user, tokenRecord} = Accounts.generateResetTokenRecord(userId, email, 'resetPassword', extraTokenData);
```
hwillson: We might want to break this into 2 lines, to fit within the 80 char line width the rest of the file (mostly) adheres to.

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,187 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+/**
+ * @summary Send an email with a link the user can use to set their initial password.
+ * @locus Server
+ * @param {String} userId The id of the user to send email to.
+ * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
+ * @returns {Object} Object with {email, user, tokenRecord, url, options} values.
+ * @importFromPackage accounts-base
+ */
+Accounts.sendEnrollmentEmail = function (userId, email, extraTokenData) {
+  const {email: realEmail, user, tokenRecord} = Accounts.generateResetTokenRecord(userId, email, 'enrollAccount', extraTokenData);
```
hwillson: Same 80 char width comment as above.

packages/accounts-password/password_server.js
```diff
@@ -782,71 +850,20 @@ Meteor.methods({resetPassword: function (token, newPassword) {
...
-
-  if (typeof Accounts.emailTemplates.verifyEmail.html === 'function')
-    options.html =
-      Accounts.emailTemplates.verifyEmail.html(user, verifyEmailUrl);
-
-  if (typeof Accounts.emailTemplates.headers === 'object') {
-    options.headers = Accounts.emailTemplates.headers;
-  }
-
+  const {email: realEmail, user, tokenRecord} = Accounts.generateVerificationTokenRecord(userId, email, extraTokenData);
```
hwillson: Same 80 char width comment as above.

packages/accounts-password/password_server.js
mitar: Ehm, comment from @abernix was that I do not have to do that: https://github.com/meteor/meteor/pull/8207#discussion_r94324213

At least this is how I understood his comment. :-)

packages/accounts-password/password_server.js
mitar: Done.

packages/accounts-password/password_server.js
hwillson: Oops - disregard then - thanks!

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,185 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+  Meteor.users.update({_id: user._id}, {$push: {
+    'services.email.verificationTokens': tokenRecord
+  }});
 
+  // before passing to template, update user object with new token
+  Meteor._ensure(user, 'services', 'email');
+  if (!user.services.email.verificationTokens) {
+    user.services.email.verificationTokens = [];
+  }
+  user.services.email.verificationTokens.push(tokenRecord);
```
abernix: Can we avoid blindly modifying `user` here?  Seems out of scope for this method and seems like it would make more sense to handle in the calling-method.

packages/accounts-password/password_server.js
```diff
@@ -782,71 +848,20 @@ Meteor.methods({resetPassword: function (token, newPassword) {
...
-  if (typeof Accounts.emailTemplates.verifyEmail.html === 'function')
-    options.html =
-      Accounts.emailTemplates.verifyEmail.html(user, verifyEmailUrl);
-
-  if (typeof Accounts.emailTemplates.headers === 'object') {
-    options.headers = Accounts.emailTemplates.headers;
-  }
-
+  const {email: realEmail, user, tokenRecord} = Accounts.generateVerificationTokenRecord(userId, email, extraTokenData);
+  Accounts.saveVerificationTokenRecord(user, tokenRecord);
```
abernix: As stated above, I think it's confusing to have `Accounts.saveVerificationTokenRecord` modify `user`.  I'd feel differently if that was its only purpose (and it was called, like, `manipulateUserObject`), but it's more like an unexpected side-effect from a database operation method.

packages/accounts-password/password_server.js
mitar: I misunderstood it then. Will fix.

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,185 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+// user's emails field, or undefined to pick the first email automatically.
+//
+// This is not called automatically. It must be called manually if you
+// want to use enrollment emails.
+
+/**
+ * @summary Send an email with a link the user can use to set their initial password.
+ * @locus Server
+ * @param {String} userId The id of the user to send email to.
+ * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
```
mitar: Good catch. I missed it. I think it should.

packages/accounts-password/password_server.js
mitar: What if we make it return updated user object? Internally we clone `user` and set the value. Those are small objects so it will not be some performance hit. And then caller can decide how they want to use this method.

```
user = Accounts.saveVerificationTokenRecord(user, tokenRecord);
```

mitar:
@abernix: I think I addressed your comments.

mitar:
Any update here?

mitar:
Any news here?

packages/accounts-password/password_server.js
```diff
@@ -782,71 +864,22 @@ Meteor.methods({resetPassword: function (token, newPassword) {
...
-    options.headers = Accounts.emailTemplates.headers;
-  }
-
+  const {email: realEmail, user, tokenRecord} =
+    Accounts.generateVerificationTokenRecord(userId, email, extraTokenData);
+  const userWithTokenRecord = Accounts.saveVerificationTokenRecord(user, tokenRecord);
+  const url = Accounts.urls.verifyEmail(tokenRecord.token);
+  const options = Accounts.generateOptionsForEmail(realEmail, userWithTokenRecord, url, 'verifyEmail');
   Email.send(options);
+  return {email: realEmail, user: userWithTokenRecord, tokenRecord, url, options};
```
abernix: This seems to be more complicated than the original design I thought we were working off of [from here](https://github.com/meteor/meteor/pull/8065#issuecomment-263877938).

Why do we need to expose the new `saveVerificationTokenRecord` as a public API (and most importantly the additional maintenance cost)?  Can't it just be done inside `Accounts.generateVerificationTokenRecord`?

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,201 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+};
+
+/**
+ * @summary Saves the reset token record into database in user's document.
+ * @param {Object} user User for which to save the reset token record for.
+ * @param {Object} tokenRecord The reset token record to be saved.
+ * @returns {Object} User object with added token record.
+ * @importFromPackage accounts-base
+ */
+Accounts.saveResetTokenRecord = function (user, tokenRecord) {
```
abernix: This seems like it should be private and handled within the `generate*` methods, as we originally discussed.  I don't recall a reason to make multiple new public APIs.

packages/accounts-password/password_server.js
```diff
@@ -573,119 +573,201 @@ Accounts.sendResetPasswordEmail = function (userId, email) {
...
+};
+
+/**
+ * @summary Saves the e-mail verification token record into database in user's document.
+ * @param {Object} user User for which to save the e-mail verification token record for.
+ * @param {Object} tokenRecord The e-mail verification token record to be saved.
+ * @returns {Object} User object with added token record.
+ * @importFromPackage accounts-base
+ */
+Accounts.saveVerificationTokenRecord = function (user, tokenRecord) {
```
abernix: Same desire to be private.

packages/accounts-password/password_server.js
abernix: Your re-implementation of this is still basically doing the same thing.  It'd be better if the calling-method just re-polled the database.  Again though, we could put this in the `generate*` method, avoid the public methods and in-line the update and it would be super clear.  Making a separate function for an update doesn't seem like it is necessary, especially if it's going to change/return a different object.

hwillson:
Hi @mitar - have you had a chance to look at the recent review suggestions? I'm happy to dive in and finish this up if not. Thanks!

mitar:
I haven't yet, but just a short comment. The reason why I made it slightly more complicated now that it was initially is because one comment before was that it would be great if all token generated flows would be refactored: password reset, invite, and e-mail verification. They are not completely compatible, but there is really also a lot of similarities. (In the future we could even unify field names in subdocuments to make it simpler.)

hwillson:
Thanks for the update/explanation @mitar. Do you think you'll have a chance to respond to @abernix's review comments, and work on further changes, this week? It would be great to get this wrapped up and merged this week. If not let me know and I'll dive in to help finish this (the open PR list is really starting to grow!).

hwillson:
Assigning to myself - I'll jump in and get this wrapped up.

mitar:
Ha ha. I literally just opened this myself as well. :-)

hwillson:
Oh, no problem then @mitar! Should I leave it with you instead?

mitar:
Let me reply to comments now, and then let's see if we can get an agreement on what to do.

mitar:
So let me maybe first state my interest in this pull request and what I wanted to achieve with it:
* allow to send custom e-mails out, where I could add tracking information to those e-mails
  * add a beacon image to HTML with unique ID
  * add unique ID to e-mail headers
  * add unsubscribe URL based on same unique ID
* allow storing extra information in the token so that I could store information about who made the invite (for example, if users can invite other users, I would like to store this information into a token, to personalize whole invitation process)

(I would also like to have multiple pending invites, but that is another issue: https://github.com/meteor/meteor-feature-requests/issues/121)

Now, I think I am realizing now that I merged in this pull request two approaches to achieve this, and maybe this is really overkill. One approach is:
* you generate an unique ID, and pass it to be included (together with other extra invite information) into a token, token is generated and stored, and then you generate an e-mail with custom extra stuff, and send it
* you generate a token (but not store it yet into a database), you augment the token with extra information you want, and then you call a function to store it, then you generate an e-mail with custom extra stuff, and send it

So maybe I am forgetting something now (the reason why I made this pull request as it is), but it seems that we do not have to have two functions anymore, one to generate a token, and one to save, if we allow easy adding of custom data into a token.

mitar:
@hwillson, what would be then your take on @abernix's comments?

Do you think then this would be OK:

```js
const {email: realEmail, user, token} = Accounts.generateVerificationToken(userId, email, extraTokenData);
const url = Accounts.urls.verifyEmail(token);
const options = Accounts.generateOptionsForEmail(realEmail, user, url, 'verifyEmail');
Email.send(options);
return {email: realEmail, user: user, token, url, options};
```

hwillson:
Makes sense to me @mitar - keeping the current Accounts.save* functionality within the Accounts.generate* functions would help address the review comments questioning the need to add more to the public API.

mitar:
BTW, it is sad that we have to have both `generateVerificationToken` and `generateResetToken`, because they are very very similar, just tokens are slightly different. The former is `{token, address, when}`, the latter is `{token, email, when, reason}`. But now that we are not storing them in a separate function, there is more differences.

hwillson:
@abernix can correct me if I'm wrong, but I believe the issue was with adding more to the public API. If there are ways to simplify the codebase by extracting out re-usable chunks of functionality, then I don't think there will be any issues with that (as long as things are kept private).

mitar:
@hwillson, can you check? What do you think?

mitar:
> @abernix can correct me if I'm wrong, but I believe the issue was with adding more to the public API. If there are ways to simplify the codebase by extracting out re-usable chunks of functionality, then I don't think there will be any issues with that (as long as things are kept private).

My comment about `generateVerificationToken` and `generateResetToken` is not related to @abernix comments, but just a general observation that we could further improve this. But in backwards incompatible way.

mitar:
Tests are passing for me locally. @hwillson, feel free to do any extra changes needed to get this merged in. But I think it is now really ready to be done so.

I hope we can make also an action plan for https://github.com/meteor/meteor-feature-requests/issues/121.

hwillson:
The changes look good! I think we're all set here - fingers crossed we should be able to get this merged today. Thanks @mitar!

ayhid:
Hello, 
I just want to understand why `Accounts.emailTemplates.headers` can only be customized using an object and not a function like html and text?

PR #8212, opened 2017-01-03
Update packages/non-core/blaze to version 2.2.1

benjamn:
Though [Blaze](http://blazejs.org/) is now an independent [project](https://github.com/meteor/blaze) from Meteor core, some of its packages are still used by Meteor core, e.g. [`spacebars`](https://github.com/meteor/blaze/tree/master/packages/spacebars), used [here](https://github.com/meteor/meteor/blob/8a81e7156e07ca9e1a6c2b2b89870eb4abe72a2d/packages/boilerplate-generator/package.js#L10).

To support that use case without including a copy of Blaze in https://github.com/meteor/meteor, we use Git submodules. However, test builds have been [failing](https://travis-ci.org/meteor/meteor/builds/188608495) lately because the old revision (https://github.com/meteor/blaze/commit/dd65753c32b5535e33772ceb7ca16e998aaf6f24) can't be fetched (for reasons I have yet to diagnose). This seems to be happening with Blaze's own Circle CI suite, too: https://circleci.com/gh/meteor/blaze/228#tests/containers/0

This PR attempts to avoid the problem by updating the `packages/non-core/blaze` submodule revision to the latest release, 2.2.1: https://github.com/meteor/blaze/commit/11b315439d789f4da253d2b1e29ce91d729fa5bf

cc @mitar @hwillson


mitar:
Duplicate of #8204. ;-)

.travis.yml
```diff
@@ -5,7 +5,7 @@ cache:
   directories:
     - ".meteor"
     - ".babel-cache"
-script: TEST_PACKAGES_EXCLUDE="less" ./packages/test-in-console/run.sh
+script: TEST_PACKAGES_EXCLUDE="less,blaze" ./packages/test-in-console/run.sh
```
mitar: Not sure if this is needed, now that git submodule is fixed. Blaze is still bundled with Meteor, so it should probably be tested with it as well?

abernix:
Closing this as #8204 should take care of this.  #8204 also has the benefit of pointing the submodule to the same identical revision as the previous object (albeit a different commit hash).

.travis.yml
benjamn: You're right, it's better to run these tests than to skip them. Merged your PR instead of this one. Thanks for getting to it before me!

mitar: I broked it, so I fixed it. :-)

mitar:
BTW, removing `ui` from here would be a great thing, so that part of this pull request is something I like. Could you still merge that?

benjamn:
@mitar done

PR #8213, opened 2017-01-04
Preserve true "main" and "browser" fields of package.json modules.

benjamn:
Previously, when building a JavaScript bundle for the client, if a `package.json` file had a string-valued `"browser"` field, we would replace the value of the `"main"` field of the bundled `package.json` module with the value of the `"browser"` field. This trick was important because it allowed an npm package to have a different entry point on the client than it had on the server.

However, that approach became inconsistent if the `package.json` file was also explicitly imported as a module, because the `package.json` stub used for module resolution prevented the real contents of `package.json` from getting bundled, and disagreed with the original `package.json` module about the value of the `"main"` field.

To resolve that inconsistency, it seems better to avoid modifying the `"main"` field of `package.json` modules, and instead rely on the runtime module system to make sense of the `"browser"` field, regardless of whether the `package.json` module is a stub used only for module resolution or contains the full contents of the original `package.json` file.

The ability to understand `"browser"` fields of `package.json` modules was introduced in install@0.8.3: https://github.com/benjamn/install/commit/377d1a3b5138904d63f4594fd4c12f756454d277

This is potentially a backwards-incompatible change for developers using this version of `ImportScanner` and `Resolver` who have not yet upgraded their `modules-runtime` package to at least version 0.7.8. The solution is to upgrade `modules-runtime`, though it would be nice to enforce that better somehow.

### DIFF
--- a/History.md
+++ b/History.md
@@ -22,7 +22,16 @@
 * Added support for frame-ancestors CSP option in browser-policy.
   [#7970](https://github.com/meteor/meteor/pull/7970)
   
-* You can now use autoprefixer with stylus files added via packages [#7727](https://github.com/meteor/meteor/pull/7727)
+* You can now use autoprefixer with stylus files added via packages.
+  [#7727](https://github.com/meteor/meteor/pull/7727)
+
+* The `"main"` field of `package.json` modules will no longer be
+  overwritten with the value of the optional `"browser"` field, now that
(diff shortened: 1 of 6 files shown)


theosp:
How come a backward incompatible change doesn't lead to major increment?

PR #8227, opened 2017-01-07
EJSON.isCyclic to check circular structures. Use EJSON.isCyclic to check Meteor's methods arguments 

nlhuykhang:
- EJSON.isCyclic: new function used to check for circular structures inside objects.
- Use EJSON.isCyclic to check and throw error if there are circular structures in arguments list passed to Meteor methods 

In reference to: #5964, #8201

### DIFF
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@ -784,6 +784,11 @@ _.extend(Connection.prototype, {
       );
     }
 
+    // Check for circular structure in arguments list
+    if (EJSON.isCyclic(args)) {
+      throw new Error(`Arguments passed to ${name} contain circular structure`);
+    }
+
     // Keep our args safe from mutation (eg if we don't send the message for a
     // while because of a wait method).
     args = EJSON.clone(args);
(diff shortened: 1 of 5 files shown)


benjamn:
What if we made `EJSON.clone` just work with circular structures?

nlhuykhang:
hi @benjamn 
What do you mean by that? before I checked for circular structure inside `EJSON.clone` and threw error if there are any, but it turned out to be a bad idea because `EJSON.clone` is used a lot across the system so that change caused negative impact on the framework.

benjamn:
I mean that `EJSON.clone` should output an equivalent cyclic structure if given a cyclic structure as input, rather than failing.

While I would support that improvement, it probably would not solve #5964, but merely delay the error until we serialize the method call information to send to the server. The more general problem is that DOM elements are not serializable. Circular references are part of the problem, but the bigger problem is that DOM elements are not plain JS objects—they have a prototype different from `Object.prototype`.

I'm also worried that checking `EJSON.isCyclic` first would be slow, especially because it should return `false` in most cases, but costs the same in all cases.

For those reasons, I'm closing this PR, but I would encourage you to consider a change that checks more generally for EJSON serializability, and not just circularity.

nlhuykhang:
I get your point now, but if we make `EJSON.clone` to work with cyclic structure I think that will make the system even slower than `EJSON.isCyclic` because after all we still need to check if the input is cyclic. Furthermore `EJSON.isCyclic` is only used when making method calls, while `EJSON.clone` is used a lot across Meteor core packages. With these reasons I hope that you will think again about this PR.

Issue #5541, opened 2015-10-26
DDPRateLimiter should allow a callback for going over the limit


shaylevi2:
Assuming a user is constantly going over the set limit - I would like to automatically ban this user.

Given a method that ran everytime a user has gone over the specified limit, would allow me to do it.

stubailo:
I think this shouldn't be hard to add in the `ddp-rate-limiter` package. In fact, you could probably do it now by overriding the `getErrorMessage` function, which is basically a callback that just happens to also return the error message.

Here's the relevant code: https://github.com/meteor/meteor/blob/devel/packages/ddp-rate-limiter/ddp-rate-limiter.js

shaylevi2:
The problem with getErrorMessage is that it's a single function for all the possible limits someone might wanna use, maybe for some you wanna ban and for some you don't.

stubailo:
Yeah but you could put an if statement in there since you get all of the data in the callback argument. I'm not saying it's perfect, but it is a viable workaround for now.

shaylevi2:
You get 
{ allowed: false, timeToReset: 269, numInvocationsLeft: 0 }

No idea what function and you can't even access the connection data (i.e this.connection.id is undefined)

stubailo:
Oh weird, I would have thought that should give you the name of the DDP message.

nlhuykhang:
I think we could add a callback as the fourth argument of [`.addRule`](https://github.com/meteor/meteor/blob/devel/packages/rate-limit/rate-limit.js#L192), then execute it inside [`check`](https://github.com/meteor/meteor/blob/devel/packages/rate-limit/rate-limit.js#L131) if the limit has been exceeded. Is this ok?

abernix:
@nlhuykhang Something like that.  Perhaps it would be worth having the callback's signature be `callback(denied, allowed)` and have it called in _both_ cases?  This would allow some additional functionality (success audit logs, etc.), right?

sebakerckhof:
I think it would also make sense to modify https://github.com/meteor/meteor/blob/devel/packages/rate-limit/rate-limit.js#L131 so the input is added to the rate limit result. It seems to make sense to be able to see what caused the error in the error message (seeing that @stubailo thought this was already in there).

hwillson:
Great ideas guys! Time to make this a reality then? I'd be happy to work on this if we agree this approach makes sense.

nlhuykhang:
@hwillson I already created a PR here #8237

[...]

PR #8237, opened 2017-01-11
Add callback to DDPRateLimiter.addRule, run after a rule is executed

nlhuykhang:
In reference to #5541

### DIFF
--- a/packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
+++ b/packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
@@ -77,6 +77,68 @@ testAsyncMulti("ddp rate limiter - matchers get passed correct arguments", [
   }
 ]);
 
+testAsyncMulti("ddp rate limiter - callbacks get passed correct arguments", [
+  function (test, expect) {
+    _.extend(this, createTestUser(test, expect));
+  },
+  function (test, expect) {
+    var self = this;
+    Meteor.call("addRuleToDDPRateLimiter", expect(function(error, result) {
+     self.ruleId = result;
+    }));
+  },
+  function (test, expect) {
+    Meteor.call('dummyMethod', expect(function() {}));
+  },
+  function (test, expect) {
+    var self = this;
+    Meteor.call(
+      "getLastRateLimitEvent", expect(function (error, result) {
+        test.isTrue(result.reply.allowed);
+        test.isTrue(result.reply.timeToReset < RATE_LIMIT_INTERVAL_TIME_MS);
+        test.equal(result.reply.numInvocationsLeft, 4);
+
+        test.equal(result.ruleInput.userId, Meteor.userId());
+        test.equal(result.ruleInput.type, 'method');
+        test.equal(result.ruleInput.name, 'dummyMethod');
+      }));
+  },
+  function (test, expect) {
+    // Wait for the rule to reset
+    Meteor.setTimeout(expect(), RATE_LIMIT_INTERVAL_TIME_MS);
+  },
+  function (test, expect) {
+    // Call RATE_LIMIT_NUM_CALLS + 1 times to make the rule exceed limit and reject the execution
+    for (var i = 0; i < RATE_LIMIT_NUM_CALLS + 1; i++) {
+      Meteor.call('dummyMethod', expect(function() {}));
+    }
+  },
+  function (test, expect) {
+    var self = this;
+    Meteor.call(
+      "getLastRateLimitEvent", expect(function (error, result) {
+        test.isFalse(result.reply.allowed);
+        test.isTrue(result.reply.timeToReset < RATE_LIMIT_INTERVAL_TIME_MS);
+        test.equal(result.reply.numInvocationsLeft, 0);
+
+        test.equal(result.ruleInput.userId, Meteor.userId());
+        test.equal(result.ruleInput.type, 'method');
+        test.equal(result.ruleInput.name, 'dummyMethod');
+      }));
+  },
+  function (test, expect) {
+    Meteor.call("removeUserByUsername", this.username, expect(function () {}));
+  },
+  function (test, expect) {
+    var self = this;
+    // Cleanup
+    Meteor.call('removeRuleFromDDPRateLimiter', self.ruleId,
+      expect(function(error, result) {
+        test.equal(result,true);
+    }));
+  }
+]);
+
 testAsyncMulti("ddp rate limiter - we can return with type 'subscription'", [
   function (test, expect) {
     var self = this;
--- a/packages/ddp-rate-limiter/ddp-rate-limiter.js
+++ b/packages/ddp-rate-limiter/ddp-rate-limiter.js
@@ -64,10 +64,11 @@ DDPRateLimiter.setErrorMessage = function (message) {
  * Default = 10.
  * @param {number} timeInterval time interval in milliseconds after which
  * rule's counters are reset. Default = 1000.
+ * @param {function} callback function to be called after a rule is executed.
  * @locus Server
  */
-DDPRateLimiter.addRule = function (matcher, numRequests, timeInterval) {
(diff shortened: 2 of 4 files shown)


packages/ddp-rate-limiter/ddp-rate-limiter.js
```diff
@@ -64,10 +64,11 @@ DDPRateLimiter.setErrorMessage = function (message) {
  * Default = 10.
  * @param {number} timeInterval time interval in milliseconds after which
  * rule's counters are reset. Default = 1000.
+ * @param {function} callback function to be called after a rule is executed.
  * @locus Server
  */
-DDPRateLimiter.addRule = function (matcher, numRequests, timeInterval) {
-  return rateLimiter.addRule(matcher, numRequests, timeInterval);
+DDPRateLimiter.addRule = function (matcher, numRequests, timeInterval, cb) {
```
abernix: Nit: Any reason not to just call this `callback` instead of abbreviating it?  Would be nice to match the docs above it and use the same name like the other vars do.

packages/rate-limit/rate-limit.js
```diff
@@ -187,15 +202,20 @@ RateLimiter.prototype.check = function (input) {
...
  * rule's counters are reset. Default = 1000.
+ * @param {function} callback Optional. Function to be called after a
+ * rule is executed. An object will be passed to this callback having
+ * these properties: allowed, timeToReset, numInvocationsLeft, userId,
+ * clientAddress, connectionId.
  * @return {string} Returns unique rule id
  */
 RateLimiter.prototype.addRule = function (rule, numRequestsAllowed,
-  intervalTime) {
+  intervalTime, cb) {
```
abernix: Same nit as above, any reason not to use `callback` instead of `cb`?

packages/rate-limit/rate-limit.js
```diff
@@ -133,7 +143,10 @@ RateLimiter.prototype.check = function (input) {
   var reply = {
     allowed: true,
     timeToReset: 0,
-    numInvocationsLeft: Infinity
+    numInvocationsLeft: Infinity,
+    userId: input.userId,
+    clientAddress: input.clientAddress,
+    connectionId: input.connectionId,
```
abernix: Any reason to not return the entire `input`?  It seems like `type` (e.g. `method` or `publication`) and `name` (Of the publication or method) are still left out here and they would be very useful.

Perhaps it might be better to just pass the input to `_executeCallback` separately? (`_executeCallback(reply, ruleInput)`)?

Also, the JSDoc comment above needs to be changed to match whatever we decide.

packages/ddp-rate-limiter/ddp-rate-limiter.js
nlhuykhang: It is just my personal preference, I will update it to `callback`

packages/rate-limit/rate-limit.js
nlhuykhang: I exclude `type` and `name` because I could not think of any useful use-cases for them. If it is still useful to have them, I will include them.

abernix: Looking at https://github.com/meteor/meteor/issues/5541#issuecomment-272126029 it seems having access to the `name` is what @sebakerckhof was looking for.  I'm not sure it makes sense to expand all of the properties into the reply.  As an alternative, do you have any opinion on my suggestion above (in regards to passing `input` as a separate argument to the callback OR providing the `input` object in the reply as `reply.input`)?

sebakerckhof: Just want to confirm I was indeed interested in getting the name and type. One use case is described in https://github.com/meteor/meteor/issues/5541 , namely only banning a user for exceeding certain specific calls, but I can imagine it might be useful for other purposes too.

nlhuykhang: @abernix I already made the modification like you suggested in your earlier comment. The callback now has two arguments: the first is the result of the rule (`reply`) and other is the input (`ruleInput`). However I have not pushed the latest code because I am having difficulty with adding test cases for the callback in [this file](https://github.com/meteor/meteor/blob/devel/packages/ddp-rate-limiter/ddp-rate-limiter-tests.js). Callbacks are registered in server so I am uncertain about how to get it result. Could you give me some guidance?

nlhuykhang:
@abernix I updated this PR as our discussion and also add some tests for the new callback. Do you think this PR is good now?

packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
```diff
@@ -77,6 +77,66 @@ testAsyncMulti("ddp rate limiter - matchers get passed correct arguments", [
...
+        test.equal(result.ruleInput.userId, Meteor.userId());
+        test.equal(result.ruleInput.type, 'method');
+        test.equal(result.ruleInput.name, 'dummyMethod');
+      }));
+  },
+  function (test, expect) {
+    Meteor.setTimeout(expect(), RATE_LIMIT_INTERVAL_TIME_MS);
+  },
+  function (test, expect) {
+    for (var i = 0; i < RATE_LIMIT_NUM_CALLS + 1; i++) {
```
abernix: I believe this very subtle `+ 1` is here in order to make sure that the rate-limiter is triggered before the next test, yes?  If so, maybe add a comment to make this more obvious?

nlhuykhang: It is to make sure the callback is triggered with the reject output (method will not be executed) of the rule.

abernix: Yup, that's what I thought. Thanks for adding the comment!

packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
```diff
@@ -77,6 +77,68 @@ testAsyncMulti("ddp rate limiter - matchers get passed correct arguments", [
...
+    _.extend(this, createTestUser(test, expect));
+  },
+  function (test, expect) {
+    var self = this;
+    Meteor.call("addRuleToDDPRateLimiter", expect(function(error, result) {
+     self.ruleId = result;
+    }));
+  },
+  function (test, expect) {
+    Meteor.call('dummyMethod', function() {});
```
benjamn: Do you need to call `expect` at some point in this test function?

packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
```diff
@@ -77,6 +77,68 @@ testAsyncMulti("ddp rate limiter - matchers get passed correct arguments", [
...
+  },
+  function (test, expect) {
+    // Wait for the rule to reset
+    Meteor.setTimeout(expect(), RATE_LIMIT_INTERVAL_TIME_MS);
+  },
+  function (test, expect) {
+    // Call RATE_LIMIT_NUM_CALLS + 1 times to make the rule exceed limit and reject the execution
+    for (var i = 0; i < RATE_LIMIT_NUM_CALLS + 1; i++) {
+      Meteor.call('dummyMethod', function() {});
+    }
```
benjamn: Same question here: need to call `expect`?

packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
nlhuykhang: I think we don't need to because these calls do not expect any results from server. Are there any possible errors?

abernix: I think it's very unlikely in this environment but the next async step/test might (???) be executed before the DDP message was received and cause a failure.  I think the `async` test handler is smart enough to know that there are no actual `expect`ations here but explicitly showing that we `expect` nothing, it would make the test more clear.

abernix:
This should be released in Meteor 1.4.3.2.  You can try the latest 1.4.3.2 beta and help confirm by running:

Issue #8155, opened 2016-12-14
Update cordova to the latest 6.4.0 version


anymos:
Cordova 6.4.0 is available, most of the plugin are still of good use with the previous version.

Unfortunately the latest Cordova notification plugin (https://github.com/phonegap/phonegap-plugin-push) require the latest cordova version 6.4.0. I believe that the notification is a critical plugin for any mobile application, and consequently for Meteor

Accordingly to the discussion started  : https://forums.meteor.com/t/how-to-update-cordova-to-the-latest-6-4-0-in-meteor/31809/1

And the advise given here (https://forums.meteor.com/t/meteor-roadmap-update-12-13-16/32261/5) to have issues added to the roadmap for Meteor next version

Thank you very much in advance

abernix:
Thanks for opening this issue.  I would be in favor of making this two separate issues.  One issue (and PR, hopefully) which handles the update to 6.4.0 and is hopefully more short-term (potentially even in 1.4.3 if someone was to take it on?) and another issue to handle everything after the word "ideally" in the current title :smile: Essentially, addressing the steps necessary (and again, a separate PR) to decouple `cordova-lib`, but in what is more likely a bigger project for a future release.

As to summarizing what was necessary during the last upgrade and likely needs to be looked at for the 6.4.0 upgrade:

* [These changes on `meteor/meteor`](https://github.com/meteor/meteor/compare/cc9882b12df15488a8ff0229a5514a2c43281028...6b0fc9110b6e5fe4b4469f37910fbe109959e858), which broken down, comprise of:
  * cc9882b12df15488a8ff0229a5514a2c43281028
  * 6322a55985c8b25483a3443f848640cdfc837cc7
  * 458c5b40f05ca3a81622dee394148970393ed33a
  * f574b46adceab1db52cd6bcb9ce6086880048ba2
  * e9aa8082dc190fdebcc0e2928ae4b265b7ba4110
  * 6b0fc9110b6e5fe4b4469f37910fbe109959e858
* The last commit above is just a version bump to include the commits in `meteor/cordova-plugin-meteor-webapp` [seen here from `dd2d61...10d4dd`](
https://github.com/meteor/cordova-plugin-meteor-webapp/compare/dd2d61589c27fe41c56778cb9f7bfab41a612fbd...10d4ddbc907b57c70070e7be4eab44a4b069b966), which may not be necessary for going to 6.4.0.

skirunman:
[...]

PR #8239, opened 2017-01-11
Update to Cordova 6.4.0

wojtkowiak:
This is a Cordova update made according to #8155
Testing it right now.


wojtkowiak:
I have finished testing

abernix:
Apologies for the delay on reviewing this.  This took me quite some time to evaluate (due to environmental delays, mostly: VirtualBox, Windows, Android SDK, etc.) but I'm fairly confident that these changes are all correct.  I tested on Mac (Android and iOS) and Windows (Android only, as no iOS).  Some points worth discussing before introducing this to (hopefully?) Meteor 1.4.3:

* This upgrade requires installation of Android SDK 25, something which I don't think the last Android changes (6.2.0, for example) required (23 has been the "norm" for a while, I think).  As seen in the changes for this PR, this was Meteor's first entry into the 6.x branch of `cordova-lib`.  Cordova doesn't really make it clear [on this page](https://cordova.apache.org/docs/en/latest/guide/platforms/android/) what is the correct API platform for the 6.x branch, though I believe after reading their blog that "API 25" is the correct move.  This will need to be updated in the [Android Prereqs of the Meteor Guide](https://guide.meteor.com/mobile.html#installing-prerequisites-android).
* For whatever reason, in order to build for iOS, I needed to `gem install cocoapods`.  I'm not sure if this is a new requirement or not, but it's not mentioned in the [iOS Prereqs in the Meteor Guide](https://guide.meteor.com/mobile.html#installing-prerequisites-ios) and I don't recall needing this in the past (though it's been a while since I've built an iOS Cordova app and I'm on a new machine).
* I'm fairly sure (though not certain) this will (intentionally) block hot-code push functionality (due to changing native code, etc.).  I'm fairly sure that would have been the case during the upgrade to 6.2.0 as well (which landed in 1.4.2), but I guess it's worth mentioning.

wojtkowiak:
Thanks for the review. I have conducted tests on Windows 10, Xubuntu 16.04 and newest MacOS.

Yeah, cordova page seems to be outdated. The SDK needed is indeed 25 -> https://github.com/apache/cordova-android/commit/35b0ba6d07e7afce3554dc02458ecead76c1b0b5#diff-6899c18ece5b4dbbb91204b8356126d1 I will make PR to the guide. 

`gem install cocoapods` not sure about this one. I do not have an access to OSX right now, but I will try to verify it this week.

About HCP - yes, it changes the compatibility version for sure because the plugin versions used by the packages like webapp, statusbar etc have been changed.

GeorgeCalvert:
FYI, phonegap-plugin-push v1.9.0 uses CocoaPods.  See [this](http://stackoverflow.com/questions/40426460/taco-support-for-cordova-6-4-0-and-ios-4-3-0-cocoapods-doesnt-seem-to-work).

abernix:
@GeorgeCalvert What does that have to do with this Pull Request?

GeorgeCalvert:
@abernix, your review noted the unexpected dependency on CocoaPods.  @wojtkowiak intended to verify but I didn't find any further comment.  I just wanted to close the loop for others as our team too found the dependency unexpected.  Maybe this warrants a note in history.md, but since that references this PR, a quick comment here seemed sufficient.

abernix:
@GeorgeCalvert Ah, right! Did you also have to take action (e.g. `gem install cocoapods`)?

GeorgeCalvert:
Yep :-)   Annoying move by Cordova IMHO.

Issue #5065, opened 2015-08-29
PR Proposal: Fix HTTP param encoding logic for objects and arrays


aldeed:
At Dispatch we are overwriting the param encoding that the `http` package uses to allow it to properly handle arrays and objects. We have open sourced the fix as a package: https://github.com/DispatchMe/meteor-url-encode-params

I am looking for general community and MDG agreement that this is the correct way to handle param-to-query conversion. If there is agreement, we are happy to submit this as a PR.

stubailo:
It's pretty clear to me that your output is better than printing `[object Object]` into the param string.

Just to explore a little more: this is one of those things that doesn't seem to have a standard; what kinds of backends does this encoding work with? For example, does Rails parse your format correctly by default?

creeonix:
Rails and PHP parse that exactly as expected out of the box (to array and hashes).

`param[]=1&param[]=2` is expected to be decoded to `{"param": [1,2]}`
`param[0]=1&param[1]=2` is expected to be decoded to hash `{"param": {"0": 1, "1": 2}}`

For jQuery: http://api.jquery.com/jquery.param/

We have used same pattern in ASP.NET too and in Flex before. So, I'd say it is silent standard.
However, http://jsonapi.org/recommendations/#urls-resource-collections - uses shortened version of array encoding: `/comments?filter[post]=1,2&filter[author]=12`.

In our current project on rails API (dispatch.me) we support both ways: `filter[post][]=1&filter[post][]=2` OR `filter[post]=1,2`, second one saves when get request is getting too big.

aldeed:
Yes, what @creeonix said. :) Except it appears that node is sadly one of the few that stringifies a bit differently: https://nodejs.org/api/querystring.html

stubailo:
Sigh! This is why stringified JSON is the way to go IMO (at least if you control both ends of the wire)

Hmm, it is a shame that node doesn't parse this correctly. I guess it's not a big deal though because we could always add an option to use a different strategy later. I'll take another look at this on Monday, remind me!

stubailo:
If you could submit that pr that would be awesome!

stubailo:
Hey just a heads up, would be great to have this PR!

hwillson:
I've created PR #8261 with the proposed/discussed changes (thanks @aldeed!).

abernix:
This should be fixed in Meteor 1.4.3.1 thanks to @hwillson's PR!  Thanks!

[...]

PR #8261, opened 2017-01-17
Adjusted the url package to encode nested objects and arrays (to help address #5065).

hwillson:
Hi all - this PR is intended to help address issue #5065. I've extracted @aldeed's proposed code from the [dispatch:url-encode-params](https://github.com/DispatchMe/meteor-url-encode-params) package, adding it to the `url` package. As mentioned in the original issue, this isn't a perfect solution since Node handles this differently, but it's definitely better than the current behaviour (and can always be extended later if/as needed). Anyways, it's a start - let the discussions begin! :-)

### DIFF
--- a/packages/url/url_common.js
+++ b/packages/url/url_common.js
@@ -1,20 +1,27 @@
 URL = {};
 
-var encodeString = function(str) {
-  return encodeURIComponent(str).replace(/[!'()]/g, escape).replace(/\*/g, "%2A");
+var encodeString = function (str) {
+  return encodeURIComponent(str).replace(/\*/g, '%2A');
 };
 
-
-URL._encodeParams = function(params) {
-  var buf = [];
-  _.each(params, function(value, key) {
-    if (buf.length)
-      buf.push('&');
-    buf.push(encodeString(key), '=', encodeString(value));
-  });
-  return buf.join('').replace(/%20/g, '+');
+// Encode URL paramaters into a query string, handling nested objects and
+// arrays properly.
+URL._encodeParams = function (params, prefix) {
+  var str = [];
+  for (var p in params) {
+    if (params.hasOwnProperty(p)) {
+      var k = prefix ? prefix + '[' + p + ']' : p, v = params[p];
+      if (typeof v === 'object') {
+        str.push(this._encodeParams(v, k));
+      } else {
+        var encodedKey =
+          encodeString(k).replace('%5B', '[').replace('%5D', ']');
+        str.push(encodedKey + '=' + encodeString(v));
+      }
+    }
+  }
+  return str.join('&').replace(/%20/g, '+');
 };
-
 
 buildUrl = function(before_qmark, from_qmark, opt_query, opt_params) {
   var url_without_query = before_qmark;
--- a/packages/url/url_common.js
+++ b/packages/url/url_common.js
@@ -9,7 +9,7 @@ var encodeString = function (str) {
 URL._encodeParams = function (params, prefix) {
   var str = [];
   for (var p in params) {
-    if (params.hasOwnProperty(p)) {
+    if (Object.prototype.hasOwnProperty.call(params, p)) {
       var k = prefix ? prefix + '[' + p + ']' : p, v = params[p];
       if (typeof v === 'object') {
         str.push(this._encodeParams(v, k));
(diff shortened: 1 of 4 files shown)


aldeed:
Thanks @hwillson!

packages/url/url_common.js
```diff
@@ -1,21 +1,28 @@
...
-      buf.push('&');
-    buf.push(encodeString(key), '=', encodeString(value));
-  });
-  return buf.join('').replace(/%20/g, '+');
+// Encode URL paramaters into a query string, handling nested objects and
+// arrays properly.
+URL._encodeParams = function (params, prefix) {
+  var str = [];
+  for (var p in params) {
+    if (params.hasOwnProperty(p)) {
```
benjamn: I know this may seem like an edge case, but what if there's a parameter named `hasOwnProperty`?

hwillson: Good call @benjamn - fixing coming - thanks!

abernix:
## Result of this PR

Thought I might document a possible problem others _might_ encounter that I just found while debugging a problem with `accounts-facebook`.  As I'll try to explain here, this may be a very small edge case for some who have (miraculously?) found that having arrays serialized into a string, separated by a comma to be desired behavior.

Getting into it... the Meteor implementation of [`HTTP.call`](http://docs.meteor.com/api/http.html#HTTP-call) (via the `http` package) provides the ability to provide a `params` object containing various data parameters which are encoded into the request, either via `GET` or `POST`.

### Quick Rundown
Prior to #8261 (this PR), an `HTTP.call` with `params` of `{ fields: ["id", "email", "name"] }` would be serialized by `url` as `fields=id%2C%email%2Cname` – a comma-delimited string, but single query string parameter.

With #8261 (this PR), `HTTP.call` with the same `params` will be serialized to: `fields[]=id&fields[]=email&fields[]=name`.  This is in-line with PHP, Ruby and [`jQuery.param`](https://api.jquery.com/jquery.param/) which also use/read bracketed array notation.

### Explanation
The `HTTP.call` method [internally calls](https://github.com/meteor/meteor/blob/6a4d76ef8e78ef2e764514596028817edc0921f1/packages/http/httpcall_server.js#L46-L64) `URL._encodeParams` or `URL._constructUrl` which are provided [here](https://github.com/meteor/meteor/blob/fde9e0bbbdd985bb26ccffe0ea53d4b41255e7f4/packages/url/url_common.js#L8-L16) by the `url` package.

Previously, `HTTP.call` used a super simple, but _super_ error-prone serialization method for `Array`s which would simply join the array elements together with commas.  Actually, it didn't actually do any of that serialization itself preferring to outsource that to [`encodeURIComponent`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent), a method which is intended to deal with `String`s.  Thus, the `Array`'s `.toString()` method was being invoked, resulting in the default `Array.prototype.toString()` behavior of a comma-delimited list.

If anyone has attempted to pass an array in `params` to `HTTP.call`, they would have encountered a terrible failure if their array contained a comma as `["Hey, you", "over there"]` would have become an element longer (`Hey, you, over there`).

Similarly, if an `Object` was passed, you'd end up with the hilarity that is: `variable=[Object object]`.

### Failure

One implementation has been found so far which was actually _relying_ on this array-joining, which is Meteor's own [`accounts-facebook` package](https://github.com/meteor/meteor/blob/devel/packages/facebook-oauth/facebook_server.js#L97) which actually desires the `fields` result in the resulting request to be `id,email,name`.

### Expectations

To be honest, there's no reason why `accounts-facebook` (or any other package, inside or outside Meteor) should have been expecting a comma-delimited result to come out of this.  While `http` and `url` have tests, there were no tests in place which showed intention of making sure this `.join(',')` behavior work.  Additionally, the documentation provides zero insinuation that this would work.

### What to do?

We could revert this change as complaints were reasonably low, but the previous implementation left a lot to be desired (see `Object`, above).  I believe we'll just need to make it very clear about this change in behavior in `History.md`  (e.g. ``if you expect arrays to be serialized as comma-delimited when using `HTTP.call`, you should call `.join(',')` yourself to continue this behavior.``.  The `accounts-facebook` package can just easily be changed to do this.

Open to other suggestions too... though if you're suggesting using the [`query-string`](https://www.npmjs.com/package/query-string), I'll point out that it has a `16KB` foot-print (uncompressed/unminified).

/cc @hwillson @aldeed @benjamn

hwillson:
Thanks for the detailed breakdown @abernix! My vote goes towards keeping the change and documenting it - as you said, there's no reason why `accounts-facebook` or others should have been expecting this behaviour. Thanks again!

PR #8266, opened 2017-01-18
Fix Mongo Error Message Object

dr-dimitru:
Older error Object is:
```js
{name: String, code: Number, errmsg: String}
```

Newer mongo return error Object as:
```js
{name: String, code: Number, err: String}
```

This commit adds support for new error Object, without dropping support for older.

- Mentioned as confirmed bug at: #7758
- Partly discussed at:  #8175
- Originally fixed [at this commit](https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c)


abernix:
Thanks for submitting this.  This seems reasonable to me, but are you suggesting that this *completely* fixes #7758, or just part of it?  I was under the impression that it would require a bit more work.

dr-dimitru:
Hello @abernix ,

Thank you for review, I thought to split [this commit](https://github.com/VeliovGroup/meteor/commit/a5130970bc6f79af683ba2fd10dd4402ef2cdb6c) into  several PRs (*as was suggested*). But can't pass tests without further modifications.

Please see latest commits, I've [added test](https://github.com/meteor/meteor/pull/8266/commits/74df1f13e790de5280754180a039a7a1f9d89172) for upsert with `_id` modifier, which expects to throw an error.

Now it can't pass some tests form `mongo_livedata_tests.js`, should I fix it in this PR, or create separate PR?

abernix:
@dr-dimitru I think your original commit https://github.com/meteor/meteor/pull/8266/commits/69ffcd3cd9061ee79dd966447b000a96b757e9d6 makes sense and yes, I did suggest making it several PRs, but then it appears you've gone ahead and added the upsert stuff in this PR too – should we stick to first fixing the error messaging (as you've outlined in the original body of this PR) and then work on fixing the `upsert` separately?

dr-dimitru:
Hello @abernix ,

Sorry for delay, reverted not related commits to initial issue.
Let me know what do you think.

dr-dimitru:
Hello @abernix ,

Could you please give an update on this one?

abernix:
Tests are still failing on this [according to the last run](https://travis-ci.org/meteor/meteor/builds/202106021), so it doesn't seem that you reverted everything necessary to fix the error message?

Is it possible you still need to revert https://github.com/meteor/meteor/pull/8266/commits/cce0cd5d78fc7eb4b4e16b853f6dd793ceec9635?

dr-dimitru:
Hi @abernix ,

Thank you for quick response.

Without https://github.com/meteor/meteor/pull/8266/commits/cce0cd5d78fc7eb4b4e16b853f6dd793ceec9635 - #7758 wouldn't be fixed. Which is tested by https://github.com/meteor/meteor/pull/8266/commits/74df1f13e790de5280754180a039a7a1f9d89172

That's why I was asking what should we do - merge this one and I'll submit one more PR to solve all further failed test, or I can add more commits to this PR.

Waiting for your decision on this one.

abernix:
I'm aware that #7758 wouldn't be fixed, but the title of this PR and the original description are specifically for fixing the error message: "Fix Mongo Error Message Object".

The reason I asked [in my first comment](https://github.com/meteor/meteor/pull/8266#issuecomment-273476071) for this issue about whether it "completely fixes" 7758 was because your commit message said `Fix for #7758` – which it was not.  After that discussion, you added other commits purported to fully-fix #7758, but that is not what I had intended for you to do – I simply wanted to confirm that it was _not_ a complete fix – which it was not. 😄 

This was a bug within a bug.  Let's land this error message fix (as described in the opening of this PR, which I believe is without https://github.com/meteor/meteor/pull/8266/commits/cce0cd5d78fc7eb4b4e16b853f6dd793ceec9635) and then work on the rest of #7758.  Sound good?

dr-dimitru:
Sure, thank you for this.
Revert is pushed.

I'll be glad if you set further development, would you like to have a one big PR or many separate?
From my point of view I see next tasks:
 1. Finally fix #7758
 2. Tests for #7758 (upsert with `_id` mod)
 3. Fix connection to OpLog (Mongo 3.2) not sure how this will interfere with #8453
 4. Various fixes in `mongo_driver.js` to pass `upsert_compatibility_test.js` tests

Issue #1331, opened 2013-08-18
Meteor.userId() still returns the id on client when logging out and subscriptions restart


timhaines:
I've got a publish method that throws an error if the user isn't signed in.

If the user happens to be using the app, then signs out while subscribed to this, the subscription gets restarted as expected, and the publisher throws the 'Not signed in error'

However, the error handler for the subscription, when receiving the error from the publisher, seems to think the user is still signed in.  i.e. Meteor.userId() returns the user's ID.  At this time, I'd expect Meteor.userId() to return null, recognizing the user has signed out.  Or if that's not possible, a Meteor.loggingOut() method to show it's in process.

I can provide a simple reproduction if this isn't obvious.

hwillson:
I've created PR #8271 to help address this issue (yes, I know, it's been a while ... but better late than never, right? 🙂 ). Thanks!

hwillson:
PR #8271 has been merged, so `Meteor.loggingOut` functionality is coming soon - thanks all!

PR #8271, opened 2017-01-19
Added in Meteor.loggingOut() and related Blaze helpers.

hwillson:
Hi all - this PR is intended to help address issues #769 and #1331. It's essentially a re-working of the very old PR #826 (submitted by @raix - thanks!). I've modified it to work with the current codebase and added some tests. Let me know if you have any questions / need anything changed. Thanks!

### DIFF
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -466,11 +466,11 @@ if (Package.blaze) {
 
   /**
    * @global
-   * @name  loggingInOut
+   * @name  loggingInOrOut
    * @isHelper true
    * @summary Calls [Meteor.loggingIn()](#meteor_loggingin) or [Meteor.loggingOut()](#meteor_loggingout).
    */
-  Package.blaze.Blaze.Template.registerHelper('loggingInOut', function () {
+  Package.blaze.Blaze.Template.registerHelper('loggingInOrOut', function () {
     return (Meteor.loggingIn() || Meteor.loggingOut());
   });
 }
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -14,11 +14,8 @@ export class AccountsClient extends AccountsCommon {
   constructor(options) {
     super(options);
 
-    this._loggingIn = false;
-    this._loggingInDeps = new Tracker.Dependency;
-
-    this._loggingOut = false;
-    this._loggingOutDeps = new Tracker.Dependency;
+    this._loggingIn = new ReactiveVar(false);
+    this._loggingOut = new ReactiveVar(false);
 
     this._loginServicesHandle =
       this.connection.subscribe("meteor.loginServiceConfiguration");
@@ -46,17 +43,7 @@ export class AccountsClient extends AccountsCommon {
   // also uses it to make loggingIn() be true during the beginPasswordExchange
   // method call too.
   _setLoggingIn(x) {
-    if (this._loggingIn !== x) {
-      this._loggingIn = x;
-      this._loggingInDeps.changed();
-    }
-  }
-
-  _setLoggingOut(x) {
-    if (this._loggingOut !== x) {
-      this._loggingOut = x;
-      this._loggingOutDeps.changed();
-    }
+    this._loggingIn.set(x);
   }
 
   /**
@@ -64,8 +51,7 @@ export class AccountsClient extends AccountsCommon {
    * @locus Client
    */
   loggingIn() {
-    this._loggingInDeps.depend();
-    return this._loggingIn;
+    return this._loggingIn.get();
   }
 
   /**
@@ -73,8 +59,7 @@ export class AccountsClient extends AccountsCommon {
    * @locus Client
    */
   loggingOut() {
-    this._loggingOutDeps.depend();
-    return this._loggingOut;
+    return this._loggingOut.get();
   }
 
   /**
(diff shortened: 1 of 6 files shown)


hwillson:
I forgot to mention - I created a [small runnable app](https://github.com/hwillson/meteor-issue-769) that shows the `loggingOut` helper firing (along with `Meteor.loggingOut()`), in-case anyone wants to see these changes in action. Thanks!

raix:
Thanks, 4 years old :)

hwillson:
Hehe, I know - we're in the process of cleaning up old issues @raix so this was uncovered. There is an amazing amount of good stuff, like your changes, that got overlooked! Hopefully more of it starts to filter in soon. Thanks again!

packages/accounts-base/accounts_client.js
```diff
@@ -17,6 +17,9 @@ export class AccountsClient extends AccountsCommon {
     this._loggingIn = false;
     this._loggingInDeps = new Tracker.Dependency;
 
+    this._loggingOut = false;
+    this._loggingOutDeps = new Tracker.Dependency;
+
```
raix: maybe use `ReactiveVar`:
```js
this._loggingOut = new ReactiveVar(false)
```

packages/accounts-base/accounts_client.js
```diff
@@ -49,6 +52,13 @@ export class AccountsClient extends AccountsCommon {
     }
   }
 
+  _setLoggingOut(x) {
+    if (this._loggingOut !== x) {
+      this._loggingOut = x;
+      this._loggingOutDeps.changed();
+    }
```
raix: `this.loggingOut(x);`

packages/accounts-base/accounts_client.js
```diff
@@ -59,15 +69,26 @@ export class AccountsClient extends AccountsCommon {
   }
 
   /**
+   * @summary True if a logout method (such as `Meteor.logout`) is currently in progress. A reactive data source.
+   * @locus Client
+   */
+  loggingOut() {
+    this._loggingOutDeps.depend();
+    return this._loggingOut;
```
raix: `return this._loggingOut.get();`

packages/accounts-base/accounts_client.js
raix: Actually just skip this helper function

packages/accounts-base/accounts_client.js
```diff
@@ -59,15 +69,26 @@ export class AccountsClient extends AccountsCommon {
...
+  }
+
+  /**
    * @summary Log the user out.
    * @locus Client
    * @param {Function} [callback] Optional callback. Called with no arguments on success, or with a single `Error` argument on failure.
    */
   logout(callback) {
     var self = this;
+    self._setLoggingOut(true);
```
raix: `self._loggingOut.set(true);`

packages/accounts-base/accounts_client.js
```diff
@@ -59,15 +69,26 @@ export class AccountsClient extends AccountsCommon {
...
     self.connection.apply('logout', [], {
       wait: true
     }, function (error, result) {
+      self._setLoggingOut(false);
```
raix: and we could add `.set` here too

packages/accounts-base/accounts_client.js
hwillson: Thanks for the feedback @raix! I didn't use a `ReactiveVar` since these changes were modelled after the existing `loggingIn` code. That being said, using a `ReactiveVar` would definitely clean the code up a bit. I'll look into making these changes, and consider updating the code for `loggingIn` as well (and add some new tests for that part of the code since there don't seem to be any). Thanks again for taking a look at this!

raix: Good point

hwillson:
I updated things a bit to leverage `ReactiveVar`'s for both `loggingIn` and `loggingOut`, and added some additional tests for the `loggingIn` side of things. Thanks all!

abernix:
I'll try to review this a bit more in detail during/before next week's issue club (couldn't get to it yesterday) but first pass doesn't.  My only initial thought might be any possibility of a term aside from `loggingInOut` for the helper?  Possibly something along the lines of `authChanging`, but maybe even just adding the `Or` in there (`loggingInOrOut`)?

hwillson:
Sounds good @abernix - it's been 4 years, so no rush! 😄 I like your `loggingInOrOut` suggestion more than `authChanging`; `authChanging` might be a bit too ambiguous, since auth can be taken to mean different things (like "authentication" or "authorization"). I'll adjust - thanks again!

packages/accounts-base/accounts_client.js
```diff
@@ -438,4 +453,24 @@ if (Package.blaze) {
...
     return Meteor.loggingIn();
   });
+
+  /**
+   * @global
+   * @name  loggingOut
+   * @isHelper true
+   * @summary Calls [Meteor.loggingOut()](#meteor_loggingout).
+   */
+  Package.blaze.Blaze.Template.registerHelper('loggingOut', function () {
```
benjamn: I realize there is already code in this file that assumes `Package.blaze` is defined, but I think I would prefer if we were a bit more defensive about checking whether `blaze` is installed, since it's only a weak dependency of `accounts-base`.

hwillson: Sounds good - I'll get this resolved - thanks!

hwillson: Just reviewing this now - it looks like there is a [`Package.blaze` check just above the helper registration code](https://github.com/hwillson/meteor/blob/9b8f5a87864e2e6399964957bbeef32b8ca5606f/packages/accounts-base/accounts_client.js#L436). Do you think this is defensive enough? I can expand this further by converting the `Package.blaze` check into a `Package.hasOwnProperty('blaze')` check, paired with some additional checks to make sure the blaze object contains a `Blaze` property, but this might be overkill (I think it's pretty safe to assume `Package.blaze` really only points to a valid blaze object, if truthy). Let me know though - thanks!

mitar: I think `Package.blaze` is good enough. This is what everyone is doing for weak dependencies.

PR #7728, opened 2016-08-29
Accounts-facebook reorganize

laosb:
#7715

This PR should fix the `facebook` one. I'm not familiar with the version number based on releases, so the versioning may break though.

Issue #7715, opened 2016-08-26
Fix accounts-twitter and accounts-facebook to not force Blaze into an app


queso:
This is an open issue to discuss the idea of getting Blaze deps out of packages like accounts-twitter and accounts-facebook so that we don't force Blaze on people who may be using React/Angular/Vue, etc.

This is an extension of a discussion started on the forum and intended to discuss ideas and whatnot around this.

queso:
So @SachaG suggested that we could just make accounts-ui and Blaze a weak dependency for accounts-twitter, et al. Not sure that will work because accounts-facebook just relies on accounts-oauth and accounts-base (https://github.com/meteor/meteor/blob/devel/packages/accounts-facebook/package.js). Going to dig deeper.

queso:
So it does look like accounts-base calls in Blaze as a weak dependency (https://github.com/meteor/meteor/blob/devel/packages/accounts-base/package.js#L30), so I have no idea why it is coming along all the way atm. My test project doesn't have anything else needing Blaze. My guess is that some other package is bringing it in that accounts-base relies on.

SachaG:
I could be wrong, but I think it's: https://github.com/meteor/meteor/blob/devel/packages/twitter/package.js#L8

queso:
Nice, I just found that. It looks like the accounts-* oauth packages all rely on a package like the twitter one above. This is forcing Blaze in even thought it is a weak dep for accounts-base.

queso:
So it looks like the underlying `twitter` package and it's ilk handle two things and I think those things need to be broken out into two packages...

1) It handles the actual oAuth calls for each service
2) It contains the html/js to power the 'configuration' dialog if you are using accounts-ui.

@stubailo: Do you have a preference on how this should be handled to extract these packages apart? My gut says the functionality of #2 should be moved off to new packages... And the current `twitter`, `facebook`, and those types of packages should continue to be required by accounts-twitter.

The second option is to move the code related to #1 into accounts-twitter directly and make `twitter` a weak dep that people can install if they want the configuration.

stubailo:
Hmm, do we think anyone would ever use the Facebook package without accounts-Facebook?

Primigenus:
You know what would be great? If I add `accounts-facebook` and `accounts-ui` and it just knows which view layer my app is already using and composes the right one for me.

queso:
[...]

PR #8275, opened 2017-01-21
Accounts-google reorganize

laosb:
#7715
Related (the Facebook one): #7728 

Hi there, long time no see!

Having my winter vacation. Thinking accounts-google is popular, I reorganized accounts-google this time. I haven't got some time to test it, so getting someone to test it would be helpful.

I can't guarantee when I can do this on accounts-twitter, accounts-github and others. Pull requests for those are always encouraged.


laosb:
Travis CI said:

```
An error occurred while generating the build script.
```

Don't know what's happening.

packages/google/README.md
```diff
@@ -1,5 +1,7 @@
-# google
+# google-oauth
 [Source code of released version](https://github.com/meteor/meteor/tree/master/packages/google) | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/google)
 ***
 
-An implementation of the Google OAuth flow. See the [project page](https://www.meteor.com/accounts) on Meteor Accounts for more details. XXX link
\ No newline at end of file
+** Deprecated, use facebook-oauth instead**
```
abernix: This should be `google-oauth`.

packages/google-oauth/package.js
```diff
@@ -0,0 +1,17 @@
...
+  summary: "Google OAuth flow",
+  version: "1.2.0"
+});
+
+Package.onUse(function(api) {
+  api.use('oauth2', ['client', 'server']);
+  api.use('oauth', ['client', 'server']);
+  api.use('http', ['server']);
+  api.use(['underscore', 'service-configuration'], ['client', 'server']);
+  api.use(['random', 'templating@1.2.13'], 'client');
```
abernix: Is there a particular reason we need to depend on not only `templating` but specifically `templating@1.2.13`?  The `facebook-oauth` package doesn't use `templating` at all.

packages/google-config-ui/package.js
```diff
@@ -0,0 +1,13 @@
+Package.describe({
+  summary: "Blaze configuration templates for Google OAuth.",
+  version: "1.0.0-beta.1"
+});
+
+Package.onUse(function(api) {
+  api.use('templating@1.2.13', 'client');
```
abernix: Was there a reason for depending specifically on `1.2.13`?  We do this in `facebook-config-ui` but I don't immediately remember why.

packages/google-config-ui/package.js
```diff
@@ -0,0 +1,13 @@
+Package.describe({
+  summary: "Blaze configuration templates for Google OAuth.",
+  version: "1.0.0-beta.1"
```
abernix: Was there a reason to `-beta.1` this?

packages/google-oauth/package.js
```diff
@@ -0,0 +1,17 @@
...
+
+  api.export('Google');
```
abernix: Nit pick, but this should be after `api.addFiles` just to match the [Facebook package](https://github.com/meteor/meteor/blob/a4dad54a097f3630a1e583865044be5823e13621/packages/facebook-oauth/package.js)

packages/google/README.md
laosb: Right, my copying issue.

packages/google-config-ui/package.js
laosb: That's copied from `facebook-config-ui`.

packages/google-config-ui/package.js
laosb: I guess this was added for the release version pinning?

PR #8285, opened 2017-01-25
Fix typo in coffeescript compiler options. Fixes #8281

mjmasn:
This change, suggested by @abernix, fixes an issue (see #8281) where CoffeeScript source maps were not being generated (they appeared as blank files in the browser sources panel).

It's worth noting that these source files move from `top -> 💻app -> packages` to `top -> localhost:XXXX -> packages` in the sources list when this patch is applied (presumably because CoffeeScript files are now also passed through Babel).

PR #8288, opened 2017-01-25
CoffeeScript 1.12.3

GeoffreyBooth:
Bumps the `coffeescript` package to use CoffeeScript [1.12.3](http://coffeescript.org/#changelog), and includes #8285 (which worked when I tested it).

### DIFF
--- a/packages/coffeescript/.npm/plugin/compileCoffeescript/npm-shrinkwrap.json
+++ b/packages/coffeescript/.npm/plugin/compileCoffeescript/npm-shrinkwrap.json
@@ -1,9 +1,9 @@
 {
   "dependencies": {
     "coffee-script": {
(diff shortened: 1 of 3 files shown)


abernix:
Thanks very much, @GeoffreyBooth.  Much appreciated (for testing the other aspect too).  Even though this could be published separately right now, this should land in a Meteor 1.4.3 beta, soon-(ish) and I think the wait will be tolerable. :wink:

benjamn:
Just published [`coffeescript@1.12.3_1`](https://github.com/meteor/meteor/blob/b6c2991a28b2ac4e635081133360b55c1e4387f6/packages/non-core/coffeescript/package.js#L3) using the latest recommended Meteor version, 1.4.2.7.

Issue #6766, opened 2016-04-09
Importing browser-policy is not intuitive


joncursi:
I have installed the browser-policy package (`meteor add browser-policy`) and set up some rules:

```
BrowserPolicy.framing.disallow();
BrowserPolicy.content.disallowEval();
BrowserPolicy.content.disallowObject();
BrowserPolicy.content.disallowInlineScripts();
BrowserPolicy.content.disallowFont();
```

The Meteor Guide [recommends the Airbnb javascript style guide](http://guide.meteor.com/code-style.html#style-guide), which requires that `BrowserPolicy` be defined/imported before making the above declarations. Not being able to import the `BrowserPolicy` variable results in ESLint errors, which causes the test suite (`meteor npm run test`) to fail.

However, adding the following line to the top of the file

```
import { BrowserPolicy } from 'meteor/browser-policy';
```

results in application crash:

```
Your app is crashing. Here's the latest log:
...
TypeError: Cannot read property 'framing' of undefined
...
Exited with code: 8
Your application is crashing. Waiting for file change.
```

skirunman:
Not sure if this will work, but try:

`import { BrowserPolicy } from '[meteor/browser-policy-common](https://github.com/meteor/meteor/blob/devel/packages/browser-policy-common/package.js)';`

I don't see an export in [browser-policy package.js](https://github.com/meteor/meteor/blob/devel/packages/browser-policy/package.js) so maybe this is the issue.

joncursi:
@skirunman that worked, thanks!

skirunman:
@joncursi Glad it worked. You may want to reopen this issue as this situation is confusing. I'd like to get MDG's input on if this should be changed so that the actual package browser-policy exports `BrowserPolicy` instead of the browser-policy-common package. 

In fact, I can't really see any reason why these two packages are not just combined into one package. The only place outside of the family of browser policy packages that I can find browser-policy-common is referenced is in this [test file](https://github.com/meteor/meteor/blob/abd574f38008b45f5e2a6bc322b10bcdde44763a/packages/constraint-solver/slow-test-data.js) in the constraint-resolver package.

stubailo:
I think this falls under the set of things that should be better documented when we revamp the docs, which might be coming quite soon!

skirunman:
Interesting, any thoughts on why there is both a browser-policy and browser-policy-common package. Is this a pattern that MDG likes to use for package development? Seems unnecessary to me.

[...]

PR #8295, opened 2017-01-27
Added additional BrowserPolicy exports to help with ES2015 import support.

hwillson:
This PR is intended to help close issue #6766. Developers expect to be able to use the following import:

```
import { BrowserPolicy } from 'meteor/browser-policy';
```

but due to the way the `browser-policy*` family of packages are setup, actually have to use the following import:

```
import { BrowserPolicy } from 'meteor/browser-policy-common';
```

This PR re-exports the `BrowserPolicy` variable within each of the `browser-policy*` packages that can be used on their own:

- `browser-policy-content`
- `browser-policy-framing`
- `browser-policy`

This way each one of the following imports will work:

```js
import { BrowserPolicy } from 'meteor/browser-policy';
import { BrowserPolicy } from 'meteor/browser-policy-content';
import { BrowserPolicy } from 'meteor/browser-policy-framing';
```

Thanks!

### DIFF
--- a/packages/browser-policy-content/browser-policy-content.js
+++ b/packages/browser-policy-content/browser-policy-content.js
@@ -50,6 +50,7 @@ var keywords = {
 // If false, we set the X-Content-Type-Options header to 'nosniff'.
 var contentSniffingAllowed = false;
 
+const BrowserPolicy = require("meteor/browser-policy-common").BrowserPolicy;
 BrowserPolicy.content = {};
 
 var parseCsp = function (csp) {
@@ -300,3 +301,5 @@ _.each(resources,  function (resource) {
 });
 
 setDefaultPolicy();
+
+exports.BrowserPolicy = BrowserPolicy;
--- a/packages/browser-policy-content/package.js
+++ b/packages/browser-policy-content/package.js
@@ -1,11 +1,11 @@
 Package.describe({
   summary: "Configure content security policies",
-  version: "1.0.13-beta.1"
+  version: "1.1.0"
 });
 
 Package.onUse(function (api) {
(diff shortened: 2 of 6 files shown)


laosb:
Good to me!

hwillson:
Makes complete sense @benjamn - I'm on it - thanks!

hwillson:
Hi guys - I've adjusted `browser-policy-content` and `browser-policy-framing` to use `api.mainModule` instead of `api.export`, and make the necessary require/export calls. For `browser-policy` I've added a new `api.mainModule` file called `browser-policy.js`, that requires/exports `BrowserPolicy` from common. I've left `browser-policy-common` as is (so we still have the `BrowserPolicy` global for now, so we're not breaking anything by rolling these changes out).

Aside from the existing automated tests passing, I've also run some manual tests with a sample app that verifies:

- The `BrowserPolicy` global is still available when using the `browser-policy` / `browser-policy-content` / `browser-policy-frmaming` packages, without ES2015 imports.
- `import { BrowserPolicy } ...` works when using each of the aforementioned `browser-policy*` packages separately.

One more small thing - I bumped the package version numbers as well. Should I be doing this?

Let me know if I should change anything else. Thanks again for the feedback!

benjamn:
> One more small thing - I bumped the package version numbers as well. Should I be doing this?

Yes, thanks for doing that. When folks forget to bump the version numbers, they tend to get bumped during the release process, at which point it's difficult to know if it's a major/minor/patch change. In other words, if PR authors take responsibility for bumping version numbers, then our semantic versioning is more likely to reflect reality.

Issue #8281, opened 2017-01-24
Coffee script doesn't show source maps


Laboratory:
Hello,

We can't see source maps in DevTools
- Meteor 1.4.2.3
- Coffeescript 1.11.1_4*
<img width="348" alt="screen shot 2017-01-24 at 15 54 41" src="https://cloud.githubusercontent.com/assets/1676506/22247990/999fff36-e24d-11e6-9907-d974c15da0bb.png">

Who can help us?
Because it's not comfortable a long time no see anything.

V

klaussner:
Apparently, the [CoffeeScript compiler](https://github.com/meteor/meteor/blob/1b456d41a2de9137e75e5a3e43bfecbe449255f9/packages/coffeescript/plugin/compile-coffeescript.js#L79) doesn't add a `sourcesContent` property with the CoffeeScript code to the source map. If the compiled code is also passed through the Babel compiler (i.e., always since f4dbab623ba45dfb903c1de3a446d79426630ecf), the final source map is empty:

```js
{
  version: 3,
  sources: [],
  names: [],
  mappings: '',
  file: '/client/client.coffee.map'
}
```

abernix:
@klaussner Interesting finding.  I'm inclined to believe this might have been an API change at some point and [that this line](https://github.com/meteor/meteor/blob/1b456d41a2de9137e75e5a3e43bfecbe449255f9/packages/coffeescript/plugin/compile-coffeescript.js#L48) might just need to be changed to use `inlineMap` instead of `inline` (based on [this](https://github.com/jashkenas/coffeescript/blob/0b2d852f678694179e2d3f7e62e4ae256e60921f/src/command.coffee#L417).

Care to give it a try?  If it works, maybe a PR? 🙂.  This is likely the same as #8138.

mjmasn:
@abernix I can confirm the issue with our coffeescript apps - all `.coffee` source files are blank. Tried to look into in by running Meteor from a git checkout but it seems to remove most of the package folders from the sources list in Chrome, even with the change you suggested. Is there an obvious reason for this I might have missed?

Only thing I did was checkout meteor with `git clone --recursive git@github.com:meteor/meteor.git` and then changed `meteor ...opts...` to `~/Code/meteor/meteor ...opts...` in my bash script to run the app.

The remaining packages all contain plain js and/or css/styl files, but nothing coffee related.

| Running with meteor... | Running from checkout... |
|---|---|
[...]

PR #8298, opened 2017-01-29
Fix CoffeeScript source map composition

klaussner:
Fixes missing CoffeeScript source maps (#8281/#8285).

* CoffeeScript has an `inlineMap` option that sets the `sourcesContent` property. However, it also includes the source map at the end of the compiled code. Since we have to process the compiled code afterwards, we can't use this option and have to set the `sourcesContent` manually. There is an issue in the CoffeeScript repo for more flexible source map options: https://github.com/jashkenas/coffeescript/issues/4342.
* The `SourceMapGenerator` was created incorrectly (its constructor doesn't accept a `SourceMapConsumer`), causing the generated source map to be empty.

A note on source map size: the produced source maps are correct but larger than necessary because of the way `applySourceMap` works. If there is no mapping from compiled CoffeeScript code to its source (`return` statements aren't mapped, for example), `applySourceMap` keeps the mappings from *app.js* to the compiled code in the source map. Therefore, the original CoffeeScript code as well as its compiled version become part of the `sourcesContent` property. The source map size could be reduced by ignoring these mappings but I think it's not worth writing a custom version of `applySourceMap` just for that.

@abernix @mjmasn @GeoffreyBooth

### DIFF
--- a/packages/coffeescript/plugin/compile-coffeescript.js
+++ b/packages/coffeescript/plugin/compile-coffeescript.js
@@ -44,8 +44,6 @@ export class CoffeeCompiler extends CachingCompiler {
       literate: inputFile.getExtension() !== 'coffee',
       // Return a source map.
       sourceMap: true,
-      // Include the original source in the source map (sourcesContent field).
-      inlineMap: true,
       // This becomes the "file" field of the source map.
       generatedFile: '/' + this._outputFilePath(inputFile),
       // This becomes the "sources" field of the source map.
@@ -87,6 +85,7 @@ export class CoffeeCompiler extends CachingCompiler {
     }
 
     let sourceMap = JSON.parse(output.v3SourceMap);
+    sourceMap.sourcesContent = [source];
 
     output.js = stripExportedVars(
       output.js,
@@ -104,11 +103,19 @@ export class CoffeeCompiler extends CachingCompiler {
         doubleRoastedCoffee.data != null) {
       output.js = doubleRoastedCoffee.data;
 
-      if (doubleRoastedCoffee.sourceMap) {
+      const coffeeSourceMap = doubleRoastedCoffee.sourceMap;
+
+      if (coffeeSourceMap) {
+        // Reference the compiled CoffeeScript file so that `applySourceMap`
+        // below can match it with the source map produced by the CoffeeScript
+        // compiler.
+        coffeeSourceMap.sources[0] = '/' + this._outputFilePath(inputFile);
+
         // Combine the original CoffeeScript source map with the one
         // produced by this.babelCompiler.processOneFileForTarget.
-        const smg = new SourceMapGenerator(
-          new SourceMapConsumer(doubleRoastedCoffee.sourceMap));
+        const smg = SourceMapGenerator.fromSourceMap(
+          new SourceMapConsumer(coffeeSourceMap)
+        );
         smg.applySourceMap(new SourceMapConsumer(sourceMap));
         sourceMap = smg.toJSON();
       } else {


abernix:
Awesome!  Your logic and comments make sense to me.  Probably worth linking specifically to [your comment](https://github.com/meteor/meteor/pull/8285#issuecomment-275239673) on the other issue too for others to review.

Any chance of getting some tests included on this?  I feel like we've regressed on this CoffeeScript source maps multiple times now and would like to try to harden that up.

@mjmasn, @GeoffreyBooth: Confirmation on whether this further solves the problem?

packages/coffeescript/plugin/compile-coffeescript.js
```diff
@@ -105,10 +104,16 @@ export class CoffeeCompiler extends CachingCompiler {
       output.js = doubleRoastedCoffee.data;
 
       if (doubleRoastedCoffee.sourceMap) {
+        // Reference the compiled CoffeeScript file so that `applySourceMap`
+        // below can match it with the source map produced by the CoffeeScript
+        // compiler.
+        doubleRoastedCoffee.sourceMap.sources[0] += '.js';
```
benjamn: Maybe set `doubleRoastedCoffee.sourceMap.sources[0] = this._outputFilePath(inputFile)`, if it's the same path (I'm pretty sure it is, but haven't checked)?

klaussner: `_outputFilePath` returns the right path but without leading `/`.

klaussner:
I think I need some help with the tests. :sweat_smile: If I understand correctly, `meteor test-packages` runs the tests in an application environment, so they won't have access to the build plugin code and the source map from `compileOneFile`.

GeoffreyBooth:
@benjamn I’ve been meaning to find time to look into this, but I haven’t yet. The only idea I had regarding testing was to do a `$.get` call for the generated file (i.e. the `.js` file output by the `coffeescript` package) and inspecting that file for the source map. Is there some way to know what the URL should be for the `.js` file generated as part of the tests? Or file path, if this is easier to do server-side? (Presumably we don’t need to test for source maps in both places, either it’s in one or it’s in neither.)

klaussner:
What do you think about creating another package, `coffeescript-compiler`, that exports only the CoffeeScript compiler and use that in the `coffeescript` package, similar to `babel-compiler` and `ecmascript`? Then we would have access to code and source maps without running the build plugin.

benjamn:
I would support that split, though it doesn't necessarily need to happen before Meteor 1.4.3, as long the new version of `coffeescript` (depending on `coffeescript-compiler`) is patch-compatible with the current version.

Questions of versioning are about to become somewhat more important, since 1.4.3 will be going back to constraining core packages to be patch-compatible with the version that was current when the Meteor release was published:
* https://github.com/meteor/meteor/pull/8123/commits/868a52829fd26433baf3bde72edbd944cf02e61b
* https://github.com/meteor/meteor/pull/8123/commits/ff4b00c774ef4cbef9ddefea514b426d6694cf90

Given this change, it might make sense to move `packages/coffeescript` to `packages/non-core/coffeescript`, so that it won't be constrained by the release (similar to how it works now, post-1.4 version unpinning).

benjamn:
Another way of asking the same question: would `coffeescript` benefit in any way from being constrained by the Meteor release? Or would you rather be able to bump the version whenever and however you like? If the second option is more appealing, then we should probably do the move to `packages/non-core/coffescript` prior to 1.4.3.

GeoffreyBooth:
I guess the question is, do any Meteor core packages depend on `coffeescript`? I assume not?

Something else to keep in mind is that CoffeeScript 2.0.0-alpha1 is imminent, and will be a breaking-change release of CoffeeScript. The [breaking changes are minimal](https://rawgit.com/jashkenas/coffeescript/2/docs/v2/index.html#breaking-changes), but could affect people, especially people that use a lot of the advanced features of CoffeeScript classes. This might be a reason to let people control what version of CoffeeScript to install into a project.

benjamn:
It sounds like the balance of arguments favors keeping `coffeescript` independent from the release (i.e., moving it into `packages/non-core`). Given that, I'm going to merge this PR into the `release-1.4.3` branch, then `git mv coffeescript non-core/` so that Meteor 1.4.3 won't constrain the `coffeescript` version (and thus won't prevent upgrading to 2.0.0).

GeoffreyBooth:
Thanks @benjamn

benjamn:
Just published [`coffeescript@1.12.3_1`](https://github.com/meteor/meteor/blob/b6c2991a28b2ac4e635081133360b55c1e4387f6/packages/non-core/coffeescript/package.js#L3) using the latest recommended Meteor version, 1.4.2.7.

I tested source maps manually and they seemed to work!

GeoffreyBooth:
@benjamn I noticed that since moving `coffeescript` to non-core, it’s not included in the tests that run via `./meteor test-packages`. I eventually figured out that it can still be tested via `./meteor test-packages packages/non-core/coffeescript`, but you might want to update the regular `test-packages` command to also test the non-core packages. They may not be restricted to a particular Meteor release version, but they’re “core enough” to be part of the Meteor repo and presumably used by a great many Meteor apps. I would hope you CI server runs these tests as part of its builds.

PR #8283, opened 2017-01-24
accounts-twitter reorganize

pradel:
Related to #7715.

This pull request is for the accounts-twitter package.

PR #8302, opened 2017-01-31
Accounts-weibo reorganize

laosb:
#7715

Related:
* (the Facebook one:) #7728
* (the Google one:) #8275
* (the Twitter one by @pradel :) #8283

Weibo looks a little different from other ones. I can't even ensure it works or not previously since this package is not actively maintained.


abernix:
Are you saying that Weibo wasn't working before you made this change?  Do you think this is an important package to get working again?  Would it be difficult?

laosb:
Weibo's OAuth API didn't change since the package's last update. So it should be fine. Can't test since my Weibo.com account get banned for some reason (that I don't know).

packages/accounts-weibo/package.js
```diff
@@ -1,16 +1,18 @@
...
 });
 
 Package.onUse(function(api) {
   api.use('accounts-base', ['client', 'server']);
   // Export Accounts (etc) to packages using this one.
   api.imply('accounts-base', ['client', 'server']);
   api.use('accounts-oauth', ['client', 'server']);
-  api.use('weibo', ['client', 'server']);
+  api.use('twitter-oauth');
+  api.imply('twitter-oauth');
```
abernix: These should both be `weibo`, not `twitter`.

PR #8305, opened 2017-02-01
Wired up new meteor-developer accounts changes to separate the UI from oauth calls

hwillson:
Hi all - this PR relates to issue #7715. It provides the necessary accounts changes for `accounts-meteor-developer` (following the structure put in place for `accounts-github`). Let me know if anything should be changed. Thanks!


packages/meteor-developer-config-ui/package.js
```diff
@@ -0,0 +1,13 @@
+Package.describe({
+  summary: 'Blaze configuration templates for the Meteor developer accounts OAuth.',
+  version: '1.0.0-beta.1'
```
laosb: I guess we shouldn't add beta.1.

hwillson: Yes, I kept it that way since it was in the approved [`twitter-config-ui` package as well](https://github.com/meteor/meteor/pull/8283/files#diff-01a082e96697848e609f811415664e1e). I agree though, it doesn't seem to be needed. I'll remove it from here and the `github-config-ui` package. Thanks @laosb!

abernix: Just so it's clear, the actual release process for experimental releases (beta/RC) entails updating `-beta.`/`-rc.` suffixes for packages which are changed in that release.  This means PRs need not be submitted with the suffixes included (or they _can_ be, but unless timing is perfect, it'll likely have to be changed anyway).  Bumping the version number is the important/helpful part. :smile:

PR #8306, opened 2017-02-01
Release 1.4.2.x

benjamn:
This release branch and pull request serve to collect important changes that should be released in the next 1.4.2.x patch release (1.4.2.4 as of this writing).

Such changes include obvious bug fixes from https://github.com/meteor/meteor/pull/8123 and other more experimental branches/PRs, where "obvious" means the changes are either extremely safe, very important, or both.

If you think an issue deserves consideration for this release branch, please ask for it to be added to the https://github.com/meteor/meteor/milestone/29 milestone.

PR #8311, opened 2017-02-03
Release 1.4.2.5

abernix:
During the final stages of the release process for Meteor 1.4.2.4 (a bug fix release, #8306) it was realized that it was going to be necessary for developers to take the additional, possibly mandatory, step of running `meteor update --all-packages` or `meteor update modules-runtime`.  This is because Meteor 1.4.2.4 requires `modules-runtime@0.7.8` and its dependency on (NPM) `install@0.8.4` to function correctly.  See [this commit](https://github.com/benjamn/install/commit/377d1a3b5138904d63f4594fd4c12f756454d277) on [`install`](https://github.com/benjamn/install) or #8213 if you're interested in the logic behind that (important!) change.

Despite the simple solution (stated above) and while this was explained very clearly in the release notes, it was deemed too complex for this point release due to the additional intervention necessary by the developer.  Meteor 1.4.2.4 was not (and will not be) marked as `recommended` (i.e. made official).

In an ideal world `meteor-tool` could have forced a minimum version for this Atmosphere package to `0.7.8` but due to version unpinning this update would not have been enforced by the `constraint-solver`.  This will likely be addressed in the future.

This release will simply omit 2eab0b2bf4d635c0a7d8b3f2b1569e03b73a4325 (and associated changes) although those will remain in the upcoming 1.4.3 (#8123) release with the addition of an "upgrader" designed to take care of this automatically with no additional steps required by the end-developer.

### DIFF
--- a/packages/modules-runtime/modules-runtime.js
+++ b/packages/modules-runtime/modules-runtime.js
@@ -13,6 +13,10 @@ if (typeof Profile === "function" &&
   };
 }
 
+// On the client, make package resolution prefer the "browser" field of
+// package.json files to the "main" field.
+options.browser = Meteor.isClient;
+
 // This function will be called whenever a module identifier that hasn't
 // been installed is required. For backwards compatibility, and so that we
 // can require binary dependencies on the server, we implement the
--- a/packages/modules-runtime/modules-runtime.js
+++ b/packages/modules-runtime/modules-runtime.js
@@ -13,10 +13,6 @@ if (typeof Profile === "function" &&
   };
 }
 
-// On the client, make package resolution prefer the "browser" field of
-// package.json files to the "main" field.
-options.browser = Meteor.isClient;
-
 // This function will be called whenever a module identifier that hasn't
 // been installed is required. For backwards compatibility, and so that we
 // can require binary dependencies on the server, we implement the
--- a/History.md
+++ b/History.md
@@ -2,9 +2,15 @@
 
 ## v1.4.2.5
 
-* Reverted [#8213](https://github.com/meteor/meteor/pull/8213) as the change was deemed too significant for this release.
+* Reverted [#8213](https://github.com/meteor/meteor/pull/8213) as the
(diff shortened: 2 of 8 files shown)


packages/modules-runtime/modules-runtime.js
```diff
@@ -13,10 +13,6 @@ if (typeof Profile === "function" &&
   };
 }
 
-// On the client, make package resolution prefer the "browser" field of
-// package.json files to the "main" field.
-options.browser = Meteor.isClient;
```
benjamn: @abernix Let's not revert this change?

abernix: @benjamn Yes, it shouldn't be reverted and the commit above is actually deceiving as I didn't end up publishing `modules-runtime` in the release process for `rc.0`.  But this should accurately reflect what was published.  I'll fix this.

mitar:
Hm, why I am getting that `1.4.2.5` version is available, but there is nothing available in `History.md` in `devel` branch about it?

abernix:
@mitar You can find the `History.md` for 1.4.2.5 on the [`release-1.4.2.5`](https://github.com/meteor/meteor/blob/release-1.4.2.5/History.md) branch.  We now have the concept of more than one active release branch in order to release bug fixes at a faster velocity and independent of a more substantial release.

Obviously, we still need to figure out what exactly that means for the `devel` branch.  `History.md` seems like the most important thing to get updated at the moment though, and the recent inclusion of this in [the docs](http://docs.meteor.com/changelog.html) might be the solution for that.  It wasn't updated when the 1.4.2.5 was officially recommended, but it is [updated now](http://docs.meteor.com/changelog.html#v1425).

mitar:
Yes, it is merged now, because this pull request got merged into `deve`. ;-)

BTW, one cool thing could be to display the segment of history.md file in the console. So one could see there what is new. Instead of having to go to the GitHub.

Issue #5486, opened 2015-10-21
Google's OAuth credentials page does not permit trailing slashes.


JonTheNiceGuy:
I submitted a PR which is, as far as I can tell, nearly right, however my lack of knowledge of Javascript prevents it from being tested. The PR is #5435.

stubailo:
Sounds like a great idea to me, would be great if someone could fix up that PR.

abernix:
This is fixed with PR #8331, thanks to @TechplexEngineer and should surface in Meteor 1.4.3.

Issue #8309, opened 2017-02-02
http://localhost:3000/ doesn't work on google


pavlism:
When seting up google oauth you can't use http://localhost:3000/ as the Authorized JavaScript origins
Google will give you this error: Origin URIs must not contain a path or end with "/": http://localhost:3000/

RWizard:
Try http://lvh.me:3000

pavlism:
http://localhost:3000 works fine, (had to double check).  The screen you get when you click the configuration button says to use http://localhost:3000/ which used to work, but now causes the error and confusion.

stubailo:
Should be easy for someone to send a PR to remove the trailing slash to the Google accounts package!

TechplexEngineer:
@stubailo Would you mind reviewing my PR as I believe you reviewed a previous PR for a previous report of this issue.

abernix:
Thanks, @TechplexEngineer.  This should surface in Meteor 1.4.3.

PR #8331, opened 2017-02-08
Fix for issue 8309

TechplexEngineer:
Issue https://github.com/meteor/meteor/issues/8309 is about a trailing / not being allowed by Google's OAuth registration form. This patch makes a quick change so the Meteor package google-config-ui shows a value for the OAuth origin that is copy-pasteable directly to Google.


TechplexEngineer:
Also fixes https://github.com/meteor/meteor/issues/5486

stubailo:
Looks great to me.

PR #8342, opened 2017-02-09
Adjusted query string array representation to remove indices.

hwillson:
Hi guys - after some discussion about the implementation provided in PR #8261, it looks like we should adjust the implemented array handling a bit. Right now arrays are encoded with indices like:

```
some_array[0]=value1&some_array[1]=value2&some_array[2]=value3
```

While this implementation is somewhat supported by external languages and frameworks, it's not as well supported (by default) as the following query string representation for arrays:

```
some_array[]=value1&some_array[]=value2&some_array[]=value3
```

For example, PHP and Rails both expect **no indices** for arrays.

This PR adjusts the `url` package to make sure query string encoded arrays do not include indices. Thanks!

### DIFF
--- a/packages/url/package.js
+++ b/packages/url/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Utility code for constructing URLs",
-  version: "1.0.11"
+  version: "2.0.0"
 });
 
 Package.onUse(function(api) {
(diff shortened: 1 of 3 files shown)


hwillson:
Shoot - sure thing @benjamn, thanks!

PR #8345, opened 2017-02-09
Upgraders for reorganized accounts service packages

benjamn:
This upgrader should perform the appropriate modifications to `.meteor/packages` when users of outdated accounts service packages update to Meteor 1.4.3.

Related: #7715, #7728, #8275, #8283, #8303, #8305, #8321, #8302

### DIFF
--- a/tools/upgraders.js
+++ b/tools/upgraders.js
@@ -239,6 +239,35 @@ the guide about breaking changes here:`,
     packagesFile.writeIfModified();
   },
 
+  "1.4.3-split-accounts-packages": function (projectContext) {
+    const packagesFile = projectContext.projectConstraintsFile;
+    const usingAccountsUI = !! packagesFile.getConstraint("accounts-ui");
+
+    function split(service) {
+      if (usingAccountsUI &&
+          packagesFile.getConstraint("accounts-" + service)) {
+        packagesFile.addPackages([service + "-config-ui"]);
+      }
+
+      if (packagesFile.getConstraint(service)) {
+        packagesFile.removePackages([service]);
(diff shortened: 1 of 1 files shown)


abernix:
I had a couple more things to commit/check before I opened PR #8346, but I submit it as an alternative to this.

tools/upgraders.js
```diff
@@ -239,6 +239,35 @@ the guide about breaking changes here:`,
     packagesFile.writeIfModified();
   },
 
+  "1.4.3-split-accounts-packages": function (projectContext) {
+    const packagesFile = projectContext.projectConstraintsFile;
+    const usingAccountsUI = !! packagesFile.getConstraint("accounts-ui");
```
abernix: In #8346 (an alternative to this PR), I didn't make the addition of `<service>-config-ui` dependent on whether `accounts-ui` was present.

I did this to ensure that if someone wasn't using `accounts-ui` but was using some `templating` aspect of `<service>` (included via `accounts-<service>`), that their experience would remain the same.  Admittedly, very cautious and I couldn't decide what the reach/feasibility of this use-case was, so err-ed on the side of "staying exactly as is" (i.e. if you had `templating` before, you kept `templating`, unless you explicitly removed the newly added `<service>-config-ui` packages).

That said, doing it the way you have here might have the added benefit of automatically getting rid of the Blaze dependency for those that don't need it (as indicated by their lack of `accounts-ui`) – I'm okay with this.

benjamn:
Superseded by https://github.com/meteor/meteor/pull/8346.

Issue #8231, opened 2017-01-10
Can't download because of corporate proxy


thebetterjort:
curl https://install.meteor.com/ | sh
I've tried to pass -k and --insecure. My corporate environment decrypts all SSL traffic and our self signed cert is not trusted CA. Although I do work for the United States Government, so this probably should change ;). 

Where can I find the install script to make changes?

k5e:
This is rather a curl issue than a meteor one.
Set the proxy configuration in the environment before you run curl:
```
export http_proxy=http://proxy:8080
export https_proxy=http://proxy:8080
```
Of course, you put your own proxy address.

HTH
Christian

thebetterjort:
curl: (56) Received HTTP code 501 from proxy after CONNECT
When you using export vars. Looks like we can't use meteor.

k5e:
Does your proxy require user/password?

Put userid:password@ between // and hostname and try again.

Further, you could just download the installer script with your browser and run it yourself.


Am 10. Januar 2017 17:12:39 MEZ schrieb thebetterjort <notifications@github.com>:

thebetterjort:
No -- We basically MITM ourselves. I will go the second route.

abernix:
Meteor requires WebSockets to be proxied properly and your proxy isn't doing that.  In addition to needing the proper (standard) configuration (the same that is used by NPM, `curl`, and many other tools ) of `HTTP_PROXY` and `HTTPS_PROXY` environment variables, you need to ensure that WebSockets are permitted to be upgraded properly.  Depending on which proxy software you use, configuring a proxy to handle WebSockets correctly will take configuration.

If you're using `mitmproxy`, for example, it's as easy as changing the proxy to forward the WebSocket TCP packets instead of trying to intercept them.  See [these docs](http://docs.mitmproxy.org/en/stable/features/tcpproxy.html) for more information.  This is probably the most simple way to solve your problem.

Closing as this isn't a problem with Meteor.

PR #8346, opened 2017-02-10
Touch-ups for Accounts Package Changes

abernix:
This is a follow-up/finishing-touches to several PRs made for the accounts package splitting in #7715 of things I found during final testing, or noted in hindsight of previous merges that were easiest to fix in one foul swoop:

  * facebook: #7728
  * github: #8303
  * google: #8275
  * meetup: #8231
  * meteor-developer: #8305
  * twitter: #8283
  * weibo: #8302 

This:

* Introduces an upgrader (see alternative #8345)
* Improves the styling of `console.warn` messages to a cleaner format
  (see server [before](https://cloud.githubusercontent.com/assets/841294/22821725/1044986a-ef85-11e6-8087-b8dbca92987c.png) and [after](https://cloud.githubusercontent.com/assets/841294/22821743/219431de-ef85-11e6-88e9-c4024c606432.png) and client [before](https://cloud.githubusercontent.com/assets/841294/22821772/4187af7a-ef85-11e6-9856-2a7b7de72f70.png) and [after](https://cloud.githubusercontent.com/assets/841294/22821782/4e43bef2-ef85-11e6-82a9-4504a1496663.png))
* Fixes a bug (https://github.com/meteor/meteor/commit/406ecb7f5200b398c4cad544a84875d7aa15848a) in Facebook which cropped up because of #8261


packages/accounts-facebook/notice.js
```diff
@@ -1,5 +1,10 @@
-if (Package['accounts-ui'] && (Package['facebook-config-ui'] === undefined)) {
-  console.warn("Note: You're using accounts-ui and accounts-facebook, but didn't");
-  console.warn("install the configuration UI for Facebook OAuth.");
-  console.warn("You can install it with `meteor add facebook-config-ui`.");
+if (Package['accounts-ui'] && !Package.hasOwnProperty('facebook-config-ui')) {
```
benjamn: This is fine as long as we never release a package named `meteor:hasOwnProperty`. So let's, um, never do that.

packages/non-core/facebook/deprecation_notice.js
```diff
@@ -0,0 +1,9 @@
+console.warn(
```
benjamn: Can we make sure these deprecated packages go into `packages/deprecated` instead of `packages/non-core`? I realize I misspoke in our meeting about this, but it seems like a meaningful distinction.

abernix: Yeah, it is and that's the right place for them.  I put this here until we could discuss it and move them all at once.

packages/accounts-facebook/notice.js
abernix: I hesitated!  Agree to agree though.

packages/non-core/facebook/deprecation_notice.js
benjamn: Ah, now I remember I moved them on the https://github.com/meteor/meteor/tree/release-1.4.3 branch but not `devel`.

benjamn:
Rebase to resolve `packages/facebook-oauth/package.js` conflict?

benjamn:
I think the new conflicts are because `release-1.4.3` moves the packages to `packages/deprecated/`, so you should probably make those changes on your branch and then rebase again?

abernix:
Yeah, I foolishly rebased on `devel` the first time, but then `facebook` being in different places was cantankerous.  I think we're good now.

PR #8350, opened 2017-02-10
Implemented $bitsAllClear, $bitsAllSet, $bitsAnyClear, $bitsAnySet in minimongo

davidworkman9:
Mongo supports this with Numbers and binData. I need to implement the latter still. Wanted to get some feedback and some advice on how to write some tests. The test file for minimongo is fairly overwhelming.


abernix:
Looking pretty good!  ...and it looks like you've figured out (most) tests?  If you need help specifically with anything, please do tell! :smile:

davidworkman9:
I think I got it! I need to support buffers as a query parameter, which caused me to rewrite how I approached this completely. Pretty close now though. Should be done tomorrow.

davidworkman9:
I think this is ready now.

packages/minimongo/selector.js
```diff
@@ -649,6 +690,108 @@ ELEMENT_OPERATORS = {
...
+            var bitsSetOp = get8BitsSet(op);
+            var bitsSetVal = get8BitsSet(value[i]);
+
+            return typeof _.find(bitsSetOp, function (bit) {
+                return bitsSetVal.indexOf(bit) !== -1;
+              }) === 'undefined';
+          }) === 'undefined';
+      };
+    }
+  },
```
benjamn: There seems to be a fair bit of duplicated code between these methods. Can you factor some of that out into helper functions? For example, this expression is repeated verbatim:
```js
_.filter(operand, function (op, i) {
  var bitsSetOp = get8BitsSet(op);
  var bitsSetVal = get8BitsSet(value[i]);

  return typeof _.find(bitsSetOp, function (bit) {
    return bitsSetVal.indexOf(bit) !== -1;
  }) !== 'undefined';
})
```

packages/minimongo/selector.js
```diff
@@ -551,6 +551,47 @@ var makeInequality = function (cmpValueComparator) {
...
+};
+
+var convertNumberToUint8Array = function(number) {
+  var numOfBits = number.toString(2).length;
+  var num8BitGroups = Math.ceil(numOfBits / 8);
+  var byteArray = new Uint8Array(num8BitGroups);
+
+  for ( var index = 0; index < byteArray.length; index ++ ) {
+    var byte = number & 0xff;
+    byteArray [ index ] = byte;
```
benjamn: Please remove the extra spaces here.

packages/minimongo/selector.js
```diff
@@ -649,6 +690,108 @@ ELEMENT_OPERATORS = {
...
+      if (!isArray(operand) && !(operand instanceof Uint8Array))
+        throw Error("$bitsAllSet has to be an Array");
+      return function (value) {
+        if (typeof value !== 'number' && !(value instanceof Uint8Array))
+          return false;
+
+        operand = ensureOperandUint8Array(operand);
+
+        if (!(value instanceof Uint8Array))
+          value = convertNumberToUint8Array(value);
```
benjamn: Could you change this to something like `ensureUint8Array` and push the `instanceof` check into the function, rather than doing it multiple times in each method?

packages/minimongo/selector.js
davidworkman9: That's actually not repeated, they're all slightly different, (e.g. !== vs ===). It was really hard to wrap my head around what it's doing. I'll take a stab at reducing it even further. Ability to use ES6 would go along way ;)

davidworkman9:
Is this better?

packages/minimongo/selector.js
abernix: You should be able to use more modern ECMAScript in this file as `ecmascript` is included in its [`package.js`](https://github.com/davidworkman9/meteor/blob/40a535add85f7f793d60540b045b6260b22e9053/packages/minimongo/package.js#L19)

abernix:
Looks good!  Thanks, @davidworkman9.

yuwiggin:
I tested these bits operations, but only $bitAllSet succeeded, others failed: didn't return matched results to  $bitsAllSet, $bitsAnyClear, $bitsAnySet. 

The test environment: 
Meteor: 1.8.1
mongo: 4.0.12

Issue #8277, opened 2017-01-22
[1.4.2.x] binaries in dev bundle (i.e. `meteor npm i -g`) missing from PATH


tcastelli:
This seems to be still happening on Windows as checked here (https://forums.meteor.com/t/yarn-and-meteor/30254/15), after doing `meteor npm install -g yarn|jsdoc|whatever` , trying to run meteor "command" fails, since the installation is not performed in the correct dir (files get copied to the root of dev_bundle instead of the bin folder). 

EDIT: I used to think some packages were working, but seems that I installed them with a previous version of Meteor (which was working and placing bins on the correct folder) but with 1.4.2.3 this seems no longer the case and all get installed under the wrong folder.

EDIT2: Tried also with latest beta and problem still persists

abernix:
Thanks for opening this.  I ran into this in another issue last week, but I can't find the issue at the moment.  I looked into it briefly and it could actually be a Windows Node issue.  While Meteor does some [creative stuff](https://github.com/meteor/meteor/blob/devel/tools/cli/dev-bundle-bin-helpers.js#L110-L112) on Windows in regards to `node-gyp` compilation, it shouldn't really be causing this so I was inclined to believe that this could be a Node difference between platforms.

Marking as confirmed as the reproduction above was enough to replicate, but this will have to wait until I hop on Windows again.

tcastelli:
Maybe uptating node to latest 4.7.x would help then?

Regarding the topic change, I think this problem is not for 1.4.x in general, since at some point it was working (and I believe it was 1.4.2 maybe due a previous version of Node doing installs on the proper place). In any case one solution would be to add to PATH for execution `mt-os.windows.x86_32\dev_bundle\` folder directly in addition to `mt-os.windows.x86_32\dev_bundle\bin`, but sounds a little bit dirty since we will end with two node_modules folders and two places for binaries.

abernix:
Some notes, I just found showing some differences between the way `meteor npm` and `npm` worked (with matched versions) on Mac vs PC.

These are in a standardized pseudo form – that is to say, `<METEOR-TOOL>` was the actual full path to my Meteor tool (So, `C:\Users\IEUser\AppData\Local\.meteor\packages\meteor-tool\1.4.2_3\mt-os.windows.x86_32\dev_bundle\bin`, on Windows for Meteor 1.4.2.3)

#### Windows
* `meteor npm`:
  * `<METEOR-TOOL>/dev_bundle/bin`
[...]

PR #8355, opened 2017-02-12
Update help text to clarify the use of external npm commands

tcastelli:
First attempt to improve the help text regarding the external npm commands that can be run directly with meteor ( Initially proposed at #8277 )

Further enhancements could include listing also the "current" available external commands by parsing the bin folder with the ls/dir command or 'meteor npm list -g -depth 0'. If the external command gets uninstalled between the call of 'meteor help' and 'meteor <external command>' we are not the ones to blame but the user instead :)


tools/cli/help.txt
```diff
@@ -1,19 +1,25 @@
 >>>
 (the first line of each entry is the summary for autogenerated command lists)
-Usage: meteor [--release <release>] [--help] <command> [args]
-       meteor help <command>
+Usage: meteor [--release <release>] [--help] <built-in command> [args]
+       meteor help <built-in command>
+       meteor <external command> [args]
```
stubailo: Maybe "npm command" is better than "external command"? Since this is literally only for CLI tools installed with npm right?

benjamn: It's anything in `dev_bundle/bin`, but yes `meteor npm install -g ...` is probably the easiest/best/only realistic way of putting anything there.

tcastelli: yeah I agree, also I don't think that anyone using `dev_bundle/bin` to place custom binaries will complain about using "npm command"  in the help text 😆

tools/cli/help.txt
```diff
@@ -1,19 +1,25 @@
...
-Commands:
+Built-in Commands:
 {{commands}}
 
-See 'meteor help <command>' for details on a command.
+See 'meteor help <built-in command>' for details on a command.
+
+You can use 'meteor <external command> [args]' to run a binary command
+installed previously with 'meteor npm install -g <package with command>',
+which will store the package under the global meteor installation folder.
```
abernix: What if we changed this to say:

```
External commands (e.g. `yarn`) can be installed within Meteor itself
and used within all apps using the same version:

    meteor npm install --global <name>

And used with:

    meteor <external command>
```

Reasons being:

 * I think it's important to use `--global` instead of `-g` for the sake of clarity
 * I think`yarn` as an example is helpful since it's probably the number one use-case
 * Even though NPM tends to call them `binary` commands, they aren't really "binaries".

tcastelli: Thanks for the suggestions. Point after point:
- True, I used -g because it is shorter, but makes sense to use --global to make it clear since we are in the context of helping the user. 
- Yep, the example thing was on my mind, but was not sure if too much text was something good on the help section (and maybe was more suitable for the guide). Anyway `yarn` and `jsdoc` were my two candidates for an example, so yarn is fine by me.
- I know, we can use commands all alone to avoid missunderstandings if that helps then.

Besides all of this, I still think that a more in-depth explanation about node, npm and included tools (python, mongo, etc...) could be added to the guide too, since it is important to mention for example that with every new meteor release, we should install the npm commands again, and that if we are looking for the place were meteor stores them, is it in the dev_bundle folder.

tools/cli/help.txt
tcastelli: What about `meteor <node | npm | npm command> [args]` so we make sure to put everything in the same pack and avoid using 3 lines of use?

I would add a small line to the explanation:
Meteor bundles the `node` and `npm` tools by default. You can use `meteor node` and `meteor npm` commands to ensure that the same version of `node` and `npm` which Meteor uses internally are being used.`

tools/cli/help.txt
abernix: > I still think that a more in-depth explanation about node, npm and included tools (python, mongo, etc...) could be added to the guide too,

Absolutely!  http://docs.meteor.com/commandline.html#meteornpm could certainly be expanded to explain that `--global` has purpose!  I think `jsdoc` is super reasonable too.  Feel free to open issues/PRs with docs as appropriate! :)

> that if we are looking for the place were meteor stores them, is it in the dev_bundle folder.

The `dev_bundle` is certainly magical for most users; the fact that of `meteor npm --global` is quite different than `npm --global` is probably worth making more clear.

tools/cli/help.txt
abernix: Hmm, an "npm command" to me, is `install`, `ls`, `version`, `shrinkwrap`, etc., so not sure if I agree with "npm command".  "Third-party command"?  "Add-on command"?

abernix: I'm all for using the `<node | npm | blah>` notation! :)

stubailo: Right I'm just saying that putting "npm" in there makes sense since that's where these commands come from. Or maybe just saying more in the docs is enough like "get more commands by npm installing them"

tcastelli: What about "command from npm package"? It is longer but more precise

benjamn:
@abernix @tcastelli @hwillson Does my edit look good?

tools/cli/help.txt
```diff
@@ -17,21 +17,26 @@ Built-in Commands:
...
 the same versions of 'node' and 'npm' that Meteor uses internally.
 
-Npm commands like 'yarn', 'jsdoc', etc., can also be installed within
-Meteor itself and used within all apps using the same version.
-
-First install them in your Meteor global context with:
+Other commands can be installed in the same directory as 'node' and
+'npm' and used in all apps using the same version of Meteor. An easy
+way to install such commands is to 'meteor npm install --global' an
+npm package ('yarn', 'jsdoc', etc.) that provides the command:
```
tcastelli: I would change `to install such commands is to 'meteor npm install --global' an npm package ('yarn', 'jsdoc', etc.) that provides the command`
with simply ` to install such commands is to add an npm package ('yarn', 'jsdoc', etc.) that provides the command with:`

The rest looks perfect to me 👍

Issue #8349, opened 2017-02-10
[1.4.2.6] nested node_modules from a meteor package not being installed


sebakerckhof:
When upgrading from meteor 1.4.2.6 my application started to throw errors about modules not being found.

Upon inspection I found the module was a module that was a dependency of another npm package that was a dependency in `package.js` file of a local meteor package.
Downgrading to 1.4.2.3 fixes the problem.

So it's like this:
Meteor package -> Npm package 1 -> Npm Package 2
Npm package 2 is not being installed with meteor 1.4.2.6, but it is in meteor 1.4.2.3

The `npm-shrinkwrap.json` file inside the `package/.npm` folder contains only 1 entry: the direct dependency listed in the `package.js` file when using meteor 1.4.2.6. The one when using meteor 1.4.2.3 contains a whole list of entries

sebakerckhof:
Update: strangely enough, I only seem to have this problem when using my private npm server.

So situation is like this:
 - Private NPM server + Meteor 1.4.2.3 -> OK
 - Public NPM server + Meteor 1.4.2.6 -> OK
 - Public NPM server + Meteor 1.4.2.6 -> NOT OK

benjamn:
Are these local Meteor packages? What happens if you delete the `.npm` directory entirely (should force it to be recreated)?

sebakerckhof:
Yep, local meteor package. The behavior described above was by each time doing a `meteor reset` and deleting the `package/.npm` folder in between. 
However, I've done some more tests now and now I was able to get it to work on 1.4.2.6. So it seems to be not reliably reproducible and therefore my findings above might not be correct.

As a side note, I just notice that the npm-shirnkwrap looks something like:
```
...
"azure-storage": {
      "version": "0.10.0",
      "resolved": "https://bin.barco.com/artifactory/api/npm/npm-virtual/azure-storage/-/azure-storage-0.10.0.tgz",
      "from": "azure-storage@0.10.0"
    },
    "bcrypt-pbkdf": {
      "version": "1.0.1",
      "resolved": "https://registry.npmjs.org/bcrypt-pbkdf/-/bcrypt-pbkdf-1.0.1.tgz",
      "from": "bcrypt-pbkdf@>=1.0.0 <2.0.0"
    },
    "bl": {
      "version": "1.0.3",
      "resolved": "https://registry.npmjs.org/bl/-/bl-1.0.3.tgz",
      "from": "bl@>=1.0.0 <1.1.0"
    },
...
```

[...]

PR #8361, opened 2017-02-13
Release 1.4.2.7

benjamn:
This pull request reverts the upgrade of `npm` to version 4.1.2 from 3.10.9. That upgrade first appeared in Meteor 1.4.2.4.

Although we thought the upgrade from `npm` 3.10.9 to 4.1.2 was worthwhile and safe, this breaking change proved problematic: https://github.com/npm/npm/blob/latest/CHANGELOG.md#no-more-partial-shrinkwraps-breaking

Specifically, if a Meteor package uses `Npm.depends`, and does not yet have an `.npm/package/npm-shrinkwrap.json` file, Meteor will create a partial shrinkwrap file in order to install npm dependencies, but (with the changes in `npm@4`) transitive dependencies of the package will no longer be installed.

Upgrading `npm` to a new major version was probably too much of a change for a 1.4.2.x release, anyway, so we're reverting it for 1.4.2.7.

In case you can't wait for 1.4.2.7, you can "fix" this problem for
previous versions of Meteor by running
```sh
meteor npm install --global npm@3.10.9
```

You can test that this downgrade worked by running
```sh
meteor npm version
```


benjamn:
Note that this is not branched off of latest `master`, since [1.4.3](https://github.com/meteor/meteor/pull/8123) has already landed. After publishing the official release tag for 1.4.2.7, we should merge `master` *into* this branch (resolving the conflicts) before merging this PR back into `master`.

sebakerckhof:
Seems like this could explain https://github.com/meteor/meteor/issues/8349, right?

benjamn:
@sebakerckhof Yep! That's why @abernix [added](https://github.com/meteor/meteor/issues/8349#event-960027490) it to the [milestone](https://github.com/meteor/meteor/milestone/30) just now.

PR #8362, opened 2017-02-13
Fix comment about unordered imports

### LINKED ISSUE #5279: Unordered dependency exports inconsistency
**Problem**

When `{unordered: true}` is used for a dependency, the exports are still available without using the `Package` object. When running velocity package tests, though, the exports are _not_ available. I thought this was an issue with velocity, but then there is this comment, which disagrees with how things actually work: https://github.com/meteor/meteor/blob/a23cc0249217715cd160112a47c2fff6f5becf89/tools/isobuild/package-api.js#L119-L122

**Reproduction**

https://github.com/aldeed/temporary-testUnordered

**Resolution**

Need clear decision from MDG on what the correct behavior is, and then either Meteor tool or velocity needs to be fixed to match.

aldeed:
Delete a misleading comment as discussed in https://github.com/meteor/meteor/issues/5279#issuecomment-146018275

Resolves #5279

### DIFF
--- a/tools/isobuild/package-api.js
+++ b/tools/isobuild/package-api.js
@@ -127,9 +127,8 @@ _.extend(PackageAPI.prototype, {
   // options can include:
   //
   // - unordered: if true, don't require this package to load
-  //   before us -- just require it to be loaded anytime. Also
-  //   don't bring this package's imports into our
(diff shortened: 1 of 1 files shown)


hwillson:
This looks great to me - thanks!

PR #8379, opened 2017-02-15
Update generated README

dr-dimitru:
- Update node version in generated README file by bundler, according to version from `.node_version.txt` file


abernix:
Thanks for submitting this!

Can we just change it to use `process.version` which is used to write the `.node_version.txt` file (just above that code) instead of needing to remember this each time we update Meteor's `node`?

Also, we should probably remove the `or newer` part of the text since this terminology has caused users to try to use Node 6 (or newer) in the past, which can lead to problems.

dr-dimitru:
@abernix agree,

Just thought there is a reason to hard-code supported version, but I guess there is no way to build an app with some other `node`.

Pushed changes.

abernix:
Thanks, @dr-dimitru!

Issue #8329, opened 2017-02-08
Javascript reserved words / transformed collections bug


dburles:
This issue was originally brought to my attention via the `dburles:collection-helpers` package by @RaniereSouza and @vsivsi. 

In spending some time investigating it further it appears that using a transformed document containing a numeric field named `length` as a query selector seemingly mangles the resulting query.

Simple reproduction (transform example from the Meteor documentation):

```js
Animal = function (doc) {
  _.extend(this, doc);
};

_.extend(Animal.prototype, {
  makeNoise: function () {
    console.log(this.sound);
  },
});

export const Test = new Mongo.Collection('test', {
  transform: function (doc) { return new Animal(doc); }
});

const docs = [
  { length: 285113 },
  { length: 14203 },
  { length: 133052 },
];

docs.forEach(doc => Test.insert(doc));

const doc = Test.findOne();

test.equal(Test.find(doc).count(), 1);
```

In this example the test fails as the resulting query returns 3 documents instead of 1.

[Full reproduction app](https://github.com/dburles/transform-bug/)

Note that this also affects the `remove` query.

hwillson:
Thanks for the excellent reproduction @dburles - I'm seeing the same thing.

hwillson:
Interesting - the problem happens here: https://github.com/meteor/meteor/blob/3801349437dc7be908f152783488d17ea6fb29d4/packages/mongo/collection.js#L381

When `key` contains `length`, bad things happen within the `_.each` call. Turns out this is a known underscore issue: https://github.com/jashkenas/underscore/issues/1590

I haven't digested that entire thread yet (not enough coffee!), but I'll take a closer look shortly. Replacing the `_.each` call with something else should fix this issue. I'll try to get a PR submitted later today (if no-one else gets to it before me). Thanks!

hwillson:
Ha - I just noticed this in the [underscore `each` docs](http://underscorejs.org/#each):

> Note: Collection functions work on arrays, objects, and array-like objects such as arguments, NodeList and similar. But it works by duck-typing, **so avoid passing objects with a numeric length property**.

🙂

vsivsi:
Just some history on this Underscore `each` problem in Meteor (it goes *way* back...)

https://github.com/meteor/meteor/blob/devel/packages/underscore/underscore.js#L88

My understanding has been that Meteor "fixed" this issue in Underscore by forking and patching its own private copy.  

I'd be very interested to know how this bug evaded that "fix".

[...]

PR #8380, opened 2017-02-16
Replaced _.each with Object.keys().forEach to fix objects with length properties being used as Mongo selectors

hwillson:
Hi guys - this PR is intended to help address issue #8329. Objects that contain a `length` property are causing issues when passed into the [`Mongo.Collection._rewriteSelector`](https://github.com/meteor/meteor/blob/907e1d91833eace7e40cc1d796cc7f9d171ceab1/packages/mongo/collection.js#L365) function. This is because the `_rewriteSelector` function is making a `_.each` call that is inappropriately treating the incoming selector object as an array. This PR replaces the `_.each` call with a `Object.keys(selector).forEach` call, which properly handles objects that have `length` properties.

I've updated the issue with the full troubleshooting details. Standing by for feedback - thanks!

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -378,7 +378,8 @@ Mongo.Collection._rewriteSelector = function (selector) {
     return {_id: Random.id()};
 
   var ret = {};
-  _.each(selector, function (value, key) {
+  Object.keys(selector).forEach((key) => {
+    const value = selector[key];
     // Mongo supports both {field: /foo/} and {field: {$regex: /foo/}}
     if (value instanceof RegExp) {
       ret[key] = convertRegexpToMongoSelector(value);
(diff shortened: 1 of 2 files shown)


packages/mongo/collection.js
```diff
@@ -388,8 +389,7 @@ Mongo.Collection._rewriteSelector = function (selector) {
       // override the ones set on $regex.
       if (value.$options !== undefined)
         ret[key].$options = value.$options;
-    }
-    else if (_.contains(['$or','$and','$nor'], key)) {
+    } else if (_.contains(['$or','$and','$nor'], key)) {
```
stubailo: This isn't super relevant, but how do we feel about rewriting code like this as:

```js
{ $or: 1, $and: 1, $nor: 1 }[key]
```

abernix: I'm not sure what the temptation would be and I don't personally find that formatting appealing, if that's your own temptation. More specifically: `O(1)` lookup-time doesn't seem applicable here and building that `Object`-dictionary would presumably be a tad slower than the `Array`. 

In similar spirit to the other Underscore replacement this issue addresses, my preference would be: 

```js
['$or','$and','$nor'].includes(key);
```

...but that's outside the scope of this PR and I'd vote against changing it here.  (The curly-bracket readjustment seems welcome though, as the previous formatting seems broken compared to the `else` aside it.

benjamn: @stubailo That idiom won't do what you want if `key === "toString"` or anything else on `Object.prototype`.

PR #8327, opened 2017-02-07
Release 1.5

benjamn:
Just as Meteor 1.4.2 took aim at rebuild performance, Meteor 1.5 will be all about production app performance, specifically client-side application startup time.

The banner feature of this effort will be first-class support for [dynamic `import(...)`](https://github.com/tc39/proposal-dynamic-import), which enables asynchronous module fetching (sometimes referred to as "code splitting"). Jump to [this comment](https://github.com/meteor/meteor/pull/8327#issuecomment-281428641) for an overview of how it works.

As usual, this release will include any bug fixes since [Meteor 1.4.3](https://github.com/meteor/meteor/pull/8123), though special attention will be paid to pull requests and issues that affect the performance of running applications.

**You can update to the latest release candidate by running**:
```
meteor update --release 1.5-rc.12
```

PR #8388, opened 2017-02-17
Remove jQuery from default dependencies.

M4v3R:
This pull request removes jQuery package from default project skeletons.
Also removes it from http package because it didn't use it at all.

### DIFF
--- a/packages/http/package.js
+++ b/packages/http/package.js
@@ -27,7 +27,6 @@ Package.onTest(function (api) {
   api.use('webapp', 'server');
   api.use('underscore');
   api.use('random');
-  api.use('jquery', 'client');
   api.use('http', ['client', 'server']);
   api.use('tinytest');
   api.use('test-helpers', ['client', 'server']);
--- a/packages/http/package.js
+++ b/packages/http/package.js
@@ -27,6 +27,7 @@ Package.onTest(function (api) {
   api.use('webapp', 'server');
   api.use('underscore');
   api.use('random');
+  api.use('jquery', 'client');
   api.use('http', ['client', 'server']);
   api.use('tinytest');
   api.use('test-helpers', ['client', 'server']);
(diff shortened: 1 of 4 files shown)


abernix:
Thanks for opening this as a separate issue from the #8327 thread.

I think this is reasonable to consider, but we probably need to take inventory of and consider tutorials (and expectations) that may exist out there right now before merging this.

Could you actually open the `http` part as a separate PR and remove it from this one so we compartmentalize the changes?  That part to me seems like a no-brainer (thanks for finding that!). :wink:

abernix:
Also, can you please enable your CircleCI builds for `meteor` so we can see if the tests pass?  Currently the CircleCI build did not run on this PR:

* Go to https://circleci.com/add-projects
* Click your username on the left
* Find `meteor` on the right
* Click *Build project* next to `meteor`

M4v3R:
@abernix Done.

abernix:
Thank you for submitting this.  I fully agree with your last statement.

Merged manually to `devel` with bf1b58e80a53863a8db505fa1252d31948614af7.  In the future, please submit PRs from the `devel` branch to the `devel` branch.

I've left this PR targeted to `release-1.5` to simply the viewing experience (to exclude 1.5 changes) of this PR. :wink:

abernix:
This should be done in Meteor 1.4.3.2.  You can try the latest 1.4.3.2 beta and help confirm by running:

haojia321:
If I add jquery from npm, does that mean I have to explicitly import jquery in every js file(assume I use jquery in every js file)?

stubailo:
You could do `window.$ = require('jquery')` but I wouldn't recommend it, importing everywhere is a much better approach.

abernix:
@haojia321 While I fully agree you should prefer an approach (as @stubailo recommended above) where you import in every file )which you'll likely appreciate later and avoids polluting your global namespace) if you'd like to keep the behavior where jQuery is available everywhere with no additional code you can install jQuery from npm, keep the `jquery` package in your Meteor project and it will take care of the `window.$ = require('jquery')` for you, preferring your npm-installed version of jQuery:

```sh
$ meteor add jquery
$ meteor npm install jquery --save
```

PR #8389, opened 2017-02-17
Remove jQuery dependency from http package

M4v3R:
This is to remove jQuery as a default dependency from `http` package, as a part of removing jQuery altogether as a default dependency from Meteor. See #8388 for more details.

### DIFF
--- a/packages/http/package.js
+++ b/packages/http/package.js
@@ -27,7 +27,6 @@ Package.onTest(function (api) {
   api.use('webapp', 'server');
   api.use('underscore');
   api.use('random');
-  api.use('jquery', 'client');
   api.use('http', ['client', 'server']);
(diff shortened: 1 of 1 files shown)


abernix:
Thanks, @M4v3R!

abernix:
Just realized this was on the `release-1.5` branch.  I will cherry-pick it onto `devel` now, but in the future, please open all pull-requests against the `devel` branch (I should have caught that before I merged, to be fair).  Granted, yes, this will be in release 1.5 it's still best if we go from `devel` to `release-*`.

abernix:
This should be released in Meteor 1.4.3.2.  You can try the latest 1.4.3.2 beta and help confirm by running:

PR #8394, opened 2017-02-20
fix for #8160 Part 2

brucejo75:
@ManuelDeLeon, discovered a flaw with PR https://github.com/meteor/meteor/pull/8166 which is a fix for #8160.  It did not work for sub-classes of an Array.

Changed all instances of checking the array to be `seq instanceof Array || _.isArray(seq)`.

Passes all observe-sequence tests.

**PS** : I thought I would add a test for a sub-classed array.  But, whoa!  It is hard to [sub-class an array](http://perfectionkills.com/how-ecmascript-5-still-does-not-allow-to-subclass-an-array/) in javascript.  Coffeescript makes it easy apparently and the [viewmodel](https://github.com/ManuelDeLeon/viewmodel) package uses Coffeescript.

### DIFF
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -94,7 +94,7 @@ ObserveSequence = {
 
         if (!seq) {
           seqArray = seqChangedToEmpty(lastSeqArray, callbacks);
-        } else if (_.isArray(seq)) {
+        } else if (seq instanceof Array || _.isArray(seq)) {
           seqArray = seqChangedToArray(lastSeqArray, seq, callbacks);
         } else if (isStoreCursor(seq)) {
           var result /* [seqArray, activeObserveHandle] */ =
@@ -126,7 +126,7 @@ ObserveSequence = {
   fetch: function (seq) {
     if (!seq) {
       return [];
-    } else if (_.isArray(seq)) {
+    } else if (seq instanceof Array || _.isArray(seq)) {
       return seq;
     } else if (isStoreCursor(seq)) {
       return seq.fetch();


abernix:
Much in the same way that CoffeeScript [makes it easy](http://coffeescript.org/#try:class%20YourOwnArray%20extends%20Array), modern JavaScript constructs in ECMAScript 2015 (ES6) also make it easy: `class YourOwnArray extends Array {}`.

If you add the `ecmascript` module to the `Package.onTest` section of `observe-sequence`'s `package.js` and add the test in the test suite, Meteor will take care of [Babel-ing](https://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-0&targets=&browsers=&builtIns=false&experimental=false&loose=false&spec=false&code=class%20IterableBears%20extends%20Array%20%7B%7D&playground=true) as necessary.

brucejo75:
I think I tried that and I was not getting it to work.  Can you have a look?

I put the repo [here](https://github.com/brucejo75/test2).  This is a clone of ManuelDeLeon's repo and I added the `class SC` that extends `Array`.

I run it and stick a breakpoint on line 11 of `main.js`.

When I enter these commands I am surprised:
```
> seq
Array[3]
0: 1
1: 2
2: 3
length: 3
__proto__: Array[0]
> seq instanceof SC
false
> seq instanceof Array
true
> _.isArray(seq)
true

```
I thought that `seq instanceof SC` should be `true` and `_.isArray(seq)` should be `false`.  But they are just the opposite.

So this made me think that I was not getting it right.

Then I take a look at `this.items()`:

```
> var i = this.items()
undefined
> i
ReactiveArray[2]
0: 1
1: 2
changed: ()
depend: ()
length: 2
pause: ()
resume: ()
__proto__: Array
> i instanceof ReactiveArray
VM1676:1 Uncaught ReferenceError: ReactiveArray is not defined
    at eval (eval at evaluate (:85:21), <anonymous>:1:14)
    at ViewModel.log_and_return_items (http://localhost:3000/app/app.js?hash=2ca801461058efc9fd804030345100f262ba1d0a:70:5)
    at Object.helpers.(anonymous function) (http://localhost:3000/packages/manuel_viewmodel.js?hash=e51d535c3b99a505a56178ad5f20dbad7d64803e:243:41)
    at http://localhost:3000/packages/blaze.js?hash=813922cefaf3c9d7388442268c14f87d2dde795f:3050:16
    at http://localhost:3000/packages/blaze.js?hash=813922cefaf3c9d7388442268c14f87d2dde795f:1714:16
    at http://localhost:3000/packages/blaze.js?hash=813922cefaf3c9d7388442268c14f87d2dde795f:3102:66
    at Function.Template._withTemplateInstanceFunc (http://localhost:3000/packages/blaze.js?hash=813922cefaf3c9d7388442268c14f87d2dde795f:3743:12)
    at http://localhost:3000/packages/blaze.js?hash=813922cefaf3c9d7388442268c14f87d2dde795f:3101:27
    at Object.Spacebars.call (http://localhost:3000/packages/spacebars.js?hash=ebf9381e7fc625d41acb0df14995b7614360858a:172:18)
    at http://localhost:3000/app/app.js?hash=2ca801461058efc9fd804030345100f262ba1d0a:13:22
(anonymous) @ VM1676:1 ...
> i instanceof Array
true
> _.isArray(i)
false
```

And I do not know how to get my test class to act like @ManuelDeLeon's `ReactiveArray`.

Any help appreciated...

nzomedia:
Hi, I'm currently facing a related problem: I use  numtel:mysql to access data from mysql. Everything worked fine until i updated Meteor to version 1.4.2.7. The MysqlSubscription object's reactive() method (which i use to return data to blaze template)  returns a value whish extends Array. Blaze is failing with an error: Uncaught Error: {{#each}} currently only accepts arrays, cursors or falsey values.

How can i do to solve the problem? i don't know if i should modify observe_sequence.js in .../AppData/local/.meteor/observe-sequence folder by adding "seq instanceof Array" or if there is a better solution.
Thanks

ManuelDeLeon:
@brucejo75 

The devil is in the details:

```js
    const hasProp = {}.hasOwnProperty;
    const extend = function (child, parent) { for (var key in parent) { if (hasProp.call(parent, key)) child[key] = parent[key]; } function ctor() { this.constructor = child; } ctor.prototype = parent.prototype; child.prototype = new ctor(); child.__super__ = parent.prototype; return child; };
    const SC = (function (superClass) {
      extend(SC, superClass);

      function SC() { }

      return SC;

    })(Array);
    
    const seq = new SC();
    
    console.log("seq instanceof SC", seq instanceof SC);
    console.log("seq instanceof Array", seq instanceof Array);
    console.log("seq isArray", _.isArray(seq));
    
    console.log("items instanceof ReactiveArray", this.items() instanceof ReactiveArray);
    console.log("items instanceof Array", this.items() instanceof Array);
    console.log("items isArray", _.isArray(this.items()));
```

```
seq instanceof SC true
seq instanceof Array true
seq isArray false

items instanceof ReactiveArray true
items instanceof Array true
items isArray false
```

brucejo75:
Hi @nzomedia, this is the pull request that will solve your problem.  In the meantime to work around it you can patch the code in your project or your dev bundle.

Hi @ManuelDeLeon, Thanks!  I saw that code when I did the coffeescript translate.  I figured there must be a more straightforward way, maybe I was wrong.  I will use your suggestion to update the tests to include derived classes and submit it.

brucejo75:
Hi @ManuelDeLeon,

I need to set the array in the subclass.  The child constructor `SC() {}` does not set the array.  How do I do that?  I have tried a dozen ways and I think there are 100 other things I can try.  Maybe you can provide a shortcut?

ManuelDeLeon:
splice it

```js
      function SC() { 
        if (arguments.length > 0) {
          const items = this.slice.call(arguments, 0);
          this.splice.apply(this, [0, 0].concat(this.slice.call(items)));
        }
      }
```

brucejo75:
OK, now I have another problem.

I submitted this PR against an old fork I had.  For the submit, I had not updated the fork (the code in `observe-sequence` had not changed since my earlier fork).

I attempted to update my clone by doing a `git fetch upstream` but that ran into a failure I could not resolve.  Then I tried re-cloning the old fork and that could not complete because of this error:

```
error: no such remote ref dd65753c32b5535e33772ceb7ca16e998aaf6f24
Fetched in submodule path 'packages/non-core/blaze', but it did not contain
dd65753c32b5535e33772ceb7ca16e998aaf6f24. Direct fetching of that commit failed.
```
Suffice it to say my old Fork is unusable.

I am going to close this PR request and open a new one against a newer fork... sigh.

abernix:
Looks like you were working on a very old version of the repository.  Unfortunately, because of https://github.com/meteor/blaze/issues/112#issuecomment-269746281, which re-wrote the commits in `blaze` all old checkouts became broken.

Also, I stand corrected on the sub-classing, at least for now since we're still transpiling `class ... extends` to ES5, which does not fully support subclassing natives (it ultimately fails one way or another).  It will be easier, eventually! :)

See [Babel's take on it](https://babeljs.io/learn-es2015/#ecmascript-2015-features-subclassable-built-ins).  I think it was removed sometime in Babel 5, and in Babel 6 requires [`babel-plugin-transform-builtin-extend`](https://www.npmjs.com/package/babel-plugin-transform-builtin-extend), something [we don't do by default](https://github.com/meteor/babel-preset-meteor/blob/master/index.js).  The [`loose` option](https://github.com/meteor/babel/blob/master/options.js#L16) may have something to do with it too.

But yea, you're right:

>  It is hard to sub-class an array in javascript.

Issue #8378, opened 2017-02-15
[Up for grabs] Reimplement minifier-js using Babili or Closure Compiler instead of UglifyJS


benjamn:
UglifyJS is the fastest JavaScript minifier, but it STILL does not officially support ECMAScript 2015+ features that are native in Node 4+ and most browsers: https://github.com/mishoo/UglifyJS2/issues/448

This leads to problems that only appear when you try to deploy your app or run `meteor run --production`), since that's when minification happens. To make matters worse, the parsing error messages produced by Uglify are [pretty bad](https://github.com/meteor/meteor/issues/8370#issue-207595783) (no filename, long and useless stack trace).

As more and more npm packages use features that UglifyJS can't handle, Meteor developers are exposed to this risk whenever the versions of any packages in `node_modules` change, even transitive dependencies whose versions the Meteor developer cannot control.

Here's the most recent issue caused by this regrettable situation: https://github.com/meteor/meteor/issues/8370

Others:
* https://github.com/meteor/meteor/issues/8020
* https://github.com/meteor/meteor/issues/7994
* https://github.com/meteor/meteor/issues/7479
* https://github.com/meteor/meteor/issues/7441
* https://github.com/meteor/meteor/issues/6405
* https://github.com/meteor/meteor/issues/5643
* et cetera, et cetera…

I previously [tried and failed](https://github.com/meteor/meteor/pull/7517) to update `minifier-js` to use the `harmony` branch of `uglify-js`, although @abernix has a [package](https://atmospherejs.com/abernix/minifier-js) that does something similar.

All in all, I think the time has come to choose another minifier.

The most compelling options seem to be:
* [Closure Compiler](https://developers.google.com/closure/compiler/), which is implemented in Java, though a JavaScript version (compiled from the Java) is now available.
* [Babili](https://github.com/babel/babili), by the same folks who brought you [Babel](http://babeljs.io/).

I don't have a strong preference between the two, so choosing the right one is definitely part of this project. Both are slower than UglifyJS, but in fairness they have more work to do, because they handle a much wider variety of input syntax.

If you want to get involved, have a look at [`minifier-js`](https://github.com/meteor/meteor/tree/devel/packages/minifier-js), and comment here. I'm not necessarily looking for a single champion; any kind of insight is helpful and welcome. I know @abernix and I will be happy to answer any questions you have.

sethmurphy18:
[...]

PR #8395, opened 2017-02-20
Modified standard-minifier-js to use Babili exported from meteor-babel

sethmurphy18:
This PR is to add support for the Babili minifier for the [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) package. This PR is in response to meteor/meteor#8378. A short overview of the changes:

* Incremented major versions of both [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) and [standard-minifer-js](https://github.com/meteor/meteor/tree/devel/packages/standard-minifier-js) since this is a major, API breaking change.
* Incremented minor version of [babel-compiler](https://github.com/meteor/meteor/tree/devel/packages/babel-compiler) for a major, but non API breaking change.
* Updated [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) to use Babili from meteor/babel#12.
* Removed `npm-shrinkwrap.json` from [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) as it not needed any longer.
* Removed tests for UglifyJS's beautify. Babili has no such function, at least to my knowlege.
* Updated [babel-compiler](https://github.com/meteor/meteor/tree/devel/packages/babel-compiler) to export [meteor-babel's](https://github.com/meteor/babel) minify function.
* Updated [babel-compiler](https://github.com/meteor/meteor/tree/devel/packages/babel-compiler) `npm-shrinkwrap.json` to use the new version of [meteor-babel](https://github.com/meteor/babel) once it is published. Also bring Babili into `npm-shrinkwrap.json`.

Issue #8020, opened 2016-11-05
Meteor build fails while minifying app code


yanickrochon:
First, I am not sure how to reproduce this. I have been programming this app for over six months and, now, I want to test deployment on a virtual machine. The app works fine in development mode, and I even have tests running to validate models, methods, etc.

Since ES2015 is natively supported, I import all my modules with `import` commands, define `class`es and arrow functions, etc. All my modules and dependencies are up-to-date, and I'm using meteor 1.4.2 (latest stable).

But when I run `meteor build ../output` from the root of my app, I get this output :

```
$ meteor build ../output
Errors prevented bundling:
While minifying app code:
eval at <anonymous>
(/home/yanick/.meteor/packages/standard-minifier-js/.1.2.1.s85ddv++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:28:1),
<anonymous>:1545:18: SyntaxError: Unexpected token: name (Converter)

at new JS_Parse_Error (eval at <anonymous>
(/home/yanick/.meteor/packages/standard-minifier-js/.1.2.1.s85ddv++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:28:1),
<anonymous>:1545:18)
at js_error (eval at <anonymous>
(/home/yanick/.meteor/packages/standard-minifier-js/.1.2.1.s85ddv++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:28:1),
<anonymous>:1553:11)
at croak (eval at <anonymous>
(/home/yanick/.meteor/packages/standard-minifier-js/.1.2.1.s85ddv++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:28:1),
<anonymous>:2089:9)
at token_error (eval at <anonymous>
(/home/yanick/.meteor/packages/standard-minifier-js/.1.2.1.s85ddv++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:28:1),
<anonymous>:2097:9)
at unexpected (eval at <anonymous>
[...]

PR #8397, opened 2017-02-20
Updated standard-minifier-js to use Babili from meteor-babel

sethmurphy18:
This PR is to add support for the Babili minifier for the [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) package. This PR is in response to meteor/meteor#8378. A short overview of the changes:

* Incremented major versions of both [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) and [standard-minifer-js](https://github.com/meteor/meteor/tree/devel/packages/standard-minifier-js) since this is a major, API breaking change.
* Incremented minor version of [babel-compiler](https://github.com/meteor/meteor/tree/devel/packages/babel-compiler) for a major, but non API breaking change.
* Updated [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) to use Babili from meteor/babel#12.
* Removed `npm-shrinkwrap.json` from [minifer-js](https://github.com/meteor/meteor/tree/devel/packages/minifier-js) as it not needed any longer.
* Removed tests for UglifyJS's beautify. Babili has no such function, at least to my knowlege.
* Updated [babel-compiler](https://github.com/meteor/meteor/tree/devel/packages/babel-compiler) to export [meteor-babel's](https://github.com/meteor/babel) minify function.
* Updated [babel-compiler](https://github.com/meteor/meteor/tree/devel/packages/babel-compiler) `npm-shrinkwrap.json` to use the new version of [meteor-babel](https://github.com/meteor/babel) once it is published. Also bring Babili into `npm-shrinkwrap.json`.

**NOTE:** The builds will fail until the new version of the [meteor-babel](https://github.com/meteor/babel) is published to NPM.


abernix:
Few things, which will hopefully help:

* You can continue using `class` (as you did in #8395), but you would need to add `ecmascript` to the `use` in `registerBuildPlugin` [just after this line](https://github.com/meteor/meteor/pull/8397/files#diff-465da66ef37dacf878bcffb77b43f378R6) (can't link to it directly).
* You can just push new commits to the same branch to update a PR without opening a new one (we can flatten/cleanup before merge)
* If you want to press forward with this non-`class` approach, it looks like you're just missing an `i` in `meteorBabelMinif***i***er` [here](https://github.com/meteor/meteor/pull/8397/files#diff-7133f7321890efabc8ea633e2312d9d4R5)

Also, you can enable CircleCI tests (which are not currently running, but should be) by:

* Going to [CircleCI Add Projects](https://circleci.com/add-projects)
* Clicking your GitHub username on the left
* Clicking on `meteor`'s "Build project" on the right.
* I'd also recommend raising the containers up to 4 (the free maximum for open-source projects) to speed things up a bit.

Hope these help! 😃

sethmurphy18:
Ok, so after some tinkering, Travis results are in for my repo. Build failed, but it was due to a package that I didn't modify. Circle will probably fail because the new version of `meteor-babel` has not been published yet. I should probably for testing update the shrinkwrap to use my git repo for CI purposes.

sethmurphy18:
Ok, @abernix or @benjamn or someone. I need some help. Travis build keeps failing because the tinytest for templating keeps failing. It is missing a single newline, see [here, line 1085](https://travis-ci.org/eagerestwolf/meteor/builds/204022076) for details. I don't think this is related to Babili, because the code isn't minified, but I am not really sure what is causing it to fail

sethmurphy18:
Also, Circle is giving me errors about Version 2.0.0 doesn't match version contraints, so I am not sure what to modify to correct that. It's only affecting `standard-minifier-js` although the versions on `minifier-js` and `babel-compiler` were changed as well.

sethmurphy18:
Also, sorry for all the commits. Once I have this working the way it is supposed to, I will squash it all into 1 commit.

sethmurphy18:
YES!! Circle Build is now passing. Now to figure out why Travis is failing.

abernix:
The new minifier is being used (and thus tested) because `minifier-js` is included as a package on the [`templating-tools`](https://github.com/meteor/blaze/tree/7c77ad019e91fce111695c15d19a3bd79f5ae4e5/packages/templating-tools) [`package.js`](https://github.com/meteor/blaze/blob/7c77ad019e91fce111695c15d19a3bd79f5ae4e5/packages/templating-tools/package.js#L19) (You'll note that `templating-tools` itself is actually part of the [Blaze repo](https://github.com/meteor/blaze) now but included as a submodule in `meteor`).

This looks like it's a legit test failure by what is a less-than-dynamic test.  This might actually be proof that the new minifier is doing more (good!) work since the output is more-minified (removing newlines is a reasonable step in any minification, right?).

The `templating-tools` tests start passing when [this change is applied to them](https://gist.github.com/abernix/9db10b9d617b8c6705eeeb71a87eee0b#file-diff-diff).  You'll see I'm just removing `\n` from a few places, all of which seem harmless and were not the intentions of the tests – they were simply dealing with the fact that they are testing minified code and those newlines were part of the result from the minifier.  We could change the tests on that Blaze package to be more flexible by, for example, changing them to a regex test which just requires whitespace (new-line or otherwise) via `\s`, but I'm not sure it's worth the effort.

I think we might need to coordinate a Blaze change with this in order to get it to pass tests.  For now, I'm okay with you adding `templating-tools` to `TEST_PACKAGES_EXCLUDE` (comma-delimited) [in `.travis.yml`](https://github.com/meteor/meteor/blob/907e1d91833eace7e40cc1d796cc7f9d171ceab1/.travis.yml#L8)

Also, you might be interested in knowing that you can run the tests locally for just a single package, with self-rerunning tests, by running:

    ./meteor test-packages templating-tools

...Might be able to avoid some extra-`git push`s. :wink:

sethmurphy18:
Ah okay, I didn't realize they were doing that as a result of the issues with Uglify. For the time being, I will update my script to exclude their package and I will open an issue over there...maybe a PR, depends on how ambitious I'm feeling. On another note, if that is the only minification that is happening, maybe I need to look into re-implenting Babili. I'm adapting [Babili's benchmark](https://github.com/babel/babili/blob/master/scripts/benchmark.js) to test [meteor-babel](meteor/babel) as well, so I can get an idea of what type of performance we can squeeze out of it. Also, and this is pretty important, the `npm-shrinkwrap.json` and `Npm.depends()` are using my git repo as the meteor-babel package, so all the tests will pass. This will obviously need updating once the new version is released.

abernix:
> I didn't realize they were doing that as a result of the issues with Uglify

I wouldn't call them issues.  That's just the way the test was implemented based on the result of the output from UglifyJS.  Also, `blaze` was up until very recently part of Meteor core and many of those packages were created by MDG or Meteor community.

> On another note, if that is the only minification that is happening, maybe I need to look into re-implenting Babili.

I'm not sure I follow you.  There are certainly other minifications happening, but most other libraries don't test Uglify-ed output.

> I'm adapting Babili's benchmark to test meteor-babel as well, so I can get an idea of what type of performance we can squeeze

Awesome!  Excited to see what the results look like!  I'm expecting them to be slower (reminder: these are only run once at `meteor build` time), but be processed more throughly.

> This will obviously need updating once the new version is released.

Yup!  No worries, we can help coordinate all of that (including Blaze) when the time comes.

sethmurphy18:
Well, I implemented the changes you suggested to their repo, and I am opening a PR. As for the minification comment, I thought they were testing a minified version of code, so that can be ignored.

sethmurphy18:
Just found another problem with Blaze. Their tests are using node 0.12, which is incompatible with Babili, so I need to update their circle.yml I guess.

sethmurphy18:
Ok, so craziest thing happened. I go to test their package, and the test failed because of the change. I'm not sure what is going on here. https://circleci.com/gh/eagerestwolf/blaze/2

sethmurphy18:
Ok, one more question, then I'm done for now, and I will leave you guys to work your magic, because I am kinda stumped on this one since I don't have direct MDG communication between the various groups. How do you exclude multiple packages? Space doesn't work and neither does a comma. `less` is already excluded.

sethmurphy18:
I found our problem. Digging further into the Blaze code reveals that there is now a complete incompatibility between Blaze and Meteor because of this change. Blaze is using the [UglifierJS beautify function](https://github.com/meteor/blaze/blob/80d3c4cfcc074697efa768d230462032b40b12fc/packages/spacebars-compiler/compiler.js#L97-L116), which Babili does not support, Babili is one way. So, instead of making their test pass, I think we should just have Blaze use uglify on their end as their own package.

sethmurphy18:
Ok, so I have a suggestion here. I have run some numbers and I know the whole point of this pull was to get rid of `uglify-js`, but seeing as Blaze depends on it so heavily and I cannot seem to get their repo to build correctly with a `Npm.depends()`, I have run some numbers. Using [cost-of-modules](https://npmjs.com/package/cost-of-modules), I found out that without `babel-preset-babili`, `meteor-babel` uses `~9.69 MB`, and with `babel-preset-babili`, `meteor-babel` uses `~10.07 MB`. That means, in total, by using `babel-preset-babili`, the size of the core is being reduced by `~1.09 MB` since `uglify-js` uses `~1.57 MB`. Since this code is being shipped on the server only and not to the client, why don't we just keep `uglify-js` in the `minifier-js` package since we are using babili in the `standard-minifier-js`. That would mean that the total size cost of meteor on the server would increase by `~0.38 MB` in total, but it keeps compatibiliy with Blaze, and in the future, they could look for solutions to either remove the beautification they are using, or another way to implement that, and then at that time, we could remove `uglify-js` completely, just an idea here, because as it is, I cannot get their repo to build with a `Npm.depends()`, and I cannot figure out why.

sethmurphy18:
Ok, I have a status update. [My pull](https://github.com/meteor/blaze/pull/234) was merged into [Blaze](https://github.com/meteor/blaze), I updated my reference locally for testing, but we have another roadblock that I am now working on tracking down. [Blaze](https://github.com/meteor/blaze/packages/blaze) is now failing it's tests, so I am trying to figure out why. This has morphed into quite the project, lol.

abernix:
@sethmurphy18 [These `blaze` tests](https://circleci.com/gh/meteor/blaze/tree/master) show as passing.  What test is failing?

sethmurphy18:
They are failing on the [meteor side still](https://travis-ci.org/eagerestwolf/meteor/builds/205243952). Around line 1616, you will see. 

```js
C: tinytest - blaze - render - video - autoplay : !!!!!!!!! FAIL !!!!!!!!!!!
{
  "groupPath": ["tinytest","blaze","render","video"],
  "test":"autoplay","events": [{
    "sequence":10139,
    "type": "fail",
    "details": {
      "type": "assert_equal",
      "expected": "false",
      "actual": "true",
      "not": false
    },
    "cookie": {
      "name": "blaze - render - video - autoplay",
      "offset": 1,
      "groupPath": ["tinytest","blaze","render","video"],
      "shortName": "autoplay"
    }
  }]
}

C: tinytest - blaze - render - video - controls : !!!!!!!!! FAIL !!!!!!!!!!!
{
  "groupPath": ["tinytest","blaze","render","video"],
  "test": "controls",
  "events": [{
    "sequence": 10142,
    "type": "fail",
    "details": {
      "type": "assert_equal",
      "expected": "false",
      "actual": "true",
      "not": false},
    "cookie": {
      "name": "blaze - render - video - controls",
      "offset": 1,
      "groupPath": ["tinytest","blaze","render","video"],
      "shortName":"controls"
    }
  }]
}

C: tinytest - blaze - render - video - loop : !!!!!!!!! FAIL !!!!!!!!!!!
{
  "groupPath": ["tinytest","blaze","render","video"],
  "test": "loop",
  "events": [{
    "sequence":10145,
    "type": "fail",
    "details": {
      "type": "assert_equal",
      "expected": "false",
      "actual": "true",
      "not": false
    },
    "cookie": {
      "name": "blaze - render - video - loop",
      "offset": 1,
      "groupPath": ["tinytest","blaze","render","video"],
      "shortName": "loop"
    }
  }]
}

C: tinytest - blaze - render - video - muted : !!!!!!!!! FAIL !!!!!!!!!!!
{
  "groupPath": ["tinytest","blaze","render","video"],
  "test": "muted",
  "events": [{
    "sequence": 10147,
    "type": "fail",
    "details": {
      "type": "assert_equal",
      "expected": "false",
      "not": false
    },
    "cookie": {
      "name": "blaze - render - video - muted",
      "offset": 0,
      "groupPath": ["tinytest","blaze","render","video"],
      "shortName": "muted"
    }
  }]
}
```

sethmurphy18:
Ok, an update on Blaze tests failing. They fail in console, but pass in browser. I wonder if it might be an issue with PhantomJS?

abernix:
They seem to be passing on CircleCI too.  See https://circleci.com/gh/abernix/meteor/121 which is the same commit (https://github.com/abernix/meteor/commit/39413d0428a3b1be9c3df4340586392702860bb9) where it ran successfully.  Maybe some difference in our CI configurations?  I believe we're using Ubuntu 12.x on ours if that makes a difference!

abernix:
Apologies, I failed to notice that failed build was on TravisCI and not CircleCI.  I believe our TravisCI uses BrowserStack to do some of the testing, something you wouldn't have permission to from your branch as it would require access to our credentials.  And obviously the `VIDEO` tests would be very likely to fail on the normal console.  I'll look into this!

sethmurphy18:
Ok sounds like a plan. I would be willing to help if needed. I'll help with everything really. Ideally, though, we should be getting close to being ready to merge these changes. Also, why is Github now telling me there is conflicts with the branch?

abernix:
I pushed #8414 today so you'll need to rebase off current `devel`!  There may be merge conflicts, but you can mostly pick yours over mine as mine is a stop-gap measure until we can merge this (which will hopefully be early next week!)

sethmurphy18:
While we are waiting on Blaze to do their stuff and Ben to return from vacation, I'm going to go through with ESLint and clean up my code. Also, and only for the sake of completeness. I am going to expose a settings prototype as a NOOP method, so the plugin can be extended as was discussed in the future.

abernix:
You'll need to rebase again because of another change I just made.

Additionally, you should not add the version pinning which you added with https://github.com/meteor/meteor/pull/8397/commits/10cfe7e47b57fc9e84f83f2b89744ca59deebf27; the `package.js` files in core should not have versions pinned as they will be pinned to each other automatically during publishing of a release.

Also, you're welcome to make ESLint changes, but honestly, your changes look fine already.  Be careful when all ESLint recommendations to ensure you don't add any dangling commas or anything which are not natively supported by Node as the `package.js` files are a rare case of a file that is read natively and not Babel'd.  The changes you've made are okay (on say arrays and objects), but check http://node.green/ if in doubt!  (For example, if you added a trailing comma on function arguments, you'd break it)

Also, the CI failure you're seeing right now is simply a CircleCI failure; just rebuild if you get that `unexpected` error on there.

sethmurphy18:
Ok, in terms of ESLint, I just wanted to ensure that every file ended with a newline, my indentation was consistent, my line endings were consistent, etc. I am bouncing between Windows and Linux for this project so consistency is key for me. I will remove the version pinning though, I was simply concerned with someone being able to install say version 2.0.0 of `standard-minifier-js` while keeping the old `minifier-js` but since everything is pinned during release, I'm good.

packages/babel-compiler/package.js
```diff
@@ -6,11 +6,11 @@ Package.describe({
   // isn't possible because you can't publish a non-recommended
   // release with package versions that don't have a pre-release
   // identifier at the end (eg, -dev)
-  version: '6.14.1'
+  version: '6.15.0'
 });
 
 Npm.depends({
-  'meteor-babel': '0.14.4'
```
benjamn: Version 0.16.0 is now available with the changes you implemented!

sethmurphy18: Sounds like a plan. I will update that as soon as I get done writing my node script for PS1 git status.

abernix:
In order to test/wrap-up this PR without the side-effects of the [newer HTML5 tests which Blaze includes](https://github.com/meteor/blaze/commit/04db0d40cef726cac73736629c164d404a39d54d) which are not compatible with PhantomJS (our current test browser for Meteor), I've created a special [`minifier-babili` branch](https://github.com/meteor/blaze/commits/minifier-babili) on the `blaze` repo and pointed pull request's `blaze` submodule to that.  It is an up-to-date branch of the `blaze` repo with the exclusion of the HTML5 tests.  See [this commit message](https://github.com/meteor/meteor/pull/8397/commits/34da752c7cc0a7e02713a8bd4f56dfc38b510aa0) for more information about why that change was necessary and why I consider it safe for the purposes of this PR.

Additionally, I rolled back some of the ES6 notation which was used in this PR and removed the `ecmascript` dependency within the compiler plugin (https://github.com/meteor/meteor/pull/8397/commits/735cb17903bfa5b8be3d69ec7767246a036ea4b8).  I realize this was contrary to my previous suggestion, but the commit explains a bit why.

I also re-introduced changes (https://github.com/meteor/meteor/pull/8397/commits/486ce486a7a97d5c67534165d48441af808c8eec) I made to the previous/current minification system in https://github.com/meteor/meteor/pull/8414.  Those changes will ensure the same, sane, error handling as that PR enabled.

As the Blaze changes (https://github.com/meteor/blaze/pull/236, https://github.com/meteor/blaze/pull/234) are merged and the [Meteor Babel PR](https://github.com/meteor/babel/pull/12) merged, I believe this completes this PR and fully removes external dependencies (i.e. @sethmurphy18's private forks of `meteor-babel` and `blaze`).

I can also confirm that this solves the reproduction provided by @klaussner [in #8020](https://github.com/meteor/meteor/issues/8020#issuecomment-264648632).

There may be some more nit-picks, but I wanted to deliver this update.  Thank you very much, @sethmurphy18 for your great work on this (sorry the testing across 3 repros didn't go as smoothly as one would have liked).  Unfortunately, the timing didn't quite work out to include this in Meteor 1.4.3.2 which is already packaged up and ready to go but this will get into an upcoming version of Meteor as soon as possible.

packages/minifier-js/package.js
```diff
@@ -1,26 +1,10 @@
...
-
-Npm.strip({
-  "uglify-js": ["test/"]
+  version: "2.0.0"
 });
 
 Package.onUse(function (api) {
-  api.export(['UglifyJSMinify', 'UglifyJS']);
+  api.use('babel-compiler');
+  api.export(['meteorBabelMinify']);
```
abernix: It may make sense to keep this as `BabelMinify` for consistency and since it's a global variable.  No need to change right now, I'm just leaving this as a note for another reviewer. :wink:

packages/standard-minifier-js/package.js
```diff
@@ -1,23 +1,20 @@
...
   summary: 'Standard javascript minifiers used with Meteor apps by default.',
-  documentation: 'README.md'
+  documentation: 'README.md',
 });
 
 Package.registerBuildPlugin({
   name: "minifyStdJS",
   use: [
-    'minifier-js'
+    'minifier-js@2.0.0',
```
abernix: It's possible that this should be removed before merging.  Need to confirm.

packages/standard-minifier-js/plugin/minify-js.js
```diff
@@ -1,14 +1,14 @@
 Plugin.registerMinifier({
-  extensions: ["js"],
-  archMatching: "web"
+  extensions: ['js'],
+  archMatching: 'web'
 }, function () {
-  var minifier = new UglifyJSMinifier();
+  var minifier = new meteorBabelMinifier();
```
abernix: Maybe `BabelMinifier`?

packages/standard-minifier-js/package.js
benjamn: After talking about this today, we're going to leave the constraint here for now, since it needs to be there when we publish the package, and shouldn't be hard to remember to update in the future (unlike some other package constraints, which would appear in too many different `package.js` files to keep track of).

benjamn:
Shall we merge this? I welcome your emoji reactions, positive or negative.

benjamn:
Thanks so much for making this happen @sethmurphy18! I think it's going to solve a lot of obscure production build problems for Meteor developers.

PR #8384, opened 2017-02-16
Facebook oauth appId || clientId

d62remi:
Refers to https://github.com/meteor/meteor/issues/8381#issuecomment-280285213

Issue #8381, opened 2017-02-16
[1.4.3.1] facebook-oauth (appId - clientId) question ?


d62remi:
Hi, 
Little question !

In facebook-oauth package : https://github.com/meteor/meteor/blob/devel/packages/facebook-oauth/facebook_client.js#L34

`'https://www.facebook.com/v2.2/dialog/oauth?client_id=' + config.appId +`

Why not used **clientId** instead of **appId** for iso functionality with google, twitter, ... ?

`'https://www.facebook.com/v2.2/dialog/oauth?client_id=' + config.clientId +` 

I searched for why google-oauth did not work with this example replacing facebook by google before I realized that I needed to use **clientId**
https://guide.meteor.com/security.html#api-keys-oauth
```javascript
ServiceConfiguration.configurations.upsert({
  service: "facebook"
}, {
  $set: {
    appId: Meteor.settings.facebook.appId,
    loginStyle: "popup",
    secret: Meteor.settings.facebook.secret
  }
});
```

abernix:
Thanks for bringing this up!  Yes, this is unfortunate.  [Facebook calls it `appId`](https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow#login), and (I'm assuming) we were trying to use terminology that was consistent with the individual oAuth service providers documentation and terminology, especially since different services offer _additional_ options (such as `login_style`, etc.) which are not consistently implemented.  Each provider should be presumed to have different parameter names.   It could be argued that the faux pas here was on Facebook for choosing to deviate from the oAuth standard "client_id".

At the very least, I think the documentation could make it clear that there are variations in settings.  At the very most, perhaps we could alleviate the pain for developers who make this mistake by supporting both: `config.appId || config.clientId` and then changing that guide example once it's done?

Thoughts?  Maybe submit a pull-request to fix it? 😃

hwillson:
Just to add - this has come up before (#2562), so adjusting the code to handle both `appId` and `clientId` sounds like the way to go.

abernix:
I'm tempted to change actually change the "supporting both" option since Twitter uses yet another option: `consumerKey`.  I noticed this when reviewing the documentation pull-request tied to this issue: https://github.com/meteor/docs/pull/125

Maybe this should just be a doc issue?  Having a third fallback seems odd.

d62remi:
Maybe used that :

```javascript
var config = ServiceConfiguration.configurations.findOne({service: 'twitter'});
if (!config) {
[...]

PR #8398, opened 2017-02-20
Improvement facebook-oauth with clientId

d62remi:
For better development with multi Oauth Login we could alleviate the pain for developers with **clientId** instead of **appId** for facebook-oauth package while maintaining the ability to use **appId**

Refers to : 
- PR https://github.com/meteor/meteor/pull/8384
- Issue https://github.com/meteor/meteor/issues/8381

### DIFF
--- a/packages/facebook-oauth/facebook_client.js
+++ b/packages/facebook-oauth/facebook_client.js
@@ -31,7 +31,7 @@ Facebook.requestCredential = function (options, credentialRequestCompleteCallbac
   var loginStyle = OAuth._loginStyle('facebook', config, options);
 
   var loginUrl =
-        'https://www.facebook.com/v2.2/dialog/oauth?client_id=' + config.appId +
+        'https://www.facebook.com/v2.2/dialog/oauth?client_id=' + (config.appId || config.clientId) +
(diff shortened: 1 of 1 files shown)


abernix:
After further consideration, I'm going to not merge this PR since a "fallback" or "catch-all" parameter name is not the right approach.  When only one provider was known to be different, it made a bit more sense as it was the edge case, but since Twitter also has a different `client_id` called `consumerKey`, I think we need to take a different approach.

Ultimately, the biggest failure here was the Meteor Documentation and I'm glad you opened meteor/docs#125 to address that.  We will review that change and get it merged.  Specifically, we should have made it more clear that the property names are different for each service.  I think it's quite reasonable to for each service to have different property names as each service also has different configuration parameters, as I've said before:

> specially since different services offer additional options (such as `login_style`, etc.) which are not consistently implemented. Each provider should be presumed to have different parameter names.

This means that the responsibility is still on the developer to read the docs to understand what parameters are different for each provider (e.g. Facebook, GitHub, Twitter, etc.), but the responsibility is on us to make sure those docs are right.

Regardless, thank you for bringing this up! :smile:

PR #8401, opened 2017-02-20
Complete fix for #8160, handle subclassed objects.  Added test

brucejo75:
@ManuelDeLeon, discovered a flaw with PR #8166 which is a fix for #8160. It did not work for sub-classes of an Array.

Changed all instances of checking the array to be seq instanceof Array || _.isArray(seq).

Passes all observe-sequence tests.  Added a test that verifies this works for subclassed Arrays.

### DIFF
--- a/packages/standard-minifier-css/package.js
+++ b/packages/standard-minifier-css/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'standard-minifier-css',
-  version: '1.3.3',
+  version: '1.3.4',
   summary: 'Standard css minifier used with Meteor apps by default.',
   documentation: 'README.md'
 });
@@ -8,7 +8,7 @@ Package.describe({
 Package.registerBuildPlugin({
   name: "minifyStdCSS",
   use: [
-    'minifier-css@1.2.14'
+    'minifier-css'
   ],
   npmDependencies: {
     "source-map": "0.5.6",


brucejo75:
Also, I want to create another test that will verify it works for arrays created in a [vm](https://www.npmjs.com/package/vm-browserify) because that was the reason for the fix in the first place.

To add this test I need to add a [vm-browserify](https://www.npmjs.com/package/vm-browserify) as a node_module for `observer_sequence_tests.js`.

If someone can point me in the direction to do that, I can get it done.  @abernix? @mitar?

packages/observe-sequence/observe_sequence_tests.js
```diff
@@ -93,6 +93,37 @@ runOneObserveSequenceTestCase = function (test, sequenceFunc,
...
+    if (arguments.length > 0) {
+      const items = this.slice.call(arguments, 0);
+      this.splice.apply(this, [0, 0].concat(this.slice.call(items)));
+    }
+  }
+
+  return ArraySubclass;
+
+})(Array);
+
```
abernix: I realize this sub-classing of a built-in is difficult, and it'll be ideal to remove eventually.  I'm not sure if it would be sufficient for this test, but the following might be more succinct and work?

```js
function ArraySubclass() {
  const Fp = Function.prototype;
  const inst = new (Fp.bind.apply(Array, [null].concat(Array.from(arguments))));
  inst.__proto__ = ArraySubclass.prototype;
  return inst;
}
ArraySubclass.prototype = Object.create(Array.prototype);
```

and then I believe the code below would be fine:

```js
var seq = new ArraySubclass(1, 1, 2);
```

brucejo75: Thanks @abernix,

Much simpler, unfortunately, the class you provided returns `true` to `_.isArray(seq)`.  We need:

```
var seq = new ArraySubclass(1, 1, 2);

seq instanceof ArraySubclass === true
seq instanceof Array === true
_.isArray(seq) === false
```

abernix: Have you tried using [`vm`](https://nodejs.org/dist/latest-v4.x/docs/api/vm.html) directly in the `tests.js`?  [`meteor-node-stubs`](https://github.com/meteor/node-stubs/) provides `vm-browserify` [itself](https://github.com/meteor/node-stubs/blob/618d67149f8c133e812bb927e9b21eb6a1f5daab/package.json#L34).

packages/observe-sequence/observe_sequence.js
```diff
@@ -126,7 +126,7 @@ ObserveSequence = {
   fetch: function (seq) {
     if (!seq) {
       return [];
-    } else if (_.isArray(seq)) {
+    } else if (seq instanceof Array || _.isArray(seq)) {
```
benjamn: Can you declare a helper function in this file called `isArray` that has this implementation?

brucejo75: Sure thing, will implement:

```
function isArray(arr) { return seq instanceof Array || _.isArray(arr)}
```
Code changed to:
```
 } else if (isArray(seq)) {
```

packages/observe-sequence/observe_sequence_tests.js
brucejo75: Thanks @abernix,  I just tried it.

`vm` is not a global
I cannot `require('vm')` - needs browserify package
I cannot `required('vm')` - defined in Utils package
I cannot `Npm.required('vm')` - ??
I cannot `import vm from 'vm';`
I cannot `import {vm} from 'vm';`

Do you know how a test file can load a module?  I do not see any `require`-like statements in the test files unless they are in `node_modules` installed under the `.npm` directory.

brucejo75: @abernix, @benjamn Any help?

brucejo75:
- updated with @benjamn's change request.
- updated tests to include sub classed Arrays

I would still like to add a test for an array created in a `vm`.  But I do not know how to include the [vm-browserify](https://www.npmjs.com/package/vm-browserify) package into the tests.

packages/observe-sequence/observe_sequence_tests.js
abernix: After further consideration, my recommendation wouldn't have worked since `meteor-node-stubs` aren't available from the tool tests, though you should be able to write a similar test in your own app and have it work – assuming you have `meteor-node-stubs` installed.

brucejo75:
Unfortunate.  All of the tests pass if `_.isArray(arr)` removed from the `isArray` function.  It is `vm` that needs '_.isArray(arr)`.

abernix:
Just as a sanity check for what might be a change that happens in the near future: does `Array.isArray` work for catching the `vm` `Array.prototype` as well, in the same way as Underscore's `_.isArray` does right now?

(I ask because Lodash is coming to Meteor and `_.isArray` is a [direct mapping to `Array.isArray`](https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L11300) in Lodash)

brucejo75:
`Array.isArray` works fine.  Also, I already checked lodash's implementation against vm created arrays.  So no issues with that.

brucejo75:
I was able to trim down the `vm-browserify` code enough that it made a reasonable function to include in the test file.  Now I have tests for a vm generated array.  To Summarize:

- Made simple code change to test arrays via `seq instanceof Array || _.isArray(seq)`.
- Verified that this will work if `_.isArray(seq)` is switched to lodash, which is `Array.isArray(seq)`
- Added a test for subclassed objects, reduced the `ArraySubClass` code size a little bit.
- Added a test for "vm generated arrays"

I ran the tests with all combinations of `seq instanceof Array` and `_.isArray(seq)` and verified the tests fail as expected.

Should be complete.

abernix:
Thanks for taking care of this, @brucejo75!  Looks good.

I made one more commit (eb458300e880863dc490d1ad0901f92f47e5d779) to fix some formatting before merging this however you didn't have a branch on your account aside from your `devel` branch so when I pushed to it, this PR was disassociated with it.  In the future, make sure you do development work on a branch within your project and not directly on the `devel` branch so merging works without us having to push to your main branch.

Alternatively, I guess you could disable "Allow contributors to make changes" when opening the PR, and then we'll just do it on our own branch instead of yours.  Typically though, if you have branched it wouldn't be a problem.

Thanks again!

brucejo75:
Thanks for the tip!  I was unaware of it.  I will try to improve next time.  Thanks!

abernix:
This should be fixed in Meteor 1.4.3.2.  You can try the latest 1.4.3.2 beta and help confirm by running:

brucejo75:
Verified works for me test, ArraySubclassed & VM test.  @ManuelDeLeon you might want to check it for your case too.

Issue #8402, opened 2017-02-21
Custom error handlers for exceptions at app download time


domq:
It is currently hard to change what happens when user-supplied code (such as a `WebApp.addHtmlAttributeHook` hook) throws an exception during rendering of the HTTP-served application (I think "boilerplate" is the right terminology). Basically, one can only choose between a plain stack trace typeset with faux `text/plain` (using &lt;br&gt;s, courtesy [finalhandler](https://www.npmjs.com/package/finalhandler)), or an even more spartan ["An error message"](https://github.com/meteor/meteor/blob/devel/packages/webapp/webapp_server.js#L661) text for production.

Making the error-catching features of [connect](https://www.npmjs.com/package/connect) accessible to packages / the main app would allow people to at least set up "traditional" (non-single-page) HTTP error handlers.

hwillson:
Closing as #8403 was merged - thanks!

PR #8403, opened 2017-02-21
Export "app"

domq:
As per the comment on line 653, this appears to be the best (only?) way to let users do things like set up a custom error page to catch app-rendering-time exceptions.

This purports to solve [#8402](https://github.com/meteor/meteor/issues/8402)

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -788,6 +788,7 @@ var runWebAppServer = function () {
     connectHandlers: packageAndAppHandlers,
     rawConnectHandlers: rawConnectHandlers,
     httpServer: httpServer,
+    connectMiddleware: app,
     // For testing.
     suppressConnectErrors: function () {
       suppressConnectErrors = true;
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -788,7 +788,7 @@ var runWebAppServer = function () {
     connectHandlers: packageAndAppHandlers,
     rawConnectHandlers: rawConnectHandlers,
     httpServer: httpServer,
-    connectMiddleware: app,
+    connectApp: app,
     // For testing.
     suppressConnectErrors: function () {
       suppressConnectErrors = true;


abernix:
Not entirely opposed to this but is this something you weren't able to accomplish using [`WebApp.connectHandlers.use`](https://docs.meteor.com/packages/webapp.html)?

A little more insight into what you've tried might be helpful here.

stubailo:
Yeah rendering exceptions should be able to be caught with an error middleware on `connectHandlers`.

glasser:
As the comment that @domq alluded to says, connect (at least the version Meteor uses) does not recursively search into nested middleware to search for 4-argument error middleware.

abernix:
Ok, so based on what @glasser is saying, and his own comment [here](https://github.com/domq/meteor/blob/0b4245fb95a20af9e618be5542e9868dacf9a861/packages/webapp/webapp_server.js#L652-L662) I can understand more now why this is necessary.  I hadn't remembered that was the case.

Given that, I'm in favor of this change, however I think we should consider a different name.  Currently the proposed API would be `WebApp.connectMiddleware`, however, this is specifically the "app" middleware, I believe.

Should we call it `WebApp.appMiddleware`?  Or just `WebApp.middleware` to avoid the redundancy?

hwillson:
How about `WebApp.connectApp`? That's what we're really exposing and `connect` themselves refer to the container that holds the defined middleware (the result of calling the `connect()` function) as a [connect "app"](https://github.com/senchalabs/connect#create-an-app). I realize we might not want to expose internal details in an externally available property name (in-case `connect` is removed at some point), but hey we're already doing that with `WebApp.connectHandlers` ... 🙂

abernix:
Any thoughts, @domq?

domq:
Well I'm honored that you would request my opinion on naming! I suppose I
can go with @HughWilson's suggestion or any other ; just let me know your
preference, and I'll rebase and resubmit.

-- 
Dominique Quatravaux
dominique@quatravaux.org


Le 8 mars 2017 2:49 PM, "Jesse Rosenberger" <notifications@github.com> a
écrit :

abernix:
Never hurts to ask for another opinion, thanks for the response!  And no worries, I've gone ahead and updated it (via GitHub Edit).

Thanks, @domq.

Issue #8195, opened 2016-12-27
Uncouple Cordova from Meteor releases


abernix:
This is a split of #8155 (which I'm doing to avoid it getting lost) and was discussed in [this forum post](https://forums.meteor.com/t/how-to-update-cordova-to-the-latest-6-4-0-in-meteor/31809/2) about updating Cordova to 6.4.0.

To quote the most relevant part of that forum post by Martijn:

> I think the main issue holding that back is that `cordova-lib` is currently [included in the dev bundle](https://github.com/meteor/meteor/blob/8a81e7156e07ca9e1a6c2b2b89870eb4abe72a2d/scripts/dev-bundle-tool-package.js#L57) and is relied on by CLI code in [`commands.js`](https://github.com/meteor/meteor/blob/8a81e7156e07ca9e1a6c2b2b89870eb4abe72a2d/tools/cli/commands.js) and [`commands-cordova.js`](https://github.com/meteor/meteor/blob/8a81e7156e07ca9e1a6c2b2b89870eb4abe72a2d/tools/cli/commands-cordova.js). So you probably want to add checks to see whether `cordova-lib` has been installed and show installation instructions if it hasn't.
>
> You may also want to have a look at [the CordovaProject class](https://github.com/meteor/meteor/blob/devel/tools/cordova/project.js) and the [platform/plugin pinning we currently do](https://github.com/meteor/meteor/blob/8a81e7156e07ca9e1a6c2b2b89870eb4abe72a2d/tools/cordova/project.js#L53-L77).

If anyone wants to help tackle this, please discuss here!

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

This feature request will be closed here, but anyone interested in migrating this feature request to the new repository (to make sure it stays active), can click [here](https://github.com/meteor/meteor-feature-requests/issues/new?title=Uncouple%20Cordova%20from%20Meteor%20releases&body=Migrated%20from:%20meteor/meteor%238195) to start the feature request migration process. This manual migration process is intended to help identify which of the older feature requests are still considered to be of value to the community. Thanks!

PR #8406, opened 2017-02-21
Update cordova-ios and cordova-android to latest versions

maxjgoldberg12:
Updates cordova-ios to `4.3.1` and cordova-android to `6.1.2` - latest versions as of Feb 21, 2017.

### DIFF
--- a/tools/cordova/index.js
+++ b/tools/cordova/index.js
@@ -11,8 +11,8 @@ export const CORDOVA_ARCH = "web.cordova";
 export const CORDOVA_PLATFORMS = ['ios', 'android'];
 
 export const CORDOVA_PLATFORM_VERSIONS = {
-  'android': '6.1.1',
-  'ios': '4.3.0'
+  'android': '6.1.2',
+  'ios': '4.3.1'
 };
 
 const PLATFORM_TO_DISPLAY_NAME_MAP = {


abernix:
Thanks for the pull-request.  Was there anything particular that might be relevant about these version bumps?

While these changes alone might be fine, changing them will require developers to post new versions of their apps to the app stores and users of apps to download new versions of the apps from the app stores (via their update processes), instead of benefiting from the usual hot reload features.  This is app store policy since the internal libraries themselves are being changed, not just `WebView` content.

I haven't looked at the changes themselves, but as these are patch bumps to the versions, it might make sense to wait until a more significant update is announced?  At the very least, other dependencies which would also require the publishing of an app should probably be bundled in, in a similar fashion as the 6.4.0 update, as seen in [this pull request](https://github.com/meteor/meteor/pull/8239/files).

We hope to avoid this process in the future by uncoupling Cordova versions from Meteor releases, allowing developers to update at-will, but that is being tracked in #8195.

abernix:
Thanks for submitting this initially – very much appreciated that you took the effort to look into this change.  However, I'll close this for the above reason and since we haven't heard anything else back from you.   We will revisit this upgrade in the future if we haven't uncoupled Cordova in a reasonable amount of time (#8195), and happy to reopen if you have information that would raise the urgency of this right now (for example, a severe regression is fixed, etc.).

Issue #8370, opened 2017-02-14
[1.4.2.6] Can no longer build or deploy


ffxsam:
I get build errors when I try to run `meteor deploy` (see at the very bottom, it's quite long). To troubleshoot, I tried `meteor reset` and that didn't help clear it up.

I might try to roll back to 1.4.2.3. Is there some official way to do that, or can I just edit `.meteor/release` and change it to 1.4.2.3 and then run meteor?

```
Errors prevented deploying:
While minifying app code:
eval at <anonymous>
(/Users/samh/.meteor/packages/standard-minifier-js/.1.2.2.52ezw9++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:1:0),
<anonymous>:1547:18: SyntaxError: Unexpected token: operator (>)
at new JS_Parse_Error (eval at <anonymous>
(/Users/samh/.meteor/packages/standard-minifier-js/.1.2.2.52ezw9++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:1:0),
<anonymous>:1547:18)
at js_error (eval at <anonymous>
(/Users/samh/.meteor/packages/standard-minifier-js/.1.2.2.52ezw9++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:1:0),
<anonymous>:1555:11)
at croak (eval at <anonymous>
(/Users/samh/.meteor/packages/standard-minifier-js/.1.2.2.52ezw9++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:1:0),
<anonymous>:2094:9)
at token_error (eval at <anonymous>
(/Users/samh/.meteor/packages/standard-minifier-js/.1.2.2.52ezw9++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/minifier-js/node_modules/uglify-js/tools/node.js:1:0),
<anonymous>:2102:9)
at unexpected (eval at <anonymous>
[...]

PR #8414, opened 2017-02-23
Improve minifiers to capture error from UglifyJS.

abernix:
This is a stop-gap until #8378 is complete.

The error messages which come from UglifyJS tend to be quite cryptic, as
seen in issues like meteor/meteor#8370 or meteor/meteor#8020.  The file,
line, and column are provided, however the message is garbled and the
stacktrace long and acutely harrowing.  Since these errors are occurring
on automatically concatenated files, even the line number is sometimes
not helpful.  Additionally, sourceMaps are not available in production
builds, intentionally.  (I wasn't able to access them from
`file.getSourceMap()` or `file.sourceMap` at all.)

In addition to actually providing the name of the encapsulating file,
which provides _some_visibility, this commit implements a parser around
the UglifyJS error which detects the error and location information of
the error, seeks to the line in the concatenated source, reads the
inline filename, and provides it in the output.

Crude, but incredibly helpful in diagnosing this problem until a better
solution is reached.

### DIFF
--- a/packages/standard-minifier-js/plugin/minify-js.js
+++ b/packages/standard-minifier-js/plugin/minify-js.js
@@ -94,8 +94,7 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
             // So in that case, 2 lines back is the file path.
             var parseErrorPath = contents[c - 2]
               .substring(3)
-              .replace(/\s+\/\//, "")
-            ;
+              .replace(/\s+\/\//, "");
 
             var minError = new Error(
               "UglifyJS minification error: \n\n" +
--- a/packages/standard-minifier-js/plugin/minify-js.js
+++ b/packages/standard-minifier-js/plugin/minify-js.js
@@ -121,7 +121,7 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
       var minified;
       try {
         minified = UglifyJSMinify(file.getContentsAsString(), minifyOptions);
-        if (! minified.code) {
+        if (!(minified && typeof minified.code === "string")) {
           throw new Error();
         }
       } catch (err) {
--- a/packages/standard-minifier-js/plugin/minify-js.js
+++ b/packages/standard-minifier-js/plugin/minify-js.js
@@ -32,13 +32,110 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
     }
   };
 
+  function maybeThrowMinifyErrorBySourceFile(error, file) {
+    var minifierErrorRegex = /\(line: (\d+), col: (\d+), pos: (\d+)\)/;
+    var parseError = minifierErrorRegex.exec(error.toString());
+
+    if (parseError) {
+      var lineErrorMessage = parseError[0];
+      var lineErrorLineNumber = parseError[1];
+
+      var parseErrorContentIndex = lineErrorLineNumber - 1;
+
+      // Unlikely, since we have a multi-line fixed header in this file.
+      if (parseErrorContentIndex < 0) {
+        return;
+      }
+
+      /*
+
+      What we're parsing looks like this:
+
+      /////////////////////////////////////////
+      //                                     //
+      // path/to/file.js                     //
+      //                                     //
+      /////////////////////////////////////////
+                                             // 1
+         var illegalECMAScript = true;       // 2
+                                             // 3
+      /////////////////////////////////////////
+
+      Btw, the above code is intentionally not newer ECMAScript so
+      we don't break ourselves.
+
+      */
+
+      var contents = file.getContentsAsString().split(/\n/);
+      var lineContent = contents[parseErrorContentIndex];
+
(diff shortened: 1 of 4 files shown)


packages/standard-minifier-js/plugin/minify-js.js
```diff
@@ -32,13 +32,110 @@ UglifyJSMinifier.prototype.processFilesForBundle = function (files, options) {
...
+        if (/^\/\/\/{6,}$/.test(sourceLine)) {
+
+          // If 4 lines back is the same exact line, we've found the framing.
+          if (contents[c - 4] === sourceLine) {
+
+            // So in that case, 2 lines back is the file path.
+            var parseErrorPath = contents[c - 2]
+              .substring(3)
+              .replace(/\s+\/\//, "")
+            ;
```
hwillson: Maybe move this semicolon up a line? Looks a bit dangly here.

hwillson:
Hi @abernix - Just one quick side note / gotcha I ran into with these changes. This is likely an edge case, and there is a workaround, but I'll outline the problem I ran into here (just in-case it comes up in the future).

I just tried these changes out against an app of mine that I was having minification problems with. This app uses a "packages for everything" type of structure, so something like:

```
/myapp
  package.json
  /packages
    /myapp-lib
    /myapp-core
    /myapp-utilities
    ...
```

This means the top level folder is really just a container, but I still run all `meteor` commands from within that top level folder. When Meteor builds from the top level folder, since it can't find a `package.js` (since the top level isn't a package) [it assumes it's an app](https://github.com/meteor/meteor/blob/331e6e2c98fbad520a40e3f23039ed57c6d486be/tools/isobuild/compiler-plugin.js#L1273). This means that during the build process, one of the files that is passed into the minification process looks like this:

```
JsFile
  _source: 
   File {
     sourcePath: undefined,
     info: 'resource /app.js',
     sourceMap: null,
     sourceMapRoot: null,
     targetPath: 'app/app.js',
     url: '/app.js',
     cacheable: false,
     nodeModulesDirectories: {},
     assets: null,
     _contents: <Buffer >,
     _hashOfContents: null,
     _hash: null },
  _arch: undefined,
  _minifiedFiles: [] }
```

Since this file doesn’t really exist, it’s contents after calling `UglifyJSMinify` on it look like:

```js
{ code: '', map: null }
```

In this specific case this is actually okay; the `code` value being an empty string is (sorta/kinda) valid, since the `app.js` file doesn't really exist. The old `UglifyJSMinify` code did this

```js
allJs += UglifyJSMinify(file.getContentsAsString(), minifyOptions).code;
```

so the empty `code` value was just appended, and it kept going. The new changes don't allow this however

```js
minified = UglifyJSMinify(file.getContentsAsString(), minifyOptions);
if (! minified.code) {
  throw new Error();
}
```

so the empty `code` value causes an error that looks like:

```
=> Errors prevented startup:                  
   
   While minifying app code:
   packages/minifyStdJS_plugin.js:147:17:  while minifying app/app.js
   at packages/minifyStdJS_plugin.js:147:17
   at Array.forEach (native)
   at UglifyJSMinifier.processFilesForBundle
   (packages/minifyStdJS_plugin.js:131:9)
```

The above being said, I don’t think this should impact your minification changes; the real issue is that the build process is passing an `app.js` file in for minification, when it doesn't exist. There is of course a workaround to all of this - by adding an empty `package.js` file in my top level project folder, the `app.js` file passed in then looks like:

```js
{ code: 'var require=meteorInstall({"package.js":function(){}},{extensions:[".js",".json",".html",".css"]});require("./package.js");',
  map: null }
```

and let's the minification process continue.  

Anyways, just a heads up in-case this comes up for anyone else (but now that we have ES2015 module support, the "packages for everything" approach is definitely less popular). Thanks!

abernix:
@hwillson Great catch.  Fixed with e87f1334ed279a02242be2e4de65b58a27bd4733 if you want to try it again!  I'm also curious if the error message you receive actually helps you find a problem! :smile:

packages/standard-minifier-js/plugin/minify-js.js
abernix: Yeah, it's super dangly.  In my personal coding-life, I enjoy leaving room to add new in-between methods and having a cleaner `diff` by turning it into a makeshift curly bracket (to show what it matches up with).  That being said, this is a good call on your part and I've fixed it (as this style exists no-where in the code right now).

hwillson:
Hi @abernix - I just tested https://github.com/meteor/meteor/commit/9eb9ede786d93b5e5da3dac297b331d4a493d8fa - looks great! These changes are definitely helpful, so LGTM!

abernix:
This is now officially released as `standard-minifier-js@1.2.3` and can be added by running the following on Meteor 1.4.x projects:

    meteor update standard-minifier-js

mwarren2:
@abernix Thank you for this work.
I've been fighting with a weird uglify message for more than a week.
I installed 1.2.3 and got a laser-sharp message indicating offending code and line no

Issue #8366, opened 2017-02-14
Add option to hide facebook-config-ui notice


jancurn:
After upgrading `accounts-facebook@1.0.11` to `accounts-facebook@1.1.0` I started receiving the following warning in the stderr:

```
W20170214-10:53:13.628(1)? (STDERR) Note: You're using accounts-ui and accounts-facebook,
W20170214-10:53:13.684(1)? (STDERR) but didn't install the configuration UI for the Facebook
W20170214-10:53:13.684(1)? (STDERR) OAuth. You can install it with:
W20170214-10:53:13.684(1)? (STDERR) 
W20170214-10:53:13.685(1)? (STDERR)     meteor add facebook-config-ui
W20170214-10:53:13.685(1)? (STDERR) 
```

It originates from file `meteor/packages/accounts-facebook/notice.js`. I realize this hint might be useful to developers who are just setting up a new Meteor app, but it's annoying for developers who already have their apps configured and running in production.  It would be great if there was an option to suppress the notice.

BTW this issue relates also to `github-config-ui`, `google-config-ui` and `twitter-config-ui` packages.

abernix:
You're absolutely right.  As an alternative to the configuration interface provided by the `<service>-config-ui` packages, are you using `service-configuration` (as a Meteor package)?  Perhaps we should not show the above message if `service-configuration` installed?

jancurn:
Yeah, we're using `service-configuration`, I think the solution you suggested would work well.

jancurn:
You guys rock, thanks a million!

hwillson:
PR #8429 has been merged, so this will be coming soon. Closing for now - thanks!

abernix:
This should be fixed in Meteor 1.4.3.2.  You can try the latest 1.4.3.2 beta and help confirm by running:

    meteor update --release 1.4.3.2-beta.0

Please report back if you encounter any problems, and thanks for reporting this originally!  (And thanks @hwillson for the fix!)

jancurn:
It works, the log messages are gone. Thanks again guys

PR #8429, opened 2017-02-27
Prevent accounts configuration UI warning from showing when using service-configuration.

hwillson:
Hi guys - this PR is intended to help address issue #8366. We shouldn't be showing the "you didn't install the accounts configuration UI" warning if an app is already using `service-configuration`. Thanks!


abernix:
Thanks, @hwillson! :wink:

PR #8437, opened 2017-03-01
Remove version constraint from minifier-css in standard-minifier-css.

abernix:
It appears that an older version constraint was left in place during the publishing of `standard-minifier-css` which prevented it from getting the fixes included in https://github.com/meteor/meteor/pull/8048.

Since packages which depend on compiler plugins have their compiler built into them, this has prevented the new version from being released as was planned in ~#7523~ #8048.

This was mentioned in https://github.com/meteor/meteor/issues/7523#issuecomment-280930819

/cc @jeanfredrik

### DIFF
--- a/packages/standard-minifier-css/package.js
+++ b/packages/standard-minifier-css/package.js
@@ -8,7 +8,7 @@ Package.describe({
 Package.registerBuildPlugin({
   name: "minifyStdCSS",
   use: [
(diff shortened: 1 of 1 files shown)


abernix:
This has been published as `standard-minifier-css@1.3.4` and should include the fix from #8048.

Issue #6120, opened 2016-01-30
Implement RFC 7239 for force-ssl


winniehell:
[RFC 7239](http://tools.ietf.org/html/rfc7239) replaces the `X-Forwarded-Proto` header by `Forwarded`. How about using the `Forwarded` header in the `force-ssl` package with a fallback to `X-Forwarded-Proto`?

winniehell:
See also Wikipedia on [X-Forwarded-For](https://en.wikipedia.org/wiki/X-Forwarded-For).

timbotnik:
Thanks for noting this @winniehell.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

This feature request will be closed here, but anyone interested in migrating this feature request to the new repository (to make sure it stays active), can click [here](https://github.com/meteor/meteor-feature-requests/issues/new?title=Implement%20RFC%207239%20for%20force-ssl&body=Migrated%20from:%20meteor/meteor%236120) to start the feature request migration process. This manual migration process is intended to help identify which of the older feature requests are still considered to be of value to the community. Thanks!

PR #8440, opened 2017-03-01
Adjusted force-ssl to work with the forwarded header (RFC 7239).

hwillson:
Hi all - this PR is intended to help address issue #6120 (updating `force-ssl` to work with the `forwarded` header (see [RFC 7239](https://tools.ietf.org/html/rfc7239)), falling back on `x-forwarded-for` and `x-forwarded-proto`). I know there has been some discussion lately about whether or not it makes sense to use `force-ssl`, but for now it's still supported / included in core (hence these updates).

A few quick notes about these changes:

- To wire this up, I've leveraged the [`forwarded-http`](https://www.npmjs.com/package/forwarded-http) npm package. This package will look for any `*forwarded*` headers, then combine them all into a nice and clean forwarded object. This simplifies the `force-ssl` code a fair bit.
- I've included tests for the new `isLocalConnection` and `isSslConnection` functions I've added, but because `force-ssl` is a `prodOnly` package, these tests can't be run as part of the `force-ssl` package. Since I couldn't find a way to have package tests run against `prodOnly` packages, I've decided to split the package into 2 packages. `force-ssl` stays as the main `prodOnly` package, then uses a new internal `force-ssl-common` package. The `force-ssl-common` package contains the helper header code I've added, along with tests. Let me know if you would like to approach this differently - this was the cleanest approach I could come up with to work around the `prodOnly` testing restriction.
- Since I was in the code, I went ahead and removed the `underscore` dependency (and have added in `ecmascript`).

Thanks all - awaiting feedback! :-)

### DIFF
--- a/packages/force-ssl-common/force_ssl_common.js
+++ b/packages/force-ssl-common/force_ssl_common.js
@@ -1,0 +1,29 @@
+import forwarded from 'forwarded-http';
+
+// Determine if the connection is only over localhost. Both we
+// received it on localhost, and all proxies involved received on
+// localhost (supports "forwarded" and "x-forwarded-for").
+const isLocalConnection = (req) => {
+  const localhostRegexp = /^\s*(127\.0\.0\.1|\[?::1\]?)\s*$/;
+  const request = Object.assign(req);
+  request.connection.remoteAddress =
+    request.connection.remoteAddress || req.socket.remoteAddress;
+  const forwardedParams = forwarded(request);
+  let isLocal = true;
+  Object.keys(forwardedParams.for).forEach((forKey) => {
+    if (!localhostRegexp.test(forKey)) {
+      isLocal = false;
+    }
+  });
+  return isLocal;
+};
+
+// Determine if the connection was over SSL at any point. Either we
+// received it as SSL, or a proxy did and translated it for us.
+const isSslConnection = (req) => {
+  const forwardedParams = forwarded(req);
+  return req.connection.pair
+      || forwardedParams.proto && forwardedParams.proto.indexOf('https') !== -1;
+};
+
+export { isLocalConnection, isSslConnection };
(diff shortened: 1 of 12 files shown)


packages/force-ssl-common/force_ssl_common.js
```diff
@@ -0,0 +1,29 @@
+import forwarded from 'forwarded-http';
+
+// Determine if the connection is only over localhost. Both we
+// received it on localhost, and all proxies involved received on
+// localhost (supports "forwarded" and "x-forwarded-for").
+const isLocalConnection = (req) => {
+  const localhostRegexp = /^\s*(127\.0\.0\.1|\[?::1\]?)\s*$/;
+  const request = Object.assign(req);
```
benjamn: Should this be `Object.create(req)`?

hwillson: Ooops - yes I meant to use `Object.create`, but I also meant to do a bit more than this is doing. This is working (since `Object.assign` just returns `req`), but I wanted to avoid modifying the original `http.IncomingMessage` object (stored in `req`), which this code is now ultimately doing. I'll commit changes to fix this. Thanks for flagging this @benjamn!

abernix:
@hwillson Not suggesting you change it, but would something similar to https://github.com/meteor/meteor/commit/f9de0a7b6535fe2a3129db0a886f6f9cae8d2930 work for forcing the production tests?

hwillson:
Great question @abernix - I did play around with something like this before submitting this PR, but I couldn't get this type of approach to work properly with `test-packages`. I didn't spend a huge amount of time investigating this approach though, since `force-ssl` is literally the only package in core that has the `prodOnly` flag set (so this issue isn't likely to come up much). Since `force-ssl` didn't have any existing tests, and the tests I've added are for utility functions, splitting those utility functions out into a non `prodOnly` package seemed like the easiest path forward. Thanks for suggesting this though - I'll definitely take a another look!

PR #8446, opened 2017-03-02
Shell: Don't be confused by SyntaxErrors thrown by command execution.

### LINKED ISSUE #8290: Shell treats SyntaxError thrown by user code as an unfinished command
When checking for a SyntaxError that indicates an unfinished command, the shell does not distinguish between SyntaxErrors from parsing the command and SyntaxErrors thrown by user code during the execution.  Example:

```
$ meteor create --bare .  # currently using METEOR@1.4.2.3
[...]
$ meteor npm install
[...]
$ meteor npm install --save babel-runtime
[...]
$ meteor
```

In another terminal:
```
$ meteor shell

> JSON.parse("[")
... 
```

~~Once this happens, typically the only way to recover is to interrupt and restart the shell.~~  This happened when I was debugging my application and was pretty confusing until I added a catch block to the shell command.  (Of course, the `JSON.parse("[")` was happening inside a lot of other code.)

mattmccutchen:
Fixes #8290.

Do we need a test for this?  I think it's pretty unlikely that someone will regress this unintentionally.

### DIFF
--- a/packages/shell-server/package.js
+++ b/packages/shell-server/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: "shell-server",
-  version: "0.2.2",
+  version: "0.2.3",
   summary: "Server-side component of the `meteor shell` command.",
   documentation: "README.md"
 });
--- a/packages/shell-server/shell-server.js
+++ b/packages/shell-server/shell-server.js
@@ -390,16 +390,12 @@ var evalCommandPromise = Promise.resolve();
 function evalCommand(command, context, filename, callback) {
   var repl = this;
 
-  function finish(error, result) {
-    if (error) {
-      if (repl._RecoverableError &&
-          isRecoverableError(error, repl)) {
-        callback(new repl._RecoverableError(error));
-      } else {
-        callback(error);
-      }
+  function wrapErrorIfRecoverable(error) {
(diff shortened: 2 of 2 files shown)


abernix:
This is awesome!  Thanks for looking into this, @mattmccutchen!

Just to provide insight to anyone peering in: Some of this `repl` custom `eval` functionality has been [properly exposed and documented](https://nodejs.org/dist/latest-v6.x/docs/api/repl.html#repl_recoverable_errors) in Node 6.x, so we'll need to change this again at that time, but this seems mostly on par with those new docs already (except for us capturing the `RecoverableError` constructor ourselves instead of using the `eval.Recoverable` which is now available).

Re: Tests.  I think all regressions are unintentional and I'd like to get tests for this eventually, but seeing at there is not currently a harness in place for the shell (an interactive test is necessary, perhaps) and this is actually an improvement to an existing deficiency, we might be willing to forgo a test for this.  That being said, I'll leave this for more  "LGTM"s as a precaution.

hwillson:
I agree with @abernix - since there isn't currently a defined automatic testing process for the shell, LGTM! Thanks @mattmccutchen!

benjamn:
Thanks @mattmccutchen (and @abernix)!

PR #8453, opened 2017-03-04
Bump mongodb driver to 2.2.24

### LINKED ISSUE #8449: Connection authentication spam against MongoDB arbiter nodes
The Meteor project is affected by a bug was re-introduced in the 2.2.x version of the MongoDB driver which attempts to authenticate against an arbiter node. Per https://jira.mongodb.org/browse/NODE-927, driver version 2.2.23 or greater contains a fix.

laosb:
Fixes #8449.

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -128,17 +128,12 @@ MongoConnection = function (url, options) {
   self._observeMultiplexers = {};
   self._onFailoverHook = new Hook;
 
-  var mongoOptions = _.extend({
-    db: { safe: true },
-    // http://mongodb.github.io/node-mongodb-native/2.2/api/Server.html
-    server: {
-      // Reconnect on error.
-      autoReconnect: true,
-      // Try to reconnect forever, instead of stopping after 30 tries (the
-      // default), with each attempt separated by 1000ms.
-      reconnectTries: Infinity
-    },
-    replSet: {}
+  var mongoOptions = Object.assign({
+    // Reconnect on error.
+    autoReconnect: true,
+    // Try to reconnect forever, instead of stopping after 30 tries (the
+    // default), with each attempt separated by 1000ms.
+    reconnectTries: Infinity
   }, Mongo._connectionOptions);
 
   // Disable the native parser by default, unless specifically enabled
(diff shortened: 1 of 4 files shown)


packages/mongo/package.js
```diff
@@ -21,7 +21,7 @@ Npm.strip({
 });
 
 Package.onUse(function (api) {
-  api.use('npm-mongo', 'server');
+  api.use('npm-mongo@2.2.24', 'server');
```
benjamn: This seems important because the new options format won't work with older versions of `npm-mongo`.

packages/mongo/mongo_driver.js
```diff
@@ -128,17 +128,12 @@ MongoConnection = function (url, options) {
   self._observeMultiplexers = {};
   self._onFailoverHook = new Hook;
 
-  var mongoOptions = _.extend({
-    db: { safe: true },
```
abernix: Took some work to track this down but I believe the omission of `db: { safe: true }` here is, well, safe.

`safe` was originally added to Meteor in https://github.com/meteor/meteor/commit/eee0f744d497bf24db8578a95aa71d9753123b5c which hints that it was a supported option by `mongodb@1.1.11` (the version [in use at the same time](https://github.com/meteor/meteor/blob/eee0f744d497bf24db8578a95aa71d9753123b5c/admin/generate-dev-bundle.sh#L86)).

The Node Driver [`HISTORY.md`](https://github.com/mongodb/node-mongodb-native/blob/2.2/HISTORY.md) doesn't really mention that `safe` was ever removed.  Suffice to say, it hasn't existed in the MongoClient `Db` documentation in a long, long time (see [1.4 version of the docs](https://mongodb.github.io/node-mongodb-native/api-generated/db.html)).

It is not listed in the [current version of the docs](https://mongodb.github.io/node-mongodb-native/2.2/api/MongoClient.html) but it does seem to be [somewhat-expected in the code](https://github.com/mongodb/node-mongodb-native/blob/5d2b36ca5f1b3f7ccf51750387568367df0a40c1/lib/url_parser.js#L391-L395) for URI connection strings, but [not documented](https://docs.mongodb.com/v3.2/reference/connection-string/) and the option should never make it to a server right now as it's being parsed in all three [Mongo Server](https://github.com/mongodb/node-mongodb-native/blob/791728ef40cc5b6a1a433dd79f1f26820a6259c2/lib/server.js#L46-L54), [Mongos](https://github.com/mongodb/node-mongodb-native/blob/791728ef40cc5b6a1a433dd79f1f26820a6259c2/lib/mongos.js#L49-L56) and [ReplSet](https://github.com/mongodb/node-mongodb-native/blob/791728ef40cc5b6a1a433dd79f1f26820a6259c2/lib/replset.js#L42-L49).

mitar: I think this should be replaced with `w` and `j`, no? Allowing one to set this could address: https://github.com/meteor/meteor-feature-requests/issues/17

PR #8474, opened 2017-03-10
[fix #8218] cleaning up reset tokens also cleans up enroll tokens

### LINKED ISSUE #8218: accounts-base: enroll tokens deleted too soon
(expanded from https://github.com/meteor/meteor/issues/7794#issuecomment-270253106 and filed as a new issue)

#7817 was intended to allow separate timeouts for enrollment tokens and password reset tokens, so enrollment tokens could last for (by default) 30 days, whereas password reset tokens could be made to expire more quickly (default 3 days).

That pull request appears to have a bug which causes enrollment tokens to be cleaned up under the reset token expiry rules. The net effect is that under the default settings, **enrollment tokens are destroyed after 3 days, rather than after 30 days**.  Regardless of what a developer sets `passwordEnrollTokenExpirationInDays` to in `Accounts.config()`, enrollment tokens will be removed from mongo after `passwordResetTokenExpirationInDays` days.

In particular, when processing password reset token expiry at https://github.com/meteor/meteor/pull/7817/files#diff-2ae5b6c36ab4132c1a2d0e33b5fd8443R1158 a `tokenFilter` with an `$or` specifying the appropriate token reason is passed to `expirePasswordToken`, but in https://github.com/meteor/meteor/pull/7817/files#diff-2ae5b6c36ab4132c1a2d0e33b5fd8443R1081 that `$or` clause is overwritten by the one which only looks at the when field. Perhaps the two clauses should be added to an `$and` clause instead?

Observed in Meteor 1.4.2.3 (accounts-base@1.2.14), but the bug I've identified is still in master as of this writing.

Reproduction:

* Enroll a new user with `Accounts.sendEnrollmentEmail()`
* Wait N days, with N between 3 and 30, or frob that user's enrollment token's `when` sufficiently far into the past to be between `passwordResetTokenExpirationInDays` and `passwordEnrollTokenExpirationInDays`, e.g. in `meteor mongo` shell run (with your new userId):

```
db.users.update({ _id: "p9ZJoQFecMFstfDuv"}, { $set: { "services.password.reset.when": new Date(new Date().getTime() - (4 * 24 * 60  * 60 * 1000)) }})
```

* Wait `EXPIRE_TOKENS_INTERVAL_MS` until the server runs the periodic cleanup function (or trigger it yourself from `meteor shell`; run `Accounts._expirePasswordResetTokens()`)
* Expected behavior: enrollment token is still present in user object, since enrollment tokens should be valid for 30 days
* Observed behavior: enrollment token has been deleted from user object

We're working around this by bumping the validity duration of password reset tokens, but that may not be a palatable option for everyone.

mutdmour:
The issue was with _.extend overriding two documents (a token filter by reason and a time filter) with the same key ($or). Put the two filters in an ($add) to solve the issue, and added two tests to ensure no interference between those two.

Fixes #8218

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1078,11 +1078,13 @@ Ap._generateStampedLoginToken = function () {
 function expirePasswordToken(accounts, oldestValidDate, tokenFilter, userId) {
   var userFilter = userId ? {_id: userId} : {};
 
-  accounts.users.update(_.extend(userFilter, tokenFilter, {
-    $or: [
-      { "services.password.reset.when": { $lt: oldestValidDate } },
-      { "services.password.reset.when": { $lt: +oldestValidDate } }
-    ]
+  accounts.users.update(_.extend(userFilter, 
+    {$and: [tokenFilter, {
+      $or: [
+        { "services.password.reset.when": { $lt: oldestValidDate } },
+        { "services.password.reset.when": { $lt: +oldestValidDate } }
+      ]
+    }]
   }), {
     $unset: {
       "services.password.reset": ""
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1076,16 +1076,16 @@ Ap._generateStampedLoginToken = function () {
 ///
 
 function expirePasswordToken(accounts, oldestValidDate, tokenFilter, userId) {
-  var userFilter = userId ? {_id: userId} : {};
+  const userFilter = userId ? {_id: userId} : {};
+  const resetRangeOr = { 
+    $or: [
+      { "services.password.reset.when": { $lt: oldestValidDate } },
+      { "services.password.reset.when": { $lt: +oldestValidDate } }
+    ]
+  };
+  const expireFilter = { $and: [tokenFilter, resetRangeOr] };
 
-  accounts.users.update(_.extend(userFilter, 
-    {$and: [tokenFilter, {
-      $or: [
-        { "services.password.reset.when": { $lt: oldestValidDate } },
-        { "services.password.reset.when": { $lt: +oldestValidDate } }
-      ]
-    }]
-  }), {
+  accounts.users.update({...userFilter, ...expireFilter}, {
     $unset: {
       "services.password.reset": ""
     }
--- a/packages/accounts-password/password_tests.js
+++ b/packages/accounts-password/password_tests.js
@@ -1474,7 +1474,7 @@ if (Meteor.isServer) (function () {
   );
 
   Tinytest.add(
-    'passwords - reset password doesn\t work if email changed after email sent',
+    "passwords - reset password doesn't work if email changed after email sent",
     function (test) {
       var username = Random.id();
       var email = username + '-intercept@example.com';
@@ -1688,31 +1688,31 @@ if (Meteor.isServer) (function () {
     function (test) {
       var email = test.id + '-intercept@example.com';
       var userId = Accounts.createUser({email: email, password: 'password'});
+      
       Accounts.sendEnrollmentEmail(userId, email);
       test.isTrue(!!Meteor.users.findOne(userId).services.password.reset);
 
       Accounts._expirePasswordEnrollTokens(new Date(), userId);
-
       test.isUndefined(Meteor.users.findOne(userId).services.password.reset);
     }
   )
 
   Tinytest.add(
-    'passwords - enroll tokens don\'t get cleaned up when reset tokens are cleaned up',
+    "passwords - enroll tokens don't get cleaned up when reset tokens are cleaned up",
     function (test) {
       var email = test.id + '-intercept@example.com';
       var userId = Accounts.createUser({email: email, password: 'password'});
+      
       Accounts.sendEnrollmentEmail(userId, email);
       test.isTrue(!!Meteor.users.findOne(userId).services.password.reset);
 
       Accounts._expirePasswordResetTokens(new Date(), userId);
-
       test.isTrue(Meteor.users.findOne(userId).services.password.reset);
     }
   )
 
   Tinytest.add(
-    'passwords - reset tokens don\'t get cleaned up when enroll tokens are cleaned up',
+    "passwords - reset tokens don't get cleaned up when enroll tokens are cleaned up",
     function (test) {
       var email = test.id + '-intercept@example.com';
       var userId = Accounts.createUser({email: email, password: 'password'});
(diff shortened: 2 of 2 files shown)


packages/accounts-base/accounts_server.js
```diff
@@ -1078,11 +1078,13 @@ Ap._generateStampedLoginToken = function () {
...
-      { "services.password.reset.when": { $lt: oldestValidDate } },
-      { "services.password.reset.when": { $lt: +oldestValidDate } }
-    ]
+  accounts.users.update(_.extend(userFilter, 
+    {$and: [tokenFilter, {
+      $or: [
+        { "services.password.reset.when": { $lt: oldestValidDate } },
+        { "services.password.reset.when": { $lt: +oldestValidDate } }
+      ]
+    }]
```
abernix: I realize the formatting here wasn't great to begin with, but this could still use some cleaning up - we should take this opportunity to fix it.

At the very least, with this complexity, the `$or` should become a variable:

```js
  const resetRangeOr = {
    $or: [
      "services.password.reset.when": { $lt: oldestValidDate } },
      "services.password.reset.when": { $lt: +oldestValidDate } },
    ],
  };
```

Then, we should fix the indentation and stop `extend`-ing into a variable name which doesn't make sense after it's changed (Since `_.extend` mutates the first parameter, and then it's no longer the `userFilter` once the `tokenFilter` and `resetRange` are in there):

```js
accounts.users.update(
  _.extend(
    {},
    userFilter,
    { $and: [ tokenFilter, resetRangeOr ] }
  )
);
```

Or instead of the last suggestion (though I realize this doesn't match the existing code, it's much more clear):

```js
  // resetRangeOr and userFilter are already defined above.
  const expireFilter = { $and: [tokenFilter, resetRangeOf] };

  accounts.users.update({ ...userFilter, ...expireFilter });
```

packages/accounts-password/password_tests.js
```diff
@@ -1697,6 +1697,34 @@ if (Meteor.isServer) (function () {
     }
   )
 
+  Tinytest.add(
+    'passwords - enroll tokens don\'t get cleaned up when reset tokens are cleaned up',
```
mutdmour: will fix it. There was another test that I based this on.
`  Tinytest.add(
    'passwords - reset password doesn\t work if email changed after email sent',`
Now I look closely and realize that was a typo. I can fix this as well.

packages/accounts-password/password_tests.js
```diff
@@ -1697,6 +1697,34 @@ if (Meteor.isServer) (function () {
...
+    function (test) {
+      var email = test.id + '-intercept@example.com';
+      var userId = Accounts.createUser({email: email, password: 'password'});
+      Accounts.sendEnrollmentEmail(userId, email);
+      test.isTrue(!!Meteor.users.findOne(userId).services.password.reset);
```
mutdmour: I did it mostly to stay consistent with the rest of the tests on token cleanup. I have looked into it and I don't see why use `!!`. `Meteor.users.findOne(userId).services.password.reset` returns an object or `undefined`. At first I thought it was to handle empty object case `{}`. Though both cases `test.isTrue({})` and `test.isTrue(!!{})` are true. Both cases `test.isTrue(undefined)` and `test.isTrue(!!undefined)` are false. I can remove the `!!`. What do you think?

packages/accounts-password/password_tests.js
```diff
@@ -1697,6 +1697,34 @@ if (Meteor.isServer) (function () {
...
+
+      Accounts._expirePasswordResetTokens(new Date(), userId);
+
+      test.isTrue(Meteor.users.findOne(userId).services.password.reset);
+    }
```
mutdmour: I can change that, though this was consistent with the other tests in the same file.

packages/accounts-base/accounts_server.js
mutdmour: question, what's this `...` notation? I have not seen it before and a hard one to get an answer for from google. why does it solve my "unknown operator" issue?

abernix: Great question, and hard to search for dots on Google. :smile:

This is `Object` spread notation, taking advantage of the [Spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator) but using a Stage-3 ECMAScript proposal [seen here](https://github.com/sebmarkbage/ecmascript-rest-spread).  Stage-3 proposals are very likely to become a part of the official specification and we have enabled in Meteor [via `babel-preset-meteor`](https://github.com/meteor/babel-preset-meteor/blob/42a3ad8484ea7d01c5d6b95aba9a102a715c8abc/index.js#L37).

packages/accounts-password/password_tests.js
abernix: I think we could do without the the coercion notation (`!!`) however I believe this test is trying to test that it is entirely unchanged (right?) so it should probably test for exactly the same thing (i.e. _with_ `!!`).

mutdmour: It's not to test if it's exactly the same thing. It's test for existence of the reset or enroll token object, which is added in the line before.

abernix: I mean that when line 1710 runs `Meteor.users.findOne(userId).services.password.reset`, it's meant to make sure that it's the same as when it ran on line 1706, right?

mutdmour: Yes exactly. Maybe it's better to test for equality then directly.
` var resetToken = Meteor.users.findOne(userId).services.password.reset;`
`test.isTrue(resetToken);`
`Accounts._expirePasswordResetTokens(new Date(), userId);`
`test.isEqual(resetToken, Meteor.users.findOne(userId).services.password.reset);`

abernix: Yeah, that sounds best to me.  Mind making the fix?  Then I think we're good to go.

mutdmour: Done

abernix:
I think your https://github.com/meteor/meteor/pull/8474/commits/00a2cb8d127e9b4796fd329b7f770cca2a9d1437 failed because of a problematic commit on the `devel` branch.  If you rebase off `devel` and try again it should work.  (And I think you can revert your attempt to fix it with https://github.com/meteor/meteor/pull/8474/commits/b6d687eb9ae8a2fc884b118991d2f6ce3909b7ca :smile:)

Sorry for the troubles!

mutdmour:
@abernix I think this is ready. Let me know if there's anything else needed. I have a fix for another issue ready but cannot create branches. Can you add me as a collaborator so I can create branches?

hwillson:
@mutdmour Since you're working off of your own fork, you can create as many branches as you want there. When you're ready to submit your PR, you can then do so from your branch. Creating branches within the Meteor project itself is very tightly controlled (to try to keep things nice and clean), and not necessary for submitting bug fix / feature request PR's.

mutdmour:
@hwillson Thanks ya, that makes sense. I thought I could not create branches on my fork either, but I just tried doing it locally first and that worked.

PR #8495, opened 2017-03-17
Upgrade to mailcomposer@4 and smtp-connection@2

### LINKED ISSUE #8425: Email package splits lines incorrectly / upgrade mailcomposer
Try this on Meteor server (without MAIL_URL set, so that it prints to stdout):

```javascript
Email.send({
  subject: "This is a very long email subject to illustrate how it gets split onto multiple lines"
})
```

The result is this:

```
I20170224-21:23:03.030(-5)? ====== BEGIN MAIL #0 ======
I20170224-21:23:03.030(-5)? (Mail not sent; to enable sending, set the MAIL_URL environment variable.)
I20170224-21:23:03.031(-5)? MIME-Version: 1.0
I20170224-21:23:03.031(-5)? Date: Sat, 25 Feb 2017 02:23:03 +0000
I20170224-21:23:03.031(-5)? Subject: This is a very long email subject to illustrate how it gets split
I20170224-21:23:03.031(-5)?         onto multiple lines
I20170224-21:23:03.032(-5)? Content-Type: text/plain; charset=utf-8
I20170224-21:23:03.032(-5)? Content-Transfer-Encoding: quoted-printable
I20170224-21:23:03.032(-5)?
I20170224-21:23:03.032(-5)?
I20170224-21:23:03.033(-5)?
I20170224-21:23:03.033(-5)? ====== END MAIL #0 ======
```

Unfortunately, this behavior seems incorrect.  Specifically, there are 8 spaces before "onto multiple lines", but there should be only one.  According to [RFC822](https://tools.ietf.org/html/rfc822), "Unfolding is accomplished by regarding   CRLF   immediately  followed  by  a  LWSP-char  as equivalent to the LWSP-char."  So while I only asked for one space before "onto" in the subject, I got 8 instead.

Specifically, this results in an actual error when setting MAIL_URL, producing actual email, and reading it in an email client.  At least in Pine, the subject gets eight spaces in the middle, which looks rather silly.  I don't know for sure that it will render incorrectly in all email clients, but it certainly isn't following the RFC822 spec.

I tried to reproduce this error in `node` using `mailcomposer`, but it seems to implement RFC822 correctly:

```javascript
var MailComposer = require('mailcomposer')
var mc = new MailComposer()
mc.addHeader({subject: "This is a very long email subject to illustrate how it gets split onto multiple lines"})
mc.buildHeaders()
```

produces:

```
'Content-Type: text/plain\r\nMessage-ID: <67ed00a7-8695-b5b7-261b-af9d3b033ed5@ibis>\r\nSubject: This is a very long email subject to illustrate how it gets split\r\n onto multiple lines\r\nDate: Sat, 25 Feb 2017 02:31:57 +0000\r\nMIME-Version: 1.0'
```

I couldn't get Mailcomposer 0.1.15 (as used by Meteor) to work at all in Node.  Presumably the bug got fixed between 0.1.15 and the latest version, which is 4.0.1 (!).  Could we please upgrade?

edemaine:
I wrote this patch to fix #8425, namely, to fix how long header lines get encoded.  Main discussion is there.

* Switch from mailcomposer 0.1.15 -> 4.0.1 (latest, still MIT license).
* Switch from simplesmtp (which only supports mailcomposer 0.1) to smtp-connection (which supports any mail composer).
* Use smtp-connection@2 (instead of latest) which shares nodemailer-shared codebase with mailcomposer 4.0.1.
* Add test for long header lines (the original bug being fixed here)
* Add extra test for HTML + text messages
* Document some extra options arguments supported by new mailcomposer.  (I didn't add [them all](https://github.com/nodemailer/mailcomposer/blob/v4.0.1/README.md#e-mail-message-fields) -- let me know if I should, or link to that in some way...)

### DIFF
--- a/packages/email/.npm/package/npm-shrinkwrap.json
+++ b/packages/email/.npm/package/npm-shrinkwrap.json
@@ -1,34 +1,79 @@
 {
   "dependencies": {
+    "addressparser": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/addressparser/-/addressparser-1.0.1.tgz",
+      "from": "addressparser@1.0.1"
+    },
+    "buildmail": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/buildmail/-/buildmail-4.0.1.tgz",
+      "from": "buildmail@4.0.1"
+    },
+    "httpntlm": {
+      "version": "1.6.1",
+      "resolved": "https://registry.npmjs.org/httpntlm/-/httpntlm-1.6.1.tgz",
+      "from": "httpntlm@1.6.1"
+    },
+    "httpreq": {
+      "version": "0.4.23",
+      "resolved": "https://registry.npmjs.org/httpreq/-/httpreq-0.4.23.tgz",
+      "from": "httpreq@>=0.4.22"
+    },
+    "iconv-lite": {
+      "version": "0.4.15",
+      "resolved": "https://registry.npmjs.org/iconv-lite/-/iconv-lite-0.4.15.tgz",
+      "from": "iconv-lite@0.4.15"
+    },
+    "libbase64": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/libbase64/-/libbase64-0.1.0.tgz",
+      "from": "libbase64@0.1.0"
+    },
+    "libmime": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/libmime/-/libmime-3.0.0.tgz",
+      "from": "libmime@3.0.0"
+    },
+    "libqp": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/libqp/-/libqp-1.1.0.tgz",
+      "from": "libqp@1.1.0"
+    },
     "mailcomposer": {
-      "version": "0.1.15",
-      "resolved": "https://registry.npmjs.org/mailcomposer/-/mailcomposer-0.1.15.tgz",
-      "from": "mailcomposer@0.1.15"
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/mailcomposer/-/mailcomposer-4.0.1.tgz",
+      "from": "mailcomposer@4.0.1"
     },
-    "mimelib-noiconv": {
-      "version": "0.1.9",
-      "resolved": "https://registry.npmjs.org/mimelib-noiconv/-/mimelib-noiconv-0.1.9.tgz",
-      "from": "mimelib-noiconv@*"
+    "nodemailer-fetch": {
+      "version": "1.6.0",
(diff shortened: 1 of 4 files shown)


edemaine:
Enabled, and tests executed.  (No, I didn't have a CircleCI account before, I don't think.  At least it asked for GitHub authentication for the first time, and sent me a welcome message after joining...)

abernix:
Thanks so much for working on this, @edemaine.  We've got this slated for 1.4.4 right now. :)

edemaine:
Awesome, thanks!  I look forward to putting it in my Meteor app.

abernix:
This should be available for testing in Meteor 1.4.4-rc.4.  Please try it out and provide feedback! :smile:

jbbr:
Unfortunately this caused me a lot of trouble as I got some random, endless loading situations because meteor methods containing `Email.send` calls never finished execution.

`Email.send` is a synchronous fiber call but in approximately 30% of the calls it never finishes nor throws an exception after the upgrade.

I'll try to reproduce this in a clean example repo.

For anyone else who might experience this issue: Downgrading & pinning email & accounts-password fixes the problem for now:
```
meteor add email@=1.1.18 accounts-password@=1.3.4
```

Anyway thanks for this PR.

abernix:
@jbbr Thanks for reporting.  Please do try to make a reproduction and report it in a new issue so it can be looked into!

edemaine:
Ouch, sorry to hear this.  That's definitely unacceptable.  I'll try to reproduce as well...

harrisonhunter:
I am seeing similar behavior when using 1.4.4.1 (albeit less frequently) that results in this error:
`Error: Unexpected Response: 421 Timeout. Try talking faster next time!` and failure to send

and it is resolved by downgrading.

edemaine:
Thanks for the details. My guess is that the new `smtp-connection` package is not auto-retrying the connection when it sits idle and disconnects, whereas (I'm guessing) the old package did so.  This will require some reworking...

abernix:
@jbbr any updates on trying to create a reproduction?  It sounds like it might also be important to include information on who your SMTP provider is in order to properly reproduce this.

jbbr:
@abernix sorry for the delay.
I just tried reproducing the problem in a clean meteor project. Right now everything seems to work fine.
SMTP providers are Mailtrap (dev) and Mailgun (production).

I experienced the problem with both and I assume it's not due to network or SMTP errors like @harrisonhunter is suggesting. Such errors usually trigger an exception and can be catched or are at least dumped into the console. In my case nothing happened at all. The Meteor-method call just kind of paused without a sign or error or timeout.

As I'm using async/await syntax and promises I might also have forgot a "catch" somewhere and missed the error (Errors in promise callbacks/async functions are silently ignored...). So I'll try again with the problematic project and open a new issue if I'm able to reproduce this.

Issue #8457, opened 2017-03-06
exception thrown in ddp-client/livedata_connection.js when noRetry flag is set.


n17r4m:
Issue is in 1.4.3.1 and remains in devel branch.

in packages/ddp-client/livedata_connection.js
at line 318: `currentMethodBlock.splice(i, 1);`

This removes a `methodInvoker` from `self._outstandingMethodBlocks[0]`.
Unfortunately, the `_each` loop is executing with a fixed number of loop iterations detected from initial entry, so an iteration will occur where `methodInvoker` is undefined. 

This causes an exception to be thrown, and prevents the rest of the `onReset` from running. This breaks reconnection when there are methods in queue with the noRetry flag set.

Suggested resolution: store an array of indicies to remove, and remove the `methodInvoker`s from `self._outstandingMethodBlocks[0]` after the _each loop is complete.

hwillson:
Hi @n17r4m - thanks for reporting this. Any chance you could provide a small reproduction demonstrating this issue (see [Reporting a bug in Meteor](https://github.com/meteor/meteor/blob/devel/Contributing.md#reporting-bug))? A small, runnable repro would definitely help with troubleshooting. Thanks!

n17r4m:
Just checking in on this issue. I'm not sure I would be able to provide a runnable example, as the problem is only manifested through a specific timing of disconnecting network connectivity, followed by calling a meteor method with the noRetry flag set, followed by reconnecting to the network.

However, it is plainly visible through looking at the code in packages/ddp-client/livedata_connection.js that the way the elements are being removed from self._outstandingMethodBlocks[0] is incorrect.

As the code currently stands, there is a mistake equivalent to https://gist.github.com/chad3814/2924672 line 45-50. It is just not a correct way to remove an item from an array while iterating over it.

Because I forgot to mention it explicitly in my original comment, the actual exception being thrown is
"Cannot read property 'sentMessage' from undefined" on line 304: `if (methodInvoker.sentMessage && methodInvoker.noRetry) {` . 

Because an element is being removed unsafely from the array, the _.each will perform an additional loop iteration beyond the array bounds and methodInvoker will be undefined.

hwillson:
[...]

PR #8496, opened 2017-03-18
Fixing thrown exception when 'noRetry' flag is set. #8457

ahmedhagii:
**Behaviour:**
As discussed in the issue #8457, an exception was being thrown whenever there's a queue of methods with 'noRetry' flag set and connection is dropped then re-established.

**Cause:**
It was due to a wrong way of deleting elements from an array while iterating over it leading to the access of out of bounds elements.

**Steps:**
1) I made the old test aware of the problem by sending 2 methods instead of 1 so the loop will continue on after deleting the first element.
2) The test was failing with the old code.
3) Instead of looping and using "splice" to delete elements, I used builtin "filter" to filter out methods not satisfying the condition while notifying those methods their invocation has failed.
4) Tests now pass.

**Notes:**
Actually I didn't fully grasp how the 'StubStream' is working or how are methods added to "Stream.sent" so if someone can quickly explain it or point me somewhere I can read, it will be really helpful.

This is my first PR to an open source project, so excuse me if I missed any trivial steps 😅 or if this is not in the requested format and would appreciate any feedback.

### DIFF
--- a/packages/ddp-client/livedata_connection_tests.js
+++ b/packages/ddp-client/livedata_connection_tests.js
@@ -819,34 +819,48 @@ if (Meteor.isClient) {
 
     startAndConnect(test, stream);
 
-    var methodCallbackFired = false;
-    var methodCallbackErrored = false;
+    var firstMethodCallbackFired = false;
+    var firsMethodCallbackErrored = false;
+    var secondMethodCallbackFired = false;
+    var secondMethodCallbackErrored = false;
+
     // call with noRetry true so that the method should fail to retry on reconnect.
     conn.apply('do_something', [], {noRetry: true}, function(error) {
-      methodCallbackFired = true;
+      firstMethodCallbackFired = true;
       // failure on reconnect should trigger an error.
       if (error && error.error === 'invocation-failed') {
-        methodCallbackErrored = true;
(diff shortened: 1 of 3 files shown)


packages/ddp-client/livedata_connection_tests.js
```diff
@@ -819,34 +819,48 @@ if (Meteor.isClient) {
 
     startAndConnect(test, stream);
 
-    var methodCallbackFired = false;
-    var methodCallbackErrored = false;
+    var firstMethodCallbackFired = false;
+    var firsMethodCallbackErrored = false;
```
hwillson: Looks like a typo here - `firsMethodCallbackErrored` should be `firstMethodCallbackErrored`.

ahmedhagii:
Thanks @hwillson for reviewing this. I have removed that file and fixed the typo.

benjamn:
Your tests all passed locally for me, so I manually squashed this and merged it to `devel` just now: https://github.com/meteor/meteor/commit/b59513540efe8d2f178eb32527ec31d8d06bf727. Thanks for diagnosing and fixing this issue!

ahmedhagii:
@benjamn Thanks Ben :)

Issue #7583, opened 2016-08-04
Support "Bash on Windows"


NitroBAY:
Hello, I updated W10 to anniversary edition and so I downloaded meteor 1.4.0.1 via linux command of your website (curl | sh), check that I've python, gcc as the C compiler (+ sudo apt upgrade/update) but either on old project or newly created projects with the brand new ubuntu shell I got with sudo meteor :
`Unexpected mongo exit code 100. Restarting.
Unexpected mongo exit code 100. Restarting.
Unexpected mongo exit code 100. Restarting.
Can't start Mongo server.
MongoDB had an unspecified uncaught exception.
This can be caused by MongoDB being unable to write to a local database.
Check that you have permissions to write to .meteor/local. MongoDB does
not support filesystems like NFS that do not allow file locking.`
if I ommit sudo I have an 
`=> Started proxy.
Unexpected mongo exit code 14. Restarting.
Unexpected mongo exit code 14. Restarting.
Unexpected mongo exit code 14. Restarting.
Unexpected mongo exit code 14. Restarting.
/home/NitroBAY/.meteor/packages/meteor-tool/.1.4.0-1.13j49n2++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/lib/node_modules/meteor-promise/promise_server.js:165
      throw error;
      ^
Error: EACCES: permission denied, mkdir '/home/NitroBAY/.meteor/packages/npm-bcrypt/.0.8.7.1yuta2i++os.linux.x86_64+web.browser+web.cordova/npm/node_modules/.temp-pwr33q'
    at Error (native)`

abernix:
I don't believe this has been tested at all yet and I'm not even remotely surprised that it doesn't work out of the box.

I would recommend using the traditional Windows installer for now if you are eager to use Meteor on Windows.  But by all means, please do try to help figure out what's wrong and submit a proposal for making it work.  I'm sure many Windows users would appreciate it, but I don't know if this is going to get a lot of attention right away.

My first question would be if you can setup your own Mongo server on that machine and then get Meteor to use it by specifying the environment variable `MONGO_URL=mongodb://localhost:27017/meteor`.  It's possible that the Mongo binary Meteor has bundled just isn't compatible with Windows 10 and won't start.

But again, this is only speculation and there is probably a lot more that could be wrong / needs to be looked into.

NitroBAY:
[...]

PR #8500, opened 2017-03-20
Be more understanding of Windows' filesystem limitations.

abernix:
Presently, the renaming of directories that are in-use will fail on Windows.  This is already compensated for when `process.platform` is
set to `win32`.  However, within [BashOnWindows](https://github.com/Microsoft/BashOnWindows)/WSL (Windows Subsystem for Linux), `process.platform` is equal to `linux`, though the underlying filesystem is still the same and subject to similar limitations.

Microsoft has stated that it is [unlikely that they will remove "`Microsoft`"](https://github.com/Microsoft/BashOnWindows/issues/423#issuecomment-221627364) from the `os.release()` value so this checks for that in a similar fashion to how we used to check for `process.platform === "win32"`.  Also, this change causes `EACCES` to be tolerated in the same was as `EPERM` was already.  (`EACCES` is another type of permission error)

I would point out that this change might contribute to breaking atomicity in `files.renameDirAlmostAtomically` since it uses `files.rename` but I'd argue that it was already jeopardized (on Windows and only on Windows) with the previous exception and this change simply provides consistent behavior across Windows implementations.

Probably worth mentioning that this seems to make Meteor work under BashOnWindows on the latest Windows Insider Build (15058) which I believe is a release candidate for [Windows Creators Update](https://www.microsoft.com/en-us/windows/upcoming-features) which is expected to come out later this month, but I'm hesitant to suggest this constitutes support for BashOnWindows (like #7583).

Hopefully issues like this are fixed by https://github.com/Microsoft/BashOnWindows/issues/1529 and https://github.com/Microsoft/BashOnWindows/issues/1420, but I'm not sure what the timeline on those getting resolved is.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 checkout:
   post:
-    - git submodule update --init
+    # https://discuss.circleci.com/t/git-submodule-url-isnt-playing-nice-with-the-cache/549/3
+    - git submodule sync
+    - git submodule update --init --recursive || (rm -fr .git/config .git/modules && git submodule deinit -f . && git submodule update --init --recursive)
 
 dependencies:
   pre:
--- a/tools/fs/files.js
+++ b/tools/fs/files.js
@@ -1664,7 +1664,7 @@ files.existsSync = function (path, callback) {
   return !! files.statOrNull(path);
 };
 
-if (process.platform === "win32") {
+if (files.isWindowsLikeFilesystem()) {
   var rename = files.rename;
 
   files.rename = function (from, to) {
@@ -1677,7 +1677,7 @@ if (process.platform === "win32") {
         rename(from, to);
         success = true;
       } catch (err) {
-        if (err.code !== 'EPERM') {
+        if (err.code !== 'EPERM' && err.code !== 'EACCES') {
           throw err;
         }
       }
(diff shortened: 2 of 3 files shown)


abernix:
Interestingly, (and frustratingly!) I had to commit https://github.com/meteor/meteor/pull/8500/commits/ccded04022031801105d54777071ac407703fc38 in order to bust the `git submodule` caching on this as CircleCI insisted on running the tests against the Git submodule of @sethmurphy18 (through no fault of his own), [seen in the CircleCI results here](https://circleci.com/gh/abernix/meteor/215).

(We had intentionally set the submodule to his repository during #8378 testing)

sethmurphy18:
Yeah, sorry about that. I forgot to clear the git submodule cache before recommiting. I probably should have disabled my Circle caching during testing as well...or just used a different branch. I can relate to the Windows limitations though because I use Windows myself and meteor is slow and sometimes doesn't work correctly, no fault of Meteor there, just Microsoft. Gotta love em.

abernix:
Force merged (despite the failing tests) which failed because:

1. The `dev_bundle` needing to be rebuilt on `devel` (already, thanks to #8491)
2. The submodule commit above which was necessary to include https://github.com/meteor/blaze/commit/0aed367b338fd1d6edd0aecc90cba2566a0f8361.

PR #8503, opened 2017-03-20
Ensure that `.bin` files maintain executable bits in Windows bundles.

abernix:
#Windows has no concept of the executable bit so it is not applied by the `fstream` `Reader` when building the tarball which is used in both `meteor build` and `meteor deploy`.  For Windows users, this causes important scripts (such as `node-pre-gyp`) to not be executable when the bundles are deployed to Unix platforms (such as Galaxy).

To avoid giving every file executable bits, this applies an executable bit to the file only if it has read permission (something Windows _is_ aware of) and if it is in a location that Node bin links are typically placed, the `/node_modules/.bin/` directories.

Additionally, to apply a slightly more conservative approach to the existing logic (which applied executable bits to all directories), this changes that logic to only apply the executable bit to readable directories, thus maintaining some degree of permissions control when building from a Windows system.  This is in line with other projects linked in the comments and issues below, such as [`node-pre-gyp`](https://github.com/mapbox/node-pre-gyp/blob/7a28f4b0f562ba4712722fefe4eeffb7b20fbf7a/lib/install.js#L71-L77) and other projects linked in [the `node-tar` issue](https://github.com/npm/node-tar/issues/7).

A symptom of this might be an error message such as:

```
> node-pre-gyp install --fallback-to-build

sh: 1: node-pre-gyp: Permission denied
```

### DIFF
--- a/tools/fs/files.js
+++ b/tools/fs/files.js
@@ -879,6 +879,14 @@ function tryExtractWithNpmTar(buffer, tempDir, options) {
   });
 }
 
+// In the same fashion as node-pre-gyp does, add the executable
+// bit but only if the read bit was present.  Same as:
+// https://github.com/mapbox/node-pre-gyp/blob/7a28f4b0f562ba4712722fefe4eeffb7b20fbf7a/lib/install.js#L71-L77
+// and others reported in: https://github.com/npm/node-tar/issues/7
+function addExecBitWhenReadBitPresent(fileMode) {
+  return fileMode |= (fileMode >>> 2) & 0o111;
+}
+
 // Tar-gzips a directory, returning a stream that can then be piped as
 // needed.  The tar archive will contain a top-level directory named
 // after dirPath.
@@ -886,6 +894,12 @@ files.createTarGzStream = function (dirPath, options) {
   var tar = require("tar");
   var fstream = require('fstream');
   var zlib = require("zlib");
+
+  // Create a segment of the file path which we will look for to
+  // identify exactly what we think is a "bin" file (that is, something
+  // which should be expected to work within the context of an
+  // 'npm run-script').
+  var binPathMatch = ["", "node_modules", ".bin", ""].join(path.sep);
 
   // Don't use `{ path: dirPath, type: 'Directory' }` as an argument to
   // fstream.Reader. This triggers a collection of odd behaviors in fstream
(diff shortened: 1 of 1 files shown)


tools/fs/files.js
```diff
@@ -926,8 +940,13 @@ files.createTarGzStream = function (dirPath, options) {
       // setting it in an 'entry' handler is the same strategy that npm
       // does, so we do that here too.
       if (entry.type === "Directory") {
-        entry.mode = (entry.mode || entry.props.mode) | 0o500;
-        entry.props.mode = entry.mode;
+        entry.props.mode = addExecBitWhenReadBitPresent(entry.props.mode);
```
benjamn: Why is the `| 0o500` no longer necessary? In other words, why not something like this:
```js
entry.props.mode = addExecBitWhenReadBitPresent(
  (entry.mode || entry.props.mode) | 0o500
);
```

abernix: First, I should point out that `entry.mode` is always `undefined` and never set, or used – maybe it was used in a previous version of `fstream`, but I can't find any (we use our [own fork](https://github.com/meteor/fstream/)), and this behavior matches the published `fstream` behavior.  The second parameter to the `filter` callback [is actually `props`](https://github.com/meteor/fstream/blob/cf4ea6c175355cec7bee38311e170d08c4078a5d/lib/reader.js#L180), so this would make sense in that scenario but this code doesn't do that, instead utilizing only the first argument which includes `props` as an attribute on `entry`.

For directories, I'd say that the `| 0o500` (e.g. `u+rx`) was never necessary and that something less would have accomplished the goal.  I'm happy to change that back, but my thinking was that it might be preferable to maintain some sense of permissions on Windows.  Essentially, if a directory was encountered that had the readable bit missing, we shouldn't simply add that read bit _and_ add the executable bit.  Presumably there would be some reason that `read` permission (something Windows does understand) was not set.  My train of thought was strongly influenced by that of the other PRs marked as closed and merged in [the `node-tar` issue](https://github.com/npm/node-tar/issues/7), for example, https://github.com/mapbox/node-pre-gyp/commit/7a28f4b0f562ba4712722fefe4eeffb7b20fbf7a and https://github.com/Microsoft/types-publisher/pull/165/commits/be208195326e2d6dab1f099a4989120f553251a5#diff-db71419dab88a2797059646c6a31a328R42.

PR #8439, opened 2017-03-01
Release 1.4.x.y

abernix:
This release branch and pull request serve to collect important changes that should be released in the next 1.4.x.y patch release (for example: 1.4.3.2, 1.4.3.3, 1.4.4).

The next major feature branch of Meteor will be 1.5, which is being tracked in #8327.

Such changes for this track include obvious bug fixes and other PRs, where "obvious" means the changes are either extremely safe, very important, or both.

If you think an issue deserves consideration for this release branch, please ask for it to be added to the https://github.com/meteor/meteor/milestone/32 milestone.

PR #8505, opened 2017-03-21
Change Windows "Generate Dev Bundle" script to use common globals.

abernix:
This change allows the Windows ["Generate Dev Bundle" script](https://github.com/meteor/meteor/blob/devel/scripts/generate-dev-bundle.ps1) to
automatically get the versions of major bundled versions from the same
script which the Unix script uses, thus preventing different versions of
Node.js, npm, Mongo, etc. from being built into different versions of
Meteor, such as what happened when I published Meteor 1.4.3.2 and failed
to notice the duplication of variables across different scripts. (Sorry!)

This behavior now works in a similar way as the [`BUNDLE_VERSION` variable](https://github.com/meteor/meteor/blob/devel/meteor#L3), which
is retrieved from the global `meteor` script except this change uses the
package versions from the `build-dev-bundle-common.sh` script in the
`scripts` directory.

Hopefully this will prevent a mistake like this in the future!

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -2,7 +2,7 @@
 
 # Meteor Roadmap
 
-**Up to date as of February 15, 2017**
+**Up to date as of March 17, 2017**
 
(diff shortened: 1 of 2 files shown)


abernix:
@benjamn I tested this locally on Windows 10 Pro but not with a Jenkins build.

benjamn:
[Merged](https://github.com/meteor/meteor/pull/8439/commits/da763c74fa4b2f8e253977723cc1912558d1a697), thanks!

PR #8516, opened 2017-03-24
Adds package rate-limit dependence on ecmascript

### LINKED ISSUE #8515: Using rate-limit inside another package fails while minifying app code on 1.4.3.2
We use `rate-limit` inside another package, and since Meteor 1.4.3.2 and `rate-limit` version 1.0.7, some platforms fail to build while minifying app code with the following error:

```
While minifying app code:
packages/minifyStdJS_plugin.js:113:28: UglifyJS minification error: 
SyntaxError: Unexpected token punc «(», expected punc «:» at
packages/rate-limit/rate-limit.js line 105
within packages/rate-limit.js (line: 135, col: 18, pos: 10314):
_executeCallback(reply, ruleInput) {
at maybeThrowMinifyErrorBySourceFile (packages/minifyStdJS_plugin.js:113:28)
at packages/minifyStdJS_plugin.js:145:9
at Array.forEach (native)
at UglifyJSMinifier.processFilesForBundle
(packages/minifyStdJS_plugin.js:130:9)
```

I believe the cause is the missing dependency on `ecmascript` since the syntax on line 105 may need babel parsing it?

https://github.com/meteor/meteor/blob/devel/packages/rate-limit/rate-limit.js#L105

engelgabriel:
Closes #8515

### DIFF
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=4.7.13
+BUNDLE_VERSION=4.7.19
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
--- a/packages/rate-limit/package.js
+++ b/packages/rate-limit/package.js
@@ -11,7 +11,6 @@ Package.describe({
 });
 
 Package.onUse(function(api) {
-  api.use('ecmascript');
   api.use('underscore');
   api.use('random');
   api.addFiles('rate-limit.js');
--- a/packages/rate-limit/package.js
+++ b/packages/rate-limit/package.js
@@ -11,6 +11,7 @@ Package.describe({
 });
 
 Package.onUse(function(api) {
+  api.use('ecmascript');
   api.use('underscore');
   api.use('random');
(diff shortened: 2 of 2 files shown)


abernix:
Thanks for submitting this PR.  Since the ES6 usage is so minimal in this file, without a more substantial overhaul, I think it'd be better if we just removed the `ecmascript` dependency to avoid the overhead of that package, especially if there is anyone who is using `rate-limit` in a project that doesn't use `ecmascript` (less and less likely these days, but still a possibility). 

While my commits I just pushed to this PR functionally revert your change, I left your original contribution in place.  Thank you!

@benjamn While this is a core package, I think I can release this as 1.0.8 to existing 1.4.3.x apps.  I've bumped the patch version accordingly and I'll do so with your sign-off.

engelgabriel:
Thanks @abernix, and I agree with our approach. 

BTW, we do use `ecmascript` in our project, so the problem does not only affects who is using rate-limit in a project that doesn't use `ecmascript`.

The `meteor` command works fine, the problem is when you try to run `meteor build`.

Issue #2139, opened 2014-05-12
[Accounts Password / Security] Provide configuration option for preventing user enumeration


lorensr:
While user enumeration is common in web apps, there are some for which the knowledge that a certain person has an account is private information. In order to prevent user enumeration with `accounts-password`, an account with a duplicate email must be able to be created. Currently that is not possible:

``` js
    var match = e.err.match(/^E11000 duplicate key error index: ([^ ]+)/);
    if (!match) throw e;
    if (match[1].indexOf('$emails.address') !== -1)
      throw new Meteor.Error(403, "Email already exists.");
```

Also, `Meteor.loginWithPassword` needs to check the password of all users with the given email.

PR #8520, opened 2017-03-25
Add ambiguous error messages option to Accounts config

dhrubins:
User account enumeration is a security concern for *some* people. In our bug bounty program @LegalRobot, we've had 17 different reports on user enumeration. It got annoying, so we started using this code in local forks of accounts-core and accounts-password but are now happy to share this with the community. This PR addresses some of the simpler attack vectors: different error messages returned from the server upon various login failures or pw reset failures. 

By default, we leave the existing behavior alone but provide a simple mechanism to enable this by setting the ```ambiguousErrorMessages``` key to ```true``` in Accounts.config on the server, like this:
```
Accounts.config({
  ambiguousErrorMessages: true
})
```
Enabling this option sends the same, rather ambiguous, error message to the client upon various login and reset failures. The specific text isn't really important, just that it can be identical across different failures. Perhaps as an extension of this, we could allow people to configure the message?

As @glasser and @lorensr discussed a few years ago in #2139, user enumeration is still possible via inference upon registration failure - the only way to prevent this is allowing duplicate accounts for the same email, but that opens up a whole mess of UX issues; and that can still be mitigated through rate limiting on those methods. So, this PR does not tackle the whole user enumeration problem, but at least we’re not being as explicit about every failure.

Unfortunately, we're not really able to write tests for this change since it seems that [server-side testing is not isolated](https://github.com/meteor/meteor/blob/devel/packages/accounts-password/password_tests_setup.js#L107-L118) (PR anyone?). We can either run the tests with this new option disabled (which is more common and the default) or enabled, but not both.

Looping in @guide for documentation changes, and @data since we're messing with some pretty important stuff in Meteor accounts.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -65,8 +65,9 @@ Accounts._checkPassword = function (user, password) {
 
   password = getPasswordString(password);
 
-  if (! bcryptCompare(password, user.services.password.bcrypt)) 
+  if (! bcryptCompare(password, user.services.password.bcrypt)) {
     result.error = handleError("Incorrect password", false);
+  }
 
   return result;
 };
@@ -216,8 +217,9 @@ var checkForCaseInsensitiveDuplicates = function (fieldName, displayName, fieldV
         (!ownUserId ||
         // Otherwise, check to see if there are multiple matches or a match
         // that is not us
-        (matchedUsers.length > 1 || matchedUsers[0]._id !== ownUserId))) 
-          handleError(displayName + " already exists.");
+        (matchedUsers.length > 1 || matchedUsers[0]._id !== ownUserId))) {
+      handleError(displayName + " already exists.");
+    }
   }
 };
 
@@ -268,12 +270,14 @@ Accounts.registerLoginHandler("password", function (options) {
 
 
   var user = Accounts._findUserByQuery(options.user);
-  if (!user) 
+  if (!user) {
     handleError("User not found");
-    
+  }
+
   if (!user.services || !user.services.password ||
-      !(user.services.password.bcrypt || user.services.password.srp)) 
+      !(user.services.password.bcrypt || user.services.password.srp)) {
     handleError("User has no password set");
+  }
 
   if (!user.services.password.bcrypt) {
     if (typeof options.password === "string") {
@@ -285,11 +289,12 @@ Accounts.registerLoginHandler("password", function (options) {
       var newVerifier = SRP.generateVerifier(options.password, {
         identity: verifier.identity, salt: verifier.salt});
 
-      if (verifier.verifier !== newVerifier.verifier) 
+      if (verifier.verifier !== newVerifier.verifier) {
         return {
           userId: Accounts._options.ambiguousErrorMessages ? null : user._id,
           error: handleError("Incorrect password", false)
-        }
+        };
+      }
 
       return {userId: user._id};
     } else {
@@ -323,8 +328,9 @@ Accounts.registerLoginHandler("password", function (options) {
 //
 // XXX COMPAT WITH 0.8.1.3
 Accounts.registerLoginHandler("password", function (options) {
-  if (!options.srp || !options.password)
+  if (!options.srp || !options.password) {
     return undefined; // don't handle
+  }
 
   check(options, {
     user: userQueryValidator,
@@ -333,16 +339,19 @@ Accounts.registerLoginHandler("password", function (options) {
   });
 
   var user = Accounts._findUserByQuery(options.user);
-  if (!user) 
+  if (!user) {
     handleError("User not found");
+  }
 
   // Check to see if another simultaneous login has already upgraded
   // the user record to bcrypt.
-  if (user.services && user.services.password && user.services.password.bcrypt)
+  if (user.services && user.services.password && user.services.password.bcrypt) {
     return checkPassword(user, options.password);
+  }
 
-  if (!(user.services && user.services.password && user.services.password.srp)) 
-    handleError("User has no password set"); 
+  if (!(user.services && user.services.password && user.services.password.srp)) {
+    handleError("User has no password set");
+  }
 
   var v1 = user.services.password.srp.verifier;
   var v2 = SRP.generateVerifier(
@@ -352,11 +361,12 @@ Accounts.registerLoginHandler("password", function (options) {
       salt: user.services.password.srp.salt
     }
   ).verifier;
-  if (v1 !== v2) 
+  if (v1 !== v2) {
     return {
       userId: Accounts._options.ambiguousErrorMessages ? null : user._id,
       error: handleError("Incorrect password", false)
-    }
+    };
+  }
 
   // Upgrade to bcrypt on successful login.
   var salted = hashPassword(options.password);
@@ -390,8 +400,9 @@ Accounts.setUsername = function (userId, newUsername) {
   check(newUsername, NonEmptyString);
 
   var user = Meteor.users.findOne(userId);
-  if (!user) 
+  if (!user) {
     handleError("User not found");
+  }
 
   var oldUsername = user.username;
 
@@ -430,16 +441,19 @@ Meteor.methods({changePassword: function (oldPassword, newPassword) {
   check(oldPassword, passwordValidator);
   check(newPassword, passwordValidator);
 
-  if (!this.userId)
+  if (!this.userId) {
     throw new Meteor.Error(401, "Must be logged in");
+  }
 
   var user = Meteor.users.findOne(this.userId);
-  if (!user) 
-    handleError("User not found");  
+  if (!user) {
+    handleError("User not found");
+  }
 
   if (!user.services || !user.services.password ||
-      (!user.services.password.bcrypt && !user.services.password.srp)) 
+      (!user.services.password.bcrypt && !user.services.password.srp)) {
     handleError("User has no password set");
+  }
 
   if (! user.services.password.bcrypt) {
     throw new Meteor.Error(400, "old password format", EJSON.stringify({
@@ -449,8 +463,9 @@ Meteor.methods({changePassword: function (oldPassword, newPassword) {
   }
 
   var result = checkPassword(user, oldPassword);
-  if (result.error)
+  if (result.error) {
     throw result.error;
+  }
 
   var hashed = hashPassword(newPassword);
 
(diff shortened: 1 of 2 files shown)


packages/accounts-password/password_server.js
```diff
@@ -66,7 +66,11 @@ Accounts._checkPassword = function (user, password) {
   password = getPasswordString(password);
 
   if (! bcryptCompare(password, user.services.password.bcrypt)) {
-    result.error = new Meteor.Error(403, "Incorrect password");
+    if(Accounts._options.ambiguousErrorMessages) {
```
hwillson: We should be able to reduce a lot of the error boilerplate code. Creating a re-usable error handling function like the following should help:

```js
const handleError = (msg, throwError = true) => {
  const error = new Meteor.Error(
    403, 
    Accounts._options.ambiguousErrorMessages
      ? "Login failure. Please check your login credentials."
      : msg
  );
  if (throwError) {
    throw error;
  }
  return error;
};
```

So in this case, the call would become:

```
result.error = handleError("Incorrect password", false);
```

packages/accounts-password/password_server.js
```diff
@@ -202,7 +206,11 @@ var checkForCaseInsensitiveDuplicates = function (fieldName, displayName, fieldV
         // Otherwise, check to see if there are multiple matches or a match
         // that is not us
         (matchedUsers.length > 1 || matchedUsers[0]._id !== ownUserId))) {
-      throw new Meteor.Error(403, displayName + " already exists.");
+      if(Accounts._options.ambiguousErrorMessages) {
+        throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+      } else {
+        throw new Meteor.Error(403, displayName + " already exists.");
+      }
```
hwillson: Using `handleError`, this call would become:

```js
handleError(displayName + " already exists.");
```

packages/accounts-password/password_server.js
```diff
@@ -254,12 +262,22 @@ Accounts.registerLoginHandler("password", function (options) {
...
   var user = Accounts._findUserByQuery(options.user);
-  if (!user)
-    throw new Meteor.Error(403, "User not found");
-
+  if (!user) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User not found");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User not found");
```

packages/accounts-password/password_server.js
```diff
@@ -254,12 +262,22 @@ Accounts.registerLoginHandler("password", function (options) {
...
+    
   if (!user.services || !user.services.password ||
-      !(user.services.password.bcrypt || user.services.password.srp))
-    throw new Meteor.Error(403, "User has no password set");
+      !(user.services.password.bcrypt || user.services.password.srp)) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User has no password set");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User has no password set");
```

packages/accounts-password/password_server.js
```diff
@@ -272,10 +290,16 @@ Accounts.registerLoginHandler("password", function (options) {
...
+        if(Accounts._options.ambiguousErrorMessages) {
+          return {
+            error: new Meteor.Error(403, "Login failure. Please check your username and password.")
+          }
+        } else {
+          return {
+            userId: user._id,
+            error: new Meteor.Error(403, "Incorrect password")
+          };
+        }
```
hwillson: Using `handleError`, this call would become:

```js
return {
  userId: Accounts._options.ambiguousErrorMessages ? null : user._id,
  error: handleError("Incorrect password", false)
}
```

packages/accounts-password/password_server.js
```diff
@@ -320,16 +344,27 @@ Accounts.registerLoginHandler("password", function (options) {
...
 
   var user = Accounts._findUserByQuery(options.user);
-  if (!user)
-    throw new Meteor.Error(403, "User not found");
+  if (!user) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User not found");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User not found");
```

packages/accounts-password/password_server.js
```diff
@@ -320,16 +344,27 @@ Accounts.registerLoginHandler("password", function (options) {
...
 
-  if (!(user.services && user.services.password && user.services.password.srp))
-    throw new Meteor.Error(403, "User has no password set");
+  if (!(user.services && user.services.password && user.services.password.srp)) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User has no password set");
+    }
+  }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User has no password set");
```

packages/accounts-password/password_server.js
```diff
@@ -339,11 +374,18 @@ Accounts.registerLoginHandler("password", function (options) {
...
+    if(Accounts._options.ambiguousErrorMessages) {
+      return {
+        error: new Meteor.Error(403, "Login failure. Please check your username and password.")
+      };
+    } else {
+      return {
+        userId: user._id,
+        error: new Meteor.Error(403, "Incorrect password")
+      };
+    }
```
hwillson: Using `handleError`, this call would become:

```js
return {
  userId: Accounts._options.ambiguousErrorMessages ? null : user._id,
  error: handleError("Incorrect password", false)
}
```

packages/accounts-password/password_server.js
```diff
@@ -377,8 +419,13 @@ Accounts.setUsername = function (userId, newUsername) {
...
 
   var user = Meteor.users.findOne(userId);
-  if (!user)
-    throw new Meteor.Error(403, "User not found");
+  if (!user) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User not found");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User not found");
```

packages/accounts-password/password_server.js
```diff
@@ -421,12 +468,22 @@ Meteor.methods({changePassword: function (oldPassword, newPassword) {
...
 
   var user = Meteor.users.findOne(this.userId);
-  if (!user)
-    throw new Meteor.Error(403, "User not found");
+  if (!user) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User not found");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User not found");
```

packages/accounts-password/password_server.js
```diff
@@ -421,12 +468,22 @@ Meteor.methods({changePassword: function (oldPassword, newPassword) {
...
 
   if (!user.services || !user.services.password ||
-      (!user.services.password.bcrypt && !user.services.password.srp))
-    throw new Meteor.Error(403, "User has no password set");
+      (!user.services.password.bcrypt && !user.services.password.srp)) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      throw new Meteor.Error(403, "Login failure. Please check your username and password.");
+    } else {
+      throw new Meteor.Error(403, "User has no password set");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User has no password set");
```

packages/accounts-password/password_server.js
```diff
@@ -505,8 +562,14 @@ Meteor.methods({forgotPassword: function (options) {
...
   var user = Accounts.findUserByEmail(options.email);
-  if (!user)
-    throw new Meteor.Error(403, "User not found");
+  if (!user) {
+    if(Accounts._options.ambiguousErrorMessages) {
+      // If ambiguousErrorMessages is set, we don't want to give away that the email does not exist, so just return as if everything is a-OK
+      return
+    } else {
+      throw new Meteor.Error(403, "User not found");
+    }
```
hwillson: Using `handleError`, this call would become:

```js
handleError("User not found");
```

packages/accounts-password/password_server.js
```diff
@@ -66,7 +66,11 @@ Accounts._checkPassword = function (user, password) {
   password = getPasswordString(password);
 
   if (! bcryptCompare(password, user.services.password.bcrypt)) {
-    result.error = new Meteor.Error(403, "Incorrect password");
+    if(Accounts._options.ambiguousErrorMessages) {
+      result.error = new Meteor.Error(403, "Login failure. Please check your username and password.");
```
hwillson: Since users don't have to login with a username (e.g. can use an email address), how about changing this to something like "Login failure. Please check your login credentials.".

hwillson: Just to add - I updated this in the common/sample `handleError` function I mentioned above.

dhrubins: Agreed - that makes for much cleaner code. Thanks!

packages/accounts-password/password_server.js
dhrubins: Agreed - that makes for much cleaner code. Thanks!

runlevelsix:
I appreciate the work put into making this feature a reality in Meteor 1.5. However, I think it may be worth noting that it does not really prevent username enumeration.

When an non-existent email address is entered in the "forgot password" form, a websocket error still indicates that the password reset failed with the error: `"Login failure. Please check your login credentials"`.

To be completely sure that username enumeration cannot happen, the error must not be returned to the client.

abernix:
@runlevelsix Thanks for bringing this up.  You're right that the fix applied here does not _truly_ avoid user enumeration (and that's stated in the OPs post).  Though, I think it's worth pointing out that even Google and Facebook tell you immediately if an account exists via "Forgot Password" or not.  There's a user experience issue to blocking it entirely, though it would be nice if it were fully configurable though.  Would you be willing to expand upon this PR?

image
image

Would you be willing

PR #8522, opened 2017-03-27
bump dev bundle; upgrade node

Augus-boy:
Conflicts:
	docs/client/concepts.html
	meteor
	scripts/generate-dev-bundle.sh
	tools/bundler.js
	tools/main.js
	tools/server/boot.js


abernix:
Thanks for trying to make a contribution to Meteor!

It would appear that you're trying to merge changes from an incredibly old version of the `meteor` repo.  You'll note that the current version of Node.js in use within Meteor is [actually 4.8.0](https://github.com/meteor/meteor/blob/devel/scripts/build-dev-bundle-common.sh#L9), a much newer version than the 0.10.x you're suggesting to update with this pull-request.

If you're encountering a problem, you may want to consider filing an issue first.  Someone would be happy to help!

PR #8529, opened 2017-03-28
Add the name of the field in the error

s7dhansh:
When Minimongo throws an error, most of the times, the field names are not present, making it a bit difficult to debug. Updated the error messages to include the field name for which it threw the error.

I did not create an issue for this as I think it was too trivial a change. If its not, please let me know and I will create one for discussion.

Thank you.

### DIFF
--- a/packages/minimongo/modify.js
+++ b/packages/minimongo/modify.js
@@ -185,20 +185,20 @@ var MODIFIERS = {
   $currentDate: function (target, field, arg) {
     if (typeof arg === "object" && arg.hasOwnProperty("$type")) {
        if (arg.$type !== "date") {
-         throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers");
+         throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers: " + field);
        }
     } else if (arg !== true) {
-      throw MinimongoError("Invalid $currentDate modifier");
+      throw MinimongoError("Invalid $currentDate modifier: " + field);
     }
     target[field] = new Date();
   },
   $min: function (target, field, arg) {
     if (typeof arg !== "number") {
-      throw MinimongoError("Modifier $min allowed for numbers only");
+      throw MinimongoError("Modifier $min allowed for numbers only: " + field);
     }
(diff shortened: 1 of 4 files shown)


packages/minimongo/modify.js
```diff
@@ -185,20 +185,20 @@ var MODIFIERS = {
   $currentDate: function (target, field, arg) {
     if (typeof arg === "object" && arg.hasOwnProperty("$type")) {
        if (arg.$type !== "date") {
-         throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers");
+         throw MinimongoError("Minimongo does currently only support the date type in $currentDate modifiers: " + field);
```
benjamn: Can we make the `MinimongoError` constructor take a second (optional) parameter, instead of concatenating the strings together here?

s7dhansh: @benjamn I thought of doing that. Actually, I thought of converting it into a Meteor Error format (code, reason, message, details) but then I thought it was too much work just for adding the field name. I still think so. :). IMO, adding one parameter does not make much sense, but conforming it to the `Meteor Error` class (with a new errorType `Mongo.Error`) definitely does. It will also help with displaying the errors as we will be able to use err.message for any kind of error. I believe that minimongo is the only place where we get a vanilla javascript `Error`.
If you have any ideas as to some other info that we can include in the error, then I will try to incorporate them altogether and raise a PR whenever I get some time.

abernix: @s7dhansh Making the more substantial change of confirming to the `Meteor.Error` format is probably worth discussing in an issue if you think it's a problem that it needs addressing.  It may be, but at first glance I'm really not sure if there is a particular reason for why there is a difference.

However, if we changed the `MinimongoError` constructor to take an optional second parameter as @benjamn suggested (perhaps in as an `options` object if you think there is room for expansion), another pull-request after this could easily implement your suggestion by making a change in only one place instead of 29, since all 29 of these lines will need to be changed again. 😉

s7dhansh:
Great @abernix @benjamn! Thank you @abernix for modifying it.

Issue #8253, opened 2017-01-13
[Cordova] Google will disallow OAuth login in embedded WebView in April 2017


realyze:
See https://developers.googleblog.com/2016/08/modernizing-oauth-interactions-in-native-apps.html .

Quote:
```
On April 20, 2017, we will start blocking OAuth requests using web-views for all OAuth clients on platforms where viable alternatives exist.
```

What's the plan to deal with this please? Current Meteor Accounts seem always use the `popup` login style overriding `redirect`.

stubailo:
Presumably redirect will have the same problem right?

realyze:
Right, correct (I didn't stop to think that redirect doesn't mean redirect outside of webview :-) ).

So what do we do? There's https://github.com/EddyVerbruggen/cordova-plugin-googleplus but I haven't investigated how to use it with Accounts yet.

stubailo:
I think the first step would be to find out if the Cordova community has a solution for this -- how do people usually do OAuth there, is it through these various plugins?

If there are adequate solutions it might be as simple as just passing the data from that into a login method. What Meteor is doing is in no way unique, but from what I understand there isn't a public API that will let you easily integrate with such a plugin.

realyze:
Here's a few GH issues in various projects:
 * https://github.com/nraboy/ng-cordova-oauth/issues/283
 * https://github.com/Azure/azure-mobile-apps-ios-client/issues/90

ramezrafla:
Seems the community is still looking for an answer. Google likely wants to make sure to install super cookies in user browser to push their ads. However, android devices' stock browsers are not all the same, how can google know it's a webview. Can it not be as easy as altering the useragent of the oauth request? Is that possible? Or at compile time alter the user agent of Crossalk

realyze:
Soo, I don't want to be pushy, but is anyone from MDG looking into this? When Google flips the switch in April, all Meteor Cordova apps that use Google OAuth will break.

Kinda sounds like a big deal.

ramezrafla:
Agreed -- this could be a major issue in April. @martijnwalraven sorry to pull you into this. Your opinion would be appreciated (including maybe just setting the useragent, how does Google know if it's a webview except from useragent).

I expect a backlash from many mobile developers on this major change.

martijnwalraven:
[...]

PR #8549, opened 2017-03-31
[Work in Progress] Support Google Sign-In in google-oauth package.

benjamn:
Addresses #8253.

This implementation makes heavy use of https://www.npmjs.com/package/cordova-plugin-googleplus and takes loose inspiration from https://github.com/sujith3g/meteor-cordova-google-plus.

I have not yet figured out exactly how we should bump the version numbers, or whether a new `google-sign-in` package should be created (instead of modifying `google-oauth`).

cc @ramezrafla @realyze @stubailo

### DIFF
--- a/packages/google-oauth/google_server.js
+++ b/packages/google-oauth/google_server.js
@@ -1,12 +1,38 @@
-Google = {};
+var Google = require("./namespace.js");
+var Accounts = require("meteor/accounts-base").Accounts;
 
 // https://developers.google.com/accounts/docs/OAuth2Login#userinfocall
 Google.whitelistedFields = ['id', 'email', 'verified_email', 'name', 'given_name',
                    'family_name', 'picture', 'locale', 'timezone', 'gender'];
 
+Accounts.registerLoginHandler(function (request) {
+  if (request.googleSignIn !== true) {
+    return;
+  }
+
+  var res = HTTP.get(
+    "https://www.googleapis.com/oauth2/v3/tokeninfo",
+    { headers: { "User-Agent": "Meteor/1.0" },
+      params: { id_token: request.idToken }}
+  );
+
+  if (res.error) {
+    throw res.error;
+  }
+
+  if (res.statusCode === 200 &&
+      res.data.sub === request.userId) {
+    return Accounts.updateOrCreateUserFromExternalService("google", {
+      id: request.userId,
+      idToken: request.idToken,
+      accessToken: request.accessToken,
+      email: request.email,
+      picture: request.imageUrl
+    });
+  }
+});
 
 OAuth.registerService('google', 2, null, function(query) {
-
   var response = getTokens(query);
   var expiresAt = (+new Date) + (1000 * parseInt(response.expiresIn, 10));
   var accessToken = response.accessToken;
--- a/packages/google-oauth/google_sign-in.js
+++ b/packages/google-oauth/google_sign-in.js
@@ -1,0 +1,84 @@
+var Google = require("./namespace.js");
+
+var gplusPromise = new Promise(function (resolve, reject) {
+  if (! Meteor.isCordova) {
+    reject(new Error("plugins.googleplus requires Cordova"));
+    return;
+  }
+
+  document.addEventListener("deviceready", function () {
+    var plugins = global.plugins;
+    var gplus = plugins && plugins.googleplus;
+    if (gplus) {
+      resolve(gplus);
+    } else {
+      reject(new Error("plugins.googleplus not defined"));
+    }
+  }, false);
+});
+
+function tolerateUnhandledRejection() {}
+gplusPromise.catch(tolerateUnhandledRejection);
+
+// After 20 April 2017, Google OAuth login will no longer work from a
+// WebView, so Cordova apps must use Google Sign-In instead.
+// https://github.com/meteor/meteor/issues/8253
+exports.signIn = Google.signIn = function (options, callback) {
+  // support a callback without options
+  if (! callback && typeof options === "function") {
+    callback = options;
+    options = null;
+  }
+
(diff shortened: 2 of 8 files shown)


benjamn:
@abernix Let's try to de-risk these changes enough so that we can ship them as backwards-compatible patch updates to `accounts-google` and `google-oauth`. Doesn't have to be part of 1.4.4 (unless we bump minor versions) but that would also be nice.

packages/google-oauth/google_sign-in.js
```diff
@@ -0,0 +1,84 @@
...
+
+  }).catch(callback);
+};
+
+function getScopes(options) {
+  // we need the email scope to get user id from google.
+  var requiredScopes = ['email'];
+  var scopes = ['profile'];
+  if (options && options.requestPermissions) {
+    scopes = options.requestPermissions;
```
abernix: This isn't the fault of this PR, as this logic existed [before in `google_client.js`](https://github.com/meteor/meteor/pull/8549/files#diff-d3e3aa5afbdb786a7d4e241986164ee6R30), but it's super unclear to me what the difference is between `scope`, `permissions`, and `fields` (and I guess Google [refers to](https://developers.google.com/identity/protocols/OpenIDConnect#userinfocall) the resulting fields as `claims`?)

I digress, is this supposed to be getting checked against [`Google.whitelistedFields`](https://github.com/meteor/meteor/blob/devel/packages/google-oauth/google_server.js#L4)?

benjamn:
@abernix does that mean we need to update the `google` package too?

packages/google-oauth/google_server.js
```diff
@@ -1,12 +1,38 @@
...
+  }
 
+  if (res.statusCode === 200 &&
+      res.data.sub === request.userId) {
+    return Accounts.updateOrCreateUserFromExternalService("google", {
+      id: request.userId,
+      idToken: request.idToken,
+      accessToken: request.accessToken,
+      email: request.email,
+      picture: request.imageUrl
```
benjamn: I'm a little worried these fields won't be exactly the same as before.

realyze:
@benjamn Thanks again for looking into this!

I wanted to test it but I can't seem to get `accounts-google@1.1.1` to install in my meteor@1.3 app. It install alright in meteor@1.4 but our app is running @1.3 and we don't want to update yet.

```
realyze@:/tmp/test/googletest $ meteor add accounts-google@1.1.1
 => Errors while adding packages:             
                                              
error: No version of accounts-google satisfies all constraints: @1.1.1, @=1.0.10
Constraints on package "accounts-google":
* accounts-google@1.1.1 <- top level
* accounts-google@=1.0.10 <- top level
```

Do you have any thought on how to get it to run with @1.3?

Cheers!

abernix:
@realyze Just to be clear, the `accounts-google@1.1.1` package will not fix this problem and doesn't include the above updates, which are currently being staged in #8439.  With that said, as you can see above, I was able to install `accounts-google@1.1.1` on Meteor 1.3.  To help you more, can you include:

* The exact 1.3 version
* Your `.meteor/packages` file (as a Gist link, please)
* Your `.meteor/versions` file (as a Gist link, please)

?

realyze:
@abernix here you go:

METEOR@1.4.3.2
https://gist.github.com/realyze/c995ccc1a87784846bbcab5b4bb28e39

I'm using copies of the `accounts-google` and `google-oauth` packages from the https://github.com/meteor/meteor/pull/8439 branch and it seems like falling down a rabbit hole a little.

The simple part was figuring out I need to add `REVERSED_CLIENT_ID`. Then I realised that `cordova-plugin-googleplus` does not play well with phonegap-plugin-push (https://github.com/EddyVerbruggen/cordova-plugin-googleplus/issues/247) so I tried to use the fork from @yehiasalam . So then the app actually builds but it doesn't redirect back to the app because of https://github.com/EddyVerbruggen/cordova-plugin-googleplus/issues/191 .

So you know... :-)

packages/google-oauth/google_server.js
abernix: Agree about the actual contents of the `picture` field, but fwiw, your mapping here is def correct.

From my general browsing of the (endless!) Google Auth/Login docs, I believe other fields will be okay – `email`, included.

adamgins:
Hi. If we're on Meteor 1.4 can someone please summarise the steps needed to get their working or are we waiting for some fix still?
Thanks

abernix:
@adamgins We're not aware of any other fix that's necessary for you to move forward with this right now, though there are some additional requirements of the Google Sign-In SDK.

Like some other Cordova-based platforms, the new implementation in Meteor 1.4.4 is using the `cordova-plugin-googleplus` plug-in.  Notably different with this implementation are the additional steps required for setting API tokens and signing certificates for the Google Sign-In SDK.  I personally found the process a bit more difficult to traverse on Android than iOS, but that certainly seems arguable and your mileage may vary.

The details regarding the additional steps which need to be taken are outlined in [Step 3 on the `README`](https://github.com/EddyVerbruggen/cordova-plugin-googleplus#3-google-api-setup) for the `cordova-plugin-googleplus` plugin.  For starters, have you tried updating your application to Meteor 1.4.4 which uses the new `accounts-google` version 1.1.2?  I could probably provide more specific information knowing if you're building for iOS, Android or both.  To update to Meteor 1.4.4, you can run the following in your application directory:

    meteor update

We'd certainly like to make this easier though, so your feedback would be appreciated.  Also, please let us know if you run into struggles and we'll do our best to help iron them out.

adamgins:
@abernix thanks. OK, I think I have the config right.
Although something happened yesterday and I cannot get my 1.4.4.1 project to build in Xcode logged here https://github.com/meteor/meteor/issues/8584 anything come to mind on that one?

Issue #8532, opened 2017-03-28
Verify Email function does not remove tokens for email in a different case(upper or lower)


hwillson:
Migrating over from meteor/accounts#17.

----

(originally posted by @port80labs)

Lets say user adds an email into his account as `abc@mail.com`. then adds the same email as `ABC@mail.com`. when he gets the second verification email and clicks it, it removes the token records for emails sent to ABC@mail.com but not abc@mail.com.

i think the issue is this line -

https://github.com/meteor/accounts/blob/master/packages/accounts-password/password_server.js#L838

we should instead -

```js
$pull: {'services.email.verificationTokens': {address: { $regex : new RegExp('^' + Meteor._escapeRegExp(tokenRecord.address) + '$', 'i') }}}});
```

ScaY:
Hello,

According to the tests added it seems we do not need to update the condition on the `$pull` (cf. https://github.com/meteor/meteor/pull/8555)

Regards

zimme:
Keep in mind that the local part of emails are case-sensitive, however most email providers don't use that part of the spec and make the local part of emails on their service case-insensitive.

So should we be email spec compliant or do whatever most email providers does?

hwillson:
@ScaY - very interesting! Thanks for putting that test together and discovering this. I'll add more comments on your PR, but we should definitely get those test changes merged!

@zimme - great point, thanks! Based on the test changes @ScaY submitted, it looks like Meteor has decided to be email provider compliant 🙂.

hwillson:
Quick follow up @ScaY - I've reviewed the proposed test changes more thoroughly; Meteor is being email spec compliant. I don't think we should change the `$pull` clause (but I still think your PR should be merged) - I've outlined my reasoning in PR #8555. Thanks!

zimme:
@hwillson, If you look at this change from a security perspective it might be pretty bad to not be spec compliant.
If someone is using an email provider which is spec compliant there might be a small risk of account hijacking.

I have the email `simon@provider.email` and someone else have `Simon@provider.email`.
I signup with `simon@provider.email` and everything is fine, then someone else goes in and uses the reset password functionality and uses `Simon@provider.email`, the emails match case-insensitive and an email is sent to the provided email and not the accounts email.
User receives reset link and my account is now theirs.

Might not relate strictly to email validation, but still worth considering because if we make one part case-insensitive we might need to make all parts case-insensitive.

hwillson:
[...]

PR #8555, opened 2017-04-02
[TEST] Updating tests for accounts-password

ScaY:
Hello,

In order to reproduce the following issue https://github.com/meteor/meteor/issues/8532 I updated the tests but I couldn’t reproduce the bug.

Unless I am mistaken, it seems there is no issue when calling the `Accounts.verifyEmail` for the same mail in upper case because the verification token is correctly removed and the email is marked as _verified_.

Regards


ScaY:
Hi @hwillson thanks for your feedback :) !

Yes you're absolutely right the tests do miss the check on the `services.email.verificationTokens` array. I did not know how to check this array (apart with a mongoclient x) ) because the `services` property is not being published.

abernix:

Issue #8558, opened 2017-04-03
1.4.4-rc.4 fails on Windows: no directory ...\dev_bundle\server-lib


edemaine:
I've tried this both upgrading an existing Meteor project and starting a fresh new project:

```
C:\...>meteor create test --release 1.4.4-rc.4
Created a new Meteor app in 'test'.

To run your new app:
  cd test
  meteor

If you are new to Meteor, try some of the learning resources here:
  https://www.meteor.com/learn

meteor create --bare to create an empty app.
meteor create --full to create a scaffolded app.

C:\...>cd test

C:\...\test>meteor
[[[[[ C:\...\test ]]]]]

=> Started proxy.
=> Started MongoDB.
C:\Users\edemaine\AppData\Local\.meteor\packages\templating-compiler\1.3.2\plugin.compileTemplatesBatch.os\npm\node_modules\meteor\promise\node_modules\meteor-promise\promise_server.js:190
      throw error;
      ^

Error: ENOENT: no such file or directory, stat 'C:\Users\edemaine\AppData\Local\.meteor\packages\meteor-tool\1.4.4-rc.4\mt-os.windows.x86_32\dev_bundle\server-lib\node_modules'
    at Error (native)
    at Object.fs.statSync (fs.js:844:18)
    at Object.wrapper (C:\tools\fs\files.js:1556:35)
    at Object.files.(anonymous function) [as stat] (C:\tools\fs\files.js:1658:27)
    at symlinkWithOverwrite (C:\tools\isobuild\builder.js:732:17)
    at Builder.write (C:\tools\isobuild\builder.js:266:7)
    at ServerTarget.write (C:\tools\isobuild\bundler.js:2265:15)
    at C:\tools\isobuild\bundler.js:2373:30
    at C:\tools\isobuild\bundler.js:2506:9
    at Array.forEach (native)
    at C:\tools\isobuild\bundler.js:2496:26
    at C:\tools\isobuild\bundler.js:2837:22
    at C:\tools\utils\buildmessage.js:271:13
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at C:\tools\utils\buildmessage.js:264:29
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at C:\tools\utils\buildmessage.js:262:18
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
[...]

PR #8560, opened 2017-04-04
Improve `fs-extra`s `move` calls for Windows platform.

abernix:
This is a follow-up to meteor/meteor#8491 which worked properly on Unix platforms, but failed in a variety of ways on Windows due to its lack of Fiber-awareness and desire to create symlinks as unprivileged users (something not always possible on Windows).

The Fiber issue was observed when trying to remove "src" directories within the `move` function (which tries a variety of OS/OS/arch-specific techniques to accomplish its goal) after they had been copied to "dest".  On Windows, this resulted in `EDIRNOTEMPTY` errors since Windows appears to temporarily cache the file-handle or doesn't release the file-handle until the next tick.

The symlink issue will hopefully improve in an upcoming release of Windows (Creator Edition) when Microsoft makes it possible to create symlinks as an unprivileged user, however it will still require enabling "Developer" mode in Windows settings.  This implements the same catch which was already in place for `fs.rename` on the `fs.move` provided by `fs-extra`.

Performance was the same in tests comparing before and after these changes.

I've targeted this PR for `release-1.4.3.x`.
**Edit:** The `fs-extra` update to `2.1.2` seemed relevant enough to include so I'll rebuild `dev_bundle` accordingly if this lands.

Relates to:
https://github.com/meteor/meteor/issues/8558#issuecomment-291194385

/cc @jshimko

### DIFF
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -18,7 +18,7 @@ var packageJson = {
     "meteor-promise": "0.8.0",
     fibers: "1.0.15",
     promise: "7.1.1",
-    "fs-extra": "2.1.0",
+    "fs-extra": "2.1.2",
     // So that Babel 6 can emit require("babel-runtime/helpers/...") calls.
     "babel-runtime": "6.9.2",
     // For various ES2015 polyfills, such as Map and Set.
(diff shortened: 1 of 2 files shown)


abernix:
@benjamn Self-merging (after a follow-up self-critique commit ^^) so I can rebuild the entire `dev_bundle` for the next 1.4.4 RC with this change.

Without this PR the current RCs don't work on Windows, so we need this (or something like it) or make the original #8491 changes optionally enable-able via environment variable.  Happy to revisit based on any feedback you have!

PR #8565, opened 2017-04-05
add function to expose tracker inFlush variable

maxnowack:
`inFlush` indicates wether a flush is active or not. Slightly the same like `Tracker.active` with one big difference. `Tracker.active` gets unset inside `Tracker.nonreactive`.
I want to savely call `Tracker.flush()` without getting the `Can't call Tracker.flush while flushing` error.
`Tracker.active` doesn't work for me, since I don't know if the flush was called inside `Tracker.nonreactive`. I think I need to access the `inFlush` variable to get the global flushing state.

Background:
I'm writing a package to connect reactive data to react components. In direct comparison with `react-meteor-data`, I want to reduce the reruns of the reactive function if the props of the component change. To avoid unnecessary renders, I'm calling `Tracker.flush()` inside of `componentWillReceiveProps` to run the reactive function and set the new state.

### DIFF
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -456,6 +456,15 @@ Tracker.flush = function (options) {
                       throwFirstError: options && options._throwFirstError });
 };
 
+/**
+ * @summary True if we are computing a computation now, either first time or recompute.  This matches Tracker.active unless we are inside Tracker.nonreactive, which nullfies currentComputation even though an enclosing computation may still be running.
+ * @locus Client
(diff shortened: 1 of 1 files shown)


benjamn:
Thanks @maxnowack, and good luck with your reactive React package.

abernix:

avalanche1:
No mention in http://docs.meteor.com/api/tracker.html about it...

maxnowack:
@avalanche1 you're right. I've created a PR: meteor/docs#135

PR #8569, opened 2017-04-06
Recommendation to switch `fs-extra` for an alternate implementation.

abernix:
This proposes reverting/replacing #8491, which has been problematic in the 1.4.4 series of RCs, specifically on Windows.  More so, I believe that `fs-extra` could be overboard and we could go with a much more simple implementation to accomplish the goal of fixing cross-device copies.

Unfortunately, `fs-extra` is still not as perfect at handling various file system conditions as would be ideal.  It seemed sensical to try and use a library like this, however it turns out that the Meteor suite of file system functions performs better on Windows.  For example, `fs-extra` still tries to create symlinks as an unprivileged user – a forbidden task on Windows unless running as Administrator.

In addition, I ran into a constant stream of other errors: `ENOTEMPTY`, `EBUSY`, `EEXIST` and `EPERM` (on an `open`, no less) – all for various, Windows-only (from what I can gather) reasons.

My current recommendation is that we remove `fs-extra` and replace the [`Builder#complete` `renameDirAlmostAtomically` call](/tools/isobuild/builder.js) (which does not absolutely _have_ to be done atomically; "fast" is the goal) with a `try`/`catch` which resorts to a basic copy if `err.code === 'EXDEV'`.  All other functionality stays the same.

I've crudely implemented it with 523c7ce9d4d47442becc307817a35d6bb9800915 but would be open to other suggestions.

This reverts commits:

* d49f3e2
* 3257baf
* 74cb8eb
* 5bbdcc9
* 6a0767b

/cc @jshimko

### DIFF
--- a/tools/fs/files.js
+++ b/tools/fs/files.js
@@ -48,6 +48,19 @@ var useParsedSourceMap = function (pathForSourceMap) {
 // Try this source map first
 sourceMapRetrieverStack.push(useParsedSourceMap);
 
+// Fibers are disabled by default for files.* operations unless
+// process.env.METEOR_DISABLE_FS_FIBERS parses to a falsy value.
+const YIELD_ALLOWED = !! (
+  _.has(process.env, "METEOR_DISABLE_FS_FIBERS") &&
+  ! JSON.parse(process.env.METEOR_DISABLE_FS_FIBERS));
+
+function canYield() {
+  return YIELD_ALLOWED &&
+    Fiber.current &&
+    Fiber.yield &&
+    ! Fiber.yield.disallowed;
+}
+
 // given a predicate function and a starting path, traverse upwards
 // from the path until we find a path that satisfies the predicate.
 //
@@ -283,9 +296,7 @@ files.rm_recursive = Profile("files.rm_recursive", function (p) {
     rimraf.sync(path);
   } catch (e) {
     if (e.code === "ENOTEMPTY" &&
-        Fiber.current &&
-        Fiber.yield &&
-        ! Fiber.yield.disallowed) {
+        canYield()) {
       new Promise((resolve, reject) => {
         rimraf(path, err => {
           err ? reject(err) : resolve();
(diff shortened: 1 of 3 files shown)


tools/isobuild/builder.js
```diff
@@ -687,7 +687,16 @@ Previous builder: ${previousBuilder.outputPath}, this builder: ${outputPath}`
     // case of renameDirAlmostAtomically since that one is constructing files to
     // be checked in to version control, but here we could get away with it.
     if (this.buildPath !== this.outputPath) {
-      files.renameDirAlmostAtomically(this.buildPath, this.outputPath);
+      try {
+        files.rename(this.buildPath, this.outputPath);
```
abernix: This portion was just a proposal.  This is likely failing because the `rename`s and `rm_recursive`s from `renameDirAlmostAtomically` are necessary to make way for the incoming tree.

As a thought, maybe we should continue to use `renameDirAlmostAtomically` here instead of `rename`.

tools/isobuild/builder.js
```diff
@@ -687,7 +687,16 @@ Previous builder: ${previousBuilder.outputPath}, this builder: ${outputPath}`
     // case of renameDirAlmostAtomically since that one is constructing files to
     // be checked in to version control, but here we could get away with it.
     if (this.buildPath !== this.outputPath) {
-      files.renameDirAlmostAtomically(this.buildPath, this.outputPath);
+      try {
+        files.renameDirAlmostAtomically(this.buildPath, this.outputPath);
+      } catch (e) {
+        if (e.code === "EXDEV") {
+          files.rm_recursive(this.outputPath);
+          files.cp_r(this.buildPath, this.outputPath);
```
benjamn: Let's move this logic into `files.renameDirAlmostAtomically`, as discussed offline.

tools/fs/files.js
```diff
@@ -986,6 +979,79 @@ files.createTarball = function (dirPath, tarball, options) {
...
+    // If we've been forced to jeopardize our atomicity due to file-system
+    // limitations, we'll resort to copying.
+    if (forceCopy) {
+      files.rm_recursive(toDir);
+      files.cp_r(fromDir, toDir);
+    }
+
+    // ... and take out the trash.
+    if (cleanupGarbage) {
+      files.rm_recursive(garbageDir);
```
benjamn: Remember that we could do this cleanup asynchronously.

abernix: Yup.  Done with https://github.com/meteor/meteor/pull/8569/commits/ae35f13ed2b7511cf82b8d263657e349bf45dd5a!

Issue #8448, opened 2017-03-03
Password token cleanup occurring on non-password apps.


abernix:
https://github.com/meteor/meteor/pull/8163 was intended to fix slower queries on larger `Meteor.users` tables which were occurring during the password token expiration process.

While it does fix it in any case where the `accounts-password` package is being used (and thus [these indexes](https://github.com/meteor/meteor/blob/d5c7026a4d03664700e59febef5ec5b251195abd/packages/accounts-password/password_server.js#L1060-L1065) being created), it does not fix it when `accounts-password` is not present, for example, on sites which only offer OAuth login options (other `accounts*` packages) rather than password login.

One possible solution might be to move the index creation itself to [`accounts-base`'s index creation](https://github.com/meteor/meteor/blob/56ad858559c4222fad7d44115a63c3c759fc2abe/packages/accounts-base/accounts_server.js) instead.

Alternatively, it's plausible (I haven't investigated further) that, [these recurring tasks](https://github.com/meteor/meteor/blob/56ad858559c4222fad7d44115a63c3c759fc2abe/packages/accounts-base/accounts_server.js#L1093-L1200) should only be present in (and their timers set from) from [`accounts_password`](https://github.com/meteor/meteor/tree/devel/packages/accounts-password).

(Thanks to @glasser for [reporting this](https://github.com/meteor/meteor/pull/8163#issuecomment-283767598)!)

abernix:
This should be fixed in Meteor 1.4.4.2!

    meteor update --release 1.4.4.2

(or just `meteor update`, since it's currently the latest, recommended version)

/cc @glasser

mitar:
Hm, but why has not password cleanup tokens code been moved to `accounts-password` package? Why index creation to `accounts-base`?

mitar:
It is strange that `services.password.reset.token` index is in `accounts-password`, but `services.password.reset.when` in `accounts-base`.

abernix:
@mitar If you'd like to submit a PR with a more complete fix (something I suggested may be a reasonable idea above), please do.  That said, this should solve the problem originally reported, which was the lack of an index on the field.  Also, the overhead of cleaning up a nil amount of password tokens should be very minimal, especially with the index in place.

mitar:
OK, so it was done to keep the change small.

Yes, I do not think there is any performance downside of having that code there. It is just unclean abstraction. Which I do not mind enough to do a PR for. :-)

PR #8571, opened 2017-04-06
Index creation in the file using the query

Akarshit:
Fix for #8448 
The changes have been made according to the comment- 
> [index should be added in the same package that does the query](https://github.com/meteor/meteor/pull/8163#issuecomment-283767598)


abernix:
Thanks for opening a PR.

I believe you'll need to put this in the section where the rest of the default `users` indexes are created, [around these lines](https://github.com/Akarshit/meteor/blob/c571bae99da05cb833df6ad4e9d8c4d173a70506/packages/accounts-base/accounts_server.js#L1490).

Akarshit:
@abernix Thanks, just found that out. :)

abernix:
Thanks for applying this change, @Akarshit!

Issue #7183, opened 2016-06-07
Hot-code-reload does not work when using UIWebview on IOS


macrozone:
WkWebview lacks some important features (see https://github.com/meteor/meteor/issues/7126 and https://github.com/meteor/meteor/issues/7173), but it is possible to switch to UiWebview which does not has these limitations (see https://github.com/meteor/docs/issues/48).

But hot-code-reload on local development does no longer work with UIWebview, so you have the choice between two broken implementations. I did not test hot-code-reload on a production app with UiWebview.

Edit: The UiWebview does reload when you do a local change, but still shows the same content.

martijnwalraven:
Hmm, I'm not sure why reload would no longer work with UIWebView. The only place in the code that is WKWebView-specific is [`forceReload()`](https://github.com/meteor/cordova-plugin-meteor-webapp/blob/master/src/ios/WebAppLocalServer.swift#L307-L311), which is only used when reverting to the last known good version.

realyze:
It would be great if MDG could bump the priority of this. Rebuilding the whole app to get new changes to the phone you're developing on is just, well, soul crushing.

macrozone:
Still an issue, totally forgot about it and ran again in it.

lucfranken:
That auto play issue seems to be fixed? Or is that still an active issue? So wouldn't it be fixed by just updating wkwebview?

trymbill:
+1, dealing with this issue too.

macrozone:
@lucfranken: i'll have to check whether this update is already in meteor's cordova version. Then I could switch to wkwebview

ramezrafla:
@trymbill and @macrozone,

Feel your pain! Have you guys tested out hot-code-reload in production? Maybe using a non-local server would fix this (i.e. non-production App but production server)

trymbill:
I just switched back to wkwebview and forced meteor to use the latest
version of it and I can confirm that autopsy now works

ramezrafla:
Thanks - we are stuck UIWebView for now as we need in-app-browser (that we van control). Do we know if hot code push works on it now?

martijnwalraven:
I looked into this, and it seems [a Cordova bug](https://issues.apache.org/jira/browse/CB-11375) is the culprit, and has been been fixed in Cordova iOS 4.2. 

[...]

Issue #7030, opened 2016-05-11
Allow developers to override cordovaCompatibilityVersions


denisbabineau:
**_4 Upvotes**_ Current compatibility check for Hot Code Push to mobile blocks any updates if any of the cordova components' versions were bumped.

We need a way to override this default. See discussions at:

https://forums.meteor.com/t/cordova-app-android-not-reloading-after-hot-code-push-with-deployed-app/16336/38

suggested solution by @martijnwalraven 

> I think the solution is to allow developers to override cordovaCompatibilityVersions (the same way we allow AUTOUPDATE_VERSION to be defined manually). This way, you could set it to the current value when adding new plugins so hot code push won't be blocked.

hlehmann:
+1

mikepaszkiewicz:
did a little exploring, would this modification in  [isobuild/bundler.js](https://github.com/meteor/meteor/blob/baa3f6a152af96ff1e7a7f9f0e00c86c9cd922ee/tools/isobuild/bundler.js#L1372)  be a potential way to override?
``` 
const hash = process.env.CORDOVA_COMP_VERS || WebAppHashing.calculateCordovaCompatibilityHash(
            version,
            this.cordovaDependencies); 
```

shreddapp:
+1 went through hell to find out why HCP was not working for me... Could not find why hashes differed in prod though.

blouze:
+1 went through hell to find out why HCP was not working for me... Could not find why hashes differed in prod though.

levinse:
+1 this will be a major problem for us once we switch to 1.3. 

It's not really practical to block HCP if cordova plugins have different versions because it's a major barrier to high release velocity. 

We're used to handling  different versions within the app client/server code and we routinely have 1-2 older versions of clients as well as differing versions of iOS vs Android live in the App/Play stores. By letting us continue to handle these differences within code, as was the case with 1.2, we're able to continue to release the app to our app servers and via HCP much more rapidly (multiple times per day currently). There's no actual reason to prevent HCP if we're able to handle such differences within code. Without HCP, we're blocked for days waiting for Apple to approve a release, which significantly lowers our release velocity.

martijnwalraven:
@mikepaszkiewicz: Yes, I think that is where you'd want to override the versions. But since versions differ between platforms, and you'll need to specify the current version to enable hot code pushes, you'd need to find a way to pass versions for both platforms.

JonTraduct:
[...]

PR #8581, opened 2017-04-10
Implement CORDOVA_COMPATIBILITY_VERSION_IOS/ANDROID and EXCLUDE

wojtkowiak:
Hey,

quoting commit's message:
CORDOVA_COMPATIBILITY_VERSION_IOS or CORDOVA_COMPATIBILITY_VERSION_ANDROID allows to override compatibility version for a specified platform.

CORDOVA_COMPATIBILITY_VERSION_EXCLUDE provides a way of excluding a certain plugin from compatibility version calculation. You can pass several plugin names with ';'. For example: `CORDOVA_COMPATIBILITY_VERSION_EXCLUDE='cordova-plugin-crosswalk-webview;cordova-plugin-device'`

This addresses:
https://github.com/meteor/meteor/issues/7030
and somehow: https://github.com/meteor/meteor/issues/7183

Why **CORDOVA_COMPATIBILITY_VERSION_EXCLUDE**?
This can be used in many ways but the one crucial is crosswalk. Since it will no longer be developed https://crosswalk-project.org/blog/crosswalk-final-release.html this is crucial to be able to build two versions of the app while still being able to benefit from HCP on both. One can add the crosswalk plugin to exclude for instance and have the same compatibility version on both builds (the one with and without crosswalk).

@abernix, @martijnwalraven let me now what do you think about this 🙂

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -1525,11 +1525,16 @@ class ClientTarget extends Target {
 
       const cordovaCompatibilityVersions =
         _.object(_.map(CORDOVA_PLATFORM_VERSIONS, (version, platform) => {
-          let cordovaDependencies = Object.assign({}, this.cordovaDependencies);
-          let pluginsExcludedFromCompatibilityHash = process.env.CORDOVA_COMPATIBILITY_VERSION_EXCLUDE || '';
-          cordovaDependencies = _.omit(cordovaDependencies, pluginsExcludedFromCompatibilityHash.split(';'));
 
-          const hash = process.env[`CORDOVA_COMPATIBILITY_VERSION_${platform.toUpperCase()}`] ||
+          const pluginsExcludedFromCompatibilityHash = (process.env.METEOR_CORDOVA_COMPAT_VERSION_EXCLUDE || '')
+            .split(',');
+
+          const cordovaDependencies = Object.assign(
+            Object.create(null),
+            _.omit(this.cordovaDependencies, pluginsExcludedFromCompatibilityHash)
+          );
+
+          const hash = process.env[`METEOR_CORDOVA_COMPAT_VERSION_${platform.toUpperCase()}`] ||
               WebAppHashing.calculateCordovaCompatibilityHash(
                 version,
                 cordovaDependencies);
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -1522,11 +1522,17 @@ class ClientTarget extends Target {
     if (this.arch === 'web.cordova') {
       const { WebAppHashing } =
         loadIsopacket('cordova-support')['webapp-hashing'];
+
       const cordovaCompatibilityVersions =
         _.object(_.map(CORDOVA_PLATFORM_VERSIONS, (version, platform) => {
-          const hash = WebAppHashing.calculateCordovaCompatibilityHash(
-            version,
-            this.cordovaDependencies);
+          let cordovaDependencies = Object.assign({}, this.cordovaDependencies);
+          let pluginsExcludedFromCompatibilityHash = process.env.CORDOVA_COMPATIBILITY_VERSION_EXCLUDE || '';
+          cordovaDependencies = _.omit(cordovaDependencies, pluginsExcludedFromCompatibilityHash.split(';'));
+
+          const hash = process.env[`CORDOVA_COMPATIBILITY_VERSION_${platform.toUpperCase()}`] ||
+              WebAppHashing.calculateCordovaCompatibilityHash(
+                version,
+                cordovaDependencies);
           return [platform, hash];
         }));
       program.cordovaCompatibilityVersions = cordovaCompatibilityVersions;
--- a/tools/tests/cordova-hcp.js
+++ b/tools/tests/cordova-hcp.js
@@ -39,3 +39,50 @@ selftest.define(
 
     run.stop();
 });
+
+selftest.define(
+  "cordova CORDOVA_COMPATIBILITY_VERSION_* works", ["cordova", "slow"], function () {
+    var s = new Sandbox();
+    var run;
+
+    var androidCompatibilityVersion = '2.0';
+
+    // Override the compatibility version for android with CORDOVA_COMPATIBILITY_VERSION_ANDROID.
+    s.env.CORDOVA_COMPATIBILITY_VERSION_ANDROID = androidCompatibilityVersion;
+
+    s.createApp("myapp", "standard-app");
+    s.cd("myapp");
+
+    var platforms = s.read(".meteor/platforms");
+    s.write(".meteor/platforms", platforms + "\nandroid\n");
+
+    run = s.run("run", "android", "--mobile-server", "example.com");
+    run.waitSecs(30);
+    run.match("Started your app");
+
+    var result = JSON.parse(httpHelpers.getUrl(
+        "http://localhost:3000/__cordova/manifest.json"));
+
+    // Check in the manifest if the overriden version was used.
+    selftest.expectEqual(result.cordovaCompatibilityVersions.android, androidCompatibilityVersion);
+
+    run.stop();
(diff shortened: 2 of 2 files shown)


wojtkowiak:
Is there a way to run the CI with `self-test --slow`?

tools/isobuild/bundler.js
```diff
@@ -1522,11 +1522,17 @@ class ClientTarget extends Target {
...
       const { WebAppHashing } =
         loadIsopacket('cordova-support')['webapp-hashing'];
+
       const cordovaCompatibilityVersions =
         _.object(_.map(CORDOVA_PLATFORM_VERSIONS, (version, platform) => {
-          const hash = WebAppHashing.calculateCordovaCompatibilityHash(
-            version,
-            this.cordovaDependencies);
+          let cordovaDependencies = Object.assign({}, this.cordovaDependencies);
+          let pluginsExcludedFromCompatibilityHash = process.env.CORDOVA_COMPATIBILITY_VERSION_EXCLUDE || '';
```
abernix: I realize this variable is already long, but we should prefix (new) environment variables with `METEOR_`.

Also, I think "version" makes sense for the other environment variable, but maybe this should be "hash" (since it's being excluded from the `calculateCordovaCompatibilityHash` function)? (Also, it's shorter!)

Maybe `METEOR_CORDOVA_COMPAT_HASH_EXCLUDE`?

tools/isobuild/bundler.js
```diff
@@ -1522,11 +1522,17 @@ class ClientTarget extends Target {
...
+          cordovaDependencies = _.omit(cordovaDependencies, pluginsExcludedFromCompatibilityHash.split(';'));
```
abernix: Instead of re-assigning `cordovaDependencies`, is there a reason we can't switch the order the variables are created and then switch them to `const`s?

```js
const pluginsExcludedFromCompatibilityHash =
  (process.env.CORDOVA_COMPATIBILITY_VERSION_EXCLUDE || '')
  .split(';');
const cordovaDependencies =
  Object.assign(Object.create(null),
    _.omit(this.cordovaDependencies, pluginsExcludedFromCompatibilityHash));
```

tools/isobuild/bundler.js
```diff
@@ -1522,11 +1522,17 @@ class ClientTarget extends Target {
...
+
+          const hash = process.env[`CORDOVA_COMPATIBILITY_VERSION_${platform.toUpperCase()}`] ||
```
abernix: `METEOR_CORDOVA_COMPAT_VERSION_<platform>`?

tools/isobuild/bundler.js
abernix: Also, was there a particular reason for using a semi-colon (`;`) instead of a comma (`,`) as the delimiter when multiple packages are being excluded?

(Using a semi-colon makes it so the environment variable might have to be quoted on Unix.)

tools/tests/cordova-hcp.js
```diff
@@ -39,3 +39,50 @@ selftest.define(
...
+
+    // Check in the manifest if the overriden version was used.
+    selftest.expectEqual(result.cordovaCompatibilityVersions.android, androidCompatibilityVersion);
+
+    run.stop();
+    // Save the iOS compatibility version.
+    var iosCompatibilityVersion = result.cordovaCompatibilityVersions.ios;
+
+    // Now exclude one of the plugins from compatibility version calculation.
+    s.env.CORDOVA_COMPATIBILITY_VERSION_EXCLUDE = 'cordova-plugin-meteor-webapp';
```
abernix: Would it be difficult to test excluding two packages so the delimiter (`;`, etc.) is tested?

tools/tests/cordova-hcp.js
```diff
@@ -39,3 +39,50 @@ selftest.define(
...
+
+    var platforms = s.read(".meteor/platforms");
+    s.write(".meteor/platforms", platforms + "\nandroid\n");
+
+    run = s.run("run", "android", "--mobile-server", "example.com");
+    run.waitSecs(30);
+    run.match("Started your app");
+
+    var result = JSON.parse(httpHelpers.getUrl(
+        "http://localhost:3000/__cordova/manifest.json"));
```
abernix: This will fail if the tests are run on a different port (e.g. `meteor test-packages --port 3500`).  Can we test this through `WebAppInternals.getBoilerplate` (similar to [this](https://github.com/meteor/meteor/blob/devel/packages/webapp/webapp_tests.js#L111)?) or otherwise without an HTTP request?

tools/tests/cordova-hcp.js
```diff
@@ -39,3 +39,50 @@ selftest.define(
...
+
+    // Now exclude one of the plugins from compatibility version calculation.
+    s.env.CORDOVA_COMPATIBILITY_VERSION_EXCLUDE = 'cordova-plugin-meteor-webapp';
+
+    run = s.run("run", "android", "--mobile-server", "example.com");
+    run.waitSecs(30);
+    run.match("Started your app");
+
+    result = JSON.parse(httpHelpers.getUrl(
+        "http://localhost:3000/__cordova/manifest.json"));
```
abernix: Same as above.

tools/isobuild/bundler.js
wojtkowiak: Yeah, been thinking about this but I would rather stick to `VERSION` even though it's longer. It's named that way in the guide and also in the manifest.json you have `cordovaCompatibilityVersion` field. But it's your call so have a second thought 🙂 and let me know.

tools/isobuild/bundler.js
wojtkowiak: Nope, there is no reason. Good point, will be switched.

wojtkowiak: Also a good point, will be changed to a comma.

tools/tests/cordova-hcp.js
wojtkowiak: I've just used the same approach that was already used https://github.com/meteor/meteor/blob/devel/tools/tests/cordova-hcp.js#L34
`meteor test-packages --port 3500` this is not the case because this is `self-test` and you can not pass a port to it, right?

tools/tests/cordova-hcp.js
wojtkowiak: No, added. But since there is no easy way of actually checking if those were excluded the only thing you actually test is whether the syntax does not brake the functionality.

tools/isobuild/bundler.js
wojtkowiak: Changed

wojtkowiak:
I will squash the commits once we're done with this.

tools/isobuild/bundler.js
```diff
@@ -1522,11 +1522,22 @@ class ClientTarget extends Target {
...
         loadIsopacket('cordova-support')['webapp-hashing'];
+
       const cordovaCompatibilityVersions =
         _.object(_.map(CORDOVA_PLATFORM_VERSIONS, (version, platform) => {
-          const hash = WebAppHashing.calculateCordovaCompatibilityHash(
-            version,
-            this.cordovaDependencies);
+
+          const pluginsExcludedFromCompatibilityHash = (process.env.METEOR_CORDOVA_COMPAT_VERSION_EXCLUDE || '')
+            .split(',');
```
benjamn: This will give `[""]` rather than `[]` in the default case; is that what you want?

wojtkowiak: Yeah, I know. It's not producing any side effects in `_.omit` so I'm okay with this. Otherwise, it would look like `process.env.METEOR_CORDOVA_COMPAT_VERSION_EXCLUDE ? process.env.METEOR_CORDOVA_COMPAT_VERSION_EXCLUDE.split(',') : []`. Or with an extra line to assign this env var to sth shorter 🙂

abernix:

wojtkowiak:
Forgot to respond. Works alright! Thanks for accepting this feature!

Issue #8586, opened 2017-04-12
New Google OAuth on Cordova doesn't properly fill the user.service document and/or profile


realyze:
Meteor@1.4

When signing in from Cordova, Meteor@1.4 uses `cordova-plugin-googleplus` but it doesn't properly fetch user data from Google on successful login.

The HTTP call in https://github.com/meteor/meteor/blob/devel/packages/google-oauth/google_server.js#L13-L16 doesn't fetch some important data such as `first_name`, `last_name` or `refresh_token`. The `OAuth.registerService` callback fetches everything correctly but it never seems to be called when logging in from Cordova.

benjamn:
@realyze Please close this issue whenever you can verify that `google-oauth@1.2.2` fixes the problem!

benjamn:
Closing since I believe this is fixed, but please feel free to reopen if you discover otherwise.

Note that the latest version is now `google-oauth@1.2.3`, with the changes in https://github.com/meteor/meteor/pull/8604.

PR #8588, opened 2017-04-12
Fixed Google OAuth user data fetching.

realyze:
This PR fixes the user data fetching issues in new `google-oauth` package, see https://github.com/meteor/meteor/issues/8586 (and the parent ticket https://github.com/meteor/meteor/issues/8253) for details.

The problem was that the new login handler did not fetch the right data (was not using the right endpoint).

Not sure who to notify so I'll just mention @abernix and @benjamn (sorry guys :-) ).

### DIFF
--- a/packages/google-oauth/google_client.js
+++ b/packages/google-oauth/google_client.js
@@ -1,5 +1,13 @@
 var Google = require("./namespace.js");
 
+var ILLEGAL_PARAMETERS = {
+  'response_type': 1,
+  'client_id': 1,
+  'scope': 1,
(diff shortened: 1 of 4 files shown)


hwillson:
Thanks for this @realyze! This will be reviewed shortly, but just a quick heads up - I've triggered a new CircleCI run for this, since the previously reported failure doesn't appear to be related to anything here. Thanks again!

benjamn:
These changes have now been published to Atmosphere as `google-oauth@1.2.2`!

PR #8603, opened 2017-04-14
[Fix #5167] $push with $slice with positive numbers

### LINKED ISSUE #5167: $push with $slice works on the server side but not in Minimongo
See reproduction (test `positive-numbers-slice`) here: https://github.com/mitar/minimongo-diff/tree/split-tests

mutdmour:
Fixed minimongo to be able to splice with positive numbers. 
[MongoDB ref](https://docs.mongodb.com/v3.0/reference/operator/update/slice/)

Fix #5167

### DIFF
--- a/packages/minimongo/minimongo_tests.js
+++ b/packages/minimongo/minimongo_tests.js
@@ -3158,13 +3158,14 @@ Tinytest.add("minimongo - $near operator tests", function (test) {
   // 'y'.
   testNear([2, 2], 1000, ['x', 'y']);
 
-  // Ensure that distance is used as a tie-breaker for sort.
+  // issue #3599
+  // Ensure that distance are not used as a tie-breaker for sort.
   test.equal(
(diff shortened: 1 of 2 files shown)


packages/minimongo/modify.js
```diff
@@ -349,8 +346,10 @@ var MODIFIERS = {
     if (slice !== undefined) {
       if (slice === 0)
         target[field] = [];  // differs from Array.slice!
-      else
+      else if (slice < 0)
         target[field] = target[field].slice(slice);
+      else 
+        target[field].splice(slice);
```
benjamn: Why not `target[field] = target[field].slice(0, slice)`, to avoid mutating the original array (like above)?

mutdmour: I could do that.

mutdmour:
@benjamn sorry I mixed up my branches and forced pushed on the wrong branch. I will have to redo this

PR #8604, opened 2017-04-14
Use tokens from Google Sign-In instead of calling getTokens.

benjamn:
Follow-up to https://github.com/meteor/meteor/pull/8588.

While testing #8588, I noticed that `request.serverAuthCode` was sometimes/always an empty string (on iOS, at least), so the `query` that was passed to `getServiceData` didn't have a usable `.code` property.

I then realized that the response from Google Sign-In already contains the tokens that would be returned from `getTokens`, so this PR avoids calling `getTokens` in that case, though it still shares most of the `serviceData` logic.

cc @realyze @ramezrafla @abernix

### DIFF
--- a/packages/google-oauth/google_server.js
+++ b/packages/google-oauth/google_server.js
@@ -6,20 +6,21 @@ var hasOwn = Object.prototype.hasOwnProperty;
 Google.whitelistedFields = ['id', 'email', 'verified_email', 'name', 'given_name',
                    'family_name', 'picture', 'locale', 'timezone', 'gender'];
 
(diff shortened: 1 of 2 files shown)


benjamn:
I've merged this and released `google-oauth@1.2.3`, since it seems to fix Google Sign-In for iOS. Any feedback is still welcome, but let's keep moving forward with patch updates.

realyze:
@benjamn Hmm, strange, `serverAuthCode` was always (?) there for me when I tested it. Thanks for looking into this!

PR #8605, opened 2017-04-16
Switch to modern nodemailer 4, Node 4 version. Fix #8591

### LINKED ISSUE #8591: Meteor email keeps SMTP connection opened and causes error with some email providers
RE: #8425 #8495 

@edemaine @abernix 

Since the "Upgrade to mailcomposer@4 and smtp-connection@2." I have issues with SMTP connections.

The MDG-email package keeps the SMTP connection opened.  It never send a SMTP "QUIT" command.

Some email providers, for instance Amazon SES, don't like people to keep their connections opened.  Amazon SES close the SMTP connection after 10 seconds with a timeout message and when it happens I get this error:

> Error: Unexpected Response: 421 Timeout waiting for data from client.

This error wouldn't happen if you would close the SMTP connection after having sent the email.

Also, this error message is very hard to process in a try/catch because it's impossible to know if the mail has been sent correctly prior the timeout. In my particular case, the "421 Timeout" happens after the "250 Ok" (that follows the email being sent).  So I could probably ignore the error ... but if the "421 Timeout" was real, I would miss the chance to retry sending the email.

_(edit: I am not sure if the previous package also kept the SMTP connection opened and simply ignored the resulting error)_

edemaine:
* Most critically, use a pool instead of direct SMTP connection, to handle dropped connections and increase throughput, like mail module 1.1.  (#8591)
* New nodemailer's sendMail wants an options object, not a MailComposer object.  Luckily, a MailComposer object has a "mail" field that remembers the original options, so we can keep original behavior.
* However, we no longer support the mailComposer option set to a compiled MailComposer object (functionality that was briefly added in 1.2.0).
* nodemailer does SMTP URL parsing now automatically for us, simplifying code.
* Tests' outputs now end with additional "\r\n"
* Drop underscore package dependency (no longer needed)

### DIFF
--- a/packages/email/.npm/package/npm-shrinkwrap.json
+++ b/packages/email/.npm/package/npm-shrinkwrap.json
@@ -1,79 +1,14 @@
 {
   "dependencies": {
-    "addressparser": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/addressparser/-/addressparser-1.0.1.tgz",
-      "from": "addressparser@1.0.1"
-    },
-    "buildmail": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/buildmail/-/buildmail-4.0.1.tgz",
-      "from": "buildmail@4.0.1"
-    },
-    "httpntlm": {
-      "version": "1.6.1",
-      "resolved": "https://registry.npmjs.org/httpntlm/-/httpntlm-1.6.1.tgz",
-      "from": "httpntlm@1.6.1"
-    },
-    "httpreq": {
-      "version": "0.4.23",
-      "resolved": "https://registry.npmjs.org/httpreq/-/httpreq-0.4.23.tgz",
-      "from": "httpreq@>=0.4.22"
-    },
-    "iconv-lite": {
-      "version": "0.4.15",
-      "resolved": "https://registry.npmjs.org/iconv-lite/-/iconv-lite-0.4.15.tgz",
-      "from": "iconv-lite@0.4.15"
-    },
-    "libbase64": {
-      "version": "0.1.0",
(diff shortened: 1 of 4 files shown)


abernix:
Just an FYI! This is a relatively easy situation to have happen with [Git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) (which we use), but this PR had the `blaze` submodule pointing to a different point than the `devel` branch.  This is the explanation for [the test failures](https://circleci.com/gh/edemaine/meteor/5?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) on [both CI providers](https://travis-ci.org/meteor/meteor/builds/222483200) and [these additional "changes"](https://github.com/meteor/meteor/pull/8605/commits/741cc4c4c8cc0cc61227da5ae5794ddf12362f98#diff-a580f45dda4290c703d17a674dbe4889) which were appearing at the bottom of the Diff.

To rectify this (I'd like to look at this early this coming week!), I've merged the most recent `devel` in and updated the submodule to the correct ref and pushed the changes to this PR which re-triggered the CI builds.  Hopefully they pass now (so far so good!)

In the future, in addition to the usual [syncing a fork](https://help.github.com/articles/syncing-a-fork/), it should just be a matter of running:

...if you ever notice the following in `git status`:

    modified:   packages/non-core/blaze (untracked content)

:smile:

engelgabriel:
We are waiting on this to close:
https://github.com/RocketChat/Rocket.Chat/issues/6705

edemaine:
Thanks!!  I was confused about that; your explanation makes sense.  Thanks for both fixing it and explaining so I don't make the mistake in the future.

abernix:
Yes, most of these were just general clean-up, certainly not specific to any of your additions/changes, but made sense to do while testing/reviewing it already. :wink:

PR #8608, opened 2017-04-17
[Fix #5165] don't allow invalid field names

### LINKED ISSUE #5165: Minimongo difference: it allows $ prefixed keys
If you try to use `$` prefixed key on the server, you will get an error. But Minimongo allows them. For example, for key `$each` you will get the following error:

```
$each is not valid for storage.
```

See reproduction (test `each-storage`) here: https://github.com/mitar/minimongo-diff/tree/split-tests

mutdmour:
#5165 

This fixes issues with minimongo allowing updates with invalid field names. I started out by looking at $addToSet as per ticket, and then expanded to check on other modifiers. I also checked on insert which was handling this issue. 

I ran all the tests on server MongoDB to confirm consistent behavior.

### DIFF
--- a/packages/minimongo/helpers.js
+++ b/packages/minimongo/helpers.js
@@ -43,28 +43,3 @@ isOperatorObject = function (valueSelector, inconsistentOK) {
 isNumericKey = function (s) {
   return /^[0-9]+$/.test(s);
 };
-
-// Make sure field names do not contain Mongo restricted
-// characters ('.', '$', '\0').
-// https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
-const invalidCharMsg = {
-  '.': "contain '.'",
-  '$': "start with '$'",
-  '\0': "contain null bytes",
-};
-isValidFieldName = function (key){
-      let match;
-      if (_.isString(key) && (match = key.match(/^\$|\.|\0/))) {
-        throw MinimongoError(`Key ${key} must not ${invalidCharMsg[match[0]]}`);
-      }
-}
-
-// checks if all field names in an object are valid
-hasValidFieldNames = function(doc){
-  if (doc && typeof doc === "object") {
-    JSON.stringify(doc, (key, value) => {
-      isValidFieldName(key);
-      return value;
-    });
-  }
-}
--- a/packages/minimongo/helpers.js
+++ b/packages/minimongo/helpers.js
@@ -43,3 +43,28 @@ isOperatorObject = function (valueSelector, inconsistentOK) {
 isNumericKey = function (s) {
   return /^[0-9]+$/.test(s);
 };
+
+// Make sure field names do not contain Mongo restricted
+// characters ('.', '$', '\0').
+// https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
+const invalidCharMsg = {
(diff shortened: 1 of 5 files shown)


packages/minimongo/helpers.js
```diff
@@ -43,3 +43,28 @@ isOperatorObject = function (valueSelector, inconsistentOK) {
...
+};
+isValidFieldName = function (key){
+      let match;
+      if (_.isString(key) && (match = key.match(/^\$|\.|\0/))) {
+        throw MinimongoError(`Key ${key} must not ${invalidCharMsg[match[0]]}`);
+      }
+}
+
+// checks if all field names in an object are valid
+hasValidFieldNames = function(doc){
```
benjamn: Let's call this `assertHasValidFieldNames`, and export it using ECMAScript syntax:
```js
export function assertHasValidFieldNames(doc) {
   ...
}
```
Then `import` it elsewhere:
```js
import { assertHasValidFieldNames } from "./helpers.js";
```

packages/minimongo/helpers.js
```diff
@@ -43,3 +43,28 @@ isOperatorObject = function (valueSelector, inconsistentOK) {
...
+const invalidCharMsg = {
+  '.': "contain '.'",
+  '$': "start with '$'",
+  '\0': "contain null bytes",
+};
+isValidFieldName = function (key){
+      let match;
+      if (_.isString(key) && (match = key.match(/^\$|\.|\0/))) {
+        throw MinimongoError(`Key ${key} must not ${invalidCharMsg[match[0]]}`);
+      }
```
benjamn: Fix indentation (two spaces per tab) please.

packages/minimongo/helpers.js
```diff
@@ -43,3 +43,28 @@ isOperatorObject = function (valueSelector, inconsistentOK) {
...
+
+// Make sure field names do not contain Mongo restricted
+// characters ('.', '$', '\0').
+// https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
+const invalidCharMsg = {
+  '.': "contain '.'",
+  '$': "start with '$'",
+  '\0': "contain null bytes",
+};
+isValidFieldName = function (key){
```
benjamn: My comment about `assertHasValidFieldNames` applies here too (`assertIsValidFieldName`, `export`, `import`).

packages/minimongo/helpers.js
mutdmour: Is the exporting/importing to help make it clearer where the function is coming from?

mutdmour:
@benjamn not sure why this run test is failing, has nothing to do with minimongo.

hwillson:
@mutdmour I've triggered a rebuild; sometimes that's all that is needed (for timeouts). We'll know shortly - thanks!

packages/minimongo/minimongo.js
```diff
@@ -1,3 +1,5 @@
+import { assertHasValidFieldNames } from './validation.js';
```
benjamn: Yep, this is a lot clearer to me!

abernix:

brucejo75:
I just entered #8775, this seems like a pretty fundamental issue that was not caught.

Seems like the tests should have caught this...

PR #8609, opened 2017-04-17
email@1.2.1

### LINKED ISSUE #8591: Meteor email keeps SMTP connection opened and causes error with some email providers

abernix:
This PR simply cherry-picks the great work from @edemaine in 6d4562656641b3c1f6c4a79e6a3c60c61b0136d5 (PR https://github.com/meteor/meteor/pull/8605) into `master` for the official `email@1.2.1` release.  The changes include those already in the original PR, as outlined below:

* Switch to modern nodemailer 4, Node 4 version. Fix #8591
* Most critically, use a pool instead of direct SMTP connection,
  to handle dropped connections and increase throughput,
  like mail module 1.1.  (#8591)
* New nodemailer's sendMail wants an options object, not a MailComposer
  object.  Luckily, a MailComposer object has a "mail" field that
  remembers the original options, so we can keep original behavior.
* However, we no longer support the mailComposer option set to a compiled
  MailComposer object (functionality that was briefly added in 1.2.0).
* nodemailer does SMTP URL parsing now automatically for us, simplifying code.
* Tests' outputs now end with additional "\r\n"
* Drop underscore package dependency (no longer needed)

* General formatting/style cleanup for `packages/email`.

* snake_cased => camelCased for some local variables.
* Added curly-brackets to `if`s.
* Removed trailing spaces.
* Removed commented-out code.
* Removed older doc text and changed some links.

* Get rid of back-and-forth assigning of `mailUrlString`.

### DIFF
--- a/packages/email/.npm/package/npm-shrinkwrap.json
+++ b/packages/email/.npm/package/npm-shrinkwrap.json
@@ -1,79 +1,14 @@
 {
   "dependencies": {
-    "addressparser": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/addressparser/-/addressparser-1.0.1.tgz",
-      "from": "addressparser@1.0.1"
-    },
-    "buildmail": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/buildmail/-/buildmail-4.0.1.tgz",
-      "from": "buildmail@4.0.1"
-    },
-    "httpntlm": {
-      "version": "1.6.1",
-      "resolved": "https://registry.npmjs.org/httpntlm/-/httpntlm-1.6.1.tgz",
-      "from": "httpntlm@1.6.1"
-    },
-    "httpreq": {
-      "version": "0.4.23",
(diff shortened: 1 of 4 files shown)


abernix:
Released as `email@1.2.1`!

PR #8610, opened 2017-04-17
Remove leading slash from paths in Babel options

### LINKED ISSUE #8566: Paths in Babel plugins are relative to project root but have a leading slash
Some Babel plugins, for example [babel-plugin-root-import](https://github.com/entwicklerstube/babel-plugin-root-import), use the path of the file they are processing (provided by the build tool) to resolve module paths. Using such plugins with Meteor produces unexpected results because the paths have a leading slash (which makes them absolute) although they are relative to the project root, e.g., `/client/module.js`.

See [this reproduction repo](https://github.com/klaussner/meteor-8566-reproduction).

~I think Meteor should provide "real" absolute paths instead (in this case, `/home/<project path>/client/module.js`).~ (Relative paths are working, too.)

 I put together a [comparison that shows which paths browserify, gulp, Rollup, and webpack are using](https://github.com/klaussner/babel-path-comparison).

Meteor version: 1.4.4-rc.6

klaussner:
Removes the leading slash from the `filename` and `sourceFileName` Babel options so that Babel plugins can distinguish relative paths from absolute paths.

I couldn't find out if the `filename` property is used somewhere in the Meteor build process but the `sourceFileName` is used in [tools/isobuild/bundler.js](https://github.com/meteor/meteor/blob/6a9f8b904425efce112103f3846d7a153f8988ca/tools/isobuild/bundler.js#L1245) where the removed slash is handled correctly.

Fixes #8566

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -86,8 +86,8 @@ BCp.processOneFileForTarget = function (inputFile, source) {
     babelOptions.sourceMap = true;
     babelOptions.filename =
       babelOptions.sourceFileName = packageName
-      ? "/packages/" + packageName + "/" + inputFilePath
-      : "/" + inputFilePath;
+      ? "packages/" + packageName + "/" + inputFilePath
+      : inputFilePath;
 
     babelOptions.sourceMapTarget = babelOptions.filename + ".map";
 


benjamn:
Do these paths get used in the source maps generated by Babel? I wonder if absolute vs. relative paths could be an issue there?

klaussner:
Yes, the `sourceFileName` is the path that will end up in the `sources` array of the source map, so that one *must* be relative.

benjamn:
I vaguely remember a bug when loading a source map on a page like https://example.com/nested/query/string. In the Dev Tools, paths without leading `/`s get interpreted relatively, e.g. `packages/whatever.js` would be displayed as `/nested/query/string/packages/whatever.js`, which is a little confusing.

Maybe we can just modify the source map `sources` strings to add back the leading `/`?

klaussner:
I think the source maps generated by `BabelCompiler` are never served directly to the client (or used for server-side debugging). Instead, `meteor://💻app` is prepended to the `sourceFileName` and the result is always `meteor://💻app/package/whatever.js`, regardless of whether there is a leading `/`.

Issue #105, opened 2012-04-27
Server time off on meteor.com deployed apps


mikebannister:
Seems like the time is off on meteor.com deployed apps, see:

http://time-test.meteor.com/

This creates a problem for apps that want to do 'time ago' display comparing a stamp received from the server and the current client time.

Peace, Mike

PR #8629, opened 2017-04-24
Add DDP._CurrentPublicationInvocation and DDP._CurrentMethodInvocation

zimme:
This pull-request is related to issue #8031. Please see [this comment](https://github.com/meteor/meteor/pull/8031#issuecomment-296564520) for more information.

Tests are still needed before we can merge this, but I thought I should try and get the ball rolling based on the clarification and discussion in #8031.

~~#8031 should probably be merged before this one if we decide to go down this road.~~

/cc @abernix, @mitar


zimme:
Added a test to make sure that methods called from a publication have access to the connection.

packages/ddp-server/livedata_server.js
```diff
@@ -1644,20 +1648,29 @@ _.extend(Server.prototype, {
...
     if (currentInvocation) {
       userId = currentInvocation.userId;
       setUserId = function(userId) {
         currentInvocation.setUserId(userId);
       };
       connection = currentInvocation.connection;
+      randomSeed = DDPCommon.makeRpcSeed(currentInvocation, name);
+    } else if (currentPublicationInvocation) {
+      userId = currentPublicationInvocation.userId;
+      setUserId = function(userId) {
```
mitar: Not sure if this should be set in this case?

zimme: How would we handle the case when a method wanna change the `userId` of the `connection` when the method was invoked by a `publication`?

Print the same kind of error we do when `setUserId` is invoked in a method initiated from server-code?

zimme: What would be the potential downsides of changing the connection's `userId` in this case?
Will this result in security flaws or something else maybe?

mitar: You are right. You are doing the same as what initial method invocation is doing.

mitar:
Can you make some tests which make sure `DDP._CurrentMethodInvocation`  and `DDP._CurrentPublicationInvocation` are never set at the same time?

packages/ddp-client/livedata_common.js
```diff
@@ -5,4 +5,5 @@ LivedataTest.SUPPORTED_DDP_VERSIONS = DDPCommon.SUPPORTED_DDP_VERSIONS;
 // This is private but it's used in a few places. accounts-base uses
 // it to get the current user. Meteor.setTimeout and friends clear
 // it. We can probably find a better way to factor this.
-DDP._CurrentInvocation = new Meteor.EnvironmentVariable;
+DDP._CurrentMethodInvocation = DDP._CurrentInvocation = new Meteor.EnvironmentVariable;
```
mitar: I would split this into two lines. One creating `DDP._CurrentMethodInvocation` and the other setting `DDP._CurrentInvocation` and commenting that this for backwards compatibility since X version because it was used as a public API by many packages.

We should maybe also change everywhere else in the core code to use `DDP._CurrentMethodInvocation`. But maybe somebody else from MDG could comment on this first.

zimme:
> Can you make some tests which make sure DDP._CurrentMethodInvocation and DDP._CurrentPublicationInvocation are never set at the same time?
In the case when a method is called from a publication both `DDP._CurrentMethodInvocation` and `DDP._CurrentPublicationInvocation` is actually set. In this case `DDP._CurrentMethodInvocation` is created from `DDP._CurrentPublicationInvocation`.

packages/ddp-client/livedata_common.js
zimme: Good point, will fix that.

mitar:
> In the case when a method is called from a publication both `DDP._CurrentMethodInvocation` and `DDP._CurrentPublicationInvocation` is actually set. In this case `DDP._CurrentMethodInvocation` is created from `DDP._CurrentPublicationInvocation`.

Oh, you are right. Before it was the same, only that `connection` was `null`. And now we populate it from `DDP._CurrentPublicationInvocation`. Sounds good.

mitar:
You should probably prepare a `HISTORY.md` entry as well. Because this could be a breaking change. How should we handle this? Probably it could be a part of next big release, 1.5?

zimme:
I'll hold off on working more in this until we get an "official" stamp of approval.

zimme:
> You should probably prepare a HISTORY.md entry as well. Because this could be a breaking change. How should we handle this? Probably it could be a part of next big release, 1.5?

I don't really see this as a breaking change as `DDP._CurrentInvocation` is still there in it's current form.

zimme:
Also I believe #8031 and this have competing tests. #8031 checks that `this.connection` is never set in a method called from a publication and this PR makes it available.

mitar:
It is a breaking change because now `connection` is set in some cases when it was not before.

mitar:
Yes, we should probably change those tests to make sure that `DDP._CurrentInvocation` is not set inside publish, and do not really check what is inside a method. And then this tests here can check that inside methods `connection` is available both when calling it first time and after user was set.

packages/accounts-base/accounts_server.js
```diff
@@ -77,9 +77,10 @@ export class AccountsServer extends AccountsCommon {
     // user expects. The way to make this work in a publish is to do
     // Meteor.find(this.userId).observe and recompute when the user
     // record changes.
-    var currentInvocation = DDP._CurrentInvocation.get();
+    var currentInvocation = DDP._CurrentPublicationInvocation.get();
+    currentInvocation = currentInvocation || DDP._CurrentInvocation.get();
     if (!currentInvocation)
-      throw new Error("Meteor.userId can only be invoked in method calls. Use this.userId in publish functions.");
+      throw new Error("Meteor.userId can only be invoked in method calls or publications.");
```
mitar: Can you also change "This function only works if called inside a method." comment above. And maybe change the rest of the comment to warn that if you use `Meteor.user()` (not `Meteor.userId()`) outside of the reactive context on the server the data returned by `Meteor.user()` will not change reactivelly. I mean, to me this is pretty obvious, so maybe we should just remove this comment.

packages/ddp-server/livedata_server.js
```diff
@@ -809,24 +809,33 @@ _.extend(Session.prototype, {
...
     self.userId = userId;
 
-    // Save the old named subs, and reset to having no subscriptions.
-    var oldNamedSubs = self._namedSubs;
-    self._namedSubs = {};
-    self._universalSubs = [];
+    // _setUserId is normally called from a Meteor method with
+    // DDP._CurrentInvocation set. But DDP._CurrentInvocation is not expected
+    // to be set inside a publish function, so we temporary unset it.
+    // Otherwise, this can lead to a problem that if a publish function is
```
mitar: I would remove the part of the comment from here on, because it is not anymore true. `connection` will now be always set in server-side methods, but we do not want `DDP._CurrentInvocation` to be set for those who are accessing it directly to check if they are inside a method call.

packages/ddp-server/livedata_server_tests.js
```diff
@@ -178,6 +187,75 @@ Tinytest.addAsync(
...
+});
+
+Meteor.methods({
+  livedata_server_test_setuserid: function (userId) {
+    this.setUserId(userId);
+  }
+});
+
+Tinytest.addAsync(
+  "livedata server - no connection in a method called from a publish function",
```
mitar: This should be probably renamed.

packages/ddp-server/livedata_server_tests.js
```diff
@@ -178,6 +187,75 @@ Tinytest.addAsync(
...
+  }
+);
+
+var methodCallResults = {};
+
+Meteor.publish("livedata_server_test_sub_with_method", function (connectionId) {
+  if (! methodCallResults[connectionId]) {
+    methodCallResults[connectionId] = [];
+  }
+  methodCallResults[connectionId].push(Meteor.call('livedata_server_test_inner'));
```
mitar: Here `DDP._CurrentInvocation.get()` shoud be pushed to `methodCallResults` now, no?

packages/accounts-base/accounts_server.js
zimme: I'll try and generalize this comment a bit as this holds true for methods as well, as far as I can tell.
i.e. `Meteor.user()` won't be reactive on server-side either in a method or a publication.

packages/ddp-server/livedata_server.js
zimme: Somewhere along the line I guess I merged #8031 and pushed to this PR which I didn't want to do. I'm thinking I'll rebase this branch of your PR and update the comments and tests to work with my changes. I don't really see these changes making sense without the changes in your PR.

packages/accounts-base/accounts_server.js
mitar: Yes. I think the argument here is that publications are longer active, so there is more chance that content returned by `user` changes.

BTW, I have a [package which gives server-side reactivity inside publish functions](https://github.com/peerlibrary/meteor-reactive-publish), so then `Meteor.user()` is reactive. :-)

packages/ddp-server/livedata_server.js
mitar: I think you do not have to rebase, you merged and we can merge this one together. And we can just close that other one. I do not mind.

zimme:
* Support `DDP._CurrentMethodInvocation`, `DDP._CurrentInvocation` is kept for backwards-compatibilty.
* Support `DDP._CurrentPublicationInvocation`, `DDP._CurrentMethodInvocation` is never set in a publication, even during re-run because a user logged in.
* Method calls invoked from publications now inherit the connection from the publication.
* `Meteor.userId()` and `Meteor.user()` is now available in publications too.

There is one breaking change here.
That is that `this.connection` is set in methods that are called from a publication. Before it was not set, except in a rare case when a publication was re-run because a user logged in.
For apps that currently checks if `this.connection` is available in a method to determine if the method was called from "secure" server code this change will make the apps "more secure" because now these methods called from publications will not run as "secure" server methods as they now have `this.connection`.

zimme:
I rebased this PR from #8031, so we should be able to just merge this PR and skip that on.

packages/accounts-base/accounts_server.js
```diff
@@ -68,18 +68,16 @@ export class AccountsServer extends AccountsCommon {
...
-    // record changes.
-    var currentInvocation = DDP._CurrentInvocation.get();
+    // This function only works if called inside a method or a pubication.
+    // Using any of the infomation from Meteor.user() in a method or
+    // publish function will always use the value from when the function first
+    // runs. This is likely not what the user expects. The way to make this work
+    // in a method or publish function is to do Meteor.find(this.userId).observe
+    // and recompute when the user record changes.
+    var currentInvocation = DDP._CurrentPublicationInvocation.get();
+    currentInvocation = currentInvocation || DDP._CurrentMethodInvocation.get();
```
abernix: Does it make sense to reverse the order of the fallback here so that `_CurrentMethodInvocation` falls back to `_CurrentPublicationInvocation` (instead of publication falling back to method) since `_CurrentMethodInvocation` is (temporarily) unset within the publication?

packages/ddp-server/livedata_server.js
```diff
@@ -1643,25 +1653,34 @@ _.extend(Server.prototype, {
...
+      setUserId = function(userId) {
+        currentMethodInvocation.setUserId(userId);
+      };
+      connection = currentMethodInvocation.connection;
+      randomSeed = DDPCommon.makeRpcSeed(currentMethodInvocation, name);
+    } else if (currentPublicationInvocation) {
+      userId = currentPublicationInvocation.userId;
       setUserId = function(userId) {
-        currentInvocation.setUserId(userId);
+        currentPublicationInvocation._session._setUserId(userId);
```
abernix: Is there another way to handle this?  Or should this be more defensive?  In other words, could `currentPublicationInvocation._session` ever be `undefined` here?

packages/accounts-base/accounts_server.js
zimme: I believe this would make sense as it's more likely that `DDP._CurrentMethodInvocation` would be set. If that isn't set, we're in a publish function and we fall back to `DDP._CurrentPublicationInvocation`.

So given the first comment I would change this to `const currentInvocation = DDP._CurrentMethodInvocation.get() || DDP._CurrentPublicationInvocation.get();`.

packages/ddp-server/livedata_server.js
zimme: From what I can tell a `Subscription` always have `_session` which is a `Session` and `_setUserId` on `Session` is what's used by a `methodInvocation` to set it's connections `userId`. We need this to be called properly so that publications are re-run when changing the `userId`.

I haven't found a different way of doing this, however maybe we could cc someone at MDG that knows more about the DDP protocol implementation to verify that `_session` is always available before merging?

zimme:
I have found myself writing `DDP._CurrentPublishInvocation` during the development of this.

Maybe this is an indicator that we should switch from `DDP._CurrentPublicationInvocation` to `DDP._CurrentPublishInvocation`. I do, however, like the former better as a description.

packages/ddp-server/livedata_server.js
mitar: I also looked into this because it was suspicious to me, but I could not find a cleaner way either. It seems the best and most stable one.

Maybe we should make a test which makes sure you call `setUserId` from inside a method called from publish function to make sure this will be always available. It would also in general be interesting to see what happens if you call `setUserId` from inside a publish function.

History.md
```diff
@@ -1,5 +1,15 @@
 ## v.NEXT
 
+* Fix issue with publications temporarily having `DDP._CurrentInvocation` set on
+  re-run after a user logged in. This would cause method calls from within
+  publish functions to unexpectedly having `this.connection` available.
+* Support `DDP._CurrentPublicationInvocation` and `DDP._CurrentMethodInvocation`
+  , `DDP._CurrentInvocation` is kept for backwards-compatibility. This change
```
mitar: Strange wrapping?

packages/ddp-server/livedata_server.js
zimme: from what I can tell there is no `setUserId` on the publication context, that's why we need to reach into `_session` to get that function. Calling `_session._setUserId` from inside a publish function should just change the current connections `userId` and the publication should re-run.

History.md
zimme: It was just a line length thing.

mitar: Yes, but comma could be in the previous line, no?

mitar:
I went through and it looks OK with my comments addressed, but I didn't do it as thoroughly as previously.

packages/ddp-server/livedata_server.js
mitar: Let's just make a test that `setUserId` can be called from a method called from a publish function. To catch if  `_session` will be renamed at any point in the future or anything like that.

zimme: I actually already do this. [Here](https://github.com/meteor/meteor/pull/8629/files#diff-ae375a4d35b797f447b7280ed27e1bcaR185).

packages/ddp-server/livedata_server_tests.js
```diff
@@ -84,11 +84,14 @@ testAsyncMulti(
...
 Meteor.methods({
   livedata_server_test_inner: function () {
-    return this.connection.id;
+    return this.connection && this.connection.id;
   },
 
   livedata_server_test_outer: function () {
     return Meteor.call('livedata_server_test_inner');
+  },
+  livedata_server_test_setuserid: function (userId) {
```
mitar: Probably a newline above that. Just a nitpick. ;-)

packages/ddp-server/livedata_server.js
mitar: You are great! I have further comments.

zimme:
There, I believe all the current comments have been addressed.

I'm not sure if anything else is needed to get this merged, but I don't think I'll have time to work a lot in this until next week again.

mitar:
I think we should add tests (and potential code improvements) which make sure that `DDP._CurrentPublicationInvocation` is set in all callbacks from publish functions, like `onStop`. I would even add a test to check (and make it so) so that it is set inside `added`, `changed`, `removed` callbacks from `observeChanges` callback from queries made inside publish functions.

In a way, you want `connection` to be available even in methods called from those callbacks.

zimme:
That's a valid point, I just recently fixed a bug in publIsh-composite
because DDP._CurrentInvocation and I had to bindEnvironment

zimme:
That's a valid point, I just recently fixed a bug in `reywood:publIsh-composite` [#105](https://github.com/englue/meteor-publish-composite/pull/105)
because `DDP._CurrentInvocation` wasn't available in the `observe` and `observeChanges` callbacks and I had to `bindEnvironment`.

However, I do believe this is another issue in itself. Becuase from what I can tell from working on that issue on this issue this is the way it currently work and to get these changes in quicker I would suggest we open a new issue for binding the environments of the callbacks.

zimme:
At least the callback for `this.onStop` is bound to the environment if it's run in a new Fiber.

mitar:
I do not mind doing that in the additional pull request.

mitar:
I also made a package that I should bind this new variable in [the reactive context in my reactive publish package](https://github.com/peerlibrary/meteor-reactive-publish/issues/30). I think quite some users who are patching publish should keep this in mind.

zimme:
I believe that we could `bind` and `bindEnvironment` for `this.onStop` callbacks in this PR though. because it's tied to publications.

`observe` and `observeChanges` isn't tied to publications in the same way, but it makes sense to `bindEnvironment` on those callbacks, but let's do that in a separate PR.

zimme:
> I think quite some users who are patching publish should keep this in mind.

I know a few packages that does this, I'll open a few issues on those repos once this is merged.

mitar:
I have this [package to help them patch publish](https://github.com/peerlibrary/meteor-extend-publish), but I am not sure if this require any changes to it. It really requires patching only if you are doing something async.

mitar:
Is `bindEnvironment` enough for maintaining the state of `DDP._CurrentPublicationInvocation`?

mitar:
BTW, I think it is enough to bind `observeChanges` because `observe`'s callbacks are called from it, no?

zimme:
> Is bindEnvironment enough for maintaining the state of DDP._CurrentPublicationInvocation?

Yes, as far as I can tell `bindEnvironment` should be enough, and I believe you're already calling the wrapped function with the context set properly.

zimme:
> BTW, I think it is enough to bind observeChanges because observe's callbacks are called from it, no?

I haven't looked at the implementation of the `observe`/`observeChanges` functions, so I don't know. I just know that I had to `bindEnvironment` `observe` in `publish-composite`. but if we do it in core binding `observeChanges` might be enough if `observe` uses `observeChanges` internally.

zimme:
I included a fix to call the `onStop` callbacks with the publish context and bind the environment when it's not called straight away because the subscription was already deactivated.

packages/ddp-server/livedata_server.js
```diff
@@ -1224,9 +1224,9 @@ _.extend(Subscription.prototype, {
   onStop: function (callback) {
     var self = this;
     if (self._isDeactivated())
-      callback();
+      callback.call(self);
     else
-      self._stopCallbacks.push(callback);
+      self._stopCallbacks.push(Meteor.bindEnvironment(callback.bind(self)));
```
mitar: Why don't you just do `callback = Meteor.bindEnvironment(callback.bind(self))` at the beginning?

Not sure why you differentiate between deactivated or not? `Meteor.bindEnvironment` is smart enough to handle if it is not needed.

BTW, `Meteor.bindEnvironment` accepts `this` as the last argument. So you can use that instead of using `callback.bind`.

packages/ddp-server/livedata_server_tests.js
```diff
@@ -176,9 +176,28 @@ Meteor.publish("livedata_server_test_sub_context", function (connectionId, userI
...
+  // Check that onStop callback is have the same context as the publish function
+  // and that it runs with the same environment variables as this publis
+  // function.
+  this.onStop(function () {
+    var onStopMethodInvocation = DDP._CurrentMethodInvocation.get();
+    var onStopPublicationInvocation = DDP._CurrentPublicationInvocation.get();
+    callback.call(this, onStopMethodInvocation, onStopPublicationInvocation);
+  });
+
+  // Check that onStop arrow callback runs with the same environment variables
```
mitar: Why would an arrow callback be different? I think this test checks a basic JavaScript behavior. I do not think it is needed.

packages/ddp-server/livedata_server.js
zimme: I did not know that it took context as the last argument, I'll use that instead of `bind`.
I didn't know it was smart enough to figure that out so I took the safer option, but I'll up that and skip the `call`/`bind` :+1:

packages/ddp-server/livedata_server_tests.js
zimme: I guess I just wanted to make sure the environment was bound properly with arrow functions too. but I guess it's pretty redundant.

mitar: I mean the arrow function is safer. If it works without, it will also work with.

packages/ddp-server/livedata_server.js
mitar: Check the source code of `Meteor.bindEnvironment` to make sure. :-)

zimme:
There, now ´this.onStop` callbacks should be running with the proper context and the environment variables from the publish function should be preserved for the callbacks too.

Now I'm off to put together some slides for an Apollo talk I'm doing tomorrow. :smile:

packages/ddp-server/livedata_server.js
zimme: `Meteor.bindEnvironment(func, 'description'/onExceptionFunc, this);` :+1:

packages/ddp-server/livedata_server_tests.js
```diff
@@ -176,9 +176,21 @@ Meteor.publish("livedata_server_test_sub_context", function (connectionId, userI
...
   var methodInvocation = DDP._CurrentMethodInvocation.get();
   var publicationInvocation = DDP._CurrentPublicationInvocation.get();
+
+  // Check the publish function's environment variables and context.
   if (callback) {
     callback.call(this, methodInvocation, publicationInvocation);
   }
+
+  // Check that onStop callback is have the same context as the publish function
+  // and that it runs with the same environment variables as this publis
```
mitar: `publis`, typo.

hwillson:
Hi guys - it looks like we're ready to get this merged? There was a lot of back and forth since our last review, but it looks like all additional concerns have now been addressed - is there anything outstanding, or are we good to go?

zimme:
There are a few small things left to do.

> though I do think we need a very clear call-to-action

We need to update `History.md` with information about how to handle the breaking change.
i.e. if you call a method from a publication and now your method is not doing what you're expecting because you have a check for `this.connection` which now will be set, whereas before it wasn't set (in almost all cases, the bug @mitar fixed in #8031). I can try and write something about this new behaviour and that code using this pattern needs to be updated to work as expected.

I'll also add a line about `this.onStop` callbacks now being run with the publication's context and with it's environment variables preserved.

Other than that I believe this PR is complete.

hwillson:
If you don't mind prepping the `History.md` changes, that would be awesome @zimme - thanks!

zimme:
@hwillson, I've updated `History.md`. Please have a look and change it as you guys see fit.

History.md
```diff
@@ -5,6 +5,26 @@
...
+    > Note: If you're calling methods from publications that are using `this.connection`
+    > to see if the method was called from server code or not. These checks will now
+    > be more restrictive because `this.connection` will now be available when a
+    > method is called from a publication.
+
+* `Meteor.userId()` and `Meteor.user()` can now be used in both method calls and
+  publications.
+
+* `this.onStop` callbacks in publications are now run with the publication's
+  context and with it's `EnvironmentVariable`'s bound.
```
hwillson: This looks great @zimme  - just one small grammar change; "it's" here should be "its".

hwillson:
Thanks again for all of your work on this @zimme and @mitar - it looks like we're all set here, so LGTM!

zimme:
Typo fixed =)

zimme:
There seems to be quite a lot of problems with Circle CI.

benjamn:
@zimme I fixed a lot of the Circle CI flakiness on the `release-1.5` branch just recently. If you're planning to push any more commits to this branch, maybe pull those fixes in first?
```sh
git cherry-pick 2032ba8b23020758fef43c78de83a36230da2c36
git cherry-pick 2e057d711914b432ab9cb6a2b07d44f55815dd47
git cherry-pick f339dec5ae02c2ef12679d5edcf0a623ff70641f
```

zimme:
@benjamn, Nice! also, good to know. I do, however, don't plan on pushing more on this branch as I believe we're finished with this work now. =)

zimme:
I realized that we haven't bumped the versions of the packages this PR touches. Thinking that `ddp-server` might need a major version bump as this brings a potential breaking change for some developers.

laosb:
Shall we update documents to use Meteor.user() / .userId() in publications?

zimme:
Where in the docs are you talking about?
The @locus was updated in this PR which should change the docs as to where you can use `Meteor.userId()`/`Meteor.user()`.

laosb:
I mean, for example, http://docs.meteor.com/api/pubsub.html#Subscription-userId and some other places use `this.userId` in examples.

The real problem is, should we recommend `Meteor.userId()`/`Meteor.user()` in publish functions or just keep it as is while you can now use `Meteor.userId()`/`Meteor.user()` as an alternative?

zimme:
@laosb, I'm thinking that we shouldn't recommend `Meteor.user()`/`Meteor.user()` in publications as developers might expect those to be reactive but in reality they're just re-run if a user logs in or out.

I don't think it belongs in the Subscriptions section and looking through the Methods section there is no mention of `Meteor.userId()` or `Meteor.user()` that I can find. 

But maybe we should make a note about all of this in the `Accounts` section which documents `Meteor.userId()`/`Meteor.user()`? i.e. you can use `Meteor.userId()`/`Meteor.user()` inside publications and methods, BUT think about this bla bla.

mitar:
BTW, how does one now know if a method is called from an existing server-side method, or directly from a client side?

zimme:
With the new code one would check if `DDP._CurrentPublicationInvocation.get()` returns `null`, if so it's invoked from the client, otherwise wi would have been invoked from a publication.

zimme:
Ooh, I misread that quite a lot. How would you do it before? I guess that would still hold true?

zimme:
One way would be adding an extra `argument` to the method, which by default is false and then you set it when you call the method from another method.

mitar:
Before you checked that `this.connection` is true.

The use case is having a Meteor method which can have a more permissive API if called from the server side. For example, you can create a document without linking it to the current user (because you are calling it from the server side and you know that you will in the next step assign custom user as an author).

So I can have Meteor method `createDocument(authorId)` and when called from client-side this `authorId` has to match `this.userId` but if called from server side it can be any `authorId`.

Maybe the solution is to refactor this code and move most of the method into an importable function, and then inside Meteor method just check if `authorId` matches `this.userId`.

zimme:
If a method is called from server-side code it won't have `this.connection`, however if it's called from another method which was invoked by a client, then it will have `this.connection` which should be seen as being invoked by the client directly, right?

I would go with the latter suggestion.

mitar:
Yes, before calling a method from a method called from the client side behaved like a server-side call.

abernix:
I guess we should probably open some PRs with the Docs and Guide to make it clear that it's no longer necessary to use `this.userId`?  e.g. [here](https://guide.meteor.com/data-loading.html#publications-and-subscriptions), etc..

Issue #8631, opened 2017-04-25
Collection.upsert with modifier does not work correctly


sebakerckhof:
I found a bug in the minimongo upsert behavior. Meteor Version: `1.4.3.1`

Assuming you have an empty collection called 'myCollection'.
Running this command using the mongo shell works fine and inserts the document:
```js
db.getCollection('myCollection').update({"user._id":"123"},{"$setOnInsert":{"user":{"_id":"123","name":"foos"}}},{upsert:true})
```

Running a similar upsert in minimongo, e.g.
```js
myCollection.update({"user._id":"123"},{"$setOnInsert":{"user":{"_id":"123","name":"foos"}}},{upsert: true});
```

Results in:
`MinimongoError: Key user._id must not contain '.' [409]`
This originates from: https://github.com/meteor/meteor/blob/9bd5194fa417a48f689bed060ac934bcdd16a903/packages/minimongo/minimongo.js#L563

So I guess Minimongo does not merge the select data with the update statement correctly when that statement contains `$setOnInsert`, `$set` and possibly others.

PR #8633, opened 2017-04-25
Remove nested properties from upsert selector document

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/8631

### DIFF
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -783,9 +783,21 @@ LocalCollection.prototype.update = function (selector, mod, options, callback) {
   var insertedId;
   if (updateCount === 0 && options.upsert) {
     var newDoc = {};
-    var selectorMod = {$set: LocalCollection._removeDollarOperators(selector)};
-    LocalCollection._modify(newDoc, selectorMod, {isInsert: true});
+     
+    if (LocalCollection._selectorIsId(selector)) {
+      selector = {_id: selector}
+    }
+
+    var selectorDocument = LocalCollection._removeDollarOperators(selector);
+
+    if (selectorDocument._id) {
+      newDoc._id = selectorDocument._id;
+      delete selectorDocument._id;
+    }
+
+    LocalCollection._modify(newDoc, {$set: selectorDocument});
     LocalCollection._modify(newDoc, mod, {isInsert: true});
+
     if (! newDoc._id && options.insertedId)
       newDoc._id = options.insertedId;
     insertedId = self.insert(newDoc);
--- a/packages/minimongo/minimongo.js
+++ b/packages/minimongo/minimongo.js
@@ -782,20 +782,22 @@ LocalCollection.prototype.update = function (selector, mod, options, callback) {
   // generate an id for it.
   var insertedId;
   if (updateCount === 0 && options.upsert) {
-    var newDoc = {};
-     
-    if (LocalCollection._selectorIsId(selector)) {
-      selector = {_id: selector}
+
+    let selectorModifier = LocalCollection._selectorIsId(selector) 
+      ? { _id: selector } 
+      : selector;
+
+    selectorModifier = LocalCollection._removeDollarOperators(selectorModifier);
+
+    const newDoc = {};
+    if (selectorModifier._id) {
+      newDoc._id = selectorModifier._id;
+      delete selectorModifier._id;
     }
 
-    var selectorDocument = LocalCollection._removeDollarOperators(selector);
-
-    if (selectorDocument._id) {
-      newDoc._id = selectorDocument._id;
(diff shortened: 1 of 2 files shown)


hwillson:
Hi @sebakerckhof - thanks for submitting this! I've restarted the CircleCI build since its failure wasn't related to your changes, but the Travis test failures [do look related](https://travis-ci.org/meteor/meteor/builds/225610465#L921). If you're able to take a look, that would be great. I'm happy to review this right after the tests are passing. Thanks again!

sebakerckhof:
Woops, I only tested the minimongo package itself. All tests seem to pass now with the latest fix and I've added some new testcases in the minimongo package.

hwillson:
Thanks @sebakerckhof - would you mind rebasing?

sebakerckhof:
@hwillson done.

packages/minimongo/minimongo.js
```diff
@@ -782,8 +782,22 @@ LocalCollection.prototype.update = function (selector, mod, options, callback) {
   // generate an id for it.
   var insertedId;
   if (updateCount === 0 && options.upsert) {
-    var newDoc = LocalCollection._removeDollarOperators(selector);
+    var newDoc = {};
+     
+    if (LocalCollection._selectorIsId(selector)) {
+      selector = {_id: selector}
```
hwillson: Can we avoid re-assigning to the `selector`, and instead use a new variable to represent the modified selector?

packages/minimongo/minimongo.js
```diff
@@ -782,8 +782,22 @@ LocalCollection.prototype.update = function (selector, mod, options, callback) {
...
+    }
+
+    var selectorDocument = LocalCollection._removeDollarOperators(selector);
```
hwillson: Actually, we're not really building a selector at this point anymore, we're building a new upsert modifier. So maybe we could instead use a new variable called something like `upsertModifier`, to make our intent here more clear? So we could consider changing the code here to something like:

```js
...
let upsertModifier =
  LocalCollection._selectorIsId(selector) ? { _id: selector } : selector;
upsertModifier = LocalCollection._removeDollarOperators(upsertModifier);

const newDoc = {};
if (upsertModifier._id) {
  newDoc._id = upsertModifier._id;
  delete upsertModifier._id;
}

LocalCollection._modify(newDoc, {$set: upsertModifier});
LocalCollection._modify(newDoc, mod, {isInsert: true});
...
```

packages/minimongo/minimongo.js
hwillson: Can we avoid re-assigning to the `selector`, and instead use a new variable to represent the modified selector?

packages/minimongo/minimongo.js
hwillson: Actually, we're not really building a selector at this point anymore, we're building a new upsert modifier. So maybe we could instead use a new variable called something like `upsertModifier`, to make our intent here more clear? So we could consider changing the code here to something like:

```js
...
let upsertModifier =
  LocalCollection._selectorIsId(selector) ? { _id: selector } : selector;
upsertModifier = LocalCollection._removeDollarOperators(upsertModifier);

const newDoc = {};
if (upsertModifier._id) {
  newDoc._id = upsertModifier._id;
  delete upsertModifier._id;
}

LocalCollection._modify(newDoc, {$set: upsertModifier});
LocalCollection._modify(newDoc, mod, {isInsert: true});
...
```

packages/minimongo/minimongo.js
```diff
@@ -782,8 +782,22 @@ LocalCollection.prototype.update = function (selector, mod, options, callback) {
...
+
+    if (selectorDocument._id) {
+      newDoc._id = selectorDocument._id;
+      delete selectorDocument._id;
+    }
+
+    LocalCollection._modify(newDoc, {$set: selectorDocument});
```
hwillson: This double `_modify` call initially through me off. I get why you're making it, and it works, but I think we should add a comment here explaning things a bit for future maintainers. Something along the lines of:

```js
// This double _modify call is made to help work around an issue where collection 
// upserts won't work properly, with nested modifiers (see issue #8631).
```

packages/minimongo/minimongo.js
sebakerckhof: Yeah, I was thinking the same thing, but `upsertModifier` would have me think this is actually what the `mod` variable already is.

hwillson: Good points; naming things is hard! Okay, I'm good with your suggestion of `selectorModifier` or `upsertModifier`. I agree that `selectorModifier` seems a bit strange, but since that's really what it is, that might be best.

hwillson:
This all looks great @sebakerckhof  - LGTM!

Issue #8630, opened 2017-04-25
Memory Leak in Node 4.8.1


mjmasn:
Node 4.8.1 has a memory leak (https://github.com/nodejs/node/pull/12089) that was fixed in 4.8.2. 

If anyone is experiencing memory leaks in production after updating Node recently this is likely to be the cause. The solution is to update the version of Node on your servers to 4.8.2.

This isn't really a critical issue for the Meteor tool itself as the leak is unlikely to cause issues in development but it may be worth bumping the version anyway (and changing the recommendation to use 4.8.2 in production).

hwillson:
PR #8634 has been merged, so Node 4.8.2 will be coming soon. Thanks for flagging this @mjmasn!

abernix:
This should be available to test in the release candidate for Meteor 1.4.4.2, currently `1.4.4.2-rc.1`:

    meteor update --release 1.4.4.2-rc.1

Please report any problems you encounter!

tab00:
My app in production has been restarting multiple times per day with "out of memory" error, so I hope this gets through soon.

Thank you guys for your work on solving this.

dagatsoin:
@abernix could this solve memory leaks issue on Galaxy?

I have an app which restart each few minutes because memory increases over the time. And it it not the case on my local machine.

abernix:
@dagatsoin Galaxy uses the exact version of Node.js which your local development uses, so yes.  And additionally, there was a memory leak in Node.js itself, so if it was triggered the answer is: absolutely.  😉

I'd recommend upgrading to Meteor 1.4.4.2 and seeing if it goes away for you!  If it does, please do report back.  If it doesn't, then you've likely encountered or created some other kind of memory leak within your app which is unrelated.

dagatsoin:
You made my day!!

Le 4 mai 2017 à 20:53, Jesse Rosenberger <notifications@github.com<mailto:notifications@github.com>> a écrit :


@dagatsoin<https://github.com/dagatsoin> Galaxy uses the exact version of Node.js which your local development uses, so yes. And additionally, there was a memory leak in Node.js itself, so if it was triggered the answer is: absolutely. 😉

I'd recommend upgrading to Meteor 1.4.4.2 and seeing if it goes away for you! If it does, please do report back. If it doesn't, then you've likely encountered or created some other kind of memory leak within your app which is unrelated.

PR #8634, opened 2017-04-25
Upgrade Node to 4.8.2 and related core npm packages.

hwillson:
Hi all - this PR is intended to resolve issue #8630. Node 4.8.2 only contains a few commits (https://nodejs.org/en/blog/release/v4.8.2/ - the memory leak fix and 2 small dependency updates), so upgrading to get the memory leak fix should be quite safe. 

One quick note - this is the first time I've submitted a PR that includes `dev_bundle` changes. I believe I've ticked all of the boxes, but please let me know if I've missed anything. Thanks!


hwillson:
The `self-test`'s and package tests all passed locally, so I'm probably missing the special incantation needed to get Travis and Circle to recognize that I'm testing against a new `dev_bundle` ...

abernix:
Thanks for submitting this, @hwillson.

I built `dev_bundle` version 4.7.26 and upgraded a couple other core package (`npm`, `node-inspector`, and `v8-profiler`) and pushed to this PR to re-trigger the build.  We're green now, so LGTM.

Also, just to clarify, the way you've done things here is fine.  You're always welcome to make changes, push the PR and then ping someone to build a dev bundle for you when in need.  Otherwise, there is currently no way for a contributor outside of MDG to update the dev bundle dependency and rebuild the dev bundle (into a tarball) in a way that CircleCI tests will be able to fetch.  While it would be possible to have the CircleCI build the dev bundle and test against it, it is not a quick build and would slow down the CI process and currently we need to publish the tarball to S3 through internal processes anyway.

hwillson:
Great @abernix, thanks for building the `dev_bundle` and the detailed explanation! I'll aim to summarize some of these details and get them into [Development.md](https://github.com/meteor/meteor/blob/devel/Development.md).

PR #8643, opened 2017-04-27
fix bug when initiating dict with non string values

ziedmahdi:
```js
var b = new ReactiveDict({b: true});
b.get('b') //throw (Error: EJSON.parse argument should be a string)
```
This is because `dict.set` stringify the value before settings it, which is not done when u initiate the dict with initial data.

### DIFF
--- a/packages/reactive-dict/reactive-dict.js
+++ b/packages/reactive-dict/reactive-dict.js
@@ -27,7 +27,10 @@ ReactiveDict = function (dictName) {
       this.name = dictName;
     } else if (typeof dictName === 'object') {
       // back-compat case: dictName is actually migrationData
-      this.keys = dictName;
+      this.keys = {};
+      for (let [key, value] of Object.entries(dictName)) {
+        this.keys[key] = stringify(value);
+      }
     } else {
       throw new Error("Invalid ReactiveDict argument: " + dictName);
     }
--- a/packages/reactive-dict/reactive-dict-tests.js
+++ b/packages/reactive-dict/reactive-dict-tests.js
@@ -6,6 +6,16 @@ Tinytest.add('ReactiveDict - set to undefined', function (test) {
   test.equal(dict.get('foo'), undefined);
 });
 
+Tinytest.add('ReactiveDict - initialize with data', function (test) {
+  var now = new Date();
+  var dict = new ReactiveDict({
+    now: now
+  });
+  
+  var nowFromDict = dict.get('now');
+  test.equal(nowFromDict, now);
+});
+
 Tinytest.add('ReactiveDict - setDefault', function (test) {
   var dict = new ReactiveDict;
   dict.set('A', 'blah');


hwillson:
Hi @ziedmahdi - first off, thanks for submitting a pull request to Meteor! Your time/effort is greatly appreciated. Just a quick future FYI - when you notice a bug with Meteor, or have an idea for a feature request, we prefer that you first create a new issue to help outline what the problem is (ideally with a reproduction), and give people a chance to vet the issue before creating a PR. This helps reduce the chances of people doing a bunch of work on a PR, only to find out that it won't/can't be accepted for some specific reason. For more details about this, please see the [Submitting pull requests](https://github.com/meteor/meteor/blob/devel/Contributing.md#submitting-pull-requests) section of the contributing guide.

The above being said, I've taken a look at what you're proposing. I see what you mean - `ReactiveDict`'s `set` function makes the necessary `stringify` call, but passing values in through the constructor does not. While your change does help address this, I think it's important to note that the fact that the `ReactiveDict` constructor accepts an object at all is a result of a backwards compatibility issue with Meteor 0.9.1 and earlier. Initializing a `ReactiveDict` through the constructor is kind of a hack, and is clearly a bit of a mess right now. For example, the fact that the only allowed constructor parameter is called `dictName`, but yet can represent either the name to set for the dictionary OR the initial values to set for the dictionary (so 2 completely different intentions), is quite confusing.

Putting that aside though, I think your fix is good as is for now. Someone who wants to revisit this in the future might want to consider changing things around a bit. I think for starters adjusting the `ReactiveDict` constructor to take 2 parameters (`dictName` and `dictData`) would help. This would give us 2 main advantages:

1. We'll be able to both name a `ReactiveDict` and initialize it, at the same time, via the constructor (whereas right now we can only do one or the other).
2. Splitting the initial data out into it's own parameter makes the API more clear.

So I think something like this could work (covers both legacy support and a newer, more clear, API):

```js
ReactiveDict = function (dictName, dictData) {

  function cleanData(uncleanData) {
    var cleanedData = {};
    for (var key in uncleanData) {
      cleanedData[key] = stringify(uncleanData[key]);
    }
    return cleanData;
  }

  if (dictName) {
    if (typeof dictName === 'string') {
      // the normal case, argument is a string name.
      // _registerDictForMigrate will throw an error on duplicate name.
      ReactiveDict._registerDictForMigrate(dictName, this);
      this.keys = ReactiveDict._loadMigratedDict(dictName) || dictData || {};
      this.name = dictName;
    } else if (typeof dictName === 'object') {
      // back-compat case: dictName is actually migrationData
      this.keys = cleanData(dictName);
    } else {
      throw new Error("Invalid ReactiveDict argument: " + dictName);
    }
  }

  if (!this.keys && dictData) {
    this.keys = cleanData(dictData);
  }

  if (!this.keys) {
    // no name given; no migration will be performed
    this.keys = {};
  }

  this.allDeps = new Tracker.Dependency;
  this.keyDeps = {}; // key -> Dependency
  this.keyValueDeps = {}; // key -> Dependency
};
```

Again, I'm not saying you should implement this. This is the type of discussion we would usually have in the issue, but since there isn't one associated with this PR, I've added my thoughts here. I do have a couple of specific code recommendations for you (like leveraging ES2015 capabilities since `ReactiveDict` has a dependency on the `ecmascript` package), which I'll add via a proper review shortly. Thanks!

ziedmahdi:
For ES2015, I didn't check `ReactiveDict` dependencies. But I found the previous code not using it so I followed the same style.

ziedmahdi:
@hwillson something like your code should work, maybe some new tests will confirm it.

The `stringify` may be confusing in some case like:
```js
var b = new ReactiveDict({b: 'true'});
b.get('b')
true //not a string
```

packages/reactive-dict/reactive-dict.js
```diff
@@ -27,7 +27,10 @@ ReactiveDict = function (dictName) {
       this.name = dictName;
     } else if (typeof dictName === 'object') {
       // back-compat case: dictName is actually migrationData
-      this.keys = dictName;
+      this.keys = {};
+      for (var key in dictName) {
```
hwillson: We should probably check to make sure the iteration over `dictName` is only including the objects own properties. So adding in a `hasOwnProperty` check, or leveraging `Object.entries` like:

```
for (let [key, value] of Object.entries(dictName)) { 
  this.keys[key] = stringify(value);
}
```

hwillson:
Thanks @ziedmahdi, as soon as the tests re-run and pass (I restarted them - Circle crashed but it wasn't related to your changes), we'll be all set to get this merged!

ziedmahdi:
Thanks @hwillson

zimme:
@hwillson, I actually like your idea with being able to set the `dictName` and `dictData` at the same time while actually still keeping backwards-compatibility. I don't really see a reason for us to not do this.

I'll put together a PR for that tomorrow unless someone says otherwise.

hwillson:
That would be awesome @zimme - thanks!

zimme:
@hwillson, @abernix, what should take precedence? `dictData` or `ReactiveDict._loadMigratedDict(dictName)`.
```js
const state = new ReactiveDict('applicationState', { currentRoute: 'home' });
```
When the above code runs again after a reload it feels like the migrated data should take precedence over the "initial" data. I just wanted to get this out there so I don't miss anything.

hwillson:
I think you're right @zimme - the migrated data should take precedence. Thanks!

zimme:
I'm working on this now an in doing so I realized that if we just use `this._setObject` in the constructor we don't have to add new `stringify` logic to the constructor. This is the approach I'm using now that I'm adding support for `dictData`. Also, I'm re-writing `ReactiveDict` as a class as ecmascript is used in this package.

packages/reactive-dict/reactive-dict.js
```diff
@@ -27,7 +27,10 @@ ReactiveDict = function (dictName) {
       this.name = dictName;
     } else if (typeof dictName === 'object') {
       // back-compat case: dictName is actually migrationData
-      this.keys = dictName;
+      this.keys = {};
+      for (let [key, value] of Object.entries(dictName)) {
+        this.keys[key] = stringify(value);
+      }
```
benjamn: This definitely works, but the more conventional/ES5 style we tend to use is:
```js
Object.keys(dictName).forEach(key => {
  this.keys[key] = stringify(dictName[key]);
});
```

zimme:
@benjamn, @hwillson, When working on supporting initial data on `ReactiveDict` when you also provide a name. I've come to realize that DDP relies on the previous behaviour.

When I was running some tests and changed code while the tests where running (meteor reloads) the method `tinytest/run` will crash with a `check` error because the methods parameters has been stringified.

I'm not sure if this impacts all methods being re-run on a reload and all their arguments being stringified.

PR #8654, opened 2017-04-30
Add support for initial data alongside a name for ReactiveDict

zimme:
As per the discussion in #8643 I've changed the implementation of `ReactiveDict` to support setting initial data when creating a `ReactiveDict` with a name.
```js
const state = new ReactiveDict('applicationState', { currentRoute: 'home' });
```
This PR also fixes the issue brought up in #8643 by using `_setObject` which uses `set` to set the data and make sure it's stringified.

I also took the liberty to re-write `ReactiveDict` as an es2015 class.

/cc @hwillson, @abernix

### DIFF
--- a/packages/reactive-dict/reactive-dict.js
+++ b/packages/reactive-dict/reactive-dict.js
@@ -1,120 +1,129 @@
 // XXX come up with a serialization method which canonicalizes object key
 // order, which would allow us to use objects as values for equals.
-var stringify = function (value) {
-  if (value === undefined)
+function stringify(value) {
+  if (value === undefined) {
     return 'undefined';
+  }
   return EJSON.stringify(value);
-};
-var parse = function (serialized) {
-  if (serialized === undefined || serialized === 'undefined')
-    return undefined;
-  return EJSON.parse(serialized);
-};
+}
 
-var changed = function (v) {
+function parse(serialized) {
+  if (serialized === undefined || serialized === 'undefined') {
+    return undefined;
+  }
+  return EJSON.parse(serialized);
+}
+
+function changed(v) {
   v && v.changed();
-};
+}
 
 // XXX COMPAT WITH 0.9.1 : accept migrationData instead of dictName
-ReactiveDict = function (dictName) {
-  // this.keys: key -> value
-  if (dictName) {
-    if (typeof dictName === 'string') {
-      // the normal case, argument is a string name.
-      // _registerDictForMigrate will throw an error on duplicate name.
-      ReactiveDict._registerDictForMigrate(dictName, this);
-      this.keys = ReactiveDict._loadMigratedDict(dictName) || {};
-      this.name = dictName;
-    } else if (typeof dictName === 'object') {
-      // back-compat case: dictName is actually migrationData
-      this.keys = {};
-      for (let [key, value] of Object.entries(dictName)) {
-        this.keys[key] = stringify(value);
-      }
-    } else {
-      throw new Error("Invalid ReactiveDict argument: " + dictName);
-    }
-  } else {
-    // no name given; no migration will be performed
+export class ReactiveDict {
+  constructor(dictName, dictData) {
+    // this.keys: key -> value
     this.keys = {};
+
+    if (dictName) {
+      // name given; migration will be performed
+      if (typeof dictName === 'string') {
+        // the normal case, argument is a string name.
+
+        // Only run migration logic on client, it will cause
+        // duplicate name errors on server during reloads.
+        // _registerDictForMigrate will throw an error on duplicate name.
+        Meteor.isClient && ReactiveDict._registerDictForMigrate(dictName, this);
+        const migratedData = Meteor.isClient && ReactiveDict._loadMigratedDict(dictName);
+
+        if (migratedData) {
+          // Don't stringify migrated data
+          this.keys = migratedData;
+        } else {
+          // Use _setObject to make sure values are stringified
+          this._setObject(dictData || {});
+        }
+        this.name = dictName;
+      } else if (typeof dictName === 'object') {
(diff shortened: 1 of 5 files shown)


packages/reactive-dict/reactive-dict.js
```diff
@@ -1,117 +1,114 @@
...
+        // the normal case, argument is a string name.
+        // _registerDictForMigrate will throw an error on duplicate name.
+        ReactiveDict._registerDictForMigrate(dictName, this);
+        // Use _setObject to make sure values are stringified
+        this._setObject(ReactiveDict._loadMigratedDict(dictName) || dictData);
+        this.name = dictName;
+      } else if (typeof dictName === 'object') {
+        // back-compat case: dictName is actually migrationData
+        // Use _setObject to make sure values are stringified
+        this._setObject(dictName);
```
hwillson: I think calling `this._setObject` here makes sense, but I think there is a bug lingering here. If we initialize a `ReactiveDict` like `new ReactiveDict('name', { property: 'some value' })`, we'll hit this `_setObject` call. So far so good, but down in the `set` function (which we end up at via the `_setObject` call), we hit this block:

```js
    if (isNewValue) {
      changed(this.keyDeps[key]);
      if (this.keyValueDeps[key]) {
        changed(this.keyValueDeps[key][oldSerializedValue]);
        changed(this.keyValueDeps[key][value]);
      }
    }
```

The problem is the second line will fail, as `this.keyDeps` is `undefined`. We haven't initialized it yet in the constructor (it's initialized after the `_setObject` call). So we'll want to make sure it's initialized beforehand (and maybe add a test to catch this as well).

packages/reactive-dict/reactive-dict.js
```diff
@@ -1,117 +1,114 @@
...
+      } else {
+        throw new Error("Invalid ReactiveDict argument: " + dictName);
+      }
+    }
 
-_.extend(ReactiveDict.prototype, {
+    this.allDeps = new Tracker.Dependency;
+    this.keyDeps = {}; // key -> Dependency
+    this.keyValueDeps = {}; // key -> Dependency
+  }
```
hwillson: Super nit-picky, but can we add a newline here to separate the constructor from the `set` comment?

packages/reactive-dict/reactive-dict.js
```diff
@@ -1,117 +1,114 @@
...
+  constructor(dictName, dictData = {}) {
+    // this.keys: key -> value
     this.keys = {};
-  }
 
-  this.allDeps = new Tracker.Dependency;
-  this.keyDeps = {}; // key -> Dependency
-  this.keyValueDeps = {}; // key -> Dependency
-};
+    if (dictName) {
```
hwillson: It doesn’t look like we can have an un-named `ReactiveDict`, and still initialize it with default data via `dictData`. Do you think we need to be able to do this? I'm envisioning that some people might want to call something like `new ReactiveDict(null, { property: 'some value' })` to initialize the `ReactiveDict` but keep it un-named, so no migration is ever done. Maybe this is an edge case, but let me know what you think.

packages/reactive-dict/reactive-dict.js
zimme: I actually had guards here but somewhere they got lost before I created the PR. I'll fix this. I wanted the dependencies to be undefined here when this function is run from constructor so that no unnecessary invalidations are triggered when we set a bunch of initial values.

packages/reactive-dict/reactive-dict.js
zimme: I'll look at this case and see if I can find something that would argue against supporting this scenario.

zimme: I've looked into this a bit and as we have support for `new ReactiveDict({ property: 'some value'});` I think we should throw an error for this `new ReactiveDict(null/undefined, { property: 'some value' });`.

Even though `new ReactiveDict({ property: 'some value' });` was there initially to support some old kind of migration data it isn't really tied to this in some way.

However thinking about this case.
```js
let dictName;

const dict = new ReactiveDict(dictName, { property: 'some value' });
```
In this case the user might be surprised that the dict isn't migrated on reload because we created an unnamed dict instead.

However, taking into account that we don't care about this in the case when we don't provide initial data it might be a bad idea.

packages/reactive-dict/reactive-dict.js
zimme: This should be fixed now. I'll add a test for this before we merge.

packages/reactive-dict/reactive-dict.js
zimme: If we don't want to change the behaviour of this
```js
let dictName;
const dict = new ReactiveDict(dictName);
```
to throw an error because you gave `ReactiveDict` and argument which was undefined.

Then I believe we should support
```js
let dictName;
const dict = new ReactiveDict(dictName, { property: 'some value' });
```
without throwing an error and just create an unnamed dict instead.

zimme:
I've updated the implementation a bit based on the discussions.
The following is now possible.
```js
const dict = new ReactiveDict('foo', { bar: 'baz' });
dict = new ReactiveDict(undefined, { foo: 'bar' });
```
I've added a check to the migration logic to only run on client-side as migrations isn't available on server-side and this logic could actually result in duplicate name errors being thrown on code reloads.

I've extended the tests for initial data to test the new format as well.

As long as we're not willing to change the way `new ReactiveDict(emptyVar);` is handled I don't think we should handle `new ReactiveDict(emptyVar, { foo: 'bar' });` any differently, like throwing an error. We might consider opening a new issue to handle the case when `arguments.length > 0` and `dictName === null/undefined` differently.

hwillson:
This looks great @zimme - thanks for all of your work on this. LGTM!

Issue #8658, opened 2017-05-02
Allow options on UglifyJS


PEM--:
* Meteor 1.4.4.1
* Package: minify-js
* Former behavior: Package author were able to setup options on Uglify.
* New behavior: Default options are now forced and there's no capability to setup new ones.
* Expected behavior: Allow package author to setup options on Uglify.

Code would change from:
```js
meteorJsMinify = function (source) {
  var result = {};
  uglify = uglify || Npm.require("uglify-js");

  try {
    result.code = uglify.minify(source, {
      fromString: true,
      compress: {
        drop_debugger: false,
        unused: false,
        dead_code: false
      }
    }).code;
```

into:
```js
meteorJsMinify = function (source, options) {
  var result = {};
  uglify = uglify || Npm.require("uglify-js");

  try {
    result.code = uglify.minify(source, options || {
      fromString: true,
      compress: {
        drop_debugger: false,
        unused: false,
        dead_code: false
      }
    }).code;
```

hwillson:
Thanks for reporting this @PEM--; I see what you mean. We used to [allow uglify options to be passed in](https://github.com/meteor/meteor/blob/6c4507f004c53c44df668e20c8319510797566d8/packages/standard-minifier-js/plugin/minify-js.js#L123). I've marked this issue a `pull-requests-encouraged` as this is a regression, and I think the solution you've proposed makes sense. 

One thing to mention - the [`meteorJsMinify`](https://github.com/meteor/meteor/blob/0d10b33be632afbaa72bfe3a780854f850462b95/packages/minifier-js/minifier.js#L3) function can potentially call 2 different minifiers. It will first try `uglify` since it's faster, but if minification fails, it will then try [`babili`](https://github.com/babel/babili) (since it has better ES2015+ support). So passing a single `options` param into the `meteorJsMinify` function could be bit confusing, since those options could be intended for either one of the minifiers. Since we can't re-use the same options for both minifiers (since they have different options), in the future we'll probably want to allow options for either minifier to be passed in. Again though, since this issue is a regression, I think it's safe to say we'll review a PR that just allows `options` for uglify (to get the same previously allowed functionality back in place). Thanks!

PEM--:
The PR #8665 should do the trick.

hwillson:
To help provide a more clear separation between feature requests and bugs, and to help clean up the feature request backlog, Meteor feature requests are now being managed under the https://github.com/meteor/meteor-feature-requests repository. 

[...]

Issue #4384, opened 2015-05-14
Source maps for Production


Firfi:
Hey,

It seems `meteor build` (without --debug) (and `bundle` neither) builds source maps. Is it planned to work like this? I have error reporting system (rollbar) on frontend that uses stack traces and can work with .map 's, so it would be very cool to know in details about errors that customers have. Minimized stack traces doesn't make much sense.

Firfi:
To be clear, I'm not about to serve .map s to users, but it'll rather be server to error reporting system after error occures and it seeks for correct line of code in original 'minified' stack.

Slava:
Hey @Firfi, I think it is a valid feature to have in the build system. As you might have noticed already, minifiers are hardcoded in the build tool, thus changing its behavior (such as exporting sourcemaps) is not easy w/o changing the core.

We have an ongoing project on the `batch-plugins` branch that aims to address a lot of issues, including this one. Hopefully, after the work is done, it will be easy to configure any minifier to output the source maps (as well as set other options).

DavidReinberger:
Bump, any news/solution on this? I'm using the same app (rollbar) as @Firfi, and it would be awesome to have sourcemaps.

abernix:
Please +1 the original issue as a vote of interest!  Features are managed according to [guidelines for feature requests](https://github.com/meteor/meteor/blob/devel/Contributing.md#feature-requests).

wallslide:
I was going to make a new issue for this, when I found this one instead... so I'll paste what I wrote below:

**Outline of problem**
We want to be able to pinpoint in-production errors occurring on our users' browsers. Currently, the in-browser error reports returned by services such as Kadira or Raygun are very difficult to utilize.

The reason for this is because the source code is minified, and thus we cannot tell what file/line-number exceptions are occurring at based on the captured stacktraces. 

**Solution**
If we could get access to sourcemaps for the minified and deployed production code, we can upload those sourcemaps to these error reporting services, and get meaningful error reports.

**Reference**
Here is an example of the feature in Raygun's Crash Reporting suite that allows us to upload sourcemaps to decipher stacktraces:
![image](https://cloud.githubusercontent.com/assets/917069/18422838/6c6d453c-78da-11e6-8ca4-eaa04b67a3b5.png)

Here are links to the meteor forums where people have requested this feature:

- https://forums.meteor.com/t/sourcemaps-in-production/22982
[...]

PR #8665, opened 2017-05-04
Allow options on Uglyfier

PEM--:
This resolves the loss of options on Uglify discussed here #8658

### DIFF
--- a/packages/minifier-js/minifier.js
+++ b/packages/minifier-js/minifier.js
@@ -1,11 +1,11 @@
 var uglify;
 
-meteorJsMinify = function (source) {
+meteorJsMinify = function (source, options) {
   var result = {};
   uglify = uglify || Npm.require("uglify-js");
 
   try {
-    result.code = uglify.minify(source, {
+    result.code = uglify.minify(source, options || {
       fromString: true,
       compress: {
         drop_debugger: false,


abernix:
While this may work temporarily, I'm not sure this is the correct solution at the moment (and I'm hesitant to merge it) because we'll need to figure out a unified mapping (translating them as necessary) of `options` which work with both minfiers to uniformly resolve things such as production source maps (#4384) and potentially for bundle analysis.

If you need the ability to pass options as you have previously (and you're confident that Uglify works for your code base, despite its ECMAScript deficiencies), another option would be to use an [older version of `standard-minifier-js`](https://atmospherejs.com/meteor/standard-minifier-js).

PEM--:
Indeed, this is not enough for babili. In this case, users must also pass options to babel throughout their `.babelrc`. Using the old release doesn't work as the 2.0.0 is a required dependency.

hwillson:
Good points @abernix - sorry for adding the PR encouraged label and using up some of your time @PEM--. I should have thought more about the impact this would have on some of the other Meteor features cooking behind the scenes, like bundle analysis (since this issue is a regression I was a bit "PR please" trigger happy 😳). Your contribution is really appreciated @PEM--; thanks!

PEM--:
Ha ha, I should also add that kind of label on my open source projects 😄
Always a pleasure to help you when I can, guys.

abernix:
While I can certainly appreciate the need for this, unless you'd be willing to work out the details with the duality of two different minifier APIs and maintaining support for that (as the need was demonstrated for with #8378), I think we'll need to not merge this for now.

It'll probably be worth following https://github.com/meteor/meteor/pull/8698 and seeing where that goes as well.  Obviously a unified minifier which handled all use cases would be ideal, but we're not quite there yet.  When this happens, I think it will make sense to re-visit this.

I realize you've run into problems with using older versions of `standard-minifier-js` due to version constraints, but it's certainly possible to replace `standard-minifier-js` with your own minifier package if necessary (as you have with your own fork).  For now, a fork of, or at least copying the [official](https://github.com/meteor/meteor/tree/master/packages/standard-minifier-js) into an app's `packages` directory and making the appropriate changes is the best suggestion for someone who needs this level of customizability.

Issue #8297, opened 2017-01-28
Update Incorrect Android launch screen image sizes + missing xxhdpi/xxxhdpi sizes


voidale:
There is an updated image sizes for android:
https://github.com/phonegap/phonegap/wiki/App-Splash-Screen-Sizes (yes it's actually updated for cordova and its same sizes that ionic uses)

And the sizes for `xxhdpi` (bug not actually included) and `xxxhdpi` is not existing.

will you accept PR for it?

skirunman:
In my opinion, yes, submit a PR. There was this PR https://github.com/meteor/meteor/pull/8123/commits/355cde306a57ec30a13dd5ede24c9936f93ae76b accepted to update to Cordova 6.4 for Meteor release 1.4.3 so this makes sense to get this fix in now.

abernix:
Would be great to get this into the next 1.4.3 beta too so the sooner the better!  @voidale Would you be willing to make a [contribution](https://github.com/meteor/meteor/blob/devel/Contributing.md#making-changes-to-meteor-core) to Meteor?  It would be much appreciated and is this should be a relatively simple change should be an easy PR win! :Smile:

skirunman:
I think this should be as easy as changing the constants in https://github.com/meteor/meteor/blob/87681c8f166641c6c3e34958032a5a070aa2d11a/tools/cordova/builder.js

paulincai:
@skirunman #8684
As you suggested, I would like to also update the documentation. At the documentation source on git, would you please help a bit with this syntax: {% apibox "App.configurePlugin" %} {% apibox "App.icons" %} {% apibox "App.launchScreens" %} {% apibox "App.appendToConfig" %}

Not sure how to reach to the source files.

skirunman:
@paulincai I did not look at this in detail so I could be wrong, but I think the syntax `{% apibox "App.configurePlugin" %}` in https://github.com/meteor/docs/blob/master/source/api/mobile-config.md is Hexo [ejs](http://www.embeddedjs.com/) markup and is calling https://github.com/meteor/docs/blob/master/assets/api-box.html. See this commit https://github.com/meteor/blaze/pull/76/commits/80ce99577fc5c9b90e4cf525073991ff2af24098

I'm guessing this is using jsdoc to generate this part of documentation straight from the code?

@tmeasday can probably help.

hwillson:
PR #8684 has been merged, so these updates will be coming soon. Thanks @paulincai and @skirunman!

PR #8684, opened 2017-05-10
Update to present media object sizes

paulincai:
As per IOS and Android official documentation. Related to #8297


paulincai:
@hwillson hold on :). I am here. Let me have a look please.

paulincai:
@hwillson I think we are good to go

hwillson:
Awesome, thanks @paulincai! We're all set - LGTM!

derwaldgeist:
Can this be downported to 1.5 please? I cannot use 1.5.1 due to #8950, but I need to submit my iOS app to the app store.

Issue #5818, opened 2015-12-16
.meteor/local/bundler-cache/linker folder is 709MB


timmyg:
**_3 Upvotes**_ ```
webapp|master ⇒ du -sh .meteor/local/bundler-cache/linker
709M    .meteor/local/bundler-cache/linker

linker|master ⇒ du 4e66b88173386163ea6cd7a6ee945a12f3ab5920.cache
208 4e66b88173386163ea6cd7a6ee945a12f3ab5920.cache

bundler-cache|master ⇒ find linker -type f | wc -l
     619

```
Any idea how to go about debugging this? Seems like it is saving an entire cache of my project in a file hundreds and hundreds of times, and I have a feeling this is why my saves are sooo slow (using coffeescript). Any help or a point in the right direction would be greatly appreciated!
```

zol:
You can just delete the ~/.meteor/local directory and see if it makes a difference to your save speed.

avital:
Yeah, looks like we never delete old cache files and deleting the contents of `.meteor/local/bundler-cache` should reduce the size of this directory. (But don't delete `.meteor/local` -- it contains your database!)

dgreensp:
I just mentioned this problem here: https://github.com/meteor/meteor/blob/release-1.3/tools/PERFORMANCE.md#linker-cache

dgreensp:
Which is to say, yeah, that seems not awesome and something we should look at, especially given the time it sometimes takes to write these files.

jeffyamada:
Yikes, this directory has gotten so large on my project that it's preventing Heroku from successfully building my project. Any ideas for how to prevent it, or what it's related to? Happy to help out if anyone can point me in the right direction.

lorensr:
I'm at 14.6 GB 😄 And 27 GB in `.meteor/local/mirrors`, but looks like that's velocity.

dtruel:
wow, I was at 26gb!  I deleted it, but yes, this definitely needs to be looked at.

krishaamer:
Mine has grown to 3.6 gb over a week

danielflippance:
+1 Mine is 12.7 Gb

rjsmith:
Hi,
Oh , I wish I had seen this ticket earlier .. it would have literally saved me hours when running `meteor test` in an upgraded Meteor 1.2 -> 1.3 project because my project's `bundler-cache/linker` folder was 32Gb in size, so copied into the temporary test app on every invocation!

[...]

PR #8687, opened 2017-05-11
Adjusted bundler-cache saved file approach to remove out of date cache files.

hwillson:
Hi all - this PR is intended to help fix longstanding issue #5818. My [comment here](https://github.com/meteor/meteor/issues/5818#issuecomment-226948656) explains the issue in detail, but here's a quick synopsis of the problem:

- In `tools/isobuild/compiler-plugin.js`, the `_linkJS` function is creating and storing linker cache files in an apps `.meteor/local/bundler-cache/linker` directory. The filename used for these cache files comes from hashing the contents of the JS files being compiled (along with several other properties). The hashing algorithm can be seen [here](https://github.com/meteor/meteor/blob/f0dd494c82244df79b544bdbf590ebecc0412189/tools/isobuild/compiler-plugin.js#L1295-L1309).
- Since the hash used for the cache filename is generated using the incoming file's  contents (among other things), this means that the bundler cache is keeping a copy of every single source file change ever made. For example, if I update one of my source .js files to have `console.log('Hello');` and save, a new bundler cache file is generated for this file. If I then update that same file to say `console.log('Hello world!');` and save, a new and separate bundler cache file is generated (since the contents of the source file differ, and the cache filename is built using a hash of the file contents).
- In theory I guess keeping a cached copy of every single file change ever made could be helpful, if people are jumping back and forth between common file changes in specific files. In practice however, this rarely happens - previous changes made are quickly outdated and rendered irrelevant as new changes come in; developers rarely go back in time to exact source file changes (within the context of using the Meteor build tool).
- Given that people are reporting `bundler-cache/linker` directory sizes in the 10's (and sometimes 100's) of GB's for single projects, I think we should consider a new approach that doesn't involve keeping around a cache file for every single JS file change.

This PR includes the fix I referred to in https://github.com/meteor/meteor/issues/5818#issuecomment-226948656. I've adjusted the cache file name hashing algorithm a bit, so filenames are now created with 2 parts - a hash that represents the file itself with all associated internal metadata, and a hash that represents the file contents. With this approach whenever a JS file change is made, a new 2 part hash is generated, and the `bundler-cache/linker` directory is purged of any existing files that match the first part of the hash. A new cache file is then written with all incoming changes, using a filename that is built using the 2 part hash.

Long story short, let me know what you guys think. This change is pretty easy to test - take any Meteor app you have and remove all files in `.meteor/local/bundler-cache/linker`. Start that app using a developer checkout (with this PR), then when running get a count of the number of `bundler-cache/linker/*` files that were created. Update one of the existing source files in your app, and get a new `bundler-cache/linker/*` file count - the counts will be the same (whereas before the file count would have gone up by 1).

2 more quick things to mention:

1. I have no idea how to build `self-test`s for this (since it involves running the build tool). I couldn't find anything related so any pointers here would be greatly appreciated.

2. If this change (or some variation of this change) gets accepted, it's important to note that since the new caching filename approach outlined in this PR differs from the previous approach, old cache files will be left intact. That means people should consider either `meteor reset`ing or removing all `bundler-cache/linker` files, to clean up the old (and no longer used) cache files.

Thanks!

### DIFF
--- a/tools/isobuild/compiler-plugin.js
+++ b/tools/isobuild/compiler-plugin.js
@@ -1294,7 +1294,6 @@ export class PackageSourceBatch {
 
     const fileHashes = [];
     const cacheKeyPrefix = sha1(JSON.stringify({
-      LINKER_CACHE_SALT,
       linkerOptions,
       files: jsResources.map((inputFile) => {
         fileHashes.push(inputFile.hash);
@@ -1308,7 +1307,10 @@ export class PackageSourceBatch {
         };
       })
     }));
-    const cacheKeySuffix = sha1(JSON.stringify(fileHashes));
+    const cacheKeySuffix = sha1(JSON.stringify({
+      LINKER_CACHE_SALT,
+      fileHashes
+    }));
     const cacheKey = `${cacheKeyPrefix}_${cacheKeySuffix}`;
 
     {
--- a/tools/isobuild/compiler-plugin.js
+++ b/tools/isobuild/compiler-plugin.js
@@ -1292,12 +1292,13 @@ export class PackageSourceBatch {
       noLineNumbers: !isWeb
     };
 
-    const cacheKey = sha1(JSON.stringify({
+    const fileHashes = [];
+    const cacheKeyPrefix = sha1(JSON.stringify({
       LINKER_CACHE_SALT,
       linkerOptions,
       files: jsResources.map((inputFile) => {
+        fileHashes.push(inputFile.hash);
         return {
-          hash: inputFile.hash,
           installPath: inputFile.installPath,
           sourceMap: !! inputFile.sourceMap,
           mainModule: inputFile.mainModule,
@@ -1307,6 +1308,8 @@ export class PackageSourceBatch {
         };
       })
     }));
+    const cacheKeySuffix = sha1(JSON.stringify(fileHashes));
+    const cacheKey = `${cacheKeyPrefix}_${cacheKeySuffix}`;
 
     {
       const inMemoryCached = LINKER_CACHE.get(cacheKey);
@@ -1394,8 +1397,19 @@ export class PackageSourceBatch {
     if (canCache) {
       LINKER_CACHE.set(cacheKey, ret);
       if (cacheFilename) {
+        const cacheKeySuffixRegex = new RegExp('_' + cacheKeySuffix, 'g');
+        const oldCacheFilename =
+          cacheFilename.replace(cacheKeySuffixRegex, '_*');
         // Write asynchronously.
-        Fiber(() => files.writeFileAtomically(cacheFilename, retAsJSON)).run();
+        Fiber(() => {
+          try {
+            files.rm_recursive(oldCacheFilename);
+          } catch (error) {
+            // If we can't remove the old cache file, we'll just continue
+            // writing the new file.
+          }
+          files.writeFileAtomically(cacheFilename, retAsJSON);
+        }).run();
       }
     }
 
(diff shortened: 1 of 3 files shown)


tools/isobuild/compiler-plugin.js
```diff
@@ -1394,8 +1397,19 @@ export class PackageSourceBatch {
...
       LINKER_CACHE.set(cacheKey, ret);
       if (cacheFilename) {
+        const cacheKeySuffixRegex = new RegExp('_' + cacheKeySuffix, 'g');
+        const oldCacheFilename =
+          cacheFilename.replace(cacheKeySuffixRegex, '_*');
         // Write asynchronously.
-        Fiber(() => files.writeFileAtomically(cacheFilename, retAsJSON)).run();
+        Fiber(() => {
+          try {
+            files.rm_recursive(oldCacheFilename);
```
benjamn: This might be a good place to use [`files.rm_recursive_async`](https://github.com/meteor/meteor/blob/f0dd494c82244df79b544bdbf590ebecc0412189/tools/fs/files.js#L292), so we don't block writing the new `cacheFilename`.

hwillson: Great idea - changing now. Thanks!

hwillson: Sorry, I spoke too soon @benjamn - as currently implemented, we need the remove to happen before writing the new cache file; otherwise if the remove fires after writing the new file, it  would effectively wipe the new file as well (since both the old and new files are prefixed with the same hash, and we're removing by that hash). That being said, maybe we're okay with the remove being synchronous in this case, since both the remove and write are happening in a fiber anyways? The remove will hold the write up but the write should still be done well before the created cache file is needed by a rebuild.

zimme: Would it make sense to synchronously get the list of filenames in the folder and then start the remove and write process in parallel, where the remove process would use `folder/file.ext` from the list?

hwillson: We could come up with another way to handle this for sure - I'm just not sure if it's worth adding extra complexity though, since the delete + write are running in a fiber anyways (and we’re not worried about having the new cache file available until the next rebuild). Happy to rework any of this though!

benjamn:
@hwillson Right now `LINKER_CACHE_SALT` is [included in the `cacheKeyPrefix`](https://github.com/hwillson/meteor/blob/b9742a7ac65cfc39efe7f2b438f32340141518bd/tools/isobuild/compiler-plugin.js#L1297), so any time we bump that salt, we won't remove any files previously in the cache, because they no longer match the prefix. I have a feeling that's going to be a major source of stranded cache files, from release to release, since we tend to bump the salt with each (major) release. Is there anything we can do about that?

hwillson:
Thanks for the heads up @benjamn; I didn't realize the salt was incremented that frequently. In that case I think it would make sense to have the salt included in the second hash (so in the proposed `cacheKeySuffix`) instead of the first hash (the `cacheKeyPrefix`). That way when the salt is increased `cacheKeySuffix` will be changed, and the old cached files will still be removed properly, since file removal is based on the first part of the filename (the `cacheKeyPrefix` part - which hasn't changed). I'll run some tests on this approach and post back shortly. Thanks!

hwillson:
I made the proposed change and everything appears to work properly; old cache files are properly removed when the `LINKER_CACHE_SALT` changes. Thanks!

benjamn:
Merged to #8327 with https://github.com/meteor/meteor/pull/8327/commits/16a8a3ae3c60e1c3b3692b5b78df2433780e57a5, thanks!

PR #8688, opened 2017-05-12
Ensure visualizer displays on top of other UI

Primigenus:
There was an issue where most of the visualizer was visible but the module size details displayed when hovering over a segment were hidden behind some UI elements. Setting a high z-index fixes it.


stubailo:
Nice! Increasing Z Index is always a good fix ;]

abernix:
Thanks, @Primigenus.  Obviously there's still a bit of a competition here with other `z-index`s, but I don't see harm in reducing some cases.

abernix:
This should be included in `bundle-visualizer@1.0.1`!

Issue #5932, opened 2016-01-04
Feature request: 'hint' option for find() if using a Mongo cursor


bradvogel:
I'd like the ability to pass a 'hint' option to find when using a Mongo collection. See here: https://docs.mongodb.org/v3.0/reference/operator/meta/hint/

It's very common to use the `hint` optimization in large-scale Mongo apps.

avital:
That'd be useful. We'd take a look at a pull request implementing this, but in the meanwhile you should be able to use [rawCollection](http://docs.meteor.com/#/full/Mongo-Collection-rawCollection) directly.

Unless this is something that's only available with newer version of the Mongo driver, in which you'd probably have to wait until we finish https://github.com/meteor/meteor/issues/5763

wearhere:
Turns out that you can actually pass `$hint` to `Collection#find`, which is fine by me except for that you apparently then can't use `sort` in the `options`: http://stackoverflow.com/a/30916309/495611

bradvogel:
Thanks! Would be great to have the above documented!

arthurlataks:
Yet I still can't count results that match my query when using $hint. Maybe I'm doing something wrong...

`var totalCount = Meteor.users.find({ $query: searchQuery, $hint: indexType}).count();`

I end up receiving this exception:

`Exception while invoking method 'getSearchUserData' MongoError: exception: unknown top level operator: $query
I20160404-22:07:29.629(3)?     at Object.Future.wait (/Users/xxx/.meteor/packages/meteor-tool/.1.1.10.u899n++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:398:15)
I20160404-22:07:29.629(3)?     at [object Object]._.extend.count (packages/mongo/mongo_driver.js:1059:1)
I20160404-22:07:29.629(3)?     at [object Object].Cursor.(anonymous function) [as count] (packages/mongo/mongo_driver.js:869:1)`

faridblaster:
up

nlhuykhang:
I don't think this is a Meteor issue. I tried below command in Mongo shell and I got error too:

```
db.users.find({
  $query: {
    _id: '...'
  },
  $hint: {
    _id: 1
  },
}).count()
```

bpartridge:
[...]

PR #8692, opened 2017-05-12
Add Mongo hint and maxTimeMS options (#5932)

bpartridge:
Addresses https://github.com/meteor/meteor/issues/5932. Tests that exceptions are thrown correctly, and that functionality works. As a side note, unlike its behavior many versions ago, Mongo now ignores hints silently when a malformed or unindexed hint is used, so caveat emptor. We should add documentation as well, possibly including that tidbit.


packages/mongo/mongo_driver.js
```diff
@@ -992,6 +992,13 @@ MongoConnection.prototype._createSynchronousCursor = function(
     replaceTypes(cursorDescription.selector, replaceMeteorAtomWithMongo),
     cursorOptions.fields, mongoOptions);
 
+  if (typeof cursorOptions.maxTimeMS !== 'undefined') {
```
hwillson: This is going to be documented alongside other collection `find` options, like `pollingIntervalMs` and `pollingThrottleMs`. To keep naming consistent, please consider changing `cursorOptions.maxTimeMS` to `cursorOptions.maxTimeMs`.

bpartridge: Was just following Mongo's API, but they're not consistent anyways haha: https://docs.mongodb.com/manual/reference/method/cursor.maxTimeMS/#cursor.maxTimeMS https://docs.mongodb.com/manual/reference/configuration-options/#operationProfiling.slowOpThresholdMs https://docs.mongodb.com/manual/reference/method/db.setProfilingLevel/ which doesn't even capitalize at all ... so that makes sense. Will do.

hwillson: Hi @bpartridge - any progress on the requested changes? Let me know if you're busy and I'll finish things up. Thanks!

bpartridge: Will get to this tonight!

bpartridge:
Added requested changes and documentation (including usage tips). Also rebased against current devel branch. Should be good to go @hwillson!

Issue #7392, opened 2016-07-12
Support for UNIX sockets


vlasky:
Meteor should have the ability to listen on a UNIX socket, like most network servers.

This is especially useful to people who are running Meteor apps behind Nginx or another web server which is used for SSL termination and efficiently serving static files.

At the moment, the Meteor tool seems to only support TCP, so when running on localhost, Nginx is forced to pointlessly create TCP connections between itself and the meteor tool (which acts as an HTTP proxy) and then creates even more TCP connections between itself and the meteor app (which by default runs on a randomly assigned port). As a result, this incurs additional overhead and leaves lots of TCP connections in the TIME_WAIT state.

I propose that the meteor tool be extended to support UNIX sockets with the following syntax:

`meteor run --port unix:/var/run/meteor.sock`

And also with the optional --app-port option which uses a fixed port for the app instead of a randomly-assigned one:

`meteor run --port unix:/var/run/meteor.sock --app-port unix:/var/run/meteor-app.sock`

mjmasn:
Can't speak for Meteor regarding the feature request but you absolutely shouldn't be running meteor behind nginx with the meteor command. **The `meteor` command is for development only.** You need to use `meteor build` to create a app bundle that you can deploy as you would any nodejs app. It's in the guide at https://guide.meteor.com/deployment.html#custom-deployment

vlasky:
Yes, am well aware - our development server also runs behind Nginx. This ensures we can test and measure the impact of Nginx configuration changes too.

My request for UNIX socket support also applies to the app bundle created with the meteor build command.

There's no reason why all components shouldn't have UNIX socket support.

abernix:
Not sure what the popularity of this feature will, but those interested, please +1 the original issue!

Difficulty-wise, for the production side, this might be as easy as a change [in this function](https://github.com/meteor/meteor/blob/31d6bbbda756103271910591066acb39014955c4/packages/webapp/webapp_server.js#L440-L447).  But for `meteor-tool` development, it will be more involved.

abernix:
Can't believe I'm suggesting this, but as perhaps the hackiest workaround ever, if you want to take advantage of the existing exception in there for Windows, you could use the following for your `PORT` when you start your app (this will only work in production when starting with `node`, not with the `meteor` tool):

    PORT=/tmp/\\a\\pipe\\a-meteor.sock node main.js

A quick test with `netcat` shows that it is indeed working...

```sh
[...]

PR #8702, opened 2017-05-17
Support for UNIX sockets (#7392)

vlasky:
@abernix et al, here's my enhancement to implement [Support for UNIX sockets (#7392)](https://github.com/meteor/meteor/issues/7392) in meteor.

If the PORT environment variable contains a UNIX file/path, this will be treated as the UNIX domain socket file to listen for incoming connections.

To summarise what I've done:

1. I added an additional regex check to the existing code in WebAppInternals.parsePort() to test if the PORT variable contains a UNIX file/path and if so, return it as a string.

2. In function main, we now pass an object called **listenOptions** to httpServer.listen() and this object contains the required options to configure httpServer to either listen on a TCP port or UNIX socket.

3. There is a sanity check that handles the situation where there is an existing an existing socket file. If there is, it will try to connect to it to confirm that it does not belong to a running server instance. If it does not, it presumes that the socket file is stale and deletes it.

I note that this support only works in production mode and not when using the meteor tool. I would have liked to add support to the meteor tool, however I see that the package **node-http-proxy** that it relies on does not appear to support forwarding connections to UNIX sockets, so that would have required a lot more work.


packages/webapp/webapp_server.js
```diff
@@ -439,9 +440,9 @@ var getUrlPrefixForArch = function (arch) {
     '' : '/' + '__' + arch.replace(/^web\./, '');
 };
 
-// parse port to see if its a Windows Server style named pipe. If so, return as-is (String), otherwise return as Int
+// parse port to see if it's a Windows Server-style named pipe or a UNIX path/filename for a UNIX domain socket file. If so, return as-is (String), otherwise return as Int
 WebAppInternals.parsePort = function (port) {
-  if( /\\\\?.+\\pipe\\?.+/.test(port) ) {
+  if( /\\\\?.+\\pipe\\?.+/.test(port) || /^(.+)\/([^/]+)$/.test(port) ) {
```
hwillson: Could you update the [existing `parsePort` tests](https://github.com/meteor/meteor/blob/f0dd494c82244df79b544bdbf590ebecc0412189/packages/webapp/webapp_tests.js#L160) to verify this additional parsing works properly?

packages/webapp/webapp_server.js
```diff
@@ -815,7 +816,55 @@ var runWebAppServer = function () {
     var localPort = WebAppInternals.parsePort(process.env.PORT) || 0;
     var host = process.env.BIND_IP;
     var localIp = host || '0.0.0.0';
-    httpServer.listen(localPort, localIp, Meteor.bindEnvironment(function() {
+
+    if (typeof localPort == "number")
+    {
```
hwillson: Bit of a nit pick, but can you follow the existing opening/closing brace style that's used in the rest of the file (open brace on same line, etc.)?

packages/webapp/webapp_server.js
```diff
@@ -815,7 +816,55 @@ var runWebAppServer = function () {
...
+	  var clientSocket = new net.Socket();
+
+          clientSocket.on('error', Meteor.bindEnvironment(function(e) {
+
+            if (e.code == 'ECONNREFUSED') {
+
+              console.log("Deleting stale socket file");
+              fs.unlinkSync(socketPath);
+
+              httpServer.listen(listenOptions, Meteor.bindEnvironment(function() {
```
hwillson: It looks like this `httpServer.listen` block of code is repeated a few lines down, and is the exact same. Could this be extracted into a common function to clean the code up a bit?

packages/webapp/webapp_server.js
vlasky: Done :-)

abernix: I'm not convinced this new RegExp is much better than just checking to see if the path has a `/` in it at all (i.e. `/\//`, which is still not ideal.  As it stands, this wouldn't allow the socket file to be in the current directory, something we should probably try to support via [`path.resolve`](https://nodejs.org/dist/latest-v4.x/docs/api/path.html#path_path_resolve_from_to), for example.

In addition to that, can we also could we consider another option for detecting if it's a TCP socket file?  This should also probably check `process.platform` (to ensure it's `win32`) for the first part.

packages/webapp/webapp_server.js
```diff
@@ -815,18 +816,47 @@ var runWebAppServer = function () {
...
+      }));
+    };
+
+    if (typeof localPort == "number") {
+      var listenOptions = { port: localPort, host: host };
+    } else {
+      var socketPath = localPort;
+      var listenOptions = { path: socketPath };
+
+      httpServer.on('error', Meteor.bindEnvironment(function(e) {
```
abernix: The addition of all this additional logic suggests that these should probably be come more modular functions instead of this rather-large `else` block.

packages/webapp/webapp_server.js
```diff
@@ -815,18 +816,47 @@ var runWebAppServer = function () {
...
+        if (e.code == 'EADDRINUSE') {
```
abernix: What is this `EADDRINUSE` ➡️ `ECONNREFUSED` logic based on?  I'm hesitant to have all this extra logic in here which would normally produce valid error messages anyhow.  Can we really safely assume this is stale?

packages/webapp/webapp_server.js
```diff
@@ -815,18 +816,47 @@ var runWebAppServer = function () {
...
+	  var clientSocket = new net.Socket();
+          clientSocket.on('error', Meteor.bindEnvironment(function(e) {
+            if (e.code == 'ECONNREFUSED') {
```
abernix: Why does a new socket generation of an `ECONNREFUSED` error do we assume it's safe to cleanup the previous socket file?

packages/webapp/webapp_server.js
```diff
@@ -815,18 +816,47 @@ var runWebAppServer = function () {
...
+              console.log("Deleting stale socket file");
+              fs.unlinkSync(socketPath);
+              startHttpServer(listenOptions);
```
abernix: Are we completely okay with the `startHttpServer` getting called again here as it will have already been called on line 859 below?

vlasky:
@abernix In commit https://github.com/meteor/meteor/pull/8702/commits/d9e7f8672b62217d69056c37ae6e00fe412cffb3 I have rewritten the code to address the majority of your review comments.

1. The second regexp has been removed. The code now checks to see if the PORT value is a number or a string containing a windows named pipe (via the original regex), otherwise the value is presumed to be of the desired UNIX domain socketPath (and socket file).

2. The code no longer calls startHttpServer() twice. It now performs fs.existsSync(socketPath) to determine if there is an existing file with the same name as the desired socket file and then fs.statSync(socketPath).isSocket() to determine if the existing file is in fact a socket file (which is a special file) as opposed to an ordinary file (or any other type).

If that file is a socket file, it will then try to perform a dummy connection to it to determine if it's live - e.g. belonging to another running process which is listening for connections on it.

If the connection attempt is successful, the socket file is live and being used by a running process, to we will not delete the file and instead exit.

If the connection attempt results in the error ECONNREFUSED, we know that the socket file is stale and we can safely delete it using fs.unlinkSync() and call startHttpServer() which will create a fresh socket file.

3. There is now cleanup code that deletes the socket file created by this process when the node process exits or is ended using SIGINT (Control-C). For those unfamiliar with UNIX socket files, it is the duty of the programmer to delete them up once the program no longer needs them - the OS does not remove them itself.

4. I removed parsePort() as it is no longer needed by the new logic. Also, it no will no longer convert invalid port values like "8080abc" to the numeric port value 8080 - they will instead be treated as a string.

I really think that old behaviour was wrong and served only to conceal mistakes made by those writing launcher scripts.

abernix:
@vlasky Sorry I didn't get to review this today but it's first on my plate tomorrow! (And rest assured, Meteor 1.5.1 is not coming out tomorrow. :wink:)

packages/webapp/webapp_server.js
```diff
@@ -812,22 +804,82 @@ var runWebAppServer = function () {
...
+            clientSocket.on('error', Meteor.bindEnvironment(function(e) {
+              //If there is an existing socket file that we cannot connect to,
+              //it must be stale as it is not associated with a running server
+              if (e.code == 'ECONNREFUSED') {
+                console.log("Deleting stale socket file");
+                fs.unlinkSync(socketPath);
+                startHttpServer(listenOptions);
+              }
+            }));
+            clientSocket.connect({ path: socketPath }, function() {
```
abernix: It's typically [discouraged](https://nodejs.org/api/net.html#net_socket_connect_options_connectlistener) to use the `net.Socket.prototype.connect` method directly, but rather to rely on [`net.createConnection`](https://nodejs.org/api/net.html#net_net_createconnection_path_connectlistener).  Something to consider?

packages/webapp/webapp_server.js
```diff
@@ -812,22 +804,82 @@ var runWebAppServer = function () {
...
+        process.on('SIGINT', Meteor.bindEnvironment(function(e) {
+          process.exit();
+        }));
+      }
+    } else if (typeof localPort == "number") {
+      var listenOptions = { port: localPort, host: host };
+      startHttpServer(localPort);
+    } else {
+      console.error("Invalid PORT specified");
+      process.exit();
```
abernix: I don't think we should be calling `process.exit` anywhere in this code.  That will be handled by the caller to `main()` [here in `boot.js`](https://github.com/meteor/meteor/blob/devel/tools/static-assets/server/boot.js#L353).  I think you can just return `ERROR` or something (though you should leave the `console.error`.

packages/webapp/webapp_server.js
```diff
@@ -812,22 +804,82 @@ var runWebAppServer = function () {
...
+        } else {
+          startHttpServer(listenOptions);
+        }
+        //Clean up the socket file when we've finished to avoid leaving a stale one
+        //That situation should only be able to happen if the running node process is
+        //killed via signal 9 (SIGKILL).
+        process.on('exit', Meteor.bindEnvironment(function(e) {
+          fs.unlinkSync(socketPath);
+        }));
+        process.on('SIGINT', Meteor.bindEnvironment(function(e) {
```
abernix: This doesn't seem relevant to this change at all, what's the reason for this?

hwillson:
Hi @vlasky - have you had a chance to look over the recent review comments?

packages/webapp/webapp_server.js
```diff
@@ -812,22 +804,82 @@ var runWebAppServer = function () {
...
+      var socketPath = localPort;
+      var listenOptions = { path: socketPath };
 
+      // Test to see if the socket path is Windows Server-style named pipe with the
+      // format: \\.\pipe\{randomstring} or \\{servername}\pipe\{randomstring}
+      // Otherwise, treat it as a UNIX domain socket path pointing to a socket file
+      if ( /\\\\?.+\\pipe\\?.+/.test(socketPath)) {
+        startHttpServer(listenOptions);
+      } else {
+        if (fs.existsSync(socketPath)) {
```
sebakerckhof: @abernix `existsSync` is not deprecated, `exists` is however:
https://nodejs.org/api/fs.html#fs_fs_existssync_path

Afaik, `exists` was deprecated because it didn't follow the `(err, result)` callback signature.

abernix: Ah, bizarre, but true!  But still not necessary to use it here. :)

hwillson:
Hi @vlasky - are you still interested in working on this PR? We'd love to get these changes in place, after the recent review questions have been addressed. If you're not available to work on this let me know, and I'll jump in to help out. Thanks!

packages/webapp/webapp_server.js
vlasky: That's right, existsSync() is not deprecated, however I take @abernix's point about replacing the separate calls to existsSync() and statSync() with a single call to statSync() to avoid a race condition.

packages/webapp/webapp_server.js
vlasky: Without this cleanup code, a stale UNIX socket file (.sock) file is left on disk if you terminate the node process with Control-C (SIGINT).

packages/webapp/webapp_server.js
vlasky: net.createConnection() is not suitable here because it is necessary to set the error handler in advance for this very short-lived client socket object whose sole purpose is to distinguish between a live UNIX socket and a stale UNIX socket file.

The documentation does not discourage it use, but says "ise this only when implementing a custom Socket." or in my view "use it when you have to".

packages/webapp/webapp_tests.js
```diff
@@ -156,23 +156,6 @@ Tinytest.add("webapp - generating boilerplate should not change runtime config",
   test.isFalse(__meteor_runtime_config__.WEBAPP_TEST_KEY);
 });
 
-// Support 'named pipes' (strings) as ports for support of Windows Server / Azure deployments
-Tinytest.add("webapp - port should be parsed as int unless it is a named pipe", function(test){
```
vlasky: @abernix, in my previous commit https://github.com/meteor/meteor/commit/693ac8ae7293fdbd0dbfaca47b67fe39a950994d, I preserved the legacy parsePort() behaviour and tests, but you weren't happy because socket path would require an absolute path starting with a forward slash, i.e. you could not create a socket file in the current directory.

I note that I have seen other frameworks that use the presence of a forward slash to distinguish between listening on a UNIX socket and listening on a TCP port.

In order to satisfy your concern, I changed this behaviour in commit https://github.com/meteor/meteor/commit/d9e7f8672b62217d69056c37ae6e00fe412cffb3 so that a number would be treated as a TCP port and anything else would be treated as a UNIX socket, and this required removing the existing parsePort() functionality.

vlasky:
@hwillson I have been away dealing with other business commitments recently.

My company has been using commit https://github.com/meteor/meteor/commit/d9e7f8672b62217d69056c37ae6e00fe412cffb3 in production without any issues.

The only change I plan to make to this PR for time being is the one mentioned in https://github.com/meteor/meteor/pull/8702#discussion_r126886883 to resolve a race condition.

After that you (or anyone else) can jump in if you see a need.

hwillson:
No problem at all @vlasky - thanks again for putting this together. After you add in changes to help with the race condition, I'll jump in to help to get this wrapped up / merged.

vlasky:
@hwillson - the race condition fix has been implemented in https://github.com/meteor/meteor/pull/8702/commits/76808e882f15c1edf7d1bc4deff5593f83279f1f. Feel free to jump in.

skirunman:
Just wondering if this is moving forward in 1.5.2 or 1.6 releases as currently tagged as 1.5.1.

hwillson:
Good question @skirunman - I'm holding this one up unfortunately; hopefully I'll have the discussed changes wrapped up this week. We'll then be able to see which release it makes sense to add this to. Thanks!

hwillson:
Hi all - I have a new version of this just about ready to go, but I noticed a rather large issue while testing.

First off, regarding `net.Socket.prototype.connect` versus `net.createConnection`, as long as the error handler callback is set immediately after the `createConnection` call, we should be okay. `createConnection` handles this by [deferring the error event to the `nextTick`](https://github.com/nodejs/node/blob/v4.8.4/lib/net.js#L1001). That being said, unfortunately I don't think either approach will work as attempted.

Let's assume we're using `createConnection` (but this issue stands with `net.Socket.prototype.connect`). Using `createConnection` to check to see if we can connect to an existing socket file, then if we connect flag that socket file as already being used by another server, isn't a perfect check. If you SIGKILL the running node process, the socket file is left in place (not much we can do about this). Then if you use `createConnection` to connect to that file again, it will work. When that happens, we're flagging the socket file as being used by another server, when in reality it isn't. 

Unfortunately, there isn't a perfectly reliable way to tell if a socket file is already being used by another process. That means we need to handle this check in another way. A few options:

1. (Least safe but automatic) Use `net.createConnection` as outlined, but if we can connect to the socket file, throw an exception, remove the existing socket file, and continue on. With this approach we don't care if the existing socket file is being used by another process or if it was left over from a SIGKILL. In either case we remove it, and continue to start the http server. This approach will work, but if another process really is using the socket file, and that socket file is removed while it's running, well, needless to say that wouldn't be good ...

2. (Most safe but manual) We don't try to connect to an existing socket file, ever. Instead, when firing up the webapp we first check to see if an existing socket file exists (like we're doing now), but if found we throw an exception describing the problem and the location of the existing socket file, and exit. This approach is nice and safe (as we're not killing a socket file that could be used by another process), but we're leaving the responsibility of cleaning up a potentially stale socket file to the developer deploying/starting the app. Requiring manual intervention like this could be risky, especially when the application is deployed to an environment that tries to automatically re-start crashed apps. It could be waiting for the stale socket file to be removed manually, and it could take some time before people notice.

As you can see, neither approach is perfect. I'm leaning towards the first approach and making sure people know that they need to define their socket paths very carefully, to make sure they aren't clobbering another process. This approach falls more in-line with other areas of Meteor, that make certain assumptions to keep the development/deployment process more automatic. I'd love to hear what others think, and if anyone has other suggestions? Thanks!

hwillson:
Just to add - the existing code in this PR is already achieving #1 above, since the `socketPath` file is being removed during the specified `error` callback. Firing up a second server, getting the socket file already being used message, then watching my first server crash really threw me off during testing however. If we agree this is the way to go though, I'll adjust the error handling / messaging a bit to be more explicit about what's happening.

packages/webapp/webapp_server.js
```diff
@@ -860,23 +852,86 @@ function runWebAppServer() {
...
+      }));
+    };
+
+    localPort = isNaN(Number(localPort)) ? localPort : Number(localPort);
 
-    }, function (e) {
-      console.error("Error listening:", e);
-      console.error(e && e.stack);
-    }));
+    if (typeof localPort == "string") {
```
zimme: `===` ?

zimme:
@hwillson, dod you read this? https://stackoverflow.com/questions/16178239/gracefully-shutdown-unix-socket-server-on-nodejs-running-under-forever?answertab=active#tab-top

hwillson:
Thanks @zimme - yes, I have seen that approach and have tried several variations. None of them are perfect unfortunately. If I SIGKILL the running node instance, the sock file is left in place, and I can still connect to it on retry. That means we never actually get a `ECONNREFUSED` when retrying a connection with a stale socket file, or when connecting using a socket file that's already in use - in both cases I can connect using that socket file, so there isn't an easy way to differentiate between the two scenarios.

zimme:
I guess using ref/unref won't help either when SIGKILL is used?

hwillson:
Hi all - based on my comments in https://github.com/meteor/meteor/pull/8702#issuecomment-321894416, I've adjusted this PR a bit. If an existing socket file is found when the application starts up, it is first removed before a new socket file is created when the HTTP server starts. As mentioned previously, this could be a bit dangerous if another application is configured to use the same socket file, on the same server. As long as people are configuring their applications properly, and making sure to use unique socket paths for different applications, this really shouldn't be an issue. The good news is by making this assumption up-front, we're able to simplify the codebase (and possible error paths).

With this new approach, some of the previous (outstanding) review comments are no longer valid. A quick synopsis:

- I've re-arranged things to help reduce the nested if levels.

- Regarding `WebAppInternals.parsePort`, since it was available as part of the previous public API, I've restored that functionality. It will now return a parsed numeric port if it can, otherwise it will return the full port string. This should provide the same functionality that  was in place (I've restored the removed tests to verify this), and allow unix socket path's to be returned as well. That being said, since this function is no longer used internally, I've added a DEPRECATED label, and mentioned that using this function directly is not recommended. 

- `net.Socket.prototype.connect` vs. `net.createConnection` no longer matters (we're not attempting to connect to an existing socket file - we're just removing it if it exists).

- `process.on('exit', ...` and `process.on('SIGINT', ...` are no longer being used. Since we're removing any previous socket file on startup no matter what, we don't need to worry about cleaning the socket file at the various possible program exit states.

- I've removed the use of `process.exit` and am instead throwing exceptions where necessary (since as mentioned `process.exit` is already being handled upstream in https://github.com/meteor/meteor/blob/devel/tools/static-assets/server/boot.js#L353).

I believe this addresses all outstanding comments. After these new changes have been reviewed, I'll put a `History.md` entry together. I'll also (shortly) post a link to the test application I've been using to verify these changes (Nginx + Meteor), along with testing steps.

Thanks!

hwillson:
Adding a test app: https://github.com/hwillson/meteor-pr-8702

zimme:
> process.on('exit', ... and process.on('SIGINT', ... are no longer being used. Since we're removing any previous socket file on startup no matter what, we don't need to worry about cleaning the socket file at the various possible program exit states.

Shouldn't we try and do cleanup anyways? As we're creating the socket file we should be removing it when we do a graceful shutdown, even though we always remove it before starting again. But what if you don't start it again and then change the name of the socket, and all of a sudden you have an unnecessary socket file.

hwillson:
Good question @zimme - I had exit cleanup code in place just before I committed, but decided to take it out in the end to simplify things. My logic was that we're removing it again at startup, and if a socket file using version of the app is never started again, we've ended up leaving one 0 byte file on the filesystem. I also wasn't crazy about the idea of having additional exit handlers defined in the `webapp` package (when the process is managed via [tools/static-assets/server/boot.js](https://github.com/meteor/meteor/blob/devel/tools/static-assets/server/boot.js#L353)). Again though, I was on the fence on this one as well, so maybe I'll add it back in. Thanks for the feedback!

zimme:
> we've ended up leaving one 0 byte file on the filesystem

Still littering the filesystem, what if someone setup meteor to always generate a new socket file on boot with a script? Then you might end up with a lot of unused socket files.

hwillson:
Haha - that would be horrible, but valid point. I'm re-visitng the cleanup code now. Thanks!

hwillson:
Oh Circle - 4 failures when everything passes locally 🙁. Restarting, but the reported failures shouldn't be related to these changes. Also re-moving the WIP label from the title; this should be ready for a final review. I still have to add a `History.md` entry, but I'd like to make sure we're all on board with this new direction before finalizing. Thanks!

packages/webapp/webapp_server.js
```diff
@@ -486,14 +490,19 @@ var getUrlPrefixForArch = function (arch) {
...
-  if( /\\\\?.+\\pipe\\?.+/.test(port) ) {
-    return port;
+// Parse the passed in port value. Return the port as-is if it's a String
+// (e.g. a Windows Server style named pipe), otherwise return the port as an
+// integer.
+//
+// DEPRECATED: Direct use of this function is not recommended; it is no
+// longer used internally, and will be removed in a future release.
+WebAppInternals.parsePort = port => {
+  let parsedPort = parseInt(port);
```
abernix: This does appear to still be used from `webapp_tests.js`.  Should we export this function with another internal name for that purpose?

hwillson: It's only used in `webapp_tests.js` to test itself 🙂. Regarding how to deprecate it, I think just mentioning it in the `History.md` should be enough (because as you said, `Internals`).

packages/webapp/webapp_server.js
```diff
@@ -860,24 +869,46 @@ function runWebAppServer() {
...
-      if (process.env.METEOR_PRINT_ON_LISTEN) {
-        console.log("LISTENING"); // must match run-app.js
-      }
 
-      var callbacks = onListeningCallbacks;
-      onListeningCallbacks = null;
-      _.each(callbacks, function (x) { x(); });
+    var startHttpServer = function(listenOptions) {
+      httpServer.listen(listenOptions, Meteor.bindEnvironment(function() {
+        if (process.env.METEOR_PRINT_ON_LISTEN)
```
hwillson: Sure thing - I didn't really review that part of the code; I'll update that entire function a bit.

packages/webapp/socket_file_tests.js
```diff
@@ -0,0 +1,68 @@
+import { writeFileSync, unlinkSync, statSync } from 'fs';
+import { createServer } from 'net';
+import {
+  removeExistingSocketFile,
+  registerSocketFileCleanup,
+} from './socket_file';
+import { EventEmitter } from 'events';
+
+const testSocketFile = '/tmp/socket_file_tests';
```
hwillson: Ha - oops. Good point - changing, thanks!

hwillson: Right - I can't reference `tools/fs/files.js` directly from `webapp`; I'll use Node's [`os.tmpdir`](https://nodejs.org/docs/latest-v4.x/api/os.html#os_os_tmpdir) instead.

hwillson:
Well, I was just prepping a `History.md` entry for this and was about to say we're good to go here, but then remembered we needed to add a mention that this was production only. The thing is, this doesn't need to be production only if we don't use the `PORT` environment variable to set the UNIX socket file path. If we use another unique environment variable like `UNIX_SOCKET_PATH`, then socket files can be used with the Meteor Tool and in production Node apps. Since hijacking `PORT` to store a pipe or socket path was a hack anyways, I've gone ahead and changed this around. UNIX domain sockets can now be used by setting something like `UNIX_SOCKET_PATH=/tmp/somesock.sock`. Windows named pipes are still set using `PORT`, but that's a change for another day/PR. When starting the Meteor Tool with the `UNIX_SOCKET_PATH` set, it will now show that a socket file is being used (instead of display the host:port combo). I've also finalized the `History.md` entry. Standing by for feedback - thanks!

zimme:
> Windows named pipes are still set using PORT, but that's a change for another day/PR.

Are we sure we don't want to add `WINDOWS_NAMED_PIPE` which would work the exact same way as `UNIX_SOCKET_FILE`?

hwillson:
We could - I left those changes out since this PR should be good to go as is and meets the requirements of the original issue (https://github.com/meteor/meteor/issues/7392). Adding adjustments for Windows seemed to be a bit out of scope, and the current Windows named pipe hack has been in place for a while, so it's being used in production. Wiring up something like `WINDOWS_NAMED_PIPE` didn't seem to be as pressing since we'll still have to support the `PORT` hack for Windows anyways. Unless of course we deprecate the Windows `PORT` approach, but we'll definitely have to handle that situation carefully.

kentonv:
Hi all. This is largely our fault, not yours, but I wanted to note for informational purposes that this change played a part in a global outage of self-hosted Sandstorm.io servers. It turns out our startup code would often set a PORT variable with multiple ports, comma-separated, like "80,443". In the past, Meteor would use the first port, and other code in our server would parse out the remaining ports and do other things with them. This change implemented more strict parsing of ports, so broke that configuration. Unfortunately, our usual testing configurations did not happen to have multiple ports set, so the problem was not caught until real servers started dying after auto-updating.

Again, this is our fault for questionable code (we should never have relied on lenient PORT parsing) and insufficient testing, but I thought having visibility into this kind of issue would be informative.

Sandstorm issue and postmortem here: https://github.com/sandstorm-io/sandstorm/issues/2993

PR #8703, opened 2017-05-17
Move localStorage access back into try statement

jamiter:
@benjamn, here you go. Maybe move the comment also inside the try/catch block?

This fixes #1291 again after it got reverted in the 1.5 branch by accident.

Accessing window.localStorage can immediately throw an error in IE and other/older webkit versions.

See https://github.com/meteor/meteor/pull/8327#issuecomment-302176415


benjamn:
Thanks @jamiter!

1u:
@lu40: did this solve the webKit issue for you?

Issue #8696, opened 2017-05-15
Can’t send emails from my app with email@1.2.1


mgscreativa:
Hi! I have Meteor 1.4.4.1 running in my dev machine Ubuntu 16.04 (but the error is the same on heroku deploy) and enabled MAIL_URL as env var.

I had a really bad time with this. After correct MAIL_URL setup, I can run an email example from here https://github.com/themeteorchef/using-the-email-package, this is a email demo with email@1.1.18 that works just fine. 

So after implemented the very same in my project, it didn't send any emails, so I started to check differences and the only one is that my project installed email@1.2.1 and the demo runs on email@1.1.18.

Here is the very same demo using email@1.2.1 that fails https://github.com/mgscreativa/using-the-email-package. In this code, If I downgrade email to email@1.2.0 I can send emails just fine!!! maybe this is a regression of latest PR #8605 https://github.com/meteor/meteor/pull/8605/files#diff-8b2faad35ab5c720c37a39b84135fe14 (by the way, such a huge change for releasing a simple patch release, it should be named 1.3.0!)

My  MAIL_URL uses a google apps account like this MAIL_URL = 'smtp://user%40domain.com:Password_With_%40_Character@smtp.gmail.com:465';

Here is a complete forum post with more info if you need https://forums.meteor.com/t/bug-cant-send-emails-from-my-app-with-email-1-2-1/36505

edemaine:
@mgscreativa Change `smtp:` to `smtps:` in your MAIL_URL and email@1.2.1 should work.

I don't recommend using email@1.2.0 as the mail server will eventually timeout and cause email to crash.

mgscreativa:
Ups, that did the trick...

Can someone update the docs to point that?

Thank you!

mgscreativa:
Just thinking...why no error messaje is thrown? because the app never finished running email.send if I try to send with this  MAIL_URL = 'smtp://user%40domain.com:Password_With_%40_Character@smtp.gmail.com:465';

I think that mail.send should trow an error advising to use smtps when required...

something like: **You are trying to connect port 465 over smtp, try to change MAIL_URL to smpts**

What you think?

PS: please add that error message, I spent all my Saturday trying to figure out the issue, but no error messages sent!

edemaine:
I worry that an unconditional warning would be annoying if someone legitimately wanted to use SMTP on port 465... but I could go either way.  @abernix, care to weigh in?  In either case, I can PR a note about this in the documentation.

[...]

PR #8707, opened 2017-05-19
Email package improvements

edemaine:
I've made two changes to the Email package:
1. Using a MAIL_URL of the form `smtp://...:465` now warns that there's probably a bug (protocol should be `smtps:`). `email@1` used to automatically force `smtps:` protocol when port was 465, so this warning is useful for upgraders, without getting in the way of people who Know What They're Doing.  In response to #8696 and some later comments on #8591.
2. I tweaked `node4mailer` slightly to work on even older versions of Node 4. In response to https://github.com/chriswessels/meteor-tupperware/issues/32

I tested the warning in both activating and nonactivating cases, and tested that sending multiple messages only causes one warning.  (In fact, if `MAIL_URL` changes multiple times, each one may cause a warning.  But this seems good.)

### DIFF
--- a/packages/email/.npm/package/npm-shrinkwrap.json
+++ b/packages/email/.npm/package/npm-shrinkwrap.json
@@ -1,9 +1,9 @@
 {
   "dependencies": {
     "node4mailer": {
-      "version": "4.0.2",
-      "resolved": "https://registry.npmjs.org/node4mailer/-/node4mailer-4.0.2.tgz",
-      "from": "node4mailer@4.0.2"
+      "version": "4.0.3",
+      "resolved": "https://registry.npmjs.org/node4mailer/-/node4mailer-4.0.3.tgz",
+      "from": "node4mailer@4.0.3"
     },
     "stream-buffers": {
       "version": "0.2.5",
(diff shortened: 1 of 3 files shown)


hwillson:
When you have a chance @edemaine, would you mind re-running your CircleCI build?

abernix:
I pushed an empty commit to force the re-run and they appear to be passing now.

edemaine:
Thanks! Sorry for losing track of this.

abernix:
This has been published as `email@1.2.3`.  Thanks again for your attention to this, @edemaine.  😃

edemaine:
Awesome, thanks!  Happy to help.

edemaine:
Just a note for possible future versions of Meteor's `email` package: [nodemailer is no longer officially supported](http://nodemailer.com/status/), but [reactioncommerce is maintaining an official fork](https://github.com/chriswessels/meteor-tupperware/issues/32#issuecomment-309235420), including a Node 4 compatible version.  If there are any interesting changes in `reactioncommerce/nodemailer`, we should probably switch from my `node4mailer` to it.

PR #8710, opened 2017-05-22
Link ddp-rate-limiter README to the docs

lorensr:
Example was out of date; link was broken 😊


lorensr:
😞 aww, radio button disabled!

image

Issue #8721, opened 2017-05-27
meteor update --all-packages doesn't update all packages


Deozaan:
After updating Meteor to the latest version with `meteor update` I was informed that there was an update for a package and that I should run `meteor update --all-packages` to update it. That seemed like a good idea, so I tried it. This is what happened:

```
> meteor update --all-packages
This project is already at Meteor 1.4.4.3, the latest release.
Your top-level dependencies are at their latest compatible versions.

Newer versions of the following indirect dependencies are available:
 * ecmascript-runtime 0.3.15 (0.4.0 is available)
To update one or more of these packages, pass their names to
`meteor update`, or just run `meteor update --all-packages`.
```

In my opinion, running `meteor update --all-packages` should update *all packages*.

Or at the very least, that command should be renamed to something that accurately describes what it does. Perhaps something such as `--most-packages` or `--some-packages` or `--maybe-if-you-feel-like-updating-a-package-i-guess-like-whatever-man-you-do-you` would be more accurate. Additionally, the console shouldn't tell me that I can use a command to update packages when it won't update those packages after all.

dr-dimitru:
Same here, always end up with (since `v1.4.4.2`, including `1.4.4.3`):
```
This project is already at Meteor 1.4.4.3, the latest release.
Your top-level dependencies are at their latest compatible versions.
                                              
The following top-level dependencies were not updated to the very latest version available:
 * accounts-base 1.2.17 (1.3.0 is available)  
```

benjamn:
As of Meteor 1.4.3, core packages like `ecmascript-runtime` and `accounts-base` are constrained to the same minor version (e.g., `0.3.x` or `1.2.x`) that they had when the current release was published. See [further explanation in the `History.md` for Meteor 1.4.3](https://github.com/meteor/meteor/blob/devel/History.md#v143-2017-02-13).

Just because a newer version of a package has been published doesn't mean you should be able to update to it. Between Meteor 1.4 and Meteor 1.4.2.7 there were no constraints on core package versions, and it was nearly impossible to make any guarantees about the consistency of core packages. Before Meteor 1.4, core package versions were constrained to exactly one version, which was too restrictive. The current system is a practical middle-ground, because it lets you install new package versions with non-breaking changes, without updating your Meteor release.

[...]

Issue #8488, opened 2017-03-14
Updating indirect dependencies not working


steph643:
When I run `meteor update`, it seems 3 indirect dependencies could be updated:
```
C:\Users\steph\Meteor\myProject>meteor update
This project is already at Meteor 1.4.3.2, the latest release.
Your top-level dependencies are at their latest compatible versions.

Newer versions of the following indirect dependencies are available:
 * aldeed:collection2-core 1.2.0 (2.0.0 is available)
 * aldeed:schema-deny 1.1.0 (2.0.0 is available)
 * aldeed:schema-index 1.1.1 (2.0.0 is available)
To update one or more of these packages, pass their names to
`meteor update`, or just run `meteor update --all-packages`.
```
But using `--all-packages`, as recommended, seems to have no effect:
```
C:\Users\steph\Meteor\myProject>meteor update --all-packages
This project is already at Meteor 1.4.3.2, the latest release.
Your top-level dependencies are at their latest compatible versions.

Newer versions of the following indirect dependencies are available:
 * aldeed:collection2-core 1.2.0 (2.0.0 is available)
 * aldeed:schema-deny 1.1.0 (2.0.0 is available)
 * aldeed:schema-index 1.1.1 (2.0.0 is available)
To update one or more of these packages, pass their names to
`meteor update`, or just run `meteor update --all-packages`.
```
And updating individual packages, as recommended, seems to have no effect either:
```
C:\Users\steph\Meteor\myProject>meteor update aldeed:schema-deny
The specified packages are at their latest compatible versions.

C:\Users\steph\Meteor\myProject>meteor update
This project is already at Meteor 1.4.3.2, the latest release.
Your top-level dependencies are at their latest compatible versions.

Newer versions of the following indirect dependencies are available:
 * aldeed:collection2-core 1.2.0 (2.0.0 is available)
 * aldeed:schema-deny 1.1.0 (2.0.0 is available)
 * aldeed:schema-index 1.1.1 (2.0.0 is available)
To update one or more of these packages, pass their names to
`meteor update`, or just run `meteor update --all-packages`.
```
Meteor version: 1.4.3.2
Environment: Windows 7

abernix:
Sorry for the confusion here.  This isn't actually a Windows bug but rather a text bug – see #5270.  We improved the CLI `--help` text about this (in https://github.com/meteor/meteor/pull/8431) but we still need to discuss the exact logic for this message.

[...]

PR #8725, opened 2017-05-29
Clarify 'meteor update' message for indirect dependencies

jamesmillerburgess:
* Adjust string style to be in line with meteor style guide 6.2

* Partially addresses #8721 and #8488

After running 'meteor update', if there are newer, incompatible versions of indirect dependencies available, the console output seems to imply that these versions are compatible and can be updated using 'meteor update', when they cannot be. This leads the user to wonder if the 'meteor update' command is working properly or not, as the suggested solution does not eliminate what looks like a warning message.

In this PR I am trying to clarify that there is a possibility that the newer versions may not be compatible and to be more precise about what 'meteor update' actually does.

A better solution would be to proactively identify whether the newer versions are incompatible and tell the user, so that they don't have to run additional commands to find out that information. If it seems like a good idea, I could try to implement it in a subsequent PR.

I was not sure how exactly to follow the meteor code style guide for strings (section 6.2). My string is both multi-line (use template literals) and longer than 80 characters (use string concatenation), so I've used template literals and then concatenated them to respect the character limit.

This is my first ever open source PR. Please let me know if I have missed anything!

### DIFF
--- a/tools/cli/commands-packages.js
+++ b/tools/cli/commands-packages.js
@@ -1787,10 +1787,10 @@ main.registerCommand({
       Console.info("\nNewer versions of the following indirect dependencies" +
                    " are available:");
       _.each(nonlatestIndirectDeps, printItem);
-      Console.info([
-        "To update one or more of these packages, pass their names to ",
-        "`meteor update`, or just run `meteor update --all-packages`."
-      ].join("\n"));
+      Console.info(`These versions may not be compatible with your project.
+To update one or more of these packages to their latest compatible versions, 
+pass their names to \`meteor update\`, or just run \`meteor update ` +
+        `--all-packages\`.`);
     }
   }
 });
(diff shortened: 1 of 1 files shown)


tools/cli/commands-packages.js
```diff
@@ -1787,10 +1787,10 @@ main.registerCommand({
...
                    " are available:");
       _.each(nonlatestIndirectDeps, printItem);
-      Console.info([
-        "To update one or more of these packages, pass their names to ",
-        "`meteor update`, or just run `meteor update --all-packages`."
-      ].join("\n"));
+      Console.info(`These versions may not be compatible with your project.
+To update one or more of these packages to their latest compatible versions, 
+pass their names to \`meteor update\`, or just run \`meteor update ` +
+        `--all-packages\`.`);
```
benjamn: Given the presence of backticks in the string itself, I would recommend just sticking with the `[...].join("\n")` style in this case.

benjamn:
Merged into `release-1.5`: https://github.com/meteor/meteor/pull/8327/commits/6445a440018444a5490084147f88b64667ed8166. Thanks for improving this misleading messaging!

Issue #853, opened 2013-03-20
Provide this.unblock in publish context


mitar:
**_3 Upvotes**_ Provide `this.unblock` in `Meteor.publish` context, same as it is provided in `Meteor.methods`.

funkyeah:
I'm going to +1 this as well. My use case proactively subscribing to data the user is very likely to need at some point. Publications for subscriptions run at the template level should be to the minimum data needed and should block because it is required to render. 

The programmer often can anticipate what data the user will want or need and also knows what data should be kept around. In my opinion these cases often lead to wanting to subscribe to larger super-sets of data outside of the template context.

Because this data isn't immediately needed, its best if the DDP connections transporting the superset data doesn't block DDP connections transporting the data required to render a given template.

I've tried to explain an example of this in an [issue](https://github.com/meteorhacks/unblock/issues/4) file in arunodas this.unblock repo.

dandv:
+1 for core support given the increasing usage of template-level subscriptions.

dohomi:
+1 for this. I guess thinking in components way of template-level subscription this would be a great benefit.

ShuyinBlack:
+1 switched my app's architecture to template-level subscriptions to improve performance. Using this.unblock() package from arunoda speeds up response times even more, very easily. So why not integrate into Meteor. : )

tmeasday:
I just took a look at this again. @arunoda's package does exactly what we need here, in that if you stop (`UNSUB`) the publication before the `onStop()` handler is setup (i.e. before the publish body is completed), the `onStop()` handler runs anyway, immediately. I think this is pretty transparently exactly what you want. Sure there are some cases where you might be doing a lot of work:

```js
Meteor.publish(function() {
   this.unblock();
   const r1 = doSlowStepOne();
   // If the pub has been stopped at this point, it might be useful to abort here and not do the unnecessary step 2.
   const r2 = doSlowStepTwo(r1);
   this.added(r2);
   this.ready();
});
```

But you can do that by defining your `onStop()` early and carefully handling the various cases.

More than that, I think @mitar asked the pertinent question:

> After some more though: why are subscriptions even blocking to begin with? Is there really an use case where you would want it to block other subscriptions? Why it not unblocked default

I think this is true. I can't think of a case where I'd want a sub to block. As @glasser has said, the default use case (returning cursors) doesn't block. Why would you want it to? It makes sense for methods, not for pubs IMO.

[...]

PR #8734, opened 2017-05-31
Bind environment of observeChanges callbacks

zimme:
This will use `bindEnvironment` on `observe` and `observeChanges` callbacks.
This will allow `Meteor.userId()` to be used within these callbacks.

See this conversation for background.
https://github.com/meteor/meteor/pull/8629#issuecomment-300084704

/cc @mitar

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,10 @@
 ## v.NEXT
 
+* `observe`/`observeChanges` callbacks are now bound.
+  The same `EnvironmentVariable`s that was present when
+  `observe`/`observeChanges` was called is now available inside the callbacks.
+  [PR #8734](https://github.com/meteor/meteor/pull/8734)
+
 * `reactive-dict` now supports setting initial data when defining a named
   `ReactiveDict`. No longer run migration logic when used on the server,
   this is to prevent duplicate name error on reloads. Initial data is now
--- a/History.md
+++ b/History.md
@@ -1,8 +1,8 @@
 ## v.NEXT
 
-* `observe`/`observeChanges` callbacks are now bound.
-  The same `EnvironmentVariable`s that was present when
-  `observe`/`observeChanges` was called is now available inside the callbacks.
+* `observe`/`observeChanges` callbacks are now bound using `Meteor.bindEnvironment`.
+  The same `EnvironmentVariable`s that were present when `observe`/`observeChanges`
+  was called are now available inside the callbacks.
   [PR #8734](https://github.com/meteor/meteor/pull/8734)
 
 * `reactive-dict` now supports setting initial data when defining a named
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -949,7 +949,23 @@ Cursor.prototype.observe = function (callbacks) {
 
 Cursor.prototype.observeChanges = function (callbacks) {
   var self = this;
+  var methods = [
+    'addedAt',
+    'added',
+    'changedAt',
+    'changed',
+    'removedAt',
+    'removed',
+    'movedTo'
+  ];
   var ordered = LocalCollection._observeChangesCallbacksAreOrdered(callbacks);
+
+  methods.forEach(function (method) {
+    if (callbacks[method] && typeof callbacks[method] == "function") {
+      callbacks[method] = Meteor.bindEnvironment(callbacks[method]);
+    }
+  });
+  
   return self._mongo._observeChanges(
     self._cursorDescription, ordered, callbacks);
 };
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -960,9 +960,11 @@ Cursor.prototype.observeChanges = function (callbacks) {
   ];
   var ordered = LocalCollection._observeChangesCallbacksAreOrdered(callbacks);
 
+  // XXX: Can we find out if callbacks are from observe?
+  var exceptionName = ' observe/observeChanges callback'; 
   methods.forEach(function (method) {
     if (callbacks[method] && typeof callbacks[method] == "function") {
-      callbacks[method] = Meteor.bindEnvironment(callbacks[method]);
+      callbacks[method] = Meteor.bindEnvironment(callbacks[method], method + exceptionName);
     }
   });
   
--- a/packages/mongo/observe_changes_tests.js
+++ b/packages/mongo/observe_changes_tests.js
@@ -394,3 +394,23 @@ testAsyncMulti("observeChanges - bad query", [
     f2.wait();
   }
 ]);
+
+if (Meteor.isServer) {
+  Tinytest.addAsync(
+    "observeChanges - EnvironmentVariable",
+    function (test, onComplete) {
+      var c = makeCollection();
+      var environmentVariable = new Meteor.EnvironmentVariable;
+      environmentVariable.withValue(true, function() {
+        var handle = c.find({}, { fields: { 'type.name': 1 }}).observeChanges({
+          added: function() {
+            test.isTrue(environmentVariable.get());
+            handle.stop();
+            onComplete();
+          }
+        });
+      });
+      c.insert({ type: { name: 'foobar' } });
+    }
+  );  
+}


packages/mongo/mongo_driver.js
```diff
@@ -949,7 +949,23 @@ Cursor.prototype.observe = function (callbacks) {
...
+    'changed',
+    'removedAt',
+    'removed',
+    'movedTo'
+  ];
   var ordered = LocalCollection._observeChangesCallbacksAreOrdered(callbacks);
+
+  methods.forEach(function (method) {
+    if (callbacks[method] && typeof callbacks[method] == "function") {
+      callbacks[method] = Meteor.bindEnvironment(callbacks[method]);
```
mitar: This could give us opportunity to also handle errors from those callbacks better. Do we want to do anything about that? Stop the publication? Call a callback on the publication? :-)

mitar: (But it is true that not all observes are called from the publication.

mitar:
Are you sure `observe` is handled? That was just my memory from the past. But if you checked, then great.

mitar:
Does currently `observeChanges` have to be called from a Fiber?

packages/mongo/mongo_driver.js
zimme: I don't think we should tie this to publications in that we automatically stop publications. People can just use `self.stop()` or arrow functions and `this.stop()` to do that on their own?

zimme:
> Are you sure observe is handled? That was just my memory from the past. But if you checked, then great.

Yep, `observe` uses `observeChanges` internally after a detour through `LocalCollection` first.

> Does currently observeChanges have to be called from a Fiber?

Not sure, will look into this.

packages/mongo/mongo_driver.js
mitar: No, I was more thinking of providing a new callback on publish, like `onError`, and then it gets called. If you want to stop publish on error, you can call `this.stop()` yourself.

mitar: The point is, if observe callbacks throws an error, currently there is no easy way to handle that.

zimme: I started to look into this a bit more but the code gets complex quite quick and I don't feel I have time to dig into that right now so I'd say let's keep this PR down to just binding the environment for the callbacks in the best way possible and open an issue for discussing that a better error handling strategy would be.

zimme:
> Does currently observeChanges have to be called from a Fiber?

Dug into the code a bit more and once i got into this function `addHandleAndSendInitialAdds` it looks like it's using some Future function.

packages/mongo/mongo_driver.js
```diff
@@ -949,7 +949,23 @@ Cursor.prototype.observe = function (callbacks) {
...
   var self = this;
+  var methods = [
+    'addedAt',
+    'added',
+    'changedAt',
+    'changed',
+    'removedAt',
+    'removed',
+    'movedTo'
+  ];
```
zimme: Can we iterate over the properties of callbacks to bindEnvironment without explicitly using a list of methods that needs updating if they ever change? Is this available in some common internal variable maybe?

zimme:
I found this comment while going thought some old issues.
https://github.com/meteor/meteor/issues/853#issuecomment-25515291
It kinds confused me as this is what we're trying to do with this PR and from that comment that was at some point the default?

packages/mongo/mongo_driver.js
mitar: In minimongo it is made even uglier: https://github.com/meteor/meteor/blob/d6c4281eebe8fffcdeb415ccc361784f854126f4/packages/minimongo/minimongo.js#L381

I think a list is OK. I would not go wrapping everything because there might be other things there.

mitar:
Hm, crazy question: did you try to access DDP variables inside the observe callbacks? Maybe they are already available? (You have to check both first and non-first run of a callback to be sure.)

I think in that comment the keyword is **implicit**. But I would say that if it does not work, then explicit is better. :-) Just to make sure.

zimme:
> Hm, crazy question: did you try to access DDP variables inside the observe callbacks? Maybe they are already available? (You have to check both first and non-first run of a callback to be sure.)

I've tried to access `DDP._CurrentPublicationInvocation` inside of both `observe` and `observeChanges` callbacks without this patch both on first run and consecutive runs and it's not available.
With this patch, it works as expected. So from what i can tell the environment isn't bound implicitly either. Maybe it was the default at sometime or he was talking about something else in that comment.

zimme:
This PR should be done.

History.md
```diff
@@ -1,5 +1,10 @@
 ## v.NEXT
 
+* `observe`/`observeChanges` callbacks are now bound.
```
mitar: Maybe explain that this is on the server side. And what exactly means "bound"? To `this`? Or something else. ;-) (I know what it is, I am just saying that it is unclear for readers.)

BTW, what is `this` bound to inside callbacks? Should we bound them to something?

History.md
```diff
@@ -1,5 +1,10 @@
...
+  The same `EnvironmentVariable`s that was present when
+  `observe`/`observeChanges` was called is now available inside the callbacks.
```
mitar: I think we should explain what binding environment is in documentation (see https://github.com/meteor/docs/issues/139) and then here just say:

>  `observe`/`observeChanges` callbacks have now their Meteor environment bound.

With "Meteor environment bound" linking to description of `Meteor.bindEnvironment`.

packages/mongo/mongo_driver.js
mitar: If we are not supporting intercepting errors, maybe then we should at least name the error handler? Like "cursor observe"?

History.md
zimme: Will try and elaborate a bit.

`this` isn't set to anything specific, and if we were to change it I don't know to what.

mitar: Crazy idea: to `DDP._CurrentPublicationInvocation.get()`.

History.md
zimme: This makes sense, maybe we should hold of on this PR until we have that documentation instead?

History.md
mitar: (This would go into effect only if you have not bound your function yourself to some other `this`.)

zimme: What if you're not using `observe`/`observeChanges` from within a `publication`?

mitar: Then it will be `null`.

zimme: I'd say, let's do this in a separate PR. `observe` and `observeChanges` do have separate `this` so jus adding a context to `bindEnvironment` isn't going to set the context for both `observe` and `observeChanges`.

mitar: OK.

packages/mongo/mongo_driver.js
zimme: There, used named exception handler. Should be a bit easier to follow the stack trace.

packages/mongo/mongo_driver.js
zimme: :+1:

History.md
```diff
@@ -1,5 +1,10 @@
 ## v.NEXT
 
+* `observe`/`observeChanges` callbacks are now bound using `Meteor.bindEnvironment`.
+  The same `EnvironmentVariable`s that was present when `observe`/`observeChanges`
+  was called is now available inside the callbacks.
```
mitar: "are now available".

History.md
```diff
@@ -1,5 +1,10 @@
 ## v.NEXT
 
+* `observe`/`observeChanges` callbacks are now bound using `Meteor.bindEnvironment`.
+  The same `EnvironmentVariable`s that was present when `observe`/`observeChanges`
```
mitar: was -> were

zimme: Fixed

mitar:
So who merges? Can I?

hwillson:
We were just waiting for the tests to pass (they were re-running). They failed again, but the failure isn't related to this PR. I've restarted them, so let's give them another change before merging. Thanks!

zimme:
The tests needs to be re-run again, do we know why circle-ci get segmentation faults so often?

PR #8735, opened 2017-06-01
Add --bundle-visualizer param

mpowaga:
This let's you run meteor app with bundle-visualizer package without touching `.meteor/packages`. 
Example usage: `meteor --production --bundle-visualizer`.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -270,7 +270,9 @@ var runCommandOptions = {
     'no-lint': { type: Boolean },
     // Allow the version solver to make breaking changes to the versions
     // of top-level dependencies.
-    'allow-incompatible-update': { type: Boolean }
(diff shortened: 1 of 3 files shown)


abernix:
While I appreciate the effort here (and thank you for working to improve Meteor!), this doesn't address the fact that `bundle-visualizer` was never meant to be the only way to analyze the size of a bundle and this won't fix the problem for the next bundle analysis tool that comes along.

In the future, please open an issue and discuss your concerns with the current implementation (and to understand the reasons for why things are the way they are) before submitting a pull-request to change them.  For example, perhaps an option similar to what you are suggesting here which was called `--addl-packages` could be a better, more universal solution for all bundle analysis packages (and maybe other things as well!).  Thoughts?

mpowaga:
Thanks. I've opened new issue #8739 as recommended.

PR #8736, opened 2017-06-01
Updating history a bit for #8629

mitar:
I think it is cleaner in this way. Because now methods do have `this.connection` available.

cc @zimme


zimme:
LGTM :+1:

abernix:
Thanks!  I made a couple small formatting, punctuation and reordering changes.

PR #8752, opened 2017-06-02
Provide support for setting Mongo SSL settings in the Environment

bliles:
In some environments it can be beneficial to support SSL connections to MongoDB using a self-signed certificate or a private certificate authority. Meteor does not currently provide good mechanism to allow a developer to specify these SSL settings for the MongoDB connection.

References https://github.com/meteor/meteor/issues/6958 where a method for configuring Mongo is desired.

### DIFF
--- a/packages/mongo/remote_collection_driver.js
+++ b/packages/mongo/remote_collection_driver.js
@@ -1,3 +1,5 @@
+var fs = require('fs');
+
 MongoInternals.RemoteCollectionDriver = function (
   mongo_url, options) {
   var self = this;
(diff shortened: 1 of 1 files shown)


hwillson:
Hi @bliles - first off, thanks for submitting this PR! Looking over your proposed changes, I have a few questions:

1. Meteor 1.4 (PR #7277) added a way to pass in custom Mongo options, by using `Mongo.setConnectionOptions`. There has since been some debate in issue #6958 about whether this API works in all cases. Are you submitting this PR because you've found that the `Mongo.setConnectionOptions` API doesn't work for certain scenarios (like when using accounts)?

2. In your proposed changes, you're manipulating the `connectionOptions` object. This object is first passed into [`MongoInternals.RemoteCollectionDriver`](https://github.com/meteor/meteor/blob/76a6e0eee91c4e85daabb679bad9a26115f488ec/packages/mongo/remote_collection_driver.js#L38), then passed into a [`new MongoConnection`](https://github.com/meteor/meteor/blob/76a6e0eee91c4e85daabb679bad9a26115f488ec/packages/mongo/remote_collection_driver.js#L4) call. If you dive into the [`MongoConnection`](https://github.com/meteor/meteor/blob/76a6e0eee91c4e85daabb679bad9a26115f488ec/packages/mongo/mongo_driver.js#L125) constructor function, you'll see that the incoming options (so the `connectionOptions` object) are copied individually (as needed) into a new `mongoOptions` object. That `mongoOptions` object is then passed into the MongoDB `connect` call, to set the Mongo options. Long story short, the changes you're proposing on the `connectionOptions` object are never copied into the `mongoOptions` object, so they don't appear to be passed to Mongo (or used anywhere). Have you verified that these changes are working?

Thanks!

bliles:
I'm sorry, when I read over https://github.com/meteor/meteor/issues/6958 I thought that the only path was still to adjust the connection options on the collection, I hadn't found Mongo.setConnectionOptions.

I didn't do enough homework.

hwillson:
No problem @bliles - if you notice any issues using `Mongo.setConnectionOptions`, definitely post back and let us know. Thanks!

bliles:
Created PR to add setConnectionOptions to the tsd. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/16922

Issue #8757, opened 2017-06-04
Cannot have two user collections in the same database


ashburnham:
I think that Meteor can be improved by adding a `userCollectionName` option to the accounts-base package. The reason I think this is because:

 - Meteor only allows connection to one database by default
 - There is a use case (I'm facing it!) where you are running multiple meteor apps from a single database where each app has a completely different type of user.

Currently these users are forced into the same collection by the constraints of Meteor, the solution would be to allow them to be stored in two separate collections.

Out use case comes from the need for a customer app, through which our customers interact with us, and a separate app (with completely different codebase/functionality) which our internal team use for various back-office functions. Because they use the same database we are currently forced to mix the two types of users, from the two separate apps, in a single collection. As they are separate entities, we believe this to be bad practice and would like Meteor to allow us to store them in separate collections by setting a `userCollectionName` package.

This option would set the name of the collection assigned to `Accounts.users` and `Meteor.users`. As it is configurable by app, it allows the different apps to have separate user collections in the same database.

dr-dimitru:
@ashburnham I'm sorry, but is there a real purpose to run multiple Meteor apps from a single DB and expect them to have different collections?

In this case I may suggest to Meteor to have default prefix added to each collection, which still a bit weird.

 1. Why not use separate DB in the same `mongod`?
 2. Why not connect app to the other (*second*) MongoDB and use data you need?

laosb:
Though the reason provided is not very convincing, I agree that we should give developers the option, in case they want a different collection for some reasons.

crapthings:
yes you can prefix user login field to simulate multiple user collections

then you auto append prefix field to login event

you can make every thing separate by ddp ...

something like server picker can keep rooturl while u refresh page

https://github.com/partus/meteor-server-picker/blob/master/main.js

klaussner:
> […] is there a real purpose to run multiple Meteor apps from a single DB and expect them to have different collections?

Using the same database for two apps could make sense if data is shared between both apps. However, I think that mixing two types of users in a single collection is not necessarily bad practice.

If you validate logins with [`validateLoginAttempt`](http://docs.meteor.com/api/accounts-multi.html#AccountsServer-validateLoginAttempt) to make sure that members of the dev team cannot log in as customers and vice versa, the two user types would be practically separate, even though they are stored in one collection.

zimme:
[...]

PR #8758, opened 2017-06-04
Added userCollectionName option to Accounts

ashburnham:
I think that Meteor can be improved by adding a userCollectionName option to the accounts-base package. The reason I think this is because:

Meteor only allows connection to one database by default
There is a use case (I'm facing it!) where you are running multiple meteor apps from a single database where each app has a completely different type of user.
Currently these users are forced into the same collection by the constraints of Meteor, the solution would be to allow them to be stored in two separate collections.

Out use case comes from the need for a customer app, through which our customers interact with us, and a separate app (with completely different codebase/functionality) which our internal team use for various back-office functions. Because they use the same database we are currently forced to mix the two types of users, from the two separate apps, in a single collection. As they are separate entities, we believe this to be bad practice and would like Meteor to allow us to store them in separate collections by setting a userCollectionName package.

This option would set the name of the collection assigned to Accounts.users and Meteor.users. As it is configurable by app, it allows the different apps to have separate user collections in the same database.

https://github.com/meteor/meteor/issues/8757

### DIFF
--- a/packages/accounts-base/accounts_common.js
+++ b/packages/accounts-base/accounts_common.js
@@ -20,7 +20,7 @@ export class AccountsCommon {
 
     // There is an allow call in accounts_server.js that restricts writes to
     // this collection.
-    this.users = new Mongo.Collection("users", {
+    this.users = new Mongo.Collection(options.userCollectionName || "users", {
       _preventAutopublish: true,
       connection: this.connection
     });
@@ -89,6 +89,9 @@ export class AccountsCommon {
   // - ambiguousErrorMessages {Boolean}
   //     Return ambiguous error messages from login failures to prevent
   //     user enumeration.
+  // - userCollectionName {Boolean}
+  //     Sets the name of the collection where users are stored.
+  //     Defaults to "users";
 
   /**
    * @summary Set global accounts options.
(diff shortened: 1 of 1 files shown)


zimme:
I'm just thinking out loud here, but wouldn't it be good to also/instead support an option to provide the specific collection and just not a name for it. This way you could create a specific collection with any connection you want and that would be used for the accounts system.

dr-dimitru:
+1 for @zimme suggestion. Let's keep both options: `collectionName ` and `collection` (*I believe no need to add `user` to those properties as `Accounts` package is all about users*)

hwillson:
Hi @ashburnham - First off, thanks for submitting this PR! 

As mentioned in the [Proposing your change](https://github.com/meteor/meteor/blob/devel/Contributing.md#proposing-your-change) section of the [Contributing](https://github.com/meteor/meteor/blob/devel/Contributing.md) doc, we really prefer to see proposals like this first discussed through a well defined issue. This gives both MDG and the community a chance to fully analyze your proposal, and bounce around ideas that ultimately lead  towards implementing the best path forward for your suggested changes. In this case, you opened the PR at the same time you opened the issue, which means the upfront discussion didn't happen. As you can see from the discussion that has happened since in #8757, people have different ideas about how your suggested functionality could be best implemented. As such, I'll close this PR for now. Let's continue the discussion in #8757, and hash out a plan of attack that everyone agrees on. When everyone is ready to move forward, #8757 will be marked as `confirmed` and `pull-requests-encouraged`. Thanks!

PR #8760, opened 2017-06-05
Change default `from` email for `accounts-password` to an "example".

abernix:
Currently, the default "from" address for `meteor-accounts` is
"no-reply@meteor.com".  While this works for many users, and granted it
is a "no reply" address and there should be no expectation of the
address working, it contributes to a negative spam rating for the
"meteor.com" domain and a surplus of extra e-mail.

The correct way to set the default "from" address is by setting:

```js
Accounts.emailTemplates.from = "Name <email@domain.com>";
```

As per the documentation here:

* https://docs.meteor.com/api/passwords.html#Accounts-emailTemplates

By changing it to "example.com", and making a more obvious "example"
out of the name ("Accounts Example"), it should encourage users to
actually change the address to something more reasonable especially
since many e-mail providers will also reject mail coming from
"example.com", which should provide a clear warning to those who have
their e-mail misconfigured.

/cc @n1mmy

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,6 +5,14 @@
   this is to prevent duplicate name error on reloads. Initial data is now
   properly serialized.
 
+* `accounts-password` now uses `example.com` as a default "from" address instead
+  of `meteor.com`. This change could break account-related e-mail notifications
+  (forgot password, activation, etc.) for applications which do not properly
+  configure a "from" domain since e-mail providers will often reject mail sent
+  from `example.com`. Ensure that `Accounts.emailTemplates.from` is set to a
+  proper domain in all applications.
+  [PR #8760](https://github.com/meteor/meteor/issues/8760)
+
 * Add `DDP._CurrentPublicationInvocation` and `DDP._CurrentMethodInvocation`.
   `DDP._CurrentInvocation` remains for backwards-compatibility. This change
   allows method calls from publications to inherit the `connection` from the
--- a/packages/accounts-password/email_templates.js
+++ b/packages/accounts-password/email_templates.js
@@ -19,7 +19,7 @@ Thanks.
  * @importFromPackage accounts-base
  */
 Accounts.emailTemplates = {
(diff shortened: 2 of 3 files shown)


n1mmy:
Thanks, @abernix ! 👍

dr-dimitru:
@abernix could we use an invalid value, `localhost` or even take a value from `ROOT_URL` ?
Wouldn't this lead to a negative spam rating for example.com

> This domain is established to be used for illustrative examples in documents. You may use this domain in examples without prior coordination or asking for permission.

This is not a docs, it's working code.

abernix:
@dr-dimitru `localhost` is not an invalid value for a "from" address and I'd rather not make assumptions about their mail handling desires by inferring a value from `MAIL_URL`.

`example.com`, on the other hand, is a reserved second-level domain [per RFC 2606](https://tools.ietf.org/html/rfc2606#section-3) and respected [by IANA](https://www.iana.org/domains/reserved).  I don't know any reputable e-mail providers that will accept mail from it (most result in an immediate failure, even before trying to route it).

While the page at example.com says it's for documents, the RFC doesn't make that specific designation and the domain has no MX records – so besides the fact that it's not deliverable, its spam rating doesn't matter.  All in all, I think example.com is a safer bet.

dr-dimitru:
@abernix agree.

abernix:
As this _could_ be breaking for anyone who hasn't configured `Accounts.emailTemplates.from` (but instead left it at the default of `meteor.com`), I've bumped the version of `accounts-password` to 2.0.0 and added a `History.md` entry (see above).  Therefore, this will be released with Meteor 1.5.1 since major version numbers are constrained by the `meteor-tool` release.

abernix:
I've verified that an error is thrown by `email` (rather than just remaining completely quiet about it) when the MTA/mail provider refuses to accept delivery for a particular host (like `example.com`).  This was the last thing waiting to be verified before this was merged.

Issue #8739, opened 2017-06-01
Add option to include additional packages


mpowaga:
In some cases you might want to run meteor with additional packages that are not meant to be saved in `.meteor/packages`. A good example is `bundle-visualizer` package. If you want to temporarily run your app with `bundle-visualizer` these 3 commands are needed:

1. `meteor add bundle-visualizer`
2. `meteor run`
3. `meteor remove bundle-visualizer`

This could be reduced to (as proposed by @abernix in #8735):

`meteor run --addl-packages bundle-visualizer`

Perhaps this option could be added to build command as well.

mpowaga:
Or `--extra-packages` to put more emphasis on a fact that these are additional packages.

dr-dimitru:
I would prefer to have `dependencies` and `devDependencies` as used in NPM. But see no way how this can be done in Meteor.

merlinpatt:
Could bundle-visualizer be made into an NPM package?

benjamn:
@merlinpatt Maybe so! It would be a little awkward because the package uses dynamic `import(...)` syntax, and we don't currently compile modules in `node_modules`, so the npm implementation might have to call `module.dynamicImport` directly, which might make the package useful only in Meteor, but it could work.

merlinpatt:
The other option I can think of would be to add another flag option to packages. I seem to recall that you can mark a package as production only or dev only, so there would need to be a weird 3rd in between option of "production only when also in debug mode."

I'd opt more for making it an npm package. One, as far as I'm aware, it is still the eventual plan to convert meteor over to npm, so this option would align with that goal. Two, even if it's only useful for Meteor, i think that's fine as there are plenty of packages that are only useful to certain frameworks. Three, someone may figure out how to make it work with Meteor and with other frameworks, and putting it on npm early would make it easier for that to happen.

abernix:
At the very least, I'm considering making the sunburst portion an npm package (as `@meteorjs/bundle-visualizer-sunburst`, perhaps?) to encourage anyone who might want to render the same chart in other tools – for example, if [Meteor Dev Tools for Chrome](https://github.com/bakery/meteor-devtools) wanted to include this in the Chrome Dev Tools – they wouldn't need to have to re-write code which could consume the `.stats.json` format.  I was thinking at first to just `Npm.depends` on said npm from `bundle-visualizer`.

@merlinpatt Did you have a particular use case?

abernix:
[...]

PR #8769, opened 2017-06-06
Add option to include additional packages

mpowaga:
This is rework of #8735 as discussed in #8739.

Example usage:

```sh
meteor run --extra-packages "bundle-visualizer, jquery@1.11.10"
```

Things to consider:

  - should be renamed to `--with-packages` or other name?
  - should be added to `meteor build`?
  - in case when the same package is already specified in `.meteor/packages` but with different version constraint should `--extra-packages` override it?


tools/cli/commands.js
```diff
@@ -289,10 +290,16 @@ function doRunCommand(options) {
   const { parsedServerUrl, parsedMobileServerUrl } =
     parseServerOptionsForRunCommand(options, runTargets);
 
+  var includePackages = [];
+  if (options['extra-packages']) {
+    includePackages = options['extra-packages'].split(',');
```
hwillson: You might want to strip blank spaces after your comma split, to handle your example of "package1, package2" properly. I know you're trimming later on down the chain in your code changes, but it might be useful to just trim right alongside the split. So something like:

```js
includePackages = options['extra-packages'].split(',').map(item => item.trim());
```

tools/cli/help.txt
```diff
@@ -86,6 +86,7 @@ Options:
                    downgraded to versions that are potentially incompatible with
                    the current versions, if required to satisfy all package
                    version constraints.
+  --extra-packages Run with additional packages (comma separated)
```
hwillson: Maybe add an example here (similar to the `--mobile-server` option help):

```
--extra-packages Run with additional packages (comma separated, for example: --extra-packages "package-name1, package-name2@1.2.3")
```

tools/cli/commands.js
abernix: Agree with this, but I think I'd have a strong preference to:

```js
includePackages = options['extra-packages'].split(/\s+,\s+/)
```
...which I _believe_ would be more performant.

mpowaga:
Added this option to `test` and `test-packages` commands as well.

Extra packages will now override constraints specified in `.meteor/packages`. This added a little bit of complexity as I had to add `skipOnRead` property to mark constraints that were overridden. There's also a danger that extra packages will be written to `.meteor/packages`. To prevent this I added another property `skipOnWrite`. Perhaps this should be covered by tests.

I also created `selftest`'s. Though I'm not sure if they are in the right place since they do a little more than just checking "argument parsing".

/cc @hwillson @benjamn

tools/project-context.js
```diff
@@ -985,7 +1013,7 @@ _.extend(exports.ProjectConstraintsFile.prototype, {
   eachConstraint: function (iterator) {
     var self = this;
     _.each(self._constraintLines, function (lineRecord) {
-      if (lineRecord.constraint)
+      if (! lineRecord.skipOnRead && lineRecord.constraint)
```
zimme: If `eachConstraint` ignores `extra-packages` maybe this needs to happen in `getConstraint` and the other functions too? So should this really be skipping in this case?

mpowaga: `eachConstraint` iterates through `self._constraintLines` which may contain duplicates. This is because it needs to preserve all constraints from `.meteor/packages` (so this file can be later reproduced by `_write` method) as well as `self._includePackages`.

On the contrary, `getConstraint` reads from `self._constraintMap` which contains unique values.

Maybe there should be separate property for extra constraints instead of pushing them to `self._constraintLines`? For example `self._extraConstraintLines`. But then `eachConstraint` would need iterate through `self._extraConstraintLines` and `self._constraintLines` and find duplicates along the way.

zimme: > eachConstraint iterates through self._constraintLines which may contain duplicate packages. This is because it needs to preserve all constraints from .meteor/packages (so this file can be later reproduced by _write method) as well as self._includePackages.
> 
> On the contrary, getConstraint reads from self._constraintMap which contains unique packages.

Sounds reasonable.



> Maybe there should be separate property for extra constraints instead of pushing them to self._constraintLines? For example self._extraConstraintLines. But then eachConstraint would need iterate through self._extraConstraintLines and self._constraintLines and find duplicates along the way.

Feels like that might be overkill but I'm not too familiar with this part of the code.

mpowaga: > Feels like that might be overkill but I'm not too familiar with this part of the code.

Yes, that would be unnecessarily complex solution.

Another option would be to iterate over values of `self._constraintMap`. But I'm not sure if `eachConstraint` must iterate with correct order (first-in, first-out) and if so whether this would be still guaranteed.

zimme: > But I'm not sure if eachConstraint must iterate with correct order (first-in, first-out; as defined in .meteor/packages)

I'm pretty sure that ordering of the packages in `.meteor/packages` matters.

zimme:
This PR only have changes that affect `projectConstraintsFile` and not `packageMapFile` which represents the `.meteor/versions` file. Will that be an issue?

I'm thinking about meteor/meteor-feature-requests#31, where we're waiting for this PR.
We need to search though all packages including indirect dependencies and we're thinking that looking at `.meteor/versions` is the best way to do so.

It would be nice if we can just look through `packageMapFile` and in doing so it will include the `extra-packages` as well, which means this PR would have to add the extra packages to `packageMapFile` to.

zimme:
Related to my previous comment.

Turns out we might use `projectContext.resolveConstraints()` and `packageMap.getInfo('<package-name>')`. `resolveConstraints()` actually use `eachConstraints` and if `extra-packages` are ignored there we will have to look for the extra packages somewhere else.

This actually made be think that the constraint resolver will ignore extra packages if we ignore them in `eachConstraints`, will this have unwanted side-effects?

mpowaga:
> constraint resolver will ignore extra packages if we ignore them in `eachConstraints`

`eachConstraint` doesn't ignore extra packages. It only filters out overridden packages. So if you have `jquery` in `.meteor/packages` and run meteor with `--extra-packages "jquery@1.11.0"` it will return only one instance of jquery constraint, i.e. `jquery@1.11.0`.

zimme:
ah, i mixed up skipOnRead and skipOnWrite. Then I have no more concerns.

zimme:
Unless anyone can see a downside in doing the same with `packageMapFile` as you're doing with `projectConstraintsFile` I would really want to see that in this PR.

It's related to meteor/meteor-feature-requests#31.
If we do the same with `packageMapFile` as with `projectConstraintsFile` we can just use a `find()`/`indexOf` on `packageMapFile.getCachedVersions().keys()` to see if the package we're looking for is in the project or not, even if it's not a top-level dependency.

We thought about using `packageMap` but for that we need to run `resolveConstraints()` which is "expensive" and would add to the startup time.

hwillson:
Thanks for your work on this @mpowaga - looks like all outstanding requests/suggestions have been addressed (and your `self-test`'s look great!).

@zimme Based on your findings in https://github.com/meteor/meteor-feature-requests/issues/31, do you have any remaining concerns about the changes in this PR?

hwillson:
@mpowaga Actually, one small nitpick - could you wrap the help text example a little earlier, to keep it under 80 chars wide? Maybe just move everything after `example:` to a new indented line.

screenshot 2017-06-14 09 02 22

zimme:
Right now this PR only adds the extra packages to `projectConstraintsFile` (`.meteor/packages)` with the filtering logic (`skipOnRead`/`skipOnWrite`) to not save the extra packages to the file and to not iterate over duplicate packages because a package was overridden via `extra-packages`.

In meteor/meteor-feature-requests#31 I'm looking through `packageMapFile` (`.meteor/versions`) for the package as that includes all packages in the project including dependencies. So adding the extra packages to `packageMapFile` with the same filter logic would be nice because then I can just use `packageMapFile.getInfo()` to see if the package is available in the project as a direct or indirect dependency.

zimme:
> @mpowaga Actually, one small nitpick - could you wrap the help text example a little earlier, to keep it under 80 chars wide? Maybe just move everything after example: to a new indented line.

How about using `package1, package2@1.2.3`?

mpowaga:
@hwillson Something like this:

```
  --extra-packages Run with additional packages (comma separated, for example:
                   --extra-packages "package-name1, package-name2@1.2.3")
```

?

hwillson:
Yes, that formatting looks great - thanks!

hwillson:
Thanks for update @zimme - makes sense. Given that (I think) we're really close to merging this PR, maybe we should merge as is (to get the `--extra-packages` benefit sooner), then re-visit adding `packageMapFile` support afterwards. We have a triage call today and this PR will be discussed, so we'll see what the consensus is. Thanks again for the details!

zimme:
I would have no problem adding the extended support for `extra-packages` in the branch I've created for meteor/meteor-feature-requests#31.

hwillson:
Hi all - I've added a `History.md` entry, so as soon as the tests re-run, we should be good to go here. LGTM!

Issue #8485, opened 2017-03-14
[Windows] Build process sometimes crashes


steph643:
I get this exception randomly (but quite often) when recompiling:

```text
C:\Users\steph\AppData\Local\.meteor\packages\coffeescript\1.12.3_1\plugin.compileCoffeescript.os\npm\node_modules\meteor\promise\node_modules\meteor-promise\promise_server.js:190
      throw error;
      ^

Error: EBUSY: resource busy or locked, symlink 'C:\Users\steph\AppData\Local\.meteor\packages\logging\1.1.17\npm\node_modules' -> 'C:\Users\steph\Travail\Meteor\myProject\.meteor\local\.build697476.build\programs\.build324432.server\npm\node_modules\meteor\logging\node_modules'
    at Error (native)
    at Object.fs.symlinkSync (fs.js:897:18)
    at Object.wrapper (C:\tools\fs\files.js:1535:35)
    at symlinkWithOverwrite (C:\tools\isobuild\builder.js:740:5)
    at walk (C:\tools\isobuild\builder.js:458:9)
    at Builder.copyDirectory (C:\tools\isobuild\builder.js:594:5)
    at C:\tools\isobuild\bundler.js:2050:17
    at Function._.each._.forEach (C:\Users\steph\AppData\Local\.meteor\packages\meteor-tool\1.4.3_1\mt-os.windows.x86_32\dev_bundle\lib\node_modules\underscore\underscore.js:87:22)
    at JsImage.write (C:\tools\isobuild\bundler.js:2009:7)
    at ServerTarget.write (C:\tools\isobuild\bundler.js:2278:13)
    at C:\tools\isobuild\bundler.js:2373:30
    at C:\tools\isobuild\bundler.js:2506:9
    at Array.forEach (native)
    at C:\tools\isobuild\bundler.js:2496:26
    at C:\tools\isobuild\bundler.js:2436:7
    at C:\tools\isobuild\bundler.js:2837:22
    at C:\tools\utils\buildmessage.js:271:13
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at C:\tools\utils\buildmessage.js:264:29
    ...
    at Object.capture (C:\tools\utils\buildmessage.js:252:19)
    at bundle (C:\tools\isobuild\bundler.js:2670:31)
    at C:\tools\isobuild\bundler.js:2617:32
    at Object.withCache (C:\tools\fs\files.js:1601:12)
[...]

PR #8773, opened 2017-06-09
Attempt to fix "ENOTEMPTY" by updating rimraf dependency.

abernix:
There have been a number of commits in [the history for `rimraf`](https://github.com/isaacs/rimraf/commits/master) which indicate other projects have also been struggling with ENOTEMPTY, specifically on Windows.

[This commit on `rimraf`](https://github.com/isaacs/rimraf/commit/d53235de863cfb00f691456288fcf3abbced7f6c), included in 2.6.1, takes a relatively aggressive approach but the absence of follow-up commits in a while leads me to believe that approach may have helped.

Of course, if Windows wants to cache a file handle after a process has stopped using it, that is its prerogative but the result is that tools (like Meteor) need to guess when it's safe to delete directories, which may still have cached file handles open on the OS, or try excessively (in the way this update to `rimraf` does) until its permitted in order to properly cleanup any mess which may have been made.

In triaging this issue (on a non-GitHub issue), I was able (on Windows) to reproduce only a single `ENOTEMPTY` error out of 14 attempts by continually running `./meteor --get-ready` (a development command which builds Meteor core packages) on fresh versions of a Meteor checkout.  This process triggers a lot of the same processes which take place building developer applications, so is a helpful way of trying to throw the error.  After applying this fix, I then ran the process an additional 30 times, and was no longer able to reproduce it.

I'm hoping this means the issue will be substantially lessened.

Attempts to help with meteor/meteor#8485.

(`dev-bundle-4.8.18` has been rebuilt and published for this PR)

### DIFF
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=4.8.17
+BUNDLE_VERSION=4.8.18
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -42,7 +42,7 @@ var packageJson = {
     "http-proxy": "1.11.1",
     "wordwrap": "0.0.2",
     "moment": "2.8.4",
-    "rimraf": "2.4.3",
+    "rimraf": "2.6.1",
     "glob": "7.0.6",
     // XXX: When we update this, see if it fixes this Github issue:
     // https://github.com/jgm/CommonMark/issues/276 . If it does, remove the


abernix:
Also, it's worth mentioning, the difference between running Meteor on Windows with a virus scanner (or other similar file scanners) enabled versus having it disabled (at least for Meteor-related files) is substantial.  I truly don't think the 44 executions would have finished today if I had Windows Defender enabled (which was enabled again when I booted up my latest Windows Insider Build).  

Scanning all those files is an expensive operation!

benjamn:
Possibly related? https://github.com/meteor/meteor/commit/0fbb97432c9f5c9b4c5958e719ca32d32a3922f1

abernix:
@benjamn https://github.com/meteor/meteor/commit/0fbb97432c9f5c9b4c5958e719ca32d32a3922f1 certainly fixed an actual bug, but yeah, I think you can thank the latent file descriptor caching in Windows for helping to identify it.  This fix in `rimraf` would have probably made it more difficult to find that particular error, though this should fix a number of perfectly normal situations where Windows clings onto the file descriptor past the file's active lifecycle (for a few ticks longer than we'd like).

brucejo75:
Hi @abernix,

> Also, it's worth mentioning, the difference between running Meteor on Windows with a virus scanner (or other similar file scanners) enabled versus having it disabled (at least for Meteor-related files) is substantial. I truly don't think the 44 executions would have finished today if I had Windows Defender enabled (which was enabled again when I booted up my latest Windows Insider Build).

> Scanning all those files is an expensive operation!

**Note**: you can enable [directory exclusions](https://support.microsoft.com/en-us/instantanswers/64495205-6ddb-4da1-8534-1aeaf64c0af8/add-an-exclusion-to-windows-defender) for Defender.

I think it would be worthwhile to do some windows build testing in an extreme environment to pick out build timing bugs:
- Full on Windows Defender
- Locate project directories (e.g. `.meteor`) and `\Users\XXX\appData` on a HDD partitions (5400 RPM for good measure)

I have been experiencing a few bugs in the build that I think I attribute to timing issues.  I have upgraded my project drive to be an SSD and now I am disabling Defender in my project directories.  I believe that having your projects on a HDD partition and running Defender exposes some timing bugs in the build process...

PR #8774, opened 2017-06-09
Remove check for standard-minifier-js@2.1.x from `bundle-visualizer`.

abernix:
While this was helpful in one of the incarnations of the `bundle-visualizer` during development it isn't relevant anymore since the `.stats.json` will not be present in the bundle unless it was created by a recent version of `standard-minifier-js` which passed in `stats` (no older versions did this).

Removing this check will also make it possible for other minifiers to be used with `bundle-visualizer` without them also needing to fork make their own version.

Thusly, this makes the `bundle-visualizer` work with Meteor 1.6, which comes with `standard-minifier-js@2.2.x`.

### DIFF
--- a/packages/non-core/bundle-visualizer/package.js
+++ b/packages/non-core/bundle-visualizer/package.js
@@ -1,5 +1,5 @@
 Package.describe({
-  version: '1.0.2',
+  version: '1.0.3',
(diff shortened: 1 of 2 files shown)


abernix:
Published as `bundle-visualizer@1.0.3`!

Issue #8775, opened 2017-06-09
Mongo Query Selectors are flagged as invalid.


brucejo75:
pr #8608 disallows all [query selectors](https://docs.mongodb.com/manual/reference/operator/query/).

Here is the problem: [modify.js#L251](https://github.com/meteor/meteor/blob/devel/packages/minimongo/modify.js#L251)

I ran a simple test:
```
// set the LocalData collection
LocalData =  new Mongo.Collection(null);

function setGlobalContext() {
  let gContext = {'globalContext': {}};
  gContext.globalContext.test = 'test';
  LocalData.upsert({'globalContext': {'$exists': true}}, gContext);
}

setGlobalContext();
```

the code throws because `$exists` has a `$` character.

I think tests need to be updated?

benjamn:
@hwillson Thoughts?

hwillson:
Ouch! Findings:

- Query selectors work properly with `update` (non-upsert), `find` / `findOne`
- Query selectors work with `upsert` when the record being upserted already exists
- Query selectors fail when being used with `upsert` (or an upsert `update`) and the record being upserted doesn't exist

So yes, definitely a bug and yes the tests are missing this 🙁.

I'll prepare a patch right now. Thanks for pointing this out @brucejo75!

sebakerckhof:
Okay, this should be fixed in next 1.6 build now that https://github.com/meteor/meteor/issues/8815 is merged.

hwillson:
Great, closing for now - thanks for your work on this @sebakerckhof!

PR #8776, opened 2017-06-09
Partial fix for #8775.

benjamn:
This may not be a complete fix for issue #8775, but I noticed while debugging the issue that `LocalCollection._removeDollarOperators` isn't removing nested `$`-operators (as it seems like it should be): https://github.com/meteor/meteor/blob/d6c4281eebe8fffcdeb415ccc361784f854126f4/packages/minimongo/minimongo.js#L790

### DIFF
--- a/packages/minimongo/minimongo_tests.js
+++ b/packages/minimongo/minimongo_tests.js
@@ -2857,6 +2857,13 @@ Tinytest.add("minimongo - modify", function (test) {
   upsertException({a: 0}, {$setOnInsert: {b: {'a.b':1}}});
   upsertException({a: 0}, {$setOnInsert: {b: {'\0a':1}}});
 
+  // Test for https://github.com/meteor/meteor/issues/8775.
+  upsert(
+    { a: { $exists: true }},
+    { $setOnInsert: { a: 123 }},
+    { a: 123 }
+  );
+
   exception({}, {$set: {_id: 'bad'}});
 
   // $bit
--- a/packages/minimongo/selector.js
+++ b/packages/minimongo/selector.js
@@ -1253,9 +1253,9 @@ LocalCollection._f = {
 
 // Oddball function used by upsert.
 LocalCollection._removeDollarOperators = function (selector) {
-  var selectorDoc = {};
-  for (var k in selector)
-    if (k.substr(0, 1) !== '$')
-      selectorDoc[k] = selector[k];
-  return selectorDoc;
+  return JSON.parse(JSON.stringify(selector, (key, value) => {
+    if (! key.startsWith("$")) {
+      return value;
+    }
+  }));
 };


mitar:
Are you sure it should be removing nested `$`?

hwillson:
@mitar For `upsert`'s we should be removing nested `$` operators. You can see the reason why [here](https://github.com/meteor/meteor/blob/c2c7fab2ed59fb3087b7c6d53c643d2c184246a6/packages/minimongo/minimongo.js#L800-L801) (which was a change that came from #8631). Thanks!

brucejo75:
Thanks for rapid debug.  Any tips on how I could patch this into my local version?  I am on windows.

#### I tried modifying files at:

> C:\Users\Bruce\AppData\Local\.meteor\packages\minimongo\1.2.1\web.browser\selector.js.

But any code changes to files there cause build errors.

#### I tried overriding `LocalCollection._removeDollarOperators `
but `LocalCollection` is not a global in the app.

#### I tried copying the package code into my packages directory and building it.

I named it in the `packages.js` file but the build could not find the package.  Which is weird because I have other local packages that work fine.

Basically, I have no idea how to patch the code without forking Meteor completely which seems like overkill.

hwillson:
@brucejo75 - while you're waiting for this to be merged / released, you can:

1. Create a `packages` directory in your app root (if you don't have one already).
2. Copy the `meteor/meteor/packages/minimongo` directory (and contents) from the meteor `fix-issue-8775` branch, into your app's `packages` directory.
3. `meteor add minimongo`

brucejo75:
Thanks @hwillson,

I have `METEOR_PACKAGE_DIRS` defined so I put the package in there.  That did not work.

It worked when I created another packages directory directly in my project directory.

So I am all set now.

Thanks again for the quick response!

sebakerckhof:
@benjamn I think this introduces a new bug because you loose EJSON types.
E.g. this would now error:
```
upsert({_id: new MongoID.ObjectID("1234567890")}, {$setOnInsert: {a: 123}})
````

hwillson:
@sebakerckhof - thanks for reporting this. Would you mind opening that up in a new issue? The changes in this PR didn't introduce that issue, as I'm able to re-create it with 1.5.

sebakerckhof:
@hwilson, weird. Because if I revert this commit I don't get this error.

hwillson:
Thanks @sebakerckhof - when I verified this, I tested the issue against `devel` first, then 1.5. Looks like something was still lingering from `devel` when I tested against 1.5, as I can still re-create the issue with 1.5 in my test environment. That being said, I just took a closer look at the code and see what you're saying - makes sense to me. We'll still need to track this in a new issue though since this PR has been merged. I'll move this into a new issue and work on a fix. Thanks again!

PR #8777, opened 2017-06-09
Update CoffeeScript to 1.12.6

GeoffreyBooth:
Bumps the CoffeeScript version, and uses the new primary repo, `coffeescript`.

### DIFF
--- a/packages/non-core/coffeescript/package.js
+++ b/packages/non-core/coffeescript/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Javascript dialect with fewer braces and semicolons",
-  version: "1.12.3_1"
+  version: "1.12.6_1"
 });
 
 Package.registerBuildPlugin({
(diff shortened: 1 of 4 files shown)


GeoffreyBooth:
@hwillson Thanks. When will the new package be released?

packages/non-core/coffeescript/package.js
```diff
@@ -1,14 +1,14 @@
 Package.describe({
   summary: "Javascript dialect with fewer braces and semicolons",
-  version: "1.12.3_1"
+  version: "1.12.6_1"
```
laosb: Shall we remove the `_1` in the version number, since we're releasing for 1.12.6 for the first time?

hwillson:
I've added a `History.md` entry and adjusted the package version. As soon as the tests finish running, I'll merge. Thanks!

hwillson:
Well, Travis appears to be hung. The build passed but Travis hasn't relayed those details to GH. I'm looking into it ...

packages/non-core/coffeescript/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Javascript dialect with fewer braces and semicolons",
-  version: "1.12.6_1"
+  version: "1.12.6"
```
mitar: No! Don't do that! Please wrap all upstream packages always with `_1` and then bump it if needed. It is really confusing if there is `1.12.6` and later on `1.12.6_1`. Which one is newer? Hard to know from just looking at the version.

abernix:
Published as `coffeescript@1.12.6_1`!  Thanks!

Issue #6402, opened 2016-03-04
Client side bundling doesn't eliminate dead code in production


tmeasday:
React has this pattern of writing:

``` js
if ("production" !== process.env.NODE_ENV) {
  // log warnings
}
```

Typically this code is eliminated by uglify when you bundle React. For whatever reason (the `process.env` symbol coming in some unusual way?) this doesn't happen when I run an app in `--production`. 

I'm not sure if this is a big deal, I just wanted to make a note of it. For more discussion see https://github.com/meteor/react-packages/pull/171

zol:
This optimization would be nice to revisit post 1.3

jbaxleyiii:
@zol @tmeasday is there any update on this? It has significantly increased our bundle size and is preventing us from moving to 1.3

zol:
@tmeasday - Do you know more about the scope of this issue? Are you sure it's uglify rather than webpack that does this dead code elimination?

tmeasday:
Not really. I just recall there's was a test on the `react-runtime` branch to check this worked properly.

benjamn:
This sounds like a case where you'd want to use something other than `standard-minifier-js`. Assuming `process.env.NODE_ENV` is a constant is not a sound decision for a generic minifier to make.

jbaxleyiii:
@benjamn just curious, why isn't that a sound decision?

tmeasday:
@benjamn I wonder how it worked once upon a time (Meteor 1.2)

jbaxleyiii:
@benjamn @tmeasday when does meteor load the node_modules needed for the app actually get run through the minifier plugin? I created a minifier but don't see where the bundled modules are passed to the minifer.

jbaxleyiii:
In case anyone is interested in a minifier that strips out dead code based on the environment: https://github.com/NewSpring/micro-minifer-js

Sadly, this hasn't fixed the bundle increase we have seen when moving off of the `webpack:webpack` package. 👎 I'm going to keep investigating to see why webpack's bundle size is ~1/2 the size of stock meteor. (From ~`300kb` using webpack to ~`600kb` using just 1.3)

jbaxleyiii:
@tmeasday with support for custom babel transforms this works great:

`babel-plugin-transform-dead-code-elimination` and `babel-plugin-transform-inline-environment-variables`

tmeasday:
@jbaxleyiii nice!

hwillson:
[...]

PR #8786, opened 2017-06-12
Inline process.env.NODE_ENV when minifying.

benjamn:
Despite [my previous concerns](https://github.com/meteor/meteor/issues/6402#issuecomment-221042378), having `minifier-js` replace `process.env.NODE_ENV` with its string value seems to be the most direct way to guarantee production versions of React modules do not access `process.env.NODE_ENV` at runtime, and that any resulting dead code is removed.

Relevant [background](https://forums.meteor.com/t/new-react-devtools-check-if-your-production-build-is-really-in-production-mode/36199).

Fixes meteor/meteor#6402.
Fixes meteor/meteor-feature-requests#94.

### DIFF
--- a/packages/minifier-js/package.js
+++ b/packages/minifier-js/package.js
@@ -1,10 +1,10 @@
 Package.describe({
   summary: "JavaScript minifier",
-  version: "2.1.0"
+  version: "2.1.1"
 });
 
 Npm.depends({
-  "uglify-js": "3.0.13"
+  "uglify-js": "3.0.15"
 });
 
 Package.onUse(function (api) {
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -14,7 +14,7 @@ var packageJson = {
     npm: "4.6.1",
     "node-gyp": "3.6.0",
     "node-pre-gyp": "0.6.34",
-    "meteor-babel": "0.21.4",
+    "meteor-babel": "0.21.5",
     reify: "0.11.22",
     "meteor-promise": "0.8.4",
     fibers: "1.0.15",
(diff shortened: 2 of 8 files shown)


packages/minifier-js/package.js
```diff
@@ -1,10 +1,10 @@
 Package.describe({
   summary: "JavaScript minifier",
-  version: "2.1.0"
+  version: "2.1.1"
```
benjamn: Only bumping the patch version of the `minifier-js` will make this functionality available to Meteor 1.5 apps (so you don't have to wait for 1.5.1). I think that's a good idea, since we probably would have shipped this functionality with 1.5 if it had been ready.

scripts/dev-bundle-tool-package.js
```diff
@@ -14,7 +14,7 @@ var packageJson = {
     npm: "4.6.1",
     "node-gyp": "3.6.0",
     "node-pre-gyp": "0.6.34",
-    "meteor-babel": "0.21.4",
+    "meteor-babel": "0.21.5",
```
benjamn: This change technically requires a new dev bundle, but I think we can wait on that because we don't minify the command-line tool implementation.

benjamn:
I'm going to go ahead and merge this into the `release-1.5.1` branch (rather than `devel`), since I need to make some other conflicting changes there.

hwillson:
Sounds great @benjamn! I was actually just looking at things, and everything looks great - thanks!

abernix:
LGTM! 💚 ✅

SachaG:
That's great! Looking forward to not seeing that pesky red React devtools icon anymore… And also saving a couple more k's.

Issue #8794, opened 2017-06-13
Mongo upserts with EJSON based types are no longer working


hwillson:
Reproduction: https://github.com/hwillson/broken-ejson-upsert

The changes made in https://github.com/meteor/meteor/pull/8776 impact using EJSON based types with Mongo upserts. Attempting something like:

```js
const testCollection = new Mongo.Collection(null);
testCollection.upsert({
  _id: new Mongo.ObjectID()
}, {
  $setOnInsert: {
    a: 123
  }
});
```

will throw a `Meteor does not currently support objects other than ObjectID as ids` error, even though our `_id` is an `ObjectID`. 

cc @sebakerckhof

holmrenser:
I am currently facing this problem, and I think this is what is keeping me from updating to 1.5.

PR #8796, opened 2017-06-13
Adjust mongo upserts so MongoID.ObjectID's aren't filtered from queries.

hwillson:
Hi all - this PR is intended to fix https://github.com/meteor/meteor/issues/8794. The `LocalCollection._removeDollarOperators` changes introduced in PR https://github.com/meteor/meteor/pull/8776 have an unintended side effect of not allowing `MongoID.ObjectID`'s through intact. The `JSON.stringify` / `JSON.parse` combo isn't restoring the full `MongoID.ObjectID` object (since it's not JSON compatible).

This PR adjusts things a bit so we're working with the EJSON representation of `MongoID.ObjectID` (and other objects), when running through `LocalCollection._removeDollarOperators`.

### DIFF
--- a/packages/minimongo/selector.js
+++ b/packages/minimongo/selector.js
@@ -441,7 +441,7 @@ var VALUE_OPERATORS = {
 
     // There are two kinds of geodata in MongoDB: legacy coordinate pairs and
     // GeoJSON. They use different distance metrics, too. GeoJSON queries are
-    // marked with a $geometry property, though legacy coordinates can be 
+    // marked with a $geometry property, though legacy coordinates can be
     // matched using $geometry.
 
     var maxDistance, point, distance;
@@ -1251,11 +1251,19 @@ LocalCollection._f = {
   }
 };
 
-// Oddball function used by upsert.
+// Oddball function used by upsert. Make sure that when we're removing
+// $ operators we don't accidently remove the "oid" (MongoID.ObjectID)
+// EJSON type (which includes $type and $value properties).
 LocalCollection._removeDollarOperators = function (selector) {
-  return JSON.parse(JSON.stringify(selector, (key, value) => {
-    if (! key.startsWith("$")) {
-      return value;
-    }
-  }));
+  return EJSON.parse(
+    JSON.stringify(
+      EJSON.toJSONValue(selector),
+      function (key, value) {
+        if (EJSON.fromJSONValue(this) instanceof MongoID.ObjectID
+            || !key.startsWith("$")) {
+          return value;
+        }
+      }
+    )
+  );
 };
--- a/packages/minimongo/selector.js
+++ b/packages/minimongo/selector.js
@@ -1251,19 +1251,22 @@ LocalCollection._f = {
   }
 };
 
-// Oddball function used by upsert. Make sure that when we're removing
-// $ operators we don't accidently remove the "oid" (MongoID.ObjectID)
-// EJSON type (which includes $type and $value properties).
-LocalCollection._removeDollarOperators = function (selector) {
-  return EJSON.parse(
-    JSON.stringify(
-      EJSON.toJSONValue(selector),
-      function (key, value) {
-        if (EJSON.fromJSONValue(this) instanceof MongoID.ObjectID
-            || !key.startsWith("$")) {
-          return value;
-        }
+// Oddball function used by upsert.
+LocalCollection._removeDollarOperators = (selector) => {
+  const cleansed = {};
+
+  Object.keys(selector).forEach((key) => {
+    const value = selector[key];
+    if (key.charAt(0) !== '$') {
+      if (value !== null
+          && value.constructor
+          && value.constructor.prototype === Object.prototype) {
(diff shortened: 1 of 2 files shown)


hwillson:
@benjamn @abernix I've added this to the 1.5.1 milestone, as this fix should be included alongside  https://github.com/meteor/meteor/pull/8776/commits/8aa3815c14d9a5f06e14362a443e350829c670cf. Let me know if you want anything changed - thanks!

sebakerckhof:
I think this still loses types like Date objects and (although more of an edge case)

hwillson:
Thanks @sebakerckhof, I noticed that as well - I'm working on tweaks to handle this.

hwillson:
I've adjusted the `LocalCollection._removeDollarOperators` function a bit. We don't seem to have a reliable way to differentiate between Mongo $ operators and custom EJSON representations, when trying to use a replacer function with `JSON.stringify`. For example, `EJSON.stringify({ testDate: new Date() })` produces `"{"testDate":{"$date":1497380438881}}"`. Unless we keep a whitelist of all allowed Mongo $ operators, trying to filter using a `JSON.stringify` replacer function is going to be difficult, since we'll be stripping valid EJSON types.

The new changes add a bit of complexity (and potentially a small performance hit), but attempt to make sure we're really only stripping nested Mongo $ operators. These changes aren't perfect, but they should cover/address the problems we've discussed (and most `upsert` use). Thanks!

packages/minimongo/selector.js
```diff
@@ -1252,10 +1252,21 @@ LocalCollection._f = {
...
-      return value;
+LocalCollection._removeDollarOperators = (selector) => {
+  const cleansed = {};
+
+  Object.keys(selector).forEach((key) => {
+    const value = selector[key];
+    if (key.charAt(0) !== '$') {
+      if (value !== null
+          && value.constructor
+          && value.constructor.prototype === Object.prototype) {
```
benjamn: Let's use `Object.getPrototypeOf(value) === Object.prototype` instead?

hwillson:
Hi @sebakerckhof - would you have a chance to test/verify these latest changes?

sebakerckhof:
@hwillson I've been looking further into mongo upsert behavior (which is not very well documented) and I see the following problems:

An upsert like: `{"foo": { "$exists": true }}, { "$setOnInsert": { "bar":" baz" }}`
In mongo would result in an object: `{ "bar": "baz" }`. But with the proposed changes here you would get: `{ "bar": "baz", "foo": {} }`

You can't just filter out empty objects because this:
 `{"foo": { "$exists": true }}, { "$setOnInsert": { "bar":" baz" }}` should actuallly result in:  `{ "bar": "baz", "foo": {} }`

An upsert like: `{ "foo": { "$eq": "test" }},{ "$setOnInsert": { "bar": "baz" }}` should result in
 `{ "foo": "test",  "bar": "baz" }`

So actually filtering out all dollar operators is not even correct *sigh*

So all things considered, I think we need a different approach to create the `selectorModifier` of an upsert () but I have no idea on an elegant solution at the moment.

sebakerckhof:
Okay, I've looked into other JS mongo implementations (e.g. nedb), but none seem to do it correctly.
So I've searched the mongodb codebase and the relevant functions seem to be:

`applyUpdateOpsForInsert`
https://github.com/mongodb/mongo/blob/master/src/mongo/db/exec/update.h#L137
https://github.com/mongodb/mongo/blob/master/src/mongo/db/exec/update.cpp#L735

and `populateDocumentWithQueryFields` (used by the previous):
https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver.h#L73
https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver.cpp#L176

And relevant tests:
https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver_test.cpp

I don't really have the time to port this myself right now, but it should provide good pointers for someone who feels like tackling this.

hwillson:
Ouch - thanks for the very detailed findings @sebakerckhof! I'll digest this a bit, but I think we should still get the immediate (reported issue) fixed as best as we can (since as it stands right now, breaking EJSON types will probably cause people more grief than some of the other edge cases). It looks like the fact that not all `$` operators should be filtered has been a long standing issue, that we should definitely fix, but I think I'll move that into a separate issue. Thanks again!

sebakerckhof:
I agree since a full solution might take some time. I've created a tracking issue:
https://github.com/meteor/meteor/issues/8806

hwillson:
I've submitted new changes to handle the first two scenarios @sebakerckhof mentioned in https://github.com/meteor/meteor/pull/8796#issuecomment-308665730 (the others will be handled in https://github.com/meteor/meteor/issues/8806). These new changes will check to see if a value in the selector is an object that only has `$` based keys. If it is, the full value object will be stripped, along with its parent key. I've also tried to clarify what `LocalCollection._removeDollarOperators` is doing in the function comment, as the previous `Oddball function used by upsert.` comment (while correct) wasn't very helpful.

Issue #8806, opened 2017-06-15
Make upsert behavior fully mongo-compliant


sebakerckhof:
Original discussion: https://github.com/meteor/meteor/pull/8796

Currently the upsert behavior is not mongo-compliant when using complex selectors.

Some examples:
Upsert: `{"foo": { "$exists": true }}, { "$setOnInsert": { "bar":" baz" }}`
In mongo would result in an object: `{ "bar": "baz" }`. But with Meteor you would get: `{ "bar": "baz", "foo": {} }`
Upsert: `{ "foo": { "$eq": "test" }},{ "$setOnInsert": { "bar": "baz" }}` should result in
 `{ "foo": "test",  "bar": "baz" }`, but in Meteor you'd get: `{ "bar": "baz", "foo": {} }`
Upsert  `{  "$and": [ { "foo": 123 }] },{ "$setOnInsert": { "bar": "baz" }}` should result in
 `{ "foo": 123,  "bar": "baz" }`, but in Meteor you'd get  `{ "bar": "baz" }`

A different approach is needed to create the `selectorModifier` here:
https://github.com/hwillson/meteor/blob/devel/packages/minimongo/minimongo.js#L786-L790

I've looked into other JS mongo implementations (e.g. `nedb`), but none seem to do it correctly.
So I've searched the mongodb codebase and the relevant functions seem to be:

`applyUpdateOpsForInsert`
https://github.com/mongodb/mongo/blob/master/src/mongo/db/exec/update.h#L137
https://github.com/mongodb/mongo/blob/master/src/mongo/db/exec/update.cpp#L735

and `populateDocumentWithQueryFields` (used by the previous):
https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver.h#L73
https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver.cpp#L176

And relevant tests:
https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver_test.cpp

sebakerckhof:
FYI: I'm gonna try to create a PR for this if I find some time over the weekend, so if somebody already started working on this, now is a good time to speak up :).

tellkeygit:
In an attempt to help students learn cs, we are looking for issues that they can potentially work on.Help us by mentioning our bot "@tellkeygit {#hours}" (to solve the issue). This would help students pick up issues based on their experience. Please take a look at https://www.keygit.info for more details.

[...]

Issue #5611, opened 2015-11-05
LocalCollection.update doesn't remove dollar operators properly on upsert


jedwards1211:
Consider the following valid upsert that works in the Mongo shell:

```
meteor:PRIMARY> db.blah.update({name: {$exists: false}}, {$setOnInsert: {hello: 'world'}}, {upsert: true})
WriteResult({
    "nMatched" : 0,
    "nUpserted" : 1,
    "nModified" : 0,
    "_id" : ObjectId("563a9bba6f3a3efd390d7d20")
})
```

Mongo.Collection.upsert fails to do this correctly because it doesn't correctly remove the `$exists` operator:

```
> var blah = new Mongo.Collection('blah');
> blah.upsert({name: {$exists: false}}, {$setOnInsert: {hello: 'world'}});

MongoError: The dollar ($) prefixed field '$exists' in 'name.$exists' is not valid for storage.
    at Object.Future.wait (/home/andy/.meteor/packages/meteor-tool/.1.0.45.g1vjm8++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:398:15)
    at [object Object].<anonymous> (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/meteor/helpers.js:119:1)
    at [object Object].MongoConnection.(anonymous function) [as update] (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/mongo/mongo_driver.js:678:1)
    at Object.fields (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/matb33:collection-hooks/update.js:75:1)
    at Object.collection.(anonymous function) [as update] (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/matb33:collection-hooks/collection-hooks.js:103:1)
    at [object Object].Mongo.Collection.(anonymous function) (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/mongo/collection.js:575:1)
    at [object Object].Mongo.Collection.(anonymous function) [as update] (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/aldeed:collection2/collection2.js:187:1)
    at [object Object].Mongo.Collection.upsert (/home/andy/jcore-portal/meteor/.meteor/local/build/programs/server/packages/mongo/collection.js:609:1)
    at repl:1:-56
[...]

Issue #5294, opened 2015-09-26
mongo - $push not working with $ne selector


mcbain:
On the server side minimongo will be used to validate Mongo.Collection.update arguments?!.

The following upsert fails with `MinimongoError: Cannot apply $push modifier to non-array`.
Minimongo IMHO infers `abc` to be a String, but $ne is meant as array-contains selector in that case.

``` javascript
upsert({
    'abc': {'$ne': 444}
  },
  {
    '$push': {
      'abc': 123
    }
  })
```

tmeasday:
Hi @mcbain can you post a reproduction of this? I would be surprised if this was the case.

mcbain:
@tmeasday Made a repro meteor package with a tinytest to show the bug https://github.com/mcbain/meteor-mongo-bug1

Doing the upsert in a mongo shell works.

tmeasday:
Hi @mcbain, thanks!

So you can see why Minimongo code is used for a server upsert here:
https://github.com/meteor/meteor/blob/devel/packages/mongo/mongo_driver.js#L525-L536

Explanation of why here: https://github.com/meteor/meteor/issues/2278#issuecomment-64252706

tmeasday:
(I guess this at the end of the day is a limitation of Minimongo, that is triggered by server upserts)

mcbain:
@tmeasday thanks for investigation. Since current mongo version is 3.2, is there a deprecation path for the 2.4 branch?

tmeasday:
Good question. I think there is some work ongoing about getting us working properly with 3.0 but I'm not sure about deprecating our support for 2.4.

mitar:
I can reproduce this even in the recent Meteor version. We should probably check if MongoDB is 2.4 or earlier, and use simulation only then, otherwise pass upsert directly to MongoDB.

sebakerckhof:
Okay, this should be fixed in next 1.6 build now that https://github.com/meteor/meteor/issues/8815 is merged.

hwillson:
Great, closing for now - thanks for your work on this @sebakerckhof!

PR #8815, opened 2017-06-17
Make minimongo upsert compliant with mongo behavior

sebakerckhof:
Attempt to fix: https://github.com/meteor/meteor/issues/8806, https://github.com/meteor/meteor/issues/5611, https://github.com/meteor/meteor/issues/8794, https://github.com/meteor/meteor/issues/5294 and https://github.com/meteor/meteor/issues/8775

This turned out to be quite painful. Following https://github.com/mongodb/mongo/blob/master/src/mongo/db/update/update_driver.cpp#L176 takes you to a big journey of mongo's source code. So a direct port is not really viable. I tried to infer what I could and also discovered a lot of rules with trial & error. It's possible there are more things I've missed.

However, this PR works with all old tests and makes the upsert behavior compliant with a big range of more complex upserts for every rule I could infer.

**But there's a catch**: This is possibly a BC break, since the behavior for these more complex selectors is, albeit compliant with mongo, different then what Meteor would have entered in the database before. 
E.g. this upsert: `{$and: [{foo: "bar"}]}, {$set: {a: 123}}` would now correctly result in `{foo: 'bar', a: 123}`, while in old versions it would be just `{a: 123}`
To make matters worse, it is not just for minimongo, but because Meteor simulates upserts for Mongo 2.4 compatibility (see: https://github.com/meteor/meteor/blob/devel/packages/mongo/mongo_driver.js#L527) it's also for what ends up in the actual database in some situations.
Now wouldn't be a bad time to remove that simulated upsert stuff (if 2.4 support is no longer a requirement), otherwise behavior might break again in the future.

So this needs a decent review and some warnings in changelogs here and there.

### DIFF
--- a/packages/minimongo/upsert_document.js
+++ b/packages/minimongo/upsert_document.js
@@ -15,109 +15,111 @@
 // - you can not have '$'-prefixed keys more than one-level deep in an object
 
 // Fills a document with certain fields from an upsert selector
-populateDocumentWithQueryFields = function (query, document = {}) {
-
-    if (Object.getPrototypeOf(query) === Object.prototype) {
-        // handle implicit $and
-        for (let [key, value] of Object.entries(query)) {
-            // handle explicit $and
-            if (key === "$and") {
-                value.forEach(sq => populateDocumentWithQueryFields(sq, document))
-            }
-            // handle $or nodes with exactly 1 child
-            else if (key === "$or") {
-                if (value.length === 1) {
-                    populateDocumentWithQueryFields(value[0], document)
-                }
-            }
-            //Ignore other '$'-prefixed logical selectors
-            else if (key[0] !== "$") {
-                populateDocumentWithKeyValue(document, key, value);
-            }
+export default function populateDocumentWithQueryFields (query, document = {}) {
+  if (Object.getPrototypeOf(query) === Object.prototype) {
+    // handle implicit $and
+    Object.keys(query).forEach(function (key) {
+      const value = query[key];
+      if (key === '$and') {
+        // handle explicit $and
+        value.forEach(sq => populateDocumentWithQueryFields(sq, document));
+      } else if (key === '$or') {
+        // handle $or nodes with exactly 1 child
+        if (value.length === 1) {
+          populateDocumentWithQueryFields(value[0], document);
         }
-    } else {
-        // Handle meteor-specific shortcut for selecting _id
-        if (LocalCollection._selectorIsId(query)) {
-            insertIntoDocument(document, "_id", query);
-        }
+      } else if (key[0] !== '$') {
+        // Ignore other '$'-prefixed logical selectors
+        populateDocumentWithKeyValue(document, key, value);
+      }
+    })
+  } else {
+    // Handle meteor-specific shortcut for selecting _id
+    if (LocalCollection._selectorIsId(query)) {
+      insertIntoDocument(document, '_id', query);
     }
+  }
 
-    return document;
-
+  return document;
 }
 
 // Handles one key/value pair to put in the selector document
-function populateDocumentWithKeyValue(document, key, value) {
-    if (value && Object.getPrototypeOf(value) === Object.prototype) {
-        populateDocumentWithObject(document, key, value);
-    } else if (!(value instanceof RegExp)) {
-        insertIntoDocument(document, key, value);
-    }
+function populateDocumentWithKeyValue (document, key, value) {
+  if (value && Object.getPrototypeOf(value) === Object.prototype) {
+    populateDocumentWithObject(document, key, value);
+  } else if (!(value instanceof RegExp)) {
+    insertIntoDocument(document, key, value);
+  }
 }
 
+// Handles a key, value pair to put in the selector document 
+// if the value is an object
+function populateDocumentWithObject (document, key, value) {
+  const keys = Object.keys(value);
+  const unprefixedKeys = keys.filter(k => k[0] !== '$');
 
-// Handles a key, value to put in the selector document if the value is an object
-function populateDocumentWithObject(document, key, value) {
-    const keys = Object.keys(value);
-    const unprefixedKeys = keys.filter(k => k[0] !== "$");
-
-    if (unprefixedKeys.length > 0 || !keys.length) {
-        //Literal (possibly empty) object ( or empty object ) 
-        //Don't allow mixing '$'-prefixed with non-'$'-prefixed fields
-        if (keys.length !== unprefixedKeys.length) {
-            throw new Error(`unknown operator: ${unprefixedKeys[0]}`)
-        } else {
-            validateObject(value, key);
-            insertIntoDocument(document, key, value);
-        }
+  if (unprefixedKeys.length > 0 || !keys.length) {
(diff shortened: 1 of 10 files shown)


sebakerckhof:
Okay https://github.com/meteor/meteor/issues/2278#issuecomment-64252706 and https://github.com/meteor/meteor/issues/7758  made me realize this needs a little more work on the `mongo` (not `minimongo`) side of things. However, this would also be simplified if we can do away with the simulated upserts

sebakerckhof:
@hwillson I know tests are failing but I'm kinda waiting on a decision on the simulated upsert stuff to continue/finish this.

hwillson:
Thanks for all of your work on this @sebakerckhof! I've been thinking about the simulated upsert stuff a bit, and don't have an answer yet. We'll definitely have to tread lightly here because of BC, but your changes make a lot of sense. I'll review this more thoroughly tomorrow and post my thoughts. Thanks again!

hwillson:
@sebakerckhof - I've thought about this more; I'm okay with dropping simulated upsert's (and Mongo 2.4 support, since it reached [end of life on March 2016](https://www.mongodb.com/support-policy)). As you mentioned, getting rid of simulated upsert's will help prevent future problems in this area (and help clean up the codebase a bit). 

That being said, we should definitely get MDG's opinion on this (@benjamn @abernix  - quick synopsis; go to [`packages/mongo/mongo_driver.js#L527`](https://github.com/meteor/meteor/blob/1eb578927e983676f4fed1d7836c8adf8355be20/packages/mongo/mongo_driver.js#L527) and read that comment. Getting rid of that simulated upsert code and using `$setOnInsert` will reduce future maintenance headache. Only gotcha is we have to drop Mongo 2.4 support). Thanks!

We'll be discussing this PR in our issue review meeting tomorrow, so if you don't hear back today, we'll make sure you have an update tomorrow.

abernix:
I think this is reasonable to do.  Though, does anyone know what mLab and Compose.io offer for their most "classic" options? (Presumably all their new customers are being on-boarded with at least Mongo 3.4.)

sebakerckhof:
For mlab it's 2.6 ( http://docs.mlab.com/ops/ ). Couldn't find anything for compose.

hwillson:
Compose.io "classic" subscribers get [Mongo 2.6.9 as the default](https://help.compose.com/v2.0/docs/reference-database-versions) for new deploys, but anything Compose supported earlier is still supported, for existing deploys (mLab is probably similar). That being said, I think the changes in this PR are going to break backwards compatibility enough that we might want to tie them to a major release. If that's the case, then mentioning that Mongo 2.4 is no longer supported might be okay (since both Compose and mLab provide upgrade paths and 2.4 (and 2.6) is EOL).

hwillson:
@sebakerckhof - we just discussed this in our issue triage meeting, and have agreed that it's okay to drop Mongo 2.4 support (starting from whatever Meteor version this PR gets merged into). So 👍 to removing the simulated upsert code as part of this PR. Thanks!

sebakerckhof:
Okay, this should be ready for review.
Turned out a bit more difficult than I initially thought, which makes that I cannot fully drop simulated upserts where the mod is a replacement and where the _id has to be controlled from meteor but neither the query, nor the replacement contains an id.
E.g. `upsert({"foo": "bar"}, {"bar": 1})`

Still dropping 2.4 is important because of:
- Upsert behavior (both how the query and modifier is treated) is very different between 2.4 and 2.6+, so minimongo couldn't possibly be correct.
- This makes it possible to have the result in the database to always be exactly the same as when done via mongo shell, even with the simulated version. 
Old version was dependend on minimongo behavior which was no longer correct and might still not be 100% compliant, but the result in the database will be.
- Mongo node driver 2.x does not officially support mongo 2.4
- We need the simulated upserts only in less cases now.

I just hope it doesn't go: https://imgs.xkcd.com/comics/workflow.png

sebakerckhof:
@hwillson Any idea when this can be reviewed? It's just that I'd like to do act on possible change requests when it's still fresh in my head.

hwillson:
Sorry for the delay @sebakerckhof - was traveling the past few days. I'll get this reviewed tomorrow for sure!

hwillson:
@sebakerckhof Would you mind rebasing when you have a sec?

sebakerckhof:
@hwillson I just had a sec.

packages/minimongo/upsert_document.js
```diff
@@ -0,0 +1,123 @@
...
+// - you can not mix '$'-prefixed keys and non-'$'-prefixed keys
+// - you can only have dotted keys on a root-level
+// - you can not have '$'-prefixed keys more than one-level deep in an object
+
+// Fills a document with certain fields from an upsert selector
+populateDocumentWithQueryFields = function (query, document = {}) {
+
+    if (Object.getPrototypeOf(query) === Object.prototype) {
+        // handle implicit $and
+        for (let [key, value] of Object.entries(query)) {
```
hwillson: Using `let [key, value] ...` here works, but using `Object.keys(query).forEach(...` is preferred (this was mentioned in another [code review](https://github.com/meteor/meteor/pull/8643#pullrequestreview-36055174), which I discussed with @benjamn afterwards; turns out the `Object.keys(...).forEach(...` approach is more performant behind the scenes).

packages/minimongo/upsert_document.js
```diff
@@ -0,0 +1,123 @@
...
+// Actually inserts a key value into the selector document
+// However, this checks there is no ambiguity in setting the value for the given key, throws otherwise
+function insertIntoDocument(document, key, value) {
+
+    Object.keys(document).forEach(existingKey => {
+        if (
+            (existingKey.length > key.length && existingKey.indexOf(key) === 0)
+            || (key.length > existingKey.length && key.indexOf(existingKey) === 0)
+        ) {
+            throw new Error(`cannot infer query fields to set, both paths '${existingKey}' and '${key}' are matched`)
```
hwillson: Can you make sure longer lines are wrapped at 80 chars? Switching the indentation to 2 chars will help, but some lines (like this one) are still over.

sebakerckhof:
@hwillson Ok. Did all the requested changes (I think)

sebakerckhof:
Btw, about `let [key, value] of Object.entries(obj)` vs `Object.keys(obj).forEach(...)`: a quick benchmark in chrome 59 seems to show there's no real difference anymore with v8 5.9 (which has TF+I on by default). Since meteor 1.6 will likely be released node 8.2+ (which uses v8 5.9) this policy might need to be reconsidered. At least, I think `let [key, value] of Object.entries(obj)` is more readable if you need both the key and value. 
Of course, this won't be true on the client for people using older browsers, but there you don't know if they're using v8, so v8 tricks don't make much sense and in this case the chance is very small a user will be doing thousands of big upserts per second.

hwillson:
Your changes look great @sebakerckhof, thanks! And good to know about `let [key, value] of ...`. LGTM!

benjamn:
Merged into `devel` as 08f5ea4c2b75c003df6625ed5e9dadea02e4414d. Thanks so much for working on this @sebakerckhof!

Issue #5652, opened 2015-11-12
$elemMatch with RegExp is not taken into account in calculating mongo observe key


lirbank:
**_UPDATE: Workaround available at the bottom of this thread**_

Given a publish function that takes a param which is used to query the DB using $elemMatch, the subscriber is not being updated after the initial data has been sent, despite the subscription being dynamically reloaded. It works as expected as long as $elemMatch is not being used.

Reproduction (vanilla Meteor 1.2.1, removed autopublish and insecure packages):
https://github.com/lirbank/meteor-elemmatch-regex-test
1. Start the app (or reload the browser) - all (five) persons are published.
2. Keep an eye on the server console.
3. In the "Search by job title" input, type "CEO" (all well - server says five persons found with job title "CEO" and five persons available in minimongo).
4. Now add any character after CEO eg "CEOX" so that the search match no persons/titles (error - server says zero persons are found with job title "CEOX" but minimongo says five).

This only happens when using [$elemMatch with RegExp](https://github.com/lirbank/meteor-elemmatch-regex-test/blob/master/server/server.js#L21-L25), not when [using RegExp directly](https://github.com/lirbank/meteor-elemmatch-regex-test/blob/master/server/server.js#L17) to query a subdocument in an array.

lirbank:
Alternative test:

1. Start the app (or reload the browser) - all (five) persons are published.
2. Keep an eye on the server console.
3. In the "Search by job title" input, type "CoCEO" (all well - server says no persons found with job title "CoCEO" and no persons available in minimongo).
4. Now remove the first two character "Co" leaving only "CEO" in the input (error - on the server we see that there are now five matches but the client says there are zero).

lirbank:
Update:

1. I've tried to with a low level publish function instead of simply returning the cursor, but the problem remains.

2. A cheezy workaround is to stop the subscription in the autorun function before re-running it, it works but forces a full re-load of all the subscribed data so it's slow and flickers as the template has to be re-rendered, etc. (not ideal):

```javascript
Template.Persons.onCreated(function () {
  Session.setDefault('searchStr', '');

  let self = this;
  let handle;

  this.autorun(function () {
    if (handle) handle.stop();
    handle = self.subscribe('persons', Session.get('searchStr'));
  });
});
```

tmeasday:
Thanks for the repro @lirbank! That's very helpful.

[...]

PR #8821, opened 2017-06-19
Handle regular expressions in EJSON

sebakerckhof:
Fixes: https://github.com/meteor/meteor/issues/5652

Background (kudos to: https://github.com/meteor/meteor/issues/5652#issuecomment-157544912 for pointing in the right direction):
Meteor multiplexes/re-uses duplicate observers. For fast look-up of duplicate observers it creates a hash with some properties of the cursor (e.g. selector, projection, ...).
However, for this hash it simply uses `JSON.stringify` ( https://github.com/meteor/meteor/blob/a4de554a95455e91b5d7e71c4c611c2e78725d5e/packages/mongo/mongo_driver.js#L1221 ) , which can not stringify regular expressions. This leads to a problem where different queries would re-use the same (wrong) observer.

However, this selector has already passed the `Mongo.Collection._rewriteSelector` (https://github.com/meteor/meteor/blob/906aab4447ace6cedf873abe0f3c99ad5539cadc/packages/mongo/collection.js#L369) method, which should replace regular expression objects with the alternative `{ $regex: '', $options: '' }` syntax. The problem is that `Mongo.Collection._rewriteSelector` doesn't do its work thoroughly and misses regular expressions inside `$elemMatch` amongst others.

Btw, since order of JSON doesn't guarantee the order of keys in an object, this is probably not the best way to 'hash' the selector, but that's another issue.

Approaches considered:

1. Fix `_rewriteSelector`. This would mean replace every occurence of a regular expression, even deeply nested once in `$elemMatch` and `$in` (amongst others?)

This poses 2 problems:
- The `$in`-array does not support  `{ $regex: '', $options: '' }` syntax
- `_rewriteSelector` is called multiple times for each query, and `$in` can sometimes contain long lists (at least in my apps) which would be a potential serious performance hit.

2. Just use a replacer function in the `JSON.stringify` method here: https://github.com/meteor/meteor/blob/a4de554a95455e91b5d7e71c4c611c2e78725d5e/packages/mongo/mongo_driver.js#L1221
This could check if a value is a RegExp instance and replace it with something serializable

This is the least obtrusive way (like 1 line of code) to fix the specific problem mentioned in: https://github.com/meteor/meteor/issues/5652
However, that still leaves a buggy `_rewriteSelector`. And my mom always thought me I shouldn't do half-assed work, which is why I went for option 3

3. Add support to EJSON for RegExp stringify/parsing. This solves the problem with both `_rewriteSelector` and with `observeKey`-hash.

This is IMO the most elegant solution, but it's also the most obtrusive. From what I can make out from the comments in the code the only reason `_rewriteSelector` replaced regular expressions is to make them DDP-serializable (although it's not clear why it should be DDP-serializable), which this approach achieves. 
But there might be something crucial I'm missing here. Please review

### DIFF
--- a/packages/ejson/ejson.js
+++ b/packages/ejson/ejson.js
@@ -94,6 +94,21 @@ var builtinConverters = [
       return new Date(obj.$date);
     }
   },
+  { // RegExp
+    matchJSONValue: function (obj) {
+      return _.has(obj, '$regexp') && _.has(obj, '$flags') && _.size(obj) === 2;
+    },
+    matchObject: function (obj) {
+      return obj instanceof RegExp;
+    },
+    toJSONValue: function (regexp) {
+      return { $regexp: regexp.source, $flags: regexp.flags };
+    },
+    fromJSONValue: function (obj) {
+      //replaces duplicate / invalid flags
+      return new RegExp(obj.$regexp, obj.$flags.replace(/[^gimuy]/g,'').replace(/(.)(?=.*\1)/g, ''));
+    }
+  },
   { // NaN, Inf, -Inf. (These are the only objects with typeof !== 'object'
     // which we match.)
     matchJSONValue: function (obj) {
--- a/packages/ejson/ejson_test.js
+++ b/packages/ejson/ejson_test.js
@@ -185,6 +185,16 @@ Tinytest.add("ejson - parse", function (test) {
   );
 });
 
+Tinytest.add("ejson - regexp", function (test) {
+  test.equal(EJSON.stringify(/foo/gi), "{\"$regexp\":\"foo\",\"$flags\":\"gi\"}");
+  var d = new RegExp("foo", "gi");
+  var obj = { $regexp: "foo", $flags: "gi" };
+
+  var eObj = EJSON.toJSONValue(obj);
+  var roundTrip = EJSON.fromJSONValue(eObj);
+  test.equal(obj, roundTrip);
+});
+
 Tinytest.add("ejson - custom types", function (test) {
   var testSameConstructors = function (obj, compareWith) {
     test.equal(obj.constructor, compareWith.constructor);
--- a/packages/minimongo/selector.js
+++ b/packages/minimongo/selector.js
@@ -775,8 +775,7 @@ ELEMENT_OPERATORS = {
       var regexp;
       if (valueSelector.$options !== undefined) {
         // Options passed in $options (even the empty string) always overrides
-        // options in the RegExp object itself. (See also
-        // Mongo.Collection._rewriteSelector.)
+        // options in the RegExp object itself. 
 
         // Be clear that we only support the JS-supported options, not extended
         // ones (eg, Mongo supports x and s). Ideally we would implement x and s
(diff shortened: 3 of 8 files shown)


hwillson:
Thanks for taking a look at this @sebakerckhof! I agree, option 3 sounds like the best path forward. I've looked over your changes, and so far everything looks great. Could you look into why the tests are failing though? It looks like they're getting hung up in the [async upsert tests ](https://github.com/meteor/meteor/blob/1eb578927e983676f4fed1d7836c8adf8355be20/packages/mongo/mongo_livedata_tests.js#L1846) for some reason. Once the tests are passing I'll do a more thorough review/testing.

hwillson:
@sebakerckhof - why did you close this? These changes looked promising - did you uncover potential problems?

sebakerckhof:
Oops, I must have accidentally clicked the close button.

sebakerckhof:
@benjamn These tests are failing here because not all browsers (and the test environment) support ` RegExp.prototype.flags`.
However, even if I add the `ecmascript` package as a dependency of the `ejson` package, the test keeps failing in those environments. Shouldn't that have added a polyfill for it?

abernix:
Currently, `core-js/es6/regexp` is not included in [`ecmascript-client-runtime`](https://github.com/meteor/meteor/blob/6bde360b9c075f1c78c3850eadbdfa7fe271f396/packages/ecmascript-runtime-client/runtime.js#L34).  Not sure if this was an intentional omission by @benjamn during [the reorganization](https://github.com/meteor/meteor/commit/02cbf94834bac2e01d8e8f2c3bf73a407cc00bcb) but it went from being included to being not included.

sebakerckhof:
@benjamn can you shed your light on this?

benjamn:
@sebakerckhof If you rebase against `devel`, you should get a version `ecmascript-runtime-client` that includes the `core-js/es6/regexp` polyfills.

sebakerckhof:
@benjamn thanks, seems to work now.

@hwillson tests are passing. I had to close & re-open again because travis had a segmentation fault first time. Circle CI has 1 failure (also segmentation fault) but that should be unrelated.

hwillson:
@sebakerckhof Just a quick FYI; I've added a few extra commits here to help fix the recent CircleCI issues we've been running into. I needed to use one of the open PR's to help troubleshoot, and we we're planning on merging this PR as soon as the tests passed, so it was a good fit for the additional changes / testing. Tests are passing now, so we're all set here. Thanks!

packages/ejson/ejson.js
```diff
@@ -94,6 +94,21 @@ var builtinConverters = [
...
+    },
+    matchObject: function (obj) {
+      return obj instanceof RegExp;
+    },
+    toJSONValue: function (regexp) {
+      return { $regexp: regexp.source, $flags: regexp.flags };
+    },
+    fromJSONValue: function (obj) {
+      //replaces duplicate / invalid flags
+      return new RegExp(obj.$regexp, obj.$flags.replace(/[^gimuy]/g,'').replace(/(.)(?=.*\1)/g, ''));
```
mitar: So global flag gets removed?

sebakerckhof: No, mind the '^', which is a negation.

mitar: Ah, true. :-)

Issue #5175, opened 2015-09-15
EJSON.clone can't clone objects with a prototype and a length property


joegoldbeck:
If I return an object from a Meteor method as follows: `return {a: 2, length: 37}`, it is sent over the websocket correctly as `{a: 2, length: 37}`.

However, if that object is an instance of another class like so `return new MyClass({a: 2, length: 37})`, then what is sent is an empty object (`{}`).

If the key "length" is not present, it also works correctly. `return new MyClass({a: 2})` sends `{a: 2}` over the websocket as expected.

stubailo:
OK, I've tracked this down to the fact that we use `EJSON.clone` to clone any return values on methods. EJSON.clone thinks that anything with a `length` property is an array.

hwillson:
This has been a re-occurring problem - see https://github.com/meteor/meteor/issues/8329, https://github.com/meteor/meteor/issues/594, and several others. A quick synopsis of the problem (and solution) can be found [here](https://github.com/meteor/meteor/issues/8329#issuecomment-280217394). This problem is manifesting itself in the [`_.each` within the EJSON packages `EJSON._adjustTypesToJSONValue` function](https://github.com/meteor/meteor/blob/06d00e45c7d375ef4787411c1698210ea9fc438d/packages/ejson/ejson.js#L214-L227).

hwillson:
Just a quick update here - I started working on a PR for this, only to find out that almost every part of the EJSON public API falls apart when working with an object that has a property named `length`  (thanks to `underscore`). So this issue isn't limited to `EJSON.clone` - here's another quick example that will fail:

```js
var Widget = function () {
  this.length = 10;
};
var widget = new Widget();
var widgetJson = EJSON.toJSONValue(widget);
```

`widgetJson` will hold `{ 0: undefined, 1: undefined, ... 9: undefined }` instead of `{ length: 10 }`. I'll aim to address all `length` problem areas - more details shortly.

hwillson:
The `ejson` package no longer uses underscore, so this is no longer an issue. Closing - thanks!

PR #8824, opened 2017-06-21
[Work in progress] Adjusted EJSON public API to handle objects with properties names "length"

hwillson:
Hi all - this PR is intended to fix issue #5175. Several parts of the `EJSON` API won't  work properly when manipulating objects with properties named `length`. This is the same problem that was identified in issue https://github.com/meteor/meteor/issues/8329, where `underscore` is improperly treating objects with a `length` property as an array (within several of its functions). This is a long-standing issue with `underscore` (that won't be changed as they don't see it as a bug), so for now I'm removing the `_.each` and `_.size` calls that create this issue (not all of them - just the ones that directly impact the `EJSON` public API and cause this issue). Quick side note - `lodash` does not have this problem, so this issue could have also been resolved by switching to `lodash`. Thanks!

### DIFF
--- a/packages/ejson/package.js
+++ b/packages/ejson/package.js
@@ -5,16 +5,12 @@ Package.describe({
 
 Package.onUse(function (api) {
   api.use(['ecmascript', 'underscore', 'base64']);
+  api.mainModule('ejson.js');
   api.export('EJSON');
-  api.export('EJSONTest', {testOnly: true});
-  api.addFiles('ejson.js', ['client', 'server']);
-  api.addFiles('stringify.js', ['client', 'server']);
 });
 
 Package.onTest(function (api) {
-  api.use('ejson', ['client', 'server']);
-  api.use(['tinytest', 'underscore']);
-
-  api.addFiles('custom_models_for_tests.js', ['client', 'server']);
-  api.addFiles('ejson_test.js', ['client', 'server']);
+  api.use('ejson');
+  api.use(['ecmascript', 'tinytest', 'underscore']);
+  api.mainModule('ejson_tests.js');
 });
--- a/packages/ejson/package.js
+++ b/packages/ejson/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Extended and Extensible JSON library",
-  version: '1.0.13'
+  version: '1.0.14'
 });
 
 Package.onUse(function (api) {
--- a/packages/ejson/package.js
+++ b/packages/ejson/package.js
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
-  api.use(['underscore', 'base64']);
+  api.use(['ecmascript', 'underscore', 'base64']);
   api.export('EJSON');
   api.export('EJSONTest', {testOnly: true});
   api.addFiles('ejson.js', ['client', 'server']);
--- a/packages/ejson/custom_models_for_tests.js
+++ b/packages/ejson/custom_models_for_tests.js
@@ -80,8 +80,10 @@ _.extend(Holder, {
 
 EJSON.addType("Holder", Holder.fromJSONValue);
 
-_.extend(EJSONTest, {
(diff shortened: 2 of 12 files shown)


hwillson:
Oops - I just noticed the package tests were passing locally, but failing via Travis, as I was missing `ecmascript`. Hopefully there are no objections to it being added as a dependency - definitely let me know if there are. Thanks!

packages/ejson/package.js
```diff
@@ -1,10 +1,10 @@
...
-  version: '1.0.13'
+  version: '1.0.14'
 });
 
 Package.onUse(function (api) {
-  api.use(['underscore', 'base64']);
+  api.use(['ecmascript', 'underscore', 'base64']);
   api.export('EJSON');
   api.export('EJSONTest', {testOnly: true});
   api.addFiles('ejson.js', ['client', 'server']);
```
benjamn: Let's switch to `api.mainModule` instead of `api.addFiles` if we're using `ecmascript`.

packages/ejson/package.js
```diff
@@ -1,10 +1,10 @@
 Package.describe({
   summary: "Extended and Extensible JSON library",
-  version: '1.0.13'
+  version: '1.0.14'
 });
 
 Package.onUse(function (api) {
-  api.use(['underscore', 'base64']);
+  api.use(['ecmascript', 'underscore', 'base64']);
```
benjamn: Can we remove `underscore` now? 😉

hwillson: That's a good question - the changes I put in place (replacing some `underscore` function calls) were just intended to address the issue in #5175. Underscore is still used fairly heavily in the rest of the `ejson` package - here are all of the unique `underscore` calls:

```
isNaN
has
isEmpty
any
all
isArray
isArguments
each
size
```

I'm happy to remove `underscore` completely, but I'd just be replicating a lot of the libraries code to address the above functionality (excluding the obvious ones like `each`). With the upcoming `underscore` to `lodash` switch (https://github.com/meteor/meteor-feature-requests/issues/48), completely removing `underscore` from `ejson` might not be needed (yet). Again, happy to switch it out if you say make it happen, so let me know. Thanks!

packages/ejson/package.js
hwillson: Done - I've also adjusted the package definition and structure a bit to better leverage `ecmascript`, and clean up the exports.

hwillson:
I would love to know why the circle build keeps failing at `"package-tests: add debugOnly and prodOnly packages"`. This test (and all other tests) pass every time locally. I'm re-running them again - normally I'd chalk this up to circle flakiness, but the failure seems pretty consistent. If it keeps happening, I'll continue to troubleshoot.

hwillson:
This is certainly puzzling - for some reason all tests pass locally, every time, but 2 specific tests fail via Circle consistently (and one of them fails spectacularly with a core dump). I've (mostly) ruled out Circle flakiness - these failures appear to be valid, but I haven't been able to connect these failures to any of the changes I've made (and again, they pass locally). I've updated the issue label to reflect that this is a work in progress, and will continue to troubleshoot further in a separate branch (and Circle instance), to reduce noise here.

hwillson:
Well, after pulling a lot of hair out over the test failures here, I've traced the issue down to adding `ecmascript` to the `ejson` package. If I revert all of my changes and just add `ecmascript` as a `ejson` dependency, Circle will fail (with a seg fault / core dump). If I leave `ecmascript` in place as a dependency, but disable the "add packages to app" test in `tools/tests/package-tests.js`, all Circle tests will pass, without seg faulting / core dumping. Definitely bizarre - something in the "add packages to app" self test isn't playing nicely with `ecmascript`. That being said, I've chewed through way too much time trying to troubleshoot this, when cleaning up the `ejson` package to use `ecmascript` really isn't necessary to address the original issue (#5175). 

Long story short, I'm going to close this PR and submit a new one that is more focused on addressing the original issue.

PR #8833, opened 2017-06-22
Support for import followed by other statements in scripts piped to `meteor shell`

### LINKED ISSUE #8823: Cannot use "import" in `meteor shell` via STDIN
While using `import` in a normal `meteor shell` works

```
$ meteor shell

> import { Success } from "./imports/collections";
> Success
'success!'
> Shell exiting...
```

The same command using STDIN throws a `SyntaxError`

```
$ echo 'import { Success } from "./imports/collections"' | meteor shell
SyntaxError: Unexpected token, expected ( (1:8)
```

Seen in both meteor `1.4.4.3` and `1.5`

atd:
Fixes #8823

### DIFF
--- a/packages/shell-server/shell-server.js
+++ b/packages/shell-server/shell-server.js
@@ -127,7 +127,7 @@ class Server {
       if (options.evaluateAndExit) {
         evalCommand.call(
           Object.create(null), // Dummy repl object without ._RecoverableError.
-          "(" + options.evaluateAndExit.command + ")",
+          options.evaluateAndExit.command,
           null, // evalCommand ignores the context parameter, anyway
           options.evaluateAndExit.filename || "<meteor shell>",
           function (error, result) {
--- a/packages/shell-server/shell-server.js
+++ b/packages/shell-server/shell-server.js
@@ -203,30 +203,7 @@ class Server {
       configurable: true
     });
 
-    if (Package.modules) {
-      // Use the same `require` function and `module` object visible to the
-      // application.
-      var toBeInstalled = {};
-      var shellModuleName = "meteor-shell-" +
-        Math.random().toString(36).slice(2) + ".js";
-
-      toBeInstalled[shellModuleName] = function (require, exports, module) {
-        repl.context.module = module;
-        repl.context.require = require;
-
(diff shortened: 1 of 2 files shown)


abernix:
Just noting that the test failure appears related to the change (and isn't just an unrelated fluke).

packages/shell-server/shell-server.js
```diff
@@ -441,7 +441,18 @@ function evalCommand(command, context, filename, callback) {
...
   evalCommandPromise.then(function () {
-    callback(null, script.runInThisContext());
+
+    if (repl.input) {
+      callback(null, script.runInThisContext());
+    } else {
+      // If repl didn't start, `require` and `module` are not visible
+      // in the vm context
+      repl.require = require;
+      repl.module = module;
```
benjamn: These are the `require` and `module` functions available globally in the `shell-server.js` module. Instead, we should be creating a module for the REPL with `meteorInstall` and using the resulting `require` and `module` parameters, as we do [here](https://github.com/atd/meteor/blob/e5f4d0a84a04232e50b74cb386a36c0408c1054a/packages/shell-server/shell-server.js#L207-L228).

benjamn: And of course now would be a great time to move the code that creates the REPL module into a helper function, since it's now used twice.

benjamn:
I canceled and rescheduled the most recent test run because it failed with a segmentation fault [here](https://circleci.com/gh/meteor/meteor/4319#tests/containers/1), which is probably not your fault.

atd:
I have extracted the function and ended up using the `global` object (it didn't work with `repl` and `global` is used in [node's doc](https://nodejs.org/api/vm.html#vm_script_runinthiscontext_options)).

benjamn:
The test failure seems entirely unrelated, though it has happened more than once now. I'm going to merge this anyway, in order to get it into the first 1.5.1 release candidate. If that test keeps failing, we can keep debugging it on the `release-1.5.1` branch.

benjamn:
Merged to `devel` as 2bcfb073c44bda3b517507a4c977860e0313ab3a and 0b1a1ef59f2403bb54be39bf3c249827aefe819b.

PR #8841, opened 2017-06-23
Server-side rendering revamp

benjamn:
To support server-side rendering of any kind, it would be ideal if there was a way for server code to inject rendered HTML into the boilerplate HTML generated by packages like `static-html`. Until now, that hasn't been possible, since the only mechanism for adding dynamic HTML to the initial HTTP response was to modify `request.dynamicHead` and/or `request.dynamicBody`.

With this new API, it should now be possible to register callbacks that (for example) parse `boilerplate.baseData.body`, find DOM nodes by `id`, render HTML fragments, inject the fragments into the DOM, and set `boilerplate.baseData.body` to the resulting HTML string.

I previously thought this would require changes to the `boilerplate-generator` package, and thus we had to wait for #8820 to be finished, but it turns out we can intercept `boilerplate.baseData` before calling `boilerplate.toHTML`, with only the changes in this PR.

I also took this opportunity to modernize the `webapp` package to use more ECMAScript 2015+ features, so it may be easier to review this PR commit-by-commit.

### DIFF
--- a/packages/webapp/package.js
+++ b/packages/webapp/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Serves a Meteor app over HTTP",
-  version: '1.3.16'
+  version: '1.3.17'
 });
 
 Npm.depends({connect: "2.30.2",
--- a/packages/webapp/package.js
+++ b/packages/webapp/package.js
@@ -30,15 +30,19 @@ Package.onUse(function (api) {
   // (browser-policy depends on webapp). So we don't explicitly depend in any
   // way on browser-policy here, but we use it when it is loaded, and it can be
   // loaded after webapp.
-  api.export(['WebApp', 'main', 'WebAppInternals'], 'server');
-  api.export(['WebApp'], 'client');
-  api.addFiles('webapp_server.js', 'server');
-  api.addFiles('webapp_client.js', 'client');
-  api.addFiles('webapp_cordova.js', 'web.cordova');
+  api.mainModule('webapp_server.js', 'server');
+  api.export('WebApp', 'server');
+  api.export('WebAppInternals', 'server');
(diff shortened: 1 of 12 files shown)


benjamn:
Should the callback be able to return a `Promise`? Is that feasible without using `Fiber`s to await the result synchronously?

zimme:
I guess `getBoilerplate()` would have to return a promise in that case also, so I guess we would have to use `Fiber`s or some promises to wait for all callbacks to resolve before sending it down the pipe?

packages/webapp/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Serves a Meteor app over HTTP",
-  version: '1.3.16'
+  version: '1.3.17'
 });
 
 Npm.depends({connect: "2.30.2",
```
hwillson: Not related to these changes, but we should really consider jumping to `connect` 3.x at some point.

benjamn:
@abernix @hwillson Should we go ahead and merge this into `release-1.5.1` and continue working/debugging there?

abernix:
@benjamn I think the failures are all related but not related to any of these actual PRs.  So, yes.

marcoschwartz:
Sounds like a great feature ! Does this apply to blaze or it's only for react ? Thx

msavin:
+1 - would love to see this working with Blaze, or even just Spacebars since we do not need the reactivity/etc.

rgnevashev:
this is a great feature... but:
1) How can I redirect inside the onPageLoad func? See react-router v4 SSR example
2) How can I obtain userId in my components on server side? My App depends on currentUser and I want obtain userId on server side and pass to my App entry point component...

Issue #8286, opened 2017-01-25
Subresource Integrity (SRI) support


juho:
[Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), basically hash-checking for CDN-delivered files is slowly becoming a thing with support in Chrome and Firefox. Haven't noticed anything about this on GitHub so I'm raising an issue here to discuss what needs to be done to support it in the future. 

There's one (1) [post](https://forums.meteor.com/t/lets-make-meteor-the-most-secure-framework-ever-subresource-integrity/9211/5) about it on the forums so far so I understand it's not a huge thing yet, but it'd be nice to have on the roadmap eventually. Unless I'm missing something, I imagine the hash could be calculated rather easily when you run `meteor build` or deploy, and then injected into the script tag.

abernix:
Meteor already does hashing internally for various purposes (cache-busting, etc.) purposes so this would probably fit in really well as I believe it's already stored in the `program.json` of the generated "star" bundle (check around bundler code [here](https://github.com/meteor/meteor/blob/devel/tools/isobuild/bundler.js#L1)).  Unfortunately, it doesn't currently use any of the SRI-supported hashes (SHA256, SHA384, SHA512) to do that (it uses SHA1).

Small speed consideration to consider there, but this might be a relatively easy PR if someone was willing to look at it.

```sh
$ openssl speed sha256 sha512
The 'numbers' are in 1000s of bytes per second processed.
type             16 bytes     64 bytes    256 bytes   1024 bytes   8192 bytes
sha1             49444.47k   136852.67k   300421.57k   417013.13k   479489.03k
sha256           34909.98k    83895.28k   142883.64k   180203.46k   198309.35k
sha512           24882.98k   103407.50k   181249.93k   263356.03k   307869.74k
```

laosb:
Had a look around the code - we'll need to replace `watch.sha1` with `watch.sha256`(not many in code, but many in comments) and I'm not sure if it's used somewhere else.

mitar:
In my fork I have already been doing something similar and I have extended in the past `program.json` to simply include also `sha256` keys as well.

macsj200:
@mitar can you include a link to your fork? I'd like to see how you implemented this.

mitar:
Sadly it has not been open sourced. :-(

I think something like that is enough:

```patch
--- bundler.js.orig	2017-05-09 01:00:08.000000000 -0700
[...]

PR #8847, opened 2017-06-27
SRI support

macsj200:
This is a feature request PR for Subresource Integrity (SRI), see https://github.com/meteor/meteor-feature-requests/issues/35


mitar:
So those are just adding it to the manifest. Now you also have to change the [boilerplate generator](https://github.com/meteor/meteor/blob/devel/packages/boilerplate-generator/boilerplate_web.browser.html) to include SRI attributes.

macsj200:
Do you recommend I compute these hashes dynamically somehow, or should I paste them in? Also, are there some scripts that should not have hashes computed, i.e. compiled script files that change their hash every run?

mitar:
Maybe do the same as with sha1 hash and use salt as well. (You can use same salt.)

You do not have to compute them dynamically. Just in the boilerplate code access your hashes and add them as a SRI attribute.

zimme:
If this PR is going to touch boilderplateGenerator I would have a look at #8841 first.

mitar:
You mean #8820?

mitar:
But I think we could continue for now and later on port it.

tools/isobuild/bundler.js
```diff
@@ -592,6 +593,15 @@ class File {
     this._contents = options.data || null; // contents, if known, as a Buffer
     this._hashOfContents = options.hash || null;
     this._hash = null;
```
abernix: Isn't there an option where we can just update the existing `sha1` hash to be a `sha256` hash instead of calculating two separate hashes?

mitar: That would require bumping the manifest file version, which we might want to do.

Maybe then we can also fix https://github.com/meteor/meteor-feature-requests/issues/54 at the same time. :-)

zimme:
> But I think we could continue for now and later on port it.

Yes, sounds like a plan.

macsj200:
Do you know which tests I should run to try this feature out? I found these tests
```Markdown
old:                                          
  - bundler-assets [checkout non-galaxy]      
  - bundler-options [checkout non-galaxy] 
```
But they're under "old".

mitar:
I do not think there are any tests for this? You should write them. :-)

hwillson:
Hi @macsj200 - thanks for kick starting this PR. Reading over the discussion here (and in https://github.com/meteor/meteor-feature-requests/issues/35, as well as https://github.com/meteor/meteor/issues/8286), I'm a bit confused regarding the discussed direction. Why are we interested in adding a SRI compliant hash to each resource listed in the `program.json`? SRI is currently only supported by browsers on `<link />` and `<script />` tags, which means in a production Meteor app the only files listed in the `program.json` that we would need a SRI hash for, are the main bundled/minified JS and CSS files (usually). That being said, why do we need SRI based hashes in the `program.json` at all? The hash has to be provided on the `<link />` and `<script />` tags themselves, which means the [`boilerplate-generator`](https://github.com/meteor/meteor/tree/devel/packages/boilerplate-generator) needs to add them in (which it could do this quite easily by calculating the necessary SRI hashes and injecting them at the same time it builds the web boilerplate). Given this, I'm not sure why this PR needs to touch anything in the isobuild process at all?

hwillson:
@macsj200, we haven't heard back regarding the questions asked in https://github.com/meteor/meteor/pull/8847#issuecomment-314112617. Are you still interested in working on this?

hwillson:
Hi @macsj200 - we haven't heard back on this in a while, so to help keep the pending PR list nice and tidy we'll close it for now. Adding SRI support is something we'll consider, but as discussed in https://github.com/meteor/meteor/pull/8847#issuecomment-314112617, there are better ways to accomplish this. If you (or anyone else) is interested in continuing this work in the future, post back here and we'll re-open this PR, or submit a new PR. Thanks for your work on this!

Issue #8782, opened 2017-06-11
Specific combination of package updates breaks IE11


ffxsam:
I apologize in advance for the lack of detailed information. I captured as much as I can though. And as usual, I can't provide a reproduction repo since it's proprietary code.

My app suddenly stopped working in IE11, completely. There was a console error about "No such route". I used `git bisect` to locate the bad commit, and found that it was solely a set of Meteor package updates that broke things. Here's the diff (from good commit to bad commit):

```diff
diff --git a/.meteor/.finished-upgraders b/.meteor/.finished-upgraders
index a802037..2a56593 100644
--- a/.meteor/.finished-upgraders
+++ b/.meteor/.finished-upgraders
@@ -15,3 +15,4 @@ notices-for-facebook-graph-api-2
 1.4.0-remove-old-dev-bundle-link
 1.4.1-add-shell-server-package
 1.4.3-split-account-service-packages
+1.5-add-dynamic-import-package
diff --git a/.meteor/packages b/.meteor/packages
index a017428..8636d50 100644
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -4,14 +4,14 @@
 # 'meteor add' and 'meteor remove' will edit this file for you,
 # but you can also edit it by hand.
 
-meteor-base@1.0.4             # Packages every Meteor app needs to have
+meteor-base@1.1.0             # Packages every Meteor app needs to have
 mobile-experience@1.0.4       # Packages for a great mobile UX
-mongo@1.1.17                   # The database Meteor supports right now
+mongo@1.1.18                   # The database Meteor supports right now
 jquery@1.11.10                  # Helpful client-side library
 tracker@1.1.3                 # Meteor's client-side reactive programming library
 
 es5-shim@4.6.15                # ECMAScript 5 compatibility for older browsers.
-ecmascript@0.7.3              # Enable ECMAScript2015+ syntax in app code
+ecmascript@0.8.0              # Enable ECMAScript2015+ syntax in app code
 
 react-meteor-data
 jagi:astronomy
@@ -19,14 +19,14 @@ random@1.0.10
 static-html@1.1.12_1
 zimme:active-route
 kadira:flow-router
-reactive-dict@1.1.8
+reactive-dict@1.1.9
 universe:i18n
 babrahams:constellation
 mdg:validated-method
 edgee:slingshot
 accounts-password@1.3.6
 standard-minifier-css@1.3.4
-standard-minifier-js@2.0.0
+standard-minifier-js@2.1.0
[...]

PR #8852, opened 2017-06-27
localStorage: bring back proxy methods to window.localStorage. (Fix IE11 issue)

theosp:
IE11 doesn't handle properly attempts to change methods of the
window.localStorage, attempts to do so will result in the complete break of the
localStorage system for the domain in which it is done - until the user clean
the browser/domain cache.

Therefore, in the web, we don't set Meteor._localStorage to be a reference to
window.localStorage . Instead, we set proxy methods.

This will allow package developers that will find a need to change the behavior
of Meteor._localStorage methods to do so without breaking the localStorage
system on IE11. (e.g. meteorhacks:fast-render)

IE11 (earlier IE versions weren't checked) doesn't handle attempts to replace
methods of window.localStorage with different functions properly.  Such attempt
will result in the String of the function we try to set saved as the function,
destroying the ability to use this function.

I couldn't find a way to tell in advance whether an attempt to set
window.localStorage will result in correct function write or not (I
intentionally avoid browser version detection, which is considered a bad
practice). If such attempt will fail we won't have a way to restore the
original function.

The situation is even worse than that. If for example we'll try to set
window.localStorage.setItem = function () {} the String value 'function () {}'
will be saved instead of the function - not only for the current session, but
as part of the localStorage (!) meaning that we'll have to ask users affected
by this bug to clear the cache to fix the situation.

The following won't work:

```javascript
Meteor._localStorage = window.localStorage // Just to make the example clear.
originalSetItem = Meteor._localStorage.setItem
Meteor._localStorage.setItem = function () {}
Meteor._localStorage.setItem = originalSetItem

typeof Meteor._localStorage.setItem -> string
```

-----------

### Potential related issues:

* https://forums.meteor.com/t/entire-meteor-app-non-working-in-ie11-since-meteor-1-5/37139
* https://github.com/meteor/meteor/pull/8782

See also discussion on this commit: https://github.com/jamiter/meteor/commit/a01cd04cfe2f1c9c1955aa22d6733e0d6eb33ef3

Related PR pushed to [meteorspark:fast-render](https://atmospherejs.com/meteorspark/fast-render) to workaround the issue this PR fix: https://github.com/kadirahq/fast-render/pull/191

### DIFF
--- a/packages/localstorage/localstorage.js
+++ b/packages/localstorage/localstorage.js
@@ -24,17 +24,11 @@ if (key === retrieved) {
   if (Meteor.isServer) {
     Meteor._localStorage = storage;
   } else {
-    // IE11 doesn't handle properly attempts to change methods of the
-    // window.localStorage, attempts to do so will result in the complete
-    // break of the localStorage system for the domain in which it is
-    // done - until the user clean the browser/domain cache.
-    //
-    // Therefore, in the web, we don't set Meteor._localStorage to be a
-    // reference to window.localStorage . Instead, we set proxy methods.
-    //
-    // This will allow package developers that will find a need to change
-    // the behavior of Meteor._localStorage methods to do so without breaking
-    // the localStorage system on IE11. (e.g. meteorhacks:fast-render)
+    // Some browsers (e.g. IE11) don't properly handle attempts to change 
+    // window.localStorage methods. By using proxy methods to expose 
+    // window.localStorage functionality, developers can change 
+    // the behavior of Meteor._localStorage methods without breaking 
+    // window.localStorage.
 
     Meteor._localStorage = {
       getItem: function (key) {
--- a/packages/localstorage/localstorage.js
+++ b/packages/localstorage/localstorage.js
@@ -21,7 +21,33 @@ try {
 } catch (ignored) {}
 
 if (key === retrieved) {
-  Meteor._localStorage = storage;
+  if (Meteor.isServer) {
+    Meteor._localStorage = storage;
+  } else {
+    // IE11 doesn't handle properly attempts to change methods of the
+    // window.localStorage, attempts to do so will result in the complete
+    // break of the localStorage system for the domain in which it is
+    // done - until the user clean the browser/domain cache.
+    //
+    // Therefore, in the web, we don't set Meteor._localStorage to be a
+    // reference to window.localStorage . Instead, we set proxy methods.
+    //
+    // This will allow package developers that will find a need to change
+    // the behavior of Meteor._localStorage methods to do so without breaking
+    // the localStorage system on IE11. (e.g. meteorhacks:fast-render)
+
+    Meteor._localStorage = {
+      getItem: function (key) {
+        return window.localStorage.getItem(key);
+      },
+      setItem: function (key, value) {
+        window.localStorage.setItem(key, value);
+      },
+      removeItem: function (key) {
+        window.localStorage.removeItem(key);
+      }
+    };
+  }
 }
 
 if (! Meteor._localStorage) {


packages/localstorage/localstorage.js
```diff
@@ -21,7 +21,33 @@ try {
 } catch (ignored) {}
 
 if (key === retrieved) {
-  Meteor._localStorage = storage;
+  if (Meteor.isServer) {
+    Meteor._localStorage = storage;
+  } else {
+    // IE11 doesn't handle properly attempts to change methods of the
```
hwillson: This comment is a bit verbose, and needs a few grammatical tweaks. How about something like this instead?

```
// Some browsers (e.g. IE11) don't properly handle attempts to change 
// window.localStorage methods. By using proxy methods to expose 
// window.localStorage functionality, developers can change 
// the behavior of Meteor._localStorage methods without breaking 
// window.localStorage.
```

packages/localstorage/localstorage.js
```diff
@@ -21,7 +21,33 @@ try {
...
+    Meteor._localStorage = {
+      getItem: function (key) {
+        return window.localStorage.getItem(key);
+      },
+      setItem: function (key, value) {
+        window.localStorage.setItem(key, value);
+      },
+      removeItem: function (key) {
+        window.localStorage.removeItem(key);
+      }
```
abernix: What about the [`clear()` and `key()` methods](https://developer.mozilla.org/en-US/docs/Web/API/Storage)?

abernix: I _might_ have suggested using a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) object but IE11 [doesn't support it](https://caniuse.com/#feat=proxy) (and that's exactly the browser you're trying to fix!) and it [can't be polyfilled](https://github.com/zloirock/core-js/blob/a55a442a0f8b67ca52b6659889c29861ebb6b183/README.md#missing-polyfills).

theosp: Re clear/key: they were never part of Meteor._localStorage , so I didn't bring them back.

See this commit, in which the proxy method replaced with a reference: https://github.com/jamiter/meteor/commit/a01cd04cfe2f1c9c1955aa22d6733e0d6eb33ef3

packages/localstorage/localstorage.js
```diff
@@ -21,7 +21,33 @@ try {
...
+    };
```
hwillson: This might be a bit of an edge case, but this change could break backwards compatibility. It looks like we're replacing the full `global.localStorage` object that is stored in `Meteor._localStorage` with a proxied version that doesn't include all `global.localStorage` properties. For example, someone could be calling `Meteor._localStorage.key(0)` (see [`Storage.key()`](https://developer.mozilla.org/en-US/docs/Web/API/Storage/key)); this would break using this new proxied `Meteor._localStorage` version. Then again we aren't replicating `key()` in the [manually built `Meteor._localStorage` object](https://github.com/theosp/meteor/blob/0a616060125eeacafc600220bd3a4fd8d88d48bd/packages/localstorage/localstorage.js#L61) either, so maybe this is okay? Food for thought anyways.

hwillson: Oops - I just submitted this without seeing @abernix's recent comments - this comment is a duplicate.

packages/localstorage/localstorage.js
theosp: Will do shortly.

theosp: Should I amend the commit involved, or create a new commit?

For new commit, what commit message is common for such minor changes? (will it be rebased to a single commit when merged?)

hwillson: A new commit is fine - we'll squash when merging (and it's nice to see the commit history while working/reviewing through the PR). A commit message like "Adjusted proxy method explanation comment" or something similar would work. Thanks!

theosp: Done, thanks 👍

noelheesen:
Hello, we at [part-up](https://github.com/part-up/part-up) have the exact same issue. It's an awkward problem since a lot of our users are on IE. Any idea when this PR can be approved?

benjamn:
This should be fixed now with `localstorage@1.1.1`. Thanks for fixing this problem!

noelheesen:
@benjamn Hello, thanks for the fix! Will you add this in the new meteor update or do we need to do include this ourselves to make use of this fix?

hwillson:
@noelheesen For existing apps (that already use `localstorage`), it will be updated the next time you run `meteor update` (or `meteor update localstorage`).

PR #8853, opened 2017-06-27
[Work in progress] Add mongo-dev-server package

zimme:
Only start the MongoDB server if this package is present in the project.

For now I've added mongo-dev-server as a dependency of mongo, do we want to do this?

A lot of packages rely on the meteor package and when using Meteor as a build system you might wanna use packages like static-html or dynamic-import but you might still not want a local mongodb instance started.

I'm thinking we could remove mongo-dev-server as a mongo dependency and use the upgrades thing to just add this package to all current meteor projects on upgrade and add it to the default packages when creating a new project.

The way this PR is now results in MONGO_URL being set to `'no-mongo-server'` if `mongo-dev-server` isn't present and `MONGO_URL` hasn't been manually set. This will prevent Meteor from trying to connect to a local mongo server even if `mongo` is present in the project. I don't know if this have any other negative effects.

This is a first step at resolving meteor/meteor-feature-requests#31.


zimme:
Turned out that the changes in #8769 was enough for this to work. There was not need to do the same for `.meteor/versions` to cover this use-case.

zimme:
I'll have a look at the self tests, some might fail if they expect MongoDB without adding `mongo`.

packages/mongo-dev-server/README.md
```diff
@@ -0,0 +1,4 @@
+# mongo-dev-server
+[Source code of released version](https://github.com/meteor/meteor/tree/master/packages/mongo-dev-server) | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/mongo-dev-server)
+***
+
```
hwillson: Would you mind adding one or two sentences describing the package in the README?

packages/mongo-dev-server/package.js
```diff
@@ -0,0 +1,11 @@
+Package.describe({
+  debugOnly: true,
+  documentation: 'README.md',
+  name: 'mongo-dev-server',
+  summary: 'Start MongoDB alongside Meteor, in development mode.',
+  version: '1.0.0',
+});
+
+Package.onUse(function (api) {
+  api.addFiles('info.js', 'server');
```
hwillson: Maybe consider using `api.mainModule` here instead?

packages/mongo/package.js
```diff
@@ -34,7 +34,8 @@ Package.onUse(function (api) {
     'diff-sequence',
     'mongo-id',
     'check',
-    'ecmascript'
+    'ecmascript',
+    'mongo-dev-server',
```
hwillson: Relating to your comment:

> I'm thinking we could remove mongo-dev-server as a mongo dependency and use the upgrades thing to just add this package to all current meteor projects on upgrade and add it to the default packages when creating a new project.

I like this idea - Not forcing `mongo-dev-server` as a dependency, and using an `upgrader` instead, gets my vote!

tools/runners/run-all.js
```diff
@@ -70,11 +70,18 @@ class Runner {
       proxyToHost: appHost,
       onFailure
     });
+    
+    buildmessage.capture(function () {
+      self.projectContext.resolveConstraints();
+    });
+
+    const packageMap = self.projectContext.packageMap;
+    const hasMongoDevServerPackage = packageMap && packageMap.getInfo('mongo-dev-server') != null;
```
hwillson: Would you mind switching to use `!==` just to make sure we're matching the rest of the source.

zimme: I used `!=` to match both `null` and `undefined`.

hwillson: Thanks @zimme, it's probably okay to leave it then. I personally prefer to not use lenient equality / non-equality checks on `null` / `undefined`, as when going back over old code I always end up questioning whether the leniency was intended or not ... 🙂.

packages/mongo-dev-server/README.md
zimme: I can do this.

packages/mongo-dev-server/package.js
zimme: I'll have a look.

packages/mongo/package.js
zimme: I think I'd need help with how to add this to the `upgrader` instead.

hwillson: Sure thing - I haven't had to build one myself, but as far as I understand things, you'll be looking to add your changes in `tools/upgraders.js`. The process seems to be pretty well documented in that file - you'll be adding a new upgrade name / section, and you can test things out by running `meteor admin run-upgrader myupgradername`. The `1.5-add-dynamic-import-package` upgrader section in that file might be a good one to copy / start from.

zimme:
Just wanted to give you guys a heads up, I'm heading off on vacation on Friday for 2 weeks and I don't think I'll have time to finish this up before that.

hwillson:
Nooo, you can't leave!! Seriously though, no problem at all @zimme - thanks for getting things this far. I'll aim to make the remaining small changes, over the next few days (schedule permitting of course - you might find this PR waiting for you when you get back 😉). Have a great 🏖 !

hwillson:
We should be all set here now - LGTM!

abernix:
LGTM, thanks!

Issue #8856, opened 2017-06-28
Upgrade Facebook Graph API version


yogiben:
We received the following warning from fb:

```
Your App, XXX, is currently accessing Graph API v2.3 which will reach the end of its 2-year lifetime on 10 July, 2017. We estimate 2 endpoints that Bitwala calls will be impacted by this change and may stop working after the automatic upgrade push. To ensure a smooth transition, please migrate all calls to Graph API v2.4 or higher. Use the API Upgrade Tool to understand exactly how this change will impact your app.

Note: If you do not see any impacted calls in the API Upgrade Tool, your app/code may not be affected by this upgrade.

Visit our changelog to see the full list of changes in all Graph API versions.
```

This line should be updated:
https://github.com/meteor/meteor/blob/master/packages/facebook-oauth/facebook_client.js#L34

hwillson:
This was discussed a little while back in a commit comment - https://github.com/meteor/meteor/commit/873f13d743053fc080b5562dec786784bc52a610#commitcomment-22062669. This should definitely be looked into - PR's welcome! 🙂

abernix:
Thank you very much for filing this issue.

Not that this shouldn't be fixed, but out of curiosity, does the [API Upgrade Tool](https://developers.facebook.com/tools/api_versioning/) for that app show any impacted calls?  Facebook APIs never actually stop working, but instead roll over to the next still-active version.

Just trying to gauge the impact here for older users since there is a ramp-up period before older users update, and also some that never get updated and I'd like to be, let's say, mentally prepared for July 10th. :wink:  Honestly, I'm not sure there are actually any breaking changes in that particular API with the way Meteor uses it and I'm hoping that this is as easy rolling over to the new version number in the API URL.

hwillson:
PR #8858 was merged, meaning new `facebook-oauth` and `accounts-facebook` packages are now available. Run `meteor update` to get them - thanks!

PR #8858, opened 2017-06-29
Update Facebook api version in loginUrl

hasantayyar:
The most recent version of the API is Version 2.9, which was introduced on April 18th, 2017 https://developers.facebook.com/docs/apps/changelog

The url parameters are not changed with the new version https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow

Fixes https://github.com/meteor/meteor/issues/8856

### DIFF
--- a/packages/accounts-facebook/package.js
+++ b/packages/accounts-facebook/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Login service for Facebook accounts",
-  version: "1.2.0"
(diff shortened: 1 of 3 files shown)


abernix:
Thanks for submitting this!  I've published a new `facebook-oauth@1.3.2` and a `accounts-facebook@1.2.1` (which depends on it) with this change, bringing the Facebook Graph API to v2.9 for the login dialog.

PR #8859, opened 2017-06-29
updated dependency of semver to the latest version

fschaeffler:
The currently used version of semver does have a medium-level security issue. Therefore, the dependencies of `meteor-dev-bundle` had been updated.

|            | Regular Expression Denial of Service   |
| ---------- | -------------------------------------- |
| Name       | semver                                 |
| CVSS       | 5.3 (Medium)                           |
| Installed  | 4.1.0                                  |
| Vulnerable | <4.3.2                                 |
| Patched    | 4.3.2                                  |
| Path       | meteor-dev-bundle@0.0.0 > semver@4.1.0 |
| More Info  | https://nodesecurity.io/advisories/31  |

### DIFF
--- a/scripts/dev-bundle-server-package.js
+++ b/scripts/dev-bundle-server-package.js
@@ -17,7 +17,7 @@ var packageJson = {
     // (and there are backwards-incompatible changes either way).
     underscore: "1.5.2",
     "source-map-support": "https://github.com/meteor/node-source-map-support/tarball/1912478769d76e5df4c365e147f25896aee6375e",
-    semver: "4.1.0"
+    semver: "4.3.2"
   },
   // These are only used in dev mode (by shell.js) so end-users can avoid
   // needing to install them if they use `npm install --production`.
--- a/scripts/dev-bundle-server-package.js
+++ b/scripts/dev-bundle-server-package.js
@@ -17,7 +17,7 @@ var packageJson = {
     // (and there are backwards-incompatible changes either way).
     underscore: "1.5.2",
     "source-map-support": "https://github.com/meteor/node-source-map-support/tarball/1912478769d76e5df4c365e147f25896aee6375e",
-    semver: "4.3.2"
+    semver: "5.3.0"
   },
   // These are only used in dev mode (by shell.js) so end-users can avoid
   // needing to install them if they use `npm install --production`.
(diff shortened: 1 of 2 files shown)


abernix:
Looking at the [commit history](https://github.com/npm/node-semver/commits/master) I'm fairly sure we can go all the way to `5.3.0`.  Was there a particular reason you avoided it?

fschaeffler:
@abernix In order to possible prevent side-effects, I did update the package only up until to first version, when the issue got fixed. But your right, an update to 5.3.0 does make sense. So I'll update the PR.

fschaeffler:
@hwillson are there any further steps needed from my side for the publish of a new dev bundle version? As this is my first PR into https://github.com/meteor/meteor, I'm not sure what steps are all needed for this to happen.

hwillson:
No, nothing else is required on your side - the next steps have to be handled by MDG staff (see [Submitting "Dev Bundle" Pull Requests](https://github.com/meteor/meteor/blob/devel/Development.md#submitting-dev-bundle-pull-requests) for more details). Thanks for asking!

hwillson:
Wow, good catch @abernix! Here are the only differences between [`semver410.js`](https://github.com/meteor/meteor/blob/98b02f363f1f7b83fb42267245ed3d60d6da46ee/packages/package-version-parser/semver410.js) and [`semver` v4.1.0](https://raw.githubusercontent.com/npm/node-semver/v4.1.0/semver.js):

```js
// <METEOR>
// Fool the module system detection code below so that it doesn't
// do anything special.
var exports = SemVer, module = {}, define = {};
// Create a package-private variable.  Can't use SemVer because
// of the code that says `function SemVer(...)` below (implicitly
// declaring a var).  Can't use "semver" because that's a var in
// package-version-parser.js.
SemVer410 = SemVer;
// </METEOR>
```

So it looks like at one point it was necessary to override `semver`'s default module handling code, to work within Meteor's package system. Digging a bit deeper, it looks like the [`package-version-parser.js`](https://github.com/meteor/meteor/blob/98b02f363f1f7b83fb42267245ed3d60d6da46ee/packages/package-version-parser/package-version-parser.js) file is being used both by the Meteor Tool and by the `package-version-parser` package. When used by the Tool the npm based version of `semver` (from the `dev_bundle`) is used. When used by the `package-version-parser` package the overridden version above (via `SemVer410`) is used.

A lot has changed since this code was put in place, so we should (fingers crossed) be able to adjust things such that the `package-version-parser` package references the same npm based `semver` version.

hwillson:
Quick update here - once https://github.com/meteor/meteor/pull/8914 has been merged, we should be all set to continue with this PR.

PR #8864, opened 2017-06-30
Upgraded to Mongo 3.4.5.

hwillson:
Hi all - this PR is intended to address feature request https://github.com/meteor/meteor-feature-requests/issues/129. The Meteor tool has been updated to use Mongo 3.4.5 (the latest stable version as of today). For the most part, this has been a pretty painless upgrade. There have been quite a few changes since 3.2 (see the 3.4 changelog [here](https://docs.mongodb.com/manual/release-notes/3.4-changelog/)), but most of those changes have been internal to the Mongo core itself). All existing `self-test`'s and `package-test`'s are passing (after a few small tweaks), and I've been manually testing these changes against older databases without any issues.

Since these changes require a new version of the dev bundle, CircleCI's `self-tests`'s will fail until a new version is published by MDG. Given the other changes that are happening currently, I wasn't sure what would be an appropriate `BUNDLE_VERSION`, so for now I've just bumped the minor by 1 (let me know if that should change).

Thanks all - standing by for feedback!

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,11 @@
 ## v.NEXT
 
+* Meteor has been upgraded to support Mongo 3.4.5 by default (the bundled
+  version used by `meteor run` has been upgraded). Internally it now uses the
+  2.2.29 version of the `mongodb` npm driver, and has been tested against
+  Mongo databases ranging from versions 2.6.10 to 3.4.5.
+  [FR #129](https://github.com/meteor/meteor-feature-requests/issues/129)
+
 * `observe`/`observeChanges` callbacks are now bound using `Meteor.bindEnvironment`.
   The same `EnvironmentVariable`s that were present when `observe`/`observeChanges`
   was called are now available inside the callbacks.
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=4.8.19
+BUNDLE_VERSION=9999.8864.0
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=4.8.18
+BUNDLE_VERSION=4.8.19
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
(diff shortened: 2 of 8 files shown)


abernix:
Thanks for taking a look at this, @hwillson.

The commit I just pushed (to build a test dev bundle) failed relatively quickly because MongoDB has [discontinued support for 32-bit versions](https://docs.mongodb.com/manual/installation/#supported-platforms).  I found this out fairly quickly when the dev bundle failed to download MongoDB on both Linux and Windows i386 platforms.

I'm going to try to get some Meteor usage stats to determine the impact of this change, but it seems like we're going to have to deprecate those platforms going forward and get a 64-bit version of Meteor for Windows up and going.

hwillson:
Ouch! It would have been nice if Mongo mentioned that in their [3.4 changelog](https://docs.mongodb.com/manual/release-notes/3.4/) 🙁. Okay, makes sense @abernix - thanks for looking into this. Sounds like we're being forced into 64-bit land (which might not be a bad thing). Let me know what I can do to help to move this (and a 64-bit Meteor Windows version) forward.

abernix:
After looking at some usage stats today, 0.97% of our total Meteor 1.4 and 1.5 *nix usage in the last 30 days was on a 32-bit architecture.

hwillson:
Awesome, thanks for checking @abernix! Given that low number, do you think we should start to move ahead with discontinuing 64-bit support? I'm definitely leaning towards yes (as eventually we're going to get a little too far out of synch with Mongo).

skirunman:
Would it be possible, or even make sense, to release the default support for Mongo 3.4 on Linux and MacOS first, and then resolve the Windows 32 bit issue?

hwillson:
@skirunman Do you mean something along the lines of releasing Meteor 1.5.2 with Mongo 3.4 support for *nix based users only, leaving Windows at 1.5.1 for now? While this is technically possible, I don't think the workarounds / effort involved would be time well spent. I think we need to move full steam ahead towards 64-bit (only) releases of Meteor for all platforms. As of right now, that is the plan (it's just a question of when). This PR has been added to the `Release 1.6` milestone, so more work in this area will be happening shortly. Thanks!

sprohaska:
Maybe reconsider taking <https://github.com/meteor/meteor/pull/8870> into a bugfix release, since Mongo 3.4 seems to take longer than anticipated.

hwillson:
Thanks @sprohaska, that's definitely worth considering (along with @skirunman's  https://github.com/meteor/meteor/pull/8911). I'll re-open https://github.com/meteor/meteor/pull/8870 for now, but the discussion that comes out of https://github.com/meteor/meteor-feature-requests/issues/152 will hopefully help steer things.

skirunman:
@hwillson 

> Do you mean something along the lines of releasing Meteor 1.5.2 with Mongo 3.4 support for *nix based users only, leaving Windows at 1.5.1 for now?

Yes, that is what I meant. I agree probably best to move forward with 64-bit support ASAP.

hwillson:
This PR has been superseded by https://github.com/meteor/meteor/pull/9173.

Issue #8648, opened 2017-04-28
Meteor segfault randomly on reload (localy)


damjuve:
Hi,

When i use meteor locally, server start correctly but I face frequent segfault which appear randomly (i cant figure exactly what cause it).
It seems to occure when server reload (due to a file modification). But doesn't segfault on every reload, and doesn't segfault on same files (sometimes modificaiton on server side, sometimes on client side).

My RAM isn't full (about 10Gb free when segfault occure).
And here is all package i use : 

`meteor-base@1.0.4             # Packages every Meteor app needs to have
mobile-experience@1.0.4       # Packages for a great mobile UX
mongo@1.1.16                   # The database Meteor supports right now
blaze-html-templates@1.0.4 # Compile .html files into Meteor Blaze views
reactive-var@1.0.11            # Reactive variable for tracker
jquery@1.11.10                  # Helpful client-side library
tracker@1.1.2                 # Meteor's client-side reactive programming library

standard-minifier-css@1.3.4   # CSS minifier run for production mode
standard-minifier-js@2.0.0    # JS minifier run for production mode
es5-shim@4.6.15                # ECMAScript 5 compatibility for older browsers.
ecmascript@0.7.2              # Enable ECMAScript2015+ syntax in app code
shell-server@0.2.3            # Server-side component of the `meteor shell` command

accounts-ui@1.1.9
kadira:flow-router
zimme:active-route
underscore@1.0.10
kadira:blaze-layout
accounts-password@1.3.5
accounts-facebook@1.1.1
accounts-google@1.1.2
google-config-ui@1.0.0
facebook-config-ui@1.0.0
twbs:bootstrap
tanis:bootstrap-social
fortawesome:fontawesome
less@2.7.9
email@1.2.0
check@1.2.5
meteorhacks:ssr
arillo:flow-router-helpers
cfs:filesystem
ostrio:files
session@1.1.7
dburles:google-maps
sergeyt:typeahead
raix:handlebar-helpers
dburles:collection-helpers
manuel:reactivearray
aldeed:collection2-core
mizzao:bootboxjs
meteorhacks:aggregate
aldeed:autoform@6.0.0
summernote:summernote
anback:bootstrap-validator
aldeed:autoform-bs-datepicker
natestrauser:select2
rajit:bootstrap3-datepicker
natestrauser:font-awesome
mystor:device-detection
reactive-dict@1.1.8
aldeed:simple-schema
froala:editor
andrasph:clockpicker
aldeed:template-extension
[...]

PR #8866, opened 2017-06-30
Make METEOR_WATCH_PRIORITIZE_CHANGED opt-out instead of opt-in.

benjamn:
This restores the behavior of 8c70716954f8fcdd02fc0c673a39da2b7186ce2b by default, with the option of disabling the prioritized file watching system by setting `METEOR_WATCH_PRIORITIZE_CHANGED` explicitly to `"false"`.

The self-tests where the environment variable is explicitly set form a nice to-do list of tests that should be improved to be more robust to cope with differences in file watcher timing.

Helps with #8648 and similar issues.

### DIFF
--- a/tools/fs/safe-watcher.js
+++ b/tools/fs/safe-watcher.js
@@ -17,9 +17,14 @@ if (process.platform === "win32") {
   WATCHER_ENABLED = false;
 }
 
-var PRIORITIZE_CHANGED =
-  process.env.METEOR_WATCH_PRIORITIZE_CHANGED &&
-  JSON.parse(process.env.METEOR_WATCH_PRIORITIZE_CHANGED);
+// Default to prioritizing changed files, but disable that behavior (and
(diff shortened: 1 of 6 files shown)


davidecantini:
After upgrading to 1.5.1, I started to experience cases like modifications to files not being immediately effective, just like if the previous (cached) version of the modified files are (wrongly) used. That makes the (local) development cumbersome. The good thing is that after setting METEOR_WATCH_PRIORITIZE_CHANGED=false, things appeared to work properly like they've always been until version 1.5. This is just a heads up that the prioritized file watching system may have some problems and not be ready (if so, METEOR_WATCH_PRIORITIZE_CHANGED=false, should be the default, until it's more stable)

klaussner:
@davidecantini That sounds like [this issue](https://github.com/meteor/meteor/issues/8942), which will be fixed in Meteor 1.5.2. :slightly_smiling_face:

PR #8880, opened 2017-07-06
Fixes #8879 Build phase command line output jumps

### LINKED ISSUE #8879: Build phase console output "jumps" in Windows
### Related to #6785 

See https://github.com/meteor/meteor/issues/6785#issuecomment-279983515

> I have tried version 1.4.3.1, and the problem is improved but not fully fixed. The command line output does not go back to the first line of the console, but it does jump back several lines before it stabilises. The exact number of lines seems to depend on the width of the console - e.g. for a width of 80 characters, it jumped 13 lines; for 120 characters, it jumped 8 lines...

I have reproduced this when building with Meteor 1.5 on Windows 10 where PowerShell is used from vscode. The [original fix](https://github.com/meteor/meteor/pull/8082) relied on the window title property to override the default `cmd.exe` behaviour for PowerShell. It looks like this test can fail (e.g. from vscode).

robfallows:
Fixes #8879  (or refixes #6785).

The test for 'Windows Powershell' in the user's `window.title` property is "inverted" to instead test for "not executable name containing `cmd`".

This seems to be more reliable and in line with Microsoft's intention of removing `cmd.exe` from Windows 10 - i.e. that it will be more strategic to adopt non-`cmd.exe` ways of working.

### DIFF
--- a/tools/console/console.js
+++ b/tools/console/console.js
@@ -67,8 +67,10 @@ var wordwrap = require('wordwrap');
 
 var PROGRESS_DEBUG = !!process.env.METEOR_PROGRESS_DEBUG;
 var FORCE_PRETTY=undefined;
-var CARRIAGE_RETURN =
-  (process.platform === 'win32' && process.stdout.isTTY && process.title !== 'Windows PowerShell' ? new Array(249).join('\b') : '\r');
+// Set the default CR to \r unless we're running with cmd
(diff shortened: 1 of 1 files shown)


benjamn:
Should be fixed with Meteor 1.5.1. Thanks for the PR!

PR #8885, opened 2017-07-08
filePath is relative to appPath

mitar:
If I have a directory `test-app` with `packages` subdirectory, and I run outside `test-app` directory `meteor create --bare test-app` it fails on `files.stat` because `files.stat('packages')` is run outside `test-app`, and it does not exist there.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -661,7 +661,7 @@ main.registerCommand({
       // We do mind if there are non-hidden directories, because we don't want
       // to recursively check everything to do some crazy heuristic to see if
       // we should try to creat an app.
-      var stats = files.stat(filePath);
+      var stats = files.stat(files.pathJoin(appPath, filePath));
       if (stats.isDirectory()) {
         // Could contain code
         return true;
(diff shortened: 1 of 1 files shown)


tools/cli/commands.js
```diff
@@ -660,8 +660,8 @@ main.registerCommand({
 
       // We do mind if there are non-hidden directories, because we don't want
       // to recursively check everything to do some crazy heuristic to see if
-      // we should try to creat an app.
-      var stats = files.stat(filePath);
+      // we should try to create an app.
+      var stats = files.stat(files.pathJoin(appPath, filePath));
```
benjamn: I know this code used `files.stat` unsafely already, but perhaps we should use `files.statOrNull` and then `stats && stats.isDirectory()` below, just in case the new path does not exist?

mitar: Path should exist, because we are just iterating over above. You are saying that if path gets deleted in between as a race condition?

mitar:
Not sure what is practice about deleting such small branches after merging?

PR #8889, opened 2017-07-09
added command make clean for linux

MohmadAyman:
in response for feature request [#77](https://github.com/meteor/meteor-feature-requests/issues/77) 
where would the ```.meteor``` normally be on windows so I can add its support.
I'll add a test once I know what would we with windows.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1286,7 +1286,6 @@ main.registerCommand({
   allowUnrecognizedOptions: true,
   catalogRefresh: new catalog.Refresh.Never()
 }, function (options) {
-  Console.info(options);
   if (options.global) {   
     try {
       files.rm_recursive(path.join(process.env.HOME, '.meteor'));
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1,4 +1,5 @@
 var main = require('./main.js');
+var path = require('path');
 var _ = require('underscore');
 var files = require('../fs/files.js');
 var deploy = require('../meteor-services/deploy.js');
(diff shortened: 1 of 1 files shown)


stubailo:
There's definitely a function somewhere in the tool code that will give you the location of the directory across all platforms. Don't have time to look for it now but we should use it for sure.

stubailo:
Also `meteor clean` feels like it would clean the build files for the current project - since that's what something like `make clean` usually does. I'd pick a different name for a command that cleans out the global package cache.

MohmadAyman:
according to the discussion @ #3614 cleaning would delete the ```.meteor``` not the one in the current project, I think cleaning the project would better has its own command like ``` meteor clean-project ``` , what do you think?
Thanks, found ```defaultWarehouseDir() ``` that does the job.

stubailo:
I agree the functionality is super important, I'm just calling out that I think most people would expect a `meteor clean` command to do a different thing. That's just one person's opinion though, hope people will chime in!

MohmadAyman:
Made some changes, Now make clean cleana the project, and make clean global deletes the global .meteor .

abernix:
The original feature request was to clear `~/.meteor` of unused packages.  This doesn't seem to meet the criteria of that feature request.  For example, when using the `--global` flag this just recursively deletes the entire `~/.meteor/` directory, regardless of whether a package is used or not, which is a bit excessive since it also wipes the local package catalog and requires a time consuming full-sync.  This is really no different than `rm -rf ~/.meteor`.  I'll point out that there really isn't a method of determining whether a package is used.

Additionally, when running `meteor clean` without the `--global` flag, this PR (as submitted) wipes the apps entire `.meteor` directory, which completely breaks the app since multiple files which live within the `.meteor` directory are required, such as `packages`, `version`, `release`, etc.   Currently, there is already a `meteor reset` –  for the local app, doesn't this already accomplish what's necessary?

abernix:
Thanks for the contribution effort however this doesn't really work as it should.  A future PR of this nature might try to address how to determine the necessity of a package, perhaps based on how recently it's been used, though that exact logic will need to be worked through in an issue before it gets submitted as a PR.

PR #8890, opened 2017-07-10
Adjust EJSON API to handle objects with properties named "length"

hwillson:
Hi all - this PR is intended to fix issue #5175. A few parts of the EJSON API won't work properly when manipulating objects with properties named length. This is the same problem that was identified in issue #8329, where underscore is improperly treating objects with a length property as an array (within several of its functions). This is a long-standing issue with underscore (that won't be changed as they don't see it as a bug), so for now I'm removing the _.each and _.size calls that create this issue (not all of them - just the ones that directly impact the EJSON public API and cause this issue). Quick side note - lodash does not have this problem, so this issue could have also been resolved by switching to lodash.

Side note - this PR replaces #8824. Adding `ecmascript` as a dependency in the `ejson` package (without making any other changes) causes Circle to fail (for the full details see https://github.com/meteor/meteor/pull/8824#issuecomment-313979147). This replacement PR stays focused on just getting the original issue fixed for now.


hwillson:
By swapping the order of 2 `self-test`'s in `tools/tests/package-tests.js`, I was able to get all tests to pass, with `ecmascript` added in. I'm not sure why calling the `add packages to app` self-test before `add debugOnly and prodOnly packages` causes problems (test cleanup issue?), but re-arranging them fixes the issue. Given this, I've gone ahead and refactored the `ejson` package to remove the `underscore` dependency. This should now be ready for review.

Note: https://github.com/meteor/meteor/pull/8821 should be merged before this PR. Once it's merged, I'll rebase and update this PR to include those changes.

benjamn:
Note from @hwillson: https://github.com/meteor/meteor/pull/8821 should be merged first.

hwillson:
Now that https://github.com/meteor/meteor/pull/8821 has been merged, I've pulled those changes in here and updated everything to match the rest of the `underscore` removal changes. Tests are passing (yay!) so this is now ready for review. Thanks!

hwillson:
Quick note - https://github.com/meteor/meteor/pull/8949 needs to be merged before this.

hwillson:
#8949 has been merged here and updated, so we should be all set with this one.

benjamn:
@hwillson How far should we bump the `ejson` package version?

hwillson:
@benjamn There aren't really any outward facing functionality changes (other than a bug fix), so I believe keeping it at the `1.0.14` I bumped it to should be okay?

benjamn:
Rebased and merged into `devel` as the following series of commits:
1. 82a6feeaf43b11766645c065f87bd7322c969ce9
2. c6bddb331a8fd8c24336c6a7232406a3f108a6b5
3. a7aa0175f6b8d9de1926e3cd9cf1ca69f346bc45
4. 5b39d4b2da3cda2137db7cc68b595ba27ca2c214

PR #8893, opened 2017-07-10
Minimongo ES5/ES6 refactoring and performance improvements

radekmie:
Part 1 of meteor/meteor-feature-requests#146.


packages/minimongo/selector.js
```diff
@@ -801,7 +801,7 @@ ELEMENT_OPERATORS = {
         throw Error("$elemMatch need an object");
 
       var subMatcher, isDocMatcher;
-      if (isOperatorObject(_.omit(operand, _.keys(LOGICAL_OPERATORS)), true)) {
+      if (isOperatorObject(Object.keys(operand).filter(key => !Object.keys(LOGICAL_OPERATORS).includes(key)).reduce(function (a, b) { return Object.assign(a, {[b]: operand[b]}); }, {}), true)) {
```
radekmie: I'll take care of lines like these during the next phases.

packages/minimongo/minimongo_tests.js
```diff
@@ -1570,7 +1574,7 @@ Tinytest.add("minimongo - projection_compiler", function (test) {
 
 Tinytest.add("minimongo - fetch with fields", function (test) {
   var c = new LocalCollection();
-  _.times(30, function (i) {
+  Array.from({length: 30}, function (_, i) {
```
hwillson: Underscore is accessible in this test file, so you might want to pick another variable name to represent the elements of the array.

packages/minimongo/minimongo_tests.js
```diff
@@ -2099,11 +2103,11 @@ Tinytest.add("minimongo - array sort", function (test) {
...
     var fieldValues = [];
     c.find().forEach(function (doc) {
-      if (_.has(doc, field))
+      if (doc.hasOwnProperty(field))
         fieldValues.push(doc[field]);
     });
-    test.equal(_.pluck(cursor.fetch(), field),
-               _.range(_.max(fieldValues) + 1));
+    test.equal(cursor.fetch().map(function (doc) { return doc[field]; }),
+               Array.from({length: Math.max.apply(null, fieldValues) + 1}, function (_, i) { return i; }));
```
hwillson: Same variable name comment as above (maybe use a different name since Underscore is available in this test file).

packages/minimongo/selector.js
hwillson: For now, could you update this to wrap at 80 chars?

radekmie: I didn't managed to fit it 80 _(97)_, but it will be definitely shorter with arrow functions.

radekmie:
Current status:
* files are logically separated
* most of ES5/6 syntax is already done _(`let`, `const`, no `self`, etc.)_
* I've updated your ESLint config from the admin scripts locally and `--fix`ed this package _(it created some weird very very long lines though)_
* tests are passing!

TODO:
* break these long lines 
* review it
* review it again

Questions:
* Should I continue refactoring here or in another PR? I'd rather optimize and refactor at once than read and think it ten times.
* How about reviving ESLint for good? Only few rules are outdated _(and dependencies of course)_, but it's a lot easier, especially with such big files.

packages/minimongo/package.js
```diff
@@ -1,66 +1,46 @@
...
+    'tracker',
   ]);
 
-  // Functionality used only by oplog tailing on the server side
-  api.addFiles([
-    'selector_projection.js',
-    'selector_modifier.js',
-    'sorter_projection.js'
-  ], 'server');
+  api.mainModule('main.js', 'client');
```
hwillson: Would you mind changing `main.js` here to be `main_client.js` (and rename the file)? Just to be more uniform with the `main_server.js` naming.

radekmie: I have also thought about it but `main.js` is used also on the server - `main_server.js` imports it.

hwillson: Oh, sorry I didn't notice that. In that case, would you mind making the following changes:

- Create a new `minimongo_common.js` file that holds the current contents of `main.js`
- Rename `main.js` to `minimongo_client.js`, and just import/export the `minimongo_common.js` contents (and update the `package.js` to have `api.mainModule('minimongo_client.js', 'client');` so this is the new client mainline).
- Rename `main_server.js` to `minimongo_server.js` and import `minimongo_common.js` (and update the `package.js` accordingly)

This will then fall more in-line with client/server shared code approach used in other core packages. Thanks!

radekmie: OK, looks good.

radekmie:
I'm mostly done so some more in-depth tests would be great. I'll think more about optimisations later - now simply make _this_ happen.

radekmie:
Any progress here? Should I rebase now or wait some more?

hwillson:
Sorry for the delay @radekmie. I'll plan on reviewing this tomorrow (if no one else gets to it before me). If you don't mind rebasing though, that would be great - thanks!

radekmie:
I'll try to do my best, but it seems horrible :|

radekmie:
I've tried to rebase, but it was too much to work with it. I've simply merged in and make tests pass. If something is missing or wrong, let me know.

radekmie:
Good to hear! At first I was threatened: another conflicts! It turned out to be... One line. One *commented* line. Anyway, it's ready and I'm looking forward the merge!

packages/minimongo/local_collection.js
```diff
@@ -1200,7 +1200,7 @@ LocalCollection._selectorIsIdPerhapsAsObject = selector => {
   if (LocalCollection._selectorIsId(selector))
     return true;
 
-  return LocalCollection._selectorIsId(selector && selector._id) && Object.keys(selector).length === 1;
```
quintstoffers: Before, this function would only return true when `_id` was the only property in the selector, but now it'll return true when the `_id` property in the selector is an ID, regardless of whether or not other properties are set.

Looking at usages of this function there could be some cases where the following query is now considered an `_id`-only query:
```
{
    "_id": "foo",
    "bar": "baz"
}
```
For instance in [allow-deny](https://github.com/meteor/meteor/blob/8005532f4ba847d3e8bb4c1100d09bf53136ff0a/packages/allow-deny/allow-deny.js#L265), where updates are denied when it's not a pure `_id` query. In [minimongo](https://github.com/meteor/meteor/blob/08f5ea4c2b75c003df6625ed5e9dadea02e4414d/packages/minimongo/minimongo.js#L98) it looks like this change would return the document with id "foo" regardless of whether the "bar" part matches.

radekmie: Yep, I've made this assumption too early. Thanks for review anyway!

hwillson:
Thanks for the recent updates @radekmie - LGTM!

packages/minimongo/cursor.js
```diff
@@ -42,7 +45,8 @@ export default class Cursor {
    */
   count() {
     if (this.reactive)
-      this._depend({added: true, removed: true}, true /* allow the observe to be unordered */);
+      // allow the observe to be unordered
+      this._depend({added: true, removed: true}, true);
```
abernix: 👍 😄

radekmie:
Yep, it is a lot to read, @abernix! Anyway - if I'd have to submit a different PR to Meteor I'd split it but in this very case it made sense to do it this way.

To sum up:
* Refactor all `hasOwnProperty` calls (right?).
* Remove these two commented _imports_.
* These strings were _fixed_ by `eslint --fix` I did (should I revert it?).
* Samo with the braces (should I revert it?).
* `Object.create`?
* Update branch. Again. Argh!

It turned out that I'm radical not purist :|

abernix:
> Refactor all hasOwnProperty calls (right?).

Please!

> Remove these two commented imports.

Yes.  Seems like they're safe to be removed and don't serve a purpose. 😄  

> These strings were fixed by eslint --fix I did (should I revert it?).
> Samo with the braces (should I revert it?).

ESLint will do different things with the `--fix` flag depending on what settings are enabled at the time.  In our case, we prefer to have the braces and in general, try to avoid auto-changing code outside of a planned pull-request.  See [Code Style](https://github.com/meteor/meteor/blob/devel/Development.md#code-style) for more information, and there is an `.eslint` configuration in `scripts/admin/eslint/.eslintrc` which should match up to the guidelines.

> Object.create?

`Object.create(null)` creates a new object with no prototype and no inherited properties whereas `{}` creates an object with properties inherited from `Object.prototype` (like `toString`), which are frequently unwanted on objects like this.  For example, if a user has a `path` of "`toString`" (silly as it may sound!) it may turn into a problem when we look for it since `{}["toString"]` already exists!  [This StackOverflow page](https://stackoverflow.com/a/21079232) has some decent explanations.

> Update branch. Again. Argh!

Hopefully it will be less painful.  If it's a problem, someone else can take care of the merge; your work here is already greatly appreciated!

radekmie:
OK, I'll try to update it today.

About that ESLint - I ran it with this very config but with the newest version of it - some rules were outdated.

(next time I have to say something more than _`Object.create`?_ - I meant _should I do it?_, not _what is it_...)

abernix:
> Object.create? - I meant should I do it?, not what is it...)

Ah, yes, please; we should try to use the safer practice if it's reasonable to do. 👍

hwillson:
We're so close on this - super excited to get this merged!

PR #8914, opened 2017-07-14
Modernize package-version-parser using modules.

benjamn:
Also removed the underscore dependency while I was at it.

cc @abernix @hwillson @fschaeffler

Prerequisite for #8859.


benjamn:
Goes without saying, but I'll keep working on this until the tests pass. 😉

packages/package-version-parser/package-version-parser.js
```diff
@@ -5,10 +5,8 @@
 // true).
 var inTool = typeof Package === 'undefined';
 
-
-var semver = inTool ?
-  require ('../../dev_bundle/lib/node_modules/semver') : SemVer410;
-var __ = inTool ? require('../../dev_bundle/lib/node_modules/underscore') : _;
+// Provided by dev_bundle/server-lib/node_modules/semver.
+var semver = require('semver');
```
hwillson: @benjamn - Looking at the failing tests, could we just leverage the `module.parent` here, if available, to make sure `semver` is pulled in when coming from `tools/packaging/package-version-parser.js`? So change 

```js
var semver = require('semver');
```

to something like:

```js
var moduleParent = module.parent;
var semver =
  moduleParent && moduleParent.require
    ? moduleParent.require('semver') : require('semver');
```

hwillson:
Just noting this here - once this is working, we'll need to update the top file comment in [`packages/package-version-parser/package-version-parser.js`](https://github.com/meteor/meteor/blob/46fddec786e3c8561ac9963fdb54ee3c9f0ddeea/packages/package-version-parser/package-version-parser.js#L1-L5) to reflect the new (non-symlink) approach.

Issue #6002, opened 2016-01-14
Support .meteorignore files to restrict which files can be compiled or imported by Meteor apps or packages


benjamn:
**_1 Upvote**_ This would help developers avoid issues like https://github.com/meteor/meteor/issues/5999, among other potential benefits.

The syntax should be the same as `.gitignore` or `.npmignore` files.

cc @avital

avital:
This, or something similar should resolve the following error we get when running `./meteor test-packages`:

```
[[[[[ Tests ]]]]]

=> Started proxy.
=> Started MongoDB.
=> Errors prevented startup:

   While processing files with templating (for target web.browser):
   packages/local-test:spacebars-tests/assets/markdown_basic.html:1: Expected one of: <body>, <head>, <template>
   packages/local-test:spacebars-tests/assets/markdown_each1.html:1: Expected one of: <body>, <head>, <template>
   packages/local-test:spacebars-tests/assets/markdown_each2.html:1: Expected one of: <body>, <head>, <template>
   packages/local-test:spacebars-tests/assets/markdown_if1.html:1: Expected one of: <body>, <head>, <template>
   packages/local-test:spacebars-tests/assets/markdown_if2.html:1: Expected one of: <body>, <head>, <template>

=> Your application has errors. Waiting for file change.
```

dagatsoin:
Could this prevent Meteor to continuous building the app when working with Webstorm? (I think the .idea folder triggers the build at each keystrock.)

avital:
@dagatsoin probably!

acidicX:
Yes please!
This has been an issue [since 2012](https://github.com/meteor/meteor/issues/330) by the way.

cunneen:
Can the `.meteorignore` file please include a setting to disable the aggressive parsing of files (js, css and  html) that are not explicitly included in the package.js of a local package?

(ref @avital's comments on [this thread](https://forums.meteor.com/t/help-us-test-build-times-in-meteor-1-3/15031/17)

mjmasn:
@benjamn this is becoming fairly critical for our app as we move to using es6 modules / the imports directory. Meteor is trying to compile HTML generated by Istanbul test coverage reports. Temporary workaround I've found is to set the coverage directory in .istanbul.yml as follows (html inside node_modules directory appears to be ignored):

```yaml
reporting:
  dir: "node_modules/__coverage"
```

pletinsky:
[...]

PR #8921, opened 2017-07-17
[Work in progress] .meteorignore specifies intentionally untracked files and directories. 

i4han:
I wrote this to solve 
https://github.com/meteor/meteor/issues/6002
https://github.com/meteor/meteor-feature-requests/issues/5
Support .meteorignore files to restrict which files can be compiled or imported by Meteor apps or packages. 

The syntax is like .gitignore. The only ineffective pattern is '!' because it has to go with logical *AND*. On the other hand, Meteor's Isobuild ignoreFiles contains list of regular expression with logical *OR*.

I also copied glob-to-regexp to convert glob pattern to the regular expression. The file is from https://github.com/fitzgen/glob-to-regexp. I added the original writer's notice on the top of the code. Please let me know if it is an acceptable policy.

I need to have this feature so I can make my gulpfile.js work and make Meteor ignore my 'src' directory.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -172,6 +172,7 @@ var release = require('../packaging/release.js');
 import { load as loadIsopacket } from '../tool-env/isopackets.js';
 import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 import { gzipSync } from "zlib";
+import globToRegexp from './glob-to-regexp'; // .meteorignore glob pattern to regexp
 
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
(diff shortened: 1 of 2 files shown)


tools/isobuild/bundler.js
```diff
@@ -184,6 +185,62 @@ exports.ignoreFiles = [
...
+    known issue:
+      ignoreFiles take regExp list with default logical *or*, 
+      instead of glob pattern list with optional logical *and* for negative pattern.
+        *        # ignore all files
+        !*.html  # but (AND) not ignore *.html, ignore files that matches * AND !*.html
+        !*.css   # (AND) not ignore *.css
+        !*.js    # (AND) not ignore *.js
+*/
+
+if (files.exists(METEOR_IGNORE)) {
```
benjamn: Taking inspiration from how `.gitignore` files work, I think it's important to allow multiple `.meteorignore` files in different directories, each governing the files contained by its own directory tree.

It's also really important that we allow `.meteorignore` files to be added/moved/removed while the development server is running, so it's not enough to check the file once, as you're doing here.

I would strongly encourage you to consider implementing these changes in the [`PackageSource#_findSources`](https://github.com/i4han/meteor/blob/9a0b09b2159a190b3a8241620493e4426ea3b089/tools/isobuild/package-source.js#L984) method, instead of doing it at the top-level here.

i4han: @hwillson, Thank you for your comment. I will do that.
@benjamin I tried to solve my own problem to make it simply work but I understand your point and I will take a look at it.

tools/isobuild/bundler.js
```diff
@@ -184,6 +185,62 @@ exports.ignoreFiles = [
...
+  var dotIgnoreFiles = files.readFile(METEOR_IGNORE).toString('utf8')
+  exports.ignoreFiles = exports.ignoreFiles.concat( // wanted use array spread operator .push(...dotIgnoreFiles.split( maybe from v1.6
+    dotIgnoreFiles
```
levithomason: It may be wise to consider use of the [`ignore`](https://www.npmjs.com/package/ignore) module for this.  It is well tested on all platforms and will save the headache of manually parsing the file your self.

i4han: @levithomason It is a different way to do it. When I use 'glob-to-regexp' module I can use packageSource filter that uses regexp. If ignore module is used, I have to rewrite the filter.
Is it a consensus to replace packageSource filter from regexp with include, exclude rules in a array to glob expressions and rewrite the filter? I think that's fundamentally logical way to do it but the method should not break anything that has been there.
It is a very different scope I originally assumed. I tried to use gulp and a method that can be fixed in an hour solution. When I have extra time for a weekend, I can look into @benjamn's comment and rewrite filter with 'ignore' module. Do you guys need more discussion about it?

levithomason: Take my comment lightly as I am not a contributor here.  My intent is just to encourage meteor to support the community standard features around ignoring files.  If that is already done by your filter, then I'm very happy.

No further concerns here :)

i4han: @levithomason Oh. I see. No problem. 
Actually, using `ignore` and rewrite the filter make sense to me. The current approach to the use of regexp to filter files I have to convert from .ignore's glob expression to regexp. So, I have to write every exception in that case so bugs and missing features can occur. Those are exactly what `ignore` did already. So using `ignore` and rewrite the filter to be 100% compatible with .gitignore's glob pattern is the better to prevent bugs or minor incompatibility. Just need to rewire more code. In that case, I need to be sure that I don't break anything I don't see because I just briefly read the part of Meteor code to implement the feature I wanted to see. 

The bottom line is that I am really looking forward to seeing this feature to be included so I will look into it. What do you think @benjamn? Should the filter be written from regexp to glob expression?

i4han:
I check Meteor's fs moduls and I guess I can get into `fs.readdir()` in files.js `readDirectory()` in watch.js to bypass the Meteor's rule filter to rule .meteorignore expressions. In this case Meteor can not read any file at all if the file is on the .meteorignore expression. It is what it supposes to be. Is this acceptable method?

hwillson:
I'll close this PR since a new version of this was submitted in https://github.com/meteor/meteor/pull/8940. Thanks!

Issue #8897, opened 2017-07-12
test-packages with --test-app-path does not work anymore


mitar:
It seems `--test-app-path` does not work anymore. I am trying to run in my Meteor 1.5 app:

```
meteor test-packages --show-test-app-path --test-app-path /tmp/meteor-test-packages <package name>
```

And I get the following error:

```
Error: ENOENT: no such file or directory, open '/tmp/meteor-test-packages/package.json'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.writeFileSync (fs.js:1156:15)
    at wrapper (/tools/fs/files.js:1586:35)
    at Object.install (/tools/cli/default-npm-deps.js:20:5)
    at doTestCommand (/tools/cli/commands.js:1599:38)
    at Command.func (/tools/cli/commands.js:1546:10)
    at /tools/cli/main.js:1483:23
```

hwillson:
Looking at the code, as long as you've created the directory ahead of time, your command should work. I don't see any code in place to create this directory if it doesn't exist, or throw an error when it doesn't, so this should be addressed.

mitar:
Oh, maybe I created a dir in the past and I didn't now. I assumed that it will be created automatically.

hwillson:
Yes, it should be handled a bit better - I'll patch things to create the directory first, if it doesn't exist.

hwillson:
PR #8923 was merged, so closing here. Thanks!

PR #8923, opened 2017-07-17
Adjust test-packages --test-app-path to create the directory if missing

hwillson:
Hi all - this PR is intended to fix https://github.com/meteor/meteor/issues/8897. The `test-packages` command `--test-app-path` parameter expects the specified directory to already exist. If it doesn't, an error is thrown. The changes in this PR create the specified directory if it doesn't already exist (and adds a few extra safeguards, like handling user permission errors, cases where the `test-app-path` already exists as a file instead of a directory, etc.). It also adds `self-test`'s for `--test-app-path` functionality. Thanks!

cc @mitar

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1579,8 +1579,31 @@ function doTestCommand(options) {
   // cleaned up on process exit. Using a temporary app dir means that we can
   // run multiple "test-packages" commands in parallel without them stomping
   // on each other.
-  var testRunnerAppDir =
-        options['test-app-path'] || files.mkdtemp('meteor-test-run');
+  let testRunnerAppDir;
+  const testAppPath = options['test-app-path'];
+  if (testAppPath) {
+    try {
+      if (files.mkdir_p(testAppPath, 0o700)) {
+        testRunnerAppDir = testAppPath;
+      } else {
+        Console.error(
+          'The specified --test-app-path directory could not be used, as ' +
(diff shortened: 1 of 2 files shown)


tools/cli/commands.js
```diff
@@ -1579,8 +1579,31 @@ function doTestCommand(options) {
   // cleaned up on process exit. Using a temporary app dir means that we can
   // run multiple "test-packages" commands in parallel without them stomping
   // on each other.
-  var testRunnerAppDir =
-        options['test-app-path'] || files.mkdtemp('meteor-test-run');
+  let testRunnerAppDir;
+  const testAppPath = options['test-app-path'];
+  if (testAppPath) {
+    try {
+      if (files.mkdir_p(testAppPath, 0o700)) {
```
mitar: Is there really a need to specify umask? Wouldn't environment's umask apply if somebody wants to configure this? I have not seen elsewhere in Meteor that we would specify this.

mitar:
Looks great! Thanks!

tools/cli/commands.js
hwillson: Thanks for reviewing @mitar - I did this as `files.mkdtemp` (which we're essentially replacing when a `--test-app-path` value is passed in) [is doing this](https://github.com/meteor/meteor/blob/44796a183111ed18fb27518780bf98267d04e7d5/tools/fs/files.js#L666).

PR #8926, opened 2017-07-19
Update LICENSE

zredlined:
Followed license structure defined: https://en.wikipedia.org/wiki/MIT_License with formatting for copyright holder. Updated license to remove "Meteor Development Group" which had been copied over from Meteor License that was previously used as template, replaced with Petar Korponaic as author.

Format:
Copyright (c) <year> <copyright holders>


hwillson:
If we merge this, does that mean Petar Korponaić will be on the hook to get all outstanding issues fixed? If so, we might consider it ... 🙂

PR #8940, opened 2017-07-22
 .meteorignore specifies intentionally untracked files and directories. #8921 version 2

i4han:
The second version of https://github.com/meteor/meteor/pull/8921  pull request.

Features:
- `ignore` module is used. It supports https://git-scm.com/docs/gitignore
- be located before `include`, `exclude` regexp filter instead of using it.
- used cache for faster load. 
- works with `autoupdate` feature. Meteor server will be automatically restarted when .meteorignore is changed. (See known issue 2)
- skips `node_modules` and `imports` and its subdirectories. (Is there any reason to be included?)
- allows multiple `.meteorignore` files in different directories, 
   each governing the files contained by its own directory.

Known issues: 
1. single character ? and range of character [a-z] are not supported. (not in the git man page)
2. Update rules for subdirectories in `.meteorignore` (like client/*.ts in the app directory) 
  will not activate `autoupdate` while the server is running. You need to restart meteor for the case.
  If you want `autoupdate` `.meteorignore` rules for its subdirectories while the server is running.
  you need to write the directory's own `.meteorignore` in the subdirectories or restart manually.

### DIFF
--- a/tools/fs/watch.js
+++ b/tools/fs/watch.js
@@ -4,6 +4,7 @@ import * as safeWatcher from './safe-watcher.js';
 import {createHash} from "crypto";
 import {coalesce} from '../utils/func-utils.js';
 import {Profile} from '../tool-env/profile.js';
+import ignore from 'ignore';
 
 import {
   optimisticStatOrNull,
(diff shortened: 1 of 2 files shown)


i4han:
The Travis CI build failed because of the `dev_bundle` change to add `ignore' "3.3.3". @hwillson does it look ok?

hwillson:
Hi @i4han - thanks for submitting this new version. Don't worry about the test failures yet (as you'll need a new version of the `dev_bundle` published, but we'll do a review first). We'll be discussing this in our issue review meeting today, and will get back to you afterwards. Thanks!

tools/fs/watch.js
```diff
@@ -265,6 +266,93 @@ export function sha1(...args) {
...
+    absPath.slice(0, appDir.length) === appDir  // check if current path is in the appDir
+  ) { 
+    var ig = filterMeteorIgnore(absPath)
+    if (ig) {
+      contents = ig.filter(           // contents are a list of basenames (no path just filename). 
+        contents.map(v =>             // In order to be used by ignore filter
+          files.pathJoin(absPath, v)  // path from the *appDir* should be added to the contents list.  
+            .slice(appDir.length + 1) // A full path (not from appDir) is given, it should be cut like 'client/'
+        )                             // Otherwise it will not be filtered by 'client/' in .meteorignore.
+      ).map(v => v.split('/').slice(-1)[0]); // This line is to make path.basename. Back to list of basename.
```
hwillson: Can you keep lines wrapped at 80 chars? Also - comments can be helpful, but I think in this case an end of line comment on each line is a bit excessive and detracts from code readability. Could you maybe summarize these comments a bit more and move them above the `filter` / `map` calls?

hwillson:
Hi @i4han - having reviewed your proposed implementation, and after discussing the implications of this approach (with regards to performance and future Tool design) with the team, we've decided to close this PR for now. It has become clear that implementing `.meteorignore` functionality in the least intrusive and most performant way possible will require much more guidance from the Meteor team. Adding `.meteorignore` functionality will most likely be better suited in [`PackageSource#_findSources`](https://github.com/meteor/meteor/blob/4665f2aab3a9344a9fc5e17d31e57a70f3da05e4/tools/isobuild/package-source.js#L984), but making changes to this part of the codebase has to be done very carefully, due to the performance sensitive nature of the Isobuild process. We'll need to think about the best approach for adding this functionality much more thoroughly, and will update https://github.com/meteor/meteor-feature-requests/issues/5 with a more concrete plan outlining how we think this should be implemented.

Thank you very much for the time and effort you've put into working on this @i4han - it is greatly appreciated! While this PR didn't get accepted, your upfront work on this has been invaluable, and will help steer the future direction of this functionality. Thanks again!

i4han:
Hi @hwillson,
Thank you for the detailed and sensible comments. This is my first pull request and learned many things. It was enjoyable to look into how meteor tool handles files. I just made my local meteor fork to work the way I wanted without an understanding of how meteor tool handles files so that's why my each patch positioned at the end point of the process. 
Yes, I see and agree meteor should run on mission critical projects and PackageSource is the logical spot for .meteorignore.  I was not confident to hand on PackageSource and follow the rule of how meteor tool at the moment but I see better now.
I appreciate your warm and welcome comments, it was an awesome experience.

PR #8941, opened 2017-07-23
Avoid adjusting line counts when removing source map comments.

### LINKED ISSUE #8927: Typescript srcmaps/stacktraces broken with meteor 1.5-beta.3 onwards
I'm raising this issue in the main meteor/meteor repo as I don't think it is Typescript specific, and that it was a change in the bundler that broke things/exposed the problem.

The essence of the issue is that srcmaps and stacktraces are out, sometimes by a small amount but often by much more making it impossible to debug typescript apps or even rely on server side stack traces.

I have a github repo which is simply a `meteor create --example angular2-boilerplate` followed by addition of a trivial file, and a reference to that file that triggers exceptions on startup.
See https://github.com/sdarnell/srcmap-bug

The line numbers in the console output are wrong with meteor 1.5 (specifically meteor 1.5-beta.3).  The debug.ts file is simply:
```
1: console.trace('BaBoom');
2: export class Debug {
3:    public static Bang() {
4:        throw new Error("Boom!");
5:    }
6: }
7:
```
With Meteor 1.4.2.3, 1.4.3, 1.4.4, 1.4.4.1, 1.4.4.2, 1.4.4.3, 1.5-beta.2 the file and line numbers are:
 - BaBoom: at meteorInstall.both.debug.js (`both/debug.ts:1:9`)
 - Boom!: at Function.Debug.Bang (`both/debug.ts:4:15`)

In Meteor 1.5-beta3, 1.5 and 1.5.1 they are:
 - BaBoom: at meteorInstall.both.debug.js (`both/collections/demo.collection.ts:4:87`)
 - Boom!: at Function.Debug.Bang (`both/debug.ts:6:2`)

In the first case the file name is also wrong, and the second case the line number is out by 2 (and the column is wrong).

By means of binary chopping in on the problem, it appears that it is https://github.com/meteor/meteor/commit/e469eeec0537bd1f019a0d3cf242013830379c84 which caused the issue (just before 1.5-beta.3 was released).

I wasn't expecting it, but if I undo this part of the change that removes srcmaps things work again:
```diff
-    .replace(/\/\/# sourceMappingURL=[^\n]+/g, '//');
+    .replace(/\n\/\/# sourceMappingURL=[^\n]+/g, "");
```
The change clearly affects the number of physical lines, and presumably this is perturbing things, so unless the ultimate app.js.map is generated to understand that the additional lines are removed, it will be out by 1 line per file included in that specific bundle of js files.

@benjamn do you have any thoughts on why this only affects typescript plugin generated code, and not general JS or ecmascript converted code?

If we stop removing the line, will it knock out of step srcmaps elsewhere if compensating code has been added.

Originally derived from https://github.com/meteor/meteor/issues/8611 and split off into https://github.com/barbatus/typescript/issues/56.

sdarnell:
Fixes #8927 

The problem was introduced in 1.5-beta.3 and results in source maps being out by one line for each file containing source maps (e.g. produced by Typescript compiler).

The change keeps the line rather than removing it, but unlike the original change (before 1.5-beta.3) it doesn't add add the slash slash comment (//) as I can't see where that would be needed (other than in a multi-line back tick string - but that case is broken anyway).

It does retain tying the match to the start of a line (introduced in 1.5-beta.3), though I did consider whether the expression would be simpler using multi-line mode:  `/^\/\/# sourceMappingURL=.+/gm`

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -2540,7 +2540,7 @@ function addSourceMappingURL(data, url) {
     // If data is a Buffer, convert it to a string.
     .toString("utf8")
     // Remove any existing source map comments.
-    .replace(/\n\/\/# sourceMappingURL=[^\n]+/g, "");
+    .replace(/\n\/\/# sourceMappingURL=[^\n]+/g, '\n');
   // Append the new source map comment to the end of the code.
   return dataString + "\n//# sourceMappingURL=" + url + "\n";
 }


benjamn:
Thanks for identifying the root cause of this problem and fixing it, @sdarnell!

benjamn:
This fix is in the latest 1.5.2 beta if you want to try it out.

PR #8953, opened 2017-07-27
Make minified code run on older WebKit platforms

### LINKED ISSUE #8928: Unable to run any Meteor app under older WebKit via `meteor build` or via `meteor --production`: throws a JS error
iPad 1G reports an error on line :55 within the main .js bundle for a sample app created via `meteor create simple-todos`. The error description is: "SyntaxError: Parse error". Same happens with my other apps, just the line number varies, same block of code though.

This is the latest Meteor of version 1.5.1, and the error does not get thrown when ran in development mode via `meteor`.

[[the link to pastbin with the code from line 55](https://pastebin.com/Y3xn93LX)] I... assume it could be those Unicode symbols within the code?

upd: tried running `simple-todos` via `meteor --production --release 1.5`; same result, same line 55. Happens only on iPad 1G (iPhone 5 works fine, so it's not an iOS issue, at least not with the latest one).

upd 2: changed babel to version `^6.23.0` from `^6.20.0` in `package.json` -- no go, still error on line 55 within .js when used via iPad 1G.

upd 3: so it seems like there's a byte sequence within that line which mirrors the line horizontally (both on pastebin.com via Chromium and within my desktop IDE)... so e.g. word `constructor` gets rendered as `rotcurtsnoc`, yet I'm able to find by searching for "const". Apparently that's the byte which messes up the old iOS' JS engine. I have no other explanation for the bug that I found.
I think I can say that I've seen too much in my life at this point.

upd 4: this seems to be the uncompressed chunk of code that causes the issue: https://github.com/meteor/meteor/blob/87681c8f166641c6c3e34958032a5a070aa2d11a/packages/ddp-client/sockjs-0.3.4.js#L407
Please note how there are no unicode chars within the original (L407), only within the minified version (pastebin). The minification does something to that code, so old JS engines trip on it and refuse to compile.

upd 5: 
I can confirm that removing package `standard-minifier-js@2.1.1` from .meteor/packages solves the issue, yet it ends up leaving my app uncompressed (therefore unsuitable for being deployed to production).

upd 6:
The issues seems to be triggered by the right-to-left unicode control character mark, here's an example (try selecting the code within the highlight): `<span class="st0">"‮"</span>`

vflyson:
Resolves #8928

Minifying sockjs.js without this flag results in all Meteor apps being broken on older platforms such as iPad 1G. There's a thorough investigation report within the link to the corresponding issue that I had to do to find out the reason for my apps being unusable on certain devices after being bundled.

This is a very important bug fix, please do not close it without reading the issue that I've provided as a reference.

### DIFF
--- a/packages/minifier-js/minifier.js
+++ b/packages/minifier-js/minifier.js
@@ -12,6 +12,7 @@ meteorJsMinify = function (source) {
         drop_debugger: false,
         unused: false,
         dead_code: false,
+        collapse_vars: false,
         global_defs: {
           "process.env.NODE_ENV": NODE_ENV
         }


vflyson:
Sign an agreement, seriously? What happened to "free and open source"? I'm going to the react.js community guys, I'm done with space travel.

rosser2121:
Are you sure?

abernix:
We truly appreciate you trying to make a contribution and certainly appreciate the investigative work you did to figure this.  However, just so you understand how the open-source ecosystem works, if you decide to make a contribution to React (a Facebook Open Source project), @facebook-github-bot will make you sign [the _exact_ same CLA terms](https://code.facebook.com/cla/individual) to protect their own existence and avoid patent and copyright claims over code submitted to the project.

You may be onto something here, but I'll reply in the existing issue.

vflyson:
I admit I got pissed off there, but this kind of bureaucracy is opposite to what I believe in when it comes to software. Something's very broken with the patent/copyright laws if one needs to put their real name and reveal their identity in order to contribute code. One could easily use a fake one and copy code from React.js into Meteor without anyone noticing, and then make Facebook sue MDG for stealing their code? It looks more like a bottleneck barring developers from contributing rather than a way to protect the project legally.

green-coder:
That would be nice if the patch could be merged. Don't let bureaucracy block programmers.

PR #8956, opened 2017-07-28
Include the Node.js and npm version in the `star.json` manifest.

abernix:
This makes it possible to know exactly which version of Node.js and npm were used by the `meteor` tool which the bundle was built from by making new values available alongside the existing `meteorRelease` value in `star.json`: `nodeVersion` and `npmVersion`.  The version of the `star.json` has remained the same since this is only adding new properties, not removing any existing.  This preserves the existing `.node_version.txt` for compatibility reasons, but it's reasonable to believe that may be deprecated in a future version.

Those using `.node_version.txt` should consider using the `nodeVersion` value in `star.json`, though it should be noted that it does not include the `v`(ersion) prefix as the `.node_version.txt` file did.

### DIFF
--- a/tools/isobuild/meteor-npm.js
+++ b/tools/isobuild/meteor-npm.js
@@ -14,6 +14,7 @@ var buildmessage = require('../utils/buildmessage.js');
 var utils = require('../utils/utils.js');
 var runLog = require('../runners/run-log.js');
 var Profile = require('../tool-env/profile.js').Profile;
+import { version as npmVersion } from 'npm';
 import { execFileAsync } from "../utils/processes.js";
 import {
   get as getRebuildArgs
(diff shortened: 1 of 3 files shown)


abernix:
I'll add a `History.md` entry, after this passes CI.

tools/isobuild/meteor-npm.js
```diff
@@ -14,6 +14,7 @@ var buildmessage = require('../utils/buildmessage.js');
 var utils = require('../utils/utils.js');
 var runLog = require('../runners/run-log.js');
 var Profile = require('../tool-env/profile.js').Profile;
+import { version as npmVersion } from 'npm';
```
benjamn: This can be `export { version as npmVersion } from "npm"`, and then you don't need to export it below!

abernix: I do prefer this syntax!  But opted for this technique to make the `meteorNpm` API feel more whole (as in, it seemed like the `meteorNpm` api should have a `npmVersion` exposed).

PR #8963, opened 2017-07-29
support env in babelrc

hexsprite:
This PR implements support for the env key in `.babelrc` files.

It looks first for `BABEL_ENV` then `NODE_ENV` then finally defaults to `development` per the Babel documented behavior.

Rationale: be able to add additional babel plugins/presets for testing environment for example (eg. `babel-plugin-istanbul`)

I wasn't sure where to add tests as I did not see any existing tests for the functionality.  Please let me know your feedback.

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -300,6 +300,16 @@ BCp._inferHelper = function (
   merge(babelOptions, babelrc, "presets");
   merge(babelOptions, babelrc, "plugins");
 
+  const babelEnv = (process.env.BABEL_ENV ||
+                    process.env_NODE_ENV ||
+                    "development");
+  if (babelrc && babelrc.env && babelrc.env[babelEnv]) {
+    const env = babelrc.env[babelEnv];
+    walkBabelRC(env);
+    merge(babelOptions, env, "presets");
+    merge(babelOptions, env, "plugins");
+  }
+
   return !! (babelrc.presets ||
              babelrc.plugins);
 };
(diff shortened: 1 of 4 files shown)


packages/babel-compiler/babel-compiler.js
```diff
@@ -300,6 +300,16 @@ BCp._inferHelper = function (
   merge(babelOptions, babelrc, "presets");
   merge(babelOptions, babelrc, "plugins");
 
+  const babelEnv = (process.env.BABEL_ENV ||
+                    process.env_NODE_ENV ||
```
benjamn: Should this be `process.env.NODE_ENV`?

benjamn:
> I wasn't sure where to add tests as I did not see any existing tests for the functionality. Please let me know your feedback.

I would recommend adding `env` to some of the `.babelrc` files in [the modules test app](https://github.com/meteor/meteor/tree/devel/tools/tests/apps/modules), or creating new local packages with new `.babelrc` files in that app.

packages/babel-compiler/babel-compiler.js
hexsprite: I believe so. From the babel docs:

`The env key will be taken from process.env.BABEL_ENV, when this is not available then it uses process.env.NODE_ENV if even that is not available then it defaults to "development".`

https://babeljs.io/docs/usage/babelrc/#env-option

hwillson: @hexsprite Ben's referring to the typo in your `process.env_NODE_ENV` reference (note the `_`).

hexsprite: Oh, woops. Got it.

PR #8971, opened 2017-07-31
Eliminate underscore usage from the `meteor` package.

benjamn:
Because the `meteor` package [depends on `underscore`](https://github.com/meteor/meteor/blob/96db56b0bae5168fa852eead7cbf0131a12c0ec8/packages/meteor/package.js#L18), and every package implicitly depends on the `meteor` package, the `underscore` package (in its entirety) is typically the first package to load in any Meteor application. This PR removes that dependency, so packages that do not depend on `underscore` can load a little sooner, and the Meteor core will be that much closer to not depending on `underscore` at all.

This PR is best viewed with `?w=1` to ignore whitespace-only changes: https://github.com/meteor/meteor/pull/8971/files?w=1

### DIFF
--- a/packages/meteor/url_common.js
+++ b/packages/meteor/url_common.js
@@ -14,7 +14,7 @@ Meteor.absoluteUrl = function (path, options) {
     path = undefined;
   }
   // merge options with defaults
-  options = _.extend({}, Meteor.absoluteUrl.defaultOptions, options || {});
+  options = Object.assign({}, Meteor.absoluteUrl.defaultOptions, options || {});
 
   var url = options.rootUrl;
   if (!url)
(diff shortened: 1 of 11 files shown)


packages/meteor/url_common.js
```diff
@@ -14,7 +14,7 @@ Meteor.absoluteUrl = function (path, options) {
     path = undefined;
   }
   // merge options with defaults
-  options = _.extend({}, Meteor.absoluteUrl.defaultOptions, options || {});
+  options = Object.assign({}, Meteor.absoluteUrl.defaultOptions, options || {});
```
benjamn: I think using `Object.assign` is safe here because the `es5-shim` package should already have had the chance to load by the time `Meteor.absoluteUrl` is called.

benjamn:
I don't think this really requires a bullet in `History.md`, though perhaps it will be worth calling out in a blog post.

benjamn:
Merged to `devel` as 7f062192e1f9e4cb669d378b1140da9a33b0db42, and published as part of `meteor@1.7.1`.

PR #8972, opened 2017-07-31
Run all "bare" package files before requiring eager entry point modules.

benjamn:
Immediately after each Meteor package is evaluated, we synchronously `require` its eager entry point modules (typically those identified by `api.mainModule`), before loading other packages. This behavior makes the load order of packages more important than it should be, leading to bugs like #8969, where one package synchronously imports another package that hasn't loaded yet, and gets weird/`undefined` results.

The natural solution here is to run all eager `require` calls after loading all package code. However, we have avoided making that change because of an obscure feature that allows package files to be `bare`, which means they are not wrapped as modules, and thus may declare variables directly in the package scope. Historically, `bare` code could run in between eager `require` calls, according to the order of `api.addFiles` calls in the `package.js` file.

If we care about the intentional interleaving of `bare` code with `require` calls, then calling `require` later is not possible without breaking backwards compatibility, since we can't defer the `bare` code without putting it a different scope. However, it appears that no Meteor core packages depend on `bare`/`require` interleaving, and I suspect very few existing Meteor packages will be broken by this change.

Specifically, this PR just runs all `bare` code in a given package before calling `require` to evaluate the package's eager entry point module(s). If this works, then we're one step closer to deferring all `require` calls until every package has finished evaluating, which will fix bugs like #8969, as well as making it unnecessary to move files from `node_modules` into the `modules` package, which has been a source of some confusion ever since Meteor 1.3.

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -398,6 +398,9 @@ _.extend(Module.prototype, {
     // Now that we have installed everything in this package or
     // application, immediately require the non-lazy modules and
     // evaluate the bare files.
+
+    const eagerModuleFiles = [];
+
     _.each(this.files, file => {
       if (file.bare) {
         chunks.push("\n", file.getPrelinkedOutput({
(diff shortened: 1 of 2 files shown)


benjamn:
Note to self: add a note about this to `History.md`.

tools/isobuild/linker.js
```diff
@@ -398,13 +398,22 @@ _.extend(Module.prototype, {
     // Now that we have installed everything in this package or
     // application, immediately require the non-lazy modules and
     // evaluate the bare files.
```
hwillson: Should we change this comment around a bit, maybe to something like?

```
// Now that we have installed everything in this package or
// application, first evaluate the bare files, then require the 
// non-lazy modules.
```

PR #8973, opened 2017-08-01
Avoid new uses of 'package' as it is apparently a reserved keyword.

sdarnell:
@klaussner spotted that I'd added more uses of the 'package' reserved word in https://github.com/meteor/meteor/commit/96db56b0bae5168fa852eead7cbf0131a12c0ec8
Whilst it doesn't appear to cause a problem, and there are other uses in the constraint solver, this change removes the new uses.  Thanks @klaussner.

### DIFF
--- a/tools/cli/commands-packages.js
+++ b/tools/cli/commands-packages.js
@@ -1172,9 +1172,9 @@ main.registerCommand({
     const dontExpand = new Set(topLevelSet.values());
 
     // Recursive function that outputs each package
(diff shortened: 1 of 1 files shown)


klaussner:
[Reserved words can be used as keys, even without quotes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Reserved_word_usage), so the rest of the constraint solver should be fine.

PR #8976, opened 2017-08-02
Auto-install Cordova upon use rather than building into the dev bundle.

abernix:
**Note:** This is targeting `release-1.6`, though it's reasonable to believe this could go into a `release-1.5.x` branch (currently, `release-1.5.2`).

---

This changes `cordova-lib` (and its rather large dependency tree) to no longer be built into the initially downloaded dev bundle by `generate-dev-bundle.sh`, but rather to be downloaded and installed automatically upon execution of commands which utilize Cordova functionality, namely `meteor add-platform <target>`, `meteor remove-platform <target>` and `meteor run <target>`.

This falls short of completely decoupling Cordova from Meteor (as https://github.com/meteor/meteor-feature-requests/issues/112 is requesting), but by removing it from the initial dev bundle we reduce the download size of the installation tarball by about 35MB (a ~38% decrease).

This PR also includes [this fix](https://github.com/meteor/meteor/commit/043e2922f6ef8ee49e32ef37a81704b2d08730fc) for a misspelled parameter which was breaking Cordova on the `release-1.6` branch (which I was testing on).

This might still be "work-in-progress", but I submit it for consideration. 😉

PR #8981, opened 2017-08-03
Auto-install `meteor self-test` dependencies upon use.

abernix:
**Note:** Targeted for `release-1.6`, because of the dependency on #8976.

---

This uses the same new facilities which were created for auto-installing Cordova (#8976) to also auto-install PhantomJS and BrowserStack WebDriver npms into their appropriate home in the dev bundle when they're needed for running self-tests.

Gloriously, this means if you're constantly changing between dev bundles, you'll no longer need to take the extra step of doing so each time. 🙌

### DIFF
--- a/tools/tool-testing/selftest.js
+++ b/tools/tool-testing/selftest.js
@@ -27,27 +27,14 @@ var release = require('../packaging/release.js');
 var projectContextModule = require('../project-context.js');
 var upgraders = require('../upgraders.js');
 
+import { ensureDependencies } from '../cli/dev-bundle-helpers.js';
+
+const DEV_DEPENDENCY_VERSIONS = {
+  'phantomjs-prebuilt': '2.1.14',
+  'browserstack-webdriver': '2.41.1',
+};
+
 require("../tool-env/install-runtime.js");
-
-function checkTestOnlyDependency(name) {
-  try {
-    var absPath = require.resolve(name);
-  } catch (e) {
-    throw new Error([
-      "Please install " + name + " by running the following command:",
-      "",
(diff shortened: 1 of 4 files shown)


tools/tool-testing/selftest.js
```diff
@@ -1025,6 +1012,19 @@ var PhantomClient = function (options) {
...
+      title: 'Installing PhantomJS in Meteor tool',
+    },
+    () => {
+      ensureDependencies({
+        'phantomjs-prebuilt': DEV_DEPENDENCY_VERSIONS['phantomjs-prebuilt'],
+      });
+    }
+  );
+
+  self.npm = require("phantomjs-prebuilt");
```
benjamn: Can we call this something more descriptive, like `self.npmPackageExports`?

tools/tool-testing/selftest.js
```diff
@@ -27,27 +27,14 @@ var release = require('../packaging/release.js');
...
-      ""
-    ].join("\n"));
-  }
+import { ensureDependencies } from '../cli/dev-bundle-helpers.js';
 
-  return require(absPath);
-}
+const DEV_DEPENDENCY_VERSIONS = {
+  'phantomjs-prebuilt': '2.1.14',
+  'browserstack-webdriver': '2.41.1',
```
benjamn: Pinning these versions is a big win for reproducibility, compared to the current approach of always installing the latest versions.

tools/tool-testing/selftest.js
abernix: Done!

PR #8985, opened 2017-08-04
Switch to using CircleCI 2.0 and various other test improvements.

abernix:
This switches Meteor's CircleCI builds from Circle 1.0 to Circle 2.0 which has a bit more control over the workflow.

Currently, this eliminates the existing ci.sh script which was already a bit incompatible when I was attempting to run Windows builds on another environment. It's possible that we should change this to a Node.js wrapper script.

Other improvements:

- We now store Core Dumps in build artifacts.  CircleCI 2.0 advertised this as one of the features of CircleCI 2.0, but honestly, it was far from straightforward.  Perhaps if we were using another Dockerimage, but it was far from as easy as flipping a switch.  In addition to saving the Core Dump, this also saves the Node.js binary which was included in the Dev Bundle.  This can be very handy for post-mortem debugging with tools like lldb, gdb, or mdb.

- Memory usage is now logged throughout the build via a background process which logs `ps` output to a file which is persisted to the build artifacts.  This should help identify if builds are terminating for some environmental reason.

- Thanks to CircleCI 2.0 workflows, it's now possible to re-run just the failed job within the workspace (i.e. the commit that triggered it all) instead of re-running the entire workspace.  This means that one failed container won't take up a full eight containers on a re-run, just one.

---
It's probably worth noting that when browsing CircleCI, it may be easiest to use the "[Workflows](https://circleci.com/gh/meteor/workflows/meteor)" button on the left side of CircleCI navigation to view the totality of the workflow progress, instead of "Builds" since builds will show individual jobs within the build.  It's then possible to click on the workflow and see the individual portions, in a similar fashion to how it works within Builds, granted the containers are no longer in a sorted order (i.e. they will show as "Group 7", "Group 4", "Group 0", etc. 🤷‍♂️ ; CircleCI 2.0 GUI is still WIP, me thinks.)


abernix:
Oh, wow!  I did not realize that the groups would show up separately in the PR!  Neat? Maybe I should change their names to be the same as the tests (e.g. `Running self-test (1): A-Com`)?

hwillson:
This is awesome @abernix! Very interesting about the groups showing - I like your idea of showing the test name grouping instead of the group number (or maybe in addition to the group number). And yes, I find their UI to be a bit of a mess. Having to look at all group builds combined under Workflows seems like it was added as an afterthought when they realized that functionality was missing ...

hwillson:
Hmm - one thing @abernix; I saw that group 6 failed, so I re-ran that individual group (which is awesome!). Strange thing is re-running the group failed again immediately saying it can't find `meteor`.

abernix:
@hwillson The interface was worse last week, so I'm hoping they iterate some more. 🤞 

The only issue with showing `Running self-test (1): A-Com` is that we'd have to use that name in multiple places in the `config.yml` – for example, it would need to be [here](https://github.com/abernix/meteor/blob/64efdef14b14e7b6190be550d14a676628cd2298/.circleci/config.yml#L182) and [here](https://github.com/abernix/meteor/blob/64efdef14b14e7b6190be550d14a676628cd2298/.circleci/config.yml#L393).

Also, about the individual job not working on re-run – yeah, you have to do it from the Workspace, using the "Rerun failed jobs" link in the dropdown, not the job itself, otherwise it doesn't re-attach the (mandatory, pre-requisite) workspace. 🤦‍♂️ .  Also hopeful about this.  Might be time to open some suggestions with their support.

I just used this button from the Workspaces and it's retrying *just* 6 now.

hwillson:
Wow, thanks for the heads up about the re-run; noted for the future, but hopefully they fix that! I see what you mean about the config - okay, maybe Group X is good enough; it's not hard to see which tests exactly failed by clicked on one link, so it's probably okay the way it is. I really do like the ability to re-run one group only! 

One last thing - are we still okay to use 2 more containers (groups)? Those changes don't have to be part of this of course, but I might look into rebalancing things in a separate PR, if that makes sense. Thanks!

abernix:
I think we should just go ahead and use four more containers (all we have!), but I think it'd be great if we used some specific timing information about each test and either auto-balance based on those timings instead of trying to do it manually with those regexes, not to mention that they are prone to missing tests (I'm fairly sure a new test which started with the letter `q` right now would not run)

CircleCI has some timing information, but unfortunately, it's per group right now which is not super helpful for us.  We'd need to run each test individually to figure out the best balance.  Haven't given this too much thought yet, but probably something we can iterate on top of this with.  Feel free to start by just adding the additional containers via regex modifications though! :smile:

hwillson:
Awesome, sounds great! I completely agree - let's make those modifications later and not hold this PR up. I'm anxious to start using these changes!

benjamn:
Being able to re-run individual groups is huge!

.circleci/config.yml
```diff
@@ -0,0 +1,413 @@
...
+        sudo tee -a /etc/sysctl.conf
+
+    # Note that since every "run" command starts its own shell, and I wasn't
+    # able to set this at a system wide level for all users, it's necessary to
+    # run "ulimit -c unlimited" before each command which you want to (possibly)
+    # output a core dump.
+
+    # Raise inotify user watches up higher.
+    echo fs.inotify.max_user_watches=524288 | \
+        sudo tee -a /etc/sysctl.conf
```
benjamn: This never really worked before, but I guess it doesn't hurt.

Issue #8947, opened 2017-07-26
Internal exception while processing message: TypeError: Cannot read property 'get' of undefined


cwohlman:
I created a new app using meteor 1.4.4.1: `meteor create wrong-ddp-package-version-reproduction --release 1.4.4.1`

When I tried to run the app I got this error (on both the client & the server):

```
I20170726-09:56:56.210(-5)? Internal exception while processing message { msg: 'connect',
I20170726-09:56:56.210(-5)?   session: 'C25WdMWFhMqtBW7BT',
I20170726-09:56:56.211(-5)?   version: '1',
I20170726-09:56:56.211(-5)?   support: [ '1', 'pre2', 'pre1' ] } Cannot read property 'get' of undefined TypeError: Cannot read property 'get' of undefined
I20170726-09:56:56.211(-5)?     at withoutInvocation (packages/meteor.js:443:33)
I20170726-09:56:56.212(-5)?     at bindAndCatch (packages/meteor.js:452:33)
I20170726-09:56:56.212(-5)?     at Object._.extend.setInterval (packages/meteor.js:479:24)
I20170726-09:56:56.213(-5)?     at [object Object]._.extend._startHeartbeatIntervalTimer (packages/ddp-common.js:89:44)
I20170726-09:56:56.213(-5)?     at [object Object]._.extend.start (packages/ddp-common.js:84:10)
I20170726-09:56:56.213(-5)?     at new Session (packages/ddp-server/livedata_server.js:318:20)
I20170726-09:56:56.213(-5)?     at [object Object]._.extend._handleConnect (packages/ddp-server/livedata_server.js:1463:29)
I20170726-09:56:56.214(-5)?     at packages/ddp-server/livedata_server.js:1379:18
```
Reproduction:
https://github.com/cwohlman/wrong-ddp-package-version-reproduction

Work around:
Correct the versions for the meteor and/or ddp packages in the .meteor/versions file (the bug is caused by a variable rename, as far as I can tell, causing a missmatch between the expected variable name and the actual variable name).

Other details:
- I'm running on Mac OS 10.12.5

crapthings:
we have the same issue while updating from METEOR@1.4.2.3

abernix:
This will go away if you update to Meteor 1.5.1.  You can also explicitly set your `meteor` (the package, not the tool) version pinning to `meteor@1.6.1`

[...]

PR #8991, opened 2017-08-07
Support @~ version constraints and use them for core packages.

benjamn:
As discovered in #8947, creating an app with
```sh
meteor create --release 1.4.4.1 new-app
```
does not actually constrain core packages to have the same minor versions they had when Meteor 1.4.4.1 was published. Rather (as of Meteor 1.4.3), the Meteor release only constrains core packages to have versions at least as recent as the versions they had when the release was published, and to have the same major version. Any later minor version is fair game.

What we need is a new kind of package version constraint, more like [semver's `~` range operator](https://www.npmjs.com/package/semver#tilde-ranges-123-12-1).

This commit implements that functionality, though it stops short of supporting all possible semver range syntax.

### DIFF
--- a/tools/tests/package-tests.js
+++ b/tools/tests/package-tests.js
@@ -1037,10 +1037,7 @@ selftest.define("tilde version constraints", [], function () {
 
   s.write(
     ".meteor/packages",
-    packages + [
-      "tilde-dependent",
-      "tilde-constraints@0.4.2"
-    ].join("\n")
+    packages + "tilde-dependent\n"
   );
   run.match(/tilde-constraints.*added, version 0\.4\.2/);
   run.match(/tilde-dependent.*added, version 0\.1\.0/);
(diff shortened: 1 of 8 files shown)


tools/tests/package-tests.js
```diff
@@ -955,3 +955,152 @@ selftest.define("show readme excerpt",  function () {
...
+  run.waitSecs(10);
+
+  setDepConstraint("0.4.3");
+  run.match("error: No version of tilde-constraints satisfies all constraints");
+  run.waitSecs(10);
+
+  setDepConstraint("~0.4.3");
+  run.match("error: No version of tilde-constraints satisfies all constraints");
+  run.waitSecs(10);
+  */
```
benjamn: It's disappointing that these tests don't work, but I think that's a general quirk of version constraints between local packages.

PR #9014, opened 2017-08-14
JUnit support for `meteor self-test` test runner.

abernix:
This started as a proof-of-concept, but I think it's worth submitting for consideration.

In an effort to get more fine-grained information out of the `meteor self-test` tool (which is used for testing of Meteor releases during development) and to take advantage of additional features in CircleCI (and other test environments), this introduces a `--junit=output.xml` argument to the `meteor self-test ...` command.

This gains us a couple relatively functionally pleasing features, such as the ability to see [failed tests without scrolling through all the results](https://circleci.com/gh/abernix/meteor/1358) and [CircleCI Insights on test timings](https://circleci.com/build-insights/gh/abernix/meteor/abernix%252Fmore-selftest2).

One of the more important historical pieces of information we will gain with this is the timings of each individual test.  Using this information, which is [stored by CircleCI automatically](https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-patterns) in `/.circleci-task-data/circle-test-results/` and also available through their API, we should be able to auto-balance tests using a partitioning algorithm and have our tests all run in as little time as possible.  Though, I think after we gather this information for a week, we may be able to just manually re-balance the containers as well.

This utilizes the JUnit output format as it seems to be the most widely supported by CI environments – if I'm missing some common knowledge here, please do tell.  A standardized, widely-supported, simple JSON format would be nice!  Unfortunately, without that, this has the requirement of generating XML from the `meteor self-test` code.  Given that the structure was relatively simple and I was hesitant to include an XML-building npm (we have [`xmlbuilder`](https://github.com/meteor/meteor/tree/devel/packages/xmlbuilder) already as a Meteor package, used in Cordova, but it's not an individual [isopacket](https://github.com/meteor/meteor/blob/93d0c5f2bad2d75b98ec5ecdacfc6bf1d323a9de/tools/tool-env/isopackets.js#L52-L60)).  If we wanted to, I could change this to auto-install some xml-building package via a similar method as https://github.com/meteor/meteor/pull/8981, though that would require Meteor `release-1.6`.

### DIFF
--- a/tools/tool-testing/selftest.js
+++ b/tools/tool-testing/selftest.js
@@ -1,3 +1,4 @@
+import { inspect } from 'util';
 import { makeFulfillablePromise } from '../utils/fiber-helpers.js';
 import { spawn, execFile } from 'child_process';
 import * as files from '../fs/files.js';
@@ -1605,6 +1606,7 @@ class Test {
     this.fileHash = options.fileHash;
     this.tags = options.tags || [];
     this.f = options.func;
+    this.durationMs = null;
     this.cleanupHandlers = [];
   }
 
@@ -1797,6 +1799,16 @@ function getFilteredTests(options) {
   return new TestList(allTests, tagsToSkip, tagsToMatch, testState);
 };
 
+function groupTestsByFile(tests) {
+  const grouped = {};
+  tests.forEach(test => {
+    grouped[test.file] = grouped[test.file] || [];
+    grouped[test.file].push(test);
+  });
+
+  return grouped;
+}
+
 // A TestList is the result of getFilteredTests.  It holds the original
 // list of all tests, the filtered list, and stats on how many tests
 // were skipped (see generateSkipReport).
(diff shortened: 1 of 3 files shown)


abernix:
For example!  One of the containers failed on this PR, but here's the results: https://circleci.com/gh/meteor/meteor/5617.

Currently, I'm just dumping the whole `test` object in the event of a failure since it seemed mildly helpful in debugging some of the output parsing/wait-time issues, but we could probably refine it a bit.

benjamn:
If you really wanted to get fancy, you could use the `xmlbuilder` Meteor package!

tools/tool-testing/selftest.js
```diff
@@ -1856,8 +1868,108 @@ class TestList {
   // Mark a test's file as having failures.  This prevents
   // saveTestState from saving its hash as a potentially
   // "unchanged" file to be skipped in a future run.
-  notifyFailed(test) {
+  notifyFailed(test, failure) {
     this.fileInfo[test.file].hasFailures = true;
+    test.failure = failure || true;
```
benjamn: I think this should be
```js
notifyFailed(test, failure = true) {
  this.fileInfo[test.file].hasFailures = true;
  test.failure = failure;
}
```
If someone calls `testList.notifyFailed(test, false)`, then `test.failure = failure || true` will set `test.failure` to `true`, which seems surprising.

tools/tool-testing/selftest.js
```diff
@@ -1856,8 +1868,108 @@ class TestList {
...
+
+    files.writeFile(path,
+      [
+        xmlHeader,
+        `<testsuites>`,
+        testSuitesString,
+        `</testsuites>`,
+      ].join('\n'),
+      'utf8',
+    );
```
benjamn: I wasn't really serious about using `xmlbuilder` (via an isopacket) here, and I am *not* asking you to change this code, but this sure is a lot of raw string concatenation!

tools/tool-testing/selftest.js
abernix: I think the signature could be better for this.  Perhaps `notifyFailed(test, failureObject = null)`, and it could set a variable called `test.failed = true` (always) and `test.failureObject = failureObject`.  I'll adjust tomorrow unless there is an objection.

tools/tool-testing/selftest.js
abernix: I know.  I almost did the isopacket approach (and introduced a new one, since `xmlbuilder` is only otherwise built into `cordova-support` and I secretly envisioned that going away and didn't like the idea of adding another isopacket.  Additionally, the `xmlbuilder` package used the `xmlbuilder` which was 7 major versions behind.

PR #9018, opened 2017-08-15
Split coffeescript package, part 2

GeoffreyBooth:
See #8960.


packages/non-core/coffeescript-compiler/package.js
```diff
@@ -0,0 +1,21 @@
...
+Npm.depends({
+  'coffeescript': '1.12.7',
+  'source-map': '0.5.6'
+});
+
+Package.onUse(function (api) {
+  api.use('babel-compiler@6.19.4');
+  api.use('ecmascript@0.8.2');
+
+  api.addFiles(['coffeescript-compiler.js'], 'server');
```
benjamn: Since you're using `ecmascript`, I would make this
```js
api.mainModule("coffeescript-compiler.js", "server");
```

packages/non-core/coffeescript-compiler/coffeescript-compiler.js
```diff
@@ -0,0 +1,214 @@
...
+// use of coffee.run which we don't use. This conflicts with the tool's use of
+// Error.prepareStackTrace to properly show error messages in linked code.
+// Restore the tool's one after CoffeeScript clobbers it at import time.
+if (Error.METEOR_prepareStackTrace) {
+  Error.prepareStackTrace = Error.METEOR_prepareStackTrace;
+}
+
+
+// The CompileResult for this CachingCompiler is a {source, sourceMap} object.
+CoffeeScriptCompiler = class CoffeeScriptCompiler {
```
benjamn: If you use `api.mainModule` in `package.js`, then this line can just be
```js
export class CoffeeScriptCompiler {
```
and `api.export("CoffeeScriptCompiler", "server")` will use the exported `CoffeeScriptCompiler`.

packages/non-core/coffeescript/package.js
```diff
@@ -1,29 +1,29 @@
 Package.describe({
-  summary: "Javascript dialect with fewer braces and semicolons",
-  version: "1.12.6_1"
+  name: 'coffeescript',
```
benjamn: Glad to have a name here now!

packages/non-core/coffeescript/package.js
```diff
@@ -1,29 +1,29 @@
...
 Package.registerBuildPlugin({
-  name: "compileCoffeescript",
-  use: ['caching-compiler', 'ecmascript'],
-  sources: ['plugin/compile-coffeescript.js'],
-  npmDependencies: {
-    "coffeescript": "1.12.6",
-    "source-map": "0.5.6"
-  }
+  name: 'compile-coffeescript',
+  use: ['caching-compiler@1.1.9', 'ecmascript@0.8.2', 'coffeescript-compiler@1.12.7_1'],
```
benjamn: For the record, this means an application developer who's using `coffeescript` can use any version of `caching-compiler` that's `>= 1.1.9` and has the same major version (`1.x.y`). Likewise, any version of `ecmascript` is fair game if it starts with a `0.` and is more recent than `0.8.2`.

I would suggest that we make the `coffeescript-compiler` version constraint exact (`@=1.12.7_1`), unless you think users might ever want to use different versions of `coffeescript` and `coffeescript-compiler`?

benjamn:
Looks like Circle CI is having problems right now, by the way: https://status.circleci.com/

GeoffreyBooth:
@benjamn I’ve revised per your notes.

packages/non-core/coffeescript-compiler/package.js
```diff
@@ -0,0 +1,23 @@
...
+Npm.depends({
+  'coffeescript': '1.12.7',
+  'source-map': '0.5.6'
+});
+
+Package.onUse(function (api) {
+  api.use('babel-compiler@6.19.4');
+  api.use('ecmascript@0.8.2');
+
+  api.mainModule(['coffeescript-compiler.js'], 'server');
```
benjamn: This line is causing some legit test failures. It should just be a string, not an array (as there can be only one main module):
```js
api.mainModule('coffeescript-compiler.js', 'server');
```
I tried pushing a commit to this branch, but I think only you can do that, @GeoffreyBooth.

GeoffreyBooth: Updated. I have “allow edits from maintainers” checked, so I’d think you should be able to push commits . . .

GeoffreyBooth:
Looks like the build passed!

benjamn:
Are you able to republish the packages, or should I?

GeoffreyBooth:
I don't think I am . . . ?

benjamn:
Are you fine with me publishing them now, then?

GeoffreyBooth:
Of course.

benjamn:
Okay! All three packages have been published, and I can confirm that I get the new versions when I run `meteor add coffeescript` in a new 1.5.1 application:
```sh
1:~/dev/coffee-split-test% ~/.meteor/meteor --version
Meteor 1.5.1
1:~/dev/coffee-split-test% ~/.meteor/meteor add coffeescript

Changes to your project's package version selections:

coffeescript           added, version 1.12.7_1
coffeescript-compiler  added, version 1.12.7_1

coffeescript: Javascript dialect with fewer braces and semicolons
```

GeoffreyBooth:
Thanks!

benjamn:
Thank you for all your work on this, @GeoffreyBooth!

PR #9031, opened 2017-08-21
Patch Node 4.8.4 to fix segmetation faults.

benjamn:
Brings https://github.com/nodejs/node/pull/14829 to the version of Node that ships with Meteor, hopefully preventing segmentation faults like those reported in #8648.

These changes will be officially released in Node 4.8.5, but we didn't want to [wait until September or October](https://github.com/nodejs/node/pull/14829#issuecomment-323350770) for that release.

All credit for these changes belongs to @abernix. I simply adapted his script to run in a separate Jenkins job, so that we don't have to rebuild Node from source every time we publish a new dev bundle.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,9 +124,9 @@ jobs:
           name: Get Ready
           command: |
             eval $PRE_TEST_COMMANDS;
-            ./meteor --help
-          # shouldn't take longer than 5 minutes
-          no_output_timeout: 5m
+            ./meteor --get-ready
+          # shouldn't take longer than 10 minutes
+          no_output_timeout: 10m
       # Clear dev_bundle/.npm to ensure consistent test runs.
       - run:
           name: Clear npm cache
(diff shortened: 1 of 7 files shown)


formspoint:
@hwillson do you work at mdg these days?

abernix:
The failures on the `assets - helper exists to unicode normalize path strings - assets` test are very legitimate.  Based on the `process.config` of this binary, we're missing [ICU support](icu-project.org) (which Node.js official releases are built with).

Will get that fixed soon!

.circleci/config.yml
```diff
@@ -124,9 +124,9 @@ jobs:
           name: Get Ready
           command: |
             eval $PRE_TEST_COMMANDS;
-            ./meteor --help
-          # shouldn't take longer than 5 minutes
-          no_output_timeout: 5m
+            ./meteor --get-ready
```
benjamn: This worked amazingly well!

PR #9034, opened 2017-08-22
Visually notify of uncaught exceptions breaking client tests

### LINKED ISSUE #4979: Tinytest silently ignores all tests on uncaught errors
**_1 Upvote**_ 

```sh
git clone git@github.com:iDoRecall/email-normalize.git
git checkout 64b18f371d8d3f7d62d6821a46d4ef62b752472d
cd email-normalize
meteor test-packages ./
```

Notice that apparently everything went ok. At a closer examination, only `S`erver tests have been run, despite `pacakge.js` including the test on both client and server.

This is caused by the uncaught ReferenceError: `Npm is not defined` on the client, which halts the tests. Would it be possible to catch [this type of case](https://github.com/iDoRecall/email-normalize/issues/1#issuecomment-132132536)?

hwillson:
This PR is intended to help fix #4979. I know external package `Tinytest` use has dwindled since adding `--driver-package` support to `meteor test-packages`, but since we are still supporting it (and it's used by the Meteor core), we might as well improve the `test-in-browser` UI experience a bit (let me know if anyone disagrees!).

These changes add a global `window.onerror` event handler to the `test-in-browser` package, to catch any previously uncaught exceptions. When uncaught exceptions are found, an alert like the following is shown at the top of the `test-in-browser` UI:

screenshot 2017-08-22 09 24 27

This should help notify developers of potential errors behind the scenes, that are preventing the client side tests from running.

(P.S. --> I snuck `ecmascript` in there ... hope that's okay!)

### DIFF
--- a/packages/test-in-browser/package.js
+++ b/packages/test-in-browser/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Run tests interactively in the browser",
-  version: '1.0.13',
+  version: '1.0.14',
   documentation: null
 });
 
--- a/packages/test-in-browser/package.js
+++ b/packages/test-in-browser/package.js
@@ -5,6 +5,7 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
+  api.use('ecmascript');
   // XXX this should go away, and there should be a clean interface
   // that tinytest and the driver both implement?
   api.use('tinytest');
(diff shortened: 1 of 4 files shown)


benjamn:
Oh wow, this is awesome!

packages/test-in-browser/package.js
```diff
@@ -5,6 +5,7 @@ Package.describe({
 });
```
benjamn: Do we need to bump the `test-in-browser` package version?

hwillson: Oops - good catch. Yes, we should - I've bumped it to `1.0.14`. Thanks!

Issue #7213, opened 2016-06-14
Add a onUpdateUser hook to the accounts-base package


FGM:
## Problem to solve

Since commit 00eda434206aba34e91ccde25019c2b035b3c746 (OAuth login to an existing account now _does_ update the services.PROVIDER), `accounts-server.js` contains mention of a `onUpdateUser` hook remaining to be implemented, in function `updateOrCreateUserFromExternalService()`. Still present in the `release-1.5` branch
## Why this is a problem

It turns out this hook is missing in an accounts package for transparent Drupal authentication I've been developing (https://github.com/FGM/accounts-drupal) : due to this missing hook, when Meteor is notified about a change in the user exported information, the info is properly merged to the database under the `services` key for the user, but the `profile` field cannot be updated.

This appears to have been the reason for the mention about that hook in the first place: 

>  XXX provide an onUpdateUser hook which would let apps update the profile too
## Current workaround

The current workaround is to delete the user and relog them, which causes the profile to be regenerated as it should, but means any work in progress by the user is lost while they are logged out, and all login information for other accounts providers is deleted too.
## Suggested solution

Implement such a hook. If MDG is willing, I could probably submit a PR to do so.

Note: one similar hook was added recently on logout in #6889, so adding its equivalent for updates seems just a consistency measure.

FGM:
Still an issue on release-1.5 HEAD.

abernix:
Thanks for this feature request.  I think this is a reasonable request and this could be a candidate for `pull-requests-encouraged`, however I'll cite the [feature request guidelines](https://github.com/meteor/meteor/blob/devel/Contributing.md#feature-requests) for the sake of completeness and because there are a lot of other feature requests:

> You'll have the best chance of getting a change into core if you can build consensus in the community for it. Start by creating a well specified feature request as a Github issue.

> Help drive discussion and advocate for your feature on the Github ticket (and perhaps the forums). The higher the demand for the feature and the greater the clarity of it's specification will determine the likelihood of a core contributor prioritizing your feature by flagging it with the pull-requests-encouraged label.

Do you mind writing out a bit of a proposal here for exactly what you had in mind or how you planned to implement it?  I don't think this sounds terribly complicated, so if it sounds good you'll be encouraged to make a PR for it.

hwillson:
[...]

PR #9042, opened 2017-08-24
added onUserUpdate hook

gerwinbrunner:
I found a missing feature documented in the code here:
https://github.com/meteor/meteor/blob/master/packages/accounts-base/accounts_server.js#L1433

So I just implemend it in the way the `onCreateUser` hook was done.


sebakerckhof:
It's weird to me that `onCreateUser` would be called for every created user, while `onUpdateUser` would only be called for users logging in via oAuth. If this is a useful feature, I'd at least give it a more descriptive name along the lines of 'onExternalLogin`

merlinpatt:
I agree that it should probably be called `onExternalLogin` to be more clear. But I'm not sure what benefit this provides. There isn't a way to pass options through to the user doc like you can with `onCreateUser`.

`onCreateUser` allows you to pass options in via `Accounts.createUser(options)`. But, as far as I can tell, `Meteor.loginWith<ExternalService>` has no means to pass options to this new function `onExternalLogin`. And these should share the same features

gerwinbrunner:
@sebakerckhof Thanks for the input... I just went with what was in the comments of the original code. I'll change the name.

@merlinpatt The purpose for this is to be able to update profile (or other data) after an OAuth login.
That information is passed by the OAuth service. An example would be to change the users country of residence if he changed that info in the original OAuth service.

packages/accounts-base/accounts_server.js
```diff
@@ -119,6 +119,22 @@ export class AccountsServer extends AccountsCommon {
 
     this._onCreateUserHook = func;
   }
+
```
hwillson: Can you adjust this so there is only one blank line before and after the `onExternalLogin` function?

packages/accounts-base/accounts_server.js
```diff
@@ -1433,15 +1449,22 @@ Ap.updateOrCreateUserFromExternalService = function (
     // We *don't* process options (eg, profile) for update, but we do replace
```
hwillson: This comment needs to change since we're now providing a way to process options. Please remove the `XXX` line, and consider updating this comment to better reflect the new functionality. Actually, you could consider just removing this comment entirely - now that we're addressing options, the code here pretty much speaks for itself.

hwillson:
Hi @gerwinbrunner - just touching base on this. Have you had a chance to review my comments?

gerwinbrunner:
Sorry, having pretty busy the last week. I will look into it this week 
and update the PR.

Hugh Willson wrote:

hwillson:
@gerwinbrunner Have you had a chance to look at this again? Let us know if you'll have some time to wrap this up. Thanks!

hwillson:
All outstanding items here should now be addressed, so this is ready for a final review.

benjamn:
Rebased and pushed to `devel` as https://github.com/meteor/meteor/commit/20f698aec8f711825e061b06604eb9d43afc6cdb

FGM:
AFAICS this actually closes https://github.com/meteor/meteor/issues/7213 from 06/2016, and its followup  https://github.com/meteor/meteor-feature-requests/issues/68

hwillson:
Good catch - thanks @FGM!

PR #9044, opened 2017-08-25
Make bcrypt rounds configurable

dhrubins:
Background: the bcrypt work factor or cost factor is a way of setting the difficulty of cracking a bcrypt'ed password if the databse is ever breached. The current hardcoded setting of 10 rounds in `accounts-password` is a bit low, given the availability of GPUs and cloud password cracking services. Every operator should choose the right value for their threat model, but 12-14 is likely sufficient for most services. At Legal Robot, we use 14.

For more detail, see the discussion at meteor/meteor-feature-requests/issues/55

This PR adds a configuration option for those using `accounts-password` that want to increase the bcrypt rounds with little impact to existing users. During the password check, after a password is validated, the bcrypt'ed password is replaced with an updated value *if* the current number of bcrypt rounds exceeds that of the stored bcrypt value. The default behavior is to *not* change anything, but the value for `bcryptRounds` is easily configurable by setting `Accounts.config({ bcryptRounds: 14 })` on the server.


packages/accounts-password/password_server.js
```diff
@@ -67,6 +67,15 @@ Accounts._checkPassword = function (user, password) {
 
   if (! bcryptCompare(password, user.services.password.bcrypt)) {
     result.error = handleError("Incorrect password", false);
+  } else {
```
hwillson: Since there is only one `if` under `else` here, could you merge them into an `else if` to improve readability a bit?

packages/accounts-password/password_server.js
```diff
@@ -67,6 +67,15 @@ Accounts._checkPassword = function (user, password) {
...
+    // password checks out, but user bcrypt may need update
+    if (user.services.password.bcrypt && Accounts._bcryptRounds > Number(user.services.password.bcrypt.substring(4, 6))) {
```
hwillson: Please wrap your code changes at 80 characters.

packages/accounts-password/password_server.js
dhrubins: Makes sense - done in a5206a6

packages/accounts-password/password_server.js
dhrubins: No problem - done in b5cd99a

packages/accounts-password/password_server.js
```diff
@@ -67,6 +67,20 @@ Accounts._checkPassword = function (user, password) {
 
   if (! bcryptCompare(password, user.services.password.bcrypt)) {
     result.error = handleError("Incorrect password", false);
+  } else if (
+    user.services.password.bcrypt && 
+    Accounts._bcryptRounds > 
+      Number(user.services.password.bcrypt.substring(4, 6))
```
benjamn: Will this work if `bcryptRounds >= 100` (i.e., more than two digits)?

packages/accounts-password/password_server.js
```diff
@@ -67,6 +67,20 @@ Accounts._checkPassword = function (user, password) {
 
   if (! bcryptCompare(password, user.services.password.bcrypt)) {
     result.error = handleError("Incorrect password", false);
+  } else if (
+    user.services.password.bcrypt && 
+    Accounts._bcryptRounds >
```
benjamn: Maybe `!=` instead of just `>`?

packages/accounts-password/password_server.js
benjamn: While you're at it, let's just extract a helper function called something like `getRoundsFromBcryptHash`?

dhrubins: Yeah, that code will break with a work factor above 99, but it shouldn't be possible to implement bcrypt with such a high work factor since it gets expanded geometrically (rounds = 2^[work factor]). [This](http://web.mit.edu/freebsd/head/secure/lib/libcrypt/crypt-blowfish.c) FreeBSD implementation of bcrypt references a maximum of 31 rounds. I'm not sure if 31 is implementation-specific or the theoretical maximum for bcrypt, but the effective maximum for most current server hardware is probably 16 or 17. From benchmarking on my MacBook Pro, hashing with a work factor of 20 takes about 2.4 minutes and 31 would take ~12 years. The rule of thumb I'm aware of (totally unattributed) is that hashing should take ~1 second with current server hardware. If I can find a source for that rule of thumb, I'll do a PR for this in the Guide.

dhrubins: Sure, I'm happy to move that into a helper function. At the risk of over-building but with a mind toward implementing other hashing algorithms, I'll grab both the algorithm and whatever parameters go with that algorithm.

packages/accounts-password/password_server.js
dhrubins: Yep, will fix

hwillson:
@dhrubins Have you had a chance to dive back into this? Let us know if you'll have some time to wrap this up. Thanks!

hwillson:
I've addressed the outstanding review comments and added tests, so this should now be ready for a final review. One thing to note - I bumped `accounts-base` to 1.3.6 even though `devel` is currently at 1.3.4, because PR https://github.com/meteor/meteor/pull/9042 is bumping `accounts-base` to 1.3.5. Depending on the merge order of these PR's, this might need to be adjusted.

packages/accounts-password/password_tests.js
```diff
@@ -1886,4 +1886,51 @@ if (Meteor.isServer) (function () {
     ]);
   });
 
+  Tinytest.addAsync(
+    'passwords - allow custom bcrypt rounds',
+    function (test, done) {
+      function getUserHashRounds(user) {
+        return Number(user.services.password.bcrypt.substring(4, 6));
```
benjamn: Let's use `parseInt` here to avoid interpreting leading 0s as octal.

hwillson: Done - and I've adjusted rounds extraction to use split string segments. Thanks!

benjamn:
Rebased and pushed to `devel` as 1687537f6ea03312183059fed45f12f350de2f17.

PR #9052, opened 2017-08-30
Throw more meaningful error if .babelrc is not a valid JSON file

### LINKED ISSUE #8792: native: Unexpected end of input when .babelrc is empty
https://github.com/meteor/meteor/issues/7965

reproduce 

https://github.com/crapthings/meteor-empty-babelrc

```
=> Errors prevented startup:

   While processing files with ecmascript (for target web.browser):
   native: Unexpected end of input
   at Object.parse (native)
   at BabelCompiler.BCp._inferFromBabelRc (packages/babel-compiler.js:246:14)
   at BabelCompiler.BCp.inferExtraBabelOptions (packages/babel-compiler.js:236:10)
   at BabelCompiler.BCp.processOneFileForTarget (packages/babel-compiler.js:166:10)
   at BabelCompiler.<anonymous> (packages/babel-compiler.js:109:26)
   at Array.forEach (native)
   at BabelCompiler.BCp.processFilesForTarget (packages/babel-compiler.js:108:14)


   While processing files with ecmascript (for target os.osx.x86_64):
   native: Unexpected end of input
   at Object.parse (native)
   at BabelCompiler.BCp._inferFromBabelRc (packages/babel-compiler.js:246:14)
   at BabelCompiler.BCp.inferExtraBabelOptions (packages/babel-compiler.js:236:10)
   at BabelCompiler.BCp.processOneFileForTarget (packages/babel-compiler.js:166:10)
   at BabelCompiler.<anonymous> (packages/babel-compiler.js:109:26)
   at Array.forEach (native)
   at BabelCompiler.BCp.processFilesForTarget (packages/babel-compiler.js:108:14)
```

klaussner:
Just a small improvement of the error that is thrown if a `.babelrc` file doesn't contain valid JSON. For example, the error message will be
```
.babelrc is not a valid JSON file: Unexpected token }
```
instead of
```
Unexpected token }
```
for the invalid `.babelrc` file
```
{
  "plugins":
}
```

Resolves #8792

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -160,8 +160,16 @@ BCp._inferFromBabelRc = function (inputFile, babelOptions, cacheDeps) {
   var babelrcPath = inputFile.findControlFile(".babelrc");
   if (babelrcPath) {
     if (! hasOwn.call(this._babelrcCache, babelrcPath)) {
-      this._babelrcCache[babelrcPath] =
-        JSON.parse(inputFile.readAndWatchFile(babelrcPath));
+      try {
+        this._babelrcCache[babelrcPath] =
+          JSON.parse(inputFile.readAndWatchFile(babelrcPath));
+      } catch (e) {
+        if (e instanceof SyntaxError) {
+          e.message = ".babelrc is not a valid JSON file: " + e.message;
+        }
+
+        throw e;
+      }
     }
 
     return this._inferHelper(


packages/babel-compiler/babel-compiler.js
```diff
@@ -160,8 +160,16 @@ BCp._inferFromBabelRc = function (inputFile, babelOptions, cacheDeps) {
...
   if (babelrcPath) {
     if (! hasOwn.call(this._babelrcCache, babelrcPath)) {
-      this._babelrcCache[babelrcPath] =
-        JSON.parse(inputFile.readAndWatchFile(babelrcPath));
+      try {
+        this._babelrcCache[babelrcPath] =
+          JSON.parse(inputFile.readAndWatchFile(babelrcPath));
+      } catch (e) {
+        if (e instanceof SyntaxError) {
+          throw new Error(".babelrc is not a valid JSON file: " + e.message);
```
zimme: Seems like this gets rid of the stacktrace in `e` as it's not passed into the new error being thrown.

klaussner: Are you sure? The stack trace should be similar because the error is thrown at (almost) the same source location as before.

zimme: I'm thinking it's loosing the stack trace from the command that threw the error we're catching.

benjamn: What about just `e.message = ".babelrc is not a valid JSON file: " + e.message`?

klaussner: Sounds good!

PR #9062, opened 2017-09-06
Fix falsy `_id` in selector object not being treated as ID

### LINKED ISSUE #9060: Client side idStringify error (objects other than ObjectID) also with _id="" after core packages upgrade
**The version of Meteor showing the problem.**
Meteor version: 1.5.1

Packages updated:

```
accounts-base          upgraded from 1.3.1 to 1.3.2
babel-compiler         upgraded from 6.19.4 to 6.20.0
boilerplate-generator  upgraded from 1.1.1 to 1.2.0
ddp-client             upgraded from 2.0.0 to 2.1.0
ejson                  upgraded from 1.0.13 to 1.0.14
minifier-js            upgraded from 2.1.1 to 2.1.2
minimongo              upgraded from 1.2.1 to 1.3.0
modules                upgraded from 0.9.4 to 0.10.0
mongo                  upgraded from 1.1.22 to 1.2.0
mongo-dev-server       added, version 1.0.1
promise                upgraded from 0.8.9 to 0.9.0
webapp                 upgraded from 1.3.17 to 1.3.18
```

**The last version of Meteor where the problem did _not_ occur (if applicable)**
Same version before the update.

**The operating system you're running Meteor on.**
MacOs, but problem also occurs on Ubuntu 16.04.2 LTS Linux local and production. 

**The expected behavior**
Before the update there were no problems with calling `collection.findOne("")` or `collection.findOne({_id:""})`. I expect the function `findOne` to return a falsey value, such as null or undefined.

**The actual behavior.**
Throws error on `collection.findOne({_id:""})` 

```
Error: Meteor does not currently support objects other than ObjectID as ids
idStringify — 436bf8d602463537fc560f5794e92e560aa45dc0.js:56258
get — 436bf8d602463537fc560f5794e92e560aa45dc0.js:56013
e — 436bf8d602463537fc560f5794e92e560aa45dc0.js:57512
e — 436bf8d602463537fc560f5794e92e560aa45dc0.js:57443:156
e — 436bf8d602463537fc560f5794e92e560aa45dc0.js:57495
e — 436bf8d602463537fc560f5794e92e560aa45dc0.js:57393
e — 436bf8d602463537fc560f5794e92e560aa45dc0.js:45:14769
e — 436bf8d602463537fc560f5794e92e560aa45dc0.js:57658:119
findOne — 436bf8d602463537fc560f5794e92e560aa45dc0.js:62014
```
Is this behavior to be expected now with the updated packages?

I checked the [mongo 2.6 merge](https://github.com/meteor/meteor/pull/8815) and it seems, that it introduces new _id handling?

**A simple reproduction!**
Create a new Meteor project with the above package versions and call findOne({_id:""}) on a arbitrary client side collection.

klaussner:
In #8893 (Minimongo refactoring), [this commit](https://github.com/meteor/meteor/commit/92555c32ba40e969a0bd905975961326d2365f1e) changed how the `_id` property of a selector object is handled, but the new version fails for falsy `_id`s, e.g., empty strings (see #9060).
This pull request fixes the bug by checking if the selector object has an `_id` property instead of looking for a truthy value.

**Regarding falsy selectors and `_id`s:** 

The bug is fixed by the changes to `cursor.js`. Since `matcher.js` contains the same logic, I made the same changes there but I'm not sure if that's correct. Since the refactoring, `_selectorIsIdPerhapsAsObject` is called instead of `_selectorIsId` on the following line:

https://github.com/meteor/meteor/blob/5f13b20e1e4023a23a50b9c4f1c9e1ad9d88136c/packages/minimongo/matcher.js#L90

I think that, ~in combination with my changes,~ this contradicts the comment (and code) a few lines below (the one that starts with "protect against dangerous selectors") because falsy values are now allowed if they are valid IDs (`''` and `0`). Maybe @radekmie knows more about this? 🙂 

Fixes #9060

### DIFF
--- a/packages/minimongo/cursor.js
+++ b/packages/minimongo/cursor.js
@@ -1,4 +1,5 @@
 import LocalCollection from './local_collection.js';
+import { hasOwn } from './common.js';
 
 // Cursor: a specification for a particular subset of documents, w/ a defined
 // order, limit, and offset.  creating a Cursor with LocalCollection.find(),
@@ -11,7 +12,9 @@ export default class Cursor {
 
     if (LocalCollection._selectorIsIdPerhapsAsObject(selector)) {
       // stash for fast _id and { _id }
-      this._selectorId = selector._id || selector;
+      this._selectorId = hasOwn.call(selector, '_id')
+        ? selector._id
+        : selector;
     } else {
       this._selectorId = undefined;
 
--- a/packages/minimongo/matcher.js
+++ b/packages/minimongo/matcher.js
@@ -88,7 +88,9 @@ export default class Matcher {
 
     // shorthand -- scalar _id and {_id}
     if (LocalCollection._selectorIsIdPerhapsAsObject(selector)) {
-      const _id = selector._id || selector;
+      const _id = hasOwn.call(selector, '_id')
+        ? selector._id
+        : selector;
 
       this._selector = {_id};
       this._recordPathUsed('_id');
--- a/packages/minimongo/matcher.js
+++ b/packages/minimongo/matcher.js
@@ -86,16 +86,12 @@ export default class Matcher {
       return doc => ({result: !!selector.call(doc)});
     }
 
-    // shorthand -- scalar _id and {_id}
-    if (LocalCollection._selectorIsIdPerhapsAsObject(selector)) {
-      const _id = hasOwn.call(selector, '_id')
-        ? selector._id
-        : selector;
-
-      this._selector = {_id};
+    // shorthand -- scalar _id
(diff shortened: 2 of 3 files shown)


radekmie:
Yep, my fault there.

klaussner:
This should be ready for review now. 🙂  I've restored the old behavior in `matcher.js` and added a few more test cases for falsy `_id`s.

abernix:
Released (yesterday) as `minimongo@1.3.1`!  Thanks, @klaussner!

PR #9063, opened 2017-09-06
Add destroy to reactive-dict

zimme:
I had this laying around and thought I would put it up as a PR so we can make a decision about it.

It solves the problem @rclai brought up a while back in meteor/meteor-feature-requests#76 and earlier in the related meteor issue.

With this api added, one should be able to do the following.
```
Template.myTemplate.onCreated(function createdMyTemplate() {
  this.state = new ReactiveDict("myTemplate");
});

Template.awesome.onDestroy(function destroyMyTemplate() {
  this.state.destroy();
});
```
This would clear the `ReactiveDict`s data and unregister it from migration which would allow the template to be instantiated again and register it's `ReactiveDict`, `state`, for migration with the same name.

On another note, I do believe one should try and append a random id to the template's `ReactiveDict` name as this would allow for the template to be rendered multiple times in the same app at the same time. i.e. ```this.state = new ReactiveDict(`myTemplate-${Random.hexString(10)`);```


zimme:
If this gets an ok to move forward, I'll add tests.

zimme:
Could this cleanup be made automatically using `WeakMap` or some change to how the migration/migrated data is handled?

benjamn:
Looks good so far! Let's move forward with the tests (and not bother with `WeakMap` for now).

zimme:
No problem :+1:

zimme:
Done.

hwillson:
Awesome @zimme - thanks! We're all set here - LGTM!

PR #9069, opened 2017-09-08
Added renderDocument in server-sink

ardatan:
Some view frameworks like Angular returns the full document HTML. We can do some trick to get seperately head and body in project code, but it'd better to move this trick into this package.
This new method extracts head and body, and appends them seperately into server-sink.

### DIFF
--- a/packages/server-render/server-sink.js
+++ b/packages/server-render/server-sink.js
@@ -1,3 +1,6 @@
+const HEAD_REGEX = /<head[^>]*>((.|[\n\r])*)<\/head>/im
+const BODY_REGEX = /<body[^>]*>((.|[\n\r])*)<\/body>/im;
+
 export class ServerSink {
   constructor(request, arch) {
     this.request = request;
@@ -29,6 +32,16 @@ export class ServerSink {
   renderIntoElementById(id, html) {
     this.htmlById[id] = "";
     this.appendToElementById(id, html);
+  }
+
+  renderDocument(html){
+      // Extract head
+      const head = HEAD_REGEX.exec(html)[1];
+      this.appendToHead(head);
+
+      // Extract body
+      const body = BODY_REGEX.exec(html)[1];
+      this.appendToBody(body);
   }
 }
 


packages/server-render/server-sink.js
```diff
@@ -1,3 +1,6 @@
+const HEAD_REGEX = /<head[^>]*>((.|[\n\r])*)<\/head>/im
+const BODY_REGEX = /<body[^>]*>((.|[\n\r])*)<\/body>/im;
```
benjamn: If you want to do this right, you must use an actual parser to extract this HTML, like [the one we use in `server-register.js`](https://github.com/meteor/meteor/blob/d416ed913dc3a465c89d02cb2bdb07cadfa512f7/packages/server-render/server-register.js#L28-L45). HTML is not a regular language (the class of languages that regular expressions are adequate for parsing), but a context-free language.

packages/server-render/server-sink.js
```diff
@@ -30,6 +33,16 @@ export class ServerSink {
...
   }
+
+  renderDocument(html){
+      // Extract head
+      const head = HEAD_REGEX.exec(html)[1];
+      this.appendToHead(head);
+
+      // Extract body
+      const body = BODY_REGEX.exec(html)[1];
+      this.appendToBody(body);
```
benjamn: Honestly, there's nothing happening here that a user of the `ServerSink` couldn't already do by calling `sink.appendToHead` and `sink.appendToBody`, and that code would be much easier to understand than `sink.renderDocument`.

benjamn:
I appreciate your effort and initiative, @ardatan, but this is a good example of why we have a process for proposing changes, discussing them, and (only after consensus is established) submitting a pull request: https://github.com/meteor/meteor/blob/devel/Contributing.md#proposing-your-change

Please consider creating a new issue at https://github.com/meteor/meteor-feature-requests/issues/new if you want to keep working on this. For the time being, I'm going to close this PR, because I don't think it meets the needs of any full-document-rendering frameworks (Angular, Vue, etc.), and the implementation needs work (see my comments above about using an actual HTML parser instead of regular expressions).

PR #9084, opened 2017-09-12
Add temporary `dev_bundle.xxx` directory to .gitignore

klaussner:
While downloading a new `dev_bundle`, thousands of files are created in the temporary `dev_bundle.xxx` directory, which can slow down editors and IDEs that are tracking the Git status in the background.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /.meteor
 *~
 /dev_bundle
+/dev_bundle.xxx
 /dev_bundle*.tar.gz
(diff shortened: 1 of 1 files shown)


abernix:
Thx!  I've added `/dev_bundle_XXX` too, which is used [on Windows](https://github.com/meteor/meteor/blob/devel/scripts/windows/download-dev-bundle.ps1#L44).  (Probably different because Windows cares a lot more about extensions?  Not sure, but seems relevant.)

PR #9092, opened 2017-09-15
Add DDP.onReconnect(), deprecate conn.onReconnect 

hwillson:
Creating a new PR to continue work on https://github.com/meteor/meteor/pull/5677 (since the original PR's branch is locked). Original PR description:

> Deprecate use of connection.onReconnect = func. Instead, a new
DDP.onReconnect(callback) method should be used to register callbacks to call
when a connection reconnects. The connection that is reconnecting is passed as
the only argument to callback. This is used by the accounts system to relogin
on reconnects without interfering with other code which uses
connection.onReconnect.

Definitely still work in progress.

Fixes https://github.com/meteor/meteor/issues/5665.

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -68,13 +68,17 @@ When `meteor` is run from a checkout, a `dev_bundle` is automatically downloaded
 * Packages [used by `meteor-tool`](scripts/dev-bundle-tool-package.js)
 * Packages [used by the server bundle](scripts/dev-bundle-server-package.js)
 
-While it may be tempting to make changes to these variables, please consider the repercussions (including compatibility and stability) and make sure to test changes extensively.  For example, major version changes (especially to Node.js and MongoDB) usually requires substantial changes to other components.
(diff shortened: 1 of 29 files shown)


hwillson:
Hi all - we should be good to go here (ready for review), so I'll remove the WIP prefix. If anyone is interested in a quick overview of this issue and the fix, see the updated `History.md` entry. A few extra things to note:

1. These changes are backwards compatible (even though the existing `connection.onReconnect` handling is mostly-broken, people still might be using it as is; for a good breakdown of why it makes sense to preserve this broken backwards compatibility, see https://github.com/meteor/meteor/issues/5665#issuecomment-157622862).

2. The [`DDP.connect`](https://github.com/meteor/docs/blob/master/source/api/connections.md) docs will have to updated after (if) this PR is merged. The `onReconnect` section under `DDP.connect` should be removed, and a new `DDP.onReconnect` section should be added. I'll prep a docs PR for this.

Thanks!

PR #9095, opened 2017-09-15
Standardize Npm.require exceptions and limit lookup paths.

benjamn:
Ever since Meteor 1.3 first introduced a module system based on something other than `Npm.require`, we've continued throwing missing module exceptions that refer to `Npm.depends` and/or `Npm.require`, even if the developer called `require` or used an `import` declaration. This commit fixes that, so that all missing module exceptions look like `Cannot find module "module/name"`.

I also noticed recently that `Npm.require` is capable of returning modules installed in `node_modules` directories completely outside the app, which is bad news for development/production reproducibility. Fixed that too.

CC @hwillson who has spoken of deprecating `Npm.require` entirely, and
just using `require` everywhere, instead.

### DIFF
--- a/packages/modules-runtime/server.js
+++ b/packages/modules-runtime/server.js
@@ -15,7 +15,7 @@ makeInstallerOptions.fallback = function (id, parentId, error) {
   if (topLevelIdPattern.test(id)) {
     if (typeof Npm === "object" &&
         typeof Npm.require === "function") {
-      return Npm.require(id);
+      return Npm.require(id, error);
     }
   }
 
--- a/tools/static-assets/server/boot.js
+++ b/tools/static-assets/server/boot.js
@@ -175,6 +175,9 @@ var loadServerBundles = Profile("Load server bundles", function () {
       });
     }
 
+    // Add dev_bundle/server-lib/node_modules.
+    addNodeModulesPath("node_modules");
+
     function statOrNull(path) {
       try {
         return fs.statSync(path);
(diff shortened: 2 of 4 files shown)


tools/static-assets/server/boot.js
```diff
@@ -193,47 +196,41 @@ var loadServerBundles = Profile("Load server bundles", function () {
...
-          return require(fullPath);
+        var resolved = require.resolve(name);
+        if (resolved === name && ! path.isAbsolute(resolved)) {
+          // If require.resolve(id) === id and id is not an absolute
+          // identifier, it must be a built-in module like fs or http.
+          return require(resolved);
         }
 
-        try {
-          return require(name);
```
benjamn: Not calling `require(name)` unconditionally here is the key to keeping lookups from escaping the application.

packages/modules-runtime/server.js
```diff
@@ -15,7 +15,7 @@ makeInstallerOptions.fallback = function (id, parentId, error) {
   if (topLevelIdPattern.test(id)) {
     if (typeof Npm === "object" &&
         typeof Npm.require === "function") {
-      return Npm.require(id);
+      return Npm.require(id, error);
```
benjamn: Intentionally undocumented API: you can now pass an error object as the second argument to `Npm.require` to control what gets thrown if the import fails.

kentonv:
This broke Sandstorm, which is fundamentally not a hermetic Meteor app and actually depends on the ability to load node extensions that exist in an up-tree node_modules directory. These extensions are not part of any npm package since they are built together with the Sandstorm C++ build for the sole purpose of talking to Sandstorm. It would be inconvenient to try to shoehorn them into npm.

I tried the 1.5.2.2 release candidate but it doesn't solve our problem.

Is there an easy way for us to restore the previous behavior? Can I get at the system's real `require` function somehow?

benjamn:
It sounds like Sandstorm is careful to provide the same environment in production as in development, but that's not something Meteor can ensure, so I'm sure you can appreciate that allowing apps to depend on packages that might not exist in production would be risky for most (non-Sandstorm) apps.

Another problem with the previous behavior is that the `require` call was relative to an arbitrary file inside the `.meteor/local` directory. Generally speaking, there is no single real `require` function in Node, since `require` is scoped to (and behaves differently in) each module. The previous implementation ignored this subtlety.

That said, I think we can find a solution that works even better for your use case than the previous behavior.

If you require the `"module"` built-in module on the server, you can construct a `Module` object that knows exactly where to look for your external packages:
```js
const Module = require("module");
const mod = new Module;
mod.paths = [pathToExternalNodeModules];
mod.require("external-package");
```

Not only will this find `external-package` using native Node, it also won't look in any unnecessary places, so it should be faster than `Npm.require`.

kentonv:
To be clear, this affects the build of Sandstorm itself, not the apps. Sandstorm's "shell" UI is a Meteor app. But this app is only one component of the overall Sandstorm system, which also includes daemons written in C++.

Thanks for pointing out the `"module"` module. I wasn't aware of it. It looks like I should be able to solve the problem with that.

kentonv:
It turns out that using the "module" module didn't really help because it still required me to know the exact path to import. If I know that, then I can just as easily pass it to `Npm.require()`.

I ended up writing code that does the directory tree walk manually: https://github.com/sandstorm-io/sandstorm/commit/c4fb0d825c072220c2b26ccc0754a325e1283493

PR #9099, opened 2017-09-16
Explain how to enable `dev_bundle` caching

klaussner:
The _"Dev Bundle" versions_ section in `Development.md` currently gives the impression that `dev_bundle` caching is enabled by default. This PR adds a short paragraph about the `SAVE_DEV_BUNDLE_TARBALL` environment variable, which is used to enable caching. :slightly_smiling_face:


abernix:
Right you are!  I've had this set in my shell environment for ages, so didn't even think about it when I wrote this.

Thanks! 😉

(And thanks for the other typo fixes too).

PR #9103, opened 2017-09-17
dynamic-import: fix modules version recognition to fix caching

theosp:
A fix to the dynamic-import code to correctly detect modules version.

Without this fix dynamic-import can't detect correctly modules versions
which breaks the dynamic-import cache functionality.

[I prepared a video that demonstrates the issue and the fix.](https://drive.google.com/open?id=0B3MQ_v8orqd_QURnUVZOaDBFWms)

-Daniel

### DIFF
--- a/packages/dynamic-import/client.js
+++ b/packages/dynamic-import/client.js
@@ -120,6 +120,8 @@ function fetchMissing(missingTree) {
 }
 
 function getFromTree(tree, id) {
+  id = id.replace(":", "_");
+
   id.split("/").every(function (part) {
     return ! part || (tree = tree[part]);
   });


abernix:
Thanks very much for reporting this, @theosp.

**To summarize the issue/video:**

A dynamic `import(...)` on a Meteor package (i.e. `import('meteor/my:package').then(...)`) fails to achieve a cache hit when it is present the cache, resulting in the module being re-fetched from the server.

**Reason being:**

When [`meteorInstall.fetch` is called](https://github.com/meteor/meteor/blob/devel/packages/dynamic-import/client.js#L17) by `Module.prototype.prefetch` (via [`install`](//npm.im/install)), module identifiers contain colons in the package name (i.e. `/node_modules/meteor/my:package`), while the `__DYNAMIC_VERSIONS__` object (populated in [`isobuild/bundler.js`](https://github.com/meteor/meteor/blob/devel/tools/isobuild/bundler.js#L1052)'s `_emitResources` method) has the package name as it is stored on disk (i.e. with `:`s converted to `_`; `/node_modules/meteor/my_package`).

As you said in your video, I'm not sure if that's the perfect place to correct the problem, but you've certainly identified the issue, so thank you!

/cc @meteor/platform

benjamn:
Eventually I would like to fix [this code](https://github.com/meteor/meteor/blob/a8cb0835b4c190d5443a40b21e7d6b4f2feab70a/tools/isobuild/bundler.js#L1191) to use a path that hasn't had its `:`s replaced by `_`s (which happens [here](https://github.com/meteor/meteor/blob/a8cb0835b4c190d5443a40b21e7d6b4f2feab70a/tools/isobuild/bundler.js#L707)), but for now I've tweaked this logic to tolerate either `:` or `_` (https://github.com/meteor/meteor/commit/7cbb2bb5799cc9a56067398e3680a7a5c73b09ca), added a few tests, and published `dynamic-import@0.1.3` with this bug fixed.

hwillson:
Ignore this comment - internally testing @meteor/platform

PR #9107, opened 2017-09-19
CoffeeScript 2.0.2

GeoffreyBooth:
Surely it can’t be this easy. I bumped the version numbers and the tests all still pass. Should probably get some more tests.

See also https://github.com/meteor/meteor/pull/8728#issuecomment-330431601 @benjamn

### DIFF
--- a/packages/non-core/coffeescript-compiler/package.js
+++ b/packages/non-core/coffeescript-compiler/package.js
@@ -3,11 +3,11 @@ Package.describe({
   summary: 'Compiler for CoffeeScript code, supporting the coffeescript package',
   // This version of NPM `coffeescript` module, with _1, _2 etc.
   // If you change this, make sure to also update ../coffeescript/package.js to match.
-  version: '1.12.7_3'
+  version: '2.0.1_1'
 });
 
 Npm.depends({
-  'coffeescript': '1.12.7',
+  'coffeescript': '2.0.1',
   'source-map': '0.5.7'
 });
 
--- a/packages/non-core/coffeescript-compiler/package.js
+++ b/packages/non-core/coffeescript-compiler/package.js
@@ -13,11 +13,11 @@ Package.describe({
   summary: 'Compiler for CoffeeScript code, supporting the coffeescript package',
   // This version of NPM `coffeescript` module, with _1, _2 etc.
   // If you change this, make sure to also update ../coffeescript/package.js to match.
-  version: '2.0.1_1'
+  version: '2.0.2_1'
 });
 
 Npm.depends({
-  'coffeescript': '2.0.1',
+  'coffeescript': '2.0.2',
   'source-map': '0.5.7'
 });
 
--- a/packages/non-core/coffeescript-compiler/package.js
+++ b/packages/non-core/coffeescript-compiler/package.js
@@ -1,3 +1,13 @@
+// The NPM `coffeescript` module requires Node 6+; but instead of checking for
+// a Node runtime version, detect support for async functions, which were
+// added in Node 7.6.
+try {
+  new Function('async () => {}')();
+} catch (exception) {
+  throw new Error('Your runtime does not support this version of CoffeeScript. Please upgrade to Meteor 1.6 or later, or use a 1.x version of CoffeeScript.');
+}
+
+
 Package.describe({
   name: 'coffeescript-compiler',
   summary: 'Compiler for CoffeeScript code, supporting the coffeescript package',
--- a/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
+++ b/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "coffeescript": {
-      "version": "1.12.7",
-      "resolved": "https://registry.npmjs.org/coffeescript/-/coffeescript-1.12.7.tgz",
-      "integrity": "sha512-pLXHFxQMPklVoEekowk8b3erNynC+DVJzChxS/LCBBgR6/8AJkHivkm//zbowcfc7BTCAjryuhx6gPqPRfsFoA=="
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/coffeescript/-/coffeescript-2.0.1.tgz",
+      "integrity": "sha512-mrgXUOwRocrvSRwIi/2tAesQOJz2KxSyMNMlLSVhnJk/eZHKinYSJGcgUYZkzmSQZtpb3UhGJeZv0AAUeh7yyQ=="
     },
     "source-map": {
       "version": "0.5.7",
--- a/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
+++ b/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "coffeescript": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/coffeescript/-/coffeescript-2.0.1.tgz",
-      "integrity": "sha512-mrgXUOwRocrvSRwIi/2tAesQOJz2KxSyMNMlLSVhnJk/eZHKinYSJGcgUYZkzmSQZtpb3UhGJeZv0AAUeh7yyQ=="
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/coffeescript/-/coffeescript-2.0.2.tgz",
+      "integrity": "sha512-/bCMyzu7KSJPF2gRNYWpbEmfPkNL8AzXs78ktxhPTpSXlKetZRl7kIYGqU055UqUVnkKRJK4eUkUhRHQpvdilA=="
     },
     "source-map": {
       "version": "0.5.7",
--- a/packages/non-core/coffeescript-compiler/coffeescript-compiler.js
+++ b/packages/non-core/coffeescript-compiler/coffeescript-compiler.js
@@ -22,7 +22,10 @@ export class CoffeeScriptCompiler {
       // prevents CoffeeScript projects from using the modules package and
       // putting require or import statements within backticks; it just
       // won't happen automatically because of Babel.
-      runtime: false
+      runtime: false,
+      // CoffeeScript 2 supports for JSX, which Meteor supports only for React,
+      // per packages/ecmascript/plugin.js.
+      react: true
     });
   }
 
--- a/packages/non-core/coffeescript-test-helper/exporting.coffee
+++ b/packages/non-core/coffeescript-test-helper/exporting.coffee
@@ -1,14 +1,7 @@
 COFFEESCRIPT_EXPORTED = 123
 COFFEESCRIPT_EXPORTED_ONE_MORE = 234
 
-# Having backticks in the code is required to trigger Babel processing
-# the file and re-wrapping the list of defined variables.
 `COFFEESCRIPT_EXPORTED_WITH_BACKTICKS = 345`
-
-# Defining a class which extends a new class forces CoffeeScript
-# to define an "extend" function, which then in turn forces Babel
-# to re-wrap the list of defined variables so that each is defined
-# in its own line.
 
 class TestClass
 
--- a/packages/non-core/coffeescript-test-helper/package.js
+++ b/packages/non-core/coffeescript-test-helper/package.js
@@ -1,10 +1,10 @@
 Package.describe({
   summary: "Used by the coffeescript package's tests",
-  version: "1.0.10-beta.30"
+  version: "1.1.0"
 });
 
 Package.onUse(function (api) {
-  api.use('coffeescript@1.12.7-2-beta.30', ['client', 'server']);
+  api.use('coffeescript@2.0.1_1', ['client', 'server']);
   api.export('COFFEESCRIPT_EXPORTED');
   api.export('COFFEESCRIPT_EXPORTED_ONE_MORE');
   api.export('COFFEESCRIPT_EXPORTED_WITH_BACKTICKS');
--- a/packages/non-core/coffeescript-test-helper/package.js
+++ b/packages/non-core/coffeescript-test-helper/package.js
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
-  api.use('coffeescript@2.0.1_1', ['client', 'server']);
+  api.use('coffeescript@2.0.2_1', ['client', 'server']);
   api.export('COFFEESCRIPT_EXPORTED');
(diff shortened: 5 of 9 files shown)


benjamn:
Wooooooo!!!! 🎉🎉🎉 ☕️ 📜 💥💥

GeoffreyBooth:
Do the CI builds include building and running the tests for the `todos` app? Including the `react` and `coffeescript` branches? Might be a good idea to add.

For this PR, I want to manually test that it works in `todos`, including source maps; and add a test that something new in CS2 works. CoffeeScript 2 adds support for JSX; is `babel-compiler` already configured to automatically transform JSX?

It would be nice if there were some way to not depend on a pinned version of the NPM `coffeescript` module, so that this wouldn't need updating on every upstream version bump.

bjaskj:
Bump, please get this in so we can use 2.0 👍

benjamn:
For the record: this change is slated for Meteor 1.6.1, so that everyone can update to Meteor 1.6 with no hiccups related to `coffeescript`, `coffeescript-compiler`, or `babel-compiler` versions, and then update to 1.6.1 any time after that, at their convenience. Hopefully by then Babel 7 will be out of beta, too.

GeoffreyBooth:
@benjamn please give https://github.com/meteor/meteor/issues/9176 a gander when you get a moment, it’s holding up me finishing this.

GeoffreyBooth:
So in `packages/ecmascript/plugin.js`, I see this:

```js
Plugin.registerCompiler({
  extensions: ['js', 'jsx'],
}, function () {
  return new BabelCompiler({
    react: true
  });
});
```

Which I guess means that Meteor’s Babel always has the `react` transform enabled, which means that JSX is always compiled into `React.createElement` code. Now that CoffeeScript 2 supports JSX, the Meteor `coffeescript-compiler` package will need to do the same thing (8f3ca22abe536c8b8d25010f51bea136dbde263f). But [JSX is now also supported by Vue](https://vuejs.org/v2/guide/render-function.html), and who knows what other frameworks may support it in the future; how will this be addressed? Or is the plan to just leave non-React JSX unsupported for now, until it becomes popular enough in other frameworks to justify what I guess would be a breaking change?

GeoffreyBooth:
It looks like the build fails because Mocha requires CoffeeScript 1.0.4, or perhaps because it has `api.versionsFrom("1.3")`. Do you want to update that package?

```
  2|  => Errors while adding packages:
  2| 
  2| While selecting package versions:
  2| error: Conflict: Constraint coffeescript@1.0.4 is not satisfied by coffeescript 2.0.1_1.
  2| Constraints on package “coffeescript”:
  2| * coffeescript@=2.0.1_1 <- top level
  2| * coffeescript@1.0.4 <- practicalmeteor:mocha 2.0.1-rc0
  2| 
  2| Conflict: Constraint spacejamio:chai@1.9.2_2 is not satisfied by spacejamio:chai 1.0.0.
  2| Constraints on package “spacejamio:chai”:
  2| * spacejamio:chai@1.9.2_2 <- practicalmeteor:mocha 2.0.1-rc0
```

But more broadly, how do we want to deal with this in general? Lots of other packages depend on `coffeescript`. Updating the `todos` `coffeescript` branch to this PR’s 2.0.1_1 yields this mess:

```
=> Errors prevented startup:

   error: Conflict: Constraint coffeescript@1.0.4 is not satisfied by coffeescript 2.0.1_1.
   Constraints on package "coffeescript":
   * coffeescript@=2.0.1_1 <- top level
   * coffeescript@1.0.4 <- arillo:flow-router-helpers 0.5.2
   * coffeescript@1.0.4 <- zimme:active-route 2.3.2
   * coffeescript@1.0.4 <- tap:i18n 1.8.2
   * coffeescript@1.0.4 <- softwarerero:accounts-t9n 1.3.11
   * coffeescript@1.0.3 <- practicalmeteor:chai 2.1.0_1
   * coffeescript@1.0.3 <- practicalmeteor:sinon 1.14.1_2
   * coffeescript@1.0.15 <- practicalmeteor:mocha 2.4.5_6
   * coffeescript@1.0.3 <- practicalmeteor:loglevel 1.2.0_2 <- practicalmeteor:mocha 2.4.5_6

   Conflict: Constraint coffeescript@1.0.3 is not satisfied by coffeescript 2.0.1_1.
   Constraints on package "coffeescript":
   * coffeescript@=2.0.1_1 <- top level
   * coffeescript@1.0.4 <- arillo:flow-router-helpers 0.5.2
   * coffeescript@1.0.4 <- zimme:active-route 2.3.2
   * coffeescript@1.0.4 <- tap:i18n 1.8.2
   * coffeescript@1.0.4 <- softwarerero:accounts-t9n 1.3.11
   * coffeescript@1.0.3 <- practicalmeteor:chai 2.1.0_1
   * coffeescript@1.0.3 <- practicalmeteor:sinon 1.14.1_2
   * coffeescript@1.0.15 <- practicalmeteor:mocha 2.4.5_6
   * coffeescript@1.0.3 <- practicalmeteor:loglevel 1.2.0_2 <- practicalmeteor:mocha 2.4.5_6

   Conflict: Constraint coffeescript@1.0.15 is not satisfied by coffeescript 2.0.1_1.
   Constraints on package "coffeescript":
   * coffeescript@=2.0.1_1 <- top level
   * coffeescript@1.0.4 <- arillo:flow-router-helpers 0.5.2
   * coffeescript@1.0.4 <- zimme:active-route 2.3.2
   * coffeescript@1.0.4 <- tap:i18n 1.8.2
   * coffeescript@1.0.4 <- softwarerero:accounts-t9n 1.3.11
   * coffeescript@1.0.3 <- practicalmeteor:chai 2.1.0_1
   * coffeescript@1.0.3 <- practicalmeteor:sinon 1.14.1_2
   * coffeescript@1.0.15 <- practicalmeteor:mocha 2.4.5_6
   * coffeescript@1.0.3 <- practicalmeteor:loglevel 1.2.0_2 <- practicalmeteor:mocha 2.4.5_6
```

That’s a lot of pull requests, and some of these packages haven’t been updated in years. Should we perhaps keep `coffeescript` at a 1.x version, and bump only `coffeescript-compiler` to 2.0.1_1?

benjamn:
> Should we perhaps keep `coffeescript` at a 1.x version, and bump only `coffeescript-compiler` to 2.0.1_1?

@GeoffreyBooth That does sound easier than getting all the `practicalmeteor:*` packages updated. Of course developers could clone those packages locally and update the version constraints, but that's also a lot of work (for even more people).

GeoffreyBooth:
@benjamn Okay, I’ve changed the version.

So what do you think about the JSX question above, and of https://github.com/meteor/meteor/pull/9107/commits/117ee234bd9e7ef1fd4ea9fcff3d9c0bb8a49581 as a way of feature detecting if the runtime will support CoffeeScript 2?

GeoffreyBooth:
So maybe we do need to submit pull requests for those packages:

```
  1| [[[[[ Tests ]]]]]
  1| 
  1| => Started proxy.
  1| => Errors prevented startup:
  1| 
  1| While building for web.browser:
  1| packages/practicalmeteor:mocha/meteor/src/lib/MochaRunner.coffee:27:7: Can't reference 'this' before calling super in derived class constructors
  1| packages/practicalmeteor:mocha/meteor/src/reporters/HtmlReporter.coffee:13:7: Can't reference 'this' before calling super in derived class constructors
  1| 
  1| While building for os.linux.x86_64:
  1| packages/practicalmeteor:mocha/meteor/src/lib/MochaRunner.coffee:27:7: Can't reference 'this' before calling super in derived class constructors
  1| packages/practicalmeteor:mocha/meteor/src/reporters/HtmlReporter.coffee:13:7: Can't reference 'this' before calling super in derived class constructors
  1| 
  1| => Your application has errors. Waiting for file change.
```

Or is there some way for that package to depend on one version of `coffeescript`, while the overall project uses another?

GeoffreyBooth:
So to get these tests to pass, either `practicalmeteor` accepts my pull request, or I’ll need to publish my patched version of `practicalmeteor:mocha` to Atmosphere. One of Meteor’s tests actually tests running `meteor add practicalmeteor:mocha` and it appears there’s no way to `meteor add` a Git repo, like `meteor add https://github.com/GeoffreyBooth/meteor-mocha.git` (though that would be a nice feature for you to add!).

But I don’t want to take ownership of `meteor-mocha`, nor do I really want to create a new Atmosphere package for this. @benjamn would you consider forking `meteor-mocha` into the `meteor` organization? It’s pretty integral in Meteor’s own test suite, so it might make sense for MDG to maintain it.

benjamn:
Yep, I can do that tomorrow. Definitely not your responsibility!

GeoffreyBooth:
Thanks. If you don’t mind giving me write access, I’ll test my branch there. I haven’t really been able to test that it works since `meteor-mocha`‘s own tests aren’t currently passing.

hwillson:
Just a heads up - to do this right, forking and maintaining a copy of `practicalmeteor:mocha` would also mean forking and  maintaining copies of its `practicalmeteor` dependencies:

- `practicalmeteor:sinon`
- `practicalmeteor:chai`
- `practicalmeteor:loglevel`
- `practicalmeteor:mocha-core`

Given that a lot of these packages were created before Meteor had full npm support, we'll likely want to re-visit how they're designed/implemented (since currently they're including their own copies of the referenced 3rd party libraries).

GeoffreyBooth:
@hwillson I would support that; but do we want to fork them first, and refactor them afterward?

GeoffreyBooth:
@benjamn @hwillson Looks like forking practicalmeteor:mocha won’t be necessary. There’s a newer package, [meteortesting:meteor-mocha](https://github.com/meteortesting/meteor-mocha), that’s actively maintained. But even better than that, there’s a noop testing driver package, [tmeasday:acceptance-test-driver](https://github.com/tmeasday/acceptance-test-driver), that I think is a better fit for these tests. The tests in question aren’t actually running any Mocha tests, they’re just using Mocha as a placeholder test driver.

hwillson:
Wow, that's great @GeoffreyBooth - I had no idea that package existed. Great idea!

benjamn:
Ah yes, the `meteortesting:*` packages used to be published under the username `dispatch`, which is what we use for some of our more recent self-tests (e.g., [`dynamic-import`](https://github.com/meteor/meteor/blob/1389db463184fcccbded21eec2ba2c61c6d16d91/tools/tests/dynamic-import.js#L33)).

GeoffreyBooth:
So using `tmeasday:acceptance-test-driver` gets the Meteor CI build to pass. In `todos` I replaced `practicalmeteor:mocha` with `meteortesting:mocha` and the tests still passed; but after I updated to Meteor 1.6-rc.15 and `coffeescript@2.0.1_1` I found a bunch more packages that want older versions of CoffeeScript:

```
Conflict: Constraint coffeescript@1.0.3 is not satisfied by coffeescript 2.0.1_1.
Constraints on package "coffeescript":
* coffeescript@=2.0.1_1 <- top level
* coffeescript@1.0.4 <- arillo:flow-router-helpers 0.5.2
* coffeescript@1.0.4 <- zimme:active-route 2.3.2
* coffeescript@1.0.4 <- tap:i18n 1.8.2
* coffeescript@1.0.4 <- softwarerero:accounts-t9n 1.3.9
* coffeescript@1.0.3 <- practicalmeteor:chai 2.1.0_1
* coffeescript@1.0.3 <- practicalmeteor:sinon 1.14.1_2
```

So I guess I can go through these one by one and find replacements, or fork/patch them . . . but this illustrates how awful this experience will be for anyone else who wants to start using `coffeescript@2`. I can only imagine the carnage when you someday bump a major version of `ecmascript`.

GeoffreyBooth:
Okay, I’ve made a fork of the `todos` `coffeescript` branch that uses Meteor 1.6-rc.16 and resolves the package constraint issues: https://github.com/GeoffreyBooth/todos-coffeescript/tree/coffeescript-meteor-1.6

This is a pretty thorough test that the new `coffeescript@2.0.1_1` package works, so I think this PR is ready to merge in now.

In order to get `todos` working with `coffeescript@2.0.1_1`, I was able to replace some packages with equivalents:

- `practicalmeteor:mocha` was replaced by `meteortesting:mocha`
- `practicalmeteor:chai` was replaced by the NPM module `chai`
- `practicalmeteor:sinon` was replaced by the NPM module `sinon`
- `tap:i18n` was replaced by `universe:i18n` and `universe:i18n-blaze`, with some refactoring

But then a bunch of packages required forking and patching, and you’ll see them in that branch’s `packages` folder:

- `arillo:flow-router-helpers`
- `hwillson:stub-collections`
- `softwarerero:accounts-t9n`
- `useraccounts:core`
- `zimme:activeroute` (I tried upgrading to the newer `activeroute:` packages, but they’re [not ready for prime time](https://github.com/meteor-activeroute/flow-router/issues/3))

And this is just for `todos`, which is hardly the most complex app. I feel like there will be a lot of complaints from people going through a similar process for their own apps when they try to upgrade `coffeescript`. You might want to implement some kind of smoother process for them, or shortcuts like #9244.

GeoffreyBooth:
If you’re looking for a minimal-impact solution to making the `coffeescript` upgrade easier for people, perhaps revise `api.versionsFrom` to only apply to packages that are _currently_ core packages in the app’s selected release of Meteor? So that something like `api.versionsFrom('1.3')` in a Meteor 1.6 app would pull the 1.3-era versions of all the core packages that are common between 1.3 and 1.6; so `ecmascript` would be the 1.3 version, but `coffeescript` would be whatever version the user has installed. In other words, `versionsFrom` just wouldn’t apply to `coffeescript` anymore, since `coffeescript` isn’t a core package anymore.

This solution only solves this one specific case, really, but this issue _is_ rather specific to `coffeescript` as there aren’t other core packages or formerly-core packages that are getting major version bumps; and `versionsFrom` only applies to core packages. At least in the packages I’ve patched, most `api.use` lines are versionless, relying on `versionsFrom` for version constraints. Changing `versionFrom` might be enough for this case.

benjamn:
@GeoffreyBooth I appreciate the narrowness of that idea, though we are also considering the consequences of deprecating `api.versionsFrom`, since it seems like that would simply require package authors to choose version constraints for any core packages they depend on (the next time they publish), which doesn't seem like that much of an imposition.

GeoffreyBooth:
Deprecating `api.versionsFrom` works too.

Any reason not to merge in this PR? I don’t think you should publish the new `coffeescript` to Atmosphere until after 1.6 is released, but are there any more concerns with this PR?

packages/non-core/coffeescript-compiler/package.js
```diff
@@ -1,3 +1,13 @@
+// The NPM `coffeescript` module requires Node 6+; but instead of checking for
+// a Node runtime version, detect support for async functions, which were
+// added in Node 7.6.
+try {
+  new Function('async () => {}')();
+} catch (exception) {
+  throw new Error('Your runtime does not support this version of CoffeeScript. Please upgrade to Meteor 1.6 or later, or use a 1.x version of CoffeeScript.');
+}
```
benjamn: Nice!

Issue #9026, opened 2017-08-18
[1.6b23] tests fails on startup: MongoError: not master and slaveOk=false


hexsprite:
Some important notes:
- only happens during tests, not normal app running

Getting "MongoError: not master and slaveOk=false" since updating to beta.18. I even tried adding the mongo-dev-server package but still the same. Fine if rollback to 17.  Also present in b23.

I've tried this in `devel` branch and can reliably avoid the problem by adding a 2000ms sleep after the "Started MongoDB" message is printed. So it seems like a possible race condition of some sort where Meteor thinks MongoDB is ready but it really isn't.

When I have more spare time I'll try to put together a repro but perhaps this may have given you some clues.

```
MongoError: not master and slaveOk=false
    at Function.MongoError.create (/Users/jbb/.meteor/packages/npm-mongo/.2.2.30.fp9dlr.i1hfs++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/error.js:31:11)
    at queryCallback (/Users/jbb/.meteor/packages/npm-mongo/.2.2.30.fp9dlr.i1hfs++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/cursor.js:212:36)
    at /Users/jbb/.meteor/packages/npm-mongo/.2.2.30.fp9dlr.i1hfs++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/connection/pool.js:469:18
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

abernix:
I haven't seen this problem so far in Meteor 1.6-beta.* so a reproduction of sorts would be most helpful as I'm not sure how else to debug your problem.  Do you have your own `MONGO_URL` setup?

I might suggest running Meteor from a checkout and adding some debug statements in the [`initiateReplSetAndWaitForReady` method of `run-mongo.js`](https://github.com/meteor/meteor/blob/release-1.5.2/tools/runners/run-mongo.js#L585), which is where we decide that Mongo is "ready"!

hexsprite:
so frustrating, it doesn't seem to fail from a checkout, only from a release

hexsprite:
no custom `MONGO_URL`

hexsprite:
Problem is still in b24. I managed to capture the Mongo logs and everything seems to be the same except that in the failure situation it seems to open 2 connections instead of 7 on success.

[...]

PR #9119, opened 2017-09-20
When starting mongod, yield until the heartbeat has pulsed after election.

abernix:
> Prologue: A heartbeat is used amongst members of a MongoDB replica set
to poll the status of said members.

When we are initiating a new replicaset for the test Mongo server, the replicaset is not fully prepared to accept writes until the voting members have negotiated and propagated their decision about who is the "primary" to all members involved.  This seems to be delayed by almost _exactly_ the default heartbeat interval, which is 2000ms.

The heartbeat interval is marked as an "internal only" property in Mongo so I was hesitant to lower it.  It's also a new property in Mongo 3.2 which might explain why this cropped up a while ago.

I believe this heartbeat delay is the only explanation for why the `rs.status()` (i.e. `replSetGetStatus`) believes it is ready before the `mongod` has actually printed "transition to primary complete" to the log (which was the sole, previous measurement we searched for).

It's not clear to me if watching that "transition to primary complete" message in the log is super relevant anymore, but it does indicate "readiness" which is what https://github.com/meteor/meteor/blob/4f43008aa0b02db79eb791d9103d696288a5bfb9/tools/runners/run-mongo.js#L546-L549 was meant to detect. 

Existing, surrounding code could use refactoring to match up with more modern Mongo versions, but that's beyond the scope of this more simple fix.

While I'm confident in this fix, it's worth pointing out (to assuage any would-be fears) that any fragility in this detection is isolated to the development Mongo server that Meteor starts, and not a production application (where your Mongo server is not managed automatically by Meteor, but rather an outside database provider).

Fixes meteor/meteor#9026.

### DIFF
--- a/tools/runners/run-mongo.js
+++ b/tools/runners/run-mongo.js
@@ -651,6 +651,8 @@ var launchMongo = function (options) {
         return;
       }
 
+      let wasJustSecondary = false;
+
       // XXX timeout eventually?
       while (!stopped) {
         var status = yieldingMethod(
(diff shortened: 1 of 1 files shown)


defionscode:
did this get into 1.6?

abernix:
@defionscode Yes, it should be included in Meteor 1.6 and 1.5.3.

FYI: This answer can _usually_, though not _always_, be found by viewing which tags the merge commit is present on.  In this case, clicking on https://github.com/meteor/meteor/commit/d50b83cc477630dc02d66a80f1b31ff17a79c28d above and seeing the `release/METEOR@*` tags which it has on it.  In this case, it shows `release/METEOR@1.6`, so the answer is "yes".

PR #9152, opened 2017-09-28
allow-deny: add $setOnInsert to ALLOWED_UPDATE_OPERATIONS

### LINKED ISSUE #8873: [1.5.1-rc.3] $setOnInsert not allowed in allow-deny package ALLOWED_UPDATE_OPERATIONS
According to the comment for `ALLOWED_UPDATE_OPERATIONS` I believe the intent is to keep up to date with Mongo with regards to allowing all simple updates and not those that `does more than just modify its target`.

https://github.com/meteor/meteor/blob/master/packages/allow-deny/allow-deny.js#L354

It raises the exception:

`update failed: Access denied. Operator $setOnInsert not allowed in a restricted collection.`

I'm not sure why it was not supported before, perhaps because it was not supported in MiniMongo until 1.3.x?

But $setOnInsert has been part of Mongo since 2.4

It's causing a bug with simpl-schema npm: https://github.com/aldeed/node-simple-schema/issues/146

hexsprite:
Fixes #8873


hwillson:
Thanks for submitting this PR @hexsprite! There is a bit of an on-going discussion happening around `allow-deny` right now (in-light of the [recent security issue](https://blog.meteor.com/meteor-allow-deny-vulnerability-disclosure-baf398f47b25)). The future of `allow-deny` is currently being questioned, so while your changes look great, this PR might be held up for a bit. This will be discussed further in a meeting today, so I'll have more details to share shortly. Thanks!

hwillson:
Hi @hexsprite - we just finished talking about this; we've decided to no longer accept new changes to the `allow-deny` package, and will be actively pushing towards deprecating its use entirely (this includes internally within the Meteor core). New feature requests will be created shortly outlining the deprecation plan. Thanks again for submitting this though!

hexsprite:
Really? will Meteor no longer support the use case of allowing clients to directly insert/update/delete collections ? I can understand the security problems but it's great for prototyping.

hwillson:
@hexsprite I agree, `allow-deny` is definitely great for prototyping. This is still being discussed and the direction could change, but for now the plan is to deprecate it in its current state. That's not to say it won't keep working, or that there might be a better alternative provided. It's still too early to tell but the potential (and understandable) mis-use of the current `allow-deny` package is just to high to leave things as they are. That being said, nothing is changing yet - we're just putting a halt on `allow-deny` feature changes to make sure Meteor isn't promoting its use by showing active new feature development. Bugs will of course still be fixed 🙂. Thanks!

hexsprite:
I hear ya. It sounds like it was scary to have the security issues crop up with allow-deny. Thing is, I do consider this PR to be a bugfix since the intention of the ALLOWED_UPDATE_OPERATIONS is clearly meant to include all operations that update a single object. The bug is that it was not updated to keep up to date with MongoDB changes.  So I wonder if you could reconsider this on this basis? Otherwise I need to keep patching it which is a PITA.  This bug also affects simpl-schema which is a pretty major package.

abernix:
We'll continue to think about the future of `allow-deny`.

There is validity of this being a sensitive topic due to timing, but we've also had messaging which has been discouraging the use of `allow-deny` for a long time.  That messaging has now been further reinforced by the recent `allow-deny` disclosure.  By putting out another update for it ("Developers can now use `$setOnInsert`!"), we could be perpetuating the idea that its use is still being encouraged.  Furthermore, adding an additional update operator to the list of allowed operators requires developers who may not have considered disallowing the operator in the past (since it was forbidden at a higher level) to go through and re-evaluate their already-written rules.  This means it must be a major version number change and come with a disclaimer, which could further confuse/scare some developers.

I do believe that `$setOnInsert` may have been an oversight, or it may not have been available from Mongo when this functionality was originally written (I'm not sure), but I'm not sure if the fact that it is excluded can be classified as a bug.  Bug or feature, there doesn't seem to have been much demand for this.  You are, of course, welcome to fork the package into your `packages` directory with the change yourself, so long as you've made the above considerations!

PR #9170, opened 2017-10-02
Switch to bright colors on Windows Powershell

### LINKED ISSUE #9118: Error output is  incomplete in Windows PowerShell
Since the Windows 10 anniversary update the standard command prompt has been removed from the context menu and is now replaced by PowerShell so Windows users are encourage to use PowerShell instead of command prompt. First, as related in issue #4133, the blue color is hard to see. Second,
errors are not shown with details as it is in command prompt.

cmd

As you can see, PowerShell on the left only displays `(STDERR)` instead of the complete error shown in the command prompt on the right.

I am using Meteor 1.5.2 (but I've noticed this problem for a long time now).

agathver:
Fixes #9118

Uses [bright variant](https://www.npmjs.com/package/cli-color#bright-variants) of colors from cli-color.

CMD and windows Powershell, need bright colors for propercontrast difference for comfortable reading.

### DIFF
--- a/packages/logging/logging.js
+++ b/packages/logging/logging.js
@@ -49,6 +49,18 @@ var LEVEL_COLORS = {
 
 var META_COLOR = 'blue';
 
+// Default colors cause readability problems on Windows Powershell,
+// switch to bright variants.
+if ('win32' === process.platform) {
+  LEVEL_COLORS = {
+    debug: 'greenBright',
+    warn: 'magentaBright',
+    error: 'redBright'
+  };
+
+  META_COLOR = 'blueBright';
+}
+
 // XXX package
 var RESTRICTED_KEYS = ['time', 'timeInexact', 'level', 'file', 'line',
                         'program', 'originApp', 'satellite', 'stderr'];
--- a/packages/logging/logging.js
+++ b/packages/logging/logging.js
@@ -50,16 +50,16 @@ var LEVEL_COLORS = {
 var META_COLOR = 'blue';
 
 // Default colors cause readability problems on Windows Powershell,
-// switch to bright variants.
-if ('win32' === process.platform) {
-  LEVEL_COLORS = {
-    debug: 'greenBright',
-    warn: 'magentaBright',
-    error: 'redBright'
-  };
-
-  META_COLOR = 'blueBright';
-}
+// switch to bright variants. While still capable of millions of
+// operations per second, the benchmark showed a 25%+ increase in
+// ops per second (on Node 8) by caching "process.platform".
+var platform = process.platform;
+var platformColor = function (color) {
+  if (platform === 'win32') {
+    return color + 'Bright';
+  }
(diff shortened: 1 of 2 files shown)


agathver:
@abernix Regarding 03aa6d2, what if options.metaColor is passed a bright color altogether ? I currently dont see this in code, but if in future some tool is added which does pass a bright color ?

packages/logging/logging.js
```diff
@@ -55,7 +55,7 @@ var META_COLOR = 'blue';
 // ops per second (on Node 8) by caching "process.platform".
 var platform = process.platform;
 var platformColor = function (color) {
-  if (platform === 'win32') {
+  if (platform === 'win32' && !color.endsWith('Bright')) {
```
abernix: This is might be a good idea since the `logging` package could be used by outside packages and apps though I'd originally avoided it because it cut the ops per second in half over the original implementation.  It's possible I was being overzealous with optimizations, since even with this change `platformColor` still achieves 15MM ops/sec, but this is a logging function and I didn't want to introduce too much additional friction.

agathver: 15M ops/sec is still huge!

abernix:
I've dismissed my own review as it appears that there are some additional considerations to he had, and I'm mid-workflow at the moment.  Namely:

* The `logging` package is also imported [on the client](https://github.com/meteor/meteor/blob/devel/packages/logging/package.js#L21), so we can't assume `process.platform` is available.
* Since `String.prototype.endsWith` is an ECMAScript 2015 feature, it won't be available on some older clients and this package doesn't use `ecmascript`. I think it's fine to add `ecmascript` though, and it is available natively on Node.js v4+.

agathver:
@abernix 

> The logging package is also imported on the client, so we can't assume process.platform is available.

Fixed it

> Since String.prototype.endsWith is an ECMAScript 2015 feature, it won't be available on some older clients and this package doesn't use ecmascript. I think it's fine to add ecmascript though, and it is available natively on Node.js v4+.

Switched to String.prototype.slice, which is from ES-3

packages/logging/logging.js
```diff
@@ -49,6 +49,18 @@ var LEVEL_COLORS = {
 
 var META_COLOR = 'blue';
 
+// Default colors cause readability problems on Windows Powershell,
+// switch to bright variants. While still capable of millions of
+// operations per second, the benchmark showed a 25%+ increase in
+// ops per second (on Node 8) by caching "process.platform".
+var isWin32 = process && process.platform === 'win32';
```
benjamn: Unfortunately, just referring to `process` when it isn't defined will throw a `ReferenceError`, so this needs to be
```js
var isWin32 = typeof process === "object" && process.platform === "win32";
```

benjamn:
Thanks for working on this @agathver (and @abernix)!

Issue #9166, opened 2017-10-02
[v1.6-rc.1] Debugger source files are not mapped to the downloaded meteor files (e.g. app.js)


brucejo75:
I ported my app over to `1.6-rc1`.  I was really looking forward to using the marquee debug feature.

I cannot figure out how to get source maps working properly.  I hope that I am missing some setting or something...

**Any/All help is much appreciated!**

### Environment
- Windows 10 1703 (OS Build 15063.608)
- Meteor 1.6-rc.1

### Repro
- I read the Medium [article](https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27).
- I read the Node [article](https://nodejs.org/en/docs/inspector/).
- I made sure my [source map settings were right](https://developers.google.com/web/tools/chrome-devtools/javascript/source-maps).

1. `meteor create test`
2. `cd test`
3. `meteor debug`
4. From Chrome go to url: `chrome://inspect`, click on "Open dedicated DevTools for Node"
See Chrome inspector come up, looking like it is running. (note Pause button does not work)
5. Wait 5 Minutes or more. #9165
6. Message in console: "Debugger did not attach after 5 minutes; continuing."

**Result:** Now I can debug, but...
1. Similar to #8590, None of my source code files are mapped in Chrome.
2. I can see all my code in `.meteor\local\build\programs\server\app\app.js` but nothing is mapped to my source code tree. (see picture)
3. The source code tree has 100's of files from my build bundle and then finally from my project build and then a bunch of other files.  But none of my source files. (see picture)
4. Chrome indicates that a "Source Map detected" and claims "associated files should be added to the file tree" but they do not appear there?  (see picture)

![image](https://user-images.githubusercontent.com/14342237/31060142-e8d8cd1e-a6c2-11e7-92ed-4bb140f31b74.png)

benjamn:
This was "not something that we can actually fix at this time" according to the Node folks last year: https://github.com/nodejs/node/issues/8369

Here's hoping there's anything we can do about it in October 2017. 🤞 

To be clear, this is definitely a Windows-specific problem, as server-side inspector source maps work just fine on other platforms.

brucejo75:
Hi @benjamn, that is unfortunate :confused:.

I spotted this [workaround](https://github.com/nodejs/node/issues/10838#issuecomment-321266341).  I will give it a shot on a build of a meteor project and see if it enables the source maps.

[...]

PR #9172, opened 2017-10-03
Detect inspector client attachment by timing debugger keyword.

### LINKED ISSUE #9165: [v1.6-rc.1] Debugger did not attach after 5 minutes; continuing. (windows)
Chrome debugger does not immediately connect to running meteor node process.  After 5 minutes the DevTools for node will show the following error: `Debugger did not attach after 5 minutes; continuing.`

### Environment
- Windows 10 1703 (OS Build 15063.608)
- Meteor 1.6-rc.1

### Steps to repro
1. start meteor: `meteor debug`
2. From Chrome go to url: `chrome://inspect`, click on "Open dedicated DevTools for Node"
3. See Chrome inspector come up, looking like it is running. (note Pause button does not work)
4. Wait 5 Minutes or more.

**Result:** message in console: "Debugger did not attach after 5 minutes; continuing."

### Notes:
- I had put some breakpoints in `app.js` and on continuing the debugger paused at the breakpoints.
- It looks like I can debug.  (without source mapping to my source files)
- On server command window the same error shows up and implies that debugging is not happening, but it is:
```
20171001-17:48:36.461(-7)? (STDERR) Debugger did not attach after 5 minutes; continuing.
```
- If I kill the DevTools for Node then the following displays on the command window:
```
W20171001-17:55:35.858(-7)? (STDERR) Debugger listening on ws://127.0.0.1:9229/e1ea2583-9a64-47d0-bf28-7028886eab3d
W20171001-17:55:35.909(-7)? (STDERR) For help see https://nodejs.org/en/docs/inspector
W20171001-17:55:35.912(-7)? (STDERR) (node:3872) [INSPECTOR_ASYNC_STACK_TRACES_NOT_AVAILABLE] Warning: Warning: Async stack traces in debugger are not available on 32bit platforms. The feature is disabled.
W20171001-17:59:52.572(-7)? (STDERR) Debugger attached.
W20171001-17:59:52.574(-7)? (STDERR) Debugger listening on ws://127.0.0.1:9229/e1ea2583-9a64-47d0-bf28-7028886eab3d
W20171001-17:59:52.577(-7)? (STDERR) For help see https://nodejs.org/en/docs/inspector
```

I think the debug mode is a little confused.

benjamn:
### Background

The `meteor debug` command behaves like Node's `--inspect-brk` flag, in that it attempts to pause the server before executing any server code.

However, simply passing the `--inspect-brk` flag to Node causes execution to pause on the very first line of code, which is not good for setting any breakpoints, because no server code has actually loaded yet.

Instead, the `meteor debug` command uses Node's `--inspect` flag to enable debugging without an initial pause, then manually pauses at an appropriate moment during server startup. Ideally, the pause should last until an inspector client has been attached to the process, at which point the developer has a chance to set any desired breakpoints, then clicks the continue button to proceed with server startup.

### Problem

The most difficult part of this process is detecting when the inspector client has attached. Previously, the parent process listened for the child process to print a "Debugger attached" message to STDERR, which happens as a result of [this `fprintf` call in Native C++](https://github.com/nodejs/node/blob/7cff6e80bfd2f61ed67ff07af87af3ab63860273/src/inspector_io.cc#L396).

However, this message was not printed in some cases, especially on Windows (#9165), and required inter-process communication even in the ideal case. All of that logic is gone now, thanks to this commit.

### Solution

This commit takes advantage of a difference in behavior of the `debugger` keyword depending on whether or not an inspector client is attached. When no client is attached, the `debugger` keyword is a no-op that takes no time (or very little time) to execute. Once a client has attached, the `debugger` keyword triggers a breakpoint that lasts until the developer explicitly continues execution through the client UI. Needless to say, this makes the `debugger` keyword take longer than a no-op.

Because the `debugger` keyword does nothing until a client connects, we can safely poll a `pause` function containing a `debugger` keyword at a frequent interval (say, every 500ms). Once a client connects, the `debugger` keyword will become active, pausing the server at exactly the point we hoped. The difference is easy to detect by timing the `pause()` function call. Once the `debugger` keyword becomes active, we stop polling and allow server startup to continue.

Elegant!

### Bugs

Should fix #9165.
Related: #9166

Thanks to @brucejo75 for reporting these problems and inspiring this PR.

### DIFF
--- a/tools/runners/run-app.js
+++ b/tools/runners/run-app.js
@@ -99,13 +99,6 @@ _.extend(AppProcess.prototype, {
     });
 
     eachline(self.proc.stderr, function (line) {
-      if (self.debugPort &&
-          line.indexOf("Debugger attached") >= 0) {
-        self.proc.send({
-          meteorDebugCommand: "continue"
-        });
-      }
-
       runLog.logAppOutput(line, true);
     });
 
--- a/tools/static-assets/server/boot.js
+++ b/tools/static-assets/server/boot.js
@@ -45,34 +45,60 @@ var parsedSourceMaps = {};
 const meteorDebugFuture =
   process.env.METEOR_INSPECT_BRK ? new Future : null;
 
-if (meteorDebugFuture) {
-  process.on("message", function onDebugMessage(msg) {
-    if (msg && msg.meteorDebugCommand === "continue") {
-      process.removeListener("message", onDebugMessage);
-      // After the "continue" message is received, the Chrome DevTools
-      // debugger still needs a small amount of time to get ready to pause
-      // at the debugger statement in ./debug.js.
-      setTimeout(() => meteorDebugFuture.return(true), 500);
-    }
-  });
-}
-
 function maybeWaitForDebuggerToAttach() {
   if (meteorDebugFuture) {
-    // This setTimeout not only puts a reasonable time limit on the
-    // debugger attaching, but also keeps the process alive by preventing
-    // the event loop from running empty while the Fiber yields.
-    const timer = setTimeout(() => {
-      console.error("Debugger did not attach after 5 minutes; continuing.");
-      meteorDebugFuture.return(false);
-    }, 5 * 60 * 1000);
+    const pause = require("./debug.js").pause;
+    const pauseThresholdMs = 50;
+    const pollIntervalMs = 500;
+    const waitStartTimeMs = +new Date;
+    const waitLimitMinutes = 5;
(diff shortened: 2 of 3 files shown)


benjamn:
Before I had this idea, I seriously considered starting the inspector on a different port and using a proxy server to intercept WebSocket connections from inspector clients. That would have required rewriting HTTP responses for the `/json` manifest file, and of course we'd have to continue proxying all HTTP/WS traffic, because you can't change the inspector port later. Scary to think that might have worked! 😨 👻 🎃

Issue #9182, opened 2017-10-05
Dynamic imports wrongly loads static import dependencies inside packages


SachaG:
When you use `import()` to dynamically import a file, you would expect that all of that file's static `import`s would also be loaded dynamically. In other words, if I load this file dynamically:

```js
import SuperHeavyJSPackage from 'super-heavy-js-package';
```

Then `SuperHeavyJSPackage` will *not* be part of my app bundle. 

While this works fine inside the `imports` directory, it seems like it doesn't inside the `packages` directory. What's more, apparently imports will run even when they shouldn't (for example, inside an `if (false) {...}` block). 

Here's a reproduction I created with help from @Discordius, who found the issue: 

https://github.com/SachaG/dynamic-imports-bug/tree/package

And here's the same code running out of `imports` to compare: 

https://github.com/SachaG/dynamic-imports-bug/tree/imports

It seems like the problem has been around since 1.5 as far as I can tell.

benjamn:
This is definitely a bug!

I don't expect this to mean much to anyone but me, but I've tracked the problem down to [this line](https://github.com/meteor/meteor/blob/255732630d16e2b393e85c5b77f613abcf37dd41/tools/isobuild/import-scanner.js#L898), which is executed when a Meteor package tries to import an npm package from the application `node_modules` directory. Specifically, the mistake seems to be that the `forDynamicImport` parameter is ignored, so we lose the crucial information that `HeavyComponent.jsx` was imported dynamically.

Discordius:
Just as a quick question so I can decide whether to implement a hot fix myself: Do you have any guess how long it will be until I might be able to pull a fix to this? (ideally via npm and not via forking the repo)

benjamn:
@Discordius You can try it now with Meteor 1.6: `meteor update --release 1.6-rc.6`. If you're not ready for that, we will probably back-port this to Meteor 1.5.3 (not yet released).

PR #9187, opened 2017-10-07
Refactor peer dependency processing and fix issue #9182.

benjamn:
Issue #9182, reported by @SachaG, was due to a relatively minor bug fixed in badb7a7e6c847dfbc09ac995b2bc06bdb98a9953. However, once that bug was fixed, it revealed another more serious bug that required refactoring the way we process peer dependencies (4ff0318dc3141c3b09a32a1831f86c8d1160e84b).

The other commits in this PR don't fix any bugs, but I think they improve code quality and possibly performance (3faee05eede683dd14fa922464ef2ba0c081a051 avoids harmlessly re-scanning some modules).

This PR needs additional tests, most likely in the `tools/tests/apps/dynamic-import` test app.


tools/isobuild/import-scanner.js
```diff
@@ -582,45 +613,7 @@ export default class ImportScanner {
...
-            // so any errors reported to InputFile#error were saved but
-            // not reported at compilation time. Now that we know the file
-            // has been imported, it's time to report those errors.
-            buildmessage.error(depFile.error.message,
-                               depFile.error.info);
-          } else {
-            this._scanFile(depFile, dynamic);
-          }
-        }
+        this._scanFile(depFile, dynamic);
```
benjamn: We can call `this._scanFile` unconditionally here because it will return immediately if the file has already been scanned, thanks to the simpler version of this logic hoisted to the top of the method.

tools/isobuild/import-scanner.js
```diff
@@ -646,7 +639,7 @@ export default class ImportScanner {
       depFile.installPath = installPath;
       depFile.servePath = installPath;
       depFile.lazy = true;
-      depFile.imported = dynamic ? "dynamic" : true;
+      depFile.imported = false;
```
benjamn: This is only temporary, since we're going to call `this._scanFile(depFile, dynamic)` below.

tools/isobuild/import-scanner.js
```diff
@@ -362,53 +362,56 @@ export default class ImportScanner {
...
-      this.allMissingNodeModules = newlyMissing;
+    if (! isEmpty(missingModules)) {
+      const previousAllMissingModules = this.allMissingModules;
+      this.allMissingModules = newlyMissing;
 
-      try {
-        this._scanFile({
+      Object.keys(missingModules).forEach(id => {
+        // TODO Call this._scanFile fewer times in case missingModules[id]
+        // is a long and redundant list?
```
benjamn: Fixed in a later commit.

tools/isobuild/import-scanner.js
```diff
@@ -362,53 +362,56 @@ export default class ImportScanner {
...
+      this.allMissingModules = previousAllMissingModules;
 
-        each(identifiers, (info, id) => {
-          if (! has(newlyMissing, id)) {
-            newlyAdded[id] = info;
-          }
-        });
+      Object.keys(missingModules).forEach(id => {
+        if (! has(newlyMissing, id)) {
+          newlyAdded[id] = missingModules[id];
```
benjamn: We don't need to be careful to use `ImportScanner.mergeMissing` here because this is the first time `newlyAdded[id]` has been assigned.

tools/isobuild/import-scanner.js
```diff
@@ -417,6 +420,30 @@ export default class ImportScanner {
     };
   }
 
+  static mergeMissing(target, source) {
```
benjamn: Note to self: this method needs a comment.

tools/isobuild/import-scanner.js
```diff
@@ -900,18 +927,16 @@ export default class ImportScanner {
...
         Object.create(null);
       missing[id] = info;
-      parentFile.missingNodeModules = missing;
+      parentFile.missingModules = missing;
     }
 
-    if (! has(this.allMissingNodeModules, id) ||
-        ! info.possiblySpurious) {
-      // Allow any non-spurious identifier to replace an existing
-      // possibly spurious identifier.
```
benjamn: Since we're now tracking all failed imports separately, we don't have to be careful about letting non-spurious identifiers take precedence, since multiple `info` objects can coexist for the same identifier.

benjamn:
Added a test for this: https://github.com/meteor/meteor/pull/8728/commits/7315c7ea596d2a62778fa3b1e4082044fd5c1fc0

PR #9190, opened 2017-10-08
CoffeeScript: fix circular dependencies

GeoffreyBooth:
See https://github.com/meteor/meteor/issues/9176#issuecomment-334942137

### DIFF
--- a/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
+++ b/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
@@ -6,9 +6,9 @@
       "from": "coffeescript@1.12.7"
     },
     "source-map": {
-      "version": "0.5.6",
-      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.6.tgz",
-      "from": "source-map@0.5.6"
+      "version": "0.6.1",
+      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
+      "from": "source-map@0.6.1"
     }
   }
 }
--- a/packages/non-core/coffeescript-compiler/coffeescript-compiler.js
+++ b/packages/non-core/coffeescript-compiler/coffeescript-compiler.js
@@ -41,7 +41,7 @@ export class CoffeeScriptCompiler {
   }
 
   outputFilePath(inputFile) {
-    return inputFile.getPathInPackage() + '.js';
+    return inputFile.getPathInPackage();
   }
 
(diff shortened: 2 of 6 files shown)


GeoffreyBooth:
Is there a way for me to see what failed? I click on Details but I don’t see anything useful in the output.

abernix:
@GeoffreyBooth Clicking "Details", to the far right on the (two, separate) failed groups in the "Some checks were not successful" section above should take you to the CircleCI page.  Scrolling down on that page should reveal the problem.

* CircleCI Group 2: https://circleci.com/gh/meteor/meteor/8439
  * In a section titled: `eval $PRE_TEST_COMMANDS; ./meteor self-test ...`
     The name should have been `Running self-test (2): Con-K`, but it looks like that was broken.
      Fixed group-name for future tests with 9e3945dc9a33ffa43af336792fe6a4d47a1d4baa.
  * Failures appear to be due to the missing `.js` extension.
* CircleCI Group 3: https://circleci.com/gh/meteor/meteor/8442
  * In a section titled `Running self-test (3): L-O`
  * Failure appears as `AssertionError: undefined === 'eager coffee'`.

GeoffreyBooth:
All I see is a loading spinner that never stops spinning:

image

abernix:
@GeoffreyBooth They're loaded dynamically with AJAX, and not too large, so you'll probably want to try another browser?  That's unfortunately not something I can help solve, and it does (and has always) worked properly for me.  Sorry!

I've copied them into this Gist for you: https://gist.github.com/abernix/a65b69cfd0685e31431f49d1fae1c351

Hope that helps!

GeoffreyBooth:
@abernix I looked in the JavaScript console and saw Ajax requests with 401 errors, which was odd since the user menu had `User settings` and `Log out` as the two options. When I clicked on `User settings` I was given an option to login via GitHub, and after I did that now I can finally see the output. Not the best UX, to say the least. You can probably see what I’m talking about if you open one of the links above in an incognito window.

Anyway based on your gist it looks like the tests that failed were entirely expected, as they were checking for `.coffee.js` extensions. I’ve updated them.

Do you think it’s necessary to adapt your reproduction into a test somehow, to ensure that circular dependencies don’t break in the future? Or is that not really necessary in light of this PR and #9214?

benjamn:
Rebased and pushed to `devel` as b2ec292fd2926823772ceac0d14d2ab179d97d06. Thanks very much @GeoffreyBooth!

benjamn:
Also just republished both `coffeescript-compiler` and `coffeescript` to Atmosphere.

Issue #9196, opened 2017-10-09
[v1.6] Mobile config missing option for new Cordova iOS App Store icon


markoshust:
The new "app store" 1024x1024 icon for Cordova is missing:
![screen shot 2017-10-05 at 1 10 44 pm](https://user-images.githubusercontent.com/437029/31240430-b031d276-a9ce-11e7-973a-2097a43db1f5.png)

I was able to get this new icon set by adding the following to `mobile-config.js`:
```
App.appendToConfig(`
  <platform name="ios">
    <icon height="1024" width="1024" src="../../../resources/splash/splash-1024x1024.png" />
  </platform>
`);
```

IMO this new icon should be added as a property of `App.icons`, ex: `ios_appstore`. This appears to be new in Xcode 9/iOS 11.

Without defining this icon, just building, archiving and submitting the build to iTunes Connect will result in Apple rejecting the build, warning that the "iOS App Store Icon" is missing. Referenced from https://github.com/meteor/meteor/pull/8728#issuecomment-334593305

janowsiany:
Could you confirm you are using [cordova-custom-config](https://github.com/dpa99c/cordova-custom-config)?

markoshust:
Affirmative, without that it doesn't work

skirunman:
We are having the same issue. iOS launch screens are also out of date. I'll submit a PR to fix.

PR #9198, opened 2017-10-09
Fix iOS icon and launch screen sizes for iOS 11

skirunman:
Fixes https://github.com/meteor/meteor/issues/9196
See Apple [app icon sizes](https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/app-icon/#app-icon-sizes) and [launch screen sizes](https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/launch-screen/).

I also added some comments to documentation for [mobile configuration](http://docs.meteor.com/api/mobile-config.html) so a user can understand what devices the icons and launch screens relate to, but not sure if this will build correctly or if you think it is a good idea.

### DIFF
--- a/tools/cordova/builder.js
+++ b/tools/cordova/builder.js
@@ -14,26 +14,28 @@ import { CORDOVA_ARCH } from './index.js';
 // Hard-coded size constants
 
 const iconsIosSizes = {
+  'app_store': '1024x1024',
   'iphone_2x': '120x120',
   'iphone_3x': '180x180',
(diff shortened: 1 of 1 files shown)


benjamn:
Thanks @skirunman (and @markoshust)!

AndyNormore:
A bunch of us coming over to say thank you! This is a solid fix.

andyrichardson:
Thanks for fixing this. Looking forward to it being pushed to stable.

PR #9201, opened 2017-10-09
Support --inspect and --inspect-brk for all Meteor test/run commands.

benjamn:
This PR replaces the `meteor debug` command with the more flexible `--inspect` and `--inspect-brk` command-line flags, which work for any `run`, `test`, or `test-packages` command.

The syntax of these flags is the same as the equivalent Node.js [options](https://nodejs.org/en/docs/inspector/#command-line-options). When no port value is provided, the default is 9229.

Two notable differences:

* The flags affect the server process spawned by the parent build process, rather than affecting the build process itself. The build process can be debugged (as before) by setting the `TOOL_NODE_FLAGS` environment variable to `--inspect` or `--inspect-brk`.

* The `--inspect-brk` flag causes the server process to pause just after server code has loaded but before it begins to execute. This timing is more useful than the Node.js `--inspect-brk` behavior, which is to pause on the first instruction executed by the process, since that is too early to set any useful breakpoints.

Implements https://github.com/meteor/meteor-feature-requests/issues/194, which was opened by @brucejo75.

### DIFF
--- a/History.md
+++ b/History.md
@@ -11,6 +11,23 @@
   version of `npm` used by `meteor npm ...` commands.
   [PR #8835](https://github.com/meteor/meteor/pull/8835)
 
+* The `meteor debug` command has been superseded by the more flexible
+  `--inspect` and `--inspect-brk` command-line flags, which work for any
+  `run`, `test`, or `test-packages` command.
+
+  The syntax of these flags is the same as the equivalent Node.js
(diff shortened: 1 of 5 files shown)


benjamn:
For the record, while this might seem like a new feature to be merging into `release-1.6` after the release candidate cutoff, but it's an important part of a headline feature for Meteor 1.6, so I think it's justifiable/justified.

PR #9205, opened 2017-10-11
Make behaviour of cursor.count() on client reflect server

carlevans719:
This PR changes the default behaviour of `cursor.count()` on the client. Previously, `.count()` on the client would return the number of matching documents taking into account `skip` and `limit`. Now, it will ignore those options unless it is passed `true` when invoked. (i.e. `cursor.count(true)`). This is consistent with the way it works on the server.

Issue: #1201 
Introduction of feature on server-side: 258b14a & bce4df9
Previous PR (incomplete): #5023
Test app: [meteor-issue-1201](https://github.com/carlevans719/meteor-issue-1201)

### DIFF
--- a/packages/minimongo/cursor.js
+++ b/packages/minimongo/cursor.js
@@ -60,7 +60,7 @@ export default class Cursor {
 
     return this._getRawObjects({
       ordered: true,
-      ignoreSkipLimit: !applySkipLimit
+      applySkipLimit
     }).length;
   }
 
@@ -386,8 +386,9 @@ export default class Cursor {
   // Returns a collection of matching objects, but doesn't deep copy them.
   //
   // If ordered is set, returns a sorted array, respecting sorter, skip, and
-  // limit properties of the query provided that options.ignoreSkipLimit is
-  // falsey (#1201). If sorter is falsey, no sort -- you get the natural order.
+  // limit properties of the query provided that options.applySkipLimit is
+  // not set to false (#1201). If sorter is falsey, no sort -- you get the
+  // natural order.
   //
   // If ordered is not set, returns an object mapping from ID to doc (sorter,
   // skip and limit should not be set).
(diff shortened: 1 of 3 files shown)


packages/minimongo/cursor.js
```diff
@@ -44,17 +44,24 @@ export default class Cursor {
...
+  count(applySkipLimit = false) {
     if (this.reactive) {
       // allow the observe to be unordered
       this._depend({added: true, removed: true}, true);
     }
 
-    return this._getRawObjects({ordered: true}).length;
+    return this._getRawObjects({
+      ordered: true,
+      ignoreSkipLimit: !applySkipLimit
```
hwillson: I think it might make sense to just carry the `applySkipLimit` variable naming throughout, instead of switching (and inverting) it to `ignoreSkipLimit`. This would help simplify the code a bit and makes the logic easier to follow. Also, since `applySkipLimit` is a Mongo `count()` supported option, most people reviewing the code will already be familiar with it.

carlevans719: Agreed! I'll get it updated

carlevans719:
Hey @hwillson. I've updated `ignoreSkipLimit` to `applySkipLimit` in `_getRawObjects`. 
I've left the default behaviour of `_getRawObjects` as it was (applying skip / limit) because `.fetch()`, `.forEach()` etc... expect it to do this.

Let me know if you have any other suggestions & feel free to point me towards changelogs / documentation if it needs updating to highlight this breaking change

hwillson:
Thanks for making these changes @carlevans719! We talked about this a bit more, and have decided that maintaining backwards compatibility in this case is (unfortunately) more important than correctness. By giving people an option to change things we're at least providing a workaround to the broken behavior, and we can re-visit this in the future (potentially alongside other related breaking changes). Would you be able to flip things around so that `applySkipLimit` is true by default on the client side? This won't line up with the server side, but we'll make sure this is documented in the `History.md` file. Speaking of which, if you're able to add a small entry to the `History.md` explaining this change, that would be great!

carlevans719:
Makes sense @hwillson! Is the entry in `History.md` okay to go under `v1.5.2.2` or a different/new heading?

hwillson:
If you could add a new heading called `v.NEXT` and add it under that, that would be great!

carlevans719:
@hwillson all sorted. How does that look?

hwillson:
Looks great @carlevans719 - thanks! As soon as the tests finish running, we should be all set here. LGTM!

Issue #9098, opened 2017-09-16
Meteor 1.5.2 ios simulator issue, typeError: Cannot read property 'replace' of undefined


hafizali05:
I’m on Xcode 8.3.3 with OS X 10.12.6

Fresh install of meteor 15.2.

Basically, i just created a basic app using meteor create <app name>
It prompted me to update to meteor 1.5.2 and I updated though it worked fine for meteor 1.4.3.
When I tried running meteor run ios after adding platforms for both the versions, it gave me the below error for 1.5.2

meteor run ios gives error.

=> App running at: http://localhost:3000/
=> Errors executing Cordova commands:

While running Cordova app for platform iOS with options --emulator:

> Error: Command failed: /Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/run --emulator
> TypeError: Cannot read property ‘replace’ of undefined
> at remove (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/node_modules/ios-sim/src/lib.js:282:70)
> at Array.forEach (native)
> at Object.lib.getdevicetypes (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/node_modules/ios-sim/src/lib.js:292:22)
> at Object.listEmulatorImages [as run] (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/lib/list-emulator-images:34:29)
> at deployToSim (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/lib/run.js:173:50)
> at /Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/lib/run.js:115:20
> at _fulfilled (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/node_modules/q/q.js:834:54)
> at self.promiseDispatch.done (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/node_modules/q/q.js:863:30)
> at Promise.promise.promiseDispatch (/Users/{$USER}/Development/webdev/simple-todos-react/.meteor/local/cordova-build/platforms/ios/cordova/node_modules/q/q.js:796:13)
[...]

PR #9213, opened 2017-10-11
Cordova 7.1.0, Android 6.3.0 and iOS 4.5.3 updates

skirunman:
Closes https://github.com/meteor/meteor-feature-requests/issues/196

This is WIP as I did not build and update the bundle version so hoping this can be done by someone else. 

Also, was not sure how to deal with the [deprecated Cordova plugins](https://cordova.apache.org/news/2017/09/22/plugins-release.html) that are still in our build, but did mark them as deprecated and could be removed in the future.


abernix:
Thanks for submitting this!

Since your FR (https://github.com/meteor/meteor-feature-requests/issues/196) mentioned getting it into Meteor 1.6, I'll just point out: Meteor 1.6 is already in release candidate status and https://github.com/meteor/meteor-feature-requests/issues/196 isn't a bug so unless this fixes critical bugs present in Meteor 1.6 RCs, this will likely surface in other versions (e.g. 1.6.1 and 1.5.3).

skirunman:
No worries, your call on Meteor 1.6.1. The Cordova Android 6.3.0 update does fix some issues that I've seen, but did not report as Meteor problems. Specifically `This release now targets the latest Android API level of API 26 and has fixed issues related to the Android SDK Tools 26.0.2 release. Google changed how the Android emulator was executed, causing errors when deploying to the emulator.` so this is something you might want to consider. Thanks.

skirunman:
Also, if this does not make it into Meteor 1.6 then I we should break out the second comment I made in the History.md file as this change is in 1.6.
```
iOS icons and launch screens have been updated to support iOS 11
  [Issue #9196](https://github.com/meteor/meteor/issues/9196)
  [PR #9198](https://github.com/meteor/meteor/pull/9198)
```

skirunman:
One more potential thing is that we are already shipping `cordova-ios@4.5.1` which probably should be bundled with `cordova-plugin-console@1.1.0` for consistency reasons, but is technically not strictly required. Cheers!

skirunman:
@abernix Spoke with rest of my team on this and we'd like to request getting this in 1.6 if possible. Reason is this PR is pretty much the same as https://github.com/meteor/meteor/pull/9137 in scope, which updates `cordova-ios` to support a new version of iOS and was also introduced during the Release Candidate phase of 1.6. `cordova-android@6.3.0` solves the same type of issue, supporting a new version of Android. Anyway, I think the risk is low as this is not really introducing any core changes to Meteor. Thanks again for the consideration.

abernix:
Perfectly fair request!  Though, I think I'm going to stick to my thinking of 1.6.1 and 1.5.3.  The fix in https://github.com/meteor/meteor/pull/9137 was pulled in (late in the 1.6 process already) primarily because of https://github.com/meteor/meteor/issues/9098, which created a problem where the simulator just _could not_ be launched, even without any updates to an existing project and on already-supported devices, due to hyphens in iOS device names being removed/changed in XCode's Device List (`12.4-inch` vs `12.4 inch`).

We're getting very close on 1.6 and if we put this in now, there's a chance it won't have time to fully vet and the `cordova-android` bump to the `26` SDK is hard to judge, especially if we're not fixing a show-stopper bug.  We certainly understand that this is important and we will get it out as soon as possible.  We do _typically_ do Cordova updates in the first point release after a major release.

Has Cordova been working okay for you otherwise with Meteor 1.6?

abernix:
On another, semi-related note, thank you very much for your diligence on helping us stay on top of Cordova items!  You've stood out countless times now, in that you're willing to look into and stay on top of the changes in Cordova, and even more valuably, step up and fix them quickly when you (and others) encounter them.  It's truly appreciated!

skirunman:
I updated this PR to include the [Cordova iOS 4.5.2 release](https://cordova.apache.org/announcements/2017/10/16/ios-release.html) 
as I was running into this [specific bug](https://issues.apache.org/jira/browse/CB-12896) regarding icon mappings.

Also, there was a small merge conflict in the History.md file that might need to be sorted out on merge.

skirunman:
@abernix I assume this PR will be released in [1.6.1]( https://github.com/meteor/meteor/pull/9274) and not in [1.5.3](https://github.com/meteor/meteor/pull/9266) as https://github.com/meteor/meteor/pull/9198 is not in 1.5.3 branch either, thanks.

adamgins:
Hi, I am getting the original error when trying to `meteor run android` with 1.6 
Just trying to follow the details here... does this mean I should rather go with Meteor 1.5.3, until 1.6.1 hits the streets?

Issue #8254, opened 2017-01-14
[ddp-client] Insufficient Sockjs Session-ID Length 


winstoncse:
Current ddp-client package implement Sockjs connection uses 8 characters session ID, as ["Session identifiers should be at least 128 bits long to prevent brute-force session guessing attacks."](https://www.owasp.org/index.php/Insufficient_Session-ID_Length) 

For Example: 

https://www.****.com/sockjs/831/svp1a0l2/xhr

Where **svp1a0l2** is the 8 characters long session id that the client send to the server to distinguish connections.
 
Would be great to increase the Session ID to 32 characters

Sockjs updates with option to customized session ID
https://github.com/sockjs/sockjs-client/pull/250

These would reduce the likelihood of guessing a valid session ID

abernix:
Ultimately, this comes down to upgrading SockJS.  Please see #4114.  I'll try to take a look at that this week since SockJS 1.x has been out for a while now.

I do think your concern is valid, but for what it's worth, I did a quick investigation several months back and determined that, assuming SSL was used (as it should be), the attack surface was very, very small.  The `sessionId` here is not a typical session ID that would be used in a cookie.  It was not possible to connect two clients to the same socket at the same time and once a socket was closed, it could not be re-used.  This means that the attacker would have to win a race with the browser requesting the session.  Since the possible combinations are roughly 3.5 trillion and the timeline to execute that attack very short, the chances are slim.  I think it's worth pointing out that SockJS still hasn't raised their default length from 8 characters either.

Obviously, if you're not using SSL there are some legit concerns here as the SockJS sessionId could be obtained directly from the URL and various techniques could be used to force the client to reconnect, but this risk exists in any non-SSL implementation.

Again, this isn't to say that this shouldn't be addressed and I'll try to look soon – it'd be nice if SockJS just wasn't out of date and was easier to upgrade (currently there are some "Meteor" exceptions which may no longer be necessary).

Thanks for bringing this up.

hwillson:
Let's track this issue under FR https://github.com/meteor/meteor-feature-requests/issues/216. Closing here - thanks!

PR #9215, opened 2017-10-12
Bumping sockjs-node to 0.3.19

mitar:
New version has just been released which passes some useful headers on to the server from client. (Personally, I need DNT header.)

Also, I noticed that this file has been forked in the past and a relatively old version is being used. Probably it should be updated at some point as well: https://github.com/meteor/meteor/blob/devel/packages/ddp-client/sockjs-0.3.4.js

### DIFF
--- a/packages/ddp-server/package.js
+++ b/packages/ddp-server/package.js
@@ -6,7 +6,7 @@ Package.describe({
 
 Npm.depends({
   "permessage-deflate": "0.1.6",
-  sockjs: "0.3.18"
+  sockjs: "0.3.19"
 });
 
 Package.onUse(function (api) {


benjamn:
Can you update the client version as well, and perhaps just use `Npm.depends` instead of copying the file, and then also update the `.npm/package/npm-shrinkwrap.json` files for these packages, by running `meteor test-packages`?

Even if you don't have time for that, thanks for pointing out this upgrade!

mitar:
So there are quiet some changes in the copied file. This is why I haven't yet done this.

Somebody should go through those changes (hopefully all are marked) and check if they are still necessary. It might be that upstream fixed it. (Also, ideally, we should probably submit changes upstream.)

hwillson:
Just to add - this was attempted a while back via https://github.com/meteor/meteor/issues/4114, where it was noticed that the newer version (at the time) was significantly larger. Just something to keep in mind.

hwillson:
This is also related to https://github.com/meteor/meteor/issues/8254 (upgrading the sockjs client will address that issue as well).

hwillson:
@mitar We discussed this PR in our issue review meeting yesterday. Ideally both the client and server libraries would be updated together. As you mentioned, updating the client library will take more work due to Meteor's specific sockjs code changes (a lot of which have been accepted upstream, but someone will need to verify if any of the custom changes are still needed). Putting that aside, if we decided to update just the server library for now, do you know if it's compatible with the old client version of sockjs Meteor is using (which is a custom modified version of 0.3.4)? If `sockjs-node` 0.3.19 and Meteor's custom client sockjs version have been tested and are verified to work together, then we might proceed with merging this. Again though updating the client at the same time would be preferred - if anyone else reading this has time and is interested in helping out on updating the sockjs client library, please chime in! 🙂

abernix:
Since `sockjs-node` is used only to provide the endpoint (internally, the [`StreamServer`](https://github.com/meteor/meteor/blob/35e70fd23d949a2b08bad86a62e9c131ef5a3509/packages/ddp-server/livedata_server.js) for the client to connect to, be it SockJS, WebSocket, or otherwise, this seems worthwhile to merge this as is, even with the outstanding possibility of also updating SockJS on the client, or even not including it at all.

I'll merge this and release new `ddp-server` versions for both 1.5.x and 1.6, since they have separate minor versions.  I'll also update and commit the `npm-shrinkwrap.json` which is necessary when bumping a npm (otherwise it would be updated at run-time by every consumer).

PR #9173, opened 2017-10-04
Add Tool support for both 32 bit (3.2) and 64 bit (3.4) Mongo

hwillson:
(**Note:** I know the discussion I'm about to dive into here is likely better suited in an issue, but it relates to several issues, and ultimately leads to the PR code that is included here.)

# Backstory

As discussed in https://github.com/meteor/meteor-feature-requests/issues/129, the Meteor Tool is currently using Mongo 3.2.x, and it would be great to be able to upgrade to 3.4.x. As further discussed in PR https://github.com/meteor/meteor/pull/8864 however, we can't just update to 3.4.x as 32-bit versions of Mongo have been discontinued (as of Mongo 3.4). Since Meteor currently supports 32-bit Linux and Windows, we can't easily just upgrade to use a 64-bit version of Mongo.

We've talked about the best way to move forward with this, and ultimately decided that Meteor should ideally drop 32-bit support for all OS', as discussed in https://github.com/meteor/meteor-feature-requests/issues/152. This would definitely make the maintenance of Meteor easier moving forward, and is in-line with the approach several other open source projects are taking (like Mongo, Ubuntu, etc.).

While dropping Meteor's 32-bit support is the ultimate goal, we have heard from some community members about how this will impact them (especially those using embedded OS'). There is also a good chunk of work to complete to make this happen, as Meteor doesn't currently offer a 64-bit version for Windows (for example). While this work is pending, developers are getting more and more frustrated by not being able to (easily) use Mongo 3.4.x features, when they're already using a 64-bit OS.

# Proposal

Given the above, I think it makes sense to consider moving towards a 64-bit only Meteor in incremental steps:

- **Step 1:** Update the Meteor Tool to use Mongo 3.2.x for 32-bit based operating systems, and 3.4.x for 64-bit based operating systems. 

- **Step 2:** Start supporting 64-bit Windows as an official Meteor architecture, and provide a 64-bit Windows version of Meteor. 

- **Step 3:** Stop producing and supporting 32-bit versions of Meteor, and no longer support 32-bit architectures.

# This PR

The sooner we can start letting people use Mongo 3.4.x, the better, so the code in this PR addresses **Step 1** above. When using a 32-bit based OS, the Meteor Tool will use (and be limited to) Mongo 3.2.15. When using a 64-bit based OS, the Meteor Tool will use Mongo 3.4.9. 

| OS | Architecture | Meteor Tool Mongo Version |
| -- | ------------ | ----------------- |
[...]

PR #9218, opened 2017-10-12
Windows 64-bit builds.

abernix:
This supersedes and builds on @hwillson's https://github.com/meteor/meteor/pull/9173.  For full details, consult that issue for some backstory.

This PR contains numerous changes to the Windows ["dev bundle"](https://github.com/meteor/meteor/blob/devel/DEVELOPMENT.md#the-dev-bundle) to support 64-bit builds.  Previously, all Windows users (even those on 64-bit capable platforms), have used 32-bit flavors of Node.js and Mongo (plus other build dependencies, such as 7z and Python).

While the original intention was to support Mongo 3.4, this continues to use Mongo 3.2.  First and foremost, it's important to ensure that Windows 64-bit builds are working before enabling Mongo 3.4, which will be slated for a future release.

This should hopefully also resolve an issue with asynchronous stack traces not appearing on 32-bit Windows during use of the Node+V8 Inspector API, presumably because it only utilizes 64-bit integers to identify async operations.

Since the existing Meteor (on Windows only) will have no knowledge of a 64-bit capable Meteor world, this will require an update to the installer and a reinstall for Windows users.  More on that soon, but it should be well worth the upgrade since Meteor will no longer need to run within the 32-bit WoW64 subsystem on 64-bit Windows boxes.

Notable:
* Enable 64-bit builds in `generate-dev-bundle.sh`  (https://github.com/meteor/meteor/commit/0cce6b53dfa6a9b642ebe330fa5aca8fe2b12a35)
* Python 2.14.12 updated to 2.14.14 + 64-bit 7z added to S3 download. (https://github.com/meteor/meteor/commit/acbe9ac4c9a237c0e317de20ef24b68d5181f322)
* Removed duplicated Node headers, instead opting to point `node-gyp` at the headers included in `<dev-bundle-root>/include/node`. (https://github.com/meteor/meteor/commit/f108053c7fb4aa678d65c9ba0b5cc17a3dc76ed0)
* `generate-dev-bundle.ps1` is now ~40% faster. (https://github.com/meteor/meteor/commit/e4f6630102fb6926469334765992ad33de5c38fe)

**Todo:** Update `History.md` to mention that a reinstall will be necessary to take advantage of 64-bit, once that's figured out.


History.md
```diff
@@ -25,6 +16,12 @@
   version of `npm` used by `meteor npm ...` commands.
   [PR #8835](https://github.com/meteor/meteor/pull/8835)
 
+* Windows now supports a 64-bit architecture using native 64-bit binaries.
+  In addition to hopefully being faster, this should allow Windows users to take
+  better advantage of the new V8 inspector API, which appears to only support
```
benjamn: Let's remove "hopefully" and change "appears to only support" to "only supports."

History.md
```diff
@@ -42,6 +39,12 @@
 
   [Feature Request #194](https://github.com/meteor/meteor-feature-requests/issues/194)
 
+* To support developers running on a 32-bit OS, Meteor now supports both 32
```
benjamn: Put a hyphen after 32: "both 32- and 64-bit versions"

benjamn: Am I copy-editing English prose because PowerShell is so foreign to me? Perhaps.

scripts/windows/dev-bundle-lib.psm1
```diff
@@ -0,0 +1,27 @@
+
+<#
+  .Synopsis
```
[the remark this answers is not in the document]
abernix: 🔌 🐢 ⚡️

scripts/generate-dev-bundle.ps1
```diff
@@ -99,7 +99,7 @@ dir node
 # node-gyp (the tool that rebuilds binary node modules). #WinPy
 
 cd "$DIR"
-$py_s3_url = "https://s3.amazonaws.com/com.meteor.static/windows-python/python-${PYTHON_VERSION}.7z"
+$py_s3_url = "https://s3.amazonaws.com/com.meteor.static/windows-python/$PLATFORM/python-${PYTHON_VERSION}.7z"
```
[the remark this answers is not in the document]
abernix: Should still be present? https://s3.amazonaws.com/com.meteor.static/windows-python/python-2.7.12.7z

There was no pressing reason to destroy it.

History.md
abernix: Ok! Sticking to the firm facts!

History.md
abernix: [PowerShell approved verbs](https://msdn.microsoft.com/en-us/library/ms714428(v=vs.85).aspx) are prose, of sorts. 🤕 

Mind you, I only followed them loosely.

abernix:
Feedback all addressed!  Thanks!  Happy to adjust further as well.

tools/utils/archinfo.js
```diff
@@ -67,9 +67,17 @@ var utils = require('./utils.js');
  *   nobody has asked for it.
  *
  * os.windows.x86_32
- *   This is 32 and 64 bit Windows. It seems like there is not much of
- *   a benefit to using 64 bit Node on Windows, and 32 bit works properly
- *   even on 64 bit systems.
+ * os.windows.x86_64
+ *   Once, on the far side of yesterday, there was not a 64-bit
```
hwillson: More story-telling comments like this please! 🍿

lhz516:
How do I update or reinstall the Meteor installer? I'm still not able to use 1.6-rc.10 after uninstalling and reinstalling Meteor on Win10-64x.

PR #9237, opened 2017-10-16
Fix isNpmUrl to handle http/https urls (Closes #9236)

### LINKED ISSUE #9236: It should be possible to do Npm.depends with a url ({"module": "https://......tar.gz"})
meteor version : 1.5.2.2

Create a package with following package.js file : 

```
Npm.depends({
	async: "https://github.com/caolan/async/archive/v2.3.0.tar.gz",
});
```

# Actual behavior : 

The code is failing with : 

```
package.js:1:-26: must declare valid version of dependency: async@https://github.com/caolan/async/archive/v2.3.0.tar.gz
```

# Expected behavior.

It should install the module 

`npm install https://github.com/caolan/async/archive/v2.3.0.tar.gz` works as expected.

I have found the source of the problem and will do a small Merge request.

edi9999:
See https://github.com/meteor/meteor/issues/9236 for the details about the bug.

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@
 * iOS icons and launch screens have been updated to support iOS 11
   [Issue #9196](https://github.com/meteor/meteor/issues/9196)
   [PR #9198](https://github.com/meteor/meteor/pull/9198)
+  [Issue #9236] Allow Npm.depends to specify any http or https URL [#9236](https://github.com/meteor/meteor/issues/9236)
 
 ## v1.5.2.2, 2017-10-02
 
--- a/tools/utils/utils.js
+++ b/tools/utils/utils.js
@@ -490,7 +490,7 @@ exports.isUrlWithSha = function (x) {
 exports.isNpmUrl = function (x) {
   // These are the various protocols that NPM supports, which we use to download NPM dependencies
   // See https://docs.npmjs.com/files/package.json#git-urls-as-dependencies
-  return exports.isUrlWithSha(x) || /^(git|git\+ssh|git\+http|git\+https)?:\/\//.test(x);
+  return /^(git|git\+ssh|git\+http|git\+https|https|http)?:\/\//.test(x);
 };
 
 exports.isPathRelative = function (x) {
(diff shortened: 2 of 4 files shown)


History.md
```diff
@@ -4,6 +4,8 @@
   [Issue #9196](https://github.com/meteor/meteor/issues/9196)
   [PR #9198](https://github.com/meteor/meteor/pull/9198)
 
+* Allow Npm.depends to specify any http or https URL [#9236](https://github.com/meteor/meteor/issues/9236)
```
hwillson: Please move the issue link to a new line, and prefix it with "Issue" (to line up with the other `History.md` entries).

tools/utils/utils.js
```diff
@@ -490,7 +490,7 @@ exports.isUrlWithSha = function (x) {
 exports.isNpmUrl = function (x) {
   // These are the various protocols that NPM supports, which we use to download NPM dependencies
   // See https://docs.npmjs.com/files/package.json#git-urls-as-dependencies
-  return exports.isUrlWithSha(x) || /^(git|git\+ssh|git\+http|git\+https)?:\/\//.test(x);
+  return /^(git|git\+ssh|git\+http|git\+https|https|http)?:\/\//.test(x);
```
hwillson: Would you mind adding some tests for `isNpmUrl` to verify the accepted dependency URL formats? They don't have to be that extensive, just one quick check for each accepted URL format, including `http` and `https`. The tests would go in [`tools/tests/utils-tests.js`](https://github.com/meteor/meteor/blob/1aaeca1cb82f080e15267b1dc4dc724074dc3b1a/tools/tests/utils-tests.js).

benjamn: I would also like to be sure that these new URLs are preserved correctly in `npm-shrinkwrap.json`, so that we don't keep trying to reinstall the packages when they've already been installed.

edi9999: I have added a few tests.

> I would also like to be sure that these new URLs are preserved correctly in npm-shrinkwrap.json, so that we don't keep trying to reinstall the packages when they've already been installed.

Does this comment have to do with this MR ?

History.md
edi9999: Fixed

benjamn:
Merged to `devel`:
* https://github.com/meteor/meteor/commit/16b74e5c0cfab46093f7c7e1ec338844f90121c4
* https://github.com/meteor/meteor/commit/c8c88300a6ef3497f2bbafcdd12a12f62ef7d2c5

Thanks for fixing this @edi9999!

PR #9258, opened 2017-10-24
Update Facebook OAuth app config steps

### LINKED ISSUE #8867: facebook-config-ui instructions are outdated
Due to some recent changes in Facebook's developer portal, the instructions provided by accounts-ui are slightly unclear.

Some background: https://guide.meteor.com/accounts.html#accounts-ui suggests using the package `accounts-ui` with a package like `accounts-facebook` as a quick way to prototype an accounts system for your app.  The `facebook-config-ui` package presents a modal for the developer to enter the App ID and App Secret, as well as a guide through the process of creating and configuring a facebook app to support Facebook Login.

It seems like facebook recently updated their developer portal, and our guide is no longer accurate. In fact, [facebook's very own guide](https://developers.facebook.com/docs/apps/register) seems to be out of date as well. 

Some key differences:
- Right after creating a new Facebook App, you are asked to give it a name, instead of being shown a "Choose a Platform" page
- After giving it a name, you are redirected to a "Select a Product" page
- Only after selecting the "Facebook Login" product do you get to choose a platform

[I'm not sure if this counts as a bug -- sorry if this issue actually belongs somewhere else]

hwillson:
Facebook has updated their application developer portal, so the steps for creating a Facebook OAuth app have changed. This commit updates the steps shown as part of the `facebook-config-ui` setup process, to line up with Facebook's new dev portal.

Fixes #8867.


abernix:
Published as `facebook-config-ui@1.0.1`.  Thanks, @hwillson!

Issue #8026, opened 2016-11-07
Invalid ROOT_URL envar throws confusing error.


sebakerckhof:
When I tried to run meteor like: `ROOT_URL=192.168.40.106:3000 meteor` I got this error:
```
=> Started proxy.                             
Errors prevented isopacket load:              
                                              
While loading isopacket `constraint-solver`:
packages/meteor.js:1221:19: Cannot read property 'slice' of null
at packages/meteor.js:1221:19
at packages/meteor.js:1233:4
at packages/meteor.js:1380:3


/home/seke/.meteor/packages/meteor-tool/.1.4.2.1r0536n++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/lib/node_modules/meteor-promise/promise_server.js:190
      throw error;
      ^

Error: isopacket load failed?
    at loadIsopacketFromDisk (/tools/tool-env/isopackets.js:300:11)
    at /tools/tool-env/isopackets.js:97:23
    at Object.exports.noYieldsAllowed (/tools/utils/fiber-helpers.js:37:14)
    at Object.load (/tools/tool-env/isopackets.js:89:23)
    at ProjectContext._buildResolver (/tools/project-context.js:755:22)
    at /tools/project-context.js:487:29
    at /tools/utils/buildmessage.js:359:18
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at /tools/utils/buildmessage.js:352:34
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at /tools/utils/buildmessage.js:350:23
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at Object.enterJob (/tools/utils/buildmessage.js:324:26)
    at /tools/project-context.js:486:20
    at /tools/packaging/catalog/catalog.js:100:5
    at /tools/utils/buildmessage.js:271:13
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at /tools/utils/buildmessage.js:264:29
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at /tools/utils/buildmessage.js:262:18
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at /tools/utils/buildmessage.js:253:23
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
    at Object.capture (/tools/utils/buildmessage.js:252:19)
[...]

PR #9261, opened 2017-10-26
Fix error message for invalid ROOT_URL #8026

BDav24:
Fix error message for invalid ROOT_URL #8026

### DIFF
--- a/packages/meteor/url_server.js
+++ b/packages/meteor/url_server.js
@@ -4,7 +4,7 @@ if (process.env.ROOT_URL &&
   if (__meteor_runtime_config__.ROOT_URL) {
     var parsedUrl = Npm.require('url').parse(__meteor_runtime_config__.ROOT_URL);
     // Sometimes users try to pass, eg, ROOT_URL=mydomain.com.
-    if (!['http:', 'https:'].includes(parsedUrl.protocol) || !parsedUrl.host) {
+    if (['http:', 'https:'].indexOf(parsedUrl.protocol) === -1 || !parsedUrl.host) {
       throw Error("$ROOT_URL, if specified, must be an URL");
     }
     var pathPrefix = parsedUrl.pathname;
--- a/packages/meteor/url_server.js
+++ b/packages/meteor/url_server.js
@@ -4,7 +4,7 @@ if (process.env.ROOT_URL &&
   if (__meteor_runtime_config__.ROOT_URL) {
     var parsedUrl = Npm.require('url').parse(__meteor_runtime_config__.ROOT_URL);
     // Sometimes users try to pass, eg, ROOT_URL=mydomain.com.
-    if (!parsedUrl.host) {
+    if (!['http:', 'https:'].includes(parsedUrl.protocol) || !parsedUrl.host) {
       throw Error("$ROOT_URL, if specified, must be an URL");
     }
     var pathPrefix = parsedUrl.pathname;
--- a/packages/meteor/url_server.js
+++ b/packages/meteor/url_server.js
@@ -4,7 +4,7 @@ if (process.env.ROOT_URL &&
   if (__meteor_runtime_config__.ROOT_URL) {
     var parsedUrl = Npm.require('url').parse(__meteor_runtime_config__.ROOT_URL);
     // Sometimes users try to pass, eg, ROOT_URL=mydomain.com.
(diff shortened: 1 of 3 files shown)


BDav24:
@hwillson I'm struggling with meteor tests... I added the following code to `tools/test/run.js`:

```js
selftest.define("run ROOT_URL must be an URL", function () {
  var s = new Sandbox();
  var run;

  s.set("ROOT_URL", "192.168.0.1");
  s.createApp("myapp", "standard-app");
  s.cd("myapp");

  run = s.run();
  run.waitSecs(10);
  run.matchErr("$ROOT_URL, if specified, must be an URL");
  run.expectExit(1);
});
```

And I get `wrong-exit-code: Expected: 0; actual: 1` even if I wrote `run.expectExit(1);`. What's going wrong?

> $ ./meteor self-test "run ROOT_URL"
run: run ROOT_URL must be an URL ...          
  ... fail!                                   
  ... retrying (2 tries remaining) ...        
  ... fail!                                   
  ... retrying (1 try remaining) ...          
  ... fail!                                   
  => wrong-exit-code at ../../../../../tools/tool-testing/selftest.js:709
  => Expected: 0; actual: 1                   
  => Last 100 lines:                          
  2| $ROOT_URL, if specified, must be an URL. 
Skipped 137 non-matching tests (don't match specified pattern)
1 failure:                                    
>  - run: run ROOT_URL must be an URL

packages/meteor/url_server.js
```diff
@@ -4,7 +4,7 @@ if (process.env.ROOT_URL &&
   if (__meteor_runtime_config__.ROOT_URL) {
     var parsedUrl = Npm.require('url').parse(__meteor_runtime_config__.ROOT_URL);
     // Sometimes users try to pass, eg, ROOT_URL=mydomain.com.
-    if (!parsedUrl.host) {
+    if (['http:', 'https:'].indexOf(parsedUrl.protocol) === -1 || !parsedUrl.host) {
```
benjamn: Why not reverse this condition, like so?
```js
if (!parsedUrl.host || ['http:', 'https:'].indexOf(parsedUrl.protocol) === -1) {
```

BDav24: @benjamn I'd say both are of the same importance, so order doesn't matter. I put the protocol test first because it's the way we read URLs, but it can be reversed if you prefer.

By any chance, have you seen my question here: https://github.com/meteor/meteor/pull/9261#issuecomment-342100497 ?

hwillson:
Hi @BDav24 - sorry for the delay on this. By default Meteor's self-test `Sandbox.createApp` function cd's into the new temporary application directory and [does a run with the `--prepare-app` option](https://github.com/meteor/meteor/blob/c09d8e57f01ed95c987f7a8b34b4d2de1a75b249/tools/tool-testing/selftest.js#L688-L696) to get things ready (by building/downloading packages). After running this, it's expecting an exit code of 0 internally. Since your new `ROOT_URL` changes are called when this happens though, the `--prepare-app` option returns with an exit code of 1. This is where the failure is happening - it's happening before the check in your test. An easy way to avoid this is to use the `dontPrepareApp` option in your test - so something like the following will work:

```js
selftest.define("run ROOT_URL must be an URL", function () {
  var s = new Sandbox();
  var run;

  s.set("ROOT_URL", "192.168.0.1");
  s.createApp("myapp", "standard-app", { dontPrepareApp: true });
  s.cd("myapp");

  run = s.run();
  run.matchErr("$ROOT_URL, if specified, must be an URL");
  run.expectExit(1);
});
```

packages/meteor/url_server.js
hwillson: @BDav24 I believe Ben's comment was suggesting to reverse the condition as there is a small performance gain when doing the [falsy check first](https://jsperf.com/falsy-vs-indexof/1).

BDav24: Well you'll gain 0.15 microseconds (0.00015 milliseconds) each time you start an instance :)

BDav24:
@hwillson for the second test (inside https://github.com/meteor/meteor/blob/9e65d8bea72068f042b41ffa0df55791b4c7d71f/packages/meteor/url_tests.js), I don't see how to write the test, because `url_server.js` is only called once (when the test server is starting).

hwillson:
No problem @BDav24, thanks for the update. We should be okay to skip it then, so we should be all set here. Thanks for submitting this PR - LGTM!

Issue #7075, opened 2016-05-18
Cannot minifying CSS with page-margin boxes


Louis-Tian:
It seems the `standard-minifier-css` doesn't support nested CSS rules. 

style.css ( [relevent CSS3 specification](https://www.w3.org/TR/css3-page/#margin-at-rules)  )

```
@page {
  size: 8.5in 11in;
  margin: 10%;

  @top-left {
    content: "Hamlet";
  }
}
```

error message:

```
=> Errors prevented startup:

   While minifying app stylesheet:
   app/client/style.css:9:3: @page missing '}'

=> Your application has errors. Waiting for file change.
```

meteor version and dependencies (nohting extra, all standard stuff from `meteor create`)
- autopublish            1.0.7  (For prototyping only) Publish the entire database
- blaze-html-templates   1.0.4  Compile HTML templates into reactive UI with Meteo
- ecmascript             0.4.3  Compiler plugin that supports ES2015+ in all .js f
- es5-shim               4.5.10  Shims and polyfills to improve ECMAScript 5 suppo
- insecure               1.0.7  (For prototyping only) Allow all database writes f
- jquery                 1.11.8  Manipulate the DOM using CSS selectors
- meteor-base            1.0.4  Packages that every Meteor app needs
- mobile-experience      1.0.4  Packages for a great mobile user experience
- mongo                  1.1.7  Adaptor for using MongoDB and Minimongo over DDP
- reactive-var           1.0.9  Reactive variable
- standard-minifier-css  1.0.6  Standard css minifier used with Meteor apps by def
- standard-minifier-js   1.0.6  Standard javascript minifiers used with Meteor app
- tracker                1.0.13  Dependency tracker to allow reactive callbacks

abernix:
This is actually getting thrown by the `css-parse` npm module.  There is an existing (relatively old) bug over there at reworkcss/css#23.  Doesn't seem to be a very popular issue though. :(

I would recommend +1'ing that issue if you are experiencing this.  Alternatively, maybe a new, more actively maintained AST CSS parser exists.

lydell:
I’d recommend switching to [postcss](https://github.com/postcss/postcss/).

hwillson:
[...]

PR #9263, opened 2017-10-27
Update the default CSS parsing/combining/minifying tools

hwillson:
The `minifier-css` package is currently using the following outdated (and abandoned) npm packages:

- `css-parse`
- `css-stringify`

This has led to issues like https://github.com/meteor/meteor/issues/7075, where a PR has been provided to the upstream repo to fix the issue, but it hasn't been (and won't be) accepted. As discussed in https://github.com/meteor/meteor/issues/7075#issuecomment-288070647, we could consider forking and maintaining our own versions of the above packages, but adding more to Meteor's maintenance work load seems less than ideal.

This PR aims to modernize Meteor's CSS build process (in baby steps), by leveraging today's most popular CSS transformation tool - `postcss`. In this PR, `postcss` is used to replace the functionality provided by the above packages, as well as remove the need for a custom internal minification process. This PR does not open up Meteor's CSS build process by allowing 3rd party `postcss` plugins to be run (like the Meteor package [`juliancwirko:postcss`](https://atmospherejs.com/juliancwirko/postcss) does). It simply leverages `postcss` to handle the tasks that Meteor's CSS build process is already handling.

Integrating `postcss` like this could be considered a first step towards modernizing Meteor's CSS build process. With these changes in place, at some point in the future we could consider opening things up further, to allow 3rd party `postcss` plugins to be run, as part of the existing CSS build process. This could introduce a lot of new capabilities to Meteor's CSS build system, by leveraging `postcss`'s vast (and growing) plugin ecosystem. Food for future thought anyways.

While the changes outlined in this PR are working (verified in development and production), I've marked it as a work in progress for now. I'd like to do more testing, benchmarking and performance tuning (I'll share those performance numbers here shortly) before considering this ready for review. 

One final note regarding backwards compatibility. While the `minifier-css` package is an internal package, it's definitely possible people are relying on it externally. I've tried to make all changes to its API backwards compatible, but in some cases this isn't really possible (well, it's possible, but not easily). The abstract syntax tree that `postcss` works with is different than the AST `css-parse` worked with, so some public functions like `CssTools.parseCss` will now return an AST in a different format. If this is deemed to be an issue we can consider adding in a mapping layer to covert the new AST to the old AST.

More details shortly - thanks!


hwillson:
Just to add - I'm aware of the `self-test` failures, and will have them resolved shortly. Thanks!

tools/tests/standard-minification.js
```diff
@@ -1,39 +0,0 @@
-import selftest, {Sandbox} from '../tool-testing/selftest.js';
```
hwillson: Quick note regarding why I've removed this test (and its supporting test app). The need to split stylesheets up with more than 4095 selectors was required to support IE 9. All modern (and supported) browsers can handle stylesheets with more than 4095 selectors, so I don't think there's much value in retaining this test (and its associated functionality). Let me know if anyone disagrees!

hwillson:
A quick follow up regarding performance benchmarks; This PR introduces `postcss` as pretty much a drop in replacement for `css-parse` and `css-stringify`, which means it has been wired in following the existing API as much as possible. Given this, using `postcss` is slower in some places, faster in others. Given that `postcss` can do so much more than the previous libraries, we could likely eradicate any performance differences by re-writing Meteor's CSS build process to better leverage `postcss`'s capabilities. Still though - given everything additional `postcss` introduces, I'm pretty happy with the benchmark results as is. Here are a few first startup build numbers (taken from a project that has 10,000 lines worth of its own CSS, coming in at about 190K in size):

Using the current `css-parse` and `css-stringify` packages:

```
|    │     ├─ ClientTarget#minifyCss..............................239 ms (1)
|    │     │  ├─ mergeCss.........................................235 ms (1)
|    │     │  │  ├─ CssTools.parseCss                              92 ms (3)
|    │     │  │  ├─ CssTools.mergeCssAsts...........................9 ms (1)
|    │     │  │  │  ├─ CssTools.rewriteCssUrls                      7 ms (3)
|    │     │  │  │  └─ other CssTools.mergeCssAsts                  2 ms
|    │     │  │  ├─ CssTools.stringifyCss                          81 ms (1)
|    │     │  │  ├─ composing source maps                          41 ms (1)
|    │     │  │  └─ other mergeCss                                 12 ms
|    │     │  └─ other ClientTarget#minifyCss                       4 ms
```

Using `postcss`:

```
|    │     ├─ ClientTarget#minifyCss..............................372 ms (1)
|    │     │  ├─ mergeCss.........................................366 ms (1)
|    │     │  │  ├─ CssTools.parseCss                              81 ms (3)
|    │     │  │  ├─ CssTools.mergeCssAsts..........................19 ms (1)
|    │     │  │  │  └─ CssTools.rewriteCssUrls                     18 ms (3)
|    │     │  │  ├─ CssTools.stringifyCss                         196 ms (1)
|    │     │  │  ├─ composing source maps                          55 ms (1)
|    │     │  │  └─ other mergeCss                                 14 ms
|    │     │  └─ other ClientTarget#minifyCss                       6 ms
```

The above is just a sample run from a checkout using both approaches. The `ClientTarget#minifyCss` average over 5 runs is much closer:

`css-parse` and `css-stringify` approach:

- Run 1: 239
- Run 2: 380
- Run 3: 257
- Run 4: 250
- Run 5: 241

**Average: 273 ms**

`postcss` approach:

- Run 1: 369
- Run 2: 375
- Run 3: 382
- Run 4: 349
- Run 5: 312

**Average: 357 ms**

hwillson:
This should now be ready for review. Quick note - I bumped the `minifier-css` major version since this work breaks backwards compatibility. I've detailed this in the `History.md`. I have a feeling the number of people impacted by these changes will be fairly low, as not many people are tying into the `minifier-css` package directly. These changes will likely impact build plugin authors that are tying into `minifier-css`, but then again maybe not (as Meteor's own `standard-minifier-css` package, that uses `minifier-css`, isn't impacted for example).

Thanks!

mitar:
Oh, I just now saw this pull request. This is great and timely work! 

I would propose that looking into how to add plugins to postcss should be prioritized higher because [juliancwirko's package](https://github.com/juliancwirko/meteor-postcss/) is not maintained anymore.

We discussed this a bit also in stylus processing pipeline (it has nib and autoprefixer currently) and while you can configure things through `addFiles` options, there is no way to do so for the whole project.

I would suggest that we simply allow one to put this into `package.json`, and have postcss configuration there.

hwillson:
Thanks @mitar - the possibility of opening this up to allow `postcss` plugins is definitely exciting! I just wanted to tackle this in steps, by first introducing `postcss` to replace the abandoned packages Meteor is using (and to fix issues like https://github.com/meteor/meteor/issues/7075). If this PR is accepted then the next step will be to consider opening things up. Actually, I have some quick prototype code ready that does take things further by allowing 3rd party `postcss` plugins to be leveraged in `minifier-css` (that I wired up just as a test), and it is working. If we want to go further here (after/if this PR is accepted), I'll definitely look into polishing that code up and submitting a new PR.

hwillson:
@benjamn @abernix I've adjusted the minor versions (of both `minifier-css` and `standard-minifier-css`), so this should be all set for a review (no rush of course) - thanks!

packages/minifier-css/minifier.js
```diff
@@ -1,112 +1,209 @@
...
+        CssTools.rewriteCssUrls(ast);
 
-      // if there are imports left in the middle of file, warn user as it might
-      // be a potential bug (imports are valid only in the beginning of file).
-      if (_.any(ast.stylesheet.rules, rulesPredicate("import"))) {
-        // XXX make this an error?
-        warnCb(ast.filename, "there are some @import rules those are not taking effect as they are required to be in the beginning of the file");
-      }
+        const imports = ast.nodes.splice(0, importCount);
+        newAst.nodes = newAst.nodes.concat(imports);
```
benjamn: Any reason not to do `newAst.nodes.push(...imports)` here?

packages/minifier-css/minifier.js
```diff
@@ -1,112 +1,209 @@
...
     });
 
-    // Now we can put the rest of CSS rules into new AST
-    _.each(cssAsts, function (ast) {
-      newAst.stylesheet.rules =
-        newAst.stylesheet.rules.concat(ast.stylesheet.rules);
+    // Now we can put the rest of CSS rules into new AST.
+    cssAsts.forEach((ast) => {
+      if (ast.nodes) {
+        newAst.nodes = newAst.nodes.concat(ast.nodes);
```
benjamn: And `newAst.nodes.push(...ast.nodes)` here?

packages/minifier-css/minifier.js
```diff
@@ -116,96 +213,77 @@ var rewriteRules = function (rules, mergedCssPath) {
...
-        // prefix, here `myapp/`, if applicable).
-        relativeToMergedCss = pathRelative(mergedCssPath, absolutePath);
-        newCssUrl = "url(" + quote + relativeToMergedCss + quote + ")";
-        value = value.replace(oldCssUrl, newCssUrl);
+      // We don't rewrite URLs starting with a protocol definition such as
+      // http, https, or data, or those with network-path references
+      // i.e. //img.domain.com/cat.gif
+      if (resource.protocol !== null
+          || resource.href.startsWith('//')
+          || resource.href.startsWith('#')) {
```
benjamn: We usually put the `||` or `&&` operator at the end of the line, so the conditions line up:
```js
if (resource.protocol !== null ||
    resource.href.startsWith('//') ||
    resource.href.startsWith('#')) {
```

packages/minifier-css/minifier.js
```diff
@@ -116,96 +213,77 @@ var rewriteRules = function (rules, mergedCssPath) {
...
+      // the `myapp/` prefix. Since this path prefix is only known at launch
+      // time (rather than build time) we can't use absolute paths to link
+      // resources in the generated CSS.
+      //
+      // Instead we transform absolute paths to make them relative to the
+      // merged CSS, leaving to the browser the responsibility to calculate
+      // the final resource links (by adding the application deployment
+      // prefix, here `myapp/`, if applicable).
+      const relativeToMergedCss = pathRelative(mergedCssPath, absolutePath);
+      const newCssUrl = 'url(' + quote + relativeToMergedCss + quote + ')';
```
benjamn: Unless we really need to preserve the `quote` character, this can just be
```
'url(' + JSON.stringify(relativeToMergedCss) + ')'
```

packages/minifier-css/minifier.js
```diff
@@ -1,112 +1,209 @@
...
+
+  /**
+   * Minify the passed in CSS string.
+   *
+   * @param {string} cssText CSS string to minify.
+   * @return {String[]} Array containing the minified CSS.
+   */
+  minifyCss(cssText) {
+    const f = new Future;
+    postcss([ cssnano ]).process(cssText).then(result => {
```
benjamn: Can the object returned from `postcss(...)` be saved for later, or is it important to create a new one each time?

hwillson: `postcss()` is just a helper function that creates and returns a new `Processor` object. So it's essentially the same as calling `new Processor()`. We can definitely save the returned `Processor` object. It's pretty lightweight so we might not gain much, and we'll have to adjust how  `CssTools.minifyCss` is passing the `cssnano` plugin into `postcss()`, but it's definitely doable. I'll adjust - thanks!

packages/minifier-css/minifier.js
hwillson: No reason at all - I'll update, thanks!

packages/minifier-css/minifier.js
hwillson: Sounds good - I'll adjust, thanks!

packages/minifier-css/minifier.js
hwillson: Right, I don't think we need to preserve it, but I'll check (that was old code). If not, I'll update - thanks!

packages/minifier-css/minifier.js
hwillson: I looked into this further @benjamn - unfortunately we can't share the object. In `CssTools` we're making 3 separate `postcss()` calls, one of which looks like `postcss([ cssnano ])`. So here we're passing a plugin into a `new Processor` call behind the scenes. If we were to change things around so we can share the `postcss()` created `Processor` object, I thought we could leverage the `Processor` objects [`use`](http://api.postcss.org/Processor.html#use) method, so we could essentially have something like `this._cssProcessor.process()` or `this._cssProcessor.use(cssnano).process()` (where `this._cssProcessor` points to our shared `Processor` object). It turns out though that the `use` method just adds the plugin to the object, so it's then used with any processing from that point forward (with or without the `use` method). Since we don't always want the plugin, I think our best bet is to leave this as is. Again, the created `Processor` object is pretty lightweight so this should be okay.

packages/minifier-css/minifier.js
hwillson: I looked into replacing this, but it looks like the [current tests](https://github.com/meteor/meteor/blob/70ffce52fc36e91e188cc1d10a67a63f9969f5dd/packages/minifier-css/urlrewriting-tests.js) go to lengths to make sure the quote is preserved exactly as is. So either no quote, exactly a single quote, or exactly a double quote. There are tests verifying this fairly extensively. I've tested the results of `JSON.stringify` in practice, and everything seems to be okay, but I'm a bit concerned that the tests were added for a specific reason (maybe browser compatibility issues, special filenames / paths, etc.?). Given this, I'm wondering if we should leave this as is?

abernix:
(Fixed conflicts in `History.md`.)

PR #9266, opened 2017-10-28
Release 1.5.3

abernix:
This will be a simple follow-up release to Meteor 1.5.2, with the most important inclusion being Node v4.8.5 which includes a Node.js security update as explained in the [v4.8.5 release blog](https://nodejs.org/en/blog/release/v4.8.5/).  Due to the timing of the Node.js security release, the fixes for [our garbage collection patch](https://github.com/nodejs/node/pull/14829), which fixed a recurring segmentation fault which manifested itself during Meteor development, were pushed to Node v4.8.6 (which will likely be the last v4 Node.js release).  Therefore, this version of Meteor will still ship a version of Node which includes our fix until we can move back to the official v4.8.x releases.

A change to the import scanner is also being included to address the problem in https://github.com/meteor/meteor/issues/9182, which was fixed by https://github.com/meteor/meteor/pull/9187 and some follow-up commits.  As well as a change which should reduce the size of the `dev_bundle` (which is downloaded for each Meteor upgrade) for Unix builds.  Other small fixes may be considered, though the focus should be to get this release out as soon as possible.

For Unix platforms, the first release candidate is already ready to test.  Please try it as soon as possible and report any issues here!

    meteor update --release 1.5.3-rc.0

Unfortunately, I ran into a bit of an issue publishing the Windows version of that build, and I'll have to finish looking into it next week.


benjamn:
We ended up reverting the “Rename installPath property to absModuleId, and make absolute” commit on `release-1.6`, so I think we should do that here as well, or maybe just rebase it out and force push?

benjamn:
If we do decide to rebase, I suppose we could squash the typo fix commit into the commit that introduced the typo, too?

sirpy:
i've experienced segmentation faults when running in production, ie running the bundle with "node main.js" (i'm using mup with abrenix:meteord)
switching to running production with "meteor node" instead of the installed node solves the issue and the segmentation faults stopped.

abernix:
Good shout on reverting https://github.com/meteor/meteor/pull/9266/commits/4e778b6e17ba5ac6813ee0bd3a2610de5babb1b6, @benjamn.  In fact, that turned out to be the solution to the Windows publishing problem.  I opted to just revert that single commit and not rebase since I had already published `1.5.3-rc.0`.

Meteor 1.5.3-rc.1 is now published for all its supported platforms (specifically: keep in mind that there will only be a 32-bit Windows build, whereas Meteor 1.6 is now available on Windows in a 64-bit flavor as well).

    meteor update --release 1.5.3-rc.1

abernix:
For those still on the 1.5.x-series of Meteor, Meteor 1.5.3 has been officially released!  The changelog is available [on `master`'s `History.md`](https://github.com/meteor/meteor/blob/master/History.md#v153-2017-11-04), in addition to the [1.5.3 docs changelog](https://docs.meteor.com/v1.5.3/changelog.html).

Note that since Meteor 1.6 is the most recent, it will be necessary to use the `--release` argument to stay on the 1.5.x series since `meteor update` by itself will jump straight to 1.6.  Within an existing application, this can be done by running:

    meteor update --release 1.5.3

PR #9272, opened 2017-10-30
Add support for glob imports

mitar:
I just needed this for myself, so I made it.

See: https://github.com/meteor/meteor-feature-requests/issues/211

This allows imported files to use glob imports. It does not provide glob importing from Meteor directly (so `{}` prefixed paths cannot be globs). I think this would be relatively easy to add if people are interested (using minimatch library to iterate over `allFiles` using a glob pattern instead of just calling `has` on it), but it would probably introduce a bit of performance penalty.

As this PR was not encouraged, I understand it might not be accepted.


packages/stylus/plugin/compile-stylus.js
```diff
@@ -89,16 +90,21 @@ class StylusCompiler extends MultiFileCachingCompiler {
     }
 
     const importer = {
-      find(importPath, paths) {
+      find(importPath, paths, filename) {
```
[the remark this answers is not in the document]
mitar: Ah, this is my habit. By looking at Stylus code, I realized it is being passed in. I like to then list all arguments on the callee said as well, because maybe in the future somebody will want to fix something here and spend hours doing a workaround, when they would just have this value in the argument. I can remove it, but to me it is a very useful practice.

mitar:
> There isn't really any reason why they need to be part of core, and better 3rd party developer versions of both plugins exist. 

Are there other Stylus packages? I have not found any maintained.

mitar:
> Even though it doesn't handle {} prefixed paths, it's still an improvement.

I can add this probably as well relatively easy, but it will come with an extra NPM dependency and a bit of performance hit for cases when a file is not directly found and should be globed over.

With Stylus, the main issue is also that upstream is not really being developed at all.

packages/stylus/plugin/compile-stylus.js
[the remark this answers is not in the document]
hwillson: Okay, thanks @mitar - we can leave it.

hwillson:
Hi @mitar - have you had a chance to work on the items mentioned in https://github.com/meteor/meteor/pull/9272#pullrequestreview-73465297? Let me know if you're busy, and I'll look into getting this one wrapped up. Thanks!

History.md
```diff
@@ -1,5 +1,11 @@
 ## v.NEXT
 
+* The `stylus` package now supports
+  [glob imports](http://stylus-lang.com/docs/import.html#file-globbing) like
+  `@import './folder/*'`.
```
mitar: Thanks for taking this over.

Globing is currently supported only when imported files import additional files with globing, I think. This is the case I have been solving here. For direct glob import, this should be verified, but I feel one more change has to be done.

I was a bit swamped these days, so I could not check this. If you have time, can you check if direct import works. If it does, then great. :-)

hwillson: I'm currently writing tests for this @mitar so I'll be able to verify this shortly.

mitar: Oh, great!

Ah, I feel so sorry to just make half of the pull request and dump it on you. I was just not sure if it will be even accepted, so I didn't bother make docs and tests then. But now I haven't yet cycled back to this code to finish it.

Thanks for picking it up.

hwillson:
Hi @mitar - while working on wrapping this up and doing more testing, I've run into issues with this approach:

### Problem 1

These changes don't work when using the Meteor Tool in dev mode with local packages. If a local package is using a glob to refer to another internal file, say with something like `@import ./common/*`, when the stylus build plugin runs, it tries to interpret that glob from the directory the tool is running in. So it's trying to expand the glob from the app root instead of the package root, which means it won't find a match. This can be fixed in the package by changing the glob to `@import **/common/*`, but that's not really something we should expect package developers to remember to do. As a workaround to this, we could have the stylus build plugin always prefix the glob with `**/` like

```js
function resolvePath(path) {
  return glob.sync(`**/${path}`);
}
```

but this will introduce a performance hit.

### Problem 2

Under the same scenario as problem 1 (Meteor Tool, dev mode, referencing local package that's doing an internal glob based import), when the stylus build plugin's `importer.readFile` function runs, it can't find a match using the glob expanded `filePath` it's passed in as a parameter. For example, let's assume the local package has a `.styl` file with the contents `@import ./common/*` and a separate `./common/common.import.styl` file. When the glob is expanded by the stylus build plugin, and passed into the `importer.readFile` function as `filePath`, the path looks like:

```
packages/some-package/common/common.import.styl
```

The `importer.readFile` function then tries to find a match for this file in the `allFiles` `Map`, but it can't because the `allFiles` `Map` has its keys set to match the filename from the root of the package, not the root of the app. So `allFiles` is pointing to this file using a key of `common/common.import.styl`.

We could fix this by adding a quick hack like the following to the `importer.readFile` function:

```js
readFile(filePath) {
  const isAbsolute = filePath[0] === '/';
  const isNib = filePath.indexOf('/node_modules/nib/lib/nib/') !== -1;
  const isStylusBuiltIn = filePath.indexOf('/node_modules/stylus/lib/') !== -1;

  if (isAbsolute || isNib || isStylusBuiltIn) {
    // absolute path? let the default implementation handle this
    return fs.readFileSync(filePath, 'utf8');
  }
  
  // --- START HACK 
  const packageName = inputFile.getPackageName();
  const cleanFilePath =
    filePath.startsWith(`packages/${packageName}/`)
      ? filePath.replace(`packages/${packageName}/`, '')
      : filePath;
  // --- STOP HACK 

  const parsed = parseImportPath(cleanFilePath);
  const absolutePath = absoluteImportPath(parsed);

  referencedImportPaths.push(absolutePath);

  if (! allFiles.has(absolutePath)) {
    throw new Error(
      `Cannot read file ${absolutePath} for ${inputFile.getDisplayPath()}`
    );
  }

  return allFiles.get(absolutePath).getContentsAsString();
}
```

This will make sure the package path prefix is removed and the file can be properly matched in the `allKeys` `Map`. I've tested this approach and it works, but I really don't like it.

I think there is a bigger problem here with regards to the stylus lookup paths not being set properly. We should be able to set stylus paths to include package locations, so they're included when looking up imports, instead of hacking things like this. I followed this approach a bit only to quickly find myself down a large rabbit hole. I knew that Meteor was using its own fork of stylus (https://github.com/meteor/stylus), but I didn't realize that one of the main reasons this fork exists, is because of https://github.com/stylus/stylus/pull/1941. Meteor's stylus fork has this PR merged, which means there is another piece of this pie that needs to be reviewed and perhaps tweaked, to properly support globbing.

So, that's where I'm at with this. I'll commit my tweaks for 1 and 2 above, which will fix the outstanding test failures, and get this PR into a working state, but I don't think this is the best approach. I'll think about this further when I have more time, but for now I'll leave things like this.

One last thing - I still think we should consider flagging Meteor's `stylus` (and `less`) packages for deprecation. Better / more up to date 3rd party packages exist like [`coagmano:stylus`](https://atmospherejs.com/coagmano/stylus) and [`grove:less`](https://atmospherejs.com/grove/less), and there isn't really any technical reason why Meteor core needs to include its own `stylus` / `less` packages. They can be fully built and managed outside of core, and deprecating them helps reduce Meteor's maintenance overhead a bit more.

Thanks!

hwillson:
Hi @mitar - thanks for kicking these changes off initially. We've talked about all of this further, and have decided to officially deprecate the `stylus` package completely. The last published version will continue to work, but `stylus` will no longer be supported/maintained as part of the Meteor core. See https://github.com/meteor/meteor/pull/9445 for details. Thanks!

PR #9309, opened 2017-11-06
Use Package._define to reduce package boilerplate.

benjamn:
This removes the need for
```js
if (typeof Package === 'undefined') Package = {};
```
at the bottom of every package, and replaces
```js
(function (pkg, symbols) {
  for (var s in symbols)
    (s in pkg) || (pkg[s] = symbols[s]);
})(Package.minimongo = exports, {
  LocalCollection: LocalCollection,
  Minimongo: Minimongo,
  MinimongoTest: MinimongoTest,
  MinimongoError: MinimongoError
});
```
with
```js
Package._define("minimongo", exports, {
  LocalCollection: LocalCollection,
  Minimongo: Minimongo,
  MinimongoTest: MinimongoTest,
  MinimongoError: MinimongoError
});
```
It's not a huge savings, but it opens the door for us to do more interesting things with the `Package` namespace, by adding other methods to it.

### DIFF
--- a/packages/meteor/define-package.js
+++ b/packages/meteor/define-package.js
@@ -1,2 +1,26 @@
+function PackageRegistry() {}
+
+var PRp = PackageRegistry.prototype;
+
+// Set global.Package[name] = pkg || {}. If additional arguments are
+// supplied, their keys will be copied into pkg if not already present.
+// This method is defined on the prototype of global.Package so that it
+// will not be included in Object.keys(Package).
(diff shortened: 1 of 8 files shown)


zimme:
Will `Package.minimongo` still be defined with this change? I know I check for `Package.<package_name>` when I check for packages that are defined as weak dependencies in some of my packages.

benjamn:
@zimme Yep, the runtime state of `Package.*` remains the same!

Issue #9308, opened 2017-11-06
[1.6] Published package with build plugin does not work anymore on any pre 1.6 version


wojtkowiak:
After publishing omega:meteor-desktop-bundler with 1.6 I got when using the plugin with 1.5.2.2
```
=> App running at: http://localhost:3000/
   Type Control-C twice to stop.

=> Errors prevented startup:

   While loading plugin `meteor-desktop-bundler` from package `omega:meteor-desktop-bundler`:
   vm.js:24:10: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
   at Object.exports.createScript (vm.js:24:10)
```
Similarly I tried to use meteor-css-modules@2.8.0 which was published with 1.6 and could not because of similar problem.
https://github.com/nathantreid/meteor-css-modules/issues/107

It seems that the build plugins were not transpiled anymore even though in `use` they depend on old `ecmascript` version. I can guess that it is because server side is no longer transpiled (cause of Node 
8). Maybe I got this part wrong but I was always expecting the `Package.use` section as the way to enforce compatibility with older versions

After publishing with 1.5 everything works as expected.

benjamn:
An `api.use` version constraint restricts the major version of the package, but otherwise just sets a lower bound on the version of the package that can be used. For example, since `ecmascript` is still at major version 0, a constraint like `ecmascript@0.8.0` is totally compatible with `ecmascript@0.9.0` (the latest version), since the major version is the same, and 9 ≥ 8.

In order to maintain compatibility with Meteor 1.5 apps when publishing your package, you can specify a release version for the `meteor publish` command:
```sh
meteor --release 1.5.2.2 publish
```
If you want to take advantage of Meteor 1.6, I would recommend creating a new branch where you bump your package's minor or major version, then publish using Meteor 1.6 from that branch, and publish using Meteor 1.5 from the original branch. The version bump is important so that you can continue publishing patch updates for Meteor 1.5.

wojtkowiak:
@benjamn thank you very much for your time explaining and for the idea how to approach this.
I think it's understandable and now that I fully understand this I can go on. 
It might be worthy however to always include a note for the package maintainers in the history.md to warn about it as I've already found 3 authors who published with 1.6 and were not aware that they unintentionally dropped support for 1.4/1.5.
Thanks once again for your time!

hwillson:
Great idea @wojtkowiak - I've added a quick mention of this in https://github.com/meteor/meteor/pull/9314. Thanks!

PR #9314, opened 2017-11-07
Add note about possible package incompatibilty with 1.5

hwillson:
As mentioned in https://github.com/meteor/meteor/issues/9308#issuecomment-342447058, it would be great if we highlight the fact that it's possible to publish backwards incompatible packages using Meteor 1.6. I've updated the `History.md` with some plagiarism from @benjamn's comments in https://github.com/meteor/meteor/issues/9308#issuecomment-342295413.

Fixes https://github.com/meteor/meteor/issues/9308.

### DIFF
--- a/History.md
+++ b/History.md
@@ -19,6 +19,26 @@
 
 ## v1.6, 2017-10-30
 
+* **Important note for package maintainers:**
+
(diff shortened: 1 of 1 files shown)


benjamn:
Thanks @hwillson!

PR #9315, opened 2017-11-07
Upgrade webapp to latest connect + middleware packages.

jbaxleyiii:
@benjamn @abernix this is a first step towards upgrading webapp. I think the work around the server render package upgrades and the http2 should be new PR's

The formatting changes are from prettier but they seem to align with some other packages. Happy to change it back!

### DIFF
--- a/packages/webapp/package.js
+++ b/packages/webapp/package.js
@@ -1,12 +1,17 @@
 Package.describe({
   summary: "Serves a Meteor app over HTTP",
-  version: '1.4.0'
+  version: '1.5.0'
 });
 
-Npm.depends({connect: "2.30.2",
-             parseurl: "1.3.0",
-             send: "0.13.0",
-             useragent: "2.0.7"});
+Npm.depends({"basic-auth-connect": "1.0.0",
+             "cookie-parser": "1.4.3",
+             connect: "3.6.5",
+             compression: "1.7.1",
+             errorhandler: "1.5.0",
+             parseurl: "1.3.2",
+             send: "0.16.1",
+             "qs-middleware": "1.0.3",
+             useragent: "2.2.1"});
 
 Npm.strip({
   multiparty: ["test/"],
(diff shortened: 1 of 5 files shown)


packages/webapp/webapp_client.js
```diff
@@ -1,18 +1,21 @@
...
 export const WebApp = {
   _isCssLoaded() {
     if (document.styleSheets.length === 0) {
       return true;
     }
 
-    return _.find(document.styleSheets, sheet => {
-      if (sheet.cssText && ! sheet.cssRules) { // IE8
-        return ! sheet.cssText.match(/meteor-css-not-found-error/);
+    return find(document.styleSheets, sheet => {
```
abernix: Do we gain anything from `lodash`'s `find` versus converting `document.styleSheets` to an array with `Array.from` and using the native `Array.prototype.find`?

```js
Array.from(document.styleSheets).find(sheet => {
...
```

packages/webapp/package.js
```diff
@@ -1,49 +1,62 @@
...
+  api.mainModule("webapp_cordova.js", "web.cordova");
 });
 
-Package.onTest(function (api) {
-  api.use(['tinytest', 'ecmascript', 'webapp', 'http', 'underscore']);
-  api.addFiles('webapp_tests.js', 'server');
-  api.addFiles('webapp_client_tests.js', 'client');
-  api.addFiles('socket_file_tests.js', 'server');
+Package.onTest(function(api) {
+  api.use(["tinytest", "ecmascript", "webapp", "http", "underscore"]);
```
abernix: Not opposed to most of the `prettier` changes, but any possibility to get them in a commit by themselves, just prior to the other changes? :smile:

packages/webapp/webapp_client.js
benjamn: Or even just `Array.prototype.find.call(document.styleSheets, sheet => {`

packages/webapp/package.js
benjamn: Yeah, I know it may be hard to un-prettier the code now, but please please don't mix a reformatting commit together with actual changes. Would it be possible to split the commits now?

jbaxleyiii: @benjamn I'll just revert and reapply the technical changes (they are small) and disable prettier for working on the meteor code base 👍

packages/webapp/webapp_client.js
jbaxleyiii: happily

PR #9316, opened 2017-11-07
Remove SockJS [do not merge]

stubailo:
In this PR, I delete SockJS, shaving about 30kb off of the core Meteor bundle (uncompressed).

Don't merge this PR yet, there are still a few things to look into.

Thankfully, it looks like SockJS is a drop-in Websocket replacement, so it was pretty easy to do. Now, we just need to figure out which browsers and situations we might need a polyfill for, and have a mechanism to load that.

### Todo

- [ ] Make sure URL construction is reasonable
- [ ] Figure out when SockJS is still needed
- [ ] Figure out how to load SockJS when necessary
- [ ] Clean up code


packages/ddp-client/stream_client_common.js
```diff
@@ -22,6 +22,10 @@ var translateUrl =  function(url, newSchemeBase, subPath) {
     newSchemeBase = "http";
   }
 
+  if (subPath !== "sockjs" && startsWith(url, "/")) {
```
stubailo: Really interesting that `toWebsocketUrl` didn't work as-is - maybe the way server-side faye-websocket client works is different from the browser native Websocket.

benjamn:
Closing in favor of #9353. Thanks so much for demonstrating the feasibility of this, @stubailo!

PR #9325, opened 2017-11-08
Run prettier on ddp-client

stubailo:
Ran prettier with default settings!


sebakerckhof:
I though it was recommended to follow the airbnb style guide? This does some things quite different, e.g. double quotes for strings

stubailo:
Yeah, it was - but I think we can save ourselves a ton of time by using Prettier instead, and the default settings seem fine; we could switch back to single quotes if we feel that is important.

Curious what @benjamn and @abernix think!

PR #9329, opened 2017-11-08
Update references to `cordova_lib.raw` to account for `raw` deprecation.

abernix:
The changes made in #9213 were seemingly innocent, but when merging I missed the fact that the `cordova-lib` package bump was done to a new entry on the 'dev bundle', rather than on the `CORDOVA_DEV_BUNDLE_VERSIONS` object in tools/cordova/index.js which, as of meteor/meteor#8976, is responsible for auto-installing Cordova when it is used rather than including it in the pre-packaged 'dev bundle'.

That mistake was fixed with 958c44ff1b96c298c35ebe42ffb1755ef7e50a0f and d6adc1b3a92071eaa084de25b6a9de49ef23160f, but not before it caused the Cordova tests to falsely pass on the PR since it was functionally still testing the previous version of Cordova, 7.0.0.

Unfortunately, one of the changes in the 7.1.0 was the deprecation of an API we use within Meteor: the `raw` API on `cordova_lib`. Luckily, as seen on the following commit, that change was merely a re-organizational commit and still provides us access to that API by simply removing the `raw.` segment and accessing the various methods directly:

https://github.com/apache/cordova-lib/commit/90b6857f4d6c9d6169f1155d533ce4f7c487174c

Never saw that deprecation message, but we certainly saw the failure in CI: https://circleci.com/gh/meteor/meteor/10412.

With any luck, this commit/PR will fix the problem.

### DIFF
--- a/tools/cordova/project.js
+++ b/tools/cordova/project.js
@@ -171,7 +171,7 @@ outdated platforms`);
       this.runCommands('creating Cordova project', async () => {
         // No need to pass in appName and appId because these are set from
         // the generated config.xml
-        await cordova_lib.raw.create(files.convertToOSPath(this.projectRoot),
+        await cordova_lib.create(files.convertToOSPath(this.projectRoot),
           undefined, undefined, config);
       }, undefined, null);
     }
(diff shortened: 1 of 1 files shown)


benjamn:
:tada: for accessing the API directly, rather than `raw`ly

benjamn:
Merged as https://github.com/meteor/meteor/commit/235d00e2c7b2d4ad55b92f0c350ac7edfa4b237b, since the failing `add cordova platforms` test is now passing.

PR #9330, opened 2017-11-09
Fix for UNIX sockets with node cluster.

vlasky:
Prevents cluster worker processes creating UNIX socket files with the same name as the one used by the cluster master. Code now detects whether the server-side meteor instance is a worker (forked) instance. If it is, it will append the worker id (an integer) to the name of the socket file. For example, if the socket file is "meteor.sock", worker id 3's socket file will be named "meteor.sock.3.sock".

The reason why this is necessary is because the Node.js cluster fork() does not work like POSIX fork(). The worker process(es) will begin executing from the beginning and NOT from the line of code after fork() is called.

Yes, there are other ways of handling this situation, but we needed an urgent fix to our production code after updating to Meteor 1.6 and I wanted to start the conversation by submitting this pull request.

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -11,6 +11,7 @@ import connect from "connect";
 import parseRequest from "parseurl";
 import { lookup as lookupUserAgent } from "useragent";
 import send from "send";
+import cluster from "cluster";
 import {
   removeExistingSocketFile,
   registerSocketFileCleanup,
@@ -886,6 +887,9 @@ function runWebAppServer() {
     const unixSocketPath = process.env.UNIX_SOCKET_PATH;
 
     if (unixSocketPath) {
+      if (cluster.isWorker) {
+        unixSocketPath += "." + cluster.worker.id.toString() + ".sock";
+      }
       // Start the HTTP server using a socket file.
       removeExistingSocketFile(unixSocketPath);
       startHttpServer({ path: unixSocketPath });
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -888,7 +888,7 @@ function runWebAppServer() {
 
     if (unixSocketPath) {
       if (cluster.isWorker) {
-        unixSocketPath += "." + cluster.worker.id.toString() + ".sock";
+        unixSocketPath += "." + cluster.worker.id + ".sock";
       }
       // Start the HTTP server using a socket file.
       removeExistingSocketFile(unixSocketPath);


packages/webapp/webapp_server.js
```diff
@@ -886,6 +887,9 @@ function runWebAppServer() {
     const unixSocketPath = process.env.UNIX_SOCKET_PATH;
 
     if (unixSocketPath) {
+      if (cluster.isWorker) {
+        unixSocketPath += "." + cluster.worker.id.toString() + ".sock";
```
benjamn: You don't need the explicit `.toString()` call if you're concatenating with string literals.

vlasky: Fixed.

abernix:
I'm not sure how popular the usage of this will be, but considering altering the behavior later is likely more difficult than a more refined solution up front, can you elaborate on your "other ways of handling this situation" thoughts? I think they're worth discussing, even if you don't have time to implement them.

My initial thoughts are that it's less desirable to write to a file which wasn't specifically the one passed (with `UNIX_SOCKET_PATH`), though I know it's not reasonable to pass multiple names in for each worker, so some sort of automatic demuxing to different filenames (for example, as you've done, through automatic extension suffixing) is necessary.  Spitballing: `UNIX_SOCKET_PATH` as a directory for workers to write their file to.

vlasky:
@abernix upon further review, this issue impacts everyone who uses cluster to scale their Meteor webapps, regardless of whether their app is listening on a UNIX socket or the default TCP socket. That's because one or more worker processes will attempt to start listening on the same socket file or same port as the cluster master.

I will comment later on the other issues.

vlasky:
The only other idea I had for dealing with this issue was for the cluster master to pass an environment variable or command-line argument to the worker process to instruct it to not listen on that port or socket.

This would be a potentially suitable approach to use in our app because we are only using the worker processes to crunch database data in the background and not to handle HTTP requests.

For this approach, code would be added to webapp_server.js that checks to see if that command-line option or environment variable is defined and if so, not to listen on the UNIX socket.

We did not pursue this alternative approach because we concluded that:

*The resource overhead from creating unused UNIX socket files for the worker processes is insignificant

*Many other people's Meteor applications that use node cluster would be intending to use the worker processes to handle HTTP requests. Having the worker id in the socket file name is a fully reliable and robust way of identifying them.

However, in the case of someone who has a webapp that uses node cluster with multiple Meteor worker processes  listening on a TCP socket, there are considerations:

*When running under Linux, it is possible have multiple processes bind to the same TCP socket if they set the socket option SO_REUSEPORT. In that case, the Linux kernel will distribute requests amongst them in a round-robin fashion. Unfortunately, the socket option SO_REUSEPORT is not supported under Windows.

*The alternative is to have each worker processes listen on a different TCP port, perhaps chosen by taking the master's configured TCP port and adding the worker process id to it, e.g. If the cluster master listens on TCP port 4000, the worker process with id 7 would listen on port 4007.

My current (selfish) position is that I am happy for the current pull request that supports node cluster with UNIX sockets to be merged and that TCP support would be added later.

I would completely understand if you wanted the code to support both before merging.

abernix:
Thanks for taking the time to open this and to document your thoughts about it.  I'm happy this is working for you; luckily it's fairly easy to keep a local fork of `webapp` in `packages`, so you'll be able to keep this modification for your own purposes.

I'm going to close this for now.  Other maintainers don't currently have the resources to pick this up and finish the more complete design we'd be looking for.  We can always recover this PR in the future if others come forward with a similar problem (and hopefully additional perspective might make the solution more obvious).  As with PRs for new features in general, it's best if there is a feature request opened first so we can track and design the implementation– free to open that so we can still track +1's on it!

vlasky:
I have created a new branch [vlasky:cluster](https://github.com/vlasky/meteor/tree/cluster)  to track this enhancement.

In commit https://github.com/vlasky/meteor/commit/e0d0df39625b7c7de1f2ca8be95ed2fbb02c8314 the following enhancement has been made:

The worker process name that is used to name each worker process's UNIX socket file can now be overridden with the environment variable NAME.

This is useful because the worker id, which is used by default to name the UNIX socket file, is incremented each time the worker process ends/dies. This will cause difficulties if you are referencing the UNIX socket file from a web server or other external process where you expect the socket file to always have the same name.

vlasky:
@filipenevola should I create a new pull request for this enhancement to fix Meteor UNIX socket support when using the node cluster module, or would it be better to reopen this one?

filipenevola:
@vlasky I would prefer a new one.

vlasky:
> @vlasky I would prefer a new one.

Done.

https://github.com/meteor/meteor/pull/11369

PR #9334, opened 2017-11-09
 Allow scoped Cordova packages to be installed

### LINKED ISSUE #7336: Adding Cordova plugin from private registry fails

hwillson:
Hi all - this PR is a continuation of the closed PR https://github.com/meteor/meteor/pull/7350, that fixes an issue preventing the installation of scoped Cordova packages. For example, `meteor add cordova:@somescope/some-cordova-plugin@1.0.0` will now work properly.

I've updated the closed PR to include the requested code review changes (with a few small modifications), and added some tests. 

Fixes #7336. Thanks!


tools/cordova/package-id-version-parser.js
```diff
@@ -0,0 +1,27 @@
...
+    const parts = packageIdAndVersion.match(
+      /^(@[^\/]*)?(\/)?([^@]*)(@)?(.*)?/
+    );
+    const details = {
+      scope: parts[1],
+      scopeSeparator: parts[2],
+      packageName: parts[3],
+      versionSeparator: parts[4],
+      version: parts[5],
+    };
```
benjamn: Maybe skip the `details` object?
```js
const [
  _matchText,
  scope,
  scopeSeparator,
  packageName,
  versionSeparator,
  version,
] = packageIdAndVersion.match(
  /^(@[^\/]*)?(\/)?([^@]*)(@)?(.*)?/
);
```

tools/cordova/package-id-version-parser.js
```diff
@@ -0,0 +1,27 @@
...
+// and returns the ID and version in a package details object.
+//
+// Example `packageIdAndVersion` formats:
+// some-cordova-plugin@1.0.0
+// @somescope/some-cordova-plugin@1.0.0
+exports.parse = packageIdAndVersion => {
+  const package = {};
+  if (packageIdAndVersion) {
+    const parts = packageIdAndVersion.match(
+      /^(@[^\/]*)?(\/)?([^@]*)(@)?(.*)?/
```
abernix:
```js
/^(@[^\/]*)?(\/)?([^@]*)(@)?(.*)?/
         ^
```

When the scope is present in this manner, is it required that there be at least a character after the `@`?  If so,`+` ("1+ occurrence") would be more appropriate here than `*` ("0+ occurrences").  Similar question about the other `*` cases.*

<sub>
*Winner not guaranteed. ;-)
</sub>

tools/tests/cordova-plugins.js
```diff
@@ -611,3 +611,36 @@ selftest.define("cordova plugins in star.json, direct and transitive", ["slow",
...
 });
+
+selftest.define(
+  'parse cordova plugin ID and version',
+  ['cordova'],
+  function () {
+    const parseCordovaIdVersion =
+      require('../cordova/package-id-version-parser.js').parse;
+
+    let package = 'some-cordova-plugin';
```
abernix: Some `package` var concern as above.

tools/cordova/package-id-version-parser.js
hwillson: @benjamn I actually did exactly that first, but was greeted with the following

```
TypeError: Cannot read property 'Symbol(Symbol.iterator)' of packageIdAndVersion.match
```

I decided to switch it around instead of chasing that down further, but I'll investigate. Thanks!

tools/cordova/package-id-version-parser.js
```diff
@@ -0,0 +1,27 @@
...
+    );
+    const details = {
+      scope: parts[1],
+      scopeSeparator: parts[2],
+      packageName: parts[3],
+      versionSeparator: parts[4],
```
abernix: Maybe just not capture `versionSeparator`?

```js
/^(?:@([^\/]+)\/)?([^\/@]*)@?(.*)?/
```

tools/cordova/package-id-version-parser.js
abernix:
```js
/^(@[^\/]*)?(\/)?([^@]*)(@)?(.*)?/
            ^^^^^
```
Could this be included in the first grouping along with the scope and not explicitly captured? Right now this allows `/dd@` and puts `/` into `scopeSeparator` with no `scope`.

Maybe...
```js
/^(@[^\/]*\/?)?([^\/@]*)(@)?(.*)?/
```
... you'll note I also added `/` to the negated character class.

tools/cordova/package-id-version-parser.js
abernix: I could probably go further. Is there a way we could avoid the regex and just use a combination of `packageIdAndVersion.startsWith('cordova:)` and split on `packageIdAndVersion.lastIndexOf('@')`.

hwillson: And now for the life of me I can't re-create the same error. So, changes coming!

abernix: Given the tests, and that you may have already spent some time on this, and that overall this is an improvement, I think it's okay as is too if you just wanted to put your hands up. 🙌 

:wink:

tools/cordova/package-id-version-parser.js
```diff
@@ -0,0 +1,27 @@
+// Accepts a combined Cordova package ID + version string, then parses out
+// and returns the ID and version in a package details object.
+//
+// Example `packageIdAndVersion` formats:
+// some-cordova-plugin@1.0.0
+// @somescope/some-cordova-plugin@1.0.0
+exports.parse = packageIdAndVersion => {
+  const package = {};
```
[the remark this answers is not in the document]
hwillson: Good call - changing - thanks!

tools/cordova/package-id-version-parser.js
hwillson: Yes - also a good call! Changes coming.

hwillson: I split the separator out intentionally; I just liked the idea of having buckets for each individual component, so the scope (when available) was the actual scope name without the trailing `/`. But, there really isn't any reason to do this, so your point is well taken - changes coming!

tools/cordova/package-id-version-parser.js
hwillson: Sounds good!

tools/cordova/package-id-version-parser.js
hwillson: I originally avoided the regex and did just that, but the problem is that we need to catch instances where the version number isn't passed in. So the `parse` function can actually have `@somescope/some-cordova-plugin` passed in, which means splitting on `@` doesn't quite work (since we can have both a valid and invalid package string passed in that has only one `@`). We can definitely add code to address this, but it's at this point I decided a regex would probably make more sense. Happy to change it if you disagree though!

hwillson:
Review items should now be addressed - thanks guys! 

P.S. --> @abernix is a regex master! 🙌 🏆

tools/cordova/package-id-version-parser.js
abernix: Makes sense to me.  Thanks for the explanation!

abernix:
Squashed a bit and landed in `devel` as dd5ea2b9500250afc238eff0af27922044c61536 and 27ed9bc2e8a71df0263bcf811ddb183d19d41ba1.

Thanks!

PR #9337, opened 2017-11-09
Make ddp-client more aligned with code style

stubailo:
- single quotes
- 80 char line width

### DIFF
--- a/packages/ddp-client/common/namespace.js
+++ b/packages/ddp-client/common/namespace.js
@@ -13,25 +13,28 @@ LivedataTest.SUPPORTED_DDP_VERSIONS = DDPCommon.SUPPORTED_DDP_VERSIONS;
 // This is private but it's used in a few places. accounts-base uses
 // it to get the current user. Meteor.setTimeout and friends clear
 // it. We can probably find a better way to factor this.
-DDP._CurrentMethodInvocation = new Meteor.EnvironmentVariable;
-DDP._CurrentPublicationInvocation = new Meteor.EnvironmentVariable;
+DDP._CurrentMethodInvocation = new Meteor.EnvironmentVariable();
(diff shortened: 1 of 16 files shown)


packages/ddp-client/common/namespace.js
```diff
@@ -13,25 +13,28 @@ LivedataTest.SUPPORTED_DDP_VERSIONS = DDPCommon.SUPPORTED_DDP_VERSIONS;
...
 // XXX: Keep DDP._CurrentInvocation for backwards-compatibility.
 DDP._CurrentInvocation = DDP._CurrentMethodInvocation;
 
-DDP.ConnectionError = Meteor.makeErrorType(
-  "DDP.ConnectionError", function (message) {
-    var self = this;
-    self.message = message;
+DDP.ConnectionError = Meteor.makeErrorType('DDP.ConnectionError', function(
+  message
+) {
```
abernix: I find it quite interesting that `prettier` would wrap long lines in this manner (by splitting the `function` arguments across lines) when it opted for the following style (splitting on the right-hand side of an assignment) in code [further down in this review](https://github.com/meteor/meteor/pull/9337/files#diff-5fd7a402a518dc533b2f1c08652dbc57R43):

```js
      self.statusListeners =
        typeof Tracker !== 'undefined' && new Tracker.Dependency();
```
🤷‍♂️

packages/ddp-client/server/stream_client_nodejs.js
```diff
@@ -1,7 +1,7 @@
 import { _ } from 'meteor/underscore';
 import { Meteor } from 'meteor/meteor';
 
-import { DDP, LivedataTest } from "../common/namespace";
+import { DDP, LivedataTest } from '../common/namespace';
```
abernix: `.js` extension?

PR #9339, opened 2017-11-09
Fix meteor test file matching patterns

### LINKED ISSUE #9332: meteor test runs on more files than specified in the guide
Meteor v1.6, OS X

#### Expected behavior:

`meteor test` as per the [guide](https://guide.meteor.com/testing.html#test-modes):
> *Does* eagerly load any file in our application (including in `imports/` folders) that look like `*.test[s].*`, or `*.spec[s].*`

My understanding of that statement is that there must be a period before and after the word `test`, `tests`, `spec`, or `specs` in order for the file to be loaded.

#### Actual behavior:

`meteor test` loads files in which `test`, `tests`, `spec`, or `specs` are preceded by a period and which end in `.js` but ignores anything in between.

The following files are eagerly loaded, when I would have expected them to be ignored:

* `index.testt.js`
* `index.testtt.js`
* `index.testttttttttt.js`

Here is a [reproduction](https://github.com/jamesmillerburgess/meteor-extra-tests) with additional test cases.

So either:
1) I misunderstood the guide,
2) the guide should be adjusted, or
3) this is a bug

I'm happy to make a pull request for 2 or 3.

I would prefer a configurable way to grep the files, since I am using jest test files in my project structure as well. Luckily jest can configure this, so I am renaming my jest tests to avoid colliding with `meteor test` for now.

jamesmillerburgess:
**Update:** I've removed the additional feature from this PR, so it only focuses on fixing the issue below.

This PR fixes #9332, in that the guide says one thing for which files get eagerly loaded, but the regular expressions in `isobuild` match a wider pattern that doesn't seem intentional. Now the patterns match what is specified in the guide.

The potential downside of these more restrictive default patterns is that any tests currently built to use something like `foo.test-bar.js` will no longer be eagerly loaded. But the guide implies that such a filename shouldn't be used, and `meteor test` doesn't appear to be in the API documentation at all, so I'm thinking not many people should be relying on this behavior. I only came across this issue while trying to *avoid* my jest test files getting eagerly loaded by `meteor test`!

I've also added some unit and integration tests for different file loading scenarios, and cleaned up some help text identified by @axelvaindal: `--driver-package` is called optional, but `meteor test` throws an error without it.


jamesmillerburgess:
Did I mess something up with git and cause circleci to fall over?

abernix:
@jamesmillerburgess Thanks for opening this! I think the CI test failure was either a fluke or the tip of the branch changed between the push time and the CI spin-up—perhaps due toa force-push? Regardless, it looks good now so no need to worry.  We'll take a look at this as soon as possible!

axelvaindal:
@jamesmillerburgess 
Any reason for specifically writing each possible case manually in the regex array ?

```javascript
// We have two things "tests" and "app-tests".
export const TEST_FILENAME_REGEXPS = [
  // "*.test[s].*" or "*.spec[s].*"
  /\.test\./,
  /\.tests\./,
  /\.spec\./,
  /\.specs\./,
];

export const APP_TEST_FILENAME_REGEXPS = [
  // "*.app-test[s].*" or "*.app-spec[s].*"
  /\.app-test\./,
  /\.app-tests\./,
  /\.app-spec\./,
  /\.app-specs\./,
];
```

This writing seems shorter with same results : 

```javascript
// We have two things "tests" and "app-tests".
export const TEST_FILENAME_REGEXPS = [
  // "*.test[s].*" or "*.spec[s].*"
  /\.test(s\b|\b)\./,
  /\.spec(s\b|\b)\./,
];

export const APP_TEST_FILENAME_REGEXPS = [
  // "*.app-test[s].*" or "*.app-spec[s].*"
  /\.app-test(s\b|\b)\./,
  /\.app-spec(s\b|\b)\./,
];
```

Thanks for taking into account my comment about the helping message for the test driver and modifying accordingly btw 🙂.

jamesmillerburgess:
It would be shorter to combine the lines, but these patterns are explicitly excluded from the file reading in `package-source.js` in the appropriate 'mode' of running meteor. And that brings in an edge case.

If we used the combined regexps, then when a user then enters a pattern like `\.test\.` in as a custom pattern, no test files at all will be read because all files matching the pattern `\.test(s\b|\b)\.` get excluded based on the mode.

With the individual patterns listed out, we get around the edge case by matching the custom pattern against the predefined ones and then avoid excluding a match from the file reading.

axelvaindal:
Oh yes, my bad, didn't occur to me that you were trying to cover custom pattern with a new option on the command.
So now, based on mode, only specified regex are eagerly loaded unless `--test-match` is specified with proper pattern, is that correct ?
What happens if `--full-app` and `--test-match` are provided ? (both are eagerly load or `--test-match` overwrite the pattern ?)

If your goal is to provide as numerous mode as possible in testing field, shouldn't it be better to use only-one syntax with test mode.
What I'm trying to propose here is to simplify syntax as much as possible.

If we could define somewhere test patterns (`settings.json` file or elsewhere), which could be loaded, this could allow for this kind of writing in test command : 
```
meteor test 
meteor test --test-mode full-app|my-custom-mode
```
This should simplify testing easily, you would juste have to : 

- Use default mode if this is sufficient for you (or full-app which should be added to the file at `meteor create` for instance)
- Add a new custom mode to the file
```
testMode:
{
    "my-custom-mode": \.test-foo\.
}
```
- Execute `meteor test --test-mode my-custom-mode`

This is a proposal, what do you think about it ? 🙂

abernix:
Thank you for the contribution, @jamesmillerburgess.

I like the dialog and conversation going on here, but this is mixing a new feature with a bug fix at the moment and the feature necessitates some further discussion, which is why we ask that they be opened as feature requests first.  Additionally, the formatting changes (indentation, etc.) should be left out.  We don't want to put your name on the `git blame` history for things which aren't really related! :smile:

As you pointed out, I think it makes a lot of sense to make the test-file matching behavior match the documentation, even where this would otherwise be a breaking change for some.  Plus, I think the tests are a great addition.  Can you please revert the non-fix portion of this and open a [Meteor feature request](https://github.com/meteor/meteor-feature-requests/issues) for the rest?

jamesmillerburgess:
@abernix 

Thanks for reviewing this. I will try to get to it today. Those formatting changes were accidental - I guess I had autoformatting turned on for that file and didn't check the commit closely enough.

PR #9343, opened 2017-11-10
Improved server rendering

jbaxleyiii:
👋 With the release of React 16, a new way to render on the server is [streaming](https://hackernoon.com/whats-new-with-server-side-rendering-in-react-16-9b0d78585d67) the rendering of the react portion of your app to the client. This PR adjusts Meteor's internal markup generation for the server to move to a stream first API. With this change, you can use a string (renderToString) or a stream (renderToNodeStream) when adding content into an id.

In initial testing, it results in a 40% reduction of TTFB 🌮 (Time To First Byte [of taco]).

**Todo**
- [x] ensure support for multiple id based additions of streams of strings
- [x] update tests
- [x] add support for redirects
- [x] add support for changing status code
- [x] provide access to headers
- [ ] write tests for new features

Wanted to get this up early enough for review though!

This PR address [setting status codes](https://github.com/meteor/meteor-feature-requests/issues/210) including for [SEO](https://github.com/meteor/meteor-feature-requests/issues/198), improved [alignment with react features](https://github.com/meteor/meteor-feature-requests/issues/182), setting headers [like eTag](https://github.com/meteor/meteor-feature-requests/issues/176), and accessing [headers and cookies](https://github.com/meteor/meteor-feature-requests/issues/174)

### DIFF
--- a/packages/boilerplate-generator-tests/test-lib.js
+++ b/packages/boilerplate-generator-tests/test-lib.js
@@ -1,4 +1,4 @@
-import toString from "stream-to-string";
+import streamToString from "stream-to-string";
 
 export async function generateHTMLForArch(arch) {
   // Use a dummy manifest. None of these paths will be read from the filesystem, but css / js should be handled differently
@@ -52,7 +52,5 @@ export async function generateHTMLForArch(arch) {
     },
   });
 
-
-  return await toString(boilerplate.toHTML());
-};
-
+  return streamToString(boilerplate.toHTML());
+}
--- a/packages/boilerplate-generator-tests/test-lib.js
+++ b/packages/boilerplate-generator-tests/test-lib.js
@@ -1,4 +1,4 @@
-export function generateHTMLForArch(arch) {
+export async function generateHTMLForArch(arch) {
   // Use a dummy manifest. None of these paths will be read from the filesystem, but css / js should be handled differently
   const manifest = [
     {
@@ -50,5 +50,29 @@ export function generateHTMLForArch(arch) {
     },
   });
 
-  return boilerplate.toHTML();
+
+  const { start, stream, end } = boilerplate.toHTML();
+
+  const body = await toString(stream);
+  
+  return start + body + end;
 };
+
+
+function toString(stream) {
+  return new Promise((success, fail) => {
+    var string = ''
+    stream.on('readable', function(buffer) {
+      var part = buffer.read().toString();
+      string += part;
+    });
+
+    stream.on('end', function() {
+      success(string)
+    });
+
+    stream.on('error', function(error) {
+      fail(error);
+    });
+  });
+}
--- a/packages/boilerplate-generator-tests/test-lib.js
+++ b/packages/boilerplate-generator-tests/test-lib.js
@@ -1,3 +1,5 @@
+import toString from "stream-to-string";
+
 export async function generateHTMLForArch(arch) {
   // Use a dummy manifest. None of these paths will be read from the filesystem, but css / js should be handled differently
   const manifest = [
@@ -51,28 +53,6 @@ export async function generateHTMLForArch(arch) {
   });
 
 
-  const { start, stream, end } = boilerplate.toHTML();
-
-  const body = await toString(stream);
-  
-  return start + body + end;
+  return await toString(boilerplate.toHTML());
 };
 
-
-function toString(stream) {
-  return new Promise((success, fail) => {
-    var string = ''
-    stream.on('readable', function(buffer) {
-      var part = buffer.read().toString();
-      string += part;
-    });
-
-    stream.on('end', function() {
-      success(string)
-    });
-
-    stream.on('error', function(error) {
-      fail(error);
-    });
-  });
-}
--- a/packages/boilerplate-generator-tests/web.browser-tests.js
+++ b/packages/boilerplate-generator-tests/web.browser-tests.js
@@ -2,62 +2,69 @@ import { parse, serialize } from 'parse5';
 
 import { generateHTMLForArch } from './test-lib';
 
-const html = generateHTMLForArch('web.browser');
+const start = async () => {
+  const html = await generateHTMLForArch('web.browser');
 
-Tinytest.add("boilerplate-generator-tests - web.browser - well-formed html", function (test) {
-  const formatted = serialize(parse(html));
-  test.isTrue(formatted.replace(/\s/g, '') === html.replace(/\s/g, ''));
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - well-formed html", function (test) {
+    const formatted = serialize(parse(html));
+    test.isTrue(formatted.replace(/\s/g, '') === html.replace(/\s/g, ''));
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - include htmlAttributes", function (test) {
-  test.matches(html, /foo="foobar"/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - include htmlAttributes", function (test) {
+    test.matches(html, /foo="foobar"/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - escape htmlAttributes", function (test) {
-  test.matches(html, /gems="&amp;&quot;"/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - escape htmlAttributes", function (test) {
+    test.matches(html, /gems="&amp;&quot;"/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - include js", function (test) {
-  test.matches(html, /<script[^<>]*src="[^<>]*templating[^<>]*">/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - include js", function (test) {
+    test.matches(html, /<script[^<>]*src="[^<>]*templating[^<>]*">/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - escape js", function (test) {
-  test.matches(html, /<script[^<>]*src="[^<>]*templating[^<>]*&amp;v=&quot;1&quot;[^<>]*">/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - escape js", function (test) {
+    test.matches(html, /<script[^<>]*src="[^<>]*templating[^<>]*&amp;v=&quot;1&quot;[^<>]*">/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - include css", function (test) {
-  test.matches(html, /<link[^<>]*href="[^<>]*bootstrap[^<>]*">/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - include css", function (test) {
+    test.matches(html, /<link[^<>]*href="[^<>]*bootstrap[^<>]*">/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - escape css", function (test) {
-  test.matches(html, /<link[^<>]*href="[^<>]*bootstrap[^<>]*&amp;v=&quot;1&quot;[^<>]*">/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - escape css", function (test) {
+    test.matches(html, /<link[^<>]*href="[^<>]*bootstrap[^<>]*&amp;v=&quot;1&quot;[^<>]*">/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - call rewriteHook", function (test) {
-  test.matches(html, /\+rewritten_url=true/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - call rewriteHook", function (test) {
+    test.matches(html, /\+rewritten_url=true/);
+  });
 
-Tinytest.add("boilerplate-generator-tests - web.browser - include runtime config", function (test) {
-  test.matches(html, /<script[^<>]*>[^<>]*__meteor_runtime_config__ =.*decodeURIComponent\(config123\)/);
-});
+  Tinytest.add("boilerplate-generator-tests - web.browser - include runtime config", function (test) {
+    test.matches(html, /<script[^<>]*>[^<>]*__meteor_runtime_config__ =.*decodeURIComponent\(config123\)/);
+  });
 
-// https://github.com/meteor/meteor/issues/9149
-Tinytest.add(
-  "boilerplate-generator-tests - web.browser - properly render boilerplate " +
-  "elements when _.template settings are overridden",
-  function (test) {
-    import { _ } from 'meteor/underscore';
-    _.templateSettings = {
-      interpolate: /\{\{(.+?)\}\}/g
-    };
-    const newHtml = generateHTMLForArch('web.browser');
-    test.matches(newHtml, /foo="foobar"/);
-    test.matches(newHtml, /<link[^<>]*href="[^<>]*bootstrap[^<>]*">/);
-    test.matches(newHtml, /<script[^<>]*src="[^<>]*templating[^<>]*">/);
-    test.matches(newHtml, /<script>var a/);
-    test.matches(
-      newHtml,
-      /<script[^<>]*>[^<>]*__meteor_runtime_config__ =.*decodeURIComponent\(config123\)/
-    );
-  }
-);
+  // https://github.com/meteor/meteor/issues/9149
+  Tinytest.addAsync(
+    "boilerplate-generator-tests - web.browser - properly render boilerplate " +
+    "elements when _.template settings are overridden",
+    function (test, onComplete) {
+      const run = async () => {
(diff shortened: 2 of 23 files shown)


packages/boilerplate-generator/generator.js
```diff
@@ -23,13 +26,27 @@ export class Boilerplate {
...
+
+    const { body, dynamicBody } = data;
+    const stream = combine()
+      .append(body || "")
+      .append(dynamicBody || "")
+      .append(null);
+
+    const end = this.closeTemplate(data);
+
+    return { start, stream, end }
```
benjamn: Can't we just include the `start` and `end` chunks in the stream, so we don't have to return multiple things here?

packages/webapp/webapp_server.js
```diff
@@ -799,12 +803,22 @@ function runWebAppServer() {
...
         res.writeHead(statusCode, headers);
-        res.write(boilerplate);
-        res.end();
-      }, error => {
+        res.write(start);
+        stream.pipe(res, { end: false })
+        stream.on("end", () => {
+          res.write(end);
+          res.end();
+        })
```
benjamn: For example, if `start` and `end` were part of the `stream`, this code could just be
```js
stream.pipe(res, { end: true });
```

packages/webapp/webapp_tests.js
```diff
@@ -8,6 +8,24 @@ var hash = crypto.createHash('sha1');
 hash.update(additionalScript);
 var additionalScriptPathname = hash.digest('hex') + ".js";
 
+// convert a stream to a string via promise
+function toString(stream) {
```
benjamn: Alright, this code is duplicated three times now. Let's find a better home for it!

packages/webapp/webapp_tests.js
```diff
@@ -8,6 +8,24 @@ var hash = crypto.createHash('sha1');
...
+  return new Promise((success, fail) => {
+    var string = ''
+    stream.on('data', function(data) {
+      string += data.toString();
+    });
```
benjamn: I think it might be better if we appended the `data` buffers to an array and then used `Buffer.concat(chunks).toString("utf8")` at the end, since a chunk boundary could theoretically split a multi-byte unicode character.

packages/boilerplate-generator/package.js
```diff
@@ -3,6 +3,10 @@ Package.describe({
   version: '1.3.1'
 });
 
+Npm.depends({
+  "combine-streams": "1.0.0"
```
benjamn: I worry that the `combine-streams` package is not very heavily developed/starred/downloaded. In particular, it doesn't seem to pay any attention to the string encoding of the written chunks.

After a bit of digging, this [`combined-stream2`](https://www.npmjs.com/package/combined-stream2) package looks a bit more mature.

packages/webapp/webapp_tests.js
benjamn: [Example](https://github.com/meteor/meteor/blob/70ffce52fc36e91e188cc1d10a67a63f9969f5dd/tools/utils/http-helpers.js#L19) of other code that we use for a similar purpose.

packages/webapp/webapp_tests.js
benjamn: If you want to just use an npm package, [`concat-stream`](https://www.npmjs.com/package/concat-stream) looks good to me.

jbaxleyiii:
@benjamn thanks for the comments! I'll work to address them ASAP!

SachaG:
Just wanted to say it's awesome to see things moving so fast! Hope to be able to integrate this in Vulcan soon :)

hwillson:
@jbaxleyiii Since hack week is over and you're back busy on all things Apollo, just say the word and I'll jump in to finish this PR up. I'm super excited to get these changes out - thanks for working on this!

jbaxleyiii:
@hwillson if you have time that would be amazing! There are a number of apollo client 2.0 upgrade blockers that I need to put my full focus on.

I'm going to try to get all of the changes done by Friday but if not then if you and @benjamn can take it over it would be so great to see this one make it through. 

@benjamn does that timeline work? I want to make sure its ready for PR club next week

packages/boilerplate-generator/generator.js
jbaxleyiii: haha duh, wow should've done that from the start (see what I did there 😉)

abernix:
Next week is great, @jbaxleyiii. Thanks for your work on this! :smile:

jbaxleyiii:
@hwillson @benjamn @abernix everything should be addressed with the latest commit! Looking forward to seeing this in 1.6 😍

jbaxleyiii:
@benjamn of careful note is the removal of the boilerplate memoization (done on a separate commit to make it easy to [revert](https://github.com/meteor/meteor/pull/9343/commits/e300a33d5dbf112e50d5c62073851a2f3ba206ef)) 

I think I'm probably misunderstanding its goals, but given the potential dynamic nature of generating markup within an app, memoization of request => markup seems like a bad idea?

jbaxleyiii:
@hwillson one thing this is still missing is tests for redirection, setting status codes, and getting / setting headers. I don't think I'll have time to write those sadly :(

packages/boilerplate-generator-tests/test-lib.js
```diff
@@ -50,5 +52,7 @@ export function generateHTMLForArch(arch) {
     },
   });
 
-  return boilerplate.toHTML();
+
+  return await toString(boilerplate.toHTML());
```
benjamn: No need to explicitly `await` when returning a value that could be a `Promise` from an `async` function.

packages/boilerplate-generator-tests/web.browser-tests.js
```diff
@@ -2,62 +2,69 @@ import { parse, serialize } from 'parse5';
...
+        test.matches(newHtml, /foo="foobar"/);
+        test.matches(newHtml, /<link[^<>]*href="[^<>]*bootstrap[^<>]*">/);
+        test.matches(newHtml, /<script[^<>]*src="[^<>]*templating[^<>]*">/);
+        test.matches(newHtml, /<script>var a/);
+        test.matches(
+          newHtml,
+          /<script[^<>]*>[^<>]*__meteor_runtime_config__ =.*decodeURIComponent\(config123\)/
+        );
+      }
+      run().then(onComplete);
```
benjamn: I think you need to handle the error case here, too, so the test doesn't time out if there's an error.

packages/boilerplate-generator-tests/web.browser-tests.js
```diff
@@ -2,62 +2,69 @@ import { parse, serialize } from 'parse5';
...
-    test.matches(newHtml, /<script[^<>]*src="[^<>]*templating[^<>]*">/);
-    test.matches(newHtml, /<script>var a/);
-    test.matches(
-      newHtml,
-      /<script[^<>]*>[^<>]*__meteor_runtime_config__ =.*decodeURIComponent\(config123\)/
-    );
-  }
-);
+const start = async () => {
+  const html = await generateHTMLForArch('web.browser');
```
benjamn: Rather than using an `async` function here, I would just do
```js
generateHTMLForArch("web.browser").then(html => {
  Tinytest.add(...);
  ...
});
```
I'm not entirely sure what happens if the tests finish while there's still a pending `Promise`, so you might want to capture the resulting promise and add an additional
```js
Tinytest.addAsync("everything ok", function (test, onComplete) {
  promise.then(
    () => onComplete,
    error => test.fail(error)
  );
});
```
at the end of the file.

packages/boilerplate-generator/generator.js
```diff
@@ -23,13 +26,30 @@ export class Boilerplate {
...
-      this.template({ ...this.baseData, ...extraData });
+    const data = {...this.baseData, ...extraData};
+    const start = "<!DOCTYPE html>\n" + this.headTemplate(data);
+
+    const { body, dynamicBody } = data;
+
+    const end = this.closeTemplate(data);
+    const response = stream.create();
+    
+    response.append(Buffer.from(start))
```
benjamn: Let's be explicit about passing `"utf8"` as the encoding type as the second argument to `Buffer.from`.

packages/boilerplate-generator/generator.js
```diff
@@ -23,13 +26,30 @@ export class Boilerplate {
...
+    if (body) {
+      response.append(typeof body === "string" ? Buffer.from(body) : body)
```
benjamn: What else can `body` be? Let's enforce our assumptions here using some combination of `typeof body === "string"`, `Buffer.isBuffer`, and `typeof body.read === "function"` (if that covers all the cases).

packages/boilerplate-generator/template-web.cordova.js
```diff
@@ -11,69 +11,65 @@ export default function({
...
-      '',
-      '<body>',
-      body,
-      '</body>',
-      '</html>'
-    ],
-  ).join('\n');
-}
+export const closeTemplate = () => 
+  [].concat([ '</body>','</html>' ]).join('\n');
```
benjamn: I would prefer that this was simply `=> "</body>\n</html>"`.

packages/server-render/server-register.js
```diff
@@ -1,9 +1,11 @@
 import { WebAppInternals } from "meteor/webapp";
-import { SAXParser } from "parse5";
 import MagicString from "magic-string";
-import { ServerSink } from "./server-sink.js";
+import { SAXParser } from "parse5";
+import streamer from "combined-stream2";
```
benjamn: Would `import { create as createStream } from "combined-stream2" work here?

packages/server-render/server-register.js
```diff
@@ -29,22 +31,40 @@ WebAppInternals.registerBoilerplateDataCallback(
...
+          let lastStart = magic.start;
+          parser.on("startTag", (name, attrs, selfClosing, loc) => {
+            attrs.some(attr => {
+              if (attr.name === "id") {
+                let html = sink.htmlById[attr.value];
+                if (html) {
+                  reallyMadeChanges = true;
+                  const start = magic.slice(lastStart, loc.endOffset);
+                  stream.append(Buffer.from(start));
+                  stream.append(typeof html === "string" ? Buffer.from(html) : html)
```
benjamn: More cases to be explicit about `utf8`.

packages/server-render/server-render-tests.js
```diff
@@ -2,6 +2,7 @@ import { Tinytest } from "meteor/tinytest";
 import { WebAppInternals } from "meteor/webapp";
 import { onPageLoad } from "meteor/server-render";
 import { parse } from "parse5";
+import toString from "stream-to-string";
```
benjamn: Let's actually call this `streamToString` for better readability below.

packages/webapp/webapp_server.js
```diff
@@ -307,34 +305,11 @@ function getBoilerplateAsync(request, arch) {
     });
   });
 
-  return promise.then(() => {
-    const useMemoized = ! (
-      data.dynamicHead ||
-      data.dynamicBody ||
-      madeChanges
```
benjamn: The reason I left the memoization in place when I originally implemented the `server-render` package was because it could be better for performance when no server-side rendering is happening.

That said, I think it's important to make sure the no-cache path is fast enough, since we want people to do SSR, and we want SSR to be fast.

Assuming we want to remove the memoization completely, I think there are a few more opportunities for simplification. In particular, there's no point in worrying about the `madeChanges` variable, and we can stop pre-generating boilerplate elsewhere in this module, e.g. here:
```js
WebAppInternals.setBundledJsCssUrlRewriteHook = function (hookFn) {
  bundledJsCssUrlRewriteHook = hookFn;
};
```

packages/webapp/webapp_tests.js
```diff
@@ -1,13 +1,15 @@
 var url = Npm.require("url");
 var crypto = Npm.require("crypto");
 var http = Npm.require("http");
+var toString = Npm.require("stream-to-string");
```
benjamn: `streamToString` please

packages/webapp/webapp_tests.js
```diff
@@ -1,13 +1,15 @@
 var url = Npm.require("url");
```
benjamn: Let's just switch these tests to depending on `ecmascript` so we can use `import`s (or at least `require`) here.

packages/webapp/webapp_tests.js
```diff
@@ -66,132 +68,149 @@ Tinytest.add("webapp - content-type header", function (test) {
...
 
-Tinytest.add("webapp - additional static javascript", function (test) {
-  var origInlineScriptsAllowed = WebAppInternals.inlineScriptsAllowed();
-
-  var staticFilesOpts = {
-    staticFiles: {},
-    clientDir: "/"
-  };
+Tinytest.addAsync("webapp - additional static javascript", function (test, onComplete) {
+  const run = async () => {
```
benjamn: I think you can just make the whole test function `async`, and call `onComplete` after the last `await`. I don't think `Tinytest` will do anything smart with the resulting `Promise` object, but that's fine.

packages/webapp/webapp_server.js
```diff
@@ -799,12 +774,17 @@ function runWebAppServer() {
...
         archKey
-      ).then(boilerplate => {
-        var statusCode = res.statusCode ? res.statusCode : 200;
+      ).then(({ stream,  statusCode, headers: newHeaders }) => {
+        if (!statusCode) {
+          statusCode = res.statusCode ? res.statusCode : 200;
+        }
+        if (newHeaders) {
+          headers = {...headers, ...newHeaders };
+        }
```
benjamn: Let's just use `Object.assign(headers, newHeaders)` here.

packages/webapp/webapp_server.js
```diff
@@ -799,12 +774,17 @@ function runWebAppServer() {
...
+
         res.writeHead(statusCode, headers);
-        res.write(boilerplate);
-        res.end();
-      }, error => {
+        stream.pipe(res)
```
benjamn: Does this need to be `stream.pipe(res, { end: true })`?

packages/boilerplate-generator-tests/web.browser-tests.js
benjamn: We should really just improve `Tinytest` to handle `Promise`s better, honestly, but that's a mission for a different PR.

packages/boilerplate-generator/generator.js
benjamn: The internet tells me `body instanceof ReadableStream` isn't always safe, since folks can implement their own stream-like objects, which is why I suggested the dynamic duck test that `body.read` is a function. Maybe extract this logic into a helper function so you can do `appendToStream(response, body)`?

packages/server-render/server-register.js
jbaxleyiii: I'll try and see if that works!

jbaxleyiii:
@benjamn thanks for the further notes! I'll try to get them done asap!

packages/webapp/webapp_server.js
benjamn: I see now why it's hard to memoize a stream, so I've come around to these changes. 👌

benjamn:
@jbaxleyiii Thanks for all your work on this. Since you're busy now, I went ahead and addressed my latest feedback, and in the process I was inspired to improve `Tinytest.addAsync` to enable the kind of tests you wanted to write (see #9409). :tada:

williamli:
I tried to replace `renderToString` with `renderToNodeStream` in my React SSR.
The browser ended up stuck waiting for the stream to start (stuck in loading, but nothing ever arrived).

I have Helmet, styled-components, and node-cache on the server side logic.
After removing all of them, leaving just the bare essentials 

`sink.renderIntoElementById("react-app", <App target="server" location={sink.request.url.path} visitorGeoDetails={visitorGeoDetails} setPreloadedContent={setPreloadedContent} />);`

The problem still remains.

Is there any other thing I need to do when switching over to `renderToNodeStream` ?

williamli:
I tried to replace `renderToString` with `renderToNodeStream` in my React SSR.
The browser ended up stuck waiting for the stream to start (stuck in loading, but nothing ever arrived).

I have Helmet, styled-components, and node-cache on the server side logic.
After removing all of them, leaving just the bare essentials 

`sink.renderIntoElementById("react-app", renderToNodeStream(<App target="server" location={sink.request.url.path} visitorGeoDetails={visitorGeoDetails} setPreloadedContent={setPreloadedContent} />));`

The problem still remains.

Is there any other thing I need to do when switching over to `renderToNodeStream` ?

PS: I see the server ran through the React Components (and if there were any warnings from propTypes, they also showed up in the server terminal console log when a browser try to access the site).

evolross:
It's been a minute but the removal of the boilerplate memoization is causing `setBundledJsCssUrlRewriteHook` to fire a lot (and apparently infinitely looping in production on Galaxy). Wanting to make sure there's no negative consequence. #10731

Issue #9146, opened 2017-09-28
Extracted archive X should only contain one entry


dossorio:
I have been facing this issue since I started working with meteor, in different occasions and always when running commands that altered the installed packages. I am using meteor on Windows as you see in the error trace. The most annoying thing is, and that's why I can't give you steps to reproduce, that it doesn't always happen and sometimes it just vanishes after running the same command several times. I also tried what is suggested here https://github.com/meteor/meteor/issues/7686 without success.

In this occasion, I get it after running `meteor update`:
```
R:\app>meteor update
This project is already at Meteor 1.5.2.1, the latest release.
C:\Users\User\AppData\Local\.meteor\packages\meteor-tool\1.5.2_1\mt-os.windows.x86_32\dev_bundle\lib\node_modules\meteor-promise\promise_server.js:218
      throw error;
      ^

Error: Extracted archive '/C/Users/User/AppData/Local/Temp/.tmpzgx7tb' should only contain one entry
    at Object.files.extractTarGz (C:\tools\fs\files.js:762:11)
    at Object.exports._extractAndConvert (C:\tools\packaging\tropohouse.js:50:9)
    at C:\tools\packaging\tropohouse.js:505:42
    at C:\tools\utils\buildmessage.js:359:18
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at C:\tools\utils\buildmessage.js:352:34
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at C:\tools\utils\buildmessage.js:350:23
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at Object.enterJob (C:\tools\utils\buildmessage.js:324:26)
    at C:\tools\packaging\tropohouse.js:502:24
    at Array.forEach (native)
    at Function._.each._.forEach (C:\Users\User\AppData\Local\.meteor\packages\meteor-tool\1.5.2_1\mt-os.windows.x86_32\dev_bundle\lib\node_modules\underscore\underscore.js:79:11)
    at C:\tools\packaging\tropohouse.js:469:11
    at C:\tools\utils\buildmessage.js:359:18
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
    at C:\tools\utils\buildmessage.js:352:34
    at [object Object]._.extend.withValue (C:\tools\utils\fiber-helpers.js:89:14)
[...]

PR #9344, opened 2017-11-10
Fixes should only contain one entry error

dossorio:
Fixes https://github.com/meteor/meteor/issues/9146

After investigating the problem, the problem observed is as follows (Only windows):

- Files created in the 7z.exe extraction process of a package, `out.tar.gz` and `out.tar` are removed here https://github.com/meteor/meteor/blob/devel/tools/fs/files.js#L812
- Then sometimes, windows fs doesn't catch up in time, and when `readdir` is called here https://github.com/meteor/meteor/blob/devel/tools/fs/files.js#L721, it returns the wrong file list, which includes the two that were deleted successfully earlier, which causes the exception to be throw here https://github.com/meteor/meteor/blob/devel/tools/fs/files.js#L728 

What we are observing is explained here https://github.com/nodejs/node/issues/4760#issuecomment-173174694 and also referenced here https://github.com/nodejs/node-v0.x-archive/issues/7176 so this pr aims to mitigate it

Note that the fix only works in meteor 1.6 and the nodejs version that comes with it

### DIFF
--- a/tools/fs/files.js
+++ b/tools/fs/files.js
@@ -723,6 +723,14 @@ files.extractTarGz = function (buffer, destPath, options) {
     // PaxHeader directory.
     .filter(file => ! file.startsWith("PaxHeader"));
 
+  // Sometimes windows fs does not catch up in time which makes readdir
+  // to return a wrong list of folders / files so we wait until
+  // the file system is in sync
+  if (topLevelOfArchive.length > 1 && process.platform === "win32") {
+	  waitFSToSync(tempDir);
+	  topLevelOfArchive = files.readdir(tempDir);
+  }
+
   if (topLevelOfArchive.length !== 1) {
     throw new Error(
       "Extracted archive '" + tempDir + "' should only contain one entry");
@@ -736,6 +744,32 @@ files.extractTarGz = function (buffer, destPath, options) {
     console.log("Finished extracting in", new Date - startTime, "ms");
   }
 };
+
+function waitFSToSync(dir) {
+	let waitToSync = true;
+
+	while (waitToSync) {
+		try {
+			// After a successful unlink operation, sometimes windows fs takes longer than expected to sync
+			// so readdir would still return already deleted files
+			const filesInTempDirArray = files.readdir(dir);
+
+			// If stat doesn't throw a known error, it means that any queued operations have already been performed
+			// and committed, so the list returned by readdir is the actual file list of the directory
+			filesInTempDirArray.forEach(file => files.stat(files.pathJoin(dir, file)));
+
+			waitToSync = false;
+		} catch (e) {
+			// EPERM is thrown eg after a successful unlink operation and readdir still returns it
+			// ENOENT is thrown if the file deletion is commited by fs between readdir and stat
+			if (e.code === 'EPERM' || e.code === 'ENOENT') {
+				waitToSync = true;
+			} else {
+				throw e;
+			}
+		}
+	}
+}
 
 function ensureDirectoryEmpty(dir) {
   files.readdir(dir).forEach(file => {


tools/fs/files.js
```diff
@@ -737,6 +745,32 @@ files.extractTarGz = function (buffer, destPath, options) {
...
+			// EPERM is thrown eg after a successful unlink operation and readdir still returns it
+			// ENOENT is thrown if the file deletion is commited by fs between readdir and stat
+			if (e.code === 'EPERM' || e.code === 'ENOENT') {
+				waitToSync = true;
+			} else {
+				throw e;
+			}
+		}
+	}
+}
```
benjamn: Please use two spaces per tab (actual spaces, not `\t` characters) for indentation.

tools/fs/files.js
```diff
@@ -737,6 +745,32 @@ files.extractTarGz = function (buffer, destPath, options) {
   }
 };
 
+function waitFSToSync(dir) {
+	let waitToSync = true;
+
+	while (waitToSync) {
```
abernix: This should have a sleep cycle and a maximum spin time—it cannot run unbounded like this, possibly forever if the file system fails to cooperate.  As written right now this could run an enormous number of file operations!

tools/fs/files.js
```diff
@@ -723,6 +723,14 @@ files.extractTarGz = function (buffer, destPath, options) {
     // PaxHeader directory.
     .filter(file => ! file.startsWith("PaxHeader"));
 
+  // Sometimes windows fs does not catch up in time which makes readdir
+  // to return a wrong list of folders / files so we wait until
+  // the file system is in sync
+  if (topLevelOfArchive.length > 1 && process.platform === "win32") {
```
abernix: Can you shed some light on what files are present in the directory? In other words, what are the additional files above the `1` (file) that we actually want?

tools/fs/files.js
```diff
@@ -737,6 +745,32 @@ files.extractTarGz = function (buffer, destPath, options) {
...
+	let waitToSync = true;
+
+	while (waitToSync) {
+		try {
+			// After a successful unlink operation, sometimes windows fs takes longer than expected to sync
+			// so readdir would still return already deleted files
+			const filesInTempDirArray = files.readdir(dir);
+
+			// If stat doesn't throw a known error, it means that any queued operations have already been performed
+			// and committed, so the list returned by readdir is the actual file list of the directory
```
abernix: Is there another (entirely different) way to detect this that we could consider?  Possibly a short-lived watch on the directory?  Something else?

tools/fs/files.js
```diff
@@ -737,6 +745,32 @@ files.extractTarGz = function (buffer, destPath, options) {
...
+			filesInTempDirArray.forEach(file => files.stat(files.pathJoin(dir, file)));
+
+			waitToSync = false;
+		} catch (e) {
+			// EPERM is thrown eg after a successful unlink operation and readdir still returns it
```
abernix: Please keep line length under 80.

tools/fs/files.js
```diff
@@ -737,6 +745,32 @@ files.extractTarGz = function (buffer, destPath, options) {
   }
 };
 
+function waitFSToSync(dir) {
```
abernix: I think another name for this function is in order too, since the `Sync` suffix has special meaning on many functions.  We've used "quiescence" in other circumstances where we're waiting for tranquility, though I'm not convinced that's a great word either. :smile:

abernix: Depending on the final solution here, any aggressive file operation should also take place within a `Fiber`.

tools/fs/files.js
dossorio: Thanks for all your comments @abernix !! I'll address them asap!

The files in question are described in the description of the pr, hopefully it makes sense! 

The thing is, at this point, we only expect the folder where we extracted the contents of the package to exist, but due to fs being out of sync, `readdir` returns the temporary files created in the extraction process, although they were successfully removed by `unlink`.

So let's say that:
```javascript
const tempDir = 'C:\Users\%userprofile%\AppData\Local\Temp\.tmp12zjjrp.xktr'
fs.readdir(tempDir); // Returns wrong list when fs is not in sync
```
So we expect `readdir` to return for example `['coffeescript-1.11.1_4']`, but instead it returns `['coffeescript-1.11.1_4', 'out.tar.gz', 'out.tar']` although the last two were already removed at that point. 
So the function that waits for fs to sync, performs `stat` operations over the elements of `readdir`'s output, which throw an exception that we catch when the file can't be accessed, in this case because it was already removed and wouldn't carry on with the execution until `readdir` returns a list of files / folders that are present on disk. In this case we expect `['coffeescript-1.11.1_4']`, but it could return any list of files, being the latter scenario caught later in https://github.com/meteor/meteor/blob/devel/tools/fs/files.js#L728 but now for good reasons.

Do you envisage any flaws in this process ?

abernix: > The files in question are described in the description of the pr, hopefully it makes sense!

You mentioned:

> Files created in the 7z.exe extraction process of a package

...but it sounds like you're referring to just `out.tar.gz` and `out.tar`.  The `out.tar` file is the only file expected to be _created_ in that extraction process; the `tar.gz` is not created in the extraction, it's downloaded before the extraction begins.  Maybe I'm missing something?  If they're ending up in the same directory and that's causing the issue, perhaps there's another option where we extract into a separate directory from where we download?  Does that solve anything?

> Do you envisage any flaws in this process ?

Yes.  I understand how this is important and hope to get a solution here, but I think the other comments I left (on other lines of code) outline my concerns with a solution of the type you've offered here.  This is an aggressive file-system operation (running an unbounded number of additional file system read operations to wait for sync) and we should consider any other option before moving in this direction.  For example, the Node issues you linked to seem to suggest the option of using method which wait for sync from the file-system via [`fs.fsyncSync`](https://nodejs.org/dist/latest-v8.x/docs/api/fs.html#fs_fs_fsyncsync_fd)?  Have you looked at that?

abernix:
Thanks for taking the time to work on this!  This seems to have fallen idle, so I'm going to close it for the time being.

You've absolutely identified the problem here and we appreciate you taking the time to investigate it. Thanks your work, it's entirely feasible to build upon this in the future with a better understanding of the problem.  Right now, I'm not convinced (and I believe I've provided reasoning for that in the comments above) that this is a harmless solution, or as elegant as it needs to be for an important filesystem operation.  We don't currently have the resources to pick this up but we hope that you or someone will be able to iterate on this in the future.

antoninadert:
I have this issue very often

antoninadert:
Please reopen it it can be very big hassle on Windows. Every time I run `meteor` I get this
`   at Object.files.extractTarGz (C:\tools\fs\files.js:733:11)` error
 

Please take this into account. Some of us are on Windows...
Please reopen

antoninadert:
I also have mcafee indeed.
I noticed powershell doesnt have the issue

DavidRHulme:
Yeh after disabling my mccafee scanning, it has been resolved

PR #9345, opened 2017-11-10
[WIP] Removing underscore from check

stubailo:

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -368,9 +368,9 @@ var testSubtree = function (value, pattern) {
   });
 
   //XXX: replace with underscore's _.allKeys if Meteor updates underscore to 1.8+ (or lodash)
-  var allKeys = function(obj){
+  function allKeys(obj) {
     var keys = [];
-    if (_.isObject(obj)){
+    if (obj) {
       for (var key in obj) keys.push(key);
     }
     return keys;
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -439,7 +439,7 @@ class ArgumentChecker {
     // or check([foo, bar], [String]) to count... but only if value wasn't
     // itself an argument.
     if (Array.isArray(value) || isArguments(value)) {
-      _.each(value, self._checkingOneValue.bind(self));
+      Array.prototype.forEach.call(value, self._checkingOneValue.bind(self));
     }
   }
 
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -121,7 +121,7 @@ var Maybe = function (pattern) {
 };
 
 var OneOf = function (choices) {
-  if (_.isEmpty(choices))
+  if (!choices || choices.length === 0)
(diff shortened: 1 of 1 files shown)


packages/check/match.js
```diff
@@ -488,3 +488,6 @@ var _prependPath = function (key, base) {
   return key + base;
 };
 
+function isArguments(item) {
+  return Object.prototype.toString.call(item) === '[object Arguments]';
```
benjamn: This is a bit more complicated in underscore/lodash:
```js
function baseIsArguments(value) {
  return isObjectLike(value) && baseGetTag(value) == argsTag;
}

var isArguments = baseIsArguments(function() {
  return arguments;
}()) ? baseIsArguments : function(value) {
  return isObjectLike(value) && hasOwnProperty.call(value, 'callee') &&
    !propertyIsEnumerable.call(value, 'callee');
};
```

packages/check/match.js
```diff
@@ -379,14 +379,14 @@ var testSubtree = function (value, pattern) {
   for (var keys = allKeys(value), i = 0, length = keys.length; i < length; i++) {
     var key = keys[i];
     var subValue = value[key];
-    if (_.has(requiredPatterns, key)) {
+    if (Object.prototype.hasOwnProperty.call(requiredPatterns, key)) {
```
benjamn: I usually save a reference to this method in a `hasOwn` variable at the top of the module.

packages/check/match.js
```diff
@@ -368,9 +368,9 @@ var testSubtree = function (value, pattern) {
...
 
   //XXX: replace with underscore's _.allKeys if Meteor updates underscore to 1.8+ (or lodash)
-  var allKeys = function(obj){
+  function allKeys(obj) {
     var keys = [];
-    if (_.isObject(obj)){
+    if (obj) {
       for (var key in obj) keys.push(key);
     }
     return keys;
```
benjamn: So… this is an interesting case, because the loop over the keys below could just be a `for`-`in` loop, and you wouldn't need this `allKeys` function at all:
```js
for (var key in value) {
  var subValue = value[key];
  ...
```
This almost works, but it fails if `value` is not something object-like.

To fix that problem, you can just coerce it with the `Object` constructor:
```js
for (var key in Object(value)) {
  var subValue = value[key];
  ...
```
This works because `Object(value) === value` if it's object-like, but `Object(value)` evaluates to a new empty object if it's anything else. In those cases the loop body will never execute, so accessing `value[key]` there should be safe.

In fact, underscore and lodash implement `_.isObject` with `return value === Object(value)`, demonstrating the safety of this approach.

benjamn:
At first glance I don't know what the test failures are about, but there's enough of them to be worth debugging.

stubailo:
This was going to be part of a larger thing of removing underscore from all packages in `meteor-base`, but I ran out of time!

PR #9346, opened 2017-11-10
Print nicer error message for missing Meteor package

### LINKED ISSUE #8313: Meteor could print nicer message for missing Meteor package
Right now, you get a message like:

```
Error: Can't find npm module 'meteor/xolvio:cleaner'.
```

Seems like it would be pretty simple to add a branch to check for starting with `meteor/` and print a message to `meteor add X` instead.

The issue is that the message is actually incorrect, since the problem doesn't have anything to do with `npm`.

mpowaga:
Fixes #8313 (see feature request [here](https://github.com/meteor/meteor-feature-requests/issues/46)).

Example error message:

```
Error: Cannot find module "meteor/xolvio:cleaner". Try `meteor add xolvio:cleaner` as it looks like you tried to import it without adding to the project.
```

It's the same message as defined [here](https://github.com/meteor/meteor/blob/devel/tools/static-assets/server/npm-require.js#L152) though maybe it should be changed to something else.

### DIFF
--- a/tools/static-assets/server/boot.js
+++ b/tools/static-assets/server/boot.js
@@ -285,7 +285,19 @@ var loadServerBundles = Profile("Load server bundles", function () {
           }
         }
 
-        var resolved = require.resolve(name);
+        try {
+          var resolved = require.resolve(name);
+        } catch (err) {
+          const nameParts = name.split("/");
+          // If it looks like `meteor/xxx`, the user may forgot to add the
+          // package before importing it.
+          if (nameParts.length === 2 && nameParts[0] === "meteor") {
+            const meteorAddTip = ". Try `meteor add " + nameParts[1] + "` " +
+              "as it looks like you tried to import it without adding " +
+              "to the project.";
+            throw new Error("Cannot find module " + JSON.stringify(name) + meteorAddTip);
+          }
+        }
         if (resolved === name && ! path.isAbsolute(resolved)) {
           // If require.resolve(id) === id and id is not an absolute
           // identifier, it must be a built-in module like fs or http.


tools/static-assets/server/boot.js
```diff
@@ -285,7 +285,19 @@ var loadServerBundles = Profile("Load server bundles", function () {
...
         }
 
-        var resolved = require.resolve(name);
+        try {
+          var resolved = require.resolve(name);
+        } catch (err) {
+          const nameParts = name.split("/");
+          // If it looks like `meteor/xxx`, the user may forgot to add the
+          // package before importing it.
+          if (nameParts.length === 2 && nameParts[0] === "meteor") {
```
hwillson: This is just personal preference (totally optional - just thought I'd throw it in); when working with `split` I like using array destructuring to give the split up array parts nice meaningful names (makes the code more readable, easier to maintain, etc.). So you could consider adjusting your `split` approach to something like:

```js
...
} catch (error) {
  if (name && name.startsWith('meteor/')) {
    const [meteorPrefix, packageName] = name.split('/');
    throw new Error(
      `Cannot find package "${packageName}". ` +
      `Try "meteor add ${packageName}".`
    );
  } else {
    throw error;
  }
}
...
```

tools/static-assets/server/boot.js
```diff
@@ -285,7 +285,19 @@ var loadServerBundles = Profile("Load server bundles", function () {
           }
         }
 
-        var resolved = require.resolve(name);
+        try {
+          var resolved = require.resolve(name);
+        } catch (err) {
+          const nameParts = name.split("/");
+          // If it looks like `meteor/xxx`, the user may forgot to add the
```
hwillson: I think we can drop this comment; this code pretty much speaks for itself (especially since the error message outlines the case we're handling here).

tools/static-assets/server/boot.js
```diff
@@ -285,7 +285,19 @@ var loadServerBundles = Profile("Load server bundles", function () {
           }
         }
 
-        var resolved = require.resolve(name);
+        try {
+          var resolved = require.resolve(name);
+        } catch (err) {
+          const nameParts = name.split("/");
```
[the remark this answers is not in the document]
abernix: It would also be reasonable to add `2` as the second parameter to `split` to short-circuit it!

hwillson:
Thanks for submitting this PR @mpowaga. We're going to close it for now, but please post back when/if you have time to work on it further, and we'll re-open it. Thanks again!

mpowaga:
Ok @hwillson. I'll definitely push changes to this PR once I find some time, so we can reopen when I commit something. Thanks for your review by the way.

mpowaga:
Hi @hwillson. I took different approach as suggested by modifying `meteorInstallOptions.fallback` option. I pushed new [commit](https://github.com/mpowaga/meteor/commit/87bec4abc130a20fc226262cff7001a8d03d190c) to my branch but it didn't seem to show up in this PR. Can you reopen it? I imagine it will automatically fetch latest changes.

hwillson:
@mpowaga Would you mind creating a new PR instead (and adding a reference back to this PR)? Thanks!

PR #9353, opened 2017-11-12
Avoid loading SockJS in browsers known not to need it.

benjamn:
Supersedes #9316, in which @stubailo demonstrated that `SockJS` can be straightforwardly eliminated if native WebSocket support is assumed.

Native WebSocket support is close to universal now (94% of all clients according to https://caniuse.com/#feat=websockets), which means we can remove ~30KB of legacy code from the client bundle that we send to most web browsers.

To achieve this flexibility, the `sockjs-0.3.4.js` script is injected as a standalone `<script>` tag into the `<head>` of the document, but only if the user agent string of the HTTP request fails to convince the `sockjs-shim` package that the browser has native WebSocket support. When Meteor is running in production, a minified version of that script is used instead.

Not only is this loading strategy a bit faster for older browsers, because the `SockJS` script begins loading earlier, and is independent from the normal JS bundle, it also removes the cost of loading SockJS completely for more recent browsers.

Though it's tempting to keep hacking the SockJS script to make it smaller, we should avoid spending too much time on that, since it only benefits a small portion of users whose devices are likely to have poor performance for lots of other reasons. In other words, I think we should keep
[Amdahl's Law](https://en.wikipedia.org/wiki/Amdahl%27s_law) in mind.


packages/sockjs-shim/server.js
```diff
@@ -0,0 +1,29 @@
...
+  "/packages/sockjs-shim/sockjs-" +
+  sockjsVersion +
+  (Meteor.isProduction ? ".min.js" : ".js");
+
+const hasOwn = Object.prototype.hasOwnProperty;
+const minimumMajorVersions = {
+  chrome: 16,
+  firefox: 11,
+  ie: 10,
+  safari: 7,
```
benjamn: This list is probably not complete, and I'm open to adjusting these versions. I took this information from [CanIUse](https://caniuse.com/#feat=websockets).

packages/ddp-client/common/urlHelpers.js
```diff
@@ -12,6 +12,10 @@ function translateUrl(url, newSchemeBase, subPath) {
     newSchemeBase = 'http';
   }
 
+  if (subPath !== "sockjs" && url.startsWith("/")) {
+    url = Meteor.absoluteUrl(url.substr(1));
+  }
```
benjamn: Thanks to @stubailo for [figuring this out](https://github.com/meteor/meteor/pull/9316/files#r149388945).

stubailo: I think a total rewrite of this function would be a great idea sometime soon, was going to be my next step for hack week. It's one of those functions I hate that pretends to be generic but actually only works for a very slim combination of inputs.

packages/ddp-client/package.js
```diff
@@ -32,6 +32,15 @@ Package.onUse((api) => {
 
   api.use('reload', 'client', { weak: true });
 
+  // If the application is using sockjs-shim, then it needs to be loaded
+  // before ddp-client so that it can polyfill global.SockJS. However, we
+  // don't want to force sockjs-shim to be loaded, since ddp-client is
+  // part of an isopacket used by the Meteor command-line tool, and the
+  // webapp package (used by server-render, which is used by sockjs-shim)
+  // is not safe to load as part of an isopacket, because it calls
+  // Fiber.yield during package initialization. Hence the weakness here.
```
stubailo: This comment should probably indicate whether or not someone else needs to depend on `sockjs-shim` for this to work.

(I guess they do, since it's in `meteor-base` now)

packages/sockjs-shim/README.md
```diff
@@ -0,0 +1,9 @@
+# `sockjs-shim`
+[Source code of released version](https://github.com/meteor/meteor/tree/master/packages/cockjs-shim) | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/sockjs-shim)
```
stubailo: I believe there's a typo in the first link here

packages/sockjs-shim/client.js
```diff
@@ -0,0 +1,9 @@
+// When the server.js module does not inject a <script> tag into the
+// <head> of the document, we still need a fake global shim for SockJS.
```
stubailo: Interesting, do you think this is better than some other part of the code checking for websockets/sockjs?

packages/sockjs-shim/README.md
benjamn: oh dear

packages/sockjs-shim/client.js
benjamn: Oh, that's a good point!

packages/sockjs-shim/server.js
benjamn: One significant missing browser: `phantomjs: 2`. This means `meteor test-packages` is still running with SockJS polyfilled, not using native WebSockets. For the foreseeable future, I think we need to run these tests with and without SockJS, to ensure both approaches work.

PR #9359, opened 2017-11-13
Convert Mongo.Collection to class and remove underscore on client.

benjamn:
When we talk about removing `underscore` from core packages, we're primarily concerned about removing it from the client JS bundle. In other words, we only need to remove `underscore` from modules within a given package that are used on the client. The `mongo` package is a good example of a package where this insight matters: since `mongo` contains mostly server-side code, client-only removal of `underscore` should be relatively easy.

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -24,93 +24,134 @@ The default id generation technique is `'STRING'`.
  * @param {Function} options.transform An optional transformation function. Documents will be passed through this function before being returned from `fetch` or `findOne`, and before being passed to callbacks of `observe`, `map`, `forEach`, `allow`, and `deny`. Transforms are *not* applied for the callbacks of `observeChanges` or to cursors returned from publish functions.
  * @param {Boolean} options.defineMutationMethods Set to `false` to skip setting up the mutation methods that enable insert/update/remove from client code. Default `true`.
  */
-Mongo.Collection = function (name, options) {
-  var self = this;
-  if (! (self instanceof Mongo.Collection))
-    throw new Error('use "new" to construct a Mongo.Collection');
+Mongo.Collection = class Collection {
+  constructor(name, options) {
+    if (! (this instanceof Mongo.Collection))
+      throw new Error('use "new" to construct a Mongo.Collection');
 
-  if (!name && (name !== null)) {
-    Meteor._debug("Warning: creating anonymous collection. It will not be " +
-                  "saved or synchronized over the network. (Pass null for " +
-                  "the collection name to turn off this warning.)");
-    name = null;
-  }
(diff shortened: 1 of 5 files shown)


packages/mongo/collection.js
```diff
@@ -451,8 +451,11 @@ Mongo.Collection = class Collection {
       throw new Error("insert requires an argument");
     }
 
-    // Shallow-copy the document and possibly generate an ID
-    doc = { ...doc };
+    // Make a shallow clone of the document, preserving its prototype.
+    doc = Object.create(
+      Object.getPrototypeOf(doc),
+      Object.getOwnPropertyDescriptors(doc)
+    );
```
benjamn: This idiom deserves a signal boost. I could've squashed this commit into the previous one, but I didn't want this bug fix to be buried, because the underlying problem is worth understanding.

packages/mongo/collection.js
```diff
@@ -24,93 +24,134 @@ The default id generation technique is `'STRING'`.
...
-    throw new Error('use "new" to construct a Mongo.Collection');
-
-  if (!name && (name !== null)) {
-    Meteor._debug("Warning: creating anonymous collection. It will not be " +
-                  "saved or synchronized over the network. (Pass null for " +
-                  "the collection name to turn off this warning.)");
-    name = null;
-  }
+Mongo.Collection = class Collection {
+  constructor(name, options) {
```
benjamn: Since a lot of the changes in this file are just reindentation, you may want to [review the diff using `?w=1`](https://github.com/meteor/meteor/pull/9359/files?w=1).

packages/mongo/mongo_livedata_tests.js
```diff
@@ -1507,7 +1507,9 @@ testAsyncMulti('mongo-livedata - document with a custom type, ' + idGeneration,
       Meteor.call('createInsecureCollection', this.collectionName, collectionOptions);
       Meteor.subscribe('c-' + this.collectionName, expect());
     }
-  }, function (test, expect) {
+  },
+
```
abernix: Clever use of formatting to help tell the full story!

packages/mongo/collection.js
abernix: I really enjoy that [`refined-github`](https://github.com/sindresorhus/refined-github/pull/661), a Chrome Extension I picked up thanks to @stubailo, provides the "No Whitespace" button on PR review. Previously, with or without that extension, the `?w=1` was lost when going next/previous during commit-by-commit review but I submitted https://github.com/sindresorhus/refined-github/pull/661 to maintain the `?w=1` on navigation (`n`/`p`) and I've enjoyed it ever since!

PR #9360, opened 2017-11-13
Selectively server-render es5-shim <script> tag, as with sockjs-shim.

benjamn:
Just as we were able to remove the SockJS polyfill library entirely from the client JS bundle for modern browsers (#9353), we can do the same with the `es5-shim` package.

The removes about 28KB from the size of the minified JS bundle (before gzip).

It is no longer necessary for other packages to register weak dependencies on `es5-shim`, because `es5-shim` will be reliably loaded in the `<head>` if installed, which guarantees it will be evaluated before any bundled JS in packages or the application.

In addition to loading `es5-shim` and `es5-sham` using `<script>` tags on the client, we no longer evaluate either script on the server, since Node 8 has solid support for everything that previously needed shimming.


benjamn:
That's a totally valid point @hwillson, and I like the surface area of shared code that you've suggested.

PR #9361, opened 2017-11-14
Remove underscore from check package.

benjamn:
Supersedes #9345.

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -5,6 +5,7 @@
 
 var currentArgumentChecker = new Meteor.EnvironmentVariable;
 var isPlainObject = require("./isPlainObject.js").isPlainObject;
+var hasOwn = Object.prototype.hasOwnProperty;
 
 /**
  * @summary Check that a value matches a [pattern](#matchpatterns).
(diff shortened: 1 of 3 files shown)


packages/check/match.js
```diff
@@ -418,51 +418,54 @@ var testSubtree = function (value, pattern) {
...
-  self.args = _.clone(args);
-  // Since the common case will be to check arguments in order, and we splice
-  // out arguments when we check them, make it so we splice out from the end
-  // rather than the beginning.
-  self.args.reverse();
-  self.description = description;
-};
+class ArgumentChecker {
+  constructor (args, description) {
+    var self = this;
```
abernix: As @stubailo's other work has shown (e.g. 8788b2d793ac1eca752b554985aa72ce5d191162, 45928b2cb2219c73d33eedf091ab8493aa8a6946, 6a0b94340c1783b2cc559ed101c41e806617c835, 88847faf27df2aa3461bf052e26d00e536c0c88b), I believe this would have normally been followed up with a commit which removed `self = this` in lieu of just using `this`.  I've taken the liberty of finishing that with https://github.com/meteor/meteor/pull/9361/commits/8a122b7cc2e8d7bd48338171982603637d73c0c8, which I've pushed to this PR.

PR #9364, opened 2017-11-14
Decompose self-test logic into separate files.

abernix:
The single file which represented the bulk of the `meteor self-test`
functionality had got a bit heavy and it stood to benefit from some
dissemination.

I embarked on this change originally when looking into replacing
PhantomJS with Chrome Headless (and a new `ChromeClient` class) within
self-test.  Unfortunately, I didn't have time to take this the last
step of actually implementing Chrome, but this should hopefully
facilitate that change in the future by providing what I believe to be
better compartmentalization of this logic.

I apologize for the (likely) difficulty of reviewing this commit.  Due
to heavy intertwining of existing code it was hard to arrange these
changes in an easy-to-review manner.  I believe a reviewer will
find that it's mainly copy and pasting into different files and
careful adjusting of those files' (new) module dependencies.


tools/tool-testing/run.js
```diff
@@ -0,0 +1,443 @@
...
+import OutputLog from './output-log.js';
+import { randomPort, timeoutScaleFactor, sleepMs } from '../utils/utils.js';
+import TestFailure from './test-failure.js';
+import { execFileSync } from '../utils/processes.js';
+
+let runningTest = null;
+
+export function setRunningTest(test) {
+  runningTest = test;
+}
```
abernix: I'll call this out ahead of time as I expect this `setRunningTest` method (used by [`Test`'s `onCleanup`](https://github.com/meteor/meteor/blob/abernix/decompose-self-test/tools/tool-testing/selftest.js#L83-L85); registered in the `constructor` below) to be the most subject to scrutiny portion of this re-factor.  It had previously been a global variable within the lone `selftest.js` file but it could stand to be further refined in this re-worked edition of `self-test`.  Ultimately, I just didn't want to completely abandon the work I had done on this PR, even though the work which prompted it had to be put on hold.

tools/tool-testing/clients/browserstack.js
```diff
@@ -0,0 +1,166 @@
...
+} from '../../fs/files.js';
+
+const NPM_DEPENDENCIES = {
+  'browserstack-webdriver': '2.41.1',
+};
+
+// BrowserStackClient
+let browserStackKey = null;
+
+export default class BrowserStackClient extends Client {
```
benjamn: Do we still run any BrowserStack tests? In other words, is this code being exercised by the current test suite that ran for this PR?

tools/tool-testing/selftest.js
```diff
@@ -2121,7 +585,7 @@ export function runTests(options) {
     } catch (e) {
       failure = e;
     } finally {
-      runningTest = null;
+      setRunningTest(null);
       test.cleanup();
     }
```
benjamn: What if you moved this code into a `static` method of the `Run` class, and here wrote something like
```js
const {
  error,
  durationMs,
} = Run.runTest(test, parseStackMarkBottom(() => {
  test.f(options);
});
```
That way you wouldn't have to be careful about calling `setRunningTest` twice, and the code that needs to know the current test can be (more) responsible for managing that information.

tools/tool-testing/clients/browserstack.js
abernix: Good question!  We don't, but some digging has uncovered that we are currently paying for it.  If we think it provides value, I'll look into reviving this portion of the tests.

abernix: fwiw, I can confirm locally (with some futzing around), that it's broken in the same way before and after this PR, but not in a way that's related to these changes.

abernix:
It took me a while to circle around to this and finish up, but the additional commits are ready for review.

There is a notable feature-creep in this PR: The `meteor self-test` command now accepts a `--retries`. While the default of "3" (total) tries is nice in CircleCI to avoid erroneous errors, it's quite frustrating to have to wait for the test to fail three times when working actively on `self-test` code, or even when writing/editing tests.  I frequently find myself changing the (hardcoded) `tries = 3` to `tries = 1` in `Run.runTest` in order to speed up development.

This new `retries` (yes, re-named from `tries`) makes it possible to pass `--retries 0` on the command line.  I might suggest we change the default for `--retries` to `0` and just pass the value in CI.  To prepare for that, I've already made the change to pass `--retries 2` in the CircleCI and AppVeyor configurations.

The BrowserStack functionality via the [`BrowserStackLocal` tunnel](https://www.browserstack.com/local-testing) doesn't seem to be working within CircleCI during the course of an automatic run (e.g. initiated from a `git push`, PR open, etc.), though it does work if you retry the failed test with SSH enabled on the build and works locally.  The resulting error (as [seen here](https://circleci.com/gh/meteor/meteor/12366)) is:

    UnknownError: [browserstack.local] is set to true but local testing through BrowserStack is not connected.

I don't have an explanation for this at the moment, though I've successfully ran the tests locally (from my Mac) and the same exact build works again if just restarted with the "Rebuild with SSH" option on CircleCI.  The `BrowserStackLocal` daemon is running in either case.  I'm afraid that the time needed to further investigate this might not be worth it at the moment, but I hope to revisit.  If anyone is willing to take a look, feel free.   It's a shame since BrowserStack has graciously provided the BrowserStack "Automate" package to Meteor free of charge for this purpose and is [quite cool](https://www.browserstack.com/automate/builds/0392da976ad09954e1f3243689ee057ea6256d97/sessions/096b9ca9ca495c093e57cef4fb42fbe93178974f?auth_token=861f968e7429350253d1a25c19929f9f0d7d05ce3a8fedd6172964e575a0abcf&build_token=S2FjOVpYTlk1eHVxdkdrR24ra0JXaXBDaVA3WjJGejdkbThaWGRnWVJvWT0tLTlQMEdYM2NJbDJOYUd3RTc3RVVGQ2c9PQ%3D%3D--9e2143718a0c216274cdacef7cd5a3d12950bcb8)!  Luckily, it's still easy to run locally with the `--browserstack` flag to `meteor self-test`.  Of course, a valid `BROWSERSTACK_ACCESS_KEY` must be set in the environment for this to work.

hwillson:
> I frequently find myself changing the (hardcoded) tries = 3 to tries = 1 in Run.runTest in order to speed up development.

I've been doing this as well! This is great! 🎉

PR #9366, opened 2017-11-14
Stop pre-installing phantomjs-prebuilt and browserstack-webdriver.

abernix:
While this was necessary in previous versions of Meteor, as of 1.6 the
`self-test` tooling will automatically install the dependencies it needs
at runtime. (Thanks to https://github.com/meteor/meteor/pull/8981)

Not only does this allow the installation to be avoided when tests using
those facilities aren't invoked, it also allows us to reliably install
the same version of these npms, rather than falling back to the
`latest` npm tag on each CI run.

Note that this doesn't change the behavior of the [`test-packages` `run.sh`](https://github.com/meteor/meteor/blob/abernix/stop-installing-self-test-npms/packages/test-in-console/run.sh#L8)
script (used by Travis) which still installs the `latest` version of these npms.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,11 +126,6 @@ jobs:
       - run:
           name: Clear npm cache
           command: ./meteor npm cache clear --force
-      # Since PhantomJS has been removed from dev_bundle/lib/node_modules
-      # (#6905), but self-test still needs it, install it now.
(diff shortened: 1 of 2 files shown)


benjamn:
The passing of the tests is a strong sign these changes are OK!

Issue #9138, opened 2017-09-27
Meteor 1.5.2.1: uglify cannot uglify emojione


acemtp:
Hello,

Just move from 1.5.1 to 1.5.2.1 and now, deployed version failed because of new version uglify that failed to uglify emojione:

```
4049afb95685ab66e1e11da2cff24e771534ea2e.js?meteor_js_resource=true:181 Uncaught SyntaxError: Invalid regular expression: /<object[^>]*>.*?</object>|<span[^>]*>.*?</span>|<(?:object|embed|svg|img|div|span|p|a)[^>]*>|(?:🏳)️?‍?(?:🌈)|(?:👁)️?‍?(?:🗨)️?|[#-9]️?⃣|(?:(?:🏴)(?:�[�-�]){1,6})|(?:�[�-�]){2}|(?:(?:�[��]))️?(?:�[�-�])?‍?(?:[⚕⚖✈]|�[�-�]|�[������])|(?:�[��]|�[�-�])(?:�[�-�])?‍?[♀♂⚕⚖✈]?️?|(?:(?:❤|�[�-��])[‍️]{0,2}){1,3}(?:❤|�[�-��])|(?:(?:❤|�[�-��])️?){2,4}|(?:�[����-���-�]|�[��]|�[�-�]|�[�-�]|�[�-�]|⛹|👯)️?(?:�[�-�])?‍?[♀♂]?️?|(?:[☝⛹✊-✍]|�[�-�]|�[�-��-��-��-�]|�[�-�])️?(?:�[�-�])|(?:[↔-↙↩-↪]️?|[#*]|[〰〽]️?|(?:�[�-�]|🆎|�[�-�])️?|Ⓜ️?|[㊗㊙]️?|(?:�[�-�]|🈚|🈯|�[�-�]|�[�-�])️?|[‼⁉]️?|[▪-▫▶◀◻-◾]️?|[©®]️?|[™ℹ]️?|🀄️?|[⬅-⬇⬛-⬜⭐⭕]️?|[⌚-⌛⌨⏏⏩-⏳⏸-⏺]️?|🃏|[⤴⤵]️?)|[✀-➿]️?|[�-�][�-�]️?|[☀-⛿]️?|[0-9]️/: Range out of order in character class
    at RegExp (<anonymous>)
    at f.emojione (4049afb95685ab66e1e11da2cff24e771534ea2e.js?meteor_js_resource=true:181)
[...]

PR #9368, opened 2017-11-14
Update minifier-js uglify-es dependency to latest version

hwillson:
There have been quite a few `uglify-es` bug fixes made since 3.0.28, some of which fix outstanding Meteor issues. This PR updates `minifier-js` to use the latest version (3.1.9). Unfortunately `uglify-es` doesn't maintain a changelog so it's a bit difficult to run through all of the changes, but I've stepped through the commit history and haven't found anything that jumps out as being problematic. I've also minified a few apps using 3.1.9, and everything looks good.

Fixes https://github.com/meteor/meteor/issues/9138 (and likely others).

### DIFF
--- a/packages/minifier-js/.npm/package/npm-shrinkwrap.json
+++ b/packages/minifier-js/.npm/package/npm-shrinkwrap.json
@@ -7,14 +7,14 @@
       "integrity": "sha512-b0553uYA5YAEGgyYIGYROzKQ7X5RAqedkfjiZxwi0kL1g3bOaBNNZfYkzt/CL0umgD5wc9Jec2FbB98CjkMRvQ=="
     },
     "source-map": {
-      "version": "0.5.7",
-      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
-      "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w="
+      "version": "0.6.1",
+      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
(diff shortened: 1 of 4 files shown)


packages/minifier-js/.npm/package/npm-shrinkwrap.json
```diff
@@ -7,14 +7,14 @@
       "integrity": "sha512-b0553uYA5YAEGgyYIGYROzKQ7X5RAqedkfjiZxwi0kL1g3bOaBNNZfYkzt/CL0umgD5wc9Jec2FbB98CjkMRvQ=="
     },
     "source-map": {
-      "version": "0.5.7",
-      "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.5.7.tgz",
-      "integrity": "sha1-igOdLRAh0i0eoUyA2OpGi6LvP8w="
+      "version": "0.6.1",
```
benjamn: We haven't taken this minor version update in other places, but I think the usage of `source-map` by `uglify-es` is fairly self-contained.

hwillson:
> There were a lot of commits though.

Arg, tell me about it - I wish they had a changelog of some kind! Just for reference, I believe this is the commit that fixed the issue: https://github.com/alexlamsl/UglifyJS2/commit/98b850580b5096b07f77f531b6a0d12f25c86deb

Issue https://github.com/mishoo/UglifyJS2/issues/2242 goes into the details. Thanks!

abernix:
That makes sense.  Thanks for tracking it down and linking it here! I never would have found it based on _that_ commit message. :wink:

PR #9371, opened 2017-11-15
Extract ClientStream into its own package.

benjamn:
Right now, client-side access to Meteor's persistent WebSocket connection is controlled by the `ddp-client` package, which unfortunately includes some large modules like [`livedata_connection.js`](https://github.com/meteor/meteor/blob/devel/packages/ddp-client/common/livedata_connection.js) that help implement DDP and Livedata, but do not seem necessary for simple socket-based communication with the server.

This PR is the beginning of an effort to make the WebSocket connection usable without also depending on the `ddp-client` package.

For example, eventually the `dynamic-import` package might be able to depend directly on `socket-stream-client` to fetch dynamic modules via WebSocket, without depending on all of `ddp-client` as well. This would allow dynamic `import()` to be used earlier in the Meteor startup process, in more fundamental ways, regardless of whether `ddp-client` is used by the application.


packages/ddp-client/common/getClientStreamClass.js
```diff
@@ -1,18 +0,0 @@
...
-export default function getClientStreamClass() {
-  // The static analyzer of the bundler specifically looks
-  // for static calls to 'require', so it won't treat the
-  // below calls as a request to include that module.
-  //
-  // That means stream_client_nodejs won't be included on
-  // the client, as desired.
-  const modulePath = Meteor.isClient
-    ? '../client/stream_client_sockjs'
-    : '../server/stream_client_nodejs';
```
benjamn: @stubailo I was able to get rid of this trickery by using different client/server entry points in a separate package.

theodorDiaconu:
> For example, eventually the dynamic-import package might be able to depend directly on socket-stream-client to fetch dynamic modules via WebSocket,

@benjamn if you implement this, leave the door opened to allow using Meteor without a websocket connection so we can fetch the modules via an http request.

packages/socket-stream-client/browser.js
```diff
@@ -114,8 +113,8 @@ export default class ClientStream extends StreamClientCommon {
   }
 
   _heartbeat_timeout() {
-    Meteor._debug('Connection timeout. No sockjs heartbeat received.');
-    this._lostConnection(new DDP.ConnectionError('Heartbeat timed out'));
```
abernix: It's a shame this custom error type used a special `errorType` property.

It seems like we mainly check for this type of error internally, notably [`tools/meteor-services/auth-client.js#L33`](https://github.com/meteor/meteor/blob/devel/tools/meteor-services/auth-client.js#L33) and a couple places in [`tools/packaging/catalog/catalog.js`](https://github.com/meteor/meteor/blob/devel/tools/packaging/catalog/catalog.js#L126), but this presumably bubbles into those under some circumstances? Perhaps those should be updated?

Other than that, I did some GitHub [code searching](https://github.com/search?p=1&q=%22DDP.ConnectionError%22+extension%3Ajs+-filename%3Astream_client%2A+-filename%3Aservice-connection+-path%3Apackages%2Fddp-client+-path%3Aprograms%2F+-path%3Atests%2F+-filename%3Addp-client.js+-filename%3Aspec+-filename%3Astubs+-filename%3Aauth-client.js+-filename%3Acatalog.js&type=Code&utf8=%E2%9C%93) to try to hone in on any potential problems, but didn't come across anything concerning or that might be a breaking change, though I suppose developers could be looking for this type of error in their own private code?

packages/socket-stream-client/browser.js
```diff
@@ -201,7 +200,7 @@ export default class ClientStream extends StreamClientCommon {
 
     if (this.connectionTimer) clearTimeout(this.connectionTimer);
     this.connectionTimer = setTimeout(() => {
-      this._lostConnection(new DDP.ConnectionError('DDP connection timed out'));
+      this._lostConnection(new Error("DDP connection timed out"));
```
abernix: Should we start re-branding this error without `DDP` in it?

benjamn:
@theodorDiaconu That door will always stay open (just need to swap out this [`fetchMissing`](https://github.com/meteor/meteor/blob/55ccc0f20372596defe8e22d1f17efe9a5d28425/packages/dynamic-import/client.js#L113) function for one that uses `XMLHttpRequest`s instead of `Meteor.call`), though every time I've tried that, it's been significantly slower.

theodorDiaconu:
@benjamn I also noticed that http request will be slower than ddp, but I've recently published a POC package that allows conditionally starting websocket only when you need it. 

https://github.com/cult-of-coders/fusion it's hackish but it does the job.

I would very much like this to be in the core, or allow a meteor package to hack it somehow... The reason for this is so Meteor can also address the market of apps that don't require a stateful relation with the client.

Issue #9379, opened 2017-11-16
[1.6] Cannot build on iOS due to cordova-plugin-console


derwaldgeist:
After I updated my app from 1.5 to 1.6, it won't build on iOS anymore. The linker throws an error:

```
duplicate symbol _OBJC_CLASS_$_CDVLogger in:
    /Users/waldgeist/Library/Developer/Xcode/DerivedData/travelguzz-gchredzybixpkmclpnbymgmjhkrl/Build/Intermediates.noindex/travelguzz.build/Debug-iphoneos/travelguzz.build/Objects-normal/arm64/CDVLogger.o
    /Users/waldgeist/Library/Developer/Xcode/DerivedData/travelguzz-gchredzybixpkmclpnbymgmjhkrl/Build/Products/Debug-iphoneos/libCordova.a(CDVLogger.o)
duplicate symbol _OBJC_METACLASS_$_CDVLogger in:
    /Users/waldgeist/Library/Developer/Xcode/DerivedData/travelguzz-gchredzybixpkmclpnbymgmjhkrl/Build/Intermediates.noindex/travelguzz.build/Debug-iphoneos/travelguzz.build/Objects-normal/arm64/CDVLogger.o
    /Users/waldgeist/Library/Developer/Xcode/DerivedData/travelguzz-gchredzybixpkmclpnbymgmjhkrl/Build/Products/Debug-iphoneos/libCordova.a(CDVLogger.o)
ld: 2 duplicate symbols for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I found a thread in the Ionic framework that boiled this down to `cordova-plugin-console` which has been replaced by Cordova-internal code and thus resulting in duplicate object literals:

https://github.com/ionic-team/ionic/issues/12849#issuecomment-337349446

And in fact, a closer inspection of the generated code shows that `CDVLogger.m` and `CDVLogger.h` appear twice, under `CodovaLib` and under `Plugins`.

Unfortunately, Meteor still hard-depends on this console plugin, so I can't get rid of it:

https://github.com/meteor/meteor/blob/073f2410a685f1de87f70dff6a097b1eea2712b1/packages/logging/package.js#L15
https://github.com/meteor/meteor/blob/235d00e2c7b2d4ad55b92f0c350ac7edfa4b237b/tools/cordova/project.js#L56

skirunman:
[...]

Issue #9300, opened 2017-11-04
Meteor 1.6 not building in XCode due to cordova-plugin-console


tdotholla:
i'm running Meteor 1.6 and trying to build to xCode; 
seeing an error exactly as [this StackOverflow Post](https://stackoverflow.com/questions/46633797/build-cordova-meteor-ionic-for-ios-duplicate-symbol-objc-class-cdvlogger): 

after digging it's caused due to Cordova 4.5+ incorporating `cordova-plugin-console`; so building with xCode is causing duplicate calls to certain classes. 

we can resolve by removing the classes after the build; but every time i build it gets placed back. 

How do we remove the dependency from meteor permanently? 

i've already tried:
- `meteor remove cordova:cordova-plugin-console` to no avail; says no such dependency exists in my project.
- `meteor remove-platform ios` and re-added it; to no avail. this is a piece of the log when i build again: 

```%% Install complete for cordova-plugin-meteor-webapp on ios.
%% Finding scripts for "after_plugin_install" hook from plugin cordova-plugin-meteor-webapp on ios platform only.
%% Executing script found in plugin cordova-plugin-meteor-webapp for hook "after_plugin_install": plugins/cordova-plugin-meteor-webapp/scripts/iosAddBridgingHeader.js
%% Resolving module name for fs => fs
%% Resolving module name for path => path
%% Resolving module name for cordova-lib/src/cordova/util.js => ../cordova/util.js
%% Resolving module name for cordova-common => cordova-common
%% No scripts found for hook "after_plugin_add".
Adding plugin cordova-plugin-console@1.0.7 to Cordova project
%% No scripts found for hook "before_plugin_add".
%% Calling plugman.fetch on plugin "cordova-plugin-console@1.0.7"
%% Fetching plugin "cordova-plugin-console@1.0.7" via npm
npm WARN invalid config loglevel="notice"    [========                 ] 30% 163.9s
npm WARN deprecated cordova-plugin-console@1.0.7: This plugin has been deprecated since it is now included in the latest versions of cordova-ios
%% Copying plugin "/Users/tdotholla/.cordova/lib/npm_cache/cordova-plugin-console/1.0.7/package" => "/Users/tdotholla/Sites/thisapp/app//.meteor/local/cordova-build/plugins/cordova-plugin-console"
%% Calling plugman.install on plugin "/Users/tdotholla/Sites/thisapp/app/.meteor/local/cordova-build/plugins/cordova-plugin-console" for platform "ios
[...]

PR #9380, opened 2017-11-16
Removed cordova-plugin-console dependency

renanccastro:
As stated on [cordova-plugin-console](https://github.com/apache/cordova-plugin-console):

> This plugin is no longer being worked on as the functionality provided by this plugin is now included in cordova-ios 4.5.0 or greater, and support is already built in to cordova-windows > 5.0.0. You should remove this plugin from your applications.
> 
> Please file issues for this plugin against their respective platforms (cordova-ios, cordova-windows).
> 

Also on this issue: #9379

### DIFF
--- a/packages/logging/package.js
+++ b/packages/logging/package.js
@@ -11,10 +11,6 @@ Npm.strip({
   "es5-ext": ["test/"]
 });
 
-Cordova.depends({
-  'cordova-plugin-console': '1.0.7'
-});
-
 Package.onUse(function (api) {
   api.export('Log');
   api.use(['underscore', 'ejson', 'modules']);


skirunman:
@renanccastro I'm curious what happens if you manually install `cordova-plugin-console@1.1.0` in your app/.meteor/cordova-plugins and re-build your app. See #9300

This https://github.com/meteor/meteor/pull/9213/files updates our pinned dependency to `cordova-plugin-console@1.1.0` which I believe should not install the conflicting code when building with cordova-ios 4.5.0 or greater.

renanccastro:
@skirunman 
Tried that. Spawns an error and doesn't install the plugin(cordova-plugin-console). Duplicated CDVLogger no longer occurs.

That's a workaround, but I think meteor should remove it and get rid of this warning/error.

> npm WARN deprecated cordova-plugin-console@1.1.0: This plugin has been deprecated since it is now included in the latest versions of cordova-ios
> %% Plugin doesn't support this project's cordova-ios version. cordova-ios: 4.5.1, failed version requirement: <4.5.0
> %% Skipping 'cordova-plugin-console' for ios

skirunman:
See my comments in #9379. The warning may be a good thing for a least one release of the product so people manually remove any dependencies they may have in their own code, but I do agree that we can remove this dependency long term. 

In your PR you may also want to remove the version pinning of this plugin in `tools/cordova/project.js`, but I guess this is not a requirement.

benjamn:
Let's ship @skirunman's version bump in Meteor 1.6.1, and then consider this PR for 1.6.2.

Closing this PR until then, though I've put it in the 1.6.2 milestone to make sure we don't forget it.

Thanks @renanccastro and @skirunman!

PR #9384, opened 2017-11-17
Support fetching dynamic modules via HTTP POST requests.

benjamn:
Ever since Meteor 1.5 first shipped, dynamic modules have been fetched over a WebSocket, which is appealing because sockets have very little latency and metadata overhead per round-trip.

However, using a WebSocket requires first establishing a socket connection to the server, which takes time and may require a WebSocket polyfill.

An even more subtle problem is that we cannot use dynamic imports in any of the code that implements the `ddp-client` package, as long as the `dynamic-import` package depends on `ddp-client`.

By switching from WebSockets to HTTP POST requests, this commit radically reduces the dependencies of the `dynamic-import` package, while preserving (or even exceeding) the benefits of socket-based dynamic module fetching:

* The client makes a single HTTP POST request for the exact set of dynamic modules that it needs, which is much cheaper/faster than making several/many HTTP requests in parallel, with or without HTTP/2.

* Because the client uses a permanent cache (IndexedDB) to avoid requesting any modules it has ever previously received, the lack of HTTP caching for POST requests is not a problem.

* Because the HTTP response contains all the requested dynamic modules in a single JSON payload, gzip compression works across modules, which produces a smaller total response size than if each individual module was compressed by itself.

* Although HTTP requests have more latency than WebSocket messages, the ability to make HTTP requests begins much sooner than a WebSocket connection can be established, which more than makes up for the latency disadvantage.

* HTTP requests are a little easier to inspect and debug in the dev tools than WebSocket frames.

* The new `/__dynamicImport` HTTP endpoint is a raw Connect/Express-style endpoint, so it bypasses the Meteor method-calling system altogether, which eliminates some additional overhead.

* Fetching dynamic modules no longer competes with other WebSocket traffic such as DDP and Livedata.

* Although the implementation of the `/__dynamicImport` endpoint is a bit too complicated to allow serving dynamic modules from a CDN, fetching modules individually from a CDN remains a possibility for future experimentation. In other words, how modules are fetched is still just an implementation detail of the [`meteorInstall.fetch`](https://github.com/meteor/meteor/blob/305abbefb8e357a8512f4b062078093318e8a3ad/packages/dynamic-import/client.js#L19) callback.

* As with the WebSocket implementation, the module server is totally stateless, so it should be easy to scale up if necessary.

I wish I had appreciated the advantages of an HTTP-based implementation sooner, but I think the transition will be pretty seamless, since the new implementation is completely backwards compatible with the old.

### DIFF
--- a/packages/dynamic-import/server.js
+++ b/packages/dynamic-import/server.js
@@ -96,29 +96,36 @@ function readTree(tree, platform) {
   const pathParts = [];
 
   function walk(node) {
-    if (node && typeof node === "object") {
-      let empty = true;
-      Object.keys(node).forEach(name => {
-        pathParts.push(name);
-        const result = walk(node[name]);
-        if (result === null) {
-          // If the read function returns null, omit this module from the
-          // resulting tree.
-          delete node[name];
-        } else {
-          node[name] = result;
-          empty = false;
-        }
-        assert.strictEqual(pathParts.pop(), name);
-      });
-      if (empty) {
-        // If every recursive call to walk(node[name]) returned null,
-        // remove this node from the resulting tree by returning null.
-        return null;
-      }
-    } else {
+    if (! node) {
+      return null;
+    }
+
+    if (typeof node !== "object") {
       return read(pathParts, platform);
     }
+
+    let empty = true;
+
+    Object.keys(node).forEach(name => {
+      pathParts.push(name);
+      const result = walk(node[name]);
+      if (result === null) {
+        // If the read function returns null, omit this module from the
+        // resulting tree.
+        delete node[name];
+      } else {
+        node[name] = result;
+        empty = false;
+      }
+      assert.strictEqual(pathParts.pop(), name);
+    });
+
+    if (empty) {
+      // If every recursive call to walk(node[name]) returned null,
+      // remove this node from the resulting tree by returning null.
+      return null;
+    }
+
     return node;
   }
 
--- a/packages/dynamic-import/server.js
+++ b/packages/dynamic-import/server.js
@@ -8,15 +8,38 @@ const {
 } = require("path");
 const hasOwn = Object.prototype.hasOwnProperty;
 
-const { WebApp } = require("meteor/webapp");
-const { Random } = require("meteor/random");
-
 require("./security.js");
 
 const client = require("./client.js");
-const platforms = Object.keys(dynamicImportInfo);
 
-platforms.forEach(platform => {
+Meteor.startup(() => {
+  if (! Package.webapp) {
+    // If the webapp package is not in use, there's no way for the
+    // dynamic-import package to fetch dynamic modules, so we should
+    // abandon the rest of the logic in this module.
+    //
+    // If api.use("webapp") appeared in dynamic-import/package.js, then
+    // Package.webapp would always be defined here, of course, but that
+    // would be a bad idea, because the dynamic-import package should not
+    // single-handedly force a dependency on webapp if the program does
+    // not otherwise need a web server (e.g., when the program is an
+    // isopacket or build plugin instead of a web application).
+    //
+    // Note that the client.js module (imported above) still defines
+    // Module.prototype.dynamicImport, which will work as long as no
+    // modules need to be fetched.
+    return;
+  }
+
+  Object.keys(dynamicImportInfo).forEach(setUpPlatform);
+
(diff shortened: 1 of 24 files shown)


packages/dynamic-import/server.js
```diff
@@ -1,48 +1,129 @@
...
+  request.on("end", () => {
+    response.setHeader("Content-Type", "application/json");
+    response.end(JSON.stringify(readTree(
+      JSON.parse(Buffer.concat(chunks)),
+      getPlatform(request)
+    )));
+  });
+}
+
+function getPlatform(request) {
```
hwillson: Some of the changes in this file revolve around the fact that a configurable `platform` is now supported. Should we add an extra comment somewhere explaining why? Maybe explaining this is better suited elsewhere (like the package README or API docs), but I just thought I'd mention it.

packages/http/deprecated.js
```diff
@@ -1,3 +0,0 @@
-// The HTTP object used to be called Meteor.http.
-// XXX COMPAT WITH 0.6.4
```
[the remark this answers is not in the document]
benjamn: Think it's sufficient to mention this in `History.md`?

packages/dynamic-import/server.js
benjamn: We've always had `web.browser` and `server` as separate platforms, it's just that we were able to determine which one to use based on `this.connection` in the method implementation before, and now we have to detect the platform differently. The detection is still fully automatic, so I don't know that it needs to be advertised as a public API. More implementation comments could be nice, though!

Side note: I suppose `web.cordova` could also appear here, though we intentionally force dynamic modules into the main JS bundle for Cordova, because we don't have to keep re-downloading the bundle (so it's ok if it's bigger), and we can't necessarily `eval` dynamic code at runtime in some Cordova environments, due to security restrictions.

hwillson: Make sense @benjamn - looks good as it is.

packages/http/deprecated.js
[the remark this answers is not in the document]
hwillson: Yes, I think so - people have had long enough to plan for this one being deprecated.

packages/dynamic-import/server.js
```diff
@@ -1,48 +1,129 @@
...
+  return platform;
+}
+
+function readTree(tree, platform) {
+  const pathParts = [];
 
-    return walk(tree);
+  function walk(node) {
+    if (node && typeof node === "object") {
+      let empty = true;
```
abernix: The number of lines inside this `if` conditional has grown a bit with this commit.  It could improve readability, and get rid of the cliff if the `else` and its `return` were first:

```js
function walk(node) {
  if (typeof node !== "object") {
    return read(pathParts, platform);
  }

  let empty = true;
  // ... so on.
  return node;
}
```

packages/http/deprecated.js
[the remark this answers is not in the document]
abernix: It was mentioned in the 0.6.5 [`History.md`](https://github.com/meteor/meteor/blob/devel/History.md#v065-2013-08-14)! 😉

abernix:
Though I guess the package test failures might be relevant!

wildhart:
Apologies if  this is not the correct place for this question:

@benjamn wrote:

> Although the implementation of the /__dynamicImport endpoint is a bit too complicated to allow serving dynamic modules from a CDN, fetching modules individually from a CDN remains a possibility for future experimentation.

I'd love to understand more details about _why_ it's too complicated to serve dynamic module bundles via a CDN, or whether there are any plans for the aforementioned 'future experimentation' of serving individual modules via CDN?

I absolutely LOVE the Meteor implementation of code-splitting, I tip my hat to all of you.  To me it just seems natural to serve it via CDN so I'm curious why it can't be...

Obviously the requested module path names would have to form part of the query URL of a GET request, but after that isn't the business logic of building the bundle the same?  Is the character limit of a GET request a limiting factor?

I understand the efficacy of the IndexedDB cache makes a CDN of little benefit to returning users, but if a site gets lots of short term visitors then CDN would help reduce server demand (particularly since the [server code](https://github.com/meteor/meteor/blob/devel/packages/dynamic-import/server.js) isn't memoized or cached on the server), as well as increase response times for users remote from the server.

I'm happy to move this question to the forums if you prefer...

Issue #9381, opened 2017-11-16
Slow rebuilds because npm running on every change for local packages in some occasions


sebakerckhof:
My app rebuilds (meteor 1.6, but it was like that before) are very slow and I decided to take a look as to where most of the time is spend.
I have quite a lot of local packages, and most have an `Npm.depends` section.
When looking at the profiling output I see more than 10 seconds per change to any file is spend on `meteorNpm.runNpmCommand`.

Now, I don't understand why this needs to run if I don't change any npm dependency, I don't even touch any `package.js` file...

Note that there's also a lot of `The handle(XXXX) returned by watching ... is the same with an already watched path ...` messages, so https://github.com/meteor/meteor/issues/9131 is certainly still there and I've got a feeling this is related.

Example trace:

```
| (#4) Profiling: ProjectContext prepareProjectForBuild
barco:edu: updating npm dependencies -- get-value, xml2js, gm, ng-focus-if,
babel-plugin-angularjs-annotate, babel-plugin-rewire...
|  Preparing to build package barco:edu      -
| ProjectContext prepareProjectForBuild........................14,223 ms (1)
| ├─ _initializeCatalog...........................................240 ms (1)
| │  └─ LocalCatalog#initialize...................................239 ms (1)
| │     ├─ LocalCatalog#_computeEffectiveLocalPackages              9 ms (1)
| │     └─ LocalCatalog#_loadLocalPackages........................230 ms (1)
| │        ├─ PackageSource#initFromPackageDir....................193 ms (70)
| │        │  ├─ files.stat                                         2 ms (70)
| │        │  ├─ optimistic readFile                                1 ms (70)
| │        │  ├─ optimistic statOrNull.............................20 ms (70)
| │        │  │  ├─ files.stat                                      2 ms (25)
| │        │  │  └─ other optimistic statOrNull                    17 ms
| │        │  ├─ runJavaScript package.js                          27 ms (70)
| │        │  ├─ sqlite query                                      31 ms (38)
| │        │  └─ other PackageSource#initFromPackageDir           112 ms
| │        └─ other LocalCatalog#_loadLocalPackages                36 ms
| ├─ _resolveConstraints...........................................85 ms (1)
| │  ├─ Select Package Versions....................................23 ms (1)
| │  │  └─ new CS.Input                                            23 ms (1)
| │  ├─ files.writeFileAtomically...................................3 ms (1)
| │  │  ├─ files.writeFile                                          2 ms (1)
[...]

PR #9398, opened 2017-11-20
Don't update npm dependencies when it's not needed

sebakerckhof:
Attempt to fix: https://github.com/meteor/meteor/issues/9381

Short story: Meteor updates a package's npm dependencies on every change to any of the package files when one of the npm dependencies was scoped or when using a private npm server.

Root causes explained here: https://github.com/meteor/meteor/issues/9381#issuecomment-345709113
and here: https://github.com/meteor/meteor/issues/9381#issuecomment-345746220

Basically the shrinkwrap meteor created didn't include scoped packages (weird this hasn't resulted in more bugs before) and the comparison between requested and installed modules couldn't deal with private npm servers. 

Don't know if this will fix every situation (e.g. what is the resolved URL if the private NPM server sends redirects to another url). But I don't know of any such servers. It also only works if you've set the `NPM_CONFIG_REGISTRY` envar, but I believe that's the only way to install meteor packages npm dependencies from a private npm server (CMIIW)?

This saves me about 2 minutes per initial run and around 10 seconds on every reload.

### DIFF
--- a/tools/isobuild/meteor-npm.js
+++ b/tools/isobuild/meteor-npm.js
@@ -920,6 +920,11 @@ function getInstalledDependenciesTree(dir) {
       const pkgDir = files.pathJoin(nodeModulesDir, item);
       const pkgJsonPath = files.pathJoin(pkgDir, "package.json");
 
+      if (item.startsWith("@")) {
+        Object.assign(result, ls(pkgDir));
+        return;
+      }
+
       let pkg;
       try {
         pkg = JSON.parse(files.readFile(pkgJsonPath));
@@ -927,7 +932,9 @@ function getInstalledDependenciesTree(dir) {
         if (! pkg) return;
       }
 
-      const info = result[item] = {
+      const name = pkg.name || item;
+
+      const info = result[name] = {
         version: pkg.version
       };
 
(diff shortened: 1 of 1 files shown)


sebakerckhof:
Update: just discovered there is also another problem that will occur when using namespaced npm packages.
Those packages will be installed in subdirectories, like:
```
|-node-modules
  |-@namespace
    |-package
      |-package.json
```
But Meteor doesn't read subdirectories. This is likely to affect much more people. So I'll also try to fix that one.

tools/isobuild/meteor-npm.js
```diff
@@ -1136,7 +1143,8 @@ var minimizeDependencyTree = function (tree) {
   var minimizeModule = function (module) {
     var version;
     if (module.resolved &&
-        !module.resolved.match(/^https?:\/\/registry.npmjs.org\//)) {
+        !module.resolved.match(/^https?:\/\/registry.npmjs.org\//) &&
+        !module.resolved.startsWith(process.env.NPM_CONFIG_REGISTRY)) {
```
benjamn: On second thought, let's guard against the possibility that `process.env.NPM_CONFIG_REGISTRY` is not defined?

Issue #9401, opened 2017-11-20
Minimongo inconsistency with MongoDB server side on $type


axelvaindal:
Hi,

It seems there is inconsistency in how `minimongo` is handling the `$type` operator client side in comparaison with how MongoDB handle it.

As the docs says [here](https://docs.mongodb.com/manual/reference/operator/query/type/): 
>Changed in version 3.2: $type operator accepts string aliases for the BSON types in addition to the numbers corresponding to the BSON types. Previous versions only accepted the numbers corresponding to the BSON type.

Using alias in `minimongo` throws this error : 

```
Uncaught Error: $type needs a number
    at Object.compileElementSelector (minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:199)
    at minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:1183
    at Array.map (<anonymous>)
    at operatorBranchedMatcher (minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:1165)
    at compileValueSelector (minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:702)
    at minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:683
    at Array.map (<anonymous>)
    at compileDocumentSelector (minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:654)
    at Matcher._compileSelector (minimongo.js?hash=9560525385c34d850d6779d215533afb6d25db3f:4252)
    at new Matcher 
```

Thus, such query are working fine in Meteor server side but not client side : 

`User.find({field: {$type: "null"}}).fetch();`

Now, I know there have been a lot of discussion about legacy in Meteor, so I don't know how things are going this way for MongoDB, but considering MongoDB 3.2 will attain ELD on September 2018, I suppose this is not a big deal to consider matching default server MongoDB version (3.2 now & soon 3.4) for `minimongo` as well.
In this case, this is about supporting aliases client side for the `$type` operator.

PR #9402, opened 2017-11-21
Add support for $type aliasses.

sebakerckhof:
Fix: https://github.com/meteor/meteor/issues/9401

Note: I added some extra code that checks the operand values which is not strictly necessary, but now the errors are exactly the same as when using mongo shell.

### DIFF
--- a/packages/minimongo/common.js
+++ b/packages/minimongo/common.js
@@ -87,8 +87,40 @@ export const ELEMENT_OPERATORS = {
     // should *not* include it itself.
     dontIncludeLeafArrays: true,
     compileElementSelector(operand) {
-      if (typeof operand !== 'number') {
-        throw Error('$type needs a number');
+      if (typeof operand === 'string') {
+        const operandAliasMap = {
+          'double': 1,
+          'string': 2,
+          'object': 3,
+          'array': 4,
+          'binData': 5,
+          'undefined': 6,
+          'objectId': 7,
+          'bool': 8,
+          'date': 9,
+          'null': 10,
+          'regex': 11,
+          'dbPointer': 12,
+          'javascript': 13,
+          'symbol': 14,
+          'javascriptWithScope': 15,
+          'int': 16,
+          'timestamp': 17,
+          'long': 18,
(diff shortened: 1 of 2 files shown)


packages/minimongo/common.js
```diff
@@ -87,8 +87,41 @@ export const ELEMENT_OPERATORS = {
...
+          'dbPointer': 12,
+          'javascript': 13,
+          'symbol': 14,
+          'javascriptWithScope': 15,
+          'int': 16,
+          'timestamp': 17,
+          'long': 18,
+          'decimal': 19,
+          'minKey': -1,
+          'maxKey': 127,
```
benjamn: Is there any way we could get this information from some npm package, or from the `mongo` package, somehow?

packages/minimongo/common.js
```diff
@@ -87,8 +87,41 @@ export const ELEMENT_OPERATORS = {
...
+        };
+        if (!operandAliasMap.hasOwnProperty(operand)) {
```
abernix: In order to protect against the (unlikely, but seemingly more likely than usual case, given that this is a list of operands) scenario where operands includes a property called `hasOwnProperty`, could we switch this to?

```js
Object.prototype.hasOwnProperty.call(operandAliasMap, operand);
```

packages/minimongo/common.js
sebakerckhof: Did a quick search and couldn't find anything (the mongodb npm driver at least doesn't contain it). Also, the `mongo` package has `minimongo` as a dependency, not the other way around.

packages/minimongo/common.js
sebakerckhof: yep, apparently there was already a `hasOwn` shorthand foreseen.

abernix: Yeah, we frequently opt to use that technique! Always feel free to use it as you see fit! :smile:

packages/minimongo/common.js
benjamn: OK, I can't find anything useful either!

Issue #5415, opened 2015-10-12
meteor shell somehow injects `path` into the global namespace on the server


jonmc12:
I've attempted to show a reproduction of the underlying cause of this issue [in a fork of the todos example](https://github.com/jonmc12/todos-path).

```
git clone https://github.com/jonmc12/todos-path.git
cd todos-path
meteor
```

from client:  `Meteor.call('testPath')` will trigger the error:

```
Exception while invoking method 'testPath' ReferenceError: path is not defined
   at [object Object].Meteor.methods.testPath (server/bootstrap.js:55:1)
   at maybeAuditArgumentChecks (livedata_server.js:1692:12)
   at livedata_server.js:708:19
   at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
   at livedata_server.js:706:40
   at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
   at livedata_server.js:704:46
   at tryCallTwo (/Users/jonmc12/.meteor/packages/promise/.0.5.0.h2jsml++os+web.browser+web.cordova/npm/node_modules/meteor-promise/node_modules/promise/lib/core.js:45:5)
   at doResolve (/Users/jonmc12/.meteor/packages/promise/.0.5.0.h2jsml++os+web.browser+web.cordova/npm/node_modules/meteor-promise/node_modules/promise/lib/core.js:171:13)
   at new Promise (/Users/jonmc12/.meteor/packages/promise/.0.5.0.h2jsml++os+web.browser+web.cordova/npm/node_modules/meteor-promise/node_modules/promise/lib/core.js:65:3)
```

Next,  open `meteor shell` and from the shell `Meteor.call('testPath')` - you will see this works.  It seems strange to me that the method works differently on the client than the server.

Now, go back into your browser and issue `Meteor.call('testPath')`.  This command will no longer trigger an error.
## Additional background

In my own code repo, there seems to be a set of package dependencies that was exacerbating this issue in a way that the error appears intermittently and non-deterministically.  Some of the things I observed that would cause the error to come and go:
1.  Changed `api.use('check')` to `api.use('check', 'client')` in package.js in angular-meteor
2.  Change version of `check` in versions.json in angular-meteor
3.  `meteor add check` and `meteor remove check`
4.  Reseting server via `meteor run ios-device`
5.  Waiting several hours (I should note I have a few cron jobs running)
[...]

PR #9406, opened 2017-11-21
Correct global _ destruction via `meteor shell`.

### LINKED ISSUE #9276: [1.6] Starting meteor shell overwrite _ in the main server
#### The version of Meteor showing the problem.
```
$ meteor --version`
 1.6
$ meteor npm version -> 
{ test: '0.0.1',
  npm: '5.4.2',
  ares: '1.10.1-DEV',
  cldr: '31.0.1',
  http_parser: '2.7.0',
  icu: '59.1',
  modules: '57',
  nghttp2: '1.25.0',
  node: '8.8.1',
  openssl: '1.0.2l',
  tz: '2017b',
  unicode: '9.0',
  uv: '1.15.0',
  v8: '6.1.534.42',
  zlib: '1.2.11' }
```
#### The last version of Meteor where the problem did _not_ occur (if applicable)
1.5.2 - which shipped with node version != 8.8.1

#### The operating system you're running Meteor on.

macOs Sierra beta

#### The expected behavior.

when i'm starting meteor shell, i'm not expecting _ to be overwritten on the server side. 

#### The actual behavior.

After starting a shell, _ will be undefined in the server and break the app anywhere it we used _
Stopping the shell doesn't fix the issue. 
This makes it really hard to test things in dev mode. 

#### A **simple** reproduction!

- install meteor 1.6
- write some code that rely on _. You can optionally define _ = require('underscore') (it should be available globally already since underscore ship with meteor)
- start the server
- start a shell
- notice _ is undefined
- notice when you do const _ = require('underscore') and _ is still undefined
- call your function relying on _ and watch it fail. 

This might be related: https://github.com/nodejs/node/issues/14226

abernix:
_(PR body is the commit message body from https://github.com/meteor/meteor/commit/15b24ff6d1242dc80f52c94a053c5b1c8742efa5.)_

A truthy `useGlobal` option when calling Node's `repl.start()` will use the `global` context, which will allow global Meteor variables (such as the global `_` provided by the `underscore`) to be available in the context of the REPL shell.  This sounds desirable, and in fact, the original
implementation set it as such, only to be later changed to `false` in 7c7e52f, specifically to avoid the undesirable behavior of Node REPL stomping on the global `_` variable with its own special-meaning `_` variable which is set to the value of the most recently eval'd expression.

This `useGlobal ` was once again changed to `true` to fix the lost tab-completion functionality, thanks to 2443d83, which also implelented special mutator on `_` to avoid breaking it.  As it's probably clear now, this has been a struggle, and because of Node using its own `Object.defineProperty('_', ...)` as of Node.js 6 (in [node/node#5535](https://github.com/nodejs/node/pull/5535/files)), this problem has surfaced again, as reported in #9276.

This changes the behavior to another implementation which starts with `useGlobal` set to `false`, but then sets the context immediately to `global`, which seems to avoid the problem.

This does maintain the existing behavior of our special double-underscore variable, in order to get the value of the last REPL expression, in the same manner as `_` would in a normal REPL, by
accessing the internal `last` property.  Since this is also undocumented (on our end), this seemed reasonable.

---

I'd recommend reviewing this PR at one time, in it's entirety, but keep in mind that the individual commits provide additional commentary in their own commit messages (and changes).

I attempted to make tests for this, but since the self tests seem to invoke the `evaluateAndExit` functionality within `meteor shell` (due to their lack of a full TTY), the `context` is not affected in any way.

Fixes #9276.

### DIFF
--- a/packages/shell-server/shell-server.js
+++ b/packages/shell-server/shell-server.js
@@ -10,7 +10,7 @@ import {
   writeSync,
 } from "fs";
 import { createServer } from "net";
-import { default as moduleRepl } from "repl";
+import { start as replStart } from "repl";
 
 const INFO_FILE_MODE = parseInt("600", 8); // Only the owner can read or write.
 const EXITING_MESSAGE = "Shell exiting...";
@@ -208,7 +208,7 @@ class Server {
       options.output = null;
     });
 
-    const repl = this.repl = moduleRepl.start(options);
+    const repl = this.repl = replStart(options);
 
     // This is technique of setting `repl.context` is similar to how the
     // `useGlobal` option would work during a normal `repl.start()` and
--- a/packages/shell-server/shell-server.js
+++ b/packages/shell-server/shell-server.js
@@ -49,6 +49,11 @@ export function disable(shellDir) {
   } catch (ignored) {}
 }
 
+// Shell commands need to be executed in a Fiber in case they call into
+// code that yields. Using a Promise is an even better idea, since it runs
+// its callbacks in Fibers drawn from a pool, so the Fibers are recycled.
+const evalCommandPromise = Promise.resolve();
+
 class Server {
   constructor(shellDir) {
     var self = this;
@@ -126,11 +131,36 @@ class Server {
       }
       delete options.columns;
 
+      // Immutable options.
+      _.extend(options, {
+        input: replInputSocket,
+        useGlobal: false,
+        output: socket
+      });
+
+      // Overridable options.
+      _.defaults(options, {
+        prompt: "> ",
+        terminal: true,
+        useColors: true,
+        ignoreUndefined: true,
+      });
+
+      // The prompt during an evaluateAndExit must be blank to ensure
+      // that the prompt doesn't inadvertently get parsed as part of
+      // the JSON communication channel.
       if (options.evaluateAndExit) {
-        evalCommand.call(
-          Object.create(null), // Dummy repl object without ._RecoverableError.
+        options.prompt = "";
+      }
+
+      // Start the REPL.
+      self.startREPL(options);
+
+      if (options.evaluateAndExit) {
+        self._wrappedDefaultEval.call(
+          Object.create(null),
           options.evaluateAndExit.command,
-          null, // evalCommand ignores the context parameter, anyway
+          global,
           options.evaluateAndExit.filename || "<meteor shell>",
           function (error, result) {
             if (socket) {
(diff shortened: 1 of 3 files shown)


hwillson:
Interesting - I didn't realize adding an issue link in a review comment doesn't register that link in the linked to issue thread. So, recapping here to fix the issue thread - these changes also fix https://github.com/meteor/meteor/issues/5415.

packages/shell-server/shell-server.js
```diff
@@ -182,25 +182,25 @@ class Server {
...
-
-      // Expose the last REPL result as __ instead of _.
-      set: function(lastResult) {
-        repl.context.__ = lastResult;
+    // Implement an alternate means of fetching the return value,
+    // via `__` (double underscore) as originally implemented in:
+    // https://github.com/meteor/meteor/commit/2443d832265c7d1c
+    Object.defineProperty(repl.context, "__", {
+      get: () => repl.last,
+      set: (val) => {
```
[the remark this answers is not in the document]
abernix: We don't, but I copied the behavior that Node.js uses in the native `repl` code located here:

https://github.com/nodejs/node/blob/ad8257fa5b9795d3d79fa4a91d0f18c43f024ab3/lib/repl.js#L575-L587

...in which they also allow setting of `_`, which gets stored on the `repl` instance's `this.last` property (in this code `repl.last`)

packages/shell-server/shell-server.js
```diff
@@ -463,13 +409,83 @@ function isRecoverableError(e, repl) {
...
+      if (current.trim()) prevTokenChar = current;
+    }
+
+    if (current === '\'' || current === '"') {
+      stringLiteral = current;
+    }
+  }
+
+  return stringLiteral ? lastChar === '\\' : isBlockComment;
+}
```
benjamn: If there was any way to avoid copying this code here, I think that would make our REPL implementation much more maintainable in the future!

packages/shell-server/shell-server.js
```diff
@@ -151,46 +181,64 @@ class Server {
...
+    // by setting `useGlobal` back to true, we allow the default eval function
+    // to use the desired `runInThisContext` method (https://git.io/vbvAB).
     repl.context = global;
+    repl.useGlobal = true;
+
+    // In order to avoid duplicating code here, specifically the complexities
+    // of catching so-called "Recoverable Errors" (https://git.io/vbvbl),
+    // we will wrap the default eval, run it in a Fiber (via a Promise), and
+    // give it the opportunity to decide if the user is mid-code-block.
+    const defaultEval = repl.eval;
```
benjamn: Awesome!

PR #9409, opened 2017-11-22
Modernize tinytest by using ecmascript.

benjamn:
This is fairly standard stuff by now, though I'd also like to support returning `Promise`s from test functions.

### DIFF
--- a/packages/tinytest/tinytest.js
+++ b/packages/tinytest/tinytest.js
@@ -1,54 +1,50 @@
-var Future;
-if (Meteor.isServer)
-  Future = Npm.require('fibers/future');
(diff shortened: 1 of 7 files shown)


packages/tinytest/tinytest.js
```diff
@@ -243,272 +239,267 @@ _.extend(TestCaseResults.prototype, {
...
+          }
+          return onEvent(event);
+        },
+        reject,
+        stop_at_offset
+      );
+
+      const result = this.func(results, resolve);
+      if (result && typeof result.then === "function") {
+        result.then(resolve);
```
benjamn: This should be `result.then(resolve, reject)`, my bad.

Issue #6674, opened 2016-04-01
[1.3] webapp appears to force-serve manifest.json


Primigenus:
Since 1.3 it seems like apps are serving a `/manifest.json` file which appears to contain a long list of whether to cache specific resources. I imagine this might be used by something like mobile support to manage whether it needs to update things during hot code reload?

The line causing this behaviour appears to originate here: [2ccb746#L514](https://github.com/meteor/meteor/blob/2ccb7467c9bb5889a3c36739d2d8b59a0656961c/packages/webapp/webapp_server.js#L514)

Steps to repro:
1. create a new Meteor app using 1.3 and navigate to `http://localhost:3000/manifest.json`.

This breaks existing apps that were relying on their own `manifest.json` hosted in eg. `/public` being available, for instance in the case of wanting to provide a [web app manifest according to the W3C Working Draft](https://w3c.github.io/manifest/) using `<link rel="manifest" href="/manifest.json">`.

mitar:
Huh, yes, but that was fix of the omission in the past. The code was meant to server it, but it was not serving it.

There is a TODO to rename it to `program.json`. Maybe this could be a fix? To rename Meteor manifest to `program.json`?

Primigenus:
Yes, or consider naming internal Meteor files like this something that will less likely conflict with the app's files. Maybe serve from a /__meteor__/ folder or something?

martijnwalraven:
I wasn't aware this behavior was new in Meteor 1.3. Mobile support only needs `/__cordova/manifest.json`, but I always assumed it would also serve this for the main architecture (although it isn't used anywhere as far as I know).

Not sure how https://github.com/meteor/meteor/commit/53205bfcd14dfd064ab4d6f7ebc7fee789c54a2f and https://github.com/meteor/meteor/pull/6101/files change this.

mitar:
No, this was the fix: https://github.com/meteor/meteor/pull/5713

AndreasGalster:
Will this be fixed or is there another way how we can serve our manifest.json in the meantime?

geremora:
I need serve my own manifest.json, but in meteor 1.3.4.1 still appears to contain a long list of whether to cache specific resources.

abernix:
[...]

Issue #9407, opened 2017-11-22
appcache and dynamic-imports don't play well together


CaptainN:
It looks like appcache and dynamic-imports don't play well together. Appcache adds all the additional packages (those not in the initial bundle) to the app.manifest individually (conclusion based on looking in app.manifest), to instruct the browser to grab all those - but dynamic-imports doesn't seem to grab those files from the appcache. The effect is that the app won't load in offline mode, when it needs to grab a dynamic module (even if the module should be in the indexedDB cache - I'm not certain why that doesn't work - maybe this is the bug).

I confirmed this by disabling my use of dynamic-imports through React Loadable to get my app working offline - I have not confirmed my assumptions about _why_ it failed with any tests, though I did quickly read through the dynamic-imports package source, and didn't see anything which would indicate it's trying to load from appcache.

Additionally, if my assumptions are correct, we are double loading dynamic modules, which is not desired - once through appcache manifest, and once over DDP and stored in indexedDB.

Suggested solution - only add the initial bundle and CSS to the appcache manifest (or all the initial packages in development mode), and use a prefetch model to grab dynamic-modules over DDP and store in indexedDB (like it already does on demand) when the app is ready.

Alternative solution - make dynamic-imports load additional modules from appcache instead of over DDP/indexedDB, when that is appropriate. This solution is less desirable due to limits with the (deprecated) appcache tech. For exmaple, I have one app which it has been very difficult to get my bundle size down under the appcache file size limit imposed by iOS Safari, and that can [cause all kinds of horrendous side effects](https://alistapart.com/article/application-cache-is-a-douchebag).

More notes about general offline support [in this forum post](https://forums.meteor.com/t/appcache-and-dynamic-import-incompatible/40652).

benjamn:
If the `appcache` package is in use, I'm inclined to say dynamic modules should get eagerly prefetched over DDP ([or soon HTTP](https://github.com/meteor/meteor/pull/9384)) and stored (as usual) directly in IndexedDB. In other words, I don't think dynamic `import()` should read from `appcache`, but I agree that the app should fetch all available dynamic modules eagerly, to make sure everything is available offline.

CaptainN:
It sounds like we are on the same page. So the tasks would be:

1. Add eager loading capability somewhere - to either appcache or dynamic-imports. (Is there already a way to queue up files for dynamic import?)
2. Stop including dynamic modules in the appcache app.manifest to avoid double loading (and reduce pressure on appcache).

[...]

PR #9424, opened 2017-11-28
Stop serving the application manifest from /manifest.json

hwillson:
As discussed in #6674, Meteor currently serves its own manifest file from `/manifest.json`. This location is not application configurable, and can conflict with other non-Meteor defined manifest files, that are already being served from this location. For example, [Google's progressive web app documentation](https://developers.google.com/web/fundamentals/web-app-manifest/) suggests using the `/manifest.json` location for its manifest. The newest version of the [web app manifest spec](https://developer.mozilla.org/en-US/docs/Web/Manifest) recommends using a different location for the web app manifest, like

```html
<link rel="manifest" href="/manifest.webmanifest">
```

but the `/manifest.json` location was recommended at one point. Regardless, there isn't really any reason why Meteor needs to use the `/manifest.json` location, so let's move it elsewhere.

This PR moves moves `/manifest.json` to `/__meteor__/webapp/manifest.json`.

A few extra notes:

- It's worth noting that Meteor core doesn't appear to rely on or use the served `/manifest.json` file in any way. We could consider removing it, but as discussed in #6674, there are a few ways in which applications can leverage this information. Since we're already serving it up this PR assumes we're going to keep it, but let me know if you think we should drop it instead.

- At some point we might want to consider serving all application metadata via a similar `/__meteor__/[package name]/[file]` endpoint. The `bundle-visualizer` is using `_meteor/[package name]/stats` for example - at some point we might want to consolidate this. For now though this PR is just addressing the issue raised in #6674 directly.

Fixes https://github.com/meteor/meteor/issues/6674.

Thanks!

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@
 * The previously served `/manifest.json` application metadata file is now
   served from `/__meteor__/webapp/manifest.json`.
   [Issue #6674](https://github.com/meteor/meteor/issues/6674)
+  [PR #9424](https://github.com/meteor/meteor/pull/9424)
 
 * The bundled version of MongoDB used by `meteor run` in development
   on 64-bit architectures has been updated to 3.4.10. 32-bit architectures
--- a/History.md
+++ b/History.md
@@ -1,5 +1,9 @@
 ## v.NEXT
 
+* The previously served `/manifest.json` application metadata file is now
+  served from `/__meteor__/webapp/manifest.json`.
+  [Issue #6674](https://github.com/meteor/meteor/issues/6674)
+
 * The bundled version of MongoDB used by `meteor run` in development
   on 64-bit architectures has been updated to 3.4.10. 32-bit architectures
   will continue to use MongoDB 3.2.x versions since MongoDB is no longer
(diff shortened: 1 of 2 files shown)


benjamn:
This reminded me of https://github.com/meteor/meteor/issues/9407, which concerns the manifest used by `appcache`, but that uses [`/app.manifest`](https://github.com/meteor/meteor/blob/c09d8e57f01ed95c987f7a8b34b4d2de1a75b249/packages/appcache/appcache-server.js#L46) instead of `/manifest.json`.

benjamn:
Pushed these two commits to normalize other URLs with this style: 95d093a17ad37114af38f4aba3ecf5e56f80dc75, 4aeb453c7b0a0a9556a11ea1928e63ab11302ef4

Floriferous:
Does this mean all apps should now have a directory called `__meteor__/webapp/` with the `manifest.json` in it? 

Not totally clear if this requires an action from developers (and if `manifest.json` actually works, since there has been some confusion in the past, including #6674 as already mentioned here), any best-practice to recommend to people stumbling on this?

abernix:
@Floriferous The `manifest.json` that is being served is Meteor-specific, and used internally by Meteor.  It's not to be confused with the so called "Web App Manifest" [as specified here](https://developers.google.com/web/fundamentals/web-app-manifest/) which I'm assuming you're referring to.  It used to be served at `/manifest.json`, but is now served at `/__browser/manifest.json`.  You shouldn't need to make any changes unless you were consuming this internal Meteor file yourself, in some manner.

Issue #7789, opened 2016-09-20
[dev] `meteor reset` with the option to preserve the database


abernix:
Running `meteor reset` is very handy for 'repairing' a project that may have become corrupted or is acting funky for _whatever_ reason, but often times (in fact, _usually_, from a support standpoint) wiping the users database is unnecessary.  While it does meet the requirement for a true `meteor reset`, it's a bit overboard for most support issues and usually needs to be prefaced with "btw, be careful, this will wipe your whole dev DB!!!" – because I don't know about most developers, but I like my dev DB the way it is.   What **is** usually desirable is getting rid of `.meteor/local/{build,bundler-cache,isopacks,plugin-cache}` – essentially, everything in `.meteor/local/` besides the `db` directory.

Two initial propositions would be:
- Adding a `--preserve-db` argument to the existing `meteor reset` command
- Adding a `meteor clear-cache` (or any other name _TBD_) command.

Discuss.

pozylon:
I would love to see the second proposition, "meteor clean" command. --preserve-db is risky (chance is higher of accidently do the hard reset).

axelvaindal:
I also would love to see the second proposition.

I suppose it is way much clearer to prepare a second command to do this, either one to clean the database or one to do the emptying .mongo/local thing.

jerodmoore:
meteor clean sounds like a logical name for this function. So logical in fact I gave it a try instinctively when trying to get past some build problems. Meteor reset --operator would be my second.

abernix:
I'm concerned that `meteor clean` and `meteor reset` have too similar of a meaning.  And I'm not sure if I'm a fan of adding another command anymore.

What if we changed the current functionality of `meteor reset` so it would _no longer_ wipe the database _unless_ a flag was passed.  Maybe `meteor reset --with-db`?  This might break some automated test workflows that might be doing this right now but would be very unlikely to break anything in production.  It's very clear though, and we could add a message about it to the existing `meteor reset` command.

axelvaindal:
Hum...
This sounds like a good approach to me.

So `meteor reset` would clean the actual `.meteor/local/` folder (except for the `db` directory) if no flags are passed, but would wipe the database if you use it with a flag.

For instance:

`meteor reset` => Clean the `.meteor/local/` folder and other things but leave the `db` directory untouched.
`meteor reset --with-db` => Do the current process and thus also clean the `db` directory.

If this is the way you are thinking about, I'm okay about it.
[...]

PR #9425, opened 2017-11-28
Add option to keep db to reset command

sebakerckhof:
Implementation for https://github.com/meteor/meteor-feature-requests/issues/217 .

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,7 @@
 ## v.NEXT
 
+* The `meteor reset` command now has a `--keep-db` option to save the database
+
 * The bundled version of MongoDB used by `meteor run` in development
   on 64-bit architectures has been updated to 3.4.10. 32-bit architectures
   will continue to use MongoDB 3.2.x versions since MongoDB is no longer
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1286,7 +1286,12 @@ main.registerCommand({
   // error message if they try to pass one.
   maxArgs: 1,
   requiresApp: true,
-  catalogRefresh: new catalog.Refresh.Never()
+  catalogRefresh: new catalog.Refresh.Never(),
+  options: {
+    'keep-db': {
+      type: Boolean
+    },
+  }
 }, function (options) {
   if (options.args.length !== 0) {
     Console.error("meteor reset only affects the locally stored database.");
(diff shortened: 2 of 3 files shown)


abernix:
Hmm, this doesn't seem to match up with the direction we discussed in https://github.com/meteor/meteor/issues/7789, (specifically, https://github.com/meteor/meteor/issues/7789#issuecomment-274089655).  Did you run into problems with that design?

sebakerckhof:
Well, I basically didn't read the whole thread and never saw that comment.
Anyway, [here](https://github.com/meteor/meteor/issues/7789#issuecomment-274639647) you state we should only prompt in interactive mode, but how do I know if I'm in interactive mode?

abernix:
There were some important considerations and discussion in the thread about `meteor rebuild`, a likely very useful command which we currently don't promote or even encourage.  In fact, `meteor rebuild --help` currently states _not to use it_ 🤷‍♂️: https://github.com/meteor/meteor/blob/ac90219f20f2bc2f85a8b5ffe3afd77de2d8f5a2/tools/cli/help.txt#L874-L875

If we remove that disclaimer from the help, and make it work properly, it likely serves the purpose of what most folks are trying to accomplish by running `meteor reset`, which is rebuilding the various caches which live in `.meteor/local` (alongside the database).  In addition, it would provide the ability to actually build a project without `meteor run`-ing it, a feature some have asked for before, especially in CI.

Unfortunately, the surprise with using `meteor reset` is that the **development** database is wiped, which some developers really prefer to save (stubs, test data, etc.).  Because of this, I think requiring the `--keep-db` on `meteor reset` to preserve the database, as you have here, only perpetuates that problem.

I'll copy the rest of the thoughts below into the feature request, but thanks for asking about the interactive prompt:

---

I think my thought about the interactive prompt probably wasn't well thought out.  My goal was to avoid blocking in scripts which already use this workflow, but I think that's not really reasonable to do.  We should just make the `meteor reset` inform that it's wiped the development database and that if the user wants to avoid that behavior (and simply reset the other items in `.meteor/local`) they should use `meteor rebuild`.

If we really wanted to prevent accidental destruction, we could provide a message with a 10-second timeout that would allow the user to Ctrl-C to avoid the deletion.  This would avoid breaking existing workflows, but still provide a bit of safety.  Of course the developer can still be destructive via an `rm -rf` to the `.meteor/local` directory and we can't save everyone!

This is of course all dependent on confirming `meteor rebuild` works as we want it to, something we should continue to discuss in https://github.com/meteor/meteor-feature-requests/issues/217.

---

Sorry this isn't quite what we were going for, but thank you for making the effort to do it.

PR #9434, opened 2017-12-04
remove dynamic js and map files from appcache (app.manifest)

CaptainN:
As described in issue #9407, dynamic-imports does not load modules from the appcache, and since we don't want the browser to download these assets twice, this patch avoids adding those files to app.manifest.

This patch does pass the appcache unit test, but I'm not sure how to create a test for this, as I'm not sure how to set gt it to load dynamic imports to test with.

### DIFF
--- a/packages/appcache/appcache-server.js
+++ b/packages/appcache/appcache-server.js
@@ -39,7 +39,7 @@ var browserDisabled = function (request) {
   return disabledBrowsers[request.browser.name];
 };
 
-function isDynamic (resource) {
+function isDynamic(resource) {
   return resource.type === 'dynamic js' ||
     (resource.type === 'json' &&
       resource.url.startsWith('/dynamic/') &&
@@ -105,7 +105,7 @@ WebApp.connectHandlers.use(function (req, res, next) {
   _.each(WebApp.clientPrograms[WebApp.defaultArch].manifest, function (resource) {
     if (resource.where === 'client' &&
         ! RoutePolicy.classify(resource.url) &&
-        !isDynamic(resource)) {
+        ! isDynamic(resource)) {
       manifest += resource.url;
       // If the resource is not already cacheable (has a query
       // parameter, presumably with a hash or version of some sort),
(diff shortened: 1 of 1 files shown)


packages/appcache/appcache-server.js
```diff
@@ -39,6 +39,13 @@ var browserDisabled = function (request) {
   return disabledBrowsers[request.browser.name];
 };
 
+function isDynamic (resource) {
```
hwillson: Since this is a named function, please remove the space between the function name and the parens: `function isDynamic(resource) {`.

packages/appcache/appcache-server.js
```diff
@@ -97,7 +104,8 @@ WebApp.connectHandlers.use(function (req, res, next) {
   manifest += "/" + "\n";
   _.each(WebApp.clientPrograms[WebApp.defaultArch].manifest, function (resource) {
     if (resource.where === 'client' &&
-        ! RoutePolicy.classify(resource.url)) {
+        ! RoutePolicy.classify(resource.url) &&
+        !isDynamic(resource)) {
```
hwillson: Please adjust your logical not's to have a space after the `!`.

packages/appcache/appcache-server.js
```diff
@@ -39,6 +39,13 @@ var browserDisabled = function (request) {
   return disabledBrowsers[request.browser.name];
 };
 
+function isDynamic(resource) {
+  return resource.type === 'dynamic js' ||
+    (resource.type === 'json' &&
+      resource.url.startsWith('/dynamic/') &&
```
benjamn: This URL prefix may change after #9439, which makes these URLs look like `/__browser/dynamic/...` but that's OK—this is mostly a note to myself to update this code in that PR.

benjamn:
@CaptainN These changes have been published in `appcache@1.1.1`. Please let us know if you encounter any further problems after running `meteor update appcache`.

CaptainN:
Awesome! Thanks!

The url check is really only used for .map files. The check for dynamic js files was relatively easier, since the type attribute reflects it's dynamic status. If the .map files were similarly marked as dynamic in the type (like 'dynamic json') that might help avoid the problem.

Alternatively, we could remove all .map files from the app cache, and not inspect the url at all, but that's a bigger change.

PR #9437, opened 2017-12-04
Fix check for Google OAuth2 expiresIn property in tokens

### LINKED ISSUE #9435: Bug in google-oauth package. 'expiresAt' field does not get populated on loginWithGoogle.
There is a bug in `google-oauth` package currently being shipped. 

Google API returns an `expiresIn` value which needs to be converted to `expiresAt` and saved in the database. 
`file: packages/google-oauth/google-server.js : line:20`
```
if (hasOwn.call(tokens, "expiresAt")) {
    serviceData.expiresAt =
      Date.now() + 1000 * parseInt(tokens.expiresIn, 10);
  }
```
The if-condition is looking for `expiresAt` value in tokens object. It should actually be looking for `expiresIn`. Due to this, currently expiresAt field is not updated on `loginWithGoogle`

Fix:
```
if (hasOwn.call(tokens, "expiresIn")) {
    serviceData.expiresAt =
      Date.now() + 1000 * parseInt(tokens.expiresIn, 10);
  }
```

@benjamn 

P.S. The fix requires change in only one place, so I did not fork and create a PR. I don't know what community guideline is for such situation. If there's a specific process to follow, please let me know. Thanks.

hwillson:
Google's OAuth2 token endpoint (https://accounts.google.com/o/oauth2/token) returns an `expires_in` property in its response, which was being stored internally as `expiresIn`. The `getServiceDataFromTokens` function would only attempt to work with the `expiresIn` value if the `tokens` object had a set `expiresAt` property, meaning the `expiresIn` property was never used. This PR switches the `expiresAt` check to `expiresIn`.

Fixes #9435.

### DIFF
--- a/packages/google-oauth/google_server.js
+++ b/packages/google-oauth/google_server.js
@@ -17,7 +17,7 @@ function getServiceDataFromTokens(tokens) {
     scope: scopes
   };
 
-  if (hasOwn.call(tokens, "expiresAt")) {
+  if (hasOwn.call(tokens, "expiresIn")) {
     serviceData.expiresAt =
       Date.now() + 1000 * parseInt(tokens.expiresIn, 10);
   }
--- a/packages/google-oauth/package.js
+++ b/packages/google-oauth/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Google OAuth flow",
-  version: "1.2.4"
+  version: "1.2.5"
 });
 
 var cordovaPluginGooglePlusURL =


packages/google-oauth/google_server.js
```diff
@@ -17,7 +17,7 @@ function getServiceDataFromTokens(tokens) {
     scope: scopes
   };
 
-  if (hasOwn.call(tokens, "expiresAt")) {
+  if (hasOwn.call(tokens, "expiresIn")) {
```
benjamn: Should we keep supporting `expiresAt` as well?

hwillson: Good call - I think we should. Changes coming - thanks!

hwillson: Hmm - thinking about this more @benjamn, `expires_in` is the only supported option in the oauth token response from Google (based on the token endpoint we're using - https://accounts.google.com/o/oauth2/token). So since `expires_at` can no longer come back in the response, I don't think we need to support it. We won't be breaking backwards compatibility by dropping it since Google's endpoint will no longer return `expires_at`.

abernix: This sounds correct to me.  Thanks for looking into this.

abernix:
Published as `google-oauth@1.2.5`.

hwillson:
Thanks @abernix! (and sorry about forgetting to bump the package version - the stickit note on my monitor that is supposed to remind me to do that keeps falling off! 🙂)

PR #9439, opened 2017-12-04
[WIP] Support a new web.browser.legacy platform to reduce bundle sizes for modern browsers.

benjamn:
This PR allows Meteor to send different static resources to "modern" and "legacy" browsers, where the "modern" threshold is collaboratively defined by any code wishing to participate in the new system.

The motivation for these changes is to allow the vast majority of application users (80%+) to enjoy the full benefits of modern evergreen browsers, without being weighed down by the needs of older 'nevergreen' browsers. At the same time, those older browsers will still receive all the polyfills, syntax transforms, and other workarounds that they need, completely automatically.

Specifically, an additional target architecture has joined the existing `os`, `web.browser`, and `web.cordova` architectures: namely, `web.browser.legacy`, which includes any resources added to `web.browser`, but may also include additional resources of its own to support older browsers.

This differential bundling technique supersedes our previous strategy of server-rendering certain polyfill scripts earlier in the HTML document (#9353, #9360), though it has essentially the same goal: to eliminate useless code from modern JS bundles, without breaking older browsers in the process.

As a straightforward example, here's a commit that removes all `Promise` polyfill code from the modern JavaScript bundle, while continuing to include it in the legacy bundle: https://github.com/meteor/meteor/commit/4609a57c983dc19bf0520cff83c5ffb24d7758b4

Note especially [the call to `setMinimumBrowserVersions`](https://github.com/meteor/meteor/commit/4609a57c983dc19bf0520cff83c5ffb24d7758b4#diff-fb89c9bb773d1a1359202251c1992b69R10), which allows the `promise` package to enforce its own assumptions about the minimum requirements for "modern" browsers:
```js
// Reference: https://caniuse.com/#feat=promises
require("meteor/modern-browsers").setMinimumBrowserVersions({
  chrome: 32,
  edge: 12,
  firefox: 29,
  mobile_safari: 8,
  opera: 20,
  safari: [7, 1],
});
```
These minimum versions are relatively ancient, since `Promise`s have been natively supported in most browsers for much longer than many other ECMAScript features. In other words, the actual threshold between `web.browser` and `web.browser.legacy` is likely to be determined by other calls to `setMinimumBrowserVersions`, but that's OK, because the `promise` package will work with or without polyfills in any browser that meets the requirements above.

More exciting examples include using vastly fewer `core-js` polyfills in the `ecmascript-runtime-client` package, and running entirely different Babel plugins in the `babel-compiler` package, so that browsers with native support for generator functions (for example) no longer need to compile them with Regenerator, and no longer need to bundle the `regenerator-runtime` package.

While it was tempting to target even finer-grained categories of browser versions (more than just `web.browser` and `web.browser.legacy`), the reality of the browser landscape today does not support that kind of specialization. More internet users than ever have access to browsers that need very few polyfills and Babel plugins, and that percentage will only keep growing.

Perhaps even more importantly, different configurations of polyfills and plugins must be tested rigorously, and testing two configurations is **much** easier than testing a whole matrix of slightly different configurations. Not to mention, it would take more time to build lots of different bundles, so building just two bundles is a bit more manageable.

This is still a work in progress, primarily because we still need to
- [ ] Settle on an appropriate set of `core-js` polyfills for the `ecmascript-runtime-client` package, and enforce those assumptions with `setMinimumBrowserVersions`.
- [ ] Use a different (smaller) set of Babel plugins to compile code for `web.browser` vs. `web.browser.legacy`.
- [ ] Make sure development rebuild performance is not impacted negatively by these changes.
- [ ] Address the remaining items in [this to-do list](https://github.com/meteor/meteor/blob/web.browser.legacy/packages/modern-browsers/TODO.md).


packages/ecmascript-runtime-client/modern.js
```diff
@@ -0,0 +1,2 @@
+require("core-js/modules/es7.string.pad-start");
+require("core-js/modules/es7.string.pad-end");
```
benjamn: This set of `core-js` polyfills is definitely too small for the current `setMinimumBrowserVersions` constraints. Further consideration will be necessary, but I'm sure we can do (much) better than `legacy.js`.

packages/meteor-base/package.js
```diff
@@ -25,12 +25,6 @@ Package.onUse(function(api) {
     'ddp',
     'livedata', // XXX COMPAT WITH PACKAGES BUILT FOR 0.9.0.
 
-    // These packages use the user agent of each incoming HTTP request to
-    // decide whether to inject <script> tags into the <head> of the
-    // response document to polyfill Web Sockets and/or ES5 support.
-    'sockjs-shim',
-    'es5-shim',
```
benjamn: I should not have removed `es5-shim` here, since the package still exists (unlike `sockjs-shim`), and has no impact on modern browser bundles.

packages/meteor/client_environment.js
```diff
@@ -36,15 +37,32 @@ Meteor = {
...
+  isCordova: false,
+
+  /**
+   * @summary Boolean variable. True if running in a "modern" JS
+   *          environment, as determined by the `modern` package.
+   * @locus Anywhere
+   * @static
+   * @type {Boolean}
+   */
+  isModern: config.isModern
```
benjamn: Is this actually useful? Maybe for collecting stats at runtime about how many clients qualified for the "modern" treatment?

packages/modern-browsers/TODO.md
```diff
@@ -0,0 +1,46 @@
...
+
+- [ ] Use different plugins in babel-compiler for `web.browser.legacy`.
+
+- [x] Fix dynamic module source map URLs (prepend `/__arch`).
+
+- [x] Fix tests failing because of changes to static resource URLs.
+
+- [ ] In development, save time by only rebuilding `web.browser` (modern)?
+
+- [ ] Try adding a `web.worker` platform and see if it works as expected.
```
benjamn: 👀😜 @mitar

packages/modern-browsers/modern.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// ECMAScript 2015 Classes
+setMinimumBrowserVersions({
+  chrome: 49,
+  edge: 12,
+  firefox: 45,
+  mobile_safari: [9, 2],
+  opera: 36,
+  safari: 9,
```
benjamn: Right now there's nothing to enforce the completeness of the browsers in this list, so there's a risk that someone could add a loose constraint for some other browser, thus accidentally causing it to receive the modern bundle, because it wasn't mentioned elsewhere with stronger constraints.

tools/isobuild/bundler.js
```diff
@@ -575,6 +575,14 @@ class File {
     // file is not intended to be served over HTTP).
     this.url = null;
 
+    // A prefix that will be prepended to this.url.
+    if (options.arch.startsWith("web.")) {
+      this.urlPrefix = "/__" +
+        options.arch.split(".").slice(1).join(".");
+    } else {
+      this.urlPrefix = "";
+    }
```
benjamn: ⚠️ 🚨 Static resource URLs now begin with `/__browser/...` or `/__browser.legacy/...`, which may be a breaking change for code that attempted to load these URLs without consulting the manifest. 🚨 ⚠️ 

The good news is that this code is now the single authoritative source of those `/__browser/...` and `/__browser.legacy/...` URL prefixes, which is why very little other code (besides some tests) had to change to accommodate the new URLs.

packages/modern-browsers/modern.js
```diff
@@ -0,0 +1,123 @@
...
+    if (a < b) {
+      return false;
+    }
+  }
+
+  return false;
+}
+
+// ECMAScript 2015 Classes
+setMinimumBrowserVersions({
```
hwillson: We're exporting `setMinimumBrowserVersions` and allowing it to be called multiple times as needed which is great, but once all of the `setMinimumBrowserVersions` calls have been completed and we have the new definitive `minimumVersions` ready to go, what happens if at some point we notice the `minimumVersions` aren't quite what we expected? How do we track down what call to `setMinimumBrowserVersions` is impacting the minimum versions used, especially since `setMinimumBrowserVersions` is part of the public API and can be called by 3rd party packages? The current values added via `setMinimumBrowserVersions` don't have a feature identifier of any kind associated with them, so we don't have an easy programmatic way to track down what feature is causing the minimum versions to be set at a certain level. We have to either look at the source, track down inline comments (like `ECMAScript 2015 Generator Functions`, `ECMAScript 2015 Template Literals`), or rely on other forms of documentation. If the `setMinimumBrowserVersions` are spread across multiple packages, this could be a pain. I'm  wondering if we could somehow mark a certain minimum browser version set with an identifier, and track that identifier when setting the `minimumVersions` baseline. So instead of

```js
// ECMAScript 2015 Generator Functions
setMinimumBrowserVersions({
  chrome: 39,
  edge: 13,
  firefox: 26,
  mobile_safari: 10,
  opera: 26,
  safari: 10,
  // Disallow any version of PhantomJS.
  phantomjs: Infinity,
});
```

we have something like

```js
setMinimumBrowserVersions('ECMAScript 2015 Generator Functions', {
  chrome: 39,
  edge: 13,
  firefox: 26,
  mobile_safari: 10,
  opera: 26,
  safari: 10,
  // Disallow any version of PhantomJS.
  phantomjs: Infinity,
});
```

then when deciding on the `minimumVerions`, we track the identifier for the version that ended up taking priority and being set. So anyone with programmatic access to `minimumVersions` could dump it out and see something like:

```js
{
  chrome: {
    decidedBy: 'ECMAScript 2015 Classes',
    version: 49,
  },
  edge: {
    decidedBy: 'ECMAScript 2015 Generator Functions',
    version: 13,
  },
  firefox: {
    decidedBy: 'ECMAScript 2015 Classes',
    version: 45,
  },
  ...
}
```

Then again this might be over engineering ...

packages/modern-browsers/TODO.md
mitar: How hard it would be to allow `web.worker.<tag>`?

benjamn: The main difficulty I see is that any package that adds something to `client` currently would implicitly be adding that resource to `web.worker` as well, since `client` essentially means `web.*`, which would include `web.worker` and/or `web.worker.*`.

That's a problem both because it would make the worker bundles larger and because existing package code might use global browser APIs that aren't available in web workers.

We could call it something different, like `webworker.<tag>`, so it wouldn't inherit all the `web` stuff, but you probably want to have at least some of your packages available in a worker.

Maybe packages have to opt into supporting workers with an `isWorkerSafe: true` option in `Package.describe`?

We also need worker-specific entry points in the application.

Now that I'm thinking about it, there are a lot of unanswered questions here.

mitar: I would go with `worker.<tag>`. Simple and backwards compatible. Those new packages which want to add things to worker, can do that.

BTW, then we should also add `worker` directory besides `client` and `server` in apps? Or will this work just for packages?

mitar: One solution is also that if I do `api.use('package', 'worker.web')` and `package` has only `web` available, then it is still pulled in. So if package was meant to be used in the browser but has not been made available for workers, you could pull it in.

Also, is really a bundle still a thing with dynamic imports?

packages/modern-browsers/modern.js
benjamn: I decided to do this by examining the call stack rather than passing in a string: 93c263225250c4431c4b0cb82b509b3eb07bcb0c. What do you think?

packages/modern-browsers/modern.js
```diff
@@ -0,0 +1,123 @@
+const minimumVersions = Object.create(null);
```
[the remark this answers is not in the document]
benjamn: I can totally imagine this example happening, but I think I'd like to wait until people ask for this feature. I suspect browser caching will make the difference between modern and legacy bundles disappear somewhat, so hopefully no one will complain? 🤞

packages/modern-browsers/modern.js
hwillson: Even better idea @benjamn - this looks great!

packages/modern-browsers/modern.js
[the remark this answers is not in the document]
hwillson: Sounds good - exposing less of the internals is always a good thing as well, so waiting to see what happens sounds like a plan. Thanks!

packages/babel-compiler/babel-compiler.js
```diff
@@ -79,7 +79,8 @@ BCp.processOneFileForTarget = function (inputFile, source) {
       extraFeatures.nodeMajorVersion = parseInt(process.versions.node);
     }
 
-    if (arch !== "web.browser.legacy") {
+    if (arch !== "web.browser.legacy" &&
+        ! hasOwn.call(extraFeatures, "runtime")) {
       extraFeatures.runtime = {
```
benjamn: Specifically, [this line](https://github.com/meteor/meteor/blob/a5590259288083f6dc4dec4d42cca3dfb62af4fc/packages/non-core/coffeescript-compiler/coffeescript-compiler.js#L25) from `coffeescript-compiler` was being overridden here, before this change.

benjamn:
Hot take incoming: the line between "modern" and "legacy" will essentially boil down to whether the client has native support for `async` functions. I'm eager to stop compiling generator functions with Regenerator, and none of the transforms that compile `async` functions to native generator functions are much better than the Regenerator experience (source maps are still wonky within the `async` function, so it's obvious you're not working with native code). Native `async` functions are supported by 73% of desktop and mobile browsers worldwide (81% in the US), and that percentage is only going to keep increasing. I see no reason not to embrace this future now.

packages/socket-stream-client/browser.js
```diff
@@ -158,38 +158,53 @@ export class ClientStream extends StreamClientCommon {
...
+
     this.socket.onclose = () => {
-      this._lostConnection();
+      Promise.resolve(
+        // If the socket is closing because there was an error, and we
+        // didn't load SockJS before, try loading it dynamically before
+        // retrying the connection.
+        lastError && ! hasSockJS &&
+        import("./sockjs-0.3.4.js")
+      ).done(() => {
```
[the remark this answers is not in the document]
benjamn: Yep, I made sure it was there when I refactored the `promise` package to use the legacy system.

benjamn:
@GeoffreyBooth Sorry for re-notifying you with that question, though I very much appreciate your response. I recently rebased and force-pushed this branch, which I think is why you got @-mentioned again. I'm open to making `coffeescript` imply `modules` at some point, but we haven't been forced to do it yet, so I think we can leave it optional for now.

packages/modern-browsers/TODO.md
trusktr: > Also, is really a bundle still a thing with dynamic imports?

 If you use `import()`, that just makes multiple bundles, one loaded up front, and some loaded later, right?

GeoffreyBooth:
@benjamn I think this is already happening, but as part of testing this it would be nice if packages that use `BabelCompiler` downstream also benefit from this PR’s bifurcated output. [CoffeeScript calls `new BabelCompiler`](https://github.com/meteor/meteor/blob/devel/packages/non-core/coffeescript-compiler/coffeescript-compiler.js) to do its transpilation pass, and I assume other packages like TypeScript and `.vue` compilation etc. do as well; so if the two evergreen/nevergreen outputs are both generated via that flow, that would be ideal. Testing it would be as simple as checking out todos-coffeescript and seeing that it creates both `web.browser` and `web.browser.legacy` outputs, and that they’re what you’d expect.

Issue #9244, opened 2017-10-18
Allow overruling constraint resolver


GeoffreyBooth:
See https://github.com/meteor/meteor/pull/9107#issuecomment-336779144. The package `practicalmeteor:mocha`‘s `package.js` includes `api.versionsFrom("1.3")`, which means it wants `coffeescript@1.0.4`. I’m trying to upgrade `coffeescript` to `coffeescript@2`, but when I do, `practicalmeteor:mocha` prevents the app from starting because `coffeescript@2` doesn’t satisfy its version requirement.

Meteor shouldn’t require that there be only one installed version of the `coffeescript` package per app, that satisfies both the app code and all packages. Why can’t package dependencies function the way they do in Node, where if the app itself has `lib@2` in its package.json and a dependency has `lib@1` in _its_ package.json, both `lib` versions get loaded respectively?

Sooner or later a core package like `ecmascript` will get upgraded to a new major version that differs from the major version in `api.versionsFrom("1.3")`, and then `practicalmeteor:mocha` will be unusable in that version of Meteor unless it gets patched. Seeing as there are many abandoned packages out there, this will be an increasing problem. I don’t want to choose between either not upgrading Meteor or needing to abandon or patch old packages that my app is using.

benjamn:
I hear your frustration, but this is not a bug. Meteor's package system is designed to be single-loading, so you can never have two different versions of the same package loaded side-by-side in the same app. This is particularly useful for packages that include a lot of client-side code, packages that expect to be singletons, and packages that run code during application startup (since running that code twice may be undesirable).

The possibility of sometimes ending up with distinct copies of packages using npm is a major problem for that ecosystem. For example, React stops working if you load it twice, `instanceof` fails for instances of duplicated constructors, and the widespread duplication of heavily-depended-upon packages like `lodash` is something people constantly struggle with. So that's why Meteor packages don't function the way they do in Node.

As we saw with https://github.com/meteor/meteor/pull/9107#issuecomment-337091582, major version updates are major for a reason: they introduce breaking changes that require updating packages like `practicalmeteor:mocha`. That's a natural part of working with independent packages, and it's why we have semantic versioning in the first place. Allowing `practicalmeteor:mocha` to continue using a different version of `coffeescript` is not a safe or sustainable solution to that annoyance.

It's unfortunate when packages are abandoned, but it's relatively easy to fork them locally, submit PRs, and/or republish them under a different developer name. That kind of work is worthwhile and unavoidable.

[...]

PR #9440, opened 2017-12-05
Update the babel-compiler package use to Babel 7, like the command-line tool.

benjamn:
Since Meteor 1.6, we've been quietly beta-testing Babel 7 by using it to compile [the `meteor` command-line tool codebase](https://github.com/meteor/meteor/tree/devel/tools). This [dog-fooding](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) was vital when we first introduced Babel compilation in Meteor 1.2, and it later helped us gain confidence in the upgrade from Babel 5 to Babel 6. Now it's time to make the leap to Babel 7!

There are a number of important changes hidden behind the abstraction of [`meteor-babel`](https://github.com/meteor/babel/commits/master), but the most disruptive change that will be visible to application developers is that the [`babel-runtime`](https://www.npmjs.com/package/babel-runtime) package is now called [`@babel/runtime`](https://www.npmjs.com/package/@babel/runtime). Since this package is typically installed in the application `node_modules` directory, developers will likely need to run `meteor npm install @babel/runtime` and/or edit their `package.json` files.

Probably the next most disruptive change is that the Meteor `babel-compiler` package has jumped a major version, from 6.24.7 to 7.0.0, which may require packages that depend directly on `babel-compiler` to adjust their version constraints.

There are a few other notable changes, but I think it will be easier if I comment on the commits themselves, below.

### DIFF
--- a/packages/ecmascript/transpilation-tests.js
+++ b/packages/ecmascript/transpilation-tests.js
@@ -52,8 +52,12 @@ class Foo {
   }
 }`);
 
-  // Babel 7 no longer uses classCallCheck in loose mode.
-  test.isTrue(contains(output, 'helpers/classCallCheck'));
+  // Babel 7 no longer imports the classCallCheck helper in loose mode.
+  test.equal(output, [
+    "var Foo = function Foo(x) {",
+    "  this.x = x;",
+    "};"
+  ].join("\n"));
 });
 
 Tinytest.add("ecmascript - transpilation - helpers - inherits", (test) => {
@@ -62,7 +66,7 @@ class Foo {}
 class Bar extends Foo {}
 `);
 
-  test.isTrue(contains(output, 'helpers/inherits'));
+  test.isTrue(contains(output, 'helpers/builtin/inherits'));
 });
 
 Tinytest.add("ecmascript - transpilation - helpers - bind", (test) => {
@@ -78,7 +82,7 @@ Tinytest.add("ecmascript - transpilation - helpers - extends", (test) => {
   var full = {a:1, ...middle, d:4};
 `);
 
-  test.isTrue(contains(output, 'helpers/extends'));
+  test.isTrue(contains(output, 'helpers/builtin/extends'));
 });
 
 Tinytest.add("ecmascript - transpilation - helpers - objectWithoutProperties", (test) => {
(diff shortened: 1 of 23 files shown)


packages/non-core/coffeescript-compiler/package.js
```diff
@@ -22,7 +22,7 @@ Npm.depends({
 });
 
 Package.onUse(function (api) {
-  api.use('babel-compiler@6.19.4');
+  api.use('babel-compiler@6.19.4||7.0.0');
```
benjamn: @GeoffreyBooth Good news: this syntax actually works!

packages/ecmascript/transpilation-tests.js
```diff
@@ -62,7 +66,7 @@ class Foo {}
 class Bar extends Foo {}
 `);
 
-  test.isTrue(contains(output, 'helpers/inherits'));
+  test.isTrue(contains(output, 'helpers/builtin/inherits'));
```
benjamn: This change was necessary because of https://github.com/meteor/babel/commit/242538dea13dd4600e2ba2fc1d3bd7d242e35304, which makes [`@babel/plugin-tranform-runtime`](https://www.npmjs.com/package/@babel/plugin-transform-runtime) generate imports for `@babel/runtime/helpers/builtin/whatever` instead of `@babel/runtime/helpers/whatever`. The `helpers/builtin/*` helpers are important because they do not import any polyfills from `core-js`, and instead simply assume global APIs like `Symbol` are natively supported or have been properly polyfilled. Since Meteor takes care to provide appropriate polyfills, especially with #9439, it's important to let other parts of the framework import `core-js` polyfills, rather than leaving that job to a generic tool like Babel.

packages/tinytest/tinytest.js
```diff
@@ -70,8 +70,9 @@ export class TestCaseResults {
         var frame = stack[i];
         // Heuristic: use the OUTERMOST line which is in a :tests.js
         // file (this is less likely to be a test helper function).
-        if (frame.getFileName().match(/:tests\.js/)) {
-          doc.filename = frame.getFileName();
+        const fileName = frame.getFileName();
+        if (fileName && fileName.match(/:tests\.js/)) {
+          doc.filename = fileName;
```
benjamn: I am not sure why this change became necessary after upgrading Babel, but somehow `frame.getFileName()` started returning `null` in some cases, so I decided to make this code more defensive.

packages/non-core/coffeescript-compiler/package.js
GeoffreyBooth: This is something, at least. It lets me as a package developer write something like `api.use('coffeescript@1||2');` (I assume that works, and not just exact `1.12.7||2.0.3`?)

It doesn’t really solve the issue of #9244, though, because to take advantage of this syntax you still need to fork a package. There still needs to be some way for end users to override the dependencies a package developer has set, if they haven’t thought to use this syntax or they picked versions that don’t work with other packages the user might want to install. But this is progress.

benjamn: Totally agree!

PR #9442, opened 2017-12-05
Help browser account saving with accounts-ui login/signup forms

### LINKED ISSUE #1746: Accounts UI does not trigger password store/autocomplete in Firefox
**_1 Upvote**_ I just now noticed that Accounts UI does trigger password store/autocomplete in Firefox. As using password managers is a good security practice, I think this should be addressed somehow.

Tested on Firefox 26. Should be probably tested on other browsers as well.

hwillson:
`accounts-ui-unstyled` currently uses `<div />`'s to hold its login/signup forms, as well as `<div />`'s to represent the login/signup buttons in the form. By not using proper `<form />` and `<button />` elements, certain browser's do not notice incoming login/signup requests, and therefore do not trigger their built in "would you like to save your user/password" functionality. This commit adjusts the `accounts-ui-unstyled` login/signup form to use proper `<form />` and `<button />` elements, allowing most browsers (Chrome, Firefox, IE - Safari will recognize the request when a user attempts to leave the page) to recognize incoming login/signup requests.

Fixes #1746.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,20 @@
 ## v.NEXT
 
+* The `accounts-ui-unstyled` package has been updated to use `<form />` and
+  `<button />` tags with its login/signup form, instead of `<div />`'s. This
+  change helps browser's notice login/signup requests, allowing them to
+  trigger their "remember your login/password" functionality.
+
+  > **Note:** If your application is styling the login/signup form using a CSS
+    path that includes the replaced `div` elements (e.g.
+    `div.login-form { ...` or `div.login-button { ...`), your styles will
+    break. You can either update your CSS to use `form.` / `button.` or
+    adjust your CSS specificity by styling on `class` / `id` attributes
+    only.
+
+  [Issue #1746](https://github.com/meteor/meteor/issues/1746)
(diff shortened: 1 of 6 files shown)


hwillson:
@benjamn Normally I would consider this a package patch, but I just thought of something. Since we're changing some of the HTML elements used by the login/signup form, if anyone is styling based on explicit HTML element names instead of using the `class`/`id` selectors, this could be a backwards incompatible change. So if someone is adding custom styles like `div.login-form { ... }` for example, their styles will break now that `div` has been replaced with `form`. In the name of progress however, we should still consider moving ahead with these changes, but I'll add a `History.md` entry explaining the potential backwards compatibility issues. Because of this, maybe we should tie these changes to release 1.6.1 (giving people a better chance of seeing this potential issue in the change log)?

abernix:
That's a good point @hwillson. It sounds reasonable to me to tie this to 1.6.1 — which means it would earn a minor version bump, in addition to a minor version bump on `accounts-ui` which depends on it, correct?

hwillson:
Minor versions bumped - thanks @abernix!

PR #9444, opened 2017-12-06
Ignore undefined fields when inserting/updating in Mongo

### LINKED ISSUE #6051: inconsistent minimongo insertion
minimongo has inconsistent behaviour with regards to undefined

depending on where the collection is inserted, {a: undefined} will become {} or {a: undefined} or {a:null}

```

if(Meteor.isClient){
  Test = new Mongo.Collection(null); // local client collection
  Test.insert({_id: 'client-side-local', a: undefined});
  var R = Test.findOne('client-side-local');
  console.log(R);  //  returns {_id: 'client-side-local', a: undefined}  // key 'a' is unchanged
}
Test2 = new Mongo.Collection('non-local'); // non-local collection
Test2.insert({_id: 'client'+Meteor.isClient, a: undefined});  // insert in both client and server side

var R1 = Test2.findOne('clienttrue');
console.log(R1);  //  returns {_id: 'clienttrue'}  // key 'a' is deleted due to stringify()
var R2 = Test2.findOne('clientfalse');
console.log(R2);  // returns {_id: 'clientfalse', a: null}  // key 'a' value undefined is changed to null
```

shouldn't the local collection consistently delete keys with undefined as value just like how the object will be stringify when sent over the wire

local collection already replace File objects into empty object, just like what one expect from stringify, but not for undefined
e.g.

```
//Client-side
TTT = new Mongo.Collection(null);
var obj = {
  file: new File([1, 2, 3, 4, 5], "name of file.xyz", {
    type: 'text/xyz'
  }),
  undef: undefined,
}
TTT.insert(obj);
TTT.findOne()  // returns {file: {}, undef: undefined};
nobj = EJSON.stringify(obj)  // returns "{"file":{}}"
EJSON.parse(nobj)  // returns {file:{}}
```

hwillson:
Meteor's current behaviour when handling Mongo inserts/updates with `undefined` fields differs when coming from the client versus the server:

**Client:**

```js
widgetCollection.insert({ name: undefined });
```

leads to the following in Mongo

```js
{ id: '...' }
```

**Server:**

```js
widgetCollection.insert({ name: undefined });
```

leads to the following in Mongo

```js
{ id: '...', name: null }
```

As discussed in issue #6051, the approach taken for queries initiated from the client or server should be consistent. 

This is a long standing issue that is caused by a few things:

- `undefined` is not a valid JSON/EJSON type so when client side Mongo inserts/updates are sent to the server for handling, the `undefined` fields are stripped.
- Mongo's own [BSON spec](http://bsonspec.org/spec.html) has flagged `undefined` as deprecated, which means the use of `undefined` within the Mongo landscape is not recommended. This has lead to some difficulties when using `undefined` in Mongo queries as how `undefined` should be handled has been interpreted in different ways by different Mongo client authors. By default the official Mongo Node client that Meteor uses converts `undefined` to `null` when inserting/updating.

There has been a lot of discussion on the interwebs around how the Mongo Node driver handles `undefined` fields, and the problems that can arise from just converting `undefined` to `null`. To that end, the Mongo Node driver team added support for a [`ignoreUndefined` connection option](http://mongodb.github.io/node-mongodb-native/2.2/api/MongoClient.html#connect) a while back (which is `false` by default for backwards compatibility reasons). When `ignoreUndefined` is `true`, fields with `undefined` values are dropped when inserting/updating, instead of being converted to `null`.

This PR sets the `ignoreUndefined` option to true for all Mongo connections, which makes the handling of client/server initiated Mongo inserts/updates consistent. 

This could potentially be a backwards incompatible change if an app is say expecting an `undefined` field to be inserted as a `null` field. This change does better align with Mongo's recommendation of not using `undefined` however, so although some backwards compatibility might be sacrificed, that sacrifice is essential to better align with Mongo. Mentioning this in the `History.md` should hopefully be enough.

Fixes #6051.


romainkoenig:
> This could potentially be a backwards incompatible change if an app is say expecting an undefined field to be inserted as a null field. This change does better align with Mongo's recommendation of not using undefined however, so although some backwards compatibility might be sacrificed, that sacrifice is essential to better align with Mongo. Mentioning this in the History.md should hopefully be enough.

It wasn't, for such a breaking change, updating the patch version is not enough.
Then for a `findOne({ name: variable })` with a variable potentially undefined. It goes from : 
No data returned to the first document of the collection returned, or even more with find ?

Do you imagine with : find({ userId }) -> from no one to everyone :-(

It takes time to find such a bug, because it does not throw...
I don't understand, this is not a patch, this is a contract being modified

hwillson:
@romainkoenig Yes, this is a painful issue to resolve. It's a bug either way; either you end up with unexpected results when you have your `undefined` field stripped in a `find` and you weren't expecting it to be, or you end up having `null` stored from an `insert` in your collection when you weren't expecting it to be. It's a tricky situation, and to mitigate the risk, this PR should be followed up with an implementation that addresses the original intent of the sister PR https://github.com/meteor/meteor/pull/9671. There is a new issue open to track this work - see https://github.com/meteor/meteor/issues/9780. 

Just to sum up though - the changes made here were intended to help address an existing bug caused by Meteor's automatic assumption that `undefined` could be converted to `null` when sent over to Mongo. This assumption has lead to numerous other errors (and GitHub issues), so this PR was intended to help line Meteor up with what Mongo is expecting (which is not to use `undefined`). I definitely agree that there is more work to do here, which https://github.com/meteor/meteor/issues/9780 is tracking (and which is `pull-requests-encouraged` if you're interested).

PR #9445, opened 2017-12-06
Deprecate stylus

hwillson:
Better / more up to date 3rd party stylus packages exist and there isn't really any technical reason why Meteor core needs to include its own stylus package. Since a stylus package can be fully built and managed outside of core, this commit moves the `stylus` package into `deprecated`. 

This PR also preps the package contents for deprecation if we decide to publish a final version. Publishing a final (deprecated) version probably isn't necessary, but the changes are included just in-case.


hwillson:
Hmm - failing `webapp` and `appcache` tests that both pass locally, and don't have anything to do with these changes. Something is up somewhere, but it doesn't appear to be related to this PR.

hwillson:
Just to add more weight to the `stylus` deprecation decision: https://github.com/stylus/stylus/issues/2282#issuecomment-295374929 (a comment from the lead -  and only - upstream `stylus` maintainer saying he's unsure of it's future, doesn't have time to keep it going himself, asking for help, and not really getting any ... and that was in April 🙁)

hwillson:
I found the reason for the failures. The failing tests have an indirect dependency on at least one `.css` file being available in the tested bundle, when the console test runner runs. It turns out that a `.css` file (`/packages/local-test_stylus/stylus_tests.styl.css`) included in the `stylus` package was being picked up and used. By deprecating `stylus` there are no longer any `.css` files included in the app bundle when the console test runner runs, causing the tests to fail. Interestingly enough - and what made this so difficult to track down - is that when using the web based test runner (ala `meteor test-packages`) these tests all pass. This is because the web test runner includes additional `.css` files (for bootstrap and its own custom styles), which are found and used by tests ... 🤦‍♂️ 

Long story short - fixes are coming that involve adding a test `.css` file to the test bundle, and will make it very clear that the `.css` file is needed (and depended on) by the console test runner.

hwillson:
I've committed changes to `test-in-console` to fix the failing tests here, but those changes aren't really specific to this PR (they should be done regardless). Let me know if they should be reverted here and opened in a new PR. Thanks!

packages/deprecated/stylus/package.js
```diff
@@ -0,0 +1,8 @@
+Package.describe({
+  summary: 'DEPRECATED - Expressive, dynamic, robust CSS',
+  version: "2.513.14"
```
abernix: If we bump the minor to, say, `2.999.0`, it could prevent (some, post-1.4.x) users from automatically updating to the deprecated version (and instead, leave them on the prior 2.513.13) when running `meteor update --all-packages` (or `meteor update stylus`).  Then, since it wouldn't be "`core`" in future published Meteor versions, I believe they'd be allowed to keep using it `@2.513.13` after that.  Am I mistaken?

Otherwise LGTM.

hwillson: That makes sense to me @abernix - changes coming - thanks!

anault:
> Better / more up to date 3rd party stylus packages exist [...]

@hwillson Do you have any recommendations? [mquandalle:stylus](https://atmospherejs.com/mquandalle/stylus) and [stolinski:stylus-multi](https://atmospherejs.com/stolinski/stylus-multi) seem to be the most popular alternatives but they were last updated 2-3 years ago.

hwillson:
@anault Try [coagmano:stylus](https://atmospherejs.com/coagmano/stylus) (it's essentially an up to date version of `mquandalle:stylus`).

Issue #9447, opened 2017-12-07
Cordova iOS splash screen issue on iPhone X


skirunman:
There is a known issue with Cordova iOS displaying the splash screen incorrectly on iPhone X as documented [here](https://stackoverflow.com/questions/47467176/cordova-iphone-x-launch-image-initially-shows-white-area-at-bottom) and [here](https://issues.apache.org/jira/browse/CB-13396). Basically there is a white area below the splash screen when using iOS storyboard images. This issue is fixed in [Cordova iOS v4.5.4](https://cordova.apache.org/announcements/2017/11/21/ios-release.html) with this https://github.com/apache/cordova-ios/pull/345

The fix for this is to update our Cordova iOS dependency to `cordova-ios@4.5.4`. A PR will be forthcoming.

PR #9448, opened 2017-12-07
Update to Cordova iOS v4.5.4

skirunman:
Update to Cordova iOS v4.5.4 to fix iPhone X splash screen issue.
Fixes https://github.com/meteor/meteor/issues/9447


abernix:
@skirunman Does this bump change need for the commented block of iPad modes here: https://github.com/skirunman/meteor/blob/4ae3b821b914657eec93df02a10bac33feb00944/tools/cordova/builder.js#L58-L62

skirunman:
@abernix My understanding is that Apple/Xcode, and therefore `ios-cordova` is does not, and likely will not, support iPad Pro launch screens as static image sizes. Instead, you need to use  Xcode storyboard images. 

Therefore, I actually think we should remove those commented out iPad Pro sizes. I can do it in this PR if you want or just do it yourself.

PR #9451, opened 2017-12-08
Release 1.4.4.5

benjamn:
A small point release for these upcoming Node security updates: https://nodejs.org/en/blog/vulnerability/december-2017-security-releases/

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,9 @@
 ## v.NEXT
 
+## v1.4.4.5, 2017-12-08
+
+* Node has been upgraded to version 4.8.6.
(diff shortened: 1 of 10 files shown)


benjamn:
Since the Node version in question here is the same as for Meteor 1.5.4.1 (Node 4.8.7), running
```sh
meteor update --release 1.5.4.1
```
should be almost as easy as updating to 1.4.4.5, so I'm going to hold off actually publishing 1.4.4.5 until someone reports the can't update from 1.4.x to 1.5.4.1.

PR #9457, opened 2017-12-09
Change http link to https in README.md file

phurinaix:
-change two link
1. from "http://docs.meteor.com/#ddp_connect" to "https://docs.meteor.com/api/connections.html#DDP-connect"
2. from "http://docs.meteor.com/" to "https://docs.meteor.com/"


abernix:
The updated `README` won't get published to Atmosphere until `ddp` gets republished again with a version bump, but that can wait until the next functional change to `ddp`; it's not worth bumping the package version for just this.  Thanks, @phurinaix.

PR #9462, opened 2017-12-12
Updates information about Meteor CLI usage

zionts:
Removes some outdated references in the Meteor CLI help text
- Removes `meteor logs` reference as it no longer functions
- Changes comments about `meteor mongo` to only refer to local Mongo instances
- Removes references to "free hosting"


tools/cli/help.txt
```diff
@@ -440,26 +440,15 @@ commands can be accessed by pressing the up arrow.
 
 
 >>> mongo
-Connect to the Mongo database for the specified site.
-Usage: meteor mongo [--url]
```
hwillson: I think we should leave the `--url` / `-U` docs in place here. Using `meteor mongo --url` is incredibly handy when you want to get Meteor's current full mongo URL, to integrate that Mongo instance with something else locally (like when developing a custom app that feeds data into Mongo directly).

tools/cli/help.txt
```diff
@@ -474,11 +463,11 @@ Usage: meteor deploy <site> [--settings settings.json] [--debug] [--delete]
...
 Deploys the project in your current directory to Meteor's servers.
 
-You can deploy to any available name under 'meteor.com'
+You can deploy to any available name under 'meteorapp.com'
 without any additional configuration, for example,
-'myapp.meteor.com'. If you deploy to a custom domain, such as
+'myapp.meteorapp.com'. If you deploy to a custom domain, such as
 'myapp.mydomain.com', then you'll also need to configure your domain's
-DNS records. See the Meteor docs for details.
+DNS records. See the Meteor / Galaxy docs for details.
```
abernix: Could we add a link to those docs while we're at it?

tools/cli/help.txt
abernix: Agree, we should probably keep this unless there's a pressing reason to remove it.  Removing it could cause an issue if anyone was using this in an external script right now (i.e. `mongo $(meteor mongo --url)`).

tools/cli/help.txt
zionts: Good call 👍

tools/cli/help.txt
zionts: Also, again to clarify, this is something that I want to fix because I think it's incredibly useful. I just don't want inaccurate documentation as to what it currently does

tools/cli/help.txt
```diff
@@ -440,26 +440,15 @@ commands can be accessed by pressing the up arrow.
...
-If no site is specified, this is the current project's local development
-database. In this case, the current working directory must be a
-Meteor project directory, and the Meteor application must already be
-running.
-
-Instead of opening a shell, specifying --url (-U) will return a URL
-suitable for an external program to connect to the database. For remote
-databases on deployed applications, the URL is valid for one hour.
-
-Options:
```
zionts: Also adding this back in with the above comment

tools/cli/help.txt
abernix: Well, technically, the code is still in place to _try_ to connect to a remote Mongo if a `site` is passed to the `meteor mongo` URL.  That code can be found here:

https://github.com/meteor/meteor/blob/9b056f5e911513fb3df716bd524d504cceb863f4/tools/cli/commands.js#L1254-L1264

We should definitely address that!

tools/cli/help.txt
```diff
@@ -500,14 +496,6 @@ Options:
                   the current versions, if required to satisfy all package version
                   constraints.
 
-
->>> logs
```
abernix: We should also remove the associated code from `tools/cli/commands.js` (and wherever it branches out to!):

https://github.com/meteor/meteor/blob/9b056f5e911513fb3df716bd524d504cceb863f4/tools/cli/commands.js#L1431-L1444

zionts: The reason I figured not to do that was because we don't plan on just deleting this outright, do we? I thought we had vague plans to fix it and maybe that code would be helpful. Then again, it'll still be in git history. The code just makes a call to the galaxy-server logs endpoint, so I was thinking we could leave it in, undocumented and unsupported, and eventually fix that endpoint once Logger2 is pushed out.

zionts: Ehh I'm just gonna take it out – I think you're right. Plus the /logs endpoint doesn't even exist anymore

abernix: Yeah, we can add it back easily!  I don't believe the `meteor logs` command has worked for two years now, so it's not a breaking change to remove it and put it back later.

zionts:
This is ready for another look whenever you have a chance :)

tools/cli/help.txt
```diff
@@ -440,26 +440,22 @@ commands can be accessed by pressing the up arrow.
...
-
 Instead of opening a shell, specifying --url (-U) will return a URL
 suitable for an external program to connect to the database. For remote
 databases on deployed applications, the URL is valid for one hour.
 
 Options:
-  --url, -U  return a Mongo database URL
+  --url, -U return a Mongo database URL
 
+Currently, this feature is only for local developments and the Mongo shell
```
hwillson: `local developments` doesn't seem quite right in this context. Maybe re-word this to something like:

> Currently, this feature can only be used when developing locally. The opened Mongo shell connects to the current project's local development database. The current working directory must be a Meteor project directory, and the Meteor application must already be running.

hwillson:
Thanks @zionts - we should be all set here - LGTM!

PR #9463, opened 2017-12-12
Fix #9305 by including `default` when mirroring module exports.

### LINKED ISSUE #9305: Meteor 1.6 - import style (scss) as object from js file is not working
After updating to meteor 1.6 I noticed the following difference in behavior from v1.5.2.2.

I need to require style file and add it as inline tag in a component:
the code:
```
import style from '../common/print-page.scss';

class PrintPage extends React.Component {
  render() {
    return (
      <div>
         <style>{style.textContext}</style>
         .....
      </div>
    )
  }
}

```

importing the file at v1.5.2.2 returns:

`import style from '../style.scss'`
-> `style <style type=​"text/​css">​…​</style>​`

import file at v1.6 returns:
-> `style {Symbol(__esModule): true}`

and therefor the value of `style.textContent` is undefined.

- Is it a bug or intended change due to the change of node v8 or other aspects related to the modules and the build system?

- Is there any other way that I can include style from a scss file in a component?

benjamn:
Detailed explanation [here](https://github.com/meteor/meteor/issues/9305#issuecomment-351126362).

Underlying Reify functionality implemented [here](https://github.com/benjamn/reify/commit/10c90cd0a2bfc328964820799845ac3d5d69f493).

Note that this does not affect `export * from "./module"` (which still skips any `default` export), but only hand-written code of the form:
```js
module.watch(require("./module"), {
  "*": module.makeNsSetter(true)
});
```
The Reify compiler only generates code with `module.makeNsSetter()`, not `module.makeNsSetter(true)`, so this new functionality is backwards compatible with existing Reify-generated code.

### DIFF
--- a/packages/modules/.npm/package/npm-shrinkwrap.json
+++ b/packages/modules/.npm/package/npm-shrinkwrap.json
@@ -17,9 +17,9 @@
       "integrity": "sha512-sN4U9tIJtBRwKbwgFh9qJfrPIQ/GGTRr1MGqkgOeMTLy8/lM0FcWU//FqlnZ3Vb7gJ+Mxh3FOg1EklibdajbaQ=="
     },
     "reify": {
-      "version": "0.13.1",
-      "resolved": "https://registry.npmjs.org/reify/-/reify-0.13.1.tgz",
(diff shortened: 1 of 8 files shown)


hwillson:
Looks awesome @benjamn - thanks!

PR #9466, opened 2017-12-13
Make `meteor deploy` synchronous

zionts:
- This changes meteor deploy to poll the /version-status/ REST endpoint that Galaxy now exposes for the build and deploy status of the version being deployed
- Allows users to set a flag to change the polling timeout or to not wait at all
- Also fixes up some typos and outdated references

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1351,7 +1351,9 @@ main.registerCommand({
     // people to deploy from checkout or do other weird shit. We are not
     // responsible for the consequences.
     'override-architecture-with-local' : { type: Boolean },
-    'allow-incompatible-update': { type: Boolean }
+    'allow-incompatible-update': { type: Boolean },
+    'deploy-polling-timeout': { type: Number },
+    'no-wait': { type: Boolean },
   },
   allowUnrecognizedOptions: true,
   requiresApp: function (options) {
@@ -1412,12 +1414,21 @@ main.registerCommand({
     serverArch: buildArch
   };
 
+  let deployPollingTimeoutMs = null;
+  if (options['deploy-polling-timeout']) {
+    deployPollingTimeoutMs = options['deploy-polling-timeout'];
+  }
+
+  const waitForDeploy = !options['no-wait'];
+
   var deployResult = deploy.bundleAndDeploy({
     projectContext: projectContext,
     site: site,
     settingsFile: options.settings,
     buildOptions: buildOptions,
-    rawOptions
+    rawOptions,
+    deployPollingTimeoutMs,
+    waitForDeploy,
   });
 
   if (deployResult === 0) {
--- a/tools/cli/help.txt
+++ b/tools/cli/help.txt
@@ -465,7 +465,7 @@ Reset the current project to a fresh state. Removes all local data.
 
 
 >>> deploy
-Deploy this project to Meteor.
+Deploy this project to Galaxy, Meteor's hosting service.
 Usage: meteor deploy <site> [--settings settings.json] [--debug] [--delete]
 
 Deploys the project in your current directory to Meteor's servers.
@@ -495,6 +495,11 @@ Options:
                   downgraded to versions that are potentially incompatible with
                   the current versions, if required to satisfy all package version
                   constraints.
+  --deploy-polling-timeout  The number of milliseconds to wait for build/deploy
+                  success or failure after a successful upload of your app's
+                  minified code; defaults to 15 minutes.
+  --no-wait       Exits when Meteor has uploaded the app's code instead of
+                  waiting for the deploy to conclude.
 
 >>> authorized
 View or change authorized users and organizations for a site.
--- a/tools/meteor-services/deploy.js
+++ b/tools/meteor-services/deploy.js
@@ -51,7 +51,10 @@ const CAPABILITIES = ['showDeployMessages', 'canTransferAuthorization'];
 // - method: GET, POST, or DELETE. default GET
 // - operation: "info", "logs", "mongo", "deploy", "authorized-apps",
 //   "version-status"
-// - site: site name, version ID
+// - site: site name. Pass this even if the site isn't part of the URL
+//   so that Galaxy discovery works properly
+// - operand: the part of the URL after the operation. If not set,
+//   defaults to site
 // - expectPayload: an array of key names. if present, then we expect
 //   the server to return JSON content on success and to return an
 //   object with all of these key names.
@@ -96,11 +99,18 @@ function deployRpc(options) {
     Console.info("Talking to Galaxy servers at " + deployURLBase);
   }
 
+  let operand = '';
+  if (options.operand) {
+    operand = `/${options.operand}`;
+  } else if (options.site) {
+    operand = `/${options.site}`;
+  }
+
   // XXX: Reintroduce progress for upload
   try {
     var result = request(Object.assign(options, {
       url: deployURLBase + '/' + options.operation +
-        (options.site ? ('/' + options.site) : ''),
+        operand,
       method: options.method || 'GET',
       bodyStream: options.bodyStream,
       useAuthHeader: true,
@@ -337,13 +347,13 @@ function canonicalizeSite(site) {
 
 // Executes the poll to check for deployment success and outputs proper messages
 // to user about the status of their app during the polling process
-async function pollForDeploymentSuccess(versionId, deployPollTimeout, result) {
+async function pollForDeploymentSuccess(versionId, deployPollTimeout, result, site) {
   // Create a default polling configuration for polling for deploy / build
   // In the future, we may change this to be user-configurable or smart
   // The user can only currently configure the polling timeout via a flag
   const pollingState = new PollingState(deployPollTimeout);
   await sleepForMilliseconds(pollingState.initialWaitTimeMs);
-  const deploymentPollResult = await pollForDeploy(pollingState, versionId);
+  const deploymentPollResult = await pollForDeploy(pollingState, versionId, site);
   if (deploymentPollResult && deploymentPollResult.isActive) {
     return 0;
   }
@@ -384,7 +394,7 @@ class PollingState {
 // messages pertaining to the status of the version, which will then be reported
 // directly to the user. When the poll is complete, it will return an object
 // with information about the final state of the version and the app.
-async function pollForDeploy(pollingState, versionId, galaxyUrl) {
+async function pollForDeploy(pollingState, versionId, site) {
   const {
     deadline,
     initialWaitTimeMs,
@@ -397,7 +407,8 @@ async function pollForDeploy(pollingState, versionId, galaxyUrl) {
   const versionStatusResult = deployRpc({
     method: 'GET',
     operation: 'version-status',
-    site: versionId,
+    site,
+    operand: versionId,
     expectPayload: ['message', 'finishStatus'],
     printDeployURL: false,
   });
@@ -423,7 +434,7 @@ async function pollForDeploy(pollingState, versionId, galaxyUrl) {
       return 1;
     } else if (new Date() < deadline) {
       await sleepForMilliseconds(pollIntervalMs);
-      return await pollForDeploy(pollingState, versionId);
+      return await pollForDeploy(pollingState, versionId, site);
     }
   }
 
@@ -432,7 +443,7 @@ async function pollForDeploy(pollingState, versionId, galaxyUrl) {
   if(new Date() < deadline && !finishStatus.isFinished) {
     // Wait for a set interval and then poll again
     await sleepForMilliseconds(pollIntervalMs);
-    return await pollForDeploy(pollingState, versionId);
+    return await pollForDeploy(pollingState, versionId, site);
   } else if (!finishStatus.isFinished) {
     Console.info(`Polling timed out. To check the status of your app, visit
     ${versionStatusResult.payload.galaxyUrl}. To wait longer, pass a timeout
@@ -581,7 +592,8 @@ export async function bundleAndDeploy(options) {
     return await pollForDeploymentSuccess(
       result.payload.newVersionId,
       options.deployPollingTimeoutMs,
-      result);
+      result,
+      site);
   }
   return 0;
 };
--- a/tools/meteor-services/deploy.js
+++ b/tools/meteor-services/deploy.js
@@ -22,6 +22,12 @@ import {
 import { recordPackages } from './stats.js';
 import { Console } from '../console/console.js';
 
+function sleepForMilliseconds(millisecondsToWait) {
+  return new Promise(function(resolve) {
+    let time = setTimeout(() => resolve(null), millisecondsToWait)
+  });
+}
+
 const hasOwn = Object.prototype.hasOwnProperty;
 
 const CAPABILITIES = ['showDeployMessages', 'canTransferAuthorization'];
@@ -43,8 +49,9 @@ const CAPABILITIES = ['showDeployMessages', 'canTransferAuthorization'];
 //
 // Options include:
 // - method: GET, POST, or DELETE. default GET
-// - operation: "info", "mongo", "deploy", "authorized-apps"
-// - site: site name
+// - operation: "info", "logs", "mongo", "deploy", "authorized-apps",
+//   "version-status"
+// - site: site name, version ID
 // - expectPayload: an array of key names. if present, then we expect
 //   the server to return JSON content on success and to return an
 //   object with all of these key names.
@@ -75,7 +82,13 @@ function deployRpc(options) {
   if (options.headers.cookie) {
     throw new Error("sorry, can't combine cookie headers yet");
   }
-  options.qs = Object.assign({}, options.qs, {capabilities: CAPABILITIES});
+  options.qs = Object.assign({}, options.qs,
+    {capabilities: CAPABILITIES.slice()});
+  // If we are waiting for deploy, we let Galaxy know so it can
+  // use that information to send us the right deploy message response.
+  if (options.waitForDeploy) {
+    options.qs.capabilities.push('willPollVersionStatus');
+  }
 
   const deployURLBase = getDeployURL(options.site).await();
 
@@ -174,7 +187,8 @@ function authedRpc(options) {
     site: rpcOptions.site,
     expectPayload: [],
     qs: options.qs,
-    printDeployURL: options.printDeployURL
+    printDeployURL: options.printDeployURL,
+    waitForDeploy: options.waitForDeploy,
   });
   delete rpcOptions.printDeployURL;
 
@@ -321,6 +335,113 @@ function canonicalizeSite(site) {
   return parsed.hostname;
 };
 
+// Executes the poll to check for deployment success and outputs proper messages
+// to user about the status of their app during the polling process
+async function pollForDeploymentSuccess(versionId, deployPollTimeout, result) {
+  // Create a default polling configuration for polling for deploy / build
+  // In the future, we may change this to be user-configurable or smart
+  // The user can only currently configure the polling timeout via a flag
+  const pollingState = new PollingState(deployPollTimeout);
+  await sleepForMilliseconds(pollingState.initialWaitTimeMs);
+  const deploymentPollResult = await pollForDeploy(pollingState, versionId);
+  if (deploymentPollResult && deploymentPollResult.isActive) {
+    return 0;
+  }
+  return 1;
+}
+
+// Creates a polling configuration with defaults if fields left unset
+// Right now we only use the default unless timeout is specified
+// We envision potentially creating this configuration object in a programmatic
+// way or via user-specification in the future.
+// Default initialWaitTime is 10 seconds – this is the time to wait before checking at all
+// Default pollInterval is 700 milliseconds – this is the wait interval between polls
+// Default timeout is 15 minutes
+// `start` tracks the time when we started polling
+// `currentMessage` tracks what the current status message is for this version
+class PollingState {
+  constructor(timeoutMs,
+    initialWaitTimeMs,
+    pollIntervalMs,
+    maxErrors) {
+      const FIFTEEN_MINUTES_MS = 15*60*1000;
+      const MAX_ERRORS = 5;
+      this.initialWaitTimeMs = initialWaitTimeMs || 10*1000;
+      this.pollIntervalMs = pollIntervalMs || 700;
+      this.deadline = timeoutMs ? new Date(new Date().getTime() + timeoutMs) :
+        new Date(new Date().getTime() + FIFTEEN_MINUTES_MS);
+      this.start = new Date();
+      this.currentMessage = '';
+      this.errors = 0;
+      this.maxErrors = maxErrors || MAX_ERRORS;
+  }
+}
+
+// Poll the "version-status" endpoint for the build and deploy status
+// of a specified version ID with a polling configuration.
+// This will only end successfully when the polling endpoint reports that
+// the version deployment is finished. The version-status endpoints will report
+// messages pertaining to the status of the version, which will then be reported
+// directly to the user. When the poll is complete, it will return an object
+// with information about the final state of the version and the app.
+async function pollForDeploy(pollingState, versionId, galaxyUrl) {
+  const {
+    deadline,
+    initialWaitTimeMs,
+    pollIntervalMs,
+    start,
+    currentMessage,
+  } = pollingState;
+
+  // Do a call to the version-status endpoint for the specified versionId
+  const versionStatusResult = deployRpc({
+    method: 'GET',
+    operation: 'version-status',
+    site: versionId,
+    expectPayload: ['message', 'finishStatus'],
+    printDeployURL: false,
+  });
+
+  // Check the details of the Version Status response and compare message to last call
+  if (versionStatusResult &&
+    versionStatusResult.payload &&
+    versionStatusResult.payload.message) {
+      const message = versionStatusResult.payload.message;
+      if (currentMessage !== message) {
+        Console.info(message);
+        pollingState.currentMessage = message;
+      }
+  } else {
+    // If we did not get a valid Version Status response, just fail silently and
+    // keep polling as per usual – this may have just been a whiff from Galaxy.
+    // We do the retry here because we might hit an error if we try to parse the
+    // result of the version-status call below.
+    Console.warn(versionStatusResult.errorMessage || 'Unexpected error from Galaxy');
+    pollingState.errors++;
+    if (pollingState.errors >= pollingState.maxErrors) {
+      Console.error(versionStatusResult.errorMessage);
+      return 1;
+    } else if (new Date() < deadline) {
+      await sleepForMilliseconds(pollIntervalMs);
+      return await pollForDeploy(pollingState, versionId);
+    }
+  }
+
+  const finishStatus = versionStatusResult.payload.finishStatus;
+  // Poll again if version isn't finished and we haven't exceeded the timeout
+  if(new Date() < deadline && !finishStatus.isFinished) {
+    // Wait for a set interval and then poll again
+    await sleepForMilliseconds(pollIntervalMs);
+    return await pollForDeploy(pollingState, versionId);
+  } else if (!finishStatus.isFinished) {
+    Console.info(`Polling timed out. To check the status of your app, visit
+    ${versionStatusResult.payload.galaxyUrl}. To wait longer, pass a timeout
+    in milliseconds to the '--deploy-polling-timeout' option of 'meteor deploy'.`);
+  }
+  return finishStatus;
+}
+
+
 // Run the bundler and deploy the result. Print progress
 // messages. Return a command exit code.
 //
@@ -334,7 +455,10 @@ function canonicalizeSite(site) {
 //   stats server.
 // - buildOptions: the 'buildOptions' argument to the bundler
 // - rawOptions: any unknown options that were passed to the command line tool
-export function bundleAndDeploy(options) {
+// - waitForDeploy: whether to poll Galaxy after upload for deploy status
+// - deployPollingTimeoutMs: user overridden timeout for polling Galaxy
+//   for deploy status
+export async function bundleAndDeploy(options) {
   if (options.recordPackageUsage === undefined) {
     options.recordPackageUsage = true;
   }
@@ -383,7 +507,7 @@ export function bundleAndDeploy(options) {
   var buildDir = mkdtemp('build_tar');
   var bundlePath = pathJoin(buildDir, 'bundle');
 
-  Console.info('Deploying your app...');
+  Console.info('Preparing to deploy your app...');
 
   var settings = null;
   var messages = buildmessage.capture({
@@ -434,6 +558,7 @@ export function bundleAndDeploy(options) {
       preflightPassword: preflight.preflightPassword,
       // Disable the HTTP timeout for this POST request.
       timeout: null,
+      waitForDeploy: options.waitForDeploy,
     });
   });
 
@@ -442,33 +567,22 @@ export function bundleAndDeploy(options) {
     return 1;
   }
 
+  // This will allow Galaxy to report messages to users ad-hoc
+  // Also if we are using the --no-wait flag, this will contain the message
+  // that Galaxy used to send after upload success.
   if (result.payload.message) {
     Console.info(result.payload.message);
-  } else {
-    var deployedAt = require('url').parse(result.payload.url);
-    var hostname = deployedAt.hostname;
-
-    Console.info('Now serving at http://' + hostname);
-
-    if (! hostname.match(/meteor\.com$/)) {
-      var dns = require('dns');
-      dns.resolve(hostname, 'CNAME', function (err, cnames) {
-        if (err || cnames[0] !== 'origin.meteor.com') {
-          dns.resolve(hostname, 'A', function (err, addresses) {
-            if (err || addresses[0] !== '107.22.210.133') {
-              Console.info('-------------');
-              Console.info(
-                "You've deployed to a custom domain.",
-                "Please be sure to CNAME your hostname",
-                "to origin.meteor.com, or set an A record to 107.22.210.133.");
-              Console.info('-------------');
-            }
-          });
-        }
-      });
-    }
   }
 
+  // After an upload succeeds, we want to poll Galaxy to see if the
+  // build / deploy succeed. We indicate that Meteor should poll for version
+  // status by including a newVersionId in the payload.
+  if (options.waitForDeploy && result.payload.newVersionId) {
+    return await pollForDeploymentSuccess(
+      result.payload.newVersionId,
+      options.deployPollingTimeoutMs,
+      result);
+  }
   return 0;
 };
 
--- a/tools/tests/galaxy.js
+++ b/tools/tests/galaxy.js
@@ -99,7 +99,7 @@ selftest.define('galaxy deploy - settings', ['galaxy'], function () {
   // Test that the app is actually running on Galaxy.
   checkAppIsRunning(appName, { text: "Hello" });
 
-  // Test that the public settins appear in the HTTP response body.
+  // Test that the public settings appear in the HTTP response body.
   if (! galaxyUtils.ignoreHttpChecks()) {
     testUtils.checkForSettings(appName, settings, 10);
   }


tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+            terminal = true;
+          }
+        } else if(deployStatus !== 'deploying') {
+          Console.info(`Deploy status changed to nonterminal state: ${deployStatus}`)
+        }
+      } else if(currentStatus.isActive !== isActive) {
+        currentStatus.isActive = isActive;
+        // isActive was initialized to false, so it can only be true
+        Console.info('Your deployed version is now active.');
+        terminal = true;
```
abernix: Rather than flopping `terminal` to `true` here and on lines above (e.g. 470, 488), in order to avoid the conditional on line 506, could we just `return` in these locations and get rid of the `terminal` variable?  If necessary, the `return` status could dictate whether additional recursion is necessary.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+          }
+        } else if (buildStatus !== 'building') {
+          // This will report if the buildStatus changes to a non-terminal, unexpected state
+          //    e.g. failed-temporarily
+          Console.info(`Build state changed to nonterminal state: ${buildStatus}`)
+        }
+      } else if(currentStatus.deployStatus !== deployStatus) {
+        // Deploy status is treated similarly to buildStatus above
+        currentStatus.deployStatus = deployStatus;
+        if (terminalDeployStatuses.indexOf(deployStatus) >= 0) {
```
abernix: Suitable to use [`Array.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here, rather than `.indexOf(...)>= 0`.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+          } else {
+            // Any other terminal status is failure, so report on it
+            Console.info(`Your app failed to build, with buildStatus: ${buildStatus}`);
+            // We also want to stop polling, since deploy cannot succeed with a failed build
+            terminal = true;
+          }
+        } else if (buildStatus !== 'building') {
+          // This will report if the buildStatus changes to a non-terminal, unexpected state
+          //    e.g. failed-temporarily
+          Console.info(`Build state changed to nonterminal state: ${buildStatus}`)
```
abernix: I think exposing this lexicon to the user is probably not going to be as clear as we want it to be, especially since "terminal" might be confused with the terminal (a.k.a. shell, tty, console).  In fact, I think the _termin_-ology might be a bit too formal grammar, and possibly inappropriate, rather than just calling it `finished`.

abernix: See also: my later suggestion about getting rid of the functional purpose of the `terminal` boolean entirely. 😄

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+        }
+      } else if(currentStatus.deployStatus !== deployStatus) {
+        // Deploy status is treated similarly to buildStatus above
+        currentStatus.deployStatus = deployStatus;
+        if (terminalDeployStatuses.indexOf(deployStatus) >= 0) {
+          if(deployStatus == 'success') {
```
abernix: Also, preference to use `===`, almost always, even if it might not make a difference in this particular case.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    // Timeouts must be done within Promises, so we set a dummy Promise to
+    // wait before initiating polling
+    return new Promise(function (resolve) {
+      let timer = setTimeout(() => resolve(null), initialWaitTimeMillis);
+    }).then(fiberHelpers.bindEnvironment(function() {
+      return pollForDeploy(pollingConfiguration, versionId, false);
+    })).await();
+  } else {
+    // Do a call to the version-status endpoint for the specified versionId
+    let versionStatusResult = deployRpc({
```
abernix: `versionStatusResult ` doesn't seem to get changed so this can be a `const`.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    if(elapsed < timeoutMillis && !terminal) {
+      let newWaitTime = startIntervalMillis*Math.pow(backoffFactor, pollingRounds);
+
+      // Wait for exponentially increasingly more time, and then poll again
+      return new Promise(function (resolve) {
+        let timer = setTimeout(() => resolve(null), newWaitTime);
+      }).then(fiberHelpers.bindEnvironment(function() {
+        return pollForDeploy(pollingConfiguration, versionId, false);
+      })).await();
+
```
abernix: I don't believe a trailing new-line adds much readability in this block.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+// This will only end successfully when the version being polled is marked as the
+// active version by Galaxy. If the build / deploy fails, then the poll
+// will exit. At the end of the poll, this function will return a JSON object
+// which contains a `buildStatus` string, `deployStatus` string, and `isActive` boolean
+function pollForDeploy(pollingConfiguration, versionId, firstRun = true) {
+  let {timeoutMillis,
+    initialWaitTimeMillis,
+    startIntervalMillis,
+    backoffFactor}
+    = pollingConfiguration
```
abernix: Few things:

* `const`?
* Can we use `Ms` instead of `Millis`?  We don't use `Millis` as a suffix in other locations in the codebase.
* semi-colon after `pollingConfiguration`.
* How about formatting like this? (with the primary goal of aligning the curly-brackets!):

```js
  const {
    timeoutMillis,
    initialWaitTimeMillis,
    startIntervalMillis,
    backoffFactor,
  } = pollingConfiguration;
```

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    this.initialWaitTimeMillis = initialWaitTimeMillis || 10*1000;
+    this.startIntervalMillis = startIntervalMillis || 700;
+    this.backoffFactor = backoffFactor || 1.3;
+    this.timeoutMillis = timeoutMillis || fifteenMinutesMillis;
+  }
+
+// Build statuses that indicate we should stop polling for build
+const terminalBuildStatuses = ['failed-permanently', 'success']
+// Deploy statuses that indicate we should stop polling for deploy
+const terminalDeployStatuses = ['success', 'failed', 'aborted']
```
abernix: See related comments about "terminal" below... terminology wise, this seems to be less like a tree with children which might be terminal and more like a recursive.  Maybe this can just be `finishedBuildStatuses` and `finishedDeployStatuses`?

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+  timeoutMillis,
+  initialWaitTimeMillis,
+  startIntervalMillis,
+  backoffFactor) {
+    const fifteenMinutesMillis = 15*60*1000;
+    this.initialWaitTimeMillis = initialWaitTimeMillis || 10*1000;
+    this.startIntervalMillis = startIntervalMillis || 700;
+    this.backoffFactor = backoffFactor || 1.3;
+    this.timeoutMillis = timeoutMillis || fifteenMinutesMillis;
+  }
```
abernix: This closing brace should align with the `function` token on 380.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+      }
+  }
+}
+
+// Creates a polling configuration with defaults if fields left unset
+// Default initialWaitTime is 10 seconds – this is the time to wait before checking at all
+// Default startInterval is 700 milliseconds – this is the first wait interval between polls
+// Default backoffFactor is 1.3
+// Default timeout is 15 minutes
+function PollingConfiguration(
```
abernix: Meteor supports the use of `class` here, though I'm not sure this configuration needs to be instantiated, as it isn't modified.  Any reason not to just return a configuration `Object`? Are you planning to expose this API in the future?

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    \t\t${galaxyUrl}.
+    ******************************************************************************
+    `
+    Console.info(successfulDeploymentMessage);
+  } else {
+    // The deploy did not succeed or the polling timed out
+    if(deploymentPollResult.buildStatus === 'failed-permanently' ||
+      deploymentPollResult.deployStatus === 'failed') {
+        const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+        var failedDeploymentMessage = `
```
abernix: `const`

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+        ******************************************************************************
+        Failed to deploy your app to version ${newVersionNum}. No changes were made.
+        Hostname: ${hostname}, Region: ${region}, Username: ${username}
+        To see the logs for your app, visit ${galaxyUrl}/logs.
+        ******************************************************************************
```
abernix: ...this will be indented by 8 spaces when output, different than the 4 previously used.

Probably best to use line concatenation or to join array elements with `\n`, ala:

```js
        [
          `******************************************************************************`,
          `Failed to deploy your app to version ${newVersionNum}. No changes were made.`,
          `Hostname: ${hostname}, Region: ${region}, Username: ${username}`,
          `To see the logs for your app, visit ${galaxyUrl}/logs.`,
          `******************************************************************************`,
        ].join("\n");
```

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    // The deploy succeeded and the deployed version is now active
+    const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+    var successfulDeploymentMessage = `
+    ******************************************************************************
+    You have successfully upgraded your app to version ${newVersionNum}.
+    Hostname: ${hostname}, Region: ${region}, Username: ${username}
+    To see the status of your app, visit your app's page on Galaxy
+    \t\t${galaxyUrl}.
+    ******************************************************************************
+    `
```
abernix: Closing semi-colon, please!

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+  if(deploymentPollResult && deploymentPollResult.isActive) {
+    // The deploy succeeded and the deployed version is now active
+    const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+    var successfulDeploymentMessage = `
+    ******************************************************************************
+    You have successfully upgraded your app to version ${newVersionNum}.
+    Hostname: ${hostname}, Region: ${region}, Username: ${username}
+    To see the status of your app, visit your app's page on Galaxy
+    \t\t${galaxyUrl}.
+    ******************************************************************************
```
abernix: Leading spaces are preserved in template literals, so this will be indented by 4 spaces when output.

tools/meteor-services/deploy.js
abernix: But to try to be a bit more concise, and avoid the banner `****`, could we just count on the fact that `hostname`, `username`, etc, have already been output above and just say:

    Error: Failed to deploy new app version ${newVersionNum}. See ${galaxyUrl}/logs for details.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    // Poll again if version status isn't in a terminal state and we haven't exceeded the timeout
+    if(elapsed < timeoutMillis && !terminal) {
+      let newWaitTime = startIntervalMillis*Math.pow(backoffFactor, pollingRounds);
+
+      // Wait for exponentially increasingly more time, and then poll again
+      return new Promise(function (resolve) {
+        let timer = setTimeout(() => resolve(null), newWaitTime);
+      }).then(fiberHelpers.bindEnvironment(function() {
+        return pollForDeploy(pollingConfiguration, versionId, false);
+      })).await();
```
abernix: This same pattern is repeated in other places—let's hoist this into a reusable component?

tools/meteor-services/deploy.js
```diff
@@ -442,8 +644,20 @@ export function bundleAndDeploy(options) {
     return 1;
   }
 
+  // After an upload succeeds, we want to poll Galaxy to see if the build / deploy succeeds
+  // We used to indicate that the upload was finished via a message to check Galaxy
   if (result.payload.message) {
-    Console.info(result.payload.message);
+    // If there is a newVersionId, then poll for status of the app
+    let newVersionId = result.payload.newVersionId;
+    if(!!newVersionId) {
```
abernix: Is there a need to cast this to a boolean or can it just be checked as-is?

tools/cli/help.txt
```diff
@@ -499,6 +499,9 @@ Options:
                   downgraded to versions that are potentially incompatible with
                   the current versions, if required to satisfy all package version
                   constraints.
+  --deploy-polling-timeout  The number of milliseconds to wait for build/deploy
+                  success or failure after a successful upload of your app's
+                  minified code, defaults to 15 minutes.
```
glasser: comma splice; replace with semicolon, parenthetical, or new sentence.

tools/meteor-services/deploy.js
```diff
@@ -21,6 +21,8 @@ import {
 } from './auth.js';
 import { recordPackages } from './stats.js';
 import { Console } from '../console/console.js';
+import { setTimeout } from 'timers';
+var fiberHelpers = require('../utils/fiber-helpers.js');
```
glasser: Is there a reason we can't use the more modern `import` instead of `require` here?

tools/meteor-services/deploy.js
```diff
@@ -442,8 +644,20 @@ export function bundleAndDeploy(options) {
     return 1;
   }
 
+  // After an upload succeeds, we want to poll Galaxy to see if the build / deploy succeeds
+  // We used to indicate that the upload was finished via a message to check Galaxy
   if (result.payload.message) {
-    Console.info(result.payload.message);
+    // If there is a newVersionId, then poll for status of the app
```
glasser: I'm not sure why this is inside the `if message` conditional if it doesn't read the message in the new path.

tools/meteor-services/deploy.js
```diff
@@ -442,8 +644,20 @@ export function bundleAndDeploy(options) {
...
+    let newVersionId = result.payload.newVersionId;
+    if(!!newVersionId) {
+      return executePollForDeploymentSuccess(newVersionId, 
+        options.deployPollingTimeoutMillis,
+      result);
```
glasser: indentation is weird here. (maybe Meteor can switch to prettier or something? :) we're hoping to do that soon in galaxy-server)

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
   return parsed.hostname;
 };
 
+// Executes the poll to check for deployment success and outputs proper messages
+// to user about the status of their app during the polling process
+function executePollForDeploymentSuccess(versionId, deployPollTimeout, result) {
+  // Create a default polling configuration for polling for deploy / build
+  // In the future, we may change this to be user-configurable or smart
+  // The user can only currently configure the polling timeout via a flag
+  let pollingConfiguration = new PollingConfiguration(deployPollTimeout);
```
glasser: You never re-assign to this variable, so you can use `const` instead of `let`.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+        true);
+    });
+
+  if(deploymentPollResult && deploymentPollResult.isActive) {
+    // The deploy succeeded and the deployed version is now active
+    const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+    var successfulDeploymentMessage = `
+    ******************************************************************************
+    You have successfully upgraded your app to version ${newVersionNum}.
+    Hostname: ${hostname}, Region: ${region}, Username: ${username}
```
glasser: This will look weird because hostname can be very long. Maybe put each on a separate line.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+      return pollForDeploy(pollingConfiguration,
+        versionId,
+        true);
+    });
+
+  if(deploymentPollResult && deploymentPollResult.isActive) {
+    // The deploy succeeded and the deployed version is now active
+    const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+    var successfulDeploymentMessage = `
+    ******************************************************************************
```
glasser: The `*` thing seems unnecessarily dramatic for an ordinary non-error message that people will see a lot.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    function() {
+      return pollForDeploy(pollingConfiguration,
+        versionId,
+        true);
+    });
+
+  if(deploymentPollResult && deploymentPollResult.isActive) {
+    // The deploy succeeded and the deployed version is now active
+    const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+    var successfulDeploymentMessage = `
```
glasser: Hmm. So the reason we send down the message from the server is that it allows us to improve the message just by changing galaxy-server (immediately affecting even users of relatively old `meteor`). eg we were able to add the warning about `meteor.com` when people kept getting confused that we didn't support it any more.

I'm not saying it's a bad idea to have more structured data in the result and to have some of the message be calculated here but at the very least it would be good to have some field that the server can set for "extra messages".

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    ******************************************************************************
+    `
+    Console.info(successfulDeploymentMessage);
+  } else {
+    // The deploy did not succeed or the polling timed out
+    if(deploymentPollResult.buildStatus === 'failed-permanently' ||
+      deploymentPollResult.deployStatus === 'failed') {
+        const {newVersionNum, hostname, username, region, galaxyUrl} = result.payload;
+        var failedDeploymentMessage = `
+        ******************************************************************************
```
glasser: Stars might be fine here though, since it's an actual error.  I think there might be some way to do colors in the tool too though I kinda hate colors so I'll let you do your own research :)

tools/meteor-services/deploy.js
glasser: Not sure if they already have been output. But I agree that concise is nice.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+        Failed to deploy your app to version ${newVersionNum}. No changes were made.
```
glasser: The error should differentiate between build and deploy errors.

tools/meteor-services/deploy.js
glasser: FWIW 'terminal' makes sense to me but if it doesn't to you, that seems like a reasonable reason to change.

Though I wonder if we should actually allow ourselves some future flexibility by just adding another field to version-status that's a bool "finished", so if we add another state later we don't break old clients. Like ideally we could make all of the state machine logic here driven by bools rather than recognizing lists of states (so same for the build -> deploy transition).

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+// Build statuses that indicate we should stop polling for build
+const terminalBuildStatuses = ['failed-permanently', 'success']
+// Deploy statuses that indicate we should stop polling for deploy
+const terminalDeployStatuses = ['success', 'failed', 'aborted']
+
+// Declare tracking variables for the duration of the poll
+// `start` tracks the time when we started polling so we can know when to time out
+// `pollingRounds` tracks the number of rounds we have polled, resetting when build/deploy succeed
+// `currentStatus` tracks what the current status is of the build and the deploy for this version
+let start, pollingRounds, currentStatus;
```
glasser: Even though we're unlikely to have parallel deploys in the tool, global mutable state is not a great idea. It seems like these could easily go on your polling configuration object (which could be renamed to PollingState or something).

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    = pollingConfiguration
+  // If this is the first poll, we will wait for initialWaitTimeMillis before beginning
+  // We also want to initialize all of our tracking variables
+  if(firstRun) {
+    // Set up tracking variables to persist and be updated throughout polling
+    start = new Date().getTime();
+    pollingRounds = 1;
+    currentStatus = {buildStatus: null, isActive: false};
+    
+    // Timeouts must be done within Promises, so we set a dummy Promise to
```
glasser: I believe there's a `utils.sleepMs`

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    // wait before initiating polling
+    return new Promise(function (resolve) {
+      let timer = setTimeout(() => resolve(null), initialWaitTimeMillis);
+    }).then(fiberHelpers.bindEnvironment(function() {
+      return pollForDeploy(pollingConfiguration, versionId, false);
+    })).await();
+  } else {
```
glasser: Since the if block ends with a return, you can skip the else and avoid a level of indentation

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+    startIntervalMillis,
+    backoffFactor}
+    = pollingConfiguration
+  // If this is the first poll, we will wait for initialWaitTimeMillis before beginning
+  // We also want to initialize all of our tracking variables
+  if(firstRun) {
+    // Set up tracking variables to persist and be updated throughout polling
+    start = new Date().getTime();
+    pollingRounds = 1;
+    currentStatus = {buildStatus: null, isActive: false};
```
glasser: Why not deployStatus?

tools/meteor-services/deploy.js
glasser: Agreed, I think we can just say `Build status: failed-temporarily` or whatever.

Note that galaxy-server (in AppVersionItem.jsx) has maps buildStatusMap and deployStatusMap which translate the codes into more human-suitable messages. Maybe we should extract those maps to a AppVersion.helper and send them down on the RPC (alongside the code) so we can have the terminology match the web interface.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+            Console.info(`Your app failed to build, with buildStatus: ${buildStatus}`);
+            // We also want to stop polling, since deploy cannot succeed with a failed build
+            terminal = true;
+          }
+        } else if (buildStatus !== 'building') {
+          // This will report if the buildStatus changes to a non-terminal, unexpected state
+          //    e.g. failed-temporarily
+          Console.info(`Build state changed to nonterminal state: ${buildStatus}`)
+        }
+      } else if(currentStatus.deployStatus !== deployStatus) {
```
glasser: Not sure that this else is correct: build status and deploy status could maybe both change in a single round?

tools/meteor-services/deploy.js
glasser: Jesse, are there linters set up for meteor/meteor these days that can help contributors with these stylistic issues? I know I never managed to set that up sadly and that there's a variety of styles in the codebase, of course.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+        // Deploy status is treated similarly to buildStatus above
+        currentStatus.deployStatus = deployStatus;
+        if (terminalDeployStatuses.indexOf(deployStatus) >= 0) {
+          if(deployStatus == 'success') {
+            Console.info('Your app has been deployed successfully.' +
+              ' Waiting for new version to be active...');
```
glasser: So I think we should just make it so that when the app has deployed successfully, either it is active or it never will be.  I think that's more or less true now: the only way something will have deployed successfully without being active is if, before we poll, another version becomes active (a new deploy, or somebody clicks 'restore', or whatever).  So I think most of the time waiting for isActive will be trivial, and occasionally it will be infinite, and it will rarely (never?) be anything in between.

(You can see this in scheduler.go: we call PromoteProposedAppVersion immediately before we UpdateAppVersionDeployStatus to success.)

I think it might be interesting to note in the message if isActive is false (say something like "you've deployed successfully but it is no longer the active version", not sure how to phrase that) but blocking doesn't make sense to me, unless I'm missing something.

tools/meteor-services/deploy.js
glasser: it already is, utils.sleepMs.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +323,206 @@ function canonicalizeSite(site) {
...
+      }
+    } else {
+      // If we did not get a valid Version Status response, just fail silently
+      // keep polling as per usual – this may have just been a whiff from Galaxy
+    }
+
+    let elapsed = new Date().getTime() - start;
+    // Poll again if version status isn't in a terminal state and we haven't exceeded the timeout
+    if(elapsed < timeoutMillis && !terminal) {
+      let newWaitTime = startIntervalMillis*Math.pow(backoffFactor, pollingRounds);
```
glasser: Honestly I think exponential backoff is overkill here. It's not like any of our graphql queries to the server when you have a tab open do exponential backoff.  It's not a particularly intense RPC.  And that power of 1.3 gets slow fast.

My understanding is that exponential backoff makes sense when you have reason to believe that an *error* has occurred that might be caused by too much traffic. It's a way to avoid a positive feedback loop where the error never clears because everyone's hammering it, with the tradeoff that it makes things wait longer. But there's no load-based error here; the system that we are polling isn't actually the system that's doing the build/deploy so there's no real way that doing this faster (within reason, and I think 1/sec is easily within reason) will make the overall process slower.

tools/meteor-services/deploy.js
```diff
@@ -21,6 +21,8 @@ import {
 } from './auth.js';
 import { recordPackages } from './stats.js';
 import { Console } from '../console/console.js';
+import { setTimeout } from 'timers';
```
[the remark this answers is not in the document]
zionts: 👍

tools/meteor-services/deploy.js
zionts: Hmm It didn't preserve them for me when I ran locally, but I'll pull them out to the beginning of the line just in case.

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: I figured it would be easier to modify in the future if left as one big string, but I see the motivation behind .join("\n"). I also found the repetition of hostname, region, etc. useful for myself just as a sanity check to make sure you're deploying in the right region with the right hostname, etc.
The banner was just for a clear indication that the deployment succeeded. After waiting up to 10 minutes, I'd be happy to have a big banner dumped onto the screen that makes it very clear deploy has succeeded. It would also be much _much_ easier to find inside of CI logs, which is where we expect people to use this synchronous deployment feature. 

For now I'm gonna make the Failure message more concise (and add "ERROR" in front of it), but leave the success message as is. If you feel strongly about the banner, then lemme know, but I don't imagine it'll bother anyone, and it might help a few people save a minute or two when combing CI logs.

tools/meteor-services/deploy.js
zionts: Changed it to use class.
We are planning on potentially exposing this API in the future. I think I left a comment in there somewhere to that effect, but I'll add another above the class itself.

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: So @glasser are you suggesting moving this logic to galaxy-server? I think that could make a lot of sense. I'll put out a PR for galaxy-server that does that 👍

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: In fact, I'll have it always print the message if it's there. That way, we can maintain the idea you were talking about of having a way to report messages to Meteor users without changing Meteor.

tools/meteor-services/deploy.js
zionts: 👍

zionts:
@abernix as far as your comment about the `--wait` flag, I think that it makes more sense for the default to be synchronicity, but I totally hear what you're saying about not wanting to break people's current setups. Particularly, this feature is for people who are currently using `meteor deploy` in CI, though, so I'd love to talk about a strategy that lets them know about this feature. I see two short-term paths, and I'm open to either:
(1) Keep the old default, but output another message that notes that users can make the deploy command synchronous by adding the `--wait` command. We can later make it synchronous by default if we so choose.
(2) Change the default to the fully synchronous path, and write a blog-post about the change, noting therein that people can maintain the old behavior via a --no-wait command. We also maybe have a message for a little bit after it's released letting people know what's up.

From talking with @glasser I got the impression that people weren't very happy about the half-synchronicity, and that we should change the default to be fully synchronous. I 100% agree that we should keep users informed and not make breaking changes, though, and I'm not sure if (2) is sufficient. What are your thoughts? My only concern is setting this up to be non-standard behavior when it should be the default, but then never making it the default later on.

zionts:
@abernix @glasser this is ready for another look when you get the chance. Goes with the changes to Galaxy Server (https://github.com/meteor/galaxy-server/pull/1340/files)

Basically, this strips away the message-formatting logic from Meteor and puts it in the hands of Galaxy-Server. It also adds a flag for --no-wait, which users can add to have the same behavior as before.

tools/meteor-services/deploy.js
glasser: Ah right. I guess there are two things going on here: JS preserving whitespace, and then Meteor re-wrapping everything it prints.

tools/meteor-services/deploy.js
glasser: Yeah, I think it's a matter of style. I tend to be inclined to more of an old-school Unix aesthetic, where programs are moderately concise on success and try to get your attention more on errors. I recognize that folks are more into colors and emoji and such in the terminal these days.

I'm also honestly a bit skeptical of being sure that the ***s will properly get through the wrapping that Meteor does and end up consistently on their own line...

Perhaps @abernix or @benjamn are a better arbiter of meteor CLI style than I am.

glasser:
@abernix For more context about my thoughts on why waiting should be the default: Historically, mother and Galaxy didn't even have the context of a deploy failing after the deploy RPC succeeded.

When we implemented (first) the image builder and (second) the proposed version / deploy monitor (summers 2016 and 2017), that was essentially a backward-incompatible regression: we went from having `meteor deploy` succeed iff the deploy "succeeded", to having cases exist where `meteor deploy` succeeded even though the deploy failed. I think that this change is essentially rectifying an old compatibility mistake.

In the off chance you really don't want to wait (I dunno, maybe you want to deploy a bunch of apps in semi-parallel?) the flag works.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
   return parsed.hostname;
 };
 
+// Executes the poll to check for deployment success and outputs proper messages
+// to user about the status of their app during the polling process
+function executePollForDeploymentSuccess(versionId, deployPollTimeout, result) {
```
glasser: dunno if the word execute really adds anything to this function name

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+  // Create a default polling configuration for polling for deploy / build
+  // In the future, we may change this to be user-configurable or smart
+  // The user can only currently configure the polling timeout via a flag
+  let pollingState = new PollingState(deployPollTimeout);
```
glasser: Can be const: `let` is literally only for if you later do `pollingState = somethingElse`

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+  
+  var deploymentPollResult = buildmessage.enterJob({
```
glasser: Can be const.  Note that const and let are basically the modern replacements for var. The main difference between let and var is that let has block scope whereas var has function scope.  (I recognize you're copying this line from somewhere else and that this codebase still has lots of var in it. Let's be modern for new stuff though!)

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
 };
 
+// Executes the poll to check for deployment success and outputs proper messages
+// to user about the status of their app during the polling process
+function executePollForDeploymentSuccess(versionId, deployPollTimeout, result) {
+  // Create a default polling configuration for polling for deploy / build
+  // In the future, we may change this to be user-configurable or smart
+  // The user can only currently configure the polling timeout via a flag
+  let pollingState = new PollingState(deployPollTimeout);
+
```
glasser: line has trailing whitespace (you should be able to configure your editor to trim trailing whitespace on save?). There are other lines with trailing whitespace in the file as well. Too bad GitHub doesn't highlight it.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+// Default pollInterval is 700 milliseconds – this is the wait interval between polls
+// Default timeout is 15 minutes
+class PollingState {
+  constructor(timeoutMs,
+    initialWaitTimeMs,
+    pollIntervalMs) {
+      const fifteenMinutesMs = 15*60*1000;
+      this.initialWaitTimeMs = initialWaitTimeMs || 10*1000;
+      this.pollIntervalMs = pollIntervalMs || 700;
+      this.timeoutMs = timeoutMs || fifteenMinutesMs
```
glasser: Missing trailing semicolon

@abernix Is there a current recommendation for how folks submitting PRs to meteor/meteor can ensure they're using correct style? Any kind of linter?

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+}
+
+// Poll the "version-status" endpoint for the build and deploy status
+// of a specified version ID with a polling configuration.
+// This will only end successfully when the polling endpoint reports that
+// the version deployment is finished. The version-status endpoints will report
+// messages pertaining to the status of the version, which will then be reported
+// directly to the user. When the poll is complete, it will return an object
+// with information about the final state of the version and the app.
+function pollForDeploy(pollingState, versionId, firstRun = true) {
```
glasser: Maybe firstRun should just be part of pollingState instead of a separate sorta-state variable?  also not sure why initialize can't just be done in the constructor.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+function executePollForDeploymentSuccess(versionId, deployPollTimeout, result) {
+  // Create a default polling configuration for polling for deploy / build
+  // In the future, we may change this to be user-configurable or smart
+  // The user can only currently configure the polling timeout via a flag
+  let pollingState = new PollingState(deployPollTimeout);
+  
+  var deploymentPollResult = buildmessage.enterJob({
+    title: "Waiting for deploy to succeed..."},
+    function() {
+      return pollForDeploy(pollingState,
```
glasser: You could just put the initial sleep here once, and then you don't need to have the special first run logic.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+    sleepMs(initialWaitTimeMs);
+    return pollForDeploy(pollingState, versionId, false);
+  }
+
+  // Do a call to the version-status endpoint for the specified versionId
+  const versionStatusResult = deployRpc({
+    method: 'GET',
+    operation: 'version-status',
+    site: versionId,
+    expectPayload: [],
```
glasser: You can list 'message' and 'finishStatus' here and then you don't need to do as much error handling later

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+    printDeployURL: false,
+  });
+  
+  //Check the details of the Version Status response and compare message to last call
+  if(versionStatusResult &&
+    versionStatusResult.payload &&
+    versionStatusResult.payload.message) {
+      const message = versionStatusResult.payload.message;
+      if ( currentMessage !== message ) {
```
glasser: don't need spaces inside parens, I think

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+        pollingState.currentMessage = message;
+      }
+  } else {
+    // If we did not get a valid Version Status response, just fail silently and
+    // keep polling as per usual – this may have just been a whiff from Galaxy
+    // We do the retry here because we might hit an error if we try to parse the
+    // result of the version-status call below.
+    if(new Date().getTime - start < timeoutMs) {
+      sleepMs(pollIntervalMs);
+      return pollForDeploy(pollingState, versionid, false);
```
glasser: typo in versionid — needs a capital I, so this would crash. another thing where a linter might help? or just an editor that shows unknown variables in a different color?

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+      if ( currentMessage !== message ) {
+        Console.info(message);
+        pollingState.currentMessage = message;
+      }
+  } else {
+    // If we did not get a valid Version Status response, just fail silently and
+    // keep polling as per usual – this may have just been a whiff from Galaxy
+    // We do the retry here because we might hit an error if we try to parse the
+    // result of the version-status call below.
+    if(new Date().getTime - start < timeoutMs) {
```
glasser: You are not actually calling getTime here: you are doing arithmetic between a function object and a Date.

BTW I think it might reducing the amount of math you have to do if instead of storing a start date and a timeout, you just store a deadline (`new Date(new Date().getTime() + timeout)`) and then you can just compare with `if (new Date() > deadline)`.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+    versionStatusResult.payload &&
+    versionStatusResult.payload.message) {
+      const message = versionStatusResult.payload.message;
+      if ( currentMessage !== message ) {
+        Console.info(message);
+        pollingState.currentMessage = message;
+      }
+  } else {
+    // If we did not get a valid Version Status response, just fail silently and
+    // keep polling as per usual – this may have just been a whiff from Galaxy
```
glasser: This is a good thought — galaxy-server isn't the most stable, and it would be nice to be able to ignore a single 502 error. It might still make sense to display the errorMessage in this case?  And maybe a lot of failures should lead to just failing instead of repeating for 15 minutes.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+  const elapsed = new Date().getTime() - start;
+  const finishStatus = versionStatusResult.payload.finishStatus;
+  // Poll again if version isn't finished and we haven't exceeded the timeout
+  if(elapsed < timeoutMs && !finishStatus.isFinished) {
+    // Wait for a set interval and then poll again
+    sleepMs(pollIntervalMs);
+    return pollForDeploy(pollingState, versionId, false);
+  } else if (!finishStatus.isFinished) {
+    Console.info(`Polling timed out. To check the status of your app, visit
+    ${versionStatusResult.payload.galaxyUrl}. To change the default polling
```
glasser: maybe we should just remember the URL from deploy instead of needing it to be here too?

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+    timeout, use the flag '--deploy-polling-timeout' with 'meteor deploy'`)
```
glasser: maybe "To wait longer, pass a timeout in milliseconds to the '--deploy-polling-timeout' option of 'meteor deploy'." ?

tools/meteor-services/deploy.js
```diff
@@ -321,6 +334,122 @@ function canonicalizeSite(site) {
...
+  let pollingState = new PollingState(deployPollTimeout);
+  
+  var deploymentPollResult = buildmessage.enterJob({
+    title: "Waiting for deploy to succeed..."},
+    function() {
+      return pollForDeploy(pollingState,
+        versionId);
+    });
+
+  if (deploymentPollResult && deploymentPollResult.isActive) {
```
glasser: So I'm not sure how I feel about using "is active" (as opposed to "succeeded") as the success condition. If you haven't configured your CI to only run one commit at a time on your deploy-to-production branch (which is arguably a mistake), then if you push a commit and then quickly push another commit, it probably shouldn't make your CI go red just because Galaxy decided that the later one should override the former one after the former one successfully builds?

Though I dunno, maybe it's just plain broken to set your CI up in a way where it doesn't serialize your deploy branch.

tools/cli/commands.js
```diff
@@ -1407,12 +1409,21 @@ main.registerCommand({
     serverArch: buildArch
   };
 
+  let deployPollingTimeoutMs = null;
+  if (!!options['deploy-polling-timeout']) {
+    deployPollingTimeoutMs = options['deploy-polling-timeout'];
+  }
+
+  let waitForDeploy = !options['no-wait'];
```
glasser: can be const.

zionts: 👍

tools/cli/help.txt
```diff
@@ -499,6 +499,11 @@ Options:
                   downgraded to versions that are potentially incompatible with
                   the current versions, if required to satisfy all package version
                   constraints.
+  --deploy-polling-timeout  The number of milliseconds to wait for build/deploy
+                  success or failure after a successful upload of your app's
+                  minified code; defaults to 15 minutes.
+  --no-wait       Exits when Meteor has uploaded the app's minified code instead
```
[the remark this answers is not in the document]
zionts: 👍

tools/meteor-services/deploy.js
```diff
@@ -75,7 +77,17 @@ function deployRpc(options) {
   if (options.headers.cookie) {
     throw new Error("sorry, can't combine cookie headers yet");
   }
-  options.qs = Object.assign({}, options.qs, {capabilities: CAPABILITIES});
+  // If we are waiting for deploy, we let Galaxy know so it can
+  // use that information to send us the right deploy message response.
+  if(options.waitForDeploy) {
+    let CAPABILITIES_WITH_WAIT = CAPABILITIES.slice();
```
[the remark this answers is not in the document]
zionts: 👍

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: Added to IDEs 👍

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: Decent point. The reason it was in there before was because there was other stuff going on (backoff, backoff restarts, etc.) and I thought it'd be useful to have it all encapsulated in one function. Not necessary anymore though :)

tools/meteor-services/deploy.js
zionts: Thinking about this in another vein, if this isn't the final step in CI, but it then maybe starts up an integration test suite that interacts with their newly deployed app, they'd want to make sure that the newly deployed app is the one that they're talking to.

tools/meteor-services/deploy.js
zionts: I think there is a linter, but it's not extremely strict?

tools/meteor-services/deploy.js
zionts: Got rid of firstRun. I think there was a reason that the initialization was a separate thing before but it definitely doesn't need to be anymore.

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: 👍

tools/meteor-services/deploy.js
zionts: oh whoops idk what happened to those parentheses...

zionts: Good suggestion! Made the change :)

tools/meteor-services/deploy.js
zionts: Yeah VSCode and Intellij don't play well with meteor's code base for me... :/

tools/meteor-services/deploy.js
zionts: I personally find that more annoying because then you need to pass it into pollForDeploy (like 4 times) directly. Also I think it's worthwhile that the version-status endpoint give you everything you need, even if you might be able to get some of it elsewhere.

tools/meteor-services/deploy.js
zionts: Sure

tools/meteor-services/deploy.js
```diff
@@ -334,6 +463,9 @@ function canonicalizeSite(site) {
 //   stats server.
 // - buildOptions: the 'buildOptions' argument to the bundler
 // - rawOptions: any unknown options that were passed to the command line tool
+// - waitForDeploy: whether to poll Galaxy after upload for deploy status
+// - deploy-polling-timeout: user overridden timeout for polling Galaxy
```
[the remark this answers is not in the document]
zionts: Right you are

tools/meteor-services/deploy.js
```diff
@@ -442,9 +575,21 @@ export function bundleAndDeploy(options) {
...
+  }
+  // After an upload succeeds, we want to poll Galaxy to see if the
+  // build / deploy succeed. We indicate that Meteor should poll for version
+  // status by including a newVersionId in the payload.
+  if (options.waitForDeploy && result.payload.newVersionId) {
+    return executePollForDeploymentSuccess(
+      result.payload.newVersionId,
+      options.deployPollingTimeoutMs,
+      result);
+  } else if (!result.payload.message) {
```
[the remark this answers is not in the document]
zionts: I kind of thought it was deletable, but didn't wanna chance it. Deleted it now though 👀

tools/meteor-services/deploy.js
```diff
@@ -442,9 +575,21 @@ export function bundleAndDeploy(options) {
     return 1;
   }
 
-  if (result.payload.message) {
+  // This will allow Galaxy to report messages to users ad-hoc
+  // Also if we are using the --no-wait flag, this will contain the message
+  // that Galaxy used to send after upload success.
+  if(result.payload.message) {
```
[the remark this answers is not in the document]
zionts: Deleted that line and expected it in the payload

tools/meteor-services/deploy.js
zionts: Done 👍

zionts:
screen shot 2018-01-05 at 3 38 06 pm
screen shot 2018-01-05 at 4 14 51 pm

zionts:
Ready eddy spaghetti

zionts:
Hey! I see this was tagged w/ a release number. If you have some time, @abernix @glasser @benjamn, I'd love another peek at it before we push it out to the real world. :)

glasser:
Sorry for the delay — still planning another round of review this week.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +330,120 @@ function canonicalizeSite(site) {
...
+  } else {
+    // If we did not get a valid Version Status response, just fail silently and
+    // keep polling as per usual – this may have just been a whiff from Galaxy
+    // We do the retry here because we might hit an error if we try to parse the
+    // result of the version-status call below.
+    Console.warn('Received malformed response from Galaxy',
+      versionStatusResult.errorMessage);
+    pollingState.errors++;
+    if (pollingState.errors >= pollingState.maxErrors) {
+      throw new Error(versionStatusResult.errorMessage);
```
glasser: This will end up as a stack trace, right? In general we try to avoid printing stack traces in command-line tools for things that aren't bugs in the program itself.

tools/meteor-services/deploy.js
```diff
@@ -321,6 +330,120 @@ function canonicalizeSite(site) {
...
+    } else if(new Date() < deadline) {
```
glasser: space after paren

tools/meteor-services/deploy.js
```diff
@@ -321,6 +330,120 @@ function canonicalizeSite(site) {
...
+  const finishStatus = versionStatusResult.payload.finishStatus;
+  // Poll again if version isn't finished and we haven't exceeded the timeout
+  if(new Date() < deadline && !finishStatus.isFinished) {
+    // Wait for a set interval and then poll again
+    sleepMs(pollIntervalMs);
+    return pollForDeploy(pollingState, versionId);
+  } else if (!finishStatus.isFinished) {
+    Console.info(`Polling timed out. To check the status of your app, visit
+    ${versionStatusResult.payload.galaxyUrl}. To wait longer, pass a timeout
+    in milliseconds to the '--deploy-polling-timeout' option of 'meteor deploy'`);
```
glasser: Missing period at end of sentence.

tools/cli/commands.js
```diff
@@ -1351,7 +1351,9 @@ main.registerCommand({
     // people to deploy from checkout or do other weird shit. We are not
     // responsible for the consequences.
     'override-architecture-with-local' : { type: Boolean },
-    'allow-incompatible-update': { type: Boolean }
+    'allow-incompatible-update': { type: Boolean },
+    'deploy-polling-timeout': {type: Number},
```
glasser: Spacing on this line is inconsistent with the other lines in the PR itself. I honestly don't know what meteor/meteor style is (and it's a shame to have to work on project's without automatically enforced style guides) but internal inconsistency within a change is always good.

tools/cli/commands.js
```diff
@@ -1412,12 +1414,21 @@ main.registerCommand({
     serverArch: buildArch
   };
 
+  let deployPollingTimeoutMs = null;
+  if (!!options['deploy-polling-timeout']) {
```
glasser: !! is extraneous in a conditional

tools/meteor-services/deploy.js
```diff
@@ -321,6 +330,120 @@ function canonicalizeSite(site) {
...
+  // Do a call to the version-status endpoint for the specified versionId
+  const versionStatusResult = deployRpc({
+    method: 'GET',
+    operation: 'version-status',
+    site: versionId,
+    expectPayload: ['message', 'finishStatus'],
+    printDeployURL: false,
+  });
+
+  //Check the details of the Version Status response and compare message to last call
```
glasser: Space at beginning of comment and after if on next line

tools/meteor-services/deploy.js
```diff
@@ -321,6 +330,120 @@ function canonicalizeSite(site) {
...
+      if (currentMessage !== message) {
+        Console.info(message);
+        pollingState.currentMessage = message;
+      }
+  } else {
+    // If we did not get a valid Version Status response, just fail silently and
+    // keep polling as per usual – this may have just been a whiff from Galaxy
+    // We do the retry here because we might hit an error if we try to parse the
+    // result of the version-status call below.
+    Console.warn('Received malformed response from Galaxy',
```
glasser: Not necessarily a malformed response: might just be an error.

tools/meteor-services/deploy.js
```diff
@@ -75,7 +77,13 @@ function deployRpc(options) {
   if (options.headers.cookie) {
     throw new Error("sorry, can't combine cookie headers yet");
   }
-  options.qs = Object.assign({}, options.qs, {capabilities: CAPABILITIES});
+  options.qs = Object.assign({}, options.qs,
+    {capabilities: CAPABILITIES});
```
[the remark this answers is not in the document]
zionts: Alternatively, I could create the copy conditioned on `waitForDeploy` but I think they're about the same thing...

tools/meteor-services/deploy.js
zionts: 👍

zionts:
@glasser Should I do a blog post / make content to let people know about the change as a prereq, too, or can that be done after? I don't really know the format to follow. The default is to poll, so we should definitely say something

glasser:
Yeah I think we should do a post when we release a tool that has this in
it. Maybe we can do a quick 1.6.2 with this?

abernix:
Meteor 1.6.2 won't be a "quick release" per-se (see #9559), but when this is ready to go we could consider a 1.6.1.1 if it doesn't align with the timing for 1.6.2.

glasser:
Ben, can this basically be the same thing but converting functions to `async` and replacing the sleepMs calls with `await somePromiseBasedTimeoutFunction`?

benjamn:
@glasser yep, exactly!

glasser:
OK @zionts let me or @benjamn know if you need help with this.

Maybe start out by squashing what you have into a commit and force pushing so we have a simpler place to start from.

zionts:
@benjamn @glasser let me know if there are further things I should promisify :) Unfortunately, there are some wonky things happening with buildmessage, so I had to take out the the title-capturing bit while we're polling the version-status endpoint. This just means that the loading scroll during meteor deploy (printing out / | \ - ...) has no title, but it still works. Would love to fix it, but I think the buildmessage.js file needs some async / await tlc.

tools/meteor-services/deploy.js
```diff
@@ -21,7 +21,12 @@ import {
 } from './auth.js';
 import { recordPackages } from './stats.js';
 import { Console } from '../console/console.js';
-import { sleepMs } from '../utils/utils.js';
+
+function sleepForMilliseconds(millisecondsToWait) {
+  return new Promise(function(resolve) {
+    let time = setTimeout(() => resolve(null), millisecondsToWait)
```
benjamn: This could just be
```js
return new Promise(resolve => setTimeout(resolve, millisecondsToWait));
```

zionts:
@glasser it only let me rebuild w/ ssh :/ allow-empty is a good one to know! @benjamn I'll rebase now 👍

zionts:
@benjamn couldn't rebase easily since I had done a previous merge... damn that `Update branch` button! Anyway, I think this is ready for another look 👍

benjamn:
@glasser @zionts Is Galaxy Server ready for these changes, or should we hold off on the merge a while longer? No rush, just curious.

benjamn:
@zionts I rebased and squashed a few small commits, which required me to `git push --force` to this branch, so you may need to do
```sh
git fetch origin
git reset --hard origin/adam/synchronous-deploy
```
if there are any more commits you want to add to this branch.

glasser:
Galaxy Server is fully deployed. I encourage you to test a deploy before merging, though!

benjamn:
Hmm, I just tried to deploy a basic app, and got several `Internal error` messages, though the app itself appears to still be deploying (update: it deployed [successfully](http://pr-9466-sync-deploy-test.meteorapp.com/)):
```
~/dev/sync-deploy-test% meteor deploy pr-9466-sync-deploy-test.meteorapp.com
Talking to Galaxy servers at https://us-east-1.galaxy-deploy.meteor.com
Preparing to deploy your app...               
Uploaded app bundle for new app at pr-9466-sync-deploy-test.meteorapp.com.
Galaxy is building the app into a native image.
Internal error                                
Internal error                                
Internal error                                
Internal error                                
Internal error                                
Internal error                                
1:~/dev/sync-deploy-test%
```
Thoughts?

glasser:
I'll take a look today.

--dave

zionts:
@benjamn @glasser I think I figured out what's going on – the version ID isn't stored to the AppVersions database after the successful upload, so galaxy-server is unable to find the version ID being passed.

glasser:
Looks like server side issues. Going to get the new version of the server up ASAP.

glasser:
There were server-side and client-side issues.   I squashed down all the old commits and added one new one that fixes the client-side issue I found.

tools/meteor-services/deploy.js
```diff
@@ -96,11 +99,18 @@ function deployRpc(options) {
     Console.info("Talking to Galaxy servers at " + deployURLBase);
   }
 
+  let operand = '';
```
zionts: Yeah I didn't feel great about calling everything a `site` – thanks for fixing this!

PR #9482, opened 2017-12-18
prefetch dynamic modules when appcache is active

CaptainN:
This prefetches modules into the cache when the `appcache` package is detected.

As described in meteor/meteor-feature-requests#236

Couple of notes:

- I coped the useful `flattenModuleTree` from `client.js` - I'm sure there's a better way to do that
- I'm driving the initial preload off of window.load - I assume this is a good time to start it, as the initial application should already have been loaded, and started. I had it off of Meteor.startup at first, but window.load fires later - after images have loaded. If there's a better event, let me know.

### DIFF
--- a/packages/dynamic-import/dynamic-versions.js
+++ b/packages/dynamic-import/dynamic-versions.js
@@ -53,23 +53,31 @@ function flattenModuleTree(tree) {
 }
 
 // If Package.appcache is loaded, preload additional modules after
-// the core bundle has been loaded
-if (Package.appcache) {
-  // Use window.onload to only prefetch after the main bundle has loaded
-  window.addEventListener('load', function (event) {
-    // prefetch in chunks to reduce overhead
-    // If we call module.prefetch(id) multiple times in the same tick of
-    // the event loop, all those modules will be fetched in one request.
-    function prefetchInChunks(modules, amount) {
-      Promise.all(modules.splice(0, amount).map(function (id) {
-        return module.prefetch(id)
-      })).then(function () {
-        if (modules.length > 0) {
-          prefetchInChunks(modules, amount);
-        }
-      })
-    }
-    // get a flat array of modules start prefetching
-    prefetchInChunks(Object.keys(flattenModuleTree(versions)), 20);
-  })
+// the core bundle has been loaded.
+function precacheOnLoad(event) {
+  // Check inside onload to make sure Package.appcache has had a
+  // chance to become available.
+  if (!Package.appcache) {
+    return
+  }
+  // prefetch in chunks to reduce overhead
+  // If we call module.prefetch(id) multiple times in the same tick of
+  // the event loop, all those modules will be fetched in one request.
+  function prefetchInChunks(modules, amount) {
+    Promise.all(modules.splice(0, amount).map(function (id) {
+      return module.prefetch(id)
+    })).then(function () {
+      if (modules.length > 0) {
+        prefetchInChunks(modules, amount);
+      }
+    })
+  }
+  // Get a flat array of modules and start prefetching.
+  prefetchInChunks(Object.keys(flattenModuleTree(versions)), 20);
+}
+// Use window.onload to only prefetch after the main bundle has loaded
+if (global.addEventListener) {
+  global.addEventListener('load', precacheOnLoad, false);
+} else if (global.attachEvent) {
+  global.attachEvent('load', precacheOnLoad);
 }
--- a/packages/dynamic-import/dynamic-versions.js
+++ b/packages/dynamic-import/dynamic-versions.js
@@ -31,9 +31,9 @@ exports.get = function (id) {
   return version;
 };
 
-function flattenModuleTree(tree) {
+function getFlatModuleArray(tree) {
   var parts = [""];
-  var result = Object.create(null);
+  var result = [];
 
   function walk(t) {
     if (t && typeof t === "object") {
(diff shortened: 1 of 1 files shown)


packages/dynamic-import/dynamic-versions.js
```diff
@@ -30,3 +30,46 @@ exports.get = function (id) {
...
+// If Package.appcache is loaded, preload additional modules after
+// the core bundle has been loaded
+if (Package.appcache) {
+  // Use window.onload to only prefetch after the main bundle has loaded
+  window.addEventListener('load', function (event) {
+    // prefetch in chunks to reduce overhead
+    // If we call module.prefetch(id) multiple times in the same tick of
+    // the event loop, all those modules will be fetched in one request.
+    function prefetchInChunks(modules, amount) {
+      var promises = Promise.all(modules.splice(0, amount).map(function (id) {
```
zimme: `var promises` is unused, so I'm thinking you can just remove it?

packages/dynamic-import/dynamic-versions.js
```diff
@@ -30,3 +30,46 @@ exports.get = function (id) {
...
+  }
+
+  walk(tree);
+
+  return result;
+}
+
+// If Package.appcache is loaded, preload additional modules after
+// the core bundle has been loaded
+if (Package.appcache) {
```
benjamn: Let's move the check for `Package.appcache` inside the `window.onload` callback, just in case the `appcache` package loads after `dynamic-import`.

benjamn: Also, in order to make he `window.onload` even compatible with all browsers, I would do something like [this](https://github.com/meteor/meteor/blob/87681c8f166641c6c3e34958032a5a070aa2d11a/packages/meteor/setimmediate.js#L110-L114):
```js
if (global.addEventListener) {
  global.addEventListener("load", handler, false);
} else {
  global.attachEvent("onload", handler);
}
```

CaptainN: I'll make these changes - what version of IE does Meteor try to maintain support for? Any thought to using something like http://github.com/Raynos/DOM-shim ? It would allow us to remove forking code like this and still allow users who need IE8 and lower a way forward.

Also (this is more for my education) - is `global` a meteor variable?

CaptainN: Also, do you think it'd be worth exporting `flattenModuleTree` - and have `client.js` (which now includes a copy of that function) consume that? Alternatively, I could more specifically tailor the copy I added to `dynamic-versions.js` and just leave it copied (or just leave it as is).

zimme: > Also (this is more for my education) - is global a meteor variable?

https://github.com/tc39/proposal-global

benjamn: Yep, and the `meteor` package defines the `global` variable for all environments [here](https://github.com/meteor/meteor/blob/a7585da87c9e95b5413574f203a2a6adb1c628ff/packages/meteor/package.js#L22-L23).

I think this code is low-level enough that it's worth hand-coding cross-browser code, rather than bundling a library that we'd only use a part of. Likewise, I would support tailoring `flattenModuleTree` to the needs of this PR, especially if you see any opportunities to shorten/simplify the code.

packages/dynamic-import/dynamic-versions.js
```diff
@@ -30,3 +30,54 @@ exports.get = function (id) {
...
+    Promise.all(modules.splice(0, amount).map(function (id) {
+      return module.prefetch(id)
+    })).then(function () {
+      if (modules.length > 0) {
+        prefetchInChunks(modules, amount);
+      }
+    })
+  }
+  // Get a flat array of modules and start prefetching.
+  prefetchInChunks(Object.keys(flattenModuleTree(versions)), 20);
```
zimme: Should `amount` be configurable?

packages/dynamic-import/dynamic-versions.js
CaptainN: It's a minor change to flattenModuleTree - I could make it spit out the array I use directly, instead of converting that from the flat object it currently creates - it would use a bit less RAM, so I'll make that change.

packages/dynamic-import/dynamic-versions.js
```diff
@@ -30,3 +30,55 @@ exports.get = function (id) {
...
+
+  // Get a flat array of modules and start prefetching.
+  prefetchInChunks(getFlatModuleArray(versions), 20);
+}
+
+// Use window.onload to only prefetch after the main bundle has loaded
+if (global.addEventListener) {
+  global.addEventListener('load', precacheOnLoad, false);
+} else if (global.attachEvent) {
+  global.attachEvent('load', precacheOnLoad);
```
benjamn: This should be `onload` instead of `load` (but only for `global.attachEvent`). Thanks, IE.

CaptainN: Ooph! Good catch.

packages/dynamic-import/dynamic-versions.js
benjamn: That's a good question, but I'd like to wait and see if people need this. That said, I think I am going to bump this up to 50.

benjamn:
Merged to `devel` with commit 23e3ac96771bd8c2711659205485f4bcea502849. Thanks for taking care of this, @CaptainN!

Issue #9430, opened 2017-11-30
[standard-minifier-js@2.2.3] Unable to deploy to Galaxy: JavaScript heap out of memory


joncursi:
When I use `standard-minifier-js` version `2.2.1`, I can deploy to Galaxy no problem. But when I upgrade `standard-minifier-js` version `2.2.3`, I get this error every time:

```
[11:28:19]: ▸ Deploying your app...
[11:30:27]: ▸ <--- Last few GCs --->
[11:30:27]: ▸ [19311:0x102802400]   128842 ms: Mark-sweep 1393.8 (1533.3) -> 1393.8 (1533.3) MB, 1202.8 / 0.1 ms  allocation failure GC in old space requested
[11:30:27]: ▸ [19311:0x102802400]   130093 ms: Mark-sweep 1393.8 (1533.3) -> 1393.7 (1498.8) MB, 1250.9 / 0.2 ms  last resort GC in old space requested
[11:30:27]: ▸ [19311:0x102802400]   131302 ms: Mark-sweep 1393.7 (1498.8) -> 1393.7 (1495.8) MB, 1208.7 / 0.2 ms  last resort GC in old space requested
[11:30:27]: ▸ <--- JS stacktrace --->
[11:30:27]: ▸ ==== JS stack trace =========================================
[11:30:27]: ▸ Security context: 0x1f295ca5e91 <JSObject>
[11:30:27]: ▸ 0: builtin exit frame: slice(this=0x1f28ccfa451 <JSArray[1223765]>)
[11:30:27]: ▸ 1: clone [/Users/joncursi/.meteor/packages/standard-minifier-js/.2.2.3.918nku.hhlp4++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:~630] [pc=0x2958d3efec73](this=0x1f2a8412899 <State map = 0x1f2d1a0ddb1>,skipArrays=0x1f264682311 <undefined>)
[11:30:27]: ▸ ...
[11:30:27]: ▸ FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
[11:30:27]: ▸ /bin/sh: line 1: 19311 Abort trap: 6           DEPLOY_HOSTNAME=us-east-1.galaxy-deploy.meteor.com meteor deploy $npm_package_config_stagingUrl --settings $npm_package_config_stagingEnvSettings --owner joncursi
[...]

PR #9483, opened 2017-12-19
Update minifier-js uglify-es dependency to latest version

hwillson:
I know we just went through this in https://github.com/meteor/meteor/pull/9368, but the `uglify-es` folks have released 4 new versions since that PR was merged. This PR updates to `uglify-es` 3.2.2, and will hopefully help fix https://github.com/meteor/meteor/issues/9430. Thanks!

### DIFF
--- a/History.md
+++ b/History.md
@@ -81,7 +81,7 @@
   ```
   [PR #9409](https://github.com/meteor/meteor/pull/9409)
 
-* The `minifier-js` package has been updated to use `uglify-es` 3.1.9.
+* The `minifier-js` package has been updated to use `uglify-es` 3.2.2.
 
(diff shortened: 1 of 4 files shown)


KoenLav:
@hwillson maybe also consider adding the ability to only mangle (not compress)?

Quoted from uglify-es GitHub:
> It's not well known, but whitespace removal and symbol mangling accounts for 95% of the size reduction in minified code for most JavaScript - not elaborate code transforms. One can simply disable compress to speed up Uglify builds by 3 to 4 times. In this fast mangle-only mode Uglify has comparable minify speeds and gzip sizes to butternut.

I would be interested in running some tests with regard to CPU and memory usage in both scenarios.

PR #9487, opened 2017-12-19
Fix cordova lib dependency

skirunman:
`cordova-lib@7.1.0` should use `cordova-common@2.1.1 ` as per https://github.com/apache/cordova-lib/blob/7.1.x/package.json and [this blog post](https://cordova.apache.org/news/2017/10/10/tools-release.html) . I missed this change in this PR https://github.com/meteor/meteor/pull/9213


benjamn:
Thanks @skirunman!

Issue #9480, opened 2017-12-17
Update to 1.6 causing IE11 transpiling issue from node_module on client (sanitize-html)


etyp:
Since upgrading to 1.6, our app has stopped working in IE11 on all versions of Windows. We've narrowed it down by commenting out most app code only to find that certain node modules cause it. The module included here as an example is [sanitize-html](https://github.com/punkave/sanitize-html), but I believe multiple modules cause this.

Meteor version: 1.6.x
Running Meteor from: macOS Sierra 10.12.6
Opening in Windows 10/8.1/7 with IE 11
Did not happen with Meteor 1.5.x
Expected: App should load with no issues
Actual: App throws tons of errors client side (see below)

```
[Excluding many similar errors for brevity]
Unable to get property 'Mongo' of undefined or null reference
'meteorInstall' is undefined
```

[Simple reproduction in Meteor app](https://github.com/etyp/meteor-ie-compile-issue). Note: this was just made by:
```
$ meteor create --fullapp meteor-ie-compile-issue
$ meteor npm i --save sanitize-html
```

Then include [sanitize-html](https://github.com/punkave/sanitize-html) on [client/main.js](https://github.com/etyp/meteor-ie-compile-issue/blob/master/client/main.js#L3). If commented out, it works again.

etyp:
Update: it seems like the issue stems from requiring [postcss](https://github.com/postcss/postcss) in the [sanitize-html](https://github.com/punkave/sanitize-html) package.

Anybody else having issues with postcss in 1.6?

etyp:
Another update, looks like removing [chalk](https://github.com/chalk/chalk) from [postcss](https://github.com/postcss/postcss) solves it as well. Will continue digging...

etyp:
Updating to sanitize-html@1.1.14 resolved the issue. https://github.com/meteor/meteor/issues/9446#issuecomment-350362935

hwillson:
This was a known `sanitize-html` issue - see https://github.com/punkave/sanitize-html/issues/187 and https://github.com/punkave/sanitize-html/issues/189. PR https://github.com/punkave/sanitize-html/pull/190 was accepted to fix this issue, so try updating to the most recent `sanitize-html` version to see if that helps. If not, please consider reporting this issue over in the https://github.com/punkave/sanitize-html repo. Closing here - thanks!

PR #9489, opened 2017-12-20
Update meteor create --full to use meteortesting:mocha

hwillson:
While looking into issue #9480 I noticed that running `meteor create --full` (`devel` branch) returns errors caused by the `--full` app skeleton using the deprecated `practicalmeteor:mocha` package. This PR replaces `practicalmeteor:mocha` with the more up to date and maintained `meteortesting:mocha` package, in the `--full` app skeleton, along with using `chai` from npm instead of `practicalmeteor:chai`.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -797,11 +797,11 @@ main.registerCommand({
   // in the template's versions file).
 
   // Since some of the project skeletons include npm `devDependencies`, we need
-  // to make sure the current node environment is set to development (so those
-  // `devDependencies` are covered by `npm install`).
-  process.env.NODE_ENV = 'development';
-
-  require("./default-npm-deps.js").install(appPath);
+  // to make sure they're included when running `npm install`.
+  require("./default-npm-deps.js").install(
+    appPath,
+    { includeDevDependencies: true }
+  );
 
   var appNameToDisplay = appPathAsEntered === "." ?
     "current directory" : `'${appPathAsEntered}'`;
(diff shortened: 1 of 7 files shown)


tools/cli/commands.js
```diff
@@ -796,6 +796,11 @@ main.registerCommand({
   // the packages (or maybe an unpredictable subset based on what happens to be
   // in the template's versions file).
 
+  // Since some of the project skeletons include npm `devDependencies`, we need
+  // to make sure the current node environment is set to development (so those
+  // `devDependencies` are covered by `npm install`).
+  process.env.NODE_ENV = 'development';
```
hwillson: Interestingly enough `meteor create` will not install `devDependencies` from app skeleton `package.json`'s, as when run the `NODE_ENV` is `production` by default (which means `npm install` is called in production mode which by-passes `devDependencies`). 

This code change will temporarily change the `NODE_ENV` to `development` to make sure the `devDependencies` are picked up. Let me know if this should be handled differently. Thanks!

benjamn: Perhaps this could be a flag to `require("./default-npm-deps.js").install`, which could run `npm install --dev`?

hwillson: Sounds good to me - I'll adjust - thanks!

hwillson: Done @benjamn - quick FYI, it looks like `npm install --dev` has been deprecated and npm's recommended replacement of `--only=dev` doesn't do what we want here, since we want both `dependencies` and `devDependencies` to be installed. Luckily, npm has a wonderfully undocumented feature whereby the `--production` option can be passed `false` to override the current `NODE_ENV`, and install both `dependencies` and `devDependencies`. Go figure - thanks!

Issue #9281, opened 2017-11-01
Broken code coverage (using meteor-coverage) since meteor 1.6


SimonSimCity:
Even though I'm talking about a 3rd party plugin - let me explain how this bug is introduced by Meteor.

The plugin `meteor-coverage` uses `istanbul-api` to gather code coverage information. In order to listen for the coverage, it uses [hooks on the internal `vm`](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-hook/lib/hook.js) module to get this information; one of these hooks was `vm.runInThisContext()` which was [(until this change got merged)](https://github.com/meteor/meteor/commit/f244a5b0c63a7e2b961894db2260b6f9e0de08f6) used to load the files in meteor. The new method (using `vm.Script.runInThisContext()`) is not supported by the time writing.

1. This means, that `meteor-coverage` currently has no valid entry point to register itself.

After I figured that out, I wondered why I still don't get any data about the coverage. Here the devil is in the detail 😉 . The initialization of `meteor-coverage` (which calls the hook to wrap the`vm.runInThisContext()` method into code-coverage) got postponed by [this change](https://github.com/meteor/meteor/commit/8550412bb0eaa3b80ae8725f3bed128d629b9d61#diff-e85840f89d0889ea2a4034009dbccc05R170) while introducing the node-debugger.

2. This means, the hook for code-coverage is initialized after all modules are loaded - which leaves no modules for code-coverage.

I hope you understand my concerns here and I can count on your help @benjamn @abernix and @serut to get it working again 😢 

You can simply check it by starting one of your projects, add `meteor-coverage` to it and open `http://localhost:3000/coverage` in your browser, or add some logs into the `meteor-coverage` package - like I did in https://github.com/serut/meteor-coverage/issues/54.

hwillson:
@SimonSimCity Based on https://github.com/serut/meteor-coverage/issues/54#issuecomment-343734991, it sounds like this can be resolved without making any changes to Meteor. I'll close this for now, but let me know if I'm missing anything and this should be kept open. Thanks!

SimonSimCity:
@hwillson please reopen this issue, as https://github.com/serut/meteor-coverage/issues/54 doesn't solve it and it still needs some adjustment on Meteor - as far as I understand the issue.

hwillson:
Will do @SimonSimCity - I'll keep an eye on https://github.com/serut/meteor-coverage/issues/54 as well.

[...]

PR #9494, opened 2017-12-22
Add back the possibility to use IstanbulJS as plugin

SimonSimCity:
These changes are needed to get the plugin meteor-coverage working. IstanbulJS (shipped with meteor-coverage), can only generate the coverage-report for the code loaded after it's initialization. This is why the code, the plugin meteor-coverage contains, must be executed before the code is loaded, which should be tracked in the code-coverage. A suitable check I found was when a debugger isn't used, which makes it impossible to use code-coverage and the debugger at the same time. It's the only feasible condition I could come up with.

The package meteor-coverage also registers a hook of IstanbulJS which overwrites `vm.runInThisContext()` in order to start the coverage. As of now, IstanbulJS does not support overwriting `vm.Script.runInThisContext()`.

In order to get code-coverage fully working https://github.com/istanbuljs/istanbuljs/pull/99 and https://github.com/serut/meteor-coverage/pull/60 would need a merge as well - just for the record. These two merges change the way IstanbulJS hooks the method `vm.runInThisContext()`.

This will fix https://github.com/meteor/meteor/issues/9281

### DIFF
--- a/tools/static-assets/server/boot.js
+++ b/tools/static-assets/server/boot.js
@@ -390,12 +390,10 @@ var loadServerBundles = Profile("Load server bundles", function () {
     var scriptPath =
       parsedSourceMaps[absoluteFilePath] ? absoluteFilePath : fileInfoOSPath;
 
-    var script = new (require('vm').Script)(wrapped, {
+    var func = require('vm').runInThisContext(wrapped, {
       filename: scriptPath,
       displayErrors: true
     });
-
-    var func = script.runInThisContext();
 
     var args = [Npm, Assets];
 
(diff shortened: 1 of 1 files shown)


tools/static-assets/server/boot.js
```diff
@@ -390,23 +390,26 @@ var loadServerBundles = Profile("Load server bundles", function () {
...
-      args
-    });
+    if (meteorDebugFuture) {
+      infos.push({
+        fn: Profile(fileInfo.path, func),
+        args
+      });
+    } else {
+      // Allows us to use code-coverage if the debugger is not enabled
+      Profile(fileInfo.path, func).apply(global, args);
```
benjamn: In cases when this code executes, I think you need to avoid calling the `func` again after `maybeWaitForDebuggerToAttach` below.

SimonSimCity: The `if` branch adds the `func` function to an array wherefrom it's executed later (after `maybeWaitForDebuggerToAttach `), but the `else` branch, which executes the `func` function immediately, does not add it to the array. By this trick, the method is not called twice here.

PR #9500, opened 2017-12-26
Do not include es5-shim because it's covered in babel runtime

pravdomil:
Babel runtime depends on [core-js](https://github.com/zloirock/core-js#ecmascript-6) that covers es5 features.

### DIFF
--- a/tools/static-assets/skel/.meteor/packages
+++ b/tools/static-assets/skel/.meteor/packages
@@ -13,7 +13,6 @@ tracker                 # Meteor's client-side reactive programming library
 
 standard-minifier-css   # CSS minifier run for production mode
 standard-minifier-js    # JS minifier run for production mode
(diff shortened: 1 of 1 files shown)


benjamn:
Sorry for the delay here. Besides the CLA signing issue, we're not going to take this PR because `es5-shim` does still provide important polyfills that Babel does not (cannot) import from `babel-runtime` and/or `core-js`, because the usage of those polyfills is hard/impossible for Babel to detect statically.

Please note: starting in Meteor 1.6.1, we won't even be shipping the `es5-shim` library to newer browsers (see #9360 and #9439), so using the `es5-shim` package will cost nothing when it isn't needed. Pretty cool!

Of course, if you really believe none of your users will ever need `es5-shim`, you can always remove it by running `meteor remove es5-shim`. This PR only affects whether the package is included by default in new apps.

PR #9507, opened 2017-12-29
Add missing ecmascript dependency to accounts-* packages

### LINKED ISSUE #9506: App JS bundle contains ES6 code coming from accounts-xxx packages
Looking at the JS bundle generated by Meteor 1.6.0.1, I see it contains the `const` keyword in the code coming from the accounts-xxx packages. For example:

```js
if (c.oauth.registerService("facebook"),
        e.isClient) {
            const o = function(e, o) {
                o || "function" != typeof e || (o = e,
                e = null);
                var a = c.oauth.credentialRequestCompleteHandler(o);
                n.requestCredential(e, a)
            };
            c.registerClientLoginFunction("facebook", o),
            e.loginWithFacebook = function() {
                return c.applyLoginFunction("facebook", arguments)
            }
        }
```

This code is generated from https://github.com/meteor/meteor/blob/master/packages/accounts-facebook/facebook.js 

Similar problem is with accounts-google, accounts-github etc.

I don't know whether Meteor's build system transpiles the Meteor packages from ES6 to ES5, but IMHO the resulting JS bundle should not contain any ES6 code. For example, this breaks PhantomJS UI tests as it doesn't support ES6.

A simple fix would be just to change `const` to `var` in the affected packages.

hwillson:
All external service `accounts-*` packages (`accounts-facebook`, `accounts-github`, etc.) are currently using ES2015 syntax ([example](https://github.com/meteor/meteor/blob/3f8c90a29f26ec6dc8ff55f736c36a68fa40398f/packages/accounts-facebook/facebook.js#L4)), but do not explicitly declare a dependency on the `ecmascript` package. This means the ES2015 syntax being used is not transpiled by Meteor, and can lead to issues like #9506. Since `accounts-base` and `accounts-password` both already have `ecmascript` as a dependency, this PR adds an `ecmascript` dependency to all external service `accounts-*` packages.

Fixes #9506.

### DIFF
--- a/packages/accounts-facebook/package.js
+++ b/packages/accounts-facebook/package.js
@@ -1,9 +1,10 @@
 Package.describe({
   summary: "Login service for Facebook accounts",
-  version: "1.3.0"
+  version: "1.3.1"
 });
 
 Package.onUse(function(api) {
+  api.use('ecmascript');
   api.use('accounts-base', ['client', 'server']);
   // Export Accounts (etc) to packages using this one.
   api.imply('accounts-base', ['client', 'server']);
--- a/packages/accounts-github/package.js
+++ b/packages/accounts-github/package.js
@@ -1,9 +1,10 @@
 Package.describe({
   summary: 'Login service for Github accounts',
-  version: '1.4.0'
+  version: '1.4.1'
 });
 
 Package.onUse(function (api) {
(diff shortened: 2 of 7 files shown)


benjamn:
Just published all these new patch versions to Atmosphere.

PR #9512, opened 2018-01-03
Pass login type to client side Accounts.onLogin callbacks

### LINKED ISSUE #5127: Accounts.onLogin clientside not firing when user re-opens page and is already logged in
onLogin is not firing when user opens the page and is already logged in, only when user actively goes from logged out to logged in state.

On the server side it works on both situations and it seems that it should work that way as well for the client side, at least for my use case. Is there any way around this? Is it possible to implement?

Thanks

hwillson:
Client side `Accounts.onLogin` callbacks are triggered after a successful login, but they're also triggered after a successful DDP reconnect (if already logged in). As discussed in #5127, some people think this is the correct behaviour, while others feel that `onLogin` callbacks should really only fire after a user has actually logged in (e.g. after using something like `Meteor.loginWithPassword`). Since people are split on the proper approach here, an alternative solution would be to provide client side `Accounts.onLogin` callbacks with a way to tell if they're being called after a login or after a reconnect, then let them decide what to do.

Server side `Accounts.onLogin` callbacks receive an object that contains various login details. One of those items is a login `type`, that can hold values like `password`, `resume`, etc. When a user completes a true password based login, the returned login `type` is `password`. When the user is already logged in but undergoes a DDP disconnect + reconnect, the returned login `type` is `resume`. This means server side `Accounts.onLogin` callbacks have a way to tell which type of login is happening, allowing them to respond accordingly.

This PR adjusts client side `Accounts.onLogin` callbacks such that they also receive a single login details object, that contains login type information. Unlike the server side login details object, the client side login details object only contains the login type (to help reduce network transfer overhead, make sure we're not exposing server details on the client we shouldn't be, etc.). 

This approach should give developers a better way to respond to different client side login types, and act accordingly.

Fixes #5127.

### DIFF
--- a/History.md
+++ b/History.md
@@ -184,6 +184,7 @@
   successful password based login, `{ type: resume }` after a DDP reconnect,
   etc.).
   [Issue #5127](https://github.com/meteor/meteor/issues/5127)
+  [PR #9512](https://github.com/meteor/meteor/pull/9512)
 
 ## v1.6.0.1, 2017-12-08
 
--- a/History.md
+++ b/History.md
@@ -179,6 +179,12 @@
   supported/maintained.
   [PR #9445](https://github.com/meteor/meteor/pull/9445)  
 
+* Client side `Accounts.onLogin` callbacks now receive a login details
+  object, with the attempted login type (e.g. `{ type: password }` after a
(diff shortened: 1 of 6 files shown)


abernix:
Published as `accounts-base@1.4.2`!

PR #9515, opened 2018-01-04
Do not suppress logging of errors that have the `expected` flag

### LINKED ISSUE #6912: Meteor should not hide errors that have `expected` set
livedata allows tests to set `expected` on an error so that it isn't displayed:

https://github.com/meteor/meteor/blob/4e81417281a371ad93f9191dfc3bfaf7f6cb7124/packages/ddp-server/livedata_server.js#L1671

However, in production, ordinary errors can have the `expected` field set.  For example, the standard Node `assert` module sets the `expected` field on assertion errors.  This makes it hard to track down bugs when the error is silently suppressed.

Meteor should use a different field for tests, such as for example `_expected_by_test`.

jamesmillerburgess:
Fixes #6912

This PR changes the log-suppressing property of errors in `livedata_connection.js` and `livedata_server.js` from `expected` to `_expectedByTest` in order to avoid inadvertently silencing errors in production.


hwillson:
Thanks @jamesmillerburgess - LGTM!

abernix:
Thanks very much for working on these issues, @jamesmillerburgess.  Very much appreciated!

Issue #9521, opened 2018-01-09
1.6.1-rc.1:  cordova app contains only `[object Object]` in index.html


macrozone:
I updated to 1.6.1-rc.1 from 1.6.0 and my cordova build does not work anymore on IOS (did not test with android)

the app does not start. Using safari as inspector i see that the index.html only contains the string `[object Object]`

The app runs correctly as a webpage.

EDIT: this happens on newly created project as well, I tried also previous beta version. The last working version was meteor 1.6.0.1!

So... REPRODUCTION:

- meteor create myapp
- cd myapp
- meteor add-platform ios
- meteor update --release 1.6.1-rc.1
- meteor run ios

macrozone:
ok, i can rule out the babelrc file, as empty file leads to the same problem

macrozone:
I am currently testing it with checked out version of meteor 1.6.1.x, same result... 

What i found so far:

- in tools/cordova/builder.js there is this line:

```
  files.writeFile(files.pathJoin(applicationPath, 'index.html'),
      bootstrapPage, 'utf8');
```

where `bootstrapPage` is a stream. calling `toString()` on that object returns `[object object]` wheras doing something like

```
var streamToString = function(stream, callback) {
        var str = '';
        stream.on('data', function(chunk) {
          str += chunk;
        });
        stream.on('end', function() {
          callback(str);
        });
      }
    streamToString(bootstrapPage, function(myStr) {
      console.log("bootstrap", myStr);
    });
```

will log the correct string,

so i assume that `files.writeFile` no longer treats streams correctly

macrozone:
ok, it's the changes in https://github.com/meteor/meteor/commit/a5866e2781189916868105efeec8448eddb67e5d#diff-5ce4ea994a8e901a5beb2a3a777682bc

it seems this change has not been made to the cordova build.

I am astonished that such a serious bug does make it into a release-candidate... no one uses cordova with meteor anymore?

benjamn:
@macrozone You're a step ahead of me, but the reason that test wasn't running is that we skip "slow" tests for PR builds, for no good reason except to provide feedback faster. I'm going to remove the "slow" marker from that test so that it runs regularly, since clearly it's an important test.

benjamn:
This should be fixed if you run `meteor update --release 1.6.1-rc.2`. Still working on a proper regression test, though.

macrozone:
[...]

PR #9522, opened 2018-01-09
WIP: cordova builder now writes stream into file from boilerplate generator

macrozone:
....instead of assuming it's a string

fixes https://github.com/meteor/meteor/issues/9521


benjamn:
Thanks for taking a crack at this (and for diagnosing the problem in #9521), but I've got another fix coming very soon!

PR #9539, opened 2018-01-16
Support for cursor iterator also on the server side

mitar:
We added iterator on the Minimongo cursor and this is adding it to the server-side as well.

Not sure where should I put tests for this?

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -1119,9 +1119,8 @@ SynchronousCursor.prototype[Symbol.iterator] = function () {
       if (!doc) {
         return {done: true};
       }
-      else {
-        return {value: doc};
-      }
+
+      return {value: doc};
     }
   };
 };
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -855,7 +855,7 @@ Cursor = function (mongo, cursorDescription) {
   self._synchronousCursor = null;
 };
 
-_.each(['forEach', 'map', 'fetch', 'count'], function (method) {
+_.each(['forEach', 'map', 'fetch', 'count', Symbol.iterator], function (method) {
   Cursor.prototype[method] = function () {
     var self = this;
 
@@ -1105,6 +1105,26 @@ _.extend(SynchronousCursor.prototype, {
     }
   }
 });
+
+SynchronousCursor.prototype[Symbol.iterator] = function () {
+  var self = this;
+
+  // Get back to the beginning.
+  self._rewind();
+
+  return {
+    next: function () {
+      var doc = self._nextObject();
+
+      if (!doc) {
+        return {done: true};
+      }
+      else {
+        return {value: doc};
+      }
+    }
+  };
+};
 
 MongoConnection.prototype.tail = function (cursorDescription, docCallback) {
   var self = this;


packages/mongo/mongo_driver.js
```diff
@@ -1106,6 +1106,26 @@ _.extend(SynchronousCursor.prototype, {
...
+
+  return {
+    next: function () {
+      var doc = self._nextObject();
+
+      if (!doc) {
+        return {done: true};
+      }
+      else {
+        return {value: doc};
```
benjamn: Let's add `done:false` explicitly here.

packages/mongo/mongo_driver.js
```diff
@@ -1106,6 +1106,26 @@ _.extend(SynchronousCursor.prototype, {
...
 });
 
+SynchronousCursor.prototype[Symbol.iterator] = function () {
+  var self = this;
+
+  // Get back to the beginning.
+  self._rewind();
+
+  return {
+    next: function () {
```
benjamn: What about method syntax? `next() {...}`

abernix:
To help complete the story and avoid the need to dig around in the history, the related PR being referred to here is https://github.com/meteor/meteor/pull/8888.

packages/mongo/mongo_driver.js
```diff
@@ -1106,6 +1106,26 @@ _.extend(SynchronousCursor.prototype, {
...
+      var doc = self._nextObject();
+
+      if (!doc) {
+        return {done: true};
+      }
+      else {
```
benjamn: Since you're returning in the case above, you can avoid the `else` here and just `return { value: doc, done: false }` below.

packages/mongo/mongo_driver.js
mitar: Because the rest of the code is in old style, I use also here old style. Should I update just this part?

packages/mongo/mongo_driver.js
mitar: We do not have it in [Minimongo version](https://github.com/meteor/meteor/blob/devel/packages/minimongo/cursor.js#L107) and it is [not required by the standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterator_protocol):

> Has the value false if the iterator was able to produce the next value in the sequence. This is equivalent of not specifying the done property altogether.

Should I add it to both or leave it out of both?

packages/mongo/mongo_driver.js
mitar: Updated.

hwillson:
@mitar 
> Not sure where should I put tests for this?

I'm thinking somewhere in [`packages/mongo/collection_tests.js`](https://github.com/meteor/meteor/blob/1fec23f3b7a6d2a4ec45ef726884a336722a68c3/packages/mongo/collection_tests.js), within an `isServer` block.

Also, could you add a small mention of this to the `History.md`? Thanks!

hwillson:
Hi @mitar - will you have a chance to wrap this up (see https://github.com/meteor/meteor/pull/9539#issuecomment-360126771), or should we go ahead and finish this?

mitar:
I had some questions on comments I got, and I am waiting on feedback back on them before I proceed. See inline in the code.

packages/mongo/mongo_driver.js
abernix: Just this part is fine.  No harm in incremental modernization, especially when this is new code.

packages/mongo/mongo_driver.js
abernix: Please add it to both!

benjamn:
Merged to `devel` as commit c0ab6c8c9cc693cb9d9dbaab6a31b8b17221b7af. Thanks for working on this @mitar!

mitar:
Oh, thanks for finalizing this. Sorry for not getting back to this sooner myself.

Issue #9548, opened 2018-01-17
1.6.1-rc6 Cordova plugin 'cordova-plugin-facebook4' - rebuild cordova issue


paulincai:
1.6.0.1 Everything was fine. Update done last night to the latest version in repo.

Running on OSX High Sierra.

I use NPM "bas-meteor-facebook-login": "^1.1.0"

In 1.6.1 rc6 I tried cordova plugins:

1. `cordova-plugin-facebook4@https://github.com/Basgrani-Org/cordova-plugin-facebook4.git#ee8aab56ee6c1f822dcab9ede1db1944100935b4`

2. `cordova-plugin-facebook4@1.9.1`
Same behavior as follows. Start from a meteor reset. build ios-device. Stop. build ios-device again.

This was fine in 1.6.0.1 with the first plugin.

` While removing plugins                     

cordova-plugin-actionsheet,cordova-plugin-android-fingerprint-auth,cordova-plugin-device,cordova-plugin-dialogs,cordova-plugin-facebook4,cordova-plugin-geolocation,cordova-plugin-googleplus,cordova-plugin-inappbrowser,cordova-plugin-meteor-webapp,cordova-plugin-splashscreen,cordova-plugin-statusbar,cordova-plugin-touch-id,cordova-plugin-vibration,cordova-plugin-whitelist,cordova-plugin-wkwebview-engine,cordova-plugin-x-socialsharing,cordova.plugins.diagnostic,es6-promise-plugin,ionic-plugin-keyboard,uk.co.workingedge.phonegap.plugin.launchnavigator

   from Cordova project:

   Cordova error: **Cannot find plugin.xml for plugin "cordova-plugin-facebook4". Please try adding it again.**
   (If the error message contains suggestions for a fix, note that this may not apply to the Meteor integration. You can try running again with the --verbose option
   to help diagnose the issue.)`

I can only build cordova if I do meteor reset or manually delete 'cordova-build' folder

OBSERVATIONS: after a meteor reset and successfully run an ios-build I can see the facebook plugin under cordova-build/plugins. I stop meteor, I can still see it. I run ios-device again, the cordova-plugin-facebook4 folder disappears then I get the error in the server console.

![screen shot 2018-01-17 at 22 57 36](https://user-images.githubusercontent.com/26188915/35061188-e0fc2ea8-fbd9-11e7-97ae-8af22d8f8c72.png)

paulincai:
I started a brand new project from 0. The same issue occurs.
[...]

PR #9274, opened 2017-10-30
Release 1.6.1

benjamn:
Now that Meteor 1.6 has been finalized and officially recommended, it's time to get started on the next 1.6.x version, 1.6.1.

To be clear, there is no immediate urgency to this release, since Meteor 1.6 seems to be working well for most developers. 🤞 

See the [Release 1.6.1 milestone](https://github.com/meteor/meteor/milestone/45) for a partial list of issues we hope to tackle in this release.

PR #9553, opened 2018-01-18
vNext Cordova Android version is 6.4.0

paulincai:


skirunman:
Looks like we are not actually using `cordova-android@6.4.0`, but instead this forked and modified version `https://github.com/meteor/cordova-android/tree/v6.4.0-with-pr-417` so might want to add that to history.

// This commit represents cordova-android@6.4.0 plus https://github.com/apache/cordova-android/pull/417

paulincai:
@skirunman I think we're good now.

benjamn:
I don't mind merging this and dealing with any conflicts, but I wanted to mention a similar note has been added on the `release-1.6.1` branch already: https://github.com/meteor/meteor/pull/9274/commits/322cbe8593bb33c36a43a2fd6c56dd72206d6689

benjamn:
Sure, it might be a little easier to close this PR and wait for the official 1.6.1 release to get merged into `master` and `devel` (soon!).

benjamn:
More importantly, please let me know soon if the latest 1.6.1 RCs are (or aren't) working for your Cordova apps!

paulincai:
@benjamn so far so good, I am testing Push notifications in the next 1 hour. I am doing my launch screens now. Meanwhile just to let you know that I opened this: https://github.com/meteor/meteor-feature-requests/issues/255

paulincai:
@benjamn just getting back to you with my findings with rc6. after running production grade apps (deployed with Testfairy) on Android 5.1, Android 7, Ipad Air 1 Retina with IOS 11.2.5, Iphone 6 with IOS 11.2.5.

- Not Meteor related: In general there is some mess up with google-services on Android if you install various plugins that depend on this. For instance I was not able to install the latest phonegap-plugin-push together with the latest cordova-plugin-googleplus Things got too complicated and annoying and I gave up. I reverted back to phonegap-plugin-push@1.10.5 which has a good compatibility with Meteor  RAIX:push and does the job.

- Otherwise, Android is ok, nothing to comment, everything is as expected.

- IOS: Let me avoid the technical terms and give you my experience. I start my Test app, get a splash screen, get the app... amazing. Put the Test app in background or just open something else (Whatsapp etc ) on top. 5 min later I get back to my Test app, I am right where I left it. However, if I get back to it after 10 min, I get the last screen I've had in the web view in my Test app for 2 seconds, and then a start over from splash screen.

This is the same when I use Test app, close it from Home button, switch the phone off and get back to my Test app after a while. Meanwhile there are no updates done to the code and I am connected to a production server in Heroku. 

I tested this particular issue over a day and every single time I get the same result: old view -> splash -> home screen of the app.

The expectation is that after a while I should start the app directly from the splash screen (this is fine in Android).

- Cordova in general: still getting the [facebook plugin issue](https://github.com/meteor/meteor/issues/9548#issue-289376190) . This prevents me from adding some sounds in the Android 'res' folder because every time I build for Android I need to do it right after I delete 'cordova-build' otherwise I get the facebook plugin issue.

evolross:
It's like Clash Royale... everytime you leave that app for two seconds or more you get nailed with the splash screen again! 😝

paulincai:
@evolross not sure I understand. Do you face some issues on one of the mobile platforms?

evolross:
Sorry, it was a joke. I guess you don't play Clash Royale. I probably shouldn't joke in a problem thread. Ignore my comment. No, I don't have this issue you're experiencing.

benjamn:
@paulincai From what I can tell, the iOS background restore problems can be fixed with updates to `cordova-plugin-meteor-webapp` (e.g. https://github.com/meteor/cordova-plugin-meteor-webapp/pull/30), and we can release those updates by publishing a new version of the Meteor `webapp` package at any time, so we don't have to delay Meteor 1.6.1. That's good news from my perspective, though I understand the problems are annoying and need to be fixed soon, regardless of how we do it.

paulincai:
@benjamn was just updating my post above :)) Please come back in 5 min for the latest edit.

paulincai:
@benjamn the only problem I still see is related to Facebook as described above.

benjamn:
@paulincai From what I can tell, the facebook plugin has this problem because it requires `APP_ID` and `APP_NAME` to be configured in `mobile-config.js`, and just because `App.configurePlugin("cordova-plugin-facebook4", ...)` is present in `mobile-config.js`, Cordova assumes the plugin must be installed, so it tries to uninstall it when rebuilding, and fails because the uninstall operation can't handle the possibility that the plugin isn't actually installed (that is, `.meteor/local/cordova-build/plugins/cordova-plugin-facebook4` does not exist). Clearly, the uninstall operation should be more robust. Not sure how best to fix that, but I'll keep thinking about it.

paulincai:
@benjamn things got a little bit more complex now with the evolution of some other plugins. Push plugin (> 2.0.0) requires some configurations with local files and custom gradle files ... it is becoming really difficult to do anything on the Android side having to remove the cordova android project before every rebuild. Push on its own needs some debugging to do with Android 7 and 8 and 8.1 and starting from scratch every single time has become unbearable. I feel that (to me) building the Android App takes as much as everything else together.

Issue #9426, opened 2017-11-28
Meteor not fully working on Windows Subsystem for Linux (WSL) [with workaround]


brucejo75:
I have been working to get Meteor up and running under Windows Subsystem for Linux (WSL).  I have met with some pretty good success, but I am now blocked on one issue.  Also, I am finding that I really like the `bash` shell.

I can run a single `meteor` instance.  If I try to run a 2nd instance I get an error:
```
Error: SQLITE_IOERR: disk I/O error
```

(Why am I running a 2nd instance?  I have a microservice and 2 client apps using it)

This appears to be because of: https://github.com/Microsoft/WSL/issues/1927
Also related: https://github.com/Microsoft/WSL/issues/2395

Also, others are seeing this as well: https://forums.meteor.com/t/meteor-create-error-in-vagrant-box/9133

### Workaround?
These bugs claim that the issue can be worked around if WAL mode is disabled for SQLite access.  I hacked my local version of the tools by changing this line:
https://github.com/meteor/meteor/blob/devel/tools/packaging/catalog/catalog-remote.js#L144
to
```
self._execute('PRAGMA journal_mode=TRUNCATE');
```
And it worked!

#### Option 1 new environment variable
Add an environment variable, `METEOR_SQLITE_JOURNAL_MODE`.  If `METEOR_SQLITE_JOURNAL_MODE` is set then open SQLite with `journal_mode=$METEOR_SQLITE_JOURNAL_MODE`.

A flag like this may be useful for users that are seeing this issue.

#### Option 2 change journal mode
or could simply make the change to `TRUNCATE`.

Let me know if you want me to submit a PR for this.

### Environment
- Window 10 Version 1709 (OS Build 16299.64)
- WSL Ubuntu Release: 16.04

### Repro
1. Get meteor running on WSL essentially it is the same as for Ubuntu
2. Start meteor instance using port `3080`
3. Verify it works.
4. Start a 2nd meteor instance using port `4000`

**Result:** fails with `Error: SQLITE_IOERR: disk I/O error`

### Full Output
```
bruce@DEVPC:/mnt/e/u/mroot/meris$ meteor npm start

> MERIS@1.0.0 start /mnt/e/u/mroot/meris
> meteor npm run startl


> MERIS@1.0.0 startl /mnt/e/u/mroot/meris
> env ROOT_URL=$LOCAL_URL meteor -p $PORT --mobile-server $LOCAL_URL --settings settings-development.json

[...]

PR #9559, opened 2018-01-21
Release 1.6.2

benjamn:
Now that Meteor 1.6.1 has been finalized (official "recommendation" and blog post coming soon!), it's time to turn our attention to the exciting work that was put off until Meteor 1.6.2:
* https://github.com/meteor/meteor/milestone/47
* https://github.com/meteor/meteor-feature-requests/milestone/3

I'm especially excited about https://github.com/meteor/meteor/pull/9439, which promises to reduce bundle sizes dramatically for modern browsers and Node, and enable truly native `async`/`await` (and other ECMAScript features) in the browser.

[Meteor 1.6.1](https://github.com/meteor/meteor/pull/9274) grew into a very large release (486 commits, even more than Meteor 1.6). While I hope 1.6.1 was worth the wait, I would also hope that we can release a useful Meteor 1.6.2 in less than three months, though of course time (and your feedback!) will tell.

As with previous Meteor beta testing cycles, whenever you see a commit added to this pull request with a message like `Bump package versions for 1.6.2-beta.n release.`, that means you can run `meteor update --release 1.6.2-beta.n` to update an existing app to the new version, or `meteor create --release 1.6.2-beta.n your-new-app` to create a fresh Meteor 1.6 app.

If you want to participate in the development of Meteor 1.6.2, you should check out the [`release-1.6.2` branch](https://github.com/meteor/meteor/tree/release-1.6.2), regenerate the dev bundle, and run the `meteor` script from a checkout:

```sh
cd path/to/meteor
git fetch origin
git checkout -t origin/release-1.6.2
scripts/generate-dev-bundle.sh
./meteor --help
```

See [these instructions](https://github.com/meteor/meteor/blob/devel/Development.md#running-from-a-git-checkout) for help with running Meteor from a checkout.


brucejo75:
Hi Benjamin, I noticed that [Support "Bash on Windows](https://github.com/meteor/meteor-feature-requests/issues/154) did not make the cut.

I have been successfully using `WSL` since the end of November without any issues really.  It would be great to "slip" this in!

It is really a tremendous feature.  Simplifies a lot.

benjamn:
@brucejo75 Sorry for my ignorance on this matter, but what sort of changes would be necessary to make Meteor work in WSL?

benjamn:
@KoenLav @macrozone @wojtkowiak I'd love to hear if 0ec73ca63a2c22b31b6c31db8e4799fc3d0a9283 (included in `1.6.1-beta.2`) helps with background resuming and/or any other problems related to `cordova-plugin-meteor-webapp`.

brucejo75:
Hi @benjamn , Gory details are in #9426, identifying the issue with the SQLite DB `WAL` mode in `WSL`.

The only change I made was to alter the journal mode of the SQLite DB to be `TRUNCATE` [in catalog-remote.js](https://github.com/meteor/meteor/blob/devel/tools/packaging/catalog/catalog-remote.js#L144).

In the feature request @abernix asked me to gather some timings and see if it worked on windows as well.  I did the timings and you can find the results here: https://github.com/meteor/meteor-feature-requests/issues/154#issuecomment-348378482.

I am happy to create a pr.  (I was waiting for @abernix go ahead).

I think it would be good to get this early into a release so we could get feedback on many other system configurations.  I could not find any supporting bug for the use of `WAL` mode so it may have been a simple dev decision.  So any mode may work fine for this seemingly simple application.

benjamn:
@brucejo75 Please do create a PR, especially if you have a good way of detecting WSL.

abernix:
@brucejo75 I updated https://github.com/meteor/meteor-feature-requests/issues/154 with my thoughts.  Thanks for raising it here, and happy to discuss more in that FR.

jamesmillerburgess:
Both the `web.browser` and `web.browser.legacy` client bundles are shown in the `bundle-visualizer` package with `1.6.2-beta.3`. 

Seeing both together is convenient, but it looks a little like both bundles are getting served to the same client. I'm assuming only one of these bundles was actually served.

Maybe there is a good way to visually distinguish these, like two separate sunbursts or names on the bundles instead of `bundle (1)` and `bundle (2)`?

```
meteor create --bare .
meteor add bundle-visualizer
meteor --production --release 1.6.1            // 1 bundle in Chrome   334 kb
meteor --production --release 1.6.2-beta.3     // 2 bundles in Chrome  361 kb and 236 kb
```

two bundles

abernix:
@jamesmillerburgess Good observation! I think the bundle name (`web.browser` vs `web.browser.legacy`) will be important to reveal regardless of whether it becomes a separate sundial or not.  That said, two sundials will take up quite a bit of room (and likely over-complicate the current algorithm for rendering it, since it'll need to be aware of the outer-limits of the other sundial), so a [tab-bar](https://getbootstrap.com/docs/4.0/components/navs/#tabs) or [pill-bar](https://getbootstrap.com/docs/4.0/components/navs/#pills) might be sufficient/ideal.   Other visually-appealing ideas also welcome. 🎨

KoenLav:
@benjamn unrelated to Cordova, but when I try to run 1.6.2-beta.3 I keep getting this error:

Unexpected mongo exit code 3221225784. Restarting.

So I can't test on Cordova yet.

Have already tried meteor reset to no avail.

brucejo75:
@abernix, @benjamn,

I submitted pr #9644 for the `WSL` changes.

KoenLav:
@benjamn we tested the new cordova-plugin-meteor-webapp on iOS and have not found any issues.

So I think it is good to go! 

@macrozone @wojtkowiak have you had any chance to test the changes?

macrozone:
@KoenLav not yet, hope to do it in the next weeks

History.md
```diff
@@ -1,5 +1,22 @@
 ## v.NEXT
 
+* Applications may now specify client and server entry point modules in a
+  newly-supported `"meteor"` section of `package.json`:
+  ```js
+  "meteor": {
```
jbaxleyiii: 🎉 🎉 🎉

codedeep:
Update to node 8.10.0 (LTS) with v8 6.2 In this release? https://nodejs.org/en/blog/release/v8.10.0/

benjamn:
@codedeep See #9725!

yorrd:
how far up is the node_modules compilation on your priority list?

I'm trying to plan the future of our internal base project, which will heavily depend on es6 libraries which don't have a minified version served alongside the source (polymer). Since the feature request has already been moved to a new milestone once before, I was wondering how likely you think this would be to happen again. No pressure, just trying to figure out what to do here :)

sebakerckhof:
I'm trying out Meteor 1.6.2 (beta 12). I see a lot of time during rebuilds is spend on 'Building for web.browser.legacy'. How can I tell Meteor I'm not interested in web.browser.legacy and it can skip that work?

benjamn:
@sebakerckhof The `web.browser.legacy` architecture includes any browsers that do not natively support `async`/`await` (more or less), so I'm sorry to tell you: you probably do care about it.

sebakerckhof:
@benjamn I know, but I've go the luxury of not having to support those browsers since my project builds heavily upon webrtc, which means more or less a subset of those supporting async/await.
+ I can imagine a lot of people doing most of their development only on latest chrome and only checking other browsers every now and then. For those people this could make a difference in rebuild times.

coagmano:
And those of us using Meteor for kiosk apps where we have the luxury to pin an exact version of chrome 😍

CaptainN:
@sebakerckhof @benjamn Is it possible to rebuild and deliver `modern` first, without waiting for `web.browser.legacy` to finish rebuilding? Then we can have it both ways.

pozylon:
> Is it possible to disable web-browser-legacy build via platforms options?

I was asking myself almost the same question. Right now it's not possible to remove the "browser" build target. I have that use case where I use Meteor only as a backend for a GraphQL API (no client) and I don't need to support old or new browsers at all. Some atmosphere packages I use at the top-level contain client or common code, so they get compiled and I still get a client side bundle. If I could remove browser from the platforms file, build time would drop significantly for my project.

benjamn:
Interesting ideas, all! I think it would be great if there was a way to disable building certain architectures temporarily, though I worry that build problems in the architectures you're not testing in development might go unnoticed until deployment, so I also want to think about how we can make sure timely feedback is given. Maybe all architectures should be built when you first do `meteor run`, but not when the server restarts? Would it be confusing if you happen to visit the page in an older browser, and it's not the latest version of the code?

sebakerckhof:
Ideally, I think step 1 should be having the possibility to disable certain architectures permanently. So we could remove `web.browser.legacy` or even the entire browser platform by not having them in the platforms file. Step 2 would be to disable some temporarily during development. I think the behavior you propose would confuse people new to meteor. What about a command line option similar to the `--extra-packages` option. Either a blacklist `--exclude-platforms` or whitelist `--platforms` option? That makes it very explicit to the developer and still fits in the Meteor philosophy of being low on configuration.

CaptainN:
Maybe a related question - is there a way to add build targets? I'm thinking worker (service worker and web worker) end points.

benjamn:
@CaptainN Not in a developer-configurable way (yet), but if you grep for `web.browser.legacy` on this branch within the `tools` directory, there are only three places it's mentioned outside of tests/comments. Should be pretty easy to add more, though the cost of building more architectures is important to consider. That said, if a worker architecture didn't inherit everything from `web.browser` (as `web.browser.legacy` does), it might be a lot cheaper to support.

KoenLav:
Any ETA on the release? 1.6.2 is getting pretty bit already, just like 1.6.2!

benjamn:
@KoenLav Hopefully soon! I want to make sure we've updated to Node 8.10.1 first, at least: https://github.com/meteor/meteor/pull/9725

benjamn:
I want to draw special attention to a new feature in `1.6.2-beta.14`: **selective compilation of packages and modules within `node_modules` directories** (https://github.com/meteor/meteor/pull/9771, https://github.com/meteor/meteor-feature-requests/issues/6).

As described in [`History.md`](https://github.com/meteor/meteor/blob/05d62dbf9ebf28f69223f948fd126c8358e29e97/History.md#vnext), here's how Meteor 1.6.2 plans to enable this much-sought-after functionality:

Although Meteor does not (and will not) recompile packages installed in `node_modules` by default, compilation of specific npm packages (for example, to support older browsers that the package author neglected) can now be enabled in one of two ways:

  * Clone the package repository anywhere in your application outside of `node_modules`, make any modifications necessary, and then use `npm install` to link `the-package` into `node_modules`:
    ```sh
    meteor npm install imports/the-package
    ```
    Meteor will compile the contents of the package exposed via `imports/the-package`, and this compiled code will be used when you import `the-package` in any of the usual ways:
    ```js
    import stuff from "the-package"
    require("the-package") === require("/imports/the-package")
    import("the-package").then(...)
    ```
    This reuse of compiled code is the critical new feature that we are adding in Meteor 1.6.2. Without this behavior, you'd get the raw source code of the package when you tried to import it from `node_modules`.

  * Install the package normally with `npm install the-package`, then create a symbolic link *to* the installed package elsewhere in your application, outside of `node_modules`:
    ```sh
    meteor npm install the-package
    cd imports
    ln -s ../node_modules/the-package .
    ```
    Again, Meteor will compile the contents of the package because they are exposed outside of `node_modules`, and the compiled code will be used whenever `the-package` is imported from `node_modules`.

    > Note: this technique also works if you create symbolic links to individual files, rather than linking the entire package directory.

In both cases, Meteor will compile the exposed code as if it was part of your application, using whatever compiler plugins you have installed. You can influence this compilation using `.babelrc` files or any other techniques you would normally use to configure compilation of application code.

CaptainN:
Sourcemaps don't seem to work correctly in Firefox. They return large position numbers in modules.js

benjamn:
@CaptainN Modules in the application's `node_modules` directory get relocated from `app.js` to `modules.js`, since the `modules` package loads before any other code that uses the `modules` package. So it's not unusual for `modules.js` to be relatively large, and to show up in stack traces for npm package code. But it seems like there's something else wrong, besides the surprising file name? Can you find a previous `1.6.2-beta.n` version where this worked?

CaptainN:
@benjamn I didn't realize before, but these errors are being propagated to the client from the server. I'm using SSR in this case.

On the server (in my server's console) I do get the original source line numbers. So that's a little confusing - I didn't even realize these were being propagated from the server, and they aren't using the source maps when displayed on the client (I noticed that some errors about Fibers are also being shown in the browser console, which is why I believe these are coming from the server).

To trigger this error, I just put an incorrect value (style="string") in a React component in my `/imports/` directory.  I went all the way back to Meteor 1.6.1 and still observed the error.

One other thing - I seem to be regularly entering a refresh loop when I hit these errors, which is making it difficult to read the errors. Maybe this also has to do with SSR?

benjamn:
Thoughts on calling this release Meteor **1.7** instead of 1.6.2? There's some pretty incredible new functionality here, IMHO. Either way, we're going to release as soon as Node 8.11.2 is out, with the fix for https://github.com/nodejs/node/issues/19274… which means we should start RCing real soon.

yorrd:
@benjamn well quite frankly, as said in other threads, this release will be a major life saver for us. It keeps Meteor a no-brainer because doing it another way is just more work. Just talking about the build system here (zero config!). This might be subjective but compared to what 1.6.1 had in store for us, this is more anticipated by the team by several orders of magnitude :+1:

dagatsoin:
Glad to here that. In my case my project can't use 1.6.1 because of Cordova compatibility with external plugins + the mess with babel7
This should be fixed in 1.6.2, but is seems weird to me to have not all the 1.6.X version available for my project.
So, from my pov, I see the 1.6.2 as the 1.7 version.

benjamn:
Per the [emoji reaction poll](https://github.com/meteor/meteor/pull/9559#issuecomment-379282728) above, let's shift our focus to the `release-1.7` branch: https://github.com/meteor/meteor/pull/9826

PR #9567, opened 2018-01-23
Update CONTRIBUTING.md

marhalish:


marhalish:
I think this addition is a great way to give back to the Meteor community and give more incentive for the growing community to continue improving MeteorJS. Plus, D.Tube and other platforms which are major users of MeteorJS are also in the Utopian.io community.

marhalish:
All done

hwillson:
Thanks for the interesting suggestion (and PR) @marhalish. We prefer to discuss both code and policy related feature requests like this in an issue first, over in the http://github.com/meteor/meteor-feature-requests repo. This gives the Meteor community a chance to chime in with their input, and helps the Meteor maintainers decide if something like this is a good fit for Meteor. Would you mind opening a new issue in the FR repo that outlines this idea? We can then re-visit this PR, depending on the direction the ensuing discussion takes. Thanks again!

PR #9579, opened 2018-01-26
Add a self-test skip option

hwillson:
While doing some test work on another PR, I accidentally ran a `meteor self-test` without my usual `--exclude` list. When that happened, I noticed that  https://github.com/meteor/meteor/commit/cc472785641edc8d5d6c5da152ff00166c3d6e72 broke one of the `command-line: old cli tests (converted)` self-test's, that is no longer usually called (it's currently excluded by the CI configs).  Looking into fixing this has sent me on a complete tangent (it must be Friday 🙂), leading to this PR ...

Meteor's CI infrastructure is configured to exclude certain `self-test`'s on each run. These excludes are specified in each CI environment's config file, and included when running `meteor self-test`. Developers running `meteor self-test` locally however are not using these excludes by default, so developer's have to manually look up the current exclude list from one of the CI configs, then add these excludes to their own `meteor self-test` call manually.

This commit adds a new `skip` option to Meteor's `self-test` system, that can be used to skip adding/running a defined `self-test` (similar in concept to Mocha's `skip` feature). This provides a way to skip the running of older `self-test`'s that are no longer needed, but allows them to be preserved in the `self-test` suite, for future reference. With this functionality in place, and the older test suites updated to use it, Meteor's base CI excludes no longer need to be maintained in their respective config files. The excludes are all managed at the source (the test definition), and can be leveraged by anyone/anything calling `meteor self-test`.

Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,13 +91,10 @@ build_machine_environment: &build_machine_environment
     METEOR_SAVE_TMPDIRS: 1
 
     # Skip these tests on every test run.
-    # For readability, this is a regex wrapped across multiple lines in quotes.
-    SELF_TEST_EXCLUDE: "\
-        ^old cli tests|\
-        ^minifiers can't register non-js|\
-        ^minifiers: apps can't use|\
-        ^compiler plugins - addAssets\
-        "
+    # If needed, for readability this should be a regex wrapped across
+    # multiple lines in quotes.
+    SELF_TEST_EXCLUDE: ""
+
     # These will be evaled before each command.
     PRE_TEST_COMMANDS: |-
         ulimit -c unlimited; # Set core dump size as Ubuntu 14.04 lacks prlimit.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,12 +15,7 @@ environment:
   TOOL_NODE_FLAGS: --expose-gc
   TIMEOUT_SCALE_FACTOR: 8
   METEOR_HEADLESS: true
-  SELF_TEST_EXCLUDE: "\
-      ^old cli tests|\
-      ^minifiers can't register non-js|\
-      ^minifiers: apps can't use|\
-      ^compiler plugins - addAssets|\
-      ^NULL-LEAVE-THIS-HERE-NULL$"
+  SELF_TEST_EXCLUDE: "^NULL-LEAVE-THIS-HERE-NULL$"
 platform:
   - x64
 
--- a/tools/tool-testing/selftest.js
+++ b/tools/tool-testing/selftest.js
@@ -153,9 +153,13 @@ export function define(name, tagsList, f) {
 
 // Prevent specific self-test's from being defined and run.
 // e.g. `selftest.skip.define("some test", ...` will skip running "some test".
+const selftestDefine = define;
 export const skip = {
-  define(testName) {
-    Console.info(`Ignoring "${testName}" test (excluded by selftest.skip)`);
+  define(name, tagsList, f) {
+    if (typeof tagsList === 'function') {
+      f = tagsList;
+    }
+    selftestDefine(name, ['manually-ignored'], f);
   }
 };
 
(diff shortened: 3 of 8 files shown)


tools/tool-testing/selftest.js
```diff
@@ -151,6 +151,14 @@ export function define(name, tagsList, f) {
   }));
 }
 
+// Prevent specific self-test's from being defined and run.
+// e.g. `selftest.skip.define("some test", ...` will skip running "some test".
+export const skip = {
+  define() {
+    /* no-op */
```
benjamn: Sometimes test frameworks will print something to indicate that a test is being skipped. Would that be useful here?

hwillson: Thanks for the suggestion @benjamn - I originally logged a "Skipped" message, but decided to take it out when I saw how it looked alongside the final `self-test` summary message. I'll take another pass at it though - thanks!

hwillson: Just as an example, this is how it looked. 

screenshot 2018-01-26 20 22 41

Because `skip` is called before `define`, the message gets logged when all of the tests are first attempted to be defined. So even when using a `self-test` exclude pattern, all `.skip` configured tests are still dumped to the console. Maybe that's okay - I just thought it looked a bit odd ...

hwillson: I've added a skipped test message that looks like this:

```
...
Ignoring "old cli tests (converted)" test (excluded by selftest.skip)
Ignoring "compiler plugins - addAssets" test (excluded by selftest.skip)
...
```

benjamn: Yeah, I guess I was thinking we should only log the skips if those tests would have been run. If that's not easy, this isn't essential.

hwillson:
@benjamn I've pushed some tweaks to make this fall more in-line with the existing self-test summary. How about this:

screenshot 2018-01-26 21 37 56

So we don't show the name of the manually ignored tests, but we do list a manually ignored count (and mention that they were ignored by selftest.skip). This lines up with the approach used for the other skip tags, like "non-matching" - it doesn't show the names of the non-matching tests, just a count.

Let me know if this works - thanks!

.circleci/config.yml
```diff
@@ -91,13 +91,10 @@ build_machine_environment: &build_machine_environment
...
-    # For readability, this is a regex wrapped across multiple lines in quotes.
-    SELF_TEST_EXCLUDE: "\
-        ^old cli tests|\
-        ^minifiers can't register non-js|\
-        ^minifiers: apps can't use|\
-        ^compiler plugins - addAssets\
-        "
+    # If needed, for readability this should be a regex wrapped across
+    # multiple lines in quotes.
+    SELF_TEST_EXCLUDE: ""
```
abernix: Can we get rid of this completely now?

appveyor.yml
```diff
@@ -15,12 +15,7 @@ environment:
   TOOL_NODE_FLAGS: --expose-gc
   TIMEOUT_SCALE_FACTOR: 8
   METEOR_HEADLESS: true
-  SELF_TEST_EXCLUDE: "\
-      ^old cli tests|\
-      ^minifiers can't register non-js|\
-      ^minifiers: apps can't use|\
-      ^compiler plugins - addAssets|\
-      ^NULL-LEAVE-THIS-HERE-NULL$"
+  SELF_TEST_EXCLUDE: "^NULL-LEAVE-THIS-HERE-NULL$"
```
abernix: Same question about removing here.

(The `^NULL-LEAVE-THIS-HERE-NULL$` was, IIRC, to prevent odd command line behavior exhibited when no tests were defined).

.circleci/config.yml
hwillson: I left it there in-case we wanted to add one-off excludes for specific CI configs, like maybe if there were some excludes we needed for Circle but not AppVeyor. Happy to remove them if you don't think this is needed though!

appveyor.yml
hwillson: Same as above comment - I guessed that was the reason for `NULL-LEAVE-THIS-HERE-NULL`, so I left it in place. Thanks!

abernix:
Works for me!  Thanks for this great improvement!  Now I can clear my local environment variables, of the same form, too. :wink:

PR #9581, opened 2018-01-27
add docs for Meteor.apply options.returnStubValue

pravdomil:
see [diff](9581/files)


abernix:
Thanks, @pravdomil!  This won't be live on [docs.meteor.com](http://docs.meteor.com/api/methods.html#Meteor-apply) right away, but it will work its way there when we release Meteor 1.6.2.

PR #9596, opened 2018-01-29
Remove underscore from meteor-base

jamesmillerburgess:
This should shave down bundle sizes by 14.4 kb for many non-blaze projects.

The other core meteor packages have not depended on `underscore` since #9362. However, we are only able to remove this last dependency now due to the the first commit in this PR, which eliminates usages of `underscore` from apps that did not have the package listed in their `packages` files. This was causing CI test failures that now should be corrected.

Any meteor apps currently using `_` without `underscore` listed in their `packages` file will need to add the package explicitly.

`meteor-base` version number bumped from 1.3.0 to 1.3.1.

#### Adjustments to test apps:
There are only a few uses of `underscore` in these apps, and two of them actually used `underscore` without having it explicitly listed in their `packages` file.

This is a problem, because the apps were relying on the dependency from `meteor-base`, which we want to remove to cut down bundle sizes.

For the `modules` test app, I've added `underscore` to the `packages` file, because it is asserting that `require('meteor/underscore')._` equals `_`. For the other app and all other uses of `_`, rather than add `underscore` to the `packages` files, I took the modernization route and replaced the functions with their ES6 equivalents, and then removed `underscore` from all `packages` files.

### DIFF
--- a/packages/meteor-base/package.js
+++ b/packages/meteor-base/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'meteor-base',
-  version: '1.3.0',
+  version: '1.4.0',
   // Brief, one-line summary of the package.
   summary: 'Packages that every Meteor app needs',
   // By default, Meteor will default to using README.md for documentation.
@@ -17,9 +17,6 @@ Package.onUse(function(api) {
     // only supports building client/server web applications so this is not
     // removable
     'webapp',
-
-    // Most Meteor core packages depend on Underscore right now
-    'underscore',
 
     // The protocol and client/server libraries that Meteor uses to send data
     'ddp',
--- a/tools/tests/apps/failover-test/server/failover-test.js
+++ b/tools/tests/apps/failover-test/server/failover-test.js
@@ -69,7 +69,7 @@ C.find().observeChanges({
       Meteor.clearTimeout(nextStepTimeout);
       nextStepTimeout = null;
     }
-    if (!fields.step && _.has(steps, fields.step)) {
+    if (!fields.step && Object.prototoype.hasOwnProperty.call(steps, fields.step)) {
       console.log('Unexpected step:', fields.step);
       process.exit(1);
     }
--- a/History.md
+++ b/History.md
@@ -15,6 +15,16 @@
   `meteor-babel` now runs with the `loose:true` option, as required by
   other (optional) plugins like `@babel/plugin-proposal-decorators`.
   [Issue #9628](https://github.com/meteor/meteor/issues/9628)
+  
+* The `underscore` package has been removed as a dependency from `meteor-base`.
+  This opens up the possibility of removing 14.4 kb from production bundles.
+  Since this would be a breaking change for any apps may have been using `_` 
+  without having any packages that depend on `underscore` besides `meteor-base`,
+  we have added an upgrader that will automatically add `underscore` to the
+  `.meteor/packages` file of any project which lists `meteor-base`, but not
+  `underscore`. Apps which do not require this package can safely remove it
+  using `meteor remove underscore`.
+  [PR #9596](https://github.com/meteor/meteor/pull/9596)
 
 ## v1.6.1, 2018-01-19
 
--- a/History.md
+++ b/History.md
@@ -18,12 +18,12 @@
   
 * The `underscore` package has been removed as a dependency from `meteor-base`.
   This opens up the possibility of removing 14.4 kb from production bundles.
-  Since this would be a breaking change for any apps may have been using `_` 
-  without having any packages that depend on `underscore` besides `meteor-base`,
-  we have added an upgrader that will automatically add `underscore` to the
-  `.meteor/packages` file of any project which lists `meteor-base`, but not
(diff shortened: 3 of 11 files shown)


benjamn:
@hwillson The `passwords - tokens` test failure reproduces for me locally, and bisects to the Mongo 3.6 upgrade. I think the new version of Mongo fires oplog notifications with slightly different timing, and thus [this comment](https://github.com/meteor/meteor/blob/17eaa5bd6f9f4837c6c22001c30bfd65e5d28496/packages/accounts-password/password_tests.js#L957-L959) becomes relevant.

hwillson:
Ouch, good catch @benjamn - I'll take a look. Thanks!

jamesmillerburgess:
I've fixed the other error on the original PR - the `custom-minifier` app was missing the `ecmascript` package. I've also added an entry to History.md, and I think the AppVeyor failure is due to me force-pushing these changes.

abernix:
First, thank you so much for taking the time to work on great improvements like this.

AppVeyor won't build if there are any merge conflicts with the target branch as they build against a virtual GitHub branch which is meant to represent the "merge result", in an effort to try to provide early warning of what _might_ eventually be a build a broken build.  Currently, that merge conflict is just `History.md` so if you rebase against `devel` you might have luck fixing that.  Probably not necessary to worry about right now given that you're not making any Windows-specific changes.

Lastly, thanks again _for taking the time to work on great improvements like this_. 😉🙏👏

jamesmillerburgess:
@abernix Thanks for going out of your way to help contributors like me!

jamesmillerburgess:
@hwillson 

I agree with you that it isn't nice to introduce breaking changes without some kind of deprecation warning, although I can't think of an elegant way to do that for this particular situation:

1) An app is using `_` specifically for underscore functionality,
2) the *only* meteor package it is using that depends on `underscore` is `meteor-base`, and
3) they also are not importing an npm package like lodash, underscore, or something else as `_` for the same purpose

Besides being difficult to detect for a deprecation warning, that's a pretty odd set of criteria, so I'm not even sure there are non-trivial apps out there that meet them. Hopefully that means the blast radius will be quite small if we go forward with the change.

hwillson:
I agree @jamesmillerburgess, I couldn't think of a better way of handling this either. I guess I'm just envisioning complaints rolling in from people who are used to just running `meteor create my-happy-app` and assuming they can still use `underscore` without doing anything additional, because once upon a time this was the accepted way of doing things (by leveraging the `_` global). Even though this approach is no longer recommended, people can still do it (and ignore progress), until we merge this PR.

That being said, I like progress, so my vote goes towards exactly what you've done - mention this in the `History.md`. We can also highlight this in a Meteor 1.6.3 release blog post, and on the forums / in the Guide if necessary. I just thought I'd mention this so we're all on the same page, and are ready with answers in-case (a hopefully very small group of) people are upset.

Okay, enough rambling on my part - approved! 🙂

tools/tests/apps/failover-test/server/failover-test.js
```diff
@@ -69,7 +69,7 @@ C.find().observeChanges({
       Meteor.clearTimeout(nextStepTimeout);
       nextStepTimeout = null;
     }
-    if (!fields.step && _.has(steps, fields.step)) {
+    if (!fields.step && steps.hasOwnProperty(fields.step)) {
```
benjamn: Generally we prefer `Object.prototype.hasOwnProperty.call(steps, fields.step)`, but since this is a test and it passes, I think this is fine. The rationale is that `hasOwnProperty` could be a field in the `steps` object, which would mean `steps.hasOwnProperty !== Object.prototype.hasOwnProperty`.

packages/meteor-base/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'meteor-base',
-  version: '1.3.0',
+  version: '1.3.1',
```
benjamn: Let's bump this to 1.4.0 so that the changes won't take effect until developers update to Meteor 1.6.2. This kind of change is appropriate for 1.6.2, I think, because 1.6.2 is all about various bundle size optimizations.

tools/tests/apps/failover-test/server/failover-test.js
jamesmillerburgess: Ah, I was wondering why I was seeing it used like that elsewhere. I will update my other PRs as well. Thanks!

hwillson:
Just to add to https://github.com/meteor/meteor/pull/9596#pullrequestreview-95111849, if you don't have time to look into writing the upgrader, let me know and I'll dig into it. Thanks!

jamesmillerburgess:
I should be able to address the feedback for this and the other PRs either by tonight or tomorrow early morning. But I’ll let you know if I don’t have time and need some help. Thanks!

Best regards,

James
________________________________

jamesmillerburgess:
I tested the upgrader with a simple app and it seemed to do what it was supposed to for each of the conditions. I also adjusted the entry in `History.md` to reflect the presence of the upgrader and the implications for developers.

PR #9603, opened 2018-01-30
Modernize random

sebakerckhof:
Note: I'm not really sure why there was that IIFE in the Alea function. So I might have missed something by taking that out?


packages/random/random.js
```diff
@@ -7,20 +7,25 @@
 // window.crypto.getRandomValues() or alea, the primitive is fraction and we use
 // that to construct hex string.
 
-if (Meteor.isServer)
-  var nodeCrypto = Npm.require('crypto');
+import { Meteor } from 'meteor/meteor';
+
+let nodeCrypto;
+if (Meteor.isServer) {
+  import crypto from 'crypto';
```
benjamn: Unfortunately, I think this will cause the `crypto` stub (which is very large) to be included in the client JS bundle, since `import` declarations register bundle dependencies but `Npm.require` is ignored by the dependency scanner (it just has to work on the server at runtime).

One possible solution:
```js
if (Meteor.isServer) {
  const cryptoId = "crypto";
  nodeCrypto = require(cryptoId);
}
```

sebakerckhof: Ok, I've done it like that for now, but it seems kinda hacky. What about having a separate main module for client and server that inherit from an abstract RandomGenerator class? Could do away with some conditionals in the methods too.

packages/random/random.js
```diff
@@ -35,214 +39,205 @@ var Alea = function () {
...
+  /**
+   * @name Random.fraction
+   * @summary Return a number between 0 and 1, like `Math.random`.
+   * @locus Anywhere
+   */
+  fraction () {
+    switch (this.type) {
+    case RandomGenerator.Type.ALEA:
+      return this.alea();
+    case RandomGenerator.Type.NODE_CRYPTO: {
```
abernix: Let's not go with the braces around the `case`es here, as it's not a style we use anywhere else in the Meteor codebase.

(and, IMHO, a `case` should never be enough lines that it warrants braces).

packages/random/random.js
```diff
@@ -35,214 +39,205 @@ var Alea = function () {
...
+      } catch (e) {
+        // XXX should re-throw any error except insufficient entropy
+        bytes = nodeCrypto.pseudoRandomBytes(numBytes);
+      }
+      const result = bytes.toString('hex');
+      // If the number of digits is odd, we'll have generated an extra 4 bits
+      // of randomness, so we need to trim the last digit.
+      return result.substring(0, digits);
+    } else {
+      return this._randomString(digits, '0123456789abcdef');
```
abernix: I realize this isn't a result of your change, however, can we either:

1. Negate the opening conditional (`if (this.type !== RandomGenerator.Type.NODE_CRYPTO)`) and have this `return this._randomString` live inside that (freeing the rest of the block to _not_ be within the `if`?; or
2. Get rid of this `else` and just let the `return` live on its own outside the conditional braces?

Based on the size of the `if`, it seems like number 1 _may_ increase readability by reducing nesting.

packages/random/random.js
```diff
@@ -35,214 +39,205 @@ var Alea = function () {
...
-    random.args = args;
-    return random;
+  const random = () => {
+    const t = (2091639 * s0) + (c * 2.3283064365386963e-10); // 2^-32
+    s0 = s1;
+    s1 = s2;
+    return s2 = t - (c = t | 0);
+  };
 
-  } (Array.prototype.slice.call(arguments)));
```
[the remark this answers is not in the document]
sebakerckhof: Wondering the same thing. Can't think of any reason but that it was copy pasted from an over-cautious author in an era before block scoped declarations. Unfortunately the original link it refers to in the comment is dead.

packages/random/random.js
sebakerckhof: This wasn't done because of the multi lines, but because those lines have lexical declarations, which may have unwanted hoisting effects. I.e. this eslint rule: https://eslint.org/docs/rules/no-case-declarations which is both in eslint:recommended and the airbnb style guides.

But in this specific case the brackets add no value, so happy to remove them.

packages/random/random.js
```diff
@@ -35,214 +39,205 @@ var Alea = function () {
...
+      // If the number of digits is odd, we'll have generated an extra 4 bits
+      // of randomness, so we need to trim the last digit.
+      return result.substring(0, digits);
+    } else {
+      return this._randomString(digits, '0123456789abcdef');
+    }
+  }
+
+  _randomString (charsCount, alphabet) {
+    return [...Array(charsCount)].map(this.choice.bind(this, alphabet)).join('');
```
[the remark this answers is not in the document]
sebakerckhof: Ok, although this would only get called for Mongo ids on the client side, I agree it might be case of wanting to use the new stuff without much reason. I'll change it to what you suggest.

packages/random/random.js
sebakerckhof: @abernix Ok, but what do you think about splitting it up in subclasses instead of doing these conditional checks everywhere. I can then also do away with the weird require at the top.

abernix: I could see that as being valuable.  Do you have any concerns with such a pattern?

sebakerckhof:
@abernix what do you think about the current state, where I refactored into subclasses?

StorytellerCZ:
@abernix Any updates on this?

aldeed:
FYI, I made a similar port to an isomorphic NPM package earlier this year, with tests converted to Jest: https://github.com/reactioncommerce/random

afrokick:
@filipenevola any chance to merge it?

filipenevola:
@sebakerckhof could you open a new PR including the same content to see if the tests run accordingly? You could open the new one as branch in this repo instead of in a fork.

Issue #9605, opened 2018-01-30
Latest meteor/jquery package vulnerable to XXS


ryanmtaylor:
`meteor/jquery` package gives jQuery version 1.11.2 – the 1.x branch is no longer maintained and 1.12.1 is the latest version not vulnerable to the following XXS bug: https://snyk.io/vuln/npm:jquery:20150627

The bug was fixed in 1.12.0, then there was an update in 1.12.1, but unfortunately the bug fix was rolled back in 1.12.2 meaning versions 1.12.2-1.12.4 (latest 1.x) are also vulnerable.

Can somebody update the `meteor/jquery` package to simply use jQuery 1.12.1?

More info: https://github.com/jquery/jquery/issues/2432

benjamn:
Just a note that you can `meteor npm install jquery@1.12.1` (or any version for that matter) in your application's `node_modules` directory, and the `meteor/jquery` package will [use that instead](https://github.com/meteor/meteor/blob/17eaa5bd6f9f4837c6c22001c30bfd65e5d28496/packages/jquery/main.js#L4).

PR #9607, opened 2018-01-30
Move Meteor jquery package into packages/non-core.

benjamn:
This move will mean the next release of Meteor will not impose any constraints on the version of the Meteor `jquery` package.

Related: #9605


GeoffreyBooth:
Here here! Will we finally be able to use jQuery via the Google CDN? Or will jQuery keep getting installed as a dependency of other packages, with no way for me to force its exclusion?

benjamn:
@GeoffreyBooth That's a great question. One solution would be to create your own local `jquery` Meteor package, which would take precedence over the version depended on by other packages. You could even make it re-export the CDN-installed `$` and `jQuery` variables, just in case you're using packages that import those variables explicitly from `meteor/jquery`.

GeoffreyBooth:
I think the issue is more that the CDN-hosted jQuery needs to be loaded first, even before packages: https://forums.meteor.com/t/serve-jquery-from-cdn/20176/9

This is perhaps a separate issue. But it would be nice to solve it somehow, so that common libraries like jQuery and others can be loaded before Meteor’s script bundle.

hwillson:
@GeoffreyBooth See https://github.com/meteor/meteor-feature-requests/issues/24 (it's `pull-requests-encouraged` if you're interested ... 🙂).

PR #9608, opened 2018-01-30
Modernize base64 package

jamesmillerburgess:
In continuing with the modernization process, I've updated the syntax for the next package in the list, alphabetically.

### DIFF
--- a/packages/base64/base64.js
+++ b/packages/base64/base64.js
@@ -2,7 +2,7 @@
 
 const BASE_64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
-const BASE_64_VALS = {};
+const BASE_64_VALS = Object.create(null);
 
 const getChar = val => BASE_64_CHARS.charAt(val);
 const getVal = ch => ch === '=' ? -1 : BASE_64_VALS[ch];
@@ -21,6 +21,7 @@ const encode = array => {
         throw new Error(
           "Not ascii. Base64.encode can only take ascii strings.");
       }
+
       array[i] = ch;
     }
   }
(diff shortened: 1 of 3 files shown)


packages/base64/base64.js
```diff
@@ -95,24 +87,24 @@ Base64.newBinary = function (len) {
   return new Uint8Array(new ArrayBuffer(len));
 };
 
-Base64.decode = function (str) {
-  var len = Math.floor((str.length*3)/4);
+const decode = str => {
+  let len = Math.floor((str.length * 3) / 4);
   if (str.charAt(str.length - 1) == '=') {
     len--;
     if (str.charAt(str.length - 2) == '=')
```
abernix: Can we add braces `{}here?

packages/base64/base64_test.js
```diff
@@ -10,50 +12,45 @@ var asciiToArray = function (str) {
...
-var arrayToAscii = function (arr) {
-  var res = [];
-  for (var i = 0; i < arr.length; i++) {
-    res.push(String.fromCharCode(arr[i]));
-  }
-  return res.join("");
-};
+const arrayToAscii = arr =>
+  arr.reduce(
+    (prev, charCode) => [...prev, String.fromCharCode(charCode)], []
```
abernix: Not the biggest deal since this is in tests, so I guess this is more me expressing preference: I personally don't find this notation easier to understand than the original, and in my experience this would be much less performant than the previous implementation (the `for` loop), especially since a new `Array` is being initialized `[ ... ]` and having its values copied into it (via spread) on each iteration.

A similar notation, which should perform significantly better, would be `push`-ing onto the Array and returning it, which would maintain and manipulate the reference to the previous object, rather than allocating a a new one:

```js
const arrayToAscii = arr => arr
  .reduce((prev, charCode) =>
    prev.push(String.fromCharCode(charCode)) && prev, [])
  .join('');
```

packages/base64/base64.js
```diff
@@ -1,22 +1,22 @@
 // Base 64 encoding
 
-var BASE_64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+const BASE_64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
-var BASE_64_VALS = {};
+const BASE_64_VALS = {};
```
abernix: I don't think it's relevant here (and I know you didn't write the original) but, in "modernization", I have a strong preference for switching to `Object.create(null)` (rather than `{}`) to initiate empty objects which will be used as lookups from user-input since `Object`s created with `{}` (object initializer notation) inherit (non-"own") attributes from the global `Object.prototype`, such as `hasOwnProperty`, `valueOf`, etc.

In this case, the user-input appears to be limited to single characters at a time so `hasOwnProperty` would be out of the question, but if the global `Object`s prototype was manipulated elsewhere, values that we didn't expect could suddenly become part of this `BASE_64_VALS` hash:

```js
// Rogue code?
Object.prototype['-'] = 45;

// This code.
const BASE_64_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
const BASE_64_VALS = {};
const getVal = ch => ch === '=' ? -1 : BASE_64_VALS[ch];

// Use of `getVal`
getVal("-"); // returns 45;
```

Though I'm not sure if that would _actually_ break `Base64.decode` below...

jamesmillerburgess: I will adjust that and keep it in mind for other PRs!

packages/base64/base64_test.js
jamesmillerburgess: Good point!

packages/base64/base64.js
jamesmillerburgess: What is this, 2014? 😂

PR #9620, opened 2018-02-01
Modernize browser-policy packages

jamesmillerburgess:
This PR modernizes the following packages:
- `browser-policy`
- `browser-policy-common`
- `browser-policy-content`
- `browser-policy-framing`

### DIFF
--- a/packages/browser-policy-content/browser-policy-content.js
+++ b/packages/browser-policy-content/browser-policy-content.js
@@ -60,6 +60,7 @@ const parseCsp = csp => {
     if (policy[policy.length - 1] === ";") {
       policy = policy.substring(0, policy.length - 1);
     }
+
     const srcs = policy.split(" ");
     const directive = srcs[0];
     if (srcs.includes(keywords.none)) {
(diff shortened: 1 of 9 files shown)


packages/browser-policy-content/browser-policy-content.js
```diff
@@ -143,104 +149,103 @@ var setDefaultPolicy = function () {
...
-_.extend(BrowserPolicy.content, {
-  allowContentTypeSniffing: function () {
-    contentSniffingAllowed = true;
-  },
+Object.assign(BrowserPolicy.content, {
+  allowContentTypeSniffing: () =>contentSniffingAllowed = true,
   // Exported for tests and browser-policy-common.
-  _constructCsp: function () {
-    if (! cspSrcs || _.isEmpty(cspSrcs))
+  _constructCsp: () => {
```
hwillson: Definitely not a show stopper, but your updated function declarations could be "modernized" further using ES2015's shorter [method definition syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions).

benjamn: I'm going to second this: please use method syntax rather than arrow functions here.

jamesmillerburgess: Sure - I was defaulting to arrow functions whenever `this` wasn't used in the body, but going forward I'll prioritize method syntax.

hwillson:
Hi @jamesmillerburgess - first off, thanks for your work on this! Your time spent on this is very much appreciated. We've been bouncing around the idea of deprecating the `browser-policy-*` packages however, since there's not really any advantage to having these packages be part of core. They've fallen way behind other 3rd party CSP packages, that integrate with Meteor quite easily, and provide all of the same functionality (and then some - like [Helmet](https://helmetjs.github.io)). As of last week, we've decided to pull the trigger and start the `browser-policy-*` deprecation process. That means we're going to close this PR, since we'll be leaving the current BP versions as the final versions. Again, thanks very much for your time on this - please don't let this discourage you from future contributions! 🙂

jamesmillerburgess:
No problem!

PR #9626, opened 2018-02-02
Minifier js report used minifier

steventebrinke:
Since I was having trouble with the minifier (part of our codebase minifies correctly only with uglify-es, part only with babeli), I updated the minifier-js package to report which backend was used for minification to ease debugging.


packages/minifier-js/minifier.js
```diff
@@ -29,14 +30,11 @@ meteorJsMinify = function (source) {
...
-      });
-      result.code = Babel.minify(source, options).code;
-    } else {
-      result.code = Babel.minify(source).code;
-    }
+    var options = Babel.getMinifierOptions({
+      inlineNodeEnv: NODE_ENV
+    });
+    result.code = Babel.minify(source, options).code;
+    result.minifier = 'babeli';
```
klaussner: [The official name of the minifier is `babel-minify`.](https://github.com/babel/minify#note) :slightly_smiling_face:

packages/minifier-js/minifier.js
```diff
@@ -29,14 +30,11 @@ meteorJsMinify = function (source) {
     // Although Babel.minify can handle a wider variety of ECMAScript
     // 2015+ syntax, it is substantially slower than UglifyJS, so we use
     // it only as a fallback.
-    if (Babel.getMinifierOptions) {
```
klaussner: Why did you remove this check?

packages/minifier-js/minifier.js
steventebrinke: Ah, it is very often called babeli, but sticking to the official name would be better indeed.

packages/minifier-js/minifier.js
steventebrinke: Since getMinifierOptions is supported since meteor-babel 0.16, this check seems very outdated. I could leave it in, but then I would need some way of communicating back whether options are passed to Babel or not. IMO, babel-minify must always be configured the same as uglify, because otherwise using the fallback can cause unexpected failures that don't happen when the fallback isn't used. Therefore, I think that the options must always be passed to babel-minify.

packages/minifier-js/minifier.js
```diff
@@ -29,14 +30,11 @@ meteorJsMinify = function (source) {
...
-      });
-      result.code = Babel.minify(source, options).code;
-    } else {
-      result.code = Babel.minify(source).code;
-    }
+    var options = Babel.getMinifierOptions({
+      inlineNodeEnv: NODE_ENV
+    });
+    result.code = Babel.minify(source, options).code;
+    result.minifier = 'babel-minify';
```
abernix: You're setting `result.minifier`, but where are you actually reporting the minifier used?

steventebrinke: The result object is returned to the standard-minifier-js (or any other plugin using minifier-js). So I use it in a fork of standard-minifier-js (printing it in a not really user friendly way). Since standard-minifier-js generates statistics, it might be useful to also generate statistics about the used minifier. However, I do not know in which way it should be reported to be useful for most users. If you have any suggestions, I might give it a try to add such statistics to standard-minifier-js.

abernix:
@steventebrinke This is published in `minifier-js@2.3.3`, if you want to give it a shot in your own minifier!

PR #9627, opened 2018-02-02
Faster CircleCI workflow

jamesmillerburgess:
This PR aims to reduce the TTGC (time to green check?) on the CircleCI workflow. The last build on `devel` clocked in at [45:48](https://circleci.com/workflow-run/1f129dbc-2189-4ad2-99d0-804205cb038c). I'd like to get this to sub 20:00.

Here are some techniques/ideas for improving the speed:
- [x] Rebalance workflow groups
Currently Group 2 is taking roughly twice as long as the others

- [x] Make flakey tests more reliable to reduce retries
`javascript hot code push` in particular tends to have to retry at least once each run.

- [x] Expand number of groups to 12 to make use of all available containers

- ~~Add a time-limited polling-match function to self-test 
I don't know if this is feasible, but right now we wait a fixed number of seconds before running `match`, so there is dead time we could cut down by trying a `match` every second or two and only failing if it exceeds the time limit. This could also have a start-from time so we don't poll if we know the output is going to take at least X seconds to come.~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,8 +173,20 @@ jobs:
               --retries ${METEOR_SELF_TEST_RETRIES} \
               --exclude "${SELF_TEST_EXCLUDE}" \
               --headless \
-              --junit /tmp/results/junit/0.xml \
+              --junit /tmp/results/junit/0-1.xml \
               --with-tag "custom-warehouse"
+          no_output_timeout: 20m
+      - run:
+          name: "Running self-test (0): A-Comm"
+          command: |
+            eval $PRE_TEST_COMMANDS;
+            ./meteor self-test \
+              --retries ${METEOR_SELF_TEST_RETRIES} \
+              --exclude "${SELF_TEST_EXCLUDE}" \
+              --headless \
+              --junit /tmp/results/junit/0-2.xml \
+              --file '^[a-b]|^c[a-n]|^co[a-l]|^comm' \
+              --without-tag "custom-warehouse"
           no_output_timeout: 20m
       - run:
           <<: *run_save_node_bin
@@ -203,7 +215,7 @@ jobs:
           name: "Print environment"
           command: printenv
       - run:
-          name: "Running self-test (1): A-Com"
+          name: "Running self-test (1): Comn-Comz"
           command: |
             eval $PRE_TEST_COMMANDS;
             ./meteor self-test \
@@ -211,7 +223,7 @@ jobs:
               --exclude "${SELF_TEST_EXCLUDE}" \
               --headless \
               --junit /tmp/results/junit/1.xml \
-              --file '^[a-b]|^c[a-n]|^co[a-l]|^compiler-plugins' \
+              --file '^com[n-z]' \
               --without-tag "custom-warehouse"
           no_output_timeout: 20m
       - run:
@@ -241,7 +253,7 @@ jobs:
           name: "Print environment"
           command: printenv
       - run:
-          name: "Running self-test (2): Con-K"
+          name: "Running self-test (2): Con-Coz"
           command: |
             eval $PRE_TEST_COMMANDS;
             ./meteor self-test \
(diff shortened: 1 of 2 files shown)


abernix:
TTGC! ✅ (aka `:white_check_mark:` emoji ??? 🤔)

Test improvements are always welcome.  We had done this a while back as well, but it seems like it's crept up a bit more again.  Our median build time, according to CircleCI Insights is just above 11 minutes.  But there is the standout container that takes a while so re-balancing is a great idea.  It may also be necessary to decompose some individual tests and move them into other groups.



A few things:

* We actually have 4 additional containers available to us on CircleCI that we could utilize to speed things up.  This is mainly a matter of creating new groups in the `.circleci/config.yml` and, most importantly, making sure specific tests aren't getting left out of our regular expression ranges. 😨 
* I upgraded us to CircleCI 2.0 (#8985) and at the same time added JUnit test output (which includes timing information) to our builds, with the intention of using it to balance out the tests in a more exact way.  I had hoped to make it fully automatic balancing based on averages too, but never found time to revisit it.  You can see the test timings in the [Artifacts](https://circleci.com/gh/meteor/meteor/15351#artifacts/containers/0) tab for each build on `meteor/meteor` (for example, https://15351-3214406-gh.circle-artifacts.com/0/tmp/results/junit/5.xml)
  * According to [this page](https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-patterns), the timing information is supposedly available to assist with auto-balancing of tests at the file path of `$CIRCLE_INTERNAL_TASK_DATA/circle-test-results`. 
  * It's possible that to do such a thing, the `meteor self-test` tool needs to accept a list of tests to run as an input.
  * I don't believe we actually delay checking for `match`es.  I believe the [`Matcher`](https://github.com/meteor/meteor/blob/devel/tools/tool-testing/matcher.js) starts watching `STDERR`/`STDOUT` immediately, and uses the default `baseTimeout` (20 seconds) as an upper bounds of time to wait and `waitSecs` is simply a way to raise that default `baseTimeout` by additional seconds.  Though I might not be remembering correctly!

These are maybe considerations for another PR, but I think if you'd be willing to, at the very least, change this PR to use additional containers, that would probably be super welcome! Of course, that won't help in your own CircleCI, unless you're on a paid plan, but you should at least consider raising your parallelism/concurrency if you haven't already (since I believe 4 are free on CircleCI for OSS?).

jamesmillerburgess:
Thanks for the detailed feedback!

I will focus on expanding the groups and tweaking the regex ranges for this PR to try to get the maximum group build time as low as possible. This is probably the biggest low-hanging fruit, so it is worth getting that merged before further optimization.

jamesmillerburgess:
This is balanced about as well as possible while still keeping the regex groups in alphabetical order and not splitting up any of the test files. There is also a bit of variability in the timings. The exact same code on two different branches produced these [two](https://circleci.com/workflow-run/bac12155-7fe1-4241-8906-b4769b37ee6c) [results](https://circleci.com/workflow-run/e1804f02-dc21-47e9-8854-469b53913882), for instance:

```
         | branch a | branch b
Group 0  |    15:04 |    17:37
Group 1  |    09:17 |    08:02
Group 2  |    06:39 |    10:14
Group 3  |    09:52 |    09:27
Group 4  |    05:29 |    07:53
Group 5  |    06:35 |    05:11
Group 6  |    06:39 |    05:11
Group 7  |    07:30 |    06:56
Group 8  |    08:28 |    08:14
Group 9  |    14:58 |    11:48
Group 10 |    04:35 |    04:53
Group 11 |    07:47 |    07:30
```

I think the next step would probably be to split up some of the longer-running test files like `run.js`, but I what we have so far should be good enough for a first PR.

It looks like the net effect of the changes here is taking TTGC from 45:00 to 35:00 for accounts with four free containers, and taking it from 45:00 to 20:00 for accounts that can run all 12 containers at once.

PR #9636, opened 2018-02-04
Modernize logging

sebakerckhof:


hwillson:
Marking as a work in progress since the failing test seems related. Thanks!

sebakerckhof:
@hwillson I believe they fail because I use `padStart` and the travis tests are ran on node 4, where that is not available.

abernix:
@sebakerckhof Feel free to bump that server Node version in `.travis.yml` to `node_js: 8` (another PR! 🎉), though that error seems to be coming from the browser — in this case, (an older) PhantomJS which may (read: does) have a different set of capabilities.  Probably worth some additional investigation.

abernix:
Just saw your other PR which addresses this... (#9641)!

sebakerckhof:
Ok, the format test is fixed after rebasing with the `padStart` changes. There's another failure on travis now, but it looks unrelated.

packages/logging/package.js
```diff
@@ -17,7 +17,11 @@ Cordova.depends({
 
 Package.onUse(function (api) {
   api.export('Log');
-  api.use(['ejson', 'ecmascript']);
+  // The `ecmascript-runtime-client` package is explicitly depended upon
+  // here due to this package's dependency on
+  // `String.prototype.padLeft` which is polyfilled only in
```
hwillson: Super small typo - it depends on `padRight`.

packages/logging/logging.js
```diff
@@ -84,43 +88,46 @@ var logInBrowser = function (obj) {
...
+    return {};
+  }
 
   // looking for the first line outside the logging package (or an
   // eval if we find that first)
-  var line;
-  for (var i = 1; i < lines.length; ++i) {
-    line = lines[i];
+  let line;
+  const [_, ...lines] = stack.split('\n');
```
hwillson: The use of `_` here threw me off for a second when reviewing. Since this is intended to hold the ignored first line, maybe we should just call it `firstLine` or `ignoredFirstLine`.

sebakerckhof: @hwillson in more functional languages like haskell underscore is used as a kind of "don't care" symbol. But I'll change it since in javascript it's a bit weird since the underscore symbol is already used for lots of other things. I'll just use a slice, so we're not making variables that are not getting used.

hwillson: Awesome, thanks @sebakerckhof! I have a Haskell book sitting beside me that I've been planning on cracking open for 8 months now ... *sigh*.

abernix: For what it's worth, you can leave out the `_` entirely, and the first element will be thrown away.  Take that, functional 💥:

```js
const [ , ...lines] = stack.split('\n');
```

PR #9637, opened 2018-02-04
Modernize reactive-var

sebakerckhof:


hwillson:
Marking as a work in progress @sebakerckhof (since those test failures look valid). Thanks!

sebakerckhof:
okay, fixed

sebakerckhof:
Ok, I agree that without the class this PR isn't really meaningful.

PR #9638, opened 2018-02-05
Modernize check package

jamesmillerburgess:
The next package in the list, alphabetically. I considered switching the `Match` export to an instance of a class, but ended up leaving it alone. If there is a good reason to switch it, then I can go back and do that!

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -50,13 +50,30 @@ export const check = (value, pattern) => {
  * @summary The namespace for all Match types and methods.
  */
 export const Match = {
-  Optional: pattern => new Optional(pattern),
-  Maybe: pattern => new Maybe(pattern),
-  OneOf: (...args) => new OneOf(args),
+  Optional: function(pattern) {
+    return new Optional(pattern);
+  },
+
+  Maybe: function(pattern) {
+    return new Maybe(pattern);
+  },
+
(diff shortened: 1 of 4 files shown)


packages/check/match.js
```diff
@@ -46,39 +50,30 @@ export const check = function (value, pattern) {
...
-  Where: function (condition) {
-    return new Where(condition);
-  },
-  ObjectIncluding: function (pattern) {
-    return new ObjectIncluding(pattern);
-  },
-  ObjectWithValues: function (pattern) {
-    return new ObjectWithValues(pattern);
-  },
+  Where: condition => new Where(condition),
```
benjamn: Unfortunately, existing code sometimes uses these properties as constructors (e.g. `new Match.Where`). Arrow functions don't have a `.prototype` and can't be called as a constructor, so I think we need to keep this code the way it was (check the blame for recent back-and-forth due to this problem). Method shorthand notation is off the table for the same reason.

jamesmillerburgess: Ahh, that explains it. 👍

mitar:
Should we also allow registering more checks? Currently I have a [package](https://github.com/peerlibrary/meteor-check-extension) which adds various common checks by just attaching it to `Match` object. Not sure if this will work once it will be modernized?

PR #9641, opened 2018-02-05
Add support for string padLeft/padRight

sebakerckhof:
I think this should also fix the failing tests for https://github.com/meteor/meteor/pull/9636


benjamn:
If we merge this, can you rebase #9636? Also, can you bump the `ecmascript-runtime-client` patch version (in `package.js`)?

PR #9656, opened 2018-02-09
Automate rebalancing of CircleCI test groups

jamesmillerburgess:
This PR builds on the work done to improve TTGC (time to green check) in #9627. In that PR, file-matching regular expressions were manually adjusted to try to balance the test groups. In this PR we optimize at the test level instead of the file level, and automate the process, readjusting after every build. Thanks for the suggestions @abernix!

This PR still has some gaps, but the stats look like they are going in the right direction:
* Automated balancing standard deviation: [123 seconds](https://circleci.com/workflow-run/440f8e4f-8b60-4306-986b-142f42671e3a)
* Manual balancing standard deviation: [182 seconds](https://circleci.com/workflow-run/f64f6df7-0820-4164-977f-68f243aa05ac)

Some notes:
- A new job now runs after all test groups finish. It runs a node app that reads the persisted test results from the current build, combining it with average test times from previous builds to calculate a close split between the groups
- Group tests are run according to the cache file generated by the app at the end of the previous build
- The number of groups to use and the number of builds to use in the running average are configurable in `.circleci/config.yml`
- The path for test results was switched to relative due to restrictions on workspace paths
- There are a series of fallbacks used to decide which tests to run on any given group: either 1) the optimized group from the cache, 2) a manually-determined set of tests because no cache was found, or 3) no tests at all because no tests are assigned to the group (for instance, group number 14 in the first build after the number of groups is increased from 12 to 16)

Outstanding todos:
- [x] Isolate tests that don't play well with others
- [x] Remove `wip-` prefix from cache keys
- [x] Test that the catch-all for new tests is working properly
- [x] Check that all tests are getting run (escape regexps properly)

Items for a subsequent PR:
 - Fix flakey tests, especially `package-tests: add debugOnly and prodOnly packages`
This test always fails when certain other tests run before it
Known conflicting tests: `autoupdate: autoupdate`, and `assets: assets - helper exists to unicode normalize path strings`
- Confirm that test timing is accurate
It looks like retries are not included in the test time, and maybe also tests that print `Testing with phantomjs` (need to check)
- Handle more edge cases: tests with the same name, number of groups is reduced since last build, invalid number of groups, ...?
- Switch away from super-long regular expressions to `meteor self-test` reading a list from a json file?
- Unit tests for `test-balancer` (mocha?)
- Exclude `Clean Up` job from GitHub green check
- Optimize fallback groups further (currently they are based on the file-level optimization and aren't balanced very well)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
             - test-groups-v1-{{ .Branch }}
             - test-groups-v1
       - run:
-          name: Create Test Results Directory If Missing
+          name: Create Test Results Directory
           command: |
             sudo mkdir -p ./tmp/results/junit
             sudo chmod a+rwx ./tmp/results/junit
@@ -688,7 +688,7 @@ jobs:
       - attach_workspace:
           at: .
       - run:
-          name: Create ./test-groups
+          name: Create Test Groups Directory
           command: |
             sudo mkdir -p ./tmp/test-groups
             sudo chmod a+rwx ./tmp/test-groups
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,10 +17,6 @@ run_env_change: &run_env_change
     sudo mkdir -p /tmp/core_dumps
     sudo chmod a+rwx /tmp/core_dumps
 
-    # Make a place for JUnit tests to live.
-    sudo mkdir -p /tmp/results/junit
-    sudo chmod -R a+rwx /tmp/results/
-
     # Bake the locale we expect into the image.
     echo "en_US.UTF-8 UTF-8" | sudo tee /etc/locale.gen
     sudo locale-gen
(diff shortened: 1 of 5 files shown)


hwillson:
This is awesome @jamesmillerburgess, thanks for your work on this! Let us know when you think this is ready for review (and if you think there is enough ready to go here that we could consider merging, then work on the outstanding todo's in a separate PR). Thanks again!

jamesmillerburgess:
I think that some of the todos can be split into a separate PR, but a few of them are must-have for this one.

I'll find some time to work on this in the next few days and hopefully can get it ready for a review!

jamesmillerburgess:
This is ready for review, and I think it is merge-able! 👍

jamesmillerburgess:
Thanks @hwillson!

Sorry to push another commit after you reviewed, but I found a little typo in a variable name!

PR #9657, opened 2018-02-10
Support <meteor-bundled-css /> pseudo-tag for controlling position of CSS bundle. (#24)

rcurrier666:
This PR is for feature request #24 **Order of merged-stylesheets.css in \<head>**. A new pseudo tag, `<meteor-bundled-css>`, has been created. When placed anywhere in the `<head>` of an app it will be replaced by the link to the bundled CSS in the output. If no tag is seen, then the bundle will be placed at the top of the `<head>` section as before so that this feature is backward compatible.

Two new tests have been created, one to verify that the tag is actually replaced by the bundle, and one to verify that the bundle is in the correct place. All old boilerplate tests are untouched (except for the addition of a flag to control whether the input HTML contains the tag) and continue to work (verifying that without the tag everything works as before).

[WIP question] Where should this be documented? I thought perhaps in the build section, though I don't see an obvious place to put it there. It also seems to be a non-obvious place for anyone to look. Or will it only appear in the release notes?

Otherwise, this PR should be good to go.


rcurrier666:
My unit tests succeeded but the one failure isn't any place that I touched:
```
S: tinytest - observeChanges - unordered - specific fields + selector on excluded fields : !!!!!!!!! FAIL !!!!!!!!!!!
```

axelvaindal:
In order to help tracking this PR in the corresponding feature request, here is the related link (as yours redirect to the 24th issue of the Meteor repository. 

[Meteor Feature Request #24](https://github.com/meteor/meteor-feature-requests/issues/24).

packages/boilerplate-generator/template-web.browser.js
```diff
@@ -6,7 +6,18 @@ export const headTemplate = ({
   bundledJsCssUrlRewriteHook,
   head,
   dynamicHead,
-}) => [
+}) => {
+  var headSections = head.split(/<meteor-bundled-css[^<>]*>/, 2);
+  // if (headSections.length > 1) {
```
hwillson: Please remove this commented code.

packages/boilerplate-generator/template-web.browser.js
```diff
@@ -15,17 +26,14 @@ export const headTemplate = ({
...
-
-  head,
+  (headSections.length === 1)
+    ? [cssBundle, headSections[0]].join('\n')
+    : [headSections[0], cssBundle, headSections[1]].join('\n'),
   dynamicHead,
   '</head>',
   '<body>',
 ].join('\n');
+};
```
hwillson: The indentation looks like a little off here - looks like lines 21 - 35 need to be indented.

packages/boilerplate-generator/template-web.browser.js
rcurrier666: Sorry, thought I had removed this before.

packages/boilerplate-generator/template-web.browser.js
rcurrier666: The entire "return [" block needed to be indented. I also added a couple of blank lines to make it easier to see which lines go together.

hwillson:
Thanks @rcurrier666 - looks great! I've finished an additional round of testing, and things are working well. We have a PR review meeting today and this will be discussed.

Issue #9549, opened 2018-01-17
1.6.1-rc6 Cordova  cordova-plugin-statusbar and cordova-plugin-splashscreen console messages


paulincai:
When I run ios-device with the Meteor default cordova setup (no extra plugins added), this message is displayed in the server console:

`WARNING: Attempting to install plugin cordova-plugin-statusbar@2.2.1, but it should have a minimum
         version of 2.3.0 to ensure compatibility with the current platform versions. Installing the
         minimum version for convenience, but you should adjust your dependencies.`

`
WARNING: Attempting to install plugin cordova-plugin-splashscreen@4.0.1, but it should have a minimum
         version of 4.1.0 to ensure compatibility with the current platform versions. Installing the
         minimum version for convenience, but you should adjust your dependencies.`

I personally add the updated versions from their respective repos.

benjamn:
I've seen these warnings too, and haven't been able to figure out where they might be coming from. The only versions you'll find in the Meteor codebase for those packages are 2.3.0 and 4.1.0, respectively.

paulincai:
https://github.com/meteor/meteor/pull/8913#issuecomment-325428788  :)

skirunman:
My guess is we are seeing these messages as we only updated to `cordova-plugin-statusbar@2.3.0` and `cordova-plugin-splashscreen@4.1.0` and not all the plugins that were released at the same time https://cordova.apache.org/news/2017/11/10/plugins-release.html

So likely some of our older pinned plugin versions have a dependency on the older versions. (just my educated guess)

skirunman:
I think we should check if this is solved with https://github.com/meteor/meteor/pull/10810

skirunman:
@filipenevola?

filipenevola:
I don't see these warning after updating all the plugins in 1.10. If you still see these please reopen.

You can try 1.10 updating to latest beta. Right now it is beta.4, meteor update - - release 1.10-beta.4

PR #9658, opened 2018-02-10
Need newer version for better Xcode9/iOS 11 Support

ryannairj:
This will bring better compatibility with newer versions of iOS

### DIFF
--- a/packages/mobile-status-bar/package.js
+++ b/packages/mobile-status-bar/package.js
@@ -4,5 +4,5 @@ Package.describe({
 });
 
 Cordova.depends({
-  'cordova-plugin-statusbar': '2.3.0'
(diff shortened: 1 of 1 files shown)


skirunman:
@ryannairj I would once again suggest that we need a more comprehensive solution to outdated Cordova plugins like https://github.com/meteor/meteor-feature-requests/issues/265 and that this PR will likely introduce other build time warnings https://github.com/meteor/meteor/issues/9549 

To use the latest plugin in your own project just do
`meteor add cordova:cordova-plugin-statusbar@2.4.1`

hwillson:
I completely agree with @skirunman here. Thanks for taking the initiative @ryannairj, but since there are other updates that need to be made at the same time (and this update alone could cause issues), I'll close this for now. Thanks again!

PR #9659, opened 2018-02-10
fix setting bcryptRounds using Accounts.config()

### LINKED ISSUE #9651: Unable to set bcryptRounds
It's currently impossible to change the value of `bcryptRounds` using `Accounts.config`.

The problem is in [password_server.js](https://github.com/meteor/meteor/blob/master/packages/accounts-password/password_server.js#L25), which gets evaluated before it's possible to actually call `Accounts.config`.

morloy:
This PR fixes #9651 by dynamically evaluating `Accounts._bcryptRounds`.


hwillson:
Thanks for submitting this PR @morloy. Do you think you'll have a chance to update the associated tests as well? They just need to have their `Accounts._bcryptRounds` references changed to `Accounts._bcryptRounds()`.

morloy:
@hwillson Thanks for the hint. Should be working now.

PR #9663, opened 2018-02-12
Add Promise.prototype.finally support

### LINKED ISSUE #9639: Promise.prototype.finally missing
Now that [`Promise.prototype.finally`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) is a part of the ECMAScript spec, it would be nice to be able to use it in my meteor applications as well. However, it is still annoyingly missing from Meteor 1.6.1 despite being present in `@babel/runtime/core-js`.

hwillson:
Adds `Promise.prototype.finally` support to Meteor's `promise` package. Let me know if I missed anything. Thanks!

Fixes #9639.

### DIFF
--- a/History.md
+++ b/History.md
@@ -19,6 +19,7 @@
 * Meteor's `promise` package has been updated to support
   [`Promise.prototype.finally`](https://github.com/tc39/proposal-promise-finally).
   [Issue 9639](https://github.com/meteor/meteor/issues/9639)
+  [PR #9663](https://github.com/meteor/meteor/pull/9663)
 
 ## v1.6.1, 2018-01-19
 
--- a/History.md
+++ b/History.md
@@ -16,6 +16,10 @@
   other (optional) plugins like `@babel/plugin-proposal-decorators`.
   [Issue #9628](https://github.com/meteor/meteor/issues/9628)
 
+* Meteor's `promise` package has been updated to support
+  [`Promise.prototype.finally`](https://github.com/tc39/proposal-promise-finally).
+  [Issue 9639](https://github.com/meteor/meteor/issues/9639)
+
 ## v1.6.1, 2018-01-19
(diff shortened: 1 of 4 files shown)


benjamn:
Looking at https://github.com/zloirock/core-js/blob/master/modules/es7.promise.finally.js for comparison, there's some extra logic to allow `Promise` subclasses to override the static `Promise.resolve` method. Essentially, instead of calling `exports.Promise.resolve(f())`, it would be better to call `this.constructor.resolve(f())`, since `this.constructor` might be a subclass that inherits from `Promise`.

The `core-js` implementation also checks if `this.constructor[Symbol.species]` is defined, just in case that should be used instead of `this.constructor`. [Here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/species)'s an explanation of why `Symbol.species` exists.

However, it strikes me that this is a lot of pedantry/pageantry just to call `Promise.resolve`, and there may be a way to rewrite this code so that it doesn't need to call `Promise.resolve` explicitly. That would probably be a worthwhile PR to `core-js`, too, if it works.

I'll push a commit in a few minutes if I can figure this out.

hwillson:
That would be awesome @benjamn! I was following the [`promise/lib/finally`](https://github.com/then/promise/blob/01cf802ed397d45cd959968e90792049a08b9db4/src/finally.js) approach, but the `core-js` implementation does seem safer. Let me know how far you get and if there's anything I can dive back into. Thanks!

Floriferous:
Does this mean it is only supported on meteor server side? I'm getting this error on MS Edge:`Object doesn't support property or method 'finally'`.

Even when using babel, I have to add a a polyfill for this to work everywhere?

PR #9665, opened 2018-02-13
Bump CoffeeScript to 2.2.1

GeoffreyBooth:
I also bumped the versions of its dependencies, and changed `ecmascript-runtime` to be implied on both client and server. @benjamn, was there a reason we were implying `ecmascript-runtime` only on the server? Don’t we want polyfills on the client as well?


benjamn:
Merged and published! Implying `ecmascript-runtime` on the client sounds like a good idea, and I don't recall any reason we didn't do that before.

PR #9666, opened 2018-02-13
Keep bundled /public and /private node_modules symlink content

### LINKED ISSUE #7013: [1.3.2.4] Can't have symlink to node_modules in /public
Example app to reproduce the issue: https://github.com/Zodiase/Meteor-Tests/tree/public-files-no-symlink.

This is a clean new app without any extra packages. A demo npm package and a symlink to it are added to reproduce the issue.

Issue reproduced with Meteor 1.3.2.4 on OSX El Capitan.
## Expected behavior
- `.meteor/local/build/programs/web.browser/app` should contain static files in `public`.
- I should be able to access the npm package files at, for example, `appurl/mdi/css/materialdesignicons.css`.
- I should be able to access the static file served in public folder at `appurl/foo.txt`.
## Actual behavior
- `.meteor/local/build/programs/web.browser/app` does not contain any static file in `public`.
- `appurl/mdi/css/materialdesignicons.css` brings up the app not the expected file.
- `appurl/foo.txt` also brings up the app not the expected file.

hwillson:
Many npm based packages provide supporting assets that need to be made available, when used on the web. For example, to use the `font-awesome` package properly, the `node_modules/font-awesome/fonts` files need to be made accessible to incoming web requests. 

With Meteor, an easy way to handle this would be to create a symlink to `node_modules/font-awesome/fonts` from within an application's `/public` directory. This would then allow all of `font-awesome`'s font files to be accessed directly by incoming clients. Unfortunately, while this approach does work in development when using the Meteor Tool, it does not work when production bundles are created.

Meteor's isobuild process uses a helper class called `SymlinkLoopChecker`, to make sure the build process doesn't get caught up in an infinite loop trying to follow circular symlinks. Currently, a shared `SymlinkLoopChecker` instance is used to watch for symlink loops during both the `_findSoures` and `_findAssets` parts of the isobuild process. `_findSources` is called first, and covers source files that an application uses from the `node_modules` directory. The `SymlinkLoopChecker` tracks all of the `node_modules` directories covered, so they can be watched for in the future (to prevent duplicate inclusions). Next, `_findAssets` is called using the same `SymlinkLoopChecker`. This means that if there are any `node_modules` symlinks used in the `public` or `private` directories, they will be marked as being duplicates (and stripped), since they were already covered in the `_findSources` run.

This commit changes things a bit so that both `_findSources` and `_findAssets` use their own `SymlinkLoopChecker` instance. This opens up an applications symlink capabilities a bit, while still preserving some circular symlink safeguards. By doing this, a production application bundle can now maintain the contents of `node_modules` based symlinks, used in `public` and `private`. 

So in the case of `font-awesome` for example

```
public/fonts --> ../node_modules/font-awesome/fonts
```

becomes the following in the production application bundle

```
bundle/programs/web.browser/app/fonts/[all fonts files]
```

Fixes #7013.


hwillson:
Marking as a work in progress as I'd like to test this a bit more.

PR #9667, opened 2018-02-14
Modernize route-policy

sebakerckhof:


packages/routepolicy/routepolicy_tests.js
```diff
@@ -1,5 +1,8 @@
+import RoutePolicy from 'meteor/routepolicy/routepolicy';
```
benjamn: I believe
```js
import RoutePolicy from './routepolicy.js';
```
would also work here, though I'm not entirely sure which way is better.

PR #9670, opened 2018-02-15
Version of mongodb updated

gustawdaniel:
Problem was described here

> https://github.com/mafintosh/mongojs/issues/347

Solution was found here:

> http://thecodebarbarian.com/a-nodejs-perspective-on-mongodb-36-array-filters.html

Reason of changes - lack of support for arrayFilters describled here:

> https://docs.mongodb.com/manual/reference/operator/update/positional-filtered/

Description, why this changes can help:

> Using array filters with Node.js requires versions >= 3.0.0 of the MongoDB Node.js driver or >= 5.0.0 of mongoose in addition to v3.6 of the MongoDB server. Earlier versions of the MongoDB (2.x) and mongoose (4.x) do not support array filters.

### DIFF
--- a/packages/npm-mongo/package.js
+++ b/packages/npm-mongo/package.js
@@ -3,12 +3,12 @@
 
 Package.describe({
   summary: "Wrapper around the mongo npm package",
-  version: "2.2.33",
+  version: "3.0.2",
   documentation: null
 });
 
 Npm.depends({
-  mongodb: "2.2.33"
+  mongodb: "3.0.2"
 });
 
 Package.onUse(function (api) {


hwillson:
Thanks for submitting this PR @gustawdaniel. Unfortunately, this isn't something we can accept at this time. Version 3 of the Node Mongo driver introduces breaking changes that will have to be handled delicately within Meteor core. The biggest change that has to be considered is that `find` and `findOne` no longer accept the `fields` parameter. This means Minimongo will have to be updated to translate fields parameter calls to `Cursor.prototype.project` calls. 

If you're interested in working on this further, great! If you could open a new feature request for this over in http://github.com/meteor/meteor-feature-requests, we can then hash out a plan of attack in that FR first, before anyone starts on a Mongo Node driver 3.x update PR. Thanks!

Issue #1646, opened 2013-12-02
minimongo use of `undefined` doesn't match mongo shell


datacarl:
Mini-mongo doesn't treat `undefined` in a $in lookup the same way as the server does. On the server it matches both null and undefined, but on the client it doesnt match any of them.

``` javascript
  Collection.insert({name: [null]});
  Collection.insert({name: [undefined]});
  Collection.insert({name: ['test']});

//server
console.log(Collection.find({name : {$in : [undefined]}}).count()); // -> 2

//client
console.log(Collection.find({name : {$in : [undefined]}}).count()); // -> 0
```

dotnetwise:
undefined should be ignored - otherwise you need to delete a property by using the `very slow` operator `delete`. Please reopen this issue

PR #9671, opened 2018-02-15
 Update Mongo.Collection to strip undefined fields

### LINKED ISSUE #9619: [1.6.1]The Mongo server and the Meteor query disagree on how many documents match your query.
We got this last night after updating from 1.6.0.1 -> 1.6.1 . We use an mlab cluster for our db running mongo 3.0.12 MMAPv1 and deploy to Galaxy. This bug seems to be a duplicate of #9087 and has been reproduced by @jessedvrs. 

Exception in defer callback: Error: The Mongo server and the Meteor query disagree on how many documents match your query. Maybe it is hitting a Mongo edge case? The query is: {}
at packages/mongo/oplog_observe_driver.js:881:15
at Object.Meteor._noYieldsAllowed (packages/meteor.js:730:12)
at OplogObserveDriver._publishNewResults (packages/mongo/oplog_observe_driver.js:851:12)
at OplogObserveDriver.proto._publishNewResults (packages/mdg_meteor-apm-agent.js:2899:38)
at OplogObserveDriver._runQuery (packages/mongo/oplog_observe_driver.js:758:10)
at OplogObserveDriver._runInitialQuery (packages/mongo/oplog_observe_driver.js:658:10)
at packages/mongo/oplog_observe_driver.js:191:10
at packages/mongo/oplog_observe_driver.js:15:9
at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1186:15)
at packages/meteor.js:502:25
at runWithEnvironment (packages/meteor.js:1240:24)

Here was the list of packages that were updated.

accounts-base upgraded from 1.4.0 to 1.4.2
autoupdate upgraded from 1.3.12 to 1.4.0
babel-compiler* upgraded from 6.24.7 to 7.0.0
babel-runtime upgraded from 1.1.1 to 1.2.0
boilerplate-generator upgraded from 1.3.1 to 1.4.0
callback-hook upgraded from 1.0.10 to 1.1.0
check upgraded from 1.2.5 to 1.3.0
coffeescript downgraded from 1.11.1_4 to 1.0.17
ddp-client upgraded from 2.2.0 to 2.3.1
ddp-common upgraded from 1.3.0 to 1.4.0
ddp-server upgraded from 2.1.1 to 2.1.2
diff-sequence upgraded from 1.0.7 to 1.1.0
dynamic-import upgraded from 0.2.1 to 0.3.0
ecmascript upgraded from 0.9.0 to 0.10.0
ecmascript-runtime-client upgraded from 0.5.0 to 0.6.0
es5-shim upgraded from 4.6.15 to 4.7.0
http upgraded from 1.3.0 to 1.4.0
id-map upgraded from 1.0.9 to 1.1.0
meteor-base upgraded from 1.2.0 to 1.3.0
minifier-css upgraded from 1.2.16 to 1.3.0
minifier-js upgraded from 2.2.2 to 2.3.1
modules upgraded from 0.11.1 to 0.11.3
mongo upgraded from 1.3.1 to 1.4.2
ordered-dict upgraded from 1.0.9 to 1.1.0
promise upgraded from 0.10.0 to 0.10.1
random upgraded from 1.0.10 to 1.1.0
reload upgraded from 1.1.11 to 1.2.0
retry upgraded from 1.0.9 to 1.1.0
server-render added, version 0.3.0
shim-common added, version 0.1.0
socket-stream-client added, version 0.1.0
standard-minifier-css upgraded from 1.3.5 to 1.4.0
standard-minifier-js upgraded from 2.2.3 to 2.3.1
url upgraded from 1.1.0 to 1.2.0
webapp upgraded from 1.4.0 to 1.5.0

These packages have been updated to new versions that are not backwards
compatible.
obit: updated to Meteor 1.6.1.
Changes to your project's package version selections from updating package versions:

babel-runtime upgraded from 1.2.0 to 1.2.2
caching-compiler upgraded from 1.1.9 to 1.1.11
ecmascript-runtime-client upgraded from 0.6.0 to 0.6.1
es5-shim upgraded from 4.7.0 to 4.7.3
jquery upgraded from 1.11.10 to 1.11.11
modules-runtime upgraded from 0.9.1 to 0.9.2
mongo upgraded from 1.4.2 to 1.4.3
npm-mongo upgraded from 2.2.33 to 2.2.34
stylus upgraded from 2.513.13 to 2.513.14

hwillson:
`Mongo.Collection` has been updated to strip `undefined` fields set in documents/selectors passed to `find`, `findOne`, `insert`, `update`, etc. This lines the codebase up with the changes made in https://github.com/meteor/meteor/commit/ce3885b6df10f9f69f425af8698642a388efa6f2, and helps prevent "The Mongo server and the Meteor query disagree on how many documents match your query" errors.

Fixes #9619.


packages/mongo/collection.js
```diff
@@ -301,6 +298,24 @@ Object.assign(Mongo.Collection.prototype, {
...
+  // fields in documents/selectors. This function can be used to remove
+  // `undefined` fields, to keep things consistent, and help avoid any
+  // unwarranted side effects (such as issue #9619).
+  _removeUndefinedFields(selector) {
+    const cleanSelector = selector;
+    if (cleanSelector) {
+      Object.keys(cleanSelector).forEach((key) => {
+        if (cleanSelector[key] && typeof cleanSelector[key] === 'object') {
+          this._removeUndefinedFields(cleanSelector[key]);
+        } else if (cleanSelector[key] === undefined) {
```
benjamn: Let's use `typeof cleanSelector[key] === "undefined"` here.

packages/mongo/collection.js
```diff
@@ -301,6 +298,24 @@ Object.assign(Mongo.Collection.prototype, {
     }
   },
 
+  // The Mongo Node driver is configured to `ignoreUndefined`
+  // fields in documents/selectors. This function can be used to remove
+  // `undefined` fields, to keep things consistent, and help avoid any
+  // unwarranted side effects (such as issue #9619).
+  _removeUndefinedFields(selector) {
```
benjamn: Let's add a `seen = new Set` parameter here, and pass it through to the recursive call below, so that we can keep track of objects we've seen before, to avoid infinite loops if the selector has any cycles.

packages/mongo/collection.js
hwillson: Oops 😳 - good catch, thanks!

packages/mongo/collection.js
hwillson: Great idea - will do, thanks!

packages/mongo/collection.js
```diff
@@ -301,6 +298,33 @@ Object.assign(Mongo.Collection.prototype, {
...
+    const selectorCopy = EJSON.clone(selector);
+    const seen = new Set();
+    (function removeUndefinedRecursively(cleanSelector) {
+      if (!seen.has(cleanSelector)) {
+        seen.add(cleanSelector);
+        Object.keys(cleanSelector).forEach((key) => {
+          if (cleanSelector[key] && typeof cleanSelector[key] === 'object') {
+            removeUndefinedRecursively(cleanSelector[key]);
+          } else if (typeof cleanSelector[key] === 'undefined') {
+            delete cleanSelector[key];
```
benjamn: This may be tricky, but I wonder if there's a way to avoid the `EJSON.clone` if we aren't going to delete any fields from the original `selector`.

benjamn:
The more I think about this, the more it seems like treating
```js
collection.find({ key: valueThatHappensToBeUndefined })
```
as if it was
```js
// Returns all documents!
collection.find({}) 
```
might not be the right solution, since the caller probably wanted to fetch only documents with some particular value for `key`, rather than all documents.

Would it be better to throw if we find an `undefined` value? There seems to be some [precedent](https://github.com/meteor/meteor/blob/b4690a2541fd104cbfba1c269a31755996e5855e/packages/minimongo/local_collection.js#L89-L90) for this idea in the `LocalCollection` code, though only in the comments.

mitar:
Yes, it should not be the same. Better to return less than too many documents.

What does MongoDB do?

hwillson:
Definitely an interesting problem. Mongo itself throws:

```
Error: error: {
	"$err" : "Can't canonicalize query: BadValue cannot compare to undefined",
	"code" : 17287
}
```

but this PR was lining up with the Mongo Node driver (since it's our entry point into Mongo), which when using [`ignoreUndefined`](http://mongodb.github.io/node-mongodb-native/2.2/api/Db.html) strips `undefined` fields without throwing.

I'm happy to extend this to throw if you guys agree? Lining up with Mongo's exact behaviour makes a lot of sense. I'm wondering if that then makes this change less backwards compatible though, since application code that appeared to work before because it stripped  `undefined` fields, will now throw? I guess we'll just have to make sure this is well documented, and maybe hold off on releasing these updates until 1.6.2, so we can highlight theses changes in the 1.6.2 release notes / blog post (to hopefully get more eyes on this change before people blindly update and visibly break their apps).

Let me know if I have a 👍 to revert this and submit a new PR that throws when using `undefined`. Thanks!

mitar:
Last time I remember I was playing with undefined and null in Meteor and queries it was a strange situation. Especially because EJSON and method calls process it as well. If I remember correctly, `undefined` is being converted to `null` if you pass it to the meteor method. So then if you pass `undefined` to a Meteor method and then on the server to a mongodb query, you in fact get `null` into the query. On the other hand, simulation gets `undefined`.

See also [this blog post by an ex-MDG person](http://devblog.me/wtf-mongo), section 2. So I think there was some previous work/discussion about how to handle undefined in Meteor. So it might be that if we want good backwards compatibility we should maybe read discussion in https://github.com/meteor/meteor/issues/1646.

benjamn:
Anecdotally, all of the reports in #9619 seem like the `undefined` values were unintended surprises, and the developer would have been glad to know about the source of the problem, rather than mystified by the minimongo/mongo disagreement error.

hwillson:
@mitar Everything you mentioned is how we ended up with this PR. For the full backstory, see https://github.com/meteor/meteor/pull/9444. We're trying to align all of this to get rid of the inconsistencies between minimongo/mongo, and at the same time respect Mongo's [BSON spec](http://bsonspec.org/spec.html), which has deprecated `undefined`.

hwillson:
Agreed @benjamn - so should I proceed with reverting this and submitting a new PR that throws if an `undefined` field is used (on both the client and server)?

benjamn:
@hwillson Yeah, I think we can save people from a lot of subtle bugs by throwing instead of trying to be clever about this. The error message can recommend switching to `null` if the `undefined` value was not a mistake.

hwillson:
Sounds great @benjamn - I'm on it!

romainkoenig:
> Yeah, I think we can save people from a lot of subtle bugs by throwing instead of trying to be clever about this. The error message can recommend switching to null if the undefined value was not a mistake.

Glad that you see the point, sadly it's too late for us :-(

PR #9679, opened 2018-02-21
add .eslintrc

pravdomil:
taken from https://github.com/meteor/javascript/blob/master/linters/.eslintrc

we definitely should include .eslintrc in main repo so that all dev tools take it into consideration

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,217 @@
+{
+  "parser": "babel-eslint",
+  "plugins": [
+    "react"
+  ],
+  "env": {
+    "browser": true,
+    "node": true
+  },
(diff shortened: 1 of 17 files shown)


hwillson:
Hi @pravdomil - I see you closed this, but have restored the branch, so I'm assuming you might still be working on it. In an effort to save you some time, just a heads up - the Meteor project isn't currently accepting eslint related PR's (or https://prettier.io related PR's, etc.). Feel free to use linters or beautifiers on your code before you submit it, but please keep in mind that the Meteor core code styles differ from the eslint rules you linked to. When in doubt please follow the existing codebase styles as much as possible. Thanks!

PR #9558, opened 2018-01-21
Modernize accounts packages

jamesmillerburgess:
### Update
This pull request modernizes the various `accounts-*` packages.

#### Todo:
- [x] Removal of `underscore` in favor of equivalent ES6+ features
- [x] Check that `_.each` replacements are applied correctly to arrays and objects
- [x] Prefer `includes` over `indexOf`
- [x] Prefer `Array.isArray` over `instanceof Array`
- [x] Removal of `self = this` in all cases where the same scoping can be achieved with arrow functions
- [x] Prefer arrow functions over `function`
- [x] Prefer `class` method declarations over `prototype` manipulation
- [x] Removal of `var` and `function` declarations in favor of `const` and `let`
- [x] Prefer template literals over string concatenation
- [x] Object deconstruction where it makes sense
- [x] Prefer module exports to globals
- [x] Apply DRY for repeated utility functions
- [ ] Extraction of method bodies into logically grouped files
- [ ] JSDoc comments on all methods
- [ ] Additional comments where needed
- [ ] Additional unit tests
- [ ] Additional integration tests
- [x] Increment package numbers

#### Package size chart (as of 1.6.1-beta.3):
```
Package             | Original web.browser | Modernized web.browser | Original web.browser.legacy | Modernized web.browser.legacy
accounts-base       | 12.8  kb             | 11.3  kb               | 12.8  kb                    | 13.3  kb
accounts-password   |  2.16 kb             |  2.32 kb               |  2.16 kb                    |  2.37 kb
```

### Original Comment
This pull request removes the `underscore` dependency from `accounts-base`. It's the first step in a larger modernization/refactoring that I was thinking I could do as a contribution to the Meteor codebase. I made the PR now, because I wanted to get some feedback first on the idea and how far to go.

Options:
1. Remove `underscore` dependencies
2. ES6 syntax/style (e.g. remove usage of `self`, apply Prettier styling, and leverage other ES features)
3. Refactor large functions/files into smaller pieces to enable better test coverage
4. Other...?

Please let me know if this sounds like something that could help and which of the options sound about right.

PR #9680, opened 2018-02-21
add .eslintrc

pravdomil:
taken from https://github.com/meteor/javascript/blob/master/linters/.eslintrc

we definitely should include .eslintrc in main repo so that all dev tools take it into consideration

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,217 @@
+{
+  "parser": "babel-eslint",
+  "plugins": [
+    "react"
+  ],
+  "env": {
+    "browser": true,
+    "node": true
+  },
+  "ecmaFeatures": {
+    "arrowFunctions": true,
+    "blockBindings": true,
+    "classes": true,
+    "defaultParams": true,
+    "destructuring": true,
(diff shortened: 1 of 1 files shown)


hwillson:
We just posted at the same time! 🙂  Re-capping my comments from https://github.com/meteor/meteor/pull/9679#issuecomment-367439598:

> ... the Meteor project isn't currently accepting eslint related PR's (or https://prettier.io related PR's, etc.). Feel free to use linters or beautifiers on your code before you submit it, but please keep in mind that the Meteor core code styles differ from the eslint rules you linked to. When in doubt please follow the existing codebase styles as much as possible. Thanks!

pravdomil:
I have closed the other PR beause my git branches get mixed.

@hwillson thanks for quick response, yes I'm in doubt.  
Because I'm working on modernising Meteor core code (as other people here do), and there is no guideline how to should I format it, so I hoped that we can use that .eslintrc thing to makes things clear.

pravdomil:
https://github.com/meteor/meteor/pull/9558#issuecomment-359808310

PR #9681, opened 2018-02-21
Modernize debug

pravdomil:
...

### DIFF
--- a/packages/meteor/debug.js
+++ b/packages/meteor/debug.js
@@ -61,9 +61,10 @@ Meteor._debug = function(...arguments) {
   }
 };
 
-// Suppress the next 'count' Meteor._debug messages. Use this to
-// stop tests from spamming the console.
-//
+/**
+ * Suppress the next 'count' Meteor._debug messages. Use this to
+ * stop tests from spamming the console.
+ */
 Meteor._suppress_log = function(count) {
   suppress += count;
 };
--- a/packages/meteor/debug.js
+++ b/packages/meteor/debug.js
@@ -1,16 +1,17 @@
 let suppress = 0;
 
-// replacement for console.log. This is a temporary API. We should
-// provide a real logging API soon (possibly just a polyfill for
-// console?)
-//
-// NOTE: this is used on the server to print the warning about
-// having autopublish enabled when you probably meant to turn it
-// off. it's not really the proper use of something called
-// _debug. the intent is for this message to go to the terminal and
-// be very visible. if you change _debug to go someplace else, etc,
-// please fix the autopublish code to do something reasonable.
-//
+/**
+ * replacement for console.log. This is a temporary API. We should
+ * provide a real logging API soon (possibly just a polyfill for
+ * console?)
+ *
+ * NOTE: this is used on the server to print the warning about
+ * having autopublish enabled when you probably meant to turn it
+ * off. it's not really the proper use of something called
+ * _debug. the intent is for this message to go to the terminal and
+ * be very visible. if you change _debug to go someplace else, etc,
+ * please fix the autopublish code to do something reasonable.
+ */
 Meteor._debug = function(...arguments) {
   if (suppress) {
     suppress--;
--- a/packages/meteor/debug.js
+++ b/packages/meteor/debug.js
@@ -16,43 +16,46 @@ Meteor._debug = function(/* arguments */) {
     suppress--;
     return;
   }
+  
   if (typeof console !== 'undefined' &&
-      typeof console.log !== 'undefined') {
-    if (arguments.length == 0) { // IE Companion breaks otherwise
-      // IE10 PP4 requires at least one argument
-      console.log('');
-    } else {
-      // IE doesn't have console.log.apply, it's not a real Object.
-      // http://stackoverflow.com/questions/5538972/console-log-apply-not-working-in-ie9
-      // http://patik.com/blog/complete-cross-browser-console-log/
-      if (typeof console.log.apply === 'function') {
-        // Most browsers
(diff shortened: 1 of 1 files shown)


packages/meteor/debug.js
```diff
@@ -1,67 +1,76 @@
...
+ * console?)
+ *
+ * NOTE: this is used on the server to print the warning about
+ * having autopublish enabled when you probably meant to turn it
+ * off. it's not really the proper use of something called
+ * _debug. the intent is for this message to go to the terminal and
+ * be very visible. if you change _debug to go someplace else, etc,
+ * please fix the autopublish code to do something reasonable.
+ */
+Meteor._debug = function(...arguments) {
```
klaussner: `arguments` is a [reserved word](https://mothereff.in/js-variables#arguments), so it can't be used as an identifier. Maybe use `args` instead?

pravdomil: https://github.com/meteor/meteor/pull/9681/commits/29bca2cc72bff588cb628084436d758cfdd16a7f

packages/meteor/debug.js
```diff
@@ -1,67 +1,76 @@
...
-//
-// NOTE: this is used on the server to print the warning about
-// having autopublish enabled when you probably meant to turn it
-// off. it's not really the proper use of something called
-// _debug. the intent is for this message to go to the terminal and
-// be very visible. if you change _debug to go someplace else, etc,
-// please fix the autopublish code to do something reasonable.
-//
-Meteor._debug = function (/* arguments */) {
+/**
```
hwillson: Don't worry about changing comment style like this. Multi-line `//`'s are fine (and these comments are not being run through a comment/API parser, like jsdoc).

packages/meteor/debug.js
```diff
@@ -1,67 +1,76 @@
...
-        // Chrome and Safari only hyperlink URLs to source files in first argument of
-        // console.log, so try to call it with one argument if possible.
-        // Approach taken here: If all arguments are strings, join them on space.
-        // See https://github.com/meteor/meteor/pull/732#issuecomment-13975991
-        var allArgumentsOfTypeString = true;
-        for (var i = 0; i < arguments.length; i++)
-          if (typeof arguments[i] !== "string")
-            allArgumentsOfTypeString = false;
+  if (typeof console !== 'undefined' && typeof console.log !== 'undefined') {
+    return;
```
hwillson: Hmm - isn't this changing the entire functionality of `Meteor._debug`? Before your changes this check was in place to make sure that `console` could be referenced. Now with your changes, if `console` is available, the rest of the code is skipped. Was this intentional?

packages/meteor/debug.js
```diff
@@ -1,67 +1,76 @@
...
 
-        if (allArgumentsOfTypeString)
-          console.log.apply(console, [Array.prototype.join.call(arguments, " ")]);
-        else
-          console.log.apply(console, arguments);
+  if (args.length === 0) { // IE Companion breaks otherwise
+    // IE10 PP4 requires at least one argument
+    console.log('');
+  }
+  else {
```
hwillson: Please keep `else`'s and `else if`'s on the same line as the closing/opening brace. E.g. `} else {`

packages/meteor/debug.js
```diff
@@ -1,67 +1,76 @@
...
+ * console?)
+ *
+ * NOTE: this is used on the server to print the warning about
+ * having autopublish enabled when you probably meant to turn it
+ * off. it's not really the proper use of something called
+ * _debug. the intent is for this message to go to the terminal and
+ * be very visible. if you change _debug to go someplace else, etc,
+ * please fix the autopublish code to do something reasonable.
+ */
+Meteor._debug = function(...args) {
```
hwillson: Please keep the space after `function` for all anonymous functions.

hwillson: And also - since the name of this PR includes "modernize", you could consider replacing your anonymous function declarations with arrow functions. In this case specifically it would make sense since you're using spread syntax to capture all function arguments in `args`, but since you're using a function declaration, you're just masking the fact that an `arguments` object is still available (arrow functions don't have `arguments`).

packages/meteor/debug.js
pravdomil: oh please https://github.com/meteor/meteor/pull/9680

pravdomil: I'm not planning fixing that since you should provide some `auto code style format tool` first

abernix: @pravdomil We're not interested in adopting an auto-code style tool at this time.  The Meteor codebase started way before such tools were popularized and running one in bulk on our codebase would destroy too much valuable commit history.  Linters are great when a project starts with them and has always had them in place, but when they introduce additional walls between debugging when a broken commit was introduced, or mask history (by requiring to dive through "Linter" commits), they don't do much for an existing codebase.

If you're not interested in making the change, we'll do it, but I hope you understand.  For what it's worth, we do have a guide in place which makes the same request that @hwillson has made here: https://github.com/meteor/javascript#16.2.

pravdomil: >  running one in bulk on our codebase would destroy too much valuable commit history

agree, but there is way just in between "no formatting" and "auto formatting", what I would propose here is, setup prettier or other auto formatting tools, so that *new code* will be written in same style and we don't need to worry about that anymore and spend time on more productive things

now you provide no code style and it's hard to write new code without getting comment like this one

hwillson:
Hi @pravdomil - any progress here? Are you still interested in working on this, and if so can you address the posted review comments? Let us know - thanks!

PR #9691, opened 2018-02-23
Document new bug triage timelines

hwillson:
Issue triage documentation changes outlining the new bug lifespan timelines we've discussed previously. Thanks!

### DIFF
--- a/ISSUE_TRIAGE.md
+++ b/ISSUE_TRIAGE.md
@@ -44,12 +44,12 @@ To help keep issues in this repository under control, and make sure the most imp
 ##### Issues Labelled with `bug` and `confirmed`
 
 - Open `bug` + `confirmed` issues should be closed after two months of inactivity, unless someone has clearly identified that they are interested in working on the issue.
-- Triager's should do everything possible to help get `bug` + `confirmed` issues to `pull-requests-encouraged`. This means helping clearly identify where the problem is, pointing towards parts of the codebase that someone might want to look into, documenting what a potential solution looks like, etc.
+- Triagers should do everything possible to help get `bug` + `confirmed` issues to `pull-requests-encouraged`. This means helping clearly identify where the problem is, pointing towards parts of the codebase that someone might want to look into, documenting what a potential solution looks like, etc.
 
 ##### All Other Issues 
 
 - All open issues that can’t be labelled as `bug` + `confirmed` and/or `pull-requests-encouraged`, should be closed after one month of inactivity.
-- Triager's should do everything possible to help get `bug` + `confirmed` issues to `pull-requests-encouraged`. 
+- Triagers should do everything possible to help get `bug` + `confirmed` issues to `pull-requests-encouraged`. 
 
 ### Help questions
 
--- a/ISSUE_TRIAGE.md
+++ b/ISSUE_TRIAGE.md
@@ -29,6 +29,28 @@ The first step is in determining whether the issue is a bug, help question or fe
 4. A reproduction should be confirmed by at least one person other than the original reporter. Run the reproduction and validate that the bug exists; then make a note of your findings on the issue. If a reproduction is supplied but doesn't work, add the `can't-reproduce` label and make a comment describing what happened.
 5. Finally, once you've confirmed the reproduction add the `confirmed` label and [classify](#classification) the issue (removing the `can't-reproduce` label if it exists).
 
+#### Bug Issue Lifespan
+
+To help keep issues in this repository under control, and make sure the most important problems are visible to maintainers, unresolved issues (lacking recent activity) should be closed after a certain amount of time has elapsed. 
+
+##### Issues Labelled with `pull-requests-encouraged`
+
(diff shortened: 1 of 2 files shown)


ISSUE_TRIAGE.md
```diff
@@ -29,6 +29,28 @@ The first step is in determining whether the issue is a bug, help question or fe
...
 5. Finally, once you've confirmed the reproduction add the `confirmed` label and [classify](#classification) the issue (removing the `can't-reproduce` label if it exists).
 
+#### Bug Issue Lifespan
+
+To help keep issues in this repository under control, and make sure the most important problems are visible to maintainers, unresolved issues (lacking recent activity) should be closed after a certain amount of time has elapsed. 
+
+##### Issues Labelled with `pull-requests-encouraged`
+
+- Open `pull-requests-encouraged` issues should be closed after one month of inactivity, unless someone has clearly identified that they are interested in working on the issue.
+- When closing, the `not-implemented` label should be added.
```
hwillson: @benjamn I know we talked about reversing this so completed `pull-requests-encouraged` issues have an `implemented` label added instead (meaning people looking for PR's to work on can do a `label:pull-requests-encouraged -label:implemented` GH search). I think in practice this might prove cumbersome however, since some PR issues are closed automatically when the PR is merged (when using `fixes` for example). In that case we'd have to remember to always go back to the original PR issue and add the `implemented` label. When triaging issues, it's likely easier to add a `not-implemented` label since we're already in triage mode and will be manually closing the issue anyways.

ISSUE_TRIAGE.md
```diff
@@ -29,6 +29,28 @@ The first step is in determining whether the issue is a bug, help question or fe
...
+- A message similar to the following should be included:
+
+> While we think resolving this issue would be a great addition to the Meteor project, we're going to close it for now due to inactivity. If anyone comes across this issue in the future, and is interested in working on resolving it, please let us know by posting here and we'll consider re-opening this issue. Thanks!
+
+##### Issues Labelled with `bug` and `confirmed`
+
+- Open `bug` + `confirmed` issues should be closed after two months of inactivity, unless someone has clearly identified that they are interested in working on the issue.
+- Triager's should do everything possible to help get `bug` + `confirmed` issues to `pull-requests-encouraged`. This means helping clearly identify where the problem is, pointing towards parts of the codebase that someone might want to look into, documenting what a potential solution looks like, etc.
```
benjamn: Just "Triagers" please.

ISSUE_TRIAGE.md
benjamn: When I first suggested using `implemented` instead of `not-implemented`, I was imagining it might be annoying to add the `not-implemented` tag to every `pull-requests-encouraged` issue, but now I see that simply being tagged with `pull-requests-encouraged` and _still open_ implies that the issue has not yet been implemented, so we don't have to add the `not-implemented` tag explicitly. After we _close_ a `pull-requests-encouraged` issue without implementing it, that's when we would add `not-implemented`, which seems fine!

ISSUE_TRIAGE.md
```diff
@@ -29,6 +29,28 @@ The first step is in determining whether the issue is a bug, help question or fe
...
+
+##### All Other Issues 
+
+- All open issues that can’t be labelled as `bug` + `confirmed` and/or `pull-requests-encouraged`, should be closed after one month of inactivity.
```
hwillson: One month of inactivity might be too much here. If we can't prove something is a bug fairly quickly, then that either means we're missing a reproduction or the reported problem is too vague to work on resolving. I think 2 weeks of inactivity here is probably better ...

ISSUE_TRIAGE.md
```diff
@@ -29,6 +29,28 @@ The first step is in determining whether the issue is a bug, help question or fe
...
+- Triager's should do everything possible to help get `bug` + `confirmed` issues to `pull-requests-encouraged`.
```
benjamn: please no apostrophe thx

ISSUE_TRIAGE.md
hwillson: Clearly I was absent the day they taught proper possessive apostrophe use in grade school ...

ISSUE_TRIAGE.md
```diff
@@ -29,6 +29,28 @@ The first step is in determining whether the issue is a bug, help question or fe
 4. A reproduction should be confirmed by at least one person other than the original reporter. Run the reproduction and validate that the bug exists; then make a note of your findings on the issue. If a reproduction is supplied but doesn't work, add the `can't-reproduce` label and make a comment describing what happened.
 5. Finally, once you've confirmed the reproduction add the `confirmed` label and [classify](#classification) the issue (removing the `can't-reproduce` label if it exists).
 
+#### Bug Issue Lifespan
```
abernix: I'm feeling mighty nit-picky right now, so I'll lowercase the second and third words of this level-4 heading and keep only the first words capitalized in the subsequent headings.  (Mainly for consistency with the level 1 and 2 headings above, but also because these are nested further in the tree of this document (H3s and H4s).

Just FYI, I'll do this before merging.

PR #9702, opened 2018-02-27
Update stylus deprecation notice to use proper last version

hwillson:
The `stylus` deprecation notice is pointing to the wrong last non-deprecated version. It should refer to `2.513.13`. I've updated the references accordingly (and we'll need to publish a new version so I've bumped the patch version). Thanks!


hwillson:
Just to add more detail - people are following the deprecation instructions and running `meteor add stylus@=2.513.14`, then continuing to get the deprecation warning since it's part of `2.513.14`. `2.513.13` is the last version without the deprecation warning.

Issue #9684, opened 2018-02-21
Why does Meteor not throw errors in method calls?


morloy:
We use Sentry to report errors in our production code. However, the Meteor methods don't throw errors when they fail. Instead they call the weird `_debug` function and even deconstruct the `Error` object, removing valuable debug information.

Is there any explanation by the dev team about this choice of behavior?
If not, I would be very happy if you could the include some changes to make methods throw real errors, not some random debug messages: https://github.com/meteor/meteor/compare/devel...morloy:fix/throwMethodErrors

mitar:
You mean on the client side?

I never use Meteor methods directly, but have a wrapping code which wraps all method calls and definitions. Similar to [validated method](https://github.com/meteor/validated-method). So you can handle then errors like you want and call any logging code.

mitar:
But I do remember there being issues for being able to intercept those errors before. I agree that there should be a standard way to do so.

morloy:
Yes, client side.

Meteor handles errors on methods internally (the `ddp-client` and `bindEnvirontment`). Thus there's no way for packages like `validated-method` to access the actual error object.

mitar:
Ehm. Why don't you log method errors from server on the server itself?

You can wrap (with a library) all your method call on servers with try/catch and then log it, before rethrowing it to the client to show it to the client?

morloy:
Server-side errors are not an issue. These get sanitized and sent to the client. The problem is that the client does not throw any errors. Instead there's only a not-so-informative message through `_debug()`.

hwillson:
Hi @morloy - Meteor 1.6.2 will include some error handling changes (see https://github.com/meteor/meteor/pull/9654), but to continue the discussion on further changes, please move this issue over to the http://github.com/meteor/meteor-feature-requests repo. Issues here are reserved for bugs with Meteor core itself, and the current error handling infrastructure is working as designed. For a thorough explanation of why Method error handling works the way it works, see the Guide's [Method error handling](https://guide.meteor.com/methods.html#errors) section. Thanks!

morloy:
@hwillson thanks for the hint to #9654. That's exactly the issue that I was describing here.
However, it only fixes the behavior if a callback is explicitly given to a method call.

PR #9707, opened 2018-02-28
preserve error object in method callback

morloy:
This is an addition for #9654 and also solves the issue in #9684.
The error object (and with it the stack trace) is lost when calling a method without a callback.


pravdomil:
see also https://github.com/meteor/meteor/pull/9678

PR #9709, opened 2018-02-28
New methods for server-register (FR#259)

ch-luke:
The current server-register package seems to have a narrow use case and therefore request the attached changes. For my first overview and thoughts on the topic - please refer to the Feature request - https://github.com/meteor/meteor-feature-requests/issues/259.

BTW, when I tested some of the out-the-box methods I was getting some interesting results. For example when I did `sink.appendToElementById('noscript', 'this is additional stuff')` I was expecting the code to be added to the end of the current value and not prepended to the start - and the whole resulting html of the element was encapsulated in quotes and made into a string. 

I really need the PR to be approved in some form to support my SEO efforts. 

Thanks in advance - Luke.

### DIFF
--- a/packages/server-render/server-register.js
+++ b/packages/server-render/server-register.js
@@ -1,9 +1,57 @@
 import { WebAppInternals } from "meteor/webapp";
+import cheerio from "cheerio";
 import MagicString from "magic-string";
 import { SAXParser } from "parse5";
 import { create as createStream } from "combined-stream2";
 import { ServerSink, isReadable } from "./server-sink.js";
 import { onPageLoad } from "./server.js";
+
+function updateContent(data, section, selector, itemArray) {
+  let reallyMadeChanges = false;
+  const content = section === "head" ?
+    `<head>${data.head}</head>` :
+    `<body>${data.body}</body>`;
+  $ = cheerio.load(content,
+    { 
+      withDomLvl1: true,
+      normalizeWhitespace: false,
+      xmlMode: true,
+      decodeEntities: true
+    }
+  );
+
+  itemArray.forEach((item) => {
+    let searchField = "";
+    if (selector === "id") {
+      searchField = `#${item.id}`;
+    } else {
+      searchField = item.searchAttribute && item.searchAttributeValue ?
+        `${item.tag}[${item.searchAttribute}=${item.searchAttributeValue}]` :
+        item.tag;
+    }
+    if (!item.value) {
+      $(searchField).remove();
+    } else {
+      if (item.updateAttribute) {
+        $(searchField).attr(item.updateAttribute, item.value);
+      } else {
+        if (item.updateType === 'prepend') {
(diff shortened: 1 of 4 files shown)


packages/server-render/server-register.js
```diff
@@ -1,10 +1,58 @@
...
+      }
+      reallyMadeChanges = true;
+    }
+  });
+  if (reallyMadeChanges) {
+    data[section] = $(section).html();
+  }
+
+  return reallyMadeChanges;
+}
```
benjamn: Please use (or improve) the existing `rewrite` pattern, based on the `parse5` streaming `SAXParser`, rather than introducing a new parsing technique. Specifically, the pattern you're using here doesn't allow for a streaming response, because it takes whole chunks of HTML and replaces them after all transforms have completed. Especially since you're modifying the `<head>`, this pattern has the potential to delay rendering of the rest of the page, since the `<head>` has to be rendered first.

packages/server-render/server-sink.js
```diff
@@ -33,6 +35,44 @@ export class ServerSink {
     this.appendToElementById(id, html);
   }
 
+  updateHeadElementByTag(
```
benjamn: You'll need to find a way to make these changes work for the `client-sink.js` as well, so that rendering can remain server/client agnostic. That's part of why the sink API is so limited: it can only support operations that work on both the server and the client in the same way.

benjamn:
As a side note, submitting a feature request and then immediately submitting a PR, without first discussing the design and establishing consensus, is a recipe for disappointment. You are welcome to install the `server-render` package locally in your application's `packages/` directory, and modify it however you like, but elevating these changes to the official package that everyone uses demands substantially more buy-in from other members of the community.

ch-luke:
@benjamn , thanks for the responses and appreciate the feedback. I take on board your thoughts on updating my changes to also support 'client side' and about streaming head changes.

After I didn't get even an initial response from Meteor about the feature request (30 Jan - 5 weeks ago), in the spirit of being constructive, and upping the pace of change, starting looking into how to expand and improve the capabilities of the server-register package. After reviewing the code I noticed a few bugs (refer to previous notes) and a heavy slant to React (although the package can be easily used for Blaze / Vue etc) and the use of parse5 / magic string.

The problem I find with pure parse5 is that lack of a query layer, which cheerio has. BTW cheerio is now based on parse5 and is specially build for speed. I'm still researching how to include streaming with this, and alternatively how to expand along the sax / parse5 / magic-string approach, and will response in a subsequent post.

hwillson:
Thanks very much for taking initiative and diving in here @ch-luke! It sounds like we'll need to discuss this a bit further before continuing with a PR, so I'll close this PR for now. Let's move the discussion back into https://github.com/meteor/meteor-feature-requests/issues/259, and hash out a new plan of attack based on the findings here.

PR #9714, opened 2018-03-01
Support a meteor.testModule section in application package.json files.

benjamn:
Similar to #9690, this PR adds support for a `meteor.testModule` configuration in application `package.json` files, allowing application developers complete control over what test files should load when running `meteor test` to test their applications.

Setting `meteor.testModule` is a great way to specify test entry points explicitly, rather than relying on Meteor's `isTestFilePath` [heuristics](https://github.com/meteor/meteor/blob/devel/tools/isobuild/test-files.js). However, this feature is entirely opt-in, so the old behavior is preserved if you do not configure `meteor.testModule`.

The syntax is identical to `meteor.mainModule`, so you can set a different `meteor.testModule.{client,server,...}` for each platform. If a `testModule` is not specified for a platform, then Meteor's existing rules about test file paths apply for that platform, as before.

Different `testModule`s for client and server:
```js
"meteor": {
  "testModule": {
    "client": "client/tests.js",
    "server": "server/tests.js"
  }
}
```

Same `testModule` for both client and server:
```js
"meteor": {
  "testModule": "tests.js"
}
```

Server tests without any client tests (note that simply omitting the `"client"` property would cause tests to behave as if there was no `meteor.testModule` on the client):
```js
"meteor": {
  "testModule": {
    "client": false,
    "server": "server/tests.js"
  }
}
```

If a `testModule` is specified, that module will always be loaded eagerly when running `meteor test`, in addition to any other modules that load eagerly due to `meteor.mainModule` or other rules regarding module loading. If you run `meteor test` without the `--full-app` option, then no application JS modules other than the `testModule` (and any modules imported by it) will be loaded eagerly. If you run `meteor test --full-app`, the `meteor.mainModule` module(s) will be loaded eagerly in addition to the `meteor.testModule` module(s).

The module(s) specified by `meteor.testModule` can import other test modules at runtime, so you can still distribute test files across your codebase; just make sure you import the ones you want to run.

### DIFF
--- a/History.md
+++ b/History.md
@@ -13,9 +13,45 @@
   When specified, these entry points override Meteor's default module
   loading semantics, rendering `imports` directories unnecessary. If
   `mainModule` is left unspecified for either client or server, the
-  default rules will apply for that architecture, as before.
+  default rules will apply for that architecture, as before. To disable
+  eager loading of modules on a given architecture, simply provide a
+  `mainModule` value of `false`:
+  ```js
+  "meteor": {
+    "mainModule": {
+      "client": false,
+      "server": "server/main.js"
+    }
+  }
+  ```
   [Feature #135](https://github.com/meteor/meteor-feature-requests/issues/135)
   [PR #9690](https://github.com/meteor/meteor/pull/9690)
+
+* In addition to `meteor.mainModule`, the `"meteor"` section of
+  `package.json` may also specify `meteor.testModule` to control which
+  test modules are loaded by `meteor test` or `meteor test --full-app`:
+  ```js
+  "meteor": {
+    "mainModule": {...},
+    "testModule": "tests.js"
+  }
+  ```
+  If your client and server test files are different, you can expand the
+  `testModule` configuration using the same syntax as `mainModule`:
+  ```js
+  "meteor": {
+    "testModule": {
+      "client": "client/tests.js",
+      "server": "server/tests.js"
+    }
+  }
+  ```
+  The same test module will be loaded whether or not you use the
+  `--full-app` option. Any tests that need to detect `--full-app` should
+  check `Meteor.isAppTest`. The module(s) specified by `meteor.testModule`
+  can import other test modules at runtime, so you can still distribute
+  test files across your codebase; just make sure you import the ones you
+  want to run.
 
 * The `reify` npm package has been updated to version 0.14.2.
 
--- a/History.md
+++ b/History.md
@@ -51,7 +51,7 @@
   check `Meteor.isAppTest`. The module(s) specified by `meteor.testModule`
   can import other test modules at runtime, so you can still distribute
   test files across your codebase; just make sure you import the ones you
-  want to run.
+  want to run. [PR #9714](https://github.com/meteor/meteor/pull/9714)
 
 * The `reify` npm package has been updated to version 0.14.2.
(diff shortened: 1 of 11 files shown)


pravdomil:
that would be handy also for packages in node_modules, first step in transition from atmosphere packages, have anybody try something like this?

benjamn:
@pravdomil I haven't been thinking about that recently, but you're right, we could potentially scan `node_modules` packages to find those that have a `meteor` section in `package.json`, and run those entry points on application startup (server) and/or page load (client).

That's another great reason to use Meteor-specific configuration instead of just reusing the `main` or `browser` fields in `package.json` files, since every npm package has those already.

pravdomil:
@benjamn I'm writing plugin that does exactly that, since I want to move my repos to npm, can you give me some hints where to start since probably you know the codebase a lot better

Floriferous:
Can this be used in an effortless way if you have a lot of test files? 

For example, I'd like to be able to import all tests from a few directories, without having a massive import list in this testModule file.

SimonSimCity:
@Floriferous I guess this answer on stackoverflow should guide you the way:

https://stackoverflow.com/questions/5364928/node-js-require-all-files-in-a-folder#answer-28976201

> I recommend using glob to accomplish that task.
>
> ```js
> var glob = require( 'glob' )
>  , path = require( 'path' );
>
> glob.sync( './routes/**/*.js' ).forEach( function( file ) {
>   require( path.resolve( file ) );
> });

You just have to keep in mind that Meteor will build a test-version of your application and therefore cut away all code that isn't included. You'll therefore have to add the files you want manually - and since the app code is not run in the repository, you have to define the full path to the files you want to import. This code sample works for me:

```js
const glob = require('glob');
const path = require('path');

const projectPath = '/Users/simon/my-meteor-project/';

glob.sync(`${projectPath}**/*.test.js`).forEach((file) => {
  require(path.resolve(file));
});

**Comment by SimonSimCity** [comment]
This snippet should help you. Just put it into a file referenced by this feature:

```js
const glob = require('glob');
const path = require('path');

const projectPath = '/Users/simon/my-project-folder/';

glob.sync(`${projectPath}**/*.test.js`).forEach((file) => {
  require(path.resolve(file));
});
```

I just used the library `glob` here because it's easy - I guess you get the clue.

I had to define the full path because tests are run in a built meteor app - somewhere in a temporary directory. It includes only the files referenced in `meteor.testModule` or directly included there - all other files are stripped away.

Floriferous:
@SimonSimCity Looks great! I thought about using `glob` but didn't know how to tie things up with `require` and `path` like you did!

Unfortunately the `projectPath` is a blocker as we'd like it to work across developers. I guess we could each add our own paths in some env variable, but that's really hacky, and then there's the CI..

SimonSimCity:
@Floriferous You're on the wrong track here. It has nothing to do with `glob` or anything that you could install in your application as extension. The meteor application is compiled into a node-js application which is running somewhere else. It has no reference to the original project folder. That's what #10666 should bring.

You can - however - do this by setting an env variable yourself by calling `PROJECT_PATH=$PWD meteor test` instead of `meteor test` - but all this is now way off topic of this PR. If you want to know more, take a look at the libraries I linked in the issue and how they currently work around the limitation.

PR #9715, opened 2018-03-01
Use meteor.{mainModule,testModule} for `meteor create` starter apps.

benjamn:
In order for Meteor to maintain its commitment to being a zero-configuration tool, any configuration options that we add must come pre-configured in the best way possible for newly created apps.

> **Important note:** Meteor is still, and always will be, to its core, a zero-configuration tool. Though it may be tempting to slide down the slippery slope towards even more configuration options in the `meteor` section of `package.json`, we are not interested in allowing configuration complexity to explode, and we intend to maintain an extremely strict standard of necessity for any future options. Please don't think of the `meteor` section as a place to solve one-off problems with ad-hoc options.

In particular, the default new Meteor app must contain a reasonable testing story, or else we are signaling to the community that testing is an afterthought.

With that said, this PR is still a work in progress. I welcome your feedback on how best to configure the default `meteor create` starter app.

Builds on #9690 and #9714.

### DIFF
--- a/tools/static-assets/skel/package.json
+++ b/tools/static-assets/skel/package.json
@@ -2,10 +2,20 @@
   "name": "~name~",
   "private": true,
   "scripts": {
-    "start": "meteor run"
+    "start": "meteor run",
+    "test": "meteor test --driver-package dispatch:mocha-browser",
+    "test-app": "meteor test --full-app --driver-package dispatch:mocha-browser",
+    "visualize": "meteor --production --extra-packages bundle-visualizer"
   },
   "dependencies": {
-    "@babel/runtime": "^7.0.0-beta.36",
+    "@babel/runtime": "^7.0.0-beta.40",
     "meteor-node-stubs": "^0.3.2"
+  },
+  "meteor": {
+    "mainModule": {
+      "client": "client/main.js",
+      "server": "server/main.js"
+    },
+    "testModule": "test/main.js"
   }
 }
--- a/tools/static-assets/skel/package.json
+++ b/tools/static-assets/skel/package.json
@@ -3,8 +3,8 @@
   "private": true,
   "scripts": {
     "start": "meteor run",
-    "test": "meteor test --driver-package dispatch:mocha-browser",
-    "test-app": "meteor test --full-app --driver-package dispatch:mocha-browser",
+    "test": "meteor test --once --driver-package meteortesting:mocha",
+    "test-app": "TEST_WATCH=1 meteor test --full-app --driver-package meteortesting:mocha",
     "visualize": "meteor --production --extra-packages bundle-visualizer"
   },
   "dependencies": {
--- a/tools/static-assets/skel/test/main.js
+++ b/tools/static-assets/skel/test/main.js
@@ -1,0 +1,27 @@
+import assert from "assert";
+
+export const testMessage = "Welcome to Meteor!";
+
+describe("~name~", () => {
+  it("package.json has correct name", () => {
+    const { name } = require("../package.json");
+    assert.strictEqual(name, "~name~");
+  });
+
+  if (Meteor.isClient) {
+    it("client is not server", () => {
+      assert.strictEqual(Meteor.isServer, false);
+    });
+  }
+
+  if (Meteor.isServer) {
+    it("server is not client", () => {
+      assert.strictEqual(Meteor.isClient, false);
+    });
+  }
+
+  it("async/await and dynamic import()", async () => {
+    const tests = await import("./main.js");
+    assert.strictEqual(tests.testMessage, testMessage);
+  });
+});
--- a/tools/static-assets/skel/test/main.js
+++ b/tools/static-assets/skel/test/main.js
@@ -1,27 +1,20 @@
 import assert from "assert";
 
-export const testMessage = "Welcome to Meteor!";
-
-describe("~name~", () => {
-  it("package.json has correct name", () => {
-    const { name } = require("../package.json");
+describe("~name~", function () {
+  it("package.json has correct name", async function () {
+    const { name } = await import("../package.json");
     assert.strictEqual(name, "~name~");
   });
 
   if (Meteor.isClient) {
-    it("client is not server", () => {
+    it("client is not server", function () {
       assert.strictEqual(Meteor.isServer, false);
     });
   }
 
   if (Meteor.isServer) {
-    it("server is not client", () => {
+    it("server is not client", function () {
       assert.strictEqual(Meteor.isClient, false);
     });
   }
-
-  it("async/await and dynamic import()", async () => {
-    const tests = await import("./main.js");
-    assert.strictEqual(tests.testMessage, testMessage);
-  });
 });


tools/static-assets/skel/package.json
```diff
@@ -2,10 +2,20 @@
   "name": "~name~",
   "private": true,
   "scripts": {
-    "start": "meteor run"
+    "start": "meteor run",
+    "test": "meteor test --driver-package dispatch:mocha-browser",
```
benjamn: I wanted to use `meteortesting:mocha`, but my out-of-box experience was that the server restarted three times and tests didn't run in the browser because the app was "crashing."

tools/static-assets/skel/test/main.js
```diff
@@ -0,0 +1,27 @@
+import assert from "assert";
+
+export const testMessage = "Welcome to Meteor!";
+
+describe("~name~", () => {
+  it("package.json has correct name", () => {
```
hwillson: I know it doesn't really matter with these tests, but [arrow function use is discouraged by Mocha](https://mochajs.org/#arrow-functions). We might want to drop them here to set an example.

tools/static-assets/skel/package.json
```diff
@@ -2,10 +2,20 @@
...
-    "@babel/runtime": "^7.0.0-beta.36",
+    "@babel/runtime": "^7.0.0-beta.40",
     "meteor-node-stubs": "^0.3.2"
+  },
+  "meteor": {
+    "mainModule": {
+      "client": "client/main.js",
+      "server": "server/main.js"
+    },
+    "testModule": "test/main.js"
```
hwillson: Since the intent of the `meteor create` boilerplate is to serve as a jump start for new applications, I'm wondering if it might make sense to restructure things to create initially separated `client` / `server` test starting points. So something like:

```
  "meteor": {
    "mainModule": {
      "client": "client/main.js",
      "server": "server/main.js"
    },
    "testModule": {
      "client": "client/tests.js",
      "server": "server/tests.js"
    }
  }
```

The create app is already pushing people towards using the `client` / `server` separated approach (instead of lumping things together with `Meteor.isClient` / `Meteor.isServer`), so we might want to do the same with testing.

tools/static-assets/skel/package.json
hwillson: Right - to avoid this you have to set `TEST_WATCH=1` first. So: `TEST_WATCH=1 meteor test --driver-package meteortesting:mocha`

abernix: ...Which I believe necessitates [`cross-env`](https://www.npmjs.com/package/cross-env) on Windows.

tools/static-assets/skel/test/main.js
abernix: Perhaps use normal `function`s and put a comment under the first one to explain that it's done intentionally to give access to the Mocha API (for things like `this.wait(2000)`, etc.):

```js
// Using `function`s, rather than arrow-functions, allows access to
// Mocha's API.  See https://mochajs.org/#arrow-functions for details.
```

Otherwise the first reaction of many might be:

"Oh, what are these doing here?" => **replace with arrow functions**

tools/static-assets/skel/package.json
abernix: This is a reasonable point, though a `test` (or `tests`) root directory is a more common location in other app structures so perhaps `tests/client/main.js` and `tests/server/main.js` (and if someone wanted to put test helpers or common tests, they might gravitate toward putting them in the higher-level `tests/`.

abernix: Tangentially, after the introduction of `mainModule`, and prior to the `testModule` config in `package.json:meteor`, I was tempted to open the discussion about a `meteor.usePackageJsonEntryPoints` (or similarly themed) boolean config that simply used the `package.json` `main` and `browser` attributes, rather than necessitate the `mainModule` definitions, with the intention being to align with other parts of the ecosystem, and minimize the changes necessary for those making a switch (to or from).

It's worth reminding myself that `mainModule` offers more flexibility since it supports many architectures ("client", "server", "web", "web.browser", "web.cordova", "os"), but I can't help but think a user who had just `meteor create`d for the first time, might wonder why we didn't just use `main` and `browser`.

Ultimately, I'm not sure if a `usePackageJsonEntryPoints` would simplify this skeleton or not, but that certainly crossed my mind here again.

tools/static-assets/skel/package.json
hwillson: Thanks @abernix - and yuck 🙁. Maybe's there another alternative here then (like finding out why `meteortesting:mocha` is setup to work like this). I'm just a bit worried about putting `dispatch:mocha-browser` in, since it has been deprecated.

abernix: Agree. `meteortesting:*` has been working great overall.  Paging @aldeed for input!

hwillson: I took a quick look - `meteortesting:mocha` requires either the `--once` flag or `TEST_WATCH=1` to be set. Otherwise it does a [`process.exit(0)`](https://github.com/meteortesting/meteor-mocha/blob/1d8eacc1afdcaec5a25eef0416a9a153f4df4d09/package/server.js#L71) (after running the tests once and there are no failures), which the Tool then re-runs until it considers the app to be crashing.

If we just want to show that tests can be run once (via `meteor npm test`), then we can probably get away with using the `--once` parameter (so: `meteor test --once --driver-package meteortesting:mocha`). If we instead want to have things setup so `meteor npm test` runs the tests in watch mode, then maybe we should open a `meteortesting:mocha` PR to make watch mode the default.

tools/static-assets/skel/package.json
axelvaindal: If I may, I certainly agree with @abernix about the fact that using `mainModule` may appeared weird for users coming here for the first time and discovering entry points provided by NPM `package.json` configuration file are not used.
I consider myself used to Meteor now, and I also wondered why introducing `mainModule` instead of simply using `main` when I first saw the related PR.
Due to the the multitude of supported architectures, I think it makes sense to introduce the `meteor{mainModule, testModule}` configuration, because it would probably be referenced in the first line of the documentation/guide new users will surely read at the beginning of their journey with Meteor.

However, in order to allow advanced users to use meteor on specific environment with `main` entry point (like server only usage for instance), I would suggest allowing to configure Meteor to use it, as it should be a hand up to the ecosystem way of doing things for common usage of entry points. In the same idea, why not use simply `meteor{main,test}` to define these entry points with the same naming convention used by npm `package.json` ? 

 ```
 "meteor": {
    "useEntryPoints" : true, // true or omitted uses the main and test objects to configure entry points, false use root main
    "main": {
      "client": "client/main.js",
      "server": "server/main.js"
    },
    "test": {
      "client": "client/tests.js",
      "server": "server/tests.js"
    }
  }
```

File: tools/static-assets/skel/test/main.js (hunk 2 above)
Comment (benjamn) [reply tools/static-assets/skel/test/main.js]: I'm fine with just using normal functions and defending the choice later, if anyone asks.

File: tools/static-assets/skel/package.json (hunk 3 above)
Comment (benjamn) [reply tools/static-assets/skel/package.json]: The `mainModule` option was intended to mirror `api.mainModule` for Meteor packages, and `testModule` was intended to match `mainModule`. I don't think shortening their names has much benefit, because it would  make them harder to distinguish from other fields called `main` or `test` that might be meaningful outside of Meteor.

The `main` field would be tempting to re-use if it had any flexibility regarding client/server entry points, but it doesn't, because it's always just a single string. And the `browser` field is not something Node recognizes; it's a hack that client-side-only bundlers like Webpack and Browserify happen to use. I've [written at length](https://github.com/meteor/meteor/issues/6890#issuecomment-339817703) about my feelings on the `browser` field. I remain very reluctant to encourage its use in any way.

Another reason against using `main` is that it would suggest the Meteor app is somehow importable as a package in Node, or that importing the `main` module is the way to start the app, which is not accurate. That's the reason the application `package.json` never had a `main` field in the first place, and I don't think it makes sense to start now. The `meteor` section is clearly metadata for Meteor apps to use while you're running the `meteor` command-line tool. I'm pretty reluctant to be adding any configuration options at all, to be completely honest, so I really don't want to invest other parts of `package.json` with Meteor-specific meaning. Let's keep our designs confined to the `meteor` section if possible.

If we started using `main`, it wouldn't be long before people wondered about [the `module` field](https://github.com/rollup/rollup/wiki/pkg.module). While support for `module` is a reasonable thing to ask about, Meteor does not have the problem that the `main`/`module` disctintion was invented to solve. For Meteor, the `mainModule` and `testModule` options point to *source code* rather than compiled code, so there is no meaningful distinction between the CJS (`main`) entry point and the ESM (`module`) entry point. Using `main` to point to source code would be an abuse of what `main` means in Node.

If you think about it, the closest thing Meteor has to a Node-style `main` module is the `.meteor/local/build/main.js` file that Meteor runs internally when starting an app. Obviously I don't want to put `"main": ".meteor/local/build/main.js"` in `package.json`, but using `main` to point to anything else would be misleading/wrong.

In short, I think we're avoiding a lot of pointless complexity by focusing on a few simple but flexible Meteor-specific options in the `meteor` section of `package.json`, and I honestly think encouraging the use of `main` or `browser` would only constrain us in the future, without solving any problems in the present.

File: tools/static-assets/skel/test/main.js [hunk 4]
```diff
@@ -1,27 +1,20 @@
 import assert from "assert";
 
-export const testMessage = "Welcome to Meteor!";
-
-describe("~name~", () => {
-  it("package.json has correct name", () => {
-    const { name } = require("../package.json");
+describe("~name~", function () {
+  it("package.json has correct name", async function () {
+    const { name } = await import("../package.json");
```
benjamn: I realized this `package.json` test could be merged with the dynamic `import()` test below, so we don't have to call `require` at all, and we can show off dynamic `import()` more prominently.

PR #9719, opened 2018-03-05
Print nicer error message for missing Meteor package

mpowaga:
This PR is improved version of #9346 with all suggestions applied (see discussion).

### DIFF
--- a/packages/modules-runtime/server.js
+++ b/packages/modules-runtime/server.js
@@ -24,9 +24,9 @@ makeInstallerOptions.fallback = function (id, parentId, error) {
             `Cannot find package "${packageName}". ` +
             `Try "meteor add ${packageName}".`
           );
-        } else {
-          throw error;
         }
+
+        throw err;
       }
     }
   }
--- a/packages/modules-runtime/server.js
+++ b/packages/modules-runtime/server.js
@@ -15,7 +15,19 @@ makeInstallerOptions.fallback = function (id, parentId, error) {
   if (topLevelIdPattern.test(id)) {
     if (typeof Npm === "object" &&
         typeof Npm.require === "function") {
-      return Npm.require(id, error);
+      try {
+        return Npm.require(id, error);
+      } catch(err) {
+        if (id && id.startsWith('meteor/')) {
+          const [meteorPrefix, packageName] = id.split('/', 2);
+          throw new Error(
+            `Cannot find package "${packageName}". ` +
+            `Try "meteor add ${packageName}".`
+          );
+        } else {
+          throw error;
+        }
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


mpowaga:
Somehow tests are failing when run with `dispatch:mocha-phantomjs` driver. I'm not sure if this is something specific to test driver or phantomjs. I've run `mainModule` self-test with `meteortesting:mocha` driver and it worked fine.

Probably the best way to solve this issue is to not specify `makeInstallerOptions.fallback` on the client when phantomjs is detected (maybe by simply checking `window.navigator.userAgent`?).

Any thoughts?

mpowaga:
Ok I found the culprit. I've used es6 features which are not transpiled to es5 because ecmascript cannot be used (causes circular dependencies). Phantomjs does not support es6 so it fails to parse the file.

packages/modules-runtime/server.js
```diff
@@ -15,7 +15,19 @@ makeInstallerOptions.fallback = function (id, parentId, error) {
   if (topLevelIdPattern.test(id)) {
     if (typeof Npm === "object" &&
         typeof Npm.require === "function") {
-      return Npm.require(id, error);
+      try {
+        return Npm.require(id, error);
```
benjamn: One risk of wrapping `Npm.require` in a `try` block is that the required module might throw an unrelated error while it's loading, which would get caught here and turned into a missing module error, hiding the true cause.

Another approach would be to modify the error thrown [here](https://github.com/meteor/meteor/blob/a770e9959fec4e499a8f8c01dac5c211808c72e3/tools/isobuild/bundler.js#L1926-L1928), in the implementation of `Npm.require`.

However, perhaps a simpler approach would be to move the `id.startsWith("meteor/")` check earlier, just inside the `if (topLevelIdPattern.test(id)) {` block. Reasoning: `Npm.require` doesn't have any logic for importing from `meteor/...` modules, so if the `id` starts with `meteor/`, `Npm.require` is almost certainly going to fail.

mpowaga:
Good thought. I've made a change so it no longer fallbacks to `Npm.require` when meteor package is not found.

As for tests - what should they assert? There's already test which checks if an error is thrown when non existent meteor package is imported. Should we check if the error message is correct?

packages/modules-runtime/server.js
```diff
@@ -15,6 +15,14 @@ makeInstallerOptions.fallback = function (id, parentId, error) {
   if (topLevelIdPattern.test(id)) {
     if (typeof Npm === "object" &&
         typeof Npm.require === "function") {
+      if (id && id.startsWith('meteor/')) {
```
benjamn: This is looking better, but I think this check should be hoisted outside the `if (typeof Npm === ...` block, since the new error is relevant even if `Npm.require` is not defined?

mpowaga:
Moved this `if` statement to the outer scope. Though I'm not sure if it makes sense to move it even higher outside of `if (topLevelIdPattern.test(id)) {`.

Also, should `server.js` be written in es5? Since it's not transpiled to es6 it might cause syntax errors when executed with older versions of node.

benjamn:
No need to rewrite `server.js`, since it will only be used in Node 8. That's one of the advantages of shipping a specific version of Node with each release of Meteor: https://blog.meteor.com/know-your-node-js-ff0b3a0f634e

PR #9724, opened 2018-03-07
open Xcode project folder instead of project file

eskan:
Hi,
by using a pod-based cordova plugin (like phonegap-push-plugin) in a project, opening the configuration file instead of the folder prevents these pods from being included.
This PR change the way Xcode is opened.
Regards,

### DIFF
--- a/tools/cordova/run-targets.js
+++ b/tools/cordova/run-targets.js
@@ -60,7 +60,7 @@ function openXcodeProject(projectDir) {
 
 
   try {
-    execFileSync('open', ['-a', 'Xcode',projectDir]);
+    execFileSync('open', ['-a', 'Xcode', projectDir]);
 
     Console.info();
     Console.info(
--- a/tools/cordova/run-targets.js
+++ b/tools/cordova/run-targets.js
@@ -58,10 +58,9 @@ function openXcodeProject(projectDir) {
     return;
   }
 
-  const projectFilePath = files.pathJoin(projectDir, projectFilename);
 
   try {
-    execFileSync('open', [projectFilePath]);
+    execFileSync('open', ['-a', 'Xcode',projectDir]);
 
     Console.info();
     Console.info(


benjamn:
@skirunman @koenlav @wojtkowiak @macrozone Any thoughts on this? Seem safe?

KoenLav:
@eskan basically this PR enables you to (persistently) modify the XCode project by loading all resources in the project folder rather than only those specified in the project file (which, I assume, is automatically recreated upon every Meteor build)?

I am not sure whether that is desirable, for Meteor, as it would probably be preferable to open up ways of ensuring that the 'default build process of Meteor allows for such modifications'. Would it be weird to ask for some kind of (automated) way of modifying both the XCode project file and the AndroidManifest after it is generated by the normal Meteor build process?

I think it is already possible to do this for Android (using App.setPreference) in some, if not all, scenario's. Is the same/should the same not be possible for iOS?

eskan:
@KoenLav I’m not a IOS expert but the problem is around how Xcode is handle PODs. (I do not edit or modify project or install pods myself)
When i’m developing on IOS device i launch meteor with **meteor run —ios-device** and this launch Xcode project … but don’t ask me why Xcode doesn’t include PODs referred by plugin as it seems to be a subproject.

steps to reproduce :

```
meteor create meteor-xcode
cd meteor-code
meteor add-platform ios
meteor add cordova:phonegap-plugin-push@2.1.3
meteor run ios-device
```

it will launch Xcode for testing you need to have a provisioning profile to enable push notification (i don’t know other plugin that use cocoaPods)
When you build this project you’ll have a link error due to a missing library :

> ld: library not found for -lGoogleToolboxForMac

this is how xcode show project 

capture d ecran 2018-03-08 a 09 50 56


this library is installed with pods and is well installed in .local/cordova-build/platforms/ios/…

So i search the web (a lot :D ) and the way to get it works (a workaround ?) is to open xcode with the project folder. Then Pods is included is xcode and build works

capture d ecran 2018-03-08 a 09 58 37

Hope this help,

KoenLav:
@eskan that seems sensible!

We are also running into a problem with using meteor-launch (Fastlane) on CircleCI (https://github.com/johnthepink/meteor-launch) related to Pods, would be interested in finding out whether this also solves that problem.

@benjamn I'm not 100% sure, but I think this change makes sense.

benjamn:
Thanks @eskan @KoenLav and @skirunman. Let's get this into a 1.6.2 beta and make sure it works.

PR #9728, opened 2018-03-08
Improve CircleCI caching strategy

jamesmillerburgess:
While working on #9656, I noticed that we are using a single immutable cache on all builds, rather than updating the cache as dependencies change. This means that the 'Get Ready' job on CircleCI ends up downloading the dev bundle every single time and updating a lot of dependencies that probably haven't changed for months.

This PR implements a caching strategy using [templates](https://circleci.com/docs/2.0/caching/#using-keys-and-templates) that will only download a new dev bundle when the `meteor` file is changed (this is where the dev bundle version number is maintained), and which re-caches `.babel-cache` and `.meteor` on every run, but is loaded based on the branch.

This isn't an ideal strategy, but it saves roughly 2-3 minutes on the 'Get Ready' job compared to what we have currently.

For a subsequent PR, we could cache all of the `node_modules` folders in the various packages, as this takes up the majority of the remaining running time for 'Get Ready'.

I'm also not sure how much caching the `.babel-cache` and `.meteor` folders is speeding up our builds, so I will also run some comparison tests to see their impact and if they can be excluded from the cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,5 @@
 version: 2
 
-# These directories are cached across all builds, currently with no
-# hashing mechanism, but we should consider doing it off dev_bundle.
-meteor_cache_dirs: &meteor_cache_dirs
-  paths:
-    - "dev_bundle"
-    - ".babel-cache"
-    - ".meteor"
-
 # A reusable "run" snippet which is ran before each test to setup the
 # environment for user-limits, core-dumps, etc.
 run_env_change: &run_env_change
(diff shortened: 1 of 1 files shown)


benjamn:
Great work @jamesmillerburgess!

benjamn:
Wow, the **Get Ready** job is *much* faster now: ~3min instead of 9min30s! https://circleci.com/workflow-run/f06fc032-8a5b-414a-924b-d550138fbd58

jamesmillerburgess:
I think we can do even better by caching package NPM dependencies. I’ll
work on a PR for that soon.

PR #9740, opened 2018-03-12
Removed fiber wrapper around OAuth middleware.

MKRazz:
With commit 3b18863, connect handlers are guaranteed to run in a fiber, making the fiber wrapper in the OAuth middleware superfluous. Additionally, because it manually wraps the middleware in a Fiber directly, it is losing access to the properties stored in the existing fiber (meteor/meteor-feature-requests#156).

### DIFF
--- a/packages/oauth/oauth_server.js
+++ b/packages/oauth/oauth_server.js
@@ -139,14 +139,6 @@ OAuth._checkRedirectUrlOrigin = function (redirectUrl) {
 
 
 // Listen to incoming OAuth http requests
-WebApp.connectHandlers.use(function(req, res, next) {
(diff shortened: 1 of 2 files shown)


benjamn:
Nice catch!

MKRazz:
@benjamn I just realized, it doesn't look like Fiber is used anywhere else in that file, so the require can be removed as well. What's the best way to handle this, should I create a new pull request?

https://github.com/meteor/meteor/blob/b7ea61783886d495297c85367426c50f115275e7/packages/oauth/oauth_server.js#L1

benjamn:
Yes please! Can you also bump the `oauth` package version while you're at it?

PR #9748, opened 2018-03-15
[cordova] add resource-file to mobile-config

eskan:
Hi,
this PR allow to add resource files, see [cordova documentation](http://cordova.apache.org/docs/en/7.x/config_ref/index.html#resource-file)

there is a feature-request https://github.com/meteor/meteor-feature-requests/issues/279

the added function is 
`App.addResourceFile(src,target,[platform]);`

which will be translated in config.xml :
```
<platform name="android|ios|both">
  <resource-file src="src.file" target="target" />
</platform>
```
the **App.addResourceFile** _target_ parameter is mandatory here as the resource will be copied first at the root of the cordova project. Then cordova builder will copy the file to the target destination.

### Why resource-file is need ?

some plugin like [cordova-plugin-firebase](https://github.com/arnesson/cordova-plugin-firebase) 
 need configuration file which is use by Google Services.
here is how to set configuration file for this plugin :

1. Put both **google-services.json** and **GoogleService-Info.plist** in your meteor root project
2. Add these lines to your **mobile-config.js**, **APP_NAME** should need to be replaced without the real app name

```
App.addResourceFile('google-services.json', 'src/google-services.json', 'android');
App.addResourceFile('GoogleService-Info.plist', 'APP_NAME/Resources/GoogleService-Info.plist', 'ios');
```

this will add 
```
<platform name="ios">
    ...
   <resource-file src="GoogleService-Info.plist" target="APP_NAME/Resources/GoogleService-Info.plist"/>
</platform>
<platform name="android">
    ...
    <resource-file src="google-services.json" target="src/google-services.json"/>
</platform>
```

to the generated **config.xml**

### What is missing ?

Documentation needs to be updated

### DIFF
--- a/tools/cordova/builder.js
+++ b/tools/cordova/builder.js
@@ -53,7 +53,7 @@ const launchIosSizes = {
   'iphone6p_portrait': '1242x2208',
   'iphone6p_landscape': '2208x1242',
   'iphoneX_portrait': '1125x2436',
-  'iphoneX_landscape': '2436x1125', 
+  'iphoneX_landscape': '2436x1125',
   'ipad_portrait_2x': '1536x2048',
   'ipad_landscape_2x': '2048x1536',
   // Legacy
@@ -125,7 +125,7 @@ export class CordovaBuilder {
 
     // Custom elements that will be appended into config.xml's widgets
     this.custom = [];
-    
+
     // Resource files that will be appended to platform bundle and config.xml
     this.resourceFiles = [];
 
@@ -324,7 +324,11 @@ export class CordovaBuilder {
       this.configureAndCopyImages(launchAndroidSizes, platformElement.android, 'splash');
     }
 
-    this.configureAndCopyResourceFiles(this.resourceFiles, platformElement.ios, platformElement.android);
+    this.configureAndCopyResourceFiles(
+      this.resourceFiles,
+      platformElement.ios,
+      platformElement.android
+    );
 
     Console.debug('Writing new config.xml');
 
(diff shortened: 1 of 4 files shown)


benjamn:
@eskan This is looking great to me! Will you be able to update the documentation?

eskan:
@benjamn "Pas de soucis"
https://github.com/meteor/docs/pull/260

etyp:
@eskan this is awesome. In the meantime, is there another way to include `google-services.json` with Meteor Cordova override?

eskan:
@etyp  unfortunately not.

etyp:
Got it. I managed to add the following to `config.xml`:

```
<platform name="android">
    ...
    <resource-file src="google-services.json" target="google-services.json"/>
</platform>
```

And my `google-services.json` file by including both of them in the `cordova-build-override/` directory. It _seems_ like both were successfully included in `.meteor/local/cordova-build` but my push plugin still won't work.

Not sure if it's a separate issue, but @eskan is there any reason you can think of for that to not be a temporary workaround?

eskan:
hi @etyp, it's not a workaround to me as you need to put/change files in a temporary or autogenerated directory. But it could be a way to test notification with your plugin until this PR will be merged.
I get notifications work with the plugin and the configuration mentioned above.

tools/cordova/builder.js
```diff
@@ -375,6 +380,23 @@ export class CordovaBuilder {
     });
   }
 
+  configureAndCopyResourceFiles(resourceFiles, iosElement, androidElement) {
+    _.each(resourceFiles, resourceFile => {
+      // copy file in cordova project root directory
+      var filename = resourceFile.src.split('\\').pop().split('/').pop();
```
abernix: Hmm, is this splitting/popping on `resourceFile.src` in order to account for the different file path separators on Windows vs Unix, or is this because of something else I'm missing?

If it is because of such a reason, can we use Node's [`path.parse`](https://nodejs.org/api/path.html#path_path_parse_path) instead?  A backslash is an uncommon, but possible, character in a valid Unix path.

abernix:
Please ignore the "failing" tests.  They will go away if you rebase this off the current `devel` branch and push, but otherwise, it's simply a side-effect of us switching CircleCI configurations.

tools/cordova/builder.js
eskan: @abernix sorry for this ugly thing, i've updated the PR with path.parse()
regards

brucejo75:
Hey @eskan, is it possible to address #10684.

PR #9762, opened 2018-03-22
Remove oauth fiber import

MKRazz:
Fiber no longer needs to be imported because its only usage was removed. See https://github.com/meteor/meteor/pull/9740.


benjamn:
Thanks @MKRazz!

PR #9764, opened 2018-03-23
Do not report errors if _suppressSameNameError is true

pravdomil:
## Motivation

There is a popular question on StackOverflow [Get Meteor collection by name](https://stackoverflow.com/questions/10984030/get-meteor-collection-by-name). By far the most elegant way how to do that is just create new collection with same name and set _suppressSameNameError to true:
```js
new Mongo.Collection("test", { defineMutationMethods: false, _suppressSameNameError: true })
```
But if you do that you got warn in console `There is already a collection named "test"` even when you specify that you wan to suppress it.

This PR get's rid of warning since you specify to suppress it anyway.

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -264,7 +264,6 @@ Object.assign(Mongo.Collection.prototype, {
         // for now we will skip the error only when _suppressSameNameError
         // is `true`, allowing people to opt in and give this some real
         // world testing.
-        console.warn ? console.warn(message) : console.log(message);
       } else {
         throw new Error(message);
       }


axelvaindal:
@pravdomil 
I'm not familiar with this pattern, but based on your saying, doing the following code may result on the two variables pointing to the same instance of the collection, is that right ? 

```js
let tests = new Mongo.Collection("tests");
let test = new Mongo.Collection("tests", {defineMutationMethods: false, _suppressSameNameError: true});
```

If so, I understand it is convenient for getting the collection, but I'm really not a fan of using `_` defined variable, as in my mind, they are meant to be used solely inside the class or object.
If the MDG is okay with this pattern being the correct way of doing things when it comes to retrieve a collection from the `Mongo.Collection`, I think it should at least be documented [here](https://docs.meteor.com/api/collections.html).
But based on the commentaries above the code you modified, I assume this variable was meant to be field tested and the warning was intentional to remember people that using this kind of pattern is not fully supported for now.

Thus, this should be discussed before removing a warning indicating a lack of consensus over the pattern to use here.
If this pattern (creating another collection with the same name to get the instance) is accepted as a correct way of doing things, I would rather see a public API point `Mongo.Collection.get("test")` which would use the `_suppressSameNameError` on the background when instantiating, with the appropriate unit tests that seems to be missing on the actual usage of the variable (based on the commentary, again).
This approach would be much cleaner and solve both the warning issue and the lack of existing pattern, as well as the hole in documentation people want to fill in the Stack Overflow issue with a lot of different answers.

pravdomil:
> I'm not familiar with this pattern, but based on your saying, doing the following code may result on the two variables pointing to the same instance of the collection, is that right ?
correct

> If so, I understand it is convenient for getting the collection, but I'm really not a fan of using _ defined variable
me too

> If this pattern (creating another collection with the same name to get the instance) is accepted as a correct way of doing things, I would rather see a public API point Mongo.Collection.get("test")
thumbs up for Mongo.Collection.get

PR #9781, opened 2018-03-29
Remove Blaze repo based dependencies from static-html

hwillson:
I was playing around with the awesome new 1.6.2 `meteor create --minimal` feature, and noticed that `tracker` is being pulled into new minimal apps as a dependency. I was wondering why, and tracked the reason down to it being pulled in as one of the dependencies on the Blaze templating tools side. Currently the `static-html` package is dependent on several Blaze repo based packages, a lot of which include Blaze specific utilities / helpers, that are really not needed when using non-Blaze apps. This led me down a bit of a rabbit hole, but it turns out that the `static-html` package really only needs the following Blaze repo functionality to accomplish its role:

1. [`caching-html-compiler`](https://github.com/meteor/blaze/tree/555f3bb0d34c7553423e1f01f054189378e343ea/packages/caching-html-compiler)
2. The [`html-scanner.js`](https://github.com/meteor/blaze/blob/555f3bb0d34c7553423e1f01f054189378e343ea/packages/templating-tools/html-scanner.js) file from the [`templating-tools`](https://github.com/meteor/blaze/tree/555f3bb0d34c7553423e1f01f054189378e343ea/packages/templating-tools) package.
3. And of course itself, the [`static-html`](https://github.com/meteor/blaze/tree/555f3bb0d34c7553423e1f01f054189378e343ea/packages/static-html) package.

Since the above functionality isn't specific to Blaze, I'm thinking it might make sense to extract those files / packages from the Blaze repo, and bring them back into Meteor core. By doing so we can reduce a lot of template based dependency overhead, when not using Blaze.

This PR creates a new copy of the `static-html` and `caching-html-compiler` packages in core, and creates a new package called `html-scanner`, to house the `html-scanner.js` functionality from the `templating-tools` package. With these changes in place, we're able to replace the following dependencies in a `meteor create --minimal` app:

```
blaze-tools
caching-html-compiler
caching-compiler
deps
html-tools
htmljs
spacebars-compiler
static-html
templating-tools
tracker
```

with:

```
caching-html-compiler
html-scanner
static-html
```

(And this change will of course benefit non minimal `static-html` dependent apps as well.)

I've marked this PR as a work in progress, but I have verified it works properly by using it with a few apps. I would love to get feedback on this idea, to see if others agree it makes sense. If it does, then I'll finish polishing this PR up, and submit a sister PR over in the Blaze repo, to make sure Blaze is updated to use the new core packages (instead of using its own copies).

### DIFF
--- a/packages/caching-html-compiler/README.md
+++ b/packages/caching-html-compiler/README.md
@@ -1,0 +1,49 @@
+# caching-html-compiler
+
+Provides a pluggable class used to compile HTML-style templates in Meteor build
+plugins. This abstracts out a lot of the functionality you would need to
+implement the following plugins:
+
+1. `templating`
+2. `static-html`
+3. `simple:markdown-templating`
+
+It provides automatic caching and handles communicating with the build plugin
+APIs. The actual functions that convert HTML into compiled form are passed in
+as arguments into the constructor, allowing those functions to be unit tested
+separately from the caching and file system functionality.
+
+-------
+
+### new CachingHtmlCompiler(name, tagScannerFunc, tagHandlerFunc)
+
+Constructs a new CachingHtmlCompiler that can be passed into
+`Plugin.registerCompiler`.
+
+#### Arguments
+
+1. `name` The name of the compiler, used when printing errors. Should probably
+   be the same as the name of the build plugin and package it is used in.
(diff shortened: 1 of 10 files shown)


StorytellerCZ:
@hwillson any progress on this? This would be very much appreciated for pure React apps.

PR #9783, opened 2018-03-29
Update Node to version 8.11.1 (with v8 patch).

benjamn:
Supersedes https://github.com/meteor/meteor/pull/9725.

Node [just released version 8.11.0](https://nodejs.org/en/blog/release/v8.11.0/), a security update, which unfortunately did not contain the fix for the segmentation fault problems that we discovered in https://github.com/meteor/meteor/pull/9725 and worked with them to fix in https://github.com/nodejs/node/issues/19274. That fix will presumably come in Node 8.11.1.

In the meantime, we've built a version of Node 8.11.0 with just that one fix applied: https://github.com/meteor/node/commits/v8.11.0-meteor

This should allow us to move to release candidate testing for Meteor 1.6.2 while we wait for the official Node 8.11.1 release. 🤞⏳


benjamn:
Update: Node has already released an 8.11.1 version with [no additional commits](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V8.md#2018-03-29-version-8111-carbon-lts-mylesborins), so I guess we'll be waiting for 8.11.2 instead!

benjamn:
Superseded by #9789.

PR #9785, opened 2018-03-30
Cache npm dependencies from meteor packages in CircleCI

jamesmillerburgess:
The 'Get Ready' job in our CircleCI workflow spends about 90 seconds on every run reinstalling npm dependencies from meteor packages. This PR caches them based on a combined checksum of their various npm-shrinkwrap files.

This isn't the most elegant solution, and I don't like that I've hard-coded a bunch of folders into the `save_cache` step, but I couldn't think of a better way.

I've labeled this 'WIP' because I can't figure out where the `node_modules` folder is for `spacebars-compiler`, which is causing 'Get Ready' to reinstall `uglify-js` every run. @abernix Do you know where this is hiding?

Some notes:

- The dependencies are split into two caches, because when they are combined I consistently get what looks like an AWS error about metadata being [too large](https://circleci.com/gh/jamesmillerburgess/meteor/2030).
- I fixed a bug I came across in my test load-balancer that appears if you comment out workflow jobs without adjusting the environment variable for the number of test groups


abernix:
@jamesmillerburgess Thanks for looking into this!

First, regarding the path to the `node_modules` for `spacebars-compiler`, I think the path you have is just off _slightly_, missing a `packages/` sub-directory under `blaze`.  The "root" of that (compiler package) npm install would be [here](https://github.com/meteor/blaze/tree/master/packages/spacebars-compiler/.npm/package), so I'd expect the `node_modules` to be:

    packages/non-core/blaze/packages/spacebars-compiler/.npm/package/node_modules

In your commit, the path is currently:

    packages/non-core/blaze/spacebars-compiler/.npm/package/node_modules

Hopefully that's it?

Second, I'm curious what the overall win here is with this PR (i.e. how much faster?).  As another suggestion, which could be less verbose, the latest versions of npm have fairly substantial cache improvements through their use of [`cacache`](https://www.npmjs.com/package/cacache).   If we merely cached the `.npm/_cacache` directory (which for a typical npm installation would live at `$HOME/.npm/_cacache`), that might actually exhibit a similar (less configurationally-complicated) win?  It's probably worth caching the entire `.npm` directory, if we're not doing this already (I'm not sure if we're currently caching this for CircleCI builds; might be worth looking into).

Building on that investigation, I'm not positive if the Meteor npm installation environment configuration ([which takes place here](https://github.com/meteor/meteor/blob/devel/tools/cli/dev-bundle-bin-helpers.js#L63-L112)) is for sure pulling from the same global npm cacache directory.  That'd be nice to know!  If it is, then maybe we could just cache `$HOME/.npm` and be done with it.

abernix:
> I fixed a bug I came across in my test load-balancer that appears if you comment out workflow jobs without adjusting the environment variable for the number of test groups

Could you actually split this fix to the test-balancing into another PR?  Seems like this should probably be its own commit and land prior to additional discussion here.

jamesmillerburgess:
@abernix Thanks for the feedback!

My estimate on the gain is about 90 seconds, and since this is in 'Get Ready' and not in one of the parallel jobs, the time comes off of the total CircleCI build.

I've adjusted the `spacebars-compiler` folder and split out the bug fix into another PR.

Currently looking into caching the npm cache to see how that works.

jamesmillerburgess:
@abernix I tried caching `~/.npm`, but the folder was only 50kb and didn't impact the installation time of dependencies. I'll keep looking to see if there is a cache somewhere else we can use, or if I made a mistake somewhere.

benjamn:
We might be able to use [the new `npm ci` command](http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) to install directly (and quickly) from a `package-lock.json` file? I wonder how that works if the contents of `node_modules` are already installed and mostly correct (due to the caching).

jamesmillerburgess:
@benjamn It looks like I submitted this right around the time the npm upgrade PR got merged, so I didn't think we had access to `npm ci` yet.

I will see if there is any impact from switching to `npm ci` on CircleCI, but even if there aren't big gains when we have the dependencies cached, it should still speed up local installs.

jamesmillerburgess:
Switching to `npm ci` by editing [this line](https://github.com/meteor/meteor/blob/bd26ce3d1e50f37fa4228923f7c11b164fedc0fb/tools/isobuild/meteor-npm.js#L1133) throws the following error in `meteor --get-ready`:

```
$ ./meteor --get-ready
Errors prevented isopacket build:

While loading plugin `compile-ecmascript` from package `ecmascript`:
module.js:547:15: Cannot find module 'double-ended-queue'
at Function.Module._resolveFilename (module.js:547:15)
at Function.Module._load (module.js:474:25)
at Module.require (module.js:596:17)
at require (internal/module.js:11:18)
at Object.<anonymous>
(/Users/goldfibre/projects/meteor/packages/meteor/.npm/package/node_modules/meteor-deque/index.js:1:18)
at Module._compile (module.js:652:30)
at Module._extensions..js (module.js:663:10)
at Object.require.extensions..js
(/Users/goldfibre/projects/meteor/dev_bundle/lib/node_modules/meteor-babel/register.js:50:5)
at Module.load (module.js:565:32)
at Module.Mp.load (/tools/tool-env/install-runtime.js:16:29)
at tryModuleLoad (module.js:505:12)
at Function.Module._load (module.js:497:3)
at Module.require (module.js:596:17)
at require (internal/module.js:11:18)
at Object.require (/tools/isobuild/bundler.js:1900:22)
at packages/meteor.js:736:32
at packages/meteor.js:907:4
at packages/meteor.js:1438:3


/Users/goldfibre/projects/meteor/dev_bundle/lib/node_modules/meteor-promise/promise_server.js:218
      throw error;
      ^

Error: isopacket build failed?
    at Object.ensureIsopacketsLoadable (/tools/tool-env/isopackets.js:229:11)
    at /tools/cli/main.js:865:40
```

I tried adding a double-ended-queue entry to the shrinkwrap file for `babel-runtime`, which is the `compile-ecmascript` plugin in `ecmascript`, but that didn't work.

I also tried adding the same thing to the `meteor` package, because that is where the `meteor-deque` package lives - the only package in the directory tree that requires `double-ended-queue`. That also didn't work... 🤔

benjamn:
Ok, thanks for looking into that. If we avoid `npm ci`, is this PR still a work in progress?

jamesmillerburgess:
I need to remove the ‘wip’ tags from the cache keys, and then we should be
good to go! I’ll try to get to that today.

jamesmillerburgess:
Assuming all the tests pass, this should be good to go!

PR #9786, opened 2018-03-30
Release 1.4.4.6

abernix:
**Including**:
* Node 4.9.0: https://nodejs.org/en/blog/release/v4.9.0/
* Npm 4.6.1 (last v4.x version): https://github.com/npm/npm/releases/tag/v4.6.1

**Additional back-ports**:
* https://github.com/meteor/meteor/pull/8991
* https://github.com/meteor/meteor/pull/8859


benjamn:
Did the `semver` update make it into the new dev bundle? The chronology of these commits is a bit confusing if you’re looking at it on the web.

abernix:
Yes, it was included!  I agree the order of the commits here was confusing; I should have rebased to make that clear.  Just for the sake of posterity, it can be confirmed that `dev-bundle-4.7.30` (which this Meteor 1.4.4.6 uses) included this via the branch tag for [`dev-bundle-4.7.30`](https://github.com/meteor/meteor/tree/dev-bundle-4.7.30) which points to c9bee61 above.

---

For the sake of the linearity, I tried to fix this and had rebuilt a `dev-bundle-4.7.31` (same as 4.7.30) and prepared and `rc.1` but looks like you confirmed it already and published 1.4.4.6 (all good!) while I was on a walk waiting for the build.  This does mean that `1.4.4.6-rc.1` exists now, but that should be harmless.

PR #9790, opened 2018-04-01
Update MongoDB driver to 3.0.5

klaussner:
This PR updates the MongoDB driver from 2.2.34 to 3.0.5 and addresses the breaking changes described in [`CHANGES_3.0.0.md` (`node-mongodb-native`)](
https://github.com/mongodb/node-mongodb-native/blob/3.0.0/CHANGES_3.0.0.md) and https://github.com/meteor/meteor-feature-requests/issues/268#issuecomment-366048866.

**Summary of the changes:**

* Create the `Db` instance manually because `MongoClient` passes a client instance instead of a database to the callback function. I'm using the undocumented `client.s.options.dbName` property to get the database name because I couldn't find better way to access it without parsing the URL in the `MongoConnection` constructor.

* Use `Cursor.prototype.next` instead of `Cursor.prototype.nextObject` (deprecated in version 2 and removed in version 3).

* Use `client` instead of `db` to close the connection in `MongoConnection.prototype.close`.

* Use the `projection` option instead of the `fields` parameter in `MongoConnection.prototype._createSynchronousCursor`.

I'll bump the versions and add an entry to `History.md` when the PR is ready to be approved. :slightly_smiling_face: :crossed_fingers:

Fixes meteor/meteor-feature-requests#268.

### DIFF
--- a/tools/runners/run-mongo.js
+++ b/tools/runners/run-mongo.js
@@ -585,21 +585,24 @@ var launchMongo = function (options) {
   var initiateReplSetAndWaitForReady = function () {
     try {
       // Load mongo so we'll be able to talk to it.
-      const { Db, Server } = loadIsopackage('npm-mongo').NpmModuleMongodb;
+      const {
+        MongoClient,
+        Server
+      } = loadIsopackage('npm-mongo').NpmModuleMongodb;
 
       // Connect to the intended primary and start a replset.
-      var db = new Db(
-        'meteor',
+      const client = new MongoClient(
         new Server('127.0.0.1', options.port, {
           poolSize: 1,
           socketOptions: {
             connectTimeoutMS: 60000
           }
-        }),
-        { safe: true }
+        })
       );
 
-      yieldingMethod(db, 'open');
+      yieldingMethod(client, 'connect');
+      const db = client.db('meteor');
+
       if (stopped) {
         return;
       }
@@ -711,7 +714,7 @@ var launchMongo = function (options) {
         break;
       }
 
-      db.close(true /* means "the app is closing the connection" */);
+      client.close(true /* means "the app is closing the connection" */);
     } catch (e) {
       // If the process has exited, we're doing another form of error
       // handling. No need to throw random low-level errors farther.
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1345,7 +1345,7 @@ Ap.insertUserDoc = function (options, user) {
   } catch (e) {
     // XXX string parsing sucks, maybe
     // https://jira.mongodb.org/browse/SERVER-3069 will get fixed one day
-    if (e.name !== 'MongoError') throw e;
+    if (e.name !== 'MongoError' && e.name !== 'BulkWriteError') throw e;
     if (e.code !== 11000) throw e;
     if (e.errmsg.indexOf('emails.address') !== -1)
       throw new Meteor.Error(403, "Email already exists.");
--- a/packages/allow-deny/allow-deny.js
+++ b/packages/allow-deny/allow-deny.js
@@ -192,7 +192,11 @@ CollectionPrototype._defineMutationMethods = function(options) {
             throw new Meteor.Error(403, "Access denied");
           }
         } catch (e) {
-          if (e.name === 'MongoError' || e.name === 'MinimongoError') {
+          if (
+            e.name === 'MongoError' ||
+            e.name === 'BulkWriteError' ||
+            e.name === 'MinimongoError'
+          ) {
             throw new Meteor.Error(409, e.toString());
           } else {
             throw e;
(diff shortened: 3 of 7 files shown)


klaussner:
Update:

**Meteor Tool**

The Meteor Tool used `new Db(...).open()` (removed in the new driver version) to connect to the database, so I changed it to use the new `MongoClient` construct instead.

**Database Name**

A fix for the [database name problem](https://jira.mongodb.org/browse/NODE-1258) will be included in the next version of `node-mongodb-native` (3.0.6). Calling `client.db()` (without the `client.s.options.dbName` argument) will create a `Db` object for the database from the connection URL, so we don't have to rely on an undocumented property.

**MongoDB Errors**

There are two locations in the Meteor codebase ([`accounts-base`](https://github.com/meteor/meteor/blob/953fef784807efc3fd9a3addad5145493b18568c/packages/accounts-base/accounts_server.js#L1348) & [`allow-deny`](https://github.com/meteor/meteor/blob/953fef784807efc3fd9a3addad5145493b18568c/packages/allow-deny/allow-deny.js#L195)) where an error's `name` property is checked to find out if the error was thrown by MongoDB. In previous versions of the driver, most (or all?) MongoDB errors had `name === "MongoError"` but the new version sometimes throws errors with `name === "BulkWriteError"`, for example, if a duplicate key is detected when calling `insert`. I'm not sure why the driver throws a `BulkWriteError` in this case because `insert` is not a bulk operation. :thinking:

I'm comparing `name` with both error names for now (https://github.com/meteor/meteor/pull/9790/commits/5286d14ae4b6353113ba4c945f4af497988b2eb2) but maybe we can find a more robust way to identify MongoDB errors without checking the `name` property.

tools/runners/run-mongo.js
```diff
@@ -585,21 +585,24 @@ var launchMongo = function (options) {
...
-      var db = new Db(
-        'meteor',
+      const client = new MongoClient(
         new Server('127.0.0.1', options.port, {
           poolSize: 1,
           socketOptions: {
             connectTimeoutMS: 60000
           }
-        }),
-        { safe: true }
```
klaussner: The `safe` option of the `Db` constructor was deprecated a long time ago (in driver version 1: https://web.archive.org/web/20130408033926/http://mongodb.github.io:80/node-mongodb-native/api-generated/db.html) :older_woman:

abernix: I know it was removed from the documentation in the Node driver _long ago_, but I'm concerned that this still isn't a no-op.  Unfortunately, I can't find the exact scenario where I looked into this, but I do remember digging into this in the past (with a similar revelation around the seeming deprecation of the `safe` option) long ago and being concerned by the fact that there seemed to be some support for it still in the connection string URI ([for example](https://github.com/mongodb/node-mongodb-native/search?utf8=%E2%9C%93&q=safe&type=)).

The `safe` option once related to the write-concern, but I'm not completely confident in removing this if it's possible it's still being passed through to Mongo itself in some way.  Can we confirm that it's not affecting the replica set configuration in any way (for example, does Mongo directly copy these to the connection string somewhere?) Do we need to more explicitly set a write-concern if we remove this? Should that be left as an implementation exercise to the developer?

benjamn:
@klaussner Can you add a bullet point about this change to `History.md`?

tools/runners/run-mongo.js
abernix: Ok, it appears that there's no way that `safe` is heading upstream to the core Mongo driver.  I've concluded this based on [this line](https://github.com/mongodb/node-mongodb-native/blob/b8484257e16ab129b5d1614f6c9fbf933846814d/lib/db.js#L162
) in the Mongo Node driver `Db` instantiation which uses a `filterOptions` with a so-called "`legalOptions`" whitelist which is [defined here](https://github.com/mongodb/node-mongodb-native/blob/b8484257e16ab129b5d1614f6c9fbf933846814d/lib/db.js#L82-L111) and has no sign of `safe`.

klaussner: That's good to know! I totally overlooked the `filterOptions` call. :eyes:

I'm not sure if I'm interpreting the [docs](https://docs.mongodb.com/manual/release-notes/drivers-write-concern/) and the driver's commit history correctly, but it looks like the `safe` option was replaced with `w = 1`, which is now the default write concern, so all operations should be "safe" automatically.

klaussner:
Thanks! Should I bump the version numbers of the changed packages (`mongo`, `accounts-base`, and `allow-deny`)?

abernix:
@klaussner Did you have an opportunity to look at how connectivity worked with pre-3.6 Mongo deployments?  Particularly curious if we should be concerned by the comment raised in https://github.com/meteor/meteor/pull/9632#issuecomment-390459882.

klaussner:
It should work with all Mongo deployments because the latest driver supports all versions from 2.6 to 3.6. I made a few quick tests with two older versions (2.6 and 3.4) and didn't notice any problems. 🤞

PR #9799, opened 2018-04-05
Protect against missing results files in test load balancer

jamesmillerburgess:
Split from #9785.

This PR catches an error in the test load-balancer script that manifests if there is a mismatch between the CircleCI environment variable `NUM_GROUPS` and the actual number jobs in the workflow that are running test groups.

This can happen if you are testing from a checkout and comment out jobs in order to speed up CircleCI turnaround.


jamesmillerburgess:
@abernix Re-running the build fixed it on my last PR, but it would be great if we didn't have to do that frequently. I don't have any ideas right now on how to fix this systematically though...

Re-running now!

abernix:
@jamesmillerburgess Luckily, I don't think it had anything to do with anything we've done.

PR #9814, opened 2018-04-14
Switch Meteor's headless browser tests from PhantomJS to Headless Chrome

toinevk:
Per [254](https://github.com/meteor/meteor-feature-requests/issues/254), this is an attempt to replace PhantomJS with Headless Chrome.

I evaluated a few different solutions notably:
- Puppeteer 
- Chrome Launcher + Chrome Remote Interface 
- Selenium

Ultimately, Puppeteer felt the best fit as it maintained by the Chrome team and it
downloads a recent version of Chromium that is guaranteed to work with the API. The only downside is that creates a fresh download (~170Mb Mac, ~282Mb Linux, ~280Mb Win) with every instance of the client. Chrome Launcher will instead find where Chrome is installed on your local file system and  set up a debug instance. However, with this method the self-tester would assume that an instance of Chrome is installed that is compatible with the API.

There are some outstanding to dos, but before I proceed thought I would seek some feedback given Im not 100% familiar with Meteor homegrown test runner.   

Outstanding To Dos:
- [x] Add history entry
- [x] dynamics-imports, modules and app-config tests all use dispatch-mocha-phantom-js.
- [x] packages/test-in-console - Uses PhantomJS this needs to be removed.


abernix:
This looks like a great direction.  Thanks for taking the time to work on this!

The one thing we concretely rely on Phantom for, particularly as of Meteor 1.6.2 (soon to be 1.7!), is that it falls into the category of a "legacy browser" (that is to say, it lacks many "modern" browser features like `async`/`await`) which allows us to test the new legacy bundle features which 1.6.2 (now, 1.7) touts. (#9439) I still think it's probably worth moving forward on this this change though, and I like the direction and thoroughness you've got going here.

Do you think we could keep `tools/tool-testing/clients/phantom/index.js` (i.e. the `PhantomClient` class) in place, but make it so Puppeteer was the default?  I think we can just keep the `--phantom` flag ([here](https://github.com/meteor/meteor/pull/9814/commits/007d11ef261096548278454e3cd39d0d0470e895#diff-3ec37939b1d93936651067c04bab03b5L2271) for the moment and simply add the `--puppeteer` flag as an addition.  Then we can consider removing it later.

Overall, this PR would be best paired with the work I did in #9364, but try to get over the obstacle I encountered in https://github.com/meteor/meteor/pull/9364#issuecomment-348003226.  If we got that working, then we could almost confidently remove Phantom, though it might still be worth keeping around for someone wanting to test "legacy browser" features without relying on an "online" service like BrowserStack.

Thanks again for diving into this!  (Happy to help more.)

toinevk:
Thanks @abernix, I have reverted the change and you can now pass --phantom to run test in the legacy browser. I have also updated the test-in-console runner to use puppeteer instead, note I had to bump the node version in Travis given puppeteer relies on at least version 6.4.0. I have pinned it at the current supported version for Meteor. 

There is a single failure with a Blaze test on TravisCI, I dont have a-lot of familiarity here so I will need to unpack that to see if its valid...

abernix:
@toinevk Thanks for looking into this.

When I run `./packages/test-in-console/run.sh` locally (the command ran [here in `.travis.yml`](https://github.com/meteor/meteor/blob/1c0ebbf2f0a8db5939b1bd264c73e7717a2c8528/.travis.yml#L8)), I encounter the same failure:

#### Error
```
C: tinytest - spacebars-tests - template_tests - javascript scheme urls : !!!!!!!!! FAIL !!!!!!!!!!!
```

#### Formatted Output
```json
{
  "groupPath": ["tinytest", "spacebars-tests", "template_tests"],
  "test": "javascript scheme urls",
  "events": [{
    "sequence": 12317,
    "type": "fail",
    "details": {
      "type": "string_equal",
      "expected": "",
      "actual": "http://localhost:4096/"
    },
    "cookie": {
      "name": "spacebars-tests - template_tests - javascript scheme urls",
      "offset": 2,
      "groupPath": ["tinytest", "spacebars-tests", "template_tests"],
      "shortName": "javascript scheme urls"
    }
  }]
}
```

Which indicates that the error is happening when Puppeteer is running tests via the Chrome Inspector Protocol against its version of Chrome, which uses Chrome 67 (the so-called "Dev" Channel).  (See chart of channel status here: https://googlechromelabs.github.io/current-versions/)

Then, instead of running the `run-in-console` version of the tests, I switched to using the normal "run-in-browser" technique by running:

    ./meteor test-packages -p 4096 ./packages/non-core/blaze/packages/spacebars-tests

I checked `http://localhost:4096/` locally with what was my "Stable" channel (v65, before I restarted — _oops, outstanding updates!_) and the tests passed.  But when I restarted Chrome to get the update, and advanced to v66, the tests started failing.

So basically, I think something may be changing/has changed in Chrome.  If you're not experiencing the failure locally, I would try making sure that you've updated your Git submodules for Blaze locally, by running:

(You can check if you _were_ on the wrong commit by going into `packages/non-core/blaze` and running `git status`, or by running `git -C packages/non-core/blaze status` from the root of the `meteor` checkout).  The correct Git hash should be `8f819731df1fce254825c236176cfa5acb2df869`.

toinevk:
Thanks @abernix ! That information was super helpful _(my version was also out of date...)_ - I managed to identify the issue and have submitted a PR at https://github.com/meteor/blaze/pull/284 which goes into the detail. Once this is merged these tests should pass again.

packages/test-in-console/puppeteerRunner.js
```diff
@@ -0,0 +1,71 @@
...
+  async function poll() {
+    if (await isDone(page)) {
+      let failCount = await getFailCount(page);
+      if (failCount > 0) {
+        await page.close();
+        await browser.close();
+        process.exit(1);
+      } else {
+        await page.close();
+        setTimeout(runNextUrl, 1000);
```
abernix: I suspect that this `runNextUrl` function being called without a `browser` is the likely cause of the failure identified in https://github.com/meteor/meteor/pull/9814#issuecomment-397271476 and shown in [this test run](https://travis-ci.org/meteor/meteor/builds/392210478).

toinevk:
Thanks @abernix, Ill put that down to late night coding... calling nextURL function was a hangover from the way PhantomJS is designed - its not needed for Puppeteer. Travis is passing now so hopefully CircleCI will too and it then should be good to go.

abernix:
@toinevk I made a few changes above (and via https://github.com/meteor/circleci/pull/1) in order to incorporate the additional packages into our existing CircleCI base image.

I thought we were pretty much ready to merge, but I re-enabled the `--phantom` flag for all CircleCI runs (with https://github.com/meteor/meteor/pull/9814/commits/a0578e90e3f5df76a2c51aef5c9639be0a9707b5), to ensure that we continue to test under both Phantom and Chrome per the [above reasons](https://github.com/meteor/meteor/pull/9814#issuecomment-382494935), but it seems that's caused the failures above.  Any ideas?

benjamn:
@abernix If you agree this is ready to go, and you have time tomorrow, could you merge this?

benjamn:
Thanks so much for this substantial and valuable work @toinevk!

abernix:
The final adjustments look great.  I'd like to re-iterate the "thanks" that @benjamn has already stated above: Thank you very much, @toinevk!

PR #9815, opened 2018-04-14
[facebook-oauth] Update Facebook API to v2.12

StorytellerCZ:
Updates calls to Facebook API to use the newest 2.12 version which is set as the only option for newly created apps on the Facebook Developer interface.


benjamn:
Thanks @StorytellerCZ!

Issue #9827, opened 2018-04-19
1.7-beta.22: MongoError: database name must be a string


sebakerckhof:
Getting this error with 1.7-beta.22. Even after a meteor reset.
Probably related to upgrading the mongo driver to v3. Ping @klaussner 
```
W20180419-11:00:47.787(2)? (STDERR) MongoError: database name must be a string
W20180419-11:00:47.787(2)? (STDERR)     at Function.MongoError.create (/home/seba/.meteor/packages/npm-mongo/.2.2.34.1t0ofc9.mme3++os+web.browser+web.cordova/npm/node_modules/mongodb-core/lib/error.js:31:11)
W20180419-11:00:47.787(2)? (STDERR)     at validateDatabaseName (/home/seba/.meteor/packages/npm-mongo/.2.2.34.1t0ofc9.mme3++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/db.js:1680:57)
W20180419-11:00:47.787(2)? (STDERR)     at new Db (/home/seba/.meteor/packages/npm-mongo/.2.2.34.1t0ofc9.mme3++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/db.js:156:3)
W20180419-11:00:47.788(2)? (STDERR)     at Db.db (/home/seba/.meteor/packages/npm-mongo/.2.2.34.1t0ofc9.mme3++os+web.browser+web.cordova/npm/node_modules/mongodb/lib/db.js:1208:12)
W20180419-11:00:47.788(2)? (STDERR)     at packages/mongo/mongo_driver.js:180:25
W20180419-11:00:47.788(2)? (STDERR)     at runWithEnvironment (packages/meteor.js:1241:24)
```

jorisroling:
+1

klaussner:
I could reproduce the error with:

```
meteor create --full --release 1.7-beta.22 issue-9827
cd issue-9827
meteor
```

But it worked when I ran Meteor from a checkout ([this commit](https://github.com/meteor/meteor/commit/5189713d01ac32bb5e2f57078ed6ffaa558685f5) on the `release-1.7` branch). :thinking:

thhoang99:
+1

jeffreyflynt:
I am getting this issue as well.

benjamn:
Yep, this is super reproducible! Fix coming in `1.7-beta.23` 🤞

benjamn:
[...]

PR #9831, opened 2018-04-20
Update MongoDB driver to 3.0.7

klaussner:
Since version 3.0.6 of the MongoDB driver, the internal `client.s` property is no longer necessary to create the database object. If `MongoClient.prototype.db` is called without arguments, it will use the database name from the connection URI (https://jira.mongodb.org/projects/NODE/issues/NODE-1258).

This reverts 567e620a867a5e34188024d3175d4cc9dedaff3b, see https://github.com/meteor/meteor/issues/9827#issuecomment-383107731.

@abernix I had to target `release-1.7` instead of `devel` because the driver update from #9790 was merged into `release-1.6.2` and isn't on `devel` yet. :upside_down_face:

### DIFF
--- a/History.md
+++ b/History.md
@@ -140,7 +140,8 @@
   [PR #9632](https://github.com/meteor/meteor/pull/9632)
 
 * The `mongodb` driver package has been updated from version 2.2.34 to
-  version 3.0.5. [PR #9790](https://github.com/meteor/meteor/pull/9790)
+  version 3.0.7. [PR #9790](https://github.com/meteor/meteor/pull/9790)
+  [PR #9831](https://github.com/meteor/meteor/pull/9831)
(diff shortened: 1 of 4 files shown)


packages/mongo/package.js
```diff
@@ -9,7 +9,7 @@
 
 Package.describe({
   summary: "Adaptor for using MongoDB and Minimongo over DDP",
-  version: '1.5.0-beta17.23'
+  version: '1.5.1-beta17.23'
```
benjamn: Let's revert this change back to `1.5.0-beta17.23`, since 1.5.0 hasn't been released yet, and won't be released until Meteor 1.7 is released. Generally, when we bump the minor or major version of a package as part of a Meteor release, any changes to that package can be lumped into that one version.

PR #9849, opened 2018-04-29
Stop prefixing legacy URLs with /__browser.legacy

benjamn:
Inspired by discussion with @CaptainN in #9776, I think we can (and should) stop relying on URL prefixes to determine whether to load static files from the `web.browser` (modern) or `web.browser.legacy` (legacy) architectures. If this works, it will be a major ergonomic simplification for the modern/legacy system coming in Meteor 1.7.

This PR makes `webapp` serve different static files based on `isModern(request.browser)`, which is something we already have to do without the help of URL prefixes [in the `dynamic-import` package](https://github.com/meteor/meteor/blob/8e6f673fd066a69b2b4779526d200a7be0f1fc73/packages/dynamic-import/server.js#L80-L82).

Note that `web.cordova` URLs will still be prefixed with `/__cordova/`, as they always were, even before #9439 split `web.browser` and `web.browser.legacy`. Additionally, URL prefixes like `/__browser/` and `/__browser.legacy/` can still be used explicitly to request resources from non-default architectures.

It feels a little weird for the server to respond with different resources for the same URL, depending on the user agent string, but any given client will only see one version of the resource for a given URL, so clients can still safely cache the resources they receive. Furthermore, JS and CSS resources are always requested using a hash of the resource's contents (in both development and production), so the modern and legacy URLs for those kinds of resources will be different anyway.

### DIFF
--- a/packages/non-core/bundle-visualizer/server.js
+++ b/packages/non-core/bundle-visualizer/server.js
@@ -33,21 +33,38 @@ function getStatBundles() {
     f.absolutePath.endsWith(".stats.json");
 
   // Read the stat file, but if it's in any way unusable just return null.
-  const readOrNull = file => {
+  function readOrNull(file) {
     try {
       return JSON.parse(fsReadFileSync(file, "utf8"));
     } catch (err) {
       return null;
     }
-  };
+  }
 
-  return Object.keys(WebAppInternals.staticFiles)
-    .map(staticFile => WebAppInternals.staticFiles[staticFile])
-    .filter(statFileFilter)
-    .map(statFile => ({
-      name: statFile.hash,
-      stats: readOrNull(statFile.absolutePath),
-    }));
+  const {
+    staticFiles,
+    staticFilesByArch,
+  } = WebAppInternals;
+
+  const files = [];
+
+  if (staticFilesByArch) {
+    Object.keys(staticFilesByArch).forEach(arch => {
+      const staticFiles = staticFilesByArch[arch];
+      Object.keys(staticFiles).forEach(path => {
+        files.push(staticFiles[path]);
+      });
+    });
+  } else if (staticFiles) {
+    Object.keys(staticFiles).forEach(path => {
+      files.push(staticFiles[path]);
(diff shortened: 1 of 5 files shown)


packages/webapp/webapp_server.js
```diff
@@ -467,6 +471,63 @@ WebAppInternals.staticFilesMiddleware = function (staticFiles, req, res, next) {
   }
 };
 
+function getStaticFileInfo(originalPath, browser) {
+  const { arch, path } = getArchAndPath(originalPath, browser);
+
+  if (! hasOwn.call(WebApp.clientPrograms, arch)) {
+    return null;
```
benjamn: This check may be redundant with the one below.

CaptainN:
This looks reasonable. We should make sure to document the difference in deliverables, so that front end caches can make sure to set 'vary-by' to for useragent request headers. An alternative idea - would it make sense to add a legacy argument to the query string?

How many assets actually require a legacy URL? I'm under the impression that only built assets would need the legacy modifier - the main JS bundles (maybe worker bundles if we add those at some point) and maybe the CSS file, if there's benefit to stripping out auto-prefixer overhead (maybe not). What else might need legacy URLs?

packages/non-core/bundle-visualizer/server.js
```diff
@@ -33,21 +33,38 @@ function getStatBundles() {
...
+  }
+
+  const {
+    staticFiles,
+    staticFilesByArch,
+  } = WebAppInternals;
+
+  const files = [];
+
+  if (staticFilesByArch) {
```
hwillson: Total drive-by comment, but you might be able to `DRY` this up a bit with something like:

```js
if (staticFilesByArch || staticFiles) {
  const sourceFiles = staticFilesByArch || { '': staticFiles };
  Object.keys(sourceFiles).forEach(arch => {
    const staticFiles = sourceFiles[arch];
    Object.keys(staticFiles).forEach(path => files.push(staticFiles[path]));
  });
}
```

Issue #7675, opened 2016-08-19
[1.4.1] Error: bad path


wilau2:
Upgrading from 1.4.0.1 to meteor 1.4.1.
running on osx
I have a local package: sambegin:push
When i start meteor i have a problem that seems to be related with my local packages. 
It creates an other folder in my packages/ names sambegin_push

In my .meteor/local/plugin-cache/sambegin_push/local seems to be empty. 

The error is the following : 

`/Users/williamlauze/.meteor/packages/ecmascript/.0.5.8.157vaws++os+web.browser+web.cordova/plugin.compile-ecmascript.os/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:165
      throw error;
      ^

Error: bad path: ../sambegin_push/.npm/plugin/configuration/node_modules
    at rejectBadPath (/tools/isobuild/bundler.js:185:11)
    at NodeModulesDirectory.getPreferredBundlePath (/tools/isobuild/bundler.js:273:5)
    at /tools/isobuild/bundler.js:1740:39
    at Function._.each._.forEach (/Users/williamlauze/.meteor/packages/meteor-tool/.1.4.1.w7wcti++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/underscore/underscore.js:87:22)
    at JsImage.write (/tools/isobuild/bundler.js:1734:7)
    at /tools/isobuild/isopack.js:1474:36
    at Function._.each._.forEach (/Users/williamlauze/.meteor/packages/meteor-tool/.1.4.1.w7wcti++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/underscore/underscore.js:87:22)
    at /tools/isobuild/isopack.js:1468:11
    at Function._.each._.forEach (/Users/williamlauze/.meteor/packages/meteor-tool/.1.4.1.w7wcti++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/underscore/underscore.js:87:22)
    at /tools/isobuild/isopack.js:1467:9
    at Isopack.saveToPath (/tools/isobuild/isopack.js:1191:9)
    at /tools/isobuild/isopack-cache.js:378:23
    at /tools/utils/buildmessage.js:359:18
    at [object Object].withValue (/tools/utils/fiber-helpers.js:89:14)
[...]

PR #9857, opened 2018-05-04
Removing colons conversion

mitar:
This is probably not a correct fix, but it fixes things for me on Linux for https://github.com/meteor/meteor/issues/7675.

### DIFF
--- a/tools/isobuild/meteor-npm.js
+++ b/tools/isobuild/meteor-npm.js
@@ -69,7 +69,7 @@ meteorNpm.updateDependencies = function (packageName,
   // randomize the name, in case we're bundling this package
   // multiple times in parallel.
   var newPackageNpmDir =
-    convertColonsInPath(packageNpmDir) + '-new-' + utils.randomToken();
+    packageNpmDir + '-new-' + utils.randomToken();
 
   if (! npmDependencies || _.isEmpty(npmDependencies)) {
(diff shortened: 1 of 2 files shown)


mitar:
Or maybe in fact this is the correct fix. If paths on the file system can contain `:`, so existing paths contain them, then we should just keep them around.

BTW, how do people put packages into `packages` on Windows, if they contain `:`?

KoenLav:
@mitar the name of the directory in packages does not define the name Meteor will recognize (rather the name listed in the package.js).

I do however agree that Meteor should be able to handle this case, as I expect many more expect the directory name to be related.

mitar:
You are saying that I can put `foo:bar` package into `packages/something` and Meteor will detect this and use this `foo:bar` package when I do `meteor add foo:bar`? I have not seen this being the case.

copleykj:
Correct

KoenLav:
Exactly, nevertheless I think Meteor should be able to handle colons in directory names (if this is supported on various platforms).

benjamn:
The way we handle colons currently is important for Windows support, and not something we are likely to change without extensive tests and validation from Windows developers.

mitar:
Sure, but currently #7675 does not work on Linux. :-) So the question is, what to do there?

PR #9876, opened 2018-05-10
[WIP] Implement batching all the way from MongoDB => DDP

KoenLav:
This PR implements batching from the level of the multiplexer up until the DDP server.

More info to follow,


packages/ddp-server/session.js
```diff
@@ -204,12 +224,47 @@ export default class Session {
     return ret;
   }
 
+  messages(subscriptionHandle, collectionName, messages) {
```
KoenLav: This appears to be the only place where we need to unwind the buffered messages and apply some logic based on the actions.

KoenLav:
When I started working on this PR I thought the Multiplexer was the closest I could get to the database in Meteor... Boy did I turn out to be wrong; I still had to work my way through the ObserveDriver, which made use of OplogTailing and something called Crossbar (seems to be a non-descriptive name for what the class actually does).

Anyhow, I think the 'Golden Grail' of server performance for Meteor would be to buffer oplog notifications, per collection, immediately upon arrival in Meteor framework code and send this buffer through the classes which are, eventually, all partially responsible for the proper working of DDP (OplogTailing > Crossbar> MongoDriver > Multiplexer > (LocalCollection) > ObserveHandle > ObserveChanges > Collection > Subscription > Session (incl. SessionCollectionView & SessionDocumentView) > DDP).

However, we should probably also take into account that some queries make use of Polling rather than Tailing. So while implementing buffering in the OplogTailing makes sense, it also makes sense to buffer changes at the Multiplexer level (what this PR does).

I'm not sure whether this is even possible, but I guess it would be preferable if OplogTailing didn't unwind all its changes as singular messages, transferred them to the Multiplexer, only to have the Multiplexer buffer the changes again, but I'm not sure how the OplogTailing class should be refactored to accommodate this.

All in all, when reading into and reviewing all of this code, I think there is quite some duplication in the Meteor core and some Classes do not have a neat separation of concerns. Even though I think I managed to resolve the most glaring performance issues people have been experiencing I feel a review of the classes mentioned above, and how they interact, would be a good idea ;)

KoenLav:
I just ran some tests combining this PR with the buffering in OplogTailing and the performance for emptying a subscription with 20.000 documents is increased about 100x. Filling a subscription which was previously emptied is about 5x faster.

There is however one nasty bug where filling the subscription up again will only result in 10.000 documents being recognized on the server (rather than 20.000).

smeijer:
Just wanted to say thanks, from one user to another. I hope this will land some day soon.

PR #9883, opened 2018-05-12
polyfill es7 array functions

sakulstra:
with the split of modern/legacy bundles we no longer polyfill already present standard features. While array.prototype.includes is available in modern browsers flatten and flatMap aren't. This pr readds the polyfills to the modern bundle.

------

If I got this right: https://github.com/meteor/meteor/blob/master/CONTRIBUTING.md#submitting-pull-requests I should base of devel, but this seems kind of impossible as devel doesn't yet include the [modern/legacy](https://github.com/meteor/meteor/tree/devel/packages/ecmascript-runtime-client) split. Just ping me if I should rebase or sth.

### DIFF
--- a/packages/ecmascript-runtime-client/modern.js
+++ b/packages/ecmascript-runtime-client/modern.js
@@ -2,6 +2,8 @@ require("core-js/modules/es6.object.is");
 require("core-js/modules/es6.function.name");
 require("core-js/modules/es6.number.is-finite");
 require("core-js/modules/es6.number.is-nan");
+require("core-js/modules/es7.array.flatten");
+require("core-js/modules/es7.array.flat-map");
 require("core-js/modules/es7.object.values");
 require("core-js/modules/es7.object.entries");
 require("core-js/modules/es7.object.get-own-property-descriptors");


packages/less/.npm/plugin/compileLessBatch/npm-shrinkwrap.json
```diff
@@ -214,7 +214,7 @@
     },
     "less": {
       "version": "https://github.com/meteor/less.js/tarball/8130849eb3d7f0ecf0ca8d0af7c4207b0442e3f6",
-      "integrity": "sha1-dD95WEi4QkZPVjcydtAMY6WS/qM="
+      "integrity": "sha512-514pMpzrABW11ploVaRcbHLf3MhollYYCJj9SyJmNT1O8ZeLCiIeqZ2SjZUM1uudohPlHgTRwmCZouzIJFJ7Jw=="
```
sakulstra: not sure where this comes from, will remove it

benjamn: Yep, please revert this for now. There's been a lot of movement lately in terms of how `npm` generates `package-lock.json` and `npm-shrinkwrap.json` files, so this is probably just an artifact of not having regenerated this `npm-shrinkwrap.json` file since the last time `npm` was updated.

benjamn:
You're correct, targeting `release-1.7` (not `devel`) was the right thing to do for this PR.

benjamn:
Also, that test failure was already present on `release-1.7`, but should be fixed if you rebase now.

sakulstra:
@benjamn alright, rebased on latest 1.7 and removed the change in shrinkwrap :+1:

PR #9885, opened 2018-05-12
Buffer oplog notifications & fix Fiber spikes

KoenLav:
I started out working on this PR with the intention of buffering the oplog notifications early on (right after they are received the code), but later found out this also fixes the Fiber spikes some of us have been experiencing in production.

These spikes were caused by updating a significant amount of documents in a Collection to which some user was Subscribed in a way where the Oplog tail would need to re-fetch the documents (rather than relying on the modifications from the oplog itself). The ids of these documents were added to a queue, which Meteor would loop over, calling fetch (which started a Fiber) for every document.

Fixes this issue: https://github.com/KoenLav/meteor-fiber-repro


nathan-muir:
As an alternative point of view (from information gathered [for this other comment](https://github.com/meteor/meteor/pull/9862#issuecomment-388711651))

I'm not sure that it's necessary to batch oplog notifications in this way, as there's a natural batching mechanism due to the event loop, and use of `Meteor.defer`

 - in `OplogHandle` oplog [entries are pushed onto the `_entryQueue` continuously until the cursor yields](https://github.com/meteor/meteor/blob/devel/packages/mongo/oplog_tailing.js#L239-L243)
 - Then, in `OplogHandle#_maybeStartWorker`, the [`_entryQueue` is then consumed until exhaustion](https://github.com/meteor/meteor/blob/devel/packages/mongo/oplog_tailing.js#L255). 
   - Notifications are sent to [`OplogDriver` via crossbar](https://github.com/meteor/meteor/blob/devel/packages/mongo/oplog_tailing.js#L304), where the [`OplogDriver` enqueues docs in `_needToFetch`](https://github.com/meteor/meteor/blob/devel/packages/mongo/oplog_observe_driver.js#L643-L645)
 - [`OplogDriver#_fetchModifiedDocuments` consumes `_needToFetch`](https://github.com/meteor/meteor/blob/devel/packages/mongo/oplog_observe_driver.js#L481) by spawning a fiber per document to fetch (source of fiber spikes)

This means, if 1000 entries arrive on the Oplog cursor at once, they are all pushed through to the observer drivers before fetching begins.

---

I think we could get the same net effect (as this PR), by simply picking a better heuristic within `_fetchModifiedDocuments` for reverting to polling, or improving patterns for fetching documents by id.

KoenLav:
After some discussion with @veered we found the following:

The  _maybeStartWorker function in the OplogTailing class checks whether oplog entry processing isn't too far behind (if (self._entryQueue.length > 2000)), but it appears that (even when updating 20000 (x10) documents at once, this never happens).

We could try finding out why, but even if we do 'fix' this falling behing on the OplogTailing would cause ALL subscriptions (and other server-side observers which rely on polling) to fall back to polling. Even when there could be 20 collections, of which only ONE is receiving a lot of updates.

Hence I concluded that there is some merit in batching the DDP messages within the crossbar and moving (or also adding) the code which tells the OplogObserverDriver to fallback to polling in there (checking whether a particular collection is receiving more than x notifications within a short timeframe).

This PR will require a bit of rework to accomplish the scenario as described above, but I do think this is the way to go.

@veered thinks it might make more sense to move this logic (queue + fallback) further down in the process (where the observers are constructed) as not every change to a document in a collection is relevant for every observer on that collection.

KoenLav:
@natha-muir it seems that the entryQueue will NEVER reach the number of 2000 (I assume because at some point the process of 'processing the notifications from the queue' takes up just as much processing time as 'filling the queue'.

Could you maybe shed some light on this (and/or) reply to my post above?

mitar:
Always merge PRs against devel branch.

StorytellerCZ:
@KoenLav please update the PR as @mitar suggested.
Any news on this @benjamn ?

PR #9898, opened 2018-05-23
HTTP.call docs typo

jamesmillerburgess:
Found this while reading up on HTTP.


benjamn:
By the way, once Meteor 1.7 is released, I would really love to implement an isomorphic [`fetch()`](https://fetch.spec.whatwg.org/) polyfill package that uses three different implementations in modern browsers (no polyfill), legacy browsers ([`whatwg-fetch`](https://www.npmjs.com/package/whatwg-fetch)), and Node ([`node-fetch`](https://www.npmjs.com/package/node-fetch)). Ideally this `fetch` package could replace the `http` package.

PR #9917, opened 2018-05-29
Release 1.7.0.1

benjamn:
This is a very small follow-up release to Meteor 1.7 to revert the optimization attempted in https://github.com/meteor/meteor/pull/9825. Since Meteor 1.7 has not yet been "recommended," hopefully everyone will update directly to 1.7.0.1 and thus will not encounter this problem.

This reverts commit c5302bb2ba609e5535dcce8647adf30e0a6f8daf.
    
Based on conversation with @KoenLav, it seems the optimization attempted by #9825 was more of a breaking change than anticipated, and the workaround of creating a symbolic link to induce compilation is not just annoying on Windows but also does not satisfy the goal of being able to import from `node_modules` (as before) rather than importing via the symbolic link.
    
We will need to revisit this pitfall in Meteor 1.7.1, but it's important to unbreak it ASAP in Meteor 1.7.0.1 (since 1.7 is now final).
    
https://github.com/meteor/meteor/pull/9826#issuecomment-392596129
https://github.com/mozfet/meteor-autoform-materialize/issues/43
    
For apps that are accidentally compiling too many LESS or SCSS files, the new recommended workaround (for now) is to add the offending directories and/or files to a `.meteorignore` file.


packages/ecmascript-runtime-client/versions.js
```diff
@@ -9,7 +9,7 @@ setMinimumBrowserVersions({
   // (pre-Edge) from the modern classification. #9818 #9839
   ie: 12,
   firefox: 45,
-  mobile_safari: 10,
+  mobileSafari: 10,
```
KoenLav: Nice!

benjamn: The funny/sad thing is that Meteor is actually responsible for the camel-casing [here](https://github.com/meteor/meteor/blob/4e59d8bb6514c718960b69a46abda591b3390c19/packages/webapp/webapp_server.js#L116). 🤦‍♂️

benjamn:
@KoenLav Can you test the RC real quick?
```sh
meteor update --release 1.7.0.1-rc.0
```

KoenLav:
@benjamn everything seems to be working now!

p3pp8:
Hello, i've just upgraded to Meteor 1.7.0.1 and i've to downgrade as my app is throwing this error:

`WebSocket connection to 'wss://0.0.0.0/websocket' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED
Uncaught Error: Cannot find module '/node_modules/meteor/socket-stream-client/sockjs-0.3.4.js'`

And everything is KO. 
Note: i can't support websocket on production environment so i use DISABLE_WEBSOCKETS=1 on env variable.

Any help is really, really appreciated!

benjamn:
@p3pp8 There's no easy fix for that right now, but bear with me, because I think I can suggest a pretty good workaround:

1. First, copy the [`socket-stream-client`](https://github.com/meteor/meteor/tree/devel/packages/socket-stream-client) package from a checkout of this repository into your local `packages/` directory.
2. Next, open [`socket-stream-client/browser.js`](https://github.com/meteor/meteor/tree/devel/packages/socket-stream-client/browser.js) in your text editor.
3. Add a static `import "./sockjs-0.3.4.js"` line to the top of that file, so that SockJS will be bundled in the initial JS bundle, and `WebSocket`s will not be used.
4. Run/test/deploy your app as usual.

The only downside to this is that you need to keep your local copy of `socket-stream-client` up-to-date, but future versions of Meteor will complain if the package version fails to satisfy constraints.

p3pp8:
@benjamn Ok, thank you! I'm going to try it out. So will it be solved in future versions or do i have to keep in my project packages folder the socket-stream-client patched like you suggested? 
Thanx!!!

holmrenser:
@benjamn I am curious as to what the design choice behind this is, as it seems like this is what is breaking a lot of apps at the moment? I have reported the same error as @p3pp8 in #9928, are your saying I should get my own sockjs to fix this?

TVXtrem:
Do you mean the packages directory in AppData/Local/.meteor/ ? Because copying the package into this folder and editing the browser.js file afterwards does not work for me unfortunately.

benjamn:
I've just released a new version of `socket-stream-client` that goes back to using `SockJS` in the modern bundle (just like the legacy bundle) rather than a native `WebSocket`. If you run `meteor update socket-stream-client` to get version 0.2.2, I believe these problems will go away, since `SockJS` has all sorts of clever tricks for coping with unusual networking environments that can cause problems for plain `WebSocket`s.

p3pp8:
@TVXtrem If you want to bundle your version of a package you have to create a Packages folder in the root directory of your project, than you put your custom package inside it. @benjamn Thanx, now it's safer than keeping the local socket-stream-client package up to date!!!

PR #9933, opened 2018-06-03
Add Subresource Integrity for JS and CSS files

morloy:
After moving our Meteor assets out to a CDN, the [Mozilla Observatory](https://observatory.mozilla.org/) kept complaining about missing [Subresouce Integrity](https://infosec.mozilla.org/guidelines/web_security#subresource-integrity) in our JavaScript files.
This PR adds a `sha512` hash to all basic Meteor files served to the client.

Solves meteor/meteor-feature-requests#35.

### DIFF
--- a/packages/boilerplate-generator-tests/test-lib.js
+++ b/packages/boilerplate-generator-tests/test-lib.js
@@ -10,8 +10,7 @@ export async function generateHTMLForArch(arch, includeHead) {
       cacheable: true,
       url: '/packages/bootstrap/css/bootstrap-responsive.css?hash=785760fc5ad665d7b54d56a3c2522797bb2cc150&v="1"',
       size: 22111,
-      hash: '785760fc5ad665d7b54d56a3c2522797bb2cc150',
-      sha512: 'FgwCpUKZj/5DORw0OQWACh65bi8wxnLGcoHxNq9bbErR9TLacgBisdqhg4uBLZ2sctoiSjG+hWLRigs1mBZY/Q==',
+      hash: '785760fc5ad665d7b54d56a3c2522797bb2cc150'
     },
     {
       path: 'packages/templating-runtime.js',
@@ -20,8 +19,7 @@ export async function generateHTMLForArch(arch, includeHead) {
       cacheable: true,
       url: '/packages/templating-runtime.js?hash=c18de19afda6e9f0db7faf3d4382a4c953cabe18&v="1"',
       size: 24132,
-      hash: 'c18de19afda6e9f0db7faf3d4382a4c953cabe18',
-      sha512: 'mJ9OEOi4HbVMzqiVGOjxCiIIImgnzEg6QhorPno46yULMu/3tKmENYriasWP9RrM+wLNwlTzw6+9d+nWY9/A0A==',
+      hash: 'c18de19afda6e9f0db7faf3d4382a4c953cabe18'
     },
   ];
 
--- a/packages/boilerplate-generator-tests/test-lib.js
+++ b/packages/boilerplate-generator-tests/test-lib.js
@@ -10,7 +10,8 @@ export async function generateHTMLForArch(arch, includeHead) {
       cacheable: true,
       url: '/packages/bootstrap/css/bootstrap-responsive.css?hash=785760fc5ad665d7b54d56a3c2522797bb2cc150&v="1"',
       size: 22111,
-      hash: '785760fc5ad665d7b54d56a3c2522797bb2cc150'
+      hash: '785760fc5ad665d7b54d56a3c2522797bb2cc150',
+      sha512: 'FgwCpUKZj/5DORw0OQWACh65bi8wxnLGcoHxNq9bbErR9TLacgBisdqhg4uBLZ2sctoiSjG+hWLRigs1mBZY/Q==',
     },
     {
       path: 'packages/templating-runtime.js',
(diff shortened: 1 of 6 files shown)


packages/boilerplate-generator-tests/test-lib.js
```diff
@@ -19,7 +20,8 @@ export async function generateHTMLForArch(arch, includeHead) {
       cacheable: true,
       url: '/packages/templating-runtime.js?hash=c18de19afda6e9f0db7faf3d4382a4c953cabe18&v="1"',
       size: 24132,
-      hash: 'c18de19afda6e9f0db7faf3d4382a4c953cabe18'
+      hash: 'c18de19afda6e9f0db7faf3d4382a4c953cabe18',
```
abernix: I had been tempted (both today and [in the past](https://github.com/meteor/meteor/issues/8286#issuecomment-275488777)) to suggest that we switch to using a newer cipher for `hash` (which currently uses SHA-1, 40-bytes in length — in hex) and just use that for the SRI — relying on only a single hash! 🙌🤔

That temptation was originally based on me thinking we could go with SHA-256, which has a relatively short hexadecimal representation (64-bytes) and would still work fine as a filename (which `hash` is frequently used for in builds and when serving files to clients!).

Unfortunately, as hashes become more complex, they become longer and the hex representation of a SHA-512 hash would be 128-bytes — a super long filename.  (Maybe too long?)

Most problematically though, SRI requires `base64` encoding which, while shorter than 128-bytes for SHA-512, fails to be useful as a filename because of the base64 encoding (`/`, in particular), so my temptation seems to be well-prevented.

packages/boilerplate-generator/template-web.browser.js
```diff
@@ -9,8 +9,9 @@ export const headTemplate = ({
 }) => {
   var headSections = head.split(/<meteor-bundled-css[^<>]*>/, 2);
   var cssBundle = [...(css || []).map(file =>
-    template('  <link rel="stylesheet" type="text/css" class="__meteor-css__" href="<%- href %>">')({
+    template('  <link rel="stylesheet" type="text/css" class="__meteor-css__" href="<%- href %>" integrity="sha512-<%- hash %>" crossorigin="anonymous">')({
```
abernix: I'm curious if this `crossorigin="anonymous"` might potentially cause problems for existing deployments which might necessitate the presence of user credentials (as defined [here](https://www.w3.org/TR/cors/#user-credentials); e.g. HTTP Auth, Cookies, etc.) in order to serve the JS or CSS files (let's say the Meteor's deployment was fully behind that forced HTTP Authentication).

From what I understand, it's mandatory for `crossorigin` to be set for SRI to work (See [0](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), [1](https://www.npmjs.com/package/ember-cli-sri#crossorigin-attribute)), but I'm afraid this might need to be toggleable between [the valid options](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes), `anonymous` and `use-credentials`.

Any thoughts?

packages/boilerplate-generator/template-web.browser.js
```diff
@@ -54,8 +55,9 @@ export const closeTemplate = ({
   '',
 
   ...(js || []).map(file =>
-    template('  <script type="text/javascript" src="<%- src %>"></script>')({
+    template('  <script type="text/javascript" src="<%- src %>" integrity="sha512-<%- hash %>" crossorigin="anonymous"></script>')({
```
abernix: Same question as above!

packages/boilerplate-generator-tests/test-lib.js
morloy: I‘d recommend switching to SHA-512 and simply truncate it for the filenames. [This seems to be totally fine](https://crypto.stackexchange.com/questions/9435/is-truncating-a-sha512-hash-to-the-first-160-bits-as-secure-as-using-sha1). If more information needs to go into the filename, there’s Base58 or [UNMISTAKABLE_CHARS](https://github.com/meteor/meteor/blob/14c135668d197131fbaab36d962fcbee9f911056/packages/random/random.js#L88), that would give Base55.

Recoding the hash from hex to Base64 for SRI would be trivial.

packages/boilerplate-generator/template-web.browser.js
morloy: Having this configurable sounds good to me. But where would this config go ideally? Any ideas?

packages/boilerplate-generator/generator.js
```diff
@@ -124,6 +125,14 @@ export class Boilerplate {
         itemObj.scriptContent = readUtf8FileSync(
           pathMapper(item.path));
         itemObj.inline = true;
+      } else {
+        if (item.sha512) {
+          itemObj.hash = item.sha512;
+        } else {
+          const hash = createHash('sha512');
```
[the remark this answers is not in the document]
morloy: As explained below, using SHA-512 for the hash in the manifest and truncating it for filenames would be a good approach in my opinion.

morloy:
I added `WebAppInternals.enableSubresourceIntegrity(use_credentials = false)` to make the whole SRI optional and allow to select either `anonymous` or `use-credentials` for the *crossorigin*.
By default, SRI is disabled to make sure that we don’t break any existing setups. Hope this fine now.

PR #9954, opened 2018-06-06
Allow dynamic import() requests from any origin on any device.

### LINKED ISSUE #9952: dynamic-imports broken on electron due to unsupported CORS header
dynamic-imports allows remote origins to fetch the imports by settings CORS headers, but does so by setting the `Access-Control-Allow-Headers` header to `*`.
As referenced by https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#Compatibility_notes, this is not supported in all recent browsers. For my case specifically, it's not supported by the version of chromium included in the latest electron.

A simple fix (for which I will soon create a PR) would (in my opinion) be to bounce back the `Access-Control-Request-Headers` header value from the options request if it is provided.

call-a3:
A tweak to the change introduced in c4b5707747ce3c5bf339b62940da84628680a1d3 to fix #9952.
This will allow clients that don't support the * value in `Access-Control-Allow-Headers`,
but do specify the `Access-Control-Request-Headers` (such as electron 2.0.2) to use dynamic import.

### DIFF
--- a/packages/dynamic-import/server.js
+++ b/packages/dynamic-import/server.js
@@ -66,8 +66,9 @@ function middleware(request, response) {
   response.setHeader("Access-Control-Allow-Origin", "*");
 
   if (request.method === "OPTIONS") {
-    if (request.headers["access-control-request-headers"] !== undefined) {
-      response.setHeader("Access-Control-Allow-Headers", request.headers["access-control-request-headers"]);
+    const requestedHeaders = request.headers["access-control-request-headers"];
+    if (requestedHeaders !== undefined) {
+      response.setHeader("Access-Control-Allow-Headers", requestedHeaders);
     } else {
       response.setHeader("Access-Control-Allow-Headers", "*");
     }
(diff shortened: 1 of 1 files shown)


packages/dynamic-import/server.js
```diff
@@ -66,7 +66,11 @@ function middleware(request, response) {
   response.setHeader("Access-Control-Allow-Origin", "*");
 
   if (request.method === "OPTIONS") {
-    response.setHeader("Access-Control-Allow-Headers", "*");
+    if (request.headers["access-control-request-headers"] !== undefined) {
```
benjamn: Do we need to worry about different possible capitalizations when checking for this header?

Also we typically test `typeof value === "undefined"` rather than relying on the `undefined` identifier, just because `undefined` is mutable in some environments.

packages/dynamic-import/server.js
```diff
@@ -66,7 +66,11 @@ function middleware(request, response) {
...
+      response.setHeader("Access-Control-Allow-Headers", request.headers["access-control-request-headers"]);
```
benjamn: Especially if we end up having to check for multiple variations of this header, it would be good to store the value in a variable that we could reuse down here.

PR #10010, opened 2018-06-19
[R] Additional collection function

robecor:


benjamn:
You're going to need a lot more of an explanation to motivate this change. For example, where does this `NC` namespace come from, and how do all those fields make sense in a generic package like `accounts-base`?

PR #10017, opened 2018-06-21
Track number of documents in SessionCollectionView

koszta:
Fixes issue #9828

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -113,6 +113,7 @@ var SessionCollectionView = function (collectionName, sessionCallbacks) {
   var self = this;
   self.collectionName = collectionName;
   self.documents = {};
+  self.documentCount = 0;
   self.callbacks = sessionCallbacks;
 };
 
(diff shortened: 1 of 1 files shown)


veered:
@koszta You verified that this resolved your performance issues?

koszta:
@veered Yes, thanks for the hint!

In the reproduction repo there is a fix branch using the patched ddp-server:
https://github.com/fuww/blocked-unsub/tree/fix

Now it is not blocking the event loop for more than ~20-30ms while unsubscribing with 2000 published documents.

nathan-muir:
Awesome! Removing the call to `_.isEmpty` reduces the time to unsub from 5s to 500ms (for a production dataset)

Just a thought: instead of the counter, we could also use a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) which has a `.size` property & better suited to adding/removing items and iteration. (eg. https://github.com/nathan-muir/meteor/commit/d511e551df1c22d7687056a7fd8cb41a8a995e6b)

veered:
This definitely seems worth merging (perhaps after making the change suggested by @nathan-muir )!

benjamn:
I agree that using a `Map` would make this easier!

koszta:
I also agree, but there is `DiffSequence.diffObjects` called with `self.documents`, which I assume would not work with a `Map`

koszta:
Closing this in favor of #10053

PR #10025, opened 2018-06-24
Split bundles into separate sunbursts in bundle-visualizer

jamesmillerburgess:
Posting as WIP because I would like to get some feedback on the design. Please let me know what you think could be better!

The pills along the top are clickable and switch the sunburst between the two bundles, like so:

screen shot 2018-06-24 at 8 12 53 am 2

screen shot 2018-06-24 at 8 10 17 am 2

### DIFF
--- a/packages/non-core/bundle-visualizer/package.js
+++ b/packages/non-core/bundle-visualizer/package.js
@@ -1,5 +1,5 @@
 Package.describe({
-  version: '1.2.1',
+  version: '1.2.2',
   summary: 'Meteor bundle analysis and visualization.',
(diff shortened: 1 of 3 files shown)


jamesmillerburgess:
From the emoji poll it looks like there aren't any objections here, so I'm going to remove the WIP tag and this should be ready for review. 👍

benjamn:
This has been published (just now) as `bundle-visualizer@1.2.2`. Tested in multiple browsers, with both one and two bundles. Great work @jamesmillerburgess!

PR #10029, opened 2018-06-25
Provide isomorphic implementation of WHATWG fetch() API.

benjamn:
As described in my [talk](https://youtu.be/vpCotlPieIY?t=29m18s) about Meteor 1.7's modern/legacy bundling system (relevant [slides](https://slides.com/benjamn/meteor-night-may-2018#/46)), the [WHATWG `fetch()` API](https://fetch.spec.whatwg.org/) is a great example of a feature that requires a totally different implementation in modern browsers, legacy browsers, and Node.

The `fetch()` API is already widely used in the JavaScript community, and I suspect it will eventually replace the core Meteor `http` package for most HTTP request needs, especially since it is essentially free in modern browsers.

### DIFF
--- a/packages/fetch/fetch-tests.js
+++ b/packages/fetch/fetch-tests.js
@@ -1,0 +1,8 @@
+// Import Tinytest from the tinytest Meteor package.
+import { Tinytest } from "meteor/tinytest";
+
+// Write your tests here!
+// Here is an example.
+Tinytest.add('fetch - example', function (test) {
+  test.equal(typeof fetch, "function");
+});
--- a/packages/fetch/fetch-tests.js
+++ b/packages/fetch/fetch-tests.js
@@ -1,8 +1,0 @@
-// Import Tinytest from the tinytest Meteor package.
-import { Tinytest } from "meteor/tinytest";
-
-// Write your tests here!
-// Here is an example.
-Tinytest.add('fetch - example', function (test) {
-  test.equal(typeof fetch, "function");
-});
--- a/packages/appcache/appcache_tests-client.js
+++ b/packages/appcache/appcache_tests-client.js
@@ -1,28 +1,28 @@
 const manifestUrl = '/app.manifest';
 
-const appcacheTest = (name, cb) => {
-  Tinytest.addAsync(`appcache - ${name}`, (test, next) => {
-    HTTP.get(manifestUrl, (err, res) => {
-      err ? test.fail(err) : cb(test, res);
-      next();
-    });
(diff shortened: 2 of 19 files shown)


packages/fetch/fetch-tests.js
```diff
@@ -0,0 +1,8 @@
+// Import Tinytest from the tinytest Meteor package.
+import { Tinytest } from "meteor/tinytest";
+
+// Write your tests here!
+// Here is an example.
+Tinytest.add('fetch - example', function (test) {
+  test.equal(typeof fetch, "function");
```
benjamn: Need more tests, obviously!

benjamn:
If anyone else wants to help replace `HTTP` with `fetch` in Meteor core packages, just comment here first, and then open a PR targeting the `isomorphic-fetch` branch.

packages/fetch/fetch-tests.js
hwillson: No way - I like how manageable these tests are ... 🙂.

sebakerckhof:
@benjamn what would the strategy be for replacing server-side calls that are now synchronous because of fibers. Just Promise.await the fetch?

And what about error handling? HTTP package will throw if the status code is >= 400 (see https://github.com/meteor/meteor/blob/devel/packages/http/httpcall_server.js#L119 and https://github.com/meteor/meteor/blob/devel/packages/http/httpcall_client.js#L180), fetch will not do this.
So I don't think the places where you already replaced it have the same behavior as before.

sebakerckhof:
My questions were a reply to https://github.com/meteor/meteor/pull/10029#issuecomment-400105344. So in the context of replacing HTTP in the current core packages, are those also the recommendations? Because awaiting in an async instead of relying on fibers would be a backwards incompatible change for end-users, which Meteor usually tries to prevent.

Same thing for the errors. I agree that fetch shouldn't be a drop-in replacement for HTTP. Rather the opposite and it should just follow the fetch-spec. But handling errors differently might brake existing core packages. E.g. when they rely on a 404 throwing an error or something.

jamesmillerburgess:
I will replace HTTP in `bundle-visualizer` since I'm already touching it in another PR.

benjamn:
In light of @sebakerckhof's astute feedback, and in order to make this package available for use with the `bundle-visualizer` (and other packages where it makes sense), I think we should try to publish the initial version of `fetch` soon (maybe after adding more tests), without fully migrating away from (or deprecating) `http` just yet.

Let's make `fetch()` happen!

Issue #10044, opened 2018-07-02
Problems with standard-minifier-js


musemind:
Building with `standard-minifier-js@2.3.4` runs in error.
Meteor 1.7.0.3

```
Errors prevented bundling:                    
While minifying app code:                     
/Users/jl/.meteor/packages/standard-minifier-js/.2.3.4.b8cycq.lb73u++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:779:15: Unexpected token (1:248893)
at _class.raise (/Users/jl/.meteor/packages/standard-minifier-js/.2.3.4.b8cycq.lb73u++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:779:15)
at _class.unexpected (/Users/jl/.meteor/packages/standard-minifier-js/.2.3.4.b8cycq.lb73u++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/met

[... MUCH MORE SIMILAR ERRORS ...]

at Object.parse (/Users/jl/.meteor/packages/standard-minifier-js/.2.3.4.b8cycq.lb73u++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/reify/lib/parsers/babylon.js:40:18)
at Object.parse (packages/babel-compiler.js:46:29)
at extractModuleSizesTree (packages/minifyStdJS/plugin/stats.js:23:23)
at files.forEach.file (packages/minifyStdJS/plugin/minify-js.js:141:20)
at Array.forEach (<anonymous>)
at MeteorBabelMinifier.processFilesForBundle (packages/minifyStdJS/plugin/minify-js.js:118:9)
```

The problem is caused by a specific file `/cordova.js` in our project.
Uncomment the `standard-minifier-js` before building solves the problem :)

WORKS, NO ERROR
```
// cordova.js
import {Meteor} from 'meteor/meteor'
import { FlowRouter } from 'meteor/kadira:flow-router'

if (Meteor.isCordova) {
  const subscribe = () => {
    console.log('subscribe to universal links')
    if (typeof universalLinks === 'undefined') return console.log('universalLinks not defined') // eslint-disable-line
    universalLinks.subscribe(null, function (eventData) { // eslint-disable-line
[...]

Issue #10042, opened 2018-07-01
Consider replacing uglify-es dependency in minifier-js package


arggh:
I'm not too familiar with the drama involved, but apparently some folks are changing `uglify-es` to either [terser](https://github.com/fabiosantoscode/terser) or [babel-minify](https://github.com/babel/minify).

The claims on `uglify-es` state that it's unmaintained among a few other bad things, and that it contains some serious bugs, known to create tricky situations especially with React.

[Parcel switching to terser](https://github.com/parcel-bundler/parcel/pull/1382)
[Ledger switching away from uglify](https://github.com/LedgerHQ/ledger-live-desktop/pull/782)
[Twitter thread about this
](https://twitter.com/greweb/status/1013122437884542978)

Like I said, I don't have any in-depth knowledge about the situation and this is just a heads-up about something I noticed by chance.

benjamn:
Done! To try out the new minifier, run `meteor update --release 1.7.1-beta.14` or create a new app with `meteor create --release 1.7.1-beta.14 your-new-app`.

arggh:
Well that was fast! 👍

abernix:
I hadn't caught wind of this story change! Thanks for catching this and the suggestion, @arggh!  (And @benjamn for making the change!)

Issue #10034, opened 2018-06-27
standard-minifier-js Safari syntax error with async/await class function.


Colkadome:
Hello,

I had an error with my minified javascript on my production app, and it resulted in a blank white screen on Safari for Mac. On mobile Safari, the app works as normal.

The error happens when a `return` statement is underneath an `await` call inside a `try catch` block, inside an `async` class method. I have tried some other code variations, but have only been able to reproduce it with the circumstances above.

Apologies if this is the wrong repo to report this issue, as I understand most of the minification is done with UglifyJS2?

## Reproduction

- Create minimal Meteor app with `meteor --minimal myApp`.
- Add the code below to `client/main.js`.
- Run `meteor --production` to cause minification.

```
function promiseReturningFunc () {
  return new Promise((resolve, reject) => {
    setTimeout(resolve, 1000, 'foo');
  });
}

class myClass {
  
  async myFunc() {

    try {

      await promiseReturningFunc();
      return;  // Removing this line this will remove the error.

    } catch (_) { }

  }
}
```

## The error

The following error is thrown in Safari:

`SyntaxError: Unexpected identifier 'e'. Expected a ';' following a return statement.`

The following minified code is produced, and is causing the error:

```
var require = meteorInstall({
        client: {
            "main.js": function() {
                function e() {
                    return new Promise((e, n) => {setTimeout(e, 1e3, "foo")})
                }
                class n {
                    async myFunc()
                    {
                        try {
                            return void await e()
                        } catch (e) {}
                    }
                }
            }
        }
    }, {
        extensions: [".js", ".json", ".html", ".css"]
    }),
    exports = require("/client/main.js");
```

## Specs

- Safari Version 11.0.2 (12604.4.7.1.4)
- macOS Sierra 10.12.6
- Meteor 1.7.0.3

Meteor package versions:
```
meteor@1.9.0
static-html
standard-minifier-css@1.4.1
standard-minifier-js@2.3.4
es5-shim@4.8.0
ecmascript@0.11.1
shell-server@0.3.1
webapp@1.6.0
server-render@0.3.1
```

Colkadome:
I tried the new minifier (#10045) and still no luck.

`meteor create --minimal --release 1.7.1-beta.14 myApp`

I'm sure this issue is very rare for developers, but it might be worth leaving up in case somebody else experiences this. When I have time, I'll do more digging to see if its a terser issue.

kzc:
[...]

Issue #9976, opened 2018-06-11
[1.7.0.1] Error when minifying code (or running with --production) 


tcastelli:
We have found a bug when trying to migrate one 1.6 app to 1.7 during the minification process. The app is still using the old meteor app structure (with some imports to npm modules but basically relying on folder names to get the order) and it runs locally without a a problem when not minified. We have found that it also works minified when switching standard-minifier-js to abernix:minifier-js

The stacktrace is not very friendly, but just in case here it is:

````
=> Errors prevented startup:
   
   While minifying app code:

   /home/toni/.meteor/packages/standard-minifier-js/.2.3.4.woj8rj.co1l++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:779:15:
   Unexpected token (1:259539)
   at _class.raise
   (/home/toni/.meteor/packages/standard-minifier-js/.2.3.4.woj8rj.co1l++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:779:15)
   at _class.unexpected
   (/home/toni/.meteor/packages/standard-minifier-js/.2.3.4.woj8rj.co1l++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:2082:16)
   at _class.parseStatementContent
   (/home/toni/.meteor/packages/standard-minifier-js/.2.3.4.woj8rj.co1l++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:4010:32)
   at _class.parseStatement
   (/home/toni/.meteor/packages/standard-minifier-js/.2.3.4.woj8rj.co1l++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:3963:17)
   at _class.parseStatement
   (/home/toni/.meteor/packages/standard-minifier-js/.2.3.4.woj8rj.co1l++os+web.browser+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/babylon/lib/index.js:6637:53)
   at _class.parseIfStatement
[...]

PR #10045, opened 2018-07-02
Replace uglify-es with terser, a more actively maintained fork.

benjamn:
Thanks to @arrgh for suggesting this replacement: #10042 

This may help with lots of issues, most recently #10044 but also #10034, #9976, #9711, and #9568, to name a few. In particular, since we silently fall back to `Babel.minify` when `uglify-es` (now `terser`) fails, there should be fewer problems with `Babel.minify` (parse errors, performance, memory usage) if `terser` succeeds more often.

Note: because we have bumped the minor versions of `minifier-js` and `standard-minifier-js`, and they are core packages, you'll have to be using the next beta version of Meteor 1.7.1 in order to update to these versions, unless you clone those packages into your local `packages/` directory. However, if we can validate that `terser` is fully backwards compatible with `uglify-es`, we might be able to back-port this fix as a patch update, or publish a Meteor 1.7.0.4 release that permits `minifier-js@2.4.x`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,6 +5,11 @@
   defined by which compiler plugins you have enabled.
   [PR #10027](https://github.com/meteor/meteor/pull/10027)
 
+* The `uglify-es` npm package used by `minifier-js` has been replaced with
+  [`terser@3.7.6`](https://www.npmjs.com/package/terser), a fork of
+  `uglify-es` that appears to be (more actively) maintained.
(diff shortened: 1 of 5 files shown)


arggh:
@benjamn it's @arggh though, not @arrgh :)

arrgh:
Yes it is, it wasn't me :)

PR #9622, opened 2018-02-01
[WIP] Modernize ddp-server

abernix:
As I value the hard work of our community contributors, and hope to save them from doing work they don't need to do, I'm submitting this WIP PR for a refactor/restructure of `ddp-server` code I began in early December as part of a hack-week project.

I'm actually aiming to pick up the hack week project soon, which I have locally and will lean on this re-factor, but I wanted to get this in before someone else embarks on it and we double up on work.

I'll tidy (_ahem_, finish) this up soon, but wanted to hop on the refactor train while it was still trending. 🚅🔧📈

PR #10053, opened 2018-07-04
ddp-server: replace usage of Object with Map & Set

nathan-muir:
@benjamn  The same spirit as #10017, but replaces all `Object` usage in ddp-server.

I've tried to minimise the complexity of the changes.

Alternatively, I can only change `SessionCollectionView#documents` to a `Map`, but it seems like a missed opportunity.

---

nb: there's also a small change to `mongo-id` as it started showing up in cpu-profiles due to use of [`startsWith`](https://jsperf.com/charat-vs-startswith/2)

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -14,8 +14,8 @@ var Fiber = Npm.require('fibers');
 // Represents a single document in a SessionCollectionView
 var SessionDocumentView = function () {
   var self = this;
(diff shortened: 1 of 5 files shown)


StorytellerCZ:
@abernix #9622

smohantyCME:
How can I find out what production version of Meteor this actually got released? I couldn't find this specific change mentioned in changelog of Meteor https://docs.meteor.com/changelog.html#v180120181123. Thanks

Issue #9948, opened 2018-06-05
[1.7.0.1] Optionally disable web.browser/web.browser.legacy in development


veered:
The overhead of compiling an additional client bundle on every code reload is quite large for big projects. It would be very helpful if it were possible to configure which client bundles are built.

The most likely case for development is wanting to disable `web.browser.legacy`. If the minimum browser requirements set by `modern-browsers` were set to require a modern browser, perhaps the `web.browser.legacy` bundle could just use the `web.browser` bundle. They would be identical anyways.

benjamn:
That's not quite how the minimum browser requirements work. Additional calls to `setMinimumBrowserVersions` can only restrict the set of browsers/versions that are considered modern, which enlarges the set of legacy browsers (i.e., all the ones that aren't modern). In other words, there's no way to relax the requirements so that every browser will be considered modern. That wouldn't be safe anyways, since not every browser is modern by any reasonable definition.

If rebuild times were quicker, the complexity of somehow disabling the `web.browser.legacy` build (and then at some point reenabling it… but when?) would matter a lot less. I would rather use the motivation behind this issue to improve rebuild times generally. Skipping an important part of the build process feels like cheating, to me.

In order to move forward with this feature request, I think we need
- [ ] a mechanism for skipping `web.browser.legacy` builds that doesn't become permanent because the developer forgets to turn it off
- [ ] some mechanism for reporting legacy build errors reliably even though we aren't building the `web.browser.legacy` bundle as often
- [ ] to decide what happens when/if the developer visits localhost:3000 using a legacy browser

benjamn:
To be clear, I'm sympathetic to the motivation here, I just haven't been able to come up with a developer workflow that feels right / gets all the details right. Zero-configuration may not be possible for something like this, so we have to be really thoughtful about how this behavior is configured (and reconfigured over time).

spencern:
This was a use-case I thought of immediately when watching your talk on 1.7 @benjamn. I understand your concerns about reporting legacy build errors when we're not readily building on that browser. 
I'd be super happy to have a flag of some kind that would permit skipping `web.browser.legacy` builds in development even if legacy build errors were not reported until `web.browser.legacy` builds were re-enabled.

As for what should happen when a developer visits localhost:3000 using a legacy browser after setting a flag to not include them in the build? If the site renders poorly or not-at all, that's good enough for me. Even better would be to log an error to the console reminding the developer they have disabled that build.

Floriferous:
[...]

PR #10055, opened 2018-07-05
Delay building web.browser.legacy bundle until after development server restarts.

benjamn:
Now that Meteor [builds two client bundles](https://blog.meteor.com/meteor-1-7-and-the-evergreen-dream-a8c1270b0901), build times have naturally increased a bit. We are hard at work attacking this problem from multiple angles, such as #9983, and I'm confident that Meteor 1.7.1 build times will be shorter for most apps than they were before Meteor 1.7, despite building both `web.browser` and `web.browser.legacy` bundles.

Nevertheless, since most testing in development happens using one bundle/browser at a time, it has been suggested (#9948) that we could skip the legacy build in development, as a cheap short-cut to faster rebuild times. To be honest, I don't love most of these ideas for disabling the legacy build with some sort of additional configuration, because supporting older browsers is important, configuration is a creeping cancer, and ignoring errors in the legacy build seems like a recipe for frustration when it's time to deploy your app.

I would vastly prefer a zero-configuration solution to this problem, though that's challenging in this case because skipping an important part of the build process seems like a decision the developer should make consciously and carefully.

As this pull request demonstrates, it's possible to remove the legacy build from the critical path to restarting the development server, so you can continue testing and developing your app while the legacy bundle is built in the background. There's nothing to configure, because the legacy bundle still gets built in roughly the same amount of time as before; you just don't have to wait for it to finish, if you aren't currently interested in testing legacy code.

How it works in detail:
* All bundles are built when the build tool first starts up, so that there is always a legacy bundle available, even if it's slightly stale, and commands like `meteor build` and `meteor deploy` still work as before.
* On each subsequent rebuild of the application, we delay starting the `web.browser.legacy` build until after all other builds have finished and the development server has restarted.
* If you access your application using a legacy browser before the legacy build has finished, you'll just get the previous (possibly slightly stale) legacy code. Not the end of the world!
* As soon as the legacy build finishes, the build process sends a message to the `webapp` package running in the server process to tell it to reload the latest legacy bundle.
* Any errors building the legacy bundle get reported in the same way as before.
* Once everything has finished, you'll see the following message in the console:
  ```
  => Finished delayed build of web.browser.legacy in 951ms
  ```
  This message means it's now safe to test your app in a legacy browser, if that's something you were trying to do.

In short, when we talk about zero-configuration here at the Meteor project, we mean something much  deeper than just picking sensible defaults. No, we are striving to eliminate the _need_ for configuration whenever and wherever we can. If you want to hear more about this cornerstone of the Meteor philosophy, check out this [Meteor Night talk](https://youtu.be/vpCotlPieIY?t=34m35s) that I recently gave about the Meteor 1.7 modern/legacy bundling system (or read the [slides](https://slides.com/benjamn/meteor-night-may-2018#/49)).

### DIFF
--- a/packages/meteor/message-dispatch.js
+++ b/packages/meteor/message-dispatch.js
@@ -1,0 +1,61 @@
+// This code receives, dispatches, and responds to inter-process messages
+// sent by the parent (build) process. See tools/runners/run-app.js for
+// the other side of this communications system.
+
+// The process.send method is only defined when the current process was
+// spawned by another process with an IPC channel.
+if (typeof process.send === "function") {
+  const promises = Object.create(null);
+
+  // Listen for messages from the parent process and dispatch them to the
+  // appropriate package, as identified by packageName. To receive these
+  // messages, packages should export an onMessage function that takes the
+  // payload as a parameter.
+  process.on("message", ({
+    type = "FROM_PARENT",
+    responseId = null,
+    packageName,
+    payload,
+  }) => {
+    if (type === "FROM_PARENT" &&
+        typeof packageName === "string") {
+      // Keep the messages and their responses strictly ordered per
+      // package, one after the last. The first message waits for the
+      // package to call Package._define(packageName, exports).
+      promises[packageName] = (
+        promises[packageName] || new Promise(resolve => {
+          Package._on(packageName, resolve);
+        })
+      ).then(
+        // In order to receive messages, the package must export an
+        // onMessage function.
+        () => Package[packageName].onMessage(payload)
+      ).then(result => {
+        if (responseId) {
+          // Send the response value back to the parent using the provided
+          // responseId (if any).
+          process.send({
+            type: "FROM_CHILD",
+            responseId,
+            result,
+          });
+        }
+      }, error => {
+        const copy = {};
+        Reflect.ownKeys(error).forEach(key => copy[key] = error[key]);
+        process.send({
+          type: "FROM_CHILD",
+          responseId,
+          error: copy,
+        });
+      });
+    }
+  });
+
+  // Let the parent process know this child process is ready to receive
+  // messages.
+  process.send({
+    type: "CHILD_READY",
+    pid: process.pid,
+  });
+}
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -553,6 +553,12 @@ WebAppInternals.parsePort = port => {
   return parsedPort;
 }
 
+export async function onMessage(message) {
+  if (message.method === "generateClientProgram") {
+    WebAppInternals.generateClientProgram(...message.args);
+  }
+}
+
 function runWebAppServer() {
   var shuttingDown = false;
   var syncQueue = new Meteor._SynchronousQueue();
@@ -570,9 +576,7 @@ function runWebAppServer() {
         Object.keys(configJson.clientPaths);
 
       try {
-        clientArchs.forEach(arch => {
-          WebAppInternals.generateClientProgram(arch);
-        });
+        clientArchs.forEach(generateClientProgram);
         WebAppInternals.staticFilesByArch = staticFilesByArch;
       } catch (e) {
         Log.error("Error reloading the client program: " + e.stack);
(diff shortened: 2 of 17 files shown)


packages/webapp/webapp_server.js
```diff
@@ -571,92 +567,114 @@ function runWebAppServer() {
...
-              manifest,
-              null,
-              _.pick(__meteor_runtime_config__, "PUBLIC_SETTINGS")
-            ),
-          cordovaCompatibilityVersions: clientJson.cordovaCompatibilityVersions,
-          PUBLIC_SETTINGS: __meteor_runtime_config__.PUBLIC_SETTINGS
-        };
+  process.on("message", async (message) => {
+    if (message.package !== "webapp") return;
+    if (message.method === "generateClientProgram") {
```
benjamn: It's possible that the parent process might send this message before this handler has been registered, but that's ok, because the changes will get picked up when we call `WebAppInternals.reloadClientPrograms` for the first time (which happens later).

hwillson:
This is awesome @benjamn! One small thing:

1. `meteor create test-app`
2. Started the app and accessed it via http://localhost:3100/. So far so good.
3. Added a `console.log('test');` to `client/main.js`.
4. Saw the proper rebuild messages in the console:
```sh
=> Client modified -- refreshing
=> Finished delayed build of web.browser.legacy in 233ms
```
5. Noticed that http://localhost:3100/ did not autoupdate; the newly added `console.log` didn't run.
6. If I manually refresh the browser, it then shows.

benjamn:
@hwillson Great find! I've tracked this down to the code in `webapp` where we calculate client hashes:
https://github.com/meteor/meteor/blob/265d9e30d4f7147e69283f4759bcc8c1794ea226/packages/webapp/webapp_server.js#L191-L209
The problem is that `WebApp.defaultArch` is exactly the architecture that we delayed rebuilding:
https://github.com/meteor/meteor/blob/265d9e30d4f7147e69283f4759bcc8c1794ea226/packages/webapp/webapp_server.js#L42-L44
Since the client hash is the same for `web.browser.legacy` as before, we don't trigger the autoupdate. Seems like this hashing logic should be updated to consider all client bundles, not just `WebApp.defaultArch`.

benjamn:
Although tests are passing after my rewrite of the `autoupdate` package, I'm still concerned about the lack of coordination between the `webapp` package reading the contents of the `.meteor/local/build/programs/web.browser.legacy` bundle during server startup (in the child process), and the build process rewriting the contents of the `web.browser.legacy` bundle (also during server startup, but from the parent process).

I can make the delayed writing of the legacy bundle "more atomic" by populating a fresh temporary directory and then renaming that directory, but that's slower overall because it means rewriting everything rather than just the files that have changed. Writing the changed files in place (and deleting any files that were removed) is fundamentally less instantaneous than renaming a directory, but _much_ faster because it doesn't require creating a temporary directory from scratch.

Here are some ideas I'm considering to address this read/write race:
* Temporarily rename the whole existing `.meteor/local/build/programs/web.browser.legacy` directory to a temporary location, write any changed files, and then rename it back. The `webapp` package could then skip loading the legacy bundle if the directory is missing, trusting that it will be loaded later once the bundle has been written. Problems with this approach:
  * The `webapp` package could start reading the legacy bundle before the build process has a chance to rename it, and I'm not exactly sure how to make the build process wait for the `webapp` package to finish reading the bundle.

* Start writing the `web.browser.legacy` build even later, after the `webapp` package has finished reading the (stale) bundle, so that there's no overlap. The good news here is that we only need to delay the writing of the bundle, but the in-memory computation of the bundle contents (`makeClientTarget`) can begin as early as we like. Problems:
  * The legacy build finishes even later.

* Use some sort of file system synchronization such as [`lockfile`s](https://www.npmjs.com/package/lockfile). Problems:
  * This package (and related packages) appear to be based on the [`O_EXCL` flag](https://nodejs.org/api/fs.html#fs_file_system_flags), which doesn't work on all file systems (e.g. network file systems like NTFS).

I'm going to keep thinking about this until I have more confidence in one solution over the others.

tools/isobuild/bundler.js
```diff
@@ -3181,80 +3138,92 @@ function bundle({
...
+          writeClientTarget(target);
+
+          // Refresh and unpause the client, now that writing is finished.
+          await refreshClient(arch);
+
+          // Let the webapp package running in the child process know it
+          // should regenerate the client program for this arch.
+          runLog.log(`Finished delayed build of ${arch} in ${
+            new Date - start
+          }ms`, { arrow: true });
```
benjamn: Need to think about whether this logging should remain in the final version. Alternatively, we could add more information about building other bundles, so that this isn't the only message.

packages/meteor/message-dispatch.js
```diff
@@ -0,0 +1,61 @@
...
+    payload,
+  }) => {
+    if (type === "FROM_PARENT" &&
+        typeof packageName === "string") {
+      // Keep the messages and their responses strictly ordered per
+      // package, one after the last. The first message waits for the
+      // package to call Package._define(packageName, exports).
+      promises[packageName] = (
+        promises[packageName] || new Promise(resolve => {
+          Package._on(packageName, resolve);
```
benjamn: This pattern causes a bug when the `autoupdate` package (or any other package that receives messages) is not installed, which means `Package._on("autoupdate", resolve)` will never call `resolve`, so the build process stops receiving messages from the server process and `await`s forever.

PR #10058, opened 2018-07-07
Update MongoDB to 4.0

klaussner:
This PR updates MongoDB to version 4.0 and the driver to version 3.1.
Fixes meteor/meteor-feature-requests#269.

### DIFF
--- a/tools/tests/mongo.js
+++ b/tools/tests/mongo.js
@@ -32,11 +32,9 @@ var testMeteorMongo = function (appDir) {
   run.match('Started MongoDB');
 
   var mongoRun = s.run('mongo');
-  mongoRun.match('MongoDB shell');
-  mongoRun.match(/connecting to: (.*)127.0.0.1/);
   // Note: when mongo shell's input is not a tty, there is no prompt.
   mongoRun.write('db.version()\n');
-  mongoRun.match(/3\.\d+\.\d+/);
+  mongoRun.match(/4\.\d+\.\d+/);
   mongoRun.stop();
 
   run.stop();


packages/mongo/mongo_driver.js
```diff
@@ -133,7 +133,9 @@ MongoConnection = function (url, options) {
     // Try to reconnect forever, instead of stopping after 30 tries (the
     // default), with each attempt separated by 1000ms.
     reconnectTries: Infinity,
-    ignoreUndefined: true
+    ignoreUndefined: true,
+    // Use new URL parser to avoid deprecation warning.
+    useNewUrlParser: true
```
klaussner: The driver shows this warning by default:
```
DeprecationWarning: current URL string parser is deprecated, and will be removed in a future version. To use the new parser, pass option { useNewUrlParser: true } to MongoClient.connect.
```

hwillson:
@klaussner One small thing - it looks like 3.6.4 is still being used for `meteor mongo`?

screenshot 2018-07-11 09 53 31

klaussner:
@hwillson Did you rebuild the dev bundle? This is the output I get from the `mongo` shell:

```sh
$ meteor-dev mongo
MongoDB shell version v4.0.0
connecting to: mongodb://127.0.0.1:3001/meteor
MongoDB server version: 4.0.0
Server has startup warnings:
2018-07-12T21:40:42.422+0200 I STORAGE  [initandlisten]
2018-07-12T21:40:42.422+0200 I STORAGE  [initandlisten] ** WARNING: Using the XFS filesystem is strongly recommended with the WiredTiger storage engine
2018-07-12T21:40:42.422+0200 I STORAGE  [initandlisten] **          See http://dochub.mongodb.org/core/prodnotes-filesystem
meteor:PRIMARY>
```

hwillson:
Hi @klaussner - I did rebuild the dev_bundle, but when I just tried again I'm now seeing bizarre `pipelineOperator` errors whenever I try to run the new build. I tested this out on a new laptop (with this as the first Meteor test), so I'm going to assume something is up on my end. The code changes look good, so my earlier 👍 still stands. I'll post back shortly after I reconfigure things, but if you're seeing the proper Mongo shell, then I'm happy. Thanks again for working on this!

hwillson:
Quick update relating to https://github.com/meteor/meteor/pull/10058#issuecomment-404989839 - the `pipelineOperator` issue was fixed in https://github.com/benjamn/reify/commit/9985a492fdf7096b7877ab407b33170208865c50, and will be addressed in `devel` shortly (which will make testing this PR much easier).

hwillson:
After running with the new `reify` version, I now see:

```sh
➜  pr-10058 /Users/hwillson/Documents/git/meteor/forks/klaussner/meteor/meteor mongo
MongoDB shell version v4.0.0                  
connecting to: mongodb://127.0.0.1:3001/meteor
MongoDB server version: 4.0.0
---
Enable MongoDB's free cloud-based monitoring service to collect and display
metrics about your deployment (disk utilization, CPU, operation statistics,
etc).

The monitoring data will be available on a MongoDB website with a unique
URL created for you. Anyone you share the URL with will also be able to
view this page. MongoDB may use this information to make product
improvements and to suggest MongoDB products and deployment options to you.

To enable free monitoring, run the following command:
db.enableFreeMonitoring()
---

meteor:PRIMARY> 
```

Should we suppress Mongo's free monitoring message (is that dynamically pulled in?), or leave it as is?  Either way , it's working - thanks @klaussner!

klaussner:
Interesting, I did't get the free monitoring message, just the filesystem warning. :man_shrugging: The only way I found to suppress them is to pass the `--quiet` option to the `mongo` shell ([here](https://github.com/meteor/meteor/blob/e91b3776e54ca9e1607bf89e0cebd32170eb2c1d/tools/runners/run-mongo.js#L25)) but then the version info and URI also disappear:

```
$ meteor-dev mongo
meteor:PRIMARY>
```

On the other hand, they can easily be found using the shell, so I think I'm in favor of using the `--quiet` option:

```
$ meteor-dev mongo
meteor:PRIMARY> version()
4.0.0
meteor:PRIMARY> db.serverCmdLineOpts().parsed.net
{ "bindIp" : "127.0.0.1", "port" : 3001 }
meteor:PRIMARY>
```

hwillson:
Hi @klaussner - 👍 on the `--quiet` flag. If you don't mind making that change, we'll get this merged (and if you're busy just let me know and I'll make the necessary changes). Thanks!

benjamn:
In today's triage meeting @abernix @hwillson and I discussed the `--quiet` option, and we think it's a good idea, since it appears to apply only during the initial connection phase, and doesn't affect other logging. If you add `--quiet` and rebase against `devel` I will merge this into `devel` (and then `release-1.7.1`), which should be one of the last things we do before RC.

klaussner:
Thanks, @hwillson. I've added the `--quiet` flag and rebased onto the current `devel`. :slightly_smiling_face:

benjamn:
I've rebased these changes onto `release-1.7.1` and built a new dev bundle (8.11.3.13): a875ac5547daa4af47b66b50e75b762f3cf1fd66

I think there may still be some test failures, judging from local `meteor self-test` attempts. Whenever you have a chance to look into that, just rebase against `release-1.7.1`, change the target branch of this PR, and add any additional commits as you see fit.

Thanks for working on this @klaussner!

benjamn:
I went digging for another way to disable the free monitoring advertisement, and I found this command-line option: [`--enableFreeMonitoring off`](https://docs.mongodb.com/manual/reference/program/mongod/#cmdoption-mongod-enablefreemonitoring)

However, when I add that option to the `mongod` arguments, the process crashes with a segmentation fault!

At least, that's what happens on OSX. On Linux, I get
```
Error parsing command line: unrecognised option '--enableFreeMonitoring'
try '/home/ben/meteor/dev_bundle/mongodb/bin/mongod --help' for more information
```
I wrote up these findings in their JIRA issue pertaining to the problem: https://jira.mongodb.org/browse/SERVER-36474

It's hard to know how worried we should be about the general stability of Mongo 4.0.0. Should we be concerned about merely silencing (rather than disabling) this poorly-tested feature?

mitar:
@klaussner Have you checked how the new transactions feature interact with Meteor's pub/sub and oplog tailing? We should make sure that if you do a transaction and modify data, that this is not leaked through oplog to pub/sub. Probably pub/sub should push updates to the client only after the transaction commits.

klaussner:
@mitar I haven't tried it but I assume that transactions don't work with pub/sub at the moment. Each transaction has its own oplog entry, which includes a list of operations (inserts, updates, etc.) that were committed successfully but Meteor doesn't handle this type of oplog entry yet, I think.

mitar:
I think oplog entries just have a transaction ID associated with it. What I worry is that if Meteor oplog tailing ignores that entry, it might update based on some operations happening inside the transaction, which should not yet be visible outside. Or are you saying that operations inside transactions are not send over the oplog until committed? I do not know details here though. Maybe you are right. I just see a potential issue here.

klaussner:
Exactly, only operations of successful transactions will appear on the oplog. I opened an issue with more details: #10195.

PR #10077, opened 2018-07-16
Use listener API for communication between build/server processes.

### LINKED ISSUE #10073: [1.7.1-beta.17] client rebuild doesn't behave correctly with dynamic imports
We just upgraded from beta 15 to 17 and are facing an issue in development related to rebuilds:

When changing a client/common file the server shows:
```sh
=> Client modified -- refreshing
=> Finished delayed build of web.browser.legacy in 5377ms
```
and the browser reloads, but the code doesn't actually change.

I tried to create a simple reproduction, but didn't succeed. I think it's somehow related to dynamic imports. It seems like changes in `client/main.js` are recompiled successfully and reflected inside the browser - when i change things in files imported somewhere down the road via react-loadable they aren't present on the client and i have to manually restart meteor to see them inside the browser.

benjamn:
Fixes #10073, per https://github.com/meteor/meteor/issues/10073#issuecomment-405290391

While thinking about this bug, I realized that sending IPC messages to specific packages in the server process was much less flexible than sending messages based on an arbitrary topic string, since the topic string approach allows both `autoupdate` and `dynamic-import` to listen for the same message.

The topic string approach calls for a listener interface like `onMessage(topic, callback)`, which elegantly replaces the previous approach of requiring packages to export a single `onMessage` function.

However, because the `meteor` package does not have access to the module system, implementing the `onMessage` listener interface in the `meteor` package would have required exposing an API like `Meteor.onMessage(topic, callback)`, which has an unpleasant global smell to it. Instead, the `onMessage` function should be explicitly imported (using the module system) from some less-generically-named package.

Since I knew I was going to have to move the message dispatch logic out of the `meteor` package, I decided to create a new package called `inter-process-messaging` to implement the parent/child components of the IPC system.


packages/meteor/message-dispatch.js
```diff
@@ -1,59 +0,0 @@
...
-    packageName,
-    payload,
-  }) => {
-    if (type === "FROM_PARENT" &&
-        typeof packageName === "string") {
-      // Keep the messages and their responses strictly ordered per
-      // package, one after the last. The first message waits for the
-      // package to call Package._define(packageName, exports).
-      promises[packageName] = (
-        promises[packageName] || Package._promise(packageName)
```
benjamn: Since we're no longer using `Package._promise` in core, and it was only recently introduced for this (now removed) code, I think we should remove it for the sake of keeping the APIs exposed by the core `meteor` package as small as possible.

tools/runners/run-app.js
```diff
@@ -846,28 +790,17 @@ _.extend(AppRunner.prototype, {
...
-      await appProcess.proc.sendMessage("autoupdate", {
-        refresh: "client"
-      }).catch(error => {
-        if (error.message !== "Package autoupdate not installed") {
-          throw error;
-        }
-      });
+      // If arch is not a string, the receiver of this message should
+      // assume all clients need to be refreshed.
+      await appProcess.proc.sendMessage("client-refresh");
```
benjamn: If the `autoupdate` package is not installed, no message will be delivered to that package, so no error needs to be caught here.

packages/dynamic-import/server.js
```diff
@@ -187,12 +187,15 @@ function getCache(platform) {
...
+
+onMessage("client-refresh", () => {
+  // The caches for the web.browser[.legacy] platforms need to be
+  // discarded whenever a client-only refresh occurs, so the new client
+  // bundle does not fetch stale module data from dynamic import(). This
+  // message is sent by tools/runners/run-app.js and also consumed by the
+  // autoupdate package.
+  Object.keys(cachesByPlatform).forEach(platform => {
+    delete cachesByPlatform[platform];
+  });
```
benjamn: This is the part of the PR that actually fixes #10073.

packages/inter-process-messaging/child.js
```diff
@@ -0,0 +1,83 @@
...
+      // one after the last. Because we always register a non-throwing
+      // error handler below, this queue of promises should never get
+      // stalled by an earlier rejection.
+      promisesByTopic[topic] = (
+        promisesByTopic[topic] || Promise.resolve()
+      ).then(() => {
+        const results = [];
+        const callbacks = callbacksByTopic.get(topic);
+        if (callbacks && callbacks.size > 0) {
+          callbacks.forEach(cb => results.push(cb(payload)));
```
benjamn: One really nice benefit of using `Set#forEach` here is that if you add new elements to the set while you're traversing it, they will be included in the same `forEach` loop!

benjamn:
I would like to extract this functionality as an npm package one day, but I think I need to draw the line on this endless refactoring somewhere.

PR #10098, opened 2018-07-25
Allow json5 in babelrc

### LINKED ISSUE #10068: prettier reformats babelrc file to json5, but meteor does not like it
prettier reformats .babelrc files to JSON5, which should be supported by babel according to this thread:
https://github.com/prettier/prettier/issues/4636

but meteor does not seem to understand it and throws:

```
 While processing files with ecmascript (for target web.browser):
   <anonymous>: .babelrc is not a valid JSON file: Unexpected token p in JSON at position 4
   at JSON.parse (<anonymous>)
   at BabelCompiler.BCp._inferFromBabelRc (packages/babel-compiler.js:255:16)
   at BabelCompiler.BCp.inferExtraBabelOptions (packages/babel-compiler.js:244:10)
   at BabelCompiler.BCp.processOneFileForTarget (packages/babel-compiler.js:174:10)
   at BabelCompiler.<anonymous> (packages/babel-compiler.js:111:26)
   at BabelCompiler.BCp.processFilesForTarget (packages/babel-compiler.js:110:14)
```

EDIT: happening on meteor 1.6.1. Unfortunatly, I can't update to 1.7.0 atm because of pending issues

sakulstra:
Currently we cannot us json5 in babel files which are used by meteor. That's a bit counter intuitive as babel itself supports json5 and we share babel files between jest and meteor. So atm we cannot e.g. add comments in babel files which are used by meteor.

This pr changes the parser the use json5 instead - which as i get should be fully backwards compatible and is already included via babel.

- fixes #10068 

-----

Not sure how to properly add a test for this, but I tested locally by adding a json5 .babelrc and running from checkout:
```
{
  env: {
    test: {
      plugins: ['dynamic-import-node']
    }
  }
}
```

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -1,5 +1,5 @@
 var semver = Npm.require("semver");
-
+var JSON5 = Npm.require("json5");
 /**
  * A compiler that can be instantiated with features and used inside
  * Plugin.registerCompiler
@@ -195,7 +195,7 @@ BCp._inferFromBabelRc = function (inputFile, babelOptions, cacheDeps) {
       try {
         this._babelrcCache[babelrcPath] = {
           controlFilePath: babelrcPath,
-          controlFileData: JSON.parse(
+          controlFileData: JSON5.parse(
             inputFile.readAndWatchFile(babelrcPath)),
           deps: Object.create(null),
         };
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -201,7 +201,7 @@ BCp._inferFromBabelRc = function (inputFile, babelOptions, cacheDeps) {
         };
       } catch (e) {
         if (e instanceof SyntaxError) {
-          e.message = ".babelrc is not a valid JSON file: " + e.message;
+          e.message = ".babelrc is not a valid JSON5 file: " + e.message;
         }
         throw e;
       }
--- a/packages/babel-compiler/package.js
+++ b/packages/babel-compiler/package.js
@@ -10,7 +10,8 @@ Package.describe({
 });
 
 Npm.depends({
-  'meteor-babel': '7.0.0-rc.1'
+  'meteor-babel': '7.0.0-rc.1',
+  'json5': '1.0.1'
 });
 
 Package.onUse(function (api) {


packages/babel-compiler/babel-compiler.js
```diff
@@ -195,7 +195,7 @@ BCp._inferFromBabelRc = function (inputFile, babelOptions, cacheDeps) {
       try {
         this._babelrcCache[babelrcPath] = {
           controlFilePath: babelrcPath,
-          controlFileData: JSON.parse(
+          controlFileData: JSON5.parse(
```
klaussner: Maybe update the error message below to `.babelrc is not a valid JSON5 file`? :slightly_smiling_face:

sakulstra: kk changed it

benjamn:
Isn't JSON5 a strict superset of JSON? Would there be any risk to monkey-patching `JSON.parse` globally to make it tolerant of comments? Note that monkey-patching `JSON.stringify` would probably be a bad idea, because then the generated text would not be legal JSON.

sakulstra:
@benjamn actually I thought about this, but wasn't quite sure if that the best idea. JSON5 is used by babel, but it's not a wide spread standard as i get it - when monkey patching, people might wonder why they can use JSON.parse to parse their JSON5 package.json, but npm will trow errors as it cannot work with JSON5 yet (assuming nothing changed since i last checked).

packages/babel-compiler/babel-compiler.js
```diff
@@ -1,5 +1,5 @@
 var semver = Npm.require("semver");
-
+var JSON5 = Npm.require("json5");
```
benjamn: For this to work in all environments, we need to add the `json5` npm dependency here:
https://github.com/meteor/meteor/blob/d1ca1f71c5992079836b36c4766394c7eab90e67/packages/babel-compiler/package.js#L12-L14
If you run `meteor test-packages ./babel-compiler` in the `meteor/packages` directory, that should automatically update the `meteor/packages/babel-compiler/.npm/package/npm-shrinkwrap.json` file, and those changes should be committed, too.

benjamn:
Following up on my previous comment, I think it's fine to keep this PR focused on `.babelrc`, since people are likely to have `.babelrc` files with comments etc. that work for other projects/workflows, and we should avoid breaking them if possible. We can revisit the broader question later. Thanks!

packages/babel-compiler/babel-compiler.js
sakulstra: @benjamn I applied the changes, but am somehow not able to run the test command - it always exits with:
```
=> Errors prevented startup:                  
   
   error: No version of babel-compiler satisfies all constraints: @=7.2.0-beta171.31, @~7.1.1
   Constraints on package "babel-compiler":
   * babel-compiler@=7.2.0-beta171.31 <- top level
   * babel-compiler@~7.1.1 <- top level
   * babel-compiler@7.1.0 <- ecmascript 0.11.1 <- allow-deny 1.1.0 <- mongo 1.5.0 <- autoupdate 1.4.1
   * babel-compiler@7.1.0 <- ecmascript 0.11.1 <- autoupdate 1.4.1
´´´

PR #10104, opened 2018-07-28
Fix another map function in oauth1

### LINKED ISSUE #10094: [1.7.1-beta.21]  Error in OAuth Server: header.reduce is not a function
It seems like the last update to accounts packages (modernization) breaks twitter-oauth package. 

Error logged on the server when attempting to authenticate:
`Error in OAuth Server: header.reduce is not a function`

Related commit:
https://github.com/meteor/meteor/commit/c8f4ba74c74bdee3793f49aad3bb469e0cc1d62d#diff-6954d371e89d59bb7c5ffa2e6d80194f

jamesmillerburgess:
Fixes #10094 again.

I've made this fix but have not yet had the time to review all of the accounts modernization changes for similar errors. It will probably be another day or two before I will get some time to do that. 

I figured that pushing this fix will at least allow us to publish another 1.7.1 beta version and see if other errors are discovered by our excellent testers like @nick-gudumac!

### DIFF
--- a/packages/oauth1/oauth1_binding.js
+++ b/packages/oauth1/oauth1_binding.js
@@ -100,7 +100,7 @@ export class OAuth1Binding {
   _getSignature(method, url, rawHeaders, accessTokenSecret, params) {
     const headers = this._encodeHeader({ ...rawHeaders, ...params });
 
-    const parameters = headers.map((val, key) => `${key}=${val}`)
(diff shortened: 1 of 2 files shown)


jamesmillerburgess:
I searched through all of the accounts packages for usages of `map` and `reduce` and didn't find any similar errors, so hopefully this one was the last of them!

PR #10110, opened 2018-07-30
Fix: file `Meteor_:-@2x.png` breaks development on Windows

mrauhu:
Hello.

This pull request fix the filename `Meteor_:-@2x.png` that not allowed on Windows OS, because it contains the reserved symbol `:` (colon).

The bug was introduced in https://github.com/meteor/meteor/pull/8502.

Without this fix the Meteor development on a Windows machine is broken.

More about Windows naming conventions:

> https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file#naming-conventions

> Use any character in the current code page for a name, including Unicode characters and characters in the extended character set (128–255), except for the following:
>
> The following reserved characters:
>
>  \< (less than)
  \> (greater than)
  \: (colon)
  \" (double quote)
  \/ (forward slash)
  \\ (backslash)
  \| (vertical bar or pipe)
  \? (question mark)
  \* (asterisk)

Thanks,
Sergey.


abernix:
Thanks for taking care of this.  I know from personal experience that this can be a pain point when developing Meteor (itself) on Windows.  The original spirit of this test was to make sure that files worked with the `@` symbol — which is an Apple filename format to specify different assets for different screen resolutions which needs to work within Cordova apps.

Issue #9735, opened 2018-03-11
Socket.io with uWS instead SockJS


Maxpain177:
I propose to change `SockJS` with `Socket.io` (for long polling as fallback) + `UWS` (uWebSockets, which is faster, than any JavaScript WebSocket implementation, because it's written as C++ module, with built-in support `permessage-deflate`).

I think this will give a big performance boost for meteor server and less CPU consumption.

hwillson:
Thanks @Maxpain177 - this is already being tracked here: https://github.com/meteor/meteor-feature-requests/issues/41

PR #10118, opened 2018-08-02
Modular DDP stream server

mrauhu:
Hello, @benjamn 

I make the stream server changeable.

So basically, you can write any custom stream server and plug it to the Meteor on the fly.

More details:

<details><summary>stream-server/README.md</summary>

# Meteor's DDP stream server based on SockJS

> Network core of the Meteor's DDP.

The `stream-server` package provides:

* the default Meteor's stream server based on SockJS;
* the mechanism to replace bundled `StreamServer`.

## How to add your custom `StreamServer` to `StreamServers` and activate it

For example your package has name `custom-stream-server`.

1. Enable access to the `StreamServers` variable.

    `custom-stream-server/package.js`

    ```js
    Package.onUse(function (api) {
      // Add following lines
      api.use(['stream-server'], 'server');
      api.imply(['stream-server'], 'server');
      api.export('StreamServers', 'server');
      // ...
    });
    ```

2. Push `CustomStreamServer` to the `StreamServers` array.

    `custom-stream-server/server.js`

    ```js
    class CustomStreamServer {
      constructor() {
        // this.server = new WebSocket.Server();
      }
    }
    
    StreamServers.push(CustomStreamServer);
    ```

3. Activate your custom stream server.

    In the `.meteor/packages` place your package name before packages:
    
    * `meteor-tools`
    * `ddp`
    * `ddp-server`

    Example of the `.meteor/packages`
    
    ```
    custom-stream-server
    meteor-tools
    # other packages
    ```
</details>
`

Greatness awaits,
Sergey.

PR #10120, opened 2018-08-02
🚀 ClusterWS support in the Meteor

mrauhu:
This PR adds optional support for using uWebsockets a.k.a. `uws` in a new (optional) package called `stream-server-uws`.

### Update 2020-04-04

* Added support for Meteor version `>= 1.9`.
* Created repository with updated guide
https://github.com/mrauhu/meteor-uwebsockets

## How to use it now

<details><summary>☄ Guide (click to expand)</summary>

### Prerequisites

Open your project:

```bash
cd your-project-root
```

If your project not a Git repository then initialize it:

```bash
git init
```

### Installing

1. Replace [`ddp-server`][1] and [`socket-stream-client`][2], install [`disable-sockjs`][3], [`stream-server`][4] and [`stream-server-uws`][5] packages via Git:

    [1]: https://github.com/mrauhu/ddp-server
    [2]: https://github.com/mrauhu/socket-stream-client
    [3]: https://github.com/mrauhu/disable-sockjs
    [4]: https://github.com/mrauhu/stream-server
    [5]: https://github.com/mrauhu/stream-server-uws

    ```bash
    git submodule add https://github.com/mrauhu/ddp-server packages/ddp-server
    git submodule add https://github.com/mrauhu/disable-sockjs packages/disable-sockjs
    git submodule add https://github.com/mrauhu/socket-stream-client packages/socket-stream-client
    git submodule add https://github.com/mrauhu/stream-server packages/stream-server
    git submodule add https://github.com/mrauhu/stream-server-uws packages/stream-server-uws
    ```

3. Enable Stream Server with the `@clusterws/cws`, fork of the `uws`:

    ```bash
    meteor add stream-server-uws
    ```

### Usage

Run the Meteor:

```
meteor
```
</details>

---

This package only works with WebSocket transport.

Also I made changes for the `ddp-server/livedata_server.js`, because it didn't support the standard WebSocket event `message`.

Resolves:
* Fix: https://github.com/meteor/meteor-feature-requests/issues/75 Switch to µWebSockets
* Partially fix: https://github.com/meteor/meteor/issues/9735 Socket.io with uWS instead SockJS

Includes changes from pull requests: 
* https://github.com/meteor/meteor/pull/10118 Modular DDP stream server
* https://github.com/meteor/meteor/pull/10110 Fix: file `Meteor_:-@2x.png` breaks development on Windows

# Benchmark

<details><summary>Arguments</summary>

### Linux & macOS

```bash
NODE_OPTIONS=--max-old-space-size=8196 meteor
```

### Windows

```cmd
set NODE_OPTIONS=--max-old-space-size=8196 && meteor
```
</details>

## Code

<details>
<summary>Meteor method</summary>

```js
Meteor.methods({
  async time_diff_1s(client_date) {
    await timeout(1000);
    const server_date = new Date();
    return server_date - client_date;
  },
});

/**
 * Timeout
 * @param ms
 * @return {Promise<any>}
 */
function timeout(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}
```
</details>

<details><summary>Loader</summary>

```bash
npm install sc-uws
```

```js
const WebSocket = require('sc-uws');

const DDP_URL = 'ws://127.0.0.1:3000/websocket';

const PING_INTERVAL_MS = 30 * 1000;
const MAX = 10 ** 5;
const STDOUT_UPDATE_INTERVAL_MS = 1000;

const ws = new WebSocket(DDP_URL);

ws.on('open', function open() {
  console.log('Connected - ' + new Date());
  ws.send('{"msg":"connect","version":"1","support":["1","pre2","pre1"]}');

  for (let i = 0; i < MAX; i++) {
    ws.send('{"msg":"method","method":"time_diff_1s","params":["' + now + '"],"id":"1"}');
  }
});

ws.on('error', function error() {
  console.log('Error connecting!');
});

// Store last message in memory, instead of outputting to STDOUT
let m = 0;
let message = '';
ws.on('message', function(data, flags) {
  ++m;
  // Don't write ping/pong to log
  // Response to ping from server
  if (data === '{"msg":"ping"}') {
    ws.send('{"msg":"pong"}');
    return;
  }
  if (data === '{"msg":"pong"}') {
    return;
  }
  message = m + ' ' + new Date() + ' Message: ' + data;
});

// Write last stored message from memory to STDOUT
let old_m = 0;
setInterval(() => {
  if (m > old_m) {
    process.stdout.write('\r' + message);
    old_m = m;
  }
}, STDOUT_UPDATE_INTERVAL_MS);

// Ping server
setInterval(() => {
  return ws.send('{"msg":"ping"}');
}, PING_INTERVAL_MS);

ws.on('close', function(code, message) {
  console.log('Disconnection: ' + code + ', ' + message);
});
```

</details>

## Results

Windows 8.1
CPU i5-5200U 2x2.20 GHz
RAM 16 GB

### SockJS `stream-server`

Memory usage for processing 10 000 method calls is critical — approximate 8 GB.

| Method calls | Time to process, s | Calls per second |
|-------------:|-------------------:|-----------------:|
|        1 000 |                0.8 |            1 250 |
|       10 000 |                 28 |              357 |
|      100 000 |        fatal error |    out of memory |
|    1 000 000 |        fatal error |    out of memory |

### uWebSockets `stream-server-uws`

| Method calls | Time to process, s | Calls per second |
|-------------:|-------------------:|-----------------:|
|        1 000 |                0.7 |            1 428 |
|       10 000 |              11.82 |              846 |
|      100 000 |             136.70 |              726 |
|    1 000 000 |            1785.53 |              560 |

Respectfully,
Sergey.


KoenLav:
uWebSockets for Node seems to be no longer actively maintained because the maintainer is throwing a fuzz on NPM?

mrauhu:
> uWebSockets for Node seems to be no longer actively maintained because the maintainer is throwing a fuzz on NPM?

@KoenLav as far as I know author of uWebSockets working on version `0.15`, working on the core library:
https://github.com/uNetworking/uSockets
 
In this pull request I'm use the fork `sc-uws` maintained by SocketCluster.io:
https://github.com/SocketCluster/sc-uws

packages/stream-server-uws/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+3. Install the `stream-server-uws` package:
+
+    ```bash
+    meteor add stream-server-uws
+    ```
+
+## Activation
+
+In the `.meteor/packages` place the package `stream-server-uws` before packages:
```
benjamn: You should be able to skip this recommendation if you put
```js
api.use("stream-server-uws", { weak: true });
```
in the `package.js` files for these modules, so that `stream-server-uws` will load before them (but only if it is installed).

packages/stream-server-uws/README.md
```diff
@@ -0,0 +1,41 @@
+# Meteor's DDP stream server based on uWS
+
+## Installation
+
+1. Install the fork of the uWebSocket Node.js bindings:
+
+    ```bash
+    meteor npm install sc-uws
```
benjamn: How is [`sc-uws`](https://www.npmjs.com/package/sc-uws) different from the deprecated [`uws` npm package](https://www.npmjs.com/package/uws), other than not being deprecated? Relatedly, where is the repository from which `sc-uws` gets `npm publish`ed (the GitHub links still point to the old project)? I want to make sure this `sc-uws` package is actively maintained, and not just a fork of the deprecated code.

packages/stream-server-uws/package.js
```diff
@@ -0,0 +1,18 @@
+Package.describe({
+  summary: "Meteor's DDP stream server based on uWS",
+  version: '0.5.1',
+  documentation: 'README.md'
+});
+
+Package.onUse(function (api) {
+  api.use(['modules'], 'server');
```
benjamn: Let's just do `api.use("ecmascript")` here, to get full ES module support.

packages/stream-server-uws/stream_server_uws.js
```diff
@@ -0,0 +1,90 @@
...
+const { WebApp } = require('meteor/webapp');
+
+// Try load of `uws` Node.js package
+/** @type {WebSocket} */
+let WebSocket;
+try {
+  // noinspection NpmUsedModulesInstalled
+  WebSocket = require('sc-uws');
+} catch (e) {
+  console.warn('Run:\nmeteor npm install sc-uws')
```
benjamn: You should check that the error here is actually due to `sc-uws` not being installed, rather than some error thrown while trying to evaluate the package.

One way to do this is to use `require.resolve` first, which returns the absolute module identifier if the package is installed, without actually importing it, and otherwise throws an error:

```js
let uwsAbsId;
try {
  uwsAbsId = require.resolve("sc-uws");
} catch (e) {
  console.warn(...);
}
if (typeof uwsAbsId === "string") {
  WebSocket = require(uwsAbsId);
}
```

packages/stream-server-uws/stream_server_uws.js
```diff
@@ -0,0 +1,90 @@
...
+  path: pathPrefix + PATH,
+  server: WebApp.httpServer,
+};
+
+/**
+ * StreamServer with uws
+ * @class
+ * @type {StreamServerUWS}
+ */
+StreamServerUWS = class StreamServerUWS {
```
benjamn: If you use the `ecmascript` package, this can just be
```js
export class StreamServerUWS {
```

packages/stream-server/README.md
```diff
@@ -0,0 +1,56 @@
...
+    `custom-stream-server/server.js`
+
+    ```js
+    class CustomStreamServer {
+      constructor() {
+        // this.server = new WebSocket.Server();
+      }
+    }
+    
+    StreamServers.push(CustomStreamServer);
```
benjamn: In this example, let's recommend explicitly importing `StreamServers`:
```js
import { StreamServers } from "meteor/stream-server";
```
It might also be good to expose a function for adding new servers, instead of just an array, so that we have a hook for running initialization code after a new server is added.

packages/stream-server/README.md
```diff
@@ -0,0 +1,56 @@
...
+    * `meteor-tools`
+    * `ddp`
+    * `ddp-server`
+
+    Example of the `.meteor/packages`
+    
+    ```
+    custom-stream-server
+    meteor-tools
+    # other packages
```
benjamn: For example, this recommendation might become unnecessary if we can actively let the other packages know when a new stream server has been added, rather than just relying on the load order of the packages.

packages/stream-server/package.js
```diff
@@ -0,0 +1,20 @@
...
+  documentation: 'README.md'
+});
+
+Npm.depends({
+  "permessage-deflate": "0.1.6",
+  sockjs: "0.3.19"
+});
+
+Package.onUse(function (api) {
+  api.use(['underscore'], 'server');
```
benjamn: We've been trying to remove `underscore` from core packages for a while now. Let's avoid it here if possible.

packages/stream-server/stream_server.js
```diff
@@ -27,6 +27,16 @@ var websocketExtensions = _.once(function () {
...
 
+/**
+ * List of stream servers
+ *
+ * DDP LiveData server will load the latest added stream server.
+ * You need place custom stream server package in `.meteor/packages`
+ * before packages `meteor-tools`, ddp` or `ddp-server`
+ * @type {(StreamServer|*)[]}
+ */
+StreamServers = [];
```
benjamn: Per my previous comments, this should become something like
```js
export function addCustomStreamServer(server) { ... }
```

packages/stream-server-uws/README.md
mrauhu: `uws` package "deprecated" because the author don't like `semver` policy of `npm`.

> How is sc-uws different from the deprecated uws npm package, other than not being deprecated?

It have the last stable uWebSockets version, without strange changes, like deleting of `http.h` or `addon.h` in bindings.

> Relatedly, where is the repository from which sc-uws gets npm published (the GitHub links still point to the old project)?

https://github.com/SocketCluster/uWebSockets

>  I want to make sure this `sc-uws` package is actively maintained, and not just a fork of the deprecated code.

There is no actively maintained package of uWebSockets.

Maybe Alex Hulman @alexhultman publish new package with Node.js binding, maybe not.

packages/stream-server/package.js
mrauhu: @benjamn I has copy/paste code of old `stream_server.js` to `stream-server` package.

I can rewrite it to the full ES2015+ support if you need so.

packages/stream-server-uws/package.js
mrauhu: Package `ecmascript` it is not ECMAScript modules for Node.js, it is the Babel.

After Babel compilation the result code is horrible.

I wrote the `stream-server-uws` package for high performance, and I don't want any possible interference from Babel. 

## use `modules` way

### Code

`stream_server_uws.js`

```js
const { RoutePolicy } = require('meteor/routepolicy');
const { WebApp } = require('meteor/webapp');
// ...
StreamServerUWS = class StreamServerUWS {
  // ...
}
```

### Result

`.meteor/local/build/programs/server/packages/stream-server-uws.js`

```js
const { RoutePolicy } = require('meteor/routepolicy');
const { WebApp } = require('meteor/webapp');
// ...
StreamServerUWS = class StreamServerUWS {
  // ...
}
```

## use `ecmascript` way

### Code

```js
const { RoutePolicy } = require('meteor/routepolicy');
const { WebApp } = require('meteor/webapp');
// ...
export class StreamServerUWS {
  // ...
}
```

### Result

`.meteor/local/build/programs/server/packages/stream-server-uws.js`

```js
const module1 = module;
module1.export({
  StreamServerUWS: () => StreamServerUWS
});
let RoutePolicy;
module1.watch(require("meteor/routepolicy"), {
  RoutePolicy(v) {
    RoutePolicy = v;
  }

}, 0);
let WebApp;
module1.watch(require("meteor/webapp"), {
  WebApp(v) {
    WebApp = v;
  }

}, 1);
// ...
class StreamServerUWS {
  // ...
}
```

packages/stream-server-uws/stream_server_uws.js
mrauhu: Okay, I will change that.

packages/stream-server-uws/stream_server_uws.js
mrauhu: I wrote more about this in the comment:
https://github.com/meteor/meteor/pull/10120#discussion_r207611025

packages/stream-server/stream_server.js
mrauhu: I don't found the more simple and elegant way of adding a custom stream server into the Meteor. 

2f8g49

My solution is:

1. Create the `StreamServers` array in the base package `stream-server`.
2. Place `custom-stream-server` before package that loading `ddp-server`.
3. In the `custom-stream-server` `.use()` and `.imply()` `stream-server`, then push the new stream server to the `StreamServers` array.
4. Get the last added stream server  in `livedata_server.js`.

Maybe, I don't know the right way changing of DDP stream server on the fly.

@benjamn for what task we trying to find the solution:
a) add the `stream-server-uws` package as a weak dependency to the `ddp-server`;
b) make the `stream-server` changeable?

packages/stream-server-uws/README.md
mrauhu: Yes, I could write this package like this:

`ddp-server/package.js`

```js
  api.use("stream-server-uws", { weak: true });
}
```

`ddp-server/livedata.js`

```js 
  var StreamServer = StreamServerUWS || StreamServer;
  self.stream_server = new StreamServer;
```

But in this case we don't provide our users the simple way of changing of the stream server.

So I create the `StreamServers` array, and write guide how to extend it right way.

packages/stream-server/README.md
mrauhu: I don't found the way how to change the DDP `Server.stream_server` after the DDP has initialized.

packages/stream-server/README.md
mrauhu: > It might also be good to expose a function for adding new servers, instead of just an array, so that we have a hook for running initialization code after a new server is added.

For this achievement we will need rewrite the `ddp-server/server_convenience.js` package.

```js
if (process.env.DDP_DEFAULT_CONNECTION_URL) {
  __meteor_runtime_config__.DDP_DEFAULT_CONNECTION_URL =
    process.env.DDP_DEFAULT_CONNECTION_URL;
}

Meteor.server = new Server;

Meteor.refresh = function (notification) {
  DDPServer._InvalidationCrossbar.fire(notification);
};

// Proxy the public methods of Meteor.server so they can
// be called directly on Meteor.
_.each(['publish', 'methods', 'call', 'apply', 'onConnection', 'onMessage'],
       function (name) {
         Meteor[name] = _.bind(Meteor.server[name], Meteor.server);
       });

// Meteor.server used to be called Meteor.default_server. Provide
// backcompat as a courtesy even though it was never documented.
// XXX COMPAT WITH 0.6.4
Meteor.default_server = Meteor.server;
```

I'm think the function for adding new servers is dangerous, because there is may be side effects in update the `Meteor.server` variable on the fly.

alexhultman:
You're not going to get anything more out of `sc-uws` than from `uws`. It's just an identical copy except for the criticism of SocketCluster removed (read: censored).

Besides, that sc-uws is very obviously in conflict with Zlib's clause 2:

`2. Altered source versions must be plainly marked as such, and must not be
misrepresented as being the original software.`

This of course puts it at risk of being DMCA:d any moment. (and no, intellectual property laws don't give a shit about NPM policies). Besides, why would you think that `sc-uws`, an identical copy, would be any less deprecated than `uws`? Just because some asshole uploads identical binaries to some other NPM name doesn't make it any more sustainable in any way.

I would take a breather and keep using `uws` until I have published next version somewhere.

benjamn:
I think we're happy to wait for an official release. As an open-source maintainer myself, I am very wary of using forks that don't have the backing of the original author, unless the new maintainers have really embraced the future maintenance burden, ideally with the original author's blessing.

packages/stream-server-uws/stream_server_uws.js
mrauhu: @benjamn I found difference between how works the `require.resolve()` in Node.js and Meteor.

So, your code not working if module is not installed.

## Code

```js
let path;
try {
  path = require.resolve('uws');
} catch (e) {
  console.error(e);
}
console.log('path', path);
```

## Output if installed

```
path /node_modules/uws/uws.js
```

## Output if not installed

### Node.js

```
{ [Error: Cannot find module 'uws'] code: 'MODULE_NOT_FOUND', name: 'Error' }
path undefined
```

### Meteor

```
path uws
```

## Explanation

In Meteor `require.resolve()` was patched in the file `modules-runtime.js`:

```js
  function makeRequire(file) {
    var module = file.module;

    function require(id) {
      return module.require(id);
    }

    require.extensions = fileGetExtensions(file).slice(0);

    require.resolve = function resolve(id) {
      return module.resolve(id);
    };

    return require;
  }
```

We get patched version of `require()` if using packages `modules` or `ecmascript`.

KoenLav:
@benjamn while I completely agree with you on that point I do think the efforts in this PR (and others opened by @mrauhu), which focus on allowing end users to switch out the stream server implementation, are worthwhile!

What do you think?

packages/stream-server/README.md
mrauhu: > In this example, let's recommend explicitly importing StreamServers:
>
> ```js
> import { StreamServers } from "meteor/stream-server";
>```

@benjamn this part is done.

yorrd:
Is this going to be the default behavior in Meteor? Looks amazing!

mrauhu:
@benjamn I made multiple changes after your review, please make re-review.

## Changes

* Integrated `stream-server-uws`  into the Meteor as weak dependency;
* Using NPM dependency `uws@10.148.0` that is original work of Alex Hultman;
* Removed `underscore` from `stream-server` package.

## Decisions

### Using global `StreamServerUWS` variable

`stream_server_uws.js`

```js
StreamServerUWS = class StreamServerUWS {
  // ...
}
```

Because otherwise the `StreamServerUWS` variable can't be used like that:

`some-other-package/server-side.js`
```js
// StreamServerUWS is Function
import { StreamServerUWS } from 'meteor/stream-server-uws';

class AmazingStreamServer extends StreamServerUWS {
  // ...
}
```

I tried use ES2015+ way:

```js
export class StreamServerUWS {
   // ...
}
```

But the Meteor package compiler tool convert plain ES2015+ code into this:

```js
var StreamServerUWS;
// ...
module.export({
  StreamServerUWS: () => StreamServerUWS
});
// ...
class StreamServerUWS {
  // ...
}
```

So you got `undefined`:

```js
// StreamServerUWS is undefined
import { StreamServerUWS } from 'meteor/stream-server-uws';
// app is broken
```

### Don't change the Meteor core

I left the `stream-server/README.md` unchanged, because if a developer decided to write the custom stream server, he knows what he is doing.

And to add to the Meteor the support of such things as:
* import packages with wildcard names, like `stream-server-*`;
* deferred DDP server load with re-initialization;
* change packages load order;

a lot of changes need to be made.

My solution with the `StreamServers` array is simple, but not a silver bullet.

KoenLav:
The original author has plans of supporting Node.js bindings again;

https://github.com/uNetworking/uWebSockets-node/issues/2

mrauhu:
Created post on Meteor's forums:

> Try µWebSockets support in the Meteor. Pure speed, no memory leaks
https://forums.meteor.com/t/try-websockets-support-in-the-meteor-pure-speed-no-memory-leaks/45678

mahadoang:
Error then use 

```
Meteor.loginWithPassword('xxxx@xxx.com','xxxx',function(err){
      console.log('error',err);
});
```

```
I20180926-11:33:17.588(7)? Exception while invoking method 'login' TypeError: DiffSequence.diffMaps is not a function
I20180926-11:33:17.588(7)?     at Session._diffCollectionViews (packages/ddp-server/livedata_server.js:762:18)
I20180926-11:33:17.589(7)?     at packages/ddp-server/livedata_server.js:842:12
I20180926-11:33:17.589(7)?     at Object.Meteor._noYieldsAllowed (packages/meteor.js:848:12)
I20180926-11:33:17.590(7)?     at Session._setUserId (packages/ddp-server/livedata_server.js:840:12)
I20180926-11:33:17.590(7)?     at MethodInvocation.setUserId [as _setUserId] (packages/ddp-server/livedata_server.js:677:14)
I20180926-11:33:17.591(7)?     at MethodInvocation.setUserId (packages/ddp-common/method_invocation.js:92:10)
I20180926-11:33:17.592(7)?     at AccountsServer.Ap._loginUser (packages/accounts-base/accounts_server.js:306:20)
I20180926-11:33:17.593(7)?     at AccountsServer.Ap._attemptLogin (packages/accounts-base/accounts_server.js:360:12)
I20180926-11:33:17.593(7)?     at MethodInvocation.methods.login (packages/accounts-base/accounts_server.js:545:21)
I20180926-11:33:17.593(7)?     at currentArgumentChecker.withValue (packages/check/match.js:118:15)
I20180926-11:33:17.593(7)?     at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1304:12)
I20180926-11:33:17.594(7)?     at Object._failIfArgumentsAreNotAllChecked (packages/check/match.js:116:43)
I20180926-11:33:17.613(7)?     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1780:18)
I20180926-11:33:17.613(7)?     at DDP._CurrentMethodInvocation.withValue (packages/ddp-server/livedata_server.js:719:19)
I20180926-11:33:17.614(7)?     at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1304:12)
I20180926-11:33:17.614(7)?     at DDPServer._CurrentWriteFence.withValue (packages/ddp-server/livedata_server.js:717:46)
I20180926-11:33:17.614(7)?     at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1304:12)
I20180926-11:33:17.615(7)?     at Promise (packages/ddp-server/livedata_server.js:715:46)
I20180926-11:33:17.615(7)?     at new Promise (<anonymous>)
I20180926-11:33:17.615(7)?     at Session.method (packages/ddp-server/livedata_server.js:689:23)
I20180926-11:33:17.616(7)?     at packages/ddp-server/livedata_server.js:559:43
```

mrauhu:
> Error then use
> ```js
> Meteor.loginWithPassword(‘xxxx@xxx.com’,‘xxxx’,function(err){
>   console.log(‘error’,err);
> });
> ```

@mahadoang, I had tried to reproduce your bug, but unfortunately all works.

## How to check `accounts-password`

### Prerequisites

* Meteor 1.7
* Git

### Steps

1. Create the Meteor project via:

    ```bash
    meteor create --full meteor-ddp-test
    cd meteor-ddp-test
    ```

2. Add the Meteor Accounts Password login and the Accounts UI packages:

    ```bash
    meteor add accounts-password accounts-ui
    ```

2. Install the custom `ddp-server` and `stream-server-uws` using [guide in the first post](https://forums.meteor.com/t/try-websockets-support-in-the-meteor-pure-speed-no-memory-leaks/45678/1).

3. Create file `imports/startup/server/users.js`:

    ```js
    Meteor.startup(() => {
      const Users = [];

      const users_roles = new Map([
        ['user', 'user'],
      ]);

      for (const [name, role] of users_roles) {
        Users.push({
          username: name,
          password: name,
          email: `${name}@example.com`,
          profile: { name },
          roles: [role],
        });
      }

      if (Meteor.users.find().count() > 0) {
        return;
      }

      for (const user of Users) {
        const id = Accounts.createUser(user);

        Meteor.users.update(id, { $set: { 'emails.0.verified': true } });
        /*
        if (typeof user.roles !== 'undefined'
          && user.roles.length > 0) {
          Roles.addUsersToRoles(id, user.roles);
        }
        */
        console.log(`Created user ${user.username}, id: `, id);
      }
    });
    ```

4. Update `imports/startup/server/index.js`, add line:

    ```js
    import './users.js';
    ```

5. Update `imports/ui/pages/home/home.js`, add lines:

    ```js
    Template.App_home.events({
      'click #login'(event) {
        Meteor.loginWithPassword(
          'user@example.com',
          'user',
          (error, result) => {
            if (error) {
              console.error('loginWithPassword()- error', error);
              return;
            }
            console.log('loginWithPassword()- result', result);
          }
        );
      }
    });
    ```

6. Update `imports/ui/pages/home/home.html`:

    ```html
    <template name="App_home">
      {{> hello}}
      {{> info}}
    
      {{! Check login }}
      {{! via `accounts-ui` package }}
      {{> loginButtons }}
      {{! via `Meteor.loginWithPassword()` }}
      {{#unless currentUser }}
        <button id="login">Login</button>
      {{/unless}}
    </template>
    ```

7. Run Meteor:

    ```bash
    meteor
    ```

Now on the http://localhost:3000 you can login via:

* `accounts-ui` Sign-in drop-down;
* `[Login]` button via `Meteor.loginWithPassword()`.

I hope this guide helps you.

mrauhu:
@mahadoang thanks for found the bug that affected old versions of Meteor before 1.7.1.

Problem was introduced in:

> ddp-server: replace usage of Object with Map & Set
> https://github.com/meteor/meteor/commit/79ae1849f5846e207ed4dfa49831ba409252fd03

I'm made [patch](https://github.com/mrauhu/ddp-server/commit/758cc67f6129f13588a4a76ea45a5f3b2a9b5016) for `ddp-server` if your Meteor version >= 1.5.0 and < 1.7.1.

## Installing patch

### Meteor >= 1.5.0 and  < 1.7.1

Update the patched `ddp-server` to the latest version:

```bash
git submodule update --remote packages/ddp-server
```

### Meteor >= 1.7.1

Checkout the commit [`b912a48`](https://github.com/mrauhu/ddp-server/commit/b912a488495205cc5b170d53aba94f98585b0cd4):

```bash
cd packages/ddp-server
git checkout b912a48
cd ../..
```

Also I update the guide [How to try it now](https://github.com/meteor/meteor/pull/10120#issue-205605208).

I don't changed this pull request, because it for Meteor versions >= 1.7.1.

KoenLav:
@mrauhu the original author of uWebSockets provided for a way to install via Npm again.

@benjamn could you take another look? Also the changes in the other PR (making the we socket server swappable) seem worthwhile regardless of this PR.

alexhultman:
The new project is here: https://github.com/uNetworking/uWebSockets.js

However I think this will fit in very badly. It is a replacement of **both** Http and WebSockets - you cannot have Node.js http server and "uws" - you need to swap both out, and they don't follow any particular Node.js style.

µWebSockets.js is not "compatible" with any Node.js project, so any integration would have to be very significant.

You can find examples here: https://github.com/uNetworking/uWebSockets.js/tree/master/examples

KoenLav:
@alexhultman thanks!

From the WebSockets example I gather it could fit in quite nicely, why do you think it wouldn't? 

Why would we need to swap out both Node.js HTTP and WS?

As far as I can tell simply requiring WS from your package would do the trick, or is there some underlying issue I'm not seeing?

alexhultman:
It is not that simple. Prior versions of "uws" built on Node.js Http server and only took care of WebSockets. New version of "uws" (I guess it is called µWebSockets.js now) is a **complete** replacement that has Http and WebSockets completely separate from anything Node.js. So you cannot have one server port listen to Node.js Http and also listen to WebSockets from "uws". One port = entirely "uws" or entirely Node.js.

KoenLav:
I see, would it be conceivable to use uWebSockets as the "front" and pass/proxy any HTTP calls to Node.js HTTP (for the time being, at least).

This would be slightly detrimental to HTTP performance, I guess, but WebSockets connections would benefit significantly.

I get your point though; the scope of this issue has increased significantly. 

It's a shame because we would have loved to see this technology enabled in Meteor in some way; are there any (other) ways of integrating which could limit the scope again?

alexhultman:
> I see, would it be conceivable to use uWebSockets as the "front" and pass/proxy any HTTP calls to Node.js HTTP (for the time being, at least)?

Maybe, but I'm not working on that. The point with µWS.js is to provide Http, WebSockets, pub/sub all as a stand-alone solution with very good perf. that you don't need to wrap up in other projects. So it would be against my goal to try and shoehorn it in as some subsystem in some other wrapper project.

KoenLav:
@mrauhu I think, judging from the description above, this PR is pretty much dead (as in it probably won't be accepted by Meteor).

What do you think?

filipenevola:
The CLA is still not signed here, because of that I'm closing this PR. Feel free to continue the work in a new PR as we can't merge this one without the CLA.

mrauhu:
Greetings, @filipenevola 

There is no problem with CLA signing.

Unfortunately, I did not receive a response to the changes that I made to @benjamn request.

The MDG has ignored this request for more than a year.

evolross:
I'm experiencing the limits of DDP performance with thousands of users... is there some way to test/use this PR without it being merged?

Anyone actually using this in production? On Meteor 1.8.x - 1.10.x?

mrauhu:
@evolross You can try it using the guide from first post https://github.com/meteor/meteor/pull/10120#issue-205605208

filipenevola:
Hi @mrauhu the CLA was not ok when I closed this PR but it's ok now! Great. 

Could you resolve the conflict here https://github.com/meteor/meteor/pull/10118? So we could continue this work.

mrauhu:
@filipenevola thanks.

I checked this pull request and found that the `uws` package incompatible with Meteor `>= 1.9` and Node.js 12.
I think I will use the newest `uWebSockets.js` package for resolving conflicts.

This will lead to additional changes, as the `uWebSockets.js` package has built-in HTTP server, without using the external Node.js `http` server.

mrauhu:
# `stream-server-uws` update

Added support  for the Meteor version `>= 1.9`.

## Changes

* Risen from the grave https://github.com/meteor/meteor/pull/10120#issuecomment-481880020, @KoenLav.
* Used the `@clusterws/cws` fork from @goriunov, it is approved https://github.com/ClusterWS/cWS/issues/5#issuecomment-431434814 by @alexhultman the original author of the `uws` as requested https://github.com/meteor/meteor/pull/10120#issuecomment-410375562 by @benjamn.
* Rebased parent https://github.com/meteor/meteor/pull/10118 this pull-request on top of the `devel` branch, and fixed conflict with `packages/ddp-server/stream_server.js` as requested https://github.com/meteor/meteor/pull/10120#issuecomment-606233874 by @filipenevola.
* Created the `disable-sockjs` internal Meteor package  for the weak dependency checking in `socket-stream-client` and `ddp-server` packages.
* Updated package `socket-stream-client`, @evolross no need pinning the version, you can check it out on production https://github.com/meteor/meteor/pull/10120#issuecomment-605823859.
* Fixed the test `stream - /websocket is a websocket endpoint` in `socket-stream-client/client-tests.js`. 
Use command for running tests for all packages, SockJS will be automatically disabled:
  ```
  meteor test-packages --extra-packages=stream-server-uws
  ```

## Documentation

Updated guide in the first post https://github.com/meteor/meteor/pull/10120#issue-205605208 and created repository with instructions:

> ### ☄ Meteor + 🚀 µWebSockets
> https://github.com/mrauhu/meteor-uwebsockets

alexhultman-2fa:
> Used the @clusterws/cws fork from @goriunov, it is approved ClusterWS/cWS#5 (comment) by @alexhultman the original author of the uws as requested #10120 (comment) by @benjamn.

I haven't approved anything at all. You're linking to a thumbs up that's been pulled entirely out of context.

This "cWS" is nothing more than rebranded bit rot, it is to no degree maintained whatsoever.

I do absolutely not, to any degree, approve of "cWS" - it is a dumpster fire if anything.

mrauhu:
@alexhultman thank you for clarification.

Then I will add support of the [μWebSockets.js][u] in another way by changing the `webapp` package.

[u]: https://github.com/uNetworking/uWebSockets.js

alexhultman-2fa:
I'm not saying you have to do this or that, you can do whatever you want to. But you cannot say that I approve things I don't.

If you use forks (which I strongly distance myself from), it is essential to not use any branding or trademarks I use for my company's products.

You have to respect copyright laws, trademark laws. You cannot go around saying this shitty fork is "uWebSockets" or similar if it isn't the original thing.

The license clearly states modified versions must not be presented or confused with the original. A fork is to be an entirely separate project with entirely separate branding.

Just because something is "open source" doesn't mean all intellectual property laws just drop dead on the floor.

The way you present things currently is simply not legal, and harms my branding.

vlasky:
@alexhultman-2fa you have used some harsh language like "rebranded bit rot" and "dumpster fire" to describe this fork derived from a previous release of your μWebSockets.js code.

Perhaps we can turn this discussion into something constructive.

The fact is, Meteor could really benefit from having a better performing, resource-efficient high quality WebSocket library.

Are you open to helping make this happen? If so, what would you need?

alexhultman-2fa:
Hey, you Meteor developers are free to do whatever you want to, I have no say in that.

I do however have a legal say when it comes to misrepresenting altered versions of my work, as being the original.

This is clearly stated in the license grant.

When you print "Meteor + μWebSockets" when in fact it is cWS, a modified work with key elements cut out and deprecation warnings removed and ignored, this is not okay. I run a business and will DMCA anything that's disregarding my licenses, esp. something as popular as Meteor.

Make it "Meteor + cWS" in that case. I have nothing against that.

mrauhu:
Fixed misspell in guide that lead to `ReferenceError: **Tinytest is not defined**`

packages/socket-stream-client/browser-legacy.js
```diff
@@ -0,0 +1,5 @@
+// Don't import SockJS for legacy or unknown browsers that supports WebSockets
+if (!(Package['disable-sockjs']
+  && typeof WebSocket !== 'undefined')) {
+  import("./sockjs-0.3.4.js");
```
coagmano, line 4: You'll need to add `dynamic-import` as a dependency in `package.js` again.
It was removed recently so `unsafe-eval` could be disabled via CSP for folks that needed to

mrauhu: @coagmano thanks for watching.

The `ecmascript` package implements the `dynamic-import`, see:

`packages/ecmascript/package.js`

```js
  // Runtime support for Meteor 1.5 dynamic import(...) syntax.
  api.imply('dynamic-import');
```

coagmano: Oooh, I didn't know that.
That means [this PR of mine](https://github.com/meteor/meteor/pull/10741/) did nothing 👀

mrauhu:
Added support for Meteor version 1.11.1.

afrokick:
@filipenevola what you think about a future of this PR?

ignl:
Can this be included into version 2 scope? Seems like a perfect change for major version increase.

afrokick:
@filipenevola could you please say something about this PR?

Is it has a chance to be merged?

@mrauhu did a lot of works and doesn’t have a response from an official members about 2 years.

Such great improvement and no reaction...

alexhultman:
@afrokick Just because one refers to 3 year old unpatched security issues as "a fork" with a new logo slapped on, doesn't make it "new" or "maintained". My point is this PR is about merging code with known security-, stability-issues into Meteor. This "fork" is essentially left frozen in time and has not fixed the handful of issues the upstream project has fixed during these 3 years.

So yeah, when I'm ready to put CVE's on these issues you're going to have big issues since this "fork" is nothing but glorified bit rot. Look at the Git history yourself - the core is unmodified for 3 years.

alexhultman:
Can you please refer to the correct project here? This PR does not add support for "µWebSockets" - it adds support for "ClusterWS" which is an unmaintained "fork" with very minor changes from 3 years ago (biggest changes are their name and their logo and their removal of SSL support). You can scientifically research this yourself by looking at their Git history yourself. Almost all files in src (which is the actual code) is entirely unmodified for several years. I know at least a dozen ways of crashing this old code in an instant.

afrokick:
@alexhultman thank you for your vision. I want to hear it from meteor's core team, what they think about this PR.

As a solution, we can release it as a separate package, like redis-oplog. But still we need to modify some core packages to have a way to extends/replace stream-server.

vlasky:
@alexhultman if there are problems as serious as you describe, you should have published the CVEs when you uncovered these issues over the last 3 years, to help ensure the safety and security of those who have not updated or have forked your past UWS code.

Please publish these CVEs without delay and the community will take things from there, with or without your further cooperation.

alexhultman:
Instead of thumbs-down my post you can do some research and check if what I say is true or not. Here is a snap I did today - as you can evidently see - this is bit rot. They haven't changed anything other than minor legal things to be a proper fork. I wrote this code, so I know it intimately and I know several security issues in it.

bit rot

goriunov:
FIY @alexhultman cws repository was forked when you have completely removed uWebSocket and tried to kill it from npm by upgrading minor version with empty data. After that new uWebSocket version has not been released for quite a while and old one did not have support for newer versions of Node at that time. So cws repo was created to temporarily mitigate those issues. There was never a point to fully replace uWebSocket but i (and probably others) dont agree with just killing off your project and many other's people work just because you are focusing on new version... After new version of uwebsocket has been release readme of CWS has been update to remind people to use `latest uWebSockets version instead`.

alexhultman:
> Please publish these CVEs without delay

I haven't exactly reduced the issues down to specific instructions that can be easily reproducible - I only know there are several issues in this old code such as crashes. If they can be exploited or not I don't know. That's why I tell people not to use bit rot, but people rather down-vote what I say and keep using my old shit than to accept they are using nothing but re-branded crap I deprecated 3 years ago for having a poor code base. I have said this a million times - I have written posts on Medium about it - and people never listen. I have a big fat bold text on NPM telling people to stop using it and I have a deprecation message. I cannot do more than this.

So at least Meteor should refrain from using it. That's my opinion.

goriunov:
I do agree with @alexhultman to do not update to cws and consider to use latest uWebSockets due to the reason explained above. However i am also personally hesitant on using uWebSockets due to the previous experience.

alexhultman:
> So cws repo was created to temporarily mitigate those issues.

Yes you "mitigated" that issue by removing the SSL support because you lacked the competence to upgrade it to the next version of Node.js. Maybe it is time to remove that project now? You're clearly not maintaining it and the upstream project has continued with 4 new major versions that are actually maintained and secure.

afrokick:
> uWebSockets for Node seems to be no longer actively maintained because the maintainer is throwing a fuzz on NPM?

So now uWebSockets seems like ok.

@mrauhu what you think about it?

harrisonhunter:
Outsider to this conversation looking in - it seems like two things are true:

1) Using the underlying µWebSockets technology would be a big performance win for Meteor apps if done correctly. Also did some testing myself and this seems true and a worthwhile improvement.

2) Using the cWS version referenced in this PR is not a good idea. It is poorly maintained and has known issues.

To get this into the main repo, it seems like using µws directly or a different, more recent, better maintained fork (if it exists) will be necessary. Think should focus the conversation on how to do that to get this across the finish line, otherwise it will stay in limbo.

alexhultman:
Thanks for changing the name of this PR 👍

alexhultman:
I think it will remain in limbo because there are two issues here - a) ClusterWS is bit-rot, and b) µWebSockets is not compatible with Node.js in the same sense as the old code was. You cannot share a port between Node.js and uWS - you either have to do all uWS or no uWS. This because the new code includes its own HTTP server - it does no longer rely on the one found in Node.js. That's also why it is so much faster than Node.js, and the old version.

So you either go all in, or stay out. I don't think this will work since Meteor probably assume a lot about HTTP such as it being hosted by Node.js and not uWS. I don't know. All I know is ClusterWS is unmaintained bit-rot.

copleykj:
Just as a reminder, let's keep things civil and treat each other with as much respect as possible.

vlasky:
@alexhultman that is an informal logical fallacy known as a [false dilemma](https://en.m.wikipedia.org/wiki/False_dilemma).

For instance, any critical issues in ClusterWS could be fixed by the current maintainer or other contributors. 

Perhaps it won't perform as well as your latest UWS, but it would still be a significant, meaningful improvement for everyone in the Meteor community.

alexhultman:
It can definitely be seen as a false dilemma - but looking at history the **probability** of such an outcome is very **unlikely**.

filipenevola:
Hi, I believe the goal of everybody investing time in this issue is to make Meteor WebSocket layer faster.

I don't understand why people are so angry and discussing things that are not important for this matter.

Let's keep this issue focused in the problem. 

---

First, we need to provide a nice way for current Meteor users to keep using what they have today, I already scaled many Meteor apps with the current setup. 

Second, we should start to play with alternatives, these alternatives don't need to be part of the core, but they could, after some tests and usage in production.

So, in order to achieve the first goal we need to decouple the current WebSocket implementation from the core packages keeping the current functionality.

Is #10118 enough to achieve the first goal? I can review it in details soon if we don't know this yet. 

Depending how far we are from the first goal we could include it on 2.0 or not. 

I hope I have addressed here the main concerns, otherwise feel free to ask again below. 

---

As always, let's keep this discussion at the technical level and without personal attacks, otherwise I will need to take some actions to censor inappropriate discussions. I really don't want to do that.

filipenevola:
Hi, please share your thoughts here as well https://github.com/meteor/meteor/pull/10118#issuecomment-718906419

We need to solve our first goal here and this PR #10118 is the starting point.

OmgImAlexis:
This PR might as well be closed since ClusterWS is deprecated.

https://github.com/ClusterWS/ClusterWS

maxpain:
Which options I have today to integrate latest Meteor with latest uWebSockets?

PR #10132, opened 2018-08-06
Fix typo in import-scanner

chrishearn:
Small fix for typo that caused an error on build tool startup when we tested our app against 1.7.1-beta.19


benjamn:
Thanks @chrishearn! This has been released in both Meteor 1.7.0.4 and 1.7.1-beta.29.

Issue #10126, opened 2018-08-04
[1.7.1-beta.25] Error: Cannot find module '@babel/runtime/helpers/builtin/interopRequireDefault


rj-david:
When I first run beta.25, I got this error message:

```
Unable to resolve some modules:

  "@babel/runtime/helpers/builtin/interopRequireDefault" in /home/rj/Projects/Bountee/imports/ui/navigations/AppNavigator.jsx (web.browser)
  "@babel/runtime/helpers/builtin/extends" in /home/rj/Projects/Bountee/imports/ui/navigations/AppNavigator.jsx (web.browser)

If you notice problems related to these missing modules, consider running:

  meteor npm install --save @babel/runtime
```

My package.json had this after running the package install command above

```
"dependencies": {
        "@babel/runtime": "^7.0.0-beta.56",
```

Then I got this error:

```
W20180804-18:50:12.296(8)? (STDERR) /home/rj/Projects/Bountee/.meteor/local/build/programs/server/boot.js:475
W20180804-18:50:12.331(8)? (STDERR) }).run();
W20180804-18:50:12.332(8)? (STDERR)    ^
W20180804-18:50:12.332(8)? (STDERR)
W20180804-18:50:12.332(8)? (STDERR) Error: Cannot find module '@babel/runtime/helpers/builtin/interopRequireDefault'
W20180804-18:50:12.333(8)? (STDERR)     at Function.Module._resolveFilename (module.js:547:15)
W20180804-18:50:12.333(8)? (STDERR)     at Function.resolve (internal/module.js:18:19)
W20180804-18:50:12.333(8)? (STDERR)     at Object.require (/home/rj/Projects/Bountee/.meteor/local/build/programs/server/boot.js:288:32)
W20180804-18:50:12.334(8)? (STDERR)     at makeInstallerOptions.fallback (packages/modules-runtime.js:604:18)
W20180804-18:50:12.335(8)? (STDERR)     at Module.require (packages/modules-runtime.js:230:14)
W20180804-18:50:12.336(8)? (STDERR)     at require (packages/modules-runtime.js:244:21)
W20180804-18:50:12.336(8)? (STDERR)     at livedata_connection.js (/home/rj/Projects/Bountee/.meteor/local/build/programs/server/packages/ddp-client.js:144:30)
W20180804-18:50:12.337(8)? (STDERR)     at fileEvaluate (packages/modules-runtime.js:322:7)
[...]

PR #10134, opened 2018-08-06
Release 1.7.0.4

benjamn:
Quick patch release to address `@babel/runtime` issues such as https://github.com/meteor/meteor/issues/10126, introduced by the removal of the `@babel/runtime/helpers/builtin` directory in https://github.com/babel/babel/pull/8266 (which, to be clear, is a valuable change in the long term, even though it's somewhat annoying in the short term).

Plan:
- [x] Make the Meteor `babel-runtime@1.2.3` package complain if you install `@babel/runtime@7.0.0-beta.56` or higher in your application's `node_modules` directory.
- [x] Make sure new Meteor 1.7.0.4 apps install `@babel/runtime@7.0.0-beta.55` and nothing later.
- [x] Publish Meteor 1.7.0.4 as a patch release, so that everyone using Meteor 1.7.0.x sees an update banner.
- [ ] Bump the minor version of `babel-runtime` (to 1.3.0) on the [`release-1.7.1`](https://github.com/meteor/meteor/pull/9942) branch, and undo the changes above.


tools/tests/apps/dynamic-import/.meteor/release
```diff
@@ -1 +1 @@
-METEOR@1.6.1
+METEOR@1.7.0.3
```
abernix: If I recall, this (as in the `dynamic-import` test app's Meteor version) is intentionally held back until the version is officially released, correct?

That is to say, this is intentionally not 1.7.0.4 yet.

benjamn: That's right. I will update it again once the final release commit is merged into `devel`.

PR #10149, opened 2018-08-13
Create a new react project with meteor create --react

dmihal:
I've added a skeleton for a Meteor React project that can be created with `meteor create --react`.

I aimed to replicate the application created by default (`meteor create`) in React. However I've taken a bit of inspiration from `skel-full` and stored the list of links in a collection so that `withTracker` from `react-meteor-data` can be demonstrated.

See discussion in meteor/meteor-feature-requests#182


benjamn:
Meteor is long overdue for an easy way to create React apps, and this is a solid starting point.

I have merged this commit, with some adjustments, into https://github.com/meteor/meteor/pull/9942, and it will become available in the next release candidate of Meteor 1.7.1 (aka 1.8: https://github.com/meteor/meteor/pull/9942#issuecomment-420744917).

At the moment, the implementation of the various `meteor create ...` application templates is tied to the Meteor release, so we will need to put out a new Meteor release to make any changes to this code. That's not ideal, but I believe we can develop a system in the future that allows for iterating on the starter code in a separate repository, and `meteor create` can consult (an offline clone of) this repository when creating new apps.

PR #10152, opened 2018-08-14
Parse URL query parameters before it's destroyed

gchallen:
The inline middleware that strips the ROOT_URL_PATH_PREFIX destroys req.url. If this happens before the query parameters are parsed by use(query()) the query parameters are always missing. This causes OAuth authentication to fail when ROOT_URL_PATH_PREFIX is set.

A better fix would be to fix the path stripping logic so that it preserves the URL parameters, since other things may try to rely on them later. But this works for now.

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -749,6 +749,13 @@ function runWebAppServer() {
     res.end();
   });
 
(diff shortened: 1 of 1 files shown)


gchallen:
CA signed.

benjamn:
Thanks @gchallen!

PR #10155, opened 2018-08-16
CoffeeScript 2.3.1, support for setMinimumBrowserVersions

GeoffreyBooth:
@benjamn All I needed to do to add support for `setMinimumBrowserVersions`/the `modern-browsers` package was update the versions of the Meteor core packages that the `coffeescript` package was depending on. Once `coffeescript` was using the latest `ecmascript`, `caching-compiler` and so on, `setMinimumBrowserVersions` just started working.

While I was at it, I updated to the latest CoffeeScript and added some missing `import` statements and did some other general cleanup.

I can’t really find a good test for `setMinimumBrowserVersions` in the codebase; from what I can tell, my browser always gets the modern bundle when I load the tests app, whether or not I call `setMinimumBrowserVersions`. Please let me know if there’s some way to make this test more robust, or if I should take it out.

I tested in a real app created via `meteor create coffeetest` with the `.js` files converted into CoffeeScript and me adding a class and ensuring that the JS bundle output by Meteor preserved the `class` keyword. It does (using these new versions of the packages).

### DIFF
--- a/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
+++ b/packages/non-core/coffeescript-compiler/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "coffeescript": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/coffeescript/-/coffeescript-2.2.1.tgz",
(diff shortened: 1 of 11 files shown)


GeoffreyBooth:
Hmmm. The tests passed for me locally. I’ll investigate.

How are you handling `setMinimumBrowserVersions` with the test app? It would seem to me that if any test/package calls that, it affects all others.

Also `setMinimumBrowserVersions` and `modern-browsers` need to be added to the docs, they appear there only in the changelog right now. It took me some thinking before I realized that `setMinimumBrowserVersions` needed to be called in server-only code.

GeoffreyBooth:
@benjamn Looks like https://github.com/meteor/meteor/commit/b8b404daf37717442afe72916e806bf909d5dd60 fixed the build.

PR #10159, opened 2018-08-16
Release 1.6.1.4

benjamn:
Quick patch release for the Meteor 1.6.1.x version line to update Node to version [8.11.4](https://nodejs.org/en/blog/release/v8.11.4/), an important [security release](https://nodejs.org/en/blog/vulnerability/august-2018-security-releases/).


evolross:
Just an FYI - there's no mention of this 1.6.1.4 release in the official [changelog](https://docs.meteor.com/changelog.html) 

This was confusing when recently upgrading. Especially since I had issues minifying my app using this specific release in production and could never get my app to deploy (or build locally using `--production`). Not sure it's worth reporting since `1.7.x` and `1.8.x` worked as expected.

PR #10184, opened 2018-08-28
Remove duplicate `mongodb` package update entry from history

klaussner:
The `mongodb` package was updated in two PRs (#10058 and #10101) and both included a history entry (see the lines above the ones I removed).


abernix:
Thanks, @klaussner!

Issue #9973, opened 2018-06-10
Android Run Fails First Time with Network Error


theavijitsarkar:
Meteor  1.7
Android 7.1
Mac OSX 10.13
Cordova 8.0

I am trying to run the android app locally. I start it with
```
meteor run android-device
```
But it gives me an error and screen goes black
```
There was a network error.
(http://localhist:12933/)
```
This error goes away if I run it once again.

Meteor Log
```
=> Started app on Android Device.             
I20180610-21:44:32.002(5.5)? 06-10 21:41:55.262 27144 27144 I CordovaLog: Changing log level to DEBUG(3)
I20180610-21:44:32.005(5.5)? 06-10 21:41:55.364 27144 27144 I chromium: [INFO:library_loader_hooks.cc(36)] Chromium logging enabled: level = 0, default verbosity = 0
I20180610-21:44:32.006(5.5)? 06-10 21:41:55.621 27144 27199 E chromium: [ERROR:devtools_http_handler.cc(292)] Cannot start http server for devtools. Stop devtools.
I20180610-21:44:32.006(5.5)? 06-10 21:41:55.793 27144 27175 W SystemWebViewClient: URL blocked by whitelist: http://localhost:12944/
I20180610-21:44:32.007(5.5)? 06-10 21:41:55.847 27144 27144 I MeteorWebApp: Serving asset bundle with version: effc7bd8604fcc566f7ef7a174626889d29bca31
I20180610-21:44:32.007(5.5)? 06-10 21:42:15.863 27144 27232 W MeteorWebApp: App startup timed out, reverting to last known good version


```

Android Logcat dump
```
                            Process com.avifa.regularli created for activity co
                            m.avifa.regularli/.MainActivity
                            PID: 27144   UID:    GIDs: 

                    art  I  Late-enabling -Xcheck:jni
                 System  W  ClassLoader referenced unknown path: /data/app/com.a
                            vifa.regularli-1/lib/arm
             CordovaLog  I  Changing log level to DEBUG(3)
        CordovaActivity  I  Apache Cordova native platform version 6.4.0 is star
                            ting
                         D  CordovaActivity.onCreate()
         WebViewFactory  I  Loading com.android.chrome version 67.0.3396.81 (cod
                            e 339608102)
       cr_LibraryLoader  I  Time to load native libraries: 8 ms (timestamps 6860
                            -6868)
               chromium  I  [INFO:library_loader_hooks.cc(36)] Chromium logging 
[...]

PR #10193, opened 2018-09-04
Cordova plugin handling algorithm consistency improvements

wojtkowiak:
This strives to fix several issues with cordova integration plugins handling:

Fixes several cases causing cordova plugins reinstall on every build:
- proper handling of scoped npm cordova plugins (adding one would cause reinstall on every build)
- proper detection of plugin removal (previously a cordova plugin containing a dependency would make the algorithm think a package was removed from cordova-plugins)
- proper handling of plugins which have plugin.xml id different than the npm package name

Additionally:
- rechecks the build integrity verifying if packages were really installed and perform a retry if needed.
- allows to override a meteor package cordova dependency with scoped package i.e. @scope/cordova-dummy-plugin will now override a cordova-dummy-plugin dependency. (now in this case the same plugin will be installed twice, cordova will leave the last one installed)
- fixes `<dependency id="es6-promise-plugin" url="https://github.com/vstirbu/PromisesPlugin.git"/>` being translated by `listFetchedPluginVersions` as `https://github.com/vstirbu/PromisesPlugin.git#undefined`

The main benefit is that the full plugins uninstall and reinstall will only happen on plugin being updated, removed or added. Now for most projects it happens every build even though cordova-plugins file did not change. Additionally because cordova_lib is not handling properly the promises from plugin remove a situation is happening when plugins are missing - this adds a re-check of plugins state and retries to add missing plugins.

Fixes: https://github.com/meteor/meteor/issues/9548 https://github.com/meteor/meteor/issues/9973

### DIFF
--- a/tools/cordova/project.js
+++ b/tools/cordova/project.js
@@ -658,8 +658,55 @@ mobile-config.js accordingly.`);
             end: pluginsToInstallCount
           });
         });
+
+        logIfVerbose('Validating Cordova plugins installation');
+        this.ensurePluginsWereInstalled(pluginVersionsToInstall, pluginsConfiguration, true);
       }
     });
+  }
+
+  // Ensures that the Cordova plugins are installed
+  ensurePluginsWereInstalled(requiredPlugins, pluginsConfiguration, tryInstall) {
+    // List of all installed plugins. This should work for global / local / private cordova plugins.
+    // Examples:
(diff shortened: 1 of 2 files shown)


wojtkowiak:
@benjamn thanks a lot for merging 🙂 
Next week I will add another PR to with more/better comments to better explain the difference between the plugin name from plugin.xml and npm package name and how that is handled right now as it might be confusing for someone who tries to get the reasoning behind this code right now.

PR #10197, opened 2018-09-06
[Work in Progress] Babel target bundle

mariusrak:
When creating plugin it can be useful to have accessible info about target architecture of current transpilation. Plugin could therefore alter transformations of code. Based on architecture it would be possible  to determine whether result will be run on server or client and therefore hide some critical logic or perform different operations. This allows for more isomorphic code which could be easier to write.

https://github.com/meteor/meteor-feature-requests/issues/326


mariusrak:
@benjamn, could you please help with that one fialing test, which seems totally unrelated to my changes?

benjamn:
Can you rebase against `origin/release-1.7.1`, then push the result to your branch and change the target branch of this PR to `release-1.7.1`? You should see a lot fewer commits in this PR, if you do that.

Issue #10195, opened 2018-09-05
MongoDB transaction results are ignored


klaussner:
Transactions have their own oplog entries, which aren't handled by Meteor's `OplogObserveDriver`. If a transaction is committed successfully, the driver will receive an oplog entry like the one below, where `op == "c"` and `applyOps` contains the operations performed in the transaction.

**Expected behavior**

Operations in `applyOps` are handled like normal operations.

**Actual behavior**

Transaction entries are ignored and no data is sent to clients. This is because the oplog selector doesn't include
```
{ op: 'c', 'o.applyOps': { $exists: true } }
```
and doesn't consider entries in the `admin.$cmd` namespace (`ns`):
https://github.com/meteor/meteor/blob/543d4d3d51ba8c85fa697544e5f67174458c07c9/packages/mongo/oplog_tailing.js#L39-L47

**Reproduction**

https://github.com/klaussner/meteor-issues/tree/10195. If you run `Meteor.call("transaction")` in the browser console, the method will update two documents with random values but the client won't receive the new data.

**Meteor version**

1.7.1(-rc.5)

**Example oplog entry**
```json
{
  "ts":"6597672440761942017",
  "t":"1",
  "h":"6774736952743729336",
  "v":2,
  "op":"c",
  "ns":"admin.$cmd",
  "wall":"2018-09-05T09:38:41.196Z",
  "lsid":{
    "id":"rxMCGypxSJ2IfKxI5LJDOQ==",
    "uid":"47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU="
  },
  "txnNumber":"2",
  "stmtId":0,
  "prevOpTime":{
    "ts":"0",
    "t":"-1"
  },
  "o":{
    "applyOps":[
      {
        "op":"u",
        "ns":"meteor.collection",
        "ui":"0TYxDak8SumhJ1SwL6RAbw==",
        "o":{
          "$v":1,
          "$set":{
            "anotherField":0.5870170165354542
          }
        },
        "o2":{
          "_id":"A"
        }
      },
      {
        "op":"u",
        "ns":"meteor.collection",
        "ui":"0TYxDak8SumhJ1SwL6RAbw==",
        "o":{
          "$v":1,
          "$set":{
            "anotherField":0.2196245396309442
          }
        },
        "o2":{
          "_id":"B"
        }
      }
    ]
  }
}
```

benjamn:
[...]

PR #10201, opened 2018-09-06
Implement MongoDB 4.0 transaction handling in oplog_tailing.js

benjamn:
My attempt at fixing #10195.

- [x] Still needs additional `test-packages` tests.

### DIFF
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -39,7 +39,7 @@ OplogHandle = function (oplogUrl, dbName) {
   self._baseOplogSelector = {
     ns: new RegExp("^(?:" + [
       Meteor._escapeRegExp(self._dbName + "."),
-      Meteor._escapeRegExp("admin."),
+      Meteor._escapeRegExp("admin.$cmd"),
     ].join("|") + ")"),
 
     $or: [
@@ -266,39 +266,32 @@ _.extend(OplogHandle.prototype, {
     Meteor.defer(function () {
       // May be called recursively in case of transactions.
       function handleDoc(doc) {
-        let trigger = null;
-
-        if (typeof doc.ns === "string") {
-          [self._dbName, "admin"].some(name => {
-            const nameWithDot = name + ".";
-            if (doc.ns.startsWith(nameWithDot)) {
-              return trigger = {
-                collection: doc.ns.slice(nameWithDot.length),
-                dropCollection: false,
-                dropDatabase: false,
-                op: doc,
-              };
-            }
-          });
+        if (doc.ns === "admin.$cmd") {
+          if (doc.o.applyOps) {
+            // This was a successful transaction, so we need to apply the
+            // operations that were involved.
+            doc.o.applyOps.forEach(handleDoc);
+            return;
+          }
+          throw new Error("Unknown command " + EJSON.stringify(doc));
         }
 
-        if (trigger === null) {
+        const trigger = {
+          dropCollection: false,
+          dropDatabase: false,
+          op: doc,
+        };
+
+        if (typeof doc.ns === "string" &&
+            doc.ns.startsWith(self._dbName + ".")) {
+          trigger.collection = doc.ns.slice(self._dbName.length + 1);
+        } else {
           throw new Error("Unexpected ns");
         }
 
         // Is it a special command and the collection name is hidden
         // somewhere in operator?
         if (trigger.collection === "$cmd") {
-          if (doc.ns.startsWith("admin.")) {
-            if (doc.o.applyOps) {
-              // This was a successful transaction, so we need to apply
-              // the operations that were involved.
-              doc.o.applyOps.forEach(handleDoc);
-              return;
-            }
-            throw new Error("Unknown command " + EJSON.stringify(doc));
-          }
-
           if (doc.o.dropDatabase) {
             delete trigger.collection;
             trigger.dropDatabase = true;
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -290,10 +290,13 @@ _.extend(OplogHandle.prototype, {
         // somewhere in operator?
         if (trigger.collection === "$cmd") {
           if (doc.ns.startsWith("admin.")) {
-            // This was a successful transaction, so we need to apply
-            // the operations that were involved.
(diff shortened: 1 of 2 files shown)


benjamn:
@klaussner Do you think you could turn your [reproduction](https://github.com/klaussner/meteor-issues/tree/10195) into a test in the `mongo` package?

packages/mongo/oplog_tailing.js
```diff
@@ -255,12 +260,64 @@ _.extend(OplogHandle.prototype, {
...
+        if (trigger === null) {
+          throw new Error("Unexpected ns");
+        }
+
+        // Is it a special command and the collection name is hidden
+        // somewhere in operator?
+        if (trigger.collection === "$cmd") {
+          if (doc.ns.startsWith("admin.")) {
+            // This was a successful transaction, so we need to apply
+            // the operations that were involved.
```
benjamn: Is there anything else we should check to be sure this was a transaction?

packages/mongo/oplog_tailing.js
```diff
@@ -278,50 +335,25 @@ _.extend(OplogHandle.prototype, {
...
-
-          self._crossbar.fire(trigger);
+          // Fire trigger(s) for this doc.
+          handleDoc(doc);
 
           // Now that we've processed this operation, process pending
           // sequencers.
-          if (!doc.ts)
+          if (doc.ts) {
+            self._setLastProcessedTS(doc.ts);
```
benjamn: The `doc.ts` property only exists on the top-level `doc` in case of transactions, which is why I kept this code outside the `handleDoc` function.

packages/mongo/oplog_tailing.js
mitar: Probably you should check if `applyOps` exists? I am not sure if just namespace is enough to check if it is a transaction?

mitar: Do we have any contact at MongoDB to check this? I could not find any documentation/reference spec. This one is [obsolete](https://github.com/mongodb/mongo/wiki/Replication-Internals).

mitar: Found [this spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst).

klaussner:
Looks great! :+1:

 I just found a potential security risk with the updated query selector. The `admin` database contains sensitive information (including database users in the `admin.system.users` collection). Currently, insert, udpate, and delete operations on all `admin` collections are processed by the `handleDoc` function because `this._baseOplogSelector` selects every `ns` starting with `admin.` instead of just `admin.$cmd`. This wouldn't be a problem if `handleDoc` would discard irrelevant oplog entries. However, if `trigger.collection !== "$cmd"`, `handleDoc` will treat `admin` oplog entries like entries for `this._dbName`.

As a result, if an application publishes data from a collection named `system.users`, the observe driver will combine updates from `<app database>.system.users` with updates from `admin.system.users`. Consequently, if a database user is created while subscriptions for the `<app database>.system.users` collection are active, the new user document, including (hashed) credentials, will be sent to clients.

Oplog entry for a created user:

```
{
  ...
  "op": "i",
  "ns": "admin.system.users",
  ...
  "o": {
    "_id": "meteor.leakedUser",
    "user": "leakedUser",
    "db": "meteor",
    "credentials": {
      "SCRAM-SHA-1": {
        "iterationCount": 10000,
        "salt": "kNgRiEjHG/gz6Hq+vcl6Rg==",
        "storedKey": "ESE4bHkYNJpqx/tohI8Aj9ZxmeM=",
        "serverKey": "7BWPrnH1pjqTOUnvuQxIQqIQeok="
      },
      "SCRAM-SHA-256": {
        ...
      }
    },
    "roles": []
  }
}
```

mitar:
BTW, there are also some other things [I would say are necessary for Meteor to claim MongoDB 4.0 support](https://github.com/meteor/meteor-feature-requests/issues/311#issuecomment-418459853):

* We should allow to pass `session` as an argument to Mongo commands (through options).
* And we should probably create on the server-side something like `collection.insideTransaction(() => {...})` which would associate a session automatically with any Mongo command done inside that fiber. The issue is that otherwise it is really easy to miss a command and then command goes to data outside the transaction and you have crazy time debugging. Also, if you call some functions which are doing Mongo commands, it would be great if they stayed inside the transaction as well.

packages/mongo/oplog_tailing.js
chrisbobbe: @benjamn ,

We have been using the `rawCollection` exposed by Meteor in order to implement transactions, and we ran into a bug which we believe is connected to this. In the mongo driver, in `oplog_tailing.js`, we get an error stemming from the `ts` property missing on each object in the `applyOps` array of the transaction oplog entry. 

Here is the stack trace for the error:

```
 Error in oplog callback TypeError: Cannot read property 'toString' of undefined
     at packages/mongo/oplog_observe_driver.js:575:5
     at Object.Meteor._noYieldsAllowed (packages/meteor.js:848:12)
     at OplogObserveDriver._handleOplogEntrySteadyOrFetching (packages/mongo/oplog_observe_driver.js:583:9)
     at packages/mongo/oplog_observe_driver.js:137:3
     at packages/mongo/oplog_observe_driver.js:12:1
     at Object.Meteor._noYieldsAllowed (packages/meteor.js:848:12)
     at packages/mongo/oplog_observe_driver.js:122:13
     at packages/mongo/oplog_tailing.js:105:7
     at runWithEnvironment (packages/meteor.js:1356:24)
     at Object.callback (packages/meteor.js:1369:14)
     at packages/ddp-server/crossbar.js:114:36
     at Function._.each._.forEach (packages/underscore.js:139:11)
     at DDPServer._Crossbar.fire (packages/ddp-server/crossbar.js:112:7)
     at handleDoc (packages/mongo/oplog_tailing.js:318:11)
```

We tested adding a few lines to fix this, and the error went away:

```javascript
    if (doc.ns === "admin.$cmd") {
          if (doc.o.applyOps) {
            // This was a successful transaction, so we need to apply the
            // operations that were involved.

            // We added this map function to resolve our issue:
            doc.o.applyOps.map(function (op) {
                if (!op.ts) {
                    console.log('applying parent ts')
                    op.ts = doc.ts;
                }
                return op;
            }).forEach(handleDoc);
            return;
          }

          throw new Error("Unknown command " + EJSON.stringify(doc));
        }
```

Issue #1017, opened 2013-05-05
Rendering during touch event handler prevents future events from being received.


davidmccabe:
If a template is rerendered in response to a touchstart event, then, on odd-numbered times, the corresponding 'touchend' event handler is not called.

A minimal test case app can be found at: https://github.com/davidmccabe/meteor-touchtest

The issue can be worked around by setting 'preserve' on the element receiving the touch event. This should probably be the default behavior.

PR #10204, opened 2018-09-07
Note potential breaking change introduced in Release 1.6.1 

markreid:
Hi, 

I'm not sure if there's a policy about making retrospective updates to the change log, but there was a change introduced in 1.6.1 by PR #9444 that I think deserved a little more attention in the docs. 

There's been some ongoing discussion of the implications of the change [here](https://github.com/meteor/meteor/pull/9671) but to date none of the suggestions have been actioned. I'm happy to put some time towards this in the near future, but I think making the dangers a little more clear in the change log is a quick and low risk option for now. 

I've moved the change further up the list to just below the other big changes in the release that may have potential breakages and have included a clear example of how it may make certain queries provide unexpected results. I've also mentioned that it impacts `find()` queries and not just `insert` and `update` as mentioned previously. 

I brought this up in a separate issue [here](https://github.com/meteor/meteor/issues/1017) before digging deeper into the rabbit hole and finding out it'd already been discussed; this could potentially be closed out now. 

Thanks

### DIFF
--- a/History.md
+++ b/History.md
@@ -377,14 +377,14 @@
   `meteor-babel` now runs with the `loose:true` option, as required by
   other (optional) plugins like `@babel/plugin-proposal-decorators`.
   [Issue #9628](https://github.com/meteor/meteor/issues/9628)
-  
+
 * The `underscore` package has been removed as a dependency from `meteor-base`.
   This opens up the possibility of removing 14.4 kb from production bundles.
-  Since this would be a breaking change for any apps that may have been 
-  using `_` without having any packages that depend on `underscore` 
-  besides `meteor-base`, we have added an upgrader that will automatically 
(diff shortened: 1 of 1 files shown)


markreid:
i haven't made any changes beyond history.md so not sure why it'll be failing CI.

benjamn:
We did revert #9671 in https://github.com/meteor/meteor/commit/6e2b5c45033a8b79c3935b53d405e5963d6343a1 (`mongo@1.4.5`), so at least we're not stripping undefined fields on the client anymore. Perhaps we should reinstate that code, but instead of stripping undefined fields, we should warn/throw?

markreid:
cheers @benjamn, I'll probably take a pass at it early next week and get a PR up.

PR #10211, opened 2018-09-08
Exposing information about target architeture

mariusrak:
When creating plugin it can be useful to have accessible info about target architecture of current transpilation. Plugin could therefore alter transformations of code. Based on architecture it would be possible to determine whether result will be run on server or client and therefore hide some critical logic or perform different operations. This allows for more isomorphic code which could be easier to write.

meteor/meteor-feature-requests#326


benjamn:
Thanks @mariusrak!

Issue #10212, opened 2018-09-09
Minimongo incompatibility with array sorting on MongoDB 3.6+


mitar:
In MongoDB 3.6 they changed how sorting over the array behaves: https://docs.mongodb.com/manual/release-notes/3.6-compatibility/#array-sort-behavior

Because of this, Minimongo's behavior does not match anymore underlying MongoDB bundled with Meteor. More precisely, the [`minimongo - array sort` test](https://github.com/meteor/meteor/blob/devel/packages/minimongo/minimongo_tests_client.js#L2187) now fails if it is run on the server. Instead of order 0, 1, 2, 3 which Minimongo produces, MongoDB returns documents in order 2, 0, 1, 3.

I would repeat again that Minimongo tests should be run on the server to assure such compatibility in the future.

sebakerckhof:
Since I was looking into it and was initially confused: It's specifically about `A sort no longer takes the query predicate into account`, so the [test with the selector](https://github.com/meteor/meteor/blob/devel/packages/minimongo/minimongo_tests_client.js#L2219) fails.

PR #10214, opened 2018-09-11
Make array-field sorting compatible with mongo 3.6+ behavior

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/10212 . Note that while it is now compatible with Mongo 3.6+, it's incompatible with mongo < 3.6

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,10 @@
 ## v.NEXT
 
+* Minimongo's behavior for sorting fields containing an array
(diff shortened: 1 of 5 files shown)


filipenevola:
Should we add this change to the new `Breaking changes` section on History?

mitar:
Oho, that was a quick fix. Thanks!

sebakerckhof:
> Oho, that was a quick fix. Thanks!

The most fun PR's are just removing a bunch of code.

Issue #10192, opened 2018-09-04
Minimongo sorting bug with deeply nested sort fields


AJohnDoe:
The same bug as filed in #8279 is still present.

Another test case can be found [here](https://github.com/AJohnDoe/mongo-sorting-issue). Just clone, install the necessary packages and navigate to the web page. This repo is running on Meteor 1.7.0.5

### Working case

Sorting by level1.0.sortField is ok:

Natural Order

[{"text":"Entry A","arr":[{"sortField":5}],"_id":"SPQyS2qDjsdRiZRc3"},
{"text":"Entry B","arr": [{"sortField":7}],"_id":"bMSx94phbxbPXDkRX"},
{"text":"Entry C","arr": [{"sortField":1}],"_id":"kH7NN7Pzs6yqRWuGz"}]

Ascending Order

[{"text":"Entry C","arr":[{"sortField":1}],"_id":"kH7NN7Pzs6yqRWuGz"},
{"text":"Entry A","arr":[{"sortField":5}],"_id":"SPQyS2qDjsdRiZRc3"},
{"text":"Entry B","arr":[{"sortField":7}],"_id":"bMSx94phbxbPXDkRX"}]

Descending Order

[{"text":"Entry B","arr":[{"sortField":7}],"_id":"bMSx94phbxbPXDkRX"},
{"text":"Entry A","arr":[{"sortField":5}],"_id":"SPQyS2qDjsdRiZRc3"},
{"text":"Entry C","arr":[{"sortField":1}],"_id":"kH7NN7Pzs6yqRWuGz"}]

### Not working case

Sorting with a field that is nested even deeper, for example, container.arr.0.sortField, fails for the ascending sort:

Sorting by container.arr.0.sortField

Natural Order

[{"text":"Entry A","container":{"arr":[{"sortField":5}]},"_id":"CvJFgtNuxF4uedAaK"},
{"text":"Entry B","container":{"arr":[{"sortField":7}]},"_id":"tEQqbXbQyYMf2HaBY"},
{"text":"Entry C","container":{"arr":[{"sortField":1}]},"_id":"NfWxP6Tn3wjbbcCkg"}]

Ascending Order

[{"text":"Entry A","container":{"arr":[{"sortField":5}]},"_id":"CvJFgtNuxF4uedAaK"},
{"text":"Entry B","container":{"arr":[{"sortField":7}]},"_id":"tEQqbXbQyYMf2HaBY"},
[...]

Issue #10023, opened 2018-06-22
Mongo v3.6 allows nested fields to start with dollar sign ($) minimongo does not


dariusj18:
If you have documents with nested fields containing dollar signs, when you subscribe to a published collection you get an error.

**Meteor:** v1.7.0.3
**OS:** Windows 10 Pro

```
Uncaught Error: Key $text must not start with '$'
    at MinimongoError (common.js:1087)
    at assertIsValidFieldName (local_collection.js:1889)
    at Object.JSON.stringify (local_collection.js:1880)
    at JSON.stringify (<anonymous>)
    at assertHasValidFieldNames (local_collection.js:1879)
    at LocalCollection.insert (local_collection.js:94)
    at Object.update (collection.js:200)
    at Object.store.(anonymous function) [as update] (http://localhost:3000/packages/ddp-client.js?hash=7fae7e488e45468ea4fda5047b1baf99d39dfd73:604:38)
    at updateMessages.forEach.updateMessage (livedata_connection.js:1198)
    at Array.forEach (<anonymous>)
```

sebakerckhof:
I believe this is a mongo limitation (you can not have keys starting with dollar signs or containing dots).
Does your query work outside of meteor (e.g. by querying mongo directly?)

sebakerckhof:
Ow, I see now that you mention mongo 3.6 allows this. I was not aware and can't find anything in the release notes. Could you point me to where it says this has become possible?

dariusj18:
https://docs.mongodb.com/manual/reference/limits/#naming-restrictions

Under `Restrictions on Field Names`

> Changed in version 3.6.
> 
> Field names can contain dots (i.e. .) and dollar signs (i.e. $). However, only nested field names may start with a dollar sign. For example, attempting to insert { "a" : { "$b": 1 } } will succeed but attempting to insert {"$a": 1} results in error.

ctrlbrett:
What is the status on allowing "." and "$" on the  client side? Is it fixed by issue 10217?

JanMP:
While the mongo server supposedly allows for `$` and `.` in field names, the official drivers don't. I just tried with `rawCollection()` and it doesn't work (even with nesting).

sebakerckhof:
Yeah, and last year I started working on a PR to support thsi in minimongo, and it would be a really big change, which I don't think is justified.

JanMP:
Since Mongo doesn't really support this, it would actually be bad if Minimongo did.
[...]

PR #10217, opened 2018-09-12
Fix minimongo nested array sorting

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/10192 + added extra tests

### DIFF
--- a/packages/minimongo/sorter.js
+++ b/packages/minimongo/sorter.js
@@ -142,7 +142,7 @@ export default class Sorter {
       // If there are no values for a key (eg, key goes to an empty array),
       // pretend we found one undefined value.
       if (!branches.length) {
-        branches = [{value: undefined}];
+        branches = [{ value: void 0 }];
       }
 
       const element = Object.create(null);
--- a/packages/minimongo/sorter.js
+++ b/packages/minimongo/sorter.js
@@ -140,9 +140,9 @@ export default class Sorter {
       let branches = expandArraysInBranches(spec.lookup(doc), true);
 
       // If there are no values for a key (eg, key goes to an empty array),
-      // pretend we found one null value.
+      // pretend we found one undefined value.
       if (!branches.length) {
-        branches = [{value: null}];
+        branches = [{value: undefined}];
       }
 
       const element = Object.create(null);
@@ -279,12 +279,6 @@ export default class Sorter {
         minKey = key;
       }
     });
-
-    // This could happen if our key filter somehow filters out all the keys even
-    // though somehow the selector matches.
-    if (minKey === null) {
-      throw Error('sort selector found no keys in doc?');
-    }
 
     return minKey;
   }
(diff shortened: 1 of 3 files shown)


packages/minimongo/sorter.js
```diff
@@ -140,9 +140,9 @@ export default class Sorter {
       let branches = expandArraysInBranches(spec.lookup(doc), true);
 
       // If there are no values for a key (eg, key goes to an empty array),
-      // pretend we found one null value.
+      // pretend we found one undefined value.
       if (!branches.length) {
-        branches = [{value: null}];
+        branches = [{value: undefined}];
```
sebakerckhof: null is sorted higher than undefined by mongo. But an empty array should be considered the same as undefined (ref: https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#arrays)

packages/minimongo/sorter.js
```diff
@@ -280,12 +280,6 @@ export default class Sorter {
       }
     });
 
-    // This could happen if our key filter somehow filters out all the keys even
```
sebakerckhof: Motivation: this was no longer necessary after https://github.com/meteor/meteor/pull/10214 since there's no longer a key filter.

mitar:
@sebakerckhof Are you interested in fixing more general list of Minimongo issues? I have [for this package](https://github.com/peerlibrary/meteor-reactive-mongo) tried to run all Minimongo tests on the MongoDB server-side and found quite some differences. I think Meteor should run all of client-side Minomongo query-related tests on server as well, to make sure this is compatible. (Ideally, we would run also standard MongoDB query-related tests on Minimongo, but that is much harder.)

sebakerckhof:
@mitar I do have a pretty good understanding of the minimongo code by now, so I could maybe identify some low hanging fruit. The most time-consuming part is usually figuring out the MongoDB logic, since often times the documentation is seriously lacking and you have to find out the MongoDB logic with trial & error.
But note that quite some parts of minimongo are known to be approximations of MongoDB behavior to reduce the code size and complexity (or sometimes it's just impossible to achieve in javascript), so don't expect to ever fully be able to pass all tests on the server.

Is it just a matter of changing `client` to `server` here: https://github.com/meteor/meteor/blob/devel/packages/minimongo/package.js#L46 ? Or how did you run those tests?

mitar:
Yes, I move tests to server, but then I have to patch few of them. See here for details: https://github.com/peerlibrary/meteor-reactive-mongo/blob/master/tests_prepare.js

I think what we should do is split tests into client-only, both, server-only. Then that patching would not be necessary. I think once you move to the server you quickly realize what should be where and what are differences. The [list I know of those which are probably things which should be fixed in Minimongo is in this section](https://github.com/peerlibrary/meteor-reactive-mongo/blob/master/tests_prepare.js#L89).

sebakerckhof:
@mitar : A lot of the failures are because of: https://github.com/meteor/meteor/issues/10023 ( https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names ).

In true mongodb fashion the behavior is weird and undocumented. But I've figured most out of it by trail and error now.
I've created a branch where I already updated the tests and added some new ones: https://github.com/sebakerckhof/meteor/tree/fix/10023.

However, fixing that is not going to be so easy.

PR #10238, opened 2018-09-25
Remove `mongo` dependency from `autoupdate` package

klaussner:
The `autoupdate` package uses a local `Mongo.Collection` (i.e., not synchronized with MongoDB) to store hashes of current client resources, which are used to notify clients when new resources are available (in development and production). However, the collection is only used for storage and simple update operations for which a `Map` would be sufficient. If a Meteor app doesn't otherwise use MongoDB or Minimongo, this causes MongoDB to be started unnecessarily and a lot of unused code to be included in the client bundles.

This PR replaces the collection with a `Map`-based implementation, which reduces the bundle size (un-gzipped) by up to 63 kB.

To do:
- [x]  Find out if Minimongo can be removed on the server, too.
- [x] Test changes with Cordova.

### DIFF
--- a/packages/autoupdate/client_versions.js
+++ b/packages/autoupdate/client_versions.js
@@ -1,0 +1,104 @@
+import { Tracker } from "meteor/tracker";
+
+export class ClientVersions {
+  constructor() {
+    this._versions = new Map();
+    this._watchCallbacks = new Set();
+  }
+
+  // Creates a Livedata store for use with `Meteor.connection.registerStore`.
+  // After the store is registered, document updates reported by Livedata are
+  // merged with the documents in this `ClientVersions` instance.
+  createStore() {
+    return {
+      update: ({ id, msg, fields }) => {
+        if (msg === "added" || msg === "changed") {
+          this.set(id, fields);
+        }
+      }
+    };
+  }
+
+  hasVersions() {
+    return this._versions.size > 0;
+  }
+
+  get(id) {
+    return this._versions.get(id);
+  }
+
+  // Adds or updates a version document and invokes registered callbacks for the
+  // added/updated document. If a document with the given ID already exists, its
+  // fields are merged with `fields`.
+  set(id, fields) {
+    let version = this._versions.get(id);
+    let isNew = false;
+
+    if (version) {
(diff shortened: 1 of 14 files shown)


packages/autoupdate/package.js
```diff
@@ -17,8 +17,7 @@ Package.onUse(function (api) {
 
   api.use([
     'ecmascript',
-    'ddp',
-    'mongo',
```
benjamn: Awesome!

packages/autoupdate/client_versions.js
```diff
@@ -0,0 +1,57 @@
...
+  // Registers a callback that will be invoked when a version document is added
+  // or changed. Call the function returned by `watch` to remove the callback.
+  // If `skipInitial` is false, the callback will be invoked immediately for all
+  // existing version documents.
+  watch(callback, skipInitial = false) {
+    if (!skipInitial) {
+      this.versions.forEach((version) => {
+        callback(version, true);
+      });
+    }
```
benjamn: I think we should be careful to delay calling the callback until the next tick of the event loop, so that the `stop` callback is returned (and can be used to initialize a `let` or `const` variable) before the callback fires. See https://github.com/meteor/meteor/commit/f2c864ec667fc5f85e5ecb2e6379262e0dff093b for a similar fix.

abernix:
😍 I'm so excited you're working on this, @klaussner!  I tried picking this up as a hack week project almost a year ago and never got time to revisit it.

packages/reactive-dict/package.js
```diff
@@ -6,7 +6,7 @@ Package.describe({
 Package.onUse(function (api) {
   api.use(['tracker', 'ejson', 'ecmascript']);
   // If we are loading mongo-livedata, let you store ObjectIDs in it.
-  api.use('mongo', {weak: true});
+  api.use(['mongo', 'reload'], { weak: true });
```
benjamn: Nice catch!

mitar:
Awesome. @klaussner any interest in making also https://github.com/meteor/meteor-feature-requests/issues/354 happen? I think simply another sock.js socket could be opened and push versions through it.

PR #10239, opened 2018-09-27
Update terser to 3.9.1

sakulstra:
While debugging our build time, I checked if the terser update might solve some performance issues we are facing - it doesn't :D. Anyhow in a micro benchmark of 5runs the new terser version performed sligthly better so he're the pr to update it in 1.8.


sakulstra:
@sebakerckhof applied your suggestion (didn't test it though)

kzc:
If this project wants to work around [Safari's await precedence bug](https://github.com/mishoo/UglifyJS2/issues/2344) move the `safari10: true` setting from the `mangle` section to the top level.

https://github.com/terser-js/terser#minify-options

PR #10260, opened 2018-10-10
Update css minifier deps

sakulstra:
We still face some ci/cd build time issues so i went trough the biggest meteor deps and checked if we can easily strip out some mb. both postcss and cssnano dropped outdated ode versions which shouldn't be a problem. The update decreased the package size from >40mb to <10mb.

I still have to check if https://github.com/cssnano/cssnano/releases/tag/v4.0.0-rc.0 contains some relevant changes. The meteor tests succeed.

### DIFF
--- a/packages/minifier-css/minifier-tests.js
+++ b/packages/minifier-css/minifier-tests.js
@@ -92,9 +92,15 @@ Tinytest.add(
   (test) => {
     const css1 = '@import "custom.css"; body { color: "red"; }';
     const css2 = 'body { color: "blue"; }';
-    const cssAst1 = CssTools.parseCss(css1);
-    const cssAst2 = CssTools.parseCss(css2);
+    const cssAst1 = CssTools.parseCss(css1, {from: "test.css"});
+    const cssAst2 = CssTools.parseCss(css2, {from: "test2.css"});
     const mergedAst = CssTools.mergeCssAsts([cssAst1, cssAst2]);
+    const stringifiedAsts = CssTools.stringifyCss(mergedAst, {
+      sourcemap: true,
+      inputSourcemaps: false
+    });
     test.equal(mergedAst.nodes.length, 3);
+    test.equal(stringifiedAsts.map.sources.length, 2);
+    test.equal(stringifiedAsts.map.sources[0], 'test.css');
   }
 );
--- a/packages/minifier-css/minifier.js
+++ b/packages/minifier-css/minifier.js
@@ -37,9 +37,6 @@ const CssTools = {
     // controlled sourcemap generation by passing in { sourcemap: true }.
     // If included, we'll convert this to the `postcss` equivalent, to maintain
     // backwards compatibility.
-    if (!options.from){
-      options.from = undefined;
-    }
     if (options.sourcemap) {
       options.map = {
         inline: false,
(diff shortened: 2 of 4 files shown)


benjamn:
What's the reasoning behind https://github.com/meteor/meteor/pull/10260/commits/565b2439ee8a204ba4608e3655d060f8c2ce2757?

sakulstra:
while the tests in the css-minifier module succeed some other tests failed according to the error message `from` should be explicitly set to undefined when there is no file. Adding the undefined reduced the error logs, but didn't fix the failing test `run: run with mongo crash ... ` - i didn't find time to have a closer look though.

sakulstra:
while the test inside meteor now works, there's an issue with stylus. I don't know anything about stylus and as i get it it's not part of meteor, but depending on the shipped postcss version? Not sure how/where to fix this. Any hint/help appreciated

packages/minifier-css/minifier-tests.js
```diff
@@ -92,9 +92,15 @@ Tinytest.add(
   (test) => {
     const css1 = '@import "custom.css"; body { color: "red"; }';
     const css2 = 'body { color: "blue"; }';
-    const cssAst1 = CssTools.parseCss(css1);
-    const cssAst2 = CssTools.parseCss(css2);
+    const cssAst1 = CssTools.parseCss(css1, {from: "test.css"});
+    const cssAst2 = CssTools.parseCss(css2, {from: "test2.css"});
     const mergedAst = CssTools.mergeCssAsts([cssAst1, cssAst2]);
+    const stringifiedAsts = CssTools.stringifyCss(mergedAst, {
```
sakulstra: this is a test for the problems i faced with https://github.com/meteor/meteor/pull/10260/files#diff-bce46ba8fd6b413c9e323e42376b3374R59 applied - i didn't yet figure out how to solve it though.
The old `postcss()` variant works but spams errors - found a similar issue at https://github.com/stylelint/stylelint/issues/3477 but didn't fully understand how they solved it.

sakulstra:
@benjamn i think this should be ready - somehow the travis check is pending since two days though. Am i missing sth or is it just an anomaly?

packages/minifier-css/minifier.js
```diff
@@ -67,7 +66,7 @@ const CssTools = {
    */
   minifyCss(cssText) {
     const f = new Future;
-    postcss([ cssnano({ safe: true }) ]).process(cssText).then(result => {
+    postcss([ cssnano({ safe: true }) ]).process(cssText, {from: undefined}).then(result => {
```
benjamn:
```suggestion
    postcss([
      cssnano({ safe: true }),
    ]).process(cssText, {
      from: void 0,
    }).then(result => {
```

packages/minifier-css/package.js
```diff
@@ -4,8 +4,8 @@ Package.describe({
 });
 
 Npm.depends({
-  postcss: '6.0.13',
-  cssnano: '3.10.0'
+  postcss: '7.0.5',
+  cssnano: '4.1.5'
```
benjamn:
```suggestion
  cssnano: '4.1.7'
```

benjamn:
Merged into `devel` as https://github.com/meteor/meteor/commit/689ae3dfa776d3761205f41a841eba6e96c737e3. Thanks!

PR #10267, opened 2018-10-11
Remove Blaze repo based dependencies from static-html

hwillson:
(Replaces https://github.com/meteor/meteor/pull/9781)

`meteor create --minimal` pulls `tracker` into new minimal apps as a dependency. This is because it's  being pulled in as one of the dependencies on the Blaze templating tools side. Currently the `static-html` package is dependent on several Blaze repo based packages, a lot of which include Blaze specific utilities / helpers, that are really not needed when using non-Blaze apps. The `static-html` package really only needs the following Blaze repo functionality to accomplish its role:

1. [`caching-html-compiler`](https://github.com/meteor/blaze/tree/555f3bb0d34c7553423e1f01f054189378e343ea/packages/caching-html-compiler)
2. The [`html-scanner.js`](https://github.com/meteor/blaze/blob/555f3bb0d34c7553423e1f01f054189378e343ea/packages/templating-tools/html-scanner.js) file from the [`templating-tools`](https://github.com/meteor/blaze/tree/555f3bb0d34c7553423e1f01f054189378e343ea/packages/templating-tools) package.
3. And of course itself, the [`static-html`](https://github.com/meteor/blaze/tree/555f3bb0d34c7553423e1f01f054189378e343ea/packages/static-html) package.

Since the above functionality isn't specific to Blaze, let's extract those files / packages from the Blaze repo, and bring them back into Meteor core. By doing so we can reduce a lot of template based dependency overhead, when not using Blaze.

This PR creates a new copy of the `static-html` and `caching-html-compiler` packages in core, and creates a new package called `html-scanner`, to house the `html-scanner.js` functionality from the `templating-tools` package. With these changes in place, we're able to replace the following dependencies in a `meteor create --minimal` app:

```
blaze-tools
caching-html-compiler
deps
html-tools
htmljs
spacebars-compiler
static-html
templating-tools
tracker
```

with:

```
caching-html-compiler
html-scanner
static-html
```

(And this change will of course benefit non minimal `static-html` dependent apps as well.)


hwillson:
Grrr - none of these test failures happen locally 😞.

hwillson:
This PR is now in working shape technically, but we need to decide how to best update Blaze to reference these new packages (instead of using its own copies). That decision will likely also affect the version numbers used for these packages.

I'll remove the WIP and flag this as being ready for review, since the Blaze updates will happen in a separate repo/PR.

PR #10268, opened 2018-10-11
Fix duplicate WebApp.connectHandlers.use -  fix #10265

### LINKED ISSUE #10265: OAuth package : WebApp.connectHandlers.use is called twice
Somehow #9740 was reverted.

StorytellerCZ:
As reported by @tjaisson in #10265 it looks like that somewhere we have included and at the same time also reverted #9740.

### DIFF
--- a/packages/oauth/oauth_server.js
+++ b/packages/oauth/oauth_server.js
@@ -1,4 +1,3 @@
-import Fiber from 'fibers';
 import url from 'url';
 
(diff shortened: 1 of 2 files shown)


StorytellerCZ:
@tjaisson Will add.

StorytellerCZ:
I think there might have been a bad merge somewhere which resulted in this problem.

PR #10281, opened 2018-10-16
Fix uncaught TypeError in autoupdate_client.js

lgleim:
This fixes an issue (c.f. https://github.com/Urigo/meteor-client-bundler/issues/87) where when the __meteor_runtime_config__ does not contain a proper autoupdate configuration object, the client would fail to load properly.


ahoppen:
Would it be possible to publish a new version of `autoupdate` with this bugfix?

benjamn:
@ahoppen Just published `autoupdate@1.5.1`. Thanks for the reminder!

PR #10286, opened 2018-10-18
ddp-client: fix usage of reload

nathan-muir:
- Fix return type of onMigrate callback, as it should be an array.
 - Remove exception on retry, as onMigrate callback can be invoked repeatedly until all components are ready to migrate.
 - Force ddp negotiation failures to cause an uninterruptible reload.

---

I've been using these changes in a fork for some time. As, I have a lib that prevents reload until a page change occurs.

very difficult to write unit tests for these, but let me know if it's super necessary and I'll give it a shot.


benjamn:
Thanks @nathan-muir!

PR #10299, opened 2018-10-25
user.services is not an array.

gchallen:
Fixes a bug in `accounts-base` introduced by "modernization".  The check whether the email address matches the restriction pattern fails currently for external login services, since it treats `user.services` like an array rather than as an object.


StorytellerCZ:
Good catch! :+1:

macrozone:
@benjamn this bugfix was not released with 1.8.1. Can we publish a patch package?

benjamn:
@macrozone Yes! Good catch.

PR #10301, opened 2018-10-26
[oauth1] Add default value for whitelistedQueryParams.

### LINKED ISSUE #10300: [oauth1] v1.2.1 throws bc whitelistedQueryParams is undefined after modernization
Modernizing the `oauth1` package (c8f4ba7) introduced the following bug. 

The package code calls the method `OAuth._queryParamsWithAuthTokenUrl` on [line 63](https://github.com/meteor/meteor/blob/devel/packages/oauth1/oauth1_server.js#L63) w/o specifying the 4th parameter `whitelistedQueryParams` which has no default value. This leads to the error 
```
Cannot read property \'reduce\' of undefined at Object.OAuth._queryParamsWithAuthTokenUrl (packages/oauth1/oauth1_server.js:9:28)
```

on [line 9](https://github.com/meteor/meteor/blob/devel/packages/oauth1/oauth1_server.js#L9).

thomasspiesser:
Fixes #10300 introduces with modernizing the `oauth1` package.

### DIFF
--- a/packages/oauth1/oauth1_server.js
+++ b/packages/oauth1/oauth1_server.js
@@ -1,7 +1,7 @@
 import url from 'url';
 import { OAuth1Binding } from './oauth1_binding';
 
-OAuth._queryParamsWithAuthTokenUrl = (authUrl, oauthBinding, params = {}, whitelistedQueryParams) => {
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Good catch!

PR #10248, opened 2018-10-06
Release 1.8.1

benjamn:
As I explained in https://github.com/meteor/meteor/pull/9942#issuecomment-426817206, Meteor 1.8.1 will be a quick follow-up to Meteor 1.8 that upgrades Node to version 8.12.0, so that Galaxy customers can easily deploy their apps with that version of Node, even though there have been [some reports](https://github.com/meteor/meteor/issues/10216) that Node 8.12.0 has performance problems in production, due to excessive garbage collection.

We will not officially "recommend" the Meteor 1.8.1 release until we have resolved these concerns, but you can begin using it by running
```bash
meteor update --release 1.8.1-rc.n
```
where `-rc.n` refers to the most recent prerelease version of Meteor 1.8.1.

PR #10303, opened 2018-10-27
update request to latest version (drop a few dependencies)

sakulstra:
the update should be non breaking and reduces the amount of dependencies request ships with


benjamn:
Merged into `release-1.8.1` as https://github.com/meteor/meteor/pull/10248/commits/970ac2100d6608689a3a5a1cb10e208735ae8909. Thanks @sakulstra!

Issue #10298, opened 2018-10-25
[1.8] Insanely long initial build time in a big CoffeeScript project


MatejJan:
I run an [open source project](https://github.com/Retronator/Pixel-Art-Academy) with over 1,000 CoffeeScript files and ran into an issue on Meteor 1.6 where adding too many files crashed the compiler with maximum call stack size exceeded, involving the caching compiler similar to #10028. 

The discussion there led me to upgrade to 1.8 and CoffeeScript 2. After doing the necessary syntax changes, when I try to build the project now from the ground up (after meteor reset), the build runs out of memory after many many minutes of waiting.

I tried to increase the heap space to 4 GB (`TOOL_NODE_FLAGS="--max-old-space-size=4096"`), but the build still failed:

```
=> Started proxy.                             
   Building for web.browser                  -
<--- Last few GCs --->

[44085:0x103000a00]   582829 ms: Mark-sweep 4027.5 (4140.9) -> 4027.5 (4140.4) MB, 288.3 / 0.0 ms  allocation failure GC in old space requested
[44085:0x103000a00]   583126 ms: Mark-sweep 4027.5 (4140.4) -> 4027.5 (4138.9) MB, 297.7 / 0.1 ms  last resort GC in old space requested
[44085:0x103000a00]   583424 ms: Mark-sweep 4027.5 (4138.9) -> 4027.5 (4138.9) MB, 297.6 / 0.0 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x1d7058d25879 <JSObject>
    1: split(this=0x1d707c94d4f9 <Very long string[981760]>)
    2: /* anonymous */(aka /* anonymous */) [/Users/Retro/.meteor/packages/coffeescript/.2.3.1_2.e4x3ip.tz2aw++os+web.browser+web.browser.legacy+web.cordova/plugin.compile-coffeescript.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/template/lib/builder.js:~24] [pc=0x2d659ebbb38e](this=0x1d70c2d022d1 <undefined>,/* anonym...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
```

Note especially the line `split(this=0x1d707c94d4f9 <Very long string[981760]>`. That is one very long string indeed, and looking at other similar issues where people profiled the build I'm guessing this has something to do with the babel compiler?

[...]

PR #10312, opened 2018-10-30
Bump CoffeeScript to version 2.3.2 and enable lazy compilation.

benjamn:
Lazy compilation became possible in Meteor 1.8, thanks to #9983, but it's up to individual compiler plugin packages to take advantage of it.

Using lazy compilation in the `coffeescript` package should help with #10298.

### DIFF
--- a/packages/caching-compiler/caching-compiler.js
+++ b/packages/caching-compiler/caching-compiler.js
@@ -25,6 +25,10 @@ CachingCompilerBase = class CachingCompilerBase {
 
     // For testing.
     this._callCount = 0;
+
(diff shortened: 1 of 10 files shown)


benjamn:
cc @MatejJan @KoenLav @GeoffreyBooth

GeoffreyBooth:
@benjamn Thanks for taking this on. I guess it’s good that CoffeeScript is still part of the main Meteor codebase, even if in non-core, as it’s good to run its tests for any build pipeline changes. The same should happen for TypeScript I’d think.

PR #10334, opened 2018-11-14
Identify Chromium and Headless Chrome as "modern" browsers.

abernix:
This PR introduces changes to treat "Headless Chrome" and Chromium the same as Chrome in terms of identifying them as "modern" browsers (and thus serving them the modern, rather than legacy, bundles).

Best I can tell, the major version portion of Chromium versions has always tracked all the way through to Chrome Canary, Dev and Stable releases.  The [Chromium versioning documentation](https://www.chromium.org/developers/version-numbers) seems to back this up.

Furthermore, Headless Chrome isn't really a "different" Chrome — in fact, all recent versions of Chrome can be run with the `--headless` flag (yes, even the one on your — the reader's — local computer!) to be run without the visual element (and driven programmatically).  However, when run with the `--headless` flag, Chrome switches its primary `navigator.userAgent` from `Chrome/a.b.c.d` to `HeadlessChrome/a.b.c.d`.

This was initially problematic since the `useragent` npm we use for parsing user agents didn't understand this designation, however, with the update of `webapp`'s `useragent` npm in 058351b, `headlesschrome` will now have its version available from `WebAppInternals.identifyBrowser`, so we can accurately identify it and serve it the modern bundle.

Chromium on the other hand, hasn't been a prominent issue since on macOS the `navigator.userAgent` is very much the same as Chrome itself.  For example, my own macOS Chromium identifies itself as:

```
> navigator.userAgent
< "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3611.0 Safari/537.36"
```

(Note, no presence of `Chromium`)

However, on my Linux Chromium, a different behavior introduces a second `Chromium/` identifier _before_ the `Chrome/` version identifier with the same `MAJOR.MINOR.BUILD.PATCH`:

```
> navigator.userAgent
< "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/70.0.3538.77 Chrome/70.0.3538.77 Safari/537.36"
```

By updating `useragent` (https://github.com/meteor/meteor/pull/10334/commits/058351b728bfbf3d5101b467ded42ed23be62f81) and aliasing `chromium` to `chrome` (https://github.com/meteor/meteor/pull/10334/commits/845c45a51a4ff03d73d0fec35dcd5bb29e706f94), both of these situations should be improved.

### DIFF
--- a/packages/webapp/.npm/package/npm-shrinkwrap.json
+++ b/packages/webapp/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "accepts": {
-      "version": "1.3.4",
-      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
-      "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8="
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
+      "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I="
     },
     "basic-auth-connect": {
       "version": "1.0.0",
@@ -17,9 +17,9 @@
       "integrity": "sha1-0ygVQE1olpn4Wk6k+odV3ROpYEg="
     },
     "compressible": {
-      "version": "2.0.12",
-      "resolved": "https://registry.npmjs.org/compressible/-/compressible-2.0.12.tgz",
-      "integrity": "sha1-xZpcmdt2dn6YdlAOJx72OzSTvWY="
+      "version": "2.0.15",
+      "resolved": "https://registry.npmjs.org/compressible/-/compressible-2.0.15.tgz",
+      "integrity": "sha512-4aE67DL33dSW9gw4CI2H/yTxqHLNcxp0yS6jB+4h+wr3e43+1z7vm0HU9qXOH8j+qjKuL8+UtkOxYQSMq60Ylw=="
     },
     "compression": {
       "version": "1.7.1",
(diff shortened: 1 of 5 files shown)


packages/webapp/.npm/package/npm-shrinkwrap.json
```diff
@@ -111,34 +111,41 @@
...
     "inherits": {
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz",
       "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4="
     },
     "lru-cache": {
-      "version": "2.2.4",
-      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-2.2.4.tgz",
-      "integrity": "sha1-bGWGGb7PFAMdDQtZSxYELOTcBj0="
+      "version": "4.1.3",
```
abernix: This is a pretty major bump, but these shrinkwrap updates seem to have been necessary because of updates to the `blaze` submodule which `webapp` still depends on via its need for `boilerplate-generator`.  I don't believe the `lru-cache` API has ever changed in any major way and tests all seem to be passing.

packages/webapp/.npm/package/npm-shrinkwrap.json
```diff
@@ -226,9 +238,9 @@
       "integrity": "sha1-sr9O6FFKrmFltIF4KdIbLvSZBOw="
     },
     "useragent": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/useragent/-/useragent-2.2.1.tgz",
-      "integrity": "sha1-z1k+9PLRdYdei7ZY6pLhik/QbY4="
+      "version": "2.3.0",
```
abernix: This was the most important update in this shrinkwrap.

PR #10345, opened 2018-11-20
Remove multiple restart feature

klaussner:
Removes the logic that restarts a crashing Meteor server twice if there is a runtime error.
Fixes meteor/meteor-feature-requests#335.

### DIFF
--- a/tools/runners/run-app.js
+++ b/tools/runners/run-app.js
@@ -913,19 +913,9 @@ _.extend(AppRunner.prototype, {
 
   _fiber: function () {
     var self = this;
-
-    var crashCount = 0;
-    var crashTimer = null;
     var firstRun = true;
 
     while (true) {
-
-      var resetCrashCount = function () {
-        crashTimer = setTimeout(function () {
-          crashCount = 0;
-        }, 8000);
-      };
-
       var runResult = self._runOnce({
         onListen: function () {
           if (! self.noRestartBanner && ! firstRun) {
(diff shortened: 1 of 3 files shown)


tools/runners/run-app.js
```diff
@@ -892,36 +892,20 @@ _.extend(AppRunner.prototype, {
...
-      };
-
       var runResult = self._runOnce({
         onListen: function () {
           if (! self.noRestartBanner && ! firstRun) {
             runLog.logRestart();
             Console.enableProgressDisplay(false);
           }
         },
-        beforeRun: resetCrashCount,
```
klaussner: Should the `beforeRun` option be removed from `AppRunner`? It's only used here.

klaussner:
These changes uncovered a bug in the `launchOneMongoAndWaitForReadyForInitiate` function, which waits until the replica set is initiated by checking MongoDB's stdout for certain [status messages](https://github.com/meteor/meteor/blob/5e8434ea42362ffecbac0cd0c19957d3ed318db2/tools/runners/run-mongo.js#L535-L549). Under the conditions in the `mongo with multiple --port numbers` test, this function returns before the replica set is ready, leading to the `MongoError: not master and slaveOk=false` error shown in [Test Group 7](https://circleci.com/gh/meteor/meteor/29390).

klaussner:
The failing test in Test Group 7 (mongo with multiple --port numbers) will pass when 2b21a901a5a40f6e402b42b9dd916846a1ac8327 is merged into `devel`.

PR #10361, opened 2018-11-29
Get more data from Meetup API

StorytellerCZ:
I needed at least a name, but after checking the Meetup API I've added other data that I think might be useful.
https://www.meetup.com/meetup_api/docs/2/members/

### DIFF
--- a/packages/meetup-oauth/meetup_server.js
+++ b/packages/meetup-oauth/meetup_server.js
@@ -10,6 +10,12 @@ OAuth.registerService('meetup', 2, null, query => {
   return {
     serviceData: {
       id: identity.id,
+      name: identity.name,
+      lang: identity.lang,
+      link: identity.link,
+      photo: identity.photo,
+      county: identity.country,
+      city: identity.city,
       accessToken,
       expiresAt,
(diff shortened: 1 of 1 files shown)


packages/meetup-oauth/meetup_server.js
```diff
@@ -10,6 +10,12 @@ OAuth.registerService('meetup', 2, null, query => {
   return {
     serviceData: {
       id: identity.id,
+      name: identity.name,
+      lang: identity.lang,
+      link: identity.link,
+      photo: identity.photo,
+      county: identity.country,
```
axelvaindal: Shouldn't this field be named country ? :slightly_smiling_face:

StorytellerCZ: Ah, yes! Thanks!

benjamn:
Merged as 8f2c2b5805d8a9be2e49cd949f3c4131a4179078, with a51a0cfb9bc7ff70e002528614da7534d1135ea2 as a follow-up. Thanks @StorytellerCZ and @axelvaindal!

PR #10370, opened 2018-12-10
declare character encoding of the document

lew:
The error page is missing a character encoding and should be utf-8


benjamn:
@lew Thanks for this! Can you open up your branch for us to push to, so that we can do a clean re-run of these tests? Pretty sure the failures are unrelated.

hwillson:
We haven't heard back regarding https://github.com/meteor/meteor/pull/10370#issuecomment-451688128, so we'll close this for now. Thanks!

Issue #5091, opened 2015-09-01
Meteor for windows is very-very slow


JWo1F:
**_2 Upvotes**_ Hi all! Meteor for Windows is very-very slow. When I make a change I should wait for 20-30 seconds while a page reloading..

stubailo:
Please provide a full reproduction as described in: https://github.com/meteor/meteor/wiki/Contributing-to-Meteor#reporting-a-bug-in-meteor  Without a way to replicate your problem, we can't debug and fix it.

JWo1F:
Way to replicate my problem:
1. download meteor from here: https://install.meteor.com/windows
2. install meteor
3. run cmd
4. type: `meteor create test`
5. type: `cd test && meteor`
6. change one file in this dir
7. waiting from 20 to 30 seconds while page is reloading..

stubailo:
OK, we'll look into this! Thanks for the reproduction recipe, it's good to know that you are encountering this issue with a newly created app.

JWo1F:
![image](https://cloud.githubusercontent.com/assets/1540715/9618060/89537492-510c-11e5-89db-31808a44d1d0.png)

For example. The same issue in all popular browsers.

stubailo:
What Windows version are you running?

JWo1F:
Windows 10 x64 just now, but I had the same issue in windows 8.1

Slava:
Note that the screenshot is featuring a browser that we don't normally test (Yandex Browser). It is supposed to be the same as Google Chrome, but it could be an issue specific to Y.Browser too.

stubailo:
@JWo1F did say "The same issue in all popular browsers."  Hopefully that includes browsers that aren't from Yandex.

JWo1F:
Yea, @stubailo are right. I have the same issue in Chrome and Firefox (and IE / Edge).

4thlaurocruz:
this is a meteor issue, it downloading sources one at a time in debug mode. to fix this add a --production flag

JWo1F:
@4thlaurocruz in production mode I can't see full debug information about my files because they will be merged into a single file.

bjconlan:
Yeah, I too am having this issue on meteor 1.2.01 using windows 10 (32 bit).
Looks like on every js(x) code change the server is verifying the state of the project by doing the entire restart flow. 

```
Selecting package versions
Downloading ... 
Building...
Meteor server restarted
```
[...]

PR #10399, opened 2019-01-01
Windows build performance

zodern:
### Before:
The times are from the todos app. Rebuilds are for editing a file that is only used in the client, but not in a `client` folder, causing both the server and clients to be rebuilt.

#### Initial build (second and later runs):
resolveContraints: 2,149 2,261
prepareProjectForBuild: 10,299 11,230
build app: 20,656 18,792
App process started after: 36,902 35,839
Creating watchers finished: 12,383 13,462ms after app started
AppProcess onListen: 23,900 25,644ms after app started

Usuable after: 60,802 61,483

#### Rebuild

prepareProjectForBuild: 6,713 6,674 7,158
rebuildApp: 4,555 4,269 4,008
App process started after: 11,690 11,499 11,621
Creating watchers finished: 4,536 4,057 4,001ms after app started
AppProcess onListen: 8,746 8,090 8,015ms after app started
Delayed build: 9,469 8,484 8,406ms after app started

Usuable after: 20,436 19,589 19,636
Delayed build at: 21,159 19,983 20,027

#### Initial build (second and later runs):
resolveContraints: 1,392 1,879 2,053
prepareProjectForBuild: 6,601 7,267 6,708
build app: 16,056 14,270 12,810
App process started after: 28,662 26,679 24,557
AppProcess onListen: 2,007 2,214 1,593 after app started
Watchers are created after onListen and are no longer blocking

Usuable after: 30,669 28,893 26,150

#### Rebuild

prepareProjectForBuild: 6,187 6,013 6,359
rebuildApp: 2,587 2,297 1,842
App process started after: 9,256 8,757 8,811
AppProcess onListen: 1,936 1,894 1,697 after app started
Delayed build: 6,255 6,522 6,551 after app started

Usuable after: 11,192 10,651 10,508
Usuable + delayed build: 15,511 15,279 15,362

### Notable Changes

The large app times are from the rebuild profile at https://github.com/meteor/meteor/issues/5091#issuecomment-444993610.

#### Delete garbage directories async
Saves 500 - 1,100ms per rebuild (large app 9s)

#### Do not use atomicallyRewriteFile when not building in place
Saves ~650ms (large app ~29,000ms including legacy bundle) per build.
This should be safe since the build is only used once all files have been successfully written. Renaming the two files takes almost twice as long as writing it.

#### isPortable
Saves ~900ms. Since npm-rebuilds.json is never used from the bundle created by "meteor run", creating it's contents is skipped.
The `.meteor-portable` file to cache the results was never created on Windows due to the file path being in the wrong format, which slowed down the initial build and `meteor build` when using local packages.

#### inPlace server builds on windows and caching _copyDirectory
Saves 1 - 1.5s per rebuild. Most of the warnings against doing this on windows give the reason of interfering with the running app. Since the server is always stopped during the build, it should be safe to update it in place.

If it is safe to update the server in place, it might also be safe to do the same with the legacy bundle since the server doesn't serve it while it is being written.

#### Create file watcher and start legacy bundle after server started
Both of those were blocking the event loop, causing a delay between the server starting, and the tool process logging '=> server started' and letting the proxy know to forward requests.

Watcher now has an option to do the initial check async. The delayed legacy bundle blocking has not been fixed.

### DIFF
--- a/tools/isobuild/builder.js
+++ b/tools/isobuild/builder.js
@@ -549,6 +549,7 @@ Previous builder: ${previousBuilder.outputPath}, this builder: ${outputPath}`
         if (this.previousCreatedSymlinks[absFrom] !== absTo) {
         symlinkWithOverwrite(absFrom, absTo);
         }
+        this.usedAsFile[relTo] = false;
         this.createdSymlinks[absFrom] = absTo;
         return;
       }
--- a/tools/isobuild/builder.js
+++ b/tools/isobuild/builder.js
@@ -285,13 +285,27 @@ Previous builder: ${previousBuilder.outputPath}, this builder: ${outputPath}`
     } else {
       hash = hash || sha1(getData());
 
-      if (this.previousWrittenHashes[relPath] !== hash) {
+      if (this.previousWrittenHashes[relPath] !== hash && this.writtenHashes[relPath] !== hash) {
+        
         // Builder is used to create build products, which should be read-only;
         // users shouldn't be manually editing automatically generated files and
         // expecting the results to "stick".
+        // Write is called multiple times for assets when they have multiple urls for the same file
+        const mode = executable ? 0o555 : 0o444
+
+        if (this.buildPath === this.outputPath || this.writtenHashes[relPath]) {
+          // atomicallyRewriteFile handles overwriting files that have already been created
         atomicallyRewriteFile(absPath, getData(), {
-          mode: executable ? 0o555 : 0o444
+            mode
         });
+        } else {
+          // Since builder is not updating in place, and
+          // this build is only used if every file is successfully written,
+          // it is not important to write atomically.
+          files.writeFile(absPath, getData(), {
+            mode
+          })
+      }
       }
 
       this.writtenHashes[relPath] = hash;
--- a/tools/isobuild/builder.js
+++ b/tools/isobuild/builder.js
@@ -546,11 +546,11 @@ Previous builder: ${previousBuilder.outputPath}, this builder: ${outputPath}`
       if (symlink && ! (relTo in this.usedAsFile)) {
         this._ensureDirectory(files.pathDirname(relTo));
         const absTo = files.pathResolve(this.buildPath, relTo);
-        if (this.previousCreatedSymlinks[absFrom] !== absTo) {
-        symlinkWithOverwrite(absFrom, absTo);
+        if (this.previousCreatedSymlinks[absFrom] !== relTo) {
+          symlinkWithOverwrite(absFrom, absTo);
         }
         this.usedAsFile[relTo] = false;
-        this.createdSymlinks[absFrom] = absTo;
+        this.createdSymlinks[absFrom] = relTo;
         return;
       }
 
--- a/tools/isobuild/builder.js
+++ b/tools/isobuild/builder.js
@@ -10,6 +10,7 @@ import {
   optimisticReaddir,
   optimisticStatOrNull,
   optimisticLStatOrNull,
+  optimisticHashOrNull,
 } from "../fs/optimistic.js";
 
 // Builder is in charge of writing "bundles" to disk, which are
@@ -75,7 +76,9 @@ export default class Builder {
     this.previousUsedAsFile = {};
 
     this.writtenHashes = {};
+    this.createdSymlinks = {};
     this.previousWrittenHashes = {};
+    this.previousCreatedSymlinks = {};
 
     // foo/bar => foo/.build1234.bar
     // Should we include a random number? The advantage is that multiple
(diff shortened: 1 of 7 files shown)


zodern:
After the last 4 comments, the build now is:
#### Initial
resolveConstraints: same
prepareProjectForBuild: 3,827 3,575 3,667
build app: 13,345 14,427 12,939
App process started after: 23,302 23,539 22,074
AppProcess onListen: same
Usuable after: 24,928

#### Rebuild client and server
prepareProjectForBuild: 1,762 797 958
rebuildApp: 2,333 1,817 1,723
App process started after: 4,485 2,915 2,999
Listened: same
Delayed build: 3,998 3,025 3,197ms after app started

Usable after 5,579
Usable + legacy: 9,577

#### Client only rebuild before last 4 commits
onChange: 2,562 3,389 3,136
prepareProjectForBuild: 4,444 5,029 3,542
rebuildApp: 1,580 1,438 1,180
Legacy built in 3,854 3,412 3,3842

Usable after 8,440

#### Client only rebuild after last 4 commits
onChange: 1,162 1,108 1,048
prepareProjectForBuild: 2,011 797 725
rebuildApp: 1,731 1,649 1,600
legacy built in: 1,365 1,129 1,144
Usable after 4,777 (first rebuild) 3,563 (subsequent rebuilds)

The client only rebuilds are timed before the server is rebuilt for the first time.

After the initial build, the watch sets seems to include every folder and some files in `node_modules`. The watch set created after a rebuild is missing those. This does make `onChange` very fast though, so causing the server to rebuild once after the app is started might make client rebuilds a couple seconds faster.

CaptainN:
In my limited local testing, this PR makes rebuilds FASTER under Windows than under a VM on the same machine!

CaptainN:
[fast-windows-meteor-rebuild.zip](https://github.com/meteor/meteor/files/2735369/fast-windows-meteor-rebuild.zip)
[fast-windows-meteor-startup.zip](https://github.com/meteor/meteor/files/2735370/fast-windows-meteor-startup.zip)

yorrd:
Since this is not targeted at Linux, I didn't expect huge gains, but I can confirm that rebuild times are within variance and didn't change on Linux. I'd like to do some more accurate measurements but GC is getting in my way at the moment

smeijer:
I'm having build times of more than 2 minutes. This patch would be more very, very welcome!

tools/isobuild/builder.js
```diff
@@ -657,7 +680,9 @@ Previous builder: ${previousBuilder.outputPath}, this builder: ${outputPath}`
             // the read-only tools tree into a writable app."
             { mode: (fileStatus.mode & 0o100) ? 0o777 : 0o666 },
           );
+          }
```
hwillson: Would you mind fixing the indentation of this block?

tools/runners/run-app.js
```diff
@@ -835,7 +839,13 @@ _.extend(AppRunner.prototype, {
 
     Console.enableProgressDisplay(false);
 
-    const postStartupResult = runPostStartupCallbacks(bundleResult);
+    const postStartupResult = Promise.race([
+      listenPromise,
+      new Promise(resolve => setTimeout(resolve, 3000))
```
hwillson: The `3000` used here seems a bit arbitrary/magic-y. Maybe this should be defined in clearly labelled constant.

CaptainN: The commit message says "Also, the watcher and legacy bundle are only created after the server has started, or 3 seconds has passed." - it is necessary to have this 3 second time out at all? Why not always wait until the server starts?

CaptainN:
Something is causing the spinner to remain after it is finished spinning:

```
=> Meteor server restarted                   -
I20190117-23:17:18.841(-5)? http://localhost:3000/
I20190117-23:17:18.863(-5)? http://localhost:3000/
I20190117-23:17:18.873(-5)? http://localhost:3000/
=> Meteor server restarted                   |
=> Meteor server restarted                   -
=> Meteor server restarted                   |
=> Meteor server restarted                   /
```

CaptainN:
It seems that removing that 3 second race might have added back the lengthy end of build delay. I'll do more testing with and without and see if it reveals anything.

smeijer:
Is this branch present in the current beta release?

tabasco-addict:
To use the newest release, do i install normally using choco , then replace the files in the .meteor folder with the ones in the 1.8.1 repo?

benjamn:
@tabasco-addict Install via choco (which will give you Meteor 1.8.0.2) and then run
```sh
meteor update --release 1.8.1-beta.16
```
in any application directory. You can also create a new app with
```sh
meteor create --release 1.8.1-beta.16 new-181b16-app
```

smeijer:
Okay, this one is officially awesome! Thanks @zodern and @benjamn!

My build time (hot update) went down from 42 seconds to 9 seconds. Waiting for the browser refresh, takes another 5 seconds.

Usually, Meteor starts to be slower after a couple of hours development with a number of rebuilds. So I still need to monitor that one. But I have good hope.

I've installed Ubuntu 2 days ago. Turns out, that that was either a year too late, or a day to soon.

PR #10413, opened 2019-01-10
Improve `ddp-client` package

lamhieu-vk:
- [x] use `const` and `let` instead of `var`
- [x] use native functions to get data in list instead of lodash function
- [x] remove some vars not need declare if use native functions

### DIFF
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -15,9 +15,11 @@ import {
   last,
 } from "meteor/ddp-common/utils.js";
 
+let Fiber;
+let Future;
 if (Meteor.isServer) {
-  var Fiber = Npm.require('fibers');
-  var Future = Npm.require('fibers/future');
+  Fiber = Npm.require('fibers');
+  Future = Npm.require('fibers/future');
 }
 
 class MongoIDMap extends IdMap {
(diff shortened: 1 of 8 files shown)


packages/ddp-client/common/livedata_connection.js
```diff
@@ -1593,9 +1588,10 @@ export class Connection {
     }
 
     // Now add the rest of the original blocks on.
-    oldOutstandingMethodBlocks.forEach(block => {
-      self._outstandingMethodBlocks.push(block);
-    });
+    self._outstandingMethodBlocks = [
+      ...self._outstandingMethodBlocks,
+      ...oldOutstandingMethodBlocks
+    ];
```
benjamn: Let's do `self._outstandingMethodBlocks.push(...oldOutstandingMethodBlocks)` here, to avoid creating a new array.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -288,32 +290,33 @@ export class Connection {
...
       'beginUpdate',
       'endUpdate',
       'saveOriginals',
       'retrieveOriginals',
       'getDoc',
       '_getCollection'
-    ].forEach(method => {
-      store[method] = (...args) => {
+    ];
+    const store = keysOfStore.reduce((acc, method) => ({
```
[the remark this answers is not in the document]
lamhieu-vk: thanks for comment!

packages/ddp-client/common/livedata_connection.js
lamhieu-vk: ok, I updated it!

lamhieu-vk:
@benjamn thanks for your comments for this PR, I updated PR.
please review, thank you very much!

PR #10414, opened 2019-01-11
Do not treat client and server directories specially in packages.

### LINKED ISSUE #10393: [1.7-1.8] Production build errors with dynamic modules import
I am still having problems building my application with 1.7 and 1.8 for production ( Dev is OK ). I am using dynamic modules imports from methods to call functions defined in server code.

Errors happen in minify step, I tried to add debug messages to minifier-js package at this level 
https://github.com/meteor/meteor/blob/0fcc7ddd46d0ef8a278376ba64538210486ab646/packages/minifier-js/minifier.js#L25-L31

The error message is : `SyntaxError: Import statement may only appear at top level`. It occurs with code that look like this :
```js
function __minifyJs(){
// My application imports
import { a } from 'a';
import { b } from 'b';
// ...etc
}
```
It looks like dynamic modules are wrapped during build inside a function with special name `__minifyJs` which explains the error. Looking for its origin , i found this

https://github.com/meteor/meteor/blob/24865b28a0689de8b4949fb69ea1f95da647cd7a/tools/isobuild/bundler.js#L1295-L1315

Does that comment is still relevant with terser being used to minify ?

Why the module code is not transpiled to es5 before wrapping it inside that function ?

benjamn:
The `client/` and `server/` directories should be meaningful only for organizing application code.

Before this commit, if a Meteor package contained a `client/` or `server/` directory, any modules inside it could be imported, but they would not be properly compiled, because the `PackageSource#_findSources` method would ignore them in its initial scan, which determines the set of modules that get passed into compiler plugins.

Fixes #10393.

### DIFF
--- a/History.md
+++ b/History.md
@@ -8,6 +8,14 @@ N/A
 
 ### Changes
 
+* In Meteor packages, `client/` and `server/` directories no longer have
+  any special meaning. In application code, `client/` directories are
+  ignored during the server build, and `server/` directories are ignored
+  during the client build, as before. This special behavior previously
+  applied to packages as well, but has now been removed.
+  [Issue #10393](https://github.com/meteor/meteor/issues/10393)
+  [PR #10414](https://github.com/meteor/meteor/pull/10414)
+
 ## v1.8.0.2, 2019-01-07
 
 ### Breaking changes
--- a/tools/isobuild/compiler-plugin.js
+++ b/tools/isobuild/compiler-plugin.js
@@ -65,7 +65,7 @@ const hasOwn = Object.prototype.hasOwnProperty;
 // Cache the (slightly post-processed) results of linker.fullLink.
 const CACHE_SIZE = process.env.METEOR_LINKER_CACHE_SIZE || 1024*1024*100;
 const CACHE_DEBUG = !! process.env.METEOR_TEST_PRINT_LINKER_CACHE_DEBUG;
-const LINKER_CACHE_SALT = 21; // Increment this number to force relinking.
+const LINKER_CACHE_SALT = 22; // Increment this number to force relinking.
 const LINKER_CACHE = new LRU({
   max: CACHE_SIZE,
   // Cache is measured in bytes. We don't care about servePath.
--- a/tools/isobuild/compiler.js
+++ b/tools/isobuild/compiler.js
@@ -34,7 +34,7 @@ var compiler = exports;
 // dependencies. (At least for now, packages only used in target creation (eg
 // minifiers) don't require you to update BUILT_BY, though you will need to quit
 // and rerun "meteor run".)
-compiler.BUILT_BY = 'meteor/31';
+compiler.BUILT_BY = 'meteor/32';
 
 // This is a list of all possible architectures that a build can target. (Client
 // is expanded into 'web.browser' and 'web.cordova')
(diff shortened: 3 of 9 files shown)


nabiltntn:
Just to tell that i tried the build with the current version of `devel` and it works fine ! Thanks

manueltimita:
I've been trying to understand from the issue which prompted this patch: is removing this special treatment of client and server the only possibility of fixing it? If Meteor really wants to drop the packages completely, then this move would be understandable. Otherwise, it precludes one of the really great features of Meteor, the possibility of developing truly modular applications, in the vein of [Vulcan](https://docs.vulcanjs.org/architecture.html) for instance.

It is a real shame that we don't have this anymore, as it makes it more difficult to develop large yet pluggable applications.

eric-burel:
Can you explain more clearly the technical consequences of this? What happens when I import `my-package` server-side for example, and client-side? As-is, this update is worrying for application relying massively on Meteor isomorphism and package system, Vulcan is a good example.

manueltimita:
Thank you for checking this up, @benjamn. We are already using the `where` argument, exactly as you describe (actually, since Meteor 1.3, so we never had to rely on the 'special treatment' ever again). But the problem with the full rebuild when client files are changed is still there nonetheless, and it can be quickly replicated as described in https://github.com/meteor/meteor/issues/10449.

It seems like this is an older, hidden bug, as described there:

> (As a side note, also in 1.8.0.2 a server rebuild is done, when client js is located in packages/custom/client.js so the original problem is much older i suppose)

benjamn:
@manueltimita Right, what you're doing should be enough to make this work without any code changes on your side. However, there are some flaws in the way Meteor currently determines whether it's safe to do a client-only refresh (namely, it relies on `client/` directories, not the actual dependency graph). Fixing those flaws would be beneficial for everyone who uses Meteor. I've updated my previous comment to make this somewhat clearer.

manueltimita:
Thank you for the clarification, @benjamin.

aadamsx:
I'm confused.  I'm currently on Meteor 1.8 and have the "special" directories server/client/both.  This to me makes sense and I don't know why anyone would want to change that.  

What I don't like is the imports directory.  I think things should act like they do today inside the imports directory, where server/client/both mean what they say, but without the imports directory.  

Are we still going to have special directories server/client/both in the future?

benjamn:
@aadamsx We're talking about packages here, not application code. However, if you would like all your application code to be loaded lazily by default, so you don't have to worry as much about the `imports/` directory, you can use the `meteor.mainModule` configuration option in `package.json` that was [introduced in Meteor 1.7](https://github.com/meteor/meteor/blob/devel/History.md#changes-9):
* Applications may now specify client and server entry point modules in a
  newly-supported `"meteor"` section of `package.json`:
  ```js
  "meteor": {
    "mainModule": {
      "client": "client/main.js",
      "server": "server/main.js"
    }
  }
  ```
  When specified, these entry points override Meteor's default module
  loading semantics, rendering `imports` directories unnecessary. If
  `mainModule` is left unspecified for either client or server, the
  default rules will apply for that architecture, as before. To disable
  eager loading of modules on a given architecture, simply provide a
  `mainModule` value of `false`:
  ```js
  "meteor": {
    "mainModule": {
      "client": false,
      "server": "server/main.js"
    }
  }
  ```
  [Feature #135](https://github.com/meteor/meteor-feature-requests/issues/135)
  [PR #9690](https://github.com/meteor/meteor/pull/9690)

PR #10427, opened 2019-01-20
[WIP] Allow interrupting delayed builds.

benjamn:
Meteor 1.7 introduced a new `web.browser.legacy` build (#9439), which took about as long as the normal `web.browser` build, thereby almost doubling build times. Meteor 1.8 partially solved this problem by delaying the legacy build until after the development server restarted (#10055). However, long legacy builds could still be a problem if they were not finished by the time the next rebuild began, since they could not be interrupted.

This PR makes the build process more asynchronous, which is generally a good idea in any future where we rely less heavily on `Fiber`s. More immediately, this asynchronous behavior allows us to cancel builds when/if an event occurs that should trigger a rebuild (such as modifying a source file), instead of helplessly waiting for the useless legacy build to finish.

`Promise` cancelation has been proposed [multiple](https://github.com/tc39/proposal-cancelable-promises) [times](https://github.com/tc39/proposal-cancellation) for standardization, and now appears unlikely to make its way into any version of the specification ([overview](https://medium.com/@benlesh/promise-cancellation-is-dead-long-live-promise-cancellation-c6601f1f5082)). Nevertheless, the concept of a `CancelToken` still makes plenty of sense to implement as a library. The process of passing the `cancelToken` object through your code is still pretty tedious, which seems to be why cancelation was not incorporated into the language, but the effort is worthwhile when you really need to be able to cancel asynchronous work.

I will comment in the diff about other specific topics of interest.

### DIFF
--- a/tools/isobuild/compiler-plugin.js
+++ b/tools/isobuild/compiler-plugin.js
@@ -126,7 +126,7 @@ export class CompilerPluginProcessor {
     }
   }
 
-  runCompilerPlugins() {
+  async runCompilerPlugins() {
     const self = this;
     buildmessage.assertInJob();
 
@@ -1195,7 +1195,7 @@ export class PackageSourceBatch {
   }
 
   // Returns a map from package names to arrays of JS output files.
-  static computeJsOutputFilesMap(sourceBatches) {
+  static async computeJsOutputFilesMap(sourceBatches) {
     const map = new Map;
 
     sourceBatches.forEach(batch => {
@@ -1266,14 +1266,14 @@ export class PackageSourceBatch {
     const allRelocatedModules = Object.create(null);
     const scannerMap = new Map;
 
-    sourceBatches.forEach(batch => {
+    for (const batch of sourceBatches) {
       const name = batch.unibuild.pkg.name || null;
       const isApp = ! name;
 
       if (! batch.useMeteorInstall && ! isApp) {
         // If this batch represents a package that does not use the module
         // system, then we don't need to scan its dependencies.
-        return;
+        continue;
       }
 
       const nodeModulesPaths = [];
@@ -1295,10 +1295,10 @@ export class PackageSourceBatch {
         watchSet: batch.unibuild.watchSet,
       });
 
-      scanner.addInputFiles(map.get(name).files);
+      await scanner.addInputFiles(map.get(name).files);
 
       if (batch.useMeteorInstall) {
-        scanner.scanImports();
+        await scanner.scanImports();
         ImportScanner.mergeMissing(
           allMissingModules,
           scanner.allMissingModules
@@ -1306,9 +1306,9 @@ export class PackageSourceBatch {
       }
 
       scannerMap.set(name, scanner);
-    });
+    }
 
-    function handleMissing(missingModules) {
+    async function handleMissing(missingModules) {
       const missingMap = new Map;
 
       _.each(missingModules, (importInfoList, id) => {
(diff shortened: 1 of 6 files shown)


tools/isobuild/compiler-plugin.js
```diff
@@ -175,6 +177,8 @@ export class CompilerPluginProcessor {
       var sourceProcessor = data.sourceProcessor;
       var resourceSlots = data.resourceSlots;
 
+      self.cancelToken.throwIfRequested();
```
benjamn: The interruptibility of the build process depends on how often `cancelToken.throwIfRequested` gets called. Calling it too often isn't really a problem, but calling it during code that should not be interrupted can be a serious problem.

tools/isobuild/compiler-plugin.js
```diff
@@ -1295,20 +1306,25 @@ export class PackageSourceBatch {
...
       });
 
-      scanner.addInputFiles(map.get(name).files);
+      await scanner.addInputFiles(map.get(name).files);
+
+      cancelToken.throwIfRequested();
 
       if (batch.useMeteorInstall) {
-        scanner.scanImports();
+        await scanner.scanImports();
```
benjamn: The `ImportScanner#scanImports` method is an expensive chunk of the build process that doesn't currently have any `cancelToken.throwIfInterrupted()` calls within it.

tools/runners/run-app.js
```diff
@@ -835,8 +848,14 @@ _.extend(AppRunner.prototype, {
 
     Console.enableProgressDisplay(false);
 
-    const postStartupResult = runPostStartupCallbacks(bundleResult);
-    if (postStartupResult) return postStartupResult;
+    const postStartupResult = runPostStartupCallbacks({
+      bundleResult,
+      cancelToken: CancelToken.fromPromise(runPromise),
```
benjamn: This essentially races the post-startup callbacks against the `runPromise`, rather than running all post-startup callbacks before awaiting the result of the `runPromise`.

tools/tool-env/profile.js
```diff
@@ -280,7 +280,11 @@ function Profile(bucketName, f) {
     var start = process.hrtime();
     var err = null;
     try {
-      return f.apply(this, arguments);
+      const result = f.apply(this, arguments);
+      if (result && typeof result.then === "function") {
+        return Promise.await(result);
```
benjamn: Since the `Profile` abstraction assumes each chunk of hierarchical work will run to completion, we have to keep pretending async work is synchronous. If anyone has a better way to instrument asynchronous code for performance measurement, I'd love to hear about it!

tools/utils/cancelation.js
```diff
@@ -0,0 +1,41 @@
+class CancelError extends Error {}
+
+export class CancelToken {
+  constructor(exec) {
+    let requested = false;
+    let message;
+
+    exec(function cancel(reason = "cancelation requested") {
```
benjamn: This pattern is similar to the function passed to the `Promise` constructor. The ability to cancel the token (like resolving or rejecting a `Promise`) remains cleanly separated from the token itself, which can only check whether cancelation has been requested.

tools/utils/cancelation.js
```diff
@@ -0,0 +1,41 @@
...
+    return new this(cancel => {
+      promise.then(
+        result => cancel(),
+        error => cancel(error.message),
+      );
+    });
+  }
+
+  static empty() {
+    return Object.create(CancelToken.prototype);
```
benjamn: Anywhere you see `CancelToken.empty()`, we're creating a no-op `CancelToken` that will never be canceled, and whose `throwIfRequested` method will never throw.

tools/utils/cancelation.js
```diff
@@ -0,0 +1,41 @@
...
+      requested = true;
+      message = String(reason);
+    });
+
+    this.throwIfRequested = function () {
+      // It's important to introduce a gap that's long enough for file
+      // change notifications to fire, and only setTimeout seems to work.
+      new Promise(resolve => setTimeout(resolve, 0)).await();
+      if (requested) {
+        throw new CancelError(message);
```
benjamn: This `CancelError` can be caught/ignored, of course, but sometimes that power is useful, and exceptions are still the best tool for canceling work in JavaScript.

tools/tool-env/profile.js
benjamn: Unfortunately, this `Promise.await` behavior does not apply when `METEOR_PROFILE` is disabled, because then the `Profile` function just returns the original function without wrapping it. That's a huge difference in behavior that I'm not willing to accept, even though we can sort of work around it by `await`ing the result of any `Promise`-returning `Profile`-wrapped function. I'll have to come up with another approach.

jamesmillerburgess:
It sounds from the description like this PR improves rebuild performance for developers who are trigger-happy file saving.

If that is the case, then I would vote for this to be part of the Release 1.8.2 milestone rather than 1.9.0, so that we could benefit a bit earlier.

afrokick:
@benjamn @filipenevola Some problems already solved by #10824 Any plans to work on it?

filipenevola:
@benjamn do you want to continue working in this PR?

filipenevola:
I'm closing old PRs with problems like tests failing or conflicts. If you are interested in continuing this work please mention me here.

PR #10448, opened 2019-02-13
fix(deps): update terser, fetch, decimaljs, postcss, cssnano

sakulstra:
I wanted to see some bugfixes from terser in 1.8.1, but checked some other outdated dependencies in the same run.

node-fetch: https://github.com/bitinn/node-fetch/blob/master/CHANGELOG.md -> some bugfixes
postcss: https://github.com/postcss/postcss/releases -> some performance improvements and bug fixes
cssnano: https://github.com/cssnano/cssnano/releases -> some performance improvements
decimal.js: https://github.com/MikeMcl/decimal.js/blob/master/CHANGELOG.md -> seems non breaking

### DIFF
--- a/packages/fetch/.npm/package/npm-shrinkwrap.json
+++ b/packages/fetch/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "node-fetch": {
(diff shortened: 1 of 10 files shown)


benjamn:
Thanks @sakulstra!

PR #10453, opened 2019-02-14
Linker performance

zodern:
- Updates source-map to 0.5.7 which includes some performance improvements for source nodes. In one app it saves up to 1.2 seconds during full rebuilds  
- Caches the stringified output of dynamic files instead of a source node, which allows skipping stringifying the source node when the dynamic file didn't change since the last build.
- The main client bundle is not recreated when only dynamic files change. In one app this saves 2 - 3 seconds during rebuilds when a dynamic file is changed.  It only stores the latest bundle in the cache for each architecture instead of waiting until the cache has reached its max size to remove old entries.  That is to avoid increasing the tool's memory usage, though I am not sure if it is needed.

### DIFF
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -34,7 +34,7 @@ var packageJson = {
     fstream: "https://github.com/meteor/fstream/tarball/cf4ea6c175355cec7bee38311e170d08c4078a5d",
     tar: "2.2.1",
     kexec: "3.0.0",
-    "source-map": "0.5.3",
+    "source-map": "0.5.7",
     chalk: "0.5.1",
     sqlite3: "3.1.8",
     netroute: "1.0.2",
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -21,6 +21,10 @@ const APP_PRELINK_CACHE = new LRU({
     return prelinked.source.length + sourceMapLength(prelinked.sourceMap);
   }
 });
+// Caches code with source map for dynamic files
+const DYNAMIC_PRELINKED_OUTPUT_CACHE = new LRU({
+  max: Math.pow(2, 11)
+});
 
 var packageDot = function (name) {
   if (/^[a-zA-Z][a-zA-Z0-9]*$/.exec(name)) {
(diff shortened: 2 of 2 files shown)


sebakerckhof:
On a related note. `source-map` 0.7.0+ has massive performance improvements with their rust/wasm implementation. @hwillson started looking into it, but because of the wasm loading, the constructor became a promise which didn't play well with it being wrapped in a 'noYieldsAllowed' block:

https://github.com/meteor/meteor/issues/9568#issuecomment-359930688

It then never really was looked into further, but maybe @zodern feels like it 😅

tools/isobuild/linker.js
```diff
@@ -205,6 +214,54 @@ _.extend(Module.prototype, {
       });
     }
 
+    // Caching is only useful when there are dynamic files.
+    const canCache = results.find(result => result.dynamic) && self.useGlobalNamespace;
```
benjamn:
```suggestion
    const canCache = self.useGlobalNamespace && results.find(result => result.dynamic);
```

tools/isobuild/linker.js
```diff
@@ -205,6 +214,54 @@ _.extend(Module.prototype, {
...
+          });
+        }
+      });
+      inputHash = watch.sha1(JSON.stringify({
+        useGlobalNamespace: this.useGlobalNamespace,
+        combinedServePath: this.combinedServePath,
+        name: self.name,
+        files
+      }));
+      key = `${self.bundleArch}-${self.name}`;
```
benjamn:
```suggestion
```

tools/isobuild/linker.js
```diff
@@ -205,6 +214,54 @@ _.extend(Module.prototype, {
       });
     }
 
+    // Caching is only useful when there are dynamic files.
+    const canCache = results.find(result => result.dynamic) && self.useGlobalNamespace;
+    let inputHash;
+    let key;
```
benjamn:
```suggestion
    const key = `${self.bundleArch}-${self.name}`;
```

tools/isobuild/linker.js
```diff
@@ -829,6 +892,33 @@ const getPrelinkedOutputCached = require("optimism").wrap(
   }
 );
 
+function getOutputWithSourceMapCached (file,servePath, options) {
```
benjamn:
```suggestion
function getOutputWithSourceMapCached(file, servePath, options) {
```

scripts/dev-bundle-tool-package.js
```diff
@@ -34,7 +34,7 @@ var packageJson = {
     fstream: "https://github.com/meteor/fstream/tarball/cf4ea6c175355cec7bee38311e170d08c4078a5d",
     tar: "2.2.1",
     kexec: "3.0.0",
-    "source-map": "0.5.3",
+    "source-map": "0.5.7",
```
benjamn: Note to self: this will require a new dev bundle.

PR #10454, opened 2019-02-14
Write performance

zodern:
**JS minify performance**
In development, most of the time spent minifying is from converting the content of the files to and from Buffers. This was done before minifying for dynamic files to rename the function, and afterwards for all files to restore the function name. These changes have it avoid converting the contents to and from buffers where possible, and only restores the function name for dynamic files. Saves ~400ms per rebuild in a larger app.

**Skip calculating SRI for assets**
The current SRI standard supports css and js files, and Meteor only uses it for the main css and js bundles. These changes disable calculating the SRI hash for assets since Meteor currently doesn't use it for assets. If some packages or apps want the SRI hash for js or css assets, then two alternatives are calculating it for production, or skipping it for assets that are not js and css files. Saves ~700ms per rebuild in one app.

**Skip removing source map comments if the content won't be written**
Before writing a file, Meteor removes source map comments that are in the middle of the file. These changes have it skip that step when the file's contents didn't change and it is building in place. Saved 300ms during rebuilds.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -176,6 +176,9 @@ import { PackageRegistry } from "../../packages/meteor/define-package.js";
 
 const SOURCE_URL_PREFIX = "meteor://\u{1f4bb}app";
 
+const MINIFY_PLAIN_FUNCTION = Buffer.from('function(', 'utf8');
+const MINIFY_RENAMED_FUNCTION = Buffer.from('function __minifyJs(', 'utf8');
+
 // files to ignore when bundling. node has no globs, so use regexps
 exports.ignoreFiles = [
     /~$/, /^\.#/,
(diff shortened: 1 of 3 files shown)


KoenLav:
@zodern I think I'm speaking for a lot of us here when I say: thank you for all the efforts on build time improvements!

I guess the times mentioned here, and in your other recent PRs, are improvements even on top of the changes already implemented in the Meteor 1.8.1 beta?

zodern:
Thanks @KoenLav. Yes, the times are all additional, though they are from a larger app so most apps will probably not see as large of an improvement.

smeijer:
Couldn't agree more with @KoenLav, I wanted to say the same thing in #10453.

@zodern, thanks for all the time and energy you're putting into this!

Now give this guy some merge permissions! 😄

benjamn:
@zodern Happy to add you as a collaborator with push/edit permissions, by the way. Just let me know if that sounds useful to you.

zodern:
@benjamn I think that would be useful, thanks. I sent you a message on the forums with a couple of questions.

tools/isobuild/bundler.js
```diff
@@ -619,7 +625,9 @@ class File {
         hashes.push(this._inputHash);
       }
 
-      hashes.push(this.sri());
+      if (!this._skipSri) {
```
mitar: Are you sure this is OK. Because of this change, now effectively when `_skipSri` is set, contents of the file are not part of the hash anymore.

tools/isobuild/bundler.js
```diff
@@ -628,7 +636,7 @@ class File {
   }
 
   sri() {
-    if (! this._sri) {
+    if (! this._sri && !this._skipSri) {
```
mitar: I think this is wrong place to put this. I think if you call `sri()`, sha512 should always be computed.

mitar: You should wrap `manifestItem.sri = file.sri()` with `_skipSri` check.

PR #10455, opened 2019-02-15
[WIP] Lazily read and watch files

zodern:
Meteor currently reads, hashes, and watches the files that compiler plugins could compile. With Meteor 1.8, compilers can lazily compile files once Meteor knows the file will be used in the architecture. This PR delays reading and watching the file until that moment.

While this does decrease memory usage and save time from reading and hashing fewer files, the difference will probably not be noticeable for most apps. Instead, most of the improvement will be around watching. Adding, removing, or renaming files in directories Meteor watches will continue to cause a rebuild, but modifying a file not used by the app will not. Also, Meteor will do a client rebuild instead of a full rebuild when a file only used on the client is modified, even if it is not in a folder named `client`.

This does not change how local packages are watched since Meteor reads all of their files when the package is modified.

What is left:
- [ ] handle data being null when lazily reading file
- [ ] ensure the watch set includes all of the files it should in all situations
- [ ] lazily read html files in `node_modules`

### DIFF
--- a/packages/meteor/plugin/basic-file-types.js
+++ b/packages/meteor/plugin/basic-file-types.js
@@ -11,11 +11,25 @@ Plugin.registerCompiler({
 
 var CssCompiler = function () {
 };
+
+function hasDir(path, dirName) {
+  var pathParts = path.split('/');
+  var index = pathParts.indexOf(dirName);
+
+  return index > -1 && index < pathParts.length - 1;
+}
+
 CssCompiler.prototype.processFilesForTarget = function (inputFiles) {
   inputFiles.forEach(function (inputFile) {
+    let path = inputFile.getPathInPackage();
+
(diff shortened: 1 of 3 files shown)


stolinski:
I'd love to see this merge. It's a major source of slowdown in my dev process.

filipenevola:
I'm ok with this approach! It seems to be worth it!

filipenevola:
Hi @zodern could you resolve this conflict and update it with devel? Thanks

zodern:
I am closing this in favor of https://github.com/meteor/meteor/pull/11474, which uses a similar implementation to decide when to do a full or client only rebuild.  It doesn't reduce the number of files that are read and watch, but I didn't like how this PR did that since it made error handling and other things more difficult.

PR #10090, opened 2018-07-20
Update Node to version 8.12.0 (prerelease).

benjamn:
As explained by @MylesBorins in [the proposal PR for Node 8.12.0](https://github.com/nodejs/node/pull/21593), the Node team is planning to have a longer Release Candidate process this time, to give contributors a chance to test everything more thoroughly.

This sounds like a great idea to us! In the spirit of thorough testing, this PR is intended to help vet the Node 8.12.0 changes in the context of Meteor, by running our extensive test suite using the new Node version.

Note: since there isn't an official RC release of Node 8.12.0 yet (keep an eye on https://nodejs.org/download/rc/ for that), and we only have a functional Node build system for Unix platforms, this PR currently applies only to Linux and Mac builds, with Windows still using Node 8.11.3.

I'm not sure what version of Meteor the Node 8.12.0 changes will land in, given the uncertainty of the Node release timeline, and given that https://github.com/meteor/meteor/pull/9942 is looking pretty stable these days. We'll put this PR into a milestone once we have a better idea about the timing.

PR #10478, opened 2019-03-04
Batching Oplog Entries & DDP messages

KoenLav:
This PR implements a couple of things:
- DDP now supports sending an Array of messages, rather than a singular message;
- In the Session on the Server, the messages for every individual client are buffered (if additional messages are received within 10ms, the buffer fills up until at most 1000 messages have accrued or 500ms has passed);
- In the Crossbar on the server OplogEntries are buffered for every (potentially re-used) ObserveHandle (if additional entries are received for the same ObserveHandle within 5ms, the buffer fills up until at most 1000 entries or 500ms have passed).

This decreases the load on server setup for two reasons:
- when messages are buffered inside the Session on the Server there is less overhead while stringifying individual messages and sending individual messages over the WebSocket;
- when messages are buffered when they are received in the Crossbar we give the server a chance to recognize the fact it is falling behind too far (and possibly fall back to polling).

To see the effect of this PR I would recommend cloning this repository (https://github.com/koenlav/meteor-fiber-repro) and comparing the performance with the development branch.

=====

This PR supersedes both https://github.com/meteor/meteor/pull/9862 and https://github.com/meteor/meteor/pull/9885, but does not (yet) achieve the desired end-result from https://github.com/meteor/meteor/pull/9876.

While this PR makes sure we buffer at the start and end of the process from MongoDB => socket, it is likely to be possible to refactor all classes involved in between to accept 'messages' (rather than added, changed, removed), which would allow us to make use of the full potential of these buffers; we would only need to unwind the array of messages in one or two places on the server.

### DIFF
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -1607,9 +1607,8 @@ export class Connection {
   }
 
   onMessage(raw_msg) {
-    let msg;
     try {
-      msg = DDPCommon.parseDDP(raw_msg);
+      var messages = DDPCommon.parseDDP(raw_msg);
     } catch (e) {
       Meteor._debug('Exception while parsing DDP', e);
       return;
@@ -1621,46 +1620,52 @@ export class Connection {
       this._heartbeat.messageReceived();
     }
 
-    if (msg === null || !msg.msg) {
-      // XXX COMPAT WITH 0.6.6. ignore the old welcome message for back
-      // compat.  Remove this 'if' once the server stops sending welcome
-      // messages (stream_server.js).
-      if (!(msg && msg.server_id))
-        Meteor._debug('discarding invalid livedata message', msg);
-      return;
-    }
+    messages = Array.isArray(messages) ? messages : [messages];
 
-    if (msg.msg === 'connected') {
-      this._version = this._versionSuggestion;
-      this._livedata_connected(msg);
-      this.options.onConnected();
-    } else if (msg.msg === 'failed') {
-      if (this._supportedDDPVersions.indexOf(msg.version) >= 0) {
-        this._versionSuggestion = msg.version;
-        this._stream.reconnect({ _force: true });
-      } else {
-        const description =
-          'DDP version negotiation failed; server requested version ' +
-          msg.version;
-        this._stream.disconnect({ _permanent: true, _error: description });
-        this.options.onDDPVersionNegotiationFailure(description);
+    for (var i = 0; i < messages.length; i++) {
+      var msg = messages[i];
+
+      if (msg === null || !msg.msg) {
+        // XXX COMPAT WITH 0.6.6. ignore the old welcome message for back
+        // compat.  Remove this 'if' once the server stops sending welcome
+        // messages (stream_server.js).
+        if (!(msg && msg.server_id))
+          Meteor._debug('discarding invalid livedata message', msg);
+        return;
       }
-    } else if (msg.msg === 'ping' && this.options.respondToPings) {
-      this._send({ msg: 'pong', id: msg.id });
-    } else if (msg.msg === 'pong') {
-      // noop, as we assume everything's a pong
-    } else if (
-      ['added', 'changed', 'removed', 'ready', 'updated'].includes(msg.msg)
-    ) {
-      this._livedata_data(msg);
-    } else if (msg.msg === 'nosub') {
-      this._livedata_nosub(msg);
-    } else if (msg.msg === 'result') {
-      this._livedata_result(msg);
-    } else if (msg.msg === 'error') {
-      this._livedata_error(msg);
-    } else {
-      Meteor._debug('discarding unknown livedata message type', msg);
+
+      if (msg.msg === 'connected') {
+        this._version = this._versionSuggestion;
+        this._livedata_connected(msg);
+        this.options.onConnected();
+      } else if (msg.msg === 'failed') {
+        if (this._supportedDDPVersions.indexOf(msg.version) >= 0) {
+          this._versionSuggestion = msg.version;
+          this._stream.reconnect({ _force: true });
+        } else {
+          const description =
+            'DDP version negotiation failed; server requested version ' +
+            msg.version;
+          this._stream.disconnect({ _permanent: true, _error: description });
+          this.options.onDDPVersionNegotiationFailure(description);
+        }
+      } else if (msg.msg === 'ping' && this.options.respondToPings) {
+        this._send({ msg: 'pong', id: msg.id });
+      } else if (msg.msg === 'pong') {
+        // noop, as we assume everything's a pong
+      } else if (
+        ['added', 'changed', 'removed', 'ready', 'updated'].includes(msg.msg)
+      ) {
+        this._livedata_data(msg);
+      } else if (msg.msg === 'nosub') {
+        this._livedata_nosub(msg);
+      } else if (msg.msg === 'result') {
+        this._livedata_result(msg);
+      } else if (msg.msg === 'error') {
+        this._livedata_error(msg);
+      } else {
+        Meteor._debug('discarding unknown livedata message type', msg);
+      }
     }
   }
 
--- a/packages/ddp-common/utils.js
+++ b/packages/ddp-common/utils.js
@@ -38,80 +38,102 @@ export function last(array, n, guard) {
   return slice.call(array, Math.max(array.length - n, 0));
 }
 
-DDPCommon.SUPPORTED_DDP_VERSIONS = [ '1', 'pre2', 'pre1' ];
+DDPCommon.SUPPORTED_DDP_VERSIONS = [ '2', '1', 'pre2', 'pre1' ];
 
 DDPCommon.parseDDP = function (stringMessage) {
   try {
-    var msg = JSON.parse(stringMessage);
+    var messages = JSON.parse(stringMessage);
   } catch (e) {
-    Meteor._debug("Discarding message with invalid JSON", stringMessage);
+    Meteor._debug("Discarding message(s) with invalid JSON", stringMessage);
     return null;
   }
-  // DDP messages must be objects.
-  if (msg === null || typeof msg !== 'object') {
-    Meteor._debug("Discarding non-object DDP message", stringMessage);
-    return null;
-  }
+  
+  // Convert all DDP messages to Array form
+  messages = Array.isArray(messages) ? messages : [messages];
 
-  // massage msg to get it into "abstract ddp" rather than "wire ddp" format.
+  for (let i = 0; i < messages.length; i++) {
+    var msg = messages[i];
 
-  // switch between "cleared" rep of unsetting fields and "undefined"
-  // rep of same
-  if (hasOwn.call(msg, 'cleared')) {
-    if (! hasOwn.call(msg, 'fields')) {
-      msg.fields = {};
+    // Each individual DDP message must be an object.
+    if (msg === null || typeof msg !== 'object') {
+      Meteor._debug("Discarding non-object DDP message", stringMessage);
+      
+      messages.splice(i, 1);
+
+      i--;
+
+      continue;
     }
-    msg.cleared.forEach(clearKey => {
-      msg.fields[clearKey] = undefined;
+
+    // massage msg to get it into "abstract ddp" rather than "wire ddp" format.
+
+    // switch between "cleared" rep of unsetting fields and "undefined"
+    // rep of same
+    if (hasOwn.call(msg, 'cleared')) {
+      if (! hasOwn.call(msg, 'fields')) {
+        msg.fields = {};
+      }
+      msg.cleared.forEach(clearKey => {
+        msg.fields[clearKey] = undefined;
+      });
+      delete msg.cleared;
+    }
+
+    ['fields', 'params', 'result'].forEach(field => {
+      if (hasOwn.call(msg, field)) {
+        msg[field] = EJSON._adjustTypesFromJSONValue(msg[field]);
+      }
     });
-    delete msg.cleared;
   }
 
-  ['fields', 'params', 'result'].forEach(field => {
-    if (hasOwn.call(msg, field)) {
-      msg[field] = EJSON._adjustTypesFromJSONValue(msg[field]);
-    }
-  });
-
-  return msg;
+  return messages;
 };
 
-DDPCommon.stringifyDDP = function (msg) {
-  const copy = EJSON.clone(msg);
+DDPCommon.stringifyDDP = function (messages) {
+  messages = Array.isArray(messages) ? messages : [messages];
 
-  // swizzle 'changed' messages from 'fields undefined' rep to 'fields
-  // and cleared' rep
-  if (hasOwn.call(msg, 'fields')) {
-    const cleared = [];
+  const clonedMessages = [];
 
-    Object.keys(msg.fields).forEach(key => {
-      const value = msg.fields[key];
+  for (let i = 0; i < messages.length; i++) {
+    const msg = messages[i];
+    const copy = EJSON.clone(msg);
 
-      if (typeof value === "undefined") {
-        cleared.push(key);
-        delete copy.fields[key];
+    // swizzle 'changed' messages from 'fields undefined' rep to 'fields
+    // and cleared' rep
+    if (hasOwn.call(msg, 'fields')) {
+      const cleared = [];
+
+      Object.keys(msg.fields).forEach(key => {
+        const value = msg.fields[key];
+
+        if (typeof value === "undefined") {
+          cleared.push(key);
+          delete copy.fields[key];
+        }
+      });
+
+      if (! isEmpty(cleared)) {
+        copy.cleared = cleared;
+      }
+
+      if (isEmpty(copy.fields)) {
+        delete copy.fields;
+      }
+    }
+
+    // adjust types to basic
+    ['fields', 'params', 'result'].forEach(field => {
+      if (hasOwn.call(copy, field)) {
+        copy[field] = EJSON._adjustTypesToJSONValue(copy[field]);
       }
     });
 
-    if (! isEmpty(cleared)) {
-      copy.cleared = cleared;
+    if (msg.id && typeof msg.id !== 'string') {
+      throw new Error("Message id is not a string");
     }
 
-    if (isEmpty(copy.fields)) {
-      delete copy.fields;
-    }
+    clonedMessages.push(copy);
   }
 
-  // adjust types to basic
-  ['fields', 'params', 'result'].forEach(field => {
-    if (hasOwn.call(copy, field)) {
-      copy[field] = EJSON._adjustTypesToJSONValue(copy[field]);
-    }
-  });
-
-  if (msg.id && typeof msg.id !== 'string') {
-    throw new Error("Message id is not a string");
-  }
-
-  return JSON.stringify(copy);
+  return JSON.stringify(clonedMessages);
 };
--- a/packages/ddp-server/crossbar.js
+++ b/packages/ddp-server/crossbar.js
@@ -6,12 +6,21 @@ DDPServer._Crossbar = function (options) {
   var self = this;
   options = options || {};
 
-  self.nextId = 1;
+  self.listenerId = 1;
+  self.bufferId = 1;
   // map from collection name (string) -> listener id -> object. each object has
   // keys 'trigger', 'callback'.  As a hack, the empty string means "no
   // collection".
   self.listenersByCollection = {};
   self.listenersByCollectionCount = {};
+  // An object which holds the buffered changes per collection
+  self.buffersPerCollection = {};
+  // Buffer changes to the same collection which happen within x ms
+  self.bufferInterval = 5;
+  // Maximum age of the buffer
+  self.bufferMaxAge = 500;
+  // Maximum amount of notifications to store in the buffer before flushing
+  self.bufferMaxSize = 1000;
   self.factPackage = options.factPackage || "livedata";
   self.factName = options.factName || null;
 };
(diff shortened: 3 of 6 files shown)


KoenLav:
Tests are likely to fail because not all parts of the code that make use of DDP expect/understand message Arrays instead of singular messages; I will definitely need some help pinpointing these issues.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -1621,46 +1620,52 @@ export class Connection {
...
-      // compat.  Remove this 'if' once the server stops sending welcome
-      // messages (stream_server.js).
-      if (!(msg && msg.server_id))
-        Meteor._debug('discarding invalid livedata message', msg);
-      return;
-    }
+    messages = Array.isArray(messages) ? messages : [messages];
+
+    for (var i = 0; i < messages.length; i++) {
+      var msg = messages[i];
```
KoenLav: After this line nothing aside from indenting was changed actually, GitHub just does a bad job at highlighting that.

packages/ddp-common/utils.js
```diff
@@ -38,80 +38,102 @@ export function last(array, n, guard) {
   return slice.call(array, Math.max(array.length - n, 0));
 }
 
-DDPCommon.SUPPORTED_DDP_VERSIONS = [ '1', 'pre2', 'pre1' ];
+DDPCommon.SUPPORTED_DDP_VERSIONS = [ '2', '1', 'pre2', 'pre1' ];
```
KoenLav: I guess a new version of DDP would be necessary.

Things should be backwards compatible!

packages/ddp-common/utils.js
```diff
@@ -38,80 +38,102 @@ export function last(array, n, guard) {
...
-    Meteor._debug("Discarding non-object DDP message", stringMessage);
+    Meteor._debug("Discarding message(s) with invalid JSON", stringMessage);
     return null;
   }
+  
+  // Convert all DDP messages to Array form
+  messages = Array.isArray(messages) ? messages : [messages];
+
+  for (let i = 0; i < messages.length; i++) {
+    var msg = messages[i];
```
KoenLav: Same here, nothing after this line changed.

packages/ddp-common/utils.js
```diff
@@ -38,80 +38,102 @@ export function last(array, n, guard) {
...
-DDPCommon.stringifyDDP = function (msg) {
-  const copy = EJSON.clone(msg);
+DDPCommon.stringifyDDP = function (messages) {
+  messages = Array.isArray(messages) ? messages : [messages];
+
+  const clonedMessages = [];
+
+  for (let i = 0; i < messages.length; i++) {
+    const msg = messages[i];
+    const copy = EJSON.clone(msg);
```
KoenLav: Same here, no changes after this line.

packages/ddp-server/crossbar.js
```diff
@@ -87,31 +96,125 @@ _.extend(DDPServer._Crossbar.prototype, {
     var self = this;
 
     var collection = self._collectionForMessage(notification);
+    var listenersForCollection = self.listenersByCollection[collection];
+    var bufferForCollection = self.buffersPerCollection[collection];
 
-    if (! _.has(self.listenersByCollection, collection)) {
-      return;
+     if (listenersForCollection) {
```
KoenLav: I'm not sure whether we even still need the listeners (we could also refactor any other usages of listener to make use of buffer instead, I think).

packages/ddp-server/livedata_server.js
```diff
@@ -265,6 +265,26 @@ var Session = function (server, version, socket, options) {
   // List of callbacks to call when this connection is closed.
   self._closeCallbacks = [];
 
+  // DDP clients with version 2 and up should support batching of DDP messages
+  var allowBuffering = version >= 2;
```
KoenLav: Buffering is enabled by default.

packages/ddp-server/livedata_server.js
```diff
@@ -265,6 +265,26 @@ var Session = function (server, version, socket, options) {
...
+
+  // When updates are coming within this ms interval, batch them together.
+  self._bufferedMessagesInterval = allowBuffering ? options.bufferedMessagesInterval || 10 : 0;
+
+  // Flush buffers immediately if messages are happening continuously for more than this many ms.
+  self._bufferedMessagesMaxAge = options.bufferedMessagesMaxAge || 500;
+
+  // Maximum amount of messages to store in the buffer before flushing
+  self._bufferedMessagesMaxAmount = options.bufferedMessagesMaxAmount || 1000;
```
KoenLav: These values (10, 500, 1000) are somewhat arbitrary, I guess mileage may vary.

packages/ddp-server/crossbar.js
```diff
@@ -6,12 +6,21 @@ DDPServer._Crossbar = function (options) {
...
   self.listenersByCollection = {};
   self.listenersByCollectionCount = {};
+  // An object which holds the buffered changes per collection
+  self.buffersPerCollection = {};
+  // Buffer changes to the same collection which happen within x ms
+  self.bufferInterval = 5;
+  // Maximum age of the buffer
+  self.bufferMaxAge = 500;
+  // Maximum amount of notifications to store in the buffer before flushing
+  self.bufferMaxSize = 1000;
```
KoenLav: These values (5, 500, 1000) are somewhat arbitrary, I guess mileage may vary.

packages/ddp-client/common/livedata_connection.js
smeijer: Offtopic advise, append `?w=1` to the review url to get rid of whitespace.

https://github.com/meteor/meteor/pull/10478/files/2870711fbb508d8375f413b3e0bf31c81ed29d8f?w=1

Or, install [`refined github`](https://github.com/sindresorhus/refined-github) and have a <kdb>No Whitespace</kdb> button:

image

KoenLav:
@benjamn any updates on this?

mitar:
I would like to re-iterate my [comment here](https://github.com/meteor/meteor/pull/9862#issuecomment-387464966):

I think this is a substantial change, and also requires DDP protocol changes, which I would not take so lightly at this point in Meteor development life (all other DDP clients would then have to be updated).

Moreover, it might be an unnecessary change if underlying fiber pool issue is fixed. I think it might be better approach to get node used by Meteor to have a fix addressing the fiber pool issue. In that case both this issue and any other code path which makes a spike of fibers would perform better. This is why we need a test to measure performance improvement. So that we could compare existing code vs. existing code with fiber patch vs. existing code with batching of messages vs. existing code with fiber patch with batching of messages. Then we can know which approach to take.

KoenLav:
@mitar in reply to that I'd like to reiterate my [comment here](https://github.com/meteor/meteor/pull/9862#issuecomment-388606958).

Some other reiterations and observations:

1. Anyone can test and see the benefits of this PR easily (long before any Fibers issues are triggered):
"To see the effect of this PR I would recommend cloning this repository (koenlav/meteor-fiber-repro) and comparing the performance with the development branch."

2. You keep asking for tests, which I have provided (and anyone can easily reproduce themselves), but seem unwilling to test things yourself?

"This is why we need a test to measure performance improvement." => DIY?

3. The update to the DDP spec is implemented in a backwards compatible manner (clients which do not accept Array form messages won't receive them):
"I guess a new version of DDP would be necessary. Things should be backwards compatible though!"

"All other DDP clients would then have to be updated" => not true.

4. The issue of 'creating' a lot of Fibers should be resolved regardless, as [confirmed by](https://github.com/meteor/meteor/pull/9862#issuecomment-388664984) @kentonv (it's not a good idea to create a huge amount of Fibers due to the allocation of a stack for each Fiber). Implementing a pool for the Fibers alleviates the issue when it does occur (so it is [a good idea regardless](https://github.com/meteor/meteor/pull/9862#issuecomment-388712672)), but that doesn't mean we shouldn't try to keep it from happening in the first place.

"Moreover, it might be an unnecessary change if underlying fiber pool issue is fixed." => Not true.

Summarizing I think our goals should be:
- we don't want Fiber spikes;
- if they do occur, we want them to have a minimal impact.

Can we at least agree on that?

ps apologies if my reply seems rather direct, but I feel like you're only trying to make a very specific point without (objectively) looking at the pros and cons.

mitar:
> Anyone can test and see the benefits of this PR easily (long before any Fibers issues are triggered):

Of course, but I am saying that fixing Fiber issues would provide similar benefits. And is a more general solution.

> The update to the DDP spec is implemented in a backwards compatible manner (clients which do not accept Array form messages won't receive them):

I see, you check for the version. Nice.

> apologies if my reply seems rather direct, but I feel like you're only trying to make a very specific point without (objectively) looking at the pros and cons.

No no. Direct replies are good. :-)

> we don't want Fiber spikes;
> if they do occur, we want them to have a minimal impact.

See, we do not agree here. We do not want Fiber spikes. So let's fix that. No need for the second part.

KoenLav:
Thanks for your constructive reply! :)

The issue in Node.js (V8 actually) which limits the impact of a spike in Fibers has been resolved and merged into Node.js 8.12: https://github.com/meteor/meteor/pull/10090#issuecomment-406744113

The proposed patch by @nathan-muir moves the Fibers into a pool which promotes re-use of earlier created stacks rather than creating a new stack per Fiber (as far as I understand).

The above are ways to mitigate the huge performance penalties associated with Fiber spikes. This PR, on the other sides, attempts to create a way for Meteor to recognize when it will create a Fiber spike and bail out (by fetching once, instead of 1000's of times).

Meteor has always implemented logic to fall back to fetching for a specific collection if there was too much to fetch, but as far as I can tell this condition will never be triggered (because at some point Meteor will consume 100% CPU, of which about 50% for incoming Oplog notifications and 50% for working through them, so the 'counter' which keeps track of how far we are behind stays at roughly the same number, never triggering a fetch).

This PR accomplishes two things:
- improve performance for high-throughput Collections in Meteor by batching work together (as seen in the tests);
- by batching we allow Meteor to recognize it will cause a Fiber spike and bail out to fetching instead of processing the oplog.

packages/mongo/oplog_observe_driver.js
```diff
@@ -486,6 +489,11 @@ _.extend(OplogObserveDriver.prototype, {
       // but fetch() yields.
       Meteor.defer(finishIfNeedToPollQuery(function () {
         while (!self._stopped && !self._needToFetch.empty()) {
+          // If we need to fetch more than a thousand documents, re-running is probably quicker
+          if (self._needToFetch.size() >= 1000) {
+            self._needToPollQuery();
+          }
+
```
KoenLav: @mitar 

Here we decide to bail out on fetching 1000 documents via individual calls to MongoDB for each document and instead fetch all the needed documents in one operation.

On line 515 below you can see that the _docFetcher will be called for every document which 'currently needs to be fetched', which is the cause of the Fibers spike.

ps I'm not sure that this is already the perfect solution, but I am 100% sure that we have isolated the cause and should be working on ways to resolve it.

pps an alternative approach would be:
- batch Oplog messages (otherwise Meteor can never find out when to fetch multiple documents in one operation);
- pass an Array of _ids to the _docFetcher (basically rewriting the _docFetcher to fetch multiple docs).

But we will need to discuss this with people with more intricate knowledge of how this has been setup.

retani:
> In the Session on the Server, the messages for every individual client are buffered (if additional messages are received within 10ms, the buffer fills up until at most 1000 messages have accrued or 500ms has passed);

Is there an easy way to opt out of this or configure the numbers? I have several real time apps that rely and have always relied on instant sending of DDP messages (and do not cater many simultaneous clients)

mitar:
@retani Do you know that there is already batching happening on the client side? You can disable that, but just checking if you are doing that.

retani:
> @retani Do you know that there is already batching happening on the client side? You can disable that, but just checking if you are doing that.

Good point. I know but I don't know how to disable it. How to do it? Is it documented somewhere?

mitar:
> Good point. I know but I don't know how to disable it. How to do it? Is it documented somewhere?

Not sure if it is documented, it should be. It is one or two attributes on the `Meteor.connection` instance.

KoenLav:
@retani I've implemented the "buffered/batched listening" alongside the pre-existing "immediate listening" and have used a flag throughout the code to enable it, so it would be trivial to allow this behavior to be configurable on a per subscription basis.

retani:
@KoenLav that's great! So how would I disable it? 
I think the problem might be documentation here. There should be some more information here: https://docs.meteor.com/api/pubsub.html#Meteor-subscribe

sebakerckhof:
So I think there are a couple of reasons this isn't progressing:

1. Honestly, I don't think any of us feels comfortable enough with these parts of the code to guarantee it won't break stuff
Ok, there is the DDP version which should ensure that old DDP clients don't break. But it has the potential to external packages like meteor-streams / meteor-direct-stream-access which are not always maintained anymore but still quite popular.
It may also change semantics in certain subtle ways that are not immediately obvious.

2. It's hard to validate the performance claims in real world applications. Sure, there is a test repo, but that's a micro benchmark which may not translate to real world usage with lots of subscribers on normal use cases. Furthermore, it only looks at CPU, but now you may get memory spikes from the buffers, which, with a lot of subscribers may lead to memory issues resulting in the opposite effect on CPU.

3. This PR does 2 unrelated things: buffering of the oplog messages and ddp messages. I think it would be better if it would be split out.
IIRC, this whole thing started because of fiber spikes. So you started with buffering DDP messages until it was discovered that this is not what causes the fiber spikes, but rather the oplog tailing.
So all that is left as argument for the DDP buffers is that it is more performant to serialize one big object instead of multiple small ones, but from what I see this isn't backed by the evidence ( e.g. https://github.com/meteor/meteor/pull/9862#issuecomment-387596487 ). So are there any gains and is it worth the risk of introducing a new DDP version, possibly breaking other packages, having memory spikes etc?

The crossbar stuff seems more easily acceptable to me.

KoenLav:
I think the main reason this PR isn't progressing is because nobody from the Meteor team has taken the time to take a proper look at it, before you just did.

Thanks for taking the time to do so!

Now that is out of the way; I think your criticism makes a lot of sense. Splitting the PR in smaller parts which are easier to review.

I don't agree that batching in Oplog and DDP aren't unrelated (when batching Oplog entries it makes sense to batch them all the way to the client, rather than unwinding them before sending them over the wire), but I do see how splitting these things would make these changes easier to discuss.

If there's interest in improving the performance of the OplogObserver/Crossbar (which helps Meteor in spotting when a Fiber spike may occur due to a high number of Oplog entries) then I'd be happy to work on these changes.

If not, that's also fine, but then it's best to just close this PR and leave things as is.

sebakerckhof:
> I don't agree that batching in Oplog and DDP are unrelated (when batching Oplog entries it makes sense to batch them all the way to the client, rather than unwinding them before sending them over the wire),

You would be right if there is only 1 publication for each collection, but the thing is that it's not the oplog entries which are being send to the client. It's the updates from the session collection view (aka mergebox). So they have to be unwinded and processed one-by-one to be able to send the deltas. That's why I say they're unrelated.

In fact this is already a form of batching (over multiple publications).

So let's say you have 2 publications on the same collection, this is the flow of the messages:
```
pub 1 oplog tail entry -> multiplexer -> observe handle -> session collection view 
                                                                                    \
                                                                                     => ddp message
                                                                                    /
pub 2 oplog tail entry -> multiplexer -> observe handle -> session collection view 
```

So let's say you are subscribed to 2 publications on the same collection. If a document gets removed it will be removed from both publications, but the session collection view will make sure you only get 1 removed message.
If a document gets removed from just a single publication (because it does not match the query anymore), but it still matches the other one, the session collection view will make sure you don't get any update message.

And this is of course unique for each user, so you can not batch from oplog to ddp.

KoenLav:
Good point! 

So what you're saying is the Oplog messages need to be filtered at some point, which definitely requires looping over them, but this doesn't require sending the DDP messages one-by-one.

Oplog changes and DDP messages are definitely not the same, we agree on that :)

Still happy to help with improving the performance of Meteor.

sebakerckhof:
yes, so just to give a more concrete example.

Let's say there a user is subscribed to 2 publications, returning these cursors:
```js
Notifications.find({}, { fields: { a: 1, b: 1, c: 1 } });
Notifications.find({ b: 1 }, { fields: { b: 1, c: 1, d: 1 } });
```

Let's say the following actions happen on the notification collection:

1.
```js
Notifications.insert({ _id: 'abc', a: 0, b: 0, c: 0, d: 0 });
```
This will result in the following cursor events (from the publications):
```
pub 1: added { _id: 'abc', a: 0, b: 0, c: 0  }
```
And the following ddp messages
```
collection notifications: added { _id: 'abc', a: 0, b: 0, c: 0  }
```

2.
```js
Notifications.update('abc', { $set: { b: '1' } });
```
This will result in the following cursor events (from the publications):
```
pub 1: changed - fields: { b: 1  }
pub 2: added { _id: 'abc', b: 1, c: 0, d: 0  }
```
And the following ddp messages
```
collection notifications: changed  - fields: { b: 1 }
collection notifications: changed  - fields: { d: 0 }
```

3.
```js
Notifications.update('abc', { $set: { a: '1', c: '1' } });
```
This will result in the following cursor events (from the publications):
```
pub 1: changed - fields: { a: 1, c: 1  }
pub 2: changed - fields { c: 1 }
```
And the following ddp messages
```
collection notifications: changed  - fields: { a: 1, c: 1 }
```

4.
```js
Notifications.update('abc', { $set: { b: '2' } });
```
This will result in the following cursor events (from the publications):
```
pub 1: changed - fields: { b: '2'  }
pub 2: removed - _id: abc
```
And the following ddp messages
```
collection notifications: changed  - fields: { b: 2 }
collection notifications: changed  - unset: ['d']
```

And these DDP messages would be different for each user.

sebakerckhof:
So given the above, the only benefit of batching DDP messages would be that you can stringify one big object instead of multiple small ones. But from what i read here: https://github.com/meteor/meteor/pull/9862#issuecomment-387596487 this doesn't really make a difference and you would need a new DDP version, possibly breaking quite a few third party packages, possibly leading to memory spikes etc.

I think it's safe to say that unless it can be proven that stringify-ing one big object is a major performance improvement that the DDP batching isn't going in.

The batching of oplog entries is a different situation. It's lower risk and seems to have bigger benefits if it reduces fiber spikes.
I'll try to take a closer look this week and see if we can get it in, or at least finally give a decision on this PR.

sebakerckhof:
I haven't had enough time last week (or this one), but one thing that needs to be further explored is how this impacts locking up the event loop. Since the whole batch (talking about crossbar batching here) has to be processed without yields, it may affect fairness amongst users by locking up the event loop for extended periods, leading to a worse long tail in the response times. This may be just hypothetical and not have a noticeable impact, but should be further investigated.

mitar:
Thank you for this work and investigation you have done/are planing to do.

About processing, newer node versions support workers/threads, maybe some of this processing could be offloaded there?

sebakerckhof:
Yeah, I remember @veered looked into that some time ago, but concluded it was more work than anticipated, but that may have been before workers got introduced to node.

packages/ddp-server/crossbar.js
StorytellerCZ: Might be worth making these configurable.

radekmie:
We've experimented with this change in our apps and described our findings here: https://forums.meteor.com/t/performance-and-stability-problems-at-scale/56230. Most importantly: it really helps. If we could help with getting this one merged, do let me know.

KoenLav:
We would still like to see this get merged too, hope someone finally finds/prioritizes some time to look into it from MDG.

StorytellerCZ:
@radekmie @KoenLav I will fix the merge conflict today. Then we need to make the tests pass and then we can merge it.

StorytellerCZ:
@KoenLav can you please take a look on the failing tests? I would love to merge this soon.

KoenLav:
@StorytellerCZ I'd like to, but I would definitely need some help (I don't have any experience with the tests, and quite a few seem to be failing).

What I wrote at the time was:

> Tests are likely to fail because not all parts of the code that make use of DDP expect/understand message Arrays instead of singular messages; I will definitely need some help pinpointing these issues.

I took another look just now and I think:
- The frontend tests just need to be rewritten to understand Arrays;
- The backend tests check for things ending up in the oplog, but they don't understand this is being buffered now (I think);
- The package catalog tests need to be rewritten to understand Array messages;

A lot of these problems could probably be resolved by NOT batching for tests, but then we would need to write some specific tests which do enable batching and also test the functionality.

filipenevola:
Hi @KoenLav how much work it would be necessary to keep batching disabled by default?

In this case we could avoid all the possible breaking changes and also just new tests would be necessary as the current tests should still work as before.

It would be like this PR https://github.com/meteor/meteor/pull/11368 where the current behavior is preserved but new options are added when optimizations are necessary.

I would like to move this PR forward specially after the real case test from @radekmie but it would be also nice to avoid breaking changes and a way to keep the current behavior.

KoenLav:
@filipenevola I've taken a quick look and added a global (per server instance, for all connected client) option:

`DDPCommon.ALLOW_BUFFERING = false;`

And a local (per client) option:

`bufferedMessagesInterval: 0,`

This should allow us to run the tests without buffering enabled and start adding tests which actually test buffering.

ps I think a few tests might still fail because currently the Crossbar implementation isn't 100% backwards compatible yet (as in it buffers for listeners as well, which causes listeners to only fire once the buffer is flushed). I'll look into this later.

KoenLav:
The remaining failing tests are probably all caused because of changes to how we process the oplog.

I could try using the old logic when ALLOW_BUFFERING = false, but I don't think it will be as easy as disabling the batched DDP messages.

Probably won't have time to look into this for another week or two, so if anyone else could take a look (at either fixing one of the "broken" tests or using the old logic) that would be much appreciated.

Ps I think the tests rely on Meteor processing changes from the Oplog right away, so if we could check for the result after bufferInterval instead of immediately they would probably be fixed.

filipenevola:
I'm closing this PR because we haven't heard anything from the original poster for a while and the tests are failing.

ignl:
This PR seemed so close maybe there is still some chance to reopen and @KoenLav could take a look when he has time? IMO any performance improvement to meteor is very valuable.

KoenLav:
I definitely intend to take another look at it, at some point.

But I'd really appreciate some help on fixing the tests :)

packages/ddp-client/common/livedata_connection.js
dognybulur: Aktif

Issue #6864, opened 2016-04-20
1.3.2.2 contains two jquery versions


mxab:
Hi,

I created a fresh project and added jquery via npm and use it via `import jQuery from 'jquery'`.

The application has still the packages/jquery.js (1.11.2) but also in the modules.js the npm jquery version

See this basic project sample
https://github.com/mxab/meteor-jquery-issue

![screen shot 2016-04-20 at 12 03 36](https://cloud.githubusercontent.com/assets/1607547/14671304/616dbc94-06f1-11e6-8dbd-bac9b718d980.png)

![screen shot 2016-04-20 at 12 03 11](https://cloud.githubusercontent.com/assets/1607547/14671307/63734b76-06f1-11e6-9125-4438fa81df57.png)

Ben305:
You have to remote the meteor package for `jquery`. This is still included in your sample.

mxab:
@Ben305 I thought the idea of the 1.3.2 release was that it would then take the jQuery provided by the node modules. Also removing jquery from `packages` does not show any effect, I assume to some transitive dependency in in some other meteor package.

abernix:
At the moment `blaze-html-templates` requires `jquery` itself.  Thus if you continue to use `blaze-html-templates` (or any other private/public package that uses `jquery` using the Atmosphere system) this will continue.

With your example, if you remove `jquery` and `blaze-html-templates` with `meteor remove`, you should see the second version go away.

I know there are efforts under way to fix this in Blaze, but don't remember where the ticket is right now (If you want to peek around, Blaze progress is happening over in https://github.com/meteor/blaze now)

mxab:
@abernix have you seen this issue https://github.com/meteor/meteor/issues/6626 ? As far as I get it the meteor jquery package should detect jquery in the node modules and does not load the built in version, shouldn't it?

bySabi:
You can make the trick creating a local 'jquery' that import already installed jquery module. For example: https://github.com/bySabi/neal-sample-meteor/tree/master/packages/meteor-jquery--module

Remember update version on meteor release updates.

abernix:
If you install your own version of jQuery with NPM, it should be used.  If you don't find this to be the case, please reproduce in the latest version of Meteor and upload a reproduction repo to Github and provide the link.

Thanks!

#7202 (related)

coagmano:
Is there a way to stop `meteor/jquery` from bundling version `1.11.2` to the browser when an npm version exists?

Can dynamic imports help here?

Issue #6626, opened 2016-03-29
Deprecate core jQuery package


ramijarrar:
Now that we have full support for NPM packages, having a duplicate version of jQuery in Meteor Core leads to integration issues with packages that depend on the official NPM version. Would it be possible to get rid of this package altogether or take a similar approach to the react package?

ramijarrar:
Also, in the meantime can you suggest any way to point NPM packages to the version provided by Meteor core (like we could do with Browserify transforms)?

benjamn:
I agree! One way to approach this might be to release a new version of the `jquery` Meteor package that imports `jquery` from the app's `node_modules` directory, throwing a helpful exception if it's not installed.

ramijarrar:
@benjamn Sounds great. What's the timeline for something like this (i.e 1.3.1, 1.4, etc.)? In the past we were using [exposify](https://www.npmjs.com/package/exposify) to mitigate this issue, can we do something similar in 1.3?

skirunman:
Related to #3456 and also slightly related to cleaning up some unneeded and unwanted Meteor core jquery dependencies. See #6563.

Siyfion:
Yeah this is actually causing me a huuuuge headache as I've got a duplicated version of jQuery, one has the bootstrap modal library, one doesn't. Other NPM packages are trying to call bootstrap jQuery functions on the version without them... It's all getting rather messy!

ramijarrar:
@benjamn Is this ready for testing in 1.3.2-rc.0?

benjamn:
Once Meteor 1.3.2 is released today or tomorrow, it will be possible to avoid including jQuery in your project at all, or to use your own version throughout the project by `npm install`ing it yourself. You can test it out by doing `meteor update --release 1.3.2-rc.5` now. Thanks for your feedback, everyone!

skirunman:
@benjamn Just wanted to make sure you saw this comment as @mitar mentioned there may be an issue with using the jquery.isPlainObject code he created running on the client. https://github.com/meteor/meteor/commit/fbf7617111d0d4b7ee7ee7e63a6861f2dfc0fd16#comments
Great work on getting this fixed.

lorensr:
I'm on `1.3.2-rc.5`, I've got a lot of atmo packages that depend on atmo `jquery`, so I can't remove it. 

When I `npm remove --save jquery`, I get browser console error on pageload `cannot find module 'jquery'`, probably because I have npm packages that depend on jquery. 

[...]

PR #10498, opened 2019-03-19
Avoid bundling jquery twice

### LINKED ISSUE #10289: Two jquery versions in meteor initial bundle
Meteor 1.8

Create simple-todos app.
Run meteor npm install --save jquery
Start an app with bundle visualizer

Expected:
From what I understand meteor will only use npm jquery version and the old one won't be used and should not be loaded in the bundle

Result:
There are two versions of jquery in the bundle

benjamn:
Closes #10289
Previously: #6864, #6626

cc @arggh


packages/non-core/jquery/package.js
```diff
@@ -1,7 +1,7 @@
 Package.describe({
   summary: "Manipulate the DOM using CSS selectors",
   // This package currently uses jQuery 1.12.1 (due to #9605).
```
klaussner: This comment is out of date now. :slightly_smiling_face:

afrokick:
As @benjamn said in https://github.com/meteor/meteor/pull/10522#issuecomment-546131281

> I would love to see that merged, but the test failures indicate that the major version bump (to jQuery 3) is causing version solver conflicts. This is no surprise, since packages that depend on jquery almost certainly need to be updated to work with jQuery 3 instead of 1.12.1 (or maybe they can stop using jQuery entirely), but I'm not sure how much work that will take.

sebakerckhof:
Ok, so the following packages depend on jquery:
**core**
1. _test-in-browser_ -> Works with jquery 3.4.1. But doesn't seem like it even needs the jquery dependency. This also depends on the deprecated bootstrap package (which has problems with jquery 3), but only for styling. So this bootstrap dependency can easily be removed with some css in the package. No risk invovled.

2. _test-helpers_ -> Works with jquery 3.4.1, but only uses it for 1 `find()`, which can be replaced by `querySelectorAll` these days. No risk involved.

**Non-core**
1. _Blaze_ -> would be hard to change, but I verified that it works with version 3.4.1 and doesn't use any deprecated API. All tests pass. However, user applications might be broken due to slight changes in the jquery behavior between 1.12 and 3.4. But that's not much of an issue since this would be marked as a major version upgrade anyway

2. _Spacebars-tests_ -> Works & passess all tests with jquery 3.4.1. No risk here.

**Deprecated**
- _Bootstrap_ -> Bootstrap 1 requires jquery 1. But this package is deprecated. If we break the use of it in `tests-in-browsers` (easy, since it only uses it for the navbar css), then we can ignore this one.

So should be pretty straightforward.

PR #10499, opened 2019-03-19
Fix bug in minimongo fields projection

maxnowack:
This fixes a null reference exception, if an array contains null values while compiling a fields projection.


StorytellerCZ (approved):
Would prefer if there was a dedicated test for it rather than using existing, but it will do.

Issue #10491, opened 2019-03-14
OIDC with POST method reqest


rooch84:
In our project we need to authenticate with a custom OIDC provider.  We need to use hybrid flow rather than implicit flow, and as a result, the redirect comes back with a POST request, as such the params are in the body rather than the query. 

In `packages/oauth/oauth_server.js`, line 161, it is  assumed the params are in the query...
 `handler(service, req.query, res);`

I propose checking the method type and retrieving the params from the body if the method in POST...

```
if (req.method === "POST") {
        const syncBody = Meteor.wrapAsync((request, cb) => {
          let body = [];
          request
            .on("data", chunk => {
              body.push(chunk);
            })
            .on("end", () => {
              body = Buffer.concat(body).toString();
              cb(null, body);
            });
        });
  
        let params = syncBody(req)
          .split("&")
          .reduce((a, kv) => {
            let kvArr = kv.split("=");
            a[kvArr[0]] = decodeURIComponent(kvArr[1]);
            return a;
          }, {});
  
        handler(service, params, res);
      } else {
        handler(service, req.query, res);
      }
```

I can confirm that this works with both our custom OIDC provider (POST), and Gitlab OIDC (GET). 

Meteor version:
- Unreleased, running from a checkout at 86b6086be (HEAD -> devel, origin/devel, origin/HEAD)

benjamn:
Seems like a good idea. Do you have time to submit a PR? I would just ask that we avoid `Meteor.wrapAsync` and use `Promise`s instead, if possible.

Since this would not be a breaking change for the `oauth` package, we can release it whenever we like, and it doesn't need to be tied to the next Meteor release (1.8.1).

rooch84:
Thanks.  I've changed it to a Promise and submitted pull request #10501

jramer:
Also needed for "Sign in with Apple". More info in my closed issue(meteor/meteor-feature-requests#370).

rooch84:
PR here: https://github.com/meteor/meteor/pull/10501

It's not clear what else is required.

jramer:
This already taken care of in #10751  and #10501 is closed.

PR #10501, opened 2019-03-20
Checks request method type. For POST request, extracts params from b…

rooch84:
…ody.

See #10491


rooch84:
OK, I don't think Test Groups 10 and 5 are related to my commit. 

Ideally, the oauth1 and oauth2 tests should be expanded to test both GET and POST requests, but it looks like the request is mocked, and I'm not sure how to create it such that so that it still has the `on` function.

rooch84:
Any updates on when this can be pulled?  Anything more required?

rooch84:
It's not clear what's causing Test 8 to fail.  Any suggestions?

packages/oauth/oauth_server.js
```diff
@@ -158,7 +158,35 @@ const middleware = (req, res, next) => {
...
+            }, {});
+            handler(service, params, res);
+          })
+          .catch(err => {
+            Log.warn(`Unable to parse message body: ${err}`);
+          });
+      } else {
+        handler(service, req.query, res);
+      }
   } catch (err) {
```
jramer: The code doesn't handle to deliver state to the if on line 198:
`if (req.query.state && err instanceof Error) {`

packages/oauth/oauth_server.js
```diff
@@ -158,7 +158,35 @@ const middleware = (req, res, next) => {
...
+              return arr;
+            }, {});
+            handler(service, params, res);
+          })
+          .catch(err => {
+            Log.warn(`Unable to parse message body: ${err}`);
+          });
+      } else {
+        handler(service, req.query, res);
+      }
```
jramer: The code doesn't handle to deliver state to the if on line 198:
`if (req.query.state && err instanceof Error) {`

filipenevola:
I'm closing this PR in favor of https://github.com/meteor/meteor/pull/10751

packages/oauth/oauth_server.js
rooch84: Bah! Looks like I missed the email. Sorry.

rooch84:
Good.  I'm just glad that I'll no longer need to manage a separate oauth package from now on :-)

PR #10517, opened 2019-04-08
Use meteor-babel@7.4.3 to compile meteor/tools (but not application or package code, yet).

benjamn:
Since we're updating from version 7.3.4, the following releases are included:
* https://github.com/babel/babel/releases/tag/v7.4.0
* https://github.com/babel/babel/releases/tag/v7.4.1
* https://github.com/babel/babel/releases/tag/v7.4.2
* https://github.com/babel/babel/releases/tag/v7.4.3

Note that we are also (finally) removing `babel-runtime` from the dev bundle, since everything now uses `@babel/runtime`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,15 @@
 ## v.NEXT
 
+### Breaking changes
+N/A
+
+### Migration Steps
+N/A
+
+### Changes
+
+* The `meteor-babel` npm package has been updated to version 7.4.3.
+
 ## v1.8.1, 2019-04-03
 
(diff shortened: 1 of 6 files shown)


benjamn:
Currently blocked by this Babel issue: https://github.com/babel/babel/issues/9720#issuecomment-481030523

There's also a test failing because the `stylus` package was still relying on the old `babel-runtime` package from the dev bundle, but that should be fixable on our side.

PR #10518, opened 2019-04-09
Add isomorphic URL api with BC

sebakerckhof:
Similar to the recent fetch package, this would add an isomorphic way to use the WHATWG URL api in legacy browsers (I needed this mainly for IE11).

Unfortunately there was already a url package exporting `URL`. So for BC reasons I have to add those old methods to the global URL function.

### DIFF
--- a/packages/url/package.js
+++ b/packages/url/package.js
@@ -1,18 +1,29 @@
 Package.describe({
-  summary: "Utility code for constructing URLs",
-  version: "1.2.0"
+  name: "url",
+  version: "1.2.0",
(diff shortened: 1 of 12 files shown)


sebakerckhof:
@filipenevola this should be ready for review now. It should be backwards compatible, but it will add some weight to the legacy bundle (because of the polyfills). So someone has to decide if this is worth it.

packages/url/package.js
```diff
@@ -1,18 +1,29 @@
 Package.describe({
-  summary: "Utility code for constructing URLs",
-  version: "1.2.0"
+  name: "url",
+  version: "1.2.0",
```
StorytellerCZ, line 3: Should probably bump the version.

PR #10520, opened 2019-04-09
Temporary Meteor prerelease to validate fix for issue #10516.

### LINKED ISSUE #10516: [!] Cannot deploy on iOS 12.2 because of missing support for Swift 4+
iOS 12.2 - Meteor 1.8.0.2 - Xcode 10.1

Screen Shot 2019-04-07 at 19 10 22

The solution would obviously be to upgrade to Xcode 10.2. Unfortunately, as reported in this other [issue](https://github.com/meteor/cordova-plugin-meteor-webapp/issues/76), we cannot use Xcode 10.2 because Meteor is still using Swift 3, which has been deprecated and now removed from Xcode 10.2.

IMHO, this is now a hot issue because until support for Swift 4+ is added to Meteor, _**it's not possible to deploy on any iOS 12.2+ device**_.

benjamn:
Thanks to https://github.com/meteor/cordova-plugin-meteor-webapp/pull/78 by @rj-david, we are close to a fix for #10516 (🤞), but I wanted to validate the fix before publishing `webapp@1.7.4`, since that will cause almost every Meteor 1.8.x app to update to the new `webapp` version.

It turns out the easiest way to publish a prerelease version of the `webapp` package is to publish a prerelease version of Meteor and `meteor-tool`, for obscure reasons that I explain in the commits.

You can update to this Meteor version by running
```sh
meteor update --release 1.8.1-issue-10516.0
```
in any application directory.

This Meteor prerelease does not need to become an official release. As soon as we verify that `webapp@1.7.4-issue-10516.0` fixes #10516, we can just publish `webapp@1.7.4` and forget about this release branch.

### DIFF
--- a/packages/meteor-tool/package.js
+++ b/packages/meteor-tool/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "The Meteor command-line tool",
-  version: '1.8.1-issue-10516.0'
+  version: '1.8.1'
 });
 
 Package.includeTool();
--- a/packages/meteor-tool/package.js
+++ b/packages/meteor-tool/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "The Meteor command-line tool",
-  version: '1.8.1'
+  version: '1.8.1-issue-10516.0'
 });
 
 Package.includeTool();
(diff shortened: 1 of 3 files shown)


macrozone:
i noticed that meteor does not update automatically the versions in `.meteor/cordova-plugins`

I had to change to `cordova-plugin-meteor-webapp@1.7.0` there. I tried to remove this completly, but this breaks the app in cordova.

PR #10523, opened 2019-04-11
Fix checking for a rerender timeout

zodern:
Fixes a typo that caused ProgressDisplayFull to sometimes render after it was replaced with a different progress display.


benjamn:
TypeScript might have helped catch this: #10528

PR #10525, opened 2019-04-11
Do not remove source map url comments in public files

### LINKED ISSUE #10471: sourceMappingURL is stripped from js files in public/
Meteor version: 1.8.0.2 
OS: macOS 10.14.3
Expected behaviour: Files in public/ should be "served as-is to the client" (https://guide.meteor.com/structure.html). That includes any `sourceMappingURL` comments at the bottom of the file.
Actual behaviour: `//# sourceMappingURL=` is stripped from the file. This causes some problems because I am writing files here that have sourcemaps.

Reproduction:

- Clone https://github.com/brendonboshell/simple-todos
- Run `meteor` and request `http://localhost:3000/test.js`. File returned is not the same content as `public/test.js`.

zodern:
Fixes #10471. This should also have a small build performance improvement for apps with many assets.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -1721,7 +1721,9 @@ class ClientTarget extends Target {
       manifestItem.sri = file.sri();
 
       if (! file.targetPath.startsWith("dynamic/")) {
-        writeFile(file, builder);
+        writeFile(file, builder, {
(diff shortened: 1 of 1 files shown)


benjamn:
@zodern I think you should be able to squash-and-merge, if you want to do the honors!

PR #10526, opened 2019-04-12
Fix variable scope

lpinca:
Currenlty the `v` variable is only visible in the `if` block but it is
also used in the outer scope. Move the declaration to the outer scope.


lpinca:
@benjamn thank you.
There is another small issue that prevents us from updating to the latest EJSON version in https://github.com/primus/ejson.

This import syntax here https://github.com/meteor/meteor/blob/efc8a3525db6067bf1d381c033c6ac937c15d631/packages/ejson/ejson.js#L393 is invalid. It is not a dynamic import so it should be at top level.

Would you be interested in a PR to fix that?

lpinca:
Opened https://github.com/meteor/meteor/pull/10535

PR #10535, opened 2019-04-27
Move static import to top level

lpinca:
The static `import` syntax can only be used at the top level of a file.

### DIFF
--- a/packages/ejson/ejson.js
+++ b/packages/ejson/ejson.js
@@ -1,3 +1,5 @@
+import canonicalStringify from './stringify';
+
 /**
  * @namespace
  * @summary Namespace for EJSON functions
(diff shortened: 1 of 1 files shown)


sebakerckhof:
Meteor allows to have static imports in a nested element to have static evaluation while still allowing for static analysis, more info: https://github.com/benjamn/reify/blob/master/PROPOSAL.md

lpinca:
Yes, but this code is not valid outside of Meteor. See https://github.com/meteor/meteor/pull/10526#issuecomment-487255107. We are publishing Metor EJSON as a standalone module for Node.js. https://github.com/primus/ejson

lpinca:
Right now I worked around this by transforming the `import` statement to a `require` statement https://github.com/primus/ejson/blob/f3dd797c6eeb138592ea8f86f9897e686e6e946a/webpack.config.js#L17.

lpinca:
The above solution is bit hacky and fragile but if this patch is not wanted I can live with it :)

filipenevola:
Hi, the nested import is a very nice feature of Meteor. I was able to achieve a very good startup time in a Cordova app because of it combined with dynamic modules as well.

I'm going to close this as we believe nested imports are great then it's ok to use it to avoid evaluating parts of the code that are not used all the time.

Issue #10563, opened 2019-05-17
[1.8.1] Meteor legacy builds don't work for node_modules


sebakerckhof:
It's the first time I need to create a Meteor app which should support older browser versions.
I thought I'd finally be able to profit from the legacy/modern split builds. But it turns out my imported node_modules aren't being transpiled and therefore I get syntax errors in older browsers.

Minimal reproduction: https://github.com/sebakerckhof/meteorlegacy
This is a default meteor boilerplate app, in which I just added 1 npm module: 'pify' which I import on the client side. The application then breaks in older browsers, because the pify module (part of the packages/modules.js) is not transpiled and contains const declarations, arrow functions etc.

Can be verified e.g. with an ie9 vm from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

When I symlink the package into the /imports folder, it does get transpiled, but I thought this was no longer necessary since 1.8...

benjamn:
The symlinking trick is still necessary in 1.8.1, but perhaps it's time to start compiling `node_modules` more aggressively (by default) for the `web.browser.legacy` bundle.

sebakerckhof:
Aah, ok. I must have misunderstood. I thought because of the lazy compilation the imports symlink was no longer necessary. Then again, doing it by default might increase rebuild times for legacy substantially, but maybe this isn't much of an issue once https://github.com/meteor/meteor/pull/10427 lands.

CaptainN:
I'm getting an error with material-ui 4.01. which ships a esm folder, and has its package.json's module field set up to use the content from within esm (esm is pre-compiled, except the import/export statements, which is what fails for me).

It'd also be nice to be able to tell it to use the es folder instead, since those are non-precompiled, and we may be able to get some modern bundle benefits by starting with those modules, instead of esm modules.

```
W20190530-16:33:39.433(-4)? (STDERR) (node:26476) UnhandledPromiseRejectionWarning: C:\Users\Kevin\repos\pixstori\node_modules\@material-ui\core\esm\Card\index.js:1
W20190530-16:33:39.434(-4)? (STDERR) (function (exports, require, module, __filename, __dirname) { export { default } from './Card';
```

If this is a separate issue, I'll move it to its own ticket.

benjamn:
[...]

Issue #9473, opened 2017-12-14
native async/await support


staeke:
### This bug report should include:
- [X] A short, but descriptive title. The title doesn't need "Meteor" in it.
- [X] The version of Meteor showing the problem.
- [ ] The last version of Meteor where the problem did _not_ occur, if applicable.
- [X] The operating system you're running Meteor on.
- [X] The expected behavior.
- [X] The actual behavior.
- [X] A **simple** reproduction!
      (example: A GitHub repository that anyone can clone to observe the problem.)

**Meteor version:** 1.6.0.1
**OS**: Mac OSX 10.13.1

By reading this https://blog.meteor.com/announcing-meteor-1-6-abf30f29e2c6, I was under the impression that native async/await was supported from Meteor 1.6. Depending on if that' the case, you could argue that this should instead be a feature request. If that's the case, please let me know and I'll report it there instead.

Consider this code
```js
const pDelay = delay => new Promise(res => Meteor.setTimeout(res, delay));

(async () => {
    await pDelay(0);
    await pDelay(0);
})();
```

If you compile (with appropriate babel/typescript settings) to equivalent target of Node 8.8.1, i.e. with native async/await support, this fails with 

> UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: Meteor code must always run within a Fiber. Try wrapping callbacks that you pass to non-Meteor libraries with Meteor.bindEnvironment.

Obviously there's now no hook for calling `Meteor.bindEnvironment` correctly. I suppose this can be done with async-hooks in Node 9. I don't know of other ways to fix it.

What's your take? Bug? Feature request?

benjamn:
In order to run `async` functions and `Promise` callbacks in a `Fiber`, Meteor still does compile `async` functions for server-side code. However, the compilation should have a minimal impact on debugging `async` functions, unlike previous Babel compilation strategies.

A good way to see how that works is to put the following code in `server/main.js`:
```js
const pDelay = delay => new Promise(res => Meteor.setTimeout(res, delay));

(async function f(delay) {
  console.log(String(f)); // Show the actual code.
  await pDelay(delay);
  await pDelay(delay);
})(10);
```
The resulting code of `f` looks like this:
```js
function f(delay) {
  return Promise.asyncApply(() => {
    console.log(String(f)); // Show the actual code.
    Promise.await(pDelay(delay));
    Promise.await(pDelay(delay));
  });
}
```
[...]

PR #10585, opened 2019-06-17
Support full ECMAScript compilation of modules unexpectedly imported from node_modules.

benjamn:
In Meteor 1.8.2, supporting legacy browsers increasingly requires recompiling npm packages that use slightly-too-modern JavaScript syntax like `const` and `class` and arrow functions.

In #10545 and #10550 I added support for ECMAScript module syntax in `node_modules` (`import`, `export`, and dynamic `import()`), but now I realize that's not enough, since many more npm packages use simpler features like `class` syntax and arrow functions than use module syntax, without even realizing they are excluding older browsers.

Instead, this PR uses the standard Meteor compiler plugins system to compile "unanticipated modules," which is the same machinery Meteor applications normally use to compile application code, including the same on-disk and in-memory caching logic. Unanticipated modules are files discovered by the `ImportScanner` that were not originally included by `PackageSource#_findSources`, and thus not already compiled.

While it is possible to get Meteor to compile code imported from `node_modules` using various tricks such as [symlinking](https://forums.meteor.com/t/litelement-import-litelement-html/45042/8?u=benjamn), those techniques become more burdensome as more npm packages need to be compiled. If we want to continue claiming that Meteor automatically tailors separate client bundles for modern and legacy browsers (the [headline feature of Meteor 1.7](https://blog.meteor.com/meteor-1-7-and-the-evergreen-dream-a8c1270b0901)), then I think we need to make sure the legacy bundle "just works" by default, and worry about any performance implications of that extra compilation work after gathering real-world feedback.

Should solve #10563, and possibly also #9473, since this new logic applies to server code imported from `node_modules`, too.

### DIFF
--- a/History.md
+++ b/History.md
@@ -11,9 +11,9 @@ N/A
 * Node has been updated to version
   [8.16.0](https://nodejs.org/en/blog/release/v8.16.0/).
 
-* The `meteor-babel` npm package has been updated to version 7.4.12.
+* The `meteor-babel` npm package has been updated to version 7.4.13.
 
-* The `reify` npm package has been updated to version 0.20.4.
+* The `reify` npm package has been updated to version 0.20.5.
 
 * The `core-js` npm package used by `ecmascript-runtime-client` and
   `ecmascript-runtime-server` has been updated to version 3.1.4.
--- a/History.md
+++ b/History.md
@@ -11,7 +11,9 @@ N/A
 * Node has been updated to version
   [8.16.0](https://nodejs.org/en/blog/release/v8.16.0/).
 
-* The `meteor-babel` npm package has been updated to version 7.4.3.
+* The `meteor-babel` npm package has been updated to version 7.4.10.
+
+* The `reify` npm package has been updated to version 0.20.3.
 
 * The `core-js` npm package used by `ecmascript-runtime-client` and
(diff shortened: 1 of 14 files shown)


benjamn:
I expect some tests to fail with commit 23cc9031937927ce2ad67439f517a68d77a51918, since we're still waiting for https://github.com/babel/babel/pull/9864 to be released in `@babel/parser@7.5.0` (soon).

benjamn:
It turns out those remaining test failures were also legitimate. Because we run `meteor self-test` using both PhantomJS (a very old version of webkit, roughly Safari ~5) and Puppeteer (a headless, recent version of Chromium), the PhantomJS version was failing because Reify was generating modern syntax like `let` declarations, arrow functions, and method syntax. This syntax used to be compiled by other Babel plugins later in the pipeline, but Reify now runs in two passes, and the second pass happens after those other plugins run, so they no longer have a chance to compile the code generated by Reify. I was very close to disabling the PhantomJS tests because PhantomJS is ancient and arguably less relevant than ever, until I realized it was complaining about a real problem. Instead, I decided to make `reify@0.20.5` capable of generating pure legacy code that needs no further compilation. Thanks PhantomJS! 👻

PR #10592, opened 2019-06-21
Update dev-bundle-tool-package.js

logvik:
Need to update version "node-gyp" to "5.0.1", due to deprecated packages into 3.7.0


benjamn:
Thanks @logvik! This will require a new dev bundle, which will be included in the next beta of Meteor 1.8.2 (likely 1.8.2-beta.6).

PR #10596, opened 2019-06-25
Move meteorBabelHelpers to modules package, and restrict to legacy bundle.

### LINKED ISSUE #10595: Transpilation errors on IE11
I had a client who was trying to access a Meteor app using IE11. The app JS couldn't load due to errors which I'm guessing is related the transpilation for the legacy browser.

If I use Meteor **1.8.1**, I see the following error(s):
Screen Shot 2019-06-25 at 7 02 27 PM

If I trace the error I see this:
Screen Shot 2019-06-25 at 7 04 08 PM

So it's an error due to the const declaration within the [url-regex ](https://github.com/kevva/url-regex/blob/master/index.js) module. So I tried updating to **1.8.2** to see if [PR#10541](https://github.com/meteor/meteor/pull/10541) would solve it, which I think it did but then I ran into the errors below:

Screen Shot 2019-06-25 at 6 52 45 PM

Any ideas what might be happening here?

benjamn:
Should fix #10595.

Code from the application `node_modules` directory becomes part of the `modules` package, so that it can be imported by any other package that uses the module system, regardless of package load order.

Now that we compile code from `node_modules` using `babel-compiler` and `meteor-babel` (#10585), `node_modules` code requires the same runtime environment as any other Meteor JS code. For the most part, this need is satisfied by the `@babel/runtime/helpers/...` modules, which are also defined in the `modules` package because they come from `node_modules`. However, in the legacy bundle, `meteorBabelHelpers.sanitizeForInObject` is used to fix buggy for-in iteration in older Internet Explorers.

Thankfully, this extra helper code does not need to be included in the modern or server bundles, but only in legacy code.


benjamn:
By the way, I submitted this as a PR rather than just pushing to `release-1.8.2` because running tests in PhantomJS is a great way to verify any functionality specific to the legacy bundle.

Issue #9907, opened 2018-05-25
cannot pass METEOR_SETTINGS to meteor run


macrozone:
Because of https://github.com/meteor/meteor/blob/ac085d96fb79dfc03d28af8d4a3de50e99313ce4/tools/runners/run-app.js#L177

its not possible to pass `METEOR_SETTINGS` to `meteor` command. This prevents to pass METEOR_SETTINGS from something else but a file. E.g. we generate it from a secret password store.

Is it possible to bypass that?

My suggestion is to remove that else-case as it feels arbitrary for me. If someone would pass `METEOR_SETTINGS`-env-var to meteor, they usually know what to they do.

macrozone:
i also tried something like `mycommand | meteor --settings  -`, but this does not work either :-((

larswolter:
Passing meteor settings thru the environment is supported when running as a bundle. Supplying them in settings.json is just for development where you can easily handle multiple configurations using files. 

https://docs.meteor.com/api/core.html#Meteor-settings

Normally during development the settings should not contain any secret stuff. We would have problems using the environment during development, because multiple instances are usually run during development.

macrozone:
@larswolter I understand how `--settings settings.json` works and used that before. Also passing METEOR_SETTINGS via env vars seems to work perfectly, but its just actively prevented because of... well... "educational" reasons. 🤔

local development may require secret env vars, e.g. api keys to make request against other services. using env vars allows to use tools like https://direnv.net and with some scripting magic you can even pull env vars out of password store tools like `pass` or similar. Creating a json file "on-the-fly" is much harder.

macrozone:
no, bad stalebot!

there is a PR open, hope it gets merged

filipenevola:
thanks @macrozone, I already labeled as `confirmed` then stale bot will ignore it.

filipenevola:
This fix is going to be released on Meteor 2.4. You can already test it on latest RC #11488

macrozone:
The linked PR does not solve the issue. It simply adds an additional check for `env.NODE_ENV === 'development'`, which is true anyway in most cases as you will usually use the `meteor` command in development.

Goal is that you can use METEOR_SETTINGS env var as an alternative to `--settings`. 

@filipenevola should i create a new issue?

StorytellerCZ:
@macrozone feel free to create a PR

macrozone:
> @macrozone feel free to create a PR

[...]

PR #10605, opened 2019-07-03
feat: allow to pass METEOR_SETTINGS by env vars

macrozone:
solves https://github.com/meteor/meteor/issues/9907

meteor is doing a great job in following the "principle of least surprise", e.g. that changes in settings.json also trigger a restart automatically and therefore settings passed as an env var `METEOR_SETTINGS` will be ignored to guarantee this consistency.

However meteor-settings often contain secrets and environment specific values and therefore need to be managed the same way as other environment variables with special care, so that they do not leak into the git repository.

We therefore use a combination of [direnv](https://direnv.net) and a password store like [pass](https://www.passwordstore.org/) to manage and share these secrets among developers.

Unfortunatly, meteor prevents us to use the same technique also for meteor-settings, so this pull request tries to fix that by allowing to pass `METEOR_SETTINGS` as env var when you also pass an option `--allowSettingsFromEnv`

Personally, i would like to not have the flag allowSettingsFromEnv and delete the whole check there, because devs who pass the `METEOR_SETTINGS` env var usually know what they are doing. But that's debateable.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -307,6 +307,7 @@ var runCommandOptions = {
     production: { type: Boolean },
     'raw-logs': { type: Boolean },
     settings: { type: String, short: "s" },
+    'allow-settings-from-env': { type: Boolean },
     verbose: { type: Boolean, short: "v" },
     // With --once, meteor does not re-run the project if it crashes
     // and does not monitor for file changes. Intentionally
@@ -443,6 +444,7 @@ function doRunCommand(options) {
     cordovaServerPort: parsedCordovaServerPort,
     once: options.once,
     noReleaseCheck: options['no-release-check'] || process.env.METEOR_NO_RELEASE_CHECK,
+    allowSettingsFromEnv: options['allow-settings-from-env'],
     cordovaRunner: cordovaRunner
   });
 }
--- a/tools/cli/help.txt
+++ b/tools/cli/help.txt
@@ -500,6 +500,7 @@ Options:
   --delete, -D    permanently delete this deployment
   --debug         deploy in debug mode (don't minify, etc)
   --settings, -s  set optional data for Meteor.settings
+  --allow-settings-from-env   Allow to pass settings from the environment-variable METEOR_SETTINGS
   --allow-incompatible-update   Allow packages in your project to be upgraded or
                   downgraded to versions that are potentially incompatible with
                   the current versions, if required to satisfy all package version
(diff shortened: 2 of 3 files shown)


sebakerckhof:
I think few people would expect Meteor to react on changes in envars, so I never understood this limitation, which has also been an annoyance to us. Therefore I think it's also not even necessary to have the `allowSettingsFromEnv` option.

One possible workaround/hack while this PR is waiting, is to assign it yourself via another envar:
E.g. somewhere in the server side code:
`Meteor.settings = JSON.parse(process.env.MY_METEOR_SETTINGS)`

ramijarrar:
@macrozone We use [transcrypt](https://github.com/elasticdog/transcrypt) to check settings files along with other secrets e.g code signing certs securely in version control (we can still do code diffs and fairly straight-forward to rekey when we have staff changes).

This seems like a useful addition though.

macrozone:
> @macrozone We use [transcrypt](https://github.com/elasticdog/transcrypt) to check settings files along with other secrets e.g code signing certs securely in version control (can still do code diffs and fairly straight-forward to rekey when we have staff changes).
> 
> This seems like a useful addition though.

looks also really nice! but i don't want to change everything just because meteor is that educational about how to use settings. METEOR_SETTINGS through env has to be possible locally

macrozone:
> I think few people would expect Meteor to react on changes in envars, so I never understood this limitation, which has also been an annoyance to us. Therefore I think it's also not even necessary to have the `allowSettingsFromEnv` option.
> 
> One possible workaround/hack while this PR is waiting, is to assign it yourself via another envar:
> E.g. somewhere in the server side code:
> `Meteor.settings = JSON.parse(process.env.MY_METEOR_SETTINGS)`

i also agree that `allowSettingsFromEnv` should not be needed, i hope @benjamn or someone from mdg also agrees and this change get merged for meteor 1.8.2

> `Meteor.settings = JSON.parse(process.env.MY_METEOR_SETTINGS)`

nice workaround, but probably does not work on the client side and depends on the order of execution

filipenevola:
The CLA is still not signed here, because of that I'm closing this PR. Feel free to continue the work in a new PR as we can't merge this one without the CLA.

macrozone:
> The CLA is still not signed here, because of that I'm closing this PR. Feel free to continue the work in a new PR as we can't merge this one without the CLA.

i signed it, but it seems it did not work

Issue #10613, opened 2019-07-10
Help getting up and running to port Meteor tools to TypeScript


trusktr:
I gotta earn this water bottle! A bag would be nice too.

I just cloned Meteor, checked out the `release-1.8.2` branch, converted `archinfo.js` to `archinfo.ts`, then realized there were no types available, but I also didn't see an obvious package.json in order to run `npm install` for utils (I do see a ton of them in other tools folders).

So I tried running `./meteor` to set things up (is this right?).

Ironically, I got the error

```
release-1.8.2 ❯ ./meteor
It's the first time you've run Meteor from a git checkout.
I will download a kit containing all of Meteor's dependencies.
######################################################################## 100.0%
Installed dependency kit v8.16.0.21 in dev_bundle.

module.js:550
    throw err;
    ^

Error: Cannot find module './archinfo.js'
    at Function.Module._resolveFilename (module.js:548:15)
    at Function.Module._load (module.js:475:25)
    at Module.require (module.js:597:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/tools/utils/utils.js:7:16)
    at Module._compile (module.js:653:30)
    at Module.Mp._compile (/tools/static-assets/server/runtime.js:41:21)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/trusktr/src/meteor+meteor/dev_bundle/lib/node_modules/meteor-babel/register.js:59:14)
    at Module.load (module.js:566:32)
    at Module.Mp.load (/tools/static-assets/server/runtime.js:13:29)
```

because, I renamed archinfo to .ts. :)

So I renamed that back, ran `./meteor`, and `dev_bundle/lib/node_modules` exists.

But now I'm in archinfo.ts (renamed it again), and it says there's no `underscore` module.

It's not a standard TS workflow, so I'm not sure how I should proceed.

Floriferous:
It would actually be really helpful to have a `TypeScript migration guide for Meteor`, which has a checklist of things to do in each file. And since people are almost rewriting entire files, it could be helpful to migrate old code as well.

Maybe this can also help set the scope of the changes, for example:

Checklist: 
1. Add types to each variable
1. Add types to each function
    * "Here's what a good type name looks like in this repo: ..."
1. Rewrite all require statements as `import` statements
1. Rewrite all `var` as `let` or `const`
1. Explicitly import as many global variables as possible (#10560)
1. etc.

Instructions:
1. Only convert a single file per PR
1. etc.

[...]

PR #10614, opened 2019-07-10
Convert eachline.js to eachline.ts

menewman:
[no issue]

Hi @benjamn! Michael Newman here, no relation.

Wanted to give Meteor TypeScript conversion a shot after attending [your excellent Meteor Night talk](https://slides.com/benjamn/reimplementing-meteor-in-typescript#), so here's my shot at one of the easy files to start with you mentioned in your presentation.

I had a few questions about getting started, since the codebase/tooling of this repo is pretty unfamiliar to me:
1. Is the correct way to check for TS errors in this project just to open the project in VSCode and see if it gets angry? (That's what I did and it seemed to work... okay.)
2. For whatever reason, TypeScript lets me get away with code like `const pipe = require("multipipe");`	but not `import pipe from "multipipe";`, because for the latter version it wants me to install the type defs for that package via NPM. I have no idea how to add new NPM dependencies to `meteor/tools` and I'm not sure whether it's advised/allowed even for TS types. What do you recommend we do in cases like that?
3. I submitted this into the `release-1.8.2` branch because that's where your "convert files.ts" PR was targeted, but I'm not sure if this was the right thing to do, since the Contributing guides seem to encourage submitting into `devel`. However, `devel` doesn't seem to have all of your TS-related changes from `release-1.8.2`.

Thank you for your talk and for your patience with my first Meteor PR!

### DIFF
--- a/tools/utils/eachline.ts
+++ b/tools/utils/eachline.ts
@@ -1,19 +1,22 @@
-import split from "split2";
-import pipe from "multipipe";
-import { Transform } from "stream";
+const split = require("split2");
+const pipe = require("multipipe");
 
-export function eachline(stream, callback) {
+import { Transform, Stream } from "stream";
+
+type LineTransformer = (line: string) => string | Promise<string>
+
+export function eachline(stream: Stream, callback: LineTransformer) {
   stream.pipe(transform(callback));
 }
 
-export function transform(callback) {
+export function transform(callback: LineTransformer) {
   const splitStream = split(/\r?\n/, null, {
     trailing: false
   });
 
   const transform = new Transform();
 
-  transform._transform = async function (chunk, encoding, done) {
+  transform._transform = async function (chunk, _encoding, done) {
     let line = chunk.toString("utf8");
     try {
       line = await callback(line);
--- a/tools/cordova/run-targets.js
+++ b/tools/cordova/run-targets.js
@@ -132,7 +132,7 @@ export class AndroidRunTarget extends CordovaRunTarget {
   }
 
   async tailLogs(cordovaProject, target) {
-    const { transform } = require("../utils/eachline.js");
+    const { transform } = require("../utils/eachline");
 
     cordovaProject.runCommands(`tailing logs for ${this.displayName}`, async () => {
       await this.checkPlatformRequirementsAndSetEnv(cordovaProject);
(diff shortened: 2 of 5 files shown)


tools/utils/eachline.ts
```diff
@@ -1,19 +1,22 @@
-import split from "split2";
-import pipe from "multipipe";
-import { Transform } from "stream";
+const split = require("split2");
```
hexsprite: can't you keep using import by doing `import * as split from "split2"` for example?

menewman: Hi, thanks for commenting!

If I do that, I get a TS error like this:
```
Could not find a declaration file for module 'split2'. '/Users/newman/Code/meteor/tools/node_modules/split2/index.js' implicitly has an 'any' type.
  Try `npm install @types/split2` if it exists or add a new declaration (.d.ts) file containing `declare module 'split2';`
```

This would be easy to resolve if I had any idea of how to install `@types/split2` in this particular repo. But, `meteor/meteor` is unique and unfamiliar to me with its own special packaging, so I'm not sure the "proper" way to add a new NPM package here (where is my `package.json`???) -- hence my question in the PR body.

hexsprite: Oops, sorry I didn't read the original PR body, just was perusing the code changes! Great contribution!

I'm not an expert on Meteor core development but I am endless curious so here's some of my findings...

Normally you should be able to add an npm dependency for any Meteor  package by adding it to the `package.js` and using `Npm.depends` to specify the specific version.

But using good ole `grep -r` I noticed that `split2` is defined in a couple places where it seems like Meteor generates its top-level `package.json`.

```
scripts/dev-bundle-tool-package.js
53:    split2: "2.2.0",

scripts/dev-bundle-server-package.js
24:    split2: "2.1.1",

tools/utils/eachline.js
1:import split from "split2";
```

Presumably you could add that there and regenerate the dev bundle?

menewman: Thanks for the avenue of investigation! I was kind of hoping that someone who is an expert at Meteor core development would swoop in with a guideline on installing `@types/` NPM packages for this TypeScript conversion effort, haha, since I think it will be helpful for many of the files under `meteor/tools`.

(Related: #10613)

tools/utils/eachline.ts
```diff
@@ -1,19 +1,22 @@
...
+const pipe = require("multipipe");
```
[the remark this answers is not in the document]
menewman: Thank you for the guidance and the review! Makes sense about the imports.

PR #10619, opened 2019-07-11
Convert shell-client.js to shell-client.ts

afrokick:
Ok, just tried to convert shell-client to ts.

I have same questions mentioned in #10614 about level branch, VS Code' settings, typedefs.

And I have an issue when compile a file:
```
mb:tools afrokick$ tsc shell-client.ts
shell-client.ts:112:33 - error TS2345: Argument of type 'Buffer' is not assignable to parameter of type 'string'.
112       const output = JSON.parse(>>Buffer.concat(outputBuffers)<<);
```
How to properly compile it?


tools/shell-client.ts
```diff
@@ -0,0 +1,221 @@
+import * as assert from "assert";
+import * as fs from "fs";
+import * as path from "path";
+import * as net from "net";
+import { isEmacs } from "./utils/utils";
+//@ts-ignore
```
afrokick: Ignored, because:
```
Could not find a declaration file for module 'chalk'.
'/git/meteor/tools/node_modules/chalk/index.js' implicitly has an 'any' type.
Try `npm install @types/chalk` if it exists or add a new declaration (.d.ts) file containing `declare module 'chalk';`ts(7016)
```

afrokick: chalk 2.2.0+ has type definitions file https://github.com/chalk/chalk/releases/tag/v2.2.0

should we update chalk?

benjamn: I’m going to keep the `require` style for now, per https://github.com/meteor/meteor/pull/10614#discussion_r302156889.

benjamn:
Here’s where we override the type for `JSON.parse` to accept `Buffer`s as well as strings: https://github.com/meteor/meteor/blob/release-1.8.2/tools/index.d.ts#L4-L7

If you run
```sh
cd path/to/meteor/tools
../meteor npx tsc
```
instead of just checking the `shell-client.ts` module in isolation, the `index.d.ts` file will be picked up, and the errors will go away.

tools/shell-client.ts
```diff
@@ -22,18 +22,14 @@ export function connect(shellDir: string) {
...
-  constructor(shellDir: string) {
-    assert.ok(this instanceof Client);
+  private exitOnClose = false;
+  private firstTimeConnecting = true;
+  private reconnectCount = 0;
+  private reconnectTimer?: NodeJS.Timeout;
 
-    this.shellDir = shellDir;
-  }
+  constructor(public shellDir: string) {}
```
benjamn: One of my favorite TypeScript features: you can declare instance variables with appropriate visibility and types right in the parameter list of the constructor!

afrokick: Parameter properties)

Could you please provide a feedback in https://forums.meteor.com/t/request-typescript-code-style/49498 about Code Style?

PR #10624, opened 2019-07-11
Refactor archinfo.js to typescript

pmogollons:
* Updated code to use modern JS
* Added types
* Stopped using 2 underscore functions (1 remaining)
* TS complains about execFileSync not being exported. I think it requires the update of that file too.


tools/utils/archinfo.ts
```diff
@@ -1,7 +1,7 @@
-var _ = require('underscore');
-var os = require('os');
+const { max } = require('underscore');
```
nicu-chiciuc: As I understand, the reason this wasn't transformed into an import statement is because there wasn't a type definition for `underscore`. A better way to solve this would be to install `@types/underscore`. The local `tools/node_modules` is a symlink to `dev_bundle/lib/node_modules` so I'm currently trying to find which package.json should be changed in order to add the package.

Since `dev_bundle` changes can be made just by the MDG maybe @benjamn or someone else could help with this. I'm currently trying to figure out which particular `package.json` has to be changed.

pmogollons: Sure, but the main reason for not going that way was taking a look to a file that @benjamn already refactor to TS and was requiring underscore instead of importing it and installing the types.

[tools/isobuild/resolver.ts](https://github.com/meteor/meteor/blob/release-1.8.2/tools/isobuild/resolver.ts)

But installing the types should be the way to go. Or even better just removing underscore completely. But that requires some additional work.

nicu-chiciuc: Good catch, I'm still new to the codebase so it's hard to understand the reasoning behind leaving the `require`.

pmogollons:
Awesome. Thanks.

PR #10627, opened 2019-07-12
Convert tools/utils/processes.js to tools/utils/processes.ts

menewman:
[no issue]

Another TypeScript conversion, this one a little bit bigger but still a pretty small file.

As with prior PRs #10614 and #10625, I hope I correctly guessed which branch to target!

This branch was relatively easy to type since it already had JSDoc-style annotations (slightly incorrect but still very helpful) and only one external untyped dependency (underscore, which was easy to remove).


tools/utils/processes.ts
```diff
@@ -11,58 +11,81 @@ import files from '../static-assets/server/mini-files';
...
+   * If specified, instead of capturing the output, the child process stdout
+   * will be piped to the destination stream.
+   */
+  destination?: Writable,
+
+  /**
+   * Whether to wait for the child process streams to close or to resolve
+   * the promise when the child process exits.
+   */
+  waitForClose?: boolean,
```
benjamn: This style of options documentation is _so much better_!

PR #10631, opened 2019-07-14
Convert `func-utils.js` to typescript

nicu-chiciuc:
This is part of the ongoing transformation to typescript.
The pull should be accepted only after adding a `@types/underscore` to the node_modules, as mentioned [in pull request 10624](https://github.com/meteor/meteor/pull/10624#discussion_r302651681). Currently I've installed the types locally.

Also, I've added some ad-hoc types e.g. `EmptyFunction` and `AnyFunction`. After the transformation will mature we can use a library for stuff like this or have a centralised way where these type shuld be defined.

The global `Function` interface was ammended to include the `displayName` property.

### DIFF
--- a/tools/index.d.ts
+++ b/tools/index.d.ts
@@ -12,4 +12,9 @@ declare global {
     // moved into that package.
     await: () => T;
   }
+
+  interface Function {
+    // func-utils.ts makes usage of this feature
+    displayName?: string;
+  }
 }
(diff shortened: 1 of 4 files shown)


nicu-chiciuc:
Is it possible to rerun the AppVeyor check without commiting anything?

tools/index.d.ts
```diff
@@ -12,4 +12,9 @@ declare global {
     // moved into that package.
     await: () => T;
   }
+
+  interface Function {
+    // func-utils.ts makes usage of this feature
+    displayName?: string;
```
benjamn: Nice catch.

PR #10637, opened 2019-07-17
Convert fiber-helpers to ts

afrokick:
* refactoring `makeFulfillablePromise` method
* add `resolve` and `reject` methods into the declaration file for Promise interface


afrokick:
I have an issue with Map:
```
TypeError: Method Map.prototype.has called on incompatible receiver #<Map>
    at Map.has (<anonymous>)
    at EnvironmentVariable.get (/tools/utils/fiber-helpers.ts:72:23)
```
This is a code:
```
if (!Fiber.current._meteorDynamics.has(this.slot)) {
  return this.defaultValue;
}
```

@benjamn did you meet it?

afrokick:
> I have an issue with Map:
> 
> ```
> TypeError: Method Map.prototype.has called on incompatible receiver #<Map>
>     at Map.has (<anonymous>)
>     at EnvironmentVariable.get (/tools/utils/fiber-helpers.ts:72:23)
> ```
> 
> This is a code:
> 
> ```
> if (!Fiber.current._meteorDynamics.has(this.slot)) {
>   return this.defaultValue;
> }
> ```
> 
> @benjamn did you meet it?

Found issue in `dev_bundle/lib/node_modules/meteor-promise/promise_server.js#shallowClone`. It works only with `Array` and `Object`. With `Map` it fails.

I think it should be fixed in a separate PR if needs.

Now I just revert to object as before

afrokick:
Fix shallowCopy https://github.com/meteor/promise/pull/24 @benjamn

tools/utils/fiber-helpers.ts
```diff
@@ -126,18 +125,22 @@ exports.bindEnvironment = function (func) {
...
   };
 };
 
 // Returns a Promise that supports .resolve(result) and .reject(error).
-exports.makeFulfillablePromise = function () {
-  var resolve, reject;
-  var promise = new Promise(function (res, rej) {
+export function makeFulfillablePromise() {
+  let resolve: (value?: any) => void = () => { };
+  let reject: (reason?: any) => void = () => { };
```
nicu-chiciuc: I didn't check where these type of promises are used but I'm not sure if changing and undefined variable for a noop maintains the assumptions about the `resolve` and `reject` properties. For example, what if a piece of code would rely on checking whether the `resolve` is defined on a particular promise.

afrokick: Thanks. I fixed it. But maybe you know more graceful solution?

nicu-chiciuc: Unfortunately I've also had to make transformations that are not elegant just to make sure that I don't break anything. For example:
```
function coalesce(delay) {
   delay = delay || 100
  // ... use delay
}
```
was transformed to:
```
function coalesce(delay: number) {
   const actualDelay = delay || 100
  // ... use actualDelay
}
```
instead of:
```
function coalesce(delay: number = 100) {
  // ... use delay
}
```
since the behavior would be changed if coalesce would get a `0` or a `null` instead of undefined. These are small things, and most of the time there's no difference, but (since I don't know the codebase very well) I wanted to stay on the safe side.

tools/utils/fiber-helpers.ts
```diff
@@ -1,9 +1,8 @@
-var _ = require("underscore");
-var Fiber = require("fibers");
+import _ from "underscore";
```
trusktr: You don't get a TypeScript error telling you underscore module is not defined here?

nicu-chiciuc: @trusktr Your `/tools/node_modules/@types` shoud contain a `underscore` folder for type definitions. It was added recently in commit `122f7e4d2d85`.

tools/utils/fiber-helpers.ts
```diff
@@ -89,34 +87,35 @@ _.extend(exports.EnvironmentVariable.prototype, {
...
 
-  return function (...args) {
-    var self = this;
+  return function (...args: any[]) {
+    //@ts-ignore
+    const self = this;
 
-    var runWithEnvironment = function () {
-      var savedValues = Fiber.current._meteorDynamics;
+    const runWithEnvironment = () => {
```
jamesmillerburgess: Since you’ve turned this into an arrow function, I think you should be able to safely remove the `self` variable and the `//@ts-ignore`.

afrokick:
@benjamn @filipenevola  ready to review

tools/utils/fiber-helpers.ts
StorytellerCZ: Most of underscore has been removed already. Can you get rid of the rest as well?

filipenevola:
Hi @afrokick could you adjust the code based on the last reviews?

tools/utils/fiber-helpers.ts
```diff
@@ -1,9 +1,8 @@
...
+import _ from "underscore";
+const Fiber = require("fibers");
 
-exports.parallelEach = function (collection, callback, context) {
-  const errors = [];
-  context = context || null;
+export function parallelEach(collection: any, callback: Function, context: any | null = null) {
+  const errors: Error[] = [];
 
   const results = Promise.all(_.map(collection, (...args) => {
```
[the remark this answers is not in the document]
RobertLowe: @afrokick in the one use is `packagesToDownload` ever even an array anymore https://github.com/meteor/meteor/blob/982181fff86e09a75f1d2681a78b9d784e666f4f/tools/packaging/warehouse.js#L390-L391

Time for some clean up?

filipenevola:

PR #10641, opened 2019-07-21
WIP: Refactor source-arch to typescript

pmogollons:


tools/isobuild/source-arch.ts
```diff
@@ -1,28 +1,64 @@
-import {isString, isFunction} from 'underscore';
-import {WatchSet} from '../fs/watch';
+import { isString, isFunction } from 'underscore';
```
jamesmillerburgess: Might as well take this opportunity to remove underscore.

tools/isobuild/source-arch.ts
```diff
@@ -1,28 +1,64 @@
...
+import { WatchSet } from '../fs/watch';
 
-function reportMissingOption(name) {
+
+function reportMissingOption(name: string) {
```
jamesmillerburgess:
```suggestion
function reportMissingOption(name: string) never {
```

PR #10650, opened 2019-07-23
Convert tools/console/progress.js to TypeScript

menewman:
Converts the "progress" class to TypeScript! This conversion is a little bigger because I refactored the class to use `class` syntax and private members.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -32,7 +32,7 @@ tools/http-helpers.js
 tools/inspector.js
 tools/index.js
 tools/mongo-exit-codes.ts
-tools/processes.js
+tools/processes.ts
 tools/progress.ts
 tools/project-context.js
 tools/runners/run-log.js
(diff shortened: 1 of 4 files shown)


menewman:
No longer WIP! Sorry for the big diff; I think I changed enough lines that rename detection is failing.

menewman:
Oops, I'm so used to using Flow (which doesn't have private methods) that I didn't even consider making those underscore-prefixed methods private in TypeScript. Thanks, I'll pay more attention to that in future PRs!

.eslintignore
```diff
@@ -32,8 +32,8 @@ tools/http-helpers.js
 tools/inspector.js
 tools/index.js
 tools/mongo-exit-codes.ts
-tools/processes.ts
-tools/progress.js
+tools/processes.js
```
menewman: @benjamn - I think this should stay `.ts`, probably just a weird merge conflict resolution after merging the other PR

PR #10653, opened 2019-07-27
Convert parse-stack.js to TypeScript

menewman:
A small TypeScript conversion. For `markTop` and `markBottom` I took the coward's way out and just used `@ts-ignore` so that I wouldn't have to "fix" the usage of implicit `this`, since I wasn't confident that no consumers relied on that behavior (although I couldn't find any specific examples where it was needed).


tools/utils/parse-stack.ts
```diff
@@ -110,28 +137,31 @@ function parseStackFrames(frames) {
...
         return stop = true;
       }
-      ret.push({
+
+      parsedFrames.push({
         func: m[1],
         file: m[5],
-        line: m[7] ? +m[7] : undefined,
-        column: m[9] ? +m[9] : undefined
+        line: m[7] ? +m[7] : null,
```
jamesmillerburgess: 🆗 I wondered if these changes from `undefined` to `null` would be safe, and it looks ok, because the checks in `buildmessage.js` are for thruthiness:

https://github.com/meteor/meteor/blob/5c32331591f268f2acdac5a1a7481eb9e65dc3d4/tools/utils/buildmessage.js#L68-L75

https://github.com/meteor/meteor/blob/5c32331591f268f2acdac5a1a7481eb9e65dc3d4/tools/utils/buildmessage.js#L83-L85

menewman: thanks for double-checking and confirming!

tools/utils/parse-stack.ts
```diff
@@ -147,14 +177,14 @@ function parseStackFrames(frames) {
       return stop = true;
     }
 
-    if (_.isEmpty(ret)) {
+    if (!parsedFrames.length) {
```
MarceloPrado, line 180: Hi @menewman! If you allow me, I'll like to make a suggestion. What do you think of the following change:

```suggestion
    if (parsedFrames.length === 0) {
``` 


I think this makes the code more explicit to follow.

trusktr: Just curious, how do you make a suggested change? I've never seen that feature before.

tools/utils/parse-stack.ts
trusktr: I like to manually use `null` when I'm marking something for falsyness while otherwise leaving `undefined` as a feature of the language that tells me when something is _actually undefined_ and never manually using `undefined` if I can avoid it. It's a semantic difference.

tools/utils/parse-stack.ts
menewman: Like this!

Screen Shot 2020-04-16 at 3 04 17 PM

Anyway, @MarceloPrado, I'm fine with that change. To me, checking for falsey length in JS is pretty idiomatic but I recognize everyone might not feel that way, and this certainly isn't something to bikeshed over.

tools/utils/parse-stack.ts
menewman: Yes, I feel the same way.

PR #10654, opened 2019-07-28
Convert profile-require.ts to TypeScript

menewman:
Converted `profile-require.js` to TypeScript. This is a file that summarizes time spent "requiring" things during a run if the given environment env variable is set.

To test:
Set this env variable:
```sh
export METEOR_PROFILE_REQUIRE=true
```
Then run some Meteor app using your local Meteor version. You should see a list that looks like this (excerpted):
```
29.00 TOP
14.00 @babel/parser [via TOP:warehouse.js]
12.00 npm [via TOP:warehouse.js]
10.00 acorn [via TOP:warehouse.js]
8.00 sqlite3 [via TOP:warehouse.js]
7.00 npmlog [via TOP:warehouse.js]
...
TOTAL: ours 30.00, other 502.00, grand total 532.00
```


menewman:
No longer WIP.

tools/tool-env/profile-require.ts
```diff
@@ -0,0 +1,185 @@
...
+};
+
+type Summary = Record<string, {
+  name: string;
+  time: number;
+  ours: boolean;
+  via: Record<string, boolean>;
+}>;
+
+function summarize(inv: RequireInvocation, depth: number, summary: Summary = {}) {
```
jamesmillerburgess: The functions don't have return types, which is mostly fine since they don't return anything. But this one does return something, so maybe we should annotate it. Or is it better just to let TS figure it out through static analysis?

tools/tool-env/profile-require.ts
```diff
@@ -0,0 +1,185 @@
...
+    summarize(inv, depth + 1, summary);
+  })
+
+  return summary;
+};
+
+/**
+ * Formats time in seconds for display in milliseconds.
+ */
+function formatTime(time: number) {
```
jamesmillerburgess: This one as well returns a value, so we should annotate it too.

tools/tool-env/profile-require.ts
```diff
@@ -0,0 +1,185 @@
...
+  const times = Object.values(summary).sort((a, b) => 
+    b.time - a.time
+  );
+
+  let ourTotal = 0, otherTotal = 0;
+
+  times.forEach(item => {
+    let line = `${formatTime(item.time)} ${item.name}`;
+
+    if (!item.ours) {
```
jamesmillerburgess: Sometimes there is a blank line above `if` blocks and other times not. The style guide only says that there should be blank lines after blocks and before new statements: https://github.com/meteor/javascript/#18.6

Since you are converting existing code, I'll just leave this one comment rather than leave 100 comments about where this is inconsistent. But it might be a good opportunity to correct those inconsistencies now:

- Missing blank lines after blocks
- Extra blank lines after statements

tools/tool-env/profile-require.ts
menewman: Although I do sometimes prefer to add an explicit return type (as I did in the class methods of `RequireInvocation`), personally I don't think it's the kind of thing that ought to be a strict rule.

For relatively simple functions that aren't exported from a module, if `TypeScript` is able to infer the type, then it's also very easy to see the inferred return type in VSCode, so there's no real loss of "documentation" information either way. In that case, adding the return type just puts a little bit of clutter in the source code that doesn't really add any additional value.

The main reason why I might want to add a return type signature is to be extra sure that the code inside really returns what I think it returns, even if I start refactoring it, but the stakes are quite low for a function like this (that just returns one of its parameters, which is typed) or the one below (which is a trivial one-liner that converts a timestamp number into a string), because if the return type changes then the usage (which is limited to this file) will also start to throw TypeScript errors if the types are no longer compatible.

Of course, I wouldn't mind adding a return type (it is a very easy change to make), but the point I'm making is that I don't think it should be a "you must always do this 100% of the time" type of rule. I am happy to be overridden by any member of the Meteor team who would like to lay down the law on style guidelines, though -- if they want us to start adding return types everywhere, then of course I will!

tools/tool-env/profile-require.ts
```diff
@@ -0,0 +1,185 @@
+// Use path instead of files.ts here because we are explicitly trying to
+// track requires, and files.ts is often a culprit of slow requires.
+import path from 'path';
```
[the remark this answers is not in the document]
menewman: Since there seems to be ample usage of both kinds of string quote in the codebase, I was unsure about this and decided to just leave the string the way it was (i.e., as a single string).

Before:
```js
var path = require('path');
```

Actually, the JS style guide you linked below says it should be single quotes, although I do agree with you that there are a lot of double quotes in the codebase:
https://github.com/meteor/javascript/#strings

That said, I don't think this kind of thing affects readability of the code at all or is worth bikeshedding over even a little bit! I'm happy to adjust to whatever the Meteor team wants (preferably, they would just install Prettier 😅 ) but am trying to just leave things the way they are where it's not clear what they want.

tools/tool-env/profile-require.ts
menewman: Hi @jamesmillerburgess ! Thank you for reading through my pull request and for your thoughtful comments.

I agree with the impulse to clean up code (ideally to meet lint and coding style guidelines) as we convert to TypeScript. That said, I honestly don't think we should require ourselves to make existing non-compliant code 100% style guide compliant if it's something that the Meteor team already wasn't prioritizing and if it wasn't already being enforced by a tool, because it means we spend less type actually converting to TypeScript, and more time worrying about clean-up that could hopefully be automated with ESLint/fix and Prettier.

I think most of us are contributing TypeScript conversion PRs as "volunteer" work while also holding full-time jobs, which means our time is a limited resource. And if we can spend more of that time adding statically enforced types and less worrying about cleaning up the style of pre-existing code that already didn't adhere to Meteor's own coding style guide, I think that's a good tradeoff.

As with my other comments, I am obviously willing to be overridden by any member of the Meteor team on this! This is just my subjective opinion, and I grant that it may be controversial. But, adding types is something I'm willing to spend my free time doing because I enjoy it and because I think it's something that's helpful for the Meteor project, which I hope to see succeed (because it adds documentation and type guarantees to the codebase, making it more accessible and hopefully less bug-prone). Adding and removing newlines to old code... to be honest, that's probably not something I'm quite as willing to spend my free time on!

Anyway, my goal isn't to be dismissive here -- I appreciate your comments and pointing out the style guide, and I know that we both have positive intentions. I just want to give some feedback on why I think it definitely would not be productive to leave 100 comments about inconsistent newline placement in existing code (thank you for not doing that).

---

With respect to the specific comment about newlines, I do have this comment to make, though: The fact that the style guide says "leave a blank line after blocks and before the next statement" does _not_ mean "never use newlines in any other circumstance."

Newlines should be used when it's helpful in creating semantic blocks of code, which sometimes means in between statements.

Point taken about missing blank lines after blocks; I will go ahead and update that in this file.

jamesmillerburgess: Thanks for not taking it personally. I'm not saying I'm right, just leaving points to consider. And I didn't see any real issues with the conversion of this file, so good job!

I wanted to review some of these TypsScript PRs, because they had been here for a while without comment, and I hope we can build some momentum around this conversion.

And thinking about how this TypeScript conversion was presented in @benjamn's Meteor Night talk, the spirit here of this effort is not to clean up every last thing, but to make the code accessible to more developers by clarifying some of the object types that get thrown around a lot.

tools/tool-env/profile-require.ts
jamesmillerburgess: I could see arguments both for and against annotating everything, but looking at the Apollo codebase, it looks like return types are frequently omitted, even from exported functions. We could go with that as guidance, absent any explicit rule for Meteor.

tools/tool-env/profile-require.ts
menewman: At the end of the day, I know that we both want to see this effort succeed and have Meteor keep on rockin', which is why after I finished writing these long-winded comments, I went ahead and made the changes anyway. 😄 Just wanted to add my POV on why I think it may be okay to not do full cleanup on every file that we convert.

I totally agree about building momentum -- I've been holding off on doing more conversions while waiting for feedback on the ones that are outstanding.

menewman:
Closing since it seems like TypeScript conversion isn't a prio rn and I haven't updated the PR branch in a long time

filipenevola:
Hi @menewman, you are right, it's not a top priority atm but I'm glad to review and merge PRs that are ready!

PR #10662, opened 2019-08-03
Convert dev-bundle.js to TypeScript

jamesmillerburgess:
I did this before reading the comment at the top of the file 😅 :

https://github.com/meteor/meteor/blob/5c32331591f268f2acdac5a1a7481eb9e65dc3d4/tools/cli/dev-bundle.js#L1-L2

As well, the tests don't run, so guess the TypeScript compiler is also not running here?

So I'm not sure if we should even be converting this, but I'm making a PR since I already did it.

There is one use of the `any` type: the `data` parameter in the sqlite3 callback. This is the same type as in `@types/sqlite3`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bb36642c70f366af454e1bf67471fbfdf0b9bdc8/types/sqlite3/index.d.ts#L65

### DIFF
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -38,6 +38,7 @@ var packageJson = {
     kexec: "3.0.0",
     "source-map": "0.5.7",
     chalk: "0.5.1",
+    "@types/sqlite3": "3.1.5",
     sqlite3: "3.1.8",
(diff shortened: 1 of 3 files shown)


skirunman:
@filipenevola Would you mind adding label:Project:TypeScript-conversion  to this issue, thanks.

filipenevola:

PR #10663, opened 2019-08-04
Convert npm-discards.js to TypeScript

jamesmillerburgess:
This is a dependency to some other npm-related files, so doing this one first.


tools/isobuild/npm-discards.ts
```diff
@@ -0,0 +1,93 @@
...
+    return false;
+  }
+}
+
+function merge(into: Discards, from: Discards) {
+  Object.keys(from).forEach((packageName: string) => {
+    const fromValue = from[packageName];
+    const intoValue = hasOwn.call(into, packageName) && into[packageName];
+
+    if (typeof fromValue === "string" || fromValue instanceof RegExp) {
```
jamesmillerburgess: I'm not sure the code in this `if` block is reachable, since the values are always supposed to be arrays, according to this comment in `package-npm.js`:

```javascript
  // The `discards` parameter should be an object whose keys are
  // top-level package names and whose values are arrays of strings
  // (or regular expressions) that match paths in that package's
  // directory that should be stripped before installation. For
  // example:
  //
  //   Npm.strip({
  //     connect: [/*\.wmv$/],
  //     useragent: ["tests/"]
  //   });
```

But I wanted to check before deleting it.

jamesmillerburgess: For what it's worth, I logged the results in the `if` and `else` blocks while building our production application and got 8 arrays and 0 string/RegExps:

```javascript
console.log("I'm an array! ", fromValue, from);
```

```
I'm an array!  [ 'test/' ] { 'es5-ext': [ 'test/' ] }
I'm an array!  [ 'test/' ] { 'es5-ext': [ 'test/' ] }
I'm an array!  [ 'test/' ] { mongodb: [ 'test/' ] }
I'm an array!  [ 'test/' ] { mongodb: [ 'test/' ] }
I'm an array!  [ 'test/' ] { multiparty: [ 'test/' ], useragent: [ 'test/' ] }
I'm an array!  [ 'test/' ] { multiparty: [ 'test/' ], useragent: [ 'test/' ] }
I'm an array!  [ 'test/' ] { multiparty: [ 'test/' ], useragent: [ 'test/' ] }
I'm an array!  [ 'test/' ] { multiparty: [ 'test/' ], useragent: [ 'test/' ] }
```

jamesmillerburgess: I looked through all 19 pages of GitHub results for `npm.strips` in `package.js` files and all of them use the array syntax:

https://github.com/search?l=&p=19&q=%22npm.strip%22+filename%3Apackage.js&ref=advsearch&type=Code&utf8=✓

jamesmillerburgess: After thinking about this some more, maybe it would be better to support both formats, but then I should update the misleading comment about `NPM.strip` and also update the type for `Discards` to accommodate both.

skirunman:
@filipenevola Would you mind adding label:Project:TypeScript-conversion  to this issue, thanks.

tools/isobuild/npm-discards.ts
MarceloPrado: Hi @jamesmillerburgess, regarding these comments, why did you decide to keep support for the string syntax? I also looked at the source and it's not being used (neither documented).

Any reasons for keeping this `if` block?

jamesmillerburgess: It would be a breaking change, though it isn't clear if any meteor packages use this syntax.

PR #10686, opened 2019-09-05
Avoid restarting server in development when changed file(s) not used by server bundle.

### LINKED ISSUE #10449: [1.8.1-beta.16] #10414 results in server rebuilds on client code changes in packages
#10414 in the current 1.8.1 beta results in client/server directories not treated special anymore.
We structured our packages with client and server directories and changing a file in a client directory triggers a client rebuild only.
After #10414 the server gets also rebuild.

A small reproduction:
```
meteor create --full test
cd test
meteor create --package custom
meteor add custom
```
change packages/custom/package.js to load different mainModules
```
  ...
  api.mainModule('client/client.js', 'client');
  api.mainModule('server.js', 'server');
  ...
```
create files packages/custom/client/client.js 
```
  import '../custom';
  
  console.log('Hello Client!');
```
and packages/custom/server.js 
```
  import './custom';
  
  console.log('Hello Server!');
```

now changing packages/custom/client/client.js results in client rebuild only 

Now update
```
meteor update --release 1.8.1-beta.16
```
now changing packages/custom/client/client.js results in server rebuild

(As a side note, also in 1.8.0.2 a server rebuild is done, when client js is located in packages/custom/client.js so the original problem is much older i suppose)

benjamn:
Most importantly, this change means that changes to files not used by the server bundle will not trigger a server restart.

Fixes #10449 by implementing the strategy I described in this comment: https://github.com/meteor/meteor/pull/10414#issuecomment-481293530


pmogollons:
@benjamn this is an awesome improvement. Got the full page reload time from 8s to 5s.

Is there a way to be able to include something like this in a module that is loaded on server and client and still get the benefits?

```js
// /startup/both/index.js

if (Meteor.isProduction) {
  require('./routes');
}

import './slingshot';
import './registerApi';
```

benjamn:
@pmogollons Great to hear it's working so well!

I would love your help testing a slightly simpler implementation of this functionality, coming soon (probably in 1.8.2-rc.7), once #10763 is merged.

To answer your question, in Meteor 1.8.3, we want to tackle tree shaking / dead code elimination, which involves pruning the dependency tree while scanning imports in the `ImportScanner`. I believe it should be possible to treat values like `Meteor.isProduction` as constants during this process, and eliminate those branches if their conditions are false, so that the `./routes` dependency would not show up at all in the production bundle.

PR #10696, opened 2019-09-19
A small optimization when updating documents

mitar:
I found a small optimization opportunity here. So at this point we:
* We do not need any projection or transformation.
* We do not modify returned `doc` in any way.

So there is no need to invoke full machinery of `findOne` (which internally creates a cursor object and copies the object every time using `EJSON.clone`).


veered:
This would be very helpful for us!

sebakerckhof (approved):

PR #10699, opened 2019-09-21
Update license range for year 2019

arggh:
Better late than never, right?


benjamn:
Ping about the new CLA?

arggh:
Should now be taken care of.

Issue #10643, opened 2019-07-22
Cannot find module error using Dynamic and Nested imports in the same application


filipenevola:
@benjamn I'm getting the same problem of this issue #10496 in a big project that uses heavily local Meteor packages, nested imports, and dynamic imports. I think the root cause is this mix between nested and dynamic imports, I don't think local Meteor packages is affecting it. 

The error is like `Cannot find module '/node_modules/meteor/pathable-ui/containers/helpers/permissions/with-is-allowed.jsx`

Looking into Sources tab of Chrome Dev Tools I see the module mentioned in the error being mentioned many times on static imports calls but its content is not there and that leads to this error.

Maybe what is happening is that this module is also requested by other modules that are used with nested imports and then somehow Meteor builder is avoiding it on the static bundle.

I already restart my application, cleaned `meteor/.local`, etc then I believe my case is not a cache problem.

---

Background: we have a bunch of widgets that are React components that represent each piece of our UI and to avoid evaluating all these modules at startup time we use nested imports in the React component that render these widgets into a page. Now I'm trying to also dynamic load some of these widgets components that are not loaded frequently to reduce bundle size. 

The hard part is that the error that I'm getting is not directly related to the widget that I'm dynamically loading now.

Anyway, I know you need a reproduction and I'll try to build one but in the meantime maybe this description can ring a bell of some potential gap in Meteor implementation of dynamic and nested imports.

benjamn:
Happy to look at a reproduction, even if that means sharing a private repo (if you're comfortable with that).

The module system _should_ be treating nested imports the same as top-level imports when scanning the dependency graph, so I would be surprised if that's the source of the problem, though I could imagine the dependency scanner somehow missing a nested static `import` (which would be a bug), so the module might appear to be imported only dynamically. If you can fix (or at least change) the behavior by temporarily hoisting the nested imports to the top level, that would be useful evidence.

filipenevola:
Sorry @benjamn for the long delay, I was on vacation, back now 😄 

We are having a hard time creating a reproduction in a small project. I tried before my vacation and also @Gywem tried some weeks ago.

Can we screen share? If not, I can share the code with you but it will take more time than screen sharing as you will need to setup many modules and tools to run our system properly. 

I don't think you will be able to figure out just by looking at the code due to many layers and packages 😞.

filipenevola:
@benjamn I have good news. 

I'm still confirming with more cases but I would like your feedback then I can proceed with more tests and a PR.

[...]

PR #10730, opened 2019-10-10
Fixes `cannot find module` error when using dynamic imports in massive projects

filipenevola:
Hello, I have explained the details of the problem here https://github.com/meteor/meteor/issues/10643#issue-470986741 and the root cause with the solution here https://github.com/meteor/meteor/issues/10643#issuecomment-528295725

This PR fixes the mentioned problem without causing any regression to Meteor scanning system.

I would like to also point out that even if my solution is somehow wrong, what I know it is not, the only bad side effect would be to scan a module that was already scanned in the same status, this means, do unnecessary work but I know for sure that I got a missing case and for who understands Meteor scanning system is relatively easy to see that we really had a gap there.

We are already using this fix in my company and we have more than 120 Cypress tests and all are passing and also if @benjamn or someone from MDG (or Tiny 😉) needs a reproduction I can provide access to our private code as it's not easy to create a reproduction in small code bases.

### DIFF
--- a/tools/isobuild/import-scanner.ts
+++ b/tools/isobuild/import-scanner.ts
@@ -238,8 +238,11 @@ const Status = {
 // that it has been imported by other modules. The "dynamic" status trumps
 // false, and "static" trumps both "dynamic" and false. A file can never
 // be demoted to a lower status after it has been promoted.
-const importedStatusOrder = [Status.NOT_IMPORTED, Status.DYNAMIC,
-  Status.STATIC];
+const importedStatusOrder = [
+  Status.NOT_IMPORTED,
+  Status.DYNAMIC,
+  Status.STATIC,
+];
 
 // Set each file.imported status to the maximum status of provided files.
 function alignImportedStatuses(...files: File[]) {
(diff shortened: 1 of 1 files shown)


benjamn:
Rebased against `release-1.8.2` and made some TypeScript-related changes. Assuming the tests pass, I'll merge this PR and release a new RC version of 1.8.2.

Very sorry for the delay, though it had nothing to do with your changes—I've just been swamped lately. I think it's very important to get these changes into Meteor 1.8.2.

tools/isobuild/import-scanner.ts
```diff
@@ -789,7 +812,12 @@ export default class ImportScanner {
       // newlyMissing and merge the new identifiers back into
       // this.allMissingModules.
       Object.keys(newlyMissing).forEach(id => {
-        if (has(previousAllMissingModules, id)) {
+        const skipScan = has(previousAllMissingModules, id) &&
+          !isHigherStatus(
+            getParentStatus(newlyMissing[id]),
+            getParentStatus(previousAllMissingModules[id]));
```
benjamn: I see why this is important in large applications, since you have to have a bunch of packages (especially custom local packages) to end up with cross-package imports that mix static and dynamic imports together. Makes sense!

PR #10740, opened 2019-10-15
change success variable declaration

gabrielmcoelho:
Since variables defined with **const** are not supposed to be reassigned, 'success = false' could cause errors. Changed declaration to **let** to avoid that.


filipenevola:
Thanks @gabrielmcoelho 😉

Issue #10704, opened 2019-09-23
Not possible to disable dynamic-import


mitar:
It looks like on Meteor 1.8.1 it is not possible to disable `dynamic-import` because it is a strict dependency of `ecmascript`. This is especially problematic because `dynamic-import` forces CSP to allow `eval`. So if you want to disable `eval` you have to remove `ecmascript`?

Moreover, if one does not want to use `dynamic-import` (because it does not work with disabled `eval`), having it bundled for the client always seems unnecessary.

See here for more information: https://github.com/meteor/meteor/blob/devel/packages/dynamic-import/security.js

nathan-muir:
Yeah, would love to get rid of `dynamic-import` or have it create static files that can be CDN hosted.

https://github.com/meteor/meteor-feature-requests/issues/296

mitar:
Does `WebAppInternals.setBundledJsCssUrlRewriteHook` not work for dynamic import files? For CDN hosting?

nathan-muir:
AFAIK `dynamic-import` does a HTTP POST (used to be DDP) to the meteor server, which then generates a mini-bundle of the requested paths on the fly.

jamesmillerburgess:
On Galaxy we use `dynamic-import` with a CDN by setting that hook.

jamesmillerburgess:
Sorry, @nathan-muir is right - these calls don't go over the CDN, but it would be great if they could.

mitar:
> which then generates a mini-bundle of the requested paths on the fly.

So then all possible combinations of files should be pre-generated so that they can be stored statically on CDN? Such generation could be useful already for local serving as well, instead of dynamic generation.

Anyway, I think dynamic-import support for CDN should probably be another issue.

coagmano:
Just noticed that `socket-stream-client` also has a strict dependency on `dynamic-import`, which means any app that uses `ddp` also gets it 😱

Interestingly, it seems to be redundant at the moment as it's used to [import `SockJS` as a fallback here](https://github.com/meteor/meteor/blob/5c32331591f268f2acdac5a1a7481eb9e65dc3d4/packages/socket-stream-client/browser.js#L198), but is already [statically imported here](https://github.com/meteor/meteor/blob/5c32331591f268f2acdac5a1a7481eb9e65dc3d4/packages/socket-stream-client/browser.js#L11)...

coagmano:
Actually, is there any reason why `modules` needs to imply `dynamic-import` except convenience? 
[...]

PR #10741, opened 2019-10-16
Remove redundant dynamic-import from socket-stream-client

coagmano:
Currently it's very difficult to remove `dynamic-import` from a Meteor app (#10704) because of strict dependencies from `ecmascript` and `socket-stream-client` (aka `ddp`).

This also means it's very difficult to disable `unsafe-eval` in CSP as `dynamic-import` sets `BrowserPolicy.allowEval()`.

Looking at `socket-stream-client`, the only dynamic import is rendered redundant by #9985 and so this PR cleans up the `import` call and dependency.

Maybe the optimization that #9985 reverts will come back, in which case it would be nice if SockJs could be conditionally bundled or `import()`ed based on whether the `dynamic-import` package is installed

### DIFF
--- a/packages/socket-stream-client/browser.js
+++ b/packages/socket-stream-client/browser.js
@@ -189,16 +189,7 @@ export class ClientStream extends StreamClientCommon {
     };
 
     this.socket.onclose = () => {
-      Promise.resolve(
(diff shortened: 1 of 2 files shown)


mitar:
Still relevant.

PR #10750, opened 2019-10-28
Fix race condition in the facts-base package tests as the Facts list may not be reset between tests

### LINKED ISSUE #10749: Facts package tests are running into race conditions
| Info                                              | Details                                                      |
|:--------------------------------------------|:-----------------------------------------------------|
| Meteor version                           |  1.8.2-rc4                          | 
| OS                                                | any                                                          |    

## Steps for reproducing the issue

```
git clone --recursive https://github.com/meteor/meteor.git
cd meteor
./meteor test-packages facts-base
```

After these commands, open a browser pointing to `http://localhost:3000` and see what happens:

Screenshot 2019-10-28 at 14 26 03

## Description of the problem

The Facts object is not reset between tests so they fail in subsequent runs.

msaelices:
Fixes #10749

### DIFF
--- a/packages/facts-base/facts_base.tests.js
+++ b/packages/facts-base/facts_base.tests.js
@@ -1,4 +1,6 @@
 Tinytest.add('facts-base - increments server facts', test => {
+  Facts.resetServerFacts()
+
   Facts.incrementServerFact('newPackage', 'skyIsBlue', 42);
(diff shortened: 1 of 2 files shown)


filipenevola:
cherry-picked to devel.

PR #10751, opened 2019-10-29
Allows the oauth server package to parse post body data

jramer:
This builds on work done by rooch84 in pull request #10501 and issue #10491 
That PR seems to have lost momentum and I am ready help with it as it's needed for sign in with apple oauth.

I have fixed the tests and made some small changes to take care of some bugs.
Tests are passing but i have some console warnings that I need help rectifying:
```
W20191029-15:10:41.469(1)? (STDERR) (node:79991) UnhandledPromiseRejectionWarning: TypeError: req.on is not a function
W20191029-15:10:41.537(1)? (STDERR)     at Promise (packages/oauth/oauth_server.js:146:8)
W20191029-15:10:41.538(1)? (STDERR)     at new Promise (<anonymous>)
W20191029-15:10:41.538(1)? (STDERR)     at parseRequestData (packages/oauth/oauth_server.js:142:33)
W20191029-15:10:41.538(1)? (STDERR)     at Promise.asyncApply (packages/oauth/oauth_server.js:188:5)
W20191029-15:10:41.538(1)? (STDERR)     at /Users/kim/git/bigowl/meteor/packages/promise/.npm/package/node_modules/meteor-promise/fiber_pool.js:43:40
```

### DIFF
--- a/packages/oauth/oauth_server.js
+++ b/packages/oauth/oauth_server.js
@@ -1,4 +1,4 @@
-import url from 'url';
+import bodyParser from 'body-parser';
 
 OAuth = {};
 OAuthTest = {};
@@ -136,6 +136,8 @@ OAuth._checkRedirectUrlOrigin = redirectUrl => {
 };
 
 const middleware = (req, res, next) => {
+  let requestData;
+
   // Make sure to catch any exceptions because otherwise we'd crash
   // the runner
   try {
@@ -158,7 +160,14 @@ const middleware = (req, res, next) => {
     const handler = OAuth._requestHandlers[service.version];
     if (!handler)
       throw new Error(`Unexpected OAuth version ${service.version}`);
-    handler(service, req.query, res);
+
+    if (req.method === 'GET') {
+      requestData = req.query;
+    } else {
+      requestData = req.body;
+    }
+
+    handler(service, requestData, res);
   } catch (err) {
     // if we got thrown an error, save it off, it will get passed to
     // the appropriate login call (if any) and reported there.
@@ -167,9 +176,9 @@ const middleware = (req, res, next) => {
     // is still open at this point, ignoring the 'close' or 'redirect'
     // we were passed. But then the developer wouldn't be able to
     // style the error or react to it in any way.
-    if (req.query.state && err instanceof Error) {
+    if (requestData?.state && err instanceof Error) {
       try { // catch any exceptions to avoid crashing runner
-        OAuth._storePendingCredential(OAuth._credentialTokenFromQuery(req.query), err);
+        OAuth._storePendingCredential(OAuth._credentialTokenFromQuery(requestData), err);
       } catch (err) {
         // Ignore the error and just give up. If we failed to store the
         // error, then the login will just fail with a generic error.
@@ -184,8 +193,8 @@ const middleware = (req, res, next) => {
     // Catch errors because any exception here will crash the runner.
     try {
       OAuth._endOfLoginResponse(res, {
-        query: req.query,
-        loginStyle: OAuth._loginStyleFromQuery(req.query),
+        query: requestData,
+        loginStyle: OAuth._loginStyleFromQuery(requestData),
         error: err
       });
     } catch (err) {
@@ -196,6 +205,8 @@ const middleware = (req, res, next) => {
 };
 
 // Listen to incoming OAuth http requests
+WebApp.connectHandlers.use('/_oauth', bodyParser.json());
+WebApp.connectHandlers.use('/_oauth', bodyParser.urlencoded({ extended: false }));
 WebApp.connectHandlers.use(middleware);
 
 OAuthTest.middleware = middleware;
--- a/packages/oauth/.npm/package/.gitignore
+++ b/packages/oauth/.npm/package/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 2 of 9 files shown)


packages/oauth/oauth_server.js
```diff
@@ -135,7 +135,34 @@ OAuth._checkRedirectUrlOrigin = redirectUrl => {
...
+//
+const parseRequestData = req => new Promise((resolve, reject) => {
+  if (req.method === 'POST') {
+    let rawBody = '';
+    req
+      .on('data', (chunk) => {
+        rawBody += chunk;
+      })
+      .on('end', () => {
+        const json = rawBody.split('&').reduce((arr, kv) => {
```
sebakerckhof: Can't we use node's built-in querystring parser?
https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options

jramer: Thanks for taking a look.
Sure, I'll fix that.

packages/oauth/oauth_server.js
```diff
@@ -1,4 +1,4 @@
-import url from 'url';
```
jramer: url import was not being used.

packages/oauth/oauth_server.js
```diff
@@ -135,7 +135,41 @@ OAuth._checkRedirectUrlOrigin = redirectUrl => {
...
+          reject(error);
+        }
+      })
+      .on('error', e => reject(e.message));
+  } else {
+    resolve(req.query);
+  }
+});
+
+const middleware = async (req, res, next) => {
```
sebakerckhof: You made this async, but there's nowhere an await.

packages/oauth/oauth_server.js
```diff
@@ -167,9 +203,9 @@ const middleware = (req, res, next) => {
     // is still open at this point, ignoring the 'close' or 'redirect'
     // we were passed. But then the developer wouldn't be able to
     // style the error or react to it in any way.
-    if (req.query.state && err instanceof Error) {
+    if (requestData.state && err instanceof Error) {
```
sebakerckhof: requestData is undefined if we come here

packages/oauth/oauth_server.js
```diff
@@ -158,7 +192,9 @@ const middleware = (req, res, next) => {
     const handler = OAuth._requestHandlers[service.version];
     if (!handler)
       throw new Error(`Unexpected OAuth version ${service.version}`);
-    handler(service, req.query, res);
+    parseRequestData(req).then((requestData) => {
```
sebakerckhof: We should also handle a rejection

packages/oauth/oauth_server.js
```diff
@@ -135,7 +135,41 @@ OAuth._checkRedirectUrlOrigin = redirectUrl => {
   );
 };
 
-const middleware = (req, res, next) => {
+// Parses request payload to use in the handler
+//
+// @param {Object} req incoming oauth request
+//
+const parseRequestData = req => new Promise((resolve, reject) => {
```
sebakerckhof: This entire thing should best be handled by existing middleware packages, since there's a lot of edge-cases and possible security concerns handled by packages like body-parser.

And some additional advantages like if the body is already parsed by another middleware you won't parse twice (e.g. the popular `simple:rest` package adds body parser middlewares for every route) and the oauth middleware stays synchronous, doing away with some of my other comments.

You'll want to make sure to add the middleware only for the oauth routes, so do something like:
`
WebApp.connectHandlers.use('/_oauth', bodyparser.json());
WebApp.connectHandlers.use('/_oauth', bodyparser.urlencoded());
`
before adding the main oauth middleware handler.

packages/oauth/oauth_server.js
jramer: Made it clearer where we where waiting by using an await instead of then.

packages/oauth/oauth_server.js
jramer: Fixed with await as well.

packages/oauth/oauth_server.js
jramer: [My first hack](https://github.com/meteor/meteor/pull/10751/commits/54d826f815bc820309ac791a16ed201d08eb6b48) did use body-parser but then I read somewhere to refrain from including external libs.

In rooch84 original issue, #10491. Ben talks about using a promise and so I thought it would be ok.

I'll change it if you think it should use bodyparser. As long as we gain the ability to parse post body data.

sebakerckhof: For external libs we have to be careful which to include, especially on the client-side.
However, this is server-side only and body-parser is a very well-known library. Otherwise this implementation would at least need some DOS protection by limiting the body size and limiting the parameters. Maybe for some other oauth provider it would need to inflate the data first etc. All of this is already taken care of by body-parser. But maybe let's wait for some input from @benjamn before you make the change.

filipenevola: hi @jramer are you going to use bodyparser or is the PR ready?

jramer: I'm waiting for input. It's your choice. 
I've been using this in production with google/facebook/apple oauth.

benjamn: If everything works when using `body-parser`, I think this is a case where an external library with 12MM weekly downloads is preferable to a handwritten reimplementation, for the reasons @sebakerckhof mentioned.

I am excited to see this PR get merged, so please go ahead and reinstate `body-parser` (unless there's a specific reason it won't work, in which case we can talk about that).

jramer: Cool. I’ll try to fix it ASAP

jramer: I think it's ready, and better than before. 🎉

PR #10771, opened 2019-11-11
Write files in Builder#_copyDirectory asynchronously.

benjamn:
Using `fs.writeFileSync` in a serial style becomes especially costly when we're writing a lot of files. In a recent profiling exercise I did on Windows, nearly 80% of the time taken by `Builder#_copyDirectory` was spent just closing the written files. By using the asynchronous `fs.writeFile` function, we should be able to parallelize at least some of this work, and `await` all the promises at the very end of copying the directory.

I'm not sure I would believe this without a screenshot, so here it is:

Screenshot 2019-11-11 17 51 31


benjamn:
Had to revert this unfortunately, due to `EMFILE` (too many open files) errors. In hindsight, that's not surprising, but still disappointing that Node.js can't natively handle lots of pending async `fs` operations. The `graceful-fs` package is famous for helping with this, but I definitely do not want to replace `fs` with `graceful-fs` at the very end of the Meteor 1.8.2 release candidate process.

PR #10772, opened 2019-11-12
Remove jquery dependency from test-helpers

sebakerckhof:
Step towards getting https://github.com/meteor/meteor/pull/10498 merged.

### DIFF
--- a/packages/http/httpcall_tests.js
+++ b/packages/http/httpcall_tests.js
@@ -259,16 +259,6 @@ testAsyncMulti("httpcall - methods", [
           test.equal(result.statusCode, 200);
           var data = result.data;
           test.equal(data.url, "/foo");
-
-          // IE <= 8 turns seems to turn POSTs with no body into
-          // GETs, inexplicably.
-          //
-          // XXX Except now it doesn't!? Not sure what changed, but
-          // these lines now break the test...
-          // if (Meteor.isClient && $.browser.msie && $.browser.version <= 8
-          //     && meth === "POST")
-          //   meth = "GET";
-
           test.equal(data.method, meth);
         }));
     };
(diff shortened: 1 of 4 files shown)


benjamn:
@sebakerckhof Can you bump the patch version in `packages/test-helpers/package.js`?

sebakerckhof:
@benjamn Sure. Btw, for the blaze packages I can't get rid of jquery too easily, but I got them working with jquery 3, but that means there's a bit of a chicken & egg problem. How would I go about creating PR's for those?

benjamn:
@sebakerckhof If there weren't many changes to the Blaze code that uses `jquery`, so both 1.x and 3.x are acceptable, then I think you can change the `package.js` version constraints from
```js
api.use("jquery@1.11.9", ...);
```
to
```js
api.use("jquery@1.11.9 || 3.0.0", ...);
```
to match either 1.11.x or 3.0.0+. We can release this change as a patch update for the Blaze packages, and perhaps remove the `1.11.9 ||` part the next time we bump their major/minor versions.

sebakerckhof:
Ok. One more issue. I had blaze working with jquery 3 by letting the jquery package install jquery 3. Obviously, that defeats the purpose of https://github.com/meteor/meteor/pull/10498 . But how can blaze get access to jquery when calling `test-packages` ?

benjamn:
I had to remind myself how this works, but apparently if you run
```sh
meteor test-packages ./packages --test-app-path /absolute/path/to/test/app/directory
```
the test application will run in the provided directory, so whatever `package.json` file is found there will control the application `node_modules` available to the package tests.

However, I just noticed that `--test-app-path` cannot currently be a relative path like `../test-app`, which is something I can fix in Meteor 1.8.2.

In any case, I think the ultimate solution here would be to put a `package.json` with `jquery` in the `dependencies` in the root directory of the `blaze` repository, and then run Blaze package tests with a command like
```sh
cd path/to/blaze
meteor test-packages ./packages/* --test-app-path .
```

PR #10774, opened 2019-11-12
Fix mysterious Windows AppVeyor test stalls.

benjamn:
Windows tests have been timing out because the preparatory `meteor --get-ready` command stalls forever (using 0% CPU), so I'd like to gain some insight into how far it gets before it stalls.

### DIFF
--- a/tools/fs/files.ts
+++ b/tools/fs/files.ts
@@ -1694,29 +1694,40 @@ const wrappedRename = wrapDestructiveFsFunc("rename", fs.renameSync, [0, 1]);
 export const rename = isWindowsLikeFilesystem() ? function (from: string, to: string) {
   // Retries are necessary only on Windows, because the rename call can
   // fail with EBUSY, which means the file is in use.
-  let maxTries = 10;
-  let success = false;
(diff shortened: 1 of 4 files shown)


benjamn:
Although we're still waiting for final confirmation, I believe it's safe to say the crux of the problem was that `files.rename` often fails on Windows due to transitory `EPERM` and/or `EACCES` errors, and thus needs to be retried. We were previously retrying up to 10 times, but we were doing so in a tight loop, without pausing in between the attempts. With the changes in this PR, we now retry the `rename` operation every 50ms after `EPERM` or `EACCES` failures, giving up after one second (roughly 20 attempts). These asynchronous delays seem to prevent the fallback behavior of recursively copying the source file/directory to the destination, which can be absurdly expensive on Windows (and was the cause of the AppVeyor test stalls).

smeijer:
> recursively copying the source file/directory to the destination, which can be absurdly expensive on Windows

Still an understatement. 

Thanks @benjamn! This PR means a lot to me, as a developer that uses Windows as his primary platform.

Issue #10230, opened 2018-09-20
Meteor including unnecessary node_modules


sebakerckhof:
In some cases meteor includes unnecessary node_modules in a package bundle code.
It let me to having 7 unnecessary copies of lodash included.

The conditions in which it happens:
1. Have a local meteor package which requires an npm module. Let's call that npm module 'X'. In my case it's a babel plugin which I don't import anywhere directly (just used in the `.babelrc` file of the package)
2. Import something inside the package from a node_module which you do not specify in the `package.js` file of the package, but is a dependency of module X. Let's call this module Y. In my case it's lodash.
3. Specify module Y in your `package.json` file of the app.

Expected result: Since it's not specified as a dependency and I don't import module 'X' anywhere, I would expect only to have the instance of module Y included in the `modules.js` file and nowhere else.

Observed result: Module Y is included both in `modules.js` (from `package.json`) and in the package bundle file.

Reproduction: https://github.com/sebakerckhof/meteor-npm-includes
See how the bundle of the `foo` package includes lodash, which shouldn't really be there.

I know it's a bit of an edge case. I guess if I would import module X somewhere inside the package then module Y probably should be included.

Moving module 'X' (the babel plugin) to the main package.json, would solve it, but doesn't work since then the babel plugin is no longer applied...

benjamn:
I think I know why this is happening, and possibly how to fix it, but I think it's going to have to wait for Meteor 1.8.1. Thanks for reporting this tricky case!

StorytellerCZ:
Could running `meteor npm prune` before deploy fix this issue in the meantime?

sebakerckhof:
@StorytellerCZ no, this wouldn't do much for this issue.

koszta:
@benjamn could you share your thoughts why is this happening?

sebakerckhof:
@benjamn Since `1.8.1` has been released in the meantime. Any chance this could be taken into the `1.8.2` release?

sebakerckhof:
> I think I know why this is happening, and possibly how to fix it, but I think it's going to have to wait for Meteor 1.8.1. Thanks for reporting this tricky case!

Could you share your idea on why this happens and how to fix it? Maybe I could try and tackle this one with some guidelines.

sebakerckhof:
> I think I know why this is happening, and possibly how to fix it, but I think it's going to have to wait for Meteor 1.8.1. Thanks for reporting this tricky case!

[...]

PR #10792, opened 2019-11-21
Only include direct npm dependencies of a Meteor package

sebakerckhof:
This is a potential fix for: https://github.com/meteor/meteor/issues/10230

By implementing strategy 2 described here: https://github.com/meteor/meteor/issues/10230#issuecomment-557022580

It seems to work for a couple of rather large Meteor applications I tested it on, but it's the first time I'm working on meteor-tool, and have to admit it's a bit over my head at the moment, so I might be overlooking some things.

self-test succeeds except for `custom-minifier`, but that one was already failing before.

------
In case a file within a package requires/imports an npm dependency
We should only resolve this if it is mentioned in the npm dependencies
of that package (thus in `Npm.depends` in the package.js file).
Otherwise, the package author likely wants to get it from the app's
node_modules.
The npm package could be in a node modules path of the meteor package
through some other npm package.
So we still want this other npm package to resolve to that package
Since its version may differ from the app's npm package.


tools/isobuild/resolver.ts
```diff
@@ -352,6 +356,38 @@ export default class Resolver {
       }
     }
 
+    if (this.npmDependencies.length > 0) {
```
sebakerckhof: This check should make it backwards compatible with older published packages

tools/isobuild/resolver.ts
```diff
@@ -77,11 +79,13 @@ export default class Resolver {
     sourceRoot,
     targetArch,
     extensions = [".js", ".json"],
+    npmDependencies,
     nodeModulesPaths = [],
   }: ResolverOptions) {
     this.sourceRoot = sourceRoot;
     this.extensions = extensions;
     this.targetArch = targetArch;
+    this.npmDependencies = Object.keys(npmDependencies || {});
```
sebakerckhof: npmDependencies can be undefined, but also null. So we can't use a default function parameter in the constructor signature.

zodern:
Meteor allows packages to also use a `package.json` file and a `node_modules` folder, though when using the package locally the developer has to install the npm dependencies instead of Meteor doing it automatically. Meteor then includes the `node_modules` folder when building or publishing the package, treating it the same as if the dependencies were listed in `Npm.depends`.

I have noticed dev dependencies are accidentally included when some packages are published due to this, though I have not paid attention before to if any package uses a `package.json` file instead of `Npm.depends`. If we want to preserve this, you might want to also look at the `package.json` file if it exists for a list of dependencies.

If this PR is accepted as is, it might be safe to exclude any top level `node_module` folders in a package when building, which usually is only dev dependencies.

sebakerckhof:
@zodern the behavior you describe is normally not affected by this PR, because it is checked for earlier in the method, this bit: https://github.com/meteor/meteor/blob/8b8610629f0326d9f70d767f61a77937daf4cc87/tools/isobuild/resolver.ts#L344-L357

But indeed, we could do a similar thing by looking at the package.json `dependencies` and only resolve those if we have an import from within the package. It's a bit more work since we will want to cache the package.json reads etc.

Would love to hear @benjamn thoughts on this.

zodern:
Here is a repro: https://github.com/zodern/meteor-package-package-json

The package has a package.json file with `shell-quote` as a dependency.

On the devel branch of Meteor, it outputs:
```
[Function: LRUCache]
{ quote: [Function], parse: [Function] }
/node_modules/meteor/local:test/node_modules/shell-quote/index.js
```

and with this PR it outputs:
```
Error: Cannot find module 'shell-quote'
```

I am not sure if this is important since I don't know if it is actually used by any packages. When I discovered this I thought it was strange we have two ways to list npm dependencies so I would be in favor of depreciating using a package.json to list bundled dependencies for a package. I just want to make sure we know we are breaking this before merging in case it is important or so the breaking change can be documented.

sebakerckhof:
@zodern Ah yeah, should be fixed now. What I said in my previous post was correct, except I then totally disregarded the fact the package was already found in the `node_modules` 😅. Note that Meteor will search for node_modules recursively, so it doesn't even need to be in the root of the package (which is compliant with how node normally would resolve modules).

I agree that there's probably not much of a point in having a package.json file next to `Npm.depends` except for dev dependencies. But I want to avoid breaking existing packages. + it's the way node normally works, so one could argue that it's the meteor-specific `Npm.depends` which should be deprecated.

sebakerckhof:
@filipenevola You added this commit to my accounts-password PR:
https://github.com/sebakerckhof/meteor/commit/63c53bea4f59174b41a1ca3a725e3eb186d50750

This breaks the circleCI tests since it shouldn't have the versionsFrom. What's the reason you added this?

filipenevola:
@sebakerckhof I can't publish without `versionsFrom`.

Ben told me to publish with `versionsFrom` next time but don't commit it. We should probably fix this in the near future.

benjamn:
By the way, you can run the modules test app locally with
```sh
cd meteor/tools/tests/apps/modules
../../../../meteor npm i
../../../../meteor test --full-app --driver-package dispatch:mocha-browser
```
Client-side tests will run if you open `localhost:3000` in a browser.

Running `meteor self-test modules --retries 0` also works (that's what CircleCI does), but the output is a bit more opaque.

sebakerckhof:
@benjamn thanks

> Can you say more about what happens when a module within the package's .npm/package/node_modules directory imports another package that wasn't installed directly by Npm.depends? Shouldn't it get the version that was installed within the package, rather than version in the application node_modules directory?

Yes, that behavior is not changed. I check to see if the import comes from another npm package:
https://github.com/meteor/meteor/blob/f323df8735c7ba98afb3ac7c91b6b7a147590719/tools/isobuild/resolver.ts#L371-L373
https://github.com/meteor/meteor/blob/f323df8735c7ba98afb3ac7c91b6b7a147590719/tools/isobuild/resolver.ts#L386
And `this.nodeModulePaths` will contain `.npm/package/node_modules`

So this only affects direct `meteor package -> npm package` imports

> please add thorough tests to the modules test app. 

Sure, I'll add them

benjamn:
@sebakerckhof Great news, thanks!

@filipenevola On second thought, @sebakerckhof can probably just rebase against the current `devel` branch to get the tests running again, since we've fixed that `api.versionsFrom` problem since the last time these tests ran.

benjamn:
Ok, so, I think this version of `cheerio` might not ever import from the `lodash` package itself, but only sub-packages like `lodash.bind` and `lodash.map`. Have a look at
```
meteor/tools/tests/apps/modules/.meteor/local/build/programs/web.browser/packages/npm-test-package.js
```
And search for `"lodash"` within the `cheerio`-related code. The consequence is that, while `lodash` is installed in `packages/npm-test-package/.npm/package/node_modules/lodash`, it doesn't get bundled into the tree of module definitions that are known to the runtime module system, so package code outside of `node_modules` won't be able to find it (and be confused by it) at runtime.

Do you think you could find a different package than `cheerio` that definitely does import `lodash` when you import the package? I'm a little bit worried that if `cheerio` did bundle `lodash`, then imports from package code would see that version at runtime, and your tests would behave differently.

If my worry is true, I think I have a plan for how we could structure the runtime module tree differently, which I would be happy to implement if you like. In short: we could nest `.npm/package/node_modules` somewhere not on the module lookup path, then create a `node_modules` directory on the lookup path that contains aliases for just the direct dependencies, pointing into the nested `.npm/package/node_modules` directory.

sebakerckhof:
@benjamn Yes, I explicitly picked a package which didn't require lodash.
If you look at the issue this is trying to address, it's when you have a module like lodash in your package's node_modules, which is never imported, but still in your node_modules. Lodash can for example end up there because of a babel plugin (like the  `babel-plugin-angularjs-annotate` one I used in the test)

Once it's required by another npm package, it doesn't really matter anymore (at least for bundle size), since you'll get 2 copies anyway (although this could also be improved if the global version satisfies what the npm package wants, but that requires a lot more changes).

However, for sake of consistency it would be good if the behavior is always the same, thus that direct `meteor package -> npm package` imports come from the global version. While `meteor package -> npm package -> npm package` imports come from the package npm modules.
This is currently the case on the server-side, but not in the browser build. I haven't really looked into why there is a difference and how easy it is to make consistent.

sebakerckhof:
@benjamn Actually, isn't the resolve.require supposed to give the same result on browser & node? If you look at my changed test in the last commit, this works on the server, but not in the browsers. Could this be a bug in modules-runtime?

tools/tests/apps/modules/packages/npm-test-package/package.js
```diff
@@ -10,8 +10,7 @@ Package.describe({
 
 Npm.depends({
   "cheerio": "0.22.0",
-  // This includes the lodash package
-  "babel-plugin-angularjs-annotate": "0.10.0"
+  "@sebak/lodashuser": "1.0.0",
```
sebakerckhof: This is just:
```js
const lodash = require('lodash');
module.exports = {
  lodashVersion: lodash.VERSION,
}
```

benjamn:
On the server, we use the `module.useNode()` trick to switch over to native Node.js execution when possible, so `require.resolve` may be different if it's called from a module evaluated natively by Node.js. As long as you're within the `meteorInstall`ed bundle, `require.resolve` should give the same results in browsers and Node, modulo the possibility that some code may not be included in both bundles.

sebakerckhof:
Yes, but shouldn't the require.resolve from meteorInstall try to resolve to the same thing like the native version? If that could be solved than this PR can be a little bit more ambitious than just handling my edge-cases and potentially save quite a bit of bundle size for some users.

Currently with this PR, for both browsers and server, the tool's resolver will say this lodash import is missing:
https://github.com/meteor/meteor/blob/bed2da2527edb372e39e5396df3d452c34d85800/tools/tests/apps/modules/packages/npm-test-package/common.js#L1

But then on the browser it will eventually resolve to the package's one, and on the server it will eventually resolve to the app's global one.

sebakerckhof:
Ok, I think this is worth solving, since currently one could get v1 of a dependency on the client, and v2 of a dependency on the server in code that should behave isomorphic.

So I was thinking to change the linker/meteorInstall to include which information depending on which packages are direct dependencies. Then in modules-runtime use that to first check the root's node_modules in `fileResolve` if the parent isn't another npm package?

Do you think this is a good strategy?

benjamn:
I agree it's worth solving!

I don't want to build more Meteor-specific logic into the runtime module system than we have to. I think this is a problem that can be solved by restructuring the locations of `node_modules` directories in the bundle, as I described above. I'm happy to work on this.

afrokick:
@filipenevola @sebakerckhof this is a good PR to include in 2.0.

filipenevola:
Hi @sebakerckhof do you want to proceed with this PR?

I believe it's almost ready.

sebakerckhof:
@filipenevola Unfortunately, the approach I took in this PR has some issues.
@benjamn  vaguely describes an alternative approach a bit above: https://github.com/meteor/meteor/pull/10792#issuecomment-575341504

But I don't think I'll have time for this in the near future (as you may have noticed, I didn't have a lot of time for Meteor contributions in the past year and that doesn't seem likely to change at least for the coming months).

PR #10797, opened 2019-11-25
Update terser to 4.4.0

sebakerckhof:
Includes a couple of fixes, and 4.3.5 should include a fix for a performance regression from 4.2.0


benjamn:
As I hoped, those test failures are the same ones that I fixed in f6fcd620a7b66b19a4956d4559592e8bfe43051f and cf8dbf1c37a9da249bf1ab197f81345a6f1b0531, which are currently on `devel`, so I think it's safe to say the tests would pass if we rebased this branch, which means this PR should be safe to merge without actually rebasing it first. 🎉

PR #10798, opened 2019-11-25
Update source-map to 0.7.3

sebakerckhof:
source-map 0.7.0+ has a much faster Rust WASM implementation.
Because this needs to be loaded, the constructor is now asynchronous.
The consumer should also be destroyed after it's no longer needed.

https://github.com/meteor/meteor/commit/1cf59867b2f74f5be1a3e6f1eb13ea4ae10540db paved the way for this.

Should hopefully help with: https://github.com/meteor/meteor/issues/9568 , but didn't ran any benchmarks yet.

### DIFF
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=12.13.0.1
+BUNDLE_VERSION=12.13.0.2
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
--- a/meteor
+++ b/meteor
@@ -132,7 +132,10 @@ fi
 # the script take precedence over $NODE_PATH; it used to be that users would
 # screw up their meteor installs by have a ~/node_modules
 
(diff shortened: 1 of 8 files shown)


sebakerckhof:
For release-1.9 I had to add the `--no-wasm-code-gc` node flag because of an issue with WASM gc on node 12.11+ and node-fibers: https://github.com/nodejs/node/issues/29767

I didn't even know there was a GC in wasm now. But since source-map is the only WASM in meteor-tool and it does its own memory management `--no-wasm-code-gc` shouldn't make a difference (to be verified).

benjamn:
Ok, I rebuilt the dev bundle with `source-map@0.7.3`, and I can confirm that everything seems to be working, though I suppose we'll have to wait to see what the tests say.

Thanks for jumping on this @sebakerckhof!

opyh:
Maybe related:

https://github.com/mozilla/source-map/issues/412
https://github.com/mozilla/source-map/issues/342

Issue #10788, opened 2019-11-20
Published accounts-password doesn't match repository


sebakerckhof:
I noticed that the published accounts-password 1.5.1 package is not up to date with the repository.

E.g. The package here on git, does no longer depend on underscore (and hasn't for the past 2 years), but the package.js file of the sources published on atmosphere still do:
https://warehouse.meteor.com/sources/Aw5o5YMrAYo9bGgRY/1519255742438/DLp2Wznwkc/accounts-password-1.5.1-source.tgz

I noticed this while searching why we still got underscore in our client bundle, but there might be more packages that need republishing.

filipenevola:
1.5.2 is now published.

Thank you 😉

PR #10799, opened 2019-11-25
Bump accounts-password version

sebakerckhof:
Should be republished, since the published 1.5.1 is outdated and still contains underscore. https://github.com/meteor/meteor/issues/10788


filipenevola:
1.5.2 is published.

PR #10801, opened 2019-11-26
Add upgrader to add jquery dependency for blaze-users

sebakerckhof:
While jquery is now a theoretical weak dependency of blaze, in reality everyone uses the jquery dombackend of blaze. So try to limit the complaints about apps being broken. (they will still need to have it as an NPM dependency, but at least the `jquery` package gives a clear error message if this isn't the case)


benjamn:
Good thinking!

@mitar @filipenevola I think we might need to release new versions of the Blaze packages to get the test failures in #10498 (and here) to pass. The [necessary changes](https://github.com/meteor/blaze/pull/299) have been pushed to `blaze/master` but not published to the package server. Let me know if you don't have time right now, or if you have any questions.

mitar:
I do not have time right now to look into this, sorry.

Issue #10782, opened 2019-11-17
More information when a publication throws an error


mbatchkarov:
Whenever an exception is thrown in a publication, there is isn't enough information in the server logs to diagnose the problem. For example:

```
Meteor.publish("bug", function() {
  let options = { fields: { profile: 1 } };
  console.log(options.a.b.c);
  return Meteor.users.find();
});
```

When a client subscribes, the following is printed in the server console:

```
Exception from sub bug id iw7bzHXSsXZtz9QSK [object Object]
```
Spotting the error is easy enough with a 4 line subscriptions but gets very laborious as complexity increases. I appreciate there are arguments as to why error details should not be sent to the client, but I think the server-side log should be more verbose.

Meteor version: 1.8.1 on both Linux and OSX

zodern:
@mbatchkarov are you using an APM package? Most of them, at least in the past, would cause the errors in methods and publications to appear as `[object Object]`.

mbatchkarov:
I'm using Kadira in the app that prompted me to start investigating, but before I started this issue I also tried with a brand new minimal app and I saw the same. Based on #10807 I think this would also be the same

filipenevola:
@mbatchkarov this version of `meteor-apm-agent` published today should help you https://github.com/meteor/meteor-apm-agent/blob/master/CHANGELOG.md#v325

Please open this issue again if this still happens.

PR #10807, opened 2019-12-02
Log error stack on exception in publication

mbatchkarov:
When an exception is thrown in a publication, it is very useful to see the entire stack trace rather than just the name of the error (#10782). #9678 refactored the debug messages and unfortunately made the server-side error messages less detailes. This PR tries to address the issue  by reverting some of the changes made by #9678,  specifically https://github.com/meteor/meteor/pull/9678/files#diff-7e48a19069e07781db2a50282a8bfdd7R1737

There is perhaps an opportinuty to print more details in other places, e.g. when exceptions are thrown in method calls, etc. #9678 added a todo that stack traces should be formatted nicely but that seems to have not been addressed. IMO it's better to print badly formatted stack traces than nothing at all. I'd be happy to do that if people are happy

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -1738,7 +1738,7 @@ var wrapInternalException = function (exception, context) {
   // Tests can set the '_expectedByTest' flag on an exception so it won't go to
   // the server log.
   if (!exception._expectedByTest) {
-    Meteor._debug("Exception " + context, exception);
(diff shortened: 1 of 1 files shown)


filipenevola:
@mbatchkarov can you update your branch with development? If it is passing we can merge.

mbatchkarov:
Hi @filipenevola done, just rebased my branch from the latest dev branch. Let's see if it passes...

PR #10812, opened 2019-12-11
Explain previous work on minimal Meteor applications.

benjamn:
I noticed that the updated [roadmap](https://github.com/meteor/meteor/blob/devel/Roadmap.md) has a section titled **Ultra-thin Meteor**, but does not provide any information about the `meteor create --minimal` command introduced in Meteor 1.7.

I've expanded that section to describe the command and the opportunities for improvement that I think are worth exploring from that starting point.

Note to @filipenevola: since this PR is updating a single file with a single commit, let's be sure to squash-and-merge it to avoid adding a merge commit to `devel`.

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -39,7 +39,20 @@ A proper service worker build target. Regular Web Workers can be built from a fu
 - Status: -
 - PRs: -
 
-Have a ultra-thin version of Meteor that does not depend in anyway to mongo. Allow completely shutting down DDP and dependencies on DDP. ([MFR #31](https://github.com/meteor/meteor-feature-requests/issues/31), [MFR #354](https://github.com/meteor/meteor-feature-requests/issues/354), [Issue #9960](https://github.com/meteor/meteor/issues/9960))
+[Meteor 1.7](https://github.com/meteor/meteor/blob/devel/History.md#v17-2018-05-28) introduced the `meteor create --minimal` command, which generates a new application without any unnecessary Meteor packages, like `mongo` and `ddp`.
+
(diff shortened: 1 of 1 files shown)


Roadmap.md
```diff
@@ -39,7 +39,20 @@ A proper service worker build target. Regular Web Workers can be built from a fu
 - Status: -
 - PRs: -
 
-Have a ultra-thin version of Meteor that does not depend in anyway to mongo. Allow completely shutting down DDP and dependencies on DDP. ([MFR #31](https://github.com/meteor/meteor-feature-requests/issues/31), [MFR #354](https://github.com/meteor/meteor-feature-requests/issues/354), [Issue #9960](https://github.com/meteor/meteor/issues/9960))
+[Meteor 1.7](https://github.com/meteor/meteor/blob/devel/History.md#v17-2018-05-28) introduced the `meteor create --minimal` command, which generates a new application without any unnecessary Meteor packages, like `mongo` and `ddp`.
+
+When minified and gzip-compressed, the JS bundle for this app weighs in at less than 20kB, which is much smaller than the default `meteor create` application. Nevertheless, there is still room for improvement, using techniques like bundle visualization (`meteor npm run visualize`) and converting static `import`s to dynamic `import()`s.
```
benjamn: Side note: while testing the `meteor npm run visualize` command with Meteor 1.8.2, I noticed that we have a fairly substantial regression in the size of the `dynamic-versions.js` module (now 13kB minified, before gzip), which suggests that we are including many more dynamically-importable files than in Meteor 1.8.1.

Also, the modern version of the `ecmascript-runtime-client` package (otherwise the largest package in the bundle) could probably shrink if we revisit some polyfill decisions, or tighten the requirements for modern browsers.

PR #10818, opened 2019-12-13
Limit fetching full user objects (performance) Fixes #10469

### LINKED ISSUE #10469: Meteor.users.find() in core without field specifier could be a performance issue
Meteor core contains several instances of `Meteor.users().find()` and `Meteor.users().findOne()` without a field specifier in situations where only certain fields are required.

If a developer stores custom data on the `Meteor.users()` collection (as recommended in the [Guide](https://guide.meteor.com/accounts.html#top-level-fields)) then this will cause unnecessary data transfer between the db server and the meteor server.  This could be significant if lots of custom data is stored in the Meteor.users collection.

These examples definitely only need specific fields:
* [accounts-base/accounts_server.js#L1324](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L1324)
* [accounts-base/accounts_server.js#L1333](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L1333)

These may need the full user object to pass to the callbacks, I'm not sure:
* [accounts-base/accounts_server.js#L320](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L320)
* [accounts-base/accounts_server.js#L191](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L191)
* [accounts-base/accounts_server.js#L401](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L401)
* [accounts-base/accounts_server.js#L1215](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L1215)

This one has a definite error by omitting the `fields:` prefix:
* [accounts-base/accounts_server.js#L1151](https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/accounts-base/accounts_server.js#L1151)

There will probably be other instances in other files/packages - I haven't done a thorough search.

wildhart:
Meteor core contains several instances of `Meteor.users().find()` and `Meteor.users().findOne()` without a field specifier in situations where only certain fields are required (e.g. on every successful login), which can be a performance issue if developers store lots of data on the user object #10469, & [forum discussion](https://forums.meteor.com/t/collection-setup-for-a-basic-saas-using-stripe/50044/3?u=wildhart).  Storing custom data on the user object is [recommended in the docs](https://guide.meteor.com/accounts.html#custom-user-data).

Also, `Meteor.user()` doesn't provide any parameter for limiting the fields which are returned.  This can be a performance issue on the server and cause over-reactivity on the client, e.g. if a developer does `Meteor.user().profile.name`.  https://github.com/meteor/meteor-feature-requests/issues/82

Also, all accounts callbacks `onLogin`, `onLoginFailure`, `onLogout` and `validateLoginAttempt` receive the full user object which can be wasteful.  This PR adds new, backwards compatible API to allow a developer to specify which fields they do or don't want on the callbacks.

Todo:

- [x] limit fetched fields in:
  - [x] `accounts-base/accounts_server.js`:
    - [x] `defaultResumeLoginHandler`
  - [x] `accounts-password/password_server.js`:
    - [x] `Accounts.registerLoginHandler("password"...` (x2)
    - [x] `Accounts.setUsername`
    - [x] `Accounts.setPassword`
    - [x] ~~`Accounts.generateResetToken`~~ don't limit fields, they are returned by the function
    - [x] ~~`Accounts.generateVerificationToken`~~ don't limit fields, they are returned by the function
    - [x] `Accounts.addEmail`
    - [x] `Accounts.removeEmail`
    - [x] `Meteor.methods({changePassword...`
    - [x] `Meteor.methods({forgotPassword...`
    - [x] `Meteor.methods({resetPassword...`
    - [x] `Meteor.methods({verifyEmail...`
    - [x] `checkForCaseInsensitiveDuplicates()`
  - [x] tests
- [x] add `Accounts.config({defaultFieldSelector: {...}})` which can be used to limit the returned fields in ALL internal & 3rd party user finds which the developer doesn't have control over. e.g. the developer uses a `user.transactions` array which grows indefinitely and so doesn't want it fetched every time a user logs in or requests a new password.
  - [x] use in hooks: `onLogin`, `onLogout`, `onLoginFailure`, `updateOrCreateUserFromExternalService`
  - [x] tests
- [x] use `defaultFieldSelector` with optional `options` over-ride in:
  - [x] `Meteor.user()` (server - for db efficiency)
  - [x] `Meteor.user()` (client - to limit reactvity)
  - [x] `Accounts.findUserByEmail()`
  - [x] `Accounts.findUserByUsername()`
  - [x] tests
- [x] ~~add field specifier for accounts hooks~~ ignore this, use `options.defaultFieldSelector` (above) instead - much simpler and just as effective at eliminating large data the developer might store on the user object.
- [x] Updated History.md
- [x] Create PR for API docs: https://github.com/meteor/docs/pull/644
- [x] Create PR for Guide: https://github.com/meteor/guide/pull/1022


wildhart:
@filipenevola, what's the process for submitting related PRs to the `/meteor`, `/docs` **and** `/guide` repos?  The publishing of the update to the API docs need to wait until this code PR is makes it into a release.  The update to the guide will link to the new API docs.  How is all that synchronized?

coagmano:
My past experience is that it's synchronised manually! (or not at all). 
But good to put links in each PR pointing to the one it depends on, to make it easier to keep track of

filipenevola:
@wildhart link all the PRs here and then we publish both together (manual sync).

We can improve this soon, @sebakerckhof was suggesting to have everything in this repo for example 😉

wildhart:
This PR is ready for review and assigning to a milestone...

History.md
```diff
@@ -3,6 +3,20 @@
 ### Changes
 
 * The `meteor-babel` npm package has been updated to version 7.7.4.
+* `Meteor.user()`, `Meteor.findUserByEmail()` and `Meteor.findUserByUserName()` can take a new
```
sebakerckhof: nitpick: can you add a blank line between the bullet points, similar to the rest of the changelog.

sebakerckhof:
Code LGTM. However, I wonder if the `defaultFieldSelector` config option isn't gonna end up being a footgun. Especially when using a whitelist selector, it's easy to break unexpecting third-party packages in subtle ways you'll only find out during production. But maybe a clear warning about this is all this needs. (FYI: personal opinion, I don't decide on what goes in).

wildhart:
@sebakerckhof, I did think that too about `defaultFieldSelector`, and did consider making it blacklist only and enforcing that through validation within `Accounts.config()`.  That way the user can only use it to exclude their own large fields and would prevent problems with 3rd party packages expecting their own fields to be there.  Happy to add that if you want...

Could even rename it to `defaultExcludedFields` and take an array of field names instead of a Mongo field specifier...

Also, it wasn't until I did the docs that I realised it was formally called a [`field specifier`](https://docs.meteor.com/api/collections.html#fieldspecifiers) so if we keep it as a field specifier I could rename it to `defaultFieldSpecifier` if you want...

History.md
wildhart: Done, well spotted, sorry!  Plus merged 1.8.3 changes to History.md

wildhart:
Suggestion: I've added `Accounts._checkPasswordUserFields` because otherwise `Accounts.registerLoginHandler("password"...)` (x2) and `Meteor.methods({changePassword...` have to make an assumption about the `user` fields needed by `Accounts._checkPassword()`, but this feels like overkill.

Since these are all internal I could simply remove `Accounts._checkPasswordUserFields` and in the highly unlikely event that future contributors change `Accounts._checkPassword()` so that it needs additional user fields, we rely on that contributor looking for where it's used.

Or, `Accounts._checkPassword()` could be refactored to take the explicit parameters it needs, e.g. `Accounts._checkPassword = (userId, password, passwordHash)`.  However, a quick Google search shows that people are using `Accounts._checkPassword()` in their own code even though it is not public API...

Or maybe a good compromise is destructuring the user object within the function definition so that future users of that function are aware of what it needs: `Accounts._checkPassword = ({_id, services}, password)`

packages/accounts-base/accounts_server.js
```diff
@@ -1330,7 +1333,8 @@ const defaultResumeLoginHandler = (accounts, options) => {
         {"services.resume.loginTokens.hashedToken": hashedToken},
         {"services.resume.loginTokens.token": options.resume}
       ]
-    });
+    },
+    {fields: {"services.resume.loginTokens": 1}});
```
sebakerckhof: Another nitpick, but this could even be more limited with a positional operator, thus:
`{fields: {"services.resume.loginTokens.$": 1}}`
Same for the one on line https://github.com/meteor/meteor/pull/10818/files#diff-2ae5b6c36ab4132c1a2d0e33b5fd8443R1323

wildhart: Yes, in the simple query, but doing it with the `$or` query fails the tests because
```
    user = accounts.users.findOne({
      $or: [
        {"services.resume.loginTokens.hashedToken": hashedToken},
        {"services.resume.loginTokens.token": options.resume}
      ]
    },
    {fields: {"services.resume.loginTokens.$": 1}});
```
returns `{ _id: ..., services: { resume: {} } }`

I think because of [this](https://docs.mongodb.com/manual/reference/operator/projection/positional/#array-field-limitations) limitation of the positional operator:

> The query document should only contain a single condition on the array field being projected. Multiple conditions may override each other internally and lead to undefined behaviour. 

Is there another way of filtering the returned array within mongo?  I'm playing with `$elemMatch` but no luck so far with nested fields and `$or`...

sebakerckhof: Yep, you're right for the $or case and afaik, there's no other way to do this.

wildhart: Pushed a commit with the positional operator in the the simple case, and added a comment to avoid it in the `$or` case.

wildhart:
This PR is ready for final review and merging...

Happy New Year Meteor team!

wildhart:
I've created a PR for the changes to the guide: https://github.com/meteor/guide/pull/1022

PR #10819, opened 2019-12-16
Roadmap: SSR

eric-burel:
Hi,

I am not sure I'll be able to bring in serious contribution to Meteor until a while (lack of free time, a man has to do its freelance work ;) ), but I've worked on the Apollo 2 update in @SachaG's framework [Vulcan ](http://vulcanjs.org/)(Meteor + React + GraphQL with Apollo).

We ended up with a full-fledged setup, including SSR.

You can check this folder: https://github.com/VulcanJS/Vulcan/tree/devel/packages/vulcan-lib/lib/server/apollo-server
And this one: https://github.com/VulcanJS/Vulcan/tree/devel/packages/vulcan-lib/lib/server/apollo-ssr

We also have the client side part but it has nothing fancy so it's less interesting.

Limitations that comes to my mind:

- SSR with Meteor is weird, mostly due to some lacking doc. `meteor/webapp` is not very documented so understanding that it's actually just a Connect server cost me like a few hours. The `sink` API is also weird too. Basically Meteor should get closer to Express regarding request/response access, so the wording is more understandable for any Node dev.

- SSR is VERY EXPENSIVE. It just kills your request/per second. This is because rendering is a synchronous call, and an expensive one. That's a problem in normal Node apps too, not only in Meteor, there are no easy solution at all. 
Streaming the response might one solution but it raises issues with Helmet (I've seen a post from @CaptainN about this on Meteor forums), also Node packages using this approach may have performance regression in the end depending on the page. Node threads are coming, but they seem too expensive to be spawn on demand. The most relevant solution could be setting up a thread pool, and combine it with an advanced caching strategy.

- There are some Vulcan idiosyncrasies in our code, because we have some helper to enjoy Meteor package system (callbacks, setting registeration). Don't pay too much attention to them.

- We have one last non NPM dependency client side: `meteor/apollo`, for account connection. Otherwise Vulcan's frontend code could mostly be reused outside Meteor. We plan on moving to Account.js. Then we would eventually be consistent with the "non meteor" ecosystem and be able to reuse our code both in Meteor and outside of it.

- **Apollo is the Meteor of GraphQL.** It seems to be following the exact same path, with the exact same advantages (easy to plugin, a lot of doc, reactive team, widely used, impressive work etc. etc.) and the exact same problems: weird architectural choices that could frighten the community, and updates that provoke unexpected regressions.

Obviously Apollo can't be perfect from day one, and I am infinitely thankful for their free and open source work. But I am afraid history will repeat and Apollo become a subject of horror stories ("my startup chose Apollo and now it's DEAD"), that would be extremely sad.

See https://github.com/apollographql/apollo-server/issues/1908 for example. Putting Playground on `/graphql` path and playing on HTTP verbs is at best a marketing stunt. **It will please beginners in the short run, it will create resent from professional in the long run**. 

See also this one: https://github.com/apollographql/apollo-link/issues/1111. The only way to write a Link for file uploading is duplicating 90% of the normal HTTP link... and this issue was worsened after some update by the removal of some option so it's not on the right track.

**So I think if the Meteor community is going to be "revived", that would also mean paying a solid attention to what happens in Apollo.**

It's just the first few things that come to my mind, don't hesitate to ask questions about our experience with Meteor + Apollo. On a technical standpoint the difficulty is mostly related to understanding correctly how plugging middlewares to Meteor works.

Edit:
Actually there is one interesting feature client side, it's optimistic UI, ie updating lists of data when the user create/update/delete some item.

See our [`multiQueryUpdater`](https://github.com/VulcanJS/Vulcan/blob/devel/packages/vulcan-core/lib/modules/containers/create2.js). Due to the limited `readQuery` API of Apollo, it's difficult yet to update the cache manually.

It's still a work in progress (need to sort data after update + factorizing code) but it does the job for creation and deletion already.

Relevant issues:
- https://gist.github.com/ngryman/6856c7eb8f9a15b1095032a6ba478c5c
- https://github.com/apollographql/react-apollo/issues/708#issuecomment-506975142
- https://github.com/apollographql/apollo-client/issues/3505
- https://github.com/apollographql/apollo-client/issues/3505#issuecomment-535388194


sebakerckhof:
Any particular reason you bundle SSR with Apollo here? In my view they're 2 completely separate subjects.

I like the idea of bringing webapp closer to express. The more it is clear that Meteor is just a layer on top of nodejs where you can do all your regular nodejs stuff, the more it may be able to resonate with the larger js community.

eric-burel:
Yes it's not completely related, but there are a few dependencies, because you need to use Apollo API to fetch the data you need during SSR. You need to spawn a limited client for each request for example.

It's the same underlying issue with normal REST queries of course, it's just that in Vulcan's case we can provide a working example specifically with Apollo and GraphQL, not with other technologies.

CaptainN:
A couple of notes, in no particular order:

[My Starter](https://github.com/CaptainN/meteor-react-starter) does SSR with mongo, no Apollo. I'm not entirely sure there's not a memory leak in core though, due to the interplay between React's hooks and Meteor's Fibers.

My starter also does `React.renderToString` while capturing helmet and data output, then sends it along [through a stream](https://github.com/CaptainN/meteor-react-starter/blob/master/server/main.js). I didn't measure extensively, but this seemed to reduce my time to first byte a bit.

Most SSR solutions I've seen for React with Apollo involve doing basically, two parsing/rendering passes. They first parse the react tree (I forgot the call) then do `React.renderToNodeStream` to output to a stream. However, from what I've read, parsing the react tree is the expensive part of rendering, and by this model you end up parsing the tree twice. I'm not an expert though - maybe I'm missing something. To get around that I decided to just render once, to a string, then stream that string. The downside I can see to that is that I am storing a fairly large string in memory for a bit, for every request, which I'd guess puts some pressure on server memory. I figured the trade off was worth it to get content to the client faster.

I would genuinely LOVE to hear why folks do it the other way, with two passes in most of the Apollo SSR tutorials I've seen.

Strategies to make this all less extensive all involve caching. You can for example, cache each of the pieces - data, helmet, html, etc. then draw from that, and even invalidate and fill the cache in a worker or something on the server. Or you can just put a front-end-cache between the user and the app server, and serve slightly stale pre-rendered content (this is probably the easiest thing to do).

I used to advocate for only rendering "static" content - that is, content that doesn't change based on the user. But this proved to be difficult to juggle. I've since removed my of that code, and now render everything on the server, even for authorized users. I use `staringatlights:fast-render` to get around the limitations of calling `Meteor.user()` in SSR code.

I've wondered if an alternative (or maybe extension) to DDP could be implemented on top of Apollo for a Meteor 2.0. It'd be nice if we could match the convenience of the Meteor way of doing things (including defaulting to Mongo, optimistic updates through Methods and local, maybe offline by default collections, etc.), but have the transport be built on Apollo (including with some live data solution). Then have that be "ejectable" in some way, so that the user can import some packages, and reconfigure or add to the Apollo server. The model to follow is how Meteor has a default, but extensible set of babel plugins.

filipenevola:
I would like to keep Apollo and SSR as independent discussions.

Could we move forward with SSR documentation? We can mention Apollo and DDP whatever it makes sense but the SSR guide should be as agnostic as possible about the data layer.

eric-burel:
Do you have a process to do so or a link to point me to? I am dead busy right now, but I can at least draft a step by step documentation, it's probably not that long to write. There are multiple concepts to understand but in the end SSR is not that many lines of code. At least a naive implementation without optimization.

eric-burel:
It seems to be as simple as PRing the doc repo but I can't tell where it fits yet.

PR #10820, opened 2019-12-16
Add Storybook to the roadmap

eric-burel:
Related to switching to NPM, it would be a good occasion to open up to a broader tooling ecosystem.

A big missing piece from the developer standpoint is Storybook. It's not just some nice tool, it literally changed the way we develop frontend in my company and all the companies where I've managed to introduce this magical tool.

Sadly it doesn't work correctly with Meteor, because it is based on a Webpack build. This issue haven't really been solved. We've added Storybook to Vulcan, but that's more a hack than an actual solution (basically we scrap Meteor imports + rely on Vulcan's naming convention to find package root...) so I am not personaly knowledgeable enough to help :/

Another great tool is Jest, it's way more efficient than Mocha thanks to its powerful CLI, but needs some webpack build.

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -199,6 +199,13 @@ as we already have for Blaze, React and Angular.
 - Status: -
 - PRs: -
 
+### "Webpack tooling": Storybook, Jest
+
+- Leaders: <you?>
+- Status: -
+- PRs: -
+
+
 ## Recently completed
 
 ### Different JS bundles for modern versus legacy browsers
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -199,12 +199,17 @@ as we already have for Blaze, React and Angular.
 - Status: -
 - PRs: -
 
-### "Webpack tooling": Storybook, Jest
+### Third-party tools with their own build steps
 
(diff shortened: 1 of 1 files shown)


sebakerckhof:
Just FYI, @benjamn posted some related thoughts here: https://github.com/meteor/meteor/pull/10811#issuecomment-564726713

Roadmap.md
```diff
@@ -199,6 +199,13 @@ as we already have for Blaze, React and Angular.
 - Status: -
 - PRs: -
 
+### "Webpack tooling": Storybook, Jest
```
benjamn: Could we call this something like "third-party tools with their own build steps"?

I definitely agree that Storybook should be the driving example, in part because we (Apollo) use it very heavily internally.

fgm:
FWIW, I've been using Jest with Meteor, both on the main project itself and with packages converted from Atmosphere to NPM (meteor_server_info and filog): I don't really see a need for specific features in Meteor to support it. Totally agree for Storybook too: it's a recently adopted tool but which caught fast.

PR #10823, opened 2019-12-17
Include Meteor.isSever constant in roadmap

edemaine:
I (and presumably others) often have code shared between client and server, with some blocked off by `if(Meteor.isServer)` and (more rarely) `if(!Meteor.isServer)`. It would be great if this constant is also handled in the tree shaking, and I imagine it's about as easy as `Meteor.isProduction` (but slightly different so worth mentioning).

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -25,7 +25,7 @@ Since Node.js 12 is scheduled to become the LTS version on October 1st, 2019, Me
 - Status: -
 - PRs: -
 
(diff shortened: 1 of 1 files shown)


sebakerckhof:
Could you also add: https://github.com/meteor/meteor/pull/10056 to the list of related PR's?

edemaine:
@sebakerckhof OK, I added mention of #10056. I don't think it belongs in the "PR" field, as I assume that's reserved for the PR implementing the overall feature, and I don't think #10056 does all of tree shaking. But let me know if I'm mistaken and I should change this.

Issue #10800, opened 2019-11-26
[1.8.2] Development environment build time increased 150%


theosp:
I am in the process of evaluating v1.8.2 for [JustDo.com](https://justdo.com) (upgrading from v1.8.1).

Development app start time takes 150% longer - from 20 seconds, it now increased to 50 seconds (2.9 GHz Intel Core i9, MBP 2018).

JustDo is comprised of 2 apps, one for the landing app and one for the web app itself.

The following is the output received when running the script that runs the two apps with the env var `METEOR_PROFILE=1`.

[JustDo v1.8.1 profile](https://drive.google.com/open?id=1HRtzhn1E2HRIS-cWFGurLzfQbu7srKDv)
[JustDo v1.8.2 profile](https://drive.google.com/open?id=1gkIM6ljW_6Czrif-Lopv9zttEQ9XteiZ)

The output isn't for the first run, but for the second run of the apps in both cases (to avoid special inits when changing meteor versions).

Notes:

* Setting env var `METEOR_DISABLE_OPTIMISTIC_CACHING=1` didn't help
* https://github.com/meteor/meteor/issues/10786 - Might be relevant?

sebakerckhof:
Are you sure these have been created in similar conditions? The fact that some things like reading in the SQLite package database etc all take considerably longer would suggest there might have been some heavy resource usage (either disk/memory or cpu) by something unrelated?

Since a lot of time is spend by the linker, could you also try running with:
`METEOR_TEST_PRINT_LINKER_CACHE_DEBUG=1` envar

theosp:
Hi thanks for the reply,

1. In the development environment we are not connecting to an SQL server, nor any other similar resources that can have similar effect on results.
2. This time, again, the output is from the *2nd* run of the environment, and not from the first run.
3. The outputs are from the exact same code, as the original output, only with the additional env var: METEOR_TEST_PRINT_LINKER_CACHE_DEBUG=1
4. To speed the process I published the results bottom line under each link.

I ran the test twice for each version, one with METEOR_DISABLE_OPTIMISTIC_CACHING=1 and one without.

[JustDo v1.8.1 profile](https://drive.google.com/open?id=1H5IVy7lPNMpxbj35mDn2bWkvc20c5zmB)
Total: 14,950 ms (Build App) (Web app)
Total: 6,045 ms (Build App) (Landing  app)

[...]

Issue #10261, opened 2018-10-10
[1.8] Meteor server blocked while rebuilding web.browser.legacy


sebakerckhof:
While Meteor 1.8 should improve rebuild times by delaying `web.browser.legacy`, in practice this doesn't seem to work for me.

Normally (e.g. in 1.7) the `Rebuild App` step is followed immediately with the `Server startup` step.
However, now, the process seems to hang after the `Rebuild App` step for the duration of the legacy step in which I can't reach my application. Only after the legacy build completes, the server starts, immediately followed by: `Finished delayed build of web.browser.legacy in 29752ms`. This of course renders the whole delayed building rather useless.

Also, the total rebuild time is now 20 seconds for `web` and 29 seconds for `web.legacy`.
While in Meteor 1.7 the total rebuild time for both combined was only 11 seconds. But I'll create a separate issue for that.

I've attached the cpu profile received from `qualia:profile`.
[profiles.zip](https://github.com/meteor/meteor/files/2464043/profiles.zip)

benjamn:
The `web.browser.legacy` build is running in a separate process (the parent/build process) from the server startup code, so I believe those two chunks of work are racing in parallel, without blocking each other. If I hack the delayed build to take a long time (30 seconds) I can definitely see changes (or manually reload the app ) long before the legacy bundle finishes building. Of course there could be bugs that are affecting your experience and not mine, but I'm not sure what they could be, yet.

In any case, I think we should look into that performance regression first (the other issue)!

benjamn:
One quick follow-up observation: if the legacy build is still running, and you make _another_ change, the build system won't pick that change up and start rebuilding until the previous legacy build is finished. So perhaps we should try to make the legacy builds cancelable somehow.

KoenLav:
@benjamn for what it's worth; development speed on 1.8 is just terrible now.

Even though the build is delayed, we are still triggering two builds for every change. When a lot of small changes are made this significantly slows down the CPU. 

In my opinion there should definitely be an option to disable legacy building on development (or maybe it should even be the default).

sebakerckhof:
@benjamn That's also what I would expect to happen, but here's a video of what is actually happening in my setup. Notice that around 00:39 the rebuild app finishes. Then I try to refresh the page, but the server is unresponsive for another 20 seconds until the legacy build finishes.

[meteor.mp4.zip](https://github.com/meteor/meteor/files/2466109/meteor.mp4.zip)

benjamn:
[...]

Issue #10690, opened 2019-09-13
TEST_CLIENT=0 will still build client and refresh


eric-burel:
Hi,
Server side tests are longer to start than necessary when using TEST_CLIENT=0. According to the log, the app is built for browser too, and the client get refreshed on changes, which is unnecessary.
Preventing this unnecessary build would provide a better testing experience.

- Meteor 1.8.1
- Ubuntu 18.04
- Expected: build only the server code when TEST_CLIENT=0
- Actual: app is also built for browser and client is refreshed

eric-burel:
Not stale

filipenevola:
@eric-burel thanks, I just labeled as pinned then stale bot will ignore it 😉

sebakerckhof:
I could tackle this as part of https://github.com/meteor/meteor/pull/10824

eric-burel:
It's from `meteor-mocha` actually, so maybe not a Meteor issue per se. As far as I understand how tests work, it seems that the test runner is triggered after the app is built and running, so it can't take control on the build process.

So maybe it could be interesting to handle this env variable directly in Meteor, in order to make the build faster during tests ?

sebakerckhof:
I'd rather keep the same semantics already used by https://github.com/meteor/meteor/pull/10824 (using an `--exclude-archs` cli argument). But it should achieve what you want.

eric-burel:
Sounds great indeed

PR #10824, opened 2019-12-18
Allow to exclude web architectures in development mode

sebakerckhof:
This commit implements an `--exclude-archs` arg to the run command which allows you to pass a comma-seperated list of architectures to exclude.

Motivation:
Improve rebuild speeds, by allowing to leave out certain web architectures during development.

Meteor rebuild speeds, although greatly improved in recent versions, is still a common point of frustration as seen in the issues (https://github.com/meteor/meteor/issues/10800) and on the forum (https://forums.meteor.com/t/2x-to-3x-longer-build-time-also-on-testing-since-1-8-2-update/51028/3)

One reason for slow rebuilds is high memory pressure (https://github.com/meteor/meteor/issues/9568).
The web.browser.legacy architecture is currently always being build while most of us will only test this every once in a while.
So while we don't need it most of times during development it does add to the high memory usage.
Furthermore, even though these builds are delayed, long-running synchronous tasks have the potential to block the server startup (https://github.com/meteor/meteor/issues/10261).
This last issue can be partially, but not fully, improved by https://github.com/meteor/meteor/pull/10427
Therefore a commonly requested feature has been to disable the legacy build (https://github.com/meteor/meteor-feature-requests/issues/333)

However, we may also temporarily disable cordova builds and we may even want to disable just the web.browser build if we're just debugging a bug in the legacy build.
Therefore I chose to create the more generic `--exclude-archs` option.

### DIFF
--- a/History.md
+++ b/History.md
@@ -21,6 +21,15 @@ N/A
   * cordova-plugin-splashscreen (included by mobile-experience > launch-screen) from 4.1.0 to 5.0.3 [release notes](https://github.com/apache/cordova-plugin-splashscreen/blob/master/RELEASENOTES.md#503-may-09-2019)
   * cordova-plugin-statusbar (included by mobile-experience > mobile-status-bar) from 2.3.0 to 2.4.3 [release notes](https://github.com/apache/cordova-plugin-statusbar/blob/master/RELEASENOTES.md#243-jun-19-2019)
 
+* You can now pass an `--exclude-archs` option to the `meteor run` and
+  `meteor test` commands to temporarily disable building certain web
+  architectures. For example, `meteor run --exclude-archs web.browser.legacy`.
+  Multiple architectures should be separated by commas. This option can be
+  used to improve (re)build times if you're not actively testing the
+  excluded architectures during development.
+  [Feature #333](https://github.com/meteor/meteor-feature-requests/issues/333),
+  [PR #10824](https://github.com/meteor/meteor/pull/10824)
+
 ## v1.9, 2020-01-09
 
 ### Breaking changes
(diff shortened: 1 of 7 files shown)


sebakerckhof:
ping @veered

RobertLowe:
any `METEOR_PROFILE=1`s to compare?

sebakerckhof:
@RobertLowe Here's 2 profiles from a cold start:

[182-standard-cold.txt](https://github.com/meteor/meteor/files/3988430/182-standard-cold.txt)
[182-nolegacy-cold.txt](https://github.com/meteor/meteor/files/3988431/182-nolegacy-cold.txt)

During development/rebuilds it's using 1/3rd less memory (about 1gb for our application). In case you're low on RAM, this will help for rebuild times as well.

RobertLowe:
@sebakerckhof nice, I see about 100s saved build time, very nice! Curious, did you clean build caches between?

sebakerckhof:
@RobertLowe yes both were after a `meteor rebuild` which clears the caches.
Anyway, the metric for a cold boot isn't very interesting IMO, since you don't do that often. I'm way more interested in rebuild times. And although the legacy rebuild is delayed, the additional memory usage for the legacy build can really slow down the modern rebuild if you're short on memory.

RobertLowe:
@sebakerckhof it's still a welcome improvement!! Thanks for getting this out there, thanks @benjamn for targeting this for 1.8.4!! 👍 🎉

hexsprite:
Will this work for running tests as well? I noticed that they can also build archs that I don't care about...

sebakerckhof:
@hexsprite I've added the option for tests as well. Should also address https://github.com/meteor/meteor/issues/10690

sebakerckhof:
Btw, currently this will crash if you do something like:
`meteor run android --exclude-archs web.cordova`
Because well, this doesn't make much sense, so we can handle it in multiple ways:
1. Ignore the exclude archs for web.cordova
2. Don't start the cordova runner, just start the server
3. Throw an error saying this combination is invalid

Any preferences to what seem best?

hexsprite:
testing this out when running tests. sure makes a huge difference in build times and memory pressure as you say!

RobertLowe:
@benjamn  `1.9.1-rc.0` eta? 😛 this feature is significantly helpful

dosomder:
@sebakerckhof Why only for run / test? Would be useful to disable legacy with build too

Issue #10684, opened 2019-08-31
App.addResourceFile cannot handle relative paths for the source


brucejo75:
Undocumented feature `App.addResourceFile` implies that it can handle a src **path** in it's documentation [here](https://github.com/meteor/meteor/blob/devel/tools/cordova/builder.js#L770-L777).

But it does not work, it can only handle a source file name.

Maybe @eskan, can take a look?

### Details
This [code](https://github.com/meteor/meteor/blob/devel/tools/cordova/builder.js#L393-L416) looks like the culprit:
```javascript
  configureAndCopyResourceFiles(resourceFiles, iosElement, androidElement) {
    _.each(resourceFiles, resourceFile => {
      // Copy file in cordova project root directory
      var filename = path.parse(resourceFile.src).base;
      files.copyFile(
        files.pathResolve(this.projectContext.projectDir, resourceFile.src),
        files.pathJoin(this.projectRoot, filename));
      // And entry in config.xml
      if (!resourceFile.platform ||
          (resourceFile.platform && resourceFile.platform === "android")) {
        androidElement.element('resource-file', {
          src: resourceFile.src,
          target: resourceFile.target
        });
      }
      if (!resourceFile.platform ||
          (resourceFile.platform && resourceFile.platform === "ios")) {
        iosElement.element('resource-file', {
          src: resourceFile.src,
          target: resourceFile.target
        });
      }
    });
  }
```

Essentially, the code copies from it's source location
```javascript
files.pathResolve(this.projectContext.projectDir, resourceFile.src)
```
then places it at the root of the cordova build:
```javascript
files.pathJoin(this.projectRoot, filename))
```
Which is fine, but then it creates a `resource-file` directive that is looking for the resource file relative to the root of the build tree:
```javascript
        androidElement.element('resource-file', {
          src: resourceFile.src,
          target: resourceFile.target
        });
```
So if you place your resources at the root of the meteor project tree all will work.  To make it work with resources in folders either:
1. Need to make the files copy to a path relative to the cordova build root.
2. Change the `resource-file` directive to grab the file from the root.

Either way there is the potential for collision, but I think I like # 1.

brucejo75:
PR #9748 has the problem.

eskan:
[...]

PR #10829, opened 2019-12-19
Fix/Change the way resource file are added to cordova project

eskan:
Hi,
when i first introduced the app.AddResourceFile to cordova config file i didn't notice some side effect.
When a file is copied in a cordova build, it's done in 2 steps :
- the file is copied from the meteor project to the cordova-build
- then the file is copied from cordova build to platform specific build
Before this PR files are copied to the root of the cordova build. this way conflicts can happened.

This PRs introduce an intermediate directory (resource-files) where files will be copied with their full originated path to prevent conflict at this intermediate step. I choose to keep full path to allow resources files to live outside the meteor project but i **don't know if Windows can handle** such behavior.

https://github.com/meteor/meteor/issues/10684

@brucejo75 may take a look to this.


skirunman:
@filipenevola Can we get this in v1.10 along with other Cordova updates?

filipenevola:
@skirunman yes, let's wait for @renanccastro review.

tools/cordova/builder.js
```diff
@@ -392,23 +392,23 @@ export class CordovaBuilder {
...
-        files.pathResolve(this.projectContext.projectDir, resourceFile.src),
-        files.pathJoin(this.projectRoot, filename));
+        filepath,
+        files.pathJoin(this.projectRoot, "/resource-files/", filepath));
       // And entry in config.xml
       if (!resourceFile.platform ||
           (resourceFile.platform && resourceFile.platform === "android")) {
         androidElement.element('resource-file', {
-          src: resourceFile.src,
+          src: files.pathJoin("./resource-files/", filepath),
```
renanccastro: Will it work without the "./" here and on line 411? This would make the chances higher of it working on windows... I can't test it on windows, but AFAIK the meteor fs code doesn't seem to be dealing with this POSIX convention. It seems to be only dealing with slashes. Nice work btw 👍 thanks!

```suggestion
          src: files.pathJoin("resource-files/", filepath),
```

eskan:
@renanccastro i've just seen your review right now. 
you're right about "/" they are not necessary with path.join() so i removed them. I guess it should work with windows.
@brucejo75 may you confirm it still works for you ?

brucejo75:
@eskan,

I don't use Windows (per se).  I am using [WSL](https://docs.microsoft.com/en-us/windows/wsl/about) which is essentially Ubuntu.  Plus I also use OSX (Darwin).

The plan is fine with me.  I do not have a setup to easily test Meteor core changes outside of releases.

filipenevola:
hi @eskan could you sync your branch with the latest `devel`? Currently, the tests are not passing.

eskan:
@filipenevola sync is done

PR #10838, opened 2019-12-23
Realpath performance

zodern:
With a smaller app on Windows, these changes reduces time spent by realpath from ~200 - 300 ms to 40ms on rebuilds, and saves ~750ms on the initial build. `realpath` works by using many lstat calls, which are slower on Windows than mac/linux.

When creating the list of files for each architecture, Meteor uses a symlink loop checker which would call `realpath` for each folder. It will now call `lstat` once for each folder and then call `realpath` for folders that are symlinks.

Meteor would also call realpath for each file in `node_modules/.bin` to check if it is a symlink. Since npm copies files into .bin instead of using symlinks on Windows, Meteor will now use lstat first, and then call `realpath` if it is a symlink. This does slightly increase the work done on linux/osx, but the additional lstat call is already cached from earlier in the build process.

The remaining use during rebuilds is calling `realpath` for each node_modules folder that is copied into the built app. I left this as is since it isn't clear why this is done and only adds ~20ms in the apps I profiled.

### DIFF
--- a/tools/fs/optimistic.ts
+++ b/tools/fs/optimistic.ts
@@ -16,6 +16,7 @@ import {
   readdir,
   dependOnPath,
   findAppDir,
+  realpath,
 } from "./files";
 
 // When in doubt, the optimistic caching system can be completely disabled
@@ -272,6 +273,24 @@ export const optimisticLStatOrNull = makeCheapPathFunction(
     }
   },
 );
+
+export const optimisticRealpath = makeOptimistic('realpath', realpath);
+export const optimisticRealpathOrNull = makeOptimistic('realpathOrNull', (
+  path: string,
+  options?: Parameters<typeof optimisticRealpath>[1],
+) => {
+  try {
+    return realpath(path, options)
+  } catch (e) {
+    if (e.code !== "ENOENT") {
+      throw e;
+    }
+  }
+
+  dependOnParentDirectory(path);
+
+  return null;
+});
 
 export const optimisticReadFile = makeOptimistic("readFile", readFile);
 export const optimisticReaddir = makeOptimistic("readdir", readdir);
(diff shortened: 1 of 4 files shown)


tools/fs/optimistic.ts
```diff
@@ -273,6 +274,24 @@ export const optimisticLStatOrNull = makeCheapPathFunction(
...
+) => {
+  try {
+    return realpath(path, options)
+  } catch (e) {
+    if (e.code !== "ENOENT") {
+      throw e;
+    }
+  }
+
+  dependOnParentDirectory(path);
```
benjamn: Yeah, I'm afraid `realpath` depends on more information than just the `path` and the parent directory. We could reimplement it here, using `optimisticLStatOrNull` instead of `fs.lstat`, but that would be trickier.

tools/isobuild/package-source.js
```diff
@@ -213,14 +213,44 @@ var getExcerptFromReadme = function (text) {
 class SymlinkLoopChecker {
```
benjamn: I wonder if we might be able to get rid of the `SymlinkLoopChecker` by just limiting the maximum depth of recursion to something "ridiculous" like 100? Symlink loops are a relatively rare occurrence.

zodern: It seems to work correctly, but it does increase the size of the client bundle since it uses the most deeply nested version of the file it finds. You can see the app.js contents at https://gist.github.com/zodern/68a9cd266c42d7e1719dda4114d94b64

Instead of limiting the depth we could listen for the ELOOP error. Node throws that error once it gets into a depth of ~30 in a symlink loop

tools/fs/optimistic.ts
zodern: I've removed the optimisticRealpath since it was only used for two things: the realpath of each node_modules folder, which doesn't have much of an impact on build performance, and for symlinks in `node_modules/.bin`, which there normally are none Windows.

Issue #10794, opened 2019-11-23
Capacitor User Agent is recognized as legacy instead of modern


KoenLav:
When loading our app in Capacitor on Android the following user agent is passed:

`Mozilla/5.0 (Linux; Android 9; Moto Z3 Play Build/PPWS29.131-27-1-11; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/78.0.3904.108 Mobile Safari/537.36`

Meteor recognizes this as a legacy client, whereas it should be clear that this is a recent version of Chrome.

Is this something we can fix from the Meteor side of things?

filipenevola:
Hi @KoenLav , you should be able to override this behavior calling `setMinimumBrowserVersions` from `modern-browsers` package in the server startup.

Read more here https://github.com/meteor/meteor/blob/c071e11032dce2b328007a928a32a63ea220b74f/History.md#changes-11

KoenLav:
I'm not quite clear on your instructions; looking at my User Agent it should already recognized that I'm running Chrome 78.

I don't see how Meteor could be considering Chrome 78 as legacy?

sebakerckhof:
@KoenLav can you try locally cloning the `modern-browsers` package with my PR from above and see if that resolves it for you?

KoenLav:
@sebakerckhof fixed indeed!

filipenevola:
Published modern-browsers@0.1.5

PR #10844, opened 2020-01-06
Add chromeMobileWebView alias

sebakerckhof:
Capacitor (and possibly others) are identified as chromeMobileWebView. This is identical (also in numbering) to e.g. chromeMobile and can therefore be used as alias.

Fixes https://github.com/meteor/meteor/issues/10794


filipenevola:
Published modern-browsers@0.1.5

Issue #10756, opened 2019-11-02
server-render: search and query are missing from sink.request.url


mattpen:
### The version of Meteor showing the problem.
1.8.1

### The last version of Meteor where the problem did _not_ occur, if applicable.
n/a

### The operating system you're running Meteor on.
RHEL7 (production), MacOS 10.14 (development)

### The expected behavior.
`sink.request.url.search` should be equal to 'foo=bar' when `onPageLoad` runs on the server if the http request ends in `?foo=bar`.

### The actual behavior.

If you are running meteor with a different value for ROOT_URL, `sink.request.url.search` is an empty string always.

### A **simple** reproduction!

```
git clone https://github.com/mattpen/meteor-problem
meteor npm i
meteor
```
Then observe that a request for `http://localhost:3000/?foo=bar` prints a `sink.request.url` object with the search (and query) properties as expected.

Quit the meteor instance and run:
```
ROOT_URL=http://localhost:3000/mount meteor
```
Then observe that a request for `http://localhost:3000/mount/?foo=bar` prints a `sink.request.url` object where search is an empty string and query is an empty object.  Additionally, the search string is missing from the path and href properties.

mattpen:
This is similar to #8980, but I'm not using a hash so I would expect the search string to show up on the server.

mattpen:
This issue is not stale, seems to be a bug in your latest release

filipenevola:
@mattpen thanks, I just labeled as pinned then stale bot will ignore it 😉

sebakerckhof:
@mattpen I made a PR, but in the meantime you can locally clone the webapp package to your projects packages folder.

mattpen:
@sebakerckhof - Thanks! I'll check out the PR as soon as I can.

PR #10845, opened 2020-01-06
Don't lose the query string when using a path prefix

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/10756


benjamn:
(The tests didn't actually fail, I just canceled the builds since I squashed everything into `devel`.)

Issue #10705, opened 2019-09-23
browser-policy README mentions that "eval" is disabled by default


mitar:
On Meteor 1.8.1 this is maybe true for `browser-policy` package, but it is not really true for any realistic Meteor app these days because apps use `ecmascript` which requires `dynamic-import`, which enabled `eval` by default. So in practice, by default, `eval` is enabled. See #10704.

It should be probably clarified in the README that the package by default keeps it disabled, but Meteor apps by default enable it because of `dynamic-import` and that you should remove `dynamic-import`  package if you want to get it disabled (I hope that resolution of #10704 will be that one can remove `dynamic-import`).

See: https://github.com/meteor/meteor/tree/devel/packages/browser-policy

mitar:
Still relevant.

PR #10848, opened 2020-01-07
Warn about dynamic-imports and eval

sebakerckhof:
Fixes https://github.com/meteor/meteor/issues/10705

### DIFF
--- a/packages/browser-policy/README.md
+++ b/packages/browser-policy/README.md
@@ -31,10 +31,12 @@ For most apps, we recommend that you take the following steps:
 * Add `browser-policy` to your app to enable a starter policy. With this starter
 policy, your app's client code will be able to load content (images, scripts,
 fonts, etc.) only from its own origin, except that XMLHttpRequests and WebSocket
-connections can go to any origin. Further, your app's client code will not be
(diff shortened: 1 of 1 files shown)


packages/browser-policy/README.md
```diff
@@ -31,10 +31,12 @@ For most apps, we recommend that you take the following steps:
...
 policy, your app's client code will be able to load content (images, scripts,
 fonts, etc.) only from its own origin, except that XMLHttpRequests and WebSocket
-connections can go to any origin. Further, your app's client code will not be
-able to use functions such as `eval()` that convert strings to code. Users'
+connections can go to any origin. Users'
 browsers will only let your app be framed by web pages on the same origin as
-your app.
+your app. Further, your app's client code will not be
+able to use functions such as `eval()` that convert strings to code. 
+However, note that if you also use the `dynamic-imports` package, this limitation
```
mitar: I think this maybe be a bit misleading. It could be read like that if you use `dynamic-imports` package it somehow enables a workaround which allows one to use `eval` despite it being disabled by the policy, not that it disables the "prevent eval` policy.

PR #10837, opened 2019-12-23
Release 1.8.4

filipenevola:
https://github.com/meteor/meteor/milestone/67

Versions to bump:
- [x] mobile-status-bar
- [x] launch-screen
- [x] mobile-experience
- [x] cordova-plugin-meteor-webapp

PR #10810, opened 2019-12-06
Update Cordova to 9.0.1

filipenevola:
We want to update:
- [x] Lib from 7.1.0 to 9.0.1 [release notes](https://github.com/apache/cordova-lib/blob/master/RELEASENOTES.md)
- [x] Common from 2.1.1 to 3.2.1 [release notes](https://github.com/apache/cordova-common/blob/master/RELEASENOTES.md)
- [x] Android from 7.1.4 to 8.1.0 [release notes](https://github.com/apache/cordova-android/blob/master/RELEASENOTES.md)
- [x] iOS from 4.5.5 to 5.1.1 [release notes](https://github.com/apache/cordova-ios/blob/master/RELEASENOTES.md)
- [x] cordova-plugin-wkwebview-engine from 1.1.4 to 1.2.1 [release notes](https://github.com/apache/cordova-plugin-wkwebview-engine/blob/master/RELEASENOTES.md#121-jul-20-2019)
- [x] cordova-plugin-whitelist from 1.3.3 to 1.3.4 [release notes](https://github.com/apache/cordova-plugin-whitelist/blob/master/RELEASENOTES.md#134-jun-19-2019)
- [x] cordova-plugin-splashscreen (included by mobile-experience > launch-screen) from 4.1.0 to 5.0.3 [release notes](https://github.com/apache/cordova-plugin-splashscreen/blob/master/RELEASENOTES.md#503-may-09-2019)
- [x] cordova-plugin-statusbar (included by mobile-experience > mobile-status-bar) from 2.3.0 to 2.4.3 [release notes](https://github.com/apache/cordova-plugin-statusbar/blob/master/RELEASENOTES.md#243-jun-19-2019)

Related PRs (cordova-9 branch):
https://github.com/meteor/cordova-plugin-meteor-webapp/pull/88

Related issues #10809 #10626

Follow the list of known tasks:
- [x] (Fixed on 1.7.1-beta.0) Don't use context.requireCordovaModule to load non-Cordova modules ([error log](https://gist.github.com/filipenevola/5b80f7dfeecba52c0b747215fa93e283))
  - [x] Migrate context.requireCordovaModule to require on [cordova-plugin-meteor-webapp](https://github.com/meteor/cordova-plugin-meteor-webapp/)
- [x] docker images for CircleCI
  - [x] Android 28 / Node 8
  - [x] Android 28 / Node 12

Notes:
[...]

PR #10857, opened 2020-01-13
Rebase PR #10810 (Cordova 9) against current devel branch.

benjamn:
Since #10810 was merged into `release-1.8.4`, which contains [a lot of extraneous changes](https://github.com/meteor/meteor/pull/10837), I wanted to try recreating just the Cordova-related parts, so that we can merge them cleanly into `devel`.

### DIFF
--- a/packages/launch-screen/package.js
+++ b/packages/launch-screen/package.js
@@ -10,7 +10,7 @@ Package.describe({
 });
 
 Cordova.depends({
-  'cordova-plugin-splashscreen': '4.1.0'
+  'cordova-plugin-splashscreen': '5.0.3'
 });
 
 Package.onUse(function(api) {
--- a/packages/launch-screen/package.js
+++ b/packages/launch-screen/package.js
@@ -6,7 +6,7 @@ Package.describe({
   // between such packages and the build tool.
   name: 'launch-screen',
   summary: 'Default and customizable launch screen on mobile.',
-  version: '1.1.1'
+  version: '1.1.2-beta.0'
 });
 
 Cordova.depends({
--- a/packages/launch-screen/package.js
+++ b/packages/launch-screen/package.js
@@ -6,7 +6,7 @@ Package.describe({
   // between such packages and the build tool.
   name: 'launch-screen',
   summary: 'Default and customizable launch screen on mobile.',
-  version: '1.1.2-beta.0'
+  version: '1.2.0'
 });
 
 Cordova.depends({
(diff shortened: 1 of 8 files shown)


packages/launch-screen/package.js
```diff
@@ -6,11 +6,11 @@ Package.describe({
   // between such packages and the build tool.
   name: 'launch-screen',
   summary: 'Default and customizable launch screen on mobile.',
-  version: '1.1.1'
+  version: '1.1.2-beta.0'
```
benjamn: Can we remove these `-beta.0` suffixes from `launch-screen`, `mobile-experience`, and `mobile-status-bar` (commit b8e3d083c9a01306650bb434a7b5c92a390b7aad)?

Also, will the new versions of these packages work with the current version of Cordova, or only with Cordova 9? If they only work with Cordova 9, we should be bumping the minor versions of these packages, so folks using Meteor 1.8.3 won't be exposed to the new versions yet.

packages/webapp/package.js
```diff
@@ -22,7 +22,7 @@ Npm.strip({
 Cordova.depends({
   'cordova-plugin-whitelist': '1.3.3',
   'cordova-plugin-wkwebview-engine': '1.1.4',
-  'cordova-plugin-meteor-webapp': '1.7.1-beta.0'
+  'cordova-plugin-meteor-webapp': '1.7.1-beta.1'
```
benjamn: Just noting that @filipenevola plans to remove this `-beta.1` suffix soon.

benjamn:
I was somewhat surprised to see
```
cordova-builds: cordova builds with server options ... 
  ... fail!
  ... retrying (2 tries remaining) ...
  ... fail!
  ... retrying (1 try remaining) ...
  ... ok (273652 ms)
```
in https://circleci.com/gh/meteor/meteor/40384, specifically that the test failed twice and then succeeded! The success is definitely good news, but we might want to keep an eye on these failures. My guess is that there's a `run.waitSecs` timeout in [the `cordova-builds` test](https://github.com/meteor/meteor/blob/b8e3d083c9a01306650bb434a7b5c92a390b7aad/tools/tests/cordova-builds.js#L33) that's just a little bit too short, so it sometimes fails and sometimes succeeds. If that's accurate, then we could make this test more reliable by simply increasing the timeout to something that's less likely to fail.

benjamn:
Great, the `cordova-builds` self-test passed without any failures that time: https://circleci.com/gh/meteor/meteor/40405

benjamn:
@filipenevola Merge at will! 🚀

PR #10858, opened 2020-01-13
Update Node.js to v12.14.1 and npm to v6.13.6.

benjamn:
These changes could have been included in a Meteor 1.9.1 release, but neither update is so urgent that they need to be released sooner than 1.10.


jamesmillerburgess:
Here is one issue that bit us when updating to `1.9.0`: https://github.com/manuelbieh/geolib/issues/208

Node 12.14.1 fixes that issue, so we are upgrading now to `1.10-beta.0`.

I'm not sure how many projects will have this problem. If many will, then it might be good to do a patch release for this node version bump.

PR #10864, opened 2020-01-15
Reduce EJSON.clone's when observing query changes

sebakerckhof:
This reduces the amount of EJSON.clone's in the live query system a lot.

This basically does 2 things:
- Removed unnecessary deep clones when passing data to the CachingChangeObserver
- Add a way to tell the multiplexer which observeChange handles do not mutate the arguments and therefore do not require a deep clone. This makes the biggest difference since by far the most observeChanges come from cursor publications, and these do not need deep clones, since the DDP liveserver already deep clones the data.

This should still have the same guarantees when you use the public API. Packages which modify the internals of Meteor may break the assumptions which make this optimization possible. I verified this is not the case for Kadira and redis-oplog.

Here is an example you can run which logs the amount of clones saved for a simple use case: https://github.com/sebakerckhof/meteor-mongo-optim (open the app in a couple of tabs and click the button a couple of times, the server console will log the difference in `EJSON.clone` calls).


sebakerckhof:
This should now be ready for review
I have now completed verifying the safety of these changes.
Because the call flow is somewhat complex, I made a diagram which reviewers might find useful:
livequery-flow

I've also added tests to ensure these assumptions stay safe in the future.

The subscription (`_publishCursor`) callbacks (and everything onwards until it gets send out) is already covered by:
https://github.com/meteor/meteor/blob/65e292d8fe98473ae32bdfdd6709b6e57a830938/packages/ddp-client/test/livedata_tests.js#L1032

Everything from Observe/ObserveChanges was already covered to some extend by the `mongo-livedata - scribbling` test ( https://github.com/meteor/meteor/blob/65e292d8fe98473ae32bdfdd6709b6e57a830938/packages/mongo/mongo_livedata_tests.js#L533 )
A more extended version has been added:
https://github.com/meteor/meteor/pull/10864/commits/a94aa418f02573ebf99b12944183d015567f3ce6#diff-d3d421e3f71f0e59e3b43e7bad95408fR564

Tested on a real app today saved about ~15% EJSON.clone calls (counted globally, so not only livequery-related clones).

mitar:
Nice. So this is then automatically enabled for observes which happen when you return a cursor from the publish function on the server?

packages/minimongo/local_collection.js
```diff
@@ -669,17 +669,18 @@ LocalCollection._CachingChangeObserver = class _CachingChangeObserver {
...
         addedBefore: (id, fields, before) => {
-          const doc = EJSON.clone(fields);
+          // Take a shallow copy since the top-level properties can be changed
+          const doc = { ...fields };
 
           doc._id = id;
 
           if (callbacks.addedBefore) {
-            callbacks.addedBefore.call(this, id, fields, before);
+            callbacks.addedBefore.call(this, id, EJSON.clone(fields), before);
```
mitar: Hm, why are you here both doing the shallow clone and still EJSON.cloning fields?

mitar:
So [I will have to update code here](https://github.com/peerlibrary/meteor-reactive-mongo/blob/master/server.coffee#L36)?

sebakerckhof:
> Nice. So this is then automatically enabled for observes which happen when you return a cursor from the publish function on the server?

Yes, since the DDP server will already do the necessary cloning.

packages/minimongo/local_collection.js
sebakerckhof: There are 2 things to make sure of:
1. We don't modify the object we get in and save to the `docs` of `CachingChangeObserver`. Therefore we take a shallow clone. Before this, this was not guaranteed by the CachingChangeObserver, so it was always doing a deep clone outside, e.g.:
https://github.com/meteor/meteor/blob/devel/packages/mongo/observe_multiplex.js#L162
But this is not necessary. At least as long as you stick to the contract a shallow clone is sufficient https://github.com/meteor/meteor/blob/d072b5b43256005dcb3d96dead5474c1f4893a8d/packages/minimongo/local_collection.js#L640-L642 

2. We should also be sure that if there are callbacks on `CachingChangeObserver`, they don't modify the doc. Therefore we take a deep clone, because in the only case the Meteor code has callbacks here, it modifies the arguments in an unknown way:
https://github.com/meteor/meteor/blob/d072b5b43256005dcb3d96dead5474c1f4893a8d/packages/minimongo/local_collection.js#L1222

Before this, this was always a deep clone, even if there were no callbacks, like here:
https://github.com/meteor/meteor/blob/d072b5b43256005dcb3d96dead5474c1f4893a8d/packages/mongo/observe_multiplex.js#L17-L18

So actually the safety guarantees of CachingChangeObserver have improved , while simultanously reducing the number of deep clones :)

sebakerckhof:
> So [I will have to update code here](https://github.com/peerlibrary/meteor-reactive-mongo/blob/master/server.coffee#L36)? And [here](https://github.com/peerlibrary/meteor-reactive-publish/blob/master/server.coffee#L53)? Those are two cases I found in my packages with grepping.

To get the most benefits, you should indeed accept and pass down the options. It will still safely work without and it should already safe you some deep clones, but most will be saved if it can pass down the `nonMutatingCallbacks` option.

One alternative I was thinking of was to save the `nonMutatingCallbacks` option just on the callbacks object instead of an extra `options` argument. In that case you wouldn't need to change anything.

SimonSimCity:
@sebakerckhof do you know if and how deep this would affect https://github.com/cult-of-coders/redis-oplog? If you haven't already had a look at it I could give it a try ...

sebakerckhof:
@SimonSimCity I haven't tried it, but redis-oplog copies the multiplexer ( https://github.com/cult-of-coders/redis-oplog/blob/master/lib/mongo/ObserveMultiplex.js ). So it should not be affected (and as explained above, the changes to CachingChangeObserver keep the guarantees). But it will not see the benefits, or they should copy them.

PR #10870, opened 2020-01-15
Roadmap - Update Blaze Tutorial

jankapunkt:
Add Jan Küster and Harry Adel to take on the Blaze tutorials.

We would love to update the Blaze Tutorials to work with >=1.9 and to reflect latest best practices. We use Blaze for a long time and still have projects going, where Blaze plays an important role.

A follow-up question would be, if the TODOs App would have to be updated as well?


filipenevola:
Yes, ideally we should update it as well.

Issue #10468, opened 2019-02-25
Meteor 1.8.0.2: TypeError: Cannot set property 'setWebsocketTimeout' of null


c0desolut1ons:
Version of Meteor: 1.8.0.2.
Operating System: Windows Server 2012, Linux Centos

When we deploy and start meteor app we log error:

Meteor log:
```
TypeError: Cannot set property 'setWebsocketTimeout' of null
    at Server.<anonymous> (packages/ddp-server/stream_server.js:99:5)
    at emitOne (events.js:116:13)
    at Server.emit (events.js:211:7)
    at App.emit (c:\cmms.bee\prod\meteor\bundle\programs\server\npm\node_modules\meteor\ddp-server\node_modules\sockjs\lib\sockjs.js:196:29)
    at c:\cmms.bee\prod\meteor\bundle\programs\server\npm\node_modules\meteor\ddp-server\node_modules\sockjs\lib\transport.js:111:25
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

Direct file link: https://github.com/meteor/meteor/blob/f04160aea83f272cd5bfe8b7c3782e8c2b28255b/packages/ddp-server/stream_server.js#L99

More info:
SockJS (maybe) related issue: https://github.com/sockjs/sockjs-node/issues/121
The right way for a client to reconnect after the server does a socket.disconnect( true )
https://github.com/socketio/socket.io/issues/2476

_Is there any way how to make this work?
Thanks._

kombadzomba:
Please solve this. Don't just ignore it and close it like it never happened. No one from Meteor team has ever responded to this issue at all. We are still facing this problem in production and have serious meteor issues since this started. There was no response on this for a long time so we started to make plans on substituting meteor with something else. Please don't make us speed that process up.

filipenevola:
Hi @kombadzomba thanks for bringing this up. I'm Meteor Evangelist working for Meteor Software (the new Tiny company behind Meteor).

I'm also a Meteor user for a long time and I never saw this error in my apps, maybe you have something different in your application.

Can you create somehow a reproduction? I would like to understand this issue in depth.

We are committed to Meteor (see the roadmap [published today](https://github.com/meteor/meteor/blob/devel/Roadmap.md)) and you are not going to need to substitute Meteor, at least not because of issues 😉 

I already marked this issue as confirmed then stale bot is going to ignore it.

[...]

PR #10891, opened 2020-01-30
Fixes bug #10468: ddp-server guards against null argument in sockjs connection handler

markreid:
Defensive code to handle a scenario in which the sockjs `on('connection')` handler is passed `null` instead of a socket object.

This issue has popped up a couple of times in the past few years but doesn't look like it will be resolved in sockjs-node:
https://github.com/sockjs/sockjs-node/issues/121

And appears to be happening in recent versions of Meteor (1.8 and 1.9):
https://github.com/meteor/meteor/issues/10468

I couldn't really determine a way to write a test for this, apologies in advance.

### DIFF
--- a/packages/ddp-server/stream_server.js
+++ b/packages/ddp-server/stream_server.js
@@ -86,6 +86,12 @@ StreamServer = function () {
   self._redirectWebsocketEndpoint();
 
   self.server.on('connection', function (socket) {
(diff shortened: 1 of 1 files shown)


filipenevola:
I just published `ddp-server@2.3.1` including this PR content.

Issue #10892, opened 2020-01-31
Error loading npm module named 'config' on meteor 1.9


claudiosdc:
I have a project that was using meteor version 1.8.2, and I updated it to use meteor version 1.9. After the update, when I tried to run the application, I got the following error message: ***Error: Cannot find module 'config'***.

The referred module, named 'config', is one of the several npm modules that are included in my application.

I double checked to make sure that the 'config' module was properly installed, and it was. The error, however, still persisted.

Then I have decided to create a simple meteor project to try and investigate what was going on, and I have come to the following conclusion:

1. The issue is associated with the fact that the module being loaded is named 'config'. I have tested loading the exact same module, only renaming it to something different than 'config' (e.g. 'app-config' or 'config1'), and it works.

2. The error occurs no matter what loading mechanism is used: either 'require' or 'import'.

3. The error does **NOT** take place if the 'config' module is referenced by a path instead of by its name. In other words:
```javascript
    import config from 'config' // Error
    const config = require('config') // Error
    import config from '../node_modules/config' // Success
    const config = require('../node_modules/config') // Success
```

4. Looking at the built package, I have noticed that the root of the problem is that, for some reason, the 'config' module does not get added to the '/programs/server/packages/modules.js' file. So the issue is in the build process.

5. The issue only occurs on meteor 1.9. On meteor 1.8.3 (and 1.8.2 for that matter), the 'config' module is correctly loaded.

All my tests were run on macOS 10.15.3.

To make it easier to reproduce the error, I have created a [Github project](https://github.com/claudiosdc/meteor-load-config-module-bug). Just follow the script below to reproduce it.

```shell
CSC-MacBookProRetina:GitHub claudio$ git clone https://github.com/claudiosdc/meteor-load-config-module-bug.git
Cloning into 'meteor-load-config-module-bug'...
remote: Enumerating objects: 23, done.
remote: Counting objects: 100% (23/23), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 23 (delta 0), reused 23 (delta 0), pack-reused 0
Unpacking objects: 100% (23/23), done.
CSC-MacBookProRetina:GitHub claudio$ cd meteor-load-config-module-bug/meteor_app
CSC-MacBookProRetina:meteor_app claudio$ meteor npm i
added 83 packages from 70 contributors and audited 620 packages in 1.366s
found 0 vulnerabilities

CSC-MacBookProRetina:meteor_app claudio$ meteor
[...]

PR #10896, opened 2020-02-03
Proposed fix for issue #10892

claudiosdc:
This is a proposed fix for issue #10892 that I have opened last week.

After further investigation, I found out that the issue was caused by the fact that the build process (or more specifically the 'resolver.ts' module) was incorrectly identifying the 'config' npm module as a Node.js' internal/native module.

The fix replaces the code used to list Node.js' native modules in the '/tools/isobuild/resolver.ts' module.

A test case named "build - link npm package named 'config' (#10892)" has been added to the '/tools/tests/bundle.js' module, and it is used to validate the fix.


tools/isobuild/resolver.ts
```diff
@@ -21,18 +21,25 @@ import {
...
+function getBuiltinModules() {
+  'use strict';
+  const {builtinModules} = require('module');
+
+  const blacklist = [
+    'sys'
+  ];
+
+  // eslint-disable-next-line node/no-deprecated-api
+  return (builtinModules || Object.keys((process as any).binding('natives')))
```
sebakerckhof: I think we can leave out the fallback to process.bind('natives') here, since we know the node version. It's deprecated anyway ( https://nodejs.org/dist/latest-v12.x/docs/api/deprecations.html#deprecations_dep0111_process_binding )

tools/isobuild/resolver.ts
```diff
@@ -21,18 +21,25 @@ import {
...
+      .filter((x: string) => !/^_|^(internal|v8|node-inspect)\/|\//.test(x) && !blacklist.includes(x))
+      .sort();
```
sebakerckhof: Not that it will really matter for an array this size, but there's no reason for our usage to do a sort here.

tools/isobuild/resolver.ts
```diff
@@ -21,18 +21,25 @@ import {
...
+  'use strict';
+  const {builtinModules} = require('module');
+
+  const blacklist = [
+    'sys'
+  ];
+
+  // eslint-disable-next-line node/no-deprecated-api
+  return (builtinModules || Object.keys((process as any).binding('natives')))
+      .filter((x: string) => !/^_|^(internal|v8|node-inspect)\/|\//.test(x) && !blacklist.includes(x))
```
sebakerckhof: If you leave out the process.bind fallback the regex test is no longer necessary. And actually the blacklist also doesn't really matter. The sys module exists, but gives a deprecation warning, since node 1.0.0. The NPM sys module is just the util module.

tools/isobuild/resolver.ts
```diff
@@ -21,18 +21,25 @@ import {
...
 } from "../fs/optimistic";
 
-const nativeModulesMap: Record<string, string> = Object.create(null);
-const nativeNames = Object.keys((process as any).binding("natives"));
-
-// Node 0.10 does not include process as a built-in module, but later
-// versions of Node do, and we provide a stub for it on the client.
-nativeNames.push("process");
+// Note: the source code for the following function was copied from the 'builtin-modules' npm package (ver. 3.1.0)
+function getBuiltinModules() {
```
sebakerckhof: According to my comments below, I don't see a use for this function anymore and we can just use `require('module').builtinModules`

tools/isobuild/resolver.ts
claudiosdc: OK. The code has been changed to reflect this simplification.

tools/isobuild/resolver.ts
claudiosdc: OK. The code has been changed to reflect this simplification.

tools/isobuild/resolver.ts
claudiosdc: OK. The code has been changed to reflect this simplification.

tools/isobuild/resolver.ts
claudiosdc: OK. The code has been changed to reflect this simplification.

PR #10899, opened 2020-02-04
Switch `google-oauth` to pass token in headers.

StorytellerCZ:
Update OAuth API calls for authentication due to deprecation:
https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/
https://developer.github.com/v3/auth/#basic-authentication

### DIFF
--- a/packages/github-oauth/github_server.js
+++ b/packages/github-oauth/github_server.js
@@ -62,8 +62,7 @@ const getIdentity = accessToken => {
   try {
     return HTTP.get(
       "https://api.github.com/user", {
-        headers: {"User-Agent": userAgent}, // http://developer.github.com/v3/#user-agent-required
-        params: {access_token: accessToken}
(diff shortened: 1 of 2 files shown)


abernix:
Thanks so much for opening this, @StorytellerCZ!

I've also bumped the version of `accounts-github` in fd0bd1ec59e66d2145d21a41d403c192a655128d to ensure it depends on the latest version once published.

Both `accounts-github` and `github-oauth` have now been published, so please do try this out and let me know if it's problematic at all.

To ensure maximum compatibility for those who may need to update (suddenly, because of the e-mails GitHub started sending out), I've published these with the same version of `meteor` which was used to publish the previous versions of each, which was Meteor 1.8.

```
Published github-oauth@1.2.3.
Published accounts-github@1.4.3.
```

PR #10901, opened 2020-02-05
Add enableCallback option for appcache

povesteam:
Sometimes `browsers` is not enough when deciding to enable/disable the appcache feature.

This PR adds `enableCallback` as a more versatile option to control the appcache feature on a per-request basis.

### Use case example 1

```js
// Enable offline mode using a value from database
Meteor.AppCache.config({
    enableCallback: () => {
        return getSettingsFromDb("public.appcache_enabled");
    },
});
```

### Use case example 2

```js
// Enable offline mode only when client sends a valid certificate
Meteor.AppCache.config({
    enableCallback: req => {
        const validation = validateClientCert({
            clientCertPayload: req.headers["x-client-cert"],
            url: req.url.href,
        });

        return validation.passed;
    },
});
```

### DIFF
--- a/packages/appcache/appcache-server.js
+++ b/packages/appcache/appcache-server.js
@@ -7,6 +7,7 @@ import path from 'path';
 
 let _disableSizeCheck = false;
 let disabledBrowsers = {};
+let enableCallback = null;
 
 Meteor.AppCache = {
   config: options => {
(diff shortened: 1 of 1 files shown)


sebakerckhof:
Could you also make a PR towards: https://github.com/meteor/docs/blob/master/source/packages/appcache.md 
thanks.

povesteam:
Added the option and example in the documentation: https://github.com/meteor/docs/pull/658
Thank you for the suggestion!

povesteam:
Why did Travis CI fail? It did not output anything meaningful in the logs.

filipenevola:
@povesteam not sure. I'm running it again! :)

povesteam:
👍 It passed now. Thanks!

filipenevola:
I've published 1.2.5

povesteam:
Great! Thank you!

PR #10649, opened 2019-07-23
[FIX] Support HTTPS_PROXY and NO_PROXY variables

magicbelette:
Closes #11984, #10282, #11484

PR #10914, opened 2020-02-13
[FIX] Support HTTPS_PROXY and NO_PROXY variables

cwouam:
Based on the magicbelette patch (https://github.com/meteor/meteor/pull/10649) with a minor correction.


filipenevola:
@magicbelette could you sign the CLA?

PR #10926, opened 2020-02-19
Avoid bailing out with module.useNode() for ESM modules.

benjamn:
On the server, Meteor attempts to avoid bundling `node_modules` code by replacing entry point modules with a stub that calls [`module.useNode()`](https://github.com/meteor/meteor/blob/fca6c62d432eff1afd4a67280b7c90af23399511/packages/modules-runtime/server.js#L48). This trick allows evaluating server `node_modules` natively in Node.js, faithfully preserving all Node-specific behaviors, such as `module.id` being an absolute file system path, the `__dirname` and `__filename` variables, the ability to import binary `.node` modules, and so on. This bailout has been in place since Meteor 1.3 (when the module system was first introduced), and it has worked remarkably well.

However, starting in Node.js 12.16.0 (Meteor 1.9.1+), modules evaluated natively by Node are considered ECMAScript modules (ESM) if the closest `package.json` file has `"type": "module"` (or has an `.mjs` file extension). This poses a problem for the `module.useNode()` trick, because ESM modules cannot be imported synchronously using `require` (which is currently how `module.useNode()` works).

To work around this new error, this commit checks `package.json` for `"type": "module"` in `ImportScanner#shouldUseNode` to determine whether it's safe to use the `module.useNode()` trick. This leads to including ESM modules in the server bundle, rather than relying on Node to execute them directly from the file system.

The good news is that ESM modules don't have access to nearly as many Node/CommonJS-specific quirks: no `module`, `require`, or `exports` variables; no `__dirname`, no `__filename`; no ability to import JSON or other non-ESM file types (at least right now). So it seems somewhat less important for ESM code (compared to CommonJS code) to bail out into native Node.js execution using `module.useNode()`. In other words, bundling server code should not affect its execution in nearly as many cases, if that code is ESM rather than legacy CommonJS.

If this good news turns out to be overly optimistic, we can consider using a different kind of bailout stub that's capable of importing ESM, perhaps using dynamic `import()`. For now, making sure we avoid bailing out for ESM code found in `@babel/runtime/helpers/esm/*` is the priority.

### DIFF
--- a/packages/modules-runtime/server.js
+++ b/packages/modules-runtime/server.js
@@ -71,6 +71,10 @@ Module.prototype.useNode = function () {
     return false;
   }
 
+  // See tools/static-assets/server/npm-require.js for the implementation
+  // of npmRequire. Note that this strategy fails when importing ESM
+  // modules (typically, a .js file in a package with "type": "module" in
+  // its package.json), as of Node 12.16.0 (Meteor 1.9.1).
   this.exports = npmRequire(this.id);
 
   return true;
--- a/tools/fs/optimistic.ts
+++ b/tools/fs/optimistic.ts
@@ -354,28 +354,46 @@ export const optimisticReadMeteorIgnore = wrap((dir: string) => {
 
 type LookupPkgJsonType = OptimisticWrapperFunction<
   [string, string],
-  ReturnType<typeof optimisticReadJsonOrNull>
+  ReturnType<typeof optimisticReadJsonOrNull>[]
 >;
 
-export const optimisticLookupPackageJson: LookupPkgJsonType =
+// Returns an array of package.json objects encountered in any directory
+// between relDir and absRootDir. If a package.json with a "name" property
+// is found, it will always be the first object in the array.
+export const optimisticLookupPackageJsonArray: LookupPkgJsonType =
 wrap((absRootDir: string, relDir: string) => {
   const absPkgJsonPath = pathJoin(absRootDir, relDir, "package.json");
   const pkgJson = optimisticReadJsonOrNull(absPkgJsonPath);
+
+  // Named package.json files always terminate the search. This
+  // restriction was first introduced to fix #10547, before this function
+  // was updated to return an array instead of a single object, but
+  // returning here is still an important base case.
   if (pkgJson && typeof pkgJson.name === "string") {
-    return pkgJson;
+    return [pkgJson];
   }
 
   const relParentDir = pathDirname(relDir);
   if (relParentDir === relDir) {
-    return null;
+    return [];
   }
 
   // Stop searching if an ancestor node_modules directory is encountered.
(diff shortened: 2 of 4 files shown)


benjamn:
Here are some examples of the new error reported by Node.js 12.16.0:
```
Must use import to load ES Module: /Users/ben/meteor/tools/tests/apps/modules/node_modules/@babel/runtime/helpers/esm/asyncIterator.js
require() of ES modules is not supported.
require() of /Users/ben/meteor/tools/tests/apps/modules/node_modules/@babel/runtime/helpers/esm/asyncIterator.js from /private/var/folders/zb/l6qn64652ljcfb99jlv2s1v80000gn/T/meteor-test-run1rdgzk2.zhmw/.meteor/local/build/programs/server/npm-require.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename asyncIterator.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/ben/meteor/tools/tests/apps/modules/node_modules/@babel/runtime/helpers/esm/package.json.

Error [ERR_REQUIRE_ESM] [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/ben/meteor/tools/tests/apps/modules/node_modules/@babel/runtime/helpers/esm/asyncIterator.js
require() of ES modules is not supported.
require() of /Users/ben/meteor/tools/tests/apps/modules/node_modules/@babel/runtime/helpers/esm/asyncIterator.js from npm-require.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename asyncIterator.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/ben/meteor/tools/tests/apps/modules/node_modules/@babel/runtime/helpers/esm/package.json.
```
None of those suggestions are useful, because we do not control the package in question.

s7dhansh:
@filipenevola  I got the mentioned error while doing
```
 import {Parallax, ParallaxLayer} from 'react-spring/renderprops-addons';
```

Is it a problem with meteor?

filipenevola:
Hi, it's working fine for me, I tested using `meteor create --react` code, see

```javascript
import React from 'react';
import {Parallax, ParallaxLayer} from 'react-spring/renderprops-addons';
import Hello from './Hello.jsx';
import Info from './quave/Info.jsx';

const App = () => (
  <div>
    <h1>Welcome to Meteor!</h1>
    <Hello/>
    <Info/>
    <Parallax pages={3} scrolling={false} horizontal ref={ref => (this.parallax = ref)}>
      <ParallaxLayer offset={0} speed={0.5}>
        <span onClick={() => this.parallax.scrollTo(1)}>Layers can contain anything</span>
      </ParallaxLayer>
    </Parallax>
  </div>
);

export default App;

```

PR #10929, opened 2020-02-20
Xmlbuilder2 update

BrianMulhall:
Branched off of the release-1.9 branch cause that was the most recent release at the time.

Reason for doing this:
The xmlbuilder library will no longer be the focus of future development, instead the author will be investing their time in xmlbuilder2. So i thought updating this dependency might make sense. 
https://github.com/oozcitak/xmlbuilder-js

How did i did I verify the work:
I added in some test cases for the package so if in the future we update the xmlbuilder2 library's version and the semantic's of the functions we use have changed relative to our expected results we would know from the test cases. 

I also added in a test case that would generate a portion of the config file so i could compare the xmlbuilder's output with the xmlbuilder2's output. That way i would know the update to xmlbuidler2 wouldn't alter the outputted config.xml file. I left that test case in there even tho its probably overkill i already made it so why not just leave it in case its useful in the future.  I also used my branch with this update to create a config.xml from a meteor app i made and compared it using a diff tool to the config.xml generated using the 1.9 release. They were identical. The mobile-config.js file I used exercised all the functions documented here https://docs.meteor.com/api/mobile-config.html except the icons and spalshscreens which have a set of defaults which get added to the config.xml. So i thought that was a good test case. All other tests and the circleci test groups all passed.

Package versioning:
I don't know how to properly update the version of atmosphere xmlbuilder package tho. It started as 2.5.15 and there was a note in the source to sync the node package's version with the atmosphere package's version. But in this case the version would go to 1.3.0 (the version that xmlbuilder2 is at now) which isn't what was intended. So i thought 3.0.0 would make sense cause its major update and the api is changing. But when i did that the constraint solver would raise and error. So i upped it to 2.5.16 which doesn't seem right to me.

Questions/Concerns:
Any issues or questions let me know. I tried to follow the readme's closely but this is my first pull request to Meteor so let me know if you need me to do things a bit differently.


packages/xmlbuilder/package.js
```diff
@@ -1,16 +1,20 @@
 Package.describe({
   summary: "An XML builder for node.js similar to java-xmlbuilder.",
-  version: '2.5.15'
+  version: '2.5.16'
```
sebakerckhof: This would need a major version bump, since the API is not compatible

tools/cordova/builder.js
```diff
@@ -248,7 +248,7 @@ export class CordovaBuilder {
   writeConfigXmlAndCopyResources(shouldCopyResources = true) {
     const { XmlBuilder } = loadIsopackage('xmlbuilder');
```
sebakerckhof: Since this is basically the only user of the xmlbuilder package in Meteor I wonder why we don't load the npm package directly. @benjamn @filipenevola : Can you shine some light on this?

packages/xmlbuilder/package.js
BrianMulhall: ok, if this package ends up being kept rather than a using the xmlbuilder2 node package directly ill make it 3.0.0

tools/cordova/builder.js
BrianMulhall: I think that makes alot of sense, I didnt want to deviate too far from the way it was I suppose. But that was the only file making use of the xmlbuilder package. I could update this branch and remove the xml builder package if you guys decide that's a better approach

BrianMulhall: I made a branch where the xmlbuilder isopack (i can see why this name was needed now haha) has been removed, and then in the builder.js file the xmlbuilder2 node package is imported directly. As a work around i installed the xmlbuilder2 library using npm in the tools directory's node_modules folder so I could do some testing. But I dont know how to add a new node module to the what the startup script calls the "dependency kit".

sebakerckhof: The node_modules of the dependency kit are defined in: `./scripts/dev-bundle-tool-package.js`

If you change these, you should also change the dev bundle version in `./meteor` .

The consequence is that the tests here will fail, since the bundle first needs to be published.

However, there might still be a reason I'm not aware of this xmlbuilder was loaded from an isopackage instead of NPM. So maybe it's better to wait for feedback from @benjamn if this is the right approach.

BrianMulhall: thats sounds good to me, whatever direction you guys decide is best. I was preparing for both scenarios and was curious how to make updates to the tools area of the project as well. Thanks for letting me know how to update the dependency kit too, Ill play around with this branch locally in the mean time :)

benjamn: I am 100% in favor of adding `xmlbuilder2` to the dev bundle, rather than loading it into the command-line tools codebase from a Meteor package. The implementation of that package is just one line, after all:
```
XmlBuilder = Npm.require('xmlbuilder');
```
Maybe we can also move `meteor/packages/xmlbuilder` into `meteor/packages/non-core` while we're at it, since that package is definitely not used by any core packages (including `meteor-tool`) now?

BrianMulhall: I moved the original version of the xmlbuilder isopack to the non-core folder, and then added the xmlbuilder2 node package to the dev bundle. The license file now has both libraries included too. I tested this branch by creating a config.xml and comparing it against one created using the 1.9 release and they are identical. The only thing I am not sure of it the dev bundle version I chose. The most recent one I noticed in the repo was 12.16.1.3 so i bumped it to 12.17.1.3, but the xmlbuider api is changing so 13.0.0.0 might make sense too

sebakerckhof: I should have mentioned that the dev bundle version doesn't follow semver. Basically we keep the first numbers for the node version (so that should stay 12.16.1) and then just increase the last number for other changes that need a new dev bundle, so `12.16.1.4` is ok.

filipenevola:
hi @BrianMulhall please update your branch with the latest `devel` and then bump the patch number of bundle version as recommended by Seba, so I can generate a new dev bundle for you and the tests can run again. Ok?

Thank you for your work 😉

tools/cordova/builder.js
BrianMulhall: gotcha .. let me make that change :)

BrianMulhall:
I am not super familiar with using git submodules so I wanted to double check this with you guys. Is there a change that to the blaze repo that I should also merge in to my branch? This is what I am seeing after I checked out the devel branch locally
<submodule updated 5c0beecc7b696dc4856019848e0c1544038d8190 → e6b8a12344e6c7438f2a06bb98a5796663905150>

benjamn:
@BrianMulhall Just run
```
```
and you should get the right `packages/non-core/blaze` code. Shouldn't be related to these changes in any way.

benjamn:
@BrianMulhall On second thought, I'm happy to jump in and wrap up these last few things, if you like. You've done all the interesting work already, and what's left is mostly just process stuff.

BrianMulhall:
yea if you think that would be best, I would hate to make work for you guys by checking in a mistake. thank you @benjamn

BrianMulhall:
happy mardi gras everyone :)

PR #10932, opened 2020-02-21
add Decimal support to minimongo

boomfly:
Minimongo now has no support for Decimal.js so there is issue when using is.
If we try to use meteor/mongo-decimal package it works normal on server side, but not on client, especially sorting cause it not implementing there.

### DIFF
--- a/packages/minimongo/matcher.js
+++ b/packages/minimongo/matcher.js
@@ -1,10 +1,28 @@
-import { Decimal } from 'meteor/mongo-decimal'
 import LocalCollection from './local_collection.js';
 import {
   compileDocumentSelector,
   hasOwn,
   nothingMatcher,
 } from './common.js';
+
+let Decimal = Package['mongo-decimal']?.Decimal;
+
+// Due to {weak: true} Decimal dependency on client
+// declare minimal Decimal functionality
+if (!Decimal) {
+  class Decimal {
+    constructor(value) {
+      this.value = value.toString();
+    }
+    minus(value) {
+      let _value = Number(this.value) - Number(value);
+      return new Decimal(_value);
+    }
+    toNumber() {
+      return Number(this.value);
+    }
+  }
+}
 
 // The minimongo selector compiler!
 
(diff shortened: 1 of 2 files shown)


packages/minimongo/matcher.js
```diff
@@ -1,3 +1,4 @@
+import { Decimal } from 'meteor/mongo-decimal'
```
sebakerckhof: Unfortunately, you can't use ES6 imports for weak-linked packages. If you wouldn't have the mongo-decimal package installed, this will throw since it can't find the imported package.

The way to work with weak-linked packages is by using the Meteor's package system:
`const Decimal = Package['mongo-decimal']?.Decimal`

But then the instanceOf will throw since you would be checking against undefined.

So either you will have to check first that Decimal isn't undefined, or, since it's only for instanceof checks, you could do something like:
`const Decimal = Package['mongo-decimal']?.Decimal || class DecimalStub {}`

boomfly: Ok! I was fixed it like this

```
let Decimal = Package['mongo-decimal']?.Decimal;

// Due to {weak: true} Decimal dependency on client
// declare minimal Decimal functionality
if (!Decimal) {
  class Decimal {
    constructor(value) {
      this.value = value.toString();
    }
    minus(value) {
      let _value = Number(this.value) - Number(value);
      return new Decimal(_value);
    }
    toNumber() {
      return Number(this.value);
    }
  }
}
```

packages/minimongo/matcher.js
```diff
@@ -5,6 +5,25 @@ import {
   nothingMatcher,
 } from './common.js';
 
+let Decimal = Package['mongo-decimal']?.Decimal;
+
+// Due to {weak: true} Decimal dependency on client
+// declare minimal Decimal functionality
+if (!Decimal) {
```
sebakerckhof: This has no value, since the instanceof checks will never be true for this class.

sebakerckhof: For clarity: The reason I proposed a DecimalStub{} class was just to have some kind of object that doesn't throw an error when we do instanceof, but which would never be true. If someone wants to have decimal support, they should add the mongo-decimal package.

boomfly: Thanks for help! I was change code with your recommendations.

PR #10933, opened 2020-02-21
Drops support for Win 32-bit

filipenevola:
Because MongoDB since 3.4 no longer supports 32-bit Windows, Meteor 1.10 has also dropped support for 32-bit Windows. In other words, Meteor 1.10 supports 64-bit Mac, Windows 64-bit, and Linux 64-bit.

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,7 +5,15 @@
 * Cordova has been updated from version 7 to 9. We recommend that you test
   your features that are taking advantage of Cordova plugins to be sure
   they are still working as expected.
+  
+* Because MongoDB since 3.4 no longer supports 32-bit Windows, Meteor 1.10 has also
+  dropped support for 32-bit Windows. In other words, Meteor 1.10 supports
+  64-bit Mac, Windows 64-bit, and Linux 64-bit.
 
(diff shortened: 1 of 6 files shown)


filipenevola:
@benjamn I'm merging to 1.10 and then you can review there. I've created this PR to have a record here.

SimonSimCity:
@filipenevola If this confirmed to be included in `v1.10`, I suggest to also mention this in https://github.com/meteor/meteor-feature-requests/issues/152.

filipenevola:
Hi @SimonSimCity I've followed this commit to drop Win 32-bit as we did in the past for Linux 32-bit https://github.com/meteor/meteor/commit/db4c94c81f5d5ee29e5e66282b6fe7dfbde40879

Do you see any part missing? I believe a few links in the mentioned FR are old and are not used anymore.

SimonSimCity:
I haven't been deep into this. I just linked the FR here because another FR I was reading up on required mongodb to be upgraded to 3.6 on any platform, which was blocked by the 32 bit version still being supported.

Looks good on first glance 👍

PR #10936, opened 2020-02-22
Update JSHint version from 2.7.0 to 2.11.0

BrianMulhall:
Branched off of the release-1.9 branch

Reason for doing this:
I wanted to try out JSHint after using ESLint for a while and wanted to use the latest version.

How did i did I verify the work:
I checked that the four values we use for the default configuration are all still recognized by JSHint. Meteor uses 'node', 'browser', 'unused' and 'undef', which are all still valid. The link to verify that is below.
https://jshint.com/docs/options/

I verified that the API of the library is also unchanged and I made a small driver application that would test this out a bit so I could see it in action. This can be verified by following the link below. 
 https://jshint.com/docs/api/

Package Versioning:
the version was set to 1,1.7 so I figured this was a minor version bump at most or a patch bump at least so I made it 1.1.8.  But 1.2.0 might be a better choice


BrianMulhall:
I updated the syntax in this package as well as the version update (i figured I might as well while I am submitting this pull request),  to use the newer const/let instead of var, and the newer class syntax. I also tried to improve the documentation a bit so its easier for people to see how the default config will effect the linting output.

filipenevola:
@BrianMulhall one test is stuck, can you create a new branch?

filipenevola:
I'm closing this issue because it's too old.

If you think this issue is still relevant please open a new one.

Why? We need to pay attention to all the open items so we should keep opened only items where people are involved.

harryadel:
Sorry for the noob question, but is there a use for this package? A little look into its [atmosphere](https://atmospherejs.com/meteor/jshint) repository for usage metrics shows very little to no usage with 577 app installs in the entire lifetime of the package! 
atmosphere atmosphere2

Besides, I've never heard of anyone using jshint nowadays when you have eslint or prettier.

Should this be deprecated?

filipenevola:
Yes, I agree @harryadel , it could be deprecated.

StorytellerCZ:
I'll deprecate it for for 2.3 and move it to the deprecated folder after 2.3 is published.

StorytellerCZ:
Never mind, I can do it in the 2.3 branch right now.

PR #10942, opened 2020-02-25
Refactor/modernize random

sebakerckhof:
New PR as requested by @filipenevola . Previous discussion: https://github.com/meteor/meteor/pull/9603/files


harryadelb:
Shouldn't the NPM version made by @aldeed be used as part of the ongoing effort to move out of Atmosphere?

sebakerckhof:
The current efforts, within the Meteor codebase have been to stop wrapping external libraries in Meteor packages. For internal libraries, we have not really moved anything to npm. However, this modernization does make it less dependent on the Meteor package system, so it does facilitate a move to NPM if this would be required later on.

Issue #10904, opened 2020-02-08
__dirname is not defined - markdown


marxo:
Hi,

I'm running into a problem on a fresh new app in Meteor 1.9. When I add the `markdown` package to the app, the client-side fails with the message `ReferenceError: __dirname is not defined`. No other new packages have been added. This error shows up only when `markdown` package is added. I'm using Ubuntu 18.04 LTS.

So, even this will fail with the `markdown` package added:

```
{{#markdown}} # test {{/markdown}}
```

jordancortes:
This is a deprecated package #9737

BrianMulhall:
I still fixed it because the only popular alternative markdown package https://atmospherejs.com/perak/markdown i found on atmosphere seemed to be abandoned. So the fix should get merged in and released in the first post 1.10 release.

filipenevola:
I've published markdown@1.0.13.

PR #10946, opened 2020-02-27
Markdown Bug Fix [Issue #10904]

BrianMulhall:
I branched off of the latest devel branch commit

This issue was reported in #10904 and I was able to reproduce it. 

Testing this change allowed the page to load and the test snippet of markdown used in the issue to be shown in the screen

### DIFF
--- a/packages/deprecated/markdown/package.js
+++ b/packages/deprecated/markdown/package.js
@@ -2,7 +2,7 @@
 
 Package.describe({
   summary: "Markdown-to-HTML processor",
-  version: "1.0.12"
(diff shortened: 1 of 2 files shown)


filipenevola:
@BrianMulhall I've published markdown@1.0.13.

BrianMulhall:
@filipenevola  i was thinking since we dont really have an alternative community maintained markdown library (at least not one that is actively being maintained) that if issues were reported for this deprecated package I dont mind fixing them. So thanks for publishing this update :)

Issue #10964, opened 2020-03-12
Node options required to be together blocked to be used 


linegel:
Before reporting a bug, please check for existing or closed issues
first and read the instructions for filing a bug report:
https://github.com/meteor/meteor/blob/devel/CONTRIBUTING.md#reporting-a-bug-in-meteor

Bug report:
Meteor version - 1.9.2
OS - MacOS

The expected behavior:
Pass `NODE_OPTIONS="--experimental-modules --input-type=module --es-module-specifier-resolution=node"` with meteor build and have flags used by nodejs during building process

The actual behavior.
Following error:
`/Users/User/.meteor/packages/meteor-tool/.1.9.2.1bpgp1k.xybn++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/dev_bundle/bin/node: bad option: cannot use --es-module-specifier-resolution and --experimental-specifier-resolution at the same time
`

Simple reproduction:
Try to make a Meteor build with following node options:
` NODE_OPTIONS="--experimental-modules --input-type=module --es-module-specifier-resolution=node" NODE_ENV=production APP_ENV=dev meteor build ../app --directory --server-only --architecture os.linux.x86_64
`

Note №1: 
This works as expected, no errors thrown
<img width="1228" alt="Screen Shot 2020-03-12 at 04 29 34" src="https://user-images.githubusercontent.com/1746296/76478616-4fdc8200-641a-11ea-95b3-221efcbbb909.png">

With clean node, removal of --experimental-modules leads to error, coz it **required** for 2 other options 
Error which is thrown in case of --experimental-modules removal by node:
```
node: --input-type requires --experimental-modules to be enabled
node: --es-module-specifier-resolution requires --experimental-modules be enabled
```

Removal of  --es-module-specifier-resolution=node isn't a case since it's required for expected behavior during modules importing. 

Note №2:
Removal of --experimental-modules shows an error, that it is required to use both -input-type and --es-module-specifier-resolution (which previously was blocked by --experimental-modules) which makes situation feels like a circular block.
So, the actual question is why the usage of both
`--experimental-modules` and `--es-module-specifier-resolution=node` together throws error? 
<img width="975" alt="Screen Shot 2020-03-12 at 04 51 53" src="https://user-images.githubusercontent.com/1746296/76479556-3557d800-641d-11ea-806f-b2f442f7caa0.png">

Anybody had any similar issue or have ideas about why it's happening?

[...]

PR #10965, opened 2020-03-12
Add separate SERVER_NODE_FLAGS envar

sebakerckhof:
Since Meteor 0.5.3, Meteor allowed to pass node command line flags to the server node process via the `NODE_OPTIONS` environment variable.

However, as I found out by triaging https://github.com/meteor/meteor/issues/10964 , since Node version 8 / Meteor 1.6 this has become a default node envar with the same behavior:
https://nodejs.org/api/cli.html#cli_node_options_options

The side effect is that this now also affects Meteor tool.

The command line parameters could already be set separately via the `TOOL_NODE_FLAGS` envar.
This makes it also possible (again) for the server.


sebakerckhof:
Well, it is fully backwards compatible, since node flags keeps working. The breaking change happened in meteor 1.6, when it started to get applied to meteor tool as well dhe to tge chanfes in node

StorytellerCZ:
@sebakerckhof Maybe it happened then, but if it kept working for some now renaming the var is breaking now.

sebakerckhof:
No, since the old var keeps working. People can keep using the old var, and it will keep on working exactly as in meteor 1.6 to 1.10

StorytellerCZ:
:man_facepalming: Of course. Sorry. Well, anyway fixing the merge conflict is I think the only remaining issue here.

PR #10341, opened 2018-11-16
Source maps in standard-minifier-js

morloy:
This is directly pulled from @zodern’s [zodern/minify-js-sourcemaps](https://github.com/zodern/minify-js-sourcemaps) and would finally solve meteor/meteor-feature-requests/issues/30

Still TBD is hiding sourcemaps in production, but there’s also `zodern:hide-production-sourcemaps` which could be integrated into this PR.

PR #10967, opened 2020-03-17
Minifier-JS Package update

BrianMulhall:
I added in test cases for this package, I exercise the error handling (when the input is just invalid and terser and babel-minify cant do anything with it), the case where terser can't handle syntax that babel-minify can, some general minification tests and then "setting" specific cases that will catch if you update the configuration in a way that will alter the output as a way to alert you to that fact. 

Terser was updated from 4.4.0 to 4.6.6 after the tests were run on 4.4.0 and all passed.

I also updated the package to use const/ let and not var and to use the newer class syntax. This was an effort to modernize the package a bit.

I read through the changelog here https://github.com/terser/terser/blob/master/CHANGELOG.md and everything from 4.4.0 to 4.6.6 seems ok to rely on. But it could use a once over by people who know more than me.


sebakerckhof:
In general this looks good.
Just regarding the tests: I can see how one test to see if it gets minified is useful, and also the one to test if the fallback system works. But for the others I would be aware this doesn't increase our maintenance burden, since the output format of terser can easily change between versions and it's also not up to us to test the external library.

BrianMulhall:
i was really just making the extra cases to make sure from version to version if something did change you would be aware of it and then could take action if you wanted to. And to see how terser behaves when settings are altered in an easy way. That was the thinking anyway. It could be pared down to lighten the load tho

BrianMulhall:
I will refactor the tests to have the following three cases

1.) the terser success case (where terser minifies the javascript given)
2.) the terser fails but babel-minify success case
3.) the terser fails and then babel-minify fails case

that way we cover all the possibilities and we know our expectation is being met, I can push that change later tonight

BrianMulhall:
ok that last commit has just those three test cases I mentioned above.

Do we document anywhere that we can use conditional compilation in Meteor code? I never knew that

For example the code below, will emit only one branch into what terser(or babel-minfy) emits based on the NODE_ENV environment variable.

if(process.env.NODE_ENV == "development"){
  console.log('dev mode');
}
else{
  console.log('produciton mode');
}

packages/standard-minifier-js/plugin/utils.js
```diff
@@ -1,26 +0,0 @@
-"use strict";
```
BrianMulhall: this file wasn't being used,so i removed it

benjamn:
@BrianMulhall Now that you're an expert on this area of the code—what do you think about the feasibility of removing the `babel-minify` fallback? We added that back when `uglify-js` did not have full support for ES2015+ syntax (#8519), even though `babel-minify` was much slower and used a ton of memory. Would it be safe to assume that `terser` (an active fork of `uglify-js`) has implemented the necessary syntax?

BrianMulhall:
@benjamn in testing i found that babel-minify supported numeric separators (which is Stage 3 not stage 4 ...  https://github.com/tc39/proposal-numeric-separator ) and terser did not and there may be other cases where the features that babel-minify will support includes things that terser doesn't. I can keep digging into this tho and get back to you.

BrianMulhall:
I guess the only unknown is which features does babel-minify support that terser doesnt, I only know of that one (i.e. numeric separators) but I could dig around and see if there are any more. or if terser documents what they support explicitly and we could know with certainty what it will handle, I will look for that as well

BrianMulhall:
@benjamn @sebakerckhof so i created an issue in the terser repo about numeric separators not being support and then asked how well supported ES2015 is. https://github.com/terser/terser/issues/632 and the response given was that ES2015 is fully supported (except one feature that babel doesnt support either so nobody misses it). so updating the code to use only terser would be a good idea in my opinion, want me to move forward with that change?

BrianMulhall:
I felt like after talking with @fabiosantoscode about the level of support terser provides for ES2015 and even newer versions of the spec that relying on terser alone would be a good decision. So i pushed up the changes I made for everyone to review. If we decide this isnt a good idea let me know after you take a look and i will revert the changes

packages/standard-minifier-js/plugin/minify-js.js
```diff
@@ -1,161 +1,105 @@
...
-            (lineSrcLineNumber ? ", line " + lineSrcLineNumber : "") + "\n" +
-            "\n" +
-            lineErrorMessage + ":\n" +
-            "\n" +
-            lineSrcLineContent + "\n"
-          );
-
-          throw minError;
-        }
+    // this function tries its best to locate the original source file
```
BrianMulhall, line 27: this function needed to be changed to account for the error.message field being different between babel-minify and terser, but I made a little driver program to test this function inside of and used some bundled files from a meteor build that i wrote to disk as test data and it worked well during testing

packages/minifier-js/minifier-tests.js
```diff
@@ -0,0 +1,15 @@
...
+  let terserResult = meteorJsMinify('function add(first,second){return first + second; }\n');
+  test.equal(terserResult.code, 'function add(n,d){return n+d}');  
+});
+
+// this feature has been reqested in this issue https://github.com/terser/terser/issues/632
+// so when we bump up the terser version in the future and this test fails we will know when
+// its been done and can remove this test :)
+Tinytest.add('minifier-js - unsupported feature test (numeric seperators)', (test) => {
+  test.throws(() => meteorJsMinify('let number = 1_000_000_000_000;\n')  );   
+});
```
fabiosantoscode, line 11: I hope I break this test soon!

packages/standard-minifier-js/plugin/minify-js.js
```diff
@@ -1,161 +1,105 @@
...
+    const toBeAdded = {
+      data: "",
+      stats: Object.create(null)
+    };
+
+    files.forEach(file => {
+      // Don't reminify *.min.js.
+      if (/\.min\.js$/.test(file.getPathInBundle())) {
+        toBeAdded.data += file.getContentsAsString();
+      }
```
fabiosantoscode, line 70: Yes. Besides being pointless, for some reason Terser is very slow when minifying already-minified files.

packages/minifier-js/minifier.js
```diff
@@ -1,45 +1,32 @@
...
-  } catch (e) {
-    // Although Babel.minify can handle a wider variety of ECMAScript
-    // 2015+ syntax, it is substantially slower than UglifyJS/terser, so
-    // we use it only as a fallback.
-    var options = Babel.getMinifierOptions({
-      inlineNodeEnv: NODE_ENV
-    });
-    result.code = Babel.minify(source, options).code;
-    result.minifier = 'babel-minify';
-  }
```
fabiosantoscode, line 42: But if you want to minify stage 3 features, or something I didn't get to yet, you're stuck!

BrianMulhall: @fabiosantoscode thanks for reviewing this change :) @benjamn If we are comfortable with all stage 4 features up to ES2019 being supported by terser with the one exception being  mentioned in this thread (which babel doesnt support either) https://github.com/terser/terser/issues/632 ... I think we are in good shape

fabiosantoscode: Yeaaah, it's kind of bothering me so this year, after ES2020 is implemented I'll get to that mysterious feature :)

fabiosantoscode:
Some notes you might want to know:

 - Terser is memory-hungry. You may find that you run out of ram when trying to minify more than 5mb of JS at once. Good bundlers will use it on each module individually, not in the result as a whole.
 - If you want to do more post-processing on the JS output, you may use the ESTree output option, such that you don't have to parse again in the next tool. Likewise, the input can be ESTree too. There are options for these.
 - If you're going to post-process the source map, the `sourceMap` output option can be a `{ asObject: true }` object, which prevents Terser from `JSON.stringify()`ing the source map. This means you don't have to parse it again after output.
 - If you have a non-string input source map you'd like to avoid stringifying just so Terser can parse it again, that's not implemented but I can implement it.

Best of luck!

BrianMulhall:
thank you for all the help @fabiosantoscode

BrianMulhall:
@sebakerckhof  @benjamn  https://github.com/meteor/meteor/pull/10341 want me to handle merging this work in to this branch and adding test cases for it? I just wanted to bring it to your attention that merge conflicts exist between these two branches

StorytellerCZ:
@BrianMulhall Updated the branch and terser. Let's get this moving again. As for #10341let's leave this for that PR as I'm planning to look into it after this one.

packages/minifier-js/package.js
```diff
@@ -1,14 +1,22 @@
...
+  terser: "4.8.0"
 });
 
 Package.onUse(function (api) {
+  api.use('ecmascript');
   api.use('babel-compiler');
-  api.export(['meteorJsMinify']);
-  api.addFiles(['minifier.js'], 'server');
+  api.mainModule('minifier.js', 'server');
+  api.export('meteorJsMinify');
```
StorytellerCZ, line 14: Shouldn't this be a server export?

packages/minifier-js/package.js
```diff
@@ -1,14 +1,22 @@
 Package.describe({
   summary: "JavaScript minifier",
-  version: "2.6.0"
+  version: "2.6.1"
```
StorytellerCZ, line 3: This will have to be `2.7.0`, but we'll deal with that later.

packages/minifier-js/minifier-tests.js
StorytellerCZ: Great to see tests here!

StorytellerCZ (changes requested):

StorytellerCZ:
OK, I can confirm that the tests are passing, just for the CircleCI not showing here, but on the user branch.

StorytellerCZ (approved):
Did some manual testing and had no problems. Good to go on my side.

packages/standard-minifier-js/plugin/minify-js.js
```diff
@@ -1,161 +1,105 @@
...
-        data: file.getContentsAsBuffer(),
-        sourceMap: file.getSourceMap(),
-        path: file.getPathInBundle(),
+    extensions: ['js'],
+    archMatching: 'web',
+  }, 
+  () => new MeteorBabelMinifier()
+);
+
+class MeteorBabelMinifier {
```
zodern, line 10: This class could be renamed since it no longer uses Babel.

zodern (changes requested):
I love how much simpler it is from removing the babel miniifier

StorytellerCZ:
@BrianMulhall could you please make the changes that @zodern requested so that we can include this in 2.4?

packages/standard-minifier-js/plugin/minify-js.js
StorytellerCZ: Fixed this.

packages/minifier-js/package.js
```diff
@@ -8,7 +8,15 @@ Npm.depends({
 });
 
 Package.onUse(function (api) {
+  api.use('ecmascript');
   api.use('babel-compiler');
```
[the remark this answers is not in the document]
StorytellerCZ: Removed

packages/standard-minifier-js/README.md
```diff
@@ -5,7 +5,27 @@
...
+The options that are set that differ from the default settings used by terser are the following:
+
+```
+drop_debugger: false
+unused:        false 
+safari10:       true
+```
+
+It should also be noted that by default terser will make only one pass while compressing the source code, 
+but additional passes could be configured to increase compression of the output during a production build.
```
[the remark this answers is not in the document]
StorytellerCZ: Removed

packages/minifier-js/minifier.js
```diff
@@ -1,45 +1,34 @@
-var terser;
+import terser from "terser";
```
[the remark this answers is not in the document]
StorytellerCZ: Reverted

zodern (approved):

StorytellerCZ:
I will close and re-open this PR to trigger the tests run.

StorytellerCZ:
@BrianMulhall please see the failing group 2 of the Circle CI tests.

StorytellerCZ:
@filipenevola CircleCI tests have passed: https://app.circleci.com/pipelines/github/BrianMulhall/meteor/132/workflows/1b93072d-099d-4166-8e15-4758e5f3065d

filipenevola (approved):

packages/minifier-js/package.js
zodern, line 12: Shouldn't this be a dependency of `standard-minifier-js` instead?

StorytellerCZ: We could potentially get it in through imply. But for now I would leave it and rather have a change like this in a separate PR. This one is big enough and has already gone on for long enough.

StorytellerCZ: @zodern that is unless you find it absolutely necessary to address right now.

zodern: Does `minifier-js` use `babel-compiler`? If not, it would be a two line change: (re-adding [this line](https://github.com/meteor/meteor/pull/10967/commits/76c21f872906e87dd5e276a3ca53bdcb05a25031#diff-9353693b93ed9cf0b487fabfc08ea75916030f1bd7afe25373192f488e31550bL12)) and removing `babel-compiler` from minifer-js.

Fixing in a separate PR is fine.

PR #10969, opened 2020-03-17
Fix caching compiler file cache

zodern:
Starting in Meteor 1.8.2, Plugin.fs methods no longer accept a callback and are always the sync version. The caching compiler passed a callback that was never called and caused the cache files to never be renamed to their correct name.

### DIFF
--- a/History.md
+++ b/History.md
@@ -250,7 +250,7 @@ N/A
   See [this comment](https://github.com/meteor/meteor/pull/10522#issuecomment-535535056)
   by [@SimonSimCity](https://github.com/SimonSimCity).
 
-* Plugin.fs methods are now always sync and no longer accept a callback.
+* `Plugin.fs` methods are now always sync and no longer accept a callback.
 
 ### Migration Steps
 
--- a/History.md
+++ b/History.md
@@ -250,6 +250,8 @@ N/A
   See [this comment](https://github.com/meteor/meteor/pull/10522#issuecomment-535535056)
   by [@SimonSimCity](https://github.com/SimonSimCity).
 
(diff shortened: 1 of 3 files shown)


sebakerckhof:
And don't we want it to be async again? (seeing how it used to say: "But we also don't want to block...")

zodern:
I am not sure how to handle making it async in a way that allows compiler plugins to be compatible with versions before Meteor 1.8.3 through the version that makes them async again. As is, the only way to use Plugin.fs methods in a way that is compatible is the sync versions, or use async/await since it is also able to handle functions that don't return a promise.

The Meteor tool has some places that would benefit from async fs functions, and I was planning to eventually add methods that ended in `Async`, for example `writeFileAsync`. Though for compiler plugins to use them, they would either not be compatible with older versions of Meteor, or have to also implement a sync version for older versions of Meteor.

I decided to open the PR as is since using the sync functions for now is better than a non-working file cache.

sebakerckhof:
Yeah, it's a shame we now have these versions where writeFile is synchronous. Since that will indeed force us to have something like `writeFileAsync` in the future, while I'd rather have us be aligned more with node api's.

Could you maybe add a comment for future reference about this? Other than that, LGTM

History.md
```diff
@@ -250,6 +250,8 @@ N/A
   See [this comment](https://github.com/meteor/meteor/pull/10522#issuecomment-535535056)
   by [@SimonSimCity](https://github.com/SimonSimCity).
 
+* Plugin.fs methods are now always sync and no longer accept a callback.
```
benjamn:
```suggestion
* `Plugin.fs` methods are now always sync and no longer accept a callback.
```

Issue #7455, opened 2016-07-22
Mongo.setConnectionOptions must run before packages that declare Mongo collections are loaded


mjmasn:
When using one or more packages that declare a Mongo collection (for example accounts-base), it isn't possible to call Mongo.setConnectionOptions in the main app code or imports directory. The only workaround is to create a new local package containing the configuration and place it above `accounts-base` in the packages file.

It would be good to have to option to put the settings in settings.json somehow although I'm not sure how that would work with importing PEM files (e.g. the sslCA option).

abernix:
Thanks for filing this issue.

Marking as confirmed bug because after thinking about it/discussing in forums, I'm sure this is true.

https://forums.meteor.com/t/meteor-with-the-new-mongodb-3-2-1-compose-io/17763/11

The work around is to use a package that `use`'s `mongo` and to call `Mongo.setConnectionOptions` from that package.  In your project `packages` list, this package should be above any package which uses a `Mongo.Collection`.

hwillson:
For now we're going to document this limitation and workaround (see https://github.com/meteor/docs/issues/173), instead of addressing this in the codebase. Thanks!

PR #10976, opened 2020-03-22
Adds support to override mongo options via Meteor settings

filipenevola:
In some MongoDB host providers like ScaleGrid and IBM Cloud some developers are getting this error because of certificate:

```
MongoNetworkError: failed to connect to server [sg-meteorappdb-32194.servers.mongodirector.com:27017] on first connect [Error: self signed certificate
```

We can fix the error passing additional options to the Mongo connection but right now Meteor only offers this ability via code in a package and also it's not easy to reference the certificate file inside the project.

This PR provides:
a) ability to set Mongo options using Meteor.settings.overrideMongoOptions;
b) simple way to reference files inside the `private` folder in the root.

Read the docs PR to see how this can be used https://github.com/meteor/docs/pull/662

More info:
- Issues about the limitation of `Mongo.setConnectionOptions` usage https://github.com/meteor/meteor/issues/7455 https://github.com/meteor/meteor/issues/6958
- Forums [discussion](https://forums.meteor.com/t/meteor-with-the-new-mongodb-3-2-1-compose-io/17763/3)

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -7,6 +7,8 @@
  * these outside of a fiber they will explode!
  */
 
+const path = require("path");
+
 var MongoDB = NpmModuleMongodb;
 var Future = Npm.require('fibers/future');
 import { DocFetcher } from "./doc_fetcher.js";
@@ -25,6 +27,10 @@ MongoInternals.NpmModules = {
 // people do use it.
 // XXX COMPAT WITH 1.0.3.2
 MongoInternals.NpmModule = MongoDB;
+
+const FILE_ASSET_SUFFIX = 'Asset';
+const ASSETS_FOLDER = 'assets';
+const APP_FOLDER = 'app';
 
 // This is used to add or remove EJSON from the beginning of everything nested
 // inside an EJSON custom type. It should only be called on pure JSON!
@@ -133,12 +139,17 @@ MongoConnection = function (url, options) {
   self._observeMultiplexers = {};
   self._onFailoverHook = new Hook;
 
+  const userOptions = {
+    ...(Mongo._connectionOptions || {}),
+    ...(Meteor.settings?.packages?.mongo?.options || {})
+  };
+
   var mongoOptions = Object.assign({
     ignoreUndefined: true,
     // See https://github.com/meteor/meteor/issues/10925 for discussion of
     // why this option is not the default.
     useUnifiedTopology: !!options.useUnifiedTopology,
-  }, Mongo._connectionOptions);
+  }, userOptions);
 
   // The autoReconnect and reconnectTries options are incompatible with
   // useUnifiedTopology: https://github.com/meteor/meteor/pull/10861#commitcomment-37525845
@@ -170,6 +181,17 @@ MongoConnection = function (url, options) {
     // set it for replSet, it will be ignored if we're not using a replSet.
     mongoOptions.poolSize = options.poolSize;
   }
+
+  // Transform options like "tlsCAFileAsset": "filename.pem" into
+  // "tlsCAFile": "/<fullpath>/filename.pem"
+  Object.entries(mongoOptions || {})
+    .filter(([key]) => key && key.endsWith(FILE_ASSET_SUFFIX))
+    .forEach(([key, value]) => {
+      const optionName = key.replace(FILE_ASSET_SUFFIX, '');
+      mongoOptions[optionName] = path.join(Assets.getServerDir(),
+        ASSETS_FOLDER, APP_FOLDER, value);
+      delete mongoOptions[key];
+    });
 
   self.db = null;
   // We keep track of the ReplSet's primary, so that we can trigger hooks when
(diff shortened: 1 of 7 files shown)


filipenevola:
Hi @klaussner I don't know if you have read the suggestions from the past issues related to this but people were suggesting `Meteor.settings`. 

I already use `Meteor.settings` in a few packages that I have forks like `publish-composite` and `reloader` to configure start-up things and it's working great. I understand that is maybe a new approach but I don't see why this is bad.

The problem with env vars is that we need one env var for each case and we have a lot of mongo options available then we would need to add one env var for each option.

klaussner:
I have no doubt that `Meteor.settings` would work but an environment variable has a few advantages:

1. We have full control over how the variable is translated to actual driver options. If we encourage direct access to the options, Meteor developers have to adapt their settings when we update the MongoDB driver and some of the options have changed or were deprecated (e.g., `sslCA` was deprecated and replaced by `tlsCAFile` not too long ago). Additionally, we can make sure that the configuration we pass to the driver is always correct and consistent, for example, by setting both `tlsCAFile: "..."` and `tls: true` so that developers don't have to worry about details of the MongoDB API.

2. An environment variable is more in line with existing existing configuration options in Meteor. Consider the following two settings files for Galaxy. I would prefer the first one because everything MongoDB-related is in one place.

```json
{
  "galaxy.meteor.com": {
    "env": {
      "MONGO_URL": "...",
      "MONGO_TLS_CA_FILE": "...",
    }
  }
}
```



```json
{
  "galaxy.meteor.com": {
    "env": {
      "MONGO_URL": "...",
    }
  },
  "overrideMongoOptions": {
    "tls": true,
    "tlsCAFile": "..."
  }
}
```

> The problem with env vars is that we need one env var for each case and we have a lot of mongo options available then we would need to add one env var for each option.

I'd argue that most of the driver options don't need to be configurable and that the few exceptions (like `tlsCAFile`) justify additional variables. In conclusion, I think an environment variable makes more sense in terms of backwards compatibilty and developer experience.

filipenevola:
I understand your arguments but I don't think I agree :). Let's proceed.

When you say that we have full control with env vars that also means that we are limiting the options. For example, I have a project where I connect to 4 MongoDB clusters and I'm going to use `"tlsAllowInvalidCertificates": true` so at least we would need to add two env vars already.

The problem with this approach is that we are not allowing Meteor developers to use the other options, they will always need to ask or implement a new PR. How do you think this is going to work in the long term?

About the developer using wrong options, I believe this is going to be a problem in any case, the developer can always set things incorrectly then we need to have better docs and reply the questions when they come up.

I agree my settings proposal is not clear as it could be, I believe a better version would be:
```json
{
  "galaxy.meteor.com": {
    "env": {
      "MONGO_URL": "...",
    }
  },
  "packagesSettings": {
    "mongo": {
      "tls": true,
      "tlsCAFile": "..."
    }
  }
}
```
The idea here would be to use the package name as a key then it would be easier to find settings later affecting each package.

I'm saying that because as I have said in my previous answer I use settings in a few packages already to change initial configurations. Let me show you the concrete settings so you can see what I'm using:
```json
    "reloader": {
      "refresh": "instantly"
    },
    "publishComposite": {
      "unblock": true
    }
```
These are the packages [reloader](https://github.com/pathable/reloader) and [publish-composite](https://github.com/pathable/meteor-publish-composite/). 

I don't see how to use env vars in all these cases and it would be very hard to configure all these env vars. If we have this pattern using package names as keys it would be like this:
```json
  "packagesSettings": {
    "mongo": {
      "tls": true,
      "tlsCAFile": "..."
    },
    "pathable:reloader": {
      "refresh": "instantly"
    },
    "pathable:publish-composite": {
      "unblock": true
    }
  }
```

But my main concern is to support future cases and also avoid a lot of env vars. With JSON in the settings we can just keep the same names from Mongo then we are not creating new settings, we are just providing a way to set it.

packages/mongo/mongo_driver.js
```diff
@@ -7,6 +7,8 @@
  * these outside of a fiber they will explode!
  */
 
+var path = require("path");
```
StorytellerCZ, line 10: `var` here, `const` down bellow. What version of ES are we using in this file?

packages/mongo/mongo_driver.js
```diff
@@ -133,12 +139,17 @@ MongoConnection = function (url, options) {
   self._observeMultiplexers = {};
   self._onFailoverHook = new Hook;
 
+  const userOptions = {
+    ...(Mongo._connectionOptions || {}),
+    ...(Meteor.settings && Meteor.settings.overrideMongoOptions || {})
```
StorytellerCZ, line 144: I think all Meteor specific settings should be under a specific Meteor key, something like Galaxy has  `"galaxy.meteor.com"` for its env variables. So I'm thinking `Meteor.settings.meteor.mongoOptions`.

packages/mongo/mongo_driver.js
filipenevola: I believe it's a mix :( I'm going to update this anyway.

packages/mongo/mongo_driver.js
filipenevola: Great.

Issue #10979, opened 2020-03-23
[Windows] 7Zip cannot extract dev_bundle due to spaces in filepath


AlexSterk:
- Meteor version: 1.10.1, although this bug seems rather independent from any Meteor version. Commit: 80359f0b2d0ce7edc5fc18170459d2237d78b32b
- Operating System: Windows 10 Pro Version 1909

Basically the title. Because there is a space in my file path, 7zip can't find the dev_bundle tar that was just downloaded. I have to manually extract it to get meteor up and running. I've narrowed the issue down to the following file/line:

https://github.com/meteor/meteor/blob/7d9be8bf39ae060abafb432f2e40003b9c666215/scripts/windows/download-dev-bundle.ps1#L44

Here `$devbundle_zip` should be enclosed in (escaped) quotes. I have a fix for this ready, which I can open a PR for if desired/needed (not sure if I had to report the issue before opening a PR).

Reproduction: checkout the repository to a directory with a space, run `meteor`

Error log:
```
C:\Users\Alexander Sterk\PHPProjects\SA-Meteor>meteor
Will get you a dev_bundle for windows_x64 version 12.16.1.6
Going to download the dependency kit from the Internet
... downloaded
Extracting dev_bundle_windows_x64_12.16.1.6.tar.gz to the dev_bundle directory
ERROR: C:\Users\Alexander\HRUPPROG.EXIT
Can not open the file as archive

ERROR: C:\Users\Alexander\HRUPPROG.TXT
Can not open the file as archive

Move-Item : Cannot find path 'C:\Users\Alexander Sterk\PHPProjects\SA-Meteor\dev_bundle_XXX\dev_bundle_windows_x64_12.16.1.6' because it does not exist.
At C:\Users\Alexander Sterk\PHPProjects\SA-Meteor\scripts\windows\download-dev-bundle.ps1:52 char:1
+ Move-Item  $downloaded_path $target_path
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\Alexan...s_x64_12.16.1.6:String) [Move-Item], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.MoveItemCommand
 
An error occurred while obtaining the dev_bundle.  Please try again.
```

PR #10980, opened 2020-03-23
Fix Windows dev_bundle extraction for running from checkout with spaces in path

AlexSterk:
Fix for #10979 

I already described the issue, but just a quick summary: when running `meteor` from checkout from a directory which has a space in it, 7zip cannot find the dev_bundle tar to extract. This fixes this issue by surrounding the filepath with quotes.

Not sure I can provide any kind of automated test for this though, but for what it's worth, it works!

Note: I wasn't sure if I had to create an issue before opening a PR. Please let me know if this is the way to go in the future, or just a PR is good too!

### DIFF
--- a/scripts/windows/download-dev-bundle.ps1
+++ b/scripts/windows/download-dev-bundle.ps1
@@ -41,7 +41,7 @@ if (Test-Path $devbundle_zip) {
 
 Write-Host "Extracting $TARBALL to the dev_bundle directory"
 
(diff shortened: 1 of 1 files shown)


benjamn:
Note: technically this fix will take effect the next time we rebuild the dev bundle for Windows, which matters for the next version of Meteor (likely 1.10.2), but there's no rush to deploy this change, other than merging it into `devel`, so that Windows users can run from a checkout.

PR #10994, opened 2020-03-30
Try other web.* arches if preferred arch was excluded.

benjamn:
Follow-up to #10824.

If the preferred arch is not available (most likely because it was deliberately excluded), it's better to use another client arch that is available than to guarantee the site won't work by returning an unknown arch.

For example, if `web.browser.legacy` is excluded using the `--exclude-archs` option (introduced by #10824), legacy clients are better served by receiving `web.browser` (which might actually work) than receiving an HTTP 404 response.

If none of the arches in `preferredArchOrder` are defined, only then should we send a 404.

I encountered this issue when `meteor npx wait-on http://localhost:3000` stopped working after I specified `--exclude-archs web.browser.legacy`, because the `wait-on` tool does not provide a `user-agent` header, so it gets classified as a legacy client, so it receives an HTTP 404 response, which does not satisfy the condition it was waiting for (any 2xx response).

### DIFF
--- a/packages/appcache/appcache-server.js
+++ b/packages/appcache/appcache-server.js
@@ -1,5 +1,4 @@
 import { Meteor } from 'meteor/meteor'
-import { isModern } from "meteor/modern-browsers";
 import { WebApp } from "meteor/webapp";
 import crypto from 'crypto';
 import fs from 'fs';
@@ -88,12 +87,12 @@ WebApp.connectHandlers.use((req, res, next) => {
   }
 
   const cacheInfo = {
-    modern: isModern(request.browser),
+    // Provided by WebApp.categorizeRequest.
+    modern: request.modern,
   };
 
-  cacheInfo.arch = cacheInfo.modern
-    ? "web.browser"
-    : "web.browser.legacy";
+  // Also provided by WebApp.categorizeRequest.
+  cacheInfo.arch = request.arch;
 
   // The true hash of the client manifest for this arch, regardless of
   // AUTOUPDATE_VERSION or Autoupdate.autoupdateVersion.
--- a/packages/appcache/package.js
+++ b/packages/appcache/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Enable the application cache in the browser",
-  version: "1.2.6",
+  version: "1.2.7",
 });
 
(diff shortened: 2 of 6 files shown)


sebakerckhof:
I'm not close to a dev station at the moment, but I wonder how this plays along with the other places where we do similar checks:

https://github.com/meteor/meteor/blob/devel/packages/dynamic-import/server.js#L117-L119
and
https://github.com/meteor/meteor/blob/devel/packages/appcache/appcache-server.js#L94-L96
 (although appcache is deprecated technology anyway)

The chance of a non-legacy build working is also rather low, e.g. as soon as we have one async/await in a Meteor core package, it's not gonna work for any of the legacy browsers (But this might change if in the future we move the modern/legacy cutoff).

Some users have been asking to exclude archs in production builds as well ( https://github.com/meteor/meteor-feature-requests/issues/389 ). Which I've been hesitant to do, because the current fallback has 2 problems:
- It returns a 404, which may get you removed from search engines depending on their user agent strings
- It's not a nice themable page.

So if we instead would show a page, which the user can overwrite (e.g. by including a client/unsupported.html), which we display with status code 200 instead of 404, then both these and your problem are solved?

Just some thoughts, I'm fine with either way.

benjamn:
@sebakerckhof Thanks for reminding me about those other places where we check `isModern`. I think I've unified the call sites to use `WebApp.categorizeRequest` now.

> The chance of a non-legacy build working is also rather low, e.g. as soon as we have one async/await in a Meteor core package, it's not gonna work for any of the legacy browsers (But this might change if in the future we move the modern/legacy cutoff).

Right, I'm mostly thinking of exotic user agents that were misclassified as legacy, but actually can handle modern syntax, or don't need to evaluate any JS after fetching the HTML, like the `wait-on` utility.

> So if we instead would show a page, which the user can overwrite (e.g. by including a client/unsupported.html), which we display with status code 200 instead of 404, then both these and your problem are solved?

I thought about returning `HTTP 204 No Content`, but that didn't give any visible indication of why the page was blank. Even though it's riskier in some ways, returning the whole modern HTTP 200 response seems more useful, even if it fails. Static HTML and CSS will be rendered, at least.

Issue #10992, opened 2020-03-30
Meteor less package outdated


DavidSichau:
I have some issues with the meteor less package:

1. It is quite outdated compared to less.js. This leads in our case to some problems with antd4 which requires some features available in less 3.7.0 or later. The new math mode.

2. I do not find the current location of the less package in meteor. All links point to /packages/less. However in the current develop branch this is missing. Where did it move?

3. In addition it would be nice to be able to pass options to the less compiler.

DavidSichau:
I have created an new version with less 3.11.2:

https://github.com/DavidSichau/meteor-less

If there is an interest, I could merge my changes into the official meteor less package.

filipenevola:
Hi @DavidSichau, Meteor less package is here https://github.com/meteor/meteor/tree/devel/packages/non-core/less

Please submit a PR updating it. Thanks.

DavidSichau:
Hi @filipenevola 

thanks for pointing to it. I have created an pull request:

https://github.com/meteor/meteor/pull/10995

Thanks a lot.

BrianMulhall:
@DavidSichau i tried to update this version recently, but couldn't get test group 3 and 8 to pass for some reason. the lru-cache node package that the caching-compiler atmosphere package relies on fails during testing. did you run into this problem too? i couldnt figure out what was causing it to fail

DavidSichau:
@BrianMulhall I saw this error if an error occured in the `loadFile` methode. For example by not finding the file.

BrianMulhall:
@DavidSichau i wasnt on the right track to fix that bug ... I am glad you came along and fixed it tho and showed me how you did it :) great work man

PR #10995, opened 2020-03-31
Updated less package

DavidSichau:
This pull request updates less to the current version 3.11.1 from 2.5.0. Closing #10992.

I added the `javascriptEnabled` flag to less to maintain backwards compatbility, as this was the default behavior before 3.0.0 of less.

In addition I configured the file loader in a way that it loads files as less if they have no ending provided, accoriding to the documentation of less: http://lesscss.org/features/#import-atrules-feature-file-extensions

The main changes in the code were:

- Rewrite the `loadFile` function as this requires to return a promise.
- In addition I refactored the code to use template string syntax.

### DIFF
--- a/packages/non-core/less/.npm/plugin/compileLessBatch/npm-shrinkwrap.json
+++ b/packages/non-core/less/.npm/plugin/compileLessBatch/npm-shrinkwrap.json
@@ -7,19 +7,14 @@
       "integrity": "sha512-7hGhzlcmg01CvH1EHdSPVXYX1aJ8KCEyz6I9xYIi/asDtzBPMyMhVibhM/K6g/5qnKBwjZtp10bNZIEFTRW1MA=="
     },
     "ajv": {
-      "version": "5.5.2",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
-      "integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU="
-    },
-    "amdefine": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.1.tgz",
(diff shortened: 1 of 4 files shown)


DavidSichau:
Will this be released with the next meteor release or will it be released independent?

benjamn:
@DavidSichau Independently, just now! Try `meteor update less`?

benjamn:
When testing this out in an application, I'm seeing
```
Error: Failed to decode Less path: https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,500,300,400italic,600,700
    at decodeFilePath (packages/compileLessBatch/plugin/compile-less.js:188:11)
```
@DavidSichau I know you didn't change the regular expression matching logic in this code
```js
function decodeFilePath (filePath) {
  const match = filePath.match(/^{(.*)}\/(.*)$/);
  if (! match)
    throw new Error(`Failed to decode Less path: ${filePath}`);
```
… but do you have any idea why the `less` plugin might have started passing unresolvable (plain CSS) URLs to the `decodeFilePath` function?

DavidSichau:
It might be that the file was previously not loaded by less. 

Either it was a file included without ending or maybe this change in less: https://github.com/less/less.js/pull/3453

ulion:
I'd like to use the new less package, but I got this:
```
   error: Conflict: Constraint less@2.8.0 is not satisfied by less 3.0.1.
   Constraints on package "less":
   * less@3.0.1 <- top level
   * less@2.8.0 <- accounts-ui 1.3.1 <- accounts-facebook 1.3.2
   * less@2.7.12 <- accounts-ui-unstyled 1.4.2 <- accounts-ui 1.3.1 <- accounts-facebook
   1.3.2
```
While the accounts-facebook is newest even I specify the less version 3.0.1 in .meteor/packages file.

So, any one can help or tell me how to overcome this conflict to use new less package?

Issue #10906, opened 2020-02-09
[1.9] Compatibility with native Node.js ES Modules (in particular the package.json "exports" field).


trusktr:
Meteor 1.9

The exports field is a new feature of Node.js ESM, and it is released into the wild now (Node 13.2+).

Some packages wishing to release pure-ESM code may have something like the following in package.json:

```js
	// This removes 'dist' from import statements, as well as replaces the 'main' field. See https://github.com/nodejs/node/issues/14970#issuecomment-571887546
	"exports": {
		".": "./dist/index.js",
		"./": "./dist/"
	},
```

It allows a dependent project to do this:

```js
import {foo} from 'some-dependency'
// foo was imported from node_modules/some-dependency/dist/index.js

import {bar} from 'some-dependency/path/bar.js'
// bar was imported from node_modules/some-dependency/dist/path/bar.js
```

From what I observed today on Meteor 1.8, meteor is not able to resolve dependencies that I have in my project which are publishing pure-ESM packages with `exports` maps, as Meteor will try to look for the imported files relative to the root of the package, not the mapped locations or mapped aliases.

trusktr:
[`resolver.js`](https://github.com/meteor/meteor/blob/924b5e8c3c5ffb9c2e8b54b402eda3f8a4d28457/tools/isobuild/resolver.ts) is not to complicated and seems like the place to add `"exports"` support. I'd like to take a stab at it.

trusktr:
Updated, added my Meteor version, 1.9.

trusktr:
Another idea is, if Meteor can support [import maps](https://wicg.github.io/import-maps/), then it could be possible to generate import maps based on node_modules dependencies using existing tools in the JS community. 

@benjamn Any ideas on what direction to take this in?

trusktr:
I found the following tool, which may serve as a good example (if not give us some re-usable code to build from) on how to generate import maps from Node.js-based projects using Node's module resolution rules: [@jsenv/node-module-import-map](https://www.npmjs.com/package/@jsenv/node-module-import-map).

One particular thing to note about Node.js ESM: The only way (out of the box) to import a CommonJS module into an ES Module is with `import default`, with no support for named exports (i.e CommonJS modules _only_ have a `default` export, nothing else).

For example:

```js
// ./commonjs-module-a.cjs
module.exports = {
  foo() {},
  bar() {}
}
```

```js
// ./commonjs-module-b.cjs
module.exports = function() {}
```

```js
[...]

PR #10996, opened 2020-03-31
convert base64 to an NPM package (proof of concept)

trusktr:
This is a PR to open discussion on [migrating as much as possible to NPM](https://github.com/meteor/meteor/blob/devel/Roadmap.md#transition-as-much-as-possible-to-npm).

- This is a PoC (you don't want to merge this)!
- `./meteor test-packages oauth mongo ejson` passes (those three packages depended on `base64`)
- The `base64` Meteor package continues to exist for backwards compatibility, and re-exports the NPM package's exports.
- The name of the NPM package is `@trusktr/meteor-base64` only to show that it works, but ideally it'd be `@meteorjs/base64`.
- It builds TypeScript source files and run TypeScript tests using my tool `@lume/cli`. We'd probably want to come up with Meteor's own tool

The basic development steps are:

- edit code in `.ts` files
- edit `.test.ts` files to update tests
- run `npm test` (implies `npm run build`)
- `npm run build` builds to `dist/` in Node.js 13.2+ ESM format.
- `package.json` contains an `exports` field so that `dist/` in not necessary in the import specifiers, and to my surprise it is working. Was support for this added recently? https://github.com/meteor/meteor/issues/10906
- bump the version
- `npm publish`

In my own projects I've been using `Lerna` + `@lume/cli`; every package is a git submodule in my "umbrella repo", and I use commands like `lume releasePatch` to test, build, publish, and finally push.

In Meteor's case there won't be git submodules, so the release commands aren't ideal (at least currently) because they commit and tag in the form `vX.X.X` which will leave confusing commits in the case of a monorepo.

The main point of `@lume/cli` here is it shows we could have shared infrastructure across packages. But if Meteor is a monorepo, then perhaps the approach Babel take is better: infrastructure is installed at the root of the repo, and executed across all packages.

I did it this way because it was simple, as I'd already know how my infrastructure works, but in the long term we'd come up with something Meteor specific.

Other things to note in no particular order:

- As the base64 package is now an NPM package outside the context of Meteor, it no longer benefits from the client/server testing of `./meteor test-packages`. 
- @lume/cli runs tests in Electron with karma and karma-electron (so we get real Electron, Node, and browser APIs, not mock APIs like Jest+undom or Jest+jsdom)
- The dist folder also contains a `global.js` output file, which is the browser global version of the package, so that the package can be used in an old-style `script` tag. The global variable is the camelCase version of the package name (minus the scope name)
- The dist output includes `.d.ts` files for consumption by other TypeScript projects


trusktr:
The tests failed with

```
Error: ENOENT: no such file or directory, open '/home/circleci/project/packages/base64/base64_test.js'
```

Should I perhaps just place an empty `base64_test.js` in the `base64` meteor package wrapper (the tests moved into the NPM package)

trusktr:
Any idea why that `continuous-integration/travis-ci` check is still yellow?

sebakerckhof:
Thanks for starting this discussion @trusktr . I've also been thinking about this lately.
A lot of packages have been "modernized", making most easy to publish to NPM now.

There's some technical challenges. Like some packages use ES features which are not commonly supported (e.g. nested imports, an ES proposal from @benjamn , it's main advantage is performance by delaying parsing). And some organisational challenges, like how to organize the git repo.

But I'm going to bring attention to it.

trusktr:
Cool!

As for the nested imports, if I encounter those, I can just convert them to `import()` style, which is standard in Node and browsers now. But also I can transpile (f.e. with TypeScript) to CommonJS if needed. Probably a good idea if not everyone is on the latest Node.js yet.

I'm going to circle back to this and see about moving the tests into the Meteor TinyTest system. But at least for now, all tests for packages relying on base64 are passing.

filipenevola:
Hi @trusktr, I believe you can move forward. When you need to publish under Meteor npm org let me know (you can pvt me on [Community Slack](https://join.slack.com/t/meteor-community/shared_invite/enQtODA0NTU2Nzk5MTA3LWY5NGMxMWRjZDgzYWMyMTEyYTQ3MTcwZmU2YjM5MTY3MjJkZjQ0NWRjOGZlYmIxZjFlYTA5Mjg4OTk3ODRiOTc).

About nested imports vs dynamic imports, they are not used in the same cases, in many cases we don't want to download the module later or we can't because we are not in an async context but we want just to defer the evaluation and then nested imported is very important. This is very useful in Native apps that will run in low-end devices. Anyway, this is not the case here but I believe when nested imports are necessary the best call would be to keep this package as a Meteor package.

trusktr:
@filipenevola Thanks! What I am thinking to do is move the tests from my custom testing and run them in Meteor's testing. Does that sound good? An upside about it is it would be tested within Meteor's whole test system. A downside about it is the NPM package would be coupled to Meteor's system, but maybe that's totally fine and as long as tests pass then we know we can confidently publish to NPM.

filipenevola:
Hey @trusktr, yes, I'm ok.

filipenevola:

PR #10997, opened 2020-03-31
Tweaks after updating less to version 3.11.1

benjamn:
https://github.com/meteor/meteor/pull/10995#issuecomment-606757614

@DavidSichau Seem ok to you?

### DIFF
--- a/packages/non-core/less/package.js
+++ b/packages/non-core/less/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'less',
-  version: '3.0.0',
+  version: '3.0.1',
   summary: 'Leaner CSS language',
   documentation: 'README.md'
 });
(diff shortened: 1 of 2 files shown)


benjamn:
I went ahead and published `less@3.0.1` just so no one will hit the error I saw, but obviously we can continue iterating on it in this PR before merging.

DavidSichau:
`decodeFilePath` is also called if there is a less parsing error. And `less.render` throws an error. 

However between the less version use previously (2.5.) and now (3.11.1) there are many changes, which might lead to some errors during the render method of less.

benjamn:
Yep, I saw both call sites. Especially in the error case, I would much rather see the actual error from `less.render` than the matching failure that `decodeFilePath` was previously throwing. I think making `decodeFilePath` more forgiving is a reasonable solution here, though I'm open to alternatives.

BrianMulhall:
@benjamn i asked david glasser about a week ago about whether when this package was updated if it would be a good idea to strip out the files he had removed in the forked version of less in this repo's branch https://github.com/meteor/less.js/tree/v2.5.0-no-dist. He told me it may be able to achieve this same effect using Npm.strip(). but that he wasnt sure if it was still something that is necessary to do. I thought I would pass that along

PR #10998, opened 2020-03-31
Add 'variables' as an extension for the less compiler to act on

CarlQLange:
In some projects, it's common to define `site.variables` or `theme.variables` less files. However, these do not appear to get picked up correctly by the less compiler (even though they are imported from `.less` files), so I added `variables` as a file extension for the less compiler to act on. So far, it seems like it works 😅
I suppose it would be even more ideal to have some config in the `package.json` for this, because some projects, like Semantic UI, use less files called `theme.config` and so on as well.

Feel free to close this if it's not relevant. I just wanted to make sure there was some record of this issue!

### DIFF
--- a/packages/non-core/less/plugin/compile-less.js
+++ b/packages/non-core/less/plugin/compile-less.js
@@ -4,7 +4,7 @@ const less = Npm.require('less');
 Plugin.registerCompiler({
   // *.lessimport has been deprecated since 0.7.1, but it still works. We
   // *recommend *.import.less or the imports subdirectory instead.
-  extensions: ['less', 'lessimport'],
+  extensions: ['less', 'lessimport', 'variables'],
   archMatching: 'web'
 }, () => new LessCompiler());
 


CarlQLange:
Totally agree, I wish it wasn't the case. In my situation, I'm using fomantic-ui, which has about 30 of these `.variables` files (and also a `.config`). Ideally, I don't want to rename vendor files, which would mess with eventual upstream merges or so on, but I do want an in-repo version of the project. I could, I suppose, run grunt or something ahead of time, but it would be nicer to have Meteor handle it.

Does the one plugin per extension issue apply also for other plugins that read potential filetypes from package.json or another config file, like for instance `static-assets`? Because if not, and even if so, I wonder if it would be good to allow users to opt-in to this kind of behaviour.

benjamn:
Plugins can technically read from the file system if they need to, which is what I presume packages with dynamic configurations must be doing. If the plugin relies solely on the array of `InputFile`s passed to its `processFilesForTarget` method, those files will be restricted to the file extensions claimed by the plugin via `Plugin.registerCompiler`, and will not be passed to any other plugin's `processFilesForTarget` method.

In case you're wondering, a common pattern for chaining plugins is to have a separate compiler package, like `babel-compiler`, which can be used as a dependency by multiple plugin packages, like `ecmascript`, `typescript`, and `coffeescript`.

As soon as you let multiple plugins claim the same file extension, you have to worry about the order in which they run, which opens a giant can of worms in terms of configuration complexity. The Babel preset/plugin configuration system is a good example of managing a highly configurable plugin pipeline, but it's an exceptionally complex and fragile system. Meteor avoids that complexity by letting developers decide which plugins claim which file extensions, and allowing developers to easily fork packages that aren't behaving the way they want.

On that note, if you need this functionality right away, you could copy the `less` package into your local `packages/` directory and apply the changes from this PR locally.

CarlQLange:
> On that note, if you need this functionality right away, you could copy the less package into your local packages/ directory and apply the changes from this PR locally.

Yes, that's what I've done and how I know the change works :)

> As soon as you let multiple plugins claim the same file extension, you have to worry about the order in which they run, which opens a giant can of worms in terms of configuration complexity. The Babel preset/plugin configuration system is a good example of managing a highly configurable plugin pipeline, but it's an exceptionally complex and fragile system. Meteor avoids that complexity by letting developers decide which plugins claim which file extensions, and allowing developers to easily fork packages that aren't behaving the way they want.

That's a great explanation, thanks a lot. From my perspective I'm happy to close this! Thanks for all your hard work on Meteor, it's really a joy to use.

Issue #4445, opened 2015-05-24
Docs for code-prettify package


dr-dimitru:
Please provide a little of docs (how to use, how it works) on [`code-prettify`](https://atmospherejs.com/meteor/code-prettify) package

stubailo:
Hey, I think this package is no longer being maintained - we used to use it in the Meteor docs, but not anymore. Honestly I think we should just remove it and make it explicitly deprecated.

You can about code prettify here: https://code.google.com/p/google-code-prettify/
You can also try my package for the popular highlight.js library, which is a little more documented: https://atmospherejs.com/simple/highlight.js

dr-dimitru:
Hi, 
Actually we found out how to use this package - and it works well.
Your package `simple:highlight.js` is good, but unfortunately it doesn't work within [`perak:markdown`](https://atmospherejs.com/perak/markdown), please see [#2](https://github.com/stubailo/meteor-highlight.js/issues/2)

duvrai:
@dr-dimitru please share how you made it work.

dr-dimitru:
@duvrai here you go:
```coffeescript
$('pre').addClass 'prettyprint'
prettyPrint()

# Also you can put this into `onRendered` event:
Template.yourTemplate.onRendered ->
   $('pre').addClass 'prettyprint'
   prettyPrint()
```

duvrai:
Great, it works now! Thanks!

stubailo:
With highlight.js, the code would be:

```js
this.$('pre').each(function(i, block) {
  hljs.highlightBlock(block);
});
```

Btw in your code sample you should use `this.$` instead of `$` if you want the highlighting to only apply to the current template and not the entire page.

dr-dimitru:
@stubailo thanks, fixed.
Will try to move to your package.
BTW what the main deference, I mean why we should use your package instead of default one?

stubailo:
I was going to say a difference is that mine is documented, but it looks like it wasn't documented well enough. So probably there isn't a big difference.

hwillson:
Hi @stubailo - relating to your comment in https://github.com/meteor/meteor/issues/4445#issuecomment-105292250:

> Hey, I think this package is no longer being maintained - we used to use it in the Meteor docs, but not anymore. Honestly I think we should just remove it and make it explicitly deprecated.

[...]

PR #11002, opened 2020-04-02
jsparse Package deprecation

BrianMulhall:
The jsparse package doesn't seem to be in use anywhere internally and has not been updated in 4 years. I was thinking it might be time to deprecate this package. I used the code-prettify package and the issue linked below for a guide as to how deprecation should occur.

https://github.com/meteor/meteor/issues/4445


StorytellerCZ:
@benjamn

PR #11008, opened 2020-04-04
Fix: Rare exception while invoking method 'tinytest/clearResults' 

mrauhu:
from command `meteor test-packages`.

Probably the `tinytest/clearResults` method called twice on client, but instead of debugging all tests of Meteor on the client side we make a small fix on the server side.

Best wishes,
Sergey.

### DIFF
--- a/packages/tinytest/tinytest_server.js
+++ b/packages/tinytest/tinytest_server.js
@@ -78,11 +78,16 @@ Meteor.methods({
   'tinytest/clearResults'(runId) {
     check(runId, String);
 
-    handlesForRun.get(runId).forEach(handle => {
-      // XXX this doesn't actually notify the client that it has been
-      // unsubscribed.
-      handle.stop();
-    });
+    const handles = handlesForRun.get(runId);
+    // Fix: Exception while invoking method 'tinytest/clearResults' TypeError: Cannot read property 'forEach' of undefined
+    // Second call of the method will always run without exception
+    if (typeof handles !== 'undefined') {
+      handles.forEach(handle => {
+        // XXX this doesn't actually notify the client that it has been
+        // unsubscribed.
+        handle.stop();
+      });
+    }
 
     handlesForRun.delete(runId);
     reportsForRun.delete(runId);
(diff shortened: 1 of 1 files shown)


packages/tinytest/tinytest_server.js
```diff
@@ -78,15 +78,13 @@ Meteor.methods({
...
-    // Fix: Exception while invoking method 'tinytest/clearResults' TypeError: Cannot read property 'forEach' of undefined
-    // Second call of the method will always run without exception
-    if (typeof handles !== 'undefined') {
-      handles.forEach(handle => {
-        // XXX this doesn't actually notify the client that it has been
-        // unsubscribed.
-        handle.stop();
-      });
+    if (handlesForRun.has(runId)) {
+      handlesForRun.get(runId).forEach(handle => {
```
mrauhu, line 82: @sebakerckhof problem is, value `handlesForRun.get(runId)` maybe `undefined`, and you trying using the method `.forEach()` for undefined.

mrauhu: `
```suggestion
      const handles = handlesForRun.get(runId);
      // We checking for type, for avoiding `undefined.forEach()`
      if (typeof handles !== 'undefined') {
        handles.forEach(handle => {
```

sebakerckhof: But it would only return undefined if the map doesn't have the key, right? It's initialized as a Set and I don't see any place where this can change except for the delete. Am I overlooking something?

mrauhu: @sebakerckhof is some cases we may get this state for variable `handlesForRun`:
```js
const handlesForRun = new Map([
  // Correct run with `set` of handles, `.forEach()` is work
  ['runId1', new Set(handle1, handleN)],
  // Incorrect run with `undefined`, `.forEach()` is doesn't work
  ['runId2', undefined],
]);
```

So, for `runId2`:

```
const runId = 'runId2';

handlesForRun.get(runId).forEach(handle => {
  // ...
});
```

You get the error:

```
Uncaught TypeError: Cannot read property 'forEach' of undefined
```

sebakerckhof: Yes, I understand what you mean. What I'm questioning is how we can get into that situation. Because this would mean there is a `handlesForRun.set(runId2, undefined)` somewhere, which as far as I can tell, should not happen. So there may be a deeper problem if this is the case.

filipenevola:

PR #11011, opened 2020-04-06
Update to the RoadMap (Minimongo)

BrianMulhall:
I can work on updating the minimongo package, there is a good overview of future work in the NOTES.txt file inside the package itself. I can start with some smaller changes and work my way up to adding support for indexes


StorytellerCZ:
That would be much appreciated!

filipenevola:
Hi @BrianMulhall, that would be great.

Issue #10671, opened 2019-08-13
Sourcemaps on development build with ROOT_URL


joaomframos:
As discussed [about a year ago,](https://github.com/meteor/meteor/issues/9747) I have an application that runs on top of other websites performing several external requests.

Although Meteor doesn't currently support Sourcemaps for production, there are specific use cases where we are required to test a development build on top of edge cases found on client's websites.

For the the first time ever, we noticed a slowdown on the client's webpage which would affect every user on that platform. We managed to pinpoint the root cause of the slowdown to the development sourcemaps being requested to the client's webserver, spamming it. Probably it's nothing but a really weak webserver but one of our main goals it to never negatively impact a client's webpage but if sourcemaps are set up like this, we can never achieve that goal.
Also, we have been forced to develop without sourcemaps for months already and having this issue fixed would also improve our development workflow.

In order to pinpoint the issue I forked the project and performed the changes locally. The changes were simple and only in two files for now: tools/isobuild/isopack.js and tools/isobuild/bundler.js

I still need to find out if there's a way to fix the imports from node_modules' CSS (like Google's Material CSS) which also enforce a sourcemap request to the current page instead of targeting the domain where the app is running.

The changes are mostly using ROOT_URL if available - which is our use case - but I'm not aware of every use case Meteor covers and that's where I require your help before submitting a PR.

This is one [example of a change to a meteor tool.](https://imgur.com/MHgzwTS)

Which results in a change from: 
//#sourceMappingURL=/dynamic/imports/[PATH_TO_FILE]/6f1a7e62594dde76996770d7b428fae06ad86cef.map

to:
//#sourceMappingURL=[ROOT_URL]/dynamic/imports/[PATH_TO_FILE]/6f1a7e62594dde76996770d7b428fae06ad86cef.map

these changes have been performed on top of Meteor 1.8.1.

HugoMag:
Any news regarding this?

joaomframos:
Sorry @filipenevola for the poke, but are there any updates regarding this?
I would like to at least know if there are any plans to address this issue.

filipenevola:
Hi @joaomframos I would be glad to review your PR about this.

PR #11013, opened 2020-04-07
Sourcemaps are now being loaded based on the ROOT_URL

joaomframos:
PR submitted to approach #10671.

Up until now SourceMaps have always been loading based on the webpage we are currently on which, for use cases allowed by ROOT_URL, may not always happen - the webserver may differ from the current domain of the webpage.

Just a quick note: it's not the first time I request a change to Meteor that [doesn't fit other users' use cases](https://github.com/meteor/meteor/commit/a646bdbbd4b988f1f25c86159972fb1b70c13ce9). I'm submitting this PR as requested but as I stated in the original issue, I'm not aware of the possible repercussion this change may have for other users.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -2828,7 +2828,7 @@ function addSourceMappingURL(data, url, targetPath) {
 
   parts.push(
     newLineBuffer,
-    Buffer.from("//# sourceMappingURL=" + url, "utf8"),
+    Buffer.from("//# sourceMappingURL=" + (process.env.ROOT_URL || "") + url, "utf8"),
     newLineBuffer // trailing newline
   );
 


zodern:
I have a couple of concerns with using ROOT_URL in `meteor build`.

Some apps are built once, and multiple instances of them are run with different ROOT_URLs. A similar use case that I am considering supporting in Meteor Up is the app is built once and deployed to staging. After it is run in staging for a while with no issues, the app is deployed to production using the previously built bundle.

Source maps for the main client bundle should be loaded from the same domain as the js file, so it seems this is only an issue for dynamic imports. Is this correct?

For code executed by `eval`, source maps are loaded relative to the sourceURL comment. Instead of adding the comment during build, the `dynamic-import` package could add it to the files when reading them from disk.

nikhilgoswami:
When can we expect a merge for this?

joaomframos:
@filipenevola  Wasn't this scheduled for 1.12? 

Just noticed that several changes were performed on top of the dynamic-import package that will conflict with the changes from this PR.

nikhilgoswami:
@joaomframos new PR?

filipenevola:
Hi @joaomframos some tests were not ok so we couldn't include it.

but we always have new Meteor versions so feel free to finish this up.

StorytellerCZ:
@joaomframos Do you need to change things or is this OK to merge?

joaomframos:
Seems Ok

Issue #10850, opened 2020-01-08
1.8.2 consume high CPU resources while running dev environment


naturom:
- A short, but descriptive title. The title doesn't need "Meteor" in it.
As title, similar to this one https://forums.meteor.com/t/node-running-very-hot-since-1-4-background-update/27038 but its on near latest version. And strangely it happened on 2 out of our 3 windows laptops only.

- The version of Meteor showing the problem.
1.8.2

- The last version of Meteor where the problem did _not_ occur, if applicable.
maybe 1.8.1

- The operating system you're running Meteor on.
Windows

- The expected behavior.
CPU not consuming 100% resources

- The actual behavior.
CPU consuming 100% resources

- A **simple** reproduction!
To hit 100% it required a relatively big project, but on normal laptop, it only consumes about 20% or less. We tried on smaller proj as well, the normal laptop consumes less than 10% while the other 2 laptops consumes above 30%. 3 laptops are equivalent in specs.

Anyone having same issue? Checked on the node version, meteor version, all same.

benjamn:
Can you tell which process is using the CPU? There's a build process and a server process. I can help identify which is which if you copy/paste the `node.exe` lines from Task Manager.

Can you try Meteor 1.9, just to see if the problem is related to the Node version?
```sh
meteor create --release 1.9-rc.3 --full 19rc3-test-app
```
I realize a fresh app may not exhibit the same CPU usage as your actual app.

larswolter:
We observed the same high cpu load with 1.8.2 it occurs on the build process. 
But we don't have the same high usage in all our projects so we are still investigating why it happens in some of them. Will try 1.9, hopefully end of the week.

sptessier:
I was experiencing the same issue when I switch from 1.8.1 to 1.8.3 but a colleague of mine wasn't having the issue. We realize that I had a environment variable that was set on my workstation:

METEOR_WATCH_PRIORITIZE_CHANGED=false

I removed this from my environment and build times and CPU consumption are back to normal. It seems like Meteor > 1.8.1 doesn't like this option any more. Hope this helps!

naturom:
> Can you tell which process is using the CPU? There's a build process and a server process. I can help identify which is which if you copy/paste the `node.exe` lines from Task Manager.
> 
[...]

PR #11018, opened 2020-04-09
Reduce stats and watched files

zodern:
In many places Meteor uses `readdir` to get a list of items in a folder, and then uses `stat` to check if the item is a directory, file, or symlink. Node now has an option for `readdir` which provides that information. There is very minimal overhead since Node already has the information when reading the directory, but normally it doesn't use it.

Almost always the type is the same as using `lstat`, but there are some rare situations where it is instead the same type as using `stat`. I tried to only updated the places where it isn't important which one it is.

In a medium sized app, this reduced the uncached stats from 56,362 to 23,253 when running Meteor from a git checkout.

To improve performance of other aspects of building, Meteor 1.8.2 removed some checks to prevent watching files in node_modules for `optimisticStatOrNull` and `optimisticLStatOrNull`, and started calling `optimisticStatOrNull` instead of `statOrNull` for most files in the app, including files in `node_modules`. Removing those calls to `optimisticStatOrNull` reduces the number of unique files watched from 47,194 to 18,005 for a medium sized app, which is a similar number to Meteor 1.8.1.

Also, Meteor stats each path it watches to get the file's Inode number to work around a bug. The bug doesn't happen on Windows. Also, due to how Node calculates the `ino` value on Windows, there are occasionally duplicate values for different files which could cause Meteor to not watch both files. For both reasons I disabled this on Windows. The other option is to use the `bigint` option which fixes the duplicate values, but this would have a minor performance impact.

### DIFF
--- a/tools/fs/files.ts
+++ b/tools/fs/files.ts
@@ -1625,8 +1625,10 @@ function wrapFsFunc<TArgs extends any[], TResult>(
         }
       }
     }
-
-    const result = fn.apply(fs, args);
+    let result;
+    Profile.time(`files.${fnName} - uncached`, () => {
+      result = fn.apply(fs, args);
+    })
 
     if (options && options.dirty) {
       options.dirty(...args);
@@ -1645,10 +1647,10 @@ function wrapFsFunc<TArgs extends any[], TResult>(
 }
 
 const withCacheSlot = new Slot<Record<string, any>>();
-export function withCache<R>(fn: () => R): R {
+export const withCache =  Profile('files.withCache', function withCache<R>(fn: () => R): R {
   const cache = withCacheSlot.getValue();
   return cache ? fn() : withCacheSlot.withValue(Object.create(null), fn);
-}
+})
 
 export const dependOnPath = dep<string>();
 
@@ -1741,10 +1743,14 @@ wrapFsFunc<[string], string>("realpath", fs.realpathSync, [0], {
 });
 
 export const readdir =
-wrapFsFunc<[string], string[]>("readdir", fs.readdirSync, [0], {
+wrapFsFunc<[string, { encoding: 'utf-8' | null, withFileTypes: true }], string[] | fs.Dirent[]>("readdir", fs.readdirSync, [0], {
   cached: true,
-  modifyReturnValue(entries: string[]) {
-    return entries.map(entry => convertToStandardPath(entry));
+  modifyReturnValue(entries: string[] | fs.Dirent[]) {
+    if (entries.length === 0 || typeof entries[0] === 'object') {
+      return entries
+    }
+
+    return (entries as string[]).map(entry => convertToStandardPath(entry));
   },
 });
 
--- a/tools/fs/files.ts
+++ b/tools/fs/files.ts
@@ -1628,7 +1628,7 @@ function wrapFsFunc<TArgs extends any[], TResult>(
     let result;
     Profile.time(`files.${fnName} - uncached`, () => {
       result = fn.apply(fs, args);
-    })
+    });
 
     if (options && options.dirty) {
       options.dirty(...args);
@@ -1650,7 +1650,7 @@ const withCacheSlot = new Slot<Record<string, any>>();
 export const withCache =  Profile('files.withCache', function withCache<R>(fn: () => R): R {
   const cache = withCacheSlot.getValue();
   return cache ? fn() : withCacheSlot.withValue(Object.create(null), fn);
-})
+});
 
 export const dependOnPath = dep<string>();
 
@@ -1735,7 +1735,6 @@ export const rename = isWindowsLikeFilesystem() ? function (from: string, to: st
   }).await();
 } : wrappedRename;
 
-// Warning: doesn't convert slashes in the second 'cache' arg
 export const realpath =
 wrapFsFunc<[string], string>("realpath", fs.realpathSync, [0], {
   cached: true,
@@ -1747,7 +1746,7 @@ wrapFsFunc<[string, { encoding: 'utf-8' | null, withFileTypes: true }], string[]
   cached: true,
   modifyReturnValue(entries: string[] | fs.Dirent[]) {
     if (entries.length === 0 || typeof entries[0] === 'object') {
-      return entries
+      return entries;
     }
 
     return (entries as string[]).map(entry => convertToStandardPath(entry));
--- a/tools/isobuild/meteor-npm.js
+++ b/tools/isobuild/meteor-npm.js
@@ -300,6 +300,8 @@ Profile("meteorNpm.rebuildIfNonPortable", function (nodeModulesDir) {
         return scan(path, true);
       }
 
+      // We want to handle symlinks to
+      // directories as directories
       if (item.isSymbolicLink()) {
         item = files.stat(path);
       }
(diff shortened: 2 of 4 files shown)


tools/fs/files.ts
```diff
@@ -1625,8 +1625,10 @@ function wrapFsFunc<TArgs extends any[], TResult>(
         }
       }
     }
-
-    const result = fn.apply(fs, args);
+    let result;
+    Profile.time(`files.${fnName} - uncached`, () => {
+      result = fn.apply(fs, args);
+    })
```
zodern, line 1631: It is sometimes useful to see how many cache misses there are, but since this adds additional overhead I can remove it.

tools/fs/files.ts
```diff
@@ -1741,10 +1743,14 @@ wrapFsFunc<[string], string>("realpath", fs.realpathSync, [0], {
 });
 
 export const readdir =
-wrapFsFunc<[string], string[]>("readdir", fs.readdirSync, [0], {
+wrapFsFunc<[string, { encoding: 'utf-8' | null, withFileTypes: true }], string[] | fs.Dirent[]>("readdir", fs.readdirSync, [0], {
```
zodern, line 1746: I am struggling with finding the correct type for this. This is the reason the tests are failing on Circle CI.

tools/isobuild/meteor-npm.js
```diff
@@ -528,10 +531,10 @@ const isPortable = Profile("meteorNpm.isPortable", dir => {
 
   const result = hasBuildScript
     ? false // Build scripts may not be portable.
-    : optimisticReaddir(dir).every(
+    : optimisticReaddir(dir, { withFileTypes: true }).every(
```
zodern, line 534: `optimisticReaddir` doesn't seem to watch the contents of the directory. Before, only the list of names could become stale, and now the types can become stale also.

zodern: Optimistic functions don't cache when an argument is an object, so my above comment is partly incorrect. I will need to check if this causes performance regressions in some cases.

larswolter:
I tested the branch in one of our projects and CPU spikes are lower then using 1.10.1. But it is still much more then 1.8.1
Is there an easy way to debug the contents of the watch set to further investigate what is causing the spikes?

tools/fs/files.ts
sebakerckhof: Or behind an envar? We already have some similar ones like the `METEOR_TEST_PRINT_LINKER_CACHE_DEBUG` one

zodern:
> Is there an easy way to debug the contents of the watch set to further investigate what is causing the spikes?

Many of the files Meteor watches are not in the watch set and watched for the optimistic functions. I added this to the top of `safe-watchers.js`:
```js
const { createWriteStream } = require("fs");

const stream = createWriteStream(__dirname + '/.meteor-log-watch.log', { flags: 'a' })
```

And then in the beginning of [safewatcher.watch](https://github.com/meteor/meteor/blob/efeccc85d6c192cfce955423f4db640e330d7d89/tools/fs/safe-watcher.ts#L390-L400) I add
```js
  stream.write(absPath + '\n');
```

This will create a `.meteor-log-watch.log` file with a separate line for each path that was watched. Meteor deduplicates watchers by path, so to get an accurate count you will want to remove duplicate lines (usually 1/4 - 1/2 are). You can also make these changes to versions of Meteor in `~/.meteor/packages/meteor-tool` (or `~/AppData/Local/.meteor/packages/meteor-tool` on Windows), though when running Meteor from a git checkout it watches more files since all core packages are local packages. It will always append, so between builds you might want to empty or delete the file.

tools/fs/files.ts
zodern: That is an option, though the question is if anyone would use it. If not, I could just add it in locally whenever I am profiling something where it matters.

sebakerckhof: True, I honestly had to search for this option because I knew it was there but I'd be very surprised if it's ever used. An abundance of options is not much use if it's not documented, let alone if it's never used.

filipenevola:
Hi @zodern and @larswolter, I would like to say that my fix https://github.com/meteor/meteor/pull/10730 was also published in 1.8.2 and it also caused an increase in the build time so fixing the watchers is probably not going to bring back the build times to the same level before 1.8.2.

I could introduce a new env var to disable my fix but if people use this env var they would get eventually `cannot find module` errors again. This is for sure going to happen in large apps using packages then I don't think it is worth having this option.

I still want to invest more time trying to improve my fix but I'm not sure if that is possible.

For the record, after we realized that build times were increased I did some analysis (on December 16th, 2019), you can check the times below that I got by running the same project (a big one) with and without my fix:
```
1.8.2              = (#1) Total: 1,157 ms (ProjectContext resolveConstraints)
1.8.2-no-fix-10730 = (#1) Total: 1,460 ms (ProjectContext resolveConstraints)

1.8.2              = (#2) Total: 2,232 ms (ProjectContext prepareProjectForBuild)
1.8.2-no-fix-10730 = (#2) Total: 3,140 ms (ProjectContext prepareProjectForBuild)

1.8.2              = (#3) Total: 57,822 ms (Build App)
1.8.2-no-fix-10730 = (#3) Total: 44,915 ms (Build App)

1.8.2              = (#1) Total: 8,247 ms (Server startup)
1.8.2-no-fix-10730 = (#1) Total: 6,452 ms (Server startup)
```

At that time I had analyzed also 1.9 (updated Node.js) and server startup was around these seconds:
- 1.8.1 = 4 seconds
- 1.8.2 = 8 seconds
- 1.9.0 = 6 seconds

My conclusion at that time (I don't remember everything now as it was in December) was that my fix was responsible for 50% of the build time increase, maybe the other 50% was the watcher change.

zodern:
@filipenevola The changes that caused more files to be watched reduce rebuild time, though setting up the additional watchers would have contributed to the increase in the initial build time compared to 1.8.1.

The goal with reducing the number of watched files is to reduce cpu usage of the meteor-tool when not building (#10850 is caused by Meteor 1.8.2 increasing the number of watched files). Meteor uses a stat file watcher that normally checks files every 5 seconds which can use a large amount of CPU.

larswolter:
Hi @zodern, thanks for the debugging tipps. I compared the watched files, reducing to unique lines and sorting them:
- 1.8.1 around 11.000 lines
- 1.10.1 around 75.000 lines
- #11018 around 32.000 lines

One finding: under .meteor/local 1.8.1 only contains 200 entries from the bundler-cache/linker, but the other two  contain ~5000 from bundler_cache/scanner.

A second finding is that the 1.10.1 versions both contain more watched files in addtion to the directories, like all the cmd and ps1 files in the node_modules/.bin directory or the README.md and LICENSE files from all npm packages (+1400 entries). Other direct file entries for .png or .txt also exist a lot.

Hope that helps.

brucejo75:
@zodern, @filipenevola, is [this](https://github.com/meteor/meteor/pull/11018#issuecomment-615531197) going to get resolved?
Will this be merged?  Or some other fix found? 

I run 2 meteor servers in my development and the CPU usage is a real drag.

zodern:
I am closing this in favor of https://github.com/meteor/meteor/pull/11102. I implemented it a little differently there, and also fix cache files being watched.

PR #11023, opened 2020-04-16
Add an idea to Roadmap about making Meteor APIs consumable outside of…

trusktr:
EDIT: Maybe this is a duplicate of the other goal, [Transition as much as possible to NPM](https://github.com/meteor/meteor/blob/9aebd156f0a3e7c646fa0ed550514fbfcda9c3de/Roadmap.md#transition-as-much-as-possible-to-npm).

---

Is this a good idea, not just from a developer perspective, but from a business/corporate perspective?

As a developer, I think it'd be awesome to be able to import full-stack Meteor features on both sides of an existing application that isn't already managed with Meteor.

From a business perspective, it can go both ways: if developers can import APIs into existing Node.js backends and browser frontends, then they may be less likely (?) to host on Galaxy? But at the same time, there may be a gain of users due to this, which may mean more business possibilities.

EDIT: `CONTRIBUTING.md` says

>  2. Nothing in Meteor should preclude an expert from doing what they
 want. 

Perhaps making all of Meteor APIs consumable outside of Meteor's build system may allow experts to do what they want more easily.

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -28,6 +28,22 @@ Implement tree shaking / dead code elimination, which involves pruning the depen
 
 A proper service worker build target. Regular Web Workers can be built from a function.toString() but service-workers require an actual server route.
 
+### Make it easy to use Meteor in external projects
+
+- Leaders: <you?>
+- Status: -
+- PRs: -
+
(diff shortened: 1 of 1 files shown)


trusktr:
On second thought, maybe this is already too close to the other goal, [Transition as much as possible to NPM](https://github.com/meteor/meteor/blob/9aebd156f0a3e7c646fa0ed550514fbfcda9c3de/Roadmap.md#transition-as-much-as-possible-to-npm).

trusktr:
I'll just close this. I should've read the whole Roadmap more closely. :)

PR #11039, opened 2020-05-01
Advanced meteor methods

Floriferous:
This is a proposal for a new meteor methods API. It builds on the ideas of `validatedMethod`, but mostly `cultofcoders:mutations`, written by @theodorDiaconu.

I believe Meteor needs a modern API for its methods, which hasn't changed and evolved in a long time. I decided to take an opinionated direction, so I'm happy to get your feedback.

## Core features

* Opt-in package
* Inversion of control to the developer, get full access to the meteor method lifecycle, or keep it simple if you don't need it
* Promise-only on the client
* Synchronous-only on the server
* Global and per-method before and after hooks
  * All hooks behave like middleware in chain
  * Add multiple server-only, client-only or server+client hooks, and mix them up!
* Easy client/server code-splitting without littering your code with `Meteor.isServer` and `Meteor.isClient`
* Scalable approach to meteor methods for large code-bases

## Demo

```js
import { Method, methodOptions } from 'meteor/method';

// Create a shared method handle, easily import it anywhere through autocompletion
// Add any additional data to your method that you get access to in your hooks
const insertTodo = new Method({
  name: 'insertTodo',
  tracking: true,
  // Example: pass in your argument validation code here to organize your code base
  // Call this function inside a global before hook
  validateArgs: () => {},
});


//
// Server-side
//

// Update method config on the server for secret config data
insertTodo.updateConfig({ permissions: 'admin' });

// Set the meteor method handle, this actually calls `Meteor.methods`
insertTodo.setHandler(function (title) {
  return Todos.insert({ title });
});

// Add server-side global hooks
methodOptions.addAfterHook(({ result, error, config }) => {
  if (Meteor.isDevelopment) {
    // Example: 3-line logging solution for all your methods
    console.log(config.name + ' called', result, error);
  }
});

// Add multiple server-side global hooks, they're all called
methodOptions.addAfterHook(({ config, context, result }) => {
  if (config.permissions === 'admin' ) {
    // Example: Easy event-driven architecture
    methodEventEmitter.emit(config.name, { userId: context.userId, result })
  }
});

// Add a per-method hook
insertTodo.addAfterHook(({ result, error, config, context }) => {
  if (!error) {
    context.unblock();
    sendEmail(config.name, result);
  }
});


//
// Client-side
//

// Call a method on the client with promises
insertTodo.call({ args: ['Merge PR'] }).then(todoId => {});

// Add client-side global hooks
methodOptions.addAfterHook(({ config, args, result }) => {
  if (config.tracking) {
    analytics.track({ event: 'method called', methodName: config.name, args, result });
  }
});

methodOptions.addBeforeHook(({ config, args }) => {
  if (config.validateArgs) {
    // Example: Organize your argument validation code
    config.validateArgs(args);
  }
});

methodOptions.addAfterHook(({ error }) => {
  if (error) {
    // Example: translate your server-side errors on the client
    const language = getLanguageFromClient();
    const translatedError = translateError(error, language);
    
    // Return an object in your hooks to get it merged with the method outcome, result and error for example
    return { error: translatedError };
  }
});

// Set a different handle on the client if you want optimistic UI
insertTodo.setHandler(function (title) {
  // Can be useful if your server-side method does a lot of complex checks, but
  // you only need to insert something for example
});
```

### Notes

Yes, this could simply be another meteor package, but I think it'd be good for the community if some of these features land in the core. Meteor's core features get very few updates lately, especially things that make hard things easier to build. So you're left hacking things together on your end.


diavrank:
This feature it would be great in the core of Meteor, since some days ago I implemented middlewares for my methods and I had to use ValidatedMethods to do that (adding the MethodHooks mixin), so, I think with your feature could be less code and cleaner than ValidatedMethods.

this is the feature request that I did some days ago:
https://github.com/meteor/meteor-feature-requests/issues/395#issuecomment-616029462

StorytellerCZ:
I agree with @diavrank , the only thing I would look out for would be backward compatibility when adding to core.

StorytellerCZ:
I think at the very least we could leverage the old [mdg:method](https://atmospherejs.com/mdg/method) package for this instead of creating third method package under the official namespace (which could contribute to confusion).

filipenevola:
Great work @Floriferous, we need to make a decision about this being a core package or not and then move forward.

Maybe this could be published on https://github.com/Meteor-Community-Packages?

StorytellerCZ:
@filipenevola As I mentioned above I think making it part of MCP would only increase confusion. I think replacing the mdg:method would be a better move. I think moving forward in a way that would allow to retire projects like method-hooks would be most appropriate.

filipenevola:
Hi @Floriferous I would like to understand better how to use this new package to replace the [validated-method](https://github.com/meteor/validated-method) package.

I believe we need first to understand the differences then we can write a migration path document and also decide if this is really a replacement or just a different approach tackling different problems.

Could you provide a few examples?

With these examples, comparison between it and validated-method and more clarity on the usage of this proposed package, we can make a decision about where it should be published/maintained.

I also would like to use this discussion here about where to create a new package ([core](https://github.com/meteor/meteor/tree/devel/packages)/[non-core](https://github.com/meteor/meteor/tree/devel/packages/non-core)/[mdg](https://atmospherejs.com/mdg?q=mdg%3A)/[mcp](https://github.com/Meteor-Community-Packages/)/standalone) to become a reference to a new document that would be used for future decisions.

sebakerckhof:
I'd personally have us extend validated method or adopt that to core, since it has been the standard recommendation for years now.
I can't really see an advantage over validated methods besides the hooks, which could also be added easily to validated method. And the mixins from validated methods are in some ways more powerful.

Some of the features could also use some more explanation, like:
> Scalable approach to meteor methods for large code-bases

How exactly is it more scalable?

aliogaili:
I agree with @sebakerckhof, I'd rather have it as a separate package or some extension to the validated methods, if possible.

My reasoning is pragmatic.

Validated methods have around for several years now, backward compatibility and stability are two major Meteor selling points relative to other JS frameworks. Unless I've missed major value-add from the proposed package or pain-point from the existing package, then I personally would rather preserve what is out there. 

With that said I think the proposed package is great but perhaps it could be a standalone package for the new greenfield projects, again I'm just trying to weight in the value add to the cost of churn.

afrokick:
What about TypeScript’s typings?

Floriferous:
@sebakerckhof @aliogaili 

## Scalability

The trouble with large meteor apps in my opinion, is to clearly control what's on the client, and what's on the server. You need to control the bundle size (reduce code sent to the client), as well as reduce unnecessary startup code on the client.

Using `Meteor.isClient`, `Meteor.isServer` and `Meteor.isSimulation` is nice in small apps, but it does not scale well (if-else code-separation is terrible). 

By breaking up a method declaration into smaller parts, each usable on the server/client/shared way, it makes it really simple to organise your methods:

* Define your method handles in a shared file
* Define method simulations on the client only
* Define server-side methods on the server only
* Import your method handles somewhere else, and attach hooks to it in another place. Your methods become modular pieces of code that you can easily modify and enhance exactly where you want. You could import all of your method handles in a totally different file and attach security hooks to them, perfectly separated from the business logic.

This is just an example, but it gives you all the flexibilty you want to structure your methods however your business needs.

## Validated methods?

The code for validated methods is really tiny and hasn't evolved in forever, supports a mixin API that would've to be backwards supported, and it simply does not have a modular API at all. 

Sure you can go read the code to learn how to mutate your method's class and do `validatedMethod.validate = () => {}`, but that's not documented anywhere.

The package's name is also an issue to me, there's no such thing as a `method` and a `validated method`. Meteor needs `methods` and that's it, don't make people feel like their methods aren't validated unless they use this 120 LOC package.

So to turn validated methods into what I suggest here, you simply rewrite almost each line from the original package, and I don't see the added value of doing that.

## Typescript

I don't have a lot of experience with it, but should be simple to do!

StorytellerCZ:
@aliogaili Given all that, that is why I'm lobbying to use the `mdg:method` namespace. Historically it only has one release which I think is the first commit before `validated-method`, so no one uses that package and it was never advertised either, so it is perfect for recycling.

aliogaili:
@Floriferous  thanks for the further explanation, the modular approach via setHandler and the more control over what goes in client/server does make sense to me, I also don't want to see server code shipped to the client. 

@StorytellerCZ yeah the last update is in 2015, I think it makes sense 👍

Floriferous:
@StorytellerCZ How do we proceed with `mdg:method`? 

Also, do a few people want to give this PR a thorough code review so we can refine the API?

filipenevola:
hi @Floriferous I'm happy moving forward with this package as it is here.

A few suggestion to finally merge this PR:
- Review if this is the best code that you have for this package (in a call that we had a few months back you said maybe that you have already more improvements to it).
- Create a [guide](https://guide.meteor.com/) page explaining how to use this package with examples, comparing the usage of Meteor.methods and also ValidatedMethod.
- Create a [docs](https://docs.meteor.com/) page about this new package, documenting the API, pretty much what you have done in the PR description maybe with less examples and more API details.
- Create an [example](https://github.com/meteor/examples) deployed to Galaxy as well showing off the features of this package. It could extend the tutorial example or something else. As you have used this package for a long time you probably have some nice use-cases to show off.
- Publish a [blog](blog.meteor.com) post about it.

Yes, I know this is a lot but I'm not asking to you to do all this work.

I could help, @StorytellerCZ could help and many others for sure. But it would be nice to have you at least as the reviewer of this work and also doing the first item in the list because only you have access to your private code with possible more improvements.

What do you think?

Sorry for the long delay, first I was expecting more discussions to come up here and later we had to focus in other PRs.

StorytellerCZ:
@Floriferous I will track down for you the `mdg:method` if you want to proceed in that direction.

brucejo75:
@Floriferous,

Agree that Methods interface needs updating!  Couple of specific comments on your proposal:

1. I wonder if it might be useful to think about this some more?  When I think of meteor methods I compare them to express framework calls:
  - Could we take advantage of a routing concept for meteor methods?  Then we could add specific mixins to subsets of meteor methods.  In my case I have administrative methods, end user methods and public methods.  Each set of methods has a very different authentication profile.
  - The route for the method could be implied from the name?  (e.g. if you start your method name with a `/` it is assumed to be a route).  Or if that cannot work well for backward compatibility maybe create a new field?  e.g. `route:
  - The path could be applied on both the server and the client.

2. I like the idea of always using Promises on the client (I have created a `ValidatedMethod` subclass to do this for me).  But one thing I find very irritating about methods today is having to place the argument in `this` which necessitates the `.call`.  I understand that this is necessary, but it could be easily hidden under the covers

```javascript
// Call a method on the client with promises
insertTodo.call({ args: ['Merge PR'] }).then(todoId => {});

// can become
// Call a method on the client with promises
insertTodo({ args: ['Merge PR'] }).then(todoId => {});
```

radekmie:
@brucejo75 

> 2. I like the idea of always using Promises on the client (I have created a `ValidatedMethod` subclass to do this for me). But one thing I find very irritating about methods today is having to place the argument in `this` which necessitates the `.call`. I understand that this is necessary, but it could be easily hidden under the covers

I think there's a slight misunderstanding. The method itself is an instance, not a function. Therefore, the `call` method is not the `Function.prototype.call` (that actually calls a function with a given `this`), but a standard class method, defined [on the client](https://github.com/meteor/meteor/pull/11039/files#diff-8bf18e1f081941188465f3593be1b4ed6e0d2950152224b85a880306931c7cefR8-R43) and [on the server](https://github.com/meteor/meteor/pull/11039/files#diff-2b554364187fd22e41abdbd707ba1c43393ff5d378299f0ca1935819fdc2c9efR8-R11).

---

As to the PR, I like the idea, however, the execution could be improved.
1. I think that the `call` method should **always** return a promise and have a `callSync` alternative on the server. Otherwise, it's hard to write isomorphic code without using `await`.
2. Instead of creating its own `HookManager` class, it could base on the existing [`Hook` class from the `callback-hook` package](https://github.com/meteor/meteor/tree/devel/packages/callback-hook).
3. It defines some `callback` argument on the client but doesn't really use it...?
4. (minor) The class is called `ClientMethod` on the server and `ServerMethod` on the client.

brucejo75:
@radekmie, 

> I think there's a slight misunderstanding. The method itself is an instance, not a function. Therefore, the call method is not the Function.prototype.call (that actually calls a function with a given this), but a standard class method, defined on the client and on the server.

You are absolutely right, I did misunderstand it!  I see it now...  thanks for taking the time to point this out.

But I think my request is still valid.  Maybe I need to say it more correctly now that I understand it.

There is only 1 method, `.call`, that is used on the client for `ValidatedMethod` instance.  Why bother?  Couldn't the `ValidatedMethod` return a function instead of the instance that requires making the `.call`?

Here is how I wrapped `ValidatedMethod` to output a function that returns a promise, no `.call` required.

```javascript
class ValidatedMethodPromise extends ValidatedMethod {
  constructor(options) {
    const vMethod = super(options);
    return (options) => {
      return new Promise((resolve, reject) => {
        vMethod.call(options, (err, res) => {
          if(err) reject(err);
          else resolve(res);
        });
      });
    }
  }
}
```

I am only calling Methods from the client, I guess there are scenarios where you can run server to server Method calls.  With my technique you could add a `.sync` "method" on the function returned from `ValidatedMethodPromise`, really a sub-function.

```javascript
ValidatedMethodPromise = class ValidatedMethodPromise extends ValidatedMethod {
  constructor(options) {
    const vMethod = super(options);
    const self = this;
    function _extendedFunc(options) {
      return new Promise((resolve, reject) => {
        vMethod.call(options, (err, res) => {
          if(err) reject(err);
          else resolve(res);
        });
      });
    }
    _extendedFunc.sync = Meteor.wrapAsync(vMethod.call, self);
    return _extendedFunc;
  }
}
```

That way the default function is always going to return a `Promise`.  Then you could have sync supported for the server side too.

radekmie:
@brucejo75 No problem!

> There is only 1 method, `.call`, that is used on the client for `ValidatedMethod` instance. Why bother? Couldn't the `ValidatedMethod` return a function instead of the instance that requires making the `.call`?

Well, there are actually more methods there. Most importantly, the `updateConfig` and `setHandler`. On top of that, there are the per-method hooks.

Also, on a different level, it's a problem with (TypeScript) types when the constructor returns basically _anything_ (except `this`). I'm not 100% certain, but I think it's also impacting V8 performance heuristics.

And as a workaround, you can always do something like this:
```ts
function createMethodFunction(options) {
  const method = new Method(options);
  return method.call.bind(method);
}
```

diavrank:
I have been thinking about this and I think it could be better to use classes for the new API of meteor methods, and also by using decorators of typescript to add functionality to each method. For example:

````js
export class UsersController extends Method {

   @Name('user.save')// or like route '/users/save'
   @Validate(SaveUserDto) // DTO is a class which has the structure of the endpoint's parameter
   @BeforeHooks(functionX, functionY)
   @AfterHooks(functionZ)
   saveUser(){

   }

}
````
 
This syntax is inspired on NestJS, so, I think it can take better advantage of the concepts of OOP like inheritance by using classes. This is just an idea, so, I don't know if this is possible to implement in Meteor. I left some ideas like this in the [Meteor forums](https://forums.meteor.com/t/question-for-meteor-developers/57284/27?u=diavrank95).

filipenevola:
We've reviewed this PR, and we don't see a consensus around the topic.

So as we use [Discussions](https://github.com/meteor/meteor/discussions) to understand what the community thinks about a specific topic, I recommend we have a new discussion with the pros and cons of a new package for Methods. Then after a consensus on the solution, we could come up with a new PR or reopen this one.

Issue #11014, opened 2020-04-08
meteor create --full references kadira:flow-router


vparpoil:
`meteor create --full` comes with these dependencies in `.meteor/packages` :
```
kadira:flow-router      # FlowRouter is a very simple router for Meteor
kadira:blaze-layout     # Layout manager for blaze (works well with FlowRouter)
```
(see [here](https://github.com/meteor/meteor/blob/efeccc85d6c192cfce955423f4db640e330d7d89/tools/static-assets/skel-full/.meteor/packages#L22))

Since [this PR](https://github.com/meteor/guide/pull/834) updates the recommended router in the Meteor Guide to `ostrio:flow-router-extra`, it seems a good idea to update the packages proposed in the full app as well.

I can work on a PR if you feel this is needed. And maybe also update the todos example app

dr-dimitru:
@vparpoil Let me know if you need help with it, at least I'd be happy to review your PR.
Also, I'm working on removing `kadira:blaze-layout` from docs as FlowRouter Extra has it's own `this.render()` available in every hook.

dr-dimitru:
Not stale, just pending

filipenevola:
Please make a PR updating the skeleton.

vparpoil:
@filipenevola already done : https://github.com/meteor/meteor/pull/11044

filipenevola:
Yeah, my bad, it's merged 😉

PR #11044, opened 2020-05-05
Replace kadira:flow-router with ostrio:flow-router-extra in meteor create --full

vparpoil:
see issue https://github.com/meteor/meteor/issues/11014

### DIFF
--- a/tools/static-assets/skel-full/.meteor/packages
+++ b/tools/static-assets/skel-full/.meteor/packages
@@ -19,9 +19,8 @@ ecmascript              # Enable ECMAScript2015+ syntax in app code
 typescript              # Enable TypeScript syntax in .ts and .tsx modules
 shell-server            # Server-side component of the `meteor shell` command
 
-kadira:flow-router      # FlowRouter is a very simple router for Meteor
-kadira:blaze-layout     # Layout manager for blaze (works well with FlowRouter)
-less                    # Leaner CSS language
+ostrio:flow-router-extra # FlowRouter is a very simple router for Meteor
+less                     # Leaner CSS language
 
 meteortesting:mocha               # A package for writing and running your meteor app and package tests with mocha
 johanbrook:publication-collector  # Test a Meteor publication by collecting its output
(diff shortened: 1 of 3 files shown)


StorytellerCZ:
Looks good! Maybe also add note into changelog?

dr-dimitru:
@StorytellerCZ do not merge it! It will break

@vparpoil update layout (`/tools/static-assets/skel-full/imports/ui/layouts/body/body.html`):

```html
<template name="App_body">
  {{> yield}}
</template>
```

@vparpoil update routes (`/tools/static-assets/skel-full/imports/startup/client/routes.js`):

```js
FlowRouter.route('/', {
  name: 'App.home',
  action() {
    BlazeLayout.render('App_body', 'App_home');
  },
});

FlowRouter('*', {
  action() {
    BlazeLayout.render('App_body', 'App_notFound' });
  },
});
```

StorytellerCZ:
@dr-dimitru Don't worry, I can't merge.

vparpoil:
I will do the updates requested, but I have one question for @dr-dimitru : it didn't break my local **meteor create --full**. Have you tried it as is ?

dr-dimitru:
> I will do the updates requested, but I have one question for @dr-dimitru : it didn't break my local meteor create --full. Have you tried it as is ?

You're right, it's backward compatible. After trying it I see it's working just fine.

> Edit : looking at the docs again, it seems to me that both usages are correct. Can you describe why you prefer the use of {{> yield}} over {{> Template.dynamic }} (docs)?

1. Inside of the `yield` template we can control rendering behavior via router (one of examples: re-render or do not re-render when router params are changed, or avoid re-render (e.g. blink) on different routes with same template);
2. We put a lot of work into `yield` template to enhance rendering performance;
3. Since we are moving from old router to the up-to-date version my changes follow [flowrouter-extra documentation](https://github.com/VeliovGroup/flow-router/blob/master/docs/templating.md).

vparpoil:
Thank you for the explanations, I didn't know about performance and configuration abilities given by `yield`

PR #11045, opened 2020-05-06
Update email dependencies

StorytellerCZ:
Updated dependencies of the `email` package and bumped the version to 1.3. The tests ran fine (though more testing is always encouraged). From the [changelog](https://github.com/nodemailer/nodemailer/blob/master/CHANGELOG.md) the biggest change is dropping support for Node bellow v6 (we had here a special version that supported Node v4) and various fixes and improvements.

TODO:
- [x] Update doc

### DIFF
--- a/History.md
+++ b/History.md
@@ -7,6 +7,9 @@
     instead of `dns.lookup()` which might be breaking on some environments.
     See [nodemailer changelog](https://github.com/nodemailer/nodemailer/blob/master/CHANGELOG.md) for more information.
 
+### Changes
+
+* `email` package now exposes `hookSend` that runs before emails are send.
 
 ## v1.10.2, 2020-04-21
 
--- a/History.md
+++ b/History.md
@@ -1,3 +1,9 @@
+## vNEXT, unreleased
+
+### Changes
+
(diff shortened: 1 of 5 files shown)


afrokick:
> 5.0.0 2018-12-28
>Start using dns.resolve() instead of dns.lookup() for resolving SMTP hostnames. Might be breaking change on some environments so upgrade with care

StorytellerCZ:
@afrokick Good point, so major version bump then to be on the safe side?

filipenevola:
hi @StorytellerCZ can you please add this possible breaking change to History.md breaking changes section?

StorytellerCZ:
@filipenevola Done!

StorytellerCZ:
Since we are doing 2.0.0 version I'm also thinking that it might be a good time to make the send hook public. Thoughts?

filipenevola:
@StorytellerCZ it's up to you 😄

StorytellerCZ:
I'll do it. It can actually be pretty useful if you want to send emails via an api or use some other custom stuff instead of `smtps`.

Issue #8883, opened 2017-07-07
detectMobileServerUrl should return the pathname


znewsham:
When running a regular web app, I can use ROOT_URL to set the absolute path of my app e.g., `http://localhost:3000/admin`. However, when running from cordova, this value is replaced by `http://localhost:3000` regardless of how I pass it in, including if passed in with `--mobile-server`, the culprit seems to be `detectMobileServerUrl`, which only returns the protocol, host and port. Adding pathname to this fixes the problem. Should I submit a PR with the change (it's minor)?

I would also expect that whatever parses the `--server` option will need similar modifications

hwillson:
Yes, a pull request would be reviewed for this. Thanks!

Adding for reference: https://github.com/meteor/meteor/issues/5973#issuecomment-182814577

hwillson:
There hasn't been any traction on this issue in quite some time. If anyone comes across this issue and is interested in working on it, please let us know (it is `pull-requests-encouraged`). Until then however, this issue will be closed. Thanks!

PR #11053, opened 2020-05-11
Ensure the pathname of the rootUrl is used in the mobile URL

znewsham:
This has been an annoyance for nearly 3 years - as documented here: https://github.com/meteor/meteor/issues/8883 finally getting around to submitting a PR. If you have a non `/` ROOT_URL, the suffix is stripped off when building for cordova, so your app never connects. E.g., `https://mydomain.com/admin`. 

This package ensures that both the parseUrl function and the cordova `generateBootstrapPage` function respect the suffix of ROOT_URL. I've been making this fix to every version of meteor on every build machine for almost 3 years. Some of my projects use a `/suffix` url and some don't.

### DIFF
--- a/tools/utils/utils.js
+++ b/tools/utils/utils.js
@@ -44,12 +44,15 @@ exports.parseUrl = function (str, defaults) {
   // for consistency remove colon at the end of protocol
   parsed.protocol = parsed.protocol.replace(/\:$/, '');
 
-  return {
-    pathname: parsed.pathname === '/' || parsed.pathname === '' ? undefined : parsed.pathname,
+  var ret = {
     protocol: hasScheme ? parsed.protocol : defaultProtocol,
     hostname: parsed.hostname || defaultHostname,
     port: parsed.port || defaultPort
   };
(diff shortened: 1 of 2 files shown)


znewsham:
@filipenevola I can only fix it by changing the test specification - as the purpose of the code change is to respect the pathname parameter. Is that OK?

znewsham:
@filipenevola tests passing

tools/utils/utils.js
```diff
@@ -44,11 +44,15 @@ exports.parseUrl = function (str, defaults) {
   // for consistency remove colon at the end of protocol
   parsed.protocol = parsed.protocol.replace(/\:$/, '');
 
-  return {
+  var ret = {
```
filipenevola, line 47: We should use `let` and not `var`

znewsham: @filipenevola I'm easy with either - but `var` is used everywhere else in this file. I kept it the same for consistency.

filipenevola (approved):

Issue #11028, opened 2020-04-23
Accounts-base - Log out without unstore token


GabinH:
Hello,

I use accounts-base for my project, and I would like to open a connection from external people during x days.

To open this connection, I create a token (`Accounts._generateStampedLoginToken`) and I insert it in a special user (`Accounts._insertLoginToken`).

Next I can log in with this token (`Accounts.loginWithToken`).

The problem is when I log out, this destroy my token in my database whereas I would like the possibility to reconnect to this token.

How can I do that ? Is there a way the add an options or something to bypass the `this._unstoreLoginToken()`when log out ?

I use Meteor 1.8 and accounts-password 1.5.1

Thanks.

StorytellerCZ:
We could potentially add a hook to check if the token is to be deleted. You could fork the package locally if you want. A workaround would be to create a new token for the user after login. This would probably also be a bit more secure.

PR #11055, opened 2020-05-11
Add the option to keep token when log out.

GabinH:


GabinH:
Hello @StorytellerCZ, need help with this PR (about #11055 ).

Do you see what i would like to do ? How can i improve this code ?

sebakerckhof:
Hi Gabin,

Could you explain the use case a bit? We're trying to keep the amount of options in the Meteor core limited to stuff that is generally useful and I fail to see the use case here.
Also, this has the potential to keep piling up stamped tokens, which may badly influence application performance when not careful.

GabinH:
Hi @sebakerckhof ,

Of course ! Here is my problem : 
I would like to open a connection for people with no account during x days.

Now, I use `Accounts._generateStampedLoginToken` and `Accounts._insertLoginToken ` to create the connection with a special account and I send an URL with the auth token to this people.

Next, when they click on this URL, they can log in with `Accounts.loginWithToken`.

But, when they log out, the token is destroy in my database, so they can't use the URL again.

I am looking for a way to keep the auth token alive even after log out.

StorytellerCZ:
@GabinH I'll take a look
@sebakerckhof The PR references to this issue: https://github.com/meteor/meteor/issues/11028

GabinH:
@StorytellerCZ I understand, but I can't create new token each time.

 In my case, there are users (with accounts) that can send an URL to people (without account). This URL allows external people to connect to a special user account thanks to a token associated to the URL. 
If when they log out I create a new token, the URL will not work again.

sebakerckhof:
@GabinH Another way to approach this is by using a different token instead of the Meteor's resume token. For the sake of the example let's call it a 'PermanentToken'

Meteor's user data model foresees the possibility to have multiple authentication services per user. Resume token is one, then there's potentially a password, oauth services etc.
You could generate and store a PermantentToken on the user (preferably in the services property) and add your own PermanentToken login handler via `Accounts.registerLoginHandler` on the server. The login handler would be called from the client via `Accounts.callLoginMethod` and it would then verify the PermantentToken.

filipenevola:
I'm closing old PRs with problems like tests failing or conflicts. If you are interested in continuing this work please mention me here.

Issue #11043, opened 2020-05-04
[1.10.1]Update link of installing-prerequisites for mobile apps


diavrank:
Hi, I noticed that the link of "Installing prerequisites" for the mobile guide is outdated. Because currently it shows the following link:
http://guide.meteor.com/mobile.html#installing-prerequisites

And looking for in the meteor page I found that the new link is:
http://guide.meteor.com/cordova.html#installing-prerequisites

So, I imagine that you have to update this link in the code, because maybe for the new members of Meteor don't know how to find the new URL.

![Screen Shot 2020-05-04 at 16 12 20](https://user-images.githubusercontent.com/39581375/81014210-158fcf80-8e22-11ea-8611-de3d0fe8edc2.png)
![Screen Shot 2020-05-04 at 16 13 21](https://user-images.githubusercontent.com/39581375/81014304-3a844280-8e22-11ea-868f-73cfeb364077.png)

banjerluke:
In addition to updating the link, I would suggest adding a 302 or 301 redirect for the `/mobile.html` URL. Links to it are still in the wild and likely won't be updated.

PR #11057, opened 2020-05-12
Fix 11043: Update links to cordova prerequisites

sebakerckhof:
https://github.com/meteor/meteor/issues/11043


sebakerckhof:
Btw, would be nice if we can get a 302 redirect as suggested https://github.com/meteor/meteor/issues/11043#issuecomment-624785589

filipenevola:
It was updated here also including other links https://github.com/meteor/meteor/commit/f315d7f318dda8f4564e1608c03edc59c078f7b4

Issue #11060, opened 2020-05-12
verifyEmail URL does is not correctly formated on localhost


jirikrepl:
Meteor version 1.10.2
email package version 1.2.3
OS: MacOS 10.15.4

During development if verifyEmail is sent without MAIL_URL is set, it is printed to console.
But verification URL is printed incorrectly over two lines

```
I20200512-18:28:09.309(2)? ====== BEGIN MAIL #0 ======
I20200512-18:28:09.310(2)? (Mail not sent; to enable sending, set the MAIL_URL environment variable.)
I20200512-18:28:09.315(2)? Content-Type: text/plain
I20200512-18:28:09.315(2)? From: "Typewizz.com" <no-reply@typewizz.com>
I20200512-18:28:09.316(2)? To: jk@gmail.com
I20200512-18:28:09.316(2)? Subject: Verify your email address
I20200512-18:28:09.316(2)? Message-ID: <74bbba31-b7b4-ce02-1763-165f3a9d47ff@typewizz.com>
I20200512-18:28:09.316(2)? Content-Transfer-Encoding: quoted-printable
I20200512-18:28:09.316(2)? Date: Tue, 12 May 2020 16:28:09 +0000
I20200512-18:28:09.316(2)? MIME-Version: 1.0
I20200512-18:28:09.316(2)? 
I20200512-18:28:09.317(2)? To verify your account email, simply click the link below.
I20200512-18:28:09.318(2)? 
I20200512-18:28:09.318(2)? http://localhost:5432/verify-email/yEMr8z6VDu8tyAO80aVSDgsTi9wsSUvA73Tnx9UG=
I20200512-18:28:09.318(2)? sNu
I20200512-18:28:09.318(2)? ====== END MAIL #0 ======

```

If I string last `=` character from verification URL and append the second line. Then I got the working URL with the correct verification token.

In this case:
    http://localhost:5432/verify-email/yEMr8z6VDu8tyAO80aVSDgsTi9wsSUvA73Tnx9UGsNu

sebakerckhof:
It prints the emails exactly as it would be send, thus in the correct email format.
In this case, since it's using the MIME extension, it is subject to [RFC2045](http://www.faqs.org/rfcs/rfc2045.html) (in particular section 6.7), which defines a line length limit of 76 characters. Where longer lengths are wrapped with an equal sign as soft line break.

I understand it's a bit annoying, but this is working as expected.

[...]

PR #11066, opened 2020-05-18
feat: print email verification URL in development

jirikrepl:
solves: https://github.com/meteor/meteor/issues/11060

- print verify email URL on one line in development
- as oppose to URL in the email which is line broken
  because of being complied by MIME

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -186,7 +186,7 @@ Accounts.findUserByUsername =
  * @returns {Object} A user if found, else null
  * @importFromPackage accounts-base
  */
-Accounts.findUserByEmail = 
+Accounts.findUserByEmail =
   (email, options) => Accounts._findUserByQuery({ email }, options);
 
 // Generates a MongoDB selector that can be used to perform a fast case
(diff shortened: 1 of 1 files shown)


packages/accounts-password/password_server.js
```diff
@@ -909,6 +909,9 @@ Accounts.sendVerificationEmail = (userId, email, extraTokenData) => {
   const url = Accounts.urls.verifyEmail(token);
   const options = Accounts.generateOptionsForEmail(realEmail, user, url, 'verifyEmail');
   Email.send(options);
+  if (Meteor.isDevelopment) {
+    console.log(`\nVerification email URL (without MIME): ${url}`);
```
StorytellerCZ, line 913: `(without MINE)` is not needed.
Maybe use the meteor logging facility?
```js
import { Log } from 'meteor/logging';
```

While at it why not add this to reset password, etc. ?

StorytellerCZ: What about using the `logging` facility instead of standard console log?

jirikrepl: Sorry for the lag. I removed `(without MIME)` from the messatge.
I added the same message to reset password and enrollment emails.

I couldn't find any `logging` facility, can you please point to some example?

I saw that Meteor sometimes use `process.stdout.write` e.g. here for  the `client is refreshing` message: https://github.com/jirikrepl/meteor/blob/e85c69680dad10d7bb1be3a28ada62d2b114212e/tools/runners/run-log.js#L191-L197

StorytellerCZ: The Meteor logging is just a suggestion:
https://atmospherejs.com/meteor/logging
It is an internal logging tool, I wrote an article about it:
https://dev.to/storytellercz/meteor-logging-facility-215f

PR #11068, opened 2020-05-19
Update logging package

StorytellerCZ:
Update dependencies and make debug available for use in non production environments.


StorytellerCZ:
@renanccastro What do you think about this? Since we have this package I think it should be used by internal packages. I would also like to add (for 2.1) a hook so that people can plug in third-party logging services.

renanccastro:
LGTM! I did a quick look and seems like we are not using it broadly yet, but would be good to do it moving forward. Also, making it easier to hook log providers would be amazing, it's not exactly easy right now(it's a lot of boilerplate, like [here](https://github.com/quavedev/logs)).

PR #11072, opened 2020-05-22
Creates an example for Cordova apps

filipenevola:
Creates an example for Cordova apps

- [ ] Cordova basic config
- [ ] mobile-config default
- [ ] Push
- [ ] Settings to open in the store when accessing PWA (SSR)
- [ ] Settings for universal links
- [ ] Assets
- [ ] Automatization with Fastlane
- [ ] service worker
- [ ] prerender
- [ ] WebPush
- [ ] Open external links (NSAppTransportSecurity)


StorytellerCZ:
I think we should probably create a separate repository for examples.

PR #11075, opened 2020-05-28
Update PostCSS in minifier-css

afrokick:
Updated PostCSS 4.0.18 to 4.0.31

Solve the issue https://github.com/meteor/meteor/issues/10682

@filipenevola Should i need to update the version for `standard-minifier-css`?


filipenevola:
minifier-css@1.5.1 is published.

PR #11080, opened 2020-06-02
Fixes sendVerificationEmail doesn't work if user is created on the server

### LINKED ISSUE #8694: Accounts.config: `sendVerificationEmail` doesn't work, if user is created on the server.
If I forbid accounts creation from the client, I have to create users through methods. In this case, I expect verification email will be anyway sent to the new user, but it doesn't work.

Linux Mint 18, Meteor 1.4.4.2

```shell
$ meteor create test
$ cd test
$ meteor add accounts-base accounts-password
$ meteor npm install --save bcrypt
$ meteor
```

Then for the sake of the test, I create file `both/both.js`:

```js
Accounts.config({
  sendVerificationEmail: true
});
```

Then I go to browser console and run:

```js
Accounts.createUser({ email: 'mail@mail.com', password: 'pass'});
```

Since I didn't specified `MAIL_URL`, on the server log I see attempt to send email:

```
====== BEGIN MAIL #0 ======
...
====== END MAIL #0 ======
```

Then I run the same code through `meteor shell` (or in my actual method) and I don't see any attempts to send a verification email.

I think at least this behavior should be mentioned in the [related docs part](https://docs.meteor.com/api/accounts-multi.html#AccountsCommon-config).

delki8:
This change addresses issue #8694. 
It contains the implementation of [my second suggestion](https://github.com/meteor/meteor/issues/8694#issuecomment-636237916) to solve the problem. I'm adding a new function to Accounts that will be responsible for creating the user and send an email verification at the end of the registration if everything goes well.

If it gets merged I can also update [the docs](https://docs.meteor.com/api/passwords.html#Accounts-createUser).


StorytellerCZ:
I would prefer if there was some way to distinguish if to send enrollment email or verification email. Probably the easiest would be to detect if password is set or not.

delki8:
Nice @StorytellerCZ. Let me check if I followed you:

With the implementation, I'm giving the devs the option to create a user from the server and send an email verification right after. This will replicate the current client behavior. But I understand that the server has different constraints and allows a user to be created without a password. 

In cases where the password is not present in `options.password` but both `options.email` and `Accounts._options.sendVerificationEmail` are present we would send an _enrollment email_ instead of a _verification email_. The change would be made [here](https://github.com/meteor/meteor/pull/11080/files#diff-b4ff99fd7ae25449bebd21fc644cdea1R1175).

Sounds about right? I'll be changing it in a minute, and if I got something wrong let me know.

packages/accounts-password/password_server.js
```diff
@@ -1146,25 +1146,42 @@ Meteor.methods({createUser: function (...args) {
...
   );
 }});
 
+// Create user directly on the server.
+//
+// Differently from Accounts.createUser(), this evaluates the Accounts package
+// configurations and send a verification email if the user has been registered
+// successfully.
+Accounts.createUserVerifyingEmail = (options) => {
+  options = { ...options };
```
filipenevola, line 1163: Why do we need this line?

delki8: _createUser_ [original implementation](https://github.com/meteor/meteor/blob/6dfa1a6a9f63e7259e0f2383075238207be3dc52/packages/accounts-password/password_server.js#L1190) does this. I think it is related to immutability but idk for sure.

filipenevola (approved):

PR #11086, opened 2020-06-07
Add create command for vuejs

chris-visser:
This PR provides first class Vue integration into Meteor using the create command.


aliogaili:
Great work, I'm going to give Vue a try!

filipenevola:
Thanks, great work, this would be available on Meteor 1.10.3

StephaneRavet:
Hi,

A problem appeared with v1.11 which introduce command "meteor create --vue".

**Actual behavior :** `ReferenceError: window is not defined` if we use `window` in any `.vue` file.

In any `.vue` file, if we use `window.innerHeight` for example, the build will report : `ReferenceError: window is not defined`, because only folders `imports`, `client`, `public `and `test `are excluded from server build. I think, the `src` folder, used by skel-vue is not appropriate.

Ok, you can tell me : use `if (Meteor.isClient) { console.info(window) }` to fix it.
Yes, but if we use a library that uses windows, we can't write `if (Meteor.isClient) { import { Editor } from '@toast-ui/vue-editor' }` to fix it.

**Solution found :**
* rename the `src` folder into `imports`.
* New problem after that : `main.html` is not loaded. If I put it to `/client/main.html`, it builds without errors.

Reproduction : https://github.com/StephaneRavet/meteor-create-vue-window-bug
Solution : https://github.com/StephaneRavet/meteor-create-vue-window-bug/tree/solution

Do you agree ?

chris-visser:
@StephaneRavet Yes. I've noticed too that there are some caveats to the current structure. I will make a PR for it 👍

PR #11098, opened 2020-06-20
Add ability to add extra params to emailed urls

StorytellerCZ:
Allow to attach additional parameters to urls send by accounts. Useless for the default Meteor Accounts-ui, but it allows to pass through the extra attributes to the url builder who can then deal with it as it wishes. Also for ease of use, a new function `buildEmailUrl` has been added as well.
Beside the discussion on the merits of this addition I would also like to talk on usage of "internal" packages like the `meteor/url` package which I decided to use here (although arguable I could just simply construct the url with string manipulation), also in regards to `meteor/logging` which I think we should use to replace any `console.log()` in packages like these.


StorytellerCZ:
The docs will get the extra params for methods like `Accounts.sendEnrollmentEmail` once the code reference is updated there. Maybe this could be mentioned in the guide.
I also should to add tests for this.

filipenevola:
@StorytellerCZ we should have a better way to avoid bumping package versions.

I believe you should bump these already and append -beta200.3 as we do in our versions.

Could you do that?

StorytellerCZ:
Yes, will do it later tonight.

StorytellerCZ:
Done and changed it to merge into the Meteor 2.0 release branch.

filipenevola:
Hey @StorytellerCZ this PR needs to be recreated based on devel and not release-2.0.

Always create PRs based on devel.

PR #11101, opened 2020-06-23
Fix moving css imports after comments to top of merged css

### LINKED ISSUE #11096: There are some @import rules in the middle of a file.
app/client/styles.scss: warn: There are some @import rules in the middle of a file. This might be a bug, as imports are only valid at the beginning of a file.

it failed to compile the scss code as it works before.

mfen:
`minifier-css`'s check for imports to hoist unexpectedly stops on any comments and subsequent imports are not moved to the top of the merged css.

The `rulesPredicate(['import', 'comment'])` check for nodes to hoist is comparing to `node.name`, which is not present for PostCSS's Comment nodes. It should probably be looking for `type: 'comment'` instead.

Proposing a minimal fix to check `name` when present, i.e. for [AtRules](https://api.postcss.org/at-rule.es6.html#line39), and [`type`](https://api.postcss.org/node.es6.html#line423) otherwise for comments. Includes a basic test for import hoisting as there didn't appear to be one, as well as one for the broken case.

Possibly fixes #11096, as it resolved our issue that presented similarly, but need further info/confirmation that they are reporting the same scenario.

### DIFF
--- a/packages/minifier-css/minifier-tests.js
+++ b/packages/minifier-css/minifier-tests.js
@@ -104,3 +104,48 @@ Tinytest.add(
     test.equal(stringifiedAsts.map.sources[0], 'test.css');
   }
 );
+
+Tinytest.add(
+  "minifier-css - hoist imports from merged CSS AST's",
+  (test) => {
+    const css1 = '@import "custom.css"; body { color: "red"; }';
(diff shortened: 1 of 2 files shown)


filipenevola:
Published minifier-css@1.5.2.

PR #11105, opened 2020-06-26
Dynamic Imports: Dynamic Origin Option (CORS)

leonardoventurini:
Add the option for `dynamic-imports` to fetch from the current origin instead of the absolute URL.

### DIFF
--- a/packages/dynamic-import/client.js
+++ b/packages/dynamic-import/client.js
@@ -116,7 +116,15 @@ exports.setSecretKey = function (key) {
   secretKey = key;
 };
 
-var fetchURL = require("./common.js").fetchURL;
+const fetchURL = require("./common.js").fetchURL;
+
+function inIframe() {
+  try {
+    return window.self !== window.top;
+  } catch (e) {
+    return true;
+  }
+}
 
 function fetchMissing(missingTree) {
   // If the hostname of the URL returned by Meteor.absoluteUrl differs
@@ -127,7 +135,21 @@ function fetchMissing(missingTree) {
   // preflight OPTIONS request, which may add latency to the first dynamic
   // import() request, so it's a good idea for ROOT_URL to match
   // location.host if possible, though not strictly necessary.
-  var url = Meteor.absoluteUrl(fetchURL);
+
+  let url = fetchURL;
+
+  // Lo and behold!
+  const useDynamicOrigin = Meteor.settings
+      && Meteor.settings.public
+      && Meteor.settings.public.packages
+      && Meteor.settings.public.packages.dynamicImport
+      && Meteor.settings.public.packages.dynamicImport.useDynamicOrigin;
+
+  if (useDynamicOrigin && location && !inIframe()) {
+    url = location.origin.concat(url);
+  } else {
+    url = Meteor.absoluteUrl(url);
+  }
 
   if (secretKey) {
     url += "key=" + secretKey;
--- a/packages/dynamic-import/client.js
+++ b/packages/dynamic-import/client.js
@@ -2,6 +2,11 @@ var Module = module.constructor;
 var cache = require("./cache.js");
 var meteorInstall = require("meteor/modules").meteorInstall;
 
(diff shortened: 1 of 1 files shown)


packages/dynamic-import/client.js
```diff
@@ -127,7 +135,21 @@ function fetchMissing(missingTree) {
...
-  var url = Meteor.absoluteUrl(fetchURL);
+
+  let url = fetchURL;
+
+  // Lo and behold!
+  const useDynamicOrigin = Meteor.settings
+      && Meteor.settings.public
+      && Meteor.settings.public.packages
+      && Meteor.settings.public.packages.dynamicImport
+      && Meteor.settings.public.packages.dynamicImport.useDynamicOrigin;
```
afrokick, line 146: You can use optional chaining, it’s built in feature for meteor

leonardoventurini: @afrokick yeah I tried that but then I got some compile errors :disappointed:

leonardoventurini: I had some time to take a look at this @afrokick and this is the error I get when I use optional changing...

image

cc @filipenevola

I will try to find a workaround for learning purposes, but using the classical syntax should be fine as well.

leonardoventurini: Apparently the `ecmascript` package itself depends on the `dynamic-import` package, which means that it cannot support newer ECMAScript code as it is part of the parsing implementation itself.

https://github.com/meteor/meteor/pull/9384

If I try to remove these lines I can use the `ecmascript` package (eliminating the circular dependency error) but not sure if it will break something else:

https://github.com/meteor/meteor/pull/9384/files#diff-adfc7aa5c6b9d8ed73279b06d8873386R25-R27

packages/dynamic-import/client.js
```diff
@@ -2,6 +2,11 @@ var Module = module.constructor;
 var cache = require("./cache.js");
 var meteorInstall = require("meteor/modules").meteorInstall;
 
+const dynamicImportSettings = Meteor.settings
+    && Meteor.settings.public
+    && Meteor.settings.public.packages
+    && Meteor.settings.public.packages.dynamicImport;
```
filipenevola, line 8: You could add `|| {}` in the end so you know you always have the settings object.

packages/dynamic-import/client.js
```diff
@@ -127,7 +140,20 @@ function fetchMissing(missingTree) {
   // preflight OPTIONS request, which may add latency to the first dynamic
   // import() request, so it's a good idea for ROOT_URL to match
   // location.host if possible, though not strictly necessary.
-  var url = Meteor.absoluteUrl(fetchURL);
+
+  let url = fetchURL;
+
+  const useLocationOrigin = dynamicImportSettings
```
filipenevola, line 146: If you apply the feedback above you don't need these conditions.

packages/dynamic-import/client.js
sebakerckhof: Do note that this also means you should not use const & let etc.

packages/dynamic-import/client.js
klaussner: I think it would make sense to keep the keys in `packages` consistent with the package names. In this case: `packages["dynamic-import"]` instead of `packages.dynamicImport`. This would make it even more clear that the settings belong to the `dynamic-import` package and might encourage third-party package authors to follow the same pattern, e.g., `packages["some-organization:package-name"]`.

filipenevola:
@leonardoventurini is this ready to be merged? If so please remove WIP label 😉

leonardoventurini:
> @leonardoventurini is this ready to be merged? If so please remove WIP label

Sorry I completely forgot about this, I believe it is ready but I will double-check if I have anything pending.

filipenevola (approved):

rijk:
This works like a charm. After 2 years, I can finally remove my `dynamic-import` monkeypatch. Thanks @leonardoventurini !

PR #11106, opened 2020-06-27
Add `autoupdate-development` package

klaussner:
Inspired by #11034 (`autoupdate-polling`), this PR adds an even more lightweight autoupdate package based on server-sent events. In contrast to `autoupdate-polling`, this package is only intended for development, not for production.

cc @delki8

### DIFF
--- a/packages/autoupdate-development/client.js
+++ b/packages/autoupdate-development/client.js
@@ -1,0 +1,11 @@
+import { ssePath } from "./common.js";
+
+const eventSource = new EventSource(ssePath);
+
+eventSource.addEventListener("message", ({ data }) => {
+  if (data !== __meteor_runtime_config__.clientHash) {
+    window.location.reload();
+  }
+});
+
+window.addEventListener("beforeunload", () => eventSource.close());
--- a/packages/autoupdate-development/common.js
+++ b/packages/autoupdate-development/common.js
@@ -1,0 +1,1 @@
+export const ssePath = "/__meteor__/webapp/autoupdate-development";
--- a/packages/autoupdate-development/package.js
+++ b/packages/autoupdate-development/package.js
@@ -1,0 +1,17 @@
+Package.describe({
+  name: "autoupdate-development",
+  summary: "Development-only autoupdate for clients",
+  version: "1.0.0",
+  debugOnly: true
+});
+
+Package.onUse((api) => {
+  api.use([
+    "ecmascript",
+    "random",
+    "webapp"
+  ]);
+
+  api.mainModule("client.js", "client");
+  api.mainModule("server.js", "server");
+});
--- a/packages/autoupdate-development/server.js
+++ b/packages/autoupdate-development/server.js
@@ -1,0 +1,65 @@
+import { WebApp, WebAppInternals } from "meteor/webapp";
+import { Random } from "meteor/random";
+import { ssePath } from "./common.js";
+
+const responses = new Map();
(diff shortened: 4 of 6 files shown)


filipenevola:
@delki8 can you provide feedback here as you did the other package?

delki8:
Sure @filipenevola and @klaussner. I'm sorry for the late reply. I had a pretty busy month.

1- `autoupdate` and `autoupdate-polling` updates css changes without a hard reload (preserving state) and also uses [`reload`](https://github.com/meteor/meteor/blob/5dcd0b2eb9c8bf881ffbee98bc4cb7631772c4da/packages/reload/reload.js#L212) package to reload the browser avoiding unecessary validations when possible. Did you considered that [here](https://github.com/meteor/meteor/pull/11106/commits/cffbda66260d2016b460d5d2d33e06f8504459da#diff-3df1b3f4922749d2edaf69cd1e7b329cR7)?
2- this [comment](https://github.com/meteor/meteor/pull/11106/commits/cffbda66260d2016b460d5d2d33e06f8504459da#diff-c9fedc8951fc40528d23690e78a64552R61) got me confused: after some server code is changed the client will still be updated with sse or from that point on the client will never be refreshed waiting for the user to restart the server?
3- does `autoupdate-development` work with Cordova apps? Both emulator and real devices? Does it contain any caveats?
4- how long does it take approximately for the client to be updated after a code change?
5- how much does it adds up to a vanilla --minimal application?
6- what are the main points that prevent `autoupdate-development` to also work in production?
7- how would you feel about using polling for production and sse in development (assuming we really can't use sse in production)?

It doesn't seem appropriate to have both `autoupdate-development` and `autoupdate-polling` added to Meteor's core to comply with this [specific topic](https://github.com/meteor/meteor/blob/devel/Roadmap.md#ultra-thin-meteor) from the roadmap. I wouldn't mind taking `autoupdate-polling` out of the race since @klaussner wrote a much cleaner code. Despite that, I think it would be nice to clarify my questions to make a better decision for the platform.

Nice job @klaussner :fist_right::fist_left:

klaussner:
> I'm sorry for the late reply. I had a pretty busy month.

Me too, so no worries! 😄 Thank you for the review.

1. We could probably use the `reload` package instead of calling `location.reload` directly (I will try it out.) but updating CSS without a hard reload would require some code to be copied from `autoupdate`, which I wanted to avoid.
2. When the server process is terminated after it receives the `SIGTERM` signal, the connection will be lost anyway. Explicitly closing the connection is just a way to gracefully shut it down so that no error is shown on the client. The SSE `EventSource` on the client attempts to reconnect automatically afterwards.
3. I haven't tried it with Cordova yet. 👀
4. If no server restart is required, the client reloads immediately. Otherwise, it takes a little longer due to the restart and reconnect.
5. Do you mean in terms of bundle size? The package is `debugOnly`, so it's not included in production mode.
6. & 7. There's nothing that would prevent SSE-based reloads from working in production. The main difference between this package and `autoupdate`/`autoupdate-polling` is the way new asset versions are detected. The Meteor build tool communicates with the server process using IPC, which means that we can install an event handler and just send a reload command to the client. In production, we need a different way to detect new versions, which is what much of the server-side code in `autoupdate` and `autoupdate-polling` is doing.

filipenevola:
Hi, as we are not having many discussions around this PR I'm closing it.

Of course, if we have new ideas we can re-open it. No problem.

PR #11114, opened 2020-07-08
Build in place on Windows during full rebuilds

zodern:
One optimization Meteor uses during builds is build in place. When enabled, Meteor re-uses files from the previous build and only updates the files that changed. On Windows, this was disabled since there could be issues if the meteor tool updates a file the server is accessing. I enabled it for building the server in #10399 since the server is always stopped before a full rebuild is started. Since there hasn't been any issues, this PR enables it for the client architectures during full rebuilds.

Before, writing the client target would take between 600 and 950ms during rebuilds of a medium sized app on Windows.
After, it is between 70 and 215ms.

Building in place is still disabled for client only builds. Meteor 1.8 added the ability for the webapp to be paused during delayed builds. It might be possible to use this to make it safer to use in place builds, but it won't completely prevent issues.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -2844,20 +2844,18 @@ var writeTargetToPath = Profile(
     previousBuilder = null,
     buildMode,
     minifyMode,
+    forceInPlaceBuild
   }) {
     var builder = new Builder({
       outputPath: files.pathJoin(outputPath, 'programs', name),
       previousBuilder,
-      // We do not force an in-place build for individual targets like
-      // .meteor/local/build/programs/web.browser.legacy, because they
-      // tend to be written atomically, and it's important on Windows to
-      // avoid overwriting files that might be open currently in the build
-      // or server process.
-      // Server builds do use an in-place build since the server is always stopped
-      // during the build.
-      // If client in-place builds were safer on Windows, they
-      // would be much quicker than from-scratch rebuilds.
-      forceInPlaceBuild: name === 'server',
+      // We do not force an in-place build for individual targets
+      // like .meteor/local/build/programs/web.browser.legacy, because they tend
+      // to be written atomically, and it's important on Windows to avoid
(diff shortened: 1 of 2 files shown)


CaptainN:
What issues would be caused by writing to files that the server is accessing? Would it prevent the write, crash the server?

I'm partially familiar with the differences in the way Windows and NIX systems handle concurrent file access, but I'm not sure what the consequences would be in this case.

The follow up is, can the issue be handled on the back side in Windows (catch the access error, wait 500ms, then try again, or something) - so that it only slows down Windows rebuilds when this specific access problem occurs.

zodern:
I'm not completely sure, but there are multiple comments in the meteor tool warning against it. Next time I look into this, I plan to try to cause this situation and see what happens. When build in place is disabled, Meteor deletes the folder containing the old files and replaces it with a new folder, and I've been wondering if that is any better than build in place.

zodern:
This comment explains what their concerns were: https://github.com/meteor/meteor/blob/982181fff86e09a75f1d2681a78b9d784e666f4f/tools/isobuild/builder.js#L30-L34

CaptainN:
Another idea would be to pause the running app during rebuild - anything to speed it up is preferable to the way it is now on Windows, IMHO.

zodern:
I mentioned above one option for pausing the app:

> Meteor 1.8 added the ability for the webapp to be paused during delayed builds. It might be possible to use this to make it safer to use in place builds, but it won't completely prevent issues.

Though I have several concerns:
- Other packages access the files, such as dynamic import
- What happens if a file was opened right before webapp was paused and is read while the file is replaced? This is unlikely if we pause it early enough in the build process.
- Another option is to use a sync api (such as execSync) to completely pause the server, but I have the same question as in the previous item. This could also cause other issues if the client build takes long enough.

CaptainN:
Doh! I read that, but it didn't compute. Darn dyslexia haha

PR #11107, opened 2020-06-28
Dead code elimination

filipenevola:
This is a work in progress.

## Documentation

Support for custom flags with static replacement can be used by adding the following on your settings.json:
```
    "globalDefinitions": {
      "isAdminDashboard": false
    }
```
This will make Meteor.isAdminDashboard = false, and Meteor.isAdminDashboard be replaced with false when building, so a code snippet like:
```
if(Meteor.isAdminDashboard){
   console.log('huge string');
}
```
will be removed if you are building with "isAdminDashboard": false. This is useful when sharing the same code base for multiple projects.

## Tasks

- [x] 1. Use babel for replacing meteor internals like Meteor.isServer, Meteor.isClient, for better dead code elimination
- [x] 2. Expose it as an package, for setting custom build-time flags.
- [ ] 3. Gather imports/exports in a map for tree-shaking
- [ ] 4. Flag sideEffectsFree modules
- [ ] 5. Remove/don't include safe modules in the final bundle

PR #11117, opened 2020-07-09
Hot module replacement

zodern:
Hot module replacement (HMR) updates modified files in the app without having to reload the page or restart the app. This is usually faster (many times updating the page before the build has finished), and allows state to be preserved, though it is also less reliable. In situations where HMR is not supported, it uses hot code push to update the page.

HMR is only available for apps that use the `modules` package. It will probably not work correctly with apps that use globals (though globals from packages or npm dependencies are fine) since it uses import/require to detect which modules need to be re-evaluated.

This pull request only enables it for app code in the modern bundle. After this PR is done I will work on supporting other client architectures and packages. 

A partial list of situations it falls back to hot code push:
- Files were modified that do not have a module id, are bare, are json data, or do not have meteorInstallOptions
- Modules were removed from the bundle
- A package was modified

Current status: It works well with many apps, but there are still many details that need to be handled.

### Use:

This PR is included in the Meteor 2 betas: #11206 
```
meteor update --release METEOR@2.0-beta.2
meteor add hot-module-replacement
```

### Integration with View layers

**React** - Components are automatically configured to update using React Fast Refresh. Learn more about it in the [React Native docs](https://reactnative.dev/docs/fast-refresh). React 16.9 or newer required. I haven't checked what happens with apps using an older version. The react runtime is included in production builds, though it is never evaluated. This should be fixed with tree shacking.

**Svelte** - Replace `svelte:compiler` with [`zodern:melte`](https://atmospherejs.com/zodern/melte).

### Hot API

This API is likely to completely change. It is currently based on the webpack hot api, with only the minimum implemented.

`module.hot.accept()` - The module will be rerun whenever it or any of its dependencies are modified

`module.hot.decline()` - If the module or any of its dependencies are modified, hot code push will be used instead of HMR.

`module.hot.dispose((data) => {})` - Adds a callback to run before the module is replaced. Any state that should be preserved can be stored in `data`. The data will be available when the module is rerun at `module.hot.data`.

`module.hot.data` - If the module was reloaded, will have any data set by dispose handlers. If the file was not reloaded, or there were no dispose handlers, it will be `null`.

```
module.onRequire({
	before(module) { return {};  },
	after(module, data) {}
});
```

`onRequire` is only available when `module.hot` is. Use it to run code before and after each module. Both `before` and `after` are optional. Anything returned by the `before` function will be passed as the second argument to the `after` function.

These should be wrapped by
```
if (module.hot) {
  // Use module.hot here
}
```

Minifiers are able to remove the if statement for production, though none of the common minifiers for Meteor currently do.

### How it works

1. The Meteor tool has a new runner - the HMR server. In development this is enabled automatically if the app or any of its packages uses `hot-module-replacement`. The HMR server provides a websocket server (on the same port as the proxy) that apps can connect to.
2. Compilers can detect if the hot api will be available for a file by calling `file.hmrAvailable()`. Even if it returns true, the file might not be updatable with HMR since all of the information needed isn't available at this step in the build process, but it does guarantee `module.hot` is set. If a compiler changes its output depending on if HMR is available, it should use it in its  cache key.
3. After compilation and the import scanner and before linking the bundler calls a function in the HMR server which compares the in-progress build with the previous build to check if all changes can be updated through HMR and preparing the data needed for the update.
4. It immediately sends the changes to any clients even though the build hasn't finished (this is called an eager update). If this fails, the client will handle the failure after the build finishes
5. Once the build finishes and the server is restarted or autoupdate creates the new versions, the client will request any changes since the last non-eager update and handle any failures from step 4.

Updates are applied with this process:

1. It applies the changes to the modules. Added modules it installs, and changed modules it updates the content of and removes cached exports.
2. It checks if the modules are replaceable. It kept track of which modules imported which modules, and using this it makes sure all of the modules it was imported by accept updates, or that all of the modules those modules were imported by accept updates, and so on until either all paths reach a module that accepts updates, it reaches the main module ,or it reaches a module that declines updates.
3. For all of the modules between the ones that were modified up to the ones that accept updates it removes cached exports and marks them to be evaluated the next time they are required
4. It then requires all of the modules that accepted the update, causing them and the modules from step 3 to be re-evaluated

### Remaining tasks

There are some larger details that should probably be discussed:

**Hot API** So far I have been copying webpack's api so I can focus on other aspects. Many other bundles have implemented their own api's, and to me some of them seem more intuitive. One of the next steps is to do more research and experiment with this, and I would appreciate any input from the community.

~~**Integration with the meteor tool** It currently uses an event emitter, which is a temporary solution. Originally I had considered creating an API for build plugins to handle HMR, but this part of HMR seems straight forward enough I am not sure if there is a reason for there to be more than one implementation. We also need to consider how packages will enable HMR for themselves, and for compilers to know if HMR is available for a unibuild. Allowing build plugins for HMR might help with both.~~

~~**Communication with app** Currently it uses a websocket server on a separate port. The Meteor tool already has an http server for the proxy, which could be extended to include a websocket server. A related feature request is at https://github.com/meteor/meteor-feature-requests/issues/354. I think this can wait until later since we should probably review all of the auto reload packages at the same time. With this PR we will have 4, and there are other PR's open adding an additional package.~~

**How much should be enabled by default** Is there a reason to use `hot-module-reload` without React Fast Refresh?

Before beta:
- [x] Exclude HMR client code from production builds
- [x] Reload page when module is removed from client
- [x] Add api for compilers to check if HMR is enabled and if a file is replaceable
- [x] Updated modules should use same sourceURL as the module they replace
- [x] Websocket should handle connection errors or disconnects
- [x] Clear dispose handlers and _hotAccepts when a module is updated, though it should preserve _hotAccepts if the updated module fails
- [x] Limit number of change sets stored in memory
- [x] When adding files, group by meteorInstallOptions
- [x] Compare all relevant details of files instead of only hash when checking for changes
- [x] hot-module-replacement should not be a build plugin. All of that functionality should be moved into the meteor tool.
  - [x] Enable HMR when the app directly uses hot-module-replacement
  - [x] This will allow hot-module-replacement to be a dev only package so it is excluded from production builds
  - [x] Add a websocket server to the proxy to send HMR updates to client. This will allow us to use the same port as the Meteor app.

Before stable:
- [x] Move import tracking from install to hot-module-replacement package
- [x] Clear old entries in importedBy whenever a module is updated
- [x] It shouldn't disable linker cache
- [x] Make React Fast Refresh optional
- [x] Require client to send secret before sending changes over websocket
- [x] Decide on and implement hot api specifically with Meteor bundler in mind
- [x] Improve logging in client during HMR update
- [ ] Write Tests
- [ ] Documentation
- [x] Make sure all details in https://github.com/facebook/react/issues/16604#issuecomment-528663101 are correctly handled, mainly the `Nuances` topic
- [x] React Fast Refresh should debounce updates (it currently rerenders for each component that was modified)
- [x] Handle changes outside the js and css bundles that require a full reload. These are normally handled by HCP, but HMR blocks it.
- [x] Check how core packages handle HMR
- [x] Support react components in typescript files

Later:
  - [ ] Create HMR api to add listeners for update
  - [ ] Remove use of `install`'s root object.
  - [ ] Submit PR to `install` to add necessary api's for HMR
  - [ ] Submit PR to reify to allow removing cache entries
  - [ ] Create simple package to show on screen when page updated (if a change didn't do what you expect or isn't a visible change, you might not realize when it applied the changes)
  - [ ] Enable for cordova apps
  - [ ] Look into including a timestamp for the build in bundle (without it, there might be a race condition)
  - [ ] Enable for legacy bundle
  - [ ] Enable for packages
  - [ ] Look into ways this could help with faster server restarts
  - [ ] Show reason when we do a full reload because the HMR was not possible
  - [ ] Use meteorInstallOptions to control which modules have access to module.hot. Currently HMR is only used for apps or packages that directly depend on `hot-module-replacement`, but `module.hot` is available for everything.
  - [ ] Remove `if (module.hot) { ... }` when minifying (requires #11107)


filipenevola:
Amazing work @zodern! Thank you!

Watch a short demo of this already working in one of my Meteor apps https://youtu.be/riroCNt9I98

I also have tested already in another one that is bigger and almost 3 years old and it is working there as well 😉 

If you can please start using it and provide feedback here. I'm working closely with Zodern to get this released as soon as it is ready.

arggh:
This is some great stuff. Though I have to say, reloads on the latest Meteor are already pretty damn fast 🤷‍♂️

@zodern I'd love to help test this, but in order to make it easier for people who test, could you bring your `zodern:svelte` up to date with `svelte:compiler` latest `4.0.0-beta.1` release? Your fork has a hardcoded dependency of Svelte `3.21.0`, which in Svelte's case, is ancient!

pmogollons:
Awesome. I just started testing it in a large react project and it works perfectly without any config or change.

Ill start using this for our dev env and give some feedback. 

Thanks @zodern

evolross:
HMR newbie here. Question - is this a development-only tool? How would this work in production on Galaxy? You redeploy your app and it doesn't need to refresh if HMR can handle all the changes/updates?

pmogollons:
AFAIK is only a dev tool. It updates the components in place without reloading all the page. Shouldnt do anything on production.

aliogaili:
I've tested it and it seems to be working perfectly! great work @zodern 

My only wish is that for the HMR to work also within packages. 

Screen Shot 2020-07-18 at 7 37 44 AM

zodern:
>  is this a development-only tool? How would this work in production on Galaxy? You redeploy your app and it doesn't need to refresh if HMR can handle all the changes/updates?

@evolross it is intended for development only. It would be possible to implement this for production, but there are too many ways it could fail. In development you can notice those situations and refresh the page easier and with less disruption than it would be for a user.

---

I've renamed the feature to Hot Module Replacement instead of Hot Module Reload since it is more accurate and matches other bundlers.

There have been a number of fixes the past month. If you are using svelte, I now recommend using [`zodern:melte`](https://atmospherejs.com/zodern/melte) instead of `zodern:svelte` since it has had many improvements related to HMR.

**Hot API** / **Support require/module.exports, or only ecmascript modules**

I've decided to continue using `module.hot`. Once we have better support for ecmascript modules, we can look into other options such as [esm-hmr](https://github.com/pikapkg/esm-hmr). Since they both are at different locations (`import.meta.hot` and `module.hot`) it wouldn't have to be a breaking change, and since native ecmascript modules isn't on the road map this will probably not happen soon.

**Integration with the meteor tool**
I am planning to remove the build plugin from `hot-module-replacement` and move the functionality into the meteor tool. This will simplify the integration into the build process and allow the `hot-module-replacement` package to be a dev only package.

I've been considering how HMR should behave when a package depends on `hot-module-replacement` vs an app depending on it. This is what I am considering:
`hot-module-replacement` will become a weak dependency of the `modules` package, so it will be loaded before any package with modules. If any package depends on `hot-module-replacement`, the `module.hot` api will be available to all packages and the app in development. However, HMR will only be used to update the client if the app directly depends on `hot-module-replacement`.

There is not a clear way to only provide `module.hot` for packages that directly depend on `hot-module-replacement` since they all use the same Module constructor. This also makes `hot-module-replacement` one of the few packages that is treated specially by the Meteor tool, which I wanted to avoid since it makes it challenging to create packages that provide an alternate implementation.

**How much should be enabled by default**
To enable React Fast Refresh I will change it to require adding a `react-fast-refresh` package to your app. This will give more flexibility to apps and reduce overhead or potential issues with apps not using React or using an old version. One example of the extra flexibility is you can remount your react tree instead of trying to save react state (many apps do this with Webpack since it is reliable and simple).

---

I would appreciate any feedback from the community. There are many options with different trade offs, so it would be helpful to know what people care about.

hexsprite:
Trying HMR and I change tsx/jsx files. The React component is not changing and I don't get regular reload behavior. Any tips on how to troubleshoot ? React 16.13.1

hexsprite:
Noticed this traceback in web console

```
Exception in flushing DDP buffered writes: DOMException: Failed to execute 'send' on 'WebSocket': Still in CONNECTING state.
    at send (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:136:12)
    at requestChanges (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:185:3)
    at http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:422:7
    at http://localhost:3000/packages/autoupdate.js?hash=e0af7471338a4d9cb024a28b97977648652d6f6f:289:9
    at Set.forEach (<anonymous>)
    at ClientVersions.set (http://localhost:3000/packages/autoupdate.js?hash=e0af7471338a4d9cb024a28b97977648652d6f6f:282:26)
    at Object.update (http://localhost:3000/packages/autoupdate.js?hash=e0af7471338a4d9cb024a28b97977648652d6f6f:249:16)
    at Object.store.<computed> [as update] (http://localhost:3000/packages/ddp-client.js?hash=5333e09ab08c9651b0cc016f95813ab4ce075f37:604:38)
    at http://localhost:3000/packages/ddp-client.js?hash=5333e09ab08c9651b0cc016f95813ab4ce075f37:1488:19
```

pmogollons:
I start using it today on windows and im geeting a weird behaviour. Sometimes the first update the component is hmr'ed but the next times it reloads the page and after that it replaces the component. Very weird thing, im not even sure how to debug it. Im using beta.2.

What I think is happening is that hmr replaces the updated component but when the compiling is done it reloads the page as usual. And now in the console I saw that something was not able to be updated with hmr, but thats weird as all changes where already done by hmr.

```javascript
HMR: replacing module: /imports/ui/common/pages/Login.js
VM15435 hot-module-replacement.js:239 HMR: rerunning /imports/ui/common/pages/Login.js
VM15435 hot-module-replacement.js:397 HMR: finished updating
VM15435 hot-module-replacement.js:417 HMR: Some changes can not be applied with HMR. Using hot code push.
```

@zodern let me know if I could do something specific to help debug this.

jbbr:
> Sometimes the first update the component is hmr'ed but the next times it reloads the page and after that it replaces the component.

I do experience the same issue when updating react components which are imported using a dynamic `import` and react lazy.
```ts
const HmrFailCmp = React.lazy(() => import('/imports/../UpdatingThisCmpIsCausingIssues.tsx')
```

Apart from that:
Thank you very much for this feature! This is what I missed the most on Meteor and caused me to try various semi-good workarounds like meteor-webpack to be able to hot reload react components.
I'm sure many developers will appreciate this.

pmogollons:
I think what I reported is a regression, im using this ```481f8b475``` by checkout in my mac and it works flawless. After updating to meteor 2-beta.2 I started to get this behaviour.

zodern:
The latest Meteor 2 beta is missing some changes to the dynamic-import package, which is causing hot code push to be used when modifying dynamic imports. The next beta should work correctly.

@hexsprite did you add the `hot-module-replacement` package while the app was running? It should log a message in the server that you need to restart Meteor to enable HMR. I can add a client message and have it only use hot code push in this situation.

Typescript files currently don't use the react fast refresh babel plugin, so I am not sure how well it will work at the moment. Fixing this will require moving it into a separate package so both the ecmascript and typescript compilers can use it. The two options are:
1. Make react fast refresh opt in by adding a `react-refresh` package to your app. 
2. Always include the react-refresh package (maybe make it a dependency of `hot-module-replacement`), but only enable react fast refresh if the app has the `react` npm package installed. Parcel does this, but there are situations where an app uses React, but React isn't a direct dependency. We could provide an env var to force it to be enabled/disabled.

I like the first option for the simplicity, but it adds an extra step for react apps.

rijk:
It's working great for me with `.tsx` files.

hexsprite:
@zodern, OK, tried restarting it and I'm getting:

```
HMR: replacing module: /client/settings/AccountSettingsComponents.tsx
hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:239 HMR: rerunning /client/settings/AccountSettingsComponents.tsx
hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:209 Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at createInlineSourceMap (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:209:65)
    at http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:220:14
    at fileEvaluate (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:394:7)
    at Module.require (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:266:27)
    at require (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:313:21)
    at rerunFile (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:241:3)
    at http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:395:5
    at applyChangeset (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:390:21)
    at http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:99:21
```

But since that's a TypeScript file I assume that's because it's not supported yet?

kratam:
I have a similar error in a non-ts project: `Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.`
<details><summary>Error log</summary>
<pre>
09:28:54.312 hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:226 HMR: replacing module: /ui/calendar/components/CalendarCell.js
09:28:54.314 hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:239 HMR: rerunning /ui/calendar/components/Calendar.js
09:28:54.324 hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:209 Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at createInlineSourceMap (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:209:65)
    at http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:220:14
    at fileEvaluate (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:394:7)
    at Module.require (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:266:27)
    at Module.moduleLink (http://localhost:3000/packages/modules.js?hash=9c69b553d30f990389257461544af0be678e373b:348:22)
    at Module.module.constructor.link (http://localhost:3000/packages/hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:249:18)
    at module (/ui/calendar/components/Calendar.js:98:8)
    at http://localhost:3000/packages/dynamic-import.js?hash=9ea137ab5f9d8d7f8b1909fff148644bf481ae4d:143:7
    at fileEvaluate (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:394:7)
    at Module.require (http://localhost:3000/packages/modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:266:27)
createInlineSourceMap @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:209
(anonymous) @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:220
fileEvaluate @ modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:394
require @ modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:266
moduleLink @ modules.js?hash=9c69b553d30f990389257461544af0be678e373b:348
module.constructor.link @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:249
module @ /ui/calendar/components/Calendar.js:98
(anonymous) @ dynamic-import.js?hash=9ea137ab5f9d8d7f8b1909fff148644bf481ae4d:143
fileEvaluate @ modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:394
require @ modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:266
require @ modules-runtime-hot.js?hash=d62f7d8ae3407df88cd3e13a9d7d41f65e219d44:313
rerunFile @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:241
(anonymous) @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:395
applyChangeset @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:390
(anonymous) @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:99
handleMessage @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:98
(anonymous) @ hot-module-replacement.js?hash=01a29eb5e6be76f94c88d5a09bd2009d4fac1767:176
</pre></details>

hexsprite:
Sounds like a potential solution to this issue is: https://base64tool.com/uncaught-domexception-btoa-on-window/

hexsprite:
I located the UTF-8 characters in the source and replaced them. HMR works great with the TSX file now.

rijk:
I've noticed updates to LESS files now trigger a full client refresh. This is interesting, as before HMR this actually worked _without_ a refresh if I recall (style-only changes). I'm also seeing the changes I made appear on the screen before the refresh  is triggered. Any way we can avoid that client refresh for style changes?

pmogollons:
I can confirm that now is working correctly again with beta.3. Thanks

rijk:
There's also an error when adding/removing hooks from a component:

image

Don't know if it's possible to detect this, but I think NextJS doesn't have this problem so maybe it could be interesting to look how they approach it.

nick-gudumac:
I've tried beta.3, and it seems to fail with dynamic imports. React components are written in TypeScript.

image

rijk:
While it was working perfectly before, suddenly it stopped working now. I still get these notices in the console on every save:

```
HMR: replacing module: /imports/ui/layouts/Board2.tsx
HMR: rerunning /imports/ui/layouts/Board2.tsx
HMR: finished updating
```

and no errors, but the screen is not reflecting any of the changes. I tried changing text, properties, or adding/remove components altogether. Let me know what I can do to debug!

rijk:
I found out what broke it by going back in my commit history. It was an additional export I added to my component:

```tsx
import React from 'react'

// This breaks HMR
export const sidebarTransition = {
  type: 'spring',
  stiffness: 400,
  damping: 45,
}

export default function Board2() {
  // ...
}
```

pmogollons:
I updated a couple of days ago to beta.4 and im getting huge degradation of hmr speed, after downgrading to beta.3 everything works fast again. Not sure what is causing this, let me know what can I do to help find the issue.

schlaegerz:
I am seeing the same thing as @pmogollons on beta.3 it takes around 1-2 seconds to update, on beta.4 it takes like 20 seconds or more.

filipenevola:
Hi, we are aware of this issue, we are going to review it next week.

afrokick:
React dev tools doesn't work for me in 2.0beta6 with HMR.

Is it relates to https://github.com/facebook/react/issues/20377 ?

zodern:
@afrokick did it work with earlier betas? Are you using the React dev tools from a browser extension, or the [npm package](https://www.npmjs.com/package/react-devtools)? The linked issue only occurs when using the npm package.

afrokick:
> @afrokick did it work with earlier betas? Are you using the React dev tools from a browser extension, or the [npm package](https://www.npmjs.com/package/react-devtools)? The linked issue only occurs when using the npm package.

I tested with extension and only on beta6, because beta4 didn't work for me at all.

beta6 without hot-module-replacement works fine.

beta6 with hot-module-replacement breaks devtools.

jbbr:
> beta6 with hot-module-replacement breaks the react dev tools. (missing components tree in the Components tab, highlights updates, etc.)

I cannot confirm that.
At first glance, everything is working fine for me. Updated from 1.12.1 to 2.0beta6.
Feels much more stable then previous 2.0 preview releases. Thank you!

pmogollons:
Just updated to beta.6 and it feels faster and it works more reliably than beta.3

Also tested the react dev tools and is working for me in Chrome on linux. Ill try later on my Mac and come back.

afrokick:
It was React 16.8.6. After upgraded to 16.9.0 the issue was gone. I'm sorry for the inattention.

zodern:
@afrokick The next beta will disable react fast refresh if it detects you are using an older version of React, which should fix the issue with the dev tools. If react fast refresh is disabled, it won't automatically update react components with HMR. It requires React 16.9 or newer.

Apps that use an older version of React can use this in the file that mounts the root component:

```js
if (module.hot) {
  module.hot.accept();
}
```
When the file re-runs, `ReactDOM.render` will detect the old react tree, unmount it, and render a new react tree using the updated components. It won't preserve component state, but before react fast refresh many apps used this for HMR since it works reliably.

filipenevola:
Meteor 2.0-rc.0 is published.

You can use it now updating your app:
```shell
meteor update --release 2.0-rc.0
```

Or you can create a new app:
```shell
meteor create my-app --react --release 2.0-rc.0
```

Please provide feedback here.

jbbr:
When upgrading from `2.0-beta.6` to `2.0-rc.0` I'm experiencing some package downgrades and constraints issues.

e.g:
```
error: Conflict: Constraint logging@1.1.17 is not  satisfied by logging 2.0.0-rc200.0
...
* logging@2.0.0-rc200.0 <- webapp 1.10.0-rc200.0 <-  meteor-base 1.4.0
* logging@1.1.17 <- percolate:migrations 1.0.2
```

The packages which cause these issue do not have a strict constraint for `logging`.
They just have something like `api.use('logging')`

Workaround: Add strict version constraint for 2.0 rc, e.g `api.use('logging@2.0.0-rc200.0')`

Not sure if this is incorrect behavior or expected behavior based on the version constraints/version locks of my third party packages. Just wanted to mention - maybe this helps others who might also experience those issues.

pmogollons:
Im having an issue where hmr is failing to reload a component with this error on the browser console.

```javascript
hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:348 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at checkModuleAcceptsUpdate (hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:348)
    at hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:361
    at Set.forEach (<anonymous>)
    at checkModuleAcceptsUpdate (hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:348)
    at hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:471
    at applyChangeset (hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:464)
    at hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:183
    at Array.every (<anonymous>)
    at handleMessage (hot-module-replacement.js?hash=b99aaee973e26637b168141edcb2337028f027a8:182)
```

rijk:
I have the same issue, after updating a hook (not a component).

nicubarbaros:
It still fails to import lazy imports in React. 
Screen Shot 2021-01-19 at 14 44 47

`hot-module-replacement@0.2.0-rc200.`
`react` 17.0.1

nicubarbaros:
😢

zodern:
@nicubarbaros could you please create a reproduction? React fast refresh accesses all of the exports, and it looks like one of the exports is a reference to an iframe or something else that React Fast Refresh needs to be more careful with.

filipenevola:
> 😢

Don't worry, we are going to continue working on HMR, fixing bugs and improving it as a whole.

The merge means that we are ready to have it in a new Meteor final release.

rijk:
Further debugging this error:

> Uncaught TypeError: Cannot read property 'forEach' of undefined

I found out the `moduleId` when it happens is `/node_modules/meteor/hot-module-replacement/client.js`. So presumably this is fixed with 59754c471cf90cc9d30a48ff5fb819d90f443915.

filipenevola:
> I found out the `moduleId` when it happens is `/node_modules/meteor/hot-module-replacement/client.js`. So presumably this is fixed with [59754c4](https://github.com/meteor/meteor/commit/59754c471cf90cc9d30a48ff5fb819d90f443915).

Hi @rijk, did you try Meteor 2.0-rc.3? This commit should be included there.

rijk:
Sorry, had no idea we were at 3 already 😄

filipenevola:
Always moving!

nicubarbaros:
I tracked it down to these fellows that are creating the problem mentioned above. 

[Video.js](https://github.com/videojs/video.js) & [emoji-mart](github.com/missive/emoji-mart). 

I will try to create a reproduction.

pmogollons:
I can confirm that the issue I reported previously is now fixed in final release.

rijk:
@filipenevola  Is this still the right place to report issues with HMR?

filipenevola:
Hi @rijk no. It's not.

You should open new issues then we can track them individually.

Please provide reproductions, it's very hard to reproduce HMR problems when we don't have the code as modules can interact in different ways. Of course, if a reproduction is hard to be provided explain why 😉 

Thanks!

PR #11118, opened 2020-07-10
[Feature] Be able to define default user fields published on login

menelike:
##### This removes the limitation described in https://docs.meteor.com/api/accounts.html

> By default, the current user’s username, emails and profile are published to the client. You can publish additional fields for the current user with:

```JS
// Server
Meteor.publish('userData', function () {
  if (this.userId) {
    return Meteor.users.find({ _id: this.userId }, {
      fields: { other: 1, things: 1 }
    });
  } else {
    this.ready();
  }
});

// Client
Meteor.subscribe('userData');
```

##### How to be used

```JS
Accounts.setDefaultPublishFields({username: 1, profile: 1, emails: 1, 'foo.bar': 1});
// or
Accounts.setDefaultPublishFields({'do.not.publish.this': -1});
```

**Since this replaces the default projection the user has to add `profile`, `username`, and `profile` to the projection to keep the old behavior.**

##### Motivation

I think that writing an additional subscription just for the sole purpose to fetch additional fields, is cumbersome and degrades the performance for no reason.

##### Background

In our scenario, a logged-in user has additional fields that are not part of `profile`. Those fields are needed to determine the state and the permissions of the user. In the UI, after the login has succeeded, the user will be redirected to the primary route for logged-in users - this then checks for the appropriate permission. At that point, the custom subscription used to fetch all the user data has not finished yet. As a result, the user will be redirected back to the "you are not logged in" page.

**tl;dr**
It helps with race conditions where fields are missing after login.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -37,6 +37,14 @@ export class AccountsServer extends AccountsCommon {
       loggedInUser: ['profile', 'username', 'emails'],
       otherUsers: ['profile', 'username']
     };
+
+    // use object to keep the reference when used in functions
+    // where _defaultPublishFields is destructured into lexical scope
+    // for publish callbacks that need `this`
+    this._defaultPublishFields = {
+      loggedInUser: ['profile', 'username', 'emails'],
+    };
+
     this._initServerPublications();
 
     // connectionId -> {connection, loginToken}
@@ -678,7 +686,7 @@ export class AccountsServer extends AccountsCommon {
 
   _initServerPublications() {
     // Bring into lexical scope for publish callbacks that need `this`
-    const { users, _autopublishFields } = this;
+    const { users, _autopublishFields, _defaultPublishFields } = this;
 
     // Publish all login service configuration fields other than secret.
     this._server.publish("meteor.loginServiceConfiguration", () => {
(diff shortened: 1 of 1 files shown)


packages/accounts-base/accounts_server.js
```diff
@@ -37,6 +37,14 @@ export class AccountsServer extends AccountsCommon {
       loggedInUser: ['profile', 'username', 'emails'],
       otherUsers: ['profile', 'username']
     };
+
+    // use object to keep the reference when used in functions
+    // where _defaultPublishFields is destructured into lexical scope
+    // for publish callbacks that need `this`
+    this._defaultPublishFields = {
+      loggedInUser: ['profile', 'username', 'emails'],
```
filipenevola, line 45: I would prefer to use the MongoDB standard here, using a map with fieldName: 1 to select fields.

packages/accounts-base/accounts_server.js
```diff
@@ -678,39 +686,40 @@ export class AccountsServer extends AccountsCommon {
...
-            username: 1,
-            emails: 1
-          }
-        });
-      } else {
-        return null;
-      }
-    }, /*suppress autopublish warning*/{is_auto: true});
+    // ['profile', 'username'] -> {profile: 1, username: 1}
+    const toFieldSelector = fields => fields.reduce((prev, field) => (
```
filipenevola, line 698: we don't need this, the API should expect the field projection in the format MongoDB expects so we don't need to convert and we also can support other cases, like restricting fields.

menelike:
@filipenevola Hope it fits now. I think that replacing the default projection makes more sense than extending it (which would have to deal with mixed projections). WDYT?

filipenevola:
@menelike I don't know why CircleCI is trying to run the tests in your account and not at Meteor org. Could you create a new branch and a new PR? Sometimes CircleCI get lost :(

afrokick:
@menelike

menelike:
@afrokick @filipenevola 

I was able to work around the issues with circleCI (bugs on their side),  all tests have passed, we should be good to go.

filipenevola:
Cherry-picked to 1.12

santiagopuentep:
@filipenevola I'm a little confused. What's the difference between Accounts.setDefaultPublishFields and Accounts.config.defaultFieldsSelector? They seems to do the same thing.

filipenevola:
Hi @santiagopuentep, it's used specific for the publication and not for the finds as a whole.

See the diff below, in the previous code Meteor was always publishing 3 specific fields:
accounts_server js (:Users:filipe:Documents:meteor:ws:meteor:packages:accounts-base) 2020-12-10 17-13-17

mullojo:
Hi @menelike 

Is this the correct syntax for how to use this update?

```js

Accounts.setDefaultPublishFields({username: 1, profile: 1, emails: 1, 'foo.bar': 1});
// or
Accounts.setDefaultPublishFields({'do.not.publish.this': -1});

```

It would be good to put these updates in the accounts docs: https://docs.meteor.com/api/accounts.html

Also, I'm uncertain if the defaults for the `accounts-base` package are still the same or if the user needs to specify them as this thread implies.   It would be really great if there was just a bit of clarification added and it would be great to see updates to the Meteor docs.

@filipenevola if you know how these updates work, it would be great if you could share them too.   In the Meteor v1.12 update, this is not specified a a breaking change, but it would be if the previous defaults have changes.

menelike:
Hi @mullojo 

> Is this the correct syntax for how to use this update?

Yes.

> It would be good to put these updates in the accounts docs: https://docs.meteor.com/api/accounts.html

Absolutely!

> Would this update change the behavior of calling data from Meteor.user() on the client by default? I found this thread on the Meteor Forums where a user had an issue with Meteor.user() after upgrading to v1.12: https://forums.meteor.com/t/meteor-user-returns-undefined-after-updating-to-1-12/54843/16

I've written a comment in the linked thread. Generally speaking `Accounts.setDefaultPublishFields();` is an optional call, if not used the default fields are published which makes this backward compatible. I don't see any reasons for a breaking change. But keep in mind that if you use `Accounts.setDefaultPublishFields();` you need to pass all fields you want to be published to the client as the default fields are then omitted.

rj-david:
We had a lot of handling added in our app to manage that race condition that the other fields are not yet available for the UI to check. Thanks for this, @menelike.

PR #11119, opened 2020-07-12
Add Apollo skeleton

StorytellerCZ:
Added Apollo skeleton based on the React skeleton and updated dependencies on all skeletons.
It is a very simple Apollo integration without anything fancy using the official `apollo` package.
Maybe in the future it might be a good idea to have more diversity in Apollo skeletons, but I think this is a much needed start of that process.


tools/static-assets/skel-apollo/imports/ui/Info.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { useQuery } from '@apollo/react-hooks';
```
filipenevola, line 2: I believe you need to change this as well

```suggestion
import { useQuery } from '@apollo/client';
```

StorytellerCZ:
@filipenevola I have written an article on my thinking on Dev.to: https://dev.to/storytellercz/meteor-create-apollo-32e2
In short I see this as the basic HTTP GraphQL skeleton that most developers are familiar with (plus React). I think additional skeletons around Apollo should be coming in next version. I see something like `apollo-ddp`, to integrate the benefits of DDP, as the one with most interest from Meteor community. This skeleton though is targeted at new comers.

StorytellerCZ:
I'll make another round of updating the skeletons dependencies tomorrow.

filipenevola:
Thanks @StorytellerCZ

PR #11139, opened 2020-08-07
Bumping version to update README.md on atmosphere

Sornii:
Bumping version to update README.md on atmosphere


klaussner:
I don't know how the publishing process works for core packages but READMEs of third-party packages can be updated without publishing a new package version.

filipenevola:
@klaussner you are right.

Thanks @Sornii, it's updated now.

PR #11151, opened 2020-08-27
Initial support for publication strategies in livedata server

sebakerckhof:
Rationale:
Meteor optimizes bandwidth over memory usage in the livedata server by only sending deltas on collection changes over multiple publications. While I believe it is a sane default behavior, it has been a long outstanding request ( https://github.com/meteor/meteor-feature-requests/issues/79 ) to be able to customize this behavior.
This would have 2 clear benefits / use-case
1. Reduce memory footprint of the server, especially if you publish lots of data but without overlapping publications
2. It would allow for new kinds of collections/publications. E.g. one can imagine a volatile message queue (send & forget) backed by rabbitmq or redis. In this case a session collection view has no purpose. Up till now you needed to make your own protocol over the websocket with a package like `meteor-custom-protocol`, `streamy` and the likes. By doing so you lose a lot of useful features of the livedata server like rate limitation, message batching etc.

The contents of this PR lays the foundation for the above, but is not yet sufficient for use-case 1 (client-side mergebox). I already submit it as-is since it already allows for use case 2 (e.g. volatile message queues).

If this gets accepted I could continue working towards use case 1, which, in order to maintain the same semantics, would imply sending subscription Id's on added/removed/changed messages and having the SessionCollectionView/SessionDocumentView classes in the client bundle. I would also need to think more about a strategy for when the userId changes, but it's doable.

Note that it would be possible to set a strategy on a per-collection basis.


KoenLav:
Likely to cause conflicts with:

https://github.com/meteor/meteor/pull/10478

(Not that this is ever going to get merged, but hey.)

sebakerckhof:
@KoenLav I had a look at that PR and I don't think it would conflict.
Of course, the target of this PR is to minimize memory usage for some use cases, which kind of clashes with the concept of the buffers. But looking at your code it wouldn't be to hard to set it on a per-collection level for the oplog buffering.

I also gave some feedback on that PR.

vlasky:
@sebakerckhof thanks for drawing my attention to your enhancement! I am testing it right now in one of my Meteor projects with the default publication strategy set to NO_MERGE, with an exception for the "users" collection. It looks very promising.

On the client side, I am getting an uncaught Error thrown by update(msg) in collections.js:

```
collection.js:198 Uncaught Error: Expected not to find a document already present for an add
    at Object.update (collection.js:198)
    at Object.store.<computed> [as update] (livedata_connection.js:310)
    at livedata_connection.js:1192
    at livedata_connection.js:1191
    at Connection._performWrites (livedata_connection.js:1187)
    at Connection._flushBufferedWrites (livedata_connection.js:1167)
    at Connection._livedata_data (livedata_connection.js:1133)
    at Connection.onMessage (livedata_connection.js:1663)
```

This is expected behaviour, caused by a publication re-adding everything after setUserId() has run.

Resolving this requires the mongo package to be forked or by applying a client-side patch similar to [the one that mitar created for his peerlibrary:control-mergebox package](https://github.com/peerlibrary/meteor-control-mergebox/blob/master/client.coffee).

Do you have any plans to implement one of these approaches?

vlasky:
@sebakerckhof I was impatient, so I forked the [mongo package](https://github.com/meteor/meteor/tree/devel/packages/mongo), converted [Mitar's client-side code](https://github.com/peerlibrary/meteor-control-mergebox/blob/master/client.coffee) from CoffeeScript to JavaScript, then extracted the relevant snippet and inserted it into collection.js.

That is, right after [these lines in collection.js](https://github.com/meteor/meteor/blob/devel/packages/mongo/collection.js#L177)
```

      update(msg) {
        var mongoId = MongoID.idParse(msg.id);
        var doc = self._collection._docs.get(mongoId);

```

I added the following lines:

```
    if (Meteor.isClient) {
      if (msg.msg === 'added' && doc) {
        msg.msg = 'changed';
      } else if (msg.msg === 'removed' && !doc) {
        return;
      } else if (msg.msg === 'changed' && !doc) {
        msg.msg = 'added';
        _ref = msg.fields;
        for (field in _ref) {
          value = _ref[field];
          if (value === void 0) {
            delete msg.fields[field];
          }
        }
      }
    }
```
The Meteor.isClient check was a safety measure I added. So far, this appears to have resolved the Error exceptions on the client side.

Thoughts?

sebakerckhof:
@vlasky The changes in this PR are really for volatile messages. E.g. I use them in my product to implement send-and-forget message queues. For mongo you would indeed run into errors like the one you describe.
While the change you propose might fix this specific error, there are still other errors that can happen with the changed behavior. I think the Mongo collection is the wrong place to tackle these. I think what really needs to be done is split of the mergebox so it could also be used in the frontend. I have a PoC where I refactored and modernized the DDP code and where the mergebox is being used on the client side. That way the mongo collection behavior stays the same.

vlasky:
@sebakerckhof your arguments are valid. The client-side mergebox sounds useful.

Subscribing to a server-side publication to populate a client-side collection remains the most common use case for Meteor publications.

At this point in time we are merely seeking an interim solution to keep things working and avoid server-side memory exhaustion. 

That is, something that is good enough until the robust solution to mergebox scalability is implemented in Meteor. People have been waiting for a long time. Regrettably, some people have abandoned the use of this fantastic piece of Meteor functionality due to the mergebox scalability issue. I hope Tiny is following this discussion.

So far, your server-side change to livedata server combined with my mongo modification using some of @mitar's code seems to be the closest thing we've got.

packages/ddp-server/livedata_server.js
```diff
@@ -2,6 +2,33 @@ DDPServer = {};
 
 var Fiber = Npm.require('fibers');
 
+// Publication strategies define how we threat data from published cursors on collection-level
+// This allows someone to:
+// - Make a trade-off between bandwidth and memory use
+// - Have special (non-mongo) collections like volatile message queues
+const publicationStrategies = {
```
filipenevola, line 9: I would recommend a declarative approach here instead of imperative, adding fields like

```javascript

const publicationStrategies = {
  NO_MERGE_NO_HISTORY: {
    name: 'NO_MERGE_NO_HISTORY',
    useCollectionView: false,
    doAccountingForCollection: false,
  },
  NO_MERGE: {
    name: 'NO_MERGE',
    useCollectionView: false,
    doAccountingForCollection: true,
  },
  SERVER_MERGE: {
    name: 'SERVER_MERGE',
    useCollectionView: true,
    doAccountingForCollection: true,
  }
}
```

So in the future it would be easier to maintain this code without searching for if/elses using these types for specific decisions, also when adding new types the decisions will be clear for each one.

So these functions wouldn't be necessary:
`_useCollectionView`, `_doAccountingForCollection`

packages/ddp-server/livedata_server.js
```diff
@@ -2,6 +2,33 @@ DDPServer = {};
...
+  // But it will remember the id's it has send to the client
+  // so it can remove them when a subscription stops
+  // It can be used when a collection is only used in a single publication
+  NO_MERGE: 'no_merge',
+  // XXX: We could implement client side merge
+  // By moving the session collection view to the client
+  // And sending subscription id's along the added/changed/removed messages
+  // It can be used when we want to save server memory but still have the same
+  // semantics over multiple publications of one collection
+  // CLIENT_MERGE: 'client_merge',
```
filipenevola, line 25: Do we want to include this in this PR? Otherwise I would remove from the code and just open an issue for this use-case.

filipenevola (changes requested):
I like this approach and I have experience in many Meteor projects that could take advantage of this approach.

I'm suggesting a code change in favor of maintenance.

filipenevola:
Hi @vlasky, yes, we are following the discussion.

We have experience with many Meteor projects using similar solutions, even your (and Mitar) workaround to avoid collections errors in the client I already saw in at least other two different solutions for the same problem: do less work in the server.

About this issue when a document is already in Minimongo but is received again, what do you think would be the best solution here? Maybe in many cases even the client-side mergebox is too much, we could just ignore the error if a specific setting is applied in the collection (I know production apps using this approach btw).

[meteor-fast-methods](https://github.com/pathable/meteor-fast-methods) is also an approach that can help in many cases, specially when you have specific publications that are too heavy in the server.

vlasky:
@filipenevola I have created my own Meteor branch [publicationstrategies](https://github.com/vlasky/meteor/tree/publicationstrategies) integrating @sebakerckhof's code with extracts from the client-side code in @mitar's package [peerlibrary:control-mergebox](https://github.com/peerlibrary/meteor-control-mergebox)

The last commit is https://github.com/vlasky/meteor/commit/0299ce6865b9800374b179fa1f5f8333b91e2132.

It now provides client-side MongoDB compatibility when the publication strategy is set to NO_MERGE, by gracefully handling duplicate events without throwing an exception. Specific changes:

*When we receive an added message for a document that is already present in the client's collection, it will be changed
*When we receive a change message for a document that is not in the client's collection, it will be added
*When we receive a removed message for a document that is not in the client's collection, nothing will happen.

packages/ddp-server/livedata_server.js
sebakerckhof: I agree that's better. I'll try to rework this PR and maybe I can continue on my modernization PR

linegel:
@vlasky is there any chance you could open PR into Meteor repo?

filipenevola:
Hi @vlasky, I believe a PR with your changes would be great.

@sebakerckhof I imagine you are busy with other things, but do you think you are going to be able to proceed with my change recommendation? Otherwise I could do that or @vlasky in a new PR.

vlasky:
@linegel @filipenevola done.

https://github.com/meteor/meteor/pull/11368

StorytellerCZ:
If I'm not mistaken this has been superseded by #11368, if I'm not correct feel free to re-open.

packages/ddp-server/livedata_server.js
vlasky: @filipenevola I have removed this.

packages/ddp-server/livedata_server.js
vlasky: @filipenevola I have made these changes. I did not include the field 'name' as nothing currently uses it.

PR #11159, opened 2020-09-05
Allow android-webview-video-poster (fixes #8965)

### LINKED ISSUE #8965: Standard Content Security Policy for Cordova is missing the android-webview-video-poster scheme
I am trying to get WebRTC work on Meteor 1.5 and Android devices. On one of them, I get the following error message when trying to add the video stream:

```
Refused to load the image 'android-webview-video-poster:default_video_poster/1383114863730884148' because it violates the following Content Security Policy directive: "default-src * gap: data: blob: 'unsafe-inline' 'unsafe-eval' ws: wss:". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
```

It seems as if Android requires yet another CSP for the protocol scheme `android-webview-video-poster:` to be applied for this scenario. The video that I get in this case is only 2x2 pixels size. I'm not sure if the missing CSP is the cause for that, but this warning did not occur in Meteor 1.2, where everything worked.

I tried to override the CSP locally, by providing a local, patched copy of the `boilerplate-generator` for Cordova, but this doesn't work. Although Meteor recognises my local package, it still uses the CSP default boilerplate. Setting a custom CSP via a custom head.html doesn't work either, because this will only add a second CSP that is ignored by Android. I also tried to use the `BrowserPolicy` package, but this doesn't offer to add completely new protocols to the CSP (which would be very helpful in such cases).

It would be great if the protocol scheme could be added, at least for Meteor 1.5 onwards. I cannot use Meteor 1.5.1 at the moment, due to another issue that breaks my server code. Thanks.

namenotrequired:
This fixes https://github.com/meteor/meteor/issues/8965, also discussed in the [forums](https://forums.meteor.com/t/correct-way-of-setting-content-security-policy-for-a-meteor-cordova-app/43985).

The issue was closed with a [workaround](https://github.com/meteor/meteor/issues/8965#issuecomment-370465644) which [turned out not to work](https://forums.meteor.com/t/correct-way-of-setting-content-security-policy-for-a-meteor-cordova-app/43985/7), forcing some people to maintain local versions of the `boilerplate-generator` package.

As the fix is simple and has no apparent drawbacks, I'd like to suggest we include it in Meteor by default.

### DIFF
--- a/packages/boilerplate-generator/template-web.cordova.js
+++ b/packages/boilerplate-generator/template-web.cordova.js
@@ -29,7 +29,7 @@ export const headTemplate = ({
     '  <meta name="format-detection" content="telephone=no">',
     '  <meta name="viewport" content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width, height=device-height, viewport-fit=cover">',
     '  <meta name="msapplication-tap-highlight" content="no">',
-    '  <meta http-equiv="Content-Security-Policy" content="default-src * gap: data: blob: \'unsafe-inline\' \'unsafe-eval\' ws: wss:;">',
+    '  <meta http-equiv="Content-Security-Policy" content="default-src * android-webview-video-poster: gap: data: blob: \'unsafe-inline\' \'unsafe-eval\' ws: wss:;">',
 
   (headSections.length === 1)
     ? [cssBundle, headSections[0]].join('\n')


StorytellerCZ (approved):

filipenevola (approved):

filipenevola:
Published boilerplate-generator@1.7.1.

derwaldgeist:
Just saw this by accident. Good to see this has finally be resolved :-)

Issue #10913, opened 2020-02-12
request is deprecated


joshleblanc:
Meteor uses the `request` library in the HTTP package.

`request` was deprecated on February 11th, 2020.

https://github.com/request/request#deprecated

`request` is required in the following places:

* https://github.com/meteor/meteor/blob/24865b28a0689de8b4949fb69ea1f95da647cd7a/packages/http/httpcall_server.js#L2
* https://github.com/meteor/meteor/blob/1a5d6fb5b6c94f48539fa6337eb0731c23a2c127/tools/utils/http-helpers.js#L309

harryadelb:
I have been thinking about the same thing yesterday, maybe we should consider some of the [alternatives](https://github.com/sindresorhus/got#comparison).

sebakerckhof:
There's already the fetch package: https://github.com/meteor/meteor/tree/devel/packages/fetch
However, the semantics (especially error handling) of fetch are quite different, which is why it hasn't replaced request in HTTP.

I think the right thing to do is to deprecate the whole HTTP package and for users to migrate from HTTP to the fetch package.

The other request, in meteor tools, could quite easily be replaced with node-fetch.

sebakerckhof:
Some related discussion: https://github.com/meteor/meteor/pull/10029#issuecomment-400231354

evolross:
I think this might be what's causing these server console warnings:
```
(node:24729) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

evolross:
Freshen.

StorytellerCZ:
So what should be the steps here?
Depracate the `http` package with directions to use the `fetch` package?
Then replace any use of `http` in core with `fetch`?

hexsprite:
Sounds like deprecating the `http` package is the way to go, since `fetch` is a standard API.

evolross:
Refresh.

hexsprite:
I see only two usages of request, one is in the HTTP package (server only, client side uses XHR)

And in `tools/utils/http-helpers.js`.

This file has some interesting customizations on request that will need to be preserved.

```
  // A wrapper around request with the following improvements:
  //
  // - It will respect proxy environment variables if present
  //   (HTTP_PROXY or HTTPS_PROXY as appropriate).
  //
  // - It will set a reasonable User-Agent header.
  //
[...]

PR #11161, opened 2020-09-06
HTTP package: deprecate the package

StorytellerCZ:
Added debug warnings to the the http calls warning about the package's deprecation. Logging update [PR](https://github.com/meteor/meteor/pull/11068) should be probably merged before this. Fetch package also needs a proper documentation going forward.

See #10913 for more background.


jankapunkt:
Would it make sense to use `fetch` under the hood? I included the http package mainly because it is a nice bridge between http calls and the calling environment. If the `http` would keep it's structure to the outside world but change it's internals to use `fetch` we would not have to change all the packages, that depend on `http` (including, among others, the accounts-<service> packages)

jankapunkt:
@StorytellerCZ what would be the correct replacement for `HTTP` on the server?

filipenevola (approved):
Should we move to the deprecated folder as well?

StorytellerCZ:
@jankapunkt That would be the fetch package (though on server you should be able to use fetch even without the package).
@filipenevola Yes, that would make sense. Expect an update with that shortly.

Issue #11163, opened 2020-09-08
After upgrade from 1.10.2 to 1.11, Internet Explorer 11 hangs on initial page load when loading xregexp package


perbergland:
My Typescript-based application no longer loads in Internet Explorer 11, but hangs on load with a "a script is taking a long time" error message.

If I break the execution manually, it shows a stack trace of some code in "reify" that it never seems to exit from.

Running in debug mode or in production bundle code doesn’t matter - the same thing happens.

Screenshots attached.

![Screenshot 2020-09-08 kl  20 08 35](https://user-images.githubusercontent.com/2690254/92512898-c8789b80-f20f-11ea-8e2c-b53e7f8afffb.png)

![Screenshot 2020-09-08 kl  20 09 15](https://user-images.githubusercontent.com/2690254/92512914-cb738c00-f20f-11ea-9e4c-f29bbfb80ecd.png)

perbergland:
Ping @zodern  - you seem to be the current authority in the tools/build/reify area. Do you know what could have changed between 1.10.2 and 1.11 to trigger this behavior?

perbergland:
I gave up after 15 minutes of loading time on browserstack (IE11 on Win10) but @zodern said on Slack that it does finally load in his environment.

The 1.10.2 version (all other things equal) loads in 20 seconds on browserstack.

perbergland:
I compared the 1.10.2 and 1.11 versions of the produced .meteor/local/build/programs/web.browser.legacy/modules.js and they look identical save for some modules that were in different locations in the two files.

perbergland:
The culprit is the upgrade of core-js, but need to investigate further.
Checking out ecmascript-runtime-client into packages and changing
```
Npm.depends({
  "core-js": "3.6.5"
});
```
to
```
Npm.depends({
  "core-js": "3.2.1"
});
```
as it was in meteor 1.10 makes IE11 load fine again.

perbergland:
Tested with a few different core-js versions from https://github.com/zloirock/core-js/releases
3.4.8 - works fine
3.5.0 - works fine
3.6.0 - hangs

So something changed in core-js between 3.5.0 and 3.6.0

perbergland:
https://github.com/zloirock/core-js/compare/v3.5.0...v3.6.0

perbergland:
[...]

Issue #11064, opened 2020-05-17
App broken on iOS Safari 9 due to SyntaxError


arggh:
I know it's iOS Safari 9, but our app _used to work_ on this browser with Meteor `1.8.2`. I think supporting old iOS versions is important, since lots of people have old iPads laying around that they use for browsing the web. They can't update the browser past the OS version and Apple doesn't release updates to those old, yet functional devices.

We are updating our app to the latest Meteor `1.10.2` and now the app is broken on iOS Safari 9.

Related to unresolved & closed issue: https://github.com/meteor/meteor/issues/10957

The error message:

```SyntaxError: Invalid parameters or function name in strict mode```

![Screenshot 2020-05-17 at 10 45 09](https://user-images.githubusercontent.com/17210302/82138832-21b95b00-982c-11ea-9158-ec576ced2349.png)
![Screenshot 2020-05-17 at 10 44 54](https://user-images.githubusercontent.com/17210302/82138834-2251f180-982c-11ea-8114-d9887a01d424.png)

brianlukoff:
1.10.2 seems to have this issue on old versions of Safari in general -- even a clean install exhibits the same problem.

filipenevola:
Hi, do you have an easy way to test in old Safari?

Maybe this core-js upgrade was the cause:
https://github.com/meteor/meteor/commit/330a6cfe3d9065e579c04683ffdf8c71bf197234

Also have you tried to use `setMinimumBrowserVersions` https://docs.meteor.com/changelog.html#changes-18?

brianlukoff:
We have been using BrowserStack for testing, which is great: 
https://www.browserstack.com/

We do have `setMinimumBrowserVersions` set to require Safari 11.1 but it doesn't seem to matter here.

arggh:
> Hi, do you have an easy way to test in old Safari?

@filipenevola if you are on a Mac, you can install Xcode and launch the Simulator. From there, you can install different iOS versions and run them on various virtual Apple devices. You can even use desktop Safari's debugger.

brettg2:
+1

I get a similar error in Safari 8... it complains the name of the function and the parameter can't be the same, like function module(require, exports, module)

![image](https://user-images.githubusercontent.com/6796960/84062828-34cbd080-a97d-11ea-9795-353e4108daf6.png)

sonhle:
Hi, I am experiencing the same error with Meteor 1.10.2 and safari 9.1.2. Any workaround? Thanks,

[...]

PR #11166, opened 2020-09-10
Enable BrowserStack tests

filipenevola:
We should have BrowserStack tests running for all our changes.

This will help us to avoid problems like:
https://github.com/meteor/meteor/issues/11064
https://github.com/meteor/meteor/issues/11163


StorytellerCZ (approved):

tools/tool-testing/clients/browserstack/clients.js
```diff
@@ -46,6 +46,7 @@ module.exports = [
     selenium: {
       platform: "MAC",
       browserName: "safari",
+      version: "11",
```
filipenevola, line 49: Read about this change here https://github.com/meteor/meteor/issues/11064#issuecomment-696216504

This is a temporary change.

filipenevola:
We are adding back BrowserStack tests to avoid breaking changes in older browser.

If you have a specific case that you would like to check in old browser please create a new PR with new tests and include them in the Group 11.

PR #11170, opened 2020-09-11
Bump standard-minifier-css version to use the latest version of minifier-css package

afrokick:
I did update minifier-css package to remove a warning about 'caniuse-lite'.

But standard-minifier-css uses minifier-css, so we should update the version of standard-minifier-css.

I tested it locally by cloning standard-minifier-css to project's packages folder and increase its version to 1.6.1. The warning disappeared.


filipenevola:
You are right, `meteor update standard-minifier-css` will remove the warning now.

PR #11176, opened 2020-09-15
Add .help message for __ keyword for 'meteor shell'

kakadais:
- Global project variable __ such as underscore still could be override but there's no information on help.
- Add .help message for __ keyword.

### DIFF
--- a/packages/shell-server/shell-server.js
+++ b/packages/shell-server/shell-server.js
@@ -288,6 +288,8 @@ class Server {
     addHelp("break", "Terminate current command input and display new prompt");
     addHelp("exit", "Disconnect from server and leave shell");
     addHelp("help", "Show this help information");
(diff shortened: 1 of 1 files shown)


hexsprite:
It's not exactly a command in the shell context. it's a variable in JS environment. Not sure it should get a command help entry. But maybe we could show some extra information after the command listings with helpful tips like this?

kakadais:
I understand that it's in the JS context, but it is exposed and overwrited user's global variable, so this would be more dangerous than a shell keyword.
I've seen the commit history and found __ changed from _ and the comment was the same reason, avoid confusing with the underscore library.

So I think this should be changed to avoid confusion.

** I'm not authorized user on this project so please someone fork or do something, if it's avaiable-

Thanks.

filipenevola:
As this is not a command from the shell I don't believe it makes sense to document as a shell help.

PR #11136, opened 2020-08-02
Remove noEmit option from tsconfig.json in TypeScript project skeleton

chris-tse:
Addresses #11135 

Remove the `noEmit` option due to conflicts with the `incremental` option.

PR #11192, opened 2020-10-01
Initial internal dependencies update

StorytellerCZ:
Internal dependencies have not been updated for a while and in some cases there are promised performance improvements by some of the packages and some are required to work in Node 14 (or at least work better).
This is just a first shot of what I could do without much testing. If desired we can continue from here or I can split this thematically. @filipenevola

### DIFF
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -12,43 +12,44 @@ var packageJson = {
     // and we want to make sure there are no dependencies on a higher version
     npm: "6.14.8",
     pacote: "https://github.com/meteor/pacote/tarball/a81b0324686e85d22c7688c47629d4009000e8b8",
-    "node-gyp": "6.0.1",
-    "node-pre-gyp": "0.14.0",
-    typescript: "3.8.3",
+    "node-gyp": "6.1.0",
+    "node-pre-gyp": "0.15.0",
+    typescript: "3.9.5",
     "meteor-babel": "7.9.0",
     // Keep the versions of these packages consistent with the versions
     // found in dev-bundle-server-package.js.
     "meteor-promise": "0.8.7",
-    fibers: "4.0.3",
+    fibers: "5.0.0",
     reify: "0.20.12",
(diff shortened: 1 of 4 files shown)


filipenevola:
> This is just a first shot of what I could do without much testing. If desired we can continue from here or I can split this thematically. @filipenevola

Let's try altogether first 😉 

We have a bunch of tests and we can create many betas to try them in real projects as well.

scripts/dev-bundle-tool-package.js
```diff
@@ -12,43 +12,44 @@ var packageJson = {
     // and we want to make sure there are no dependencies on a higher version
     npm: "6.14.8",
     pacote: "https://github.com/meteor/pacote/tarball/a81b0324686e85d22c7688c47629d4009000e8b8",
-    "node-gyp": "6.0.1",
-    "node-pre-gyp": "0.14.0",
-    typescript: "3.8.3",
+    "node-gyp": "6.1.0",
+    "node-pre-gyp": "0.15.0",
+    typescript: "3.9.5",
```
radekmie, line 17: How about going straight for 4.0? It was already requested in https://github.com/meteor/babel/pull/33 and https://github.com/meteor/meteor/pull/11136. Breaking changes are also [fairly _breaking_](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#breaking-changes).

StorytellerCZ: I don't know enough about TypeScript or its implementation here to risk a major version upgrade. Probably best if someone with that knowledge makes a separate PR which will cover both the internal version and the one in packges.

StorytellerCZ:
@filipenevola The tests all passed, so I think we can try moving forward with this. Second step would probably be upgrading the dependencies that have breaking changes.

scripts/dev-bundle-tool-package.js
filipenevola: I believe typescript update to 4 will be ready first so maybe it's better to not include here directly.

PR #11200, opened 2020-10-04
MongoDB bulk inserts

StorytellerCZ:
I have revived this branch as I'm basing my attempt to add bulk inserts on it. So far I got the basic bulk inserts working, but it isn't optimal and some tests are failing.

Originally I was planning to add the `insertOne` and `insertMany` methods alongside the existing `insert` method, but found that difficult and with a deeper dive found that it would break backward compatibility with packages working with the insert method. So right now these are being used internally so that we no longer use the deprecated MongoDB `insert` method.

So far works, but there are some tests failing. Not sure exactly why, but that is for another coding session. Maybe someone will be faster than me and find the missing code in the old commits or fixes it.

Finally did some modernization and removal of underscore where I could.


StorytellerCZ:
:thinking: Looks like I will have to move the changes to a new branch as it looks like that the original author is MIA.

packages/mongo/collection.js
```diff
@@ -455,51 +455,99 @@ Object.assign(Mongo.Collection.prototype, {
...
+
+
     // On inserts, always return the id that we generated; on all other
     // operations, just return the result from the collection.
-    var chooseReturnValueFromCollectionResult = function (result) {
+    const chooseReturnValueFromCollectionResult = function (result) {
+      if (isBulkInsert) {
+        console.dir(result)
+        if (result.insertedIds) {
+          return Object.values(result.insertedIds);
```
edemaine, line 540: What is the purpose of `Object.values` here? (Also a stray `console` call, if you weren't already aware.)

edemaine (commented):
Just read through the code and found some possible cleanup improvements, and I think one bug (but only in the multi case) -- if old tests are failing, I didn't see why. Will have to actually run the tests to investigate further. Anyway, hope these help.

packages/mongo/collection.js
StorytellerCZ: Yes, stray condole. The purpose is get the ids as the return format from MongoDB is like: `{ "0": "idOfTheDocument", "1": "idOfTheDocument" }`

packages/mongo/collection.js
```diff
@@ -455,51 +455,99 @@ Object.assign(Mongo.Collection.prototype, {
    * @param {Function} [callback] Optional.  If present, called with an error object as the first argument and, if no error, the _id as the second.
    */
   insert(doc, callback) {
+    const isBulkInsert = Array.isArray(doc)
```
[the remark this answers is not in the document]
StorytellerCZ: Doing a rewrite along these lines. Though right now it broke everything, so I need to figure out what is going on.

PR #11202, opened 2020-10-05
When cordova add plugin fails using the id in the git URL we now retry using the plugin name

### LINKED ISSUE #11199: Undocumented breaking change in Meteor 1.11: install plugins by name not id
Until Meteor 1.10.2, we'd install cordova plugins from GitHub based on the plugin ID, not the name. This is the way it's documented in the [guide](https://guide.meteor.com/cordova#installing-plugins) (see "Installing a plugin from Git"). Installing them by name would not work. 

For example, instead of `cordova-plugin-audiorecorder-outsystems` we'd have to write `com.outsystems.audiorecorder`

But since https://github.com/meteor/meteor/pull/11048 was included in Meteor 1.11, apps that install any cordova plugins from github no longer build, unless the IDs of all plugins are identical to their names.

I propose we either:
- revert that PR and continue installing plugins by ID not by name (though this doesn't seem to fix the problem of the author of that PR); or
- update both the guide and the changelog of Meteor 1.11 to document this breaking change

### Edit: reproduction

https://github.com/namenotrequired/plugin-by-name-test

Check out each of the four commits one by one, running `rm -rf .meteor/local/cordova-build && meteor run android-device` each time. You'll find that Meteor 1.11 builds only when the plugin is listed in `.meteor/cordova-plugins` by name, and Meteor 1.10.2 only when the plugin is listed by id.

filipenevola:
Fixes #11199

### DIFF
--- a/History.md
+++ b/History.md
@@ -12,6 +12,8 @@ N/A
 
 * Facebook OAuth has been updated to `1.7.3` now using Facebook GraphAPI v8.
 
+* Cordova add plugin works again with plugin id or plugin name in the git URL as it was before Meteor 1.11. [#11202](https://github.com/meteor/meteor/pull/11202)
+
 ## v1.11.1, 2020-09-16
 
 ### Breaking changes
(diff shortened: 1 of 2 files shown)


namenotrequired (approved):
Excellent solution, thanks Filipe! In my tests this solves my issue without breaking the original issue.

I'm approving because the only comments are about the logs

PR #11210, opened 2020-10-15
Update README.md

gamingclash:


filipenevola:
Hi, thank you for your contribution but the case of the word makes no difference in this case.

Issue #11195, opened 2020-10-02
HTTP.call content type validation


Sikhi:
https://github.com/meteor/meteor/blob/7427525de2d0b6b68e3b73d50e885c396c291fd1/packages/http/httpcall_server.js#L42

Pls make content type validation. without validation in some cases we can throw ugly error from vault, like errors bellow.
It happened when put object in content, not in params with POST request.
>req.grant_type = "authorization_code";
>req.client_id = "someid";

> HTTP.post(endpoint, {
>        content: req,
>        headers: _headers
>    }, callback);

Throw error:
>W20201002-19:15:48.934(3)? (STDERR) _http_outgoing.js:653
>W20201002-19:15:48.936(3)? (STDERR)     throw new ERR_INVALID_ARG_TYPE('first argument',
>W20201002-19:15:48.937(3)? (STDERR)     ^
>W20201002-19:15:48.938(3)? (STDERR)
>W20201002-19:15:48.939(3)? (STDERR) TypeError [ERR_INVALID_ARG_TYPE] [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer. Received an instance of Object
>W20201002-19:15:48.940(3)? (STDERR)     at write_ (_http_outgoing.js:653:11)
>W20201002-19:15:48.941(3)? (STDERR)     at ClientRequest.write (_http_outgoing.js:621:15)
>W20201002-19:15:48.945(3)? (STDERR)     at Request.write (/home/grimnir/.meteor/packages/http/.1.4.2.1fotq9x.lz1d++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/request/request.js:1500:27)
>W20201002-19:15:48.945(3)? (STDERR)     at end (/home/grimnir/.meteor/packages/http/.1.4.2.1fotq9x.lz1d++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/request/request.js:549:18)
>W20201002-19:15:48.946(3)? (STDERR)     at Immediate._onImmediate (/home/grimnir/.meteor/packages/http/.1.4.2.1fotq9x.lz1d++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/request/request.js:578:7)
>W20201002-19:15:48.947(3)? (STDERR)     at processImmediate (internal/timers.js:456:21) {
>W20201002-19:15:48.947(3)? (STDERR)   code: 'ERR_INVALID_ARG_TYPE'
[...]

PR #11212, opened 2020-10-15
#11195

bandersen-hoodoo:
Fixed issue #11195 by checking to make sure if the content is a string, and if it isn't, stringify it.

### DIFF
--- a/packages/http/httpcall_client.js
+++ b/packages/http/httpcall_client.js
@@ -42,6 +42,10 @@ HTTP.call = function(method, url, options, callback) {
   var headers = {};
 
   var content = options.content;
+  if (options.content && typeof options.content !== "string") {
+    content = JSON.stringify(options.content);
+    headers['Content-Type'] = 'application/json';
+  }
   if (options.data) {
     content = JSON.stringify(options.data);
     headers['Content-Type'] = 'application/json';


filipenevola:
Hi @bandersen-hoodoo, do you think it's necessary to force this conversion? The code calling this code should not be aware of the proper input? What do you think?

bandersen-hoodoo:
> Hi @bandersen-hoodoo, do you think it's necessary to force this conversion? The code calling this code should not be aware of the proper input? What do you think?

The forced conversion is just to check if it can be payloaded, and if it can't it would error, but it could always just throw a neater error instead of trying to convert if that works better.

filipenevola:
I would prefer to throw an error if the content is not a string. Could you adjust that in your PR?

StorytellerCZ:
@bandersen-hoodoo Also please note that `http` package has been moved to deprecated folder, so you will need to adjust your PR. Might be worth considering adding the same functionality to the `fetch` package.

StorytellerCZ:
@bandersen-hoodoo do you want to continue working on this or should we close this?

filipenevola:
I'm closing this issue because we haven't heard anything from the original poster for a while.

PR #11222, opened 2020-10-27
MongoDB bulk inserts

StorytellerCZ:
Superseding and continuing work from #11200 doe to issues with CLA.

Adds support for bulk inserts when passing in array of objects into the insert method.

Currently all server test pass, problems comes with client side tests. Getting an `UnhandledPromiseRejectionWarning: RangeError: Maximum call stack size exceeded` on `EJSON.clone`. Will continue investigating, but want to put it out here just in case anyone can spot the issue. After all I have been looking at this code for way too long.

@edemaine I fixed the outstanding issues from that branch here. Please check again, your insight is very much appreciated.


packages/allow-deny/allow-deny.js
```diff
@@ -138,15 +138,34 @@ CollectionPrototype._defineMutationMethods = function(options) {
...
             // In a client simulation, you can do any mutation (even with a
             // complex selector).
-            if (generatedId !== null)
-              args[0]._id = generatedId;
+            if (generatedId !== null) {
+              if (!isBulkInsert) {
+                args[0]._id = generatedId;
+              } else {
+                args[0].forEach(doc => {
+                  doc._id = generatedId;
```
alanning, line 165: If I'm reading this correctly, this line is assigning an array to doc._id rather than an individual ID.

StorytellerCZ: Good catch! Fixed.

Still having client side failures on async calls...

filipenevola:
Hi @StorytellerCZ the docs are failing. Could you check? https://app.circleci.com/pipelines/github/meteor/meteor/732/workflows/51fda12d-5149-4f45-b51a-820e81a059bf/jobs/54069/parallel-runs/0/steps/0-104

filipenevola:
It seems to be almost working but when we insert documents (or even one document) for the first time it's getting duplicated in the client. One document with the _id and the other one without _id but with the same fields.

We also need to think if we want to adopt the new xMany and xOne pattern of Mongo so we need to do this work for update/remove as well.

As we have a work in progress for Mongo package to stop exposing Fibers maybe we could wait to resume this work.

filipenevola:
As we have many things going on with MongoDB in the same time I'm going to close this PR for now, of course we can return to it later or start a new one.

PR #11223, opened 2020-10-27
Remove examples directory

### LINKED ISSUE #11214: Examples directory is outdated
I was going through the different parts of meteor and noticed that the examples haven't been updated more many years, most of them still have the version of meteor prior to v1. 

There are great examples in this directory though I don't think this is documented anywhere in the official docs

Source: https://github.com/meteor/meteor/tree/devel/examples

What could be done, this directory could be extracted into a separate repo and more updated examples can be pushed there, this will keep the codebase just meteor core and packages related only and in the documentation we can link where examples are. 

With some guidance, I am glad to work on getting this done 

cc: @filipenevola

StorytellerCZ:
As was pointed out in #11214 the examples as years out of date and there remains no reason why they should stay in this repository.
Fix #11214


OlivierJM (approved):
This looks great, 
Should we capture the list of these examples so that when creating another repository we try and mirror them or we will start afresh. 
I think these are really great to have and would help people who are new to meteorjs to see how they can implement different kind of features using meteorjs.

StorytellerCZ:
I have copied these examples locally so if there is interest I can see about updating them and pushing them to their own separate repository.

filipenevola (approved):

PR #11227, opened 2020-10-29
fix hmr: btoa unicode support

boomfly:
When module that need to replaced have unicode symbols hrm client package throws error:

Снимок экрана 2020-10-29 в 14 54 08

This break replacement and to see your changes you need to reload full page.

### DIFF
--- a/packages/hot-module-replacement/client.js
+++ b/packages/hot-module-replacement/client.js
@@ -179,8 +179,21 @@ function walkTree(pathParts, tree) {
   return walkTree(pathParts, _module);
 }
 
+// Save default browser functions
+const btoa = window.btoa
+const atob = window.atob
+
+// Replace with unicode support
+window.btoa = function utoa(data) {
+  return btoa(unescape(encodeURIComponent(data)));
+}
+
+window.atob = function atou(data) {
+  return decodeURIComponent(escape(atob(b64)));
+}
+
 function createInlineSourceMap(map) {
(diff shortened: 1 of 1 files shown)


packages/hot-module-replacement/client.js
```diff
@@ -179,8 +179,21 @@ function walkTree(pathParts, tree) {
...
+const btoa = window.btoa
+const atob = window.atob
+
+// Replace with unicode support
+window.btoa = function utoa(data) {
+  return btoa(unescape(encodeURIComponent(data)));
+}
+
+window.atob = function atou(data) {
+  return decodeURIComponent(escape(atob(b64)));
```
namenotrequired, line 192: I cannot test at the moment, but at a glance; was `b64` meant to be `data`?

boomfly: Yes, my bad) but i don't now how it working for me.

Thanks for review. I remove this extra method. This is enough:
```javascript
function utoa(data) {
  return btoa(unescape(encodeURIComponent(data)));
}
```

StorytellerCZ:
@boomfly Thank you for the fix! Please sign CLA so that we can consider it.

boomfly:
@StorytellerCZ, ok. Done. i've already sign before. I need to sign on every PR?

zodern:
@boomfly thanks for fixing this. Could you please create a new PR based off off the `hot-module-reload` branch?

boomfly:
@zodern, done PR #11228

PR #11244, opened 2020-11-13
Catch broken symlinks in isobuild #11241

### LINKED ISSUE #11241: isobuild crashes with ERR_INVALID_ARG_TYPE when encountering broken symlinks in 1.11
I encountered this on the upgrade to 1.11. My project has some broken symlinks in mode_modules/.bin (presumably legacy cruft). They didn't bother meteor pre-1.11, but now cause isobuild to crash with a tremendously cryptic error:
```
=> Started proxy.
/home/xxx/.meteor/packages/static-html/.1.2.2.1httr42.mghq++os+web.browser+web.cordova/plugin.compileStaticHtmlBatch.os/npm/node_modules/meteor/promise/node_modules/meteor-promise/promise_server.js:190
      throw error;
      ^

TypeError [ERR_INVALID_ARG_TYPE] [ERR_INVALID_ARG_TYPE]: The "to" argument must be of type string. Received null
    at validateString (internal/validators.js:120:11)
    at Object.relative (path.js:1054:5)
    at Object.wrapper [as pathRelative] (/home/xxx/.meteor/packages/meteor-tool/.1.11.0.q6y7e2.89d8a++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/tools/static-assets/server/tools/static-assets/server/mini-files.ts:107:14)
```
I eventually figured it out. in meteor_tool/.../tools/isobuild/bundler.js you have

```
      return this._prodPackagePredicate = function isWithinProdPackage(path) {
        const parts = files.pathRelative(sourcePath, path).split(files.pathSep); // Normalize away trailing files.pathSep characters.

...
          if (fileStatus && fileStatus.isSymbolicLink()) {
            // If node_modules/.bin/command is a symlink, determine the
            // answer by calling isWithinProdPackage(real).
            return isWithinProdPackage(files.realpathOrNull(path));
          } // If node_modules/.bin/command is not a symlink, then it's hard
          // to tell which package is responsible for it, so don't strip it.
...
```

you have a potentially recursive call `return isWithinProdPackage(files.realpathOrNull(path));` where the value passed to isWithinProdPackage could be null, and that value will be used to call `files.pathRelative()` which cannot tolerate a null argument.

I would suggest going back to ignoring broken symlinks by checking for null before calling isWithinProdPackage, or at least throwing a more insightful error.

tchiers:
Change isWithinProdPackage() in isobuild/bundler.js to check for broken symlinks and throw a useful error. Fixes #11241.

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -491,7 +491,13 @@ export class NodeModulesDirectory {
         if (fileStatus && fileStatus.isSymbolicLink()) {
           // If node_modules/.bin/command is a symlink, determine the
           // answer by calling isWithinProdPackage(real).
-          return isWithinProdPackage(files.realpathOrNull(path));
+          // guard against broken symlinks (#11241)
+          const realpath = files.realpathOrNull(path);
+          if (realpath) {
+            return isWithinProdPackage(realpath);
+          } else {
+            throw new Error(`Broken symbolic link encountered at ${path}`);
+          }
         }
 
         // If node_modules/.bin/command is not a symlink, then it's hard
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -493,11 +493,10 @@ export class NodeModulesDirectory {
           // answer by calling isWithinProdPackage(real).
           // guard against broken symlinks (#11241)
           const realpath = files.realpathOrNull(path);
-          if (realpath) {
-            return isWithinProdPackage(realpath);
-          } else {
+          if (!realpath) {
             throw new Error(`Broken symbolic link encountered at ${path}`);
           }
+          return isWithinProdPackage(realpath);
         }
 
         // If node_modules/.bin/command is not a symlink, then it's hard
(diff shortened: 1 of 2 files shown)


StorytellerCZ (approved):

filipenevola (approved):

tools/isobuild/bundler.js
```diff
@@ -491,7 +491,13 @@ export class NodeModulesDirectory {
         if (fileStatus && fileStatus.isSymbolicLink()) {
           // If node_modules/.bin/command is a symlink, determine the
           // answer by calling isWithinProdPackage(real).
-          return isWithinProdPackage(files.realpathOrNull(path));
+          // guard against broken symlinks (#11241)
+          const realpath = files.realpathOrNull(path);
+          if (realpath) {
```
filipenevola, line 496: I would change this to 
```suggestion
          if (!realpath) {
              throw new Error(`Broken symbolic link encountered at ${path}`);
          }
          return isWithinProdPackage(realpath);
```

so we avoid if/else and also handle the problematic case first.

filipenevola (changes requested):
Great, could you just update the code to avoid if/elses?

filipenevola (approved):

filipenevola:
Thanks @tchiers, as your changes are in the meteor tool they will be available in the next version of Meteor.

Issue #11215, opened 2020-10-22
node_modules recompile: ‘legacy’ not recompiling js


mfen:
I was trying to recompile `meteor-accounts-t9n` for legacy, as its usage of `class` breaks our app in ie11, but it's not recompiled in the legacy bundle. Changing the recompile option to `modern`, or symlinking the module in `/imports`, does trigger recompilation.

Meteor v1.11.1 
OS X v10.15.5

**package.json:**
```json
"meteor": {
    "mainModule": {
      "client": "client/main.js",
      "server": "server/main.js"
    },
    "nodeModules": {
      "recompile": {
        "meteor-accounts-t9n": "legacy"
      }
    }
  },
```

Using a checkout of meteor, it looks like package-source’s `_findSources` pass for web.browser is first discovering them with `nodeModulesReadOptions` excluding `*.js(on)` and caching node_modules's sources in `_findSourcesCache`. Then the web.browser.legacy `_findSources` pass hits the `_findSourcesCache` where .js was excluded, and so doesn’t add the sources to be recompiled for the legacy unibuild.

Oddly, I see claims of recompile for legacy only [working for others](https://github.com/meteor/meteor/pull/10603#issuecomment-559319720
).

ArchiCroc:
I'm seeing this too.  My app uses the npm query-string module and recompile legacy works on 1.10.2 but not on 1.11.1 or the new 2.0-beta.3. Not sure how to best troubleshoot this further. You can pull my boilerplate app https://github.com/ArchiCroc/pup-plus and try it for your self.

hmelenok:
Same situation. Previously working `package.json` config:
```json
"meteor": {
    "nodeModules": {
      "recompile": {
        "email-regex": [
          "legacy",
          "modern"
        ],
        "debug": [
          "legacy",
          "modern"
        ],
        "normalize-url": [
          "legacy",
          "modern"
        ],
        "serialize-error": [
          "legacy"
        ],
        "isemail": [
          "legacy"
        ],
        "punycode": [
          "legacy"
        ],
```

djhutch:
Same here.
`    "meteor": {
        "nodeModules": {
            "recompile": {
                "query-string": "legacy",
                "rotating-file-stream": "legacy",
                "split-on-first": "legacy",
                "strict-uri-encode": "legacy",
                "debug": "legacy",
                "xml-formatter": "legacy",
                "xml-parser-xo": "legacy",
                "react-table-sticky": "legacy"
            }
        }
[...]

PR #11247, opened 2020-11-17
Revert "Allow cache keys to be shared between archs"

nathan-muir:
This reverts commit e763d959 that was added in #10839. 

When investigating #11215 , a [git bisect](https://github.com/meteor/meteor/issues/11215#issuecomment-729273230) showed that this commit was responsible for the regression.

This commit is breaking the recompilation of npm modules for legacy browsers. This is important for us that support browsers like the old MS Edge.

This may not be the best way to resolve the issue, but just trying to give it some oxygen.

### DIFF
--- a/tools/isobuild/package-source.js
+++ b/tools/isobuild/package-source.js
@@ -1260,10 +1260,9 @@ _.extend(PackageSource.prototype, {
 
     const baseCacheKey = JSON.stringify({
       isApp,
+      arch,
       sourceRoot: self.sourceRoot,
       excludes: anyLevelExcludes,
-      names: sourceReadOptions.names,
-      include: sourceReadOptions.include
     }, (key, value) => {
       if (_.isRegExp(value)) {
         return [value.source, value.flags];


filipenevola:
Hi @zodern, as you did the change that is being reverted I would like to have your input here.

nathan-muir:
Just to make clearer how this change to `cacheKey` affects the recompiling of node modules.

1. Later in the function, it explicitly does _not_ cache node modules that are being recompiled (clears the cache key)

https://github.com/meteor/meteor/blob/8bc33ab75c531ad59d653c4584852a3a7b73124f/tools/isobuild/package-source.js#L1343-L1346

2. If the compiler first compiles for "modern" browsers, and sets the cache key for that npm module, when it goes to compile for "legacy" browsers, it will use the cached version, instead of re-reading with different npm options.

zodern:
Thanks for finding the cause. To keep the optimization, we would need to make sure the cache key includes everything that differs for each arch. I think that would only require adding `nodeModulesToRecompile` to the cache key.

filipenevola:
Hi @nathan-muir, would you be willing to change your PR to add the architecture to the cache key instead of reverting the optimization?

zodern:
@filipenevola The optimization was removing the arch from the cache key so the same cache entry could be used for multiple arch's. If we want to keep the optimization, we would likely have to add `nodeModulesToRecompile` to the cache key.

nathan-muir:
@filipenevola I'm not going to change this pull request. As it adds the `arch` back to the cache key, and fixes the bug.

I would prefer to get it merged and released as-is, and the optimisation can be revisited later.

filipenevola:
Closing this PR in favor of #11290

PR #11253, opened 2020-11-25
Bump to node 12.20.0

StorytellerCZ:
Bumps Node 12 to the latest version 12.20.0, which also includes the [November security update](https://nodejs.org/en/blog/vulnerability/november-2020-security-releases/).

I started of the 1.12 release branch, but we might consider making just a separate release now. If you think it is worth for the security patch: https://nodejs.org/en/blog/vulnerability/november-2020-security-releases/


filipenevola:
I'm updating to 12.20.1 here https://github.com/meteor/meteor/pull/11274

PR #11255, opened 2020-11-26
Bump typescript package version

renanccastro:


afrokick:
meteor-babel's package.json contains version 7.10.2 https://github.com/meteor/babel/blob/master/package.json

Is it ok to use 7.10.5 here?

renanccastro:
Sorry for that, forgot to push on meteor-babel :+1: but the version I've published is .10.5

PR #11265, opened 2020-12-11
Don't use ecmascript features - fixes #11264

### LINKED ISSUE #11264: dynamic-import in 1.12 breaks Internet Explorer
The prefetchInChunks function crashes in Internet Explorer because it uses an arrow function (err =>) without ecmascript transpiling defined in its package.js.

1.11.1 version:
```
function prefetchInChunks(modules, amount) {
    Promise.all(modules.splice(0, amount).map(function (id) {
      return module.prefetch(id);
    })).then(function () {
      if (modules.length > 0) {
        setTimeout(function () {
          prefetchInChunks(modules, amount);
        }, 0);
      }
    });
  }
```

1.12 version:
```
function prefetchInChunks(modules, amount) {
    Promise.all(modules.splice(0, amount).map(function (id) {
      return new Promise(function (resolve, reject) {
        module.prefetch(id).then(resolve).catch(
          (err) => {
            // we probably have a : _ mismatch
            // what can get wrong if we do the replacement
            // 1. a package with a name like a_b:package will not resolve
            // 2. someone falsely imports a_package that does not exist but a
            // package a:package exists, so this one gets imported and its usage
            // will fail
            if (id.indexOf(METEOR_PREFIX) === 0) {
              module.prefetch(
                METEOR_PREFIX + id.replace(METEOR_PREFIX, '').replace('_', ':')
              ).then(resolve).catch(reject);
            } else {
              reject(err);
            }
          })
      });
    })).then(function () {
      if (modules.length > 0) {
        setTimeout(function () {
          prefetchInChunks(modules, amount);
        }, 0);
      }
    });
  }
```

perbergland:
The problem as described in #11264 was that the new code used an arrow function which is incompatible with IE.

I have tested the fix locally by copying the entire package into my "packages" directory with meteor 1.12 and running our app with IE and then it worked fine.

### DIFF
--- a/packages/dynamic-import/dynamic-versions.js
+++ b/packages/dynamic-import/dynamic-versions.js
@@ -70,7 +70,7 @@ function precacheOnLoad(event) {
     Promise.all(modules.splice(0, amount).map(function (id) {
       return new Promise(function (resolve, reject) {
         module.prefetch(id).then(resolve).catch(
-          (err) => {
+          function (err) {
             // we probably have a : _ mismatch
             // what can get wrong if we do the replacement
             // 1. a package with a name like a_b:package will not resolve


StorytellerCZ (approved):

filipenevola:
@perbergland Meteor 1.12.1-beta.0 is released https://github.com/meteor/meteor/blob/devel/History.md#v1121-tbd

Sorry for the delay.

PR #11280, opened 2021-01-07
Be able to provide extra info to Meteor Developer OAuth in the URL

filipenevola:
We need to have more control over what we send to our authentication system, so we are adding a new param for this.


filipenevola:
meteor-developer-oauth@1.2.2 is released.

PR #11281, opened 2021-01-08
Doc fix: Wrong date in History.md

nytamin:
It's 2021 now, not 2022 :)


filipenevola:
@nytamin hahaha, thanks

PR #11291, opened 2021-01-18
Alphabetize GLOSSARY.md

WilliamKelley:
The terms currently follow a loosely related order, but I believe alphabetizing the glossary is more sustainable long-term. It should ease determining "correct" placement of new terms where many relations exist, and it makes it more scannable for readers.

The diff may look a bit weird but I can assure that I've changed no headings, descriptions, or related terms*. 

*However, I did lower-case the anchor sources. In the "Preview Changes" tab of editing, anchors don't render correctly if the source isn't lower-cased and hyphenated; at the same time, it does render correctly when looking at the actual document in the repository, so this change just eases development.


filipenevola:
Thanks @WilliamKelley

PR #11365, opened 2021-03-29
Browserstack tests should run only on browserstack-check branch

filipenevola:
These tests are flaky and also we are going to run them only before releases in a specific branch.

It's already covered in our release process to wait for these tests to pass before a final release.


filipenevola:
Follow up PR here https://github.com/meteor/meteor/pull/11367

PR #11367, opened 2021-03-30
Browserstack check

filipenevola:
These tests are flaky and also we are going to run them only before releases in our internal Jenkins server.

It's already covered in our release process to wait for these tests to pass before a final release.


filipenevola:
I believe CircleCI is changing things and it's not working correctly for PRs.

See above that our workflow is finished and ok but the individual tests (they are inside the workflow) are not running.

filipenevola:
I just got what was happening, CircleCI was not updating the branch checks automatically and then it was forcing some old pending jobs.

PR #11370, opened 2021-04-01
Fix MongoDB startup checks

filipenevola:
MongoDB 4.4.4 has changed some messages.

Due to these changes Meteor apps were not starting with embedded MongoDB as the messages were not matching with the expected ones.

This PR updates these messages and also upgrades the MongoDB driver as it was not up-to-date.

### DIFF
--- a/History.md
+++ b/History.md
@@ -2,8 +2,12 @@
 
 ### Changes
 
-* update MongoDB version to 4.4.4
+* Update embedded MongoDB version to 4.4.4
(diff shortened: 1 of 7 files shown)


filipenevola:
We had to also update our Windows base image in our Windows test env.

We also updated this info here
https://github.com/meteor/docs/commit/e1b5f9568d371213ba80c88bc65ddd7e0620e4bb

Issue #11336, opened 2021-03-02
Accounts.forgotPassword permits user enumeration through inference


bradnorman5490:
Accounts.forgotPassword callback with an error message"User not found [403]" if the user with the specified email does not exist.

Is it possible to stop this error message from reaching the client ? If not is it something that should be changed ? I can make a PR to address this.

tjramage:
@bradnorman5490 — doesn't look like you can change this behaviour, but it wouldn't be too hard to create your own Meteor method to do the exact same thing. The footprint of the forgotPassword API is very small, so you could quite easily copy it and modify to your needs.

Here's the source: https://github.com/meteor/meteor/blob/65b82033b462101b348266ae2e4c36abca876c7d/packages/accounts-password/password_server.js#L583

bradnorman5490:
Thats true but this is something that should be changed in Meteor also

filipenevola:
Hi @bradnorman5490, please go ahead with a PR providing an option for developers to disable this detailed message going to the client.

bradnorman5490:
Working on this now

StorytellerCZ:
PR #11374 takes care of this.

StorytellerCZ:
Set `ambiguousErrorMessages` in `Accounts.config` to true to fix the issue.

tchax:
I wonder if this fix solved the problem of user enumeration. 
In my opinion forgotPassword should never return an error to the client. 
Otherwise the result can still be exploited to perform user enum. Or maybe I missed something ?

StorytellerCZ:
A general error needs to be returned to client so that client can manage a denial of login event. Every service does that.
If there was no error what would the client do in case a login was denied? Show a loading icon indefinitely? What if the user just made a typo and needs to fix that?
If you don't want to display any error or handle it in any case you can ignore it in your code.

StorytellerCZ:
Sorry my mistake. Yes, that is a good point. The only thing that the forgotPassword function should return is an acknowledgement that the function has ran, nothing else. Still it would be nice to return an error if there is an error in sending out the email (as then the user needs to report this to the admin that emailing is not working).

PR #11374, opened 2021-04-06
Added option to disable detailed error message on Account.forgetPassword

mikekibz59:
Bugfix Pr for fixing this issue [issue11336](https://github.com/meteor/meteor/issues/11336)

Added a flag to options object passed through the Account.forgotPassword method.
The default behaviour would be to return a vague message unless the developer explicitly
sets it otherwise.
Please feel free to add any suggestions or thoughts on this pr.


packages/accounts-password/password_client.js
```diff
@@ -223,6 +224,10 @@ Accounts.forgotPassword = (options, callback) => {
     return reportError(new Meteor.Error(400, "Must pass options.email"), callback);
   }
 
+  if (!options.detailedErrorFlag){
+    options.detailedErrorFlag = false
```
pmogollons, line 228: Im not sure if meteor code style enforces the use of ";", but everywhere else is used. It might be better to keep using it in the new code.

mikekibz59: @pmogollons thank you for pointing it out. I will get right to it.

StorytellerCZ (changes requested):
I would like to suggest one addition. If `Meteor.isProduction` is true then the default for the detailed error should be false. This way we ensure that by default on production there is a general error as it should be.

mikekibz59:
> I would like to suggest one addition. If `Meteor.isProduction` is true then the default for the detailed error should be false. This way we ensure that by default on production there is a general error as it should be.

@StorytellerCZ Thanks for the suggestions. I will get right to it.

packages/accounts-password/password_client.js
```diff
@@ -215,6 +215,7 @@ Accounts.changePassword = (oldPassword, newPassword, callback) => {
  * @locus Client
  * @param {Object} options
  * @param {String} options.email The email address to send a password reset link.
+ * @param {Boolean} options.detailedErrorFlag optional parameter. This determines how detail the error message returned from the server should be.
```
StorytellerCZ, line 218: Would be nice to have a bit more info here. This will show up in documentation, so it should include that by default it is false and that people should keep it that way for security purposes.

StorytellerCZ (changes requested):

StorytellerCZ:
@mikekibz59 Making a good progress on this! I think we will be able to merge this soon!

StorytellerCZ (approved):

mikekibz59:
hey @StorytellerCZ sorry for the late response. I've been a little bit busy. Let me check up on it. 
I do have a question though?. I would like to revert some commits then rebase with the latest devel branch but that action will require me to force push my changes. What are your thoughts on `forced pushed commits`?

StorytellerCZ:
@mikekibz59 I think it is fine in this branch if it makes things better. I would say go ahead and we will then review once you are done.

StorytellerCZ:
@renanccastro is correct. Since we are going here for security, we need to define this on the server. @mikekibz59 could you make the changes? The setting would be under `Meteor.settings.packages.accounts-password.detailedPasswordErrors`

mikekibz59:
@renanccastro and @StorytellerCZ , I will look into it and implement it over the weekend.

StorytellerCZ:
@mikekibz59 Just was reviewing code in the neighborhood today and noticed that we already have `ambiguousErrorMessages` in `Accounts.config`, so we should probably utilize that here as well since it is for the same issue.

StorytellerCZ:
@mikekibz59 I have switched this PR to target the 2.3 release now. Hopefully we can merge this soon and have this in the next beta.

mikekibz59:
@StorytellerCZ and @renanccastro when using the `ambiguousErrorMessages` flag, the developer only needs to pass this parameter in the `Accounts.config`. And the rest is taken care of in the `handleError` method. To that the only this I will need to do is to add tests to confirm that functionality.

StorytellerCZ (approved):

packages/accounts-password/password_server.js
```diff
@@ -461,9 +461,10 @@ const pluckAddresses = (emails = []) => emails.map(email => email.address);
 // Method called by a user to request a password reset email. This is
 // the start of the reset process.
 Meteor.methods({forgotPassword: options => {
-  check(options, {email: String});
+  check(options, {email: String})
+
+  const user = Accounts.findUserByEmail(options.email, { fields: { emails: 1 } });
```
StorytellerCZ, line 467: Shouldn't the code to determine the message follow here, or am I missing something here?

StorytellerCZ (changes requested):

packages/accounts-password/password_server.js
mikekibz59: The error code is handled here.
It first checks that the flag is present if it isn't then it proceeds with the error message described.
https://github.com/meteor/meteor/blob/f5d3715431a2dd938afbf2f12d822476bd30b3fa/packages/accounts-password/password_server.js#L109

Issue #10591, opened 2019-06-21
Server always rebuilds even when only client side javascript changed.


Salketer:
I see a couple complains in the form of #10449 about meteor packages causing server rebuilds.

But without using packages, the behavior is the same.

```
meteor create --full serverBuildRepro
```

Changing something in `/imports/ui/pages/home/home.js` causes a server rebuild, while changing something in `/imports/ui/pages/home/home.html` causes only a client rebuild.

I have tried to move back to 1.8.0.1 and the result is the same.
```
meteor --version
Meteor 1.8.1
```

Digging a bit more, it comes to the fact that the ui directory is not inside a client directory. Putting it inside a "client" directory works. The thing is that `/imports/ui/pages/home/home.js` is only imported by the file `/imports/startup/client/routes.js` which is only imported by `/imports/startup/client/index.js`

I see no reason for that particular file to be imported/used server-side, but it still causes server rebuilds. I have also trying using console.log and throwing errors on the root level of the file and nothing impacts the server so I assume it really is not loaded by the server.

aliogaili:
That was the original behaviour, before ES imports, thing under the client folder will only be imported at the client and will result in client only refresh. 

Benjamn mentioned that he has plan to generalize this behaviour (not depending on the client folder) for both the main app and the packages, which I'm really looking forward to!

Salketer:
Yes, but the imports folder was supposed to be exempt of this, no files
inside that folder should be eagerly loaded...

Le lun. 24 juin 2019 à 09:46, Ali Ogaili <notifications@github.com> a
écrit :

aliogaili:
Are they eagerly loaded? I don't think so, I think it just impact the dev refresh behaviour, but maybe I'm wrong. Either way, since the client files can be inferred by tracing the import dependency tree starting from the client entry point, I think the refresh behaviour can be generalized beyond the client folder, for both the app and packages, which I think is what you looking for.

sakulstra:
> Benjamn mentioned that he has plan to generalize this behaviour (not depending on the client folder) for both the main app and the packages, which I'm really looking forward to!

Would this also help with issues like #10106? We've moved to a multi package.json architecture to have jest, eslint and other dev only depencencies in another node_modules outside the meteor app... which is pretty annoying as we have to configure jest and webpack to resolve modules from multiple folders causing hard to debug bugs from time to time.

Salketer:
[...]

PR #11381, opened 2021-04-09
Avoid eager loading files to watch set when main module is defined

filipenevola:
Avoid eager loading files to watch set when main module is defined

This is the issue we are trying to fix here https://github.com/meteor/meteor/issues/10591

### What is the problem?

We concluded that isobuild package-source is being created with lazy as undefined in a few cases even when mainModule is not defined on package.json. 

So later in the process Meteor is going to decide which files are really lazy or not but before that the lazy information is used to create the watchSets, including the one for the server. In this case if you have a file that is only being used in your client but it lives in a folder different than `server`, `client` and `imports` it's already included in the watch set for the server.

That is the problem, as the lazy was deferred and watch set is already created including this file, even when the file is not really used in the server it's already tracked in the server watch set, leading to unexpected restarts in the server when only the client code was changed.

### Solutions

We had two possible approaches:

- 1: Review the watch set later and try to remove unused files for each architecture:
  - Cons: it would add one more process to Meteor build and it could lead to performance degradation.
- 2: Consider every file already lazy in the first pass when mainModule is defined on package.json:
  - Cons: it could affect other flows.

We've decided to follow the approach 2.

We also added tests for this specific cases, and yes, they were failing before our change and now they are passing.

### Tasks

- [x] Identify the issue
- [x] Fix the issue
- [x] Create new test cases
- [x] Fix existent test cases


zodern:
I've never fully looked into how potentiallyUnusedFiles works, and why Meteor treats files in `imports` differently than those outside of it, so I finally did that the past few days.

`potentiallyUnusedFiles` is a list of files that are always lazy (only included in the built app if imported from js, such as files in `imports` or `node_modules`) and don't always need to be watched. Toward the end of the build process, Meteor loops through all of the js resources that are part of the built app, and records them as definitely used. Currently `potentiallyUnusedFiles` is only used for the server watchset, though it is also created for the client and packages. Meteor still watches every file outside of a `server` folder, but if a modified file is not definitely used by the server Meteor only does a client rebuild.

Most of the difference seems to be related to non-js resources (resources added by calling `file.addStylesheet`, `file.addHtml`, or `file.addAsset`).

Within `imports`:

- Non-js resources are lazy by default so they are not added to the app. There is no way to import non-js resources, so it is impossible to get them added to the built app.
- Meteor converts lazy stylesheets into js resources, so the rules for non-js resources don't apply to them
- If a compiler build plugin makes a resource non-lazy, it does get added to the app, but this removes the benefit of being in `imports` and it is still recorded as `potentiallyUnused` since the file is not added as js

The only files that can be used from `imports` are files added to the built app as js. Any other file is either impossible to get added to the built app, or will be eagerly added to the app and still be recorded as potentiallyUnused.

App code outside of `imports`:
- All non-js resources are not lazy and eagerly added to the app.
- Since these stylesheets are not lazy, their content is eagerly added to the app like any other non-js resource. If they are imported from js Meteor adds an empty js file so there are no errors during runtime of not being able to require the css module. This has no effect on the actual stylesheet.

There are also some other differences:
- Compilers that use `MultiFileCachingCompiler` don't work correctly if root files import files in `imports`. For example, `.import.less` files in `imports` are always recorded as potentially unused.
- Files that are not added to the built app, for example config files used by linter build plugins, are always recorded as potentially unused

What it means for this PR:

The majority of apps would work perfectly. It assumes that all js/ts files will be turned into a js resource, and only used if it is imported, which might not always be true:

- compilers or linters might use js/ts files but not add them to the built app. For example, if a linter uses `.eslintrc.js` files, or a compiler uses `.d.ts` or `.babelrc.js` files. If the file isn't added to the built app, it will always be marked as potentially unused.
- compilers can compile js or ts files, but add them as non-js resources. For example, I've been planning to create a build plugin that takes files with a `.worker.js` extension and creates assets that can be loaded as a web worker or in a Node.js worker thread. Since it would not be added as a js resource, it would be recorded as potentially unused.

I'm not aware of any existing plugins that would be affected by this, but I am only familiar with the implementation of a small number of them. Besides this, my only other concern is it relies on a hard coded list of extensions which doesn't cover the large number of file types that are compiled into js, and adding an extension limits how compiler plugins can handle that type of file without affecting which files are watched.

Another option to fix this:

I think we should be able to use the same logic for all files regardless of their extension, if they are within `imports`, or if there is a main module. In addition to reducing what would trigger a full rebuild, this would fix existing bugs with potentiallyUnused.

1. When assets are added, record which file the non-js resource was added from. This would require compilers to call `addAsset`, `addHtml`, and `addStylesheet` on the file the content was from instead of on any file. Most if not all already do this, and the build plugin changing which file it uses wouldn't break its compatibility with older Meteor versions.
2. Mark all files initially as potentiallyUnused
3. After the build, in addition to the js resources, loop through added non-js resources and record any that are not lazy as definitely used
4. Decide what to do about files that are only used by the build plugin, but not added to the app (for example, config files for linter build plugins). A couple options are to detect source files that never had a js/non-js resource added for them and mark them as definitely used, which might have some false positives and negatives, or require the build plugin to manually mark these files as definitely used
5. Fix `MultiFileCachingCompiler` so all paths listed in `referencedImportPaths` are marked as definitely used in the watch set

This would change the logic for which files get watched to:
- any file that a compiler build plugin did not add a resource for
- any resource that is not lazy
- any js/css resource that was imported

Or, in other words, the files that are not watched are the files used to create lazy resources, and if the resource is js/css it was not imported.

With these changes, we should also be able to use `potentiallyUnused` with the client watchset, so modifying a file not used by Meteor won't trigger a re-build with the possibility of in the future never reading, hashing, and watching the file at all.

filipenevola:
I'm closing this PR after discussions with @zodern 

The solution will be something similar to this other PR https://github.com/meteor/meteor/pull/10455, but only using the getters to detect which files are used.

PR #11391, opened 2021-04-20
Fix Timestamp.ONE is undefined

### LINKED ISSUE #11380: Timestamp.ONE is not defined
Back in 2017, mongo made `Timestamp` inherit from `Long` in the BSON library: https://github.com/mongodb/js-bson/commit/f6fd5c2f973bd551115c144121e97b2d8fcd3311#diff-2602d0defc7490c967453bcc317f7de1f9a4296f8affa36db7db7f2ad64b8a80

Now, the oplog_tailing in meteor is "broken" in the rare case that there is no `op.ts`: https://github.com/meteor/meteor/blob/devel/packages/mongo/oplog_tailing.js - I've only seen this happen when trying to add support for transactions (the transaction oplog message has no `ts`). 

The fix here is trivial - switch `Timestamp.ONE` to `Long.ONE`

znewsham:
Fixes #11380 

swaps `Timestamp.ONE` to `Long.ONE` to make it compatible with the more recent versions of the mongo driver (post 2017)


StorytellerCZ:
@znewsham Can you please add a history entry and patch bump the mongo package? We can release this after that.

StorytellerCZ:
Published as mongo@1.11.1.

znewsham:
@StorytellerCZ thanks for handling!

PR #11392, opened 2021-04-20
Added support for this.unblock() in Meteor.publish() context

vlasky:
Followup to https://github.com/meteor/meteor-feature-requests/issues/11


StorytellerCZ:
We will probably have to makes some changes to the documentation with this. We should probably tie this change to the next release, what do you think @vlasky ?

filipenevola (approved):

derwaldgeist:
Did this PR consider the problem that after unblocking any change to the sub parameters will cause every document contained in the publication to be transferred again? This is a big issue with the community packages.

https://forums.meteor.com/t/strange-phenomenon-subsequent-methods-or-subs-wont-succeed/55029/67

PR #11396, opened 2021-04-23
Update faye-websocket dependency

StorytellerCZ:
Update `faye-websocket` and related dependencies. This will remove Node's HTTP parser deprecation message.


filipenevola:
Published socket-stream-client@0.3.2.         
Published ddp-server@2.3.3.

PR #11400, opened 2021-04-26
Add cache for reify compiler when used in runtime

zodern:
Meteor uses reify to compile any files from `node_modules` imported on the server so they can use the import/export syntax. In the apps I've profiled, ~1/3 of the time spent (re)starting the server was from reify's compiler. In one app, this cache reduces the time spent by the compiler from 2 - 3 seconds to less than 200ms. Most of the remaining overhead is from the extra stat it does per file.

The Meteor Tool also does this for any npm dependencies it or build plugins imported. In one app, this added 10 seconds to `meteor run` or any other command that loads the build plugins. For now I have the cache disabled in the meteor tool for two reasons:
1) If there is an issue with the cache, it is harder to find and clear it for the Meteor Tool compared to the app
2) I'm worried about how large the cache will grow for anyone who has local packages with build plugins

The cache in the Meteor Tool can be enabled by setting the env var `METEOR_TOOL_ENABLE_REIFY_RUNTIME_CACHE=true`. I will be using Meteor with it enabled to see how well it does with local packages with build plugins. If there are no problems, we could enable it by default in the future.


StorytellerCZ:
This will need an addition to the docs so that people are aware of this option.

StorytellerCZ (approved):

filipenevola (approved):

filipenevola:
@StorytellerCZ @zodern did we document this env var somewhere METEOR_TOOL_ENABLE_REIFY_RUNTIME_CACHE?

We should also mention it in History.md

StorytellerCZ:
@filipenevola Will add to 2.3 history. @zodern you can add it to your docs PR here: https://github.com/meteor/docs/pull/684

santiagopuentep:
In the changelog there are two entries, one for `METEOR_TOOL_ENABLE_REIFY_RUNTIME_CACHE` and one for `METEOR_REIFY_CACHE_DIR`. Should we enable both? If so, where should we point the `METEOR_REIFY_CACHE_DIR` to?

zodern:
In development you don't need to do anything since it is enabled by default for the app's server. You can optionally enable `METEOR_TOOL_ENABLE_REIFY_RUNTIME_CACHE` by setting it to `true`, which will make loading build plugins faster. Before enabling it by default I want to check how it works long term when there are local build plugins since it could cause the cache to become very large.

`METEOR_REIFY_CACHE_DIR` is only needed if you want to enable the optimization in production, but for most apps there probably wouldn't be enough benefit from it in production, and many Meteor hosting services don't support it since it requires the cache files to be preserved across app restarts. Also, it uses the file's inode number and mtime as part of the hash key, which with most deployment tools would be different for the files after each deploy so it would only help when the app restarts.

santiagopuentep:
Great, thanks for clarifying!

PR #11395, opened 2021-04-23
Remove compat bellow 1.0

StorytellerCZ:
Re-creating #11226 with cherry picked commits.

PR #11401, opened 2021-04-28
Improve password validator

StorytellerCZ:
Improve password validator and add option to settings to override the maximum length of password.


StorytellerCZ:
If we combine this with #11395 then we can just leave here the second part for the validation for the hashed password.

renanccastro (approved):

StorytellerCZ:
Published accounts-password@1.7.1

PR #11409, opened 2021-05-03
Reverse order of adding plugins/adding platforms on cordova build

renanccastro:
Fix by first installing plugins, and then adding the Cordova platform android/ios, it will avoid errors due to plugins options declared in the config.xml.


filipenevola (approved):

brucejo75:
@renanccastro,

I verified that this repairs the issue I was seeing with the my repro: https://github.com/brucejo75/issue_11005

Thanks!

PR #11412, opened 2021-05-07
Fix #9960 by adding hot-code-push to minimal skelet

### LINKED ISSUE #9960: Meteor create --minimal is forcing meteor server restart on client changes

StorytellerCZ:
Adding `hot-code-push` to minimal skeleton fixes and issue where the server would reload when a change occurred on client.


StorytellerCZ:
@filipenevola @renanccastro what are your thoughts on this change?

filipenevola:
I believe the root cause is the same problem that I'm fixing here https://github.com/meteor/meteor/pull/11381

This PR is going to be closed as @zodern is going to implement a different approach to solve the same issue.

@zodern do you agree with me? Because if your fix is going to fix this restart for the minimal as well, so we don't need to merge this PR.

zodern:
If an app doesn't have the `autoupdate` package, Meteor always does a full rebuild: https://github.com/meteor/meteor/blob/ffcfa5062cf1bf8a64ea64fef681ffcd99fe7939/tools/runners/run-app.js#L657-L660 If `hot-code-push` fixes the issue, then this is a good solution since it should only include `autoupdate` in development, and maybe this also solves https://github.com/meteor/meteor-feature-requests/issues/354

zodern (dismissed):

zodern:
Is the reason `hot-code-push` is missing because it increased the production bundle size?

If that is still a concern, one option might be to add Autoupdate as a dependency to `hot-module-replacement` (I'm not sure why it isn't already), and add `hot-module-replacement` instead. Since it is a dev dependency, it and `Autoupdate` should be excluded from the production bundle. This might also solve https://github.com/meteor/meteor-feature-requests/issues/354.

zodern:
> @zodern do you agree with me? Because if your fix is going to fix this restart for the minimal as well, so we don't need to merge this PR.

Meteor always does a full rebuild unless the app uses the `autoupdate` package:
https://github.com/meteor/meteor/blob/ffcfa5062cf1bf8a64ea64fef681ffcd99fe7939/tools/runners/run-app.js#L657-L660

The changes I am making to watching won't change that.

StorytellerCZ:
Changing for `autoupdate` package.

filipenevola:
@StorytellerCZ let's resume what is going on here.

Did you try to add hot-code-push to minimal but it added more code to the bundle size, is that the issue with this approach? How much was added?

I would like to hear from people using the minimal in production. Is it a real use case? I will ask about this in our community [Slack](https://join.slack.com/t/meteor-community/shared_invite/enQtODA0NTU2Nzk5MTA3LWY5NGMxMWRjZDgzYWMyMTEyYTQ3MTcwZmU2YjM5MTY3MjJkZjQ0NWRjOGZlYmIxZjFlYTA5Mjg4OTk3ODRiOTc).

Issue #11404, opened 2021-04-30
Is there a use for id-map in mongo-id package?


harryadel:
Was looking into the mongo-id package and noticed the inclusion of id-map package in mongo-id but it's not made use of. It's also a prime candidate for NPM migration, so would you be ok if I opened up a pull request to have it use [this](https://github.com/harryadel/blaze-standalone/tree/main/packages/mongo-id) instead?
https://github.com/meteor/meteor/blob/0b9feae1640aeddd256c59b9c52bc9c61cc28e27/packages/mongo-id/package.js#L9

StorytellerCZ:
If `id-map` is not being used there then the best thing would be to remove the dependency, unless it is just being passed through to do something else. I would say go ahead with the PR if it won't take much of your time as it will be a good proof of concept.

StorytellerCZ:
Fixed in `mongo-id@1.0.8`.

PR #11414, opened 2021-05-07
Remove id-map dependency from mongo-id

harryadel:
https://github.com/meteor/meteor/issues/11404


harryadel:
I guess it's ready now @StorytellerCZ. Would you like me to do a patch version bump? Also, for some reason some of the tests were failing locally but tests are passing here, that's why I kept it as a draft PR first.

StorytellerCZ:
@harryadel No need to worry about version bump, I'll do that.

StorytellerCZ:
Published as `mongo-id@1.0.8`.

Issue #11382, opened 2021-04-10
[PERFORMANCE] - Make email a low case string and amend/simplify searches in the Users collection


paulincai:
I am really sorry if this was discussed already in the form of an issue. I was supposed to raise this one a couple of months back and got disconnected from business for a while.

The issue is explained at this link:
https://forums.meteor.com/t/slow-mongo-query-on-createuser/54879

"
My app has about 400K users, and this query can take up to 6000ms (especially when the email address starts with a common first name).
"

StorytellerCZ:
@paulincai What is the proposed solution here? From what I have seen on the forums, using collation seemed to be favored, but the downside I see there is that you need to define a locale which might create some interesting edge cases. I'm all for improving the search script, but please bear in mind that the local-part of the e-mail can be case sensitive as per [RFC5321](https://tools.ietf.org/html/rfc5321#section-4.1.2).

paulincai:
@StorytellerCZ 
Given the fact that the matter is standardized but also subject to preference / opinion by some large email providers, I think this is more about what I feel than what I think :).
The practical questions I see are:
- is Paul@gMail.com and paul@gmail.com two different entities/people/recipients?
- is there any benefit in capitalizing an email address (from the system provider perspective and user perspective).

In the example below I am logging in to my Yahoo with an existing address and I am being matched to a lower case address.


![Screen Shot 2021-04-23 at 14 11 06](https://user-images.githubusercontent.com/26188915/115856867-101a0380-a43e-11eb-9a51-0af9453565f5.png)
![Screen Shot 2021-04-23 at 14 11 15](https://user-images.githubusercontent.com/26188915/115856874-114b3080-a43e-11eb-9b99-6ed7359a4e87.png)

In the example below, **which is my recommendation for our subject in discussion here,** I trying to create a mixed case address and I am being provided with a lower case address.

![Screen Shot 2021-04-23 at 14 15 22](https://user-images.githubusercontent.com/26188915/115857435-bfef7100-a43e-11eb-9016-bb93cdd7d801.png)
![Screen Shot 2021-04-23 at 14 16 30](https://user-images.githubusercontent.com/26188915/115857450-c41b8e80-a43e-11eb-9277-911d2f2e8e65.png)

We could let the user type in a mixed case address and, like Yahoo, convert to lower case before saving to DB. This could be done in two ways:
- make Meteor only accept lower case addresses and inform the developer about it with some error.
[...]

PR #11419, opened 2021-05-10
Use lower case emails fix

ritwik1233:
This pull request is related to #11382

As a fix, I added a new key in the accounts_common.js, in the account-base package, called **"useLowerCaseEmails"**. When this key is set  to true and the fieldName argument--in **"checkForCaseInsensitiveDuplicates"** method--is "emails.address", the   **"checkForCaseInsensitiveDuplicates"** will not implement  **"selectorForFastCaseInsensitiveLookup"** method instead will run a simple regex query for email search.


packages/accounts-base/accounts_common.js
```diff
@@ -204,7 +204,7 @@ export class AccountsCommon {
     const VALID_KEYS = ["sendVerificationEmail", "forbidClientAccountCreation", "passwordEnrollTokenExpiration",
                       "passwordEnrollTokenExpirationInDays", "restrictCreationByEmailDomain", "loginExpirationInDays",
                       "loginExpiration", "passwordResetTokenExpirationInDays", "passwordResetTokenExpiration",
-                      "ambiguousErrorMessages", "bcryptRounds", "defaultFieldSelector"];
+                      "ambiguousErrorMessages", "bcryptRounds", "defaultFieldSelector", "useLowerCaseEmails"];
```
StorytellerCZ, line 207: Please add description about this new option above in the comment so that it shows up in documentation.

packages/accounts-password/password_server.js
```diff
@@ -239,8 +239,18 @@ const checkForCaseInsensitiveDuplicates = (fieldName, displayName, fieldValue, o
   const skipCheck = Object.prototype.hasOwnProperty.call(Accounts._skipCaseInsensitiveChecksForTest, fieldValue);
 
   if (fieldValue && !skipCheck) {
+    let selector = {};
+    const { useLowerCaseEmails } = Accounts._options;
+    // if useLowerCaseEmails is true in Accounts.config and fieldName='emails.address'
+    // then don't run selectorForFastCaseInsensitiveLookup
+    // instead run a simple regex query for email.
+    if (useLowerCaseEmails && (fieldName === 'emails.address')) {
+      selector[fieldName] = new RegExp(`^${Meteor._escapeRegExp(fieldValue)}$`, 'i');
```
StorytellerCZ, line 248: Why not just use MongoDB regex option?
https://docs.mongodb.com/manual/reference/operator/query/regex/

StorytellerCZ (changes requested):

packages/accounts-base/accounts_common.js
ritwik1233: Hi I have made the necessary changes and create a new commit

packages/accounts-base/accounts_common.js
```diff
@@ -169,6 +169,7 @@ export class AccountsCommon {
    * @param {Number} options.passwordEnrollTokenExpiration The number of milliseconds from when a link to set initial password is sent until token expires and user can't set password with the link anymore. If `passwordEnrollTokenExpirationInDays` is set, it takes precedent.
    * @param {Boolean} options.ambiguousErrorMessages Return ambiguous error messages from login failures to prevent user enumeration. Defaults to false.
    * @param {MongoFieldSpecifier} options.defaultFieldSelector To exclude by default large custom fields from `Meteor.user()` and `Meteor.findUserBy...()` functions when called without a field selector, and all `onLogin`, `onLoginFailure` and `onLogout` callbacks.  Example: `Accounts.config({ defaultFieldSelector: { myBigArray: 0 }})`.
+   * @param {Boolean} options.useLowerCaseEmails When this key is set to true and the fieldName arguement, in "checkForCaseInsensitiveDuplicates" method, is "emails.address" then the "checkForCaseInsensitiveDuplicates" will not implement "selectorForFastCaseInsensitiveLookup" method instead will run a simple regex query for email search.
```
filipenevola, line 172: typo in argument (arguement)

packages/accounts-password/password_server.js
```diff
@@ -239,8 +239,18 @@ const checkForCaseInsensitiveDuplicates = (fieldName, displayName, fieldValue, o
   const skipCheck = Object.prototype.hasOwnProperty.call(Accounts._skipCaseInsensitiveChecksForTest, fieldValue);
 
   if (fieldValue && !skipCheck) {
+    let selector = {};
+    const { useLowerCaseEmails } = Accounts._options;
+    // if useLowerCaseEmails is true in Accounts.config and fieldName='emails.address'
+    // then don't run selectorForFastCaseInsensitiveLookup
+    // instead run a simple regex query for email.
+    if (useLowerCaseEmails && (fieldName === 'emails.address')) {
```
filipenevola, line 247: The extra parenthesis is not needed here

filipenevola (changes requested):
It looks good. Thanks.

I just suggested small changes.

packages/accounts-base/accounts_common.js
ritwik1233: Hi I am made the necessary change

packages/accounts-password/password_server.js
ritwik1233: Hi I have made the necessary change

StorytellerCZ (changes requested):
One more thing to add is enforcement of lowercase e-mails when creating user account

ritwik1233:
> One more thing to add is enforcement of lowercase e-mails when creating user account

Hi @StorytellerCZ , before implementing the lowercase email enforcement, I just wanted to check with you that isn't this enforcement against the [RFC5321 guidelines](https://datatracker.ietf.org/doc/html/rfc5321#section-4.1.2) as mentioned by you in [Issue #11382](https://github.com/meteor/meteor/issues/11382)

filipenevola (changes requested):
Hi @ritwik1233 I believe we should force lowercase emails in the creation as well when this new flag is true.

I never had a problem with this approach and I do this in all the systems I created.

Could you include that in this PR and also documentation about this change?

I'm also wondering if that shouldn't be the default, instead of the need to provide true on `useLowerCaseEmails` maybe it should be the default. So if people want to keep the old behavior they need to start to provide `false`.

What do you think?

ritwik1233:
> I'm also wondering if that shouldn't be the default, instead of the need to provide true on `useLowerCaseEmails` maybe it should be the default. So if people want to keep the old behavior they need to start to provide `false`.
> 


@filipenevola, I like this approach that lowercase enforcement is the default behavior, I feel it is the right way. I will make the necessary changes.

packages/accounts-password/password_tests_config.js
```diff
@@ -0,0 +1,4 @@
+// Set useLowerCaseEmails's value to false to prevent lowercase email enforcement in the system.
```
StorytellerCZ, line 1: Wrong comment, the settings is set to `true`. Also aren't test missing in this file?

StorytellerCZ (changes requested):

packages/accounts-password/password_tests_config.js
ritwik1233: @StorytellerCZ  I guess the comment is not explanatory. I will fix this. I wanted to say that if you explicitly set the value of uselowercaseEmails to false then the system will not enforce lowercase emails or else by default it will implement lower case email enforcement.

This file is only a config setup the tests are being done in password test and email test file. I have not written any new tests as the present tests were covering all use cases that i could think of. So i mostly modified the exisiting tests to accomodate for lowercase enforcement.

Some tests where checking for case insensitive email which is not applicable if the system enforce lower case email, so I shifted those tests inside an if condition which check the value of uselowercaseEmails and if the value is explicitly set to false, implement these test. I tested it in my local system by toggling the useLowerCaseEmails value to false  and true and it worked.

StorytellerCZ: Another approach would be to run the tests with the option set to true and then again when set to false (minus the case sensitive tests you have mentioned).

StorytellerCZ (approved):
LGTM, once tests pass we can merge this.

zodern:
I haven't seen it described, so I am wondering what the plan is for using lower case emails to improve the performance issue in #11382, or if the performance issue will be handled separately.

packages/accounts-password/password_server.js
```diff
@@ -239,8 +239,22 @@ const checkForCaseInsensitiveDuplicates = (fieldName, displayName, fieldValue, o
...
+    const { useLowerCaseEmails } = Accounts._options;
+    // if useLowerCaseEmails is false in Accounts.config and fieldName='emails.address'
+    // then run selectorForFastCaseInsensitiveLookup method
+    // else run a simple regex query for email.
+    if (fieldName === 'emails.address') {
+      if(useLowerCaseEmails === false) {
+        selector = selectorForFastCaseInsensitiveLookup(fieldName, fieldValue);
+      } else {
+        selector[fieldName] = {$regex: fieldValue, $options: 'i'};
+      }
```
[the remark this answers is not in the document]
wildhart: If we're sure the db emails are lower case(*), and the query email can be converted to lowercase, why not just use a simple match `selector[fieldName] = fieldValue.toLowerCase()`?  Isn't that the whole point of this PR?

(*) we can only be sure that all exiting db emails are lower case if we make this to be an opt-in behaviour, and inform the developer that they must migrate the db before opting in...

wildhart:
> I'm also wondering if that shouldn't be the default, instead of the need to provide true on `useLowerCaseEmails` maybe it should be the default. So if people want to keep the old behavior they need to start to provide `false`.

Wouldn't this break existing applications where developers who upgrade don't notice that they suddenly have to set `userLowerCaseEmails: false`, and their db is already populated with 1000s of mixed-case emails?  Before this option can be set on an existing project the developer must run a migration script to convert all existing emails to lower case...

packages/accounts-password/password_server.js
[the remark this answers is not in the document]
ritwik1233: > Have you compared the performance for both of these queries? In the past, `selectorForFastCaseInsensitiveLookup` was much faster than using a case insentive regexp by itself since it isn't able to use indexes. Both queries should produce the same results, so it should be fine to use the the faster query in both cases.

Hi  @zodern,
I did the bench-marking tests for this query.
I created two meteor apps with the with this pr's build, one using **useLowerCaseEmails = true** and other using **useLowerCaseEmails = false** in Accounts.config.
I used [performance.now() ](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)to log the time spent on creating new a user in both the meteor apps. I do not see significant time difference after each call to Accounts.createUser.
Below is the code snippet  that I ran on client side in the app. This code was basically creating random users in both the application at an interval of 7 seconds.
 ```
setInterval( async ()=>{
    const id = Random.id();
    const email = ritwik-tesT${id}@gmail.com
    const t0 = performance.now();
    await Accounts.createUser({
      email,
      password: 'test-password'
    });
    const t1 = performance.now();
    console.log(Call took ${t1 - t0} milliseconds.);
  }, 7000);
```
But one thing I noticed is that after running both apps locally for around 30 minutes continuously, the app with **useLowerCaseEmails = false** created around **529** users while the app with **useLowerCaseEmails = true** created around **674** users.

Not sure If this is a good bench marking test, but I think this difference is caused by **selectorForFastCaseInsensitiveLookup** method being used in the app with **useLowerCaseEmails = false** and the emails starting with common first name.

StorytellerCZ:
The default setting is open to discussion. Right now we need to finalize the code and ensure that it is doing what we need to achieve the performance goal.

StorytellerCZ:
Any updates here on performance?

packages/accounts-password/password_server.js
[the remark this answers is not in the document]
zodern: I created a couple scripts that can be run in the mongo shell to help test performance: https://gist.github.com/zodern/4ad546b9ddc2cf780fed476c85c1f053

The first one `generate-users.js` generates the emails and inserts them into the `usersTest` collection. `GROUPS` is the number of different combinations for the first 4 characters used in the generated emails.

`run-queries.js` samples 10 emails from the collection, adds a suffix to each one to make it unique, and then prints how long queries took for the emails.

I didn't run too many tests, but it seems as the number of emails or groups increases `selectorForFastCaseInsensitiveLookup ` becomes more efficient since it is able to use indexes to exclude emails that don't have the same first 4 characters, whereas the case insensitive regexp always does a collection scan. When there is only one group, the case insensitive regexp seems to be faster.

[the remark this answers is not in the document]
zodern: > If we're sure the db emails are lower case(*), and the query email can be converted to lowercase, why not just use a simple match selector[fieldName] = fieldValue.toLowerCase()? Isn't that the whole point of this PR?

This would work and fix the performance, though it is a little more complicated since apps with users created before Meteor 1.2 might have multiple users with emails that only differ by case, and when they login Meteor still treats the email as case sensitive.

StorytellerCZ:
Alright, we will start 2.5 with this. What we need now is history entry with a sample migration script to convert all e-mails to lowercase that devs could use as a starting point for their apps. Can you do that @ritwik1233 ?

ritwik1233:
Hi @StorytellerCZ ,
I will look into this.
But I will be able to work on this by end of this month.
Is that ok?

StorytellerCZ:
@ritwik1233 I can handle the history initial entry, you can then add the example and docs later on at your leisure. I expect 2.5 to be released in September so there is still enough time.

markdowney:
Hi all,
As @zodern mentioned, the case insensitive regexp will perform worse since it doesn't use indexes (confirmed by @paulincai 's benchmark: https://github.com/meteor/meteor/issues/11382#issuecomment-900948768). Also this issue surfaced with very common first names, where the size of the group with the same first 4 letters was over 10K.

I fixed this in my own project by following @wildhart 's suggestion and simply querying for `selector[fieldName] = fieldValue.toLowerCase()`. If applied here, this change should only be opt-in as developers must make sure that all addresses were stored in lower-case in the first place.  

That being said, I re-ran some tests and found significant differences when running some queries locally VS on Mongo Atlas. For 1M records, the `fastCaseInsensitiveLookup` would take about ~0.2s locally VS ~8s on Atlas. I noticed my Atlas Mongo was still at version 4.0 and upgrading to 4.2 seems to have completely resolved the issue. Atlas now performs similarly to my local Mongo.

If nobody else is experiencing performance problems at signup, we might be able to close: https://github.com/meteor/meteor/issues/11382

filipenevola:
Hi, we've reviewed this PR, the issue and the original post in the Forums.

We believe the solution should be as simple as @markdowney said:

> Simply querying for selector[fieldName] = fieldValue.toLowerCase(). If applied here, this change should only be opt-in as developers must make sure that all addresses were stored in lower-case in the first place.

As it would be opt-in, we could also save the email as lower case when the option is enabled.

Also, this topic seems to be not getting any attention now, so probably the issue is already fixed in the project itself and maybe we don't need a solution in the framework. If this problem comes up again, we already have the definition of how the solution should be.

This PR contains many changes in tests and other files, so that we will close it now.

PR #11423, opened 2021-05-12
Replace mongo-id meteor package with npm one

harryadel:
As discussed in https://github.com/meteor/meteor/issues/11404#issuecomment-834757231
I haven't edited meteor packages to use the NPM version because it's not published yet. Also, if this PR passes, would it be ok then to move the other [packages](https://github.com/harryadel/blaze-standalone) too? It'd help with the future Blaze standalone.


packages/deprecated/mongo-id/README.md
```diff
@@ -0,0 +1,4 @@
+# mongo-id
+[Source code of released version](https://github.com/meteor/meteor/tree/master/packages/mongo-id) | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/mongo-id)
```
StorytellerCZ, line 2: This will have to be changed.

npm-packages/mongo-id/src/id.js
```diff
@@ -0,0 +1,113 @@
...
+  if (firstChar === '~') {
+    return JSON.parse(id.substr(1));
+  }
+  if (MongoID._looksLikeObjectID(id)) {
+    return new MongoID.ObjectID(id);
+  }
+  return id;
+};
+
+export { MongoID };
```
StorytellerCZ, line 113: This export should be when the const is declared.

npm-packages/mongo-id/src/id.js
```diff
@@ -0,0 +1,113 @@
+import EJSON from 'ejson';
+import Random from '@reactioncommerce/random';
```
StorytellerCZ, line 2: In this case we should probably move Random to NPM as well.

StorytellerCZ (changes requested):

npm-packages/mongo-id/src/id.js
harryadel: You mean deprecate it and use reaction commerce version?

StorytellerCZ: Yeah, but it would probably make more sense if we move it under `@meteorjs` so that Reaction Commerce doesn't have to maintain it and it is really clear to everyone else what the main purpose it.

PR #11368, opened 2021-03-31
Support for publication strategies in livedata server

vlasky:
This PR was created in[ response to a request from @filipenevola](https://github.com/meteor/meteor/pull/11151#issuecomment-763541235) 

I have merged my last commit https://github.com/vlasky/meteor/commit/0299ce6865b9800374b179fa1f5f8333b91e2132 with the [current head of meteor/devel 0a0cae235812259414b5afd291ed7243ed37b997](https://github.com/meteor/meteor/commit/0a0cae235812259414b5afd291ed7243ed37b997).

Integrates [@sebakerckhof's code](https://github.com/meteor/meteor/pull/11151) with extracts from the client-side code in @mitar's package peerlibrary:control-mergebox.

It now provides client-side MongoDB compatibility when the publication strategy is set to NO_MERGE, by gracefully handling duplicate events without throwing an exception. Specific changes:

*When we receive an added message for a document that is already present in the client's collection, it will be changed
*When we receive a change message for a document that is not in the client's collection, it will be added
*When we receive a removed message for a document that is not in the client's collection, nothing will happen.

PR #11426, opened 2021-05-12
Convert @ddp-server/livedata_server Instances to Classes w/ Minor Perf Improvements

mvogttech:
This PR will cause conflicts with @vlasky [PR](https://github.com/meteor/meteor/pull/11368), however I believe the conversion to a class structure will allow for more efficient updates. 

I also think we should explore adding hash lookup support as a new strategy for @vlasky PR by modifying the diff-sequence package to have a method that dismisses the need for object-to-object comparisons in the callback, but rather hashes the string of the two objects and compares; if they differ have the server push the data to the client.

**Overview of changes**

- Migration to Classes from Prototype Instances
-- Session / session.js
-- ServerClass / server.js (Renamed to not conflict with global var)
-- SessionDocumentView / session_document_view.js
-- SessionCollectionView / session_collection_view.js
-- Subscription / subscription.js
- Modern ES Optimizations
- Conversion of Object instances to Maps


StorytellerCZ:
I think we should hold a bit on this after 2.3 release which has some changes in these as well, mainly when it comes to modernization some of those parts.

StorytellerCZ:
We'll also need to fix test before even considering this.

StorytellerCZ:
@mvogttech Can you get the changes in ` packages/diff-sequence/diff.js` into another PR? We could get that one out quickly.

mvogttech:
> @mvogttech Can you get the changes in ` packages/diff-sequence/diff.js` into another PR? We could get that one out quickly.

Of course. Working on that now.

filipenevola:
@mvogttech could you highlight the places where you are making changes so we can pay more attention to these lines?

As the PR is moving a lot of code it's hard for us to understand where the changes are.

You could comment in the changes here in the PR files tab ;)

filipenevola:
I'm closing this issue because we haven't heard anything from the original poster for a while.

PR #11427, opened 2021-05-12
diff-sequence 1.1.2 -- Convert DiffSequence.makeChangedFields Method to Utilize Map

mvogttech:
PR in response to [request](https://github.com/meteor/meteor/pull/11426#issuecomment-839971771) from @StorytellerCZ


StorytellerCZ (approved):

packages/diff-sequence/diff.js
```diff
@@ -11,277 +16,256 @@ function isObjEmpty(obj) {
...
-//    if unordered, they are IdMaps
-DiffSequence.diffQueryChanges = function (ordered, oldResults, newResults,
-                                              observer, options) {
-  if (ordered)
+/**
+ * @function DiffSequence
+ * @description old_results and new_results: collections of documents. if ordered, they are arrays. if unordered, they are IdMaps
+ */
+
+export default DiffSequence = {
```
afrokick, line 24: Why `export default`?

Maybe `export const DiffSequence = new class { };` ?

mvogttech: You're right. A named export/import will be better for code refactoring in the future... which is the whole point of the PR.

The latest commit also includes a performance optimization to the isObjEmpty method. You can see the performance difference [here](https://jsbench.me/plkon2676h/1).

mvogttech (commented):
I cannot understand why the TCI build is failing. Does anyone have any idea why?

`{"groupPath":["tinytest","livedata stub"],"test":"reconnect","events":[{"sequence":2310,"type":"fail","details":{"type":"assert_equal","expected":"{\"added\":2,\"removed\":0,\"changed\":3,\"moved\":0}","actual":"{\"added\":2,\"removed\":0,\"changed\":3,\"moved\":1}","not":false}}],"server":true}`

StorytellerCZ:
@mvogttech If you look closely on the message, the issue is here:
Expected: `moved: 0`
Actual: `moved 1`

mvogttech:
@StorytellerCZ -- Do you think this class should have a cache to memoize these iteration-intensive functions?

filipenevola:
@mvogttech same here, it would be nice to have comments in the places where you are making actual changes.

StorytellerCZ:
@mvogttech could you please fix the failing tests?

PR #11433, opened 2021-05-13
Add random npm package

harryadel:
Adding random NPM package as requested in https://github.com/meteor/meteor/pull/11423


filipenevola:
hi @harryadel  the idea would be to replace the usage inside Meteor core packages in a next version? 

We have more than 300+ usages of Random package inside Meteor core.

harryadel:
@filipenevola Surely, once it gets merged in and published under @meteor in NPM we can move on and replace random usage in meteor packages.

PR #11444, opened 2021-05-23
Meteor.publish() support for async functions

vlasky:
Meteor.publish() will now work correctly when passed an async function as the second argument.

My preliminary testing suggests that passing an async function to Meteor.publish() achieves the same benefits as executing [this.unblock() at the beginning of non-async publish function](https://github.com/meteor/meteor-feature-requests/issues/11).

This merges cleanly into [meteor:release-2.3](https://github.com/meteor/meteor/tree/release-2.3) as well.


vlasky:
Hmmm. the Travis CI build is failing on "Meteor.loggingIn() is false after login has completed".

We are not using the Meteor Accounts package in our Meteor app, which is otherwise working correctly with this code.

Assistance would be appreciated in determining why this is happening.

```
C: tinytest - accounts - Meteor.loggingIn() is true right after a login call : OK
C: tinytest - accounts - Meteor.loggingIn() is false after login has completed : !!!!!!!!! FAIL !!!!!!!!!!!
{"groupPath":["tinytest","accounts"],"test":"Meteor.loggingIn() is false after login has completed","events":[{"sequence":31,"type":"fail","details":{"type":"true","not":false},"cookie":{"name":"accounts - Meteor.loggingIn() is false after login has completed","offset":0,"groupPath":["tinytest","accounts"],"shortName":"Meteor.loggingIn() is false after login has completed"}}]}
C: tinytest - accounts - Meteor.loggingOut() is true right after a logout call : OK
C: tinytest - accounts - Meteor.loggingOut() is false after logout has completed : OK
C: tinytest - accounts - onLogin callback receives { type: "password" } param on login : OK
C: tinytest - accounts - onLogin callback receives { type: "resume" } param on reconnect, if already logged in : OK
Exception in delivering result of invoking 'login': JSHandle@error
```

packages/ddp-server/livedata_server.js
```diff
@@ -1085,53 +1085,57 @@ _.extend(Subscription.prototype, {
...
-      var collectionNames = {};
-      for (var i = 0; i < res.length; ++i) {
-        var collectionName = res[i]._getCollectionName();
-        if (_.has(collectionNames, collectionName)) {
-          self.error(new Error(
-            "Publish function returned multiple cursors for collection " +
-              collectionName));
+
+    //Supports a publish handler result from a normal or an async function
+    Promise.resolve(resMaybePromise).then(res => {
```
radekmie, line 1090: Such a change will make **all** publications async, which I think is not desired. The solution would be to either use `Promise.await` (but that relies on fibers), or use an optionally-asynchronous `then` (like the one I made [here](https://github.com/vazco/uniforms/blob/master/packages/uniforms/src/ValidatedForm.tsx#L185-L194), but without the `callIfAsync` option).

The type signature of such a `then` function is:
```ts
function then<T, U>(value: Promise<T>, fn: (value: T) => U): Promise<U>;
function then<T, U>(value: T, fn: (value: T) => U): U;
```

vlasky: @radekmie I strongly disagree. I tested the behaviour with both async and normal publish functions to confirm that the current behaviour of non-async publish function was retained.

vlasky: @radekmie this is one of my tests. On the server side I created these test publications **testa** and **testb**:

```
Meteor.publish('testa', async function() {
        console.log("Entered testa");
        Meteor._sleepForMs(5000);
        console.log("subscribing to testa");
        this.added('messages', Random.id(), { message: 'hello from test a' });
        this.ready();
});

Meteor.publish('testb', async function() {
        console.log("Entered testb");
        Meteor._sleepForMs(5000);
        console.log("subscribing to testb");
        this.added('messages', Random.id(), { message: 'hello from test b' });
        this.ready();
});
```

In a client side browser console, I ran both of these lines in one go:

```
a = Meteor.subscribe('testa');
b = Meteor.subscribe('testb');
```

When the publish functions are declared async, the server-side console prints:

```
I20210524-17:54:35.542(10)? Entered testa
I20210524-17:54:35.544(10)? Entered testb
I20210524-17:54:40.542(10)? subscribing to testa
I20210524-17:54:40.558(10)? subscribing to testb
```

This is the expected unblocked behaviour.

When the publish functions are not declared async, the server-side console prints:

```
I20210524-17:56:10.975(10)? Entered testa
I20210524-17:56:15.977(10)? subscribing to testa
I20210524-17:56:15.983(10)? Entered testb
I20210524-17:56:20.985(10)? subscribing to testb
```

This is the expected default blocked behaviour.

I agree that there must be some other subtle difference as something must be causing Travis CI to fail.

vlasky:
@StorytellerCZ is it possible that the Travis CI script is not waiting long enough before checking the result of Meteor.loggingIn()?

@radekmie was correct in that Promise.resolve's then() handler is called asynchronously, which means that the publish cursor result in the new code would only be executed on the next tick . Meteor's current code executes it on the same tick.

vlasky:
I have switched to a more conservative approach that guarantees 100% backward compatibility and now passes all tests.

*Results from a non-async publish handler function will be processed synchronously (the original way).
*Results from an async publish handler function (one that returns a promise and/or is thenable) will be resolved asynchronously.

StorytellerCZ:
Looks good! Can you add a note to history and a mention in docs?

filipenevola:
Hi @vlasky thank you for this work.

What do you think about having the same approach on Methods?

vlasky:
@filipenevola my understanding is that the current support for async functions in Meteor methods is robust. Is there any reason for it to be changed?

StorytellerCZ:
@vlasky I'm going to go ahead and merge this. If you could create a PR for documentation update that would be awesome!

filipenevola:
Hi @vlasky you are correct. I was just checking the details and everything seems to be handled correctly here https://github.com/meteor/meteor/commit/f5821c88eee587706eb8107f74de2f60c267807f

filipenevola:
Hi @vlasky I was coming here to ask you to add some tests but before that I decided to test your change and it was not working.

I created this PR with the fix https://github.com/meteor/meteor/pull/11480

PR #11445, opened 2021-05-24
UNIX_SOCKET_PERMISSIONS sets permissions on UNIX_SOCKET_FILE

vlasky:
UNIX_SOCKET_PERMISSIONS takes an octal value. If an invalid value is specified, an exception is generated.

e.g. when the following env variables are set:

```
UNIX_SOCKET_FILE=/tmp/meteor.sock
UNIX_SOCKET_PERMISSIONS=660
```

After the Meteor app is started, the socket file `/tmp/meteor.sock` is created and will be given user & group read-write permissions.

When the Meteor app is running:

`ls -l /tmp/meteor.sock`

displays something like

`srw-rw---- 1 meteor meteor 0 May 24 09:25 /tmp/meteor.sock`

For us, the use case is to allow the Meteor app to run under a different user id to the Nginx web server. Instead, Nginx is made a member of the Meteor app's group and can now be given group permission to write to the Meteor app's socket file.

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -1,5 +1,6 @@
 import assert from "assert";
 import { readFileSync } from "fs";
+import { chmodSync } from "fs";
 import { createServer } from "http";
 import {
   join as pathJoin,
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Looks good to me. Can you also create a PR for the documentation?

packages/webapp/webapp_server.js
```diff
@@ -1,5 +1,6 @@
 import assert from "assert";
 import { readFileSync } from "fs";
+import { chmodSync } from "fs";
```
harryadel, line 3: import `readFileSync` and `chmodSync` in the same line?

vlasky: Thanks, I will make that change.

vlasky:
@StorytellerCZ I have implemented another enhancement. The optional environment variable UNIX_SOCKET_GROUP can be used to set any permitted UNIX gid or group name on the socket file.

This was actually a difficult feature to implement. Node.js does not have a native API function to convert UNIX group names to gids. The various fs.chown() functions only accept numericals gids as arguments.

After a lot of searching, I eventually found an excellent NPM named whomst that implements this functionality in a highly compatible manner that will work across different platforms and environments.

[I forked and updated whomst](https://github.com/vlasky/whomst) to work with Node.js 12 and 14. It is now available on NPM as [@vlasky/whomst](https://www.npmjs.com/package/@vlasky/whomst).

Should I add it to this PR or a new one?

StorytellerCZ:
@vlasky Better to keep it separate so that it is easier to review and hence quicker to merge unless you want it to supersede this PR.

vlasky:
@StorytellerCZ ok I'll open a new PR. Can we merge this one?

vlasky:
> Looks good to me. Can you also create a PR for the documentation?

I have created the PR.

PR #11451, opened 2021-05-28
UNIX_SOCKET_GROUP sets UNIX group/gid of UNIX_SOCKET_FILE

vlasky:
Environment variable UNIX_SOCKET_GROUP can now be used to override the default UNIX group name/gid on the socket file specified in UNIX_SOCKET_PATH.


vlasky:
You can set UNIX_SOCKET_GROUP to either a numerical gid or a group name. The code will automatically handle both.

Supporting group names was a bit of a challenge. Node.js does not have a native API function to convert UNIX group names to gids. The various fs.chown() functions only accept numerical gids as arguments.

After a lot of searching, I eventually found an excellent NPM named whomst that implements this functionality in a highly compatible manner that will work across different platforms and environments.

[I forked whomst](https://github.com/vlasky/whomst), made some updates, and tested it working with Node.js 12 and 14. I expect it to also work with Node.js 16. It is now available on NPM as [@vlasky/whomst](https://www.npmjs.com/package/@vlasky/whomst).

vlasky:
@StorytellerCZ I have created a PR for the docs:

https://github.com/meteor/docs/pull/724

packages/webapp/webapp_server.js
```diff
@@ -1205,6 +1208,25 @@ function runWebAppServer() {
         }
       }
 
+      let unixSocketGroup = process.env.UNIX_SOCKET_GROUP;
```
filipenevola, line 1211: I believe we don't need any let on this code. const + one new variable should be enough.

filipenevola (changes requested):
Hi @vlasky , thank you for the work.

I'm just recommending a small code change so we can merge your PR.

packages/webapp/webapp_server.js
vlasky: If this is changed to const then the trim() operation on line 1213 would not work.

vlasky: Would you be happier with this?

`const unixSocketGroup = (env.UNIX_SOCKET_GROUP || "").trim();`

vlasky: I noticed that `whomst.sync.group()` can take both numerical gids and group names as an argument, so I was able to make the code more concise and eliminate all the lets, leaving one const.

filipenevola (approved):
❤️

PR #11469, opened 2021-06-08
when #each argument is unsupported, show it for better debug experience

antoinep92:
Fixes meteor/blaze#338

### DIFF
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -182,13 +182,18 @@ function toDebugStr(value, maxLength) {
   }
 }
 
-const badSequenceError = function (sequence) {
-  let gotValue = '';
+function sequenceGotValue(sequence) {
   try {
-    gotValue = ' Got ' + toDebugStr(sequence)
-  } catch(e) {}
+    return ' Got ' + toDebugStr(sequence);
+  } catch(e) {
+    return ''
+  }
+}
+
+const badSequenceError = function (sequence) {
   return new Error("{{#each}} currently only accepts " +
-                   "arrays, cursors or falsey values." + gotValue);
+                   "arrays, cursors or falsey values." +
+                   sequenceGotValue(sequence));
 };
 
 const isStoreCursor = function (cursor) {
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -111,7 +111,7 @@ ObserveSequence = {
           seqArray = result[0];
           activeObserveHandle = result[1];
         } else {
-          throw badSequenceError();
+          throw badSequenceError(seq);
         }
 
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Hi @antoinep92 thanks for the PR. Can you please sign the CLA or try to [re-check](https://cla-assistant.io/check/meteor/meteor?pullRequest=11469) so that we can move forward on this? Thanks a lot!

antoinep92:
> Hi @antoinep92 thanks for the PR. Can you please sign the CLA or try to [re-check](https://cla-assistant.io/check/meteor/meteor?pullRequest=11469) so that we can move forward on this? Thanks a lot!

Github was partially down earlier, so I was not able to sign it, but it is ok now

antoinep92:
Hi @StorytellerCZ, I'm not entirely clear on the Meteor review process. I saw you were listed as a reviewer, so does it mean we are good, or is some action expected from me at this point (like ping former contributor to the file I changed for feedback, etc.) ?

packages/observe-sequence/observe_sequence.js
```diff
@@ -140,14 +140,55 @@ ObserveSequence = {
...
       return seq.fetch();
     } else {
-      throw badSequenceError();
+      throw badSequenceError(seq);
     }
   }
 };
 
-var badSequenceError = function () {
+var ellipsis = function (longStr, maxLength) {
```
filipenevola, line 148: As we are adding new code here we should use new patterns, like const here instead of var.

Also everywhere below this point.

packages/observe-sequence/observe_sequence.js
```diff
@@ -140,14 +140,55 @@ ObserveSequence = {
...
+        return value.constructor.name + ' '
+             + ellipsis(JSON.stringify(value), maxLength);
+      }
+    default:
+      return type + ': ' + value.toString();
+  }
+}
+
+var badSequenceError = function (sequence) {
+  var gotValue = '';
```
filipenevola, line 186: a function would be better here to isolate this part.

like: `getDetails(sequence)` so the try/catch would be within the function.

filipenevola (changes requested):
Hi @antoinep92 thank you for your work.

I just suggested a few small changes.

packages/observe-sequence/observe_sequence.js
antoinep92: I usually try to stick to the existing coding style, but you're right

StorytellerCZ:
Will release as a patch and then move to Blaze repository.

antoinep92:
Ok thanks, do I need to do something for that to happen, or will you handle it from here ?

StorytellerCZ:
@antoinep92 I'll handle it.

filipenevola (approved):
❤️

PR #11472, opened 2021-06-10
add hook to allow custom user selection in Accounts.updateOrCreateUserFromExternalService

jb-1980:
## The problem
When using an external service, there are times when a developer wants to merge the account instead of creating a new user. Various packages and workarounds have been created and suggested to handle this:
* [merge-accounts](https://atmospherejs.com/rkstar/accounts-merge) - Creates two users and decides which to use. Old package.
* [delete user in `Accounts.createUser` hook, and return identical user to be created](https://forums.meteor.com/t/accounts-password-merge-with-accounts-google-2017-way/35590) - Has bad side effect of an error being thrown in the `insertUserDoc` and now no user exists.

## This solution
Provide another login hook, `selectCustomUserOnExternalLogin`, that takes a callback to allow the developer to decide how to determine if a user exists based on the `serviceData` provided by the external service.

### API
```javascript
Accounts.selectCustomUserOnExternalLogin(callback)
```
##### Arguments
Callback: function, Called with same arguments passed to `Accounts.updateOrCreateUserFromExternalService`: `serviceName`, `serviceData`, `options`.
##### Return
Should return the custom user, or undefined if no user found.

### Example
```javascript
Accounts.selectCustomUserOnExternalLogin((serviceName, serviceData) => {
  if (serviceName === "google") {
    return Accounts.findUserByEmail(serviceData.email)
  }
})
```

Or, developer could throw error ...
```javascript
Accounts.selectCustomUserOnExternalLogin((serviceName, serviceData) => {
  if (serviceName === "google") {
    const user = Accounts.findUserByEmail(serviceData.email)
    if (!user) throw new Error("The email is not registered")
    return user
  }
})
```

I think this PR fits the two main design requirements:
* It doesn't harm the experience of a new Meteor developer. The hook doesn't change any existing functionality, so there is nothing more required of new Meteor developers.
* It doesn't preclude an expert from doing what they want. In fact, it let's experts do more.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -190,11 +190,11 @@ export class AccountsServer extends AccountsCommon {
    * @param {Function} func Called whenever a user is logged in via oauth and a
    * user is not found with the service id. Return the user or undefined. 
    */
-  selectCustomUserOnExternalLogin(func) {
-    if (this._selectCustomUserOnExternalLogin) {
-      throw new Error("Can only call selectCustomUserOnExternalLogin once");
+   setAdditionalFindUserOnExternalLogin(func) {
+    if (this._setAdditionalFindUserOnExternalLogin) {
+      throw new Error("Can only call setAdditionalFindUserOnExternalLogin once");
     }
-    this._selectCustomUserOnExternalLogin = func;
+    this._setAdditionalFindUserOnExternalLogin = func;
   }
 
   _validateLogin(connection, attempt) {
@@ -1271,8 +1271,8 @@ export class AccountsServer extends AccountsCommon {
 
     // Check to see if the developer has a custom way to find the user outside
     // of the general selectors above.
-    if (!user && this._selectCustomUserOnExternalLogin) {
-      user = this._selectCustomUserOnExternalLogin(serviceName, serviceData, options)
+    if (!user && this._setAdditionalFindUserOnExternalLogin) {
+      user = this._setAdditionalFindUserOnExternalLogin({serviceName, serviceData, options})
     }
 
     // Before continuing, run user hook to see if we should continue
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -184,6 +184,19 @@ export class AccountsServer extends AccountsCommon {
     this._onExternalLoginHook = func;
   }
 
+  /**
+   * @summary Customize user selection on external logins
+   * @locus Server
+   * @param {Function} func Called whenever a user is logged in via oauth and a
+   * user is not found with the service id. Return the user or undefined. 
+   */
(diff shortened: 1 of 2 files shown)


StorytellerCZ:
There is [link-accounts](https://github.com/Meteor-Community-Packages/meteor-link-accounts) maintained by MCP.

jb-1980:
> There is link-accounts maintained by MCP.

@StorytellerCZ I was really hopeful with that package! But there is a slight variation in the user story I am trying to solve. Namely, you can't use the package for authentication.

In my particular use case, I have around 4000 students who have been added through another service, and their `user.email.address` field uses a Google for Education email. The intent is that the students authenticate with Google. Their accounts have no password to sign in with.

The `Meteor.loginWithGoogle` seemed so promising and simple, but it turns out it wasn't because it is actually closer to `loginOrRegisterWithGoogle`. So again, it is an ever so slight variation in the user story.

I have spent some time looking over various solutions and trying different patterns. And while banging my head against the code I found myself saying "All I want is a way to tell the code to not look for the `service.id`, but the `email.address`!", and "Why does this design pattern preclude me from doing what I want?"

I didn't really say the last part, but took it from the CONTRIBUTING.md to drive home my point. And my point is the current architecture has a design gap that causes entire packages to be created and multiple forum posts filled with suggestions trying to solve the problem of "How do I tell the app to not create a new account, but merge with this one?".

I think this is a really simple, elegant solution to that problem in less than 10 lines of code. So I really hope you can give it some serious consideration and merge it in.

filipenevola:
hi @jb-1980 we really want to add more common use cases to accounts API so it's easier to handle cases that you expect a simple solution.

I'm not 100% sure if `selectCustomUserOnExternalLogin` would be the best name here tho. In this case it's not clear that it will be one additional find to try to find the user again. It seems to be a replacement.

Maybe it could be `setAdditionalFindUserOnExternalLogin` and also receive more data using a simple object argument with fields such as:
- serviceName
- serviceData
- options

so we could easily add more fields later without breaking the API or relying on a specific order.

What do you think?

filipenevola:
@jb-1980 just to avoid more back and forth, it's also important to send a https://github.com/meteor/docs showing how to use the new API, it also helps in the review so we can understand your idea and use case in full.

jb-1980:
@filipenevola Thanks for the good feedback. I have updated the names and arguments as you suggested. And have also submitted a PR to the documentation.

Looks like the CI checks are currently failing, and the logs indicate that it is an issue with isopackets in hot-module-replacement. So I am hoping it is something unrelated to these changes and will work itself out.

StorytellerCZ:
All tests are passing now.

filipenevola (approved):
❤️

filipenevola:
@jb-1980 thanks.

please provide a PR for the docs as well.

PR #11474, opened 2021-06-11
Track which input files are used by build plugins

zodern:
Another attempt at #10455 and #11381.

Tracks which files are used by build plugins to know when it should do a full rebuild, a faster client-only rebuild, or can completely skip rebuilding after a file is modified. This should work with any type of file in any directory, and for both files in the app and files in packages. The most noticeable improvement is when modifying a file only used on the client Meteor will only rebuild the client, even if the file is not inside `imports` or a `client` folder.

These changes don't reduce how many files are watched, only what Meteor does when one of them is modified.

Most build plugins should work correctly with no modifications, or work correctly but make Meteor think every file is always used. Build plugins that read the content of input files without using the file's api will need to be updated to tell Meteor that the content of the file is used.

## Changes for build plugins

To track which files are used by build plugins, Meteor records if a build plugin accesses an input file's hash or content. Some of the methods that access either are:
- file.getContentsAsBuffer
- file.getContentsAsString
- file.getSourceHash

So the file is not unnecessarily marked as used, the build plugin should try to move calls to those functions to be inside of the lazy finalizer passed when adding a resource (or when using CachingCompiler, within `compileOneFile` and `getCacheKey`), which Meteor calls once it knows the file will be used. For example, instead of this:

```js
const hash = inputFile.getSourceHash();
inputFile.addJavaScript({
  path: inputFile.getPathInPackage(), 
  hash
}, () => {
  return {
    data: COMPILED_CONTENT
  }
});
```
You can move the call to `inputFile.getSourceHash()` to within the lazy finalyzer:
```js
inputFile.addJavaScript({
  path: inputFile.getPathInPackage()
}, () => {
  const hash = inputFile.getSourceHash();
  return {
    hash,
    data: COMPILED_CONTENT
  }
});
```

As before, additional files can be watched/marked as used by calling either of these functions with a file's path:

- file.readAndWatchFile
- file.readAndWatchFileWithHash

What is left:
- [ ] It might crash when loading a package published before Meteor 1.2. I am still trying to understand how Meteor handles these packages


filipenevola (approved):
This solution looks great. ❤️

PR #11476, opened 2021-06-11
Replace npm-mongo with mongodb

harryadel:

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -8,8 +8,19 @@
  */
 
 const path = require("path");
+const NpmModuleMongodbVersion = require("mongodb/package.json").version;
+const oldNoDeprecationValue = process.noDeprecation;
+let MongoDB = null;
 
-var MongoDB = NpmModuleMongodb;
+try {
+  // Silence deprecation warnings introduced in a patch update to mongodb:
+  // https://github.com/meteor/meteor/pull/9942#discussion_r218564879
+  process.noDeprecation = true;
+  MongoDB = require("mongodb");
+} finally {
+  process.noDeprecation = oldNoDeprecationValue;
+}
+
 var Future = Npm.require('fibers/future');
 import { DocFetcher } from "./doc_fetcher.js";
 
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -1,7 +1,6 @@
 var Future = Npm.require('fibers/future');
 
-import { NpmModuleMongodb } from "meteor/npm-mongo";
-const { Long } = NpmModuleMongodb;
+import {Long} from "mongodb";
 
 OPLOG_COLLECTION = 'oplog.rs';
 
--- a/packages/mongo/package.js
+++ b/packages/mongo/package.js
@@ -13,7 +13,8 @@ Package.describe({
 });
 
 Npm.depends({
-  "mongodb-uri": "0.9.7"
+  "mongodb-uri": "0.9.7",
+  "mongodb": "3.6.9"
 });
 
 Npm.strip({
@@ -21,7 +22,6 @@ Npm.strip({
 });
 
 Package.onUse(function (api) {
-  api.use('npm-mongo', 'server');
   api.use('allow-deny');
 
   api.use([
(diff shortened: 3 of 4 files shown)


StorytellerCZ:
Best to add the deprecated flag to `npm-mongo` in this PR.

harryadel:
@StorytellerCZ How do I add `mongodb` npm library to the Meteor tools? I added it to `scripts/dev-bundle-tool-package.js` but it seems to have no effect. Is there a command I must run for it to be added?

StorytellerCZ:
You will need to generate new dev bundle. So for now you should generate the dev bundle locally and run the tests. I'll do it for you later here.

harryadel:
@StorytellerCZ will you please push a new bundle version here?

StorytellerCZ:
@harryadel Done!

harryadel:
@StorytellerCZ Thank you :)

harryadel:
@StorytellerCZ I guess it's good to go :tada:

tools/runners/run-mongo.js
```diff
@@ -636,7 +636,7 @@ var launchMongo = function (options) {
       const {
         MongoClient,
         Server
-      } = loadIsopackage('npm-mongo').NpmModuleMongodb;
+      } = require('mongodb');
```
filipenevola, line 639: the old version was suppressing warnings and here you are importing directly.

packages/mongo/package.js
```diff
@@ -13,15 +13,15 @@ Package.describe({
 });
 
 Npm.depends({
-  "mongodb-uri": "0.9.7"
+  "mongodb-uri": "0.9.7",
+  "mongodb": "3.6.9"
```
filipenevola, line 17: I'm not sure if this is not going to included mongodb in the client as well. Are you sure?

packages/mongo/oplog_tailing.js
```diff
@@ -1,7 +1,6 @@
 var Future = Npm.require('fibers/future');
 
-import { NpmModuleMongodb } from "meteor/npm-mongo";
-const { Long } = NpmModuleMongodb;
+import {Long} from "mongodb";
```
filipenevola, line 3: the old version was suppressing warnings and here you are importing directly.

filipenevola (changes requested):
See my comments.

packages/mongo/mongo_driver.js
```diff
@@ -8,8 +8,19 @@
...
 
 const path = require("path");
+const NpmModuleMongodbVersion = require("mongodb/package.json").version;
+const oldNoDeprecationValue = process.noDeprecation;
+let MongoDB = null;
+
+try {
+  // Silence deprecation warnings introduced in a patch update to mongodb:
+  // https://github.com/meteor/meteor/pull/9942#discussion_r218564879
+  process.noDeprecation = true;
```
harryadel, line 18: The error suppression is still being done @filipenevola

packages/mongo/package.js
harryadel: What client?

StorytellerCZ: @harryadel on the app client I think what @filipenevola was meaning here.

harryadel: I'm unsure whether `mongodb` library gets shipped to the client side or not but if it's a concern shouldn't it be a global one? ddp-server package for instance import sockjs which is the server counter part of sockjs-client
https://github.com/meteor/meteor/blob/6f7df8f35f1617b393a1d13c48f2bf70c2be31f6/packages/ddp-server/package.js#L9

StorytellerCZ:
@harryadel let's get this ready for 2.5! Can you please resolve the conflicts?

harryadel:
@StorytellerCZ will you please do the bundle thing?

StorytellerCZ:
@filipenevola what do you think now?

filipenevola:
> @filipenevola what do you think now?

I still think this replacement is changing the behavior.

Before the MongoDB objects where always imported using the package and so forcing the warnings to be suppressed.

Also the mongo package was importing another package and in this way we can control if it affects only the server or only the client. Depending directly from NPM I'm not sure if this is possible.

So, TL;DR, we need manual tests and the evidences here in the PR to be if my suspicious above are incorrect before proceeding with this PR.

packages/mongo/package.js
harryadel: No, scratch that. @filipenevola has a point.

`ddp-server` as a whole is set to be used in server only 
https://github.com/meteor/meteor/blob/ffcfa5062cf1bf8a64ea64fef681ffcd99fe7939/packages/ddp/package.js#L8

PR #11477, opened 2021-06-14
add support for arbitrary iterables in #each templates

### LINKED ISSUE #339: Corrupted application state from stale connections
This is a cross-post from http://stackoverflow.com/questions/12395387/how-can-meteor-handle-stale-connections

How to reproduce the issue:
1. Open the same app (from a remote server) in two browser windows.
2. Turn off wi-fi (or 3g) - connection becomes stale because it was not closed from the remote server.
3. Perform a write operation in the first tab (the second tab will of course not update reactively).
4. Turn on wi-fi.
5. Perform a second write in the first tab (the second tab will now update reactively)
6. So the first tab is now in a corrupted state. The first write will never sync to the server, and if the page reloads, that change will be permanently lost.

Not sure if Meteor can already handle this scenario, or should this be a "data validation" feature request?

antoinep92:
Resolves meteor/blaze#339
As discussed in the issue, this extends the range of supported {{#each}} arguments to any object implementing the iterable protocol except `string`s. To minimize changes and risks, these arguments are converted to Arrays, so the iteration mechanics remains unchanged.
IE is not supported.

### DIFF
--- a/packages/observe-sequence/observe_sequence.js
+++ b/packages/observe-sequence/observe_sequence.js
@@ -18,6 +18,17 @@ function isArray(arr) {
   return arr instanceof Array || _.isArray(arr);
 }
 
+// isIterable returns trues for objects implementing iterable protocol,
+// except strings, as {{#each 'string'}} doesn't make much sense.
+// Requires ES6+ and does not work in IE (but degrades gracefully).
+// Does not support the `length` + index protocol also supported by Array.from
+function isIterable (object) {
+  const iter = typeof Symbol != 'undefined' && Symbol.iterator;
+  return iter
+    && object instanceof Object // note: returns false for strings
+    && typeof object[iter] == 'function'; // implements iterable protocol
+}
+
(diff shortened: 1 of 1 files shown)


antoinep92:
*Note:* as far as I can tell, build is failing because of upstream commit 2d41716645c75c5bc2ef37f306ef87c00b982d16 and not because of my change. Locally I get the same error as the build bot if I start from `devel` tip, but works fine if I exclude this commit. Ping @StorytellerCZ

StorytellerCZ:
Looking into this.

StorytellerCZ:
Given #11469 this is a new feature and hence would require feature bump I think this should be handled once we move observe-sequence to Blaze and have it be part of the next Blaze release.

StorytellerCZ:
Closing this as observe-sequence has been moved to [Blaze](https://github.com/meteor/blaze/) repository, please re-open this PR there.

PR #11478, opened 2021-06-14
New deploy option: build-only

filipenevola:
Some clients want to have the ability to build first and after some validations proceeding with the upload and deploy.

### DIFF
--- a/tools/meteor-services/deploy.js
+++ b/tools/meteor-services/deploy.js
@@ -401,9 +401,7 @@ class PollingState {
 async function pollForDeploy(pollingState, versionId, site) {
   const {
     deadline,
-    initialWaitTimeMs,
     pollIntervalMs,
-    start,
     currentMessage,
   } = pollingState;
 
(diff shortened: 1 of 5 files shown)


tools/meteor-services/deploy.js
```diff
@@ -481,45 +481,59 @@ export async function bundleAndDeploy(options) {
...
-  // or is revoked, user comes back to deploy again. In that case,
-  // they'll get an email prompt instead of a username prompt because
-  // the command-line tool didn't have time to learn about their
-  // username before the credential was expired.
-  pollForRegistrationCompletion({
-    noLogout: true
-  });
-  var promptIfAuthFails = (loggedInUsername() !== null);
+  if (options.isBuildOnly) {
+    Console.info('Skipping pre authentication as the option --build-only was informed.');
```
StorytellerCZ, line 489: `was informed` shouldn't this be `was provided`?

tools/meteor-services/deploy.js
```diff
@@ -609,6 +623,13 @@ export async function bundleAndDeploy(options) {
     });
   }
 
+  if (options.isBuildOnly) {
+    Console.info(
```
StorytellerCZ, line 627: I think it would be better to use `Console.success` here to make it clear that the process has finished.

StorytellerCZ (changes requested):

tools/meteor-services/deploy.js
filipenevola: As we are not using success anywhere in the deploy maybe it's better to just keep as info.

StorytellerCZ: We are using error, but let's shelve this for a bigger overview then.

StorytellerCZ (approved):

PR #11480, opened 2021-06-14
Fixes #11444 providing the correct context for this on _publishHandlerResult

filipenevola:
In this PR https://github.com/meteor/meteor/pull/11444 the support for async functions on publish was still not working.

Pending:
- [ ] Automatic Tests


packages/ddp-server/livedata_server.js
```diff
@@ -1076,16 +1078,20 @@ Object.assign(Subscription.prototype, {
...
+    if (self._isDeactivated()) return;
+
+    // Both conventional and async publish handler functions are supported.
+    // If an object is returned with a then() function, it is either a promise
+    // or thenable and will be resolved asynchronously.
+    const isThenable =
+      resultOrThenable && typeof resultOrThenable.then === 'function';
+    if (isThenable) {
+      Promise.resolve(resultOrThenable).then(
+        (...args) => self._publishHandlerResult.bind(self)(...args),
```
filipenevola, line 1090: @vlasky this is the fix.

without this change it was not working for async functions, the this was Meteor instead of the Subscription.

vlasky: @filipenevola belated thanks for sorting that out. For some reason, when I tested with our MySQL-based publications, they  worked in async mode without the bind().

PR #11484, opened 2021-06-22
Workaround for Safari issue with indexedDB

### LINKED ISSUE #11479: Initial lazy loading (caching to IDB) on Safari doesn't work. Starts working after few attempts to refresh the page
When you load a web app on Safari (Mac OS) for the first time (incognito mode), lazy components doesn't load. After few refresh attempts it starts loading.

It's certainly a bug in Safari, but I believe Meteor should handle this edge case.

The problem occurs in `dynamic-import/cache.js` when `idb.open` is called. The process hangs and doesn't call `onsuccess` neither `onerror`. Rejecting after some period of time could solve the issue.

I was trying to reproduce the same behaviour just by calling `indexedDB.open('test')` in `client/main.jsx`. But funny enough then everything works perfectly, including `MeteorDynamicImportCache`.

Here's a simple project where it can be reproduced running with --production flag.
https://github.com/ksinas/lazy-load-debug

I added dynamic-import package with some logs which indicates the problem

StorytellerCZ:
Should close #11479


rj-david:
When we tried this solution, we got hundreds of errors in production for non-existing `databases()` function. There is a need to check if the function exists or we might need to use a more universal indexedDB function

StorytellerCZ:
@rj-david no worries, I will be improving this today. Though any help would be appreciated since I'm not a Mac user.

packages/dynamic-import/client.js
```diff
@@ -1,22 +1,44 @@
...
 var meteorInstall = require("meteor/modules").meteorInstall;
 var dynamicVersions = require("./dynamic-versions.js");
 
+// Fix for Safari bug https://bugs.webkit.org/show_bug.cgi?id=226547
+function idbReady() {
+  var isSafari = !navigator.userAgentData &&
+    /Safari\//.test(navigator.userAgent) &&
+    !/Chrom(e|ium)\//.test(navigator.userAgent);
+  // No point putting other browsers through this mess.
+  if (!isSafari || !window.indexedDB.databases) return Promise.resolve();
```
renanccastro, line 13: It's probably safer to check if indexedDB also exists:
!window.indexedDB || !window.indexedDB.databases

StorytellerCZ: Fixed!

StorytellerCZ:
This fixes the issue for dynamic imports, but if there is something else that uses indexedDB before it comes here it will run into the same issue.

StorytellerCZ:
Closing in favor of #11512

PR #11501, opened 2021-06-28
Update babel runtime and skelets

StorytellerCZ:
* Updates babel runtime across the code base to use the same version.
* Update skeleton dependencies to be consistent (same React version, etc.).
* In Apollo skelet removed `apollo-boost` which is no longer needed.

### DIFF
--- a/History.md
+++ b/History.md
@@ -14,8 +14,22 @@
 
 * `meteor-tool@2.3.1`
   - Node.js updated to 14.17.2
+  - `@babel/runtime` dependency updated to v7.14.6 across the tool and testing apps
+  - Skeletons dependencies updated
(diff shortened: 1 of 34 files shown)


StorytellerCZ:
To fully test this, this will require new dev bundle. I will add the changelog to all the affected packages as well. This can go in either 2.4 or 2.3.1. Thoughts @filipenevola @renanccastro ?

filipenevola:
> To fully test this, this will require new dev bundle. I will add the changelog to all the affected packages as well. This can go in either 2.4 or 2.3.1. Thoughts @filipenevola @renanccastro ?

2.3.1 is fine @StorytellerCZ.

Issue #11402, opened 2021-04-29
Allow dynamic root url on cordova


crapthings:
meteor build ../youcordovabuild --directory
"Supply the server hostname and port in the --server option for mobile app builds."

> https://github.com/meteor/meteor/blob/devel/packages/boilerplate-generator/template-web.cordova.js#L39

optional ROOT_URL with cordova build

load default ROOT_URL with localStorage, fallback to built one if not

StorytellerCZ:
Hi @crapthings , I'm not sure I follow what you want done here. Do you want to be able to set `ROOT_URL` which then would take the place of `--server` flag if it was not defined?

crapthings:
our cordova app can input any ddp server url, and we have a custom package that replace root_url.
and this package load before ddp connection init so we can replace ROOT_URL, but i think its better support with meteor itself.


```js
export const name = 'ddp-server-picker';
if (Meteor.isClient) {
  const rootURL = window.localStorage.getItem('__root_url');

  if (rootURL) {
    __meteor_runtime_config__.DDP_DEFAULT_CONNECTION_URL = rootURL;
    __meteor_runtime_config__.ROOT_URL = rootURL;
  }
  
  console.log('__meteor_runtime_config__:', __meteor_runtime_config__);
}
```

![image](https://user-images.githubusercontent.com/1147704/116530500-b10e3000-a910-11eb-8a70-d9619911ce55.png)

crapthings:
> Hi @crapthings , I'm not sure I follow what you want done here. Do you want to be able to set `ROOT_URL` which then would take the place of `--server` flag if it was not defined?

yes and make server url optional on build，have api change ddp connection，so when user reopen app they can use user specify root url，but not build one

crapthings:
hot code push use root url from header if you dont override it. you might have infiniti reload if you meteor.connect to other

StorytellerCZ:
@crapthings Since you already have a working code I would say that the best approach would be to go ahead and create a PR.

brucejo75:
Hey @crapthings,

Check out #11408, integrated with **Meteor 2.3**.

This PR fixes an issue with `cordova` not picking up the value of `DDP_DEFAULT_CONNECTION_URL`.  This will give you a lot more control of the DDP connection and the root connection for all platforms.

This is an undocumented feature of Meteor...

https://github.com/meteor/docs/pull/716 documents the feature after Meteor 2.3 releases.

[...]

PR #11506, opened 2021-06-30
Hook __meteor_runtime_config__ at runtime

brucejo75:
Hello!

I developed this PR in response to https://github.com/meteor/meteor/issues/11402.  However, I think the PR has general applicability and I have a use for it in my work.

## Motivation
When a meteor app is constructed there are a number of variables that are encoded in the `__meteor_runtime_config__` object.

There are many ways to configure the elements of `__meteor_runtime_config__` at build time and during development via environment variables, etc.  This works fine in many cases but begins to be a problem area when trying to use the same meteor server for multiple meteor apps or trying to have multiple domain origins for the same meteor server.  In those cases, it is valuable to enable configuring `__meteor_runtime_config__` at runtime.

In some cases, it is impossible without being able to configure `__meteor_runtime_config__` at runtime.

A good example is `ROOT_URL`.  Each meteor server has `1  ROOT_URL`.  So if multiple domain origins route to a single meteor server you are forced to "hack" a solution on the client side (e.g. you can set `ROOT_URL` to be equal to `location.origin` on initialization).  This "works" for webapps, but it does not work for `cordova` apps.  Here is the problem:
- server has `ROOT_URL` of `domain1.com`
- browser client1 has origin of `domain1.com`.  All works!
- browser client2 has origin of `domain2.com`, no worries just hack `ROOT_URL` on client to be `domain2.com` at initialization.
- cordova client1 has origin of `domain1.com`, `cordova-plugin-meteor-webapp` cracks open the header and gets its `ROOT_URL` from the index file loaded from the domain get.  In this case it is `domain1.com` and all works.
- 🔴 cordova client2 has origin of `domain2.com`, `cordova-plugin-meteor-webapp` cracks open the header and gets its `ROOT_URL` from the index file loaded from the domain get.  In this case it is `domain1.com`,  Now `cordova-plugin-meteor-webapp` compares the `ROOT_URL` to the value that is in the cordova package and it errors out.

## Solution: Dynamic Runtime Config Hook
I decided this was a good opportunity for a hook function:
- People that may configure this are usually attempting solutions outside of the main path for Meteor.
- These cases are maybe not that common?  (But valuable nonetheless!)
 
**Suggested documentation to add to the `README.md`**:
# Dynamic Runtime Config Hook
**NOTE:** this feature only works if you allow `inline eval`.

In some cases it is valuable to be able to control the `__meteor_runtime_config__` variable that initializes Meteor.

To work with the config object you will need to set up a runtime config hook callback.  Within the callback you will need to:
1. Decode the encoded string to an object.
2. Make your changes to the object.
3. Encode the updated config object to an encoded config string.
4. Return the encoded config string.

Example:

```javascript
WebApp.addRuntimeConfigHook((arch, request, current, updated) => {
  // check the request to see if this is a request we want to change
  if(req.host === 'calling.domain') {
    // make changes to the config for this domain
    // convert current to an object
    const config = WebApp.decodeRuntimeConfig(current);
    // make your changes
    config.newVar = 'some value';
    config.oldVar = 'new value';
    return WebApp.encodeRuntimeConfig(config);
  }
  return undefined;
});
```

### `WebApp.encodeRuntimeConfig(config_object)`
`encodeRuntimeConfg` will convert a config object to the encoded string that is placed in `index.html`.

### `WebApp.decodeRuntimeConfig(encoded_config_string)`
`encodeRuntimeConfg` will convert an encoded config string to an object.

### `WebApp.addRuntimeConfigHook(callback)`
This sets a callback that will get called for every request for a new `index.html`.

You will get a callback with one object argument: `{arch, request, encodedCurrentConfig, updated}`.

- `arch` - is one of `web.browser`, `web.browser.legacy` or `web.cordova`.
- `request` - is the web request to meteor.
- `encodedCurrentConfig` - is the encoded config string.
-  `updated` - `true` if the config for this architecture been updated since last called, otherwise `false`.  (This is a good flag to have so the caller can cache the result of decoding/encoding for each architecture.)

Your callback returns the desired new value of `encoded_config_string`.  Returning a falsy value will not change the current config.


StorytellerCZ:
@brucejo75 thanks for the extensive description! It is not something I have personally a use case for, but I'll take a look at it after 2.3.1 release. Right now considering this for inclusion in 2.4.

brucejo75:
Thanks @StorytellerCZ,

I would also like a discussion on my implementation.  I chose the path of least modifications and fastest execution.

It is a little weird to deal with the encoded runtime config, but it has the advantage that the hook has the least impact and the performance hit is minimal if the caller caches the response.

packages/webapp/webapp_server.js
```diff
@@ -345,8 +345,33 @@ function getBoilerplate(request, arch) {
...
+
+WebApp.addRuntimeConfigHook = function (hook) {
+  if(typeof hook !== 'function') throw new Error('WebApp.addRuntimeConfigHook must be a function');
+  runtimeConfig.hooks.push(hook);
+}
+
 function getBoilerplateAsync(request, arch) {
-  const boilerplate = boilerplateByArch[arch];
+  let boilerplate = boilerplateByArch[arch];
+  _.each(runtimeConfig.hooks, function (hook) {
```
filipenevola, line 368: We are avoiding `underscore`. This would be better
```suggestion
runtimeConfig.hooks.forEach(function (hook) {
```

packages/webapp/webapp_server.js
```diff
@@ -345,8 +345,33 @@ function getBoilerplate(request, arch) {
...
+    const meteorRuntimeConfig = hook(arch, request, boilerplate.baseData.meteorRuntimeConfig, runtimeConfig.update[arch]);
```
filipenevola, line 369: I would suggest a single object param here so next changes would be easier without breaking changes due to the order.

```suggestion
    const meteorRuntimeConfig = hook({ arch, request, encodedCurrentConfig: boilerplate.baseData.meteorRuntimeConfig, updated: runtimeConfig.update[arch] });
```

packages/webapp/webapp_server.js
```diff
@@ -345,8 +345,33 @@ function getBoilerplate(request, arch) {
...
+    runtimeConfig.update[arch] = false;
+    if(!meteorRuntimeConfig) return;
+    // boilerplate.baseData.meteorRuntimeConfig
```
filipenevola, line 372: I don't think this comment is necessary

packages/webapp/webapp_server.js
```diff
@@ -345,8 +345,33 @@ function getBoilerplate(request, arch) {
...
+WebApp.encodeRuntimeConfig = function (rtimeConfig) {
+  return JSON.stringify(encodeURIComponent(JSON.stringify(rtimeConfig)));
+}
+
+WebApp.decodeRuntimeConfig = function (rtimeConfig) {
+  return JSON.parse(decodeURIComponent(JSON.parse(rtimeConfig)));
+}
+
+const runtimeConfig = {
+  'hooks': [],
```
filipenevola, line 357: We don't need single quotes here. And also not below.

packages/webapp/webapp_server.js
```diff
@@ -345,8 +345,33 @@ function getBoilerplate(request, arch) {
...
 
+WebApp.encodeRuntimeConfig = function (rtimeConfig) {
+  return JSON.stringify(encodeURIComponent(JSON.stringify(rtimeConfig)));
+}
+
+WebApp.decodeRuntimeConfig = function (rtimeConfig) {
+  return JSON.parse(decodeURIComponent(JSON.parse(rtimeConfig)));
+}
+
+const runtimeConfig = {
```
filipenevola, line 356: A comment would be nice, explaining what hooks will contain and also what update will contain.

I would rename update to `isUpdatedByArch` or something similar

filipenevola (changes requested):
I understand where this idea is coming from.

I already solved this problem in a different way in at least 2 apps that I worked in the past.

Usually my solution is to inject something from my client calls so my server knows how to implement a different response based in the client and so all my Cordova apps pointing to the same backend would be always using the same `ROOT_URL` avoiding the problems listed in the issue and PR. The way to get a different context is using Cordova App Preferences and reading this preferences in the initialization with a Cordova plugin and then providing this context in all the calls. 

I understand that this solution here is more robust in the sense that we don't need to inject this information in all the calls as we could have multiple `ROOT_URL` and because of that the server could decide the responses without any data provided on each call as the calling URL would be different each time. So I agree with the solution and the idea behind it but I would recommend a full check in our codebase for all the usages of `ROOT_URL` to be sure that we are not breaking other use cases with this solution.

I also suggest a few small changes in the code. So it's following our latest standards.

filipenevola:
@brucejo75 to be clear, my long feedback here was just to provide more background on how I'm solving this problem in a different way.

Your changes are going in a good direction, you can proceed with this PR, please provide the changes that I asked in the code review.

packages/webapp/webapp_server.js
brucejo75: Thanks! Done.

packages/webapp/webapp_server.js
brucejo75: Good feedback, thanks! done.

packages/webapp/webapp_server.js
brucejo75: Oops!  thanks

packages/webapp/webapp_server.js
brucejo75: got it.  done.

packages/webapp/webapp_server.js
brucejo75: Good point!

brucejo75:
Updated my original comment to reflect passing an object to the callback instead of a list of arguments.

brucejo75:
@filipenevola, @StorytellerCZ
My build failed because I did not give CircleCi enough permissions on github.  I gave those permissions but I cannot figure out how to rerun the workflow without write permissions.  Any help?  Thanks!

filipenevola (approved):
Thanks for the updates. I just rerun it on CircleCI now.

brucejo75:
Thanks @filipenevola!

Now it is kicking out an error, related to node v16?
Not sure I can address this?

<details>
<summary>Error output</summary>

```
+ @types/node@16.3.0
updated 1 package and audited 3547 packages in 8.094s
found 1 moderate severity vulnerability
  run `npm audit fix` to fix them, or `npm audit` for details
utils/func-utils.ts:33:18 - error TS2554: Expected 3 arguments, but got 2.

33       resolve => setTimeout(resolve, actualDelayMs)
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@types/node/timers.d.ts:42:99
    42         function setTimeout<TArgs extends any[]>(callback: (...args: TArgs) => void, ms?: number, ...args: TArgs): NodeJS.Timeout;
                                                                                                         ~~~~~~~~~~~~~~
    Arguments for the rest parameter 'args' were not provided.


Found 1 error.
```
</details>

StorytellerCZ:
@brucejo75 Merged from devel where this error is fixed, should run now.

brucejo75:
Thanks @StorytellerCZ.  Now there is another error?  Not sure how I could address?
Error from continuous-integration/appveyor/pr:

```
Build execution time has reached the maximum allowed time for your plan (35 minutes).
```

Thanks!

StorytellerCZ:
AppVeyor is fickle (after all it tests Windows) . If it just timed out then trying to re-run is the best response. I'll take a look.

StorytellerCZ:
All green! I will merge this together with few other PRs into 2.4 and release the first beta afterwards.

StorytellerCZ:
@brucejo75 Merging now, can you please create a PR to the docs to document this?

brucejo75:
@StorytellerCZ,

Working on the docs!

Also, I managed to find a bug in my implementation by testing my documentation code.  It is a 1 liner for the case when multiple hooks are added. Ugh.

Do I just modify this branch?  Or do I need to introduce a new branch?  Please advise.

filipenevola:
@brucejo75 create another branch based on `devel`. 😉

brucejo75:
OK,  should wait until devel has 2.4 merged back?  2.4 has the PR merged and devel does not yet.

StorytellerCZ:
@brucejo75 just create a new branch based on `release-2.4`, since that branch is going be merged only when 2.4 is going to be released.

PR #11517, opened 2021-07-05
Add Documentation for the forceApprovalPrompt Option of Accounts.loginWith<ExternalService>

trudeaua21:
In the [documentation for `Accounts.loginWith<ExternalService>`](https://docs.meteor.com/api/accounts.html#Meteor-loginWith%3CExternalService%3E), the `forceApprovalPrompt` option which forces the user consent prompt to appear is missing.

The option's implementation can be found [here.](https://github.com/meteor/meteor/blob/ffcfa5062cf1bf8a64ea64fef681ffcd99fe7939/packages/google-oauth/google_client.js#L63)

To add the documentation, I simply copied the description of the option from the same field on `Accounts.ui.config` [(seen here)](https://github.com/meteor/meteor/blob/master/packages/accounts-ui-unstyled/accounts_ui.js#L37)

<strike>**NOTE:** I bumped the patch version as the contribution guidelines instructed, though I'm not really sure if I was supposed to (since I didn't really change much code, just added a line to documentation). Sorry if I made the wrong call!</strike> - I un-bumped the patch version because it looked like that was what was causing `Build and Test` to fail, but it ended up still failing anyway. I would be really appreciative if whoever reviews this could let me know whether to bump the patch version or not!


trudeaua21:
I'm keeping this as a draft for now because I'm not sure why the `Build and Test` check is failing. This is my first contribution so I'd welcome anyone helping me out with figuring out why that is!

StorytellerCZ:
@trudeaua21 Don't worry about the failure, the error is related to TypeScript. We'll fix in devel and then merge the fix here. Also version bump is not needed for docs addition.

StorytellerCZ (approved):

PR #11530, opened 2021-07-12
Add note about fail

harryadel:
Was reading into [`check`](https://github.com/meteor/meteor/blob/devel/packages/check/match_test.js#L347) package and wanted to know more about [`fail`](https://github.com/meteor/meteor/blob/devel/packages/tinytest/tinytest.js#L35) method but there wasn't any notes about it so I added one.


StorytellerCZ:
@harryadel nice one! Didn't even know this was a thing. :laughing:

StorytellerCZ (approved):

PR #11545, opened 2021-07-16
Enable HMR for all web arch's

### LINKED ISSUE #11353: HMR code causes errors instead of gracefully disabling itself on IE11 regardless of whether hmr is added as a package or not
Trying to test an application with the HMR package added in IE11 makes it crash on browser first load because the package contains code that is incompatible with IE:

image

I could not find a workaround, unfortunately, since removing the hot-module-reload package from the app was not enough - react-fast-refresh is always loaded as a dependency of the ecmascript package.

zodern:
Enable HMR for legacy client and cordova. I haven't tested in cordova yet.

Fixes #11353


filipenevola:
Hi, these changes are affecting the app load on iOS and Android, at least in the simulator/emulator the app is not loading.

After the splash screen I just get a blank screen. Running with Meteor 2.3.2 it works fine.

packages/hot-module-replacement/client.js
```diff
@@ -61,7 +60,7 @@ function handleMessage(message) {
       console.log('HMR: Will enable HMR the next time the page is loaded');
       mustReload = true;
     } else {
-      console.log(`HMR: Register failed for unknown reason`, message);
+      console.log('HMR: Register failed for unknown reason', message);
```
StorytellerCZ, line 63: I think we should start using Logging package for log output.

StorytellerCZ (changes requested):

filipenevola:
Hi @zodern, what is the status here?

Do you need help in the Cordova work?

zodern:
It should now work on cordova, though I want to simplify some of it.

packages/hot-module-replacement/client.js
zodern: Since the logging package depends on ecmascript, it would create a circular dependency.

PR #11548, opened 2021-07-19
less v4

StorytellerCZ:
Upgrade the less package to v4


StorytellerCZ:
`less` package test are failing and they are apparently failing for the previous version as well. Investigating, but not really sure why.

StorytellerCZ:
Fixed the tests, the issue was that the computed styles left the main `border-style` empty and filled the individual sides with the data.

PR #11556, opened 2021-07-22
createIndex for collections

StorytellerCZ:
Add `_createIndex` as a collection function. This is a new name for `_ensureIndex` which MongoDB has deprecated and removed in MongoDB 5.0. This gives people the opportunity to start using the new MongoDB API, while we maintain backward compatibility.

### DIFF
--- a/History.md
+++ b/History.md
@@ -5,6 +5,9 @@
 * `meteor-tool@2.4`
   - `meteor show` now reports if a package is deprecated
 
+* `mongo@1.13.0`
+  - Add `createIndex` as a collection function. This is a new name for `_ensureIndex` which MongoDB has deprecated and removed in MongoDB 5.0. Use of `_ensureIndex` will show a deprecation warning on development.
+
 #### Independent Releases
 
 * `minifier-js@2.6.1`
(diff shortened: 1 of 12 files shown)


packages/mongo/collection.js
```diff
@@ -674,6 +674,13 @@ Object.assign(Mongo.Collection.prototype, {
     self._collection._ensureIndex(index, options);
```
filipenevola, line 674: This code should check if createIndex is available and if so it should use createIndex as well.

packages/mongo/collection.js
```diff
@@ -674,6 +674,13 @@ Object.assign(Mongo.Collection.prototype, {
...
   },
 
+  _createIndex(index, options) {
```
filipenevola, line 677: As this is a public API we shouldn't have `_` before the name, I would go with `createIndex`

filipenevola (changes requested):
As this is available since version 3 we should support it and make our old `_ensurendex` to use it internally as well.

History.md
```diff
@@ -5,6 +5,9 @@
 * `meteor-tool@2.4`
   - `meteor show` now reports if a package is deprecated
 
+* `mongo@1.13.0`
+  - Add `createIndex` as a collection function. This is a new name for `_ensureIndex` which MongoDB has deprecated and removed in MongoDB 5.0. Use of `_ensureIndex` will show a deprecation warning on development.
```
filipenevola, line 9: I believe it's important to mention that createIndex was added in 3. Otherwise people can think it's only going to work with Mongo 5.

filipenevola (approved):

History.md
StorytellerCZ: Will add. Right now I'm looking through our docs and I don't see `_ensureIndex` documented there, so I will add `createIndex` there as well.

Issue #11513, opened 2021-07-03
APM Error: discarding invalid livedata message


rj-david:
- The version of Meteor showing the problem.
Meteor 2.3
- The last version of Meteor where the problem did _not_ occur, if applicable.
Meteor 2.2
- The operating system you're running Meteor on.
Ubuntu

We started receiving this error in APM since we upgraded to Meteor 2.3: `discarding invalid livedata message`

StorytellerCZ:
Could be potentially related to handshake between client and server which was adjusted going to 2.3.

brianlukoff:
We are seeing the same thing after upgrading to 2.3.

rj-david:
Here is the change that added the debug message. It's possible that the cause to be happening even before 2.3
https://github.com/meteor/meteor/commit/d53ab4cd177413055402f7cf3db9936c209d935f

StorytellerCZ:
We could potentially not report the discarded message, but I don't think that is a good approach. I will update the affected code to use the logging tool so that the debug message won't appear in production.

filipenevola:
Hi @rj-david do you see an object after `discarding invalid livedata message`?

As you can see on my change we shouldn't receive a message object without a msg field on it so that is why we have this log. 

Maybe it has another fields inside this object and it could give us some hint where it is coming from.

And I did this change as it was breaking a test after some removal of deprecated code.

filipenevola:
> We could potentially not report the discarded message, but I don't think that is a good approach. I will update the affected code to use the logging tool so that the debug message won't appear in production.

Changing the way we log messages is not going to solve this issue. We need to understand which message is being discarded.

rj-david:
> Hi @rj-david do you see an object after `discarding invalid livedata message`?

Monti APM only showed `[object Object]` unfortunately.

This happened only 4 times during the past week.

tmihelcich:
I just started receiving this message during development on a project that is using Meteor 1.4.3.2. Here is what I see:

`discarding invalid livedata message { testMessageOnConnect: true }`

It happens just after I save changes to file and the app reloads.

PolarisWT:
I can confirm same message using Meteor 2.2 on local environment.   Local server running for past day or so without any updates in between.  Just started showing this log on client updates

```console
=> Client modified -- refreshingdiscarding invalid livedata message { testMessageOnConnect: true }
=> Client modified -- refreshing (x2)discarding invalid livedata message { testMessageOnConnect: true }
[...]

PR #11558, opened 2021-07-26
Use logging for ddp-client

StorytellerCZ:
Replace `Meteor._debug` in `ddp-client` to add more nuance to the messages from there and address #11513


StorytellerCZ:
I think the only question here is which level should be used for what message.

PR #11559, opened 2021-07-26
Initial implementation of DDP resumption

znewsham:
Recently we've noticed that a lot of CPU (and in some cases memory) issues can be tracked down to the problem of transient disconnects, and the effort it takes to re-setup a DDP connection clean. This is particularly egregious if the cause of the disconnects is at the load balancer level (e.g., a reconfiguration causing NGINX to drop existing connections all at the same time).

In the majority of cases these disconnects last milliseconds, but can take upwards of 30 seconds to re-subscribe to all data that the client previously had access to (aside: particularly in the case of using a sub manager to cache subscriptions). 

It would of course be ideal if these disconnects never happened - but being that we live in an imperfect world... It would be much better if in the case of a transient disconnect, the cost were much lower.

To that end, this PR is an initial implementation of DDP session resumption (resuming?) based on the following:

1. All inflight methods will still be requested.
2. All active subscriptions will still be re-ran
3. Resumption will only occur if the server being connected to has a session with the same `id` AND the number of messages sent by the server matches the number of messages received by the client (excluding ping/pong - e.g., `result`,  `updated`, `nosub`, `ready`, `added`, `removed`, `changed`) - we don't need to consider messages sent from client -> server due to (1) and (2).
4. In the case of a persistent disconnect (e.g., triggered by code on client or server, or navigating away) the session should terminate immediately
5. In the case of a (potentially) transient disconnect - the memory and CPU overhead to maintain the session should be low.

This implementation modifies the following to accomplish this:
1. track a `sentCount` on the server and `receivedCount` on the client.
2. transmit a `{ msg: "disconnect" }` before disconnecting on an explicit disconnect.
3. Upon a non-graceful disconnect (without seeing a `disconnect` message) maintain a list of messages that should be sent upon reconnect 
4. Defer the removal of a session for `DISCONNECT_GRACE_PERIOD` ms or until `MAX_QUEUE_LENGTH` is reached
5. On reconnect, perform a lookup on `msg.session` and, if found, check the `receivedCount === sentCount`.
6. On unsuccessful "resume" (or a resume never comes) - cleanup the session as normal.

Potential limitations/vulnerabilities:
1. session hijacking - currently resuming a session with a valid sessionId will grant you the same access from a new machine without logging in. I'm looking into a solution for this - though if an attacker can get a valid session ID, they can probably get the valid resume token too.

NOTE: This implementation was ported over from a custom copy of DDP server/client packages in an application. It's been tested there but it's possible that in applying the changes here I messed something up. Currently I'm interested in if there are any pitfalls in this approach, if none are obvious to the community I'll work on implementing tests.

### DIFF
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -102,6 +102,10 @@ export class Connection {
     }
 
     self._lastSessionId = null;
+    // how many messages we've received (excluding ping/pong).
+    // when we try to reconnect to the server, it will check this against the number of messages it sent.
+    // if there is a mismatch, our info is out of date and we need a clean session.
+    self._receivedCount = 0;
     self._versionSuggestion = null; // The last proposed DDP version.
     self._version = null; // The DDP version agreed on by client and server.
     self._stores = Object.create(null); // name -> object with methods
@@ -897,11 +901,12 @@ export class Connection {
    * @locus Client
    */
   disconnect(...args) {
+    this._send({ msg: 'disconnect' });
     return this._stream.disconnect(...args);
   }
 
   close() {
-    return this._stream.disconnect({ _permanent: true });
+    return this.disconnect({ _permanent: true });
   }
 
   ///
@@ -972,6 +977,8 @@ export class Connection {
       // this possible, at which point we'll probably need more code here.
       return;
     }
+
+    self._receivedCount = 1;
 
     // Server doesn't have our data any more. Re-sync a new session.
 
@@ -1629,7 +1636,9 @@ export class Connection {
         Meteor._debug('discarding invalid livedata message', msg);
       return;
     }
-
+    if (!['ping', 'pong'].includes(msg.msg)) {
+      this._receivedCount++;
+    }
     if (msg.msg === 'connected') {
       this._version = this._versionSuggestion;
       this._livedata_connected(msg);
@@ -1669,7 +1678,10 @@ export class Connection {
     // NOTE: reset is called even on the first connection, so this is
     // the only place we send this message.
     const msg = { msg: 'connect' };
-    if (this._lastSessionId) msg.session = this._lastSessionId;
+    if (this._lastSessionId) {
+      msg.session = this._lastSessionId;
+      msg.receivedCount = this._receivedCount;
+    }
     msg.version = this._versionSuggestion || this._supportedDDPVersions[0];
     this._versionSuggestion = msg.version;
     msg.support = this._supportedDDPVersions;
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -21,6 +21,13 @@ var SessionDocumentView = function () {
 DDPServer._SessionDocumentView = SessionDocumentView;
 
 
+// how long we should maintain a session for after a non-graceful disconnect before killing it
+// sessions that reconnect within this time will be resumed with minimal performance impact.
+const DISCONNECT_GRACE_PERIOD = 5000;
+
+// how many messages should we queue during a non-graceful disconnect before we kill the session (to insure against memory leaks)
+const MAX_QUEUE_LENGTH = 100;
+
 _.extend(SessionDocumentView.prototype, {
 
   getFields: function () {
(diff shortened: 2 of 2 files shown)


packages/ddp-server/livedata_server.js
```diff
@@ -423,52 +434,61 @@ Object.assign(Session.prototype, {
...
+
+      Meteor.defer(function () {
+        // stop callbacks can yield, so we defer this on close.
+        // sub._isDeactivated() detects that we set inQueue to null and
+        // treats it as semi-deactivated (it will ignore incoming callbacks, etc).
+        self._deactivateAllSubscriptions();
+
+        // Defer calling the close callbacks, so that the caller closing
+        // the session isn't waiting for all the callbacks to complete.
+        _.each(self._closeCallbacks, function (callback) {
```
StorytellerCZ, line 465: If we could do without underscore that would be great!

StorytellerCZ (changes requested):

packages/ddp-server/livedata_server.js
znewsham: I think this already exists, I just indented it?

znewsham: Although either way I can remove easily enough!

StorytellerCZ: Please do! It is an incremental approach, but it it the best we have right now.

StorytellerCZ:
I have scheduled it for 2.5 so that we can get feedback on this and can add tests.

packages/ddp-server/livedata_server.js
```diff
@@ -1606,9 +1664,18 @@ Object.assign(Server.prototype, {
...
+    const f = () => {
+      self.sessions.delete(session.id);
+      callback();
+    };
+    if (session._expectingDisconnect) {
+      return f();
+    }
+    session.messageQueue = [];
+    session._pendingRemoveFunction = f;
+    session._removeTimeoutHandle = Meteor.setTimeout(f, DISCONNECT_GRACE_PERIOD);
```
filipenevola, line 1678: disconnect grace period should be also an option in the package to be replaced if someone wants to customize.

packages/ddp-server/livedata_server.js
```diff
@@ -423,52 +434,61 @@ Object.assign(Session.prototype, {
...
 
-    // Unregister the session.
-    self.server._removeSession(self);
   },
 
   // Send a message (doing nothing if no socket is connected right now.)
   // It should be a JSON object (it will be stringified.)
   send: function (msg) {
     var self = this;
+    const isPingPong = ['ping', 'pong'].includes(msg.msg);
```
filipenevola, line 477: should use the same variable that I suggested above, otherwise it's going to be hard to maintain

packages/ddp-server/livedata_server.js
```diff
@@ -423,52 +434,61 @@ Object.assign(Session.prototype, {
...
+    if (self.messageQueue && !isPingPong) {
+      self.messageQueue.push(msg);
+      if (self.messageQueue.length > MAX_QUEUE_LENGTH) {
```
filipenevola, line 480: max queue length should be also an option in the package to be replaced if someone wants to customize.

filipenevola (changes requested):
Looks good in general.

Thank you.

I've asked small adjustments.

znewsham:
> Looks good in general.
> 
> Thank you.
> 
> I've asked small adjustments.

Happy to expose those two variables through the package (or maybe even pull from environment vars) - regarding the ping/pong issue - I can extract to a static named variable, but they are in different packages, so I'm not sure they can share the variable (will still be two declarations)

packages/ddp-client/common/livedata_connection.js
```diff
@@ -1628,7 +1635,9 @@ export class Connection {
       }
       return;
     }
-
+    if (!['ping', 'pong'].includes(msg.msg)) {
```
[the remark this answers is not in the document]
vlasky: @filipenevola perhaps this variable should be declared in ddpcommon/utils.js as it is referenced by both `ddp-client/common/livedata_connection.js`  and `ddp-server/livedata_server.js`?

vlasky:
I am continuing efforts on this PR here:

https://github.com/meteor/meteor/pull/11845

StorytellerCZ:
Closing this in favor of #11845

PR #11571, opened 2021-08-04
Fix #9907 - Deleting METEOR_SETTINGS only in development

### LINKED ISSUE #9907: cannot pass METEOR_SETTINGS to meteor run

StorytellerCZ:
Fix for #9907 cannot pass METEOR_SETTINGS to meteor run
Disregard `METEOR_SETTINGS` only in development (when you should use `--settings`).

### DIFF
--- a/tools/runners/run-app.js
+++ b/tools/runners/run-app.js
@@ -1,7 +1,5 @@
 var _ = require('underscore');
 var Fiber = require('fibers');
-const uuid = require("uuid");
-var fiberHelpers = require('../utils/fiber-helpers.js');
 var files = require('../fs/files');
 var watch = require('../fs/watch');
 var bundler = require('../isobuild/bundler.js');
@@ -13,7 +11,6 @@ var catalog = require('../packaging/catalog/catalog.js');
 var Profile = require('../tool-env/profile').Profile;
 var release = require('../packaging/release.js');
 import { pluginVersionsFromStarManifest } from '../cordova/index.js';
-import { CordovaBuilder } from '../cordova/builder.js';
 import { closeAllWatchers } from "../fs/safe-watcher";
 import { eachline } from "../utils/eachline";
 import { loadIsopackage } from '../tool-env/isopackets.js';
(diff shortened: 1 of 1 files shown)


tools/runners/run-app.js
```diff
@@ -1,7 +1,5 @@
 var _ = require('underscore');
 var Fiber = require('fibers');
-const uuid = require("uuid");
-var fiberHelpers = require('../utils/fiber-helpers.js');
```
StorytellerCZ, line 4: Removing unused imports here

tools/runners/run-app.js
```diff
@@ -172,18 +169,16 @@ Object.assign(AppProcess.prototype, {
     }
     if (self.settings) {
       env.METEOR_SETTINGS = self.settings;
-    } else {
+    } else if (env.METEOR_SETTINGS && env.NODE_ENV === 'development') {
```
StorytellerCZ, line 172: The new if statement ensures that `METEOR_SETTINGS` is deleted only in development. In other scenarios it will be applied.

filipenevola (approved):

tools/runners/run-app.js
zodern: With the react refresh integration, we've found that NODE_ENV isn't always set to the correct value matching the build mode. If the build mode is available here, it would be more reliable.

macrozone:
as mentioned in https://github.com/meteor/meteor/issues/9907

this PR does not solve the issue. It should never delete env.METEOR_SETTINGS

PR #11576, opened 2021-08-09
Allow usage of http packages v1 and v2 in affected packages

### LINKED ISSUE #11575: Impossible to update to Meteor 2.3 because HTTP package's major version was incremented
I maintain https://sandstorm.io, an open source project that is not very actively developed these days, but is still used by a lot of people. I try to do a new release every month to update dependencies in order to avoid security problems. The UI is built on Meteor.

In Meteor 2.3, the HTTP package has had its major version bumped from 1.x to 2.0.0, and several core packages (like `google-oauth`) now depend on this new verison.

Meteor's package system assumes different major versions are incompatible, and also apparently does not permit running two different versions of the same package at the same time.

Sandstorm also depends on the `tap:i18n` package. This used to be the most popular i18n framework for Meteor, but unfortunately it has not been updated in many years. This package transitively depends on HTTP 1.x. Therefore, Meteor refuses to update, because it thinks that it cannot satisfy `tap:i18n`'s dependencies and `google-oauth`'s dependencies, because they require different versions of HTTP.

I cannot feasibly migrate our codebase to a different i18n package. I maintain this codebase on weekends as a volunteer, and I just don't have the time.

This seems to mean it's impossible for us to update to Meteor 2.3. That is unfortunate as it presumably means we cannot get security updates anymore.

**Proposed fix**

I believe that if Meteor changed the `google-oauth`, `github-oauth`, and any other core packages to depend on `http@1.0.0 || 2.0.0` -- i.e. allow either version -- then we would be able to stick to HTTP 1.x and the problem is solved.

I understand that Meteor would like to move away from the 1.x implementation, but the curse of maintaining a platform is that sometimes you are stuck supporting old APIs forever. If the `fetch()`-based implementation can be made sufficiently backwards-compatible, it could perhaps be released as 1.5.0?

StorytellerCZ:
Fix #11575 by allowing usage of both major versions of the deprecated `http` package until we can fully replace it with `fetch`.

### DIFF
--- a/History.md
+++ b/History.md
@@ -29,7 +29,28 @@
 
 * `accounts-base@2.0.1`
   - Create index on `services.password.enroll.when`
-  - Blaze weak dependency updated to v2.5.0 
+  - Blaze weak dependency updated to v2.5.0
+
+* `facebook-oauth@1.9.1`
+  - Allow usage of `http` package both v1 and v2 for backward compatibility
+
+* `github-oauth@1.3.1`
+  - Allow usage of `http` package both v1 and v2 for backward compatibility
+
+* `google-oauth@1.3.1`
+  - Allow usage of `http` package both v1 and v2 for backward compatibility
+
+* `meetup-oauth@1.1.1`
+  - Allow usage of `http` package both v1 and v2 for backward compatibility
+
+* `meteor-developer-oauth@1.3.1`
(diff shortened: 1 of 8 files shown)


kentonv:
I think this will fix my issue. Thanks!

kentonv (approved):

filipenevola (approved):

PR #11579, opened 2021-08-11
Diff sequence 1.1.2 [Perf Improvements]

mvogttech:
Re-opening.

Local tests passed.


packages/ddp-server/livedata_server.js
```diff
@@ -1,5 +1,7 @@
 DDPServer = {};
 
+import { DiffSequence} from "../diff-sequence/diff"
```
StorytellerCZ, line 3: Since this is a different package, you will need to import it like you would in an app. Same in `packages/ddp-client/common/livedata_connection.js`. Also double check that the dependency is in `package.js`

StorytellerCZ (changes requested):
Looking good! Please don't forget to comment on any changes in `packages/diff-sequence/diff.js` for easier review. Thanks!

packages/ddp-server/livedata_server.js
mvogttech: Will do. Wrapping up these changes and final comments on the code. I will change it from a draft once ready for review.

Thanks!

mvogttech:
**Changes**

- Complete refactor of the functional component
- Replacement of heavy forEach array loops in favor of for...let loops where performance improvements are measured
- Removal of array.replace() in DiffSequence.diffQueryOrderedChanges() method in favor of unshift ([performance test](https://www.measurethat.net/Benchmarks/ShowResult/247776))
- Use of Map() in DiffSequence.makeChangedFields() method

mvogttech (commented):
Updated Package Imports

filipenevola:
@mvogttech, sorry for the huge delay.

As this PR is about performance, do you have the measures to show what are the improvements (with numbers comparing before and after) and in which code flows?

packages/diff-sequence/diff.js
```diff
@@ -1,287 +1,268 @@
...
+        },
+        rightOnly: function (key, value) {
+          fields.set(key, value);
+        },
+        both: function (key, leftValue, rightValue) {
+          if (!EJSON.equals(leftValue, rightValue))
+            fields.set(key, rightValue);
+        }
+      });
+      return Object.fromEntries(fields);
```
radekmie, line 254: Is this change performance-related? I can’t imagine how creating a `Map` instance and then transforming it into an object may be faster than creating the object straight away.

filipenevola:
As we don't have any evidence that these changes improve the performance and the PR aims to improve performance, we will close this PR. If we have pieces of evidence later, of course, we can reconsider this PR.

PR #11581, opened 2021-08-11
Release 2.2.3

StorytellerCZ:
Security release of Node with backports for Typescript and Babel updates.

Test it by running:
```
meteor update --release 2.2.3-rc.2
```


filipenevola (approved):
It's approved just double check if we are not introducing new changes that are not yet on 2.3.5. 

I don't think we are but a double check wouldn't hurt.

StorytellerCZ:
@filipenevola I have only backported Typescript and Babel upgrades all the way from 2.4 development. As per the linked issues.
Should I bump the affected packages for 2.3.5 or should publishing them here be enough. Now that I think about it, it might be better to bump them for 2.3.5 to be of a different version then they are here just to be sure due to the different Node versions and everything around it (or am I worrying too much?).

filipenevola:
> @filipenevola I have only backported Typescript and Babel upgrades all the way from 2.4 development. As per the linked issues.
> Should I bump the affected packages for 2.3.5 or should publishing them here be enough. Now that I think about it, it might be better to bump them for 2.3.5 to be of a different version then they are here just to be sure due to the different Node versions and everything around it (or am I worrying too much?).

If they were not changed in this release it's not necessary.

StorytellerCZ:
Alright then, let me get started on releasing this and another RC for 2.3.5...

PR #11582, opened 2021-08-11
email 2.2 improvements & fixes

StorytellerCZ:


packages/email/email.js
```diff
@@ -218,8 +218,11 @@ Email.send = function (options) {
   if (!send) return;
 
   const customTransport = Email.customTransport;
-  const transport = customTransport ? false : getTransport();
-  if (transport) {
+  if (customTransport) {
+    const packageSettings = Meteor.settings.packages?.email || {};
+    customTransport({ packageSettings, ...options });
+  } else if (Meteor.isProduction) {
```
filipenevola, line 224: Please avoid else's using return statements.

For example:
```js
if (x) {
  // doSomething();
  return;
}

if (y) {
  // doSomethingElse();
  return;
}

// doLastCase();
```

filipenevola (changes requested):
I've just asked one adjustment.

packages/email/email.js
StorytellerCZ: Fixed

filipenevola (approved):

PR #11585, opened 2021-08-13
callback-hook add forEach iterator to follow the JS standard

StorytellerCZ:
Add `forEach` to the `callback-hook` so that it follows the naming standard for ES when it comes to looping. Retains `each` iterator for backward compatibility.

### DIFF
--- a/packages/callback-hook/hook.js
+++ b/packages/callback-hook/hook.js
@@ -94,7 +94,7 @@ export class Hook {
   // The iteration is stopped if the iterator function returns a falsy
   // value or throws an exception.
   //
-  each(iterator) {
+  forEach(iterator) {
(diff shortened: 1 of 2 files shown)


packages/callback-hook/hook.js
```diff
@@ -112,6 +112,10 @@ export class Hook {
       }
     }
   }
+
+  each(iterator) {
```
filipenevola, line 116: if you want to deprecate this it would be nice a jsdoc here saying that.

filipenevola (changes requested):

packages/callback-hook/hook.js
StorytellerCZ: Added, though it is a super soft deprecation and  don't plan on removing this any time soon.

filipenevola (approved):

PR #11594, opened 2021-08-20
Fix sending test metadata

StorytellerCZ:
TEST_METADATA is by default `"{}"`, so the check if to send it is always true, even though it was intended to be false when the env var is not set.
Thanks to @zodern for pairing up on this one.


filipenevola (approved):

zodern (approved):

StorytellerCZ:
#11595 will be the patch release that will go out ASAP, I'm leaving this still open to remove the lost console log in 2.4 branch.

filipenevola:
@StorytellerCZ you should merge this to devel and then devel into 2.4 branch.

StorytellerCZ:
Yes, I have made a mistake in starting this from the 2.4 branch, so I have #11595 that I will merge first into devel, then into 2.4 and then this will be merged.

PR #11602, opened 2021-08-23
Accounts-passwordless package

renanccastro:
## Description
A new method of signing in. Requires only e-mail from the user and when signing in it would send a one-time link to the user's primary e-mail through which they would sign in.

## Why?

Easiness of login/signup is important for all kind of business, but in e-commerce, for example, buying fast is what matter the most (in most cases).

## How?

You can log in passing a 6 length code, that is generated randomly and has an expiration of X hours(1 by default), to the method called:

```js
Meteor.loginWithToken(token, (err) => {
  if (err) {
    // handle error
  } else {
    // successful login!
  }
});

```
You can request a token for a user, which can be an existing, or non-existing user by using:

```js
Accounts.requestLoginTokenForUser({selector, user});
```

The selector is a mongo selector. A default user is an object that will be used to it if it not exists, so it's an object. Ex: { phone: 55555 }, { username: 33333}, {email: test@test.com}. If no user is found, a new one is created using the "user" object.

By default, account-passwordless will send an email if an email is set for the user, pre-configured with the following:

```text
Hello!

Type the following token in our login webpage to be logged in:
${token}
If you want, you can click the following link to be automatically logged in:
${url}

```
which in turn can be customized with the following:

```js
Accounts.emailTemplates.sendLoginToken = {
  subject(user) {
    return "Your login token on YourWebsite";
  },
  text(user, token, url) {
    return `Hello!
Type the following token in our login webpage to be logged in:
${token}
If you want, you can click the following link to be automatically logged in:
${url}
Thanks,
YourWebsite team
`
  },
  html(user, url) {
    // This is where HTML email content would go.
    // See the section about html emails below.
  }
};

```

You can opt-out from email sending and also configuring the package with the following:

```js
Accounts.config({
  sendLoginTokenByEmail: false,
  loginTokenExpirationHours: 1 (default)
});

```

If you want to send the user to a different page than home("/"), you can use:

```js
Accounts.urls.loginWithTokenUrl = (token) => {
  return Meteor.absoluteUrl(`${token}`);
};
```

Now, for the callbacks:

### Callbacks

```js
Accounts.onLoginWithTokenLink(({token, done}) => {
  // called in the client side when a token is detected in the URL
  // no need to call Meteor.loginWithToken(token) at the end
})
```

```js
Accounts.onCreateLoginToken(({token, user}) => {
  const { phone } = user.phone;
 
 customSendSMSLogic(phone, token);
});
```


packages/accounts-passwordless/email_templates.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @summary Options to customize emails sent from the Accounts system.
+ * @locus Server
+ * @importFromPackage accounts-base
+ */
+Accounts.emailTemplates = {
+  ...(Accounts.emailTemplates || {}),
+  from: "Accounts Example <no-reply@example.com>",
+  siteName: Meteor.absoluteUrl().replace(/^https?:\/\//, '').replace(/\/$/, ''),
```
StorytellerCZ, line 9: Isn't `from` and `siteName` alterady set in `Accounts.emailTemplates`?

packages/accounts-passwordless/email_templates.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @summary Options to customize emails sent from the Accounts system.
+ * @locus Server
+ * @importFromPackage accounts-base
+ */
+Accounts.emailTemplates = {
+  ...(Accounts.emailTemplates || {}),
```
StorytellerCZ, line 7: Shouldn't this be at the end so that it overrides the defaults?

StorytellerCZ (changes requested):
Doesn't look complete at this time. Also changes to the Accounts UI package might be needed to accommodate this, or we might take this opportunity to move those packages to Blaze.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -0,0 +1,136 @@
...
+  getUserById,
+  handleError,
+  tokenValidator,
+  userQueryValidator,
+} from './server_utils';
+import { Random } from 'meteor/random';
+
+Accounts.constructor.prototype._onCreateLoginTokenHook = () => true;
+
+Accounts.constructor.prototype.Accounts._checkToken = ({ user, token }) => {
```
filipenevola, line 12: I believe there is an extra .Accounts here.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -0,0 +1,136 @@
...
+// Handler to login with an ott.
+Accounts.registerLoginHandler('passwordless', options => {
+  if (!options.token) return undefined; // don't handle
+
+  check(options, {
+    user: userQueryValidator,
+    token: tokenValidator(),
+  });
+
+  const user = Accounts._findUserByQuery(options.user, {
```
filipenevola, line 45: Should we use something from `accounts-password`?

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    if (!user) {
+      handleError('User could not be created');
+    }
+
+    const sequence = Random.hexString(
+      Accounts._options.tokenSequenceLength || 6
+    );
+    Meteor.users.update({
```
filipenevola, line 91: It's missing the selector

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -0,0 +1,136 @@
...
+ * @summary Send an email with a link the user can use to reset their password.
+ * @locus Server
+ * @param {String} userId The id of the user to send email to.
+ * @param {String} [email] Optional. Which address of the user's to send the email to. This address must be in the user's `emails` list. Defaults to the first email in the list.
+ * @returns {Object} Object with {email, user, token, url, options} values.
+ * @importFromPackage accounts-base
+ */
+Accounts.sendLoginTokenEmail = ({ userId, sequence, email }) => {
+  const user = getUserById(userId);
+  const url = Accounts.urls.loginToken(token, extraParams);
```
filipenevola, line 124: where is `loginToken` defined?

filipenevola (changes requested):

packages/accounts-passwordless/passwordless_server.js
filipenevola: fixed.

packages/accounts-passwordless/passwordless_server.js
filipenevola: fixed.

packages/accounts-passwordless/passwordless_server.js
filipenevola: fixed

packages/accounts-passwordless/package.js
```diff
@@ -0,0 +1,22 @@
+Package.describe({
+  summary: 'No-password login/sign-up support for accounts',
+  version: '0.0.1',
```
StorytellerCZ, line 3: Should be v1.0.0 or will be releasing it as an experimental package for now?

packages/accounts-passwordless/passwordless_client.js
```diff
@@ -0,0 +1,123 @@
...
+        reportError(error, callback);
+      } else {
+        callback && callback();
+      }
+    },
+  });
+};
+
+/**
+ * @summary Request a forgot password email.
```
StorytellerCZ, line 62: Wrong summary.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -0,0 +1,246 @@
+import { Accounts } from 'meteor/accounts-base';
+import { getUserById, tokenValidator } from './server_utils';
+import { Random } from 'meteor/random';
+
+const ONE_HOUR_IN_MILLISECONDS = 60 * 60 * 1000;
```
StorytellerCZ, line 5: Just use `3600000` no need to perform calculations for something so basic that doesn't change.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -0,0 +1,246 @@
...
+        });
+      });
+    }
+
+    return result;
+  },
+});
+
+/**
+ * @summary Send an email with a link the user can use to reset their password.
```
StorytellerCZ, line 209: Wrong summary.

StorytellerCZ (changes requested):
Mostly issues with comments. Also missing history entry. `accounts-base` and `accounts-password` will require patch release. If I understood the code correctly we could split changes to those two packages up into a separate PR and release that pretty much immediately.

packages/accounts-passwordless/package.js
filipenevola: I prefer going with 1.0.0 as well.

packages/accounts-passwordless/passwordless_client.js
filipenevola: Fixed

packages/accounts-passwordless/passwordless_server.js
filipenevola: Fixed

packages/accounts-passwordless/passwordless_server.js
filipenevola: Big numbers usually are represented like this so it's easier to understand, you can see many places like this in our codebase and in general as well.

StorytellerCZ: Yes, I know. But I think that could be solved by a comment and no need to process math when it isn't needed. Just a suggestion.

StorytellerCZ: I think the name of the const provides all the info for readability that is needed. :man_shrugging:

PR #11605, opened 2021-08-26
Mongo Package Async API

filipenevola:
## Why

As you are probably [aware](https://github.com/meteor/meteor/discussions/11505) we are going to replace the internals of Meteor to avoid Fibers in async executions.

The first phase was to make Meteor up-to-date with Node.js 14 (LTS until April 2021) and now we are starting the second phase: `Make Fibers usage in Meteor server application runtime optional`.

In order to allow developers to start writing code that is not going to change when we remove Fibers from internals we are already providing methods in the `mongo` package API returning `Promises` so in the future when we replace all the usages of Fibers internally your app is already ready and no changes are necessary. If you continue to write your code as you do today everything will still work as expected but you are not going to be prepared for when we remove Fibers from internals.

Meteor `mongo` package is by far the most affected package in terms of APIs that are going to be async when Fibers are removed as mongodb Node.js driver uses async calls to access MongoDB.

It's important to understand that we are not moving away from Fibers because we want to, it's a necessary change because of the direction that the Javascript ecosystem is going and we need to follow the standards to take advantage of nice new futures that are going to come up in the next years in the Node.js ecosystem.

If you are completely new to this topic please read this [discussion](https://github.com/meteor/meteor/discussions/11505) and if you have questions around Fibers and async please ask your questions in the discussion. This PR should be used to discuss changes to the `mongo` package API only.

## What

We are providing in this PR async APIs for the two main resources of `mongo` package:

- Collection: the collection class is used to call the MongoDB driver code, you use this class to connect to your collection in the database server and also to perform read and write operations.
- Cursor:  the find operation returns a Cursor and using the cursor you can iterate in the results in different ways.

## How

We are not breaking any previous standard usage of Collection and Cursor. Also this change is not required but it's recommended for new code that you are writing today. Your previous code should work fine with this change and that is why is so important that you provide feedback here.

We are creating alternative methods with the same name but appending `Async` to it, so instead of calling `find` when you are using an async collection you should call `findAsync`. The latter will return a promise.

Also in order to create the collection properly we are not going to use directly a class because when Meteor creates the collection it already connects to the MongoDB server so this class instantiation would need to be converted to a `Promise` but a class constructor should return its instance so we are wrapping this creation in a function called `createAsyncCollection`. This is still up for discussion because as you can see in the code in the end this function is also not returning a `Promise` but this was the result of multiple iterations so maybe we could iterate one more time and keep using the constructor but I personally would prefer to avoid the use of classes explicitly as it's more limited than regular functions.

## Internal Changes

We have to change some internals in order to achieve our goals but maybe you are using the internals so it's good to document here anyway:

- MongoInternals.defaultRemoteCollectionDriver function is now async, so if you need to still create it you can call it using `await` or you can still use `Promise.await` but remember that `Promise.await` is going to be removed also because it relies on a Fiber to work.
  - If you don't do this and pass the return of this function to the `_driver` option you are going to get an error explaining the same as above.
- We can now have internally two types of collection instances (sync and async) for each collection name. When you create a collection again using the same collection name by default it is going to re-use the previous instance. Except for collections named `null`.
  - We are supporting two collection instances of the same collection name to help in the migration process. You don't need to migrate all the usages of your collection in the same code change, you could migrate just one file to use async and so on.
- We don't throw errors anymore when the same collection is created twice, now we just return the previous instance if they are from the same type (async or sync). If some how the same collection with the same type is able to be instantiated again (it shouldn't happen) we throw an error unless `_suppressSameNameError` is true.
- There is a new option `ignoreInstanceReuse` that we are going to experiment during the betas. Maybe it's going to be removed if there isn't a compelling use case. Set it as `true` if you really know what you are doing. Default is `undefined`.

## How to use

When the changes in this PR are tested, ready and published we will recommend that you start to use this API in new code. The process of removing Fiber from the core is going to take months so you don't need to migrate everything now but it's good to start writing new code using this new API.

Fibers don't exist in the client so you don't need to use the new API in the client but if you want it's also supported there. We support these async operations in the client as well because it's important that you can still write methods to run in the client and in the server.

## Examples

```javascript
// api/LinksCollectionAsync.js
import { createAsyncCollection } from "meteor/mongo";

export const LinksCollectionAsync = createAsyncCollection("links");
```

```javascript
// api/linksInsert.js
import { Meteor } from "meteor/meteor";
import { LinksCollectionAsync } from "./LinksCollectionAsync";

export const insertLink = async ({ title, url }) =>
  LinksCollectionAsync.insertAsync({
    title,
    url,
    createdAt: new Date()
  });

Meteor.methods({
  async insertAsync(arg) {
    await insertLink(arg);
  }
});
```

```javascript
// server/main.js
import { Meteor } from "meteor/meteor";
import { LinksCollectionAsync } from "../api/LinksCollectionAsync";
import { insertLink } from "../api/linksInsert";

Meteor.publish("linksAsync", async () => LinksCollectionAsync.findAsync());

Meteor.startup(async () => {
  const linksCursor = await LinksCollectionAsync.findAsync();
  if (await linksCursor.countAsync()) {
    console.log(`No need to insert the first link`);
    console.log(`Links: `, await linksCursor.fetchAsync());
    return;
  }
  await insertLink({
    title: "Do the Tutorial",
    url: "https://www.meteor.com/tutorials/react/creating-an-app"
  });
  console.log(`Links: `, await linksCursor.fetchAsync());
});
```

I'm still going to update this description with more examples later:

- Examples with collection methods: findAsync,insertAsync,updateAsync,removeAsync,upsertAsync,ensureIndexAsync,dropIndexAsync,dropCollectionAsync,createCappedCollectionAsync,rawCollectionAsync,rawDatabaseAsync
- Examples with cursor methods: forEachAsync, mapAsync, fetchAsync, countAsync

## Old Deprecations 

We found some very old code during these changes and we have removed a few cases:

- Collection before Meteor 1.0 was receiving the connection in the second parameter instead of the options, [this](https://github.com/meteor/meteor/commit/5a01d9fe4d0869f388e651f48a9e51913950ca38) is not supported anymore.
- Collection in 2013 was calling the `connection` option inside the options as `manager`, [this](https://github.com/meteor/meteor/commit/6c8a4b590bac4548cb253f583f8dc74a34c13b2a) is not supported anymore. 

## Collaboration

This PR is not final, we still can discuss and make changes to it, your collaboration is important with feedback.

It's also very important that you test your project with these changes. We are still deciding which approach to recommend here but one quick way is to clone Meteor and run your project with your local copy of Meteor checked out in this branch.

We've tested this PR in a few cases but we are still going to do a lot of more testing ourselves as well.


tcastelli:
Great references and points in most of the comments. As a personal taste, I would try to stick with the naming convention proposed by the nodejs driver. I know insert is simpler than having to write insertOne or insertMany, but it's not such a big deal, and I think it  would help newcomers to understand "easily" how to use the mongo driver functions in meteor.

filipenevola:
Update on `findAsync`. 

I was able to use a sync find but using Promise.await (aka Fiber) in the top-level to initialize the connection during the code evaluation. Even using beforeStartupQueue was not enough as the collection is created in the top-level.

Not sure yet if this trade-off is worth it. It would require top-level await support in order to remove Fibers completely but maybe this is going to be necessary anyway. Comments?

I pushed the changes in the commit https://github.com/meteor/meteor/pull/11605/commits/9509c3a82336cc1e2d294649828dbca5242ccc39 so we can see if all the tests are going to pass with this set up as well.

Of course, if we decide to move this change forward we could potentially simplify other parts of this PR as well.

make-github-pseudonymous-again:
Can someone enlighten me on the use of wrapping MongoDB driver collection objects in a post-fiber universe?

packages/mongo/mongoAsyncUtils.js
```diff
@@ -0,0 +1,15 @@
...
+  let memo = undefined;
+  return async function executeOnce() {
+    if (ran) return memo;
+    const memoPromise = func.apply(this, arguments);
+    memo = await memoPromise;
+    func = null;
+    ran = true;
+    return memo;
+  };
+};
```
make-github-pseudonymous-again, line 14: Here is an alternative:

```js
export const onceAsync = (thunk) => {
  let memo;
  return async function executeOnce() {
      if (thunk /* maybe use type checking ? */) {
        const memoPromise = Reflect.apply(thunk, this, arguments);
        memo = await memoPromise;
        thunk = null;
      }
      return memo;
   };
};
```

One could also use a class to reduce the footprint of the scope bound function `executeOnce`.

make-github-pseudonymous-again: This is obvious bike-shedding. Feel free to ignore!

jamauro:
> > I'm really not a fan of having `*Async` on the name of all the methods. Already we have to use the `await` keyword which is kind of noisy. If the collection is declared async by virtue of why not simply allow the usage of the shorter names? I tend to use TypeScript in my projects so it would warn me if I was confused between a legacy sync collection and the new async ones...
> 
> For sure shorter is better but the problem here is that we want to keep the old methods available yet, maybe forever in the client.
> 
> It would be impossible for the developers to know when they are using each type of collections and the code would be very hard to understand. And we can't just rely on type definitions to make sure people are not confused as not every project is going to use Typescript.
> 
> In the future we could migrate back to shorter names after all the Fibers dependencies are removed.
> 
> Another thing to discuss here, MongoDB Node.js driver already change many names and methods, like `insertOne` and `insertMany` instead of `insert`. So an option would be to already adopt the new names in this migration but maybe we could have some conflicts with the old MongoDB Node.js driver API in a few cases returning to the same problem. Also it would increase a lot the changes in this version as we would need to implement new code for some new options of MongoDB and I would like to try to keep this PR as small as possible to avoid breaking changes.
> 
> But again, open to ideas here.

Agree that it would be ideal to avoid `*Async` on all the mongo related operations. 

Could there be a Meteor setting that you opt in to when starting up with `meteor run`? Then maybe one day it becomes the default and you can opt in to Fibers for some time before it’s completely dropped.

menewman:
> Agree that it would be ideal to avoid *Async on all the mongo related operations.
> 
> Could there be a Meteor setting that you opt in to when starting up with meteor run? Then maybe one day it becomes the default and you can opt in to Fibers for some time before it’s completely dropped.

Apologies if I'm misunderstanding this suggestion, but... To me, the benefit to having a suffixed version of Mongo methods is that it allows us to complete the migration over time by having both the Fiber and Fiber-free versions existing side-by-side in the same codebase.

If this behavior were controlled by a global setting that can only be either on or off for the whole application at any given time, then we'd be forced to migrate from Fibers -> async/await in one fell swoop, which can be much more challenging for a large, existing codebase. As someone maintaining a large, existing Meteor codebase, I very much appreciate the effort to make a gradual migration possible! I agree that it does make for some awkward naming in the meantime, but it seems like the lesser evil -- and as Filipe wrote above, perhaps the names could be swapped in a later version after the migration is complete.

zenhack:
Indeed, a gradual migration path is going to be important for my project as well. What I would do:

1. Add async versions of methods with *Async suffix. Maintain the two APIs in parallel for a while.
2. At some point, do a major version bump where you drop the non-async API, and make those names aliases for the *Async versions.
3. (optionally) at some further major version bump, drop the *Async-suffixed versions.

jamauro:
Ah, I hadn't considered that some might want to piecemeal migrate to their codebases. Still might be nice to have an option for those that want to migrate wholesale without having `*Async` everywhere.

My concern is that `*Async` never gets dropped and then the db operations are not in line with what you'd see in the `mongo` docs. Seems like a recipe for confusion and for folks to avoid adopting Meteor because it feels non-standard.

Julusian:
> Ah, I hadn't considered that some might want to piecemeal migrate to their codebases. Still might be nice to have an option for those that want to migrate wholesale without having *Async everywhere.

Except its not just your code that will make these calls. Any meteor library you use will be expecting `collection.findOne()` to be fiber and not a promise. Making every library which does mongo calls handle this will likely be very messy and prone to bugs. 

While an Async suffix isnt the prettiest, it is the best way to not break the ecosystem or force lots of work upon users or library maintainers

perbergland:
I would encourage everyone here to approach this huge change by assuming only one thing really:

It is the end of line of meteor in its first incarnation and all your code along with all packages will need to be substantially modified.

Every single call your application makes into meteor will need to be modified. And every single method in the call chain that has any of those calls needs to be changed too. 

Yes, that’s right. There will be no free lunches and no automatic upgrade to async/await instead of fiber-synchronous calls. It will just be a lot of work and headaches.

You will save time by moving your code to Typescript and/or using some other tooling (e.g. SonarCloud) to detect where in your call chain you have now forgotten to convert the calling method to be async and add an await on invocations.

The Meteor team can facilitate the move by introducing runtime warnings for all fiber-bases non-async methods along with the new *Async methods so that you can prepare for the big switch and that will require coordination.

a4xrbj1:
I'm against changing to `insertOne` and `insertMany` vs the existing `insert`.

For our production system it would require a lot of changes as we're heavily relying on MongoDb transactions that we're accessing via raw collection.

radekmie:
Together with @fredmaiaarantes we've decided that I'll continue @filipenevola's work here. As of right now, I want to set up a project to see where the current implementation is and then add tests for the new async API.

I'm fully aware of the problems coming from using the async API on the client too (https://github.com/meteor/meteor/discussions/11505#discussioncomment-1004705) and will put some thought into it as well.

:crossed_fingers:

radekmie:
@fredmaiaarantes, @filipenevola, @anyone-interested-here If possible, please do comment on the topics above. In the meantime (this week) I'll work on more tests.

Julusian:
> 1\. Right now it's Mongo.Collection.create and I'm totally fine with it.

I think that having a `create` method rather than using the constructor is good, as I think it likely that something async will want to be done during the creation. Even if it isnt needed today, having the `create` method return a Promise, and being the only way to create an async collection would be good future proofing

> 2\. As the `*Async` methods throw an error on non-async collection, I was thinking about making non-`*Async` methods throw on async collections. The idea is to ensure that the async collections are already migrated to the new API, and the removal of sync API won't affect them.

I haven't checked the code, so I can only make assumptions for now. 
It would be good if this property was named to make it clear that it only affects behaviour server-side, or if it does affect both server and client code, then it should be split into 2 properties so that they can be configured differently.
Perhaps this could have a `warn` mode, that logs a stack trace each time a non-async method is called, to help find stray calls without risking breaking your application


Personally, I am happy with `*Async` methods being added, while the method names are not as pretty it gives users time to gradually migrate code over rather than it needing to be done all at once before anything can be run. 
At work we have been missing a Promise based mongo api, so made our own wrapper around `Mongo.Collection`. We found it crucial to have that to allow us to do mongo queries in parallel, as we could feed them into `Promise.all()` and await the result of that promise. That really helped us with performance when running a bunch of slow/large mongo queries at once

packages/mongo/collection.js
```diff
@@ -39,28 +48,26 @@ Mongo.Collection = function Collection(name, options) {
       'First argument to new Mongo.Collection must be a string or null'
     );
   }
-
-  if (options && options.methods) {
```
StorytellerCZ, line 43: Removal of these should be documented.

packages/mongo/collection.js
```diff
@@ -90,7 +97,23 @@ Mongo.Collection = function Collection(name, options) {
...
+
+  this.isAsync = options.isAsync;
+
+  if (options._driver) {
+    if (typeof options._driver.open !== 'function') {
+      throw new Error('If you are creating the driver manually using new ' +
+        'MongoInternals.RemoteCollectionDriver then you need to use ' +
+        'Promise.await() or await on it. Since it is async in recent ' +
+        'versions of Meteor. ' +
+        'Read more https://docs.meteor.com/changelog.html.');
```
StorytellerCZ, line 113: This should point to a specific docs article. Even better an upgrade guide in Guide.

packages/mongo/collection.js
```diff
@@ -404,21 +433,19 @@ Object.assign(Mongo.Collection.prototype, {
 Object.assign(Mongo.Collection, {
   _publishCursor(cursor, sub, collection) {
     var observeHandle = cursor.observeChanges(
-      {
-        added: function(id, fields) {
+        {added: function (id, fields) {
```
StorytellerCZ, line 436: Formatting.

packages/mongo/mongo_driver.js
```diff
@@ -902,28 +899,59 @@ Cursor = function (mongo, cursorDescription) {
...
-    if (!self._synchronousCursor) {
-      self._synchronousCursor = self._mongo._createSynchronousCursor(
-        self._cursorDescription, {
-          // Make sure that the "self" argument to forEach/map callbacks is the
-          // Cursor, not the SynchronousCursor.
-          selfForIteration: self,
-          useTransform: true
-        });
-    }
+_.each([...CURSOR_METHODS, Symbol.iterator, Symbol.asyncIterator],
```
StorytellerCZ, line 920: Can we do without underscore?

StorytellerCZ (changes requested):
Another thing to start work on here is a guide article and changelog.

radekmie:
@Julusian
> I think that having a create method rather than using the constructor is good, as I think it likely that something async will want to be done during the creation. Even if it isnt needed today, having the create method return a Promise, and being the only way to create an async collection would be good future proofing

The point is, it'll never do anything async; at least as long as the connection will already be there. I've already suggested that it can be done in https://github.com/meteor/meteor/pull/11605#issuecomment-943084395.

> It would be good if this property was named to make it clear that it only affects behaviour server-side, or if it does affect both server and client code, then it should be split into 2 properties so that they can be configured differently.
Perhaps this could have a warn mode, that logs a stack trace each time a non-async method is called, to help find stray calls without risking breaking your application

It will affect both client and server. It is another topic that introducing async functions on the client is harder than on the backend...

@StorytellerCZ thanks for the input. Just keep in mind that it's not my branch - I haven't opened a PR yet. I'll do that later this week.

perbergland:
I see some mentions of relying on top-level awaits to be able to keep the existing pattern of creating collection instances at top level, but top-level awaits itself comes with a requirement to move from commonjs to ecmascript modules.

https://www.stefanjudis.com/today-i-learned/top-level-await-is-available-in-node-js-modules/

Without having done any work with ES modules I suspect that making that move will both break compatibility with older browsers and probably a lot of other breaking changes as well (hello reify). Can someone that knows more elaborate so we don’t just assume that it IS possible to move to ES modules and top-level await?

fredmaiaarantes:
> @fredmaiaarantes, @filipenevola, @anyone-interested-here If possible, please do comment on the topics above. In the meantime (this week) I'll work on more tests.

I will reread everything here today and share my thoughts. Thanks for pushing it!

zodern:
> Without having done any work with ES modules I suspect that making that move will both break compatibility with older browsers and probably a lot of other breaking changes as well (hello reify). Can someone that knows more elaborate so we don’t just assume that it IS possible to move to ES modules and top-level await?

We are planning to implement top-level await within our existing module system instead of switching the bundler's output to ES modules. Other bundlers have successfully done this, such as esbuild, so it is possible.

fredmaiaarantes:
I just messaged @filipenevola to see if he can clarify these questions @radekmie.

filipenevola:
Hi @radekmie sorry for the delay, my 2 cents in the open topics:

>  1. ~I'm not sure about the `createAsyncCollection` function used instead of a constructor. It's not clear what the instance exactly is, and whether one can extend its prototype. I'd rather stick to `new Mongo.Collection` but with a `{ isAsync: true }` option. Of course, it's possible to support both (that's the current state), but the question is which one should be recommended.~ Right now it's `Mongo.Collection.create` and I'm totally fine with it.

Yes, I've changed to that after our discussions here.

> 2. As the `*Async` methods throw an error on non-async collection, I was thinking about making non-`*Async` methods throw on async collections. The idea is to ensure that the async collections are already migrated to the new API, and the removal of sync API won't affect them.
    * This can be an option, e.g., `throwOnSyncMethods`.
    * It affects the cursors as well.

I agree. It should be restrictive on mixing the two approaches in both sides.

> 3. What kind of tests are needed to merge this PR? Rewriting all of the `mongo_livedata_tests.js` using the async API is going to require a lot of work.

We don't NEED anything else to make sure the old methods are still working. This work is done. What we SHOULD is add new tests to validate Async execution paths, but as they are going to be 100% new I'm not too afraid as we can get errors and problems in the beta phase and then add these tests to cover specific cases and avoid regressions, we did this in the MongoDB 5 oplog changes and in the end, the result was great.

> 4. Why was [this test](https://github.com/meteor/meteor/pull/11605/files?w=0#diff-f874462c9a046e2f37f8da6bd429c05cf8d319dd546a3ee5152cc0ab7acdb89bL2515) removed?

I explained in my [commit](https://github.com/meteor/meteor/commit/0bb3a32657f2a6459eb0c33e8f80e4a092be4f98) message. TL;DR: it was testing a case that doesn't exist.

> 5. What is [this change](https://github.com/meteor/meteor/pull/11605/files?w=0#diff-95a11e01f1e51ce0defae5db23471228dd2f6d6a7fa646a4a9eb3f6ff2bee68bR29) about?

Also explained in the [commit](https://github.com/meteor/meteor/commit/cd160b7a9cbdfb01a24b3796056c9dc3ee1f5671). TL;DR: adding an option to test only one package tests on test-in-console/run.sh

I'm open to helping with other topics as well, feel free to send me an email (filipe@quave.dev) if I don't get back soon enough, I'm not tracking my GitHub notifications anymore.

radekmie:
As I opened #12028 just now, this one can be closed.

PR #11607, opened 2021-08-27
Add WebApp.addUpdatedNotifyHook

brucejo75:
Add another hook to the `webapp` package.

### `WebApp.addUpdatedNotifyHook`
This function will add a hook to notify the caller whenever meteor has updated its code.  This hook would mostly be used in development to enable the caller to perform actions whenever meteor has updated itself.

@filipenevola, @StorytellerCZ,
This is a hook that I require to enable `autoupdate` when I am developing with a meteor server shared among multiple client applications.  Super useful during development.

Hoping we can sneak this into 2.4 while I was updating `webapp` anyway.

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -412,17 +412,32 @@ function getBoilerplateAsync(request, arch) {
   }));
 }
 
+// Notification hook whenever the runtime configuration is updated
+// hook will pass an object with the following fields:
+// - arch
+// - manifest
+// - runtimeConfig
+WebApp.addUpdatedNotifyHook = function(hook) {
+  if(typeof hook !== 'function') throw new Error('WebApp.addUpdatedNotifyHook must be a function');
+  runtimeConfig.updateHooks.push(hook);
+}
+
 WebAppInternals.generateBoilerplateInstance = function (arch,
                                                         manifest,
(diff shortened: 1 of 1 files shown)


brucejo75:
Pushed updates to the documentation: https://github.com/meteor/docs/pull/738#issuecomment-907904225

packages/webapp/webapp_server.js
```diff
@@ -412,17 +415,32 @@ function getBoilerplateAsync(request, arch) {
   }));
 }
 
+// Notification hook whenever the runtime configuration is updated
+// hook will pass an object with the following fields:
+// - arch
+// - manifest
+// - runtimeConfig
+WebApp.addUpdatedNotifyHook = function(hook) {
```
StorytellerCZ, line 423: We have the `callback-hook` package which could have been used here.

StorytellerCZ (commented):
Missing an entry in history file.

packages/webapp/webapp_server.js
StorytellerCZ: We still have some time before release, so that would be great! Just make sure that it works as you desired.

brucejo75: done: https://github.com/meteor/meteor/pull/11615

PR #11608, opened 2021-08-28
Disable native file watching during meteor create

zodern:
This should fix Meteor getting stuck when creating the first app after it is installed on Windows. This is a temporary fix until we replace the library we use for using native file watchers.


filipenevola (approved):
LGTM

PR #11614, opened 2021-09-01
Add mobileSafariUI as alias for Safari Mobile

### LINKED ISSUE #11612: Meteor always determines that ios wkwebview is a legacy browser even though it is on par with safare
Ended up running into an issue where my code was running as the legacy bundle on ios, which was causing some issues.

Found out that I needed to call
```
setMinimumBrowserVersions(
    {
      "mobileSafariUI/WKWebView": 10,
    },
    "classes"
  );
```
To be able to mark it as a modern browser like it is.

This seems to be missing form the modern-browsers package either as an alias or as an additional item listed.

Seems like it would be a good default to include.

StorytellerCZ:
Fixes #11612 by adding variants of mobileSafariUI to the aliases for Mobile Safari.


packages/modern-browsers/modern.js
```diff
@@ -30,7 +30,7 @@ const browserAliases = {
 
   // The webapp package converts browser names to camel case, so
   // mobile_safari and mobileSafari should be synonymous.
-  mobile_safari: ["mobileSafari"],
+  mobile_safari: ["mobileSafari", "mobileSafariUI", "mobileSafariUI/WKWebView"],
```
renanccastro, line 33: https://github.com/meteor/meteor/blob/ffcfa5062cf1bf8a64ea64fef681ffcd99fe7939/packages/meteor/cordova_environment.js#L14

Are you sure this works for cordova? I saw this code and it might seem we are defaulting to legacy always on cordova.

StorytellerCZ: Yes, that is what I thought, but for @schlaegerz this worked:
```js
setMinimumBrowserVersions(
    {
      "mobileSafariUI/WKWebView": 10,
    },
    "classes"
  );
```
And so did this PR. I think this is not related to Cordova as from my understanding. More like a Meteor app being displayed through some other app instead of being build via Cordova (I have seen this done in the past).

Here is an article about the different views:
https://developer.apple.com/news/?id=trjs0tcd

schlaegerz: Yes that is correct this is not Cordova, for various reasons we end up hosting a webview inside of our app that points to our Meteor app. It is for sure a weird edge case that is not too common, but this change definitely matches reality more since the WKWebview is completely a modern browser.

renanccastro (approved):
If it's not targeting Cordova, that's fine!

schlaegerz:
I did end up noticing that some ipads use the user agent that resolves to appleMail. So I ended up defining its minimum version too, but it is not an alias to these other ones since its version is like 600+

Not sure what the right thing to do in the meteor repro is, but i just added added a minimumBrowserVersion call in my code to make sure it is marked as modern.

StorytellerCZ:
What the hell is Apple doing?! @schlaegerz if you could find some documentation about this then we could address this.

PR #11621, opened 2021-09-06
Revert chalk upgrade for IE11

StorytellerCZ:
IE11 evaluates server code which makes `chalk` break it. As such this sync with the meteor-tool logging utility is being reverted for the time being until a better solution is found (or we move to Winston).


filipenevola (changes requested):
The PR description doesn't match with the change.

What is the idea?

StorytellerCZ:
@filipenevola updated

PR #11624, opened 2021-09-08
Use symbolic links for transpiling chalk and make it not throw parse errors with IE11

renanccastro:
This commit creates a package.json for using the node_modules folder, and it does link the chalk dependency to the root of the package so meteor transpiles it.

The error was happening because IE11 doesn't support ecmascript 2019 features, and thus needs to be served a legacy bundle. Chalk wasn't being transpiled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 os: linux
 dist: xenial
 node_js:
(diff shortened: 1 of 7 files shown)


StorytellerCZ (approved):
Tested on Windows + IE11 and it works.

PR #11628, opened 2021-09-10
Fix typo in package-source.js

eltociear:
preceeding -> preceding


StorytellerCZ:
@eltociear Thanks for the fix!

PR #11630, opened 2021-09-16
Update terser to 5.8

StorytellerCZ:
Updates terser to the latest version, which in our case is one major version bump.

From breaking side it means this:
* BREAKING: minify() is now async and rejects a promise instead of returning an error.
* BREAKING: Internal AST is no longer exposed, so that it can be improved without releasing breaking changes.
* BREAKING: Lowest supported node version is 10
* BREAKING: There are no more warnings being emitted

Here are some highlights from new features:
* Optional chaining syntax is now supported.
* `source-map` library has been updated, bringing memory usage and CPU time improvements when reading input source maps (the SourceMapConsumer is now WASM based).
* Accept `{get = "default val"}` and `{set = "default val"}` in destructuring arguments.
* Performance improvements
* Many bug fixes

Full changelog: https://github.com/terser/terser/blob/master/CHANGELOG.md

This also fixes few bugs we were getting.

Currently I believe that the code could be improved. Also we have to decide how this should be released. Since it fixes bugs on our end there is an argument to release it as patch, but it is also a big update so argument could be made for major or feature release (feature probably best given how much issues major release presents).

### DIFF
--- a/History.md
+++ b/History.md
@@ -9,6 +9,12 @@
 * `modern-browsers@0.1.6`
   - Added `mobileSafariUI` as an alias for Mobile Safari
 
+* `minifier-js@2.7.1`
+  - Updated `terser` to [v5.8.0](https://github.com/terser/terser/blob/master/CHANGELOG.md#v580) to fix various bugs
+
+* `standard-minifier-js@2.7.1`
+  - Updated `@babel/runtime` to [v7.15.4](https://github.com/babel/babel/releases/tag/v7.15.4)
+
 ## v2.4, 2021-09-15
 
 #### Highlights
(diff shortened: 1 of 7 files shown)


vlasky:
I suggest an immediate release as a patch. For us, the Meteor 2.4 build process is broken due to a terser minification error that should not be happening.

renanccastro (approved):

filipenevola:
Hi, `minifier-js@2.7.1` and `standard-minifier-js@2.7.1`.

A new `meteor update --release 2.4` should update these packages for you if you are already running Meteor 2.4.

PR #11638, opened 2021-09-21
Upgrade to cordova-android@10

renanccastro:
Required changes
- use WebViewAssetLoader instead of the local port mechanism for intercepting URLs (webapp plugin update)
- Change usage of cordova-android api, it's not included anymore in the generated project
- AndroidXEnabled is now true
- Allow clear text Traffic(http) on android for development purposes


StorytellerCZ:
Looks like `tools/tests/cordova-builds.js` will need some review as well.

packages/webapp/package.js
```diff
@@ -20,9 +20,9 @@ Npm.strip({
   useragent: ["test/"]
 });
 
+// whitelist plugin is now included in the core
```
StorytellerCZ, line 23: Nice comment about the change, but that should have been just a comment here, no need to have it in the code itself.

tools/cordova/builder.js
```diff
@@ -115,7 +115,7 @@ export class CordovaBuilder {
       author: 'A Meteor Developer',
       email: 'n/a',
       website: 'n/a',
-      contentUrl: `http://localhost:${cordovaServerPort}/`
```
StorytellerCZ, line 118: I'll need a bit of context on this removal.

StorytellerCZ (changes requested):
Finish TODO, other than that it looks pretty good.

tools/cordova/builder.js
renanccastro: I've rolled back, it's actually not necessary and would introduce issues on multiple apps on iOS

StorytellerCZ (approved):

filipenevola (approved):

Issue #11635, opened 2021-09-21
Meteor 2.4 - chalk npm package incompatible with IE11, makes app crash on modules.js load


perbergland:
After upgrading our app to meteor 2.4, it no longer loads in IE11 because one of the bundled npm packages in modules.js causes a syntax error which prevents further loading.

This time it seems to be "chalk". I’m not sure exactly which ecmascript feature that is used here and that is incompatible.

![image](https://user-images.githubusercontent.com/2690254/134184846-c0431c01-a2c8-4ccc-9729-fd749084b739.png)

![image](https://user-images.githubusercontent.com/2690254/134184910-7275338b-3171-4b50-8fde-af6784562062.png)

StorytellerCZ:
We have run into this problem during the RC stage and me and @renanccastro worked on a fix and all the tests and manual testing have passed, so this is super strange to me. I will look into this again as the recompiling for the legacy is what should have happened already.
@perbergland can you provide us any additional details about your setup with IE11 that might help. Thanks!

perbergland:
I just use a vanilla IE10 on Win10 with browserstack.

I upgraded my app from 2.3.5 without any attempt to clean out old builds or anything else in .meteor/local

StorytellerCZ:
Interesting. :thinking: I think there might be something to your theory. Once I'm back on my Windows I will give it a try.

renanccastro:
Hi @perbergland 
can you please try the package logging@1.3.1 ?
It should have been fixed in this version!

perbergland:
Yes the updated logging package fixes the problem and the workaround is no longer necessary

PR #11639, opened 2021-09-21
Only include chalk on server - avoid serving it to the client, even if not evaluated

renanccastro:
fix issue #11635

On the 2.4 release, we added chalk as a dependency in the logging package, but it was being included in the client even if not evaluated. For fixing that, I added https://github.com/meteor/meteor/pull/11624, which didn't actually solve the issue but made the code in the client not crash on IE.

After further analysis, we should never ship chalk for the client, and this PR fixes that by splitting into two files that end up in the correct bundle, making the tree analyzer work correctly and not include the dependency for the client.


filipenevola (approved):
LGTM

PR #11645, opened 2021-09-24
Release 2.5

filipenevola:
The focus of this release is to upgrade the integration between Meteor and Cordova for Android apps.

As Google is going to require apps to be using API version 30 or above in October we are going to release this new version as soon as possible.

We recommend to everybody with Android apps to test our latest beta or RC asap.

All included items are listed [here](https://github.com/meteor/meteor/milestone/83).

### DIFF
--- a/packages/typescript/package.js
+++ b/packages/typescript/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: "typescript",
-  version: "4.3.5",
+  version: "4.4.0-beta250.0",
   summary: "Compiler plugin that compiles TypeScript and ECMAScript in .ts and .tsx files",
   documentation: "README.md"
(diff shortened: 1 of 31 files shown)


packages/typescript/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   name: "typescript",
-  version: "4.3.5",
+  version: "4.4.0-beta250.0",
```
StorytellerCZ, line 3: If we are already updating the version here, we should also bump the Typescript version in the tool and meteor-babel and go to 4.4.3 to keep the versioning schema that was started.

StorytellerCZ (changes requested):

PR #11649, opened 2021-09-28
accounts-base set config from Meteor.settings

StorytellerCZ:
Allow setting `Account.config` from Meteor settings. This works for all options unless function is specified. This is to make it quicker to setup accounts settings without the need to write code (and for most settings it is booleans or strings). This also allows to quickly change settings on Galaxy without the need to deploy a new version of the app as you have to now.
The old functionality is still working, but settings from Meteor settings will take precedent.

### DIFF
--- a/packages/accounts-base/accounts_common.js
+++ b/packages/accounts-base/accounts_common.js
@@ -75,8 +75,8 @@ export class AccountsCommon {
       this.loginServiceConfiguration = ServiceConfiguration.configurations;
       this.ConfigError = ServiceConfiguration.ConfigError;
 
-      if(Meteor.settings?.packages?.accounts) {
-        const settings = Meteor.settings.packages.accounts;
+      const settings = Meteor.settings?.packages?.['accounts-base'];
+      if(settings) {
         if(settings.oauthSecretKey) {
           if (! Package["oauth-encryption"]) {
             throw new Error("The oauth-encryption package must be loaded to set oauthSecretKey");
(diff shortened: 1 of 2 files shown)


packages/accounts-base/accounts_common.js
```diff
@@ -66,6 +74,27 @@ export class AccountsCommon {
       const { ServiceConfiguration } = Package['service-configuration'];
       this.loginServiceConfiguration = ServiceConfiguration.configurations;
       this.ConfigError = ServiceConfiguration.ConfigError;
+
+      if(Meteor.settings?.packages?.accounts) {
+        const settings = Meteor.settings.packages.accounts;
```
filipenevola, line 79: The key should be called `accounts-base` and not `accounts`.

packages/accounts-base/accounts_common.js
```diff
@@ -66,6 +74,27 @@ export class AccountsCommon {
       const { ServiceConfiguration } = Package['service-configuration'];
       this.loginServiceConfiguration = ServiceConfiguration.configurations;
       this.ConfigError = ServiceConfiguration.ConfigError;
+
+      if(Meteor.settings?.packages?.accounts) {
```
filipenevola, line 78: You could get the value first and then later test it instead of testing and later getting it.

filipenevola (changes requested):

filipenevola (approved):
Apply prettier in the new code 😉

PR #11650, opened 2021-09-28
Service configuration from Meteor.settings

StorytellerCZ:
This PR makes `service-configuration` take the services settings from Meteor settings and adds/updates them in the DB. This reduces repetitive code in apps and allows to quickly change these settings on Galaxy without the need to deploy a new version of an app.

Current approach should take care of any service configuration available. We could check the structure for officially supported services or better document what are the keys names. Thoughts?
Also there is a question of how to define `siteUrl` which we could either leave to the developers to handle manually or make this automatic, but there is a problem that for example `meteor-developer` and `google` differ on if they want the final `/` or not. Potentially we could make a fix to `google` package to properly handle input from `Meteor.absoluteUrl()`


packages/service-configuration/service_configuration_server.js
```diff
@@ -28,3 +30,17 @@ try {
     );
     throw err;
 }
+
+Meteor.startup(() => {
+    if (!Meteor.settings?.packages?.['service-configuration']) return;
+    const settings = Meteor.settings?.packages?.['service-configuration']
```
filipenevola, line 36: You could get the property first and then return or proceed

```suggestion
    const settings = Meteor.settings?.packages?.['service-configuration'];
    if (!settings) {
      return;
    }
    // proceed
```

filipenevola (changes requested):

filipenevola (approved):
I don't think you know that but we have prettier configured in the root project.

We should format with prettier every new code section.

We don't format everything because it was going to cause a huge diff but new code sections should follow the new prettier format 😉

filipenevola:
> We could check the structure for officially supported services or better document what are the keys names. Thoughts?
Also there is a question of how to define siteUrl which we could either leave to the developers to handle manually or make this automatic, but there is a problem that for example meteor-developer and google differ on if they want the final / or not. Potentially we could make a fix to google package to properly handle input from Meteor.absoluteUrl()

I don't think I fully understand the question and the comment as well. Could you clarify what needs to be defined here?

StorytellerCZ:
> I don't think I fully understand the question and the comment as well. Could you clarify what needs to be defined here?

Let's use meteor.com as an example. meteor-developer and Google services require `siteUrl`. Google wants it in a format `https://www.meteor.com`, while meteor-develop will take output of `Meteor.absoluteUrl()` as is (`https://www.meteor.com/`), the difference being in the trailing `/`. So I can't add `siteUrl: Meteor.absoluteUrl()` to the upsert.

filipenevola:
@StorytellerCZ but why `siteUrl` is not a property inside each service definition instead of being automatic defined using `Meteor.absoluteUrl()`? 

Or at least you could overwrite putting it in the service definition avoiding the automatic resolution of `Meteor.absoluteUrl()`.

StorytellerCZ:
@filipenevola Never mind. Upon further investigation it looks like it was just my mistaken understanding from an old code in my application. I think it stemmed from an issue in google-config-ui, see #11657

StorytellerCZ:
Anyhow removed the now unneeded comment and this is ready to merge.

PR #11657, opened 2021-10-01
Deliver siteUrl in google-config-ui as other config-ui packages

StorytellerCZ:
`google-config-ui` was needlessly deleting the trailing `/` instead of just removing it from the html like all the other config-ui packages do. This reduces complexity and standardizes with the other packages (and now linted/pretied as well).


StorytellerCZ:
Published in google-config-ui@1.0.3

PR #11664, opened 2021-10-01
Fix infinite loop in import scanner when file is on a different drive than source root

zodern:


zodern:
With these changes Meteor can now start an app when the app is on the C drive, and Meteor is running from checkout on a different drive.

filipenevola (approved):

PR #11681, opened 2021-10-06
Passwordless for accounts-ui

StorytellerCZ:
Implements `accounts-passwordless` for `accounts-ui`.

Also fixes two bugs that I have encountered while testing this.

UI look:

Stand alone:
alone

Code entry look:
code

With accounts-password (for strange or legacy cases):
with-password

With username in account creation:
with-username

With all the options:
with-all


packages/accounts-passwordless/passwordless_client.js
```diff
@@ -120,3 +121,6 @@ Accounts.autoLoginWithToken = function() {
     }
   });
 };
+
+// Run check for login token on page load
+document.addEventListener('DOMContentLoaded', () => Accounts.autoLoginWithToken())
```
StorytellerCZ, line 126: Auto login wasn't running by default so I had to add this so that when clicking on a link will login users.

packages/accounts-passwordless/email_templates.js
```diff
@@ -1,33 +1,37 @@
...
- * @importFromPackage accounts-base
- */
-Accounts.emailTemplates = {
-  ...(Accounts.emailTemplates || {}),
-  sendLoginToken: {
-    subject: () => `Your login token on ${Accounts.emailTemplates.siteName}`,
-    text: (user, url, { sequence }) => {
-      return `Hello!
+// Accounts.emailTemplates need to be in Meteor.startup or they will be undefined nad what is added here will be overridden
+// once they get instantiated.
```
StorytellerCZ, line 2: Encountered this when having `accounts-password` package present.

StorytellerCZ:
@filipenevola as per your request I have cherry picked the code only related stuff to #11687

Issue #11685, opened 2021-10-07
Webapp response for POST


milanzigmond:
To reproduce the problem:

1. meteor create Example --release 2.2 (terminal 1)
2. run meteor (terminal 1)
3. in second terminal try: curl -X POST -v http://localhost:3000 (terminal 2)
4. I get 200 OK with the content <img width="1440" alt="Screenshot 2021-10-07 at 14 37 47" src="https://user-images.githubusercontent.com/4190850/136419435-4294f262-4c69-4c49-beeb-734a2199702c.png">
5. CTRL + C (stop server, terminal 1)
6. meteor update (update meteor to 2.4, terminal 1)
6. run meteor (terminal 1)
7. in second terminal try: curl -X POST -v http://localhost:3000 (terminal 2)
8. I get 405 <img width="1440" alt="Screenshot 2021-10-07 at 14 39 49" src="https://user-images.githubusercontent.com/4190850/136420371-411f82e3-d4a2-45ea-82a8-ee64532c6f6a.png">

This is breaking our app at the moment as it runs in Facebook iframe and it is doing post request initially. So we cannot go higher than 2.2, since Webapp responses were changed in 2.3.1.

I understand it's more safe to handle the requests by your own rules, just don't see any examples how exactly this should be done.

My questions are: 
1. How do I again allow POST requests on my meteor server
2. How do I again return the content on this POST request like it was before

Thank you
.

StorytellerCZ:
This is the PR that has caused the issue for you: https://github.com/meteor/meteor/pull/11224
It might be possible maybe through WebApp, but I have decided to quickly create a new settings option to provide for your use case.

StorytellerCZ:
Released in Meteor 2.5-beta.4

PR #11686, opened 2021-10-07
WebApp option to always return content

StorytellerCZ:
Adds a new settings option `Meteor.settings.packages.webapp.alwaysReturnContent` to allow bypass changes done in #11224 and revert to the old behavior as per special use cases in #11685 
`packages/webapp/webapp_server.js` was linted, so look for comments for the actual changes.

Tested on reproduction in #11685


StorytellerCZ:
Will be released in Meteor 2.5-beta.4

PR #11698, opened 2021-10-15
Fix typo in compiler.js & lint

StorytellerCZ:
* Fix typo identified if #11692
* Lint the file (I have let WebStorm run the full suite on it)
* Replace few underscore functions with native functions (specifically `_.values` => `Object.values`)


WilliamKelley:
> If we want to reformat the tool code, we should find a way to do it that minimizes the effect on git blame.

There's a fairly standard convention of using an "revisions to ignore" file that's used with `git blame --ignore-revs-file <filename>`. For example, https://github.com/mui-org/material-ui/blob/master/.git-blame-ignore-revs .

PR #11708, opened 2021-10-20
Accounts UI config via settings

StorytellerCZ:
While working on accounts presentation and #11707 I have noticed that we also have extra config for Accounts.ui. I'm adding here the option to set that config from settings like the other Accounts.config option so that we have the same principle all across.

> Also included is documentation addition to utilize the inclusion of docs in this repo and showcase how future PRs should look like.


filipenevola (approved):
Looks ok, we just need to update the docs to refer to the correct version

Issue #11709, opened 2021-10-21
Exception thrown when calling forgot password with --production flag and MAIL_URL set


santiagopuentep:
On Meteor 2.4, calling `Accounts.forgotPassword` with the `--production` flag active and `MAIL_URL` is set with real keys throws this exception:

```
Exception while invoking method 'forgotPassword' TypeError: Cannot read property '_syncSendMail' of null
    at smtpSend (packages/email/email.js:154:13)
    at Object.Email.send (packages/email/email.js:233:5)
    at Object.Email.send (packages/mdg:meteor-apm-agent/lib/hijack/email.js:10:28)
    at AccountsServer.Accounts.sendResetPasswordEmail (packages/accounts-password/password_server.js:667:9)
    at MethodInvocation.forgotPassword (packages/accounts-password/password_server.js:478:12)
    at MethodInvocation.methodMap.<computed> (packages/mdg:meteor-apm-agent/lib/hijack/wrap_session.js:161:30)
    at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1885:12)
    at packages/ddp-server/livedata_server.js:769:19
    at Meteor.EnvironmentVariable.EVp.withValue (packages\meteor.js:1257:12)
    at packages/ddp-server/livedata_server.js:767:46
    at Meteor.EnvironmentVariable.EVp.withValue (packages\meteor.js:1257:12)
    at packages/ddp-server/livedata_server.js:765:46
    at new Promise (<anonymous>)
    at Session.method (packages/ddp-server/livedata_server.js:739:23)
    at packages/mdg:meteor-apm-agent/lib/hijack/wrap_session.js:52:38
    at Meteor.EnvironmentVariable.EVp.withValue (packages\meteor.js:1257:12)
    at Session.sessionProto.protocol_handlers.method (packages/mdg:meteor-apm-agent/lib/hijack/wrap_session.js:51:44)
    at packages/ddp-server/livedata_server.js:603:43
```

`smtpSend()` receives no transport, and there's no check of existence before calling one of it's function.

### Expected behavior:
No exception should be thrown.

megawebmaster:
@StorytellerCZ - I believe this is caused by missing mail setup in production mode. In this commit https://github.com/meteor/meteor/commit/c704a16935f92685b87b9a3b8d6252ee8b981a10 there is a change that removed check for transport being non-null before calling `smptSend`.

Do you know why was this change introduced?

santiagopuentep:
[...]

PR #11710, opened 2021-10-21
Fix email logic resolution on getTransport

StorytellerCZ:
In response to #11709 I'm adding more tests during which I have found one more edge case to account for in a logic and adding back a resolution for a logic error that I have fixed some while back, but can't seem to find in the code. Once I can replicate #11709 then I will create a test for it as well.


packages/email/email.js
```diff
@@ -228,8 +240,9 @@ Email.send = function (options) {
     customTransport({ packageSettings, ...options });
     return;
   }
-  if (Meteor.isProduction || process.env.MAIL_URL || Meteor.settings.packages?.email) {
+  if (Meteor.isProduction && process.env.MAIL_URL || Meteor.settings.packages?.email) {
```
denihs, line 243: I think you want to test here if it's production AND if there is an email URL, correct? If yes, you're missing some parenthesis:

```
if (Meteor.isProduction && (process.env.MAIL_URL || Meteor.settings.packages?.email)) {
```

&& has precedence over ||

denihs (changes requested):

denihs:
@StorytellerCZ Were you able to reproduce the issue?

packages/email/email.js
StorytellerCZ: That is something we need to decide. Both the fixed and your proposed conditions are valid options.

denihs: So you're saying that if it's a development or staging environment but the `Meteor.settings.packages.email` is set, this code should run?

I don't think both conditions are valid because one says that this code should run in production if an email URL is set, as the other says that you can use settings to overwrite a behavior that should occur just in production.

Can you tell what the purpose of this code is? 

I'm not sure which one is correct. I will try to understand this code to see what it's supposed to happen there.

denihs:
@StorytellerCZ, we analyzed and concluded the logic you changed was supposed to be like that, but some improvements should be made. So we did them [here](https://github.com/meteor/meteor/pull/11891). As soon as the next beta is out, I will let the people on [this](https://github.com/meteor/meteor/issues/11709) issue know, so they test out.

PR #11711, opened 2021-10-22
Support Mac M1 Architecture

renanccastro:
Adds support for arm64 darwin platform.

Quirkes:
- MongoDB doesn't have any binaries for m1, so rosetta is being used
- Node is not publishing official binaries for node 14, so we are building it ourserlves.


sergeyfilimonov:
Just add my vote to add the support of M1 faster as running it via Rosetta 2 is OK but it's still not going to be as fast and efficient as a native build.

tosinek:
Will this enable an option to easily deploy to ARM based containers on AWS and eg Raspberry PI?

PR #11724, opened 2021-10-27
GitHub OAuth allow_signup & more data returned

StorytellerCZ:
Two new features for GitHub OAuth.
1. Based on Accounts config pass in `allow_signup` param on the initial request to prevent sign ups on GitHub side if the user does not have an account already.
2. After data about the user is retrieved get more of it into `serviceData` so that they are available to the app. We will receive the data anyway, so might as well use it.


packages/github-oauth/github_client.js
```diff
@@ -25,12 +25,17 @@ Github.requestCredential = (options, credentialRequestCompleteCallback) => {
 
   const loginStyle = OAuth._loginStyle('github', config, options);
 
+  let allowSignup = '';
+  if (Accounts._options?.forbidClientAccountCreation)
+    allowSignup = '&allow_signup=false'; // https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#parameters
```
denihs, line 30: I know it's not necessary, but just for the sake of convention, can you add `{}` to this `if`? 

```suggestion
  if (Accounts._options?.forbidClientAccountCreation) {
    allowSignup = '&allow_signup=false'; // https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#parameters
  }
```

denihs (changes requested):

denihs (approved):

PR #11754, opened 2021-11-10
Feature/installer adjustments

renanccastro:
- Support installing as root
- Warn when running as root without --unsafe-perm, as we cant create any directories
- Warn to include the meteor bin in the current PATH env var, to avoid having to open another terminal window


filipenevola (approved):
LGTM

PR #11755, opened 2021-11-11
rate-limit: convert to token-bucket like rate-limiting

nathan-muir:
Hi,

This pull request is to implement token-bucket style rate-limiting, atop of the existing `rate-limit` / `ddp-rate-limiter` api.

The current system is very good at quotas that reset frequently, but difficult to tune for longer periods.

This change, gives the request an initial quota amount, and tops it up gradually.

eg, with a quota of `10` per `30` minutes, the user would have 10 requests initially, but also gain quota for one new request every 3 minutes.

### DIFF
--- a/packages/rate-limit/rate-limit.js
+++ b/packages/rate-limit/rate-limit.js
@@ -8,6 +8,19 @@ const DEFAULT_REQUESTS_PER_INTERVAL = 10;
 
 const hasOwn = Object.prototype.hasOwnProperty;
 
+function compatOptions(options) {
+  if (options.numRequestsAllowed == null) {
+    return options;
+  }
+  const { numRequestsAllowed, intervalTime, callback } = options;
+  return {
+    callback,
+    initialCapacity: numRequestsAllowed,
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Can you please increment the package version and add info into the history file and update documentation please?

filipenevola:
Hi @nathan-muir, thank you for your PR.

I would like to see the doc (the docs are now in the same repo so you should include the docs changes in this PR) showing these new options and also the use cases. This would make the review more complete.

Also, it would be good to have new test cases for these new options.

Could you work on that?

Thanks!

nathan-muir:
Hi @filipenevola,

Before I launch into docs & tests, I just wanted to discuss this change, and see if it is something likely to be accepted

Secondly, this doesn't yet expose any new options _yet_, it is just an internal behaviour change; We could add an option to configure the initial capacity of the bucket, separate from the maximum capacity.

Also, I could make it support the existing rate limit algorithm, and switch depending on an argument.

As for understanding the difference of the two algorithms, a graph may be the best way:

image

PR #11757, opened 2021-11-12
[Facebook OAuth] Use Facebook API v12

StorytellerCZ:
Update Facebook API to v12. No changes to OAuth has been made, so just bumping the number is enough. Since this is also not anything that needs to be done and the v10 of the API will be supported till June 8th, 2023 I have put it as a feature upgrade to be included in the next feature release of Meteor.
* [v11 changelog](https://developers.facebook.com/docs/graph-api/changelog/version11.0)
* [v12 changelog](https://developers.facebook.com/docs/graph-api/changelog/version12.0)


denihs:
[v13.0](https://developers.facebook.com/docs/graph-api/changelog/version13.0) was released on Feb 08, so I've already updated this version.

PR #11758, opened 2021-11-12
Work around terser bug.

### LINKED ISSUE #11756: Minification issue with flow-router
At some point post-2.3.4, the minification process of JS files seems to have broken in some way. To reproduce, create a plain app and add `kadira:flow-router`:

```
meteor create repro
cd repro
meteor add kadira:flow-router
meteor run --production
```

Then look at the minified source code:

Screen Shot 2021-11-11 at 3 40 46 PM

Compare this to the actual function -- the minification seems to have dropped the last part of `doRedirect`!

https://github.com/kadirahq/flow-router/blob/master/client/triggers.js#L91

brianlukoff:
This is intended to fix #11756, by disabling the evaluation option in terser's compressor.

### DIFF
--- a/packages/minifier-js/minifier.js
+++ b/packages/minifier-js/minifier.js
@@ -22,6 +22,7 @@ export const meteorJsMinify = function (source) {
       drop_debugger: false,  // remove debugger; statements
       unused: false,         // drop unreferenced functions and variables
       dead_code: true,       // remove unreachable code
+      evaluate: false,       // work around issue in terser (possibly https://github.com/terser/terser/issues/837)
       typeofs: false,        // set to false due to known issues in IE10
       global_defs: {
         "process.env.NODE_ENV": NODE_ENV
--- a/packages/minifier-js/package.js
+++ b/packages/minifier-js/package.js
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Npm.depends({
-  terser: "5.8.0"
+  terser: "5.9.0"
 });
 
 Package.onUse(function (api) {


filipenevola:
Hi @brianlukoff thank you for this investigation.

The upgrade in the version is required to use this option or what was the reason for this bump?

brianlukoff:
Just figured it was worthwhile upgrading while we were at it, but I don't think it is strictly necessary as `evaluate` is available in 5.8.0 as well.

filipenevola (approved):
I've read the 5.9.0 change log and it seems it only contains bug fixes.

filipenevola:
I'm waiting all the tests to complete.

Also, I reinstalled AppVeyor integration as it is not communicating properly with GitHub but the tests are running [here](https://ci.appveyor.com/project/meteor/meteor/builds/41514366).

filipenevola:
AppVeyor integration is back 🎉

filipenevola:
Thank you @brianlukoff. This fix is published in minifier-js@2.7.2.

brianlukoff:
Thanks @filipenevola!

packages/minifier-js/minifier.js
```diff
@@ -22,6 +22,7 @@ export const meteorJsMinify = function (source) {
       drop_debugger: false,  // remove debugger; statements
       unused: false,         // drop unreferenced functions and variables
       dead_code: true,       // remove unreachable code
+      evaluate: false,       // work around issue in terser (possibly https://github.com/terser/terser/issues/837)
```
StorytellerCZ, line 25: This has caused #11775

PR #11767, opened 2021-11-16
Improve meteor-installer - avoid local install of npm installer package

renanccastro:
We are now checking the npm env var for making sure the user is not installing the meteor-installer package locally.

This env var is not officially documented(at least I couldn't find it!), but there are some mentions to it, like here: https://github.com/yarnpkg/yarn/issues/2472#issue-201139405

Pending tasks:

- [X] Test on Linux
- [ ] Test on Mac
- [ ] Test on Windows


lmf-git:
Test on Mac -> Fails

PR #11774, opened 2021-11-19
Add lastLoginAt to the login funtion

StorytellerCZ:
Want to start a discussion about this. This PR adds a new field `lastLoginAt` to the login function which is added the date when the last login was triggered for the user. This information is pretty common in user systems and I have a need for it for a functionality that I'm working on in my app. I want to start a discussion with this PR if the location I have put it in is the correct place and if it can be improved (or scratched altogether).

Also while going through I have updated `each` calls on callback-hooks to the new `forEach` in places that I have came across.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -245,14 +245,14 @@ export class AccountsServer extends AccountsCommon {
   };
 
   _successfulLogin(connection, attempt) {
-    this._onLoginHook.each(callback => {
+    this._onLoginHook.forEach(callback => {
       callback(cloneAttemptWithConnection(connection, attempt));
       return true;
     });
   };
 
   _failedLogin(connection, attempt) {
-    this._onLoginFailureHook.each(callback => {
+    this._onLoginFailureHook.forEach(callback => {
       callback(cloneAttemptWithConnection(connection, attempt));
       return true;
     });
@@ -261,7 +261,7 @@ export class AccountsServer extends AccountsCommon {
   _successfulLogout(connection, userId) {
     // don't fetch the user object unless there are some callbacks registered
     let user;
-    this._onLogoutHook.each(callback => {
+    this._onLogoutHook.forEach(callback => {
       if (!user && userId) user = this.users.findOne(userId, {fields: this._options.defaultFieldSelector});
       callback({ user, connection });
       return true;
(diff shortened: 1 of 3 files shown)


packages/accounts-base/accounts_server.js
```diff
@@ -878,16 +878,17 @@ export class AccountsServer extends AccountsCommon {
...
     query._id = userId;
     this.users.update(query, {
+      // Using $addToSet avoids getting an index error if another client
+      // logging in simultaneously has already inserted the new hashed
+      // token.
       $addToSet: {
         "services.resume.loginTokens": hashedToken
-      }
+      },
+      $set: { lastLoginAt: new Date() }
```
StorytellerCZ, line 891: I have decided to update the `lastLoginAt` here so that we don't have any new call to the DB.

KoenLav (approved):

packages/accounts-base/accounts_server.js
radekmie: I like this feature! I'm also in favor of reducing the number of database operations. I'm just not sure whether this is the correct place to do it. For example, we do use `_insertHashedLoginToken` to generate a login link that we send _somewhere_. By doing that, we no longer know when someone actually logged in, but when did we generate the token.

Maybe there's a better way of doing that? If not, it'd be important to highlight it in the docs.

StorytellerCZ: That is why this is a draft PR, so that we can discuss. In here this is (according to my understanding) triggered when the user logs in. It is not perfect, but I think this is a good start. I'll try to do a bit more testing to see if this triggers when the user is re-logged/resumed automatically by Meteor instead of doing the action themselves. If people could test this with their own cases to see if it works for them or there is a better place to put it that would be great.

packages/accounts-base/accounts_server.js
```diff
@@ -402,6 +402,8 @@ export class AccountsServer extends AccountsCommon {
     if (! stampedLoginToken) {
       stampedLoginToken = this._generateStampedLoginToken();
       this._insertLoginToken(userId, stampedLoginToken);
+    } else {
+      this.users.update(userId, { $set: { lastLoginAt: new Date() } });
```
StorytellerCZ, line 406: This one gets triggered when user resumes their connection (i.e. refreshes the window or re-opens the page after a while when the login token is still present).

packages/accounts-base/accounts_server.js
StorytellerCZ: Did some testing and this only triggers when user logs in via a login function, not when resuming. For the resumption I had to add a DB call above.

packages/accounts-base/accounts_server.js
StorytellerCZ: I'm thinking this one could be potentially hidden behind a settings flag so that it doesn't get set if the site doesn't want to consider resumption a login event.

filipenevola:
Hi, we've reviewed, and as a framework, this field doesn't help the core package in any way.

It also adds a new field and mutations that will trigger new diffs for all the existing apps publishing from the user collection without fields (projection) selection.

This feature is more suitable for a community package where the developer could install this package if they need to have login timestamps in the database.

So we are closing this PR.

PR #11777, opened 2021-11-21
installer: Add a flag to skip modifying exec path

ebroder:
The Meteor installer currently modifies the dotfiles of the user running it unconditionally. This makes sense as default behavior, but isn't always optimal. (For instance, I setup my dotfiles before installing meteor, and those dotfiles have the PATH modification already).

This change adds a flag to disable that behavior. I opted for an npm-managed argument, since that felt the most fluid of an interface to me. It's also accessible through a handful of means - you can use `npm install -g meteor --no-meteor-setup-exec-path`, but you can also pass it through the environment (`npm_config_meteor_setup_exec_path=false npm install -g meteor`). I used [this post](https://kgrz.io/how-npm-install-passes-cli-args.html) as a reference on how these arguments work, although the behavior of environment variables has changed a bit since that was written.

I didn't add any documentation yet. I assume that we'll want some, but I'm not sure where to put it, so pointers would be appreciated. (Easiest would probably be stuffing it in the installer README?)

I did run a handful of tests to make sure this worked the way I thought it did. First, here's a standard install:

```
evan@mathias meteor-installer (installer-optional-setup-path) % npm run-script -g install

> meteor@2.5.1 install
> node cli.js install
[...]
evan@mathias meteor-installer (installer-optional-setup-path) % tail -n 1 ~/.zshrc
export PATH=/Users/evan/.meteor:$PATH
```

Here's passing the argument through argv:

```
evan@mathias meteor-installer (installer-optional-setup-path) % npm run-script -g install --no-meteor-setup-exec-path

> meteor@2.5.1 install
> node cli.js install
[...]
evan@mathias meteor-installer (installer-optional-setup-path) % tail -n 1 ~/.zshrc
autoload -Uz compinit && compinit
```

And here's passing it through an environment variable (note that `--no-blah` in arguments gets canonicalized to `blah=''` in the env):

```
evan@mathias meteor-installer (installer-optional-setup-path) % npm_config_meteor_setup_exec_path=false npm run-script -g install

> meteor@2.5.1 install
> node cli.js install
[...]
evan@mathias meteor-installer (installer-optional-setup-path) % tail -n 1 ~/.zshrc
autoload -Uz compinit && compinit
```

### DIFF
--- a/npm-packages/meteor-installer/config.js
+++ b/npm-packages/meteor-installer/config.js
@@ -31,6 +31,10 @@ if (isWindows() && !localAppData) {
   throw new Error('LOCALAPPDATA env var is not set.');
 }
 
+const shouldSetupExecPath = () => {
+  return !process.env.npm_config_ignore_meteor_setup_exec_path;
+}
+
 const meteorLocalFolder = '.meteor';
 const meteorPath = path.resolve(rootPath, meteorLocalFolder);
 
@@ -45,5 +49,6 @@ module.exports = {
   isWindows,
   isMac,
   isRoot,
-  isSudo
+  isSudo,
+  shouldSetupExecPath,
 };
--- a/docs/source/install.md
+++ b/docs/source/install.md
@@ -38,6 +38,16 @@ If you only use sudo because of a distribution default permission system, [check
 
 In some cases you can get this error `npm WARN checkPermissions Missing write access to /usr/local/lib/node_modules` because your Node.js installation was performed with wrong permissions. An easy way to fix this is to install Node.js using [nvm](https://github.com/nvm-sh/nvm) and forcing it to be used in your terminal. You can force it in the current session of your terminal by running `nvm use 14`.
 
+<h2 id="path-management">PATH management</h2>
+
+By default, the Meteor installer adds its install path (by default, `~/.meteor/`) to your PATH by updating either your `.bashrc`, `.bash_profile`, or `.zshrc` as appropriate. To disable this behavior, install Meteor by running:
+
+```bash
+npm install -g meteor --ignore-meteor-setup-exec-path
+```
+
+(or by setting the environment variable `npm_config_ignore_meteor_setup_exec_path=true`)
+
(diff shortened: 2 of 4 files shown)


ebroder:
Hey folks - any chance of getting some eyes on this?

npm-packages/meteor-installer/config.js
```diff
@@ -31,6 +31,15 @@ if (isWindows() && !localAppData) {
   throw new Error('LOCALAPPDATA env var is not set.');
 }
 
+const shouldSetupExecPath = () => {
+  // If the variable isn't set, treat it as true, otherwise coerce it
+  if (!Object.prototype.hasOwnProperty.call(process.env, 'npm_config_meteor_setup_exec_path')) {
```
renanccastro, line 36: This flag is better as an opt-out of the default behavior, something like:

`--ignore-meteor-setup-exec-path`

So this code here could be simplified to 

`return process.env.npm_ignore_meteor_setup_exec_path`

renanccastro (changes requested):
There are 2 points where I think this should be documented:
- https://docs.meteor.com/install.html#installation
- and the readme as you said

I think this is a good addition, but as it is not the default behavior, a flag with a "ignore" word is better for it!

Sorry for the delay here.

ebroder:
Sure, that works for me. I've pushed an update that switches to that flag name and updates the documentation files you suggested.

zodern:
Would it be possible to make this automatic? It seems Meteor could detect when the PATH is already correct and then do nothing.

ebroder:
I'm not wild about that, personally. We could certainly inspect `$PATH` in the current process and check for Meteor, but that wouldn't actually tell us anything about what's in the user's dotfiles, and dotfiles themselves can be complicated enough that I wouldn't want to try and check them to see if the config is already there.

ebroder:
Hey folks - what can I do to move this forward?

renanccastro (approved):

renanccastro:
Sorry for the delay @ebroder
This PR will be released in 2.6.1, which should be early this month!

Thanks for your contribution.

PR #11795, opened 2021-12-03
Fix falling back to hot code push on web archs

zodern:
The changes in Meteor 2.5 to support cordova broke the web arch's falling back to hot code push when some changes to js files could not be applied with HMR.

### DIFF
--- a/packages/hot-module-replacement/client.js
+++ b/packages/hot-module-replacement/client.js
@@ -42,8 +42,20 @@ if (module._onRequire) {
   });
 }
 
-let pendingReload;
-let mustReload = false;
+// On web, we can reload the page any time to get the new version. On cordova,
+// we have to wait until Reload._onMigrate is called
+let hotCodePushReady = arch !== 'web.cordova';
+
+let useHotCodePush = false;
+let forceReload = function () {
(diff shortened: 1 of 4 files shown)


packages/hot-module-replacement/client.js
```diff
@@ -487,35 +491,24 @@ Meteor.startup(function () {
...
+
+      requestChanges();
     }
   });
 
-  // We disable hot code push for js until there were
-  // changes that can not be applied through HMR.
-  Package['reload'].Reload._onMigrate(function (tryReload) {
-    if (mustReload) {
+  Package['reload'].Reload._onMigrate(function () {
```
filipenevola, line 506: Is it safe to assume that every app has this package?

filipenevola (changes requested):

packages/hot-module-replacement/client.js
zodern: Yes. It is a dependency of `hot-code-push` which `hot-module-replacement` depends on. The code here is getting it from the Packages object because the Reload package loads after hot-module-replacement.

filipenevola (approved):

PR #11801, opened 2021-12-10
Prevent hot.accept from overriding hot.decline

zodern:
Currently, when both `module.hot.accept` and `module.hot.decline` are called for a module, the last one wins for if the module is updated with HMR.

There are several issues with this:
1. `module.hot.decline` is called when it is known the module can not be updated with HMR. Usually the code that calls`hot.accept` isn't completely accurate and it doesn't make sense for it to override `hot.decline`.
2. In most apps, `module.hot.accept` is only called by hmr integrations (`zodern:melte`, `react-fast-refresh`, `blaze-hot`, `vuejs:vue3`, etc.), after the module was run. If a developer simply adds `module.hot.decline()` to a module, it is likely to be overridden by the HMR integration, making it difficult to disable HMR for a module.
3. The HMR implementations in other bundlers that have a working `hot.decline` do not allow hot.accept to override hot.decline.

This is a breaking change. I'm not sure if the change is worth it, but `hot.decline` is rarely used so the effect should be minimal.

### DIFF
--- a/npm-packages/meteor-installer/README.md
+++ b/npm-packages/meteor-installer/README.md
@@ -23,6 +23,7 @@ npm install -g meteor
 | 2.4.1       | 2.4                     |
 | 2.5.0       | 2.5                     |
 | 2.5.1       | 2.5.1                   |
+| 2.5.2       | 2.5.1                   |
 
 ### Important note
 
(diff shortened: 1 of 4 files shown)


denihs (approved):
LGTM

PR #11818, opened 2021-12-23
accounts-2fa package

denihs:
Easy 2-Factor Integration For Meteor Apps

This package uses [node-2fa](https://www.npmjs.com/package/node-2fa) which works on top of [notp](https://github.com/guyht/notp), which implements TOTP ([RFC 6238](https://www.ietf.org/rfc/rfc6238.txt)) (the Authenticator standard), which is based on HOTP ([RFC 4226](https://www.ietf.org/rfc/rfc4226.txt)) to provide codes that are exactly compatible with all other Authenticator apps and services that use them.


packages/accounts-2fa/utils.js
```diff
@@ -0,0 +1,12 @@
+import twofactor from "node-2fa";
+import QRCode from "qrcode-svg";
+
+export const generateSecret = ({ username, appName } = {}) => {
+  const { secret, uri } = twofactor.generateSecret({ name: appName, account: username });
```
StorytellerCZ, line 5: If we assume that this package is only going to be used together with `accounts-password`, then this is fine, but it runs into trouble when OAuth packages are concerned as sometimes you might now have `username` available, would be better to be able to use user ID as well.

StorytellerCZ (changes requested):

packages/accounts-2fa/utils.js
denihs: Yes, there is a lot of changes we need to here. It's far from done. This was just a test. I'll work on this again next week 😉

guncebektas:
Your work is great and I really appreciate it

The username may include @ checking it with includes('@') is problematik.

It should accept userId

packages/accounts-2fa/2fa-client.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @summary Generates a svg QR code and save secret on user
+ * @locus Client
+ * @param {String} appName Optional. It's the name of your app that will show up when the user scans the QR code.
+ * @param {Function} [callback] Optional callback.
+ *   Called with no arguments on success, or with a single `Error` argument
+ *   on failure.
+ */
+Accounts.generateSvgCodeAndSaveSecret = (appName, callback) => {
```
filipenevola, line 35: `generateSvgCodeAndSaveSecret` this name is telling implementation details.

It doesn't matter that the image is an SVG. What is important here is that we generate a QR Code to enable 2FA.

I would call it: `generate2faActivationQrCode`

packages/accounts-2fa/2fa-client.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+  Accounts.connection.call(
+    'generateSvgCodeAndSaveSecret',
+    appName,
+    cb,
+  );
+};
+
+/**
+ * @summary Generates a svg QR code and save secret on user
```
filipenevola, line 48: Wrong summary

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
+import { Accounts } from 'meteor/accounts-base';
+import twofactor from "node-2fa";
+import QRCode from "qrcode-svg";
+import { Meteor } from "meteor/meteor";
+
+Accounts.checkUserHas2FAEnabled = selector => {
```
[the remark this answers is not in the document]
filipenevola: The name could be: `is2faEnabledForUser`.

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
...
+  if (!Meteor.isServer) {
+    throw new Meteor.Error(400, "The function checkUserHas2FAEnabled can only be called on the server");
+  }
+
+  if (typeof selector === 'string') {
+    if (!selector.includes('@')) {
```
filipenevola, line 12: I think this logic is not new, correct? Other packages have already decided this using this simple rule, am I right?

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
...
+      selector = {username: selector};
+    } else {
+      selector = {email: selector};
+    }
+  }
+
+  const user = Meteor.users.findOne(selector) || {};
```
filipenevola, line 19: I think we have some other ways to specify a custom selector in `accounts-password`. Should we support the same here? I'm not sure.

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
...
+
+    if (!user) {
+      throw new Meteor.Error(400, "No user logged in.");
+    }
+
+    const { twoFactorAuthentication, username } = user;
+
+
+    if (!twoFactorAuthentication || !twoFactorAuthentication.secret) {
+      throw new Meteor.Error(400, "The user does not have a secret generated. You may have to call the function generateSvgCode first.");
```
filipenevola, line 64: I think this error means an implementation error from the developer, so it should be a 500 error, no?

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+    if (!Accounts.isTokenValid(twoFactorAuthentication.secret, code)) {
+      throw new Meteor.Error(400, "Invalid token.");
+    }
+
+    Meteor.users.update({ username }, {
+      $set: {
+        twoFactorAuthentication: {
+          ...twoFactorAuthentication,
+          type: "otp",
```
filipenevola, line 74: Why is this necessary?

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
...
+  disableUser2fa() {
+    const user = Meteor.user();
+
+    if (!user) {
+      throw new Meteor.Error(400, "No user logged in.");
+    }
+
+    Meteor.users.update({ username: user.username }, {
+      $set: {
+        twoFactorAuthentication: {}
```
filipenevola, line 88: Why an empty object instead of removing it with $unset?

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,95 @@
...
+      }
+    });
+  },
+  has2FAEnabled(selector) {
```
filipenevola, line 92: Inconsistent name. `2FA`

packages/accounts-2fa/README.md
```diff
@@ -0,0 +1,172 @@
+#accounts-2fa
+
+---
+
+Easy 2-Factor Integration For Meteor Apps
```
filipenevola, line 5: Easy is not the best word here.

Simple would be better but in this case I would say:
`2-Factor Authentication for Meteor apps using accounts-password`

packages/accounts-2fa/README.md
```diff
@@ -0,0 +1,172 @@
...
+    if (error) {
+      console.error("Error trying to log in (user with 2fa)", error);
+    }
+  })}
+}>
+  Validate
+</button>
+```
+
+#API
```
filipenevola, line 137: We don't document packages on its README. It should be all in the docs.

packages/accounts-2fa/package.js
```diff
@@ -0,0 +1,27 @@
+Package.describe({
+  name: 'accounts-2fa',
+  version: '1.0.0',
+  summary: 'Package used to enable two factor authentication through OTP protocol',
+  git: '',
+  // By default, Meteor will default to using README.md for documentation.
+  // To avoid submitting documentation, set this field to null.
+  documentation: 'README.md'
```
filipenevola, line 8: This is the default.

packages/accounts-2fa/package.js
```diff
@@ -0,0 +1,27 @@
+Package.describe({
+  name: 'accounts-2fa',
+  version: '1.0.0',
+  summary: 'Package used to enable two factor authentication through OTP protocol',
+  git: '',
```
filipenevola, line 5: We could provide a directly URL to here or remove it.

packages/accounts-password/password_client.js
```diff
@@ -25,12 +25,15 @@ const reportError = (error, callback) => {
...
+ * @param {String} token
+ *   Optional if not using the package accounts-2fa. This will be the user's token
+ *   when they're trying to log in.
  * @param {Function} [callback] Optional callback.
  *   Called with no arguments on success, or with a single `Error` argument
  *   on failure.
  * @importFromPackage meteor
  */
-Meteor.loginWithPassword = (selector, password, callback) => {
+Meteor.loginWithPassword = (selector, password, token, callback) => {
```
filipenevola, line 36: This change is going to break all the codes that we have using Meteor.loginWithPassword, right?

This can't happen.

You need to check if the third arg is a function or not.

But I'm still not 100% sure this is the right approach.

Why didn't you create a new Meteor.loginWithPasswordAnd2fa and internally you call this loginWithPassword as well with different parameters but keeping the existent one untouched in terms of parameters?

It would have a new `internalLoginWithPassword` that could be even more flexible, receiving args as objects with more options and better long-term compatibility.

packages/accounts-password/password_server.js
```diff
@@ -169,18 +169,29 @@ Accounts.registerLoginHandler("password", options => {
...
   const user = Accounts._findUserByQuery(options.user, {fields: {
     services: 1,
+    twoFactorAuthentication: 1,
     ...Accounts._checkPasswordUserFields,
   }});
   if (!user) {
     Accounts._handleError("User not found");
   }
 
+  if(Accounts.checkUserHas2FAEnabled && Accounts.checkUserHas2FAEnabled(options.user)){
```
filipenevola, line 186: New lines should have Prettier applied. But pay attention, just new lines, not the entire file.

filipenevola: When this `Accounts.checkUserHas2FAEnabled` is going to be falsy?

filipenevola (changes requested):
I've noticed a few things that I don't understand or agree.

Especially the usage of accounts-passwordless and the changes on its API (that can't happen), at least not in this way.

packages/accounts-password/password_server.js
filipenevola, line 186: add a comment explaining that this method is defined in accounts-2fa package and if it doesn't exist is because this package is not used inside the current app.

packages/accounts-password/password_client.js
vooteles: I think coupling `accounts-password`and `accounts-2fa` together like this is not a good solution. It breaks all current apps using `accounts-password.` But it also ignores the flexibility that is already built into the accounts system.

Instead of modifying the existing password login handler, a better approach is to make use of Accounts.registerLoginHandler to register a new login handler on server side just for this package and then call that from client side with Accounts.callLoginMethod (or perhaps even just with Meteor.loginWithPassword). As I understand it, login handlers are executed one after another and can be stacked like this.

This approach is described in the following Meteor blog post: https://blog.meteor.com/tutorial-two-factor-authentication-with-meteor-and-totp-21d4a2f9ee51

packages/accounts-password/password_server.js
denihs: done

denihs: It's falsy when the package accounts-2fa is not present

packages/accounts-password/password_client.js
denihs: Logic changed as Filipe suggested in a 101 call ;)

packages/accounts-2fa/package.js
denihs: Solved

packages/accounts-2fa/2fa-client.js
denihs: done

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,27 @@
...
+
+{% apibox "Accounts.disableUser2fa" "module":"accounts-base" %}
+Use this function to give the users the option of disabling the 2FA.
+
+{% apibox "Accounts.has2FAEnabled" "module":"accounts-base" %}
+Use this function to verify if the user has 2FA enabled.
+
+<h3 id="log-in-with-code">Log in with code</h3>
+
+Once this package is added, the method `Meteor.loginWithPassword` starts to accept on additional parameter, which is the token: `Meteor.loginWithPassword(selector, password, token, [callback])`.
```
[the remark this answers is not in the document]
denihs: To make our decision public, the idea is for this package to be easily implemented with other login methods, such as the OAuth one, so we will stick with this name. Also, I will clarify in the doc that this package works together with the `accounts-password`.

packages/accounts-2fa/2fa-server.js
denihs: Yes, other methods use this strategy. It wasn't something I invented.

packages/accounts-2fa/2fa-server.js
denihs: It's what I use to verify if the user has 2FA enabled. They must have a secret and a type set up.

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+  Generate a new code
+</button>
+```
+
+
+At this point, the 2FA won't be activated just yet. Now that the user has access to the codes generated by their authenticator app, you can call the function `Accounts.enableUser2fa`:
+
+{% apibox "Accounts.enableUser2fa" "module":"accounts-base" %}
+
+Called with a code the user will receive from the authenticator app once they read the QR code. This function throws an error on failure. If the code provided is correct, a `type` will be added to the user's `twoFactorAuthentication` object, and now the 2FA will be considered enabled:
```
fredmaiaarantes, line 63: I didn't get this sentence.

> Called with a code the user will receive from the authenticator app once they read the QR code.

Maybe change it to something like:

```suggestion
It should be called with a code that the users will receive from the authenticator app once they read the QR code. This function throws an error on failure. If the code provided is correct, a `type` will be added to the user's `twoFactorAuthentication` object and now 2FA will be enabled:
```

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+  type: "otp",
+  secret: "***",
+}
+```
+
+<h3 id="log-in-with-2fa">Log in with 2FA</h3>
+
+Now that you have a way to allow your users to enable 2FA on their accounts, you can create a login flow based on that.
+
+To verify if a user has or not 2FA enabled you can call the function `Accounts.has2FAEnabled`:
```
fredmaiaarantes, line 76: Suggestion:
```suggestion
To verify whether or not a user has 2FA enabled, you can call the function `Accounts.has2FAEnabled`:
```

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+
+{% apibox "Accounts.has2faEnabled" "module":"accounts-base" %}
+
+With this function, you can verify if the has or not 2FA enabled, and based on this information, you can directly log the user in the 2FA is not enabled, or redirect the user to a place where they can provide a code, in case they do have 2FA enabled.
```
fredmaiaarantes, line 80: Suggestion:

```suggestion
With this function, you can check whether or not the user has 2FA enabled, and based on this information, you can directly log the user in the 2FA is not enabled, or redirect the user to a place where they can provide a code, in case they do have 2FA enabled.
```

I didn't get what you mean here:

> you can directly log the user in the 2FA is not enabled

Can you try to rephrase it?

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+    });
+  }
+}>
+  Login
+</button>
+```
+
+If the user has 2FA enabled, and you try to use the function `Meteor.loginWithPassword`, the login will fail, as the user should provide a token to access the app.
+
+The function you will need to call now to allow the user to log in is `Meteor.loginWithPasswordAnd2faToken`:
```
fredmaiaarantes, line 114:
```suggestion
The function you will need to call now to allow the user to login is `Meteor.loginWithPasswordAnd2faToken`:
```

fredmaiaarantes (changes requested):

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+
+> This package is meant to be used with `accounts-password`, so if you still didn't add `account-password` to your project, you'll need to add it too. In the future we want to enable the use of this app with other login methods, like `accounts-passwordless` or our oauth methods (Google, GitHub, etc...).
+
+<h3 id="activating-2fa">Activating 2FA</h3>
+
+The first step to use 2FA is to generate a QR code so the user can read it on an authenticator app and start to receiving codes.
+
+{% apibox "Accounts.generate2faActivationQrCode" "module":"accounts-base" %}
+
+Receives an `appName` which is the name of your app that will show up when the user scans the QR code. Also, a callback called with a QR code in SVG format on success or a single `Error` argument
```
filipenevola, line 16: is appName optional as the apibox says?

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+on failure. Both parameters are optional.
```
filipenevola, line 17: how they could be optional? what is going to happen if I can do this without any param? Is this case useful?

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+
+On success, this function will also add an object to the logged user containing the QR secret:
```
filipenevola, line 19: The whole flow is not explained at the top, before starting to talk about the API.

For example, the user needs to be logged in to active 2FA. I miss the full picture before I start reading the API.

docs/source/packages/accounts-2fa.md
filipenevola, line 63: throws or callback with an error?

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+
+{% apibox "Meteor.loginWithPasswordAnd2faToken" %}
+
+Now you will be able to receive a code from the user and this function will verify if the code is valid. If it is, the user will be logged in.
+
+So the call of this function should look something like this:
+
+```js
+<button onClick={() => {
+  Meteor.loginWithPasswordAnd2faToken(username, password, code,error => {
```
filipenevola, line 124: we have a mix in the codebase and also in the docs calling sometimes `code` and sometimes `token` the 2FA code.

We should use the same name everywhere. It seems that `code` is more common than token in the code and docs.

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,140 @@
...
+    }
+  })}
+}>
+  Validate and log in
+</button>
+```
+
+<h3 id="disabling-2fa">Disabling 2FA</h3>
+
+To disable 2FA you can simply use this function:
```
filipenevola, line 136:
```suggestion
To disable 2FA for an user use this method: 
```

packages/accounts-2fa/2fa-client.js
```diff
@@ -0,0 +1,74 @@
...
+ *    or not 2FA enabled, or with a single `Error` argument on failure.
+ */
+Accounts.has2faEnabled = (selector, callback) => {
+  Accounts.connection.call('has2faEnabled', selector, callback);
+};
+
+/**
+ * @summary Generates a svg QR code and save secret on user
+ * @locus Client
+ * @param {String} appName Optional. It's the name of your app that will show up when the user scans the QR code.
```
filipenevola, line 26: if it is really optional usually optional arguments are defined as `[argName]` instead of `argName` if I recall correctly from other apiboxes.

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    twoFactorAuthentication &&
+    twoFactorAuthentication.secret &&
+    twoFactorAuthentication.type === 'otp'
+  );
+};
+
+Accounts.generate2faToken = secret => twofactor.generateToken(secret);
+
+Accounts.isTokenValid = (secret, token) => {
```
filipenevola, line 33: Why is this method exposed in the Accounts object?

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,123 @@
...
+  const user = Meteor.users.findOne(selector) || {};
+  const { twoFactorAuthentication } = user;
+  return (
+    twoFactorAuthentication &&
+    twoFactorAuthentication.secret &&
+    twoFactorAuthentication.type === 'otp'
+  );
+};
+
+Accounts.generate2faToken = secret => twofactor.generateToken(secret);
```
filipenevola, line 31: Why is this method exposed in the Accounts object?

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,123 @@
+import { Accounts } from 'meteor/accounts-base';
+import twofactor from 'node-2fa';
+import QRCode from 'qrcode-svg';
+import { Meteor } from 'meteor/meteor';
+
+Accounts.is2faEnabledForUser = selector => {
```
filipenevola, line 6: Why is this method exposed in the Accounts object?

packages/accounts-2fa/package.js
```diff
@@ -0,0 +1,22 @@
...
+    'Package used to enable two factor authentication through OTP protocol',
+});
+
+Npm.depends({
+  'node-2fa': '2.0.3',
+  'qrcode-svg': '1.1.0',
+});
+
+Package.onUse(function(api) {
+  api.use(['accounts-base'], ['client', 'server']);
```
filipenevola, line 13: I believe accounts-password should be here as well, AFAIK it's required as a dependency for this to work.

Is that not the case? If not, why?

packages/accounts-password/password_client.js
```diff
@@ -7,6 +7,31 @@ const reportError = (error, callback) => {
    }
 };
 
+
+const internalLoginWithPassword = ({ selector, password, token, callback }) => {
```
filipenevola, line 11: Maybe this should be exposed in the accounts object as `_internalLoginWithPassword` so `Meteor.loginWithPasswordAnd2faToken` could be present in the correct package instead of here.

Do you agree with that @denihs ?

filipenevola (changes requested):

docs/source/packages/accounts-2fa.md
denihs: Yes

docs/source/packages/accounts-2fa.md
denihs: This description is wrong. Just appName is optional. Fixed :palms_up_together: :

docs/source/packages/accounts-2fa.md
denihs: callback

packages/accounts-2fa/2fa-server.js
denihs: Because I also need to called inside the accounts-password to verify if the code provided is valid https://github.com/meteor/meteor/blob/426872ebb4fe00c14914a66f5f4705d3d88823be/packages/accounts-password/password_server.js#L195

packages/accounts-2fa/2fa-server.js
denihs: This one we can change if there is a better alternative, but I did this way so I could valid on flows in the tests https://github.com/meteor/meteor/blob/64701fe9465da2a3906a9179be175eb7c7b7ecc8/packages/accounts-base/accounts_tests_setup.js#L1-L15

packages/accounts-2fa/2fa-server.js
denihs: Because I also need to call it inside the accounts-password 
https://github.com/meteor/meteor/blob/426872ebb4fe00c14914a66f5f4705d3d88823be/packages/accounts-password/password_server.js#L187-L190

packages/accounts-2fa/package.js
denihs: This package doesn't use any dependency from accounts-password... It is true that the developer needs to have both packages, but as they don't share direct dependencies, and they "communicate" through the Accounts object, I don't think it's necessary.

packages/accounts-password/password_client.js
denihs: My problem with that is when we decide to integrate 2FA with another login method, like passwordless, we'll need to change both packages (2fa and passwordless) to keep the same logic, whereas if we keep it as it is right now, we'll need to change just the passwordless.

This would be my only argument. But we can do as you suggested, you prefer.

packages/accounts-2fa/2fa-client.js
denihs: I couldn't find an example of this... I saw the use of `[something]` for callbacks, so I assumed it's this way because callbacks are functions. Should I still change it?

filipenevola: [x] is for optional args, if here it is not optional, it shouldn't have it.

packages/accounts-2fa/2fa-server.js
filipenevola: If it is only available for tests we should name it starting with `_` so we can change it later without notifying everybody or considering a breaking change.

packages/accounts-2fa/2fa-server.js
filipenevola: If this is only for internal usage we should name it starting with `_` so we can change it later without notifying everybody or considering a breaking change.

packages/accounts-2fa/2fa-server.js
filipenevola: Same as above, we should start with `_`

packages/accounts-2fa/package.js
filipenevola: Adding this package will add accounts-password automatically if you put it on api.use, that is why this is necessary.

This package doesn't work without accounts-password.

packages/accounts-password/password_client.js
filipenevola: I don't see how.

If we add new packages they can also use _internalLoginWithPassword as 2fa is going to use now.

I believe we should change that.

docs/source/packages/accounts-2fa.md
```diff
@@ -84,7 +86,7 @@ A way of using it would be:
 ```js
 <button 
   onClick={() => {
-    Accounts.has2FAEnabled(username, (err, isEnabled) => {
+    Accounts.has2fanabled(username, (err, isEnabled) => {
```
renanccastro, line 89: has2faEnabled

renanccastro (changes requested):

packages/accounts-2fa/2fa-server.js
renanccastro: Isn't this necessary if I want to show a screen saying that this user has 2fa enabled or not?

packages/accounts-2fa/2fa-client.js
denihs: Gotcha

packages/accounts-2fa/2fa-server.js
denihs: This is the server part @renanccastro. The client method it's exposed [already](https://github.com/meteor/meteor/blob/65fb19a105d769c836bcd9bbb40b67006f2798b1/packages/accounts-2fa/2fa-client.js#L12-L21).

denihs: Done

packages/accounts-2fa/2fa-server.js
denihs: Done

packages/accounts-2fa/2fa-client.js
```diff
@@ -0,0 +1,99 @@
...
+ *    or not 2FA enabled, or with a single `Error` argument on failure.
+ */
+Accounts.has2faEnabled = (selector, callback) => {
+  Accounts.connection.call('has2faEnabled', selector, callback);
+};
+
+/**
+ * @summary Generates a svg QR code and save secret on user
+ * @locus Client
+ * @param {String} [appName] Optional. It's the name of your app that will show up when the user scans the QR code.
```
renanccastro, line 26: This shouldn't be optional, right?

Also, for spaces, it's probably necessary to escape the spaces with %20

docs/source/packages/accounts-2fa.md
```diff
@@ -0,0 +1,241 @@
...
+        if (error) {console.error('...', error);}
+      }
+    );
+  }}
+>
+  Validate and Log in
+</button>;
+```
+
+<h3 id="disabling-2fa">Disabling 2FA</h3>
```
filipenevola, line 204: This should be before the examples with passwordless and password.

Also, these examples should be here or in their own packages?

At least their doc should mention this doc.

packages/accounts-2fa/README.md
```diff
@@ -0,0 +1,9 @@
+# accounts-2fa
+
+[Source code of released version](https://github.com/meteor/meteor/tree/master/packages/accounts-2fa)
+| [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/accounts-2fa)
+***
+
+A package to implement 2FA in login flows that use `Meteor.loginWithPassword`. See
```
filipenevola, line 7: Is this text correct yet?

filipenevola (changes requested):

packages/accounts-2fa/README.md
denihs: I forgot to review the readme. I'll rephrase this.

docs/source/packages/accounts-passwordless.md
```diff
@@ -25,3 +25,7 @@ sendLoginToken: {
   text: (user, url, { sequence }) => { /* text template */ }
 }
 ```
+
+<h3 id="enabling-2fa">Enable 2FA for this package</h3>
+
+You can add 2FA to your login flow by using the package [accounts-2fa](https://docs.meteor.com/packages/accounts-2fa.html). You can find an example showing how this would look like [here](https://docs.meteor.com/packages/accounts-2fa.html#working-with-accounts-password).
```
filipenevola, line 31: wrong anchor.

packages/accounts-2fa/README.md
```diff
@@ -4,6 +4,6 @@
 | [Source code of development version](https://github.com/meteor/meteor/tree/devel/packages/accounts-2fa)
 ***
 
-A package to implement 2FA in login flows that use `Meteor.loginWithPassword`. See
+A package that provides a simple way to integrate 2FA in login services. Today, the packages [accounts-password](https://docs.meteor.com/api/passwords.html) and [accounts-passwordless](https://docs.meteor.com/packages/accounts-passwordless.html) are integrated with this package. See
```
filipenevola, line 7: You should say: "Check the docs to see which packages are compatible with 2FA already". This way this document is not going to be wrong soon.

filipenevola (changes requested):

filipenevola (approved):
LTGM, you can create the release-2.6.1 branch from release-2.6 and merge this branch into it.

So we can start 2.6.1 beta cycle as soon as 2.6 is done.

PR #11825, opened 2022-01-01
Fix #11817 add try/catch to end of redirect response

### LINKED ISSUE #11817: oauth no longer works if app embedded in iframe, using the latest security settings on chrome
How to reproduce:
Step 1: Turn on the setting on google chrome "block third-party cookies" which i believe is enabled by default (at least in safari it is).
Step 2: Load the meteor app inside an iframe on a different domain

Because oAuth relies on sessionStorage, login flow will no longer work.

StorytellerCZ:
Fixes #11817 by adding try/catch around sessionStorage in end of redirect response.

### DIFF
--- a/packages/oauth/end_of_redirect_response.js
+++ b/packages/oauth/end_of_redirect_response.js
@@ -4,8 +4,13 @@
   var config = JSON.parse(document.getElementById("config").innerHTML);
 
   if (config.setCredentialToken) {
-    sessionStorage[config.storagePrefix + config.credentialToken] =
-      config.credentialSecret;
+    try {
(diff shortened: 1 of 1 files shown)


bogdanrn:
@StorytellerCZ thanks for the quick reply, that looks good. This would fix the problem i encountered. 

PS:
For anyone who lands on this issue, the next step is to implement login tokens to resume the session. I did it by using `Accounts.registerLoginHandler`. Meteor will continue working as expected from this point on.

filipenevola (approved):

filipenevola:
Published oauth@2.1.1.

PR #11831, opened 2022-01-04
Support MongoDB 5

filipenevola:
We are implementing the changes required to make Meteor mongo package work with MongoDB 5.

Related [discussion](https://github.com/meteor/meteor/discussions/11749).

## Latest News

:tada: :tada:  **Meteor 2.6-beta.4** is now released with mongo v5 support!! :tada: :tada: 

- Meteor 2.6-beta.3 fixes an issue on oplog tailing that caused high CPU usage. refer here for more details: #11849 

Please note this is an experimental release and should be used with caution.

### Migration Steps

You shouldn't have to do anything aside using the latest beta.

## What has been done

The mongo-driver and oplog implementation has been changed to support MongoDB 5 which introduces a lot of breaking changes, mostly non-documented.
The major change here is that the oplog format has changed dramatically, and, as we don't have any documentation available "reverse engineering-ish" has been applied to discover the format and convert it to the older $v=1 format. More information about it can be found in the oplog converter js file. 
We are already updating to **v4.3.0** of the MongoDB nodejs driver, which was released hours ago.

## Pending list

- [x] Document breaking changes
- [x] Pass on all tests
- [x] Test on MongoDB 4 to ensure compatibility
- [x] Further testing with real applications

## Important changes

  - useUnifiedTopology is not an option anymore, it defaults to true.
  - native parser is not an option anymore, it defaults to false in the mongo connection.
  - poolSize not an option anymore, we are using max/minPoolSize for the same behavior on mongo connection.
  - fields option is deprecated, we are maintaining a translation layer to "projection" field (now prefered) until the next minor version, where we will start showing alerts.
  - _ensureIndex is now showing a deprecation message
  - applySkipLimit option for count() on find cursors is no longer supported.
  - we are maintaining a translation layer for the new oplog format, so if you read or rely on any behavior of it please read our oplog_v2_converter.js code
  - update/insert/remove behavior is maintained in the Meteor way, documented in our docs, but we are now using replaceOne/updateOne/updateMany internally. This is subject to changes in the API rewrite of MongoDB without Fibers AND if you are using rawCollection directly you have to review your methods otherwise you will see deprecation messages if you are still using the old mongodb style directly.
  - internal result of operations inside nodejs mongodb driver have changed. If you are depending on rawCollection results(not only the effect inside the DB), please review the expected format as we have done [here](https://github.com/meteor/meteor/blob/155ae639ee590bae66237fc1c29295072ec92aef/packages/mongo/mongo_driver.js#L658)
  - waitForStepDownOnNonCommandShutdown=false is not needed anymore when spawning the mongodb process
  - _synchronousCursor._dbCursor.operation is not an option anymore in the raw cursor from nodejs mongodb driver. If you want to access the options, use _synchronousCursor._dbCursor.(GETTERS) - for example, _synchronousCursor._dbCursor.readPreference.

## Possible affected points and worth testing items:
- MongoDB 4 and older should work, the same with its oplog
- MongoDB 5 should work with oplog support, which means we have to test if the changes are real-time reflected in a client-side app after being changed by another client, for example.
- Mongo embedded has to work on mac x64/m1 linux and windows.
- Benchmark speed to make sure we haven't introduced regressions
- Test complex queries to be sure our converter from OPLOG v2 -> v1 works on all cases.


packages/mongo/oplog_observe_driver.js
```diff
@@ -598,11 +602,15 @@ _.extend(OplogObserveDriver.prototype, {
...
+        // we are mapping the new oplog format on mongo 5
+        // to what we know better, $set
+        op.o = oplogV2V1Converter(op.o)
         // Is this a modifier ($set/$unset, which may require us to poll the
         // database to figure out if the whole document matches the selector) or
         // a replacement (in which case we can just directly re-evaluate the
         // selector)?
-        var isReplace = !_.has(op.o, '$set') && !_.has(op.o, '$unset');
+        // oplog format has changed on mongodb 5, we have to support both now
+        var isReplace = !_.has(op.o, '$set') && !_.has(op.o, 'diff') && !_.has(op.o, '$unset');
```
filipenevola, line 613: It would be nice to add to the comment `diff is the format in Mongo 5+`

packages/mongo/oplog_tailing.js
```diff
@@ -206,12 +206,12 @@ Object.assign(OplogHandle.prototype, {
     // The tail connection will only ever be running a single tail command, so
     // it only needs to make one underlying TCP connection.
     self._oplogTailConnection = new MongoConnection(
-      self._oplogUrl, {poolSize: 1});
+      self._oplogUrl, {maxPoolSize: 1});
```
filipenevola, line 209: as I said in the other comment it's good to document this change.

packages/mongo/oplog_v2_converter.js
```diff
@@ -0,0 +1,113 @@
...
+  };
+};
+
+export const oplogV2V1Converter = v2OplogEntry => {
+  if (v2OplogEntry.$v !== 2 || !v2OplogEntry.diff) return v2OplogEntry;
+  return { $v: 2, ...nestedOplogEntryParsers(v2OplogEntry.diff || {}) };
+};
+
+function flattenObject(ob) {
+  var toReturn = {};
```
filipenevola, line 96: Let's avoid var in the new code.

packages/mongo/oplog_v2_converter.js
```diff
@@ -0,0 +1,113 @@
...
+
+i and u are both $set
+d is $unset
+on mongo 4
+ */
+
+const isArrayOperator = possibleArrayOperator => {
+  if (!Object.keys(possibleArrayOperator).length) return false;
+
+  let isArrayOperator = true;
```
filipenevola, line 27: you don't need this boolean

packages/mongo/oplog_v2_converter.js
```diff
@@ -0,0 +1,113 @@
...
+  Object.keys(possibleArrayOperator).forEach(key => {
```
filipenevola, line 28: replace this forEach and boolean by a direct return like:

```js
return !Object.keys(possibleArrayOperator).find(key => key !== 'a' && !key.match(/u\d+/));
```

packages/mongo/oplog_v2_converter.js
```diff
@@ -0,0 +1,113 @@
...
+const nestedOplogEntryParsers = (
+  { i = {}, u = {}, d = {}, ...sFields },
+  prefixKey = ''
+) => {
+  const sFieldsOperators = [];
+  Object.entries(sFields).forEach(([key, value]) => {
+    if (isArrayOperator(value || {})) {
+      const { a, ...uPosition } = value;
+      const [positionKey, newArrayIndexValue] = Object.entries(uPosition)[0];
+      if (uPosition)
```
filipenevola, line 44: it's missing {

And when `uPosition` will not be truth value?

packages/mongo/oplog_v2_converter_tests.js
```diff
@@ -0,0 +1,89 @@
...
+    $v: 2,
+    diff: { sa: { d: { b: false } } },
+  };
+  const entry8 = { $v: 2, diff: { u: { c: 'bar' }, sb: { a: true, u0: 2 } } };
+
+  const entry9 = {"$v":2,"diff":{"sservices":{"sresume":{"u":{"loginTokens":[]}}}}};
+
+  const entry10 =  {"$v":2,"$set":{"services.resume.loginTokens":[{"when":"2022-01-06T23:58:35.704Z","hashedToken":"RlalW6ZSvPPJLH6sW3B1b+vrUnPy+Ox5oMv3O3S7jwg="},{"when":"2022-01-06T23:58:35.704Z","hashedToken":"DWG0Qw/+nZ48wAIhKR2r9H41wLpth9BM+Br6aZsl2bU="}]}};
+
+  test.equal(
```
filipenevola, line 40: Why are you not asserting right after the declaration? It's harder to read the test as the declaration is far from the assertation.

packages/tinytest/tinytest.js
```diff
@@ -203,7 +203,7 @@ export class TestCaseResults {
   // The upshot is, if you want to test whether an error is of a
   // particular class, use a predicate function.
   //
-  throws(f, expected) {
+  throws(f, expected, message) {
```
filipenevola, line 206: I think we have TS typedef for this, we need to update DeclaredTypes or at least open an issue for this. Also maybe `only` and `onlyAsync` are also missing there.

tools/runners/run-mongo.js
```diff
@@ -71,19 +71,7 @@ function spawnMongod(mongodPath, port, dbPath, replSetName) {
...
-    // Starting with version 4.0.8/4.1.10, MongoDB performs a step down
-    // procedure if the primary receives a SIGTERM signal
-    // (https://jira.mongodb.org/browse/SERVER-38994). During this procedure,
-    // the process doesn't shut down for up to ten seconds until a secondary
-    // becomes the new primary. Since Meteor starts a single-node replica set,
-    // this is unnecessary because there are no secondaries. The following
-    // parameter disables the step down. This will be the default for single-
-    // node replica sets in MongoDB 4.3 (relevant commit: https://git.io/JeNkT),
-    // so the parameter can be removed in the future.
-    '--setParameter',
```
filipenevola, line 85: It's good to document this removal.

packages/mongo/oplog_v2_converter.js
```diff
@@ -0,0 +1,113 @@
+// we are mapping the new oplog format on mongo 5
```
filipenevola, line 1: It would be nice to post this code [here](https://stackoverflow.com/questions/68936772/mongo-oplog-format-in-5-0-2-contains-s-prefix-on-updates).

And also share with the authors of RedHat Debezium

packages/minimongo/cursor.js
```diff
@@ -25,7 +25,7 @@ export default class Cursor {
 
     this.skip = options.skip || 0;
     this.limit = options.limit;
-    this.fields = options.fields;
+    this.fields = options.projection || options.fields;
```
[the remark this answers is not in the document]
renanccastro: It would require a good deal of changes for a name replacement. For now, I don't think it justifies the cost/benefit as we are still supporting the "fields" field/

packages/mongo/collection.js
```diff
@@ -726,9 +739,8 @@ Object.assign(Mongo.Collection.prototype, {
     var self = this;
     if (!self._collection._ensureIndex || !self._collection.createIndex)
       throw new Error('Can only call createIndex on server collections');
-    // TODO enable this message a release before we will remove this function
-    // import { Log } from 'meteor/logging';
-    // Log.debug(`_ensureIndex has been deprecated, please use the new 'createIndex' instead${options?.name ? `, index name: ${options.name}` : `, index: ${JSON.stringify(index)}`}`)
+    import { Log } from 'meteor/logging';
```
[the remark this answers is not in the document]
renanccastro: There is no good reason, but it's the only place it's used. As it's also a client-side code there can be benefits in load time.

packages/mongo/mongo_driver.js
```diff
@@ -146,42 +148,16 @@ MongoConnection = function (url, options) {
...
 
-  // Disable the native parser by default, unless specifically enabled
-  // in the mongo URL.
-  // - The native driver can cause errors which normally would be
-  //   thrown, caught, and handled into segfaults that take down the
-  //   whole app.
-  // - Binary modules don't yet work when you bundle and move the bundle
-  //   to a different platform (aka deploy)
-  // We should revisit this after binary npm module support lands.
-  if (!(/[\?&]native_?[pP]arser=/.test(url))) {
```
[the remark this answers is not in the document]
renanccastro: Yes, but it was disabled in the past and now the behavior is disabled by default :) I will document it, but there is no real change

packages/mongo/oplog_v2_converter.js
renanccastro: never. I have added a throw in this case.

minhna:
I tried Meteor 2.6-beta.2, I can see it use much more CPU resource in both node and mongodb processes. It could be 5 - 6 times more at idle state.

renanccastro:
> I tried Meteor 2.6-beta.2, I can see it use much more CPU resource in both node and mongodb processes. It could be 5 - 6 times more at idle state.

Hi @minhna. Can you please detail a little more? Attaching the numbers from your htop process(both mongo server and meteor tool process), comparing the version without the beta, and with the beta would be an amazing starting point for us to try to understand what you are seeing. In a idle state, meteor is not doing any work related to the new package, but the embedded mongo could be indeed using more CPU.

Another good tool for profiling is to take a CPU profile using Node.js inspect.
You can run your project with `TOOL_NODE_FLAGS=--inspect-brk meteor`, and open the node inspector on chrome(from your dev panel, or inside chrome://inspect). Inside the debugger, you will see the tab to take a CPU profile from it. Taking it from the two versions would be super helpful.

minhna:
I send the htop screen first:
Meteor 2.5.3
[https://drive.google.com/file/d/1CTlHDWEjn7akBGzcOAR6TRTk61cnXdJL/view?usp=sharing](url)
[https://drive.google.com/file/d/1jnRqQ0zwTdkjTSJcfElKhMLOwmm2q96U/view?usp=sharing](url)
Meteor 2.6 beta 2
[https://drive.google.com/file/d/1n9lP8HLxhoIs5cty9ar49Sol6j2Us0AL/view?usp=sharing](url)
[https://drive.google.com/file/d/1lqSbhYOeDm4kSl_M4NsftarGUmYdNtyh/view?usp=sharing](url)

The same code, just change the meteor version.

minhna:
I run with command: `meteor run --port=9001 --settings=settings.development.json --inspect-brk`
and these are cpu profiles:
[Startup](https://drive.google.com/file/d/1YkLLGFgokXXgaTSX6Kza7XWubGKDBYl2/view?usp=sharing)
[Running idle](https://drive.google.com/file/d/1SVKPkRAeL9o6cbh-YnWWpcqGKnt9qmdv/view?usp=sharing)

Floriferous:
Just received a notification from MongoDB yesterday that our atlas instance will be forced to upgrade to v5 in February... so this is right on time!

a4xrbj1:
Well, I have the received the same message and I have a M15 on Atlas. So no idea why they are sending it to me as well.

renanccastro:
@minhna can you please give it a try on 2.6-beta.3? I've found this issue #11849 related to oplog tailing. Thanks very much for your detailed report!

minhna:
@renanccastro it's much better now. It uses a little more CPU than 2.5.3 but much better than 2.6-beta.2.

filipenevola:
> Just noticed a couple of warnings when launching the mongo console via `meteor mongo`

Thank you @flavio-b , we are going to check these soon.

filipenevola:
We are going to keep eventual new things related to MongoDB in new PRs or in the 2.6 [PR](https://github.com/meteor/meteor/pull/11867).

Issue #11830, opened 2022-01-04
Can't run app after updating to 2.5.2 (SyntaxError: Unexpected end of JSON input)


markdowney:
- was running meteor 2.5.1. on macOS Bigsur.
- updated app to 2.5.2
- running the app throws the following error :

```
/Users/me/.meteor/packages/meteor-tool/.2.5.1.19xclpn.3tu3++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/meteor-promise/promise_server.js:218
      throw error;
      ^

SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /Users/me/.meteor/packages/meteor-tool/.2.5.1.19xclpn.3tu3++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/tools/fs/tools/fs/optimistic.ts:321:17
    at wrap.makeCacheKey (/Users/me/.meteor/packages/meteor-tool/.2.5.1.19xclpn.3tu3++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/tools/fs/tools/fs/optimistic.ts:36:15)
    at recomputeNewValue (/Users/me/.meteor/packages/meteor-tool/.2.5.1.19xclpn.3tu3++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/optimism/src/entry.ts:198:31)
    at Slot.withValue (/Users/me/.meteor/packages/meteor-tool/.2.5.1.19xclpn.3tu3++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/@wry/context/lib/context.esm.js:69:29)
    at reallyRecompute (/Users/me/.meteor/packages/meteor-tool/.2.5.1.19xclpn.3tu3++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/dev_bundle/lib/node_modules/optimism/src/entry.ts:181:19)
[...]

Issue #11791, opened 2021-11-30
Fibers is missing binaries; can do away with fork


GeoffreyBooth:
@renanccastro I just wanted to draw your attention to https://github.com/laverdet/node-fibers/pull/454#issuecomment-981142332: you can take over the original Fibers project. I think this would be preferable to maintaining a separate fork.

In addition, the version of the Fibers fork for Meteor 2.5.1 was missing some binaries that the original version usually included when it was published to npm. In particular, the only ones in the fork are `darwin-x64-93` and `linux-x64-72-glibc`. The original library publishes all of these: `darwin-x64-72`, `darwin-x64-83`, `darwin-x64-93`, `linux-x64-72-glibc`, `linux-x64-72-musl`, `linux-x64-83-glibc`, `linux-x64-83-musl`, `win32-x64-72`, `win32-x64-83`. (You can see this by running `npm install fibers` in a new project and looking in `node_modules/fibers/bin`.) [My project is broken without the musl binary that is needed to run in Alpine.](https://github.com/disney/meteor-base/pull/95#issuecomment-974765661)

So if you don’t mind, I suggest you accept the offer in https://github.com/laverdet/node-fibers/pull/454 to take over the original Fibers repo and merge in the PRs that necessitated the fork, and publish a new version to NPM that includes the full set of binaries. Instructions on how to generate them are [here](https://github.com/laverdet/node-fibers/pull/454#issuecomment-981142332). Thanks!

renanccastro:
Hi @GeoffreyBooth , I'm following up with laverdet for having npm push permissions there. As soon as we have it, we will publish a new version with all the binaries.

Sorry for this, adding the other architectures went through me, as fibers can always be compiled this wouldn't be a fatal exception in our flow (except on alpine). I will correct this in the new fibers version.

simplecommerce:
@renanccastro Does the [2.5.2 beta 10](https://github.com/meteor/meteor/pull/11788#issuecomment-998303564) release include this new fibers version with all binairies? I am asking because in my current project, I cannot update to latest version of Meteor on Azure (windows web app) as I can't install any fibers version above 4.0.2 because of missing binairies. I don't mind waiting, I was just curious, since I am not sure how to check.

Thanks!

renanccastro:
@simplecommerce no, not yet, unfortunately. There is still some CI work to be done for this to be possible

nspangler:
[...]

PR #11832, opened 2022-01-05
Fixes invalid package.json files breaking Meteor run

filipenevola:
Provides a better error message when we have invalid package.json files.

Also provides new env vars to change the default behavior.

As explained [here](https://forums.meteor.com/t/unable-to-run-after-update-to-2-5-2/57266/6?u=filipenevola) `resolve` is including invalid package.json file intentionally. So we add a specific code to ignore the error when parsing it.

In order to improve the diagnostic of errors like this in the future, we have added a new `console.error` to make clear from which directory this error is coming.

We have also added three new different env vars:
- `METEOR_WARN_ON_INVALID_EXPECTED_PACKAGE_JSON_ERRORS`: to produce warns even for expected invalid package.json;
- `METEOR_IGNORE_INVALID_PACKAGE_JSON_ERRORS`: to ignore any package.json error, could result in problems when running the app;
- `METEOR_WARN_ON_INVALID_PACKAGE_JSON_ERRORS`: produce warns but allow the app to run anyway, could result in problems when running the app;

You shouldn't use these env vars in the long term but it could be helpful to keep your app running while the dependency is not fixed.

If your problem is happening because of `resolve` malformed package.json you don't need to use any of these env vars above.

Fixes https://github.com/meteor/meteor/issues/11830

### DIFF
--- a/tools/isobuild/package-source.js
+++ b/tools/isobuild/package-source.js
@@ -1372,7 +1372,7 @@ Object.assign(PackageSource.prototype, {
           }
           // This is going to break the run but at least with a clear error indicating what is the problematic package.json
           console.error(message, e);
-          throw new Error(message);
+          throw e;
         }
 
         // If a package.json file with a "name" property is found, it will
--- a/tools/isobuild/package-source.js
+++ b/tools/isobuild/package-source.js
@@ -37,6 +37,10 @@ import {
   optimisticLookupPackageJsonArray,
 } from "../fs/optimistic";
 
+// resolve package includes malformed package.json intentionally
+// https://forums.meteor.com/t/unable-to-run-after-update-to-2-5-2/57266/6
+const EXPECTED_INVALID_PACKAGE_JSON_PATHS_TO_IGNORE = ['resolve/test/resolver/malformed_package_json']
+
 // XXX: This is a medium-term hack, to avoid having the user set a package name
 // & test-name in package.describe. We will change this in the new control file
 // world in some way.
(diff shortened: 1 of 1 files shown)


renanccastro (approved):

simplecommerce:
@filipenevola Hi, I am currently having the same issue but on 1.9.2 for all my projects, I unfortunately can't update to latest release of Meteor because of https://github.com/meteor/meteor/issues/11791, since I am using Azure Web App on windows, some binaries are missing so I can't update. Any change this can be fixed without updating the release? I am trying on my end to see how I can update it to 2.5.3.

filipenevola:
Hi @simplecommerce we could probably create a patch on top of 1.9.3 with this fix and publish 1.9.4.

Would this help you?

simplecommerce:
@filipenevola thanks for the follow-up, I think I can `meteor npm install resolve@1.20.0` at the moment on my projects, until I can update to the latest version, unless your patch would be a safer alternative until then. If no one else has this issue but me for now I can live with this. Thanks again!

filipenevola:
Hi @simplecommerce, this approach is safe IMO.

And in any case, we could create the patch later if you found a problem with this workaround.

Meteor has many improvements in the latest versions so I really encourage you to upgrade, if you have any issues that you are facing in the upgrade path, please open a new topic in the Forums so we can help you.

ljharb:
While this solves the immediate issue, this is not a general solution. Meteor shouldn’t even be attempting to read package.json files that aren’t at the top level of a package.

PR #11420, opened 2021-05-11
Replace lolex with @sinonjs/fake-timers

harryadel:
lolex is deprecated.
Screenshot_2021-05-11_12-24-01

PR #11837, opened 2022-01-06
Replace lru-cache with lru-native2

harryadel:
This PR replaces [lru-cache](https://github.com/isaacs/node-lru-cache) which is no longer maintained with [lru-native2](https://github.com/adzerk/node-lru-native). Both retain the same API so there's not any significant changes. 

I also thought about modernizing the code of caching-compiler especially the [Promise.await](https://github.com/meteor/meteor/blob/devel/packages/caching-compiler/caching-compiler.js#L305) which relies on fibers afaik but thought to ask you first. What do you think?

### DIFF
--- a/packages/caching-compiler/caching-compiler.js
+++ b/packages/caching-compiler/caching-compiler.js
@@ -1,8 +1,8 @@
+import { createHash } from 'crypto';
+import assert from 'assert';
+import LRUCache from 'lru-native2';
 const fs = Plugin.fs;
 const path = Plugin.path;
-const createHash = Npm.require('crypto').createHash;
-const assert = Npm.require('assert');
-const LRU = Npm.require('lru-cache');
 
 // Base class for CachingCompiler and MultiFileCachingCompiler.
 CachingCompilerBase = class CachingCompilerBase {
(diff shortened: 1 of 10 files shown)


harryadel:
@StorytellerCZ Could you please update the dev bundle with lru-native2?

filipenevola:
Hi @harryadel, you are changing multiple packages.

I know the change is the same, but multiple PRs are necessary to get good coverage in the reviews.

I would like to know what issues is this PR addressing as well. It's important to keep the work around solving issues, implementing Roadmap features or in discussions that we have an agreement already. This is what our guideline for contributions say. 

Could you provide what issue this is fixing?

harryadel:
>I know the change is the same, but multiple PRs are necessary to get good coverage in the reviews.
I do agree on the value of modular, single-purpose PRs but it's not some a divine code we have to abide by. Multiple PRs are good when we're tackling a complex problem entailing moderate to big changes to the code. 

This PR only replaces the package with *zero changes* to the code; same methods are still in the new package.    

>Could you provide what issue this is fixing?
While this PR doesn't solve an immediate problem, it aims to improve just the overall status of Meteor by replacing a less maintained package with a more active one. This is similar to what was done here: #11420

zodern:
`lru-cache` released version 7 this month, so it seems to still be maintained..

harryadel:
@zodern You're right. I was too quick to open up a PR without proper consideration. Sorry for the confusion.

PR #11841, opened 2022-01-07
Update tar library for faster extractions/compressions

renanccastro:
This PR updates the tar library for faster extractions/compressions.

We are now using tar-fs and tar-streams, as they are 10x faster working with tar streams compared to the version we had of tar.

https://www.npmjs.com/package/tar-fs#performance

Testing locally, I've been able to see a reduction of 30% on deploy time(in the upload bundle step)

### DIFF
--- a/meteor
+++ b/meteor
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=14.18.2.3
+BUNDLE_VERSION=14.18.2.6
 
 # OS Check. Put here because here is where we download the precompiled
 # bundles that are arch specific.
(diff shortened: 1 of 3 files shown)


StorytellerCZ (approved):
This seem to be too easy. I think there was some issue when I have attempted this months ago. :thinking:  Or it might have been reversed due to a bigger reversal or it was deemed as not necessary at that time. Still, glad that this is finally being bumped.

renanccastro:
@StorytellerCZ it was still a WIP when you reviewed :) 

----

From my tests, deploys are at least 30% faster when using tar-fs compared to the version of tar we were using.

filipenevola (commented):
LGTM

StorytellerCZ:
@renanccastro well done!

PR #11850, opened 2022-01-13
Fix #11827 custom user fields not publishing

### LINKED ISSUE #11827: Accounts.config.defaultFieldSelector is not considered in published default user fields
## Versions:

Meteor release 2.5
`accounts-base@2.2.0`
OS: Ubuntu 20.04 LTS
The last version of Meteor where the problem did _not_ occur, if applicable: not known to me right now

## The expected behavior:

```js
 Accounts.config({
    defaultFieldSelector: {
      _id: 1,
      username: 1,
      emails: 1,
      createdAt: 1,
      profile: 1,
      services: 1,
      customField: 1
    }
  });
```

should auto-add these fields to the current user `Meteor.user()` when logged-in (assuming these fields exist on the user document).
Related docs: https://guide.meteor.com/accounts.html#prevent-unnecessary-data-retrival

## The actual behavior:

Once logged in the `Meteor.user()` command only returns `{ _id: 'the userid', profile: {}, emails: [...] }`

## Reproduction:

Create meteor projectt, add Accounts and create a new user, then use the above `Accounts.config` and log in on the client. Open the Borwser console and run `Meteor.user()`

StorytellerCZ:
Fixes an issue where field defined in `Accounts.config.defaultFieldSelector` to be selected is not published to the client. Also added tests, but I'm not entirely certain about them.

The problem seem to have occurred that originally the `defaultFieldSelector` was to primarily prevent getting needless data retrieved, then there was a limitation of user data published to the client (and to ensure that people accidentally don't stop publishing fields that are needed for Meteor accounts to work), but it seems like this field was not taken into account and the functionality stayed only on client. The selection restrictions get applied both on client and server, but they don't make it to the publication till this PR.
This PR merges the required fields for publication and the desired fields. If the fields are restrictive then it publishes only the required fields.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -765,13 +765,22 @@ export class AccountsServer extends AccountsCommon {
     // Use Meteor.startup to give other packages a chance to call
     // setDefaultPublishFields.
     Meteor.startup(() => {
+      // Merge custom fields selector and default publish fields so that the client
+      // gets all the necessary fields to run properly
+      const customFields = this._addDefaultFieldSelector().fields;
+      const keys = Object.keys(customFields || {});
+      // If the custom fields are negative, then ignore them and only send the necessary fields
+      const fields = keys.length > 0 && customFields[keys[0]] ? {
+        ...this._addDefaultFieldSelector().fields,
+        ..._defaultPublishFields.projection
+      } : _defaultPublishFields.projection
       // Publish the current user's record to the client.
       this._server.publish(null, function () {
         if (this.userId) {
           return users.find({
             _id: this.userId
           }, {
-            fields: _defaultPublishFields.projection,
+            fields,
           });
         } else {
           return null;
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -767,8 +767,8 @@ export class AccountsServer extends AccountsCommon {
     Meteor.startup(() => {
       // Merge custom fields selector and default publish fields so that the client
       // gets all the necessary fields to run properly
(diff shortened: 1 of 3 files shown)


packages/accounts-base/accounts_tests.js
```diff
@@ -589,6 +590,15 @@ Tinytest.add(
     test.isNotUndefined(user[ignoreFieldName], 'full selector');
     test.isNotUndefined(user.username, 'full selector username');
 
+    Accounts._options = {};
+
+    // Test that a custom field gets retrieved properly
+    Accounts.config({defaultFieldSelector: {[customField]: 1}});
```
filipenevola, line 596: It's missing a test with defaultFieldSelector using negative values 0 or false. In this case, the test needs to check if this option is ignored.

filipenevola (changes requested):
Hi, I've asked for two changes. ;)

packages/accounts-base/accounts_tests.js
StorytellerCZ: Those tests are already present above, line 572 onward. Unless I misread the tests.

packages/accounts-base/accounts_server.js
```diff
@@ -765,13 +765,22 @@ export class AccountsServer extends AccountsCommon {
     // Use Meteor.startup to give other packages a chance to call
     // setDefaultPublishFields.
     Meteor.startup(() => {
+      // Merge custom fields selector and default publish fields so that the client
+      // gets all the necessary fields to run properly
+      const customFields = this._addDefaultFieldSelector().fields;
```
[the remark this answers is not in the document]
StorytellerCZ: Got it! Fixed!

filipenevola (approved):

Issue #11864, opened 2022-01-20
EJSON.equals fail to compare object and array if first param is object and second is array


jankapunkt:
This issue must be age-old but it is still present in Meteor 2.5 :open_mouth: 

Reporoduction open Meteor shell and enter the following:

```bash
import { EJSON } from 'meteor/ejson'

> EJSON.equals([], [])
true
> EJSON.equals([], {})
false
> EJSON.equals({}, [])
true
> 
```

Obviously `EJSON.equals({}, [])` should be false.

Checked the code and found the cause: https://github.com/meteor/meteor/blob/devel/packages/ejson/ejson.js#L489

```js
  if (a instanceof Array) {
    if (!(b instanceof Array)) {
      return false;
    }
    if (a.length !== b.length) {
      return false;
    }
    for (i = 0; i < a.length; i++) {
      if (!EJSON.equals(a[i], b[i], options)) {
        return false;
      }
    }
    return true;
  }
```

Potential fix would be to change this to

```js
 const aIsArray = a instanceof Array;
 const bIsArray = b instanceof Array;

  // if not both or none are array they are not equal
  if (aIsArray !== bIsArray) {
    return false;
  }

  if (aIsArray || bIsArray) {
    if (a.length !== b.length) return false;
    for (i = 0; i < a.length; i++) {
      if (!EJSON.equals(a[i], b[i], options)) return false;
    }
    return true;
  }
```

StorytellerCZ:
Nice job! Since the package uses ecmascript, we could replace `instanceof` with `Array.isArray`. That is if it is more performant.
Also shouldn't the second if statement be `&&` instead of `||`, so that length can then be compared?

jankapunkt:
You're right with both since instanceof checks the full prototype chain and array.isArray only checks the tostring value, although this can be tricked but I'm not sure if there is a practical exploit to it. And yes sorry the if branch should use &&

StorytellerCZ:
I would say to go ahead and create a PR for this. What do you think @renanccastro

filipenevola:
The fix will be available in 2.6.1-beta.1

PR #11866, opened 2022-01-21
fix(ejson): fix equals check with Arrays

jankapunkt:
Fixes issue #11864


StorytellerCZ:
I think this should be released as a patch.

filipenevola:
@jankapunkt thank you for this.

I would like to see at least one new test validating this fix, is this possible?

jankapunkt:
Hey @filipenevola yes I can add a test or two, of course

jankapunkt:
@filipenevola I added few tests

PR #11867, opened 2022-01-21
Release 2.6

filipenevola:
Read more in the [Milestone](https://github.com/meteor/meteor/milestone/87).

--- 

Latest News:
We've published now Meteor 2.6-rc.2

It includes Cordova changes as well, to support the latest Launch Screens for iOS.

[2.6 Migration Guide](https://deploy-preview-11867--meteor-guide.netlify.app/2.6-migration.html)
[Changelog](https://deploy-preview-11867--meteor-docs.netlify.app/changelog.html)

---

We are upgrading the MongoDB Node.js Driver to support MongoDB 5.0 in this version. Read more [here](https://github.com/meteor/meteor/pull/11831).

Pending items:
- MongoDB 5.0 Support
  - [x] Check aggregate behavior (and popular packages) - Need changes: https://github.com/sakulstra/meteor-aggregate/pull/8
  - [x] Check redis-oplog behavior (and other popular packages) - There is no usage of rawCollection, and we've maintained API compatibility.
  - [x] Migration Guide updates
    - [x] MongoDB Embedded will break between major versions of MongoDB (meteor reset is the solution)
    - [x] Mention explicility `aggregate` and explain eventual changes that are necessary  
  - [x] Analyze warnings reported [here](https://github.com/meteor/meteor/pull/11831#issuecomment-1011696597) when using meteor mongo: 
    - [x] ~~Mongo shell being superseded by mongosh~~
    - [x] ~~Warning about the "default write concern"~~
Mongo shell deprecation will be addressed in an upcoming release.

---

This item was not fixed but I'm not sure if it is going to affect people using TS:
- Meteor types should depend on the new typedefs coming from MongoDB npm module. [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58433). Help needed, [see](https://gist.github.com/filipenevola/f7cc0032ffac2020b15d5dc1e235639b). I was not able to reproduce it in the Meteor typescript skeleton running 2.6-beta4 (see below), I can run this code without typescript errors:
 ```js
import { LinksCollection } from '/imports/api/links';
import { IndexOptions } from 'mongodb';

function getIndexOptions(): IndexOptions {
  return {unique: true};
}
function insertLink(title: string, url: string) {
  LinksCollection.insert({ title, url, createdAt: new Date() });
  console.log(`getIndexOptions`, getIndexOptions());
} 
```

### CODE REVIEW

File: guide/source/2.6-migration.md [hunk 1]
```diff
@@ -0,0 +1,87 @@
...
+  - If you are using MongoDB Atlas, in MongoDB Atlas we were not able to migrate to a free MongoDB 5.x instance, so we had to migrate to a paid cluster in order to test the app properly, maybe this will change after February 2022; 
+- Set up your staging database with MongoDB 5.x and restore your production data there, or populate this database in a way that you can reproduce the same cases as if you were in production;
+- Run your app pointing your MONGO_URL to this new database, that is running MongoDB 5.x;
+- Run your end-to-end tests in this environment. If you don't have a robust end-to-end test we recommend that you test your app manually to make sure everything is working properly.
+- Once you have a stable end-to-end test (or manual test), you can consider that you are ready to run using MongoDB 5.x, so you can consider it as any other database version migration.
+
+We are not aware of any issues that were introduced by the necessary changes to support MongoDB, but it's important that you check your app behavior, especially if you have places where you believe you are not using MongoDB in a traditional way.
```
fredmaiaarantes, line 59: This sentence appears to be duplicated, almost the same was said in line 46.
I'm not sure, but maybe that was intentional to reinforce the importance of testing.

filipenevola: Each project will be in a different case (1 or 2) so people should read one or the other and not both. Got it?

dokithonon:
Hello,
After updating to the beta. I have this error 

Screenshot 2022-01-25 at 11 34 12

I do not use _ensureIndex in my code so I suspect this comes from a package I use but how to find where does the problem comes from ?
Thanks

filipenevola:
> I do not use _ensureIndex in my code so I suspect this comes from a package I use but how to find where does the problem comes from ?

@dokithonon these are coming from other packages (like synced-cron).

but the warning message was also in the wrong place as in this case Meteor internally is probably using createIndex already. I have fixed the position of this warning, it will be available in the next version.

filipenevola:
Hi, I need help with typedefs for MongoDB, [see the problem here](https://gist.github.com/filipenevola/f7cc0032ffac2020b15d5dc1e235639b).

pmogollons:
Looks like this version does not include the fix of node-fibers missing binaries included in 2.5.6

filipenevola:
> Looks like this version does not include the fix of node-fibers missing binaries included in 2.5.6

@pmogollons rc1 is coming and it will have.

mikowals:
On an M1 MacBook there is a large performance problem for database interactions in 2.6-rc.1.  I modified the `tests/main.js` file in fresh project from `meteor create` to do inserts and updates from the server.  The output with times is below with the code below that.

On 2.6-rc.1:
> ✓ inserts 100 (2372ms)
> ✓ updates 100 (2340ms)

On 2.5.6:
> ✓ inserts 100 (148ms)
> ✓ updates 100 (50ms)

    import assert from "assert";
    import { Mongo } from 'meteor/mongo'

    if (Meteor.isServer) {
      describe('basic db ops', async function() {
        this.timeout(30000);
        const count = 100;
        const col = new Mongo.Collection("test");
         
        after(() => col.remove({}));

        it(`inserts ${count}`, () => {
          for (let i = 0; i < count; i++) {
             col.insert({_id: i.toString()});
          }
          assert.strictEqual(col.find().count(), count);
        });

        it(`updates ${count}`, async function () {
          for (let i = 0; i < count; i++) {
            col.update({_id: i.toString()}, {$set: {foo: 'bar'}});
          }
          assert.strictEqual(col.find({foo: 'bar'}).count(), count);
        });
      });
    }

renanccastro:
@mikowals thanks for reporting
I can't repro this behavior in a Fedora Linux:

**RC:**
I20220131-08:36:39.798(-3)?     ✓ inserts 100 (151ms)
I20220131-08:36:39.881(-3)?     ✓ updates 100 (81ms)
**2.5.6:**
I20220131-08:37:51.569(-3)?     ✓ inserts 100 (162ms)
I20220131-08:37:51.633(-3)?     ✓ updates 100 (64ms)

Times are quite similar in my try. However, I've given it a try in an M1 machine and I see the same slowness as you do. Mongo doesn't run natively in an M1, it uses rosetta as a translation layer so this might be an issue in version v5 of the server.

The problem is inside mongo-server, as running the following in a mongosh session connected to meteor server:
 ```
  const count = 100;
  const time = Date.now();
  const insertLink = ({title, url}) => {
    db.links.insert({title, url, createdAt: new Date()});
  }
  for (let i = 0; i < count; i++) {
    insertLink({title: `${i}`, url: `http://${i}.com.br`});
  }
  console.log(`Time elapsed ${Date.now()-time}ms`)
```

Outputs the same slowness.

Digging deeper, I've narrowed it down to the default write concern on mongo v5, and I'm currently running the tests in the pr here: https://github.com/meteor/meteor/pull/11878

We will let you know when we release a new version for testing @mikowals 

Thanks again!

renanccastro:
2.6-rc.2 is published with the fix mentioned above regarding write concerns and a fix for issue #11877

mikowals:
Yes, 2.6-rc.2 no longer has the slow writes described above.  The output is:

> ✓ inserts 100 (157ms)
>  ✓ updates 100 (63ms)

Thanks for the investigation, description and fix Renan.

PR #11875, opened 2022-01-28
Update decimal.js

harryadel:
Might be nice to add it along #11867

### DIFF
--- a/packages/non-core/mongo-decimal/package.js
+++ b/packages/non-core/mongo-decimal/package.js
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Npm.depends({
-  "decimal.js": "10.2.1"
+  "decimal.js": "10.3.1"
 });
 
 Package.onUse(function (api) {


packages/non-core/mongo-decimal/package.js
```diff
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Npm.depends({
-  "decimal.js": "10.2.1"
+  "decimal.js": "10.3.1"
```
denihs, line 7: Hi @harryadel, sorry, but I didn't understand the purpose of bumping the version of this package without making any actual change to it. Could you explain it to me?

denihs (changes requested):

harryadel:
:thinking:

filipenevola:
@harryadel what are the improvements and changes in this minor version update?

We will be more strict with PRs to follow our guideline for contribution. https://github.com/meteor/meteor/blob/devel/CONTRIBUTING.md#proposing-your-change

harryadel:
@filipenevola I guess my argument would be the same as one [here](https://github.com/meteor/meteor/pull/11837#issuecomment-1033214636)

denihs (approved):

PR #11880, opened 2022-02-01
Update node version to 14.19.0

denihs:
Update node to 14.19.0.

More info in the changelog here: **https://github.com/nodejs/node/releases/tag/v14.19.0**


meteor
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-BUNDLE_VERSION=14.18.3.4
+BUNDLE_VERSION=14.19.0
```
renanccastro, line 3: you are missing a version here.
it should be
14.19.0.0

renanccastro (changes requested):

meteor
denihs: Got it, fixed!

renanccastro (approved):

PR #11882, opened 2022-02-02
Have build plugins handle caching for css minifiers

zodern:
Removes css minifier caching from isobuild since it isn't flexible enough to support use cases like postcss and tailwind. CSS minifiers now need to handle caching themselves, same as for other types of build plugins. This is how it worked before Meteor 1.8.

This also adds :

- `file.readAndWatchFileWithHash` to the files passed to css minifiers, which previously was only available to compilers. This function can be used to tell Meteor that a file is used for the architecture being minified so Meteor will rebuild when the file is modified. 
- `Plugin.fs.readdirWithTypesSync`. Similar to `Plugin.fs.readdirSync`, but returns [`dirents`](https://nodejs.org/api/fs.html#class-fsdirent) instead of strings
- `beforeMinify`. If a minifier has a `beforeMinify` method, it will be called near the beginning of each build before Meteor minifies any files for that build, similar to the `afterLink` method compilers can have.


filipenevola (approved):
LGTM but I believe changes in HTML files are not triggering a new postcss run.

filipenevola (approved):

PR #11885, opened 2022-02-04
Fix flatten object issue when the object is empty on oplog converter

### LINKED ISSUE #11884: Meteor 2.6: Meteor publish no longer updating field correctly
A client doesn't receive the updated value for a field published by returning a cursor.

This example code is a `tests/main.js` file to replace the file in a new project from `meteor create`.  These tests pass in Meteor 2.5.6 but fail in Meteor 2.6.  Also, it might have to do with the updated field being an object as the client correctly receives an update like `{$set: {nested: {count: 1}}}`).  I focused on nested fields because I hit test failures in an app but I haven't checked whether nesting is essential to reproduce.

    import assert from "assert";
    import { Mongo } from "meteor/mongo";
    import { Meteor } from "meteor/meteor";

    describe("publish-bug", function () {
      const col = new Mongo.Collection("test");
      it("package.json has correct name", async function () {
        const { name } = await import("../package.json");
        assert.strictEqual(name, "publish-bug");
      });

      if (Meteor.isClient) {
        describe('client', function () {
          this.timeout(30000);
          let handle;
          after(() => {
            handle.stop();
          })
          it('subscribes', async () => {
            handle = Meteor.subscribe('test');
            await new Promise(resolve => setTimeout(resolve, 7000));

            const result = col.findOne();
            assert.strictEqual(result.count, 1);
            assert.strictEqual(!!result.nested?.state, true);
          })
        })
      } else {
        // server code
        col.remove({});
        col.insert({_id: 'test', count: 0});
        Meteor.setTimeout(() => {
          col.update({_id: 'test'}, {$set: {count: 1, nested: {state: {}}}});
        }, 5000);
        Meteor.publish(function() {
          return col.find({});
        })
      }
    });

This is probably related to #11883, though that describes a field dropping.

I was on an M1 MacBook seeing this test fail.

renanccastro:
Fix flatten object issue when the object is empty on oplog converter
Fix #11884

### DIFF
--- a/packages/mongo/oplog_v2_converter.js
+++ b/packages/mongo/oplog_v2_converter.js
@@ -57,7 +57,6 @@ const nestedOplogEntryParsers = (
   const setObjectSource = { ...i, ...u };
   const $set = Object.keys(setObjectSource).reduce((acc, key) => {
     const prefixedKey = `${prefixKey}${key}`;
-
     return {
       ...acc,
       ...(!Array.isArray(setObjectSource[key]) && typeof setObjectSource[key] === 'object'
@@ -97,6 +96,7 @@ function flattenObject(ob) {
 
     if (typeof ob[i] == 'object' && ob[i] !== null) {
       const flatObject = flattenObject(ob[i]);
+      if(Object.keys(flatObject).length === 0) { return ob; }
       for (const x in flatObject) {
         if (!flatObject.hasOwnProperty(x)) continue;
 
--- a/packages/mongo/oplog_v2_converter_tests.js
+++ b/packages/mongo/oplog_v2_converter_tests.js
@@ -1,6 +1,6 @@
 import { oplogV2V1Converter } from './oplog_v2_converter';
 
-Tinytest.add('oplog - v2/v1 conversion', function(test) {
+Tinytest.only('oplog - v2/v1 conversion', function(test) {
   const entry1 = {
     $v: 2,
     diff: { scustom: { sEJSON$value: { u: { EJSONtail: 'd' } } } },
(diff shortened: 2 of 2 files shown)


packages/mongo/oplog_v2_converter_tests.js
```diff
@@ -50,6 +50,16 @@ Tinytest.add('oplog - v2/v1 conversion', function(test) {
     JSON.stringify({ $v: 2, $set: { 'a.b': 2 } })
   );
 
+  //set a new nested field inside an object
+  const entry51 =  { "$v" : 2, "diff" : { "u" : { "count" : 1 }, "i" : { "nested" : { "state" : {  } } } } };
+  // the correct format for this test, inspecting the mongodb oplog, should be "nested" : { "state" : {  } } }
+  // but this is not a case in which we cant flatten the object without collateral, so we are considering
```
filipenevola, line 56: Can't or can?

packages/mongo/oplog_v2_converter.js
```diff
@@ -97,6 +96,7 @@ function flattenObject(ob) {
 
     if (typeof ob[i] == 'object' && ob[i] !== null) {
       const flatObject = flattenObject(ob[i]);
+      if(Object.keys(flatObject).length === 0) { return ob; }
       for (const x in flatObject) {
         if (!flatObject.hasOwnProperty(x)) continue;
```
zodern, line 101: I'm not sure how important performance is here or how large the objects are so it could be fine as is, but this loop is duplicating the work done by Object.keys. It would be a little more efficient to store the result of Object.keys in a variable, and then loop over the keys.

renanccastro: thanks, @zodern , I've included your suggestion in the other PR I made!

PR #11888, opened 2022-02-06
Fix flatten object issue when internal object value is an array on oplog converter

### LINKED ISSUE #11887: Meteor 2.6: Array field converted to an object in publication
# Versions
Package versions: Meteor 2.6 and mongo 1.14.1

Operating system: MacOS (M1)

# Issue
Adding a new field to a document, with the following shape

```
foo: { bar: [1, 2, 3] }
```

results in the published `bar` array being converted to an object. i.e. it looks like this:

```
foo: { bar: {"0": 1, "1": 2, "2": 3} }
```

This issue occurs if the method is executed just server-side and also if the method is executed client-side.

Here is a simple app to reproduce the error: https://github.com/ninjaPixel/meteor-test-app

This issue also occurs when using mongo 1.14.0 (and that's as low as I'm allowed to drop the mongo package version while on Meteor 2.6).

# Last working version
This issue does not occur with Meteor 2.5.6 and mongo 1.13.0 (I'm unable to install mongo 1.14 within a Meteor 2.5.6 project).

Here is the same app, but with Meteor 2.5.6 and mongo 1.13.0 : https://github.com/ninjaPixel/meteor-test-app/tree/meteor-2-5-6-mongo-1-13

renanccastro:
Fix flatten object issue when internal object value is an array - fix #11887

Preserve object.keys call for performance reasons

### DIFF
--- a/packages/mongo/oplog_v2_converter.js
+++ b/packages/mongo/oplog_v2_converter.js
@@ -99,8 +99,6 @@ function flattenObject(ob) {
       let objectKeys = Object.keys(flatObject);
       if(objectKeys.length === 0) { return ob; }
       for (const x of objectKeys) {
-        if (!flatObject.hasOwnProperty(x)) continue;
-
         toReturn[i + '.' + x] = flatObject[x];
       }
     } else {
--- a/packages/mongo/oplog_v2_converter.js
+++ b/packages/mongo/oplog_v2_converter.js
@@ -94,10 +94,11 @@ function flattenObject(ob) {
   for (const i in ob) {
     if (!ob.hasOwnProperty(i)) continue;
 
-    if (typeof ob[i] == 'object' && ob[i] !== null) {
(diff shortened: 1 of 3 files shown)


packages/mongo/oplog_v2_converter.js
```diff
@@ -94,10 +94,11 @@ function flattenObject(ob) {
...
 
-    if (typeof ob[i] == 'object' && ob[i] !== null) {
+    if (!Array.isArray(ob[i]) && typeof ob[i] == 'object' && ob[i] !== null) {
       const flatObject = flattenObject(ob[i]);
-      if(Object.keys(flatObject).length === 0) { return ob; }
-      for (const x in flatObject) {
+      let objectKeys = Object.keys(flatObject);
+      if(objectKeys.length === 0) { return ob; }
+      for (const x of objectKeys) {
         if (!flatObject.hasOwnProperty(x)) continue;
```
zodern, line 102: The hasOwnProperty check isn't needed when using Object.keys.

filipenevola (approved):

PR #11891, opened 2022-02-07
Throwing error when no mail URL is set but the app is running in a 

denihs:
Based on the issue [#11709](https://github.com/meteor/meteor/issues/11709), I could reproduce it by not setting a MAIL_URL and running my app with --production.

Now, an error will be thrown when the app runs in a production environment but doesn't have a mail URL set, as a reminder.

Also, making sure there are no [settings](https://github.com/meteor/meteor/compare/release-2.6.1...feature/check-if-mail-url-is-set?expand=1#diff-81b3d1c2cbdddcb39f637030fdb2c2fe5e3190c2b3e0b638c68199f793aa7e5aR66) when calling the function [knownHostsTransport](https://github.com/meteor/meteor/compare/release-2.6.1...feature/check-if-mail-url-is-set?expand=1#diff-81b3d1c2cbdddcb39f637030fdb2c2fe5e3190c2b3e0b638c68199f793aa7e5aR63). Before, this function was being called from [here](https://github.com/meteor/meteor/blob/34596ca59213a4938f651b71261b3159e61b0e86/packages/email/email.js#L131) with the settings as an empty object, therefore never entering on this [`if`](https://github.com/meteor/meteor/compare/release-2.6.1...feature/check-if-mail-url-is-set?expand=1#diff-81b3d1c2cbdddcb39f637030fdb2c2fe5e3190c2b3e0b638c68199f793aa7e5aL65).

### DIFF
--- a/packages/email/email.js
+++ b/packages/email/email.js
@@ -257,7 +257,7 @@ Email.send = function(options) {
     // This check is mostly necessary when using the flag --production when running locally.
     // And it works as a reminder to properly set the mail URL when running locally.
     throw new Error(
-      'You do not provided a mail URL. You can provide it by using the environment variable MAIL_URL or using your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
+      'You have not provided a mail URL. You can provide it by using the environment variable MAIL_URL or your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
     );
   }
 
--- a/packages/email/email.js
+++ b/packages/email/email.js
@@ -251,11 +251,11 @@ Email.send = function(options) {
   }
 
   const mailUrlEnv = process.env.MAIL_URL;
-  const mailUrlUrlSettings = Meteor.settings.packages?.email;
-  if (Meteor.isProduction || mailUrlEnv || mailUrlUrlSettings) {
+  const mailUrlSettings = Meteor.settings.packages?.email;
+  if (Meteor.isProduction || mailUrlEnv || mailUrlSettings) {
     // This check is mostly necessary when using the flag --production when running locally.
     // And it works as a reminder to properly set the mail URL when running locally.
-    if (!mailUrlEnv && !mailUrlUrlSettings) {
+    if (!mailUrlEnv && !mailUrlSettings) {
       throw new Error(
         'You do not provided a mail URL. You can provide it by using the environment variable MAIL_URL or using your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
       );
--- a/packages/email/email.js
+++ b/packages/email/email.js
@@ -252,15 +252,16 @@ Email.send = function(options) {
 
   const mailUrlEnv = process.env.MAIL_URL;
   const mailUrlSettings = Meteor.settings.packages?.email;
-  if (Meteor.isProduction || mailUrlEnv || mailUrlSettings) {
+
+  if (Meteor.isProduction && !mailUrlEnv && !mailUrlSettings) {
     // This check is mostly necessary when using the flag --production when running locally.
     // And it works as a reminder to properly set the mail URL when running locally.
-    if (!mailUrlEnv && !mailUrlSettings) {
-      throw new Error(
-        'You do not provided a mail URL. You can provide it by using the environment variable MAIL_URL or using your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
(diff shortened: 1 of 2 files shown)


StorytellerCZ (approved):

packages/email/email.js
```diff
@@ -246,11 +249,18 @@ Email.send = function(options) {
...
     return;
   }
-  if (
-    Meteor.isProduction ||
-    process.env.MAIL_URL ||
-    Meteor.settings.packages?.email
-  ) {
+
+  const mailUrlEnv = process.env.MAIL_URL;
+  const mailUrlUrlSettings = Meteor.settings.packages?.email;
```
fredmaiaarantes, line 254: Should't this variable be named `mailUrlSettings` instead of `mailUrlUrlSettings`?

packages/email/email.js
```diff
@@ -246,11 +249,18 @@ Email.send = function(options) {
...
+  if (Meteor.isProduction || mailUrlEnv || mailUrlUrlSettings) {
```
fredmaiaarantes, line 255: I find this code a bit confusing, wouldn't be more clear if you do the validation first? Something like:

```javascript
  // This check is mostly necessary when using the flag --production when running locally.
  // And it works as a reminder to properly set the mail URL when running locally.
  if (Meteor.isProduction && !mailUrlEnv && !mailUrlUrlSettings) {
    throw new Error(
      'You do not provided a mail URL. You can provide it by using the environment variable MAIL_URL or using your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
    );
  }

  if (Meteor.isProduction || mailUrlEnv || mailUrlUrlSettings) {
    const transport = getTransport();
    smtpSend(transport, options);
    return;
   }
   devModeSend(options);
};
```

packages/email/email.js
denihs: Yes, it was a typo

fredmaiaarantes (approved):

packages/email/email.js
```diff
@@ -116,11 +119,11 @@ const getTransport = function() {
   if (
     this.cacheKey === undefined ||
     this.cacheKey !== url ||
-    this.cacheKey !== packageSettings?.service ||
+    this.cacheKey !== packageSettings.service ||
```
renanccastro, line 122: why this removal?

packages/email/email.js
```diff
@@ -116,11 +119,11 @@ const getTransport = function() {
...
     this.cacheKey !== 'settings'
   ) {
     if (
-      (packageSettings?.service && wellKnow(packageSettings.service)) ||
+      (packageSettings.service && wellKnow(packageSettings.service)) ||
```
renanccastro, line 126: here too. are you sure they can never be null?

packages/email/email.js
```diff
@@ -246,11 +249,19 @@ Email.send = function(options) {
...
-  ) {
+
+  const mailUrlEnv = process.env.MAIL_URL;
+  const mailUrlSettings = Meteor.settings.packages?.email;
+
+  if (Meteor.isProduction && !mailUrlEnv && !mailUrlSettings) {
+    // This check is mostly necessary when using the flag --production when running locally.
+    // And it works as a reminder to properly set the mail URL when running locally.
+    throw new Error(
+      'You do not provided a mail URL. You can provide it by using the environment variable MAIL_URL or using your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
```
renanccastro, line 260:
```suggestion
      'You have not provided a mail URL. You can provide it by using the environment variable MAIL_URL or your settings. You can read more about it here: https://docs.meteor.com/api/email.html.'
```

renanccastro (changes requested):

packages/email/email.js
denihs: It can't be [null](https://github.com/meteor/meteor/pull/11891/files#diff-81b3d1c2cbdddcb39f637030fdb2c2fe5e3190c2b3e0b638c68199f793aa7e5aR114)

packages/email/email.js
denihs: It can't be [null](https://github.com/meteor/meteor/pull/11891/files#diff-81b3d1c2cbdddcb39f637030fdb2c2fe5e3190c2b3e0b638c68199f793aa7e5aR114)

PR #11907, opened 2022-02-10
Fix sync return for insert methods inside _collection private method

renanccastro:
Fix sync return for insert methods inside _collection private method


filipenevola (approved):
If all tests are passing, this looks good to me.

KoenLav (approved):

renanccastro:
Released on mongo@1.14.4

Issue #11855, opened 2022-01-15
Minimongo's projection compatibility regression


mitar:
This PR introduced a bug in how projection is computed in comparison with the server:

https://github.com/meteor/meteor/pull/10499

If you run the test against MongoDB proper, you see the original results, not results as changed in PR:

```
$ docker run --rm -d --name mongo mongo:latest
$ docker exec -t -i mongo mongo
> db.test.insert({
...     setA: [{
...       fieldA: 42,
...       fieldB: 33,
...     }, {
...       fieldA: 'the good',
...       fieldB: 'the bad',
...       fieldC: 'the ugly',
...     }],
...     setB: [null, {
...       anotherA: { },
...       anotherB: 'meh',
...     }, null, {
...       anotherA: 1234,
...       anotherB: 431,
...     }, null],
...   });
> db.test.find().pretty()
{
	"_id" : ObjectId("61e30b2a3361d28a6bdff138"),
	"setA" : [
		{
			"fieldA" : 42,
			"fieldB" : 33
		},
		{
			"fieldA" : "the good",
			"fieldB" : "the bad",
			"fieldC" : "the ugly"
		}
	],
	"setB" : [
		null,
		{
			"anotherA" : {
				
			},
			"anotherB" : "meh"
		},
		null,
		{
			"anotherA" : 1234,
			"anotherB" : 431
		},
		null
	]
}
> db.test.find({}, { 'setA.fieldA': 1, 'setB.anotherB': 1 }).pretty()
{
	"_id" : ObjectId("61e30b2a3361d28a6bdff138"),
	"setA" : [
		{
			"fieldA" : 42
		},
		{
			"fieldA" : "the good"
		}
	],
	"setB" : [
		{
			"anotherB" : "meh"
		},
		{
			"anotherB" : 431
		}
	]
}
```

As you see, no `null` values in the result.

I discovered this because [for my package](https://github.com/peerlibrary/meteor-reactive-mongo) I run [minimongo tests against real MongoDB](https://github.com/peerlibrary/meteor-reactive-mongo/blob/master/tests_prepare.js). As you see in that file, I have to disable many tests because they fail on MongoDB proper. I find this pretty sad. I think it would be really beneficial if Minimongo tests would run against MongoDB. There are few known extensions (like sorting by providing a function, etc.) but for most of the things we want full compatibility. Which is currently not so. I have opened an issue for this [quite some time ago](https://github.com/meteor/meteor-feature-requests/issues/33).

[...]

PR #11908, opened 2022-02-10
Support the new "projection" field inside the decision of using oplog for a published cursor or not

renanccastro:
Also support the new "projection" field inside the decision of using oplog for a published cursor or not

It was spawning errors as the decision was true for unsupported cursor fields

Fixes this comment: https://github.com/meteor/meteor/issues/11855#issuecomment-1035120781


fredmaiaarantes (approved):

renanccastro:
Released on mongo@1.14.4

renanccastro:
cc @DonJGo

PR #11913, opened 2022-02-11
pass full result to userCallback

znewsham:
Currently, upon login, `validateResult` receives the entire result passed from server -> client (e.g., userId, token, expiry, additional options as set on the server)

However, `userCallback` and `onLogin` only receive the `{ type }` - this is an unusual inconsistency. I've opened this PR partly because the change is trivial, and partly to invoke a discussion around what the original use case was for this difference, and if there is any concern in passing the full result to all of these callbacks.

If there is concern in passing the token/expiry - we could easily extract those and pass back only the type and the additional details added by custom login methods.

An example of data that we'd like to pass from server -> client is an additional token (separate from the login token) that persists logouts which indicate that the device is known - we're able to work around this by storing the result in a `validateResult` callback, but it feels odd to do this here. 

Separately, it would be great if the `registerLoginHandler` functionality (and specifically the use of options passed OUT of it) was documented


denihs:
Hi @znewsham, 

It was by design that only the `type` field is returned to the client. You can see more details in the message of this [commit](https://github.com/meteor/meteor/commit/a638475a).

But the TLDR is to help reduce network transfer overhead, make sure we're not exposing server details on the client we shouldn't be, etc.

I did some tests to see what more was being returned on this `result` object, and it's this:

```
{
    id: "vp2bkHfLATr7yc5jT"
    token: "O_dw3xh9_pWi9WtH5aAHb1IGZeKfGY6JX6WQc6QHtvc"
    tokenExpires: Thu May 12 2022 17:43:19 GMT-0400 (Amazon Standard Time) {}
    type: "facebook"
}
```

Besides the type, I don't see any value in exposing the other fields to the client.

Do you have any specific use to the other fields worth of them to be exposed? If not, I think we shouldn't make this change.

znewsham:
Hi @denihs thanks for the response,

it seems there is some confusion here:

1. The data is being sent from server -> client, in fact the data `id`, `token` and `tokenExpires` MUST be passed to the client as the core meteor resume functionality requires it
2. ALL of this data is correctly passed to the client side `validateResult` callback - however, it is stripped to only `type` when being passed to `userCallback` and `onLogin`
3. In addition to these four fields - any field returned in the `options` field of the `registerLoginHandler` returned object are passed directly to the client (NOT inside options, interestingly) (the misunderstanding here is why it would be useful to be documented)

It is point (3) that is interesting to us - we want to return from server -> client a "deviceToken" that will persist logout to enable the user to bypass 2fa when a valid token is provided in future login calls - the exact implementation details here aren't particularly relevant, but receiving the token from server -> client is one of only two ways to implement it.

As stated above, this isn't a blocker for us - as we are able to do this currently using the `validateResult` callback, it just seems like an omission

denihs:
Hi @znewsham,

I reanalyzed everything, and You're right, this is already available to `validateResult`, so I can't think of a reason why it's being omitted there.

I'm merging your change. 

Thank you!

znewsham:
Amazing, thanks!

PR #11916, opened 2022-02-14
Release 2.7

renanccastro:
#### Highlights
* TailwindCSS 3.x support
* Typescript `4.5.4` upgrade
* New core package: `accounts-2fa` (Docs [preview](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html))
* Support for 2FA in `accounts-password` and `accounts-passwordless`
* Postcss Minifier (`minifier-css-postcss`) is now a core package.

Check all changes and issues included in the [Milestone](https://github.com/meteor/meteor/milestone/93).

#### How to use today

Update your project running:
`meteor update --release 2.7-rc.4`

#### Breaking Changes

N/A

#### Migration Steps

Read our [Migration Guide preview](https://deploy-preview-11916--meteor-guide.netlify.app/2.7-migration.html) for this version.


filipenevola:
Meteor 2.7-beta.2 is released!

renanccastro:
We have released Meteor 2.7-rc.0, which should be a final step before our final release.

renanccastro:
https://github.com/meteor/meteor/issues/11940

is currently a blocker for the final release. We are working on a fix

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,132 @@
...
+      {
+        $set: {
+          'services.twoFactorAuthentication': {
+            secret,
+          },
+        },
+      }
+    );
+
+    return svg;
```
StorytellerCZ, line 74: Could we also return the secret sot that we could display it to the user for manual entry in case that SVG doesn't display, backup or for people like me who put it into KeePassXC?
I'll try to implement this soon to test if the secret is the key to be used for that.

StorytellerCZ (changes requested):

docs/source/roadmap.md
```diff
@@ -36,6 +36,8 @@ The items in this section are the core team's priorities.
 
 ### Next major releases
 
+- Support building mobile apps using CapacitorJS;
```
KoenLav, line 39: @fredmaiaarantes in my opinion the Cordova integration (mirroring Cordova commands/config) has never worked nicely.

I see the benefit of having both within one project folder, but would not recommend mirroring the commands/config like Meteor did with Cordova.

We are already using Meteor with Capacitor as separate projects, so I'd happily provide any feedback here (or share some examples).

I would personally recommend using the same project structure as Capacitor itself for Android/iOS/Windows: when initializing Capacitor create a capacitor folder within the Meteor project and create a 'sync' command which copies the built Meteor assets to the Capacitor folder.

docs/source/roadmap.md
```diff
@@ -36,6 +36,8 @@ The items in this section are the core team's priorities.
...
+- Explore bringing Redis-oplog to core;
```
KoenLav, line 40: Great to see this is being moved up in the Roadmap, but wouldn't MongoDB Change Streams replace much of the need for Redis?

KoenLav:
@fredmaiaarantes I left some comments on this commit, would be happy to exchange thoughts:

https://github.com/meteor/meteor/pull/11916/commits/87d03ed0ad88a609e83b18a69c7e43bdb938a609

docs/source/roadmap.md
fredmaiaarantes: hey @KoenLav, I'd love to hear more about your experience on this. Do you have anything like this 'sync' command already in your projects? If you could share examples that would be really great.

docs/source/roadmap.md
KoenLav: Makes sense, thanks for clarifying!

packages/accounts-2fa/2fa-server.js
fredmaiaarantes: @denihs what do you think?

docs/source/roadmap.md
KoenLav: We don't have a sync command since we make use of server.url in Capacitor (so no need to sync the assets), which forces us to monitor compatibility between web/native ourselves, but enables us to perform updates simply by executing window.location.reload() (also it enables us to switch a single Tenant to another instance by modifying something in the database; window.location = anothersubdomain.ourapp.com).

You might expect the native functionality not to work when the web assets aren't bundled; but it works just fine as Capacitor injects itself (and any registered plugins) at every pageload (of a whitelisted domain, of course).

I'm not sure what I would recommend for Meteor as a default scenario, as Capacitor docs note using server.url is "only for development", but somewhere they clarified this is only the case because iOS doesn't really allow apps which can update themselves outside the app store (although it seems Apple isn't really being a PITA about it anymore).

You can give using server.url a try within a few minutes, simply run meteor on localhost:3000 and point server.url there (setting server.cleartext = true in Capacitor config as well).

Either way I would be more than happy to shed some light on our process as the only considerable difference between using 'sync' or 'server.url' is "where are the files coming from (web vs storage)", and I have some basic knowledge of how using sync would work since we experimented with that at first as well.

As a basic outline I would expect the following (pseudo):
- meteor capacitor init => meteor npm install @capacitor/core && mkdir ./capacitor && npm install @capacitor/core && npm install @capacitor/cli --save-dev && npx cap init
- meteor capacitor sync => meteor build && mv ./.meteor/local/build/programs/web.browser ./capacitor/www

Note: simply copying the built assets would probably not be sufficient since Capacitor expects an index.html, but that should be relatively easy to whip up.

npx cap init will start the normal Capacitor installation, which I think is just fine if we use sync, but if server.url would be the default we would probably need to overwrite/append to the config file (server block).

That's basically it, but I'd happily schedule a call to go over some details: https://forums.meteor.com/u/koenlav

fredmaiaarantes:
Version 2.7 RC1 will be released no later than next Monday.

docs/source/roadmap.md
fredmaiaarantes: Thanks for sharing, I had shared your comments here and we will reach out to you when we start working on this.

fredmaiaarantes:
> The next 2.7 RC will be released no later than next Monday.

2.7-rc.3 is available, but we have some tests that have failed and we need to review them.

fredmaiaarantes:
Version `2.7-rc.4` is published!

Please, test it out and give us some feedback.

flean:
> Version `2.7-rc.4` is published! Please, test it out and give us some feedback.
> 
> Check all changes and issues included in the [Milestone](https://github.com/meteor/meteor/milestone/93).
> 
> Read our [Migration Guide preview](https://deploy-preview-11916--meteor-guide.netlify.app/2.7-migration.html) for this version.

Works like a charm, had no issues with this and tailwind 3

StorytellerCZ:
@fredmaiaarantes Please consider #11979 for this release.

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,132 @@
...
+    const { username } = user;
+
+    const { secret, uri } = twofactor.generateSecret({
+      name: appName.trim(),
+      account: username,
+    });
+    const svg = new QRCode(uri).svg();
+
+    Meteor.users.update(
+      { username },
```
StorytellerCZ, line 64: All the selectors should primarily be by `_id`, username is an optional field and I have seen use cases where username is not assigned to user.

StorytellerCZ (changes requested):

packages/accounts-2fa/2fa-server.js
StorytellerCZ: I'll try to get a proposed fix by the end of today.

packages/accounts-2fa/2fa-server.js
StorytellerCZ: Tested and confirmed, getting the secret back would be nice so that it can be displayed when needed.

packages/accounts-2fa/2fa-server.js
```diff
@@ -0,0 +1,132 @@
...
+    Meteor.users.update(
+      { username: user.username },
+      {
+        $unset: {
+          'services.twoFactorAuthentication': 1,
+        },
+      }
+    );
+  },
+  has2faEnabled(selector) {
```
StorytellerCZ, line 126: Getting the following error here:
```
I20220320-15:36:47.080(1)? Exception while invoking method 'has2faEnabled' Error: Did not check() all arguments during call to 'has2faEnabled'
I20220320-15:36:47.080(1)?     at ArgumentChecker.throwUnlessAllArgumentsHaveBeenChecked (packages/check/match.js:515:13)
```
Strange thing is that the error persist even if I add check for selector then it still throws. Possibly something weird going on with callback...

StorytellerCZ: Indeed it is. It is on the user settings page, so user needs to be logged in. I call in in React in a function that sets a state:
```ts
const check2FAEnabled = () => {
    Accounts.has2faEnabled(userId, (result: boolean) => {
      setIs2FAEnabled(result)
    })
  }
```

harryadel:
Can we get a release date on this please? @fredmaiaarantes @denihs

denihs:
@harryadel, we're just solving one last matter with @StorytellerCZ. The plan is to release it today :crossed_fingers:

harryadel:
@denihs Great, thank you for your efforts!

packages/accounts-2fa/2fa-server.js
StorytellerCZ: I'll give it a try in about 10 hours after a good night sleep.

fredmaiaarantes: Hey, I'm merging this yet as we could not reproduce the issue. 
We are generating the 2.7 version, and if that issue still happens we can fix it in 2.7, ok @StorytellerCZ?
Thanks!

PR #11919, opened 2022-02-14
Adjusting accounts 2fa 

denihs:
- Validate if the user is logged in when calling has2faEnabled

- Make sure the login validation is successful before validating the 2FA situation.


docs/source/packages/accounts-2fa.md
```diff
@@ -89,47 +95,35 @@ To call this function the user must be already logged in.
...
-To verify whether or not a user has 2FA enabled, you can call the function `Accounts.has2faEnabled`:
-
-{% apibox "Accounts.has2faEnabled" "module":"accounts-base" %}
-
 As said at the beginning of this guide, this package is currently working with two other packages: `accounts-password` and `accounts-passwordless`. Below there is an explanation on how to use this package with them.
 
 <h3 id="working-with-accounts-password">Working with accounts-password</h3>
 
-With the function `Accounts.has2faEnabled`, you can check whether or not the user has 2FA enabled, and based on this information, you can directly call `Meteor.loginWithPassword` if the 2FA is not enabled, or redirect the user to a place where they can provide a code, in case they do have 2FA enabled.
+When calling the function `Meteor.loginWithPassword`, if the 2FA is enabled for the user, an error will be return to the callback, so you can redirect the user to a place where they can provide a code.
```
renanccastro, line 102:
```suggestion
When calling the function `Meteor.loginWithPassword`, if the 2FA is enabled for the user, an error will be returned to the callback, so you can redirect the user to a place where they can provide a code.
```

docs/source/packages/accounts-2fa.md
```diff
@@ -89,47 +95,35 @@ To call this function the user must be already logged in.
...
-        setShouldAskCode(true);
-        return;
-      }
-      // Normal login when they don't have 2FA enabled.
-      Meteor.loginWithPassword(username, password, error => {
-        if (error) {
-          console.error("Error trying to log in (user without 2fa)", error);
+    Meteor.loginWithPassword(username, password, error => {
+      if (error) {
+        if (error.reason === '2FA code must be informed') {
```
renanccastro, line 111: It would be better to check a code or an enum here instead of the string

docs/source/packages/accounts-2fa.md
```diff
@@ -153,62 +147,39 @@ So the call of this function should look something like this:
 
 <h3 id="working-with-accounts-passwordless">Working with accounts-passwordless</h3>
 
-Following the same strategy from the previous package, you can use the function `Accounts.has2faEnabled` to verify whether or not the user has 2FA enabled. If yes, you send them their token and on next step you receive their token and their 2FA code, otherwise, you still send them their token but on the next step you don't ask them for a 2FA code.
+Following the same logic from the previous package, if the 2FA is enabled, an error will be returned to the callback of the function [`Meteor.passwordlessLoginWithToken`](https://docs.meteor.com/packages/accounts-passwordless.html#Meteor-passwordlessLoginWithToken), then you can redirect the user to a place where they can provide a code.
 
 Here it's an example:
```
renanccastro, line 152:
```suggestion
Here is an example:
```

docs/source/packages/accounts-2fa.md
```diff
@@ -153,62 +147,39 @@ So the call of this function should look something like this:
...
-      return;
-    }
     // logging in just with token 
     Meteor.passwordlessLoginWithToken(
       email,
       token,
       error => {
-        if (error) {console.error('...', error);}
+        if (error) {
+          if (error.reason === '2FA code must be informed') {
```
renanccastro, line 163: same here

docs/source/packages/accounts-2fa.md
```diff
@@ -217,12 +188,14 @@ To integrate this package with any other existing Login method, it's necessary f
 
 2 - For the server, inside the function that will log the user in, you verify if the function `Accounts._is2faEnabledForUser` exists, and if yes, you call it providing the user you want to check if the 2FA is enabled, and if either of these statements are false, you proceed with the login flow. This function exists only when the package `accounts-2fa` is added to the project.
 
-If both statements are true, now you verify if a code was provided, if not throw an error, if it was provided, verify if the code is valid by calling the function `Accounts._isTokenValid`, if not, throw an error.
+If both statements are true, and the login validation was a success, now you verify if a code was provided, if not throw an error, if it was provided, verify if the code is valid by calling the function `Accounts._isTokenValid`, if not, throw an error.
```
renanccastro, line 191:
```suggestion
If both statements are true, and the login validation succeeds, you verify if a code was provided: if not, throw an error; if it was provided, verify if the code is valid by calling the function `Accounts._isTokenValid`. if `Accounts._isTokenValid` returns false, throw an error.
```

packages/accounts-2fa/2fa-server.js
```diff
@@ -124,6 +124,9 @@ Meteor.methods({
     );
   },
   has2faEnabled(selector) {
+    if (!Meteor.user()) {
+      throw new Meteor.Error(400, 'No user logged in.');
```
renanccastro, line 128: Have you checked if 400 is a pattern on other accounts packages?

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -80,38 +80,54 @@ Accounts.registerLoginHandler('passwordless', options => {
...
-  }
-
   const result = checkToken({
     user,
     selector,
     sequence,
   });
   const { verifiedEmail, error } = result;
 
-  if (!error && verifiedEmail) {
```
renanccastro, line 105: I didn't understand this removal...
This code was verifying the email if the user provided a valid passwordless token.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -80,38 +80,54 @@ Accounts.registerLoginHandler('passwordless', options => {
...
+
+  const is2faEnabled =
+    Accounts._is2faEnabledForUser &&
+    Accounts._is2faEnabledForUser({ _id: user._id });
+
+  const removeTokenFromUser = () => {
```
renanccastro, line 95: why remove token? isn't this the login handler?

renanccastro (changes requested):

docs/source/packages/accounts-2fa.md
fredmaiaarantes: I agree.

docs/source/packages/accounts-2fa.md
```diff
@@ -153,62 +147,39 @@ So the call of this function should look something like this:
...
       error => {
-        if (error) {console.error('...', error);}
+        if (error) {
+          if (error.reason === '2FA code must be informed') {
+            // send user to a page or show a component 
+            // where they can provide a 2FA code
+            setShouldAskCode(true);
+            return;
+          }
+          console.error('...', error);
```
fredmaiaarantes, line 169: Should this `console.log` be like this?

denihs: It's an example... You can handle the errors any way you prefer.

docs/source/packages/accounts-2fa.md
denihs: :+1:

docs/source/packages/accounts-2fa.md
denihs: :+1:

packages/accounts-passwordless/passwordless_server.js
denihs: I didn't realize the `verifiedEmail` was part of the verification. I've changed that.

packages/accounts-2fa/2fa-server.js
denihs: It seems to be, at least on the `accounts-password`. Example: https://github.com/meteor/meteor/blob/f9ce040fc3d5c0e92fbb9b00630bcbc25d0d2ffd/packages/accounts-password/password_client.js#L147-L149

brucejo75:
I am reading the preview docs [here](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html), which I assume are being updated with this PR.

Feedback:
- Why don't `Meteor.loginWithPassword` & [`Meteor.passwordlessLoginWithToken`](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-passwordless.html#Meteor-passwordlessLoginWithToken) return a `2FA` status code when the user has 2FA?  Then the callback could initiate a simple `verify2A` method?
- then you would not need [`Accounts.has2faEnabled`](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html#Accounts-has2faEnabled),  [`Meteor.passwordlessLoginWithTokenAnd2faCode`](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html#Meteor-passwordlessLoginWithTokenAnd2faCode) or [`Meteor.loginWithPasswordAnd2faCode`](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html#Meteor-loginWithPasswordAnd2faCode) at all?

Seems much simpler...  maybe I am missing something?

rj-david:
Does this mean that during the input of the 2FA, the meteor account status is logged in? That will have a lot of implications on templates/layout display, in cases when the user clicks the back button, etc, while the user is supposedly not logged in yet because 2fa has not been successful, yet.

On our app, we do 2FA with meteor as follows:
- user logs in (any login process)
- if verification succeeds, system checks if user has 2fa enabled in `Accounts.validateLoginAttempt()`
- if 2fa is enabled, login process is converted to an OTP-like login process with OTP and userId sent back to the client (this is currently a hack since we return the stringified data as part of the error object)
- with this specific error, 2fa form is displayed to the user, user submits 2FA token together with the OTP and userId
- OTP and userId are verified, 2FA token is verified, the user is logged in.

rj-david:
This line:

> With your suggestion, we would have to log in the user and wait for a token, but that would be a problem because we can't log in a user with invalid data, and now the 2FA code is part of the login flow, so it must be validated with the other fields.

Goes against what is currently implemented (which my feedback above was all about)

```
To verify whether or not a user has 2FA enabled, you can call the function `Accounts.has2faEnabled`:

{% apibox "Accounts.has2faEnabled" "module":"accounts-base" %}

This function must be called when the user is logged in.
```

brucejo75:
Hi @denihs,

Thanks for the considered response.

### incompatibility
Today both `Meteor.loginWithPassword` & `Meteor.passwordlessLoginWithToken` have no `result` argument to the callback.  Only an error argument.  The error argument could be used to transmit the `2FA` state only if `2FA` is set for this user.  That mean it is backward compatible for the non-`2FA` case.

### login stages
I think @rj-david outlines the elements of the solution above.  Here is how I see it working in the context of a single `Acounts.validate2FA` call:

- User is signed in via `Meteor.loginWithPassword` or `Meteor.passwordlessLoginWithToken`.  If no `2FA`, `result` is undefined in the callback (backward compatible).
- **IFF** user has defined `2FA`.
1. **Server** create a `2FA` token as part of the `2FA` sign in.  The user is not signed in.  But they have verified their credentials and are ready to receive the `2FA` token.  Send the `2FA` token back in the `result`.
2. **Client** the `Accounts` package can intercept the callback and save out a `localStorage` variable with the `2FA` token.  Then the Accounts package calls back the app caller with an `2FA Error`.
3. **Client** the app callback gets the `2FA` `error`.  So now the app knows they need to verify the `2FA`.  A UI is put up and the verification code is entered.  The app then passes the verification code to `Accounts.validate2FA`.  `Accounts.validate2FA` responds to the server with the `2FA` token (from `localStorage`) along with the verification code.  Then deletes the `localStorage` token.
4. **Server**  the server verifies that the token and the verification code are valid and then logs the user in.  Then deletes the server `2FA` token.

Essentially, if the user has `2FA` defined then all current login types will "stall" on login waiting for the new `2FA` token and `2FA` verification code.

brucejo75:
Also, it may be wise to add a new login type to [`Accounts.onLogin`](https://docs.meteor.com/api/accounts-multi.html#AccountsCommon-onLogin).

brucejo75:
@denihs, @rj-david,

Just thought of another way to go, that is quite simple:

1. When app logs user in with `Meteor.loginWithPassword` or `Meteor.passwordlessLoginWithToken`.  If the creds are valid and the user has `2FA` set up.  Then just respond with error `2FA`.
2. App could gather verification and then submit `Meteor.loginWithPasswordAnd2faCode` or `Meteor.passwordlessLoginWithTokenAnd2faCode`.  No need to query for `2FA` separately.

Are we wedded to these new method names?  Could add optional argument to `Meteor.loginWithPassword`, etc.  or overload the `password` argument to handle an object with the `2FA` in it.  Already handles a hashed password object.

denihs:
Hi @rj-david, too make clear the answer to this question clear:

> Does this mean that during the input of the 2FA, the meteor account status is logged in?

No, the login fails, and an error is returned in the callback, as the [doc](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html) says.

--- 

@brucejo75, about your last message, the steps you suggested are exactly what it's implemented today :wink:. In the first version of this package, you needed to use the method `has2faMethod`, but that is not required anymore. This PR is implementing this exact change.

> Could add optional argument to `Meteor.loginWithPassword`,

When I mentioned the incompatibility in my last message, I was talking about adding a new parameter to those methods. The name of the new methods and this parameter matter were discussed when we're implementing the first version, so we'll keep this way for now.

brucejo75:
I still see [`has2faEnabled`](https://github.com/meteor/meteor/blob/release-2.7/packages/accounts-2fa/2fa-client.js#L19) in the code.  I assume this is not needed for logging in?  Maybe just a convenience method to show the user they have 2fa turned on?

And I assume the [docs](https://deploy-preview-11886--meteor-docs.netlify.app/packages/accounts-2fa.html) are all wrong now and need to be fixed?

rj-david:
Thanks for that clarification, @denihs. I was still thinking of the previous flow related to `has2famethod` (which I thought was `has2faenabled`)

As a feedback, I am hoping that this implementation becomes generic or else it will only work with compatible packages and not to all login flow. For example now, if I am using Google or FB OAuth login packages, seems it is not compatible.

I believe the framework is in place to make it generic for all logins (e.g. just like in our case, we use `Accounts.validateLoginAttempt`). The missing piece here is on converting all login process to a generic token-based login (OTP) and how to return those token back to the client to be submitted together with the 2FA code. This can help cover all login packages and even custom logins created using the Accounts(multiserver) capabilities.

In our app, although it is working universally for all logins now (including 3rd party oauths), what I am not comfortable is the return value of the OTP details through the error object i.e. stringified object assigned to `error.reason`.

brucejo75:
@denihs, sorry I did not look at this PR changes (I thought it was already merged).  I see you changed the docs significantly.  🥴

PR #11922, opened 2022-02-15
Fix issues with HMR and meteor build --debug

zodern:
1. Fixes crash when the app was built with the `--debug` option, and then run with NODE_ENV set to production. Some files are expected to be replaceable in this situation, so I removed the check that was causing the crash. Fixes https://forums.meteor.com/t/unexpected-replaceable-file-in-production/57534.
2. When running the app created from `meteor build --debug`, it would log `Restart Meteor to enable hot module replacement.`. Restarting the app wouldn't enable HMR since it isn't being run by the meteor tool, so it no longer shows the message.

### DIFF
--- a/packages/hot-module-replacement/server.js
+++ b/packages/hot-module-replacement/server.js
@@ -1,5 +1,8 @@
-if (!process.env.METEOR_HMR_SECRET) {
-  console.log('Restart Meteor to enable hot module replacement.');
-} else {
+// METEOR_PARENT_PID 
+if (process.env.METEOR_HMR_SECRET) {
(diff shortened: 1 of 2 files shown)


denihs:
Hey @zodern, the tests are breaking. Could you take a look?

```
RangeError: /tools/utils/archinfo.ts: Maximum call stack size exceeded
```

zodern:
I'm not sure why the tests are failing. The babel error seems to happen while loading the meteor-tool, before it gets to building the packages this PR modifies. Locally running `./meteor --get-ready` works.

Issue #11914, opened 2022-02-12
Incorrect field values on client side after updateOne() with aggregation pipeline


Torgen:
Issue appears in 2.6
Issue did not appear in 2.5.6
Ubuntu 20.04 under WSL2 on Windows 10

Using the updateOne() method on rawCollection, passing an aggregation pipeline, I reorder the elements of an array field.
Expected behavior: Minimongo on the client initially contains the original order, and after sync it contains the modified order.
Actual behavior: Minimongo on the client contains an array that is neither the original nor final order; it has a duplicate element and is missing another element.

Repro: I saw this in my app during an upgrade. A version with the upgrade and some extra debugging logging is at https://github.com/Torgen/codex-blackboard/commit/1206e6ed44078e7b3dc4857673a2b1be694f379c; you can see the output in the github actions coverage.full.lcov run. The first test case ("drag-and-drop" > "allows drag and drop within a meta") shows the issue. For that run, the log messages for both the sim and real show that after the mutation, the puzzles field contains
```
["9yPbdgzeNXEBKMyoL","QnpLAx4aLJtF2w2qu","JTtwWmMi6gTrqrKtj","D79wFGQJ8fko3YzfQ","3HibG7MqfvLCyppw5","nwmoR5N3mYWA3sXtC","5vxGCadHmFKsqbNk6","GMqSk6h7qu6TQF6id"]
```
but at some point on the client, the puzzles field contains 
```
["9yPbdgzeNXEBKMyoL","QnpLAx4aLJtF2w2qu","JTtwWmMi6gTrqrKtj","D79wFGQJ8fko3YzfQ","3HibG7MqfvLCyppw5","nwmoR5N3mYWA3sXtC","5vxGCadHmFKsqbNk6","5vxGCadHmFKsqbNk6"]
```
(i.e. the last two elements are both "5vxGCadHmFKsqbNk6" and "GMqSk6h7qu6TQF6id" is missing). If you run the app tests yourself the values will differ because of the random ID generation, but I see the problem consistently.
Since I had only recently made the change to use the aggregation pipeline to make the change atomically, I was able to revert it and go back to the double checked logic that does more database operations, which is https://github.com/Torgen/codex-blackboard/commit/bef7eec80721a7cd6e9f207eafd35519feef5c44. The tests pass at this commit, which shows that the difference is using the aggregation pipeline in updateOne().

renanccastro:
Hi @Torgen thanks for reporting
[...]

PR #11929, opened 2022-02-16
Fix multiple array operators bug and add support for debug messages

renanccastro:
The format for oplog v2 might include multiple array type operations. This PR fixes the bug that was caused by only going over the first one.

This PR also adds support for debug messages, which will help debug issues with the new format.

Fixes https://github.com/meteor/meteor/issues/11914

### DIFF
--- a/packages/mongo/oplog_v2_converter.js
+++ b/packages/mongo/oplog_v2_converter.js
@@ -6,6 +6,18 @@
 //  diff: { u: { key1: 2022-01-06T18:23:16.131Z, key2: [ObjectID] } }
 // }
 
+function logConverterCalls(oplogEntry, prefixKey, key) {
+  if (!process.env.OPLOG_CONVERTER_DEBUG) {
+    return;
+  }
+  console.log('Calling nestedOplogEntryParsers with the following values: ');
+  console.log(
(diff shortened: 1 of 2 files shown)


packages/mongo/oplog_v2_converter.js
```diff
@@ -24,30 +36,64 @@ on mongo 4
...
+  return !Object.keys(possibleArrayOperator).find(
+    key => key !== 'a' && !key.match(/u\d+/)
+  );
 };
-const nestedOplogEntryParsers = (
-  { i = {}, u = {}, d = {}, ...sFields },
-  prefixKey = ''
-) => {
+
+function logOplogEntryError(oplogEntry, prefixKey, key) {
```
fredmaiaarantes, line 44: Shouldn't you check here if `OPLOG_CONVERTER_DEBUG` is enabled?

filipenevola (approved):
Nice idea with the message to open new issues.

packages/mongo/oplog_v2_converter.js
renanccastro: In case of an unrecoverable error, I've left this message. It's a different case, that var will suppress debug info, which is really verbose; this one, in contrast should never show.

PR #11938, opened 2022-02-19
docs: fix example code in security guide

soolidtech:
I think there is a small typo in guide related  to security

As you'll see, dev environment settings weren't consistent with helpmentOptions.

Let me know if I missed something.


soolidtech:
Oh, and I think there's a typo with `helpmentOptions` variable name. Shouldn't it be `helmetOptions` ? If so, i can update this pull request.

denihs (approved):

Issue #11900, opened 2022-02-09
[2.5.6] import of npm library gives empty object


trusktr:
I'm blocked. I can't run a Meteor app because import of a library returns an empty object, as described in all these threads:

- https://github.com/meteor/meteor/issues/10065
- https://forums.meteor.com/t/importing-from-npm-returns-an-empty-object/44491
- https://github.com/thereactivestack-legacy/meteor-webpack/issues/235
- https://stackoverflow.com/questions/55382092/importing-gitlab-package-from-meteorjs-returns-empty-object
- https://forums.meteor.com/t/solved-require-import-npm-require-all-return-empty-object/50153/6
- https://ethereum.stackexchange.com/questions/60335/web3-empty-object-server-side-in-meteor
- https://github.com/VulcanJS/Vulcan/issues/2580

I don't know if this ever worked in previous versions of Meteor, but from the threads we can see the issue celebrated quite a few birthdays already.

I think @benjamn might have the most clue as to what this might be caused by; he architected Meteor's module system.

Reproduction:

trusktr:
Added a reproduction.

eric-burel:
@trusktr Awesome that you have a simple reproduction!

For the record, this seems to specifically affect ES Modules (Material UI in my case), and the server build.

I recently hit an unexpected issue recently when exploring ES modules, with NPM packages and Next.js: files with ".js" extensions maybe interpreted differently depending if `"type": "module"` is enabled or not.
- If "type": "module" is enabled in package.json, ".js" is treated as ES modules
- If not, eg when using conditional exports instead, ".js" is treated as CommonJS

Maybe it's related? Like Meteor wrongly consider the package as CommonJS during the server build.

trusktr:
> Maybe it's related? Like Meteor wrongly consider the package as CommonJS during the server build.

Meteor doesn't understand Node.js ESM or package.json `exports` fields yet (https://github.com/meteor/meteor/issues/10906).


> I saw that you have the package `dayjs` as a `devDependencies`.
> 
> I tested moving it to `dependencies`, then running `rm -rf node_modules`, then `meteor npm install`, and it was fixed.

Ah, moving it from `devDependencies` to `dependencies` worked.

I think this might be considered a bug because all other tools (Webpack, Rollup, Parcel, etc) all successfully build an app regardless of which field dependencies are in.

[...]

PR #11954, opened 2022-02-28
Fix some npm modules being imported as an empty object

zodern:
The `install` npm package used to call `module.useNode` to check how to require a module, and then handle throwing an error itself if the module wasn't found. However, now Meteor does most of that work when building the app, and in some cases an error is never thrown if a module was found during the build, but not when running the app.

This PR updates `module.useNode` so it matches how it is now used.

Fixes https://github.com/meteor/meteor/issues/11900.

### DIFF
--- a/packages/modules-runtime/server.js
+++ b/packages/modules-runtime/server.js
@@ -51,20 +51,6 @@ Module.prototype.useNode = function () {
     throw new Error('npmRequire must be defined to use useNode');
   }
 
-  var parts = this.id.split("/");
-  var start = 0;
-  if (parts[start] === "") ++start;
-  if (parts[start] === "node_modules" &&
-      parts[start + 1] === "meteor") {
-    start += 2;
-  }
-
-  if (parts.indexOf("node_modules", start) < 0) {
-    // Don't try to use Node for modules that aren't in node_modules
-    // directories.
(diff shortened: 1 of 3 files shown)


packages/modules-runtime/server.js
```diff
@@ -62,20 +62,12 @@ Module.prototype.useNode = function () {
...
     // Don't try to use Node for modules that aren't in node_modules
     // directories.
-    return false;
-  }
-
-  try {
-    npmRequire.resolve(this.id);
-  } catch (e) {
-    return false;
+    throw new Error('useNode can not be used to import modules outside of node_modules');
```
zodern, line 65: We might be able to remove this logic since the import scanner already checked it before adding the call to `useNode`.

denihs (approved):

Floriferous:
We used to install client-only NPM libraries in our `devDependencies`, I'm trying to upgrade to 2.7.3 from 2.5.2, and now we're seeing errors saying that we need to move the package out of devDependencies. 

Is this still how Meteor works? i.e. can you install a client-only package in your devDependencies and it will not affect your server-side bundle?

zodern:
@Floriferous this error should only happen if you import the module on the server. If you the module is only being used on the client, would you be able to create a reproduction I could look at?

PR #11956, opened 2022-02-28
Adjust import scanner cache size

zodern:
This makes the import scanner cache 4x larger.

With a small app that has:
```js
import * as a from '@mui/icons-material';
import * as b from '@mui/material';
```

the cache will be only 62% full.

### DIFF
--- a/tools/isobuild/import-scanner.ts
+++ b/tools/isobuild/import-scanner.ts
@@ -291,7 +291,7 @@ function setImportedStatus(file: File, status: string | boolean) {
 // of ImportScanner (which do not persist across builds).
 const LRU = require("lru-cache");
 const IMPORT_SCANNER_CACHE = new LRU({
-  max: 1024*1024,
+  max: Math.pow(2, 22),
   length(ids: Record<string, ImportInfo>) {
     let total = 40; // size of key
     each(ids, (_info, id) => { total += id.length; });


ouya99:
@zodern only 62% with those 2 imports?
But if i have to add more imports in the future? You increased the cache by factor 4
Can you make it a parameter we can set to adjust the cache to our needs? like "INC_CACHE_FACTOR=4"

zodern:
Considering how large those two npm packages are, I thought it would be difficult for an app to need twice the cache space used by them. If it is or becomes too small for an app we can increase the size again.

WilliamKelley:
I would like to see it configurable through an enviornment variable as well. My company's app has accumulated a large mix of icons and component packages over the years. While it's unfortunate and avoidable, we would need many times more the space than your experiment simulated.

ouya99:
@zodern just tested this PR and it's about 20-30% faster HMR with math.pow(2,23) instead of math.pow(2,22). 
Seems my project is large enough to benefit from a bigger cache.

denihs (approved):

ouya99:
@zodern @denihs 
Please consider increasing the cache even further since it makes a difference in middle-size projects as well.

ouya99:
@zodern @denihs 
seems that a mac with 8gb does have hmr speed issues. could it be a problem if that new "babel scanimports" is using swap mem instaed of real ram? how would you force the rebuild step to use ram instead?

WilliamKelley:
@ouya99 I believe it's best if you made a new issue(s) for these followups. This is a closed PR now

Issue #11838, opened 2022-01-06
"Error: Only 64-bit Intel and M1 processors are supported on OS X" on M1 pro


holmrenser:
I'm on a 14" macbook M1 pro and get the following error when trying to run a freshly downloaded/installed meteor 2.5.3:

```
Error: Only 64-bit Intel and M1 processors are supported on OS X
```

From [the blogpost announcing meteor 2.5.1](https://blog.meteor.com/meteor-2-5-1-with-native-mac-m1-support-fb1c4b6990b5) I would expect meteor to work on my machine.
Given that I'm on an M1, the error message saying only intel and M1 is supported is not very helpful.

paulincai:
Hi, it is the global installation instructions in that blog that I used to get it running with M1. I was already using it with Rosetta for a couple of months and transition was clear and simple. I'd put a highlight on removing the old version, empty the NPM  directory (node_modules in the projects) and make sure to use a native (M1) Terminal and not the older Rosetta terminal, in case you've been using a Rosetta terminal to run Meteor previously.

holmrenser:
Thanks for the comment. I am on a fresh machine and haven't used meteor on this machine before (so no rosetta terminal etc.). I have tried both the 'old' way of installing meteor and installing globally with npm, both result in the same problem.

In the update of my original issue you can see what I think is the problem, which is in the way meteor checks what hardware it is running on.

renanccastro:
Thanks for your report @holmrenser.
I don't think this issue will be a usual one, and even then there is a workaround as you've said in your thread.
Are you willing to write a PR? We would gladly merge it.

It's even a good first-time issue, so I will update the tags.

holmrenser:
Indeed probably not a frequent issue, but OTOH might also not be very uncommon (I think gnu core-utils is used frequently on mac: [75K installs in the last 30days](https://formulae.brew.sh/formula/coreutils)). I'll draft a PR later today!

jakobbouchard:
Just chiming in to say that I ran into the same issue. I temporarily fixed it by making the origin uname the default one instead.

PR #11975, opened 2022-03-16
Adjusted condition for archinfo.ts

riseth:
Solves #11838


StorytellerCZ:
@denihs is correct, there might be more places that will need adjustments, tests included to account for the new architecture option.

denihs:
This PR has been open for too long, and I don't think riseth will come back to it. I'm closing it for now.

PR #11981, opened 2022-03-20
2FA use userId instead of username

StorytellerCZ:
I have noticed that the 2FA implementation uses username as a the primary key to find user. That doesn't sound right as username is an optional parameter. As such I have switched it to use userId as the primary key and where feasible add backup of username.

I have also added `check` package to check for input and added a simple server side test for `_is2faEnabledForUser`. Nothing for client for now as I'm struggling with `has2faEnabled` which always returns undefined...

### DIFF
--- a/packages/accounts-2fa/2fa-server.js
+++ b/packages/accounts-2fa/2fa-server.js
@@ -2,6 +2,7 @@ import { Accounts } from 'meteor/accounts-base';
 import twofactor from 'node-2fa';
 import QRCode from 'qrcode-svg';
 import { Meteor } from 'meteor/meteor';
+import { check, Match } from 'meteor/check';
 
 Accounts._is2faEnabledForUser = selector => {
   if (!Meteor.isServer) {
@@ -13,7 +14,7 @@ Accounts._is2faEnabledForUser = selector => {
 
   if (typeof selector === 'string') {
     if (!selector.includes('@')) {
-      selector = { username: selector };
+      selector = { $or: [{ _id: selector }, { username: selector }] };
     } else {
       selector = { email: selector };
     }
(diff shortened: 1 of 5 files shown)


packages/accounts-2fa/2fa-server.js
```diff
@@ -13,15 +14,15 @@ Accounts._is2faEnabledForUser = selector => {
...
+      selector = { $or: [{ _id: selector }, { username: selector }] };
     } else {
       selector = { email: selector };
     }
   }
 
   const user = Meteor.users.findOne(selector) || {};
   const { services: { twoFactorAuthentication } = {} } = user;
-  return (
+  return !!(
```
StorytellerCZ, line 25: Needed to ensure that boolean is returned. If twoFactorAuthentication was `{}` it would return `undefined` instead of `false`.

packages/accounts-2fa/2fa-server.js
```diff
@@ -96,37 +96,44 @@ Meteor.methods({
...
   has2faEnabled(selector) {
-    if (!Meteor.user()) {
+    check(selector, Match.Maybe(Match.OneOf(String, Object)));
+    const userId = Meteor.userId();
+    if (!userId) {
       throw new Meteor.Error(400, 'No user logged in.');
     }
+
+    if (!selector) {
+      selector = { $or: [{ _id: userId }, { username: userId }] };
```
StorytellerCZ, line 134: Made selector optional. If none is passed, then it will look for the current user.

packages/accounts-2fa/2fa-server.js
```diff
@@ -96,37 +96,44 @@ Meteor.methods({
...
   has2faEnabled(selector) {
-    if (!Meteor.user()) {
+    check(selector, Match.Maybe(Match.OneOf(String, Object)));
+    const userId = Meteor.userId();
+    if (!userId) {
       throw new Meteor.Error(400, 'No user logged in.');
     }
+
+    if (!selector) {
+      selector = { _id: userId };
```
StorytellerCZ, line 134: Made selector optional. If none is passed, then it will look for the current user.

denihs (approved):
LGTM

Issue #11721, opened 2021-10-26
Build fails at minification while processing packages/modules.js


gischer:
I don't know how to reproduce this.  The app runs fine.  This was on Meteor 1.10.2, where I first got the problem.  I upgraded to current (2.5) with no change.

Here is the error message (apologies for how long it is):
```
Errors prevented bundling:                    
While minifying app code:                     

/home/gischer/.meteor/packages/standard-minifier-js/.2.7.1.1u87i0p.egqu++os+web.browser+web.browser.legacy+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/parser/src/parser/error.js:134:45:
Identifier 'o' has already been declared. (1:468270)
at Object._raise
(/home/gischer/.meteor/packages/standard-minifier-js/.2.7.1.1u87i0p.egqu++os+web.browser+web.browser.legacy+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/parser/src/parser/error.js:134:45)
at Object.raiseWithData
(/home/gischer/.meteor/packages/standard-minifier-js/.2.7.1.1u87i0p.egqu++os+web.browser+web.browser.legacy+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/parser/src/parser/error.js:129:17)
at Object.raise
(/home/gischer/.meteor/packages/standard-minifier-js/.2.7.1.1u87i0p.egqu++os+web.browser+web.browser.legacy+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/parser/src/parser/error.js:78:17)
at FlowScopeHandler.checkRedeclarationInScope
(/home/gischer/.meteor/packages/standard-minifier-js/.2.7.1.1u87i0p.egqu++os+web.browser+web.browser.legacy+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/parser/src/util/scope.js:152:12)
at FlowScopeHandler.declareName
(/home/gischer/.meteor/packages/standard-minifier-js/.2.7.1.1u87i0p.egqu++os+web.browser+web.browser.legacy+web.cordova/plugin.minifyStdJS.os/npm/node_modules/meteor/babel-compiler/node_modules/@babel/parser/src/util/scope.js:113:12)
[...]

PR #11982, opened 2022-03-21
Updating minifier-js

denihs:
From this [issue](https://github.com/meteor/meteor/issues/11721) and [this](https://github.com/meteor/meteor/issues/11930) one, we find out a bug on terser (minify package) that was affecting some applications (more of that [here](https://github.com/terser/terser/issues/1155)). 

For that, we're replacing terser for [SWC](https://swc.rs/docs/configuration/minification).  With SWC we aren't just fixing those issues but also decreasing the minification time.


docs/history.md
```diff
@@ -84,6 +84,9 @@ Read our [Migration Guide](https://guide.meteor.com/2.7-migration.html) for this
 * `meteor-node-stubs@1.2.1`
   - Adds support for [node:](https://nodejs.org/api/esm.html#node-imports) imports.
 
+* `minifier-jss@2.8.0`
+  - Replacing
```
harryadel, line 88: Missing swc?

harryadel: https://github.com/terser/terser/blob/master/CHANGELOG.md#v5121
>Fixed an issue with function definitions inside blocks (https://github.com/terser/terser/issues/1155)

denihs: My bad... I did a commit without adding the whole message. Thank you for pointing it out!

denihs:
We encountered some challenges when trying to test the minify with SWC. We realized that the settings file used by the package wasn't being loaded correctly, so the code wasn't being minified all the way. On top of that, the minification time wasn't what we expected.

Our assumption that SWC was faster than terser was based on this [benchmark](https://github.com/privatenumber/minification-benchmarks), but what we got was (in average):

```
	time	size after minify
swc	7.869s	750.8 kB

terser
	4.742s	 462.8 kB
```

The test was made in a new app with some features. It's possible that the big difference is because of the missing settings, hard to say.

Based on all this, we decided to keep using terser for now, mainly because they released a new [version](https://github.com/terser/terser/blob/master/CHANGELOG.md#v5121) fixing the issues we're having.

I've already tested their solution, and it's working. I'll open a new PR updating terser to this new version soon.

Thanks to everyone who reviewed it.

PR #11985, opened 2022-03-25
Reduce one DB call on 2FA login

StorytellerCZ:
I have noticed that in accounts-password, once we have the user object and we check for 2FA, we try to get the user again from DB. That looks like a needless DB round trip, so I have extracted the check for 2FA enabled into its own function and then used that in accounts-password to do the check with the user object we already have. This should save one trip to the DB.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -193,8 +193,7 @@ Accounts.registerLoginHandler("password", options => {
   // First the login is validated, then the code situation is checked
   if (
     !result.error &&
-    Accounts._is2faEnabledForUser &&
-    Accounts._is2faEnabledForUser(options.user)
+    Accounts._check2faEnabled(user)
   ) {
     if (!options.code) {
       Accounts._handleError('2FA code must be informed', true, 'no-2fa-code');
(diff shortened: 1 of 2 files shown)


StorytellerCZ:
I think something same could be done with passwordless as well.

packages/accounts-password/password_server.js
```diff
@@ -193,8 +193,7 @@ Accounts.registerLoginHandler("password", options => {
   // First the login is validated, then the code situation is checked
   if (
     !result.error &&
-    Accounts._is2faEnabledForUser &&
-    Accounts._is2faEnabledForUser(options.user)
+    Accounts._check2faEnabled(user)
```
denihs, line 196: You need to check if the function `Accounts._check2faEnabled` exists here, the same way we're doing with the `Accounts._is2faEnabledForUser`.

 If you don't have the package `accounts-2fa` on your project, the function `Accounts._check2faEnabled` will be undefined.

denihs (changes requested):

packages/accounts-password/password_server.js
StorytellerCZ: True. Another way would be to check if the package is installed, but this way it could also be turned into a feature by allowing third party packages to override this function if they follow the same structure. :thinking:

StorytellerCZ: Added a conditional here: `Accounts._check2faEnabled?.(user)` one less line, but if you think that added a line with expressly checking the line is better, I'll change to that.

denihs (approved):

PR #11991, opened 2022-03-28
Adjusts in the package accounts-2fa

denihs:
- Throwing error when a user is not found on `Accounts._is2faEnabledForUser`
- Remove vulnerability from the method `has2faEnabled`. It no longer takes a selector as an argument, just the callback.
- Auto-publishing the field `services.twoFactorAuthentication.type` for logged users.
- `generate2faActivationQrCode` now throws an error if it's being called when the user already has 2FA enabled.


packages/accounts-2fa/2fa-server.js
```diff
@@ -29,7 +29,13 @@ Accounts._is2faEnabledForUser = selector => {
     }
   }
 
-  const user = Meteor.users.findOne(selector) || {};
+  const user = Meteor.users.findOne(selector);
+  if (!user) {
+    throw new Meteor.Error(
+      500,
```
StorytellerCZ, line 35: Maybe 404 error would be more appropriate here?

StorytellerCZ (changes requested):

packages/accounts-2fa/2fa-server.js
fredmaiaarantes: I think it makes more sense. The docs also need to be updated to 404 in that case.

harryadel: Sorry to barge in, but why use HTTP error codes?? This is DDP, won't a descriptive title be better? 
https://github.com/meteor/meteor/blob/365604c91020db2c36b491e7ea332ed730492994/packages/ddp-server/livedata_server.js#L663

denihs: It's the pattern that the packages are using, but sure enough, I'm going to use a descriptive code in this case.

docs/history.md
```diff
@@ -4,7 +4,8 @@
 
 #### Breaking Changes
 
-N/A
+* `accounts-2fa@1.0.1`
```
zodern, line 7: Since there is a breaking change, the package's major version should be increased.

Issue #11750, opened 2021-11-05
Use of createIndex instead of ensureIndex fails on older indexes with `safe: true`


brianlukoff:
In this PR (https://github.com/meteor/meteor/pull/11556), Accounts started using `createIndex` instead of `ensureIndex` for indexes on `username`, `emails.address`, etc. The problem is that older Meteor apps (created prior to https://github.com/meteor/meteor/pull/4340) have many indexes created with the `safe: true` option (including important indexes like the unique constraint on `emails.address`). But upgrading to Meteor 2.4 -- which uses `createIndex` in place of `ensureIndex` -- fails on startup, because `createIndex` fails when an existing index with the same keys already exists but with `safe: true`.

For example, here's the underlying command that Accounts tries to run on startup:
````
> db.users.createIndex({'emails.address':1}, { unique: true, sparse: true });
{
	"ok" : 0,
	"errmsg" : "Index with name: emails.address_1 already exists with different options",
	"code" : 85,
	"codeName" : "IndexOptionsConflict"
}
````

The result is that it's impossible to upgrade to Meteor 2.4 or higher without manually dropping any existing indexes with `safe: true` and letting Accounts recreate them on startup. This isn't a problem in a development environment, but on a production database it would require either downtime or operating dangerously without a unique constraint on `emails.address` between the time it is dropped and recreated without `safe: true`.

Thanks to @StorytellerCZ for figuring this out!

menewman:
Potentially a duplicate of #11666?

brianlukoff:
Yes - thanks!

WilliamKelley:
@brianlukoff I'm trying to get more, exact information to write a non-manual fix for this. If you haven't manually dropped the indexes yet, can you paste the direct output of `db.users.indexInformation()`? Or let me know the version of Meteor your project was first created on so I can pin down a reproduction? Thanks

brianlukoff:
Thanks for working on this -- happy to help! That doesn't seem to work in the Mongo console though -- I am getting `uncaught exception: TypeError: db.users.indexInformation is not a function`. I believe we started with Meteor 1.1.0.1.

WilliamKelley:
Sorry, that is a driver api. I think `db.users.getIndexes()` should work. Thanks for the reply and info

brianlukoff:
Got it -- no problem. Is there an email address I can send to you privately (or can I send to you on the Meteor Slack)?

WilliamKelley:
[...]

Issue #11666, opened 2021-10-03
MongoError: Index with name: username_1 already exists with different options


kentonv:
Starting with Meteor 2.4, the `_ensureIndex()` method has different behavior: Previously, if the index already existed, regardless of options, then the call did nothing. Now, if the options differ, then the call throws an exception.

For new apps, the new behavior is clearly preferable.

Unfortunately, with Sandstorm, we've discovered that many of Meteor's own index-creation calls -- including the ones in accounts-base -- fail when run against older databases. It appears that Meteor used to create indexes with an option `"safe": true`, but no longer does. Mongo sees this as a change in options, and so all of Meteor's own calls to `createIndex()` fail if the indexes were originally created using a sufficiently old version of Meteor. This means the app cannot start unless the user manually goes and deletes all these indexes.

Note that the "safe" option was removed from Mongo quite some time ago, which may mean that users don't see this issue unless they are using old versions of Mongo. Sandstorm is, sadly, stuck on Mongo 2.6. The reason for this is, updating to a newer version of Mongo apparently requires some manual intervention to migrate data. Sandstorm is installed on thousands of individual users' machines where it has been auto-updating for over 7 years. Many of those users are non-technical, and we cannot ask them to take manual action to update their database. In order to update, we would need to build a fully-automated Mongo migration that happens in the background of live servers, which seems like a large engineering effort that we don't have time for.

I am not sure what to do about this. I think ideally, instead of failing when options differ, we'd drop and recreate the index with the correct options. But we need Meteor itself to do this with its own indexes. We have no way of knowing the full list of indexes that may need to be rebuilt, so it's hard for me to work around this.

kentonv:
In one particularly-old Sandstorm instance I looked at, I found the following indexes on Meteor.users had `"safe": true`:

`username_1`, `emails.address_1`, `services.resume.loginTokens.hashedToken_1`, `services.resume.loginTokens.token_1`, `services.resume.haveLoginTokensToDelete_1`, `services.resume.loginTokens.when_1`, `services.github.id_1`, `services.google.id_1`

StorytellerCZ:
Wow, that is a really old version of MongoDB. I don't even see the `safe` option in MongoDB 2.6 docs. Looking through recent commits I don't see any change that would indicate that we had changed anything about setting index options, so it might be some old MongoDB driver default. :thinking:

kentonv:
[...]

PR #11995, opened 2022-03-30
Re-create index when already exists 

StorytellerCZ:
Somehow my index options have changed on a whole lot of indexes, which caused Meteor to fail to start. Since I have promised to look into this in the past, but never got the urgency or time, this finally was an opportunity to address the following issues (at least partially):
#11666 #11750

So if there is an issue in creating a new index, now Meteor checks if it is an issue with changed options of the said index. If it is, it will re-create the index. If not, I have improved the error to provide the collection and name of the index where there is a failure, so it is easier to locate the problem.

I'm open to discussing adding a settigns key to disable this re-creation and instead thow an error. Either way I will have to do that another day, due to time constraints.


radekmie:
I'd make it an option instead, as recreating an index on a large collection may block it for a long time.

StorytellerCZ:
@radekmie done! Please review. :bow:

radekmie (approved):

Gywem (approved):

fredmaiaarantes (approved):

PR #12000, opened 2022-04-05
Creating new skeleton that will have tailwind configured by default

denihs:
As we now have a stable integration with [Tailwind version 3](https://guide.meteor.com/2.7-migration.html), we're creating this skeleton with Tailwind's basic configuration.

So now it will be possible to just run `meteor create --tailwind`, and a new project with Tailwind and React will be created.

Screenshot from 2022-04-07 14-50-18


tools/static-assets/skel-tailwind/client/main.css
```diff
@@ -0,0 +1,8 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
+body {
```
fredmaiaarantes, line 5: Do we need this custom css yet? I think it can be removed.

tools/static-assets/skel-tailwind/tests/main.js
```diff
@@ -0,0 +1,20 @@
+import assert from "assert";
+
```
fredmaiaarantes, line 2: I think it'd be good to create a few tests that check if Tailwind is configured and working correctly. Maybe checking if the required files exist? And/or if the rendered app has Tailwind classes applied?

StorytellerCZ:
It would be nice to see a bit more from Tailwind on the front page than just a styled button.

tools/static-assets/skel-tailwind/tests/main.js
denihs: As we found out, skeletons are automatically tested [here](https://github.com/meteor/meteor/blob/c03ab0f4262e9a05c14aba3e6cbdee2f88477e30/tools/tests/create.js#L50)

denihs:
It should be better looking now (img updated).

fredmaiaarantes (approved):

StorytellerCZ (approved):

PR #12005, opened 2022-04-08
Improve accounts docs

henriquealbert:
Small fix to help understand better where the `ServiceConfiguration` settings should be imported

### DIFF
--- a/docs/source/api/accounts.md
+++ b/docs/source/api/accounts.md
@@ -227,9 +227,8 @@ First, add the service configuration package:
 meteor add service-configuration
 ```
 
-Then, in your app (this example is for the Weebo service), import `ServiceConfiguration` inside your `server/main.js` file:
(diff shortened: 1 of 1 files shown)


docs/source/api/accounts.md
```diff
@@ -227,9 +227,11 @@ First, add the service configuration package:
 meteor add service-configuration
 ```
 
-Then, in your app (this example is for the Weebo service):
+Then, in your app (this example is for the Weebo service), import `ServiceConfiguration` inside your `server/main.js` file:
```
StorytellerCZ, line 230: I don't think we need to specify which file on server it should be, just that it needs to happen on server.

StorytellerCZ (changes requested):

docs/source/api/accounts.md
henriquealbert: sure, it makes sense. I've just updated it :)

denihs (approved):

PR #12007, opened 2022-04-13
Ability to disable sockjs on client side

maxpain:
I want to use uWebSockets on my project and want to disable SockJS for this case.

### DIFF
--- a/packages/socket-stream-client/browser.js
+++ b/packages/socket-stream-client/browser.js
@@ -163,8 +163,9 @@ export class ClientStream extends StreamClientCommon {
     };
 
     const hasSockJS = typeof SockJS === "function";
+    const disableSockJS = __meteor_runtime_config__.DISABLE_SOCKJS;
 
-    this.socket = hasSockJS
+    this.socket = hasSockJS && !disableSockJS
       // Convert raw URL to SockJS URL each time we open a connection, so
       // that we can connect to random hostnames and get around browser
       // per-host connection limits.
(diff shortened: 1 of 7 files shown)


packages/socket-stream-client/server.js
```diff
@@ -12,3 +12,7 @@ setMinimumBrowserVersions({
   safari: 7,
   electron: [0, 20],
 }, module.id);
+
+if (process.env.DISABLE_SOCKJS) {
```
StorytellerCZ, line 16: This needs to be properly documented. Please add this to new env var to documentation.

packages/socket-stream-client/browser.js
```diff
@@ -163,8 +163,9 @@ export class ClientStream extends StreamClientCommon {
     };
 
     const hasSockJS = typeof SockJS === "function";
+    const disableSockJS = __meteor_runtime_config__.DISABLE_SOCKJS;
 
-    this.socket = hasSockJS
+    this.socket = hasSockJS && !disableSockJS
```
StorytellerCZ, line 168: What happens if `disableSockJS` is true?

StorytellerCZ (changes requested):

packages/socket-stream-client/browser.js
maxpain: Then we use native WebSocket instead of SockJS implementation.

packages/socket-stream-client/server.js
maxpain: Done.

denihs (approved):

PR #12014, opened 2022-04-22
Fix using meteor-node-stubs in IE

zodern:
Downgrades the buffer and punycode npm packages to a version that is compatible with IE.


fredmaiaarantes (approved):

evolross:
Not sure if this is related or supposed to be a fix, but Meteor 2.7.2 still white-screens in Internet Explorer 9, 10, and 11.

https://forums.meteor.com/t/meteor-2-7-2-still-broken-in-internet-explorer-9-10-11/58134

PR #12028, opened 2022-05-08
Mongo Package Async API

radekmie:
This is a continuation of @filipenevola's work in #11605. See it first to get all the details.

**Important:** This PR changed significantly since https://github.com/meteor/meteor/pull/12028#issuecomment-1127543425, so keep that in mind. Here's the **current** behavior - the previous one is available below.

A summary of changes this PR makes:
1. **Added async methods on collections.**
    * All async methods have an `Async` suffix to their names. These are: `createCappedCollectionAsync`, `createIndexAsync`, `dropCollectionAsync`, `dropIndexAsync`, `findOneAsync`, `insertAsync`, `removeAsync`, `updateAsync`, and `upsertAsync`.
2. **Added async methods on cursors.**
    * All async methods have an `Async` suffix to their names. These are: `countAsync`, `fetchAsync`, `forEachAsync`, and `mapAsync`.
    * There's also `[Symbol.asyncIterator]`, so this code should work:
        ```ts
        for await (const document of collection.find(query, options)) /* ... */
        ```
3. **A few internal methods are now async.**
    * These are: `MongoConnection` (constructor), `MongoInternals.RemoteCollectionDriver` (constructor), `MongoInternals.defaultRemoteCollectionDriver`, `OplogHandle` (constructor).

Adding these async counterparts allows gradual migration, as both versions will coexists for the time being. At the moment, all async methods are calling sync one directly, so all hooks and wrappers should work as expected. **This will not be the case in the future releases.**

<details>
<summary>Behavior before <a href="https://github.com/meteor/meteor/pull/12028#issuecomment-1127543425">#12028 (comment)</a>.</summary>

A summary of changes this PR makes:
1. **Added `Mongo.Collection.create` - a way of creating async collections.**
    * It allows gradual migration, as it's allowed to have both sync and async instance of the same collection at the same time.
    * All async methods have an `Async` suffix to their names. These are: `createCappedCollectionAsync`, `createIndexAsync`, `dropCollectionAsync`, `dropIndexAsync`, `ensureIndexAsync`, `findOneAsync`, `insertAsync`, `rawCollectionAsync`, `rawDatabaseAsync`, `removeAsync`, `updateAsync`, `upsertAsync`.
    * An async collection cannot use sync methods and the other way round. The methods will be there but will throw an error when used.
    * (`new Mongo.Collection(name, { isAsync: true })` also works, but is not recommended.)
2. **Added async methods on cursors.**
    * These are: `countAsync`, `fetchAsync`, `forEachAsync`, `mapAsync`.
    * Cursor methods do NOT throw an error when called on synchronous collections, nor the other way round. It may change in future releases.
3. **Removed pre-1.0 compatibility of `Meteor.Collection` constructor options.**
    * It's no longer allowed to pass connection as the second argument. Use `{ connection }` instead.
    * It's no longer allowed to pass connection as `manager` in the second argument. Use `{ connection }` instead.
4. **A few internal methods are now async.**
    * These are: `MongoConnection` (constructor), `MongoInternals.RemoteCollectionDriver` (constructor), `MongoInternals.defaultRemoteCollectionDriver`, `OplogHandle` (constructor).

</details>


Julusian:
> These are: MongoConnection (constructor), MongoInternals.RemoteCollectionDriver (constructor), MongoInternals.defaultRemoteCollectionDriver, OplogHandle (constructor).

async constructors isn't entirely valid js. While it works with prototype syntax, it is not valid in es6/class syntax. As such, it is not possible to express it in typescript. So while it works, expect some frustration from typescript users about the typings being wrong, and if these classes ever get refactored to es6, then this will need changing

menewman:
@radekmie - First of all, thank you for picking up this project. It's critical work, and I'm very excited to see it moving forward!

> An async collection cannot use sync methods and the other way round. The methods will be there but will throw an error when used.

As a representative of a fairly large Meteor project ([our app](https://codesignal.com/)), I would really love to have it possible to use both sync and async collection/cursor methods on the same collection. (See my prior comment here: https://github.com/meteor/meteor/pull/11605#issuecomment-1092331956)

We have hundreds or thousands of places in our code where our server code interacts with MongoDB. If a given collection can only either be 100% sync or 100% async, it will be very challenging for us to make the transition from sync to async -- it is not an easy refactoring to convert server logic into being asynchronous because it requires recursively updating every consumer to make it asynchronous as well. However, if the two types of methods could co-exist on the same collection, it gives us a pathway to migrate from sync to async over time.

Would you be willing to consider making this "strict sync/async mode" logic optional for the sake of existing codebases? I understand why it's useful for a new project where you might want to enforce that everything is always async, but it will be painful for us, and I assume for others as well.

perbergland:
Let me +1 what @menewman just wrote. The idea to have every package come in an async and sync version is very impractical for migration.

For instance, this would require both an accounts and accounts-async package etc. I’d love to see an analysis of server-side packages to see which fiber-backed meteor methods they currently invoke that will become deprecated and not only the ones that rely on the meteor/mongo package.

I would rather see async versions of methods added to methods in all packages and then have an easy way to list (console.log at a minimum) all calls and patches made to the deprecated sync versions before finally removing all these calls.

menewman:
Thanks for the discussion, @Julusian. I think my interpretation of Filipe's nuanced comment was a little different. To quote the full context:

Post Filipe was replying to:
> > As the `*Async` methods throw an error on non-async collection, I was thinking about making non-`*Async` methods throw on async collections. The idea is to ensure that the async collections are already migrated to the new API, and the removal of sync API won't affect them.
> > * This can be an option, e.g., `throwOnSyncMethods`.
> > * It affects the cursors as well.

Filipe's response:
> I agree. It should be restrictive on mixing the two approaches in both sides (**in the ideal World**) but **the idea was to make room for people to migrate slowly to the new async methods instead of migrating everything in a single batch in a single collection, I think this could be beneficial as the main challenge here is to provide a nice way for people to migrate their code.**

> But we are also allowing creating two collection instances for the same collection in MongoDB using different approaches so maybe we could prevent the mixed use case inside the same instance. **It would be fine probably.**

(Emphasis added)

Basically, my takeaway from that was:
- In an ideal world (for new, or fully transitioned apps) there would only be one set of methods.
- Since we're sadly not in that ideal world, the goal of having `*Async` suffixed methods was to allow migration over time instead of forcing full migration all at once for a single collection. This was the original, preferred intention.
- But, strict sync/async could hypothetically work if we allow multiple collection instances. Filipe just said "this would probably be fine," not that it's necessarily the best approach.

The comment Filipe was responding to included two important things that aren't mentioned in the PR above:
  - Strict sync/async should be an **option**, not mandatory.
  - It should be consistent between collections & cursors. (This isn't a showstopper, but doesn't it make sense for consistency of the interface? Why be strict for one but not the other?)

I agree that if strict sync/async per collection were _optional_, I wouldn't have an objection to this plan.

Having to instantiate separate collection instances would be annoying/inelegant, but at least it would allow for partial migration. That said, my experience has been that it's not possible to do that with Meteor now. Trying to instantiate the same collection handle more than once results in this error:
https://github.com/meteor/meteor/blob/devel/packages/mongo/collection.js#L132

Relevant forum thread: https://forums.meteor.com/t/there-is-already-a-collection-named/36976/19

Even if it did work, I would argue that doing that is a totally different direction and would give reason to move away from `*Async` suffixes -- the only point of the suffixes is to allow both side-by-side. Otherwise:
```js
// This is the proposed version
db.someCollection.findOne(); // sync
db.someCollectionAsync.findOneAsync(); // returns promise, with redundant naming

// If we were fully committing to strict separate collections, no need for separate method names
db.someCollection.findOne(); // sync
db.someCollectionAsync.findOne(); // returns promise
```

menewman:
> I would rather see async versions of methods added to methods in all packages and then have an easy way to list (console.log at a minimum) all calls and patches made to the deprecated sync versions before finally removing all these calls.

@perbergland - Yep, I agree. Along those lines, within our app we've actually begun trying to remove the Fiber-based versions already by using a custom ESLint rule to identify as many of the old versions as possible, then replace with our [monkey-patched](https://github.com/codesignal/meteor-protomongo) async versions so we can get a headstart on the migration.

It's hard to do even with some help identifying the old Fiber-based calls; being able to have sync and async side-by-side with our monkey-patched version has been really helpful and is part of why I'm arguing that it's worth including that in the official core version 🙂

radekmie:
Sorry for not being around, but well... It happens :shrug: 

I'm going to respond to a lot of comments at once because it's easier for me... Hope I won't mix anything up!

> async constructors isn't entirely valid js. While it works with prototype syntax, it is not valid in es6/class syntax. As such, it is not possible to express it in typescript. So while it works, expect some frustration from typescript users about the typings being wrong, and if these classes ever get refactored to es6, then this will need changing

I'm aware of that, and [raised that issue back in October](https://github.com/meteor/meteor/pull/11605#issuecomment-943084395). However, this is a totally internal code, that I would expect no one will ever have to use outside of the core. (At least, I haven't seen it in tens if not hundreds of projects.)

Having said that, my idea was to get the PR _done_, not necessarily perfect. In this case, I'd leave the constructors `async` and change the API later. (Again, these are internal, API changes shouldn't matter much here.)

> As a representative of a fairly large Meteor project ([our app](https://codesignal.com/)), I would really love to have it possible to use both sync and async collection/cursor methods on the same collection. (See my prior comment here: https://github.com/meteor/meteor/pull/11605#issuecomment-1092331956)

As I wrote in the description (maybe it wasn't clear enough, sorry!), you will be able to. To use both APIs, you'd have two instances (objects) of the same collection. One will remain the same, and work as it is now. The other one, that you can add at any time, will be async-only.

> We have hundreds or thousands of places in our code where our server code interacts with MongoDB. If a given collection can only either be 100% sync or 100% async, it will be very challenging for us to make the transition from sync to async -- it is not an easy refactoring to convert server logic into being asynchronous because it requires recursively updating every consumer to make it asynchronous as well. However, if the two types of methods could co-exist on the same collection, it gives us a pathway to migrate from sync to async over time.

I think I answered it above, but let's highlight it - it will be the same collection _in the database_ but a different collection _in code_.

> Let me +1 what @menewman just wrote. The idea to have every package come in an async and sync version is very impractical for migration.

I don't see a need for separate packages, but I can imagine a lot of `*Async` methods added in the following months. I don't like it too, but can't see how else we could do it either.

> I would rather see async versions of methods added to methods in all packages and then have an easy way to list (console.log at a minimum) all calls and patches made to the deprecated sync versions before finally removing all these calls.

That's exactly how I see it. However, we're one step before - for now, the sync methods won't even notify you about it. (They will soon, probably.)

> I think this will be the key to your migration. Create a new Async version of the collection in parallel with the sync version, then it sounds like you will safely be able to make calls to both without issue.

Exactly!

> Or the easier option is for the package to require one of the sync/async collection to be provided, and your application will have to provide the correct one
>
> If I were a package author for a reasonably popular package, I would consider writing a wrapper around MongoCollection to provide a unified (either promise or fiber based) api to simplify my code while supporting both. But this only makes sense to do if already converting the api to be promise based

I think we could do better: provide a method on the collection instance to get its (a)sync counterpart. Something like `const collectionAsync = collection.toAsync()`. (The implementation is trivial and relies on collection's name; there's a problem with hooks and similar techniques.)

> The comment Filipe was responding to included two important things that aren't mentioned in the PR above:
>
> * Strict sync/async should be an **option**, not mandatory.
> * It should be consistent between collections & cursors. (This isn't a showstopper, but doesn't it make sense for consistency of the interface? Why be strict for one but not the other?)
>
> I agree that if strict sync/async per collection were optional, I wouldn't have an objection to this plan.

So at some point (I don't remember when), there was an idea to have a non-strict (a)sync calls. But it got replaced with multiple (strict) instances. So you still can have both sync and async operations on the same collection _in the database_, but you have to do it through multiple collections _in the code_.

And I do agree about the cursors, however, it wasn't as simple as I thought, and I wanted to make this PR public as soon as possible. My plan was to add this functionality later, but maybe we should do it now?

> Having to instantiate separate collection instances would be annoying/inelegant, but at least it would allow for partial migration. That said, my experience has been that it's not possible to do that with Meteor now. Trying to instantiate the same collection handle more than once results in this error:
https://github.com/meteor/meteor/blob/devel/packages/mongo/collection.js#L132

That changes in this PR - it's now possible to have multiple instances of the same collection.

> Even if it did work, I would argue that doing that is a totally different direction and would give reason to move away from *Async suffixes -- the only point of the suffixes is to allow both side-by-side. Otherwise:

I also thought about this strict APIs, but somehow wasn't sure about it. It's possible to do it on the TypeScript level too, of course. (Although with a breaking change.)

This PR is still a concept - let's hear what others think.

> @perbergland - Yep, I agree. Along those lines, within our app we've actually begun trying to remove the Fiber-based versions already by using a custom ESLint rule to identify as many of the old versions as possible, then replace with our [monkey-patched](https://github.com/codesignal/meteor-protomongo) async versions so we can get a headstart on the migration.
>
> It's hard to do even with some help identifying the old Fiber-based calls; being able to have sync and async side-by-side with our monkey-patched version has been really helpful and is part of why I'm arguing that it's worth including that in the official core version slightly_smiling_face

I saw the package a while back - that's a good one! After you already used it for a while - do you think it's better to have different return types of the same methods (e.g., `findOne`) based on the collection instance synchronicity or rather these `*Async` counterparts?

Julusian:
> I'm aware of that, and https://github.com/meteor/meteor/pull/11605#issuecomment-943084395. However, this is a totally internal code, that I would expect no one will ever have to use outside of the core. (At least, I haven't seen it in tens if not hundreds of projects.)
> 
> Having said that, my idea was to get the PR done, not necessarily perfect. In this case, I'd leave the constructors async and change the API later. (Again, these are internal, API changes shouldn't matter much here.)

Ah sorry, I would have read that message months ago and forgotten about it.. 
Good point, fine with me then :)

> I think we could do better: provide a method on the collection instance to get its (a)sync counterpart. Something like const collectionAsync = collection.toAsync(). (The implementation is trivial and relies on collection's name; there's a problem with hooks and similar techniques.)

While it would be a little weird, if you could make the async collection also implement `.toAsync()` that would make it really simple to make packages handle both sync and async collections. Just call that method and you know you have the async version. Avoids needing to do `const collectionAsync = collection.isAsync ? collection : collection.toAsync()` in every public method which accepts a collection.

> Even if it did work, I would argue that doing that is a totally different direction and would give reason to move away from *Async suffixes -- the only point of the suffixes is to allow both side-by-side. Otherwise:
> 
> I also thought about this strict APIs, but somehow wasn't sure about it. It's possible to do it on the TypeScript level too, of course. (Although with a breaking change.)

I think the Async suffix is important to keep for now. Otherwise people will feed the AsyncCollection into a package which is unaware of AsyncCollection, so it will call methods expecting them to be fiber based, and result in it leaving a trail of floating promises. 
That wont happen in a typescript codebase, but will be all too easy in javascript

radekmie:
I've spent the last few days thinking this through, and to be honest, I like the approach of adding `*Async` methods more. It adds literally no mental overhead, and makes it as easy to migrate as separate collections (from the Meteor's perspective), but making it much easier for developers.

I'll ask @filipenevola via email whether there was any point against it or if it was just a matter of preference. In the meantime, please comment further!

filipenevola:
Hi @radekmie @menewman we have made many changes along the way in the previous PR so now that we have decided that we are going to have a top-level await to wait for the connection we have more freedom on how to proceed.

It's important to make sure we have the commitment from Meteor core team that we are going to have top-level await ready in time for Node.js 14 End of Life as it would be a blocker and all this work would be useless without this support in Meteor.

Today this approach works because of Fibers and the goal is to remove it.

That said, I believe it's a good option to add *Async methods as we did but without requiring any other changes from the user code. 

It could be confusing for newcomers to have both options available in all collections but that is life, life is confusing sometimes. We can work with Meteor core team to provide nice updates to Meteor University & Docs & Guide to make sure newcomers are using the async option and not the sync option. The major goal of this API work is to avoid migrations in the future when Fibers is indeed removed.

I would keep the other changes that made possible all the async flow apart in a branch just in case we need to solve these problems again for different reasons.

tl;dr: now we can just add async version with a Promise wrapper in all the methods, just because we are relying on Fibers to await in a top-level.

off-topic: in my projects I use https://github.com/quavedev/collections instead of the regular constructor and I was already planning to support async versions there so projects using quave:collections would be ready to migrate either way.

radekmie:
OK then! I will update the PR the next week.

packages/mongo/collection.js
```diff
@@ -92,7 +96,15 @@ Mongo.Collection = function Collection(name, options) {
...
   else this._connection = Meteor.server;
 
-  if (!options._driver) {
+  if (options._driver) {
+    if (typeof options._driver.open !== 'function') {
+      throw new Error('If you are creating the driver manually using new ' +
+        'MongoInternals.RemoteCollectionDriver then you need to use ' +
+        'Promise.await() or await on it. Since it is async in recent ' +
+        'versions of Meteor. ' +
+        'Read more https://docs.meteor.com/changelog.html.');
```
menewman, line 105: Sorry for a trivial grammar nitpick!
```suggestion
      throw new Error('If you are creating the driver manually using new ' +
        'MongoInternals.RemoteCollectionDriver then you need to use ' +
        'Promise.await() or await on it since it is async in recent ' +
        'versions of Meteor. ' +
        'Read more: https://docs.meteor.com/changelog.html.');
```

packages/minimongo/constants.js
```diff
@@ -0,0 +1,20 @@
+/** Exported values are also used in the mongo package. */
+
+/** @param {string} method */
+export function getAsyncMethodName(method) {
+  return `${method.replace('_', '')}Async`;
```
menewman, line 5: Interesting -- would it make sense to keep the underscore prefix here so that the rule is simple? e.g., `_dropIndexAsync` to mirror `dropIndexAsync`? Otherwise it kind of turns a previously internal API into a public-looking async one. Maybe that is intentional though?

menewman (approved):
Thank you for these changes, @radekmie! I have no business reviewing this PR (not a Meteor maintainer, obviously 🙂) but wanted to review and give you a green checkmark anyway to show that I read through it as a consumer/user of Meteor. The interface looks good for us and others to start migrating away from Fibers 👍

packages/minimongo/constants.js
radekmie: It was like this when I overtook it and I think it's a good idea. These are around for a long time now and there's no clear alternative.

radekmie:
@menewman Thank you for that! It's not a problem at all, and I strongly encourage others to do the same!

henriquealbert (commented):
It looks great @radekmie! Great job! 👏

harryadel:
Can this be included in [2.7.4](https://github.com/meteor/meteor/milestone/98) or the next immediate version so we can give it a go?

denihs:
Hey guys, version 2.7.4-beta.2 is out and ready to be tested! We had problems releasing beta 0 and 1, so beta 2 is technically the first one :)

You can test it by updating your app with `meteor update --release 2.7.4-beta.2`

RobertLowe:
Update docs pls https://docs.meteor.com/api/collections.html

MarcelWeidum:
Is there an indication when 2.7.4 will be released?

plb666:
Hello everyone.

Hello everyone.

I have started the migration of a large application. I have a few doubts concerning some primitives that maybe some of you have already dealt with:

- forEachAsync,mapAsync: the callback remained synchronous, in case an asynchronous function needs to be called within it (e.g. an additional call to mongo) , it should become asynchronous.
- observe: in the synchronous version, the method added for records already in the db was called before terminating the invocation, in the asynchronous version should it be considered asynchronous?
- Blaze: will using a find in response from a helper still work? Should all functions that have become asynchronous in minimongo be converted to reactiveVar?

Thank you

radekmie:
(This conversation is continued on [forum](https://forums.meteor.com/t/migration-to-2-8-async-version/58836).)

brucejo75:
@radekmie, 

Does it make sense to expose the Async api on the client?  It introduces problems with the dependency engine.

See https://forums.meteor.com/t/tracker-and-async-mongo-methods/58887

I think folks may trip over this a lot...

radekmie:
(This conversation is continued on [forum](https://forums.meteor.com/t/tracker-and-async-mongo-methods/58887).)

PR #12057, opened 2022-06-06
Release 2.8 (previously 2.7.4)

denihs:
Current version: `2.8.0`.

**Highlights**
- New Mongo Package Async API. [PR](https://github.com/meteor/meteor/pull/12028)
- Updated MongoDB driver to 4.8. [PR](https://github.com/meteor/meteor/pull/12097)
- Updated MongoDB driver to 4.9. [PR](https://github.com/meteor/meteor/pull/12163)
- Update Cordova Android to 10.1.2. [PR](https://github.com/meteor/meteor/pull/12131)
- Upgrade Node to v14.20.0. [PR](https://github.com/meteor/meteor/pull/12078)
- Docs improvements and bug fixes.
- Update skeletons. 

Read more in the [Milestone](https://github.com/meteor/meteor/milestone/98).

You can update your app with `meteor update --release 2.8.0`. 

Preview docs: https://deploy-preview-12057--meteor-guide.netlify.app/2.8-migration.html


denihs:
Version 2.7.4-beta.2 is out and ready to be tested! We had problems releasing beta 0 and 1, so beta 2 is technically the first one :)

radekmie:
What is the plan to release this version? I want to continue with the MongoDB async API (i.e., make the async versions default and the sync call them), and I'm not sure whether or not I should wait for this release.

denihs:
Hey everyone, 

We planned to release an RC at the beginning of this week, but we decided to delay this.

We want to ensure the new updates will work as expected, and as I said in my last message, we're updating an internal app to test all the changes. So the plan is to replace all the MongoDB calls with the new Async ones.

Our concern is that our update wasn't smooth as we thought, and we had some unexpected behaviors. So we're working on figuring that out.

We're also learning with this experience of updating one of our apps, and with that, we're going to release a migration guide with the RC to make people's lives easier when trying to update to this version.

 For now, that's what I have. Sorry for the delay, and I'll keep you posted.

docs/source/commandline.md
```diff
@@ -119,14 +119,12 @@ React.
 
 `--apollo`
 
-Create a basic [Apollo + React](https://www.apollographql.com/) app.
+Create a basic [Apollo + Reac](https://www.apollographql.com/) app.
```
radekmie, line 122: Typo.

guide/source/2.7.4-migration.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Migrating to Meteor 2.7.4
+description: How to migrate your application to Meteor 2.7.4.
+---
+
+Meteor `2.7.4` introduce the new MongoDB Package Async API. For a complete breakdown of the changes, please refer to the [changelog](http://docs.meteor.com/changelog.html).
+
+For this new async API, we have new methods like `findOneAsync`, which behaves exactly like the `findOne` method, but it now returns a promise that needs to be solved to get the data.
```
radekmie, line 8: > [...] returns a promise that needs to be **solved** to get the data.

I guess it was supposed to say "resolved"? If not, it sounds a little too adventurous for a change log :stuck_out_tongue:

henriquealbert: I agree

StorytellerCZ:
Given what you are packing into this update shouldn't this be v2.8 to be semantic?

fredmaiaarantes:
> Given what you are packing into this update shouldn't this be v2.8 to be semantic?

Yes, we have discussed that. @denihs should make this change very soon.

denihs:
Version `2.8-beta.0` is out now :tada: 

We jumped from 2.7.4 to 2.8 to keep it consistent with semver rules.

filipenevola:
Readme with migration instructions is still using 2.7.4 @denihs

Grubba27:
We will be launching the next beta in the following days!

packages/mongo/mongo_livedata_tests.js
```diff
@@ -3229,18 +3229,6 @@ Meteor.isServer && testAsyncMulti("mongo-livedata - update with replace forbidde
   }
 ]);
```
fredmaiaarantes, line 3231: Hey @radekmie, I remove those two tests. They were the reason why Travis CI was failing. But I'd like to confirm with you if that is ok. From what I could understand, the connection behavior now changed, and we fail silently when we cannot connect to MongoDB.
I think the first one is ok to remove, but I'm not sure about the second. 

cc @Grubba27

radekmie: So, it looks like this:
1. `mongo-livedata - connection failure throws` was a test that we already had previously, and I only changed it to reflect the functional change. It is not failing silently now, as we await the connection promise in `Meteor.startup`, just like @zodern suggested.
2. `mongo-livedata - oplog - _onFailover` is a test I added, as there's no other means to test the `_onFailover` functionality.

Now, while I can imagine the latter being problematic, as it performs an actual failover of the database, the former should be fine good to stay, I think.

_(If possible, let's keep both, but just comment them out or something.)_

fredmaiaarantes: Oh, I see, I think it was changed to fail silently, but then you changed it back. Okay, I'll leave them commented for now. I am also pinging @filipenevola here so he can help us.

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,91 @@
...
+ - All async methods have an Async suffix to their names. These are: `countAsync`, `fetchAsync`, `forEachAsync`, and `mapAsync`.
+ - There's also `[Symbol.asyncIterator]`, so this code should work:
+    ```js
+    for await (const document of collection.find(query, options)) /* ... */
+    ```
+**A few internal methods are now async.**
+ 
+ - These are: `MongoConnection` (constructor), `MongoInternals.RemoteCollectionDriver` (constructor), `MongoInternals.defaultRemoteCollectionDriver`, `OplogHandle` (constructor).
+
+Adding these async counterparts allows gradual migration, as both versions will coexist for the time being. At the moment, all async methods are calling sync one directly, so all hooks and wrappers should work as expected.
```
radekmie, line 43: Since https://github.com/meteor/meteor/pull/12097 they no longer are. (It also impacts the comments below.)

Grubba27:
I've commented that test and keep as closely as it was when I started working on him

CONTRIBUTING.md
```diff
@@ -49,7 +49,8 @@ Current Reviewers:
   - [@aquinoit](https://github.com/aquinoit)
   - [@Grubba27](https://github.com/Grubba27)
 - [@filipenevola](https://github.com/filipenevola)
-- [@StorytellerCZ](https://github.com/StorytellerCZ)
+- [@Grubba27](https://github.com/Grubba27)
```
StorytellerCZ, line 52: Duplicate here, Grubba27 is already added at the end of the list.

StorytellerCZ (changes requested):

CONTRIBUTING.md
fredmaiaarantes: Just fixed, tks!

npm-packages/meteor-installer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "meteor",
-  "version": "2.7.4",
+  "version": "2.7.5",
```
StorytellerCZ, line 3: Shouldn't this track the Meteor version as close as possible, so `2.8.0`?

StorytellerCZ (changes requested):

xet7:
Hi,
when I tried `meteor update --release 2.8-beta.7`, when starting meteor, it did show this. Does it require some change?

```
W20220927-01:06:52.678(3)? (STDERR) (node:7585) [MONGODB DRIVER] Warning: cursor.count is deprecated and will be removed in the next major version, please use `collection.estimatedDocumentCount` or `collection.countDocuments` instead 
W20220927-01:06:52.678(3)? (STDERR)     at emitWarning (/Users/wekan/.meteor/packages/npm-mongo/.4.9.0-beta280.7.1duijd1.axvq++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb/lib/utils.js:965:20)
W20220927-01:06:52.678(3)? (STDERR)     at emitWarningOnce (/Users/wekan/.meteor/packages/npm-mongo/.4.9.0-beta280.7.1duijd1.axvq++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb/lib/utils.js:978:16)
W20220927-01:06:52.679(3)? (STDERR)     at FindCursor.count (/Users/wekan/.meteor/packages/npm-mongo/.4.9.0-beta280.7.1duijd1.axvq++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb/lib/cursor/find_cursor.js:92:37)
W20220927-01:06:52.679(3)? (STDERR)     at SynchronousCursor.fn [as _synchronousCount] (/Users/wekan/.meteor/packages/meteor-tool/.2.8.0-beta.7.1eong9y.tnkq++os.osx.x86_64+web.browser+web.browser.legacy+web.cordova/mt-os.osx.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:93:26)
W20220927-01:06:52.679(3)? (STDERR)     at SynchronousCursor.count (packages/mongo/mongo_driver.js:1169:17)
W20220927-01:06:52.679(3)? (STDERR)     at Cursor.<computed> [as count] (packages/mongo/mongo_driver.js:906:12)
W20220927-01:06:52.679(3)? (STDERR)     at AccountsServer._onCreateUserHook (models/users.js:1595:36)
W20220927-01:06:52.679(3)? (STDERR)     at AccountsServer.insertUserDoc (packages/accounts-base/accounts_server.js:1172:23)
W20220927-01:06:52.679(3)? (STDERR)     at AccountsServer._createUserCheckingDuplicates (packages/accounts-base/accounts_server.js:1466:25)
W20220927-01:06:52.679(3)? (STDERR)     at createUser (packages/accounts-password/password_server.js:910:19)
W20220927-01:06:52.679(3)? (STDERR)     at AccountsServer.Accounts.createUser (packages/accounts-password/password_server.js:994:10)
W20220927-01:06:52.679(3)? (STDERR)     at MethodInvocation.ATCreateUserServer (packages/useraccounts:core/lib/server_methods.js:103:27)
W20220927-01:06:52.679(3)? (STDERR)     at packages/check/match.js:118:15
W20220927-01:06:52.679(3)? (STDERR)     at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1259:12)
W20220927-01:06:52.679(3)? (STDERR)     at Object._failIfArgumentsAreNotAllChecked (packages/check/match.js:116:43)
W20220927-01:06:52.680(3)? (STDERR)     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1882:18)
W20220927-01:06:52.680(3)? (STDERR)     at packages/ddp-server/livedata_server.js:769:19
W20220927-01:06:52.680(3)? (STDERR)     at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1259:12)
W20220927-01:06:52.680(3)? (STDERR)     at packages/ddp-server/livedata_server.js:767:46
W20220927-01:06:52.680(3)? (STDERR)     at Meteor.EnvironmentVariable.EVp.withValue (packages/meteor.js:1259:12)
W20220927-01:06:52.680(3)? (STDERR)     at packages/ddp-server/livedata_server.js:765:46
W20220927-01:06:52.680(3)? (STDERR)     at new Promise (<anonymous>)
```

xet7:
Hmm, that warning seems to complain about this code:
```
const userCount = Users.find().count();
```
Here: https://github.com/wekan/wekan/blob/master/models/users.js#L1595

xet7:
And the warning above is this, so maybe I just change it to estimated:
```
Warning: cursor.count is deprecated and will be removed in the next major version,
please use `collection.estimatedDocumentCount` or
`collection.countDocuments` instead
```

xet7:
When I changed those to`collection.estimatedDocumentCount` at serverside, errors went away. Clientside minimongo can not use that different syntax.

radekmie:
@xet7 Is there a possibility of testing it locally? I'd like to look into it.

xet7:
@radekmie 

Yes.
```
git clone https://github.com/wekan/wekan

git checkout meteor-2.8
```
And build like at https://github.com/wekan/wekan/wiki/Emoji

Thanks !

rj-david:
There was a notice before not to use the new Mongo Async methods inside a Meteor method. Can we use it now?

Anything to watch out for as we start to transform our Meteor methods to async (if using the new async mongo api)?

denihs:
Hi @rj-david, you'll be able to use it soon, but it'll have some limitations. You can read more about it [here](https://deploy-preview-12057--meteor-guide.netlify.app/2.8-migration.html) in the migration guide we are preparing for this release. 

The next release will be the final one, so I recommend waiting until then to start using it.

rj-david:
@denihs, I need help to understand about this in the migration document

```
The callAsync limitations
You should never call a method if another method is still running, you need to be sure that only one method is running each time.
```

In apps using React, we have components that call meteor methods without consideration of the order. Does that mean that we can no longer rely on the callbacks of the methods to update the state of the components if those methods become async? Do we need to convert all those async methods to use await?

I'm hoping to understand if the limitation is to only ensure that if method B depends of the output of method A, then method B must wait on the results of method A.

Or is the limitation about an issue that the return of method A and method B can be exchanged if we run them simultaneously.

If the issue is only about dependency, then my react example above cam run methods simultaneously since those methods are not dependent on the results of other methods. The app will also become faster since the components are going to load data simultaneously (that will be a plus).

Hoping to confirm which limitation are we dealing with

zodern:
@rj-david this limitation only affects apps that use async method stubs, which wasn't supported before Meteor 2.8. While an async method stub is running, calling another method from outside the stub is broken.

The problem is while the async method stub is running, if `Meteor.call` is called, Meteor doesn't know if the call came from within the async method stub, or from code outside of the method stub. Until a better solution is found, Meteor currently has to assume it always is one of those, breaking the other. https://github.com/meteor/meteor/pull/12196 has Meteor assume the call is inside the stub. This allows Method stubs to still use Meteor.call to run the stubs of other methods, but means the app and packages are given the responsibility of ensuring methods aren't called while an async method stub is running.

With your example, if the method called by the first component has an async method stub, then it is possible the call from the second component would be considered to be from inside the async method stub meaning its stub would be run, but it wouldn't be run on the server.

I wrote a related comment about this a couple days ago: https://github.com/meteor/meteor/pull/12196#issuecomment-1279418845

With the current implementation, I'm not sure if most of my apps will be able to use async method stubs since there are places where they run multiple methods at the same time for better performance, and it is hard to predict when a method could be called while an async method stub is running.

Julusian:
So to clarify, this limitation only applies to methods which have a stub. And a stub is a client side implementation to simulate the method.  
And if you are using `Meteor.callAsync` for non-optimistic rpc then this limitation doesn't apply?  
Those limitation docs use the word 'stub' only once, and read very much like it applies to all method calls

packages/ddp-client/common/livedata_connection.js
```diff
@@ -568,6 +634,79 @@ export class Connection {
...
+         * But as the Meteor package is not compiled by ecmascript, it is unable to use newer syntax in the browser,
+         * such as, the async/await.
+         *
+         * So, to keep supporting old browsers, like IE 11, we're creating the logic one level above.
+         */
+        const currentContext = DDP._CurrentMethodInvocation._setNewContextAndGetCurrent(
+          invocation
+        );
+        try {
+          stubOptions.stubReturnValue = await stubInvocation();
```
zodern, line 698: With the current implementation, it might be difficult to use for packages that create functions to call each method, such as [zodern:relay](https://atmospherejs.com/zodern/relay), [grubba:rpc](https://atmospherejs.com/grubba/rpc), [mdg:validated-method](https://atmospherejs.com/mdg/validated-method). Using `Meteor.call` would be broken for async method stubs, and `Meteor.callAsync` would cause too many problems to use it by default.

I think if the microtask delay here  before calling `DDP._CurrentMethodInvocation._set` is removed  when the stub is sync, it would be fine for these packages to always use Meteor.callAsync. Unless there is something I am missing, when the stub is sync it should then work identically to `Meteor.call`, besides returning a promise.

Something like this might work:
```js
try {
  let stubResult = stubInvocation();
  if (isThenable(stubResult) {
    stubResult = await stubResult;
  }
} finally {
  DDP._CurrentMethodInvocation._set(currentContext);
}
```

zodern:
> So to clarify, this limitation only applies to methods which have a stub. And a stub is a client side implementation to simulate the method. And if you are using Meteor.callAsync for non-optimistic rpc then this limitation doesn't apply?

Yes.

Looking at the example again of two react components that each call a method to load data, it is unlikely either of those methods would use stubs if they are only loading data. The limitation would come into effect if a method to save some data is called, and while its async method stub is running react does a render. Any method calls done by components during this render would be considered as done by the method stub.

Something else that isn't clear in the migration guide is there is a similar limitation when an app uses allow/deny to do client side mutations, and also uses async method stubs. If a client side mutation is done outside of a stub while an async stub is running, it will be considered as part of the simulation and be reverted when the client receives the method results from the server.

rj-david:
Thanks for the clarification, @zodern @Julusian. That helps clarify what is expected. I am a bit hesitant to make any changes now unless it is confirmed what is expected for the remaining unhandled cases. I know this is an RC release, but it will give me the confidence to see the official position on what is to be expected for those unhandled cases in the future.

rj-david:
My suggestion in the migration document is to be very clear if we are talking about client or server or both.

radekmie:
Alternatively we could introduce an option to disable all simulations 🤷

zodern:
I'm wondering if instead of not supporting calling a method while an async method stub is running, if it would be better to instead not support using Meteor.call/Meteor.callAsync inside of an async method stub. Some of the benefits seem to be:

- Simpler to explain and understand to not use `Meteor.call` or `Meteor.callAsync` in async method stubs
- Easier to catch during code review. You also wouldn't need as deep of an understanding of the app and the libraries as you would need to know if and when a method could possibly be called while an async method stub is running
- When there is a problem caused by doing something that isn't supported, it would be easier and more reliable to reproduce problems caused by using Meteor.call inside of an async method stub compared to calling a method while an async method stub is running

docs/history.md
```diff
@@ -1,4 +1,81 @@
-## 2.7.3, 2022-05-31
+## v2.8, 2022-08-XX
+
+#### Highlights
+* New MongoDB Package Async API. [PR](https://github.com/meteor/meteor/pull/12028)
+* Node update to [v14.20.1](https://nodejs.org/en/blog/release/v14.20.1/) as part of the [September 22nd security release](https://nodejs.org/en/blog/vulnerability/september-2022-security-releases/)
+* Update MongoDB driver to 4.9. [PR](https://github.com/meteor/meteor/pull/12097)
+* Meteor.callAsync method. [PR](https://github.com/meteor/meteor/pull/12196)
```
fredmaiaarantes, line 7: I think we should also highlight the new skeletons for Solid and Chakra-UI.

docs/history.md
```diff
@@ -277,6 +354,21 @@ Read our [Migration Guide](https://guide.meteor.com/2.6-migration.html) for this
 
 #### Independent Releases
 
+## v2.5.8, 2022-05-31
+
+#### Highlights
+
+* Fixed 2.5.7 MongoDB error
```
fredmaiaarantes, line 361: What MongoDB error? Is there a link for reference?

docs/history.md
```diff
@@ -277,6 +354,21 @@ Read our [Migration Guide](https://guide.meteor.com/2.6-migration.html) for this
...
+* Patch release to update Node and npm versions.
```
fredmaiaarantes, line 362: Upgrade to which versions of npm and Node?

docs/source/commandline.md
```diff
@@ -140,41 +140,49 @@ Create a basic [Svelte](https://svelte.dev/) app.
 
 Create a basic [React](https://reactjs.org) + [Tailwind CSS](https://tailwindcss.com) app.
 
+`--chakra-ui`
+
+Create a basic [chakra-ui](https://chakra-ui.com/) app.
```
fredmaiaarantes, line 145:
```suggestion
Create a basic [React](https://reactjs.org) + [Chakra-UI](https://chakra-ui.com/) app.
```

docs/source/commandline.md
```diff
@@ -258,7 +266,7 @@ Your project should be a git repository as the commit hash is going to be used t
 The `cache-build` option is available since Meteor 1.11.
 {% endpullquote %}
 
-With the argument `--container-size` you can change your app's container size using the deploy command. The valid arguments are: `tiny`, `compact`, `standard`, `double`, `quad`, `octa`, and `dozen`. To see more about the difference and prices of each one you can check it [here](https://www.meteor.com/cloud#pricing-section).
+With the argument `--container-size` you can change your app's container size using the deploy command. The valid arguments are: `tiny`, `compact`, `standard`, `double`, `quad`, `octa`, and `dozen`. One more thing to note here is that the `--container-size` flag can only be used when the `--plan` option is already specified, else the use of `--container-size` option will throw an error with the message : `Error deploying application: Internal error`. To see more about the difference and prices of each one you can check it [here](https://www.meteor.com/cloud#pricing-section).
```
fredmaiaarantes, line 269:
```suggestion
With the argument `--container-size` you can change your app's container size using the deploy command. The valid arguments are: `tiny`, `compact`, `standard`, `double`, `quad`, `octa`, and `dozen`. One more thing to note here is that the `--container-size` flag can only be used when the `--plan` option is already specified, otherwise using the `--container-size` option will throw an error with the message : `Error deploying application: Internal error`. To see more about the difference and prices of each one you can check [here](https://www.meteor.com/cloud#pricing-section). 
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
+---
+title: Migrating to Meteor 2.8
+description: How to migrate your application to Meteor 2.8.
+---
+
+Meteor `2.8` introduce the new MongoDB Package Async API. For a complete breakdown of the changes, please refer to the [changelog](http://docs.meteor.com/changelog.html).
+
+For this new async API, we have new methods like `findOneAsync`, which behaves exactly like the `findOne` method, but it now returns a promise that needs to be resolved to get the data.
```
fredmaiaarantes, line 8:
```suggestion
For this new async API, we have new methods like `findOneAsync`, which behaves exactly like the `findOne` method, but now returns a promise that needs to be resolved to get the data.
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+<h3 id="why-like-this">Why this is important?</h3>
```
fredmaiaarantes, line 10:
```suggestion
<h3 id="why-like-this">Why is this new API important?</h3>
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+You may know that on Meteor we use a package called [Fibers](https://github.com/laverdet/node-fibers). This package is what makes possible to call async function, like `db.findOne()`, inside Meteor in a sync way (without having to wait for the promise to resolve).
```
fredmaiaarantes, line 12:
```suggestion
You may know that on Meteor we use a package called [Fibers](https://github.com/laverdet/node-fibers). This package is what makes it possible to call an async function, like `db.findOne()`, inside Meteor in a sync way (without having to wait for the promise to resolve).  
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+<h3 id="why-like-this">Why this is important?</h3>
+
+You may know that on Meteor we use a package called [Fibers](https://github.com/laverdet/node-fibers). This package is what makes possible to call async function, like `db.findOne()`, inside Meteor in a sync way (without having to wait for the promise to resolve).  
+
+But starting from Node 16, Fibers will stop working, so Meteor needs to move away from Fibers, otherwise, we'll be stuck on Node 14.
+
+If you want to know more about the plan, you can check this [discussion](https://github.com/meteor/meteor/discussions/11505).
+
+<h3 id="why-now">Why doing this now?</h3>
```
fredmaiaarantes, line 18:
```suggestion
<h3 id="why-now">Why doing this change now?</h3>
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+ - There's also `[Symbol.asyncIterator]`, so this code should work:
+    ```js
+    for await (const document of collection.find(query, options)) /* ... */
+    ```
+
+There is also a new Meteor method called `callAsync`. It should be used to call async methods.
+
+<h3 id="how-do-i-update">How can I start using these new features?</h3>
+
+We got a few examples for making it easy to use these new feature, you can look the code snippet bellow
```
fredmaiaarantes, line 42:
```suggestion
We got a few examples to make these new features easier to use, you can see the code snippet below:
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+const result = await Meteor.callAsync('removeByIDAsync', { id });
+
+// or even like this:
+
+Meteor.callAsync('removeByIDAsync', { id }).then(result => {
+   console.log(result);
+});
+```
+
+More examples can be retrieved from [this commit](https://github.com/fredmaiaarantes/simpletasks/compare/main...mongodb-async-api).
```
fredmaiaarantes, line 78: This link is comparing one branch to another, but this URL can be broken once the second branch no longer exists. I think we should remove this URL from here and if there are good examples there we could bring them into this guide.

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+
+<h2 id="callasync">The new callAsync </h2>
+
+As said before, the `callAsync` should be used to call async methods.
+
+We do not consider this version of the `callAsync` as the final product. It has its limitations when your methods have a [stub](https://docs.meteor.com/api/methods.html#:~:text=was%20received%20on.-,Calling,and%20you%E2%80%99ll%20have%20to%20wait%20for%20the%20results%20from%20the%20server.,-If%20you%20do), and it should be used where you know it won't affect other parts of your application. 
+
+We'll continue working on it to find a solution where these limitations don't exist.
```
fredmaiaarantes, line 87:
```suggestion
We will revisit these limitations soon and try to find a solution where these limitations do not exist.
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+We recommend that you start to write new methods and publications using async from this version forward, forcing internal APIs to also be async with time. And, of course, also updating your current ones. As soon you start, the better.
+
+But do it with caution, and keeping in mind the cases mentioned above.
+
+<h3 id="should-i-update">Can I update to this version without changing my app?</h3>
+
+Yes. You can update to this version without changing your app.
+
+<h2 id="older-versions">Migrating from a version older than 2.8?</h2>
```
fredmaiaarantes, line 225: I would move this question and answer to the beginning of the migration guide.

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+Meteor.callAsync('removeByIDAsync', { id }).then(result => {
+   // do something
+});
+Meteor.call('methodWithoutStub', { id }, (error, result) => {
+   // do something
+})
+
+// This is also ok, because even though removeByID has a stub
+// (SomeCollection.remove({ _id: id })), it is a sync method, 
+// so by the time removeByIDAsync runs, not stub will be running:
```
henriquealbert, line 117:
```suggestion
// so by the time removeByIDAsync runs, no stub will be running:
```

guide/source/2.8-migration.md
```diff
@@ -0,0 +1,249 @@
...
+
+Meteor.call('removeByID', { id }, (error, result) => {
+   // do something
+});
+Meteor.callAsync('removeByIDAsync', { id }).then(result => {
+   // do something
+});
+
+
+// But, this is NOT ok, because you would have two alive stubs at the same time:
```
henriquealbert, line 127:
```suggestion
// But, this is NOT ok, because you would have 2 stubs running at the same time:
```
Wouldn't this make it more clear?

henriquealbert (approved):

packages/ddp-client/common/livedata_connection.js
denihs: Implementing this on this [PR](https://github.com/meteor/meteor/pull/12275).

PR #12066, opened 2022-06-24
Validates required Node.js version

fredmaiaarantes:
Today, Meteor only works with **Node.js 14.x**. People often have problems when trying to install Meteor using a higher version. Therefore, this PR adds a verification code that compares the installed version of NodeJs with the required version. It also updates NPM dependencies. After Fiber removal, we must remove/change this validation.


denihs (approved):

henriquealbert (approved):
LGTM

zodern:
Why doesn't the installer support Node 16 or newer? When I created the initial version of the npm installer (before Node 16 was published), one of my goals was to support a wide range of Node versions to make it easier for users. Meteor doesn't care what version of Node is installed globally, so I thought the installer should try to follow that same idea as much as can be reasonably done.

npm-packages/meteor-installer/package.json
```diff
@@ -9,16 +9,19 @@
...
-    "tmp": "^0.1.0"
+    "semver": "^7.3.7",
+    "tar": "^6.1.11",
+    "tmp": "^0.2.1"
   },
   "bin": {
     "meteor-installer": "cli.js"
+  },
+  "engines": {
+    "node": "<=14.19.3"
```
zodern, line 25: Would using something like `<15` work? That way it would include any new 14.x releases.

fredmaiaarantes:
The installer works, but Meteor itself does not. Maybe, we should add this validation in Meteor tools instead of in the installer?

zodern:
> The installer works, but Meteor itself does not. Maybe, we should add this validation in Meteor tools instead of in the installer?

Meteor comes with its own copy of Node so it should always use the correct version. The only place where that wouldn't be true is when running the bundle from a meteor app created with `meteor build`, and for that there is a check for the node version at https://github.com/meteor/meteor/blob/966698a93f56c72bd67b46f315ec2a599a40dd5c/tools/static-assets/server/boot.js#L13-L21.

fredmaiaarantes:
Currently, due to Fibers, Meteor doesn't work with Node 16.x. So maybe the place you pointed out should have its validation changed to accept only Node.js 14.x? It would avoid people trying to use Node.js >=16 and getting unexpected errors.

henriquealbert:
At least when I tried installing Meteor on my M1, I installed it using `npm install -g meteor` and it installed with node 16 but it doesn't work. Meteor also doesn't install node if you install it with the npm command. That's why @fredmaiaarantes is doing these verifications before letting people install it

If you install meteor using the curl command, it installs the correct version by default

zodern:
I just reinstalled Meteor on my m1 mac with Node 16 and version 2.7.4 of the npm installer, and it mostly worked. Meteor still has its own copy of node:

```
$ node -v
v16.15.1
$ meteor node -v                             
v14.19.3
```

Some issues I ran into:
- It incorrectly thought I was not installing Meteor globally (either due to changes in npm or due to the node version manager I am using), so I had to set the env var `npm_config_global='true'` to get it to install.
- Creating an app or running `meteor npm install` for a 2.7.4 app fails due to the app's package.json file missing a `versions` field. I hadn't tried creating an app with Meteor 2.7.4 before, but I can still reproduce if I change the global node version to 14.x

Running Meteor 2.7.4 and Meteor 2.6.1 apps worked correctly.

rj-david:
Isn't the problem more about npm 8 than node 16 (which comes with npm 8 by default)

npm-packages/meteor-installer/package.json
fredmaiaarantes: Now I'm using "node": "<=14.x", and "npm": "<=6.x". This is just a warning, and it will not generate an error.

fredmaiaarantes:
I changed the code to generate just a warning instead of an error.
It will check if "node": "<=14.x", and "npm": "<=6.x".

PR #12081, opened 2022-07-12
[Refactoring] modernize `tools/run-updater.js`

afrokick:
- convert Updater to es6 class
- remove run-updater.js from .eslintignore
- fix a small bug

### DIFF
--- a/tools/runners/run-updater.js
+++ b/tools/runners/run-updater.js
@@ -1,60 +1,52 @@
-var Console = require('../console/console.js').Console;
+import { Console } from '../console/console';
 
-var Updater = function () {
-  var self = this;
-  self.timer = null;
-};
+const CHECK_UPDATE_INTERVAL = 3 * 60 * 60 * 1000; // every 3 hours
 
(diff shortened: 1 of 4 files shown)


tools/runners/run-updater.js
```diff
@@ -1,60 +1,52 @@
...
-  },
+  }
 
-  // Returns immediately. However if an update check is currently
+  // Returns immediately. However, if an update check is currently
   // running it will complete in the background. Idempotent.
-  stop: function () {
-    var self = this;
+  stop() {
+    if (!this.timer) return;
```
afrokick, line 47: There was a bug here.

We want clearInterval only if it exists like this.timer && clearInterval(this.timer)

StorytellerCZ (approved):
Good first step. Next one will be to move it to TypeScript 😉

henriquealbert (approved):

fredmaiaarantes:
> Good first step. Next one will be to move it to TypeScript 😉

Makes sense, gradual steps like these are most welcome!

PR #12082, opened 2022-07-12
Allow setting a custom ddp-rate-limit message per rule

wreiske:
Added new `DDPRateLimiter.setErrorMessageOnRule` function that allows developers to specify custom error messages for each rule instead of being limited to one global error message for every rule. This is nice because it adds some clarity to what rules triggered which errors, allowing for better UX. It also opens the door for i18nable error messages per rule instead of the default English error message. 

## Example with function

```js
const setupGoogleAuthenticatorRule = {
  userId(userId) {
    const user = Meteor.users.findOne(userId);
    return user;
  },
  type: 'method',
  name: 'Users.setupGoogleAuthenticator',
};

// Add the rule, allowing up to 1 google auth setup message every 60 seconds
const ruleId = DDPRateLimiter.addRule(setupGoogleAuthenticatorRule, 1, 60000);
DDPRateLimiter.setErrorMessageOnRule(ruleId, function (data) {
  return `You have reached the maximum number of Google Authenticator attempts. Please try again in ${Math.ceil(data.timeToReset / 1000)} seconds.`;
});
```

chrome_RexK4LUU2C

## Example with string

```js
// Add the rule, allowing up to 1 google auth setup message every 60 seconds
const ruleId = DDPRateLimiter.addRule(setupGoogleAuthenticatorRule, 1, 60000);
DDPRateLimiter.setErrorMessageOnRule(ruleId, 'Example as a single string error message');
```
image


packages/ddp-rate-limiter/ddp-rate-limiter.js
```diff
@@ -10,9 +10,25 @@ let errorMessage = (rateLimitResult) => {
     'trying again.';
 };
 
+// Store rule specific error messages.
+const errorMessageByRule = [];
```
afrokick, line 14: `{}` or `Map` more suitable than `[]` for key-value

wreiske:
@StorytellerCZ , thought on this PR? Would love your review / feedback. Thanks!

StorytellerCZ (changes requested):
The packages should have feature bump and we need a changelog entry, other than that LGTM.

wreiske:
> The packages should have feature bump and we need a changelog entry, other than that LGTM.

@StorytellerCZ , would you like for me to do this? Let me know and I can dive in....

PR #12072, opened 2022-07-06
Refactoring/Remove unused imports from `tools` folder

afrokick:
Some cleanup before migration to TypeScript)

PR #12084, opened 2022-07-13
Refactoring/Remove unused imports from `tools` folder

denihs:
This PR was originally created by @afrokick and merged in the branch release-2.7.4 https://github.com/meteor/meteor/pull/12072.

But internally, we decided that it would be best to test this properly in a new beta version, and the primary focus right now should be just the MongoDB changes in version 2.7.4. 

So we're reopening this PR, and we'll merge it on version 2.7.5.


afrokick:
Is it more convenient for you to create PR to develop instead of release branch for such refactoring?

denihs:
It would be ideal, but in this case is not necessary because soon, the `devel` branch will have the changes from this release, and the new release branch will contain the same changes, so we should be fine.

Also, sorry for the confusion here. I was hasty in merging the branch. The MongoDB changes are pretty delicate, so it's actually best not to add more things on top of it.

denihs:
As we're creating a new beta version, 2.8, I will merge this now.

PR #12090, opened 2022-07-17
Remove useraccounts from Guide

StorytellerCZ:
This is a proposal to remove the `useraccounts` section from the Guide. This is due to the fact that `useraccounts` packages are no longer maintained and it is causing confusion to people thinking that these packages are official.


henriquealbert (approved):
cc @denihs

Grubba27 (approved):
LGTM!

StorytellerCZ:
I have this as a draft so that there can be discussion on this topic if needed. I will make this into a ready PR in my next round of checking PRs.

Grubba27:
I believe there is no problem in removing this section. As you said in the PR description, it is an unmaintained package.
The only possible downside is the lack of docs if any user is using this package, but I guess the things that we are removing here are also there in their Github repo.

fredmaiaarantes (approved):

PR #12095, opened 2022-07-25
[WIP] - Enable top level await on Meteor (server only)

matheusccastroo:
[WIP] - Enable top level await on Meteor.

This is our current plan (subject to changes): 
- [x] Support TLA with Babel compilation.
  - [x] `babel-preset-meteor` was nested in another empty dir, moved it up to follow the same style as the other packages.
  - [x] Add babel plugin to support TLA in `babel-preset-meteor`.
  - [x] Add plugin on reify babel parser options.
- [X] Adapt build/bundle system
  - [X] Files added with `addFiles` with the `isAsync` option should have an `async closure`: adapt linker.
     - [X] Create new `isAsync` for `addFiles` option
  - [X] When linking, we should make the closures that we create `async` and await for then to resolve when needed: `getHeader`, `_getClosureHeader` and `getClosureFooter` methods.
  - [X] Adapt `VM` on `boot.js` to run with an `async` script, and correctly wait for its evaluation.
    - [ ] Check if the `Profile` would work with how it is now.
  - [X] Adapt `VM` on `runJavaScript` function to work with async script.
     - [ ] Check if the `Profile` would work with how it is now.
  - [X] Adapt `install` so that `require` can work with promises too.
  - [ ] Adapt `reify` so that `module.link` can work with promises too.
- [ ] Improve documentation along the way, when possible.
  - [x]  Include `TOOL_NODE_FLAGS` as a tip on `DEVELOPMENT.md`.
- [ ] Find what may have broke and fix it. 

At the current stage we are able to support top level await on meteor packages and on user code, but only at the server.
It's important to note that for now, the top level await on user code is still swapped to `Promise.await`. This will change after we remove Fibers.

One point that we need to think is about top level await support on the client. It would be basically following the same steps as here, but not scoping it to the server arch only. However, this brings some issues: what about the legacy bundle? Should we still support it? If yes, then I guess our only option would be to pass such files to Babel again for polyfill. I will create a discussion in `Github Discussions` or the forum about this.

Notes:
1. If you want to check this out locally, you will need to generate the `dev bundle` with the current `reify/babel/babel-presets` version.
2. You will probably have issues with cache, so the best thing to do after any change is to `meteor_checkout reset` and then `meteor_checkout`


denihs:
Published versions:

babel-preset-meteor@7.10.1-beta.0
meteor-babel@7.16.0-beta.4
 babel-compiler@7.9.1

PR #12100, opened 2022-07-27
Feature/fibers check enabled by flag

edimarlnx:
Create method to check if Fibers is enabled by flag `DISABLE_FIBERS`.

### DIFF
--- a/packages/meteor/helpers.js
+++ b/packages/meteor/helpers.js
@@ -171,3 +171,5 @@ function logErr(err) {
     );
   }
 }
+
+Meteor.isFibersEnabled = global.isFibersEnabled;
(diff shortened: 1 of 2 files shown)


radekmie:
What is this for, actually?

denihs:
Hi @radekmie, 

As you may know, we'll not remove Fibers at once. Instead, we'll have both Fibers and non-fibers code running together.

We will use this EV to know if Fibers is disabled or not and, based on that, take the necessary action.

zodern:
The [plan](https://github.com/meteor/meteor/discussions/11505#discussioncomment-2737839) was to create a new Meteor package, for example `fibers`, and use its existence to know if fibers are enabled for an app. Is using an environment variable temporary until that is implemented or is the plan changing?

denihs:
We wanted to start with a flag because it's simpler to test. So AFAIK, it's temporary, and the way it's done now, it's really simple to change it to the package option later.

denihs (approved):

PR #12101, opened 2022-07-27
Create `Email.sendAsync` method without using Fibers.

edimarlnx:
Create `Email.sendAsync` method without using Fibers.

- [x] Create `sendAsync` method to send email asynchronous
- [x] Create asynchronous tests with all current synchronous tests
- [x] Create doc for new `sendAsync` method


eduwr (approved):

packages/email/email.js
```diff
@@ -227,6 +231,7 @@ Email.customTransport = undefined;
  * @param {Object[]} [options.attachments] Array of attachment objects, as
  * described in the [nodemailer documentation](https://nodemailer.com/message/attachments/).
  * @param {MailComposer} [options.mailComposer] A [MailComposer](https://nodemailer.com/extras/mailcomposer/#e-mail-message-fields)
+ * @deprecated in 2.8
```
StorytellerCZ, line 234: I don't think the deprecation notice here makes sense, since eventually all the async functions will be merged into the regular ones.

StorytellerCZ (changes requested):

packages/email/email.js
radekmie: I thought we want to go with flow like this:
1. Only sync.
2. Both sync and async (with a suffix).
3. Only async (with a suffix).
4. Only async (without a suffix).

If so, then I'd mark them as deprecated. If not, then the flow looks like this:
1. Only sync.
2. Both sync and async (with a suffix).
3. Only async (without a suffix).

And we don't need a deprecation.

edimarlnx: The correct strategy is:

1. Only sync. (Current)
2. Both sync and async (with a suffix).
3. Only async (without a suffix).

Thank you! @StorytellerCZ and @radekmie

packages/email/email.js
```diff
@@ -227,26 +225,67 @@ Email.customTransport = undefined;
...
+  if (!send) {
+    return;
+  }
 
-  const customTransport = Email.customTransport;
-  if (customTransport) {
+  if (Email.customTransport) {
     const packageSettings = Meteor.settings.packages?.email || {};
-    customTransport({ packageSettings, ...options });
+    Email.customTransport({ packageSettings, ...email });
```
radekmie, line 288: Isn't `Email.customTransport` expected to be async?

packages/email/email.js
```diff
@@ -263,8 +302,10 @@ Email.send = function(options) {
...
   if (mailUrlEnv || mailUrlSettings) {
     const transport = getTransport();
-    smtpSend(transport, options);
+    smtpSend(transport, email);
     return;
   }
-  devModeSend(options);
+  return devModeSendAsync(email, stream).catch((err) => {
+    throw err;
+  });
```
radekmie, line 310:
```suggestion
  return devModeSendAsync(email, stream);
```

packages/email/email.js
```diff
@@ -138,40 +137,39 @@ const getTransport = function() {
...
+// Testing hooks
+EmailTest.resetNextDevModeMailId = function () {
+  nextDevModeMailId = 0;
 };
 
-const devModeSend = function(mail) {
-  let devModeMailId = nextDevModeMailId++;
-
-  const stream = output_stream;
+const devModeSendAsync = async function (mail, stream) {
```
radekmie, line 150: There's no point in making this one `async` if it returns a `Promise` directly.

edimarlnx: You're right. Thanks!

packages/email/email.js
edimarlnx: Good question!
Changed to return value if use the custom transport.

packages/email/email_tests.js
```diff
@@ -1,329 +1,250 @@
...
-    test.matches(contents2, /^====== BEGIN MAIL #1 ======$/m);
-    test.matches(contents2, /^From: qux@example.com$/m);
-    test.matches(contents2, /^To: baz@example.com$/m);
-
+Tinytest.add('[Sync] email - hooks stop the sending', function (test) {
+  // Register hooks
+  const hook1 = Email.hookSend((options) => {
+    // Test that we get options through
+    test.equal(options.from, 'foo@example.com');
+    console.log('EXECUTE');
```
eduwr, line 81: Unnecessary log on test

docs/source/api/email.md
```diff
@@ -83,6 +83,27 @@ Meteor.call(
...
+Meteor.methods({
+  sendEmail(to, from, subject, text) {
+    // Make sure that all arguments are strings.
+    check([to, from, subject, text], [String]);
+
+    // Let other method calls from the same client start running, without
+    // waiting for the email sending to complete.
+    this.unblock();
+
+    Email.sendAsync({ to, from, subject, text }).catch(err => {
```
radekmie, line 101: It'd be good to either add `return` or `await` here, to make sure the method returns only when this call is finished.

packages/email/email.js
```diff
@@ -227,27 +225,66 @@ Email.customTransport = undefined;
...
+    return;
+  }
 
-  const customTransport = Email.customTransport;
-  if (customTransport) {
+  if (Email.customTransport) {
     const packageSettings = Meteor.settings.packages?.email || {};
-    customTransport({ packageSettings, ...options });
-    return;
+    return Email.customTransport({ packageSettings, ...email });
```
radekmie, line 287: It's important to document, that if `Email.customTransport` returns a `Promise`, it'll be part of the `sendAsync` (and thus `send`) flow, i.e., if it takes a long time to process, it'll be included as well (previously it was called but not awaited).

packages/email/email.js
```diff
@@ -227,27 +225,66 @@ Email.customTransport = undefined;
...
+ * @param {Object[]} [options.attachments] Array of attachment objects, as
+ * described in the [nodemailer documentation](https://nodemailer.com/message/attachments/).
+ * @param {MailComposer} [options.mailComposer] A [MailComposer](https://nodemailer.com/extras/mailcomposer/#e-mail-message-fields)
+ * @param {Object} [options.stream] Output stream to write email on development environment
+ * object representing the message to be sent.  Overrides all other options.
+ * You can create a `MailComposer` object via
+ * `new EmailInternals.NpmModules.mailcomposer.module`.
+ */
+Email.sendAsync = async function (options) {
+  const { stream = process.stdout, ...rest } = options;
```
StorytellerCZ, line 273: This destructioning doesn't make much sense to me. It should either get `mailComposer` for use on the line bellow or it should be `const stream = options.stream || process.stdout` or `const { stream = process.stdout } = options;`
So that on the next line you just use the full options object (unless that is an issue).
Also stream should be defined just before it is used on line 306 so that you don't needlessly define it here when the function can be returned before it is even used.

StorytellerCZ (changes requested):

packages/email/email.js
filipenevola: we are not creating any unnecessary resource, declaring in the top makes it easy to see how it is going to define if no stream property is provided.

filipenevola (approved):
LGTM

packages/email/email.js
edimarlnx: I changed and removed it from this method, but I preserved it on the devModeSendAsync method because it's required in async tests to work.

packages/email/email.js
edimarlnx: I changed it in the send synchronous method to preserve the current behavior.
Please see the changes here https://github.com/meteor/meteor/pull/12101/files#diff-81b3d1c2cbdddcb39f637030fdb2c2fe5e3190c2b3e0b638c68199f793aa7e5aR233. Is it make sense for you?

radekmie: My point was different: because `Email.sendAsync` now does `return Email.customTransport(...)`, doing `await Email.sendAsync` will not only wait for the code of `Email.sendAsync` itself, but also the `Email.customTransport`, due to promise chaining. I think it's a good change, but it must be documented.

Besides that, you can't just return `Email.sendAsync`, as it'll return a promise (it was `undefined` before). And because this promise is not `.catch`ed anywhere, it'll raise an unhandled promise rejection on error.

PR #12102, opened 2022-07-27
Use async on `Meteor.startup`

edimarlnx:
Use async on `Meteor.startup`

### DIFF
--- a/packages/hot-module-replacement/client.js
+++ b/packages/hot-module-replacement/client.js
@@ -50,7 +50,7 @@ let useHotCodePush = false;
 let forceReload = function () {
   useHotCodePush = true;
   // Wait until Reload package has been loaded
-  Meteor.startup(function () {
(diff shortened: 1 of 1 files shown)


packages/hot-module-replacement/client.js
```diff
@@ -50,7 +50,7 @@ let useHotCodePush = false;
 let forceReload = function () {
   useHotCodePush = true;
   // Wait until Reload package has been loaded
-  Meteor.startup(function () {
+  Meteor.startup(async function () {
```
zodern, line 53: This file isn't compiled by ecmascript, so it can only use syntax that is supported by legacy browsers.

edimarlnx: We will don't make these changes to the client side.
So I will close this PR.

Thank you for your review.

PR #12103, opened 2022-07-27
Update package facts-base publish method to async

eduwr:
Facts base package uses publish method that will be async.


packages/facts-base/facts_base_server.js
```diff
@@ -56,7 +56,7 @@ Facts.resetServerFacts = function () {
 // called?
 Meteor.defer(function () {
   // XXX Also publish facts-by-package.
-  Meteor.publish(FACTS_PUBLICATION, function () {
+  Meteor.publish(FACTS_PUBLICATION, async  function () {
```
radekmie, line 59: It'd be nice to document the fact that publications can be async - it's only in the changelog now.

PR #12104, opened 2022-07-27
Convert startup method on launch screen package to async

eduwr:
Launch screen package has a startup method that is going to be async.

### DIFF
--- a/packages/launch-screen/default-behavior.js
+++ b/packages/launch-screen/default-behavior.js
@@ -6,7 +6,7 @@ var handle = LaunchScreen.hold();
 
 var Template = Package.templating && Package.templating.Template;
 
-Meteor.startup(function () {
+Meteor.startup(async function () {
   if (! Template) {
     handle.release();
   } else if (Package['iron:router']) {


packages/launch-screen/default-behavior.js
```diff
@@ -6,7 +6,7 @@ var handle = LaunchScreen.hold();
 
 var Template = Package.templating && Package.templating.Template;
 
-Meteor.startup(function () {
+Meteor.startup(async function () {
```
radekmie, line 9: Is this a safe change? I mean, `Meteor.startup` doesn't handle async functions in any way, i.e., it runs them and forgets about the results. Making all `startup` calls async will potentially change the ordering, which may impact both the behavior and the performance. Also, as it's not waiting for the results, it'll make `Call Meteor.startup hooks` profiling section useless.

edimarlnx: We will don't make these changes to the client side.
So I will close this PR.

About call async in `Meteor.startup`: 
Yes, the order will change, but we will work on it. First on the server side and then on the client side.

Thank you for your review.

PR #12105, opened 2022-07-27
Create async method `CssTools.minifyCssAsync`

edimarlnx:
Create async method `CssTools.minifyCssAsync` to replace method `CssTools.minifyCss` in future


packages/minifier-css/minifier.js
```diff
@@ -63,25 +63,35 @@ const CssTools = {
...
+    const f = new Future();
+    CssTools.minifyCssAsync(cssText)
+      .then((res) => f.return(res))
+      .catch((error) => f.throw(error));
     // Since this function has always returned an array, we'll wrap the
     // minified css string in an array before returning, even though we're
     // only ever returning one minified css string in that array (maintaining
     // backwards compatibility).
-    return [minifiedCss];
+    return f.wait();
```
radekmie, line 77: @zodern Can we use `Promise.await` in here (i.e., build plugins)? It'd make the code much cleaner.

zodern: Yes, we can.

edimarlnx: I did it.

Thank you for the suggestion!

packages/minifier-css/minifier.js
```diff
@@ -65,23 +64,24 @@ const CssTools = {
...
+   */
+  async minifyCssAsync(cssText) {
+    return postcss([cssnano({ safe: true })])
+      .process(cssText, {
+        from: void 0,
+      })
+      .then((result) => [result.css])
+      .catch((error) => {
+        throw new Error(error);
+      });
```
radekmie, line 84: This could also be simplified with `Promise.await`.

packages/minifier-css/minifier-async-tests.js
```diff
@@ -0,0 +1,51 @@
...
+
+Tinytest.addAsync(
+  '[Async] minifier-css - simple CSS minification',
+  (test, onComplete) => {
+    const promises = TEST_CASES.map(([css, expected, desc]) =>
+      CssTools.minifyCssAsync(css).then((minifiedCss) => {
+        test.equal(minifiedCss[0], expected, desc);
+      })
+    );
+    Promise.all(promises).then(() => onComplete());
```
radekmie, line 49: `addAsync` also works with promises, therefore this test could look like this:
```suggestion
  async (test) => {
    for (const [css, expected, desc] of TEST_CASES) {
      const minifiedCss = await CssTools.minifyCssAsync(css);
      test.equal(minifiedCss[0], expected, desc);
    }
```

It also fixes the problem that right now it doesn't handle rejection correctly.

packages/minifier-css/minifier.js
edimarlnx: But `Promise.await` is a Fibers method. Right?

radekmie: Ah, right :facepalm:

packages/minifier-css/minifier-async-tests.js
edimarlnx: I changed the code to handle rejections but didn't use await because I wanted execution out of order.

radekmie (approved):

PR #12106, opened 2022-07-28
Create async js minifier

eduwr:
The intent of this PR is to create a js minifier without using the wrapAsync.

I also changed Meteor minifier to use the async minifer.


packages/minifier-js/minifier.js
```diff
@@ -47,3 +47,37 @@ export const meteorJsMinify = function (source) {
...
+
+  try {
+    const terserResult = await terserMinify(source, options);
+    const response = {
+      code: terserResult.code,
+      minifier: "terser"
+    };
+    return response;
+  } catch (e) {
+    throw e;
```
radekmie, line 81: There's no need to `try` if you're rethrowing the same error.

packages/minifier-js/minifier.js
```diff
@@ -1,14 +1,14 @@
...
+const terserMinify = async (source, options, callback = () => {}) => {
   terser = terser || Npm.require("terser");
   try {
     const result = await terser.minify(source, options);
     callback(null, result);
     return result;
   } catch (e) {
     callback(e);
-    return e;
+    throw e;
```
radekmie, line 11: That's a rather big change - where is this coming from? If it's only because `meteorJsMinifyAsync` needs that, maybe it'd be easier to inline it instead? Removing the synchronous version will be easier then.

packages/standard-minifier-js/plugin/minify-js.js
```diff
@@ -63,37 +63,43 @@ class MeteorMinifier {
...
+        Plugin.nudge();
       }
       else {
         let minified;
-        try {
-          minified = meteorJsMinify(file.getContentsAsString());
-        }
-        catch (err) {
+        // Async Implementation
+        meteorJsMinifyAsync(file.getContentsAsString()).then(min => {
```
radekmie, line 76: The `forEach` here won't wait for these promises.

packages/minifier-js/minifier.js
eduwr: You're right, actually it was a mistake. The behavior should be similar to the synchronous method, so it should just return the error. But I think the inline approach is better since we are planning to remove it later.

packages/minifier-js/minifier.js
eduwr: Inline approach will solve this.

packages/standard-minifier-js/plugin/minify-js.js
eduwr: yep. now it is being called in order, but if there is a big file in the middle probably will change the order. I'm inverting the logic to deal with the promises after all are completed.

StorytellerCZ:
@eduwr any update on this?

denihs:
This PR doesn't seem to be relevant anymore. I'm going to close it.

PR #12110, opened 2022-08-02
Fix MongoDB async call inside a Method

denihs:
While testing version 2.7.4-beta.2, @fredmaiaarantes encountered this error:

Screenshot from 2022-08-02 13-24-28

This error comes from the package [`allow-deny` ](https://github.com/meteor/meteor/blob/9c76ea63c3f3f8b9f6432dd5bdb607e4282edcd2/packages/allow-deny/allow-deny.js#L163-L167). 

I did some tests inside the package until I saw this [call](https://github.com/meteor/meteor/compare/release-2.7.4...fix/allow-deny-error?expand=1#diff-e145c1d05aed3f6761172f40f6135e46fdf3852a9af912bbab0ba6f771e67d4dL903). 

Inspecting the value of the `this` object, I saw the methods weren't declared directly inside the object, which means some magic was happening in the background, and a copy of the original methods was being called, so it wasn't getting the right validation from the package `allow-deny` (that's what I'm guessing). 

So now I'm calling the methods from `this._collection`.

Inspecting this object, you'll find the values:

```
_collection: {
      find: [Function (anonymous)],
  findOne: [Function (anonymous)],
  insert: [Function (anonymous)],
  update: [Function (anonymous)],
  upsert: [Function (anonymous)],
  remove: [Function (anonymous)],
  _ensureIndex: [Function (anonymous)],
  createIndex: [Function (anonymous)],
  _dropIndex: [Function (anonymous)],
  _createCappedCollection: [Function (anonymous)],
  dropCollection: [Function (anonymous)],
  rawCollection: [Function (anonymous)]
},
```
That's also why I changed `_dropCollection` to `dropCollection`.


Grubba27 (approved):
looks good!!

StorytellerCZ (approved):

radekmie:
I'm not 100% sure, but this _may_ break some of the hooks packages (as in: don't call them for `*Async` methods). It looks like `matb33:collection-hooks` already uses [`_collection`](https://github.com/Meteor-Community-Packages/meteor-collection-hooks/blob/2090adde3d82e883af3b9fac21f6b55a6d96e99b/collection-hooks.js#L66), but with an exception for `upsert`, which may be affected by this change.

zodern:
@denihs there was a discussion about this error on slack at https://meteor-community.slack.com/archives/C01SWPPE81F/p1658764821424219. The problem is async/await isn't supported in method stubs.

filipenevola:
I'm closing this PR as this solution is not appropriate. It was just skipping Meteor collection at all.

PR #12115, opened 2022-08-05
Exit on MongoDB connection error

radekmie:
As suggested by @zodern in https://github.com/meteor/meteor/pull/12097#issuecomment-1199782349, the app should fail on MongoDB connection error. In this PR I added such a check (as well as a test for it).


Grubba27 (approved):
It looks ok.
Don not know why that test on group 9 is failing

Grubba27:
The failed tests do not seem to be related to the PR. I changed the test a little to make it work, but I'm not sure if it is the right move: 
```js
    const sourceMap = e.stack.split(":")[index];
    const r = (sourceMap === "17" || sourceMap === "18");
    selftest.expectEqual(r, true);
```
in ``tools/tests/source-maps.js``

radekmie:
@Grubba27 After a rebase it fails with the same errors like https://github.com/meteor/meteor/pull/12116#issuecomment-1214069103.

fredmaiaarantes:
> The failed tests do not seem to be related to the PR. I would change the test a little to make it work, but I'm not sure if it is the right move:
> 
> ```js
>     const sourceMap = e.stack.split(":")[index];
>     const r = (sourceMap === "17" || sourceMap === "18");
>     selftest.expectEqual(r, true);
> ```
> 
> in `tools/tests/source-maps.js`

The flaky test we fixed was not the one failing here @Grubba27. It was failing on this PR (https://github.com/meteor/meteor/pull/12125), and it was related to the source-maps tests. Now we need to look into these other ones.

PR #12116, opened 2022-08-05
Improved `oplogV2V1Converter` implementation.

radekmie:
As I mentioned in #12107, I find the current implementation of the `oplogV2V1Converter` inefficient but also hard to understand. In this PR, I simplified it and made it more performant by skipping a lot of unnecessary object copies.

Additionally, I adjusted the tests, so it's easier to add more, as well as reduced logs to one per call to make it easier to debug. (The error message that was there before was no longer reachable.)

### DIFF
--- a/packages/mongo/oplog_v2_converter.js
+++ b/packages/mongo/oplog_v2_converter.js
@@ -1,174 +1,122 @@
-// we are mapping the new oplog format on mongo 5
-// to what we know better, $set and $unset format
-// new oplog format ex:
-// {
-//  '$v': 2,
-//  diff: { u: { key1: 2022-01-06T18:23:16.131Z, key2: [ObjectID] } }
-// }
+// Converter of the new MongoDB Oplog format (>=5.0) to the one that Meteor
+// handles well, i.e., `$set` and `$unset`. The new format is completely new,
+// and looks as follows:
+//
+//   { $v: 2, diff: Diff }
+//
+// where `Diff` is a recursive structure:
+//
+//   {
+//     // Nested updates (sometimes also represented with an s-field).
+//     // Example: `{ $set: { 'foo.bar': 1 } }`.
+//     i: { <key>: <value>, ... },
+//
+//     // Top-level updates.
+//     // Example: `{ $set: { foo: { bar: 1 } } }`.
+//     u: { <key>: <value>, ... },
+//
+//     // Unsets.
+//     // Example: `{ $unset: { foo: '' } }`.
+//     d: { <key>: false, ... },
+//
+//     // Array operations.
+//     // Example: `{ $push: { foo: 'bar' } }`.
+//     s<key>: { a: true, u<index>: <value>, ... },
+//     ...
+//
+//     // Nested operations (sometimes also represented in the `i` field).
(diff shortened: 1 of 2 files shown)


packages/mongo/oplog_v2_converter.js
```diff
@@ -1,174 +1,122 @@
...
 
-  // Handle flat keys.
-  if (u) {
-    Object.entries(u).forEach(([key, value]) => {
-      const prefixedKey = `${prefixKey}${key}`;
-      $set[prefixedKey] = value;
-    });
+function convertOplogDiff(oplogEntry, diff, prefix) {
+  if (process.env.OPLOG_CONVERTER_DEBUG) {
+    console.log(`convertOplogDiff(${JSON.stringify(oplogEntry)}, ${JSON.stringify(diff)}, ${JSON.stringify(prefix)})`);
```
Grubba27, line 66: is it okay to have console .log here ?

radekmie: That's just like it was before: https://github.com/meteor/meteor/pull/12116/files#diff-d214660d46c242c90c987f272897584a9c3ea71e1cad282d58b5f43f31152b35L13-L18

packages/mongo/oplog_v2_converter.js
```diff
@@ -1,174 +1,122 @@
...
+}
 
-  // Handle flat keys.
-  if (u) {
-    Object.entries(u).forEach(([key, value]) => {
-      const prefixedKey = `${prefixKey}${key}`;
-      $set[prefixedKey] = value;
-    });
+function convertOplogDiff(oplogEntry, diff, prefix) {
+  if (process.env.OPLOG_CONVERTER_DEBUG) {
```
zodern, line 65: Accessing properties of `process.env` is relatively slow since Node always gets the latest value of the env var. It might be good to store the value in a variable so it only has to be accessed once.

Grubba27 (approved):
Looks good to me!

zodern (approved):

Grubba27:
I've fixed the group 9 flaky test and refactored it. Perhaps rebasing this branch with its target may make the CI green?

radekmie:
@Grubba27 The first couple of MongoDB errors (error code 48) are rather _normal_, as they suggest that there's some other MongoDB instance and/or port is taken (weird, but OK). But then, there's that:

```
  1| Unexpected mongo exit code 23. Restarting.
  1| Unexpected mongo exit code 46. Restarting.
  1| Unexpected mongo exit code 47. Restarting.
  1| Can't start Mongo server.
  1| MongoDB exited due to excess clock skew
```

And that's _super_ weird. Mostly because none of them is [on the list of exit codes](https://www.mongodb.com/docs/v5.0/reference/exit-codes/), but also the clock skew part. I have no idea what's going on there, but I can't imagine how this PR could cause it.

Grubba27:
That is really odd. I would try to see the internals as a last resort to understand what is the context for these errors, but I do agree that the PR does not have any relation with so many bugs :/ 
I've tried looking for some errors in mongo, and the only file that includes errors I found was [this](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml) one but I don't think it would be useful. Guess I could look in some other repo?

PR #12125, opened 2022-08-10
Fixed `MongoConnection._onFailover` hook (fixes #12118).

### LINKED ISSUE #12118: Mongodb client.topology is undefined
When testing the new Meteor version 2.8-beta.0 in one of our internal apps, an error came up on this [line](https://github.com/meteor/meteor/blob/54964c88e7843defd6e9fc25c894df715dcdad73/packages/mongo/mongo_driver.js#L203-L202). `self.client.topology` is undefined.

This app tries to connect to a MongoDB database in version 5.0.10.

I've tried version 2.8-beta.0 in an app that connects to a database in version 3.6, and it worked fine.

@radekmie thinks it looks like we should call `on` on the client itself and use the `topologyDescriptionChanged` event. 

We're going to test this theory soon.

radekmie:
As reported by @denihs in #12118, the internal `topology` object is not always available, and as such, we can no longer rely on it. What is more, the current implementation relies on the `join` event of the aforementioned `topology`, and I couldn't find any documentation regarding it (I suspect it didn't work for a long time now).

Now, I thought that the solution will be to use the `topologyDescriptionChanged` instead, but it wasn't the case. The new implementation, proposed in this PR, uses the `serverDescriptionChanged` event, as it has all of the information we need.

It also allowed me to get rid of the `_primary` field on the `MongoConnection` object, but I'd need to check whether it won't be needed on a replica set with more nodes (~so far, I tested it on a set with one node~).

**EDIT:** ~I can't make the tests run with a proper replica set. Does anyone else have this problem?~ I was missing `MONGO_OPLOG_URL`.

### DIFF
--- a/tools/tests/run.js
+++ b/tools/tests/run.js
@@ -202,7 +202,11 @@ selftest.define("run errors", function () {
   run.match("Can't start Mongo server");
   run.match("MongoDB exited because its port was closed");
   run.match("running in the same project.\n");
-  run.expectEnd();
+  // TODO pr 12125: the problem is that the buff continues with the value:
+  //  2| Browserslist: caniuse-lite is outdated. Please run:
+  //  2|   npx browserslist@latest --update-db
+  //  2|   Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
+  // run.expectEnd();
   run.forbid("Started MongoDB");
   run.expectExit(254);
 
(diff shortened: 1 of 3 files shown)


Grubba27 (approved):
In the projects that I`ve tried this branch everything turned out okay.
LGTM!

tools/tests/run.js
```diff
@@ -202,7 +202,11 @@ selftest.define("run errors", function () {
   run.match("Can't start Mongo server");
   run.match("MongoDB exited because its port was closed");
   run.match("running in the same project.\n");
-  run.expectEnd();
+  // TODO pr 12125: the problem is that the buff continues with the value:
+  //  2| Browserslist: caniuse-lite is outdated. Please run:
+  //  2|   npx browserslist@latest --update-db
+  //  2|   Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
+  // run.expectEnd();
```
zodern, line 209: @denihs you can update the dependencies of the `minifier-css` package to fix this message. I've never done it before, but I think you can by deleting the package's `.npm` folder so Meteor re-installs the npm deps. MDG had decided a few years ago to not disable the browserlist warning so it would remind anyone maintaining Meteor to update it occasionally.

radekmie:
@Grubba27 I've rebased this PR as well, but Travis tests are still failing with tons of `undefined`s. What do you think, can we merge it?

Grubba27:
I think we could! What do you think @fredmaiaarantes?

fredmaiaarantes:
> I think we could! What do you think @fredmaiaarantes?

Yep, the release branch has the same error, we can merge this PR and investigate the issue from there.

PR #12140, opened 2022-08-15
Update standard-minifier-css dependencies

fredmaiaarantes:
Update the `standard-minifier-css` package dependencies.
This update will avoid issues with Browserslist and caniuse-lite.


fredmaiaarantes:
This branch PR should be created from devel, I will create a new one.

PR #12151, opened 2022-08-19
Feat(callAsync): async method option to Meteor.call

Grubba27:
Hello there! I've seen a fair amount of Meteor projects using this ``wrapper`` to make meteor.call async by promising their callback into a Promise. 
My Idea was to add this small change to make the developer experience a little bit better.


pmogollons:
We also use this for our projects. I think its backwards compatible, just requires to update docs.

Grubba27:
If no callback is provided, I wanted to implement this; it would become a promise based on an option to do something like meteor.call().toPromise() but the problem is that you can use Meteor.call without passing any callbacks as shown [here](https://github.com/meteor/meteor/blob/38c0d70eb4d64dbb9342d5fa7f68e990deb1f0c1/packages/ddp-client/common/livedata_connection.js#L541)
so the second best option is to follow what we are doing in Mongo and place an async postfix in the method you can see these methods being tested [here](https://github.com/meteor/meteor/blob/99e93bde4168a4c287fe371908dc02aae2fbf33d/packages/mongo/collection_async_tests.js#L5)

Grubba27:
Tested using checkout all seems okay :)
Captura de Tela 2022-08-25 às 20 02 08

henriquealbert (approved):
It looks good to me, but let's see what the community thinks about it. Maybe you could create a post on the Forums?

packages/ddp-client/common/livedata_connection.js
```diff
@@ -551,6 +551,27 @@ export class Connection {
...
+  /**
+   * @memberOf Meteor
+   * @importFromPackage meteor
+   * @alias Meteor.callAsync
+   * @summary Invokes asynchronously a method passing any number of arguments.
+   * @locus Anywhere
+   * @param {string} name Name of method to invoke
+   * @param args Optional method arguments
+   */
+  async callAsync(name, ...args) {
```
matheusccastroo, line 563: No need to use the `async` keyword here, since you are already returning a promise.

packages/ddp-client/test/livedata_connection_tests.js
```diff
@@ -466,7 +466,7 @@ Tinytest.add('livedata stub - this', function(test) {
 });
 
 if (Meteor.isClient) {
-  Tinytest.add('livedata stub - methods', function(test) {
+  Tinytest.add('livedata stub - methods', async function(test) {
```
matheusccastroo, line 469: Shouldn't we use the `Tinytest.addAsync` here?

Grubba27:
Adjusted @matheusccastroo :)

radekmie:
While I do like having it in the core, in its current form it doesn't bring any value over an in-app (we have it in a few apps) or in-package implementation (e.g., [`deanius:promise`](https://github.com/deanrad/meteor-promise)).

I would see it relevant, if it'd:
1. Introduce `applyAsync`, that'd work like `apply` but with a `Promise`-based API.
2. Make `applyAsync` truly `Promise`-based, i.e., make it Fibers-free.
3. Make `callAsync` use `applyAsync`, just like `call` uses `apply`.
4. Make `apply` use `applyAsync`, by simply `Promise.await` it on the server and `.then` on the client (plus `.catch` to silent an unhandled rejection).

In this way, once everyone would use `callAsync` / `applyAsync`, we'd be free from Fibers on this end.

Grubba27:
You got a great point! @radekmie 
Prob. I Will make it use applyAsync since it is going to be shipped when this pr [here](https://github.com/meteor/meteor/pull/12152/files) is done. I Will make these changes. Thx a lot for the Review :)

Grubba27:
I will close this PR in favor of @denihs [PR](https://github.com/meteor/meteor/pull/12196).

PR #12152, opened 2022-08-19
Fix problem when publishing async methods

edimarlnx:
When publishing async methods with `Meteor.methods`, we need to call with await to preserve execution order. 
This PR is changing the code of method `Meteor.call` to calling internal async implementation if the published method is async, or continue calling sync method if the published method is sync.

Fix problem when publishing async methods.


edimarlnx:
Related to this PR https://github.com/meteor/meteor/pull/12110 and this Slack discussion https://meteor-community.slack.com/archives/C01SWPPE81F/p1658764821424219

filipenevola:
I've tested with Fred's branch from [here](https://github.com/fredmaiaarantes/simpletasks/tree/mongodb-async-api) and it's working now.

I'm wondering how we could make sure the async contexts are being preserved correctly.

@zodern any ideas how we could try to break it in the client?

filipenevola (approved):

Grubba27 (approved):

zodern:
The original problem was sometimes code in a method stub thought it was running outside of a method stub. It seems these changes invert the problem. Code in a method stub will now always know it is in a method stub, but code outside of a method stub could sometimes think it is inside one.

Here are two examples to demonstrate it. It requires an app with `insecure`, `autopublish`, and `tracker`, and you can put this code in a file that is imported on both the client and server:

```js
// set up
var coll1 = new Mongo.Collection('coll1');
var coll2 = new Mongo.Collection('coll2');

if (Meteor.isServer) {
  coll1.remove({});
  coll1.insert({ _id: '1', a: 1, b: 1 });

  coll2.remove({});
}

function sleep(ms) {
  return new Promise(resolve => setTimeout(resolve, ms));
}

// Test 1
Meteor.methods({
  async 'test1'() {
    coll1.update({ _id: '1' }, { $inc: { a: 1 } });

    await sleep(100);
  }
});

Test1 = function () {
  Tracker.autorun(() => {
    const doc = coll1.findOne({ _id: '1' });
    console.log('a', doc.a, 'b', doc.b);
  });

  Meteor.call('test1', () => {
    console.log('method finished - both a and b should be increased by "1"');
  });

  // If coll1.update thinks it is inside of a method stub, the change will be
  // reverted when the client receives the server result for the method 
  coll1.update({ _id: '1' }, { $inc: { b: 1 } });
}

// Test 2
Meteor.methods({
  async create() {
    await sleep(50);
    coll2.remove({});
    coll2.insert({});
    coll2.insert({});
    coll2.insert({});
  },
  async nothing() {
    await sleep(500);
  }
});

Test2 = function () {
  Tracker.autorun(() => {
    let docs = coll2.find({}, { fields: { _id: 1 } }).fetch();
    if (docs.length > 0) {
      console.log('test 2 - coll2 ids', docs);
    }
  });

  Meteor.call('create');
  Meteor.call('nothing', () => {
    console.log('test 2 finished - "test 2 - coll2 ids" should only have logged once');
  });
}
```

The first test will work correctly in released versions of Meteor (only because all side effects are done before the `await`), and currently fails with this PR. You can run it by calling `Test1()` in the browser's console.

When it runs, it:
1) Calls the `test1` method, which increases `a`
2) Does a client side mutation to increase `b`. If this mutation correctly thinks it is outside of a method stub, then this change will persist. If it instead thinks it is inside a method stub, when the client receives the method result from the server the change will be reverted so the method stub's results match what happened on the server.

If you run it with this PR, `a` will be increased, but `b` will never be.

For test 2, this is checking what happens when two method stubs run in parallel, and you can run it by calling `Test2()` in the browser's console. When inserting documents in a method, Meteor ensures that on the server it generates the same document ids that were generated in the stub. However, for this to work correctly, `Collection.insert` must be able to get the correct value for the `DDP._CurrentMethodInvocation` environment variable when running in the stub. If `DDP._CurrentMethodInvocation` has the wrong method stub invocation, then the ids will change when the client receives the method results from the server.

This test will not work in released versions of Meteor since it is an issue specific to async method stubs. If you run it with this PR, you will notice that the id's are logged twice, and the second time they are all different.

-----

I think there are some related issues with the server implementation, though they will be much easier to fix, and a couple other things that will need to be adjusted, but I will wait to see what the plan for this PR is before looking into those.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -568,6 +575,51 @@ export class Connection {
...
+   * @param {Boolean} options.noRetry (Client only) if true, don't send this method again on reload, simply call the callback an error with the error code 'invocation-failed'.
+   * @param {Boolean} options.throwStubExceptions (Client only) If true, exceptions thrown by method stubs will be thrown instead of logged, and the method will not be invoked on the server.
+   * @param {Boolean} options.returnStubValue (Client only) If true then in cases where we would have otherwise discarded the stub's return value and returned undefined, instead we go ahead and return it. Specifically, this is any time other than when (a) we are already inside a stub or (b) we are in Node and no callback was provided. Currently we require this flag to be explicitly passed to reduce the likelihood that stub return values will be confused with server return values; we may improve this in future.
+   * @param {Function} [asyncCallback] Optional callback; same semantics as in [`Meteor.call`](#meteor_call).
+   */
+  async applyAsync(name, args, options, callback) {
```
Grubba27, line 608: If this method is going to be exposed to the client it needs to be added in the array here ``packages/ddp-client/client/client_convenience.js``

Grubba27:
Hey @radekmie!! Correct me if I misunderstood something @denihs; with the new async API, we can lose scope context. This is a real issue, and we have not yet found it. There are proposals like [this](https://github.com/legendecas/proposal-async-context) one that were presented in TC39 on June 4th, 2020. 
We had some hope to solve this using zone.js, but as seen in this [PR](https://github.com/meteor/meteor/pull/12176) it did not work out. So we are documenting where these cases would happen because we do not have async context and them we will be ready to roll 2.8

Issue #10672, opened 2019-08-13
[1.8.1] Async methods are unblocked without specifying this.unblock()


jeffm24:
I'm not sure sure whether or not this is intended behavior, but I've noticed that methods defined as async on the server are not being awaited, which effectively makes them unblock as if you called this.unblock() within them. This does not happen if you use Promise.await() instead of native await. You can see this is the following example:

Server:
```javascript
const asyncTimeout = ms => new Promise(resolve => setTimeout(resolve, ms));

Meteor.methods({
    async unblockedMethod() {
        await asyncTimeout(10000);
        return 'unblockedMethod';
    },
    blockedMethod() {
        Promise.await(asyncTimeout(10000));
        return 'blockedMethod';
    },
    anotherMethod() {
        return 'anotherMethod';
    }
});
```

Client:
```javascript
function testAsyncAwait () {
    Meteor.call('unblockedMethod', console.log);
    Meteor.call('anotherMethod', console.log);
}
function testPromiseAwait() {
    Meteor.call('blockedMethod', console.log);
    Meteor.call('anotherMethod', console.log);
}
```

Running the first function will result in 'anotherMethod' logging immediately, and then 'unblockedMethod' logging in 10 seconds. On the other hand the second function works as you would expect; the Promise.await blocks the second method from running until the first has finished, so you see 'blockedMethod' log before 'anotherMethod'.

Because of the existence of this.unblock(), it seems like unintended behavior to be able to achieve the same result without explicitly calling it. A very easy fix that I've been employing is to override the Meteor.methods function to wrap all method calls in a Promise.await() call. This ensures that defining them as async will not cause them to unblock unless this.unblock() is called. It seems like this would be pretty easy to implement into the method handling code, or at least this behavior should be documented in the official Methods documentation.

zodern:
This will need to be fixed for some methods in core packages to be able to use async/await.

denihs:
I suspected this issue was going to be solved on this [PR](https://github.com/meteor/meteor/pull/12196), and it was. 

Before, this function [maybeAuditArgumentChecks](https://github.com/meteor/meteor/pull/12196/files#diff-173e69ea0353a765b98017d67abc45ec7ce1449178466dc74738324db83f9183L765-R791), was returning a Promise (when the method was declared async) without waiting for the result.

denihs:
We're going to check if it's the case and follow up here later.

nesbtesh:
[...]

PR #12156, opened 2022-08-23
Change Accounts and Oauth to use Async methods.

edimarlnx:
Change Accounts and Oauth to use Async methods.

[Work in Progress]

In this PR, we are starting changes to remove dependencies of Fibers in packages related to Accounts and OAuth. 

In this phase, we will change all codes that depend on Fibers to work synchronously to async methods. 
For example: When waiting for HTTP requests, methods with Meteor.wrapAsync, etc.

Change Oauth packages
- [x] Google
- [x] Github
- [x] Facebook
- [x] Meteor developer
- [x] Password
- [x] Twitter
- [x] Meetup
- [x] Weibo

Test accounts packages
- [x] Google
- [x] Github
- [x] Facebook
- [x] Meteor developer
- [x] Password
- [x] Twitter
- [x] Meetup 
- [x] Weibo 

_* The OAuth packages Meetup and Weibo didn't test because we don't have service accounts for these services._


radekmie (commented):
It's important to document that a few accessible functions (e.g., `Accounts._checkPassword`) are now `async`. I use it in my projects, and I don't think that's uncommon.

Similarly, a guide on how to migrate custom `loginWithX` packages is crucial here (even if it'd say "it'll just work").

zodern:
I think https://github.com/meteor/meteor/issues/10672 will need to be fixed since there will probably be problems if some of these methods are not blocking, especially the login and logout methods.

edimarlnx:
> It's important to document that a few accessible functions (e.g., `Accounts._checkPassword`) are now `async`. I use it in my projects, and I don't think that's uncommon.
I changed it to preserve compatibility mode. 


> Similarly, a guide on how to migrate custom `loginWithX` packages is crucial here (even if it'd say "it'll just work").
I will see it.

Thank you!

packages/facebook-oauth/facebook_server.js
```diff
@@ -1,16 +1,16 @@
 Facebook = {};
 import crypto from 'crypto';
-import { Accounts } from 'meteor/accounts-base';
+import {Accounts} from 'meteor/accounts-base';
```
StorytellerCZ, line 3: This seems to be some linting happening here. I don't think this is the correct one.

StorytellerCZ (changes requested):

perbergland:
There are a bunch of sync mongodb updates left in accounts_server.js (`users.update(...)`), esp for the token handling method like _insertHashedLoginToken (an example of a method we use in our code to implement SSO login).

All these methods need to become async too and use the mongo async methods, preferrably using the established pattern of adding an async version for now. Is that also planned for 2.9 or is that work saved for a 2.9.x release?

packages/google-oauth/google_server.js
```diff
@@ -56,98 +62,113 @@ const getServiceDataFromTokens = tokens => {
...
+  return returnValue;
 };
 
-Accounts.registerLoginHandler(request => {
+Accounts.registerLoginHandler(async (request) => {
   if (request.googleSignIn !== true) {
     return;
   }
-
+  console.log({ request });
```
Grubba27, line 83: Shouldn't this be removed?

packages/google-oauth/google_server.js
```diff
@@ -56,98 +62,113 @@ const getServiceDataFromTokens = tokens => {
...
-
-  return Accounts.updateOrCreateUserFromExternalService("google", {
-    id: request.userId,
-    idToken: request.idToken,
-    accessToken: request.accessToken,
-    email: request.email,
-    picture: request.imageUrl,
-    ...result.serviceData,
-  }, result.options);
+  console.log({ result });
```
Grubba27, line 110: And this one

PR #12157, opened 2022-08-24
WIP - Fibers optional on Server

matheusccastroo:
WIP - Make Fibers optional on server.

Will write a better description when we have something more concrete here.


packages/ddp-server/livedata_server.js
```diff
@@ -330,9 +330,15 @@ var Session = function (server, version, socket, options) {
...
-  }).run();
+  if (Meteor._isFibersEnabled) {
+    Fiber(function() {
+      self.startUniversalSubs();
+    }).run();
+  } else {
+    ASL.run(Meteor._getAslStore, function () {
+      self.startUniversalSubs();
+    });
+  }
```
KoenLav, line 341: A lot of code follows the same pattern, so it may be useful to create a function Meteor._runAsync which recognizes whether it should use Fibers or native Async/Await.

Maybe even better: we could create a wrapper package for the Fibers library (if not already there) and perform the check in there (so Fiber().run would use either Fibers or Async/Await)?

matheusccastroo: @KoenLav Thanks for the suggestions! Yeah, I will try to create the wrapper and if it doesn't work, create a helper like you said. I'm just doing one by one for now to better understand the places that will be affected, since sometimes this requires the whole chain to be asynchronous.

KoenLav:
Looking at the timeline for Node.js v14 (EOL April 2022), is anyone pursuing "Phase 1.1 - Fibers running on Node.js 16 (optional)" in parallel with this PR?

https://github.com/meteor/meteor/discussions/11505

What are the major blockers left in this PR? I was looking at usages of Fiber and Future, and they seem quite limited/scoped.

Fiber seems easy to replace everywhere, but it appears Future is used for two purposes: (1) awaiting a value later in time and (2) simply halting execution of code until something is completed.

Simply halting execution seems easy enough to fix, but waiting for a value (without making the function async) seems difficult? Specifically Meteor.wrapAsync, but in other places as well.

Is there already a plan to address this?

matheusccastroo:
@KoenLav The idea to have Fibers running on node 16 was discarded, you can check more here: https://github.com/meteor/meteor/discussions/11505#discussioncomment-2737839.

On this PR we basically have the app starting already, but many things have a funky behavior. Right now we are working with the tests Meteor already have to ensure everything works (`Mongo, DDP-Server etc`). For example, right now I'm checking the oplog implementation, which stopped working correctly.

Unfortunately replacing Fibers is not always so straightforward, since the code was built upon this, so "everything needs" Fibers to work correctly.  

> Specifically Meteor.wrapAsync, but in other places as well.

As the `wrapAsync` resolved a future when the function callback runs, we swapped it for `promisify`, which creates a promise that resolves when the callback runs (with all the needed bindings etc). The downside is that the whole chain needs to be async, but there is no escape from that.

KoenLav:
Ah thanks for clarifying about Fibers on Node.js 16 (it seemed like quite a challenge indeed).

We're invested in making sure Meteor enables us to meet the LTS deadline of Node.js 14, so if I can offer any help I'd be glad to.

For instance: I've looked into quite some Meteor (and specifically oplog/DDP) internals when I was working on [this PR](https://github.com/meteor/meteor/pull/10478) and I tried to simplify the SynchronousQeue in [this PR](https://github.com/meteor/meteor/pull/12215).

If there are any specific issues which are still challenging I'd be happy to look into them and see if I can offer some suggestions :)

npm-packages/meteor-babel/runtime.js
```diff
@@ -11,19 +11,22 @@ Module.prototype.resolve = function (id) {
 
 require("@meteorjs/reify/lib/runtime").enable(Module.prototype);
 
-require("meteor-promise").makeCompatible(
-  global.Promise = global.Promise ||
-    require("promise/lib/es6-extensions"),
-  require("fibers")
-);
+if (!!!process.env.DISABLE_FIBERS) {
```
denihs, line 14: Only one ! should be enough :)

> if (!process.env.DISABLE_FIBERS) {

denihs (changes requested):

StorytellerCZ:
Is this still relevant?

harryadel:
@StorytellerCZ I know the point of any upcoming 2.x releases is to ease the transition but I believe it doesn't get easier than this. There'd have come a time where users must put their feet down and fix/migrate their application. It's best to focus all of our efforts towards 3.x releases.

PR #12165, opened 2022-08-28
Handled invalid dates in Minimongo Matcher (fixes #12063).

### LINKED ISSUE #12063: When `The Mongo server and the Meteor query disagree on how many documents match your query.`, socket is broken.
Using Meteor 2.7.3 on windows for development and debian/kubernetes for production.

Expected behavior: Do something else: Get an error, or close the connection and reconnect, make the client aware, or let the client as is, like a regular error that can happen on a publish.
Actual behavior: The client can't send anything back to the server.

Minimal reproduction:
https://github.com/chneau/meteor-bug 
```bash
npm i
npm start
```
On the UI you'll see ways to create links via methods or Collection (allowed server side), and a button to trigger the bug.
The bug is basically subscribing to a publish that trigger a specific mongo error.

```ts
const invalidDate = new Date("");
Meteor.publish("links.bugged", () => LinksCollection.find({ $or: [{ createdAt: { $gte: invalidDate } }] }));
```

Which trigger this error:

```bash
The Mongo server and the Meteor query disagree on how many documents match your query. Cursor description:  CursorDescription {
  collectionName: 'links',
  selector: { '$or': [ [Object] ] },
  options: { transform: null }
}
```

Unfortunately, I was not able to reproduce the exact same error I have on production:
> When the error happens from any client, all clients loose the "real time" part.
They can still call methods to mutate data, but they need to refresh the page to actually see the changes.
This was fixed by ensuring the dates passed to the publishes are not invalid/null.
>> It was looking like the server lost connection with the oplog, and no more real time updates was happening. 

If you try to subscribe to a publish that raise an error:
```ts
Meteor.publish("links.error", () => {
  throw new Error("Error");
});
```
At least the client can still do things, but not with the `Meteor query disagree` one.

radekmie:
As reported in #12063, Minimongo doesn't handle the inequality operators (e.g., `$gt`) on invalid date objects (e.g., `new Date('')`) correctly, resulting in a broken connection. I couldn't figure out why the connection breaks (yet...?), but I implemented handling of the invalid date objects to fix it in the first place.

It's not stated in the MongoDB docs, but by checking the `explain` result, it looks like invalid dates are handled just like `new Date(0)`, and that's what I did. Here's how I checked it:
```ts
// db.example.find({createdAt: {$gt: new Date('')}}).explain().queryPlanner
{
    "namespace" : "meteor.example",
    "indexFilterSet" : false,
    "parsedQuery" : {
        "createdAt" : {
            "$gt" : ISODate("1970-01-01T00:00:00.000Z")
        }
    },
    "queryHash" : "77E9ED6B",
    "planCacheKey" : "0CF3FA65",
    "maxIndexedOrSolutionsReached" : false,
    "maxIndexedAndSolutionsReached" : false,
    "maxScansToExplodeReached" : false,
    "winningPlan" : {
        "stage" : "COLLSCAN",
        "filter" : {
            "createdAt" : {
                "$gt" : ISODate("1970-01-01T00:00:00.000Z")
            }
        },
        "direction" : "forward"
    },
    "rejectedPlans" : []
}
```

### DIFF
--- a/packages/minimongo/matcher.js
+++ b/packages/minimongo/matcher.js
@@ -262,8 +262,8 @@ LocalCollection._f = {
     if (ta === 9) { // Date
       // Convert to millis.
       ta = tb = 1;
-      a = a.getTime();
-      b = b.getTime();
+      a = isNaN(a) ? 0 : a.getTime();
+      b = isNaN(b) ? 0 : b.getTime();
     }
 
     if (ta === 1) { // double
--- a/packages/minimongo/minimongo_tests_client.js
+++ b/packages/minimongo/minimongo_tests_client.js
@@ -439,8 +439,16 @@ Tinytest.add('minimongo - selector_compiler', test => {
   // dates
   const date1 = new Date;
   const date2 = new Date(date1.getTime() + 1000);
+  const date3 = new Date('');
   match({a: date1}, {a: date1});
   nomatch({a: date1}, {a: date2});
+  match({a: date3}, {a: date3});
+  nomatch({a: date1}, {a: date3});
+  nomatch({a: date3}, {a: date1});
+  match({a: {$gt: date3}}, {a: date1});
+  match({a: {$gte: date3}}, {a: date1});
+  nomatch({a: {$lt: date3}}, {a: date1});
+  nomatch({a: {$lte: date3}}, {a: date1});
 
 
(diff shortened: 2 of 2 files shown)


Grubba27 (approved):
Sounds good

Issue #12022, opened 2022-04-28
Bad naming for setPublicationStrategy variable (documentation fix probably)


kolyasya:
Hello, yesterday I was using publication strategies for a first time and was confused with:
[getPublicationStrategy(publicationName)](https://docs.meteor.com/api/pubsub.html#getPublicationStrategy)

Obviously `publicationName` should be a `collectionName` which we use when publish documents on a client. At least it was the only way to make it working :)

Do I think about it wrong or should the documentation be fixed?

<img width="839" alt="image" src="https://user-images.githubusercontent.com/3371143/165687614-1313d5d2-59bb-42cf-bf92-541836f5c898.png">

hexsprite:
Maybe it's unclear but the name of a publication is distinct and different from the name of a collection. There's no 1 to 1 mapping. I can have a `users` collection and publish the data as `loggedInUsers` based on a query from that collection. Or a publication could be "virtual" and have no collection backing it....

kolyasya:
> Maybe it's unclear but the name of a publication is distinct and different from the name of a collection. There's no 1 to 1 mapping. I can have a `users` collection and publish the data as `loggedInUsers` based on a query from that collection. Or a publication could be "virtual" and have no collection backing it. Yes, you can publish to a publication with the same name as the collection but that's not always the case hence the distinction.

I'm not sure if it was answer to the issue or addition :-)

I understand that publication name is different from a collection name and that is why `publicationName` variable confuses. Because there should be a `collectionName` which we'll publish into.

kolyasya:
Just checked the source code to make sure, even there it is used with `collectionName`

<img width="646" alt="image" src="https://user-images.githubusercontent.com/3371143/165752248-5501c5f8-9ac9-4266-8d38-61a33ac3414d.png">

hexsprite:
Ahh, well maybe I misunderstood and I will defer to someone else who get this more :)

WilliamKelley:
No, publication name is the correct entity name.

A publication strategy is how changes in a publication are handled and communicated to clients. It is specific to a publication, not a collection.

This should be clear from the [description above](https://docs.meteor.com/api/pubsub.html#Publication-strategies).

radekmie:
Maybe @sebakerckhof could help here?

kolyasya:
Crazy that such simple thing is not yet fixed...

Grubba27:
It was merged and will be correted in the next release :tada:

[...]

PR #12166, opened 2022-08-29
fix: rename setPublicationStrategy and getPublicationStrategy arguments.

fredmaiaarantes:
Rename setPublicationStrategy and getPublicationStrategy arguments.

Fix issue: [Bad naming for setPublicationStrategy variable (documentation fix probably) #12022](https://github.com/meteor/meteor/issues/12022)

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -1517,7 +1517,7 @@ Object.assign(Server.prototype, {
   },
 
   /**
-   * @summary Set publication strategy for the given publication. Publications strategies are available from `DDPServer.publicationStrategies`. You call this method from `Meteor.server`, like `Meteor.server.setPublicationStrategy()`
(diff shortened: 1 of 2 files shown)


packages/ddp-server/livedata_server.js
```diff
@@ -1520,30 +1520,30 @@ Object.assign(Server.prototype, {
    * @summary Set publication strategy for the given publication. Publications strategies are available from `DDPServer.publicationStrategies`. You call this method from `Meteor.server`, like `Meteor.server.setPublicationStrategy()`
```
radekmie, line 1520: Both JSDocs still suggest that it's set per publication, not collection.

radekmie (approved):

packages/ddp-server/livedata_server.js
fredmaiaarantes: Thanks! I just changed them as well.

fredmaiaarantes:
The variable name initially was 'collectionName,' not 'publicationName.' Later, when docs were added, I believe the variable name was changed by mistake.

Grubba27 (approved):
Looks good

PR #12168, opened 2022-08-29
Feat migration guide for 2.8

Grubba27:
Relates to [this topic](https://github.com/meteor/meteor/pull/12057#discussion_r956599272])
Also corrected a typo in changelog

PR #12170, opened 2022-08-31
chore: added how can I use section to change log

Grubba27:
As discussed in this thread #12168 with @radekmie and @StorytellerCZ 
I've added this new section: How can I start using these new features?
Any opinions, @denihs?

### DIFF
--- a/guide/source/2.8-migration.md
+++ b/guide/source/2.8-migration.md
@@ -44,7 +44,7 @@ We got a few examples for making it easy to use these new feature, you can look 
 ```js
 // Before 2.8, we would use something like this
 export const removeByID = ({ id }) => {
-  SomeCollection.remove(id);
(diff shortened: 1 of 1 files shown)


guide/source/2.8-migration.md
```diff
@@ -37,6 +37,33 @@ Here are the newly added methods (you can see this description and the code [her
...
     ```
 
+<h3 id="how-do-i-update">How can I start using these new features?</h3>
+
+We got a few examples for making it easy to use these new feature, you can look the code snippet bellow
+
+```js
+// Before 2.8, we would use something like this
+export const removeByID = ({ id }) => {
+  SomeCollection.remove(id);
```
radekmie, line 47: I'd use `{ _id: id }` here as well to make them even more similar.

radekmie (approved):

PR #12174, opened 2022-09-09
Type definitions for core packages

piotrpospiech:
### Description

This is one of several PRs that move type definitions from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/meteor) into packages. It will enable using `zodern:types` for core packages' types.

It contains types for two packages: `ejson` and `fetch`. Types for other packages will be added later in different pull requests. There are two reasons for that:

1. To test the current solution of using `zodern:types` package on the production.
2. Several type definitions from DefinitelyTyped will go into different repositories for example [react-packages](https://github.com/meteor/react-packages) or [blaze](https://github.com/meteor/blaze). It will be easier to test this solution first and add other packages later than to make changes in several repositories.

### About this approach

Users can test implemented types with [zodern:types](https://github.com/zodern/meteor-types#meteor-apps) package. It needs to be installed on their Meteor project. However, a user has to remove previously used `@types/meteor` from the project because types will be redeclared.

### DIFF
--- a/packages/ejson/package.js
+++ b/packages/ejson/package.js
@@ -5,7 +5,7 @@ Package.describe({
 
 Package.onUse(function onUse(api) {
   api.use(['ecmascript', 'base64']);
-  api.addAssets('ejson.d.ts', ['client', 'server']);
+  api.addAssets('ejson.d.ts', 'server');
   api.mainModule('ejson.js');
   api.export('EJSON');
 });
--- a/packages/ejson/package.js
+++ b/packages/ejson/package.js
@@ -5,6 +5,7 @@ Package.describe({
 
 Package.onUse(function onUse(api) {
   api.use(['ecmascript', 'base64']);
+  api.addAssets('ejson.d.ts', ['client', 'server']);
   api.mainModule('ejson.js');
   api.export('EJSON');
 });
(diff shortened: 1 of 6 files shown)


Grubba27:
Hey, @piotrpospiech, is there anything that we can do to help you with this PR?

piotrpospiech:
> Hey, @piotrpospiech, is there anything that we can do to help you with this PR?

Hey @Grubba27, yes, you can do a code review. After CR it will be ready to merge.

packages/ejson/package.js
```diff
@@ -5,6 +5,7 @@ Package.describe({
 
 Package.onUse(function onUse(api) {
   api.use(['ecmascript', 'base64']);
+  api.addAssets('ejson.d.ts', ['client', 'server']);
```
radekmie, line 8: 1. Do we have to add it as an asset?
2. Do we have to add it as an asset **on the client**?

@piotrpospiech, @zodern.

radekmie (approved):

packages/ejson/package.js
piotrpospiech: 1. Yes, without adding it to the assets, it won't be added to the package-types
2. I tested with just a server. It worked, so I will change it.

PR #12176, opened 2022-09-09
Isolate call context with Zone.js

edimarlnx:


StorytellerCZ:
Why and why this solution?

edimarlnx:
> Why and why this solution?

These changes are a test to isolate the context when asynchronous methods are called. But, it doesn't work because Zone.js know how to deal with async/await.

I am closing this PR.

PR #12175, opened 2022-09-09
The rest of type definitions for core packages

piotrpospiech:
The rest of the type definitions for core packages.

### Migration Guide

1. Remove `@types/meteor` package
2. Install [`zodern:types`](https://github.com/zodern/meteor-types) package
3. Follow [installation guide for the Meteor Apps](https://github.com/zodern/meteor-types#meteor-apps) to update `tsconfig.json`

PR #12180, opened 2022-09-14
Evolved meteor methods

Grubba27:
# RPC Package
Minimal opt-in package for calling meteor methods with e2e type-safety. 

## Demo

PoC is [here](https://github.com/Grubba27/charm-meteor-app)

## Explanation 
As we have seen in this [PR/Discussion](https://github.com/meteor/meteor/pull/11039), Meteor and its users would benefit a lot from a modern implementation of Validated Methods(evolved meteor methods).
 I'm willing to bring back this discussion with a new view and with another implementation. Most of this code and ideas have come from @zodern [zodern:relay](https://packosphere.com/zodern/relay). Also, I've made some design decisions that we should all try and decide on the best approach.

### Key decisions
- Convention over configuration (should work out of the box as intuitive as possible);
- Incremental(It should be easy to move from ``Meteor.call`` to this solution);
- All types defined
- Functional over Class based solution

### Non-decisions(can change)
- Zod (made development easier and is a good practice, I think, but we could be Schema-validator agnostic)

### How to use it?

```js
// consider that we have these methods below:
/**
 * Insert a task for the logged user.
 * @param {{ description: String }}
 * @throws Will throw an error if user is not logged in.
 */
const insertTask = ({ description }) => {
  check(description, String);
  checkLoggedIn();
  TasksCollection.insert({
    description,
    userId: Meteor.userId(),
    createdAt: new Date(),
  });
};
/**
 * Insert a task for the logged user.
 * @param {{ newDescripiton: String , taskId: String}}
 * @throws Will throw an error if user is not logged in.
 */
const updateTask = ({ newDescripiton, taskId }) => {
  check(newDescripiton, String);
  checkLoggedIn();
  TasksCollection.update(
    { _id: taskId },
    { $set: { description: newDescripiton } }
  );
};

/**
 * Check if user is logged in and is the task owner.
 * @param {{ taskId: String }}
 * @throws Will throw an error if user is not logged in or is not the task owner.
 */
const checkTaskOwner = ({ taskId }) => {
  check(taskId, String);
  checkLoggedIn();
  const task = TasksCollection.findOne({
    _id: taskId,
    userId: Meteor.userId(),
  });
  if (!task) {
    throw new Meteor.Error('Error', 'Access denied.');
  }
};

/**
 * Remove a task.
 * @param {{ taskId: String }}
 * @throws Will throw an error if user is not logged in or is not the task owner.
 */
export const removeTask = ({ taskId }) => {
  checkTaskOwner({ taskId });
  TasksCollection.remove(taskId);
};

/**
 * Toggle task as done or not.
 * @param {{ taskId: String }}
 * @throws Will throw an error if user is not logged in or is not the task owner.
 */
const toggleTaskDone = ({ taskId }) => {
  checkTaskOwner({ taskId });
  const task = TasksCollection.findOne(taskId);
  TasksCollection.update({ _id: taskId }, { $set: { done: !task.done } });
};

function publishTasks() {
  return TasksCollection.find({ userId: this.userId });
}

// in the old way:
Meteor.methods({
  insertTask,
  removeTask,
  toggleTaskDone,
});

Meteor.publish('tasksByLoggedUser', function publishTasks() {
  return TasksCollection.find({ userId: this.userId });
});

// in the front end we would call like this:
Meteor.call('insertTask', {description: 'some'})

// or like this:
const isLoading = useSubscribe('tasksByLoggedUser');

// using RPC
const Tasks = createRouter('tasks')
  .addMethod('insert', z.object({ description: z.string() }), insertTask)
  .addMethod('update', z.object({ newDescripiton: z.string(), taskId: z.string() }), updateTask)
  .addMethod('remove', z.object({ taskId: z.string() }), removeTask)
  .addMethod('toggleDone', z.object({ taskId: z.string() }), toggleTaskDone)
  .addPublication('byUser', z.any(), publishTasks)
  .build();

// Tasks schema is like this: {
//   "insert": ({description: string}) => void;
//   "update": ({newDescripiton: string, taskId: string)} => void;
//   "remove": ({taskId: string}) => void;
//   "toggleDone": ({taskId: string}) => void;
//   "byUser": () => Mongo.Cursor;
// }

// in the front end we would call like this:
Tasks.insert({ description }) 
// ˆ?  fully typed 

// Subscriptions can be made like this
 const isLoading = useSubscribe(Tasks.byUser.config.name);

```

### What do we have in this Alfa?
- Minimal opt-in package;
- Fully typed;
- hooks


Grubba27:
https://user-images.githubusercontent.com/70247653/190198357-b851d64d-3a4f-4e17-9de5-6adb0db86322.mov

Some video while doing tests

DavidSichau:
Thanks for your effort to help to make meteor more type-safe. We struggled with this for a while and developed our own approach to make meteor type-safe. We added type-safe meteor methods on top of meteor without any major changes in the underlying foundation. It is optional and can be added incrementally to an application, which allows a gradual adaption. 

I wrote a summary of our approach here: https://docs.expert.ethz.ch/Type-Safe-Meteor-Methods-cfee32a0f69e45aaaa2729ea57653107

# Sugestions or opinions

 - Think about declaration merging, as this would allow moving logic to typescript. 
 - If you make changes to the core it would be easily possible to make `Meteor.call` type-safe, instead of our `callMethod` helper.
 - Instead of using zod or any other library just provide an interface that developers can choose their own schema validation like iots or zod.

radekmie:
I haven't read the code (yet), but here are my two cents:
1. I agree with @DavidSichau, that forcing the `zod` package here is not a good idea. Meteor _almost_ since ever (it's a long time now) suggested using `check` for simple things and SimpleSchema for more complex ones. I don't say that `zod` is not a good modern replacement, but I don't see why it couldn't be more general.
2. I also would rather see `Meteor.methods`/`Meteor.apply`/`Meteor.call` (and their async counterparts) typesafe rather than some solutions built on top of that. It's totally possible with declaration merging, and even though it's slightly more verbose, it's more natural (I think).
    * It also has zero runtime overhead - there are no "artificial method objects" or imports on the client side. Everything disappears in runtime.
3. Instead of adding a new core package, I'd rather release it as a community one and see whether or not people will adopt it. Increasing the API surface adds complexity and adds more packages to take care of, with no clear benefit. Once it's adopted and battle-tested by multiple users, it's totally fine to add it to the core.

And as for the example code: it's not "good" either:
1. The methods are still doing `check` calls, suggesting that their inputs are not validated, which is not true.
2. None of them is correctly typed, as without any type annotations or proper usage of inference, their inputs are all `any`.
    * The code in the video is already better, as it correctly uses inference to get full type safety. However, I do find it extremely hard to watch, as the Copilot (I guess) is constantly suggesting random expressions.
3. The definition of `checkLoggedIn` is missing. It'd be fine, but it's actually crucial - I think it should use the `this.userId` from the method context, but it's not passed to it. (Actually, `this.userId` is not even available there, as top-level arrow functions won't have it.)

Akryum:
Here is what I've been doing (slightly modified to leave out the vue-meteor-tracker parts): https://gist.github.com/Akryum/1a26bf4cc82b9a5cd63035aea1a374e4

Grubba27:
Liked a lot implementation @Akryum! 
This PR/ Discussion idea is to devise a solution for modernizing this side of Meteor.
Most can agree that ``createMethod``, like the one you shared, is seen in many projects. We could provide an abstraction to use these primitives, as @radekmie said.
My key objectives and we can discuss some other ways to implement them:

- _Develop something that both JS and TS users could benefit from._
 My first thought was of something like tRPC has done with types but if you are not using TS you can't use their full potential.

- _Create something/intuitive simple that can create a good Developer Experience out of the box._
 That is where I got the idea to make something like the ``createRouter`` its seems really something that you see in the node world. Using something like declaration merge would make the end user do some configuration, It is an option but I'm considering others before this one. I've made a gist crafting something in this [sense](https://gist.github.com/Grubba27/64020a70268d332d2cb0af7baf3ff5ef).

- _Validator Agnostic._
Because it is easy for development and I think brings good practice, I've chosen Zod, but with a simple callback and generics, I could create some custom validation.

Grubba27:
Hey @radekmie, good to see you here; how are you doing? 
Thanks a lot for this review/feedback. Really appreciate it.
What you have said makes a lot of sense and can be a solution.
You mean having a type API that I can merge my types and my Meteor.call being a key of one of these?
That sounds like a good solution. Here is the [gist](https://gist.github.com/Grubba27/a13043dc91a23060f2f8ae07785ee4c1) that I made while I was thinking about this issue
I think as well that we should be doing things in the core using the same APIs that were battle-tested like the ones from ``mdg:validated-method``

Another solution would be to have some kind of codegen like running something like ``meteor generate types``. And it would generate those types with declaration merging.

-----
Do you have more ideas? I would love to hear!

denihs (approved):
LGTM!

jankapunkt:
I'd like to add one more thing: please do not forget documentation. I know with TS the code is so much more self-explaining but it can sometimes also be much more verbose and confusing.

For example: `createRouter` implies to me, that there are HTTP routes or any other form of client-side routing involved. If this is not the case it should definitely be explained in which context this "Router" is to be used and how it differs from i.e. an express router or a flow-router on the client.

Grubba27:
Hey @jankapunkt good to see you here! 
> I'd like to add one more thing: please do not forget documentation.

I was in the process of making it. We do have one now. Any feedback on it would be awesome 🙏 

> createRouter implies to me, that there are HTTP routes or any other form of client-side routing involved. 

I choose the name because seems intuitive. 
But other option were ``createModule``

Akryum:
I have updated my gist to use `z.tuple` which allows tuples `[1,2] as const` in JS too:
https://gist.github.com/Akryum/1a26bf4cc82b9a5cd63035aea1a374e4

It has:
- Simple signature: `(name, args, handler, options) => call`
- Allows for multiple arguments like `Meteor.call`

Example result in JS:

```js
// @ts-check

import { z } from 'zod'
import { createMethod } from './util2'

export const hello = createMethod('hello', null, () => {
  return 'Hello world!'
})

hello()

export const withParams = createMethod('withParams', z.tuple([z.string(), z.number()]), (name, age) => {
  return `Hello ${name}! You are ${age} years old.`
})

withParams()
withParams('John', 30)
withParams('John', 'Hello')
withParams('John', 30, 2)
```

image

image

Example result in TS:

```ts
import { z } from 'zod'
import { createMethod } from './util2'

export const hello = createMethod('hello', null, () => {
  return 'Hello world!'
})

hello()

export const withParams = createMethod('withParams', z.tuple([z.string(), z.number()]), (name, age) => {
  return `Hello ${name}! You are ${age} years old.`
})

withParams()
withParams('John', 30)
withParams('John', 'Hello')
withParams('John', 30, 2)
```

image

Akryum:
```js
const Tasks = createRouter('tasks')
  .addMethod('insert', z.object({ description: z.string() }), insertTask)
  .addMethod('update', z.object({ newDescripiton: z.string(), taskId: z.string() }), updateTask)
  .addMethod('remove', z.object({ taskId: z.string() }), removeTask)
  .addMethod('toggleDone', z.object({ taskId: z.string() }), toggleTaskDone)
  .addPublication('byUser', z.any(), publishTasks)
  .build()
```

I find this chain of functions to be confusing. Also it doesn't allow exporting each method individually easily from the look of it.
IMO importing methods from a module is already enough, there is no need to add complexity (and more JS shipped to the user):

```js
export const insert  = createMethod(...)
export const update  = createMethod(...)
```

```js
import * as Tasks from '/imports/api/tasks-methods'

Tasks.insert(...)
```

Grubba27:
> I have updated my gist to use z.tuple which allows tuples [1,2] as const in JS too:
https://gist.github.com/Akryum/1a26bf4cc82b9a5cd63035aea1a374e4

I liked a lot your implementation! 

> The name adds to the confusion ("Am I creating an URL router?") 😹

Okay, I'm on my way to changing this API 😆.
 I will remove this one. I think the ESM way is *way* better too

Given these comments, I think we are getting somewhere. I'm having some doubts about how I could implement multiple-schema validators.

Having support for Zod, Yup, and Simpl is a must, and also have the choice of just using it as a callback.
the last is a little bit awkward like you have two callbacks but that can be an option.

packages/rpc/lib/createPublication.ts
```diff
@@ -0,0 +1,36 @@
+import {Config, Subscription, SubscriptionCallbacks} from "./types";
+import {z} from "zod";
+import {isThenable} from "./utils/isThenable";
+import {RateLimiterConfig} from "./utils/RateLimiterConfig";
+
+export const createPublication: Subscription =
+  <Name extends string, S extends z.ZodUndefined | z.ZodTypeAny, T>
+  (name: Name, schema: S, run: (this: Meteor.MethodThisType, args: z.output<S>) => T, config?: Config<S, T>) => {
```
Akryum, line 8: `this: Meteor.MethodThisType` is wrong, should be `Subscription`

Akryum:
Updated the gist with subscriptions that I just implemented in my app:
https://gist.github.com/Akryum/1a26bf4cc82b9a5cd63035aea1a374e4

Automatic inference <3

image

image

radekmie:
@Grubba27
> You mean having a type API that I can merge my types and my Meteor.call being a key of one of these?
> That sounds like a good solution. Here is the [gist](https://gist.github.com/Grubba27/a13043dc91a23060f2f8ae07785ee4c1) that I made while I was thinking about this issue
>
> Another solution would be to have some kind of codegen like running something like meteor generate types. And it would generate those types with declaration merging.

No, I meant something completely different. Here it is:
```tsx
declare module 'meteor/meteor' {
  namespace Meteor {
    // The interface we'll "merge" method definitions to. It's initially empty,
    // and will grow with each added method.
    interface MethodsMap {}

    // Helper types.
    type Methods = keyof MethodsMap;
    type MethodParams<Method extends Methods> = Parameters<MethodsMap[Method]>;
    type MethodResult<Method extends Methods> = ReturnType<MethodsMap[Method]>;

    // I've added a "*Typed" suffix here just to make it work with the current
    // typings. Ideally, it wouldn't be here. The idea is that every time we
    // define a new method or call it, its type has to be already "registered",
    // i.e., added to the `MethodsMap` interface.
    //
    // I've omitted `apply` for brevity, but it's totally fine to add it as well.
    function methodsTyped<MethodList extends Methods>(
      methods: {
        [Method in MethodList]: Method extends never
          ? never
          : (
              this: MethodThisType,
              ...args: MethodParams<Method>
            ) => MethodResult<Method>;
      },
    ): void;
    function callTyped<Method extends Methods>(
      name: Method,
      // TODO: Add optional callback. (It won't be a problem with async API.)
      ...args: MethodParams<Method>
    ): MethodResult<Method>;
  }
}

// This piece of code can be added to any file in the project, including `.d.ts`
// ones, and it relies on interface declaration merging. That means there are
// no runtime implications whatsoever.
//
// Additionally, these can be generated from the source code by inferring the
// types of the `Meteor.methodsTyped` calls, or via some other logic.
//
// Actually, it's possible to use an API similar to the one in this PR, i.e.,
// registering a method with a name and schema, which then would be used to
// generate standard `Meteor.methods`/`Meteor.call` calls along with their type
// definitions.
//
// Without code generation, you can think of it in the same way how GraphQL
// schemas work - we can define them upfront and then let the backend implement
// them while the front end already knows what will be there.
declare module 'meteor/meteor' {
  namespace Meteor {
    interface MethodsMap {
      example: (foo: number, bar: string) => string;
    }
  }
}

Meteor.methodsTyped({
  // Both `foo` and `bar` are typed, and so is the return type. Keep in mind,
  // that it's still possible to call it without checking the type, e.g.,
  // directly from the browser.
  example(foo, bar) {
    return `foo = ${foo.toFixed(2)}, bar = ${bar}`;
  },

  // This does not type, as it's not registered in the `MethodsMap`.
  thisMethodDoesNotExist() {},
});

// Typed as `string`. Of course, we'd have to change it to an async API or type
// the callbacks, as it's not possible to type the client and server code
// differently (without a lot of overhead).
Meteor.callTyped('example', 10.24, 'something');
```

_(And thanks for asking! I'm better now, but still off for probably two weeks.)_

radekmie:
And just to make it clear - I'm totally up for some package like the one described here, with or without suggestions from all of the people in the comments (although, making it `zod`-only would be a missed opportunity, I think). I'm just _rather against_ adding it to the core just now.

Grubba27:
@radekmie, I really liked how simple yet functional your implementation is!!

> (although, making it zod-only would be a missed opportunity, I think)

Yeah, so my thoughts, given the comments I've received on this, are to be at least Simpl-schema open. At least on the alfa version is to have support on these two and later on or the community can help bring more schema validators.

for the implementation: 
```ts

// quick implementation
 if (schema.constructor.name === "SimpleSchema") {
// user is using SimpleSchema
}

 if (/* Zod validation */) {
// user is using Zod
}
```

> I'm just rather against adding it to the core just now.

I would ask why this opinion. 
Surely this package could start as a community package and then be moved to the core. This is what you mean? 

At the same time, I think a more native solution is really neat. The user can choose how and if he wants to validate data or just have a prototype.
I'm only thinking about how this could affect the JavaScript-only projects;

Grubba27:
Gave some more tests into it and I'm kind of liking the results:'

```ts

declare module 'meteor/meteor' {
  namespace Meteor {

    interface PublicationsMap {
    }

    type Publications = keyof PublicationsMap;
    type PublicationParams<Publication extends Publications> = Parameters<PublicationsMap[Publication]>;
    type PublicationResult<Publication extends Publications> = ReturnType<PublicationsMap[Publication]>;

    function publish<Publication extends Publications>(
      name: Publication,
      publication: (this: Subscription, ...args: PublicationParams<Publication>) => PublicationResult<Publication>,
      options?: { is_auto: boolean }
    ): void;

    function subscribe<Publication extends Publications>(
      name: Publication,
      args?: PublicationParams<Publication>,
      callback?: (error: Error | PublicationResult<Publication>) => void,
    ): Meteor.SubscriptionHandle;
    function subscribe<Publication extends Publications>(
      name: Publication,
      callback: (error: Error | PublicationResult<Publication>) => void,
    ): Meteor.SubscriptionHandle;

    interface MethodsMap {
    }

    type Methods = keyof MethodsMap;
    type MethodParams<Method extends Methods> = Parameters<MethodsMap[Method]>;
    type MethodResult<Method extends Methods> = ReturnType<MethodsMap[Method]>;

    function methods<MethodList extends Methods>(
      methods: {
        [Method in MethodList]: Method extends never
          ? never
          : (
            this: Meteor.MethodThisType,
            ...args: MethodParams<Method>
          ) => MethodResult<Method>;
      },
    ): void;

    function call<Method extends Methods>(
      name: Method,
      ...args: MethodParams<Method>
    ): MethodResult<Method>;
    function call<Method extends Methods>(
      name: Method,
      args: MethodParams<Method>,
      callback: (error: Error | undefined, result: MethodResult<Method>) => void,
    ): void;

  }
}

```

@Akryum @radekmie, any thoughts on this? 
I was thinking something along this lines:
Still make a new package just with these implementations
like, RPC.methods / RPC.call just being reference to Meteor.call / Meteor.methods but being typed: 

```ts
const RPC = {
  call:  Meteor.call,
  methods: Meteor.methods,
};
```
Or add that typing to this [PR](https://github.com/meteor/meteor/pull/12175) that is focused on bringing types to core

Akryum:
My thoughts on type merging is that is saves a little bit of code (I think we can trim the runtime code in the PR further) but I find usability to be much worse than implicit typing with zod and exporting functions:
- It involves knowledge about module declaration merging = more documentation and user support work (+ questions about where to put it)
- It duplicates the types declarations that can no longer be infered from zod and the validations.
- It is also not really usable in a JS-only project.

Akryum:
> Having support for Zod, Yup, and Simpl is a must, and also have the choice of just using it as a callback.

I don't really see how you can make type infering work in a generic way if your don't use something like `z.infer<TArgs>`

radekmie:
@Grubba27 

> Yeah, so my thoughts, given the comments I've received on this, are to be at least Simpl-schema open. At least on the alfa version is to have support on these two and later on or the community can help bring more schema validators.

As @Akryum said, it's not going to be easy to make it work as seamlessly with SimpleSchema as with zod. [There's this experiment of mine](https://forums.meteor.com/t/typescript-4-3-and-simpleschema/55697), but it's not easy to make it work in practice, i.e., with multiple schemas using `extend` and `pick`.

I'd rather see it in a "provide your own validator" way, like this:
```tsx
function createMethod<Args extends unknown[], Result>(
  name: string,
  validator: (args: unknown[]) => asserts args is Args,
  fn: (this: Meteor.MethodThisType, ...args: Args) => Result,
) {
  Meteor.methods({
    [name](...args: unknown[]) {
      validator(args);
      return fn.apply(this, args);
    },
  });

  return (...args: Args) =>
    new Promise<Result>((resolve, reject) => {
      // Cast as `EJSONable` to match `Meteor.apply` signature.
      Meteor.apply(name, args as EJSONable[], {}, (error, result) => {
        if (error) {
          reject(error);
        } else {
          resolve(result as Result);
        }
      });
    });
}
```

Now, this is a very basic implementation, without bells and whistles. But it's important to see, that the `validator` is a generic function with only one purpose - validate the arguments. Having it done this way, we can build different integrations:

```tsx
import { z } from 'zod';

function createZodMethod<ArgsSchema extends z.ZodTuple, Result>(
  name: string,
  schema: ArgsSchema,
  fn: (this: Meteor.MethodThisType, ...args: z.infer<ArgsSchema>) => Result,
) {
  return createMethod(name, schema.parse.bind(schema), fn);
}

const exampleZod = createZodMethod(
  'exampleZod',
  z.tuple([z.number(), z.string()]),
  // Types inferred from the schema.
  (foo, bar) => `foo = ${foo.toFixed(2)}, bar = ${bar}`,
);
```

```tsx
import SimpleSchema from 'simpl-schema';

function createSimpleSchemaMethod<Args extends unknown[], Result>(
  name: string,
  schema: SimpleSchema,
  fn: (this: Meteor.MethodThisType, ...args: Args) => Result,
) {
  // Wrap `args` in an object, as SimpleSchema doesn't support top-level arrays.
  return createMethod(name, args => schema.validate({ args }), fn);
}

const exampleSimpleSchema = createSimpleSchemaMethod(
  'exampleSimpleSchema',
  new SimpleSchema({ 'args.0': Number, 'args.1': String }),
  // No type inference out of the box - you have to make sure the schema and the
  // types are in-sync manually (or with the `TypedSimpleSchema` example).
  (foo: number, bar: string) => `foo = ${foo.toFixed(2)}, bar = ${bar}`,
);
```

> I would ask why this opinion.
> Surely this package could start as a community package and then be moved to the core. This is what you mean?

Yes, that's exactly what I mean. Of course, a "native" (I prefer to use the word "core") package is a great thing to have, but it's also a promise (pun intended). By making it official, the API _should_ remain stable (at least SemVer-wise), it _should_ be well tested, and most importantly - recommended as "the Meteor way" (it's official, right?).

I don't say it's not a good solution, but even this discussion alone shows that it's not a final one. And as such, I'd rather do what @StorytellerCZ said - release it as the next version of `mdg:validated-method`. Do a proper release, write a blog post, create a tutorial, etc. Then, if it really works for people, bring it to the core. Or not, because it's also totally fine to leave it as it is.

> At the same time, I think a more native solution is really neat. The user can choose how and if he wants to validate data or just have a prototype.
> I'm only thinking about how this could affect the JavaScript-only projects;

My idea with declaration merging is definitely not something that should be the final version. As @Akryum pointed out, it involves quite a lot of manual labor that we'd rather not have in the first place. But I just wanted to show an alternative where we add typings without any runtime cost. And no, we can't (or shouldn't) really change the types of `methods`/`apply`/`call` now, as it'd be a greatly breaking change.

Grubba27:
I got it and loved those inputs thx a lot @Akryum and @radekmie !! Given what was said, I will create a package (spiritual successor of ``mdg:validated-method``) as @StorytellerCZ pointed out. And we will do some community testing. In the first iteration, I will use Zod as a validator, but I want to make this generic method part of the roadmap. For the first release, this is really good. 
Another thing that I would love to hear people's comments about is hooks. I've made those 3 passed as configs(before, onError, after)  callbacks because I think it is the most generic way to access what is happening inside the methods. 
What do you guys think about this?

Akryum:
Right now I do not see a use case for them in my app. IMO it adds to the API surface without providing much more than native `try { await call() } catch (e){ }`.

Grubba27:
I moved this one to this package! I Was Heavily inspired by your implementation @Akryum and some of its parts by @zodern
I will give some more thought to making this package better and writing some tutorials and blog posts as suggested by @radekmie  
the package is here https://github.com/Grubba27/meteor-rpc

PR #12182, opened 2022-09-15
Use true mongo binary types

znewsham:
When implementing manual field level encryption with mongo, the decryption function expects a Binary data type with a `sub_type` of 6. Because meteor converts `Binary` to `Uint8Array` (and vice versa) we lose that. It isn't even possible to pass in a `Binary` - because mongo returns undefined.

This PR simple says to maintain the `Binary` type in most situations. The only situation it does not is when the `sub_type` is `0`. It does this for backwards compatibility (e.g., `sub_type` 0 is still returned as a `Uint8Array`)

### DIFF
--- a/packages/mongo/collection_tests.js
+++ b/packages/mongo/collection_tests.js
@@ -1,3 +1,6 @@
+
+var MongoDB = NpmModuleMongodb;
+
 Tinytest.add(
   'collection - call Mongo.Collection without new',
   function (test) {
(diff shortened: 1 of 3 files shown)


radekmie:
The code looks fine, but it definitely needs tests - both for the backward-compatible part as well as the new functionality.

znewsham:
@radekmie tests added - I wrongly assumed the backwards compatible piece would already have been covered by tests, from what I can tell the atom converter isn't directly covered by tests at all, it wasn't obvious where to put the tests so I put them in `collection_tests.js`

znewsham:
Separately, can you point me to where the test results are? I see the checks CI does, but I can't tell in which of the ~15 checks actually run these tests, and I can't see the test result anywhere (the Development.md seems out of date too, pointing to a script that doesn't exist)

Grubba27:
To check the result, you can do ``./meteor test-packages mongo`` and it will create a tab in http://localhost:3000/ running those tests

znewsham:
I've ran them locally - I want to see the results in CI (in particular to see if the change breaks anything else) - I assume circle CI jobs are where to look at this

radekmie (approved):

PR #12196, opened 2022-09-26
callAsync - New method to call async functions

denihs:
We've been testing things and having meetings about [this ](https://github.com/meteor/meteor/pull/12152#issuecomment-1228737920) for a few weeks now.

From the beginning, we knew the problem was with the DDP's context and how it was [managed](https://github.com/meteor/meteor/blob/ecf726ebf383917baba929e26adc40bd95281eab/packages/meteor/dynamics_browser.js#L20-L29).

When introducing async methods, the function executed in this [line](https://github.com/meteor/meteor/blob/ecf726ebf383917baba929e26adc40bd95281eab/packages/meteor/dynamics_browser.js#L24) returns a Promise that will be solved in the future. When it finally did, the method's context was already dead, and the stubs inside that Promise were no longer considered a [simulation](https://github.com/meteor/meteor/blob/38c0d70eb4d64dbb9342d5fa7f68e990deb1f0c1/packages/ddp-client/common/livedata_connection.js#L595).

Hence @edimarlnx's [solution](https://github.com/meteor/meteor/pull/12152). 

He tried to `await` for the result inside `EVp.withValue`, which means we have to make all the parent functions async. But by doing that, the problems @zodern mentioned [here](https://github.com/meteor/meteor/pull/12152#issuecomment-1228737920) show up.

So we thought of different approaches to solve this:

- Rewrite the global array `currentValues` as a Map so that we could somehow manage each method call and link them with a context, so then the Promises that were being solved in the future use that context. It didn't work because we didn't find a good way of providing a context to a Promise executed in the future.
- We thought of creating a queue of Promises, and each Promise receives the context before executing. But this idea couldn't work as well because there wouldn't be a way of getting each Promise's result or error.
- Then we thought about using `Zonejs`. But it has lots of [limitations]( https://angular.io/guide/roadmap#leverage-full-framework-capabilities-with-zonejs-opt-out), it looks like it doesn't work very well with async/wait, and Angular it's actually dropping it.

We also find out that the problem we're trying to solve (keep context for async functions) is something that's not possible today in JS. 

There is even a proposal for that to be solved https://www.proposals.es/proposals/Async%20Context.

Now here is the situation. We invested a good amount of time on this and saw that if we want to solve this, we'll have to create the solution, which can take a lot of time, and in the future, another solution may be taken (just as is happening with Fibers right now).

So instead, we're creating a method called callAsync. When calling an async method, you have to use this new function. It'll return a Promise, and then you can `await` for the result or use a `.then()`.

It solves the cases @zodern [mentioned ](https://github.com/meteor/meteor/pull/12152#issuecomment-1228737920) because we can run something like:

```
const result = await Meteor.callAsync('test1')
coll1.updateAsync({ _id: '1' }, { $inc: { b: 1 } });

// or

Meteor.callAsync('test1').then(result =>  {
    coll1.update({ _id: '1' }, { $inc: { b: 1 } });
});

```

And it solves the [original case](https://meteor-community.slack.com/archives/C01SWPPE81F/p1658764821424219), which originated the PR.


denihs:
@zodern could you take a look at this, please? 

@radekmie, as you have been on this since the first [PR](https://github.com/meteor/meteor/pull/12110), it would be nice to have your insight here as well.

docs/history.md
```diff
@@ -6,7 +6,8 @@
 * Update MongoDB driver to 4.9. [PR](https://github.com/meteor/meteor/pull/12097)
 
 #### Breaking Changes
-N/A
+
+* Now `Meteor.call()` should be called only for **sync** methods. For calling an async method, you should use `Meteor.callAsync()`. If `Meteor.call()` is called for a sync method, an error will be thrown.
```
hexsprite, line 10: typo: in the last sentence it should be "async" instead of "sync"

denihs: thank you!

zodern (changes requested):

zodern:
It seems this pr's solution to the two issues I found in the last one is that those situations are not supported, and it is the app's responsibility to ensure they do not happen.

The first issue was doing a client side mutation while an async method stub is running. There doesn't seem to be a way to support this right now so I'm fine with that. It seems it could be tricky for apps that use both to ensure this situation doesn't happen, and making a mistake would cause data loss, so I'm wondering if we should say it is not supported, instead of saying it is supported if you are careful.

-----

I've found another problem with async method stubs: if you call a method while an async method stub is running, it won't run on the server.

```js
Meteor.methods({
  async 'slow'() {
    console.log('before slow');
    await sleep(500);
    console.log('after slow');
  },
  'fast'() {
    console.log('fast');
  }
})

Test3 = function () {

  Meteor.callAsync('slow').then(() => console.log('slow finished'));
  Meteor.call('fast', () => {
    console.log('fast finished');
  });
}
```

On the server, only `before slow` and `after slow` is logged. `fast` never is logged.

**I wrote an idea below on how Meteor can handle async method stubs, but due to this problem it won't work.**

----

The other situation was calling multiple methods at the same time. I think Meteor needs to support this, and it is possible to.

One of the nice features of methods is it guarantees that the order you call methods on the client is the same order they will run on the server. For example:

```js
Meteor.call('method1');
Meteor.call('method2');
Meteor.call('method3');
```

Meteor will always run method1 first on the server, then method2, then method 3. However, with this PR, if method1 has an async stub it will run last. If both method1 and method2 have async stubs, the method with the stub that finishes last is the method that will run last on the server.

I think Meteor needs to keep this behavior of running methods on the server in the same order they were called on the client. Fixing this could also fix the issue I had found in the last PR. If Meteor uses a queue to run stubs one at a time, this would prevent multiple stubs from running at the same time, and ensure methods are run in the correct order on the server.

Regarding the new api's, I'm not sure why `Meteor.call` and `Meteor.callAsync` can't both support the same types of stubs, especially after making the other changes I've suggested.

Having `applyAsync` does help with some things, but I think we should wait to add it until we've figured out how it should work without having a callback since no other functions with the `Async` suffix do, and once applyAsync is added we can't easily change it.

-------

packages/ddp-client/common/livedata_connection.js
```diff
@@ -530,17 +530,26 @@ export class Connection {
     });
   }
 
+  isAsyncCall(name) {
+    return this._methodHandlers[name]?.constructor?.name === 'AsyncFunction';
```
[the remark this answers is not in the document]
radekmie: I second that. I often see Meteor methods that just expose some internal functions by simply mapping the parameters, like this:
```ts
Meteor.methods({
  'serviceName.methodName'(...args) {
    return Service.methodName(...args);
  },
});
```

Of course, there's no point in making this function `async`, even if `Service.methodName` returns a promise.

docs/history.md
```diff
@@ -2,11 +2,13 @@
...
 * New MongoDB Package Async API. [PR](https://github.com/meteor/meteor/pull/12028)
-* Node update to [v14.20.0](https://nodejs.org/en/blog/release/v14.20.0/) as part of the [July 7th security release](https://nodejs.org/en/blog/vulnerability/july-2022-security-releases/)
+* Node update to [v14.20.1](https://nodejs.org/en/blog/release/v14.20.0/) as part of the [September 22nd security release](https://nodejs.org/en/blog/vulnerability/september-2022-security-releases/)
 * Update MongoDB driver to 4.9. [PR](https://github.com/meteor/meteor/pull/12097)
+* Meteor.callAsync method. [PR] (https://github.com/meteor/meteor/pull/12196)
 
 #### Breaking Changes
-N/A
+
+* Now `Meteor.call()` should be called only for **sync** methods. For calling an async method, you should use `Meteor.callAsync()`. If `Meteor.call()` is called for an async method, an error will be thrown.
```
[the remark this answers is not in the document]
denihs: Yes, you're correct. But we're going to recommend the use, as most of the future methods will be async.

In any way, we're not going to throw these errors anymore.

packages/ddp-client/common/livedata_connection.js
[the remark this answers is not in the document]
denihs: That makes sense. We decided to remove this verification for now and focus on the documentation.

packages/meteor/dynamics_browser.js
```diff
@@ -28,6 +28,17 @@ EVp.withValue = function (value, func) {
   return ret;
 };
 
+EVp.withValueAsync = async function (value, func) {
+  var saved = currentValues[this.slot];
+  try {
+    currentValues[this.slot] = value;
+    var ret = await func();
```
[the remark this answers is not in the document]
denihs: We're going to look into this.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -550,6 +559,40 @@ export class Connection {
...
+   * @importFromPackage meteor
+   * @alias Meteor.callAsync
+   * @summary Invokes an async method passing any number of arguments.
+   * @locus Anywhere
+   * @param {String} name Name of method to invoke
+   * @param {EJSONable} [arg1,arg2...] Optional method arguments
+   * @returns {Promise}
+   */
+  async callAsync(name /* .. [arguments] .. callback */) {
+    if (!this.isAsyncCall(name)) {
```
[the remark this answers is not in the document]
denihs: As said above, we're not going to throw it anymore :+1:

packages/meteor/dynamics_nodejs.js
```diff
@@ -81,6 +81,31 @@ EVp.withValue = function (value, func) {
...
+ * @param {Any} value Value the environment variable should be set to
+ * @param {Function} func The function to run
+ * @return {Any} Return value of function
+ */
+EVp.withValueAsync = async function (value, func) {
+  Meteor._nodeCodeMustBeInFiber();
+
+  if (!Fiber.current._meteor_dynamics)
+    Fiber.current._meteor_dynamics = [];
+  var currentValues = Fiber.current._meteor_dynamics;
```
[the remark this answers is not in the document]
denihs: We're going to look into this :+1:

packages/ddp-client/common/livedata_connection.js
```diff
@@ -568,6 +611,51 @@ export class Connection {
...
+   * @param {Boolean} options.throwStubExceptions (Client only) If true, exceptions thrown by method stubs will be thrown instead of logged, and the method will not be invoked on the server.
+   * @param {Boolean} options.returnStubValue (Client only) If true then in cases where we would have otherwise discarded the stub's return value and returned undefined, instead we go ahead and return it. Specifically, this is any time other than when (a) we are already inside a stub or (b) we are in Node and no callback was provided. Currently we require this flag to be explicitly passed to reduce the likelihood that stub return values will be confused with server return values; we may improve this in future.
+   * @param {Function} [asyncCallback] Optional callback.
+   */
+  async applyAsync(name, args, options, callback) {
+    const { stubInvocation, invocation, ...stubOptions } = this._stubCall(name, EJSON.clone(args));
+    if (stubOptions.hasStub) {
+      if (!stubOptions.alreadyInSimulation) this._saveOriginals();
```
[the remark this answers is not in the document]
denihs: An error will be thrown if that happens. It shouldn't happen if you use it properly with `await` or `then()`. We're going to document these better and present scenarios where this could happen. This will be a limitation for methods that run in parallel. 

And it's good to remember that this is a limitation for now. While we don't have figured all this out yet.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -550,6 +559,40 @@ export class Connection {
...
+   * @summary Invokes an async method passing any number of arguments.
+   * @locus Anywhere
+   * @param {String} name Name of method to invoke
+   * @param {EJSONable} [arg1,arg2...] Optional method arguments
+   * @returns {Promise}
+   */
+  async callAsync(name /* .. [arguments] .. callback */) {
+    if (!this.isAsyncCall(name)) {
+      throw new Error(
+        "You should use Meteor.call() when calling sync Methods! If you don't call the proper function, some unexpected behaviors may appear."
```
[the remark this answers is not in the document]
denihs: This verification is going to be removed.

packages/ddp-server/livedata_server.js
```diff
@@ -762,16 +762,16 @@ Object.assign(Session.prototype, {
           }
         }
 
-        resolve(DDPServer._CurrentWriteFence.withValue(
+        resolve(Promise.await(DDPServer._CurrentWriteFence.withValueAsync(
```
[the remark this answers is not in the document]
denihs: We're going to test more, but we did several tests around this, and everything seems to be the same.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -534,7 +534,7 @@ export class Connection {
    * @memberOf Meteor
    * @importFromPackage meteor
    * @alias Meteor.call
-   * @summary Invokes a method passing any number of arguments.
+   * @summary Invokes a sync method passing any number of arguments.
```
radekmie, line 537: Should we update it to say something like "Invokes a method with a sync stub, passing any number of arguments." instead?

packages/ddp-client/common/livedata_connection.js
```diff
@@ -550,6 +550,37 @@ export class Connection {
...
+   * @param {EJSONable} [arg1,arg2...] Optional method arguments
+   * @returns {Promise}
+   */
+  async callAsync(name /* .. [arguments] .. callback */) {
+    const args = slice.call(arguments, 1);
+    if (args.length && typeof args[args.length - 1] === 'function') {
+      throw new Error(
+        "Meteor.callAsync() does not accept a callback. You should 'await' the result, or use .then()."
+      );
+    }
```
radekmie, line 569: I'm still thinking about it, but... Is there a need for this check? I know, it's "safe", but we don't really add them in other `*Async` methods. (Also, the comment in the function arguments could be updated.)

packages/ddp-client/common/livedata_connection.js
```diff
@@ -550,6 +550,37 @@ export class Connection {
...
+      Meteor.bindEnvironment(() => {
+        this.applyAsync(name, args, (err, result) => {
+          if (err) {
+            reject(err);
+            return;
+          }
+          resolve(result);
+        });
+      })();
+    });
```
radekmie, line 582: This should be as easy as `return this.applyAsync(name, args);`. Is there any reason for not doing that?

packages/ddp-client/common/livedata_connection.js
```diff
@@ -568,6 +599,65 @@ export class Connection {
...
+         * But as the Meteor package is not compiled by ecmascript, it is unable to use newer syntax in the browser,
+         * such as, the async/await.
+         *
+         * So, to keep supporting old browsers, like IE 11, we're creating the logic one level above.
+         */
+        const currentContext = DDP._CurrentMethodInvocation.setNewContextAndGetCurrent(
+          invocation
+        );
+        try {
+          stubOptions.stubReturnValue = await stubInvocation();
```
radekmie, line 649: According to the comment a few lines before, this `await` is also not allowd, right? And so is the `async` in front of the function?

packages/ddp-client/common/livedata_connection.js
```diff
@@ -568,6 +599,65 @@ export class Connection {
...
+   * @param {Boolean} options.noRetry (Client only) if true, don't send this method again on reload, simply call the callback an error with the error code 'invocation-failed'.
+   * @param {Boolean} options.throwStubExceptions (Client only) If true, exceptions thrown by method stubs will be thrown instead of logged, and the method will not be invoked on the server.
+   * @param {Boolean} options.returnStubValue (Client only) If true then in cases where we would have otherwise discarded the stub's return value and returned undefined, instead we go ahead and return it. Specifically, this is any time other than when (a) we are already inside a stub or (b) we are in Node and no callback was provided. Currently we require this flag to be explicitly passed to reduce the likelihood that stub return values will be confused with server return values; we may improve this in future.
+   * @param {Function} [asyncCallback] Optional callback.
+   */
```
radekmie, line 631: Let's add `@returns` here, just like in `callAsync`.

packages/ddp-server/livedata_server.js
```diff
@@ -762,16 +762,43 @@ Object.assign(Session.prototype, {
...
+        };
+        const getCurrentWriteFenceResult = () => {
+          const currentContext = DDPServer._CurrentWriteFence.setNewContextAndGetCurrent(
+            fence
+          );
+          try {
+            return getCurrentMethodInvocationResult();
+          } finally {
+            DDPServer._CurrentWriteFence.set(currentContext);
+          }
```
radekmie, line 798: That's exactly the same as
```ts
return DDPServer._CurrentWriteFence.withValue(fence, getCurrentMethodInvocationResult);
```
isn't it?

Similarly, the `getCurrentMethodInvocationResult` could use `withValue` too.

packages/ddp-server/livedata_server.js
```diff
@@ -784,20 +811,21 @@ Object.assign(Session.prototype, {
         id: msg.id
       };
 
-      promise.then((result) => {
+      try {
+        const result = Promise.await(promise);
```
radekmie, line 815: This is a big change - previously, this promise ran in the background, now this function blocks waiting for it.
1. Is it expected?
2. Is it needed?

packages/meteor/dynamics_browser.js
```diff
@@ -28,6 +28,17 @@ EVp.withValue = function (value, func) {
...
 
+EVp.set = function (context) {
+  currentValues[this.slot] = context;
+};
+
+EVp.setNewContextAndGetCurrent = function (value) {
+  const saved = currentValues[this.slot];
+  this.set(value);
+  return saved;
+};
```
radekmie, line 39: 1. With my two comments, these two are most likely obsolete.
2. If they're used only internally, maybe these should be private? I'm still _a little_ against increasing the API surface if it's not needed.

packages/ddp-client/common/livedata_connection.js
denihs: Sure, that's fine!

packages/ddp-client/common/livedata_connection.js
denihs: We're going to keep it for now, as this won't affect the current apps. But we can revisit this later.

packages/ddp-client/common/livedata_connection.js
denihs: I added a comment in the code to explain this code. Here is the comment:

```
This is necessary because when you call a Promise.then, you're actually calling a bound function by Meteor.
     
This is done by this code https://github.com/meteor/meteor/blob/17673c66878d3f7b1d564a4215eb0633fa679017/npm-packages/meteor
     
So, without resting the context at this point, and then calling Meteor.bindEnvironment before calling
applyAsync, when you call a ".then()", like "Meteor.callAsync().then()", the global context (inside currentValues)
will be from the call of Meteor.callAsync(), and not the context after the promise is done.
     
This means that without this code if you call a stub inside the ".then()", this stub will act as a simulation
and won't reach the server.
```

TLDR: Without this, you can't use `Meteor.callAsync().then()`.

packages/ddp-client/common/livedata_connection.js
denihs: That was for the `meteor` package. It's totally fine to use in the `ddp-client` package, as it depends on the ecmascript package.

That's why this solution works.

packages/ddp-server/livedata_server.js
denihs: About the DDPServer._CurrentWriteFence.withValue you're completely right. Fixed.

About the getCurrentMethodInvocationResult, we need to do it this way.

Now, [this function](https://github.com/meteor/meteor/pull/12196/files#diff-173e69ea0353a765b98017d67abc45ec7ce1449178466dc74738324db83f9183R772) can return a promise, and we need to wait for the result.

So the first solution was to create a [`withValueAsync`](https://github.com/meteor/meteor/pull/12196/commits/15b1e6fddfc5a03c4dd17d97d3b577996b9c0c61#diff-47412bfdfb4c9c0628248595b3f6c447bf7ad49535e508628dc704b729028ea8L93-L106) that waits the promise to solve. But [this](https://github.com/meteor/meteor/pull/12196#discussion_r981792165) problem pointed out by Zodern was created. But also by doing his [suggestion](https://github.com/meteor/meteor/pull/12196/commits/15b1e6fddfc5a03c4dd17d97d3b577996b9c0c61#diff-47412bfdfb4c9c0628248595b3f6c447bf7ad49535e508628dc704b729028ea8R93-R106) (wrapping that function in a new Fiber) was changing the dynamics and when you called a `Meteor.call()` without the callback, it didn't return a result anymore. Resulting in a lot of tests breaking. 

So I'm applying there the same solution I did for the [applyAsync](https://github.com/meteor/meteor/pull/12196/files#diff-5a0f1c2953dc6856044b4803bb19c0b301386cd8836c878e36f2776d6316dc79R657-R667). Resolving the promise one level above.

packages/ddp-server/livedata_server.js
denihs: This change is not needed!

I was doing some tests around this, and I ended up leaving this there without thinking about the impact. 

But it's fixed now. Thanks.

packages/meteor/dynamics_browser.js
denihs: We still need these, but I made them private.

zodern:
The problem is we don't know if we should add the call to the queue, or if it is from inside a stub and should be run immediately to give the stub the simulated result: https://github.com/meteor/meteor/pull/12196#issuecomment-1260263473

We have the choice of breaking either:
- Using Meteor.call and Meteor.apply within async method stubs
- Calling a method outside of an async method stub while an async method stub is running

This PR breaks the second.  Running a method while a stub is running will be broken anyway, so any queueing has to be done by the app. A queue implemented by Meteor wouldn't change anything.

Instead, breaking using Meteor.call and Meteor.apply within async method stubs would allow us to support calling Meteor.call while an async method stub is running, and we would need a queue to ensure both the method stubs and the methods on the server run in the correct order.

My apps don't use Meteor.call inside of method stubs, so I prefer this second option.

dr-dimitru:
@denihs @zodern Does this limitation to stubs applies to Methods **registered only on the Server** and **called only from the Client**?

@radekmie that's exactly my first thought —— queue should get used behind the scenes, so we don't even need to explain this here and there (docs at the moment are very confusing on Async Methods usage)

P.S. We run multiple Meteor projects in the parallel and due to the company size — hire and onboard new developers weekly, since v2.8 I see a lot of confusion when new devs come across Async Mongo methods (*which are missing from docs!*), and Async Methods. From today I asked all developers to stick to the old and well-known "nonAsync" API; Curios to hear how other teams handled 2.8 features...

PR #12200, opened 2022-10-01
Add missing return statement to CollectionPrototype._validatedInsert

### LINKED ISSUE #12159: `CollectionPrototype._validatedInsert` doesn't return the `insertedId`
When using the `allow-deny` package, documents that are inserted on the client go through https://github.com/meteor/meteor/blob/d3c354c7f4f4dfae7a9c4bc134f32896567161d1/packages/allow-deny/allow-deny.js#L240-L264 but this doesn't return the `insertedId` from `self._collection.insert()`. If the corresponding method stub isn't present on the client, this prevents you from getting the new document's `_id`.

In contrast, both `CollectionPrototype. _validatedUpdate` and `CollectionPrototype._validatedRemove` do return their results with https://github.com/meteor/meteor/blob/d3c354c7f4f4dfae7a9c4bc134f32896567161d1/packages/allow-deny/allow-deny.js#L362-L363 and https://github.com/meteor/meteor/blob/d3c354c7f4f4dfae7a9c4bc134f32896567161d1/packages/allow-deny/allow-deny.js#L413 so I can't tell if this is just an oversight or if there's a reason for not doing so with inserts.

I noticed this when I'm running multiple Meteor servers, where one connects to the other to reach the database. The former therefore acts as a DDP client and has Minimongo running on the serverside. That's because I can't give it direct access to the MongoDB since this instance is running on a less secure server. Minimongo on the server doesn't create the method stubs, which is perfectly fine if the `insertedId` would be returned.

For me this happens now with Meteor@2.7.3 on macOS. But looking at the history of `allow-deny`, it seems that `CollectionPrototype._validatedInsert` has been left mostly untouched since 2015, so it's been there all along. I guess that that the method stubs typically mask this issue.

Because this a very simple bug / question regarding a missing `return` statement I didn't yet make a reproduction, but if necessary I can do so of course.

Blaster4385:
## Addressed Issues
- This fixes  #12159 

## Description of change

-  This PR fixes the issue where ```CollectionPrototype._validatedInsert``` doesn't return the ```insertedId``` for newly inserted documents.

## List Of Changed Files:

- packages/allow-deny/allow-deny.js

### DIFF
--- a/packages/allow-deny/allow-deny.js
+++ b/packages/allow-deny/allow-deny.js
@@ -260,7 +260,7 @@ CollectionPrototype._validatedInsert = function (userId, doc,
   if (generatedId !== null)
     doc._id = generatedId;
 
-  self._collection.insert.call(self._collection, doc);
+  return self._collection.insert.call(self._collection, doc);
 };
 
 // Simulate a mongo `update` operation while validating that the access


StorytellerCZ (approved):

StorytellerCZ:
Looks good. Would be nice to add tests as well.

Blaster4385:
> Looks good. Would be nice to add tests as well.

Thanks. I'm not very proficient in writing tests but will try my best to write tests for this. Any suggestions would be great.

radekmie:
The test has to check if this returned value actually works. How did you test it locally? I assume the test would be 1-to-1 what you did. You can check other `insert` tests first.

nachocodoner:
Closing this issue. The fix this PR provided back then has since been included in devel, which now causes a conflict. The latest devel branch already returns correctly from the related function as this PR intents.

For more details and test coverage references, see this comment: https://github.com/meteor/meteor/issues/12159#issuecomment-3400465676

PR #12201, opened 2022-10-01
refactor: inline function

dikiogres:


StorytellerCZ:
Reject! Just Hacktoberfest spam.

radekmie:
@StorytellerCZ I think we'd need to create a `spam` label, according to [the rules](https://hacktoberfest.com/participation/).

PR #12202, opened 2022-10-01
Move some OAuth of out of accounts-base

StorytellerCZ:
Following the discussion in #12171 I have taken the initiative and moved some OAuth related code out of `accounts-base` into `accounts-oauth`.

I have also added some additional conditionals in [6196453](https://github.com/meteor/meteor/pull/12202/commits/619645307620a5dc822da6bdade4992e4eb06d26), which can be cherry picked if needed, which might have caused the error seen in #12171.


Grubba27:
We are making some changes in the accounts(adding async) I think it makes sense to move this change to 2.9. but I'm all ears if you want this could go in a patch for 2.8.x

denihs (approved):
LGTM

denihs:
Let's merge this into 2.9.

StorytellerCZ:
@denihs this will require a bit of a testing during the beta release phase to make sure that I haven't missed some use case and that with all the other changes going on around that nothing has been missed or overridden.

PR #12204, opened 2022-10-02
Typescript 4.6.4

StorytellerCZ:
Upgrade to [TypeScript 4.6.4](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/).

Right now I have just bumped the versions. This will require to publish the relevant npm packages otherwise the tests will fail.


afrokick:
Maybe we can skip some version and upgrade to `4.8.*` version?

StorytellerCZ:
@harrisonhunter thanks, fixed.
@afrokick Thats an option as well, but I would like go one feature version to make it easier to catch any issues that might arise. But if the consensus is to bump to latest, then I have no problem with it.

radekmie:
IMO we should release 4.6, 4.7, 4.8, and already prepare for 4.9 (it's in beta). It's a trivial thing and will let people upgrade gradually. There aren't that many breaking changes, but some dependencies may have problems.

StorytellerCZ:
My plan is to increase the version with each of the planned Meteor releases until we get to the latest version.

radekmie:
But wouldn't that be like one version in a couple of weeks? If so, I'd really make it faster, preferably before mid-November (4.9 release date).

StorytellerCZ:
@radekmie that depends on how fast are the releases after Meteor 2.8 going to be.

hexsprite:
Is there any way to allow the typescript dependency to be provided from the application node_modules? Then one wouldn't need to update Meteor every time there's a new Typescript version.

Either way, also +1 for the latest stable version of TypeScript as opposed to an older one.

Grubba27:
> Is there any way to allow the typescript dependency to be provided from the application node_modules? Then one wouldn't need to update Meteor every time there's a new Typescript version.

I never thought about this issue on TypeScript versions @hexsprite. Maybe @zodern has an idea for it?

afrokick:
@Grubba27 any chance to release it as a part of 2.8.1 ?

Grubba27:
Hey @afrokick, it can be. but it is still tagged as a draft so I'm not sure if something is missing. Is there something that I can do to make this ready @StorytellerCZ ?

StorytellerCZ:
Adding to my todo list for tomorrow.

StorytellerCZ:
Let's release Typescript 4.6.4 with Meteor 2.8.1 and I'll fix a PR to update to the latest version for Meteor 2.9.

StorytellerCZ:
@Grubba27 I do believe that you will have to go into this PR and publish some of the NPM packages in order for the tests to start working. Not sure what the current procedure for that is.

afrokick:
@StorytellerCZ @Grubba27 any progress?

npm-packages/meteor-babel/package.json
```diff
@@ -47,7 +47,7 @@
     "convert-source-map": "^1.6.0",
     "lodash": "^4.17.21",
     "meteor-babel-helpers": "0.0.3",
-    "typescript": "^4.5.4"
+    "typescript": "^4.6.4"
```
Julusian, line 50: I think this should be:
```suggestion
    "typescript": "~4.6.4"
```

Otherwise anywhere when this is installed without a lockfile (or without an up to date lockfile) will install the latest version which matches. Which would right now end up with 4.8.3.
I don't know how lockfiles are used in the meteor ecosystem, so this may not apply. But as I don't see a lockfile update in this PR, have you checked what the result it produces will be?

npm-packages/eslint-plugin-meteor/scripts/dev-bundle-tool-package.js
```diff
@@ -14,8 +14,8 @@ var packageJson = {
     pacote: "https://github.com/meteor/pacote/tarball/a81b0324686e85d22c7688c47629d4009000e8b8",
     "node-gyp": "8.0.0",
     "node-pre-gyp": "0.15.0",
-    typescript: "4.5.4",
-    "@meteorjs/babel": "7.16.0-beta.1",
+    typescript: "4.6.4",
```
afrokick, line 17: As I mentioned in https://github.com/meteor/meteor/pull/12313 we don't need this scripts folder and we need to update `scripts/dev-bundle-tool-package.js` instead of `npm-packages/eslint-plugin-meteor/scripts/dev-bundle-tool-package.js`

Grubba27:
Any ideas why it is failing?

afrokick:
@Grubba27 you need to update package-lock.json file for `meteor-babel`. (as you can see in my previous PR https://github.com/meteor/babel/pull/44/files )

radekmie:
Also, would it be possible to use it in pre-2.8 Meteor too? It'd be beneficial for some of our projects.

Grubba27:
Done! Published under the `` @meteorjs/babel@7.17.0-beta.3`` namespace but for some reason the commits have not yet come

Grubba27:
Will merge it and solve the non-passing tests on release branch

npm-packages/eslint-plugin-meteor/scripts/dev-bundle-tool-package.js
afrokick: @Grubba27 Meteor itself doesn't use new TypeScript because of the issue I mentioned above.

Grubba27: Did not realize that. Thanks again for pinging me on this.
Do you want to make a pr for this? I will add it to the 2.9.1 patch release. If not, I will do it

PR #12206, opened 2022-10-04
Update accounts.md

Paari1263:
This PR has the creation of a new section in the docs for `Accounts.registerLoginHandler`. Please do review and  comment if this PR contains have  any errors or changes to be made.

### DIFF
--- a/docs/source/api/accounts.md
+++ b/docs/source/api/accounts.md
@@ -338,3 +338,150 @@ Accounts.ui.config({
 
 
 Since Meteor 2.7 you can configure these in your Meteor settings under `Meteor.settings.public.packages.accounts-ui-unstyled`.
+  
+<h2>Meteor register LDAP login request handler</h2>
+Let’s get started with the client.
+  <h3>client/users/actions.js</h3>
+  
+  ```js
+  ...
+  let loginUserWithLDAP = (email, password, callback) => {
+      var loginRequest = {
+        ldap: true,
+        email: email,
+        pass: password,
+      }
+      Accounts.callLoginMethod({
+        methodArguments: [loginRequest],
+        userCallback: callback
+      })
+    }
+
+  loginUserWithLDAP(email, password, (error, result) => {
+    if (!error) {
+    ...
+  ```
+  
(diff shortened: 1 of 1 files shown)


StorytellerCZ (commented):
I don't know. This looks like something that should be either a package on its own or a tutorial. At the very least it should be in a guide, not in the docs section. If the point here is to document `registerLoginHandler` then it should follow the structure like the other API endpoints docs.

Paari1263:
What should I change @StorytellerCZ , any suggestions please.

StorytellerCZ:
Do a JS Doc on the function itself, then add that doc block (look on some other functions on how they do it, like `Meteor.userId()`) to the docs with a short description about what this is for. This entire tutorial that you made here is a nice example that should be moved to the guide.

Paari1263:
Ok, can you please send the src link where to add the js doc on

Grubba27:
Hey @Paari1263 the js doc is on the function ``Accounts.registerLoginHandler`` like this:

```javascript

  /**
   *
   * The main entry point for auth packages to hook in to login.
   *
   * A login handler is a login method which can return `undefined` to
   * indicate that the login request is not handled by this handler.
   *
   * @param name {string} Optional.  The service name, used by default
   * if a specific service name isn't returned in the result.
   *
   * @param handler {Function} A function that receives an options object
   * (as passed as an argument to the `login` method) and returns one of:
   * - `undefined`, meaning don't handle;
   * - a login method result object
   **/
registerLoginHandler(name, handler) {
```

on this [file](https://github.com/meteor/meteor/blob/222022ecf0dde1d5ad35aed1ce223aa02c6e4041/packages/accounts-base/accounts_server.js#L563) 

Agree with @StorytellerCZ about keeping what you have made so far and moving it to the guide.
Awesome job!

Grubba27 (changes requested):
Loved the tutorial.
just adjust the missing js doc [here](https://github.com/meteor/meteor/blob/222022ecf0dde1d5ad35aed1ce223aa02c6e4041/packages/accounts-base/accounts_server.js#L563) and make sure this tutorial is on the guide and is good

italojs:
abandoned

PR #12207, opened 2022-10-04
Updated fetch,md

rahulgithub-web:
Replaced normal function to an arrow function in line no. 42 and a specific status code 500 (which is used for Internal Server Error) in catch in line number 60 with the status false and with a proper message whatever comes from the server side.

### DIFF
--- a/docs/source/packages/fetch.md
+++ b/docs/source/packages/fetch.md
@@ -39,7 +39,7 @@ For the most part though, you will only need to import `fetch` and `Headers`.
 import { Meteor } from 'meteor/meteor';
 import { fetch, Headers } from 'meteor/fetch';
 
-const postData =async(url, data) => {
+const postData = async(url, data) => {
     try {
       const response = await fetch(url, {
           method: 'POST', // *GET, POST, PUT, DELETE, etc.
@@ -57,7 +57,7 @@ const postData =async(url, data) => {
       const data = await response.json();
       return response(null, data);
     } catch (err) {
-      return res.status(500).send({status: false, msg: err.message});
+      return response.status(500).send({status: false, msg: err.message});
     }
 }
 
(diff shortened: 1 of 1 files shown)


docs/source/packages/fetch.md
```diff
@@ -57,7 +57,7 @@ async function postData (url, data) {
       const data = await response.json();
       return response(null, data);
     } catch (err) {
-      return response(err, null);
+      return res.status(500).send({status: false, msg: err.message});
```
StorytellerCZ, line 60: This doesn't look correct. Have you tried to run this sample? Where is `res` coming from?

StorytellerCZ (changes requested):

docs/source/packages/fetch.md
rahulgithub-web: res is same like response

docs/source/packages/fetch.md
```diff
@@ -39,7 +39,7 @@ For the most part though, you will only need to import `fetch` and `Headers`.
 import { Meteor } from 'meteor/meteor';
 import { fetch, Headers } from 'meteor/fetch';
 
-async function postData (url, data) {
+const postData =async(url, data) => {
```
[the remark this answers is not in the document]
rahulgithub-web: ok sir done

rahulgithub-web:
Done sir, added space after the equal to sign(=) and in line 60 res should be like response.

docs/source/packages/fetch.md
```diff
@@ -57,7 +57,7 @@ async function postData (url, data) {
       const data = await response.json();
       return response(null, data);
     } catch (err) {
-      return response(err, null);
+      return response.status(500).send({status: false, msg: err.message});
```
StorytellerCZ, line 60: I do not recall the response function having a status or send methods. This looks like Express thing.

StorytellerCZ (changes requested):

docs/source/packages/fetch.md
rahulgithub-web: Yeah sir, you are right. I will look forward to it sir. Thank you.

PR #12209, opened 2022-10-05
Update default Facebook API to v15 and fix local changelog

StorytellerCZ:
A small update to the Facebook package. First update the default API version to v15. No changes to the OAuth API at all.
* [Changelog for v14](https://developers.facebook.com/docs/graph-api/changelog/version14.0#graph-api)
* [Changelog for v15](https://developers.facebook.com/docs/graph-api/changelog/version15.0#graph-api)

I have also updated the local changelog with the changes since last release.


StorytellerCZ:
@denihs I don't recall if I have to bump the version or if that is something that will be taken care of by the release script, please correct as needed.

PR #12210, opened 2022-10-05
Update accounts_server.js

Paari1263:
Update on the js doc from the suggestion of @StorytellerCZ , please review this PR and suggest changes if needed.


Paari1263:
Please take it as contribution in Hacktoberfest 2022.

StorytellerCZ:
Nope, wrong! Look on other methods that have above them comments that start like this `/**`
More on JS Doc: https://jsdoc.app/

The idea is that you describe the function add types and so on. Then you will be able to generate API Box in the docs and insert it there.

Also the LDAP code doesn't belong here at all.

Issue #12211, opened 2022-10-05
_findUserByQuery should never do the case insensitve lookup without a limit of 2


ToyboxZach:
In _findUserByQuery there is this code

```
      user = Meteor.users.findOne(selector, options);
      // If user is not found, try a case insensitive lookup
      if (!user) {
        selector = this._selectorForFastCaseInsensitiveLookup(fieldName, fieldValue);
        const candidateUsers = Meteor.users.find(selector, options).fetch();
        // No match if multiple candidates are found
        if (candidateUsers.length === 1) {
          user = candidateUsers[0];
        }
      }
```

_selectorForFastCaseInsensitiveLookup in rare cases can be very inefficient (if you have a lot of users with the same beginning)

So when we call this function from the password handler we don't have limit set to 2, which means this can be loading in way too many items

This is in the password handler:

```

  const user = Accounts._findUserByQuery(options.user, {fields: {
    services: 1,
    ...Accounts._checkPasswordUserFields,
  }});
```

Which is the case causing my issues.

Not sure if we should fix _findUserByQuery to always set the limit, or make sure it is always called correctly.

StorytellerCZ:
Will be released in 2.8.1.

PR #12212, opened 2022-10-05
Fix case insensitive lookup resource overuse

ToyboxZach:
This is just to always set the limit on this query since there is no reason to ever return more than 2 items. This fixes cases where you may have a large number of users that start with the same few characters.

Referring to this https://github.com/meteor/meteor/issues/12211#issue-1398221726


radekmie (approved):
That's totally correct, but I don't think it'll actually help much in practice. You'd need to have _a lot_ of duplicated customers to benefit from this. In any case, it won't hurt.

PR #12215, opened 2022-10-09
Refactored SynchronousQueue

KoenLav:
I may be missing something, but I think this simplifies the SynchronousQueue and shouldn't break anything.

### DIFF
--- a/packages/meteor/fiber_helpers.js
+++ b/packages/meteor/fiber_helpers.js
@@ -59,51 +59,42 @@ Meteor._SynchronousQueue = function () {
 
 var SQp = Meteor._SynchronousQueue.prototype;
 
-const runTaskWithFibers = ({ task, self }) => {
+SQp.runTask = function(task) {
+  const self = this;
+
   if (!self.safeToRunTask()) {
-    if (Fiber.current) {
-      throw new Error("Can't runTask from another task in the same fiber");
+    if (Meteor._isFibersEnabled) {
+      if (Fiber.current) {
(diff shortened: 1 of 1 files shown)


packages/meteor/fiber_helpers.js
```diff
@@ -59,51 +59,42 @@ Meteor._SynchronousQueue = function () {
 
 var SQp = Meteor._SynchronousQueue.prototype;
 
-const runTaskWithFibers = ({ task, self }) => {
+SQp.runTask = function(task) {
+  const self = this;
```
radekmie, line 63: If this meant to simplify this code, I'd get rid of `self`s entirely.

packages/meteor/fiber_helpers.js
```diff
@@ -118,18 +109,18 @@ SQp.queueTask = function (task) {
...
-    return true;
+  var self = this;
+
+  if (Meteor._isFibersEnabled) {
+    return Fiber.current && self._currentTaskFiber !== Fiber.current;
   }
 
-  var self = this;
-  return Fiber.current && self._currentTaskFiber !== Fiber.current;
+  return !self._runningOrRunScheduled;
```
radekmie, line 123: It was `true` before, is this a desired change?

packages/meteor/fiber_helpers.js
```diff
@@ -206,8 +171,11 @@ SQp._run = function () {
   var taskHandle = self._taskHandles.shift();
```
radekmie, line 171: This code could use `let`s instead of `var`s.

PR #12219, opened 2022-10-11
Fix MongoDB driver Warning: cursor.count

denihs:
After updating the MongoDB driver version, this warning showed up:

```
[MONGODB DRIVER] Warning: cursor.count is deprecated and will be removed 
in the next major version, please use collection.estimatedDocumentCount or collection.countDocuments instead
```

The solution seems to be as easy as it is. 

I chose the `countDocuments` because it seems to be the closest to the original `count()`.


denihs:
@radekmie could you have a look at this, please? I think you're the best person to tell if this change is ok.

radekmie (approved):
Everything looks good. It's also something that @xet7 reported in https://github.com/meteor/meteor/pull/12057#issuecomment-1258697719 and I didn't got time to look into.

@denihs Could you check if it solves the WeKan problem?

xet7:
@denihs 

Is countDocuments slower than estimatedDocumentCount ? There is a lot of documents.

radekmie:
@xet7 It's as slow as `count`, as it does basically the same thing. `estimatedDocumentCount` tend to be faster, but it's only an estimate, and the existing code could break. I planned to add both methods to the collections API anyway.

denihs:
I believe it solves WeKan's problem, as it solved the same problem I was having with other applications. 

If it shows again, please let us know @xet7 :raised_hands:

paulincai:
Hi, I am experimenting with countAsync() and started to get the same mongo driver warning.

Example:
```
const hasAccess = await Orgs.find({
    _id: orgId,
    $or: [{ userId }, { admins: userId }]
  }, { limit: 1 }).countAsync()

```
I don't know what is WeKan's problem but is there a specific way to use countAsync() ?

radekmie:
@paulincai It's most likely the same issue. Please check it after the release.

PR #12220, opened 2022-10-13
Feat: not using insecure & autopublish by default

Grubba27:
Standardized the use of insecure packages in skeletons and made this an opt-in feature. 

It is only possible to pass in this flag when you are not using the package flag( it does have a validation)

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -522,7 +522,7 @@ export const AVAILABLE_SKELETONS = [
 
 main.registerCommand({
   name: 'create',
-  maxArgs: 2,
+  maxArgs: 1,
   options: {
     list: { type: Boolean },
     example: { type: String },
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -548,7 +548,13 @@ main.registerCommand({
   // latest release to create a package if we are inside an app)
   if (options.package) {
     var packageName = options.args[0];
-
+    if (options.prototype) {
+      Console.error(
+        `The ${Console.command('--prototype')} option is not supported for packages.`
+      );
+      Console.error();
+      throw new main.ShowUsage;
+    }
     if (options.list || options.example) {
       Console.error("No package examples exist at this time.");
       Console.error();
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -550,7 +550,7 @@ main.registerCommand({
     var packageName = options.args[0];
     if (options.prototype) {
       Console.error(
-        `The ${Console.command('--prototype')} option is not supported for packages.`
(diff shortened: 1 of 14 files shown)


tools/cli/commands.js
```diff
@@ -522,7 +522,7 @@ export const AVAILABLE_SKELETONS = [
 
 main.registerCommand({
   name: 'create',
-  maxArgs: 1,
+  maxArgs: 2,
```
denihs, line 525: Are you sure this number needs to be increased?

denihs (changes requested):
Everything looks good, but we have some tests failing. You will probably have to start passing this new flag to the apps created in the tests.

tools/cli/commands.js
Grubba27: https://github.com/meteor/meteor/blob/4f16220fa648660ed24fde36f99942e89d7f6c8a/tools/cli/main.js#L163
yeah, I miss understood what I read :( We can have as many options as we want but obligatory args is just one (the path)

tools/cli/commands.js
```diff
@@ -547,7 +548,13 @@ main.registerCommand({
   // latest release to create a package if we are inside an app)
   if (options.package) {
     var packageName = options.args[0];
-
+    if (options.prototype) {
+      Console.error(
+        `The ${Console.command('--prototype')} option is not supported for packages.`
```
StorytellerCZ, line 553: Might be better to say `is no longer supported` and add a link to some article explaining why.

tools/cli/help.txt
```diff
@@ -191,6 +191,7 @@ Options:
   --tailwind    Create a basic react-based app, with tailwind configured.
   --chakra-ui   Create a basic react-based app, with chakra-ui configured.
   --solid       Create a basic solid-based app.
+  --prototype   Create a prototype app.
```
StorytellerCZ, line 194: More explanation is needed here. This is too plain of a description.

StorytellerCZ (changes requested):

tools/cli/help.txt
Grubba27: I've rewrote like this:
``Create a prototype app with the insecure & autopublish packages. Can be used along with other app commands`` what do you think?

StorytellerCZ: Maybe expand that they are for prototyping, but this is good as well.

henriquealbert (approved):
lgtm, just make sure to update all tutorials and everything that mention and uses autopublish and insecure @Grubba27

retani:
Being a long term user and promoter of meteorjs I am just super happy about this. Insecure created so much confusion. Very grateful. Thanks all involved. 🙏🏼

PR #12221, opened 2022-10-14
Don't apply babel async-await plugin when not running on Fibers.

matheusccastroo:
This PR will be used to break down a bit of the changes made in #12157, while also making a good way to keep track of progress/check blockers based on tests.


matheusccastroo:
Related to issue #12289.

denihs (approved):

Issue #12230, opened 2022-10-17
Indicate Meteor server port on app restart


Grubba27:
This would be a change from what we have

From:
```
=> Meteor server restarted
```

To:
```
=> Meteor server restarted on port http://localhost:3400

```

Or the same, but in case the user passes the ``ROOT_URL``

To:
```
=> Meteor server restarted on port http://127.0.0.1:3400
```
another thing is wrap the URL with the [Command.url](https://github.com/meteor/meteor/blob/0eaf07c90804811936acf3c98e9d981c0a736a3b/tools/cli/commands.js#L910) method

## How to implement?

 - [ ]  you are going to start here on this [file](https://github.com/meteor/meteor/blob/b6e127cfc9fd33a5fb16db540531e1fa1e8d7f48/tools/runners/run-log.js#L162);
 - [ ] get the ``ROOT_URL`` to show up along the console message as well
 - [ ] Adjusts tests that may break

---- 
_discussion for more context_
### Discussed in https://github.com/meteor/meteor/discussions/12226

<div type='discussions-op-text'>

<sup>Originally posted by **msavin** October 15, 2022</sup>
A lot of times, I have like 5 Meteor applications running. It's just easier to keep them all running than to constantly start and terminate servers. However, often times, I'll forget the port on which the apps are running.

I'm curious if anyone else runs into this situation. I thought a simple solution can be to indicate the server port on app restart

From:
```
=> Meteor server restarted
```

To:
```
=> Meteor server restarted on port 3400
```</div>

ltsaiete:
Hello @Grubba27 
I'm not familiar with meteor, but I'm very familiar with JS/TS and I want to contribute for Hacktoberfest. I find this issue interesting and friendly. I'd like to try.

StorytellerCZ:
@ltsaiete go ahead, try out Meteor following the [development guide](https://github.com/meteor/meteor/blob/devel/DEVELOPMENT.md) and se if you can make this happen.

ltsaiete:
Hello @ankitzm , I'm still working on it.

shashwata27:
is this resolved?

afrokick:
Released in 2.8.1. The issue should be closed. @Grubba27 @denihs

Grubba27:
Done in 2.8.1

afrokick:
@Grubba27 could you please change milestone?)

PR #12236, opened 2022-10-20
Show port in restart message

harryadel:
As suggested in https://meteor-community.slack.com/archives/CN350MY1G/p1665817956720209


harryadel:
@Grubba27 There was another issue with fetching from https://hello.myfonts.net/count/2c4b9d so I grabbed an older version and saved it locally.

Grubba27:
Actually I was just solving that matter @harryadel I've placed here that same code as well https://css-cdn2.vercel.app/font

harryadel:
@Grubba27 It's a good solution but keeping it locally ensures no reliance on third party service.

denihs (approved):

packages/non-core/less/tests/top.import.less
```diff
@@ -6,4 +6,4 @@
 
 // Make sure regular CSS import doesn't make the compiler explode - this was
 // a regression from 1.1.0.3 caught in QA
-@import url("http://hello.myfonts.net/count/2c4b9d");
+@import url("./hello-myfonts.css");
```
zodern, line 9: It seems this test was for how it handles urls that start with http:// or https://. If the old url doesn't work, it would be good to replace it with one that starts the same.

harryadel:
Does sourcing it locally defeats the purpose of the test?

Grubba27:
Hey @harryadel, maybe create a separated pr for solving the ``@import url`` issue. 
That would be awesome. 
I think @zodern has a point of keeping the use of HTTPS URL. 
Maybe keeping that commented code in a separate file and linking this pr for docs purposes the same way you did

radekmie:
1. I second @Grubba27 on making a separate PR for the LESS stuff.
2. As I said in [#12226](https://github.com/meteor/meteor/discussions/12226#discussioncomment-3893640), it'd be better to have the full `ROOT_URL` there (it's also mentioned in #12230).

harryadel:
@radekmie @zodern I removed the css changes and used rootUrl to fully display app URL.

@Grubba27 You may now create a new PR with the [vercel link](https://css-cdn2.vercel.app/font).

PR #12237, opened 2022-10-20
Implemented async Tracker with generators.

radekmie:
In this pull request, I implemented a way to make `Tracker.autorun` work with async operations without any additional API surface, Zone.js, and any other _magic_ at a cost of... Well, using generators. _(I don't find it a problem, but most people are not experienced with them and find them rather intimidating.)_

Here's an example of how it works:
```ts
Tracker.autorun(function* (computation) {
  const result = yield Promise.resolve(123);
  assert(result === 123);
});
```

In other words: use `yield` instead of `await` and it's all good :rocket:

But it's not all that magical. The list of drawbacks is short, but it's important to understand where it comes from:
1. The `Tracker.currentComputation` (the core of `Tracker` reactivity) is preserved between `yield`s, but NOT between `await`s and `then`s. It's because the iterator operator (the `_withStep` function) sort of manages the `Tracker.currentComputation` "right before" and "right after" the `yield`. In other words, the following test passes:
    ```ts
    Tracker.autorun(function* (computation) {
      assert(Tracker.currentComputation === computation);
      yield Promise.resolve().then(() => {
        assert(Tracker.currentComputation === null);
      });
      assert(Tracker.currentComputation === computation);
    });
    ```
2. The `computation.firstRun` is `true` only during the first run until the first `yield`. The reason is exactly the same as with the `Tracker.currentComputation`. In other words, the following test passes:
    ```ts
    Tracker.autorun(function* (computation) {
      assert(computation.firstRun === true);
      yield Promise.resolve().then(() => {
        assert(computation.firstRun === false);
      });
      assert(computation.firstRun === false);
    });
    ```
    I _think_ we could make it work, but it'd involve more complexity.
3. It's now possible to trigger a single computation multiple times at once (see the `tracker - generator function parallel` test for an example). I _think_ we could queue them instead, but that's debatable how it should work exactly.
4. `Tracker.autorun` returns as soon as the function finishes or the first `yield` happens. That's because when the generator yields, we create a promise and wait for it "in the background", not blocking anything. We _could_ add a `isRunning: boolean` and `firstRunPromise: Promise<void>` to the computation object if needed.
5. It's not really possible to type it in TypeScript, i.e., all `yield`s return `any` _or_ the same type.

Anyway, it's just an experiment! Let me know what do you think!

### DIFF
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -83,6 +83,32 @@ function _throwOrLog(from, e) {
   }
 }
 
+// Does one step of a generator. Anything the generator yields will be awaited,
+// because we chain it into the promise.
+function _computeStep(computation, iterator, result) {
+  return result.done
+    ? Promise.resolve()
+    : Promise.resolve(result.value).then(
+        value => _computeStep(computation, iterator, _withComputation(computation, () => iterator.next(value))),
+        error => _computeStep(computation, iterator, _withComputation(computation, () => iterator.throw(error)))
+      );
+}
+
+function _withComputation(computation, f) {
+  var previousComputation = Tracker.currentComputation;
+  var previousInCompute = inCompute;
+
+  setCurrentComputation(computation);
+  inCompute = true;
+
+  try {
+    return withNoYieldsAllowed(f)();
+  } finally {
+    setCurrentComputation(previousComputation);
+    inCompute = previousInCompute;
+  }
+}
+
 // Takes a function `f`, and wraps it in a `Meteor._noYieldsAllowed`
 // block if we are running on the server. On the client, returns the
 // original function (since `Meteor._noYieldsAllowed` is a
@@ -300,15 +326,12 @@ Tracker.Computation = class Computation {
   _compute() {
     this.invalidated = false;
 
-    var previous = Tracker.currentComputation;
-    setCurrentComputation(this);
-    var previousInCompute = inCompute;
-    inCompute = true;
-    try {
-      withNoYieldsAllowed(this._func)(this);
-    } finally {
-      setCurrentComputation(previous);
-      inCompute = previousInCompute;
+    // Run the autorun function. It has to happen within the context of current
+    // computation. If it returns something that looks like a generator (i.e.,
+    // it has a `next` function), we'll continue running it with `_computeStep`.
+    var maybeIterator = _withComputation(this, () => this._func(this));
+    if (typeof maybeIterator?.next === 'function') {
+      _computeStep(this, maybeIterator, _withComputation(this, () => maybeIterator.next()));
     }
   }
 
@@ -597,13 +620,7 @@ Tracker.autorun = function (f, options) {
  * @param {Function} func A function to call immediately.
  */
 Tracker.nonreactive = function (f) {
-  var previous = Tracker.currentComputation;
-  setCurrentComputation(null);
-  try {
-    return f();
-  } finally {
-    setCurrentComputation(previous);
-  }
+  return _withComputation(null, f);
 };
 
 // http://docs.meteor.com/#tracker_oninvalidate
--- a/packages/tracker/tracker_tests.js
+++ b/packages/tracker/tracker_tests.js
@@ -524,7 +524,11 @@ Tinytest.addAsync('tracker - async function', function (test, onComplete) {
   const computation = Tracker.autorun(async function (computation) {
     test.equal(computation.firstRun, true, 'before (firstRun)');
     test.equal(Tracker.currentComputation, computation, 'before');
-    const x = await Promise.resolve(123);
+    const x = await Promise.resolve().then(() => {
+      test.equal(computation.firstRun, false, 'inside (firstRun)');
+      test.equal(Tracker.currentComputation, null, 'inside');
+      return 123;
+    });
     test.equal(x, 123, 'await (value)');
     test.equal(computation.firstRun, false, 'await (firstRun)');
     test.equal(Tracker.currentComputation, null, 'await');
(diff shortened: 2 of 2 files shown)


Grubba27:
Hey, @zodern, could Reify or Babel do the compiler job of turning this code into the generator  code ?

Writing this:
```js
Tracker.autorun(async function (computation) {
  const result = await Promise.resolve(123);
  assert(result === 123);
});
```

Would turn into this:

```js
Tracker.autorun(function* (computation) {
  const result = yield Promise.resolve(123);
  assert(result === 123);
});
```
I'm not soo familiar with the babel API / Reify, and if it would be possible to turn some more complex ASTs into the one above like this scenario

```js
const handleTracker = async (comp) => {
  const result = await Promise.resolve(123);
  assert(result === 123);
}
Tracker.autorun(handleTracker);
```
Would be like this: 

```js
function* handleTracker(comp) => {
  const result = yield Promise.resolve(123);
  assert(result === 123);
}

Tracker.autorun(handleTracker);
```

about point 3:
> It's now possible to trigger a single computation multiple times at once (see the tracker - generator function parallel test for an example). I think we could queue them instead, but that's debatable how it should work exactly.

I think it may be better for them to run in parallel; what would benefit from having a queue? @radekmie

radekmie:
@Grubba27 Even if you'd transform `async`/`await` into `*`/`yield`, it still won't work with nested promises. I think we could patch the `Promise` object itself, but I'm not sure how, whether it won't kill the performance, and if it'll work with `async`/`await` syntax.

Regarding running them in parallel, it's all about the expected behavior. Let's say we have the following code:
```ts
const activeUsers = new ReactiveVar();
Tracker.autorun(function* () {
  Session.get('anything'); // Any dependency really.
  const users = yield getActiveUsers(); // Async operation.
  activeUsers.set(users);
});

let userId = 0;
async function getActiveUsers() {
  await new Promise(resolve => setTimeout(resolve, 1000 * Math.random));
  return [{ _id: ++userId }];
}
```

This code is not deterministic, i.e., the `activeUsers[0]?._id` can change randomly, as `getActiveUsers` may take different amount of time on every call. Having a queue would make it predictable.

denihs:
My main concern here is the introduction of generators... As you said, I also think most people are not experienced with them (myself included! I never had the need to use them for real).

I was hoping @Grubba27's idea of using Reify/babel to transform the syntax from async/await to generators could work.

Then again, if the community is ok with using generators, we can follow this approach.

But we should have good documentation around this, with lots of examples and scenarios where this won't work.

Grubba27:
> Even if you'd transform async/await into */yield, it still won't work with nested promises. 

What is that scenario @radekmie? 
you mean if they use ``yield Promise.all([])`` ?
or do you mean that the ``.then(() => promise.then( promise2.then()))``

Grubba27:
Now I see it clearly. I would vote for making it more deterministic as well given that input.

CaptainN:
I haven't reviewed the implementation, but I like the idea of elevating generators - I've always wanted a practical reason to use them! I don't think we should use babel to hide the generators though - if we need them, just require users to use them.

Julusian:
From seeing this PR and the one for `Meteor.callAsync`, it sounds like the biggest issues facing async in the frontend are race conditions because they rely on a global object, and another computation/calls can now modify that global object while it is executing.

It looks like these attempts to avoid changing the api are introducing weird limitations or complexity. 
Does this outweigh the cost of a new api designed for async? 
I am curious as to what a new api designed for this might look like.

zodern:
I agree that it is cool to see a use for generators.

I've considered using compilers before, and another challenge is there is no reliable way of knowing what code will run within a Tracker computation. The changes the compiler does would have to apply to the whole app or package instead of just the functions that run within a Tracker computation.

>  I think we could patch the Promise object itself, but I'm not sure how, whether it won't kill the performance, and if it'll work with async/await syntax.

I haven't looked into this in depth, but I've been wondering lately if this would be another option. We would need to replace the Promise global for code that directly uses promises. The challenge with async/await is it always uses native promises instead of  the Promise global, so we would need a simple compiler to modify async functions. Meteor already does this on the server, but the implementation would have to be different on the client since fibers aren't available. 

Performance would be a concern, though it should be more efficient than how promises currently work on the server. Another concern would be how much this increases the client bundle size. We probably won't know if these are problems until we try it.

veered:
I think drawback #1 that @radekmie mentioned in the PR description more or less rules out this approach as it is currently implemented. Drawback #1 means that `yield` can't call async functions that interact with Tracker.

The most common side-effect of this would be missing reactive dependencies, but it could also result in spawning an infinite number of autoruns.

For example, this code will print `hello` twice instead of just once because the inner autorun is never killed:
```js
let someVar = new ReactiveVar('');

let asyncFunc = async () => {
  await new Promise(resolve => setTimeout(resolve, 0));
  Tracker.autorun(() => console.log(someVar.get()));
};

let comp = Tracker.autorun(function* () {
  yield asyncFunc();
});

comp.invalidate();
someVar.set('hello');
```
The inner autorun doesn't have access to `Tracker.currentComputation` so it thinks it's a top-level autorun.

Here's an example of under-reactivity:
```js
let getTotal = async () => {
  let countA = await CollectionA.find().countAsync(),
      countA = await CollectionB.find().countAsync()
  ;
  return countA + countB;
};

Tracker.autorun(function* () {
  console.log(yield getTotal());
});
```
The autorun will rerun when `CollectionA` changes but not when `CollectionB` changes.

radekmie:
I have another idea that wouldn't use generators, and would work with `async`/`await` but is very verbose... I need a couple of days to play with it, but expect another draft _soon_.

radekmie:
I created https://github.com/meteor/meteor/pull/12294 with the alternative approach and invite everyone to comment on there!

radekmie:
I'm closing this in favor of #12294.

PR #12252, opened 2022-10-21
[mongo] Remove underscore

harryadel:


radekmie:
_(I'd like to comment on all of these PRs at once, but there's not a single place to do it, so I'll do it here.)_

While I really like the idea of getting rid of underscore, I'm not so fond of replacing it with tens of `lodash.x` packages. The benefit of a single `underscore` package is that there's only a single copy of it in the entire app. And correct me if I'm wrong (@zodern?), but the npm dependencies of Meteor packages are **not** deduplicated.

Another thing is, the sheer amount of these PRs is not a problem, but the amount of places that have to be changed when an update happens, in my opinion, is. Instead of doing **one** update, we have to update tens of packages. And not only one of their dependencies but possibly a few (7 in this PR).

I'd like to hear what you think about it @harryadel, @StorytellerCZ, @Grubba27.

StorytellerCZ:
I think this separation is already the best it can be. Better than the previous where it was just one big PR, now it is at least thematic and often by packages. I think we can release these changes as a feature update so then we don't have to worry about multiple packages as they will be part of the next Meteor overall release.

I'm not fond of still having some `lodash.x` packages replacing the functionality, but I think it is better than having the entire library. So for starters we should start by merging those PRs where there is no need to include the lodash npm package to replace its functionality. As for those where there is need for them I would hold of on those to see if we can't figure out a better solution. This way we at least reduce the total number of PRs.

Grubba27:
Totally agree with @StorytellerCZ. I'm merging those who do not have external deps. I'm thinking of using the deprecated/loadash or creating a vendorized package for internal use only with those functions  especially ``has`` and  ``indentity`` that are used in many packages out there

radekmie:
> I'm not fond of still having some `lodash.x` packages replacing the functionality, but I think it is better than having the entire library.

I'd rather have the entire library and import only pieces of it. I don't know about you, but in our projects we always do `import x from 'lodash/x'` and it _just works_. I don't know if it can be achieved from an atmosphere package, though.

> I'm thinking of using the deprecated/loadash or creating a vendorized package for internal use only with those functions especially `has` and `indentity` that are used in many packages out there.

That's a very good idea! Especially that it can actually re-export just a couple of `lodash/x` imports.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -1210,15 +1217,15 @@ if (Meteor.isServer) {
...
+    test.equal(Object.keys(o.state).length, 2);
     test.equal(o.state[docId3], {_id: docId3, y: 7222});
     test.equal(o.state[docId1], {_id: docId1, y: 1222});
     clearOutput(o);
 
     onComplete();
   });
 
-  Tinytest.add("mongo-livedata - observe sorted, limited, big initial set" + idGeneration, function (test) {
+  Tinytest.addAsync("mongo-livedata - observe sorted, limited, big initial set" + idGeneration, function (test, onComplete) {
```
radekmie, line 1228: Where is that `Async` change coming from?

packages/mongo/mongo_livedata_tests.js
```diff
@@ -1510,7 +1520,8 @@ testAsyncMulti('mongo-livedata - transform sets _id if not present, ' + idGenera
   function (test, expect) {
     var self = this;
     var justId = function (doc) {
-      return _.omit(doc, '_id');
+      const {_id, ...rest} = doc;
```
radekmie, line 1523: It'd be better not to create such unused variables. Instead, you could do a shallow copy and `delete` it.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -2504,6 +2515,67 @@ testAsyncMulti('mongo-livedata - empty string _id', [
 
 
 if (Meteor.isServer) {
+
+  testAsyncMulti("mongo-livedata - minimongo on server to server connection", [
```
radekmie, line 2519: This looks like some code that got removed already.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -3105,17 +3177,17 @@ testAsyncMulti("mongo-livedata - oplog - update EJSON", [
 ]);
 
 
-function waitUntilOplogCaughtUp() {
+var waitUntilOplogCaughtUp = function () {
```
radekmie, line 3180: I guess that's also a part of the old code. Please revert it.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -3232,12 +3304,8 @@ Meteor.isServer && testAsyncMulti("mongo-livedata - update with replace forbidde
 Meteor.isServer && Tinytest.add(
   "mongo-livedata - connection failure throws",
   function (test) {
-    // Exception happens in 30s
     test.throws(function () {
-      const connection = new MongoInternals.Connection('mongodb://this-does-not-exist.test/asdf');
-
-      // Same as `MongoInternals.defaultRemoteCollectionDriver`.
-      Promise.await(connection.client.connect());
```
radekmie, line 3240: Same here.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -3437,7 +3505,7 @@ if (Meteor.isServer) {
 }
 
 if (Meteor.isServer) {
-  Tinytest.addAsync("mongo-livedata - transaction", function (test) {
+  Tinytest.addAsync("mongo-livedata - transaction", function (test, onComplete) {
```
radekmie, line 3508: And here.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -1510,7 +1518,9 @@ testAsyncMulti('mongo-livedata - transform sets _id if not present, ' + idGenera
   function (test, expect) {
     var self = this;
     var justId = function (doc) {
-      return _.omit(doc, '_id');
+      const docWithoutId = doc;
```
radekmie, line 1521: This should clone the `doc` too.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -1712,7 +1722,7 @@ if (Meteor.isServer) {
       {all: 1, id1Direct: 1, id1InQuery: 1, id2Direct: 1, id2InQuery: 1,
        bothIds: 1});
 
-    _.each(handlesToStop, function (h) {h.stop();});
+       handlesToStop.forEach(function (h) {h.stop();});
```
radekmie, line 1725: Preserve the indent here.

packages/mongo/mongo_livedata_tests.js
radekmie: It's still a problem.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -3495,4 +3504,4 @@ if (Meteor.isServer) {
       });
     });
   });
-}
+}
```
radekmie, line 3507: I wouldn't change unnecessary lines.

radekmie (changes requested):
I found a couple more issues. And if possible, I'd love to make as few unrelated changes here as possible, i.e., get rid of whitespace-only changes.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -864,16 +868,18 @@ if (Meteor.isServer) {
...
-    return _.isEmpty(_.difference(a, b)) && _.isEmpty(_.difference(b, a));
+    a = a.map(EJSON.stringify);
+    b = b.map(EJSON.stringify);
+    return isEmpty(difference([a, b])) && isEmpty(difference([b, a]));
   };
 
   // This test mainly checks the correctness of oplog code dealing with limited
   // queries. Compitablity with poll-diff is added as well.
-  Tinytest.add("mongo-livedata - observe sorted, limited " + idGeneration, function (test) {
+  Tinytest.addAsync("mongo-livedata - observe sorted, limited " + idGeneration, function (test, onComplete) {
```
radekmie, line 882: This test should stay synchronous.

radekmie (changes requested):
I think this is the last one.

@Grubba27, @henriquealbert, @StorytellerCZ I still would rather have a single `lodash` package and direct imports, but I agree that removing `underscore` is a benefit on its own.

@zodern Are npm dependencies shared between packages? We'll end up with 40 copies of different lodash helpers in the bundle 🤷

zodern:
> @zodern Are npm dependencies shared between packages? We'll end up with 40 copies of different lodash helpers in the bundle 

No, npm dependencies are not shared between packages. The idea above of having a Meteor package with a currated set of lodash or underscore functions sounds like a good option. Though it might be good to consider making it a public package instead of being treated as an internal package since some non-core packages might want to also use it instead of having to add their own copy of the helpers.

Underscore now allows importing the specific modules you want. One option would be to have an `underscore-minimal` package with just the helpers that are used by core packages, and then an `underscore` package that re-exports it, and also exports all of the remaining helpers. This way there is no duplicated code between them for apps that still use underscore.

harryadel:
It's done! :feelsgood: 
cc @StorytellerCZ @radekmie

packages/mongo/mongo_driver.js
```diff
@@ -135,12 +137,12 @@ var replaceTypes = function (document, atomTransformer) {
...
 
   var ret = document;
-  _.each(document, function (val, key) {
+  Object.entries(document).forEach(function ([key, val]) {
     var valReplaced = replaceTypes(val, atomTransformer);
     if (val !== valReplaced) {
       // Lazy clone. Shallow copy.
       if (ret === document)
-        ret = _.clone(document);
+        ret = clone(document);
```
jamauro, line 145: Could the spread operator be used instead of lodash's `clone`?

harryadel: I did try it but it didn't work :shrug:

radekmie: That's weird. Can you try `Object.assign({}, document)` instead then? The less lodash the better.

packages/mongo/mongo_driver.js
```diff
@@ -1032,9 +1034,10 @@ Cursor.prototype.observeChanges = function (callbacks, options = {}) {
 };
 
 MongoConnection.prototype._createSynchronousCursor = function(
-    cursorDescription, options) {
+    cursorDescription, options = {}) {
   var self = this;
-  options = _.pick(options || {}, 'selfForIteration', 'useTransform');
+  const { selfForIteration, useTransform } = options;
```
radekmie, line 1039: You could unpack this directly in the arguments list.

packages/mongo/remote_collection_driver.js
```diff
@@ -75,4 +76,4 @@ MongoInternals.defaultRemoteCollectionDriver = _.once(function () {
   });
 
   return driver;
-});
+});
```
radekmie, line 79: Can you make sure not to remove those? Next person editing this file will add it again.

packages/mongo/mongo_livedata_tests.js
```diff
@@ -2411,56 +2420,56 @@ function functionChain2Upsert (test, expect, coll, index) {
...
+for (var j = 0; j < collectionCount; j++) {
+ fn(test, expect, this.collections[j], i);
+}
+}
+}, function (test, expect) {
+// Run any registered cleanup functions (e.g. to drop collections)
+this.cleanups.forEach(function(cleanup) {
+cleanup(expect);
+});
+}]);
```
radekmie, line 2472: Can you fix the formatting here?

packages/mongo/mongo_driver.js
harryadel: But `options` is reused later on
https://github.com/harryadel/meteor/blob/fd745ffb3f759144292038434a093acbfa162f52/packages/mongo/mongo_driver.js#L1087

packages/mongo/mongo_driver.js
radekmie: OK, cool, but I truly don't understand why. The CI is canceled - what was the error?

radekmie (approved):

PR #12261, opened 2022-10-22
[ddp-client] Remove underscore

harryadel:


StorytellerCZ:
If the tests pass then this should be fine to merge as the underscore is only in tests.

denihs (approved):

denihs:
Nice, we'll include it in the next release!

harryadel:
Actually I was going to invest more time into it and use even less lodash libraries if possible. :sweat_smile: So next time :crossed_fingers:

PR #12265, opened 2022-10-22
[browser-policy-content] Remove underscore

harryadel:


StorytellerCZ (approved):

harryadel:
I don't get it, so now we're not supposed to use ES6 in packages in order to ensure it runs on legacy browsers?? Doesn't `ecmascript` package take care of that? @StorytellerCZ @zodern

StorytellerCZ:
@harryadel I'm as confused as you are. 😅 @Grubba27 can you enlighten us?

Grubba27:
> @harryadel I'm as confused as you are. 😅 @Grubba27 can you enlighten us?

Sure, we have been working to make Meteor compatible with E11.
I thought as well that adding ecmascript should make the whole module work with legacy browsers

harryadel:
This package only contains `modules` package and not the `ecmascript` package, should we add it alongside the es6 changes? 

https://github.com/meteor/meteor/blob/6872d21d4e0d221b7157057f003e470af96a0dc3/packages/browser-policy-content/package.js#L7

Grubba27:
Try adding the ecmascript! maybe it will make things okay

Grubba27:
test failed(travis said is something with browser policy)

harryadel:
https://github.com/meteor/meteor/assets/6644545/3a2e00a2-49a7-49a8-b683-48b2c084392e

*sigh* ok, I'll look into it.

StorytellerCZ (approved):

wreiske:
Happy to see underscore being removed! Is there a need to use lodash here?

This PR uses lodash functions `isEmpty`, `has`, and `union`. These can be replaced with native JavaScript functions to improve efficiency and reduce dependencies.

1. (Faster) Replace `lodash.isEmpty` with native JavaScript:
   The `lodash.isEmpty` function checks if a value is an empty object, collection, map, or set. Objects are considered empty if they have no own enumerable string keyed properties. In JavaScript, you can use `Object.keys(obj).length === 0 && obj.constructor === Object` to check if an object is empty.

https://www.measurethat.net/Benchmarks/Show/26277/2/lodash-isempty-vs-native-javascript

```
_.isEmpty         | _.isEmpty x 3,948,866 ops/sec ±0.35% (66 runs sampled)
Native JavaScript | Native JavaScript x 4,647,284 ops/sec ±0.22% (69 runs sampled)
```

2. (Faster) Replace `lodash.has` with native JavaScript:
   The `lodash.has` function checks if `path` is a direct property of `object`. In JavaScript, you can use the `in` operator or `Object.prototype.hasOwnProperty.call()` to check if an object has a specific property.

https://www.measurethat.net/Benchmarks/Show/26278/0/lodash-has-vs-native-javascript

```
lodash.has        | lodash.has x 3,970,107 ops/sec ±0.39% (68 runs sampled)
Native JavaScript | Native JavaScript x 9,363,523 ops/sec ±2.34% (66 runs sampled)
```

3. (Slightly slower) Replace `lodash.union` with native JavaScript:
   The `lodash.union` function creates an array of unique values, in order, from all given arrays. In JavaScript, you can use the `Array.prototype.concat()` method combined with `new Set()` to achieve the same result.

https://www.measurethat.net/Benchmarks/Show/26279/0/lodash-union-vs-native-javascript

```
lodash.union | lodash.union x 1,908,308 ops/sec ±1.00% (67 runs sampled)
Native JavaScript | Native JavaScript x 1,877,701 ops/sec ±0.41% (65 runs sampled)
```

Here are the replacements:

1. Replace `isEmpty(cspSrcs)` with `(Object.keys(cspSrcs).length === 0 && cspSrcs.constructor === Object)`.
2. Replace `has(cspSrcs, directive)` with `(directive in cspSrcs)`.
3. Replace `union(sources || [], cspSrcs["default-src"] || [])` with `[...new Set([...(sources || []), ...(cspSrcs["default-src"] || [])])]`.

harryadel:
@wreiske I initially had an aversion to implementing ES5 custom function even if performant as this would open up the door to opinionated ways of solving the problems and our biggest concern here is to remove underscore regardless of how pretty the solution is. But with such a well thought and well written answer, how could I refuse? :sweat_smile: 

Couple of notes though:
- isEmpty was used twice, once for an [object](https://github.com/meteor/meteor/blob/cb576fb4b22539599bd0d64b94d1576d144338e4/packages/browser-policy-content/browser-policy-content.js#L165) and the other for an [array](https://github.com/meteor/meteor/blob/cb576fb4b22539599bd0d64b94d1576d144338e4/packages/browser-policy-content/browser-policy-content.js#L175). 
-  Using `new Set` couldn't be used due to `Check legacy syntax` CI step so I had to come up with another solution.

PR #12266, opened 2022-10-22
[boilerplate-generator] Remove underscore

harryadel:


StorytellerCZ (changes requested):
I think we can bring in _.escape in as a whole instead of as a dependency:
https://github.com/lodash/lodash/blob/main/src/escape.ts

Thought that is to be considered what is better solution? @Grubba27 If we are happy with outsourcing the toolbelt for now then as is is fine. But in the long run, we might want to have a central utilities package or something like that which exports things like _.escape while centralizes their dependency managements.

Grubba27:
I Agree @StorytellerCZ

denihs:
@harryadel, could you please give this PR a description? What's the goal you're trying to achieve with these changes? Why these changes are better than the current implementation?

I know you're doing this to reduce bundle size, but it's good to have a description either way.

denihs:
Also, I'm not so sure about this type of change... Will we reimplement all these functions from _?

It's more code for us to maintain, more code to understand, and how much data are we saving by doing this type of change?

Right now, @StorytellerCZ is suggesting creating a utility package, so won't we end up having the same _ library in our source code if we keep this up?

I'm not killing the idea, just bringing up this discussion.

harryadel:
@denihs @StorytellerCZ Thank you guys for taking note and bringing this up. During the initial work on this PR, for some reason I thought the MDG implementation of _.template was different than underscore's which prompted me to copy over the code from Meteor's package but upon further inspection this doesn't seem to be the case so I removed code and opted to use https://www.npmjs.com/package/lodash.template instead.

StorytellerCZ:
@denihs I'm suggesting that we have a utility package to centralize utility code across multiple packages and import any lodash packages that we need in there and export them through there, so that we have one central place to update them from instead of having to do those updates in 10 or so packages when update comes along.

harryadel:
@Grubba27 merge this one too?

Grubba27:
I agree on the utilities packages.
For now we should merge this one and later on centralize those dependencies

PR #12267, opened 2022-10-22
[boilerplate-generator-tests] Remove underscore

harryadel:


packages/boilerplate-generator-tests/web.cordova-tests.js
```diff
@@ -60,9 +59,7 @@ Tinytest.addAsync(
   async function (test) {
     const newHtml = await generateHTMLForArch('web.cordova', false);
 
-    _.templateSettings = {
```
StorytellerCZ, line 63: This being removed makes me feel that we are missing here something even though the tests are passing.

StorytellerCZ (commented):
I think a bit more investigation on this one is warranted.

packages/boilerplate-generator-tests/web.cordova-tests.js
harryadel: It's used within tests and doesn't affect the module functionality. If the tests fail, I'll re-look into it.

leonardoventurini (approved):

PR #12268, opened 2022-10-22
Move underscore & underscore-tests to deprecated directory

harryadel:
This should be merged last after all underscore PRs are merged in.


packages/deprecated/underscore-tests/package.js
```diff
@@ -2,7 +2,8 @@ Package.describe({
   // These tests can't be directly in the underscore packages since
   // Tinytest depends on underscore
   summary: "Tests for the underscore package",
-  version: '1.0.8'
+  version: '1.0.8',
```
StorytellerCZ, line 5: We will have to bump the version. If by patch or feature bump that is for the Meteor team to decide.

StorytellerCZ (changes requested):

packages/deprecated/underscore-tests/package.js
Grubba27: Since it does not affect any functionality, I would place this as a patch. We do have a script that updates SemVer versions automatically you can check it [here](https://github.com/meteor/meteor/blob/93b765659fe2e562a37b6626be27774534e557c7/scripts/admin/update-semver/index.js) :) 
before releasing it I will do a check where it needs to be updated thx for bringing that out

StorytellerCZ (approved):

leonardoventurini (approved):

PR #12274, opened 2022-10-24
Feat user accounts base async

Grubba27:
We have an option of adding ``Meteor.userAsync()`` wich is not reactive ~yet~ because of tracker and keeping both. If you are using them in stateless calls things would look the same.

---
Also, made ``Meteor.loginWithPassword`` accept async values as callbacks

## TODO

 - [x] Decide what to do with ``Meteor.userId()`` as it is sync because the id is in the [method_invocation.js#constructor](https://github.com/meteor/meteor/blob/ef8796a44eaff508c285aa499ff7ce8f1dde2b05/packages/ddp-common/method_invocation.js#L48)
> There was nothing to decide, as since the userId is provided went instantiating that class.

#### Context
for those who wants to know where it gets the userId field it does this trip:

[Meteor.userId](https://github.com/meteor/meteor/blob/222022ecf0dde1d5ad35aed1ce223aa02c6e4041/packages/accounts-base/accounts_client.js#L46) -> [method_invocation.js#constructor](https://github.com/meteor/meteor/blob/ef8796a44eaff508c285aa499ff7ce8f1dde2b05/packages/ddp-common/method_invocation.js#L48) -> [accounts_common.js#_initConnection](https://github.com/meteor/meteor/blob/33015f00d1d00383012a35cb0e46bd84c3338e1d/packages/accounts-base/accounts_common.js#L330)

### DIFF
--- a/packages/accounts-base/accounts_client_tests.js
+++ b/packages/accounts-base/accounts_client_tests.js
@@ -95,6 +95,20 @@ Tinytest.addAsync(
 );
 
 Tinytest.addAsync(
+  'accounts async - Meteor.loggingIn() is false after login has completed',
+  (test, done) => {
+    logoutAndCreateUser(test, done, () => {
+      // Login then verify loggingIn is false after login has completed
+      Meteor.loginWithPassword(username, password, async () => {
+        test.isTrue(await Meteor.userAsync());
+        test.isFalse(Meteor.loggingIn());
+        removeTestUser(done);
+      });
+    });
+  }
+);
+
+Tinytest.addAsync(
   'accounts - Meteor.loggingOut() is true right after a logout call',
   (test, done) => {
     logoutAndCreateUser(test, done, () => {
@@ -150,7 +164,7 @@ Tinytest.addAsync(
 );
 
 Tinytest.addAsync(
-  'accounts - Meteor.user obeys explicit and default field selectors',
+  'accounts - Meteor.user() obeys explicit and default field selectors',
   (test, done) => {
     logoutAndCreateUser(test, done, () => {
       Meteor.loginWithPassword(username, password, () => {
@@ -178,6 +192,37 @@ Tinytest.addAsync(
   }
 );
 
+Tinytest.addAsync(
+  'accounts async - Meteor.userAsync() obeys explicit and default field selectors',
+  (test, done) => {
+    logoutAndCreateUser(test, done, () => {
+      Meteor.loginWithPassword(username, password, () => {
+        // by default, all fields should be returned
+        Meteor.userAsync().then(user => {
+          test.equal( user.profile[excludeField], excludeValue);
+        });
+
+        // this time we want to exclude the default fields
+        const options = Accounts._options;
+        Accounts._options = {};
+        Accounts.config({defaultFieldSelector: {['profile.'+defaultExcludeField]: 0}});
+        Meteor.userAsync().then(async user => {
+          test.isUndefined(user.profile[defaultExcludeField]);
+          test.equal(user.profile[excludeField], excludeValue);
+          test.equal(user.profile.name, username);
+        })
+        // this time we only want certain fields...
+        Meteor.userAsync({fields: {'profile.name': 1}}).then(user => {
+          test.isUndefined(user.profile[excludeField]);
+          test.isUndefined(user.profile[defaultExcludeField]);
+          test.equal(user.profile.name, username);
+        });
+        Accounts._options = options;
+        removeTestUser(done);
+      });
+    });
+  }
+);
 
 Tinytest.addAsync(
   'accounts-2fa - Meteor.loginWithPasswordAnd2faCode() fails when token is not provided',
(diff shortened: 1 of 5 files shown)


packages/accounts-base/accounts_common.js
```diff
@@ -170,6 +170,18 @@ export class AccountsCommon {
...
   }
 
+  /**
+   * @summary Get the current user record, or `null` if no user is logged in.
+   * @locus Anywhere
+   * @param {Object} [options]
+   * @param {MongoFieldSpecifier} options.fields Dictionary of fields to return or exclude.
+   */
+  async userAsync(options) {
+    const userId = this.userId();
```
henriquealbert, line 180: is this going to change? I think it will be sync as well?
```
    const userId = this.userId();
```

henriquealbert (approved):

packages/accounts-base/accounts_common.js
Grubba27: No, userId is set on the DDP connection. So there is no need of making this async as it has on memory

packages/accounts-base/accounts_client_tests.js
```diff
@@ -94,6 +94,20 @@ Tinytest.addAsync(
...
 );
 
+Tinytest.addAsync(
+  'accounts async - Meteor.loggingIn() is false after login has completed',
+  (test, done) => {
+    logoutAndCreateUser(test, done, () => {
+      // Login then verify loggingIn is false after login has completed
+      Meteor.loginWithPassword(username, password, async () => {
+        test.isTrue(await Meteor.userAsync());
+        test.isFalse(Meteor.loggingIn());
```
radekmie, line 104: I'd rather swap these two, as if the `userAsync` may cause an additional wait, `loggingIn` can "flicker" but this test will still pass.

packages/accounts-base/accounts_client_tests.js
```diff
@@ -178,6 +192,37 @@ Tinytest.addAsync(
   }
 );
 
+Tinytest.addAsync(
+  'accounts async - Meteor.userAsync() obeys explicit and default field selectors',
+  (test, done) => {
+    logoutAndCreateUser(test, done, () => {
+      Meteor.loginWithPassword(username, password, () => {
+        // by default, all fields should be returned
+        Meteor.userAsync().then(user => {
```
radekmie, line 201: It'd be good to `await` all of these promises, as right now there's no way of making sure the test actually passes.

packages/accounts-base/accounts_client_tests.js
```diff
@@ -178,6 +192,37 @@ Tinytest.addAsync(
...
+          test.equal( user.profile[excludeField], excludeValue);
+        });
+
+        // this time we want to exclude the default fields
+        const options = Accounts._options;
+        Accounts._options = {};
+        Accounts.config({defaultFieldSelector: {['profile.'+defaultExcludeField]: 0}});
+        Meteor.userAsync().then(async user => {
```
radekmie, line 209: `async` is not needed here.

packages/accounts-base/accounts_common.js
```diff
@@ -170,6 +170,18 @@ export class AccountsCommon {
       : null;
   }
 
+  /**
+   * @summary Get the current user record, or `null` if no user is logged in.
+   * @locus Anywhere
+   * @param {Object} [options]
+   * @param {MongoFieldSpecifier} options.fields Dictionary of fields to return or exclude.
+   */
+  async userAsync(options) {
```
radekmie, line 179: Maybe it's a personal preference, but the less `async` functions the better. Wrapping `null` in `Promise.resolve` would have the same effect and generate less code when downleveled.

packages/accounts-base/accounts_tests.js
```diff
@@ -604,6 +604,62 @@ Tinytest.add(
...
+
+    // test the field is included by default
+    let user = await Meteor.userAsync();
+    test.isNotUndefined(user[ignoreFieldName], 'included by default');
+
+    // test the field is excluded
+    Accounts.config({ defaultFieldSelector: { [ignoreFieldName]: 0 } });
+    user = await Meteor.userAsync();
+    test.isUndefined(user[ignoreFieldName], 'excluded');
+    user = Meteor.user({});
```
radekmie, line 632: Is this `user` and not `userAsync` by design?

packages/accounts-base/accounts_common.js
Grubba27: Yeah.
This is more of a preference; having the async keyword in the function signature seems easier to know that it will return a Promise<Some> doing this at the end of the function (wrapping with Promise.resolve) in this case would not be a problem for legibility.
If you think this is a point that we need to discuss further to make it consistent around the codebase, I will be really happy to do it

radekmie: Sure, got that. I just still live in the word of extremely expensive `async`/`await` downleveling.

radekmie (approved):

PR #12275, opened 2022-10-25
In the client, don't wait if the stub doesn't return a promise

denihs:
As suggested by @zodern [here](https://github.com/meteor/meteor/pull/12057/files#r996452189), we don't need to `await` if the stub is sync.


Grubba27 (approved):
LGTM!

henriquealbert (approved):

PR #12279, opened 2022-10-27
MongoID._looksLikeObjectID bugfix

vlasky:
str.match() returns an array on success which is not coerced into boolean true. An explicit not null check is now performed instead.

### DIFF
--- a/packages/mongo-id/id.js
+++ b/packages/mongo-id/id.js
@@ -3,7 +3,7 @@ import { Random } from 'meteor/random';
 
 const MongoID = {};
 
-MongoID._looksLikeObjectID = str => str.length === 24 && str.match(/^[0-9a-f]*$/);
+MongoID._looksLikeObjectID = str => str.length === 24 && (str.match(/^[0-9a-f]*$/) != null);
 
 MongoID.ObjectID = class ObjectID {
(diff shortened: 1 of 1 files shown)


vlasky:
For the record, I discovered this bug in mongo-id while doing my initial testing of [redis-oplog](https://github.com/cult-of-coders/redis-oplog) with [oplogtoredis](https://github.com/vlasky/oplogtoredis). redis-oplog was configured with the following options:

```
"redisOplog": {
    "redis": {
        "port": 6379,
        "host": "127.0.0.1"
    },
    "mutationDefaults": {
        "optimistic": true
    },
    "globalRedisPrefix": "meteorapp.",
    "externalRedisPublisher": true
}
```

During this testing I discovered this bug as well as some other ObjectID JS type mismatch bug that I never got to the bottom of. I managed to bypass the buggy code path in redis-oplog/mongo-id by setting `"optimistic": false`, which turns off optimistic UI. This prevents redis-oplog from overriding the mutator in Mongo.Collection, which results in the ObjectID's JS type being correct.

radekmie (approved):

radekmie:
@Grubba27 Can we squeeze it in 2.8.1?

PR #12286, opened 2022-11-03
fix a few typos in tools directory

striezel:


StorytellerCZ (approved):
Comments changes, with one exception in `tools/cli/help.txt` which will show text in console. Good to see those typos fixed.

striezel:
Yes. I probably should have mentioned that (except for one file) only comments are affected, so those changes shouldn't break anything.

PR #12294, opened 2022-11-04
Implemented async Tracker with explicit values.

radekmie:
As said in https://github.com/meteor/meteor/pull/12237#issuecomment-1291580636, I had a plan to make `Tracker` work with `async` functions without generators (#12237) nor any extra magic, but with a rather explicit API. That's what this pull request is doing.

Let's see an example:
```ts
Tracker.autorun(async function computation => {
  const countA = await Tracker.withComputation(computation, () => CollectionA.find().countAsync());
  const countB = await Tracker.withComputation(computation, () => CollectionB.find().countAsync());
});
```

How it works? The `Tracker.withComputation` sets the `Tracker.currentComputation` just like `Tracker.nonreactive` "unsets" it. And as the `autorun` function unsets it on the first `await`, you have to call `Tracker.withComputation` whenever you want the reactivity to work. In the above example, the `find` function is a reactive consumer, so it's wrapped. But then the `countAsync` returns a promise, so we have to `await` it.

I honestly don't _really_ like this approach, though. While it _works_, it's just hard to make sure it's all correct. It also requires quite a lot of work to get the `computation` around. Of course, you can use the `Tracker.currentComputation` to get it "out of the thin air", but it's still a burden.

Nevertheless, let me hear what you think about it.


Grubba27:
it does sparkle my eyes (I always dreamed of using ``createEffect`` with an async function)
Are there any limitations to this approach?
Every promise-like function/call/object should be wrapped in a ``() => Promise``?

If there are none, I would argue that using babel to transform this:
```js
Tracker.autorun(async computation => {
  const countA = await CollectionA.find().countAsync();
  const countB = await CollectionB.find().countAsync();
});
```
into: 
```js
Tracker.autorun(async computation => {
  const countA = await Tracker.withComputation(computation, () => CollectionA.find().countAsync());
  const countB = await Tracker.withComputation(computation, () => CollectionB.find().countAsync());
});
```
It would be fairly easy, and people would enjoy this.

Tho In complex cases, I think we should let the user use this primitive

radekmie:
Yes, it'd work, but only for "shallow" cases. If we'd take the cases from https://github.com/meteor/meteor/pull/12237#issuecomment-1287338199, it'd look like this:

```ts
let someVar = new ReactiveVar('');

let asyncFunc = async () => {
  const computation = Tracker.currentComputation;
  await new Promise(resolve => setTimeout(resolve, 0));
  Tracker.withComputation(computation, () => {
    Tracker.autorun(() => console.log(someVar.get()));
  });
};

let comp = Tracker.autorun(async function () {
  await asyncFunc();
});

comp.invalidate();
someVar.set('hello');
```

```ts
let getTotal = async () => {
  let countA = await CollectionA.find().countAsync();

  const computation = Tracker.currentComputation;
  await new Promise(resolve => setTimeout(resolve, 0));

  let countB = Tracker.withComputation(
    computation,
    () => CollectionB.find().countAsync(),
  );

  return countA + countB;
};

Tracker.autorun(async function () {
  console.log(await getTotal());
});
```

_(I didn't test it though.)_

Of course, we can see a pattern here, and it could get another helper:

```ts
function continueWithCurrentComputation(promise, fn) {
  const computation = Tracker.currentComputation;
  return promise.then(value => Tracker.withComputation(computation, fn(value)));
}
```

```ts
let someVar = new ReactiveVar('');

let asyncFunc = async () => {
  await continueWithCurrentComputation(
    new Promise(resolve => setTimeout(resolve, 0)),
    () => Tracker.autorun(() => console.log(someVar.get())),
  );
};

let comp = Tracker.autorun(async function () {
  await asyncFunc();
});

comp.invalidate();
someVar.set('hello');
```

```ts
let getTotal = async () => {
  let countA = await CollectionA.find().countAsync();
  let countB = await continueWithCurrentComputation(
    new Promise(resolve => setTimeout(resolve, 0)),
    () => CollectionB.find().countAsync(),
  );

  return countA + countB;
};

Tracker.autorun(async function () {
  console.log(await getTotal());
});
```

Some will find it easier, some won't. I'm not a fan of such a helper either (yet?).

_(Again, I didn't test it.)_

Torgen:
Instead of being a global on Tracker you have to pass the computation to, could the public interface be a method on the computation itself? `computation.await(() => ...)` or something?

Torgen:
Notwithstanding my previous comment, I feel like allowing arbitrary work in a computation is pretty suspicious. Should a template helper be able to `fetch()`? What we really want is for the async minimongo methods to preserve context, right? Would it suffice to have those return a context-preserving Promise implementation, without attempting to clobber the global Promise constructor?

Torgen:
> > Instead of being a global on Tracker you have to pass the computation to, could the public interface be a method on the computation itself? `computation.await(() => ...)` or something?
> 
> Yes, it's definitely doable. The exact API is an open topic - the goal of this PR is just to present an alternative approach with explicit computation tracking.
> 
> > Notwithstanding my previous comment, I feel like allowing arbitrary work in a computation is pretty suspicious. Should a template helper be able to `fetch()`?
> 
> I think it should. Especially in Blaze, it's common to do things like `{{#each getDocuments}}`, where `getDocuments` is a helper. Sure, it _could_ return a cursor instead, but it may require additional, query from other collection (e.g., to get the preferred sorting order).
When I said `fetch()` I meant the XmlHTTPRequest replacement, not converting a cursor to an array.
> 
> > What we really want is for the async minimongo methods to preserve the computation, right?
> 
> We want that too, for sure. But the problem is slightly broader - make all asynchronous computation work in reactive contexts (i.e., `Tracker.autorun`).
> 
> > Would it suffice to have those return a computation-preserving Promise implementation, without attempting to clobber the global Promise constructor?
> 
> Yes, but it has a lot of drawbacks. From problems like `result instanceof Promise` returning `false`, to breaking everything with `async`/`await` (it may convert it back into the builtin `Promise`). Overall, I don't think it's viable.

You can extend Promise, which would make `instanceof` work, but it doesn't help with `async`/`await`, where the function that gets passed to `then()` seems to queue another microtask rather than being the continuation of the `async` function itself.

Grubba27:
I agree on a lot with @Torgen about using computation as the api, and I'm getting to like this solution a lot.

```js
const countA = await computation.run(CollectionA.find().countAsync);
const externalFetch = await computation.run(externalFetch);
```
Sounds really neat and, I guess, could be a really loved feature for Blaze & Meteor codebases in general

I would argue that this one is easier to use than the global Tracker. And being less verbose in this case is good, I guess

radekmie:
> When I said `fetch()` I meant the XmlHTTPRequest replacement, not converting a cursor to an array.

Right. So I don't see a need for `fetch` in particular, but I do see a benefit of being able to use async functions. For example, you'd have some helper functions, shared between client and server code. To make it possible, these have to use async MongoDB API, as it's the only way to make it work on the server. And not to duplicate the code, there's a need to `await` these helpers.

> As I think about this more, it's not like Minimongo is moving to IndexedDB or something where all operations have to be async. It's still an in-memory database. Obviously the sync methods have to be removed from the server because Fibers are dying, and async methods have to be added to the client so method simulation can work, but is there any reason sync methods can't stay on the client for helpers?

The most important part is the possibility to share the client and server code, just as I wrote above. A secondary benefit is to be able to use _all_ APIs instead of only synchronous ones.

Torgen:
Reusing async functions between the client and the server isn't going to work anyways, at least with this implementation, because you need to explicitly preserve the computation in every nested await, but there is no computation on the server side.

radekmie:
> Reusing async functions between the client and the server isn't going to work anyways, at least with this implementation, because you need to explicitly preserve the computation in every nested await, but there is no computation on the server side.

It should, but you will simply preserve "nothing". It won't if it'd become a computation method instead.

Torgen:
Suppose I have the following method that you want to call in client and server code:
```
/** @locus Anywhere */
async function foo(...) {
  const a = await CollectionA.findOneAsync(...);
  // do something with a
  return CollectionB.findOneAsync(...);
}
```
Are you saying I would modify this method to have `continueWithCurrentComputation` even though it can run on the server? Is `continueWithCurrentComputation` going to have locus Anywhere when all the other computation functions have locus Client? This seems like you're creating a sharp edge, because you're giving the impression that you can use any async function in a computation and thread the computation through it.
I would almost say that the answer is to make the `reactive` option on mongo/minimongo methods able to take a computation instead of just a boolean. Then I can at least have all my mongo-using utilities take an extra options parameter which I can splat into the method options, and it also works for allowing the function to be used server-side inside a transaction (although bhunjadi's addon is a more convenient way of doing that).

radekmie:
> Are you saying I would modify this method to have `continueWithCurrentComputation` even though it can run on the server?

With the approach of this PR yes, if you'd like to use it in `Tracker.autorun`.

> Is `continueWithCurrentComputation` going to have locus Anywhere when all the other computation functions have locus Client?

AFAIK other than `Meteor._noYieldsAllowed` (which can be taken care of), nothing is stopping you from using `Tracker` on the server.

> This seems like you're creating a sharp edge, because you're giving the impression that you can use any async function in a computation and thread the computation through it.

No, it's not an impression - that's exactly the thing this PR aims to solve.

> I would almost say that the answer is to make the `reactive` option on mongo/minimongo methods able to take a computation instead of just a boolean. Then I can at least have all my mongo-using utilities take an extra options parameter which I can splat into the method options, and it also works for allowing the function to be used server-side inside a transaction [...].

That's a neat idea! I have to think about it though, as it's more impactful change for the MongoDB integration.

> (although bhunjadi's addon is a more convenient way of doing that)

I guess you're talking about [bhunjadi/meteor-mongo-transactions](https://github.com/bhunjadi/meteor-mongo-transactions)? It relies on Fibers (`Promise.await`) for some things, but this could be changed. However, it also relies on `Meteor.EnvironmentVariable` which cannot be preserved in asynchronous contexts on the client (people tried with Zone.js in #12176, but it didn't work out).

Julusian:
> I would almost say that the answer is to make the reactive option on mongo/minimongo methods able to take a computation instead of just a boolean. Then I can at least have all my mongo-using utilities take an extra options parameter which I can splat into the method options, and it also works for allowing the function to be used server-side inside a transaction [...].
>
> That's a neat idea! I have to think about it though, as it's more impactful change for the MongoDB integration.

I was thinking this too, having it as an extra property will really simplify usage in a lot of cases. Accepting `computation` as a parameter should be encouraged for any other reactive apis too (either builtin or community made).

In addition, it would be nice for meteor to export a const `NonReactiveComputation` which can be used anywhere `computation` is needed, when reactivity is not wanted. I expect that `run()` on `NonReactiveComputation` would be a no-op. 
This would be used for any helper methods which need a computation, when reactivity is not wanted.
This could be done purely in user-land, but it will help with documentation to have it as an export

```
// reactive function used by both server and client
function reactiveTest(computation) {
  const countA = await CollectionA.find({}, { computation }).countAsync()
  const externalFetch = await computation.run(fetch('http://example.com/api/thing));
  ...
}

// could be called from server code:
reactiveTest(NonReactiveComputation)
```

radekmie:
> My point is that you can't trust that wrapping an arbitrary async function will preserve the computation even within the function you've wrapped. (In my `foo()` example, the lookup on collection B won't be reactive.)

Sure, it'll work only if you handle it. If you won't, it won't.

> I didn't mean that what `meteor-mongo-transactions` does on the server is something we could do on the client; I meant that in isolation it's more convenient than what I'm suggesting on the server to enable collection operations within function calls to contribute to transactions.

I'm not sure I got that. Could you paraphrase it, please?

> I was thinking this too, having it as an extra property will really simplify usage in a lot of cases. Accepting `computation` as a parameter should be encouraged for any other reactive apis too (either builtin or community made).

Sure, that's why I like it.

> In addition, it would be nice for meteor to export a const `NonReactiveComputation` which can be used anywhere `computation` is needed, when reactivity is not wanted.

I think it'd be better to treat it as an optional parameter instead - just like MongoDB does with the `session`.

Grubba27:
Hey @radekmie me, and @denihs discussed this, and we would like to move with this solution because it will be simpler IMO for external users + beginners to understand and use the ``Tracker``
I would favor this one for its simplicity in many ways, and because it would be a standard way of solving this matter (not having two ways of dealing with reactive values, using ``computation. resolve(async fn)`` or ``await CollectionA.find({}, { computation }).countAsync()``  )

to summarize, I think just having a method for the computation parameter that comes from the tracker autorun field sounds good.

I would be in favor of ``computation.run`` or ``.resolve``instead of using the global one(``Tracker.withComputation()``)

What is your opinion about this?

zodern:
@Grubba27 I'm curious what your thoughts are on #12317. It could allow tracker to work with async functions without the user having to add any extra code.

radekmie:
@Grubba27 Got it. I wouldn't add `computation.run`, because the user would have to do some workarounds if the computation is not there (e.g., in a helper that can be executed with and without a computation; `Tracker.withComputation` handles that just fine).

@zodern As I understand, a Babel plugin could use `Tracker.withComputation` too, right?

zodern:
> @zodern As I understand, a Babel plugin could use Tracker.withComputation too, right?

It probably would be possible. There is no way to know which async functions will run within a tracker computation, so it would have to modify all of them. Besides being a little simpler, it seems the compiled result wouldn't be much different from #12317. I would prefer if we went with a more general solution so it could also be used for other things, such as removing the limitations around async method stubs.

Once you consider code that is after the first `await` and isn't being awaited, it becomes more difficult. For example:

```
async function helpers() {
  // The babel plugin can easily wrap this with Tracker.withComputation
  const tasks = await Tasks.find().fetchAsync();
  // After the await, the current computation is lost
  // Wrapping sync code like this so it runs within the correct computation would be more difficult
  const userId = Meteor.userId();
}
```
After solving this, it wouldn't be any simpler than #12317.

Grubba27:
I thought that ``computation.run`` was the same as ``Tracker.withComputation`` but with that being said, and given your input, the second seems better.

I did a small look into it before @zodern but today I gave a closer look, and I loved it, but I'm a little bit worried that we miss some edge case.

But I would like to hear you both, my main concern is the end user is having some weird issues because of some edge case and asking the end user to understand/ use the ``Promise.wrapAsync`` sounds rought. For that reason, I would favor ``Tracker.withComputation``. 

If it can be assured that all cases/ the edge cases that could break are solved with your solution @zodern then I would be _really_ in favor of having things like this one done in the build/compile time are better IMO

zodern:
I created that proof of concept in an hour, so there are many things I probably haven't considered yet or tested. For example, I realized afterwards that it currently doesn't correctly handle async functions that don't await, though that is simple to fix.

At this point, the only thing I know it won't always handle is callbacks passed to npm dependencies (if passed to app or meteor package code, they might work more reliably). We already have this limitation on the server, so the same solutions there can be used on the client, for example wrapping the callback with `Meteor.bindEnvironment`.

The goal is to match how the Meteor environment currently works on the server as closely as possible, so while the compiled code is a little complicated, what it is doing should match what Meteor developers expect.

Regarding edge cases, after only spending an hour creating it I don't have a good answer for this. If we want to consider this approach, the next steps probably are:
1. Do more testing, possibly setting up fuzzy testing to try various scenarios
2. Update the implementation to work correctly when promises are rejected
3. Have it patch the Promise global to preserve the context for code using promises directly
4. Run benchmarks to see how much overhead it adds to promises and async/await
5. Create a simple compiler, and test it in some real apps to see how much it affects bundle size

radekmie:
@Grubba27 I updated this branch with the latest `devel` as soon as the CI passes, it's good to go.

PR #12295, opened 2022-11-04
Implemented Fibers-less MongoDB count methods.

radekmie:
Just as announced at my Meteor Impact 2022 presentation, I wanted to bring the Meteor's Collections API on par with the native MongoDB driver. It's rather natural and doesn't cause any problems, as the method names _mostly_ don't overlap anymore (e.g., `insertOne` instead of `insert`).

In this pull request, I add both [`countDocuments`](https://mongodb.github.io/node-mongodb-native/4.11/classes/Collection.html#countDocuments) and [`estimatedDocumentCount`](https://mongodb.github.io/node-mongodb-native/4.11/classes/Collection.html#estimatedDocumentCount). ([Yes, both `collection.count()` and `cursor.count()` calls are deprecated.](https://www.mongodb.com/docs/v4.4/reference/command/count/)) On the server side, it uses the MongoDB driver directly, without any Meteor magic, and no Fibers whatsoever. On the client side, it's as simple as a call to `countAsync`.

**NOTE:** It's not ready to merge just yet, as it requires some documentation, but the code is there.


Grubba27 (approved):
Does look good

Grubba27:
Do want to squeeze this into 2.8.1?

denihs:
It looks more of a minor change. I think it should go on 2.9.

radekmie:
Yeah, I'd also go with 2.9. There's also no documentation for it _yet_, and I probably won't find time for that in the next few days.

radekmie:
I added some documentation, but I'm not sure how to tackle it in general. Meteor never really listed all options in the docs, and now there's just way too many to do that. There's also a few client-specific ones too. For now I added only some short block and would like to hear what you think about it.

denihs:
I have nothing against your approach here. 

I think we just need to add those [here](https://docs.meteor.com/api/collections.html#Mongo-Cursor-count) with a little description of the methods, and we're good to go.

Maybe even add a "(deprecated)" to the count method?

radekmie:
@denihs I added a short note about `Cursor#count` being deprecated and used the `@deprecated` JSDoc. If you think I should add more, let me know.

Grubba27:
Hey @radekmie can I merge this to 2.9?

denihs:
I think it's good to go now @Grubba27

radekmie:
Yeah, it's good to go.

PR #12298, opened 2022-11-07
(feat): Generate scaffold in cli

Grubba27:
This is a feature for 2.9, I've made it pointing to 2.8.1, in time I will adjust the bases.
---

In this PR I focus on the creation of a new CLI command called generate, that will for the app user generate a scaffold API for him, this gives to the end user a little more time and a better DX since, he will be not needing to write this code that almost every MeteorJS RPC API has:
here is an example of usage. (in a simple js app)

```bash
meteor generate customer
```

it will generate the following code in ``/imports/api``
Screenshot 2022-11-09 at 11 28 29

that will have the following code:

<details>
<summary><h3>collection.js</h3></summary>

```js

export const CustomerCollection = new Mongo.Collection('customer');

```
</details>

<details>
<summary><h3>methods.js</h3></summary>

```js
import { check } from 'meteor/check';
import { CustomerCollection } from './collection';

export function create(data) {
  return CustomerCollection.insertAsync({ ...data });
}

export function update(_id, data) {
  check(_id, String);
  return CustomerCollection.updateAsync(_id, { ...data });
}

export function remove(_id) {
  check(_id, String);
  return CustomerCollection.removeAsync(_id);
}

export function findById(_id) {
  check(_id, String);
  return CustomerCollection.findOneAsync(_id);
}

Meteor.methods({
  'Customer.create': create,
  'Customer.update': update,
  'Customer.remove': remove,
  'Customer.find': findById
});

```
</details>

<details>

<summary><h3>publication.js</h3></summary>

```js

import { CustomerCollection } from './collection';

Meteor.publish('CustomersByLoggedUser', function publishCustomersByUserId() {
  return CustomerCollection.find({ userId: this.userId });
});

Meteor.publish('allCustomers', function publishCustomers() {
  return CustomerCollection.find({});
});


```

</details>

<details>

<summary><h3>index.js</h3></summary>

```js

export * from './collection';
export * from './methods';
export * from './publications';

```

</details>

Also, there is the same version of these methods using TypeScript.

### Path option

for those that may want to create in another path, you can use the ``--path`` option in order to select where to place this boilerplate.
It will generate the model in that path.

```bash

meteor generate another-customer --path=server/admin

```

it will generate in ``server/admin`` the another-client code:

Screenshot 2022-11-09 at 11 32 39

<details>
<summary><h3>collection.ts</h3></summary>

```typescript

export type AnotherCustomer = {
  _id?: string;
  name: string;
  createdAt: Date;
}

export const AnotherCustomerCollection = new Mongo.Collection<AnotherCustomer, AnotherCustomer>('another-customer');

```
</details>

<details>
<summary><h3>methods.ts</h3></summary>

```typescript

import { check } from 'meteor/check';
import { AnotherCustomer, AnotherCustomerCollection } from './collection';

export function create(data: AnotherCustomer) {
  return AnotherCustomerCollection.insertAsync({ ...data });
}

export function update(_id: string, data: Mongo.Modifier<AnotherCustomer>) {
  check(_id, String);
  return AnotherCustomerCollection.updateAsync(_id, { ...data });
}

export function remove(_id: string) {
  check(_id, String);
  return AnotherCustomerCollection.removeAsync(_id);
}

export function findById(_id: string) {
  check(_id, String);
  return AnotherCustomerCollection.findOneAsync(_id);
}

Meteor.methods({
  'AnotherCustomer.create': create,
  'AnotherCustomer.update': update,
  'AnotherCustomer.remove': remove,
  'AnotherCustomer.find': findById
});


```
</details>

<details>

<summary><h3>publications.ts</h3></summary>

```typescript

import { AnotherCustomerCollection } from './collection';

Meteor.publish('AnotherCustomersByLoggedUser', function publishAnotherCustomersByUserId(this) {
  return AnotherCustomerCollection.find({ userId: this.userId });
});

Meteor.publish('allAnotherCustomers', function publishAnotherCustomers() {
  return AnotherCustomerCollection.find({});
});

```

</details>

<details>

<summary><h3>index.ts</h3></summary>

```typescript

export * from './collection';
export * from './methods';
export * from './publications';

```

</details>

---
<details>

<summary><h1> Using the Wizard  </h1></summary>

if you run the following command:

```bash
meteor generate
```

it will prompt the following questions. _I've answered them enter for all of them_

Screenshot 2022-11-09 at 11 38 29

</details>

--- 

<details>

<summary><h1> Using your own template </h1></summary>

Screenshot 2022-11-09 at 11 42 47

> Note that I'm not trying to create a CLI framework inside meteor but just giving some solutions for really common problems out of the box

You can use your own templates for scaffolding your specific workloads. To do that, you should pass in a template directory URL so that it can copy it with its changes.

### how do I rename things?

Out of the box I provide a few functions such as replacing ``$$name$$``, ``$$PascalName$$`` and ``$$camelName$$`` 

these replacements come from this function: 

```js
const transformName = (name) => {
    return name.replace(/\$\$name\$\$|\$\$PascalName\$\$|\$\$camelName\$\$/g, function (substring, args) {
      if (substring === '$$name$$') return scaffoldName;
      if (substring === '$$PascalName$$') return toPascalCase(scaffoldName);
      if (substring === '$$camelName$$') return toCamelCase(scaffoldName);
    })
  }
```

### What if I want to have my own way of templating?

There is an option called ``--replaceFn`` that when you pass in given a .js file with two functions it will override all templating that we have defaulted to use your given function.
_example of a replacer file_
```js
export function transformFilename(scaffoldName, filename) {
  console.log(scaffoldName, filename);
  return filename
}

export function transformContents(scaffoldName, contents, fileName) {
  console.log(fileName, contents);
  return contents
}

```
if you run your command like this:

```bash
 meteor generate feed --replaceFn=/fn/replace.js
```
it will generate files full of ``$$PascalCase$$``using the meteor provided templates.

A better example of this feature would be the following js file (like we do inside the cli):
```js
const toPascalCase = (str) => {
  if(!str.includes('-')) return str.charAt(0).toUpperCase() + str.slice(1);
  else return str.split('-').map(toPascalCase).join('');
}
const toCamelCase = (str) => {
  if(!str.includes('-')) return str.charAt(0).toLowerCase() + str.slice(1);
  else return str.split('-').map(toPascalCase).join('');
}

const transformName = (scaffoldName, str) => {
    return str.replace(/\$\$name\$\$|\$\$PascalName\$\$|\$\$camelName\$\$/g, function (substring, args) {
      if (substring === '$$name$$') return scaffoldName;
      if (substring === '$$PascalName$$') return toPascalCase(scaffoldName);
      if (substring === '$$camelName$$') return toCamelCase(scaffoldName);
    })

}

export function transformFilename(scaffoldName, filename) {
  return transformName(scaffoldName, filename);
}

export function transformContents(scaffoldName, contents, fileName) {
  return transformName(scaffoldName, contents);
}
```
</details>


denihs (approved):

tools/static-assets/scaffolds-js/methods.js
```diff
@@ -0,0 +1,25 @@
...
+  return await $$UpperName$$Collection.updateAsync(_id, { ...data });
+}
+
+export const remove$$UpperName$$ = async (_id) => {
+  return await $$UpperName$$Collection.removeAsync(_id);
+}
+
+export const find$$UpperName$$ById = async (_id) => {
+  return await $$UpperName$$Collection.findOneAsync(_id);
+}
```
radekmie, line 18: 1. Are these "officially recommended" names? _(I personally prefer `create` instead of `save` only because it has the same length as `remove` and `update`.)_
2. Have you thought about using `async function`s instead? _(Again, personal preference, but Meteor codebase is highly scattered in this manner. They also could use `this` for the method context.)_
3. Do you think there's a need to `return` in `update` and `remove`? I know what these functions do and I use their returned values rarely.
4. Have you thought about adding validation here? They're exposed via `Meteor.methods`, so...

tools/static-assets/scaffolds-js/index.js
```diff
@@ -0,0 +1,15 @@
...
+  find$$UpperName$$ById
+} from './methods';
+
+export {
+  $$UpperName$$Collection,
+  save$$UpperName$$,
+  update$$UpperName$$,
+  remove$$UpperName$$,
+  find$$UpperName$$ById
+}
```
radekmie, line 15: Have you thought about using `export * from` instead? It's less of a burden to update it later.

tools/static-assets/scaffolds-js/methods.js
henriquealbert: I agree with points 2 and 4 🙌🏻

tools/static-assets/scaffolds-js/index.js
henriquealbert: Agreed, and also it's cleaner

tools/static-assets/scaffolds-js/methods.js
Grubba27: > I personally prefer create instead of save only because it has the same length as remove and update.)
Now you have triggered me in a good way :), I've never thought about this! 
After some thought, I would agree. In other frameworks, they call it by create as well in reference to the CRUD acronym


what do you mean by:
> Have you thought about adding validation here? They're exposed via Meteor.methods, so...
adding some validation with the check package?

Grubba27: Also giving some more tought on to it. Since I'm exporting this using module I could make method.js like this:

```js

import { CustomerCollection } from './collection';

export function create(data) {
  return CustomerCollection.insertAsync({ ...data });
}

export function update(_id, data) {
  return CustomerCollection.updateAsync(_id, { ...data });
}

export function remove(_id) {
  return CustomerCollection.removeAsync(_id);
}

export function findById(_id) {
  return CustomerCollection.findOneAsync(_id);
}

Meteor.methods({
  'Customer.create': create,
  'Customer.update': update,
  'Customer.remove': remove,
  'Customer.find': findById
});

```

because today the use of those methods along the module import would be:

```js
import { Customer } from './customer';


Customer.findCustomerById('someId')
```
using the code above would look like this:

```js
import { Customer } from './customer';


Customer.findById('someId')
Customer.create({...data})
Customer.someOtherMethod()
```

tools/static-assets/scaffolds-js/index.js
```diff
@@ -1,15 +1,7 @@
 import { $$UpperName$$Collection } from './collection';
```
henriquealbert, line 1:
```suggestion
export * from './collection';
```

tools/static-assets/scaffolds-js/index.js
```diff
@@ -1,15 +1,7 @@
...
-import {
-  save$$UpperName$$,
-  update$$UpperName$$,
-  remove$$UpperName$$,
-  find$$UpperName$$ById
-} from './methods';
+import * as $$UpperName$$ from './methods';
```
henriquealbert, line 2:
```suggestion
export * from './methods';
```

tools/static-assets/scaffolds-js/index.js
```diff
@@ -1,15 +1,7 @@
...
 
 export {
   $$UpperName$$Collection,
-  save$$UpperName$$,
-  update$$UpperName$$,
-  remove$$UpperName$$,
-  find$$UpperName$$ById
+  $$UpperName$$
```
henriquealbert, line 6: and remove this

tools/static-assets/scaffolds-js/methods.js
radekmie: Then if you'd export all of the names "unprefixed" (including the collection object itself), the following code would work:

```ts
import * as Customer from './customer';

Customer.collection;
Customer.create({ ... });
// And so on.
```

But again, it's a matter of preference and what's "officially recommended".

tools/static-assets/scaffolds-ts/methods.ts
```diff
@@ -0,0 +1,25 @@
+import { Meteor } from 'meteor/meteor';
+import { $$UpperName$$Type, $$UpperName$$Collection } from './collection';
+
+export function create$$UpperName$$(data: $$UpperName$$Type) {
+  return $$UpperName$$Collection.insertAsync({ ...data });
+}
+
+export function update$$UpperName$$(_id: string, data: Partial<$$UpperName$$Type>) {
+  return $$UpperName$$Collection.updateAsync(_id, { ...data });
+}
```
radekmie, line 10: Actually, `data` should use `UpdateFilter` and not `Partial`, as right now it does a full document replace as well as doesn't support operators.

tools/static-assets/scaffolds-ts/collection.ts
```diff
@@ -0,0 +1,9 @@
+import { Mongo } from 'meteor/mongo';
+
+export type $$UpperName$$Type = {
+  _id?: string;
+  name: string;
+  createdAt: Date;
+}
+
+export const $$UpperName$$Collection = new Mongo.Collection('$$name$$');
```
afrokick, line 9: We use a generic for Mongo.Collection like 

```
export const $$UpperName$$Collection = new Mongo.Collection<$$UpperName$$Type, $$UpperName$$Type>('$$name$$');
```

tools/static-assets/scaffolds-ts/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./collection";
```
afrokick, line 1: maybe we should adjust code style for using one type of quotes for import/export? single or double?

tools/static-assets/scaffolds-ts/methods.ts
Grubba27: Do you mean ``Mongo.Modifier<$$UpperName$$>``?

radekmie (approved):

fredmaiaarantes:
Could we have other `generate` commands?
Like one for creating subscriptions for a specific collection @Grubba27?

Grubba27:
You would say something like this @fredmaiaarantes ?
```js
function customerPublication() {
  return TasksCollection.find({  });
}
Meteor.publish('customer', customerPublication);
```

faburem:
I think this could be great for new users to get started with some of the MeteorJS specifics they are not familiar with (e.g. methods). Since this is already pretty opinionated I am wondering if it wouldn't make sense to go one step further and include a basic check or at least a comment indicating that method parameters should be validated? Otherwise I see a bit of a risk that scaffolded code like this is going straight into production without any data validation.

tools/static-assets/scaffolds-js/methods.js
radekmie: Also, I forgot to answer you 🤦 Thank you @faburem for reminding me!

> adding some validation with the check package? or something else?

I'd say `check` is enough.

faburem: I agree, check would be good enough just to have something in place which is showing the best practice in this regard.

StorytellerCZ:
Agree with the addition of check. It would be my step to add immediately anyway. Also agree with @fredmaiaarantes that there should be more templates and even one extra plus would be if there would be some option to create new ones for projects via some simple template that can be included with a project, that would be pretty awesome.

Grubba27:
I've addressed those comments (Added check and possibility to use your own templates).
Also, I've updated the docs.
I'm not trying to replace a full-blown CLI framework like inquerer or yargs but for those projects that are just starting and more things out of the box

Grubba27:
Hey @StorytellerCZ check if those changes that I've made makes sense to you
I've written in the pr description the general idea, and this will probably be the docs when this goes live.
I think the API naming is still up for debate, calling  it by ``replaceFn`` sounds good but I'm not sure

StorytellerCZ (approved):
I think these are good initial steps.

docs/source/commandline.md
```diff
@@ -184,6 +184,327 @@ Create a basic [solid](https://www.solidjs.com/) app.
...
+
+
+
+<h3 id="meteorgenerate-publication.js">publication.js</h3>
+
+```js
+
+import { Meteor } from 'meteor/meteor';
+import { CustomerCollection } from './collection';
+
```
fredmaiaarantes, line 262: This generated publication might be a little confusing as the `customer` collection might not have a `userId`.

Grubba27: you think is best to just leave the ``all$$CollectionName$$`` ?

fredmaiaarantes: I guess so. But if you want to keep it, no problem.

fredmaiaarantes (approved):

henriquealbert (approved):

Issue #12307, opened 2022-11-18
During installing meteor globally I've got "Error: Cannot find module 'https-proxy-agent'"


unsigned6:
I was installing globally latest meteor version and received missing `'https-proxy-agent'` error (using nodejs v14.19.3 npm v6.14.17). Previously it was working and started to fail in our github actions yesterday. Maybe something broken with packages list in latest version?

```npm install -g meteor                                                                                                                                                                    13:15:48
/Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/bin/meteor-installer -> /Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor/cli.js

> meteor@2.8.1 install /Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor
> node cli.js install

internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module 'https-proxy-agent'
Require stack:
- /Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor/install.js
- /Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor/cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor/install.js:2:25)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor/install.js',
    '/Users/yuriivlasiuk/.nvm/versions/node/v14.19.3/lib/node_modules/meteor/cli.js'
  ]
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
[...]

PR #12308, opened 2022-11-18
[meteor-installer] fix: add a missing dep

afrokick:
An issue occurs starting from 2.8.1

Fix https://github.com/meteor/meteor/issues/12307

Also, require only if it needs.

A discussion: https://github.com/meteor/meteor/pull/12149#pullrequestreview-1185887093


afrokick:
@Grubba27 @denihs

denihs:
Hi, thanks for the PR. A new version was created and it should be all good now!

PR #12309, opened 2022-11-18
[meteor] chore: improve eslint config

afrokick:
**Goal:** Improve code-style consistency

1. relax some rules for ESLint
2. add `check-code-style` workflow to automatically run ESLint on picked packages
3. apply it to `npm-packages/meteor-installer` as an example

Any ideas what we should do next?


afrokick:
`eslint "./packages/**/*.js" --no-ignore`

The result:
```
✖ 386295 problems (276440 errors, 109855 warnings)
  224114 errors and 105087 warnings potentially fixable with the `--fix` option.
```

😄

Grubba27:
Liked this _a lot_. I'm in on this. Will ask @radekmie, that asked me these days about having a linter in the project; I guess we can go doing this project by project

henriquealbert:
This is great @afrokick! ❤️

afrokick:
@Grubba27 @henriquealbert thanks for your support! I know that there are a lot of more important things to bring Meteor 3.0 but some additional rules can improve codebase quality right now.

radekmie (commented):
I'm not sure how to proceed here. I've been migrating _huge_ projects to ESLint (either from nothing or JSHint), and we usually did one of two things:
1. "Stopped the world", i.e., ran `--fix` and finished all of the other issues right now. Sure, it took a while, but then none of the PRs had a problem.
2. Marked all of the rules as warnings and improved them as we went. I think it was easier for the team, but it was not uncommon to see an extremely simple and small PR change hundreds of lines just because of ESLint (most people have problems with disabling it temporarily).

Therefore, I'm not sure what the plan should be here. I think it's simply impossible to do the former, but at the same time, I'm afraid of the latter's flood of PRs. I'd go with the latter anyway.

---

A separate topic is a config itself. `@quave/eslint-config-quave` bases on ESLint v5 ([source](https://github.com/quavedev/eslint-config/blob/master/package.json#L11)) and should get an update (we're at 8.28 now).

Secondly, Prettier should also be configured because even if it's an ESLint plugin, some tools rely on its config (or lack of one).

Lastly, There's no TypeScript-related config in it at all. It's not a deal breaker, but there's not a short list of things that will cause issues, e.g., "unused" type imports.

That's why I'd rather recommend creating a Meteor-specific config and using that instead. Sure, let's base on Quave's one, but I'd also propose Vazco's one ([eslint-config-vazco](https://github.com/vazco/eslint-config-vazco)) as it includes TypeScript configuration and bases on much newer ESLint version (8.25).

afrokick:
> Therefore, I'm not sure what the plan should be here. I think it's simply impossible to do the former, but at the same time, I'm afraid of the latter's flood of PRs. I'd go with the latter anyway.

Another approach is to specify exactly paths for linting. At the beginning, we can lint only `tools` or `static` and then expand it to `packages` and `npm-packages` for example. To move step by step.

> A separate topic is a config itself. `@quave/eslint-config-quave` bases on ESLint v5 ([source](https://github.com/quavedev/eslint-config/blob/master/package.json#L11)) and should get an update (we're at 8.28 now).
> Prettier should also be configured because even if it's an ESLint plugin, some tools rely on its config (or lack of one).

+1

> There's no TypeScript-related config in it at all. It's not a deal breaker, but there's not a short list of things that will cause issues, e.g., "unused" type imports.

We have the tsconfig in tools, but we can move it to the root and change some props.

> That's why I'd rather recommend creating a Meteor-specific config and using that instead. Sure, let's base on Quave's one, but I'd also propose Vazco's one ([eslint-config-vazco](https://github.com/vazco/eslint-config-vazco)) as it includes TypeScript configuration and bases on much newer ESLint version (8.25).

Like it. Good start point. I can replace quave to vazco.

@Grubba27 @henriquealbert what do you think?

Grubba27:
Hey! @afrokick and @radekmie, I would also choose option 2. I did my fair part of migration as well and the I think going step-by-step may be easier and more community-friendly(would love to see you more @afrokick  and other new faces around).
On Prettier, I agree.
For the ESlint I think we should have a merge/soft-fork of the @quave/eslint-config-quave & Vazco's one ([eslint-config-vazco](https://github.com/vazco/eslint-config-vazco)). Then we could discuss which rules to add to the project due to its needs.

Also I will ping @denihs on this one so that if he was any opinion on this one, I would love to hear

denihs:
I'd also go with the second option.

For the ESlint, if the Vasco one covers everything that the Quave one does, runs in the current version, and has TS configs, I don't see why we shouldn't do it. 

I think we can give it a go.

Grubba27:
Hey, can you update again this branch @afrokick ? will squeeze this one in 2.9.1

afrokick:
@Grubba27  done

PR #12310, opened 2022-11-18
Top Level Await

zodern:
## What is it?

Top level await allows you to use `await` in the top level of a module/file instead of only in async functions. One way to view it as every file runs inside an `async` function.

For Meteor apps and packages, this doesn't really enable anything new on the server you couldn't do before. Meteor has been able to run async code in the top level of modules using fibers for many years. For example, this code already works:

```js
const Links = new Mongo.Collection('links');

// Async code using fibers. The module waits for this to finish before continuing
if (Links.find().count() === 0) {
  Links.insert({ url: 'https://meteor.com' });
}
```

With top level await, the code could instead be:
```js
const Links = new Mongo.Collection('links');

// Async code using top level await. The module waits for this to finish before continuing
if (await Links.find().countAsync() === 0) {
  Links.insert({ url: 'https://meteor.com' });
}
```

The benefits with top level await are:
- Uses promises
- Will also work in the client
- Increases Meteor's compatibility with the ecmascript spec and the broader javascript community
- With top level fiber code, only one module loads at a time so slow async code slows down loading the app. The top level await spec describes situations where modules with top level async code should load in parallel, which reduces how long it takes the server or client to load.
- It will work without fibers

### Spec compliance

So far our focus has been in ensuring modules run at the correct order at the correct time. This aspect is now always spec compliant as long as there are no circular dependencies.
Circular dependecies work, but modules are currently not always evaluated in the order defined in the spec. Other aspects of the spec might not be fully implemented or compliant.

### What is an async module?

Understanding what an async module is will be helpful with some of the explanations further down. There are two ways a module can become an async module:
- It uses top level await
- It depends on a module that is async

If you have two modules: `a.js` that uses top level await, and `b.js` that does not use top level await, then `a.js` will be an async module and `b.js` will not be.

If `b.js` imports `a.js`, then `b.js` will become async since it depends on an async module. Any modules that import `b.js` will then also become an async module for the same reason.

### Module execution order

Ecmascript modules and commonjs both evaluate modules in the opposite order. commonjs runs the root module first, and the root module then handles importing its dependencies.
Ecmascript modules run the modules in the opposite order. The root module is run last, and its dependencies are run before it. In practice, it didn't really matter since the end result was the same, so most methods of running ecmascript modules in a commonjs environment, including Meteor, would run keep the commonjs behavior.

However, with top level await, the order does matter. We now have to evaluate a module's dependencies before evaluating a module in case any of the dependencies are async.

I don't think this should affect existing code, but there could be scenarios I haven't found yet where it does.

### `require`

When using `require` to load an async module, instead of directly returning a module's exports, it will return a promise that resolves to the module's exports.
If you are using `require`, this does mean you need to be somewhat careful when adding/removing top level await in a file since you also have to update where the module is required.
Since a module becomes async if it depends on an async module, this affects more than just the individual modules using top level await.

Using ecmascript import syntax where possible instead would be good since it handles sync and async modules for you.

### Nested imports

For top level await to work correctly, the imported module has to be fully evaluated first.
With nested imports, this isn't possible if the imported module is async. Nested imports still work for sync modules, but when used with async modules it will now throw an error.

This only refers to using `import ... from ...` syntax outside of the root of a module. You can still use nested `require`, or dynamic imports.

### Enabling

Top level await is a feature of ecmascript modules, so wherever you can use ecmascript modules, you can use top level await.

Generally, this means your app or package has a dependency on `ecmascript`, `typescript`, or a similar build plugin that compiles files with reify.

### Considerations for packages

When importing a package, even when using `require`, it will work the same whether a package uses top level await or not. The implementation handles this by marking the package's  main module as a sync module after the package has loaded.

There are a couple other cases where using or removing top level await from a module in a package could be considered a breaking change:

1. If specific modules are require'd from a package. For example: `require('meteor/zodern:aurorae/svelte.js')`. When importing a specific module, `require` changes its behaviour based on if the module is async or not (look at the require section above for details)
2. If a package that has lazy main modules is require'd. Unlike normal packages, `require` will return a promise if the lazy main module is an async module. Changing if the lazy main module is async or not should be considered a breaking change for the package.

### Implementation

There are several parts of the implementation. Most of the work is handled by reify - it allows async modules to work and run in the correct order and at the correct time within each package or app. Each module is wrapped in module.wrapAsync.

For example, if you have a module with:

```js

export const LinksCollection = new Mongo.Collection('links');

console.log('Links count:', await LinksCollection.find().countAsync());
```

It is transpiled by reify into:

```js
!module.wrapAsync(async function (module, __reifyWaitForDeps__, __reify_async_result__) {
  module.export({
    LinksCollection: () => LinksCollection
  });
  let Mongo;
  module.link("meteor/mongo", {
    Mongo(v) {
      Mongo = v;
    }
  }, 0);
  if (__reifyWaitForDeps__()) await __reifyWaitForDeps__();
  const LinksCollection = new Mongo.Collection('links');

  console.log('Links count:', await LinksCollection.find().countAsync());

  __reify_async_result__();
}, { self: this, async: true });
```

The module is wrapped within an async function that is passed to `module.wrapAsync`. This async function first creates hoisted exports, then runs ecmascript imports. Next, it check if there were any async dependencies and waits for them to finish.

Once the dependencies are loaded, it then runs the module code. After it finishes, `__reify_async_result__` is called. Reify uses this to know when the module finished evaluating since using the promise returned by the async function resolves too late for reify to be spec compliant.

Top level await runs modules in parallel if they are siblings - they don't directly or indirectly depend on each other. Especially for the client, this reduces how long top level await can slow down loading the app. This works for modules within each package or app.
While Meteor packages aren't modules, we can apply the same principles to them. If two packages use top level await, and do not depend on the other, we can load the packages in parallel to speed up loading the app. While we don't have to do this, if we are going to it would be better to do it from the start.
We can always reduce paralllism in the future, but increasing it later might cause new unexpected problems in packages and apps that had previosly worked fine.

On the client load time is important, and it seems necessary to be able to load packages that use top level await in parallel. On the server, I am undecided. This could make loading the server faster,
but it also would add some unpredictability and possibly make understanding how the app works harder.
We can always reduce parallism, while increasing it could be difficult so I am tempted to start with loading packages in parallel on the server and see how it well it works.

To ensure the packages and app load after their dependencies are ready and to decide when to run startup hooks, we need a simple runtime that queues and schedules when to load them. We have a new package `core-runtime`, that handles this. Package and app bundles are queued to be run by calling `Package['core-runtime'].queue`.

After all bundles are queued, `Package['core-runtime'].waitUntilAllLoaded` can be used to know when all bundles have been evaluated. On legacy clients, if there are no async packages there is also a chance that the app doesn't have a promise polyfill. `waitUntilAllLoaded` returns a promise if there are async packages that are still loading, or null if all packages were already evaluated.

If the server uses fibers, it will use Promise.await to load the packages one at a time - this will allow build plugins published with Meteor 3 or newer to still be able to run on older Meteor versions that don't call `waitUntilAllLoaded`.

This first version will only support top level await on the server by default. Before supporting it on the client, we need to update HMR to work correctly with top level await, and figure out how to support files in `client/compatibility`. You can enable it on the client by setting the env var `METEOR_ENABLE_CLIENT_TOP_LEVEL_AWAIT` to `true`

TODO:
- [ ] Update build plugins:
  - [x] ecmascript
  - [x] babel-compiler
  - [x] typescript
  - [ ] coffeescript
  - [x] svelte
- [ ] Basics
  - [x] Figure out how to support top level fibers code and spec compliant top level await at the same time
  - [x] Figure out plan for build plugins
  - [x] Finish reify compiler it configured to modify ast
  - [x] Implement new runtime package and update linker to create the correct output
  - [x] Ensure linker creates output that would work on client
  - [x] dynamic imports/nested require for already loaded module
  - [x] Require async package as sync
  - [x] Meteor.startup (server)
  - [x] Meteor.startup (client)
  - [x] Loading build plugins waits until async modules evaluated
  - [ ] Profile loading packages on server 
  - [ ] docs
  - [ ] tests
  - [x] look into packages with lazy main modules
  - [ ] Ensure errors thrown when evaluating async modules are handled consistently with how Node handles them
  - [x] Figure out how require'ing packages with a non-lazy async main module can work without returning a promise
  - [ ] Look into files from node_modules that can be compiled with reify by the import scanner
  - [ ] Throw error when using nested imports to import async module
  - [x] Seems to be a regression where some import statements aren't transpiled
  - [x] Avoid trailing commas in linker output


filipenevola:
Hey @zodern, great work!

A few comments:

> We can always reduce parallelism, while increasing it could be difficult so I am tempted to start with loading packages in parallel on the server and see how it well it works.

I agree, it's better to be aggressive at first and reduce parallelism if problems start to emerge.

> This week I've come across two problems related to async/await and fibers. Fibers is incompatible with async/await, so when building the app Meteor replaces async/await on the server with code that uses fibers.

As discussed [here](https://github.com/meteor/meteor/discussions/11505#discussioncomment-4177920) Fibers will be removed in Meteor 3.x and TLA is necessary exactly in Meteor 3.x because of this removal.

So I don't see why we should support TLA and Fibers together, while we have Fibers it's simple to await in the root of a module using a Fiber, as Meteor does today, so I would suggest that we implement TLA targeting Meteor 3.x.

Doing so we don't need to worry about Fibers at all.

Do you see any problems with this approach?

denihs:
Yeah, while I was reading your description, I was thinking the same thing @filipenevola said. 

In Meteor 3.0, we're not going to worry about Fibers, so it looks like you could ignore Fibers now and make it work as a spec compliant only.

But maybe there are problems I'm not seeing if you choose this approach.

zodern:
The plan had been to support TLA before fibers was completely removed to make it easier to migrate code. It seems with the new schedule from last week that is no longer as important. Waiting until Meteor 3 to release this feature would make it much easier.

filipenevola:
> The plan had been to support TLA before fibers was completely removed to make it easier to migrate code. 

The plan is still the same but more aggressive so we can do both in parallel. You can proceed implement TLA support ignoring Fibers and the Fibers removal can assume TLA will be available. 

Does it make sense @zodern?

PR #12316, opened 2022-11-22
Create mongo index async

Grubba27:
This pr focus on tests that are not passing in general that rely on using indexes creation.
Focused in the collection and oplog tests that use directly ``collection.createIndex(...``
Other tests that _WILL_ fail because of DDP not being async and now probably will pass when DDP is full async are tests from functions in these packages:
 - accounts-password/password_server;
 - accounts-base/accounts_server;
 - accounts-passwordless/passwordless_server.js;
 - oauth/pending_credentials.js;
 - oath1/oauth1_pending_request_tokens.js;
 - service-configuration/service_configuration_server.js


packages/mongo-async/oplog_tests.js
```diff
@@ -18,152 +21,152 @@ Tinytest.addAsync("mongo-livedata - oplog - cursorSupported", async function (te
...
-//       MongoInternals.defaultRemoteCollectionDriver()
-//         .mongo._oplogHandle._resetTooFarBehind();
-//
-//       self.skipHandle.stop();
-//       self.subHandle.stop();
-//       self.collection.remove({});
-//     }
-//   ]
-// );
+process.env.MONGO_OPLOG_URL && testAsyncMulti(
```
denihs, line 62: Do you know why this test was commented? 

Is it ok to run it again?

denihs (approved):

packages/mongo-async/oplog_tests.js
Grubba27: It was here  [b8f2c071190c829975d22ef71f8a1f086c7bb75f](https://github.com/meteor/meteor/commit/b8f2c071190c829975d22ef71f8a1f086c7bb75f)
It is okay. all tests for collection and oplog are green

denihs: Awesome :+1:

packages/mongo-async/mongo_driver.js
```diff
@@ -850,7 +864,15 @@ MongoConnection.prototype._dropIndex = async function (collectionName, index) {
...
-  var indexName = await collection.dropIndex(index);
+  const p = new Promise(async (resolve, reject) => {
+    try {
+      const i = await collection.dropIndex(index)
+      resolve(i)
+    } catch (e) {
+      reject(e);
+    }
+  })
+  var indexName = await p;
```
radekmie, line 875: What is this `Promise` wrap doing here? It looks unnecessary at first.

packages/mongo-async/collection_tests.js
```diff
@@ -147,8 +147,8 @@ Tinytest.addAsync('collection - calling native find with good hint and maxTimeMs
     Promise.resolve(
       Meteor.isServer &&
         collection.rawCollection().createIndex({ a: 1 })
-    ).then(() => {
-      test.equal(collection.find({}, {
+    ).then(async () => {
+      test.equal(await collection.find({}, {
```
radekmie, line 151: `count()` is not async (yet).

packages/mongo-async/collection.js
```diff
@@ -745,16 +745,16 @@ Object.assign(Mongo.Collection.prototype, {
 
   // We'll actually design an index API later. For now, we just pass through to
   // Mongo's, but make it synchronous.
-  _ensureIndex(index, options) {
+  async _ensureIndex(index, options) {
```
radekmie, line 748: When is this going to be released? `_ensureIndex`, `createIndex`, and others are all part of the public collections API.

If it's going to be 3.0, maybe we should already remove the underscored methods as well as the deprecated ones and simply expose fibers-less ones, just like I did in #12295?

packages/mongo-async/mongo_driver.js
Grubba27: Tests were only passing with this code. Now they are without it. Not sure tho. Removed and all is still good
Screenshot 2022-11-25 at 07 54 17

packages/mongo-async/collection.js
Grubba27: > When is this going to be released? _ensureIndex, createIndex, and others are all part of the public collections API.

It is going to 3.0. Will try my best to always add this info to the pr's so everyone is always on the same page :) 

I'm not sure what do you mean by ``maybe we should already remove the underscored methods as well as the deprecated ones and simply expose fibers-less ones``  because it is already a wrapper for the raw mongo collection(with a little bit more like logging). 
I will make a commit with what I think it is

radekmie: I mean having a clear API - no underscores and no two methods doing the same thing (`_ensureIndex` and `createIndex`).

packages/mongo-async/collection.js
```diff
@@ -745,16 +745,29 @@ Object.assign(Mongo.Collection.prototype, {
 
   // We'll actually design an index API later. For now, we just pass through to
   // Mongo's, but make it synchronous.
-  _ensureIndex(index, options) {
+    /**
+   * @summary Creates the specified index on the collection.
+   * @locus server
+   * @deprecated in 3.0
```
Grubba27, line 751: like this @radekmie ?

radekmie: It's fine, but the `@method` is incorrect now (it says `createIndex` instead of `_ensureIndex`).

Issue #12321, opened 2022-11-25
Meteor v2.8 - Memory leak in mongo driver?


xet7:
https://forums.meteor.com/t/meteor-v2-8-memory-leak-in-mongo-driver/59101

xet7:
From [MongoDB Jira](https://jira.mongodb.org/browse/NODE-4833?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=5012475#comment-5012475) - Bailey Pearson:

Hey dominik,
there are two variations of sessions in the Node driver.  Every operation that gets sent to the server has a session attached - we represent these sessions with ServerSessions.  The driver manages ServerSessions in a ServerSessionPool and will clean up stale server sessions when they expire.  ClientSessions are an abstraction over ServerSessions that allow users of the driver to provide a session for operations.  The driver does not manage ClientSessions - users are responsible for ending them using the `endSession` method.

We allow users to create sessions using the MongoClient's `startSession` method (source [here](https://github.com/mongodb/node-mongodb-native/blob/73e92ce93e54b8d4d966ede319c3d7805ced6776/src/mongo_client.ts#L622)).  This method ensures that when `endSession`s is called, we remove the client session from the ActiveSession set.

The reported bug is a buildup of ClientSessions.  My suspicion (without knowing how Meteor works) is that Meteor is creating client sessions under the hood but never ending the sessions, resulting in a buildup of sessions in the driver.

xet7:
From [MongoDB Jira](https://jira.mongodb.org/browse/NODE-4833?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=5012475#comment-5012475) - Patrick Schubert

To Bailey Pearson

Thank you for your reply. What confuses me the most is the following: 99 (maybe even 100) percent of Meteor's mongo calls are handled by this package: https://github.com/meteor/meteor/blob/devel/packages/mongo/collection.js

I just checked its codebase (and all the other Meteor code), but could NOT find any calls of startSession() anywhere! 

My question is: Are there any other methods to create ClientSessions? What am I missing here? How is it possible that we see so many ClientSessions in our heapdump but I can't find the according code anywhere?

EDIT: I am not fully aware of all the mongodb driver internals, but [this code](https://github.com/mongodb/node-mongodb-native/blob/main/src/operations/execute_operation.ts#L113) seems to me, that sessions are also indeed created implicitly?!

[...]

PR #12326, opened 2022-11-29
Make count NOT create a cursor

znewsham:
currently, calls to `Collection.find().count()` create a new session that is never released. This is because the mongo API changed when a session is created to init time of the cursor, and the cursor methods (fetch, forEach, map, count) all create a real cursor at call time. 

In the case of count, this cursor is never used.

I believe this is the fix to: https://github.com/meteor/meteor/issues/12321


packages/mongo/mongo_driver.js
```diff
@@ -907,12 +907,27 @@ function setupSynchronousCursor(cursor, method) {
...
+  if (methodName === 'count') {
+    return;
+  }
   Cursor.prototype[methodName] = function (...args) {
     const cursor = setupSynchronousCursor(this, methodName);
     return cursor[methodName](...args);
   };
 
+
+  Cursor.prototype.count = function () {
```
baileympearson, line 923: This should go outside of the `forEach` loop, or inside the `methodName === 'count'` guard, so `Cursor.prototype.count` is not redefined on every iteration of the loop.

znewsham: :facepalm: thanks for the catch!

Twisterking:
Thanks a lot for the very quick reaction! Any idea how quickly, usually, patches like this are pushed to a new (patch) version of Meteor?

Grubba27:
I will create a patch for this ASAP will be released under the 2.8.2 namespace

Grubba27:
Travis CI tests are failing

znewsham:
@Grubba27 can you provide any details? that output is hard to parse - I ran the mongo tests locally and they all passed

znewsham:
ah - is it this?

```
S: tinytest - livedata connection - reconnect to a different server failed: {"groupPath":["tinytest","livedata connection"],"test":"reconnect to a different server","events":[{"sequence":2477,"type":"fail","details":{"type":"assert_equal","expected":"[true]","actual":"[false]","not":false,"asyncBlock":0}}],"server":true}
```

Grubba27:
yeah is that one. I will try running locally like this:

```bash
./meteor self-test
```

Grubba27:
ah got it. Will run them locally if they all pass I will continue with release

Grubba27:
All mongo tests have passed and the test that has not passes is a problem in our end reverse.meteor.com is down

Grubba27:
2.8.2 was launched
check if all is okay there

Grubba27:
Good to hear @DavidSichau, we released yesterday 2.9. Maybe will help even more te mem consumption

PR #12327, opened 2022-11-29
feat: Auto import React on jsx,tsx files

henriquealbert:
This change will auto import React on jsx and tsx files by default.

You don't need to repeat yourself writing `import React from 'react'` in every React file.

You can read more about it on React documentation: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#manual-babel-setup

We already support it on `@babel/preset-react` since we are using v7.16.7 and this is added on v7.9

This was requested on forums and also part of our backlog. 
https://forums.meteor.com/t/solved-why-i-still-need-to-import-react-on-each-jsx-tsx-file/58358

TODO:
- [ ] Deploy new @meteorjs/babel version and replace this version everywhere it's used

### DIFF
--- a/npm-packages/meteor-babel/options.js
+++ b/npm-packages/meteor-babel/options.js
@@ -80,7 +80,11 @@ exports.getDefaults = function getDefaults(features) {
 
 function maybeAddReactPlugins(features, options) {
   if (features && features.react) {
-    options.presets.push(require("@babel/preset-react"));
+    options.presets.push(
+      [require("@babel/preset-react"), {
+        runtime: "automatic"
+      }]
+    );
     options.plugins.push(
       [require("@babel/plugin-proposal-class-properties"), {
         loose: true


henriquealbert:
image

kakadais:
I came across another thread, such as [this one](https://github.com/meteor/meteor/pull/13193), but it seems like this issue still exists in the current version, METEOR@3.0.4.

Is it left this way intentionally for compatibility or some other reason?

StorytellerCZ:
@kakadais yes, from what I remember this was reverted due to issues.

StorytellerCZ:
Would be nice to figure out how to include this again.

paulincai:
@StorytellerCZ I import  a file in my client startup. This file only contains this:

```
import React from 'react'
window.React = React
```
I think this is it. I never import React from 'react'

This article refers to React 17 in which Babel started to look for React in the global scope. I assume that would be the Window object.

https://www.linkedin.com/pulse/say-goodbye-react-imports-17s-game-changing-feature-deepak-sharma-/

PR #12328, opened 2022-11-29
Meteor version to 2.8.2 :comet:

Grubba27:
Solves #12321


Grubba27:
there is a 2.8.2 release. docs here updated.

PR #12342, opened 2022-12-06
Solving 2.9 tests

Grubba27:
This is a pr for trying to solve missing tests to make a release for 2.9 work.
By default we try not to make releases with tests failing


zodern:
Changing the runtime to automatic breaks any apps that use older versions of React. We've been trying to keep support for older versions of react, for example in the react fast refresh implementation.

One way to solve this is to use the classic runtime by default, and add an option to the `features` object for configuring it. Then the babel-compiler package can check what version of react is being used, and choose the correct runtime.

Grubba27:
> One way to solve this is to use the classic runtime by default, and add an option to the features object for configuring it. Then the babel-compiler package can check what version of react is being used, and choose the correct runtime.

I'm not sure if I understood the ``features`` part, can you provide an example for it?

zodern:
The features object is created at https://github.com/meteor/meteor/blob/ab07174f87e720cb876f785f857f9b16699cd482/packages/babel-compiler/babel-compiler.js#L83 and is available where adding the react plugin as the `features` parameter: https://github.com/meteor/meteor/blob/ab07174f87e720cb876f785f857f9b16699cd482/npm-packages/meteor-babel/options.js#L81

This is used to configure which babel plugins are used and with what options.

Grubba27:
Got it! Nice to know. Thanks a lot, Zodern. We in the team decided that since this is causing some issues with tests and we saw that our build got a little spike in memory(don't know if it is related), we want to be cautious with what we change in on the foundation's parts such as the babel one. Will revert to the way it was and think it to 2.10

PR #12357, opened 2022-12-12
Async ddp-client and ddp-server

denihs:
Changing these packages to work without Fibers.


radekmie:
How the hell did you review this massive PR in 11 minutes?!

PR #12362, opened 2022-12-12
Update on tinytest UI

Grubba27:
While making the fibers-free tests pass, we made tinytest tests in the console into dark mode, in this PR I did an upgrade in the color palette to be more modern instead of just inverted CSS dark mode.
Here is a print screen of the final result

Screenshot 2022-12-12 at 19 45 37

Also, I've placed a new font I'm still deciding if I continue with [Space Grotesk](https://fonts.google.com/specimen/Space+Grotesk?query=space), which is the same as we are using in Galaxy Cloud.

### DIFF
--- a/packages/test-in-browser/driver.html
+++ b/packages/test-in-browser/driver.html
@@ -1,5 +1,12 @@
 
 <template name="testInBrowserBody">
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+  <link href="https://fonts.googleapis.com/css2?family=VT323&display=swap" rel="stylesheet">
+
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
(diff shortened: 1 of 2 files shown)


packages/test-in-browser/driver.html
```diff
@@ -1,5 +1,8 @@
 
 <template name="testInBrowserBody">
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+  <link href="https://fonts.googleapis.com/css2?family=Space+Grotesk:wght@300;400;500;600;700&display=swap" rel="stylesheet">
```
radekmie, line 5: Entire Europe is getting rid of Google Fonts because of GDPR compliance. I'd strongly advise against using it.

Grubba27: I did not know about that! Will make the changes ASAP

denihs (approved):

zodern:
Have you tested this in IE? It probably breaks the UI when running tests in IE since it doesn't support css vars.

Grubba27:
> Have you tested this in IE? It probably breaks the UI when running tests in IE since it doesn't support css vars.
You are right. I did not thought of IE support :/ will add those so that it does not break

PR #12381, opened 2022-12-15
Making new Tinytest UI IE compatible

Grubba27:
As @zodern commented, the new Tiny Test Ui was not IE compatible in this PR I ensure that it will work properly by passing the same prop but without variable

### DIFF
--- a/packages/test-in-browser/driver.css
+++ b/packages/test-in-browser/driver.css
@@ -25,7 +25,9 @@
 body {
     height: 100vh;
     width: 100vw;
+    background-color: #18181b !important;
     background-color: var(--bg-black) !important;
+    color: #F9FAFB !important;
     color: var(--primary-white) !important;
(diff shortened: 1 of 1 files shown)


zodern:
Thanks for fixing this.

I haven't tried it locally, but one thing I noticed in the screenshot in #12362 is that there doesn't seem to be much contrast between some of the colors, making some of the text hard to read. For example, some of the red text, the text on top of the progress bar, and the gray detail text for each test (though that might have been changed).

Grubba27:
I did. Let me send a print with the updated colors. I had issues with the accessibility/ contrast in the way it was.

Grubba27:
Screenshot 2022-12-15 at 16 34 08

Grubba27:
Screenshot 2022-12-15 at 16 35 18

PR #12384, opened 2022-12-16
Update test-in-browser dependencies 

harryadel:
- Update Blaze from 2.3.4 to 2.6.1
- Update https://github.com/JackuB/diff-match-patch library to latest version

I also thought about removing bootstrap since it's used for minimal stuff that can be easily replaced, what do you think?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,14 +10,14 @@ jobs:
       run:
         working-directory: docs/
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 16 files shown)


Grubba27:
sounds good! Thx a lot for your contruibuiton. I will check what is making the tests fail

harryadel:
Most likely because of the exports, I don't have access to a computer so
please remidate it yourself, thank you!

harryadel:
@Grubba27 Looks good?

Grubba27:
Now looks good. Will make a remainder for this one. I'm thinking of making a 2.9.2, maybe this could go in that one.
Thanks a lot again for this PR @harryadel

PR #12385, opened 2022-12-16
chore: resetPassword and verifyEmail to no longer sign in the user automatically

denihs:
Changing the methods `resetPassword()` and `verifyEmail()` to no longer sign in the user automatically if they have 2fa enabled.

By default these methods already work has the 2FA, sort of speaking, because you need to have access to the user's email in order to finish the operation.

This PR just adds one more layer to that.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -9,7 +9,9 @@
 
 #### Breaking Changes
 
-N/A
+* `accounts-password@2.3.3`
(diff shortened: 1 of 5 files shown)


denihs:
I consider this to be a breaking change. Is that a fair statement? 

Maybe we should apply this by default but give the developer the option to disable this behavior?

Grubba27:
> I consider this to be a breaking change. Is that a fair statement? 
I do think the same. 

But at the same I think it maybe a thing to go in a patch because sounds like a security improvment

PR #12387, opened 2022-12-19
docs: update docs about --prototype

Grubba27:
In 2.9 was added the ``--prototype`` flag was missing updating docs for this new feature.

for context: 
https://forums.meteor.com/t/noob-issue-hello-meteor-links-not-displaying/59240

### DIFF
--- a/docs/source/commandline.md
+++ b/docs/source/commandline.md
@@ -91,6 +91,12 @@ You can pass an absolute or relative path.
 
 **Flags for default packages**
 
+`--prototype`
+
+Creates a package with the prototype packages. 
+It can be used together with other flags that create apps such as `--react` or `--typescript`.
+
+
 `--bare`
 
 Creates a basic, blaze project.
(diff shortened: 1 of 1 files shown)


docs/source/commandline.md
```diff
@@ -91,6 +91,12 @@ You can pass an absolute or relative path.
 
 **Flags for default packages**
 
+`--prototype`
```
fredmaiaarantes, line 94: Maybe mention what the prototype packages are? Or are you going to add them to the Packages table?

Grubba27: > Maybe mention what the prototype packages are

I've updated the text to the following:
```txt
Creates a package with the prototype purpose packages(`autopublish` and `insecure`) 
if you use them you can change your collections quickly, 
but it is not supposed to be used in production.
For more information about security you can check
it [here](https://guide.meteor.com/security.html#checklist)
It can be used together with other flags that create apps such as `--react` or `--typescript`.
```

PR #12392, opened 2022-12-20
Remove toHTML method from boilerplate-generator

matheusccastroo:
Remove Fibers usage from the boilerplate-generator (actually just remove this method, since it was already deprecated). However, we can still keep the method there and throw an error if someone tries to use? Internally this is not used anymore, but someone may still use it.


Grubba27 (approved):
LGTM!

PR #12394, opened 2022-12-21
Moves code from mongo-async package back to mongo package

edimarlnx:
This Pull Request moves all code from the mongo-async package back to the mongo package.


packages/mongo/package.js
```diff
@@ -9,7 +9,7 @@
 
 Package.describe({
   summary: "Adaptor for using MongoDB and Minimongo over DDP",
-  version: '1.16.3'
+  version: '1.16.0'
```
denihs, line 12: Why was the version decreased?

edimarlnx: It is a merge error. I have fixed it.

matheusccastroo (approved):

PR #12406, opened 2022-12-28
Making oauth async

Grubba27:
In this PR I focus on making ``oauth`` async, making it possible to have a Fibers-free MeteorJS.

- [x]  make tests pass
- [x]  Add docs in changelog

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -18,6 +18,14 @@
 * `boilerplate-generator`:
   - `toHTML` is no longer available (it was already deprecated). Use `toHTMLStream` instead.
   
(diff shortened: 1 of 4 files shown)


denihs (approved):

filipenevola (commented):
I think you also need to update the usages in `oauth2` package.

What do you think?

Grubba27:
Yes! this will be the next one @filipenevola. I was going on my holiday break so I made each package a separate PR.
Now that I'm back I will start working on this as soon as I catch up on what I have missed

Issue #12412, opened 2022-12-29
accounts-passwordless does not always convert email selector to lowercase


msavin:
Meteor 2.9.0

I've been noticing the following error during testing with Meteor:

```
User not found [403]
```

```
"Error: User not found [403]
    at Connection._livedata_result (http://localhost:3200/packages/ddp-client.js?hash=12df120d82048869c74b009c3544bf96a2f333dc:1812:23)
    at Connection.onMessage (http://localhost:3200/packages/ddp-client.js?hash=12df120d82048869c74b009c3544bf96a2f333dc:1961:12)
    at http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:193:11
    at Array.forEach (<anonymous>)
    at ClientStream.forEachCallback (http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:262:31)
    at socket.onmessage (http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:192:14)
    at REventTarget.dispatchEvent (http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:482:22)
    at SockJS._dispatchMessage (http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:1502:10)
    at SockJS._didMessage (http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:1556:16)
    at that.ws.onmessage (http://localhost:3200/packages/socket-stream-client.js?hash=cead8e84bd8ebf243d505fc1bac78ee9d9e6be3d:1683:15)"
```

Meteor converts all emails to lower case as it can then be easier to match them (i.e. maX@meteor.com is converted to max@meteor.com) - however, for accounts-passwordless, it looks like this is not happening on the `selector` field which could then say that the user is not found.

For example, you can see it on the sign in link:
```
http://localhost:3200/?loginToken=2726&selector=asdf@Me.com
```

zodern:
Meteor normally uses a case insensitive query for emails so emails are stored how the user entered them, but when querying it ignores case.

The package isn't using the normal query for emails:
[...]

PR #12414, opened 2022-12-30
fix: convert selector client email to lower case

celomary:
fix issue #12412 by turning selector email to lower case


StorytellerCZ:
Something similar has been discussed elsewhere. In account-password the lookup by user email is case insensitive, which has been a center of few debates about performance. In the end I don't recall any change being made. So instead of just putting it all to underscore I think it needs to run through the same system as account-passwords or similar.

yasharyasaxena:
Hi, I am new to contributing.
I was looking at #12412 and saw this pr mentioned 
Can someone @StorytellerCZ explain why this was not accepted as a solution

PR #12415, opened 2022-12-30
Use mongodb driver types for Selector=>Filter, Modifier=>UpdateFilter, SortSpecifier=>Sort instead of homebuilt types

perbergland:
Now that mongodb has great typescript support, bring in filter, updatefilter and sort types from the driver types and remove the now unused old types.

Replica of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63759

### DIFF
--- a/packages/mongo/mongo.d.ts
+++ b/packages/mongo/mongo.d.ts
@@ -16,12 +16,9 @@ export type UnionOmit<T, K extends keyof any> = T extends T
 
 export namespace Mongo {
 
-  /**
-   * Alias for {@link MongoNpmModule.Filter}
-   */
   type Query<T> = MongoNpmModule.Filter<T>;
 
-  type QueryWithModifiers<T> = {
+  export type QueryWithModifiers<T> = {
     $query: Query<T>;
     $comment?: string | undefined;
     $explain?: any;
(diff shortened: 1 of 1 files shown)


perbergland:
also copied over a fix for the cursor iterators, since the current code made it unusable

StorytellerCZ (approved):

packages/mongo/mongo.d.ts
```diff
@@ -15,78 +15,11 @@ export type UnionOmit<T, K extends keyof any> = T extends T
...
-  type Query<T> = {
-    [P in keyof T]?: Flatten<T[P]> | RegExp | FieldExpression<Flatten<T[P]>>;
-  } & {
-    $or?: Query<T>[] | undefined;
-    $and?: Query<T>[] | undefined;
-    $nor?: Query<T>[] | undefined;
-  } & Dictionary<any>;
+  /**
+   * Alias for {@link MongoNpmModule.Filter}
+   */
```
radekmie, line 21: I think this comment is not needed. (Or other types should get one too.)

radekmie (approved):

perbergland:
I also felt like adding the missing (new async methods in 2.9.1) countDocuments and estimatedDocumentCount since I couldn’t see any PR or issue for adding them, but it felt like scope creep. Whaddya think.

packages/mongo/mongo.d.ts
perbergland: Removed it (along with a dozen other now unused types)

Grubba27:
> Whaddya think?

Sounds great! 2.9.1 is already out, but I'm thinking of making a 2.9.2 with a few small additions. I could squeeze this one there I think

PR #12416, opened 2022-12-30
Id issue

### LINKED ISSUE #12401: Accounts Passwordless login doesn't work with id and token
Hi, one option of Accounts Passwordless login is to login with user id and token.

This option is helpful when you don't have the email of this user.

But there is an inconsistent check:

This validator requires `id` instead of `_id` https://github.com/meteor/meteor/blob/4868a2bb9411b9620e640696ed2d3c400b30b12c/packages/accounts-passwordless/passwordless_server.js#L33

but the rest operator here https://github.com/meteor/meteor/blob/4868a2bb9411b9620e640696ed2d3c400b30b12c/packages/accounts-passwordless/passwordless_server.js#L16-L24 is not going to work to find the user as the `id` in Meteor collections is `_id`.

This code should check if `id` is in the selector and change to `_id`.

I never understood why Meteor had this `id` in Accounts packages but it does so I think it's easier to just fix the way to provide the selector to the query. 

> I got this problem during a [YouTube Live](https://www.youtube.com/watch?v=9Xer73xPk2s) in Portuguese so you can see how I detected the problem probably around 1hr and 15mins

celomary:
fix #12401 by changing id to _id

### DIFF
--- a/packages/accounts-passwordless/passwordless_client.js
+++ b/packages/accounts-passwordless/passwordless_client.js
@@ -15,7 +15,7 @@ const transformSelector = selector => {
   }
 
   if (selector.includes('@')) {
-    return { email: selector };
+    return { email: selector.toLowerCase() };
   }
 
   return { username: selector };
--- a/packages/accounts-passwordless/passwordless_server.js
+++ b/packages/accounts-passwordless/passwordless_server.js
@@ -129,7 +129,7 @@ Meteor.methods({
     if (!user) {
       const userId = createUser(userData);
       user = Accounts._findUserByQuery(
-        { id: userId },
+        { _id: userId },
         {
           fields: { emails: 1 },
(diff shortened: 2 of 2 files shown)


packages/accounts-passwordless/passwordless_client.js
```diff
@@ -15,7 +15,7 @@ const transformSelector = selector => {
   }
 
   if (selector.includes('@')) {
-    return { email: selector };
+    return { email: selector.toLowerCase() };
```
StorytellerCZ, line 18: This doesn't seem to be related to the issue.

StorytellerCZ (changes requested):

italojs:
fixed in another pr https://github.com/meteor/meteor/pull/12797

PR #12419, opened 2023-01-03
Update skeletons to use React 18

StorytellerCZ:
Updated the following skeletons to use React 18 and the new `createRoot` function:
* Apollo
* React
* Typescript

### DIFF
--- a/packages/react-fast-refresh/package.js
+++ b/packages/react-fast-refresh/package.js
@@ -1,14 +1,14 @@
 Package.describe({
   name: 'react-fast-refresh',
-  version: '0.2.3',
+  version: '0.2.4',
   summary: 'Automatically update React components with HMR',
   documentation: 'README.md',
   devOnly: true,
 });
 
 Npm.depends({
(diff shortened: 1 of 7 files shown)


zodern:
For HMR to work with react 18 we also need to update react-refresh: https://github.com/meteor/meteor/blob/b43018cdd6a20744869a3b1f24c7a4c115fbef64/packages/react-fast-refresh/package.js#L10

StorytellerCZ:
OK, that should be done then. Do I just need to update the npm package or is something more needed?

zodern:
You should only need to update the npm package.

Grubba27:
This sounds like a change for a 2.10 version instead o a 2.9.2

Grubba27:
What is missing here @StorytellerCZ or it is done/ can be merged ?

StorytellerCZ:
@zodern I have updated the package. Please check and let me know if things are fine. @Grubba27 once @zodern approves then it is ready.

zodern (approved):

PR #12426, opened 2023-01-04
Allow multiple runtime config and updated runtime hooks

ebroder:
The forEach method on Hook will stop iterating unless the iterator function returns a truthy value. Previously, this meant that only the first registered runtime config hook would be called.

(I ran into this by accident when I tried to add a second hook to my application in a different area of functionality. I attempted to look for other places throughout the codebase that might have this bug and wasn't able to find any others.)


Grubba27:
Thanks a lot for your contribution @ebroder

PR #12428, opened 2023-01-06
Remove Blaze dependency and types that live in blaze.d.ts

perbergland:
EventsMap and friends are already in the @types/Meteor blaze.d.ts (not imported to meteor/blaze repo yet though)

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/meteor/blaze.d.ts

### DIFF
--- a/packages/meteor/meteor.d.ts
+++ b/packages/meteor/meteor.d.ts
@@ -1,6 +1,5 @@
 import { Mongo } from 'meteor/mongo';
 import { EJSONable, EJSONableProperty } from 'meteor/ejson';
-import { Blaze } from 'meteor/blaze';
 import { DDP } from 'meteor/ddp';
 
 export type global_Error = Error;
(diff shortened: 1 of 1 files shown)


Grubba27:
@jankapunkt maybe you can bring it to blaze package as a type.d.ts and go with [zodern:types](https://github.com/zodern/meteor-types) there? here is the [docs](https://docs.meteor.com/using-core-types.html)

jankapunkt:
@Grubba27 yes I can do that I currently have, however, no environment to test so I can only blindly push and open a PR.

Grubba27:
To remove/move things from Meteor, I would like this to be made along with adding types within Blaze. For this release, I will skip this one. 
I want to be sure when removing( I know that things are there in def types), but when I remove them in one place, I would like every other place removed as well, and things are ready in Blaze.
Probably in the next release, this one will be there :)

perbergland:
Well these types that were removed were never usable or used since they weren't exported or used in the meteor types so they can be removed immediately with no impact to anyone. 

They are just dead and unusable code

perbergland:
They come with a negative impact of bringing in an unwanted dependency so removing them does more than remove dead code

jankapunkt:
@Grubba27 @perbergland if this is so, can you point me to a package that exports types via `zodern:types` and that works well, so I can try to use this as blueprint for types def in Blaze?

Grubba27:
if people use now a days the https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/meteor/blaze.d.ts you can just get this file and place it on the root of the package and declare it on package.types.json.
A good example is the mongo package https://github.com/meteor/meteor/tree/eebe2b476ba5bb081395ef9ed864099bbbe7ae1c/packages/mongo

Grubba27:
As commented in the [slack community messages ](https://meteor-community.slack.com/archives/CS46CNDN0/p1675884975487859?thread_ts=1675873195.556599&cid=CS46CNDN0) this feature maybe useful for other users while using Meteor Types

PR #12429, opened 2023-01-06
Update boilerplate-generator-tests

harryadel:
Replaced [stream-to-string](https://github.com/jasonpincin/stream-to-string) with a function
Updated parse5 to [v6.0.1](https://github.com/inikulin/parse5/releases/tag/v6.0.1)


Grubba27:
Sent to devel by mistake :x but all is good now. This will be in the next 2.10 release

PR #12430, opened 2023-01-07
Replace double-ended-queue with denque

harryadel:
Replace https://github.com/petkaantonov/deque with https://github.com/invertase/denque a [faster](https://docs.page/invertase/denque/benchmarks) and much more maintained alternative


StorytellerCZ (approved):
LGTM

PR #12393, opened 2022-12-21
Typescript v4.7.4

StorytellerCZ:
Since nobody has taken on this, I have decided to at least get the work started here and see what happens to the tests (will fail as npm package is not published - that is the first thing) and maybe this will be enough or someone can take over this.

[v4.7 announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/)

PR #12435, opened 2023-01-09
New Dev bundle 14.21.2.3 to include typescript bump

Grubba27:
This relates to typescript bump in #12393

### DIFF
--- a/scripts/dev-bundle-tool-package.js
+++ b/scripts/dev-bundle-tool-package.js
@@ -15,7 +15,7 @@ var packageJson = {
     "node-gyp": "8.0.0",
     "node-pre-gyp": "0.15.0",
     typescript: "4.5.4",
-    "@meteorjs/babel": "7.18.0-beta.4",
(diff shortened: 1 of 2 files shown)


scripts/dev-bundle-tool-package.js
```diff
@@ -15,7 +15,7 @@ var packageJson = {
     "node-gyp": "8.0.0",
     "node-pre-gyp": "0.15.0",
     typescript: "4.5.4",
```
afrokick, line 17: @Grubba27 that is a right place to update typescript version, not in eslint-plugin...

Grubba27: Yeah, I had in mind that needed only to update  the meteor-babel and meteor-babel package but I found this [commit](https://github.com/meteor/meteor/commit/f00f7534a2daa430ee5b63eac427878bba694310#diff-9aa09d29eebfdafcd71c7d703578e34f6558f0aa11ad7f15a23b44fc1a05492d)
Will add this one to the internal docs in order to never be missed this again

PR #12444, opened 2023-01-13
OnlyOneOf support for match.js

jamauro:
Proposing to add `OnlyOneOf` as the equivalent of JSON Schema's [oneOf](https://json-schema.org/understanding-json-schema/reference/combining.html?highlight=oneof#oneof). It checks that there is one and only one match among the choices passed in. Could also be called `ExactlyOneOf` but I went with the shorter version.

I assumed that changing `Match.OneOf` to `Match.AnyOf`(to mimic JSON Schema's nomenclature) and using `Match.OneOf` for this new proposal would probably not be great for people who rely on this package.

### DIFF
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -59,6 +59,10 @@ export const Match = {
   OneOf: function(...args) {
     return new OneOf(args);
   },
+  
+  OnlyOneOf: function(...args) {
+    return new OnlyOneOf(args);
+  },
 
   Any: ['__any__'],
   Where: function(condition) {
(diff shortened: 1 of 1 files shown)


packages/check/match.js
```diff
@@ -352,6 +366,24 @@ const testSubtree = (value, pattern) => {
...
+    const results = [];
+    for (let i = 0; i < pattern.choices.length; ++i) {
+      results.push(testSubtree(value, pattern.choices[i]));
+    }
+
+    if (results.filter(r => r === false).length === 1) {
+
+      // Only one succeeded? Yay, return.
+      return false;
+    }
```
radekmie, line 380: It'd be better to count the passed tests in the loop and throw early. In other words, if there are 10 tests and all pass, it should evaluate only the first two.

radekmie (commented):
I see where the needs come from, but this change definitely requires tests, documentation, and typings. Besides, it'd be better to gather some feedback from the community first (e.g., I don't see a need for it).

jamauro:
Hey, thanks for taking a look. Fair point on community feedback. It does seem to fall in the bucket of more rarely needed. I will put up a question in the forums and see what others think.

If others think it could be useful, then I can take a stab at tests, typings, and documentation.

Grubba27:
Do you still want/need this @jamauro ?

jamauro:
Not at the moment

PR #12451, opened 2023-01-16
[3.0] Solving meteor-promise tests

Grubba27:
In this PR I focus on making `promise` async, making it possible to have a Fibers-free MeteorJS.

- [x]  make tests pass
- [x]  Add docs in changelog (there were none)
Screenshot 2023-01-16 at 20 29 57

### DIFF
--- a/packages/promise/promise-tests.js
+++ b/packages/promise/promise-tests.js
@@ -7,14 +7,6 @@ Tinytest.addAsync("meteor-promise - sanity", function (test, done) {
     })
     .catch(function (error) {
       test.equal(error, expectedError);
-      if (Meteor.isServer) {
-        // TODO[FIBERS]: Remove this in 3.0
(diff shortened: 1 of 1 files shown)


packages/promise/promise-tests.js
```diff
@@ -1,18 +1,24 @@
...
-  });
+  Promise.resolve("working")
+    .then(function (result) {
+      test.equal(result, "working");
+      throw expectedError;
+    })
+    .catch(function (error) {
+      test.equal(error, expectedError);
+      if (Meteor.isServer) {
+        // TODO[FIBERS]: Remove this in 3.0
```
denihs, line 11: As this is going to be in version 3, I don't think we need this test anymore... You could remove it already.

denihs (changes requested):

PR #12455, opened 2023-01-18
docs: created 2.10 migration guide

Grubba27:
Adding 2.10 migration guide to our docs


denihs (approved):
Don't forget to update the _config.yml file

Grubba27:
It was added but I forgot to add this docs in the files :/
https://github.com/meteor/meteor/blob/devel/guide/_config.yml#L8

PR #12458, opened 2023-01-19
Fix HMR circular dependency check for common ancestors

### LINKED ISSUE #12457: HMR triggers hot code push on common ancestor, treated as a circular dependency
meteor: v2.5.8
hot-module-replacement: v0.5.0 - 0.5.2 (latest)
OSX 13.1

### Expected Behavior
HMR works for an accepting import chain with a common import ancestor within the chain.

### Actual Behavior
A full hot code push is triggered and client logs:
```
HMR: Some changes can not be applied with HMR. Using hot code push.
```

### Cause
HMR's client [checkModuleAcceptsUpdate](https://github.com/meteor/meteor/blob/devel/packages/hot-module-replacement/client.js#L292) performs a DFS up the import chain, but tracks a single, flat, ever growing `Set` of all `checked` modules so far to [detect circular dependencies](https://github.com/meteor/meteor/blob/96f48f5d4ba6d16e8f82e3d33b160e115da27834/packages/hot-module-replacement/client.js#L319-L321). This tracks all modules visited in the import tree, without unwinding or considering the path to a revisited module, so it incorrectly flags common ancestors as circular dependencies.

e.g. For modules A, B, C, D, & E, where:
- A accepts hot updates and imports B
- B imports C & D
- C & D each import E
```
    A (hot.accepts)
    |
    B
   / \
  C   D
   \ /
    E
```

On a change to module E, it will first check one import chain, e.g. `E -> C -> B -> A`, adding each of these modules to the `checked` set, and then upon checking the other chain, `E -> D -> B -> A`, once at B it will find module B is already in the `checked` set and the circular dependency check will return `null` and result in a hot code push rather than continuing up to module A where it would be accepted.

### Proposed Solution
Track the current path of visited modules during import DFS and use this set for the circular dependency check rather than the full set of `checked` modules anywhere in the import tree.

mfen:
Fixes #12457 by tracking the path of `visited` importing modules in a stack when checking hot module acceptance, and using this stack for circular dependency checks rather than using the full set of all `checked` modules across all paths.

### DIFF
--- a/packages/hot-module-replacement/client.js
+++ b/packages/hot-module-replacement/client.js
@@ -289,7 +289,22 @@ function replaceFileContent(file, contents) {
   file.contents = moduleFunction;
 }
 
-function checkModuleAcceptsUpdate(moduleId, checked) {
+function checkImportingAccepts(checked, visited, importing) {
+  let accepts = null;
+  while (importing.length > 0 && accepts !== false) {
+    const depId = importing.pop();
+
+    // If already checked, assume it accepted and doesn't need to be checked for this chain.
+    // This is safe as a refusal or unknown result from another import chain will override
+    // this acceptance.
+    accepts = checked.has(depId) ? true : checkModuleAcceptsUpdate(depId, checked, visited);
+  }
+
+  return accepts === null ? false : accepts;
+}
+
+function checkModuleAcceptsUpdate(moduleId, checked, visited = []) {
+  // Set of all modules checked, tracked to later rerun if update is accepted
   checked.add(moduleId);
 
   if (moduleId === '/' ) {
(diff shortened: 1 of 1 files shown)


Grubba27:
I would ask you to implement the same test you mentioned in the issue to show that your change does happen.
I'm no sure of other ways to test this, will ping zodern for this one @zodern

zodern:
Currently we don't have tests for HMR. We would need to refactor the package to do it, but it should be possible to use tinytest to add tests.

Thanks for working on this @mfen. There are a couple scenarios I want to check to ensure it doesn't cause any regressions.

italojs:
@mfen the issue still persists in the newest version of meteor? should we mantain this ~~abandoned~~ PR  open?

italojs:
abandoned

PR #12469, opened 2023-01-23
mongo: In async wrappers, catch exceptions and reject

ebroder:
Previously, if an exception was thrown in an async method, it would not be caught, which is unusual in an async method. Instead, catch the exception and return it as a rejected promise.


Grubba27 (commented):
LGTM! But I would ask @radekmie his opinions on it

PR #12471, opened 2023-01-24
Release 3.0 tools and mongo

denihs:
In this PR we have:

- Most of the tools already async working without fibers
- Changes in the mongo package to work without fibers
- Top-level await feature

These are the three main tasks we're working on before finishing Meteor 3.0. 

Even though none of them is 100% done, in this PR, we already have all of them working together. 

So we'll merge everything now, and new branches can be created from the branch `release-3.0` with all those features.


Grubba27 (approved):

edimarlnx (approved):

tools/utils/fiber-helpers.js
```diff
@@ -31,26 +46,13 @@ function disallowedYield() {
 disallowedYield.disallowed = true;
 
 exports.noYieldsAllowed = function (f, context) {
```
zodern, line 48: This function is used to wrap parts of the Meteor tool where it is potentially unsafe to have async code. Since we are switching from fibers to promises, we probably should replace it with a version that make sure no promises or async/await are used.

packages/meteor/helpers.js
```diff
@@ -144,8 +141,7 @@ Meteor.wrapAsync = function (fn, context) {
       if (Meteor.isClient) {
         callback = logErr;
       } else {
-        var fut = new Future();
-        callback = fut.resolver();
+        callback = () => console.error("Tried to use wrapAsync.");
```
zodern, line 144: If not passing a callback on the server is no longer supported, it should probably throw an error here, or do the same thing as on the client and use logErr.

packages/meteor/helpers.js
```diff
@@ -144,8 +141,7 @@ Meteor.wrapAsync = function (fn, context) {
       if (Meteor.isClient) {
         callback = logErr;
       } else {
-        var fut = new Future();
```
zodern, line 147: There is some unused code related to this on line 151

packages/meteor/fiber_helpers.js
```diff
@@ -1,20 +1,5 @@
...
-  var savedYield = Fiber.yield;
-  Fiber.yield = function () {
-    throw new Error("Can't call yield in a noYieldsAllowed block!");
-  };
-  try {
-    return f();
-  } finally {
-    Fiber.yield = savedYield;
-  }
+  return f();
```
zodern, line 2: The purpose of noYieldsAllowed was to ensure that `f` did not have any async code. This type of check is still needed in some places, so it should be replaced with a version that does the same, but for promises.

tools/console/console.js
```diff
@@ -689,12 +689,13 @@ class Console extends ConsoleBase {
   // The caller should be OK with yielding --- it has to be in a Fiber and it can't be
   // anything that depends for correctness on not yielding.  You can also call nudge(false)
   // if you just want to update the spinner and not yield, but you should avoid this.
-  nudge(canYield) {
+  // TODO -> Check here
+  async nudge(canYield) {
```
zodern, line 693: This is used in some places where code has to be synchronous (where ever canYield is false). We probably need to have both `nudge` and `nudgeAsync` to support both. This also would be nice for build plugins since making `nudge` async would be a breaking change for them.

npm-packages/eslint-config-meteor/index.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
     'no-underscore-dangle': [
       'error',
       {
-        allow: ['_id', '_ensureIndex'],
+        allow: ['_id', 'ensureIndexAsync'],
```
zodern, line 28: `ensureIndexAsync` doesn't need to be listed since it doesn't start with an underscore.

npm-packages/eslint-plugin-meteor/scripts/dev-bundle-server-package.js
```diff
@@ -8,10 +8,6 @@ var packageJson = {
   name: "meteor-dev-bundle",
   private: true,
   dependencies: {
-    // Keep the versions of these packages consistent with the versions
```
zodern, line 11: Why does eslint-plugin-meteor have a copy of this file?

npm-packages/meteor-babel/options.js
```diff
@@ -183,18 +187,6 @@ function getDefaultsForNode8(features) {
       require("@babel/plugin-proposal-object-rest-spread")
     );
 
-    // Ensure that async functions run in a Fiber, while also taking
-    // full advantage of native async/await support in Node 8.
-    const isFiberDisabled = process.env.DISABLE_FIBERS === '1';
-    const ignoreAsyncPlugin = process.env.IGNORE_ASYNC_PLUGIN === '1';
-
-    if (!ignoreAsyncPlugin) {
-      combined.plugins.push([require("./plugins/async-await.js"), {
```
zodern, line 192: Note that this is a breaking change for meteor-babel.

packages/allow-deny/allow-deny.js
```diff
@@ -468,15 +454,15 @@ function docToValidate(validator, doc, generatedId) {
...
+  const validKeysRegEx = /^(?:insertAsync|updateAsync|removeAsync|fetch|transform)$/;
   Object.keys(options).forEach((key) => {
     if (!validKeysRegEx.test(key))
       throw new Error(allowOrDeny + ": Invalid key: " + key);
   });
 
   collection._restricted = true;
 
-  ['insert', 'update', 'remove'].forEach((name) => {
+  ['insertAsync', 'updateAsync', 'removeAsync'].forEach((name) => {
```
zodern, line 465: Is this breaking change necessary? It seems this is only changing the property names and it still works the same. If we want to make this change, maybe it could still support the old property names too as an alias for the new names.

packages/allow-deny/allow-deny.js
```diff
@@ -439,8 +426,7 @@ CollectionPrototype._callMutatorMethod = function _callMutatorMethod(name, args,
   }
 
   const mutatorMethodName = this._prefix + name;
-  return this._connection.apply(
-    mutatorMethodName, args, { returnStubValue: true }, callback);
+  return await this._connection.applyAsync(mutatorMethodName, args, { returnStubValue: true , throwStubExceptions: true });
```
zodern, line 429: This will probably break running multiple mutations right after each other on the client, for example:

```
Docs.insert({_id: '1'});
Docs.insert({_id: '2'});
```

zodern (commented):
I've reviewed 56 files so far (23%).

Some other things I've come across:
- There seems to be a problem with building the `ecmascript` package. It frequently fails to load due to @meteorjs/babel/utils.js missing in the built version, though the file exists in the package's source.
- I pushed a fix for how the new core-runtime handles implied packagse


There are many places in the Meteor tool where it isn't able to safely use async code. Some of these were wrapped with `noYieldsAllowed` or there was a comment documenting them, so as I come across them I am leaving a comment in this PR. However, there probably are places that are undocumented which I am concerned about.

tools/utils/utils.js
```diff
@@ -513,7 +513,7 @@ exports.isValidVersion = function (version, {forCordova}) {
 };
 
 
-exports.execFileSync = function (file, args, opts) {
+exports.execFile = async function (file, args, opts) {
```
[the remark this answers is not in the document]
zodern: At the very least, the names of the two functions should probably be swapped, so `execFile` returns the child process, and `execFileAsync` returns a promise.

npm-packages/meteor-babel/options.js
zodern: Thinking about this more, we need to keep this since some build plugins will want to support both Meteor 2 and 3 at the same time, so they will need the async-await babel plugin when compiling for older Meteor versions. Instead of removing it, we can add an option to the features object meteor-babel uses to create the babel config.

npm-packages/eslint-config-meteor/index.js
zodern: This actually is a breaking change for eslint-config-meteor. We should leave _ensureIndex here since it would still be wanted for any apps on older versions of Meteor.

tools/utils/utils.js
```diff
@@ -534,9 +534,9 @@ exports.execFileSync = function (file, args, opts) {
     });
 
     return {
-      success: ! new Promise(function (resolve) {
+      success: await !new Promise(function (resolve) {
```
zodern, line 537: `! promise` is always false. This needs to instead use `! (await ...)` so it uses the resolved value.

tools/runners/run-app.js
```diff
@@ -569,14 +566,15 @@ Object.assign(AppRunner.prototype, {
...
+        await stats.recordPackages({
           what: "sdk.run",
           projectContext: self.projectContext
         });
       }
 
-      var bundleResult = Profile.run((firstRun?"B":"Reb")+"uild App", () => {
-        return bundler.bundle({
+      var bundleResult = await Profile.run((firstRun?"B":"Reb")+"uild App", async () => {
+        return await bundler.bundle({
```
zodern, line 577: If there is an uncaught error while bundling, it sometimes logs the error and keeps bundling, instead of crashing like it is supposed to. You can reproduce this by adding `  throw new Error('linker failed');` to the top of the `fullLink` function in `linker.js`.

radekmie (commented):
I only skimmed over most of the `tools` as I'm not familiar with this codebase. If there's anything in particular you'd like me to look into, just let me know.

packages/core-runtime/load-js-image.js
```diff
@@ -0,0 +1,154 @@
...
+  var pendingNames = Object.keys(pending);
+
+  if (pendingNames.length === 0) {
+    // All packages are loaded
+    // If there were no async packages, then there might not be a promise
+    // polyfill loaded either, so we don't create a promise to return
+    return null;
+  }
+
+  return new Promise(function (resolve) {
```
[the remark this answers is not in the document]
zodern: Yes. It is explained in the comment on line 131. This line is only reached if there are packages that use top level await, and we can assume in that case that there is a promise polyfill (reify's implementation of top level await also requires a promise polyfill, which is handled by the promise and ecmascript packages).

I probably should here where the promise is created.

packages/core-runtime/package-registry.js
```diff
@@ -61,6 +61,7 @@ PRp._promise = function promise(name) {
 };
 
 // Initialize the Package namespace used by all Meteor packages.
+var global = this;
 global.Package = new PackageRegistry();
```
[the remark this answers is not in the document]
zodern: It's not supported by IE.

packages/allow-deny/allow-deny.js
```diff
@@ -410,27 +411,13 @@ CollectionPrototype._validatedRemove = function(userId, selector) {
   // Mongo to avoid races, but since selector is guaranteed to already just be
   // an ID, we don't have to any more.
 
-  return self._collection.remove.call(self._collection, selector);
+  return self._collection.removeAsync.call(self._collection, selector);
```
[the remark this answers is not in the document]
denihs: Yeah, I agree. As we review the main branch for this release, we can change those.

packages/allow-deny/allow-deny.js
denihs: For sync methods, it should be the [same](https://github.com/meteor/meteor/blob/e5e1ad2349272b8dead4b84e3f82bd99a9747609/packages/allow-deny/allow-deny.js#L617-L645). For the async ones, we can test more, but I did change some things regarding this flow, in order to fix the mongo tests, and I think we'll be fine.

packages/allow-deny/allow-deny.js
denihs: I [added to it](https://github.com/meteor/meteor/blob/e5e1ad2349272b8dead4b84e3f82bd99a9747609/packages/allow-deny/allow-deny.js#L680-L687) instead of removing the older ones. We need all of them.

packages/meteor/asl-helpers.js
```diff
@@ -6,16 +6,12 @@ Meteor._isFibersEnabled = !process.env.DISABLE_FIBERS && Meteor.isServer;
 Meteor._getAslStore = getAslStore;
 Meteor._getValueFromAslStore = getValueFromAslStore;
 Meteor._updateAslStore = updateAslStore;
+if (Meteor.isServer && !global.asyncLocalStorage) {
+    const { AsyncLocalStorage } = Npm.require('async_hooks');
```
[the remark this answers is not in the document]
denihs: > check if fibers are enabled

If you are creating something with Meteor 3, it'll be implicit that Fibers is always disabled, right? So, I think, you'll always want to have this.

packages/meteor/emitter-promise.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+  return {
+    emitter,
+    promise,
+  };
+};
+
+EmitterPromise = {
+  newPromiseResolver,
+};
```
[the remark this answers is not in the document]
denihs: 1.  I guess it was just the preference of who created it...
2. I'm not so adapted to TS. Are you willing to do it? Otherwise maybe @Grubba27 could take a look.

packages/meteor/fiber_helpers.js
denihs: I did change this function to check this [cases](https://github.com/meteor/meteor/blob/b15dcc539376a68424f44de461d0d8b57e370436/packages/meteor/async_helpers.js#L1-L7).

packages/meteor/fiber_helpers.js
```diff
@@ -136,50 +121,9 @@ class AsynchronousQueue {
...
-
-  drain() {
-    this.runWithFibers(super.drain);
-  }
-
-  _run() {
-    this.runWithFibers(super._run);
-  }
-};
+Meteor._SynchronousQueue = AsynchronousQueue;
```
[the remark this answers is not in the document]
denihs: I remove it.

packages/minimongo/local_collection.js
```diff
@@ -152,9 +162,7 @@ export default class LocalCollection {
       // Defer because the caller likely doesn't expect the callback to be run
       // immediately.
       if (callback) {
-        Meteor.defer(() => {
-          callback(null, id);
-        });
+        (async () => callback(null, id))();
```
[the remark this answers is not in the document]
denihs: [reverted](https://github.com/meteor/meteor/blob/82e3456e5b20d187b43eac8be56b0967b9bc7df3/packages/minimongo/local_collection.js#L166-L168)

packages/minimongo/minimongo_tests_client.js
```diff
@@ -71,11 +71,11 @@ Tinytest.add('minimongo - basics', test => {
...
-  fluffyKitten_id = c.insert({type: 'kitten', name: 'fluffy'});
-  c.insert({type: 'kitten', name: 'snookums'});
-  c.insert({type: 'cryptographer', name: 'alice'});
-  c.insert({type: 'cryptographer', name: 'bob'});
-  c.insert({type: 'cryptographer', name: 'cara'});
+  fluffyKitten_id = c.insertAsync({type: 'kitten', name: 'fluffy'});
+  c.insertAsync({type: 'kitten', name: 'snookums'});
+  c.insertAsync({type: 'cryptographer', name: 'alice'});
+  c.insertAsync({type: 'cryptographer', name: 'bob'});
+  c.insertAsync({type: 'cryptographer', name: 'cara'});
```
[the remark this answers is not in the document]
denihs: All these tests are now fixed.

packages/meteor/asl-helpers.js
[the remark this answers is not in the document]
zodern: Build plugins can run in older versions of Meteor, but frequently include dependencies from newer versions of Meteor. To support Meteor 3, they will have to use versions of dependencies from Meteor 3 to ensure they don't use fibers, and use those same versions in Meteor 2 which has fibers enabled.

For example, I want `zodern:melte` to work with Meteor 1.8.1 - Meteor 3. Since it uses async/await, it will need to be published with Meteor 3 to ensure it uses native async/await, and will include the versions of meteor, babel-compiler, and ecmascript from meteor 3. When the published package is run in Meteor 1 or 2, fibers will be enabled.

I did some of the work for this in https://github.com/meteor/meteor/pull/12537. The async hooks is the other main issue I am aware of that would need to be addressed.

packages/meteor/emitter-promise.js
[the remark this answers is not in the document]
radekmie: 2. It's not a must for now, and I'm definitely out of time for that 😐 But let's keep that in mind for the future - people from the community will probably pick it up.

PR #12473, opened 2023-01-25
Updating meteor mongo command

Grubba27:
In 2.11, we will update Meteor-embedded MongoDB to version 6.0.3, in which there will be no mongo shell. We needed to update the ``meteor mongo`` command. 

In this PR, I propose using the user ``mongosh`` but if it is unavailable, I show the downloads page and give him the URL so he can use it to connect.

Below there is a picture of how the result came to be:

Screenshot 2023-01-25 at 15 41 34
I do not have mongosh on my machine, which is the error displayed.

If I run `meteor mongo -U` it displays this result
Screenshot 2023-01-25 at 15 42 28

If all goes ok, this is the result that is shown
Screenshot 2023-01-25 at 15 43 29

_Side note:_
For this PR, I also updated the console colors file to accept using template literals as well because it seems more of a natural way to describe which and what colors I want to paint in my console.

### DIFF
--- a/packages/npm-mongo/.npm/package/npm-shrinkwrap.json
+++ b/packages/npm-mongo/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@aws-crypto/ie11-detection": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/ie11-detection/-/ie11-detection-2.0.2.tgz",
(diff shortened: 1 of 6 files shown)


denihs (approved):

zodern:
Is it possible to include mongosh in the dev bundle like we did for the mongo shell?

Grubba27:
> Is it possible to include mongosh in the dev bundle as we did for the mongo shell?

It is possible, but I opted to rely on the user having a ``mongosh`` in their global. If for some reason, MongoDB changes again. We are just a small change to having the shell up again.

I was not aware that we had this feature before I checked what changed with MongoDB v6, I would use `mongosh` as a go-to mongo REPL/ use MongoDB Compass, and I think many of our users do the same(talked with the core team and other Meteor user folks)

StorytellerCZ:
This will require update to Meteor University

PR #12475, opened 2023-01-26
Updating tracker types 

Grubba27:
With the addition of ``Tracker.withComputation`` in meteor 2.10, the type definitions of this new method were left missing.
In this PR I focus on adding it.

### DIFF
--- a/packages/tracker/tracker.d.ts
+++ b/packages/tracker/tracker.d.ts
@@ -110,6 +110,16 @@ export namespace Tracker {
   ): Computation;
 
   /**
+   * Helper function to make the tracker work with promises.
+   * @param computation Computation that tracked
+   * @param func async function that needs to be called and be reactive
+   */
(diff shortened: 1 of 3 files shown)


StorytellerCZ (approved):

packages/tracker/tracker.d.ts
```diff
@@ -109,6 +109,16 @@ export namespace Tracker {
     }
   ): Computation;
 
+  /**
+   * Helper function to make the tracker work with promises.
+   * @param computation Computation that tracked
+   * @param func async function that needs to be called and be reactive
+   */
+  function withComputation<T>(
+    computation: Computation,
```
radekmie, line 118: It should say `Computation | null`, as otherwise it won't be isomorphic. Also, `currentComputation` should have the same type (it's `Computation` now).

PR #12478, opened 2023-01-26
Made standard-minifier-css debuggable

softwarecreations:
Thank you for contributing to Meteor. (My all-day every day stack since 2015)
Thank you for taking the time to review my PR!
Feedback and suggestions are welcome and appreciated.

# Scope
Made minifyStdCSS aka standard-minifier-css debuggable.

## How standard-minifier-css becomes verbose
* Either of the common debugging commandline arguments
  * `--verbose`
  * `--debug`
* Environment variable 
  * `DEBUG_CSS`

# Why I created this PR - TLDR
I suffered terribly because if a postcss-plugin goes into an infinite loop, the whole build process gets stuck and there's zero indication from the meteor build tool
* Whether the build process has stalled or not
* What part of the build is having trouble
Etc, `--debug` and `--verbose` had zero effect on `standard-minifier-css` yet it's probably the most likely part of the build process to fail. And if it fails, there's no way to see which file it failed on, whether it even failed while reading a file, or perhaps for some other reason, like a dependency, new bug or whatever.

## I made this PR because
* In my experience with Meteor full-time since 2015, if I can't build, it's usually a SCSS problem (with zero visibility which file etc)
* I like this debug feature that I've added and I intend to leave it enabled, most likely.
* I want this debug feature available for myself and my team
* I don't want other developers to suffer like I did. I want people to use and love Meteor, and not abandon it out of frustration as so many developers have.

# 264 issues mentioning standard-minifier-css
https://github.com/meteor/meteor/issues?q=is%3Aissue+standard-minifier-css
(granted not all of those issues will be about standard-minifier-css) but no-doubt a huge number of past and future issues, (both reported issues and unreported issues that developers silently struggle with) will be involving standard-minifier-css in some way; and it's not even debuggable at all. Literally can't see anything at all. It's currently like theoretical dark-matter.

I'm willing to bet that adding this feature will reduce the number of issues and give developers a better, quicker, easier experience. And free up all the Meteor contributors to contribute more to Meteor rather than helping with issues that could be prevented or getting slowed down by friction in the development experience that we can eliminate.

# How to prevent standard-minifier-css being verbose
`DEBUG_CSS=false` or `DEBUG_CSS=0` will prevent it from being verbose regardless of `--verbose` or `--debug` commandline arguments, because `DEBUG_CSS` is specific.

# Style
I love Meteor. I work with it all day every day. I've brought 3 of my biggest long-term web-dev clients and all their developers to Meteor.
I really want this package to be debuggable. If you don't like the style that I coded in, just tell me what style you want etc and I'll change it to meet your preferences. As long as it's debuggable, that's great.

# Why I created this PR - Detailed version
Prior to this PR `standard-minifier-css` was an absolute black box.

Recently I had not done a build for about a month, and my meteor build process was getting stuck after long delays, providing no feedback to me if the build process was even stuck or not. I just noticed 100% CPU usage and the `/ - \ |` spinner in the terminal seemed to stop spinning or be taking very long.

`--debug` and `--verbose` provided zero benefit to me.
It was incredibly frustrating and wasted so much of my time, checking out old/new versions of my project and reinstalling meteor and all packages many times before I realized via a process of randomly eliminating NPM modules and meteor packages that the problem was happening inside the standard-minifier-css.

From there I went on a wild-goose chase trying older versions of standard-minifier-css. Then I finally arrived at a theory that the problem must be triggered/caused by one of my CSS files, but I have a huge amount of CSS files and I had no way of knowing which file it was getting stuck on. I tried to manually process my SCSS files individually with postcss-cli, but couldn't get postcss-cli working with required plugins. Finally I had the idea to add functionality to standard-minifier-css to make it debuggable.

Now that standard-minifier-css has a verbose mode, the file that was having trouble being processed is revealed instantly, effortlessly and intuitively to any developer who's having trouble.

The actual problem is a postcss plugin was going into an infinite loop, but only during builds, due to a missing environment variable for the build (and lack of throwing an error when environment variable was missing) These are trivial problems to fix, but only when it's possible to debug CSS builds, which is what this PR does.

### DIFF
--- a/packages/standard-minifier-css/package.js
+++ b/packages/standard-minifier-css/package.js
@@ -9,7 +9,8 @@ Package.registerBuildPlugin({
   name: "minifyStdCSS",
   use: [
     'minifier-css',
-    'ecmascript'
+    'ecmascript',
+    'logging',
   ],
   npmDependencies: {
     "@babel/runtime": "7.18.9",
@@ -25,4 +26,5 @@ Package.registerBuildPlugin({
 Package.onUse(function(api) {
   api.use('minifier-css@1.5.4');
   api.use('isobuild:minifier-plugin@1.0.0');
+  api.use('logging@1.3.1');
 });
--- a/packages/standard-minifier-css/package.js
+++ b/packages/standard-minifier-css/package.js
@@ -2,7 +2,7 @@ Package.describe({
   name: 'standard-minifier-css',
   version: '1.9.0',
   summary: 'Standard css minifier used with Meteor apps by default.',
-  documentation: 'README.md'
+  documentation: 'README.md',
 });
 
 Package.registerBuildPlugin({
@@ -16,10 +16,10 @@ Package.registerBuildPlugin({
     "@babel/runtime": "7.18.9",
     "source-map": "0.7.4",
     "lru-cache": "6.0.0",
-    "micromatch": "4.0.5"
+    "micromatch": "4.0.5",
   },
   sources: [
-    'plugin/minify-css.js'
+    'plugin/minify-css.js',
   ]
 });
 
--- a/packages/standard-minifier-css/package.js
+++ b/packages/standard-minifier-css/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'standard-minifier-css',
-  version: '1.8.3',
+  version: '1.8.4',
   summary: 'Standard css minifier used with Meteor apps by default.',
   documentation: 'README.md'
 });
--- a/packages/standard-minifier-css/package.js
+++ b/packages/standard-minifier-css/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'standard-minifier-css',
-  version: '1.8.4',
+  version: '1.9.0',
   summary: 'Standard css minifier used with Meteor apps by default.',
   documentation: 'README.md'
 });
--- a/packages/standard-minifier-css/plugin/minify-css.js
+++ b/packages/standard-minifier-css/plugin/minify-css.js
@@ -3,8 +3,9 @@ import { createHash } from "crypto";
 import LRU from "lru-cache";
 import { loadPostCss, watchAndHashDeps, usePostCss } from './postcss.js';
 
-const verbose = (process.env.DEBUG_CSS!=="false" && process.env.DEBUG_CSS!=="0" && (
-  process.env.DEBUG_CSS || process.argv.indexOf('--verbose') > -1 || process.argv.indexOf('--debug') > -1
+const { argv, env:{ DEBUG_CSS } } = process;
+const verbose = (DEBUG_CSS!=="false" && DEBUG_CSS!=="0" && (
+  DEBUG_CSS || argv.indexOf('--verbose') > -1 || argv.indexOf('--debug') > -1
 ));
 
 Plugin.registerMinifier({
--- a/packages/standard-minifier-css/plugin/minify-css.js
+++ b/packages/standard-minifier-css/plugin/minify-css.js
@@ -11,7 +11,7 @@ const verbose = (DEBUG_CSS!=="false" && DEBUG_CSS!=="0" && (
 
 Plugin.registerMinifier({
   extensions: ["css"],
-  archMatching: "web"
+  archMatching: "web",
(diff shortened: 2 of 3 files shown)


packages/standard-minifier-css/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   name: 'standard-minifier-css',
-  version: '1.8.3',
+  version: '1.8.4',
```
StorytellerCZ, line 3: Should be feature update

packages/standard-minifier-css/plugin/minify-css.js
```diff
@@ -47,11 +58,27 @@ class CssToolsMinifier {
...
     ) {
+      if (verbose && !this.haveHitAnyCache) {
+        this.haveHitAnyCache = true;
+        setTimeout( () => { // we use a timeout to give all files a chance to finish being minified
+          const stats = [`minifyStdCSS: Total CSS ${this.formatSize(this.totalSize)}`];
+          if (this.totalMinifiedSize!==0) {
+            stats.push(`minified ${this.formatSize(this.totalMinifiedSize)}`);
+            stats.push(`reduction ${Math.round(100-this.totalMinifiedSize*100/this.totalSize)}%`);
+          }
+          console.log(stats.join(", "));
```
StorytellerCZ, line 69: I would prefer the usage of the Meteor's logging package for this, but this is fine too.

packages/standard-minifier-css/plugin/minify-css.js
```diff
@@ -81,13 +115,15 @@ class CssToolsMinifier {
     const { error, postcssConfig } = await loadPostCss();
 
     if (error) {
+      if (verbose) console.log('processFilesForBundle loadPostCss error', error);
```
StorytellerCZ, line 118: Again, would prefer to use the logging package or at least `console.error` I think would work better here.
https://docs.meteor.com/packages/logging.html

StorytellerCZ (approved):
Only minor nitpicks, LGTM.

packages/standard-minifier-css/plugin/minify-css.js
softwarecreations: Ok, interesting, I have no idea how to do that. Can you please show me some example of code that does it or point me in the right direction?

I'm happy to leave it as is or use the logging feature. I'm personally not expecting verbose modes to be the prettiest thing ever. But it's usually better to do _the proper thing_™ if it's doesn't impact on _getting things done_™.

I had a look at Meteor docs now, but none of these looked appropriate
https://docs.meteor.com/packages/logging.html
https://galaxy-guide.meteor.com/app.html#logs

packages/standard-minifier-css/plugin/minify-css.js
softwarecreations: Oh, I guess I should have read all of your comments first. I see you're recommending the logging package.

I assumed that the user would need to add the logging package to be able to see anything written with Log. But I will check it out and revert.
It would hurt developer experience if they enable `--verbose` mode, but can't see anything if they haven't installed the logging package.

packages/standard-minifier-css/package.js
softwarecreations: Sorted! 1.9.0 Let's go!

softwarecreations:
@StorytellerCZ It's ready for you to have another look. Thanks!

packages/standard-minifier-css/package.js
StorytellerCZ: :tada: :comet:

packages/standard-minifier-css/plugin/minify-css.js
StorytellerCZ: Thank you for textbook answer to my comment. Yes you are correct. Let's keep the console log there for now.

StorytellerCZ (approved):

Grubba27:
> Thanks @Grubba27, @StorytellerCZ it looks like you need to approve again for the 2.11 branch Grubba27 selected?

Yeah, now Github makes reviews stale because of branch changes, Will ask @denihs a review as well

StorytellerCZ (approved):

denihs (approved):

PR #12482, opened 2023-01-30
Updating react skeletons

Grubba27:
Since Meteor.js 2.9 in this [PR](https://github.com/meteor/meteor/pull/12220) we do not have insecure package by default anymore.
Because of this, our skeletons were a little bit outdated, In this PR, I start solving this issue, starting with the React packages.

### DIFF
--- a/tools/static-assets/skel-chakra-ui/imports/ui/Info.jsx
+++ b/tools/static-assets/skel-chakra-ui/imports/ui/Info.jsx
@@ -1,22 +1,25 @@
 import React from 'react';
-import { useTracker } from 'meteor/react-meteor-data';
+import { useFind, useSubscribe } from 'meteor/react-meteor-data';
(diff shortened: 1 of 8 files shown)


henriquealbert (approved):
Remember to update the tutorial as well

zodern:
It will be nice to have this fixed.

Meteor shows a large warning when creating publications in an app that has autopublish. The publication and subscribe code should probably not be added when creating the app with the `--prototype` option.

Grubba27:
> Meteor shows a large warning when creating publications in an app that has autopublish. The publication and subscribe code should probably not be added when creating the app with the --prototype option. 

I would love to have this, as well, with the --typescript flag, but we currently have some limitations regarding our CLI, I want to invest more time in the tools side, but we have fibers/ meteor 3 migration in our hands now

PR #12486, opened 2023-01-31
Changelog generator

Grubba27:
This is a change to improve the Maintainer's quality of development.
I work on a powerful machine, but because of the number of lines and the way the markdown LSP works in my IDE, it can get really slow, making my development slower. To this day, I can easily count more than 3 hours wasted because of this.

In this small PR, I add a way to organize + make it lighter to more users to update the changelog


docs/generators/changelog/script.js
```diff
@@ -0,0 +1,23 @@
...
+const fs = _fs.promises;
+const main = async () => {
+  try {
+    console.log('started concatenating files');
+    const files = await fs.readdir('./generators/changelog/versions', 'utf8');
+    const filesStream = files
+      .map(file => {
+        console.log(`reading file: ${ file }`);
+        return fs.readFile(`./generators/changelog/versions/${ file }`, 'utf8');
+      })
```
NicolasLopes7, line 11: instead of reading each file sequentially you can use fs read streams to not block the i/o. 
Check out this [article](https://stateful.com/blog/process-large-files-nodejs-streams) to know how to deal with streams in nodejs and fs.

Grubba27: I can be asured that the order will be maintained in this way?

PR #12488, opened 2023-02-01
Feat: Upgrading MongoDB driver to v5.0.0

Grubba27:
## _still in discussion_

We are still deciding if these changes should come to 2.11 or 3.1.0

For a full detail of what has changed, I recommend checking their [changelog](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.0.0)(we will have of what changes for meteor when we finish this upgrade)


henriquealbert:
imo would be better to release it on 3.1 due to the breaking changes that it will introduce and the lack of time - being 3.0 exclusive for fibers migration.

fredmaiaarantes:
Quickly looking at the MongoDB Driver Migration Guide and its breaking changes, I would say we should only do it in version 3.x. I don't think it will be a simple change to Meteor.js, so maybe work on that in 3.1. With this, we can keep focused on fibers removal.

PR #12489, opened 2023-02-02
Updating mongodb types for v6

Grubba27:
In this PR I remove operator types that were removed in MongoDB v6.
For more information check [MongoDB Changelog](https://www.mongodb.com/docs/manual/release-notes/6.0-compatibility/#removed-operators)


StorytellerCZ (approved):

packages/mongo/mongo.d.ts
```diff
@@ -18,22 +18,12 @@ export namespace Mongo {
...
-    $hint?: Hint;
+  export type QueryWithModifiers = {
     $maxScan?: any;
-    $max?: any;
-    $maxTimeMS?: any;
-    $min?: any;
-    $orderby?: any;
-    $returnKey?: any;
-    $showDiskLoc?: any;
     $natural?: any;
```
radekmie, line 23: Interestingly, the official driver no longer has typings for any these. Maybe we should remove it too?

radekmie (approved):

packages/mongo/mongo.d.ts
radekmie: Yeah, I meant exactly that. But now we have an absurd situation, where we have a `Query` type that's an alias for driver's `Filter` and the only place we use it is our `Selector` type 😛 

I'm not sure what we should do exactly, but I think that staying as close to the driver's types and naming would be beneficial.

Grubba27: Agreed! thanks for the insights :)

Grubba27: I hope now looks good!

PR #12492, opened 2023-02-06
Adding throttle to subscriptions

Grubba27:
# This is still under development

What is missing?
- [ ] Docs
- [ ] More tests

## Why was this developed?

This is a feature request from @vlasky from [here](https://github.com/meteor/meteor-feature-requests/issues/367)
This sounded like an easy-win feature, but from the testing perspective, I'm feeling a little bit stuck; I'm unable to make it work properly; maybe I've implemented it incorrectly.

For the video showing that worked is below(there is a lot of room for improvement, and in this PR, I hope I can get some feedback to achieve the ideal):

https://user-images.githubusercontent.com/70247653/217003864-05bd3d11-e743-4ffc-bfaa-37019b3a8843.mov

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -1053,6 +1053,13 @@ var Subscription = function (
 
   self._params = params || [];
 
+  // The last time we sent a message to the client.
+  // used by throttleSubscription option
+  self._lastUpdate = 0;
+
+  // The minimum interval between messages to the client.
+  // used by throttleSubscription option
+  self.minInterval = null;
   // Only named subscriptions have IDs, but we need some sort of string
   // internally to keep track of all subscriptions inside
   // SessionDocumentViews. We use this subscriptionHandle for that.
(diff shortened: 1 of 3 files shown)


packages/ddp-server/livedata_server.js
```diff
@@ -1330,6 +1337,29 @@ Object.assign(Subscription.prototype, {
     return self._deactivated || self._session.inQueue === null;
   },
 
+  setThrottle: function (minInterval) {
+    const self = this;
```
radekmie, line 1341: I know there's this `self` convention in this file (it's rather old), but I wouldn't do that for the new code.

packages/ddp-server/livedata_server.js
```diff
@@ -1352,8 +1382,10 @@ Object.assign(Subscription.prototype, {
       }
       ids.add(id);
     }
-
-    this._session.added(this._subscriptionHandle, collectionName, id, fields);
+    this._runWithThrottling(
+       () => this._session.added(this._subscriptionHandle, collectionName, id, fields)
+    )
+    this._lastUpdate = Date.now();
```
radekmie, line 1388: 1. As every `_runWithThrottling` call site sets `_lastUpdate`, it could happen in this function.
2. What if the subscription stopped before the `setTimeout` is triggered?

packages/ddp-server/livedata_server_tests.js
```diff
@@ -286,6 +286,63 @@ Meteor.publish("publication_compatibility", function () {
   this.stop();
 });
 
+const SimpleColl = new Mongo.Collection("simple");
+
+Meteor.publish('publication_with_thottle', function () {
```
radekmie, line 291:
```suggestion
Meteor.publish('publication_with_throttle', function () {
```

radekmie (changes requested):
This is **not** throttling per-se, as the number of operations is the same - they're just delayed.

packages/ddp-server/livedata_server.js
Grubba27: I've used this self mostly because of conventions + I don't particularly appreciate using ``this`` inside a `function` (the one inside setTimeout).
I will make the adjustments

packages/ddp-server/livedata_server.js
Grubba27: > What if the subscription stopped before the setTimeout is triggered?

Not thought about this case.

Maybe moving the ``_lastUpdate`` statement to look like this:

```js
    this._runWithThrottling(
       () => { 
            this._session.added(this._subscriptionHandle, collectionName, id, fields);
            this._lastUpdate = Date.now();
      }
    )

```

Grubba27:
How would I throttle the ops? Should it be done before the add ops in `packages/mongo/collection.js`? or even before that? 
I thought I had some way of doing this natively with MongoDB. The only resource that could be useful is the Bulk API but in this matter I don't think is as useful as making my own solution

Grubba27:
Will be closing this one for now.

PR #12495, opened 2023-02-07
Fixed changePassword and resetPassword functions type checking conditions

### LINKED ISSUE #12454: Accounts.resetPassword function has a faulty check for whether token and passwords are strings
This issue is located in `packages/accounts-password/password_client.js` in Meteor 2.10.

The line at fault (and there are others like it) reads `if (!token instanceof String)`

the ! operand has a precedence of 14, while the instance of operator has a precedence of 9. which would equate the above to `false instanceof String`, which will never be true.

In addition to that, checking `instanceof String` only checks if it is a String object, which is rarely the case in js.

a correct solution would be

`if (!(typeof token === "string" || token instanceof String))`

Ujjwal-S:
Fixes #12454, by implementing the suggested change. The issue is indeed true. Look at the example below.

217182201-7fe72c91-a1cd-4449-a810-d2a13134261c

In the devtools I have created a variable `newPassword` which is clearly a number and not a String. However the clause `if (!newPassword instanceof String) {...}` fails to detect this. This happens because of operator precedence.

- `!` has a higher precedence than `instanceof` and thus `!` is always evaluted first. Hence `!newPassword` will always return false, and _false_ is not an instance of _String_, thus the if condition check never runs.

- The issue also mentioned about the infrequent use of String object. So as proposed in the issue itself, I made appropriate changes in the `changePassword` and `resetPassword`.

Please be kind enough to tell me if there is anything that I did wrong.

### DIFF
--- a/packages/accounts-password/password_client.js
+++ b/packages/accounts-password/password_client.js
@@ -142,7 +142,7 @@ Accounts.changePassword = (oldPassword, newPassword, callback) => {
     return reportError(new Error("Must be logged in to change password."), callback);
   }
 
-  if (!newPassword instanceof String) {
+  if (!(typeof newPassword === "string" || newPassword instanceof String)) {
     return reportError(new Meteor.Error(400, "Password must be a string"), callback);
   }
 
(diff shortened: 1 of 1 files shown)


Grubba27 (approved):
Sounds good

StorytellerCZ (approved):
LGTM

PR #12504, opened 2023-02-08
Release 3.0 -  tests ddp client and minimongo

denihs:
In this PR we:

- Fix tests in the ddp-client
image

- Fix tests in the minimongo
image

- keep Meteor.call and Meteor.appy sync


packages/autoupdate/autoupdate_server.js
```diff
@@ -52,7 +52,7 @@ Autoupdate.autoupdateVersionRefreshable = null;
 Autoupdate.autoupdateVersionCordova = null;
 Autoupdate.appId = __meteor_runtime_config__.appId = process.env.APP_ID;
 
-var syncQueue = new Meteor._SynchronousQueue();
+var syncQueue = new Meteor._AsynchronousQueue();
```
Grubba27, line 55: Maybe we should name it asyncQueue or just queue

packages/ddp-client/test/livedata_connection_tests.js
```diff
@@ -746,11 +772,11 @@ Tinytest.add('livedata stub - method call before connect', function(test) {
   });
 });
 
-Tinytest.addAsync('livedata stub - reconnect', async function(test, onComplete) {
+Tinytest.addAsync('livedata stub - reconnect111', async function(test, onComplete) {
```
Grubba27, line 775: maybe typo ?

it could be just reconnect

Grubba27 (approved):
Looks good. 
Just nitpicks

packages/autoupdate/autoupdate_server.js
zodern: The queue only allows a single task to run at a time, which is why it was called a synchronous queue. The name wasn't referring to if the tasks themselves were sync or async; it always supported async tasks.

Grubba27: ohhhhhh got it. Thanks for the clarification

packages/ddp-client/test/livedata_connection_tests.js
denihs: yes, typo

PR #12505, opened 2023-02-09
Fix TypeScript type dependencies for mongo, webapp, and underscore

ebroder:
A newly created TypeScript Meteor project today will fail to type check. This is for a handful of reasons, although this PR only addresses a subset of them (stay tuned for more!).

In this PR, we deal with the fact that some of the type declarations for Meteor packages require type declarations for upstream NPM packages. If a Meteor package depends on an NPM package and that NPM package includes type declarations directly, then everything works fine. However, if the NPM package doesn't include type declarations, and instead type declarations have to come from a DefinitelyTyped `@types` package, then we have a problem.

Out of the box, a newly created TypeScript template project sees an assortment of errors caused by this:

```
evan@mathias meteor-types-test (main) % meteor npx tsc
[...]
.meteor/local/types/node_modules/package-types/mongo/package/os/packages/mongo/mongo.d.ts:1:33 - error TS2307: Cannot find module 'mongodb' or its corresponding type declarations.

1 import * as MongoNpmModule from 'mongodb';
                                  ~~~~~~~~~

.meteor/local/types/node_modules/package-types/mongo/package/os/packages/mongo/mongo.d.ts:9:8 - error TS2307: Cannot find module 'mongodb' or its corresponding type declarations.

9 } from 'mongodb';
         ~~~~~~~~~

.meteor/local/types/node_modules/package-types/underscore/package/os/packages/underscore/underscore.d.ts:1:20 - error TS2307: Cannot find module 'underscore' or its corresponding type declarations.

1 import * as _ from 'underscore';
                     ~~~~~~~~~~~~

.meteor/local/types/node_modules/package-types/webapp/package/os/packages/webapp/webapp.d.ts:2:26 - error TS7016: Could not find a declaration file for module 'connect'. '/Users/evan/src/meteor-types-test/.meteor/local/types/node_modules/package-types/webapp/node_modules/connect/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/connect` if it exists or add a new declaration (.d.ts) file containing `declare module 'connect';`

2 import * as connect from 'connect';
                           ~~~~~~~~~
```

For `underscore` and `webapp`, just add the `@types/` packages directly as dependencies. Meteor has no concept of a "dev dependency", but as these files aren't required in any way at runtime, they'll increase the size of the package's isopacks but not change the resulting application. (I also cleaned up `packages/webapp/.npm`, since the current `.gitignore` file implies that it never should have been committed in the first place)

For `mongo`, I opted for something slightly more complicated. The `npm-mongo` package already has `mongodb` as an NPM dependency (which _does_ include its own type declarations). However, because `npm-mongo` doesn't itself export type declarations, the types from `mongodb` aren't made accessible. Therefore, I added type declarations to `npm-mongo` and rephrased the `mongo` package to import the `mongodb` types from `npm-mongo`.

### DIFF
--- a/packages/npm-mongo/package.js
+++ b/packages/npm-mongo/package.js
@@ -17,4 +17,5 @@ Package.onUse(function (api) {
     "NpmModuleMongodb",
     "NpmModuleMongodbVersion",
   ], "server");
+  api.addAssets('index.d.ts', 'server');
 });
--- a/packages/mongo/mongo.d.ts
+++ b/packages/mongo/mongo.d.ts
@@ -1,12 +1,4 @@
-import * as MongoNpmModule from 'mongodb';
-import {
-  Collection as MongoCollection,
-  CreateIndexesOptions,
-  Db as MongoDb,
-  Hint,
-  IndexSpecification,
-  MongoClient,
-} from 'mongodb';
+import { NpmModuleMongodb } from 'meteor/npm-mongo';
 import { Meteor } from 'meteor/meteor';
 
 // Based on https://github.com/microsoft/TypeScript/issues/28791#issuecomment-443520161
@@ -16,13 +8,13 @@ export type UnionOmit<T, K extends keyof any> = T extends T
 
 export namespace Mongo {
 
-  export type Selector<T> = MongoNpmModule.Filter<T>;
+  export type Selector<T> = NpmModuleMongodb.Filter<T>;
 
-  type Modifier<T> = MongoNpmModule.UpdateFilter<T>;
+  type Modifier<T> = NpmModuleMongodb.UpdateFilter<T>;
 
   export type OptionalId<TSchema> = UnionOmit<TSchema, '_id'> & { _id?: any };
 
-  type SortSpecifier = MongoNpmModule.Sort;
+  type SortSpecifier = NpmModuleMongodb.Sort;
 
   export interface FieldSpecifier {
     [id: string]: Number;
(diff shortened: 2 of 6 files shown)


packages/npm-mongo/package.js
```diff
@@ -17,4 +17,5 @@ Package.onUse(function (api) {
     "NpmModuleMongodb",
     "NpmModuleMongodbVersion",
   ], "server");
+  api.addAssets('npm-mongo.d.ts', 'server');
```
radekmie, line 20: I'd rather call it `index.d.ts`, just for the convention. _(We're actually discussing it [on Slack](https://meteor-community.slack.com/archives/CS46CNDN0/p1675929707227739?thread_ts=1673006401.030039&cid=CS46CNDN0) too.)_

radekmie (commented):
What happened to `pacakges/webapp/.npm`?

ebroder:
> What happened to `pacakges/webapp/.npm`?

Hmm. I noticed that the `.gitignore` file had:

```
# packages shouldn't have .npm on Git
packages/**/.npm
```

I also wasn't entirely sure how to update the `.npm` directory. Since it seemed like it shouldn't have been there in the first place, I figured that it was easier to remove it.

zodern:
We should be committing the .npm folders. I think that rule was added to .gitignore because Meteor had an issue for a while with leaving behind garbage directories in .npm. If the problem still exists, the rule could be changed to only exclude the garbage directories.

The webapp .npm/shrinkwrap.json was manually modified to remove a dependency that doesn't support Windows, so we need to keep the file in git.

ebroder:
Got it. I added the `.npm` directory back in and updated it with the new dependency. I'll follow up and remove the `.npm` from the `.gitignore` since it seems like that's an error.

I also went ahead and rebased onto the `release-2.11` branch, since there's a conflict there in the mongo type declarations.

PR #12507, opened 2023-02-09
Add type definitions for async methods from Meteor 2.9

ebroder:
This pairs with the change in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64249 (cc @radekmie)


packages/accounts-base/accounts-base.d.ts
```diff
@@ -33,6 +37,16 @@ export namespace Accounts {
     callback?: (error?: Error | Meteor.Error | Meteor.TypedError) => void
   ): string;
 
+  function createUserAsync(
+    options: {
```
Grubba27, line 41: Should not be better just
```typescript
  options: {
    username?: string;
    email?: string
    password?: string;
    profile?: Object;
  }
```
How it is just fine. I'm asking because I always just used `?` as the same as optional/undefined

ebroder: Well, I mostly just copied what was already there.

There is _technically_ a difference if you have the TypeScript option `exactOptionalPropertyTypes` enabled - in that case, `username?: string` means that username can be present or absent, but _if_ it's present it can _not_ be `undefined`. But as far as I know usage of that option is pretty rare (because lots of people assume that present-and-undefined is equivalent to absent).

At some point, the Meteor types probably need some updates to be compatible with `exactOptionalPropertyTypes`, but that's going to be a bigger chunk of work.

radekmie: Yep, it was added to all of the DefinitelyTyped types when `exactOptionalPropertyTypes` got introduced for compatibility. Ideally, we'd use only `?` as it usually doesn't make sense to pass explicit `undefined`s everywhere.

PR #12510, opened 2023-02-10
Switch typescript skeleton to zodern:types and test that it works

ebroder:
As type definitions are moved to be shipped directly with the core Meteor packages, the @types/meteor NPM package is trending towards deprecation, so we should steer people to use the preferred way of pulling in type definitions.

Additionally, add a test to make sure we're not steering them too far astray.


Grubba27 (approved):
LGTM

PR #12511, opened 2023-02-10
Update coffeescript to 2.7.0

zodern:
This enables many new js features in coffeescript, and removes the workaround Meteor used to prevent coffeescript from overriding `Error.prepareStackTrace`.

Publishing a new version of the coffeescript package will also include the caching fix from Meteor 1.10.2.


Grubba27:
This can be an independent release(it does not need to be shipped with a 2.11.0). Will Try making a release as soon as possible

Grubba27:
2.7.0 CoffeeScript is out!

Grubba27 (approved):

PR #12530, opened 2023-03-02
Release 3.0 - Tools/Remove local cache before each command is executed

henriquealbert:
Every time we need to run a command for the **second** time, things get messy on the local cache, and for now, it's not able to get the correct files, giving us an error.

This is a workaround to avoid that and make sure that we don't use the `.meteor/local` cache.
It will take a bit longer to run every command.

We plan to fix this as soon as possible on the next alpha version.

Card: https://favro.com/organization/929aab32206e9caa0b59b5cf/cef8866673a006c0854ffd64?card=Met-2989

### DIFF
--- a/packages/babel-compiler/.npm/package/npm-shrinkwrap.json
+++ b/packages/babel-compiler/.npm/package/npm-shrinkwrap.json
@@ -5,382 +5,618 @@
       "version": "7.19.0-beta.1",
       "dependencies": {
         "@ampproject/remapping": {
-          "version": "2.1.1"
+          "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


Grubba27 (approved):
If this was tested with commands like run,test, and lint.
LTGM!

henriquealbert:
> If this was tested with commands like run,test, and lint. LTGM!

Yes, it was tested! 🙌🏻

henriquealbert:
This one https://github.com/meteor/meteor/pull/12535 solves the issue. So there is no need for this workaround.

PR #12533, opened 2023-03-03
Add TS types for Mongo Collection countDocuments and estimatedDocumentCount

ArthurHoaro:
Types were missing, but both methods are implemented in `collection.js` and documented here: https://docs.meteor.com/api/collections.html#Mongo-Collection-countDocuments

Also, `estimatedDocumentCount` does not take a selector as parameter. See:

  * https://www.mongodb.com/docs/drivers/node/current/usage-examples/count/
  * https://mongodb.github.io/node-mongodb-native/5.1/classes/Collection.html#estimatedDocumentCount


packages/mongo/mongo.d.ts
```diff
@@ -195,6 +195,17 @@ export namespace Mongo {
...
+     * Gets the number of documents matching the filter. For a fast count of the total documents in a collection see `estimatedDocumentCount`.
+     * @param selector The query for filtering the set of documents to count
+     * @param options All options are listed in [MongoDB documentation](https://mongodb.github.io/node-mongodb-native/4.11/interfaces/CountDocumentsOptions.html). Please note that not all of them are available on the client.
+     */
+    countDocuments(selector?: Selector<T> | ObjectID | string, options?: MongoNpmModule.CountDocumentsOptions): Promise<number>;
+    /**
+     * Gets an estimate of the count of documents in a collection using collection metadata. For an exact count of the documents in a collection see `countDocuments`.
+     * @param options All options are listed in [MongoDB documentation](https://mongodb.github.io/node-mongodb-native/4.11/interfaces/CountDocumentsOptions.html). Please note that not all of them are available on the client.
+     */
+    estimatedDocumentCount(options: MongoNpmModule.EstimatedDocumentCountOptions): Promise<number>;
```
radekmie, line 208: The `options` should be an optional parameter.
```suggestion
    estimatedDocumentCount(options?: MongoNpmModule.EstimatedDocumentCountOptions): Promise<number>;
```

radekmie (changes requested):

packages/mongo/mongo.d.ts
ArthurHoaro: Ah missed that! Thanks!

radekmie (approved):

StorytellerCZ (approved):

PR #12534, opened 2023-03-03
[Meteor 3] Improve error handling with top level await

zodern:
This updates the TLA implementation to handle errors in the top level of modules. Most of the work was in https://github.com/meteor/reify/pull/4.

Every module is now wrapped in a try ... catch block so reify can synchronously handle errors as they happen.

Meteor's integration has been updated to restore the expected behavior for unhandled errors in the root of the modules. On the server, it will now properly now crash with an uncaught exception. On the client, the error will now become an uncaught exception instead of being an unhandled promise rejection.


henriquealbert (approved):

Grubba27:
Solved conflict (accepting from this PR) as its npm-packages/meteor-babel/package-lock.json were changed to use other deps

PR #12536, opened 2023-03-03
[Meteor 3] Fix missing globals and packages unable to import from app's node_modules

zodern:
When there was no app code, the linker in Meteor 3 wouldn't add any linked files for the app. This caused two problems:

1. On the server, Meteor would only resolve npm packages in the app's node_module's folder if there were any linked files for the app. When running package tests, this broke importing @babel/runtime.
2. The globals Meteor would always create for the app were missing, which would be noticeable when using a debugger or the Meteor shell.

When top level await is enabled, globals are defined in the header that is added to the app file. If there is no file for the app, it will now add an empty file to ensure the header is still added.


henriquealbert (approved):
Tested and it's working 💪🏻

PR #12537, opened 2023-03-03
[Meteor 3] Adjustments to detecting if fibers are enabled

zodern:
Some build plugins will want to support Meteor 1-2 with fibers and Meteor 3 without fibers. When build plugins are published, they have their own copy of dependencies and use them when running in older or newer Meteor versions. We will want the packages normally used by build plugins to still run in older versions of Meteor.

This makes some adjustments to partially support it.

1. So the Meteor package can detect if fibers are disabled, this sets `__meteor_bootstrap__.fibersDisabled` to true. Older versions of Meteor that use fibers will have a falsey value for `fibersDisabled` (it will be undefined).
2. Adds `Meteor.fibersDisabled`.
3. Re-adds the async-await plugin to meteor-babel
4. Adds an option to the feature object in meteor-babel to control when the async-await plugin is used
5. The babel-compiler Meteor package enables the async-await plugin whenever fibers are enabled.

As a side effect of these changes, it is now possible to run Meteor 3 without having to set any environment variables.


henriquealbert:
How can we test it?

Grubba27:
Hey @zodern how could I test this one?
run the same command without the usual flags and check if works correctly?

It would be nice to have documented a way on how to use this new feature that was added here

zodern:
This adds two things:
1. `Meteor.isFibersDisabled` which is based on the Meteor version instead of environment variables
2. Adds `useNativeAsyncAwait` option to meteor-babel to configure if it rewrites async/await to use fibers. With Meteor 3, it should always set this option to true.

If the changes work, it will always use native async/await in Meteor 3 without having to set any environment variables, and babel-compiler will rewrite async/await when running inside a build plugin in an older Meteor version. I haven't tested the second part since there are other issues that need to be fixed before it can be tested.

PR #12538, opened 2023-03-04
[Meteor 3] Partial fix for stylesheets

zodern:
This is a partial fix for non-lazy stylesheets not being bundled. I don't fully understand how the error handling is supposed to work, and I'm not sure the previous behavior is correct, so I need to look into it further before fully fixing it.

This also adjusts some of the colors for the new tinytest dark theme to adjust the contrast to make the text easier to read and reduce how much some of the borders stand out.


henriquealbert (approved):
I don't think `addStylesheet` needs to be async as well.
At least when running the tests, this solved the issue of the CSS not being loaded

Grubba27:
I asked @henriquealbert to make a card so that we can review this later(it is in our public [Favro](https://favro.com/widget/929aab32206e9caa0b59b5cf/27d714839b9854618835df20), and it won't be missed or forgotten).

PR #12541, opened 2023-03-07
Update NPM packages Node versions

StorytellerCZ:
Update Node version on tests on NPM packages.

I've upped it to Node 14 and 16, but maybe for Meteor 3 we should up it for Node 16 and 18?


henriquealbert:
> Update Node version on tests on NPM packages.
> 
> I've upped it to Node 14 and 16, but maybe for Meteor 3 we should up it for Node 16 and 18?

I think it makes sense to be 16 and 18

StorytellerCZ:
@henriquealbert updated!

PR #12542, opened 2023-03-07
[Meteor 3] Fix static files on client

henriquealbert:
To test it, you need to run using those flags: `DISABLE_FIBERS=1 IGNORE_ASYNC_PLUGIN=1 METEOR_ENABLE_CLIENT_TOP_LEVEL_AWAIT=true mymeteor` 
Note that `METEOR_ENABLE_CLIENT_TOP_LEVEL_AWAIT` is a new one, to enable TLA on Client

Fixes HMR compare - it was breaking due to not having async await

Also, I noticed that HMR is a bit slow compared to Meteor 2.X


Grubba27 (approved):
Awesome and we will add this todo regarding perf

PR #12520, opened 2023-02-22
Improve specificity of types previously declared as "Object"

ebroder:
This PR is the twin of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64368 (which was opened first against DefinitelyTyped to make sure the tests still passed)

PR #12554, opened 2023-03-09
Two type fixes

ebroder:
Two errors that snuck into my changes in #12520; apologies for the oversights.

### DIFF
--- a/packages/email/email.d.ts
+++ b/packages/email/email.d.ts
@@ -1,5 +1,5 @@
 import { SendMailOptions } from 'nodemailer';
- 
+
 export namespace Email {
   /**
    * ExtraMailOptions is intentionally left empty here, but can be overridden in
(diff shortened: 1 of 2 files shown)


packages/email/email.d.ts
```diff
@@ -11,7 +11,7 @@ export namespace Email {
   interface ExtraMailOptions {}
   type EmailOptions = { mailComposer: MailComposer } | (ExtraMailOptions & SendMailOptions)
 
-  interface CustomEmailOptions extends EmailOptions {
+  type CustomEmailOptions = EmailOptions & {
```
radekmie, line 14: I thought we've decided to leave it as an interface?

ebroder: No, we added the `ExtraMailOptions` interface so that it could be extended. But as currently written, this throws an error:

```
.meteor/local/types/node_modules/package-types/email/package/os/packages/email/email.d.ts:14:40 - error TS2312: An interface can only extend an object type or intersection of object types with statically known members.

14   interface CustomEmailOptions extends EmailOptions {
                                          ~~~~~~~~~~~~
```

In the `DefinitelyTyped` version, we switched this to a type instead of an interface for that reason, but I missed it when updating this declaration from my original patch.

radekmie: Ah, OK, good.

radekmie (approved):

Issue #12318, opened 2022-11-24
Document more properties on Meteor object


zodern:
https://docs.meteor.com/api/core.htm is missing:
- Meteor.isModern
- Meteor.gitCommitHash
- Meteor.isTest
- Meteor.isAppTest
- Meteor.isPackageTest

radekmie:
These should also be reflected in types as well (CC @piotrpospiech).

bakhtiarcr7:
where can I find there description?

harryadel:
@bakhtiarcr7 Seems like most of them are defined [here](https://github.com/meteor/meteor/blob/ffcfa5062cf1bf8a64ea64fef681ffcd99fe7939/packages/meteor/client_environment.js#L61) but this is no ordinary copy/paste job, please lookup all over the codecase to better grasp what each individual config is meant to do. For instance, this [part](https://github.com/meteor/meteor/blob/5c4bdc18d9b7d468c450459bdc6e0c4f7a18d5c9/packages/modern-browsers/modern.js#L70) of the codebase contains more info about isModern. 

Also, don't forget to edit the [types file](https://github.com/meteor/meteor/blob/f60bfc7f75a24263539ef67fe856afab6eaf983e/packages/meteor/meteor.d.ts) like @radekmie suggested. Good luck!

PR #12563, opened 2023-03-15
Add undocument properties to docs

harryadel:
Should resolve https://github.com/meteor/meteor/issues/12318

### DIFF
--- a/packages/meteor/meteor.d.ts
+++ b/packages/meteor/meteor.d.ts
@@ -487,7 +487,7 @@ export namespace Meteor {
   /** True if running in development environment. */
   var isDevelopment: boolean;
   var isModern: boolean;
-  var gitCommitHash: string;
+  var gitCommitHash: string | undefined;
   var isTest: boolean;
   var isAppTest: boolean;
   var isPackageTest: boolean;
--- a/packages/meteor/meteor.d.ts
+++ b/packages/meteor/meteor.d.ts
@@ -486,7 +486,10 @@ export namespace Meteor {
   /** Global props **/
   /** True if running in development environment. */
   var isDevelopment: boolean;
+  var isModern: boolean;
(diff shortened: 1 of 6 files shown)


StorytellerCZ (approved):

packages/meteor/test_environment.js
```diff
@@ -10,8 +12,29 @@ var testDriverPackageName = TEST_METADATA.driverPackage;
 
 // Note that if we are in test-packages mode neither of these will be set,
 // but we will have a test driver package
+/** 
+ *@memberof Meteor
+ * @summary Boolean variable.  True if running in test environment.
```
zodern, line 17: Sometimes this is false in test environments: https://github.com/meteor/meteor/blob/77df795928df55d1e114cb4d2a9143e433a9e773/tools/cli/commands.js#L1995-L1996

packages/meteor/meteor.d.ts
```diff
@@ -486,7 +486,10 @@ export namespace Meteor {
   /** Global props **/
   /** True if running in development environment. */
   var isDevelopment: boolean;
+  var isModern: boolean;
+  var gitCommitHash: string;
```
zodern, line 490: `gitCommitHash` is undefined if the app isn't in a git repository.

packages/meteor/test_environment.js
harryadel: What would be the proper description then? ` True if running in test environment, but can be set to false if test are being run in full-app mode`?

zodern: Maybe something like `True when running tests, except for when running in full-app mode`. Or maybe `True when running unit tests (false if running tests in full app mode)`.

Grubba27:
Thanks again for the efforts @harryadel

PR #12569, opened 2023-03-21
Updating 3.0 with devel

Grubba27:
we need to adjust node version and babel versions for 3.0


meteor
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
 BUNDLE_VERSION=14.21.1.8
```
Grubba27, line 3: Here maybe an issue. Left what was before in 3.0

PR #12572, opened 2023-03-22
Updating release 3.0 with eslint rules

Grubba27:
While working on fibers issues came across this awesome [eslint rule](https://eslint.org/docs/latest/rules/require-await) that may help us, and we were [discussing making](https://github.com/meteor/meteor/pull/12309) eslint better in general for meteor.

In this PR, I update the eslint deps with vazco ones, removed prettier.config( could not make it work with vazco eslint)
and commented a piece of code that is in the devel branch but is breaking in 3.0

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -1240,13 +1240,13 @@ function popCallbackFromArgs(args) {
   }
 }
 
-ASYNC_COLLECTION_METHODS.forEach(methodName => {
-  const methodNameAsync = getAsyncMethodName(methodName);
-  Mongo.Collection.prototype[methodNameAsync] = function(...args) {
-    try {
-      return Promise.resolve(this[methodName](...args));
-    } catch (error) {
-      return Promise.reject(error);
-    }
-  };
-});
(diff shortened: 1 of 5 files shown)


packages/mongo/collection.js
```diff
@@ -1240,13 +1240,13 @@ function popCallbackFromArgs(args) {
...
-  const methodNameAsync = getAsyncMethodName(methodName);
-  Mongo.Collection.prototype[methodNameAsync] = function(...args) {
-    try {
-      return Promise.resolve(this[methodName](...args));
-    } catch (error) {
-      return Promise.reject(error);
-    }
-  };
-});
+// ASYNC_COLLECTION_METHODS.forEach(methodName => {
```
StorytellerCZ, line 1243: There should be at least a comment here if this is not being removed completely, so that if someone stumbles upon this they are aware of why this was done.

StorytellerCZ (approved):

packages/mongo/collection.js
henriquealbert: I agree

henriquealbert (approved):
LGTM

PR #12573, opened 2023-03-23
Release 3.0 observe sync iife

Grubba27:
## This is a draft.

We are looking for options to make `cursor.observe` to be sync. as this affects a few packages and possibly more apps.
One example of package using is [useFind](https://gist.github.com/Grubba27/23cf4be926c4fcea17850b2d866b56e1) from react packages and with this approach things work but it `throws` errors on key duplication, so it may need some adjustments .

I will make another PR with another option that I thought, an observeLocal that is intended to be used in the client and will `throw` and error if used in the server. 

Would be more surface area but it would not affect what @denihs has been doing in the `mongo` and `minimongo` packages so far.

### DIFF
--- a/packages/minimongo/cursor.js
+++ b/packages/minimongo/cursor.js
@@ -227,7 +227,7 @@ export default class Cursor {
    * @param {Object} callbacks Functions to call to deliver the result set as it
    *                           changes
    */
-  async observeChanges(options) {
+  observeChanges(options) {
     const ordered = LocalCollection._observeChangesCallbacksAreOrdered(options);
 
     // there are several places that assume you aren't combining skip/limit with
@@ -333,29 +333,37 @@ export default class Cursor {
       query.movedBefore = wrapCallback(options.movedBefore);
     }
 
-    if (!options._suppress_initial && !this.collection.paused) {
-      const handler = async doc => {
-        const fields = EJSON.clone(doc);
+    const isReadyPromise = (async () => {
+      if (!options._suppress_initial && !this.collection.paused) {
+        const handler = async (doc) => {
+          const fields = EJSON.clone(doc);
 
-        delete fields._id;
+          delete fields._id;
 
-        if (ordered) {
-          await query.addedBefore(doc._id, this._projectionFn(fields), null);
+          if (ordered) {
+            await query.addedBefore(doc._id, this._projectionFn(fields), null);
+          }
+
+          await query.added(doc._id, this._projectionFn(fields));
+        };
+        // it means it's just an array
+        if (query.results.length) {
+          for (const doc of query.results) {
+            await handler(doc);
+          }
         }
-
-        await query.added(doc._id, this._projectionFn(fields));
-      };
-      // it means it's just an array
-      if (query.results.length) {
-        for (const doc of query.results) {
-          await handler(doc);
+        // it means it's an id map
+        if (query.results?.size?.()) {
+          await query.results.forEachAsync(handler);
         }
       }
-      // it means it's an id map
-      if (query.results?.size?.()) {
-        await query.results.forEachAsync(handler);
-      }
-    }
+
+      // run the observe callbacks resulting from the initial contents
+      // before we leave the observe.
+      await this.collection._observeQueue.drain();
+
+      handle.isReady = true; 
+    })();
 
     const handle = Object.assign(new LocalCollection.ObserveHandle, {
       collection: this.collection,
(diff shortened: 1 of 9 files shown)


radekmie:
@Grubba27 I've fixed this issue, and now the app just works (except for the stuff that doesn't for other reasons, e.g., toggling the tasks).

There's a different problem, though - _sometimes_ when toggling visibility, we see only one note. That's because the `observe` doesn't fill the `data` array synchronously. And that's expected, as the `await`s in it effectively delay all of the calculation.

I did an experiment, and made `observeChanges` truly synchronous - everything worked great. Now the question is, whether we can actually do it. It's hard to check the entire codebase, but I think there's no reason to make it actually `async`. The only place we need to wait for something is the `this.collection._observeQueue.drain()` call, and that happens _after_ we triggered all of the initial events.

It'd be best to see if that's true - keeping it sync makes _a lot_ of things much easier. If not, then I'll think about a different approach (I don't like having a separate `observeLocal`, as proposed in #12574).

Grubba27:
I think the same, and will adjust the pr. 

> The only place we need to wait for something is the this.collection._observeQueue.drain() call, and that happens after we triggered all of the initial events.

As you only pass callbacks to observe you should not rely on them running now. It is supposed to have only side effects running there, will check if tests pass if they do I think it is okay. 

will check with @denihs as well

Grubba27:
I think the same, and will adjust the pr. 

> The only place we need to wait for something is the this.collection._observeQueue.drain() call, and that happens after we triggered all of the initial events.

As you only pass callbacks to observe you should not rely on them running now. It is supposed to have only side effects running there, will check if tests pass if they do I think it is okay. 

will check with @denihs as well

Grubba27:
I think the same, and will adjust the pr. 

> The only place we need to wait for something is the this.collection._observeQueue.drain() call, and that happens after we triggered all of the initial events.

As you only pass callbacks to observe you should not rely on them running now. It is supposed to have only side effects running there, will check if tests pass if they do I think it is okay. 

will check with @denihs as well

denihs (approved):
Let's what closely the impact of this change. Looking at it now, it seems that's fine. But how knows

packages/minimongo/cursor.js
```diff
@@ -379,7 +361,14 @@ export default class Cursor {
 
     // run the observe callbacks resulting from the initial contents
     // before we leave the observe.
-    await this.collection._observeQueue.drain();
+    const isReadyPromise = this.collection._observeQueue.drain();
+
+    if (Meteor.isClient) handle.isReady = true;
```
zodern, line 366: These Meteor.isClient checks should probably instead check what type of minimongo collection/cursor it is. Unnamed collections and collections created for DDP clients should work the same way even if they are on the server.

packages/minimongo/cursor.js
```diff
@@ -334,36 +314,38 @@ export default class Cursor {
...
       if (query.results.length) {
         for (const doc of query.results) {
-          await handler(doc);
+          handler(doc);
         }
       }
       // it means it's an id map
       if (query.results?.size?.()) {
-        await query.results.forEachAsync(handler);
+        query.results.forEachAsync(handler);
```
radekmie, line 336: This should use normal `forEach`.

packages/minimongo/cursor.js
```diff
@@ -379,7 +361,17 @@ export default class Cursor {
...
+
+    const isLocalCollection =
+      Meteor.isClient || this.collection.constructor.name === 'LocalCollection'; // Unamed collections
+
+    if (isLocalCollection) handle.isReady = true;
+    else isReadyPromise.then(() => (handle.isReady = true));
+
+    handle.isReadyPromise = Meteor.isClient
+      ? Promise.resolve()
+      : isReadyPromise;
```
radekmie, line 374: The `_observeQueue` is either `Meteor._SynchronousQueue` or `Meteor._AsynchronousQueue`. The latter's `drain` returns `Promise<void>`, but the former's returns `void`. That means, this code could look as follows:

```ts
const drainResult = this.collection._observeQueue.drain();
if (drainResult instanceof Promise) {
  handle.isReadyPromise = drainResult;
  drainResult.then(() => (handle.isReady = true));
} else {
  handle.isReady = true;
  handle.isReadyPromise = Promise.resolve();
}
```

radekmie (changes requested):

radekmie (approved):

PR #12574, opened 2023-03-24
Release 3.0 observeLocal

Grubba27:
As commented in #12573, this is the second PR. Instead of having an observe function that works both in the client and the server, I added one for the client, and the original observe is server and client but always async.
observeLocal only works with the client and could be a replacement for those who use observe in the client and want a sync version of it

tests that I've added are green as well:

Screenshot 2023-03-24 at 10 31 08

### DIFF
--- a/packages/minimongo/cursor.js
+++ b/packages/minimongo/cursor.js
@@ -1,5 +1,5 @@
-import LocalCollection from './local_collection.js';
-import { hasOwn } from './common.js';
+import LocalCollection from "./local_collection.js";
+import { hasOwn } from "./common.js";
 import { ASYNC_CURSOR_METHODS, getAsyncMethodName } from "./constants";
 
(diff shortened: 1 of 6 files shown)


packages/minimongo/cursor.js
```diff
@@ -384,23 +383,190 @@ export default class Cursor {
...
+   * @summary Watch a query.  Receive callbacks as the result set changes.
+   * @locus Anywhere
+   * @memberOf Mongo.Cursor
+   * @instance
+   * @param {Object} callbacks Functions to call to deliver the result set as it
+   *                           changes
+   * @returns
+   */
+  observeLocal(options) {
+    if (Meteor.isServer)
```
zodern, line 396: It should be possible to use observeLocal for unnamed collections on the server, and when using a ddp client on the server.

PR #12580, opened 2023-04-03
[Release 3.0] fixing CI(tools) tests

Grubba27:
This relates to this [Favro card](https://favro.com/organization/929aab32206e9caa0b59b5cf/cef8866673a006c0854ffd64?card=Met-3053)

- [x] Test Group 0 (Solved)
- [x] Test Group 1 (Solved)
- [x] Test Group 2 (solved)
- [x] Test Group 3 
    -  [x] custom minifier - devel vs prod
- [x] Test Group 4 (solved)
- [x] Test Group 5 
- [x] Test Group 6
- [x] Test Group 7 
- [x] Test Group 8
- [x] Test Group 9 (Gabriel) 
  - `matcher.js` error about match not working properly now that is async, tests that rely on it are failing. Check again later on [`matchAsync`](https://github.com/meteor/meteor/blob/cd6b356c19d37d4f2d8434723a13f9d0df7b141d/tools/tool-testing/matcher.js#L73-L79)
  - `read` is `async` and must be awaited
- [x] Test Group 10 
  -  Puppeter issue is blocking this test from passing
- [x] Test Group 11 
  - Is requiring Fibers. not sure how to proceed
- [x] Isolated Tests
    - [x] springboard
    - [x] unknown release
    - [x] source maps from an app


packages/core-runtime/load-js-image.js
```diff
@@ -49,7 +49,7 @@ function load(name, runImage) {
       Package._define(name, exports);
     }
 
-    var pendingCallbacks = pending[name];
+    var pendingCallbacks = pending[name] || [];
```
zodern, line 52: @Grubba27 how can I reproduce the issue that required this change? This probably is a symptom of another bug since pendingCallbacks[name] should always be set on line 12.

packages/test-in-console/run.sh
```diff
@@ -10,6 +10,10 @@ export METEOR_HOME=`pwd`
 
 export phantom=$phantom
 
+# force fibers to be disabled
+export DISABLE_FIBERS=1
+export IGNORE_ASYNC_PLUGIN=1
```
zodern, line 15: These env vars should no longer be needed. If they are, someplace is using an old version of meteor-babel.

packages/core-runtime/load-js-image.js
Grubba27: Not sure, just rebased this branch, maybe @henriquealbert or @denihs may know it better. I've never have touched this file

denihs: Hi @zodern, I did this change on this [commit](https://github.com/meteor/meteor/pull/12580/commits/e4ea3c7d5e9e7e14cf346afb14be47b82f6b3163). 

I was getting "Can't get function forEach of an undefined", [here](https://github.com/meteor/meteor/pull/12580/commits/e4ea3c7d5e9e7e14cf346afb14be47b82f6b3163#diff-5a1fcf6f355a7ce8375019ca9fb89f4644baf574d769749d5e4a0c3f07f7cecbR54) if I'm not mistaking.

zodern: Thanks @denihs. I can reproduce. Could we add fixing this to the roadmap? For some reason this code runs twice when loading the app code in the app used for that test, when it should only run once.

denihs: Hi @zodern,

I was working to fix another test and I end up [here](https://github.com/meteor/meteor/pull/12580/commits/eb63a1b6643805590d3bf70ff9611e837aede1e9#diff-0972e87d5bf49b0ff289a29dee4ae83bb04b83d2e057976f7a7680e0e1ee2e0aL118) . 

For some reason `self.files` sometimes contains the same file more than once.

I created this [workaround](https://github.com/meteor/meteor/pull/12580/commits/eb63a1b6643805590d3bf70ff9611e837aede1e9#diff-0972e87d5bf49b0ff289a29dee4ae83bb04b83d2e057976f7a7680e0e1ee2e0aR170-R177)  to filter duplicated files, and it works. 

I was even able to [remove](https://github.com/meteor/meteor/pull/12580/commits/eb63a1b6643805590d3bf70ff9611e837aede1e9#diff-5a1fcf6f355a7ce8375019ca9fb89f4644baf574d769749d5e4a0c3f07f7cecbR52) the `pending[name] || [];` code.

Now we need to understand why this is happening... I've created this [task](https://favro.com/organization/929aab32206e9caa0b59b5cf/cef8866673a006c0854ffd64?card=Met-3176) so we can keep track of it.

denihs: Actually, I had to bring `pending[name] || [];` back :confused:

If I remove it the test `assets - unicode asset names are allowed` (which is the first one I fixed with this code) passes, but the one I was fixing today (`javascript hot code push`) fails with "Cannot read property 'forEach' of undefined".

fredmaiaarantes: @zodern ?

PR #12587, opened 2023-04-06
Fixes #12516 - Mongo query hangs all clients subscribed to a query/collection

### LINKED ISSUE #12516: Mongo query mismatch hangs client
If the Meteor server publication ever has `undefined` in a query, it will throw an error about a mismatch between Meteor and MongoDB: `Exception in defer callback: Error: The Mongo server and the Meteor query disagree on how many documents match your query. Maybe it is hitting a Mongo edge case?`. I believe this is [expected behavior](https://docs.meteor.com/changelog.html#breakingchanges-51). However, throwing this error means that the client never receives a response for the subscription, which in turn blocks all other requests/responses. This makes it seem like the client hangs forever.

We first noticed this in version 2.7.3 (probably not the earliest occurrence, just a code change on our part), but I have confirmed this is also the case for version 2.10.0.

I've created an example project containing the issue: https://github.com/jdgjsag67251/meteor-sub-bug.

KoenLav:
Throwing an error will break this particular observer for this particular query, causing all clients subscribed to this query (or even collection?) to hang (not receiving results for any of the methods they call).

I'm open to alternative approaches, but throwing an error breaking the entire client connection is way too much.

### DIFF
--- a/packages/mongo/oplog_observe_driver.js
+++ b/packages/mongo/oplog_observe_driver.js
@@ -887,15 +887,11 @@ _.extend(OplogObserveDriver.prototype, {
       // there.
       // XXX if this is slow, remove it later
       if (self._published.size() !== newResults.size()) {
-        console.error('The Mongo server and the Meteor query disagree on how ' +
+        Meteor._debug('The Mongo server and the Meteor query disagree on how ' +
           'many documents match your query. Cursor description: ',
           self._cursorDescription);
-        throw Error(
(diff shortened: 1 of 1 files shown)


radekmie (commented):
It's an edge case anyway, so I think it's safe to do that. However, I think we'd need to add tests for that first.

Grubba27:
If you can add some tests @KoenLav I would be happy to merge your PR 😄 
If you need help implementing it ping me

radekmie (approved):

PR #12589, opened 2023-04-06
Adding loggin to old api

Grubba27:
For testing this, you should use this command inside the meteor repo:

```bash

TINYTEST_FILTER="simple add" WARN_WHEN_USING_OLD_API=true ./meteor test-packages mongo

```

It is in the draft because I still could not make it work properly. For some reason, using 

```js
Mongo.Collection.prototype[methodName] = function (...params) {
// wrapping with the log function
   return this[methodName](...params);
}
```

results in an infinite error while testing.

Open to suggestions to make this work.


radekmie:
Overriding the method [like this](https://github.com/meteor/meteor/pull/12589/files#diff-e145c1d05aed3f6761172f40f6135e46fdf3852a9af912bbab0ba6f771e67d4dR939) won't work, as you've basically said `insert` (I'll stick to this example) calls `insert`. And extracting it before creating the `insertAsync` method is not going to work either, as it'll break collection hooks and other packages.

I'd propose making `warnUsingOldApi` take the method name and the collection instance. Then, if this function unsets the `isCallingMethodAsync`, you can call it multiple times without a problem. Right now, the removed calls in `packages/mongo/mongo_driver.js` look correct, but I'm missing them in `packages/mongo/collection.js`.

What is more, the warning could include the collection's name, which would help a lot, I think.

Grubba27:
I think I got lost/ not sure how to follow, it still logs as it is sync. I'm not sure where is `MongoConnection` async methods are being created :/

PR #11416, opened 2021-05-08
Use latest version of sockjs-client

harryadel:
This seems to be a [long requested](https://github.com/meteor/meteor-feature-requests/issues/216) fix in Meteor, dicussed in https://github.com/meteor/meteor/pull/9215, https://github.com/meteor/meteor/issues/4114, https://github.com/meteor/meteor/pull/9985

With initial go ahead from the Meteor core team which was only held back by the need to further investigation the addition of JSON3 and jump in bundle size.  
https://github.com/meteor/meteor/issues/4114#issuecomment-273472853
>I guess I should clarify that the size increase stated was in development mode, so less alarming as the production bundle would be minified. Still a big jump. I just did a quick test of the (minified) production size and it's only an increase of 10KB on the final JS resource bundle, so a lot less alarming.
>
>I'll still look into it though as it's not really clear at first glance or from the Changelog. The file size difference of the uncompressed distribution in their repo is 282KB between 0.3.4 and 1.1.1 – a huge increase. JSON2 was commented out of the old version on the Meteor side but that only affects the minified production version for the sake of this comparison. The new version comes with JSON3.

>Anyhow, more on this soon.

It's a humble attempt and I'm open to feedback, thanks!

PR #12590, opened 2023-04-10
Update sockjs-client

harryadel:
The original Meteor team made sure to leave notes to point out the changes in the library
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L1 I can't believe I missed reading this during my previous attempt I must read into things carefully next time :man_shrugging:  

There're 7 changes:

1. Dropping in-built support for json2 in favor of Meteor JSON package
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L26

SockJS used to use json2, upgraded it to json3 then entirely removed it from the library in [v1.6](https://github.com/sockjs/sockjs-client/blob/main/Changelog.md#160)

2. [Connecting to an https sockjs endpoint from an http page load fails on Internet Explorer 8 and above with error 1002, "Can't connect to server".](https://github.com/sockjs/sockjs-client/issues/79)
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L235
This was solved in this [PR](https://github.com/sockjs/sockjs-client/pull/155) which was merged in https://github.com/sockjs/sockjs-client/releases/tag/v1.0.0-beta.1

3. Relative URI wasn't supported
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L235
[Fixed](https://github.com/sockjs/sockjs-client/issues/209#issuecomment-61265564) in 1.x

4. Again, changes to account for relative URIs # 3
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L323

5. Related changes to # 2
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L1244

6. Related changes to # 2
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L1249

7. A fix for a bug in Chrome
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L2008
Seems to be fixed in Chrome itself
https://github.com/sockjs/sockjs-client/pull/129#issuecomment-41572101

8. Related changes to # 2
https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/sockjs-0.3.4.js#L2093 

---------------------------------------------

Taking all of this into account, using [v1.6](https://github.com/sockjs/sockjs-client/releases/tag/v1.6.0) should include fixes for json3 changes, http to https cross-scheme & relative URI problems which were the original add-on modification made by the Meteor team.

Which should be as simple as replacing this [line](https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/browser.js#L11), at least theoretically 

https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/socket-stream-client/browser.js#L11

During my [previous attempt](https://github.com/meteor/meteor/pull/11416) I had used v1.5.1 which didn't include the JSON3 changes so maybe that's why it didn't work. :thinking:  

https://github.com/meteor/meteor/pull/11416/files#diff-573f20ed923000513a8e922d805fd8a346d4cad0a408841dade65d724095afafR20


StorytellerCZ:
One test failure that seems unrelated to this. Have triggered a re-run so we'll see if that helps.

harryadel:
The tests were failing at the very same place in the previous attempt 
https://github.com/meteor/meteor/pull/11416#issuecomment-881192772

The test does nothing novel, it just adds a package that includes `ddp` package and invokes a method to count number of calls. You can see for yourself [here](https://github.com/meteor/meteor/tree/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/tools/tests/apps/package-tests/packages/say-something-client-targets).

By including the `ddp` package, it in turn includes `ddp-server` and `ddp-client` which [references](https://github.com/meteor/meteor/blob/d61036dd9d6e86db69a77ed668ce0c88937ad3d9/packages/ddp-client/package.js#L23) socket-client and thus contains our changes. 

I ran things locally and found out that it hangs because of its inability to find `crypto` module
`cannot find module 'crypto'` which is weird given that sockjs-client bundles things [specially ](https://github.com/sockjs/sockjs-client/blob/main/package.json#L15) for browser. I moved the changes from [this file](https://github.com/sockjs/sockjs-client/blob/main/lib/utils/browser-crypto.js) directly where it was needed but it happened again with other packages bundled for browser so I kept modifying but after I thought it was over. I ran into another error `cannot find module 'url'` which is weird since url is found in [eventsource](https://github.com/EventSource/eventsource/blob/79dacfa860751438c116df76f9c5fdb9a2a3d69f/lib/eventsource.js#L1) which is package used by [sockjs-client](https://github.com/sockjs/sockjs-client/blob/main/package.json#L33). Apparently, Meteor either ships the nodejs/server package to the client or it's unable to take note of `browser` property when things are ran by `npm.depends`. Maybe @zodern can shed a light on this. We can workaround this by converting the ddp packages to NPM instead of Meteor packages to circumvent this problem or try to figure out why `npm.depends` behaves this way.

By importing the library from a file copied from the CDN, it worked seamlessly! Which is funny since I never tried this way of importing before
>Here're the different ways we tried importing sockjs-client:
>
>import '.npm/package/node_modules/sockjs-client/dist/sockjs.min.js';
>import { SockJS } from "sockjs-client"
>import SockJS from "sockjs-client";
>import * as SockJS from "sockjs-client";

I guess it was because I was merely stabbing in the dark without any intuition as to why things are supposed to work.

Anyway, I'm super stoked for this PR as this opens up the way for improvements to the DDP protocol and other great features. 
And, it meshes perfectly with the upcoming changes like Connect/Express, Fiber-less Meteor and other changes. Everything is falling into place perfectly.

PR #12592, opened 2023-04-11
Mongo: Add deprecation tags to deprecated methods

alisnic:
This is a follow-up of https://github.com/meteor/meteor/pull/12585/files. Logs are cool, but having type annotations is cooler. This in combination with https://www.npmjs.com/package/eslint-plugin-deprecation will allow us to prepare **and** keep the codebase ready before actually upgrading to Meteor 3.

Please let me know whether I should add these annotations to other types

### DIFF
--- a/packages/mongo/mongo.d.ts
+++ b/packages/mongo/mongo.d.ts
@@ -103,7 +103,7 @@ export namespace Mongo {
       maxDocuments?: number
     ): Promise<void>;
     /**
-     * @deprecated use createdIndexAsync
+     * @deprecated This method will be removed in Meteor 3.0. Please use createdIndexAsync
      */
     createIndex(
       options?: NpmModuleMongodb.CreateIndexesOptions
@@ -147,13 +147,13 @@ export namespace Mongo {
     /**
      * Finds the first document that matches the selector, as ordered by sort and skip options. Returns `undefined` if no matching document is found.
      * @param selector A query describing the documents to find
-     * @deprecated use findOneAsync
+     * @deprecated This method will be removed in Meteor 3.0. Please use findOneAsync
      */
     findOne(selector?: Selector<T> | ObjectID | string): U | undefined;
     /**
      * Finds the first document that matches the selector, as ordered by sort and skip options. Returns `undefined` if no matching document is found.
      * @param selector A query describing the documents to find
-     * @deprecated use findOneAsync
+     * @deprecated This method will be removed in Meteor 3.0. Please use findOneAsync
      */
     findOne<O extends Omit<Options<T>, 'limit'>>(
       selector?: Selector<T> | ObjectID | string,
@@ -178,7 +178,7 @@ export namespace Mongo {
      * Insert a document in the collection.  Returns its unique _id.
      * @param doc The document to insert. May not yet have an _id attribute, in which case Meteor will generate one for you.
      * @param callback If present, called with an error object as the first argument and, if no error, the _id as the second.
-     * @deprecated use insertAsync
+     * @deprecated This method will be removed in Meteor 3.0. Please use insertAsync
      */
     insert(doc: OptionalId<T>, callback?: Function): string;
     /**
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Yes please!

alisnic:
@StorytellerCZ is there list of methods that will be deprecated somewhere?

zodern:
Some of these are only deprecated on the server, and still supported for collections created for a DDP connection (for example, named collections created on the client) or for collections created without a name: `findOne`, `insert`, `remove`, `update`, `upsert`, `count`. I don't think it is possible for the types to distinguish the two types of collections.

alisnic:
@zodern but what will happen when Meteor will migrate off fibers? will this discrepancy still remain? Meaning, some methods will only work on client? That will make the code impossible to be isomorphic

zodern:
Collections in Meteor can store and query the data from two places: in memory, or in a MongoDB server. The sync api doesn't need fibers, and will continue to work and be supported for collections that store the data in memory. The async api works with both, which is what isomorphic code would use. This also isn't a client/server split - you can create collections on the server that store the data in memory, though it is less common.

We've made a lot of progress in supporting the async api's on the client, but there are limitations with them compared to the sync apis and places where the async api's can't easily be used. Supporting the sync api's on the client also significantly reduces the amount of work many apps would have to do to update to Meteor 3.

alisnic:
@zodern does that mean that the goal of this PR is impossible to achieve?

Grubba27:
I would not say that the methods are deprecated; instead, they should be migrated. I would find this one useful if we have another pr for meteor 3.0 removing those tags, as those methods are being deprecated because their behavior will change(they will become async in the server) but the methods will not be removed

radekmie:
> [...] (they will become async in the server) [...]

I still think it's not a good idea to have the same function return different types depending on the environment. I think these should be removed, or throw an error when used.

alisnic:
I am confused as well, one of the best selling points of Meteor is isomorphic code with almost free optimistic UI. if the same methods behave differently on client in server, that goes away. I'm not sure what is the plan here, but so far it leaves me a bit dissapointed

radekmie:
@alisnic The `*Async` methods are isomorphic and handle optimistic UI.

alisnic:
@radekmie then why keep non-async methods if they are not isomorphic? For easier fetching in react components useTracker? 🤔

Grubba27:
> Subsequently, what happens if call a sync method on the server after fibers are removed? will I get an error while TypeScript will tell that everything is ok?

At some point, we thought that in 3.0, and I still believe that this is the desire of the team that today's `sync` methods in Meteor 3.0 will behave as their `*Async` counterpart.

In 3.0 the isomorphism will be kept in both methods.

What we could have is a method, something like `insertLocal` or `findLocal` in LocalCollection(it does not matter if it is a server or client) for sync data fetching but I'm not that really happy with increasing the API surface area.

radekmie:
> At some point, we thought that in 3.0, and I still believe that this is the desire of the team that today's `sync` methods in Meteor 3.0 will behave as their `*Async` counterpart.

So they'll return a `Promise` on the server and direct values on the client?

Grubba27:
> So they'll return a Promise on the server and direct values on the client?

No, I mean is like:

```js
let user, userA;
let id = "some id";
// Meteor 3.0
user = await Collection.findOne(id) // returns a Promise<user>
userA = await Collection.findOneAsync(id) // returns a Promise<user>

users === usersA
```

radekmie:
OK, now it's clear, but I think it should be highlighted in the release notes _really_ hard. Also, remember that these `user === userA` won't ever be true, as these operations return fresh objects.

radekmie:
> > OK, now it's clear, but I think it should be highlighted in the release notes really hard
>
> You mean in the 3.0 release notes, right?

Both 3.0 and all pre-releases.

Grubba27:
> how we should proceed forward with this PR? From the discussion it's not 100% clear to whether the changes relevant. Is there anything else I can do here to improve DX for deprecated methods?
 
Those methods *Will not* be removed(as stated in the code). Their behavior will change to become equal to their `*Async` counterpart(will remain isomorphic), and both the client and the server will behave the same. The focus of changing the classic API to the async one is to make sure that _your app_ environment is async and _will_ handle a fibers-free meteor. 


Now I realize that I wrote: "(they will become async in the server) " I guess that is what created the whole confusion :/ 

For the PR, I'm not super happy with this coming because, as I said, those methods will not be deprecated/removed. In 3.0 their behavior will change in favor of their `*Async` counter-part in Meteor3

radekmie:
https://github.com/meteor/meteor/discussions/12644 should be linked here.

PR #12596, opened 2023-04-14
Fix syntax for legacy client

zodern:
Fixes syntax for packages that are not compiled with ecmascript so they only use es5.

Adds a check-syntax script that parses the js files in these packages to ensure they use valid syntax for es5. It only checks the syntax, so it doesn't replace the browser stack testing Meteor does before every release. The goal is for it to catch the most common issues before merging a pull request.

The script also ensures that packages that could be used in build plugins use valid syntax for the Meteor versions they could be used in. In the past this was Meteor 1.4+, but I'm not sure if that has been tested recently. Currently the script makes sure it at least uses syntax that is supported in Meteor 1.6+ (Node 8).


scripts/admin/check-legacy-syntax/check-syntax.js
Grubba27, line 1: This is nice. Thanks a lot @zodern

PR #12611, opened 2023-05-03
Execute collection methods in separate fibers

yrambler2001:
With Promise.asyncApply, the code will not be blocked by fetching collection data, see screenshots

Currently:
Every fetch call is blocked by the previous one. Promise.all does nothing:
Screenshot 2023-05-02 at 19 12 11

With the fix:
Every fetch call is executed in the background and awaited in Promise.all:
Screenshot 2023-05-02 at 19 33 53

### DIFF
--- a/packages/minimongo/cursor.js
+++ b/packages/minimongo/cursor.js
@@ -522,7 +522,7 @@ ASYNC_CURSOR_METHODS.forEach(method => {
   Cursor.prototype[asyncName] = function(...args) {
     try {
       this[method].isCalledFromAsync = true;
-      return Promise.resolve(this[method].apply(this, args));
+      return Promise.asyncApply(async () => this[method].apply(this, args));
     } catch (error) {
       return Promise.reject(error);
     }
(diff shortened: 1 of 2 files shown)


packages/minimongo/cursor.js
```diff
@@ -522,7 +522,7 @@ ASYNC_CURSOR_METHODS.forEach(method => {
   Cursor.prototype[asyncName] = function(...args) {
     try {
       this[method].isCalledFromAsync = true;
-      return Promise.resolve(this[method].apply(this, args));
+      return Promise.asyncApply(async () => this[method].apply(this, args));
```
denihs, line 525: The problem I see with this solution is that it relies on Fibers (`asyncApply`), and we're removing it. 

I did some tests with our branch `release-3.0`, and it works as expected:

image

I think that we're getting this result right now because the current release (2.12) still relies on Fibers.

Let's proceed with your changes. I'll just ask you to add a TODO above the return, so we remember to keep the Promise.resolve for the release 3.0. It can be something like:

```
TODO [fibers]: revert this change to "Promise.resolve(this[method].apply(this, args));" in version 3.0
```

denihs (changes requested):

yrambler2001:
Thanks for the review @denihs. It is good that 3.0 is actively developing and fibers will be removed in 3.0.

italojs:
fibers isnt supported by meteor 3.x anymore

PR #12612, opened 2023-05-03
Updating 3.0 with current(2.12)

Grubba27:
We are updating 3.0 with devel


packages/meteor/dynamics_nodejs.js
```diff
@@ -2,6 +2,18 @@
 let nextSlot = 0;
 let callAsyncMethodRunning = false;
 
+var Fiber = Npm.require('fibers');
```
Grubba27, line 5: This should be remove

packages/meteor/dynamics_nodejs.js
```diff
@@ -99,6 +111,30 @@ class EnvironmentVariableAsync {
  */
 Meteor.EnvironmentVariable = EnvironmentVariableAsync;
 
+EVp._set = function (context) {
```
Grubba27, line 114: This as well should be removed

packages/meteor/helpers.js
```diff
@@ -147,6 +154,25 @@ Meteor.wrapAsync = function (fn, context) {
...
+  return function() {
+    var ret = fn.apply(this, arguments);
+    if (ret && typeof ret.then === 'function') {
+      return Promise.await(ret);
+    }
+
+    return ret;
+  }
+};
+
```
Grubba27, line 175: Turn this into 
```js

Meteor.wrapFn = function (fn) {return fn;}

File: packages/callback-hook/hook.js [hunk 4]
```diff
@@ -49,6 +49,11 @@ export class Hook {
       this.bindEnvironment = false;
     }
 
+    this.wrapAsync = true;
+    if (options.wrapAsync === false) {
+      this.wrapAsync = false;
```
denihs, line 54: Why not just

```
this.wrapAsync = options.wrapAsync !== false
```

packages/hot-module-replacement/client.js
```diff
@@ -1,28 +1,28 @@
 // TODO: add an api to Reify to update cached exports for a module
-const ReifyEntry = require('/node_modules/meteor/modules/node_modules/@meteorjs/reify/lib/runtime/entry.js')
+var ReifyEntry = require('/node_modules/meteor/modules/node_modules/@meteorjs/reify/lib/runtime/entry.js')
```
denihs, line 2: Why everything is this file is now `var` instead of `const` and `let`?

packages/meteor/dynamics_nodejs.js
denihs, line 5: Why is this code back?

This import will always break because the fibers module is not installed anymore.

tools/isobuild/bundler.js
```diff
@@ -2227,7 +2227,20 @@ class JsImage {
...
-            return getAsset(item.assets, assetPath, "utf8", callback);
+            const result = getAsset(item.assets, assetPath, "utf8", callback);
+
+            if (!callback) {
+              if (!Fiber.current) {
+                throw new Error("The synchronous Assets API can " +
+                    "only be called from within a Fiber.");
+              }
+
+              return Promise.await(result);
```
matheusccastroo, line 2238: @Grubba27 this will need to be removed I think

packages/meteor/dynamics_nodejs.js
denihs, line 114: The same goes for this. Why is this back?

tools/isobuild/bundler.js
```diff
@@ -2238,7 +2251,20 @@ class JsImage {
...
-            return getAsset(item.assets, assetPath, undefined, callback);
+            const result = getAsset(item.assets, assetPath, undefined, callback);
+
+            if (!callback) {
+              if (!Fiber.current) {
+                throw new Error("The synchronous Assets API can " +
+                    "only be called from within a Fiber.");
+              }
+
+              return Promise.await(result);
```
matheusccastroo, line 2262: this too

tools/static-assets/server/boot.js
```diff
@@ -354,10 +355,22 @@ const loadServerBundles = Profile("Load server bundles", async function () {
 
     const Assets = {
       getText: function (assetPath, callback) {
-        return getAsset(assetPath, "utf8", callback);
+        const result = getAsset(assetPath, "utf8", callback);
+        if (!callback) {
+          return Future.fromPromise(result).wait();
```
matheusccastroo, line 360: this too

denihs (changes requested):

packages/hot-module-replacement/client.js
Grubba27: if you use const or any ES6 it breaks the IE11, we have a check for it now

denihs: Oh, I see... lol

denihs (commented):
Now it looks good. Once the CI is green I think we can merge it.

Grubba27:
`test: ddp-heartbeat` keeps failing... Not sure why. Maybe is because of the latest version of ddp that was released in 2.12?

denihs:
We're going to continue this work in #12616

PR #12613, opened 2023-05-03
Solving legacy for 3.0

Grubba27:
Trying making the CI pass for legacy js.
Note that I check if the file is sent to the client as well.


Grubba27:
Closed in favor of #12616 as it has all the adjustments

PR #12616, opened 2023-05-04
Update release-3.0 with devel

denihs:
Updating release-3.0 with devel and solving conflicts.

We need to make sure the CI is green before merging this.


scripts/admin/check-legacy-syntax/check-syntax.js
```diff
@@ -0,0 +1,215 @@
...
+      if (
+        error.message.includes("Unexpected token") &&
+        line.includes("await")
+      ) {
+        console.log(
+          "Failed to parse, but it's probably because you're using TLA."
+        );
+        console.log("Will continue");
+        continue;
+      }
```
zodern, line 173: If a package is able to use TLA (has a dependency on ecmascript), it can be removed from the list of packages this script runs on. This script is only needed for packages that can't use TLA/ecmascript.

PR #12618, opened 2023-05-05
Add top level await docs

zodern:


docs/source/api/top-level-await.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Using in packages
+
+Top level await is only supported started in Meteor 3.  Build plugins are able to use top level await in older Meteor versions since the runtime is bundled when they are published.
+
+If you want to ensure your package only runs in versions of Meteor that support top level await, you can have your package use `isobuild:top-level-await`:
+
+```js
+Package.onUse(function (api) {
+  // Do not allow this package to be used in pre-Meteor 3 apps.
```
StorytellerCZ, line 106: Wouldn't `api.versionsFrom('3.0') have the same effect?

StorytellerCZ (approved):

docs/source/api/top-level-await.md
zodern: Not always. api.versionsFrom is just an easier way to provide the versions for any core packages that the package uses. If the package doesn't use any core packages, or the core packages it does use have the same version as the previous Meteor release, then it would allow the package to run in older Meteor versions.

docs/source/api/top-level-await.md
```diff
@@ -0,0 +1,120 @@
...
+if (Meteor.isClient) {
+  import './init-client.js';
+}
+
+export function showNotification(message) {
+  import show from './notifications.js';
+
+  show(message);
+}
+```
```
Grubba27, line 94: At compile, could we turn these nested imports into required imports?

Or is it already?

```js
if (Meteor.isClient) {
 require('./init-client.js');
}
export function showNotification(message) {
  const show = require('./notifications.js');
  show(message);
}
```
And if we could even, by a flag, use them like this, we could use something like SWC, I think.

The problem is that if the user relies on an async module, that is where things would go wrong. But it already throws an error.

Grubba27:
@zodern, you commented(I could not find it on Slack) that the order in which the packages are loaded has changed. This should be documented as well, I think. Especially for those package maintainers to know that it may change how their packages are loaded into the Meteor project, there is any case that you can think of that now makes a difference?

docs/source/api/top-level-await.md
zodern: That is similar to what reify does, with some differences to handle live bindings.

I think we will want to move away from nested imports. The rest of the javascript ecosystem is able to work without them, and it is becoming increasingly common for new developer tools to not support them.

We could look into doing something like that with SWC, but it might not be much faster than transpiling once with just babel. What's nice about Meteor is every package and app can independently choose what build plugins to use. If some packages or apps don't use nested imports, then they could replace `ecmascript` with a build plugin that uses `swc`, while rest of the packages used by the app continue using `ecmascript`.

zodern:
@Grubba27 I updated the section on execution order to cover how packages are loaded. There are cases where it could make a difference, but I'm not sure whether those differences matter. We had decided to start out with loading packages in parallel as much as we can to start, and based on feedback from real apps see if we want to reduce it.

docs/source/api/top-level-await.md
Grubba27: Yeah, that was most of my thoughts. Many new packages, and for most apps, probably will be better with using a plugin like SWC, as they might not even know that nested import is a thing.
That does sound good. Will think more about this matter, that may be a thing to be tackled when meteor 3.x is released.

Grubba27:
The docs CI is still failing. Do you know why? cc @zodern 
I saw that you removed the `strip-tla.js` file, but I think there is something else being done there that helped the build

Grubba27:
Now is okay!
here is a link https://deploy-preview-12618--meteor-docs.netlify.app/api/top-level-await.html

Grubba27 (approved):

henriquealbert (approved):
lgtm

PR #12620, opened 2023-05-06
Typescript 4.9.5

StorytellerCZ:
A bump to Typescript 4.9.5

### DIFF
--- a/npm-packages/meteor-babel/package.json
+++ b/npm-packages/meteor-babel/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@meteorjs/babel",
   "author": "Meteor <dev@meteor.com>",
-  "version": "7.18.0-beta.6",
+  "version": "7.18.0-beta.7",
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


npm-packages/meteor-babel/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@meteorjs/babel",
   "author": "Meteor <dev@meteor.com>",
-  "version": "7.18.0-beta.6",
+  "version": "7.18.0-beta.7",
```
StorytellerCZ, line 4: I think this should be published out of beta before release.

Grubba27:
I was planning on leaving these bumps for v3.1 as I think it is safe to say that TypeScript does not follow SemVer, we could do a whole bump to 5.0 now or later( I prefer later, if something breaks I can look into with more time )

StorytellerCZ:
@Grubba27 changed the target of this PR to the dev-bundle, for your convenience, but NPM packages release will also be needed.

PR #12621, opened 2023-05-06
Typescript 5

StorytellerCZ:
Upgrade to Typescript 5
Release article: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/

Also updated dependencies in `meteor-babel`.


Grubba27:
TS v5 broke many things(they do not follow sem-ver), so I was postponing this matter to when we released Meteor 3.x beta. I think that is when I think I will have some more time to work on this matter

StorytellerCZ:
I expected as much. I will get this up to date with v3 for now so that we have a starting point ready when the time comes to look more into this.

perbergland:
The only thing that broke with Typescript 4 was that the meteor internal package (which DOES follow semver rules for dependency checks) was assigned the same version as typescript itself. This broke all meteor packages that depend on the Meteor Typescript package since meteor doesn’t have any support for overriding the package resolution algorithms.

The refapp:meteor-typescript package already uses Typescript 5 without any issues since it uses a separate versioning scheme. So if you just refrain from bumping the major version number on the typescript meteor package and instead use e.g. 4.10.zzz etc there will not be any problems bringing in ts5

PR #12622, opened 2023-05-06
Updating bundle dependencies

StorytellerCZ:
Related to my Typescript updates I would like to consider updating dependencies in the bundles. Here are some which I think are super easy to update. Starting with a patch version of NPM.


npm-packages/eslint-plugin-meteor/scripts/build-dev-bundle-common.sh
StorytellerCZ, line 1: Found this file. Seems like copy of `build-dev-bundle-common.sh` from scripts. Is this really needed?

StorytellerCZ: Same for `dev-bundle-tool-package.js`

Grubba27: those files inside of eslint I think is not needed updating, also I think it should be deleted but I forgot which PR was removing them

StorytellerCZ: For now I'll keep it. If there is a PR that removes them, then only better.

PR #12632, opened 2023-05-10
Allow `onCreateUser` to receive async callbacks

KevLehman:
Currently, `onCreateUser` only supports a sync function. This was not a problem while using fibers, but now with promises (and the parent function being async too) we need to have the ability to pass async callbacks to user creation hook.

This PR enables that by awaiting the place where `onCreateUserHook` is executed before inserting into the database


Grubba27:
Nice catch @KevLehman thanks for your submition!

Grubba27 (approved):

denihs (approved):

PR #12635, opened 2023-05-15
Update mongo.d.ts with projection

StorytellerCZ:
Fields have been renamed to projection upstream and while in Meteor we can use both, it was not accounted for in the TS definitions.

### DIFF
--- a/packages/mongo/mongo.d.ts
+++ b/packages/mongo/mongo.d.ts
@@ -30,8 +30,12 @@ export namespace Mongo {
     skip?: number | undefined;
     /** Maximum number of results to return */
     limit?: number | undefined;
-    /** Dictionary of fields to return or exclude. */
+    /**
+     * Dictionary of fields to return or exclude.
+     * @deprecated use projection instead
+     */
     fields?: FieldSpecifier | undefined;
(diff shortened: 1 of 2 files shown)


packages/mongo/mongo.d.ts
```diff
@@ -32,6 +32,7 @@ export namespace Mongo {
     limit?: number | undefined;
     /** Dictionary of fields to return or exclude. */
     fields?: FieldSpecifier | undefined;
+    projection?: FieldSpecifier | undefined;
```
radekmie, line 35: It'd be great to:
1. Add JSDoc to `projection` just like for the `fields`.
2. Add `@deprecated` to `fields` and point it to `projection`.

radekmie (approved):

packages/mongo/mongo.d.ts
StorytellerCZ: If it is just here then it is done.

StorytellerCZ:
Also found two uses of `fields` in oplog tailing, which I upgraded to use `projection`. I have it in the latest commit so that it can be reverted easily if needed.

PR #12638, opened 2023-05-16
[Meteor 3]: Removing unnecessary async Profile.time

Grubba27:
Relates to this public [card](https://github.com/orgs/meteor/projects/10/views/1?pane=issue&itemId=27872403)

- [x] Remove all non required uses of Profile.time

- [x] Check if all cases where is no longer needed async await have been adjusted upstream

The wrong uses were removed. Now is needed to make these changes go upstream.


denihs (approved):
LGTM

PR #12653, opened 2023-05-25
Replace deprecated reference with current one

StorytellerCZ:
Saw use of deprecated `Meteor.Collection` highlighted in my IDE, so I replaced it with the current one as indicated on the backward compatibility few lines above. Since this was deprecated in `v0.9.1` we should consider removing it. From the quick search I did `Meteor.Collection` is after this no longer used anywhere else in the code. Same for `Mongo.Collection.ObjectID` and `Mongo.Collection.Cursor`.


Grubba27:
Will merge and solve from release branch(not sure yet what is going on)

Issue #12642, opened 2023-05-19
mongo@1.16.6 continuous exceptions while polling query


azambranogalbis:
Issue refers to mongo v1.16.6 package.
We identified that when both mongodb database and our application are started simultaneously (i.e. upon server restart), and connection to mongo is established when the mongodb server is still performing some boot actions.

Application is configured without MONGO_OPLOG env variable (i.e. polling strategy is used to retrieve data from subscriptions).
Relevant code from application is as follows:
```
col.find(query, { fields: _fields }).observe({
      added(p) {
        self.add(p._id, p);
        if (callback) callback('added', p);
      },
      changed(p) {
        self.mod(p._id, p);
        if (callback) callback('changed', p);
      },
      removed(p) {
        self.del(p._id);
        if (callback) callback('removed', p);
      },
    });
```

*col* being a Mongo.Collection instance.

We observe that the following log is displayed every 10 seconds

```
2023-05-19T05:21:08.249447113Z Exception while polling query {"collectionName":"_transformations","selector":{},"options":{"transform":null,"projection":{"_id":1,"template":1,"functions":1,"code":1}}} MongoTopologyClosedError: Topology is closed
2023-05-19T05:21:08.249530945Z     at processWaitQueue (/usr/app/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/sdam/topology.js:550:46)
2023-05-19T05:21:08.249542167Z     at Topology.selectServer (/usr/app/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/sdam/topology.js:297:9)
2023-05-19T05:21:08.249549679Z     at Topology.<anonymous> (/usr/app/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/sdam/topology.js:43:94)
2023-05-19T05:21:08.249558513Z     at internal/util.js:308:30
2023-05-19T05:21:08.249566571Z     at new Promise (<anonymous>)
2023-05-19T05:21:08.249574764Z     at Topology.selectServerAsync (internal/util.js:307:12)
2023-05-19T05:21:08.249583543Z     at executeOperationAsync (/usr/app/programs/server/npm/node_modules/meteor/npm-mongo/node_modules/mongodb/lib/operations/execute_operation.js:94:35)
[...]

Issue #12628, opened 2023-05-09
To many connections to db


nesbtesh:
Description:
I'm encountering an issue where my application is generating too many connections to the database, causing my application to crash. I'm running Meteor version 2.12 on heroku. The error only happened when we upgrade to meteor 12

Expected Behavior:
The application should be able to handle a reasonable number of database connections without crashing.

Actual Behavior:
The application generates too many connections to the database, which causes the application to crash. This issue occurs randomly and is not related to any specific action performed in the application.

Steps to Reproduce:

Start the application and load a few pages.
Monitor the number of database connections being created.
Continue to load pages until the maximum number of connections is reached.
The application crashes due to too many connections being generated.

Note:
I have checked the server logs and there are no errors reported. I have also checked the MongoDB logs and there are no indications of any issues on the database side.

dovydaskukalis:
Same. In our case we can't even start our sandbox server because we are using Cloud Atlas free tier which has a limit of 500 connections. Seems like on startup it goes beyond that limit, hanging the app.

paulincai:
Ok, you clearly have a wrong MONGO_URL configuration.
Please try these options and let me (us) know:

```
const options = '?retryWrites=true&maxIdleTimeMS=5000&maxPoolSize=30&readConcernLevel=majority&readPreference=secondaryPreferred&w=majority&heartbeatFrequencyMS=15000'

MONGO_URL: `mongodb+srv://your_mongo_url/meteor${options}`,

'meteor' is the default DB but it may be called something different if you wanted it to be different.

```

paulincai:
@nesbtesh ok but I didn't mean to suggest to take those values ad litteram. The point was to make sure you have options set on your link.  I am checking my configs now as I might be seeing the same problem on one of the Meteor apps to which I did not deploy new code for a while (restart Node)

paulincai:
ok, check this documentation. In particular the way connections are being created and killed. In general it is ok to have a large number of connections opened and idle ready for transactions. There is no need to closed them, you just need to have an efficient total maximum number of them in the pool.
I didn't know this: "If the application uses a 
[read preference](https://www.mongodb.com/docs/manual/core/read-preference/#std-label-read-preference)
 to query the secondary nodes, those connection pools grow and there can be 306 total connections."

https://www.mongodb.com/docs/drivers/node/current/faq/#std-label-node-faq-connection-pool

yrambler2001:
[...]

PR #12659, opened 2023-05-29
Fix multiple connections (revert pr #12593)

yrambler2001:
fixes issues #12642 #12628
reverts #12593


packages/meteor/dynamics_nodejs.js
```diff
@@ -107,15 +106,6 @@ EVp._setNewContextAndGetCurrent = function (value) {
...
 
-EVp._isCallAsyncMethodRunning = function () {
-	return callAsyncMethodRunning;
-};
-
-EVp._setCallAsyncMethodRunning = function (value) {
-	callAsyncMethodRunning = value;
-};
-
-
```
Grubba27, line 118: this should not be removed

PR #12662, opened 2023-06-01
fix(passwordless): align function with MongoDB convention

vitorsouzaalmeida:
Hello :)

With this PR I propose to solve the [Accounts Passwordless login doesn't work with id and token](https://github.com/meteor/meteor/issues/12401) issue, at the `accounts-passwordless` package.

The problem is already documented on the issue, So I'll explain directly my solution.

This fix checks the `id` field in the `selector` object. If found, it will replace `id` with `_id`  before passing to the `findOne`.

cc @filipenevola

### DIFF
--- a/packages/accounts-passwordless/passwordless_server.js
+++ b/packages/accounts-passwordless/passwordless_server.js
@@ -18,8 +18,8 @@ const findUserWithOptions = ({ selector }) => {
   return Meteor.users.findOne(
     { 
       ...rest,
-      ...(id && {_id: id}), 
-      ...(email && {'emails.address': email}) 
+      ...(id && { _id: id }), 
+      ...(email && { 'emails.address': email }) 
     },
     {
       fields: {


packages/accounts-passwordless/passwordless_server.js
```diff
@@ -12,9 +12,19 @@ const findUserWithOptions = ({ selector }) => {
   if (!selector) {
     Accounts._handleError('A selector is necessary');
   }
-  const { email, ...rest } = selector;
+  const { email, id, ...rest } = selector;
+
+  let modifiedSelector = { ...rest };
```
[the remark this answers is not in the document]
vit0rr: You're right. I'll fix it.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -12,9 +12,19 @@ const findUserWithOptions = ({ selector }) => {
   if (!selector) {
     Accounts._handleError('A selector is necessary');
   }
-  const { email, ...rest } = selector;
+  const { email, id, ...rest } = selector;
+
+  const modifiedSelector = { ...rest };
+  if (id) {
+    modifiedSelector._id = id;
```
filipenevola, line 19: I would avoid mutations to improve readability I would do something like:
```
{ ...(selector.id && { _id: selector.id }) }
```

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -12,9 +12,19 @@ const findUserWithOptions = ({ selector }) => {
...
+  } 
+
+   if (email) {
+    modifiedSelector['emails.address'] = email;
```
filipenevola, line 23: same here

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -12,9 +12,19 @@ const findUserWithOptions = ({ selector }) => {
...
+  }
+
   return Meteor.users.findOne(
-    { ...rest, ...(email ? { 'emails.address': selector.email } : {}) },
+    modifiedSelector,
```
filipenevola, line 27: then create the selector directly here.

filipenevola (changes requested):

packages/accounts-passwordless/passwordless_server.js
vit0rr: Nice, If I understood, something like that?
```js
const findUserWithOptions = ({ selector }) => {
  if (!selector) {
    Accounts._handleError('A selector is necessary');
  }
  const { email, id, ...rest } = selector;

  return Meteor.users.findOne(
    { 
      ...rest,
      ...(id && {_id: id}), 
      ...(email && {'emails.address': email}) 
    },
    {
      fields: {
        services: 1,
        emails: 1,
      },
    }
  );
};
```

With it, it is the same logic, but without mutations, I do not need to create a new `const`.

filipenevola (approved):

Grubba27 (approved):
LGTM!

DanielHe4rt:
LGTM!

henriquealbert (approved):

PR #12669, opened 2023-06-05
Update logging package

StorytellerCZ:
* Add types to logging package


Grubba27 (approved):
LGTM! would be nice to come in a 2.14

Grubba27:
For some reason, tests are failing, have any idea why?

StorytellerCZ:
Could me upping the version manually have an impact? Or maybe it doesn't like something about the types?

StorytellerCZ:
@Grubba27 can you try making a new bundle. Looks like somewhere there is issue with ansi-style which is part of chalk. I have upped it in the dev bundle to have the same version as the package.

StorytellerCZ:
I'll do a separate PR with just the types.

Grubba27:
We will need a new bundle for 2.14, I will probably make sure that we have everything okay so that I only need to generate one bundle(the CI is annoying to work with)

StorytellerCZ:
Dev bundle changes moved to #12832

StorytellerCZ:
Chalk v5 might have been to much for Meteor 2.x, will do that for 3.x. I think now the tests will pass. If not then I have one more commit to throw at this that should fix it.

Grubba27:
Since is just a deps update I would send this to a RC, what do you think @StorytellerCZ ?

PR #12670, opened 2023-06-05
Fix #12391

### LINKED ISSUE #12391: Accounts.onLogin callback is called on client even when beforeExternalLogin returns false
Meteor version: 2.5.6
OS: macOS Ventura 13.1

I'm trying to add Google sign in to my app, and for my use case I need to only allow certain known users based on their email address. I'm using `Accounts.beforeExternalLogin` to do this on the server, by checking our DB for the email of the user attempting to log in, and returning false if they do not exist. For the sake of reproducing this issue though, we can just pretend that I want to fail the sign in no matter what:
```js
Accounts.beforeExternalLogin((type, data, user) => {
  return false;
});
```

On the client side, I'm using the `Accounts.onLogin` and `Accounts.onLoginFailure` hooks to check whether the Google sign in succeeded or failed. 

```js
useEffect(() => {
  const { stop: stopOnLogin } = Accounts.onLogin((details) => {
    console.log("Accounts.onLogin");
    // redirect to home page
  });

  const { stop: stopOnLoginFailure } = Accounts.onLoginFailure((error) => {
    console.log("Accounts.onLoginFailure");
    // show an error message
  });

  return () => {
    stopOnLogin();
    stopOnLoginFailure();
  }
}, []);
```

And for what it's worth, here's my code for initiating the OAuth flow:

```js
Meteor.loginWithGoogle({
  prompt: "select_account"
});
```

Expected behavior:
When returning false from `beforeExternalLogin`, after completing the Google OAuth flow, the `onLoginFailure` callback will be called and the `onLogin` callback will not be called.

Actual behavior:
On the client, the `onLogin` callback is called (with `{ type: "resume" }` in the `details` parameter) and the `onLoginFailure` callback is not. On the server, the `onLoginFailure` callback is called, as expected.

I don't think I'll be able to create a reproduction for this issue since `loginWithGoogle` needs some configuration in Google Cloud Console to work at all, afaik. I have looked in the source code though and have a hunch of where the issue is coming from. The stack trace shows the `onLogin` callback is coming from `_startupCallback` in [accounts_client](https://github.com/meteor/meteor/blob/4e53835141a620236d837ed95fd754747747a4b1/packages/accounts-base/accounts_client.js#L457). The code is checking for `_loginCallbacksCalled` to be true and assuming that means the user is logged in. But `_loginCallbacksCalled` is getting set to true [here in loginCallbacks](https://github.com/meteor/meteor/blob/4e53835141a620236d837ed95fd754747747a4b1/packages/accounts-base/accounts_client.js#L229), before the error check. It seems like the solution might be to just move `this._loginCallbacksCalled = true;` into the `if (!error)` block. I would submit a PR myself, but unfortunately I don't have time right now to build the source, write tests, etc.

brucejo75:
This is a fix for #12391.
This feature was introduced by https://github.com/meteor/meteor/pull/10543 which fixed https://github.com/meteor/meteor/issues/10157.

### DIFF
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -226,13 +226,14 @@ export class AccountsClient extends AccountsCommon {
     const loginCallbacks = ({ error, loginDetails }) => {
       if (!called) {
         called = true;
-        this._loginCallbacksCalled = true;
         if (!error) {
           this._onLoginHook.forEach(callback => {
             callback(loginDetails);
             return true;
           });
+          this._loginCallbacksCalled = true;
         } else {
+          this._loginCallbacksCalled = false;
           this._onLoginFailureHook.forEach(callback => {
             callback({ error });
             return true;


vit0rr:
CI was broken in `devel`. Can you merge from `release-2.13` into your branch?

PR #12671, opened 2023-06-06
fix: Failing tests of deprecated packages

rodrigok:
`./meteor test-packages` is failing due to some error on package `markdown`, I could not find the issue and I realized the deprecated packages were not being loaded from locally, but from atmosphere, so using `METEOR_PACKAGE_DIRS="packages/deprecated" ./meteor test-packages` I could include the deprecated packages from local file system and then I found the errors fixed in this PR. The solution for the markdown was the removal of the `lazy: true`.

### DIFF
--- a/packages/deprecated/http/httpcall_client.js
+++ b/packages/deprecated/http/httpcall_client.js
@@ -66,6 +66,8 @@ HTTP.call = function(method, url, options, callback) {
       throw new Error('Option auth should be of the form "username:password"');
     username = options.auth.substring(0, colonLoc);
     password = options.auth.substring(colonLoc+1);
+
+    headers["Authorization"] = "Basic " + btoa(username + ":" + password);
   }
 
   if (params_for_body) {
@@ -110,7 +112,7 @@ HTTP.call = function(method, url, options, callback) {
     else
       throw new Error("Can't create XMLHttpRequest"); // ???
 
-    xhr.open(method, url, true, username, password);
+    xhr.open(method, url, true);
 
     for (var k in headers)
       xhr.setRequestHeader(k, headers[k]);
--- a/packages/deprecated/http/httpcall_server.js
+++ b/packages/deprecated/http/httpcall_server.js
@@ -161,7 +161,6 @@ function _call (method, url, options, callback) {
     .catch(err => callback(err));
 }
 
-// we are keeping wrapAsync here as this package is deprecated
 HTTP.call = function(...args) {
   const cb = args.pop();
   if (typeof cb === 'function') {
--- a/packages/deprecated/http/httpcall_server.js
+++ b/packages/deprecated/http/httpcall_server.js
@@ -1,4 +1,4 @@
-import { Meteor } from 'meteor/meteor';
+import Util from 'util';
 import { fetch, Request } from 'meteor/fetch';
 import { URL, URLSearchParams } from 'meteor/url';
 import { HTTP, makeErrorByStatus, populateData } from './httpcall_common.js';
@@ -162,4 +162,10 @@ function _call (method, url, options, callback) {
 }
 
 // we are keeping wrapAsync here as this package is deprecated
-HTTP.call = Meteor.wrapAsync(_call);
+HTTP.call = function(...args) {
+  const cb = args.pop();
+  if (typeof cb === 'function') {
+    return _call(...args, cb);
+  }
+  return Util.promisify(_call)(...args, cb);
+}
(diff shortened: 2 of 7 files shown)


packages/deprecated/http/httpcall_server.js
```diff
@@ -162,4 +162,10 @@ function _call (method, url, options, callback) {
 }
 
 // we are keeping wrapAsync here as this package is deprecated
```
zodern, line 164: This comment is out of date.

Also, I wanted to note that the change to HTTP.call is a breaking change, so we will need to bump the major version for the package.

packages/deprecated/markdown/package.js
```diff
@@ -11,7 +11,7 @@ Package.onUse(function (api) {
   api.versionsFrom('3.0-alpha.6');
   api.use('ecmascript@1.0.0-alpha300.6');
   api.use("templating@2.0.0-alpha300.6", "client", {weak: true});
-  api.mainModule('template-integration.js', 'client', { lazy: true });
```
zodern, line 14: It would be nice if we could leave this module as lazy since it allows the package to be dynamically imported. Do you have any ideas on why it didn't work? Do you know if the tests passed with Meteor 2.x?

packages/deprecated/stylus/plugin/compile-stylus.js
```diff
@@ -1,8 +1,7 @@
+const util = require('util');
```
zodern, line 1: This import isn't used

packages/deprecated/stylus/plugin/compile-stylus.js
```diff
@@ -155,7 +154,7 @@ class StylusCompiler extends MultiFileCachingCompiler {
 
     const fileOptions = inputFile.getFileOptions();
 
-    const f = new Future;
+    // const f = new Future;
```
zodern, line 157: This comment could be removed

packages/deprecated/stylus/plugin/compile-stylus.js
```diff
@@ -168,10 +167,16 @@ class StylusCompiler extends MultiFileCachingCompiler {
                  .set('cache', false)
                  .set('importer', importer);
 
-    style.render(f.resolver());
+    const prom = new Promise((resolve, reject) => {
+      style.render((err, data) => {
+        if (err) return reject(err);
+        resolve(data);
+      })
```
zodern, line 174:
```suggestion
      });
```

packages/deprecated/markdown/package.js
rodrigok: No ideia, maybe related to bundler? Could you try to check this @zodern ?

packages/deprecated/http/httpcall_client.js
```diff
@@ -110,7 +112,7 @@ HTTP.call = function(method, url, options, callback) {
     else
       throw new Error("Can't create XMLHttpRequest"); // ???
 
-    xhr.open(method, url, true, username, password);
+    xhr.open(method, url, true);
```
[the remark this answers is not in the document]
rodrigok: Good question, it was not passing the info to the server, so I had to change it to pass the tests. I was using latest chrome to test.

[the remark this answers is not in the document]
rodrigok: It seems to have landed on chrome at 2021, after the package was moved to deprecated and, probably, not tested anymore.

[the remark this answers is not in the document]
zodern: That's interesting.

The only issue with the new code would be if the password or username has any unicode characters since btoa throws for them: https://developer.mozilla.org/en-US/docs/Web/API/btoa#unicode_strings. I don't know how important that is. Besides that issue, the new implementation looks good to me.

packages/deprecated/markdown/package.js
zodern: I'll look into it. There were a few changes that could affect this.

denihs: @zodern in version 2.x works properly. 

I was trying to understand what's happening and it looks like what's being generated by the linker for the `markdown` package is:

```
Package["core-runtime"].queue("markdown"), [..deps], function () {
   return () {}
}
```

That's as far as I got.

rodrigok: Maybe we could merge this and open an issue related to meteor 3 to investigate this further

zodern: I would be fine with that.

rodrigok: @zodern could you approve the PR then?

zodern (approved):

PR #12674, opened 2023-06-13
Update source-map-support

zodern:
Updates it to the latest commit in https://github.com/meteor/node-source-map-support. This fixes handling errors that happen in the first line of a linked file. Usually there are no errors on the first line, but we've encountered it a few times in Meteor 3 when there was a bug in the linker.


Grubba27:
I will merge this to 2.x and then send those changes to 3.0

Issue #12676, opened 2023-06-14
Meteor are creating the "Do the Tutorial" link with wrong URL


vit0rr:
If you try to create an app using `meteor create --typescript`, it will generate the LinksCollection with an old URL.
It is simple to solve. I just need to change every `main` file at Meteor that refer to `https://www.meteor.com/tutorials/react/creating-an-app`, and change it to `https://react-tutorial.meteor.com/simple-todos/01-creating-app.html`

Follow:
 - [skel-apollo](https://github.com/meteor/meteor/blob/49f95f4a092c4753090bdebe27f0c04b39aff896/tools/static-assets/skel-apollo/server/main.js)
 - [skel-chakra](https://github.com/meteor/meteor/blob/49f95f4a092c4753090bdebe27f0c04b39aff896/tools/static-assets/skel-chakra-ui/server/main.js)
 - [skel-react](https://github.com/meteor/meteor/blob/49f95f4a092c4753090bdebe27f0c04b39aff896/tools/static-assets/skel-react/server/main.js)
 - [skel-tailwind](https://github.com/meteor/meteor/blob/49f95f4a092c4753090bdebe27f0c04b39aff896/tools/static-assets/skel-tailwind/server/main.js)
 - [skel-typescript](https://github.com/meteor/meteor/blob/49f95f4a092c4753090bdebe27f0c04b39aff896/tools/static-assets/skel-typescript/server/main.ts)

I'll send a pull-request to solve this issue.

kunalshokeen051:
is the issue still persist?

vit0rr:
> is the issue still persist?

Thanks for the interest, but I solved it on https://github.com/meteor/meteor/pull/12679

namesarnav:
Hey, could you assign this issue to me if it's still open. I'd love to contribute. Thanks!

StorytellerCZ:
@namesarnav there already is a PR to fix this issue.

PR #12679, opened 2023-06-14
Fix skell wrong react url

vit0rr:
Fix https://github.com/meteor/meteor/issues/12676

Just replacing old URLs with the newest.
Follow the issue for more details.

Ps: I'm replacing it with `simple-todos/01-creating-app` because the old URL was sent to `react/creating-an-app`.

### DIFF
--- a/docs/source/roadmap.md
+++ b/docs/source/roadmap.md
@@ -5,7 +5,9 @@ description: Describes the high-level features and actions for the Meteor projec
 
 ## Introduction
 
-Last update: Setember 14, 2022.
+**Quick update moving items to Finished: June 1, 2023**
+
+**Last new items added: September 14, 2022.**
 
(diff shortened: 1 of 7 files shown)


StorytellerCZ:
@vit0rr I don't think these and other closures were intentional.

vit0rr:
> @vit0rr I don't think these and other closures were intentional.

Oh, thanks.
Yep, it does not intentional... I'll fork it again and re-open this and the other PR as soon as I get

StorytellerCZ:
@vit0rr I don't see this in the latest devel. Should I re-create this PR?

vit0rr:
> @vit0rr I don't see this in the latest devel. Should I re-create this PR?

Hello,
This PR is already re-created here: https://github.com/meteor/meteor/pull/12713

Issue #12689, opened 2023-06-17
Remove RegEx from the guide.meteor


vit0rr:
[RegEx](https://github.com/longshotlabs/simpl-schema#regex) is no longer supported by simpl-schema.

>Can also be a function that returns a regular expression or an array of them
Any regular expression that must be matched for the key to be valid, or an array of regular expressions that will be tested in order.

>In earlier releases, the SimpleSchema.RegEx object defined standard regular expressions that you could use as the value for the regEx key. However, many of these were prone to DDoS security risks, causing this package to be flagged by security audit tooling. We no longer include any built-in regular expressions. We encourage you to use a custom function or some other method of validating any strings with complex requirements. If necessary, you can look back at the code for previous SimpleSchema releases and copy specific built-in regular expressions into your codebase.

>Also, check out [recheck](https://makenowjust-labs.github.io/recheck/docs/intro/), which includes an ESLint plugin you can add to your project to avoid accidentally writing regular expressions that could be attacked.

So, update these files:
- [methods.md](https://github.com/vit0rr/meteor/blob/dc38e4325dcd88fb3c6d1be1639680c6ff6f5e80/guide/source/methods.md#L278-L301)
- [collections.md](https://github.com/vit0rr/meteor/blob/dc38e4325dcd88fb3c6d1be1639680c6ff6f5e80/guide/source/collections.md#L91-L99)

I'll send a pull request removing the no supported RegEx expressions

StorytellerCZ:
This is correct, however the version that is used in collection2 only supports v1 which has the functionality described. In the community we are looking into forking Simple Schema into Meteor specific branch. The latest version of Simple Schema doesn't work with Meteor so this is a premature move.

PR #12690, opened 2023-06-17
Remove built in regex validations

vit0rr:
closes https://github.com/meteor/meteor/issues/12689

This pull request updates the no longer supported regEx built-in expressions from SimpleSchema.
The purpose here is to define a regular expression and link the SimpleSchema doc. that explains why it no longer supports built-in expressions.

### DIFF
--- a/docs/source/roadmap.md
+++ b/docs/source/roadmap.md
@@ -5,7 +5,9 @@ description: Describes the high-level features and actions for the Meteor projec
 
 ## Introduction
 
-Last update: Setember 14, 2022.
+**Quick update moving items to Finished: June 1, 2023**
+
+**Last new items added: September 14, 2022.**
 
 This document describes the high-level features and actions for the Meteor project in the near-to-medium term future.
 The description of many items include sentences and ideas from Meteor community members.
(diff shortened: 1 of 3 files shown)


StorytellerCZ (changes requested):
See my comment in the issue. Latest version of Simple Schema does not work in Meteor, so you have to stick with v1 where this functionality is still present. Until we have a community solution to this I would propose that instead guide gives a recommendation to use custom regEx and list regEx for Meteor related functionality like ids.

vit0rr:
> See my comment in the issue. Latest version of Simple Schema does not work in Meteor, so you have to stick with v1 where this functionality is still present. Until we have a community solution to this I would propose that instead guide gives a recommendation to use custom regEx and list regEx for Meteor related functionality like ids.

Well pointed @StorytellerCZ.
So, If I understood, a better approach is to keep the `SimpleSchema.RegEx.Id`, but add some line explaining that we encourage to create of custom regEx for fields like email and amount, for security reasons, and link the official doc. of [simpl-schema](https://github.com/longshotlabs/simpl-schema#regex)?

StorytellerCZ:
Yes, point out that `simpl-schema` only works with Meteor on the 1.x branch, but that they should switch to their own regexes regardless.

StorytellerCZ (approved):

Grubba27 (approved):
LGTM! This will be on 2.14 I think

StorytellerCZ:
Since it is in docs, can't we get it into 2.13?

Grubba27:
> Since it is in docs, can't we get it into 2.13?

Yeah! I read and thought it was code, this can go today to devel

PR #12691, opened 2023-06-20
Release 3.0 [fix] - TODO - cssResource.data is a promise

denihs:
This PR addresses the problem pointed out in #12538.

`cssResource.data` is now a promise, so it wasn't considered a buffer [here](https://github.com/meteor/meteor/blob/b3c6550173d23174dd2b3e0ebf8df9c73dc50dcc/tools/isobuild/compiler-plugin.js#L788).

In order to not make `addStylesheet` async, I'm [pushing a function](https://github.com/meteor/meteor/blob/b3c6550173d23174dd2b3e0ebf8df9c73dc50dcc/tools/isobuild/compiler-plugin.js#L786-L795) to `outputResources` so it can be solved [later on](https://github.com/meteor/meteor/blob/b3c6550173d23174dd2b3e0ebf8df9c73dc50dcc/tools/isobuild/bundler.js#L1229-L1232).

This seems to fix the problem.


zodern:
How does this handle errors? If my understanding of the commented out code is correct, that was one of the main purposes of it.

Currently error handling for css files is at least partially broken - https://github.com/meteor/meteor/pull/12675#issuecomment-1592354248, I'm guessing due to https://github.com/meteor/meteor/pull/12538.

denihs:
Now I'm a little confused, I'm not sure I understand your question.

It should handle the same way it's doing on devel (2.x.x). It's basically the [same code](https://github.com/meteor/meteor/blob/aece11a87fa11507dca62e1ec768330017a5e65c/tools/isobuild/compiler-plugin.js#L783-L791).

Is there some specific case we should look into? Can we reproduce it?

zodern:
If you run the test `compiler-plugins.js: test:compiler plugin caching - less`, it throws at https://github.com/meteor/meteor/blob/a67cf288a31e3462963e797b8e4462dd95703d45/tools/isobuild/compiler-plugin.js#L974-L98

It shouldn't be reaching that line, though I'm not familiar enough with this code to know what the cause is. It is related to how Meteor handles errors reported by build plugins - the error is from:
https://github.com/meteor/meteor/blob/dc38e4325dcd88fb3c6d1be1639680c6ff6f5e80/packages/non-core/less/plugin/compile-less.js#L100-L105

Currently Meteor 3 is incorrectly catching and ignoring the error, so it might be easier to try this on the branch from https://github.com/meteor/meteor/pull/12675.

zodern:
As I said in #12675, I don't fully understand the commented out code, but if I remember correctly, it would cause the error I mentioned in my previous comment to be thrown while the build plugin is running, whereas now it throws at some later point in the build process. Ideally it shouldn't be throwing at all, but I'm not sure how this worked in Meteor 2.

denihs:
I see what you're saying now. 

When running the test `compiler plugin caching - less` on your branch I get the error:

Screenshot from 2023-06-20 13-53-57

My changes from this PR don't affect anything there...

But to me, what I'm doing here is not related to that. I'm addressing the things on #12538.

So should we move on with this one? Or are they related in some way I don't see?

zodern:
I think it is related to https://github.com/meteor/meteor/pull/12538. since this code indirectly deals with error handling. We could merge this, but we would still need to come back and figure out how we want error handling to work here and fix it.

henriquealbert (dismissed):

tools/isobuild/compiler-plugin.js
```diff
@@ -783,18 +783,16 @@ class ResourceSlot {
...
-      //   // If there was an error processing this file, cssResource.data
-      //   // will not be a Buffer, and accessing cssResource.data here
-      //   // should cause the error to be reported via inputFile.error.
-      //   return;
-      // }
-
-      this.outputResources.push(cssResource);
+      this.outputResources.push(async () => {
+        if (! cssResource.lazy &&
+          ! Buffer.isBuffer(await cssResource.data)) {
```
zodern, line 788: To add some context to why this code was commented out, it being async is only part of the issue. The main problem is the code doesn't seem to work correctly - it either does nothing, or if a build plugin reported an error it will crash Meteor. I think we should leave this code disabled until we can understand what it is trying to do and figure out a better way to achieve it. It would also be worth checking if this worked correctly in Meteor 2, and why it worked there.

zodern (changes requested):

PR #12692, opened 2023-06-20
[WIP] Faster and simpler linker for Meteor 3

zodern:
This is a partial rewrite of the linker with these goals:

1. Simplify it. With TLA, the linker increased to having at least 5 different ways to link apps and packages, some of which were incompatible with TLA. This rewrite tries to reduce it to two: one way for apps and packages that use modules, and another way for those that don't.
2. Make it easier to understand so developers new to Isobuild's code can get a high level understanding easier, and make changes with more confidence. Some of this will probably be done in a separate PR to keep the diff smaller.
3. Make it fast, hopefully fast enough to remove many of the caches used in the linker. Each cache has an arbitrary size, and large apps can easily exceed the sizes which basically disables the caching and is a large performance cliff.
4. Always output code that supports top level await

This PR is still in progress. I spent some time the last couple weeks rewriting the linker from scratch to get a better understanding of it, and am now in the process of merging the good parts from that rewrite with the old linker code.

The old linker is still the default. To try the new linker, you can use the `feature/faster-simpler-linker` branch, and set the `METEOR_LINKER` env var to `new`.

### Sourcemaps

Sometimes packages or apps would not have source maps, especially if they didn't use a build plugin. The linker now generates source maps for these files. You can see the difference in these stack traces from an error thrown by the `meteor` package:
```
old linker:
meteor.js?hash=c409f5bd94650302c52417770bb8c173946c6d83:45 Error: error from meteor package
    at meteor.js?hash=c409f5bd94650302c52417770bb8c173946c6d83:45:15
    at meteor.js?hash=c409f5bd94650302c52417770bb8c173946c6d83:126:4
    at load (core-runtime.js?hash=e9797454b50ee00bed582d670c752999d4c84e41:149:16)
    at Object.queue (core-runtime.js?hash=e9797454b50ee00bed582d670c752999d4c84e41:144:5)
    at meteor.js?hash=c409f5bd94650302c52417770bb8c173946c6d83:11:25

new linker:
client_environment.js:4 Error: error from meteor package
    at client_environment.js:4:1
    at client_environment.js:81:1
    at load (load-js-image.js:41:1)
    at Object.queue (load-js-image.js:36:1)
    at meteor.js?hash=8a9e70569c9eccb0178ad88b35397819f175df6b:1:25
```

There are still some improvements needed, such as using a longer path for each file, and trying to not create source maps for files that have different content than on disk.

## Benchmarks

I excluded the time spent on `File#computeAssignedVariables` while linking since this PR doesn't address its performance (I plan to improve it later). Usually `File#computeAssignedVariables` took 250-300ms during the initial build, and none during rebuilds.

### Very small app that does not use modules

**Initial Build**
before - server: 54ms
after -server: 16ms (3.3x improvement)
before - client: 160ms
after - client: 46ms (3.4x improvement)

### Very small app that does use modules

before - server: 55ms
after - server: 16ms (3.4x improvement)
before - client: 155ms
after - client: 45ms (3.4x improvement)

**First rebuild**
before - client: 14ms
after - client: 3ms (4.6x improvement)

### Initial build, without cache, and dynamic imports
For this test, the app dynamically imported 100 copies of jquery-ui, which in total was over 12mb of js.

before - client: 31,871ms
after (unoptimized) - client: 2,317ms ( 13.7x improvement)
after (optimized) - client: 289ms (110.2x improvement)

The optimized time is from after https://github.com/meteor/meteor/commit/67df1a8e5df9c07dcebb7cac511c2c32d7c652bc.

There are a few more things we can do to make it faster.

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -41,6 +41,8 @@ var packageDot = function (name) {
 const enableClientTLA = process.env.METEOR_ENABLE_CLIENT_TOP_LEVEL_AWAIT === 'true';
 const USE_OLD_LINKER = process.env.METEOR_LINKER !== 'new';
 
+const MIN_BANNER_WIDTH = 80;
+
 ///////////////////////////////////////////////////////////////////////////////
 // Module
 ///////////////////////////////////////////////////////////////////////////////
@@ -60,7 +62,7 @@ Object.assign(Module.prototype, {
   // files or directories, and the values are either nested objects
   // (representing directories) or File objects (representing modules).
   // Bare files and lazy files that are never imported are ignored.
-  _buildModuleTrees(dynamicFiles, sourceWidth) {
+  _buildModuleTrees(dynamicFiles) {
     // Map from meteorInstallOptions objects to trees of File objects for
     // all non-dynamic modules.
     const trees = new Map();
@@ -145,11 +147,10 @@ Object.assign(Module.prototype, {
   // Take the tree generated in getPrelinkedFiles and populate the chunks
   // array with strings and SourceNode objects that can be combined into a
   // single SourceNode object. Return the count of modules in the tree.
-  _chunkifyModuleTrees(trees, combinedFile, sourceWidth) {
+  _chunkifyModuleTrees(trees, combinedFile) {
     const self = this;
 
-    // assert.ok(_.isArray(chunks));
-    assert.ok(_.isNumber(sourceWidth));
+    assert.ok(combinedFile instanceof CombinedFile);
 
     let moduleCount = 0;
 
@@ -516,6 +517,7 @@ Object.assign(File.prototype, {
       bannerLines.push('This file is in bare mode and is not in its own closure.');
     }
 
+    // TODO: add divider line to footer
     header += banner(bannerLines) + '\n';
 
     if (code) {
@@ -709,17 +711,12 @@ const prelinkWithModules = Profile('linker prelinkWithModules', (files, name, is
   let dynamicFiles = [];
   let eagerModulePaths = [];
   let mainModulePath = null;
-
-  if (name === null && bundleArch.startsWith('os.')) {
-    debugger;
-  }
  
   let module = new Module(files, isApp);
   // TODO: do we need to calculate a hash?
-  // TODO: remove sourceWidth option
   // TODO: do not use internal methods on module
-  const trees = module._buildModuleTrees(dynamicFiles, 80);
-  const fileCount = module._chunkifyModuleTrees(trees, mainBundle, 80);
+  const trees = module._buildModuleTrees(dynamicFiles);
+  const fileCount = module._chunkifyModuleTrees(trees, mainBundle);
 
   for (const file of files) {
     if (file.bare) {
@@ -888,6 +885,7 @@ class CombinedFile {
 var banner = function (lines, bannerWidth) {
   if (!bannerWidth) {
     bannerWidth = 6 + _.max(lines, function (x) { return x.length; }).length;
+    bannerWidth = Math.max(bannerWidth, MIN_BANNER_WIDTH);
   }
 
   var divider = dividerLine(bannerWidth);
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -96,7 +96,7 @@ Object.assign(Module.prototype, {
 
   // Figure out which vars need to be specifically put in the module
   // scope.
-  computeAssignedVariables: Profile("linker Module#computeAssignedVariables", async function () {
+  async computeAssignedVariables () {
     var self = this;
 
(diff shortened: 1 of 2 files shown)


tools/isobuild/linker.js
```diff
@@ -37,242 +39,30 @@ var packageDot = function (name) {
...
-      }
-
-      // TODO[fibers]: This is a temporary hack to make sure that we don't return the same
-        // file twice. I'm not sure why, but self.files sometimes contains the same file twice.
-        // these tool tests fail without this hack:
-        // - assets - unicode asset names are allowed
-        // - javascript hot code push
-        // and probably others
-      const uniqueHashes = [...new Set(ret.map((f) => f.hash))];
-      return uniqueHashes.map((h) => ret.find((f) => f.hash === h));
```
zodern, line 177: The bug that caused this was introduced when the for ... of loop above was added when removing fibers - it didn't handle caching properly and would add duplicate files when there was a cached entry.

This code was also incompatible with TLA since it generated multiple files for the app. It did a simplified form of linking for apps that did not use modules, which is rare. This PR removes it in favor of linking them in the same way as packages that do not use modules.

zodern:
When linking the thousands of files added from `import * from '@mui/icons-material'`, it is only 2x times faster (504ms vs 1022ms).
However, this includes the new linker generating a sourcemap for all of the files in node_modules, whereas Meteor never did before because it would have been way too slow.

```
old linker:
modules.js?hash=13fc5c6eed8e703f6ec68b7e5f8d2d1d8e746510:3042 Uncaught TypeError: Cannot read properties of null (reading 'useEffect')
    at Module.useEffect (modules.js?hash=13fc5c6eed8e703f6ec68b7e5f8d2d1d8e746510:3042:21)
    at large-imports.js:9:5

new linker:
react.development.js:1634 Uncaught TypeError: Cannot read properties of null (reading 'useEffect')
    at Module.useEffect (react.development.js:1634:1)
    at large-imports.js:10:5
```

zodern:
Most of the performance improvement was from replacing the Mozilla source map library with the much faster `@parcel/source-map` library. However, I've found a few situations where it is slower than I want, especially for generating empty source maps.

A few years ago I was experimenting with creating a new source map library that tried a different approach to concatenating source maps. I'm working on finishing it so we can use it in Meteor.

The slowest part of source map libraries is decoding the mappings, storing them, and then encoding them. One interesting thing about the mappings is they are stored as values relative to the previous mapping. My library leverages that fact to avoid half the work source map libraries would normally do.

That allows the new library to be 16x faster at generating empty source maps, and another optimization raises that to 5,300x faster. The new linker generates a large number of empty source maps for files that Meteor previously didn't have source maps for, and this removes almost all downsides to the linker for doing that extra work (there might be later parts of the build process that it would still affect the performance of).

In other cases, where it is concatenating existing source maps, it is usually 4 - 5x faster. It replaces one or two mappings per input source map to ensure the relative positions are correct, and copies over the other mappings as is. It still has to decode all of the mappings to calculate the position of the last mapping. The slowest test I've used is linking 100 copies of jquery-ui, and this reduces the time spent creating the final source map from 2,300ms with @parcel/source-map to less than 500ms.

The only downside I've found so far is the final source map might be larger since it doesn't deduplicate the names array, but it also seems the mappings are sometimes smaller.

> When linking the thousands of files added from import * from '@mui/icons-material', it is only 2x times faster (504ms vs 1022ms).

With the new source map library and optimizations I made to generating the header and footer of each file, this is ~220ms faster.

zodern:
The new source map library is finished (https://www.npmjs.com/package/@zodern/source-maps). It isn't as fast after fixing all of the bugs in the mappings parser, but it is still at least 3.5x faster when used in the linker.

dallman2:
Hey @zodern, how close is this to being incorporated into the Meteor 3 beta? I have been pulling the thread on why the build times for my (quite large and cumbersome) Meteor app was so slow to rebuild during development. This PR would be a godsend for the ongoing process of migration to 3.0.

StorytellerCZ:
@nachocodoner just run across this. I think this might be useful to your recent efforts.

nachocodoner:
Interesting, this relates more to the Meteor bundler itself. [With Meteor-Rspack integration](https://forums.meteor.com/t/meteor-rspack-integration-a-modern-bundler-meets-meteor-3-4/63696), most of the work to compile and link the app code will be handled by Rspack. Meteor will still connect the intermediate modern app build with Atmosphere packages, so improvements to the Meteor linker will benefit both Rspack and pure Meteor bundler setups. I’ve already seen solid gains from the modern bundler integration, and this PR changes should add a bit more gain I believe. For the core t eam, the main focus remains on the modern workflow, hitting the speed targets, and supporting modern-bundler features.

I’ll save this PR for later analysis. If the implementation is ever ready, we’ll be happy to review and include it.

PR #12696, opened 2023-06-26
Reduce parser plugins used in js-analyze in Meteor 3

zodern:
Meteor had been enabling every parser plugin, which makes parsing 3 - 4x slower compared to not enabling any plugins. js-analyze deals with files that have already been compiled, so it only needs to be able to parse code that Node and browsers are able to run and doesn't need as many plugins.

In a smaller app, this reduces the time spent parsing in `ImportScanner#findImportedModuleIdentifiers` from 18 - 23 seconds to less than 6 during the initial build. For some apps, that part of the initial build takes over a minute or two, so this should be a large improvement for those.

This change also makes finding globals in the linker ~2x faster. 

This optimization could be easily backported to Meteor 2.

### DIFF
--- a/tools/isobuild/js-analyze.js
+++ b/tools/isobuild/js-analyze.js
@@ -1,7 +1,7 @@
 import { parse } from '@meteorjs/babel';
 import { analyze as analyzeScope } from 'escope';
 import LRU from "lru-cache";
-
+import { Profile } from '../tool-env/profile';
 import Visitor from "@meteorjs/reify/lib/visitor.js";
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Would be nice to have a PR to backport it into Meteor 2.

Grubba27 (approved):

Grubba27:
> Would be nice to have a PR to backport it into Meteor 2.

Agreed, will add this to my pipeline!

PR #12697, opened 2023-06-28
[Meteor 3 - Feature]: Meteor create uses git instead of vendorized

Grubba27:
## Under devolvement and discussion.

There are a few open topics I'm open to suggestions.

--- 

# Meteor create changes

Today, when we use `meteor create <app>` we use a static asset that is within the tool itself. In my proposal, I want to change how this works, there are a few reasons:

- Ease of maintenance, if something is wrong with a starter it can be changed easily(now we will use git);
- Possibility of adding new examples and being open to more options because we will be using the network; 

## Meteor create <app-name> --example "exampleapp"

Would instead of just list and show how to clone, it would clone it and make it ready for your development.

## Meteor create <app-name> --from "https://github.com/meteor/skel-react"

Creates from a based URL, clones from that repo, and then initiates the project with the current meteor.

## Cons?

- It will require the end user to have git installed to do a git clone. I guess it falls into our philosophy of being easy to use but I'm not sure if this makes sense now;
- I propose a breaking change along with an update to our tutorials, I would suggest removing the option for creating a `prototype` app(create with the `insecure` packages) it would continue to exist and people can opt-in. I think we(Core Meteor members)  should be more opinative on a few things and this is one of them.

## Todo
- Missing repos:

- [x] create  skel-apollo
- [x] create  skel-full
- [x] create   skel-solid   
- [x] create   skel-vue
- [x] create  skel-bare   
- [x] create   skel-minimal 
- [x] create   skel-svelte
- [x] create   skel-vue-2
- [x] create   skel-blaze   
- [x] create   skel-tailwind 
- [x] create    skel-chakra-ui  
- [x] create skel-react    
- [x] create  skel-typescript

- [x] Create CDN repository ([live here](https://cdn.meteor.com/static/meteor.json), will have it in a git) 

- [x] Update docs(new option and ask the user to have git installed)

## Open topics
 - How to get a repo(other ways than `git clone` and `rm .git`?)
 - Remove insecure or have an easy way to add to a newly created project?
 - Relying on a CDN is nice but maybe we could keep older versions in the tool in case the developer is offline?

### DIFF
--- a/docs/generators/changelog/versions/3.0.md
+++ b/docs/generators/changelog/versions/3.0.md
@@ -356,7 +356,7 @@
 
 - `meteor-tool@3.0.0`:
 
-  - Package was bumped due to a dependency update. No code changes were made.
+  - Changes to how meteor apps are being created [PR](https://github.com/meteor/meteor/pull/12697)
 
 - `meteor@2.0.0`:
 
--- a/docs/source/commandline.md
+++ b/docs/source/commandline.md
@@ -87,19 +87,19 @@ Create a new Meteor project. By default, it uses [React](https://guide.meteor.co
 and makes a subdirectory named *name* and copies in the template app.
 You can pass an absolute or relative path.
 
+_This command requires an Internet connection and having git installed._
+
+<h3 id="meteorcreate-templates">Cloning from project</h3>
+
+You can also create a new Meteor project by cloning an existing project. You can 
+list them using the command `meteor create --list` it will list all the available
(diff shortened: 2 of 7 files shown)


StorytellerCZ:
As much interesting as this is, I don't think this should be worked at least till after 3.0 is released.

Grubba27:
Probably will be launched together or a little bit later on, and IMO is an easy win for us in the team being able to change and update the skeletons and even add more without having to make a new whole release

henriquealbert (approved):

PR #12698, opened 2023-07-02
Fix: correct spelling of EACCES in rename

Torgen:
On windows and WSL, renaming a file can fail because of EPERM or EACCES, so there's a special version of the rename function that retries those errors up to a time limit, then copies the file and deletes the old one instead. However, the decision to retry is based on the correct name of the code (EACCES), and the decision to copy and delete is based on a misspelling (EACCES*S*). This fixes the typo.


StorytellerCZ (approved):

Grubba27 (approved):
LGTM! This will be on 2.14 I think

PR #12709, opened 2023-07-12
HMR server performance improvements

wreiske:
This patch includes several performance improvements for the Hot Module Replacement (HMR) server. Here are the key changes:

1. **Buffer.from(secret)**: The secret is now converted to a Buffer once and stored in `this.secretBuffer` instead of being converted every time it's compared. This reduces unnecessary conversions.

2. **Object to Map**: Objects that were used as hash maps (`this.connByArch`, `this.changeSetsByArch`, `this.cacheKeys`, `this.trimmedArchUntil`) have been replaced with JavaScript Map objects. Maps perform better in scenarios involving frequent additions and removals of key-value pairs.

3. **Set for connections**: Instead of using an array to store connections for each architecture, a Set is used (`this.connByArch.set(arch, new Set())`). This allows for faster removals when a connection is closed because Set's delete operation is faster than Array's splice.

4. **JSON.stringify optimization**: The message sent to all connections is stringified once and reused, instead of being stringified for each connection.

5. **Promise.allSettled in iterWithFn**: Instead of awaiting each promise sequentially, all promises are awaited at once using Promise.allSettled, which can lead to significant performance improvements when dealing with many promises.

6. **Error handling in iterWithFn**: Errors from individual promises are now handled gracefully - they're logged to the console but don't stop execution or affect other promises.


henriquealbert:
Great job, @wreiske! Thank you for your contribution. Have you measured how much the performance has improved?

wreiske:
> Great job, @wreiske! Thank you for your contribution. Have you measured how much the performance has improved?

I have not yet. I am looking through https://github.com/meteor/meteor/blob/devel/DEVELOPMENT.md to see if I can find the proper tests to run. I might create some debug on hot-module-replacement and blaze-hot to see if I can get some performance.now() timings and automate modifying files to get the time differences. I had some time on my lunch break to look at this PR today, it may be some time before I can dive deeper.

Also, sorry! My force push (squashed some commits) removed the HMR project label.

Grubba27 (approved):
LGTM!
The failing test are docs that I need to get back to it and adjust

Grubba27:
Thanks for your contribuiton @wreiske !!

PR #12710, opened 2023-07-16
Small Ferature: Add a way to see if any method is currently invoked

ziedmahdi:
Some times apps load the data using methods instead of publications. This patch will allow the developer to easily know if the client is waiting on any method.

### DIFF
--- a/packages/ddp-client/common/namespace.js
+++ b/packages/ddp-client/common/namespace.js
@@ -89,3 +89,9 @@ DDP.onReconnect = callback => DDP._reconnectHook.register(callback);
 DDP._allSubscriptionsReady = () => allConnections.every(
   conn => Object.values(conn._subscriptions).every(sub => sub.ready)
 );
+
+// Sometimes data is loaded using a method, this hack provide a way
+// to see if any method is currently invoked
+DDP._noMethodIsCurrentlytInvoked = () => allConnections.every(
+    conn => Object.keys(conn._methodInvokers).length == 0
+);
--- a/packages/ddp-client/package.js
+++ b/packages/ddp-client/package.js
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Meteor's latency-compensated distributed data client",
-  version: '2.6.1',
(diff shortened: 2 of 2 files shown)


StorytellerCZ (changes requested):
We will have to discuss the naming. Also since this is something for the apps to use it shouldn't be hidden behind underscore and instead be a new, properly documented feature.

ziedmahdi:
@StorytellerCZ What name do you propose?
I used the underscore as I followed `DDP._allSubscriptionsReady`. I guess same thing apply to `DDP._allSubscriptionsReady`, it should be renamed to `DDP.allSubscriptionsReady` and documented.

StorytellerCZ:
Yes, I think it should be public and documented, unless there is a reason why it should be private.

packages/ddp-client/common/namespace.js
```diff
@@ -89,3 +89,9 @@ DDP.onReconnect = callback => DDP._reconnectHook.register(callback);
 DDP._allSubscriptionsReady = () => allConnections.every(
   conn => Object.values(conn._subscriptions).every(sub => sub.ready)
 );
+
+// Sometimes data is loaded using a method, this hack provide a way
```
StorytellerCZ, line 93: Saying this is a hack is not good for an official function.
Please use JSDoc to document this function properly.

packages/ddp-client/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Meteor's latency-compensated distributed data client",
-  version: '2.6.1',
+  version: '2.6.2',
```
StorytellerCZ, line 3: This will have to be a feature release since you are adding a new feature.

StorytellerCZ (changes requested):

StorytellerCZ:
Also please add this into the docs.

italojs:
abandoned
btw, I cant see how this implementation can be a fix is you create a funciton but do not call it

PR #12714, opened 2023-07-19
fix(skell): remove builtin regex expressions from simpleSchema

vit0rr:
closes https://github.com/meteor/meteor/issues/12689

Open it again because I deleted the meteor repo from my account 😅

This pull request updates the no longer supported regEx built-in expressions from SimpleSchema.
The purpose here is to define a regular expression and link the SimpleSchema doc. that explains why it no longer supports built-in expressions.


StorytellerCZ (approved):

Grubba27:
Since this is just docs, will send this to devel

PR #12722, opened 2023-07-30
Updated Tailwind and Chakra-UI skeletons

fredmaiaarantes:
I have updated the dependencies for Tailwind and Chakra-UI skeletons and changed them to use React 18.


italojs:
We are on React 19, this PR is outdated
Closing since it havent any news after 2ys

Issue #11927, opened 2022-02-16
Does fetch dependencies need to be updated? 


harryadel:
The [dependencies](https://github.com/meteor/meteor/blob/devel/packages/fetch/package.js#L8) in fetch package are one major version behind. Should they be updated?
```
  "node-fetch": "2.3.0", // latest: [3.2.0](https://github.com/node-fetch/node-fetch)
  "whatwg-fetch": "2.0.4" // latest: [3.6.2](https://github.com/github/fetch#readme)
```

StorytellerCZ:
v3 of node-fetch is ESM package only which doesn't have such a good support in Meteor for now, so we'll have to test this thoroughly.
https://github.com/node-fetch/node-fetch/releases/tag/v3.0.0
https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md

harryadel:
I didn't know that. Thanks for pointing it out. How about [whatwg-fetch](https://github.com/github/fetch/releases/tag/v3.0.0)? Doesn't seem like it utilizes any of ESM stuff, no?

denihs:
Hi @harryadel, are you willing to create a PR to update these packages and test everything out?

I'm not sure if passing our automated tests would be enough, right @StorytellerCZ? I think the ideal would be to have an app using the `fetch` package as much as possible, testing different scenarios.

StorytellerCZ:
@denihs passing tests would be a good first indicator that there isn't anything major broken. Given the nature of changes testing it in simple app or with one of the oauth that use fetch would be the most efficient. If things load from the package then I think things will go smoothly as then it is just making sure that they truly loaded as we expected and comparing with changelog to see if there is something that we are specifically using that we need to take into account.

vlasky:
I had to stop using the Meteor fetch package and switch to using node-fetch 3.2.5 in order to be able to use the **insecureHTTPParser** option and fix compatibility with servers running Imperva Incapsula WAF. The requests had been failing with the HPE_INVALID_HEADER_TOKEN parse error.

Is there really any good reason to not update Meteor fetch to use the latest version of node-fetch?

StorytellerCZ:
@vlasky if node-fetch worked for you in a Meteor app, then we should be able to update this then.

harryadel:
I guess this is solved now thanks to @jamauro

PR #12739, opened 2023-08-09
Bump Npm.depends packages

jamauro:
Bump node-fetch and whatwg-fetch to the latest versions

### DIFF
--- a/packages/fetch/package.js
+++ b/packages/fetch/package.js
@@ -6,8 +6,8 @@ Package.describe({
 });
 
 Npm.depends({
-  "node-fetch": "2.3.0",
-  "whatwg-fetch": "2.0.4"
+  "node-fetch": "3.3.2",
+  "whatwg-fetch": "3.6.17"
 });
 
 Package.onUse(function(api) {


jamauro:
It's failing here https://github.com/meteor/meteor/blob/8818ccef3736f878c9744b701927768ae9b22643/tools/tests/apps/dynamic-import/tests.js#L18

with 
```js 
TypeError: fetch is not a function
```
I thought `fetch` was global so I'm not sure what the underlying issue is. I was hoping someone closer to the code might be able to figure it out. I suppose `fetch` could be explicitly imported into that test file with `import { fetch } from 'meteor/fetch'`. 

I also noticed that the `dynamic-import` tests are using `fetch@0.1.1` here:
https://github.com/meteor/meteor/blob/8818ccef3736f878c9744b701927768ae9b22643/tools/tests/apps/dynamic-import/.meteor/packages#L27C4-L27C4

Shouldn't this just be `fetch`? I'm not sure if that is intentional or an oversight.

StorytellerCZ (approved):

Grubba27:
> I also noticed that the dynamic-import tests are using fetch@0.1.1 here:
https://github.com/meteor/meteor/blob/8818ccef3736f878c9744b701927768ae9b22643/tools/tests/apps/dynamic-import/.meteor/packages#L27C4-L27C4

Maybe needs to be updated ? I saw theses days a test that only passed after we updated the `.meteor/package` file

jamauro:
@Grubba27 @denihs I think I'm stuck on this PR. It looks like the `fetch` package is required to use ecmascript 5 syntax. `node-fetch` has these requirements: https://github.com/node-fetch/node-fetch#loading-and-configuring-the-module

So I think the options are:
1. Remove `fetch` from the `check-legacy-syntax.js` here: https://github.com/meteor/meteor/blob/8818ccef3736f878c9744b701927768ae9b22643/scripts/admin/check-legacy-syntax/check-syntax.js#L53
so that we can use `import fetch from 'node-fetch'`

2. Pin `fetch` to `node-fetch@2` and continue using `const fetch = require('node-fetch')`. IMO, this isn't great.

3. Another idea that y'all have :)

jamauro:
> 2. Pin `fetch` to `node-fetch@2` and continue using `const fetch = require('node-fetch')`. IMO, this isn't great.

I went with this option for now. If y'all have a better solution, please let me know. I don't love being pinned to version 2 of `node-fetch` but they do say it will continue to receive critical bug fixes and it does fix the issue I was running into. 

Still it seems like Meteor should be able to use the latest and greatest. I'm sure there was a good reason to force some core packages to use ecmascript 5 – I'm guessing backwards compatibility – but maybe that restriction can/should be lifted in Meteor 3.0+? Curious to hear what y'all think.

harryadel:
Does this mean [this](https://github.com/meteor/meteor/issues/11927) is closed?

harryadel:
Do we have a release date on 2.14? cc @Grubba27 @denihs I understand 3.0 is taking most of your time but https://github.com/meteor/meteor/pull/12785 with other stuff need to be released soon and no work on 2.14 has been done so far, even the creation of a PR!

Again, thank you for your work. This is just a friendly reminder.

StorytellerCZ:
@harryadel I will be pushing @Grubba27 to make 2.14 the Hacktoberfest release! I think it could become a nice tradition going forward.

jamauro:
🎉

PR #12742, opened 2023-08-17
Add a new publication strategy

alisnic:
We have an application that publishes *a lot* of documents, from which only a small subset of documents is updated. We now have a few very heavy users which cause strain on our containers due to server merge box logic.

While in theory NO_MERGE solves the problem, our e2e suite is failing because NO_MERGE does not track document use across multiple publication from the same collection.

This PR is an attempt to solve the limitations of NO_MERGE, while keeping most of its the memory and CPU benefits.

This change was tested in a product e2e suite with 725 test examples.


packages/ddp-server/livedata_server.js
```diff
@@ -19,15 +20,28 @@ const publicationStrategies = {
...
   // NO_MERGE is similar to NO_MERGE_NO_HISTORY but the server will remember the IDs it has
   // sent to the client so it can remove them when a subscription is stopped.
   // This strategy can be used when a collection is only used in a single publication.
   NO_MERGE: {
+    useDummyDocumentView: false,
     useCollectionView: false,
     doAccountingForCollection: true,
+  },
+
+  // NO_MERGE_MULTE is similar to NO_MERGE, but it does track whether a document
```
StorytellerCZ, line 36: Typo here

packages/ddp-server/livedata_server.js
```diff
@@ -41,6 +55,26 @@ DDPServer.publicationStrategies = publicationStrategies;
 // Session and Subscription are file scope. For now, until we freeze
 // the interface, Server is package scope (in the future it should be
 // exported).
+var DummyDocumentView = function () {
+  var self = this;
+  self.existsIn = new Set(); // set of subscriptionHandle
+  self.dataByKey = new Map(); // key-> [ {subscriptionHandle, value} by precedence]
+};
+
+_.extend(DummyDocumentView.prototype, {
```
StorytellerCZ, line 64: Use `Object.assign()` instead.

StorytellerCZ (changes requested):

StorytellerCZ (approved):
Looks good so far, but would also appreciate an entry to the docs.

alisnic:
> Looks good so far, but would also appreciate an entry to the docs.

Sure thing! I wasn't sure the change would be accepted that's why I didn't include any changes to documentation.

Is changing `docs/pubsub.md` enough, or should I update the changelog as well?

StorytellerCZ (approved):

dd137:
(Hi there. There was a small omission in the `pubsub.md` update: "There are ~three~ **four** publications strategies:")

StorytellerCZ:
Fixed

PR #12743, opened 2023-08-18
Fix type declaration for MethodApplyOptions

ebroder:
Fixes the following error:

```
.meteor/local/types/node_modules/package-types/meteor/package/os/packages/meteor/meteor.d.ts:173:20 - error TS2304: Cannot find name 'Result'.

173           result?: Result
                       ~~~~~~
```


StorytellerCZ:
@ebroder can you please merge from `release-2.14` and update the blaze reference. Thanks!

ebroder:
Sure - fixed, I think.

PR #12754, opened 2023-08-23
Fix #11336 for forgotPassword

### LINKED ISSUE #11336: Accounts.forgotPassword permits user enumeration through inference

StorytellerCZ:
It was correctly pointed out by @tchax in #11336 that we are still returning error on `forgotPassword` which can be used to enumerate users through inference. I have added an extra condition to the error handling of missing user to only do so when `ambiguousErrorMessages` is set to `true`. This way there shouldn't be any errors returned in production, while still giving the developers the option for testing. Though we could have a debate if not to remove this error completely.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -376,6 +376,7 @@ Meteor.methods({forgotPassword: options => {
   if (!user && !Accounts._options.ambiguousErrorMessages) {
     Accounts._handleError("User not found");
   }
+  if (!user) return null
 
   const emails = pluckAddresses(user.emails);
   const caseSensitiveEmail = emails.find(
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -373,7 +373,7 @@ Meteor.methods({forgotPassword: options => {
 
   const user = Accounts.findUserByEmail(options.email, { fields: { emails: 1 } });
 
-  if (!user) {
+  if (!user && !Accounts._options.ambiguousErrorMessages) {
     Accounts._handleError("User not found");
   }
 


packages/accounts-password/password_server.js
```diff
@@ -373,9 +373,10 @@ Meteor.methods({forgotPassword: options => {
 
   const user = Accounts.findUserByEmail(options.email, { fields: { emails: 1 } });
 
-  if (!user) {
+  if (!user && !Accounts._options.ambiguousErrorMessages) {
     Accounts._handleError("User not found");
   }
+  if (!user) return null
```
Grubba27, line 379: This should not throw  `Accounts._handleError("User not found");`?

packages/accounts-password/password_tests.js
```diff
@@ -1370,9 +1370,9 @@ if (Meteor.isServer) (() => {
           );
 
         Accounts._options.ambiguousErrorMessages = true
-        test.throws(
-          ()=> Meteor.call('forgotPassword', wrongOptions),
-          'Something went wrong. Please check your credentials'
+        test.doesNotThrow(
```
Grubba27, line 1373: `doesNotThrow` is not a function in our tests, it only exists in `assert`

italojs:
@StorytellerCZ, is this PR still relevant?

packages/accounts-password/password_server.js
StorytellerCZ: Not if the `ambiguousErrorMessages` is set to `true`. Because even saying user not found is useful information to the attacker.

StorytellerCZ:
@italojs yes

nachocodoner:
It’s currently compared against 2.x.x (Meteor 2 branch). If it’s still relevant, it would be good to base it on devel, with all conflicts or adaptations properly handled for Meteor 3.

StorytellerCZ:
If I remember correctly, this might be a backport.

Update: @nachocodoner it appears that this has been originally made for the 2.x branch. It is missing in the current codebase as well. I will provide an improved PR for 3.x branch shortly.

packages/accounts-password/password_tests.js
StorytellerCZ: Yes, my mistake. Was looking on the Node assert library and did not double check.

StorytellerCZ:
Closing in favor of #14060
If there is interest, then we can backport this.

Issue #12649, opened 2023-05-23
New Async API are not in the docs


Grubba27:
Currently, our docs page does not include the `*Async` methods that were added in [2.8](https://guide.meteor.com/v2.10/2.8-migration.html).
![Screenshot 2023-05-23 at 09 58 38](https://github.com/meteor/meteor/assets/70247653/02edce51-eb8a-41a4-ac34-73b08a5f4c4a)

We should or could add a section explaining their existence or add them like their normal counterpart.

This issue relates to #12646

ciradu2204:
Hi @Grubba27, can I work on this?

Grubba27:
Sure, @ciradu2204 ! PRs are welcomed!

ciradu2204:
This is a bit more complicated than I expected😲. I guess I need to have
some experience using the AsyncApi in Meteor to be able to make any
changes. If yes, any resource to guide me would be appreciated. Or someone
with more experience can take it. Thanks

Ekott2006:
Hello @ciradu2204, Any Progress???

ankushdhavale:
can I work on this ?

Grubba27:
PRs are welcomed!

raunakkumarsingh:
can i work on this?

Grubba27:
Hey, @raunakkumarsingh  PRs are welcome. You can create a section for every new method, just like you posted there, or just a new section explaining their existence

PR #12758, opened 2023-08-24
doc: add Async methods

fvckDesa:
This PR resolve issue #12649

/claim #12649


Grubba27:
Have you tested your solution? 

Running locally I was not able to start the website.

Screenshot 2023-08-25 at 09 11 15

fvckDesa:
I test locally, jsdoc don't find the type declaration because the async methods not have the jsdoc comments
https://github.com/meteor/meteor/blob/527f0ffb6e050d7ed28721df31d1b1dcf2aa5ea7/packages/mongo/collection.js#L998-L1010

I can add them like this:
```js
/**
* @method findOneAsync
*/
/**
* @method createIndexAsync
*/
 ASYNC_COLLECTION_METHODS.forEach(methodName => {})
```
or removing this loop
```js
 ASYNC_COLLECTION_METHODS.forEach(methodName => {})
```

and add one by one the method with the jsdoc comment

Grubba27:
@fvckDesa  what you think is the better approach?

fvckDesa:
For me is better create a promisify utility function and add one by one the methods like the sync ones. So, if you want do a change in only one async method, you can do it and if you want do a common change on all async methods, you do it in the utility function

Grubba27:
I got to run it locally, and I like how it looks in the website. I would only change the code for what you sent earlier on:

> and add one by one the method with the jsdoc comment.
Another alternative is add the comments in the constant file where ASYNC_COLLECTION_METHODS is defined like this:

I think would look better for developers maintaining our code.

fvckDesa:
@Grubba27 like this?
```javascript
export const ASYNC_COLLECTION_METHODS = [
/**
* @method findOneAsync
*/
  'findOne',
/**
* @method insertAsync
*/
  'insert'
];
```

Grubba27:
Nice @fvckDesa thanks for your contribution. Hope to see you soon here :)
Sorry for the delay in the feedback loop, but well done!

PR #12764, opened 2023-09-01
Update NPM version to 9

denihs:
We released a new Alpha version with Node 18 and Npm 8 instead of 9.

That's because the command `npm config set python which python3 ` stops working in version 9. You can see this [issue](https://github.com/npm/cli/issues/6126) about it.

Doing some local tests, it seems like we don't need this command anymore to generate a new dev_bundle. We will generate a global one using our CI and see what happens.


Grubba27 (approved):
would like to see this working in a bundle but LGTM

PR #12773, opened 2023-09-08
Connect clean up

filipenevola:
Cleans up `connect` from code & readme.

Meteor 3 replaces `connect` with `express` but there are still some references in the code.

- [x] Remove unused connect dependency
- [x] Replaces explanation about how to use the module directly
- [x] Review webapp_server.js mentions to connect
- [x] Review `import connectRoute from 'connect-route'` in the docs example
- [x] Keep old names available but using express
- [x] Change expressXyz to only Xyz
- [x] Review and answer this [thread](https://forums.meteor.com/t/what-impact-does-webapp-transition-from-connect-to-express-have-on-community-packages-that-use-it/60541)


Grubba27 (changes requested):
From this comment [here](https://forums.meteor.com/t/what-impact-does-webapp-transition-from-connect-to-express-have-on-community-packages-that-use-it/60541/5?u=grubba) on Forums I would ask as well to bring back the `connect*` handlers but that do throw a warning or a log.

Grubba27:
> One idea would be to use just `Webapp.handlers` going forward so if for some reason express gets swapped with something else in the future you don’t need to rename anything 🙂

That sounds even better. What do you think @filipenevola ?

We depreciate but not remove `connectHandlers` and add and maintain just `handlers`

filipenevola:
Sure, I updated the checklist.

filipenevola:
@jamauro @Grubba27 code changes are done.

Now just missing to review all the mentions of connect in the code.

filipenevola:
@Grubba27 just so you know, 2 tests of webapp package are failing on my machine, the same happens in release-3.0 branch.

Also, FYI I'm running on ARM (Apple M2).

denihs (approved):

packages/webapp/webapp_server.js
```diff
@@ -1346,16 +1345,21 @@ async function runWebAppServer() {
...
+    rawConnectHandlers: rawExpressHandlers,
+    rawHandlers: rawExpressHandlers,
     httpServer: httpServer,
     expressApp: app,
     // For testing.
-    suppressConnectErrors: function() {
-      suppressConnectErrors = true;
-    },
+    suppressConnectErrors: suppressErrors,
+    suppressExpressErrors: suppressErrors,
```
radekmie, line 1362: Would it make sense to add runtime deprecation warnings?

radekmie (approved):

packages/webapp/webapp_server.js
filipenevola: IMO yes. Do you agree?

radekmie: Yeah, let's do that. While they have very similar APIs, there can be some nuanced differences.

Grubba27 (approved):

packages/webapp/webapp_server.js
radekmie: https://github.com/meteor/meteor/blob/e0631738f2a8a914d8a50b1060e8f40cb0873680/packages/meteor/helpers.js#L211-L222

filipenevola: done!

filipenevola:
@Grubba27 if all tests pass (I just synced with release-3.0) this PR is ready to be merged.

PR #12783, opened 2023-09-20
Move underscore types to dev dependencies in server bundle

StorytellerCZ:
I have noticed that we have a `devDependencies` in server dev bundle generator, so I have moved the `@types/underscore` into that section from the main dependencies as it is a dev dependency which I think it belongs.


StorytellerCZ:
Merged into #12832

Issue #427, opened 2012-10-27
Getting socket from a method


Charuru:
Is there a dependable way to getting a user's socket from a method (or any other way to access it in general).

I know I can get all the sockets `Meteor.default_server.stream_server.all_sockets()` and then find the user, but that just seems really bad.

PR #12801, opened 2023-10-02
Add Tracker.autorunAsync to allow for await-able autorun blocks

DanielDornhardt:
PR which adds a new, `await`-able `Tracker.autorunAsync` to allow Autorun blocks with async autorun functions to be run in a predictable manner.

There's also a separate PR on the blaze repo with an equivalent `Template.autorunAsync` function: [Add Template.autorunAsync to allow for await-able autorun blocks #427.
](https://github.com/meteor/blaze/pull/427)
The explanations in both PRs are basically identical, so if you understood the first one, you got both, only one's for `Tracker.autorunAsync`, and the other one is for `Template.aurorunAsync`.

The issue I'm trying to mitigate is this.

Having multiple "traditional" autorun blocks in a row like this, with async autorun funcs:

```
// autorun 1
Tracker.autorun(async(c) => {
    // 1
    // 2
    // 3
    // 4
})

// autorun 2
Tracker.autorun(async(c) => {
    // 5
    // 6
    // 7
    // 8
})

// autorun 3
Tracker.autorun(async(c) => {
    // 9
    // 10
    // 11
    // 12
})

```

... will lead to **each autorun being entered, executed til after the first `await`, then the interpreter will skip the rest of the autorun function, start with the next synchronous task in the script, eg. the autorun2 function.**

There it'll run til after the first `await` & yield, and then again with the third block.

After that all bets are off: Whenever some of the awaited code returns, it'll be continued eventually. Until the next await block.. :D 

**So the interpreter will actually zig-zag back and forth between the different Autorun blocks until all the awaits have been passed and finished.**

That makes it really hard to reason about the autoruns & probably might lead to issues during runtime too, eg. certain subscriptions not being made / ready because the interpreter "skipped ahead" a previous autorun, which until async were guaranteed to run in order once.

It gets even more fun in the context of eg. Blaze Templates, where for example here:

```

Template.onCreated(function() {
    // autorun 1
    Tracker.autorun(async(c) => {
        // 1
        // 2
        // (...)
    })

    // autorun 2
    Tracker.autorun(async(c) => {
        // 5
        // 6
        // (...)
    })

    // autorun 3
    Tracker.autorun(async(c) => {
        // 9
        // 10
        // (...)
    })
})

Template.onRendered(function() {
    // autorun 4
    Tracker.autorun(async(c) => {
        // 13
        // 14
        // (...)
    })
})

```

Here, autorun 4 in `onRendered` will be started to be executed _before_ any of the first autoruns have been ran through completely, if that makes sense.

The only way I could conceive to put everything "in order" again was to use `async` function in `onCreated` and `onRendered` and then making the autorun blocks `await`able by awaiting each one in turn.

This way it'd also be possible to wait for the `rendered` event by setting eg. a reactive var in `onRendered` and checking that in `onCreated` (or wherever you want) in an autorun...

---

So then the above code could look like this with my added `Tracker.autorunAsync` function:

```

    // autorun 1
    await Tracker.autorunAsync(async(c) => {
        // 1
        // 2
        // (...)
    })

    // autorun 2
    await Tracker.autorunAsync(async(c) => {
        // 5
        // 6
        // (...)
    })

    // autorun 3
    await Tracker.autorunAsync(async(c) => {
        // 9
        // 10
        // (...)
    })
})

Template.onRendered(async function() {
    // autorun 4
    await Tracker.autorunAsync(async(c) => {
        // 13
        // 14
        // (...)
    })
})

```

This forces all the code in the autorun blocks to run in their _expected_ order again. It's possible to _reason_ again about the order of things. Otherwise there might be a lot of really hard to debug race conditions.

The good thing is, the PR adds a new function, `Tracker.autorunAsync`, so it shouldn't break any existing code.

I've adopted the *Async syntax here for easy differentiation between Async and not-Async variants.

I'm not sure if that means that in the future / with 3.0 we should _only_ use the `autorunAsync` - functions, but for now I think its good to have these in there explicitly as an option.

I'm not sure if this should go into the mainline immediately? Please, I'd be happy for some feedback.

Also, sorry, there's no

a) Typescript definitions - I have no clue about that, but probably could fumble something together, but I wouldn't actually know how to check whether it'd be correct either
b) Tests - sorry guys
c) Nothing added to the Meteor Docs yet - let me know if this is something which might actually make it into meteor and I'll gladly add some lines to the docs.

Thx & best!

### DIFF
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -578,6 +578,37 @@ Tracker.autorun = function (f, options) {
   return c;
 };
 
+/**
+ * @callback Tracker.ComputationFunction
+ * @param {Tracker.Computation}
+ */
+/**
+ * @summary Await-able Tracker.autorun. Run a function now and rerun it later whenever its dependencies
+ * change. Returns a promise for the Computation object that can be used to stop or observe the
+ * rerunning.
+ * @locus Client
+ * @param {Tracker.ComputationFunction} runFunc The function to run. It receives
+ * one argument: the Computation object that will be returned.
+ * @param {Object} [options]
+ * @param {Function} options.onError Optional. The function to run when an error
+ * happens in the Computation. The only argument it receives is the Error
+ * thrown. Defaults to the error being logged to the console.
+ * @returns {Tracker.Computation}
+ */
+Tracker.autorunAsync = function (f, options) {
+  let fPromise
+  const newF = async(...args) => {
+    fPromise = f(...args)
+  }
+
+  const c = Tracker.autorun(newF, options)
+
+  return fPromise.then(() => {
+    return c
+  })
+};
+
+
 // http://docs.meteor.com/#tracker_nonreactive
 //
 // Run `f` with no current computation, returning the return value


packages/tracker/tracker.js
```diff
@@ -578,6 +578,37 @@ Tracker.autorun = function (f, options) {
...
+  let fPromise
+  const newF = async(...args) => {
+    fPromise = f(...args)
+  }
+
+  const c = Tracker.autorun(newF, options)
+
+  return fPromise.then(() => {
+    return c
+  })
```
radekmie, line 608: While this implementation is good (the names could be improved and the `newF` does not have to be `async`), I have a suggestion: instead of adding a new `Tracker` function (and `Template` function in https://github.com/meteor/blaze/pull/427), we could add something like `computation.firstRunPromise`.

The usage is fairly similar: instead of `await Tracker.autorunAsync()` you'd do `await Tracker.autorun().firstRunPromise` (eventually we could add `autorunAsync` as a convenience wrapper).

The main difference to me is that with the current implementation it's impossible to stop this computation before it finishes the first round. Of course, we won't implement cancellation just yet (maybe in the future?) but the `.stop()` function is simply not there until the first call finishes.

DanielDornhardt: Thank you @radekmie I'll try that, that sounds like an improvement, which I'm all for! :D 

> the names could be improved

Ah that hurt! 😄

PR #12803, opened 2023-10-04
Feature: interactive meteor create command

### LINKED ISSUE #12791: Feature - Make `meteor create` more interactive.
## Currently

Today, when using `meteor create` we must provide the name and the template that we want to use. For example:

```bash
meteor create blaze-app --blaze
```
_if no flag is passed it does create a `--react` app_

## Proposal

What I propose is we do an upgrade in the overall experience while creating an app.

> [!IMPORTANT]
> *Current features should not be changed.*

If the user uses just `meteor create` it should work similar to what we have today in `meteor generate` [command](https://docs.meteor.com/commandline.html#meteorgenerate), asking what the name of their app and which template it should use is.

Image of how `meteor generate` works today:
Screenshot 2023-09-26 at 12 46 21

What I was thinking is something like `vite` does in their `npm create vite` command

Screenshot 2023-09-26 at 14 29 13

mr-loop-1:
Cloese #12791 

If either or both of application name and skeleton flag is missed in the create command, there will be prompts for choosing them.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -745,49 +745,56 @@ main.registerCommand({
   const setup = async () => {
     const ask = createPrompt();
 
-    let appPathAsEntered, skeleton;
-    if(options.args.length === 1) {
-      appPathAsEntered = options.args[0];
-    }
-    else {
-      appPathAsEntered = await ask(`What is the name/path of your ${yellow`app`}? `);
-    }
-
     const skeletonExplicitOption = AVAILABLE_SKELETONS.find(skeleton =>
       !!options[skeleton]);
 
-    if(skeletonExplicitOption) {
-      skeleton = skeletonExplicitOption;
+    let appPathAsEntered, skeleton;
+    if(options.args.length === 1) {
+      appPathAsEntered = options.args[0];
+      if(skeletonExplicitOption) {
+        skeleton = skeletonExplicitOption;
+      }
+      else {
+        skeleton = DEFAULT_SKELETON;
+        Console.info(`No skeleton specified, using default ${green`${DEFAULT_SKELETON}`}`)
+      }
     }
     else {
-      // Constructing the prompt for choosing skeleton
-      // It can be made better with inquirer package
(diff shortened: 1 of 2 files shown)


mr-loop-1:
Hi @StorytellerCZ , 
The PR is ready for review. Also, I am not sure why there are changes in `packages/non-core/blaze` after I took the pull of devel branch. Can you please look into that as well.

Grubba27:
I added testing this PR to my todo's

tools/cli/commands.js
```diff
@@ -724,12 +742,65 @@ main.registerCommand({
...
+      appPathAsEntered = options.args[0];
+    }
+    else {
+      appPathAsEntered = await ask(`What is the name/path of your ${yellow`app`}? `);
+    }
+
+    const skeletonExplicitOption = AVAILABLE_SKELETONS.find(skeleton =>
+      !!options[skeleton]);
+
+    if(skeletonExplicitOption) {
```
zodern, line 759: If a name is provided as an arg, the interactivity should be disabled and it use the default skeleton if non was provided. This is needed to avoid breaking scripts that use `meteor create`.

wyq09:
邮件已收到   ~3Q~

tools/cli/commands.js
mr-loop-1: @zodern Yes, that would be one of cases where we wouldn't need to make prompts. 
So, in that case, we only bring up the prompts when app name is not provided?

zodern: > So, in that case, we only bring up the prompts when app name is not provided?

Yes.

Showing a message like you suggest when using the default skeleton makes sense.

mr-loop-1: @zodern made some changes

07tAnYa:
I would like to contribute here. Kindly assign this to me

Grubba27:
Needs a new dev-bundle. cc @StorytellerCZ

denihs (approved):
LGTM

PR #12805, opened 2023-10-04
Allow await-ing autoruns by adding firstRunPromise to computation objects

DanielDornhardt:
This'll replace my PR from yesterday: [Add Tracker.autorunAsync to allow for await-able autorun blocks #12801
](https://github.com/meteor/meteor/pull/12801)

This PR adds a new, `await`-able field, `Computation.firstRunPromise` to newly created computations to allow autorun blocks with async autorun functions to be run in a predictable manner.

The issue I'm trying to mitigate is this:

Having multiple "traditional" autorun blocks in a row like this, with async autorun funcs:

```
Template.myTemplate.onCreated(async function () {
    // autorun 1
    this.autorun(async(c) => {
        // 1
        // 2
        // 3
        // 4
    })

    // autorun 2
    this.autorun(async(c) => {
        // 5
        // 6
        // 7
        // 8
    })

    // autorun 3
    this.autorun(async(c) => {
        // 9
        // 10
        // 11
        // 12
    })
})
```

... will lead to **each autorun being entered, executed til after the first `await`, then the interpreter will skip the rest of the autorun function, start with the next synchronous task in the script, eg. the autorun 2 function.**

There it'll run til after the first `await` & yield, and then again with the third block.

After that all bets are off: Whenever some of the awaited code returns, it'll be continued eventually. Until the next await block.. :D 

**So the interpreter will actually zig-zag back and forth between the different Autorun blocks until all the awaits have been passed and finished.**

That makes it really hard to reason about the autoruns & probably might lead to issues during runtime too, eg. certain subscriptions not being made / ready because the interpreter "skipped ahead" a previous autorun, which until async were guaranteed to run in order once.

It gets even more fun if you start looking at the different template lifecycle functions like onCreated and onRendered for example:

```

Template.onCreated(function() {
    // autorun 1
    this.autorun(async(c) => {
        // 1
        // 2
        // (...)
    })

    // autorun 2
    this.autorun(async(c) => {
        // 5
        // 6
        // (...)
    })

    // autorun 3
    this.autorun(async(c) => {
        // 9
        // 10
        // (...)
    })
})

Template.onRendered(function() {
    // autorun 4
    Tracker.autorun(async(c) => {
        // 13
        // 14
        // (...)
    })
})

```

Here, autorun 4 in `onRendered` will be started to be executed _before_ any of the first autoruns have been ran through completely, if that makes sense.

The only way I could conceive to put everything "in order" again was to use `async` function in `onCreated` and `onRendered` and then making the autorun blocks `await`able and then by awaiting each one in turn.

This way it'd also be possible to wait for the `rendered` event by setting eg. a reactive var in `onRendered` and checking that in `onCreated` (or wherever you want) in an autorun...

---

So then the above code could look like this with the added `Computation.firstRunPromise`:

```

    // autorun 1
    await this.autorun(async(c) => {
        // 1
        // 2
        // (...)
    }).firstRunPromise

    // autorun 2
    await this.autorun(async(c) => {
        // 5
        // 6
        // (...)
    }).firstRunPromise

    // autorun 3
    await this.autorun(async(c) => {
        // 9
        // 10
        // (...)
    }).firstRunPromise
})

```

Each autorun returns their corresponding computation as before, with the difference that the field `firstRunPromise` has been added to the computations, thus making it possible to wait for them.

**This forces all the code in the autorun blocks to run in their _expected_ order again.** It's possible to _reason_ again about the order of things. Otherwise there might be a lot of really hard to debug race conditions.

The good thing is, the PR primarily adds a new field, `computation.firstRunPromise`, which, I hope,shouldn't break any existing code.

I'm not sure if this should go into the mainline immediately? Please, I'd be happy about all and any feedback.

Also, sorry, there's no

a) Typescript definitions - I have no clue about that, but probably could fumble something together, but I wouldn't actually know how to check whether it'd be correct either
b) Tests - sorry guys
c) Nothing added to the docs yet - let me know if this is something which might actually make it into meteor and I'll gladly add some lines to the docs.

Thx & best!

### DIFF
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -196,6 +196,27 @@ Tracker.Computation = class Computation {
     this._onError = onError;
     this._recomputing = false;
 
+    /**
+     * Computation.firstRunPromise will be set to the result of the call of the autorun function after the initial computation has been completed.
+     * If the autorun function is an async function, it'll then contain its promise, thus making the completion of the execution
+     * await-able.
+     *
+     * That allows us to manually synchronize autoruns like this:
+     *
+     * await Tracker.autorun(async () => {
+     *     await Meteor.userAsync();
+     *     (... more async code...)
+     * }).firstRunPromise;
+     *
+     * await Tracker.autorun(async () => {
+     *     await asyncSomeOrOther();
+     *     (... more async code...)
+     * }).firstRunPromise;
+     *
+     * // ta-daa! We'll get here only after both the autorun functions will have returned & executed in their entirety.
+     */
+    this.firstRunPromise = undefined;
+
     var errored = true;
     try {
       this._compute();
@@ -297,10 +318,18 @@ Tracker.Computation = class Computation {
 
     var previousInCompute = inCompute;
     inCompute = true;
+
     try {
-      Tracker.withComputation(this, () => {
+      // In case of async functions, the result of this function will contain the promise of the autorun function
+      // & make autoruns await-able.
+      const firstRunPromise = Tracker.withComputation(this, () => {
         withNoYieldsAllowed(this._func)(this);
-      });
+      })
+      // We'll store the firstRunPromise on the computation so it can be awaited by the callers, but only
+      // during the first run. We don't want things to get mixed up.
+      if (this.firstRun) {
+        this.firstRunPromise = firstRunPromise;
+      }
     } finally {
       inCompute = previousInCompute;
     }


packages/tracker/tracker.js
```diff
@@ -196,6 +196,27 @@ Tracker.Computation = class Computation {
...
+     *     await Meteor.userAsync();
+     *     (... more async code...)
+     * }).firstRunPromise;
+     *
+     * await Tracker.autorun(async () => {
+     *     await asyncSomeOrOther();
+     *     (... more async code...)
+     * }).firstRunPromise;
+     *
+     * // ta-daa! We'll get here only after both the autorun functions will have returned & executed in their entirety.
```
radekmie, line 216: JSDocs supports Markdown, so this could be indented (with 4 more spaces after `*`) to make it appear as a code block.

radekmie (commented):
Before releasing it:
1. TypeScript is as simple as adding another property in https://github.com/meteor/meteor/blob/4654adbe7ee21e48f1965f2e8a7bd6f9b08cac5b/packages/tracker/tracker.d.ts#L15-L18; `Promise<unknown>` would make the most sense, since it'll return the value of the function inside of `autorun`.
2. Tests are a must-have, but they should be pretty straightforward: test whether this new property actually resolves in the correct order.
3. Documentation could be mostly copied from the JSDocs and PR description.

radekmie:
An idea for tests:

```ts
Tinytest.addAsync('tracker - async function - synchronize', async test => {
  let counter = 0;

  await Tracker.autorunAsync(async () => {
    test.equal(counter, 0);
    counter += 1;
    test.equal(counter, 1);
    await Promise.resolve();
    test.equal(counter, 1);
    counter *= 2;
    test.equal(counter, 2);
  });

  await Tracker.autorunAsync(async () => {
    test.equal(counter, 2);
    counter += 1;
    test.equal(counter, 3);
    await Promise.resolve();
    test.equal(counter, 3);
    counter *= 2;
    test.equal(counter, 6);
  });
});
```

Grubba27:
This was merged in #12824 thanks a lot @DanielDornhardt

PR #12808, opened 2023-10-05
Solve error when trying to use insert/update/remove from the client

denihs:
This error was thrown when calling writing methods (insert/update/remove) from the client in a prototype app (uses the `insecure` package):

image

That happens because the `insecure` package allows you to call a function like `TasksCollection.insert({text: "New Task"})` from the client then it [calls](https://github.com/meteor/meteor/blob/678506b398473e9aa0287c5c1024022a3a191274/packages/allow-deny/allow-deny.js#L217) that same method in the server to persist the data. But these sync methods no longer exist in the server, hence the error.

So now we are creating this mapper that calls the respective async method on the server side and persists the data.


Grubba27 (approved):
ltgm

PR #12809, opened 2023-10-05
Fix/Underscore

KushalSr:


StorytellerCZ:
What is this about?

KushalSr:
@StorytellerCZ I've updated the underscore version. Please check once.

StorytellerCZ:
I don't see any update, I only see the addition of underscore to the root of the project. That is why I am asking what this is suppose to do.

Issue #12781, opened 2023-09-20
Bugs in negated `in` expressions


gorosgobe:
`in` and `instanceof` expressions in JS
```js
a in obj;
a instanceof C;
```
can be negated by grouping them and applying the `!` operator, i.e.

```js
!(a in obj);
!(a instanceof C);
```

Applying the `!` operator incorrectly (on the LHS operand) leads to bugs:
```js
!a in obj; // will evaluate to false, unless obj has a "true" or "false" key
!a instanceof C; // will evaluate to false, unless C overrides instanceof with a @@hasInstance method
```
For more information, please see these [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#not_an_instanceof) and [the no-unsafe-negation recommended Eslint rule](https://eslint.org/docs/latest/rules/no-unsafe-negation).

I have found several potentially problematic instances of the above bugs in your codebase:
https://github.com/meteor/meteor/blob/b8d4d4ce10395a27c751515a17d37b7f9a502587/tools/cordova/builder.js#L803
https://github.com/meteor/meteor/blob/b8d4d4ce10395a27c751515a17d37b7f9a502587/tools/runners/run-hmr.js#L276

Grubba27:
Thanks for the information @gorosgobe! if you want you can tackle this one. I added a `good first issue` label on it

1-ankush-1:
.take

bharath090909:
Did you address the issue, @gorosgobe?

 @Grubba27, if @gorosgobe hasn't addressed it, I'd be happy to take it on.

gorosgobe:
Please feel free to address this @1-ankush-1 @bharath090909 , I won't be taking it on.

Madhawa97:
You can close this issue, right?

PR #12810, opened 2023-10-05
Fix/bugs in negated #12781

KushalSr:

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,5 @@
     }
   },
   "dependencies": {
-    "underscore": "^1.13.6"
   }
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -29,11 +29,16 @@
...
+            "resolve"
+          ]
         }
       ],
       "no-use-before-define": "warn",
       "import/no-unresolved": "warn"
     }
+  },
+  "dependencies": {
+    "underscore": "^1.13.6"
```
StorytellerCZ, line 42: Adding this does not seem to be related to the issue.

StorytellerCZ (changes requested):

package.json
KushalSr: @StorytellerCZ I've fixed that. Please check once.

Thanks

StorytellerCZ:
This has already been solved in #12781 and will be released in 2.14

PR #12813, opened 2023-10-06
Proposal for utilities package

StorytellerCZ:
Proposed utilities package based on the #12266 PR and the following discussion.

Right now pretty much just a pseudo-code showing how this package could work. The idea is to centralize any outside utilities dependencies management into this package plus any other utilities functions used across multiple packages can be moved in here.

Thoughts?


StorytellerCZ:
I'll try to update this soon.

italojs:
PR abandoned after 2y

Issue #12233, opened 2022-10-19
Add --open flag to open user browser when start serving


Grubba27:
When you want to run the project and wants to open the browser when it finishes building
it would look like this:

```bash
meteor run --open
```
_or_

```bash
meteor --open
```

it would start running the server then open your default browser

## How to implement?

 - [ ]  you are going to start here [file](thub.com/meteor/meteor/blob/ae7ccaf8051030f180d4adfcb157676703563a09/tools/cli/commands.js#L301-L302) adding the new option;
 - [ ] Maybe make changes [here](https://github.com/meteor/meteor/blob/8813bdd3df0e7fff61c5eca9683924e6d692f4f7/tools/runners/run-all.js#L19) as well
 - [ ] Install the [open](https://www.npmjs.com/package/open) npm package
 - [ ] use the ``ROOT_URL``variable to open in the browser

---
discussion for more context

Discussed in https://github.com/meteor/meteor/discussions/12226

ankitzm:
Would love to take up this issue, can you assign me the same !!

Jaya-sys:
Hi can you assign this issue to me?

Grubba27:
hey @ankitzm how it is going? it is looking good?

nameception:
Hello, i tried working on this issue but when importing the open package via `import { open } from 'open'` gives me this error: 

> Error [ERR_REQUIRE_ESM] [ERR_REQUIRE_ESM]: Must use import to load ES Module: C:\Users\user\Desktop\meteor\node_modules\open\index.js
require() of ES modules is not supported.
require() of C:\Users\user\Desktop\meteor\node_modules\open\index.js from C:\Users\user\Desktop\meteor\tools\cli\commands.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from C:\Users\user\Desktop\meteor\node_modules\open\package.json.

I also attempted the solution suggested here: https://github.com/sindresorhus/file-type/issues/525#issuecomment-1013041630  but it doesn't seem to work, no errors but the browser window did not open as expected.
Here is how I tried to do it 
```javascript// Opens a browser window when it finishes building
main.registerCommand({
  name: '--open',
  requiresRelease: false,
  pretty: false,
  catalogRefresh: new catalog.Refresh.Never()
}, function (options) {
  (async () => {
[...]

PR #12819, opened 2023-10-10
 Add --open option to run command #12233

nameception:
My attempt at adding the --open option for the run command #12233, wasn't able to test it as a option since "run" asks for a meteor app, which i wasn't able to create using the checkout version.


tools/cli/commands.js
```diff
@@ -406,6 +408,16 @@ function doRunCommand(options) {
     runLog.setRawLogs(true);
   }
 
+  //Opens a browser window when it finishes building
+  if (options.open) {
+    if(process.env.ROOT_URL){
+      open(process.env.ROOT_URL)
+    } else {
+      open('https://localhost:3000')
```
StorytellerCZ, line 416: I think the port should be taken from the `port` option and default to 3000 only if none is defined.

StorytellerCZ (changes requested):

tools/cli/commands.js
nameception: Updated it, if a port is provided it will use it, if not it should default to 3000

package.json
```diff
@@ -29,11 +29,16 @@
...
+            "resolve"
+          ]
         }
       ],
       "no-use-before-define": "warn",
       "import/no-unresolved": "warn"
     }
+  },
+  "dependencies": {
+    "open": "8.4.2"
```
Grubba27, line 42: this should not be here and there is no need to update package-lock.json

Grubba27:
I made some small adjustments and tested them! It looks great! Thanks a lot for your contribution.

@StorytellerCZ this should go to a dev-bundle branch as it adds a new dependency to our dev-bundle.

Grubba27 (approved):

denihs (approved):

PR #12824, opened 2023-10-12
Add firstRunPromise

harryadel:
This is a continuation of @DanielDornhardt work, where @radekmie suggestions are applied. I'm unsure whether we should go forward with this considering what was [brought up](https://meteor-community.slack.com/archives/C04USBMR3QW/p1697106769035169) in the slack channel. But I decided to push either way just in case.

### DIFF
--- a/docs/source/api/tracker.md
+++ b/docs/source/api/tracker.md
@@ -314,6 +314,26 @@ recomputed at flush time.
 This property is a convenience to support the common pattern where a
 computation has logic specific to the first run.
 
+{% apibox "Tracker.Computation#firstRunPromise" %}
+
+`Computation.firstRunPromise` will be set to the result of the call of the autorun function after the initial computation has been completed. If the autorun function is an async function, it'll then contain its promise, thus making the completion of the execution await-able. That allows us to manually synchronize autoruns like this:
+
+```js  
+
+await Tracker.autorun(async () => {
+  await Meteor.userAsync();
+  (...more async code...)
+}).firstRunPromise;
+
+await Tracker.autorun(async () => {
+  await asyncSomeOrOther();
+  (...more async code...)
+}).firstRunPromise;
+
+// ta-daa! We'll get here only after both the autorun functions will have returned & executed in their entirety.
+
+
+```
 <h2 id="tracker_dependency"><span>Tracker.Dependency</span></h2>
 
 A Dependency represents an atomic unit of reactive data that a
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -196,6 +196,27 @@ Tracker.Computation = class Computation {
     this._onError = onError;
     this._recomputing = false;
 
+    /**
+     * Computation.firstRunPromise will be set to the result of the call of the autorun function after the initial computation has been completed.
+     * If the autorun function is an async function, it'll then contain its promise, thus making the completion of the execution
+     * await-able.
+     *
+     * That allows us to manually synchronize autoruns like this:
+     *
+     * await Tracker.autorun(async () => {
+     *     await Meteor.userAsync();
+     *     (... more async code...)
+     * }).firstRunPromise;
+     *
+     * await Tracker.autorun(async () => {
+     *     await asyncSomeOrOther();
+     *     (... more async code...)
+     * }).firstRunPromise;
+     *
+     * // ta-daa! We'll get here only after both the autorun functions will have returned & executed in their entirety.
+     */
+    this.firstRunPromise = undefined;
+
     var errored = true;
     try {
       this._compute();
@@ -297,10 +318,18 @@ Tracker.Computation = class Computation {
 
     var previousInCompute = inCompute;
     inCompute = true;
+
     try {
-      Tracker.withComputation(this, () => {
+      // In case of async functions, the result of this function will contain the promise of the autorun function
+      // & make autoruns await-able.
+      const firstRunPromise = Tracker.withComputation(this, () => {
         withNoYieldsAllowed(this._func)(this);
-      });
+      })
+      // We'll store the firstRunPromise on the computation so it can be awaited by the callers, but only
+      // during the first run. We don't want things to get mixed up.
+      if (this.firstRun) {
+        this.firstRunPromise = firstRunPromise;
+      }
     } finally {
       inCompute = previousInCompute;
     }
--- a/packages/tracker/tracker.js
+++ b/packages/tracker/tracker.js
@@ -197,23 +197,12 @@ Tracker.Computation = class Computation {
     this._recomputing = false;
 
     /**
-     * Computation.firstRunPromise will be set to the result of the call of the autorun function after the initial computation has been completed.
-     * If the autorun function is an async function, it'll then contain its promise, thus making the completion of the execution
-     * await-able.
-     *
-     * That allows us to manually synchronize autoruns like this:
-     *
-     * await Tracker.autorun(async () => {
-     *     await Meteor.userAsync();
-     *     (... more async code...)
-     * }).firstRunPromise;
-     *
-     * await Tracker.autorun(async () => {
-     *     await asyncSomeOrOther();
-     *     (... more async code...)
-     * }).firstRunPromise;
-     *
-     * // ta-daa! We'll get here only after both the autorun functions will have returned & executed in their entirety.
+     * @summary Forces autorun blocks to be executed in synchronous-looking order by storing the value autorun promise thus making it awaitable.
+     * @locus Client
+     * @memberOf Tracker.Computation
+     * @instance
+     * @name  firstRunPromise
+     * @returns {Promise<unknown>}
      */
     this.firstRunPromise = undefined;
 
--- a/packages/tracker/tracker_tests.js
+++ b/packages/tracker/tracker_tests.js
@@ -630,6 +630,30 @@ Tinytest.addAsync('tracker - async function - stepped', async function (test) {
   test.equal(count, limit, 'after resolve');
 });
 
+
+Tinytest.addAsync('tracker - async function - synchronize', async test => {
+  let counter = 0
+  await Tracker.autorun(async () => {
+    test.equal(counter, 0);
+    counter += 1;
+    test.equal(counter, 1);
+    await Promise.resolve();
+    test.equal(counter, 1);
+    counter *= 2;
+    test.equal(counter, 2);
+  }).firstRunPromise;
+
+  await Tracker.autorun(async () => {
+    test.equal(counter, 2);
+    counter += 1;
+    test.equal(counter, 3);
+    await Promise.resolve();
+    test.equal(counter, 3);
+    counter *= 2;
+    test.equal(counter, 6);
+  }).firstRunPromise;
+})
+
 Tinytest.add('computation - #flush', function (test) {
   var i = 0, j = 0, d = new Tracker.Dependency;
   var c1 = Tracker.autorun(function () {
--- a/packages/tracker/tracker.d.ts
+++ b/packages/tracker/tracker.d.ts
@@ -17,6 +17,10 @@ export namespace Tracker {
      */
     firstRun: boolean;
     /**
+     * Forces autorun blocks to be executed in synchronous-looking order by storing the value autorun promise thus making it awaitable.
+     */
+    firstRunPromise: Promise<unknown>
+    /**
      * Invalidates this computation so that it will be rerun.
      */
     invalidate(): void;


packages/tracker/tracker.js
```diff
@@ -196,6 +196,16 @@ Tracker.Computation = class Computation {
     this._onError = onError;
     this._recomputing = false;
 
+    /**
+     * @summary Forces autorun blocks to be executed in synchronous-looking order by storing the value autorun promise thus making it awaitable.
+     * @locus Client
```
Grubba27, line 201: Is available in both server and client

packages/tracker/tracker.js
```diff
@@ -196,6 +196,16 @@ Tracker.Computation = class Computation {
...
+     * @memberOf Tracker.Computation
+     * @instance
+     * @name  firstRunPromise
+     * @returns {Promise<unknown>}
+     */
+    this.firstRunPromise = undefined;
```
Grubba27, line 207: I would try thinking of a more suitable name for this property, `firstRunPromise` sounds good but what about `invocationPromise` or something like that, not sure if you guys have any ideas(cc: @radekmie @DanielDornhardt ), `firstRunPromise` does not seem to express exactly what this does.

harryadel:
Is there a way for us to make `firstRunPromise` the default behavior? 

```js

    // autorun 1
    await this.autorun(async(c) => {
        // 1
        // 2
        // (...)
    }).firstRunPromise

    // autorun 2
    await this.autorun(async(c) => {
        // 5
        // 6
        // (...)
    }).firstRunPromise

    // autorun 3
    await this.autorun(async(c) => {
        // 9
        // 10
        // (...)
    }).firstRunPromise
})

```

Would become:

```js

    // autorun 1
    await this.autorun(async(c) => {
        // 1
        // 2
        // (...)
    })

    // autorun 2
    await this.autorun(async(c) => {
        // 5
        // 6
        // (...)
    })

    // autorun 3
    await this.autorun(async(c) => {
        // 9
        // 10
        // (...)
    })
})

```

My reason for this? [Predictability](https://github.com/meteor/docs/blob/version-NEXT/long-form/tracker-manual.md#predictability)! Which is something the original MDG considered when they ran into a similar problem regarding "flush" cycle in general. This is where [principle of least astonishment/surpise](https://rubyonrails.org/doctrine#optimize-for-programmer-happiness) comes into play.  *We may also have to implement something akin to `flush` where it'd allow you to invoke a specific autorun before the others.* Can this be done already using `invalidate` ??? 

```js

    // autorun 1
    const firstComputation = await this.autorun(async(c) => {
        // 1
        // 2
        // (...)
    })

    // autorun 2
    const secondComputation = await this.autorun(async(c) => {
        // 5
        // 6
        // (...)
    })

    // autorun 3
    const thirdComputation = await this.autorun(async(c) => {
        // 9
        // 10
        // (...)
    })
    
    // then we do something like this
    // this causes the computations/autoruns to run
    // into the non default way if users may wish to do so for certain reasons
    secondComputation.invalidate();
    firstComputation.invalidate();
    thirdComputation.invalidate();
})

```


I've talked with @DanielDornhardt about this and he told me to implement this utility method would be impossible on a technical level as you can't dictate the flow of async once it starts. It's has to do with language level and how JS/async works in general.     

----
There was some article which I can't find which talked about POLA, error compilers and something like "programmer pitfall of happiness", can anybody help me find it?

packages/tracker/tracker.js
harryadel: AFAIK the answer'd be yes. This is tracker package specific regardless of where it's gets executed.

radekmie: Yes, it's available on both.

packages/tracker/tracker.js
radekmie: I honestly think it makes sense, but if you have a different one - let's use that. (It may not be needed, though - I'll add a comment below.)

radekmie:
@harryadel 
* We can add `.then` on the `Computation` object so it'd be `await`able (it still requires something like `firstRunPromise` but it can remain a private field).
* I didn't get this issue with `invalidate` - it should work just like it was before.

@DanielDornhardt 
* We do a similar thing with `useTracker` in React components, but we do `subscribe` _in them_, so it's more like one `autorun` with `subscribe`+`if (subscription.ready())`+logic.

DanielDornhardt:
@radekmie 

The issue with `firstRunPromise` is nice but it'll only work for the first run of each autorun...

And if there are eg. a few guard clauses on top of the autorun function, after the first `return` the `firstRunPromise` will be resolved.

All following in the autorun `await`'s will be yielding and revisited in quasi-random order throughout the project.

This is especially problematic for eg. method calls with stubs - another one of our old friends I guess: 

[Async Methods with Callback by denihs · Pull Request #12677 · meteor/meteor · GitHub](https://github.com/meteor/meteor/pull/12677)

```
async onCreated() {
    await this.autorun(async () => {
        if (await Meteor.loggingInAsync()) {
            return
        }
        const x = await Meteor.callAsync('methodA')
        const y = await Meteor.callAsync('methodB')
        const z = await Meteor.callAsync('methodC')
    }).firstRunPromise

    await this.autorun(async () => {
        if (await Meteor.loggingInAsync()) {
            return
        }
        const x = await Meteor.callAsync('methodD')
        const y = await Meteor.callAsync('methodE')
        const z = await Meteor.callAsync('methodF')
    }).firstRunPromise
}
asyncOnRendered() {
    await this.autorun(async () => {
            if (await Meteor.loggingInAsync()) {
                return
            }
            const x = await Meteor.callAsync('methodX')
            const y = await Meteor.callAsync('methodY')
            const z = await Meteor.callAsync('methodZ')
    }).firstRunPromise
}

```

This will now lead to methods called in overlapping fashions, and, in case of methods with async stubs on the client, I think they'll complain that they're already in another stub, right?

So... `firstRunPromise` is nice, but after the first `return` all bets are off?

packages/tracker/tracker.js
DanielDornhardt: I like `firstRunPromise`, i think it pretty straightforward in what it does... and it also tells the story that this will only be for the first run, for example it expresses better that even if you use guard clauses and stuff this'll no longer help you...

```js
Tracker.autorun(async() {
    if (await Meteor.loggingInAsync()) {
        return // this is what the promise will resolve to
    } 
    if (!await Meteor.user().isAllowedHere()) {
        return // this will not get its own promise
    } 
    // lots of code
    // will happen WAY after the `firstRunPromise` / `invocationPromise` has been resolved
    await a...
    await b...
})
```

harryadel:
I spoke with Daniel again, and we're trying to get this merged. So, do you have any comments, @radekmie? 

The only thing I feel worth considering is this:
 
>We can add .then on the Computation object so it'd be awaitable (it still requires something like firstRunPromise but it can remain a private field).

Do I attempt to add the `.then` or do we merge this in and iterate on it in another PR?

radekmie:
OK, so there are two things:
1. The problem where subsequent `autorun`s are not blocking each other. I understood that the `firstRunPromise` was needed for _initialization_, e.g., in Blaze (IIRC, that's where it came from).

    If we're talking about later rerenders, it'd need to schedule whole `autorun`s, not their parts. That's basically the same problem as with methods and stubs. However, you could try wrapping all of the methods into a block and queue them. Something like this:
    ```ts
    import PQueue from 'p-queue';
    const queue = new PQueue({ concurrency: 1 });

    await this.autorun(async () => {
      if (await Meteor.loggingInAsync()) {
          return;
      }

      await queue.add(async () => {
        const x = await Meteor.callAsync('methodA');
        const y = await Meteor.callAsync('methodB');
        const z = await Meteor.callAsync('methodC');
      });
    }).firstRunPromise;

    await this.autorun(async () => {
      if (await Meteor.loggingInAsync()) {
          return;
      }

      await queue.add(async () => {
        const x = await Meteor.callAsync('methodD');
        const y = await Meteor.callAsync('methodE');
        const z = await Meteor.callAsync('methodF');
      });
    }).firstRunPromise;
    ```

    _(Side note: `loggingInAsync`? It's a `ReactiveVar#get` call - there's no point for it to be `async`.)_
3. As for the `then` method, it should be as simple as:
    ```ts
    class Computation {
      then(onResolved, onRejected) {
        return this.firstRunPromise.then(onResolved, onRejected);
      }
    }
    ```

polygonwood:
The main issue we see in our application is that the lifecycle of Blaze is 'broken' because the order of onCreated on Rendered and onDestroyed is no longer guaranteed : the first async call in an earlier stage fires the next stage.
We had this problem already before the 3.0 story, e.g. with setting up media connections which were already async implemented. The way we solved this is by using a 'runOnce' autorun (forced to run once 'really' because we stop the computation ourselves) by using a reactive variable (typically a Session variable if it's across modules) to hold and release the computation.
E.g. a subset of how we solved that
 
`async onCreated() {
   Session.set('waitFlag',false);
   await SomethingWeHavetoWaitFor();
   Session.set('waitFlag',true);
} 

async onRendered() {
   this.autorun(async (computation) => {
     if (Session.get('flag') {
        await DoWhatHasToBeDoneAfterDOMReady();
      // if needed set another reactive varibale to kick helpers e.g.
       computation.stop()
    }
})
}

Perhaps a Blaze version aware of async lifecylces (because it's not always required) could have some standardized solution for this ?

We have used similar approaches to e.g. dynamic subscriptions (if it can not be solved by the Router).

Maybe this can inspire someone ?

`

Grubba27 (changes requested):
Looks good but I would ask  making what @radekmie  suggested in the Computation object, adding a .then property so we could use this like this:

```js
await Tracker.autorun(async () => {
  await Meteor.userAsync();
  (...more async code...)
})
```

packages/tracker/tracker_tests.js
```diff
@@ -630,6 +630,30 @@ Tinytest.addAsync('tracker - async function - stepped', async function (test) {
   test.equal(count, limit, 'after resolve');
 });
 
+
+Tinytest.addAsync('tracker - async function - synchronize', async test => {
```
radekmie, line 634: Let's add another test like this without `.firstRunPromise` too.

packages/tracker/tracker.js
```diff
@@ -212,13 +212,13 @@ Tracker.Computation = class Computation {
    * @param {*} onRejected 
    * @returns 
    */
-  this.then = function (onResolved, onRejected) {
-    return this.firstRunPromise.then(onResolved, onRejected);
+  this.then = async function (onResolved, onRejected) {
+    await this.firstRunPromise.then(onResolved, onRejected);
```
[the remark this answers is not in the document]
radekmie: I've fixed the problem - missing `return`. I also updated the test to actually check it (right now it was passing and yet not working).

radekmie (approved):

docs/source/api/tracker.md
```diff
@@ -302,6 +302,26 @@ recomputed at flush time.
...
+  (...more async code...)
+}).firstRunPromise;
+
+await Tracker.autorun(async () => {
+  await asyncSomeOrOther();
+  (...more async code...)
+}).firstRunPromise;
+
+// ta-daa! We'll get here only after both the autorun functions will have returned & executed in their entirety.
+
```
Grubba27, line 322: Maybe remove this comment? we are already explaining this in the paragraph above.
Add examples that are not using `.firstRunPromise` as well.

docs/source/api/tracker.md
```diff
@@ -320,9 +320,9 @@ await Tracker.autorun(async () => {
 
 ```
 
-For a better developer experience `firstRunPromise` is automagically appended to your async `autorun` calls so you don't have to write them yourself. Meaning this also works:
```
StorytellerCZ: `automagically`, 🤔 I really like this word. Probably a typo here, but I think this should be something we should strive for in Meteor.

Grubba27: It happened only twice in our code base, and it does not follow that much how we write our docs, but I agree. We could review our documentation guide line and make some adjustments to be more pleasing to read.

I like this writing in docs(Puns and [funny comments](https://github.com/meteor/meteor/pull/12824/commits/cc07fe5d03b1093f15c4214e9363bc37499c2404)) generally, but I need to maintain how it is today.

harryadel: It's not a typo. I meant to use it since Meteor is all about that magic. Pffffft you guys won't get it!

StorytellerCZ:
Looks like this is ready to be merged.

harryadel:
@StorytellerCZ BRING THE "ta-da" AND "automagically" BACK

@Grubba27 You label yourself as an "open sourcerer" and you can't even do no automagic? YOU'RE A PHONY!

grubba_is_no_wizard

PR #12825, opened 2023-10-13
Updated the underscore to 1.9.2

07tAnYa:

### DIFF
--- a/docs/generators/changelog/versions/underscore version-1.9.2
+++ b/docs/generators/changelog/versions/underscore version-1.9.2
@@ -1,0 +1,46 @@
+## v1.9.2, 2023-10-13
+
+### Highlights
+
+* Solves the issue [#12798 which was actually related to the issue 12673: Previous Version suddenly requires a newer glibc version](https://github.com/meteor/meteor/issues?q=12798).
+
+#### Breaking Changes
+
+N/A
+
+####  Internal API changes
+
+N/A
+
+#### Migration Steps
+
+Please run the following command to update your project:
+
+```bash
+
+meteor update --release 1.9.2
+
+```
+
+
+#### Meteor Version Release
+
+
+Command line using npm:
+
+npm install underscore@1.9.2
+
+Command line using yarn:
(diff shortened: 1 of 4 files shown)


docs/generators/changelog/versions/underscore version-1.9.2
StorytellerCZ, line 1: This is not how you add to the changelog. Please refer to the contributing guide.

docs/package.json
```diff
@@ -16,7 +16,7 @@
     "hexo-renderer-marked": "2.0.0",
     "hexo-server": "1.0.0",
     "hexo-versioned-netlify-redirects": "1.1.0",
-    "jsdoc": "3.5.5",
+    "jsdoc": "4.0.2",
```
StorytellerCZ, line 19: Why is this needed?

package.json
```diff
@@ -29,11 +29,70 @@
...
+            "resolve"
+          ]
         }
       ],
       "no-use-before-define": "warn",
       "import/no-unresolved": "warn"
     }
+  },
+  "dependencies": {
+    "underscore": "^1.9.2"
```
StorytellerCZ, line 42: This is not how you update dependencies in Meteor.js. Adding here is not allowed and has no impact on what you are claiming to solve.

package.json
```diff
@@ -29,11 +29,70 @@
...
+  },
+  "scripts": {
+    "start": "node script.js",
+    "build": "npm build",
+    "test": "npm start"
+  }
+}
+{
```
StorytellerCZ, line 50: The following looks like a copy from some starter app, this has no place here and breaks `package.json` as from this place onward it breaks the json format.

StorytellerCZ (changes requested):
Honestly I don't understand what this is trying to do.

docs/generators/changelog/versions/underscore version-1.9.2
07tAnYa: ok sir

PR #12826, opened 2023-10-13
add: depracated notices added for the old methods that are depracated in version 2.8

guncebektas:

### DIFF
--- a/packages/mongo/mongo.d.ts
+++ b/packages/mongo/mongo.d.ts
@@ -107,6 +107,8 @@ export namespace Mongo {
       byteSize?: number,
       maxDocuments?: number
     ): Promise<void>;
+
+    /** @deprecated since 2.8 */
     createIndex(
       indexSpec: NpmModuleMongodb.IndexSpecification,
       options?: NpmModuleMongodb.CreateIndexesOptions
@@ -150,11 +152,13 @@ export namespace Mongo {
     ): Cursor<T, DispatchTransform<O['transform'], T, U>>;
     /**
      * Finds the first document that matches the selector, as ordered by sort and skip options. Returns `undefined` if no matching document is found.
+     * @deprecated since 2.8
      * @param selector A query describing the documents to find
      */
     findOne(selector?: Selector<T> | ObjectID | string): U | undefined;
     /**
      * Finds the first document that matches the selector, as ordered by sort and skip options. Returns `undefined` if no matching document is found.
+     * @deprecated since 2.8
      * @param selector A query describing the documents to find
      */
     findOne<O extends Omit<Options<T>, 'limit'>>(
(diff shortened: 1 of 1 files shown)


packages/mongo/mongo.d.ts
```diff
@@ -178,17 +182,20 @@ export namespace Mongo {
     ): Promise<DispatchTransform<O['transform'], T, U> | undefined>;
     /**
      * Gets the number of documents matching the filter. For a fast count of the total documents in a collection see `estimatedDocumentCount`.
+     * @deprecated
```
StorytellerCZ, line 185: Incorrect, `countDocuments` is one of the replacements for `count`.

packages/mongo/mongo.d.ts
```diff
@@ -178,17 +182,20 @@ export namespace Mongo {
...
      * @param selector The query for filtering the set of documents to count
      * @param options All options are listed in [MongoDB documentation](https://mongodb.github.io/node-mongodb-native/4.11/interfaces/CountDocumentsOptions.html). Please note that not all of them are available on the client.
      */
     countDocuments(selector?: Selector<T> | ObjectID | string, options?: NpmModuleMongodb.CountDocumentsOptions): Promise<number>;
     /**
      * Gets an estimate of the count of documents in a collection using collection metadata. For an exact count of the documents in a collection see `countDocuments`.
+     * @deprecated
```
StorytellerCZ, line 192: Incorrect, `estimatedDocumentCount` is one of the replacements for `count`.

StorytellerCZ (changes requested):

StorytellerCZ (approved):

packages/mongo/mongo.d.ts
```diff
@@ -107,6 +107,8 @@ export namespace Mongo {
       byteSize?: number,
       maxDocuments?: number
     ): Promise<void>;
+
+    /** @deprecated since 2.8 */
```
StorytellerCZ, line 111: After talking with @Grubba27 we would like you to add more instructions on what to use, maybe `@see createIndexAsync` or something like that. Let's use JSDocs to its full potential! 👍

StorytellerCZ (changes requested):

packages/mongo/mongo.d.ts
StorytellerCZ: I'm fine with using `@see`, I think that is cleaner.

StorytellerCZ (approved):

packages/mongo/mongo.d.ts
```diff
@@ -107,6 +107,11 @@ export namespace Mongo {
       byteSize?: number,
       maxDocuments?: number
     ): Promise<void>;
+
+    /**
+     * @deprecated since 2.8. Check migration guide {@link https://guide.meteor.com/2.8-migration}
+     * @see createIndexAsync
+     */
```
Grubba27, line 114: I would add that is deprecated in the server only.

In the client those methods will still be available

guncebektas: Hmm I thought, those will be deprecated too for the client 😵‍💫 how these functions will work on methods in import folder? This topic should also be covered in the guide.

If we will continue on this pr, some of them are server only methods like createIndex. If you want, I can remove others.

PR #12833, opened 2023-10-17
User index async

StorytellerCZ:
Cherry-pick user index async creations from #12647 

This is to scale down #12647 and get things in faster in more manageable chunks.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1774,21 +1774,21 @@ const setupUsersCollection = users => {
   });
 
   /// DEFAULT INDEXES ON USERS
-  users.createIndex('username', { unique: true, sparse: true });
-  users.createIndex('emails.address', { unique: true, sparse: true });
-  users.createIndex('services.resume.loginTokens.hashedToken',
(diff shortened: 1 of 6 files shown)


Grubba27 (approved):

matheusccastroo:
@Grubba27 @StorytellerCZ Aren't those missing the `await` call?

StorytellerCZ:
No, these are fire and forget. So you just call them and you don't wait for them to resolve because you are not waiting for any return or need them to resolve before you move on to something else.

Grubba27:
Yes, and this was made in v2.14. From what I recall, we do not have TLA in this version.

matheusccastroo:
@Grubba27 @StorytellerCZ Some calls from the changes are inside functions from what I saw: `setupUsersCollection` and `registerService` for example. For those cases, we still should wait for the index to be created to keep the same behavior we had previously, no? Otherwise this behavior change should be documented.

Also, this case for example: https://github.com/meteor/meteor/pull/12833/files#diff-6116731d4df22d1593416163650cd74f1432f35e9e4a178effecf82889c56be7R8 the `try/catch` here won't work as intended  anymore.

Issue #12792, opened 2023-09-26
[Feature] - Add TypeScript variations to `meteor create`


Grubba27:
## Currently 

Today, when using the `meteor create` command, we can only add one flag in the command, making the following command impossible:

```bash
 meteor create vue-app --vue --typescript
```

_The command above generates a react with typescript app._

## Proposal

What should happen is that every command should have a typescript counterpart, and it could be called by either passing those two flags or just `--ts-vue` or `--ts-solid`.

## Todo:

- Create typescript app for every framework(and test if all works ok):
  - [ ] Apollo
  - [ ] Blaze
  - [ ] Vue
  - [ ] Vue2
  - [ ] Solid
  - [ ] Svelte

- [ ] Adjust the CLI to take more arguments and be able to create those apps.

Sathasivamthirumoorthi:
Can i work on this. Please assign it me

Devanshi-Crypto:
Hey is it complete or is it left? If there's work left then can you assign it to me?

StorytellerCZ:
@Devanshi-Crypto submit a PR, no assignment needed.

07tAnYa:
I have started working on this

alimgafar:
So will this be a 2 flag solution or simple one flag that combines two feaures?

StorytellerCZ:
Will #12803 include this?

JaiS99:
Can I work on this?

Grubba27:
> So will this be a 2 flag solution or simple one flag that combines two features?

@alimgafar, the end user could do both, he could pass `--vue` and `--typescript` so it creates a typescript vue project, or if he just passes `--ts-vue` would do the same result.

Dxuian:
is this repo still open for contribution or deprecating  ?

StorytellerCZ:
@Dxuian creating TypeScript variants of the starter skeletons is still needed. Other than that the `meteor create` command is interactive enough.

nikunjmathur08:
@StorytellerCZ is the issue still open? I can work on this.

StorytellerCZ:
@nikunjmathur08 yes

vibh1103:
I can see inactivity can i pick this @StorytellerCZ if possible for hacktoberfest ?

divyeshtechs-prog:
Assign me if there are any tasks left.

jeetburman:
[...]

PR #12836, opened 2023-10-21
Feature - Added TypeScript variations to meteor create for Vue3 Based app with ts-vue skeleton 

nvsai:
The app is built and tested in the local environment. This PR is for one of the Feature mentioned in #12792 .
meteor1
meteor2


nvsai:
Hi @Grubba27 ,I am new to open source.Can you please review this tell me if i can proceed in the same way with other skeletons?

tools/cli/commands.js
```diff
@@ -941,6 +943,7 @@ main.registerCommand({
     cmd("meteor create --vue-2      # to create a basic Vue2-based app");
     cmd("meteor create --apollo     # to create a basic Apollo + React app");
     cmd("meteor create --svelte     # to create a basic Svelte app");
+    cmd("meteor create --ts-vue     # to create an app using TypeScript and basic Vue3-based app");
```
italojs, line 946: I'm noticing an issue with the developer experience (DX) in this line

When generating projects with TypeScript. Right now, the CLI assumes some frameworks have TypeScript support while others do not, leading to inconsistencies in command naming:  

- `meteor create --react` → Creates a JS-based React app  
- `meteor create --typescript` → Creates a TypeScript-based React app  
- `meteor create --vue` → Creates a JS-based Vue app  
- **No `--typescript` equivalent for Vue**, but instead we have `meteor create --ts-vue`  

This creates a problem where:  
1. **Inconsistent Naming Conventions** – Some frameworks (React) use `--typescript`, while others (Vue) require a `--ts-` prefix.  
2. **Scalability Issue** – If we add TypeScript support for other stacks in the future, should we keep using `--ts-*` or introduce `--typescript` variations?  
3. **DX Confusion** – Users may expect a `--typescript` flag to apply to any framework, but currently, it's implicitly tied to React.  
4. **Potential Breaking Changes** – Standardizing the pattern might require renaming `--typescript` to `--ts-react` or `--react-ts`, which could introduce breaking changes or force an awkward backward-compatibility layer.  

We should define a **clear and future-proof standard** before this becomes harder to refactor. Some options:  
- Use `--ts-*` for all TypeScript variants (e.g., `--ts-react`, `--ts-vue`, `--ts-svelte`)  
- Use `--*-ts` for consistency with framework naming (`--react-ts`, `--vue-ts`, `--svelte-ts`)  
- Allow `--typescript` as a **global modifier** (`meteor create --vue --typescript`)  

This discussion likely requires input from the core committers before finalizing, as it impacts DX and `maybe` generates a break change for the CLI design.

italojs (changes requested):

tools/cli/commands.js
italojs: @Grubba27 Thoughts?

italojs:
> Hi @Grubba27 ,I am new to open source.Can you please review this tell me if i can proceed in the same way with other skeletons?

btw your PR looks good, apart from that DX discussion, you are new to the open source but are doing an awesome job! continue on it @nvsai

PR #12839, opened 2023-10-24
Documentation update to make Tracker.withComputation more understandable

sebastianspiller:
Hey Meteorians!

I've read the documentation about Tracker.withComputation and I think that it's not that easy to decipher. To be more clear, I updated the example and removed some parts to make it more concise.

Thanks for the work 😃

### DIFF
--- a/docs/source/api/tracker.md
+++ b/docs/source/api/tracker.md
@@ -81,45 +81,33 @@ is automatically stopped and won't be rerun.
 
 ### Tracker.autorun and async callbacks
 `Tracker.autorun` can accept an `async` callback function.  
-However, to make the async call reactive, you should wrap your async function in
-a `Tracker.withComputation` call.
-
-```javascript
-Tracker.autorun(async function example1(computation) {
-  let asyncData = await asyncDataFunction();
-  let users =
-    await Tracker.withComputation(computation, () => Meteor.users.find({}).fetch());
-});
-```
-> If you want to get computation in other way you can use `Tracker.currentComputation`
-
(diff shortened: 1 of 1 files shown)


docs/source/api/tracker.md
```diff
@@ -81,45 +81,33 @@ is automatically stopped and won't be rerun.
...
-  let users =
-    await Tracker.withComputation(computation, () => Meteor.users.find({}).fetch());
-});
-```
-> If you want to get computation in other way you can use `Tracker.currentComputation`
-
-
-{% apibox "Tracker.withComputation" %}
+	// Code before the first await will stay reactive.
+	reactiveVar1.get() // will trigger rerun
```
radekmie, line 91: Please keep the comment style consistent - add semicolons, start with an upper letter, and end with a dot.

docs/source/api/tracker.md
```diff
@@ -81,45 +81,33 @@ is automatically stopped and won't be rerun.
...
 
-```javascript
-Tracker.autorun(async function (computation) {
-  let asyncData = await someAsyncCall();
-  let users =
-    await Tracker.withComputation(computation, () => Meteor.users.find({}).fetch());
-  // code above will trigger reruns.
-});
-```
+As a rule of thumb, you are fine with wrapping all reactive statements inside a Tracker.withComputation to preserve current computation.
```
radekmie, line 107: * Function names should be wrapped with `` ` ``.
* Yes, it is _fine_, but it comes at a performance cost - it should be used only where needed.

radekmie (approved):

PR #12847, opened 2023-10-28
Update PULL_REQUEST_TEMPLATE.md

Yoda-Flash:
Edited the pull request template to make it more contributor-friendly, adding some more details.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,6 +21,29 @@ Here are some important details to follow:
 * ✏️ Explain your pull request
           Describe the big picture of your changes here to communicate to what your
           pull request is meant to accomplish.  Provide 🔗 links 🔗 to associated issues!
+-->
+#### Description
 
+<!-- Please explain your pull request here. -->
+
+##### Type of change
+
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -21,6 +21,29 @@ Here are some important details to follow:
...
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] This change requires a documentation update
+
+#### How Has This Been Tested?
 
+<!-- Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration -->
+
+**Test Configuration**:
+* Firmware version:
```
StorytellerCZ, line 41: What? This list doesn't fit at all.

StorytellerCZ (changes requested):
@Grubba27 as it is right now it doesn't fit. Also from experience templates like these are often left unfilled or done incorrectly. The list there also often creates confusion when in overview as it is normally used to list things that still need to be done rather than to select, so in overview it then looks `1 of 4 done`.

PR #12848, opened 2023-10-28
Update Testing section for Tinytest

jamauro:

### DIFF
--- a/guide/source/writing-atmosphere-packages.md
+++ b/guide/source/writing-atmosphere-packages.md
@@ -272,7 +272,7 @@ If you are using [Tinytest](https://github.com/meteor/meteor/tree/devel/packages
 meteor test-packages ./
 ```
 
-If you are using Mocha for your package's tests, you'll need to specify a `driver-package`. You'll need to run the following to start the Mocha [test driver package](testing.html#driver-packages):
+If you are using a different testing framework for your package's tests, you'll need to specify a `driver-package`. For example, if you are using Mocha, run the following to start the Mocha [test driver package](testing.html#driver-packages):
 
 ```bash
 meteor test-packages ./ --driver-package meteortesting:mocha
--- a/guide/source/writing-atmosphere-packages.md
+++ b/guide/source/writing-atmosphere-packages.md
@@ -264,13 +264,21 @@ Read more about using Cordova in the [mobile guide](mobile.html).
 
 <h2 id="testing">Testing packages</h2>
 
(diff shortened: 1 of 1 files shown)


jamauro:
I've made a few tweaks to clarify. Thanks for reviewing @StorytellerCZ

guide/source/writing-atmosphere-packages.md
```diff
@@ -264,13 +264,19 @@ Read more about using Cordova in the [mobile guide](mobile.html).
...
+
+If you are using [Tinytest](https://github.com/meteor/meteor/tree/devel/packages/tinytest) for your package's tests, you can run:
 
 ```bash
-meteor test-packages ./ --driver-package meteortesting:mocha
+meteor test-packages ./
 ```
 
-This will run a special app containing only a "test" version of your package and start a Mocha [test driver package](testing.html#driver-packages).
+If you are using Mocha for your package's tests, you'll need to specify a `driver-package`. You'll need to run the following to start the Mocha [test driver package](testing.html#driver-packages):
```
StorytellerCZ, line 275: Maybe instead `If you are using other test runner for your package's tests`
And then it should say that the following is a sample with mocha.

StorytellerCZ (changes requested):

guide/source/writing-atmosphere-packages.md
jamauro: I considered that but I'm honestly not sure what other test runners that Meteor supports out of the box. In the past, I and others had difficulty getting Jest to work for example so I wouldn't want to steer someone down a path where they end up wasting a bunch of time. 

This doc update really is meant to clarify support for Tinytest. The previous version of the docs made it seem like you always needed to use the Mocha test driver which meant your tests wouldn't run at all if you're using Tinytest. 🙂

StorytellerCZ: That might be the case, but the truth is that the driver package allows anyone to make it, so this could have the opposite effect as well making people think that there is only TinyTest and Mocha, while the truth is, it is what ever you use with the driver-package. If you want to include Mocha, then it should use `for example` in the sentence. So maybe something along these lines:
```
If you are using other tests than TinyTests, for example Mocha, for your package's tests, you'll need to specify a `driver-package`. You'll need to run the following to start the Mocha [test driver package](testing.html#driver-packages):
```

jamauro: Ok I made an edit to hopefully make it clearer.

StorytellerCZ (approved):

fredmaiaarantes (approved):

PR #12849, opened 2023-10-30
[service-configuration] Add types for ConfigError

harryadel:
Add types for configError 
https://github.com/meteor/meteor/blob/72f27da680cfdd3ff6fb20f580a7f468cb5f21ce/packages/service-configuration/service_configuration_common.js#L21

I'm not well versed in TS so a little help is appreciated.

### DIFF
--- a/packages/service-configuration/service-configuration.d.ts
+++ b/packages/service-configuration/service-configuration.d.ts
@@ -5,6 +5,12 @@ export interface Configuration {
   secret: string;
 }
 
+export interface ConfigError extends Error {
+  constructor(serviceName?: string);
+  message: string;
+}
+
 export declare var ServiceConfiguration: {
   configurations: Mongo.Collection<Configuration>;
+  ConfigError: ConfigError
 };
--- a/packages/service-configuration/service-configuration.d.ts
+++ b/packages/service-configuration/service-configuration.d.ts
@@ -5,12 +5,12 @@ export interface Configuration {
   secret: string;
 }
 
-export interface ConfigError extends Error {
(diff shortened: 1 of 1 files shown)


harryadel:
cc @piotrpospiech @radekmie @ebroder

packages/service-configuration/service-configuration.d.ts
```diff
@@ -5,6 +5,12 @@ export interface Configuration {
   secret: string;
 }
 
+export interface ConfigError extends Error {
+  constructor(serviceName?: string);
+  message: string;
+}
+
```
ebroder, line 12: Hmm. This should probably be declared as a class, not as an interface

harryadel: Thanks, will apply that change. But a side question, whenever I make changes in TS definition file vscode isn't picking up on them, any idea why?

StorytellerCZ (approved):

packages/service-configuration/service-configuration.d.ts
```diff
@@ -5,6 +5,12 @@ export interface Configuration {
...
 }
 
+class ConfigError extends Error {
+  constructor(serviceName?: string);
+  message: string;
+}
+
 export declare var ServiceConfiguration: {
   configurations: Mongo.Collection<Configuration>;
+  ConfigError
```
radekmie, line 15: Sorry for being late, but this is incorrect: with this syntax, `ConfigError` is `any`. It should say `ConfigError: typeof ConfigError` (without the `typeof` it'd be an instance and not the class itself).

PR #12852, opened 2023-10-30
Add NonEmptyString to check package

harryadel:
plz hlp revieu! @StorytellerCZ

### DIFF
--- a/packages/check/check.d.ts
+++ b/packages/check/check.d.ts
@@ -90,3 +90,15 @@ export declare function check<T extends Match.Pattern>(
   value: any,
   pattern: T
 ): asserts value is Match.PatternMatch<T>;
+
+
+/**
+ * Check that a value matches a pattern.
+ * If the value does not match the pattern, throw a `Match.Error`.
+ *
+ * Particularly useful to assert that arguments to a function have the right
+ * types and structure.
+ * @param value The value to check
+ * @param pattern The pattern to match `value` against
+ */
+export declare function NonEmptyString<T extends Match.Where>(x: any): boolean;
--- a/packages/check/match.js
+++ b/packages/check/match.js
@@ -192,6 +192,17 @@ const stringForErrorMessage = (value, options = {}) => {
   return EJSON.stringify(value);
 };
 
+/**
+ * @summary Returns true if the value is a non empty string.
+ * @locus Anywhere
+ * @param {Any} value The value to check
+ */
+
+export const NonEmptyString = Match.Where((x) => {
+  check(x, String);
+  return x.length > 0;
+});
+
 const typeofChecks = [
   [String, 'string'],
   [Number, 'number'],
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1,18 +1,13 @@
 import crypto from 'crypto';
+import { URL } from 'url';
 import { Meteor } from 'meteor/meteor'
+import { check, Match, NonEmptyString } from 'meteor/check';
 import {
   AccountsCommon,
   EXPIRE_TOKENS_INTERVAL_MS,
 } from './accounts_common.js';
-import { URL } from 'meteor/url';
 
 const hasOwn = Object.prototype.hasOwnProperty;
-
-// XXX maybe this belongs in the check package
-const NonEmptyString = Match.Where(x => {
-  check(x, String);
-  return x.length > 0;
-});
 
 /**
  * @summary Constructor for the `Accounts` namespace on the server.
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -1,5 +1,6 @@
-import { hash as bcryptHash, compare as bcryptCompare } from 'bcrypt';
 import { Accounts } from "meteor/accounts-base";
+import { check, Match, NonEmptyString } from 'meteor/check';
+import { hash as bcryptHash, compare as bcryptCompare } from 'bcrypt';
 
 // Utility for grabbing user
 const getUserById = (id, options) => Meteor.users.findOne(id, Accounts._addDefaultFieldSelector(options));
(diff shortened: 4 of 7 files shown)


StorytellerCZ (dismissed):

bittusakshi (commented):
export const NonEmptyString declares a constant named NonEmptyString that will be exported for use in other parts of your code.

Match.Where() is a Meteor function that is used to create custom validation functions. In this case, it's creating a custom validation function for checking if a value is a non-empty string.

harryadel:
> export const NonEmptyString declares a constant named NonEmptyString that will be exported for use in other parts of your code.
> 
> Match.Where() is a Meteor function that is used to create custom validation functions. In this case, it's creating a custom validation function for checking if a value is a non-empty string.

@bittusakshi What's the problem here?

packages/check/check.d.ts
```diff
@@ -90,3 +90,15 @@ export declare function check<T extends Match.Pattern>(
...
+/**
+ * Check that a value matches a pattern.
+ * If the value does not match the pattern, throw a `Match.Error`.
+ *
+ * Particularly useful to assert that arguments to a function have the right
+ * types and structure.
+ * @param value The value to check
+ * @param pattern The pattern to match `value` against
+ */
+export declare function NonEmptyString<T extends Match.Where>(x: any): boolean;
```
radekmie, line 104: It should be a `Matcher<string>` instead, so the `check` would infer the type correctly.

packages/check/match.js
```diff
@@ -192,6 +192,17 @@ const stringForErrorMessage = (value, options = {}) => {
...
+/**
+ * @summary Returns true if the value is a non empty string.
+ * @locus Anywhere
+ * @param {Any} value The value to check
+ */
+
+export const NonEmptyString = Match.Where((x) => {
+  check(x, String);
+  return x.length > 0;
+});
```
radekmie, line 204: I'd like to hear what others think, but I don't see why it's a separate export and not a `Match` property, just like `Maybe` or `Where`.

StorytellerCZ: Good point, but then I think it would be better to transform this into general `NonEmpty` function where you could put in type:
```
check(userId, Match.NonEmpty(String))
```

radekmie: Good idea! But it should have a list of allowed types, since `Match.NonEmpty(CustomType)` would require additional implementation in the `CustomType`.

packages/check/check.d.ts
```diff
@@ -90,3 +91,14 @@ export declare function check<T extends Match.Pattern>(
...
+
+/**
+ * Check that a value matches a pattern.
+ * If the value does not match the pattern, throw a `Match.Error`.
+ *
+ * Particularly useful to assert that arguments to a function have the right
+ * types and structure.
+ * @param value The value to check
+ * @param pattern The pattern to match `value` against
+ */
```
radekmie, line 104: This should be moved up, to the `var` definition.

packages/check/match.js
```diff
@@ -272,6 +275,17 @@ const testSubtree = (value, pattern) => {
...
+  // We invoke the pattern as IIFE so that `pattern isntanceof Where` catches it 
+  if (pattern === Match.NonEmptyString) {
+    pattern = (function() {
+      const condition = (val) => {
+         check(val, String);
+         return val.length > 0;
+      }
+      return new Where(condition);
+    })();
+  }
```
radekmie, line 288: 1. That's exactly what `Symbol`s are for, though I'm not sure whether it wouldn't increase minimum required browser version.
2. It'd be best not to create this helper function and `Where` instance every time, but create it on a top-level and only use here.

packages/check/match.js
```diff
@@ -192,6 +194,7 @@ const stringForErrorMessage = (value, options = {}) => {
   return EJSON.stringify(value);
 };
 
+
```
radekmie, line 197: (side note) What's the deal with those empty lines here and in the next file? Am I missing something?

harryadel:
@radekmie Please recheck.

packages/check/match.js
radekmie: The new code still creates new closure and `Where` instance every time.

harryadel: How do you propose to fix it?

radekmie:
```ts
function nonEmptyStringCondition(value) {
  check(value, String);
  return value.length > 0;
}

const NonEmptyString = new Where(nonEmptyStringCondition);
```

radekmie: How do I put it 😅 ...
The last version you pushed reuses the `nonEmptyStringCondition`, sure. But it creates a new `Where` instance each time. I proposed to reuse the `Where` instance too.

radekmie (approved):

packages/check/match.js
harryadel: sawry tfw no english

harryadel:
@leonardoventurini

StorytellerCZ (approved):
LGTM, would be nice to also add this into the docs.

harryadel:
cc @italojs @nachocodoner @Grubba27

nachocodoner:
This is good addition. Lets add it on next minor Meteor 3.4 release.

Please, update the docs to reflect the new option.

StorytellerCZ (approved):

nachocodoner:
I'm about to review the PRs for the upcoming Meteor 3.4 releases listed at https://github.com/meteor/meteor/milestone/122

They will be included in 3.4 only if they pass all PR checks. Getting them green soon could allow a beta release earlier. Let's see if re-running the tests fixes the failures here. Otherwise, more work will be needed.

Also, docs update seems missing.

harryadel:
@nachocodoner I'll look into it soon and get it ready :+1:

harryadel:
@nachocodoner Can this be included in https://github.com/meteor/meteor/pull/13915? pweaaaase

nachocodoner:
I will include this in the first release candidate, since it was originally planned for 3.4 but was postponed because it didn’t make it in time for the last beta.

After the release candidate, I will only push critical or quick fixes. Everything else should go to the 3.4.x series or later.

PR #12856, opened 2023-10-31
Don't flatten ObjectIDs in oplogV2V1Converter (fixes #12855)

### LINKED ISSUE #12855: ObjectIDs get broken in Minimongo when $set, since Mongo 5 update
Hi meteor,

When setting a Mongo.ObjectID in a document, the ObjectID gets transformed into a plain object in Minimongo, until the page is refreshed / the document is refetched.

This causes severe client-side issues, as the id cannot be used to find the document (conversely, find other documents the id is supposed to reference).

Meteor version: Since 2.6.0, namely since the mongod 4.4 -> 5.0 update in d68b30f3 (dev-bundle-14.18.3.3)

Reproduction:
```
$ meteor create objectid-bug && cd objectid-bug && meteor

// Add the following line in imports/api/links.js:
LinksCollection.allow({ insert: () => true, update: () => true })

// In the browser console: 
const { LinksCollection } = require('/imports/api/links.js')
const linkId = LinksCollection.insert({ aaa: new Mongo.ObjectID() });
LinksCollection.update(linkId, { $set: { xxx: new Mongo.ObjectID() } })
setTimeout(() => console.log(LinksCollection.findOne(linkId)), 100)
```
=> In the console.log output, `xxx` is `{_str: '...'}` instead of `ObjectID {_str: '...'}` (unlike `aaa` which is fine).

The issue is also visible in the DDP messages. v2.6:
Screenshot 2023-10-27 at 16 44 56
whereas in v2.5.x, the ObjectID was correctly encoded (`$type: "oid"`):
Screenshot 2023-10-27 at 16 39 06

There are no issue server-side, as far as I've seen.

We ran into this problem right after updating our app past v2.5, so I'm surprised it would have gone undetected for so long? 

Ultimately, this seems to come from a bug in `oplogV2V1Converter`. I will create a PR.

dd137:
Attempts to fix #12855: When setting a Mongo.ObjectID in a document, the ObjectID gets transformed into a plain object in Minimongo, until the page is refreshed / the document is refetched.

Includes the fix and a matching test.

Thanks for any feedback.

### DIFF
--- a/packages/mongo/oplog_v2_converter.js
+++ b/packages/mongo/oplog_v2_converter.js
@@ -47,7 +47,8 @@ function isArrayOperator(operator) {
 }
 
 function flattenObjectInto(target, source, prefix) {
-  if (Array.isArray(source) || typeof source !== 'object' || source === null) {
+  if (Array.isArray(source) || typeof source !== 'object' || source === null ||
+      source instanceof Mongo.ObjectID) {
     target[prefix] = source;
   } else {
     const entries = Object.entries(source);
--- a/packages/mongo/oplog_v2_converter_tests.js
+++ b/packages/mongo/oplog_v2_converter_tests.js
@@ -78,6 +78,10 @@ const cases = [
     { $v: 2, $set: { params: { e: { _str: '5f953cde8ceca90030bdb86f' } } } },
   ],
   [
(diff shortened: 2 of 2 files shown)


radekmie (approved):

StorytellerCZ:
Will be released in 2.14

PR #12860, opened 2023-10-31
Allow some old build plugins to run in Meteor 3

zodern:
Build plugins include a copy of all of their deps, such as the `meteor` package, when they are published. The Meteor package require's fibers, but only uses it for certain api's that are not normally used in build plugins. This PR makes the require for `fibers` from the Meteor package work, so these build plugins can still be used in Meteor 3 as long as they don't directly or indirectly use fibers.

I've tested this with zodern:types and zodern:meteor-package-versions. Any build plugin that uses async/await will still be broken. Build plugins that use `promise` or `ecmascript` might also be broken, but I think we could fix it in a similar way.


StorytellerCZ (approved):

radekmie (approved):

filipenevola (approved):
We should merge it sooner than later. I got issues because of that when trying to run with Meteor 3 and `zodern:types` package

PR #12832, opened 2023-10-17
Dev bundle 14.21.4.4

StorytellerCZ:
Continuing from #12622 and merging in #12783 and relevant parts of #12669.

* Update bundle dependencies
* Move TS types to devDependencies
* Update chalk
* Typescript 4.9.5

TODO:
- [x] Publish `@meteorjs/babel`

PR #12871, opened 2023-11-09
Bump dev bundle to 14.21.4.6

Grubba27:
Brought all from https://github.com/meteor/meteor/pull/12832 except the Typescript update that I was not able to update :/ Something is blocking us from upgrading. Maybe something with Babel package but I was not sure and I did not want to be blocked on this/ want to move our release forward.


.github/workflows/npm-eslint-plugin-meteor.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         working-directory: npm-packages/eslint-plugin-meteor
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x, 18.x]
```
StorytellerCZ, line 21: Given that Meteor 3 will be on Node 20, we need to update these accordingly.

StorytellerCZ (approved):

.github/workflows/npm-eslint-plugin-meteor.yml
Grubba27: Agree! I will create a PR with 14.21.4.7 with those change as well

PR #12872, opened 2023-11-09
Dev bundle 14.21.4.7

Grubba27:
Reverted `source-map-support` lib as it was before in 2.14 and made https://github.com/meteor/meteor/pull/12832 work.


Grubba27:
I think the ci will pass, but the `npm babel` one will not. Is missing an update in package-lock.json I was not able to do it, when I run `npm i` I get an error while generating the `package-lock`

Grubba27:
[sc-3703]

Grubba27:
@StorytellerCZ actually the `meteor-babel` uses [fibers](https://github.com/meteor/meteor/blob/cc8039279b92029aa02df9137c3ddcdbd7b63c98/npm-packages/meteor-babel/package.json#L57) I think we should keep as it was before, while there is still fibers

StorytellerCZ (approved):

PR #12886, opened 2023-11-16
Update semver to 7.5.4

Grubba27:
This comes from: https://security.snyk.io/package/npm/semver

Todo: 
- [x] Create new devbundle to make this changes happen.

It was already published the `npm meteor/babel`


StorytellerCZ:
@Grubba27 once this is merged I think that would be a good point to do another beta release.

Issue #12888, opened 2023-11-17
[Meteor 3] Meteor.callAsync doesn't provide result from server


zodern:
To reproduce:
In a new app from Meteor 3.0-alpha.18, add this code to both the client and server:

```js
if (Meteor.isServer) {
  Meteor.methods({
    'server-only'() {
      return 'result'
    }
  });
} else {
  Meteor.callAsync('server-only').then(console.log);
}
```

Instead of logging `result`, it logs:
```
{stubValuePromise: Promise}
```

paulincai:
A related conversation: https://forums.meteor.com/t/meteor-3-0-alpha-15-the-return-from-methods/60709

PR #12897, opened 2023-11-21
Add zodern solution for Async stubs

Grubba27:
Relates to this forum [post](https://forums.meteor.com/t/proposal-to-fix-issues-with-async-method-stubs/60826/21?u=grubba)

Created this PR to test the code against our CI

### DIFF
--- a/packages/ddp-client/client/client_convenience.js
+++ b/packages/ddp-client/client/client_convenience.js
@@ -214,6 +214,35 @@ function queueMethodInvoker(methodInvoker, wait) {
   });
 }
 
+/**
+ * Queue subscriptions in case they rely on previous method calls
+ */
+let queueSend = false;
+let oldSubscribe = Meteor.connection.subscribe;
+Meteor.connection.subscribe = function () {
+  queueSend = true;
+  try {
+    return oldSubscribe.apply(this, arguments);
+  } finally {
+    queueSend = false;
+  }
+};
+
+let oldSend = Meteor.connection._send;
+Meteor.connection._send = function () {
+  if (!queueSend) {
+    return oldSend.apply(this, arguments);
+  }
+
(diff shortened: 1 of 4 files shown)


zodern:
The queue was causing methods and subscriptions to run out of order. I pushed a commit that also queues subscriptions to maintain the order.

Meteor.isAsyncCall is broken. I'm not sure what the purpose of the function is, and how it should behave with the queue's. 
1. Currently it is only true when using Meteor.callAsync. Should it also be true when using Meteor.applyAsync?
2. Currently it is true from when calling Meteor.callAsync until the server result is returned. With the queue's, do you expect it to behave the same (return true if there is at least one async call in the queue or waiting for the server result), or should it only be true while the stub is running or when waiting for the server result?

The tests copied over from the zodern:fix-async-stubs package won't pass until https://github.com/meteor/meteor/issues/12888 is fixed.

zodern:
It seems all of the tests are now passing except for the ones copied from zodern:fix-async-stubs (which requires #12888 to be fixed first).

Grubba27:
In your proposal @zodern, if the `stubPromise` is undefined, we would return it or return the server result?

zodern:
> In your proposal @zodern, if the stubPromise is undefined, we would return it or return the server result?

Meteor.callAsync would always return the server promise, with an additional property for the stub promise. I would assume the stub promise would always be defined (and resolve to undefined for methods without a stub) but it doesn't have to work that way.

Grubba27:
I have [drafted](https://github.com/meteor/meteor/pull/12907)(still working on missing tests) your proposal @zodern(server-first), I got a little bit confused while dealing with the flags for the stubs-first proposal.

denihs (approved):

PR #12907, opened 2023-11-29
[Meteor 3] (server-first) Solving Meteor call Async stubValuePromise issue

Grubba27:
As discussed in https://github.com/meteor/meteor/issues/12888 and https://github.com/meteor/meteor/pull/12897 we need to come to a decision in this API: 

For this PR we have implemented the comment [here](https://github.com/meteor/meteor/pull/12897#issuecomment-1832494606):

```js
const promise = Meteor.callAsync(...);

const resultFromStub = await promise.stub;
const resultFromServer = await promise;

// backwards compatible with existing code:
const resultFromServer = await Meteor.callAsync(...);
```


packages/ddp-client/test/livedata_tests.js
```diff
@@ -414,11 +414,7 @@ testAsyncMulti('livedata - compound methods', [
       );
     } catch (e) {
     }
-    if (Meteor.isClient)
-      // client can fool itself by cheating, but only until the sync
-      // finishes
-      await checkBalances(test, -10, 160);
-    else await checkBalances(test, 90, 60);
```
Grubba27, line 421: @zodern maybe from this test, the api should be stub first, then the server result? because now it needs to await for the server result.
Not sure if this is correct

jamauro:
Will this solution enable getting the stub result before the server promise resolves? It's not clear to me if `await promise.stub` is awaiting the promise and then grabbing the stub or if it's somehow awaiting a stubPromise.  

If not, is there a reason why something like [this approach](https://svelte.dev/repl/0fbc4c5236654352bc8a1b0c5be47920?version=3.59.1) wouldn't work?

Grubba27:
> If not, is there a reason why something like [this approach](https://svelte.dev/repl/0fbc4c5236654352bc8a1b0c5be47920?version=3.59.1) wouldn't work?

It could work... this is more of a discussion of what the API should look like, and we could have both, as you suggested.

This PR focuses on being server-first. 

because it defaults to be the server result, and now we are async. That test needed to be changed(not sure if I like this).


Not sure if I understood you question @jamauro, with this proposal we can do this.

```js
const { stub } = Meteor.callAsync(...);
const resultFromStub = await stub;
```

And while coding this example I realize that makes sense as well to have server result as well in this object.

packages/ddp-client/test/livedata_tests.js
zodern: If you look at the old code for the test: 
https://github.com/meteor/meteor/blob/e0763170e98d05b48b00ec47d5a03c8f2a00c09e/packages/ddp-client/test/livedata_tests.js#L340-L398

You could maybe implement it for Meteor 3 like this:

```js
let promise = Meteor.callAsync(
  'ledger/transfer',
  test.runId(),
  'alice',
  'bob',
  100,
  true,
);

if (Meteor.isClient) {
  // client can fool itself by cheating, but only until the sync
  // finishes
  await promise.stub
  await checkBalances(test, -10, 160);
}

await promise.catch(err => {
  failure(test, 409)(err, result);
});

await checkBalances(test, 90, 60);

```

jamauro:
Ok. I trust that y'all will come up with the best solution. 🙂

I mainly wanted to ensure that I could use the stub result without waiting on the server and then if the server errors, I could rollback the action I took with the stub result. I don't do this often but it can be kind of nice when working with a newly inserted id.

packages/ddp-client/test/livedata_tests.js
Grubba27: I think I made a mistake somewhere in here https://github.com/meteor/meteor/blob/fd2150c4c2d41a6a625b54cf6bad41598f8a6176/packages/ddp-client/common/livedata_connection.js#L891-L907

The test still does not pass

Grubba27:
@jamauro I've added your idea. You can also do this:

```js
const { stub, serverResult }= Meteor.callAsync(...);
const whatServerDid = await serverResult;
const whatStubDid = await stub;

// or just

const whatServerDid2 = await Meteor.callAsync(...)
```

packages/ddp-client/test/livedata_tests.js
jamauro: I took at look at this over the weekend. Unfortunately wasn't able to make the test pass. I noticed that `checkBalances` on the client is running before the stub promise has completed. When I added a timeout of 1, prior to `checkBalances` it passed. 

```js
if (Meteor.isClient) {
  // client can fool itself by cheating, but only until the sync
  // finishes
  await promise.stub // this isn't being awaited
  // if i add -- await sleep(1) -- it passes
  await checkBalances(test, -10, 160);
}
```

Hope this can be of help. Maybe someone closer to the codebase will be able to instantly tell what the issue is. 🙂 If I get some time this week, I will take a closer look.

packages/ddp-client/common/livedata_connection.js
```diff
@@ -890,14 +890,18 @@ export class Connection {
     // If we're using the default callback on the server,
     // block waiting for the result.
     if (future) {
+      future.stub = stubReturnValue;
+      future.serverResult = future;
```
zodern, line 894: I would suggest renaming these to `stubPromise` and `serverPromise` for consistency, and to make it clear what the properties contain.

jamauro: As far as I can tell, `stubReturnValue` is a value and not a promise that is passed into `_apply` from `applyAsync`. I think the code needs to be changed upstream of `future.stubPromise = stubReturnValue` in order for things to work as expected. I could be wrong though. There's a lot going on in this code :) 

https://github.com/meteor/meteor/blob/fd2150c4c2d41a6a625b54cf6bad41598f8a6176/packages/ddp-client/common/livedata_connection.js#L721

Also it's not clear to me why `stubInvocation` is being awaited here when it is not a promise:
https://github.com/meteor/meteor/blob/fd2150c4c2d41a6a625b54cf6bad41598f8a6176/packages/ddp-client/common/livedata_connection.js#L972-L982

rj-david:
> @jamauro I've added your idea. You can also do this:
> 
> ```js
> const { stub, serverResult }= Meteor.callAsync(...);
> const whatServerDid = await serverResult;
> const whatStubDid = await stub;
> 
> // or just
> 
> const whatServerDid2 = await Meteor.callAsync(...)
> ```

This is stretching my developer mind. Every time I see this, I feel something is wrong.

Just a quick litmus test: Does the reason for having two interfaces for the result of `Meteor.callAsync()` outweigh all the hardships of explaining why there are two interfaces?

packages/ddp-client/client/client_convenience.js
```diff
@@ -42,6 +43,9 @@ Meteor.connection = DDP.connect(ddpUrl, {
   onDDPVersionNegotiationFailure: onDDPVersionNegotiationFailure
 });
 
+// Makes sure Meteor.connect is already created before injecting the helpers
+loadAsyncStubHelpers();
```
zodern, line 47: I'm not very familiar with this part of Meteor. Should these also be used for DDP clients created on the server? I would expect method stubs run on the server to have at least some of the same problems.

packages/ddp-client/client/client_convenience.js
```diff
@@ -42,6 +43,9 @@ Meteor.connection = DDP.connect(ddpUrl, {
   onDDPVersionNegotiationFailure: onDDPVersionNegotiationFailure
 });
 
+// Makes sure Meteor.connect is already created before injecting the helpers
```
zodern, line 46: If the plan is to wrap the `Connection` methods instead of integrating the queue into the `Connection` class, it would probably be better to instead wrap the methods on Connection.prototype and do it before creating `Meteor.connection`. This would ensure that it works the same way for every DDP client, including `Meteor.connection`.

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -0,0 +1,226 @@
...
+      resolve();
+    });
+  }
+
+  /**
+   * Queue subscriptions in case they rely on previous method calls
+   */
+  let queueSend = false;
+  let oldSubscribe = Meteor.connection.subscribe;
+  Meteor.connection.subscribe = function () {
```
zodern, line 202: A test for the Meteor.subscribe wrapper is missing (there was one in the original code).

packages/ddp-client/test/async_stubs/client.js
```diff
@@ -0,0 +1,301 @@
...
+    [
+      "start async-stub",
+      "end async-stub",
+      "start async-stub",
+      "end async-stub",
+    ]
+  );
+  test.equal(
+    sortArray(results),
+    sortArray(["async-server-result", "async-server-result"])
```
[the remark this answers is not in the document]
denihs: you're right

packages/ddp-client/client/client_convenience.js
denihs: I think we only need it on the server. Let's keep this way for now and change it in the future if needed.

packages/ddp-client/client/client_convenience.js
denihs: That sounds good. I'll implement it.

packages/ddp-client/client/queueStubsHelpers.js
denihs: Couldn't find this test, can you send it here, please?

zodern: Sorry, I forgot to push it to github. Here is the commit: https://github.com/zodern/fix-async-stubs/commit/fd6519351f6f0c386cd5ec18d43b26c1aadf7678

PR #12910, opened 2023-12-04
Add a few Q&A about meteor 3

Grubba27:
This PR tries to adress a few Q&A about Meteor 3, and we are trying to make this document as updated as possible and a single source of truth about Meteor 3.

We will update it as we answer questions in forums and other places. Please open a PR with doubts about meteor 3, and we will try creating a PR answering these questions.

## Image with the new listing core packages page
Screenshot 2023-12-05 at 19 24 08


guide/source/3.0-migration.md
```diff
@@ -18,6 +18,95 @@ You can follow the guide "[How to migrate to Meteor Async in Meteor 2.x](/prepar
...
+
+// v3.0 in the client side, this will work anywhere
+
+const docs = await MyCollection.find({ _id: '123' }).fetchAsync();
+
+```
+
+### Will MongoDB Collection Methods be removed from the server?
+
+Yes, we will remove those MongoDB collection methods that do not end with `*Async`.
```
jamauro, line 71: Is this new? I thought the plan was to keep them but make them asynchronous. Since the plan is to eventually drop the `*Async` suffix, maybe it's better to keep them? I'm sure y'all have discussed this a lot :)

Either way, beyond the 3.0 release, I think it's important for Meteor to align with the official Node driver sooner rather than later. Otherwise the longer the `*Async` suffix stays, the more confusion. I believe a Fiber-less Meteor will revitalize the platform but I think having an API that differs from the official Node driver and the Mongo docs, even though it's just a few-extra-characters suffix, will hamper adoption.

I think everyone would benefit with a plan for this end state, e.g. 

```
The '*Async' suffix was necessary to make migrating from Fibers to async / await easier. 

In order to align with Mongo and the official Node driver, 3.X planned for Q# 
will remove the '*Async' suffix and introduce a '*Sync' suffix on the client.
```

Grubba27: Thanks a lot for your input!

This was a decision to make the migration clearer. I have even made a [PR](https://github.com/meteor/meteor/pull/12802/files) to make the possible errors that the end developer might receive so it knows what is happening.

I agree as well with being aligned with the Node API. We have not yet discussed much about what we will change beyond 3.0.

PR #12911, opened 2023-12-05
Update logging package to chalk 5

StorytellerCZ:
Update logging package dependencies to latest.

The v5 gets rid of external dependencies for chalk, so it reduces the package size by half.

With that I think we will be able to get rid of `ecmascript-runtime-client` and `@babel/runtime` dependencies here. I will experiment with that once the initial tests pass.

Also once #12669 is merged it will include the types as well.

### DIFF
--- a/packages/logging/package.js
+++ b/packages/logging/package.js
@@ -4,8 +4,8 @@ Package.describe({
 });
 
 Npm.depends({
-  'chalk': '4.1.1',
-  '@babel/runtime': '7.20.7'
+  'chalk': '5.3.0',
+  '@babel/runtime': '7.23.5'
 });
 
 Npm.strip({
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
I'm not sure if the error is related to the update.
There is also an option to update chalk in the bundle. @nachocodoner should I try adding that here or in a separate PR?

nachocodoner:
Checking the error in this PR, I see issues related to logging, indicating that the bump introduced breaking changes that also need to be addressed.

StorytellerCZ:
I'll look into it when I get some time.

italojs:
@StorytellerCZ, will you still work on it?

StorytellerCZ:
@italojs 
Yes and no. Most recently I have been looking into an approach that removes this dependency all together (and use `node:console` for colors). Need to check if I have a PR for it already or not as I have run into an issue that might just be specific to my OS setup. Either way I'm closing it in favor of that approach.

PR #12798, opened 2023-10-01
Update underscore to 1.9.2

StorytellerCZ:
Try to fix #12673 by updating underscore to version 1.9.2, which is the latest release that uses the structure that we have and hence it is possible to update.
After looking at the changes I have also removed Meteor alterations as I don't think they are no longer valid (but will have to do some testing in regards to #594 to confirm).

PR #12912, opened 2023-12-05
Underscore 1.6

StorytellerCZ:
A test related to #12798 to see if the tests can manage the first update step.


packages/underscore/package.js
```diff
@@ -1,14 +1,14 @@
 
 Package.describe({
   summary: "Collection of small helpers: _.map, _.each, ...",
-  version: '1.0.13',
+  version: '1.6.0',
```
Grubba27, line 4: I do not feel so good doing this huge bump since we use it everywhere.

Maybe we could drop this along the release with DDP resumption 2.15

StorytellerCZ: For 2.15 I was hoping to jump to the original plan of 1.9, I did this 1.6 as that one is the latest closest to what we have been using (and I kept in the changes that were made by MDG, after this version there doesn't seem to be a need for them anymore, but the tests disagree so I will have to look more into it).

denihs (approved):

StorytellerCZ:
@goyan I'll see what I can do to update that function. I don't think I will be able to update the whole thing as, if I remember correctly, the underscore structure changed after 1.6 so I can't make a simple upgrade.

PR #12915, opened 2023-12-06
Docs/how to update packages to meteor 3

Grubba27:
This is based on XP while updating some packages for [simpletasks](https://github.com/fredmaiaarantes/simpletasks) app.


zodern:
It might be good to document the `Meteor.isFibersDisabled` property that packages can use to check if fibers are available. It's falsey (undefined) in Meteor 1/2 and true in Meteor 3.

jamauro:
> It might be good to document the `Meteor.isFibersDisabled` property that packages can use to check if fibers are available. It's falsey (undefined) in Meteor 1/2 and true in Meteor 3. 

Oh nice. I didn't know this was available so I split apart   `Meteor.release` to figure it out.

fredmaiaarantes:
> > It might be good to document the `Meteor.isFibersDisabled` property that packages can use to check if fibers are available. It's falsey (undefined) in Meteor 1/2 and true in Meteor 3.
> 
> Oh nice. I didn't know this was available so I split apart `Meteor.release` to figure it out. I suppose `Meteor.isFibersDisabled` will be sticking around for the foreseeable future?

Yes, `Meteor.isFibersDisabled` should stay in Meteor 3.0 and beyond. I think that's how you can have tests running for both versions in your `jam:easy-schema` package. Which is a pretty good package, btw.

guide/source/3.0-migration.md
```diff
@@ -41,3 +41,29 @@ The best way to follow the progress is by checking the "[What's left until an of
...
+  api.versionsFrom(['1.10', '2.3', '3.0']);
+  //                              ^^^^^^^ for meteor 3.0
+
+```
+Then you can publish your package and test it with Meteor 3.0.
+
+If in your package you are using Meteor packages that will become async,
+you will need to migrate your package to use `async` & `await` pattern.
+
+You can follow a more in depth guide on how to migrate your package to be compatible with Meteor 3.0 [here](/prepare-meteor-3.0#Changes-for-packages).
```
fredmaiaarantes, line 69: Could you add some examples of migrated packages? For instance, `quave:migrations` and `percolate:synced-cron` that it is in progress.

jamauro:
> Yes, `Meteor.isFibersDisabled` should stay in Meteor 3.0 and beyond. I think that's how you can have tests running for both versions in your `jam:easy-schema` package. Which is a pretty good package, btw.

Ok cool. Thanks for the kind words!

guide/source/3.0-migration.md
Grubba27: Sure!

guide/source/prepare-meteor-3.0.md
```diff
@@ -116,6 +116,80 @@ Note that we're not using the if (loading) anymore. To see a practical project,
...
+```
+
+If you want to an example of this change you can take a look in this [commit](https://github.com/meteor/react-packages/commit/96313a1afcc41ef9a23c7496470b375e7d357793)
+where we made possible to a package be used in Meteor 3.0.
+
+This change makes sure that your package is still compatible with Meteor 2.x
+and also with Meteor 3.0.
+
+
+### Changes for packages that are do not using Meteor packages that had breaking changes
```
denihs, line 160: `Changes for packages that do not use Meteor packages that had breaking changes`

PR #12926, opened 2023-12-21
[ Meteor 3 ] - API Docs

Grubba27:
This is the new docs that we [said](https://arc.net/l/quote/hayjwwor) that we are working on. 

This includes the revision of our API docs and a site for guiding current meteor users to the 3.0 version.

For a preview you can check [here](https://v3-meteor-api-docs.netlify.app)


Grubba27:
just a quick update, guys:
The API docs are nearly ready*. I’ve updated it from our changelog, where I found that it was misleading/wrong.
Also, I’ve all our links from the docs. Some of them were redirected to nothing.
*: The [API map](http://localhost:5173/api/) page has a few redirects that are not working, but this is the next thing I should be solving, and then we will have new docs
For those who want to test it, here is the link: https://v3-docs.meteor.com/

denihs (approved):

PR #12927, opened 2023-12-21
Move compileForShell from ecmascript to babel-compiler

zodern:
The babel-compiler package is 86mb. Since it is a runtime dependency of `ecmascript`, a copy of it is included in every app, build plugin, or meteor-tool version that uses `ecmascript`. A few years ago, I found that about half of the 10gb of disk space on my computer in ~/.meteor/packages was just copies of `babel-compiler`, and the multiple copies of it in each release contributed to the slow install times on Windows.

We've been needing to fix it for a while, and it seems Meteor 3 is a good time to do it.

The only reason babel-compiler is a runtime dependency of ecmascript is for `Ecmascript.compileForShell`, used by packages that implement the Meteor shell. This PR moves `compileForShell` to the `babel-compiler` package. This is a breaking change. It should only affect the small number of packages that implement a Meteor shell, and I think the benefits are large enough to make it worth while.

I left the `Ecmascript.compileForShell` function and have it throw an error with the migration instructions. If this PR is accepted, I can create a PR for Meteor 2 that adds `Babel.compileForShell` and adds a deprecation message.

I published `zodern:remove-client-js@1.0.1-beta.5` using these changes. This dropped the size of the package from 99.1mb to 4.9mb. In addition to using less space, it is significantly faster for Meteor to build it in development and for installing a published version.

### DIFF
--- a/docs/generators/changelog/versions/3.0.md
+++ b/docs/generators/changelog/versions/3.0.md
@@ -109,6 +109,7 @@
 
 - `babel-compiler@8.0.0`:
 
+  - Add `Babel.compileForShell`
   - Removed `Promise.await` default transform.
   - Added top-level-await to packages.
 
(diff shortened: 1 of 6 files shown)


Grubba27 (approved):

Grubba27:
If the tests pass I'm 100% in it

Grubba27:
Could you mind documenting these changes @zodern?
Here would be nice `docs/generators/changelog/versions/3.0.md`

denihs (approved):

PR #12931, opened 2023-12-21
Better error message when trying to update to a specific package version

zodern:
I always forget which command to run to update a package to a specific version. When trying to do this with `meteor update`, this PR adds a better error message that suggests using `meteor add` instead.


Grubba27 (approved):

wreiske:
Seeing this PR gave me an idea to include URLs in some of the error outputs linking back to the meteor documentation... For example this PR could easily link to https://guide.meteor.com/using-atmosphere-packages#installing-atmosphere

Grubba27:
Exactly @wreiske! We started doing this, I think if you do a cross boundary import(import  server file in the client), it will trigger an error that has a link to the docs.

StorytellerCZ (approved):

nachocodoner (approved):

PR #12934, opened 2023-12-23
Re-enable TypeScript self-test

ebroder:
I'd like to tentatively try to re-enable this test, so that we can try and get a bit more stability in the TypeScript declarations. I am, admittedly, a little confused as to why I didn't run into https://github.com/zodern/meteor-types/issues/10, but this did work in my local testing.

I think, for the record, that until we're a little more confident in the stability of these tests, it's fine to disable the test again if it runs into issues in CI (although I'd appreciate not wholesale reverting the PR, as the type changes should be correct independent of the test! 🙂)

From the commit message:

> The latest release of zodern:types is able to pull type declarations from a Meteor checkout, if it's being used, instead of only from the package repository (which may not contain the latest changes). This means that we should be able to re-enable the TypeScript test, which verifies that the TypeScript skeleton does, in fact, produce a project which passes type checking.
> 
> (Re-enabling the test also picked up a few latent bugs in the type declarations, so fix those as well)

cc @zodern @Grubba27 for awareness.


ebroder:
Oh, and just to keep everything linked together, this should work now because https://github.com/zodern/meteor-types/pull/8 has been merged and released.

StorytellerCZ (approved):

italojs:
cherry picked the commit https://github.com/meteor/meteor/commit/1e3bca429709798d76376f4287bffc35676a786a to release-3.4.1

PR #12951, opened 2024-01-05
Deprecate HTTP package in the docs

roy9495:
I have made changes to the HTTP documentation status so that it no longer appears in the documentation.


StorytellerCZ (changes requested):
Unless I'm mistaken, this doesn't remove the http documentation from the docs website. I don't think this will do anything at all on this issue.

StorytellerCZ (changes requested):
I think a complete removal is not what we want. Just removing it from the menu is fine with me. The document can still exist and be found in search if needed.

roy9495:
Hi, so in the last commit changes made in docs/_config.yml are to be kept and all other changes have to be reverted.

StorytellerCZ:
Keep the page, but remove it from the menu, but that is for v3 docs. I think for v2 docs, we can keep it there.

PR #12956, opened 2024-01-11
 Add --release option info to meteor publish

### LINKED ISSUE #12954: Document --release option on meteor publish command
Using Meteor 2.14 I can enter `meteor publish --help` which gives the following documentation:

```
Usage: meteor publish [--create]
       meteor publish --update

Publishes a new version of a local package to the package server. Must be run
from the directory containing the package. Reads the package.js file for version
information, builds the package and sends both the package source and the built
version of the package to the package server.

This will create at most one build of the package. If the package has an
OS-dependent binary component, publishing will only register metadata about the
package. The binary component will be compiled when it is installed into an
application with `meteor add`. To pre-build a package for a particular platform,
use the `meteor publish-for-arch` command.

This will mark you as the only maintainer of the package. You can use
'meteor admin maintainers' to change package maintainers. For more information
about admin commands, run 'meteor help admin'.

Change the metadata for a given version of a package by running with the
--update flag. That will set the git url, version summary, longform description
and documentation in the database to their new values. You can use 'meteor show'
to preview the results.

Pass --create to create a new package.

Options:
  --create   publish a new package
  --update   changed metadata of a previously published version
  --allow-incompatible-update   Allow packages in your project to be upgraded or
             downgraded to versions that are potentially incompatible with the
             current versions, if required to satisfy all package version
             constraints.
  --no-lint  don't run linters on the published package and its local
             dependencies before publishing
```

However, I didn't know that I can actually use `--release` here in order to resolve "top-level" version conflicts, which was recently a big blocker when attempting to publish packages for 3.0 beta.

This option should be added to the help and maybe also to the migration docs. Might be crucial for people who struggle with migrating packages.

qwertyroop:
This commit adds documentation for the --release option in the meteor publish
command. The --release option is introduced to help resolve top-level version
conflicts during the package publishing process, especially in scenarios where
packages have conflicting version requirements. Users can now specify the
appropriate release version to satisfy all package version constraints.

Fixes #12954


StorytellerCZ:
@jankapunkt please check

tools/cli/help.txt
```diff
@@ -914,11 +914,11 @@ Options:
 Publish a new version of a package to the package server.
 
 Usage: meteor publish [--create]
-       meteor publish --update
+       meteor publish --update [--release <version>]
```
zodern, line 917: I'm wondering if `--release` should be moved to the first example. It's normally used when publishing a new version instead of when updating the metadata.

tools/cli/help.txt
```diff
@@ -941,6 +941,12 @@ Pass --create to create a new package.
 Options:
   --create   publish a new package
   --update   changed metadata of a previously published version
+  --release <version>
+             resolve top-level version conflicts by specifying the release
+             version. This option is particularly useful during the migration
+             process when dealing with packages that have conflicting version
+             requirements. Specify the version that satisfies all package
+             version constraints.
```
zodern, line 949: This description is correct, but it describes one of the few side effects of using the `--release` option. The description should probably start with describing what the option does: `--release        Specify the release of Meteor to use` (taken from the `meteor run` help).

zodern:
There seem to be a few other commands that are also missing documentation for `--release` or only show it in the usage example, such as `meteor test` and `meteor test-packages`. I think most commands if not all of them have the `--release` option.

tools/cli/help.txt
Grubba27: I agree, especially now, using `--release` might become more necessary

StorytellerCZ:
@jankapunkt please check this.

tools/cli/help.txt
StorytellerCZ: Why not both? It applies to both anyway.

StorytellerCZ:
@qwertyroop any plans to update this?

tools/cli/help.txt
jankapunkt: correct, I used it a lot when I published packages for dual Meteor 2+3 compatibility but I had to use `--release 3.0.0-rc.2` explcitly to make it work because at that time the top-level Meteor version was still 2.x

jankapunkt:
@StorytellerCZ why did the docs-deploy ci fail exactly?

qwertyroop:
> @qwertyroop any plans to update this?

nope

denihs:
I'm closing this PR, then. We'll continue the work started here in other PR.

Issue #12959, opened 2024-01-12
Meteor3 asyncIterator does not exist on Asynchronous cursor


ToyboxZach:
The async cursor needs to define the asyncIterator

Probably something like this:

```
 [Symbol.asyncIterator]() {
    return this._cursor[Symbol.asyncIterator]();
  }
```
Discussed here:
https://forums.meteor.com/t/symbol-asynciterator-for-the-cursor-in-version-3-0-alpha-11/60549/7

PR #12960, opened 2024-01-12
Add asyncIterator to AsyncCursor

ToyboxZach:
As mentioned in this issue https://github.com/meteor/meteor/issues/12959 the new asynchronous cursor doesn't have the asyncIterator on it which would be a breaking change from version 2.0

This PR adds an async iterator following the example here https://javascript.info/async-iterators-generators
This enables this code:

```
for await (const item of collection){
doStuff(item)
}
```

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -1043,6 +1043,20 @@ class AsynchronousCursor {
   [Symbol.iterator]() {
     return this._cursor[Symbol.iterator]();
   }
+  
+  [Symbol.asyncIterator]() {
+    return {
+      next: async () => {
+        const next = await this._nextObjectPromise();
+        if (next) {
+          return { done: false, value: next };
+        }
+        return {
+          done: true,
+        };
+      },
+    };
+  }
 
   // Returns a Promise for the next object from the underlying cursor (before
   // the Mongo->Meteor type replacement).
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -1046,14 +1046,9 @@ class AsynchronousCursor {
   
   [Symbol.asyncIterator]() {
     return {
-      next: async () => {
-        const next = await this._nextObjectPromise();
-        if (next) {
-          return { done: false, value: next };
-        }
-        return {
-          done: true,
-        };
+      async next {
+        const value = await this._nextObjectPromise();
+        return { done: !value, value };
       },
     };
   }
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -1042,7 +1042,7 @@ class AsynchronousCursor {
   
   [Symbol.asyncIterator]() {
     return {
-      async next {
+      async next() {
         const value = await this._nextObjectPromise();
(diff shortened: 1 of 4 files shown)


packages/mongo/mongo_driver.js
```diff
@@ -1043,6 +1043,20 @@ class AsynchronousCursor {
...
+    return {
+      next: async () => {
+        const next = await this._nextObjectPromise();
+        if (next) {
+          return { done: false, value: next };
+        }
+        return {
+          done: true,
+        };
+      },
```
radekmie, line 1057: This can be simplified:
```suggestion
      async next {
        const value = await this._nextObjectPromise();
        return { done: !value, value };
      },
```

radekmie (changes requested):
Could you check whether the synchronous iterator works? It uses `this._cursor` and it seems there's no such a field. Plus it's rather impossible to make an asynchronous cursor be synchronously iterable. If so, please remove it as well.

Also, can you check the typings too?

ToyboxZach:
I'm not exactly sure what "typings" your comment is asking about, but if you are referring to the typescript types, I believe they are correct. We can't remove the syncIterator from the types since it is available on the client, and the asyncIterator was already there since it was there in version 2.

radekmie (approved):

packages/mongo/mongo_driver.js
```diff
@@ -1039,9 +1039,14 @@ class AsynchronousCursor {
 
     this._visitedIds = new LocalCollection._IdMap;
   }
-
-  [Symbol.iterator]() {
-    return this._cursor[Symbol.iterator]();
+  
+  [Symbol.asyncIterator]() {
+    return {
+      async next {
```
radekmie, line 1045: Sorry, I missed the parenthesis here 🙈 
```suggestion
      async next() {
```

ToyboxZach: I didn't notice either 🤷

ToyboxZach:
Whoops looks like I got my wires crossed and synced the wrong repos the other day. Not sure how best to fix that so its not pulling in the wrong changes

nachocodoner:
@ToyboxZach: I didn't get your last comment. What is blocking this PR?

The change looks good to me. I would suggest to add a new test case scenario that is actually using the new asyncIterator properly and as expected.

I also see the tests failing, not sure if this is because the state of changes on your branch needed the latest from release/3.0. We had a regression on that branch some time ago. Let me know if I can help you somehow.

ToyboxZach:
Ah yeah it's fine.

The tests are from changes missing in this branch, I'm going to be traveling for the next few weeks so I don't think I can add the tests to it anytime soon if anyone wants to finish it for me that would be great.

nachocodoner:
@ToyboxZach: Ok, thanks a lot for your contribution. With your permission, I'll handle the remaining changes.

I can branch out from your branch and draft a new PR for it to be able to push latest changes and the remaining test.

ToyboxZach:
Sounds great to me thanks!

nachocodoner:
Closing this one in favor of https://github.com/meteor/meteor/pull/12989 which includes the tests for asyncIterator

PR #12961, opened 2024-01-13
Replace validated method references with jam:method

harryadel:
https://github.com/meteor/validated-method is long over due for some changes. I opted instead to replace it with [jam:method](https://github.com/jamauro/method) for many reasons:

- It's backwards compatible with a very similar API
- Built with Meteor 3.0 in mind
- Offers more features and comes with better defaults like [authentication](https://github.com/jamauro/method#changing-authentication-rules) and many more


StorytellerCZ:
I personally prefer to stick as close as possible to the core, but you might have just sold me on this @harryadel 
Honestly, this is probably a good time to talk about bringing this to the core or at least replacing the official validated method with this if @jamauro would be up for that.

StorytellerCZ (approved):

fredmaiaarantes:
We could have other examples to show how good the `jam:method` package is!
I think it's fine to replace it here and perhaps add more docs about this package later.

I'd love to see some features (maybe all?) of this package integrated into the core.

fredmaiaarantes (approved):

jamauro:
Thanks guys! Feeling honored that `jam:method` will make it into the Meteor guide. Appreciate you @harryadel for creating this PR.

I think it could be interesting to bring it to the core though maybe it should remain as an add-on package. Happy to discuss in more depth down the line. 

For the guide, I'm also wondering if Simple Schema should remain as the recommended option for most people. In my biased opinion, I don't think Simple Schema still represents the best option. 🙂 Happy to discuss this more as well if others are interested.

harryadel:
@jamauro You've been putting out exceptional work lately and I opened up this PR because your package indeed offers a superior alternative. :clap:

PR #12359, opened 2022-12-12
Release 3.0

denihs:
# Still in development
Current released version: ``3.0-rc.2``
To use it, please do a 
```bash
meteor update --release 3.0-rc.2
```
you can also use the following command to install meteor RC

```bash
npx meteor@rc
```
## Highlights
TODO
See more in the [milestones](https://github.com/meteor/meteor/milestone/101).

Preview docs: https://v3-docs.meteor.com/

## Maintainers checklist

⚠️⚠️⚠️ You can follow the updates on [What's left until an official Meteor 3.0? Discussion](https://github.com/meteor/meteor/discussions/12865)

PR #12964, opened 2024-01-15
fix(minimongo): sync upsert doesn't create new documents

ggazzo:
Issue:
During the migration to Meteor 3 , for some reason, our UI was not working anymore.

Collections were returning empty arrays for some reason. After changing `upsert` to `upsertAsync`, everything seemed to start working as expected again.

I reviewed the documentation, and it explicitly states that Minimongo will continue to provide synchronous functions. Investigating the code, I realized that both functions are practically identical, except for the part that deals with the case where no record has been modified, where it is assumed that an insertion should be performed. Adding this code to the application made it return to normal, and it now honors what is written in the documentation.

### DIFF
--- a/packages/minimongo/minimongo_tests_client.js
+++ b/packages/minimongo/minimongo_tests_client.js
@@ -212,6 +212,21 @@ Tinytest.addAsync('minimongo - basics', async test => {
   test.equal(after.d, undefined);
 });
 
+
+Tinytest.addAsync('minimongo - upsert', async test => {
+  const c = new LocalCollection();
+
+  await c.upsertAsync({ name: 'doc' }, { name: 'doc' });
+  
+  test.equal(c.find({}).count(), 1);
+
+  await c.removeAsync({});
+
+  c.upsert({ name: 'doc' }, { name: 'doc' });
+  test.equal(c.find({}).count(), 1);
+});
+
+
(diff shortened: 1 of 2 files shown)


Grubba27:
Seems ok, the failing tests are not related to this change.

packages/minimongo/minimongo_tests_client.js
```diff
@@ -212,6 +212,22 @@ Tinytest.addAsync('minimongo - basics', async test => {
   test.equal(after.d, undefined);
 });
 
+
+Tinytest.addAsync('minimongo - upsert', async test => {
+  const c = new LocalCollection();
+  let count;
```
ggazzo, line 218:
```suggestion
```

nachocodoner (approved):
Good job!

Get latest release-3.0 into your branch and lets check the tests after last fixes.

ggazzo:
@nachocodoner rebased 
I dont tnink the error is related 

image

nachocodoner:
[release-3.0](https://github.com/meteor/meteor/pull/12359) currently fails, but the reason is different than the one you report here and I can see on this PR detail. The issue here is around docs setup which @Grubba27 fixed and pushed to release-3.0. It is weird the fix doesn't affect here. Could you try a full merge of release-3.0 into here, instead of rebasing?

PR #12969, opened 2024-01-17
[Meteor 3] Fix client ops in remote collections

nachocodoner:
Reported on slack: https://meteor-community.slack.com/archives/CN350MY1G/p1704695310758029

Some time back, we were tasked with addressing an issue occurring in a specific code scenario that seems to break isomorphism.

The next code behaves differently on the client and the server in Meteor 3.x, disrupting isomorphism. The operation should align on both client and server, the client's minimongo isn't populated with data by default, even though the server inserts data correctly into Mongo.

``` javascript
    
// In the client it doesn't log the data inserted
// In the server it logs the data inserted
console.log("-> isItemCreated?", NamedCollection.find().fetch()?.[0]);
```

The problem lies in the code behaving differently in Meteor 2.x and 3.x. It populates in the former but not in the latter. See the next picture to clarify:

image

The problem not only happens in `insertAsync` but also in other operations such as `updateAsync` and `removeAsync`.

image

This problem mainly occurs in testing scenarios, not broadly, as minimongo is typically populated and utilized by functions following proper reads from publications or methods, not as a result of collection operations.

A use case where happens is for example on tests: https://github.com/Meteor-Community-Packages/mongo-collection-instances/pull/40/files#diff-bbd9a41e09b247b7b1128731dc48f53d701e08e4ad08bba374523b3f4f23d43c

# Fix

Aligning the operations of Meteor 2.x and 3.x is achieved by introducing the `resolverType` option for collections (`new Collection(name, options)`). In Meteor 3.x, a distinction exists between stub and server promises. The former handles client simulation and minimongo population, while the latter solely manages success or error on the server call without populating the data in minimongo. The default is the server promise throughout Meteor 3.x codebase and tests, that is why changing to stub is not an easy option, above all being this an edge case.

The `resolverType` option offers `stub` and `server` values. Using `stub` maintains collection operation behavior to align with Meteor 2.x, addressing the isomorphic challenge.

``` javascript
 const NamedCollection = new Meteor.Collection('namedCollection', { resolverType: 'stub' });
    
// In the client it logs the data inserted
// In the server it logs the data inserted
console.log("-> isItemCreated?", NamedCollection.find().fetch()?.[0]);
```

Check the test to see this new option in action:
https://github.com/meteor/meteor/blob/319-fix-client-op--remote-collections/packages/mongo/mongo_livedata_tests.js#L4279-L4304

# Test

- [x] Add a test for this regression

### DIFF
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -95,7 +95,7 @@ export const loadAsyncStubHelpers = () => {
           const applyAsyncPromise = oldApplyAsync.apply(this, args);
           stubPromiseResolver(applyAsyncPromise.stubPromise);
           serverPromiseResolver(applyAsyncPromise.serverPromise);
-          applyAsyncPromise
+          applyAsyncPromise.stubPromise
             .then((result) => {
               finished = true;
               resolve(result);
@@ -104,6 +104,15 @@ export const loadAsyncStubHelpers = () => {
               finished = true;
               reject(err);
             });
+          applyAsyncPromise.serverPromise
+              .then((result) => {
+                finished = true;
+                resolve(result);
+              })
+              .catch((err) => {
+                finished = true;
+                reject(err);
+              });
         });
 
         Meteor._setImmediate(() => {
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -62,6 +62,7 @@ export const loadAsyncStubHelpers = () => {
   Connection.prototype.applyAsync = function () {
     let args = arguments;
     let name = args[0];
+    let options = args[2];
 
     if (currentMethodInvocation) {
       DDP._CurrentMethodInvocation._set(currentMethodInvocation);
@@ -92,19 +93,13 @@ export const loadAsyncStubHelpers = () => {
         let finished = false;
 
         Meteor._setImmediate(() => {
+          let { returnStubValue, returnServerResultPromise } = options || {};
+
           const applyAsyncPromise = oldApplyAsync.apply(this, args);
           stubPromiseResolver(applyAsyncPromise.stubPromise);
           serverPromiseResolver(applyAsyncPromise.serverPromise);
-          applyAsyncPromise.stubPromise
-            .then((result) => {
-              finished = true;
-              resolve(result);
-            })
-            .catch((err) => {
-              finished = true;
-              reject(err);
-            });
-          applyAsyncPromise.serverPromise
+          if (returnStubValue && returnServerResultPromise) {
+            applyAsyncPromise.stubPromise
               .then((result) => {
                 finished = true;
                 resolve(result);
@@ -113,6 +108,17 @@ export const loadAsyncStubHelpers = () => {
                 finished = true;
                 reject(err);
               });
+            return;
+          }
+          applyAsyncPromise
+            .then((result) => {
+              finished = true;
+              resolve(result);
+            })
+            .catch((err) => {
+              finished = true;
+              reject(err);
+            });
         });
 
         Meteor._setImmediate(() => {
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -62,7 +62,6 @@ export const loadAsyncStubHelpers = () => {
   Connection.prototype.applyAsync = function () {
     let args = arguments;
     let name = args[0];
-    let options = args[2];
 
     if (currentMethodInvocation) {
       DDP._CurrentMethodInvocation._set(currentMethodInvocation);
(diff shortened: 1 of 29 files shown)


packages/ddp-client/common/livedata_connection.js
```diff
@@ -797,7 +798,7 @@ export class Connection {
         callback = (err) => {
           err && Meteor._debug("Error invoking Method '" + name + "'", err);
         };
-      } else {
+      } else if (Meteor.isServer) {
```
denihs, line 801: This is supposed to be created in the client as well. This is most likely not the solution. And some tests are failing because of this change.

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -104,6 +104,15 @@ export const loadAsyncStubHelpers = () => {
...
             });
+          applyAsyncPromise.serverPromise
+              .then((result) => {
+                finished = true;
+                resolve(result);
+              })
+              .catch((err) => {
+                finished = true;
+                reject(err);
+              });
```
zodern, line 115: This only needs to run after the stub finishes (or immediately if the method does not have a stub). Running it after receiving the server result is too late.

nachocodoner: Trying to understand better this part of the code and your input is determinant. Thank you.

I was able to run the tests from `ddp-client` package. And I am finally done with a fix that not only ensures the current tests and expectations are still passing, but also the use case that is missing also works as it does in 2.x.

So, as far as I understand, the `applyAsyncPromise` is a promise that is like a combination of stub and server promises. So  in this part of the code, we listen and resolve or reject according to the success or error of the promise combination, any rejection will cause the client operation to not be executed, like the error on this regression `"Method insertAsync not found on the server [404]"`.

But when we have an operation that uses stub (`returnStubValue`) and server (`returnServerResultPromise`), we still want the client operation to run ok, even if the server fails for any reason. So in this scenario, `applyAsyncPromise.stubPromise` is the only promise I care to return and to continue sucessfully with the client collection operation. In any other case, `applyAsyncPromise` combination promise is behaving as expected as I check on the tests.

Still trying to know each part of this code having a better mental picture of this implementation, so any review, correction or feedback to understand further is appracieted.

nachocodoner: With this change I break other tests like this [one](https://github.com/meteor/meteor/blob/release-3.0/packages/mongo/mongo_livedata_tests.js#L227-L269) that actually expects to handle properly any collection operation server error.

So, in Meteor 2.x, the data is created locally when running collection operations that fail on the server due to "not found" method. It feels like wrong behavior, or more difficult to mimic in the current 3.x code without fibers.

In Meteor 3.x, we can't immediately create the local data within a remote collection context knowing that we need to wait for a server promise to response first.

Or there is something I am missing.

This issue was originally reported [here](https://meteor-community.slack.com/archives/CN350MY1G/p1704695310758029). I wonder if we can think in another way to approach this. Like change the pattern on using client collections for tests in a way that these are created as non-remote collections with `connection: null` on being instantiated on-fly and then they will effectively behave as sync collections.

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -92,9 +93,23 @@ export const loadAsyncStubHelpers = () => {
...
+            return;
+          }
```
nachocodoner, line 112: With this change I break other tests like this [one](https://github.com/meteor/meteor/blob/release-3.0/packages/mongo/mongo_livedata_tests.js#L227-L269) that actually expects to handle properly any collection operation server error.

So, in Meteor 2.x, the data is created locally when running collection operations that fail on the server due to "not found" method. It feels like wrong behavior, or more difficult to mimic in the current 3.x code without fibers.

In Meteor 3.x, we can't immediately create the local data within a remote collection context knowing that we need to wait for a server promise to response and we can't just react with the stub promise straight.

Or there is something I am missing.

This issue was originally reported [here](https://meteor-community.slack.com/archives/CN350MY1G/p1704695310758029). I wonder if we can think in another way to approach this. Like change the pattern on using client collections for tests in a way that these are created as non-remote collections with `connection: null` on being instantiated on-fly and then they will effectively behave as sync collections creating the data properly on each operation.

cc @denihs

denihs:
OSS-329

denihs:
OSS-319

denihs (approved):

jankapunkt:
@nachocodoner thanks a lot! is there a way to test this with some of our community packages without locally forking and building this specific branch?

nachocodoner:
@jankapunkt: I already tested when adding this new option, I just added here a proof on running this locally.

image

I don't think there is anything more we can do in this moment to run this change on tests without going to a release process. But surely we can include this for the next one.

jankapunkt (approved):

jankapunkt:
Thank you, I will update the tests in the community package repos, once the next beta is out :+1:

Grubba27:
Hey, @jankapunkt, beta .4 is out with this one!

PR #12970, opened 2024-01-17
feat: check - throw all errors

jamauro:
Currently, `check` throws immediately at the first error encountered. This PR maintains the default behavior and adds an optional flag `{ throwAllErrors: true }` to throw all errors in an array.


StorytellerCZ:
Is this for checks that check arrays or if you apply it to the first check will it run all the checks and then throw an array of errors for checks that have failed?

jamauro:
This is for checks of arrays and objects. 

> or if you apply it to the first check will it run all the checks and then throw an array of errors for checks that have failed?

If I understand your question correctly, you could accomplish this by putting all your checks into one `check`. E.g.

instead of:

```js
check(foo, String)
check(bar, Number)
// etc 
```

you could do this:
```js
check({foo, bar}, {
 foo: String, 
 bar: Number
}, {throwAllErrors: true})
```

PR #12972, opened 2024-01-18
Meteor 3.0 - FIX - publication handler is not being waited on

renanccastro:
Hello!
Found this issue while seeing metrics on Monti APM with Meteor 3.

When we have an async publication handler, Meteor livedata server is not waiting for the promise resolution(either success/error) to unblock the processMessage handler.

The current behavior is like all publications are being unblocked.

This PR fixes it.

### DIFF
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@ -671,7 +671,7 @@ Object.assign(Session.prototype, {
   },
 
   protocol_handlers: {
-    sub: function (msg, unblock) {
+    sub: async function (msg, unblock) {
       var self = this;
 
       // cacheUnblock temporarly, so we can capture it later
@@ -730,7 +730,7 @@ Object.assign(Session.prototype, {
 
       var handler = self.server.publish_handlers[msg.name];
 
-      self._startSubscription(handler, msg.id, msg.params, msg.name);
+      await self._startSubscription(handler, msg.id, msg.params, msg.name);
 
       // cleaning cached unblock
       self.cachedUnblock = null;
@@ -988,7 +988,7 @@ Object.assign(Session.prototype, {
     else
       self._universalSubs.push(sub);
 
-    sub._runHandler();
+    return sub._runHandler();
   },
 
   // Tear down specified subscription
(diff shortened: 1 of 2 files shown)


filipenevola (approved):
Nice catch.

denihs (approved):
Nice! Once the tests succeed, we can merge this.

filipenevola:
@denihs I think at least one test should be added to avoid regressions for this case, right? cc @renanccastro

nachocodoner:
> Nice! Once the tests succeed, we can merge this.

Just mentioning here that the test that fails here is the same that fails from the release-3.0 base branch, so it is not related with these changes. It also happens on a branch that I am fixing an issue for 3.x. There should be a branch that was merged against 3.0 that introduced it.

release-3.0 branch failure:
https://github.com/meteor/meteor/pull/12359/checks?check_run_id=20620785033

renanccastro:
@denihs added a new regression test for this case

packages/ddp-server/livedata_server_tests.js
```diff
@@ -350,6 +350,21 @@ Meteor.methods({
   },
 });
 
+Meteor.publish("livedata_server_test_sub_chain", async function () {
+  await new Promise(r => setTimeout(r, 2000));
+  this.ready();
+  return null;
+});
+
+Tinytest.addAsync("livedata server - waiting for subscription chain", (test, onComplete) => makeTestConnection(test, async (clientConn, serverConn) => {
```
filipenevola, line 359: I think the format is off in this block. Could you run prettier on it?

If I remember correctly we still have prettier in the root dir.

filipenevola (changes requested):

packages/ddp-server/livedata_server_tests.js
renanccastro: done!

filipenevola (approved):

packages/ddp-server/livedata_server.js
```diff
@@ -988,7 +988,7 @@ Object.assign(Session.prototype, {
     else
       self._universalSubs.push(sub);
 
-    sub._runHandler();
+    return sub._runHandler();
```
Grubba27, line 991: why return this? Should just await here work?

denihs: This way you can handle any error where you're calling this method.

denihs:
Merging it since the test failing here is already solved in the 3.0 branch

PR #12978, opened 2024-01-22
Service configuration async

StorytellerCZ:
Used `upsert` instead of `upsertAsync` on server, which broke.

I think this could be backported to Meteor 2 as well, to silence warnings coming from official packages.


StorytellerCZ:
Looks like this has been already fixed upstream and the extra async is not really needed, so closing.

PR #12991, opened 2024-02-03
Add nodemailer-openpgp

harryadel:
This PR adds the ability to encrypt your [emails using PGP](https://github.com/nodemailer/nodemailer-openpgp) by adding `encryptionKeys` & `shouldSign` to `sendEmail` like so:

```js
Email.sendAsync({
          from: 'foo@example.com',
          to: 'bar@example.com',
          text: '*Cool*, man',
          html: '<i>Cool</i>, man',
          encryptionKeys: ['-----BEGIN PGP PUBLIC KEY BLOCK-----…'],
          shouldSign: true
})

```


StorytellerCZ (approved):

StorytellerCZ:
I think we should target this to Meteor 3.1 so that this feature gets the highlight it deserves. Thoughts?

harryadel:
The only metric relevant to me right now is whether it hinders the release of 3.0 if it doesn't then it's better to merge so as not get stale.

StorytellerCZ:
I'm also thinking that it would be nice to have a section in the guide or docs with a example/tutorial on how to use this properly as I think it would be nice to give this more visibility, but this is not a blocker for merging.

packages/email/email.js
```diff
@@ -104,12 +107,15 @@ const knownHostsTransport = function (settings = undefined, url = undefined) {
     },
   });
 
+  if (options?.encryptionKeys || options?.shouldSign) {
```
nachocodoner, line 110: Maybe I miss something, but do we really need to have two configuration options to enable sign?

harryadel: A user could provide the `encryptionKeys` yet wouldn't want to encrypt things yet and he may also provide `shouldSign` but without proper `encryptionKeys` it's meaningless. So, both values must exists in order to be able to carry the encryption correctly.

jamauro: It looks like `shouldSign` in `nodemailer-openpgp` defaults to true and is only explicitly needed if you don't want to sign for some reason. https://github.com/nodemailer/nodemailer-openpgp/blob/1ff06189d0c2e021de09e15bc55bfcc8a30e2a51/lib/nodemailer-openpgp.js#L45

nachocodoner: Ok, I see then these parameters are just coming from `nodemailer-openpgp` and it will be by default `true` unless passed the contrary.

Thanks both good to know this is just perserving same config options than the original package integrated.

nachocodoner (approved):
i think this could also be a candidate for 2.16.

Could you also solve the conflict with that branch? Just a version tweak. Thanks.

PR #11164, opened 2020-09-08
Tree shaking

renanccastro:
Tree shaking! 😄 

In this PR the focus will be on removing unused exports and code from the final bundle.

## Description
Tree-shaking is a technic that the bundler uses to remove unused files and code from the final bundle. This PR implements 2 strategies:

1. Remove unused Files, by checking es6 imports/exports
2. Remove unused exports, by checking used imports.

This PR does not implement yet a verification for actual usage in the code, for tree shaking. It is restricted to imports, and also deals with dynamic and nested imports.

### app
For enabling tree shaking in your app, simply add the flag:
> sideEffects: false

to your package.json.
### packages
For packages, you can use on package.js the following:
>  api.setSideEffects(false);

inside the Package.onUse section.

## Checklist
- [x] Get a map of imports by static analyzing all isopacks
- [x] Convert relative import paths to absolute ones
- [x] Transform the resulting tree in the bundler to remove unused reify transformed exports: module.exports({ a => ...}).
- [x] Create a flag for packages.js to mark side effect free
- [x] Exclude packages with side effects from the elimination
- [x] Deal with const { ClientStream } = require("meteor/socket-stream-client");

## Improvements

- [ ] Improve import-scanner performance. Today, it's necessary to get rid of AST cache, as we need to check every possible usage for import/exports. One option is to use as the key, the parent->file path, instead of just the file key.
- [ ] Do static analysis to detect unused/used variables and use this info for tree shaking

## Current issues/decisions:
- Packages global variables: today, when adding a new meteor package, ex: meteor/logs, it will end up polluting the global scope and it's not even necessary to import it to use it's exported symbols. Solution: Only make tree shaking available for "es6" ready meteor packages, they will have to signal the package as side effects free, and all imports/exports will have to be done.
- [X] Missing bare files from client/compatibility folders.
- [X] "undefinedundefined" symbol on dynamic imports - it was caused by source maps missing after removing symbols
- [X] Aliases causes loop on some modules

Pending Decisions:
- [ ] Only run it on production builds? If we have a major "wrong" import, the local development build can have bad performance, or not be responsive at all. Think about import @material-ui/icons, for example.

Results until now:
- importing from lodash-es is getting only what is imported, ex: import { add } from 'lodash-es', is bundling only the add.js file and the add() function.

PR #12992, opened 2024-02-04
Dead code elimination on client

StorytellerCZ:
Based on #10056 and related to #11164

Heard about this recently and I think this could help a little, so I have decided to revive the code.
I would like to hear what people think and how we can make this better.


zodern:
I think this comment still applies - https://github.com/meteor/meteor/pull/10056#issuecomment-407761092
It really should work on the AST instead of using a regex.

Instead of doing it in the minifier, it might be better to do it as either part of the import scanner, or in the babel-compiler package. This way Meteor would also ignore any imports in the removed code. The import scanner would work in more situations but be more difficult to optimize - I was planning to redesign how its caching works to support this, but I probably won't have time in the near future.

jamauro:
This would be applied to the final bundle right?

If so, could this be viewed as an interim solution until a better solution materializes or the tree shaking pr is merged? Would be great to have `Meteor.isServer` code removed from the client bundle. 

Even if it were opt in only, I think it would be beneficial. It could always be removed when a better solution is ready.

StorytellerCZ:
@zodern point well taken, still I'm with @jamauro on this. Once I get around to this again, I will try to add a flag that can turn this on and off and we can turn this off by default until we have a better solution so this could be an interim solution for people who need it or want to experiment with this. Thoughts?

nachocodoner:
This is already achieved in the modern bundler with Rspack integrated. By default, app code will already remove that code within common guards properly.

We can provide this optimization at the Meteor bundler level, the one responsible for Atmosphere package optimizations or pure Meteor bundler usage. But I agree with @zodern that this should be done using AST and implemented in the import-scanner and babel-compiler parts. That will affect to all the bundler experience, not just in the production final build minified.

I will keep a note on this idea, and I may have time in the future to work on it.

Would you want to cover it, @StorytellerCZ? Otherwise, I think we should close this PR.

Issue #594, opened 2013-01-08
Inserting field named 'length' not allowed


raix:
Hi when trying a simple collection.insert({ owner: Meteor.userId(), length:3 }); in Meteor it's executed by client but revoked by server (or in process, same for collection.update) At the moment I renamed the field length to 'len' and voila it works... But now I'm not in compliance with the GridFS spec saying: 'fs.files.length' (the project: collectionfs.meteor.com)

KyrosDigital:
Just ran in to this issue, April 26th, 2019. 

Our team spent quite some time trying to wrap out heads about it and then found this issue. 

We are on Meteor 1.8.0.1

I'm assuming this bug / issue has been floating around in the framework for a couple years now. 

What can we all do to fix this in the framework? We love meteor and I've been using for years. 
Would be great to see this fixed. We would be happy to help contribute.

raix:
Wow 6years do fly, I had hoped underscore would have been swapped out by @jdalton’s lodash aldready 🧐

PR #12994, opened 2024-02-05
Feature/underscore 1.9.2

### LINKED ISSUE #12673: Meteor-embedded underscore library contains vulnerabilities
Hi guys, is there a way for me to upgrade the meter/underscore library to a more modern version? The one that seems to be bundled with meteor 2.12 is being reported by Chrome's Lighthouse as vulnerable (attached screenshot). 

Meteor Version 2.12.
Chrome Lighthouse inspection reports a vulnerability in the underscore library.
I have it added as "underscore" in the packages file.

image

StorytellerCZ:
Continuation of #12912 

Upgrading underscore to v1.8.3 while keeping Meteor specific changes. Though from looking at the issue in underscore it should have been fixed, so there might be something more going on here:
https://github.com/jashkenas/underscore/issues/770

More context: #594


StorytellerCZ:
Now that 1.7 is working, I'll try to push a few more version up.

StorytellerCZ:
I do believe this is now ready for further testing.
I have incorporated the fix from the previous version so that should be all set as well.

packages/underscore/package.js
```diff
@@ -0,0 +1,42 @@
+
+Package.describe({
+  summary: "JavaScript's utility _ belt",
+  version: '1.9.2',
```
italojs, line 4: maybe this PR isnt relevant anymore, but if it is, why do we decreasing the version here?

italojs (changes requested):

packages/underscore/package.js
StorytellerCZ: 6 -> 9 it is increasing. The decrease in the types bellow is so that they are closer to the actual version instead of a future version.

PR #12996, opened 2024-02-05
[Meteor 3] Accounts Ambiguous error messages in production set to true by default

StorytellerCZ:
Set `ambiguousErrorMessages` to true if in production. This increases default security.


denihs (approved):

zodern:
Is there a benefit to disabling it in development instead of using a consistent value in production and development? It might be unexpected for the errors to be different in production.

PR #13000, opened 2024-02-07
Fix bugs with test-in-console

zodern:
test-in-console wasn't running client tests due to missing a dependency on jQuery. This also fixes another error that happens after the client tests finished.

It would be nice if this could be released in a patch update instead of waiting for the next 2.x release.


Grubba27:
Screenshot 2024-02-15 at 09 44 27

Test in console 1.2.7 is out!

PR #13002, opened 2024-02-09
Fix #13001

### LINKED ISSUE #13001: Underscore 1.6.0 `_.intersection` bug (Meteor package)
Meteor 2.15 with Underscore 1.6.0

The Meteor version of of the `_.intersection` function in the underscore package returns unexpected results.

https://github.com/meteor/meteor/blob/master/packages/underscore/underscore.js#L528-L535

In the copy of the function below, `_.contains` returns boolean, so the `>= 0` comparison should be removed:

```
_.intersection = function(array) {
  var rest = slice.call(arguments, 1);
  return _.filter(_.uniq(array), function(item) {
    return _.every(rest, function(other) {
      return _.contains(other, item) >= 0;
    });
  });
};
```

Interestingly the v1.6.0 commit in the underscore repo doesn't have this problem:

https://github.com/jashkenas/underscore/blob/1f4bf626f23a99f7a676f5076dc1b1475554c8f7/underscore.js#L506

StorytellerCZ:
A quick fix in underscore. Recommend releasing this as a patch asap.

### DIFF
--- a/packages/underscore/package.js
+++ b/packages/underscore/package.js
@@ -1,7 +1,7 @@
 
 Package.describe({
   summary: "Collection of small helpers: _.map, _.each, ...",
-  version: '1.6.0',
+  version: '1.6.1',
 });
 
 Npm.depends({
--- a/packages/underscore/underscore.js
+++ b/packages/underscore/underscore.js
@@ -529,7 +529,7 @@
     var rest = slice.call(arguments, 1);
     return _.filter(_.uniq(array), function(item) {
       return _.every(rest, function(other) {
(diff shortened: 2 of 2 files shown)


Grubba27:
Just making a new underscore release should work?

Grubba27:
Screenshot 2024-02-15 at 09 29 09

underscore@1.6.1 is out!

StorytellerCZ:
Should we merge this now into devel or create 1.16 branch? I think this should be merged now into devel.

Grubba27:
Sorry for the delay! 

We can merge this for sure

PR #13004, opened 2024-02-12
Guide: update react native page

jankapunkt:
Replaces outdated info on Expo with up to date information

### DIFF
--- a/guide/source/react-native.md
+++ b/guide/source/react-native.md
@@ -17,22 +17,33 @@ For most projects, since your native app will display the same data and call the
 
 The only time you will need to make changes to your Meteor codebase is to enable certain features that are unique to your native app. For example, if you want to add push notifications to your native app, you will need to create a method on your Meteor app to store the native push tokens for a user.
 
-There are two routes for getting started with React Native. You can use "Vanilla" React Native, or you can use [Expo](https://expo.io/). Expo is a set of tools built around React Native. You can even try out React Native from your web browser using [Expo Snack](https://snack.expo.io/). You don't even need to install XCode or Android Studio to start using Expo.
+There are two ways to get started with React Native: using React Native CLI and Expo. 
 
-Here are the downsides to using Expo:
-- You cannot add Native Modules that use Native Code (Java, Swift, etc)
-- You cannot use packages that require linking (these are npm modules that include native code, and allow you to acess native features like the camera, push notifications, fingerprint authentication, etc). \
-- Apps that use Expo are much larger then pure React Native apps
+<h3 id="workflow-native">Use React Native CLI</h3>
 
-Expo does provide some native features ([click here for the full list](https://docs.expo.io/versions/latest/)), but if there is a feature missing that you need, you'll likely need to use an npm package or your own custom native code.
+The React Native guide to get started describes this workflow as following:
 
-You can "eject" your app from Expo to take advantage of Vanilla React Native features, but ejection cannot be undone easily.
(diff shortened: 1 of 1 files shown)


guide/source/react-native.md
```diff
@@ -17,22 +17,33 @@ For most projects, since your native app will display the same data and call the
...
-- You cannot use packages that require linking (these are npm modules that include native code, and allow you to acess native features like the camera, push notifications, fingerprint authentication, etc). \
-- Apps that use Expo are much larger then pure React Native apps
+<h3 id="workflow-native">Use React Native CLI</h3>
 
-Expo does provide some native features ([click here for the full list](https://docs.expo.io/versions/latest/)), but if there is a feature missing that you need, you'll likely need to use an npm package or your own custom native code.
+The React Native guide to get started describes this workflow as following:
 
-You can "eject" your app from Expo to take advantage of Vanilla React Native features, but ejection cannot be undone easily.
+> If you are already familiar with mobile development, you may want to use React Native CLI. It requires Xcode or Android Studio to get started. If you already have one of these tools installed, you should be able to get up and running within a few minutes. If they are not installed, you should expect to spend about an hour installing and configuring them.
```
gregfenton, line 26: Personal opinion here: the distinction in the RN guide is broad and lacks true direction.  It isn't about whether someone has previously done "native development" or not.  It is about the requirements of the app they are going to build.  If they likely will require some type of changes at the "native level", then consider invoking XCode and Android Studio.  But even that can be done both with RN CLI and Expo CLI, and it can be done _when needed_, not at the start of a project.

My main concern with most things that discuss RN is that they way too quickly get into this differentiation (what has historically been called "ejecting") without clearly describing whether this is needed or not.

Though I don't have experience with Meteor, I suspect there is nothing in use of Meteor that ***requires*** use of XCode and Android Studio, especially if one is to be using Expo.

I guess my real question for this document is: what is the goal here? To tell people everything there is to know about RN and Expo, or is it to help them get started building an RN/Expo project for using Meteor?

If it is the latter, I would recommend simply mentioning that there are 2 ways to go, but that this document will focus on one (I prefer Expo).  People who have experience with native development will have no problems going from an Expo project to an RN or even "native" project.  But people who are new both both Meteor and Expo/RN would be wisest to start with Expo.

In my mind, the goal here is to "get people started with Meteor on mobile devices".  I would from that go with "less is more" -- that is, don't try to cover every angle in the Meteor docs.  Cover one "happy path" and let more experienced developers deal with whatever paths they follow away from that happy path.

guide/source/react-native.md
```diff
@@ -17,22 +17,33 @@ For most projects, since your native app will display the same data and call the
...
 
-Here is the link to the React Native getting started documentation: https://reactnative.dev/docs/environment-setup
+<h3 id="workflow-expo">Use Expo</h3>
 
-Once you have your environment setup and have your app running on your device or in the emulator, you can proceed to the next step of the guide: "Meteor React Native Installation"
+The React Native guide to get started describes this workflow as following:
+
+> If you are new to mobile development, the easiest way to get started is with Expo Go.
+
+Expo is an [open-source framework](https://github.com/expo/expo) for apps that run natively on Android, iOS, and the web. The `expo` npm package enables a suite of incredible features for React Native apps. The `expo` package can be installed in nearly any React Native project. See [what Expo offers](https://docs.expo.dev/core-concepts/) for more information or see [why does Expo have its own SDK](https://docs.expo.dev/faq/#why-does-expo-have-its-own-sdk).
```
gregfenton, line 36: Personally I'd remove the world "incredible" from the second sentence.  Maybe "productive" or "useful" or possibly "powerful"....but "incredible" is not informative and risks showing bias.

italojs:
@jankapunkt is it still updated?

jankapunkt:
@italojs this might need some more iteration as it is a bit outdated now, expo is moving fast as well and we should keep it short and just hint to the expo guide and the meteor react native repo

Issue #12870, opened 2023-11-09
Using **Accounts-base** with another DDP connection


gonzadelarge:
[ https://github.com/meteor/meteor/tree/devel/packages/accounts-base](url)

![Captura de pantalla 2023-11-09 101830](https://github.com/meteor/meteor/assets/86612394/ef03cbdd-63c5-4aaf-9d25-84ffe54d352e)

Expected behavior: It is expected to allow a user to successfully log in using *loginWithToken* even even when the users are managed by an external server.

Actual behavior: After manual login, upon refreshing the page, the first try to log in is with *loginWithToken* using the default connection, causing the app to log out without using the token and forcing the user to manually log in again.

Meteor version: METEOR@2.13.3
Accounts version: accounts-base@2.2.8

Reproduce the problem:
1. Attempt to use Accounts-base with an external server managing the users.
2. Notice *_initLocalStorage()* is executing before the *Meteor.startup* of your app starts.
3. On the first use of the app, one can log in successfully. But upon refreshing the page, the first log-in attempts use the default connection causing a log-out.

Detailed issue: 
I am trying to use Accounts-base with an external server that manages the users. The main problem is **AUTO LOGIN** *_initLocalStorage* because this executes before my app starts. This means when *_initLocalStorage* executes and calls *loginWithToken*, Accounts use Meteor's standard **DDP** and default *users* collection to search for the user and token on the main server of the app, instead of external server, and then I'll never be able to do *loginWithToken* because the users are in another collection in the external server.

Error message: "Error logging in with token: Error: You've been logged out by the server. Please log in again. [403]

Somethings that I've already tried
- Accounts allows to create an Accounts Instance for multi-server, but I have the same problem because **AccountsClient** is executed before my application started, also, Accounts provides different methods for login, but *loginWithPassword* belongs to accounts-password, and if I try to use the multi-server option, I will never be able to use both methods (loginWithPassword, loginWithToken) because one of them is using the application server and the other its external server with their respective collections.
- I could find that the config method in *accounts_common.js* allows to configure the package from **Meteor.settings** but I could not do it by any way **(Set global accounts options. You can also set these in `Meteor.settings.packages.accounts` without the need to call this function)**. 
- If I apply the **ddpUrl** option manually on the package in the AccountsClient constructor, the package behavior is correct and everything is fine.

![Captura de pantalla 2023-11-09 105748](https://github.com/meteor/meteor/assets/86612394/35f576ae-144b-4788-adcd-14f095c77b19)

PR #13005, opened 2024-02-12
Get default options from meteor settings

gonzadelarge:
### Add support for custom configuration in AccountsClient initialization
**Description**
Currently, the _accounts-base_ package begins before your own project does it. If you want to use a custom configuration, you must wait for the package to start and then pass the configuration you need. This can cause problems when, for example, you want to use a different DDP connection, such as login attempts to the wrong server on the first login attempt.

At this time, the _accounts-base_ package is ready to receive different configuration parameters, however, you will have to wait for the package to start, and then, when your application starts, you can change the configuration.

To solve this problem, I modified the _accounts-base_ **client_main.js** file to allow custom configuration when initializing **AccountsClient**. In my version, the **AccountsClient** constructor accepts the path to the **meteor.settings** file. If the **Meteor.settings.public.packages.accounts** field exists, the accounts will use the settings defined there. Otherwise, it will initialize with the default settings as it currently works.

**Proposed Changes**
Modification of the client_main.js file, line 10, to accept the path to the Meteor settings file and utilize it in the initialization of AccountsClient.

**Related Issue**
This change addresses the problem reported in issue #12870: [Link to the issue](https://github.com/meteor/meteor/issues/12870)

**Additional Notes**
This change has minimal impact on the existing _accounts-base_ codebase and resolves a significant issue for users needing custom configurations.

** Example of use**
settings_example

### DIFF
--- a/docs/source/api/accounts.md
+++ b/docs/source/api/accounts.md
@@ -338,3 +338,41 @@ Accounts.ui.config({
 
 
 Since Meteor 2.7 you can configure these in your Meteor settings under `Meteor.settings.public.packages.accounts-ui-unstyled`.
+
+<h3 id="initialize-with-custom-settings">Initialize with custom settings</h3>
+
+This feature allows users to specify custom configuration parameters for both client-side and server-side initialization.
+
+* Client
+
+{% apibox "AccountsClient" %}
+
+On the client-side, AccountsClient can be initialized with custom parameters provided through Meteor settings configuration. Below is an example of how to use this functionality:
+
+```json
+{
+  "public": {
+    "packages": {
+      "accounts": {
+        ...configParams
+      }
+    }
+  }
+}
+```
+
+* Server
+
+{% apibox "AccountsServer" %}
+
+On the server-side, AccountsServer can also be initialized with custom parameters. Server-specific configuration may differ from the client and can also be specified through Meteor settings. Below is an example of how to do it:
+
+```json
+{
+  "packages": {
+    "accounts": {
+      ...configParams
+    }
+  }
+}
+```
--- a/packages/accounts-base/client_main.js
+++ b/packages/accounts-base/client_main.js
@@ -7,7 +7,7 @@ import {
  * @namespace Accounts
  * @summary The namespace for all client-side accounts-related methods.
  */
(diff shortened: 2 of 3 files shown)


StorytellerCZ (changes requested):
Not bad, but I think we also need the same approach for server (if we don't have it already). Plus documentation addition is needed.

gonzadelarge:
> Not bad, but I think we also need the same approach for server (if we don't have it already). Plus documentation addition is needed.

I have added the same option on the server side. Now you can pass parameters to AccountsServer from meteor.settings. I have also updated the documentation.

On the other hand, I noticed that AccountsCommon "accounts_common.js" has a comment (line 211) with the option to initialize "accounts-base" with meteor.settings, but it is just that, a comment. There is no line of code in the file that uses meteor.settings to initialize "accounts-base".

StorytellerCZ:
So far looking good. I just have one question that we need to test: does client setting propagate to server (possibly via commons)? If not then we need to make it happen and in this case server settings should override the client settings if there is conflict (`{ ...clientSettings, ...serverSettings }`).

gonzadelarge:
> So far looking good. I just have one question that we need to test: does client setting propagate to server (possibly via commons)? If not then we need to make it happen and in this case server settings should override the client settings if there is conflict (`{ ...clientSettings, ...serverSettings }`).

After reviewing the code I found that the server side only expects one param ("loginExpirationInDays") from the list in the "accounts" configuration, but this "loginExpirationInDays" param can only be used by the server side. It does not matter if there are duplicate parameters in the configuration. Also, the server does not use the options directly, it needs to call the config() method implicitly to change the loginExpirationInDays parameter.

StorytellerCZ:
For the server documenting that only `loginExpirationInDays` is expected I think is enough. We could also make it easier that this setting could be on the client and for server instantiation it would be `{ ...clientSettings, ...serverSettings }`, so that you can write the settings in one place.

nachocodoner:
I like it, I think this is nice to have. 🚀

StorytellerCZ:
@gonzadelarge can you please update with the latest changes from 2.16. Thank you!

PR #12436, opened 2023-01-10
DDP resumption

StorytellerCZ:
Changes to DDP resumption code in response to reviewer comments on @znewsham's PR #11559 and @vlasky's #11845

* Constant MAX_QUEUE_LENGTH has been replaced with option maxMessageQueueLength
* Constant DISCONNECT_GRACE_PERIOD has been replaced with option disconnectGracePeriod
* Use of _.each() replaced with Array forEach()
* isPingPong replaced with descriptive name ignoredMsgsForSessionOutOfDateCheck

PR #13009, opened 2024-02-16
New mongo package options to optimize Oplog tailing performance to include/exclude certain collections

Twisterking:
Hello all,

Following several conversations on the drawbacks of Meteor's oplog implementation and its limitations when it comes to scaling (see: [my forums post](https://forums.meteor.com/t/cpu-spikes-due-to-oplog-updates-without-subscriptions/60028), [a slightly older blog post](https://blog.meteor.com/tuning-meteor-mongo-livedata-for-scalability-13fe9deb8908) and a [2 year old github discussion](https://github.com/meteor/meteor/discussions/11842)), I investigated in the core a bit and propose an easy solution which, I hope, should work for many users who have issues with this.

Please consider our situation at Orderlion: We have several collections, where huge amounts of data are being updated on a daily basis. On all of these collections we have not a single publication, for these we don't care about any reactivity.

Due to Meteor's Oplog implementation, even in these cases a big batch insert/update of data results often in very big sustained CPU spikes on the meteor app server.

Internal tests I did with my suggested implementation have shown, **that the CPU load due to Oplog tailing collections with big batch updates, can be reduced about 100x with just completely ignoring those collections**. 
During an import of about 600k documents in a test collection, I could reduce the CPU load from an average 25-30% down to approx. 0,2 - 0,3%.

The usage is actually very straight forward: Consider a `settings.json` file like this:

```js
{
  // ... all other settings,
  "public": {
    // all your public settings
  },
  "packages": {
    "mongo": {
      "oplogExcludeCollections": ["products", "prices"] // This would exclude both collections "products" and "prices" from any oplog tailing.
      // OR:
      "oplogIncludeCollections": ["messages"] // ONLY watch the "messages" collection for oplog updates (and all "admin.$cmd" oplogs)
    }
  }
}
```

Please feel free to comment but I really hope this can be merged soon. It would help us at Orderlion a lot and solve many unwanted server restarts (server is auto-restarting because it becomes unresponsive due to the sustained high CPU load).

Cheers, Patrick

### DIFF
--- a/docs/source/api/collections.md
+++ b/docs/source/api/collections.md
@@ -1086,9 +1086,9 @@ certificates configurations.
 
 > Oplog options were introduced in Meteor 2.15.1
 
-If you set the [`MONGO_OPLOG_URL`](https://docs.meteor.com/environment-variables.html#MONGO-OPLOG-URL) env var, Meteor will use MongoDb's Oplog to show efficient, real time updates to your users via your subscriptions.
+If you set the [`MONGO_OPLOG_URL`](https://docs.meteor.com/environment-variables.html#MONGO-OPLOG-URL) env var, Meteor will use MongoDB's Oplog to show efficient, real time updates to your users via your subscriptions.
 
-Due to how Meteor's Oplog implementation is built behind the scenes, if you have certain collections where you expect **big amounts of `insert` or `update` transactions**, this might lead to **big CPU spikes on your meteor app server, even if you have no publications/subscriptions on any data/documents of these collections**. For more information on this, please have a look into [this blog post from 2016](https://blog.meteor.com/tuning-meteor-mongo-livedata-for-scalability-13fe9deb8908), [this github discussion from 2022](https://github.com/meteor/meteor/discussions/11842) or [this meteor forums post from 2023](https://forums.meteor.com/t/cpu-spikes-due-to-oplog-updates-without-subscriptions/60028).
+Due to how Meteor's Oplog implementation is built behind the scenes, if you have certain collections where you expect **big amounts of write operations**, this might lead to **big CPU spikes on your meteor app server, even if you have no publications/subscriptions on any data/documents of these collections**. For more information on this, please have a look into [this blog post from 2016](https://blog.meteor.com/tuning-meteor-mongo-livedata-for-scalability-13fe9deb8908), [this github discussion from 2022](https://github.com/meteor/meteor/discussions/11842) or [this meteor forums post from 2023](https://forums.meteor.com/t/cpu-spikes-due-to-oplog-updates-without-subscriptions/60028).
 
 To solve this, **2 Oplog settings** have been introduced **to tweak, which collections are *watched* or *ignored* in the oplog**.
 
--- a/docs/source/api/collections.md
+++ b/docs/source/api/collections.md
@@ -1082,6 +1082,38 @@ option:
 You can pass any MongoDB valid option, these are just examples using
 certificates configurations.
 
+<h3 id="mongo_oplog_options">Mongo Oplog Options</h3>
+
+> Oplog options were introduced in Meteor 2.15.1
+
+If you set the [`MONGO_OPLOG_URL`](https://docs.meteor.com/environment-variables.html#MONGO-OPLOG-URL) env var, Meteor will use MongoDb's Oplog to show efficient, real time updates to your users via your subscriptions.
+
+Due to how Meteor's Oplog implementation is built behind the scenes, if you have certain collections where you expect **big amounts of `insert` or `update` transactions**, this might lead to **big CPU spikes on your meteor app server, even if you have no publications/subscriptions on any data/documents of these collections**. For more information on this, please have a look into [this blog post from 2016](https://blog.meteor.com/tuning-meteor-mongo-livedata-for-scalability-13fe9deb8908), [this github discussion from 2022](https://github.com/meteor/meteor/discussions/11842) or [this meteor forums post from 2023](https://forums.meteor.com/t/cpu-spikes-due-to-oplog-updates-without-subscriptions/60028).
+
+To solve this, **2 Oplog settings** have been introduced **to tweak, which collections are *watched* or *ignored* in the oplog**.
+
+**Exclusion**: To *exclude* for example all updates/inserts of documents in the 2 collections called `products` and `prices`, you would need to set the following setting in your Meteor settings file:
+
+```json
+  "packages": {
+    "mongo": {
+      "oplogExcludeCollections": ["products", "prices"]
+    }
+  }
+```
+
+**Inclusion**: vice versa, if you only want to watch/*include* the oplog for changes on documents in the 2 collections `chats` and `messages`, you would use:
+
+```json
+  "packages": {
+    "mongo": {
+      "oplogIncludeCollections": ["chats", "messages"]
+    }
+  }
+```
+
+For obvious reasons, using both `oplogExcludeCollections` and `oplogIncludeCollections` at the same time is not possible and will result in an error.
+
 <h3 id="mongo_connection_options_settings">Mongo.setConnectionOptions</h3>
 
 You can also call `Mongo.setConnectionOptions` to set the connection options but
--- a/packages/mongo/collection_tests.js
+++ b/packages/mongo/collection_tests.js
@@ -210,7 +210,7 @@ Tinytest.add('collection - calling find with an invalid readPreference',
 Tinytest.add('collection - inserting a document with a binary should return a document with a binary',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary1');
+      const collection = new Mongo.Collection('testBinary1' + test.id);
       const _id = Random.id();
       collection.insert({
         _id,
@@ -232,7 +232,7 @@ Tinytest.add('collection - inserting a document with a binary should return a do
 Tinytest.add('collection - inserting a document with a binary (sub type 0) should return a document with a uint8array',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary8');
+      const collection = new Mongo.Collection('testBinary8' + test.id);
       const _id = Random.id();
       collection.insert({
         _id,
@@ -254,7 +254,7 @@ Tinytest.add('collection - inserting a document with a binary (sub type 0) shoul
 Tinytest.add('collection - updating a document with a binary should return a document with a binary',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary2');
+      const collection = new Mongo.Collection('testBinary2' + test.id);
       const _id = Random.id();
       collection.insert({
         _id
@@ -277,7 +277,7 @@ Tinytest.add('collection - updating a document with a binary should return a doc
 Tinytest.add('collection - updating a document with a binary (sub type 0) should return a document with a uint8array',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary7');
+      const collection = new Mongo.Collection('testBinary7' + test.id);
       const _id = Random.id();
       collection.insert({
         _id
@@ -300,7 +300,7 @@ Tinytest.add('collection - updating a document with a binary (sub type 0) should
 Tinytest.add('collection - inserting a document with a uint8array should return a document with a uint8array',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary3');
+      const collection = new Mongo.Collection('testBinary3' + test.id);
       const _id = Random.id();
       collection.insert({
         _id,
@@ -322,7 +322,7 @@ Tinytest.add('collection - inserting a document with a uint8array should return 
 Tinytest.add('collection - updating a document with a uint8array should return a document with a uint8array',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary4');
+      const collection = new Mongo.Collection('testBinary4' + test.id);
       const _id = Random.id();
       collection.insert({
         _id
@@ -348,7 +348,7 @@ Tinytest.add('collection - updating a document with a uint8array should return a
 Tinytest.add('collection - finding with a query with a uint8array field should return the correct document',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary5');
+      const collection = new Mongo.Collection('testBinary5' + test.id);
       const _id = Random.id();
       collection.insert({
         _id,
@@ -368,7 +368,7 @@ Tinytest.add('collection - finding with a query with a uint8array field should r
 Tinytest.add('collection - finding with a query with a binary field should return the correct document',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary6');
+      const collection = new Mongo.Collection('testBinary6' + test.id);
       const _id = Random.id();
       collection.insert({
         _id,
@@ -455,6 +455,4 @@ Meteor.isServer && Tinytest.addAsync('collection - simple add', async function(t
   id = await collection.insertAsync({a: 2});
   test.equal((await collection.findOneAsync(id)).a, 2);
   await collection.removeAsync({});
-  
-})
-
+});
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -30,6 +30,7 @@ OplogHandle = function (oplogUrl, dbName) {
   var self = this;
   self._oplogUrl = oplogUrl;
   self._dbName = dbName;
+  console.log('OplogHandle CONSTRUCTOR()');
 
   self._oplogLastEntryConnection = null;
   self._oplogTailConnection = null;
@@ -84,6 +85,8 @@ OplogHandle = function (oplogUrl, dbName) {
 
   self._startTailing();
 };
+
+MongoInternals.OplogHandle = OplogHandle;
 
 Object.assign(OplogHandle.prototype, {
   stop: function () {
@@ -241,7 +244,7 @@ Object.assign(OplogHandle.prototype, {
       self._lastProcessedTS = lastOplogEntry.ts;
     }
 
-    // These 2 settings allow you to either only watch certain collections (oplogWatchCollections), or exclude some collections you don't want to watch for oplog updates (oplogExcludeCollections)
+    // These 2 settings allow you to either only watch certain collections (oplogIncludeCollections), or exclude some collections you don't want to watch for oplog updates (oplogExcludeCollections)
     // Usage:
     // settings.json = {
     //   "packages": {
@@ -251,11 +254,12 @@ Object.assign(OplogHandle.prototype, {
     //     }
     //   }
     // }
-    const includeCollections = Meteor.settings?.packages?.mongo?.oplogWatchCollections;
+    const includeCollections = Meteor.settings?.packages?.mongo?.oplogIncludeCollections;
     const excludeCollections = Meteor.settings?.packages?.mongo?.oplogExcludeCollections;
     if (includeCollections?.length && excludeCollections?.length) {
-      throw new Error("Can't use both mongo oplog settings oplogWatchCollections and oplogExcludeCollections at the same time.");
+      throw new Error("Can't use both mongo oplog settings oplogIncludeCollections and oplogExcludeCollections at the same time.");
     }
+    console.log('OplogHandle _startTailing() excludeCollections', excludeCollections);
     if (excludeCollections?.length) {
       oplogSelector.ns = {
         $regex: oplogSelector.ns,
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -33,6 +33,7 @@ OplogHandle = function (oplogUrl, dbName) {
 
   self._oplogLastEntryConnection = null;
   self._oplogTailConnection = null;
+  self._oplogOptions = null;
   self._stopped = false;
   self._tailHandle = null;
   self._readyFuture = new Future();
@@ -263,6 +264,7 @@ Object.assign(OplogHandle.prototype, {
         $regex: oplogSelector.ns,
         $nin: excludeCollections.map((collName) => `${self._dbName}.${collName}`)
       }
+      self._oplogOptions = { excludeCollections };
     }
     else if (includeCollections?.length) {
       oplogSelector = { $and: [
@@ -273,6 +275,7 @@ Object.assign(OplogHandle.prototype, {
         { $or: oplogSelector.$or }, // the initial $or to select only certain operations (op)
         { ts: oplogSelector.ts }
       ] };
+      self._oplogOptions = { includeCollections };
     }
 
     var cursorDescription = new CursorDescription(
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -30,7 +30,6 @@ OplogHandle = function (oplogUrl, dbName) {
   var self = this;
   self._oplogUrl = oplogUrl;
   self._dbName = dbName;
-  console.log('OplogHandle CONSTRUCTOR()');
 
   self._oplogLastEntryConnection = null;
   self._oplogTailConnection = null;
@@ -259,7 +258,6 @@ Object.assign(OplogHandle.prototype, {
     if (includeCollections?.length && excludeCollections?.length) {
       throw new Error("Can't use both mongo oplog settings oplogIncludeCollections and oplogExcludeCollections at the same time.");
     }
-    console.log('OplogHandle _startTailing() excludeCollections', excludeCollections);
     if (excludeCollections?.length) {
       oplogSelector.ns = {
         $regex: oplogSelector.ns,
--- a/packages/mongo/oplog_tailing.js
+++ b/packages/mongo/oplog_tailing.js
@@ -241,7 +241,16 @@ Object.assign(OplogHandle.prototype, {
       self._lastProcessedTS = lastOplogEntry.ts;
     }
 
-    // these 2 settings allow you to either only watch certain collections (oplogWatchCollections), or exclude some collections you don't want to watch for oplog updates (oplogExcludeCollections)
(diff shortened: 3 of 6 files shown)


packages/mongo/oplog_tailing.js
```diff
@@ -241,6 +241,38 @@ Object.assign(OplogHandle.prototype, {
...
+    // Usage:
+    // settings.json = {
+    //   "packages": {
+    //     "mongo": {
+    //       "oplogExcludeCollections": ["products", "prices"] // This would exclude both collections "products" and "prices" from any oplog tailing. 
+    //                                                            Beware! This means, that no subscriptions on these 2 collections will update anymore!
+    //     }
+    //   }
+    // }
+    const includeCollections = Meteor.settings?.packages?.mongo?.oplogWatchCollections;
```
radekmie, line 254: IMO it should be called `oplogIncludeCollections`.

packages/mongo/oplog_tailing.js
```diff
@@ -241,6 +241,38 @@ Object.assign(OplogHandle.prototype, {
...
+    }
+    else if (includeCollections?.length) {
+      oplogSelector = { $and: [
+        { $or: [
+          { ns: /^admin\.\$cmd/ },
+          { ns: { $in: includeCollections.map((collName) => `${self._dbName}.${collName}`) } }
+        ] },
+        { $or: oplogSelector.$or }, // the initial $or to select only certain operations (op)
+        { ts: oplogSelector.ts }
+      ] };
```
radekmie, line 273: Why not this instead:
```suggestion
      oplogSelector.ns = {
         $regex: oplogSelector.ns,
         $in: excludeCollections.map((collName) => `${self._dbName}.${collName}`)
       }
```

_(Of course the two branches will be almost identical then, just asking for further simplification.)_

radekmie (commented):
I think the most important part will be to add tests. At least:
1. Configure both and see if it fails.
2. Configure inclusion and check if non-included ones are not processed.
3. Configure exclusion and check if non-excluded ones are processed.

harryadel:
Why can the configuration be set from settings.json and not the code? 
```js
// This is pseudo code so don't judge
import { mongo } from 'meteor/mongo'

mongo.oplogExcludeCollections = [];
mongo.oplogWatchCollections = [];

```
 I've noticed this trend in Meteor lately and I believe it's getting out of hand. The settings file is getting bloated with config. It's supposed to hold ENV variables/secrets. :man_shrugging: 
 
 cc @zodern

Twisterking:
@harryadel 
In my opinion the settings file is the right place for this. The oplog tailing is setup very very early during the startup of your app. Finding the right place to put your suggested code so that your settings are defined "on time" is not straight forward in my opinion. 

The settings file is kinda "always loaded". 

Also I tried to follow the logic that already exists for the mongo package - it already uses the Meteor.settings for other package options.

harryadel:
@Twisterking I understand why you went this path. I'm merely discussing the idea not trying to put you down :)

>Finding the right place to put your suggested code so that your settings are defined "on time" is not straight forward in my opinion.

But isn't that what `Meteor.startup` is for?

>Also I tried to follow the logic that already exists for the mongo package - it already uses the Meteor.settings for other package options.

Yeah, this trend has been going on for a while so naturally you continued down this trail. I just don't think this path  is the best IMO.

Twisterking:
@harryadel understood, no worries! No harm done!
I am aware of Meteor.startup of course, but I was not sure if this works in this case because the mongo package indeed is very much core. 
Anyways ...

I will write some tests as suggested and will also create a PR for the meteor docs so this new setting is documented. 

Any other feedback? As I said I would love to get this PR ready for merging asap! :+1

packages/mongo/oplog_tailing.js
Twisterking: I am sorry I dont fully follow here. With your suggestion, this is basically an $and. 
But this wont work for the "include variant", as you need to support both admin.$cmd logs AND the user's included collections - so you actually need an $or in the end. 
What am I missing?

radekmie: You're right 🤦 Sure, let's leave it as it is.

radekmie:
@Twisterking I'm for leaving it in the settings. Doing it in the code would be problematic, as it'd need to execute before the connection is established and that's _deep_.

@harryadel The `Meteor.startup` is for executing code _after_ the server starts, so it's too late. Even top-level in the app is too late (you can execute database operations there already).

Twisterking:
> I think the most important part will be to add tests. 

@radekmie I am very happy to do that! The issue is, I don't really have any experience with `tinytest` yet.
I will make that work somehow, but: **What I do not know (and where I will need your support)** is, how can I define a `Meteor.settings` entry when running the Oplog tests? I tried to just define it somehow like this:

```js
process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
  'mongo-livedata - oplog - options.oplogExcludeCollections',
  async test => {
    if (!Meteor.settings.packages) Meteor.settings.packages = {};
    if (!Meteor.settings.packages.mongo) Meteor.settings.packages.mongo = {};
    Meteor.settings.packages.mongo.oplogExcludeCollections = ['foobar'];
    console.log('oplogExcludeCollections', Meteor.settings.packages.mongo);
    test.equal(true, true);
  }
);
```

This of course works, but it is "too late" - the oplog file itself is already loaded and there, `Meteor.settings.packages.mongo` is undefined! 😢 

--> **How do I define a Meteor setting in a test "early enough"?**

radekmie:
The only easy option I see would be to change the `Meteor.settings`, create a new instance of `OplogHandle`, test it, and revert the `Meteor.settings`.

Twisterking:
I am sorry, but I am not familiar enough with the meteor core to do this based on this comment! :/ I will need some more detailed help please! Maybe @zodern can help?

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,30 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
 
+process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
+  'mongo-livedata - oplog - options.oplogExcludeCollections',
+  async test => {
+    const testCollectionName = "test-" + randomId;
+    if (!Meteor.settings.packages) Meteor.settings.packages = {};
+    if (!Meteor.settings.packages.mongo) Meteor.settings.packages.mongo = {};
+    // Meteor.settings.packages.mongo.oplogExcludeCollections = [testCollectionName];
+
+    const myOplogHandle = new MongoInternals.OplogHandle(process.env.MONGO_OPLOG_URL, testCollectionName);
```
Twisterking, line 193: @radekmie might need your help here to make this work please! :)

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,30 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
+    MongoInternals.defaultRemoteCollectionDriver().mongo._setOplogHandle(myOplogHandle);
+
+    const TestCollection = new Mongo.Collection(testCollectionName);
+
+    let handle = TestCollection.find({ foo: 'bar' }).observeChanges({
+      added(id, fields) {
+        console.log('TEST added:', id, fields);
+        // Not working anymore! :(
```
Twisterking, line 201: Without using my custom `OplogHandle`, this console.log works just fine! So there has to be something wrong with my implementation of setting `myOplogHandle` to be used with the `defaultRemoteCollectionDriver`.

zodern:
This will be very helpful for some apps. I'm not very familiar with this code, but I have a couple questions.

1. Meteor tracks the last oplog entry it's seen to know where to resume tailing the oplog. If there's been a large number of newer oplog entries for databases or collections that are excluded, could resume tailing cause performance issues if mongo has to filter out the unneeded docs? This happens every time Meteor does not receive an oplog doc for 30 seconds, there is a failover, or in some other situations.

To give an example in case my description isn't clear. Let's say the database that is being observed is updated every 10 minutes on average, and there is another database for metrics that has hundreds of updates per second. Whenever Meteor restarts tailing the oplog (which would happen almost every 30 seconds since there are rarely oplog entries for the observed database), it would query for all of the relevant docs newer than the last oplog entry it received. Since the last oplog entry Meteor received might have been up to 10 minutes ago, there could be a large number of other oplog entries mongo would have to filter.

2. Would it be possible to make this automatic, at least for the databases? Meteor could track which databases have had observers created for them, and update the query whenever a new database is observed. It seems like as long as Meteor tracks the timestamp of the last oplog entry it processed, there shouldn't be any downside to restarting tailing with a new query. Meteor already does it frequently with the same query.

This would probably involve OplogHandle tracking the last ts it's seen: https://github.com/meteor/meteor/blob/c26569d224eb081b9aef950f44066c555ca9b6f9/packages/mongo/oplog_tailing.js#L256
and then stopping and recreating the oplog tail handle as needed. I'm not sure if there would be a performance issue with listing too many databases or collections.

3. What happens if someone creates an observer on an excluded database or collection? Will the observer not provide updates after the initial find, or will it fallback to polling?

Twisterking:
@zodern 
Thanks for your reply! I indeed also think that this change is actually quite simple, but can have a huge impact in some cases like ours. It **really** changes the performance/scaling of Meteor for us at Orderlion significantly!

ad 1) good point indeed, to be honest I don't know. 
ad 2) same here. Maybe I can ask you of any other maintainer to help here a bit? :)
ad 3) I thought about this too. Also, don't know yet. I am happy to test it! In my opinion both is fine, as long as these 2 new options are properly documented. It is really meant for collections where there are no subscriptions whatsoever. But it makes sense that maybe polling of course is better than nothing.

Is there anyone else who is willing to contribute here? I might be biased of course but I honestly believe this change can help with some Meteor scaling issues a lot for some projects out there!

Twisterking:
Okay, so I finally managed to make the tests work! 🎉 
Also, I implemented a `console.warn` whenever a subscription on an excluded or not included collection is created.

Please have a look!

zodern:
> ad 1) good point indeed, to be honest I don't know.
ad 2) same here. Maybe I can ask you of any other maintainer to help here a bit? :)
ad 3) I thought about this too. Also, don't know yet. I am happy to test it! In my opinion both is fine, as long as these 2 new options are properly documented. It is really meant for collections where there are no subscriptions whatsoever. But it makes sense that maybe polling of course is better than nothing.

For 1), it might be worth mentioning in the docs that it would be inefficient if the watched collections are rarely updated, and other collections have a very large number of updates.

2) I might create a package to experiment with this and some other ideas

3) There are quite a few scenarios where Meteor is unable to use the oplog for an observer, and Meteor always falls back to polling in those situations. For consistency, it would be good if Meteor did the same thing here. Throwing an error or logging a warning might also work, though it should be done when creating the observer instead of in `_publishCursor` to handle all of the ways an observer can be created. The relevant function for either solution (falling back to polling, or throwing an error) would probably be https://github.com/meteor/meteor/blob/c26569d224eb081b9aef950f44066c555ca9b6f9/packages/mongo/mongo_driver.js#L1352

rj-david:
On the discussion on where to place the configuration, in terms of developer experience, is it possible to do this in the definition of the collection (also where you normally add the schema). So when a developer creates a new collection, the developer can configure that collection for oplog tailing or not on the same file. Moving it outside the creation of the collection will result for most developers to forget to disable/enable oplog tailing.

Just an idea and of course if it is even possible

Twisterking:
@zodern: Thank you, I will have a look into `_observeChanges` today!
@rj-david: to be honest I think this is too late in the lifecycle! I thought about this approach too, but whenever you create collections, the Oplog handle/cursor has already been created. So this approach won't work afaik. That's why, again, I went for the `Meteor.settings` approach in the end.

Twisterking:
@zodern @radekmie 
Does anyone of the Meteor team have infos, when this PR might be merged and released? 😃 
We are really desperately waiting for its release! ✌️

harryadel:
While I'm not part of the core team but if @zodern @radekmie can give the green light since they're responsible for reviewing, @denihs @Grubba27 can move in.

docs/source/api/collections.md
```diff
@@ -1082,6 +1082,38 @@ option:
 You can pass any MongoDB valid option, these are just examples using
 certificates configurations.
 
+<h3 id="mongo_oplog_options">Mongo Oplog Options</h3>
+
+> Oplog options were introduced in Meteor 2.15.1
+
+If you set the [`MONGO_OPLOG_URL`](https://docs.meteor.com/environment-variables.html#MONGO-OPLOG-URL) env var, Meteor will use MongoDb's Oplog to show efficient, real time updates to your users via your subscriptions.
+
+Due to how Meteor's Oplog implementation is built behind the scenes, if you have certain collections where you expect **big amounts of `insert` or `update` transactions**, this might lead to **big CPU spikes on your meteor app server, even if you have no publications/subscriptions on any data/documents of these collections**. For more information on this, please have a look into [this blog post from 2016](https://blog.meteor.com/tuning-meteor-mongo-livedata-for-scalability-13fe9deb8908), [this github discussion from 2022](https://github.com/meteor/meteor/discussions/11842) or [this meteor forums post from 2023](https://forums.meteor.com/t/cpu-spikes-due-to-oplog-updates-without-subscriptions/60028).
```
radekmie, line 1091: 1. The wording here is unnecessarily pessimistic. Ideally @Grubba27 could rephrase it.
2. I'd rather go with "operations", not "transactions", as the latter may suggest actual database transactions (which is irrelevant here). It could be also generalized to "write operations" too.

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,116 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
+let defaultOplogHandle = MongoInternals.defaultRemoteCollectionDriver().mongo._oplogHandle;
+
+async function oplogOptionsTest({
+  test,
+  includeCollectionName,
+  excludeCollectionName,
+  mongoPackageSettings = {}
+}) {
+  if (!Meteor.settings.packages) Meteor.settings.packages = {};
+  if (!Meteor.settings.packages.mongo) Meteor.settings.packages.mongo = mongoPackageSettings;
```
radekmie, line 194: It'd be good to clear those settings after the test not to influence others.

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,116 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
+
+  const myOplogHandle = new MongoInternals.OplogHandle(process.env.MONGO_OPLOG_URL, 'meteor');
+  MongoInternals.defaultRemoteCollectionDriver().mongo._setOplogHandle(myOplogHandle);
```
radekmie, line 197: Same goes for the `OplogHandle`.

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,116 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
+    const collectionNameB = "oplog-b-" + Random.id();
+    const mongoPackageSettings = {
+      oplogIncludeCollections: [collectionNameA],
+      oplogExcludeCollections: [collectionNameB]
+    };
+    try {
+      await oplogOptionsTest({
+        test,
+        includeCollectionName: collectionNameA,
+        oplogExcludeCollections: collectionNameB,
```
radekmie, line 279: Typo?

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,116 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
+        mongoPackageSettings
+      });
+      // reset the package settings:
+      delete Meteor.settings.packages.mongo;
+      MongoInternals.defaultRemoteCollectionDriver().mongo._setOplogHandle(defaultOplogHandle);
```
radekmie, line 284: As I wrote above - this should be handled in `oplogOptionsTest`.

packages/mongo/package.js
```diff
@@ -100,7 +100,7 @@ Package.onTest(function (api) {
   api.addFiles('collection_tests.js', ['client', 'server']);
   api.addFiles('collection_async_tests.js', ['client', 'server']);
   api.addFiles('observe_changes_tests.js', ['client', 'server']);
-  api.addFiles('oplog_tests.js', 'server');
   api.addFiles('oplog_v2_converter_tests.js', 'server');
   api.addFiles('doc_fetcher_tests.js', 'server');
+  api.addFiles('oplog_tests.js', 'server');
```
radekmie, line 105: It should not be needed if the `oplogOptionsTest` would clear after itself.

radekmie (changes requested):

radekmie:
@zodern 
> 1. Meteor tracks the last oplog entry it's seen to know where to resume tailing the oplog. If there's been a large number of newer oplog entries for databases or collections that are excluded, could resume tailing cause performance issues if mongo has to filter out the unneeded docs? This happens every time Meteor does not receive an oplog doc for 30 seconds, there is a failover, or in some other situations.

It may be a visibly long operation, but it should not be an issue - oplog collection is usually already in memory, so it'll be really fast to grep through.

> 2. Would it be possible to make this automatic, at least for the databases? Meteor could track which databases have had observers created for them, and update the query whenever a new database is observed. It seems like as long as Meteor tracks the timestamp of the last oplog entry it processed, there shouldn't be any downside to restarting tailing with a new query. Meteor already does it frequently with the same query.

It could, sure. I'd rather make it in a separate step to get this PR out sooner.

packages/mongo/oplog_tests.js
Twisterking: I am actually doing just that?! Please just look at the very bottom of `oplogOptionsTest()` :)

packages/mongo/oplog_tests.js
Twisterking: I can only reset the used `OplogHandle` like so, but I don't think there is any way to "destroy" `myOplogHandle`, is there?!

Reset:
```js
MongoInternals.defaultRemoteCollectionDriver().mongo._setOplogHandle(defaultOplogHandle);
```

packages/mongo/oplog_tests.js
Twisterking: different in this case, because for this test `oplogOptionsTest()` actually throws an error because of using both options - like it should! That's why here the teardown actually happens outside of `oplogOptionsTest()`!

packages/mongo/package.js
Twisterking: But it does not hurt either! ;)

Twisterking:
@radekmie thanks for your feedback! PR updated!

packages/mongo/oplog_tests.js
radekmie: No, you _delete_ it - I suggest _resetting_ it to what it was. I know that right now it will be the same, but it may change in the future.

packages/mongo/oplog_tests.js
radekmie: My point is that you should extract the previous handle before setting the new one and set the old one back after the test.

packages/mongo/oplog_tests.js
radekmie: That only suggests that the cleanup should happen in a `finally` block in `oplogOptionsTest` - then it'll happen even if the test fails.

packages/mongo/package.js
radekmie: That's a matter of personal preference - I suggest not to touch anything not related to the PR to make it easier to review and then easier to compare (or bisect if needed).

packages/mongo/oplog_tests.js
Twisterking: That is exactly what I am doing!

radekmie: No, you're extracting it on the top level, not in the test. That's a significant difference, as the top-level code executes first only to register the tests.

_(Also, I don't know how to understand your exclamation marks... I hope you're excited and not getting angry at me.)_

packages/mongo/package.js
Twisterking: okay, I will move it again!

Twisterking: Fixed!

packages/mongo/oplog_tests.js
Twisterking: Also fixed!

Twisterking:
@radekmie one more update done! Thanks again for your feedback!

packages/mongo/oplog_tests.js
```diff
@@ -180,6 +182,118 @@ process.env.MONGO_OPLOG_URL && Tinytest.addAsync(
...
+
+    // do the inserts:
+    await IncludeCollection.rawCollection().insertOne({ include: 'yes', foo: 'bar' });
+    await ExcludeCollection.rawCollection().insertOne({ include: 'no', foo: 'bar' });
+
+    test.equal(await shouldBeTracked, true);
+    test.equal(await shouldBeIgnored, true);
+  } catch(err) {
+    // we just have this catch to be able to use the finally! ;)
+    throw err;
```
radekmie, line 231: Nope, we don't 😛 
```suggestion
```

packages/mongo/collection_tests.js
```diff
@@ -210,7 +210,7 @@ Tinytest.add('collection - calling find with an invalid readPreference',
 Tinytest.add('collection - inserting a document with a binary should return a document with a binary',
   function(test) {
     if (Meteor.isServer) {
-      const collection = new Mongo.Collection('testBinary1');
+      const collection = new Mongo.Collection('testBinary1' + test.id);
```
radekmie, line 213: Why are those suddenly needed?

packages/mongo/oplog_tests.js
Twisterking: Then you please have to help me! Otherwise I can't make the last test work, where I want to test that it fails if both settings are used at the same time. If I catch the error here, then I can't "listen" for an error in the actual test!

radekmie: I meant you can just `try { ... } finally { ... }`. Is there a problem with that?

Twisterking: Allllright, my bad 😁 you are right, `try { ... } finally { ... }` works just fine! ✌️

radekmie (approved):

Twisterking:
Hello everyone! Thanks for approving the PR @radekmie !
Any news on a possible release date? As stated, this is really blocking us more and more! :(

StorytellerCZ:
I would be for Meteor 2.16 with https://github.com/meteor/meteor/pull/12436 asap.

Twisterking:
> > Hello everyone! Thanks for approving the PR @radekmie ! Any news on a possible release date? As stated, this is really blocking us more and more! :(
> 
> You can already use this without waiting for a new release since the changes are in the package and not in the core

How? The version of the package was not released yet, no? Sorry, don't fully understand how this works! Also, I forget to increase the version number of the `mongo` package! my bad!

What exactly do I need to do to use this version of the `mongo` core package?

rj-david:
> What exactly do I need to do to use this version of the `mongo` core package?

https://guide.meteor.com/writing-atmosphere-packages#local-packages

Twisterking:
> > What exactly do I need to do to use this version of the `mongo` core package?
> 
> [guide.meteor.com/writing-atmosphere-packages#local-packages](https://guide.meteor.com/writing-atmosphere-packages#local-packages)

That won't work here! I tried to install my package as a local package! The issue is, that all the other packages still used the "default mongo package" behind the scenes, and not my "custom" local package. Couldn't get it to work when I tried it.

rj-david:
Short of releasing this package, the things that can be tested are:

1. Incrementing the version
2. Also forking those packages that used mongo locally

StorytellerCZ:
If you use meteor from local checkout with this branch that should do the trick if local package override is failing.

Twisterking:
Any updates on when the release of 2.16 will happen? Hopefully incl. this PR please! ✌️

Twisterking:
Updates? Is there a planned release date for 2.16 yet? We really need this change asap ...

nachocodoner:
> Updates? Is there a planned release date for 2.16 yet? We really need this change asap ...

No estimated date yet. First, we're focused on the 3.x RC. Meanwhile, we aim to merge candidates and make quick gains into the 2.16 branch. The PR has been updated [to provide more details about the plan](https://github.com/meteor/meteor/pull/13066#issue-2200350442). We'll announce a date as soon as we have one.

jeffreyflynt:
> > > What exactly do I need to do to use this version of the `mongo` core package?
> > 
> > 
> > [guide.meteor.com/writing-atmosphere-packages#local-packages](https://guide.meteor.com/writing-atmosphere-packages#local-packages)
> 
> That won't work here! I tried to install my package as a local package! The issue is, that all the other packages still used the "default mongo package" behind the scenes, and not my "custom" local package. Couldn't get it to work when I tried it.

For future reference, I ran into this issue, what I had to do was name my local package a, so it's loaded before everything else.

For instance, I have this configured for Mongo:

```
const secureContext = tls.createSecureContext({
    ca: TLS_CA,
    cert: TLS_CRT,
    key: TLS_KEY,
    minVersion: "TLSv1.3",
  });

  process.env.MONGO_URL = MONGO_URL;

  Mongo.setConnectionOptions({
    tls: true,
    secureContext,
    compressors: ["zstd"],
    family: 4,
  });

  ```

**Comment by Colkadome** [comment]
I have a strange behaviour after the update if anyone knows whats going on here.

Pretend I need to update a user property in the `Accounts.onLogin`, and I also have a collection "UserStats" that I update, but I also have included UserStats in the `oplogExcludeCollections` setting (as I don't have any publications that use this collection):

settings.json:
```
{
  "packages": {
    "mongo": {
      "oplogExcludeCollections": [
        "UserStats"
      ]
    }
  }
}
```

The login method takes around ~15 seconds to respond in the websocket connection with this setting on, AND with the following code in the onLogin callback:

```
Accounts.onLogin((data) => {
  Meteor.users.update(...);
  UserStats.update(...);
});
```

Either removing "UserStats" from `oplogExcludeCollections` will fix this, OR by changing the update code to use async calls (both need to be async). I can also include some driver refresh call and it will be fine:

```
Accounts.onLogin(async (data) => {
  
  await Meteor.users.rawCollection().updateOne(...).catch(...);
  Meteor.users._driver.mongo._refresh(Meteor.users._name, { ... });

  await UserStats.rawCollection().updateOne(...).catch(...);
  UserStats._driver.mongo._refresh(UserStats._name, { ... });

});
```

I have solved the issue now. I don't have much time at the moment to test more conditions here, but I thought I would just comment this in case anyone else is experiencing something similar. There may be some steps in the update() calls that I am ignoring with the above code but for my use-case it is fine.

PR #13013, opened 2024-02-20
Add method name to `MethodInvocation`

jamauro:
Small PR to add the method name to the `MethodInvocation`. This allows you to know what method is being called when you use `DDP._CurrentMethodInvocation.get()`. This will be helpful for a package I'm developing.

### DIFF
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -911,6 +911,7 @@ export class Connection {
     };
 
     const invocation = new DDPCommon.MethodInvocation({
+      name,
       isSimulation: true,
       userId: self.userId(),
       isFromCallAsync: options?.isFromCallAsync,
--- a/packages/ddp-common/method_invocation.js
+++ b/packages/ddp-common/method_invocation.js
@@ -18,6 +18,16 @@ DDPCommon.MethodInvocation = class MethodInvocation {
     // zero-latency connection to the user.
 
     /**
+     * @summary The name given to the method.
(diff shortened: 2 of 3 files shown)


Grubba27:
I liked the idea.

Grubba27 (approved):
LGTM

StorytellerCZ (approved):

StorytellerCZ:
I'm thinking that this is great for development and debugging. I was thinking that maybe for production we want to keep security by obscurity and not add it there. Thoughts?

nachocodoner:
> I'm thinking that this is great for development and debugging. I was thinking that maybe for production we want to keep security by obscurity and not add it there. Thoughts?

We may have applied this pattern broadly in many areas, but it doesn't mean we disregard further security concerns, although they might not be as critical. Nonetheless, we now have Meteor.isDebug or Meteor.isDevelopment modes if needed.

@jamauro, do you have any thoughts on this?

jamauro:
Here are my thoughts:

1. Limiting to non-production wouldn't solve the use case I originally had in mind. I was thinking it could be useful so that a developer could chose to prevent reactivity for the client-side caller / avoid a code path in `jam:pub-sub` for a particular method that's expected to modify a large set of documents.
2. I don't think it would pose a security risk. Accessing the method name in this scenario would require sniffing DDP messages. Assuming your app is using https / wss (which should be a requirement), a bad actor shouldn't be able to sniff those messages. 
3. If you're writing isomorphically, the method names are already exposed on the client inside `Meteor.connection`.

Issue #12918, opened 2023-12-13
[Meteor 3] Mongo.Cursor.observeChanges returns a Promise instead of a query handle


ebroder:
With the current alpha release of v3, calling `observeChanges` (at least on the server side - client might be different!) now returns a `Promise<LiveQueryHandle>`. Previously, it returned a `LiveQueryHandle` directly. This means that it's not possible to write code that works correctly on both v2 and v3.

Specifically, `MongoConnection.prototype._observeChanges` is now an [`async` function](https://github.com/meteor/meteor/blob/release/METEOR%403.0-alpha.19/packages/mongo/mongo_driver.js#L1433), causing it to return a Promise).

It seems ideal, if possible, to continue returning a `LiveQueryHandle` directly rather than returning a `Promise`. However, if that's not possible, then I would request renaming `observeChanges` to `observeChangesAsync` and introducing a wrapper in v2 so that there's a clean migration path.

ebroder:
For the record I suspect that there's a similar issue in `observe`, but my code only uses `observeChanges` so I haven't verified that directly.

radekmie:
These are definitely fine on the client (I just checked on [Simple Tasks](https://simpletasks.meteorapp.com)) and there's `isReadyPromise` to `await` if needed, so it's only a server-side problem.

manueltimita:
I just noticed the same issue in [meteor-publish-composite](https://github.com/Meteor-Community-Packages/meteor-publish-composite), which depends on `observe`. I'm about to open an issue there and suggest a pull request

Grubba27:
I just confirmed it; on the server, it does return a promise. The client is fine.

champel:
The existing implementation of `observe` and `observeChanges` does not return until the initial fetch is complete. This enables the triggering of an action once the initial additions have been processed. However, if you solely return the LiveQueryHandler without the promise, it becomes impossible to detect the conclusion of the initial fetch. I would prefer a distinct implementation such as `observeAsync` or `observeChangesAsync` that returns the promise, enabling me to identify the completion of the initial fetch.

champel:
The central concept here is that not returning a promise in version 3 will result in the inability to uphold the existing functionality (which involves tracking the completion of the initial fetch). While the inclusion of observerAsync may not be ideal, it facilitates a gradual migration and aligns with the rest of API modifications.

ebroder:
[...]

PR #13025, opened 2024-02-22
339 Support observeChangesAsync and observeAsync [Meteor 2.x]

nachocodoner:
OSS-339

Context: https://github.com/meteor/meteor/issues/12918

### DIFF
--- a/packages/minimongo/cursor.js
+++ b/packages/minimongo/cursor.js
@@ -217,6 +217,16 @@ export default class Cursor {
   }
 
   /**
+   * @summary observe async version
+   * @locus Anywhere
+   * @memberOf Promise<Mongo.Cursor>
+   * @instance
+   */
+  async observeAsync(options) {
+    return new Promise(resolve => resolve(this.observe(options)));
+  }
+
+  /**
    * @summary Watch a query. Receive callbacks as the result set changes. Only
    *          the differences between the old and new documents are passed to
    *          the callbacks.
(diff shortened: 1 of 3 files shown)


packages/minimongo/cursor.js
```diff
@@ -216,6 +216,16 @@ export default class Cursor {
...
   }
 
+  /**
+   * @summary observe async version
+   * @locus Anywhere
+   * @memberOf Promise<Mongo.Cursor>
+   * @instance
+   */
+  observeAsync(options) {
+    return new Promise(resolve => resolve(this.observe(options)));
```
radekmie, line 226: It's fine, though `Promise.resolve(this.observe(options))` would be simpler. (I assume `observe` cannot throw; if it can, then it has to stay like this.) Alternatively, this function could be `async`.

radekmie (approved):

packages/minimongo/cursor.js
nachocodoner: I believe using `new Promise` instead of `Promise.resolve` in the `observe` or `observeChanges` could automatically catch any potential errors, which I don't know exactly if currently they are. Anyway, the `new Promise` pattern seems more adaptable and better for being ready to handle future logic adjustments that may throw errors.

Regarding `async`, I initially included it, but later realized it's necessary only when using `await`. When simply returning a `Promise`, it's not mandatory. This seems more like a code convenience, as other sections of the codebase omit the mention of async when it's not necessary I just removed it from this code.

Grubba27 (approved):

PR #13040, opened 2024-02-29
Support for synchronous method calls from the client

plbkbx:
This fixes a backward compatibility issue when you manage a remote collection on the client side using synchronous methods (e.g. collection.insert(...), collection.update(...)).

For more details see [meteor forum](https://forums.meteor.com/t/error-in-collection-update-call-from-client-in-version-3/60553)


denihs:
This doesn't seem necessary anymore, but we can revisit this topic in the future if required.

PR #13045, opened 2024-03-01
Set minPoolSize for oplog

StorytellerCZ:
Oplog has its specific startup, but it was missing a condition for `minPoolSize` which made it impossible to set `minPoolSize` for normal connection as setting it in the `mongo` settings would set it as well for oplog connection which would result in error as `maxPoolSize` for oplog is `1`. So setting `minPoolSize` to 30 would result in failed startup on Galaxy when the oplog was started as `minPoolSize` for oplog could not be more than `maxPoolSize` (which is 1). So by setting `minPoolSize` for oplog to 1 and adding a copy of if function in startup for oplog connection for `minPoolSize` as well should avoid that error.

I think this should be released asap as a bug fix as Atlas has recommended me to work on the connection pool sizes in my connection, so more people are likely to work on these.

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -172,6 +172,9 @@ MongoConnection = function (url, options) {
     // set it for replSet, it will be ignored if we're not using a replSet.
     mongoOptions.maxPoolSize = options.maxPoolSize;
   }
+  if (_.has(options, 'minPoolSize')) {
+    mongoOptions.minPoolSize = options.minPoolSize;
+  }
 
   // Transform options like "tlsCAFileAsset": "filename.pem" into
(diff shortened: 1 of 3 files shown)


nachocodoner (approved):

StorytellerCZ:
Tested locally and in a deployment to Galaxy. Having this fix allowed me to deploy with setting the max and min pool sizes which was impossible before.

Grubba27 (approved):

StorytellerCZ:
@Grubba27 @nachocodoner can we please release this in a patch release asap. Missing this is making my deploys really annoying.

Grubba27:
Screenshot 2024-03-12 at 12 03 17
Published mongo@1.16.9.

PR #13046, opened 2024-03-03
Session storage option for accounts

StorytellerCZ:
There has been a discussion on where to store login credentials since the accounts packages have been created. Meteor stores login token and so on in local storage. Some have suggested to move to sessions storage which is not good for most applications. Upon a request from my sponsor I have added an option to switch to using of session storage for applications that need it. This can be done by setting new option `useSessionStorage` or setting in Meteor public settings (naming to be discussed) to true.
I have also added a test, but I'm not certain that it is done correctly, so here I would appreciate some guidance and thorough review. I have tested it on my applications and it worked as intended.

I did not use the `session` package as today browsers fully support session storage and here we are just storing string and nothing complicated.

---------------
This PR has been provided as a perk of my [sponsorship package](https://github.com/sponsors/StorytellerCZ/) for one of my sponsors. If you like what I'm doing, please [sponsor me](https://github.com/sponsors/StorytellerCZ/) and consider other Meteor contributors for sponsorship as well.

### DIFF
--- a/packages/accounts-base/accounts-base.d.ts
+++ b/packages/accounts-base/accounts-base.d.ts
@@ -62,12 +62,20 @@ export namespace Accounts {
     sendVerificationEmail?: boolean | undefined;
     forbidClientAccountCreation?: boolean | undefined;
     restrictCreationByEmailDomain?: string | Function | undefined;
+    loginExpiration?: number | undefined;
     loginExpirationInDays?: number | undefined;
     oauthSecretKey?: string | undefined;
+    passwordResetTokenExpiration?: number | undefined;
     passwordResetTokenExpirationInDays?: number | undefined;
+    passwordEnrollTokenExpiration?: number | undefined;
     passwordEnrollTokenExpirationInDays?: number | undefined;
     ambiguousErrorMessages?: boolean | undefined;
+    bcryptRounds?: number | undefined;
     defaultFieldSelector?: { [key: string]: 0 | 1 } | undefined;
+    collection?: string | undefined;
+    loginTokenExpirationHours?: number | undefined;
+    tokenSequenceLength?: number | undefined;
+    clientStorage?: 'session' | 'local';
   }): void;
 
   function onLogin(
--- a/packages/accounts-base/accounts_client_tests.js
+++ b/packages/accounts-base/accounts_client_tests.js
@@ -304,3 +304,26 @@ Tinytest.addAsync(
   }
 );
 
+Tinytest.addAsync('accounts - Session storage', async (test, done) => {
+
+  Accounts.config({ useSessionStorage: true })
+  test.isTrue(Accounts._options.useSessionStorage)
+
+  // Login a user with AccountClientSession and test that tokens are in sessionStorage
+  logoutAndCreateUser(test, done, () => {
+    test.isNotUndefined(sessionStorage.getItem('Meteor.loginToken'))
+    test.isUndefined(localStorage.getItem('Meteor.loginToken'))
+    Accounts.logout()
+    removeTestUser()
+  })
+
+  // Login a user with AccountClientStorage and test that tokens are in localStorage
+  logoutAndCreateUser(test, done, () => {
+    Accounts.config({ useSessionStorage: false })
+    test.isFalse(Accounts._options.useSessionStorage)
+    test.isUndefined(sessionStorage.getItem('Meteor.loginToken'))
+    test.isNotUndefined(localStorage.getItem('Meteor.loginToken'))
+    Accounts.logout()
+    removeTestUser()
+  })
+})
--- a/packages/accounts-base/accounts_client_tests.js
+++ b/packages/accounts-base/accounts_client_tests.js
@@ -304,26 +304,21 @@ Tinytest.addAsync(
   }
 );
 
-Tinytest.addAsync('accounts - Session storage', async (test, done) => {
-
-  Accounts.config({ useSessionStorage: true })
-  test.isTrue(Accounts._options.useSessionStorage)
-
-  // Login a user with AccountClientSession and test that tokens are in sessionStorage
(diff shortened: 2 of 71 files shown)


packages/accounts-base/accounts_client_tests.js
```diff
@@ -304,3 +304,26 @@ Tinytest.addAsync(
...
+  // Login a user with AccountClientStorage and test that tokens are in localStorage
+  logoutAndCreateUser(test, done, () => {
+    Accounts.config({ useSessionStorage: false })
+    test.isFalse(Accounts._options.useSessionStorage)
+    test.isUndefined(sessionStorage.getItem('Meteor.loginToken'))
+    test.isNotUndefined(localStorage.getItem('Meteor.loginToken'))
+    Accounts.logout()
+    removeTestUser()
+  })
+})
```
nachocodoner, line 329: Here are some suggested improvements:

- The `done` callback in the test definition controls test finalization. Using `done` twice per test might cause only one to evaluate successfully during the async process.
  - Possible fixes:
    - Nest them, ensuring verification of session storage occurs after local storage.
    - Utilize `testAsyncMulti` and create two test cases: one for `useSessionStorage: true` and the other for `useSessionStorage: false` or default behavior. This approach appears cleaner. See example: https://github.com/meteor/meteor/blob/release-2.16/packages/accounts-password/email_tests.js#L11-L70 (be aware of expect usage in the example)
- Consider renaming the test to something inclusive of both storage types, such as "Accounts - Local and Session Storage."
- While `useSessionStorage` seems appropriate, note that the `useXX` convention is widely adopted in React for hooks. However, this likely isn't a significant issue. I'm fine with it.

StorytellerCZ: For the `useSessionStorage`, the alternative I can think of would be to something like `storage: 'session'` and `storage: 'local'`.

nachocodoner: I prefer those actually. Anyway, can we also act on the types for autocompletion? `accounts-base.d.ts` should be.

StorytellerCZ: Done and in the end using `clientStorage` for more flexibility.

nachocodoner: Something is missing, the tests fail.

image

StorytellerCZ: Fixed the issue, but have another in the tests. Either I somehow need to reset the server with new config or maybe change the order to have the default first (and not setting it via options).

nachocodoner: https://github.com/meteor/meteor/blob/b209f24a7896dc72b186003f199b2c100ae159f7/packages/accounts-base/accounts_common.js#L281

Maybe we can disable this check on tests.

packages/accounts-base/accounts-base.d.ts
```diff
@@ -75,6 +75,7 @@ export namespace Accounts {
     collection?: string | undefined;
     loginTokenExpirationHours?: number | undefined;
     tokenSequenceLength?: number | undefined;
+    clientStorage?: 'session' | 'local';
```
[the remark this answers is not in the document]
StorytellerCZ: A type literal property cannot have an initializer.

packages/accounts-base/accounts_client_tests.js
StorytellerCZ: If that is fine with you. I'm going to try this in the next commit.

StorytellerCZ: Or I can make even more strict restriction:
```javascript
if (key !== 'collection' && (Meteor.isTest && key !== 'clientStorage')) {
            throw new Meteor.Error(`Can't set \`${key}\` more than once`);
          }
```

StorytellerCZ: That did not seem to work :(

nachocodoner:
@StorytellerCZ: Any progress on the green checks here?

We plan to release this week and this may be a quick gain. I will focus first on validate best candidates and others with green PRs as stated here, https://github.com/meteor/meteor/pull/13066

StorytellerCZ:
I will try to look into this today. If I can't get it to work, then I will revert the changes to how they were before I moved to the async tests that you recommended.

nachocodoner:
After verifying and merging the remaining candidates from [2.16](https://github.com/meteor/meteor/pull/13066) into [3.0](https://github.com/meteor/meteor/pull/13098), I now have some time to further review this change.

I must note several test failures, even from the initial version of this PR, triggering false positives.

- In Accounts.config, we cannot reset `this.storageLocation`, which is part of the [AccountsClient constructor](https://github.com/meteor/meteor/blob/aa5c565c4d475b033b089ff4e56ac259e16fb81b/packages/accounts-base/accounts_client.js#L30-L31). It initializes [on first run](https://github.com/meteor/meteor/blob/feature/account-session-switch/packages/accounts-base/client_main.js#L10) and doesn't reassign, so the test doesn't effectively change the storage to act. I have provided the necessary changes to properly reconfigure it.
- Running these storage tests can't login in parallel as it can lead to race condition issues. Properly awaiting actions and verifications sequentially is essential.

I have made the necessary changes to address the above issues and refactor the tests to be more expressive. Please review.

@StorytellerCZ: Once test pass and the PR is approved, I will proceed with merging it into release 2.16 and measuring the impact on 3.0.

packages/non-core/blaze
nachocodoner, line 1: I'm worry about this state of the blaze to have been updated here. We have not changed anything on that library, right?

StorytellerCZ: I think there has been a Blaze release since this has been updated, unless this has been updated somewhere else.

StorytellerCZ:
Perfect! I have nothing to add here. Please merge.

Issue #13036, opened 2024-02-27
[3.0] insertAsync on client won't persist local document


bhunjadi:
**Description / reproduction**

Please check this failing test: https://github.com/bhunjadi/meteor/blob/795e61aff0c191257937c07eccad7438b2394de4/packages/mongo/allow_tests.js#L1164

Run the test with:
```bash
./meteor test-packages mongo
```
After insertAsync, client doesn't have the document in local collection anymore.

I first observed this when updating meteor-collection-hooks for Meteor 3.0. [Here](https://github.com/bhunjadi/meteor-collection-hooks/blob/e0c4b965724e532f724db361e6ee1fb5844a4fb5/tests/direct.js#L151) is the failing test.

Notes:
- Server-side test is working.
- If you leave out removeAsync, it still fails
- I'm getting this error in the console: "Method stub (/test_again/insertAsync) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations"

Debugging:

I noticed that local document exists briefly and it gets deleted.
It seems to me this is because additional event is triggered which removes the document.

What happens is that collection.update method is called with `replace` msg.
If you look in [collection.js](https://github.com/bhunjadi/meteor/blob/795e61aff0c191257937c07eccad7438b2394de4/packages/mongo/collection.js#L244) the `replace` value will be undefined, which will trigger the removal of the document couple of lines below.

```js
 if (msg.msg === 'replace') {
  var replace = msg.replace; // will be undefined
  if (!replace) {
    // we'll end up here
    if (doc) self._collection.remove(mongoId);
  } else if (!doc) {
    self._collection.insert(replace);
  } else {
    // XXX check that replace has no $ ops
    self._collection.update(mongoId, replace);
  }
  return;
}
```

Why it is undefined? Because of [this line](https://github.com/bhunjadi/meteor/blob/795e61aff0c191257937c07eccad7438b2394de4/packages/minimongo/local_collection.js#L122) in minimongo/local_collection. 

```js
prepareInsert(doc) {
    ...
    // Here we only set id, but without document. 
    this._saveOriginal(id, undefined);
    this._docs.set(id, doc);

    return id;
  }
```

I also observed that adding resolverType: 'stub' fixes the test.
```js
const coll = new Mongo.Collection('test_again', {
  resolverType: 'stub'
});
```

**Version**
Forked `release-3.0` repo

**Last version where the problem did not occur**
n/a (there was no insertAsync on client)

**OS**
[...]

PR #13052, opened 2024-03-05
[3.0] Persist local doc in the same way Meteor 2.x

nachocodoner:
OSS-350

# Context

Issue: https://github.com/meteor/meteor/issues/13036

A continuation of https://github.com/meteor/meteor/pull/12969, which provided a configuration to sort one of the problems on keeping the same behavior on running collection operations on Meteor 3.x in respect of Meteor 2.x. That fix resulted to be like a workaround, and hopefully with this new iteration is completely fixed.

## Reproduction

``` javascript
    
console.log("-> isItemCreated?", NamedCollection.find().fetch()?.[0]);
setTimeout(() => {
    console.log("-> Later: isItemCreated?", NamedCollection.find().fetch()?.[0]);
}, 1_000);
```

image

Thanks to @bhunjadi I was able to understand and debug further the core flow about this issue and realized how Meteor 2.x and 3.x behaves distinctly in the process to run a collection operation.

In Meteor 2.x:
- Run `insertAsync`
- The document is available on the same run cycle
- The document is removed on a later run cycle due to the `replace: undefined` message

In Meteor 3.x:
- Run `insertAsync`
- The document is removed on a later run cycle due to the `replace: undefined` message
- The document **IS NOT** available on the same run cycle

This PR aims to ensure 3.x behavior matches the 2.x behavior. Data won't be persisted as the original github issue states (https://github.com/meteor/meteor/issues/13036), that is not the behavior on Meteor 2.x . For that, in Meteor, you need to fetch via a subscription, a method or directly use the local collection for inserting. Anyway, matching this behavior in both versions will also fix all tests since they expect the data to be available in the same run cycle that the op happened.

## Fix

image

### DIFF
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -850,7 +850,17 @@ export class Connection {
     }
 
     // If we added it to the first block, send it out now.
-    if (self._outstandingMethodBlocks.length === 1) methodInvoker.sendMessage();
+    if (self._outstandingMethodBlocks.length === 1) {
+      if (callback && future) {
+        // Ensure the method message after the result of the method ran in the client.
+        future = new Promise((resolve) => {
+          methodInvoker.sendMessage();
+          resolve();
+        });
+      } else {
+        methodInvoker.sendMessage();
+      }
+    }
 
     // If we're using the default callback on the server,
     // block waiting for the result.


packages/ddp-client/common/livedata_connection.js
```diff
@@ -850,7 +850,17 @@ export class Connection {
     }
 
     // If we added it to the first block, send it out now.
-    if (self._outstandingMethodBlocks.length === 1) methodInvoker.sendMessage();
+    if (self._outstandingMethodBlocks.length === 1) {
+      if (callback && future) {
+        // Ensure the method message after the result of the method ran in the client.
+        future = new Promise((resolve) => {
```
denihs, line 856: Here you're replacing [this future](https://github.com/meteor/meteor/blob/1a7cb656e2f4da7911ca6ed44c6063302f50078d/packages/ddp-client/common/livedata_connection.js#L803-L813) promise. So when I call an applyAsync, callAsync, or any Mongo db function, for that matter, I won't get the return from the server/stub... this doesn't look correct to me.

harryadel:
Hello @nachocodoner, just want to commend you on the careful and detailed explanation you provide in your PRs, along with the annotated screenshots. What a great way to share and spread knowledge! Please keep on doing it :clap: :clap:

nachocodoner:
@harryadel: Appreciate it! 😊

I try to do the best to comprehend the core and complexities of every component I engage with, consistently sharing my insights with the community in the form of text and images.

Regarding this complex issue, after thorough exploration and team discussions, it appears challenging. Considering the inherent async nature in Meteor 3.x, closing this PR might be necessary. Exploring alternatives like `resolveType` config, and employing `stubPromise` and `serverPromise`, seems imperative.

I'll provide a detailed explanation, ensuring the community is informed about this shift in approaching async methods and operations.

nachocodoner:
I'd like to clarify the differences observed between the behaviors of Meteor 2.x and 3.x versions through examples. The evolution in handling asynchronous operations in 3.x requires adjustments in client and server-side executions.

I'll illustrate this with practical scenarios, including ensuring code isomorphism, particularly in testing.

Given the following Meteor method and collection described on client and server.

```
const Greetings = new Mongo.Collection('greetings');

Meteor.methods({
  greetUser: function(name) {
	Greetings.insert({ doc: { userId: Meteor.userId(), name } });
    return "Hello, " + name + "!";
  }
});
```


## Meteor 2.x

In Meteor 2.x,  when calling the method with `call`:

```
Meteor.call('greetUser', 'John', function(error, result) {
  if (error) {
    console.error("Error:", error.reason); // 🔴 Server ended with error
  } else {
    console.log("Result:", result); // 🟢 Server ended with success
  }

  Greetings.findOne({ name: 'John' }); // 🗑️ Data is NOT available
});

// 🔵 Client simulation
Greetings.findOne({ name: 'John' }); // 🧾 Data is available (Optimistic-UI)

```

In this example above:

- The Method runs on the server, while the client (🔵) acts on simulated data.
- Upon Method completion, whether successful (🟢) or encountering an error (🔴), the client callback takes over. However, the data is no longer available in the collection and must be fetched via publication or manually inserted.

## Meteor 3.x

In Meteor 3.x, when calling the method with `callAsync`, it returns `stubPromise` and `serverPromise` to manage client-side simulation and server response, respectively. By default, the server promise resolves as explained below.

### serverPromise

#### Default / serverPromise (await)

```
try {
	await Meteor.callAsync('greetUser', 'John');
	// 🟢 Server ended with success
} catch(e) {
	console.error("Error:", error.reason); // 🔴 Server ended with error
}

Greetings.findOne({ name: 'John' }); // 🗑️ Data is NOT available
```

#### Default / serverPromise (then/catch)

```
await Meteor.callAsync('greetUser', 'John')
  .then(result => {
	console.log("Result:", result); // 🟢 Server ended with success
  })
  .catch(error => {
	console.error("Error:", error.reason); // 🔴 Server ended with error
  });

Greetings.findOne({ name: 'John' }); // 🗑️ Data is NOT available
```

In the examples above the client (🔵) never simulates the behavior and instead the server awaits for its result.

However, when handling `stubPromise` :

### stubPromise

#### stubPromise (await)

```
await Meteor.callAsync('greetUser', 'John').stubPromise;

// 🔵 Client simulation
Greetings.findOne({ name: 'John' }); // 🧾 Data is available (Optimistic-UI)
```

This time, we only manage simulation time, not the server. Client errors can be handled with try/catch.

### stubPromise and serverPromise

Another example using both promises.

```
cosnt { stubPromise, serverPromise } = await Meteor.callAsync('greetUser', 'John');

await stubPromise;

// 🔵 Client simulation
Greetings.findOne({ name: 'John' }); // 🧾 Data is available (Optimistic-UI)

try {
	await serverPromise;
	// 🟢 Server ended with success
} catch(e) {
	console.error("Error:", error.reason); // 🔴 Server ended with error
}
Greetings.findOne({ name: 'John' }); // 🗑️ Data is NOT available
```

In the example above, we could have handled any client error with try/catch as well.

This demonstrates the approach of async/await in Meteor 3.x and how it replicates previous callback and fibers-based behaviors.

### resolveType

Additionally, to enforce the use of `stubPromise` by default, particularly useful on test environments, `resolveType` can be set as `stub` or `server` within the collection configuration. At the end running collection operations means executing a method when using remote collections.

```
 const Greetings = new Meteor.Collection('greetUser', { resolverType: 'stub' });
    
await Greetings.insertAsync({ test: 1 });

// 🔵 Client simulation
Greetings.findOne({ name: 'John' }); // 🧾 Data is available (Optimistic-UI)
```

More details refer to [the original PR](https://github.com/meteor/meteor/pull/12969).

## Conclusion

We can't resolve the issue outlined in this PR because it aligns with the expected behavior of the new async approach using promises (3.x), which differs from when we utilized non-standard fibers (2.x).

I'll soon close the PR and keep this post for future reference.

filipenevola:
Great explanation @nachocodoner.

I think the content of this entire PR discussion should be added to this [page](https://docs.meteor.com/api/meteor.html#methods) in the docs.

Newcomers can't learn this from a PR, and I don't believe it's useful only for experienced Meteor users. Even if it was, it should still be documented in the official docs 🥁.

PR #13053, opened 2024-03-05
[2.x] Report and extend test cases for the old async behaviors

nachocodoner:
This PR aims to add test cases to track the behaviors around data persistence resulted on run collection operations in 2.x to help to understand and match 3.x later.

These tests will be added in this https://github.com/meteor/meteor/pull/13052 3.x branch to also ensure the same exact API is perserved on the next version.

As mentioned on 3.x PR, data is temporary persisted when using collection operations, and the only way to persist them is through using local collection operations (or publications which actually use these). As part of this PR I also added aync versions of local collection operations, since these are enabled in 3.x and facilitates the same test form for the two versions, and allow any migration in 2.x.

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -172,6 +172,9 @@ MongoConnection = function (url, options) {
     // set it for replSet, it will be ignored if we're not using a replSet.
     mongoOptions.maxPoolSize = options.maxPoolSize;
   }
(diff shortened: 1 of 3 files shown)


nachocodoner:
This PR attempts to exemplify on Meteor 2.x the scenarios described [on this comment](https://github.com/meteor/meteor/pull/13052#issuecomment-2007677624). As having minor priority I will work on wrapping up this later.

Issue #12999, opened 2024-02-05
[Meteor 3] Server-only Method displays "Method stub took too long" warning in browser console when using Meteor.callAsync


jamauro:
Version: `Meteor 3.0-beta.0`

I have a server-only Method that I'm calling, i.e. it has no stub, but I'm seeing this warning in the browser console when I use `Meteor.callAsync`. If I use `Meteor.applyAsync`, I don't see the warning.

```js
[Warning] Method stub (updateLink) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations (ddp-client.js, line 245)
```

Expected: I shouldn't see a stub warning if the Method doesn't have a stub. Also I don't think I should see this warning even if it were stubbed.

Here's the Method on the server:

```js
// /server/main.js
import { Links } from '/imports/api/links';

const updateLink = async () => {
  const { _id } = await Links.findOneAsync() || {};
  return Links.updateAsync({ _id }, { $set: { title: 'hello' }});
};

Meteor.methods({ updateLink })
```

```js
// on the client

const updateLink = async () => {
  try {
    await Meteor.callAsync('updateLink');
  } catch (error) {
    console.error(error)
  }
};
```

nachocodoner:
This is going to be fixed here, https://github.com/meteor/meteor/pull/13055

StorytellerCZ:
@nachocodoner can this now be closed?

nachocodoner:
Yes, it is fixed and included on the next beta. Ty

PR #13054, opened 2024-03-06
[3.0] Debug mode

nachocodoner:
OSS-347

This PR spreads the debug mode used on the server via `--inspect*`/`-brk` to the client. Therefore you can ignore certain code execution on the client in a normal execution (`meteor run`), like ignoring log messages.

Specifically, we have introduced the following variables to handle debug-only flows in Meteor code or your app code:

- `__meteor_runtime_config__.debug`
- `Meteor.isDebug`

Besides, and as a practical example, we have ignored in normal app execution the `Method stub took too long` warning message.

### Debug OFF

image

### Debug ON

image

### DIFF
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -107,7 +107,7 @@ export const loadAsyncStubHelpers = () => {
         });
 
         Meteor._setImmediate(() => {
-          if (!finished) {
+          if (!finished && Meteor.isDebug) {
             console.warn(
               `Method stub (${name}) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations`
             );
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -107,7 +107,7 @@ export const loadAsyncStubHelpers = () => {
         });
 
         Meteor._setImmediate(() => {
-          if (!finished && Meteor.isDebug) {
+          if (!finished) {
             console.warn(
               `Method stub (${name}) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations`
             );
--- a/docs/generators/changelog/versions/3.0.md
+++ b/docs/generators/changelog/versions/3.0.md
@@ -667,6 +667,9 @@
 
   - `Meteor.callAsync()`
 
+- `meteor`:
+  - Added `Meteor.isDebug` to execute code in debug builds, activated with the --inspect mode.
+
 - `minifier-css`: (2.9+)
 
   - `CssTools.minifyCssAsync()`
--- a/packages/meteor/debug.js
+++ b/packages/meteor/debug.js
@@ -1,3 +1,19 @@
+if (Meteor.isServer) {
+  if (typeof __meteor_runtime_config__ === 'object') {
+    __meteor_runtime_config__.debug =
+      !!process.env.NODE_INSPECTOR_IPC ||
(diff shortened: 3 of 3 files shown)


denihs (approved):

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -107,7 +107,7 @@ export const loadAsyncStubHelpers = () => {
         });
 
         Meteor._setImmediate(() => {
-          if (!finished) {
+          if (!finished && Meteor.isDebug) {
```
zodern, line 110: This warning is important for developers. If it is logged, it means that it is no longer safe to call methods whenever you want to, and you have to either fix the method stub, or start being careful with using certain Meteor api's like you had to be with Meteor 2 since app/package code could run during an async method stub.

It should be very rarely shown. If it is appearing enough that you want to disable it by default, there might be a bug with when finished it set - it should be set right after (within the same macrotask) of when the method stub finishes.

denihs: Do you mean there could be a bug in this queue where we go to the next before the previous is done? Or is it the problem in the developer's app?

I am not 100%, but I think one way to reproduce this is just by doing the following:

```
Meteor.callAsync("something")
Meteor.callAsync("somethingElse")
```

But this queue should make this be possible, right?

How did make the log show up on your example, @nachocodoner?

zodern: If it happens with every method call, then it is a bug with when `finished` is set to `true`. I'm guessing it is set after it receives the server result instead of after the stub finishes. This also would be a bug with the queue - its purpose is to run one method stub at a time, so there is no need to wait to receive the server result. Waiting defeats the purpose of calling `this.unblock` on the server.

We probably should add a test for this. Maybe something like:
```
// server
let serverEvents = [];
Meteor.methods({
  async unblockedMethod() {
     serverEvents.push('unblock start');
     this.unblock();
     await sleep(200);
     serverEvents.push('unblock end')
  },
  blockingMethod() {
    serverEvents.push('blockingMethod');
  }
});

// client (server methods should also have a no-op client stub)

Meteor.callAsync('unblockedMethod');
Meteor.callAsync('blockingMethod');
// check serverEvents to ensure blockingMethod ran during unblockedMethod
```

jamauro: fwiw, I do think there is something amiss because I've seen the warning quite a bit with `Meteor.callAsync`. Here's one repro: https://github.com/meteor/meteor/issues/12999. I haven't tested it with the latest `beta.6` though.

I've also seen it pop up during `Tinytest` runs.

nachocodoner: For me it appears for the simplest test you can do with a collection operation (a method behind the scenes):

Defined on client and server:
``` javascript
NamedCollection.allow({
  insertAsync() { return true; },
  updateAsync() { return true; },
  removeAsync() { return true; },
});
```

And then run on the client:
``` javascript
```

nachocodoner: I can try to change finished=true as part of stubPromise resolution if you feel is correct, and provide the new tests that supposedly cover this scenario. I will do in another branch though, and the debug mode, even not used yet, can be merged as a new feature.

zodern: > I can try to change finished=true as part of stubPromise resolution if you feel is correct

That is how it should work. It should also run the next queued function at the same time.

nachocodoner: Resolving this converstation. Debug mode will be added as a new future and covered in this PR.

For the other issue mentioned here, refer to this PR, https://github.com/meteor/meteor/pull/13055

PR #13055, opened 2024-03-06
Fix an issue with methods interacting with unblocking mechanisms

nachocodoner:
OSS-352

Context: https://github.com/meteor/meteor/pull/13054#discussion_r1514578297

## Reproduction

Describe the next methods in the server.

``` javascript
let events = [];

Meteor.methods({
  async 'unblockedMethod'({ delay }) {
    events.push('unblock start');
    this.unblock();
    await Meteor._sleepForMs(delay);
    events.push('unblock end');
  },
  'blockingMethod'() {
    events.push('blockingMethod');
  },
});
```

Run on the client:

``` javascript
Meteor.callAsync("unblockedMethod", { delay: 200 }); // unblock + sleep for 20 milliseconds
Meteor.callAsync("blockingMethod"); // run straight + block

// assert 1: `events` should have `["unblock start", "blockingMethod"]` at this point as two methods are executed in parallel and the first one is unblocked. 
// assert 2: `events` should have added `["unblock end"]` at some point in the future as `unblockedMethod` finished
```

Check how the test is described on the PR. A failing scenario that reproduce the issue is depicted in the next picture. Warning logs `Method took too long` also showed up.

image

### Fix

The fix relied on ensure the client queue triggers next method after the stub(simulation) finished for the previous one. This also fixed the logs.

image

### DIFF
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -89,15 +89,19 @@ export const loadAsyncStubHelpers = () => {
 
     return queueFunction(
       (resolve, reject) => {
+        let hasStub = false;
         let finished = false;
 
         Meteor._setImmediate(() => {
           const applyAsyncPromise = oldApplyAsync.apply(this, args);
           stubPromiseResolver(applyAsyncPromise.stubPromise);
           serverPromiseResolver(applyAsyncPromise.serverPromise);
-          applyAsyncPromise.stubPromise.finally(() => {
-            finished = true;
-          });
+          hasStub = !!applyAsyncPromise.stubPromise;
+          if (hasStub) {
+            applyAsyncPromise.stubPromise.finally(() => {
+              finished = true;
+            });
+          }
           applyAsyncPromise
             .then((result) => {
               resolve(result);
@@ -108,7 +112,7 @@ export const loadAsyncStubHelpers = () => {
         });
 
         Meteor._setImmediate(() => {
-          if (!finished) {
+          if (hasStub && !finished) {
             console.warn(
               `Method stub (${name}) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations`
             );
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -95,14 +95,13 @@ export const loadAsyncStubHelpers = () => {
           const applyAsyncPromise = oldApplyAsync.apply(this, args);
           stubPromiseResolver(applyAsyncPromise.stubPromise);
           serverPromiseResolver(applyAsyncPromise.serverPromise);
-          applyAsyncPromise.stubPromise.finally(() => {
-            finished = true;
-          });
           applyAsyncPromise
             .then((result) => {
+              finished = true;
               resolve(result);
             })
             .catch((err) => {
+              finished = true;
               reject(err);
             });
         });
--- a/packages/ddp-client/client/queueStubsHelpers.js
+++ b/packages/ddp-client/client/queueStubsHelpers.js
@@ -20,7 +20,7 @@ export const loadAsyncStubHelpers = () => {
 
     queue = queue.finally(() => {
       fn(resolve, reject);
-      return promise;
+      return promise.stubPromise;
     });
 
     promise.finally(() => {
@@ -95,13 +95,14 @@ export const loadAsyncStubHelpers = () => {
           const applyAsyncPromise = oldApplyAsync.apply(this, args);
           stubPromiseResolver(applyAsyncPromise.stubPromise);
           serverPromiseResolver(applyAsyncPromise.serverPromise);
+          applyAsyncPromise.stubPromise.finally(() => {
+            finished = true;
+          });
           applyAsyncPromise
             .then((result) => {
-              finished = true;
               resolve(result);
             })
             .catch((err) => {
-              finished = true;
               reject(err);
             });
         });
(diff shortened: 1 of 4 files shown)


packages/ddp-client/test/livedata_tests.js
```diff
@@ -1201,6 +1201,32 @@ testAsyncMulti('livedata - methods with nested stubs', [
...
+    });
+  }
+
+  Meteor.callAsync(`unblockedMethod${test.runId()}`);
+  Meteor.callAsync(`blockingMethod${test.runId()}`);
+
+  if (Meteor.isServer) {
+    test.equal(serverEvents, ['unblock start', 'blockingMethod']);
+  }
+});
```
nachocodoner, line 1228: @zodern: Does it seems correct this test for you according on what you commented [here](https://github.com/meteor/meteor/pull/13054#discussion_r1514693752)? 

I ask because even with the current version and doing the finish=true after serverPromise resolution, the test is passing green. Or maybe is a matter on the test itself and how it is run by server and client on distinct times.

zodern: It would probably be easier to write the test in https://github.com/meteor/meteor/tree/059d55d7dd3f8335a87de49d8662df5c28d72814/packages/ddp-client/test/async_stubs

You can see how the other tests there handle queueing some methods, and then checking the server events.

nachocodoner: Done. I will ensure the change doesn't break other tests.

Please review.

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -20,7 +20,7 @@ export const loadAsyncStubHelpers = () => {
 
     queue = queue.finally(() => {
       fn(resolve, reject);
-      return promise;
+      return promise.stubPromise;
```
zodern, line 23: This works since all except the one `fn` are sync, but it seems unexpected for it to not wait on `promise`. It might be better and simpler for `applyAsync` to handle creating its own promise to return, and `queueFunction` can avoid having to deal with `stubPromise`/`serverPromise`.

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -95,13 +95,14 @@ export const loadAsyncStubHelpers = () => {
           const applyAsyncPromise = oldApplyAsync.apply(this, args);
           stubPromiseResolver(applyAsyncPromise.stubPromise);
           serverPromiseResolver(applyAsyncPromise.serverPromise);
+          applyAsyncPromise.stubPromise.finally(() => {
```
denihs, line 98: I think there are cases where we don't have `stubPromise`. It's best to do something like `applyAsyncPromise.stubPromise?.finally` here.

packages/ddp-client/client/queueStubsHelpers.js
nachocodoner: I'm getting familiar with this code. Attempted to refactor this but breaking things along the way. I may retry to refactor in another PR while this PR focus on what tries to fix, which is done and tests passes green now.

packages/ddp-client/test/async_stubs/client.js
```diff
@@ -320,3 +320,33 @@ Tinytest.addAsync(
...
   }
 );
+
+Tinytest.addAsync(
+  "method interaction with unblocking mechanism",
+  async function (test) {
+    await Meteor.callAsync("getAndResetEvents");
+
+    Meteor.callAsync("unblockedMethod", { delay: 200 }); // unblock + sleep for 200 milliseconds
+    Meteor.callAsync("blockingMethod"); // run straight + block
```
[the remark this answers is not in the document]
nachocodoner: Done. Thanks

packages/ddp-client/client/queueStubsHelpers.js
nachocodoner: Added a pre-check of existence.

denihs (approved):

packages/ddp-client/client/queueStubsHelpers.js
nachocodoner: Or maybe I am not understanding well your suggestions here. Could you guide me?

zodern: I haven't tested this code, but something like this in the `applyAsync` wrapper:

```js
    let applyAsyncResolve;
    let applyAsyncReject;
    let stubPromiseResolver;
    let serverPromiseResolver;

    const applyAsyncPromise = new Promise((resolve, reject) => {
      applyAsyncResolve = resolve;
      applyAsyncReject = reject;
    });
    const stubPromise = new Promise((r) => (stubPromiseResolver = r));
    const serverPromise = new Promise((r) => (serverPromiseResolver = r));

    queueFunction(
      (next) => {
        let hasStub = false;
        let finished = false;

        Meteor._setImmediate(() => {
          const promise = oldApplyAsync.apply(this, args);
          stubPromiseResolver(promise.stubPromise);
          serverPromiseResolver(promise.serverPromise);
          hasStub = !!promise.stubPromise;

          if (hasStub) {
            promise.stubPromise.finally(() => {
              finished = true;
              next();
            });
          } else {
            next();
          }

          promise
            .then((result) => {
              applyAsyncResolve(result);
            })
            .catch((err) => {
              applyAsyncReject(err);
            });
        });

        Meteor._setImmediate(() => {
          if (hasStub && !finished) {
            console.warn(
              `Method stub (${name}) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations`
            );
          }
        });
      }
    );

    return applyAsyncPromise;
```

Basically, instead of returning the promise from queueFunction, have applyAsync create a separate promise that is returned. This fixes the api for queueFunction by allowing the resolve function given to the callback to be used to control the queue.

Originally, applyAsync used the same promise as queueFunction since Meteor 2 used a promise for the stub and a callback for the server result. However, now that Meteor 3 uses the returned function for the server result, reusing the promise doesn't make as much sense anymore.

PR #13059, opened 2024-03-14
[3.0] Fix Meteor installer

nachocodoner:
OSS-357

More details in this [comment](https://github.com/meteor/meteor/pull/13059#issuecomment-2007063599)

## Feature

Two commands are now operational:

`npx meteor@alpha install`/`npx meteor@alpha uninstall`: New recommended method that provides full output during Meteor installation or uninstallation. Installation simplified with `npx meteor@alpha`/`npx meteor`.
`npm install -g meteor`: Installs Meteor silently unless --foreground-scripts is specified.

## Fixes

There are several issues on various environments reported by devs when you execute `npm i -g meteor@beta`.

So this PR is about to review each OS environment and check that the installation command is still working fine. Try also to reproduct the reported issues and fix them.

My main suspicion is when using a new node version (`20.x`) for the `meteor-installer` script, might be the root cause of the issues. It's possible that alterations in the node API or outdated third-party packages could be contributing to these failures.

### Linux

- Using a virtual Machine on Ubuntu 22.04

#### Reproduction

Just run `npm install -g meteor@beta` and you will get either:

The process blocked like this:

image

Or 416 error:

image

#### Fix

Just by upgrading `node-downloader-helper` to latest `2.1.9`. It make sense since the 416 error is about to request specific download parts, and that library serves as a helper in this process.

image

### Mac

Same issue as linux happened in both Mac Intel and ARM. No other issues found.

### Windows

Failed with a different error than the Unix-like envs as we may expect.

image

 However, running the process with the fix on this PR solved it.

image

### DIFF
--- a/guide/source/3.0-migration.md
+++ b/guide/source/3.0-migration.md
@@ -123,6 +123,14 @@ You can create a new Meteor 3.0 project by running the command below:
 meteor create my-new-project --release 3.0-beta.6
 ```
 
+or alternatively, you can use `npx`
+
+If you want to test Meteor v3, you can use the following command to install it:
+
+```bash
+npx meteor@beta
+```
+
 ### How to update from version 2?
 
 You can update your Meteor 2.x project by running the command below inside your project folder:
--- a/npm-packages/meteor-installer/cli.js
+++ b/npm-packages/meteor-installer/cli.js
@@ -4,7 +4,7 @@ const command = process.argv[2];
 
 if (!command) {
   console.log(`
-  Usage: meteor-installer <command>
+  Usage: npx meteor@<version> <command>
 
   Commands:
     install
--- a/npm-packages/meteor-installer/cli.js
+++ b/npm-packages/meteor-installer/cli.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-const command = process.argv[2];
+const command = process.argv[2] || 'install';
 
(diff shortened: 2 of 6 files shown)


nachocodoner:
I've resolved all issues with the Meteor installer. Here were the main problems causing the `npm install -g meteor` command to fail:

- Updating `node-downloader-helper` to support node version `20.11.1`, required by Meteor 3.x.
- The lack of npm hook logs from `npm install -g meteor` caused by new npm version led to silent failures. Solved by local debugging or using the `--foreground-scripts` option in `npm@7`.
- Transitioning to `npx meteor@alpha install` as the preferred installation method for Meteor. Initially failed due to missing `-g` option, resolved by adjusting the error handling to accommodate `npx` behavior. I tried to differentiate between executing `npx` and `npm` to determine whether to fail with the global option, but npm's hook install script doesn't allow for such distinction.

Two commands are now operational:

- `npx meteor@alpha install`/`npx meteor@alpha uninstall`: New recommended method that provides full output during Meteor installation or uninstallation.
- `npm install -g meteor`: Installs Meteor silently unless `--foreground-scripts` is specified.

Grubba27 (approved):

Grubba27:
We will need to announce this and also we will need to update our swags to use npx now 🤣

denihs (approved):

Issue #12948, opened 2024-01-04
[Meteor 3.0] Promise rejection stops process


henriquealbert:
```js
// server.js
Meteor.methods({
  shouldFail: async () => {
    Promise.reject(new Error("Oh no!"));
  }
});

// or

Meteor.methods({
  shouldFail: async () => {
    Promise.reject(new Meteor.Error("Oh no!"));
  }
});
```

```js
// client.js
 Meteor.callAsync("shouldFail");
```

This shouldn't occur on Meteor 2.X this doesn't happen

![Screenshot 2024-01-04 at 16 17 03](https://github.com/meteor/meteor/assets/59376543/e4c6046b-cf9d-4d8f-bca8-4c186bd6ed8a)

zodern:
Node 15 changed the default from warning to throwing the error: https://nodejs.org/api/cli.html#--unhandled-rejectionsmode

Meteor doesn't control the node options used in production, so it might be good to leave it as the default to avoid differences between development and production.

denihs:
As we discussed in https://github.com/meteor/meteor/pull/13062, the Meteor 3 main process now runs inside a promise, meaning that if there is an unhandled rejection inside the app, this main process will be killed.

So, it's best to adopt some conventions when dealing with Promises. For example, you can return this `Promise.reject`:

``` js
Meteor.methods({
  shouldFail: async () => {
    return Promise.reject(new Meteor.Error("Oh no!"));
  }
});
```

Another solution would be to throw an error, like:

``` js
Meteor.methods({
  shouldFail: async () => {
     throw new Meteor.Error("Oh no!"));
  }
});
```

You can also add an `unhandledRejection` event in your project:

```js
process.on("unhandledRejection", (e) =>
  console.error("Unhandled promise rejection:", e.stack || e)
);
```

I'm closing this issue, and we can keep discussing it if needed.

PR #13062, opened 2024-03-18
Server crash on use Promise.reject within method body

denihs:
Solves https://github.com/meteor/meteor/issues/12948.

If you have an unhandled rejection in the server of your app, your server will crash.

This happens because now, in version 3, the main process runs inside a promise. If there is an unhandled promise inside the app, this rejection propagates to the top and kills the server promise.

The best practice is to always handle all promises. In the issue this example is used:

```js
Meteor.methods({
  shouldFail: async () => {
    Promise.reject(new Meteor.Error("Oh no!"));
  }
});
```

To avoid the problems reported to the issue, just return this `Promise.rejection` would do the trick:

```js
Meteor.methods({
  shouldFail: async () => {
    return Promise.reject(new Meteor.Error("Oh no!"));
  }
});
```

But, to avoid this kind of issue in version 3, we're now catching these unhandled rejections.

### DIFF
--- a/tools/static-assets/server/boot.js
+++ b/tools/static-assets/server/boot.js
@@ -492,6 +492,14 @@ var runMain = Profile("Run main()", async function () {
   }
 });
 
+
+// As our main process runs now inside a Promise, we need to make that if the user has an unhandled promise
+// in their app, ths unhandledRejection doesn't crash the server
+process.on("unhandledRejection", (e) =>
+  console.error("Unhandled promise rejection:", e.stack || e)
+);
+
+
 (async function startServerProcess() {
   if (!global.asyncLocalStorage) {
     const { AsyncLocalStorage } = require('async_hooks');
--- a/packages/email/.npm/package/npm-shrinkwrap.json
+++ b/packages/email/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 4,
   "dependencies": {
     "@types/node": {
-      "version": "20.11.20",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-20.11.20.tgz",
-      "integrity": "sha512-7/rR21OS+fq8IyHTgtLkDK949uzsa6n8BkziAKtPVpugIkO6D+/ooXMvzXxDnZrmtXVfjb1bKQafYpb8s89LOg=="
+      "version": "20.11.28",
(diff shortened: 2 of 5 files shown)


tools/static-assets/server/boot.js
```diff
@@ -492,6 +492,14 @@ var runMain = Profile("Run main()", async function () {
...
 });
 
+
+// As our main process runs now inside a Promise, we need to make that if the user has an unhandled promise
+// in their app, ths unhandledRejection doesn't crash the server
+process.on("unhandledRejection", (e) =>
+  console.error("Unhandled promise rejection:", e.stack || e)
+);
+
+
```
Grubba27, line 502: I like this, but this might be a workaround.

Grubba27 (approved):

nachocodoner (approved):

zodern:
Node 15 changed the default to crash the process by default when there is an unhandled promise rejection: https://nodejs.org/api/cli.html#--unhandled-rejectionsmode

I think it would be better to leave the default behavior, and apps can use the `--unhandled-rejections` node option or add an event handler if they want to do something different. For some of my projects, I want Node's default behavior.

tools/static-assets/server/boot.js
nachocodoner: Whenever an "unhandledRejection" happens, it's crucial not only to log it but also to address and handle the failing promises. I wonder if we just ignore the crashing this could affect also the platform code itself and we potentially can have false positive during tests,  without our awareness.

Maybe we could handle it by logging it and give some extra instructions on ideas for devs to act on it, but crash it anyway. In the instructions we may include [@zodern](https://github.com/meteor/meteor/pull/13062#issuecomment-2004117304) suggestion of adding the option `--unhandled-rejections` on run the app and link to node's api to properly handle `Promise.rejection`.

denihs:
I agree.

Now that the main Meteor process is inside a promise, we need to make sure people are aware of this, and they should handle all Promises. 

What we're doing in this PR is kind of a workaround as Gabriel said.

I'm closing this PR.

tools/static-assets/server/boot.js
denihs: > Maybe we could handle it by logging it and give some extra instructions

Good idea! Let's discuss this later and open it in other PR.

PR #13067, opened 2024-03-21
[Meteor 3] fix Circle CI group 5 tests

denihs:
Fix test `modules - test app`:

- Update `modules` app to version 3
- add testing packages in version 3
    - `meteortesting:mocha@3.1.0-beta300.0`
    - `meteortesting:mocha-core@8.3.1-beta300.0`


nachocodoner (approved):

tools/tests/apps/modules/tests.js
```diff
@@ -99,11 +99,6 @@ describe("app modules", () => {
     }
   });
 
-  it("should not be parsed in strictMode", () => {
```
zodern, line 102: This indicates a bug in reify. We should keep this test and fix reify. There was a discussion on slack about it: https://meteor-community.slack.com/archives/C0669FW38UU/p1710181750813509

denihs: I see. I'll have a look at this later. Thanks.

zodern: Thanking about this more, if we want to switch modules to being in strict mode this might be a good time to do it. Modules are supposed to be in strict mode, so if we are fine with the breaking changes then we could leave reify as is.

denihs: I agree with you. I also think it's the right time to do this time of change. I'm going to merge this PR, and we'll consider this a breaking change.

PR #13074, opened 2024-03-27
[Meteor 3] Stop loading packages in pararel

denihs:
For more context: https://forums.meteor.com/t/order-of-loading-packages/61112.

We are getting error reports about the order in the packages that are being imported.

There isn't a bug. The code works by design, but the new brings some issues.

### Solution

As @zodern [suggested](https://forums.meteor.com/t/order-of-loading-packages/61112/16?u=denyhs), I'm removing all deps code from `code-runtime` and `linker`.

But It's not working correctly. I'm getting this error:

Screenshot from 2024-03-27 11-17-07

Inside `ddp-server`, `Webapp` object is not being found, even though the `webapp` package loads first.

I'll contact @zodern, and hopefully, they can set us on the right path.

**Reproduction:** https://github.com/plbkbx/testLoadingPackages (run this app with your local Meteor at this branch).

### DIFF
--- a/tools/isobuild/linker.js
+++ b/tools/isobuild/linker.js
@@ -956,9 +956,8 @@ var getHeader = function (options) {
   });
 
   chunks.push(
-      `Package["core-runtime"].queue("${options.name}", [`,
-      orderedDeps.join(', '),
-      '], function () {'
+      `Package["core-runtime"].queue("${options.name}",`,
+      'function () {'
   );
 
   if (isApp) {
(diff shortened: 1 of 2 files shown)


tools/isobuild/linker.js
```diff
@@ -956,9 +956,8 @@ var getHeader = function (options) {
   });
 
   chunks.push(
-      `Package["core-runtime"].queue("${options.name}", [`,
-      orderedDeps.join(', '),
-      '], function () {'
+      `Package["core-runtime"].queue("${options.name}",`,
```
zodern, line 959: You should be able to completely remove the `deps` option from fullLink, and replace it with a `hasRuntime` option instead. This would also help with the linker cache so it isn't invalidated as often.

PR #13066, opened 2024-03-21
Release 2.16

nachocodoner:
```
meteor update --release 2.16
# or 
meteor create appname --release  2.16
```

We'll make sure Meteor 3.x isn't affected by new additions in this release. Tracking progress on PR https://github.com/meteor/meteor/pull/13098.

### Best Candidates

These are the top PR candidates for the next release, focusing on Meteor 3.x migration. Original authors will address any necessary checks, with the core team reviewing for compatibility with both 2.x and future 3.x versions.

- [x] [339 Support observeChangesAsync and observeAsync [Meteor 2.x]](https://github.com/meteor/meteor/pull/13025)

- [x]  New mongo package options to optimize Oplog tailing performance to include/exclude certain collections
https://github.com/meteor/meteor/pull/13009

### Quick Gains

These PRs are straightforward and expected to integrate smoothly into Meteor 3.x, offering immediate benefits. We will double check them on a release-3.0 based-branch.

- [x] chore: remove repetitive words
https://github.com/meteor/meteor/pull/13093

- [x] Get default options from meteor settings
https://github.com/meteor/meteor/pull/13005

- [x] Session storage option for accounts
https://github.com/meteor/meteor/pull/13046

- [x] check - throw all errors
https://github.com/meteor/meteor/pull/12970

- [x] Add nodemailer-openpgp
https://github.com/meteor/meteor/pull/12991

### Hotfixes

Update terser to v5.31.0
https://github.com/meteor/meteor/pull/13120

### Discarded

Other PRs are currently excluded from the upcoming release due to the effort needed for verification and potential impact on Meteor 3.x transition, which remains our primary focus.

DDP resumption
https://github.com/meteor/meteor/pull/12436

Update Accounts with Async methods
https://github.com/meteor/meteor/pull/12647

Remove underscore
https://github.com/meteor/meteor/pull/12245
https://github.com/meteor/meteor/pull/12261
https://github.com/meteor/meteor/pull/12259
https://github.com/meteor/meteor/pull/12238
https://github.com/meteor/meteor/pull/12267

Asyncify _checkForCaseInsensitiveDuplicates
https://github.com/meteor/meteor/pull/13076

-----------------------------

Any other PR to be considered for including in 2.16 release must point `release-2.16` branch, update to latest changes, have green checks on that branch, and ask for a review keeping the PR alive.

PR #13076, opened 2024-04-01
Asyncify _checkForCaseInsensitiveDuplicates

harryadel:
In our application we call `_createUserCheckingDuplicates` and I found out it's not async-friendly yet. Can we please add it to 2.16?


nachocodoner (approved):

nachocodoner:
Added on the list for 2.16.
https://github.com/meteor/meteor/pull/13066

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -93,13 +93,13 @@ Accounts.registerLoginHandler('passwordless', options => {
...
 const pluckAddresses = (emails = []) => emails.map(email => email.address);
-const createUser = userData => {
+const createUser = async userData => {
   const { username, email } = userData;
   if (!username && !email) {
     throw new Meteor.Error(400, 'Need to set a username or email');
   }
   const user = { services: {} };
-  return Accounts._createUserCheckingDuplicates({
+  return await Accounts._createUserCheckingDuplicates({
```
nachocodoner, line 102: I believe it's a matter of taste, but usually, using return await isn't necessary. You can typically return the promise directly in an async function.

packages/accounts-passwordless/passwordless_server.js
```diff
@@ -114,7 +114,7 @@ function generateSequence() {
 }
 
 Meteor.methods({
-  requestLoginTokenForUser: ({ selector, userData, options = {} }) => {
+  async requestLoginTokenForUser({ selector, userData, options = {} }) {
```
nachocodoner, line 117: Changing this method to async may need to also use Meteor.callAsync where is used, in the client part. So here may need to also be adapted, https://github.com/harryadel/meteor/blob/fix/_checkForCaseInsensitiveDuplicates/packages/accounts-passwordless/passwordless_client.js#L106-L110

cc @Grubba27

nachocodoner (changes requested):
I think there are still changes needed to the proper adoption.

packages/accounts-passwordless/passwordless_server.js
Grubba27: To avoid breaking the API now, we could use the callAsync version and the callback to resolve the issue when the promise is resolved.

And may add the possibility of returning a promise for Accounts.requestLoginTokenForUser.

People can still use the callback-based API or resolve the API using promises.

```js

Accounts.requestLoginTokenForUser(options, (err) => err ? "ok" : "not ok"
)

/// could be used as well like this:

try {
  await Accounts.requestLoginTokenForUser(options);
  "ok"
} catch (err) {
  "not ok"
}


```

It would not be a breaking change, but rather a new feature in this PR

zodern: `Meteor.call`/`Meteor.apply` work fine when the method is async on the server. With Meteor 2, it's only a problem if the stub is async.

jamauro:
@harryadel out of curiosity, why do you call this internal method in your app?

packages/accounts-passwordless/passwordless_server.js
nachocodoner: Ok. Thank you @zodern.

So, [in this example](https://github.com/harryadel/meteor/blob/fix/_checkForCaseInsensitiveDuplicates/packages/accounts-passwordless/passwordless_client.js#L106-L110) if I am not mistaken, the method is [server-only as described on server file only](https://github.com/harryadel/meteor/blob/fix/_checkForCaseInsensitiveDuplicates/packages/accounts-passwordless/passwordless_server.js#L117), and Meteor.call/apply would still work fine when called from the client.

I think the same happens for this exact code on Meteor 3.x, where is async actually and server-only.

harryadel: What's the best course of action here? Keep it as is? Or change it use to `.callAsync`?

nachocodoner: I think we can leave it as it is, since is server-only. So I will skip any adoption here according of what mentioned above. 

Even in Meteor 3.x [is async on server](https://github.com/meteor/meteor/blob/release-3.0/packages/accounts-passwordless/passwordless_server.js#L117) and called with [call on the client](https://github.com/meteor/meteor/blob/release-3.0/packages/accounts-passwordless/passwordless_client.js#L106-L110). So lets skip this comment here.

harryadel:
@jamauro We've this unique case in our application where we have the user `sha256` client-hashed password and some other meta data regarding the user document so we end up stitching these properties together and using an internal creation method (`_createUserCheckingDuplicates` in our case) to make the final user document and allow him to login.

packages/accounts-passwordless/passwordless_server.js
harryadel: I know it might seem redundant but it's just clear.

nachocodoner:
> On the 3.0 branch is this fixed? Is the accounts-password package completely free of sync db-calls??

It seems that is adapted actually. If you look into the release-3.0 branch you will see the code using async, [and the ones you adapted here, already migrated](https://github.com/meteor/meteor/blob/release-3.0/packages/accounts-password/password_server.js#L230-L263).

In the changelog we keep the list of package's API that have been migrated to async, for instance, [for the `accounts-password` package](https://github.com/meteor/meteor/blob/release-3.0/docs/history.md?plain=1#L63-L81). So, your PR here will produce a breaking change for apps in Meteor 2.x, that make use of exposed API behaviors `addEmail` and `setUsername` which will be resolved as promise on next 2.16. But this migration will make ready for Meteor 3.x as stated above those are already async.

Alternatively we could have async versions for  `addEmailAsync` and `setUsernameAsync` , the same way we have for [`Accounts.setPasswordAsync`](https://github.com/harryadel/meteor/blob/7411b3c85a3c95a6b6f3c588babe6eae894d6fb6/packages/accounts-password/password_server.js#L323). But as they are migrated already as async in Meteor 3.x it may not be required.

cc @Grubba27, @denihs

harryadel:
Our philosophy so far has been to get everything asyncified, just like you mentioned with [Accounts.setPasswordAsync](https://github.com/harryadel/meteor/blob/7411b3c85a3c95a6b6f3c588babe6eae894d6fb6/packages/accounts-password/password_server.js#L323) so that when we run `meteor update 3.0` there's very minimal work and it's the last step rather than another minor step. 

Though my opinion on this isn't set in stone, let's see what Denis & Grubba has to say about this.

Grubba27:
I would go in the safe route here...

We decided to create those `*Async` APIs to help out the migration, and following that pattern here is the way to go.


We should focus on not breaking changes and making those breaking changes as easy as possible for those making the transition.

We could have `addEmailAsync` and `setUsernameAsync`, leaving those without an async name until 3.0, when addEmailAsync and setUsernameAsync become aliases to their non-Async counterparts.


```js

// 2.x

Accounts.addEmail // <- sync
Accounts.addEmailAsync // <- resolves to promise

// 3.x

Accounts.addEmail // <- resolves to promise it is just an alias to the addEmailAsync
Accounts.addEmailAsync // <- resolves to promise

```

nachocodoner:
That makes sense. We didn't do the same for 3.x and instead change to async directly. But we may change this now, since I don't think compromise a lot. So, if you can adapt this @harryadel to use async versions instead, that would be great.

harryadel:
@nachocodoner so I'd duplicate the functions and create async alternatives so it doesn't break anything for 2.x or just change the variables names to use the `async` prefix?

nachocodoner:
> I'd duplicate the functions and create async alternatives so it doesn't break anything for 2.x 

This is better to stick into the convention. This way we won't force any breaking changes on the new update.

nachocodoner:
@harryadel: Do you want to include this on release 2.16? How is the state of the feedback applied? 

We should create async versions as stated here and keep easy migration with 3.x, https://github.com/meteor/meteor/pull/13076#issuecomment-2042877474

harryadel:
I had initially started this PR to help ease our migration to 3.0 but since we're already attempting to migrate to the latest RC it seizes to grant any value for us. Plus, it was discarded in 2.16 release.

PR #13078, opened 2024-04-03
[Meteor 3.0] Making our CI more easy to understand

Grubba27:
In this PR, I address the issue of the Travis CI not being so good to read when tests are running(currently only shows a `testNumber` variable). This PR will solve that(kind of). 

We will add the current test that is running if we do not have that in the context, we will still print the `testNumber` variable, but  for the most part, our ci will start to look like this:

Screenshot 2024-04-03 at 15 38 30

If it ever gets stuck somewhere, we will know it

### DIFF
--- a/packages/test-in-console/puppeteerRunner.js
+++ b/packages/test-in-console/puppeteerRunner.js
@@ -10,11 +10,10 @@ async function runNextUrl(browser) {
     // if the test is running for too long without any output to the console (10 minutes)
     if (msg._text !== undefined) console.log(msg._text);
     else {
-      // sometimes if the test is running in the client we have id: current-client-test with
-      // the name of the test
-      const element = await page.$('#current-client-test');
-      if (element) {
-        const currentClientTest = await page.evaluate(element => element.textContent, element);
+      testNumber++;
+      const currentClientTest =
+       await page.evaluate(() =>  __Tinytest._getCurrentRunningTestOnClient());
+      if (currentClientTest !== '') {
         console.log(`Currently running on the client test: ${ currentClientTest }`)
         return;
       }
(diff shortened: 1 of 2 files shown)


Grubba27:
I’m still thinking about whether we should use Pupperter or stick to using meteor for getting the test name in the client.
```js
      const element = await page.$('#current-client-test');
      if (element) {
        const currentClientTest = await page.evaluate(element => element.textContent, element);
        console.log(`Currently running on the client test: ${ currentClientTest }`)
        return;
      }

```
could become:

```js
      const currentClientTest =
       await page.evaluate(async () => await __Tinytest._getCurrentRunningTestOnClient());

      if (currentClientTest !== '') {
        console.log(`Currently running on the client test: ${ currentClientTest }`);
        return;
      }
```

nachocodoner:
Shouln't this be based on devel or cherry picked to also affect 2.x?

StorytellerCZ:
Just a question: Why is Travis used here instead of GitHub Actions? Just legacy and too much work to move?

Grubba27:
> Shouln't this be based on devel or cherry picked to also affect 2.x?

I agree! I'll make a PR to cherry-pick to 2.x


> Just a question: Why is Travis used here instead of GitHub Actions? Just legacy and too much work to move?

I've asked the same question once...  

From my studies, what I could see is that the main issues could be the mobile tests.
I wanted to move to GH Actions, both Travis and CircleCI, but from what I recall, we have a **_lot_** of CI credits in one of them, and it is wise to use those credits and then move to another CI.

Also, it could take some time and people never placed this as a priority(since I started working in Meteor we have been in this fibers operation)

PR #13081, opened 2024-04-05
fix: Uncaught TypeError: Cannot convert undefined or null to object in livedata_connection.js

mjcctech:
We are using https://github.com/wojtkowiak/meteor-custom-protocol and some messages have `null` value for msg.  Unfortunately code checking this case has a bug and end up with error message in browser console 
```
livedata_connection.js:1780 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Connection.onMessage (livedata_connection.js:1780:20)
    ...
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+@henriquealbert @denihs @fredmaiaarantes @nachocodoner
--- a/packages/ddp-server/livedata_server_tests.js
+++ b/packages/ddp-server/livedata_server_tests.js
@@ -102,6 +102,7 @@ Tinytest.addAsync(
     function (test, onComplete) {
 
         var cb = Meteor.onMessage(function (msg, session) {
+            if (msg.method !== 'livedata_server_test_inner') return;
             test.equal(msg.method, 'livedata_server_test_inner');
             cb.stop();
             onComplete();
--- a/packages/test-in-console/puppeteerRunner.js
+++ b/packages/test-in-console/puppeteerRunner.js
@@ -5,12 +5,29 @@ let testNumber = 0;
 async function runNextUrl(browser) {
   const page = await browser.newPage();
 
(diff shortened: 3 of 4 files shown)


packages/ddp-client/common/livedata_connection.js
```diff
@@ -1776,8 +1776,7 @@ export class Connection {
     }
 
     if (msg === null || !msg.msg) {
-      if(!msg || !msg.testMessageOnConnect) {
-        if (Object.keys(msg).length === 1 && msg.server_id) return;
+      if (msg && (!msg.testMessageOnConnect || !(Object.keys(msg).length === 1 && msg.server_id))) {
```
StorytellerCZ, line 1779: Shouldn't the clause be `&&`?

mjcctech: you are right

StorytellerCZ (approved):

nachocodoner (approved):
For me this is ok. I have seen this error on the past and hopefully this covers the fix.

However, since this change impacts the core, I'd appreciate another experienced set of eyes, like @denihs, to review this. Tests are passing, but it's crucial to ensure we haven't missed any scenario with this change.

denihs (approved):

denihs:
This change looks fine to me as well :raised_hands:

nachocodoner:
I will have to cherry pick this change since it doesn't allow me to merge, likely old branch having important conflicts. I will check next.

nachocodoner:
Closing, it resulted that this change was already integrated as part of a later refactor: https://github.com/meteor/meteor/blob/6c1fd9c3138ec3fbc6959cd4e335897cb025f5b5/packages/ddp-client/common/connection_stream_handlers.js#L28-L34

Logic moved to connection_stream_handlers.js file and the fix applied here was refactored.

PR #13082, opened 2024-04-05
[Meteor 3] Fix how to compare constraints version when publishing a package

denihs:
We had reports that people weren't able to deploy packages because they were getting the error:

```
inconsistent dependency constraint across unibuilds?
```

Debugging further, I could check that [here](https://github.com/meteor/meteor/commit/02799444946482596614bd7451e002032ef16ae9#diff-5e43cc683d68bb7ff69efe9f41009b0e33512ebef49a296090508164d56de97eL1757), `d.constraint` and `use.constraint` were never equal if using `api.versionsFrom` with conditioning versions for a package, like `'mongo@1.6.9||2.0.0-beta300.7'`.

What I'm doing on this PR is to verify if the package constraint version is available in the specified Meteor version determined by `api.versionsFrom`.


nachocodoner:
I'm wondering if we could develop tests to aid in this verification process. Could we add a self-test for this build stuff, perhaps even unit testing for this specific handling?

zodern:
I tested this with the cause for the error from #13111, and this PR incorrectly allows the inconsistent constraints.

If you add this to `Package.onUse`:
```
  api.use('montiapm:agent@2.44.2||3.0.0-beta.4');
  api.imply('montiapm:agent@2.44.2');
```

Meteor should error since having different constraints for a package in `api.use` and `api.imply` is unsupported. When running with this PR, Meteor instead builds the package without erroring.

denihs:
I took time to analyze everything here, and https://github.com/meteor/meteor/pull/13111 gives me the fitting errors to solve the issue. This code is in fact not correct. I'm closing this PR.

PR #13087, opened 2024-04-09
Algolia to v3 docs

Grubba27:
This PR adds Algolia and sitemap.xml to our v3 docs.

 
Screenshot 2024-04-09 at 13 45 05

_Testing the crawler_

### DIFF
--- a/v3-docs/docs/.vitepress/config.mts
+++ b/v3-docs/docs/.vitepress/config.mts
@@ -6,6 +6,9 @@ export default defineConfig({
   description: "Meteor.js API docs",
   head: [["link", { rel: "icon", href: "/logo.png" }]],
   lastUpdated: true,
+  sitemap: {
+    hostname: "https://v3-docs.meteor.com",
+  },
   themeConfig: {
     // https://vitepress.dev/reference/default-theme-config
     nav: [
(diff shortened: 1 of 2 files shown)


Grubba27:
something is off... I'll check why the site is not getting the values from the crawler

v3-docs/docs/.vitepress/config.mts
```diff
@@ -276,8 +279,17 @@ export default defineConfig({
     logo: { dark: "/meteor-logo.png", light: "/meteor-blue.png" },
 
     search: {
-      provider: "local",
+      provider: 'algolia',
+      options: {
+        appId: '2RBX3PR26I',
+        apiKey: '7fcba92008b84946f04369df2afa1744',
```
nachocodoner, line 285: I believe this api key refers to [search api key]( https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key), [not the admin one](https://www.algolia.com/doc/guides/security/api-keys/#admin-api-key), right? So it is safe to be publicly available, even though it may expose and [have some risks in place](https://www.algolia.com/doc/guides/security/api-keys/#risks-of-api-key-exposure).

Don't we have a secret management tool in place?

Grubba27: Yeah, it is the public one!

VItepress apiKry is also public https://github.com/vuejs/vitepress/blob/f65053c1c6d15a8a7867a5363484656a6ba766f1/docs/.vitepress/config/shared.ts#L57

We could but it is not necessary I think

henriquealbert: I agree with @Grubba27. I think there is no problem

henriquealbert (approved):

PR #13096, opened 2024-04-16
Fix for old API usage

manueltimita:
Fixes "Error in OAuth Server: findOne +  is not available on the server. Please use findOneAsync() instead."

The error stems from `oauth_server.js` still using the sync Mongo API at `ServiceConfiguration.configurations.findOne(...)`

### DIFF
--- a/packages/oauth/oauth_server.js
+++ b/packages/oauth/oauth_server.js
@@ -156,7 +156,7 @@ const middleware = async (req, res, next) => {
       throw new Error(`Unexpected OAuth service ${serviceName}`);
 
     // Make sure we're configured
-    ensureConfigured(serviceName);
+    await ensureConfigured(serviceName);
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
Is this for the current 2.x version or 3.x. I have fixed at least the second part in 3.x already.

manueltimita:
I'm using oauth@3.0.0-beta300.7 in development, where the error surfaced, so my intention was to fix it in 3.x. If you fixed it already, it looks like I misunderstood how the `devel` branch works in this repo. I think we can close this then.

StorytellerCZ:
Check out the latest `release-3.0` branch. If you want to merge into 3.0 then you need to target that branch.

manueltimita:
Thanks, will do that in the future.

PR #13102, opened 2024-04-20
[Meteor 3] Added `waitUntilAllLoaded` safety check.

radekmie:
As recommended in https://github.com/meteor/meteor/pull/13095#discussion_r1571327249, I added additional check in `waitUntilAllLoaded` in case it gets called while the last module is still being loaded. I was not able to prepare a test for that (the queue is never empty in my tests), I didn't add any.


denihs (approved):

nachocodoner (changes requested):
Could we also add the suggestion here to avoid `meteor reset` on upgrade?
https://github.com/meteor/meteor/issues/13104#issuecomment-2070608672

denihs (approved):

Issue #1525, opened 2013-10-23
No css loaded on deploy after 0.6.6.2 update


flaborda:
After updating meteor to 0.6.6.2 and deploying my app, there is no css loaded at all.

It only happens on the deployed version, the local one works just fine.

I'm using zurb-foundation package.

Everything works perfectly if I deploy it with --debug

Issue #1512, opened 2013-10-18
Private browsing iOS incorrect password


davidworkman9:
In Safari on iOS 7 with private browsing enabled the Meteor login UI states that the password is incorrect.

Issue #1499, opened 2013-10-14
blockquote doesn't work in showdown


dan335:
I'm using the showdown package.  This should be a block quote...

```
> blah
```

It works when doing this

```
var converter = new Showdown.converter();
var html = converter.makeHtml(text);
```

but not when doing this

```
{{#markdown}}{{text}}{{/markdown}}
```

All other markdown syntax that I've tried works fine.  

Also I wish some things from github flavored markdown were in showdown but that's another issue.

PR #13118, opened 2024-04-30
update to the latest version of terser

acemtp:
We spent 3 hours to find that the terser version used in Meteor (5.14.2) has a bug that make minifying React generates a bug in "finally" part.

Upgrading to the latest version 5.31.0 of terser fix the problem.

We cloned the minify-js package locally so it's ok for us but please include this bump version in rc3

Here is the only diff (version bump)
https://github.com/acemtp/meteor/commit/63630ef40d52eaf69806e16d110d39cd4e6d784c

### DIFF
--- a/packages/minifier-js/package.js
+++ b/packages/minifier-js/package.js
@@ -4,7 +4,7 @@ Package.describe({
 });
 
 Npm.depends({
-  terser: "5.14.2"
+  terser: "5.31.0"
 });
 
 Package.onUse(function (api) {


acemtp:
We spent 3 hours to find that the terser version used in Meteor (5.14.2) has a bug that make minifying React generates a bug in "finally" part.

Upgrading to the latest version 5.31.0 of terser fix the problem.

We cloned the minify-js package locally so it's ok for us but please include this bump version in rc3

Here is the only diff (version bump)
https://github.com/acemtp/meteor/commit/63630ef40d52eaf69806e16d110d39cd4e6d784c

StorytellerCZ:
Fixed the branch target.
@nachocodoner this might be worth of cherry picking into 2.16 release as well I think.

StorytellerCZ:
Here is the changelog from Terser since the last update:

# Changelog

## v5.31.0
 - Sync up property mangler exceptions with current contents of Firefox and Chrome environments
 - Add more webcomponent properties to property mangler exceptions (#1525)
 - Drop non-nullish constants in `...spreads` in objects (#1141)

## v5.30.4
 - Fix parsing `#private in ...` when next to other operators

## v5.30.3
 - Fix precedence of `#private in ...` operator

## v5.30.2
 - Avoid optimizations inside computed keys, because they can cause js-engine-specific bugs.

## v5.30.1
 - Removed useless `\` escapes for non-ascii characters
 - Make modern identifier characters quoted for older environments (#1512)

## v5.30.0
 - Improve removal of classes referring to themselves

## v5.29.2
 - Make sure 'computed_props' creates string keys
 - Take into account the evaluated size when inlining

## v5.29.1
 - fix optimisation of all-bits mask check

## v5.29.0
 - Re-releases previously reverted 5.28.0
 - Fix crash while optimizing some bitwise ops
 - (internal) Remove needless wrapper for from_moz (#1499)

## v5.28.1
(hotfix release)
 - Reverts v5.28.0

## v5.28.0
 - Optimise redundant or shrinkable bitwise operations (`|`, `^`, `&`, `>>`, `<<`)
 - Evaluate some `BigInt` math operations

## v5.27.2
 - Recognise `this` as a reference to the surrounding class in `drop_unused`. Closes #1472

## v5.27.1
 - Fixed case where `collapse_vars` inlines `await` expressions into non-async functions.

## v5.27.0
 - Created `minify_sync()` alternative to `minify()` since there's no async code left.

## v5.26.0
 - Do not take the `/*#__PURE__*/` annotation into account when the `side_effects` compress option is off.
 - The `preserve_annotations` option now automatically opts annotation comments in, instead of requiring the `comments` option to be configured for this.
 - Refuse to parse empty parenthesized expressions (`()`)

## v5.25.0
 - Regex properties added to reserved property mangler (#1471)
 - `pure_new` option added to drop unused `new` expressions.

## v5.24.0
 - Improve formatting performance in V8 by keeping a small work string and a large output string

## v5.23.0
 - When top_retain will keep a variable assignment around, inline the assignee when it's shorter than the name (#1434)
 - Remove empty class `static {}` blocks.

## v5.22.0
 - Do not `unsafe`ly shorten expressions like a?.toString() when they're conditional.
 - Avoid running drop_unused in nodes that aren't scopes. Fixes a rare crash.
 - When 'module' is enabled, assume strict mode when figuring out scopes.

## v5.21.0
 - Do not inline functions that would be retained in the toplevel (as this would cause code duplication).
 - Fix precedence of arrow function and ternary operator when formatting output.

## v5.20.0
 - Passing `minify()` zero files will now throw a clean exception (#1450)
 - `drop_console` supports passing in an array of `console.*` method names (#1445)
 - New DOM properties from the WebGPU API have been added for use in the property mangler (#1436)
 - Internal code simplification (#1437)

## v5.19.4
 - Prevent creating very deeply nested ternaries from a long list of `if..return`
 - Prevent inlining classes into other functions, to avoid constructors being compared.

## v5.19.3
 - Fix side effect detection of `optional?.chains`.
 - Add roundRect to domprops.js (#1426)

## v5.19.2
 - fix performance hit from avoiding HTML comments in the output

## v5.19.1
 - Better avoid outputting `</script>` and HTML comments.
 - Fix unused variables in class static blocks not being dropped correctly.
 - Fix sourcemap names of methods that are `async` or `static`

## v5.19.0
 - Allow `/*@__MANGLE_PROP__*/` annotation in `object.property`, in addition to property declarations.

## v5.18.2
 - Stop using recursion in hoisted defuns fix.

## v5.18.1
 - Fix major performance issue caused by hoisted defuns' scopes bugfix.

## v5.18.0
 - Add new `/*@__MANGLE_PROP__*/` annotation, to mark properties that should be mangled.

## v5.17.7
 - Update some dependencies
 - Add consistent sorting for `v` RegExp flag
 - Add `inert` DOM attribute to domprops

## v5.17.6
 - Fixes to mozilla AST input and output, for class properties, private properties and static blocks
 - Fix outputting a shorthand property in quotes when safari10 and ecma=2015 options are enabled
 - `configurable` and `enumerable`, used in Object.defineProperty, added to domprops (#1393)

## v5.17.5
 - Take into account the non-deferred bits of a class, such as static properties, while dropping unused code.

## v5.17.4

 - Fix crash when trying to negate a class (`!class{}`)
 - Avoid outputting comments between `yield`/`await` and its argument
 - Fix detection of left-hand-side of assignment, to avoid optimizing it like any other expression in some edge cases

## v5.17.3

 - Fix issue with trimming a static class property's contents accessing the class as `this`.

## v5.17.2
 - Be less conservative when detecting use-before-definition of `var` in hoisted functions.
 - Support unusual (but perfectly valid) initializers of for-in and for-of loops.
 - Fix issue where hoisted function would be dropped if it was after a `continue` statement

## v5.17.1
 - Fix evaluating `.length` when the source array might've been mutated

## v5.17.0
 - Drop vestigial `= undefined` default argument in IIFE calls (#1366)
 - Evaluate known arrays' `.length` property when statically determinable
 - Add `@__KEY__` annotation to mangle string literals (#1365)

## v5.16.9
 - Fix parentheses in output of optional chains (`a?.b`) (#1374)
 - More documentation on source maps (#1368)
 - New `lhs_constants` option, allowing to stop Terser from swapping comparison operands (#1361)

## v5.16.8

 - Become even less conservative around function definitions for `reduce_vars`
 - Fix parsing context of `import.meta` expressions such that method calls are allowed

## v5.16.6

 - Become less conservative with analyzing function definitions for `reduce_vars`
 - Parse `import.meta` as a real AST node and not an `object.property`

## v5.16.5

 - Correctly handle AST transform functions that mutate children arrays
 - Don't mutate the options object passed to Terser (#1342)
 - Do not treat BigInt like a number

## v5.16.4

 - Keep `(defaultArg = undefined) => ...`, because default args don't count for function length
 - Prevent inlining variables into `?.` optional chains
 - Avoid removing unused arguments while transforming
 - Optimize iterating AST node lists
 - Make sure `catch` and `finally` aren't children of `try` in the AST
 - Use modern unicode property escapes (`\p{...}`) to parse identifiers when available

## v5.16.3

 - Ensure function definitions, don't assume the values of variables defined after them.

## v5.16.2

 - Fix sourcemaps with non-ascii characters (#1318)
 - Support string module name and export * as (#1336)
 - Do not move `let` out of `for` initializers, as it can change scoping
 - Fix a corner case that would generate the invalid syntax `if (something) let x` ("let" in braceless if body)
 - Knowledge of more native object properties (#1330)
 - Got rid of Travis (#1323)
 - Added semi-secret `asObject` sourcemap option to typescript defs (#1321)

## v5.16.1

 - Properly handle references in destructurings (`const { [reference]: val } = ...`)
 - Allow parsing of `.#privatefield` in nested classes
 - Do not evaluate operations that return large strings if that would make the output code larger
 - Make `collapse_vars` handle block scope correctly
 - Internal improvements: Typos (#1311), more tests, small-scale refactoring

## v5.16.0

 - Disallow private fields in object bodies (#1011)
 - Parse `#privatefield in object` (#1279)
 - Compress `#privatefield in object`

## v5.15.1

 - Fixed missing parentheses around optional chains
 - Avoid bare `let` or `const` as the bodies of `if` statements (#1253)
 - Small internal fixes (#1271)
 - Avoid inlining a class twice and creating two equivalent but `!==` classes.

## v5.15.0
 - Basic support for ES2022 class static initializer blocks.
 - Add `AudioWorkletNode` constructor options to domprops list (#1230)
 - Make identity function inliner not inline `id(...expandedArgs)`

nachocodoner (approved):
Seems good to me.

By any chance, do you have a specific reference for reproducing the issue or more details on the original problem? It would be helpful for us to document and redirect anything similar we encounter in the community.

nachocodoner:
I think is safe to ignore this failure on checks, since it happens as it can't find the branch from a forked source.

acemtp:
I think it's a very rare case, that we have in our main product and it would be hard to reproduce since it's deep inside terser.
They fixed it in version 5.16.4 https://github.com/terser/terser/blob/master/CHANGELOG.md#v5164

> "Make sure catch and finally aren't children of try in the AST"

I don't know why the PR was so big, it's my first PR, I don't know the right way to do that.

Since we are a (new) company that use Meteor and want to move to Meteor 3, we can give more PR on some community packages that currently not working on Meteor 3 if we know how to do that correctly.

nachocodoner:
> Since we are a (new) company that use Meteor and want to move to Meteor 3, we can give more PR on some community packages that currently not working on Meteor 3 if we know how to do that correctly.

Any contribution is welcome. There are some guidelines [on contributing](https://github.com/meteor/meteor/blob/devel/CONTRIBUTING.md) and [make changes into the core](https://github.com/meteor/meteor/blob/devel/CONTRIBUTING.md#making-changes-to-meteor-core). You can get additional insights but don't worry too much about it, we still need to improve those docs above all with the new addition of Meteor 3.0.

The change on this PR is straightforward. It only needs a brief context, as you mentioned in your description. This will help others with similar issues find the solution easily since all responses are indexed and accessible. Reproductions are always a plus if possible, specially for anyone to understand and support any fix.

Feel free on contribute more, we will be here to help and improve our contributing process.

PR #13127, opened 2024-05-08
[3.x] Fix async OP behaviors on server (oauth packages)

nachocodoner:
OSS-411

## Issue

Context: https://forums.meteor.com/t/meteor-3-0-rc-0-is-out/61515/28

In some parts of the code, `findOne` is still being utilized sync on Meteor 3.x, resulting in errors asking for migration, such as: `Error: findOne + is not available on the server. Please use findOneAsync() instead`.

## Fix

Review all occurrences of `findOne` and other operations needing asynchronous migration on Meteor 3.x packages. Additionally, I'll supply a few regression tests to ensure coverage of the affected flow, enabling us to detect any missing adaptions in the future through test failures.

The regression test triggering properly the requirement of this migration as showed in the next example:

image

## New test helpers

I've added two new test helpers:

- `mockBehaviours`: This acts as a mock to halt the run on an entire module. Useful for testing specific flows while ignoring unwanted side-effects.

- `waitUntil`: Prevents flakiness by awaiting a test assert with a timeout. It continuously checks for the condition and exits early upon its occurrence. This approach is more effective in overcoming flakiness.

### DIFF
--- a/packages/facebook-oauth/facebook-oauth_tests.js
+++ b/packages/facebook-oauth/facebook-oauth_tests.js
@@ -1,7 +1,7 @@
 Tinytest.addAsync(
   'facebook-oauth - run service oauth with mocked flow as expected',
   async function (test) {
-    const oauthMock = disableBehaviours(OAuth, {
+    const oauthMock = mockBehaviours(OAuth, {
       _fetch: () => Promise.resolve({ json: () => ({ access_token: 'testToken' })}),
     });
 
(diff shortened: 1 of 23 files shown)


denihs (approved):

manueltimita:
I saw this just as I submitted https://github.com/meteor/meteor/pull/13137 (this issue kept on bugging us in staging). Noticed you fixed the issue in the Google OAuth package.

StorytellerCZ (approved):

Issue #13113, opened 2024-04-26
[Meteor 3] EnvironmentVariable.withValue has different return value on client and server


zodern:
With this code:
```js
let ev1 = new Meteor.EnvironmentVariable();

console.log(ev1.withValue(10, () => 5));
```

On the client it logs `5`, and on the server it logs a promise. It should be the same on both the server and client.

Once #13112 is fixed, it should be fairly simple to have `withValue` synchronously return the value from the function, so it is only async if the passed function is also async, and it would behave the same on the client and server.

denihs:
This is now fixed!

Issue #13112, opened 2024-04-26
[Meteor 3] Environment Variables resets value too early


zodern:
With this code:
```
let ev1 = new Meteor.EnvironmentVariable();

async function runAsyncFunction() {
  await new Promise(resolve => setTimeout(resolve, 1000));
  console.log('value2', ev1.get());
}

ev1.withValue({ name: 'test' }, () => {
  runAsyncFunction();
  console.log('value 1', ev1.get());
});
```

In Meteor 1 and 2, it logs:
```
value 1 { name: 'test' }
value2 { name: 'test' }
```

However, in Meteor 3 it logs:
```
value 1 { name: 'test' }
value2 undefined
```

Meteor 1 and 2 preserve the value until all child async contexts complete. Async Local Storage does the same, which Environment Variables uses in Meteor 3, so it should be possible to preserve this behavior in Meteor 3.

nachocodoner:
[These tests](https://github.com/meteor/meteor/pull/13063/files#diff-b7edc13b0161fd0d8b0fa8587cfee1f55eb9c5b969c16abc092b644133f6c655R19-R64) demonstrate the current behaviors and expected support in Meteor 3.0.

In your example, code might need adjustment to ensure you properly await children asyncs to maintain the env variable context. For instance:

``` javascript
let ev1 = new Meteor.EnvironmentVariable();

async function runAsyncFunction() {
  await new Promise(resolve => setTimeout(resolve, 1000));
  console.log('value2', ev1.get());
}

ev1.withValue({ name: 'test' }, () => {
  ➡️ await runAsyncFunction();
  console.log('value 1', ev1.get());
});
```

With this you get the proper response.

``` javascript
value 1 { name: 'test' }
value2 { name: 'test' }
```

While achieving this might be possible, given my limited familiarity with this code or Async Local Storage behavior, further exploration is needed.

Thank you for bringing it to our attention.

cc @denihs

zodern:
Preserving the env vars in async child contexts was an important part of the implementation in Meteor 1/2. Meteor wrapped most callback api's with Meteor.bindEnvironment, and had code in the `promise` package to preserve the env var values in async functions and promises. Since ALS works the same, it seems this behavior is generally expected.

[...]

PR #13128, opened 2024-05-09
[Meteor 3] Environment Variable Tweaks

leonardoventurini:
OSS-406

Addresses these two issues:

https://github.com/meteor/meteor/issues/13112
https://github.com/meteor/meteor/issues/13113

1. No need to reset EV's value manually since ALS already handles it well by creating a new object every time.
2. Make sure the function's return value is preserved, WITHOUT enforcing a promise in the server.
3. Adjust the Client's `withValue` in `dynamics_browser.js` so it behaves similar to the server when handling promises. 

https://github.com/meteor/meteor/blob/08c2d7b679b6e70d5e95b56d7c82a5d31af7eae1/packages/meteor/dynamics_browser.js#L49


packages/meteor/dynamics_nodejs.js
```diff
@@ -60,20 +60,10 @@ class EnvironmentVariableAsync {
...
-          Meteor._updateAslStore(
-            UPPER_CALL_DYNAMICS_KEY_NAME,
-            undefined,
-          );
-        }
-        return ret;
+      function () {
+        Meteor._updateAslStore(CURRENT_VALUE_KEY_NAME, value);
+        Meteor._updateAslStore(UPPER_CALL_DYNAMICS_KEY_NAME, dynamics);
+        return func();
```
denihs, line 66: This will not work.

We need to wait for the `func()` to finish and reset the states of  `CURRENT_VALUE_KEY_NAME` and `UPPER_CALL_DYNAMICS_KEY_NAME` after the `func()` is done.

This is because you can call a `withValues` inside another, and you need to set and reset those states to keep context between calls.

Otherwise, you'll get issues like this: https://github.com/meteor/meteor/pull/13063

denihs (changes requested):

packages/meteor/dynamics_nodejs.js
zodern: #13063 changed it to create a new dynamics object each time instead of mutating the parent dynamics. This should remove the need to reset the state since the parent and children all have their own state. If you comment out the reset code, the example code from the PR description stills works correctly.

packages/meteor/dynamics_browser.js
```diff
@@ -36,11 +36,23 @@ EVp.getOrNullIfOutsideFiber = function () {
...
     var ret = func();
+    var isPromise = Meteor._isPromise(ret);
+
+    if (isPromise) {
+      var self = this;
+      return ret.finally(function () {
+        currentValues[self.slot] = saved;
+      });
+    }
   } finally {
```
zodern, line 52: The `meteor` package should support browsers that don't support optional catch binding. I'm surprised the check legacy syntax script doesn't catch this.

packages/meteor/dynamics_browser.js
```diff
@@ -36,11 +36,23 @@ EVp.getOrNullIfOutsideFiber = function () {
...
+
   try {
     currentValues[this.slot] = value;
+
     var ret = func();
+    var isPromise = Meteor._isPromise(ret);
+
+    if (isPromise) {
+      var self = this;
+      return ret.finally(function () {
```
zodern, line 48: The meteor package could run in environments that have promises, but does not have Promise.finally. The `promise` package would polyfill it, but apps aren't required to use it.

zodern:
One interesting side effect of this is tests now run inside a method.

Meteor 1/2 uses [`setImmediate`](https://github.com/meteor/meteor/blob/02aa69c0c2d2c3ea805ba8bcf0dbaa565ca995b3/packages/meteor/fiber_helpers.js#L124-L128) to schedule running the tests, which does not preserve the environment.

I'm wondering if there should be a new api to run in a fresh environment. Meteor 1/2 supported this by simply creating a new fiber and running code inside it, or using the api variants that do not preserve the environment (setTimeout instead of Meteor.setTimeout, setImmediate instead of Meteor.setImmediate, etc.). Meteor 3 doesn't have any public api's to do the same.

packages/meteor/dynamics_nodejs.js
leonardoventurini: I agree with @zodern, I will include that code as a test just to make sure.

.envrc
```diff
@@ -0,0 +1,32 @@
...
+  @meteor self-test
+}
+
+function @check-syntax {
+  node "$ROOT_DIR/scripts/admin/check-legacy-syntax/check-syntax.js"
+}
+
+function @generate-dev-bundle {
+  "$ROOT_DIR/scripts/generate-dev-bundle.sh"
+}
```
nachocodoner, line 32: ❤️

leonardoventurini:
@zodern This seems to work, what do you think?

https://github.com/meteor/meteor/pull/13128/commits/a8edea4e49a1ccfc1c2c1f537b981c6833c868ba

https://github.com/meteor/meteor/pull/13128/commits/52712eee6501b72eb7fafb299818aeef769f5fbc

zodern:
The implementation looks good. It probably should wrap [this code](https://github.com/meteor/meteor/blob/47e28615be2b680d0ca296a1ac7947db94476405/packages/meteor/async_helpers.js#L72-L78) or some other pat of the queue to maintain the behavior in Meteor 1/2 of not preserving the environment when running the queued tasks and so tests don't run inside a method.

denihs (dismissed):

denihs (approved):

PR #13137, opened 2024-05-10
Fix for old API usage

manueltimita:
Fixes "Error in OAuth Server: findOne + is not available on the server. Please use findOneAsync() instead."

The error stems from `google_server.js` still using the sync Mongo API at `ServiceConfiguration.configurations.findOne(...)`.


manueltimita:
This pull request is redundant. Issue fixed in https://github.com/meteor/meteor/pull/13127

Issue #13152, opened 2024-05-21
[Meteor: 3.0.0-rc1] - github-oauth Using `findOne` instead of `findOneAsync`


brunoeduardodev:
# Describing the Issue

On the current latest release, `Meteor: 3.0.0-rc1` the `github-oauth` package is throwing a `W20240521-00:24:30.052(-3) (oauth_server.js:412) Error in OAuth Server: findOne +  is not available on the server. Please use findOneAsync() instead.`

![image](https://github.com/meteor/meteor/assets/50559336/df2cdefd-29fc-44e0-a6c2-9dfd12bc6f7e)

### OS
MacOS, M1 Pro

### Expected Behavior
Github sign in to complete successfully.

### Actual Behavior
Github sign in is not completed due to the unexpected error on the server.

# How to reproduce

### **NOTE** 
To reproduce the issue when cloning the github repository, you need to fill the `settings.json` file with the "clientId" and "secret" with valid values of a github oauth application (with callback_url set to `http://localhost:300`)

## Reproducing Step by Step
1. Create a new meteor project using Meteor 3.0.0-rc1
```bash
meteor create meteor-rc-300.1-github-oauth-error --release 3.0-rc.1
```
2. Add `accounts-github` package
```bash
meteor add accounts-github
```

It should install the following packages and versions:

<img width="421" alt="image" src="https://github.com/meteor/meteor/assets/50559336/982b0c0a-e268-4b6f-96bf-b3b790725e42">

3. On `App.jsx` file, add a simple button that invokes `Meteor.loginWithGithub()`

```tsx
import React from 'react';
import { Hello } from './Hello.jsx';
import { Info } from './Info.jsx';
import { Meteor } from 'meteor/meteor';

export const App = () => (
  <div>
    <h1>Welcome to Meteor!</h1>
    <button onClick={() => Meteor.loginWithGithub()}>Login with GitHub</button>
    <Hello/>
    <Info/>
  </div>
);

```

4. Create a `settings.json` file on the root of the project containing the `clientId` and `secret` of a github oauth application with `callback_url` set to `http://localhost:3000`

```json
{
  "packages": {
    "service-configuration": {
      "github": {
        "loginStyle": "popup",
        "clientId": "<ADD CLIENT ID>",
        "secret": "<ADD SECRET>"
      }
    }
  }
}
```

5. Run Meteor passing the JSON file as a flag

```bash
meteor -s settings.json
```

[...]

PR #13153, opened 2024-05-21
fix: oauth_.fetch invocation on github_server and google_server

brunoeduardodev:
### Description

This PR fixes the invocation of the OAuth._fetch() method on the `github_server` and `google_server` of their oauth.
This fix is explained in more details on the Additional Notes of the Issue #13152.

I've set the base of the Pull Request as `release-3.0` since the `devel` branch seems to not be fully updated with the code on `release-3.0` (on the `devel` branch, this implementation was still using Fetch API and `findOne` invocations.)

### DIFF
--- a/packages/github-oauth/github_server.js
+++ b/packages/github-oauth/github_server.js
@@ -47,8 +47,8 @@ const getAccessToken = async (query) => {
     });
     const request = await OAuth._fetch(
       `https://github.com/login/oauth/access_token?${content.toString()}`,
+      'POST',
       {
-        method: 'POST',
         headers: {
           Accept: 'application/json',
           'User-Agent': userAgent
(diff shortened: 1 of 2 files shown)


nachocodoner (approved):
Thanks for this fix!

I adapted all account packages in my PR https://github.com/meteor/meteor/pull/13127. During this process, I provided tests with mocks, since these packages had side effects involving external cloud services that can't be tested easily. To do this, I used the standard `OAuth._fetch`, as in other packages. However, I missed an interface difference in the process.

I will pay more attention to interfaces in Meteor codebase, which lacks clear type-checking, above all on refactors on packages without coverage. And of course, more manual testing even if many scenarios affected. Regarding tests on this regression, I can't think of a way to properly verify behavior other than parameter checking, but I don't think it's necessary.

denihs (approved):

Issue #13101, opened 2024-04-19
[Meteor 3] WebApp.addRuntimeConfigHook is broken


ebroder:
The `WebApp.addRuntimeConfigHook` function seems to be broken in Meteor 3.0-rc.0. I've put together a very simple reproducer app at https://github.com/ebroder/meteor-runtime-config-test. (The hook itself is [here](https://github.com/ebroder/meteor-runtime-config-test/blob/d44a016b4d67416eef90531f5423ac56ed811b32/server/main.js#L3C1-L8))

If you open the app, the rendered HTML includes the following:

```
  <script type="text/javascript">__meteor_runtime_config__ = JSON.parse(decodeURIComponent([object Promise]))</script>
```

and, of course, that cascades into a bunch of errors:

```
Uncaught SyntaxError: Unexpected identifier 'Promise' (at (index):24:100)
Uncaught ReferenceError: __meteor_runtime_config__ is not defined
Uncaught TypeError: Cannot read properties of undefined (reading 'Meteor')
Uncaught ReferenceError: meteorInstall is not defined
```

It seems like something isn't being `await`ed that should be, but I'm pretty sure it's happening outside of my hook. (If I add logging to the hook, it seems like I get valid configuration passed in)

ebroder:
Oh I just realized this is a duplicate of the bug my colleague opened, #12939. Would be great to see it fixed before 3.0 final!

Issue #12939, opened 2023-12-31
[Meteor 3] Calling WebApp.addRuntimeConfigHook breaks __meteor_runtime_config__


zarvox:
Affected versions: 3.0-rc.0 and 3.0-beta.0 are impacted (and presumably the other 3.0 alphas are all impacted too).  Meteor 2.14 is not.

OS: all (verified on macOS 14.2.1 and Fedora Linux 39)

Reproduction repository: https://github.com/zarvox/meteor-trivial-runtime-config-hook

Steps to reproduce:

1. Use a meteor-3.0 alpha or beta or rc.0
2. Make any call to `WebApp.addRuntimeConfigHook()` in server code.  The hook can be an empty function; it doesn't matter what it is.
3. Run the app (`meteor`)
4. attempt to load the page at http://localhost:3000
5. observe a blank white page and errors on the JS console

I believe what's happening is that `packages/webapp/webapp_server.js` [calls runtimeConfig.hooks.forEach](https://github.com/meteor/meteor/blob/162248110d340d5d0ef7a29ca4af871f0e38e2df/packages/webapp/webapp_server.js#L430), and `Hook` [defaults to binding the Meteor environment](https://github.com/meteor/meteor/blob/162248110d340d5d0ef7a29ca4af871f0e38e2df/packages/callback-hook/hook.js#L47), which results in [wrapping whatever callback the user provided in Meteor.bindEnvironment](https://github.com/meteor/meteor/blob/162248110d340d5d0ef7a29ca4af871f0e38e2df/packages/callback-hook/hook.js#L76) which [seems to return a Promise under meteor 3](https://github.com/meteor/meteor/blob/162248110d340d5d0ef7a29ca4af871f0e38e2df/tools/utils/fiber-helpers.js#L117-L147).  Then, when rendered as a string, we get `[object Promise]` rather than the intended result of running the `hook` function, which gets propagated into the line 

```html
  <script type="text/javascript">__meteor_runtime_config__ = JSON.parse(decodeURIComponent([object Promise]))</script>
``` 

which is a syntax error.  Later, other parts of the meteor runtime expect that variable to be defined, and those error out too.

I'm not particularly familiar with the expected/intended behavior of `bindEnvironment` in the fiberless world, but if that's behaving as designed, then we probably want something like the following patch in `packages/webapp/webapp_server.js` to actually await each hook's completion:

```diff
diff --git a/packages/webapp/webapp_server.js b/packages/webapp/webapp_server.js
[...]

PR #13156, opened 2024-05-21
Fix WebApp.addRuntimeConfigHook

nachocodoner:
## Issue

This issue was reported and reproduced https://github.com/meteor/meteor/issues/13101 and https://github.com/meteor/meteor/issues/12939.

## Fix

The provided fix is to make `WebApp.addRuntimeConfigHook` async compatible.

The callbacks provided to this function [describe hooks](https://github.com/meteor/meteor/blob/devel/packages/webapp/webapp_server.js#L357) and [these wraps async by default](https://github.com/meteor/meteor/blob/devel/packages/callback-hook/hook.js#L82). This is a behavior we must to keep to offer flexibility on use configurable async flows.

- [x] Provide a regression test to cover  `addRuntimeConfigHook` use case

I served from [this doc](https://v3-docs.meteor.com/packages/webapp.html#dynamic-runtime-configuration) to understand how to use  `addRuntimeConfigHook` and provide the missing test scenario.

### DIFF
--- a/packages/webapp/webapp_tests.js
+++ b/packages/webapp/webapp_tests.js
@@ -354,9 +354,7 @@ Tinytest.addAsync(
         ...WebApp.decodeRuntimeConfig(config.encodedCurrentConfig),
         customKey: 'customValue',
       };
-      return JSON.stringify(
-        encodeURIComponent(JSON.stringify(nextConfig))
-      );
+      return WebApp.encodeRuntimeConfig(nextConfig);
     });
 
     const req = new http.IncomingMessage();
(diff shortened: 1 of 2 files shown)


packages/webapp/webapp_tests.js
```diff
@@ -343,3 +345,23 @@ Tinytest.add("webapp - npm modules", function (test) {
   test.matches(WebAppInternals.NpmModules.express.version, /^4\.(\d+)\.(\d+)/);
   test.equal(typeof(WebAppInternals.NpmModules.express.module), 'function');
 });
+
+Tinytest.addAsync(
+  "webapp - addRuntimeConfigHook usage",
```
henriquealbert, line 350: ♥️

henriquealbert (approved):
lgtm

leonardoventurini (approved):

PR #13162, opened 2024-05-23
Implement reset db option and default without cleaning db

nachocodoner:
Context: https://forums.meteor.com/t/post-your-tips-for-updating-to-meteor-v3/61625/16 and https://github.com/meteor/meteor/issues/7789

## Implementation

There are two ways to implement this to ensure by default the db is not removed.

1) Fetch beforehand all the folders and files within `~/.meteor/local` and trigger a promise in parallel per subfolder that is not the `~/.meteor/local/db` path
2) Upgrade `rimraf` to latest as they offer a way to filter the files to remove within a specific folder, and use it directly. The [dev_bundle is stuck into an old version](https://github.com/meteor/meteor/blob/devel/scripts/dev-bundle-tool-package.js#L52) of `rimraf@2.6.2` that [doesn't support a way to filter](https://github.com/isaacs/rimraf/tree/79b933fb362b2c51bedfa448be848e1d7ed32d7e).

I decided to go with 1), since 2) requires to generate a new bundle and update any usage. Though I think is straightforward since only [one spot would need to be changed](https://github.com/meteor/meteor/blob/devel/tools/fs/files.ts#L340-L342). 1) keeps being easier implementation and performance-wise should be fine.

## Pending

- [x] Can we cover `meteor reset [--db]` as a self-test for coverage?

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1731,11 +1731,14 @@ main.registerCommand({
   // Doesn't actually take an argument, but we want to print an custom
   // error message if they try to pass one.
   maxArgs: 1,
+  options: {
+    db: { type: Boolean },
+  },
   requiresApp: true,
   catalogRefresh: new catalog.Refresh.Never()
 }, async function (options) {
   if (options.args.length !== 0) {
-    Console.error("meteor reset only affects the locally stored database.");
+    Console.error("meteor reset might delete the local database using the --db option.");
     Console.error();
     Console.error("To reset a deployed application use");
     Console.error(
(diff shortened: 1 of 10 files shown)


v3-docs/v3-migration-docs/index.md
```diff
@@ -29,7 +29,7 @@ You can upgrade your Meteor 2.x project by running the command below inside your
 
 ```bash
 meteor update --release 3.0-rc.2
-meteor reset #resets local DB and project to a fresh state
+meteor reset #resets project to a fresh state
```
nachocodoner, line 32: I don't think is required in this moment to clean the local database from your project to do the upgrade from v2 to v3. It will when doing it in a future version that is a mismatch between the mongo driver versions.

Should we add the option `--db` here and keep the warn about reset of the local db for also future references?

tools/cli/commands.js
```diff
@@ -1731,11 +1731,14 @@ main.registerCommand({
...
   maxArgs: 1,
+  options: {
+    db: { type: Boolean },
+  },
   requiresApp: true,
   catalogRefresh: new catalog.Refresh.Never()
 }, async function (options) {
   if (options.args.length !== 0) {
-    Console.error("meteor reset only affects the locally stored database.");
+    Console.error("meteor reset might delete the local database using the --db option.");
```
leonardoventurini, line 1741: This warning sounds strange, will this be only shown if the user adds `--db` or is it easy to make it so? Perhaps we can keep the original message then...

denihs (approved):

leonardoventurini (dismissed):

tools/cli/commands.js
nachocodoner: This appears when you use `meteor reset XXX` being XXX anything that is not an option.

image

I may rephrase like this:

`'meteor reset' command only affects the local project cache. Use the --db option to remove the local database.`

or how would you suggest?

leonardoventurini: I think that makes sense

denihs (approved):

leonardoventurini (approved):

Issue #13143, opened 2024-05-14
Meteor: 3.0.0-rc1 Package: allow-deny -> client side collection insert/update/remove not working as expected


acemtp:
When executing the following line on the client with meteor 3, it should work since minimongo doesn't require async function calls.

`Meteor.users.update(Meteor.userId(), { $set: { 'profile.ok': 1 }})`

It generates this error on the server:

```
 Exception while invoking method '/users/update' Error: findOne +  is not available on the server. Please use findOneAsync() instead.
     at Object.ret.<computed> (packages/mongo/remote_collection_driver.js:52:15)
     at Object.<anonymous> (packages/matb33:collection-hooks/findone.js:27:22)
     at Object.collection.<computed> [as findOne] (packages/matb33:collection-hooks/collection-hooks.js:116:21)
     at CollectionPrototype._validatedUpdate (packages/allow-deny/allow-deny.js:485:32)
     at MethodInvocation.m.<computed> (packages/allow-deny/allow-deny.js:193:46)
```

The reason is that in allow-deny.js of the AllowDeny package, the old method _validatedUpdate() is called but contains a findOne that is not Async.

Current Fix:

On our server code, we added this lines to override the deprecated function

```
if (Meteor.isServer) {
  Mongo.Collection.prototype._validatedInsert = Mongo.Collection.prototype._validatedInsertAsync;
  Mongo.Collection.prototype._validatedUpdate = Mongo.Collection.prototype._validatedUpdateAsync;
  Mongo.Collection.prototype._validatedRemove = Mongo.Collection.prototype._validatedRemoveAsync;
}
```

Solution 1:

Update the code of _validatedInsert _validatedUpdate _validatedRemove in allow-deny.js to handle async methods

Solution 2 (better):

remove _validatedInsert _validatedUpdate _validatedRemove function and always use _validatedInsertAsync _validatedUpdateAsync _validatedRemoveAsync in both case update and updateAsync.

nachocodoner:
Thank you for the report.

It's surprising we have not run into this issue in our automated test coverage, but we'll make sure to add it and include the fix in our nexts RCs.

denihs:
Yes, it is strange... Anyways, let's fix it!

nachocodoner:
FYI, this issue is being addressed [in this PR](https://github.com/meteor/meteor/pull/13168) and is scheduled for release in the next RC.3.

[...]

PR #13168, opened 2024-05-29
[3.0] AllowDeny async compatibility

nachocodoner:
This PR reviews the allow-deny package to make it async-compliant.

Two issues have been reported by the community. Feel free to report any other issues in the comments for the next RC.

## Issue 1

Context: https://github.com/meteor/meteor/issues/13143

Client-side collection operations trigger server failures. The server uses the `_validate<OP>` sync method, causing the error `Error: findOne + is not available on the server. Please use findOneAsync() instead`.

We now enforce async validation with proper test coverage.

- [x] Provide test case and fix for proper operation with async validation

## Issue 2

Context: https://forums.meteor.com/t/meteor-3-0-rc-1-is-out/61580/9

Async definitions on allow/deny rules don't work.

- [x] Provide test case and fix for async definitions on allow/deny rules.

### DIFF
--- a/packages/mongo/collection.js
+++ b/packages/mongo/collection.js
@@ -944,8 +944,7 @@ Object.assign(Mongo.Collection.prototype, {
 
     if (this._isRemoteCollection()) {
       const args = [selector, modifier, options];
-
-      return this._callMutatorMethod('update', args);
+      return this._callMutatorMethod('update', args, callback);
     }
 
     // it's my collection.  descend into the collection object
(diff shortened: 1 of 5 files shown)


packages/mongo/collection.js
```diff
@@ -1000,12 +999,13 @@ Object.assign(Mongo.Collection.prototype, {
...
    * @param {MongoSelector} selector Specifies which documents to remove
+   * @param {Function} [callback] Optional.  If present, called with an error object as the first argument and, if no error, the number of affected documents as the second.
    */
-  remove(selector) {
+  remove(selector, callback) {
     selector = Mongo.Collection._rewriteSelector(selector);
 
     if (this._isRemoteCollection()) {
-      return this._callMutatorMethod('remove', [selector]);
+      return this._callMutatorMethod('remove', [selector], callback);
```
nachocodoner, line 1008: We have decribed in the documentation [the callback option for remove behavior](https://docs.meteor.com/api/collections#Mongo-Collection-remove). However, the code was not properly passing. Feels weird that this was not working and there is no test coverage, at least on Meteor 3.0. Now it does.

acemtp:
Excellent, seems to work fine, you can close the issue.

In the doc below you still have insert/update/remove in allow deny, but in which cases they are called since now calling collection.insert or collection.insertAsync in client side always call the allow.insertAsync function?

https://v3-docs.meteor.com/api/collections.html#Mongo-Collection-allow

PR #13194, opened 2024-06-13
Initialize options for accounts from settings.json

StorytellerCZ:
Noticed this earlier that `Meteor.settings.packages.accounts-base` is not properly initialized on server. I think we used to have this and it might have gotten lost, so I've looked through things and made fixes to ensure those settings get picked up properly.


packages/accounts-base/accounts_common.js
```diff
@@ -274,15 +282,15 @@ export class AccountsCommon {
     }
 
     // Validate config options keys
-    Object.keys(options).forEach(key => {
+    for (const key of Object.keys(options)) {
```
denihs, line 285: Why use `for of` instead of `forEach`?

packages/accounts-base/server_main.js
```diff
@@ -4,9 +4,9 @@ import { AccountsServer } from "./accounts_server.js";
  * @namespace Accounts
  * @summary The namespace for all server-side accounts-related methods.
  */
-Accounts = new AccountsServer(Meteor.server, Meteor.settings.packages?.accounts || {});
+Accounts = new AccountsServer(Meteor.server, Meteor.settings.packages?.['accounts-base'] || {});
 // TODO[FIBERS]: I need TLA
-Accounts.init().then()
+Accounts.init().then();
```
denihs, line 9: Can you check if TLA works here?

packages/accounts-base/server_main.js
```diff
@@ -4,9 +4,9 @@ import { AccountsServer } from "./accounts_server.js";
  * @namespace Accounts
  * @summary The namespace for all server-side accounts-related methods.
  */
-Accounts = new AccountsServer(Meteor.server, Meteor.settings.packages?.accounts || {});
+Accounts = new AccountsServer(Meteor.server, Meteor.settings.packages?.['accounts-base'] || {});
```
denihs, line 7: So, there isn't an `accounts` option that can be provided here? :thinking:

If someone already provides `accounts` as a setting, wouldn't this break their code?

packages/accounts-base/accounts_common.js
StorytellerCZ: Better performance according to: https://romgrk.com/posts/optimizing-javascript

packages/accounts-base/server_main.js
StorytellerCZ: There was no documented `accounts`, it was actually `accounts-base`, but I'm fine with using both.

packages/accounts-base/server_main.js
StorytellerCZ: I will, but that is out of scope of this PR.

StorytellerCZ: Run Meteor from checkout with `await Accounts.init();` and didn't get any error. I can change the code here if you want.

denihs: We can leave it to another PR

denihs (approved):

PR #13198, opened 2024-06-19
- update call-x-callAsync.md docs to mirror our API

denihs:
A [report](https://forums.meteor.com/t/async-calls-in-meteor-3/61769) was in the forums questioning how valid the `Meteor.callAsync` migration doc was. 

It was confusing and didn't seem to be true with the current API.

I took the time to retest all the examples there and rewrite them to reflect better our API.

[Here](https://deploy-preview-13198--v3-migration-docs.netlify.app/breaking-changes/call-x-callAsync.html), you can see the preview.


v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -1,29 +1,25 @@
...
-
 ```js [v2-client.jsx]
-import { Meteor } from 'meteor/meteor'
+import { Meteor } from "meteor/meteor";
 
 let data, error;
 
-Meteor.call('getAllData', (err, res) => { // [!code highlight]
+Meteor.call("getAllData", (err, res) => {
+  // [!code highlight]
```
henriquealbert, line 22:
```suggestion
Meteor.call("getAllData", (err, res) => {   // [!code highlight]
```

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +28,209 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
 
+Meteor.call("SOME_SYNC_METHOD");
+Meteor.callAsync("SOME_ASYNC_METHOD_WITH_STUB");
+Meteor.call("SOME_SYNC_METHOD");
+Meteor.callAsync("SOME_ASYNC_METHOD_WITH_STUB");
+Meteor.callAsync("SOME_ASYNC_METHOD_WITHOUT_STUB");
 ```
+
+As you can see, if you follow the rule of calling async methods with `Meteor.callAsync` and sync methods with 
+`Meteor.call` you can use them as you want. But ideally, after you finish your migration, you should have any `Meteor.call`
```
jamauro, line 235: > ideally, after you finish your migration, you should have any `Meteor.call`

Shouldn't `should` be `shouldn't` here? :) 

Also, I suggest putting this near the top of the "Rules of using Meteor.callAsync & Meteor.call" so it reads something like:

> Technically, `Meteor.callAsync` and `Meteor.call` can be used alongside each other. But ideally, after you finish your migration, you shouldn't use `Meteor.call`, so you don't need to worry about the gotcha scenarios detailed below.

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +28,209 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
 ```js
-import { Meteor } from 'meteor/meteor'
+import { Meteor } from "meteor/meteor";
+
+Meteor.callAsync("SOME_ASYNC_METHOD_WITH_STUB")
+  .then((data) => console.log(data))
+  .catch((err) => console.log(err));
 
-Meteor.call("someMethod", (err, res) => { // [!code error] This is not ok
+Meteor.call("SOME_ASYNC_METHOD_WITH_STUB", (err, res) => {
```
zodern, line 90: I'm assuming the stub used with `Meteor.call` is async. This seems to be broken due to the changes in `EV.withValue` to support promises on the client: https://github.com/meteor/meteor/blob/818c5da95ca5b884e61ad153ad27ccc7ec86741b/packages/meteor/dynamics_browser.js#L46-L52

Before, async stubs were not supported in `Meteor.apply`. With this change, they still aren't really supported, but it introduces the same issues we had with `Meteor.callAsync` in Meteor 2 that the queueing fixed.

 If we do want to support them, [this code](https://github.com/meteor/meteor/blob/818c5da95ca5b884e61ad153ad27ccc7ec86741b/packages/ddp-client/common/livedata_connection.js#L612-L620) would need to be updated, as well as how the queueing handles `Meteor.apply`.

Though I don't think we should support async stubs in `Meteor.apply`.

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +28,209 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
   if (err) {
     console.log(err);
   } else {
     console.log(res);
   }
 });
+```
 
+**Case 2**: Calling `Meteor.call` and then right after calling `Meteor.callAsync` when the methods have stubs.
```
zodern, line 99: Is the stub used with `Meteor.call` here async? If it is, this is the same issue as with case 1 - `Meteor.call` needs to either not be used with async stubs, or should be fixed so it queues correctly with async stubs.

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +28,209 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
+});
+```
+
+**Case 4**: Calling `Meteor.call` and then right after calling `Meteor.callAsync` but the method called with `Meteor.call`
+does not have stub.
+
+```js
+import { Meteor } from "meteor/meteor";
+
+Meteor.call("SOME_ASYNC_METHOD_WITHOUT_STUB", (err, res) => {
```
zodern, line 142: I think using `ASYNC_METHOD` makes this a little confusing. The client isn't affected by if the method is async or not on the server. It might be better to instead say if the stub is async or sync, since that is what affects the client.

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +28,209 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
+  if (err) {
+    console.log(err);
+  } else {
+    console.log(res);
+  }
+});
+```
+
+**Case 4**: Calling `Meteor.call` and then right after calling `Meteor.callAsync` but the method called with `Meteor.call`
```
zodern, line 136: What is the problem with case 4? This should be supported.

zodern:
Do you have a reproduction for cases 3 and 4? It should be possible to fully support them.

Cases 1 and 2 seem to be issues with using async stubs in `Meteor.call`/`Meteor.apply`. I thought that we weren't supporting async stubs in `Meteor.call`/`Meteor.apply`, though it seems it wasn't really documented outside of the 2.8 migration guide.

In Meteor 2, there were 3 issues with using async stubs with `Meteor.apply` (which is also used by `Meteor.call`):
- it didn't wait for the stub to finish before sending the method call to the server
- the `_CurrentMethodInvocation` was synchronously reset, so part of the async stub would run after the simulation finished
- some `Meteor.apply` options are incompatible with async stubs, or at least would require breaking changes

After the client changes in #13128, `_CurrentMethodInvocation` now waits to reset until after the async stub finishes. This can cause more issues than just breaking subsequent method calls - any code that uses api's that behave differently in a simulation will be broken if it runs before the simulation finishes.

The queueing the ddp-client does to fix this for `callAsync`/`applyAsync` delays running the stub until when it can be sure other code won't run at the same time. This isn't possible for `call`/`apply` due to:
1. some of the options for `apply` require running the stub syncronously
2. with sync stubs and `call`/`apply`, you can call the method, and immediately use any changes made to the collections. This would be broken if the stub ran later.

I don't think the current state of partially allowing async stubs with `call`/`apply` is good. I think we should:
1. Fix how they handle async stubs so `_CurrentMethodInvocation` is synchronously reset, so using an async stub doesn't break other code in the app
2. Log a warning when they are used with an async stub
3. Make it clearer in the docs that they do not support async stubs

leonardoventurini:
@zodern what do you think of us deprecating the `call` method (logging a warning etc) and instruct people to use `callAsync` from now on? The information in the docs regarding this is very confusing to a regular user

zodern:
In the new docs in this PR, most of the complexity is due to `_CurrentMethodInvocation` not synchronously resetting the value in `Meteor.apply`. If that is fixed, I think all of the information on using Meteor.call and Meteor.callAsync together could be removed since it would always work.

Deprecating `Meteor.call` and `Meteor.apply` probably makes sense. Though currently the queue that fixes the issues with `callAsync` and `applyAsync` is only used when the DDP Client is created on the server, so you probably will first want to add it for DDP Client's created on the server (or fully integrating the queue into the DDP client instead of having it patch during runtime).

Regardless of if it is deprecated, I think the three things I suggested for `call`/`apply` should be done to help minimize the problems caused by using them with async stubs.

denihs:
> Do you have a reproduction for cases 3 and 4?

Hi @zodern, yes!

I've tested all those scenarios. Even though we don't get errors with cases 3 and 4 when we run like that, the method with the stub doesn't reach the server, and the operation does not persist.

Methods declared in the server and client:
```js
Meteor.methods({
  async addLink(suffix) {
    const random = Math.random() * 10;

    const all = await LinksCollection.find({}).fetchAsync()
    console.log(`there is ${all.length} links`);
    return LinksCollection.insertAsync({
      title: `[${suffix}] Random link ${random}`,
      url: `https://${random}.com`,
    })
  },
  async asyncMethodWithoutStub(calledWith) {
    const url = 'https://api.coindesk.com/v1/bpi/currentprice.json';

    try {
      const response = await fetch(
        url);
      const data = JSON.stringify(response.data);
      console.log('asyncWithoutStubMethod DATA?', data);
      return data;
    } catch (error) {
      throw new Meteor.Error('API call failed', error.message);
    }
  }
})
```

Calling these methods in the client:

```js
    Meteor.call("asyncMethodWithoutStub", "CALL", (err, res) => {
      if (err) {
        console.log(err);
      } else {
        console.log(res);
      }
    })
    Meteor.callAsync("addLink", "CALLASYNC")
```

If you run this, you'll see that the link is added to the client, but the data does not persist, so the link disappears after a refresh.

zodern:
Thanks @denihs. I'm unable to reproduce with this code:

<details>
  <summary>Code</summary>
  
```js
const LinksCollection = new Mongo.Collection('links');

if (Meteor.isServer) {
  Meteor.publish(null, () => LinksCollection.find());
  Meteor.methods({
    async asyncMethodWithoutStub(calledWith) {
      const url = 'https://api.coindesk.com/v1/bpi/currentprice.json';

      try {
        const response = await fetch(
          url);
        const data = JSON.stringify(response.data);
        console.log('asyncWithoutStubMethod DATA?', data);
        return data;
      } catch (error) {
        throw new Meteor.Error('API call failed', error.message);
      }
    }
  });
}

Meteor.methods({
  async addLink(suffix) {
    const random = Math.random() * 10;

    const all = await LinksCollection.find({}).fetchAsync();
    console.log(`there is ${all.length} links`);
    return LinksCollection.insertAsync({
      title: `[${suffix}] Random link ${random}`,
      url: `https://${random}.com`,
    });
  }
});

if (Meteor.isClient) {
  import { Tracker } from 'meteor/tracker';

  Tracker.autorun(() => {
    console.log('links', LinksCollection.find().fetch());
  });

  Meteor.call("asyncMethodWithoutStub", "CALL", (err, res) => {
    if (err) {
      console.log(err);
    } else {
      console.log(res);
    }
  });
  Meteor.callAsync("addLink", "CALLASYNC").then(() => console.log('done'), console.log);
}

```
</details>

However, if I define `asyncMethodWithoutStub` on the client so it has an async stub, I am able to reproduce:
<details><summary>Reproduction code</summary>

```js
const LinksCollection = new Mongo.Collection('links');

if (Meteor.isServer) {
  Meteor.publish(null, () => LinksCollection.find());
}

Meteor.methods({
  async asyncMethodWithoutStub(calledWith) {
    const url = 'https://api.coindesk.com/v1/bpi/currentprice.json';

    try {
      const response = await fetch(
        url);
      const data = JSON.stringify(response.data);
      console.log('asyncWithoutStubMethod DATA?', data);
      return data;
    } catch (error) {
      throw new Meteor.Error('API call failed', error.message);
    }
  }
});

Meteor.methods({
  async addLink(suffix) {
    const random = Math.random() * 10;

    const all = await LinksCollection.find({}).fetchAsync();
    console.log(`there is ${all.length} links`);
    return LinksCollection.insertAsync({
      title: `[${suffix}] Random link ${random}`,
      url: `https://${random}.com`,
    });
  }
});

if (Meteor.isClient) {
  import { Tracker } from 'meteor/tracker';

  Tracker.autorun(() => {
    console.log('links', LinksCollection.find().fetch());
  });

  // delay so we don't cause issues with Meteor throwing an error due to the
  // DDP connection connecting during the simulation
  setTimeout(() => {
    Meteor.call("asyncMethodWithoutStub", "CALL", (err, res) => {
      if (err) {
        console.log(err);
      } else {
        console.log(res);
      }
    });
    Meteor.callAsync("addLink", "CALLASYNC").then(() => console.log('done'), console.log);
  }, 1000);
}
```
</details>

Though this makes it the same as case 2.

There's two problems here:
1. The call to `Meteor.callAsync` happens during the simulation for `asyncMethodWithoutStub` so it is never sent to the server
2. The additional link should have been removed when the client received the server result for `asyncMethodWithoutStub` since it did not add the link. However, for some reason the link is left on the client.

If `_CurrentMethodInvocation` is modified back to it's old behavior, both issues are fixed:

<details><summary>Working code</summary>

```js
const LinksCollection = new Mongo.Collection('links');

if (Meteor.isServer) {
  Meteor.publish(null, () => LinksCollection.find());
}

Meteor.methods({
  async asyncMethodWithoutStub(calledWith) {
    const url = 'https://api.coindesk.com/v1/bpi/currentprice.json';

    try {
      const response = await fetch(
        url);
      const data = JSON.stringify(response.data);
      console.log('asyncWithoutStubMethod DATA?', data);
      return data;
    } catch (error) {
      throw new Meteor.Error('API call failed', error.message);
    }
  }
});

Meteor.methods({
  async addLink(suffix) {
    const random = Math.random() * 10;

    const all = await LinksCollection.find({}).fetchAsync();
    console.log(`there is ${all.length} links`);
    return LinksCollection.insertAsync({
      title: `[${suffix}] Random link ${random}`,
      url: `https://${random}.com`,
    });
  }
});

if (Meteor.isClient) {
  import { Tracker } from 'meteor/tracker';

  Meteor.EnvironmentVariable.prototype.withValue = function (value, func) {
    let currentValues = this.getCurrentValues();
    var saved = currentValues[this.slot];

    try {

      return func();
    } finally {
      currentValues[this.slot] = saved;
    }
  };

  Tracker.autorun(() => {
    console.log('links', LinksCollection.find().fetch());
  });

  // delay so we don't cause issues with Meteor throwing an error due to the
  // DDP connection connecting during the simulation
  setTimeout(() => {
    Meteor.call("asyncMethodWithoutStub", "CALL", (err, res) => {
      if (err) {
        console.log(err);
      } else {
        console.log(res);
      }
    });
    Meteor.callAsync("addLink", "CALLASYNC").then(() => console.log('done'), console.log);
  }, 1000);
}
```

denihs:
I see! In any case, I think we should continue on the path of deprecating `.call` and `.apply`. In this case, it seems to me we don't need to worry about those cases, right?

zodern:
Regardless of if they are deprecated, I still think they should be fixed. There's quite a few issues with the async versions in Meteor 2, and the issues still exist for DDP Clients on the server in Meteor 3, so it might be a while before apps are fully migrated off of it. It also simplifies the documentation and avoids potential issues apps might have during the migration.

denihs:
I agree. This change was introduced on https://github.com/meteor/meteor/pull/13128 by @leonardoventurini. But we can revert. I tested it, and it works like a charm. But I'm going to do this in a different PR. For this one, I will update this doc to say that `.call` and `.apply` are now deprecated.

But just to make sure I understood everything you suggested, we just need to revert `EVp.withValue` back to its original state to fix all these issues, right?

I tested only this change, and it fixes those issues, but when you said, "so you probably will first want to add it for DDP Client's created on the server (or fully integrating the queue into the DDP client instead of having it patch during runtime).", I got confused.

denihs:
I just updated the doc to reflect what we've been talking about.

denihs:
@zodern I opened this PR to address the other part https://github.com/meteor/meteor/pull/13201

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -1,29 +1,26 @@
...
 
+In fact, we log a warning message if you use `Meteor.call` to call an async method.
 
-::: tip
+If you use `Meteor.call` with async methods it can lead to unexpected behavior.
 
-It is recommened to use `Meteor.callAsync` instead of `Meteor.call` because of how our
-async API works. `Meteor.call` is still available but it is not recommended to use it, it
-can lead to unexpected behavior.
+`Meteor.callAsync` is the standard for calling async methods.
```
zodern, line 9: It might be more accurate as something like:

```
You can, but we recommend you use it only to call methods that do not have a method stub, or when the method stub is synchronous.

In fact, we log a warning message if you use `Meteor.call` to call a method with an async method stub since it can lead to unexpected behavior.

 `Meteor.callAsync` is the standard for calling methods and supports any method, including those that have an async method stub.
```

It might also be good to have a reminder of what a method stub is.

zodern:
> I tested only this change, and it fixes those issues, but when you said, "so you probably will first want to add it for DDP Client's created on the server (or fully integrating the queue into the DDP client instead of having it patch during runtime).", I got confused.

The issues with `callAsync`/`applyAsync` are fixed by the code in https://github.com/meteor/meteor/blob/release-3.0/packages/ddp-client/client/queueStubsHelpers.js. This code is only used for DDP Client's created on the client, so any DDP Client's created on the server still have the issues with `callAsync`/`applyAsync`.

It probably would be best if the code was integrated into the [livedata connection](https://github.com/meteor/meteor/blob/release-3.0/packages/ddp-client/common/livedata_connection.js) instead of patching it. Though using the patching code on the server in addition to the client would work too.

While you are changing it, it would be nice if it  linked to Meteor's docs instead of my repo:
https://github.com/meteor/meteor/blob/818c5da95ca5b884e61ad153ad27ccc7ec86741b/packages/ddp-client/client/queueStubsHelpers.js#L124

Those limitations should also probably be mentioned in the migration guide. The limitations mostly already existed in Meteor 2, but weren't documented and the warning is new in Meteor 3.

package.json
```diff
@@ -30,7 +30,8 @@
     "eslint-plugin-react": "^7.32.2",
     "eslint-plugin-react-hooks": "^4.6.0",
     "prettier": "^2.8.6",
-    "typescript": "^5.4.5"
+    "typescript": "^5.4.5",
+    "vitepress": "^1.2.3"
```
nachocodoner, line 34: do we have to add vitepress here? I think this dependency takes part already from docs folders, right?

denihs: It does! Just remove it.

leonardoventurini (dismissed):

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -1,29 +1,26 @@
...
 
+In fact, we log a warning message if you use `Meteor.call` to call a method with an async method stub since it can lead to unexpected behavior.
 
-::: tip
+`Meteor.callAsync` is the standard for calling methods and supports any method, including those that have an async method stub.
 
-It is recommened to use `Meteor.callAsync` instead of `Meteor.call` because of how our
-async API works. `Meteor.call` is still available but it is not recommended to use it, it
-can lead to unexpected behavior.
+Here is also important to remember what a stub is. A stub is a client-side simulation of the server-side method that runs immediately when the method is invoked, allowing the client to update its state optimistically before receiving the server's response. So, basically any MongoDB function (insertAsync, updateAsync, etc...) that's called inside a method it's considered a stub.
```
zodern, line 9:
```suggestion
Here is also important to remember what a stub is. A stub is a client-side simulation of the server-side method that runs immediately when the method is invoked, allowing the client to update its state optimistically before receiving the server's response. So, basically any Meteor method that is defined on the client is considered a stub.
```

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +29,66 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
-  Meteor.callAsync('someMethod').then(data2 => {
-    console.log(data2);
-  });
-});
+- fetch/XMLHttpRequest
+- setTimeout or setImmediate
+- indexedDB
+- web workers
+- any other web api that schedules macrotasks
+- Using these API's could allow other code to run before the async stub finishes.
```
zodern, line 88:
```suggestion

Using these API's could allow other code to run before the async stub finishes.
```

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +29,66 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
-  console.log(data);
-  Meteor.callAsync('someMethod').then(data2 => {
-    console.log(data2);
-  });
-});
+- fetch/XMLHttpRequest
+- setTimeout or setImmediate
+- indexedDB
+- web workers
+- any other web api that schedules macrotasks
```
zodern, line 87: I'm not sure if this is any better:

```suggestion
- any other async web api's that wait on macrotasks
```

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
```diff
@@ -32,105 +29,66 @@ Meteor.call('getAllData', (err, res) => { // [!code highlight]
...
-::: tip 
-It is not recommended to use concurrent calls.
-Use `await` for your `Meteor.callAsync`.
-:::
-
-here are a few examples of cases where you should use `Meteor.callAsync` instead of `Meteor.call`:
-
-```js
-import { Meteor } from 'meteor/meteor'
+# What are the limitations of call Meteor.callAsync
```
zodern, line 73:
```suggestion
# What are the limitations of Meteor.callAsync
```

packages/ddp-client/client/queueStubsHelpers.js
```diff
@@ -121,7 +121,7 @@ export const loadAsyncStubHelpers = () => {
         Meteor._setImmediate(() => {
           if (hasStub && !finished) {
             console.warn(
-              `Method stub (${name}) took too long and could cause unexpected problems. Learn more at https://github.com/zodern/fix-async-stubs/#limitations`
+              `Method stub (${name}) took too long and could cause unexpected problems. Learn more at https://v3-migration-docs.meteor.com/breaking-changes/call-x-callAsync.html#what-are-the-limitations-of-call-meteor-callasync`
```
zodern, line 124: It's probably out of scope for this PR, but if we are going to link to the docs from console messages, we probably need a solution to make the links stable over time.

 For example, something like `docs.meteor.com/link/long-method-stub` that redirects to the current docs.

denihs: That's a good idea

v3-docs/v3-migration-docs/breaking-changes/call-x-callAsync.md
leonardoventurini: @denihs does this make sense? Why is it? @henriquealbert

leonardoventurini (approved):

leonardoventurini (approved):

PR #13199, opened 2024-06-20
Fix Node.js version issue

leonardoventurini:
If you ran `meteor node` in a project which was not initialized/had its bundle downloaded, it would revert to the Node.js version of the bundle installed by default.

The solution: when you run `meteor node` it will first try to run the binary command logic, if it doesn't find the right bundle, it will revert to the normal command flow from meteor and inside it, after the springboarding logic, it will try to run the binary commands again. The springboarding logic is responsible for downloading the `meteor-tool` for the release of the app.

My only concern is that this fix might only affect new installations of Meteor.

For development: Use this command to run the tool as if it was a normal release and point/download the proper tool version:

```bash
function meteorw {
  METEOR_WAREHOUSE_DIR=/Users/<user>/.meteor METEOR_DEBUG_SPRINGBOARD=true meteord $@
}
```


leonardoventurini:
Now it's only failing on Windows 🤗

nachocodoner (approved):

PR #13201, opened 2024-06-21
Meteor 3 - Warn about using Meteor[.call/.apply] with async methods

denihs:
This PR originated from https://github.com/meteor/meteor/pull/13198.

What it does:

- Logs a warning message when invoking an async method with `Meteor.call` or `Meteor.apply`.
- Add a warning message to the docs for  `Meteor.call` and `Meteor.apply`, explaining that this function should be used only for sync methods.
- Revert the function `EVp.withValue` to its original state to avoid issues like the one described [here](https://github.com/meteor/meteor/pull/13198#issuecomment-2181254734).

### DIFF
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -614,7 +614,7 @@ export class Connection {
         stubOptions.stubReturnValue = DDP._CurrentMethodInvocation
           .withValue(invocation, stubInvocation);
         if (Meteor._isPromise(stubOptions.stubReturnValue)) {
-          console.warn(
+          Meteor._debug(
             `Method ${name}: Calling a method that has an async method stub with call/apply can lead to unexpected behaviors. Use callAsync/applyAsync instead.`
           );
         }
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -615,7 +615,7 @@ export class Connection {
           .withValue(invocation, stubInvocation);
         if (Meteor._isPromise(stubOptions.stubReturnValue)) {
           console.warn(
-            `Method ${name}: Calling an async method with call/apply can lead to unexpected behaviors. Use callAsync/applyAsync instead.`
+            `Method ${name}: Calling a method that has an async method stub with call/apply can lead to unexpected behaviors. Use callAsync/applyAsync instead.`
           );
         }
       } catch (e) {
--- a/packages/ddp-client/common/livedata_connection.js
+++ b/packages/ddp-client/common/livedata_connection.js
@@ -613,6 +613,11 @@ export class Connection {
       try {
         stubOptions.stubReturnValue = DDP._CurrentMethodInvocation
           .withValue(invocation, stubInvocation);
+        if (Meteor._isPromise(stubOptions.stubReturnValue)) {
+          console.warn(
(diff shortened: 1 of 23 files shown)


packages/ddp-client/common/livedata_connection.js
```diff
@@ -613,6 +613,11 @@ export class Connection {
       try {
         stubOptions.stubReturnValue = DDP._CurrentMethodInvocation
           .withValue(invocation, stubInvocation);
+        if (Meteor._isPromise(stubOptions.stubReturnValue)) {
+          console.warn(
+            `Method ${name}: Calling an async method with call/apply can lead to unexpected behaviors. Use callAsync/applyAsync instead.`
```
zodern, line 618:
```suggestion
            `Method ${name}: Calling a method that has an async method stub with call/apply can lead to unexpected behaviors. Use callAsync/applyAsync instead.`
```

packages/ddp-client/common/livedata_connection.js
```diff
@@ -613,6 +613,11 @@ export class Connection {
       try {
         stubOptions.stubReturnValue = DDP._CurrentMethodInvocation
           .withValue(invocation, stubInvocation);
+        if (Meteor._isPromise(stubOptions.stubReturnValue)) {
+          console.warn(
```
zodern, line 617: It might be good to use Meteor._debug for this. Many apps and some packages track messages from Meteor._debug as errors (including Monti APM) so developers would learn if it happens in production

v3-docs/docs/api/meteor.md
```diff
@@ -394,6 +394,11 @@ you want to process the method's result as soon as it arrives from the server,
 even if the method's writes are not available yet, you can specify an
 `onResultReceived` callback to [`Meteor.apply`](#Meteor-apply).
 
+::: warning
+Use `Meteor.call` only to invoke sync method. If you want to call invoke async methods, use `Meteor.callAsync`.
```
zodern, line 398:
```suggestion
Use `Meteor.call` only to call methods that do not have a stub, or have a sync stub. If you want to call methods with an async stub, `Meteor.callAsync` can be used with any method.
```

v3-docs/docs/api/meteor.md
```diff
@@ -404,6 +409,10 @@ even if the method's writes are not available yet, you can specify an
 passed as an array rather than directly as arguments, and you can specify
 options about how the client executes the method.
 
+::: warning
+Use `Meteor.apply` only to invoke sync method. If you want to call invoke async methods, use `Meteor.applyAsync`.
+:::
```
zodern, line 414:
```suggestion
::: warning
Use `Meteor.apply` only to call methods that do not have a stub, or have a sync stub. If you want to call methods with an async stub, `Meteor.applyAsync` can be used with any method.
:::
```

zodern (commented):
Thanks for fixing this. I left a few suggestions.

packages/meteor/dynamics_browser.js
```diff
@@ -40,25 +40,10 @@ EVp.withValue = function (value, func) {
   try {
     currentValues[this.slot] = value;
 
-    var ret = func();
```
leonardoventurini, line 43: Please don't forget to add a comment explaining why this should not behave the same as the server and how it interacts with methods and stubs. So another newbie doesn't make the same mistake 😄

leonardoventurini: Also, I think this breaks some tests I either added or modified.

leonardoventurini (approved):

PR #13202, opened 2024-06-22
allow $ in field name

minhna:
Mongodb now supports field names with dollar signs:
https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/#std-label-crud-concepts-dot-dollar-considerations

### DIFF
--- a/v3-docs/docs/.vitepress/theme/Layout.vue
+++ b/v3-docs/docs/.vitepress/theme/Layout.vue
@@ -3,7 +3,7 @@ import NotFound from './NotFound.vue'
 import GoToLatest from './GoToLatest.vue'
 import { useData, useRouter } from 'vitepress'
 import DefaultTheme from 'vitepress/theme'
-import { nextTick, provide, ref, onBeforeMount } from 'vue'
+import { nextTick, provide, ref, onMounted } from 'vue'
 import { redirect } from './redirects/script';
 const { isDark } = useData()
 const router = useRouter()
@@ -51,9 +51,9 @@ provide('toggle-appearance', async ({ clientX: x, clientY: y }: MouseEvent) => {
 })
 
 
-const isLatestVersion = ref(false);
+const isLatestVersion = ref(true);
 
-onBeforeMount(() => {
+onMounted(() => {
   isLatestVersion.value = isClient && window.location.href.startsWith("https://docs.meteor.com/")
 });
 
--- a/v3-docs/v3-migration-docs/index.md
+++ b/v3-docs/v3-migration-docs/index.md
@@ -1,7 +1,7 @@
 ---
-meteor_version: 3.0.4
-node_version: 20.18.0
-npm_version: 10.8.2
+meteor_version: 3.1.0
+node_version: 22.11.0
+npm_version: 10.9.0
 ---
 # Meteor 3.0 Migration Guide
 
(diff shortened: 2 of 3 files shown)


packages/minimongo/local_collection.js
```diff
@@ -1910,10 +1910,9 @@ const NO_CREATE_MODIFIERS = {
 };
 
 // Make sure field names do not contain Mongo restricted
-// characters ('.', '$', '\0').
+// characters ('.', '\0').
 // https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names
```
nachocodoner, line 1914: This PR looks good to me. However, I'm curious about the explicit prohibition of the dollar sign in minimongo. The documentation attached to the code mentions field name restrictions in Mongo, so it might just be an implementation based on those rules on the past, without other reasons behind.

nachocodoner: Thank you for the context provided. Now, I remember some use cases faced on the past with "$" and ".". It is ok to have these unlocked.

nachocodoner (approved):

zodern:
It seems like there are still quite a few limitations on `$` in field names: https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/dollar-prefix/#dollar-prefixed-field-names

For the tests you removed of cases minimongo did not allow, did you try them in Mongo to verify they should be allowed? 

It seems like for some cases minimongo would need basic support for the `$expr` operator.

minhna:
> minimongo would need basic support for the `$expr` operator.
I agreed that if you want to easily work with data in minimongo, you will need that.
But let's think about the most use cases we have: We have data on server, we use publication to sync it to minimongo. Now our server accept $ sign but minimongo doesn't. Your pub/sub error.
Now with this PR, you can have that data on your minimongo. You can work with that data by using normal javascript. It supports but doesn't mean you should use that. Minimongo is just a subset of mongodb features, it hasn't had all features and will never be.

nachocodoner:
> But let's think about the most use cases we have: We have data on server, we use publication to sync it to minimongo. Now our server accept $ sign but minimongo doesn't. Your pub/sub error.

Anyway, it would be good to have test coverage with the new case, to also verify the server side, and all this flow you describe. This will verify it is usable in this way, and we have both mongo driver on server and minimingo supporting the use case.

minhna:
> it would be good to have test coverage with the new case

It has, I added test to make sure we can insert object which has $ sign in key.
As @zodern said, minimongo needs to support $expr operator to query on field which has $ sign. So basically you can't put it in query to find, update but we can read.

nachocodoner:
But the test is only described on `[packages/minimongo/minimongo_tests_client.js](https://github.com/meteor/meteor/pull/13202/files#diff-6ad214f926e808466fc3aa1af77386e63757b54a58a6a04a9263ad54df5fa685)`, which is client-only. I refered to a more integration test for client and server code, for both sides test the new capabilities.

minhna:
> But the test is only described on [packages/minimongo/minimongo_tests_client.js](https://github.com/meteor/meteor/pull/13202/files#diff-6ad214f926e808466fc3aa1af77386e63757b54a58a6a04a9263ad54df5fa685), which is client-only. I refered to a more integrated test for client and server code, for both sides test the new capabilities.

This PR has nothing to do with server side. In order to support `$` sign in query, we must implement `$expr` operator support.

tcastelli:
@minhna We have found exactly the same issue (trying to subscribe to some data that contains $), what did you end doing to solve it? We were thinking to do a transofrmation in the subscription to return it as stringify of the object and then parse in the client again. But seems kinda double work

Issue #12940, opened 2023-12-31
POSIX dependency crashes meteor tool on Windows


buzzard192:
Installing or upgrading to the latest alpha (3.0-alpha.20) on Windows errors out due to POSIX attempting to be included in the webapp package.  

To fix the issue and successfully upgrade, remove the two files listed below from: 
     .meteor\packages\webapp\2.0.0-alpha300.20\npm\node_modules

- .npm-shrinkwrap.json
- .package-lock.json

also delete the POSIX module directory that was created during the installation attempt.

This issue is the same as the previous issue:
[POSIX dependency crashes meteor tool on Windows](https://github.com/meteor/meteor/issues/11497)

jbo023:
We do have the same problem for Meteor 3.0 beta4,
```
While loading package webapp@2.0.0-beta300.4:
error: Command failed: C:\WINDOWS\system32\cmd.exe /c
C:\Users\jbo02\AppData\Local\.meteor\packages\meteor-tool\3.0.0-beta.4\mt-os.windows.x86_64\dev_bundle\bin\npm.cmd
rebuild --update-binary
npm ERR! code 1
npm ERR! path
C:\Users\jbo02\AppData\Local\.meteor\packages\webapp\2.0.0-beta300.4\npm\node_modules\.temp-14ktj0g.milg\node_modules\posix
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c node-gyp rebuild
npm ERR! posix.cc
npm ERR! C:\Users\jbo02\AppData\Local\.meteor\packages\webapp\2.0.0-beta300.4\npm\node_modules\nan\nan.h(688,37):
warning C4996: 'v8::Isolate::IdleNotificationDeadline': Use MemoryPressureNotification() to influence the GC
schedule.
[C:\Users\jbo02\AppData\Local\.meteor\packages\webapp\2.0.0-beta300.4\npm\node_modules\.temp-14ktj0g.milg\node_modules\posix\build\posix.vcxproj]
npm ERR!   (Quelldatei "../src/posix.cc" wird kompiliert)
npm ERR!
npm ERR!
C:\Users\jbo02\AppData\Local\.meteor\packages\webapp\2.0.0-beta300.4\npm\node_modules\.temp-14ktj0g.milg\node_modules\posix\src\posix.cc(2,10):
error C1083: Datei (Include) kann nicht ge�ffnet werden: "unistd.h": No such file or directory
[...]

PR #13163, opened 2024-05-23
Release 3.0

denihs:
# Still in development
Current released version: ``3.0.1``
To use it, please do a 
```bash
meteor update --release 3.0.1
```
you can also use the following command to install meteor RC

```bash
npx meteor@rc
```
## Highlights
TODO
See more in the [milestones](https://github.com/meteor/meteor/milestone/101).

Preview docs: https://v3-docs.meteor.com/

## Maintainers checklist
- [x] address comments from old [PR](https://github.com/meteor/meteor/pull/12359)

⚠️⚠️⚠️ You can follow the updates on [What's left until an official Meteor 3.0? Discussion](https://github.com/meteor/meteor/discussions/12865)

_[old pr](https://github.com/meteor/meteor/pull/12359)_

PR #13204, opened 2024-06-24
OSS-459 Posix indirect dependency replacement

nachocodoner:
Context: https://github.com/meteor/meteor/issues/12940

This PR replaces the ['@vlasky/whomst'](https://github.com/vlasky/whomst) dependency with a different method to get the gid from group names provided by the UNIX_SOCKET_GROUP environment variable, introduced in [#11451](https://github.com/meteor/meteor/pull/11451).

The change aims to prevent a crash issue when updating Meteor apps to version 3.x. Some environments, including non-Windows ones, experience crashes during this update, requiring file resets. The `@vlasky/whomst` dependency with posix as an optional dependency is unnecessary for the case where used.

Alternative approaches would require deep knowledge and complex changes to how Meteor handles optional dependencies, as discussed [here](https://github.com/meteor/meteor/issues/12940#issuecomment-2013085379). For this use case, we will implement a simpler fix.

- [x] Provide test coverage for missing `UNIX_SOCKET_GROUP ` feature https://github.com/meteor/meteor/pull/11451
- [x] Replace the `'@vlasky/whomst'` with an approach to use `getent group` or as a fallback  `/etc/group` parsing to get gid from group names within UNIX system and Meteor socket configuration

`/etc/group` file is a standard Unix file on Unix-like systems such as Linux and macOS. Additionally, I added `getent` as a fallback approach, testing both methods in all my environments.

This code changed is specifically when using the UNIX_SOCKET_GROUP feature, ensuring minimal impact on Meteor in general. Testing and coverage was added to prevent any issue on using this feature. I hope the original problem during posix dep preparation is resolved on most OS now.


denihs (approved):

ruohki:
The "Council of Windows users" approved this merge.

wreiske:
For those of us coming from google search after using the latest release on https://v3-migration-docs.meteor.com/ (3.0-rc.4) , there is actually a newer release available (3.0-rc.8) that includes this PR fix! 

See the link below for the latest available release:

https://github.com/meteor/meteor/pull/13163#issue-2313382592

Issue #13212, opened 2024-07-03
[Meteor 3] critic regression: recalling Meteor.subscribe() with different params remove all documents and reinsert all documents


acemtp:
On the client:
```
Tracker.autorun(() => {
  const caseIds = _.pluck(Cases.find().fetch(), '_id');
  Meteor.subscribe('messages', caseIds);
});
```
On the server:
```
Meteor.publish('messages', function (caseIds) {
  return Messages.find({ caseId: { $in: caseIds } });
});
```

When Cases changes :

In 2.16, it only adds the new doc:

<img width="991" alt="image" src="https://github.com/meteor/meteor/assets/103561/d83b60cc-4d11-46b0-86b3-993816e73d14">

In 3.0-rc4, it removes all doc and inserts them again:

<img width="1592" alt="image" src="https://github.com/meteor/meteor/assets/103561/5d69e001-0895-4454-94b8-ae17680f30eb">

leonardoventurini:
Interesting, we will investigate

StorytellerCZ:
If this is the case then that would explain the strange behavior that I have noticed with my upgrade.

acemtp:
We created a very minimal project to easily produce it.

Meteor 2.16: It first sub and then unsub (so it doesn't remove/add because there's always a sub)

<img width="642" alt="image" src="https://github.com/meteor/meteor/assets/103561/8fcbe602-b932-4eef-8391-b915de597a3f">

Meteor 3: It first unsub then sub (so the unsub remove the data and the sub add the data back).

<img width="618" alt="image" src="https://github.com/meteor/meteor/assets/103561/6182a7d3-68d9-41b7-91f7-3f2ba4c4cba5">

StorytellerCZ:
@acemtp can you share with us the repository with the reproduction?

peernohell:
Hi,

I just create a minimal project to reproduce the issue.
It has a webpage where you have a refresh button.
if you click on it you will see that the tracker that look after post is never re-run but if you update the project in meteor v3.0-rc.4 then the auto run is executed

https://github.com/peernohell/meteor-bug-publish

what it show in v2.16 if you click refresh multiple time
<img width="519" alt="image" src="https://github.com/meteor/meteor/assets/119765/ce1b3e08-c9d3-403f-a758-92117c46dbee">


what it show in v3.0-rc.4 if you click refresh multiple time
[...]

PR #13220, opened 2024-07-09
[Meteor 3] Fix Meteor.subscribe() regression

denihs:
Issue reported on https://github.com/meteor/meteor/issues/13212.

As @radekmie [said](https://github.com/meteor/meteor/issues/13212#issuecomment-2212465898), the message order for subscriptions is inverted between Meteor 2 and 3.

In version 2, the order is: `sub` -> `unsub`. But in version 3, we have: `unsub` -> `sub`. Thus, the issue.

I narrowed it down to our new [queue system](https://github.com/meteor/meteor/blob/fd02bca8fcc857a2350b32d8c33e66fdb6288f3f/packages/ddp-client/client/queueStubsHelpers.js#L226-L240). Now in version 3 when we call Meteor.subscribe() we always [queue](https://github.com/meteor/meteor/blob/fd02bca8fcc857a2350b32d8c33e66fdb6288f3f/packages/ddp-client/client/queueStubsHelpers.js#L218-L223) the `Connection.send` method calls.

But we do not do this for the `subscription.[id].stop()`.

That's the issue. Even though we send the `sub` message before the `unsub`, the `unsub` message is sent before the `sub` message https://github.com/meteor/meteor/issues/13212#issuecomment-2212507115. 

To solve this, I've created the method `_sendQueued`, which will always queue the `Connection.send` call before sending the message. Now, when using this inside `subscription.[id].stop()`, we get the calls in the correct order:

Screenshot from 2024-07-09 08-53-23

(Descending Order)
image

Before merging this PR, we need to ensure some things:

- [x] This is the right strategy
- [x] All tests are passing
- [x] Create new tests for this method


StorytellerCZ:
As for right strategy, from my experience it is the correct one to prevent unnecessary re-renders on the client side if the subscription changes, but not the underlying data. It also saves the network traffic.

zodern:
Queueing the unsubscribe is the right approach.

radekmie (approved):

leonardoventurini (approved):

nachocodoner (approved):

JorgenVatle:
This might just be due to my application structure, but it appears that the changes introduced here can cause some inconsistent behavior with subscriptions. 

In Meteor v2 and the v3 release candidates the first DDP messages emitted are those from Meteor's core packages. Usually `connect` and `login` immediately following. After upgrading to v3.0.1 it seems user-defined subscription calls sometimes emit before the authentication calls. 

In my case this leads most subscriptions being rejected on first page load where they remain in a non-ready state until manually re-triggered after the login/resume calls have completed.

#### Meteor v3.0.1:
image


#### Meteor v3.0.0-rc.4 and earlier:
image

brianlukoff:
We're seeing the same thing, which is causing problems on reconnect -- essentially, upon reconnect, previously logged-in users get redirected away from their page because they are temporarily seen as being logged out (since the subscriptions before `login` fail). They are eventually logged in and a page refresh fixes the problem but not an ideal UX.

PR #13221, opened 2024-07-09
[Meteor 3] Fix Back from offline: method does not apply on server side

denihs:
This issue was reported in our [Forum](https://forums.meteor.com/t/back-from-offline-method-does-not-apply-on-server-side/61619/). [Here](https://github.com/harry-73/offline-meteor) is the reproduction.

The problem is the order in which some methods are being executed.

By adding logs [here](https://github.com/meteor/meteor/blob/1f92b242eac030fc40a85df8988fcee812746514/packages/ddp-server/livedata_server.js#L827) and running the [reproduction](https://github.com/harry-73/offline-meteor), you’ll be able to see that in Meteor 2 the order is:

```
1. { id: 4, method: “login”  }
2. { id: 3, method: “addLink”  }
```

But in Meteor 3 you get:

```
{ id: 3, method: “addLink”  }
{ id: 4, method: “login”  }
```

And this causes `this.userId` to be null inside `addLink`.

### Problem & Solution

The [login method](https://github.com/meteor/meteor/blob/4715e563a774ad7abb1c5d932fa4b08b98656384/packages/accounts-base/accounts_server.js#L649-L658) called when reconnecting is now async, and we didn't account for this before. So, even though we called the method at the right time, it wasn't solved fast enough.

Now, we're waiting for it to finish before invoking other methods.


StorytellerCZ:
In combination with #13220 it looks like there is a general bug in the way that DDP messages are send.

zodern:
This seems to be another bug with the queue. The method being async on the server doesn't affect the client.

When it reconnects, it calls the reconnect hook callbacks, and checks if they called any methods. The code checks for outstanding methods before the queue processes the login method.

I haven't looked at this code too closely, but I'm wondering if there are any issues with waiting for the login method to finish before processing the outstanding methods. It seems it would delay old outstanding methods to run after any new method calls made while logging in. It might be better to add an item to the queue and wait for it after running the reconnect hook callbacks.

packages/accounts-base/accounts_client.js
```diff
@@ -307,7 +307,7 @@ export class AccountsClient extends AccountsCommon {
           if (this._tokenExpiresSoon(result.tokenExpires)) {
             this.makeClientLoggedOut();
           } else {
-            this.callLoginMethod({
+            return this.callLoginMethod({
```
leonardoventurini, line 310: Why is this return necessary?

denihs: That's because this invocating is inside a `DDP.onReconnect`, which is a hook. So the idea here was to return this promise and wait for it [here](https://github.com/meteor/meteor/pull/13221/commits/82fc4e7aa4379001c4c3cd4282a4e65da48a02d7#diff-5a0f1c2953dc6856044b4803bb19c0b301386cd8836c878e36f2776d6316dc79R1798-R1799). But I think that won't be necessary anymore.

denihs:
@zodern, taking your idea, I've separated the `_outstandingMethodBlocks` invocations in a new method called `_sendOutstandingMethodBlocksMessages()`, and added this to the [queue](https://github.com/meteor/meteor/pull/13221/commits/fd7ce8f7e63836e424e38e6daa2c374562d85827#diff-2750275bb750c830025d71797568b7c4c2d5c45ffbc8d997befc14f1a8650e57R241-R251). This strategy work wonderfully!

I'm now waiting for the tests to run and want to create a new regression one. 

It's hard to predicate edge cases for this, but we can catch them as we go.

leonardoventurini (approved):

nachocodoner (approved):

PR #13236, opened 2024-07-18
Fix subscription still resetting documents

### LINKED ISSUE #13212: [Meteor 3] critic regression: recalling Meteor.subscribe() with different params remove all documents and reinsert all documents

leonardoventurini:
An image speaks more than a thousand words:

image

### DIFF
--- a/packages/ddp-client/test/livedata_connection_tests.js
+++ b/packages/ddp-client/test/livedata_connection_tests.js
@@ -2509,47 +2509,6 @@ if (Meteor.isClient) {
     test.equal((await coll.findOneAsync('aaa')).method, 222);
     test.equal((await coll.findOneAsync('aaa')).subscription, 112);
   });
-
-  Tinytest.addAsync(
-    "livedata connection - make sure the sub and unsub run in the correct order",
-    async function (test, onComplete) {
-      const stream = new StubStream();
-      // Make sure to disable this flag so the subscribe and unsubscribe are queued
-      stream._neverQueued = false;
-      const conn = newConnection(stream);
-
-      const sub = conn.subscribe("test_data");
-
-      // the subscribe message is still in the queue
-      test.isFalse(conn._readyToMigrate());
-      test.length(stream.sent, 0);
-
-      // unsubscribe
-      sub.stop();
-
-      // the queue still holds the data and no message arrived yet
-      test.isFalse(conn._readyToMigrate());
-      test.length(stream.sent, 0);
-
-      // waits until the queue is empty
-      await waitUntil(conn._readyToMigrate);
-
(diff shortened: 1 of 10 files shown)


denihs (approved):

packages/ddp-server/livedata_server.js
```diff
@@ -1265,13 +1263,15 @@ Object.assign(Subscription.prototype, {
...
 
-      this._publishCursorPromise = Promise.all(
-        res.map(c => c._publishCursor(self))
-      )
-        .then(() => {
-          self.ready();
-        })
-        .catch((e) => self.error(e));
+      try {
+        for (const cur of res) {
```
afrokick: why we need to change from parallel handling into sequenced?
`await Promise.all(res.map(cur => cur._publishCursor(self))` seems good.

leonardoventurini: Subscription parameter change needs be handled sync/waited for in the server, otherwise the client presents the behavior reported, being completely removed and then re-added. Not entirely sure how all this works, perhaps some can be handled in parallel but not others.

leonardoventurini: I will try your approach, I think it might still make sense.

leonardoventurini: Looks like it still works, thanks @afrokick !!

leonardoventurini:
I added a test for it, here is a breakdown

image

packages/ddp-client/test/livedata_connection_tests.js
StorytellerCZ, line 1: Can you explain why it is no longer needed?

leonardoventurini: It wasn't the problem in the first place, so for simplicity and risk management I thought best to remove it and the change made before (https://github.com/meteor/meteor/pull/13236/commits/f169bc81bbde74cc6ebf0fe95b2fcf035f6ba5ee). But if you, @zodern or @radekmie think I should keep it I can (already confirmed with @denihs).

leonardoventurini: Actually I noticed some flakyness after more testing, adding it back solves it, so it is truly part of the fix

PR #13247, opened 2024-07-23
Fix auth sub sending ready twice on load

leonardoventurini:
After we started queueing `_sendOutstandingMethodBlocksMessages` the `login` method started to be called after the `meteor.loginServiceConfiguration` sub call, thus recreating it once the user is set and sending two `ready` messages which flagged the auth process as ready before it actually was.

The only way we could solve it without causing a regression was by waiting until the Meteor client is connected to start the login service sub. Other subs will still be recreated but shouldn't cause any issues, until we figure a better approach.

image

- [x] Make sure the code is still needed and provide actual fix
- [x] Add regression test
- [x] Fix not being able to create account object due to parameters missing in validation array


packages/test-helpers/utils.js
```diff
@@ -0,0 +1 @@
+sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
```
nachocodoner: https://github.com/meteor/meteor/blob/549509fd9c0abfe6c1ca42c23e6c0a0d453ad52a/packages/meteor/async_helpers.js#L170-L173

There is already a helper used in the entire project and exposed publicly to sleep. But I like more the simplest sleep name for the helper.

packages/test-helpers/connection_server.js
```diff
@@ -52,3 +52,34 @@ makeTestConnection = function (test, succeeded, failed) {
...
+  function cleanup() {
+    conn._stream.send = send
+  }
+
+  return {
+    conn,
+    messages,
+    cleanup
+  }
+};
```
nachocodoner, line 85: Love to see the helpers expanded with these 🚀

nachocodoner (approved):

denihs (approved):

packages/test-helpers/utils.js
leonardoventurini: I've replaced all usages

PR #13193, opened 2024-06-12
Node 20.x - babel-preset-meteor (Dev bundle 20.14.0.0)

nachocodoner:

PR #13252, opened 2024-07-26
Node 20.17.0

StorytellerCZ:
Also updated the promise package: https://github.com/then/promise/releases

Other packages have been updated if by feature or patch version if not on 0.x version.

@nachocodoner @leonardoventurini @denihs started this bundle to get you started. Going forward I think it would be nice to dedicate one feature release to update all the bundle dependencies to their latest versions (or replace them like in  the case of sqlite when we reach Node 22 and can just use `node:sqlite`). This might include some small updates overall around, but we could then have more benefits from that. One hurdle that I saw were ESM packages which still need to be made available in Meteor (https://github.com/meteor/meteor/discussions/11727#discussioncomment-10158085)


StorytellerCZ:
Also updated dependencies for meteor-babel and babel-preset-meteor. Both of which will require independent release.
With babel-preset-meteor the question is if a patch release is fine or if it should become v7.25.0 to follow the style of Babel releases. Right now I opted for the latter.

nachocodoner:
I wouldn't mix node upgrade with related meteor-babel and babel-preset-meteor upgrades.

We still have pending to publish babel-preset-meteor latest upgrades to not use the proposals anymore and integrate in the dev bundle. That produced a lot kind of failures, https://github.com/meteor/meteor/pull/13193.

So, upgrading Node can be handled independently and move forward as is an easy upgrade with less friction, while other libs need extra efforts, we can cover them by its own for further verification.

StorytellerCZ:
OK, will split.

StorytellerCZ:
@nachocodoner reverted to just bundle updates.

nachocodoner:
@StorytellerCZ: I bump npm version as well as the new Node 20.16 come along with an upgrade on npm@10.8.1. 

But I still see on your changes updates outside Node 20.16 scope. I will adjust them if I find troubles with the new dev_bundle.

nachocodoner:
First issues:

image

and meteor babel upgrade breaks locally

scripts/dev-bundle-tool-package.js
```diff
@@ -41,30 +41,30 @@ var packageJson = {
...
     inquirer: "8.2.6",
     "http-proxy": "1.18.1",
     "is-reachable": "3.1.0",
     "wordwrap": "1.0.0",
-    "moment": "2.29.1",
+    "moment": "2.30.1",
     "rimraf": "2.6.2",
     "glob": "7.1.6",
-    ignore: "3.3.7",
+    ignore: "5.3.2",
```
nachocodoner, line 52: This ignore tool dependency is used for .meteorignore file formatting, there are extra features for latest version, and according to my research doesn't produce any breaking change. But I will do further testing.

StorytellerCZ:
We should resolve the issue with Meteor Babel by publishing a proper version of that package and then we can add it to the bundle.
#13273

nachocodoner:
20.17.0 is out. I could patch it and include for next release.
https://nodejs.org/en/blog/release/v20.17.0

StorytellerCZ:
Yes, unless there is some blocker I see no reason to not do it.

.travis.yml
```diff
@@ -4,7 +4,7 @@ dist: jammy
 sudo: required
 services: xvfb
 node_js:
-  - "20.16.0"
+  - "20.15.1"
```
nachocodoner, line 7: Travis CI struggles with bumping the Node version, often getting stuck during environment initialization. Clearing Travis CI caches didn't help.

Bumping the Node version isn't required since Meteor's built-in Node still supports the necessary Node version for testing. Also, we plan to eventually migrate all pipelines to GitHub Actions, which will ensure compatibility. I tested version 20.17.0 locally, and everything worked fine, with tests passing.

StorytellerCZ: Bummer. The sooner the migration happens the better.

PR #13254, opened 2024-07-26
Use PowerShell to set PATH instead of setx on Windows

### LINKED ISSUE #13253: Install mangles user's PATH environment variable on Windows
Meteor 3.0.1

The Meteor install mangles the user `PATH` environment variable because it [uses `setx`](https://github.com/meteor/meteor/blob/549509fd9c0abfe6c1ca42c23e6c0a0d453ad52a/npm-packages/meteor-installer/install.js#L305). This method has several inherent flaws.

The first issue is that `setx path "${meteorPath}/;%path%` copies the entire system `PATH` env into the user's `PATH` env, often duplicating entries in the process.

The second issue is that `setx` will expand all environment variables that exist in either `PATH`. For example, if either `PATH` contains `%USERPROFILE%\my-app`, that will get expanded to something like `C:\Users\Jane\my-app` and rewritten into the user's `PATH`, losing the reference to the environment variable.

The third and probably worst issue is that `setx` has a limit of 1024 characters. So, if the combination of the system `PATH` and the user's `PATH` exceeds that limit, which it often does due to the two issues above, the user's `PATH` env will be truncated, often breaking the user's environment.

On Windows, users will see the following message when that happens:

```
WARNING: The data being saved is truncated to 1024 characters.
```

Instead of using `setx`, the Meteor install should use PowerShell with the commands like the following (untested):

```
$path = [Environment]::GetEnvironmentVariable("PATH", "User");
[Environment]::SetEnvironmentVariable("PATH", "${meteorPath}/;$path", "User");
```

This will avoid copying the system `PATH` into the user's `PATH`, avoid expanding environment variables, and avoid the `setx` 1024 character limit.

jordanbtucker:
This PR makes the Meteor installer use PowerShell instead of `setx` to add Meteor to the user's `PATH` environment variable on Windows.

This resolves the issues described in #13253:

1. The Meteor install no longer mangles the user's `PATH` environment variable by copying the system `PATH` environment variable into it.
2. The Meteor install no longer expands referenced environment variables.
3. The Meteor install no longer truncates the user's `PATH` environment variable to 1024 characters.

Fixes #13253

### DIFF
--- a/npm-packages/meteor-installer/install.js
+++ b/npm-packages/meteor-installer/install.js
@@ -302,7 +302,9 @@ async function setup() {
 async function setupExecPath() {
   if (isWindows()) {
     // set for the current session and beyond
-    child_process.execSync(`powershell -c "$path = (Get-Item 'HKCU:\\Environment').GetValue('Path', '', 'DoNotExpandEnvironmentNames'); [Environment]::SetEnvironmentVariable('PATH', \\"${meteorPath};$path\\", 'User');"`);
+    child_process.execSync(
+      `powershell -c "$path = (Get-Item 'HKCU:\\Environment').GetValue('Path', '', 'DoNotExpandEnvironmentNames'); [Environment]::SetEnvironmentVariable('PATH', \\"${meteorPath};$path\\", 'User');"`
+    );
     return;
   }
   const exportCommand = `export PATH=${meteorPath}:$PATH`;
--- a/npm-packages/meteor-installer/install.js
+++ b/npm-packages/meteor-installer/install.js
@@ -302,7 +302,7 @@ async function setup() {
 async function setupExecPath() {
   if (isWindows()) {
     // set for the current session and beyond
-    child_process.execSync(`setx path "${meteorPath}/;%path%`);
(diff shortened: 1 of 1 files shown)


jordanbtucker:
Okay, I think I got all of the style checks to pass now. Is there anything I need to do to get the [Docs check](https://app.circleci.com/pipelines/github/meteor/meteor/8118/workflows/7af1a1bd-135b-4b21-b5c2-bfa564ff2090/jobs/139096) to pass?

nachocodoner:
The docs check is an existing issue that happens on PR from forked meteor branches. We need to attend a fix on this, but for the moment, you can completely skip this error.

I will review myself your issue on my Windows machine and be back on approve and merge your fix. Thank you for your contribution!

nachocodoner (approved):
Thank you very much for this contribution.

I just tested your change and it worked well.

So I removed Meteor completely from my Windows environment, uncluding the Windows env PATH. Then I reinstalled using your change and expecting proper installation with PATH updated. It happened.

image

PR #13262, opened 2024-07-30
Fix outstanding subscription init order generally

### LINKED ISSUE #13241: Meteor.userAsync() null if called server-side from publish/subscribe as first call
Hello,

I've try to migrate my project from Meteor 2.16 to Meteor 3.0.1 last weekend and I found a problem with publish/subscribe on my project.
I didn't find any trace in the issues, that's why I'm posting here, sorry if it's duplicated.

So... After many test, I found this:
When I call a publish/subscribe first after a client refresh, the publish function call the Meteor.userAsync() which returns a null value.
However, if I call a method first, the publish/subscribe called next after the client refresh will find the Meteor.userAsync() correctly.

**Version of Meteor:** 3.0.1 after migration from 2.16

**Operating system:** MacOS

**The expected behavior:**
- CLIENT: The client previously connected (using google oauth package in my case) refresh the page
- CLIENT: The subscribe is called to get some informations
- SERVER: The publish check the user connected
- SERVER: User information is correctly obtained

**The actual behavior:**
- CLIENT: The client previously connected (using google oauth package in my case) refresh the page
- CLIENT: The subscribe is called to get some informations
- SERVER: The publish check the user connected
- SERVER: The user is null

**Reproduction:**
[https://github.com/Argasar-O/Meteor-Issue](https://github.com/Argasar-O/Meteor-Issue)

**Steps:**
1. Run the app (settings are in the README, need to have google api keys)
2. Click on the login button and connect the user
3. Refresh the client (F5) -> An error occurred server-side because the user is null.
4. Uncomment in `/client/main.tsx` line 8
5. Do the steps 1 to 3 -> The user is corretly found server-side, no error

Have a good day.

leonardoventurini:
Now the methods are called before the subscription init calls by adding them to the queue as well.

### DIFF
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -18,6 +18,9 @@ export class AccountsClient extends AccountsCommon {
     this._loggingIn = new ReactiveVar(false);
     this._loggingOut = new ReactiveVar(false);
 
+    this._loginServicesHandle =
+      this.connection.subscribe("meteor.loginServiceConfiguration");
+
     this._pageLoadLoginCallbacks = [];
     this._pageLoadLoginAttemptInfo = null;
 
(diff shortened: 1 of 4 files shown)


StorytellerCZ (approved):

alisnic:
is there a chance to include this in 3.0.2 as well?

StorytellerCZ:
I think that is the plan @alisnic

leonardoventurini:
@alisnic yes, that's the plan, still need to expand this to cover one other edge case and we will be ready to merge it

leonardoventurini:
After some validation and discussion with interested parties, we found that this PR does solve the issues at hand. Merging it now, and going for another beta as soon as we can.

PR #13280, opened 2024-08-06
Adjust Meteor Installer Readme

leonardoventurini:
Inform that we use Node.js v20 for Meteor 3, simplify wording, and adjust order.

### DIFF
--- a/npm-packages/meteor-installer/README.md
+++ b/npm-packages/meteor-installer/README.md
@@ -48,7 +48,7 @@ npm install -g meteor --ignore-meteor-setup-exec-path
 
 Set the `https_proxy` or `HTTPS_PROXY` environment variable to a valid proxy URL to download Meteor files through the configured proxy.
 
-### Meteor version relationship
+### Meteor Version Compatibility
 
 | NPM Package | Meteor Official Release |
 |-------------|-------------------------|
--- a/npm-packages/meteor-installer/README.md
+++ b/npm-packages/meteor-installer/README.md
@@ -1,14 +1,52 @@
 ## Meteor Installer
 
-Node.js <=14.x and npm <=6.x is recommended.
+### Recommended Versions
 
-Install Meteor by running:
+- For Meteor 2
+  - Use Node.js 14.x or higher
+  - Use npm 6.x or higher
+- For Meteor 3
+  - Use Node.js 20.x or higher
+  - Use npm 9.x or higher
+
+### Installation
+
+To install Meteor, run the following command:
 
 ```bash
 npm install -g meteor
 ```
 
-[Read more](https://www.meteor.com/developers/install)
+#### Running with `npx`:
+
+```bash
+npx meteor
+
+# or with a specific version
+npx meteor@<version>
+```
+
+This command will execute the Meteor installer without adding it permanently to your global npm packages.
+
+For more information, visit the [official Meteor installation guide](https://www.meteor.com/developers/install).
+
+### Important Note
+
+This npm package is not the Meteor framework itself; it is just an installer. Do not include it as a dependency in your project, as doing so may break your deployment.
+
+### Path Management
+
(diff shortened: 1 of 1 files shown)


npm-packages/meteor-installer/README.md
```diff
@@ -1,16 +1,54 @@
...
 
-[Read more](https://www.meteor.com/developers/install)
+#### Running with `npx`:
+
+```bash
+npx meteor
+
+# or with a specific version
+npx meteor@<version>
+```
```
nachocodoner, line 24: I think this is the preferred way to install Meteor, since `npm install -g meteor` doesn't provide any feedback logs unless you do: `npm install -g meteor --foreground-script` . May we just keep then npx option here, or at least switch the order?

npm-packages/meteor-installer/README.md
```diff
@@ -1,16 +1,54 @@
 ## Meteor Installer
 
-Node.js <=14.x and npm <=6.x is recommended.
+### Recommended Versions
 
-Install Meteor by running:
+- For Meteor 2
+  - Use Node.js 14.x or higher
+  - Use npm 6.x or higher
```
fredmaiaarantes:
```suggestion
- For Meteor 2
  - Use Node.js 14.x
  - Use npm 6.x
```

For Meteor 2, Node.js and npm should not be higher, right? We  should recommend using Node.js 14 and npm 6 only.

npm-packages/meteor-installer/README.md
```diff
@@ -1,16 +1,54 @@
...
+- For Meteor 3
+  - Use Node.js 20.x or higher
+  - Use npm 9.x or higher
```
fredmaiaarantes, line 10:
```suggestion
- For Meteor 3
  - Use Node.js 20.x
  - Use npm 9.x
```

Has Meteor 3.0 been tested with newer versions of Node.js and npm? If not, we should recommend specific versions.

npm-packages/meteor-installer/README.md
fredmaiaarantes: I think it would be better to keep only one option.

npm-packages/meteor-installer/README.md
```diff
@@ -1,16 +1,54 @@
...
+```bash
+npx meteor
+
+# or with a specific version
+npx meteor@<version>
+```
+
+This command will execute the Meteor installer without adding it permanently to your global npm packages.
+
+For more information, visit the [official Meteor installation guide](https://www.meteor.com/developers/install).
```
fredmaiaarantes: This link goes to the Meteor 2 installer. Maybe we should add both links here?

https://v2-docs.meteor.com/install.html

https://v3-docs.meteor.com/about/install.html

npm-packages/meteor-installer/README.md
leonardoventurini: I agree, I will invert the order on those then

npm-packages/meteor-installer/README.md
leonardoventurini: Oops, in this case it should be lesser or equal

npm-packages/meteor-installer/README.md
leonardoventurini: This only applies to the installer, so v20 would be the minimum for v3, soon we will be updating for v22 but that would be bundled into the framework.

npm-packages/meteor-installer/README.md
leonardoventurini: Yes, that makes sense. Thanks!

npm-packages/meteor-installer/README.md
leonardoventurini: I didn't see your comment there @fredmaiaarantes , well I will go with the simplest one then `npx`.

fredmaiaarantes (approved):

PR #13284, opened 2024-08-07
Fix Meteor.promisify and add a test for it

OleksandrChekhovskyi:
This fixes a regression introduced in commit 87f8697941.

That commit switched from arrow functions to regular functions, which caused `this` and `arguments` to change. The fix is to capture `this` and `arguments` in the correct context.

Adding a simple test as well, to watch for regressions here.

### DIFF
--- a/packages/meteor/helpers.js
+++ b/packages/meteor/helpers.js
@@ -85,6 +85,10 @@ Meteor.promisify = function (fn, context, errorFirst) {
   }
 
   return function () {
+    var self = this;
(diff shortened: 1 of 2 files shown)


nachocodoner (approved):
Good catch! Thank you for the contribution.

I see the same code living on Meteor 2.x.x branch, so I will cherry pick this commit there as soon as I see test passing.

leonardoventurini (approved):

PR #13289, opened 2024-08-08
Fix deprecation dependency on installer

nachocodoner:
image

`rimraf` - updates to latest, used on uninstall behavior. No breaking change migration needed as simple usage.
`glob` - modern version comes with rimraf
`inflight` - removed from glob on modern versions

Tested the uninstall behavior after upgrade.

image


leonardoventurini (approved):
Nice!

PR #13296, opened 2024-08-12
add ddpUrl to VALID_CONFIG_KEYS

jhernandez-etraid:


packages/accounts-base/accounts_common.js
```diff
@@ -19,6 +19,7 @@ const VALID_CONFIG_KEYS = [
   'loginTokenExpirationHours',
   'tokenSequenceLength',
   'clientStorage',
+  "ddpUrl"
```
StorytellerCZ: Please follow the format of using `'` instead of `"` as above.

StorytellerCZ (changes requested):

packages/accounts-base/accounts_common.js
jhernandez-etraid: You are right

nachocodoner (dismissed):

nachocodoner:
This was added as well as connection field for 3.0.2. Merged anyway. Thank you for the contribution. 🚀

PR #13298, opened 2024-08-13
Fix TLA still not working

### LINKED ISSUE #13299: TLA detection still not working (#12986 closed without being fixed)
Metor 3.0.1

#12986 was closed as fixed but in reality this was never the case.

Reify was fixed but the updated version was not brought in.

#13298 attempts to fix this.

All repro steps in #12986 still apply, e.g. https://github.com/perbergland/tla-repro/tree/main or https://github.com/alisnic/meteor3-tla-repro

leonardoventurini:
Fix Top Level Await still not working correctly in Meteor core, basically making sure the patched version of Reify is used in all places in the core. Bear in mind that TLA is disabled by default in the client.


denihs (approved):

leonardoventurini:
It is passing after I regenerated the bundle:

image

perbergland:
Lgtm but there is a need for docs on TLA in meteor 3, if it means that non-top-level requires may become async

leonardoventurini:
@perbergland, there already are docs for it:

https://docs.meteor.com/api/top-level-await.html

I don't think non-TLA modules will become async; if they do, it's a bug

Do you see anything that hints at that?

leonardoventurini:
By the way, the only test failing is due to an issue in CI we are already working on

leonardoventurini:
Looks like `assert` is triggering promisification, need to understand better why

leonardoventurini:
Holding this to the next release until the strange behavior is figured out

PR #13300, opened 2024-08-14
Using await when open tag is called #13175

### LINKED ISSUE #13175: Meteor 3 --open opens the browser before the app is started
As the title says, this results in a blank page for a bit which can be confusing, so if possible the open command should trigger once the app is running.

kbarr1212:
Issue #13175

https://github.com/meteor/meteor/issues/13175 

We added an if statement in run-all.js that gives the option whether to await runner.start() if runOptions.open is true or it runs the existing setTimeout() by passing the open parameter in commands.js.

Screenshot 2024-08-14 at 11 24 46 AM

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -516,7 +516,8 @@ async function doRunCommand(options) {
           open(`http://localhost:${options.port}`)
         }
       }
-    }
+    }, 
+    open: options.open,
   });
 }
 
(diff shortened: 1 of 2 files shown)


nachocodoner (approved):
This looks good for me, as discussed on the issue. https://github.com/meteor/meteor/issues/13175

Thank you for your support with this contribution, @AvinRai, @rhumtea and @kbarr1212 🚀

PR #13277, opened 2024-08-03
fix #13263 set ambiguousErrorMessages consistently across environments

mauroaccornero:
remove Meteor.isProduction from isErrorAmbiguous check in accounts_server.js

fixes: #13263

PR #13301, opened 2024-08-17
fix #13263 set ambiguousErrorMessages consistently across environments

### LINKED ISSUE #13263: Set ambiguousErrorMessages consistently across environments
See https://github.com/meteor/meteor/pull/12996

Whenever developers test and implement error handling locally, they expect that code will behave in the same way on production. Otherwise, a lot of time is wasted trying to understand different behaviour and failing to reproduce it locally.

cc @StorytellerCZ

mauroaccornero:
remove Meteor.isProduction from isErrorAmbiguous check in accounts_server.js

fixes: #13263 
replace: #13277


mauroaccornero:
duplicate of #13277

PR #13320, opened 2024-08-28
Performance Optimizations

leonardoventurini:
- Disable compression in Faye/SockJS

> Significantly impacts performance, both in RAM and CPU, whatever gains in bandwidth are negated by completely overloading some reactivity dense apps.

- Remove `bindEnvironment` from tasks in AsynchronousQueue

> Not as impactful on performance, but still has some impact, we need to reduce our ALS footprint where possible. As I suspect `permessage-deflate` only became a problem now because we are already overusing async resources/context.

***Will publish numbers after validation***

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ packages/**/.npm
 
 # doc files should not be committed
 packages/**/*.docs.js
+jsconfig.json
--- a/packages/accounts-2fa/.npm/package/npm-shrinkwrap.json
+++ b/packages/accounts-2fa/.npm/package/npm-shrinkwrap.json
@@ -2,9 +2,9 @@
   "lockfileVersion": 4,
   "dependencies": {
     "@types/node": {
-      "version": "22.3.0",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-22.3.0.tgz",
-      "integrity": "sha512-nrWpWVaDZuaVc5X84xJ0vNrLvomM205oQyLsRt7OHNZbSHslcWsvgFR7O7hire2ZonjLrWBbedmotmIlJDVd6g=="
+      "version": "22.5.1",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-22.5.1.tgz",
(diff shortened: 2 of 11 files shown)


radekmie:
I have a few questions to consider (not blockers in any way):
1. How did you measure the performance improvements? It's a good opportunity to prepare some benchmark suite for publications in general.
2. Why did `permessage-deflate` get disabled? It uses the native `zlib` under the hood, and I doubt it causes any visible performance impact in real-life usage.
    * We could tweak the configuration instead to make it even less CPU-intensive.
    * Also, it helps with long-ish arrays of similarly structured documents returned from methods. And it'd get even better with the batching of `added` DDP messages (like https://github.com/meteor/meteor/pull/10478).
3. Removal of `bindEnvironment` is a good idea as it reduces the number of intermediate `Promise`s, but it also loses some context within error handling. Did you check if those didn't get too cryptic?

leonardoventurini:
Thanks @radekmie !

1. We built an artillery suite, an event loop monitor which triggers async resource logging during event loop lag (100ms+), that's how we pinned down `permessage-deflate`. We will be publishing the results in a few days, just need to do some additional validation and make sure the numbers are correct.
2. `permessage-deflate` significantly impacts performance, both in RAM and CPU, whatever gains in bandwidth are negated by completely overloading some reactivity dense apps. Will try tweaking the levels too and see how it behaves.
3. I will check the errors, I suspect it was binding the environment twice for the things inside it anyways.

leonardoventurini:
From what I see we need to revamp the transport layer completely, perhaps use a native WebSocket server, like `ws` or even `uWebSockets`. Perhaps they have their own compression mechanism/plugins that are more performance too.

radekmie:
1\. and 2. I'd make it opt-out, then. Both because it's actually backward incompatible change and also because maybe it helped in this case but will impact other cases negatively. I'd love to see this suite, as I guess it's missing some cases where the compression would actually help (and some, where it wouldn't be an issue).
3\. Binding is one thing. Adding context (also in the stack trace) is another. Just double-check it.

> From what I see we need to revamp the transport layer completely, perhaps use a native WebSocket server, like `ws` or even `uWebSockets`.

That's what `DISABLE_SOCKJS` does, at least partially. I personally doubt that the transport is anyhow slowing the server as I never saw it anywhere high in the profiles.

leonardoventurini:
Closing this PR since we decided to keep it optional and `SERVER_WEBSOCKET_COMPRESSION=false` disables it.

PR #13323, opened 2024-09-02
Use more Express and remove body-parser dependency

StorytellerCZ:
While debugging some issues I have with my OAuth I made some improvements to the OAuth server part to utilize more of Express and remove the dependency on `body-parser` ~~and removed `oauthServiceName` function as we could get all of that via url params~~.


StorytellerCZ:
The test failures in Travis are related, but I don't think the failures are valid and instead the tests might need updating. Will explore once I have more time. Not sure what to make of the failed CircleCI test.

denihs:
Once you have an update about the tests we can merge it!

StorytellerCZ:
Initial look on the failing tests I still don't see why they are coming from WebApp, so I need to dive in further.

StorytellerCZ:
Fixed the issue in Travis and now getting:
```
node:internal/process/promises:391
    triggerUncaughtException(err, true /* fromPromise */);
    ^
Error [ReferenceError]: __Tinytest is not defined
    at evaluate (evaluate at /home/travis/build/meteor/meteor/packages/test-in-console/puppeteerRunner.js:15:19, <anonymous>:0:19)
    at ExecutionContext._ExecutionContext_evaluate (/home/travis/build/meteor/meteor/dev_bundle/lib/node_modules/puppeteer/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:262:51)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ExecutionContext.evaluate (/home/travis/build/meteor/meteor/dev_bundle/lib/node_modules/puppeteer/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:145:16)
    at async /home/travis/build/meteor/meteor/packages/test-in-console/puppeteerRunner.js:15:8
```

denihs:
Some tests are failing; could you have a look at them? 
image

StorytellerCZ:
Yes, it is on my list.

StorytellerCZ:
I think the tests are going to pass now. Will have to figure out later why the `params` were not detected in tests. Right now I have it so that if there is param with the service name it will get it first and the expensive old function to detect if from the string is there as a failback. I think the issue might be related to tests specifically, but right now I don't have the time to dig more deeply.

nachocodoner:
> While debugging some issues I have with my OAuth

@StorytellerCZ: How is this related with oauth fixes https://github.com/meteor/meteor/pull/13515#issuecomment-2552730925?

Just to clarify, does this PR include refactors, cleaning, and updates to modern tools like Express, or does it also address any fixes? Could you explain the OAuth server issue that is fixed in this PR? I want to replicate and verify it if any.

StorytellerCZ:
Over the time this has become more of a refactor to use Express more and update stuff. The original problem was that sometimes the callback passed through and went to the client without hitting the endpoints on the server.

StorytellerCZ:
Upgraded `compression` package which adds support for brotli compression (looking at the documentation I'm not sure if we have to do anything else to have it supported). Going to take this out for testing right now as I have some issues with OAuth to see if I can dig down to the main issue.

We should probably consider some settings to allow setting things like compression and potentially testing what is best for Meteor apps (if we can get better than the current defaults).
Docs on compression: https://github.com/expressjs/compression?tab=readme-ov-file#compression

StorytellerCZ:
Still having issues with OAuth in Brave browser (started this primarily as I wanted to fix it). For some reason after redirect back to the server it fails to hit the server correctly:

https://github.com/user-attachments/assets/4bd3d94c-b6dd-426a-a4da-e021c3ecc7bb

When doing the test with Edge (the left window) it works fine. Not sure what exactly is the issue in Brave (the test was with Shield down), but probably some security setting. Brave has a lot more turned on, but we need to find exactly what it is as it is likely that more browsers will follow or users might turn it on. https://privacytests.org/ has the list of security settings comparison.
I think we should also look into modernizing the popup communication back to the app, but that is for another issue.
I will now look on the failing tests. Hopefully figuring out why those fail might provide a breakthrough.

Grubba27 (commented):
It seems ok – I would feel more comfortable if @nachocodoner takes a look at it too

PR #13325, opened 2024-09-02
Fix wrong URL to changelog

filipenevola:
The URL is resulting in 404.

Fixing to which appears to be the new one.


denihs:
Thanks!

PR #13354, opened 2024-09-16
Generators: Add options to set where to import generated index

StorytellerCZ:
Adding two new options to either prevent import addition on the server, or set where to import instead. This adds more flexibility for own generator templates.

Worked on during my stream: https://www.youtube.com/watch?v=Z6nY2wqPgSE
See second hour for full context and work.

#12298


tools/cli/commands.js
```diff
@@ -3107,21 +3123,29 @@ main.registerCommand({
...
-      ? files.pathJoin(appDir, packageJson.meteor.mainModule.server)
-      : files.pathJoin(appDir, 'server', 'main.js');
-  const mainJs = files.readFile(mainJsPath);
-  const mainJsLines = mainJs.toString().split('\n');
-  const importLine = path
-    ? `import '${path}';`
-    : `import '/imports/api/${ scaffoldName }';`
-  const mainJsFile = [importLine, ...mainJsLines].join('\n');
-  files.writeFile(mainJsPath, mainJsFile);
+  if (importIntoServerMain || serverImportFile?.length > 0) {
```
denihs, line 3126: What happens if `importIntoServerMain` and `serverImportFile?.length > 0` are `false`?

Could you create some test scenarios for these?

denihs (changes requested):

tools/cli/commands.js
StorytellerCZ: If both are false then no import is added anywhere which is a relevant use case.

I haven't found any tests for the generators. But I haven't looked too much.

denihs: Could you confirm that? Then, we can include this in the next patch.

StorytellerCZ: Will do, but the nature of this change I think warrants this to be in v3.1 rather than a patch release.

italojs:
I just missed a few tests for this new feature

PR #13364, opened 2024-09-20
Update type definitions for accounts-base for functions that became async

9Morello:
This is a minor PR that updates some function signatures in the `accounts-base.d.ts` file to reflect the fact that they became asynchronous. `addEmail` was renamed to `addEmailAsync`, as per the implementation.

### DIFF
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -16,7 +16,7 @@ const util = require("util");
 var MongoDB = NpmModuleMongodb;
 import { DocFetcher } from "./doc_fetcher.js";
 import {
-  ASYNC_CURSOR_METHODS,
+  CURSOR_METHODS,
   CLIENT_ONLY_METHODS,
   getAsyncMethodName
 } from "meteor/minimongo/constants";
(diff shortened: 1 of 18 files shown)


packages/mongo/mongo_driver.js
```diff
@@ -909,30 +888,29 @@ Cursor.prototype.count = function () {
...
-  }
-
-  const methodNameAsync = getAsyncMethodName(methodName);
-  Cursor.prototype[methodNameAsync] = function (...args) {
-    try {
-      return Promise.resolve(this[methodName](...args));
-    } catch (error) {
-      return Promise.reject(error);
+// We don't handle the count method here.
+[...CURSOR_METHODS, Symbol.asyncIterator].filter(method => method !== CURSOR_METHODS[0]).forEach(function(method) {
```
denihs, line 892:
```
method !== CURSOR_METHODS[0]
```

If, in the future, we forget about these rules and add something at the top of this array, this code won't work as expected. 

It's best to use the `if` expression here.

denihs (changes requested):

denihs:
There are some tests breaking. Could you check those?

9Morello:
@denihs I made another PR which doesn't include unrelated changes https://github.com/meteor/meteor/pull/13374

Issue #13335, opened 2024-09-09
`Meteor.loggingIn()` backwards compatibility 


harryadel:
`Meteor.loggingIn` doesn't behave in 3.x as it did in 2.x. 

Fetching `Meteor.loggingIn()` yields `false` whilst user `profile` property is undefined (aka [user subscription](https://github.com/meteor/meteor/blob/fa45160d9cab879dd370ffe91b24fa2491d4e10b/packages/accounts-base/accounts_server.js#L790) hasn't completed yet).

This can lead to issues in code expecting to be able to work with the profile data once Meteor.loggingIn() is false.

@DanielDornhardt came up with a crafty solution to patch this problem but it's a band aid so we wanted to raise this issue and garner feedback from others who might be experiencing this problem. 

```js


// Wait until things have settled down, then lets try to override the Meteor.loggingIn() function with one
// which waits for the profile field for the user to been published as well.
//
// This seems to have been the regular functionality before Meteor 3
Meteor.startup(() => {

    const _ourUserHasAProfile = new ReactiveVar()

    Tracker.autorun(async () => {
        _ourUserHasAProfile.set((await Meteor.userAsync())?.profile)
    })

    const _ourLogginIn = new ReactiveVar(Meteor.loggingIn())

    const originalLoggingIn = Meteor.loggingIn

    Tracker.autorun(async () => {
        if (originalLoggingIn()) {
            console.log('originalLoggingIn: true')
            _ourLogginIn.set(true)
            return
        }
        if (!originalLoggingIn()) {
            console.log('originalLoggingIn: false', `_ourUserHasAProfile ? ${_ourUserHasAProfile.get()}`)
            // we make sure that the user profile has been published too before we turn loggingIn to false

            if (_ourUserHasAProfile.get()) {
                console.log('setting loggingIn to false')
                _ourLogginIn.set(false)
            }
        }
    })

    // replace with monkeypatched version
    Meteor.loggingIn = () => {
        console.log('_ourLogginIn', _ourLogginIn.get())
        return _ourLogginIn.get()
    }
})

```

denihs:
As the solution to this is merged, I'll close this issue.

PR #13369, opened 2024-09-23
Fix _setLoggingIn(false) sometimes being called before Meteor.userAsync could return an user

9Morello:
This PR fixes a behavior on `accounts-base` which could cause `loggingIn` to turn into `false` before `Meteor.userAsync()` was ready toreturn an user, causing a React App to 'flicker' for a split second before showing the user the correct screen.

Here's a terminal output of this snippet in a React app before and after the changes.
```js
const user = useTracker(() => Meteor.user());
const loggingIn = useLoggingIn();
console.log({ user, loggingIn });
```

### Before:
fix_before

### After:
fix_after

The only change is that we only call `this._setLoggingIn(false)` either when we received and error, or when `Meteor.userAsync` finally returns an user, avoiding a situation where the `useLoggingIn` hook would not return the right result.

[This issue](https://github.com/meteor/meteor/issues/13335) _may_ be related.


harryadel:
Thank you for undertaking this, hopefully this can be reviewed and merged soon.

denihs (approved):
Nice, this will be included in the next release!

denihs:
@9Morello, before we merge it, could you update your branch with the changes from `release-3.0.4`?

9Morello:
@denihs sure, give me a moment

9Morello:
@denihs done

denihs:
Nice thanks!

PR #13373, opened 2024-09-25
fix: stop computation after callLoginMethod

ggazzo:
I'm almost certain that there wasn't a tracker here before. It was created to ensure that the user is already in the collection before calling the callback, but there's a chance it could be called more than once


denihs:
Hi @ggazzo, before we merge it, could you update your branch with `release-3.0` and solve these conflicts?

ggazzo:
@denihs 
not exactly sure what you did with my branch with those merges, but I rebased it targeting to `release-3.0.4` is that what you asked for?

denihs:
Yes, thank you! I changed the base branch.

denihs (approved):

Issue #13260, opened 2024-07-29
Meteor 3.0.1: 'npm' is not a Meteor command. See 'meteor --help'


maka-io:
On Windows, after a fresh install of 3.0.1, and then creating a new project I cannot install npm modules:

```
PS C:\Users\mjc\new-app> meteor --version
Meteor 3.0.1
PS C:\Users\mjc\new-app> meteor npm install axios
'npm' is not a Meteor command. See 'meteor --help'.
PS C:\Users\mjc\new-app> ls


    Directory: C:\Users\mjc\new-app


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         7/29/2024   9:22 AM                .meteor
d-----         7/29/2024   9:21 AM                client
d-----         7/29/2024   9:21 AM                imports
d-----         7/29/2024   9:22 AM                node_modules
d-----         7/29/2024   9:21 AM                server
d-----         7/29/2024   9:21 AM                tests
-a----         7/29/2024   9:21 AM             15 .gitignore
-a----         7/29/2024   9:22 AM          34850 package-lock.json
-a----         7/29/2024   9:21 AM            645 package.json


PS C:\Users\mjc\new-app>
```

krymancer:
when I ran `meteor npm` outside the project it download meteor@2.16.0 and npm works outside the project, but when using meteor 3 I have the same problem

nachocodoner:
I run into this issue this morning when I was testing Meteor 3 in Windows from fresh installation.

Taking a note on this to fix for next patches. For the moment use global node pointing 20.x version.to run npm command.

Thank you.

krymancer:
> For the moment use global node pointing 20.x version.to run npm command.

Just to clarify, you mean using `npm <command>` instead of `meteor <npm> command` ? 

Thank you

nachocodoner:
> Just to clarify, you mean using npm <command> instead of meteor <npm> command ?

Yes, I mean that. If you struggle with managing multiple Node.js versions in your environment, use [`volta`](https://docs.volta.sh/guide/getting-started) as your Node.js manager. Make sure to use Node.js 20 and the corresponding npm version for Meteor 3 apps.

gustawdaniel:
I have the sam problem and additionally by

```
curl https://install.meteor.com/ | sh
```

that is mentioned in v3 docs https://v3-docs.meteor.com/about/install.html

i sitll installing v2

```
[...]

PR #13090, opened 2024-04-11
[3.x] Update Node 20.12.2, Npm 10.5.0 and Typescript 5.4.5 (Dev bundle 20.12.2.0)

nachocodoner:
OSS-386
OSS-385

-----

Summary of this PR [here](https://github.com/meteor/meteor/pull/13090#issuecomment-2113923381)

### Fix on Windows tests

After the 20.12.2 node.js updates, calling `spawn` with a .cmd or .bat file requires the `shell: true` option. This PR adjusts our usage accordingly.

#### Update

Tests are fixed on Windows as seen on this picture:
image

However, it doesn't finish properly [reaching this code](https://github.com/meteor/meteor/blob/typescript-5.4.4/scripts/windows/appveyor/test.ps1#L22-L26), and it times out at the end.

And locally, I can't seem to reproduce the same, locally it finishes properly.

image

Trying to debug the code around spawn of child processes, since maybe one is never stop after last adjusts. But hard to verify it since locally behaves differently.

PR #13377, opened 2024-09-30
Fix Meteor npm subcommand among others [Meteor 3]

nachocodoner:
OSS-528

Context: https://github.com/meteor/meteor/issues/13260

### Issue

[In this PR](https://github.com/meteor/meteor/pull/13090#issuecomment-2113923381) we adapted major of the Node child spawn processes in Windows, which got a breaking change, requiring pass `shell: true`. I remember some tests reacted on this, but it seems other places needed to be adapted, and failed silently. And this issue was one of them.

image

### Fix

Adding the missing `shell: true` seems to have solved the problem. Hopefully it doesn't break other flows as I experienced on the past on upgrading Node to the breaking change version and adapting some specific spawn child processes.

image

I added a regression tests for us to be able to identify any regression on `meteor npm install` command on Windows.

### DIFF
--- a/scripts/windows/ci/install.ps1
+++ b/scripts/windows/ci/install.ps1
@@ -19,6 +19,12 @@ If ($LASTEXITCODE -ne 0) {
   throw "Updating submodules failed."
 }
 
+# The `meteor npm install` subcommand should work
+& "$meteorBat" npm install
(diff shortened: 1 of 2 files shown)


denihs (approved):
Nice!!!

PR #13378, opened 2024-10-01
[Meteor 3] DDP resumptions

StorytellerCZ:
Application of #12436 for Meteor 3


packages/ddp-server/livedata_server.js
```diff
@@ -511,52 +516,61 @@ Object.assign(Session.prototype, {
     // Destroy this session, even if it's not registered at the
     // server. Stop all processing and tear everything down. If a socket
     // was attached, close it.
-
-    // Already destroyed.
-    if (! self.inQueue)
+    if (self._removeTimeoutHandle)
```
zodern: If I understand this code correctly, if the session was already disconnected, and waiting to reconnect, the app calling `close` won't prevent it from reconnecting and the session to resume. It seems this would be unexpected.

Also, the comment doesn't seem directly related to this if statement. The previous code had a gap after the comment to show that it was separate.

zodern: To add some context, the docs have said that after calling `close`, the `The client is free to reconnect, but will receive a different connection with a new id if it does.` (in https://docs.meteor.com/api/meteor.html#Meteor-onConnection).

KoenLav:
When we migrated a part of our infrastructure to Google Cloud Run we were again reminded of the need for this functionality.

Below you can see a graph of the CPU utilization, where you can see that there are spikes every 5 minutes, because the default connection timeout in Cloud Run is 5 minutes, recreating a connection and re-fetching all "master data" for all clients.

image

ps we have increased the setting to 60 minutes (the maximum) since, but would still appreciate any progress on this PR to alleviate the situation further :)

StorytellerCZ:
@KoenLav have you tried out this branch to see what impact it has?

KoenLav:
@StorytellerCZ to be honest I'm not really sure on how to run a branch of Meteor in development, much rather deploy it into production?

StorytellerCZ:
@KoenLav it is actually pretty simple. Say you checkout this branch into `~/Web/meteor`, then replace any call to `meteor` with `~/Web/meteor/meteor` and it will run Meteor from the checkout code. So if you then deploy through this it will deploy that version to production.

vlasky:
Can we please get this one incorporated into Meteor? It has been repeatedly postponed due to some issue with the test script

Lots of effort has been put into this enhancement and it has a huge benefit on performance - something that the Meteor team has focused on in recent releases.

StorytellerCZ:
@henriquealbert @nachocodoner I have merged from develop, but I've only done the conflict resolution superficially. Any help would be appreciated with this.

vlasky:
@StorytellerCZ @nachocodoner I asked ChatGPT o3 for an opinion on why the DDP Resumption enhancements fail in the test scripts. It thinks the fix is very simple. Sounds like it could be on the right track:

### The gist of what went wrong

-   **Two self-tests in the “livedata” suite fail only when the whole test-matrix is executed.**  
    When they are run in isolation they pass – exactly what @Grubba27 reported in the PR discussion [github.com](https://github.com/meteor/meteor/pull/12436)
    
-   The common thread behind those two tests is that they depend on a **clean life-cycle of DDPRouter sessions**:
    
    1.  A client disconnects (gracefully or because the transport drops).
        
    2.  The server keeps its `Session` object around for *disconnectGracePeriod* ms so that the client can **resume** it.
        
    3.  If the client fails to resume in time, `_removeSession()` is supposed to destroy the session **once**.
        
-   In the resumption implementation on this branch that life-cycle is broken in two places:
    
    | Where | What the code does today | Why the tests blow up |
    | --- | --- | --- |
    | **`livedata_server.js` – `_onClose`** | Schedules `_removeSession` on **every** transport-level close *without cancelling the previous timer*. | Multiple timers fire later, so the same session is destroyed several times. In the full test run the stray timers from an earlier test fire while the next test is already running, corrupting its expectations. |
    | **`Connection.disconnect()` in `livedata_connection.js`** | Sends an explicit `{msg:"disconnect"}` **and** immediately tears down the WebSocket. | That extra message increments the client-side `_receivedCount`, so on the next `connect` the client tells the server it has already seen one more message than the server actually sent. The server concludes the session is “out-of-date” and refuses to resume it, which is exactly what the failing tests assert *should* not happen. |
    
-   Because the default `disconnectGracePeriod` was also raised to **15 s**, the stray timers almost always fire late enough to overlap with the next test run, which is why the problem only surfaces when the whole suite is executed together [github.com](https://github.com/meteor/meteor/pull/12436).
    

### How to make the tests (and production) green again

1.  **Debounce `_removeSession`**
    
    js
    
    CopyEdit
    
    ```
    if (self._removeTimeoutHandle) {
      clearTimeout(self._removeTimeoutHandle);
    }
    self._removeTimeoutHandle = Meteor.setTimeout(() => {
      self._removeSession(session);
    }, disconnectGracePeriod);
    ```
    
    Cancelling the old timeout ensures the session is removed exactly once.
    
2.  **Do not mutate `_receivedCount` for the synthetic `disconnect` message**, or (simpler) **stop sending that message at all** – the server already knows the transport closed.
    
    js
    
    CopyEdit
    
    ```
    // Connection.disconnect
    return this._stream.disconnect(...args);
    ```
    
    Removing the extra increment keeps the client/server message counters in sync so the resume handshake succeeds.
    
3.  **Short-circuit the grace period in self-tests** by injecting a much smaller `disconnectGracePeriod` (e.g. `20` ms) via `Meteor.server.options`, so even if a timer leaks it expires before the next test begins. That makes the test suite deterministic and keeps the production default at 15 s.
    

Once those three changes are in place the two “flaky” tests pass consistently, `_removeSession` is called exactly once per session, and resumption works both in tests and in real deployments.

### You

Are you able to show the required code changes as a diff?

### ChatGPT

Below is a *minimal* patch (unified **git diff** format) that removes the two race-conditions we discovered and makes the “livedata” self-tests pass reliably.

diff

CopyEdit

```
diff --git a/packages/ddp-server/livedata_server.js b/packages/ddp-server/livedata_server.js
index 1b2c3d4..5e6f7a8 100644
--- a/packages/ddp-server/livedata_server.js
+++ b/packages/ddp-server/livedata_server.js
@@
 Session.prototype._onClose = function (socket /* …other args… */) {
   const self = this;
 
-  // Start (another) grace-period timer that will eventually
-  // remove the session if the client fails to resume it.
-  Meteor.setTimeout(() => self._removeSession(socket.id),
-                    self._disconnectGracePeriod);
+  // Cancel any previous timer so _removeSession() is called **once**
+  // per session – multiple overlapping timers were tearing the
+  // session down in the middle of the next test.
+  if (self._removeTimeoutHandle) {
+    Meteor.clearTimeout(self._removeTimeoutHandle);
+  }
+
+  self._removeTimeoutHandle = Meteor.setTimeout(() => {
+    self._removeSession(socket.id);
+    self._removeTimeoutHandle = null;          // tidy up
+  }, self._disconnectGracePeriod);
 };
```

diff

CopyEdit

```
diff --git a/packages/ddp-client/livedata_connection.js b/packages/ddp-client/livedata_connection.js
index 9fedcba..0123456 100644
--- a/packages/ddp-client/livedata_connection.js
+++ b/packages/ddp-client/livedata_connection.js
@@
 DDPConnection.prototype.disconnect = function (...args) {
-  // Tell the server we are going away *and* drop the transport.
-  if (this.status().connected) {
-    this._send({ msg: 'disconnect' });
-  }
-  return this._stream.disconnect(...args);
+  // Just drop the transport; the server’s close handler already
+  // starts the grace-period timer.  Sending the synthetic
+  // `{msg:"disconnect"}` increments `_lastReceivedMsg` on the
+  // client side and breaks the resume handshake.
+  return this._stream.disconnect(...args);
 };
```

diff

CopyEdit

```
diff --git a/packages/ddp-server/tests/livedata/livedata_tests.js b/packages/ddp-server/tests/livedata/livedata_tests.js
index aa00bb1..bb11cc2 100644
--- a/packages/ddp-server/tests/livedata/livedata_tests.js
+++ b/packages/ddp-server/tests/livedata/livedata_tests.js
@@
 Meteor._serverTestShortGrace = function () {
-  // Production default is 15 000 ms, but that lets timers from one
-  // test fire while the next test is running.
-  this.defaultDisconnectGracePeriod = 15000;
+  // Shrink the grace period during tests so stray timers expire
+  // before the next test starts.
+  this.defaultDisconnectGracePeriod = 20;
 };
```

### How the patch fixes the failures

1.  **Debounce `_removeSession`** – Clearing the previous timer before setting a new one guarantees each session is removed exactly once, so a timer from an earlier test can’t fire in the middle of the next test run.
    
2.  **Stop emitting the synthetic `"disconnect"` message** – That extra message used to bump the client’s “last-received” counter, making the server think the client was out-of-sync and refuse to resume the session.
    
3.  **Shrink the grace period while the self-tests run** – Even if a timer does leak, it now expires before the framework moves on to the next test, eliminating the intermittent failures seen when the whole suite is executed.

StorytellerCZ:
I'll give this a try. Though looking at the provided patches I'm not seeing locally any of those lines.

StorytellerCZ:
Well, would you look at that! @vlasky, @KoenLav can you please take it out for spin in your application to make sure that things still work as you want? With the tests now passing I think we can move forward with the review and merger.

Grubba27:
CI seems good; I'll try reviewing the code again

StorytellerCZ:
Merging from `develop` to see how the tests do again, just to be sure.

KoenLav:
@StorytellerCZ we're not yet on Meteor v3, so for us it will take some time before we can get to this, sadly.

vlasky:
@StorytellerCZ I have installed the ddp-client and ddp-server packages from your PR in a recently migrated project now running Meteor 3.3. Everything is running smoothly so far.

vlasky:
@nachocodoner now that the tests pass, can we please have DDP resumptions included in upcoming Meteor release 3.4?

docs/source/api/connections.md
```diff
@@ -121,16 +121,19 @@ connection is already closed, the callback will be called immediately.
...
+
+<dl class="objdesc">
+{% dtdd name:"Meteor.server.options.disconnectGracePeriod" type:"Number" default:"15000" %}
+  Defines how long (in milliseconds) we should maintain a session for after a non-graceful disconnect before destroying it. Sessions that reconnect within this time will be resumed with minimal performance impact. Defaults to `15000`.
+{% enddtdd %}
+
+{% dtdd name:"Meteor.server.options.maxMessageQueueLength" type:"Number" default:"100" %}
+Determines how many messages we should queue during a non-graceful disconnect before we destroy the session, to insure against memory leaks. Defaults to `100`.
+{% enddtdd %}
+</dl>
```
Grubba27: We could have this in the v3-docs/docs ?

zodern:
I think my comment at https://github.com/meteor/meteor/pull/13378/files/6653e9ab800ed0e037fe95f8cac71884c8b4847e#r1819275299 still applies.

docs/source/api/connections.md
StorytellerCZ: Yes

Grubba27:
Hey @vlasky!

on: 

> can we please have DDP resumptions included in upcoming Meteor release 3.4?

Tests are passing, docs are there, but I think we are missing some reviews. I would love to see Zodern's comments addressed, and possibly if @radekmie sees something in this PR, I would like his comments (if he makes any) to be addressed too.

I'll take a look, but I'm not super familiar with this code path.

With that in mind, if we get all of this, then I think we would be in an excellent spot for testing.

packages/ddp-server/livedata_server.js
StorytellerCZ: Will restore the spacing from the comment. For the rest I will look into it when I get some time, which will take a while, so if there is anyone else interested then feel free to take on.

vlasky: @StorytellerCZ looks like a simple fix:

Replace:

```
    if (self._removeTimeoutHandle)
      return;
```
with
```     
    if (self._removeTimeoutHandle) {
      Meteor.clearTimeout(self._removeTimeoutHandle);
      self._removeTimeoutHandle = null;
    }
    
    self._expectingDisconnect = true;
```

StorytellerCZ: Applied the suggestion. Let's see what the tests say.

vlasky:
> Tests are passing, docs are there, but I think we are missing some reviews. I would love to see Zodern's comments addressed, and possibly if @radekmie sees something in this PR, I would like his comments (if he makes any) to be addressed too.
> 
> I'll take a look, but I'm not super familiar with this code path.

Thank you @Grubba27. Appreciate your time.

@Zodern - I think the latest commit has addressed your concern. Would appreciate if you could have another look.

packages/ddp-server/livedata_server.js
```diff
@@ -391,6 +410,13 @@ Object.assign(Session.prototype, {
       return;
     }
 
+    if (msg_in.msg === 'disconnect') {
+      if (_.has(self.protocol_handlers, msg_in.msg)) {
```
radekmie: Can we replace it with a non-underscore alternative?

packages/ddp-server/livedata_server.js
```diff
@@ -287,51 +292,65 @@ Object.assign(Session.prototype, {
...
-
-    // Unregister the session.
-    self.server._removeSession(self);
   },
 
   // Send a message (doing nothing if no socket is connected right now).
   // It should be a JSON object (it will be stringified).
   send: function (msg) {
-    const self = this;
+    var self = this;
```
radekmie: I guess such changes are a result of rebasing the old PR, but it'd be nice to keep the better version as it is.

packages/ddp-server/livedata_server.js
```diff
@@ -287,51 +292,65 @@ Object.assign(Session.prototype, {
...
+    const ignoredMsgsForSessionOutOfDateCheck = ['ping', 'pong'];
```
radekmie: This method will be called _very_ often - moving this array out of here would reduce the GC pressure. (Alternatively, the below condition can be inlined into two `===`s.)

packages/ddp-client/common/livedata_connection.js
```diff
@@ -1439,4 +1562,154 @@ export class Connection {
...
+
+    if (msg === null || !msg.msg) {
+      if(!msg || !msg.testMessageOnConnect) {
+        if (Object.keys(msg).length === 1 && msg.server_id) return;
+        Meteor._debug('discarding invalid livedata message', msg);
+      }
+      return;
+    }
+    const ignoredMsgsForSessionOutOfDateCheck = ['ping', 'pong'];
+    if (!ignoredMsgsForSessionOutOfDateCheck.includes(msg.msg)) {
```
radekmie: The same check happens here.

radekmie (commented):
Overall, I only have a bunch of nitpicks. Though I'm not the most knowledgeable about this part of the codebase.

packages/ddp-server/livedata_server.js
StorytellerCZ: Yes

packages/ddp-server/livedata_server.js
StorytellerCZ: I'll move this out so that we can add to it in the future if needed.

packages/ddp-server/livedata_server.js
StorytellerCZ: Fixed

StorytellerCZ:
Fixed @radekmie comments. Let's see how the tests do.

vlasky:
@StorytellerCZ @nachocodoner I've continued our valuable work on DDP resumptions in https://github.com/meteor/meteor/pull/14051.

StorytellerCZ:
Closing in favor of https://github.com/meteor/meteor/pull/14051

PR #13380, opened 2024-10-01
Node v22.11.0

StorytellerCZ:
Node v22.11.0 for Meteor 3.1.
~I suspect that this will be bumped to Node v22.10.0 before the final release as LTS version comes around, but we should get this in asap to start testing and preparing.~

The images for testing on CircleCI and so on will have to be updated as well.


nachocodoner:
Great. We’ll handle this when we move to Meteor 3.1.

Did you flag any breaking changes we need to address when upgrading to this Node major LTS version?

StorytellerCZ:
# Potentially breaking
## 22.0.0
https://github.com/nodejs/node/releases/tag/v22.0.0
* esm: drop support for import assertions

## 22.1.0
https://github.com/nodejs/node/releases/tag/v22.1.0

## 22.2.0
https://github.com/nodejs/node/releases/tag/v22.2.0

## 22.3.0
https://github.com/nodejs/node/releases/tag/v22.3.0

## 22.4.0
https://github.com/nodejs/node/releases/tag/v22.4.0

## 22.5.0
https://github.com/nodejs/node/releases/tag/v22.5.0

## 22.6.0
https://github.com/nodejs/node/releases/tag/v22.6.0

## 22.7.0
https://github.com/nodejs/node/releases/tag/v22.7.0
* Module syntax detection is now enabled by default

## 22.8.0
https://github.com/nodejs/node/releases/tag/v22.8.0

## 22.9.0
https://github.com/nodejs/node/releases/tag/v22.9.0

# Opportunities
## 22.0.0
* WebSocket client
* (SEMVER-MAJOR) lib: enable WebSocket by default

## 22.1.0
* module: implement NODE_COMPILE_CACHE for automatic on-disk code caching
* (SEMVER-MINOR) url: implement parse method for safer URL parsing (possible improvements in `webapp`)

## 22.5.0
* http: expose websockets
* lib: add node:sqlite module
* module: add __esModule to require()'d ESM
* stream: pipeline wait for close before calling the callback

## 22.6.0
* Experimental TypeScript support via strip types

## 22.7.0
* Experimental transform types support
* lib: rewrite AsyncLocalStorage without async_hooks

## 22.8.0
* New JS API for compile cache

## 22.9.0
* New API to retrieve execution Stack Trace

nachocodoner:
> 22.7.0
> lib: rewrite AsyncLocalStorage without async_hooks https://github.com/nodejs/node/pull/48528
> --experimental-async-context-frame

I am particularly interested in how this will positively impact Meteor 3's performance. I have been researching, and it seems that the issue with AsyncLocalStorage using async_hooks has been a significant concern for applications built on previous Node versions. Our benchmarks and metrics also highlight this inefficiency. The new version may already provide some performance gains in our core usage of AsyncLocalStorage. After upgrading, we will run the benchmarks to verify this with `--experimental-async-context-frame` enabled.

cc @leonardoventurini

.travis.yml
```diff
@@ -4,7 +4,7 @@ dist: jammy
 sudo: required
 services: xvfb
 node_js:
-  - "20.15.1"
+  - "22.9.0"
```
nachocodoner: @denihs: Revert this since it causes the CI to fail, check out this comment for further information, https://github.com/meteor/meteor/pull/13252#discussion_r1731330223

We could try to fix it as part of this PR, but maybe is good to focus on the Node upgrade and adapt possible breaking changes if any, then focus more on the CI upgrade. But CI Node version doesn't matter too much, since Meteor dev_bundle already uses the proper Node version to pass the checks.

StorytellerCZ: Reverted. If possible, I think it would be better to spend some time to migrate Travis test to either CircleCI or GitHub to reduce the number of CI services.

nachocodoner: Yes, we are working towards using GitHub Actions for all checks. We have already moved some checks, but there are still many left. Still, it's better to spend time migrating the checks rather than fixing issues with services like Travis.

StorytellerCZ:
Added Typescript 5.6.3 patch update to this as Typescript 5.6.2 will be in Meteor 3.0.4.

StorytellerCZ:
Node 22.10.0 has been released:
https://github.com/nodejs/node/releases/tag/v22.10.0

```
New "module-sync" exports condition
This release introduces a "module-sync" exports condition that's enabled when
require(esm) is enabled, so packages can supply a synchronous ES module to the
Node.js module loader, no matter if it's being required or imported. This is
similar to the "module" condition that bundlers have been using to support
require(esm) in Node.js, and allows dual-package authors to opt into ESM-first
only on newer versions of Node.js that supports require(esm) to avoid the
dual-package hazard.
```

StorytellerCZ:
Since node has been updated to 22.10.0, should we make a new branch for the dev bundle?

scripts/generate-dev-bundle.ps1
```diff
@@ -378,6 +379,10 @@ $npmToolArgs = @{
 }
 Add-NpmModulesFromJsBundleFile @npmToolArgs
 
+# Leaving these probably doesn't hurt, but are removed for consistency w/ Unix.
+Remove-Item $(Join-Path $dirLib 'package.json')
+Remove-Item $(Join-Path $dirLib 'package-lock.json')
```
nachocodoner: Why do we need to remove these files?

In fact, in a recent fix I discovered that these files don't need to be removed in par on what happens with Unix dev_bundle, for few issues with Cordova in Windows to be fixed. ([context](https://github.com/meteor/meteor/pull/13388))

denihs: hmm... I'm going to bring these back

StorytellerCZ:
Node 22.11.0 has been released:
https://github.com/nodejs/node/releases/tag/v22.11.0
* LTS
* OpenSSL v3

leonardoventurini (approved):

Issue #13250, opened 2024-07-24
Meteor 3 add-platform android breaks meteor installation on windows 11


ibassoftware:
I’ve encountered a problem when attempting to add the Android platform to a Meteor project on a Windows 11 system. This is a fresh Meteor 3.0.1 installation by the way. Here are the details of the issue:

When running the command meteor add-platform android, the following error occurs:

`
Error: Error: couldn't install npm package cordova-common@4.0.2: Command failed: C:\Windows\system32\cmd.exe /c C:\Users\rrd\AppData\Local\.meteor\packages\meteor-tool\3.0.1\mt-os.windows.x86_64\dev_bundle\bin\npm.cmd install cordova-common@4.0.2
node:internal/modules/cjs/loader:1148
  throw err;
  ^

Error: Cannot find module 'C:\Users\rrd\AppData\Local\.meteor\packages\meteor-tool\3.0.1\mt-os.windows.x86_64\dev_bundle\lib\node_modules\npm\bin\npm-cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
    at Module._load (node:internal/modules/cjs/loader:986:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v20.15.1
`

After this error, attempting to run any meteor command results in the following error:

`
node:internal/modules/cjs/loader:1148
  throw err;
  ^

Error: Cannot find module '@meteorjs/reify/lib/runtime'
Require stack:
- C:\Users\rrd\AppData\Local\.meteor\packages\meteor-tool\3.0.1\mt-os.windows.x86_64\tools\static-assets\server\runtime.js
- C:\Users\rrd\AppData\Local\.meteor\packages\meteor-tool\3.0.1\mt-os.windows.x86_64\tools\tool-env\install-reify.js
- C:\Users\rrd\AppData\Local\.meteor\packages\meteor-tool\3.0.1\mt-os.windows.x86_64\tools\tool-env\install-babel.js
- C:\Users\rrd\AppData\Local\.meteor\packages\meteor-tool\3.0.1\mt-os.windows.x86_64\tools\index.js
    at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
    at Module._load (node:internal/modules/cjs/loader:986:27)
[...]

PR #13388, opened 2024-10-03
Fix Cordova on Windows

nachocodoner:
Context: https://github.com/meteor/meteor/issues/13250, https://github.com/meteor/meteor/issues/13250, [forum post #1](https://forums.meteor.com/t/meteor-3-add-platform-android-breaks-meteor-installation-on-windows-11/61894/2), [forum post #2](https://forums.meteor.com/t/meteor-missing-modules/62354)

This PR continues solving Cordova development, this time on Windows, since it seems that latest Node versions produced a problem on the Meteor bundle that require dependencies to be locked.

## Issue

Using the `meteor add-platform android` command while setting up the Cordova project with the necessary npm dependencies corrupts the `dev_bundle`, removing critical dependencies. This can leave the entire Meteor installation unusable, requiring a complete reinstallation of Meteor. In practice, Cordova development does not work on the latest Meteor versions (it works on WSL, however).

After thorough debugging, I identified the issue. The problem is that the `dev_bundle` generated for Windows is missing the creation of the `package.json` context and the lock file in the dev_bundle's `lib/` directory. During the preparation of the Cordova project, it creates these files but removes existing critical dependencies (like npm) that are not locked, leading to the complete break of Meteor setup and the reported issues.

See the attached picture for clarification.

image

## Solution

We create a Meteor `dev_bundle` specifically for Unix and Windows systems. After reviewing the code, [I found that we explicitly removed these files](https://github.com/meteor/meteor/blob/devel/scripts/generate-dev-bundle.ps1#L381-L383) in earlier Meteor versions. However, newer Node versions require these dependencies to fix the dev_bundle. Likewise, for Unix, as part of Meteor 3, [we specifically copied these files](https://github.com/meteor/meteor/blob/devel/scripts/generate-dev-bundle.sh#L151-L154).

This PR simply updates the Windows scenario to align with the same expectation as Unix.

## Testing

The issue is solved:

image

However, I am performing tests to verify Cordova setup would work well now in raw Windows. I found another issue now with cordova running itself, so after adding proper Java and Android SDK envs, when running other error faced now ([error](https://github.com/user-attachments/assets/cfea1b6c-6e25-45fc-9e95-dc4129023ce0))

This PR is going to be a full review of Windows Cordova and ensure Cordova works as expected.

### DIFF
--- a/scripts/generate-dev-bundle.ps1
+++ b/scripts/generate-dev-bundle.ps1
@@ -378,10 +378,6 @@ $npmToolArgs = @{
 }
 Add-NpmModulesFromJsBundleFile @npmToolArgs
 
-# Leaving these probably doesn't hurt, but are removed for consistency w/ Unix.
-Remove-Item $(Join-Path $dirLib 'package.json')
-Remove-Item $(Join-Path $dirLib 'package-lock.json')
-
 Write-Host "Done writing node_modules build(s)..." -ForegroundColor Magenta
 
 Write-Host "Removing temp scratch $dirTemp" -ForegroundColor Magenta
--- a/tools/cordova/project.js
+++ b/tools/cordova/project.js
@@ -461,9 +461,9 @@ to Cordova project`, async () => {
       // As per Npm 8, we need now do inject a package.json file
       // with the dependencies so that when running any npm command
       // it keeps the dependencies installed.
-      const packageLock = JSON.parse(files.readFile(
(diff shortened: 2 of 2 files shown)


denihs (approved):

nachocodoner:
I am going to merge this for the purpose to include this fixes integrated into next release beta. However, the fix of Cordova in Windows will continue, since is not fixed completely yet.

PR #13389, opened 2024-10-03
Node 20.18.0

StorytellerCZ:
Node v20.18.0 just got released:
https://github.com/nodejs/node/releases/tag/v20.18.0

I don't see anything that should prevent us from using this release.


StorytellerCZ:
I think including this should be good enough to then release RC for 3.0.4

nachocodoner:
I think I will branch out a new branch. I see something weird on this branch related with blaze, maybe will happen the same on new branch but just double checking.

.travis.yml
```diff
@@ -4,7 +4,7 @@ dist: jammy
 sudo: required
 services: xvfb
 node_js:
-  - "20.15.1"
+  - "20.18.0"
```
nachocodoner, line 7: Also, I made a note about sticking into 20.15.1, and not upgrade further, https://github.com/meteor/meteor/pull/13252#discussion_r1731330223

We will research this further. We just want to include for the moment the Node upgrade as part of the next 3.0.4 release.

.circleci/config.yml
```diff
@@ -178,7 +178,7 @@ jobs:
           command: |
             eval $PRE_TEST_COMMANDS;
             cd dev_bundle/lib
-            ../../meteor npm install @types/node@20.10.5 --save-dev
+            ../../meteor npm install @types/node@20.18.0 --save-dev
```
nachocodoner, line 181: This version doesn't exist for `@types/node`. Or at least CI was failing because of this.

PR #13401, opened 2024-10-07
Update dependencies in project root

StorytellerCZ:
Updated dependencies in project root and resolved their conflicts.
Still could potentially update further, but I kept to feature updates and then major version upgrades only when required by another dependency.


denihs (approved):

italojs:
PR outdated

PR #13402, opened 2024-10-08
Official roles package

StorytellerCZ:
Import history of the roles package.
After this will need to integrate tests and documentation.


StorytellerCZ:
@nachocodoner @denihs any idea why CircleCi fails? There are no tests added right now, so it should not be affected.

denihs:
The issue was with the `api.versionsFrom` set. I think now everything will become green.

StorytellerCZ:
Of course! Should have noticed that!

StorytellerCZ:
I think we can merge this and the next step is to migrate tests and documentation.

denihs:
Done! The first published version will be done in the next beta/rc 🙌

StorytellerCZ:
Great! I will see if I can find some time to get documentation included, though maybe we might want to hold on that for one more version so that people in the known can be the early adopters and find bugs and in the next version we will add docs and guide for everyone. Thoughts?

PR #13406, opened 2024-10-09
feat: refactor mongo and update mongo driver to v6

leonardoventurini:
- Remove unecessary `bindEnvironment` calls (ALS) from high-frequency real-time code.
- Split large files and refactor code to TypeScript.
- Clean up database before running tests which helps when running them locally on restarts
- Prevent Tinytest invocation context from trickling down to actual test function


packages/mongo/observe_handle.ts
```diff
@@ -0,0 +1,37 @@
+let nextObserveHandleId = 1;
```
harryadel, line 3: Are we going full TS from now on? :sweat_smile: @leonardoventurini

leonardoventurini: Yes, the idea is to gradually refactor the `mongo` package and new things to be built using TypeScript if possible.

The real-time logic is complex and hard to understand as it is now, by refactoring we would increase quality, performance and reduce the barrier for contributions in the long term. It's the perfect starting point for committing to TypeScript.

Real-time is one of the key features of Meteor and needs to be performant and reliable so people want to use it. In this case I see TypeScript as a requirement.

The same way you choose the materials for a building, in some scenarios you need to opt for harder and more resilient materials.

Any thoughts?

harryadel: Nothing really from my end, I was just curious as rarely nothing is implemented in TS in Meteor. I thought the only stuff we might ever need to implement in TS are the core [tools](https://github.com/meteor/meteor/tree/devel/tools).

leonardoventurini: Ok :) I would say core tools would come right after `mongo` in priority, architecturally they are a centerpiece but in a more user centric view real-time is where the most value is for Meteor, and we are seeing performance challenges after moving to ALS that we need resolved or fully understood.

Real-time/mongo has direct impact on performance, DX, UX and adoption, getting it in great shape allows us to do the same to other key areas more easily.

packages/mongo/oplog_observe_driver.js
```diff
@@ -113,9 +118,6 @@ OplogObserveDriver = function (options) {
 
   self._requeryWhenDoneThisQuery = false;
   self._writesToCommitWhenWeReachSteady = [];
-
-
-
  };
 
 _.extend(OplogObserveDriver.prototype, {
```
StorytellerCZ: This should be `Object.assign`

leonardoventurini: I think that change is outdated 😄

nachocodoner (approved):

PR #13408, opened 2024-10-09
Upgrade Express 5

nachocodoner:
OSS-571

This PR upgrades Express to the latest version 5. This will be included in the Meteor 3.1.x release.

You can find the full list of migration steps and breaking changes [here](https://expressjs.com/en/guide/migrating-5.html).

I've found only one breaking change in the Meteor Core. I’ve identified one regarding `request.query`, which is now read-only. We currently write to this field as part of an Express middleware for other endpoints, such as OAuth. I will provide tests to compare the current query parsing behavior with the new behavior and explore solutions to address this breaking change.

However, **this upgrade in general may require numerous migration steps for existing apps that fully utilize Express**, so we will decide on the release timing, whether to include it with other changes or isolate it.

### DIFF
--- a/packages/webapp/package.js
+++ b/packages/webapp/package.js
@@ -5,8 +5,8 @@ Package.describe({
 
 Npm.depends({
   "cookie-parser": "1.4.6",
-  express: "4.21.0",
-  "@types/express": "4.17.21",
+  express: "5.0.1",
+  "@types/express": "5.0.0",
   compression: "1.7.4",
   errorhandler: "1.5.1",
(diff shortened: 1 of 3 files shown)


StorytellerCZ:
I think this might be better for Meteor 3.1 than a patch release.

nachocodoner:
> I think this might be better for Meteor 3.1 than a patch release.

Yes. The reason is compared with 3.0.4 is to get latest version and devbundle. That will be merged soon and spread to 3.1. It will be changed then.

PR #13409, opened 2024-10-10
Express 4.21.1

StorytellerCZ:
I would like to suggest to include this PR in Meteor 3.0.1 instead of #13408 as that one is more suited for Meteor 3.1.
This is just a patch update of `express`.


nachocodoner:
We have released the 3.0.4-rc.0. This is just a patch of a version so there won't be any problem to release it in the official release.

I will communicate to the team, and we will likely merge. Thanks

PR #13421, opened 2024-10-18
Pass extra to Accounts.urls.loginToken

StorytellerCZ:
Implementing suggestion from #13413


denihs (approved):

nachocodoner:
Are you aware if we support types on this API?

StorytellerCZ:
@nachocodoner which one? `Accounts.sendLoginTokenEmail` or `Accounts.urls.loginToken`. I haven't checked, but in either case I don't think that would have any impact on this PR.

nachocodoner (approved):

PR #13423, opened 2024-10-18
feat: refactor `bindEnvironment` and `EnvironmentVariable`

### LINKED ISSUE #13420: Meteor 3: server-side `bindEnvironment` doesn't preserve all values
Consider the following server-side code in a fresh application:

```js
const varA = new Meteor.EnvironmentVariable("a");
const varB = new Meteor.EnvironmentVariable("b");

Meteor.startup(async () => {
  varA.withValue(1, () => {
    varB.withValue(2, () => {
      Meteor.defer(() => {
        console.log(varA.get()); // undefined
        console.log(varB.get()); // undefined
      });
    });
  });

  varA.withValue(1, () => {
    varB.withValue(2, () => {
      Meteor.defer(
        Meteor.bindEnvironment(() => {
          console.log(varA.get()); // undefined
          console.log(varB.get()); // 2
        })
      );
    });
  });
});
```

Invoking `Meteor.bindEnvironment` should preserve the values of the active environment variables, but instead only the variable that is set last is preserved, and other variables return `undefined`. 

Meteor 2.16 does indeed print 1, 2, 1, 2, as expected.

Looking at [the implementation](https://github.com/meteor/meteor/blob/024061123a49ff1d4b89d69453ec9263ef8599e1/packages/meteor/dynamics_nodejs.js#L171), I don't understand how it should work. The comments seem a bit dated, but it appears that only the `CURRENT_VALUE_KEY_NAME` value is restored, which is the recent variable's value in this case. That value and all others seem to be stored in the `UPPER_CALL_DYNAMICS_KEY_NAME`, but that one is not (re)stored.

Strangely, `Meteor.defer` doesn't preserve any values although it uses `bindEnvironment` under the hood, but perhaps that has the same root cause.

The issue reproduces with a fresh Meteor 3.0.3 application on Linux.

leonardoventurini:
After removing fibers, we actually shouldn't need `bindEnvironment,` but we can keep it there to preserve some backward compatibility.

The `EnvironmentVariable` is actually very similar to `AsyncLocalStorage,` and we don't need to reinvent the wheel there at an additional cost of performance.

This fixes https://github.com/meteor/meteor/issues/13420


packages/meteor/setimmediate.js
```diff
@@ -1,141 +1,4 @@
...
-
-
-Meteor._setImmediate =
-  useSetImmediate() ||
-  usePostMessage() ||
-  useTimeout();
+/**
+ * Make sure that `setImmediate` works both on the server and the client. Uses a microtask in the client.
+ */
+Meteor._setImmediate = Meteor.isServer ? setImmediate : fn => Promise.resolve().then(fn)
```
zodern: setImmediate needs to schedule a macrotask. The purpose of the api is to schedule some work after any already queued macrotasks, without adding unnecessary delays like setTimeout would (which is why the browser polyfills usually use postMessage). Using a microtask would do the opposite - schedule the work to run before any other macrotask.

For example, I use it in one app to break apart a large amount of synchronous work so it finishes as quickly as possible without making the page unresponsive. Microtasks would instead block the browser from running event handlers or rendering until it finished.

This also will break how the DDP queueing works on the client with method stubs since it relies on setImmediate creating macrotasks.

packages/meteor/dynamics_nodejs.js
```diff
@@ -15,6 +14,7 @@ const SLOT_CALL_KEY = "slotCall";
 class EnvironmentVariableAsync {
   constructor() {
     this.slot = nextSlot++;
+    this.als = new AsyncLocalStorage()
```
zodern, line 17: The profiling I did a few months ago seemed to indicate that the number of ALS stores could affect performance - each store has to propagate its values when a new resource is created.

I'm not sure how much of an effect it would have, but the current implementation has some duplicate and unnecessary work that could be optimized.

packages/meteor/dynamics_nodejs.js
```diff
@@ -159,40 +120,19 @@ Meteor.bindEnvironment = (func, onException, _this) => {
...
-          return ret;
-        },
-        _this,
-        {
-          callId: `bindEnvironment-${Math.random()}`,
-          [SLOT_CALL_KEY]: currentSlot,
-        }
-      );
-    };
+  return function (...args) {
```
zodern, line 129: I'm not sure this matches the existing behavior:

```
let cb;
ev.withValue(5, () => {
  cb = Meteor.bindEnvironment(() => console.log(ev.get());
});
 
// This should log 5, but with this PR it seems it would log 4
ev.withValue(4, () => cb());
```

packages/meteor/setimmediate.js
leonardoventurini: Oh, I see. Will adjust, thanks!

packages/meteor/dynamics_nodejs.js
leonardoventurini: Yeah, still trying to understand this better and also see the performance impact of certain behaviors. This PR is very experimental still 😄

packages/meteor/dynamics_nodejs.js
leonardoventurini: I am trying to understand if that would be preferrable compared to having complex logic in `bindEnvironment`... if CPU is a little worse and the event loop is fine then it would be great. Also, Node.js is going to bring optimizations for ALS in the next releases which might help in this scenario.

packages/meteor/setimmediate.js
leonardoventurini: I adjusted and included a comment there with this info.

leonardoventurini:
image

packages/meteor/dynamics_nodejs.js
zodern: > compared to having complex logic in bindEnvironment

The implementation is overly complex. I created #13425 to get it down to one piece of state stored in ALS.

leonardoventurini:
No longer needed

PR #13428, opened 2024-10-24
Update 3.forms-and-events.md

JacobKGilbert:
In the instructions to add onSubmit and onChange to TaskForm.jsx there was an unused import for TaskCollections. Removing this had no effect on the useability of the code in this step. Also, to avoid confusion for new users, a brief description of how Meteor.callAsync can work without importing Meteor into TaskForm.jsx.


JacobKGilbert:
Also, it might be helpful for first time users to have an explanation just under that block as to why Meteor.callAsync can be run without importing Meteor into this file.

StorytellerCZ (approved):

StorytellerCZ:
I think instead of explanation it would be better to add the import, so that we teach best practices and prepare for the future.

italojs (approved):

Issue #13218, opened 2024-07-05
Update MongoDB driver to v6


StorytellerCZ:
So just noticed that we are two major versions behind the MongoDB Node driver.
Currently on 4.17.2. I would like to bring this up for an upgrade.

I think the reason why this has not been done is that [v5](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.0.0) of the package has removed callbacks, which is an issue for Meteor 2.x and bellow, but I think that for Meteor 3 it should be fine.

So steps for upgrade going forward:
## [5.0.1](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.0.0)
* Remove any remaining callbacks from server or make a workaround with the compatibility package

## [5.1.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.1.0)
* Check if we need to do anything with the new `bigint`

## [5.2.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.2.0)
Doesn't seem like we need to do anything here.

## [5.3.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.3.0)
* cursor forEach has been deprecated, we need to do the same, I think this is going to be huge as I have seen quiet a few things build on top of this.

## [5.4.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.4.0)
More deprecations here, but nothing that should impact us,  I think.

## [5.5.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.5.0)
Nothing of note

## [5.6.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.6.0)
* Node 20 support
I think this is the initial release that we should really target.

## [5.7.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.7.0)
* Lot of deprecations
* Support for change stream split events - I think stuff like this will be become crucial for moving of oplog tailing
* Search indexes

## [5.8.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.8.0)
Should be fine

## [5.9.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.9.0)
* Fixes memory leak with change streams

## [6.0.0](https://github.com/mongodb/node-mongodb-native/releases/tag/v6.0.0)
* Minimum Node v16
* We might need to install `socks` dependency
* Callbacks removed from ClientEncryption's encrypt, decrypt, and createDataKey methods
[...]

PR #13429, opened 2024-10-24
Upgrade Mongo driver 5.x

nachocodoner:
OSS-572

Context: https://github.com/meteor/meteor/issues/13218

This PR upgrades the Mongo driver [to version 5.x](https://github.com/mongodb/node-mongodb-native/releases/tag/v5.0.0). Since this version has breaking changes, we’ll assess how it impacts Meteor core. If the changes are straightforward, we might also upgrade to version 6.x. The goal is to include at least one major version update in Meteor 3.1.

The main breaking change in Mongo driver 5.x is the removal of callbacks in favor of Promises. We will adapt to this case by case and ensure all test coverage behaves properly. This will be also a major change for those devs that use older mongo driver directly.


nachocodoner:
Closing in favor of https://github.com/meteor/meteor/pull/13434, as that PR has resolved test failures, with most tests now passing. It appears we’re less dependent on Mongo Node driver callbacks within the core.

PR #13432, opened 2024-10-25
test: bring test-in-console to github workflow

leonardoventurini:
the goal is to drop travis ci in favor of github workflow/action, however it has been challenging

---

fun fact:

the bundled mongodb was using rosetta still, and sometimes the process would hang and I could not kill it not even with fire, by using the `arm64` version the issue stops happening:

https://github.com/meteor/meteor/pull/13432/commits/7f79ade8cd9a63f4b4fcdeaec671f9669db23b69


leonardoventurini:
it takes a long time to start in gh workflow compared to travis ci, perhaps it is downloading chrome or smth

leonardoventurini:
image

StorytellerCZ:
Not too fond of an outright delete. Given that the package will exists we should archive the code somewhere for the future. Maybe even better to update the offending dependency and releasing the fix before moving it into its own repository and archiving it.

leonardoventurini:
> Not too fond of an outright delete. Given that the package will exists we should archive the code somewhere for the future. Maybe even better to update the offending dependency and releasing the fix before moving it into its own repository and archiving it.

yeah me neither, I just deleted in order to focus on the big problem, once that is resolved I will try dealing with the package. for now the workflow times out for some unknown reason, only on github

leonardoventurini:
Not sure, what dark sorcery is happening on Github, parking this for now

PR #13440, opened 2024-10-31
fix(minimongo): `update/remove` with operator only applies one record

ggazzo:
checking the code we can see that in the past we were used to do `fn(doc, id) === false`
in this pr https://github.com/meteor/meteor/pull/12504 was changed to `!fn(doc, id)`

`fn(doc, id) === false` and `!fn(doc, id)` are slightly different but causes huge problem if we are considering/ignoring undefined

if we use operators like `$in`, `uptade`/`remove` uses  `_eachPossiblyMatchingDocSync` which returns `undefined`

```diff
if (specificIds) {
-      specificIds.some(id => {
-        const doc = this._docs.get(id);
-        if (doc) {
-          return fn(doc, id) === false;
-        }
-       });
+      for (const id of specificIds) {
+        const doc = this._docs.get(id);
+        if (doc && !fn(doc, id)) {
+          break
+        }
+      }
}
```

with that only one record is affected by the operation


ggazzo:
@denihs do you need me to do anything else here?

Grubba27:
`tinytest - minimongo - updateAsync` multiple using query operator failed seems to be failing

italojs:
@ggazzo do you have plans to fix this PR?

PR #13445, opened 2024-11-04
feat: performance improvements

leonardoventurini:
OSS-604

Due to the nature of the iterative research, this PR ended up large, mixing plain refactor and file splits with actual improvements, I will point out the actual improvements to facilitate review.

In sum, this PR should greatly improve performance in Meteor 3, but requires further validation (WIP).

---

- Refactored AsynchronousQueue to support parallel processing with configurable batching
- Optimized MongoDB observers to prioritize initial document adds
- Added batched, concurrent task execution with optional ordered processing
- Performance improvements through parallel processing and better promise handling

https://github.com/meteor/meteor/pull/13445/commits/12993e9f2e8c53112b37c81f5bca7319d1745e1a

---

Improve compression defaults and include optimized version of `permessage-deflate` which supports setting a threshold of message size to start compressing.

This reduces CPU and memory significantly. Apps that consume large amounts of bandwidth but have CPU to spare can configure this via env var.

https://github.com/meteor/meteor/pull/13445/commits/b5fd7ac4308bec917d041bec3b75bab8b56dfa25

https://github.com/leonardoventurini/permessage-deflate2/pull/1

---

Added docs for `Performance > WebSocket Compression`:

https://github.com/meteor/meteor/pull/13445/commits/c73515470f5a4a623b8403231fbfa6c8472cd5b1

---

Other changes that didn't produce noticeable improvements, but worth keeping:

https://github.com/meteor/meteor/pull/13445/commits/748f7c4f29d87907af8eaeaa9e87d3c09fdef8c0
https://github.com/meteor/meteor/pull/13445/commits/4e8c7cf1815298da77f206c2b94d9b52f12e5e60
https://github.com/meteor/meteor/pull/13445/commits/6b5248e2367c061f0c8c34704f93166ca7f1c9d8


leonardoventurini:
Not sure if it is just flaky or an actual issue, adding it here:

image

alisnic:
This is very exciting 🤩

packages/ddp-client/client/client_convenience.js
```diff
@@ -1,6 +1,6 @@
 import { DDP } from '../common/namespace.js';
 import { Meteor } from 'meteor/meteor';
-import { loadAsyncStubHelpers } from "./queueStubsHelpers";
+import { loadAsyncStubHelpers } from "./queue_stub_helpers";
```
Grubba27, line 3: loved the name change!

packages/meteor/asynchronous_queue.js
Grubba27, line 1: This could be an easy win to typescript.


We only have 8 usages in our codebase

packages/mongo/observe_handle.ts
```diff
@@ -15,11 +15,14 @@ export class ObserveHandle {
...
-  _movedBefore?: (...args: any[]) => void;
-  _removed?: (...args: any[]) => void;
+  public initialAddsSentResolver: (value: void) => void = () => {};
+  public initialAddsSent: Promise<void>
+
+  _added?: (...args: any[]) => Promise<void>;
+  _addedBefore?: (...args: any[]) => Promise<void>;
+  _changed?: (...args: any[]) => Promise<void>;
+  _movedBefore?: (...args: any[]) => Promise<void>;
+  _removed?: (...args: any[]) => Promise<void>;
```
Grubba27:
```suggestion
  _added?: (...args: any[]) => Promise<void> | void;
  _addedBefore?: (...args: any[]) => Promise<void> | void;
  _changed?: (...args: any[]) => Promise<void> | void;
  _movedBefore?: (...args: any[]) => Promise<void> | void;
  _removed?: (...args: any[]) => Promise<void> | void;
```
We could even do something like:

`class ObserveHandle<T = any>` and those methods be something like `(...args: T[]) => Promise<void> | void;`

And in the instance of the handle we could pass in the type of the collection(at least the name idk)

packages/mongo/observe_multiplex.ts
```diff
@@ -140,7 +140,7 @@ export class ObserveMultiplexer {
     return !!this._isReady;
   }
 
-  _applyCallback(callbackName: string, args: any[]): void {
+  async _applyCallback(callbackName: string, args: any[]) {
     this._queue.queueTask(async () => {
```
Grubba27: is really necessary turning this function to async? (this is not awaitable–as we do not [return the `scheduleRun` in AsyncQueue](https://github.com/meteor/meteor/blob/d27946d0c0bdc8a9e41dd4b37e828327d480560c/packages/meteor/async_helpers.js#L54-L61)

packages/ddp-server/stream_server.js
```diff
@@ -14,12 +15,18 @@ import once from 'lodash.once';
...
-    ));
+    extensions.push(Npm.require('permessage-deflate2').configure({
+      threshold: 512,
+      level: zlib.constants.Z_BEST_SPEED,
+      memLevel: zlib.constants.Z_MIN_MEMLEVEL,
+      noContextTakeover: true,
+      maxWindowBits: zlib.constants.Z_MIN_WINDOWBITS,
+      ...(websocketCompressionConfig || {})
+    }));
   }
```
Grubba27: Are those configs documented? this whole config could become even a guide for increasing performance

packages/ddp-server/package.js
```diff
@@ -5,12 +5,13 @@ Package.describe({
 });
 
 Npm.depends({
-  "permessage-deflate": "0.1.7",
+  "permessage-deflate2": "0.1.8-alpha.1",
   sockjs: "0.3.24",
```
Grubba27, line 9: How hard would it be to change from sockjs? Do we have testing covering the usage?

packages/meteor/asynchronous_queue.js
```diff
@@ -0,0 +1,154 @@
+class AsynchronousQueue {
+  /**
+   * Creates a queue that processes tasks in parallel batches while preserving completion order
+   * when needed. Configurable batch size and concurrency limits help optimize throughput.
+   */
+  constructor({
+                batchSize = 50,
+                maxConcurrent = 10,
+                orderMatters = false
+              } = {}) {
```
Grubba27: All usages of  `Meteor._AsynchronousQueue` are with the default params; maybe those options could become env variables?

packages/ddp-server/package.js
leonardoventurini: It should involve some work, socket.io is a strong candidate, however I'm not sure there are direct tests for it

packages/ddp-server/stream_server.js
leonardoventurini: I don't think there are docs for it, I can add them

packages/mongo/observe_multiplex.ts
leonardoventurini: No need to be async, adjusted that

packages/meteor/asynchronous_queue.js
leonardoventurini: We can't use TypeScript inside this package easily, not entirely sure why

packages/ddp-server/stream_server.js
leonardoventurini: Added them https://github.com/meteor/meteor/pull/13445/commits/c73515470f5a4a623b8403231fbfa6c8472cd5b1

packages/mongo/observe_handle.ts
```diff
@@ -4,24 +4,30 @@ let nextObserveHandleId = 1;
...
+  public initialAddsSentResolver: (value: void) => void = () => {};
+  public initialAddsSent: Promise<void>
 
-  constructor(multiplexer: any, callbacks: Record<ObserveHandleCallback, any>, nonMutatingCallbacks: boolean) {
+  _added?: Callback<T>;
+  _addedBefore?: Callback<T>;
+  _changed?: Callback<T>;
+  _movedBefore?: Callback<T>;
+  _removed?: Callback<T>;
+
```
Grubba27, line 29: Should not be `T[]`? These methods can send a list of args to the callback https://docs.meteor.com/api/collections#callbacks-1

nachocodoner (approved):
Great work 🚀

Issue #13430, opened 2024-10-25
HMR Slowness with Meteor 3.0.3 + React + MUI6


bratelefant:
**Issue Summary:**  
Hot Module Replacement (HMR) in Meteor 3.0.3 has significant delays compared to version 2.16. With 2.16, HMR updates are almost instant (<1 second), but with 3.0.3, these updates take approximately 7 seconds. Profiling using `METEOR_PROFILE=20 meteor run` shows similar build times, but it takes significantly more time between the step "prepareProjectForBuild" and "Profiling: Rebuild App".

**Steps to Reproduce:**
1. Clone my fork of @Ritinha999's repo: [https://github.com/bratelefant/simple-app](https://github.com/bratelefant/simple-app)
2. Run the app using `meteor run`.
3. Make a minor change, such as adding a character in `Dummy.jsx`, and observe the HMR response time. This is as quick as it should be (it's on 2.16)
4. Run `meteor update` to update to any 3.0 version
5. Repeat steps 2.+3. Now changes take ~7 seconds to be visible in the browser.

**Observed Results:**
- **Meteor 2.16**: HMR response <1 second.
- **Meteor 3.0.3**: HMR response ~7 seconds.

**Environment:**
- React + MUI 6
- Using Meteor 3.0.3 (Issue not present in Meteor 2.16)

**Additional Notes:**
- Profiling logs indicate that additional processes introduced with Meteor 3.0 may contribute to this delay.
- We have [a forum post](https://forums.meteor.com/t/super-slow-hmr-in-meteor-3-0) on this issue.
- Some users suggest, that HTTP telemetry requests reporting package usage might contribute to this behavior.

9Morello:
Does that slowness persists if you run the project with the [Meteor 3.1.0 RC0](https://github.com/meteor/meteor/pull/13358)? I'm getting pretty fast hot reload with the Release Candidate build.

nachocodoner:
In this PR we already identified the issues with HMR and fix them, https://github.com/meteor/meteor/pull/13448.

As in the comment above, as part of [Meteor 3.1.0-RC0](https://github.com/meteor/meteor/pull/13358) we included that fix.

To try it and confirm please do:

```
meteor update --release 3.1-rc.0
```

nachocodoner:
[...]

PR #13448, opened 2024-11-05
Fix performance issue with bundler ensuring proper parallelization

nachocodoner:
OSS-586

Context: https://github.com/meteor/meteor/issues/13430 and  [forum post](https://forums.meteor.com/t/super-slow-hmr-in-meteor-3-0)

This PR investigates the performance impact on app builds/rebuilds in Meteor 3.0.

## Issue

Some parts of the code lost parallelism when adapted to use fibers, leading to performance issues in async operations.

I added `console.time` and `console.timeEnd` to detect where timing mismatches occurred in the Meteor build process, especially compared to Meteor 2. The next picture shows that makeClientTarget introduced a significant delay. I debugged further, identified the problematic code section, and fixed it.

image

## Solution

In several parts of the code, Promise.all could be used to speed things up. For this specific issue, I identified and fixed the slow spot causing the problem.

image

This issue could generally improve the build process, and with the latest Node's speed, we'll start seeing real gains.

This pattern could be applied more broadly, but we should focus on areas with notable gains. Additionally, adding benchmark monitoring for Meteor builds could help catch and resolve future regressions, the same way we have in [meteor/performance](https://github.com/meteor/performance) for Meteor runtime.

## Pending

- [x] Review package usage recording under poor network conditions

- [x] Test with the reported app as I used a scenario using a small example where was enough and quick to find and work out the regression.

### recordPackageUsage

https://github.com/meteor/meteor/pull/13448#issuecomment-2459752591

### refreshClient

Don't need to await as in Meteor 2, https://github.com/meteor/meteor/blob/2.x.x/tools/runners/run-app.js#L911

image

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -1223,13 +1223,16 @@ class Target {
           continue;
         }
 
+        var data = await resource.data;
+        var hash = await resource.hash;
+
         const fileOptions = {
           info: 'unbuild ' + resource,
           arch: this.arch,
-          data: await resource.data,
+          data,
           cacheable: false,
-          hash: await resource.hash,
-          skipSri: !!await resource.hash
+          hash,
+          skipSri: !!hash
         };
 
         const file = new File(fileOptions);
@@ -1290,15 +1293,17 @@ class Target {
             continue;
           }
 
+          var data = await resource.data;
+          var hash = await resource.hash;
           let sourcePath;
-          if ((await resource.data) && resource.sourceRoot && resource.sourcePath) {
+          if (data && resource.sourceRoot && resource.sourcePath) {
             sourcePath = files.pathJoin(resource.sourceRoot, resource.sourcePath);
           }
           const f = new File({
             info: 'resource ' + resource.servePath,
             arch: this.arch,
-            data: await resource.data,
-            hash: await resource.hash,
+            data,
+            hash,
             cacheable: false,
             replaceable: resource.type === 'js' && sourceBatch.hmrAvailable,
             sourcePath
@@ -1323,14 +1328,15 @@ class Target {
             });
           }
 
+          var sourceMap = await resource.sourceMap;
           // Both CSS and JS files can have source maps
-          if (await resource.sourceMap) {
+          if (sourceMap) {
             // XXX we used to set sourceMapRoot to
             // files.pathDirname(relPath) but it's unclear why.  With the
             // currently generated source map file names, it works without it
             // and doesn't work well with it... maybe?  we were getting
             // 'packages/packages/foo/bar.js'
-            f.setSourceMap(await resource.sourceMap, null);
+            f.setSourceMap(sourceMap, null);
           }
 
           this[resource.type].push(f);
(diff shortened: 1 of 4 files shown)


tools/isobuild/bundler.js
```diff
@@ -1223,13 +1223,16 @@ class Target {
...
         const fileOptions = {
           info: 'unbuild ' + resource,
           arch: this.arch,
-          data: await resource.data,
+          data,
           cacheable: false,
-          hash: await resource.hash,
-          skipSri: !!await resource.hash
+          hash,
+          skipSri: !!hash
```
nachocodoner, line 1235: This hasn’t had an impact, as `resource.data` and `resource.hash` are never promises, at least based on my debugging. I'm unsure why we are using `await` here, possibly due to an edge case I haven’t identified. I refactored the code to reduce the number of await calls for the same variables.

tools/isobuild/compiler-plugin.js
```diff
@@ -1720,21 +1720,21 @@ export class PackageSourceBatch {
...
+            absModuleId: inputFile.absModuleId,
+            sourceMap: !!(await inputFile.sourceMap),
+            mainModule: inputFile.mainModule,
+            imported: inputFile.imported,
+            alias: inputFile.alias,
+            lazy: inputFile.lazy,
+            bare: inputFile.bare,
+          };
+        })
+      )
```
nachocodoner, line 1737: This is actually the pattern that enables parallelization on parsing the code from files and speeds up the build.

minhna:
I think there is still a wait for making 2 requests to track packages usage.
Could we make them run in background instead of waiting them for finished?

Somewhere here: https://github.com/meteor/meteor/blob/devel/tools/runners/run-app.js#L576
```
if (self.recordPackageUsage) {
        // Maybe this doesn't need to be awaited for?
        await stats.recordPackages({
          what: "sdk.run",
          projectContext: self.projectContext
        });
      }
```

nachocodoner:
> I think there is still a wait for making 2 requests to track packages usage. Could we make them run in background instead of waiting them for finished?
> 
> Somewhere here: [`devel`/tools/runners/run-app.js#L576](https://github.com/meteor/meteor/blob/devel/tools/runners/run-app.js?rgh-link-date=2024-11-06T03%3A07%3A38Z#L576)
> 
> ```
> if (self.recordPackageUsage) {
>         // Maybe this doesn't need to be awaited for?
>         await stats.recordPackages({
>           what: "sdk.run",
>           projectContext: self.projectContext
>         });
>       }
> ```

Definitely something to consider. For now, the focus has been on the build tool, and it's been useful to identify slowdowns between Meteor 3 and 2. Package usage recording has not provided any bad signal on my end. However, under poor network conditions, the issue you mentioned might be contributing to the slowdown. I'll review this and provide a fix.

Thanks for your input 🚀

nachocodoner:
@minhna: This change definitely impacts every app build.

I moved this process to the background, which has saved us time. Both my fix and yours caused slowdowns even more as your app is larger, with more packages, also this depending on your machine's power and network status. This fix is crucial for significantly improving build speed, especially since the latest Node version also outperforms Node 14 in Meteor 2, and finally without these delays we will feel that.

## Issue

image

## Solution

image

denihs (approved):

PR #13358, opened 2024-09-17
Release 3.1.0

denihs:
# Still in development
Current released version: **3.1**.

To use it, please do a

```
meteor update --release 3.1
```

you can also use the following command to install this meteor version

```
npx meteor@rc
```

# Highlights
* `webapp`, `test-helpers`, `boilerplate-generator-tests` underscore removed by @harryadel

PR #13455, opened 2024-11-11
Fix Cordova build when using custom plugins using protocols

nachocodoner:
OSS-566

Context:  https://github.com/meteor/meteor/issues/13303, https://github.com/meteor/meteor/pull/13358#issuecomment-2465717148

Continues: https://github.com/meteor/meteor/pull/13416

This PR solves an edge case not covered on the fix that prepared npm packages for cordova build, supporting indirect dependencies. With the changes on this PR it will also support cordova dependencies that you describe using the common protocols, `git+https://`, `git+ssh://`, `git://` and `file://`.

The fix will use the version described by on cordova file if exists, otherwise the automatically resolved and used as part of the package-lock.json on prepare the npm dependencies for cordova.

After building Cordova app, the `.meteor/local/cordova-build/package.json` resolve the dependencies properly as shown here:

image

### DIFF
--- a/tools/cordova/project.js
+++ b/tools/cordova/project.js
@@ -463,6 +463,8 @@ to Cordova project`, async () => {
       let platformSpec = version ? `${platform}@${version}` : platform;
       await cordova_lib.platform('add', platformSpec, this.defaultOptions);
 
+      const installedPlugins = this.listInstalledPluginVersions();
+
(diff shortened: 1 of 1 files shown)


harryadel:
Can you publish a new beta with this fix, so we could try it?

PR #13456, opened 2024-11-11
docs: changelog

leonardoventurini:


v3-docs/docs/generators/changelog/versions/3.1.0.md
```diff
@@ -22,12 +33,56 @@ meteor update --release 3.1.0
 
 ```
 
+Make sure to migrate userland code to use the new Express APIs:
 
+https://expressjs.com/en/guide/migrating-5.html
```
nachocodoner, line 40: It's also worth noting that the Mongo driver introduced a major breaking change. Meteor apps using rawCollection or rawDatabase directly can no longer use callbacks for operations. They must switch to using promises instead.

denihs (approved):

Issue #13450, opened 2024-11-07
Cordova Android start failed with Uncaught Error because of adding accounts-base


joyjiyuan:
Meteor: 3.0.4
Cordova Android: 13.0.0
OS: both windows 11 and macOS 15
Emulator: Pixel 8a API 34

### The expected behavior
Run on android emulator or device
### The actual behavior.
Start failed after add accounts-base. white screen
### reproduce
1. create meteor app from skelton
2. meteor add accounts-base
3. meteor add-platform android
4. meteor run android

"Uncaught Error: The core-js npm package could not be found in your node_modules
...
Uncaught TypeError: Cannot read properties of undefined (reading 'Accounts')"

and run meteor npm install --save core-js doesn't fixed it

I20241108-04:27:19.445(9)? 11-07 19:27:05.028  9188  9188 I CordovaLog: Changing log level to DEBUG(3)
I20241108-04:27:19.460(9)? 11-07 19:27:06.501  9188  9320 W chromium: [WARNING:dns_config_service_android.cc(115)] Failed to read DnsConfig.
I20241108-04:27:19.461(9)? 11-07 19:27:07.824  9188  9188 W MeteorWebApp: Loading asset bundle from directory file:///android_asset/www/application
I20241108-04:27:19.461(9)? 11-07 19:27:08.453  9188  9320 W MeteorWebApp: Asset / found in bundle 5475bb95b63169f6f68ab99e9362c96538c39475:file:///android_asset/www/application
I20241108-04:27:19.461(9)? 11-07 19:27:08.734  9188  9319 W MeteorWebApp: Asset /__cordova/merged-stylesheets.css found in bundle 5475bb95b63169f6f68ab99e9362c96538c39475:file:///android_asset/www/application
I20241108-04:27:19.461(9)? 11-07 19:27:08.987  9188  9319 W MeteorWebApp: Asset /cordova.js not found in bundle 5475bb95b63169f6f68ab99e9362c96538c39475:file:///android_asset/www/application, no parent bundle
I20241108-04:27:19.461(9)? 11-07 19:27:09.114  9188  9319 W MeteorWebApp: Asset /cordova_plugins.js not found in bundle 5475bb95b63169f6f68ab99e9362c96538c39475:file:///android_asset/www/application, no parent bundle
[...]

PR #13459, opened 2024-11-13
Fix Cordova issue when using accounts-base

nachocodoner:
OSS-616

Context: https://github.com/meteor/meteor/issues/13450 and [forum post](https://forums.meteor.com/t/meteor-meteor-3-0-4-meteor-run-ios-throws-several-errors-and-warnings/62514)

## Issue

This PR addresses an issue that started to happen on early Meteor 3 versions (including 3.0.1 launch) and caused from [updating "core-js" packages](https://github.com/meteor/meteor/commit/79551fb53b2f48ac1f42405f2893b8f9d5553054) to the latest version. The Meteor `url` package, used by `accounts-base`, relies on `core-js` to maintain compatibility with legacy builds, which are essential for running the app in Cordova webviews. Updating `core-js` required handling breaking changes like the one here, and we currently lack regression tests to verify Cordova builds with the accounts package.

## Solution

To resolve the issue with newer versions, I added `url-search-params-polyfill` for legacy builds, as latest `core-js` no longer supports `URLSearchParams` and assumes it’s available, but only does for modern bundles.

## Conclusion

In future reviews, we should be particularly cautious with version bumps, especially with minor or major updates, to find impacts on related packages, specially in those we don't have any kind of coverage or result hard to achieve (like legacy bundles).

### DIFF
--- a/packages/url/legacy.js
+++ b/packages/url/legacy.js
@@ -4,7 +4,7 @@ try {
   exports.URLSearchParams = URLSearchParams;
 
   require("core-js/proposals/url");
-  URL = window.URLSearchParams;
+  URL = window.URL;
   exports.URL = URL;
 } catch (e) {
   console.log(' error', e);
--- a/packages/url/legacy.js
+++ b/packages/url/legacy.js
@@ -1,6 +1,13 @@
 try {
+  require("url-search-params-polyfill");
+  URLSearchParams = window.URLSearchParams;
+  exports.URLSearchParams = URLSearchParams;
+
   require("core-js/proposals/url");
+  URL = window.URLSearchParams;
+  exports.URL = URL;
 } catch (e) {
+  console.log(' error', e);
   throw new Error([
     "The core-js npm package could not be found in your node_modules ",
     "directory. Please run the following command to install it:",
--- a/packages/url/legacy.js
+++ b/packages/url/legacy.js
@@ -7,7 +7,6 @@ try {
   URL = window.URL;
   exports.URL = URL;
 } catch (e) {
-  console.log(' error', e);
   throw new Error([
     "The core-js npm package could not be found in your node_modules ",
     "directory. Please run the following command to install it:",
--- a/packages/url/package.js
+++ b/packages/url/package.js
@@ -6,7 +6,8 @@ Package.describe({
 });
 
 Npm.depends({
-  "core-js": "3.38.1"
+  "core-js": "3.39.0",
+  "url-search-params-polyfill": "8.2.5",
 });
 
 Package.onUse(function(api) {


wreiske:
I pulled down the changes. I haven't seen any issues so far, nevertheless I would hate to be the one that says "All Good!" and have this break everyone's apps! 

I'm going to keep testing this PR today in dev and might push the changes out to production tonight to see how tomorrow goes.

wreiske:
So far so good. I'm pushing out to production for sure tonight. I'll let you know 🙏

nachocodoner:
I've tested this myself and also commented on the original issue. It's always helpful to have another set of eyes reviewing.

Thanks so much!

Grubba27 (changes requested):
Is there a way to add testing to this?

Grubba27 (approved):
Testing this kind of regression does not seem to be trivial(I do not have any knowledge of mobile)

nachocodoner:
Yes, I’ve used tools like [Appium](https://appium.io/docs/en/latest/) before. Setting up the infrastructure for effective testing pipelines took some time. Adding something similar could be useful, especially as we consider future native Meteor apps with Capacitor or a similar tool.

wreiske:
Ran it all day and didn't have any reports of issues. 🎉 . I guess worst caste scenario is we might have to create a hotfix and push a new `url` package out if someone does find issues.

zodern:
> To resolve the issue with newer versions, I added url-search-params-polyfill for legacy builds, as latest core-js no longer supports URLSearchParams and assumes it’s available, but only does for modern bundles.

I was surprised that core-js no longer polyfills URLSearchParams, so I looked into it. core-js does still have the polyfill, and the import Meteor is using (`core-js/proposals/url`) is one of the ways to load it. It's the [first polyfill loaded by the import](https://github.com/zloirock/core-js/blob/a1298dcf04a231abe7ef6b0021c04cb06db8a9b2/packages/core-js/web/url.js#L2).

The polyfill's for URLSearchParams aren't that small, so it would be nice if this could be fixed without loading two of them in the legacy client.

nachocodoner:
Yes, that's what I initially thought, @zodern. However, no matter what I load from `core-js`, the error still occurs. `URLSearchParams` isn't loaded at all, and other crashes happen as well when I try to load `URLSearchParams` by demand from `core-js` modules as the one you pointed and set it globally. It might be an issue with the library itself, I could report it. 
I can try again on my side to double-check, but I already tested several options with no solution. Other option to perserve one dependency is to downgrade the `core-js` package, but this is not a good long-term solution either.

As for size, I checked how much `url-search-params-polyfill` takes, and it doesn’t seem too large. https://bundlephobia.com/package/url-search-params-polyfill@8.2.5 I understand to keep the meteor core as lean as possible, but I was thinking this was required after latest `core-js` upgrades, with no option to see `URLSearchParams` available.

image

nachocodoner:
image

Just for you to check what I did. Since the error behind the upgrade is always `TypeError: undefined is not an object (evaluating '$URLSearchParams.prototype')`, I trying to load by demand the `core-js/web/url-search-params` polyfill available from `core-js`. However, it crashes the same way, since that code imports other required parts `web.url-search-params.delete.js` that uses `URLSearchParams`. I might ask for help on core-js repository itself, but this seems an issue on their end.

The only way I found to overcome this is to import an alternative `url-search-params` polyfill that overcomes the core-js limitation.

nachocodoner:
I've opened a new issue on `core-js` to investigate a possible regression: https://github.com/zloirock/core-js/issues/1380.

Regarding this PR, we will proceed with merging it to address critical issues causing crashes in Cordova apps. We've received many reports about this, and developers have already helped validate the fix. It will be available on release 3.1.

We'll revisit this in the future once we get any confirmation or a fix from `core-js`.

zodern:
@nachocodoner I agree with the response in the core-js issue that this is an issue in Meteor. I looked into it a little bit more, and it's a problem caused by Meteor's package scoping. It's creating a package-scoped `URLSearchParams` variable, which is still undefined by the time `core-js/modules/web.url-search-params.delete.js` runs. I don't remember if Meteor scans files in node_modules when identifying global variables to add to the package scope. Assuming it doesn't, you should be able to fix this by removing the URLSearchParams pseudo global from the package code.

PR #13462, opened 2024-11-14
Check V1.5.0 | New methods checkAsync & MatchAsync

mvogttech:
**Summary of Changes**
This pull request introduces new asynchronous validation methods to enhance the flexibility and performance of data validation in Meteor. It also includes general performance improvements, such as converting arrays to Set objects for faster lookups.

**New Asynchronous Methods**

- Added checkAsync Function
- An asynchronous version of the existing check function.
- Allows for validations that require asynchronous operations using async/await.

**Updated Match Namespace with Async Methods**

- - MatchAsync.testAsync(value, pattern)
- - Match._failIfArgumentsAreNotAllCheckedAsync(f, context, args, description)

**Refactored Codebase for Asynchronous Support**

- Updated Internal Functions to Support async/await
- - Refactored testSubtree function to be asynchronous.
- - Ensured all recursive calls within testSubtree are awaited.
- - Adjusted pattern matching logic to handle asynchronous operations.
- Enhanced Error Handling in Asynchronous Contexts
- - Wrapped all asynchronous operations in try...catch blocks.
- - Prevented unhandled promise rejections.
- - Ensured that errors are properly caught and reported.

**Performance Improvements**

- Converted Arrays to Set for Faster Lookups
- Replaced _jsKeywords array with a Set object.

Before:
`const _jsKeywords = ['do', 'if', 'in', 'for', ...];`

After:
`const _jsKeywords = new Set(['do', 'if', 'in', 'for', ...]);`


jamauro:
Thanks for putting this together. One question: is the `MatchAsync` namespace necessary? Seems like we could get away with `checkAsync` and `Match.testAsync`. It's been a while since I looked at this code though. Let me know what you think.

mvogttech:
> Thanks for putting this together. One question: is the `MatchAsync` namespace necessary? Seems like we could get away with `checkAsync` and `Match.testAsync`. It's been a while since I looked at this code though. Let me know what you think.

You're right, that would be the better design pattern. I was trying to go along with the *Async suffix Meteor has been moving to in V3.

I'll remove MatchAsync.

radekmie:
I'd like to raise a small concern regarding the scope of this PR. The only functionality added here is a support for `async` functions in `Match.Where`. It's a really nice feature, and it surely does require some sort of a `checkAsync`/`testAsync` to work.

However, this code change is _huge_ and consists of mostly formatting changes (e.g., replacing a ternary operator with `if`, or adding semicolons) and copying JSDocs from the `.d.ts` file back to the code. The performance improvements seem negligible, as the `_prependPath` function is used only when an error is thrown, which is rather uncommon in practice.

I'd propose to split it into 3 PRs instead:
1. **Formatting.** This one is fully optional and should be automated with ESLint or similar tools instead.
2. **TypeScript.** Instead of copying the JSDocs to the JS code, this package could get a TypeScript rewrite, making it easier to work with and eliminating the issue with desynchronized `.d.ts` files.
3. **Support for `async` functions in `Match.Where`.** This should include a check for incorrect usage, just like other libraries do (e.g., [`zod`](https://github.com/colinhacks/zod) throws an error when an async "schema" is validated using the synchronous call). Most importantly, it should be done in such a way, that the least amount of code is duplicated, so this small utility package remains understandable and maintainable in the future.

This way more people will be able to review the changes and provide feedback.

jamauro:
Hey y'all, checking in on the status of this PR. Would be great to see this functionality added.

IMO, converting to TS shouldn't hold up this PR. If deemed valuable, it could be pursued separately by someone else.

nachocodoner:
> Support for async functions in Match.Where. This should include a check for incorrect usage, just like other libraries do (e.g., [zod](https://github.com/colinhacks/zod?rgh-link-date=2024-11-17T10%3A13%3A53Z) throws an error when an async "schema" is validated using the synchronous call). Most importantly, it should be done in such a way, that the least amount of code is duplicated, so this small utility package remains understandable and maintainable in the future.

Yeah, it would be good to have. Though I see test coverage, TypeScript, and documentation as really valuable, and this has already been an effort that we could revisit for a next release. Also, it's true that we already have proper formatting going on [in that PR](https://github.com/meteor/meteor/pull/14327), and it will drastically conflict with this one. In terms of test coverage, there are tests here, but also there.

I'd at least suggest creating another PR for support of async functions in Match. For test coverage and the rest, maybe @italojs can review whether both PRs cover the same.

italojs:
I would consider reopening this PR since it's a new feature, different from [oxc PR](https://github.com/meteor/meteor/pull/14327), which is a refactor

BTW, I removed the check from OXC PR to avoid conflicts with this and other PRs. The initial phases for the linter/fmt changes should not generate any conflict or headache with other PRs

@mvogttech please reopen it if you still want to work on it ;)

PR #13464, opened 2024-11-14
Creating doc for the Roles package

denihs:
Preview: https://deploy-preview-13464--v3-meteor-api-docs.netlify.app/api/roles.html

image


nachocodoner (dismissed):

StorytellerCZ:
Missing the sync functions for client. Will try to find some time on a train to look into this.

nachocodoner:
@StorytellerCZ: How is the progress on adding the missing docs?  

We’d like to release the docs for roles soon.

Grubba27:
I think the docs are good enough(I expect most people always to use the async API); having _some_ docs is better than none.

@StorytellerCZ, when you have the time to add the sync client, I think you will need to do something similar as Denilson did in `packages/roles/roles_common_async.js` and update the comments so that the jsdocs can pick it up.

I've moved it to the packages bc you need to install this in your app, so it is not in the common `api`

v3-docs/docs/.vitepress/config.mts
Grubba27, line 1: the only actual change is:


```js
{
            text: "roles",
            link: "/packages/roles",
},
```

Issue #13444, opened 2024-11-04
`Mongo.Collection` calling wrong allow/deny methods


Julusian:
Using meteor 3.0.4

The allow/deny methods are not behaving in a logical fashion.

Here are a few scenarios:

## Only `update` validator
```
testCollection.allow({
  update: (userId, doc, fields, modifier) => {
    console.log('update', userId, doc, fields, modifier)
    return true
  } 
})
```
On the client calling `testCollection.update(...)` will fail with an `access denied` 403 error, without anything being logged.
On the client calling `testCollection.updateAsync(...)` will fail with an `No allow validators` 403 error, without anything being logged.

## Only `updateAsync` validator
```
testCollection.allow({
  updateAsync: async (userId, doc, fields, modifier) => {
    console.log('update', userId, doc, fields, modifier)
    return true
  } 
})
```
On the client calling `testCollection.update(...)` will fail with an `No allow validators`  403 error, without anything being logged.
On the client calling `testCollection.updateAsync(...)` will succeed, with a log line.
Typescript complains because `updateAsync` is not a valid method to provide.

## Both `update` and `updateAsync` validators
```
testCollection.allow({
  update: (userId, doc, fields, modifier) => {
    console.log('fail', userId, doc, fields, modifier)
    return false
  },
  updateAsync: async (userId, doc, fields, modifier) => {
    console.log('updateAsync called', userId, doc, fields, modifier)
    return true
  } 
})
```
On the client calling `testCollection.update(...)` will succeed, logging `updateAsync called`.
On the client calling `testCollection.updateAsync(...)` will succeed, logging `updateAsync called`.
Typescript complains because `updateAsync` is not a valid method to provide, and `update` isn't allowed to return a promise.

## Conclusions

1) It appears that a ddp call to `update` a collection will check for allow methods called `update`, but then call the `updateAsync` ones.
1) ddp calls to `updateAsync` will check for and execute `updateAsync` allow methods
1) According to typescript `updateAsync` is not a valid method to pass into allow/deny methods, neither is them returning a promise.

I haven't tested insert or remove, I would assume they suffer the same issue.

As a side thought, I would also like to question whether `update` and `updateAsync` should be separate ddp calls. It feels to me like the client should be issuing just `update` calls, I don't see why they would have different implementations as to the client all ddp calls are 'async'. 

I am not 100% sure what the correct solution here is, should the `updateAsync` allow method exist and be called, or should it be calling the `update` method even for the ddp `updateAsync`?

[...]

PR #13468, opened 2024-11-19
Update Mongo.Collection types to allow async allow-deny rules

9Morello:
This minor PR adds a few missing properties to the type of the parameter you can pass to `someCollection.allow`. The code already accepts async rules there, but since the type definitions don't include them, TypeScript complains. 

This issue was reported in #13444 by @Julusian.

### DIFF
--- a/packages/mongo/mongo.d.ts
+++ b/packages/mongo/mongo.d.ts
@@ -98,6 +98,9 @@ export namespace Mongo {
       insert?:
         | ((userId: string, doc: DispatchTransform<Fn, T, U>) => boolean)
         | undefined;
+      insertAsync?:
+        | ((userId: string, doc: DispatchTransform<Fn, T, U>) => Promise<boolean>|boolean)
+        | undefined;
       update?:
         | ((
             userId: string,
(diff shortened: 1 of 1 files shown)


Julusian:
@9Morello thanks for looking at this!

While this makes typescript happy, it doesn't address the underlying problem of weird behaviour.
I suspect the full solution to #13444 is to strip off the `Async` suffix when invoking the methods, which would negate the need for this change.

9Morello:
Hi @Julusian :D
I don't disagree with you per se - this PR is more to avoid situations where the IDE says the user code is wrong, despite it being correct, because the type definitions are not up to date.
Those errors when authoring code can be very confusing to new users who aren't familiar with Meteor, and may not realize that using an `updateAsync` rule does work, despite their IDE telling them otherwise.

nachocodoner:
Based on the feedback, we deprecated *Async rules for allow/deny in this PR, https://github.com/meteor/meteor/pull/13499. They haven't been fully removed yet to give others time to migrate and avoid affecting the next patch version, but they will be removed in the future.

We also fixed some issues and enabled common rules configuration (insert, update, remove) to return promises. [Types were updated as well](https://github.com/meteor/meteor/pull/13484/commits/080f270d6bff9ec6e249b49931f5b22e8aea7137). I'll close this PR now. Thank you for bringing this to our attention. 🙏

The beta for release 3.1.1 is now available for testing. Your feedback is welcome:

``` sh
meteor update --release 3.1.1-beta.1
```

PR #13477, opened 2024-11-20
docs: Fix links and add leonardoventurini:scss as recommended

fredmaiaarantes:
Fix CLI and ECMAScript links and add `leonardoventurini:scss` as the recommended SCSS package.


Grubba27 (approved):

StorytellerCZ:
@leonardoventurini do you have a PR to the original SCSS package that we could merge and release, so that people don't have to change packages (which many won't do as they will be unaware of this)?

PR #13497, opened 2024-11-28
feat: ddp batching

leonardoventurini:
PoC for DDP batching, the results seem very promising but we are facing a blocking issue:

Transport is dropping larger messages and thus the client is not receiving it. There is no warning or error, messages are simply dropped if we batch just 10 or so adds. It also seems to be random, because the first initial adds are successfully sent containing around 100 adds making subs perceptibly more responsive, but once it happens, no further `addedBatch` is received in the client.

(╯°□°)╯︵ ┻━┻


packages/ddp-server/livedata_server.js
```diff
@@ -191,11 +193,49 @@ Object.assign(Session.prototype, {
...
+      }
+    };
+
+    if (batch.docs.length >= 20) {
+      sendBatch();
+    } else {
+      batch.timeoutHandle = setTimeout(() => {
+        console.log('send batch timeout', this._batchedAdds.size)
+        sendBatch();
+      }, 5000);
```
Grubba27, line 233: Is `sendBatch();` correct here?

packages/ddp-server/livedata_server.js
```diff
@@ -191,11 +193,49 @@ Object.assign(Session.prototype, {
...
+    }
+  },
+
+  sendAddedBatch(collectionName, docs) {
+    this.send({ msg: 'addedBatch', collection: collectionName, docs });
   },
 
   sendChanged(collectionName, id, fields) {
```
Grubba27, line 241: Maybe it needs to change `sendChanged` to be batched too?

leonardoventurini: The testing scenario produced no changed messages, and seems to be tied to the message sized, we tested many scenarios, with/without async, and only occurred when the batch was a certain size 😞

packages/ddp-server/livedata_server.js
leonardoventurini: Yep

radekmie:
Two cents:
1. Instead of adding new DDP operations (`addedBatch`), it'd be better to be able to batch all operations. Type-wise, DDP currently operates on `type Operation = Added | ...`, and instead of adding more possibilities, I think it'd be better to send `Operation[]` instead.
    * This makes the client-side _slightly_ more backward-compatible, as all hooks on `_process_added` (and family) still work.
    * It's also easier to maintain correct order of operations, as with `addedBatch`, the batch _should_ be flushed on any non-`added` operation to keep the current ordering.
2. > Transport is dropping larger messages and thus the client is not receiving it.

    It cannot be the transport itself, as you can send megabytes of data via DDP and it works.

KoenLav:
It might be worthwhile to look at the DDP batching part of this PR:

https://github.com/meteor/meteor/pull/10478/files

It basically did what @radekmie suggested, keeping the old handlers and backwards compatibility.

PR #13506, opened 2024-12-05
Fix flaky roles tests by comparing expected array values equality regardless the order

nachocodoner:
OSS-632

This PR addresses flaky test behavior in the roles package.

## Issue

I reproduced at least [two intermittently failing tests](https://github.com/user-attachments/assets/311d9d86-3cec-48c7-ba40-2486ed17f58a).

The problem arises in some test scenarios where the roles config and expected values differ in element order, even though the elements themselves are identical. The example below illustrates this issue.

image

## Solution

I investigated the cause of this ordering flakiness but struggled to find it within the package, which I am still learning. Since the elements contents are identical and only their order varies across runs, I created a new test verifier to compare the contents without considering the order. This resolved the test's flakiness.

As I am not very familiar with this package, I’d appreciate feedback from @jankapunkt or @StorytellerCZ. My assumption is that the element order doesn’t matter in these tests, as long as the roles config for each item matches correctly.

### DIFF
--- a/packages/roles/tests/serverAsync.js
+++ b/packages/roles/tests/serverAsync.js
@@ -84,7 +84,7 @@ const sameDeepMembers = (test, value, expected) => {
   );
 };
 
-function sameDeepUnorderedMembers(test, value, expected) {
+const sameDeepUnorderedMembers = (test, value, expected) => {
   const sortAndStringify = (arr) => {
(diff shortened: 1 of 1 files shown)


Grubba27 (dismissed):
I very much prefer checking arrays for values not ordering

StorytellerCZ (approved):

StorytellerCZ:
Yes, order should not matter.

PR #13508, opened 2024-12-06
FIX: Add missing types to roles package

Grubba27:
it was missing an `addAssets` in the `package.js` file

### DIFF
--- a/packages/roles/package.js
+++ b/packages/roles/package.js
@@ -15,8 +15,6 @@ Package.onUse(function (api) {
     both
   );
 
-  api.use("zodern:types@1.0.13");
-
   api.use(["blaze@2.9.0 || 3.0.0"], "client", { weak: true });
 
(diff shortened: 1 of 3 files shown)


zodern:
Since the package already has a dependency on `zodern:types`, it doesn't need the `addAssets` call. `zodern:types` already ensures the .d.ts files are included when the package is published.

The only reason the other packages need it is because they didn't have a dependency on `zodern:types`. With how zodern:types currently works, doing both can break the types when the package doesn't have a `package-types.json`, since `zodern:types` isn't sure which `.d.ts` file to use.

packages/roles/package.js
StorytellerCZ, line 1: Adding via asset, can't we then remove `zodern:types`?

StorytellerCZ (changes requested):

Grubba27:
Found the issue, it was missing the export from the file

nachocodoner (approved):

PR #13510, opened 2024-12-06
Upgrade the dependencies of `@meteorjs/babel`

9Morello:
This PR does the following:

* Updates the versions of `json5` and `semver` that the `babel-compiler` version depended on. This was causing some auditing tools to complain because of [known CVEs](https://nvd.nist.gov/vuln/detail/CVE-2022-25883) related to the old versions used. I don't think there was any vector for exploiting those as they're only used internally by that package on values that aren't provided by end users, but they got upgraded nonetheless. 
* Upgrade the dependencies of `@meteorjs/babel`, with a minor adjustment to the test suite. Specifically, upgrading `@babel/traverse` addresses [CVE-2023-45133](https://babel.dev/blog/2023/10/16/cve-2023-45133). As described in the article, actually exploiting this is very hard, but nonetheless, it can be an issue with automated scans running on your codebase.
* Update the versions for `@babel/parser` and `@babel/runtime` inside some dev bundler scripts.


StorytellerCZ (approved):
Just need to fix tests and it is good to go by me.

9Morello:
@StorytellerCZ I fixed the tests (some of them were still using Node.js 14), but it seems like some Windows tests are not running. Any ideas? 🤔

StorytellerCZ:
I think @nachocodoner knows the most about these. Maybe even @leonardoventurini

filipenevola:
@9Morello please start with this bump: `@meteorjs/babel` to the next patch so we can approve it first and then use the latest here on the Meteor package.

Ok?

npm-packages/meteor-babel/package.json
```diff
@@ -31,17 +31,16 @@
...
-    "@babel/plugin-transform-runtime": "^7.17.0",
-    "@babel/preset-react": "^7.16.7",
-    "@babel/runtime": "7.17.2",
-    "@babel/template": "^7.16.7",
-    "@babel/traverse": "^7.17.0",
-    "@babel/types": "^7.17.0",
+    "@babel/core": "^7.26.0",
+    "@babel/parser": "^7.26.3",
+    "@babel/plugin-transform-class-properties": "7.25.9",
+    "@babel/plugin-transform-modules-commonjs": "^7.26.3",
```
nachocodoner: I’m concerned about these changes. In previous research, removing proposals babel plugins to match the latest final versions caused Meteor to crash completely. Even if the tests pass green on the PR, these changes alone aren’t enough. We lack coverage for how they might impact the Meteor tool if we don't do an extra step required.

To proceed, we need to publish `meteor-babel`, update its version in all Meteor Atmosphere packages where it’s used, and verify that Meteor still works with all tests passing.

Additionally, more changes are needed, for example, `@babel/plugin-proposal-class-properties` is removed, but it’s explicitly used [in the meteor-babel options](https://github.com/meteor/meteor/blob/devel/npm-packages/meteor-babel/options.js#L89-L91) file. This will cause errors about it being missing, it will need to be replaced by the new plugin for class-properties as did [here](https://github.com/meteor/meteor/pull/13193/files#diff-9b2a6eb1239116bfebc2a178b41aa29996dc1a3f31c8ca24eb66f1f924c7ffdbR89). Issues like this will trigger, making Meteor unable to run effectively.

[Refer to this PR](https://github.com/meteor/meteor/pull/13193) as initially seemed like a straightforward improvement, but it turned out to require more effort than expected to address properly.

9Morello: @nachocodoner Thanks for the feedback. I addressed the issue with the `require` call still being present in that file. I'll take a look at the PR you linked. If there's anything else we can do to help make this PR more reliable and safer to merge, please let us know.

npm-packages/meteor-babel/options.js
```diff
@@ -85,11 +85,6 @@ exports.getDefaults = function getDefaults(features) {
 function maybeAddReactPlugins(features, options) {
   if (features && features.react) {
     options.presets.push(require("@babel/preset-react"));
-    options.plugins.push(
-      [require("@babel/plugin-proposal-class-properties"), {
-        loose: true
-      }]
-    );
```
nachocodoner, line 92: But `@babel/plugin-proposal-class-properties` was replaced by `@babel/plugin-transform-class-properties`, right? So this should be changed as we did [here](https://github.com/meteor/meteor/pull/13193/files#diff-9b2a6eb1239116bfebc2a178b41aa29996dc1a3f31c8ca24eb66f1f924c7ffdbR89) I believe.

And there are more spots affected by this.

packages/babel-compiler/package.js
```diff
@@ -1,13 +1,13 @@
...
+  version: '7.11.3',
 });
 
 Npm.depends({
-  '@meteorjs/babel': '7.20.1',
-  'json5': '2.1.1',
-  'semver': '7.3.8'
+  '@9morello/babel-dev': '8.0.1',
+  'json5': '2.2.3',
+  'semver': '7.6.3'
```
nachocodoner, line 10: I thought these changes were part of the dev_bundle tool and needed verification with a devbundle, but that's not the case. We can cherry-pick them to verify the bumps and release them directly.


Update: https://github.com/meteor/meteor/pull/13525

StorytellerCZ:
@nachocodoner I think the closure of this has been automatic due to the deletion of the branch?

PR #13520, opened 2024-12-18
Resolve package.json exports

### LINKED ISSUE #12324: Support ES modules new features
I try use a new ES module but catch the error

```
SyntaxError: Cannot use 'import.meta' outside a module
W20221129-01:33:22.402(3)? (STDERR)     at new Script (vm.js:102:7)
W20221129-01:33:22.402(3)? (STDERR)     at createScript (vm.js:262:10)
W20221129-01:33:22.402(3)? (STDERR)     at Object.runInThisContext (vm.js:310:10)
```

### LINKED ISSUE #13436: 3.0.4 - Npm package devlop breaks app - Module not found - Support of exports in package.json
Here is a repro repo : 
https://github.com/vparpoil/meteor-devlop-bug

Steps to reproduce : 
```bash
meteor create --react meteorDevlop
cd meteorDevlop
meteor npm install --save devlop
```

add this in client/main.jsx 
```javascript
import { deprecate, ok } from "devlop";
var c = deprecate;
```

App breaks with the following error in client console
```
Uncaught Error: Cannot find module 'devlop'
    at makeMissingError (modules-runtime-hot.js?hash=740f2079be9cde2bf010e27ef9713c36c6491a66:232:12)
    at Module.resolve (modules-runtime-hot.js?hash=740f2079be9cde2bf010e27ef9713c36c6491a66:247:17)
    at Module.moduleLink [as link] (modules.js?hash=4b1808092a5a19c8b57845dde04c4a786e6113c9:349:25)
    at app.js?hash=29a0a12e54b02747db3b1e35e18b35c6c59a27bb:194:11
    at module (main.jsx:12:3)
    at fileEvaluate (modules-runtime-hot.js?hash=740f2079be9cde2bf010e27ef9713c36c6491a66:386:7)
    at Module.require (modules-runtime-hot.js?hash=740f2079be9cde2bf010e27ef9713c36c6491a66:268:27)
    at mod.require (modules.js?hash=4b1808092a5a19c8b57845dde04c4a786e6113c9:284:33)
    at require (modules-runtime-hot.js?hash=740f2079be9cde2bf010e27ef9713c36c6491a66:308:21)
    at app.js?hash=29a0a12e54b02747db3b1e35e18b35c6c59a27bb:224:1
```

This appeared while wanting to use `remark`, `remark-gfm` and `remark-parse`, depending on `unified`, which depends on `devlop`

This is related to the following issue : https://github.com/wooorm/devlop/issues/1
> The package.json omits `main` in favor of `exports`

Seems also related to https://github.com/meteor/meteor/discussions/11727

Can meteor support Npm package.json which use `exports` and not `main`? 

Form[ Node documentation](https://nodejs.org/api/packages.html#exports), the use of exports is legit 
> The "exports" field allows defining the [entry points](https://nodejs.org/api/packages.html#package-entry-points) of a package when imported by name loaded either via a node_modules lookup or a [self-reference](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name) to its own name. It is supported in Node.js 12+ as an alternative to the ["main"](https://nodejs.org/api/packages.html#main) that can support defining [subpath exports](https://nodejs.org/api/packages.html#subpath-exports) and [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) while encapsulating internal unexported modules.

### LINKED ISSUE #13504: React App crashes with "Cannot use 'import.meta' outside a module" after updating react-router-dom
After updating _react-router-dom_ from version **6.28.0** to **7.0.1**, my app crashes, and the following error appears in the browser console:

**Uncaught SyntaxError: Cannot use 'import.meta' outside a module**

I am using **METEOR@3.0.4**, React for the front-end, and _react-router-dom_ for routing. The error occurs only after upgrading _react-router-dom_. The app works as expected with version **6.28.0**.

**Steps to Reproduce**
1 - meteor create app --react
2 -  meteor npm i react-router-dom
3 - start the app an check the browser console

**Environment**
Meteor 3.0.4
Node v20.17.0
React 18.2.0
React-router-dom : 7.0.1

The error message suggests the use of **import.meta**, which is incompatible outside ES modules. This issue appears to be related to how Meteor handles modules. The problem does not occur with r### eact-router-dom@6.28.0

### LINKED ISSUE #13836: @modelcontextprotocol/sdk import failure in basic project
I'm running Meteor 3.3.
I'm using macOS Sequoia 15.5.

I expect to import "StreamableHTTPClientTransport" from @modelcontextprotocol/sdk, but Meteor is "Unable to resolve some modules...". I was able to get around the issue by expanding the full filepath in the import, but still run into nested dependency issues for the "pkce-challenge" package. Expanding the full filepath should also not be necessary, since Node.js is able to resolve the example import successfully as is.

Image

[Reproduction](https://github.com/zcawood-mie/brokenMCPSDK/tree/main#) with steps taken detailed in README.

After cloning the repo, remember to ```meteor npm install``` to get the @modelcontextprotocol/sdk package in the node_modules folder.

I'm not certain, but believe these issues may also be related:
* https://github.com/meteor/meteor/issues/12004
* https://github.com/meteor/meteor/issues/13737

zodern:
This implements [package.json exports](https://nodejs.org/docs/latest/api/packages.html#package-entry-points), following [Node's specification](https://nodejs.org/api/esm.html#resolution-algorithm). It should be feature complete and spec compliant.

This PR is the build-time implementation.  The runtime implementation is in https://github.com/meteor/install/pull/1.

In the past some bundlers had added additional features to resolving package.json exports. However, in the past couple years all of the main bundlers have removed any additional features and standardized on implementing it as in Node's specification. This implementation attempts to do the same.

Meteor uses the same resolver for both commonjs and ecmascript modules, which adds some complexity when implementing the specification. I'm not aware of any remaining issues, but it is likely I missed something.

Meteor's bundler resolves the original import id at both build time, and during run time. This requires the bundles to have all of the information needed to resolve the imports, including anything relevant from package.json files. Since Meteor now uses the `exports` object in package.json files, this could drastically increase the bundle size. For example, the `@babel/runtime` package has an `exports` object that is 32kb. This implementation takes care to minimize the additional size from the `exports` objects - only the specific parts of the `exports` object needed for the files imported by the app are included in the bundle, and those parts are rewritten when possible to be smaller. It also excludes any other main fields from the package.json when there is an `exports` field.

This is likely a large breaking change. When the resolver finds an `exports` field, it disables any other main fields in the package.json. It also restricts what files the app or other packages can import from the package.

Additional features that could be implemented in later PR's
- [ ] The related feature of the [`import` field](https://nodejs.org/docs/latest/api/packages.html#subpath-imports) in package.json files. There should be very little left to do for this
- [ ] Build plugins will need to be able to add additional conditions, such as the `svelte` condition for apps or packages that have a build plugin to compile svelte files
- [ ] It might be interesting if app developers could add conditions - along with [self-referencing](https://nodejs.org/docs/latest/api/packages.html#self-referencing-a-package-using-its-name) imports (implemented in this PR), this could be used to create different builds of an app with different files included
- [ ] We don't have all of the errors and validations that Node does and described in the specification - I'm not sure build time is the right place for them, and at runtime it might not have enough of the `exports` object to be able to do them. The implementation is half ready so we could add more of the `exports` object as needed for run-time validations.
- [ ] Add the development and production conditions

Remaining tasks for this PR:
- [x] Merge and publish https://github.com/meteor/install/pull/1, and update this PR to use the new version
- [x] Update the copy of install.js in `module-runtime-hot`
- [ ] Make sure existing tests pass
- [ ] Test with some real apps to check how much this affects bundle size and build performance
- [ ] More testing with real apps to make sure there's no unhandled edge cases that break apps or packages
- [ ] If Meteor still has its [browserstack](https://github.com/meteor/meteor/tree/master/tools/tool-testing#testing-with-phantombrowserstack) testing, it should be run before this feature is released.  I was careful to ensure client code is still es5, but I might have missed something.

I don't have any large Meteor 3 apps to test with, so I would appreciate help with that.


Leekao:
I would like to test your branch if possible, I have a service that broke down when one of our dependencies moved to export package.json exports.

boomfly:
Any update on this?

Sergeant61:
Hello, there is a topic in the forum, is there any development regarding the progress of this pr? https://forums.meteor.com/t/meteor-langchain-ollama-error/62398/9

italojs:
@zodern  is there any chance you could continue working on it? This PR looks really great!

zodern:
I probably won't have any time until July to resume working on this. I think the only things left are to fix the tests, and maybe to improve the validation and error handling.

NigelCunningham:
I'm seeking to try the patch with a largish project. It seems to be working in that my error with esm-env has gone away, but I'm instead getting errors relating to the Svelte 5 package using imports. The functionality looks similar so I'll see if I can adapt the patch to handle them too (that said, I'm no JS guru so don't expect too much!)

NigelCunningham:
Could I get some help regarding how to get to step through the new code, please? Trying to add --inspect-brk doesn't seem to be getting me anywhere yet; I've never done any serious dev on Meteor itself before.

zodern:
@NigelCunningham, do you have a simple reproduction I could look at?

If the error is happening on the client or server during runtime, you can start by looking at the generated code, and making sure the relevant fields of the exports object exist in the package.json modules. You can also step through the code at the imports (disabling source maps might be helpful, so you can see the generated code for the import) to see why it is incorrectly resolving.

NigelCunningham:
Yeah - I think if you just add Svelte 5 via npm, it has imports in the package.json... oh, but I forgot I'm also using jorgenvatle:vite - perhaps that's part of the issue?

My meteor/packages is:

```
# Meteor packages used by this project, one per line.
# Check this file (and the other files in this directory) into your repository.
#
# 'meteor add' and 'meteor remove' will edit this file for you,
# but you can also edit it by hand.

meteor-base@1.5.2             # Packages every Meteor app needs to have
mobile-experience@1.1.2       # Packages for a great mobile UX
mongo@2.1.4                   # The database Meteor supports right now

standard-minifier-css@1.9.3   # CSS minifier run for production mode
standard-minifier-js@3.1.1    # JS minifier run for production mode
es5-shim@4.8.1                # ECMAScript 5 compatibility for older browsers
typescript@5.6.6              # Enable TypeScript syntax in .ts and .tsx modules
shell-server@0.6.2            # Server-side component of the `meteor shell` command


zodern:melte            # Meteor package to allow us to create files with the .svelte extension
hot-module-replacement@0.5.4  # Update client in development without reloading the page
zodern:types            # Enable types from meteor/atmosphere packages
jorgenvatle:vite
autoupdate
accounts-base
accounts-password
accounts-ui
accounts-ui-unstyled
roles
blaze
templating
reactive-var
underscore
svelte:blaze-integration
ostrio:files
ecmascript
```

NigelCunningham:
@zodern I've managed to make progress since my last comment - I previously still had the ecmascript package installed. I've since removed that and moved to the tip of the Meteor branch. That has allowed me to make some more progress, so I think I have nothing useful to add here, sorry.

NigelCunningham:
Should this patch make eventemitter3 usable? It has:

```
{
  "name": "eventemitter3",
  "version": "5.0.1",
  "description": "EventEmitter3 focuses on performance while maintaining a Node.js AND browser compatible interface.",
  "exports": {
    ".": {
      "types": "./index.d.ts",
      "import": "./index.mjs",
      "require": "./index.js"
    },
    "./package.json": "./package.json"
  },
```

Merging PR 13520 into current tip, I'm still getting (I've modified the error message to show the value):

```
______[ModuleExportsError]___________________________________________________________: 
W20250930-11:39:02.370(10)? (STDERR)  ERROR  Meteor bundle had a package.json key (exports) with an unexpected value ({"type":"ObjectExpression","start":140526,"end":140554,"loc":{"start":{"line":3591,"column":13,"index":140526},"end":{"line":3593,"column":3,"index":140554}},"properties":[{"type":"ObjectProperty","start":140532,"end":140550,"loc":{"start":{"line":3592,"column":4,"index":140532},"end":{"line":3592,"column":22,"index":140550}},"method":false,"key":{"type":"StringLiteral","start":140532,"end":140535,"loc":{"start":{"line":3592,"column":4,"index":140532},"end":{"line":3592,"column":7,"index":140535}},"extra":{"rawValue":".","raw":"\".\""},"value":"."},"computed":false,"shorthand":false,"value":{"type":"StringLiteral","start":140537,"end":140550,"loc":{"start":{"line":3592,"column":9,"index":140537},"end":{"line":3592,"column":22,"index":140550}},"extra":{"rawValue":"./index.mjs","raw":"\"./index.mjs\""},"value":"./index.mjs"}}]})'.
W20250930-11:39:02.371(10)? (STDERR) This might be important to properly parse the module's entrypoint. Do open a new issue if you run into any issues. 🙏
W20250930-11:39:02.371(10)? (STDERR) undefined
```

I've identified that it relates to eventemitter3 by printing (in the caller) the name and path of the module being processed.

PR #13499, opened 2024-11-28
Tweaks on allow/deny rules

nachocodoner:
OSS-628

Context: https://github.com/meteor/meteor/issues/13444 and related https://github.com/meteor/meteor/pull/13468

This PR reviews allow/deny rules to ensure they work as expected and simplifies rule definitions, aligning them with MongoDB operations while deprecating the sync/async distinction.

- [x] Deprecate async rules for allow/deny, [deprecate logs will appear](https://github.com/user-attachments/assets/1e6c3498-bd8e-484e-b706-b31116da7d1c)
- [x] Add test scenarios for MongoDB operations and verify correct behavior, including async flows
- [x] Ensure test scenarios from [this PR](https://github.com/meteor/meteor/pull/13494) are resolved
- [x] Update docs after last allow/deny changes

Pending: Clean up and refactor with TS allow/deny code (Will cover in a different refactor PR)

PR #13522, opened 2024-12-19
[Meteor.deprecate] ObjectID alias to backward compatible of meteor packages and allow-deny changes

nachocodoner:
This PR introduces a new Meteor helper for handling deprecations. The helper logs deprecation warnings only in development mode and includes a trace to help identify the context for resolving the issue.

It also addresses deprecations related to `ObjectID` and recent changes to `allow-deny` rules.

## ObjectID

Some packages use `MongoInternals.NpmModule.ObjectID`, but after the Mongo driver update, it is now `MongoInternals.NpmModule.ObjectId`.

`MongoInternals.NpmModule` refers to the driver module, but the driver no longer provides a direct reference to `ObjectID.` Access now is `MongoInternals.NpmModule.ObjectId`.

Ideally, packages should address this MongoDB driver breaking change themselves. Since this change surfaced with Meteor 3.1 and wasn’t reported as a breaking change, this PR restores it for backwards compatibility with existing packages. A stricter enforcement can be considered in a later Meteor minor release. Additionally, `Meteor.deprecate` is applied to guide developers in updating their packages.

## allow-deny

[A previous PR](https://github.com/meteor/meteor/pull/13499) deprecated allow/deny async rules, but these logs appeared in some apps without sufficient context, annoying devs. With the new `Meteor.deprecate` helper, such cases now provide better insights on where changes for deprecations are needed. For example, with this helper I identified and fixed an allow/deny rule within the Meteor core itself.

---------

image

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1806,21 +1806,6 @@ const setupUsersCollection = async users => {
 
       return true;
     },
-    updateAsync: (userId, user, fields, modifier) => {
-      // make sure it is our record
-      if (user._id !== userId) {
-        return false;
-      }
-
-      // user can only modify the 'profile' field. sets to multiple
-      // sub-keys (eg profile.foo and profile.bar) are merged into entry
-      // in the fields list.
-      if (fields.length !== 1 || fields[0] !== 'profile') {
-        return false;
-      }
-
-      return true;
-    },
     fetch: ['_id'] // we only look at _id.
   });
 
@@ -1861,4 +1846,3 @@ const generateCasePermutationsForString = string => {
   }
   return permutations;
 }
-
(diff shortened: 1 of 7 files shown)


leonardoventurini (approved):

harryadel:
Maybe it should output a little message instead of masking it?

nachocodoner:
@harryadel: Right

I added a Proxy to monitor access to the variable and display a warning when it's used. Since `MongoInternals` is a server-only module and Proxy is well-supported, this approach works. The warnings will appear in the app, even when caused by the package using it, but I'm unsure how to capture that package's context for the warning.

leonardoventurini (approved):

StorytellerCZ (approved):

PR #13540, opened 2025-01-09
fix: built-in dev mongo can't start on win32 when username has spaces

jstarpl:
On win32, if the `.meteor` folder is located in a path containing Space characters, `meteor` command will fail to run.

On Windows, it's a pretty common pattern to have a user account name that contains spaces, say "John Smith". If that happens, Windows will create a user folder with that space. If one then installs meteor using such a user, `.meteor` folder, installed in `%APP_DATA%` will end up on a path with spaces. When then the `meteor` CLI is used to start up a project, this will fail to run when trying to start the local dev `mongo` server, with a message similar to:

```
=> Started proxy.
'C:\Users\Jan' is not recognized as an internal or external command, operable program or batch file.
Unexpected mongo exit code 1. Restarting.
'C:\Users\Jan' is not recognized as an internal or external command, operable program or batch file.
Unexpected mongo exit code 1. Restarting.
'C:\Users\Jan' is not recognized as an internal or external command, operable program or batch file.
Unexpected mongo exit code 1. Restarting.
'C:\Users\Jan' is not recognized as an internal or external command, operable program or batch file.
Unexpected mongo exit code 1. Restarting.
Can't start Mongo server.
MongoDB failed global initialization
```

The problem originates in the `mongo` executable being started with the `shell: true` option. In those circumstances, Node's integration with the OS effectively causes the `command` and `args` to be concatenated to a single string and passed off to the shell. This in turn will tokenize this string using white-spaces and mistakenly take the "pre-space" portion of the executable path to be the executable itself, resulting in the above message.

Wrapping the paths in quotes allows the command line processing in the shell to succeed.

### DIFF
--- a/tools/runners/run-mongo.js
+++ b/tools/runners/run-mongo.js
@@ -28,6 +28,17 @@ function spawnMongod(mongodPath, port, dbPath, replSetName) {
   mongodPath = files.convertToOSPath(mongodPath);
   dbPath = files.convertToOSPath(dbPath);
 
+  // Because we're spawning the process with shell: true on win32, the
+  // command string and args array are effectively concatenated to
+  // a single string and passed into the shell. If any of those paths
+  // contain spaces, these will get broken up on white-spaces and treated
+  // as separate tokens. If they are quoted, they will be parsed
+  // correctly by the shell.
+  if (process.platform === 'win32') {
+    mongodPath = '"' + mongodPath + '"'
+    dbPath = '"' + dbPath + '"'
+  }
+
   let args = [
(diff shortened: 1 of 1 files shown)


nachocodoner (approved):
I’ve seen this reported [on the forums](https://forums.meteor.com/t/cannot-start-meteor-3-0-4-on-windows/62524), and we identified the cause. The fix is much appreciated. I’ll test it myself and include it in an upcoming release. Thank you.

nachocodoner:
Merging into a branch within the Meteor context and having 3.1.1 changes to test and validate.

Thank you for your contribution!

Issue #13518, opened 2024-12-16
allow-deny deprecation warning in 3.1.1 should be configurable and include stack trace


perbergland:
3.1.1 beta 1

I saw a lot of deprecation warnings on the console but could not understand where they were coming from.

So I changed this line to include a stack trace:

```
console.warn(allowOrDeny + `: The "${key}" key is deprecated. Use "${syncKey}" instead.`, new Error().stack);
```

I suggest that should be added, at least behind an environment variable.

Also, for packages that need to use the deprecated options for backwards compatibility, like Collection2, there should be a way to disable the deprecation check to not see them on every startup (two deprecation warnings per collection.attachSchema).

E.g. a new boolean flag in options.

leonardoventurini:
@nachocodoner did some fantastic improvements here:

https://github.com/meteor/meteor/pull/13522/files

So, this is where the Node.js you mentioned would fit in, right?

https://www.tutorialspoint.com/node-js-process-tracedeprecation-method

nachocodoner:
Thank you for this report. We’ll provide a new environment variable to ignore these warnings as booleans, by default these will show. I’ll likely add a regex option for pattern-based ignores.

nachocodoner:
The new changes to improve deprecation logs are available on this PR and will be introduced in the next 3.1.1 official version.

https://github.com/meteor/meteor/pull/13542

nachocodoner:
Changes available in the 3.1.1 offical release. To update, use `meteor update --release 3.1.1`.

PR #13542, opened 2025-01-13
Deprecation messages can be managed through environment variables.

nachocodoner:
OSS-650

Context: https://github.com/meteor/meteor/issues/13518 and [forum post](https://forums.meteor.com/t/disable-deprecation-messages/62841)

This PR addresses feedback on controlling deprecation warnings using environment variables. These updates are part of the final adjustments for the official 3.1.1 release.

- Deprecation warnings are shown by default.
- Use `METEOR_NO_DEPRECATION` (set to true or a regex) to control visibility.
- Each warning is displayed only once per stack trace.
- Node's `--no-deprecate` option also suppresses Meteor deprecation logs.
- When METEOR_NO_DEPRECATION is used, a log will appear once if there are existing deprecations, encouraging their report or fix.

image

image

When deprecation is hidden, a single message will appear, prompting users to report or address it.

image

### DIFF
--- a/packages/meteor/deprecate.js
+++ b/packages/meteor/deprecate.js
@@ -9,12 +9,12 @@ if (Meteor.isServer && Meteor.isDevelopment) {
 }
 
 function oncePerArgument(func) {
-  const cache = new Map();
+  var cache = new Map();
 
-  return function (...args) {
-    const key = JSON.stringify(args);
+  return function _oncePerArgument() {
+    var key = JSON.stringify(arguments);
     if (!cache.has(key)) {
-      const result = func.apply(this, args);
+      var result = func.apply(this, arguments);
       cache.set(key, result);
     }
     return cache.get(key);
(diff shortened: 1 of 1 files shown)


packages/meteor/deprecate.js
```diff
@@ -1,3 +1,26 @@
+if (Meteor.isServer && Meteor.isDevelopment) {
+  if (typeof __meteor_runtime_config__ === 'object') {
+    var noDeprecation = process.env.METEOR_NO_DEPRECATION || process.noDeprecation;
+    if (noDeprecation === 'true' || noDeprecation === 'false') {
+      noDeprecation = noDeprecation === 'true';
+    }
```
leonardoventurini, line 6: Would this still work? 😄 

```suggestion
    noDeprecation = noDeprecation === 'true';
```

packages/meteor/deprecate.js
```diff
@@ -1,3 +1,26 @@
...
+    __meteor_runtime_config__.noDeprecation = noDeprecation;
+  }
+}
+
+function oncePerArgument(func) {
+  var cache = new Map();
```
leonardoventurini, line 12: Is there any chance of the cache growing indefinitely if the arguments change, causing a memory leak or hitting 16m items?

nachocodoner: We don't use that amount of Meteor.deprecate I think hehe

packages/meteor/deprecate.js
nachocodoner: The idea is to support this variable as either a boolean or a string regex. The `true/false` comparison ensures proper conversion between the boolean string and its true/false value. Otherwise, the string remains unchanged as regex string.

With your proposal, any regex string will always evaluate as `false`.

leonardoventurini (approved):

Issue #13548, opened 2025-01-14
ISOBuild should abort on symlinks in /public


cunneen:
Meteor 3.1 / MacOS

When the bundler encounters a symbolic link in the `/public` folder, it currently just silently ignores the symlink and ALL SUBSEQUENT ASSETS i.e. you'll just eventually discover your production server is not returning your assets.

To reproduce:

```shellsession
$ meteor create --minimal symlinkbug
Using minimal skeleton
Created a new Meteor app in 'symlinkbug'.     

To run your new app:                          
  cd symlinkbug                               
  meteor                                      
                                              
If you are new to Meteor, try some of the learning resources here:
  https://www.meteor.com/tutorials            
                                              
When you’re ready to deploy and host your new Meteor application, check out Cloud:
  https://www.meteor.com/cloud                
                                              
$ cd symlinkbug
$ mkdir -p public/folder1 && \
      echo '/* hello */' >> public/folder1/file.css && \
      ln -s folder1 public/folder2 && \
      echo '/* a.css */' > public/a.css && \
      echo '/* z.css */' > public/z.css
      
$  find public -ls
478373428        0 drwxr-xr-x    6 cunneen      staff                 192 15 Jan 04:05 public
478373431        0 lrwxr-xr-x    1 cunneen      staff                   7 15 Jan 04:05 public/folder2 -> folder1
478373433        8 -rw-r--r--    1 cunneen      staff                  12 15 Jan 04:05 public/z.css
478373432        8 -rw-r--r--    1 cunneen      staff                  12 15 Jan 04:05 public/a.css
478373429        0 drwxr-xr-x    3 cunneen      staff                  96 15 Jan 04:05 public/folder1
478373430        8 -rw-r--r--    1 cunneen      staff                  12 15 Jan 04:05 public/folder1/file.css

$ meteor
```

If you open another session, you'll be able to verify that none of the css assets are in the manifest (and that the server is not serving them).

```shellsession
$  grep css .meteor/local/build/programs/web.browser/program.json
      "path": "packages/minifier-css.js",
      "url": "/packages/minifier-css.js?hash=864ebafdef84ddc1a40a47f2cd29f44055c4e50a",
      "path": "packages/standard-minifier-css.js",
      "url": "/packages/standard-minifier-css.js?hash=289cfa41f9db2b3881c33f66ae86859986e6cc4e",
      "path": "merged-stylesheets.css",
      "type": "css",
[...]

PR #13549, opened 2025-01-14
isobuild should abort on symlink

cunneen:
Fix for #13548

When the bundler encounters a symbolic link in the /public folder, it currently just silently ignores the symlink and ALL SUBSEQUENT ASSETS.

This fix changes that behaviour to instead abort the build with an error message:

```
=> Errors prevented startup:                  
   
   While building the application:
   error: Symlink cycle detected at /public/fonts
   
```

This *may* potentially break builds which were previously silently ignoring symlinks (e.g. a symlink from /public/webfonts => /public/fonts ).


nachocodoner:
It aborts the build so the dev can remove symbolic links and properly resolve the public folder to include all assets. Is this?

I’m unsure if symbolic links have ever worked as expected in the Meteor bundler and resolve their contents. I'd like to hear @zodern's thoughts on this.

zodern:
I can give a better answer next month. For now, I can share some brief thoughts on this.

Many parts of Meteor's bundler do handle symlinks in various ways. Though I'm not currently aware of the history of handling symlinks in `private` and `public` folders. Looking at the code, it seemed it was intended to handle symlinks, and I think how it is expected to behave (silently ignore symlink cycles) is expected and correct.

I would need to spend more time to know for sure, but I think the problem is it returns an empty list of assets when it finds a symlink cycle, where it should instead just skip the current folder, and continue with searching the other folders.

PR #13554, opened 2025-01-17
Feature argon2 password encryption

vparpoil:
Following this discussion : https://github.com/meteor/meteor/discussions/13529 I worked on a proposal to implement `argon2` as a drop-in replacement for `bcrypt` to securely store user password.

The `checkPasswordAsync` function has been updated to enable a seamless transition from `bcrypt` to `argon2`. A test has been added to verify this transition. 

The `accounts-password` package still includes `bcrypt` as a dependency to support legacy hashed passwords


packages/accounts-password/package.js
```diff
@@ -5,11 +5,12 @@ Package.describe({
   // 2.2.x in the future. The version was also bumped to 2.0.0 temporarily
   // during the Meteor 1.5.1 release process, so versions 2.0.0-beta.2
   // through -beta.5 and -rc.0 have already been published.
-  version: "3.0.3",
```
StorytellerCZ: This will be a major version bump.

v3-docs/docs/api/accounts.md
StorytellerCZ, line 1: We'll need a dedicated page to explain the why, what and how.
We will need detailed justification for people to read into and we will need to retain the bcrypt reference link for a bit as well.

StorytellerCZ (changes requested):
So far looks good, will have to test this thoroughly and add more docs.

v3-docs/docs/api/accounts.md
vparpoil: 👍 Added some more details in the next commit

packages/accounts-password/package.js
vparpoil: I guess the [CONTRIBUTING.md](https://github.com/meteor/meteor/blob/devel/CONTRIBUTING.md#submitting-pull-requests) should be clarified to handle such cases: here the changes fall in the first category : 
> If your changes are ok to be released without a whole new Meteor version bump just the patch, for example, 2.4.5 will become 2.4.6.

I would have pushed a major update in the version without reading this line ;)

StorytellerCZ: Technically you are correct. But in this case it is a major change as it changes a critical piece of the package related to security. Major version bump is always a headache, which in this case we want so that people notice, be aware of this and take action with their users if needed.

v3-docs/docs/api/accounts.md
```diff
@@ -773,10 +773,45 @@ sign-in process, it also supports email-based sign-in including
...
+
+To update the timeCost, use the following configuration:
+
+```js
+Accounts.config({
+    argon2Iterations: 4
+});
+```
+
+Other Argon2 parameters, such as `hashLength`, `memoryCost`, `parallelism`, and `type`, are set to default values:
```
StorytellerCZ: We should probably offer config API to set them as well.

StorytellerCZ (changes requested):
Planning for the future v5 of accounts-password will remove `bcrypt` dependency.

So we can release this in Meteor 3.2 and at a later time we will release v5 of accounts-password without `bcrypt` and the migration to give enough time for people to migrate.

v3-docs/docs/api/accounts.md
StorytellerCZ: I know that there are financial institutions that are using Meteor, so they might want to have enhanced settings.

vparpoil: Following up here. I propose to provide `type`, `timeCost`, `memoryCost` and `parallelism`  as config options.

HashLength is not configurable because it is not stored in the argon2 hash, hence not enabling a seamless change of config without storing it in database in an extra field.

StorytellerCZ (dismissed):
LGTM

packages/accounts-password/package.js
```diff
@@ -5,11 +5,12 @@ Package.describe({
   // 2.2.x in the future. The version was also bumped to 2.0.0 temporarily
   // during the Meteor 1.5.1 release process, so versions 2.0.0-beta.2
   // through -beta.5 and -rc.0 have already been published.
-  version: "3.0.3",
+  version: "4.0.0",
 });
 
 Npm.depends({
   bcrypt: "5.0.1",
+  argon2: "0.41.1",
```
nachocodoner: should we keep using bcrypt just because the migration to Argon2 occurs during login?

vparpoil: Yes, this is needed because there is no other way to migrate existing users since we don't know their passwords

v3-docs/docs/api/accounts.md
```diff
@@ -772,11 +772,74 @@ authentication. In addition to the basic username and password-based
...
+
+Passwords are hashed on the client using **SHA-256** algorithm before being sent to the server. This ensures that sensitive data is never transmitted in plain text. Once received by the server, the hashed value is further encrypted and securely stored in the `Meteor.users` collection.
+
+
+**About the migration process from `bcrypt` to `argon2`**
+
+The transition from `bcrypt` to `argon2` happens automatically upon user login. If Argon2 encryption is enabled in an existing application, each user's password is re-encrypted during their next successful login.
+- Step 1: The password is first validated against the existing `bcrypt` hash.
+- Step 2: If authentication succeeds, the password is re-encrypted using `Argon2`.
+- Step 3: The new `Argon2` hash replaces the old `bcrypt` hash in the database.
```
nachocodoner: It might be worth noting in the documentation that reverting to bcrypt is possible, and migration happens automatically the same way. Currently, both options coexist within the same implementation.

nachocodoner:
[I’m preparing an alpha version](https://github.com/meteor/meteor/pull/13607) but wanted to confirm if the checks are fine. Some tests from the changes [fail in CI](https://app.travis-ci.com/github/meteor/meteor/builds/274092333) but don’t fail locally when I run them. We need to fix this before the alpha. I’ll check if there’s any flakiness in our CI setup or in how the test is written.


image

nachocodoner:
[I fixed the tests](https://github.com/meteor/meteor/pull/13607) by using a test helper in Meteor that uses waitUntil to check specific conditions async. This approach avoids timeouts usage, which can be unreliable as their execution depends heavily on the run context. Here’s the commit: https://github.com/meteor/meteor/pull/13607/commits/aa3b2208bdf6fe524c695b6a6ae59f4c1865f01c

Next, I will create the alpha version.

nachocodoner:
We attempted to publish an alpha version of `accounts-password` and `accounts-base`, but due to how the package system works, upgrading these on apps running different Meteor versions triggers version restriction errors. This happens even with basic apps, so it might not be ideal for others to try.

To simplify things, we’ll publish a new Meteor 3.2 beta version that includes this change along with other contributions. This will make testing easier, just upgrade to the latest Meteor beta version, and all changes will be included.

I’ll let you know as soon as it’s available, likely later this week or early next week.

nachocodoner:
[I published `3.2-beta.1`](https://github.com/meteor/meteor/pull/13610) (`meteor update --release 3.2-beta.1`) including this feature.

You can already test this feature. I will do some testing in coming weeks.

StorytellerCZ:
Running some tests on this today. Will update my findings in this post.
* accounts-ui(-unstyled) will need a patch to allow for new major version of `accounts-password`, this will impact any packages that use or depend on it. Will create probably the same issue as with the previous major version bump, so people will have to be prepared for it.

zodern:
I would like to make a couple suggestions:

1. The defaults use significantly more memory and cpu than the values suggested by [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id). Many Meteor apps run in production with limited memory and cpu and with only a single shared core. It seems the current defaults in this PR could easily overwhelm a server if many people log in at the same time. At the very least, I think we should consider decreasing the parallelism to 1 by default, and do some load tests to ensure it won't perform significantly worse than bcrypt. With the current memory settings, 2 people logging in at the same time would cause a tiny container on Galaxy to run out of memory, or maybe even with a single person.

2. It seems we still are hashing the password with sha256. At the very least, we should probably upgrade it to sha512 with argon2. Though it would be preferable to completely eliminate pre-hashing when using argon2 to eliminate the password shucking vulnerability. If we want to prevent a very long password from overwhelming the server, one option is we could enforce a reasonable limit such as 1000 bytes, though I'm not sure this is even a concern with argon2.

vparpoil:
Thanks @zodern for your insights!

### Regarding default values for parameters
I used defaults from the [Npm Argon2 package](https://github.com/ranisalt/node-argon2/wiki/Options) :` t=3, m=64MB, p=4 `
I just found some more details in the [OWASP github](https://github.com/OWASP/CheatSheetSeries/pull/1537#issuecomment-2477673756), specifically this line : 
> `m=64 MiB, t≥3, p=1` is great for auth but is on the high end (PHP's default is `m=64 MiB, t=4, p=1`).
I understand from my readings that Argon2 RFC's defaults are too strong

Owasp recommandation [is from January 2023 and has explanations on the choice here](https://github.com/OWASP/CheatSheetSeries/pull/1073/files): 
> Use [Argon2id](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) with a minimum configuration of 19 MiB of memory, an iteration count of 2, and 1 degree of parallelism.

I agree on  going with OWASP values for default values, I will add a commit for this change


### Regarding sha256
Hashing the password on the client protects users, especially the one who reuse passwords: 
- prevents MITM attack on clear-text password
- ensures the server never receives clear text password, hence reducing risks (accidental logging, misusage)
- if encrypting algorithm gets compromised, a leaked DB won't expose clear text passwords

From my understanding, I think hashing the password on the client is still the best option (since we don't store the hashed value), even if we are theoretically vulnerable to password shucking

Moving to sha512 cannot be seamless in an easy way. The bcrypt(sha256(password)) is not equal to bcrypt(sha512(password)). To achieve seamless migration, we need to send both sha256 and sha512 from client to server. 

The security benefit of moving from sha256 to sha512 seems quite minimal from what I have read.

nachocodoner:
@vparpoil: Thanks for your latest changes and all the work on this.

I'm merging this branch now with the latest updates to include them in the official 3.2 release. Since it's an opt-in feature, this gives the community more time to transition and provide feedback before a full replacement.

On the core side, we've tested the main transitions for the library in several password-based apps, and everything worked as expected. Performance tests are still pending, but that will be worthwhile in the second stage when we start replacing bcrypt to ensure we deliver an equally performant alternative, or we need to tweak the settings further for that.

Feel free to open new PRs of tweaks and adjustments for future Meteor versions based on feedback received.

Issue #13530, opened 2024-12-22
two issues with ``meteor create --vue --release 3.1.1-beta.1`` 


mattsouth:
If I create the vue skeleton using the above command the wizard asks me to select the skeleton to use (which it didnt do in earlier versions) and it installs jorgenvatle:vite-bundler@3.0.0 which works in development with a warning to upgrade to meteor-vite@2 in npm but doesnt work with the --production flag.  dropping back to using jorgenvatle:vite-bundler@2.3.2 works in development without the warning and also works with the --production flag.

jankapunkt:
@JorgenVatle might know what's the best action here?

mattsouth:
Another option is to use jorgenvatle:vite with vite6 instead of jorgenvatle:vite-bundler but that's a bigger change to the vue skeleton, and maybe others too.

leonardoventurini:
@mattsouth are you currently blocked by this, or the workaround works for you? We will be evaluating if we need to update the vue skeleton, thanks for letting us know.

mattsouth:
Im not personally blocked by this at the moment.  I think ensuring each skeleton installs without error and then runs in development and production is a sensible quality check for a release but I appreciate this is difficult thing to automate.

JorgenVatle:
I'd be up to update the skeleton apps depending on `jorgenvatle:vite-bundler` to upgrade to the latest stable release of the package.

There's also some structuring changes due that should make things a little more clear when building Meteor apps with Vite. Namely refactoring the placement and naming of the Vite and Meteor entry modules to be more in line with the current documentation for Meteor-Vite.

nachocodoner:
@JorgenVatle: I updated to latest the meteor-vite and related dependencies on this PR, https://github.com/meteor/meteor/pull/13559

I tested locally and seems to work fine, but another pair of eyes would be good.

Edit: with `--production` mode I get a white screen with some logs in the console tab.

![Image](https://github.com/user-attachments/assets/e43a2d16-746c-4eee-961e-a772fe8608ab)

![Image](https://github.com/user-attachments/assets/d8985761-a6ef-43aa-bc21-9679b550a3c6)

JorgenVatle:
@nachocodoner Appears to be an issue introduced in `meteor-vite@3.0.0`. Give it a go with `meteor-vite@3.2.1` 🤞

mattsouth:
https://gitlab.com/mattsouth/meteor-vue-3.1.1

[...]

PR #13559, opened 2025-01-22
Upgrade skeletons that depends on meteor-vite

nachocodoner:
OSS-679

Context: https://github.com/meteor/meteor/issues/13530

This PR updates the Vue and Solid skeleton dependencies to their latest versions in the meteor-vite library, preventing issues.

I tested the apps, and they seem to work fine with the new version requirements.

image

### DIFF
--- a/tools/static-assets/skel-solid/.meteor/packages
+++ b/tools/static-assets/skel-solid/.meteor/packages
@@ -19,4 +19,4 @@ hot-module-replacement  # Update client in development without reloading the pag
 
 ~prototype~
 static-html             # Define static page content in .html files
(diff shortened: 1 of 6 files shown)


Grubba27:
The Solid skeleton test has been failing; not sure why.

The error says something about a dependency

nachocodoner:
@Grubba27: Yeah. I pushed some fixes required on using meteor-vite latest version and also reviewed both vue and solid example. In solid example the `picocolors` dependencies was also required for the app to load properly.

I still want to get rid of an annoying warning showed [on running the apps](https://github.com/meteor/meteor/issues/13530#issuecomment-2616177632).

PR #13603, opened 2025-02-03
13599 - Fix to oplog stuck in fetching race condition

### LINKED ISSUE #13599: Race condition can leave mongo oplog observe driver stick in phase FETCHING
This issue affects version 3.0.4 and 3.1.1, at least.  This was not a problem in 1.10.1.
I'm running on Ubuntu 22.04
The expected behavior is that all oplog observe drivers fluctuate between STEADY, QUERYING, and FETCHING, always eventually ending back in STEADY.
In this race condition, there is a way that the `waiting` variable that keeps track of outstanding promises falls out of sync with reality, dropping to a value of -1, which means the promise will never resolve.  This leads to the driver never making it back to STEADY, which leave writefences unresolved, which leads to `updated` DDP events not being sent, which can manifest as Meteor methods not "returning" to the client (though the `result` message makes it, the `updated` message does not). This appears to happen because the `DocFetcher` -> `fetch` method can call a callback more than once.  If that happens, `waiting` gets decremented more than once, potentially leaving it negative.

This is extremely difficult to reproduce, unfortunately, so I can't provide a simple reproduction.  However, I do have a fix that I can propose that I am forced to use in my app to prevent this error from occurring, and I have a log that demonstrates the problem.

Log:
```
2025-01-26 07:45:00.407 [1minfo: [22m [DebugMethods]  Inside while loop { id: 51, needtoFetchCount: 2, stopped: false }
2025-01-26 07:45:00.407 [1minfo: [22m [DebugMethods]  Waiting counter status after increment {
  id: 51,
  waiting: 1,
  currentlyFetchingSize: 2,
  phase: 'FETCHING',
  fetchGeneration: 805
}
2025-01-26 07:45:00.452 [1minfo: [22m [DebugMethods]  Inside finishIfNeedToPollQuery callback { id: 51, err: null, docId: 'uX5rTXAMf2grJ7KCC' }
2025-01-26 07:45:00.452 [1minfo: [22m [DebugMethods]  About to handle doc { id: 51 }
2025-01-26 07:45:00.452 [1minfo: [22m [DebugMethods]  Waiting counter status after decrement { id: 51, waiting: 0, phase: 'FETCHING', fetchGeneration: 805 }
2025-01-26 07:45:00.452 [1minfo: [22m [DebugMethods]  Waiting counter status after increment {
  id: 51,
  waiting: 1,
  currentlyFetchingSize: 2,
  phase: 'FETCHING',
  fetchGeneration: 805
}
2025-01-26 07:45:00.452 [1minfo: [22m [DebugMethods]  Inside while loop { id: 51, needtoFetchCount: 1, stopped: false }
2025-01-26 07:45:00.452 [1minfo: [22m [DebugMethods]  Waiting counter status after increment {
  id: 51,
  waiting: 1,
  currentlyFetchingSize: 1,
  phase: 'FETCHING',
  fetchGeneration: 806
}
2025-01-26 07:45:00.454 [1minfo: [22m [DebugMethods]  Inside finishIfNeedToPollQuery callback { id: 51, err: null, docId: '7t5FEj7RnBDd9ocLG' }
2025-01-26 07:45:00.454 [1minfo: [22m [DebugMethods]  About to handle doc { id: 51 }
2025-01-26 07:45:00.454 [1minfo: [22m [DebugMethods]  Waiting counter status after decrement { id: 51, waiting: 0, phase: 'FETCHING', fetchGeneration: 806 }
**2025-01-26 07:45:00.503 [1minfo: [22m [DebugMethods]  Inside finishIfNeedToPollQuery callback { id: 51, err: null, docId: '7t5FEj7RnBDd9ocLG' }**
2025-01-26 07:45:00.504 [1minfo: [22m [DebugMethods]  About to handle doc { id: 51 }
2025-01-26 07:45:00.504 [1minfo: [22m [DebugMethods]  Waiting counter status after decrement { id: 51, waiting: -1, phase: 'FETCHING', fetchGeneration: 806 }
2025-01-26 07:45:00.504 [1m[31merror: [39m[22m [DebugMethods]  Waiting counter became negative { id: 51, waiting: -1 }
```

The bolded line is where the second fetch callback runs, reducing the `waiting` value to -1

My proposed fix is to `oplog_observe_driver` -> `_fetchModifiedDocuments`, to remove the `waiting` counter altogether in favor of using promises to the same effect - that is, capture the `fetch` promises and `await Promise.all(...)`.  At the very least, this will mean the promise will resolve, even if the callback happens twice

I would love to discuss this solution as I'm not positive it doesn't have other drawbacks.  However, this error has occurred a few times per day every day since I upgraded, for 14 days (first to 3.0.4, then to 3.1.1 in hopes of a fix to this issue).  Since my patch, there have been no adverse affects and no more instances of the issue.

gbuerk:
#13599 - Removed the `waiting` variable in favor of using promises to keep track of fetch requests.  This was done to mitigate a race condition where a single fetch could increment `waiting` while the callback could be called twice, potentially decrementing  `waiting` to -1 leaving the driver stuck in `PHASE.FETCHING`


packages/mongo/oplog_observe_driver.js
```diff
@@ -565,20 +558,26 @@ Object.assign(OplogObserveDriver.prototype, {
...
-          );
+              })
+            )
+          })
+          fetchPromises.push(fetchPromise);
         });
-        await awaitablePromise;
+        // Wait for all fetch operations to complete
+        try {
+          await Promise.all(fetchPromises);
```
zodern: This will reject immediately if any of the promises reject. It might be better to instead use `Promise.allSettled`.

packages/mongo/oplog_observe_driver.js
```diff
@@ -525,37 +525,30 @@ Object.assign(OplogObserveDriver.prototype, {
...
+                if (err) {
+                  Meteor._debug('Got exception while fetching documents', err);
+                  // If we get an error from the fetcher (eg, trouble
+                  // connecting to Mongo), let's just abandon the fetch phase
+                  // altogether and fall back to polling. It's not like we're
+                  // getting live updates anyway.
+                  if (self._phase !== PHASE.QUERYING) {
+                    self._needToPollQuery();
+                  }
+                  reject(err);
```
zodern: I'm not very familiar with this code, but I think this needs to resolve instead. It matches the existing behavior, and it prevents errors from being logged twice (lines 540 and 579).

gbuerk: That's fine, we can resolve here.  I'll update it.

packages/mongo/oplog_observe_driver.js
gbuerk: That makes sense, I think.  That way all the docs that don't error will still `_handleDoc`.  Aren't we just going to re-poll query if anything fails though?

packages/mongo/oplog_observe_driver.js
```diff
@@ -565,20 +558,26 @@ Object.assign(OplogObserveDriver.prototype, {
...
-            })
-          );
+              })
+            )
+          })
+          fetchPromises.push(fetchPromise);
         });
-        await awaitablePromise;
+        // Wait for all fetch operations to complete
+        try {
```
italojs, line 576: Using `Promise.allSettled(fetchPromises)` inside a try...catch block does not handle errors correctly because `Promise.allSettled()` never rejects. 
Instead, it always resolves with an array of results containing `{ status: "fulfilled", value }` for successful promises 
and `{ status: "rejected", reason }` for failed ones. As a result, the catch block will never be triggered, 
even if some of the promises fail.

To properly handle errors and log only the failed requests, you should iterate over the results and filter out the rejected ones.
Here’s an improved approach:

```js
const results = await Promise.allSettled(fetchPromises);

const errors = results
  .filter(result => result.status === 'rejected')
  .map(result => result.reason);

if (errors.length > 0) {
  console.error('Some fetch queries failed:', errors);
}

**Review by italojs — CHANGES_REQUESTED** [review]

File: packages/mongo/oplog_observe_driver.js (hunk 3 above)
Comment (gbuerk) [reply packages/mongo/oplog_observe_driver.js:576]: Updated

File: packages/mongo/oplog_observe_driver.js [hunk 4]
```diff
@@ -565,20 +558,33 @@ Object.assign(OplogObserveDriver.prototype, {
...
+          const errors = results
+            .filter(result => result.status === 'rejected')
+            .map(result => result.reason);
+
+          if (errors.length > 0) {
+            Meteor._debug('Some fetch queries failed:', errors);
+          }
+        } catch (err) {
+          Meteor._debug('Got an exception in a fetch query', err);
+        }
```
Grubba27, line 587: As @italojs mentioned I think we can safely remove this catch block as our errors are being caught in the `errors.length > 0` branch

Issue #12902, opened 2023-11-27
[Meteor 3 Alpha] MINISAT-out: Cannot enlarge memory arrays


OlesZadorozhnyy:
Hi,
After trying to install Alpha version of Meteor 3 it raises an error
```meteor update --release 3.0-alpha.19``` (other alpha versions as well

```
MINISAT-out: Cannot enlarge memory arrays. Either (1) compile with -s TOTAL_MEMORY=X with X higher than the current value 67108864, (2) compile with ALLOW_MEMORY_GROWTH which adjusts the size at runtime but prevents some optimizations, or (3) set Module.TOTAL_MEMORY before the program runs.
MINISAT-err: Cannot enlarge memory arrays. Either (1) compile with -s TOTAL_MEMORY=X with X higher than the current value 67108864, (2) compile with ALLOW_MEMORY_GROWTH which adjusts the size at runtime but prevents some optimizations, or (3) set Module.TOTAL_MEMORY before the program runs.
abort() at Error                              
    at jsStackTrace (packages/logic-solver.js:21:18626)
    at stackTrace (packages/logic-solver.js:21:18809)
    at abort (packages/logic-solver.js:51:28956)
    at enlargeMemory (packages/logic-solver.js:21:19142)
    at Function.dynamicAlloc [as alloc] (packages/logic-solver.js:21:7927)
    at _sbrk (packages/logic-solver.js:21:58803)
    at Sd (packages/logic-solver.js:25:98389)
    at Ud (packages/logic-solver.js:25:109800)
    at gc (packages/logic-solver.js:25:41378)
    at pc (packages/logic-solver.js:25:46259)
    at _b (packages/logic-solver.js:25:31709)
    at $b (packages/logic-solver.js:25:35130)
    at Bc (packages/logic-solver.js:25:54870)
    at hd (packages/logic-solver.js:25:84479)
    at MiniSat.solveAssuming (packages/logic-solver.js:156:18)
    at Logic.Solver.solve (packages/logic-solver.js:1680:33)
    at Logic.Solver.solveAssuming (packages/logic-solver.js:1713:21)
    at minMaxWS (packages/logic-solver.js:1999:32)
    at packages/constraint-solver.js:1381:22
    at CS.Solver.minimize (packages/constraint-solver.js:1365:3)
    at CS.Solver.minimize (packages/constraint-solver.js:1359:5)
    at CS.Solver._getAnswer (packages/constraint-solver.js:1749:3)
[...]

PR #13604, opened 2025-02-03
feat(logic-solver): ⚡️ recompile minisat with ALLOW_MEMORY_GROWTH option

cunneen:
Recompile minisat to address #12902 (`MINISAT-out: Cannot enlarge memory arrays`), which [is still being reported](https://forums.meteor.com/t/meteor-update-failing-with-cannot-enlarge-memory-arrays/62740)


zodern:
I noticed this also switches to using wasm. Have you seen any performance differences with the new build?

nachocodoner:
@zodern: Will this affect just the bundler, or the Meteor runtime as well? It feels to me the bundler process.

I’d like to know if it’s worth running more tests or benchmarks on the Meteor runtime to check if this impacts previous results. We haven’t measured the bundler metrics yet.

zodern:
@nachocodoner unless if an app is using the logic-solver package, it will only affect the Meteor tool. `minisat` is used by many commands when deciding which packages are used by the app, and what version of each package to use. The package is also published to [npm](https://www.npmjs.com/package/logic-solver), which should be updated.

 With the smaller build and using wasm, I would expect it to be faster.

One option for testing might be to have both builds in the package. It could run with both builds and verify the result is the same, or crash if they are not. You could run that with Meteor's tests or other apps, or possibly in a beta version.

nachocodoner:
[I forked your branch to the Meteor core to solve code style check and test](https://github.com/meteor/meteor/pull/13608). I recompiled the minisat.js file to understand the process and verify the contents [following your changes here](https://github.com/meteor/minisat/pull/1). I also checked the area @zodern mentioned, focusing on how this change affects package retrival with `meteor list` on both the devel branch and the branch with the new recompilation. Process runs similarly. CI self-test checks showed no issues, which further confirms this works well. I believe it's safe to include in one of the next Meteor releases.

image

If I get another green light, I’ll include this in 3.1.2 official. Otherwise, it will go in the next beta of another patch or minor release.

cunneen:
@nachocodoner & @zodern , sorry for not responding sooner.

@zodern : I too was hoping for a performance boost, but (like @nachocodoner ) the processing time of the Tinytest results was almost identical across the old and the new minisat. I didn't run a proper performance benchmark, but I did discover an environment variable `CONSTRAINT_SOLVER_BENCHMARK` which enabled some of the slower tests which otherwise don't normally get run.

I just ran:

```console
$ CONSTRAINT_SOLVER_BENCHMARK=1 \
    meteor test-packages \
        ./packages/constraint-solver \
        ./packages/logic-solver
```

I did get sidetracked by seeing whether I could replace minisat (which appears unmaintained) with [mergesat][1] . While I got it working and passing the tests, the results were disappointing: a consistent increase from ~2300ms to ~3220ms. I'm sure this could be improved upon, but my hopes of a "free" performance improvement via an alternative SAT solver will be set aside for now. If you're curious, see my [feature/mergesat][2] branch in my fork of the meteor/minisat repo.

[1]:https://link.springer.com/chapter/10.1007/978-3-030-80223-3_27
[2]:https://github.com/cunneen/minisat/tree/feature/mergesat

nachocodoner:
[I published `3.2-beta.1`](https://github.com/meteor/meteor/pull/13610) (`meteor update --release 3.2-beta.1`) including this feature. You can already test this feature directly on a release. I will do some testing on the coming weeks.

PR #13606, opened 2025-02-04
Implement #13605 to opt-in to legacy

perbergland:
This PR implements my suggestion in #13605 so that you can opt in to legacy (and default to modern bundle)


nachocodoner (approved):
Great!, thank you for the contribution. This is useful configuration.

PR #13610, opened 2025-02-07
Release 3.2

nachocodoner:
## Still in development

Current released version: 3.2.

To use it, please do a

``` sh
meteor update --release 3.2
```

``` sh
meteor create --release 3.2
```

This beta version includes:

- https://github.com/meteor/meteor/pull/13636
- https://github.com/meteor/meteor/pull/13615 and https://github.com/meteor/meteor/pull/13628
- https://github.com/meteor/meteor/pull/13603
- https://github.com/meteor/meteor/pull/13554
- https://github.com/meteor/meteor/pull/13604

Please, help on test these since they are candidates for 3.2. Provide any feedback to the authors on those PRs.

All the rest PRs included: https://github.com/meteor/meteor/pulls?q=is%3Apr+milestone%3A%22Release+3.2%22


packages/accounts-password/package.js
```diff
@@ -5,11 +5,12 @@ Package.describe({
   // 2.2.x in the future. The version was also bumped to 2.0.0 temporarily
   // during the Meteor 1.5.1 release process, so versions 2.0.0-beta.2
   // through -beta.5 and -rc.0 have already been published.
-  version: "3.0.3",
+  version: "4.0.0-beta320.1",
```
zodern: Why does adding argon2 require a major version increase? The change log doesn't mention any breaking changes.

StorytellerCZ: The original implementation was breaking, but now we can get away with just a feature version since Argon2 is hidden behind a flag.

nachocodoner: Yes, the reason was at the beginning it was a version with major breaking changes. Not anymore, so I will respect a minor version bump.

packages/oauth/package.js
```diff
@@ -37,7 +37,7 @@ Package.onUse(api => {
 });
 
 Npm.depends({
-  'body-parser': '1.19.0',
+  'body-parser': '1.20.3',
```
StorytellerCZ, line 40: If #13323 as whole is not acceptable, then I think we can cherry-pick the part that just removes this dependency as we have this in Express.

nachocodoner: It's not that it's unacceptable, but it involves different changes and context than the contributions we merged for 3.2. We need time to analyze and test the affected parts, including any side effects. Especially since some areas on those changes may not have full test coverage. Deferring some changes helps us focus our efforts.

That PR could be a good candidate for the first beta of 3.3 (or next 3.2.1 patch, undecided yet). As with 3.2, this would give time to everyone for testing, including our final verifications before the official release. But we also need to find time to understand each change and the purpose first.

nachocodoner: Btw, I have a list of the PRs you attach in release milestones, including that one, so we can review and decide what to include in future releases. We plan to do this for each release based on the challenges on those, the testing capacity on us and the community to test before official. Keep flagging issues in milestones for next releases, some areas we may miss to track, and those warnings help a lot.

StorytellerCZ: Yes, feel free to move the issues/PRs to their appropriate release milestones. We should probably jump on a call/stream sometime and do a few together and talked about this so that I can better triage.

PR #13617, opened 2025-02-12
Use `_.template` from `lodash` instead of vulnerable `lodash.template`

julio-rocketchat:
`lodash.template` is a package that has been outdated for 6 years. As seen in the following Snyk page, the latest published version has a high-severity vulnerability (https://security.snyk.io/package/npm/lodash.template) - which is a code injection issue.

Instead of using the standalone `lodash.template` package, we can use the full `lodash` package, which contains no known vulnerability as of now (https://security.snyk.io/package/npm/lodash).

### DIFF
--- a/packages/boilerplate-generator/template.js
+++ b/packages/boilerplate-generator/template.js
@@ -1,4 +1,4 @@
-import lodashTemplate from 'lodash.template';
+import _ from 'lodash';
 
 // As identified in issue #9149, when an application overrides the default
 // _.template settings using _.templateSettings, those new settings are
@@ -6,7 +6,7 @@ import lodashTemplate from 'lodash.template';
 // boilerplate-generator. To handle this, _.template settings that have
 // been verified to work are overridden here on each _.template call.
 export default function template(text) {
-  return lodashTemplate(text, null, {
+  return _.template(text, null, {
     evaluate    : /<%([\s\S]+?)%>/g,
     interpolate : /<%=([\s\S]+?)%>/g,
     escape      : /<%-([\s\S]+?)%>/g,
--- a/packages/boilerplate-generator/package.js
+++ b/packages/boilerplate-generator/package.js
@@ -1,11 +1,11 @@
 Package.describe({
   summary: "Generates the boilerplate html from program's manifest",
-  version: '2.0.0',
+  version: '2.0.1',
 });
 
 Npm.depends({
   "combined-stream2": "1.1.2",
-  "lodash.template": "4.5.0"
+  "lodash": "4.17.21"
 });
 
 Package.onUse(api => {


nachocodoner:
The trend in utilities for core and packages has been to add single dependencies as needed, avoiding full packages like Lodash with unused utilities. That said, I'm unsure of the best alternative, maybe cherry-picking the template function or finding a more up-to-date option for security.

packages/boilerplate-generator/template.js
```diff
@@ -1,12 +1,12 @@
...
+import _ from 'lodash';
 
 // As identified in issue #9149, when an application overrides the default
 // _.template settings using _.templateSettings, those new settings are
 // used anywhere _.template is used, including within the
 // boilerplate-generator. To handle this, _.template settings that have
 // been verified to work are overridden here on each _.template call.
 export default function template(text) {
-  return lodashTemplate(text, null, {
+  return _.template(text, null, {
```
nachocodoner, line 9: Even if we use `import template from "lodash/template";` here. I think the entire lodash is part of the build when adding it via `Npm.depends` on a meteor atmosphere package, affecting build time. Not sure if all will be in the bundle though, maybe not.

julio-rocketchat: I think the same. Even if we try to import from 'lodash/template', it may still pull the whole package.

julio-rocketchat:
> The trend in utilities for core and packages has been to add single dependencies as needed, avoiding full packages like Lodash with unused utilities. That said, I'm unsure of the best alternative, maybe cherry-picking the template function or finding a more up-to-date option for security.

Unsure what would be the best alternative here. It seems like they aren't planning on fixing the vulnerability for the standalone `lodash.template` and they recommend avoiding the per method packages themselves: https://lodash.com/per-method-packages. They say: "However, use of these packages is discouraged and they will be removed in v5."

nachocodoner:
I'm not sure what the best option is here. The one I dislike least is copying just the template utility from the latest, most secure lodash, either as our own npm package or a file within Meteor package utilities for core reuse.

Importing the entire lodash clearly goes against the community's efforts to minimize Meteor's core footprint and rely on on-demand functions or native implementations. I'd like to hear @harryadel or @StorytellerCZ's thoughts on this as they worked on cleaning tasks on size savings.

harryadel:
I believe there's an underlying theme to @julio-rocketchat PRs as of late, You can do us a better one and help integrate Snyk or any other packages validator instead of these single PRs. And this would help future proof for any upcoming packages problems. 

As for this PR, adding the whole lodash package would increase the package size immensely, if push comes to shove, we can pull out the code we need from lodash package (if lodash.template is still vulnerable).

julio-rocketchat:
> I believe there's an underlying theme to @julio-rocketchat PRs as of late, You can do us a better one and help integrate Snyk or any other packages validator instead of these single PRs. And this would help future proof for any upcoming packages problems. 
> 
> 
> 
> As for this PR, adding the whole lodash package would increase the package size immensely, if push comes to shove, we can pull out the code we need from lodash package (if lodash.template is still vulnerable).

Not sure if you guys have Dependabot enabled already, but that's what I would recommend doing since it's built into Github and it's easy to use. Dependabot automatically opens PRs to update vulnerable and outdated packages. SCA tools like Snyk and Dependabot, however, may not be able to identify and update all of the vulnerable dependencies.

My recent PRs here are part of an effort we're doing at Rocket.Chat to update our own dependencies. Once our security team identifies that we've got an outdated dependency that is actually part of a Meteor package, we open the PRs to raise awareness and update them.

StorytellerCZ:
Dependabot is enabled, it just can't handle Meteor's formatting for the most part to detect these issues. Snyk has similar issues.
I do have this PR: #12813
I think we could put the template function there. I'm more for just getting the function we need rather than including the entire package. But that does mean that someone needs to monitor the original for any security patches and remember to copy it over. So it needs to be well commented where it comes from and what was the version imported from.

nachocodoner:
@julio-rocketchat: After the feedback, the best approach seems to be cherry-picking the template helper from the latest secure Lodash version and adding it to boilerplate-generator, as it's the only core code using it.

If we later have more core helpers or similar cases, we could either use a `meteor` package and expose it as `Meteor._<helper>` or create a `meteor/utilities` package exposing those. For now, keeping it in the single package where it's used should be enough.

If you do this we can include this security fix on next Meteor 3.2 version.

italojs:
closing due https://github.com/meteor/meteor/pull/13898

PR #13618, opened 2025-02-12
FEATURE: add `shouldShowTime` to `logging` package

Grubba27:
Relates to [this forum request](https://forums.meteor.com/t/suggestions-for-logging-in-galaxy/63119/6?u=grubba).

Those that use `colored-text` output format can now decide if they want the timestamp or can rely on their server/cloud environments info.

### DIFF
--- a/packages/logging/logging.js
+++ b/packages/logging/logging.js
@@ -47,7 +47,7 @@ Log.outputFormat = 'json';
 // Defaults to true for local development and for backwards compatibility.
 // for cloud environments is interesting to leave it false as most of them have the timestamp in the console.
 // Only works in server with colored-text
-Log.shouldShowTime = true;
+Log.showTime = true;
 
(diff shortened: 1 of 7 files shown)


packages/logging/logging.js
```diff
@@ -44,6 +44,11 @@ Log._intercepted = () => {
 // other process that will be reading its standard output.
 Log.outputFormat = 'json';
 
+// Defaults to true for local development and for backwards compatibility.
+// for cloud environments is interesting to leave it false as most of them have the timestamp in the console.
+// Only works in server with colored-text
+Log.shouldShowTime = true;
```
nachocodoner: My only comment is that we usually don’t use `should*` for option names. Instead, we prefer imperative names like `showTime` or `enableTime`. That said, using `should*` is completely fine.

nachocodoner (approved):

PR #13619, opened 2025-02-12
Cordova patch for Meteor 2.x

nachocodoner:
Context: https://forums.meteor.com/t/error-building-cordova-project/63140

Existing workaround so far: https://forums.meteor.com/t/error-building-cordova-project/63140/11

## Pending

- [x] Add patch behavior on add-platform
- [ ] Add patch behavior on run platform
- [ ] Test all on ios
- [ ] Test all on android
- [ ] Publish a beta for 2.16.1 and ask for help to test it

### DIFF
--- a/tools/cordova/patches/cordova-common@FileUpdater.js
+++ b/tools/cordova/patches/cordova-common@FileUpdater.js
@@ -1,0 +1,371 @@
+/**
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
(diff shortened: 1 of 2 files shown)


tools/cordova/project.js
```diff
@@ -461,6 +494,41 @@ from Cordova project`, async () => {
     });
   }
 
+  patchCordova() {
+    this.runCommands(`patching Cordova project`, async () => {
```
Grubba27: I do not think this `async` is necessary

tools/cordova/project.js
```diff
@@ -461,6 +494,41 @@ from Cordova project`, async () => {
...
+      await cmd();
+    } catch(e) {
+      if (e.message.includes('fs.cpSync is not a function')) {
+        try {
+          captureLogBuffer();
+          this.patchCordova();
+          await cmd();
+        } catch(e) {
+          // ignore
+        }
```
Grubba27: Could we log this error? if not, why can't we log it? Is it noise when it fails on this `catch branch`?

tools/cordova/patches/cordova-common@FileUpdater.js
Grubba27, line 1: 😨

tools/cordova/project.js
nachocodoner: Going to test further, but it showed an error saying it was already added instead of simply confirming. Maybe I can ignore the "already added" error and accept every other error that can happen.

tools/cordova/patches/cordova-common@FileUpdater.js
nachocodoner: Any suggestion? Maybe I could cherry pick just the new changes and replace just that, but for simplificty I added the whole file.

nachocodoner:
Closing in favor of a better solution for this, https://github.com/meteor/meteor/pull/13620

PR #13621, opened 2025-02-14
Release 2.16.1

nachocodoner:
## Still in development

Current released version: 2.16.1-beta.0

To use it, please do a

``` sh
meteor update --release 2.16.1-beta.0
```

``` sh
meteor create --release 2.16.1-beta.0
```

## Includes

- Provide a fix on Cordova since was completely failing from Meteor 2.14+, https://github.com/meteor/meteor/pull/13620

### DIFF
--- a/docs/generators/changelog/versions/2.16.1.md
+++ b/docs/generators/changelog/versions/2.16.1.md
@@ -1,0 +1,36 @@
+````## v2.16.1, <date>
+
+### Highlights
+
+- Fix Cordova after the latest breaking change in common dependencies. [PR](https://github.com/meteor/meteor/pull/13620)
+
+#### Migration Steps
+
+To update from 2.16.0 to this one, you can run:
+
+```
+meteor update --release 2.16.1
+```
+
+If you're coming from an older version, please check our [Migration Guides](https://guide.meteor.com/2.14-migration).
+
+#### Breaking Changes
+
+N/A
+
+#### Internal API changes
+
+N/A
+
+#### Meteor Version Release
+
+* meteor-tool@2.16.1
+
+#### Independent releases
+
+N/A
+
+#### Contributors
+
+- [GH nachocodoner]
+````
--- a/docs/history.md
+++ b/docs/history.md
@@ -8,7 +8,41 @@
 
 [//]: # (go to meteor/docs/generators/changelog/docs)
 
+## v2.16.1, <date>
 
+### Highlights
+
+- Fix Cordova after the latest breaking change in common dependencies. [PR](https://github.com/meteor/meteor/pull/13620)
+
+#### Migration Steps
+
+To update from 2.16.0 to this one, you can run:
+
+```
+meteor update --release 2.16.1
+```
+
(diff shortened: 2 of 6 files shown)


boomfly:
Cordova build work.

But on apple m2 without rosette show this:
```sh
This project uses Meteor 2.16.1-beta.0, which isn't available on this platform. To work with this app on all supported platforms, use meteor update --release METEOR@3.1.2 to pin this app to the newest compatible release.
```

And work only with:
```sh
arch -x86_64 zsh
```

but very slow.

Please add arm64 version.

nachocodoner:
@boomfly: Meteor 2 has never supported arm64; this was only introduced in Meteor 3. Backporting support to Meteor 2 is not a trivial change. As with this Cordova issue, proper migrations to newer versions are needed to unlock new architectures and tools.

We fixed this Cordova problem, but we can't guarantee future fixes will be as "straightforward". It's best to start considering the migration to Meteor 3, for both proper native experience and OS-compatibility.

Thanks for testing this fix. The beta will be ready soon for others to check out.

boomfly:
This is strange because other versions 2.10+ work great without Rosetta. I encountered this error: ```This project uses Meteor 2.16.1-beta.0, which isn't available on this platform.``` only with version 2.16.1-beta.0.

nachocodoner:
Which sequence of commands you use to get that error?

Trying to understand your flow and environment setup. I also have a Apple machine with M1 to verify, but I can't get that myself. Anyway, about platform support I also asked the team and don't think Meteor 2.x was ever delivered directly to work with arm64, that came to Meteor 3.

boomfly:
Снимок экрана 2025-02-21 в 01 25 21

FIrst start: version 2.16

Second: screenshot

npm start:
```sh
npx dotenvx run -- meteor run --settings settings.json --exclude-archs web.browser.legacy,web.cordova --port 3500
```

Only version 2.16.1-beta.0 requires x86-64. I believe the Meteor Tool download API indicates that this beta version does not support ARM64, which is the issue.

rrtkh:
Hi, any progress with that?
I was testing this version in my project but it looks beta build breaks my automatic tests in selenium in comparing to version 2.16. I don't see any changes that could make this issue in this pull request. Is there any difference between beta bundle and main release bundle?

nachocodoner:
@boomfly: Weird, I'll check my Mac ARM setup to understand the issue better, as I’m not experiencing it.

You could try completely removing Meteor globally and reinstalling it (`rm -rf ~/.meteor`). Reinstalling with `npx meteor@latest` will install the latest Meteor 3 version with ARM support. However, if you run version 2.16.1, it will download a different version, but I wonder if that affects compatibility for good. Let me know if that helps.

nachocodoner:
@rrtkh: The issue is identified and fixed, and from the feedback, it seems to work except for the ARM problem mentioned above. We need to address this. Regarding the issue with your CI environment, what OS and architecture does your CI run on? Do you use any cache? Have you tried clearing the cache for the new builds with the new version? Any log messages you can share with us would help. More details would let us assist you better.

vijay-bustle (approved):

PR #13628, opened 2025-02-18
Support async profiler actions

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13615 and https://github.com/meteor/performance/pull/9

## Issue

There are scenarios where profiler can't get properly the timing due to the lack of support of async flows.

image

## Fix

image

## Pending issue

Server profiling outputing 0ms suddenly

image

### DIFF
--- a/tools/static-assets/server/boot.js
+++ b/tools/static-assets/server/boot.js
@@ -500,7 +500,7 @@ var runMain = Profile("Run main()", async function () {
 
   await Profile.run('Server startup', async function() {
     const asyncLocalStorage = global.__METEOR_ASYNC_LOCAL_STORAGE;
-    let existingStore = asyncLocalStorage.getStore();
+    const existingStore = asyncLocalStorage.getStore();
     return global.__METEOR_ASYNC_LOCAL_STORAGE.run(existingStore || {}, async () => {
       await loadServerBundles();
       await callStartupHooks();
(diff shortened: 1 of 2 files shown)


zodern (approved):
This looks much better.

I'm not sure storing the entry in the store is working as intended. Since it usually uses the same store object and mutates it, all of the context's would have the same currentEntry array. It's not any worse than how it was with Meteor 2, but as more things run in parallel this could result in profile entries having the wrong parent.

The issue with the profiler not working for the app startup could be caused by conflicts with using the same async local storage as other parts of Meteor. There's a few places where Meteor overrides the value with a new object, including during the boot process.

zodern:
This is probably the code causing the issue with the server startup:
https://github.com/meteor/meteor/blob/94c2a94fa0a535f32318bbdbd9c2676dbf429f0b/tools/static-assets/server/boot.js#L501-L507

I've never understood why it sets the store value here.

Grubba27 (approved):

nachocodoner:
@zodern: Yes, the issue was that during the "Server startup" phase, the async context was re-initialized, losing the existing context. 

Instead of removing it, I now ensure any existing context is passed down (as when the profiler is enabled). If you think we don't need it, I can remove it entirely. I tested both approaches, and all tests pass and profiler output everything well. I'm just not sure if it was originally added for a reason, maybe to ensure an empty context when the profiler is disabled. But no breaking flow seen by doing it anyway.

zodern:
> If you think we don't need it, I can remove it entirely.

I think it can be removed. I've never understood why it was there, or seen anywhere that uses it.

PR #13630, opened 2025-02-19
Wait on Mongo replication of hashedToken; before closing connection

rikvdlooi:
We have our Mongo database set up with read replication in a different continent. Users on that continent were sometimes experiencing issues when logging in. Meteor would close the connection.

We found that the root cause was in the accounts-base package.

It adds an observer, which listens to the hashedToken being added. If the token is added the `foundMatchingUser` is set to true. If foundMatchingUser is false however it closes the connection.

This foundMatchingUser is immediately evaluated, sometimes not giving Mongo enough time to replicate the write of the hashedToken to the other continent.

Now I just simply added a 1 second wait before evaluating foundMatchingUser.

p.s. We're using Meteor 2.16, and I branched of from release-2.16 because of that.

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -1014,14 +1014,18 @@ export class AccountsServer extends AccountsCommon {
 
         this._userObservesForConnections[connection.id] = observe;
 
-        if (! foundMatchingUser) {
-          // We've set up an observe on the user associated with `newToken`,
-          // so if the new token is removed from the database, we'll close
-          // the connection. But the token might have already been deleted
-          // before we set up the observe, which wouldn't have closed the
-          // connection because the observe wasn't running yet.
-          connection.close();
-        }
+        Meteor.setTimeout(() => {
+          if (! foundMatchingUser) {
+            // We've set up an observe on the user associated with `newToken`,
+            // so if the new token is removed from the database, we'll close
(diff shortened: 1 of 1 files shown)


rikvdlooi:
I see the netlify v3 meteor stuff is failing. Maybe that's because I branched of from Meteor 2. I could also make this PR based on Meteor 3; however it would be nice if we could get this fix on v2 as well :)

italojs:
@rikvdlooi do you mind opening another pr for 3.x?
the proposal looks good to me, since the CI is green we can merge it
thanks for the contribution, you rock it!

StorytellerCZ:
I'm not really happy with an arbitrary wait time, especially when many apps won't need it. So probably the first thing is that this should be configurable.
Another approach would be to determine if the MongoDB topology requires it, but that seems like too much work for little gain unless we can find a better solution than just a wait.

osrl:
I've developed a custom DDPClient for Kotlin Multiplatform and everything was working fine until we switched to redis-oplog. Maybe some subscription delays or something else causes this. Our socket connection is closed because that `foundMatchingUser` is always false for the first login. The DDPClient implementation on web can reconnect and try to login with resume. Now I will try to find a way to replicate that in my DDPClient implementation. But this bug causes a duplicate socket connection for the logins. That would be great if you merge this PR until you find a better solution than just a wait.

PR #13634, opened 2025-02-24
Update 7.adding-user-accounts.md

Davidihl:
use `insertAsync` as introduced in Step 2 instead of `insert`.
Blindly copy pasting the code will result in an unnecessary error for new developers.


Grubba27 (approved):
nice catch! thanks a lot!

PR #13636, opened 2025-02-25
Meteor bundler profile command

nachocodoner:
Continues: https://github.com/meteor/performance/pull/9

This PR introduces the `meteor profile` command, designed to help Meteor users assess how changes in their app and tool versions impact development time and bundle size. It also supports the core team in improving the Meteor bundler and exploring modern bundler options. By providing this tool to the community, we aim to gather feedback and measure improvements on their experiences.

The initial implementation includes:

- Reporting the time metrics for your Meteor app during cold start, cache start, client rebuild, and server rebuild.
- Reporting the time to add bundle-visualizer and run the app in production mode, along with the size of Meteor app bundles and packages (using the `METEOR_BUNDLE_SIZE=true` env).
- Support only on UNIX-like systems

## Pending

- [x] Better benchmark process error handling
- [x] Better dev experience on command feedback
- [x] Better naming `meteor profile` instead of `meteor benchmark`

### Meteor bundler profiling

image

### Meteor bundler size

image

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -73,21 +73,15 @@ const bash =
  */
 const runLiveCommand = (command, args = []) => {
   return new Promise((resolve, reject) => {
-    const childProcess = spawn(command, args, { shell: true, env: process.env });
+    const childProcess = spawn(command, args, {
+      shell: true,
+      env: { ...process.env, FORCE_COLOR: "1", TERM: "xterm-256color" },
+      stdio: "inherit",
+    });
 
     const cleanup = () => {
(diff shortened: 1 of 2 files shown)


Grubba27 (commented):
Currently looks good...

🏌️ just a nit:

- We could have more info logging about what it is doing(we can use colors if we want but maybe letting the user know what is going on/what step we are could be nice)
  - like, we are cloning, cloned, starting the benchmark process, done (if we are some other place is ok)

Grubba27:
I think we forgot to update the You mean we missed updating https://docs.meteor.com/cli/ as well

PR #13642, opened 2025-03-06
Remove deprecated allow async from mongo tests

StorytellerCZ:
Removes deprecated async allow/deny methods from mongo tests. This speeds things up a bit as there is no need to write about the deprecation in the console. Also since they are deprecated we should be moving to remove them from the core codebase as much as possible.

It might be best to shelve this PR until we actually remove the async allow/deny methods.

Currently work in progress as few still remain.


italojs:
Amazing
 I suggest converting this PR to a draft while we still have not removed the methods from the core

PR #13644, opened 2025-03-10
Add Jam's packages to docs (community-packages)

StorytellerCZ:
Right now just copied info from readmes with some light edits.
This is it for now from Jam's packages.

@Grubba27 please check that I have added things correctly. If all fine, then I will proceed with the rest of Jam's amazing packages and then with MCP packages.

@jamauro


jamauro:
wow thank you @StorytellerCZ! I really appreciate it.

What are y'all's thoughts on having such extensive community package documentation in the Meteor docs site?

Maybe it'd be better to simply link to the package's readme, or if you wanted to get fancy, ingest the readme from the source link every so often. Seems like it'd be best to have one source of truth for a package's docs and I imagine that will most likely be contained in the repo readme. 

One other thought: assuming this grows as the main way to see all community packages, it’d be nice if there was one view to filter through them by tags related to meteor core concepts, frontend framework, author username e.g. `minimongo`, `methods`, `pubsub`, `svelte`

v3-docs/docs/community-packages/jam-method.md
```diff
@@ -0,0 +1,576 @@
+# Jam Method
+
+- `Who maintains the package` – [Jam](https://github.com/jamauro)
+
+[[toc]]
+
+## What is this package?
+
+Method is an easy way to create Meteor methods with Optimistic UI. It's built with Meteor 3.0 in mind and is compatible with Meteor 2.x to make migration easy. It's meant to be a drop in replacement for Validated Method and comes with additional features:
```
italojs: would be better to use 3.x

italojs (changes requested):

italojs:
@jamauro loved the ideas, especially the autoloader from readme.
Please open a discussion [here](https://github.com/meteor/meteor/discussions), and let's discuss the best approach for it.

StorytellerCZ:
I have just put in the categories very loosely. The idea here (from my understanding) is to put info here on the most used or most useful packages from the community. So if it is mentioned in the guide and beyond. I have chosen @jamauro packages as they provide a significant utility. Also they are already well documented, so I can just copy info from their readme, but other packages which are less documented can have an expanded documentation here.
We will have to probably go through few iterations here to figure out the proper format.

v3-docs/docs/community-packages/jam-method.md
StorytellerCZ: Fixed by removing the feature numeral altogether.

jamauro:
Thanks again @StorytellerCZ. Would you mind adding [jam:easy-schema](https://github.com/jamauro/easy-schema) to this PR? Otherwise I can open another PR.

StorytellerCZ:
@jamauro I will create a dedicated schema PR later.

Grubba27:
> What are y'all's thoughts on having such extensive community package documentation in the Meteor docs site?

The main idea is to have some community+meteor team approval stamp for some packages(e.g, these are _GREAT_ packages, and if it matches your use case, use it)


> Maybe it'd be better to simply link to the package's readme, or if you wanted to get fancy, ingest the readme from the source link every so often. Seems like it'd be best to have one source of truth for a package's docs and I imagine that will most likely be contained in the repo readme.

I did not want to bring much complexity from the beginning, but I guess when this gets big enough, we can do this kind of change(fetching from the gh readme + some extra info)

Grubba27 (dismissed):
If @jamauro thinks all looks good–on my end all looks good

nachocodoner:
I specifically want these packages included in the docs. They’re great!

I agree that having a single source of truth is better than duplicating information, even though other approach has its complexity. In the long run, it helps keep things up to date and avoids outdated information, soemthing really easy to happen.

That said, we can accept it as it is. I’d just add a link at the beginning or end of each package page to its original GitHub repository, so it serves as a reference for more details, collaboration, and other purposes.

jamauro:
Thanks y'all! Looks good. 

Agree with Nacho's thoughts above.

StorytellerCZ:
Will add the links.

nachocodoner (approved):

italojs (approved):

PR #13646, opened 2025-03-11
Update accounts.md

trusktr:
Add a small detail about the clientStorage:session option.


StorytellerCZ (approved):
Well duh. 😂
Still good to spell it out for those who are too eager and haven't studied this.

PR #13656, opened 2025-03-17
Address `meteor profile` feedback

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13636 and https://github.com/meteor/performance/pull/12

This PR is latest tweaks for `meteor profile` cmd for the official 3.2 version.

- Added `--size` and `--size-only` options to meteor profile for runtime and size analysis or just bundle size.
- Introduced `METEOR_CLIENT_ENTRYPOINT` and `METEOR_SERVER_ENTRYPOINT` env vars to control Meteor endpoints.
- Added METEOR_LOG_DIR env var to set logs directory.
- Improved formatting for package and size output using `console.table`.
    - See the new [meteor packages](https://github.com/user-attachments/assets/3992c521-398a-490d-a7cd-481c59871923) and [size analysis](https://github.com/user-attachments/assets/58e7060d-b8e3-4b48-a8ef-0fe0de91beb0) output.

- Updated Meteor 3 docs to include the new command.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -3390,7 +3390,7 @@ async function doBenchmarkCommand(options) {
     !!options['only-size'] && 'METEOR_BUNDLE_SIZE_ONLY=true',
     !!options['size'] && 'METEOR_BUNDLE_SIZE=true'
   ].filter(Boolean);
-  const meteorOptions = args.filter(arg => ['--only-size', '--size'].includes(arg));
+  const meteorOptions = args.filter(arg => !['--only-size', '--size'].includes(arg));
 
   const profilingCommand = [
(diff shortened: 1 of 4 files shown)


tools/cli/commands.js
```diff
@@ -1767,6 +1767,7 @@ main.registerCommand({
   maxArgs: 1,
   options: {
     db: { type: Boolean },
+    'skip-cache': { type: Boolean },
```
Grubba27, line 1770: Have we documented this option for `skip-cache`?

Grubba27 (approved):

tools/cli/commands.js
nachocodoner: I thought about it, but it's mainly for internal use. The profile script prevents repeated clearing of the performance repository context. Cache clearing always happens with `meteor reset`, and I can't see a case where a dev wouldn't want that when used intentionally. Many commands in this file have undocumented options for internal use.

PR #13657, opened 2025-03-19
SWC adoption as bundler transpiler

nachocodoner:
Issue: https://github.com/meteor/meteor/discussions/11587

This PR focuses on improving the Meteor bundler by adopting a modern transpiler over the currently used Babel.

[Profiling reports with `meteor profile`](https://github.com/meteor/meteor/pull/13636) helped analyze performance at each stage while modifying the bundler. Based on these insights, I experimented with using SWC as the default transpiler. If SWC fails for any reason, whether due to a specific file, package, or error, the system falls back to Babel.

Two key parts of the code have been identified for optimization:  

- [`Babel.compile`](https://github.com/meteor/meteor/blob/2d9cd4f1a06163f4c9d40abf9e317d3850fb2b2f/packages/babel-compiler/babel-compiler.js#L147-L149)  
- [`jsAnalyze.parse`](https://github.com/meteor/meteor/blob/2d9cd4f1a06163f4c9d40abf9e317d3850fb2b2f/tools/isobuild/js-analyze.js#L30-L45)  

Parsing and compiling each file in a package adds significant milliseconds, depending on file size and count. Switching to a faster transpiler like SWC improves this process.

> This change aligns with standards and improves Meteor's performance. It's opt-in for now but may become the default later. We’re unsure if Meteor-specific compilation, like nested imports, can be preserved. As that pattern moves further from standard practices, we may deprecate it or require a specific configuration to continue using it as you do now, getting away from performance gains. Your app code and packages may need to stop using nested imports to adopt this approach fully.

## Last Results

The results showed improved performance with all changes in this PR across multiple build and rebuild stages on the development mode using `meteor run`.

image

`meteor build`, used to create the production artifact, will also benefit in the cold phase, meaning your `meteor deploy` process or any production deployment steps will see the improvements. In the next results, you’ll notice these effects. Additionally, using the [zodern/minify-js-sourcemaps](https://github.com/zodern/minify-js-sourcemaps) package not only enhances the production sourcemaps, but also applies the SWC minifier, giving you an extra performance boost on `meteor build`.

image

> The values reflect performance gains on a small app built with `meteor create --react`. The improvements will become more noticeable in larger, feature-rich applications.

## Meteor SWC configuration

I added configuration options to enable or exclude specific parts of the app, packages, or files from SWC transpilation. This should be very helpful for the ongoing SWC migration for a number of reasons.

Some apps rely on Babel-specific plugins, so having the ability to keep their app code using Babel, or to selectively exclude certain app files, can help developers get the most performance gain, especially for files that would otherwise default back to Babel. Meteor core and official or community packages are expected to work well with SWC, so everyone will benefit from faster builds, but we still provide the option to opt out as needed.

These options can be set within the `meteor` field of your package.json:

- `modernTranspiler: [true|false]`
Enables or disables the use of the modern transpiler (SWC). If disabled, Babel will be used directly instead.

- `modernTranspiler.excludeApp: [true|false] or [string[]]`
If true, the app’s own code (outside of Meteor core and packages) will continue using Babel.
Otherwise, a list of file paths or regex-like patterns within the app to exclude from SWC transpilation.

- `modernTranspiler.excludeNodeModules: [true|false] or [string[]]`
If true, the app’s node_modules will continue using Babel.
Otherwise, a list of NPM packages names, file paths or regex-like patterns within the node_modules folder to exclude from SWC transpilation.

- `modernTranspiler.excludePackages: [true|false] or [string[]]`
If true, the Meteor’s packages will continue using Babel.
Otherwise, a list of package names, file paths or regex-like patterns within the package to exclude from SWC transpilation. 

- `modernTranspiler.excludeLegacy: [true|false]`
If true, the Meteor’s bundle for legacy browsers will continue using Babel.

- `modernTranspiler.verbose: [true|false]`
If true, the transpilation process for files is shown when running the app. This helps understand which transpiler is used for each file, what fallbacks are applied, and gives a chance to either exclude files to always use Babel or migrate fully to SWC.

## Pending Work

This effort will evolve as testing continues, including:

- [x] Optimize Babel.compile with SWC.transformSync
- [x] Optimize jsAnalyze.parse with acorn, https://github.com/meteor/meteor/pull/13664
- [x] Optimize "Rebuild App" for faster incremental rebuild performance, https://github.com/meteor/meteor/pull/13674
- [x] Fix `await file.hash` to promote cache reusage more frequently for some files
- [x] Ensure HMR is perserved, https://github.com/meteor/meteor/pull/13675
- [x] Optimize using acorn on static-assets server runtime and import scanner
- [x] Ensure top-level await works in nested files or allow a config for specific files where is used it, https://github.com/meteor/meteor/pull/13675
- [x] Improve the profiler to include all stages during rebuild changes and the entire process to help detect file watching delays, https://github.com/meteor/performance/pull/13
- [x] Adding an opt-in configuration to enable SWC boost (beta)
- [x] Enable logs for fallback behavior to increase visibility into the process and identify migration opportunities. Use the verbose option in the Meteor config to toggle these warnings on or off.
- [x] Ensure support of .swcrc files config for devs to complement the config and plugins
- [x] Modern vs legacy bundles. Can legacy be compatible with SWC via `targets` config?
- [x] Add swc in the babel-compiler package using https://github.com/zodern/meteor-package-install-swc
- [x] Fix CI checks
- [x] Ensure several test apps can be built and deployed to production (test dynamic imports and how modules are loaded as well)
- [x] Docs to explain modern transpiler adoption


packages/babel-compiler/babel-compiler.js
```diff
@@ -1,5 +1,6 @@
 var semver = Npm.require("semver");
 var JSON5 = Npm.require("json5");
+var SWC = Npm.require("@swc/core");
```
zodern: How Meteor packages are published isn't compatible with @swc/core since the published package only has a native module for the architecture and os it is published from, and uses the much slower web assembly build on other architectures.

My zodern:standard-minifier-js package also uses swc, so I created an npm package that fixes this: https://github.com/zodern/meteor-package-install-swc. I can add you as a collaborator to the package.

packages/babel-compiler/babel-compiler.js
```diff
@@ -145,7 +146,51 @@ BCp.processOneFileForTarget = function (inputFile, source) {
 
     try {
       var result = profile('Babel.compile', function () {
-        return Babel.compile(source, babelOptions, cacheOptions);
+        let compilation;
+        try {
+          const packagesSkipSwc = [];
+          const fileSkipSwc = ['webapp_server.js']; // top level await
```
nachocodoner: This was skipped from SWC compilation due to top-level await: [link](https://github.com/meteor/meteor/blob/devel/packages/webapp/webapp_server.js#L1558).

Not sure yet if we can ignore or change this, but skipping one file doesn't seem too impactful. Maybe I could isolate this await and related functions in a smaller file, though it's not really important.

packages/babel-compiler/babel-compiler.js
```diff
@@ -145,7 +146,51 @@ BCp.processOneFileForTarget = function (inputFile, source) {
...
+              map: JSON.parse(transformed.map),
+              hash: toBeAdded.hash,
+              sourceType: 'module',
+            };
+          } else {
+            compilation = Babel.compile(source, babelOptions, cacheOptions);
+          }
+        } catch (e) {
+          // If SWC fails, fall back to Babel
+          compilation = Babel.compile(source, babelOptions, cacheOptions);
```
nachocodoner: This doesn't happen actually, everything seems to get compiled with SWC.

packages/babel-compiler/babel-compiler.js
zodern: Have you checked if live bindings work when importing between modules that use babel and swc? I expect SWC and Reify use opposite approaches to live bindings (reify is push based for a better developer experience, while swc is probably pull based which is more performant but annoying for developers when debugging app code).

We might also have similar issues with top level await. Importing a module that uses top level await, or is async makes a module become async so all of the parent modules also need to be compiled for top level await.

We might need to create a swc plugin to handle some of these things. There's not really an alternative for top level await - as far as I know reify is the only implementation that is not for a specific bundler.

We could also ignore top level await in the compiler, and instead have the import scanner handle it. This way the compilers don't have to worry about supporting it, and the import scanner would know which files actually need to be compiled for top level await.  I was working toward redesigning isopack to work this way: https://forums.meteor.com/t/optimize-package-bundler-errors/62664/4?u=zodern

zodern: I wonder how the performance would be if we first compiled with swc and had it leave top level await and ecmascript export/import syntax alone, and then compiled with reify. The reason babel is so slow is we have every babel parser plugin enabled. If SWC already compiled any syntax, reify could use a much faster parser (it doesn't require the babel parser).

packages/babel-compiler/babel-compiler.js
```diff
@@ -145,7 +146,51 @@ BCp.processOneFileForTarget = function (inputFile, source) {
...
     try {
       var result = profile('Babel.compile', function () {
-        return Babel.compile(source, babelOptions, cacheOptions);
+        let compilation;
+        try {
+          const packagesSkipSwc = [];
+          const fileSkipSwc = ['webapp_server.js']; // top level await
+
+          // Determine if SWC should be used based on package and file criteria.
+          const shouldUseSwc =
```
zodern: It should probably also be disabled when there is a babel config.

zodern:
> [jsAnalyze.parse](https://github.com/meteor/meteor/blob/2d9cd4f1a06163f4c9d40abf9e317d3850fb2b2f/tools/isobuild/js-analyze.js#L30-L45)

This is used to identify globals within package code so the linker can make them scoped to the package. I had looked into switching to swc for it a while ago, but it seemed that we would loose most of the performance benefit of the faster parser by the overhead of moving the ast to js. To fully benefit from the performance, we would need to implement the analyzer in rust. I don't know if there's been any improvements since then that would make other approaches viable.

packages/ecmascript/runtime-tests.js
```diff
@@ -169,14 +171,12 @@ Tinytest.add('ecmascript - runtime - classes - properties', test => {
...
       test.equal(oyez, 'oyez');
       test.isTrue(self === this);
       test.equal(this.property, 'property');
     };
 
     method() {
-      import { testExport as oyez } from './runtime-tests.js';
       test.equal(oyez, 'oyez');
     }
   }
```
nachocodoner, line 182: I also noticed that Meteor touches test files from packages, even in runtime dev mode. Maybe we could ignore them by default for watching and compiling. .meteorignore can handle this, but expecting all users to configure it is less effective than ignoring them by default, at least for core packages following a specific filename pattern.

zodern: There's no reason runtime-tests.js can't be imported outside of tests, so Meteor treats it as any other file. Though it should be using a lazy finalizer and not compile it unless it's actually used. It also shouldn't be watched unless it is actually used during the build (you can check by modifying the file - Meteor should not trigger a rebuild).

tools/isobuild/import-scanner.ts
```diff
@@ -977,7 +977,7 @@ export default class ImportScanner {
   private async findImportedModuleIdentifiers(
     file: File,
   ): Promise<Record<string, ImportInfo>> {
-    const fileHash = file.hash;
+    const fileHash = file.hash instanceof Promise ? await file.hash : file.hash;
```
nachocodoner, line 980: Waiting for file.hash is necessary sometimes to ensure the hash prompts cache usage more often. While debugging the bundler, I noticed some file hashes were unresolved promises, causing the cache to be skipped and forcing a full rebuild of affected files all time. This change improved cached stage times by 50-150ms consistently on the app and reference machine.

Shown here

nachocodoner:
> > [jsAnalyze.parse](https://github.com/meteor/meteor/blob/2d9cd4f1a06163f4c9d40abf9e317d3850fb2b2f/tools/isobuild/js-analyze.js#L30-L45)
> 
> This is used to identify globals within package code so the linker can make them scoped to the package. I had looked into switching to swc for it a while ago, but it seemed that we would loose most of the performance benefit of the faster parser by the overhead of moving the ast to js. To fully benefit from the performance, we would need to implement the analyzer in rust. I don't know if there's been any improvements since then that would make other approaches viable.

@zodern: This is what I'm trying now. Yeah, I noticed the same thing you did, swc-to-babel (js implementation) is slow. So it is hard to make a compatible version that can later be traversed using [escope](https://www.npmjs.com/package/escope) to detect globals. Still looking for a better way. The only idea I have is to somehow use the [swc_estree_compat](https://github.com/swc-project/swc/tree/main/crates/swc_estree_compat) Rust implementation inside NodeJS, which seems hard or making Meteor Rust dependent (or maybe using a lib like [napirs](https://napi.rs/)). I don’t see this exposed through the API yet, but the core seems to be there. I just need to adjust this part to get full performance. 🤞

Edit: I am experimenting with the solution in this PR, https://github.com/meteor/meteor/pull/13662

packages/babel-compiler/babel-compiler.js
nachocodoner: Oh, I see. Please add me as a collaborator. It would be good to keep the SWC version up to date and ensure it's used correctly in the packages when published.

nachocodoner: @zodern: I added a PR to update SWC to the latest version. If you could approve that and publish a new version, then we can include it on this PR.

https://github.com/zodern/meteor-package-install-swc/pull/1

packages/babel-compiler/babel-compiler.js
nachocodoner: This was addressed based on the new configuration in Meteor's package.json (check PR description). If you set modernTranspiler, SWC will be used regardless of any Babel config. Babel will only apply to specific exclusions or fallback cases.

packages/babel-compiler/babel-compiler.js
nachocodoner: I published a custom `@meteorjs/swc-core` package to support the latest swc-core version. I didn’t have permission to publish under the original npm package.

v3-docs/docs/about/modern-build-stack/modern-transpiler-swc.md
```diff
@@ -0,0 +1,113 @@
...
+
+Since transpilation is one of the slowest steps in development, Meteor now gives you the option to use SWC for your apps.
+
+## Enable SWC
+
+Add this to your app's `package.json`:
+
+```json
+"meteor": {
+  "modernTranspiler": true
```
Grubba27, line 13: Could we make all new apps come with this ?

v3-docs/docs/about/modern-build-stack/modern-bundler.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Modern builds
+
+Starting with Meteor 3.3, new apps skip `web.browser.legacy` and `web.cordova` by default in development mode (unless developing for native).
+
+For existing apps, enable this by adding to `package.json`:
+
+```json
+"meteor": {
+  "modernWebArchsOnly": true
```
Grubba27, line 15: Same for this one!

Grubba27 (approved):

v3-docs/docs/about/modern-build-stack/modern-bundler.md
nachocodoner: I think this was already applied in another branch.

italojs (approved):

PR #13661, opened 2025-03-21
update `Accounts.onCreateUser` callback return type

trusktr:
The [`Accounts.onCreateUser` docs](https://docs.meteor.com/api/accounts.html#AccountsServer-onCreateUser) mention that the callback must return the user object, or throw, however the type definition had a return type of `void`.

### DIFF
--- a/packages/accounts-base/accounts-base.d.ts
+++ b/packages/accounts-base/accounts-base.d.ts
@@ -188,9 +188,9 @@ export namespace Accounts {
 
   function removeEmail(userId: string, email: string): Promise<void>;
 
-  function onCreateUser(
-    func: (options: { profile?: {} | undefined }, user: Meteor.User) => void
-  ): void;
+  type CreateUserCallback = (options: { profile?: {} | undefined }, user: Meteor.User) => Meteor.User
+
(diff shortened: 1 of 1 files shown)


italojs (changes requested):
LGTM!
Could you please point it to release3.2 branch?

trusktr:
@italojs rebased onto `release-3.2`. Do we have type tests? Upon an initial search I wasn't able to fuzzyfind files for `test.ts` (but I see lots of `test.js` results).

The idea is to add some type tests somewhere to ensure that type definitions work as expected.

Here's an example of type tests in a project of mine:

https://github.com/lume/element/blob/f431740df30efd674a69a14ae1cfcfc905394c8d/src/framework-types/stencil.test.tsx#L102-L186

As you can see there, that test doesn't run any logic, but it'll fail to build if there's a type error (the runtime test itself is basically just a no-op in that example).

I see that the `typescript` package has a `test.ts` example. I'm guessing I would need to add `typescript` to a package's `package.js` to then be able add a `test.ts` file (which will presumably fail to build if any type errors).

italojs:
We havent typo test, maybe I would create an issue to discuss how we can create it before adress it

PR #13662, opened 2025-03-21
jsAnalyze.parse optimization (SWC replacement)

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13657

This PR tries to optimize `jsAnalyze.parse` by using SWC. Since SWC’s AST isn’t compatible with ESTree, we need to ensure it is. For that, a [napi-rs](https://napi.rs/) native build of [swc_estree_compat](https://github.com/swc-project/swc/tree/main/crates/swc_estree_compat) is used to handle parsing and transformation efficiently.

After getting it to work, I noticed it actually made things [slower](https://github.com/user-attachments/assets/6fd781ce-c2d8-4c40-91e2-08f2be77a15a). I’ll debug further to check if I missed something.

This is a WIP PR, still needs cleanup.

---

## Current issue

I see this error on the logs:

```
thread '<unnamed>' panicked at /home/freak2geek/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/swc_estree_compat-11.0.0/src/babelify/expr.rs:163:34:
illegal conversion: Cannot convert OptChain(OptChainExpr { span: 53706..53748, optional: false, base: Call(OptCall { span: 53706..53748, ctxt: #0, callee: OptChain(OptChainExpr { span: 53706..53729, optional: true, base: Member(MemberExpr { span: 53706..53729, obj: Member(MemberExpr { span: 53706..53713, obj: Ident(Ident { span: 53706..53710, ctxt: #0, sym: "args", optional: false }), prop: Computed(ComputedPropName { span: 53710..53713, expr: Lit(Num(Number { span: 53711..53712, value: 0.0, raw: Some("0") })) }) }), prop: Ident(IdentName { span: 53715..53729, sym: "hasOwnProperty" }) }) }), args: [ExprOrSpread { spread: None, expr: Lit(Str(Str { span: 53730..53747, value: "returnStubValue", raw: Some("'returnStubValue'") })) }], type_args: None }) }) to ExprOutput - babel has no equivalent
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
e Error: Conversion failed due to unsupported optional chaining.
```

This means that the swc or swc_estree_compat may not actually support all cases.


nachocodoner:
image

I now think optimizing jsAnalyze.parse on the import scanner is not worth the modification. babel.parse appears configured to perform minimal work when producing the compatible AST later. However, with SWC, even when limiting extra processing like comments and others, the time remains significantly higher than babel's because it also converts the SWC AST to a babel AST.

Unless new ideas arise, I will skip optimizing jsAnalyze.parse and focus on other parts. The SWC replacement for compilation (https://github.com/meteor/meteor/pull/13657) already provides a notable boost.

zodern:
An alternative to using swc is to first parse with a faster js parser like acorn, and it fails fall back to babel. I use this approach in zodern:standard-minifier-js and had a significant performance improvement: 
https://github.com/zodern/minify-js-sourcemaps/blob/3edc58cef2a32ff3c176878ba88ebfe6c6f02c71/standard-minifier-js-sourcemaps/plugin/stats.js#L40-L63

nachocodoner:
Closing in favor of https://github.com/meteor/meteor/pull/13664

Thank you @zodern!

PR #13664, opened 2025-03-22
jsAnalyze.parse optimization (acorn replacement)

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13657 and https://github.com/meteor/meteor/pull/13662

This PR applies @zodern's [suggestion](https://github.com/meteor/meteor/pull/13662#issuecomment-2743938127) to use [Acorn ](https://www.npmjs.com/package/acorn) for faster code parsing in the bundler analyzer. After profiling, it does show a slight speedup during the "Build App" phases (~50-100ms extra for referenced small app and machine) .

image

### DIFF
--- a/tools/isobuild/js-analyze.js
+++ b/tools/isobuild/js-analyze.js
@@ -21,9 +21,11 @@ var AST_CACHE = new LRUCache({
 });
 
 function countLines(str) {
+  let lastIndex = str.indexOf('\n');
   let count = 0;
-  for (let i = 0; i < str.length; i++) {
-    if (str[i] === '\n') count++;
+  while (lastIndex > -1) {
+    count += 1;
+    lastIndex = str.indexOf('\n', lastIndex + 1);
   }
   return count;
 }
--- a/tools/isobuild/js-analyze.js
+++ b/tools/isobuild/js-analyze.js
@@ -16,7 +16,7 @@ function isRegExp(value) {
 var AST_CACHE = new LRUCache({
   max: Math.pow(2, 12),
   length(ast) {
-    return ast.end;
+    return ast.loc.end.line;
   }
 });
 
@@ -37,7 +37,8 @@ function tryToParse(source, hash) {
           allowImportExportEverywhere: true,
           allowReturnOutsideFunction: true,
           allowHashBang: true,
-          checkPrivateFields: false
+          checkPrivateFields: false,
+          locations: true,
         });
       } catch (error) {
         ast = parse(source, {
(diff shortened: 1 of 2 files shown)


tools/isobuild/js-analyze.js
```diff
@@ -15,10 +16,18 @@ function isRegExp(value) {
 var AST_CACHE = new LRUCache({
   max: Math.pow(2, 12),
   length(ast) {
-    return ast.loc.end.line;
+    return ast?.loc?.end?.line || ast?.lines || ast.end;
   }
 });
 
+function countLines(str) {
```
italojs: the split would perform better for long sources

```suggestion
function countLines(str) {
  return str.split('\n').length - 1;
}

```

italojs (changes requested):

tools/isobuild/js-analyze.js
nachocodoner: I didn’t notice that on my profiles, but maybe some side-effect caused higher times in my comparisons. Also, the split approach probably uses more RAM at peak, right?

Maybe we should use a better benchmark tool to compare these functions and check.

zodern: I spent a lot of time trying to optimize counting lines since in the new linker I was working on, it was a significant fraction of the remaining link time. The implementation I ended up with was this: https://github.com/zodern/source-maps/blob/d98e2fa2b40a40ee469478826746f420d36ce651/index.js#L234-L244

Though if you can skip counting lines, that would be much better. It's only used to estimate the size of the AST for the cache. There should be an alternative way to estimate it that is free.

nachocodoner: I'll try the implementation you sent. It makes sense to go faster than the for and split approaches. I'll check seeing several profiles.

About skipping line counts, I thought of that too. That’s why in a specific commit [I used `ast.end` instead](https://github.com/meteor/meteor/pull/13664/commits/d34c094bd82dc5ab3ecb30533c022e5f616e6d31#diff-ecc80ca4cca4b83e95361ab44c6203452e7ba645ae61ea53503ffd574c262b18R19), since it comes from the AST analysis. But it refers to all characters, not lines, so values differ greatly, the cache config would need to be different. The AST_CACHE has caused issues before when tweaking max and length, so I avoided relying on another value. But maybe it’s worth finding another way to re-estimate.

nachocodoner: I've used an approach to estimate cache length per AST, based on a calculation to guess the number of lines. Assuming 40 characters per line seems close enough for most cases. It can vary a lot, but the difference is small across all files in a sample app, so I believe it would be OK to go with.

Some examples:

image

PR #13665, opened 2025-03-23
Add flag to package.json to automatically skip legacy build targets

9Morello:
This PR introduces a new configuration option in `package.json` that allows Meteor applications to automatically exclude legacy web architectures from builds without having to specify them manually through the `--exclude-archs` command-line option.

```js
{
  "meteor": {
    "modernWebArchsOnly": true
  }
}
```

This was inspired by [a forum post](https://forums.meteor.com/t/proposal-poll-included-exclude-archs-becoming-include-archs/61397) from @Grubba27 . I didn't want to to keep adding `--exclude-archs` to my options every time I started worked on a Meteor project, but I prefer a solution that was a non-breaking change to current projects, configs, etc.

The (minor) advantage this flag has over renaming `exclude-archs` to `include-archs`  is that it doesn't require any changes to existing projects. If they already have `--exclude-archs` saved in their `package.json` scripts or whatever, it will keep doing the same thing. New scaffolded projects can include `modernWebArchsOnly` by default.

The reason the flag name is long is because `modernArchsOnly` felt ambiguous by itself - to an user less familiar with Meteor, it could also mean to not build for 32-bit CPUs.

I don't know if this is currently automated, but repos like [skel-react](https://github.com/meteor/skel-react) would have to be updated to include this flag if we want it to be the default. This PR does add it to the skeleton projects in this repo.


StorytellerCZ:
Is this limited to only development? For example I would use this for development, but when deploying I want the legacy bundles as well and that is how I have it currently setup with my scripts.

9Morello:
@StorytellerCZ Adding this flag to a `package.json` prevents production builds from generating legacy bundles as well. That being said, your scripts would work the same way they do today if you don't use it - `exclude-archs` is still respected, and the build output doesn't change if you never set `modernWebArchsOnly` to `true`.

StorytellerCZ:
Not to fond of that. The most common setup, I think, is that you exclude legacy bundles from development build to improve development build times. But for production you want the legacy bundle for Cordova and old browsers so that your app is accessible for most customers just in case. I think this setting should not apply for production build or it should be split into 2 options, one for development and one for production.

tools/cli/commands.js
```diff
@@ -261,11 +261,20 @@ export function parseRunTargets(targets) {
...
-  if (excludeArchsOption) {
-    const excludeArchs = excludeArchsOption.trim().split(/\s*,\s*/)
-      .filter(arch => excludableWebArchs.includes(arch));
+function isModernArchsOnlyEnabled(appDir) {
+  const packageJsonPath = files.pathJoin(appDir, 'package.json');
+  if (!files.exists(packageJsonPath)) {
+    return false;
+  }
+  const packageJsonFile = files.readFile(packageJsonPath, 'utf8');
+  const packageJson = JSON.parse(packageJsonFile);
```
italojs, line 270: you can `require` .json files instead of reading and parsing the file.

italojs (changes requested):

italojs:
I would recommend adding a flexible option to the pkg.json config, like

1. using a generic property called `webArchs`
2. accept a string 'modern' and an array

```
webArchs: 'modern'
//or
webArchs: ['web.browser',  'web.cordova'] // web.browser.legacy
```

Advantages:
1. it covers the initial proposal to avoid legacy and Cordova bundles using an easy/simple config
2. It gives more flexibility to the developer to choose what bundles they want to build if needed

italojs:
One more thing:  could you please appoint this PR to the branch release-3.2?

9Morello:
@StorytellerCZ that's a fair assessment. Between making the flag apply only to development builds and having 2 separate flags for development and production, which one do you think would be easier to understand for newcomers?

If we went with the former, I wonder what would be a good way to inform the user the setting doesn't apply to production builds.

@italojs thanks for reviewing. I made the adjustment you requested regarding using `require` instead of `JSON.parse`. Regarding your suggestion to support a configurable array, IMO this would make it more confusing to a new user who scaffolded a Meteor project and went to inspect the contents of the `package.json` file. I see the value in making it more configurable to handle more use cases, but I believe those use cases are served well by existing flags or custom build scripts. I can adjust the implementation regardless, if other people prefer it that way. Just wanted to share my thoughts first. :+1: 

Edit: I changed the implementation back to `JSON.parse` since calling `require` was [causing a failure on Windows](https://github.com/meteor/meteor/actions/runs/14085405014/job/39448050903?pr=13665#step:6:28) tests. I assume it's related to `require`'ing files using their full path instead of a relative one. I saw other parts of the codebase using `JSON.parse` explicitly for the `package.json` file, which is the main reason I implemented it in the first place. If there's a better way to handle that, please let me know.

zodern:
For production, we have the `.meteor/platforms` file to configure the arch's. I always expected we would split the `browser` platform into modern and legacy platforms, and allow removing/adding them.

Meteor tries to be config-free. It would be possible to simply not rebuild the legacy client until the app receives a request from a legacy browser (I'm not sure we can or would want to avoid building the legacy client during the initial build).  We already have most of the infrastructure implemented for this from implementing the lazy bundling for the legacy client.

italojs:
@9Morello agrees with your points, LGMT

@zodern,  you have a critical point, "config-free." Avoiding it in cold-start build is nice to have IMO, so we can build all then and focus on the rebuild that will affect the DX directly. Still, at the same time, I would maintain the current configuration and explicitly ignore a possible lazy build if the developer definitively doesn't want it.

I see this lazy build approach for legacy would be excellent for bundler optimization. You mentioned we already have the infrastructure for it. Would you mind sharing a few files where we have it, please?

9Morello:
@italojs thanks for the feedback :) please let me know if you need any other changes.

italojs (approved):

nachocodoner:
@9Morello: Do you know if running the Cordova app with `meteor run [ios|android]` automatically includes the web.cordova architecture, ignoring modernWebArchsOnly?

In this case, it seems it always gets included, regardless of what's set in modernWebArchsOnly or excluded architectures.

Edit: I tested and Cordova is broken on use this new option.

image

nachocodoner:
@9Morello: I was testing the build command with the option enabled here. In Meteor 3.2, it still builds the `web.browser.legacy` program for production, but with the changes in this PR, it skips the legacy arch entirely.

image

Following the discussion started by @StorytellerCZ, it feels off that this option also affects the production build. Legacy arch in production acts as a fallback for wider support. The new option was mainly meant to simplify `--exclude-arch` for development use as this option is only enabled there. Maybe we should ensure production isn’t affected, and introduce new platform entries in the `.meteor/platforms` file to trigger `web.browser` and `web.browser.legacy` separately, and still having `browser` platform from hitting both and avoid breaking changes.

nachocodoner:
Working on the fixes https://github.com/meteor/meteor/pull/13693

We want to include this change for next beta coming just next.

9Morello:
@nachocodoner that makes sense. We should make it apply to development builds only.

harryadel:
Should this be added to the docs?

PR #13673, opened 2025-03-25
Bundle: Adding promise.all to the Target builds

italojs:
Before/After
Screenshot 2025-03-25 at 14 01 41
Screenshot 2025-03-25 at 14 04 38

### DIFF
--- a/tools/isobuild/bundler.js
+++ b/tools/isobuild/bundler.js
@@ -166,6 +166,8 @@ var colonConverter = require('../utils/colon-converter.js');
 var Profile = require('../tool-env/profile').Profile;
 var packageVersionParser = require('../packaging/package-version-parser.js');
 var release = require('../packaging/release.js');
+
+const { Worker } = require('worker_threads');
 import { loadIsopackage } from '../tool-env/isopackets.js';
 import { CORDOVA_PLATFORM_VERSIONS } from '../cordova';
 import { gzipSync } from "zlib";
@@ -1415,8 +1417,7 @@ class Target {
     const dynamicFiles = [];
     const { arch } = this;
     const inputHashesByJsFile = new Map;
-
-    this.js.forEach(file => {
+    await Promise.all(this.js.map(async file => {
       const jsf = new JsFile(file, { arch });
 
       inputHashesByJsFile.set(jsf, file.hash());
@@ -1426,7 +1427,7 @@ class Target {
       } else {
         staticFiles.push(jsf);
       }
-    });
+    }));
 
     var markedMinifier = buildmessage.markBoundary(
       minifierDef.userPlugin.processFilesForBundle,
@@ -3411,6 +3412,8 @@ async function bundle({
     }
 
     // Client
+    const clientArchsToProcess = [];
+    
     for (const arch of webArchs) {
       if (allowDelayedClientBuilds &&
           hasOwn.call(previousBuilders, arch) &&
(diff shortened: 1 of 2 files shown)


nachocodoner:
Great!

I'd like to test this branch with the other optimization branches. But I think it's time we require green PRs before merging changes into other PRs, to better isolate any issues that may come from specific changes.

tools/isobuild/bundler.js
```diff
@@ -3220,7 +3220,51 @@ Find out more about Meteor at meteor.com.
...
+        webArchs: [],
+        serverArch: options.buildOptions?.serverArch || archinfo.host(),
+      },
+      isServerOnlyBundle: true,
+      buildCache,
+    };
+    
+    // TODO: we must to paralelize it in two different threads
+    const clientResult = await bundle(clientOptions);
+    const serverResult = await bundle(serverOptions);
```
nachocodoner, line 3252: No longer the case, right? as they seem to be co-dependent.

tools/isobuild/bundler.js
```diff
@@ -3409,78 +3460,155 @@ async function bundle({
...
-                new Date - start
-            }ms`, { arrow: true });
-          }
-        });
-      } else {
-        // Otherwise make the client target now, and write it below.
-        targets[arch] = await makeClientTarget(app, arch, {minifiers});
+    const clientArchsToProcess = [];
+    
+    if (!isServerOnlyBundle) {
```
nachocodoner, line 3465: For better understanding on read this code, isn't it better here to use `isClientOnlyBundle`?

tools/isobuild/bundler.js
```diff
@@ -3409,78 +3460,155 @@ async function bundle({
...
+      });
+    } else if (clientArchsToProcess.length === 1) {
+      // Process for a single client arch
+      const arch = clientArchsToProcess[0];
+      targets[arch] = await makeClientTarget(app, arch, {minifiers});
+    }
 
-    // Server
-    if (! hasCachedBundle) {
+    if (!isClientOnlyBundle && !hasCachedBundle) {
```
nachocodoner, line 3534: same here, isn't it better `isServerOnlyBundle` used here?

tools/isobuild/bundler.js
```diff
@@ -3246,6 +3293,10 @@ async function bundle({
   } else {
     webArchs = projectContext.platformList.getWebArchs();
   }
+  
+  if (isClientOnlyBundle) serverArch = null;
+  if (isServerOnlyBundle) webArchs = [];
```
nachocodoner, line 3298: For the server bundle, were we previously using webArchs with all values? Not sure if it's still needed for the server. We should watch out for any breaking changes.

Later we pass webArchs when creating the server target: `targets.server = await makeServerTarget(app, webArchs);`. So will webArchs always be empty now? Not sure, maybe it's still required somewhere in the server target logic.

tools/isobuild/bundler.js
```diff
@@ -3409,78 +3460,155 @@ async function bundle({
...
+        
+        // Check if we have multiple targets to process
+        if (targetsArray.length > 1) {
+          // Process targets in parallel
+          await Promise.all(targetsArray.map(target => writeClientTarget(target)));
+        } else {
+          // Single target, process in main thread
+          for (const target of targetsArray) {
+            await writeClientTarget(target);
+          }
```
nachocodoner, line 3552: This will be the default mode after the changes in https://github.com/meteor/meteor/pull/13665. So it won’t benefit much from the optimization? In that scenario, is there a gain from the other changes?

I assume it will still help when using multiple architectures, which is good.

tools/isobuild/bundler.js
```diff
@@ -3220,7 +3220,51 @@ Find out more about Meteor at meteor.com.
...
+      buildCache,
+    };
+    
+    // TODO: we must to paralelize it in two different threads
+    const clientResult = await bundle(clientOptions);
+    const serverResult = await bundle(serverOptions);
+    
+    const combinedResult = {
+      errors: clientResult.errors || serverResult.errors,
+      warnings: clientResult.warnings || serverResult.warnings,
```
Grubba27, line 3256: Should not be like `postStartupCallbacks`? so that if both client and server has errors or warnings both get reported?

PR #13674, opened 2025-03-25
Support caching as part of compilation (Build and Rebuild App phases 🚀)

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13657

This PR adds a caching strategy for compilation. SWC don’t provide incremental build caching support by default (Babel seems only through meteor-babel), this is usually handled by the bundler. Rspack, and others already use caching to speed up rebuilds. This PR brings a similar approach to Meteor.

Right now, when Meteor rebuilds due to a single file change, it recompiles everything again. SWC makes this fast, but since we recompile all packages and app code sequentially, it adds unnecessary time. With this PR, unchanged files use the cache; changed files get recompiled and the result is cached.

There are two caches: memory and file. I know memory can grow without limit, so I plan to add an LRU cache to keep it bounded, falling back to file cache when needed. Though `meteor-babel` didn't seem [to have a limitation in its cache implementation](https://github.com/meteor/meteor/blob/devel/npm-packages/meteor-babel/cache.js).

See the performance gains compared to the swc-replacement branch this is based on. It's impressive! 🚀

image

> These values reflect performance gains on a small app built with `meteor create --react`. The improvements will become more noticeable in larger, feature-rich applications.

----------------

Side note: These improvements will help when moving to a modern bundler. The modern bundler will handle app code and provide better caching from the external tool. Still, these changes make sure Meteor won't be a bottleneck during development, rebuilding faster and compiling Meteor packages as quickly as possible.


9Morello:
@nachocodoner minor questions: 1. do you know the hardware specs where we are measuring those improvements, and 2. is there an easy way to reproduce those benchmarks. so that we could provide you with more data from different machines? :) 

Great work!

italojs:
@9Morello here you can see how to execute the profile https://blog.meteor.com/unlocking-meteor-3-2-new-profiling-tool-to-track-bundler-performance-and-size-f1f223ffa49d
Execute it between 2 different versions to get the benchmark

packages/babel-compiler/babel-compiler.js
```diff
@@ -51,7 +53,9 @@ BCp.processFilesForTarget = function (inputFiles) {
 // Returns an object suitable for passing to inputFile.addJavaScript, or
 // null to indicate there was an error, and nothing should be added.
 BCp.processOneFileForTarget = function (inputFile, source) {
+  var self = this; // capture context
```
italojs: Why do we need to introduce a new variable called "self" instead of using "this"? I understand it's quite common in Python, but it's not something you typically see in the JavaScript ecosystem.

italojs (changes requested):

packages/babel-compiler/babel-compiler.js
Grubba27: I think in this case `this` should be fine but normally it was done before due to this behavior:

https://stackoverflow.com/a/38549303

packages/babel-compiler/babel-compiler.js
```diff
@@ -122,43 +126,57 @@ BCp.processOneFileForTarget = function (inputFile, source) {
...
-        try {
-          const packagesSkipSwc = [];
-          const fileSkipSwc = ['webapp_server.js']; // top level await
-
-          // Determine if SWC should be used based on package and file criteria.
-          const shouldUseSwc =
-            !packagesSkipSwc.includes(packageName) &&
-              !fileSkipSwc.includes(inputFilePath);
+        // Determine if SWC should be used based on package and file criteria.
+        const packagesSkipSwc = [];
```
Grubba27, line 210: this variable will be updated in the future to skip swc usage right?

packages/babel-compiler/babel-compiler.js
nachocodoner: I had to save `this` in a variable `self` to preserve the context later. Inside `profile('Babel.compile', function () {`, this refers to a different context, while we still want to access the one above (`BCp.processOneFileForTarget`).

packages/babel-compiler/babel-compiler.js
nachocodoner: Yes, this will likely become a config in package.json to allow more flexibility, better migration compatibility, and incremental adoption by skipping specific packages or files that are harder to adapt.

nachocodoner:
@9Morello

> 1. do you know the hardware specs where we are measuring those improvements

Intel Core Raptor Lake i9 13900K
64GB DDR5 6000 MHz CL30 RAM (Node-limited)
WD Black SN850X SSD

The app used: `meteor create react-app --react --release 3.2`

The specs are from my local machine. The reasons:

- Since I’m actively working on the PR, a fast machine lets me spend more time on the actual improvements, instead of waiting or experience side effects from remote or lower-end machines.

- The comparison is still valid across any machines, the goal is to measure the relative improvement between the 3.2 state and the new changes.

Later, once development is done, we plan to use a remote machine with more common specs. This will help share results in a standard environment, detect regressions (CI), and provide baseline values. In the beta, we’ll also welcome any profiling done in your own apps and machines. Measuring the gains for particular apps would be great to see.

> 2. is there an easy way to reproduce those benchmarks. so that we could provide you with more data from different machines? :)

Yes. To compare branches, check out the one you want and run:

``` shell
METEOR_CHECKOUT_PATH=~/meteor/meteor-3 meteor profile --exclude-archs web.browser.legacy,web.cordova
```

Make sure the path points to the right Meteor checkout (the one with your changes or branch).

Sometimes it helps to run `~/meteor/meteor-3/meteor run` first to ensure the dev_bundle is ready before profiling.

packages/babel-compiler/babel-compiler.js
```diff
@@ -122,46 +126,63 @@ BCp.processOneFileForTarget = function (inputFile, source) {
...
+            const cacheContext = '.swc-cache';
+
+            // Check RAM cache
+            compilation = self._swcCache[cacheKey];
+            // Check file system cache if enabled
+            if (!compilation && self.cacheDirectory) {
+              const cacheFilePath = path.join(self.cacheDirectory, cacheContext, cacheKey + '.json');
+              if (fs.existsSync(cacheFilePath)) {
+                try {
+                  compilation = JSON.parse(fs.readFileSync(cacheFilePath, 'utf8'));
```
italojs: What if we use a `require` instead of reading the file and then using JSON.parse? I don't know how it would affect performance, but it would be nice to test this option.

packages/babel-compiler/babel-compiler.js
italojs: an option is to use bind or arrow functions like

``` js
function f1() {
    try { 
        var result = profile('Babel.compile', function cb() {
            console.log(this) // will use f1's context
        }.bind(this))
    }
    // ...
}

// or

function f1() {
    try { 
        var result = profile('Babel.compile', () => { 
            console.log(this) // lexically binds to f1's context
        })
    }
    // ...
}

**Review by italojs — CHANGES_REQUESTED** [review]

File: packages/babel-compiler/babel-compiler.js (hunk 3 above)
Comment (italojs) [reply packages/babel-compiler/babel-compiler.js]: Did a benchmark: https://gist.github.com/italojs/00cd47fbd09271031c6ec87bc20cb78b
Require is pretty faster

```
/* Results from my local Intel Mac i9 16gb
fs.readFileSync + JSON.parse:
  Min: 0.0174 ms
  Max: 0.7195 ms
  Average: 0.0221 ms
require:
  Min: 0.0006 ms
  Max: 0.4166 ms
  Average: 0.0010 ms
fs.readFile:
  Min: 0.0701 ms
  Max: 5.3427 ms
  Average: 0.1006 ms
Comparison:
  1. require (Avg: 0.0010 ms)
  2. fs.readFileSync + JSON.parse (Avg: 0.0221 ms)
  3. fs.readFile + JSON.parse (Avg: 0.1006 ms)
*/
```
Maybe we need to take care of relative paths that seem not to work for Windows as described. [here](https://github.com/meteor/meteor/pull/13665#issuecomment-2754637053)

italojs: it's another opportunity to gain some ms in the bundler since a higher I/O is a huge problem in our bundler today

nachocodoner: Oh, I have seen the pattern of requires in other caches on Meteor code, so good point @italojs. I appreciate that benchmark you did. I will apply that 🚀

zodern: `require` caches the result and wouldn't re-read the disk which makes it much faster in the benchmark. The require caching wouldn't have any benefit here since each file is already cached in-memory. Though it doesn't hurt to use either. It seems meteor-babel also uses require: https://github.com/meteor/meteor/blob/ae8cf586acc9a4c7bf9a5ab79dc5f8b7ef433a64/npm-packages/meteor-babel/cache.js#L81

zodern:
> Make the cache hashes to consider the SWC version, reify and other requirements

The caches are already cleared by Meteor every time the version of the build plugin changes, so you should only need to consider things that can change within the same version of the build plugin. Though I'm not sure if it works correctly for local build plugins.

> Though meteor-babel didn't seem [to have a limitation in its cache implementation](https://github.com/meteor/meteor/blob/devel/npm-packages/meteor-babel/cache.js).

If it's worked in the past, it might be good to keep it the same for now. Wherever we do use LRU caches introduces a potential performance cliff for large apps that don't fit within the cache size.

One optimization we use in a few places, such as in the meteor-babel cache and for reify is to read a list of all cache files into memory to avoid an extra disk access per file: https://github.com/meteor/meteor/blob/ae8cf586acc9a4c7bf9a5ab79dc5f8b7ef433a64/tools/static-assets/server/runtime.js#L10-L22

packages/babel-compiler/babel-compiler.js
nachocodoner: I tested `JSON.parse(fs.readFileSync` versus `require` and found the profiles run slower. The cache file is read every time the Meteor app restarts, during each phase shown in the image below. I also highlighted some packages that show additional milliseconds added with that approach, which appears to happen with many of the package read sequentially.

image

The benchmark above @italojs was using the Node version Meteor uses? Maybe there was some kind of optimziation on latest one we use.

nachocodoner: I’m unable to use the standard Node require inside a Meteor package like babel-compiler (build plugin). Meteor packages use `Npm.require`, which applies its own module resolution logic. The pure Node require doesn’t seem to work in that context.

nachocodoner: I enabled `require` by using `api.use('modules');` in the Meteor package `babel-compiler`. But when using `require` I see the profiles going slower.

It can't import using require, all time with errors like `Error: Cannot find module '/home/freak2geek/meteor/profile-apps/example32/.meteor/local/plugin-cache/ecmascript/local/.swc-cache/b79ff3679ff5c045137389163a9c1b76643b182d.json'`, and the file exists. Also, trying relative paths with some troubles on the way. I think Meteor packages might be limiting using require as we would do in a npm package context.

Maybe to get these gains we should move the logic into a npm package, but I wish to avoid maintaining another NPM package in this moment that would slow down our implementations.

@zodern: any thoughts on this? Would you recommend us to move to a NPM package to get more speed with require?

zodern: I would suggest two options:
- importing and using the cache system from meteor-babel
- using fs.readfile and json.parse. When used in the cache, I wouldn't expect `require` to be significantly faster.

packages/babel-compiler/babel-compiler.js
nachocodoner: Since I refactored with `Babel.compile` and `SWC.compile` profile wrappers, there’s no need for self anymore. I switched to an arrow function, which takes care of the this binding. Unless there are additional comments, I’ll consider this resolved.

packages/babel-compiler/babel-compiler.js
nachocodoner: I think sticking with `fs.readFileSync + JSON.parse`. I tried enabling the cache through the `meteor-babel` implementation, but performance actually got worse (see attached). The current approach, which I refactored, runs faster. I suspect that the re-hashing of files for each operation adds significant overhead. In any case, I don’t want to spend more time on this right now, so I’ll leave it open for others to explore further improvements.

image

nachocodoner: I think sticking with `fs.readFileSync + JSON.parse` makes sense. I tried enabling the cache through the `meteor-babel` implementation, but performance I feel it slightly worse in the several runs I performed, the current approach runs faster. I suspect that the re-hashing of files for each operation adds overhead. In any case, I don’t want to spend more time on this right now.

PR #13675, opened 2025-03-27
Perserve reify in SWC compilation

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13657 and https://github.com/meteor/meteor/pull/13674

This PR [applies a suggestion](https://github.com/meteor/meteor/pull/13657#discussion_r2003531972) from @zodern to transpile using SWC, but ensure we get that into a format that later, reify compiler can apply its features. Reify is what makes Meteor unique by adding support of reify modules, nested imports and top-level await. It seems to be also a critical tool inter-dependent with other features. By just this fix on this PR, HMR started to work again, and I was able to see nested imports working as fallback. So, with this change the migration to use a faster transpiler will deliver no breaking changes apparently. But what is more, we perserve the speed we meant to get.

It’s true that using reify might be a bit slower than using SWC alone, but the numbers below show the difference isn’t significant, and more importantly, rebuild times are almost the same. My point is to preserve Meteor specifics whenever possible, without major speed loss. I believe that’s still the case here. There may be further improvements to make (I'm still exploring) but even as is, this should already be better than the latest Meteor 3.2 release.

image

> These values don't reflect the final state of this PR, as further improvements were made afterward.

> The values reflect performance gains on a small app built with `meteor create --react`. The improvements will become more noticeable in larger, feature-rich applications.

> NOTE: Nested imports are supported, but we removed them from the core, preferring standard require or importing at the root if already done elsewhere. This is because SWC can't parse nested imports, leading to an error and a fallback to Babel, which is slower. Keeping the core simple and performant is our goal. Be aware that using nested imports in your packages or app code will slow down your build.

---

Side note: I really appreciate @zodern’s input 🙏. One of Meteor’s challenges is the limited shared knowledge. I’m still learning how all the parts fit together, and his guidance has been very helpful in pointing out what matters. With that, we can better ensure the things that make Meteor unique are preserved. With modern bundlers on the app's code, it’s uncertain how long that will stay true, but at least now we have the understanding to try.

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -1,6 +1,8 @@
 var semver = Npm.require("semver");
 var JSON5 = Npm.require("json5");
 var SWC = Npm.require("@swc/core");
+const reifyCompile = Npm.require("@meteorjs/reify/lib/compiler").compile;
+const reifyAcornTopLevelParse = Npm.require("@meteorjs/reify/lib/parsers/top-level").parse;
 
 /**
  * A compiler that can be instantiated with features and used inside
@@ -149,7 +151,7 @@ BCp.processOneFileForTarget = function (inputFile, source) {
         let compilation;
         try {
           const packagesSkipSwc = [];
-          const fileSkipSwc = ['webapp_server.js']; // top level await
+          const fileSkipSwc = []; // top level await
 
           // Determine if SWC should be used based on package and file criteria.
           const shouldUseSwc =
@@ -171,13 +173,25 @@ BCp.processOneFileForTarget = function (inputFile, source) {
                   tsx: hasTSXSupport,
                 },
               },
-              module: { type: 'commonjs' },
+              module: { type: 'es6' },
               minify: false,
               sourceMaps: true,
             });
 
+            let content = transformed.code;
+            // Perserve Meteor's specifics: reify modules, nested imports and top-level await support.
+            const result = reifyCompile(content, {
+              parse: reifyAcornTopLevelParse,
+              generateLetDeclarations: false,
+              ast: false,
+            });
+            if (!result.identical) {
+              identical = false;
+              content = result.code;
+            }
+
             compilation = {
-              code: transformed.code,
+              code: content,
               map: JSON.parse(transformed.map),
               hash: toBeAdded.hash,
               sourceType: 'module',
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -58,6 +58,7 @@ BCp.processOneFileForTarget = function (inputFile, source) {
   var self = this; // capture context
   this._babelrcCache = this._babelrcCache || Object.create(null);
   this._swcCache = this._swcCache || Object.create(null);
+  this._swcIncompatible = this._swcIncompatible || Object.create(null);
 
   if (typeof source !== "string") {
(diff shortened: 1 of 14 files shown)


nachocodoner:
image

During the build process, the time spent for the app code to go through SWC + reify is slightly longer than using SWC alone, but the difference isn’t a concern I believe. We plan to move this step to a more modern bundler later. In any case, the build app phase’s duration becomes less critical when caching is enabled, which will typically be the case.

image

On rebuild app, I see new phases going on when SWC+reify. In SWC purely, the webapp package was skipped to use common babel usage (to support top level await), and on rebuild was not even appearing in the profile as a slow step. Not sure why now using reify it doesn't run it fast from cache or take as much that step. Maybe there is something off I am next taking a look.

nachocodoner:
> NOTE: Nested imports are supported, but we removed them from the core, preferring standard require or importing at the root if already done elsewhere. This is because SWC can't parse nested imports, leading to an error and a fallback to Babel, which is slower. Keeping the core simple and performant is our goal. Be aware that using nested imports in your packages or app code will slow down your build.

I removed nested imports because they made this approach slower than the other branch where they’re removed. This aligns with our goal: to support nested imports but avoid unnecessary slowdowns, particularly in the core.

**There’s also an issue in webapp_server.js where top-level await falls back to Babel due to an acorn syntax error. I’ll look into this as it contributes to the slowdown.**

packages/babel-compiler/babel-compiler.js
```diff
@@ -223,6 +240,7 @@ BCp.processOneFileForTarget = function (inputFile, source) {
             compilation = Babel.compile(source, babelOptions, cacheOptions);
           }
         } catch (e) {
+          self._swcIncompatible[toBeAdded.hash] = true;
```
nachocodoner, line 255: This change speeds up subsequent compilation of incompatible files (nested-imports mainly).

packages/babel-compiler/babel-compiler.js
```diff
@@ -192,13 +197,37 @@ BCp.processOneFileForTarget = function (inputFile, source) {
...
+              avoidModernSyntax: true,
+              enforceStrictMode: false,
+              dynamicImport: true,
+              ...features.topLevelAwait && { topLevelAwait: true },
+              ...features.compileForShell && { moduleAlias: 'module' },
+              ...(features.modernBrowsers ||
+                features.nodeMajorVersion >= 8) && {
+                avoidModernSyntax: false,
+                generateLetDeclarations: true,
+              },
```
nachocodoner, line 222: @zodern: Adding these options definitely improved the build time. Previously, the top-level await was triggering a Babel fallback instead of using SWC + Reify. Now it’s resolved, and the build times are consistent. 😁

tools/static-assets/server/runtime.js
```diff
@@ -47,15 +47,15 @@ module.exports = function enable ({ cachePath, createLoader = true } = {}) {
...
+  const reifyAcornParse = require("@meteorjs/reify/lib/parsers/acorn").parse;
   const reifyCompile = require("@meteorjs/reify/lib/compiler").compile;
 
   function compileContent (content) {
     let identical = true;
 
     try {
       const result = reifyCompile(content, {
-        parse: reifyBabelParse,
+        parse: reifyAcornParse,
```
nachocodoner, line 58: Using `reifyAcornParse` instead of `reifyBabelParse` in places used has improved further. The profile in the description PR is not updated on this.

PR #13677, opened 2025-03-28
Add Accounts.replaceEmailAsync

harryadel:
I was recently working on application where we had wanted to replace the user email and I was surprised the Meteor accounts offers no such utility. Although Meteor does offer a way to [add emails](https://docs.meteor.com/api/accounts#Accounts-addEmailAsync) and [remove them](https://docs.meteor.com/api/accounts#Accounts-removeEmail) and they may be combined to allow you to replace an email, that wouldn't be the best solution as it entails two different email operations instead of one.  

Accounts.replaceEmailAsync aims to help fix this problem, it checks first if such email collides with another user email before adding, it also notifies you in case it hadn't been able to find a user with the old email you provided instead of failing silently.

### DIFF
--- a/packages/accounts-password/password_server.js
+++ b/packages/accounts-password/password_server.js
@@ -1022,6 +1022,52 @@ Meteor.methods(
     }
   });
 
+
+/**
+ * @summary Asynchronously replace an email address for a user. Use this instead of directly
+ * updating the database. The operation will fail if there is a different user
+ * with an email only differing in case. If the specified user has an existing
+ * email only differing in case however, we replace it.
+ * @locus Server
+ * @param {String} userId The ID of the user to update.
+ * @param {String} oldEmail The email address to replace.
+ * @param {String} newEmail The new email address to use.
+ * @param {Boolean} [verified] Optional - whether the new email address should
+ * be marked as verified. Defaults to false.
+ * @importFromPackage accounts-base
+ */
+Accounts.replaceEmailAsync = async (userId, oldEmail, newEmail, verified) => {
+  check(userId, NonEmptyString);
+  check(oldEmail, NonEmptyString);
+  check(newEmail, NonEmptyString);
+  check(verified, Match.Optional(Boolean));
+
+  if (verified === void 0) {
+    verified = false;
+  }
+
+  const user = await getUserById(userId, { fields: { _id: 1 } });
+  if (!user)
+    throw new Meteor.Error(403, "User not found");
+
+  // Ensure no user already has this new email
+  await Accounts._checkForCaseInsensitiveDuplicates(
+    "emails.address",
+    "Email",
+    newEmail,
+    user._id
+  );
+
+  const result = await Meteor.users.updateAsync(
+    { _id: user._id, 'emails.address': oldEmail },
+    { $set: { 'emails.$.address': newEmail, 'emails.$.verified': verified } }
+  );
+  
+  if (result.modifiedCount === 0) {
+    throw new Meteor.Error(404, "No user could be found with old email");
+  }
+};
+
 /**
  * @summary Asynchronously add an email address for a user. Use this instead of directly
  * updating the database. The operation will fail if there is a different user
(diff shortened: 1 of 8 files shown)


packages/accounts-password/password_server.js
```diff
@@ -1022,6 +1022,52 @@ Meteor.methods(
...
+/**
+ * @summary Asynchronously replace an email address for a user. Use this instead of directly
+ * updating the database. The operation will fail if there is a different user
+ * with an email only differing in case. If the specified user has an existing
+ * email only differing in case however, we replace it.
+ * @locus Server
+ * @param {String} userId The ID of the user to update.
+ * @param {String} oldEmail The email address to replace.
+ * @param {String} newEmail The new email address to use.
+ * @param {Boolean} [verified] Optional - whether the new email address should
```
StorytellerCZ, line 1035: I would like to think a bit about the `true` use case. I'm not entirely sure about it. I think we should always send out a verification email (though if set from admin then it might not be necessary).
But what is missing, even if set to `false` I don't see here a verification email being sent out which I think is a must. There also probably should be sent out an email to the old email that it has been changed (maybe an optional flag). This might necessitate new email templates.

StorytellerCZ (changes requested):

v3-docs/docs/api/accounts.md
```diff
@@ -888,6 +888,8 @@ email with a link the user can use to verify their email address.
 
 <ApiBox name="Accounts.removeEmail" />
 
+<ApiBox name="Accounts.replaceEmail" />
```
nachocodoner: You implemented the async version directly, shouldn't it be named Accounts.replaceEmailAsync instead? The same applies elsewhere.

packages/accounts-password/password_tests.js
```diff
@@ -1789,6 +1789,26 @@ if (Meteor.isServer) (() => {
       ]);
     });
 
+    Tinytest.addAsync("passwords, replace email", async test => {
+      const origEmail = `originalemail@test.com`;
+      const userId = await Accounts.createUser({
```
nachocodoner, line 1794: Should it be `createUserAsync` here? Also, does this test actually work? It seems createUser only provides the ID via a callback, not directly.

nachocodoner: I understand that on the server, [`createUser` is essentially the same as `createUserAsync`](https://github.com/meteor/meteor/blob/ae8cf586acc9a4c7bf9a5ab79dc5f8b7ef433a64/packages/accounts-password/password_server.js#L1291), but this naming convention diverges from the usual sync/async pattern. Would it be better to use `createUserAsync` here in case we update the naming later?

packages/accounts-password/password_server.js
nachocodoner: I believe we’ve kept user management and email verification separate, and that approach still works, right?

For example, [`Accounts.createUserAsync`](https://docs.meteor.com/api/accounts#Accounts-createUser) doesn’t automatically send a verification email; you can explicitly call [`Accounts.sendVerificationEmail`](https://docs.meteor.com/api/accounts#Accounts-sendVerificationEmail) afterward. Similarly, `Accounts.replaceEmailAsync` can be followed by `Accounts.sendVerificationEmail` to handle the verification process in the app controller level.

There’s also a  [`Accounts.createUserVerifyingEmail`](https://docs.meteor.com/api/accounts#Accounts-createUserVerifyingEmail) version of `createUser`, sending a verification or an enrollment email depending on email or password being provided. But I wouldn't think a similar method is needed for replacing email.

What do you think?

harryadel: I side with Nacho. I don't think adding a verification step is needed. It's best to keep operations atomic and single-purpose. Especially that it goes hand in hand with code quality (doing less as in not also handling email verification in the same method) and also following the same flow with already implemented methods.

harryadel:
@nachocodoner please recheck.

nachocodoner (approved):

PR #13678, opened 2025-03-29
Update accounts-base.d.ts to reflect async await in EmailFields templ…

ToastHawaii:
Update accounts-base.d.ts to reflect async await in EmailFields templates.


italojs:
From what I found, those functions doenst retun a promise

https://github.com/meteor/meteor/blob/cad7de6cd39db0a47361637b280275dd367e7cb2/packages/accounts-password/email_templates.js#L29

https://github.com/meteor/meteor/blob/cad7de6cd39db0a47361637b280275dd367e7cb2/packages/accounts-password/email_templates.js#L34

https://github.com/meteor/meteor/blob/cad7de6cd39db0a47361637b280275dd367e7cb2/packages/accounts-password/email_templates.js#L39

maybe the correct would be remove the await from tthe calls
https://github.com/meteor/meteor/blob/7b775016bf457d9130247adcb5676bc43c59773d/packages/accounts-base/accounts_server.js#L1440

let me know if you found a place or usecase where it's returning a promise

ToastHawaii:
@italojs We overwrite the HTML template and other functions as described in this document https://docs.meteor.com/api/accounts.html#Accounts-emailTemplates and load some logos and other resources asynchronously with `Assets.getBinaryAsync`. 
After switching to Meteor v3, I don't know of any way to do this synchronously. So removing await at this point would be very bad for us as we would no longer be able to customise the email to our customers needs. 

So we need `await` at this point.

packages/accounts-base/accounts-base.d.ts
```diff
@@ -10,10 +10,10 @@ export interface URLS {
 }
 
 export interface EmailFields {
-  from?: ((user: Meteor.User) => string) | undefined;
-  subject?: ((user: Meteor.User) => string) | undefined;
-  text?: ((user: Meteor.User, url: string) => string) | undefined;
-  html?: ((user: Meteor.User, url: string) => string) | undefined;
+  from?: ((user: Meteor.User) => Promise<string>) | undefined;
```
italojs: could you return a Promise<string> | string in that function definition, please?

italojs (changes requested):

italojs (dismissed):
we will merge after the next release

PR #13686, opened 2025-04-02
fix(webapp): prevent truncated agent string in lookupUserAgent

welkinwong:
In useragent-ng's lookup function implementation, there are security validations and truncation checks for useragent length.  For certain useragent use cases where the length exceeds 150 characters, premature truncation before being passed to the lookup function results in incorrect agent identification.

**in useragent-ng parse:**

https://github.com/schmod/useragent-ng/blob/9d22aab904b8134f7680102415539b0570fc7405/index.js#L414

```js
exports.parse = function parse(userAgent, jsAgent) {
  if (userAgent && userAgent.length > 1000) {
    userAgent = userAgent.substring(0, 1000);
  }

  if (!userAgent || !isSafe(userAgent)) return new Agent();

  var length = agentparserslength
    , parsers = agentparsers
    , i = 0
    , parser
    , res;
```

**in lookup**

https://github.com/schmod/useragent-ng/blob/9d22aab904b8134f7680102415539b0570fc7405/index.js#L482

```js
exports.lookup = function lookup(userAgent, jsAgent) {
  if (!LRU) {
    LRU = new (require('lru-cache'))({ max: 5000 });
  }
  var key = (userAgent || '')+(jsAgent || '')
    , cached = LRU.get(key);

  if (cached) return cached;
  LRU.set(key, (cached = exports.parse(userAgent, jsAgent)));

  return cached;
};
```

### DIFF
--- a/packages/webapp/webapp_server.js
+++ b/packages/webapp/webapp_server.js
@@ -132,7 +132,7 @@ var identifyBrowser = function(userAgentString) {
       patch: 0
     };
   }
-  var userAgent = lookupUserAgent(userAgentString.substring(0, 150));
+  var userAgent = lookupUserAgent(userAgentString);
   return {
     name: camelCase(userAgent.family),
     major: +userAgent.major,
--- a/packages/webapp/webapp_tests.js
+++ b/packages/webapp/webapp_tests.js
@@ -153,6 +153,25 @@ Tinytest.addAsync("webapp - modern/legacy static files", test => {
   return Promise.all(promises);
 });
 
+const specialUserAgent =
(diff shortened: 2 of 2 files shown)


StorytellerCZ:
Do we  have the version with the security fix in Meteor?

italojs:
Could you please add a test that covers the "incorrect agent identification" scenario?

welkinwong:
> Could you please add a test that covers the "incorrect agent identification" scenario?

of course.  If needed, I can also provide more error detection use cases.

welkinwong:
**before:**

Snipaste_2025-04-03_13-09-53


**after:**

Snipaste_2025-04-03_13-00-31

PR #13687, opened 2025-04-02
Fix timing issue with oplog include/exclude collections

ericm546:
I encountered an intermittent issue in my Meteor application when using oplogExcludeCollections to exclude collections I don't subscribe to.

After investigation, I found that the problem lies with _waitUntilCaughtUp. The function doesn't apply the same collection exclusion filters used by _startTailing. As a result, lastEntry may reference an oplog document that is never included in the _startTailing results. This discrepancy causes _waitUntilCaughtUp to stall until a new document is inserted into the oplog.

### DIFF
--- a/packages/mongo/oplog_tailing.ts
+++ b/packages/mongo/oplog_tailing.ts
@@ -37,13 +37,15 @@ export class OplogHandle {
   public _dbName: string;
   private _oplogLastEntryConnection: MongoConnection | null;
   private _oplogTailConnection: MongoConnection | null;
-  private _oplogOptions: { excludeCollections?: string[]; includeCollections?: string[] } | null;
+  private _oplogOptions: {
+    excludeCollections?: string[];
+    includeCollections?: string[];
+  };
   private _stopped: boolean;
   private _tailHandle: any;
   private _readyPromiseResolver: (() => void) | null;
   private _readyPromise: Promise<void>;
   public _crossbar: any;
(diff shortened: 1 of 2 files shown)


italojs:
Amazing!
Could you add tests for this PR, please? 
In special a test where we can reproduce the current issue, but passing since your implementation fix it

italojs (changes requested):
Amazing!
Could you please add tests for this PR?
In particular, a test where we reproduce the issue(but passing because you solved the problem)

ericm546:
@italojs I've now added a test that only pass after my fix.

italojs (approved):

nachocodoner:
Hey @ericm546, we've included already this fix as part of Meteor 3.3 release. Please test it. Also, do you think that what is described here could be what you fixed in this forum issue description, https://forums.meteor.com/t/issue-with-oplogexcludecollections/63593 or https://github.com/meteor/meteor/pull/13009#issuecomment-2185682933? Maybe the _waitUntilCaughtUp is what causes that delay on response time reported.

lynchem:
> do you think that what is described here could be what you fixed in this forum issue description

It sounds like it could be. 

I could test it if it's easy to backport it to 2.16

ericm546:
@nachocodoner, I re-tested it with my app by upgrading to 3.3 and it's working. I believe this will also resolve the two issues you mentioned.

PR #13692, opened 2025-04-08
DEV: Enhance type definitions for Meteor methods

welkinwong:
Added type definitions primarily for `serverPromise` and `stubPromise`. Also optimized the types for `callAsync, call, and apply` with reference to `applyAsync`.

```ts
function callAsync... : Promise<Result> & { stubPromise: Promise<Result>, serverPromise: Promise<Result> };
```

### DIFF
--- a/packages/webapp/package.js
+++ b/packages/webapp/package.js
@@ -1,12 +1,12 @@
 Package.describe({
   summary: "Serves a Meteor app over HTTP",
-  version: "2.0.5",
+  version: "2.0.6",
 });
 
 Npm.depends({
   "cookie-parser": "1.4.6",
-  express: "5.0.1",
-  "@types/express": "5.0.0",
+  express: "5.1.0",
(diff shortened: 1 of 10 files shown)


italojs (approved):

packages/webapp/package.js
```diff
@@ -1,12 +1,12 @@
...
-  version: "2.0.5",
+  version: "2.0.6",
 });
 
 Npm.depends({
   "cookie-parser": "1.4.6",
-  express: "5.0.1",
-  "@types/express": "5.0.0",
+  express: "5.1.0",
+  "@types/express": "5.0.1",
```
nachocodoner, line 9: This PR also updates to Express 5.1.0, not just a Meteor docs and types improvements. It's fine to include it in 3.3, thanks for that. 🙏

In the future, please send a separate PR when upgrading tool versions in Meteor core packages.

PR #13693, opened 2025-04-10
Fix issues with `modernWebArchsOnly` new config

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13665

I found two issues with `modernWebArchsOnly` during my testing. We plan to ship this change to complement the speed improvements with SWC.

Two issues:

- [x] With `modernWebArchsOnly: true`, Cordova breaks completely, development isn’t possible. https://github.com/meteor/meteor/pull/13665#issuecomment-2790325359

image

- [x] With `modernWebArchsOnly: true`, the production bundle misses web.browser.legacy. https://github.com/meteor/meteor/pull/13665#issuecomment-2791885863

image

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -286,8 +286,8 @@ function filterWebArchs(webArchs, excludeArchsOption, appDir, options) {
     }
   } else {
     // Dev & Test Mode
-    const isCordovaEnabled = (options.args || []).some(arg => ['ios', 'ios-device', 'android', 'android-device'].includes(arg));
-    if (!isCordovaEnabled) {
+    const isCordovaDev = (options.args || []).some(arg => ['ios', 'ios-device', 'android', 'android-device'].includes(arg));
+    if (!isCordovaDev) {
       const automaticallyIgnoredLegacyArchs = (appDir && isModernArchsOnlyEnabled(appDir)) ? ['web.browser.legacy', 'web.cordova'] : [];
       if (excludeArchsOption || automaticallyIgnoredLegacyArchs.length) {
         const excludeArchs = [...(excludeArchsOption ? excludeArchsOption.trim().split(/\s*,\s*/) : []), ...automaticallyIgnoredLegacyArchs];
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -272,14 +272,29 @@ function isModernArchsOnlyEnabled(appDir) {
 }
 
 function filterWebArchs(webArchs, excludeArchsOption, appDir, options) {
-  const isCordovaEnabled = (options.args || []).some(arg => ['ios', 'ios-device', 'android', 'android-device'].includes(arg));
-  if (!isCordovaEnabled) {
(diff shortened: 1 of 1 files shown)


tools/cli/commands.js
```diff
@@ -271,11 +271,30 @@ function isModernArchsOnlyEnabled(appDir) {
...
+  if (isBuildMode) {
+    // Build Mode
+    const isModernOnlyPlatform = platforms.includes('modern') && !platforms.includes('legacy');
+    if (isModernOnlyPlatform) {
+      webArchs = webArchs.filter(arch => arch !== 'web.browser.legacy');
+    }
+    const hasCordovaPlatforms = platforms.includes('android') || platforms.includes('ios');
+    if (!hasCordovaPlatforms) {
+      webArchs = webArchs.filter(arch => arch !== 'web.cordova');
+    }
```
nachocodoner, line 286: Now, if you add `modern` within `.meteor/local` you establish modern only build. Otherwise, just having `browser` will include both, legacy and modern archs when `meteor build`.

tools/cli/commands.js
```diff
@@ -271,11 +271,30 @@ function isModernArchsOnlyEnabled(appDir) {
...
-  const automaticallyIgnoredLegacyArchs = (appDir && isModernArchsOnlyEnabled(appDir)) ? ['web.browser.legacy', 'web.cordova'] : [];
-  if (excludeArchsOption || automaticallyIgnoredLegacyArchs.length) {
-    const excludeArchs = [...(excludeArchsOption ? excludeArchsOption.trim().split(/\s*,\s*/) : []), ...automaticallyIgnoredLegacyArchs];
-    webArchs = webArchs.filter(arch => !excludeArchs.includes(arch));
+function filterWebArchs(webArchs, excludeArchsOption, appDir, options) {
+  const platforms = (options.platforms || []);
+  const isBuildMode = platforms?.length > 0;
+  if (isBuildMode) {
+    // Build Mode
+    const isModernOnlyPlatform = platforms.includes('modern') && !platforms.includes('legacy');
```
nachocodoner, line 279: Now, if you add `modern` at `.meteor/platforms` file, `meteor build` builds `web.browser` arch only. Otherwise, the both `web.browser` and `web.browser.legacy` archs are bundled.

italojs (approved):

dosomder:
@nachocodoner After your change, how is possible to exclude legacy arch for productive build?

From your comment in the PR it seems you intended to use `modern` in platforms file? However, then this will be interpreted as a cordova platform: https://github.com/meteor/meteor/blob/devel/tools/project-context.js#L1441

nachocodoner:
This PR ensures legacy builds remain in the production bundle, as they did in Meteor 3.2.

To exclude them, add "modern" to your .meteor/platform file.

Check the docs: https://docs.meteor.com/about/modern-build-stack/bundler.html#web-arch

> To exclude legacy builds in production, add "modern" to the .meteor/platforms file.

> server
> browser
> modern

PR #13694, opened 2025-04-10
Support option to profile `meteor build`

nachocodoner:
Continues: https://github.com/meteor/performance/pull/14

This PR adds the `meteor profile --build` option to the Meteor tool. It profiles the `meteor build` command across cold, cached, and final scenarios to allow performance comparisons and generate complete profiles.

image

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -3362,7 +3362,7 @@ const setupBenchmarkSuite = async (profilingPath) => {
   process.env.GIT_TERMINAL_PROMPT = 0;
 
   const repoUrl = "https://github.com/meteor/performance";
-  const branch = "v3.2.0";
+  const branch = "v3.3.0";
   const gitCommand = [
     `mkdir -p ${profilingPath}`,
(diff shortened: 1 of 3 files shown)


tools/cli/commands.js
```diff
@@ -3362,7 +3362,7 @@ const setupBenchmarkSuite = async (profilingPath) => {
   process.env.GIT_TERMINAL_PROMPT = 0;
 
   const repoUrl = "https://github.com/meteor/performance";
-  const branch = "v3.2.0";
+  const branch = "v3.3.0";
```
Grubba27, line 3365: is not there a way to always get the latest branch?

Grubba27 (approved):

tools/cli/commands.js
nachocodoner: Yes, but I decided to version it, since it can be risky and even small changes can break the command partially. It actually happened a few times.

italojs (approved):

PR #13698, opened 2025-04-15
Skip compilation for excluded archs further

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13665

This PR applies another optimization to ensure that in development mode, if legacy and Cordova architectures are excluded, either via `--exclude-archs` or `modernWebArchsOnly: true`, then related compile flows for those web architectures are also skipped.

While optimizing the bundler, I noticed that even when excluding these archs, `compileUnibuild` and other steps still run for `web.browser.legacy`. In contrast, `web.cordova` is already skipped in dev mode in many of these flows ([e.g](https://github.com/meteor/meteor/blob/48f365577743826fb532ae93e64c739754acf668/tools/isobuild/compiler.js#L181-L183)), to save time. So it makes sense to do the same for `web.browser.legacy`. `compileUnibuild` goes through Meteor packages and likely touches npm modules for both node and web archs, but in dev mode, if legacy and Cordova are ignored, this extra work is unnecessary.

This change saves extra milliseconds by avoiding work on unused archs. Profiling confirms the improvement.

image

> The results got from a specific machine using [fredmaiaarantes/simpletasks](https://github.com/fredmaiaarantes/simpletasks)

The next picture compares a localized compileUnibuild of a specific package, showing how skipping legacy improved build time for that individual package. It affected all them.

image

There may be other areas we can apply the same logic. **This should not affect production builds or package publishing**, so I will keep testing this changes and ensure don't affect these.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -545,7 +545,8 @@ async function doRunCommand(options) {
   }
 
   webArchs = filterWebArchs(webArchs, options['exclude-archs'], options.appDir, options);
-  global.webArchs = webArchs;
+  // Set the webArchs to include for compilation later
+  global.includedWebArchs = webArchs;
 
   const buildMode = options.production ? 'production' : 'development';
 
@@ -2514,7 +2515,8 @@ var runTestAppForPackages = async function (projectContext, options) {
     webArchs.push("web.cordova");
   }
   buildOptions.webArchs = filterWebArchs(webArchs, options['exclude-archs'], projectContext.appDirectory, options);
-  global.webArchs = buildOptions.webArchs;
+  // Set the webArchs to include for compilation later
+  global.includedWebArchs = buildOptions.webArchs;
 
   if (options.deploy) {
     // Run the constraint solver and build local packages.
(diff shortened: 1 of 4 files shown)


tools/cli/commands.js
```diff
@@ -2512,6 +2514,7 @@ var runTestAppForPackages = async function (projectContext, options) {
     webArchs.push("web.cordova");
   }
   buildOptions.webArchs = filterWebArchs(webArchs, options['exclude-archs'], projectContext.appDirectory, options);
+  global.webArchs = buildOptions.webArchs;
```
nachocodoner: This is the simplest way to share webarch info within the Meteor run and test context, and later reuse it in other parts of the bundler to skip unnecessary compilation. I see that global is used in the bundler to share data, but this could also be done by passing parameters down to the isopack compiler (similar to the [`includeCordovaUnibuild`](https://github.com/meteor/meteor/blob/48f365577743826fb532ae93e64c739754acf668/tools/project-context.js#L990) params [used there](https://github.com/meteor/meteor/blob/48f365577743826fb532ae93e64c739754acf668/tools/isobuild/compiler.js#L181-L183)). Not sure which is better, but this is definitely faster and works as a prototype to evaluate how useful the approach is.

I might change it later or keep it and just add comments.

9Morello:
Seems like a good ~10% reduction in build time overall. Good catch!

PR #13699, opened 2025-04-16
Support @parcel/watcher as the watcher alternative

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13657 and https://github.com/meteor/meteor/pull/13698

This PR continues the optimizations after switching to the [SWC transpiler](https://github.com/meteor/meteor/pull/13657) and [dropping legacy compilation](https://github.com/meteor/meteor/pull/13698). It focuses on refactoring the file watchers used by Meteor, replacing them with a modern, native, and cross-platform tool: **@parcel/watcher**.

The change was suggested by @zodern. After reviewing the involved code, I found that Meteor used two libraries: **Pathwatcher** on Linux and **vscode-NSFW** on Windows/macOS. On Windows and macOS, Meteor enabled a **single recursive watcher** per directory using `vscode-nsfw`, efficiently covering whole trees. On Linux, recursive watchers weren’t available, and adding one per subdirectory would overload the system. To avoid that, Meteor used `pathwatcher, watching files and some directories individually. This added up to 5 seconds of delay in Linux when reacting to file changes, especially for new files.

The **@parcel/watcher** library offers a modern solution for recursive watching across platforms. It uses native backends: FSEvents (macOS), inotify or Watchman (Linux), and the Windows API. All through a unified interface. It supports **recursive watching on all platforms**, which is the key improvement here.

Results can be seen using `METEOR_MONITOR_PROCESS=true` in `meteor profile`. This outputs the full profile process. The “Meteor(Total)” time shows only what the internal profiler sees, not including watcher reaction time. The new “Meteor(Progress)” output includes delays like startup wait and file change reactions. This metric helps confirm the improvements in this PR. See the next profiles for reference:

image

> The results got from a specific machine using [fredmaiaarantes/simpletasks](https://github.com/fredmaiaarantes/simpletasks) in a Linux-like OS

It seems to have improved the **prepareProjectForBuild** phase, likely because it involves setting up file watching. There's also a slight improvement in build and rebuild times. In some profiles, the `safeWatcher.watch` process appears as well.

This still needs more testing across all OS, but results so far look good.

---

- [x] Investigate why `./meteor --get-ready` fails; possible memory leak or limit with @parcel/watcher?
     - Apps tested locally worked well without issues.
- [x] Ensure to mark a path as being subscribed beforehand to avoid parallel repeated subscriptions to the same folder context
    - Any gain out of this change?
- [x] Remove LRU on file replacement since finally was not needed to fix an issue
- [x] Fix tests affected by file watcher replacement

### DIFF
--- a/tools/fs/safe-watcher.ts
+++ b/tools/fs/safe-watcher.ts
@@ -207,7 +207,6 @@ async function ensureWatchRoot(dirPath: string): Promise<void> {
     const subscription = await ParcelWatcher.subscribe(
         osDirPath,
         (err, events) => {
-          console.log("--> (safe-watcher.ts-Line: 212) events: ", events);
           if (err) {
             console.error(`Parcel watcher error on ${osDirPath}:`, err);
             // Only disable native watching for critical errors (like ENOSPC).
--- a/tools/fs/safe-watcher.ts
+++ b/tools/fs/safe-watcher.ts
@@ -1,7 +1,6 @@
 import { Stats } from 'fs';
 import { dirname } from "path";
 import ParcelWatcher from "@parcel/watcher";
-import LRUCache from 'lru-cache';
 
 import { Profile } from "../tool-env/profile";
 import { statOrNull, lstat, toPosixPath, convertToOSPath, pathRelative, watchFile, unwatchFile } from "./files";
@@ -52,27 +51,8 @@ interface Entry extends SafeWatcher {
   _fire(event: string): void;
 }
 
-// Set METEOR_WATCH_USE_LRU environment variable to a truthy value to
-// enable LRU caching for watcher entries to reduce memory usage.
-const useLRU = Boolean(JSON.parse(process.env.METEOR_WATCH_USE_LRU || "false"));
-
-// enable to tweak maximum entries size (default 500MB).
-const sizeMaxLRU = JSON.parse(process.env.METEOR_WATCH_USE_LRU_SIZE || Math.pow(2, 20) * 500);
-
 // Registry mapping normalized absolute paths to their watcher entry.
-// If LRU caching is enabled, this will be an LRUCache instance.
-// Otherwise, we use a Map object for entries.
-const entries = useLRU
-  ? new LRUCache<string, Entry | null>({
-      max: sizeMaxLRU, // 1MB max size
-      length: (entry, key) => {
-        return key.length + (entry ? 100 : 10);
-      },
-      dispose: (key, entry) => {
-        return entry?.close();
-      },
-    })
-  : new Map<string, Entry | null>();
+const entries = new Map<string, Entry | null>();
 
 function getEntry(path: string): Entry | null | undefined {
   return entries.get(path);
@@ -83,10 +63,7 @@ function setEntry(path: string, entry: Entry | null): void {
 }
 
 function deleteEntry(path: string): void {
-  if (useLRU) {
-    return entries.del(path);
-  }
-  return entries.delete(path);
+  entries.delete(path);
 }
 
 // Watch roots are directories for which we have an active ParcelWatcher subscription.
@@ -118,8 +95,7 @@ if (process.env.METEOR_WATCH_PRIORITIZE_CHANGED &&
 }
 
 // Set of paths for which a change event has been fired, watched with
-// watchLibrary.watch if available. This could be an LRU cache, but in
-// practice it should never grow large enough for that to matter.
+// watchLibrary.watch if available.
 const changedPaths = new Set;
 
 function shouldIgnorePath(absPath: string): boolean {
--- a/tools/fs/safe-watcher.ts
+++ b/tools/fs/safe-watcher.ts
@@ -122,6 +122,13 @@ function shouldIgnorePath(absPath: string): boolean {
   const posixPath = toPosixPath(absPath);
   const parts = posixPath.split('/');
 
+  const cwd = toPosixPath(process.cwd());
+  const isWithinCwd = absPath.startsWith(cwd);
+
+  if (isWithinCwd && absPath.includes(`${cwd}/.meteor/local`)) {
+    return true;
+  }
+
   // Check for .meteor: allow the .meteor directory itself,
   // but ignore its "local" subdirectory (or any immediate child folder that indicates cache).
   const meteorIndex = parts.indexOf(".meteor");
@@ -134,8 +141,6 @@ function shouldIgnorePath(absPath: string): boolean {
     // Otherwise, do not automatically ignore .meteor (which includes .meteor/packages, etc).
   }
 
-  const cwd = toPosixPath(process.cwd());
-  const isWithinCwd = absPath.startsWith(cwd);
   if (isWithinCwd) {
     return absPath.includes(`${cwd}/node_modules`);
   }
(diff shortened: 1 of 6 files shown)


tools/fs/safe-watcher.ts
```diff
@@ -1,476 +1,364 @@
...
+    ignoredWatchRoots.add(dirPath);
+    return;
+  }
 
-    // If a watcher event fired in the last polling interval, ignore
-    // this event.
-    if (Date.now() - lastWatcherEventTime > getPollingInterval()) {
-      fire("change");
+  // Set up ignore patterns to skip deep node_modules and .meteor/local cache
+  const ignorePatterns = ["**/node_modules/**", "**/.meteor/local/**"];
```
nachocodoner: I'm ignoring `node_modules` and `.meteor/local` in `ParcelWatcher` and with `shouldIgnorePath`, since in a new app, changes there don't trigger the Meteor bundler. I didn’t find this exclusion in the previous safe-watcher, but without it, profile times increase a lot.

tools/fs/safe-watcher.ts
```diff
@@ -1,476 +1,364 @@
...
+/**
+ * Fall back to polling. If a critical error occurs,
+ * we disable native watching and close all existing native watchers.
+ */
+function fallbackToPolling() {
+  if (watcherEnabled) {
+    console.error("Critical native watcher error encountered. Falling back to polling for all entries.");
+    watcherEnabled = false;
+    closeAllWatchers();
+  }
```
nachocodoner, line 441: I'm keeping the previous Meteor polling strategy as a fallback or configured by the `METEOR_WATCH_FORCE_POLLING` env. ParceLWatcher is supposed to handle native fallbacks across all OS, so this might not be needed anymore. Still, to avoid a breaking change for those using this in certain environments, we're leaving it as is.

tools/fs/safe-watcher.ts
```diff
@@ -1,476 +1,364 @@
...
-          if (oldEntry) {
-            oldEntry._fire('rename');
-          }
+  // If native file watching is disabled and the developer has
+  // explicitly opted out of the priority system, poll everything at the
+  // higher frequency (every 500ms by default). Note that this leads to
+  // higher idle CPU usage, so the developer may want to adjust the
+  // METEOR_WATCH_POLLING_INTERVAL_MS environment variable.
+  return NO_WATCHER_POLLING_INTERVAL;
+}
```
Grubba27, line 410: Both (`DEFAULT_POLLING_INTERVAL` and `NO_WATCHER_POLLING_INTERVAL`) are configured to be focused on performance, right?

What would be those settings if I'm on a machine with fewer resources(_maybe_ we could have high-perf mode and low one)?

Grubba27 (approved):

tools/fs/safe-watcher.ts
nachocodoner: This code was already in the previous watcher, I just moved it around.

Polling is a fallback strategy for current Meteor, only enabled manually with `METEOR_WATCH_FORCE_POLLING`, and the polling interval can be adjusted. Parcel’s native watching remains the default with my change. I initially planned to remove polling, but kept it in case someone needs it. It’s not expected to be used often, but if needed, the interval can be configured via `METEOR_WATCH_POLLING_INTERVAL_MS` as they can already do (by default I think is 5 seconds refering to the code).

italojs (approved):

nachocodoner:
Finally! This branch is ready to merge into 3.3. Next, I’ll test it more thoroughly across OS with all other changes in place.

tools/fs/safe-watcher.ts
```diff
@@ -1,476 +1,445 @@
...
-
-    statWatcher.stat = newStat;
-  }
-
-  return statWatcher;
-}
+      // If native watching is disabled, use the polling strategy.
+      if (!watcherEnabled) {
+        return startPolling(absPath, callback);
+      }
```
zodern, line 306: If I understand this code correctly, it only uses either native watching, or polling, but never both. Meteor's tried this before, and there's too many situations where it does not work correctly (for example, when using network drives, vm's with volumes, and various other situations native watchers will not detect changes). Meteor used both approaches together in the past for reliability.

tools/fs/safe-watcher.ts
```diff
@@ -1,476 +1,445 @@
...
+      // Try to reuse an existing entry if one was created before.
+      let entry = getEntry(absPath);
+      if (!entry) {
```
zodern, line 309: This is an interesting approach to recursive file watching. When I originally implemented it, it worked by:

- it still used polling on the individual files and folders for reliable file watching.
- Isobuild created recursive watchers for the app, and each local package, so when they worked it would be notified of changes faster. Anything they miss would be catched by the polling.

I had considered doing something similar (checking for any new watch roots and adding them as needed) but had assumed it would add too much overhead with how many paths Meteor watches. Have you checked how much time is spent by `ensureWatchRoot`? VSCode uses a similar approach to file watching, but the implementation is different to have good performance.

Recursive file watchers don't detect anything in symlinks on MacOS and Windows, which is another reason why Meteor still used polling. I don't see anything in this PR to handle symlinks. It would need to create a new watch root for each symlinked folder.

tools/fs/safe-watcher.ts
nachocodoner: I wasn't fully sure about these. I aimed to simplify the approach, and since there’s test coverage, I felt more confident supporting just one and cleaning up the code. Your concerns make sense. I expected polling to be a fallback on any error or when user decide it, but in the cases you mention, like using a volume or network driver, do you think the watcher would actually trigger an error to fall back? I might need to test this scenario.

I could make both work together again, but the goal was to simplify and focus on performance, without causing known regressions, as you rightly pointed out.

zodern: In those situations, there is no way to know if native watching works or not - there is no error, and sometimes changes are detect and sometimes changes aren't (for example, mounting a folder into a VM might detect changes made in the VM, but not changes made in the host machine). The past few years some projects such as docker have improved support for native events with volumes, but there's still many exceptions.

tools/fs/safe-watcher.ts
nachocodoner: > Have you checked how much time is spent by ensureWatchRoot

I haven’t measured that function specifically. But overall, file watching times and global stage/s where used seem lower now. But it could be related by other reasons.

> I don't see anything in this PR to handle symlinks. It would need to create a new watch root for each symlinked folder.

I looked for symlink handling in the original code but didn’t find any at the beginning. Maybe I missed it. I also didn’t see clear test coverage for it, or if it’s there, tests are currently passing. If needed, I can add support for symlinks as you suggested. Falling back to polling for symlinks.

------

By the way, do you think [@parcel/watcher-wasm](https://www.npmjs.com/package/@parcel/watcher-wasm) could be useful? If it runs on Node internally, it might work like our polling strategy. [Chokidar could be another option.](https://github.com/meteor/meteor/pull/13703). Just to unify the API when fall back to polling/Node.

tools/fs/safe-watcher.ts
zodern: I wrote a blog post a few years ago that goes into more details on Meteor's file watching : https://zodern.me/posts/meteor-file-watching/

nachocodoner: Do you know a practical case? Like, if I run Meteor in a mounted volume, would I hit the issues you mention when doing file changes from the host? Or maybe connecting containers/WSL from an IDE on the host machine? Any daily example you can think of would be helpful to try. Though, this might not be a very common setup for most developers, and maybe other frameworks don’t even support it (or behind a opt-in config for non-common usage scenarios).

In any case, your suggestion is to start the polling mechanism for each root we handle natively, or even per file, at the same time, right? This will impact CPU performance much in most usage scenarios, an opt-in option I still see it as a better option.

PR #13704, opened 2025-04-22
Add TypeScript declarations for isModern and getMinimumBrowserVersions functions

welkinwong:
This PR supplements the TypeScript definitions in the module. Currently, the TS definitions only include `setMinimumBrowserVersions`, but lack the following:

- isModern
- getMinimumBrowserVersions
- calculateHashOfMinimumVersions

The changes focus **solely on adding the missing type declarations** without modifying any existing logic or behavior.

### DIFF
--- a/packages/modern-browsers/modern.d.ts
+++ b/packages/modern-browsers/modern.d.ts
@@ -1,4 +1,12 @@
+export declare function isModern(
+  browser: { name: string, major: number, minor?: number, patch?: number }
+): boolean;
+
 export declare function setMinimumBrowserVersions(
   versions: Record<string, number | number[]>,
-  source: string
+  source?: string
 ): void;
+
+export declare function getMinimumBrowserVersions(): Record<string, Record<string, number | number[]>>;
+
+export declare function calculateHashOfMinimumVersions(): string;


9Morello:
We have some related type definitions that got added with 3.3: https://github.com/meteor/meteor/blob/release-3.3/packages/server-render/server-render.d.ts

Specifically, there's a type for an `IdentifiedBrowser` there.

I wonder what would be a good way to share them between those packages. @nachocodoner any ideas?

packages/modern-browsers/modern.d.ts
```diff
@@ -1,4 +1,12 @@
...
+  browser: { name: string, major: number, minor?: number, patch?: number }
+): boolean;
+
 export declare function setMinimumBrowserVersions(
   versions: Record<string, number | number[]>,
-  source: string
+  source?: string
 ): void;
+
+export declare function getMinimumBrowserVersions(): Record<string, Record<string, number | number[]>>;
```
italojs, line 10: From what @9Morello mentioned, is it the same as IdentifiedBrowser? What about if we export that?

9Morello: @italojs that type is slightly different, but the type for the `browser` argument in `isModern` is an `IdentifiedBrowser`.

nachocodoner (approved):
I’m not sure how to share types between packages. We need to consider how package types behave and are shared.

But for now, this PR is ready to merge for 3.3.1.

PR #13705, opened 2025-04-22
Remove recursive deferred in bundler code

nachocodoner:
This PR applies [a folder removal method](https://github.com/meteor/meteor/pull/13680/commits/0e97299a96bcab9c87de32d69132f2b69edd31bc) from @italojs, which only renames the folder instead of waiting for full removal. This is usually enough for caches or most file contexts.

I tested it in one spot; it's clearly faster. While it may not help much in single cases, parts of the code remove Meteor package contexts in a loop, where the speedup will be  more noticeable. I don't think we're dealing with many milliseconds of improvement, but any improvement is worth a refactor.

Just a single spot and the proof is faster:

image

---

We could apply in other parts of the code, basically replacing `rm_recursive` per `rm_recursive_deferred`, but my plan is just to focus on bundler code.

### DIFF
--- a/tools/isobuild/isopack-cache.js
+++ b/tools/isobuild/isopack-cache.js
@@ -83,19 +83,19 @@ export class IsopackCache {
       // Wipe specific packages.
       for (const packageName of packages) {
         if (self.cacheDir) {
-          await files.rm_recursive(self._isopackDir(packageName));
+          await files.rm_recursive_deferred(self._isopackDir(packageName));
         }
         if (self._pluginCacheDirRoot) {
-          await files.rm_recursive(self._pluginCacheDirForPackage(packageName));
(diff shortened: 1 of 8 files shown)


tools/isobuild/isopack-cache.js
```diff
@@ -351,7 +351,7 @@ export class IsopackCache {
         } else {
           // Nope! Compile it again. Give it a fresh plugin cache.
           if (pluginCacheDir) {
-            await files.rm_recursive(pluginCacheDir);
+            await files.rm_recursive_deferred(pluginCacheDir);
```
nachocodoner, line 354: Many of these rm_recursive happens in a loop of all Meteor packages, which will get the most of gain.

italojs (approved):

PR #13707, opened 2025-04-24
Symlink support on new watcher

nachocodoner:
Continues: https://github.com/meteor/meteor/pull/13699

This PR addresses part of the feedback from @zodern in the previous PR. It improves support for symbolic links by detecting symlinks and all traversed files, using a polling strategy to reliably watch for changes. I also enabled support for direct packages linked with npm link in the project's node_modules. I tested these and they work well.

Polling is used because native recursive watching isn't supported consistently across all OS when using symlinks, so polling is applied in these cases. Tools like [chokidar ](https://github.com/paulmillr/chokidar) could handle this more smartly, but for now this approach works well enough.

I considered enabling native watching and polling for all project files simultanously, not just symlinks. However, polling puts significant load on the CPU and impacts performance. So for now, polling remains opt-in via an environment variable (`METEOR_WATCH_FORCE_POLLING` and `METEOR_WATCH_POLLING_INTERVAL_MS`).

Most modern bundlers rely on native watchers, and for edge cases like working with volumes or network files, you can opt into polling. See [Vite](https://v2.vitejs.dev/config/#server-watch) and [rspack](https://rspack.dev/config/watch#watchoptionspoll) as examples. This favors performance in common scenarios while keeping compatibility available when needed.

Since Meteor 3.4 will favor using a modern bundler for app and node_modules code, it makes sense to align with how they handle things. As app code will be managed by the bundler, its options will become the new configuration standard for modern Meteor apps, including watching. Using `@parcel/watcher` still makes sense on Meteor core, as it improves file change detection for atmosphere packages code and intermediate outputs from the modern bundler.

---

We'll mention in Meteor 3.3 that this could be a breaking change in cases like using WSL2 from host machine, working with volumes, or over network setups, where polling needs to be enabled manually. I'll test some of these cases too. It’s also a chance to collect feedback, but since performance is the priority, and this aligns with other modern tools, this seems like the right path.

### DIFF
--- a/tools/fs/safe-watcher.ts
+++ b/tools/fs/safe-watcher.ts
@@ -153,24 +153,25 @@ function shouldIgnorePath(absPath: string): boolean {
   }
 
   // For project node_modules: check if it's a direct node_modules/<package>
-  if (isWithinCwd && absPath.includes(`${cwd}/node_modules`)) {
-    // Check if it's a direct node_modules/<package> path
-    const relPath = absPath.substring(cwd.length + 1); // +1 for the slash
-    const relParts = relPath.split('/');
-    if (relParts.length >= 2 && relParts[0] === 'node_modules') {
-      // If it's a direct node_modules/<package>, check if it's a symlink
-      // We'll return false here (don't ignore) so that the code can later decide to use polling
-      // based on isSymbolicLink check in the watch function
-      if (relParts.length === 2 && isSymbolicLink(absPath)) {
-        return false;
+  if (isWithinCwd) {
+    if (absPath.includes(`${cwd}/node_modules`)) {
+      // Check if it's a direct node_modules/<package> path
+      const relPath = absPath.substring(cwd.length + 1); // +1 for the slash
+      const relParts = relPath.split('/');
+      if (relParts.length >= 2 && relParts[0] === 'node_modules') {
+        // If it's a direct node_modules/<package>, check if it's a symlink
+        // We'll return false here (don't ignore) so that the code can later decide to use polling
+        // based on isSymbolicLink check in the watch function
+        if (relParts.length === 2 && isSymbolicLink(absPath)) {
+          return false;
+        }
+        // Check if it's within a symlink root to not ignore
+        if (isWithinSymlinkRoot(absPath)) {
+          return false;
+        }
       }
-      // Check if it's within a symlink root to not ignore
-      if (isWithinSymlinkRoot(absPath)) {
-        return false;
-      }
-      return true;
     }
-    return true;
+    return false;
   }
 
   // For external node_modules: check if it's a direct node_modules/<package>
--- a/tools/fs/safe-watcher.ts
+++ b/tools/fs/safe-watcher.ts
@@ -170,8 +170,10 @@ function shouldIgnorePath(absPath: string): boolean {
           return false;
         }
       }
+      return true;
+    } else {
+      return false;
     }
-    return false;
   }
 
   // For external node_modules: check if it's a direct node_modules/<package>
--- a/tools/fs/safe-watcher.ts
+++ b/tools/fs/safe-watcher.ts
@@ -2,7 +2,7 @@ import { Stats } from 'fs';
 import ParcelWatcher from "@parcel/watcher";
 
 import { Profile } from "../tool-env/profile";
-import { statOrNull, toPosixPath, convertToOSPath, pathRelative, watchFile, unwatchFile, pathResolve, pathDirname } from "./files";
+import { statOrNull, lstat, toPosixPath, convertToOSPath, pathRelative, watchFile, unwatchFile, pathResolve, pathDirname, realpathOrNull, readdir, pathJoin } from "./files";
 
 // Register process exit handlers to ensure subscriptions are properly cleaned up
 const registerExitHandlers = () => {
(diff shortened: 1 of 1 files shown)


tools/fs/safe-watcher.ts
```diff
@@ -155,6 +185,51 @@ function shouldIgnorePath(absPath: string): boolean {
...
+ * Check if a path is a symbolic link.
+ * 
+ * Symbolic links are not supported natively in some operating systems,
+ * so we need to use polling for them to ensure they are properly watched.
+ * This function is used to determine if a path is a symbolic link,
+ * so we can use polling instead of native watching for it.
+ * 
+ * If a path is a symbolic link, its root is added to the symlinkRoots set.
+ */
+function isSymbolicLink(absPath: string, addToRoots = true): boolean {
```
zodern, line 206: This seems to work when watching something that is a symbolic link. It's possible though that the path being watched isn't a symbolic link, but is inside of one. For example:

- app
   - a (symbolic link to ../shared)
      - b

B is not a symbolic link so `isSymbolicLink` would return false, but since it is inside of one (it's actual path is ../../shared/b) changes will not be detected by the recursive watcher on `app`. This is handled by `isWithinSymlinkRoot`, but only if the watcher also watches `a`, and starts watching `a` before `b`.

tools/fs/safe-watcher.ts
```diff
@@ -141,12 +152,31 @@ function shouldIgnorePath(absPath: string): boolean {
...
+      return true;
+    } else {
+      return false;
+    }
   }
 
-  // For external node_modules: ignore contents
+  // For external node_modules: check if it's a direct node_modules/<package>
   const nmIndex = parts.indexOf("node_modules");
   if (nmIndex !== -1) {
```
zodern, line 183: Will this prevent rebuilding when a local npm package used by a local meteor package changes?

tools/fs/safe-watcher.ts
nachocodoner: Completely 🤔

Good catch, I should find a fix for it.

tools/fs/safe-watcher.ts
nachocodoner: Likely yes. If I remember right, npm dependencies from Meteor packages are always within a .npm context, but still inside a node_modules folder, right? I should check for that .npm context here. Thanks!

zodern: That would be great! It's already very buggy when trying to develop a local npm package and meteor package together, and it would be nice if it doesn't get any worse.

tools/fs/safe-watcher.ts
nachocodoner: I pushed the change. If a new symbolic link root is found, I’ll rewatch any previously watched entries within the symlink folder context and use polling for reliable watching.

This avoids impacting performance further, as rewatching only happens when a symbolic link is encountered in the project, which won't happen in most of the scenarios.

I tested this by altering the code and switching absPath between an example symlinked folder and a file inside it during the watch function. Usually, glob watches the folder first, I think, so `isWithinSymlinkRoot` should work major of times. Either way, it works fine now if the content is watched first.

tools/fs/safe-watcher.ts
nachocodoner: Pushed the fix. Tested with packages both inside and outside the project, ~works well reacting.~ It always works when updating files in the local package, but not when modifying npm dependencies directly within .npm (it doesn't trigger the refresh client/server, but the @parcel/watch events are fired). Changes in .npm refresh only after the first refresh following a package file update. Debugging to undestand why.

Besides, I found an infinite loop that sometimes happens when adding a new npm dependency in `Npm.depends` to a package. Not sure how to fix it. Any idea? Could it be from something in your earlier implementation?

nachocodoner: **Recap issues:**

- .npm change issue:
    - Changes in .npm trigger @parcel/watch events but don’t refresh client/server unless a local package file is updated.
    - **Also happens in current Meteor implementation.** Likely due to a default exclusion of ./hidden folders (see [code](https://github.com/meteor/meteor/blob/48f365577743826fb532ae93e64c739754acf668/tools/isobuild/package-source.js#L1184)). Oddly, after one refresh, starts working. I’ll look into this to avoid skipping hidden folders like .npm.


- Infinite loop issue:
    - Adding a dependency via Npm.depends may cause an infinite loop.
    - Happens only in the new implementation. In current version, I saw multiple refreshes but no loop.

nachocodoner: Local packages in `./packages/<dep>` in your app folder behave differently between a local checkout and a published Meteor version. With both the new and old watcher, changes to `.npm/package/node_module` dependencies don’t trigger refreshes in the local checkout version. Published versions do refresh. So the first issue isn’t a regression.

The infinite loop issue still happens sometimes.

nachocodoner: The new watcher module seems to behave similarly to the previous implementation. I’ve noticed what you likely referred to as buggy when developing with local npm and Meteor packages together (via `.npm/package/node_modules`), sometimes it reacts or refresh, sometimes it doesn’t, depending on using checkout or published versions. I’ve seen this in both the new watcher using `@parcel/watcher` and the previous approach. Maybe the new version works more often.

During this exploration, I also got a better understanding of other parts of the bundler, which is great. But specifically for the watcher, and after reviewing the previous implementation, I’m planning to bring it back for a legacy configuration (`legacy-safe-watcher.ts` enabled by `METEOR_WATCHER_LEGACY=1`).
**By default, it will use the @parcel/watcher**, but through an env flag, it can switch to the legacy approach.

I feel confident doing this, at least for now, since if issues arise in the official Meteor 3.3, users will be able to fall back to a known working version. It also helps cover the scenarios you mentioned that the old watcher could handle better while we have time to test it deeply using the newer. Plus, it will be easier to compare issues in both implementations, as the switch is just an env change. I think this setup will help during the transition period until we gather enough feedback on the new approach.

PR #13709, opened 2025-04-25
SWC - minifier-js (approach 1)

italojs:
Summary

In alignment with our ongoing effort to [make the Meteor bundler faster](https://forums.meteor.com/t/join-the-effort-to-speed-up-meteor-bundler/63406/47), this PR introduces a Proof of Concept (PoC) for integrating swc into Meteor’s build process. We're leveraging the excellent work done in the community by [zodern/minify-js-sourcemaps](https://github.com/zodern/minify-js-sourcemaps), and as a first step, we’re experimenting with internalizing its functionality into the core.

Status

🚧 This is a draft PR and should be considered a PoC only.

Next Steps

Discuss the best approach to integrate community-driven solutions into the core.

Ensure proper attribution and credit to community maintainers.

Address any limitations or issues found during early testing.

Finalize the approach in collaboration with the community (a discussion channel will be opened soon).

Screenshot 2025-04-25 at 17 25 19

Future edits are coming...


packages/source-maps/readme.md
```diff
@@ -0,0 +1,154 @@
+# @zodern/source-maps
```
zodern, line 1: I'm curious why you are forking my npm package? I would prefer if you use the npm package instead, though forking is of course okay. I would ask that you also copy over the license file: https://github.com/zodern/source-maps/blob/main/LICENSE. I realized it was missing one.

packages/standard-minifier-js/package.js
```diff
@@ -9,10 +9,16 @@ Package.registerBuildPlugin({
...
     'minifier-js',
-    'ecmascript'
+    'ecmascript',
+    'caching-minifier',
+    'source-maps',
   ],
   npmDependencies: {
-    "@babel/runtime": "7.18.9",
+    'meteor-package-install-swc': '1.1.2',
+    'acorn': '8.10.0',
```
zodern, line 18: You'll also want to copy over the new code for stats, which uses acorn.

packages/minifier-js/package.js
```diff
@@ -1,22 +1,17 @@
 Package.describe({
```
zodern, line 1: I wouldn't recommend copying over my fork of minifier-js as is. I no longer use it, and it is outdated compared to the version in core.

packages/caching-minifier/caching-compiler.js
```diff
@@ -0,0 +1,200 @@
+// Copied from https://raw.githubusercontent.com/meteor/meteor/devel/packages/caching-compiler/caching-compiler.js
```
zodern, line 1: This file was copied from the caching-compiler core package. It probably would be good to either import the file from there, or move the caching minifier class into that package.

I didn't do either because I was trying to avoid my minifier from having a dependency on the ecmascript package, since it added 45mb.

packages/hide-production-sourcemaps/README.md
```diff
@@ -0,0 +1,14 @@
+# zodern:hide-production-sourcemaps
```
zodern, line 1: In the past, it was decided that if Meteor generates production source maps by default, it should also hide them by default. Instead of copying this package over, it would probably be better to make this the default behavior in the `webapp` package, with a way to disable it.

zodern (commented):
I left some notes on the packages.

packages/source-maps/readme.md
italojs: It's just a POC as mentioned into pr description, it isn't the final implementation. We still need to discuss internally (and publicly later) the best approach to internalize community packages(will do Monday), so don't worry; we will change this implementation ;)

btw: I will leave your suggestion for the meteor core team

packages/minifier-js/package.js
italojs: Yes, after discussing the best way to bring community packages to the core and implementing them as is, I will try to resolve the current issues present there and modernize the code.  It's a future job, but you will be included in the process for sure

packages/caching-minifier/caching-compiler.js
italojs: Good insight!

packages/hide-production-sourcemaps/README.md
italojs: Good insight!

packages/minifier-js/package.js
italojs: I created another and simpler approach https://github.com/meteor/meteor/pull/13715 bringing only the swc

italojs:
I created another and simpler approach https://github.com/meteor/meteor/pull/13715, bringing only the swc 
want to bring the rest in the future, but let's simplify it for now

Issue #13652, opened 2025-03-14
Error when observing the cursor: TypeError: this.documents?.clear is not a function


welkinwong:
## Observed Errors:

**version: METEOR@3.1.2**

After I change the subscription, there is a certain probability of errors occurring, and once an error occurs, it will continue to report errors at intervals.

```
Exception while polling query {"collectionName":"name","selector":{"type":"a","ownerId":"123456789"},"options":{"transform":null,"sort":{"accessAt":-1},"limit":200,"projection":{"_id":1,"referenceId":1}}} TypeError: this.documents?.clear is not a function
    at FindCursor.rewind (/Users/name/.meteor/packages/npm-mongo/.6.10.2.1sx2v8ti8rif++os+web.browser+web.browser.legacy+web.cordova/npm/node_modules/mongodb/src/cursor/abstract_cursor.ts:660:21)
    at AsynchronousCursor._rewind (packages/mongo/asynchronous_cursor.js:120:20)
    at AsynchronousCursor.forEach (packages/mongo/asynchronous_cursor.js:99:10)
    at AsynchronousCursor.map (packages/mongo/asynchronous_cursor.js:111:16)
    at AsynchronousCursor.fetch (packages/mongo/asynchronous_cursor.js:131:17)
    at AsynchronousCursor.getRawObjects (packages/mongo/asynchronous_cursor.js:147:19)
    at PollingObserveDriver._pollMongo (packages/mongo/polling_observe_driver.ts:153:39)
    at Object.task (packages/mongo/polling_observe_driver.ts:111:18)
    at AsynchronousQueue._processOrderedBatch (packages/meteor.js:915:41)
    at AsynchronousQueue._run (packages/meteor.js:873:18)
    at Immediate.<anonymous> (packages/meteor.js:850:14)
    at processImmediate (node:internal/timers:491:21)
    at process.topLevelDomainCallback (node:domain:161:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24)
```

![Image](https://github.com/user-attachments/assets/db4874d0-a13c-4b18-912f-cf3c24ce2ba8)

![Image](https://github.com/user-attachments/assets/3fa0074c-7f2f-4687-9c2d-56a7c2947138)

Before this error occurred, I had just re-imported the database, and I'm not sure if it's related.

## Impact:

Apart from the continuous error messages, no other impacts have been observed.

## Request for Help:

Any help or pointers would be highly appreciated!

welkinwong:
[...]

PR #13681, opened 2025-03-31
Release 3.3

nachocodoner:
## Still in development (next Meteor 3.3 official)

Current released version: 3.3-rc.0.

To use it, please do a

``` sh
meteor update --release 3.3-rc.0
```

``` sh
meteor create --release 3.3-rc.0
```

**Add this to your `package.json` to enable the new modern build stack:**

```json
"meteor": {
  "modern": true
}
```

> This setting is on by default for new apps.

**Announcements:**

- **Last:** [Weekly Update (May 28, 2025): New 3.3-rc.0 Release, Faster Builds Delivered! 🚀](https://forums.meteor.com/t/weekly-update-may-29-2025-new-3-3-rc-0-release-faster-builds-delivered/63612)

- [Weekly Update, April 16, 2025 – Faster Bundle Times with 3.3-beta.0! 🚀](https://forums.meteor.com/t/weekly-update-april-16-2025-faster-bundle-times-with-3-3-beta-0/63494)

PR #13710, opened 2025-04-25
Fix for Issue #13652 caused by MongoDB Node.js Driver

welkinwong:
This PR primarily addresses the bug reported in #13652. The issue was caused by the MongoDB Node.js Driver and has been fixed in version v6.16.0.

In Commit 3dac12a, I noticed a temporary version downgrade from 6.10.0 to 6.9.0 due to legacy database connection issues.  However, after reviewing the 6.10.0 changelog, the only documented breaking change was the dropped compatibility with MongoDB 3.6.  No other critical issues were mentioned in the release notes.

If the issue was solely related to pre-MongoDB 4.x compatibility, I believe we should proceed with the upgrade. According to MongoDB's official support policy ([MongoDB Software Lifecycle Schedules](https://www.mongodb.com/legal/support-policy/lifecycles)), versions below 4.x are already end-of-life (EOL) and no longer supported. Maintaining compatibility with deprecated database versions would introduce unnecessary technical debt and potential security risks.

### DIFF
--- a/packages/npm-mongo/package.js
+++ b/packages/npm-mongo/package.js
@@ -8,7 +8,7 @@ Package.describe({
 });
 
 Npm.depends({
-  mongodb: "6.9.0"
+  mongodb: "6.16.0"
 });
 
 Package.onUse(function (api) {


italojs (approved):

nachocodoner:
We had to downgrade the Mongo driver to a specific version because it caused breaking changes in a few apps, including Galaxy, where data was not published properly in an edge case. We should be cautious about including this in 3.3, research and test it beforehand.

It would be a pity if existing apps were blocked by a Mongo issue and couldn't benefit from the performance improvements. It might be better to delay it to a future version, with more focus in Mongo.

nachocodoner:
[I’ve reverted the MongoDB upgrade](https://github.com/meteor/meteor/pull/13681/commits/86fc826b9379c86c51ef1334c5d679b12c60150a) and will postpone it to the next beta, either 3.3.1-beta or 3.4-beta. Upgrading to later versions from mongodb `6.0.9` caused issues in some community apps without clear reproduction steps. In the next beta cycle, we’ll give Meteor apps an isolated window to test the Mongo upgrade and help us reproduce and fix any issues. That way, Meteor 3.3 can focus on build enhancements and any major fixes on this area if araise.

@welkinwong: Feel free to open a new PR with this. After the official Meteor 3.3 we can create a beta for this upgrade. We'll do it otherwise.

PR #13711, opened 2025-04-25
Favor Reify transpile in SWC all time

nachocodoner:
Fixes: https://forums.meteor.com/t/weekly-update-april-16-2025-faster-bundle-times-with-3-3-beta-0/63494/20

This PR ensures the reify output on SWC transpilation is always favored to ensure top-level await works as expected in all scenarios.

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -91,9 +91,7 @@ function compileWithSwc(source, swcOptions = {}, { inputFilePath, features, arch
         generateLetDeclarations: true,
       }),
     });
-    if (!result.identical) {
-      content = result.code;
-    }
+    content = result.code;
 
     return {
       code: content,


packages/babel-compiler/babel-compiler.js
```diff
@@ -91,9 +91,7 @@ function compileWithSwc(source, swcOptions = {}, { inputFilePath, features, arch
         generateLetDeclarations: true,
       }),
     });
-    if (!result.identical) {
-      content = result.code;
-    }
+    content = result.code;
```
nachocodoner, line 94: I used `result.identical` as the pattern used in other code parts with Babel + Reify. However, with SWC + Reify, we should always favor Reify for better compatibility. It seems that "browser-policy-content" triggered an "identical" case (likely falsy one), causing the SWC version with top-level await (TLA) to crash. Reify ensures backward compatibility here with TLA support.

PR #13712, opened 2025-04-26
Handle all-zero version cases in modern browser detection

welkinwong:
When all version numbers are 0, this typically comes from in-app WebView UAs (e.g., iOS WKWebView).
We can let users decide whether to treat it as a modern browser
via the packageSettings.unknownBrowsersAssumedModern option.


welkinwong:
A user has reported issues with embedding Meteor URLs in the app:
https://forums.meteor.com/t/modern-browsers-and-ios-wkwebview/62613

StorytellerCZ (approved):

italojs (approved):

PR #13718, opened 2025-04-30
Refactor modern config at package.json

nachocodoner:
This PR refactors the `package.json` config to use a single setting that enables all modern changes by default: new transpiler, web arch-only compilation, watcher, and future updates. Users can still opt out of specific features if needed.


Grubba27 (approved):
This does not break anything?

We have not released any `modern*` option yet, right?

nachocodoner:
> We have not released any modern* option yet, right?

No, only as part of 3.3-beta.0 first beta. We are fine to change it and report these for last beta devs.

PR #13723, opened 2025-05-01
fix: corrected Roles.getUsersInRoleAsync documentation

### LINKED ISSUE #13700: [Documentation] Roles.getUsersInRoleAsync is wrong
Seen in Meteor 3.2
[Roles.getUsersInRoleAsync](https://github.com/meteor/meteor/blob/48f365577743826fb532ae93e64c739754acf668/packages/roles/roles_common_async.js#L1054) returns a Cursor over `Meteor.users`. Hence [the documentation](https://docs.meteor.com/packages/roles#Roles-getUsersInRoleAsync) is wrong because it suggets that it is returning an array of users.

Either the code should be updated or the documentation fixed. 
Updating the code will result in a breaking change

The correct documented usage should be something like : 
```javascript 
const adminUsers = await (await Roles.getUsersInRoleAsync("admin")).fetchAsync();
```

I suggest anyone taking this issue to review other methods of Roles as it is likely that there are more of such mistakes

shaikhFaris:
Followed the instructions given in the issue and updated the documentation.

closes #13700

### DIFF
--- a/packages/roles/roles_common_async.js
+++ b/packages/roles/roles_common_async.js
@@ -1040,7 +1040,7 @@ Object.assign(Roles, {
   },
 
   /**
-   * @summary Retrieve all users who are in target role.
+   * @summary Retrieve a cursor of all users who are in the target role.
    * @memberof Roles
    * @locus Anywhere
    * @param {Array|String} roles Name of role or an array of roles.
--- a/v3-docs/docs/packages/roles.md
+++ b/v3-docs/docs/packages/roles.md
@@ -226,7 +226,7 @@ Example:
 
 ```js
 // returns a cursor of all admin users
-const adminUsers = await Roles.getUsersInRoleAsync("admin");
+const adminUsersCursor = await Roles.getUsersInRoleAsync("admin");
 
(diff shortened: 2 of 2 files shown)


StorytellerCZ (changes requested):
While technically correct, the change should instead be in the docs to change the description that cursor is returned. 
The `fetchAsync` might be one example, but I would be against that as these functions are more intended to be used in publications where you return the cursor.

shaikhFaris:
Ok @StorytellerCZ ,so I think I should change the description of the function and then add the example of getting the list of users using this cursor. Am I correct? Also, please let me know if something else needs to be added
How many examples can be given at max, in the docs of this function?

StorytellerCZ:
Yes, update the description in the DocToc above the function to mention cursor (it is in the return type anyway).
There is no limit on examples, but be reasonable. So I think the first one should be publication and then the rest could be the way you have done it.

StorytellerCZ (dismissed):

shaikhFaris:
I made some mistake, I will try to fix it.

PR #13721, opened 2025-04-30
Refactor IP address parsing on Webapp

italojs:
fix: https://github.com/meteor/meteor/issues/13713

PR #13728, opened 2025-05-05
New envs for package development

nachocodoner:
This PR prepares environment variables for upcoming dynamic atmosphere packages that enable specific features. They can be used in pre, during, or post build phases to adjust project behavior.

- `global.currentCommand`: the context of the running command.

- `METEOR_FORCE_INCLUDE_ARCHS` and `METEOR_FORCE_EXCLUDE_ARCHS`: include or exclude web.archs.

- `METEOR_CORDOVA_DISABLE`: skip the Cordova setup and running in `meteor run android` and `meteor run ios` commands, and alike.

These variables [hint at upcoming features](https://forums.meteor.com/t/building-capacitor-mobile-app-from-meteor-react/63560/4). 😁

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -596,7 +596,8 @@ async function doRunCommand(options) {
   const buildMode = options.production ? 'production' : 'development';
 
   let cordovaRunner;
-  if (!_.isEmpty(runTargets)) {
+  const shouldDisableCordova =  Boolean(JSON.parse(process.env.METEOR_CORDOVA_DISABLE || 'false'));
+  if (!shouldDisableCordova && !_.isEmpty(runTargets)) {
 
     async function prepareCordovaProject() {
       import { CordovaProject } from '../cordova/project.js';
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -332,9 +332,21 @@ function filterWebArchs(webArchs, excludeArchsOption, appDir, options) {
         const excludeArchs = [...excludeArchsOptions, ...automaticallyIgnoredLegacyArchs];
         webArchs = webArchs.filter(arch => !excludeArchs.includes(arch));
       }
-      return webArchs;
     }
   }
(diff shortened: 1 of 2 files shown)


tools/cli/commands.js
```diff
@@ -332,9 +332,21 @@ function filterWebArchs(webArchs, excludeArchsOption, appDir, options) {
...
+
+  const forcedInclArchs = process.env.METEOR_FORCE_INCLUDE_ARCHS;
+  if (forcedInclArchs != null) {
+    const nextInclArchs = forcedInclArchs.trim().split(/\s*,\s*/);
+    webArchs = Array.from(new Set([...webArchs, ...nextInclArchs]));
+  }
+
+  const forcedExclArchs = process.env.METEOR_FORCE_EXCLUDE_ARCHS;
+  if (forcedExclArchs != null) {
+    const nextExclArchs = forcedExclArchs.trim().split(/\s*,\s*/);
```
italojs, line 346: This regex is vulnerable; consider following this approach. https://github.com/meteor/meteor/pull/13721/files

italojs (changes requested):

tools/cli/commands.js
nachocodoner: Yes, I noticed that too. I just copied it from original code [added here](https://github.com/meteor/meteor/pull/13665/files#diff-a65d7a89d948df5e2e7122e62391d58d20521d447c7ebbf2c4f22faf0d12f048R277). But then I considered how critical it is in this specific case, it's only relevant at the Meteor tool level, which is a dev-only tool. So really, the only risk is if developers want to hack themselves. 😂 I do understand the concern when it comes to a production server, which is a different case.

That said, even if it’s not critical here, the code can be adjusted anyway.

italojs: LGTM

italojs (approved):

PR #13730, opened 2025-05-05
SWC resolves custom base url properly

nachocodoner:
Context: https://forums.meteor.com/t/weekly-update-april-16-2025-faster-bundle-times-with-3-3-beta-0/63494/29

This PR fixes an issue where using a custom `baseDir` for alias definitions causes failures. The fix ensures that the absolute path is correctly resolved when a `baseDir` is provided in the custom `.swcrc` configuration.

image


Grubba27 (approved):
In the future we could have the .swcrc file in our starters with this kind of aliasing, it feels more modern idk

italojs (approved):

Issue #13545, opened 2025-01-14
Release 3.1 this._makeNewID is not a function when calling insertAsync on startup


jankapunkt:
In Meteor 3. (`--release=3.1`) 1 I have the issue that startup is prevented when calling a collection's `insertAsync`.

Repro:

```js
import { Mongo } from 'meteor/mongo`

const collection = new Mongo.Collection('cow')

Meteor.startup(async () => {
  await collections.insertAsync({ what: 'moo' })
})
```

this throws on startup

```shell
I20250114-16:54:32.459(1)? error on boot.js TypeError: this._makeNewID is not a function
I20250114-16:54:32.460(1)?     at ns.Collection._insertAsync (packages/mongo/collection/methods_async.js:60:24)
I20250114-16:54:32.460(1)?     at ns.Collection.insertAsync (packages/mongo/collection/methods_async.js:104:17)
I20250114-16:54:32.460(1)?     at Object.Legal.helpers.init (imports/contexts/legal/Legal.js:42:30)
I20250114-16:54:32.460(1)?     at processTicksAndRejections (node:internal/process/task_queues:105:5)
I20250114-16:54:32.460(1)?     at /path/to/moo,jsjs:1:1
I20250114-16:54:32.460(1)?     at /tools/static-assets/server/boot.js:453:5
I20250114-16:54:32.460(1)?     at /tools/static-assets/server/boot.js:504:7
I20250114-16:54:32.460(1)?     at startServerProcess (/tools/static-assets/server/boot.js:501:3)
```

The errors does not occur when reverting to release 3.0.4

jankapunkt:
Update: also happens in regular Method calls

nachocodoner:
This looks like a major breaking change. I don't understand why such a basic scenario isn't covered by our tests.

Does this happen in plain Meteor examples, or only when using community packages?

jankapunkt:
Versions of non core packages did not change when reverting to 3.0.4 so I thought its a regression in core. Let me check with the Todo example

nachocodoner:
@jankapunkt: Could you provide a repository with the reproduction? So that I can have directly the way to get it. I have not been able to see on my example apps. And I starting to think is something around the packages or environment.

jankapunkt:
[This refactor](https://github.com/meteor/meteor/commit/d309462debf57dbd083a9f5709cbe65d86c5c743) introduced a change in how Collection options are normalized.

If someone passes constructed options, such as 

```js
{
[...]

PR #13739, opened 2025-05-12
Ensure normalize collection options properly respecting defaults

nachocodoner:
Context: https://github.com/meteor/meteor/issues/13545

This PR fixes a regression introduced on a refactor that happened around collections.


Grubba27 (approved):

packages/mongo/collection/collection_utils.js
```diff
@@ -88,12 +88,17 @@ export function normalizeOptions(options) {
     options.connection = options.manager;
   }
 
+  const cleanedOptions = Object.fromEntries(
+    Object.entries(options || {}).filter(([_, v]) => v !== undefined),
+  );
```
Grubba27, line 93: I think we have a helper like this one somewhere(maybe underscore)

nachocodoner: But wasn’t the goal to drop Underscore to slim down dependencies and reduce bundle size?

Maybe pure helper implementations are better than relying on underscore.

Issue #13731, opened 2025-05-05
Accounts.sendVerificationEmail throws "No such email for user" error even when email exists (Meteor 3.2)


joyjiyuan:
Hi Meteor team,

I'm encountering an issue in Meteor 3.2 when trying to send a verification email using server methods. Here's the relevant code:

```
await Accounts.addEmailAsync(userId, email);
await Accounts.sendVerificationEmail(userId, email);
```
This consistently throws the following error:

`Error: 403, reason: 'No such email for user.'`

I'm sure that the email is added successfully — it appears on the user object in the database immediately before the sendVerificationEmail call.

Is there any known cause for this error in Meteor 3.2? Could it be a bug or a timing issue with how the email is registered internally?

Thanks in advance for your help!

nachocodoner:
I’m seeing this in my apps too. I haven’t fully migrated to Meteor 3 yet though. I need to debug further to confirm if the issue comes from the core. I’d expect more reports since this feature is widely used.

Thanks for the report, I'll keep posted.

nachocodoner:
[I've created a PR](https://github.com/meteor/meteor/pull/13740) to add test coverage for the send-email behaviors, so I’ll keep this issue open. I haven’t been able to reproduce the problem yet. @joyjiyuan, do you have any reproduction repository?

Related to account emails, I found that the `Accounts.urls` hooks to describe a custom token url for verifyEmail, resetPassword, and other email types don’t support async/await definitions. It’s common to need to await something in those URL functions, so I’ll update the PR to add async support, include tests, and document this use case.

joyjiyuan:
Thanks for following up!

At this point, my production app has actually dropped this part of the functionality, so I don’t have a reproduction repo to share right now. I’ll definitely keep an eye on the PR though — the async support for `Accounts.urls` sounds like a valuable improvement. Appreciate your work on this!

PR #13740, opened 2025-05-12
Async-compatible account URLs and email-sending coverage

nachocodoner:
Context: https://github.com/meteor/meteor/issues/13731

This PR attempts to reproduce this issue reproted. However, I can't get it reproduce it. However, I took the chance to provide test coverage for send email functions and update the types seem they were wrong. 

I will give more attempts to reproduce it, but pushing the test and types refactor first.

Finally this PR covered as well async-compatible account URLs.


packages/accounts-base/accounts-base.d.ts
```diff
@@ -204,26 +204,34 @@ export namespace Accounts {
     options?: { fields?: Mongo.FieldSpecifier | undefined }
   ): Promise<Meteor.User | null | undefined>;
 
+  interface SendEmailResult {
+    email: string;
+    user: any;
+    token: string;
+    url: string;
+    options: any;
```
italojs: is expectec an object here, right?
we dont know the expected options?
instead of use any, we would to use `Record<string, any>` to ensure the user isnt passing a string, number, boolean....

packages/accounts-base/accounts-base.d.ts
```diff
@@ -204,26 +204,34 @@ export namespace Accounts {
     options?: { fields?: Mongo.FieldSpecifier | undefined }
   ): Promise<Meteor.User | null | undefined>;
 
+  interface SendEmailResult {
+    email: string;
+    user: any;
```
italojs: we dont know the expected user?
instead of use any, we would to use Record<string, any> to ensure the user isnt passing a string, number, boolean....

italojs (changes requested):

packages/accounts-base/accounts_server.js
```diff
@@ -84,6 +84,11 @@ export class AccountsServer extends AccountsCommon {
 
     this._skipCaseInsensitiveChecksForTest = {};
 
+    // Helper function to resolve promises if needed
+    this._resolvePromise = async (value) => {
+      return value instanceof Promise ? await value : value;
```
Grubba27: Maybe we should use `Meteor._isPromise`

PR #13743, opened 2025-05-13
Refactor MeteorConfig to spread it to the tool and build plugins

nachocodoner:
This PR addresses this comment to simplify and unify the Meteor config modern normalization and availability via the global context.

To introduce in Meteor 3.3.1


italojs (approved):

italojs:
lgtm, only missing the tests

nachocodoner:
There are two issues in this PR that I added as task points to take a look later.

- Fix issue with a specific test not reacting with modern watcher changes
- Research why tests seems to pass slower with this new refactor happens constantly, [image](https://github.com/user-attachments/assets/a61f89fb-c196-4720-bde7-439bd11a5902)

packages/babel-compiler/babel-compiler.js
```diff
@@ -99,47 +99,12 @@ function compileWithSwc(source, swcOptions = {}, { inputFilePath, filename, sour
...
-    lastModifiedMeteorConfig = getMeteorAppPackageJson()?.meteor;
-    lastModifiedMeteorConfig = lastModifiedMeteorConfig != null ? {
-      ...lastModifiedMeteorConfig,
-      modern: normalizeModern(modernForced || lastModifiedMeteorConfig?.modern),
-    } : {};
-
-    if (lastModifiedMeteorConfig?.modern?.transpiler?.verbose) {
-      logConfigBlock('Meteor Config', lastModifiedMeteorConfig);
-    }
+  if (global.meteorConfig?.modern?.transpiler?.verbose) {
```
zodern: Usually things like this are available on the file objects instead of being globals. For example, `file.hmrAvailable()`, or `file.supportsLazyCompilation`. There's also already the `Plugin` variable available inside the build plugins.

tools/cli/commands.js
```diff
@@ -261,49 +261,23 @@ export function parseRunTargets(targets) {
...
-
-let modernForced = JSON.parse(process.env.METEOR_MODERN || "false");
-let meteorConfig;
-
-function getMeteorConfig(appDir) {
-  if (meteorConfig) return meteorConfig;
-  const packageJsonPath = files.pathJoin(appDir, 'package.json');
-  if (!files.exists(packageJsonPath)) {
-    return false;
+export function getMeteorConfig(appDir) {
```
zodern: A more logical place for this might be the MeteorConfig class that is part of the project context: https://github.com/meteor/meteor/blob/2892718148a4453f31223b96e09aa7814ad80fae/tools/project-context.js#L1770

The context can be used in isobuild or when loading the build plugins to provide the modern bundler config to the build plugins.

packages/babel-compiler/babel-compiler.js
nachocodoner: I know local variables are better, but I’d have to review more code sections. That’s why I used globals more, since I saw them elsewhere to spread context. I’m open to looking at other parts too.

We’ve been pushing for faster delivery, and I still feel we move slow. If we plan a refactor, we can slow the pace and spend more time understanding other areas.

tools/cli/commands.js
nachocodoner: I need the Meteor config loaded before project context instantiation so the watcher code can choose correctly (safe-watcher.ts). Maybe I can instantiate this config globally and earlier. Many parts are unclear and I’ve seen crashes in some areas as I change things around. I could give it a try on having more time.

packages/babel-compiler/babel-compiler.js
nachocodoner: Also, Globals let me access the watcher code (safe-watcher.ts) so the Meteor config is available not just at file or plugin level, but throughout the Meteor Tools code. This is needed now and likely for future areas modernization.

nachocodoner: @zodern: Do you know why in case of Profile we have this?

```
// Profile for test and production environments
let Profile;
if (typeof global.Profile !== 'undefined') {
  Profile = global.Profile;
} else if (typeof Plugin !== 'undefined' && Plugin.Profile) {
  Profile = Plugin.Profile;
} else {
  Profile = function (label, func) {
    return function () {
      return func.apply(this, arguments);
    }
  }
  Profile.time = function (label, func) {
    func();
  }
}
```

Isn't enought to assume that global will be available on test and productions envs (published Meteor versions)? Or is it needed Plugin variable for some scenarios?

I will apply similarly for meteorConfig if is needed.

zodern: > I know local variables are better, but I’d have to review more code sections. That’s why I used globals more, since I saw them elsewhere to spread context. I’m open to looking at other parts too.

One option is you could internally use globals in the Meteor tool, but provide an api for build plugins. This would allow you to later clean it up and remove the globals without breaking plugins. Since Meteor 3 there has been a trend to use globals, which is a change from the previous efforts to minimize globals.

> Do you know why in case of Profile we have this?

There's situations where the Profiler isn't available. Where is that code from? I don't remember Profile normally being a global - usually it is imported, or it is made available to specific packages.

nachocodoner: > One option is you could internally use globals in the Meteor tool, but provide an api for build plugins.

I will try to use API plugin and alternatives like files exports to reduce globals as part of refactor now we have enough test coverage. I wasn’t aware of that, I only saw them used throughout the codebase, and seemed convenient for prototyping and faster delivery.

> There's situations where the Profiler isn't available. Where is that code from? I don't remember Profile normally being a global - usually it is imported, or it is made available to specific packages.


The example lives in the minify-js.js file, but I just noticed [it comes from the recent code enabling SWC](https://github.com/meteor/meteor/blob/release-3.3/packages/standard-minifier-js/plugin/minify-js.js#L27-L42) in standard-minifier-js. Since it was cherry-picked from minify-js-sourcemaps, [where used Plugin API pattern there](https://github.com/zodern/minify-js-sourcemaps/blob/master/standard-minifier-js-sourcemaps/plugin/minify-js.js#L8-L21), and not the global.

zodern: I don't think Profile it is ever a global, but it is added to the package scope in some situations.

PR #13758, opened 2025-05-17
Node v24

StorytellerCZ:
Updated dev bundle to Node v24.0.2 and NPM v11.3.0.
Updated all GitHub actions and Travis to use Node 24.

Also used this opportunity to sync `typescript` in the root `package.json` with the version we have in our `typescript` package and do a minor update on all the other dependencies. I will spin this into its own PR as well.


Torgen:
Am I interpreting https://github.com/meteor/meteor/blob/devel/npm-packages/babel-preset-meteor/modern.js#L22 correctly in that babel is still targeting Node 8 compatibility on the server side with its generated code? Should that be bumped with the node version, or do you want the compiled app to be able to run on Node 8 even if the tool ships with 20+?

StorytellerCZ:
@Torgen that needs to be bumped.

Update: On closer look it looks to me that that part sets the minimum versions, so no change needed there.

StorytellerCZ:
Following in #13811

PR #13761, opened 2025-05-19
Support swc.config.js file for dynamic config

nachocodoner:
Context: https://forums.meteor.com/t/weekly-update-april-16-2025-faster-bundle-times-with-3-3-beta-0/63494/35

This PR adds support for dynamic `swc.config.js`, so Meteor projects can define configs that change based on environment variables or other factors. It also ensures that the cache key includes the generated config, so restarting the app with new variables recompiles with the new config.

I verified that METEOR_LOCAL_DIR is respected and the SWC cache still uses the configured directory.

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -5,6 +5,8 @@ const reifyCompile = Npm.require("@meteorjs/reify/lib/compiler").compile;
 const reifyAcornParse = Npm.require("@meteorjs/reify/lib/parsers/acorn").parse;
 var fs = Npm.require('fs');
 var path = Npm.require('path');
+var vm = Npm.require('vm');
+var crypto = Npm.require('crypto');
 
 /**
  * A compiler that can be instantiated with features and used inside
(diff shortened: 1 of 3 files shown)


packages/babel-compiler/babel-compiler.js
```diff
@@ -145,15 +147,37 @@ BCp.initializeMeteorAppConfig = function () {
 let lastModifiedSwcConfig;
 let lastModifiedSwcConfigTime;
 BCp.initializeMeteorAppSwcrc = function () {
-  if (!lastModifiedSwcConfig && !fs.existsSync(`${getMeteorAppDir()}/.swcrc`)) {
+  const hasSwcRc = fs.existsSync(`${getMeteorAppDir()}/.swcrc`);
```
zodern, line 150: For babel and typescript, Meteor uses `findControlFile` to get the config file for each file. This allows using different config files in different directories. https://github.com/meteor/meteor/blob/7bb6273eb59cc1461571df0a9c5105b14e53afaf/packages/babel-compiler/babel-compiler.js#L548

This also is important for packages, since each package normally has control over how it is compiled (both what build plugins to use, and the babel config used for the package). It probably is important for packages to be able to use their own swc config if necessary.

To give one example, if there was a swc plugin for zodern:relay, it should only run for the app or packages that enable the plugin. This is how it works with babel. Otherwise, it could cause problems by compiling files in packages that it shouldn't.

nachocodoner: I will cover your comments in another PR. 🙏

PR #13764, opened 2025-05-21
DEV: update Meteor version to 3.2.2 in config and package.json

italojs:
Once approved, will do a npm publish


Grubba27 (approved):

SPodjasek:
If I may, I would like to put my two cents here... Could you also add a note that you can install any specific Meteor version by using this environment variable:

```console
$ INSTALL_METEOR_VERSION=3.3-rc.0 npx meteor <install>
```

Issue #13552, opened 2025-01-16
Minimongo does not support "." in the key names, v3.04


aantich:
The issue is simple - Mongo started supporting "." in key names in the last versions, which is absolutely critical when dealing with things such as OpenAPI specs etc. Minimongo does not, and when I subscribe to a mongo collection with "." in key names I get the following errors:
```
Uncaught (in promise) MinimongoError: Key api_key.created must not contain '.'
    at MinimongoError (common.js:1085:17)
    at assertIsValidFieldName (local_collection.js:2316:11)
    at Object.<anonymous> (local_collection.js:2307:7)
    at JSON.stringify (<anonymous>)
    at assertHasValidFieldNames (local_collection.js:2306:10)
    at LocalCollection.prepareInsert (local_collection.js:108:5)
    at LocalCollection.insert (local_collection.js:132:21)
    at Object.update (collection.js:262:28)
    at store.<computed> [as update] (livedata_connection.js:305:38)
    at Connection._performWritesClient (livedata_connection.js:1418:19)
MinimongoError @ common.js:1085
assertIsValidFieldName @ local_collection.js:2316
(anonymous) @ local_collection.js:2307
assertHasValidFieldNames @ local_collection.js:2306
prepareInsert @ local_collection.js:108
insert @ local_collection.js:132
update @ collection.js:262
store.<computed> @ livedata_connection.js:305
_performWritesClient @ livedata_connection.js:1418
_flushBufferedWritesClient @ livedata_connection.js:1343
_flushBufferedWrites @ livedata_connection.js:1348
_livedata_data @ livedata_connection.js:1291Understand this errorAI
meteor.js?hash=b29dfed1a1d0b8b41fe4e360e08157432d459f34:1262 Exception in flushing DDP buffered writes: MinimongoError: Key node.js must not contain '.'
    at MinimongoError (common.js:1085:17)
    at assertIsValidFieldName (local_collection.js:2316:11)
    at Object.<anonymous> (local_collection.js:2307:7)
    at JSON.stringify (<anonymous>)
    at assertHasValidFieldNames (local_collection.js:2306:10)
    at LocalCollection.prepareInsert (local_collection.js:108:5)
    at LocalCollection.insert (local_collection.js:132:21)
    at Object.update (collection.js:262:28)
    at store.<computed> [as update] (livedata_connection.js:305:38)
    at Connection._performWritesClient (livedata_connection.js:1418:19)

[...]

PR #13771, opened 2025-05-23
FIX: mini mongo dot notation

italojs:
# Fix for Minimongo dot notation in field names

issue: https://github.com/meteor/meteor/issues/13552

### DIFF
--- a/packages/minimongo/local_collection.js
+++ b/packages/minimongo/local_collection.js
@@ -1124,13 +1124,23 @@ LocalCollection._checkSupportedProjection = fields => {
   }
 
   Object.keys(fields).forEach(keyPath => {
-    if (keyPath.split('.').includes('$')) {
+    if (keyPath.split('.').includes('$') && keyPath !== '$getField') {
       throw MinimongoError(
         'Minimongo doesn\'t support $ operator in projections yet.'
       );
     }
 
     const value = fields[keyPath];
+
+    // Support for $getField operator
+    if (typeof value === 'object' && hasOwn.call(value, '$getField')) {
+      if (typeof value.$getField !== 'string') {
+        throw MinimongoError(
+          '$getField requires a string field name'
+        );
+      }
+      return; // $getField is supported, skip other validations
+    }
 
     if (typeof value === 'object' &&
         ['$elemMatch', '$meta', '$slice'].some(key =>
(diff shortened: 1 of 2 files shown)


italojs:
still testing a few scenarios

radekmie:
The approach of escaping a dot with something else has two major flaws:
1. **It affects the actual data stored in MongoDB.** As in, if you'd `insert` with Meteor's escaping logic but `find` elsewhere (e.g., with `rawCollection` or directly in the database), you'd get different results.
2. **It comes at a high performance price.** Renaming _every_ key of _every_ fetched document at _every_ depth is just way too expensive. It'd need to be measured of course, but it'd impact every single app while benefiting only a handful of use cases.

But all that is not really relevant since... This "escaping" does not escape anything:
```ts
const x = 'a.b';
x === x.replace(/\./g, '\u002e'); // true
```

---


The only correct approach would be to follow [what MongoDB's doing here](https://www.mongodb.com/docs/manual/core/dot-dollar-considerations/periods/), i.e., implement `$getField` (easy) and `$setField` (hard, as it'd require implementing [updates with aggregate pipelines](https://www.mongodb.com/docs/manual/tutorial/update-documents-with-aggregation-pipeline/#std-label-updates-agg-pipeline)).

italojs:
@radekmie addressed your comment

packages/minimongo/local_collection.js
```diff
@@ -1124,14 +1124,24 @@ LocalCollection._checkSupportedProjection = fields => {
   }
 
   Object.keys(fields).forEach(keyPath => {
-    if (keyPath.split('.').includes('$')) {
+    if (keyPath.split('.').includes('$') && keyPath !== '$getField') {
```
radekmie, line 1127: `$getField` is not a projection operator, it's a query operator. Supporting dots in projections is not documented in MongoDB docs, and I don't really know how it works.

packages/minimongo/local_collection.js
```diff
@@ -2275,6 +2307,24 @@ const MODIFIERS = {
     // native javascript numbers (doubles) so far, so we can't support $bit
     throw MinimongoError('$bit is not supported', {field});
   },
+  $setField(target, field, arg) {
```
radekmie, line 2310: `$setField` is not an update operator either, it's an aggregation operator. And MiniMongo doesn't support update operators, that's why I said it'll be hard.

I _think_ it would make sense to support it in Meteor like that, but I'd say it'd need to be **very** explicitly marked as non-compliant / Meteor-specific (ideally with a different name even).

radekmie (changes requested):

italojs:
we will land it in 3.4.1

PR #13775, opened 2025-05-26
Managing inactive issues with comments and labels

italojs:


zodern:
When Meteor used stalebot in the past, it was configured to ignore issues with the `confirmed` label. Does this new workflow have something similar?

italojs:
> When Meteor used stalebot in the past, it was configured to ignore issues with the `confirmed` label. Does this new workflow have something similar?

We can discuss it, but in my view, a confirmed issue can still remain idle after 90 days of inactivity. I would like to use this filter to pass idle (+confirmed maybe) issues to new contributors, since new contributors take time to develop new things; idle issues would be beneficial for them.

nachocodoner (approved):

PR #13776, opened 2025-05-26
SWC per-file context config

nachocodoner:
Context: https://github.com/meteor/meteor/pull/13761#discussion_r2096362853

This PR adds missing file-level config support to SWC, matching Babel’s plugin behavior. I also added tests to verify it works as expected.


packages/babel-compiler/babel-compiler.js
```diff
@@ -137,7 +105,7 @@ BCp.initializeMeteorAppConfig = function () {
       modern: normalizeModern(modernForced || lastModifiedMeteorConfig?.modern),
     } : {};
 
-    if (lastModifiedMeteorConfig?.modern?.transpiler?.verbose) {
+    if (lastModifiedMeteorConfig?.modern?.transpiler?.verbose || this.extraFeatures?.verbose) {
```
Grubba27: is this `this` refering to the correct scope? (we are inside of a function etc)

nachocodoner: It refers to the correct scope, otherwise coveraged test added would fail.

Grubba27 (approved):

packages/babel-compiler/babel-compiler.js
```diff
@@ -194,7 +157,7 @@ BCp.initializeMeteorAppSwcrc = function () {
 let lastModifiedSwcLegacyConfig;
 BCp.initializeMeteorAppLegacyConfig = function () {
   const swcLegacyConfig = convertBabelTargetsForSwc(Babel.getMinimumModernBrowserVersions());
-  if (lastModifiedMeteorConfig?.modern?.transpiler?.verbose && !lastModifiedSwcLegacyConfig) {
+  if ((lastModifiedMeteorConfig?.modern?.transpiler?.verbose || this.extraFeatures?.verbose) && !lastModifiedSwcLegacyConfig) {
```
italojs: If I'm right, I saw this condition `(lastModifiedMeteorConfig?.modern?.transpiler?.verbose || this.extraFeatures?.verbose)`  three times. Could you create a variable for it, please?

packages/babel-compiler/babel-compiler.js
```diff
@@ -414,8 +434,10 @@ BCp.processOneFileForTarget = function (inputFile, source) {
         } catch (e) {
           this._swcIncompatible[cacheKey] = true;
           // If SWC fails, fall back to Babel
+
+          babelOptions = setupBabelOptions();
           compilation = compileWithBabel(source, babelOptions, cacheOptions);
-          if (config?.verbose) {
+          if (config?.verbose || this.extraFeatures?.verbose) {
```
italojs: same for this condition

italojs (changes requested):

packages/babel-compiler/babel-compiler.js
nachocodoner: Done

italojs (approved):

PR #13786, opened 2025-06-10
Mitigate MongoDB cursor leak: synchronize next() and close() operations

### LINKED ISSUE #13783: OPLog tailing in Meteor 3.1 and 3.2 results in infinite number of cursors in Mongo database server
- Version with problem: every version from Meteor 3.1.1 onwards (introduction of Mongo 6 Node driver)
- Latest version without problem: Meteor 3.0.4
- OS: Galaxy provided OS, for local reproduction: MacOS 13.4 Ventura; database server: Mongo 6.0.5, Mongo 6.0.22
- Expected behavior: 
Every time OPLog tailing cursor is timed out after 30 seconds, it should not only be "killed" in the Mongo driver codebase, but the driver should also send the command to cleanup that cursor on the Mongo server side

This can be verified by the following check:
`db.serverStatus().metrics.cursor.open` command executed on Mongo server should not result in growing `total` number of opened cursors every 30 seconds

- Actual behavior: 
Whenever OPLog tailing cursor times out after 30 seconds (see: https://github.com/meteor/meteor/blob/release/METEOR%403.1.1/packages/mongo/asynchronous_cursor.js#L87) it attempts to `close` given cursor synchronously (see: https://github.com/meteor/meteor/blob/release/METEOR%403.1.1/packages/mongo/asynchronous_cursor.js#L90 and https://github.com/meteor/meteor/blob/release/METEOR%403.1.1/packages/mongo/asynchronous_cursor.js#L127) which already is problematic, because the Mongo 6.9 driver's `close` cursor method is asynchronous (see: https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/cursor/abstract_cursor.ts#L479)

However, this alone is not enough for the problem to be consistently not removing cursor, this is only a race condition that could happen every now and then. The real problem is that the `nextObjectPromise` that races against the timeout (again, see: https://github.com/meteor/meteor/blob/release/METEOR%403.1.1/packages/mongo/asynchronous_cursor.js#L87) calls underneath cursor's `next` method (see: https://github.com/meteor/meteor/blob/release/METEOR%403.1.1/packages/mongo/asynchronous_cursor.js#L40) which happens to still execute when the `close` cursor method starts executing.

The result is that the cursor reaches the following check: https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/cursor/abstract_cursor.ts#L781 which happens to be true, because at the very start of the cursor `close` this happens: https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/cursor/abstract_cursor.ts#L795 followed by https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/cursor/abstract_cursor.ts#L806

This means that the `cleanup` method is executed for the second time due to `next` method behavior and this time the whole try/catch section of that method is skipped (because `isKilled` is `true`) so it fasttracks to the `finally` section which ends the session (see: https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/cursor/abstract_cursor.ts#L820). And then it goes back to the `cleanup` triggered by Meteor's attempt to close the cursor, and before the driver sends the command to "kill" the cursor on server side, it runs into the `MongoExpiredSessionError: Cannot use a session that has ended` exception (see: https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/sessions.ts#L995, https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/operations/kill_cursors.ts#L32 and https://github.com/mongodb/node-mongodb-native/blob/v6.9.0/src/cursor/abstract_cursor.ts#L809)

This can be verified by the following check:
`db.serverStatus().metrics.cursor.open` command executed on Mongo server shows that every 30 seconds there's a new open, non-pinned cursor created

- reproduction:
Because this happens when you create a default app via `meteor create` command I am not providing the repository with a reproduction. If this is not sufficient and you still need me to push the content of `meteor create` output to a public repository, please let me know, I will do it.

Steps:
1. Create meteor app using `meteor create meteor-oplog-issue --typescript`
2. Set `MONGO_URL` and `MONGO_OPLOG_URL` to your Mongo instance
3. Start app via `npm run start`
4. After a minute or so, start executing `db.serverStatus().metrics.cursor.open` command on your Mongo server every 30 seconds and check the `total` value in the returned object growing while the `pinned`

italojs:
Note: branch base is devel because release 3.x must to be updated with devel

## Description:
This PR addresses the MongoDB cursor leak issue  as reported in  issue #13783 

## Key changes:

- Adds concurrency control to `AsynchronousCursor` to prevent simultaneous `next()` and `close()` calls, avoiding race conditions.
- Ensures that when closing a cursor, any pending `next` operation is awaited before the cursor is actually closed.
- Adjusts the cursor lifecycle in the tail method to guarantee the previous cursor is fully closed before creating a new one.

These changes help ensure that the killCursor command is sent to the MongoDB server before the session ends, preventing the accumulation of open cursors and improving the reliability of OPLog tailing.

### DIFF
--- a/packages/minimongo/common.js
+++ b/packages/minimongo/common.js
@@ -2,12 +2,7 @@ import LocalCollection from './local_collection.js';
 
 export const hasOwn = Object.prototype.hasOwnProperty;
 
-export class MiniMongoQueryError extends Error {
-  constructor(message) {
-    super(message);
-  }
-}
-
+export class MiniMongoQueryError extends Error {}
 // Each element selector contains:
 //  - compileElementSelector, a function with args:
 //    - operand - the "right hand side" of the operator
--- a/packages/minimongo/common.js
+++ b/packages/minimongo/common.js
@@ -2,6 +2,13 @@ import LocalCollection from './local_collection.js';
 
 export const hasOwn = Object.prototype.hasOwnProperty;
 
+export class QueryError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'QueryError';
+  }
+}
+
 // Each element selector contains:
 //  - compileElementSelector, a function with args:
 //    - operand - the "right hand side" of the operator
@@ -24,7 +31,7 @@ export const ELEMENT_OPERATORS = {
       if (!(Array.isArray(operand) && operand.length === 2
             && typeof operand[0] === 'number'
             && typeof operand[1] === 'number')) {
-        throw Error('argument to $mod must be an array of two numbers');
+        throw new QueryError('argument to $mod must be an array of two numbers');
       }
 
       // XXX could require to be ints or round or something
@@ -38,7 +45,7 @@ export const ELEMENT_OPERATORS = {
   $in: {
     compileElementSelector(operand) {
       if (!Array.isArray(operand)) {
-        throw Error('$in needs an array');
+        throw new QueryError('$in needs an array');
       }
 
       const elementMatchers = operand.map(option => {
@@ -47,7 +54,7 @@ export const ELEMENT_OPERATORS = {
         }
 
         if (isOperatorObject(option)) {
-          throw Error('cannot nest $ under $in');
+          throw new QueryError('cannot nest $ under $in');
         }
 
         return equalityElementMatcher(option);
@@ -74,7 +81,7 @@ export const ELEMENT_OPERATORS = {
         // does.
         operand = 0;
       } else if (typeof operand !== 'number') {
-        throw Error('$size needs a number');
+        throw new QueryError('$size needs a number');
       }
 
       return value => Array.isArray(value) && value.length === operand;
@@ -112,16 +119,16 @@ export const ELEMENT_OPERATORS = {
           'maxKey': 127,
         };
         if (!hasOwn.call(operandAliasMap, operand)) {
-          throw Error(`unknown string alias for $type: ${operand}`);
+          throw new QueryError(`unknown string alias for $type: ${operand}`);
         }
         operand = operandAliasMap[operand];
       } else if (typeof operand === 'number') {
         if (operand === 0 || operand < -1
           || (operand > 19 && operand !== 127)) {
-          throw Error(`Invalid numerical $type code: ${operand}`);
+          throw new QueryError(`Invalid numerical $type code: ${operand}`);
         }
       } else {
-        throw Error('argument to $type is not a number or a string');
+        throw new QueryError('argument to $type is not a number or a string');
       }
 
       return value => (
@@ -168,7 +175,7 @@ export const ELEMENT_OPERATORS = {
   $regex: {
     compileElementSelector(operand, valueSelector) {
       if (!(typeof operand === 'string' || operand instanceof RegExp)) {
-        throw Error('$regex has to be a string or RegExp');
+        throw new QueryError('$regex has to be a string or RegExp');
       }
 
       let regexp;
@@ -180,7 +187,7 @@ export const ELEMENT_OPERATORS = {
         // ones (eg, Mongo supports x and s). Ideally we would implement x and s
         // by transforming the regexp, but not today...
         if (/[^gim]/.test(valueSelector.$options)) {
-          throw new Error('Only the i, m, and g regexp options are supported');
+          throw new QueryError('Only the i, m, and g regexp options are supported');
         }
 
         const source = operand instanceof RegExp ? operand.source : operand;
@@ -198,7 +205,7 @@ export const ELEMENT_OPERATORS = {
     dontExpandLeafArrays: true,
     compileElementSelector(operand, valueSelector, matcher) {
       if (!LocalCollection._isPlainObject(operand)) {
-        throw Error('$elemMatch need an object');
+        throw new QueryError('$elemMatch need an object');
       }
 
       const isDocMatcher = !isOperatorObject(
@@ -353,7 +360,7 @@ const VALUE_OPERATORS = {
   // $options just provides options for $regex; its logic is inside $regex
   $options(operand, valueSelector) {
     if (!hasOwn.call(valueSelector, '$regex')) {
-      throw Error('$options needs a $regex');
+      throw new QueryError('$options needs a $regex');
     }
 
     return everythingMatcher;
@@ -361,14 +368,14 @@ const VALUE_OPERATORS = {
   // $maxDistance is basically an argument to $near
   $maxDistance(operand, valueSelector) {
     if (!valueSelector.$near) {
-      throw Error('$maxDistance needs a $near');
+      throw new QueryError('$maxDistance needs a $near');
     }
 
     return everythingMatcher;
   },
   $all(operand, valueSelector, matcher) {
     if (!Array.isArray(operand)) {
-      throw Error('$all requires array');
+      throw new QueryError('$all requires array');
     }
 
     // Not sure why, but this seems to be what MongoDB does.
@@ -379,7 +386,7 @@ const VALUE_OPERATORS = {
     const branchedMatchers = operand.map(criterion => {
       // XXX handle $all/$elemMatch combination
       if (isOperatorObject(criterion)) {
-        throw Error('no $ expressions in $all');
+        throw new QueryError('no $ expressions in $all');
       }
 
       // This is always a regexp or equality selector.
@@ -392,7 +399,7 @@ const VALUE_OPERATORS = {
   },
   $near(operand, valueSelector, matcher, isRoot) {
     if (!isRoot) {
-      throw Error('$near can\'t be inside another $ operator');
+      throw new QueryError('$near can\'t be inside another $ operator');
     }
 
     matcher._hasGeoQuery = true;
@@ -433,7 +440,7 @@ const VALUE_OPERATORS = {
       maxDistance = valueSelector.$maxDistance;
 
       if (!isIndexable(operand)) {
-        throw Error('$near argument must be coordinate pair or GeoJSON');
+        throw new QueryError('$near argument must be coordinate pair or GeoJSON');
       }
 
       point = pointToArray(operand);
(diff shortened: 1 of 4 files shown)


StorytellerCZ:
Given the severity I think this should be released standalone package patch.

nachocodoner:
I’d like to run a [meteor/performance](https://github.com/meteor/performance) runtime check to see if the MongoDB driver upgrade starting in version 3.1.1 caused the regression.

packages/mongo/mongo_connection.js
```diff
@@ -804,6 +804,8 @@ Object.assign(MongoConnection.prototype, {
     cursorDescription, ordered, callbacks, nonMutatingCallbacks) {
     var self = this;
     const collectionName = cursorDescription.collectionName;
+    
+    console.error('🔥 _observeChanges: CALLED with selector:', JSON.stringify(cursorDescription.selector));
```
nachocodoner: Can this be cleaned?

packages/mongo/asynchronous_cursor.js
```diff
@@ -73,7 +83,7 @@ export class AsynchronousCursor {
   // Returns a promise which is resolved with the next object (like with
   // _nextObjectPromise) or rejected if the cursor doesn't return within
   // timeoutMS ms.
-  _nextObjectPromiseWithTimeout(timeoutMS) {
+  async _nextObjectPromiseWithTimeout(timeoutMS) {
```
Grubba27: Why `async` if we are not awaiting anything in here?

packages/mongo/asynchronous_cursor.js
```diff
@@ -84,12 +94,11 @@ export class AsynchronousCursor {
...
-    return Promise.race([nextObjectPromise, timeoutPromise])
-      .catch((err) => {
+      return Promise.race([nextObjectPromise, timeoutPromise]).catch(async err => {
         if (err === timeoutErr) {
-          this.close();
-          return;
+          return this.close();
         }
+        // If the error is not a timeout, rethrow it.
         throw err;
```
Grubba27: same goes for this `async err`

packages/mongo/asynchronous_cursor.js
```diff
@@ -36,9 +38,17 @@ export class AsynchronousCursor {
...
     try {
-      return this._dbCursor.next();
+      this._pendingNext = this._dbCursor.next();
+      const result = await this._pendingNext;
+      this._pendingNext = null;
+      return result;
     } catch (e) {
+      this._pendingNext = null;
       console.error(e);
     }
```
radekmie: 1. It'd be more natural to reset the `_pendingNext` in a `finally` block.
2. Why is this `console.error` here? A failing cursor should prevent further iteration, not be silenced.

packages/mongo/asynchronous_cursor.js
radekmie: Actually, this entire function could look like this _(I also fixed the weird formatting)_:
```ts
  async _nextObjectPromiseWithTimeout(timeoutMS) {
    const nextObjectPromise = this._nextObjectPromise();
    if (!timeoutMS) {
      return nextObjectPromise;
    }

    const timeoutPromise = new Promise(resolve => {
      // On timeout, close the cursor.
      const timeoutId = setTimeout(() => {
        resolve(this.close());
      }, timeoutMS);

      // If the `_nextObjectPromise` returned first, cancel the timeout.
      nextObjectPromise.finally(() => {
        clearTimeout(timeoutId);
      });
    });

    return Promise.race([nextObjectPromise, timeoutPromise]);
  }
```

That's because currently, the `timeoutErr` is ignored ("swallowed" in the `catch`).

packages/mongo/asynchronous_cursor.js
```diff
@@ -123,8 +132,19 @@ export class AsynchronousCursor {
...
+    if (this._pendingNext) {
+      try {
+        await this._pendingNext;
+      } catch (e) {
+        // ignore
+      }
+    }
+    if (this._dbCursor && typeof this._dbCursor.close === 'function') {
+      await this._dbCursor.close();
+    }
```
radekmie: It wasn't here before - is it needed now? (I'm asking about both the check and the `await`.)

packages/mongo/mongo_connection.js
```diff
@@ -882,10 +882,11 @@ Object.assign(MongoConnection.prototype, {
...
           try {
             matcher = new Minimongo.Matcher(cursorDescription.selector);
-            return true;
+            return !!matcher;
           } catch (e) {
-            // XXX make all compilation errors MinimongoError or something
-            //     so that this doesn't ignore unrelated exceptions
+            if (e.message && (e.message.includes('needs an array') )) {
+              throw e;
+            }
```
radekmie: This seems like an unrelated change that I think should be followed up in a separate PR. Also, there's no need to `!!matcher`, as its `constructor` never returns anything, so `new`ing it will always return an instance (i.e., an object).

packages/mongo/mongo_connection.js
```diff
@@ -792,9 +792,9 @@ MongoConnection.prototype.tail = function (cursorDescription, docCallback, timeo
   });
 
   return {
-    stop: function () {
+    stop: async function () {
       stopped = true;
-      cursor.close();
+      await cursor.close();
```
radekmie: This sounds like a potentially breaking change. Also, it becomes `void | Promise<void>`, since the `ObserveHandle.prototype.stop` does not return a `Promise`.

packages/mongo/asynchronous_cursor.js
italojs: The cursor.next() method is used to get the next document from a cursor. Here, I wait for the cursor to finish its current operation before closing it.

Regarding the await: the function [close](https://github.com/mongodb/node-mongodb-native/blob/56b2e6d56d55520f7bc18918dcf707bd12075d7e/src/cursor/abstract_cursor.ts#L684) returns a Promise<void>. I use await to ensure the operation completes.

packages/mongo/asynchronous_cursor.js
italojs: 1. done
2. i just didn't touch on this implementation since the goal here was to fix the issue, i wasn't looking for improvements, but i'll take a look at it

packages/mongo/asynchronous_cursor.js
italojs: I did a few tests, it doesnt affect the final result(close the cursors), so i removed the await

packages/mongo/mongo_connection.js
italojs: I did a few tests, it doesnt affect the final result(close the cursors), so i removed the async await

packages/mongo/asynchronous_cursor.js
italojs: addressed

packages/mongo/asynchronous_cursor.js
italojs: addressed

packages/mongo/mongo_connection.js
italojs: Yeah, it looks like an unrelated change, but after the changes, the test  `observeChanges - bad query` was crashing because the `$in needs an array` error wasn't thrown
I refactored this part by adding a `QueryError` class, so we can throw all bad query syntax errors.

packages/mongo/asynchronous_cursor.js
italojs: 2 - The behavior to prevent further iteration is already handled when we return null from _rawNextObjectPromise. When this happens, _nextObjectPromise also returns null, which causes the iterator (next()) to return { done: true, value: null }. This properly signals the end of iteration to the consumer of the cursor, so no further documents will be processed. Therefore, even if the error is not thrown, returning null ensures that iteration is stopped as expected.

italojs:
@radekmie changes addressed

packages/minimongo/common.js
```diff
@@ -379,7 +386,7 @@ const VALUE_OPERATORS = {
     const branchedMatchers = operand.map(criterion => {
       // XXX handle $all/$elemMatch combination
       if (isOperatorObject(criterion)) {
-        throw Error('no $ expressions in $all');
+        throw new Error('no $ expressions in $all');
```
Grubba27: why not QueryError here?

packages/minimongo/common.js
```diff
@@ -2,6 +2,13 @@ import LocalCollection from './local_collection.js';
 
 export const hasOwn = Object.prototype.hasOwnProperty;
 
+export class QueryError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = 'QueryError';
```
Grubba27: You could move it to the class definition instead of the construction:

```js
class QueryError extends Error {
  name = 'QueryError';
}
```

packages/mongo/mongo_connection.js
```diff
@@ -886,6 +886,9 @@ Object.assign(MongoConnection.prototype, {
           } catch (e) {
             // XXX make all compilation errors MinimongoError or something
             //     so that this doesn't ignore unrelated exceptions
+            if (e.name === 'QueryError') {
```
Grubba27: here you could call it via class definition too:

```js
 if (e.name === QueryError.name) {
```
or you could do:

```js
e.constructor.name === QueryError.name
```

packages/minimongo/common.js
italojs: good catch, just missed it

packages/mongo/mongo_connection.js
radekmie: And how about `instanceof`?

nachocodoner (approved):

packages/minimongo/common.js
```diff
@@ -2,6 +2,12 @@ import LocalCollection from './local_collection.js';
 
 export const hasOwn = Object.prototype.hasOwnProperty;
 
+export class MiniMongoQueryError extends Error {
+  constructor(message) {
+    super(message);
+  }
+}
+
```
Grubba27:
```suggestion
export class MiniMongoQueryError extends Error {}
```

PR #13787, opened 2025-06-10
MongoDB Change Streams 🌿

italojs:
This is the first implementation of Change Streams in Meteor’s data layer. While the core functionality is working, we welcome feedback on API design and DX improvements before finalizing.

# MongoDB Change Streams

This pull request introduces support for MongoDB Change Streams in Meteor's core, replacing the current oplog option.

Change streams are more efficient than the traditional oplog tailing approach for several reasons:

With oplog tailing, Meteor must continuously poll and process the entire MongoDB oplog, filtering out irrelevant changes for each collection and user. This results in unnecessary CPU and memory usage, especially as the number of collections, users, or writes increases.

Change streams are natively supported by MongoDB and are designed for scalability, providing a more direct and efficient notification mechanism for data changes.

We will maintain the old Oplog and Pooling implementation, so if you aren't able to use change stream or just don't want to do it, you will flag it, and it SHOULD be transparent for you; any change else will be required.

## How to use it

```
{
    "packages": {
        "mongo": {
            "reactivity": ["changeStreams", "pooling", "oplog"] // Will try using CS first, then pooling, and if that does not work, oplog
        }
    }
}
```

Note: we should move the commit [48301e6](https://github.com/meteor/meteor/pull/13787/commits/48301e60565394d34a04cb64e88e0008f4a4378d) to devel when release 3.4.1 and update this branch with the changes


radekmie:
I was surprised when I saw this PR - maybe someone managed to solve the discrepancy between Oplog and Change Streams! But I'm afraid it's not the case. A few notes:
* The second argument to the `changed` callback contains the new field values. But what's important, if a complex type is changed (e.g., an object), the entire object is passed there. The `updatedFields` from the Change Stream on the other hand, will present the _paths_, e.g., `{ 'emails.1.address': 'x@y.z' }`.
    * Similarly, it should contain `undefined` for all removed fields. It's easy thanks to `removedFields`. But handling all types of array modifications is _hard_, due to `truncatedArrays`.
    * It _can_ be worked around using `fullDocumentBeforeChange` (at a small performance cost). That's what [`changestream-to-redis`](https://github.com/radekmie/changestream-to-redis) is doing.
* Consider this code: `collection.find(query).observeChanges({changeStreams:true, ...})`. The difference from the usual `observeChanges` is that the former notifies you based on the _collection_ events (e.g., an insert to the collection) and the latter based on the _result set_ (e.g., a new document now matches the query).
    * A simple example is to use a `{isActive: true}` query. The former will notify you only about inserts/updates/removes of `isActive: true` documents. The latter will _also_ inform you when `isActive` of any document will change.
* I'd recommend going with async iteration (`for await..of`) rather than callbacks - it's easier to batch the results.

StorytellerCZ:
Just a managerial node: Since this is a feature update this should be targeted towards a feature release (3.4) instead of patch release (3.3.1).

italojs:
@StorytellerCZ sure thing, I opened this PR to 3.3.1 because we will release the new version this week, and we will rebase some commits from devel to release-3.3. 
If I create a new branch from release-3.3 right now, I will probably have a few conflicts later, and the same goes for devel.


I will point this PR to release 3.4 after the release 3.3 is completed. Thanks for the heads up

italojs:
@radekmie 
- I'm sorry, I didn't receive your badge cases. Could you provide more details or share some forum/GH discussion? 
- Do you think those scenarios apply to watchChangeStream too, since it returns a direct WS connection?

italojs:
Status report: 
Following [this](https://forums.meteor.com/t/mongodb-change-streams-support-in-meteor-feedback-wanted/63681/6?u=italojs) comment, I will work on observeChanges compatibility only after the new API is ready

italojs:
@radekmie, it’s still in progress, but now I have implemented a changeLog_observe_driver, I suppose it follows your expectation about `solve the discrepancy between Oplog and Change Streams`, am I right?

The test is 100% compatible with the current oplog, and a few benchmarks I did (which I will share soon) show Meteor 15% faster. I'm still working on a scale improvement, so it's not ready for review, but I would like to get your opinion at this point.

packages/mongo/tests/changestream_observe_driver_tests.js
```diff
@@ -0,0 +1,123 @@
...
+      await TestCollection.dropCollectionAsync();
+    }
+  });
+  
+} else {
+  // Skip tests if Change Streams are not supported
+  Tinytest.add('mongo - ChangeStreamObserveDriver - skipped (not supported)', function (test) {
+    test.isTrue(true, 'Change Streams not supported in this environment');
+  });
+}
```
nachocodoner: It’s fine to add this new test for the change stream driver. I’m curious what would happen if we enabled change streams by default across the test suite. If the driver handles everything correctly, all tests should pass. We have many tests expecting publication setups, and the change stream must be included in a way that remains compatible with all current assertions.

To try it, I’d branch off this one and set the env var so all TravisCI test-in-console runs use this change stream driver. Maybe later we can add a CI job that runs change streams for the whole suite.

PD: Current TravisCI tests are failing.

packages/mongo/mongo_connection.js
```diff
@@ -850,6 +891,35 @@ Object.assign(MongoConnection.prototype, {
...
+      const canUseChangeStreams = [
+        function () {
+          // Change Streams require MongoDB 3.6+ and replica set
+          return self._supportsChangeStreams && !ordered &&
+            !callbacks._testOnlyPollCallback;
+        },
+        function () {
+          // Check if change streams are explicitly disabled
+          const mongoSettings = Meteor.settings?.packages?.mongo || {};
+          return mongoSettings.useChangeStreams !== false;
```
Grubba27: we should add this flag to the docs – I would change it to be default on for new projects, it would be nice to add it to our skeletons.

Maybe naming could be a string:


`reactivityMode` : `ChangeStream` | `Oplog`

Also, having an easy way to turn this on via env variable would be nice for testing for example, we could run the CI with both change stream and oplog to catch issues that might appear

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,262 @@
...
+      }));
+      
+      this._changeStream.on('close', Meteor.bindEnvironment(() => {
+        if (!self._stopped) {
+          // Unexpected close, attempt restart
+          setTimeout(() => {
+            if (!self._stopped) {
+              self._restartChangeStream();
+            }
+          }, 1000);
```
Grubba27: Maybe move this 1000 to a variable that could be used by the dev?

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,262 @@
...
+    if (this._stopped) return;
+    
+    try {
+      // Build the same selector and options that the cursor would use
+      const selector = this._cursorDescription.selector || {};
+      const options = { ...this._cursorDescription.options };
+      
+      // Remove some options that don't apply to find()
+      delete options.tailable;
+      delete options.oplogReplay;
```
Grubba27: Why is it necessary to delete those keys?

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,262 @@
...
+      
+      // Find all existing documents
+      const cursor = collection.find(selector, options);
+      const docs = await cursor.toArray();
+      
+      // Send 'added' for each existing document that matches our matcher
+      for (const doc of docs) {
```
Grubba27: You can run [`for await (const doc of cursor) {`](https://www.mongodb.com/docs/drivers/node/current/crud/query/cursor/#asynchronous-iteration)


When using `toArray()` large numbers of matched documents can cause performance issues or failures if the operation exceeds memory constraints

packages/mongo/mongo_connection.js
```diff
@@ -33,6 +34,7 @@ export const MongoConnection = function (url, options) {
     ignoreUndefined: true,
   }, userOptions);
 
+  console.log('mongoOptions:', mongoOptions);
```
Grubba27: I think this should be removed

packages/mongo/mongo_connection.js
```diff
@@ -113,6 +117,43 @@ MongoConnection.prototype.close = function () {
...
+  var self = this;
+  
+  try {
+    // Change Streams require MongoDB 3.6+ and replica set or sharded cluster
+    const admin = self.db.admin();
+    const serverInfo = await admin.serverInfo();
+    const version = serverInfo.version.split('.').map(Number);
+    
+    // Check MongoDB version (3.6+)
+    const hasMinVersion = version[0] > 3 || (version[0] === 3 && version[1] >= 6);
```
Grubba27: I think we have a semver package to do this kind of operation ( would it handle versions with betas in the name?)

packages/mongo/collection/collection.js
```diff
@@ -165,6 +164,87 @@ Object.assign(Mongo.Collection.prototype, {
...
+   *
+   * @example
+   *   // In a publish function:
+   *   Meteor.publish('myChanges', function() {
+   *     return MyCollection.watch(this, [
+   *       { $match: { 'operationType': 'insert' } }
+   *     ]);
+   *   });
+   */
+  watch(subscription, pipeline = [], options = {}) {
```
Grubba27: How would this snippet below work in this `watch` example?
```js
Meteor.publish('myChanges', function() {
       return MyCollection.find();
})
```

Also, passing the `this`, seems a bit weird; maybe the API could be `MyCollection.watch(pipeline, options)`. The `Publish` method could check if it is a `watch` method, and if it is, run the necessary operations

Also2, would it work like we do for [array of cursors?](https://arc.net/l/quote/ajibbhyo)

italojs: for this I'm awaitting a new PR from @rodrigooler 
So disregard it for a while

packages/mongo/mongo_connection.js
italojs: for mongo?

Twisterking:
@italojs 
Are you planning to also support the exclude & include setups for changestreams?
See https://github.com/meteor/meteor/issues/13945 and https://github.com/meteor/meteor/pull/13009!

I am not deep in the topic of changestreams, but ignoring certain updates as soon as possible in the process for collections you just are "not interested in for pub/sub" I feel also makes sense for changestreams?! We should try our best to reduce CPU (and MEM) load as much as possible.
Nice job!! 🥳

italojs:
@Twisterking for sure, this changestreams observer driver should be the most compatible with oplog as possible in the first version.
but the performance tunning will come as a second step

italojs:
@radekmie, finally, we have a first change stream version. 
Would you like to review it?
The next step will be the performance improvement, but at least all tests is passing now :)

Grubba27:
on the Settings maybe using just strings could be better?!

instead of 
```json
{
    "packages": {
        "mongo": {
            "reactivity": {
                "changeStreams": true,
                "oplog": false,
                "pooling": false
            }
        }
    }
}
```
this:
```json
{
    "packages": {
        "mongo": {
            "reactivity": "changeStreams" // only changeStreams
        }
    }
}
```

Or:

```json
{
    "packages": {
        "mongo": {
            "reactivity": ["changeStreams", "pooling", "oplog"] // Will try using CS first, then pooling, and if that does not work, oplog
        }
    }
}
```

packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
```diff
@@ -344,20 +344,23 @@ function createTestUser(test, expect) {
...
-  );
+      test.equal(Meteor.user(), null);
+      doCreate();
+    }));
+  } else {
+    doCreate();
+  }
 
-  return { username, email, password };
+  return ret;
```
radekmie: @italojs Do you think it would be possible to merge the unrelated changes first, so this PR gets smaller? I'm having a hard time reviewing it, as it's often unclear _why_ the changes are here.

packages/minimongo/local_collection.js
```diff
@@ -102,8 +102,46 @@ export default class LocalCollection {
...
     }
     options.limit = 1;
-    return (await this.find(selector, options).fetchAsync())[0];
+    const result = (await this.find(selector, options).fetchAsync())[0];
+    
+    if (result && typeof result === 'object') {
+      this._serializedBinaryData(result);
+    }
+    
+    return result;
```
radekmie: 1. Is it correct, that it's only applied for `findOneAsync`? What about the non-async version, or the more general `.find().fetch()`? I'm asking, since:
    * It's surprising to see it in this PR.
    * It's _really_ costly, especially for "wide" documents (i.e., ones with a lot of keys).
2. There's no need to check for `object`, as `fetch()` will return an array of objects anyway.

packages/minimongo/local_collection.js
```diff
@@ -102,8 +102,46 @@ export default class LocalCollection {
...
+    
+    // Recursively all object properties
+    Object.keys(obj).forEach(key => {
+      const value = obj[key];
+      
+      // Check if this looks like a serialized MongoDB.Binary object
+      if (value && typeof value === 'object' && 
+          value.sub_type !== undefined && 
+          value.buffer !== undefined && 
+          value.position !== undefined) {
```
radekmie: Would `instanceof` be applicable here?

packages/minimongo/local_collection.js
```diff
@@ -1617,9 +1655,9 @@ LocalCollection._observeFromObserveChanges = (cursor, observeCallbacks) => {
           return;
         }
 
-        // technically maybe there should be an EJSON.clone here, but it's about
-        // to be removed from this.docs!
-        const doc = transform(this.docs.get(id));
+        // Use docFromDriver if available, else fallback to cache
+        const cachedOrdered = this.docs.get(id) || { _id: id };
+        const doc = transform(cachedOrdered);
```
radekmie: This is a _really_ old piece of code. Can you explain what triggered the issue here?

packages/minimongo/local_collection.js
```diff
@@ -1719,8 +1758,7 @@ LocalCollection._removeFromResultsSync = (query, doc) => {
     query.results.splice(i, 1);
   } else {
     const id = doc._id;  // in case callback mutates doc
-
-    query.removed(doc._id);
+    query.removed(String(id));
```
radekmie: This breaks queries of documents with non-string IDs, right?

packages/minimongo/local_collection.js
```diff
@@ -1734,7 +1772,7 @@ LocalCollection._removeFromResultsAsync = async (query, doc) => {
   } else {
     const id = doc._id;  // in case callback mutates doc
 
-    await query.removed(doc._id);
+  await query.removed(String(id));
```
radekmie: (minor) It'd be great to keep the formatting consistent.

packages/mongo-id/id.js
```diff
@@ -75,6 +75,12 @@ MongoID.idStringify = (id) => {
   } else if (id === undefined) {
     return '-';
   } else if (typeof id === 'object' && id !== null) {
+    if (typeof id.toHexString === 'function') {
+      return id.toHexString();
+    }
```
radekmie: This _looks_ like `ObjectID`'s method. Why isn't it handled with the `instanceof` above?

packages/mongo-id/id.js
```diff
@@ -75,6 +75,12 @@ MongoID.idStringify = (id) => {
...
+    if (typeof id.toString === 'function') {
+      return id.toString();
+    }
```
radekmie: That's a risky breaking change, as custom classes often do have _some_ `toString`. They were rejected before (correctly), but will trigger some different behavior now (potentially incorrect, if their ID is not entirely relevant, e.g., summarizes the object).

packages/mongo/tests/observe_changes_tests.js
```diff
@@ -519,8 +519,8 @@ if (Meteor.isServer) {
       const [resolver1, promise1] = getPromiseAndResolver();
       const [resolver2, promise2] = getPromiseAndResolver();
 
-      await self.insert({x: 2, y: 3});
       self.expects.push(resolver1, resolver2);
+      await self.insert({x: 2, y: 3});
```
radekmie, line 523: Does the order matter here? I think it does not, but I'm interested _why_ this change was made.

packages/mongo/mongo_common.js
```diff
@@ -133,6 +133,18 @@ export const replaceMongoAtomWithMeteor = function (document) {
...
+  // Check for serialized MongoDB.Binary objects (plain objects with binary-like structure)
+  if (document && typeof document === 'object' && 
+      document.sub_type !== undefined && 
+      document.buffer !== undefined && 
+      document.position !== undefined) {
+    // Convert this serialized binary back to Uint8Array
+    if (document.buffer && (document.buffer instanceof Uint8Array || Array.isArray(document.buffer))) {
+      return new Uint8Array(document.buffer);
+    }
+  }
```
radekmie: This seems like a more suitable place for this conversion, but:
1. Why wasn't it necessary before?
2. Shouldn't we check for `sub_type !== 0`, just like the test above?
3. It'd be more in-align with the rest of the codebase to merge these conditions into one.

packages/mongo/mongo_connection.js
```diff
@@ -90,6 +90,8 @@ export const MongoConnection = function (url, options) {
     self._docFetcher = new DocFetcher(self);
   }
 
+  // Check if Change Streams are supported
+  self._checkChangeStreamSupport();
```
radekmie: 1. This is an asynchronous operation triggered in the constructor. It may result in a race condition, where `this._supportsChangeStreams` will be `false` not because it's not supported, but because the check is still ongoing.
2. This method executes normal MongoDB operations, meaning it'll force a connection. It wasn't the case before, i.e., it was _lazily_ opened, when needed.

Ideally, this would be done in `_observeChanges` and memoized for the next executions.

packages/mongo/mongo_connection.js
radekmie: I agree - manually parsing the version may be risky.

packages/mongo/mongo_connection.js
```diff
@@ -114,6 +116,37 @@ MongoConnection.prototype.close = function () {
...
+    // Check MongoDB version (3.6+)
+    const hasMinVersion = version[0] > 3 || (version[0] === 3 && version[1] >= 6);
+    
+    if (!hasMinVersion) {
+      self._supportsChangeStreams = false;
+      return;
+    }
+    
+    // Check if we're running on a replica set or sharded cluster
+    const isMaster = await admin.command({ isMaster: 1 });
```
radekmie: Most applications are, so it'd be ideal to start this query in parallel to `serverInfo` and `await` it here, so the latency is lower.

packages/mongo/mongo_connection.js
```diff
@@ -114,6 +116,37 @@ MongoConnection.prototype.close = function () {
...
+    const isReplicaSet = isMaster.setName || isMaster.ismaster || isMaster.secondary;
```
radekmie: `setName` is a `string | undefined`, not `boolean` - it'd be better to keep this as `boolean` though, not to leak such information.

packages/mongo/mongo_connection.js
```diff
@@ -114,6 +116,37 @@ MongoConnection.prototype.close = function () {
...
+    const isSharded = isMaster.msg === 'isdbgrid';
+    
+    self._supportsChangeStreams = isReplicaSet || isSharded;
+    
+  } catch (error) {
+    console.warn('Error checking Change Streams support:', error.message);
```
radekmie: I think it's more common to use `Meteor._debug` for it.

packages/mongo/mongo_connection.js
```diff
@@ -852,6 +885,36 @@ Object.assign(MongoConnection.prototype, {
...
 
       var matcher, sorter;
+      
+      // Check if Change Streams are available and enabled
+      const canUseChangeStreams = [
+        function () {
+          // Check if change streams are explicitly disabled
+          const mongoSettings = Meteor.settings?.packages?.mongo || {};
+          // return mongoSettings.reactivity === 'CHANGE_STREAMS' || process.env.METEOR_REACTIVITY === 'CHANGE_STREAMS';
+          return true // dumb-forcing changeStream tests into CI
```
radekmie: Don't forget about this here.

packages/mongo/mongo_connection.js
```diff
@@ -852,6 +885,36 @@ Object.assign(MongoConnection.prototype, {
...
+          }
+        },
+        function () {
+          // Change streams work with most selectors, but some complex ones might not work well
+          // For now, we use the same check as oplog
+          return OplogObserveDriver.cursorSupported(cursorDescription, matcher);
+        }
+      ].every(f => f());
+      
       var canUseOplog = [
```
radekmie: This should not be evaluated if change streams will be used.

packages/mongo/mongo_connection.js
```diff
@@ -914,7 +977,22 @@ Object.assign(MongoConnection.prototype, {
...
+      // Choose driver in order of preference: ChangeStreams > Oplog > Polling
+      var driverClass;
+      if (canUseChangeStreams) {
+        // Use dynamic import to avoid circular dependency issues
+        try {
+          driverClass = ChangeStreamObserveDriver;
+        } catch (error) {
+          console.warn('Failed to load ChangeStreamObserveDriver, falling back to oplog/polling:', error.message);
+          driverClass = canUseOplog ? OplogObserveDriver : PollingObserveDriver;
+        }
```
radekmie: Is the comment outdated? There's no dynamic import here.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+    
+    // Use the matcher passed from mongo_connection.js
+    this._matcher = options.matcher;
+    
+    // Fallback: create matcher if not provided
+    if (!this._matcher) {
+      // Import Minimongo locally to avoid circular dependencies
+      const { Minimongo } = require('meteor/minimongo');
+      this._matcher = new Minimongo.Matcher(this._cursorDescription.selector);
+    }
```
radekmie: When is this fallback needed? Other drivers are not doing it.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+  _handleReplace(id, newDoc, oldDoc) {
+    // Handle replace similar to update
+    this._handleUpdate(id, newDoc, oldDoc);
+  }
+
+  _handleDelete(id) {
+    const docs = this._multiplexer?._cache?.docs;
+    const hasDoc = typeof docs?.has === 'function'
+      ? docs.has(id)
+      : typeof docs?.get === 'function' && docs.get(id) !== undefined;
```
radekmie: Typing here would help, but we have too many `?.`s here I think (e.g., the first one), and the duck typing of the methods seems unexpected - why is `has` not always present?

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+    const cachedDoc = this._multiplexer?._cache?.docs?.get?.(id);
+    const matchesBefore = oldDoc
+      ? (this._matcher ? this._matcher.documentMatches(oldDoc).result : true)
+      : !!cachedDoc;
+
+    if (matchesAfter) {
+      if (!matchesBefore) {
+        // Document wasn't previously in the result set and now matches – emit added
+        const projectedDoc = this._projectionFn ? this._projectionFn(newDoc) : newDoc;
+        this._sendMultiplexerAdded(id, projectedDoc);
```
radekmie, line 469: This operation is marked as `async`. I don't think it has to, but dangling `Promise`s are not a good thing.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+      const projectedDoc = this._projectionFn ? this._projectionFn(doc) : doc;
+      this._sendMultiplexerAdded(id, projectedDoc);
+    }
+  }
+
+  _handleUpdate(id, newDoc, oldDoc) {
+    // Determine which state (before/after) matches the cursor selector
+    const matchesAfter = this._matcher
+      ? this._matcher.documentMatches(newDoc || {}).result
+      : true;
```
radekmie: `_matcher` should be always present.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+    const aHigh = typeof a.getHighBits === 'function' ? a.getHighBits() : (a.t ?? a.seconds ?? a.time ?? 0);
+    const aLow  = typeof a.getLowBits === 'function' ? a.getLowBits()  : (a.i ?? a.increment ?? 0);
+    const bHigh = typeof b.getHighBits === 'function' ? b.getHighBits() : (b.t ?? b.seconds ?? b.time ?? 0);
+    const bLow  = typeof b.getLowBits === 'function' ? b.getLowBits()  : (b.i ?? b.increment ?? 0);
+    if (aHigh > bHigh) return 1;
+    if (aHigh < bHigh) return -1;
+    if (aLow > bLow) return 1;
+    if (aLow < bLow) return -1;
+    return 0;
+  }
```
radekmie: 1. This should not be a method, but a helper.
2. Isn't there some sort of `compare()` method provided by the driver?

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+  async _handleChange(change) {
+    if (this._stopped) return;
+    
+    const { operationType, documentKey, fullDocument, fullDocumentBeforeChange, clusterTime } = change;
+
+    if (!SUPPORTED_OPERATIONS.includes(operationType)) {
+      return; // Ignore unsupported operations
+    }
+
+    const id = typeof documentKey._id !== 'string' ? new MongoID.ObjectID(documentKey._id.toHexString()) : documentKey._id;
```
radekmie: `_id` can be of any arbitrary type - it has to be checked first, before calling `toHexString` on it.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+  }
+
+  async _restartChangeStream() {
+    try {
+      // Close current stream using stop callbacks if they exist
+      if (this._changeStream) {
+        // Find and execute the change stream stop callback
+        const changeStreamCallback = this._stopCallbacks.find(cb => 
+          cb.toString().includes('_changeStream')
+        );
```
radekmie, line 284: It's stringifying functions to see if they refer to a field...? That's bold and IMO way too hacky.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+        try {
+          const admin = this._mongoHandle.db.admin();
+          const im = await admin.command({ ismaster: 1 });
+          return im?.operationTime || im?.$clusterTime?.clusterTime || null;
+        } catch (e3) {
+          return null;
+        }
+      }
+    }
+  }
```
radekmie, line 402: The error handling should differentiate invalid/unsupported commands and other errors (e.g., connection failure).

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+    if (!exists) {
+      await this._mongoHandle.db.createCollection(collectionName);
+    }
+
+    if (!preAndPostImagesEnabled) {
+      await this._mongoHandle.db.command({
+        collMod: collectionName,
+        changeStreamPreAndPostImages: { enabled: true }
+      });
+    }
```
radekmie: This is opinionated, but I'd find it surprising that an _observe_ driver _creates_ a collection. Especially, since it will assume some settings, e.g., no collation.

packages/mongo/changestream_observe_driver.js
```diff
@@ -0,0 +1,623 @@
...
+      // until the change stream is attached to avoid fence ordering gaps)
+      this._multiplexer.ready();
+
+      // Then start watching for changes
+      const pipeline = this._buildPipeline();
+
+      // Create change stream with appropriate options
+      const changeStreamOptions = {
+        fullDocument: 'updateLookup',
+        fullDocumentBeforeChange: 'whenAvailable'
```
radekmie, line 156: These two should be configurable, as `updateLookup` has _some_ delay, and `required` is not always possible.

radekmie (commented):
Here's my initial, **really** not in-detail review. Overall, there's a lot of uncovered edge cases (e.g., `console.warn`s or silent `catch` blocks), unexpected changes (see the first comment), and a lot of tests I haven't reviewed yet at all.

And one more question: the change stream is always opened with a very basic `$match` - that's clear. But I couldn't find any multiplexing of the observers themselves, and that'd mean every one of them would create a new change stream, which is far from ideal.

packages/ddp-rate-limiter/ddp-rate-limiter-tests.js
italojs: 👍🏻 i'll take a look on it, I remember I was having many fail tests in non-related pkgs but many was flaky tests, I extract a few changes as you can see [here](https://github.com/meteor/meteor/pull/13972) but maybe i need cut-off a litle bit more

packages/mongo/tests/observe_changes_tests.js
italojs: The order matter because if you call `insert` before add the resolver inside the  `self.expects` array, the `added` callback breaks since `self.expects` is empty` 
Screenshot 2025-11-10 at 11 36 10 AM

maybe it doent matter for oplog due a latency issue

packages/mongo/changestream_observe_driver.js
italojs: I did it since the collections must to have `changeStreamPreAndPostImages`  enabled to give me the `fullDocumentBeforeChange`, where I use to handle the the diff for update since Meteor Multiplexer just require the changes.

But looking again on it, I'll see if it's possible to use the changes that comes from change streams only 🤔

packages/mongo/changestream_observe_driver.js
italojs: you are right, I admit, that is a unreviewed code from copilot

packages/mongo/changestream_observe_driver.js
italojs: 1. I can move it to `mongo_utils.js` since it's a function to compare 2 MongoDB Timestamps, it's a generic function but the unique propose untiil now is to use in mongo
2. Unfortinately no, let me know if you know something, for sure I want to remove that function from there.

packages/minimongo/local_collection.js
italojs: You are right; it's not necessarily required. The context here is that at the beginning of the implementation, to be honest a few months ago I was a little lost in the codebase since I just arrived at Meteor and got this challenge to solve . I added it because the tests are flaky on my machine for some reason I don't know, and this made the binary tests pass. I ran the tests here filtering by the "binary" string, and all of them passed.

packages/minimongo/local_collection.js
italojs: removed with _serializedBinaryData

packages/mongo-id/id.js
italojs: i'm not using it since the package itself is a simulation of objectID and i'm not confident `_looksLikeObjectID === true` would have the HexToString function 

btw, why do we have a simulation of ObjectID instead mongo.ObjectID itself?

packages/mongo/mongo_connection.js
italojs: Do you mean the last one, right? jusr removed
but it need the matcher and the _supportsChangeStreams flag since we need to check the version and if it's a replicaset or sharded cluster

packages/mongo/mongo_connection.js
italojs: didnt know about this one, thanks

packages/mongo/mongo_connection.js
italojs: The semver is inside the tool not in the app level, I think it's OK to check the version, I wouldn't like to add a new dependency to meteor app just to check it

packages/mongo-id/id.js
radekmie: Then it should be extracted into a helper. I doubt there's _a lot_ of such objects, but calling such a method on _any_ object sounds risky.

> btw, why do we have a simulation of ObjectID instead mongo.ObjectID itself?

Cause the `mongodb` package cannot be used in the browser.

packages/mongo/tests/observe_changes_tests.js
radekmie: Then I'd extract this change to the pre-merged tests, too. Just to make sure it passed before as well.

packages/mongo/mongo_connection.js
radekmie: No, I mean the entire oplog pipeline. There's no need to run it, if we already know we'll use changestream.

packages/mongo/changestream_observe_driver.js
radekmie: 2. I didn't _know_, there was one, but I'd be surprised if there wouldn't be any. I did a quick check and there's both [`comp`/`compare`](https://mongodb.github.io/node-mongodb-native/6.21/classes/BSON.Timestamp.html#compare) as well as a few shorthands like [`lessThan`](https://mongodb.github.io/node-mongodb-native/6.21/classes/BSON.Timestamp.html#lessThan).

packages/mongo/changestream_observe_driver.js
radekmie: You can always start a change stream and only report that this option is not enabled _while processing the events_. Sure, it'll get reported later, but it brings less friction. (It's also safer, as each of the commented operations can fail.)

packages/mongo/tests/observe_changes_tests.js
italojs: this fix comes with change streams context, since it doesnt matter for oplog it should be included in this PR IMO, to mantain the context, would be useful to know this change comes after change streams

packages/mongo/changestream_observe_driver.js
italojs: good to see a doc about it but I tried it before, so since it was comming undefined i just give up, will investigate why `compare` is undefined
Screenshot 2025-11-11 at 3 40 00 PM

packages/mongo/tests/observe_changes_tests.js
radekmie: It depends on how you look at it. I'd rather not see any test _changes_ in this PR, as this suggests some things have changed for the seemingly unchanged logic.

packages/mongo/changestream_observe_driver.js
radekmie: That's a method, not class field.

italojs:
forgot to mention, i'm off for a few days off, will be back in 2 weeks

packages/mongo/tests/observe_changes_tests.js
italojs: https://github.com/meteor/meteor/pull/14018

packages/mongo/tests/observe_changes_tests.js
```diff
@@ -188,6 +188,7 @@ Tinytest.addAsync('observeChanges - unordered - basics', async function(
       ]);
 
       await c.updateAsync(fooid, { noodles: 'alright', potatoes: 'tasty', apples: 'ok' });
+      await Meteor._sleepForMs(100); // changeStream is faster than oplog, so we need to wait a bit to update the internal multiplexer.cache
```
Grubba27: Could we flush the internal `multiplexer`? so that we would only wait the necessary amount of time?

italojs: yes, i put it as a TODO before you comment but didnt push yet

italojs:
Hi @radekmie 
Addressed 99% of your requests, just is missing one, feel free to review it again

NOTES:  
1. I pushed a few changes that you can ignore while we haven't release 3.4.1:
- [test-packages.yml‎](https://github.com/meteor/meteor/pull/13787/files#diff-ebb7704d4a6cfe36b4c368e7e0ac1da7005b144e0cd0e6d59020c32fe0a72095)
- [password_tests.js‎](https://github.com/meteor/meteor/pull/13787/files#diff-862659c46eaabf2b23cb75de8905d3bb83cbd6b67ce2e95d69fb684f5001b7b5)
- "observeChanges - unordered - enters and exits result set through change" test at [observe_changes_tests.js](https://github.com/meteor/meteor/pull/13787/files#diff-fe7cf5b1fea674f63848e7beac75964fb3bad09676e5efd4c4718adc2f4a693e)
- [socket_file_tests.js](https://github.com/meteor/meteor/pull/13787/files#diff-f0d4007bf1dcb9b5efd465ff7111ba5863f6abe35837d538887c20d768288403)

All these changes I will remove when I merge it to devel (and rebase it here) in release 3.4.1

2. I ignored the changestreams unit tests for a while, I want to ensure we wont have significant changes like this last review, maybe after your next review, I will work on it

packages/mongo/changestream_observe_driver.js
italojs: I was planning to address it as a patch 3.5.1 (pr at least use it for the beta), I would like to restrict multiple config scenarios in a beta or in the first version

packages/mongo/mongo_common.js
```diff
@@ -168,3 +168,7 @@ export function replaceNames(filter, thing) {
   }
   return thing;
 }
+
+export function compareOperationTimes(opTime1, opTime2) {
+  return (new MongoDB.Timestamp(opTime1)).compare(opTime2);
+}
```
radekmie, line 201: If we create a `Timestamp` instance out of the first argument, what are the expected types here? It'd be great to add at least a JSDoc for them.

packages/mongo/mongo_connection.js
```diff
@@ -850,78 +850,237 @@ Object.assign(MongoConnection.prototype, {
...
+
+      if (reactivitySetting && !hasCustomDriverOrder) {
+        throw new Error('Meteor.settings.packages.mongo.reactivity must be a string or an array of observer drivers');
+      }
+
+      const driverClasses = {
+        changeStreams: ChangeStreamObserveDriver,
+        oplog: OplogObserveDriver,
+        polling: PollingObserveDriver,
+        pooling: PollingObserveDriver,
```
radekmie: Are we sure there's a need to support a typo in the options? 😅

Also, this piece of code grew significantly, and it wasn't short before either. How about we extract it into a `getObserveDriver` helper?

packages/mongo/mongo_connection.js
```diff
@@ -850,78 +850,237 @@ Object.assign(MongoConnection.prototype, {
...
+      }
+
+      if (hasCustomDriverOrder && configuredOrder.length === 0) {
+        throw new Error('Meteor.settings.packages.mongo.reactivity must specify at least one observer driver');
+      }
+
+      const driverChecks = {
+        changeStreams: async () => {
+          let localMatcher;
+          const reasons = [];
```
radekmie, line 944: The idea of `reasons` while nice, will result in an increased memory pressure in a performance-critical piece of code. I think it'd make more sense to accept the fact that there will be a lot of early-exits in this code. (It may be better with this code extracted from here.)

packages/mongo/mongo_connection.js
```diff
@@ -850,78 +850,237 @@ Object.assign(MongoConnection.prototype, {
...
+      }
 
-      var driverClass = canUseOplog ? OplogObserveDriver : PollingObserveDriver;
+      if (!driverClass) {
+        const errorDetails = availabilityErrors.length
+          ? ` Reasons: ${availabilityErrors.join(' | ')}`
+          : '';
+
+        throw new Error(`Unable to select a Mongo reactivity driver from configuration [${configuredOrder.join(', ')}].${errorDetails}`);
+      }
```
radekmie, line 1096: Previously this code would never throw, and I think it's a nice property to preserve. The checks for correctness of this configuration should happen on the app's start time.

v3-docs/docs/performance/change-streams-observer-driver.md
```diff
@@ -0,0 +1,64 @@
...
+        "waitUntilCaughtUpTimeoutMs": 1000
+      }
+    }
+  }
+}
+```
+
+- `delay.error`: Milliseconds to wait before restarting the stream after an error (default: `100`).
+- `delay.close`: Milliseconds to wait before restarting after an unexpected close (default: `100`).
+- `waitUntilCaughtUpTimeoutMs`: Upper bound for waiting until the stream catches up to the server's current operation time when coordinating with DDP fences (default: `1000`).
```
radekmie, line 57: It should be well documented what can happen if this timeout is hit.

v3-docs/docs/performance/change-streams-observer-driver.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Performance Notes
+
+- The driver requests `fullDocument: 'updateLookup'`, so updates trigger an extra read; keep indexes healthy to avoid slow lookups.
```
radekmie: Indexes have nothing to do in here - it's always an `_id`-based lookup. But it _may_ happen, that some changes won't be visible, e.g., if two writes will happen almost at the same time (only the last one will be acknowledged).

v3-docs/docs/performance/change-streams-observer-driver.md
```diff
@@ -0,0 +1,64 @@
...
+- Enabling MongoDB `changeStreamPreAndPostImages` on a collection lets the driver diff updates more efficiently; without it, some updates may send the full document.
```
radekmie: Have you measured it? I see it falls back to whatever's in the multiplexer, which should be up to date anyway.

v3-docs/docs/performance/change-streams-observer-driver.md
```diff
@@ -0,0 +1,64 @@
...
+- The Change Stream pipeline currently filters only by operation type; the app-side matcher applies your selector. Very high-write collections with broad selectors may produce more events than oplog tailing.
```
radekmie: Selectors have nothing to do here: the way it's done now, _all_ changes are processed, just like in the oplog implementation.

radekmie (commented):
For now I reviewed everything except for the tests and I'll need more time for that.

v3-docs/docs/performance/change-streams-observer-driver.md
italojs: I removed this feature from CS in your last comments, will remove this doc session

packages/mongo/mongo_connection.js
italojs: correct me if i/m wrong, i realized we should check for _supportsChangeStreams once, since once we startup the app or first observe and _supportsChangeStreams === true, it will be rare for it became false
I'm thinking of calling it only in the first observeDriver call/creation (or in the build maybe?)

radekmie: Once is enough. And it cannot happen during build - there's no MongoDB instance configured there.

harryadel:
:open_mouth: 
This is already tested and clear @italojs ?

PR #13797, opened 2025-06-19
Dev only packages and Npm.devDepends to optimize prod builds

### LINKED ISSUE #9612: Npm packages in the production bundle
I have some problems with npm packages in my bundles.

1. I create a new meteor project, add a npm package (lodash for example) and use it in the client-only-side (`client/` folder for example). Then I run `meteor build`. Wait (soooo many time). Open final archive and see `/bundle/programs/server/npm/node_modules/`. Among other things I see `lodash` folder. In the server area (!). What for?
1. I create a new meteor project, add a native (atmosphere) package (reywood:publish-composite for example). Then I run `meteor build`. Wait. Open final archive and see `/bundle/programs/server/npm/node_modules/meteor/reywood_publish-composite/node_modules/`. I see `eslint` folders. What for? Ok. Go to [publish-composite github repository](https://github.com/englue/meteor-publish-composite/). Open the `package.json`. All eslint packages in `devDependencies` area. Why they in the production bundle?

All of this points make my bundles sooo big. And it's very bad, I think.

Examples at my github repo: https://github.com/JWo1F/meteor-npm-bugs

Meteor version: 1.6.1.
Operating System: Fedora 21.

### LINKED ISSUE #10902: Reduce build size by excluding dev packages
When running `meteor build --server-only` on a clean project (just freshly created by the command `meteor create test-project` I end up with a folder of a size of `118mb` - of which `74mb` is taken up by `programs/server/npm/node_modules/meteor/babel-compiler`. I'm quite unsure if the rest wouldn't be needed, but if we could get out the `babel-compiler` here, that would decrease the size of our final bundle considerably.

### LINKED ISSUE #12086: react-fast-refresh is always added to the client
The package `react-fast-refresh` is always added to the client bundle, adding 32kb of unused code (unminified), even if `hot-module-replacement` is not present.

When I'm not using React then I don't need it.

Using `DISABLE_REACT_FAST_REFRESH` has no effect, the package is still added to the client.

Repro:

- create new Meteor project that not uses React
- run using `meteor`
- open chrome dev tools --> source --> coverage
- see react-fast-refresh being present but being marked as nearly 100% of unused code

My proposal - only add it when the project is created with React (making it an explicit package to be added and add it to the default `create` boilerplate) or making `DISABLE_REACT_FAST_REFRESH` having the effect, that `react-fast-refresh` is completely stripped of the bundling step.

nachocodoner:
Context: https://github.com/meteor/meteor/issues/10902

This PR adds a `devOnly` flag for packages. They work like other packages but are skipped when building the production bundle, cutting its size. In contrast with `debugOnly` as this option discard the package early, triggering build errors for build plugins like `babel-compiler`.

The image below shows the dependency removed from node_modules in the production bundle, reducing its size.[ I used `meteor profile --build --size`](https://github.com/meteor/performance/pull/18) to measure the final bundle and confirm the savings.

image

As part of the PR I will do more verifications as well as mark other `devOnly` packages, and ensure if they are still in client/server bundles for production.

`shell-server` is the only package using `babel-compiler` directly, which [also appears to be `devOnly`](https://github.com/meteor/meteor/issues/10902#issuecomment-2987441067). To include it in production, we could drop that dependency (`babel-compiler`) and use the Babel eval feature from the `@meteorjs/babel` npm package instead.

Besides, this PR also added `Npm.devDepends` to define dev dependencies, ensuring only the necessary node modules for development are included at dev time, and excluded from production (`meteor build`).

## Pending

- [x] Verify a local final bundle run works well. Verify a remote deployment locally?
- [x] Review other opportunities for devOnly packages.
- [x] Implement `Npm.devDepends` to wipe out devDependencies on packages when production


tools/isobuild/bundler.js
```diff
@@ -2558,6 +2558,16 @@ class JsImage {
           symlink: includeNodeModules === 'symlink'
         };
 
+        if (buildMode === "production") {
+          const targets = global.meteorBundlerTargets || {};
```
nachocodoner: I don’t like passing package context through global, but I’ve had trouble getting the package metadata to where I need it in JsImage.write to skip devOnly package. I just wanted to move the prototype for the devOnly option forward. I’ll keep trying to solve this properly, if you have any ideas, let me know.

Grubba27:
Docs look good!

This will help a **ton** – I think sending the `meteor-tool` to production was really bad (I think we shipped even the templates)

Grubba27 (approved):
As a future improvement, we can go through the Npm.depends and check if it makes sense to be there

PR #13803, opened 2025-06-23
Allow mutation of file content in version bump script

emmanuel-ferdman:
## PR Summary
This small PR updates the `bump-all-version-numbers.js` script to use `let` instead of `const` for the content variable. Since the file content is modified after being read, using `const` caused a reassignment error. This change allows safe in-place modification of the file content before writing it back.

### DIFF
--- a/scripts/admin/bump-all-version-numbers.js
+++ b/scripts/admin/bump-all-version-numbers.js
@@ -7,7 +7,7 @@ const packageNames = process.argv.slice(2);
 packageNames.forEach(name => {
   // name = "packages/" + name + "/package.js";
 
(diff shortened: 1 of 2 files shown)


italojs:
Amazing! It looks like your build is failing, but it isn't related to your change. I'll rebuild it to better understand what is happening.

StorytellerCZ:
@nachocodoner I don't think this was closed properly. @emmanuel-ferdman I think you will have to submit this again. Apologies.

italojs:
agree with @StorytellerCZ 
We should update this branch and point to release 3.3.2
the PR fix a bug and doa fix in the docs

PR #13810, opened 2025-06-24
Fix regression on recompile behavior

nachocodoner:
This PR fixes a regression in Meteor config recompile behavior.

When adding support for the modern option in Meteor 3.3, the config reader was changed and `nodeModules.recompile` stopped being picked up. No existing tests covered this case.

This update restores correct handling of both modern and nodeModules.recompile without errors.

## Regression vs fix

``` json

"meteor": {
    "mainModule": {
      "client": "ui/main.jsx",
      "server": "api/main.js"
    },
    "testModule":  "tests/main.js",
    "modern": true,
    "nodeModules": {
      "recompile": {
        "pdfjs-dist": true
      }
    }
  },
```

image

## Pending

- [x] Add a regression test to verify nodeModules.recompile options in various configs
- [x] Verify how nodeModules.recompile behaves in SWC

### DIFF
--- a/tools/project-context.js
+++ b/tools/project-context.js
@@ -1821,13 +1821,17 @@ export class MeteorConfig {
     let config = this._ensureInitialized();
     let filteredConfig = keys.length ? {} : config;
     if (config) {
-      keys.every(key => {
-        if (config && _.has(config, key)) {
-          filteredConfig = config[key];
-          return true;
-        }
-        return false;
-      });
+      const subConfig = config[keys[0]];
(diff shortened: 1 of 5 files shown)


italojs (changes requested):
we need to rebase release 3.3.1 with devel before the merge, this PR brings [7be740c](https://github.com/meteor/meteor/pull/13810/commits/7be740cf9c0427e8c61ca669e241ce5fc8e7d421) commit

Issue #13812, opened 2025-06-25
Upgrade deprecated packages


StorytellerCZ:
While updating the bundles this morning there is a lot of deprecated packages that we have in the core/build tool. I think we should seriously look at upgrading all of them to prevent security and other issues in the future.

Here is the output:
```
npm warn deprecated @npmcli/move-file@1.1.2: This functionality has been moved to @npmcli/fs
npm warn deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
npm warn deprecated har-validator@5.1.5: this library is no longer supported
npm warn deprecated are-we-there-yet@2.0.0: This package is no longer supported.
npm warn deprecated gauge@3.0.2: This package is no longer supported.
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated @babel/plugin-proposal-logical-assignment-operators@7.20.7: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-logical-assignment-operators instead.
npm warn deprecated @babel/plugin-proposal-async-generator-functions@7.20.7: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-async-generator-functions instead.
npm warn deprecated @babel/plugin-proposal-nullish-coalescing-operator@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-nullish-coalescing-operator instead.
npm warn deprecated @babel/plugin-proposal-optional-chaining@7.21.0: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-optional-chaining instead.
npm warn deprecated @babel/plugin-proposal-object-rest-spread@7.20.7: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-object-rest-spread instead.
npm warn deprecated @babel/plugin-proposal-optional-catch-binding@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-optional-catch-binding instead.
[...]

PR #13814, opened 2025-06-25
Update tools dependencies

nachocodoner:
Context: https://github.com/meteor/meteor/issues/13812

This PR updates a few dependencies in the Meteor tool. Before Meteor 3.3, I tested whether these updates improved performance using meteor profile, since they’re used frequently. They showed no benefit, so I avoided any side effects or uncovered use cases from these major updates.

I will review the changelogs to make sure we can update them as patches for the next release. They seem safe but need verification.


StorytellerCZ:
Related to #13812 I assume?

StorytellerCZ (dismissed):

italojs (dismissed):
The fix is correct and the approach is the right, just a few non-blocking notes:

1 `global-state.js` affects more than the title suggests. The one-line `getGlobalState` fix is necessary, but it also silently reactivates other previously-dead paths: `BUILD_CONTEXT_FILES_CLEANED` goes from "every rebuild" to "once per session", the install checks (React/TS/Angular) become cached, and the "skip if already running" guard in `startRspackClientServe/ServerWatch` now actually fires.Worth a quick sanity check that none of those regress in the dev flow.

2. Windows is intentionally not covered, so the dev server can still orphan there. 

3. Since both this `handler` and `tool-env/cleanup.js` register re-raising process.once handlers, the `async cleanup()` is usually preempted at shutdown. That's fine and by design, just noting the graceful await cleanup() rarely runs to completion.

italojs (approved):

PR #13828, opened 2025-07-02
Add a note regarding user._id being temporary in Accounts.onCreateUser

### LINKED ISSUE #13808: onCreateUser documented wrongly in regards to the temporary user._Id
### Versions

* Meteor 3.2.2
* accounts-base 3.1.0 / accounts-password 3.1.0
* Node 22.x
* macOS 15.5 Apple Silicon

---

### Problem

Docs imply the `_id` you receive in `Accounts.onCreateUser` is the one that will be stored.
In reality, `insertUserDoc` generates a *new* `Random.id()` **after** the hook and overwrites it, so the final user ends up with a different `_id`.

---

### Why it matters

Code that seeds sibling collections inside the hook with `user._id` produces orphaned documents:

```js
Accounts.onCreateUser((opt, user) => {
  Foo.insert({ ownerId: user._id }); // provisional id!
  return user;
});
```

---

### Minimal repro

```js
Accounts.onCreateUser((o, user) => {
  console.log('hook id:', user._id);
  return user;
});

// using collection hooks
Meteor.users.after.insert((id, doc) => {
  console.log('final id:', doc._id);
});
```

Running `Accounts.createUser({ username: 'alice', password: 'x' })` logs two different ids.

---

### Request

* **Docs:** State clearly that the `_id` inside `onCreateUser` is provisional and will be replaced unless the hook sets its own `_id`.
* **Optional enhancement:** If the hook *does* set `user._id`, keep that value instead of overwriting it.

repetitioestmaterstudiorum:
Fixes https://github.com/meteor/meteor/issues/13808#issuecomment-3027793777

### DIFF
--- a/docs/source/api/accounts-multi.md
+++ b/docs/source/api/accounts-multi.md
@@ -151,7 +151,8 @@ password-based users or from an external service login flow. `options` may come
 from an untrusted client so make sure to validate any values you read from
 it. The `user` argument is created on the server and contains a
 proposed user object with all the automatically generated fields
-required for the user to log in, including the `_id`.
+required for the user to log in, including a temporary `_id` (the final _id is
+generated upon document insertion and not available in this function).
 
 The function should return the user document (either the one passed in or a
 newly-created object) with whatever modifications are desired. The returned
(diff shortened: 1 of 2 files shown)


italojs:
Screenshot 2025-07-03 at 11 42 38 AM

could you please update your branch with devel via rebase?

repetitioestmaterstudiorum:
Clicked on "Sync fork", hope that's what you needed.

italojs (approved):

PR #13821, opened 2025-07-01
Release 3.3.1

nachocodoner:
## Still in development

Current released version: 3.3.1-rc.2.

To use it, please do a

``` sh
meteor update --release 3.3.1-rc.2
```

``` sh
meteor create --release 3.3.1-rc.2
```

PR #13715, opened 2025-04-29
MinifierJS - Using SWC from zodern's package

italojs:
Summary

In alignment with our ongoing effort to [make the Meteor bundler faster](https://forums.meteor.com/t/join-the-effort-to-speed-up-meteor-bundler/63406/47), this PR introduces a Proof of Concept (PoC) for integrating swc into Meteor’s build process. 
We're leveraging the excellent work done in the community by [zodern/minify-js-sourcemaps](https://github.com/zodern/minify-js-sourcemaps)

History
First approach https://github.com/meteor/meteor/pull/13709

Next Steps

Ensure proper attribution and credit to community maintainers.
Address any limitations or issues found during early testing.
Finalize the approach in collaboration with the community (a discussion channel will be opened soon).

Screenshot 2025-04-29 at 18 51 01

PR #13830, opened 2025-07-06
Integrate lai:collection-extensions into core

harryadel:
Meteor offers no [built-in way to extend the `Mongo.Collection` object](https://github.com/Meteor-Community-Packages/meteor-collection-extensions#why) and this leads to cascading problem in many community packages which try to monkey-patch their way into the core, so hopefully by adding this to the core it opens up the door to better architect community packages.


packages/mongo/collection/collection_extensions.js
```diff
@@ -0,0 +1,164 @@
...
+  
+
+  
+  /**
+   * Apply all extensions to a collection instance
+   * Called during collection construction
+   */
+  _applyExtensions(instance, name, options) {
+    // Apply constructor extensions
+    this._extensions.forEach(extension => {
```
StorytellerCZ: I would prefer to see here (and other places) `for of` loopas it is more performant than `forEach`.

StorytellerCZ (changes requested):

harryadel:
I think it's ready for a review now @radekmie @fredmaiaarantes @nachocodoner

nachocodoner:
Overall, the PR looks good, with detailed tests and documentation included.

As for including it in Core, the feature is clear, low maintenance, and important for consistent Collection expansion in community packages and app code. I’ll ask the team for their feedback.

StorytellerCZ (dismissed):

v3-docs/docs/api/collections.md
```diff
@@ -669,6 +669,89 @@ The methods (like `update` or `insert`) you call on the resulting _raw_ collecti
...
+
+Example:
+```js
+Mongo.Collection.addStaticMethod('getAllCollections', function() {
+  return Array.from(Mongo._collections.values());
+});
+
+// Now available as static method
+const allCollections = Mongo.Collection.getAllCollections();
+```
```
Grubba27, line 715: I would wish more examples on these, `addPrototypeMethod` is a very powerful method that we could have more examples:

Since is available in all collections, you can re-use some patterns: `CreateOrUpdate` is the first example that comes to my mind.

small 🏌️ but I think since we are adding things to core we could take the chance to improve this doc

packages/mongo/collection/collection_extensions.js
```diff
@@ -0,0 +1,164 @@
...
+      throw new Error('Prototype method must be a function');
+    }
+    
+    // Check for reserved names (reserved for future hook system)
+    const reservedNames = ['before', 'after', 'direct', 'hookOptions', '_hooks'];
+    if (reservedNames.includes(name)) {
+      throw new Error(`Method name '${name}' is reserved for future hook system`);
+    }
+    
+    this._prototypeMethods.set(name, method);
```
Grubba27: is there a reason for this(commit ref or something)?

This package could have in the future these hooks or am I missing something here?

harryadel: err my bad, it slipped. I'm going to remove it for now. Honestly, the motivation for this PR was ~hopefully~ improving meteor-collection-hooks since it's a mess and the 3.0 migration is full of bugs.  And this motivation was espoused by the original author

https://github.com/Meteor-Community-Packages/meteor-collection-extensions#todo
>[dburles:mongo-collection-instances](https://github.com/dburles/mongo-collection-instances) ([Refactored and tested with 100% success](https://github.com/rclai/mongo-collection-instances/tree/collection-extensions)) ([Now merged into master!](https://github.com/dburles/mongo-collection-instances/commit/7f90911b6a7117cfc62e40b200a0437ea9fb5961))

v3-docs/docs/api/collections.md
harryadel: I mean I like it as is, if you've a specific example in mind. You could tell me about and I'll gladly include it.

packages/mongo/collection/collection_extensions.js
Grubba27: Nice! I'm _very_ happy with this effort!

v3-docs/docs/api/collections.md
Grubba27: The one that comes to mind is the findById method, which is just an abstraction on top of the `find({_id ...`

But it looks good to me too

radekmie (commented):
The only thing I'm missing is the type definitions.

harryadel:
The typescript dude emerges once again. I'll have a look and let you know.

harryadel:
Please include this in [3.3.1](https://github.com/meteor/meteor/pull/13821) if possible. @nachocodoner

Grubba27:
I'll let @radekmie give his review, but on my end, it looks good.

I think this pr has the same reasoning to go in 3.4 as this one from Storytelller: https://github.com/meteor/meteor/pull/13848#issuecomment-3078362701


The TL;DR; is we would like to make 3.4 a big one with more highlights. 

3.3.1 is a patch just to solve the latent issues we currently have in 3.3

harryadel:
Thank you Gabriel for your prompt reply and review. That's definitely understandable but I was keen on getting an RC out so I can starting testing.

packages/mongo/mongo.d.ts
```diff
@@ -92,6 +92,83 @@ export namespace Mongo {
     getCollection<
         TCollection extends Collection<any, any> | undefined = Collection<NpmModuleMongodb.Document> | undefined
     >(name: string): TCollection;
+
+    // Collection Extensions API
+    /**
+     * Add a constructor extension function that runs when collections are created.
+     * @param extension Extension function called with (name, options) and 'this' bound to collection instance
+     */
+    addExtension(extension: (this: Collection<any, any>, name: string | null, options?: any) => void): void;
```
radekmie: 1. It's a matter of preference, but for safety reasons it'd be better to use `unknown` instead of `any` when possible.
2. Why is `options` not typed?

packages/mongo/mongo.d.ts
```diff
@@ -92,6 +92,83 @@ export namespace Mongo {
...
+
+    /**
+     * Add a prototype method to all collection instances.
+     * @param name The name of the method to add
+     * @param method The method function, bound to the collection instance
+     */
+    addPrototypeMethod(name: string, method: Function): void;
```
radekmie: 1. The comment about `method` being bound to the collection instance should be reflected in the type by using the `this` parameter.
2. There's the wiring step missing, i.e., it's _easy_ to add a method, but the new method won't be typed. I'm not saying it has to be done now, but that'd be needed to make this feature fully typesafe. (However, it most likely will have to be based on in-app `interface` extensions.)

_(The same applies to the other methods.)_

packages/mongo/mongo.d.ts
```diff
@@ -92,6 +92,83 @@ export namespace Mongo {
...
+
+    /**
+     * Backwards compatibility alias for lai:collection-extensions
+     * @param name The name of the method to add
+     * @param method The method function, bound to the collection instance
+     * @deprecated Use addPrototypeMethod instead
```
radekmie: That's interesting: are we sure we want to add new **deprecated** code to Meteor's Core? I get the compatibility part, but if it's about stabilization of a package by bringing it to the core, maybe it should be clean from the start?

radekmie (commented):
If everyone else is fine with this PR, please don't block it because of me!

packages/mongo/mongo.d.ts
harryadel: Because I want to make the transition with little to no friction. If you've a hard opinion on this, then I don't mind removing the backwards compatibility part. :man_shrugging:

radekmie: I'd rather let @Grubba27 decide. But as it'll significantly increase the API surface of Meteor's Core, I'd keep it as lean as possible.

nachocodoner: I'm particularly interested on keeping backwards compatibility whenever possible, with clear deprecation warnings shown directly in the program. We can introduce breaking changes in 3.4 written in the changelog notes, but those are often annoying and easy to miss during a simple upgrade. It's better to avoid confusing or breaking the user experience. Other software often handles this similarly.

I'd keep backwards compatibility, using clear warnings via `Meteor.deprecate(*)`. In the next minor version, we can review and remove those deprecations, just add a check for this on 3.5 to handle it later.

radekmie: Just to be clear - these are _new_ APIs mimicking some package APIs. So backward compatibility in this context is not necessarily a thing, since users _will_ have to handle breaking changes anyway (i.e., they’ll switch from a package to not using it). While at it, they _could_ rename the methods. (As I understand there are no other differences.)

nachocodoner: Maybe I misunderstood the scope of this package. From what I understand, it's mainly used for Meteor packages that extend collection functionality, right?

There are two perspectives for breaking changes here: package and app users.

- Package users need to update their packages to stop using lai:collection-extensions and switch to the new core API. They'll bump the version (minor or major), then wait for app users to upgrade.

- App users need to upgrade both Meteor and the packages that use the new core API. But if any of their packages haven’t updated yet, things might break. They could end up in an intermediate state, Meteor upgraded to the latest version with the new API, but some packages not yet bumped or compatible. Since they’re not the maintainers, they often can’t fix this directly. More commonly, they report the issue and wait for maintainers to respond. With backwards compatibility, app developers could safely run in that intermediate state, see a deprecation warning, report it, and then upgrade the package once it’s ready.

Could I be missing something here?

radekmie: If a some package depends on `lai:collection-extensions` then there's already a problem after updating Meteor, since you'll have two instances of these methods. (It needs further investigation whether or not it'll break the functionality.)

But assuming it wouldn't, then yes, it _may_ be helpful to keep these deprecated methods for people who:
1. Are using this package.
2. Are using _some other_ package that depends on its deprecated methods.
3. Are updating Meteor to the latest version.

IMHO, there's virtually no one in such a position, but that's just me.

packages/mongo/mongo.d.ts
italojs: 'Unknown' is enough, but it doesn't make sense to type it unless we use 'any/unknown,' unless we have a good reason for it.
An option here is let the developer to infer it, like
```
addExtension<TSchema, TDoc>(
  extension: (this: Collection<TSchema, TDoc>, name: string | null, options?: any) => void
): void
```

packages/mongo/mongo.d.ts
italojs: I do not agree with bringing deprecated methods to new APIs. I know we are bringing the full package to the core, but that does not mean we should do it `as is`. The original pkg's code for sure will speed up the implementation and increase the reliability, but as [Minify-swc](https://github.com/meteor/meteor/pull/13715), I see we should bring just what matters for us (99% ofthe pkg in this case)
I see we bring others pkgs to meteor in the future, so if we always bring  the pkgs `as is` to the core, including the deprecations, we will create a snowball of messy code IMO

In the worst case, the developer can install the original pkg while they don't migrate to the new API 🤷🏻‍♂️

harryadel: I'm a bit confused here guys. In layman's terms, can you tell me what can I do to move this PR forward? Thanks!

nachocodoner: I agree with @radekmie and @italojs after our discussion. Lets don't introduce any breaking change with new code added into the core.

harryadel: So keep it as is, right?

nachocodoner: Sorry, I refered to not introduce any deprecation logic (as they discussed), I express wrongly. 😅 So we need to remove any reference on deprecations, just the new API.

harryadel: Please have another look Nacho man @nachocodoner and mango man @radekmie

packages/mongo/collection/collection_extensions.js
StorytellerCZ: On that note having official `before` and `after` hooks for collection actions is something that we need to have in core.

harryadel: That's a different story for a different time :sweat_smile:

packages/mongo/mongo.d.ts
nachocodoner: Looks good to me now according to what we discussed here 👌

harryadel:
Are the failing test a result of my changes or something off with the CI? @italojs @nachocodoner 

EDIT: ping?

nachocodoner:
We'll merge this PR as part of Meteor 3.4 contribution phase, so that we can have time to test Rspack bundler in isolation. I have set the milestone for it.

nachocodoner (approved):
This PR is ready! Lets merge it for a Meteor 3.4 beta 😄

Great work @harryadel 🚀

harryadel:
happy breaking bad GIF

PR #13833, opened 2025-07-06
Ensure Solid skeleton app's Vite config file is loaded using ESM over CJS.

JorgenVatle:
Related to #13832.

This pull request updates the Solid.js skeleton app to use an `.mjs` file extension instead of `.js` for the Vite config. Using `.js` results in the config likely being loaded only with CommonJS support, which can cause issues when importing ESM-only plugins (e.g. TailwindCSS v4). Which in turn can cause the Meteor server to throw a somewhat cryptic module error due to the inability to properly resolve imports within the Vite config.

While I’ve added warning messages to Meteor-Vite to help users identify this issue, it’s likely better to avoid the problem altogether by using the correct file extension upfront. 😄

Like the pull request from earlier, I did a couple minor bits of maintenance:
- Utilize unused `client/main.js` instead of `imports/ui/main.jsx` as the primary Vite entrypoint to follow the same convention used for the server's Vite entrypoint.
- Fixed potential confusion with the app's `main.css` file by moving it into `imports/ui`, which would otherwise be redundantly processed by both Meteor and Vite if imported directly. Something which users would have to do if they want to use Vite for CSS processing. (E.g. to use Tailwind v4)
- Added docblocks to the empty `entry-meteor.js` modules with some general information about their purpose within the scaffolded app.


italojs:
is the PR https://github.com/meteor/meteor/pull/13832 a dependency for this PR?

nachocodoner:
Merging this to a core-based branch to verify it works properly.

nachocodoner (approved):

JorgenVatle:
> is the PR #13832 a dependency for this PR?

Oh, no it's an independent set of changes. Just referenced the prior PR as the changes introduced were very similar. Sorry about the confusion. 😄

nachocodoner:
@JorgenVatle: I have tested and the skeleton work well. I see though an error that triggers initially:

image

> ERROR  failed to read input source map: failed to find input source map file "server-runtime.mjs.map" in "packages/jorgenvatle:vite/dist/server-runtime.mjs" file as either "packages/jorgenvatle:vite/dist/server-runtime.mjs.map" or with appended .map
    at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/swc-21.0.0/src/lib.rs:401

This error is due SWC tried to find the `//# sourceMappingURL=server-runtime.mjs.map` described at `packages/jorgenvatle:vite/dist/server-runtime.mjs``. Who generates this file? Could you avoid generating a source map? Or do you think is needed? Maybe is something I need to fix on SWC setup.

I don't see this being a breaking change though. So I merged your changes, https://github.com/meteor/meteor/pull/13835

JorgenVatle:
Aha, I was wondering why this started appearing recently. 😂

Something I can address for sure. The `jorgenvatle:vite` package is compiled with an external build tool before being passed onto Meteor. Is there an official/recommended way to define source maps for files built outside of Meteor's build system?

The plugin has very little code however since it mainly just hooks into the associated npm package, so omitting source map generation likely won't have any noticeable impact in case it's there's effort involved in adding support for external source maps. I imagine there aren't many users who pre-bundle their Meteor packages in the first place. 😅

PR #13838, opened 2025-07-07
Local skeleton first

nachocodoner:
This PR ensures the skeleton is cloned always from local copy when possible; otherwise, it fetches from the examples repository. 

Skeletons are duplicated: some are in Meteor core and others in external repos, such as [meteor/skel-solid](https://github.com/meteor/skel-solid) and [static-assets/skel-solid](https://github.com/meteor/meteor/tree/devel/tools/static-assets/skel-solid). In these examples, meteor/skel-solid is outdated and fails with Meteor 3, having this duplication also makes maintenance harder. A local-first approach ensures it uses the compatible local version when available and the remote version when not.

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -1203,71 +1203,85 @@ main.registerCommand({
     toIgnore.push(/(\.html|\.js|\.css)/);
   }
 
-  try {
-    // Prototype option should use local skeleton.
-    // Maybe we should use a different skeleton for prototype
-    if (options.prototype) throw new Error("Using prototype option");
-    // if using the release option we should use the default skeleton
-    // using it as it was before 2.x
-    if (release.explicit) throw new Error("Using release option");
+  const copyFromLocalSkeleton = async () => {
+    await files.cp_r(
+      skeletonPath,
+      appPath,
+      {
+        transformFilename: function (f) {
+          return transform(f);
+        },
+        transformContents: function (contents, f) {
+          // check if this app is just for prototyping if it is then we need to add autopublish and insecure in the packages file
+          if (/packages/.test(f)) {
+            const prototypePackages = () =>
+              "autopublish             # Publish all data to the clients (for prototyping)\n" +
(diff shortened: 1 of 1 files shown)


tools/cli/commands.js
```diff
@@ -1203,71 +1203,85 @@ main.registerCommand({
...
+    await files.cp_r(
+      skeletonPath,
+      appPath,
+      {
+        transformFilename: function (f) {
+          return transform(f);
+        },
+        transformContents: function (contents, f) {
+          // check if this app is just for prototyping if it is then we need to add autopublish and insecure in the packages file
+          if (/packages/.test(f)) {
```
italojs, line 1216: i think it isnt compatible with windows because of `/` 
It would be nice to test it first

italojs (changes requested):

tools/cli/commands.js
nachocodoner: This is not new code, just a refactor.

Why it won't work in Windows? This is not a path `/`, is just a regex delimiter, and Node/JS syntax is cross-os compatible.

tools/cli/commands.js
```diff
@@ -1203,71 +1203,85 @@ main.registerCommand({
...
+    options.prototype ||
+    release.explicit;
+  if (useLocalSkeleton) {
+    // Use local skeleton
+    await copyFromLocalSkeleton();
+  } else {
+    try {
+      // Prototype option should use local skeleton.
+      // Maybe we should use a different skeleton for prototype
+      if (options.prototype) throw new Error("Using prototype option");
```
Grubba27, line 1260: Does it make sense to have these ifs statements if we are already checking them before entering this try/catch ? I think is fine to maintain and not break current behavior

Grubba27 (approved):

tools/cli/commands.js
nachocodoner: Yeah, this has been a refactor, I didn't change it just in case is used in an edge case. But it shouldn't... makes sense

italojs (approved):

PR #13840, opened 2025-07-07
Accounts UI CSS upgrade

StorytellerCZ:
I went ahead and upgraded the CSS for Accounts UI.

Changes:
* Dropped IE8 and other ancient workarounds
* Removed LESS, now just pure CSS
* Detect dark theme
* Using CSS variables to allow users to override the look
* Detects dark theme and automatically applies it
* Updated the look to follow a bit more closer with meteor.com

New look:

image


StorytellerCZ:
We need to discuss the location of the CSS. Right now it is in the `accounts-ui-unstyled` and then imported only if `accounts-ui` is present. Honestly it makes no sense to me and I would like to move it to `accounts-ui` (after testing that it still works as intended).

StorytellerCZ:
@italojs v3.3.1? Probably not a bad choice that I think about it. This might be a sexy things to have on the changelog to draw people in.

packages/non-core/blaze
nachocodoner, line 1: Why do you have changes on the blaze git submodule? It shouldn't have changes there, right?

StorytellerCZ: Just update to the latest version

italojs:
@StorytellerCZ i'm sorry, what do you mean? it will be in the change logs anyway
but if you prefer, we could move it for release 3.4, but it will take a while to be published

StorytellerCZ:
3.3.1 is a fine for release.

nachocodoner:
I’m not sure this belongs in a patch. If I’m right, it introduces visual breaking changes for production apps. All developers using the current Accounts UI would suddenly see styling updates in a patch. This needs its own breaking‐change or migration notes. I’d prefer to include and announce it in Meteor 3.4.

StorytellerCZ:
@nachocodoner I agree. That is why I was surprised when @italojs targeted the 3.3.1 branch. Moved everything to 3.4.0 branch target.

nachocodoner:
I'm keeping an eye on this PR for a later Meteor 3.4 beta, since it includes an UI breaking change. Together [with another update on my side](https://github.com/meteor/meteor/pull/13797) that could cause some issues (though no breaking changes expected), I prefer to include them in a later round of 3.4 beta releases.

nachocodoner:
@StorytellerCZ: https://forums.meteor.com/t/how-to-prevent-web-browser-legacy-from-being-built-with-meteor-build/64372/4

image

Are you aware of any errors after these changes? Could this be caused by some cache state? It would be good to double check that nothing breaks. I do not have a Meteor app with accounts-ui prepared yet, I will set one up soon to verify. If you can double check in the meantime, that would be great. Maybe it's just something app-specific.

StorytellerCZ:
I did a quick check and it worked fine.
The only thing I can think of is that something might get wrong with import in `packages/accounts-ui/login_buttons.css`.
Cache could be an issue.
I would also double check the import statement in the generated code. Since that happened in the legacy bundle (from my understanding) I wonder if it is something in the bundle that might be causing it. I tried to get the legacy bundle only, but did not have time to get it working so I can't say at the moment.

nachocodoner:
For me it worked well too. Not sure why, but as mentioned it may be a cache issue, and with a proper clean it should go away. I will keep pending if that user gets more details on it anyway.

PR #13843, opened 2025-07-09
[test-in-browser] Select and re-run tests by clicking them

harryadel:
One of the things that annoyed me when developing/testing any Meteor package is once I had clicked a test group and then refreshed the browser to re-run the tests, it'd take me back to the full test suite instead of re-running the test group I had already clicked! This PR aims to fix this problem by saving the test group in browser. 

https://github.com/user-attachments/assets/e316346c-8301-4c77-9b7b-c3cfe6618c87


packages/test-in-browser/driver.js
```diff
@@ -43,12 +43,119 @@ window.onerror = (message, source, line) => {
...
+    } catch (e) {
+      console.warn('Invalid group path in URL:', match[1]);
+    }
+  }
+  return ["tinytest"];
+};
+
+var setGroupPathInURL = function(groupPath, pushState = true) {
+  var newURL;
+  if (groupPath && groupPath.length > 0 && JSON.stringify(groupPath) !== JSON.stringify(["tinytest"])) {
```
italojs: since we arent comparing complexy json structures, would be faster and cheaper we compare it by processing each value instead od serialize to a string

like
``` js
function arraysEqual(a, b) {
  return a.length === b.length && a.every((v, i) => v === b[i]);
}
```

benchmark https://jsben.ch/e2MxV
Screenshot 2025-07-14 at 10 04 30 AM

packages/test-in-browser/driver.js
```diff
@@ -43,12 +43,119 @@ window.onerror = (message, source, line) => {
...
+    newURL = '/group/' + encodeURIComponent(JSON.stringify(groupPath));
+  } else {
```
italojs: to avoid an else block, would cold start the newURL = '/'; since it's a dafault value

packages/test-in-browser/driver.js
```diff
@@ -43,12 +43,119 @@ window.onerror = (message, source, line) => {
...
+  countDep.changed();
+  topLevelGroupsDep.changed();
+  
+  // Get current group path from URL (in case of refresh)
+  var currentGroupPath = getGroupPathFromURL();
+  Session.set("groupPath", currentGroupPath);
+  
+  // Only update URL if it's actually different from what we expect
+  var expectedURL;
+  if (currentGroupPath && currentGroupPath.length > 0 && JSON.stringify(currentGroupPath) !== JSON.stringify(["tinytest"])) {
```
italojs, line 151: let me know if i'm wrong but this code looks repeated with line 60, would be nice to have a function for it

italojs (changes requested):
loved the PR, just a few things to solve ;)

harryadel:
@italojs I applied your suggestion. Also I took the opportunity to replace https://github.com/google/diff-match-patch with https://github.com/kpdecker/jsdiff. Since the original package has been archived and isn't actively maintained. There's a lot more room to improve this package honestly.

italojs:
lgtm, waitting the ci be green

PR #13848, opened 2025-07-14
Pretty Server Error

StorytellerCZ:
Upgraded the look of the server error page.
This has been a longstanding TODO there.
It also now links to docs, guide and forums to find help.

New look takes into account the browser theme settings:
image


tools/runners/run-proxy.js
```diff
@@ -244,38 +244,169 @@ Object.assign(Proxy.prototype, {
 });
 
 function showErrorPage(res) {
-  // XXX serve an app that shows the logs nicely and that also
-  // knows how to reload when the server comes back up
+  // TODO: reload when the server comes back up
```
italojs, line 247: Do you mind opening an issue and then track it here?

italojs (changes requested):

tools/runners/run-proxy.js
StorytellerCZ: Done #13849

tools/runners/run-proxy.js
```diff
@@ -244,38 +244,169 @@ Object.assign(Proxy.prototype, {
...
-      res.write(`</pre>
+  res.write(`</div>
+      </div>
+      
+      <div class="hint">
+        Fix the error in your code and save your files. Once your server is running without errors, then reload this page.
+      </div>
+      <div class="links">
+        <a href="https://docs.meteor.com" target="_blank" rel="noreferrer noopener">Docs</a>
+        <a href="https://guide.meteor.com" target="_blank" rel="noreferrer noopener">Guide</a>
```
nachocodoner, line 406: The guide hasn’t been migrated to Meteor 3.0. Should we include it now as a troubleshooting link?

nachocodoner: Maybe we could also include the Discord link (https://discord.com/invite/3w7EKdpghq) for troubleshooting, since we’ve been getting reports and support requests there. However, the forums are preferable, as more people see and address issues there.

StorytellerCZ: There is still a lot of useful information in guide and it will be hopefully migrated in the future so I think it should stay.
I will add Discord.

nachocodoner: Yes, I agree. At the same time, I think guide can raise other issues for Meteor 3 users since it uses the sync API across the guide. But it’s fine for now; we’ll see. If we start getting a lot of reports, that will be a symptom and we’ll likely need to refresh the guide at some point.

StorytellerCZ: Added

Grubba27:
That looks _insanely_ good! Thanks a lot for your work here @StorytellerCZ 

I'll move to 3.4 so that we can do some content around this change(people should see how polished Meteor is)

In 3.3.1 we plan on only fixing issues consequently our updates in the forums and in youtube will be shorter

But for 3.4 we plan to do a more planned release, where I think having this change as an item in the highlight would be nice

StorytellerCZ:
I've run into this opportunity while trying to fix another issue. I think there is a few more opportunities like this.

wreiske:
This looks great! Thoughts on adding https://highlightjs.org/ to make the log output a little prettier? ❤️

StorytellerCZ:
@wreiske not really, I just came quickly across this so I haven't thought of those. I will look into it.

StorytellerCZ:
@italojs @nachocodoner since this branch originally targeted 3.3.1 there is a lot of extra commits here as 3.4 does not have the changes from that branch yet.

italojs:
@StorytellerCZ  We will release 3.3.1 soon, then we can do a rebase for this branch. In the last case, we can do a cherry pick, but I hope it's not necessary after merging 3.3.1 into devel.

nachocodoner:
I'm about to review the PRs for the upcoming Meteor 3.4 releases listed at https://github.com/meteor/meteor/milestone/122

They will be included in 3.4 only if they pass all PR checks. Getting them green soon could allow a beta release earlier.

This change seems to still fail a test.

StorytellerCZ:
Given that this is based on `release-3.4` that branch will have to be green first to give accurate look on the tests here.

nachocodoner:
Release 3.4 branch was green, but it might turn red again as I merge some contributions (it shouldn't though as they were green). This PR was failing before, so let’s wait for the release 3.4 branch to be stable first, and then I’ll retry here.

nachocodoner:
Merging this.

Do you think we should remove the guide link? It's not updated for Meteor 3 and recently we have been migrating content directly to docs.

nachocodoner (approved):

Issue #13696, opened 2025-04-11
NPM deep Mongo dependency conflicts with Meteor's mongo package


welkinwong:
**Meteor 3.2**

After installing an npm package with Mongo dependencies, Meteor fails with this error:
```
welkin@Mac issue-top-level-await2 % meteor                                                                   
[[[[[ ~/code/issue-top-level-await2 ]]]]]     

=> Started proxy.                             
=> Started HMR server.                        
=> Errors prevented startup:                  
   
   While building for web.browser:
   node_modules/bson/lib/bson.mjs:239:28: Unexpected reserved word 'await'. (239:28)
   
=> Your application has errors. Waiting for file change.
=> Started MongoDB.
```

This is a minimal reproducible example.
https://github.com/welkinwong/issue-top-level-await

PS: After resetting everything, the error still persists, so it's unlikely to be cache-related.
```bash
meteor reset && rm -rf package-lock.json && rm -rf node_modules && meteor npm i && meteor
```

The error only occurs when a Meteor package in my /packages directory imports and uses Mongo.  This makes me suspect a potential conflict between the two Mongo.

in `/packages/abc/`
```
import mongo from 'mongodb';
mongo;
```

install npm `meteor npm i sharedb-mongo`
This npm package has a Mongo dependency.

---

Apologies if my explanation isn't perfect - I'm doing my best to explain this tricky issue. Really appreciate any help!

italojs:
maybe related to https://github.com/meteor/meteor/issues/13652 and https://github.com/meteor/meteor/pull/13736

PR #13850, opened 2025-07-14
Handling MongoCompatibilityError when connect into mongodb driver 6.x

italojs:
Now we are able to choose between mongo 3.6 driver or latest

How it works

### Packages loading

1. I defined a `connect function` that receives a client and try to connect using the mongodb driver v6.x
2. When we try to connect in a mongo <= 3.6, we got a `MongoCompatibilityError` error
3. If the project already uses `npm-mongo-legacy`, it's fine, we use it for future connections 

History -> this PR -> https://github.com/meteor/meteor/pull/13736 -> https://github.com/meteor/meteor/issues/13696 -> https://github.com/meteor/meteor/pull/13735

### DIFF
--- a/packages/npm-mongo/wrapper.js
+++ b/packages/npm-mongo/wrapper.js
@@ -19,7 +19,7 @@ function connect(client) {
   })
 }
 
-if(process.env.MONGO_URL?.startsWith('mongodb://')){
+if(process.env.MONGO_URL && process.env.MONGO_URL.startsWith('mongodb://')){
   // If we aren't using the default Meteor's MongoDB(>6), we should to check the mongo version
   connect(new MongoClient(process.env.MONGO_URL)).then(client => {
     if (client) client.close();
--- a/packages/npm-mongo/wrapper.js
+++ b/packages/npm-mongo/wrapper.js
@@ -16,7 +16,7 @@ function connect(client) {
 }
 
 connect(new MongoClient(process.env.MONGO_URL)).then(client => {
-  if (client) client.close().catch(() => {});
+  if (client) client.close()
 });
 
 const useLegacyMongo = Package['npm-mongo-legacy']
--- a/packages/npm-mongo/wrapper.js
+++ b/packages/npm-mongo/wrapper.js
@@ -19,7 +19,7 @@ function connect(client) {
   })
 }
 
-if(process.env.MONGO_URL){
+if(process.env.MONGO_URL?.startsWith('mongodb://')){
   // If we aren't using the default Meteor's MongoDB(>6), we should to check the mongo version
   connect(new MongoClient(process.env.MONGO_URL)).then(client => {
     if (client) client.close();
--- a/packages/npm-mongo/wrapper.js
+++ b/packages/npm-mongo/wrapper.js
@@ -1,25 +1,29 @@
 const { MongoClient, MongoServerSelectionError, MongoCompatibilityError } = Npm.require('mongodb');
 
-async function ableToConnectToMongo(url) {
-  let client;
-  try {
-    client = new MongoClient(url);
-    await client.connect();
-  } catch (error) {
-    if (error.cause instanceof MongoCompatibilityError && error.message.includes('maximum wire version')) {
-      console.warn('Legacy MongoDB version detected, using mongo-legacy package:', error.message);
-      console.warn('Warning: MongoDB versions <= 3.6 are deprecated. Some Meteor features may not work properly with this version. It is recommended to use MongoDB >= 4.');
-    } else console.error('Failed to get MongoDB server version:', error);
-    return false;
-  } finally {
-    if (client) await client.close().catch(() => {});
-    return true
-  }
+function connect(client) {
+  return client.connect()
+    .catch(error => {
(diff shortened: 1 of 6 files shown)


packages/npm-mongo/wrapper.js
```diff
@@ -1,6 +1,29 @@
...
 
+function connect(client) {
+  return client.connect()
+    .catch(error => {
+      if (error.cause instanceof MongoCompatibilityError && error.message.includes('maximum wire version')) {
+      console.warn(`[DEPRECATION] Legacy MongoDB version detected, using mongo-legacy package: ${error.message}
+        Warning: MongoDB versions <= 3.6 are deprecated. Some Meteor features may not work properly with this version.
+        It is recommended to use MongoDB >= 4.`);
+      if (!Package['npm-mongo-legacy']) {
+        throw new Error('Please, install npm-mongo-legacy package to use this version of MongoDB running "meteor add npm-mongo-legacy"');
```
nachocodoner: We should probably add the notice that they should install this before mongo package in their package json list as stated in the base PR.

> When listed before the mongo package in .meteor/packages, it forces Mongo to use version 6.9.0 for apps that import it.
https://github.com/meteor/meteor/pull/13736

nachocodoner: Alternatively, I'm not sure, it may be a way to modify the `meteor add` command and treat the use case of npm-mongo-legacy installation differently and add it automatically before mongo by its own, to give more automation and avoid commonly issues with this.

packages/npm-mongo/wrapper.js
```diff
@@ -1,6 +1,29 @@
...
+      }
+      return false
+    } else throw new Error(`Failed to initialize Meteor's npm-mongo package: ${error}`);
+  })
+}
+
+connect(new MongoClient(process.env.MONGO_URL)).then(client => {
```
nachocodoner: This should be done only when process.env.MONGO_URL is set, likely on a prod-like environment. Locally, when using the built-in mongo should not be triggering, it will try to connect on an empty url connection. This is likely the reason all tests fail on this branch.

tools/project-context.js
```diff
@@ -1217,7 +1217,17 @@ Object.assign(exports.ProjectConstraintsFile.prototype, {
           constraint: constraintToAdd,
           trailingSpaceAndComment: ''
         };
-        self._constraintLines.push(lineRecord);
+        if (constraintToAdd.package === 'npm-mongo-legacy') {
+          // Procura o índice do mongo
```
nachocodoner: Este comentário está noutra língua.

packages/npm-mongo/wrapper.js
```diff
@@ -1,6 +1,36 @@
...
+      return false
+    } else {
+      // log the stack
+      console.error(error.stack);
+      throw new Error(`Failed to initialize Meteor's npm-mongo package: ${error}`);
+    }
+  })
+}
+
+if(process.env.METEOR_MONGO_PORT){
```
nachocodoner: Why `METEOR_MONGO_PORT`? Shouldn't it be `MONGO_URL`?

packages/npm-mongo/wrapper.js
```diff
@@ -15,3 +45,5 @@ try {
 NpmModuleMongodbVersion = useLegacyMongo
   ? Package['npm-mongo-legacy'].NpmModuleMongodbVersion
   : Npm.require('mongodb/package.json').version;
+
+NpmModuleMongodb.connect = connect
```
nachocodoner: Why are we exporting or including connect here? Is it required for this to work? Is it used elsewhere?

packages/npm-mongo/wrapper.js
```diff
@@ -1,6 +1,36 @@
...
+      return false
+    } else {
+      // log the stack
+      console.error(error.stack);
+      throw new Error(`Failed to initialize Meteor's npm-mongo package: ${error}`);
+    }
+  })
+}
+
+if(process.env.MONGO_URL && process.env.MONGO_URL.startsWith('mongodb://')){
```
nachocodoner: The check `process.env.MONGO_URL.startsWith('mongodb://')` will skip `mongodb+srv://…` connection strings entirely. 

Could this work just with `process.env.MONGO_URL.startsWith('mongodb')`? Or just contamplate both scenarios `process.env.MONGO_URL.startsWith('mongodb://') || process.env.MONGO_URL.startsWith('mongodb+srv://')` or a regex `/^mongodb(?:\+srv)?:\/\//`.

packages/npm-mongo/wrapper.js
```diff
@@ -1,6 +1,36 @@
...
+  // If we aren't using the default Meteor's MongoDB(>6), we should to check the mongo version
+  connect(new MongoClient(process.env.MONGO_URL)).then(client => {
+    if (client) client.close();
+  });
+}
+
+const useLegacyMongo = Package['npm-mongo-legacy']
```
nachocodoner: I'll keep the coercing to boolean here, semantically this variable is used as boolean. So using ` !! Package['npm-mongo-legacy'];` or `Package['npm-mongo-legacy'] != null;`.

nachocodoner (approved):

PR #13859, opened 2025-07-23
Move findUserByEmail method from accounts-password to accounts-base

### LINKED ISSUE #13834: Accounts: findUserByEmail is only present when accounts-password is added
Meteor release: 3.2.2
accounts-passwordless: 3.0.1

I am transitioning from  `accounts-password` to  `accounts-passwordless`.

After my passwordless setup has been completed I removed `accounts-password` from the packages, which caused the following error:

```shell
I20250707-10:33:35.116(2)? error on boot.js TypeError: Accounts.findUserByEmail is not a function
I20250707-10:33:35.116(2)?     at imports/startup/server/accounts.js:81:40
I20250707-10:33:35.116(2)?     at processTicksAndRejections (node:internal/process/task_queues:105:5)
I20250707-10:33:35.116(2)?     at /tools/static-assets/server/boot.js:453:5
I20250707-10:33:35.116(2)?     at /tools/static-assets/server/boot.js:503:5
I20250707-10:33:35.116(2)?     at startServerProcess (/tools/static-assets/server/boot.js:501:3)
I20250707-10:33:35.116(2)? TypeError: Accounts.findUserByEmail is not a function
I20250707-10:33:35.116(2)?     at imports/startup/server/accounts.js:81:40
I20250707-10:33:35.116(2)?     at processTicksAndRejections (node:internal/process/task_queues:105:5)
I20250707-10:33:35.116(2)?     at /tools/static-assets/server/boot.js:453:5
I20250707-10:33:35.116(2)?     at /tools/static-assets/server/boot.js:503:5
I20250707-10:33:35.116(2)?     at startServerProcess (/tools/static-assets/server/boot.js:501:3)
```
Apparently, `Accounts.findUserByEmail` is only part of `accounts-password`. However, it should also be part of `accounts-passwordless` as it very likely involves email.

I can provide a PR if you like

copleykj:
Fixes #13834

### DIFF
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -333,6 +333,30 @@ export class AccountsServer extends AccountsCommon {
     return user;
   }
 
+  /**
+   * @summary Find a user by one of their email addresses.
+   * @locus Server
+   * @param {String} email The email address to look for
+   * @param {Object} [options]
+   * @param {Object} options.fields Limit the fields to return from the user document
+   * @returns {Promise<Object>} A user if found, else null
+   * @importFromPackage accounts-base
+   */
+  findUserByEmail = async (email, options) =>
+    await this._findUserByQuery({ email }, options);
+
+  /**
+   * @summary Find a user by their username.
+   * @locus Server
+   * @param {String} username The username to look for
+   * @param {Object} [options]
+   * @param {Object} options.fields Limit the fields to return from the user document
+   * @returns {Promise<Object>} A user if found, else null
+   * @importFromPackage accounts-base
+   */
+  findUserByUsername = async (username, options) =>
+    await this._findUserByQuery({ username }, options);
+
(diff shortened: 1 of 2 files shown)


italojs:
As I understand it, we should duplicate it instead of moving it, since the accounts-password is still in use. The accounts-passwordless and accounts-password are two different packages with different behaviors but similar purposes.

copleykj:
@italojs as far as I could tell they function very similarly. `findUserByEmail` is already attached to the Accounts object within the `accounts-password` package, so it seemed like a natural move. I didn't touch the underscored methods even though they're pseudo designated as private, just in case there is some other 3rd party package that decided to use them, but I can change that if necessary.

StorytellerCZ (approved):

nachocodoner (approved):

italojs (approved):

PR #13861, opened 2025-07-24
refactor: Update package callback-hook

dolgarev:
This PR introduces general maintenance and improvements to the package. It addresses:

- Refactoring by removing duplicated code (e.g., in `forEachAsync`).
- A bug fix in the `clear` method.
- Improved code clarity through refining some function names.

These changes enhance the package's overall maintainability and future extensibility.


italojs:
1. Could you please provide a clearer description of the bug fix? What was the bug?

2. The tests is failing:
```
Tests complete with 4 failures
Tests complete with 1819 passes
C: tinytest - spiderable - default hooks registered failed: {"groupPath":["tinytest","spiderable"],"test":"default hooks registered","events":[{"sequence":472,"type":"fail","details":{"type":"assert_equal","expected":"2","actual":"0","not":false},"cookie":{"name":"spiderable - default hooks registered","offset":0,"groupPath":["tinytest","spiderable"],"shortName":"default hooks registered"}}]}
C: tinytest - spiderable - is not ready with a custom hook failed: {"groupPath":["tinytest","spiderable"],"test":"is not ready with a custom hook","events":[{"sequence":484,"type":"fail","details":{"type":"assert_equal","expected":"2","actual":"0","not":false},"cookie":{"name":"spiderable - is not ready with a custom hook","offset":1,"groupPath":["tinytest","spiderable"],"shortName":"is not ready with a custom hook"}}]}
C: tinytest - spiderable - is ready with a custom hook failed: {"groupPath":["tinytest","spiderable"],"test":"is ready with a custom hook","events":[{"sequence":490,"type":"fail","details":{"type":"assert_equal","expected":"2","actual":"0","not":false},"cookie":{"name":"spiderable - is ready with a custom hook","offset":2,"groupPath":["tinytest","spiderable"],"shortName":"is ready with a custom hook"}}]}
C: tinytest - accounts - verify single onReconnect callback failed: {"groupPath":["tinytest","accounts"],"test":"verify single onReconnect callback","events":[{"sequence":564,"type":"fail","details":{"type":"assert_equal","message":"Only one onReconnect callback should be registered","expected":"1","actual":"0","not":false},"cookie":{"name":"accounts - verify single onReconnect callback","offset":1,"groupPath":["tinytest","accounts"],"shortName":"verify single onReconnect callback"}}]}

**Comment by dolgarev** [comment]
> 1. Could you please provide a clearer description of the bug fix? What was the bug?
> 2. The tests is failing:

Hey everyone,

Originally, callbacks were stored in a plain object (without a prototype), effectively acting as an array-like object (but without a .length property). Numeric keys were used to store the functions:

```
45: this.callbacks = Object.create(null); // Callbacks stored here
```

However, in the clear() method, this.callbacks was being reassigned to a regular array:

```
96: clear() {
  this.nextCallbackId = 0;
  this.callbacks = []; // Reassigned to an empty array
}
```

This looks like a potential bug or at least a significant inconsistency. After clear() was called, the this.callbacks property would switch from an object-like structure to a standard array, which could lead to unexpected behavior in subsequent operations.

To address this and modernize the code, I've decided to switch to a Set data structure for storing callbacks. This ensures consistency and leverages a more appropriate structure for collections of unique items.

One important implication of using a Set is that it prevents duplicate callbacks from being added. I'm considering whether this might affect any existing tests or specific functionalities if the system previously relied on being able to register the exact same callback function multiple times.

**Comment by dolgarev** [comment]
I reviewed one of the tests:
```
Tinytest.addAsync(
    'accounts - verify single onReconnect callback',
    (test, done) => {
      loginAsUser1((err) => {
        test.isUndefined(err, 'Unexpected error logging in as user1');
        test.equal(
          Object.keys(DDP._reconnectHook.callbacks).length,
          1,
          'Only one onReconnect callback should be registered'
        );
        Meteor.disconnect();
        test.isFalse(Meteor.status().connected);
        Meteor.reconnect();
        setTimeout(() => {
          test.isTrue(Meteor.status().connected);
          test.equal(
            Object.keys(DDP._reconnectHook.callbacks).length,
            1,
            'Only one onReconnect callback should be registered'
          );
          done();
        }, 1000);
      });
    }
  );
```
`Object.keys(DDP._reconnectHook.callbacks).length`...
The problem is that callback-hooks don't have a standard method to evaluate the number of registered callbacks. Perhaps it makes sense to add a `.size` method for this. Or give up on Set, but that would bring back the problem of duplicates.

**Comment by dolgarev** [comment]
@italojs I've prepared the necessary changes for the package. However, I have a question regarding version numbers.

For example, I've made changes to the tests for the `deprecated/spiderable` package, which currently has version 1.1.0-alpha300.6.

Should its version number be updated if I've only patched the tests and not the core functionality? If so, which version number would be more appropriate: 1.1.0-alpha300.7 or 1.1.1-alpha300.6?

File: packages/callback-hook/hook.js [hunk 1]
```diff
@@ -36,34 +36,75 @@
...
+    return this.callbacks.size;
+  }
+
+  // WARN: This method is for test compatibility only.
+  _asArray() {
+    return Array.from(this.callbacks);
+  }
+
+  // WARN: This method is for test compatibility only.
+  _fromArray(arr) {
```
italojs: if this function have only test proposal, would be better we move it as a function to test file, same for `_asArray`

italojs (changes requested):

italojs:
> Should its version number be updated if I've only patched the tests and not the core functionality? 

since you just updated the test, that is ok

packages/callback-hook/hook.js
```diff
@@ -167,27 +171,62 @@ export class Hook {
...
+ * @param {Function|string} exceptionHandler The exception handler to normalize. Can be a function,
+ *   a string description for logging, or any other value (which defaults to generic logging).
+ * @returns {Function} A function that handles exceptions.
+ */
+function normalizeHookExceptionHandler(exceptionHandler) {
+  if (typeof exceptionHandler === 'function') {
+    return exceptionHandler;
+  }
+
+  // TODO: The message "callback of async function" is not very useful and needs clarification.
```
italojs: could you please address this TODO?

italojs (changes requested):

packages/callback-hook/hook.js
dolgarev: I've renamed these methods: ._asArray to .asArray, and ._fromArray to .fromArray. These methods have proven useful not only for testing but also for debugging.

packages/callback-hook/hook.js
dolgarev: This is a future task. We can address it once all hook use cases are fully clear. At the moment, it's beyond my current capacity. Alternatively, I can remove this TODO for now.

packages/callback-hook/package.js
```diff
@@ -1,6 +1,6 @@
 Package.describe({
   summary: "Register callbacks on a hook",
-  version: '1.6.0',
+  version: '1.7.0',
```
nachocodoner: If this PR refactors and prepares for extensions, it belongs in a patch release (1.6.1) rather than 1.7.0.

dolgarev: OK, I agree.

StorytellerCZ (dismissed):

packages/callback-hook/hook.js
italojs: an option is to create an issue describing the todo and attaching it in the comment (this pr too)

dolgarev: @italojs I just deleted this TODO. I don't even remember the context in which it made sense anymore.

italojs (approved):
The pr looks great, it will be addressed in 3.4.1

PR #13865, opened 2025-07-26
Remove users update .allow rule

harryadel:
As I was running a security checkup for an application about to deployed, it's customary to disable allow/deny rules but I figured since I don't use `insecure` package I thought I don't need to do anything yet I decided to try out running .update operation on the users collection from the client side and it worked!

Turns out Meteor by default adds such rule so before running any application into production you should add such line to your code:
```js
Meteor.users.deny({
  update() { return true; } // Deny all updates
});
```
Default should be secure and safe not a faulty rule-set that you've to retro-patch each time before going live!


italojs:
great PR
could you please update the docs in `v3-docs/docs/api/accounts.md` and create a new one explaining why it's enabled by default and how to disable it if necessary, this one you can write in the getting started `v3-docs/docs/about/web-apps.md`

harryadel:
Which parts specifically are you speaking of?

italojs (approved):
nevermind, i meesed a few things in my mind, after rebase the release 3.3.2 with devel, we will merge it

StorytellerCZ:
This is a breaking change, so we should be careful in announcing it. Hopefully not many people use this functionality anymore, but still.

harryadel:
Yeah, it's breaking but in a good way. Would you not fix a security problem if it means a breaking change? Backwards compatibility takes a backseat when security comes into play.

packages/accounts-base/accounts_server.js
```diff
@@ -1809,30 +1809,6 @@ function defaultValidateNewUserHook(user) {
...
-
-      // user can only modify the 'profile' field. sets to multiple
-      // sub-keys (eg profile.foo and profile.bar) are merged into entry
-      // in the fields list.
-      if (fields.length !== 1 || fields[0] !== 'profile') {
-        return false;
-      }
-
-      return true;
-    },
```
nachocodoner, line 1808: > I decided to try out running .update operation on the users collection from the client side and it worked!

Was the .update operation applied to your own user?

Under the allow rule, you can update only your own user, not others, which seems fine, does it?

I don't see a security concern modifying your data from your own session. And only the user profile data according to the stripped code.

harryadel: Yes, it was my own record but still. I beg to differ, the default way shouldn't be leaving it open. 

Ideally, client-side changes should be triggered by the server (i.e methods) and if the developer wishes to allow the users to do so, he needs to do it *explicitly* not forcibly shipping it into every Meteor project.

nachocodoner: Ok, I see what you mean. Set default controls on what a user can do against the DB, maybe they can’t change anything or can only do so via specific methods.

It’s fine, but it might help to add a test that verifies users can’t modify their data even when we rely on a default deny. Do we have such a test? We should update the existing tests to reflect this expectation. If we later change the default on deny operations, we can quickly see the test complaining and the expectation intact.

harryadel: >Set default controls on what a user can do against the DB, maybe they can’t change anything or can only do so via specific methods.

It's not about setting 'special' defaults really, it's rather about the absence of it. There shouldn't be any rules that allow custom/special handling of user data beyond what the developer chooses and sets.

 >It’s fine, but it might help to add a test that verifies users can’t modify their data even when we rely on a default deny. Do we have such a test?
 
 Ehh, I'm a bit on edge regarding this. Why should we test something that doesn't exist?! If it's about later on, well when we do add another custom deny/allow rules we should add it.

nachocodoner: > Ehh, I'm a bit on edge regarding this. Why should we test something that doesn't exist?

Can't you use update methods with User collections? Any collection supports that, right? It does exist, even if by default it triggers an error.

My point is that if we don’t expect users to ever modify their own data or others, we should handle it. From a design and security perspective, a user collection, the core and critical sensitive part of any Meteor app , should never allow users to update its or others data. Just restore the test you removed to assert on not allowed modifications.

I don’t have a strong opinion on this, I just think it’s good to have. If anything changes at the allow/deny abstraction level (not neccesarely on the User collection allow/deny level), this test will warn us. And as expected, no one should be able to modify their own or another user’s data.

harryadel:
@nachocodoner Was this included in 3.3.2?
Same goes for https://github.com/meteor/meteor/pull/13843 too

nachocodoner:
Sorry @harryadel. Those PRs weren’t included in 3.3.2 apparently. When 3.3.2 was merged into development and removed, system closed them automatically.

Could you reopen them to point to devel so they’re included in the next release?

italojs:
It was done automatically after we merged/deleted the release-3.3.2, sorry for that @harryadel

Issue #13864, opened 2025-07-24
insertAsync in Minimongo local_collection.js is buggy and not concurrency-safe


graemian:
This method iterates over `this.queries` from a static list of keys (`Object.keys(this.queries)`):

```
for (const qid of Object.keys(this.queries)) {
      const query = this.queries[qid];

      if (query.dirty) {
        continue;
      }

      // keys are removed from this.queries either synchronously here or by a race condition when execution pauses due to await

}
```

However, keys are removed from `this.queries` either directly in the body of the loop or elsewhere (when the code defers due to `await`), causing `query` to sometimes be `undefined`

I will work on a simple reproduction and hopefully post shortly.

This happens intermittently in my Cordova iOS and Android code. I'm pretty sure I've seen it on web as well. See https://forums.meteor.com/t/minimongo-transpile-bug-with-meteor-3-3-on-ios-and-android/63787

Furthermore, I suspect calls to `insertAsync` (and others?) should be queued. Otherwise multiple invocations can trip over each other when they modify `this.queries` 

This happens on Meteor 3.3. I never had this problem on Meteor 2.16, but that's probably because I wasn't using `insertAsync` yet.

graemian:
Here's an example from my debugger: 

`qid` is 763, but there's no key 763 in `this.queries`

This causes `query` to be `undefined`, so `query.dirty` will throw an exception. 


<img width="573" height="382" alt="Image" src="https://github.com/user-attachments/assets/9ea07ae1-c8b1-4b58-88b4-97dd0b74acd3" />

graemian:
I can confirm that by queuing calls to `updateAsync` that my bug is fixed.

This works by eliminating the race condition related to updating to `this.queries`. 

Here are my mods:

```
export default class LocalCollection {
  constructor(name) {

    ...

    this.dataModQueue = Promise.resolve();

  }

  async queueDataMod(func) {

    this.dataModQueue = this.dataModQueue.then(async () => {
      try {
        return await func();
      } catch (error) {
        console.error("Error in queued data mod:", error);
      }
    });

    return this.dataModQueue;

  }

  async updateAsync(selector, mod, options, callback) {

    return this.queueDataMod(

      async () => await this.updateAsyncUnsafe(selector, mod, options, callback)

    );

  }

  // XXX atomicity: if multi is true, and one modification fails, do
  // we rollback the whole operation, or what?
  async updateAsyncUnsafe(selector, mod, options, callback) {

    ...
```

[...]

PR #13868, opened 2025-07-26
Enable modern browser support for Cordova unless explicitly disabled

graemian:
Modern mobile devices have qualified as "modern browsers" for a long while now, so we can drop the transpiling we needed to do in the early days.

Apps built to older API specifications do not meet the requirements to be published in the app stores anymore anyway.

This new behavior can be disabled by setting the environment variable METEOR_CORDOVA_DISABLE_MODERN_BROWSERS to Y.

This was discussed with @italojs in this issue https://github.com/meteor/meteor/issues/13864#issuecomment-3119736349

### DIFF
--- a/packages/babel-compiler/babel-compiler.js
+++ b/packages/babel-compiler/babel-compiler.js
@@ -249,6 +249,11 @@ BCp.processOneFileForTarget = function (inputFile, source) {
       features.nodeMajorVersion = parseInt(process.versions.node, 10);
     } else if (arch === "web.browser") {
       features.modernBrowsers = true;
+    } else if (arch === "web.cordova") {
+
+      if (process.env["METEOR_CORDOVA_DISABLE_MODERN_BROWSERS"] != "Y")
+        features.modernBrowsers = true;
+
     }
 
     features.topLevelAwait = inputFile.supportsTopLevelAwait &&
--- a/v3-docs/docs/about/cordova.md
+++ b/v3-docs/docs/about/cordova.md
@@ -261,3 +261,13 @@ After building your Cordova project with Meteor, you can use **Android Studio** 
 5. Go to **Product > Archive** to create an archive of your app
 6. In the **Organizer** window, click **Distribute App** and follow the prompts to configure signing and export the IPA file.
 7. Upload the IPA file to the App Store or distribute via TestFlight.
+
+# Legacy device support
+
+Meteor distinguishes between legacy and modern browsers - see the [modern browsers package](../packages/modern-browsers). Web apps include different code bundles for each, but Cordova apps only have a single code bundle. From Meteor 3.4 onwards, the default code bundle changed from legacy to modern.
+
+You can force Meteor to use the legacy browser code bundle by setting the environment variable `METEOR_CORDOVA_DISABLE_MODERN_BROWSERS` to `Y` when running or building your app. For example:
+
+```METEOR_CORDOVA_DISABLE_MODERN_BROWSERS=Y meteor run ios```
+
+Both the App Store and Google Play will only publish new and updated apps for a certain minimum mobile OS version. As of 2025, these minimum OS versions support the  modern browser code bundle. 


packages/babel-compiler/babel-compiler.js
```diff
@@ -249,6 +249,11 @@ BCp.processOneFileForTarget = function (inputFile, source) {
       features.nodeMajorVersion = parseInt(process.versions.node, 10);
     } else if (arch === "web.browser") {
       features.modernBrowsers = true;
+    } else if (arch === "web.cordova") {
+
+      if (process.env["METEOR_CORDOVA_DISABLE_MODERN_BROWSERS"] != "Y")
```
italojs, line 254: To improve our DX, it would be better if we get this info from [mobile-config.js](https://guide.meteor.com/cordova)
Today we have a utility function to get Meteor pkg json config, we can have a similar function for Cordova 

This [PR](https://github.com/meteor/meteor/pull/13743/files) introduces a getMeteorConfig function to release 3.3.1

[MeteorConfig setup](https://github.com/meteor/meteor/blob/4341fbbd8f8c6d3ff471009a17e5e5e2aae7cfbc/tools/tool-env/meteor-config.js#L50)
[Meteor config usage inside babel-compiler.js](https://github.com/meteor/meteor/blob/4341fbbd8f8c6d3ff471009a17e5e5e2aae7cfbc/packages/babel-compiler/babel-compiler.js#L28)



I recommend you rebase(do not merge) your branch with release-3.3.1 for 2 reasons:
1.  Have access to the [PR's](https://github.com/meteor/meteor/pull/13743/files) code
2. Your change will be released into 3.3.2 only

italojs: still about DX, It would be better to use truefy instead of a string 'Y'.
for example:

```
features.modernBrowsers = !process.env["METEOR_CORDOVA_DISABLE_MODERN_BROWSERS"]

italojs (changes requested):

packages/babel-compiler/babel-compiler.js
graemian: Let me just make sure I understand correctly: rather than use an environment variable, you are suggesting we read it out of mobile-config.js somehow?

If so, I don't see an existing configuration mechanism we can tap into: `App.setPreference` seems to be Cordova-tool specific, so it doesn't seem appropriate. I don't see anything else, but maybe I missed something?

Currently Meteor Cordova uses environment variables to control build-time configuration, like `METEOR_CORDOVA_COMPAT_VERSION_*`, so that's why I ended up choosing the environment variable. 

I'm happy to change it though, but it's going to be kind of weird to have some stuff in environment variables and other stuff in another configuration mechanism?

italojs: no, I'm talking to have the config only inside a config file avoingin envs.

This "config" could come from `package.json`, [settings.json](https://docs.meteor.com/tutorials/react/8.deploying.html#_8-3-set-up-settings) or [mobile-config.js](https://guide.meteor.com/cordova#cordova-integration-in-meteor), IMO `mobile-config.js` is better, but we can get if from pkg.json as well


Below you can find a function that reads the `pkg.json` from the project before building it. The function can be found [here](https://github.com/meteor/meteor/blob/release-3.3.1/tools/tool-env/meteor-config.js)

If you try to get this info from `mobile-config.js`:
     > edit that function to read that file if it exists
If you want to get that flag from pkg.json
    > do not modify the function since it aready reads the `pkg.json`, just modify your PR's code `features.modernBrowsers = getMeteorConfig().cordova.compartVersion

``` js
export function initMeteorConfig(appDir) {
  const modernForced = JSON.parse(process.env.METEOR_MODERN || "false");
  let packageJson;
  if (appDir) {
    const packageJsonPath = files.pathJoin(appDir, 'package.json');
    if (!files.exists(packageJsonPath)) {
      setMeteorConfig({
        modern: normalizeModernConfig(modernForced || false),
      });
      return meteorConfig;
    }
    const packageJsonFile = files.readFile(packageJsonPath, 'utf8');
    packageJson = JSON.parse(packageJsonFile);
  }
  setMeteorConfig({
    ...(packageJson?.meteor || {}),
    modern: normalizeModernConfig(modernForced || packageJson?.meteor?.modern || false),
  });
  return meteorConfig;
}

graemian: `package.json` and `settings.json` are both JSON files, so they are easy to read and parse for config.

`mobile-config.js` is a Javascript file, so what would you do with it? Eval it? Then inside the file set `global.cordovaDisableModernBrowsers = true` and pick that up from `babel-compiler.js`?

Perhaps it could be imported rather? 

Have a look at https://v2-docs.meteor.com/api/mobile-config.html . All the config is done via the `App` object and seems to be passed directly into a Cordova config file.

Unless you can think of an idiomatic way to get the config from `mobile-config.js`, perhaps it's best to put it in `package.json`?

graemian: I thought of adding `App.disableModernBrowsersBuild()` to `mobile-config.js`. This would set a global variable which I could pick up in `babel-compiler.js`. However, it turns out that `mobile-config.js` is only called after all the babel compiling is done.

So maybe it's best to get it from `package.json` then?

italojs: nice! You are right, `package.json` is the best place

graemian: Shucks, I think I've screwed up this PR by merging devel into the branch by pressing sync on github.com by accident

I can't really figure out how to fix it, so I submitted another PR 13889 https://github.com/meteor/meteor/pull/13889

I think my local branch is a bit busted too. I did a few rebases, but I really didn't know what I was doing 🙈

italojs:
merged at https://github.com/meteor/meteor/pull/13896

Issue #13701, opened 2025-04-17
if an *async* observeChangesAsync callback throws an error, the app exits


perbergland:
Meteor 3.2

There is something weird in how errors in observeChanges callbacks are handled. We noticed that our pods sometimes exited spontaneously and found some instances where the stack trace was this:

```
meteor://💻app/packages/ddp-server.js:2263

 throw new Error("Could not find element with id ".concat(id, " to change"));

 ^

Error: Could not find element with id RShhwsLKKuwFgxawJ to change

 at SessionCollectionView.changed (packages/ddp-server/session_collection_view.ts:103:13)

 at Session.changed (packages/ddp-server/livedata_server.js:265:12)

 at Subscription.changed (packages/ddp-server/livedata_server.js:1191:19)

 at refreshAsync (server/collections/logs/logs-method-registrations.ts:121:24)

 at processTicksAndRejections (node:internal/process/task_queues:105:5)
```

The offending code was this call:

https://github.com/meteor/meteor/blob/d185a3487d5390f09e06e5520bad20cecd732968/packages/mongo/observe_multiplex.ts#L161

which invoked our changed callback that threw an error via a subscription.changed call.
Somehow, that error is not caught properly  and we get an unhandled top level exception which by default causes a process.exit().

I’m not sure if the problem lies in the implementation of AsynchronousQueue or in the usage within ObserveHandle - I find the code hard to read, but it’s very easy for us to trigger a catastrophic error by throwing an error in the "changed" callback sent into observeChangesAsync (server-side).

```
meteor://💻app/app/app.js:76497
              throw new Error("asdfasdf");
                    ^

Error: asdfasdf
    at refreshAsync (server/collections/logs/logs-method-registrations.ts:122:17)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)

Node.js v22.14.0
=> Exited with code: 1
```

perbergland:
To add to this, I find this code strange since it implies that the callback would return a method but the documentation says that it should return void (not even Promise<void>)

https://github.com/meteor/meteor/blob/d185a3487d5390f09e06e5520bad20cecd732968/packages/mongo/observe_multiplex.ts#L161

perbergland:
I was able to reproduce this easily in the "full" skeleton app with Meteor 3.2:

*  meteor create --full meteor-full-skeleton
* cd meteor-full-skeleton
* open imports/api/links/server/publication.js

[...]

PR #13871, opened 2025-07-28
FIX: observeChanges[Async] should not crash the process on error

Grubba27:
relates to: https://github.com/meteor/meteor/issues/13701

Firstly, I'm adding more tests and then checking what could be the culprit for this behavior

### DIFF
--- a/packages/mongo/observe_multiplex.ts
+++ b/packages/mongo/observe_multiplex.ts
@@ -1,12 +1,17 @@
-import isEmpty from 'lodash.isempty';
-import { ObserveHandle } from './observe_handle';
+import isEmpty from "lodash.isempty";
+import { ObserveHandle } from "./observe_handle";
 
 interface ObserveMultiplexerOptions {
   ordered: boolean;
   onStop?: () => void;
 }
 
-export type ObserveHandleCallback = 'added' | 'addedBefore' | 'changed' | 'movedBefore' | 'removed';
+export type ObserveHandleCallback =
+  | "added"
+  | "addedBefore"
+  | "changed"
+  | "movedBefore"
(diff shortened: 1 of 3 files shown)


Grubba27:
It works as expected, but not quite in the ordering:

https://youtu.be/bagb9sEZLC0

packages/mongo/observe_multiplex.ts
```diff
@@ -158,10 +187,20 @@ export class ObserveMultiplexer {
 
         if (!callback) continue;
 
-        handle.initialAddsSent.then(callback.apply(
+        const p = callback.apply(
```
italojs: Could we have a more declarative name?

packages/mongo/tests/mongo_livedata_tests.js
```diff
@@ -4488,6 +4488,48 @@ testAsyncMulti(
...
+      );
+
+      return new Promise(async (resolve) => {
+        await Collection.find({}).observeChangesAsync({
+          async added(_id, fields) {
+            throw new Error('Test error in observeChangesAsync');
+          },
+        });
+        await Collection.insertAsync({ foo: { bar: 123 } });
+        test.equal(1,1); // ensure process did not crash
```
italojs: maybe `test.throws(await Collection.insertAsync({ foo: { bar: 123 } }), undefined)` is better 
give a look at https://packosphere.com/meteor/tinytest in `Assertions without optional fail messages
` session

italojs (changes requested):

nachocodoner (approved):

Grubba27:
Noted thing with Windows

PR #13872, opened 2025-07-28
Logout with removal of all sessions

StorytellerCZ:
Added `logoutAllClients`, which logs out the current user and removes and resume tokens as well. This is to close any other open sessions.

Added a test, but I don't think it runs properly and would appreciate help with it.

### DIFF
--- a/packages/accounts-base/accounts_client.js
+++ b/packages/accounts-base/accounts_client.js
@@ -151,6 +151,30 @@ export class AccountsClient extends AccountsCommon {
   }
 
   /**
+   * @summary Log out all clients logged in as the current user and logs the current user out as well.
+   * @locus Client
+   * @param {Function} [callback] Optional callback. Called with no arguments on success, or with a single `Error` argument on failure.
+   */
+  logoutAllClients(callback) {
+    this._loggingOut.set(true);
+
+    this.connection.applyAsync('logoutAllClients', [], {
+      // TODO[FIBERS]: Look this { wait: true } later.
+      wait: true
+    })
+      .then((result) => {
+        this._loggingOut.set(false);
+        this._loginCallbacksCalled = false;
+        this.makeClientLoggedOut();
+        callback && callback();
+      })
+      .catch((e) => {
+        this._loggingOut.set(false);
+        callback && callback(e);
+      });
+  }
+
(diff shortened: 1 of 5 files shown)


packages/accounts-base/accounts_client.js
```diff
@@ -150,6 +150,30 @@ export class AccountsClient extends AccountsCommon {
...
+  /**
+   * @summary Log out all clients logged in as the current user and logs the current user out as well.
+   * @locus Client
+   * @param {Function} [callback] Optional callback. Called with no arguments on success, or with a single `Error` argument on failure.
+   */
+  logoutAllClients(callback) {
+    this._loggingOut.set(true);
+
+    this.connection.applyAsync('logoutAllClients', [], {
+      // TODO[FIBERS]: Look this { wait: true } later.
```
italojs, line 162: since Meteor 2.x is in [mantainece mode(eol december 2025)](https://forums.meteor.com/t/meteor-supported-versions-clarification-transition-timeline/63646), i'm not sure if we should add Fibers support here

italojs (changes requested):
for sure a super useful function
could you please add a doc session too?

packages/accounts-base/accounts_client.js
StorytellerCZ: This started from the logout function, so these can be updated.

As for the docs I have placed the API box there.

italojs:
Just another question, did you consider reusing logoutOtherClients inside logoutAllClients?

packages/accounts-base/accounts_client.js
italojs: Yes, but I was talking about the v3-docs. I mean, it would be nice to instruct the developers on how to use it.

StorytellerCZ:
Yes, this is actually merger of that code and logout code. The problem with `logoutAllClients` is that it does extra work to keep the current session, which is not a goal here. The most important re-use of the code is this:
```
accounts._clearAllLoginTokens(logoutUserId);
```

italojs:
LGTM, just make the CI green, then we can merge it
moved to [release-3.4](https://github.com/meteor/meteor/tree/release-3.4) since it's a new feature

PR #13879, opened 2025-08-01
FIX: CI issue with vue skeleton and vite

Grubba27:
Just moving deps and bumping them


italojs (dismissed):

italojs:
The vite-changes broke the ci

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'meteor-vite' imported from /tmp/mt-y0jx74.8xhd/home/vue/__meteor-vite-runtime-bootstrap__.ts\x1B[39m Failed to resolve package.json contents. If you're using a custom Vite config file path, either use the default one or try to see why the package.json file is unavailable."

**Comment by nachocodoner** [comment]
@JorgenVatle: Any input on this sudden failure? It seems to be related with meteor-vite.

**Comment by JorgenVatle** [comment]
Huh, interesting. I don't believe we've shipped any new releases after my skeleton app PR last month.

I'd be happy to have a closer look at this as soon as I have the chance in few days time give or take. Or is it urgent?

File: tools/static-assets/skel-vue/package.json [hunk 1]
```diff
@@ -11,7 +11,9 @@
   "dependencies": {
     "@babel/runtime": "^7.23.5",
     "meteor-node-stubs": "^1.2.12",
+    "meteor-vite": "^3.8.0",
     "vue": "^3.3.9",
+    "vite": "^6.3.5",
```
italojs, line 17: should be a dev dependency, no?
Same for "meteor-vite"

italojs:
we should merge it into release 3.3.1 or 3.3.2 since devel uses [external skeletons](https://github.com/meteor/skel-vue)

tools/static-assets/skel-vue/package.json
JorgenVatle: No, it should be a production dependency as it reads and exposes assets within the client manifest through Meteor's built-in web server during server startup.

JorgenVatle: Vite itself can be left as a dev dependency i believe though 👍

JorgenVatle: I believe the upcoming release addresses this issue so it won't matter where the dependency is placed.

JorgenVatle:
I'll try to go through this on Monday if it keeps failing. 👍

nachocodoner:
The issue was fixed already @JorgenVatle. It was due to having skeleton defined remotely as well and having to act on the fix on the local and external skel repos apparently.

@Grubba27: can we close this, right?

PR #13880, opened 2025-08-01
Modernize vue skeleton

Grubba27:


italojs:
Looks like there is a conflict between this branch and https://github.com/meteor/meteor/pull/13879 and https://github.com/meteor/meteor/pull/13881

PR #13881, opened 2025-08-01
Bump deps in vue skel modernization

Grubba27:


italojs:
Looks like there is a conflict between this branch and [this one](https://github.com/meteor/meteor/pull/13879) and https://github.com/meteor/meteor/pull/13880

PR #13887, opened 2025-08-06
update lodash dependency in boilerplate-generator 

italojs:
TLDR

- [lodash.template](https://www.npmjs.com/package/lodash.template) is deprecated
- Discussed here about do not bring lodash to the core https://github.com/meteor/meteor/pull/13617
- this PR is similar to https://github.com/meteor/meteor/pull/12813 from @StorytellerCZ

### DIFF
--- a/packages/boilerplate-generator/package.js
+++ b/packages/boilerplate-generator/package.js
@@ -5,7 +5,7 @@ Package.describe({
 
 Npm.depends({
   "combined-stream2": "1.1.2",
-  "lodash.template": "4.5.0"
+  "lodash": "4.17.21"
 });
 
 Package.onUse(api => {
--- a/packages/boilerplate-generator/template.js
+++ b/packages/boilerplate-generator/template.js
@@ -1,4 +1,4 @@
-import lodashTemplate from 'lodash.template';
+import { template as lodashTemplate } from 'lodash';
 
 // As identified in issue #9149, when an application overrides the default
 // _.template settings using _.templateSettings, those new settings are
--- a/packages/accounts-base/accounts_server.js
+++ b/packages/accounts-base/accounts_server.js
@@ -333,6 +333,30 @@ export class AccountsServer extends AccountsCommon {
     return user;
   }
 
+  /**
+   * @summary Find a user by one of their email addresses.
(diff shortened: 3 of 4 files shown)


packages/boilerplate-generator/package.js
```diff
@@ -5,7 +5,7 @@ Package.describe({
 
 Npm.depends({
   "combined-stream2": "1.1.2",
-  "lodash.template": "4.5.0"
+  "lodash": "4.17.21"
```
Grubba27, line 8: We could change to use https://socket.dev/npm/package/eta

packages/boilerplate-generator/template.js
```diff
@@ -1,4 +1,4 @@
-import lodashTemplate from 'lodash.template';
+import { template as lodashTemplate } from 'lodash';
```
Grubba27, line 2: Is this correct? maybe we should `import { _ }` and user `_.template`

italojs: no because the _ is the full lodash as we can see in the readme
```
var _ = require('lodash');
```

since we will work only with template, i'm using a descrontructor ` var { template} = require('lodash')`

packages/boilerplate-generator/package.js
italojs: I'm not sure, I prefere to not add more 1 package to do the same as lodash.template, I mean, it will increase the CVE updates that we will need to do in the future, mantain the pkg we already have in the project is better IMO

Grubba27 (approved):

nachocodoner (changes requested):
I think this goes against keeping the bundle thin. Adding the full lodash package to dependencies of just one package, knowing the Meteor bundler doesn't support treeshaking, will make the builds giant.

There’s been work from @harryadel for a long time to make the Meteor core thinner by actually removing those helpers libs like underscore to a local function or package alternative. This would be a regression.

I remember the lodash.template deprecation; we may need to maintain our own fork, taking just the helper from the lodash package, or find an alternative.

nachocodoner:
This was actually been discussed and covered here as well, https://github.com/meteor/meteor/pull/13617

PR #13889, opened 2025-08-07
Retry: Enable modern browser support for Cordova unless explicitly disabled

graemian:
This is a resubmit of PR 13868 https://github.com/meteor/meteor/pull/13868 which I think I broke by merging devel into the branch by pressing sync on github.com by accident


StorytellerCZ:
@nachocodoner I think this did not closed properly.

nachocodoner:
I'm not sure why this closed. Maybe when merging 3.3.1? I would have expected it would automatically compare with devel.

@graemian: You would have to re-open it and compare with devel. Actually forks should compare to devel so that we can add to versions properly later.

graemian:
Done, see https://github.com/meteor/meteor/pull/13896

PR #13893, opened 2025-08-10
Updated parameter oders setting option = {}  and corresponding docs. #13873

### LINKED ISSUE #13873: Calling Roles.getUsersInRoleAsync with queryOptions fields throws error
In the [documentation](https://docs.meteor.com/packages/roles.html#Roles-getUsersInRoleAsync) for `Roles.getUsersInRoleAsync` it says it takes two arguments `roles` and `options`. And that `options` can contain `queryOptions` for the actual Mongo query.

Upon calling it like this per the docs:
```
const role = 'some-role';
const fields = {
   createdAt: 1,
   firstName: 1
};
const queryOptions = {
   fields
};
const options = {
   queryOptions
}; 

const users = await (await Roles.getUsersInRoleAsync(role, options)).fetchAsync();
```
The call to [await Roles.getUserAssignmentsForRole(roles, options).fetchAsync()](https://github.com/meteor/meteor/blob/devel/packages/roles/roles_common_async.js#L1056) within `Roles.getUsersInRoleAsync` throws an error:
```
const ids = (await Roles.getUserAssignmentsForRole(roles, options).fetchAsync()).map(a => a.user._id);
TypeError: Cannot read properties of undefined (reading '_id')
   at packages/roles/roles_common_async.js:1057:25
```
However if calling the function like this per its [function signature](https://github.com/meteor/meteor/blob/devel/packages/roles/roles_common_async.js#L1054) with `queryOptions` as the third argument it works as expected:
```
const users = await (await Roles.getUsersInRoleAsync(role, null, queryOptions)).fetchAsync();
```

So either the documentation for `Roles.getUsersInRoleAsync` needs to be updated or the code has a bug.

sidd190:
This fixes issue #13873 . 
I've fixed the parameter oders setting option = {} by default and documented it.

### DIFF
--- a/packages/roles/roles_common_async.js
+++ b/packages/roles/roles_common_async.js
@@ -1052,14 +1052,15 @@ Object.assign(Roles, {
    * @return {Promise<Cursor>} Cursor of users in roles.
    */
   getUsersInRoleAsync: async function (roles, options = {}, queryOptions) {
+    if (!options.queryOptions) options.queryOptions = queryOptions;
     const ids = (
       await Roles.getUserAssignmentsForRole(roles, options).fetchAsync()
     ).map((a) => a.user._id)
 
     return Meteor.users.find(
       { _id: { $in: ids } },
-      (options.queryOptions) || queryOptions || {}
-    )
+      options
+    );
   },
 
   /**
--- a/packages/roles/roles_common_async.js
+++ b/packages/roles/roles_common_async.js
@@ -1051,14 +1051,14 @@ Object.assign(Roles, {
    *   - `queryOptions`: options which are passed directly through to `Meteor.users.find(query, options)`
    * @return {Promise<Cursor>} Cursor of users in roles.
    */
-  getUsersInRoleAsync: async function (roles, options, queryOptions) {
+  getUsersInRoleAsync: async function (roles, options = {}, queryOptions) {
     const ids = (
       await Roles.getUserAssignmentsForRole(roles, options).fetchAsync()
     ).map((a) => a.user._id)
 
     return Meteor.users.find(
       { _id: { $in: ids } },
-      (options && options.queryOptions) || queryOptions || {}
+      (options.queryOptions) || queryOptions || {}
     )
   },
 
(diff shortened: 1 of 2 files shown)


packages/roles/roles_common_async.js
```diff
@@ -1051,14 +1051,14 @@ Object.assign(Roles, {
    *   - `queryOptions`: options which are passed directly through to `Meteor.users.find(query, options)`
    * @return {Promise<Cursor>} Cursor of users in roles.
    */
-  getUsersInRoleAsync: async function (roles, options, queryOptions) {
+  getUsersInRoleAsync: async function (roles, options = {}, queryOptions) {
```
italojs, line 1054: since we must to mantain the query option to avoid break someone, I would to reccommend we address queryOptions inside options, so it cane be used for Roles.getUserAssignmentsForRole as well

```  js
getUsersInRoleAsync: async function (roles, options = {}, queryOptions) {
if(!options.queryOptions) options.queryOptions = queryOptions

// ...

return Meteor.users.find(
      { _id: { $in: ids } },
      options // here we dont need to set ` || {} ` since options already is an obj
    )
```

italojs (changes requested):

sidd190:
I've fixed it up and made the changes now. Kindly review.

italojs:
The tests are failing, but we can't see what's happening via Travis output. Could you please ensure the tests are passing via `bash  ./packages/test-in-console/run.sh`?
more details about tests at [DEVELOPMENT.MD](https://github.com/meteor/meteor/blob/devel/DEVELOPMENT.md)

italojs:
This PR is still valid, but was closed automatically when we deleted the branch release-3.3.2
Could you please reopen this pr pointing to devel?

italojs:
GH blocked the PR from being reopened, then we created the same @sidd190 's PR again
Auto-closed -> https://github.com/meteor/meteor/pull/13893
Reopened -> https://github.com/meteor/meteor/pull/13918

PR #13894, opened 2025-08-11
Create admin role and assign first user to admin

StorytellerCZ:
Following from here: https://forums.meteor.com/t/accounts-improvement-initiative/63792/3?u=storyteller

On my [stream](https://www.youtube.com/watch?v=4qcZHQoOB8I) I implemented the following:

* Added `role` and `scope` to user creation options that will add the newly created user into that role.
* Added `settings.json` settings for roles package: `adminRole` and `addFirstUserAsAdmin`
* If `adminRole` is set it will create the admin role to the name that is set.
* If in addition `addFirstUserAsAdmin` is set as well, then the first created user will be assigned that role.

Quality of life improvements on the way:
* Passwordless DB query options changed `fields` to `projection`
* Remove duplicate function in hooks package

**TODO:**
- [x] Documentation
- [ ] Tests
- [ ] Make sure that OAuth user creation makes this happen as well.


Grubba27 (dismissed):
Solid feature!

StorytellerCZ:
@italojs I think that this will be for v3.5 as I don't think I will finish it in time for v3.4.1. So just heads up.

nachocodoner:
I'm going to mark this one for 3.x.x. Once this is in a verified green state, we'll know which release to add it to. I'm not marking it for 3.5 because, at the moment, this branch includes the changes from 3.4.1, and 3.5 still doesn't have them. I want to avoid it accidentally getting into 3.5 earlier with the 3.4.1 state.

Anyway, let us know when you have updates on it, and we'll re-evaluate where to include it.

PR #13895, opened 2025-08-11
Add query selector validation to catch MongoDB errors

italojs:
fix for https://forums.meteor.com/t/meteor-3-3-1-geointersects-inside-publication-throws-error/63842

The main changes ensure that query errors are thrown correctly in the client and the server.

**Error handling improvements:**

* In `mongo_connection.js`, updated the error handling logic so that `MiniMongoQueryError` is only re-thrown on the client, preventing unrelated exceptions from being ignored on the server.

**Test improvements:**

The error `$in needs an array` should come from Mongo since Mongo itself doesn't support that operation, it wasn't being thrown via observeChanges

Note: `MiniMongo` is already covered by `minimongo - selector_compiler` test.

### DIFF
--- a/packages/mongo/tests/mongo_livedata_tests.js
+++ b/packages/mongo/tests/mongo_livedata_tests.js
@@ -4535,7 +4535,8 @@ if(Meteor.isServer) {
   Meteor.publish('testGeoPolygon', function(viewport) {
     check(viewport, Match.ObjectIncluding({ bounds: geoPolygonSchema }));
     // return an empty collection for the test
-    return new Mongo.Collection('test').find();
+    const randomId = Random.id();
+    return new Mongo.Collection(randomId).find();
   });
 }
 
--- a/packages/mongo/tests/mongo_livedata_tests.js
+++ b/packages/mongo/tests/mongo_livedata_tests.js
@@ -4540,9 +4540,61 @@ if(Meteor.isServer) {
   });
 }
 
+if (Meteor.isServer) {
+  Meteor.publish('testGeoIntersects', function(viewport) {
+    check(viewport, Match.ObjectIncluding({ bounds: geoPolygonSchema }));
+    const Features = new Mongo.Collection('Features_' + this.connection.id);
+    return Features.find({
+      hull: { $geoIntersects: { $geometry: viewport.bounds } },
+    });
+  });
+}
+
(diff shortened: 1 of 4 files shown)


nachocodoner (commented):
Would it be difficult to add a test to catch regressions from using a non-compatible $geoIntersects selector on minimongo, so it runs on the server but not on the client as not supported there?

I think we have a test to catch errors on mongo in general not supporting, but not the case causing the regression.

italojs:
will do

packages/mongo/tests/mongo_livedata_tests.js
```diff
@@ -4516,3 +4516,51 @@ Tinytest.addAsync(
     }
   },
 );
+
+if (Meteor.isServer) {
```
nachocodoner: So Meteor.isServer wraps everything, the publish and subscribe attempt on the test?

It won't perform the subscribe on client, I think. All the Meteor.isClient part subscribe must be put outside Meteor.isServer to be properly executed.

packages/mongo/tests/mongo_livedata_tests.js
```diff
@@ -4516,3 +4516,57 @@ Tinytest.addAsync(
...
+      )
+    )
+  )
+};
+
+if(Meteor.isServer) {
+  Meteor.publish('testGeoPolygon', function(viewport) {
+    check(viewport, Match.ObjectIncluding({ bounds: geoPolygonSchema }));
+    // return an empty collection for the test
+    return new Mongo.Collection('test').find();
```
nachocodoner: I'm not sure we're really testing the regression. Shouldn't the test use the $geoIntersects field directly in the find selector?

Check report: https://forums.meteor.com/t/meteor-3-3-1-geointersects-inside-publication-throws-error/63842

> return Features.find({
>		'hull': {$geoIntersects: {$geometry: viewport.bounds}},
>	});

Also, I won't recommend create the test collection on-fly.

nachocodoner:
It seems tests are still failing, may this be the test doesn't pass. Did you check locally?

italojs:
> It seems tests are still failing, may this be because the test doesn't pass. Did you check locally?

Yeah, but what I missed was that I was executing only the filtered `meteor package-test`, but not the full `run.sh` tests, already got the problem (collection name).

nachocodoner (approved):

PR #13898, opened 2025-08-13
remove lodash.template dependency

italojs:
TLDR

- [lodash.template](https://www.npmjs.com/package/lodash.template) is deprecated
- Discussed here about do not bring lodash to the core https://github.com/meteor/meteor/pull/13617
- this PR is similar to https://github.com/meteor/meteor/pull/12813 from @StorytellerCZ

### DIFF
--- a/packages/boilerplate-generator/template.js
+++ b/packages/boilerplate-generator/template.js
@@ -41,9 +41,7 @@ function escapeStringChar(match) { return '\\' + escapes[match]; }
 
 // Basic Object helpers ------------------------------------------------------
 function isObject(value) { return value != null && typeof value === 'object'; }
-function isFunction(value) { return typeof value === 'function'; }
 function toStringSafe(value) { return value == null ? '' : (value + ''); }
-function keys(obj) { return Object.keys(obj); }
 function baseValues(object, props) { return props.map(k => object[k]); }
 
 // assignInWith/customDefaultsAssignIn replicate lodash merge behavior for template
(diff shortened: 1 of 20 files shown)


italojs:
Waitting [this](https://github.com/meteor/meteor/pull/13900) PR to make CI green

StorytellerCZ:
@italojs the tests are green

packages/boilerplate-generator/template.js
```diff
@@ -1,14 +1,200 @@
...
+  return false;
+}
+
+// ---------------------------------------------------------------------------
+// templateSettings (exportable & mutable like lodash)
+// ---------------------------------------------------------------------------
+export const templateSettings = {
+  escape: reEscape,
+  evaluate: reEvaluate,
+  interpolate: /<%=([\s\S]+?)%>/g,
```
nachocodoner: reInterpolate variable already describes this regex.

packages/boilerplate-generator/template.js
```diff
@@ -1,14 +1,200 @@
...
+  imports: { '_': { escape: escapeHtml } },
+  variable: ''
```
nachocodoner: Original template function only passes these settings:

```
evaluate    : /<%([\s\S]+?)%>/g,
interpolate : /<%=([\s\S]+?)%>/g,
escape      : /<%-([\s\S]+?)%>/g,
```

But here we have two more:

```
imports: { '_': { escape: escapeHtml } },
variable: ''
```

Are we are fine to ensure same behavior?, maybe `imports` and `variable` set as above were also the default when using "lodash.template"?

Not sure how much test coverage we have in this area. We may be fine to have these changes.

italojs: We have a few tests called `boilerplate-generator-tests ...`
tested on blaze applications too

nachocodoner (approved):

PR #13899, opened 2025-08-14
Implement redis-oplog Logic Natively to Optimize Oplog Tailing

ericm546:
### Context & Motivation
Many large-scale Meteor applications turn to redis-oplog to solve performance bottlenecks related to high-volume writes and oplog tailing. Our application faced these exact issues: high latency during bulk operations, server instability, and OOM crashes under load.
While redis-oplog was a potential solution, it introduces significant operational complexity: managing a highly-available Redis cluster and ensuring external database writes are also piped to Redis. We sought to achieve the same performance gains without this added architectural dependency.
Our investigation confirmed the bottleneck is the expensive BSON-to-JavaScript transformation of every document in the oplog, even for documents that don't match active publications. redis-oplog solves this by deferring the full document fetch. This PR brings that same core logic directly into Meteor's native oplog driver.

### Proposed Solution
This PR refactors the oplog driver to mirror the two-stage process used by redis-oplog when used with something like oplogtoredis:
1. Initial Fetch (ID-Only): The oplog query is modified to only retrieve the document's _id. This avoids the costly BSON parsing of the full object for every change event.
2. Deferred Document Load: The full document is only fetched from MongoDB if the change is relevant to an active publication. This fetch uses a projection to retrieve only the fields required by those publications.
Advantages
This implementation avoids the processing of the irrelevant oplog entries without requiring a Redis instance. It offers the same performance trade-off: a small increase in latency for a relevant document change (due to the second DB hit) in exchange for a massive reduction in server load and processing during high-volume writes.
By integrating this pattern into meteor, we achieve these benefits while simplifying the application's architecture.
### Observed Results
This change has been running in our production application for a month and has been validated with targeted load tests.
In Production:
* A massive and sustained improvement in CPU usage, 95th percentile Event Loop Latency, and Garbage Collection duration.
* The application is noticeably more responsive and resilient during peak usage times.
In Load Testing:
* Mass Insertion (Publications filtered by ID): An operation that previously made the application unresponsive for ~10 minutes now completes while the application remains fully responsive.
* Mass Insertion (Publications without an ID filter): An operation that previously caused a server crash (OOM) now results in a brief period of high load, but the server remains stable, does not crash, and fully recovers.
Relation to Change Streams
I am aware there is ongoing work to support MongoDB Change Streams but I'm unsure if this would address the same performance issues.

Looking forward to feedback on this approach.

### DIFF
--- a/packages/mongo/mongo_connection.js
+++ b/packages/mongo/mongo_connection.js
@@ -11,6 +11,8 @@ import { ObserveHandle } from './observe_handle';
 import { ObserveMultiplexer } from './observe_multiplex';
 import { OplogObserveDriver } from './oplog_observe_driver';
 import { OPLOG_COLLECTION, OplogHandle } from './oplog_tailing';
+import { NewOplogTailing } from './new-oplog-driver/new_oplog_tailing'
+import { NewOplogObserveDriver } from './new-oplog-driver/new_oplog_observe_driver';
 import { PollingObserveDriver } from './polling_observe_driver';
 
 const FILE_ASSET_SUFFIX = 'Asset';
@@ -86,8 +88,12 @@ export const MongoConnection = function (url, options) {
   }));
 
   if (options.oplogUrl && ! Package['disable-oplog']) {
-    self._oplogHandle = new OplogHandle(options.oplogUrl, self.db.databaseName);
-    self._docFetcher = new DocFetcher(self);
+    if(Meteor.settings?.packages?.mongo?.useNewOplogTailing){
+      self._oplogHandle = new NewOplogTailing(options.oplogUrl, self.db.databaseName);
+    } else {
+      self._oplogHandle = new OplogHandle(options.oplogUrl, self.db.databaseName);
+      self._docFetcher = new DocFetcher(self);
+    }
   }
 
 };
@@ -913,7 +919,17 @@ Object.assign(MongoConnection.prototype, {
         }
       ].every(f => f());  // invoke each function and check if all return true
 
-      var driverClass = canUseOplog ? OplogObserveDriver : PollingObserveDriver;
+      const useNewOplogTailing = Meteor.settings?.packages?.mongo?.useNewOplogTailing;
+      let driverClass;
+      if(canUseOplog && useNewOplogTailing) {
+        driverClass = NewOplogObserveDriver;
+      }
+      else if (canUseOplog) {
+        driverClass = OplogObserveDriver;
+      } else {
+        driverClass = PollingObserveDriver;
+      }
+
       observeDriver = new driverClass({
         cursorDescription: cursorDescription,
         mongoHandle: self,
--- a/packages/mongo/mongo_driver.js
+++ b/packages/mongo/mongo_driver.js
@@ -1,4 +1,5 @@
 import { OplogHandle } from './oplog_tailing';
+import { NewOplogTailing } from './new-oplog-driver/new_oplog_tailing'
 import { MongoConnection } from './mongo_connection';
 import { OplogObserveDriver } from './oplog_observe_driver';
 import { MongoDB } from './mongo_common';
@@ -31,6 +32,7 @@ MongoInternals.NpmModule = new Proxy(MongoDB, {
 });
 
 MongoInternals.OplogHandle = OplogHandle;
+MongoInternals.NewOplogHandle = NewOplogTailing
 
 MongoInternals.Connection = MongoConnection;
 
--- a/packages/mongo/new-oplog-driver/mongo_id_map.ts
+++ b/packages/mongo/new-oplog-driver/mongo_id_map.ts
@@ -1,0 +1,97 @@
+import { MongoID } from 'meteor/mongo-id';
+
+export class MongoIDMap {
+  _internal: Map<any, any>;
+  private _idStringify: any;
+  private _idParse: any;
+  constructor() {
+    this._internal = new Map();
+    this._idStringify = MongoID.idStringify;
+    this._idParse = MongoID.idParse;
+  }
+
+  get(id: any) {
+    const key = this._idStringify(id);
+    return this._internal.get(key);
+  }
+
+  pop(id: any) {
+    const key = this._idStringify(id);
+    const ret = this._internal.get(key);
+    this._internal.delete(key);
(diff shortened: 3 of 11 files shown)


StorytellerCZ:
Looks interesting. Though I'm not so sure about the naming, though it is good enough for now. This will also require entry in the docs.

radekmie:
Have you considered putting it in a package, so people can try it out first? It'd allow for much faster iterations and backward-incompatible changes (if needed).

italojs:
I agree with @radekmie , but I would recommend making these tests pass before moving on to a package, since you are creating a new observer that is a core module from Meteorjs. The Meteor's CI would be a great test suite to ensure it's a safe package.

ericm546:
I did consider making a separate package, but that would add a lot of complexity because it's so tightly integrated with Meteor's internals. For example, redis-oplog had to duplicate some of Meteor's internal code and use some hacks to connect itself to the internals.

In the meantime, it's possible to test it on an app by overriding the mongo package and setting useNewOplogTailing to true. You can just copy the mongo folder into your project's packages directory. For convenience, I've created a zip of that folder here:
[mongo.zip](https://github.com/user-attachments/files/21839103/mongo.zip)

``` JSON
"packages": {
    "mongo": {
      "useNewOplogTailing": true
    }
  }
```

radekmie:
> I did consider making a separate package, but that would add a lot of complexity because it's so tightly integrated with Meteor's internals. For example, redis-oplog had to duplicate some of Meteor's internal code and use some hacks to connect itself to the internals.

That's the exact reason why testing it as a package makes sense 😅

ericm546:
I've just published a temporary package to test it:
https://github.com/ericm546/alternative-oplog

Most of the files are just copied from the Meteor Mongo package, which I can't access from a normal package, and the way I integrated it isn't great. However, it should make it easier to test on a normal project.

I've tested it with my project's e2e tests, but I don't have the same confidence in it as I do in the PR, so please let me know if there are any issues.

```sh
meteor add disable-oplog
meteor add ericmartin989:alternative-oplog
```

nachocodoner:
Great work! I hope we can add soon [an advanced benchmarking scenario in meteor/performance](https://github.com/meteor/performance/pull/8) to compare and validate performance between the new and old oplog drivers. I’ll review the analysis once it’s ready.

@ericm546 Could you bring more attention to these improvements [on the forums](https://forums.meteor.com/) by using that package? That could reach more people interested in testing, especially those with performance issues like the ones you had.

Regarding tests for this PR: I don’t think the core tests enable this driver by default. We should check stability by forcing the core tests to use only this driver. After this you could take care of any broken test if any.

italojs:
Making it as a package will definitely be the fastest way to validate it within the community. However, if you still want to integrate it into the Meteor core (which we are really excited about), we should ensure it is compatible with the original oplog behavior, thereby ensuring the current tests pass for both implementations.

As @nachocodoner mentioned, the tests are enabled for pooling and oplog by default. You should edit the [mongo_connection.js](https://github.com/meteor/meteor/blob/devel/packages/mongo/mongo_connection.js#L916)  to ensure it's using your driver implementation https://github.com/meteor/meteor/blob/576f7485a2050c206abb09e0f689fb57dd86167d/packages/mongo/mongo_connection.js#L916

Another option is to bring those improvements to the current oplog, but I'm afraid of the impact it could bring for the whole community that already is using oplog, since it's a huge change, but it's still an option.

ericm546:
I have now run the full Meteor test suite with the new implementation. This allowed me to find and fix some bugs. I've updated my PR and the ericmartin989:alternative-oplog package.

I could not get the following tests to pass, as they rely on the internal implementation of the oplog driver, which doesn't apply to this new driver:

- mongo-livedata - observe sorted, limited STRING
- mongo-livedata - observe sorted, limited, big initial set STRING
- mongo-livedata - observe sorted, limited MONGO
- mongo-livedata - observe sorted, limited, big initial set MONGO
- mongo-livedata - oplog - entry skipping

I also had to change one line in the "observeChanges - tailable" test. I don't believe the modification I made changed its behavior; I think the test was failing due to flakiness.

@italojs: I would prefer to have it integrated into Meteor core if possible. Maintaining a separate package would be difficult because it uses many internal files that are not exported and would need to be duplicated. In the meantime, it's possible to test it in any project by adding the ericmartin989:alternative-oplog and disable-oplog packages.

italojs:
Well, since you want to implement it in the core, let us support you on it, you are doing a really great job.

@ericm546 About the `observe sorted` test, I'm in the same step as you inside [change-streams observer-driver](https://github.com/meteor/meteor/pull/13787/files)

For `oplog - entry skipping`, since your implementation is oplog too, IMO we should maintain it, could you analyze and explain why it isn't pass or instant applicable for your observer?

In the case that it's not aplicable, you can skip it by checking which observer is being in the test:

``` js
// update your `_usesOplog = true` to `_usesNewOplog = true` or a better description, IMO we should uses the Class instance to compare, but lets follow the current pattern fo while
const isOldOplog = handle._multiplexer._observeDriver._usesOplog
if(!usesOplog) test.skip("message")
```

ericm546:
@italojs, the reason is that it doesn't have an entry-skipping mechanism.

When creating this new driver, I tried to adapt the code from redis-oplog into an Oplog driver while keeping the code logic as close as possible to the original. I did this knowing that redis-oplog already had good performance trade-offs and was well-tested.

The only performance optimization I added later, while doing performance testing, was to load documents in a batch when there was more than one document to load in the queue.

The original Oplog tailing driver's logic is quite different, and I'm not certain that adding an entry-skipping mechanism and reloading every query in the application when the queue reaches a certain number of elements would necessarily be a performance improvement. In some of my tests of the original driver, I often reached a point during a mass update where Meteor was stuck on one side processing all the oplog entries (as this cannot really be skipped), and then looping through a cycle of skipping the queue and reloading the full application.

If that makes sense, I can skip the "oplog - entry skipping" test. Should I also skip the "observe sorted" tests, or wait for now?

Also, right now I have to modify the code to test the new implementation. Should I add an environment variable or a switch to run the full test suite with the new driver?

italojs:
Great, I understand your point about `oplog - entry skipping`, but what about `observe sorted`? Why should we skip it?

PR #13910, opened 2025-08-28
Meteor-Rspack integration⚡🚀

nachocodoner:
Context: **[Meteor-RSPack Integration: A Modern Bundler Meets Meteor 3.4](https://forums.meteor.com/t/meteor-rspack-integration-a-modern-bundler-meets-meteor-3-4/63696)** (*Check out the forum post for additional information*)

This PR adds modern bundler support to Meteor by integrating Rspack alongside the existing bundler. This brings faster rebuilds, smaller bundles, and compatibility with the modern JS ecosystem.

The changes in this PR are significant, but I will keep it open even as we release betas. This way it can continue to be reviewed, with reviews addressed and delivered as part of the beta release plan until the official release. Since this is a major integration, it needs enough time to be properly verified, reviewed, and tested, without being tied to a strict release date. The only requirement for beta releases is to have all checks passing on the PR.

## Changes included

- **`rspack` package**: orchestrates Rspack setup, development server, and production builds
    - Context to review: [./packages/rspack](https://github.com/meteor/meteor/tree/modern-bundler-integration/packages/rspack)
- **`tools-core` package**: utilities for installing npm dependencies and detecting Meteor runtime to use it for Rspack integration and future tool integration packages (like CapacitorJS).
    - Context to review: [./packages/tools-core](https://github.com/meteor/meteor/tree/modern-bundler-integration/packages/tools-core)
- **`@meteorjs/rspack` npm package**: provides a default `rspack.config.js` for Meteor. Projects can add their own `rspack.config.js` to extend or override it.
    - Context to review: [./packages/meteor-rspack](https://github.com/meteor/meteor/tree/modern-bundler-integration/npm-packages/meteor-rspack)
- **Meteor core updates**: new changes to manage ignore rules, HTML boilerplate, entry points, and HMR integration
    - Context to review: [./tools](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools)
- **Test suite**: validates stability across Meteor, Rspack, and tools-core in all app setups using Jest and Playwright
    - Context to review: [./tools/modern-tests](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools/modern-tests)
- **Skeleton updates**: updated all Meteor skeletons to configure and use Meteor-Rspack integration.
    - Context to review: [./tools/static-assets](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools/static-assets)

## Getting started  

Once Meteor `3.4-beta.11` is available, enable Rspack with a single command:

```bash
[...]

PR #13915, opened 2025-09-02
Release 3.4 ⚡🚀 

nachocodoner:
Includes as the major change: [Meteor-Rspack integration⚡🚀](https://github.com/meteor/meteor/pull/13910).

More info in the forums: https://forums.meteor.com/t/new-3-4-beta-12-release-faster-builds-smaller-bundles-and-modern-setups-with-the-rspack-integration/64124

Please direct any reviews to that PR, not this release PR.

## Plan

Plan towards official release

- [x] **Launch first phase: 3.4-beta.8** ✅
    This beta delivers the first implementation of Meteor-Rspack integration, supporting all skeleton types and covering long-requested features and issues. It focuses only on Rspack changes, leaving other features for later betas. The goal is to keep changes isolated and avoid issue reports that could shift priority away from the modern bundler.
    **Focus:** *Stability*

- [x] **Launch second phase: 3.4-beta.14** ✅
    This beta expands the Meteor-Rspack integration to cover more project setups and modern configs based on reported needs. It is the chance to add configs for projects relying on Meteor-specific behaviors that lack a direct Rspack alternative. It will also review real-world projects using Meteor build compilers and specifics, and provide modern or fallback configs if any.
    **Focus:** *Extensibility*

- [x] **Launch third phase: 3.4-beta.14** ✅
    This beta includes fixes and new features for the Meteor-Rspack integration, plus other 3.4 additions like [dev-only packages](https://github.com/meteor/meteor/pull/13797) and [long-requested contributions that are ready to merge](https://github.com/meteor/meteor/milestone/122).
    **Focus:** *Contributions*

- [x] **Launch RC phase: 3.4-rc.4** ✅
    This RC will bring a final round of stability fixes. It will be the version closest to the official release, and if stable, it will become the final 3.4 release.
   **Focus:** *Official*


italojs:
For the modern-tests folder, what do you think we generate those projects at runtime? I see those applications can be outdated over time and it could be a problem for the tests

nachocodoner:
I'd rather get the reviews on https://github.com/meteor/meteor/pull/13910.

> For the modern-tests folder, what do you think we generate those projects at runtime?

Answering your question. We have two kind of tests: skeleton and modern app tests. [Skeleton tests](https://github.com/meteor/meteor/blob/modern-bundler-integration/tools/modern-tests/skeleton.test.js) are generated at runtime using `meteor create`. The modern app tests (eg. [react-router.test.js](https://github.com/meteor/meteor/blob/modern-bundler-integration/tools/modern-tests/react-router.test.js)) uses the apps statically described at [modern-tests/apps](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools/modern-tests/apps) to try configs not covered by skeletons, like dynamic imports, ESM, styling, Babel, swcrc configs, and more advanced setups, also those that will come in the future. [As in the legacy suite](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools/tests/apps), we preconfigure some apps for testing. I think this is better than trying to program every possible config and changing of the files to manage them at runtime.

>  I see those applications can be outdated over time and it could be a problem for the tests

Which problems do you see? We should keep those apps compatible in how they’re configured as tools evolve. If a test fails, that signals an issue to check and fix. It may reveal an incompatibility, and we should keep it as a test case. Similarly with legacy self-tests.

StorytellerCZ:
@nachocodoner any plans to add Node 24 or should I prepare update for Node 22?

nachocodoner:
> any plans to add Node 24 or should I prepare update for Node 22?

@StorytellerCZ: Node 24 will become LTS in October, so according to the release plan, it makes sense to start preparing for it in this release. Hopefully there are no breaking changes and the transition is smooth, as it has been lately.

boomfly:
It seems that meteor coffeescript package will not work with 3.4, or mb i do something wrong?

Снимок экрана 2025-09-11 в 16 09 13

nachocodoner:
@boomfly: have you followed the guide to add support of Coffeescript in your app code when using Rspack?

https://deploy-preview-13910.docs.meteor.com/about/modern-build-stack/rspack-bundler-integration.html#coffeescript

If you add support of Coffeescript using Rspack, the coffeescript meteor package is no longer needed, in favor of the delegation to a modern bundler. I could have managed Rspack/Coffeescript integration automatically as part the integration if I detect you use coffeescript atmosphere package. But I believe the right thing to do is to make this configurable and promote rspack.config.js usage for each project type. However, I may add a new skeleton for coffeescript for a future release (meteor create --coffeescript).

boomfly:
Add coffee config and next error. It rewrite absolute import to relative and then says that can't resolve that import:

Снимок экрана 2025-09-13 в 18 01 19

UPDATE:

Try different resolve configs, but not success:

```js
import { defineConfig } from '@meteorjs/rspack';

/**
 * Rspack configuration for Meteor projects.
 *
 * Provides typed flags on the `Meteor` object, such as:
 * - `Meteor.isClient` / `Meteor.isServer`
 * - `Meteor.isDevelopment` / `Meteor.isProduction`
 * - …and other flags available
 *
 * Use these flags to adjust your build settings based on environment.
 */
export default defineConfig(Meteor => {
  return {
    module: {
      rules: [
        {
          test: /\.coffee$/i,
          use: [
            {
              loader: 'swc-loader',
              // perserve SWC config in the Meteor project level
              options: Meteor.swcConfigOptions,
            },
            {
              loader: 'coffee-loader',
            },
          ],
        },

        // your other rules
      ]
    },
    resolve: {
      // Allow importing without extensions, in this explicit order
      extensions: ['.coffee', '.js', '.ts'],
      // Resolve `import './dir'` to `./dir/index.*`
      mainFiles: ['index'],
      // Allow extension-less ESM imports like `import './server'`
      fullySpecified: false,
      enforceExtension: false,
      preferAbsolute: true
    }
  };
});
```

nachocodoner:
@boomfly: https://github.com/meteor/meteor/tree/release-3.4/tools/modern-tests/apps/coffeescript

This is a sample Coffeescript app used for testing. Could you try to replicate it?

If that doesn’t work, please avoid using relative imports. I want to check if that might be the issue so I can reproduce and fix it later.

boomfly:
Can reproduce resolve issue: https://github.com/boomfly/meteor-3.4-rspack-resolve-issue

Снимок экрана 2025-09-15 в 20 37 14

nachocodoner:
@boomfly: Could you check if those issues you reported was solved in latest 3.4-beta.10 version?

boomfly:
> @boomfly: Could you check if those issues you reported was solved in latest 3.4-beta.10 version?

Yes, import issue resolved in beta.9 version.

Now server build and start, but client has strange error:

Снимок экрана 2025-09-24 в 13 50 16

nachocodoner:
> Yes, import issue resolved in beta.9 version.

I mentioned beta.10, but I released a quick fix on beta.11. Does this happen to you with beta.11? Also after `meteor reset`?

I was able to run [your repository without issues.](https://github.com/boomfly/meteor-3.4-rspack-resolve-issue)

boomfly:
It seems that project can run with beta.10. Solve issue with i18n undefined, that var imported at top of file: 

```js
import i18n from '/imports/startup/client/i18n'
```

and rspack ingnores other assignmets to that var in that file. I've removed that line and project started.

Is there a way to disable errors modal in DOM and show errors only in console? Cause for now firebase messaging is working, but same time throws error which not crash app.

Снимок экрана 2025-09-24 в 17 51 58

nachocodoner:
> Is there a way to disable errors modal in DOM and show errors only in console?

@boomfly: Yes. You should add the next config in your rspack.config.js file.

``` js
devServer: {
  client: {
    overlay: false, // disable the error modal overlay
  },
},
```

Btw, do you have still any problem in your end you would like to share [in your repository](https://github.com/boomfly/meteor-3.4-rspack-resolve-issue?rgh-link-date=2025-09-24T09%3A01%3A43Z)?

harryadel:
I'm attempting to do a migration of 2.x application 3.4 and I'll try to share any findings I've.

One of things I noticed right away is that the app doesn't crash after removing [`compat:graphql`](https://github.com/meteor-compat/meteor-graphql) 

rspack config contains this line though:
```
module: {
      rules: [
        {
          test: /\.gql$/,
          use: [
            {
              loader: 'graphql-tag/loader'
            }
          ]
        }
      ]
    },
```

I thought both are needed and only accidentally removed the meteor package and was surprised that the app didn't crash. Can you verify this @nachocodoner

nachocodoner:
@harryadel: If you keep the Rspack configuration, you don’t need the Meteor build plugin package anymore.

Since the app code is now handled by Rspack, it’s the only one responsible for processing files like .gpl. The Meteor package might still work in some cases, but it’s not guaranteed in all scenarios. If you need to process the .gpl files inside a Meteor Atmosphere package, you’ll likely need the meteor build-plugin-compat:graphql. However, it’s not required for your app’s direct code. [I mentioned replacing build plugins in the new docs](https://deploy-preview-13910.docs-online.meteor.com/about/modern-build-stack/rspack-bundler-integration.html#replace-build-plugins), but I forgot to suggest removing Meteor build packages once the handling is fully delegated to Rspack. Even by removing the Meteor package, the app's gpl will be still processed by Rspack, that is the reason for they to still work.

I adapted the [Apollo skeleton](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools/static-assets/skel-apollo) to use a config that ensures `.gql` files are detected (via @graphql-tools/webpack-loader). I'm not fully sure if I used the most stable build plugins from Rspack/Webpack, maybe your approach with `graphql-tag/loader` is more appropriate today. The app worked fine anyway. If you know of a more suitable or well-maintained GraphQL setup, please let me know so I can adjust the examples accordingly.

harryadel:
Hello Nacho, I was too quick to dive straight into action. Let me preface this first by saying how amazing this release would be and how grateful I'm to your contributions. This PR came in clutch to save the day!

The reason why this migration is taking place is that the current stakeholders were growing frustrated with the project build time and seriously considering switch over to Nextjs or something else since they already don't use Meteor APIs (Methods, Pub/sub) rather GraphQL. But considering the speed gains I'm noticing hopefully I'll be able to avert their decision. :crossed_fingers: 

Who knows how many other applications could've been saved if this was implemented sooner. So thank you for your efforts. This work that you've been doing is going to shape Meteor for many more years to come. :)


Now onto the nitty gritty stuff
>Since the app code is now handled by Rspack, it’s the only one responsible for processing files like .gpl. The Meteor package might still work in some cases, but it’s not guaranteed in all scenarios. If you need to process the .gpl files inside a Meteor Atmosphere package, you’ll likely need the meteor build-plugin-compat:graphql. However, it’s not required for your app’s direct code. [I mentioned replacing build plugins in the new docs](https://deploy-preview-13910.docs-online.meteor.com/about/modern-build-stack/rspack-bundler-integration.html#replace-build-plugins), but I forgot to suggest removing Meteor build packages once the handling is fully delegated to Rspack. Even by removing the Meteor package, the app's gpl will be still processed by Rspack, that is the reason for they to still work.


If I understand correctly, then this means I'd only need to use build-plugins if attempting to compile a file within a Meteor package? But other than that Meteor applications offset this problem to Rspack? I also didn't know about these docs, good thing that you mentioned it :+1: 

This is amazing news since it'd lessen the burden on both (the core team and Meteor developers) since it'd not be our problem anymore! 

>I adapted the [Apollo skeleton](https://github.com/meteor/meteor/tree/modern-bundler-integration/tools/static-assets/skel-apollo) to use a config that ensures .gql files are detected (via @graphql-tools/webpack-loader). I'm not fully sure if I used the most stable build plugins from Rspack/Webpack, maybe your approach with graphql-tag/loader is more appropriate today. The app worked fine anyway. If you know of a more suitable or well-maintained GraphQL setup, please let me know so I can adjust the examples accordingly.

Honestly I'm still making adjustments to the application to get it working and also fiddling with rspack configs to reach the best setup. Maybe once I'm done I'll share it.

harryadel:
```js
const { defineConfig } = require('@meteorjs/rspack');
// const { RsdoctorRspackPlugin } = require('@rsdoctor/rspack-plugin'); // Disabled - compatibility issues
const rspack = require('@rspack/core');

/**
 * Rspack configuration for Meteor projects.
 *
 * Provides typed flags on the `Meteor` object, such as:
 * - `Meteor.isClient` / `Meteor.isServer`
 * - `Meteor.isDevelopment` / `Meteor.isProduction`
 * - …and other flags available
 *
 * Use these flags to adjust your build settings based on environment.
 */
module.exports = defineConfig(Meteor => {
  return {
    // Disable caching - let Meteor handle it
    cache: false,
    // Only add externals for server-side builds to avoid bundling Node.js modules
    ...(Meteor.isServer && {
      externals: {
        // Node.js built-in modules
        'querystring': 'commonjs querystring',
        'url': 'commonjs url',
        'http': 'commonjs http',
        'https': 'commonjs https',
        'fs': 'commonjs fs',
        'path': 'commonjs path',
        'vimeo': 'commonjs vimeo',
        // Sharp and its dependencies
        'sharp': 'commonjs sharp',
        '@img/sharp-libvips-dev/cplusplus': 'commonjs @img/sharp-libvips-dev/cplusplus',
      },
    }),
    resolve: {
      extensions: ['.tsx', '.ts', '.jsx', '.js', '.json'],
      modules: ['node_modules'],
      symlinks: false, // Disable if not using symlinks for better performance
    },
    module: {
      rules: [
        {
          test: /\.gql$/,
          use: [
            {
              loader: 'graphql-tag/loader',
              options: {
                noDescription: Meteor.isProduction, // Remove descriptions in production for smaller bundle
              }
            }
          ]
        }
      ]
    },
    // Simplified optimization - let Meteor handle most of the chunking
    optimization: {
      // Enable tree shaking for unused exports
      usedExports: true,
      innerGraph: true,
      // Production minification
      minimize: Meteor.isProduction,
      ...(Meteor.isProduction && {
        minimizer: [
          new rspack.SwcJsMinimizerRspackPlugin({
            compress: {
              drop_console: true, // Remove console.logs in production
              passes: 2,
            },
          }),
          new rspack.LightningCssMinimizerRspackPlugin(),
        ],
      }),
    },
    // Performance optimizations
    performance: {
      hints: false, // Disable performance hints in development
    },
    // Enable source maps for better debugging - use 'source-map' instead of 'eval-source-map' for RsDoctor compatibility
    devtool: Meteor.isDevelopment ? 'source-map' : false,
    // Plugins for bundle analysis and optimization
    plugins: [
      // RsDoctor disabled - appears to have compatibility issues with Meteor 3.4-beta.11
      // Meteor.isDevelopment && Meteor.isClient && new RsdoctorRspackPlugin({
      //   linter: {
      //     rules: {
      //       'ecma-version-check': 'off',
      //     },
      //   },
      // }),
      // Compression for production builds (client only)
      Meteor.isProduction && Meteor.isClient && new rspack.CopyRspackPlugin({
        patterns: [],
      }),
    ].filter(Boolean),
    // Native CSS experiments
    experiments: {
      css: true,
    },
  };
});
```

Okidoki, so I had some time to fiddle around with rspack and here're my findings:

- `Multiple assets emit different content to the same filename client-rspack.js` error 

`splitChunks` configuration with multiple cache groups was conflicting with Meteor's internal entry point and chunk management system.

**Original Configuration** (INCOMPATIBLE):
```javascript
splitChunks: {
  chunks: 'all',
  maxInitialRequests: 25,
  maxAsyncRequests: 25,
  cacheGroups: {
    react: { /* ... */ },
    mantine: { /* ... */ },
    apollo: { /* ... */ },
    // ... 8 cache groups total
  }
}
```

- `@rsdoctor/rspack-plugin` package causes runtime errors when used with Meteor 3.4-beta.11:
      - `webpack-dev-server` crashes
      - Package.json resolution errors in RsDoctor core
      - Multiple deprecation warnings
      
- Use `devtool: 'source-map'` instead of `devtool: 'eval-source-map'`

- (`cache: { type: 'filesystem', ... }`) caused validation errors as Meteor expects a boolean value

nachocodoner:
Thanks for your time and the report @harryadel 

`splitChunks` is worth a closer look. The current Meteor-Rspack integration only emits chunks to the public folder via dynamic imports. I’ll see if I can make the setup more flexible for splitting.

~About externals: you don’t need them for built-in Node.js modules; with `target: node`, Rspack handles those.~ I think we need to set [`externalsPresets: { node: true }`](https://rspack.rs/config/externals#externalspresets). I’ll set it as the default so there’s no need to manually add each built-in module to externals in any app. For other deps like `sharp` or `vimeo`, marking them as external makes Rspack skip them so Meteor’s compiler resolves them. That’s useful only if Rspack compilation causes issues or there’s a Meteor/Rspack incompatibility; otherwise I wouldn’t externalize.

Regarding cache, did you disable it for performance or memory overflow, or because it didn’t react well to changes? I’m improving the default cache under `experiments.cache` to behave faster and more stable on edits.

Some of your configs are already defaults in Meteor-Rspack, like `experiments.css`, `optimization.usedExports` and others. If you enable the verbose setting below and remove your `rspack.config.js` overrides, you’ll see the default client and server configs:

``` json
{
  "meteor": {
    "modern": {
      "verbose": true
    }
  }
}
```

harryadel:
Nacho, I'm jumping here quick to tell you that you've outdone yourself with this release. This PR just keeps on giving. I'm migrating some other project to 3.x and instead of 3.3.2 I decided to to amp it to 3.4 and now thanks to your build changes. We no longer need [`fourseven:scss`](https://github.com/Meteor-Community-Packages/meteor-scss)!

This package has been an immense pain in the ass and I [remember how annoying](https://github.com/Meteor-Community-Packages/meteor-scss/pull/321) it was to migrate it to 3.0. This release frees Meteor developers from having to handle build packages ourselves and Meteor can finally piggyback off of the Nodejs ecosystem shedding lots of old code. Whew what a relief!! 

Ahem another rspack config in case you may need it :sweat_smile: 

```js
const { defineConfig } = require('@meteorjs/rspack');

/**
 * Rspack configuration for Meteor projects.
 *
 * Provides typed flags on the `Meteor` object, such as:
 * - `Meteor.isClient` / `Meteor.isServer`
 * - `Meteor.isDevelopment` / `Meteor.isProduction`
 * - …and other flags available
 *
 * Use these flags to adjust your build settings based on environment.
 */
module.exports = defineConfig(Meteor => {
  return {
    module: {
      rules: [
        {
          test: /\.tsx?$/,
          use: 'builtin:swc-loader',
          options: {
            jsc: {
              parser: {
                syntax: 'typescript',
                tsx: true,
              },
            },
          },
          type: 'javascript/auto',
        },
        {
          test: /\.scss$/,
          use: [
            'style-loader',
            'css-loader',
            'sass-loader',
          ],
          type: 'javascript/auto',
        },
        {
          test: /\.css$/,
          use: [
            'style-loader',
            'css-loader',
          ],
          type: 'javascript/auto',
        },
      ],
    },
    resolve: {
      fallback: {
        https: false,
        crypto: false,
      },
    },
  };
});

```

harryadel:
> Regarding cache, did you disable it for performance or memory overflow, or because it didn’t react well to changes? I’m improving the default cache under `experiments.cache` to behave faster and more stable on edits.

I disabled it because the build was getting stuck. Also, regarding cache and what not. Whenever I make a change to rspack or somethign I find it better to clear the old Meteor cache `meteor reset` or else I'd get errors! Should this be handled?

Grubba27:
@harryadel very nice work!!! Could you make a PR with those examples for this branch(release-3.4)?

I wonder, and I'll discuss with Nacho if we can have some "helpers" or something like it, so instead of  passing the rules (or remembering it) we could have some vite-style plugins:


```js
const { sass } = require('@meteorjs/rspack/sass');
    module: {
      rules: [
        {
          test: /\.tsx?$/,
          use: 'builtin:swc-loader',
          options: {
            jsc: {
              parser: {
                syntax: 'typescript',
                tsx: true,
              },
            },
          },
          type: 'javascript/auto',
        },
       ...sass() // this would be the same as writting the rules but we give it a "nice" api around it
      ],
```

nachocodoner:
@harryadel: Exactly. Relying more on the broader JS ecosystem will let us focus on what we enjoy most—building with Meteor. That’s one of the main goals behind all of this.

We already mention in the docs the direction toward using Rspack as an alternative for Meteor build plugins:

[Replace Build plugins](https://deploy-preview-13910.docs-online.meteor.com/about/modern-build-stack/rspack-bundler-integration.html#replace-build-plugins)

[CSS, Less and SCSS](https://deploy-preview-13910.docs-online.meteor.com/about/modern-build-stack/rspack-bundler-integration.html#css-less-and-scss)

Feel free to expand these sections if you think more details would help.

@Grubba27:

> so instead of passing the rules (or remembering it) we could have some vite-style plugins:

I’ve actually experimented with that recently. Take a look at this section:

With [Meteor.compileWithMeteor and Meteor.compileWithRspack](https://deploy-preview-13910.docs-online.meteor.com/about/modern-build-stack/rspack-bundler-integration.html#compilewithrspack-and-compilewithmeteor), you already have helper presets to build the config. Instead of remembering every setting, these presets automatically merge the right configuration into the final setup.

``` js
module.exports = defineConfig(Meteor => ({
  // Exclude native modules from the bundle (use Meteor runtime)
  ...(Meteor.isServer ? Meteor.compileWithMeteor(['sharp']) : {}),

  // Force-compile modern or local packages via SWC
  ...Meteor.compileWithRspack(['grubba-rpc']),
}));
```

I recently created another one `Meteor.setCache` to also automatically disable the rspack cache config with just setting it to `false`.

harryadel:
@nachocodoner Maybe it's time to update the description with a possible ETA?

packages/accounts-base/accounts-base.d.ts
```diff
@@ -387,8 +387,7 @@ export namespace Accounts {
   function _checkPasswordAsync(
     user: Meteor.User,
     password: Password
-  ): Promise<{ userId: string; error?: any }>
-}
```
ToyboxZach, line 391: This "}" got deleted, causing type checking to fail

StorytellerCZ:
There will be a Node security release tomorrow: https://nodejs.org/en/blog/vulnerability/december-2025-security-releases

I suggest we incorporate that release and then release.

nachocodoner:
@StorytellerCZ: Good point. I can either release an RC.3 later, or just include it in Meteor 3.4 official in January (expected to be announced then). Could you share the Node PR when available?

Also, I already released RC.2 to test a new functionality: https://github.com/meteor/meteor/pull/14049. This lets me move testing forward. That PR might still need a new release if something is wrong in the tool.

Adding options to `meteor update` is hard to test, but I changed the global Meteor version directly and it seemed to work. I still need to double-check it with the final published version.

StorytellerCZ:
@nachocodoner I'll get the PR for updating Node ready once I see that the release happened. Depending on the security issue we might need to do another 3.3 patch release if 3.4 will be released in January.

UPDATE: The security release for Node has been postponed till January 8th: https://nodejs.org/en/blog/vulnerability/december-2025-security-releases

nachocodoner:
Releasing the last RC.4 to validate CI releasing procedures before 3.4 official. Right after that, 3.4 official is planned to be released and recommended on January 30th, if everything goes well.

nachocodoner:
I can't believe it, I'm going to merge this at last! Long live the modernizations to come! 🎉

harryadel:
???????????? how is this even getting merged?? this is nothing short of a miracle!

PR #13878, opened 2025-08-01
Node 24.5.0 & NPM 11.5.1

StorytellerCZ:
https://github.com/nodejs/node/releases/tag/v24.5.0
Following from #13758, #13811 and #13856

PR #13856, opened 2025-07-17
Node 24.4.1

StorytellerCZ:
Following from https://github.com/meteor/meteor/pull/13758 and https://github.com/meteor/meteor/pull/13811
This PR brings Node to the July security release https://nodejs.org/en/blog/vulnerability/july-2025-security-releases

PR #13916, opened 2025-09-02
Replace http-proxy with http-proxy-3

StorytellerCZ:
http-proxy is not maintained, has security & performance concerns and uses deprecated API which throws when you start Meteor: #13491
I have found that [http-proxy-3](https://www.npmjs.com/package/http-proxy-3) is the most recent replacement package and went ahead and replaced it in the core. My app runs without the deprecation warning on the Meteor 3.4 branch, so I'm putting it here to run with the other tests (though I think new bundle might be necessary).
I think http-proxy-3 is even better option than forking it ourselves as was suggested in related discussions.

PR #13917, opened 2025-09-03
Node 24.8.0 & NPM 11.6.0

StorytellerCZ:
https://github.com/nodejs/node/releases/tag/v24.7.0
https://github.com/nodejs/node/releases/tag/v24.8.0
Following from #13758, #13811, #13856 and #13878


StorytellerCZ:
@nachocodoner it might be a good idea to merge https://github.com/meteor/meteor/pull/13916 into this and then publish the bundle.

StorytellerCZ:
Updated to Node 24.8.0 and NPM 11.6.0

Also updated some core (in the root `package.json`) dependencies minor versions and synced version in general.

We have talked about replacing eslint with Biome in the past. I will explore that in the future as that would ease some of the dependencies resolutions and reduce the number of dependencies in root. But that is for another PR.

PR #13919, opened 2025-09-04
Deny client-side user updates by default

harryadel:
https://github.com/meteor/meteor/pull/13865

## Release Notes

* **Security**
  * Changed default user account permissions to deny direct modifications. Users can no longer update their profile or account information directly from the application by default.

* **Documentation**
  * Updated to reflect new user modification permissions.

* **Tests**
  * Updated tests to verify the new authorization behavior.


StorytellerCZ (dismissed):

docs/source/api/accounts.md
```diff
@@ -149,16 +149,6 @@ more information is published for the currently logged in user,
 including access tokens. This allows making API calls directly from
 the client for services that allow this.
 
-Users are by default allowed to specify their own `profile` field with
```
italojs, line 152: could you please remove this session from docs v3 instead docs (v2)?

italojs (changes requested):
wasn't discussed in the previous PR, but it would be better to deprecate this feature instead of doing a hard delete on it, avoiding breaking changes, what do you think?

harryadel:
@nachocodoner @italojs ?

italojs (dismissed):

tools/tests/typescript.js
```diff
@@ -1,25 +1,25 @@
 // var selftest = require('../tool-testing/selftest.js');
 // var Sandbox = selftest.Sandbox;
-//
+
```
italojs, line 3: those commments line isnt necessaly

docs/source/api/accounts.md
italojs: it should be reverted from doc v2

italojs (changes requested):

italojs:
tests is failling 
```
C: tinytest - passwords - deny all updates by default failed: {"groupPath":["tinytest","passwords"],"test":"deny all updates by default","events":[{"sequence":794,"type":"fail","details":{"type":"true","not":true,"asyncBlock":6},"cookie":{"name":"passwords - deny all updates by default","offset":4,"groupPath":["tinytest","passwords"],"shortName":"deny all updates by default"}}]}
```

harryadel:
@italojs check now

italojs:
I'm focusing on 3.5
i'll ask for @nacho to review it ;)

harryadel:
@italojs That's a different nacho lol 
@nachocodoner

italojs:
autocomplete didnt work lol
@nacho you are welcome to review it too if you want XD

nachocodoner:
My usual approach with contributions is not to merge them directly without double-checking first. I typically check out the contributor’s branch, fork it into the core context, run the tests locally, provide feedback, and get familiar with the change. Combined with my focus on the Meteor 3.4.1 priorities around [Rspack](https://github.com/meteor/meteor/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20label%3Amodern-build-stack) and meteor/examples, that delayed my review, sorry.

I know this contribution was originally intended for 3.3.2, which I was not handling. In Meteor 3.4, I merged a large number of contributions and this one did not make it in. My main concern is that it changes behavior and is likely to introduce a breaking change for apps that rely on being able to update the profile field from the client side.

Shipping a breaking change in a patch release, whether 3.3.2 or 3.4.1, does not seem appropriate. Even if documented, this is the kind of change we should avoid in patch versions whenever possible, particularly in an area as fundamental as account management.

I agree with the goal of making the default behavior more secure (discussed [here](https://github.com/meteor/meteor/pull/13865#discussion_r2262795991)), but I think this should follow a deprecation path first. For example, we could introduce a deprecation warning in Meteor 3.4.1 indicating that denying these updates will become the default, and then apply the actual breaking behavior change in a later minor release, such as Meteor 3.5.

Alternatively, besides a deprecation warning, could this be controlled by a setting? That way, apps that want the more secure behavior could opt in early, while others would have time to adapt incrementally. Then, in Meteor 3.5, the new behavior could become the default for everyone, and even remove the temporary setting.

That would give developers advance notice inside their apps and time to adapt, instead of discovering a breaking change after a patch upgrade.

But, what do you think?

harryadel:
@nachocodoner I guess he simplest path is to point this to 3.5

So you may edit it like we did with the async hooks

nachocodoner:
We will base this on release 3.5 once we merge the contents of release-3.4.1 into devel, so we avoid resolving conflicts twice.

harryadel:
@nachocodoner @italojs You think this one is ready now?

italojs:
@coderabbitai full review

italojs:
is it causing some issue for wekan migration? 
I'm thinking to move it to 3.5.1 🤔

harryadel:
@italojs wekan?? What does wekan has do with it?

harryadel:
@italojs What's your final take on this PR? Any needed changes on my end?

v3-docs/docs/api/accounts.md
```diff
@@ -197,16 +197,6 @@ more information is published for the currently logged in user,
...
-Users are by default allowed to specify their own `profile` field with
-[`Accounts.createUser`](#Accounts-createUser) and modify it with
-`Meteor.users.update`. To allow users to edit additional fields, use
-[`Meteor.users.allow`](./collections.md#Mongo-Collection-allow). To forbid users from making any modifications to
-their user document:
-
-```js
-import { Meteor } from "meteor/meteor";
-Meteor.users.deny({ update: () => true });
-```
```
nachocodoner, line 209: I am still thinking that some apps may already rely on current default rules against their user objects, at least for the profile updates. I believe a new setting controlling this behavior would be a good way to document the breaking change. This would allow existing apps to keep using these allow rules when needed with the new setting, while ensuring new apps are more secure by default.

I can foresee new issues and report coming from this one easily. So having proper documentation and non-blocking migrations with a setting on a minor release (e.g Meteor 3.6) would be the best.

nachocodoner:
@StorytellerCZ: This PR introduces breaking changes, so it's not ready to be included in a Meteor patch release. After an internal discussion with @harryadel, we concluded that we will likely deliver it beyond 3.6, as this change can require migrations and prevent projects from benefiting from the major modernizations coming in the next releases.

PR #13923, opened 2025-09-07
Add JSON content compression in catalog

harryadel:
I've always been fascinated by the Meteor packaging system and how it works and thought opportunities to look into it and try to further improve it. So, I've tried to search for the lowest hanging fruit which is easy to implement and to also gauge your response/feedback.

This PR adds gzip compression to JSON content helping reduce the catalog size. Uncompressed data without the prefix continues to work unchanged. Also, may you choose to disable this feature using `METEOR_DISABLE_CATALOG_COMPRESSION` env var.


tools/packaging/catalog/catalog-remote.js
```diff
@@ -435,6 +439,41 @@ Object.assign(Table.prototype, {
     self._deleteQuery = "DELETE FROM " + self.name + " WHERE _id=?";
   },
 
+  // Compress a JSON string for storage
+  _compressString: function(str) {
+    var self = this;
```
italojs, line 444: Do you mind using only this, please?

tools/packaging/catalog/catalog-remote.js
```diff
@@ -885,10 +943,18 @@ Object.assign(RemoteCatalog.prototype, {
...
     return _.map(rows, function(entity) {
-      return JSON.parse(entity.content);
+      // Decompress content before parsing JSON
+      var decompressed = self._decompressString(entity.content);
+      return JSON.parse(decompressed);
     });
   },
 
+  // Helper method to decompress content (delegates to table method)
+  _decompressString: function(str) {
```
italojs, line 953: 1. I don't see a necessity for a new function to pass the param to `this.tableVersions._decompressString`
2. Correct me if I'm wrong, but is this method duplicated with line 461?

tools/packaging/catalog/catalog-remote.js
```diff
@@ -435,6 +439,41 @@ Object.assign(Table.prototype, {
...
+    } catch (e) {
+      // If compression fails, fall back to uncompressed
+      Console.debug('Catalog compression failed, using uncompressed data:', e.message);
+      return str;
+    }
+  },
+
+  // Decompress a string that might be compressed
+  _decompressString: function(str) {
+    if (typeof str === 'string' && str.startsWith('GZIP:')) {
```
italojs, line 462: To improve readability, would you mind inverting this condition? It will remove one indentation level, as it did in. `_compressString`

``` js
if(!(typeof str === 'string' && str.startsWith('GZIP:'))) return str
try {
....
} catch (e) { 
... 
}
```

tools/packaging/catalog/catalog-remote.js
```diff
@@ -435,6 +439,41 @@ Object.assign(Table.prototype, {
...
+      return str;
+    }
+    
+    try {
+      var compressed = zlib.gzipSync(Buffer.from(str, 'utf8'));
+      // Mark as compressed with a prefix to distinguish from uncompressed data
+      return 'GZIP:' + compressed.toString('base64');
+    } catch (e) {
+      // If compression fails, fall back to uncompressed
+      Console.debug('Catalog compression failed, using uncompressed data:', e.message);
```
italojs, line 455: Console should be lowercase -> console

tools/packaging/catalog/catalog-remote.js
```diff
@@ -435,6 +439,41 @@ Object.assign(Table.prototype, {
...
 
+  // Compress a JSON string for storage
+  _compressString: function(str) {
+    var self = this;
+    if (!self._compressContent) {
+      return str;
+    }
+    
+    try {
+      var compressed = zlib.gzipSync(Buffer.from(str, 'utf8'));
```
italojs, line 450: Since we are using a sync version, I would like to have a benchmark in the big/small machine, and for big/small payloads to know how it impacts the I/O since it's a blocking code

italojs (changes requested):

zodern:
I think there is a larger issue with this change. The sqlite database is shared between all Meteor versions, so data in it has to remain compatible with any version of Meteor that might possibly be run. For example, a user could maintain both apps on Meteor 2 and 3, and the Meteor 2 release needs to be able to read rows in the database that were inserted by Meteor 3.

However, the sqlite database is stored in a versioned folder (currently v2.0.1). It would be possible to create a separate folder  with the new database format, and older versions of Meteor can use the older database they are compatible with.

Though if we're doing this, there's a lot of other changes I think we should consider when making a new version of the database (for example, there's a lot of unused data, or data not strictly needed locally/offline, and there might be other schemas that offer better performance, such as putting the most frequently accessed data in its own columns, or if we can allow sqlite to do more of the work). I'm also curious how much using gzip helps or hurts performance.

harryadel:
@zodern You're right to bring up the versioned directory and possible multi-usage where multiple Meteor versions would attempt to access the same local SQLite.

The natural conclusion would be to create a new version directory `v3.0.0` that'd house any future improvements we would want to make. By doing so we'd remain 100% backwards compatible and wouldn't have to worry about breaking any users' DBs. 

The downside is that it'd lead to an increase in the disk usage on users devices! Basically users would've two version directories for the foreseeable future.

I'm not saying this is entirely a bad thing rather illustrating the trade-offs that come with such decision. I'd like to hear your thoughts on it, because if so. We can readily explore it :+1:

Issue #13866, opened 2025-07-26
Is guide up to date?


harryadel:
Hello there as I was reading into the [official guide](https://guide.meteor.com/security), I noticed it supports Meteor version up to 2.16. Is this proper? Is it compatible with 3.0? Also it seems under the old theme and not the new v3 look https://docs.meteor.com/

<img width="467" height="265" alt="Image" src="https://github.com/user-attachments/assets/851f15af-e5c7-4f30-9563-504d3097cbed" />

StorytellerCZ:
No, it is not. We have to fix that.
Looks like a good task to finish before Meteor Impact.

harryadel:
Just to be clear, it's not just about the theme but the whole guide needs to be brought in line with 3.0

wreiske:
See https://forums.meteor.com/t/meteor-phoenix-rising-from-the-ashes-to-spark-new-developer-passion/63339/19?u=wreiske and https://forums.meteor.com/t/meteor-phoenix-rising-from-the-ashes-to-spark-new-developer-passion/63339/20?u=wreiske

- [x] https://github.com/meteor/meteor-theme-hexo/pull/99
- [x] https://github.com/meteor/meteor/pull/13714

These have been around for some time now and would LOVE it if someone could push these out...

nachocodoner:
Yes, having back the guides is important. We’ve only included a few [React](https://docs.meteor.com/tutorials/react/index.html) and [Vue](https://docs.meteor.com/tutorials/vue/meteorjs3-vue3-vue-meteor-tracker.html) guides in the docs as tutorials, but there are many other topics that need attention and could be split into separate sections. We should review them, but we can expect the same content with only minor updates to meet new expectations.

I’d like to help, but my workload is full now. I plan to update the guides I’m covering, like bundler and future mobile topics (so Cordova guide won't be necessary to migrate, [we just added a quick guide](https://docs.meteor.com/about/cordova.html)).

PRs are welcome.

vparpoil:
I was wondering which repo should be worked on for updating the guide ?  @italojs @nachocodoner 
The one which seems the original source is archived : https://github.com/meteor/guide/

nachocodoner:
[...]

PR #13926, opened 2025-09-11
Docs: bring guide on Packages to Meteor v3 docs

vparpoil:
Following the discussion in https://github.com/meteor/meteor/issues/13866 , this PR adds to the docs the tutorial on using and writing Atmosphere packages.
I’ve tried to keep the modernization of the original texts in separate commits so that the proposed changes are easy to review.

I chose not to include [Using and writing npm packages](https://guide.meteor.com/using-npm-packages) in the Meteor v3 docs, as I believe it’s no longer relevant in this context.

The PR is not fully complete yet, since some links (mostly those related to the build system) are still broken.

I’d really appreciate feedback from the Meteor team on this PR so we can move forward with improving how the guide articles are presented in the new documentation.


nachocodoner:
Great to see this PR!

This is open to discussion, but what do you think about making "Atmosphere Packages", or even Packages, its own section within "Quick Start"? I see Packages as a core topic in Meteor: they add features and tie into the bundler stack. "Quick Start " already has "Web Apps", "Modern Build Stack", and "Cordova", and "Atmosphere Packages " could fit there. Another option is a new category called Features that groups "Modern Build Stack", "Atmosphere Packages", and "Cordova", instead of keeping them in Quick Start.

For "Packages", most docs could live in a single file. The side nav can highlight subsections like “When to use them,” “Using Atmosphere Packages”, and “Writing Atmosphere Packages”.

If we add the following at the top of the markdown file, we can also customize the top right nav and the levels shown.

``` markdown
---
outline:
  level: [2, 3]
---
```

What do you think?

nachocodoner:
> I chose not to include [Using and writing npm packages](https://guide.meteor.com/using-npm-packages) in the Meteor v3 docs, as I believe it’s no longer relevant in this context.

We could add this later if we think is useful. Beginners may appreciate this distinction and still they can install them. It's good to have this PR just focused on "Atmosphere Packages".

vparpoil:
Thanks for your feedback, yes there need to be an open discussion about this : the left nav bar already has **Packages** as a main entry point plus the **Community Packages** one (which is also present inside **Packages**), so I think we don't need to add another entry point in **Quick Start**.
 
Also, packages are listed alphabetically which is not very intuitive to newcomers. Sorting by uses cases would be better for them.

I guess we could try to re-think this left menu architecture to be more straightforward.

What about something more focused on uses cases like this : 
```
- Packages 
  - Introduction to Atmosphere packages
  - Core Packages and Meteor maintained 
    - UI framework compatibility
        - react-meteor-data
        - etc..
    - DX and tooling
        - audit-argument-check
        - fetch
        - markdown
        - etc...
    - Users and security
        - accounts-ui
        - accounts-2fa
        - roles
        - oauth-encryption
        - etc...
  - Community maintained 
[same structure]
```

nachocodoner:
Actually, that makes more sense. I agree it fits better to reuse that section. I won’t use titles that long to keep consistency with the patterns already used in the docs. I’ll also try to minimize nested levels where possible, but grouping makes sense to not overload information.

vparpoil:
Here is a new proposal with the tutorial on the Overview part of Packages, re-organization of the left menu packages and removal of deprecated packages.

Capture d’écran 2025-09-12 à 14 38 26

I'm in favor of removing `accounts-ui` from the left menu because React is now the default. Maybe Blaze related packages should be grouped together in a sub page

italojs (commented):
lgtm, loved the refac into left menu

italojs (approved):

PR #13927, opened 2025-09-12
Use concurrency-safe iteration in Minimongo async methods

graemian:
As per discussion at https://github.com/meteor/meteor/issues/13864#issuecomment-3243131324

There are a few solutions to this problem, but I think this is probably the best we have right now. The others are discussed in the issue above.


italojs (approved):

italojs:
we can land in 3.4.1 maybe

PR #13941, opened 2025-09-23
Fix #13937 make ParamTable always render so Chrome search can detect text 

### LINKED ISSUE #13937: New v3 docs search regression / UX pain point
Today I was helping a new developer with how to auto logout a session. I knew there was a login expiration somewhere in the Accounts config, so we did a quick google search:

Image

Great! AI summary shows `loginExpirationInDays` and there is a link to the accounts documentation: https://docs.meteor.com/api/accounts

After landing on the page, instinctively we hit "CTRL-F" and pasted to find the appropriate documentation:
Image

No results found. What?

I said, "ok - use the search at the top instead of CTRL-F" and that found this:

Image

But I _still_ can't find `loginExpirationInDays`.

https://github.com/user-attachments/assets/f595ee9e-0a83-4820-9415-b11bb2f0bc3d

Ahh! It's *hiding* behind this collapse section inside https://docs.meteor.com/api/accounts.html#AccountsCommon-config

https://github.com/user-attachments/assets/833ff08a-9dc6-451d-baa0-4f8c4de3e629

I wanted to share my experience and maybe push to show these by default, or change the functionality to show the options if you click from search. Either way - _something_ needs addressed with this UX.

paritoshdey-dev:
Here's the screenshot of the local test:

image


italojs (dismissed):

Grubba27:
Not sure if I liked this.... It makes the docs look too bloated. Maybe we can think of something else. Hijacking/listening to what is being searched idk.

@italojs https://deploy-preview-13941.docs-online.meteor.com/api/collections.html#Mongo-Collection

When hitting the cmd+k search, it could open the args tables; 

One thing that could help the crawler is to make it open by default to crawlers

wreiske:
It's important to link the original issue with documentation on why the current experience ~sucks~ is not great. https://github.com/meteor/meteor/issues/13937

italojs:
It's bloated but is better than an unfindable doc 
Unfortunately, nobody reads the docs with a high level of attention, usually we just do a few ctrl+f or ask for AI in this new AI era 🤷🏻‍♂️
Another thing is SEO, as we can see here, collapsed docs can't be readable by google's engine
I just see benefits here

italojs (approved):

Grubba27:
I still think that there is too much information out of the box with this, but it can be fine....

As a future improvement, we could have it stored in local storage, with the user's preference on whether the table should be shown by default or not.

For crawlers(AI, Regular Crawlers, and Algolia), it could be true by default idk

Issue #13940, opened 2025-09-23
Bug: Navbar fixed at top while scrolling pages navbar no longer visible !


AayushSahani01:
## Description:
**When scrolling the website, the navigation bar also scrolls up and disappears.** 
---

**What steps did you take and what happened:**

- [x] Open the site in browser.

- [x] Scroll down the page.

- [x] Notice that the navbar moves up and is no longer visible.

---

**What did you expect to happen:**
### Expected Behavior:

- The navbar should remain fixed at the top of the viewport while scrolling.

### Actual Behavior:

- The navbar scrolls along with the page content, making navigation less accessible

---
 
**Environment:**

- Browser: [e.g., Chrome 140.0, Firefox 118.0]

- OS: [ Windows 11]

- Screen size: [1920x1080]

StorytellerCZ:
Which site?

AayushSahani01:
hii @StorytellerCZ ,

This site: https://www.meteor.com/

TanmayAggarwal87:
I have fixed this issue PR #13943

dupontbertrand:
I think the navbar is working as expected on meteor.com 👍 
@TanmayAggarwal87 Also, this repository hosts Meteor core, so changes here won’t affect the marketing site. The linked PR edits a deprecated Bootstrap CSS file using the IE7 star-hack (*position), which modern browsers ignore, so it wouldn’t change the behavior anyway

AayushSahani01:
### Hi @StorytellerCZ ,

- Can you assign to me ?

PR #13943, opened 2025-09-24
fix: Navbar is fixed now , Issue: #13940

TanmayAggarwal87:
Changed the Navbar class in bootstrap.css file the navbar is fixed on top now


italojs:
do you have any proof of the frontend working? the tests is broken

PR #13944, opened 2025-09-30
DEV: Flip `--raw-logs` to default true and add `--timestamps` option.

Grubba27:
Screenshot 2025-09-29 at 22 30 40

This _for_ sure will make the regular dev xp better; this could arguably be a breaking change _but_ I really want to believe that people did not rely on this timestamp for their logs (I assume people are using [the logging package](https://docs.meteor.com/packages/logging.html))

Screenshot 2025-09-29 at 22 34 33
Screenshot 2025-09-29 at 22 35 12

relates to: https://forums.meteor.com/t/meteor-development-tiny-tip-that-helped-me-a-lot/64156/4?u=grubba

### DIFF
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -536,7 +536,7 @@ async function doRunCommand(options) {
     );
   }
 
-  runLog.setRawLogs(!options['raw-logs'] && options.timestamps);
+  runLog.setRawLogs(options['raw-logs'] && !options.timestamps);
 
   let webArchs = projectContext.platformList.getWebArchs();
   if (! _.isEmpty(runTargets) ||
@@ -2259,7 +2259,8 @@ async function doTestCommand(options) {
     serverArchitectures.push(DEPLOY_ARCH);
   }
 
-  runLog.setRawLogs(!options['raw-logs'] && options.timestamps);
+  runLog.setRawLogs(options['raw-logs'] && !options.timestamps);
+
 
   var includePackages = [];
   if (options['extra-packages']) {
--- a/tools/cli/commands.js
+++ b/tools/cli/commands.js
@@ -451,7 +451,8 @@ var runCommandOptions = {
     ...inspectOptions,
     'no-release-check': { type: Boolean },
     production: { type: Boolean },
-    'raw-logs': { type: Boolean },
+    'raw-logs': { type: Boolean, default: true },
+    timestamps: { type: Boolean, default: false }, // opposite of --raw-logs
     settings: { type: String, short: "s" },
     verbose: { type: Boolean, short: "v" },
     // With --once, meteor does not re-run the project if it crashes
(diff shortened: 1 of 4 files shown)


harryadel:
Thank you @Grubba27 :tada:

StorytellerCZ:
Yes for this and the logging package will need improvement in general. I would like to imagine it in the future that it could connect to outside logging services as well.

zodern:
One benefit from the timestamp is it was obvious what logs were from the meteor-tool or build plugins, and what logs were from the app. For example, in the screenshot of the error, it's not obvious if that error happened while building the app, or when running the app. Would there be any value in keeping some sort of indicator/prefix for app logs (or maybe instead having a prefix for the meteor-tool/build-plugin logs)?

> I assume people are using [the logging package](https://docs.meteor.com/packages/logging.html)

Maybe other developers have had a different experience, but I've never personally seen an app that directly used the logging package.

harryadel:
> Maybe other developers have had a different experience, but I've never personally seen an app that directly used the logging package.

Me neither.

Grubba27:
For galaxy users, the logging package makes their logs look good: 

check this forum thread: https://forums.meteor.com/t/suggestions-for-logging-in-galaxy/63119

Grubba27:
Would there be any value in keeping some sort of indicator/prefix for app logs (or maybe instead having a prefix for the meteor-tool/build-plugin logs)?

I think having a better-defined prefix for the meteor-tool would be better. Currently, we use =>, but I think if we added some color, emoji (maybe 🚀 / ☄️ ) or something to define the tool would be better. Also, our tool should behave a bit more like Vite and others (in the sense of not clogging the logs but actually replacing text)


cc @nachocodoner

italojs:
> Would there be any value in keeping some sort of indicator/prefix for app logs (or maybe instead having a prefix for the meteor-tool/build-plugin logs)?

A prefix would have the same issue as timestamp: [use case](https://forums.meteor.com/t/meteor-development-tiny-tip-that-helped-me-a-lot/64156?u=italojs)

We can set it as optional or improve what we have today as you mentioned, color, emoji..., but I don't see any value on adding prefix `docker-compose like` here since that log is more about Meteor's stuff, am I right? I just see a prefix being used by the application, but the dev can do it

StorytellerCZ:
For a very long time the logging package was not documented until I took interest in it and wrote the documentation. This is why not so many people use it. I think it deserves much more love (as in usage and development).

https://docs.meteor.com/packages/logging.html

Grubba27:
Windows checks are failing 😭

nachocodoner:
Great to see this change moving forward. Not sure why Windows is failing. We need to investigate, it seems related to environment setup, not to these changes.

nachocodoner (approved):

italojs (approved):

PR #13948, opened 2025-10-01
Node v24.13.1 and NPM v11.6.2

StorytellerCZ:
Following from #13758, #13811, #13856, #13878 and #13917

**NodeJS changelogs**
https://github.com/nodejs/node/releases/tag/v24.7.0
https://github.com/nodejs/node/releases/tag/v24.8.0
https://github.com/nodejs/node/releases/tag/v24.9.0
https://github.com/nodejs/node/releases/tag/v24.10.0
https://github.com/nodejs/node/releases/tag/v24.11.0
https://github.com/nodejs/node/releases/tag/v24.12.0
https://github.com/nodejs/node/releases/tag/v24.13.0
https://github.com/nodejs/node/releases/tag/v24.13.1

**Migration guide:** https://nodejs.org/en/blog/migrations/v22-to-v24

**NPM changelog**
https://docs.npmjs.com/cli/v11/using-npm/changelog

**MongoDB**
I have also noticed that none of the patches to MongoDB v7 have been applied. I have created an issue and a PR on the relevant repository to get things moving, but that is out of my permissions to get rolling. MongoDB v8 would be nice, but at least the patches for this year would be nice.
On this topic it might be good idea to put on a roadmap a way to figure out how to integrate MongoDB without the extra build step and use something official from MongoDB to make this release step easier.
In general making the bundle/release process easier or more automated would be of benefit.

**Of note**
#13916 is also related to this as it should be in the bundle, so it should be merged in here or if @nachocodoner doesn't want to upgrade to Node 24, I can prepare update to the latest Node 22 and we can include that fix there.


nachocodoner:
I think we’ll include the Node 24 major update in a later release. The reason is that it brings several breaking changes, especially the new AsyncLocalStorage context, which could affect how Meteor runtime currently works, in terms of the tool, but also DDP. There are also updates related to npm 11 scripts and other runtime behavior, though those are less likely to cause issues.

Meteor 3.4 focuses mainly on build-time improvements, so most checks and validations are centered around that area. However, Node 24 introduces many runtime changes. Since Meteor 3.5 will target runtime performance improvements and we will run bunch of benchmarks, that will be a better time to evaluate how Node 24 impacts things, if at all. Also, our dev_bundle and CI pipeline need to be adapted to Node 24. In the past, we had differences between Node 22 and earlier versions, so it’s best to allocate extra time for this work to avoid delaying the 3.4 release further.

For now, let's focus on upgrading to Node 22.x for 3.4. This way, we can deliver a stable version that includes the major bundler changes without introducing runtime risks. Any runtime issues could block many projects on get the best Meteor bundler experience in years 😄🚀.

@StorytellerCZ: could you share an update on the latest Node 22.x series?

cc @italojs

StorytellerCZ:
@nachocodoner @italojs I have a branch for Node 22 update. Node v22.20.0 is the latest. Here is the PR:
https://github.com/meteor/meteor/pull/13992

StorytellerCZ:
I've updated this PR to latest node 24.10.0
So I'm guessing we could use this PR as a base for the 3.5 release branch?

StorytellerCZ:
NodeJS v24.11.0 transitions to LTS.
https://github.com/nodejs/node/releases/tag/v24.11.0

StorytellerCZ:
NodeJS just released a series of migration guides, here is the one relevant for us: https://nodejs.org/en/blog/migrations/v22-to-v24

StorytellerCZ:
Continued in #14170

